Issue #638, opened 2020-05-04
Error model


kvark:
Our current error model is a bunch of `assert!` calls in the function bodies. This *can* be made safe (in Rust's sense of safety), but it's not matching the upstream spec, and it's not desirable as a model exposed to users.

The upstream has a model like "internal contagious nullability", where things. at some point after they are used wrong, become internally invalid. They spread that invalid state onto things that use them, and so forth. The user would then get an error scope asynchronously with the exact problem.

I wonder if we can start reshaping the current model to propagate errors up to the user instead of asserting. This means the native users will get the errors right away, exactly in the way they expect a Rust library to error. This shouldn't block us in implementing the error scopes API for the Web specifically, it's just going to be unused (or underused) when compiling for native.

kvark:
Thinking about this more: at the lowest level, the errors are synchronous (gfx-hal, validation on the server). At the highest level, errors are deferred (the CIN model of WebGPU API). There is a layer in the middle where we'd start to deferring errors, therefore. Once we do that, all the higher levels are left with no choice.

One option is to handle all errors in wgpu-core and return `Result` things to Rust users. That would mean - Gecko, Servo, and wgpu-native would need to implement the error model on their sides. If there is too much code/logic duplication between them, it would make sense to move it in wgpu-core and have the errors deferred there.

kvark:
Marking for 0.6 release. We've already started converting the methods to return errors, e.g. #706.
The idea is that this is the most sensible behavior for wgpu-core, and the lazy WebGPU semantics can and should be implemented on top. More over, the mechanism of error propagation in browsers is very specific to implementation, so having Servo and Gecko implement it separately is needed anyway.

kvark:
One unknown here is how exactly the browsers will handle it, and how they will associated IDs with something that is in an "error" state, and thus failed to even create. @kunalmohan has hit this problem now in Servo, and I've been thinking on the ways forward. We figured out that in a multiprocess implementation, the client doesn't even need to know if something is in an error state, unless asked specifically by the user. So everything should be handled by the server. Here are some ideas on how to do it:

## 1. Invalid Id errors

[...]

PR #813, opened 2020-07-18
Add buffer creation error type

GabrielMajeri:
**Connections**
Part of #638 

**Description**
Adds an error type for the buffer creation functions.

**Testing**
Checked with core and ran test on player.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -680,24 +680,20 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             resource::BufferUse::EMPTY
         } else if desc.usage.contains(wgt::BufferUsage::MAP_WRITE) {
             // buffer is mappable, so we are just doing that at start
-            match map_buffer(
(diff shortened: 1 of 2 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -699,6 +694,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                     };
                 }
                 Err(e) => {
+                    // TODO: return error on failure?
                     log::error!("failed to create buffer in a mapped state: {:?}", e);
```
GabrielMajeri, line 698: I cannot think of a scenario when we couldn't map a buffer at creation time and then it would work later. So I'm thinking it's better to return an error here :thinking:?

GabrielMajeri:
It seems the checks have been cancelled due to an outage: https://www.githubstatus.com/

wgpu-core/src/device/mod.rs
kvark: if it can't happen, let's just assert on that

PR #814, opened 2020-07-18
Error type for texture creation

GabrielMajeri:
**Connections**
Part of #638 

**Description**
Creates an error type for texture creation functions.

**Testing**
Checked with core and player.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -55,7 +55,7 @@ fn own_label(label: &Label) -> String {
 }
 
 pub const MAX_COLOR_TARGETS: usize = 4;
-pub const MAX_MIP_LEVELS: usize = 16;
+pub const MAX_MIP_LEVELS: u32 = 16;
 pub const MAX_VERTEX_BUFFERS: usize = 16;
 pub const MAX_ANISOTROPY: u8 = 16;
 pub const SHADER_STAGE_COUNT: usize = 3;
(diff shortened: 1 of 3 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -3035,3 +3034,14 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+
+#[derive(Clone, Debug, Error)]
+pub enum CreateTextureError {
+    #[error("D24Plus textures cannot be copied")]
+    CannotCopyD24Plus,
+    #[error(
+        "texture descriptor mip level count ({0}) must be less than device max mip levels ({})",
+        MAX_MIP_LEVELS
+    )]
+    InvalidMipLevelCount(usize),
```
kvark, line 3046: we should use the same type we use for specifying the mipmap level count in the descriptor (`u32`?)

wgpu-core/src/device/mod.rs
```diff
@@ -53,7 +54,7 @@ fn own_label(label: &Label) -> String {
 }
 
 pub const MAX_COLOR_TARGETS: usize = 4;
-pub const MAX_MIP_LEVELS: usize = 16;
+pub const MAX_MIP_LEVELS: u32 = 16;
```
GabrielMajeri, line 57: Converted this to a `u32`, to match the type in the descriptor. It makes more sense to cast this to an `usize` in the texture state storage in `track/texture.rs`

PR #816, opened 2020-07-18
Error types for buffer mapping API

GabrielMajeri:
**Connections**
Part of #638 

**Description**
Adds safe error handling for buffer mapping/unmapping functions.

**Testing**
Checked with core, and ran tests with player.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -9,7 +9,8 @@ use crate::{
     hub::{GfxBackend, Global, GlobalIdentityHandlerFactory, Hub, Input, Token},
     id, pipeline, resource, span, swap_chain,
     track::{BufferState, TextureState, TrackerSet},
-    validation, FastHashMap, LifeGuard, MultiRefCount, PrivateFeatures, Stored, SubmissionIndex,
+    validation::{self, check_buffer_usage, MissingBufferUsageError},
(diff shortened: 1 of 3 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -2905,14 +2898,16 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
         device.lock_life(&mut token).map(buffer_id, ref_count);
+
+        Ok(())
     }
 
     pub fn buffer_get_mapped_range<B: GfxBackend>(
         &self,
         buffer_id: id::BufferId,
         offset: BufferAddress,
         _size: Option<BufferSize>,
```
GabrielMajeri, line 2909: I just noticed this argument, what is it for? Is the intention to check that the mapped range has a valid size?

kvark: we need to check and validate it, but we aren't using it in native implementation. A mapped range is just a pointer.
The parameter came from the WebGPU API spec, where it's needed to know how much to synchronize between content and GPU processes

kvark:
please rebase

kvark:
bors +

GabrielMajeri:
Rebased, should be merge-able now.

I've learned the compromise of many small PRs: easier to review, much more merge conflicts :sweat_smile:

PR #817, opened 2020-07-18
Safe error handling in instance module

GabrielMajeri:
**Connections**
Part of #638.

**Description**
Replaces some assertions and unwraps in `instance.rs` with error types.

**Testing**
Checked with core.

### DIFF
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -250,10 +250,6 @@ pub enum RequestDeviceError {
     NoGraphicsQueue,
     #[error(transparent)]
     DeviceCreationError(#[from] hal::device::CreationError),
-    #[error("adapter storage buffer offset alignment not compatible with WGPU")]
-    IncompatibleStorageBufferOffsetAlignment,
-    #[error("adapter uniform buffer offset alignment not compatible with WGPU")]
-    IncompatibleUniformBufferOffsetAlignment,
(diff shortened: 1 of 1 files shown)


wgpu-core/src/instance.rs
```diff
@@ -685,20 +678,16 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
-                0,
-                BIND_BUFFER_ALIGNMENT % limits.min_uniform_buffer_offset_alignment,
-                "Adapter uniform buffer offset alignment not compatible with WGPU"
-            );
+            if BIND_BUFFER_ALIGNMENT % limits.min_storage_buffer_offset_alignment != 0 {
+                return Err(RequestDeviceError::IncompatibleStorageBufferOffsetAlignment);
+            }
+            if BIND_BUFFER_ALIGNMENT % limits.min_uniform_buffer_offset_alignment != 0 {
+                return Err(RequestDeviceError::IncompatibleUniformBufferOffsetAlignment);
+            }
```
GabrielMajeri, line 690: I'm not sure if it's worth keeping two different variants for these types, or if I should merge them into a single generic one.

My reasoning is: the user can't do much anyway if their device doesn't support WebGPU alignments.

kvark: actually, just turn them back into assertions. It's *not* expected to happen, it means the specification is wrong.

PR #819, opened 2020-07-19
swap_chain_present returns SwapChainStatus

rukai:
**Connections**
Needed by https://github.com/gfx-rs/wgpu-rs/pull/452

**Description**
For `swap_chain_present` moves the PresentError cases out of the Err and into the Ok.
This means we can `.unwrap()` the result in wgpu and it will only panic on an error that the user cant handle.

**Testing**
Ran the wgpu-rs examples.

### DIFF
--- a/wgpu-core/src/swap_chain.rs
+++ b/wgpu-core/src/swap_chain.rs
@@ -42,7 +42,7 @@ use crate::{
 };
 
 use hal::{
-    self, device::Device as _, queue::CommandQueue as _, window::PresentError,
+    self, device::Device as _, queue::CommandQueue as _,
     window::PresentationSurface as _,
 };
 use thiserror::Error;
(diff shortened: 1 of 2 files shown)


wgpu-core/src/swap_chain.rs
```diff
@@ -245,14 +243,21 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             }
         }
 
-        Ok(())
+        Ok(match result {
+            Ok(None) => (SwapChainStatus::Good),
+            Ok(Some(_)) => (SwapChainStatus::Suboptimal),
+            Err(err) => match err {
+                hal::window::PresentError::OutOfMemory(_) => SwapChainStatus::OutOfMemory,
```
kvark, line 250: some of these are not really recoverable by wgpu-rs users
`OutOfMemory` and `DeviceLost` are similar to other places where this can happen, and we don't currently provide any way to handle these gracefully.
So maybe we should really keep the statuses for those things that belong to recoverable errors? I think `OutOfData` and `SurfaceLost` are only those.

rukai: Shall I remove those variants from SwapChainStatus for both swap_chain_get_current_texture_view and swap_chain_present in this PR?

rukai: I went ahead and did that.

PR #822, opened 2020-07-19
Replace borrowed slices in Descriptor structs with copy on write slices

Kimundi:
**Connections**
No known issues, this PR was motivated by discussion in the matrix channels with wgpu project maintainers.

**Description**
Various `[...]Descriptor` structs use borrowed slices for fields with variable number of elements. This is fine for the primary usecase of directly passing the struct instances as function parameters, but problematic if you want to store owned copies of them (say for a caching layer).

By changing all `&'a [T]` into `Cow<'a, [T]>`, these structs can now be used for both usecases

**Breaking change**
In order to change to `Cow`, two `Copy` derives had to be removed.

**Testing**
No API has fundamentally changed or extended, so the exsiting tests should be enough.


wgpu-types/src/lib.rs
```diff
@@ -1472,29 +1475,35 @@ pub struct BindGroupEntry<R> {
 
 /// Describes a group of bindings and the resources to be bound.
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
-pub struct BindGroupDescriptor<'a, L, B> {
+pub struct BindGroupDescriptor<'a, L, B>
+where
+    [B]: ToOwned<Owned = Vec<B>>,
```
kvark, line 1480: why is this needed? If we have `[B]` we should always be able to just call `to_vec()` and get a `Vec`, we shouldn't need these extra bounds

player/src/lib.rs
```diff
@@ -209,8 +209,8 @@ impl GlobalPlay for wgc::hub::Global<IdentityPassThroughFactory> {
                 self.device_create_bind_group_layout::<B>(
```
kvark, line 209: The main motivator behind `Cow` was actually to let the tracing use the same types. Currently `src/device/trace.rs` have to redefine a bunch of types to be owned. If we switch `wgt` stuff to `Cow`, we can use it directly in the `trace` module.
For example this variant we are processing here could just be `CreatebindGroupLayout(Id<id::BindGroupLayoutId>, wgt::BindGroupLayoutDescriptor<'a>)`. The current approach with named fields was different because we couldn't reuse the descriptor types. Once we can, there is no need to name fields have anything other than `(id, desc)` tuple variants.

wgpu-types/src/lib.rs
Kimundi: I haven't investigated what the exact issue is yet, but if I leave the associated type constraint off, derives like `Debug` stop working

player/src/lib.rs
Kimundi: Ah, I wasn't aware of that.

Kimundi: I've now mostly reduced the duplication among the tracing API types. The player can now simply pass the descriptors from `Action` into the right wgpu APIs, and owned variants with `'static` params can be created via the new `ToStatic` trait.

Sadly the implementation of `ToStatic` is relativly verbose, as its a recursivly delegating trait like `Clone` - but that also means that we could turn it into an relativly simple custom derive attribute that lives on crates.io if we wanted to remove that boilerplate.

player/src/lib.rs
```diff
@@ -201,83 +201,25 @@ impl GlobalPlay for wgc::hub::Global<IdentityPassThroughFactory> {
...
-                self.device_create_bind_group_layout::<B>(
-                    device,
-                    &wgt::BindGroupLayoutDescriptor {
-                        label: Some(label),
-                        entries,
-                    },
-                    id,
-                )
-                .unwrap();
+            A::CreateBindGroupLayout { id, desc } => {
```
kvark, line 204: nit: naming these isn't useful any more, we should just have `A::CreateBindGroupLayout(id, desc)`

Kimundi:
Regarding `ToStatic`, I see multiple ways to continue here:

1. Remove the trait, and just hard-code all the verbose `Foo<'a>` -> `Foo<'static>` transformations where they are needed
    - Pro: Simple
    - Con: Verbose; Can not reuse the "make descriptor 'static" operation outside of wgpu
2. Remove the trait, but provide all implementations as functions or methods on the corresponding type.
    - Pro: No extra trait
    - Con: Not much of a difference, and prevents generic handling of the structs
3. Keep the trait, but provide the trait impls via a custom derive crate
    - Pro: Implementation is neat and easily extensible
    - Con: More up-fron efford; Extra build dependency

kvark:
don't want these to be lost:
bovines
cows

PR #825, opened 2020-07-21
Add depth clamping support

kvark:
**Connections**
Implements https://github.com/gpuweb/gpuweb/pull/900

**Description**
Depth clamping is useful for shadow mapping and stuff. We'd want to use it in our `shadow` example.

**Testing**
Untested, should work though :)

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -131,6 +131,18 @@ bitflags::bitflags! {
     #[cfg_attr(feature = "trace", derive(Serialize))]
     #[cfg_attr(feature = "replay", derive(Deserialize))]
     pub struct Features: u64 {
+        /// By default, polygon depth is clipped to 0-1 range. Anything outside of that range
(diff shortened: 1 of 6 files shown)


wgpu-types/src/lib.rs
```diff
@@ -131,6 +131,18 @@ bitflags::bitflags! {
...
+        /// is rejected, and respective fragments are not touched.
+        ///
+        /// With this extension, we can force clamping of the polygon depth to 0-1. That allows
+        /// shadow map occluders to be rendered into a tighter depth range.
+        ///
+        /// Supported platforms:
+        /// - desktops
+        /// - some mobile chips
+        ///
+        /// This is an upstream feature.
```
cwfitzgerald, line 144: nit: users probably don't know what upstream is, so maybe something like "This is a web and native feature" or something like that.

kvark:
bors r=cwfitzgerald

kvark:
bors r=cwfitzgerald

PR #832, opened 2020-07-23
Return errors from device functions

GabrielMajeri:
**Connections**
Part of #638 

**Description**
Lots of changes, but they should be easily reviewable commit-by-commit.

- Error types for most of the fallible resource creation errors in `device/mod.rs`

- Removed assertions and replaced them with error types

- All of the `BufferMap`, `BufferNotMapped` etc. errors were united in a single `BufferAccessError`, since it was pretty weird to have so many overlapping error types.

- Removed all `unwrap`s of `gfx-hal` errors - they are now returned to the caller.

**Testing**
Checked with core and tested with `wgpu-rs` (see https://github.com/gfx-rs/wgpu-rs/pull/469)

### DIFF
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -6,6 +6,7 @@ use crate::{
     device::SHADER_STAGE_COUNT,
     id::{BindGroupLayoutId, BufferId, DeviceId, SamplerId, TextureViewId},
     track::{TrackerSet, DUMMY_SELECTOR},
+    validation::{MissingBufferUsageError, MissingTextureUsageError},
     FastHashMap, LifeGuard, MultiRefCount, RefCount, Stored, MAX_BIND_GROUPS,
 };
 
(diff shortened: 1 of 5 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -1913,6 +1879,8 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 (spv, module)
             }
             pipeline::ShaderModuleSource::Wgsl(code) => {
+                // TODO: refactor the corresponding Naga error to be owned, and then
+                // display it instead of unwrapping
                 let module = naga::front::wgsl::parse_str(&code).unwrap();
```
GabrielMajeri, line 1884: I've tried to use `?` to return this error, but I got a lifetime error. The issue seems to be that `code` is local to this function (yet I think it should work, since the returned lifetime is the same as the shader source's lifetime) :confused:

wgpu-core/src/binding_model.rs
```diff
@@ -25,25 +26,55 @@ use std::{
 use thiserror::Error;
 
 #[derive(Clone, Debug, Error)]
-pub enum BindGroupLayoutError {
+pub enum CreateBindGroupLayoutError {
```
GabrielMajeri, line 29: I've renamed these errors to indicate they happen on resource creation.

wgpu-core/src/binding_model.rs
```diff
@@ -239,30 +268,32 @@ pub struct BindGroupLayout<B: hal::Backend> {
...
     #[error("device does not have required feature: {0:?}")]
     MissingFeature(wgt::Features),
+    #[error("push constant range (index {index}) provides for stage(s) {provided:?} but there exists another range that provides stage(s) {intersected:?}. Each stage may only be provided by one range")]
+    MoreThanOnePushConstantRangePerStage {
+        index: usize,
+        provided: wgt::ShaderStage,
+        intersected: wgt::ShaderStage,
+    },
+    #[error(transparent)]
+    OutOfMemory(#[from] hal::device::OutOfMemory),
```
kvark, line 286: same here

wgpu-core/src/device/mod.rs
```diff
@@ -1705,39 +1678,35 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
                     Br::TextureViewArray(ref bindings_array) => {
-                        assert!(
-                            device.features.contains(wgt::Features::SAMPLED_TEXTURE_BINDING_ARRAY),
-                            "Feature SAMPLED_TEXTURE_BINDING_ARRAY must be enabled to use TextureViewArrays in a bind group"
-                        );
+                        if !device
+                            .features
+                            .contains(wgt::Features::SAMPLED_TEXTURE_BINDING_ARRAY)
+                        {
+                            return Err(CreateBindGroupError::SampledTextureBindingArrayNotEnabled);
```
kvark, line 1696: perhaps it would be easier to just have `MissingFeature` error variant?

wgpu-core/src/device/mod.rs
kvark: I think we need an explicit lifetime:
```rust
    pub fn device_create_shader_module<'a, B: GfxBackend>(
        &self,
        device_id: id::DeviceId,
        source: pipeline::ShaderModuleSource<'a>,
        id_in: Input<G, id::ShaderModuleId>,
    ) -> Result<id::ShaderModuleId, CreateShaderModuleError<'a>> {
```

wgpu-core/src/device/mod.rs
```diff
@@ -2987,41 +2940,94 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
 
 #[derive(Clone, Debug, Error)]
 pub enum CreateBufferError {
-    #[error("`MAP` usage can only be combined with the opposite `COPY`, requested {0:?}")]
-    UsageMismatch(wgt::BufferUsage),
+    #[error(transparent)]
+    BindError(#[from] hal::device::BindError),
```
kvark, line 2944: We don't really want to propagate HAL errors up. They are *not* expected.
The idea of `wgpu` is that it's a safe abstraction over HAL, it has logic to ensure that HAL is used safely, and all errors except for DeviceLost and OutOfMemory are unexpected and should just be unwrapped internally.

GabrielMajeri:
I've updated the PR:
- The raw `gfx-hal` errors are no longer returned. Only the `OutOfMemory`/`DeviceLost` errors get reported, everything else is unwrapped.
- I've replaced the long `SampledTextureBindingArrayNotEnabled` with a `MissingFeatures` variant.
- I've tried returning a `wgsl::ParseError<'a>` from the `create_shader_module` function, but I still encounter an error:
```
error[E0515]: cannot return value referencing local variable `code`
    --> wgpu-core/src/device/mod.rs:1919:65
     |
1919 |                 let module = naga::front::wgsl::parse_str(&code)?;
     |                                                           ----- ^ returns a value referencing data owned by the current function
     |                                                           |
     |                                                           `code` is borrowed here
```
I think I understand the reason. We take in a `ShaderModuleSource<'a>` by _value_.
* If the `Cow` is borrowed, then returning an error type with a reference to those strings won't be an issue.
* If the `Cow` is owned, then _we_ own the shader source string, which means returning an error would point to a string owned by a local variable, which is not allowed.

I think taking `ShaderModuleSource` by reference would be a solution, and it shouldn't be much of an issue upstream.

kvark:
```rust
let module = naga::front::wgsl::parse_str(&code)?;
```
Maybe the problem is `code` being moved in? What if you use `ref code` in the patter match, and just `parse_str(code)` here?

GabrielMajeri:
I've squashed the commits and dropped the `ParseError` change for now. At this point, I'd rather just dedicate a separate PR for it.

kvark:
that's a good decision, thank you!
bors r+

kvark:
bors retry

PR #833, opened 2020-07-24
Switch pipeline flag to mutation of depth/stencil

kvark:
**Connections**
Reported on the matrix

**Description**
If the pass mutates depth/stencil, but the pipeline doesn't, it's not a bug!

**Testing**
on it...


kvark:
bors r=startoaster

Issue #289, opened 2019-08-16
API tracing


kvark:
API tracing is a feature that allows to serialize the workload that the library is processing over a period of time, so that it can be replayed later, potentially on a different machine and with a different GPU backend. It opens up a way to capture bugs (for investigating either the implementation or the application issues), share interesting workloads (as a form of art), or even do benchmarking.

At this point, it's clear that WebGPU as an API needs some tooling for workload capturing. There probably will appear some JS-based tools developed by the users and browser vendors. This issue is focuses on the native targets. Let's see what options we have to proceed:

## A: Intercept in Rust, serialize with Serde.

This basically means transforming all the workload inputs (object descriptors, handles, etc), into pure Rust structs that implement Serde traits and recording them.

What this gives us:
  - Convenience of Rust and its ecosystem (i.e can use RON format as destination).
  - Can start the work today, not blocked by anything
  - Can build an internal testing and benchmarking suite based on the captured format
  - If used in conjunction with #72, could support recording from a particular point (as opposed to the start of an application).

What's not clear is where this hookup should live. It could be a separate layer on top of `wgpu-native`, it could be a part of `wgpu-remote`, or be a part of `wgpu-rs` even.

## B: develop a backend for RenderDoc and integrate with it

Given that RenderDoc became a central piece of the developer graphics tools, and it's open source, there is an opportunity here to start collaborating.

What this gives us:
  - RenderDoc is not an afterthought, we get it from day 1
  - convenience of the GUI

## C: wait for the native tooling to emerge

We are not the only party interested in this technology. Obviously, Google is either already in process, or will attempt to build some tooling for tracing the API at the C level. Something like `wgpuTrace` as an analogy to `vkTrace` and `apiTrace`.

For us this plan means letting the things to advance by Google and assisting where time allows, focusing on interoperability with wgpu-native implementation (since clearly a hypothetical `wgpuTrace` will be catering to Dawn first and foremost).

What this gives us:
  - path of least resistance

---

Now, I'm personally leaning towards (A) since I don't think this is going to be a huge amount of work, I think it could play well with #72, it will allow us to inspect and modify captures in RON by hands, and build some testing/benchmarking without relying on external suites.

My understanding is that when/if `wgpuTrace` becomes a thing, we'll just set up a tooling for bidirectional conversion to its binary format, effectively mimicking how ApiTrace trace is different from the dump.

This is a heavy RFC, please share your thoughts!

m4b:
[...]

PR #835, opened 2020-07-24
Convert all logging to tracing and add fmt logger

cwfitzgerald:
**Connections**

#289 ish.

**Description**

This converts us fully to tracing, allowing traces to show up in logs.

https://github.com/gfx-rs/wgpu-rs/pull/476 should be considered at the same time as this PR.

**Testing**

Ran examples in wgpu-rs.

### DIFF
--- a/player/src/bin/play.rs
+++ b/player/src/bin/play.rs
@@ -17,7 +17,12 @@ fn main() {
     #[cfg(feature = "winit")]
     use winit::{event_loop::EventLoop, window::WindowBuilder};
 
-    env_logger::init();
+    wgc::logging::subscriber::initialize_default_subscriber(
+        std::env::var("WGPU_CHROME_TRACE")
+            .as_ref()
+            .map(Path::new)
+            .ok(),
+    );
 
     #[cfg(feature = "renderdoc")]
     #[cfg_attr(feature = "winit", allow(unused))]
--- a/wgpu-core/src/logging/subscriber/chrome.rs
+++ b/wgpu-core/src/logging/subscriber/chrome.rs
@@ -1,0 +1,170 @@
+use crate::logging::subscriber::CURRENT_THREAD_ID;
+use parking_lot::Mutex;
+use std::{
+    borrow::Cow,
+    fmt,
+    io::{self, Write as _},
+    path::Path,
+    sync::Arc,
+    time::Instant,
+};
+use tracing::field::{Field, Visit};
+use tracing::{span, Event, Metadata, Subscriber};
+use tracing_subscriber::{layer::Context, registry::LookupSpan, Layer};
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+enum EventType {
+    Begin,
+    Event,
+    End,
+}
+
+/// A layer to add to a [`tracing_subscriber::Registry`] to output to a chrome
+/// trace.
+///
+/// If you want an easy "set and forget" method of installing this and normal
+/// tracing logging, call [`initialize_default_subscriber`].
+#[derive(Debug, Clone)]
+pub struct ChromeTracingLayer {
+    file: Arc<Mutex<std::fs::File>>,
+    start_time: Instant,
+    process_id: u32,
+}
+
+impl ChromeTracingLayer {
+    /// Create a trace which outputs to the given file. The file will be cleared if it exits.
+    pub fn with_file(file: impl AsRef<Path>) -> io::Result<Self> {
+        std::fs::File::create(file).map(|mut file| {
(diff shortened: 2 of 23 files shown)


player/src/bin/play.rs
```diff
@@ -17,7 +17,13 @@ fn main() {
     #[cfg(feature = "winit")]
     use winit::{event_loop::EventLoop, window::WindowBuilder};
 
-    env_logger::init();
+    #[cfg(feature = "subscriber")]
```
kvark, line 20: Player doesn't have this feature. Since it's a debugging tool, we should probably just enable it in `wgc` dependency for the player. But the `WGPU_CHROME_TRACE` is probably not the best way to tell the player where to put the trace. I think, ideally, we'd want some `--trace <path>` parameter on command line? Anyway, we can follow-up with this, not a blocker.

wgpu-core/src/logging/subscriber/chrome.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+pub struct FmtLayer;
+
+impl<S> Layer<S> for FmtLayer
+where
+    S: Subscriber + for<'span> LookupSpan<'span>,
+{
+    fn on_event(&self, event: &Event<'_>, ctx: Context<'_, S>) {
+        let time = chrono::Local::now().format("%Y-%m-%d %H-%M-%S%.6f");
```
kvark, line 28: I wonder if we should make this optional, even when subscriber feature is ON. We can do all the things without printing the exact date. We can count time using `std::time::Duration`.

wgpu-core/src/logging/subscriber/chrome.rs
```diff
@@ -0,0 +1,67 @@
...
+            Level::TRACE => "TRACE",
+        };
+
+        match *metadata.level() {
+            Level::ERROR | Level::WARN => eprintln!(
+                "[{} {}]({})({}): {}",
+                time,
+                level,
+                span_string,
+                metadata.module_path().unwrap_or("no module"),
```
kvark, line 54: nit: let's move this out

wgpu-core/src/logging/subscriber/chrome.rs
```diff
@@ -0,0 +1,67 @@
...
+        let level = match *metadata.level() {
+            Level::ERROR => "ERROR",
+            Level::WARN => "WARN",
+            Level::INFO => "INFO",
+            Level::DEBUG => "DEBUG",
+            Level::TRACE => "TRACE",
+        };
+
+        match *metadata.level() {
+            Level::ERROR | Level::WARN => eprintln!(
```
kvark, line 49: Somewhat unfortunate to duplicate this. I suppose doing a `format!` call is undesirable since it's `String`-allocating?
We could do `format_args!` and then pick the stderr or stdout for a single [write](https://doc.rust-lang.org/stable/std/fmt/fn.write.html) call

wgpu-core/src/logging/subscriber/fmt_layer.rs
```diff
@@ -0,0 +1,171 @@
...
+            CURRENT_THREAD_ID.with(|v| *v),
+            instant_scope,
+        )
+        .unwrap();
+    }
+}
+
+impl Drop for ChromeTracingLayer {
+    fn drop(&mut self) {
+        println!("drop");
```
kvark, line 116: is this temporary?

cwfitzgerald: Whooooops, that shouldn't have escaped my ide :)

wgpu-core/src/logging/subscriber/chrome.rs
cwfitzgerald: Eh, I'll just nuke it entirely, I doubt that absolute times matters to graphics applications as much as it does to say web server applications where multiple logs have to be synced.

player/src/bin/play.rs
cwfitzgerald: We should probably grab one of the mini arg parsing libs to use for player, I'm personally partial to preco-args, but doesn't ultimately matter.

wgpu-core/src/logging/subscriber/chrome.rs
cwfitzgerald: fmt::write takes fmt::Write which isn't really implemented by much, so I should be able to manage to turn this into a single `write!` call maybe...

cwfitzgerald:
Pushed twice, first push addressed comments, second push swapped the file names of `fmt_layer.rs` and `chrome.rs` and rebased.

Issue #834, opened 2020-07-24
BindGroupLayout[0] does not exist (sometimes)


tiberiusferreira:
Sorry, this is one of those "sometimes crashes" bugs.

OS: MacOS 10.15.6
wgpu-rs version: e785b213e409460649da5c78068f653d2634bafc

Message:
```
---- tensors::gpu_tensor::gpu_ops::bmm::tests::simple_rank_2_mm stdout ----
thread 'tensors::gpu_tensor::gpu_ops::bmm::tests::simple_rank_2_mm' panicked at 'BindGroupLayout[0] does not exist', /Users/tiberiodarferreira/.cargo/git/checkouts/wgpu-53e70f8674b08dd4/0591c48/wgpu-core/src/device/life.rs:560:20
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Metal validation does not report anything, see another instance below with Metal Validation:

```
running 25 tests
2020-07-24 05:43:34.407 gpu_compute-68282a59c1913004[19070:1079565] Metal API Validation Enabled
test tensors::gpu_tensor::gpu_ops::fill_with::tests::fill_with ... ok
test tensors::gpu_tensor::gpu_ops::clone::tests::can_clone ... FAILED
test tensors::gpu_tensor::gpu_ops::clone::tests::can_clone_non_uniform_shape ... ok
test tensors::gpu_tensor::gpu_ops::bmm::tests::simple_rank_2_mm ... ok
test tensors::gpu_tensor::indexing::index_to_dim_stride::allows_start_equals_end ... ok
test tensors::gpu_tensor::indexing::index_to_dim_stride::can_index ... ok
test tensors::gpu_tensor::indexing::index_to_dim_stride::can_index_edge_cases ... ok
test tensors::gpu_tensor::indexing::index_to_dim_stride::cant_index_more_elements_than_dimension_has ... ok
test tensors::gpu_tensor::indexing::index_to_dim_stride::doesnt_allow_end_smaller_than_start ... ok
test tensors::gpu_tensor::indexing::test ... ok
test tensors::gpu_tensor::shape_changing::broadcast_tests::broadcast_works_from_scalar ... ok
test tensors::gpu_tensor::shape_changing::broadcast_tests::broadcast_works_with_additional_unit_dim ... ok
test tensors::gpu_tensor::shape_changing::broadcast_tests::simple_broadcast_with_skipping_works ... ok
[...]

PR #838, opened 2020-07-24
wait for buffer to be done in the player

kvark:
**Connections**
Fixes our code enough to replay #834 without issues. Doesn't help to solve the original problem though.

**Description**
There are two things in here:
  1. don't deduplicate the BGLs if we are not generating new IDs at this layer. This helps Servo/Gecko/player. cc @kunalmohan 
  2. have an option in `buffer_destroy` to *actually* kill it, at the cost of blocking on GPU sometimes. This is required for the player, since the very next command may try to reuse the ID.

**Testing**
Tested on the trace in #834

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -579,6 +579,17 @@ impl<B: GfxBackend> Device<B> {
                 .unwrap()
         }
     }
+
+    fn wait_for_submit(&self, submission_index: SubmissionIndex, token: &mut Token<Self>) -> Result<(), WaitIdleError> {
+        if self.last_completed_submission_index() <= submission_index {
+            log::info!("Waiting for submission {:?}", submission_index);
+            self.lock_life(token)
+                .triage_submissions(&self.raw, true)
+                .map(|_| ())
+        } else {
+            Ok(())
+        }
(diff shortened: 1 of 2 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -1199,8 +1217,9 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         let device = &device_guard[device_id];
 
         // If there is an equivalent BGL, just bump the refcount and return it.
-        // Warning: this isn't valid logic when `id_in` is provided.
-        {
+        // This is only applicable for identity filters that are generating new IDs,
+        // so their inputs are `PhantomData` of size 0.
+        if mem::size_of::<Input<G, id::BindGroupLayoutId>>() == 0 {
```
cwfitzgerald, line 1222: Is there any cleaner way of doing this, this feels _really_ hacky

kvark: I agree! but I wasn't able to properly detect this in a simple way.
since it's solving a real problem, and it's correct, let's proceed and follow up if necessary

kvark:
bors r=cwfitzgerald

PR #839, opened 2020-07-24
Fix push constant pipeline invalidation

### LINKED ISSUE #821: Validation error DescriptorSetNotBound when using (compute-)pipline without push constants after one with
The moment I start using push constants in a compute pipeline, some (unclear what's necessary for that) following compute pipelines (in the same pass) that don't use push constants will give me this validation error:

```
  VALIDATION [UNASSIGNED-CoreValidation-DrawState-DescriptorSetNotBound (-840888189)] : Validation Error: [ UNASSIGNED-CoreValidation-DrawState-DescriptorSetNotBound ] Object 0: handle = 0x1928a4a6530, name = Encoder: Frame Main, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xcde11083 | VkPipeline 0xa7e05300000000d6[] uses set gfx-rs/wgpu-rs#1 but that set is not bound.
    object info: (type: COMMAND_BUFFER, hndl: 1728896984368, name: Encoder: Frame Main)
```
everything else equal, the moment I stop using push constants the problem goes away again

After the validation error the app gets stuck.
I created a trace by running it in an optimized build (no validation error) which seems to run fine. Resulting trace has a lot of noise, but the player is able to repro the issue for me (using wgpu-core a0ed09a6a83dc0d16c6ad711cf139a5fcfbb9c35)

[trace.zip](https://github.com/gfx-rs/wgpu-rs/files/4943384/trace.zip)

cwfitzgerald:
**Connections**

Fixes #821

**Description**

We need to invalidate everything if push constants change. This code is kinda ugly but this is try three and it's the least terrible I got.

**Testing**

Pinging @Wumpf :)

### DIFF
--- a/wgpu-core/src/command/compute.rs
+++ b/wgpu-core/src/command/compute.rs
@@ -251,10 +251,12 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                     // Rebind resources
                     if state.binder.pipeline_layout_id != Some(pipeline.layout_id.value) {
                         let pipeline_layout = &pipeline_layout_guard[pipeline.layout_id.value];
-                        state.binder.pipeline_layout_id = Some(pipeline.layout_id.value);
-                        state
-                            .binder
-                            .reset_expectations(pipeline_layout.bind_group_layout_ids.len());
+
+                        state.binder.change_pipeline_layout(
+                            &*pipeline_layout_guard,
+                            pipeline.layout_id.value,
+                        );
+
                         let mut is_compatible = true;
 
                         for (index, (entry, &bgl_id)) in state
--- a/wgpu-core/src/command/render.rs
+++ b/wgpu-core/src/command/render.rs
@@ -1103,10 +1103,12 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                     // Rebind resource
                     if state.binder.pipeline_layout_id != Some(pipeline.layout_id.value) {
                         let pipeline_layout = &pipeline_layout_guard[pipeline.layout_id.value];
-                        state.binder.pipeline_layout_id = Some(pipeline.layout_id.value);
-                        state
-                            .binder
-                            .reset_expectations(pipeline_layout.bind_group_layout_ids.len());
+
+                        state.binder.change_pipeline_layout(
+                            &*pipeline_layout_guard,
+                            pipeline.layout_id.value,
+                        );
+
                         let mut is_compatible = true;
 
                         for (index, (entry, &bgl_id)) in state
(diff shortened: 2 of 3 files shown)


wgpu-core/src/command/render.rs
```diff
@@ -1103,10 +1103,26 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
                         let pipeline_layout = &pipeline_layout_guard[pipeline.layout_id.value];
-                        state.binder.pipeline_layout_id = Some(pipeline.layout_id.value);
-                        state
+
+                        let old_pipeline_layout_id = state
                             .binder
-                            .reset_expectations(pipeline_layout.bind_group_layout_ids.len());
+                            .pipeline_layout_id
+                            .replace(pipeline.layout_id.value);
+                        let reset_start = if let Some(old_id) = old_pipeline_layout_id {
```
kvark, line 1111: can we move that logic into `binder` perhaps to not copy between the compute/render?

wgpu-core/src/command/compute.rs
```diff
@@ -251,10 +251,16 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
                         let pipeline_layout = &pipeline_layout_guard[pipeline.layout_id.value];
-                        state.binder.pipeline_layout_id = Some(pipeline.layout_id.value);
-                        state
+
+                        let old_pipeline_layout_id = state
                             .binder
-                            .reset_expectations(pipeline_layout.bind_group_layout_ids.len());
+                            .pipeline_layout_id
+                            .replace(pipeline.layout_id.value);
+                        let old_pipeline_layout =
```
kvark, line 259: would it be simpler to just follow up with a `map` here (instead of `old_pipeline_layout_id` intermediate)?

wgpu-core/src/command/compute.rs
```diff
@@ -251,10 +251,16 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                     // Rebind resources
                     if state.binder.pipeline_layout_id != Some(pipeline.layout_id.value) {
                         let pipeline_layout = &pipeline_layout_guard[pipeline.layout_id.value];
-                        state.binder.pipeline_layout_id = Some(pipeline.layout_id.value);
-                        state
+
+                        let old_pipeline_layout_id = state
```
kvark, line 255: I was thinking more like something `state.binder.change_pipeline_layout(&*pipeline_layout_guard, pipeline.layout_id.value)`, which internally do all the things, including `reset_expectations`

wgpu-core/src/command/compute.rs
cwfitzgerald: The reason I went for a temporary here is because replace isn't just a chain, it's the operation setting the new value of `pipeline_layout_id` so to call attention to it, it gets it's own line.

cwfitzgerald:
Third time's the charm 😄

PR #840, opened 2020-07-26
Add builder for TextureDescriptor

cwfitzgerald:
**Connections**

https://github.com/gfx-rs/wgpu-rs/issues/478 https://github.com/gfx-rs/wgpu-rs/issues/470

**Description**

Adds a mut based builder pattern to TextureDescriptor as a test.

**Testing**

Will test in wgpu-rs...

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -12,6 +12,9 @@
 use serde::{Deserialize, Serialize};
 use std::{borrow::Cow, num::NonZeroU32, ops::Range};
 
+pub use into_cow::IntoCow;
+mod into_cow;
+
 /// Integral type used for buffer offsets.
 pub type BufferAddress = u64;
 /// Integral type used for buffer slice sizes.
@@ -1302,6 +1305,42 @@ impl<L> TextureDescriptor<L> {
             format: self.format,
             usage: self.usage,
         }
+    }
+}
+
+/// Texture descriptor builder methods
+impl<'a> TextureDescriptor<Option<Cow<'a, str>>> {
+    pub fn new(
+        size: Extent3d,
+        dimension: TextureDimension,
+        format: TextureFormat,
+        usage: TextureUsage,
+    ) -> Self {
+        Self {
+            label: None,
+            size,
+            mip_level_count: 1,
+            sample_count: 1,
+            dimension,
+            format,
+            usage,
+        }
+    }
+
+    pub fn label<L>(&mut self, label: impl IntoCow<'a, str>) -> &mut Self {
+        self.label = Some(label.into_cow());
+        self
+    }
+
+    pub fn mip_level_count(&mut self, count: u32) -> &mut Self {
+        // TODO: validate this size is sane
+        self.mip_level_count = count;
+        self
+    }
+
+    pub fn sample_count(&mut self, count: u32) -> &mut Self {
+        self.sample_count = count;
+        self
     }
 }
 
--- a/wgpu-types/src/into_cow.rs
+++ b/wgpu-types/src/into_cow.rs
@@ -1,0 +1,45 @@
+use std::borrow::Cow;
+
+/// Implements conversion into a cow.
+pub trait IntoCow<'c, CowType: ?Sized + ToOwned> {
+    fn into_cow(self) -> Cow<'c, CowType>;
+}
+
+impl<'c, T: Clone> IntoCow<'c, [T]> for &'c [T] {
+    fn into_cow(self) -> Cow<'c, [T]> {
+        Cow::Borrowed(self)
+    }
+}
+
+impl<'c> IntoCow<'c, str> for &'c str {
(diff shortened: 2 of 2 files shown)


wgpu-types/src/lib.rs
```diff
@@ -1574,7 +1614,7 @@ pub struct RenderPipelineDescriptor<'a, L, D> {
     /// Bitmask that restricts the samples of a pixel modified by this pipeline. All samples
     /// can be enabled using the value `!0`
     pub sample_mask: u32,
-    /// When enabled, produces another sample mask per pixel based on the alpha output value, that
+    /// When enabled, produces another sample mask per pixel based on theT alpha output value, that
```
GabrielMajeri, line 1617: Is this intentional?

cwfitzgerald: No, just missed it when quickly committing.

wgpu-types/src/lib.rs
```diff
@@ -30,6 +30,61 @@ pub const COPY_BUFFER_ALIGNMENT: BufferAddress = 4;
 /// Alignment all push constants need
 pub const PUSH_CONSTANT_ALIGNMENT: u32 = 4;
 
+/// Implements conversion into a cow.
+pub trait IntoCow<'c> {
```
kvark, line 34: nit: move into a module?

wgpu-types/src/lib.rs
```diff
@@ -1303,6 +1358,49 @@ impl<L> TextureDescriptor<L> {
...
+            size,
+            mip_level_count: 1,
+            sample_count: 1,
+            dimension,
+            format,
+            usage,
+        }
+    }
+
+    pub fn build(&self) -> Self {
```
kvark, line 1380: is this necessary? If all our methods receive `&Descriptor`, most users don't need this `build()` step.

wgpu-types/src/lib.rs
```diff
@@ -1303,6 +1358,49 @@ impl<L> TextureDescriptor<L> {
     }
 }
 
+/// Texture descriptor builder methods
+impl<'l> TextureDescriptor<Option<Cow<'l, str>>> {
```
kvark, line 1362: We should use a different name for the lifetime. How about `'a`?
The `'l` you have is probably meant for labels, but other descriptors have other stuff too, and we'd better use a single lifetime name consistently.

wgpu-types/src/lib.rs
```diff
@@ -1303,6 +1358,49 @@ impl<L> TextureDescriptor<L> {
...
+            format,
+            usage,
+        }
+    }
+
+    pub fn build(&self) -> Self {
+        Self::clone(self)
+    }
+
+    pub fn label<L>(&mut self, label: L) -> &mut Self
```
kvark, line 1384: given that we are going to use this pattern a lot (a generic that is into-cowed), could we use the shorter syntax for this? I.e. `pub fn label(&mut self, label: impl IntoCow<'l, CowType = str>) -> &mut Self`

wgpu-types/src/lib.rs
```diff
@@ -30,6 +30,61 @@ pub const COPY_BUFFER_ALIGNMENT: BufferAddress = 4;
 /// Alignment all push constants need
 pub const PUSH_CONSTANT_ALIGNMENT: u32 = 4;
 
+/// Implements conversion into a cow.
+pub trait IntoCow<'c> {
+    type CowType: ?Sized + ToOwned;
```
kvark, line 35: We never have a case where we `CowType` is unknown to us. Therefore, we can turn it into a generic parameter:
```rust
pub trait IntoCow<'c, C> {
  fn into_cow(self) -> Cow<'a, C>;
}
```
The benefit of this would be less typing, i.e. `IntoCow<'a, str>` instead of `IntoCow<'a, CowType = str>`, and I don't think there are downsides for our case.

wgpu-types/src/lib.rs
cwfitzgerald: No, not necessarily. I guess people can just use clone if they need to put it somewhere.

wgpu-types/src/lib.rs
cwfitzgerald: Yeah, agreed. I thought I needed a manual where, but it doesn't turn out needed.

cwfitzgerald:
Updated and rebased.

wgpu-types/src/lib.rs
```diff
@@ -12,6 +12,9 @@
 use serde::{Deserialize, Serialize};
 use std::{borrow::Cow, num::NonZeroU32, ops::Range};
 
+pub use into_cow::*;
```
kvark, line 15: let's specifically do `pub use into_cow::IntoCow` instead of the glob import

cwfitzgerald:
Concerns addressed!
bors r+

cwfitzgerald:
bors r=kvark

PR #843, opened 2020-07-28
Reformat the alignment errors, add a stride check

kvark:
**Connections**
Addresses [one of the things](https://community-tc.services.mozilla.com/tasks/JQlmdXSiSK2ufM7mX8F1Tg/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FJQlmdXSiSK2ufM7mX8F1Tg%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L1999) Servo hit on the CTS

**Description**
This change affects the unaligned* error messages: instead of printing the exact value, they now refer to the constant in the code. On one hand, this involves an extra redirection for anyone who hits this. But on the other hand, it makes them aware of the actual constants they should be respecting. I think it's better in the long term, especially if we change any of these.

**Testing**
Not tested, but should work!

r? @GabrielMajeri

### DIFF
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -66,10 +66,7 @@ pub enum CreateBindGroupError {
     SingleBindingExpected,
     #[error("unable to create a bind group with a swap chain image")]
     SwapChainImage,
-    #[error(
-        "buffer offset {0} must be a multiple of {}",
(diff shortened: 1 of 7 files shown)


kvark:
Thank you!
bors r=GabrielMajeri

PR #845, opened 2020-07-28
Finish error model refactor

GabrielMajeri:
**Connections**
I think this is the last part of #638. I've reviewed all the remaining `unwrap`s and `assert`s in the code, and these should be the last ones left which ought to return errors (the remaining ones seem to uphold internal invariants).

**Description**
Implements error handling for various conditions, which are then returned to the caller. Including, but not limited to:
- running out of memory when creating a command pool
- running out of memory when creating a frame buffer for a render pass
- invalid dimensions when creating a texture

**Testing**
Tested with core and player.

### DIFF
--- a/wgpu-core/src/conv.rs
+++ b/wgpu-core/src/conv.rs
@@ -7,6 +7,9 @@ use crate::{
     resource, PrivateFeatures,
 };
 
+use std::convert::TryInto;
+use thiserror::Error;
+
 pub fn map_buffer_usage(usage: wgt::BufferUsage) -> (hal::buffer::Usage, hal::memory::Properties) {
     use hal::buffer::Usage as U;
     use hal::memory::Properties as P;
@@ -429,11 +432,6 @@ pub fn map_vertex_format(vertex_format: wgt::VertexFormat) -> hal::format::Forma
     }
 }
 
-fn checked_u32_as_u16(value: u32) -> u16 {
-    assert!(value <= ::std::u16::MAX as u32);
-    value as u16
-}
-
 pub fn is_power_of_two(val: u32) -> bool {
     val != 0 && (val & (val - 1)) == 0
 }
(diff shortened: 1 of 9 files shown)


wgpu-core/src/device/life.rs
```diff
@@ -304,6 +304,7 @@ impl<B: hal::Backend> LifetimeTracker<B> {
         let done_count = self
             .active
             .iter()
+            // TODO: return device lost from `get_fence_status`
             .position(|a| unsafe { !device.get_fence_status(&a.fence).unwrap() })
```
GabrielMajeri, line 308: I'm not sure how I could get a `Result` out of the `position` closure :thinking:

wgpu-core/src/conv.rs
```diff
@@ -446,28 +444,48 @@ pub fn map_texture_dimension_size(
...
+}
+
+#[derive(Clone, Debug, Error)]
+pub enum MapTextureDimensionSizeError {
+    #[error("1D images must have height set to 1")]
+    InvalidHeight,
+    #[error("sample count {0} is invalid")]
+    InvalidSampleCount(u32),
+    #[error(transparent)]
+    ValueTooLarge(#[from] TryFromIntError),
```
kvark, line 488: This is meant to be for array sizes only, right?
We'll need to have that limit officially (on the array size), and this error would then be `TooManyArrayLayers(value)` or something like this. We shouldn't expose `TryFromIntError`, since the actual limit is always smaller than the u16::max

wgpu-core/src/device/life.rs
kvark: for this case I think it's fine to even have `unwrap_or(false)`

wgpu-core/src/conv.rs
kvark: the `TryFromIntError` is still here?

GabrielMajeri: Fixed now

PR #846, opened 2020-07-28
Make level and layer count for texture view optional

kvark:
**Connections**
implements https://github.com/gpuweb/gpuweb/pull/945

**Description**
`NonZeroU32` is more idiomatic here

**Testing**
untested, but should work

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -10,7 +10,7 @@
 
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
-use std::{borrow::Cow, ops::Range};
+use std::{borrow::Cow, num::NonZeroU32, ops::Range};
 
 /// Integral type used for buffer offsets.
 pub type BufferAddress = u64;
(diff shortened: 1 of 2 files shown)


wgpu-types/src/lib.rs
```diff
@@ -1343,11 +1343,11 @@ pub struct TextureViewDescriptor<L> {
     /// Base mip level.
     pub base_mip_level: u32,
     /// Mip level count. Must be at least one. base_mip_level + level_count must be less or equal to underlying texture mip count.
```
cwfitzgerald, line 1345: Doc comment now needs adjustment.

wgpu-types/src/lib.rs
```diff
@@ -1343,11 +1343,11 @@ pub struct TextureViewDescriptor<L> {
...
-    pub level_count: u32,
+    pub level_count: Option<NonZeroU32>,
     /// Base array layer.
     pub base_array_layer: u32,
     /// Layer count. Must be at least one. base_array_layer + array_layer_count must be less or equal to the underlying array count.
```
cwfitzgerald, line 1349: same here

kvark:
Does are updated
bors r=cwfitzgerald

PR #847, opened 2020-07-28
Gecko-requested refactor

kvark:


aloucks:
This is much nicer. The previous naming didn't have much meaning and was confusing. It would be nice to see this taken a bit further and use `cmd_buffer` instead of `cmdbuf`. In either case, it's better than hair grooming utensils!

PR #853, opened 2020-07-29
Implement TEXTURE_COMPRESSION_BC extension

### LINKED ISSUE #852: Compressed Textures
Per https://github.com/gpuweb/gpuweb/issues/144#issuecomment-456897457, compressed textures will be included in the WebGPU spec. However, I can't find much else referencing them. Are there plans to add support to wgpu-rs/wgpu-core in the near term? Or even a sense of what form they may take?

cwfitzgerald:
**Connections**

Closes #852.

**Description**

Adds support for BCn textures as specified in the upstream issue. This also shores up the validation and copy logic to work well with the block oriented nature of compressed textures. ETC2 and ASTC should fall out of this easily.

**Testing**

No wgpu-rs changes were needed, however I tested it with the following changes in wgpu-rs: https://github.com/cwfitzgerald/wgpu-rs/commit/96c05ef4bf34c955ec5123a3ebeee1ab98916dc6.

### DIFF
--- a/wgpu-core/src/conv.rs
+++ b/wgpu-core/src/conv.rs
@@ -392,6 +392,141 @@ pub(crate) fn map_texture_format(
                 H::D32SfloatS8Uint
             }
         }
+
+        // BCn compressed formats
+        Tf::Bc1RgbaUnorm => H::Bc1RgbaUnorm,
+        Tf::Bc1RgbaUnormSrgb => H::Bc1RgbaSrgb,
+        Tf::Bc2RgbaUnorm => H::Bc2Unorm,
+        Tf::Bc2RgbaUnormSrgb => H::Bc2Srgb,
+        Tf::Bc3RgbaUnorm => H::Bc3Unorm,
+        Tf::Bc3RgbaUnormSrgb => H::Bc3Srgb,
+        Tf::Bc4RUnorm => H::Bc4Unorm,
+        Tf::Bc4RSnorm => H::Bc4Snorm,
+        Tf::Bc5RgUnorm => H::Bc5Unorm,
+        Tf::Bc5RgSnorm => H::Bc5Snorm,
+        Tf::Bc6hRgbSfloat => H::Bc6hSfloat,
+        Tf::Bc6hRgbUfloat => H::Bc6hUfloat,
+        Tf::Bc7RgbaUnorm => H::Bc7Unorm,
+        Tf::Bc7RgbaUnormSrgb => H::Bc7Srgb,
+    }
+}
+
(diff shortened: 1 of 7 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -948,6 +948,15 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             None => (),
         };
 
+        let texture_features = conv::texture_features(desc.format);
+        if texture_features != wgt::Features::empty() && !device.features.contains(texture_features)
```
kvark, line 952: nit: is the first check necessary?

wgpu-core/src/conv.rs
```diff
@@ -392,6 +392,141 @@ pub(crate) fn map_texture_format(
...
+        | Tf::Bc5RgUnorm
+        | Tf::Bc5RgSnorm
+        | Tf::Bc6hRgbUfloat
+        | Tf::Bc6hRgbSfloat
+        | Tf::Bc7RgbaUnorm
+        | Tf::Bc7RgbaUnormSrgb => (4, 4),
+    }
+}
+
+pub fn texture_features(format: wgt::TextureFormat) -> wgt::Features {
```
kvark, line 474: Could we merge this with `texture_block_size` function? There is a lot of correlation between them: requiring a specific extension comes with a block size, so the combined `match` is going to be roughly as big as a single one of the two. I understand that the code needs this in two different places, but here we can save the code, so we should try it.

kvark:
> No wgpu-rs changes were needed, however I tested it with the following changes in wgpu-rs

Can we convert the skybox example (having notoriously big texture) to DDS?

wgpu-core/src/device/mod.rs
cwfitzgerald: I'm really not sure, I added it out of an abundance of causion as I know bitflags does weird things with empty flags. If you think it's not I can remove it, I just wasn't sure at the time.

wgpu-core/src/conv.rs
cwfitzgerald: Right this second we can, but once we get to astc textures, the match is going to be very different as basically every format in atsc has a different block size. I'm also not sure how exactly etc2 textures work block wise.

cwfitzgerald:
Definitely. I'll look at this as a way to test all of the different formats as we build them up.

wgpu-core/src/conv.rs
kvark: ok, we can keep it on the radar then as we roll in ASTC

Issue #851, opened 2020-07-29
Builder patterns for descriptors


kvark:
**Is your feature request related to a problem? Please describe.**
Since the bovine intervention of #822, we now have a problem in wgpu-rs, since `Cow` is overly verbose.
Another problem is extending structures with extensions without breaking the API. We tried to roll it out previously, but then faced issues, and had to roll back in #787. With builders, the path to non-exhaustive is open.

**Describe the solution you'd like**
Follow the pattern established in #840 for other descriptors. This is done with help of our `IntoCow` trait.

**Describe alternatives you've considered**
Plain structures. Ends up overly verbose, and are hard to expand.

Andful:
I am currently working on it. Hopefully will have a first draft of the builder functions soon.

jprudil:
Would it be possible to use [derive_builder](https://docs.rs/derive_builder/0.9.0/derive_builder/)? It's a well-known crate and requires only a `#[derive(Builder)]` to get started, it also supports carrying over doc comments, intentionally leaving out some fields, custom traits etc. The current builder for `TextureDescriptor` already suffers from quite a lot of code duplication for something that could (potentially) be automated.

Andful:
I do somewhat agree with @jprudil 's idea. I made a pull request with the work-in-progress changes. Not all the structs were given builder pattern but already an additional 500 lines of code is added to a file of 2000 lines of code. Most of the code is very repetitive.
Even though I agree that we should have an automated process to generate the builder implementation, I don't think derive_builder quite matches our use case for the following reasons:
* Not all fields should have a default value, but the user should explicitly pass values for variables with no trivial default value in the function `new`. In the example in #840, those are `size`, `dimension`, `format`, `usage`.
* In case the struct has a `label` field with a generic type `L`, we what to implement the builder pattern only for `L` being `Option<Cow<'a, str>>` (because that is the specialization used in `wgpu-rs`).
* Some setter functions that sets a `Cow` field should accept a `IntoCow` argument and call `into_cow` to set the field.

jprudil:
To address your first concern (provided I understand it correctly): Default values are opt-in with derive_builder. If your member/struct implements the Default trait, you can specify `#[builder(default)]` to make it optional, you can also specify `#[builder(default = "expr")]` where expr is an arbitrary expression of compatible type for that specific member/struct. In case you don't provide defaults, calling the final build method will error out with a message saying what exactly is missing.

kvark:
[...]

PR #856, opened 2020-07-31
Implementation of the builder pattern

Andful:
**Connections**
This pull request addresses the issue #851 

**Description**
Add a builder for wgpu types.

**Testing**
TODO

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -111,7 +111,7 @@ impl From<Backend> for BackendBit {
 
 /// Options for requesting adapter.
 #[repr(C)]
-#[derive(Clone, Debug, Default, PartialEq, Eq, Hash)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "trace", derive(Serialize))]
 #[cfg_attr(feature = "replay", derive(Deserialize))]
 pub struct RequestAdapterOptions<S> {
@@ -120,6 +120,31 @@ pub struct RequestAdapterOptions<S> {
     /// Surface that is required to be presentable with the requested adapter. This does not
     /// create the surface, only guarantees that the adapter can present to said surface.
     pub compatible_surface: Option<S>,
+}
+
+impl<S> RequestAdapterOptions<S> {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn power_preference(&mut self, preference: PowerPreference) -> &mut Self {
+        self.power_preference = preference;
+        self
+    }
+
+    pub fn compatible_surface(&mut self, surface: S) -> &mut Self {
+        self.compatible_surface = Some(surface);
+        self
+    }
+}
+
+impl<S> Default for RequestAdapterOptions<S> {
+    fn default() -> Self {
+        Self {
+            power_preference: PowerPreference::default(),
+            compatible_surface: None,
+        }
+    }
 }
 
 bitflags::bitflags! {
@@ -357,6 +382,27 @@ pub struct DeviceDescriptor {
     pub shader_validation: bool,
 }
 
+impl DeviceDescriptor {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn features(&mut self, features: Features) -> &mut Self {
+        self.features = features;
+        self
+    }
+
+    pub fn limits(&mut self, limits: Limits) -> &mut Self {
+        self.limits = limits;
+        self
+    }
+
+    pub fn shader_validation(&mut self) -> &mut Self {
+        self.shader_validation = true;
+        self
+    }
+}
+
 bitflags::bitflags! {
     /// Describes the shader stages that a binding will be visible from.
     ///
@@ -472,6 +518,23 @@ impl BlendDescriptor {
     }
 }
 
+impl BlendDescriptor {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn factors(&mut self, src_factor: BlendFactor, dst_factor: BlendFactor) -> &mut Self {
+        self.src_factor = src_factor;
+        self.dst_factor = dst_factor;
+        self
+    }
+
+    pub fn operation(&mut self, operation: BlendOperation) -> &mut Self {
+        self.operation = operation;
+        self
+    }
+}
+
 impl Default for BlendDescriptor {
     fn default() -> Self {
         BlendDescriptor::REPLACE
@@ -493,6 +556,32 @@ pub struct ColorStateDescriptor {
     pub color_blend: BlendDescriptor,
     /// Mask which enables/disables writes to different color/alpha channel.
     pub write_mask: ColorWrite,
+}
+
+impl ColorStateDescriptor {
+    pub fn new(format: TextureFormat) -> Self {
+        Self {
+            format,
+            alpha_blend: BlendDescriptor::default(),
+            color_blend: BlendDescriptor::default(),
+            write_mask: ColorWrite::default(),
+        }
+    }
+
+    pub fn alpha_blend(&mut self, blend: BlendDescriptor) -> &mut Self {
+        self.alpha_blend = blend;
+        self
+    }
+
+    pub fn color_blend(&mut self, blend: BlendDescriptor) -> &mut Self {
+        self.color_blend = blend;
+        self
+    }
+
+    pub fn write_mask(&mut self, mask: ColorWrite) -> &mut Self {
+        self.write_mask = mask;
+        self
+    }
 }
 
 /// Primitive type the input mesh is composed of.
@@ -578,6 +667,34 @@ pub struct RasterizationStateDescriptor {
     pub depth_bias: i32,
     pub depth_bias_slope_scale: f32,
     pub depth_bias_clamp: f32,
+}
+
+impl RasterizationStateDescriptor {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn front_face(&mut self, front_face: FrontFace) -> &mut Self {
+        self.front_face = front_face;
+        self
+    }
+
+    pub fn cull_mode(&mut self, cull_mode: CullMode) -> &mut Self {
+        self.cull_mode = cull_mode;
+        self
+    }
+
+    pub fn clamp_depth(&mut self) -> &mut Self {
+        self.clamp_depth = true;
+        self
+    }
+
+    pub fn depth_bias(&mut self, bias: i32, slope_scale: f32, clamp: f32) -> &mut Self {
+        self.depth_bias = bias;
+        self.depth_bias_slope_scale = slope_scale;
+        self.depth_bias_clamp = clamp;
+        self
+    }
 }
 
 /// Underlying texture data format.
@@ -828,6 +945,39 @@ pub struct DepthStencilStateDescriptor {
 }
 
 impl DepthStencilStateDescriptor {
+    pub fn new(format: TextureFormat, depth_compare: CompareFunction) -> Self {
+        Self {
+            format,
+            depth_write_enabled: false,
+            depth_compare,
+            stencil_front: StencilStateFaceDescriptor::default(),
+            stencil_back: StencilStateFaceDescriptor::default(),
+            stencil_read_mask: 0,
+            stencil_write_mask: 0,
+        }
+    }
+
+    pub fn depth_write(&mut self) -> &mut Self {
+        self.depth_write_enabled = true;
+        self
+    }
+
+    pub fn stencil_front(&mut self, stencil: StencilStateFaceDescriptor) -> &mut Self {
+        self.stencil_front = stencil;
+        self
+    }
+
+    pub fn stencil_back(&mut self, stencil: StencilStateFaceDescriptor) -> &mut Self {
+        self.stencil_back = stencil;
+        self
+    }
+
+    pub fn stencil_masks(&mut self, read: u32, write: u32) -> &mut Self {
+        self.stencil_read_mask = read;
+        self.stencil_write_mask = write;
+        self
+    }
+
     pub fn needs_stencil_reference(&self) -> bool {
         !self.stencil_front.compare.is_trivial() || !self.stencil_back.compare.is_trivial()
     }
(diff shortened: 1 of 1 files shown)


wgpu-types/src/lib.rs
```diff
@@ -122,6 +122,25 @@ pub struct RequestAdapterOptions<S> {
...
 
+impl<S> RequestAdapterOptions<S> {
+    pub fn new() -> Self {
+        Self {
+            power_preference: PowerPreference::default(),
+            compatible_surface: None,
+        }
+    }
+
+    pub fn power_preference(&mut self, preference: PowerPreference) -> &mut Self {
```
aloucks, line 133: I generally think of builders as taking `mut self` as opposed to `&mut self` so that the final value is movable. @kvark what do you think?

Andful: I do somewhat agree. That would mean that the return value should be `Self`. @cwfitzgerald was there a particular reason for passing the value by reference over moving the value around?

cwfitzgerald: The reasoning behind using `&mut` builders is that it is more flexible when doing complex init, and because all our descriptors are accepted by reference, it can be passed in as is with no issues.

kvark: Let's proceed with `&mut self` like originally done by @cwfitzgerald . If users want to move things out, they can always `clone()`, but I imagine this is going to be a minority case.

wgpu-types/src/lib.rs
```diff
@@ -357,6 +376,31 @@ pub struct DeviceDescriptor {
     pub shader_validation: bool,
 }
 
+impl DeviceDescriptor {
+    pub fn new() -> Self {
+        Self {
+            features: Features::default(),
```
kvark, line 382: for things that derive `Default` we should just do `Self::default()` here

wgpu-types/src/lib.rs
```diff
@@ -580,6 +671,49 @@ pub struct RasterizationStateDescriptor {
...
+        self.front_face = front_face;
+        self
+    }
+
+    pub fn cull_mode(&mut self, cull_mode: CullMode) -> &mut Self {
+        self.cull_mode = cull_mode;
+        self
+    }
+
+    pub fn clamp_depth(&mut self, clamp: bool) -> &mut Self {
```
kvark, line 696: let's remove the boolean parameter

wgpu-types/src/lib.rs
```diff
@@ -580,6 +671,49 @@ pub struct RasterizationStateDescriptor {
...
+        self.clamp_depth = clamp;
+        self
+    }
+
+    pub fn depth_bias(&mut self, bias: i32) -> &mut Self {
```
kvark, line 701: let's combine all 3 depth bias parameters into a single call: they are related, and they mostly get set together

wgpu-types/src/lib.rs
```diff
@@ -828,6 +962,48 @@ pub struct DepthStencilStateDescriptor {
...
+            depth_write_enabled: false,
+            depth_compare: CompareFunction::Undefined,
+            stencil_front: StencilStateFaceDescriptor::default(),
+            stencil_back: StencilStateFaceDescriptor::default(),
+            stencil_read_mask: 0,
+            stencil_write_mask: 0,
+        }
+    }
+
+    pub fn depth_write_enabled(&mut self, enable: bool) -> &mut Self {
```
kvark, line 977: let's remove the parameter
I think we should rename the method to just `depth_write`, as other booleans don't have "_enabled" suffix

wgpu-types/src/lib.rs
```diff
@@ -1177,6 +1386,32 @@ impl<L> BufferDescriptor<L> {
...
+        self.label = Some(label.into_cow());
+        self
+    }
+
+    pub fn size(&mut self, size: BufferAddress) -> &mut Self {
+        self.size = size;
+        self
+    }
+
+    pub fn mapped_at_creation(&mut self, mapped: bool) -> &mut Self {
```
kvark, line 1409: let's remove the boolean argument

wgpu-types/src/lib.rs
```diff
@@ -1262,6 +1510,43 @@ pub struct SwapChainDescriptor {
     pub present_mode: PresentMode,
 }
 
+impl SwapChainDescriptor {
+    pub fn new() -> Self {
```
kvark, line 1514: let's accept the size here, since the descriptor isn't useful without it

wgpu-types/src/lib.rs
```diff
@@ -1262,6 +1510,43 @@ pub struct SwapChainDescriptor {
...
+        self.usage = usage;
+        self
+    }
+
+    pub fn format(&mut self, format: TextureFormat) -> &mut Self {
+        self.format = format;
+        self
+    }
+
+    pub fn width(&mut self, width: u32) -> &mut Self {
```
kvark, line 1534: and we can remove these

wgpu-types/src/lib.rs
```diff
@@ -1282,6 +1567,25 @@ pub struct RenderPassDescriptor<'a, C: Clone, D> {
     pub depth_stencil_attachment: Option<D>,
 }
 
+impl<'a, C: Clone, D> RenderPassDescriptor<'a, C, D> {
+    pub fn new() -> Self {
```
kvark, line 1571: similarly here, let's accept the color attachments in the `new()`

wgpu-types/src/lib.rs
```diff
@@ -1282,6 +1567,25 @@ pub struct RenderPassDescriptor<'a, C: Clone, D> {
...
+        Self {
+            color_attachments: Cow::Borrowed(&[]),
+            depth_stencil_attachment: None,
+        }
+    }
+
+    pub fn color_attachments(&mut self, color_attachments: impl IntoCow<'a, [C]>) -> &mut Self {
```
kvark, line 1578: and remove this one

wgpu-types/src/lib.rs
```diff
@@ -1737,6 +2083,60 @@ pub struct RenderPipelineDescriptor<'a, L, D> {
...
+        self.sample_count = count;
+        self
+    }
+
+    pub fn sample_mask(&mut self, mask: u32) -> &mut Self {
+        self.sample_mask = mask;
+        self
+    }
+
+    pub fn alpha_to_coverage_enabled(&mut self, enabled: bool) -> &mut Self {
```
kvark, line 2134: let's remove the argument, and strip the "_enabled" suffix from the method

wgpu-types/src/lib.rs
```diff
@@ -1786,6 +2227,17 @@ pub struct RenderBundleDescriptor<L> {
     pub label: L,
 }
 
+impl<'a> RenderBundleDescriptor<Option<Cow<'a, str>>> {
+    pub fn new() -> Self {
+        Self { label: None }
```
kvark, line 2232: similarly, let's call `Self::default()` for consistency

wgpu-types/src/lib.rs
```diff
@@ -2076,3 +2563,18 @@ pub struct TextureCopyView<T> {
     /// The base texel of the texture in the selected `mip_level`.
     pub origin: Origin3d,
 }
+
+impl<T> TextureCopyView<T> {
+    pub fn new(texture: T, origin: Origin3d) -> Self {
```
kvark, line 2568: let's remove the origin from here (defaulting to `ZERO`), and have a method for it

wgpu-types/src/lib.rs
```diff
@@ -122,6 +122,25 @@ pub struct RequestAdapterOptions<S> {
     pub compatible_surface: Option<S>,
 }
 
+impl<S> RequestAdapterOptions<S> {
+    pub fn new() -> Self {
+        Self {
+            power_preference: PowerPreference::default(),
```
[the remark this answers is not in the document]
Andful: `S: Default` is required. I will implement default for `RequestAdapterOptions`

wgpu-types/src/lib.rs
```diff
@@ -357,6 +376,31 @@ pub struct DeviceDescriptor {
...
+        self.features = features;
+        self
+    }
+
+    pub fn limits(&mut self, limits: Limits) -> &mut Self {
+        self.limits = limits;
+        self
+    }
+
+    pub fn shader_validation(&mut self, validation: bool) -> &mut Self {
```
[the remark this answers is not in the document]
Andful: Should the function name be `enable_shader_validation` in your opinion?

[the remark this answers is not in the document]
kvark: uh, that's a hard one... this choice affects all the boolean entry points.
My feeling is that just `shader_validation()` without prefixes or postfixes is fine. @cwfitzgerald @grovesNL please share your thoughts if you feel strongly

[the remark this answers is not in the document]
cwfitzgerald: I don't have particular feelings on the positive side, either thing() or enable_thing() work fine. Going to negative (which we don't right now) it should be disable_thing().

wgpu-types/src/lib.rs
Andful: Should `depth_write_enabled` and `depth_compare` be set together? And also, should `depth_compare` be initialized with `CompareFunction::Undefined`?

wgpu-types/src/lib.rs
```diff
@@ -828,6 +945,44 @@ pub struct DepthStencilStateDescriptor {
 }
 
 impl DepthStencilStateDescriptor {
+    pub fn new(format: TextureFormat) -> Self {
+        Self {
+            format,
+            depth_write_enabled: false,
+            depth_compare: CompareFunction::Undefined,
```
Andful, line 952: Should `depth_compare` be initialized as `CompareFunction::Undefined`? My worry is that this might fail silently.

wgpu-types/src/lib.rs
```diff
@@ -828,6 +945,44 @@ pub struct DepthStencilStateDescriptor {
...
+
+    pub fn depth_write(&mut self) -> &mut Self {
+        self.depth_write_enabled = true;
+        self
+    }
+
+    pub fn depth_compare(&mut self, compare: CompareFunction) -> &mut Self {
+        self.depth_compare = compare;
+        self
+    }
```
Andful, line 968: Should `depth_write` and `depth_compare` be combined?

wgpu-types/src/lib.rs
```diff
@@ -111,7 +111,7 @@ impl From<Backend> for BackendBit {
 
 /// Options for requesting adapter.
 #[repr(C)]
-#[derive(Clone, Debug, Default, PartialEq, Eq, Hash)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
```
Andful, line 114: Unfortunately because of [this issue](https://github.com/rust-lang/rust/issues/26925), I have to manually implement `Default`.

wgpu-types/src/lib.rs
```diff
@@ -472,6 +516,27 @@ impl BlendDescriptor {
     }
 }
 
+impl BlendDescriptor {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn src_factor(&mut self, factor: BlendFactor) -> &mut Self {
```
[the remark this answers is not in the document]
aloucks: I agree that it's probably not useful to only set one of them, but this breaks consistency and expectations. When looking for these functions you generally expect the function name to match the field `1:1`. I'd argue against using `factors(..)` simply to keep everything consistent.

wgpu-types/src/lib.rs
kvark: Sorry, I should have said. We need to accept it as a `new()` parameter instead of having a separate method.

wgpu-types/src/lib.rs
kvark: nope

wgpu-types/src/lib.rs
```diff
@@ -1262,6 +1478,33 @@ pub struct SwapChainDescriptor {
     pub present_mode: PresentMode,
 }
 
+impl SwapChainDescriptor {
+    pub fn new(width: u32, height: u32) -> Self {
+        Self {
+            usage: TextureUsage::OUTPUT_ATTACHMENT,
+            format: TextureFormat::Bgra8UnormSrgb,
```
Andful, line 1485: This default value is ok?

Andful:
Made the (hopefully) last change. Could you check my last comment?

wgpu-types/src/lib.rs
Andful: At least [here](https://github.com/gfx-rs/wgpu-rs/blob/ffa9199402e1436ba525007754e161b08ab12bcc/examples/framework.rs#L201) it uses `TextureFormat::Bgra8Unorm` but it also states `// TODO: Allow srgb unconditionally`

wgpu-types/src/lib.rs
```diff
@@ -1502,7 +1759,40 @@ pub struct TextureViewDescriptor<L> {
...
+        self.level_count = level_count;
+        self
+    }
+
+    pub fn array_layers(&mut self, base_array_layer: u32, array_layer_count: Option<NonZeroU32>) -> &mut Self {
+        self.base_array_layer = base_array_layer;
+        self.array_layer_count = array_layer_count;
+        self
+    }
+
```
Andful, line 1795: @kvark Could you double check this? I accidentally missed its implementation previously.

wgpu-types/src/lib.rs
```diff
@@ -1502,7 +1759,47 @@ pub struct TextureViewDescriptor<L> {
...
+    pub fn array_layers(
+        &mut self,
+        base_array_layer: u32,
+        array_layer_count: Option<u32>,
+    ) -> &mut Self {
+        self.base_array_layer = base_array_layer;
+        self.array_layer_count = array_layer_count;
+        self
+    }
+
```
Andful, line 1802: @kvark Could you double check this? I accidentally missed its implementation previously. Also, should `level_count` and  `array_layer_count` checked to be different from `0`?

kvark: sorry, I should have spotted that. Please separate these both into methods that change base vs count.
thank you for the heads up!

wgpu-types/src/lib.rs
kvark: I think it's ok defaulting to `Bgra8UnormSrgb` here. WASM users can always work around by using non-default, and we want to fix this in WASM too

kvark:
@Andful thank you! Would you be interested to update wgpu-rs based on this PR?

Andful:
Yes, I would be. It is only the example code, is it not?

kvark:
> Yes, I would be. It is only the example code, is it not?

likely, yes

PR #776, opened 2020-07-10
Add Error State to Storages

kunalmohan:
**Connections**
_Link to the issues addressed by this PR, or dependent PRs in other repositories_
fix #772 

**Description**
_Describe what problem this is solving, and how it's solved._
This aims to replace `VecMap` with a simple `Vec<Element<T>>` to keep track of valid and error Ids in Storages. This lays the foundations for #638. A new `enum Element<T>` is introduced as described in the linked issue.
The `Error` Variant is not used anywhere at the moment.

**Testing**
_Explain how this change is tested._
A Successful build is the only check done at the moment.

PR #857, opened 2020-07-31
Separate valid internal IDs from external ones

### LINKED ISSUE #638: Error model

kvark:
**Connections**
Closes #638 
wgpu-rs update - https://github.com/gfx-rs/wgpu-rs/pull/494

**Description**
The core change here is to allow user-specified IDs to be in the "Error" state that was introduced in #776 .

This is done by defining an internal `Valid<I>` wrapper. Now, the hub storages can be indexed by this valid wrapper only. For regular IDs, we have to go through `storage.get(index)`, which returns a `Result`. It still panics if the ID is totally garbage though, we don't want to handle what can't be expected here.

All the other changes come mostly as a consequence of that:
  - new "Invalid*" error variants are added
  - the error types are undergone sever refactoring
  - new `command/draw.rs` module for stuff shared between render passes and bundles
  - functions to generate error IDs for all the types
  - various renames, e.g. `comb` -> `cmd_buf`

The expected use by wgpu-rs is unchanged. So far, I don't think we'd be generating Error IDs, but we can always reconsider.
For browsers though, if `device_create_xxx` failed, they would generate and error ID. It will occupy the slot up until the corresponding JS object is dropped.

**Testing**
Tested on wgpu-rs examples

### DIFF
--- a/player/src/bin/play.rs
+++ b/player/src/bin/play.rs
@@ -67,7 +67,7 @@ fn main() {
         Some(trace::Action::Init { desc, backend }) => {
             log::info!("Initializing the device for backend: {:?}", backend);
             let adapter = global
-                .pick_adapter(
+                .request_adapter(
                     &wgc::instance::RequestAdapterOptions {
                         power_preference: wgt::PowerPreference::Default,
                         #[cfg(feature = "winit")]
@@ -82,7 +82,7 @@ fn main() {
                 )
                 .expect("Unable to find an adapter for selected backend");
 
-            let info = gfx_select!(adapter => global.adapter_get_info(adapter));
+            let info = gfx_select!(adapter => global.adapter_get_info(adapter)).unwrap();
             log::info!("Picked '{}'", info.name);
             gfx_select!(adapter => global.adapter_request_device(
                 adapter,
--- a/wgpu-core/src/id.rs
+++ b/wgpu-core/src/id.rs
@@ -53,8 +53,8 @@ impl<T> From<SerialId> for Id<T> {
 
 impl<T> Id<T> {
     #[cfg(test)]
-    pub(crate) fn dummy() -> Self {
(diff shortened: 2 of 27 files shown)


wgpu-core/src/id.rs
```diff
@@ -110,6 +110,14 @@ impl<T> Ord for Id<T> {
     }
 }
 
+/// An internal ID that has been checked to point to point to
```
kunalmohan, line 113: Remove the extra `point to`

player/src/bin/play.rs
```diff
@@ -82,7 +82,7 @@ fn main() {
                 )
                 .expect("Unable to find an adapter for selected backend");
 
-            let info = gfx_select!(adapter => global.adapter_get_info(adapter));
+            let info = gfx_select!(adapter => global.adapter_get_info(adapter)).unwrap();
```
cwfitzgerald, line 85: Not in scope, but we likely want an `unwrap_pretty` type thing here too.

kvark: These are actually not expected to fail at all, unless we are hand-writing them.
I would be happy to accept such PR anyway though :)

kvark:
bors r=cwfitzgerald

PR #860, opened 2020-08-01
Update naga and fix type widths

kvark:
**Connections**
Related to https://github.com/gfx-rs/naga/pull/98

**Description**
Naga's "width" has type `Bytes`, so we should treat it as such.

**Testing**
How did it use to work? Oh right, it didn't.


kvark:
bors r=mechanical

PR #863, opened 2020-08-05
fixed lifetime inconsistency when assigning string literal as label

Andful:
**Connections**
It is connected to the PR #856 

**Description**
When assigning the `label` field with a string literal, the compiler would require that the lifetime `'a` is `'static`, whether we only need the string literal to outlive `'a` (which it clearly does because it's lifetime is `'static`)
**Testing**

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -1365,7 +1365,7 @@ impl<'a> BufferDescriptor<Option<Cow<'a, str>>> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


kvark:
I don't understand what this PR is trying to do. I'm pretty sure the changes here are non-functional, i.e. there is no user code that would compile differently with this PR than without it.

> whether we only need the string literal to outlive 'a (which it clearly does because it's lifetime is 'static)

The only way to outlive a `'static` lifetime is to be static.

PR #865, opened 2020-08-05
Replace u32 with NonZeroU32 in TextureView

kunalmohan:
**Connections**

**Description**
We now expose separate methods under `Global` to register error/invalid resources. So the check for `mip_level_count = 0` and `array_layer_count = 0` can be made on the client-side in Servo, and directly register an error resource instead of trying to create a real one.

**Testing**
Not yet tested. Will be done wgpu-rs examples.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -10,7 +10,7 @@
 
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
-use std::{borrow::Cow, ops::Range};
+use std::{borrow::Cow, num::NonZeroU32, ops::Range};
 
 pub use into_cow::IntoCow;
 mod into_cow;
@@ -1750,13 +1750,13 @@ pub struct TextureViewDescriptor<L> {
     /// Mip level count.
     /// If `Some(count)`, `base_mip_level + count` must be less or equal to underlying texture mip count.
     /// If `None`, considered to include the rest of the mipmap levels, but at least 1 in total.
-    pub level_count: Option<u32>,
+    pub level_count: Option<NonZeroU32>,
     /// Base array layer.
     pub base_array_layer: u32,
     /// Layer count.
     /// If `Some(count)`, `base_array_layer + count` must be less or equal to the underlying array count.
     /// If `None`, considered to include the rest of the array layers, but at least 1 in total.
-    pub array_layer_count: Option<u32>,
+    pub array_layer_count: Option<NonZeroU32>,
 }
 
 impl<'a> TextureViewDescriptor<Option<Cow<'a, str>>> {
@@ -1786,7 +1786,7 @@ impl<L> TextureViewDescriptor<L> {
     }
 
     pub fn level_count(&mut self, level_count: u32) -> &mut Self {
-        self.level_count = Some(level_count);
+        self.level_count = Some(NonZeroU32::new(level_count).unwrap());
         self
     }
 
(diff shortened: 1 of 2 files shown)


wgpu-types/src/lib.rs
```diff
@@ -1785,7 +1785,7 @@ impl<L> TextureViewDescriptor<L> {
         self
     }
 
-    pub fn level_count(&mut self, level_count: u32) -> &mut Self {
+    pub fn level_count(&mut self, level_count: NonZeroU32) -> &mut Self {
```
kvark, line 1788: this should accept `u32` still, and just panic on zero

wgpu-types/src/lib.rs
```diff
@@ -1795,7 +1795,7 @@ impl<L> TextureViewDescriptor<L> {
         self
     }
 
-    pub fn array_layer_count(&mut self, array_layer_count: u32) -> &mut Self {
+    pub fn array_layer_count(&mut self, array_layer_count: NonZeroU32) -> &mut Self {
```
kvark, line 1798: same here

wgpu-types/src/lib.rs
```diff
@@ -1786,7 +1786,8 @@ impl<L> TextureViewDescriptor<L> {
     }
 
     pub fn level_count(&mut self, level_count: u32) -> &mut Self {
-        self.level_count = Some(level_count);
+        self.level_count
```
kvark, line 1789: why are we replacing instead of just assigning to it?

wgpu-types/src/lib.rs
```diff
@@ -1796,7 +1797,8 @@ impl<L> TextureViewDescriptor<L> {
     }
 
     pub fn array_layer_count(&mut self, array_layer_count: u32) -> &mut Self {
-        self.array_layer_count = Some(array_layer_count);
+        self.array_layer_count
+            .replace(NonZeroU32::new(array_layer_count).expect("Invalid array_layer_count"));
```
kvark, line 1801: nit: I think just `unwrap()` is fine. We can hardly expect anybody to pass `0` there, and it will be fairly clear why this fails

wgpu-types/src/lib.rs
kunalmohan: No particular reason. Is there a downside to it?

kvark: assignment is simpler. In the lack of other criteria, choose the simplest one

kunalmohan:
Tested on wgpu-rs examples, works fine!

wgpu-types/src/lib.rs
```diff
@@ -1786,7 +1786,7 @@ impl<L> TextureViewDescriptor<L> {
     }
 
     pub fn level_count(&mut self, level_count: u32) -> &mut Self {
-        self.level_count = Some(level_count);
+        self.level_count = Some(NonZeroU32::new(level_count).unwrap());
```
kvark, line 1789: now that I see it, I wonder if we should just do `level_count = NonZero::U32(...)`

kunalmohan: We would be checking for 0 then. Or should we add a separate `if` for that?

kvark: yeah, I mean, it's a builder, and passing 0 explicitly is a weird case
anyhow, we can keep that in mind, no need to do these changes, and there isn't an agreement on it yet :)

PR #870, opened 2020-08-09
Implicit layout

### LINKED ISSUE #868: Make pipeline layout objects optional
See https://github.com/gpuweb/gpuweb/pull/543 `getBindGroupLayout`.
Apparently, WebGPU CTS relies a lot on this behavior :/

When creating pipelines, the layout would now be optional. If not provided, it's derived from the shader source according to the set of rules outlined in the spec. Users can then get the bind group layout object, to create the bind groups with.

Not much we can do - this is in upstream spec now.

**Additional context**
This may be blocked on https://github.com/gfx-rs/naga progress and WGSL adoption - it's easier to extract the data from naga modules.

kvark:
**Connections**
Closes #868

**Description**
The implementation can be split into 3 parts:
  1. reflecting the shader for binding expectations, and building a bind entry map from it, merging them between stages. This is only done for shaders that can be reflected, and we error on the rest, for now.
  2. based on this info, create new bind group layouts and pipeline layouts. The tricky part here is that we can't generate the ID out of thin air, so we have to pass them into the `create_xx_pipeline` function, which now also returns the number of IDs it consumed, allowing the client to free the rest.
  3. API changes in the descriptors, new methods to obtain the bind group layouts from a pipeline

**Testing**
This isn't tested, but I think it's fine: it doesn't affect the old path, and we'll be testing the new path while improving Naga and our reflection anyway.

### DIFF
--- a/player/src/lib.rs
+++ b/player/src/lib.rs
@@ -248,7 +248,7 @@ impl GlobalPlay for wgc::hub::Global<IdentityPassThroughFactory> {
             }
             A::CreateComputePipeline(id, desc) => {
                 self.device_maintain_ids::<B>(device).unwrap();
-                self.device_create_compute_pipeline::<B>(device, &desc, id)
+                self.device_create_compute_pipeline::<B>(device, &desc, id, None)
                     .unwrap();
             }
             A::DestroyComputePipeline(id) => {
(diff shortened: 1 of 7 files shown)


cwfitzgerald:
Looks great!
bors r+

PR #872, opened 2020-08-10
Respect texture view aspect on creation

kvark:
**Connections**
Fixing https://github.com/gfx-rs/wgpu-rs/issues/498 (but needs wgpu-rs update then)

**Description**
We added the aspect field to texture view descriptor but didn't respect it properly.
We do that now, and also disallow Depth-Stencil sampled views

**Testing**
Used the modified water example for testing.

### DIFF
--- a/wgpu-core/src/resource.rs
+++ b/wgpu-core/src/resource.rs
@@ -276,6 +276,11 @@ pub enum CreateTextureViewError {
     InvalidMipLevelCount { requested: u32, total: u8 },
     #[error("TextureView array layer count + base array layer {requested} must be <= Texture depth/array layer count {total}")]
     InvalidArrayLayerCount { requested: u32, total: u16 },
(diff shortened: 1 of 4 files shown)


wgpu-core/src/resource.rs
```diff
@@ -276,6 +276,8 @@ pub enum CreateTextureViewError {
     InvalidMipLevelCount { requested: u32, total: u8 },
     #[error("TextureView array layer count + base array layer {requested} must be <= Texture depth/array layer count {total}")]
     InvalidArrayLayerCount { requested: u32, total: u16 },
+    #[error("Aspect {requested:?} is not in the source texture ({total:?}")]
```
cwfitzgerald, line 279: Nit: missing closing parens

kvark:
Addressed all the notes, thanks for the review!
bors r=cwfitzgerald

PR #873, opened 2020-08-12
Grand cleanup and refactor of the descriptors in the API

### LINKED ISSUE #848: Refactor TextureViewDescriptor to match upstream spec better
In WebGPU [spec](https://gpuweb.github.io/gpuweb/#dictdef-gputextureviewdescriptor), the descriptor has all the fields with meaningful default values. In `wgpu-types` today, the format and dimensions of the view are required fields. And `texture_create_view` method accepts `Option<&wgt::TextureViewDescriptor<Label>>`. We should instead make it so `TextureViewDescriptor::default()` is implemented, and matches the case in WebGPU where the user doesn't specify any fields, i.e `texture.createView({})`.

In order to implement `Default` for this structure, we need to turn `format` and `dimension` into options. Then we'd implement `Default`, and then we'd refactor `texture_create_view` into accepting a descriptor without an option. Finally, we'll update `wgpu-rs` to use the new API.

### LINKED ISSUE #871: Move subscriber logic out into a separate crate
Subscriber feature seems to be enabling the code paths that don't need to be in `wgpu`. Only the final consumer/application would care about this.

Create a little crate called `wgpu-subscriber` and move the stuff in there

leave as is

kvark:
**Connections**
Cleanup follows https://github.com/gfx-rs/wgpu-rs/pull/501
Fixes #871 
Fixes #848

**Description**
There is a lot of small and big things crumbled in here.

The major one is where descriptors live, and how they are parametrized. Logic is the following:
  - if something is useful by wgpu-rs's API leave it in `wgpu-types`
  - if *in addition* it's useful to `wgpu-native`(i.e. has `repr(C)`),  we possibly parametrize it (currently, only label is).
  - otherwise, the type is moved to `wgpu-core` and stripped of generics
  - remove all the builders

Some medium-sized things:
  - moves the subscriber out (#871)
  - moves the `RenderCommand` into the `draw` module (which is meant to contain shared things)
  - makes `TextureViewDescriptor` fields optional, but receives it directly now (#848)
  - moves stencil stuff out into `StencilStateDescriptor`, so that we can derive `Default` for it
  - stop accepting raw strings for labels in the `Device` API: neither the clients, or gfx-rs need that shape, and it can't be safe

Some smaller things:
  - using NonZeroU8 for anisotropy
  - using NonZeroU32 for descriptor count
  - putting sampler addressing modes into an array
  - add labels to command buffers and pipeline layouts
  - improves errors for exceeding binding limits

**Testing**
Tested on https://github.com/gfx-rs/wgpu-rs/pull/503


wgpu-core/src/device/trace.rs
```diff
@@ -14,13 +14,13 @@ type FileName = String;
...
-) -> wgt::RenderBundleEncoderDescriptor {
-    wgt::RenderBundleEncoderDescriptor {
-        label: Some(super::own_label(&label).into()),
-        color_formats: context.attachments.colors.to_vec().into(),
+pub(crate) fn new_render_bundle_encoder_descriptor<'a>(
+    label: Option<&'a str>,
+    context: &'a super::RenderPassContext,
+) -> crate::command::RenderBundleEncoderDescriptor<'a> {
+    crate::command::RenderBundleEncoderDescriptor {
+        label: label.map(Cow::Borrowed),
```
Kimundi, line 22: It might be more consistent to import `Borrowed` directly everywhere.

kvark: I don't think `wgpu` crate imports `Borrowed` directly anywhere

wgpu-types/src/lib.rs
```diff
@@ -830,12 +841,17 @@ pub struct StencilStateDescriptor {
 }
 
 impl StencilStateDescriptor {
-    pub fn needs_ref_value(&self) -> bool {
-        !self.front.compare.is_trivial() || !self.back.compare.is_trivial()
+    pub fn is_enabled(&self) -> bool {
+        (self.front != StencilStateFaceDescriptor::IGNORE
+            || self.back != StencilStateFaceDescriptor::IGNORE)
+            && self.read_mask != 0
```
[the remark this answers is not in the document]
kvark: well, the old test was wrong :) you can have a perfectly functional stencil test without writing anything

kvark:
bors r=cwfitzgerald,grovesNL

wgpu-core/src/device/trace.rs
Kimundi: Ah sorry, could have sworn I did the same as in wgpu-rs there.

PR #875, opened 2020-08-13
Add labels to pipelines

kvark:
This isn't functional until gfx-hal-0.6 is published, but the API is there.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1594,7 +1594,7 @@ dependencies = [
 
 [[package]]
 name = "wgpu-subscriber"
-version = "0.5.0"
+version = "0.1.0"
 dependencies = [
  "parking_lot",
  "thread-id",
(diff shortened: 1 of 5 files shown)


kvark:
Also, changes the subscriber version to 0.1, since it's not tied in any way to the rest of the crates. We could even move it out completely in a repo, and should consider that. cc @cwfitzgerald 
bors r=trivial

cwfitzgerald:
Back when I first started working on it, there was interest in moving the chrome tracing infra into tracing itself. The only current issue would be that we allow marking certain events as relevant to traces, which isn't a generic solution.

I am in general against moving things into their own repo, but this is independent enough that I agree we should consider it.

kvark:
@cwfitzgerald I'd be totally cool with it staying, but we'll be questioned about why it's there, since the whole repo is synced to mozilla-central :(
Would you want to perform the move-out? I created https://github.com/gfx-rs/subscriber and invited you into it.

PR #877, opened 2020-08-15
Add image cube array feature.

StarArawn:
**Connections**
None

**Description**
Enable `IMAGE_CUBE_ARRAY` in vulkan and dx12. Note: Currently this will cause metal to break as the `available_features` from the physical device will never include `IMAGE_CUBE_ARRAY`. Thus I'm opening this as a draft until gfx-hal metal can check and pass the feature on.

**Testing**

I can test this with my repo using wgpu-rs here which uses cube arrays:
https://github.com/StarArawn/harmony/

### DIFF
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -652,7 +652,8 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 | hal::Features::FRAGMENT_STORES_AND_ATOMICS
                 | hal::Features::NDC_Y_UP
                 | hal::Features::INDEPENDENT_BLENDING
-                | hal::Features::SAMPLER_ANISOTROPY;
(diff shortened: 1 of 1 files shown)


kvark:
> this will cause metal to break as the

I don't understand this. Please clarify. If Metal backend doesn't expose this feature, you can't enable it in `wgpu` and thus you can't break anything.

StarArawn:
@kvark 
Ahh, sorry I mean if you use metal it will falsely warn about the feature not being available when it actually is.

kvark:
ok, lemme fix that in Metal for gfx-0.6 release

kvark:
don't wait for it though

kvark:
@StarArawn could you update this?

kvark:
Could you squash please?

PR #885, opened 2020-08-19
Improve buffer and texture overrun error messages

### LINKED ISSUE #884: Copy buffer-to-buffer error
**Description**
Error:
> copy would end up overruning the bounds of the destination buffer/texture

**Repro steps**
Code is in https://github.com/nical/lyon/blob/c921b427b6708665cf586302812170923fe302df/examples/wgpu_svg/src/main.rs#L478-L484

Looks like it should be working.

cwfitzgerald:
**Connections**

Closes #884.

**Description**

The previous errors about buffer and texture overruns didn't tell the user any information about which buffer, how long wgpu thought the copy was, or how long it thought the buffer was. This makes the error message much better.

Unfortunately a braking change, so can't backport :(

**Testing**

Tested on #884's issue.

### DIFF
--- a/wgpu-core/src/command/transfer.rs
+++ b/wgpu-core/src/command/transfer.rs
@@ -26,6 +26,19 @@ pub(crate) const BITS_PER_BYTE: u32 = 8;
 pub type BufferCopyView = wgt::BufferCopyView<BufferId>;
 pub type TextureCopyView = wgt::TextureCopyView<TextureId>;
 
+#[derive(Clone, Debug)]
+pub enum CopySide {
+    Source,
+    Destination,
+}
+
+#[derive(Clone, Debug)]
+pub enum TextureErrorDimension {
+    X,
+    Y,
+    Z,
+}
+
 /// Error encountered while attempting a data transfer.
 #[derive(Clone, Debug, Error)]
 pub enum TransferError {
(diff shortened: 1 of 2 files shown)


wgpu-core/src/command/transfer.rs
```diff
@@ -19,13 +19,45 @@ use hal::command::CommandBuffer as _;
...
-use std::iter;
+use std::{fmt, iter};
 
 pub(crate) const BITS_PER_BYTE: u32 = 8;
 
 pub type BufferCopyView = wgt::BufferCopyView<BufferId>;
 pub type TextureCopyView = wgt::TextureCopyView<TextureId>;
 
+#[derive(Clone, Debug)]
+pub enum ResourceSide {
```
cwfitzgerald, line 30: If you have a better name, hit me.

wgpu-core/src/command/transfer.rs
```diff
@@ -19,13 +19,45 @@ use hal::command::CommandBuffer as _;
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Source => write!(f, "source"),
+            Self::Destination => write!(f, "destination"),
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub enum TextureErrorDimension {
```
cwfitzgerald, line 45: Ditto

wgpu-core/src/command/transfer.rs
kvark: `CopySide`?

wgpu-core/src/command/transfer.rs
```diff
@@ -19,13 +19,45 @@ use hal::command::CommandBuffer as _;
...
 pub type BufferCopyView = wgt::BufferCopyView<BufferId>;
 pub type TextureCopyView = wgt::TextureCopyView<TextureId>;
 
+#[derive(Clone, Debug)]
+pub enum ResourceSide {
+    Source,
+    Destination,
+}
+
+impl fmt::Display for ResourceSide {
```
kvark, line 35: why not just derive it?

wgpu-core/src/command/transfer.rs
```diff
@@ -19,13 +19,45 @@ use hal::command::CommandBuffer as _;
...
+}
+
+#[derive(Clone, Debug)]
+pub enum TextureErrorDimension {
+    X,
+    Y,
+    Z,
+}
+
+impl fmt::Display for TextureErrorDimension {
```
kvark, line 51: why not derived?

wgpu-core/src/command/transfer.rs
cwfitzgerald: Because then it would be the wrong capitalization for the error message

kvark: sure, but it's an error message, so it's fine. I.e. you can put it into quotes to separate from the rest of the message
but it would save code, and that's a good thing to have

cwfitzgerald: Fixed.

PR #893, opened 2020-08-24
Improve texture creation validation

kvark:
**Connections**
Fixes https://github.com/gfx-rs/wgpu-rs/issues/528#issuecomment-678853886

**Description**
Improves the mipmap levels check.
Note that it also removes the `TooManyLayers` error, and instead produces max `u16` value. This is to be followed up with a check for *real* limits, which has to happen regardless, and trying to catch it on the number conversion step seems wasteful.

**Testing**

### DIFF
--- a/wgpu-core/src/resource.rs
+++ b/wgpu-core/src/resource.rs
@@ -213,10 +213,17 @@ pub struct Texture<B: hal::Backend> {
     pub(crate) life_guard: LifeGuard,
 }
 
+#[derive(Clone, Debug)]
+pub enum TextureErrorDimension {
+    X,
+    Y,
+    Z,
+}
+
 #[derive(Clone, Debug, Error)]
 pub enum TextureDimensionError {
-    #[error("too many layers ({0}) for texture array")]
-    TooManyLayers(u32),
+    #[error("Dimension {0:?} is zero")]
+    Zero(TextureErrorDimension),
     #[error("1D textures must have height set to 1")]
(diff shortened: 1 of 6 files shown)


wgpu-core/src/resource.rs
```diff
@@ -215,8 +215,8 @@ pub struct Texture<B: hal::Backend> {
 
 #[derive(Clone, Debug, Error)]
 pub enum TextureDimensionError {
-    #[error("too many layers ({0}) for texture array")]
-    TooManyLayers(u32),
+    #[error("one of the dimensions is zero")]
+    Zero,
```
cwfitzgerald, line 219: Could use [`TextureErrorDimension`](https://github.com/gfx-rs/wgpu/blob/master/wgpu-core/src/command/transfer.rs#L36) to make this error better.

cwfitzgerald: And if you want to get real fancy, to detect this, you could do

```rust
match (width, height, depth) {
    (0, _, _) => return Err(Zero(TextureErrorDimension::X)),
    (_, 0, _) => return Err(Zero(TextureErrorDimension::Y)),
    (_, _, 0) => return Err(Zero(TextureErrorDimension::Z)),
    _ => {}
}
```

kvark:
Thanks! Using that enum now.
bors r=cwfitzgerald

kvark:
bors r=cwfitzgerald

PR #895, opened 2020-08-24
Improve validation in B2B copies and RenderCommands

kunalmohan:
**Connections**

**Description**
Validation for the following have been added-
- Validate all parameters in B2B copy even if `copy_size == 0`.
- Check for copy operations issued within same buffer.
- `RenderPass.setViewport()`
- `RenderPass.setScissorRect()`

**Testing**
Tested with CTS in Servo. More tests pass now. Will test on wgpu-rs now.

### DIFF
--- a/wgpu-core/src/command/render.rs
+++ b/wgpu-core/src/command/render.rs
@@ -316,6 +316,8 @@ pub enum RenderPassError {
     Encoder(#[from] CommandEncoderError),
     #[error("attachment texture view {0:?} is invalid")]
     InvalidAttachment(id::TextureViewId),
+    #[error("attachments have different sizes")]
+    MismatchAttachments,
     #[error("attachment's sample count {0} is invalid")]
     InvalidSampleCount(u8),
     #[error("attachment with resolve target must be multi-sampled")]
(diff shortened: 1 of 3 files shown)


kunalmohan:
I am not sure when `BufferUsage::QUERY_RESOLVE` is used, but would it make sense to add it to wgpu-types at the moment?

kvark:
We don't have the occlusion queries implemented yet, so the `BufferUsage` can be added once we have those.

wgpu-core/src/command/render.rs
```diff
@@ -568,6 +571,9 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                         .map_err(|_| RenderPassError::InvalidAttachment(at.attachment))?;
                     add_view(view)?;
 
+                    attachment_width = attachment_width.min(view.extent.width);
```
kvark, line 574: one thing missing is making sure that all attachments have the same size

wgpu-core/src/command/transfer.rs
```diff
@@ -46,6 +46,8 @@ pub enum TransferError {
     InvalidBuffer(BufferId),
     #[error("texture {0:?} is invalid")]
     InvalidTexture(TextureId),
+    #[error("Source and destination cannot be the same buffer")]
+    AttemptToCopyWithinBuffer,
```
kvark, line 50: Maybe "SameSourceDestinationBuffer"?

PR #901, opened 2020-08-27
Allow copying from depth textures

### LINKED ISSUE #900: Copying from depth texture to buffer fails
**Description**
Copying from depth textures fails with `unexpected depth format`. 

**Repro steps**
In short:
- create texture with `Depth32Float` format
- call `device.copy_texture_to_buffer`

Will try to provide a complete code example soon.

Code panics with `unexpected depth format` ([here](https://github.com/gfx-rs/wgpu/blob/v0.6/wgpu-core/src/conv.rs#L444)). Expect it to work fine and copy the texture to the target buffer.

**Platform**
wgpu v0.6

malu:
**Connections**
Fixes #900 

**Description**
Copying from depth textures failed with `unexpected depth format` before. Fix the cause in `conv::texture_block_size`. Add checks around the place where it would have (rightfully) failed before.

Seeing as this adds another variant to `TransferError`, this is not backwards compatible. The compatible alternative would be `panic`ing (which is what currently happens via `unreachable!`).
Also, I could understand if you think `TextureFormat::is_depth_format` isn't worth it to have as public api. We could pull it into `wgpu-core` in that case. It would need a doc comment in case we want to keep it there.

Should this instead be based on master instead of v0.6? Just let me know and I can rebase.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -809,6 +809,17 @@ pub enum TextureFormat {
     Bc7RgbaUnormSrgb = 51,
 }
 
+impl TextureFormat {
+    pub fn is_depth_format(self) -> bool {
+        match self {
+            TextureFormat::Depth32Float
+            | TextureFormat::Depth24Plus
+            | TextureFormat::Depth24PlusStencil8 => true,
+            _ => false,
+        }
+    }
+}
+
 bitflags::bitflags! {
     /// Color write mask. Disabled color channels will not be written to.
     #[repr(transparent)]
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -809,17 +809,6 @@ pub enum TextureFormat {
     Bc7RgbaUnormSrgb = 51,
 }
 
-impl TextureFormat {
-    pub fn is_depth(self) -> bool {
-        match self {
-            TextureFormat::Depth32Float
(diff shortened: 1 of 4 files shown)


kvark:
> Should this instead be based on master instead of v0.6? Just let me know and I can rebase.

Since this PR is breaking, it should definitely be targeted at `master`.
We can produce a sibling non-breaking PR for 0.6

wgpu-types/src/lib.rs
```diff
@@ -809,6 +809,17 @@ pub enum TextureFormat {
     Bc7RgbaUnormSrgb = 51,
 }
 
+impl TextureFormat {
+    pub fn is_depth(self) -> bool {
+        match self {
+            TextureFormat::Depth32Float
```
kvark, line 815: Another thing I forgot to mention on the review (sorry!).
We don't want copies from depth24plus formats. Only Depth32Float is allowed.

malu: Ah, don't worry. Thanks for checking.

In that case, I think I will remove this function from wgpu-types since it doesn't seem useful anymore. And instead add `is_valid_copy_destination_texture_format` and `is_valid_copy_source_texture_format` to wgpu-core. Which are again quite long, but since we can't add it to `TextureFormat` inside of wgpu-core it seems like a reasonable choice.

PR #903, opened 2020-08-28
fix for mixed STORAGE_STORE and STORAGE_LOAD

kocsis1david:
**Connections**
None

**Description**
A buffer that contains dynamic data is used for both uniforms and storage read. This was a problem for wgpu and it gave a `PendingTransition` error.

**Testing**
It's a trivial fix, but I don't know if there's an example for it in wgpu-rs that can be used for testing.

It seems that the PR "Sync changes from mozilla-central" is already solving the same issue.


kvark:
Also, a candidate to backporting to 0.6

PR #904, opened 2020-08-28
[0.6] Allow copying from Depth32Float formatted textures

malu:
**Connections**
Backport of #901 

**Description**
Copying from textures with format `Depth32Float` was reaching an `unreachable` and hence was impossible while copying from such textures should be well-defined and safe.


kvark:
argh, that's the `generator` issue that's still not fixed

Issue #905, opened 2020-08-28
[meta] patch 0.6.1


kvark:
- [x] #903
- [x] #901 (needs to be adjusted for backwards-compat) - #904
- [x] #886
- [x] #883
- [x] #912
- [x] #916

PR #906, opened 2020-08-28
[0.6] More backports

malu:
**Connections**
Remaining backports listed in #905.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # Change Log
 
 ## v0.6.1 (tbd)
-  - allow copying from Depth32Float textures
+  - allow copying from `Depth32Float` textures
+  - fix usage tracking of read-only buffer storages
 
 ## v0.6 (2020-08-17)
   - Crates:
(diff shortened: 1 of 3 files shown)


kvark:
actually, can we exclude the wgpu-types version bump? I don't want to release a patch version of wgpu-types just for this, yet

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## v0.6.1 (tbd)
   - allow copying from Depth32Float textures
+  - fix for mixed STORAGE_STORE and STORAGE_LOAD
```
kvark, line 5: nit: since `STORAGE_STORE` is not something users see, we should rather say something like "fix usage tracking of read-only buffer storages"

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## v0.6.1 (tbd)
   - allow copying from Depth32Float textures
```
kvark, line 4: let's quote the actual type here i.e. `Depth32Float`

malu: Sorry, not sure what you mean? Just the backticks?

CHANGELOG.md
malu: Ah, I see. I just copied the commit message. I'll fix that.

Issue #450, opened 2020-01-14
Is OpenGL Supported now?


zicklag:
With #183 getting closed in wgpu-rs does that mean that OpenGL is now supported here and we can update the checkmark on the README?

kvark:
GL was removed some time after we added it. Currently, it's not supported.
We'll be happy to accept patches that re-enable it. This may become a priority for us in the future, but it's not the most pressing thing right now.

zicklag:
Cool, thanks for the clarification. :+1:

I'm good with closing this then, unless you think it should be left open. :slightly_smiling_face:

kvark:
Well, we don't have an issue open for GL, so I'm thinking of leaving this one open.

zicklag:
At this point, would all that is necessary to get this working be to add the GL gfx backend code just like the Vulkan backend code? I was just messing around with it ( I really hardly any idea what I'm doing ) and I think I got everything except `wgpu-native` to compile on Linux. With `wgpu-native` I hadn't figured out how to do the window creation for the GL backend yet.

I pretty much just went through and whenever there were feature gates for Linux I added a section that would use the Gl backend instead of the Vulkan backend if the GL feature was enabled. It was a little haphazard and needs to be organized, but as far as I can tell the last step was window creation.

Am I in the right direction or is there more that is going to need to be done?

kvark:
@zicklag the problem here is to minimize any deviation from the existing paths. Ideally, the code should support GL fully transparently, but this requires gfx-backend-gl to be smarter about its surfaces, i.e. https://github.com/gfx-rs/gfx/issues/3133

You can post your code as a draft PR, and we'll be happy to see if you are moving in the right direction :)

zicklag:
OK, cool. After looking around more I see how the surface creation is `unimplemented!()`. I'll do some looking around to see if I can get a better grasp on the API. If I come up with anything I'll draft a PR.

zicklag:
So, as far as I understand, as shown in the gfx quad example, you can [create an OpenGL surface](https://github.com/gfx-rs/gfx/blob/master/examples/quad/main.rs#L143) from a context, split from a windowed context, that was built with Glutin, but in `wgpu-native` I we already have a raw window handle that we want to create the surface for, and I don't know how to create a glutin Context from the raw window handle.

It looks like, if I can get a Context from the raw window handle, I will be able to create the Surface and there shouldn't really be any more code branching for GL at that point, similar to the Quad example.

[...]

PR #907, opened 2020-08-29
Implement OpenGL Backend For Unix Platforms

zicklag:
**Connections**
Requires: https://github.com/gfx-rs/gfx/pull/3340 ( merged )
Should be merged first: https://github.com/gfx-rs/wgpu/pull/910
Works towards: https://github.com/gfx-rs/wgpu/issues/450

**Description**
Integrates the GFX GL backend for Unix platforms

**Testing**
Runs the basic cube and triangle examples  for wgpu-rs, but not without rendering artifacts on the cube.


zicklag:
OK, I haven't nailed down the dependency versions yet, but the rest of the code should be ready for review.

The most controversial organizational change that I've included in this ( sorry, maybe should have been a separate PR ) was I used the [`cfg_aliases`](https://docs.rs/cfg_aliases) crate to replace the `backends!` macro throughout the code. I think it is much more readable to use plain Rust `cfg` checks instead of a macro in all of the cases that we previously used `backends!` and now you can use the same `cfg` syntax anywhere that the `backends!` macro could not handle such as in struct field definitions.

kvark:
Note that `wgpu` always works with released gfx-rs versions. So we can't merge this until we get something usable in gfx-rs 0.6 branch and publish the GL backend.

wgpu-core/build.rs
```diff
@@ -0,0 +1,15 @@
+fn main() {
+    // Setup cfg aliases
+    cfg_aliases::cfg_aliases! {
```
kvark, line 3: I'm a bit concerned about introducing a build script here. However, I do agree that using `cfg` in the code is much much nicer, so this seems like a good change to do.

wgpu-core/build.rs
```diff
@@ -0,0 +1,15 @@
...
+        // Vendors/systems
+        apple: { any(target_os = "ios", target_os = "macos") },
+        linux: { all(unix, not(apple)) },
```
kvark, line 6: do we need this one? appears to only be used in the `vulkan` define, so we can omit it here

wgpu-core/build.rs
zicklag: Yeah it does add a build script but `cfg_aliases` is just a `macro_rules!` macro with no deps so it shouldn't increase build times by a noticeable amount and it also only resolves to simple `println!` statements that are interpreted by Cargo so it isn't dependent on any host environment or anything like that either.

wgpu-core/build.rs
zicklag: We don't really, I wasn't sure if it would help indicate the intent a little bit more, but if you prefer to omit it that's fine by me.

zicklag:
> I agree the configuration change should land independently prior to GL stuff

OK, I'll extract it out to a separate PR that should be merged before this one.

zicklag:
> Note that wgpu always works with released gfx-rs versions. So we can't merge this until we get something usable in gfx-rs 0.6 branch and publish the GL backend.

Yeah, I figured. I'll put the GFX git commit hashes in there to fix the deps just so we can verify that CI will pass.

zicklag:
OK, the two commits have been separated. Once #910 is merged this PR will only have the second commit in it.

After separating the commits you can see that this was actually *really* simple from the WGPU side thanks to the abstraction provided by GFX, which is really awesome.

zicklag:
Just found that the port to the newer version of GFX broke the Vulkan backend on Linux at least. I'm working on fixing that. It's some panic on the unwrap for [`get_surface_mut`](https://github.com/gfx-rs/wgpu/blob/bba82724a84e22dd23cd341a06be6909a3511370/wgpu-core/src/hub.rs#L746).

wgpu-core/Cargo.toml
```diff
@@ -49,16 +49,26 @@ package = "wgpu-types"
...
+gfx-backend-metal = { git = "https://github.com/gfx-rs/gfx.git", rev = "c9fcd08956eeddac8f6d85b02794fc5e3d1717f5" }
+gfx-backend-vulkan = { git = "https://github.com/gfx-rs/gfx.git", rev = "c9fcd08956eeddac8f6d85b02794fc5e3d1717f5", optional = true }
 
 [target.'cfg(all(unix, not(target_os = "ios"), not(target_os = "macos")))'.dependencies]
-gfx-backend-vulkan = { version = "0.6.1" }
+gfx-backend-vulkan = { git = "https://github.com/gfx-rs/gfx.git", rev = "c9fcd08956eeddac8f6d85b02794fc5e3d1717f5", features = ["x11"]}
+
+[target.'cfg(unix)'.dependencies]
+# TODO: Make x11 feature optional?
+gfx-backend-gl = { git = "https://github.com/gfx-rs/gfx.git", rev = "c9fcd08956eeddac8f6d85b02794fc5e3d1717f5", features = ["x11"] }
```
zicklag, line 60: Should X11 be a Cargo feature for WGPU?

wgpu-core/src/instance.rs
```diff
@@ -342,18 +344,22 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 let map = |inst| {
                     inst
                     .as_ref()
-                    .and_then(|inst| inst.create_surface(handle).ok())
+                    .and_then(|inst| inst.create_surface(handle).map_err(|e| {
+                        eprintln!("Error: {:?}", e);
+                    }).ok())
```
zicklag, line 349: I ran into an `UnsupportedWindowHandle` error when I accidentally left out the X11 feature, but because the code previously just ran `.ok()` I had no way of knowing what was wrong. I'm not sure if this is the right fix or not, but it definitely helps if your surface creation fails. :wink:

This is in a separate commit that can be trivially moved to another PR or just dopped completely, it is up to you.

wgpu-core/Cargo.toml
kvark: No. The idea (so far) is that if the crate that creates a window and passes it to wgpu wants to support x11, it can enable the x11 dependency on gfx-backend-vulkan directly (like wgpu-rs does).

kvark:
gfx-rs GL backend is now published. Time to revive this!

zicklag:
Yay! I'll see about getting this back up to date tomorrow probably. :smiley:

zicklag:
There you go! It looks like surfman is failing to compile on iOS, unfortunately. Not sure why we hadn't hit that until now.

Also the cube example doesn't run:

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/zicklag/.cargo/registry/src/github.com-1ecc6299db9ec823/gfx-backend-gl-0.6.0/src/command.rs:1174:30
stack backtrace:
   0: rust_begin_unwind
             at /rustc/ffa2e7ae8fbf9badc035740db949b9dae271c29f/library/std/src/panicking.rs:483:5
   1: core::panicking::panic_fmt
             at /rustc/ffa2e7ae8fbf9badc035740db949b9dae271c29f/library/core/src/panicking.rs:85:14
   2: core::panicking::panic
             at /rustc/ffa2e7ae8fbf9badc035740db949b9dae271c29f/library/core/src/panicking.rs:50:5
   3: core::option::Option<T>::unwrap
             at /home/zicklag/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:386:21
   4: <gfx_backend_gl::command::CommandBuffer as gfx_hal::command::CommandBuffer<gfx_backend_gl::Backend>>::bind_graphics_descriptor_sets
             at /home/zicklag/.cargo/registry/src/github.com-1ecc6299db9ec823/gfx-backend-gl-0.6.0/src/command.rs:1172:40
   5: wgpu_core::command::render::<impl wgpu_core::hub::Global<G>>::command_encoder_run_render_pass_impl
             at /home/zicklag/git/zicklag/wgpu-workspace/wgpu/wgpu-core/src/command/render.rs:1010:29
   6: wgpu_core::command::render::<impl wgpu_core::hub::Global<G>>::command_encoder_run_render_pass
             at /home/zicklag/git/zicklag/wgpu-workspace/wgpu/wgpu-core/src/command/render.rs:414:9
   7: <wgpu::backend::direct::Context as wgpu::Context>::command_encoder_end_render_pass
             at ./src/backend/direct.rs:1417:9
   8: <wgpu::RenderPass as core::ops::drop::Drop>::drop
             at ./src/lib.rs:2340:13
   9: core::ptr::drop_in_place
             at /home/zicklag/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:175:1
  10: <cube::Example as cube::framework::Example>::render
             at ./examples/cube/main.rs:412:9
  11: cube::framework::start::{{closure}}
             at ./examples/cube/../framework.rs:297:17
  12: winit::platform_impl::platform::sticky_exit_callback
             at /home/zicklag/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.22.2/src/platform_impl/linux/mod.rs:698:5
  13: winit::platform_impl::platform::x11::EventLoop<T>::run_return
             at /home/zicklag/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.22.2/src/platform_impl/linux/x11/mod.rs:312:21
  14: winit::platform_impl::platform::x11::EventLoop<T>::run
             at /home/zicklag/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.22.2/src/platform_impl/linux/x11/mod.rs:390:9
  15: winit::platform_impl::platform::EventLoop<T>::run
             at /home/zicklag/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.22.2/src/platform_impl/linux/mod.rs:645:35
  16: winit::event_loop::EventLoop<T>::run
             at /home/zicklag/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.22.2/src/event_loop.rs:149:9
  17: cube::framework::start
             at ./examples/cube/../framework.rs:230:5
  18: cube::framework::run
             at ./examples/cube/../framework.rs:307:5
  19: cube::main
             at ./examples/cube/main.rs:419:5
  20: core::ops::function::FnOnce::call_once
             at /home/zicklag/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
[0.222763 ERROR]()(no module): Heaps not completely freed before drop. Utilization: MemoryHeapUtilization { utilization: MemoryUtilization { used: 917504, effective: 263424 }, size: 18446744073709551615 }
[0.222821 ERROR]()(no module): Heaps not completely freed before drop. Utilization: MemoryHeapUtilization { utilization: MemoryUtilization { used: 16777216, effective: 262996 }, size: 18446744073709551615 }
[0.222881 ERROR]()(no module): Not all allocations from LinearAllocator were freed
[0.222901 ERROR]()(no module): Memory leak: SizeEntry(768) is still used
[0.222924 ERROR]()(no module): Memory leak: SizeEntry(6144) is still used
[0.222942 ERROR]()(no module): Memory leak: SizeEntry(49152) is still used
[0.222959 ERROR]()(no module): Memory leak: SizeEntry(256) is still used
[0.222992 ERROR]()(no module): Memory leak: SizeEntry(262144) is still used
[0.223013 ERROR]()(no module): Memory leak: SizeEntry(65536) is still used
[0.223039 ERROR]()(no module): DescriptorAllocator is dropped
```

Is this potentially related to your recent image changes @kvark ? The triangle example works, though.

Otherwise the changes should pretty much be in order I think.

wgpu-core/src/instance.rs
kvark: let's make this `tracing::warn!` instead of `eprintln`.

wgpu-core/src/lib.rs
```diff
@@ -212,6 +214,8 @@ macro_rules! gfx_select {
             wgt::Backend::Dx12 => $global.$method::<$crate::backend::Dx12>( $($param),* ),
             #[cfg(windows)]
             wgt::Backend::Dx11 => $global.$method::<$crate::backend::Dx11>( $($param),* ),
+            #[cfg(unix)]
```
kvark, line 217: why isn't this code using the cfg's established by the build.rs?

kvark (commented):
Thank you for the amazing work!
A few small things here

kvark:
Oh, for iOS - we shouldn't be building OpenGL for iOS at all. So if it falls under `unix`, we should change build.rs to make it so it's `unix` but not `iOS`.

wgpu-core/src/lib.rs
zicklag: Oh, I'm trying to remember for sure. I knew I should have added a comment when I ran into this. :man_facepalming:  I think there may have been an issue because it was in a macro. Like the CFG alias didn't work because it was evaluated before the alias existed somehow.

I may be wrong, though because I can't figure out the real reason that would happen looking back. I'll try to use the aliases and see if it still breaks or if I was confused earlier.

zicklag:
> Oh, for iOS - we shouldn't be building OpenGL for iOS at all. So if it falls under unix, we should change build.rs to make it so it's unix but not iOS.

Oh, yeah, of course. Easy fix.

wgpu-core/src/lib.rs
zicklag: Yeah, I tried it and it didn't work with the aliases. I'm not actually sure exactly why, but unfortunately in this macro at least we have to use the expanded cfg forms. I added a comment stating such.

:thinking: It's still a little confusing because I've used the cfg aliases in the other `macro_rules` macros I'm pretty sure.

zicklag:
OK, now I'm a little confused why surfman is being pulled in at all. The only thing that has a dependency on `surfman` should be `gfx-backend-gl` and is already excluding the `surfman` dep on iOS:

```toml
[target.'cfg(all(unix, not(target_os = "ios")))'.dependencies]
surfman = { version = "0.3", features = ["sm-raw-window-handle"], optional = true }
```

Could clippy be being weird about this and including the dependency when it isn't supposed to?

kvark:
@zicklag you are looking at the `master` branch for `gfx` but using the 0.6 released version, which is a bit different:
```
[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
surfman = { version = "0.3", features = ["sm-raw-window-handle"] }
```
Really, we shouldn't even try to build the GL backend on iOS, so this line needs to be changed:
```
[target.'cfg(unix)'.dependencies]
gfx-backend-gl = { version = "0.6" }
```

zicklag:
Oh, yes, that's right. Thank you. I don't know how I missed that. :laughing: Fixed it ( assuming CI passes, now ).

kvark (approved):

kvark:
Thank you for all the amazing work!

zicklag:
Glad I could help!

Unfortunately it looks like we've got to wait for Canonical to fix their servers or something. The apt repos are failing on CI everywhere at the moment.

kvark:
bors retry

Issue #807, opened 2020-07-17
[meta] Feature Testing Checklist


kvark:
Now that we got #786, we should start taking advantage of it by introducing local "playtests".
We can take `wgpu-rs` [example matrix](https://github.com/gfx-rs/wgpu-rs/blob/d0825a219511c28bf74f8cbef9056b55fde4e04e/examples/README.md#feature-matrix) as a baseline:

- [ ] vertex attributes
- [ ] instancing
- [ ] lines and points
- [ ] sampled color textures
- [ ] storage textures
- [ ] binding array
- [ ] comparison samplers
- [ ] subresource views
- [ ] cubemaps
- [ ] multisampling
- [ ] off-screen rendering
- [ ] stencil testing
- [ ] depth testing
- [ ] depth biasing
- [ ] read-only depth
- [ ] blending
- [ ] render bundles
- [ ] compute passes
- optional extensions
- [ ] - binding indexing
- [ ] - push constants
- [ ] WGSL shaders

Note-1: we don't have to be super detailed in testing all the features in the world. It's just one of the testing layers in "defence in depth" onion of layers, above gfx-rs running on Warden and Vulkan CTS, and below wgpu-rs and WebGPU CTS. Common sense should be applied to weight the maintenance overhead versus the test coverage. I think getting a test for an entry in the feature matrix is a great start.

Note-2: it's not clear what shaders we should be using here. I suggest we start with a single shader module that outputs a full-screen quad (without vertex attributes) in VS, and writes down a solid fixed color to the output in FS. Perhaps, we can write test shaders in WGSL right away as this thing is getting more and more mature?

DevOrc:
I'd be interested in working on this, but I was curious about a few things. 

1) Should these tests be in the player/tests directory or in a separate dedicated crate?
2) For the shaders, is WGSL ready enough to be used? I have never used it but would be wiling to learn it

If you had any other insights that would be appreciated

kvark:
1. `player` is already a dedicated development-only crate, so it's a good place to put the tests in. More specifically, into `player/tests`
2. I don't think WGSL is ready just yet. We can go a long way with just a single SPIR-V shader that is compiled from source off-line and just fills out the screen with some color, for now.

> If you had any other insights that would be appreciated

We'd love to chat on "#wgpu:matrix.org"!

cwfitzgerald:
[...]

PR #909, opened 2020-08-29
Add Quad play test

DevOrc:
**Connections**
#807 

**Description**
Adds a play test that renders a white quad to a texture. The texture is then copied to a buffer and checked for accuracy with a file of expected bytes (All 0xFF). 

**Testing**
Cargo test succeeds


player/tests/data/quad.ron
```diff
@@ -0,0 +1,169 @@
...
+            usage: (
+                bits: 27,
+            ),
+        )),
+        CreateTextureView(
+            id: Id(0, 1, Empty),
+            parent_id: Id(0, 1, Empty),
+            desc: (
+                label: Some("Output Texture View"),
+                format: Some(Rgba8Unorm),
```
kvark, line 40: Since these are the same as in the original, we can skip this and use `None` for both the format and `dimension`.

player/tests/data/quad.ron
```diff
@@ -0,0 +1,169 @@
...
+                    ),
+                ],
+                target_depth_stencil: None,
+            ),
+            CopyTextureToBuffer(
+                src: (
+                    texture: Id(0, 1, Empty),
+                    mip_level: 0,
+                    array_layer: 0,
+                    origin: (
```
kvark, line 145: I'm thinking that we should sprinkle `serde(default)` on all those fields with default values, so that writing tests is more egonomic

player/tests/data/quad.ron
```diff
@@ -0,0 +1,169 @@
...
+            sample_mask: 4294967295,
+            alpha_to_coverage_enabled: false,
+        )),
+        Submit(1, [
+            RunRenderPass(
+                base: (
+                    commands: [
+                        SetPipeline(Id(0, 1, Empty)),
+                        Draw(
+                            vertex_count: 6,
```
kvark, line 111: can we instead draw a single big triangle? i.e. have the vertex count of 3

player/tests/data/quad.vert
```diff
@@ -0,0 +1,18 @@
...
+    vec2(-1, -1),
+    vec2(-1, 1),
+    vec2(1, 1),
+    vec2(-1, -1),
+    vec2(1, 1),
+    vec2(1, -1)
+);
+
+void main() {
+    gl_Position = vec4(positions[gl_VertexIndex], 0.0, 1.0);
```
kvark, line 17: for a single large triangle, we'd do something like `gl_Position = vec4(gl_VertexIndex == 2 ? 2.0 : -1.0, gl_VertexIndex == 1 ? 2.0 : -1.0, 0.0, 1.0);

player/tests/data/quad.vert
```diff
@@ -0,0 +1,18 @@
+#version 450
+
+out gl_PerVertex {
+    vec4 gl_Position;
+};
+
+const vec2 positions[6] = vec2[6](
```
kvark, line 7: There is an unfortunate bug on SPIRV-Cross version that we are currently using, related to constant arrays like this. So it would be better to just generate the positions in the code.

player/tests/data/quad.ron
DevOrc: So should the serde defaults be enabled for tracing, or do you want to leave them just for replays?

kvark: let's leave them for replays only, so that we don't overburden the compile times when only the tracing feature is on

DevOrc:
I made those fixes and added serde(default) to the TextureCopyView and the TextureViewDescriptor. IDK if you want more defaults but those two seemed like a reasonable place to start.

PR #910, opened 2020-08-30
Replace backends! Macro with CFG Aliases

zicklag:
**Connections**
Needs to be merged before: https://github.com/gfx-rs/wgpu/pull/907

**Description**
This change makes it easier to conditionally compile code based on graphics backends by adding `#[cfg]` aliases for the backends such as `vulkan`, `metal`, etc. This makes the code easier to read and maintain.

**Testing**
Tested the WGPU-rs cube and boids examples and they work as normal on a Linux machine with Vulkan.

### DIFF
--- a/wgpu-core/src/macros.rs
+++ b/wgpu-core/src/macros.rs
@@ -2,140 +2,56 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-macro_rules! backends {
-    // one let statement per backend
-    (
-        #[vulkan] let $vk_field:pat = $vk_expr:expr;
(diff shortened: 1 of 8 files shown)


wgpu-core/src/macros.rs
```diff
@@ -2,140 +2,56 @@
...
+                $map
+            }
+        )*
     };
 
     // a struct constructor with one field per backend with mapped data
     (
         let map = |$backend:pat| $map:block;
-        $Struct:path {
+        $Struct:ident {
```
zicklag, line 41: For reasons I cannot yet fathom, the macro will not work in the `cargo test` anymore unless this is switched to `ident`.  I think this will still work in most cases and the tests pass after this change. I'm not sure if this is a problem or not.

zicklag:
It seems that windows nightly is failing, but I don't know that that is because of my changes. It's failing to compile another crate.

PR #912, opened 2020-08-30
Fix write-only stencil state descriptors from not working - fixes #911

### LINKED ISSUE #911: Stencil testing broken between v0.5 and v0.6
**Description**
We make extensive use of stencil testing over at ruffle, as Flash games often used one shape to mask another. After updating to v0.6 of wgpu-rs, any game with masking seems to be broken. I've managed to port the relevant code back over on top of the wgpu-rs `hello-triangle` example code, to use stencil testing to turn two overlapping triangles into a diamond. It works in v0.5 but when ported to v0.6, it only shows the clear colour.

~For some reason, I can't open the v0.6 version in RenderDoc to see how the output compares, and trying the trace feature doesn't appear to do anything.~
I've managed to get RenderDoc to work when running in dx12, and I can see that the up triangle (which should set the stencil state) isn't setting the correct stencil state. Pictures of that at the end of the report.

**Repro steps**
I've written one commit for 0.5 and 0.6 which should be effectively the same, and it mimics a stripped down version of how we use stencils at Ruffle. It (should) display a diamond, by first drawing a triangle going down - but only writing `1` to the stencil buffer and colour output disabled - and then one going up which requires the stencil bit `1` to be set.

It should look like this: 

Commit for 0.5: https://github.com/Dinnerbone/wgpu-rs/commit/69a16aede8ea9c623cf88ab160f9a1a5994c3131
Commit for 0.6: https://github.com/Dinnerbone/wgpu-rs/commit/43fcfc830b1b312f61733afe009cb465e4db2a07

Stencil testing should work. In the example above, you should see a diamond - currently you see nothing, as the test fails so no pixels are allowed to show.

**Extra materials**
Unfortunately ~I'm not able to get the 0.6 example to run in RenderDoc, and~ running with `mkdir -p trace && WGPU_TRACE=trace cargo run --features trace --example hello_triangle` is giving me nothing.
I'm attempting to run a debug now.

I have ran a bisect on wgpu-rs repo and found that it was introduced in http://github.com/gfx-rs/wgpu-rs/pull/503.

In 0.5 (working), the stencil state of the first draw looks like this: 

In 0.6 (broken), the stencil state of the first draw looks like this: 

**Platform**
I've only confirmed the issue on Windows 10 x64 so far.
It affects (at least) Vulkan, DX12 and DX11.

Dinnerbone:
**Connections**
This fixes [#911 - Stencil testing broken between v0.5 and v0.6](https://github.com/gfx-rs/wgpu/issues/911).

**Description**
Write-only stencil states (read 0, write >0) are being treated as if they are disabled, which causes pipelines to act as though they don't have any stencil state set at all. This worked prior to commit 2473c259718e8b8803b86d4480e5f0f54d4048e4 (introduced in PR #873). As far as I can tell, this works fine in Vulkan, Metal, DX12 and DX11 as we've been using this approach over at Ruffle for a while now.

**Testing**
You can view the reproduction case in #911 for manual testing. I have confirmed that this fix makes that case work as expected.

I couldn't find any automated tests for wgpu-types to copy and add for this case. If that's wanted then please let me know what the best approach is.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -855,7 +855,7 @@ impl StencilStateDescriptor {
     pub fn is_enabled(&self) -> bool {
         (self.front != StencilStateFaceDescriptor::IGNORE
             || self.back != StencilStateFaceDescriptor::IGNORE)
-            && self.read_mask != 0
+            && (self.read_mask != 0 || self.write_mask != 0)
     }
     pub fn is_read_only(&self) -> bool {
         self.write_mask == 0


wgpu-types/src/lib.rs
```diff
@@ -855,7 +855,7 @@ impl StencilStateDescriptor {
     pub fn is_enabled(&self) -> bool {
         (self.front != StencilStateFaceDescriptor::IGNORE
             || self.back != StencilStateFaceDescriptor::IGNORE)
-            && self.read_mask != 0
+            && (self.read_mask != 0 || self.write_mask != 0)
```
grovesNL, line 858: @kvark this is the check I was also wondering about in https://github.com/gfx-rs/wgpu/pull/873#discussion_r469667951

kvark: yeah, sorry, looks like I screwed this up 😓

grovesNL: No problem :) I just wanted to link it here in case it was intentional

kvark:
@Dinnerbone could you make the same PR against the 0.6 branch, and add a small commit that bumps wgpu-types patch version and adds a small line to the changelog?

kvark:
I've done the backporting myself, and this fix is now available in `wgpu-types-0.6.1`

PR #916, opened 2020-09-02
Flush staging init buffers

kvark:
**Connections**
Related to https://github.com/gfx-rs/gfx/pull/3362
With the new use of `write_buffer`/`write_texture`, the results on D3D11 regressed.

**Description**
It just happened to be the case that all the backends returned coherent staging buffers, used internally for `write_*` and `mapped_at_creation == true`. However, D3D11 backend currently doesn't have the coherent memory at all, and it helped me to see that we were missing a flush() on the staging buffer.

**Testing**
Untested, but should work (tm).


cwfitzgerald (approved):
LGTM
bors r+

PR #918, opened 2020-09-04
Pass [u8] instead of [u32] for push constants

PENGUINLIONG:
**Connections**
This is a follow-up of https://github.com/gfx-rs/wgpu-rs/issues/544.

**Description**
Becuase `&[u8]` is broadly recognized as a type for buffers, it makes more sense to type push constant buffers as `&[u8]` rather than `&[u32]`. Detailed rationale can be found in [this issue](https://github.com/gfx-rs/wgpu-rs/issues/544).

**Testing**
This is trivially tested with `cargo test`.

### DIFF
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -1041,12 +1041,12 @@ pub mod bundle_ffi {
     ) {
         span!(_guard, DEBUG, "RenderBundle::set_push_constants");
         assert_eq!(
-            offset & 3,
+            offset & (wgt::PUSH_CONSTANT_ALIGNMENT - 1),
             0,
             "Push constant offset must be aligned to 4 bytes."
         );
         assert_eq!(
-            size_bytes & 3,
+            size_bytes & (wgt::PUSH_CONSTANT_ALIGNMENT - 1),
             0,
             "Push constant size must be aligned to 4 bytes."
         );
@@ -1057,7 +1057,7 @@ pub mod bundle_ffi {
 
         pass.base.push_constant_data.extend(
             data_slice
-                .chunks_exact(4)
+                .chunks_exact(wgt::PUSH_CONSTANT_ALIGNMENT as usize)
                 .map(|arr| u32::from_ne_bytes([arr[0], arr[1], arr[2], arr[3]])),
         );
 
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -232,7 +232,8 @@ impl RenderBundle {
                     let pipeline_layout = &pipeline_layout_guard[pipeline_layout_id];
 
                     if let Some(values_offset) = values_offset {
-                        let values_end_offset = (values_offset + size_bytes / 4) as usize;
+                        let values_end_offset =
+                            (values_offset + size_bytes / wgt::PUSH_CONSTANT_ALIGNMENT) as usize;
                         let data_slice = &self.base.push_constant_data
                             [(values_offset as usize)..values_end_offset];
 
@@ -1036,14 +1037,30 @@ pub mod bundle_ffi {
         stages: wgt::ShaderStage,
         offset: u32,
         size_bytes: u32,
-        data: *const u32,
+        data: *const u8,
     ) {
         span!(_guard, DEBUG, "RenderBundle::set_push_constants");
-        let data_slice = slice::from_raw_parts(data, (size_bytes / 4) as usize);
+        assert_eq!(
+            offset & 3,
+            0,
+            "Push constant offset must be aligned to 4 bytes."
+        );
+        assert_eq!(
+            size_bytes & 3,
+            0,
+            "Push constant size must be aligned to 4 bytes."
+        );
+        let data_slice = slice::from_raw_parts(data, size_bytes as usize);
         let value_offset = pass.base.push_constant_data.len().try_into().expect(
             "Ran out of push constant space. Don't set 4gb of push constants per RenderBundle.",
         );
-        pass.base.push_constant_data.extend_from_slice(data_slice);
+
+        pass.base.push_constant_data.extend(
+            data_slice
+                .chunks_exact(4)
+                .map(|arr| u32::from_ne_bytes([arr[0], arr[1], arr[2], arr[3]])),
+        );
+
         pass.base.commands.push(RenderCommand::SetPushConstant {
             stages,
             offset,
--- a/wgpu-core/src/command/compute.rs
+++ b/wgpu-core/src/command/compute.rs
@@ -356,7 +356,8 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                     values_offset,
                 } => {
                     let end_offset_bytes = offset + size_bytes;
-                    let values_end_offset = (values_offset + size_bytes / 4) as usize;
+                    let values_end_offset =
+                        (values_offset + size_bytes / wgt::PUSH_CONSTANT_ALIGNMENT) as usize;
                     let data_slice =
                         &base.push_constant_data[(values_offset as usize)..values_end_offset];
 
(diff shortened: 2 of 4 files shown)


PENGUINLIONG:
It failed some clippy tests because I cast `*const u8` to `*const u32`. Sounds like I have to retype the push constant pointers in `gfx-hal`... That's a lot of penetration. Should I step ahead?

wgpu-core/src/command/bundle.rs
```diff
@@ -1036,14 +1040,20 @@ pub mod bundle_ffi {
...
     ) {
         span!(_guard, DEBUG, "RenderBundle::set_push_constants");
-        let data_slice = slice::from_raw_parts(data, (size_bytes / 4) as usize);
+        let data_slice = slice::from_raw_parts(data, size_bytes as usize);
         let value_offset = pass.base.push_constant_data.len().try_into().expect(
             "Ran out of push constant space. Don't set 4gb of push constants per RenderBundle.",
         );
+
+        let aligned_len = ((data_slice.len() + 3) / 4) * 4;
+        pass.base.push_constant_data.reserve(aligned_len);
```
cwfitzgerald, line 1052: Instead of changing internal data to use u8s, we should convert to u32s in this step. As we are copying anyway, we can use a loop of (something like):

```rust
base.push_constant_data.extend(data_slice.chunks_exact(4).map(|arr| {
    u32::from_ne_bytes([
        arr[0],
        arr[1],
        arr[2],
        arr[3],
    ])
}));
```

Then assert somewhere (probably wgpu-rs) that size is a multiple of four. There are ways of doing this that allows it to be not a multiple of 4, but all types in glsl are of size 4 anyway, so I don't think that's something we need to support.

wgpu-core/src/command/compute.rs
```diff
@@ -481,14 +487,20 @@ pub mod compute_ffi {
         pass: &mut ComputePass,
         offset: u32,
         size_bytes: u32,
-        data: *const u32,
+        data: *const u8,
     ) {
-        span!(_guard, DEBUG, "RenderPass::set_push_constant");
-        let data_slice = slice::from_raw_parts(data, (size_bytes / 4) as usize);
+        span!(_guard, DEBUG, "ComputePass::set_push_constant");
```
cwfitzgerald, line 492: Good catch.

wgpu-core/src/command/compute.rs
```diff
@@ -481,14 +487,20 @@ pub mod compute_ffi {
...
+        let data_slice = slice::from_raw_parts(data, size_bytes as usize);
         let value_offset = pass.base.push_constant_data.len().try_into().expect(
             "Ran out of push constant space. Don't set 4gb of push constants per ComputePass.",
         );
+
+        let aligned_len = ((data_slice.len() + 3) / 4) * 4;
```
cwfitzgerald, line 498: All the same comments apply here.

kvark:
> It failed some clippy tests because I cast *const u8 to *const u32. Sounds like I have to retype the push constant pointers in gfx-hal... That's a lot of penetration. Should I step ahead?

I don't think this is required. Casting pointers to `*const u32` should be safe as long as the alignment is enforced. So if your code enforces the alignment, and clippy still complains, than we need to silence clippy for these casting spots.

cwfitzgerald (changes requested):
Good call on realizing that that was too deep, we want the conversion to happen here. I have added notes on what should be done to properly convert.

Thanks for spearheading this!

wgpu-core/src/command/bundle.rs
```diff
@@ -232,10 +232,14 @@ impl RenderBundle {
                     let pipeline_layout = &pipeline_layout_guard[pipeline_layout_id];
 
                     if let Some(values_offset) = values_offset {
-                        let values_end_offset = (values_offset + size_bytes / 4) as usize;
+                        let values_end_offset = (values_offset + size_bytes) as usize;
                         let data_slice = &self.base.push_constant_data
                             [(values_offset as usize)..values_end_offset];
 
+                        let data_slice = std::slice::from_raw_parts(
```
kvark, line 239: we need to assert that `values_offset % 4 == 0`

cwfitzgerald:
Also a rustfmt pass is needed.

wgpu-core/src/command/bundle.rs
```diff
@@ -1036,14 +1040,20 @@ pub mod bundle_ffi {
...
+        data: *const u8,
     ) {
         span!(_guard, DEBUG, "RenderBundle::set_push_constants");
-        let data_slice = slice::from_raw_parts(data, (size_bytes / 4) as usize);
+        let data_slice = slice::from_raw_parts(data, size_bytes as usize);
         let value_offset = pass.base.push_constant_data.len().try_into().expect(
             "Ran out of push constant space. Don't set 4gb of push constants per RenderBundle.",
         );
+
+        let aligned_len = ((data_slice.len() + 3) / 4) * 4;
```
kvark, line 1051: not sure if the codegen will be different, but it may be better to write as `(((data_slice.len() - 1) | 3) + 1)`

wgpu-core/src/command/bundle.rs
kvark, line 1052: instead of doing this, we could just do `pass.base.push_constant_data.resize(aligned_len, 0);`

PENGUINLIONG:
Oh, now I see what you talked about. There is an alignment requirement for the entire allocation, not just the size. I remember it's not enforced in Vulkan so I was a bit lost in your discussion before.

wgpu-core/src/command/bundle.rs
PENGUINLIONG: I'm pondering whether I should pad unaligned data, or I just assert it to panic; and it should be noted that [16 bit types](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_KHR_16bit_storage.html) or even [8 bit types](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_KHR_8bit_storage.html) are allowed in GLSL and have been used quite extensively on mobile platforms. So I think it is something worth dealing with in this crate.

PENGUINLIONG: Well I thought I don't have to zero the memory that way.

PENGUINLIONG:
I looked up a couple of sources and I found no requirement on the base address of host side memory to set push constants, so I think it's not necessary to enforce 4-byte alignment for addresses on the host side, while the size alignment is already sufficient. I'm gonna leaving this to further discussion. See:
- [ID3D12GraphicsCommandList::SetGraphicsRoot32BitConstants](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12graphicscommandlist-setgraphicsroot32bitconstants)
- [vkCmdPushConstants](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCmdPushConstants.html)

kvark:
We can't leave these unaligned, since gfx-rs API expects `&[u32]` right now, and you can't get that form unaligned pointers.

cwfitzgerald:
Additionally, due to the design of our internals we are already paying for a memcpy of push constants to the internal buffer, doing a conversion to u32 should be basically free on top of that.

PENGUINLIONG:
Oh, i see. I just realized there are some extra costs for unaligned RW. Will fix it soon.

wgpu-core/src/command/bundle.rs
```diff
@@ -1036,14 +1036,30 @@ pub mod bundle_ffi {
         stages: wgt::ShaderStage,
         offset: u32,
         size_bytes: u32,
-        data: *const u32,
+        data: *const u8,
     ) {
         span!(_guard, DEBUG, "RenderBundle::set_push_constants");
-        let data_slice = slice::from_raw_parts(data, (size_bytes / 4) as usize);
+        assert_eq!(
+            offset & 3,
```
kvark, line 1043: let's use the `PUSH_CONSTANT_ALIGNMENT` constant

wgpu-core/src/command/bundle.rs
```diff
@@ -1036,14 +1036,30 @@ pub mod bundle_ffi {
...
+        );
+        let data_slice = slice::from_raw_parts(data, size_bytes as usize);
         let value_offset = pass.base.push_constant_data.len().try_into().expect(
             "Ran out of push constant space. Don't set 4gb of push constants per RenderBundle.",
         );
-        pass.base.push_constant_data.extend_from_slice(data_slice);
+
+        pass.base.push_constant_data.extend(
+            data_slice
+                .chunks_exact(4)
```
kvark, line 1059: why not just `chunk`? we already checked the size to be aligned to 4

kvark (commented):
A few more touches are needed!
Also, please squash it :)

wgpu-core/src/command/bundle.rs
cwfitzgerald: This was my doing in the code I suggested, iirc the code generated for chunk_exact was better than chunk as it can statically guarantee the accesses won't panic.

cwfitzgerald: See https://godbolt.org/z/Wf619s

cwfitzgerald:
Any update on this?

PENGUINLIONG:
It should be good to go, I think. Did I miss something?

kvark:
The last review noted that `PUSH_CONSTANT_ALIGNMENT` are missing in a few places

PR #919, opened 2020-09-04
[0.6] Don't bind groups that aren't expected by the layout

kvark:
**Connections**
Fixes https://github.com/gfx-rs/wgpu-rs/issues/550

**Description**
Our `is_valid` was used in 2 different contexts with slightly different meaning. One of the was wrong.
Now it returns an option, and each context can reinterpret it as it needs.

**Testing**
Tested on that repro case in the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## wgpu-core-0.6.3 (2020-09-04)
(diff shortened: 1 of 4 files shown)


cwfitzgerald (approved):
Gorgeous! Ship it!

kvark:
bors r=cwfitzgerald

PR #920, opened 2020-09-05
Don't bind groups that aren't expected by the layout 

kvark:
**Connections**
Sibling of #919 for master.

**Description**

**Testing**
In addition to the fix itself, this comes with a new regression playtest!


wgpu-core/src/command/bind.rs
```diff
@@ -231,7 +232,7 @@ impl Binder {
     fn compatible_count(&self) -> usize {
         self.entries
             .iter()
-            .position(|entry| !entry.is_valid())
+            .position(|entry| !entry.is_valid().unwrap_or(false))
```
grovesNL, line 235: nit: `Option<bool>` is a little bit confusing to read, maybe we could special case the layout check here (e.g. `entry.expects_layout() && !entry.is_valid()`) so the original function could keep returning `bool`

grovesNL (approved):

wgpu-core/src/command/bind.rs
kvark: My problem is, I don't know what to return in this case where the PR code has `None`.
Perhaps the most idiomatic solution would be to return an `enum` of 3 variants. I'm not sure if it's worth it given the verbosity increase, but I'm fine if anybody wants to follow-up with this.

kvark:
bors r=grovesNL

wgpu-core/src/command/bind.rs
grovesNL: Oops, I meant to write `!entry.expects_layout() || !entry.is_valid()`, so it would be short-circuited if `expects_layout` is `false`. In that case the return value of `is_valid` could continue returning `true` for `None`.

Yeah I thought about an enum too, but agree it might be overkill :)

PR #690, opened 2020-06-01
Implement Anisotropic Filtering Extension and Expose Extension Interface

cwfitzgerald:
**Connections**

Resolves #568.

**Description**

This PR solves two tightly related problems:
 - Adds the ability to create a sampler with anisotropic filtering
 - Adds the webgpu interface for getting the limits/extensions on adapters and devies.

**Testing**

I have tested this against my codebase which uses it. Adding an example in wgpu-rs on how to use extensions/limits would be good, especially as we have native/webgpu extensions, but can come as a future enhancement, once we feel the extension situation has stabilized a bit.

**TODO**

- [x] Allow wgpu-rs to call limit/extension interface (https://github.com/gfx-rs/wgpu-rs/pull/338)
- [ ] Determine how wgpu-native is going to adopt this functionality.
- [ ] After discussing #691, what changes need to be made.

PR #921, opened 2020-09-05
add `polygon_mode: PolyonMode` to `RasterizationStateDescriptor` to allow drawing wireframes

manuel-woelker:
**Description**
This adds `polygon_mode: PolyonMode` to `RasterizationStateDescriptor` to allow drawing in wireframes and point mode

I added `hal::Features::NON_FILL_POLYGON_MODE` to the `wishful_features` in `instance.rs`. I don't know what side effects this might have.

**Testing**
Tested locally on my Windows machine with Vulkan backend.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -298,6 +298,16 @@ bitflags::bitflags! {
         ///
         /// This is a web and native feature.
         const ADDRESS_MODE_CLAMP_TO_BORDER = 0x0000_0000_0100_0000;
+        /// Allows the user to set a non-fill polygon mode in [`RasterizationStateDescriptor::polygon_mode`]
+        ///
+        /// This allows drawing polygons/triangles as lines (wireframe) or points instead of filled
+        ///
+        /// Supported platforms:
+        /// - DX12
+        /// - Vulkan
+        ///
+        /// This is a native only feature.
+        const NON_FILL_POLYGON_MODE = 0x0000_0000_0200_0000;
         /// Features which are part of the upstream WebGPU standard.
         const ALL_WEBGPU = 0x0000_0000_0000_FFFF;
         /// Features that are only available when targeting native (not web).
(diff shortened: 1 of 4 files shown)


kvark:
Thank you for the PR!

This capability is not a part of upstream API: https://gpuweb.github.io/gpuweb/
See the investigation in https://github.com/gpuweb/gpuweb/issues/137

It means that we can only expose it as a native-only extension.

manuel-woelker:
Would this mean integrating it with the extension mechanism implemented in #690?

Is there anything else I should be aware of?

kvark:
You'd add a bit to https://github.com/gfx-rs/wgpu/blob/730313177e3a45c9ba53b2016d4332755c656b44/wgpu-types/src/lib.rs#L148 within `ALL_NATIVE` range.

kvark:
And you'd only expose it if gfx-hal exposes `NON_FILL_POLYGON_MODE` for this adapter

manuel-woelker:
Thanks for the info, I'll give it a shot!

kvark (approved):
Nice!
Could you squash it please?

manuel-woelker:
I added the feature flag, trying to mirror the other flags.

I pushed a branch of wgpu-rs that modifies the cube example to test line fill mode here:
https://github.com/manuel-woelker/wgpu-rs/blob/fb-polygon-mode/examples/cube/main.rs

Let me know if I missed anything!

I see the build is failing due to the new LabeledContextError, I'll rebase and squash.

wgpu-types/src/lib.rs
```diff
@@ -611,6 +641,7 @@ impl Default for CullMode {
 pub struct RasterizationStateDescriptor {
     pub front_face: FrontFace,
     pub cull_mode: CullMode,
+    pub polygon_mode: PolygonMode,
```
cwfitzgerald, line 644: This member should have a comment on it saying it can only be non-fill with the feature on.

manuel-woelker: Good point, I added the docs.

kvark (approved):
Thank you!
bors r+

manuel-woelker:
Thanks for the great and "realtime" feedback!

kvark:
Please join us for even more real time feedback on "#wgpu:matrix.org"!

PR #922, opened 2020-09-06
Add LabeledContextError

scoopr:
Purpose for this is to add more context to an error.
create_render_pipeline error is improved, by showing the action
(creating render pipeline) and the user provided label, to aid with
the investigation.

Example error
before:
```
wgpu error: Validation error

Caused by:
    error in stage VERTEX: unable to find an entry point matching the Vertex execution model

thread 'main' panicked at 'Handling wgpu errors as fatal by default', src/backend/direct.rs:1457:5
```

after:
```
wgpu error: Validation error

Caused by:
    Creating render pipeline (label: Some("hello-triangle pipeline"))
    error in stage VERTEX
    unable to find an entry point matching the Vertex execution model

thread 'main' panicked at 'Handling wgpu errors as fatal by default', src/backend/direct.rs:1457:5
```

### DIFF
--- a/wgpu-core/src/pipeline.rs
+++ b/wgpu-core/src/pipeline.rs
@@ -183,11 +183,25 @@ pub enum CreateRenderPipelineError {
     },
     #[error("missing required device features {0:?}")]
     MissingFeature(wgt::Features),
-    #[error("error in stage {flag:?}: {error}")]
+    #[error("error in stage {flag:?}")]
     Stage {
         flag: wgt::ShaderStage,
+        #[source]
         error: StageError,
     },
+}
+
+impl CreateRenderPipelineError {
+    pub(crate) fn with_label<'a>(
+        self,
+        label: &Option<Cow<'a, str>>,
+    ) -> crate::LabeledContextError<CreateRenderPipelineError> {
+        crate::LabeledContextError {
+            source: self,
+            description: "Creating render pipeline",
+            label: label.as_ref().map(|inner| inner.to_string()),
+        }
+    }
 }
 
 bitflags::bitflags! {
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -2595,7 +2595,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         implicit_pipeline_ids: Option<ImplicitPipelineIds<G>>,
     ) -> Result<
         (id::RenderPipelineId, pipeline::ImplicitBindGroupCount),
-        pipeline::CreateRenderPipelineError,
+        crate::LabeledContextError<pipeline::CreateRenderPipelineError>,
     > {
         span!(_guard, INFO, "Device::create_render_pipeline");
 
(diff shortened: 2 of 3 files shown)


wgpu-core/src/pipeline.rs
```diff
@@ -183,13 +183,27 @@ pub enum CreateRenderPipelineError {
...
 
+impl CreateRenderPipelineError {
+    pub(crate) fn context<'a>(
+        self,
+        label: &Option<Cow<'a, str>>,
+    ) -> crate::LabeledContextError<CreateRenderPipelineError> {
+        crate::LabeledContextError {
+            source: self,
+            description: "Creating render pipeline",
+            label: label.as_ref().map(|inner| inner.to_owned().into_owned()),
```
kvark, line 202: probably could be `inner.to_string()`

kvark (commented):
Looks like a nice improvement!
What is the longer term idea here? Do you want to convert all the errors to the "labeled" ones?

scoopr:
Didn't think through all the error sites yet, but at least resource creation when there is a label readily available, I think it would be rude not to show it.

It would nice to show the label on incorrect use later on as well, but I guess the label is not retained anywhere accessible? Perhaps we could retain the labels, maybe under a feature flag?

Otherwise adding more context to errors would be nice, but more case-by-case basis. 
Wouldn't it be wild to have the errors propagate all the way from naga? ;)

wgpu-core/src/pipeline.rs
```diff
@@ -183,13 +183,27 @@ pub enum CreateRenderPipelineError {
...
+    #[error("error in stage {flag:?}")]
     Stage {
         flag: wgt::ShaderStage,
+        #[source]
         error: StageError,
     },
 }
 
+impl CreateRenderPipelineError {
+    pub(crate) fn context<'a>(
```
kvark, line 195: Can we iterate a bit on this name? Generally, no-verb methods should be used for pure accessor functions.
Maybe this one should be `with_label`?

kvark:
> Perhaps we could retain the labels, maybe under a feature flag?

We totally could retain them, at least in debug.

> Wouldn't it be wild to have the errors propagate all the way from naga? ;)

It could be fine. In WebGPU, the errors are supposed to be rather free-form, so the more information we provide - the better.

wgpu-core/src/pipeline.rs
scoopr: I'm fine with anything, I originally called it `named` but that felt a bit disconnected what it was doing. Then I thought I was adding more context to the error, akin to the `context` in anyhow.

scoopr:
Addressed the nit and renamed to `with_label`

scoopr:
Some of the CI errors seemed to be just network errors

scoopr:
Were there still some concerns about this? or would you prefer that I pile up more of these in the same pull request, so you can see better how the change feels?

kvark:
I asked on the chat and forgot to follow up here, sorry!
bors r+

kvark:
As discussed in #927, we'd want to revert this PR and then add support for the labeled context to wgpu-rs instead of wgpu

Issue #894, opened 2020-08-24
create_texture with too many mip levels triggers validation error


benfrankel:
**Description**
Calling `Device::create_texture` with a `TextureDescriptor` that has `mip_level_count > 32 - width.min(height).min(depth).leading_zeros()` (aka number of times you can do `width >>= 1; height >>= 1; depth >>= 1;` before they all reach `0`) triggers a validation error (`VUID-VkImageCreateInfo-mipLevels-00958`).

**Repro steps**
1. Patch `wgpu-rs`'s mipmap example in `examples/mipmap/main.rs` with `let size = 1 << (mip_level_count - 1) - 1;` so the generated texture is too small for `9` mipmap levels.
2. Run the example (`cargo run --example mipmap`).

**Expected vs observed behavior**
- **Expected:** Error or panic.
- **Observed:** Validation error.

**Platform**
- Linux 5.7.12-arch1-1
- X11
- wgpu v0.6.0

benfrankel:
Tried passing `mip_level_count` of `0`, got `VUID-VkImageCreateInfo-mipLevels-00947` and a segfault.

kvark:
If we are computing the mipmap level count incorrectly, it should be easy to fix! Would you want to try making a patch?

benfrankel:
I took a closer look at #893, specifically this line:

```rust
if mip_level_count == 0 && mip_level_count > kind.compute_num_levels() as u32 {
```

The `&&` should be an `||`, but otherwise, that should actually fix this issue completely (mipmap level count seems to be computed correctly). I'll make a patch to fix the `&&` and improve the `InvalidMipLevelCount` error message.

PR #923, opened 2020-09-08
Fix invalid mip level count check

benfrankel:
**Connections**
Fixes https://github.com/gfx-rs/wgpu/issues/894
Continues from https://github.com/gfx-rs/wgpu/pull/893

**Description**
Invalid mip level counts are allowed through `Device::create_texture`, triggering Vulkan validation errors. A fix was applied in https://github.com/gfx-rs/wgpu/pull/893, but the fix needs a minor correction to work properly.

Also added back the `MAX_MIP_LEVELS` check, because it's technically possible for `MAX_MIP_LEVELS` to be less than `kind.compute_num_levels()`, e.g. if one of the given dimensions is very big (2^16). 

Error message is updated to reflect the change in behavior: there are now three ways for a mip level count to be invalid: it's equal to zero; it's greater than `MAX_MIP_LEVELS`; or it's greater than the largest mip level count allowed for the given size.

**Testing**
Tested by manually modifying the `mipmap` example in `wgpu-rs` to pass various edge case values for `size` and `mip_level_count`.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -515,7 +515,10 @@ impl<B: GfxBackend> Device<B> {
         let usage = conv::map_texture_usage(desc.usage, aspects);
 
         let mip_level_count = desc.mip_level_count;
-        if mip_level_count == 0 && mip_level_count > kind.compute_num_levels() as u32 {
+        if mip_level_count == 0
+            || mip_level_count > MAX_MIP_LEVELS
+            || mip_level_count > kind.compute_num_levels() as u32
+        {
             return Err(resource::CreateTextureError::InvalidMipLevelCount(
                 mip_level_count,
             ));
--- a/wgpu-core/src/resource.rs
+++ b/wgpu-core/src/resource.rs
@@ -239,7 +239,7 @@ pub enum CreateTextureError {
     CannotCopyD24Plus,
     #[error(transparent)]
     InvalidDimension(#[from] TextureDimensionError),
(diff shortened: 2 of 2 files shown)


benfrankel:
While testing this, I ran into a couple Vulkan validation errors regarding allowed image size: `VUID-VkImageCreateInfo-usage-00964` when width exceeds `maxFramebufferWidth` and `VUID-VkImageCreateInfo-usage-00965` when height exceeds `maxFramebufferHeight`.

I also ran into an additional validation error regarding allowed mip level count: `VUID-VkImageCreateInfo-mipLevels-02255` when mip level count exceeds `maxMipLevels` for a given image format, e.g. `maxMipLevels=15` for `VK_FORMAT_R8G8B8A8_SRGB`.

wgpu-core/src/device/mod.rs
```diff
@@ -515,7 +515,10 @@ impl<B: GfxBackend> Device<B> {
         let usage = conv::map_texture_usage(desc.usage, aspects);
 
         let mip_level_count = desc.mip_level_count;
-        if mip_level_count == 0 && mip_level_count > kind.compute_num_levels() as u32 {
+        if mip_level_count == 0
+            || mip_level_count > MAX_MIP_LEVELS
```
kvark, line 519: This should be handled differently in the end: we need the max texture size in the limits, and we'd be validating that texture size here, and also making sure that `MAX_MIP_LEVELS` is sane at the adapter creation. This can be followed up with, not a blocker

kvark (approved):
thank you!
bors r+

PR #782, opened 2020-07-11
Add check while trying to remove uninserted Ids

kunalmohan:
**Connections**
_Link to the issues addressed by this PR, or dependent PRs in other repositories_
An attempt to fix #781 
regression from #776 

**Description**
_Describe what problem this is solving, and how it's solved._
When we used `VecMap`, it simply returned `None` for even out of bounds access to the map (We depended on it returning `None`). After #776 , we get a panic. So adding a simple index check before accessing it fixes this issue.

**Testing**
_Explain how this change is tested._
Tested on wgpu-rs examples with the changes in https://github.com/gfx-rs/wgpu-rs/pull/430. All examples run fine except the `cube` which segfaults.

PR #925, opened 2020-09-11
Allow dropping of error resources

kunalmohan:
**Connections**

**Description**
wgpu currently panics when trying to drop an error resource. `unregister_locked()` now returns `Option<T>` to handle this.

**Testing**
Tested with wgpu-rs examples and CTS in Servo.

### DIFF
--- a/wgpu-core/src/device/life.rs
+++ b/wgpu-core/src/device/life.rs
@@ -354,9 +354,8 @@ impl<B: GfxBackend> LifetimeTracker<B> {
                 if trackers.bundles.remove_abandoned(id) {
                     #[cfg(feature = "trace")]
                     trace.map(|t| t.lock().add(trace::Action::DestroyRenderBundle(id.0)));
-                    let res = hub.render_bundles.unregister_locked(id.0, &mut *guard);
 
-                    if let Some(res) = res {
+                    if let Some(res) = hub.render_bundles.unregister_locked(id.0, &mut *guard) {
                         self.suspected_resources.add_trackers(&res.used);
                     }
                 }
@@ -371,9 +370,8 @@ impl<B: GfxBackend> LifetimeTracker<B> {
                 if trackers.bind_groups.remove_abandoned(id) {
                     #[cfg(feature = "trace")]
                     trace.map(|t| t.lock().add(trace::Action::DestroyBindGroup(id.0)));
-                    let res = hub.bind_groups.unregister_locked(id.0, &mut *guard);
 
-                    if let Some(res) = res {
+                    if let Some(res) = hub.bind_groups.unregister_locked(id.0, &mut *guard) {
                         self.suspected_resources.add_trackers(&res.used);
 
                         let submit_index = res.life_guard.submission_index.load(Ordering::Acquire);
@@ -396,9 +394,8 @@ impl<B: GfxBackend> LifetimeTracker<B> {
                 if trackers.views.remove_abandoned(id) {
                     #[cfg(feature = "trace")]
                     trace.map(|t| t.lock().add(trace::Action::DestroyTextureView(id.0)));
-                    let res = hub.texture_views.unregister_locked(id.0, &mut *guard);
 
-                    if let Some(res) = res {
+                    if let Some(res) = hub.texture_views.unregister_locked(id.0, &mut *guard) {
                         let raw = match res.inner {
                             resource::TextureViewInner::Native { raw, source_id } => {
                                 self.suspected_resources.textures.push(source_id.value);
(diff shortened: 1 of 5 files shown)


wgpu-core/src/device/life.rs
```diff
@@ -355,7 +355,10 @@ impl<B: GfxBackend> LifetimeTracker<B> {
                     #[cfg(feature = "trace")]
                     trace.map(|t| t.lock().add(trace::Action::DestroyRenderBundle(id.0)));
                     let res = hub.render_bundles.unregister_locked(id.0, &mut *guard);
-                    self.suspected_resources.add_trackers(&res.used);
+
+                    if let Some(res) = res {
```
kvark, line 359: nit: can we just combine this as `if let Some(res) = hub.render_bundles.unregister_locked(..) { ... }` here and in other cases?

wgpu-core/src/hub.rs
```diff
@@ -451,10 +451,11 @@ impl<T, I: TypedId + Copy, F: IdentityHandlerFactory<I>> Registry<T, I, F> {
...
     }
 
-    pub fn unregister_locked(&self, id: I, guard: &mut Storage<T, I>) -> T {
-        let value = guard.remove(id).unwrap();
+    pub fn unregister_locked(&self, id: I, guard: &mut Storage<T, I>) -> Option<T> {
+        let value = guard.remove(id);
         //Note: careful about the order here!
         self.identity.free(id);
+        //Returning None is legal if it's an error ID
         value
```
kvark, line 459: I believe we are missing a few things:
  1. the same kind of modification of `unregister()`
  2. the logic in `fn remove()` that would replace an `Error` entry with `Vacant`

Speaking of which, I think `remove` should panic if the element is `Vacant`, or the index is out of range. Currently it returns an `Option`, which is ambiguous (was it a bad id? or was it an error element? these are different cases).

kvark (changes requested):
Thank you!
This is a great change. We need to fix a little bit more, and we'll be good to go :)

wgpu-core/src/hub.rs
kunalmohan: For out of range access, I purposely returned `None` in https://github.com/gfx-rs/wgpu/pull/782 to match with `VecMap` implementation.
I will try refactoring the code that makes these out of range calls.

kunalmohan: https://github.com/gfx-rs/wgpu/blob/f6f7210b6323e643ff0000f3a4b25b2bb911e848/wgpu-core/src/device/mod.rs#L3457

This is code that makes out of range calls to `remove()`

kvark: uuhhh

kvark: it seems to be quite an edge case. Maybe we can roll out a different `try_remove` method for this?

wgpu-core/src/device/mod.rs
```diff
@@ -2528,10 +2529,11 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         let (cmdbuf, _) = hub
             .command_buffers
             .unregister(command_encoder_id, &mut token);
-
-        let device = &mut device_guard[cmdbuf.device_id.value];
-        device.untrack::<G>(&hub, &cmdbuf.trackers, &mut token);
-        device.cmd_allocator.discard(cmdbuf);
+        if let Some(cmdbuf) = cmdbuf {
```
kunalmohan, line 2532: I could not combine the `if let` statement with the function call here (and some more places) because of the mutable borrow of `token` in `unregister()` call.

wgpu-core/src/swap_chain.rs
```diff
@@ -248,40 +248,44 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
-        };
-        let queue = &mut device.queue_group.queues[0];
-        let result = unsafe { queue.present(B::get_surface_mut(surface), image, sem) };
-
-        tracing::debug!(trace = true, "Presented. End of Frame");
-
-        for fbo in sc.acquired_framebuffers.drain(..) {
-            unsafe {
-                device.raw.destroy_framebuffer(fbo);
+        if let Some(view) = view {
```
kvark, line 251: can we just early out here if it's None? I.e. something like
let (view_maybe, _) = hub.texture_views.unregister(view_id.value.0, &mut token);
let view = view_maybe.ok_or(SwapChainError::Invalid)?;
// continue as usual

wgpu-core/src/device/mod.rs
kvark: not sure what you wanted to do. Could you clarify?

kunalmohan: Similar to what you suggested in the previous review, i.e.
```rust
if let (Some(cmdbuf), _) = hub
            .command_buffers
            .unregister(command_encoder_id, &mut token) { ... }
```

kvark: ah I see. Thanks for clarifying!

kvark (approved):

PR #926, opened 2020-09-11
Update naga with new image API

kvark:
**Connections**
Gets us https://github.com/gfx-rs/naga/pull/187, https://github.com/gfx-rs/naga/pull/185, https://github.com/gfx-rs/naga/pull/182, partially https://github.com/gfx-rs/naga/pull/169, and https://github.com/gfx-rs/naga/pull/167

**Description**
Brings us to a point where all wgpu-rs shaders are successfully validated 🎉 

**Testing**
wgpu-rs examples

### DIFF
--- a/wgpu-core/src/validation.rs
+++ b/wgpu-core/src/validation.rs
@@ -79,8 +79,8 @@ pub enum BindingError {
     WrongSamplerComparison,
     #[error("derived bind group layout type is not consistent between stages")]
     InconsistentlyDerivedType,
-    #[error("texture format {0:?} isn't recognized")]
-    UnknownStorageFormat(wgt::TextureFormat),
+    #[error("texture format {0:?} is not supported for storage use")]
+    BadStorageFormat(wgt::TextureFormat),
 }
 
 #[derive(Clone, Debug, Error)]
@@ -96,8 +96,8 @@ pub enum InputError {
 pub enum StageError {
     #[error("shader module is invalid")]
     InvalidModule,
-    #[error("unable to find an entry point matching the {0:?} execution model")]
-    MissingEntryPoint(naga::ShaderStage),
+    #[error("unable to find an entry point at {0:?} stage")]
+    MissingEntryPoint(wgt::ShaderStage),
     #[error("error matching global binding at index {binding} in set {set} against the pipeline layout: {error}")]
     Binding {
         set: u32,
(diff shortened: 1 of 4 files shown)


wgpu-core/src/validation.rs
```diff
@@ -170,6 +170,70 @@ fn get_aligned_type_size(
...
+        Sf::R32Float => Tf::R32Float,
+        Sf::Rg16Uint => Tf::Rg16Uint,
+        Sf::Rg16Sint => Tf::Rg16Sint,
+        Sf::Rg16Float => Tf::Rg16Float,
+        Sf::Rgba8Unorm => Tf::Rgba8Unorm,
+        Sf::Rgba8Snorm => Tf::Rgba8Snorm,
+        Sf::Rgba8Uint => Tf::Rgba8Uint,
+        Sf::Rgba8Sint => Tf::Rgba8Sint,
+        Sf::Rgb10a2Unorm => Tf::Rgb10a2Unorm,
+        Sf::Rg11b10Float => Tf::Rg11b10Float,
```
cwfitzgerald, line 224: A couple of these formats appear to be missing from the inverse transform. Is this intentional?

cwfitzgerald (approved):
A possible concern, but otherwise looks good!

wgpu-core/src/validation.rs
kvark: This is a little bit tricky. So right now, the set of formats supported for storage is smaller than what Naga has.
We could reduce Naga's set, but then it would stop being useful for non-WebGPU-related things, e.g. for gfx-rs, where more formats can be storage.

`map_storage_format_to_naga` only has this limited subset.
In `map_storage_format_from_naga` I decided to map all the formats, so that it's 100% succeeds. Then, if see that an inappropriate format is used, I can generate a `BadStorageFormat` error. If I had the `map_storage_format_from_naga` returning an `Option`/`Result`, I'd have to figure out a different error variant for it...

kvark:
thank you for reviewing!
bors r=cwfitzgerald

kvark:
bors retry

wgpu-core/src/validation.rs
cwfitzgerald: Ah alright; seems reasonable.

kvark:
bors retry

kvark:
bors retry

cwfitzgerald:
bors retry

kvark:
bors retry

PR #927, opened 2020-09-12
More wrapping with LabelContextError

scoopr:
Continuing on wrapping more context to some error messages, these wrap the creation error for Buffer, Texture and ComputePipeline.

I'm not entirely happy that I had to undo some try-operator `From` short-circuiting with more explicit error construction (the more verbose `map_err` stuff).

Otherwise this is fairly straightforward

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -1261,17 +1261,19 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         device_id: id::DeviceId,
         desc: &resource::TextureDescriptor,
         id_in: Input<G, id::TextureId>,
-    ) -> Result<id::TextureId, resource::CreateTextureError> {
+    ) -> Result<id::TextureId, crate::LabeledContextError<resource::CreateTextureError>> {
         span!(_guard, INFO, "Device::create_texture");
 
         let hub = B::hub(self);
         let mut token = Token::root();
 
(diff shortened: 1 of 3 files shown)


scoopr:
I suppose in lines of modified code, this would of been neater:
```rust
fn create_foo_impl(desc:FooDesc) -> Result<Foo, CreateFooError> { /* .. */ }
fn create_foo(desc:FooDesc) -> Result<Foo, LabelContextError<Foo>> 
{
   create_foo_impl(desc).map_err( |err| err.with_label(&desc.label) )
}
```
but seems a bit hacky maybe.

kvark:
Does this have to be done in `wgpu` itself? Since:
  - wgpu-rs passes the descriptors in, so it knows the label
  - it is the thing that unwraps today
  - it has a single point of contact for every function (i.e the "xxx_impl" comes naturally)

Wouldn't this be nicer to move to `wgpu-rs` completely?

scoopr:
Well.. yes that would be cleaner in many fronts. Then all users of wgpu-core would need to remember to do that to have a consistent error messages. Also tiny fear that starts to be more substance in wgpu-rs than just wrapping/language stuff.

But if you are OK with that, I can try the other path too

kvark:
Yes, I think that's ok. The other clients of wgpu-core would be able to choose what exactly they need to include in the error message. The current selection of "label only" is a bit arbitrary and opinionated anyway.

PR #929, opened 2020-09-14
Add swap_chain_error

scoopr:


scoopr:
I basically copy-pasted another method, I hope it is correct (not even sure how to test)

kvark (approved):
bors r+

kvark:
Sorry, I almost missed an important thing.
Swapchain IDs are not generated, at all. Instead, we get a Surface ID, which is backend-less, and set the backend part in it to the one coming from the device. That means a surface can't have more than one swapchain for a given backend, and that's ok.
We need to figure out a solution for erroring swapchains...

scoopr:
Oh, OK.

That means, places that use the swapchain id just assume that it can extract the surface from it, ignoring that it may be invalid?

Then it is above my threshold of effort I’m willing to give it right now, and I’ll just close this one.

kvark:
yes, sorta. In places where the backend is known, there is a 1:1 relationship between the surface and the swapchain.
We may need to redesign this aspect if it causes a problem with the errors...

PR #931, opened 2020-09-15
Retain labels for resources

scoopr:
This is wip, mostly a conversation starter.

I've attempted to have some labels show up in the errors, outside of only creation errors.
Steps to do that include,
1. Retain the label in the underlying struct (here, `Buffer`). It is guarded by `debug_assertions`
2. Add the label option to the error variant, and show it
3. Add label option to the `Element::Error` variant, so that invalid ids retain the label also. Didn't guard it with `debug_assertions` yet, but probably should. `buffer_error` now takes the label option.
4. Added a query for invalid id labels
5. Add `Global::buffer_label` for returning the label from the struct, or the invalid id.
6. Change the error variants to query the `buffer_label` when creating the error

with that (and little bit of fudging to get in to the error state), I can get
```
wgpu error: Validation error

Caused by:
    In CommandEncoder::copy_buffer_to_buffer
    buffer (0, 1, Metal) (label: Some("Staging buffer")) is invalid
```
instead of 
```
wgpu error: Validation error

Caused by:
    In CommandEncoder::copy_buffer_to_buffer
    buffer (0, 1, Metal) is invalid
```

Some of the hub handling etc. is a bit iffy for me, I really have no idea what I'm doing.

The point would be to massage this to the point that there would be a simple steps to follow on how to decorate all the errors with labels to whatever they refer to.

Other ideas I had, included trying to have `impl Debug for Id<BufferId>` to print out the label (from the struct, wouldn't work for invalid ids), but in that context I don't think I have access to the Global, so I'm not sure how resolve the id to anything useful. I suppose there could a whole separate singleton storage for id to label maps which could be easier to access in any context. Then the errors could just store the Id, and not have a separate label field.

### DIFF
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -316,6 +316,23 @@ pub struct BindGroupLayout<B: hal::Backend> {
     pub(crate) desc_counts: DescriptorCounts,
     pub(crate) dynamic_count: usize,
     pub(crate) count_validator: BindingTypeMaxCountValidator,
+    #[cfg(debug_assertions)]
+    pub(crate) label: String,
+}
+
+impl<B: hal::Backend> crate::hub::Resource for BindGroupLayout<B> {
+    const TYPE: &'static str = "BindGroupLayout";
+
+    fn life_guard(&self) -> &LifeGuard {
+        unreachable!()
+    }
+
+    fn label(&self) -> &str {
+        #[cfg(debug_assertions)]
+        return &self.label;
+        #[cfg(not(debug_assertions))]
+        return "";
+    }
 }
 
 #[derive(Clone, Debug, Error)]
@@ -488,6 +505,14 @@ impl<B: hal::Backend> PipelineLayout<B> {
     }
 }
 
+impl<B: hal::Backend> crate::hub::Resource for PipelineLayout<B> {
+    const TYPE: &'static str = "PipelineLayout";
+
+    fn life_guard(&self) -> &LifeGuard {
+        &self.life_guard
+    }
+}
+
 #[repr(C)]
 #[derive(Clone, Debug, Hash, PartialEq)]
 #[cfg_attr(feature = "trace", derive(Serialize))]
@@ -582,6 +607,14 @@ impl<B: hal::Backend> Borrow<RefCount> for BindGroup<B> {
 impl<B: hal::Backend> Borrow<()> for BindGroup<B> {
     fn borrow(&self) -> &() {
         &DUMMY_SELECTOR
+    }
+}
+
+impl<B: hal::Backend> crate::hub::Resource for BindGroup<B> {
+    const TYPE: &'static str = "BindGroup";
+
+    fn life_guard(&self) -> &LifeGuard {
+        &self.life_guard
     }
 }
 
--- a/wgpu-core/src/command/allocator.rs
+++ b/wgpu-core/src/command/allocator.rs
@@ -8,6 +8,9 @@ use crate::{
     PrivateFeatures, Stored, SubmissionIndex,
 };
 
+#[cfg(debug_assertions)]
+use crate::LabelHelpers;
+
 use hal::{command::CommandBuffer as _, device::Device as _, pool::CommandPool as _};
 use parking_lot::Mutex;
 use thiserror::Error;
@@ -80,9 +83,11 @@ impl<B: GfxBackend> CommandAllocator<B> {
         device: &B::Device,
         limits: wgt::Limits,
         private_features: PrivateFeatures,
+        label: &crate::Label,
         #[cfg(feature = "trace")] enable_tracing: bool,
     ) -> Result<CommandBuffer<B>, CommandAllocatorError> {
         //debug_assert_eq!(device_id.backend(), B::VARIANT);
+        let _ = label; // silence warning on release
         let thread_id = thread::current().id();
         let mut inner = self.inner.lock();
 
@@ -126,6 +131,8 @@ impl<B: GfxBackend> CommandAllocator<B> {
             } else {
                 None
             },
+            #[cfg(debug_assertions)]
+            label: label.to_string_or_default(),
         })
     }
 }
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -12,8 +12,8 @@ use crate::{
     id, pipeline, resource, span, swap_chain,
     track::{BufferState, TextureSelector, TextureState, TrackerSet},
     validation::{self, check_buffer_usage, check_texture_usage},
-    FastHashMap, Label, LifeGuard, MultiRefCount, PrivateFeatures, Stored, SubmissionIndex,
-    MAX_BIND_GROUPS,
+    FastHashMap, Label, LabelHelpers, LifeGuard, MultiRefCount, PrivateFeatures, Stored,
+    SubmissionIndex, MAX_BIND_GROUPS,
 };
 
 use arrayvec::ArrayVec;
@@ -273,7 +273,7 @@ impl<B: GfxBackend> Device<B> {
             mem_allocator: Mutex::new(mem_allocator),
             desc_allocator: Mutex::new(descriptors),
             queue_group,
-            life_guard: LifeGuard::new(),
+            life_guard: LifeGuard::new("<device>"),
             active_submission_index: 0,
             trackers: Mutex::new(TrackerSet::new(B::VARIANT)),
             render_passes: Mutex::new(FastHashMap::default()),
@@ -489,7 +489,7 @@ impl<B: GfxBackend> Device<B> {
             full_range: (),
             sync_mapped_writes: None,
             map_state: resource::BufferMapState::Idle,
-            life_guard: LifeGuard::new(),
+            life_guard: LifeGuard::new(desc.label.borrow_or_default()),
         })
     }
 
@@ -581,7 +581,7 @@ impl<B: GfxBackend> Device<B> {
                 levels: 0..desc.mip_level_count as hal::image::Level,
                 layers: 0..kind.num_layers(),
             },
-            life_guard: LifeGuard::new(),
+            life_guard: LifeGuard::new(desc.label.borrow_or_default()),
         })
     }
 
@@ -721,6 +721,8 @@ impl<B: GfxBackend> Device<B> {
                 .count(),
             count_validator,
             entries: entry_map,
+            #[cfg(debug_assertions)]
+            label: label.unwrap_or("").to_string(),
         })
     }
 
@@ -822,7 +824,7 @@ impl<B: GfxBackend> Device<B> {
                 value: id::Valid(self_id),
                 ref_count: self.life_guard.add_ref(),
             },
-            life_guard: LifeGuard::new(),
+            life_guard: LifeGuard::new(desc.label.borrow_or_default()),
             bind_group_layout_ids: desc
                 .bind_group_layouts
                 .iter()
@@ -901,6 +903,14 @@ impl<B: hal::Backend> Device<B> {
                 self.raw.destroy_framebuffer(fbo);
             }
         }
+    }
+}
+
+impl<B: hal::Backend> crate::hub::Resource for Device<B> {
+    const TYPE: &'static str = "Device";
+
+    fn life_guard(&self) -> &LifeGuard {
+        &self.life_guard
     }
 }
 
@@ -1156,10 +1166,18 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         Ok(())
     }
 
-    pub fn buffer_error<B: GfxBackend>(&self, id_in: Input<G, id::BufferId>) -> id::BufferId {
+    pub fn buffer_label<B: GfxBackend>(&self, id: id::BufferId) -> String {
+        B::hub(self).buffers.label_for_resource(id)
+    }
+
+    pub fn buffer_error<B: GfxBackend>(
+        &self,
+        id_in: Input<G, id::BufferId>,
+        label: Option<&str>,
+    ) -> id::BufferId {
         B::hub(self)
             .buffers
-            .register_error(id_in, &mut Token::root())
+            .register_error(id_in, label.unwrap_or(""), &mut Token::root())
     }
 
     pub fn buffer_destroy<B: GfxBackend>(
@@ -1304,10 +1322,18 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         Ok(id.0)
     }
 
-    pub fn texture_error<B: GfxBackend>(&self, id_in: Input<G, id::TextureId>) -> id::TextureId {
+    pub fn texture_label<B: GfxBackend>(&self, id: id::TextureId) -> String {
+        B::hub(self).textures.label_for_resource(id)
+    }
+
+    pub fn texture_error<B: GfxBackend>(
+        &self,
+        id_in: Input<G, id::TextureId>,
+        label: Option<&str>,
+    ) -> id::TextureId {
         B::hub(self)
             .textures
-            .register_error(id_in, &mut Token::root())
+            .register_error(id_in, label.unwrap_or(""), &mut Token::root())
     }
 
     pub fn texture_destroy<B: GfxBackend>(
@@ -1567,7 +1593,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             extent: texture.kind.extent().at_level(desc.base_mip_level as _),
             samples: texture.kind.num_samples(),
             selector,
-            life_guard: LifeGuard::new(),
+            life_guard: LifeGuard::new(desc.label.borrow_or_default()),
         };
         let ref_count = view.life_guard.add_ref();
 
@@ -1593,10 +1619,11 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
     pub fn texture_view_error<B: GfxBackend>(
         &self,
         id_in: Input<G, id::TextureViewId>,
+        label: Option<&str>,
     ) -> id::TextureViewId {
         B::hub(self)
             .texture_views
-            .register_error(id_in, &mut Token::root())
+            .register_error(id_in, label.unwrap_or(""), &mut Token::root())
     }
 
     pub fn texture_view_drop<B: GfxBackend>(
@@ -1726,7 +1753,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 value: id::Valid(device_id),
                 ref_count: device.life_guard.add_ref(),
             },
-            life_guard: LifeGuard::new(),
+            life_guard: LifeGuard::new(desc.label.borrow_or_default()),
             comparison: info.comparison.is_some(),
         };
         let ref_count = sampler.life_guard.add_ref();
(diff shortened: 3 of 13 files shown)


kvark:
Thank you for the PR!

What was the original motivation for this? What problem are you solving? Just trying to get better error messages?

What is the end goal that you see in this direction? Every error on any object carrying around the label strings?

scoopr:
Yes, in the rust community it is pretty easy to get spoiled by really good error messages, I think. And having an error message that pinpoints the whole context is, I think, such a breath of fresh air after years of looking at `GL_INVALID_OPERATION`.

Showing the label of all the involved resources would put it really far in that direction. I haven't researched what different errors would look like before/after, but if wgpu knows the context of an error, it would be really nice to show it. Especially in debug builds.

I would rather have the errors have the ids of the resources than the label itself, there would be less cloning of the label, and most of the work would just be in Error Display time. And it would give the possibility of showing even more context of the resource, if useful.

kvark:
Thank you for explanation!

> I would rather have the errors have the ids of the resources than the label itself

This appears to not be possible, since we need the global context in order to grab the labels.

What I was thinking, originally, is that: if you see an error that is confusing, you could record an API trace. Then you'd be able to see what descriptors (including the label, but also other useful things) associated with any IDs mentioned in the error messages.

> Showing the label of all the involved resources would put it really far in that direction

With that other workflow in mind (using an API trace), my concern was whether labels are going to be enough. If we'd then decide to go all the way and basically provide the full descriptor (passed to resource creation), then it would seem inferior to just making the users to record API traces. If, on the other hand, labels is all we need, then I think it's a good improvement, assuming it doesn't add too much code complexity and maintenance load.

TL;DR: I think the feature is useful. We should probably just do a `String` instead of `Option<String>`, leaving it empty if there is no label.

scoopr:
I removed the Option.

I also tried to use the `cfg(debug_assertions)` more carefully, though I'm not totally sure about how I feel about it. Retaining the labels in release seem certainly superfluous for wgpu-rs applications. For browsers it would certainly be compiled with release, so missing them there would be kind of odd – and then the error messages claiming an empty string as labels suddenly become more confusing than helpful. 

Oh, and the `_error` methods are going to have an api change as they accept the label, but this is still implementing it only for `buffer_error`.

I'm a bit annoyed that I'm again making the error generation more verbose, and adding those labels to all kinds of error variants all over is going to be a bit messy and annoying work.
I kinda wish I could augment the errors with `invalid_buffer_error.add_note("Source buffer label {}",label)` or some such. Might read nicely in the error message if they would be formatted like the compiler notes. Then the errors themselves wouldn't carry the label. So yeah, I'm a bit concerned about the maintenance burden as well.

> What I was thinking, originally, is that: if you see an error that is confusing, you could record an API trace. 

For complex cases, you may be right, that reaching for other tools might be better. Focusing on making the simple cases crystal clear is certainly enough. There is value to have the simpler error explained in-situ, without trying to reproduce with tools.

kvark (commented):
Thinking about this some more, I believe this logic should (once again!) land in `wgpu-rs`. It can resolve all the errors and replace the IDs (which are totally useless for wgpu-rs consumer) to the labels.

kvark:
Notably, the public methods like `buffer_label` that this PR has would be required for these wgpu-rs changes. Also, it may be useful to make the error types generic over the indices, so that we could replace them by labels in wgpu-rs.

kvark:
Heads up - this PR is still very much needed, just focused on bringing the logic on wgpu-rs side.

kvark:
Here is a more concrete proposal. First of all, the goal we are pursuing here is to have more descriptive messages. So in a sense you don't actually *need* the proper error structs with `String` in them. You just need a resulting error message string.

I think we can do this by just being smart in the error handler. For example, transfer operations in wgpu-rs could use the following code to produce an error string:
```rust
pub trait PrettyError {
    fn fmt_pretty(&self, f: &mut fmt::Formatter<'_>, context: &Context) -> Result<(), fmt::Error>;
}

impl PrettyError for wgc::command::TransferError {
    fn fmt_pretty(&self, f: &mut fmt::Formatter<'_>, context: &Context) -> Result<(), fmt::Error> {
        use wgc::command::TransferError as Te;
        match *self {
            Te::InvalidBuffer(id) => {
                let name = context.0.buffer_label(id);
                f.debug_tuple("InvalidBuffer").field(&name).finish()
            }
            Te::InvalidTexture(id) => {
                let name = context.0.texture_label(id);
                f.debug_tuple("InvalidTexture").field(&name).finish()
            }
            _ => self.fmt(f), // relies on the classy `Debug` implementation
        }
    }
}
```

As you can see, it skips all the boilerplate for variants that don't contain IDs. It takes care of the ID variants only, and that boilerplate turns out to be pretty light. We'll need that trait implemented for all the errors that we are currently unwrapping with `PrettyResult`. The `PrettyResult` implementation would need to be adjusted to use this new `fmt_pretty`.

And of course, we'd need those `buffer_label`, `texture_label`, etc, in `wgpu-core` in order to query this information back.

scoopr:
Ahh, but I need to traverse the errors.. by going through `err.source()`, which I would then have to `downcast_ref` to the `PrettyError`, but you can't downcast to a trait, but only to the concrete type? .. which would mean in wgpu-rs there would be this massive downcast attempt to any known type that implements `PrettyError`?

With matching concrete type, I did manage to make it work for a single instance!
```
wgpu error: Validation Error

Caused by:
    In CommandEncoder::copy_buffer_to_buffer
    Copy error
    destination buffer/texture is missing the `COPY_DST` usage flag
      destination label=staging_buffer
```

For that I added the buffer id to `TransferError::MissingCopyDstUsageFlag`, and matched that in the `fmt_pretty`. Also I couldn't easily use the `std::fmt::Formatter` machinery, so the `fmt_pretty` returns a `String` currently.

And this of course means that the `wgpu::Error` needs to carry the formatted `String` around, and I also some of the indenting is now applied all over the place .

scoopr:
The latest commits makes this pull just about retaining the labels and related things. The formatting would then happen in in wgpu-rs side, with gfx-rs/wgpu-rs#617

wgpu-core/src/device/mod.rs
```diff
@@ -1156,10 +1160,37 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
-            .register_error(id_in, &mut Token::root())
+    pub fn buffer_label<B: GfxBackend>(&self, buffer_id: id::BufferId) -> String {
+        let hub = B::hub(self);
+        let mut token = Token::root();
+        let (buffer_guard, _) = hub.buffers.read(&mut token);
+
+        #[allow(unused_variables)]
+        match buffer_guard.get(buffer_id) {
+            Ok(buffer) => {
+                let buffer = buffer; // silence warning
```
kvark, line 1171: aren't we already allowing unused variables?

wgpu-core/src/device/mod.rs
```diff
@@ -1156,10 +1160,37 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+                #[cfg(debug_assertions)]
+                return buffer.label.clone();
```
kvark, line 1173: we shouldn't need `return`

wgpu-core/src/hub.rs
```diff
@@ -151,6 +154,16 @@ impl<T, I: TypedId> Storage<T, I> {
         result
     }
 
+    pub(crate) fn label_for_invalid_id(&self, id: I) -> &str {
+        let (index, _, _) = id.unzip();
+        match &self.map[index as usize] {
```
kvark, line 159: let's not match on a reference, please

wgpu-core/src/hub.rs
```diff
@@ -80,6 +80,9 @@ impl IdentityManager {
 enum Element<T> {
     Vacant,
     Occupied(T, Epoch),
+    #[cfg(debug_assertions)]
+    Error(Epoch, String),
```
kvark, line 84: I think we can carry `String` unconditionally here. If it's empty, it doesn't cost much, and the `Occupied` already takes more size, so adding `String` to `Error` variant wouldn't change the total size of the `enum`.

wgpu-core/src/hub.rs
```diff
@@ -256,6 +278,7 @@ impl<B: hal::Backend> Access<ComputePipeline<B>> for BindGroup<B> {}
 impl<B: hal::Backend> Access<RenderPipeline<B>> for Device<B> {}
 impl<B: hal::Backend> Access<RenderPipeline<B>> for BindGroup<B> {}
 impl<B: hal::Backend> Access<RenderPipeline<B>> for ComputePipeline<B> {}
+impl<B: hal::Backend> Access<RenderPipeline<B>> for Root {}
```
kvark, line 281: let's not modify this list please. Until we get a stronger mean of protection, we should strive to keep this minimal.
For your purposes, all of these things (`RenderPipeline`, `Buffer`, `Texture`) are accessible in the `Device` context. So you can lock the device guard for reading first, and then get any of these.

scoopr:
Alright, I organized some stuff differently, in attempt to DRY it a bit. I added the `label_for_resource` to `Registry`, but that meant it needed to have a trait to ask the label from the type, so that what `HasLabel` is.

It kind of goes around the Token system, I'm not entirely sure how important that is here, but it would of had to make some custom token with all access anyway, as I understood it.

scoopr:
Implemented retaining of the labels for the rest of the types.
I can squash them if needed

wgpu-core/src/binding_model.rs
```diff
@@ -406,6 +408,8 @@ pub struct PipelineLayout<B: hal::Backend> {
     pub(crate) life_guard: LifeGuard,
     pub(crate) bind_group_layout_ids: ArrayVec<[Valid<BindGroupLayoutId>; MAX_BIND_GROUPS]>,
     pub(crate) push_constant_ranges: ArrayVec<[wgt::PushConstantRange; SHADER_STAGE_COUNT]>,
+    #[cfg(debug_assertions)]
+    pub(crate) label: String,
```
kvark, line 412: I'm wondering... if this is on all objects, and `life_guard` is in all objects, can we put the label inside it?

scoopr: At least I can't see it on `BindGroupLayout` or `CommandBuffer`. But I guess it could be moved for the ones that do have it?

kvark: Yes. It looks like all the things that have `life_guard` also need the label, so we should move it in.
For others few we can have it aside, like your PR does now.

scoopr:
Moved some labels to `LifeGuard`, hid some repeating code behind `LabelHelpers`.

scoopr:
Oh, also, I didn't yet update my -rs side locally, so the latest hasn't seen as much testing, but It Compiles(tm)

wgpu-core/src/command/allocator.rs
```diff
@@ -8,6 +8,9 @@ use crate::{
     PrivateFeatures, Stored, SubmissionIndex,
 };
 
+#[allow(unused_imports)]
+use crate::LabelHelpers;
```
kvark, line 12: Could we guard it as `#[cfg(debug_assertions)]` instead of `unused_imports`?

wgpu-core/src/hub.rs
```diff
@@ -20,6 +20,9 @@ use crate::{
     Epoch, Index,
 };
 
+#[allow(unused_imports)]
+use crate::LabelHelpers;
```
kvark, line 24: same here

wgpu-core/src/hub.rs
```diff
@@ -166,9 +177,10 @@ impl<T, I: TypedId> Storage<T, I> {
         self.insert_impl(index as usize, Element::Occupied(value, epoch))
     }
 
-    pub(crate) fn insert_error(&mut self, id: I) {
+    #[allow(unused_variables)]
```
kvark, line 180: why does it need unused variables?

wgpu-core/src/hub.rs
```diff
@@ -178,7 +190,7 @@ impl<T, I: TypedId> Storage<T, I> {
                 assert_eq!(epoch, storage_epoch);
                 Some(value)
             }
-            Element::Error(_) => None,
+            Element::Error(_, ..) => None,
```
kvark, line 193: uh, that's a weird thing, why not just `Error(..)`?

wgpu-core/src/lib.rs
```diff
@@ -166,14 +187,19 @@ impl Drop for MultiRefCount {
 struct LifeGuard {
     ref_count: Option<RefCount>,
     submission_index: AtomicUsize,
+    #[cfg(debug_assertions)]
+    pub(crate) label: Label<'static>,
```
kvark, line 191: any reason to not make it a `String`?

wgpu-core/src/hub.rs
```diff
@@ -759,6 +792,141 @@ pub trait GfxBackend: hal::Backend {
     fn get_surface_mut(surface: &mut Surface) -> &mut Self::Surface;
 }
 
+pub trait HasLabel {
```
kvark, line 795: It seems to me that this trait repeats logic that we pretty much have already in the code.

First, the `label` thing. We have `Borrow<RefCount>` implemented for all the resources, and all of them just return `&life_guard.ref_count`. So we should instead probably do something like:

```rust
trait Resource {
  const TYPE: &'static str;
  fn life_guard(&self) -> &LifeGuard;
}
```

We'd then use this trait in the state tracker instead of relying on `Borrow<RefCount>`.

Secondly, the "type" thing is already specified in `Hub::new` function when registries are created. We should unite the two, possibly by having something like this:
```rust
impl<T: Resource, I: TypedId, F: IdentityHandlerFactory<I>> Registry<T, I, F> {
    fn new_resource(backend: Backend, factory: &F) -> Self {
      Self::new(backend, factory, T::TYPE)
    }
}
```
And then call `Registry::new_resource` instead of `Registry::new` for everything that implements `Resource`.

kvark (commented):
Great work here! I have a few suggestions on how we can make it better, and avoid repeating ourselves in the codebase.

wgpu-core/src/lib.rs
scoopr: So yeah, since I noticed this `Label` typed used elsewhere, I figured might as well use it here too. The main benefit is that some of the internal labels don't need to be constructed as `String`, but can utilize the `Cow<'static, str>` here, for example in `swap_chain_get_current_texture_view` the label for the `TextureView`. 

If it made a heap allocation here, it would be tempting to leave it be a empty string (which I think doesn't heap allocate). But it would actually be nice to have the internally created resource to be labeled too.

But this is a tiny thing, I'm fine plain 'ol `String` as well.

kvark: I think we should be able to deal with `swap_chain_get_current_texture_view` later. For now, just make it something like
```rust
if cfg!(debug_assertions) { "swapchain".to_string() } else { String::new() }
```

wgpu-core/src/hub.rs
scoopr: There are few of them that don't have `LifeGuard`, quickly glancing at least `CommandBuffer` and  `BindGroupLayout`.

I'm not sure if there is really a semantic relation with `LifeGuard` and having a label, or is it just something convenient many of them have and could be coerced to function as something common for all of them? Might just be a naming thing.

The type name makes sense though.

scoopr:
Fixed the stupid mistakes, but did not yet touch the haslabel/lifeguard/typename thing.

wgpu-core/src/hub.rs
kvark: Just convenience, I suppose. I don't feel strongly about this. If you feel that it shouldn't be a part of `LifeGuard`, feel free to discard that idea.

scoopr: Ooh, didn't notice that the `kind` str is already given at `Registry::new`, should I just ask it from the `Storage`?

kvark: That's one way to reduce duplication, I suppose.
What I suggested above was to use the `Resource::TYPE` to *give* it to the storage, instead

scoopr:
I tried the `Resource` trait, I think it turned out ok, I use the `life_guard()` of it for the default impl for `label()` but it can be then implemented when there is no `life_guard` to be had.

It wasn't obvious to me how the `Borrow<RefCount>` should be replaced, but I didn't try to dig too deeply.

I should probably prepare a PR in `wgpu-rs` that just does the minimal braking change update (the labels for the error-ids), and handle the error message notes separately

scoopr:
A toned down the use of `Label` in preference to plain 'ol `String`. What do you think now?

scoopr:
Made the `*_error` methods take `Option<&str>`. I noticed while looking at the wgpu-rs side, that the input is almost always like that, and the call sites looked silly doing conversions to `Option<String>`.

wgpu-core/src/lib.rs
```diff
@@ -58,6 +58,22 @@ type Epoch = u32;
 pub type RawString = *const c_char;
 pub type Label<'a> = Option<Cow<'a, str>>;
 
+trait LabelHelpers<'a> {
+    fn new_borrowed(s: &'a str) -> Label<'a>;
```
kvark, line 62: since it's used only once, maybe we don't need this helper?

wgpu-core/src/lib.rs
```diff
@@ -163,17 +179,22 @@ impl Drop for MultiRefCount {
...
     ref_count: Option<RefCount>,
     submission_index: AtomicUsize,
+    #[cfg(debug_assertions)]
+    pub(crate) label: String,
 }
 
 impl LifeGuard {
-    fn new() -> Self {
+    #[allow(unused_variables)]
+    fn new(label: String) -> Self {
```
kvark, line 191: Can we instead do something like this:
```rust
pub fn new(label: &str) -> Self {
// something under `cfg(debug_assertions)` inside
}
```

This would allow all the call sites to not care about `debug_assertions`

kvark (commented):
Thank you! It's looking great, I think we are almost done.
There is a few more suggestions, plus we'd need this squashed for landing.

wgpu-core/src/lib.rs
scoopr: I don't understand, how does the callsite here care about `debug_assertions`?

It could be `&str` though I guess

kvark: Ah, I misread. It will only affect the call sites with static strings, like
```rust
            #[cfg(debug_assertions)]
            life_guard: LifeGuard::new("<device>".to_string()),
            #[cfg(not(debug_assertions))]
            life_guard: LifeGuard::new(String::new()),
```

wgpu-core/src/swap_chain.rs
```diff
@@ -183,7 +191,10 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                         layers: 0..1,
                         levels: 0..1,
                     },
-                    life_guard: LifeGuard::new(),
+                    #[cfg(debug_assertions)]
+                    life_guard: LifeGuard::new("<SwapChain View>"),
```
kvark, line 195: no need to split that

wgpu-core/src/device/mod.rs
```diff
@@ -273,7 +273,10 @@ impl<B: GfxBackend> Device<B> {
             mem_allocator: Mutex::new(mem_allocator),
             desc_allocator: Mutex::new(descriptors),
             queue_group,
-            life_guard: LifeGuard::new(),
+            #[cfg(debug_assertions)]
+            life_guard: LifeGuard::new("<device>"),
```
kvark, line 277: no need to split that

wgpu-core/src/instance.rs
```diff
@@ -212,11 +224,22 @@ impl<B: hal::Backend> Adapter<B> {
             raw,
             features,
             limits,
-            life_guard: LifeGuard::new(),
+            #[cfg(debug_assertions)]
+            life_guard: LifeGuard::new("<Adapter>"),
```
kvark, line 228: same here

wgpu-core/src/device/mod.rs
```diff
@@ -3576,7 +3623,10 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         }
 
         let swap_chain = swap_chain::SwapChain {
-            life_guard: LifeGuard::new(),
+            #[cfg(debug_assertions)]
+            life_guard: LifeGuard::new("<SwapChain>"),
```
kvark, line 3627: and here

kvark (commented):
Last note, and please squash!

wgpu-core/src/swap_chain.rs
scoopr: Oh duh, of course :)

scoopr:
_squash_

kvark:
Thank you!
bors r+

PR #935, opened 2020-09-22
Properly implement compute pass usage scopes

### LINKED ISSUE #934: Using buffer for draw/dispatch_indirect which is also being written to should fail
Right now it is possible to have a buffer bound for writing but at the same time pass it to dispatch/draw_indirect. Vulkan RenderDoc capture shows that the buffer is transitioned to (only) INDIRECT_COMMAND_READ_BIT, i.e. the write state is ignored (sadly, this seems to be something that is not caught by the Vulkan validation layer) ignoring the fact that it is still bound for writing.
Wgpu should validate this and fail.

kvark:
**Connections**
Fixes #934 
Closes #933

**Description**
The old barriers in a compute pass were totally wrong. They didn't combine usages across the bind groups, or with indirect buffer.
This change implements the *actual* semantics as seen by the spec. At dispatch time we gather the usages across bind groups needed by the active pipeline, possibly add the indirect buffer into the mix, and that combined used is what we transition into.

I think we can find a few interesting ideas on how to make this faster. In particular - how to avoid a separate `TrackerSet` per pass. But for now correctness is more important.

**Testing**
Tested on wgpu-rs examples

### DIFF
--- a/wgpu-core/src/command/bind.rs
+++ b/wgpu-core/src/command/bind.rs
@@ -219,6 +219,13 @@ impl Binder {
         }
     }
 
+    pub(super) fn list_active(&self) -> impl Iterator<Item = Valid<BindGroupId>> + '_ {
+        self.entries.iter().filter_map(|e| match e.provided {
+            Some(ref pair) if e.expected_layout_id.is_some() => Some(pair.group_id.value),
+            _ => None,
+        })
+    }
+
     pub(super) fn invalid_mask(&self) -> BindGroupMask {
         self.entries.iter().enumerate().fold(0, |mask, (i, entry)| {
             if entry.is_valid().unwrap_or(true) {
--- a/wgpu-core/src/track/mod.rs
+++ b/wgpu-core/src/track/mod.rs
@@ -480,6 +480,24 @@ impl<I: Copy + fmt::Debug + TypedId> ResourceState for PhantomData<I> {
 
 pub const DUMMY_SELECTOR: () = ();
 
+#[derive(Clone, Debug, Error)]
+pub enum UsageConflict {
+    #[error(
+        "Attempted to use buffer {id:?} as a combination of {combined_use:?} within a usage scope."
+    )]
+    Buffer {
+        id: id::BufferId,
+        combined_use: resource::BufferUse,
+    },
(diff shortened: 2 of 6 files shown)


wgpu-core/src/command/bind.rs
```diff
@@ -219,6 +219,13 @@ impl Binder {
         }
     }
 
+    pub(super) fn list_active<'a>(&'a self) -> impl 'a + Iterator<Item = Valid<BindGroupId>> {
```
cwfitzgerald, line 222: nit: this could be `fn list_active(&self) -> impl Iterator<Item = Valid<BindGroupId>> + '_`

wgpu-core/src/track/mod.rs
```diff
@@ -480,6 +480,22 @@ impl<I: Copy + fmt::Debug + TypedId> ResourceState for PhantomData<I> {
 
 pub const DUMMY_SELECTOR: () = ();
 
+#[derive(Clone, Debug, Error)]
+pub enum UsageConflict {
+    #[error("buffer {id:?} combined use is {combined_use:?}")]
```
cwfitzgerald, line 485: Could these errors provide a bit more description of what they are/what the user has to do to fix them? If I saw this error with no other context I would be very confused and have to go digging.

cwfitzgerald (changes requested):
Some nits/surface level comments, but looks good to me :)

wgpu-core/src/track/mod.rs
kvark: reworded this a bit... still could be improved

cwfitzgerald: That's significantly better, I think the only question I have would be "what is a usage scope?", as that terminology isn't on the api surface at all.

kvark: Usage scopes are defined in upstream spec. We could also have something in wgpu or wgpu-rs docs. Not sure how to link to that from the error message though.

cwfitzgerald: Fair enough, this is a doc issue so out of scope.

cwfitzgerald (approved):

PR #936, opened 2020-09-22
Prevent an invalid texture from being registered in device_create_texture

kunalmohan:
**Connections**

**Description**
Since we checked the usage of a `Feature` in `device_create_texture()` after registering the new `Id`, there could be a case where it returns an error after registering the texture as a valid one, which further panics when the client tries to register it as an `error_id`.

To solve this, the check for the usage of a `Feature` is now done before registering the new `Id`.

**Testing**
Tested with wgpu-rs examples and CTS in Servo.


kvark (approved):
nice, thank you!
bors r+

PR #938, opened 2020-09-23
Remove erroneous extra feature check

cwfitzgerald:
**Connections**

None, found during investigation of DI issue.

**Description**

This line of code just shows up twice in a row, checking the same feature and setting the same feature.

**Testing**

Untested, but shouldn't affect anything.


kvark (approved):
bors r+

PR #942, opened 2020-09-26
Fix clippy warnings about using if-let over match

lberrymage:
The specific lint triggered was [clippy::single_match](https://rust-lang.github.io/rust-clippy/master/index.html#single_match).

**Description**
Details found in the link above

**Testing**
N/A

### DIFF
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -931,19 +931,16 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             .register_identity(id_in, render_bundle, &mut token);
 
         #[cfg(feature = "trace")]
-        match device.trace {
-            Some(ref trace) => {
-                use crate::device::trace;
-                let (bundle_guard, _) = hub.render_bundles.read(&mut token);
-                let bundle = &bundle_guard[id];
-                let label = desc.label.as_ref().map(|l| l.as_ref());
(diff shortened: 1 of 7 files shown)


wgpu-core/src/command/bundle.rs
```diff
@@ -931,19 +931,16 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
-                let bundle = &bundle_guard[id];
-                let label = desc.label.as_ref().map(|l| l.as_ref());
-                trace.lock().add(trace::Action::CreateRenderBundle {
-                    id: id.0,
-                    desc: trace::new_render_bundle_encoder_descriptor(label, &bundle.context),
-                    base: BasePass::from_ref(bundle.base.as_ref()),
-                });
-            }
-            None => {}
+        if let Some(ref trace) = device.trace {
```
kvark, line 934: This has a higher MSRV (minimum supported Rust version), which is the reason it's not written this way originally.
Could you check what the exact MSRV is required with this change?

lberrymage: Oh whoops, I didn't realize that. MSRV for this is 1.43.0.

kvark: Hmm, it looks to be [matching Firefox 78](https://wiki.mozilla.org/Rust_Update_Policy_for_Firefox), so we should should be good.

Issue #861, opened 2020-08-02
API traces of crashed applications are not closed properly


kvark:
**Is your feature request related to a problem? Please describe.**
A trace is a sequence of actions. The sequence must end with "]". If an application crashed, it doesn't properly close the trace. See https://github.com/gfx-rs/wgpu/issues/785#issuecomment-667682847

**Describe the solution you'd like**
~~In `player`, if RON is failed to load, we should attempt to just add the trailing "]" and load again.~~
We should ensure that "]" is inserted on the drop of the tracing logic. There shouldn't be any manual patching on the player side.

**Describe alternatives you've considered**
~~Not doing anything - that's annoying to the users.~~
Do the patching on the player side.

**Additional context**
The `Drop` is already implemented, but not called because of https://github.com/rust-windowing/winit/issues/1850
Patching the RON is also related to backend substitution. The playtesting infra replaces the backend of a capture with the current one, and the player could do that too, automatically.

Kriyszig:
Can I give this issue a try?
I can create a preliminary patch which adds a trailing ']' if the RON loading fails and tries to load it again.
If there is a more novel way to go about this maybe we can discuss it over the patch?
Thank you

kvark:
Something I missed previously - we need to double-check if `]` can be inserted during writing. Maybe in a `drop()` implementation for something?

kvark:
I looked at this some more. I believe the current logic we have is already correct, and traces should be auto-closed. What's blocking is behavior, however, are these upstream issues:
  - https://github.com/zakarumych/gpu-descriptor/pull/11
  - https://github.com/rust-windowing/winit/issues/1850 (Apple systems only)

kvark:
The descriptor update has been published in https://crates.io/crates/gpu-descriptor/0.1.1. That should get us the desired behavior on Windows and Linux.
The fix for the winit issue on Apple is coming in https://github.com/rust-windowing/winit/pull/1853
We can consider this addressed (since we don't depend on `winit`).

PR #944, opened 2020-09-27
player: Add a trailing ']' to the trace if player fails to load the RON file

Kriyszig:
**Connections**
**Issue:** #861 

**Description**
In case of a panic, the RON trace ends without appending a ']' at
the end. Loading this trace as it is will cause a failure while
loading as this doesn't follow the specifications for a valid
RON file.
In case of an error while loading the RON file, we append the ']'
at the end of file and retry. If the error was indeed a missing ']'
the trace is loaded correctly.
If the load fails second time, this means that the trace needs a
manual review for correctness but appending ']' solves most of the
cases of invalid RON file obtained as a result of panic.

**Approach**

The `from_reader` function for RON can be found here - https://github.com/ron-rs/ron/blob/master/src/de/mod.rs#L48
Internally it works as follows

```rust
let mut bytes = Vec::new();
rdr.read_to_end(&mut bytes)?;
from_bytes(&bytes)
```
It reads the bytes into a vector and calls `from_bytes`

As we need to append thr `]` without modifying the trace to keep implementation pure, we extract the file content into a vector and call `from_bytes` on it - this returns a `Result` and in case we come across an error we retry by appending a `]` at the end of the vector and trying again.

**Testing**
The command `cargo test` was run in the player directory which shows all green on my local setup (GNU/Linux)

Being new contributor to this project I'll highly value any and all feedback you can give.
If you feel something needs to be changed / improved or added please let me know and I'll get on it asap!

Thank you

### DIFF
--- a/player/src/bin/play.rs
+++ b/player/src/bin/play.rs
@@ -10,6 +10,7 @@ use wgc::device::trace;
 
 use std::{
     fs,
+    io::Read,
     path::{Path, PathBuf},
 };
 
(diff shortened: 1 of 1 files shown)


kvark (commented):
This is great, thank you for the PR!
My main question before merging this would be - did you look into making the trace writer to add the `]` on `wgpu` side? This way, the player would need no workaround logic, and the produced RON would be valid by itself, which is useful for highlighting and stuff.

Kriyszig:
> did you look into making the trace writer to add the ] on wgpu side? 
Ah! I haven't yet.
I believe the process would be to find potential spots the program can panic at and add a trailing ']' to the trace.
I can look into this if you would like and come back with the finding

Thank you for the quick review.

kvark:
I mean, consider that path to be unfeasible if there is more than one place to patch. My hope is that - most of the crashes are panics, and panics still trigger `drop` on our objects, so we should have some space to finish the job, by just carefully handling `drop` of the tracing struct.

kvark:
Closing because of https://github.com/gfx-rs/wgpu/issues/861#issuecomment-773444836

PR #957, opened 2020-10-05
Enable RBA feature where available

kvark:
**Connections**
Related to https://github.com/gpuweb/gpuweb/issues/955

**Description**
Until we have a better plan in place, enabling RBA is the least we can do.

**Testing**
Should work!


grovesNL (approved):
bors r+

PR #958, opened 2020-10-05
[0.6] fix device feature requests

kvark:
**Connections**
Looks like we were requesting a little bit too much?
Also includes #936  and #957

**Description**
Fix the features requested.

**Testing**
Confirmed by the virtue of `println!` :)


cwfitzgerald (approved):
Well that's a helluva oversight. LGTM

kvark:
Bors r=cwfitzgerald

PR #963, opened 2020-10-06
Add repr(C) on mapping structs

kvark:
**Connections**
Required for https://phabricator.services.mozilla.com/D92636

**Description**
Allows C clients to avoid redefining those structs.

**Testing**
No need


kvark:
bors r=trivial

PR #966, opened 2020-10-08
Immediate resource destruction and freeing

### LINKED ISSUE #964: Implement buffer and texture destroy()
WebGPU upstream has that: in a lot of cases we want to free the GPU memory right away, even if some CPU objects still refer to it. Using any of these objects is invalid after that happens.

This is what `destroy()` would do on buffers and textures.

Rely on GC.

**Additional context**
Would help DungeonFog (cc @krupitskas ) - they resize textures and rely on internal GC to clean up memory.

kvark:
**Connections**
Fixes #964

**Description**
We are making it so a buffer or a texture can have their native resources freed while they are still referenced, so without waiting for GC.

In addition, the PR adds a few missing cases where error IDs should have been handled, like at render pass encoding.

**Testing**
Tested on wgpu-rs examples, see https://github.com/gfx-rs/wgpu-rs/pull/591

### DIFF
--- a/player/src/lib.rs
+++ b/player/src/lib.rs
@@ -149,6 +149,9 @@ impl GlobalPlay for wgc::hub::Global<IdentityPassThroughFactory> {
                 self.device_maintain_ids::<B>(device).unwrap();
                 self.device_create_texture::<B>(device, &desc, id).unwrap();
             }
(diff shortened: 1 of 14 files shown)


cwfitzgerald (approved):
LGTM!

bors r+

Issue #950, opened 2020-10-01
Run playtests on CI


kvark:
**Is your feature request related to a problem? Please describe.**
We started rolling out the data-based reference tests - [playtests](https://github.com/gfx-rs/wgpu/blob/97d1a7eeb9926ca6a894d78581ea03375029b420/player/tests/test.rs). They can be run locally as a part of `cargo test`. It would be great to run them to some extent on CI.

**Describe the solution you'd like**
Depending on a platform the CI runs on, there can be software-only implementations available:
  - Windows CI can use the WARP adapter
  - Linux CI can use "lavapipe" on Vulkan, or "llvmpip" on GL
  - macOS CI could possibly just run Metal?

**Describe alternatives you've considered**
Not running CI.

**Additional context**
https://www.phoronix.com/scan.php?page=news_item&px=Mesa-Vulkan-Lavapipe

kvark:
Fixed by #968

PR #968, opened 2020-10-08
CI playtests for Dx12, Vulkan

DevOrc:
**Connections**
#950

**Description**
Implements CI testing for Vulkan and DX12. Also, cargo test stdout will no longer be captured so that people can verify that the playtests actually ran. 

Note: If no adapters are found, nothing will run and the tests will pass. This might become a problem if the installation no longer works and no one notices. 

**Testing**
Passed CI Tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
           - name: MacOS Nightly
             os: macos-10.15
             channel: nightly
-            build_command: cargo test
+            build_command: cargo test -- --nocapture
             additional_core_features:
             additional_player_features:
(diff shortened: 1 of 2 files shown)


kvark (approved):
This is great, thank you!

I wish we could call those CI entries right. Currently, you have to know where to look for the tests, i.e. on Ubuntu it's just called `Run sudo add-apt-repository ppa:oibaf/graphics-drivers;` rather misleadingly.

On the good side, I'm excited to see Vulkan llvm and D3D12 tests running 🚀 
We'll need to see why macOS adapters are not found at some points...

bors r+

PR #446, opened 2020-01-11
Prefer discrete GPU for PowerPreference::Default

aloucks:

PR #969, opened 2020-10-09
Never prefer CPU devices over GPUs

myfreeweb:
Mesa (development repo) now has a software renderer for Vulkan (`lavapipe`), it is incomplete for now and can't handle iced, so my app exploded after a mesa rebuild :)

Looks like someone forgot to push CPU renderers to the lowest priority.

(This commit should be cherry-pickable onto master I think, since this code hasn't really changed)

### DIFF
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -480,7 +480,8 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             return Err(RequestAdapterError::NotFound);
         }
 
-        let (mut integrated, mut discrete, mut virt, mut other) = (None, None, None, None);
+        let (mut integrated, mut discrete, mut virt, mut cpu, mut other) =
+            (None, None, None, None, None);
 
         for (i, ty) in device_types.into_iter().enumerate() {
             match ty {
@@ -493,7 +494,10 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 hal::adapter::DeviceType::VirtualGpu => {
                     virt = virt.or(Some(i));
                 }
-                _ => {
+                hal::adapter::DeviceType::Cpu => {
+                    cpu = cpu.or(Some(i));
+                }
+                hal::adapter::DeviceType::Other => {
                     other = other.or(Some(i));
                 }
             }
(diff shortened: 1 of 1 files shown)


aloucks:
Yes, please! It's very odd that the `Default` power preference selects my integrated GPU over the discrete one. Chrome, Firefox, and every game I've played always prefer the discrete.

This was fixed in #446, however @kvark suggested checking for a battery and selecting the GPU based on that. The power detection code was later removed. I don't really understand the reasoning behind the current logic. As far as I know mobile and laptops don't really have multiple GPUs. Preferring an integrated GPU over discrete is generally only relevant on desktop and arguably the discrete GPU is always want the end user wants.

wgpu-core/src/instance.rs
```diff
@@ -493,6 +494,9 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 hal::adapter::DeviceType::VirtualGpu => {
                     virt = virt.or(Some(i));
                 }
+                hal::adapter::DeviceType::Cpu => {
+                    cpu = cpu.or(Some(i));
+                }
                 _ => {
```
kvark, line 500: let's list this variant explicitly as `hal::adapter::DeviceType::Other`

kvark (commented):
Nice, thank you! Just one small thing to address here

kvark:
@aloucks if you know you need the high-performance one, you should be passing that preference. Same for integrated. The `Default` is only for those who don't care - that's the current logic.
The reason we cut out the `battery` dependency is simply because you can do it yourself: hook up `battery` and then form the `PowerPreference` based on what it reports.

aloucks:
@kvark I understand that it can be set explicitly. I'm trying to understand the reasoning behind the behavior for `Default`. Why prefer integrated when a discrete GPU exists? Generally, the only time two or more adapters will be present is on desktop systems and desktops are not running on battery power.

If the author of the code specifies `Default` (as is in all the examples) and doesn't expose a way to change it at runtime, the end user gets stuck with using the "slow" GPU. I have the integrated intel GPU enabled for testing purposes but I don't want anything to select it by default.

kvark:
> Generally, the only time two or more adapters will be present is on desktop systems and desktops are not running on battery power.

This is incorrect. Many laptops have discrete GPUs in addition to integrated ones. On these laptops, choosing the discrete would mean faster power draw, which is undesirable if on battery.

I think what we need is to remove `Default` variant entirely. It's not useful much today, and we can have the same semantic by just `impl Default for PowerPreference`. Thoughts  - @grovesNL @cwfitzgerald ?

aloucks:
@kvark What about removing the `Default` variant and *not* having an `impl Default`? This would make it explicit and deterministic. 

Alternatively, an `impl Default` could check an environment variable such as `WGPU_DEFAULT_POWER_PREFERENCE`.

grovesNL:
@kvark agreed, removing it sounds good to me

wgpu-core/src/instance.rs
myfreeweb: done

cwfitzgerald:
Full agree. Let the user either choose a preference, or manually choose an adapter themselves (if on native).

kvark:
@myfreeweb would you mind creating the same PR against `master`?

PR #970, opened 2020-10-12
Fix coalescing of descriptor sets

kvark:
**Connections**
Fixes https://github.com/gfx-rs/wgpu-rs/issues/592

**Description**
I forgot that all the descriptors in a write have to be the same type and visibility. This should be fixed now.

**Testing**
Tested on wgpu-rs examples, see https://github.com/gfx-rs/wgpu-rs/pull/593

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -2355,19 +2355,40 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 }
             }
 
-            if let Some(start_binding) = write_map.keys().next().cloned() {
-                let descriptors = write_map
-                    .into_iter()
-                    .flat_map(|(_, list)| list)
-                    .collect::<Vec<_>>();
-                let write = hal::pso::DescriptorSetWrite {
-                    set: desc_set.raw(),
-                    binding: start_binding,
-                    array_offset: 0,
-                    descriptors,
-                };
+            if !write_map.is_empty() {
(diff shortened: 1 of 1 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -2355,19 +2355,40 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+                    let layout = &bind_group_layout.entries[&binding];
+                    let ty = match layout.ty {
+                        wgt::BindingType::UniformBuffer { .. } |
+                        wgt::BindingType::StorageBuffer { .. } => DescriptorType::Buffer,
+                        wgt::BindingType::Sampler { .. } => DescriptorType::Sampler,
+                        wgt::BindingType::SampledTexture { .. } |
+                        wgt::BindingType::StorageTexture { .. } => DescriptorType::TextureView,
+                    };
+                    if layout.visibility == prev_stages && ty == prev_ty {
+                        prev_stages = layout.visibility;
```
grovesNL, line 2378: Looks like these assignments should be under the `else` branch instead?

kvark: Oh wow, nicely spotted! I was refactoring this branch to the opposite and forgot this piece :sweat_smile: 
Now I wonder how/why it works even

kvark:
bors r=grovesNL

kvark:
bors r=grovesNL

PR #980, opened 2020-10-13
Group binding writes by wgt::BindingType

### LINKED ISSUE #979: vkUpdateDescriptorSets call on image & sampler at once, leading to validation error
When putting a StorageTexture followed by a SampledTexture in a bind group layout, creating any bind group will result in this error:
```
VALIDATION [VUID-VkWriteDescriptorSet-descriptorCount-00317 (-284942183)] : Validation Error: [ VUID-VkWriteDescriptorSet-descriptorCount-00317 ] Object 0: handle = 0xedbd50000000010, type = VK_OBJECT_TYPE_DESCRIPTOR_SET; | MessageID = 0xef042099 | vkUpdateDescriptorSets() failed write update validation for VkDescriptorSet 0xedbd50000000010[] with error: Attempting write update to VkDescriptorSet VkDescriptorSet 0xedbd50000000010[] allocated with VkDescriptorSetLayout VkDescriptorSetLayout 0x731f0f000000000a[] binding index #2 (1 from dstBinding offset) with a different stageFlag and/or descriptorType from previous bindings. All bindings must have consecutive stageFlag and/or descriptorType across a VkWriteDescriptorSet. The Vulkan spec states: All consecutive bindings updated via a single VkWriteDescriptorSet structure, except those with a descriptorCount of zero, must have identical descriptorType and stageFlags (https://vulkan.lunarg.com/doc/view/1.2.141.0/windows/1.2-extensions/vkspec.html#VUID-VkWriteDescriptorSet-descriptorCount-00317)
object info: (type: DESCRIPTOR_SET, hndl: 1070683532407341072)
```

Created a repro here inside the wgpu-rs' compute sample
https://github.com/Wumpf/wgpu-rs/commit/48ab7b606697ad43e38eab96d726046238d90dc8

(haven't played around too much with the specifics, so could be that there's some more details in here that make the stars align so that things go wrong. Don't even know if this is compute shader specific.. probably not? ;))

kvark:
**Connections**
Follows up https://github.com/gfx-rs/wgpu/pull/970
Fixes #979

**Description**
The problem was that Vulkan restriction applies to its own descriptor types, as specified in the layout. I originally interpret this as types specified in the descriptor set. So we erroneously considered `StorageTexture` and `SampledTexture` to be in the same descriptor write.

This PR makes it use the `wgt::BindingType` instead. It's a bit richer than Vulkan side, but still correct. We are just more conservative than we have to be.

I think gfx-hal API could be better here. Filed https://github.com/gfx-rs/gfx/issues/3408 to look more.

**Testing**
Tested on #979 test case (thanks!)

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -701,13 +701,13 @@ impl<B: GfxBackend> Device<B> {
                     .map_or(1, |v| v.get() as hal::pso::DescriptorArrayIndex), //TODO: consolidate
                 stage_flags: conv::map_shader_stage_flags(entry.visibility),
                 immutable_samplers: false, // TODO
-            })
-            .collect::<Vec<_>>(); //TODO: avoid heap allocation
+            });
+        let desc_counts = raw_bindings.clone().collect();
 
         let raw = unsafe {
             let mut raw_layout = self
                 .raw
-                .create_descriptor_set_layout(&raw_bindings, &[])
+                .create_descriptor_set_layout(raw_bindings, &[])
                 .or(Err(DeviceError::OutOfMemory))?;
             if let Some(label) = label {
                 self.raw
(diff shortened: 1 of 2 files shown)


cwfitzgerald (approved):
LGTM!

bors r+

Issue #981, opened 2020-10-13
`wgpu_render_pass_insert_debug_marker` outputs garbled text for multiple markers


kazimuth:
**Description**
I'm debugging a wgpu application with RenderDoc, I noticed that inserting multiple debug markers into a single render pass results in strange output.

**Repro steps**
Note: I found this via `wgpu-rs` but I believe the issue is with `wgpu-core`. I ran the following code, integrated into a larger example that submitted and ran the render pass every frame.
```rust
render_pass.insert_debug_marker("this is a test");
render_pass.insert_debug_marker("hello");
render_pass.insert_debug_marker("how are you");
render_pass.insert_debug_marker("i love to walk on the beach and feed seagulls to the clams");
```

**Expected vs observed behavior**
When I run the example and capture a frame with RenderDoc, I see the following debug output:
```
this is a test
this 
this is a t
this is a testhellohow are youi love to walk on the beach
```
As opposed to what it should be.

**Extra materials**
The error is in the following pieces of code:

https://github.com/gfx-rs/wgpu/blob/f963193be193022fb02ac85d5d0fb3088bf7373e/wgpu-core/src/command/render.rs#L1943-L1957

https://github.com/gfx-rs/wgpu/blob/f963193be193022fb02ac85d5d0fb3088bf7373e/wgpu-core/src/command/render.rs#L1523-L1528

Because the `InsertDebugOutput` struct only keeps track of the length of the message and not its starting point, subsequent debug messages after a first one will slice through previous messages before getting to their own content.

**Platform**
OpenSUSE Tumbleweed, Mesa Vulkan, wgpu-rs 0.6.0, wgpu-core 0.6.4, wgpu-types 0.6.1

kvark:
Great finding, and you practically spelled out a fix! Would you mind turning this into a PR?

kazimuth:
Sure, no problem.

kazimuth:
Closed by #982

PR #982, opened 2020-10-13
Fix debug markers in render passes (#981)

kazimuth:
Hey, that was easy. Didn't change anything in the debug groups markers / command pass markers, since afaict they work correctly.

I didn't add a test here because I'm not sure how to use the `player` testing apparatus, but I figure since I'm just adding a line of code that's already there for compute passes ([here](https://github.com/gfx-rs/wgpu/blob/master/wgpu-core/src/command/compute.rs#L452)) it should be alright. Let me know if there's anything else I should add.


kvark (approved):
ohhh, that was indeed just a copy-paste error
Thank you for fixing this!
bors r+

Issue #983, opened 2020-10-13
Vulkan hangs after a certain render sequence, and either panics, hangs or loses device


Dinnerbone:
**Description**
It's a little hard to turn this into a small repro case so please forgive the vagueness.

After submitting a frame to wgpu whilst using Vulkan backend, Vulkan seems to become unstable and this manifests itself in a few ways:
- A hang of the application
- Graphics device crashes and PC dies (this happened to me a few times)
- The submit seemingly returns okay but nothing actually happened, and the device will become lost the next time we try to draw a frame

Our application has two ways to reproduce the bug:
- When rendering to a window, we repeatedly submit frames as a typical game would. This often just locks up but sometimes will gracefully give you an error about the device being lost.
- Rendering one single frame to a texture, saved to disk.

In this second case, we perform the following sequence of events:
- Create a texture
- Draw a frame to a command encoder
- Submit the command encoder to the queue
- Copy the texture buffer to disk, much like the capture example in wgpu-rs

Seemingly the submit returns okay but the texture is completely empty, when we'd expect to see some graphics in it. The application then freezes (at least, for me on windows - this seems to vary) when dropping wgpu::Instance. For reference, the image it spits out should be identical to [this one](https://raw.githubusercontent.com/ruffle-rs/visual-tests/master/tests/drawing_api/fills_and_lines/actual.png).

I've taken a trace of this single-frame capture and had to manually close the toml as the recording can't finish. This seems to freeze when played back, but I'm unable to get renderdoc to play nice and see anything from it.

This worked for us in the past, I think as soon as 24 days ago I was running this without issues. The same code, unchanged, no longer works today.

**Repro steps**
I haven't been able to create a minimal reproducible example, but you can see it in our project with the following steps:
- Grab [this swf](https://github.com/ruffle-rs/visual-tests/raw/master/tests/drawing_api/fills_and_lines/test.swf)
- Clone [Ruffle](https://github.com/ruffle-rs/ruffle/)
- `cargo run --package=ruffle_desktop -- test.swf` if you want to see it visually, with multiple frames
- `cargo run --package=exporter -- test.swf` if you want to see the single frame saved to a texture on disk

You can apply this commit to reduce the amount of rendering done to the bare minimum that still crashes, with that particular swf: https://github.com/Dinnerbone/ruffle/commit/b4f173dbbc9db0128cd2c591d71efbd49e024852

**Expected vs observed behavior**
[...]

PR #986, opened 2020-10-14
[0.6] Allign stencil reference between state setting and pipeline creation

kvark:
**Connections**
Found this when running Ruffle from #983 

**Description**
There are 2 problems:
  1. setting the stencil reference can't be valid if the pipeline doesn't expect this (d'oh)
  2. our code that created the pipeline used slightly stricter conditions than the code setting the stencil, so these diverged a tiny bit

**Testing**
Tested on Ruffle

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## wgpu-core-0.6.5 (2020-10-14)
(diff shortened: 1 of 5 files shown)


cwfitzgerald (approved):
Ship it!

bors r+

PR #987, opened 2020-10-14
Align stencil reference flags between pipeline creation and setting

kvark:
Sibling of #986 for master


kvark:
bors r=cwfitzgerald

PR #990, opened 2020-10-15
Simplify descriptor writes

kvark:
**Connections**
Reverts #980 and #970
Takes advantage of https://github.com/gfx-rs/gfx/pull/3410

**Description**
Weighting all the pros and cons, I figured it's easier to adjust the Vulkan backend a little than treating it as a lowest common denominator for this issue.

**Testing**
Tested on wgpu-rs examples (on Vulkan).


cwfitzgerald (approved):
bors r+

PR #998, opened 2020-10-20
Fix regression in device_create_swap_chain

qthree:
**Description**
My app works fine with wgpu 0.6 but after upgrade to master branch of `wgpu-rs` it started to panic inside `Device::create_swap_chain`.

**Repro steps**
That happens when there are at least 3 windows opened right after app's startup. My [viewports](https://github.com/qthree/viewports) crate can reproduce this issue. Steps:
```
git clone -b wgpu-master https://github.com/qthree/viewports
cd viewports
RUST_BACKTRACE=1 cargo run --example wgpu --all-features
```
First run is OK, but if you drag two "virtual" windows out of native window and then restart app, it panics.

**Connections**
It's caused by the new behavior of `hub::Storage::remove` introduced in #925, it panics while trying to remove Element::Vacant.

**Proposed solution**
Previously `Storage::remove` returned `None` on `Element::Vacant` and this behavior was expected by `device::Global::device_create_swap_chain`. At least, [this code](https://github.com/gfx-rs/wgpu/blob/8ce2530b6963982ef9a58f7efac682d13ce4c508/wgpu-core/src/device/mod.rs#L3561) doesn't look like it should panic implicitly.
After #925 `device::Global::device_create_swap_chain` was changed to use new `Storage::try_remove`, and it takes into account out of range access, but panics on Element::Vacant.
Since that's the only place where `Storage::try_remove` is used, i changed this function's body back to the body of old `Storage::remove`, and left new `Storage::remove` as is, because it's used in other places.

**Testing**
Since `Storage::try_remove` is used only in one place, `device::Global::device_create_swap_chain`, there shouldn't be any impact on another pieces of code. Also, function code is as same as on stable wgpu 0.6, so there shouldn't be any new bugs. Still, this fixed  regression encountered by me, and `cargo test` of `wgpu-core` is passing.
I'm not sure if I should add unit test to cover this case, to save it from regressions in future (and what meaningful test would look like).

**Alternatives**
Don't call `Storage::try_remove` if it's first swapchain creation for surface. Probably, by checking id returned by `surface_id.to_swap_chain_id(B::VARIANT)` with `Storage::contains`. But `Storage::contains` will panic on out of range id.

### DIFF
--- a/wgpu-core/src/hub.rs
+++ b/wgpu-core/src/hub.rs
@@ -183,13 +183,19 @@ impl<T, I: TypedId> Storage<T, I> {
         }
     }
 
-    //Prevents panic on out of range access.
+    // Prevents panic on out of range access, allows Vacant elements.
     pub(crate) fn try_remove(&mut self, id: I) -> Option<T> {
-        let (index, epoch, backend) = id.unzip();
+        let (index, epoch, _) = id.unzip();
         if index as usize >= self.map.len() {
-            return None;
+            None
+        } else if let Element::Occupied(value, storage_epoch) =
+            std::mem::replace(&mut self.map[index as usize], Element::Vacant)
+        {
+            assert_eq!(epoch, storage_epoch);
+            Some(value)
+        } else {
+            None
         }
-        self.remove(I::zip(index, epoch, backend))
     }
 
     pub(crate) fn iter(&self, backend: Backend) -> impl Iterator<Item = (I, &T)> {


kvark:
What about the `Error` variant though? It seems a bit strange to treat it as `Vacant` there, and the old code couldn't be expecting it either. Maybe we could still treat `Error` as an error?

qthree:
This, I'm not sure. Right now this code used only on swap chain recreation (window resizing), so it's probably makes sense to ignore error from the last swap chain creation, but it's not like we have code that sets up swap chain error. So it doesn't make any difference in practice.
More importantly, with my next PR (`Drop` for `SwapChain`), this code can become redundant at all. But this PR still can be accepted as regression bug fix.
Panic on `Error` or not - that's your call, I've not dived that deep yet, so i just copied old behavior.

kvark:
ok, seems reasonable.
bors r+

PR #1000, opened 2020-10-20
Elide redundant set_pipeline calls.

Kimundi:
**Description**
This adds an check in each `set_pipeline()` call for wether the same pipeline id has already been set previously. This should cause free performance wins for suboptimal usage of the wgpu API, while having neglible overhead otherwise.

An example scenario for where this would be useful is a game that just blindly sets all render state for each object, but has few actually different objects:

```rust
for game_obj in game_objs {
    rpass.set_pipeline(...);
    rpass.set_bind_group(...);
    rpass.set_vertex_buffer(...);
    rpass.draw(...);
}
```

**Testing**
Ideally we would have tests that check that pipeline changes have been ellided, but I'm not sure how to write them in the current codebase.

**Future work**
- We could extend this kind of redundant state change detection to most `.set_*` methods on `RenderPass`es.
- If we want to guarantee this behavior in the API, we should also document it.

### DIFF
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -38,7 +38,7 @@
 #![allow(clippy::reversed_empty_ranges)]
 
 use crate::{
-    command::{BasePass, DrawError, RenderCommand, RenderCommandError},
+    command::{BasePass, DrawError, RenderCommand, RenderCommandError, StateChange},
     conv,
     device::{
         AttachmentData, DeviceError, RenderPassContext, MAX_VERTEX_BUFFERS, SHADER_STAGE_COUNT,
@@ -515,6 +515,7 @@ struct State {
     raw_dynamic_offsets: Vec<wgt::DynamicOffset>,
     flat_dynamic_offsets: Vec<wgt::DynamicOffset>,
     used_bind_groups: usize,
+    pipeline: StateChange<id::RenderPipelineId>,
 }
 
 impl State {
@@ -698,6 +699,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 raw_dynamic_offsets: Vec::new(),
                 flat_dynamic_offsets: Vec::new(),
                 used_bind_groups: 0,
+                pipeline: StateChange::new(),
             };
             let mut commands = Vec::new();
             let mut base = bundle_encoder.base.as_ref();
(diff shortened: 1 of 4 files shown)


kvark:
Thank you for making this PR!
My main concern with it is consistency of the code. So far, the heavy work of figuring out what bind groups need to be rebound happens at the point of running a pass, not encoding it. So if we land the PR like this, we'd have some of the state deduplication logic running during encoding, and some during finishing a pass. It would be nice to have that compact in one place.

Kimundi:
Alright, then I'll move it to that location. :)

Kimundi:
I've moved the state change detection to the actual execution of the render/compute passes, and also added a tiny abstraction around the state change detection for better readability and to make the future extension to other state change detections easier.

wgpu-core/src/command/bundle.rs
```diff
@@ -60,6 +60,8 @@ use std::{
 };
 use thiserror::Error;
 
+use super::StateChange;
```
kvark, line 63: please use the existing `use crate::{}` for adding new things

wgpu-core/src/command/compute.rs
```diff
@@ -24,6 +24,8 @@ use wgt::{BufferAddress, BufferUsage, ShaderStage};
 
 use std::{fmt, iter, str};
 
+use super::StateChange;
```
kvark, line 27: same here

wgpu-core/src/command/mod.rs
```diff
@@ -270,3 +270,19 @@ where
...
+#[derive(Debug)]
+struct StateChange<T> {
+    last_state: Option<T>,
+}
+
+impl<T: Copy + PartialEq> StateChange<T> {
+    fn new() -> Self {
+        Self { last_state: None }
+    }
+    fn already_set(&mut self, new_state: T) -> bool {
```
kvark, line 283: this method is mutating, but it's named as if it's an immutable check, which is confusing.
Maybe we can call it just `set`?

wgpu-core/src/command/render.rs
```diff
@@ -45,6 +45,8 @@ use std::{
     str,
 };
 
+use super::StateChange;
```
kvark, line 48: same here - please add to an existing `use` statement

kvark (commented):
Thank you for updating!
Please address the last notes and squash the commits to proceed

kvark:
Please squash the stuff!

Kimundi:
I'm on it! :) I usually do a squashed force push as the very last step, to keep history around for as long as there might be comments about the actual changes.

kvark (approved):
bors r+

kvark:
Sorry, I just figured out that we have some duplication of logic now between `pipeline: OptionalState,` and your new field. We don't need the former. Could you remove it, and make it so the correctness check `is_ready` considers your `last_pipeline` to be some?

Kimundi:
Sure!

kvark (approved):
This is very sleek, thank you!
bors r+

kvark:
bors retry

PR #1007, opened 2020-10-24
Improve cubemap validation

### LINKED ISSUE #952: Creating a CubeMap View from Image With Dimension D3 Causes Erronious "Not Enough Memory Left" Error on Vk
**Description**

All cubemap images must be made with dimension D2. If you make it with D3, wgpu validation should catch it. Currently, when you make the cubemap texture view, it returns a non-sensical out-of-memory error.

**Repro steps**

Clone https://github.com/cwfitzgerald/wgpu-rs/tree/d3-cubemap and run the skybox example. Panics on vulkan, returns black if set to DX12. Unknown behavior on mac.

Expected: Errors with useful error message.
Observed: Errors with out of memory error.

**Extra materials**

https://github.com/gfx-rs/wgpu/blob/master/wgpu-core/src/device/mod.rs#L1406 assumes any call failure is due to an out of memory condition, causing the error message.

https://github.com/gfx-rs/wgpu/blob/master/wgpu-core/src/device/mod.rs#L1346-L1355 validation should just need to be added here. Should be fairly easy.

cwfitzgerald:
**Connections**

Closes #952.

**Description**

We were missing depth-size and texture dimension validation when creating Cubemap and CubemapArray views.

I split the depth-size errors so I can give a more helpful error message in both cases.

**Testing**

Tested on the skybox example when made intensionally incorrect.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -27,7 +27,9 @@ use hal::{
 };
 use parking_lot::{Mutex, MutexGuard};
 use thiserror::Error;
-use wgt::{BufferAddress, BufferSize, InputStepMode, TextureDimension, TextureFormat};
+use wgt::{
+    BufferAddress, BufferSize, InputStepMode, TextureDimension, TextureFormat, TextureViewDimension,
+};
 
 use std::{
     borrow::Cow,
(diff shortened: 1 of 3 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -1478,7 +1480,51 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         let device = &device_guard[texture.device_id.value];
 
         let view_kind = match desc.dimension {
-            Some(dim) => conv::map_texture_view_dimension(dim),
+            Some(dim) => {
+                use hal::image::Kind;
+
+                let valid_view_dim = match texture.dimension {
```
kvark, line 1486: can we rewrite this in a simpler way without `matches!`, like:
```rust
let texture_dim = match dim {
  TextureViewDimension::D1 => TextureDimension::D1,
  ... => TextureDimension::D2,
  TextureViewDimension::D3 => TextureDimension::D3,
};
if texture.dimension != texture_dim {
  return Err(..);
}
```
We could even put this conversion into `wgt` if it's useful for public

wgpu-core/src/device/mod.rs
```diff
@@ -1478,7 +1480,51 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+                if !valid_view_dim {
+                    return Err(
+                        resource::CreateTextureViewError::InvalidTextureViewDimension {
+                            view: dim,
+                            image: texture.dimension,
+                        },
+                    );
+                }
+
+                match (dim, texture.kind) {
```
kvark, line 1507: this could be rewritten somewhat cleaner as:
```rust
if let Kind::D2(_, _, depth, _) = texture.kind {
    match dim {
      TextureViewDimension::Cube if depth != 6 => return Err(...),
      TextureViewDimension::CubeArray if depth % 6 != 0 => return Err(...),
      _ => {}
    }
}
```

wgpu-core/src/device/mod.rs
cwfitzgerald: Seems like it could be, I'll add it there.

kvark (approved):
Thanks!
bors r+

PR #1008, opened 2020-10-27
Add helpers to convert passes to serialized forms

kvark:
**Connections**
Stuff I needed to fix in the upcoming Gecko WebGPU update.

**Description**
It allows us to use the tracing `Command` more aggressively.

**Testing**
Tested in Gecko, doesn't need testing in wgpu


grovesNL (approved):
bors r+

PR #1012, opened 2020-10-30
Check array layer counts

kvark:
**Connections**
Prevents https://github.com/gfx-rs/wgpu-rs/issues/601

**Description**
Follows the spec, yo!

**Testing**
Not tested, but not terribly complicated either.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -1479,7 +1479,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             .ok_or(resource::CreateTextureViewError::InvalidTexture)?;
         let device = &device_guard[texture.device_id.value];
 
-        let view_kind =
+        let view_dim =
             match desc.dimension {
                 Some(dim) => {
                     use hal::image::Kind;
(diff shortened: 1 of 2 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -1513,14 +1513,17 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
                 }
                 None => match texture.kind {
-                    hal::image::Kind::D1(_, 1) => hal::image::ViewKind::D1,
-                    hal::image::Kind::D1(..) => hal::image::ViewKind::D1Array,
-                    hal::image::Kind::D2(_, _, 1, _) => hal::image::ViewKind::D2,
-                    hal::image::Kind::D2(..) => hal::image::ViewKind::D2Array,
-                    hal::image::Kind::D3(..) => hal::image::ViewKind::D3,
+                    hal::image::Kind::D1(..) => wgt::TextureViewDimension::D1,
+                    hal::image::Kind::D2(_, _, depth, _)
+                        if depth > 1 && desc.array_layer_count.is_none() =>
```
cwfitzgerald, line 1521: This seems weird. I would expect if I pass `None` for texture kind, but pass an array layer count, I would get a D2Array texture not a D2. If this isn't what the spec dictates, so be it, but it caught my eye.

cwfitzgerald (commented):
One note, otherwise LGTM

wgpu-core/src/device/mod.rs
kvark: IIRC, one of the points of confusion was:
What if you pass an array layer count of 1 ? how do we figure out what you actually mean?

So we are open to make the spec cleaner, but this is what we have to date.

kvark:
Thank you for the review!
bors r=cwfitzgerald

PR #1018, opened 2020-11-09
Add context for errors originating from commads in a Pass

Kimundi:
**Description**
This adds some error context information for some errors that can happen in a `{Render,Compute}Pass` to make error messages more clear.

Combined with https://github.com/gfx-rs/wgpu-rs/pull/613, errors in passes will look like this:
```
wgpu error: Validation error

Caused by:
    In a RenderPass
    In a draw command
    index 3600 extends beyond limit 38
```

### DIFF
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -38,7 +38,10 @@
 #![allow(clippy::reversed_empty_ranges)]
 
 use crate::{
-    command::{BasePass, DrawError, RenderCommand, RenderCommandError, StateChange},
+    command::{
+        BasePass, DrawError, MapPassErr, PassErrorScope, RenderCommand, RenderCommandError,
+        StateChange,
+    },
     conv,
     device::{
         AttachmentData, DeviceError, RenderPassContext, MAX_VERTEX_BUFFERS, SHADER_STAGE_COUNT,
@@ -656,15 +659,45 @@ impl State {
 
 /// Error encountered when finishing recording a render bundle.
 #[derive(Clone, Debug, Error)]
-pub enum RenderBundleError {
+pub(super) enum RenderBundleErrorInner {
     #[error(transparent)]
     Device(#[from] DeviceError),
     #[error(transparent)]
-    RenderCommand(#[from] RenderCommandError),
+    RenderCommand(RenderCommandError),
     #[error(transparent)]
     ResourceUsageConflict(#[from] UsageConflict),
     #[error(transparent)]
     Draw(#[from] DrawError),
+}
+
+impl<T> From<T> for RenderBundleErrorInner
+where
+    T: Into<RenderCommandError>,
+{
+    fn from(t: T) -> Self {
+        Self::RenderCommand(t.into())
+    }
+}
+
+/// Error encountered when finishing recording a render bundle.
+#[derive(Clone, Debug, Error)]
+#[error("{scope}")]
+pub struct RenderBundleError {
+    scope: PassErrorScope,
+    #[source]
+    inner: RenderBundleErrorInner,
+}
+
+impl<T, E> MapPassErr<T, RenderBundleError> for Result<T, E>
+where
+    E: Into<RenderBundleErrorInner>,
+{
+    fn map_pass_err(self, scope: PassErrorScope) -> Result<T, RenderBundleError> {
+        self.map_err(|inner| RenderBundleError {
+            scope,
+            inner: inner.into(),
+        })
+    }
 }
 
 impl<G: GlobalIdentityHandlerFactory> Global<G> {
@@ -675,13 +708,16 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         id_in: Input<G, id::RenderBundleId>,
     ) -> Result<id::RenderBundleId, RenderBundleError> {
         span!(_guard, INFO, "RenderBundleEncoder::finish");
+        let scope = PassErrorScope::Bundle;
+
         let hub = B::hub(self);
         let mut token = Token::root();
         let (device_guard, mut token) = hub.devices.read(&mut token);
 
         let device = device_guard
             .get(bundle_encoder.parent_id)
-            .map_err(|_| DeviceError::Invalid)?;
+            .map_err(|_| DeviceError::Invalid)
+            .map_pass_err(scope)?;
         let render_bundle = {
             let (pipeline_layout_guard, mut token) = hub.pipeline_layouts.read(&mut token);
             let (bind_group_guard, mut token) = hub.bind_groups.read(&mut token);
@@ -712,12 +748,15 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                         num_dynamic_offsets,
                         bind_group_id,
                     } => {
+                        let scope = PassErrorScope::SetBindGroup;
+
                         let max_bind_groups = device.limits.max_bind_groups;
                         if (index as u32) >= max_bind_groups {
-                            Err(RenderCommandError::BindGroupIndexOutOfRange {
+                            return Err(RenderCommandError::BindGroupIndexOutOfRange {
                                 index,
                                 max: max_bind_groups,
-                            })?
+                            })
+                            .map_pass_err(scope);
                         }
 
                         let offsets = &base.dynamic_offsets[..num_dynamic_offsets as usize];
@@ -729,25 +768,32 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                             .map(|offset| *offset as wgt::BufferAddress)
                             .find(|offset| offset % wgt::BIND_BUFFER_ALIGNMENT != 0)
                         {
-                            Err(RenderCommandError::UnalignedBufferOffset(offset))?
+                            return Err(RenderCommandError::UnalignedBufferOffset(offset))
+                                .map_pass_err(scope);
                         }
 
                         let bind_group = state
                             .trackers
                             .bind_groups
                             .use_extend(&*bind_group_guard, bind_group_id, (), ())
-                            .map_err(|_| RenderCommandError::InvalidBindGroup(bind_group_id))?;
+                            .map_err(|_| RenderCommandError::InvalidBindGroup(bind_group_id))
+                            .map_pass_err(scope)?;
                         if bind_group.dynamic_binding_info.len() != offsets.len() {
-                            Err(RenderCommandError::InvalidDynamicOffsetCount {
+                            return Err(RenderCommandError::InvalidDynamicOffsetCount {
                                 actual: offsets.len(),
                                 expected: bind_group.dynamic_binding_info.len(),
-                            })?
+                            })
+                            .map_pass_err(scope);
                         }
 
                         state.set_bind_group(index, bind_group_id, bind_group.layout_id, offsets);
-                        state.trackers.merge_extend(&bind_group.used)?;
+                        state
+                            .trackers
+                            .merge_extend(&bind_group.used)
+                            .map_pass_err(scope)?;
                     }
                     RenderCommand::SetPipeline(pipeline_id) => {
+                        let scope = PassErrorScope::SetPipeline;
                         if state.pipeline.set_and_check_redundant(pipeline_id) {
                             continue;
                         }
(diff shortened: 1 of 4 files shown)


Kimundi:
I updated the PR to add error context to all commands in a RednerPass/ComputePass/RenderBundle, and it got a bit verbose. 

There are two potential, mutual exclusive, courses of action I can take to make the code a bit nicer:

- I can replace the `with` scope wrappers with a simpler system that selects the command context at runtime via a single additional register-sized variable.
- I can move some of the error sources in the `[...]ErrorInner` back into the main `[...]Error` struct to need less `map_err` calls outside of the command `match` blocks.

Kimundi:
I decided to push the variant that sets the error contexts dynamically, as it has a more isolated impact on the code

wgpu-core/src/command/bundle.rs
```diff
@@ -667,13 +667,81 @@ pub enum RenderBundleError {
     Draw(#[from] DrawError),
 }
 
+/// Error encountered when performing a render pass.
+#[derive(Clone, Debug, Error)]
+pub enum RenderBundleError {
+    #[error(transparent)]
+    Inner(#[from] RenderBundleErrorInner),
```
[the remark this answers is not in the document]
Kimundi: Honestly I just did it like this to make use of `thiserror`s `#[error]` annotations in a simple way, as I was under the impression that I would have to write the Error implementations manually if I make it a regular struct - I might be wrong about that though. In any case I thought that it doesnt matter much as it won't take up more or less space in the types memory layout anyway.

wgpu-core/src/command/bundle.rs
```diff
@@ -667,13 +667,81 @@ pub enum RenderBundleError {
...
+    DrawIndexed(#[source] RenderBundleErrorInner),
+    #[error("In a indirect draw command")]
+    DrawIndirect(#[source] RenderBundleErrorInner),
+    #[error("In a indexed indirect draw command")]
+    DrawIndexedIndirect(#[source] RenderBundleErrorInner),
+}
+
+type RenderBundleErrorCtx = fn(RenderBundleErrorInner) -> RenderBundleError;
+
+fn set_error_ctx(command: &RenderCommand, err_ctx: &mut RenderBundleErrorCtx) {
```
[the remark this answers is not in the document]
Kimundi: Ah sure, that would be a trivial change

wgpu-core/src/command/bundle.rs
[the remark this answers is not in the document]
Kimundi: Actually nevermind that, just re-read the docs of thiserror, you can use it with structs as well of course. Then that leaves just the issue that I'm not sure it would be relevantly shorter if I re-write as a struct (and it would also lose the functionality of the `Inner` variant, which does _not_ attach an error context and as such can attach the error messages direct to the RenderCommand itself)

I'll prototype it though, lets see how it will look.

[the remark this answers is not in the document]
Kimundi: How about this?
```rust

fn error_context(command: ComputeCommand) -> ComputePassCommandError {
    match command {
        ComputeCommand::SetBindGroup { .. } => ComputePassCommandError::SetBindGroup,
        ComputeCommand::SetPipeline(_) => ComputePassCommandError::SetPipeline,
        ComputeCommand::SetPushConstant { .. } => ComputePassCommandError::SetPushConstant,
        ComputeCommand::Dispatch(_) => ComputePassCommandError::Dispatch,
        ComputeCommand::DispatchIndirect { .. } => ComputePassCommandError::DispatchIndirect,
        ComputeCommand::PushDebugGroup { .. }
        | ComputeCommand::PopDebugGroup
        | ComputeCommand::InsertDebugMarker { .. } => ComputePassCommandError::Debug,
    }
}

/// Error encountered when performing a compute pass.
#[error("{command}")]
pub struct ComputePassError {
    command: ComputePassCommandError,
    #[source]
    inner: ComputePassErrorInner,
}

pub enum ComputePassCommandError {
    #[error("In the pass parameters")]
    Pass,
    #[error("In a set_bind_group command")]
    SetBindGroup,
    #[error("In a set_pipeline command")]
    SetPipeline,
    #[error("In a set_push_constant command")]
    SetPushConstant,
    #[error("In a dispatch command")]
    Dispatch,
    #[error("In a indirect dispatch command")]
    DispatchIndirect,
    #[error("In a debug command")]
    Debug,
}
```

I could store the `RenderCommand` directly, but then I would still need a match to map to the error messages - it would remove the enum I suposed, but it seems kinda unclean to carry the payload of the commands around if we only care about the descriminant.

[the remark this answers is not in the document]
Kimundi: I could also make it shorter by storing the error message as a field:

```rust
type ComputePassCommandError = Option<&'static str>;

fn error_context(command: ComputeCommand) -> ComputePassCommandError {
    Some(match command {
        ComputeCommand::SetBindGroup { .. } => "In a set_bind_group command",
        ComputeCommand::SetPipeline(_) => "In a set_pipeline command",
        ComputeCommand::SetPushConstant { .. } => "In a set_push_constant command",
        ComputeCommand::Dispatch(_) => "In a dispatch command",
        ComputeCommand::DispatchIndirect { .. } => "In a indirect dispatch command",
        ComputeCommand::PushDebugGroup { .. }
        | ComputeCommand::PopDebugGroup
        | ComputeCommand::InsertDebugMarker { .. } => "In a debug command",
    })
}

/// Error encountered when performing a compute pass.
#[error("{command}")]
pub struct ComputePassError {
    command: &'static str,
    #[source]
    inner: ComputePassErrorInner,
}

impl ComputePassError {
    fn new(command: ComputePassCommandError, inner: ComputePassErrorInner) -> Self {
        ComputePassError {
            command: command.unwrap_or("In the pass parameters"),
            inner,
        }
    }
}
```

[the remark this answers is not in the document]
kvark: The last one looks good to me, thank you for experimenting!
I was thinking that just debug-printing `RenderCommand` would be simple and effective, too, since it would avoid the need to remap it into anything. But your last `error_context` looks simple enough.

Kimundi:
Alright, I updated everything to the string-based error contexts and squashed it, should be good now. :)

wgpu-core/src/command/bundle.rs
```diff
@@ -667,13 +669,45 @@ pub enum RenderBundleError {
...
+        self.render_bundle_encoder_finish_inner::<B>(bundle_encoder, desc, id_in, &mut command)
+            .map_err(|inner| RenderBundleError::new(command, inner))
+    }
+
+    fn render_bundle_encoder_finish_inner<B: GfxBackend>(
+        &self,
+        bundle_encoder: RenderBundleEncoder,
+        desc: &RenderBundleDescriptor,
+        id_in: Input<G, id::RenderBundleId>,
+        err_ctx: &mut RenderBundleCommandError,
```
kvark, line 709: Can we get away from this mutable semantic entirely? Generally we should try to keep this "functional style". Mixing the procedural and function in the same semantics (error passing) is awkward, and would impede maintenance of this code.

Can we instead just have `-> Result<id::RenderBundleId, RenderBundleError>`? whenever there is a spot that returns an error, we'd just map that error and insert the appropriate command description in place, i.e.

```rust
something()
  .map_err(|error| RenderBundleError {
    command: "me does things wrong",
    inner: error.into(),
  })?

Comment (Kimundi) [reply wgpu-core/src/command/bundle.rs:709]: I did something like that initially, but it got very verbose and error prone in my opinion, which is why I went with this blanked approach. 

I can try changing the mutable context setting to a more declarative style, but it probably would make a bigger diff as I need to capture all subcommands in a closure.

Comment (kvark) [reply wgpu-core/src/command/bundle.rs:709]: Could you show how it looks for one small thing, e.g. compute errors? Just to get the idea

File: wgpu-core/src/command/bundle.rs [hunk 4]
```diff
@@ -656,17 +656,45 @@ impl State {
...
+/// Error encountered when finishing recording a render bundle.
+#[derive(Clone, Debug, Error)]
+#[error("{command}")]
+pub struct RenderBundleError {
+    command: &'static str,
+    #[source]
+    inner: RenderBundleErrorInner,
+}
+
+fn pass_err<E>(command: &'static str) -> impl FnOnce(E) -> RenderBundleError
```
kvark, line 688: would we be able to copy the result if it's `impl Fn(E) -> ...` instead of `FnOnce`?
That would allow us to do `let pass_ctx = pass_err("In a bundle parameter");` right away and avoid repeating the `pass_err` calls through arms of the match

wgpu-core/src/command/bundle.rs
```diff
@@ -712,12 +743,15 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                         num_dynamic_offsets,
                         bind_group_id,
                     } => {
+                        let err_ctx = "In a set_bind_group command";
+
                         let max_bind_groups = device.limits.max_bind_groups;
                         if (index as u32) >= max_bind_groups {
                             Err(RenderCommandError::BindGroupIndexOutOfRange {
```
kvark, line 750: can we be smarter in these cases? i.e.
```rust
Err(pass_err(err_ctx)(RenderCommandError::{..}.into()))?
```
Otherwise, it's a bit strange to construct an `Err` only to follow-up with `map_err`...

kvark (commented):
Thank you!
I think this looks shippable. Just a few final notes, plus a squash is needed.

wgpu-core/src/command/bundle.rs
Kimundi: I have thought about that, but that doesn't actually solve the issue because different uses of this function want to instantiate the closure with different `E` types.

wgpu-core/src/command/bundle.rs
Kimundi: Yes that would be possible, I just did a blanked pass with the map functions for now, but I can clean up those cases.

wgpu-core/src/command/bundle.rs
```diff
@@ -656,17 +656,47 @@ impl State {
 
 /// Error encountered when finishing recording a render bundle.
 #[derive(Clone, Debug, Error)]
-pub enum RenderBundleError {
+pub enum RenderBundleErrorInner {
```
kvark, line 659: now that we are returning a different errors, do we still need all the `Inner` ones to be public?

wgpu-core/src/command/bundle.rs
Kimundi: With the `map_pass_err`, this is a moot point now.

wgpu-core/src/command/bundle.rs
Kimundi: AH, indeed not.

Kimundi: Actually, due to the trait impl it would have to be `pub(super)` at the least.

kvark: sure, `pub(super)` is fine

wgpu-core/src/command/bundle.rs
```diff
@@ -678,9 +681,9 @@ where
 
 /// Error encountered when finishing recording a render bundle.
 #[derive(Clone, Debug, Error)]
-#[error("{command}")]
+#[error("{ctx}")]
 pub struct RenderBundleError {
-    command: &'static str,
+    ctx: PassErrorCtx,
```
kvark, line 686: maybe we should call it `scope` with `ErrorScope` instead?

kvark (commented):
I think this is an improvement!

wgpu-core/src/command/bundle.rs
Kimundi: Sure!

wgpu-core/src/command/bundle.rs
```diff
@@ -675,13 +708,16 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         id_in: Input<G, id::RenderBundleId>,
     ) -> Result<id::RenderBundleId, RenderBundleError> {
         span!(_guard, INFO, "RenderBundleEncoder::finish");
+        let err_ctx = PassErrorScope::Bundle;
```
kvark, line 711: can we mass-rename this to `err_scope` or just `scope`?

Kimundi: Ah sure, that would make sense in combination with the other change :)

kvark (approved):
Alright, looks like you beat the final boss!
Thank you for all the patience :)
bors r+

Issue #978, opened 2020-10-13
Make wgpu_render_pass_set_index_buffer to accept an IndexFormat


kvark:
This would match the latest upstream API changes of https://github.com/gpuweb/gpuweb/pull/943

kvark:
Issue is currently blocked on WebGPU discussions...

grovesNL:
Fixed by #1019

PR #1019, opened 2020-11-10
Match upstream API changes for IndexFormat

kejor:
**Connections**
#978 

**Description**
Match upstream API changes for IndexFormat.

- `index_format` from `VertexStateDescriptor` is now a `Option`
- `RenderCommand::SetIndexBuffer` now has a `index_format` attribute
- The index format for execution is now set when matched both `RenderCommand::SetPipeline` and `RenderCommand::SetIndexBuffer`.

**Testing**
Tested with the examples of https://github.com/kejor/wgpu-rs/commit/e5f82a21972842806774629534ab97051043489d

---

Should I implement this?https://github.com/gpuweb/gpuweb/blob/0855ad9b97d4fa6a8cb4d51740675a21044f099c/spec/index.bs#L3703

### DIFF
--- a/player/tests/data/quad.ron
+++ b/player/tests/data/quad.ron
@@ -86,7 +86,7 @@
             ],
             depth_stencil_state: None,
             vertex_state: (
-                index_format: Uint16,
+                index_format: None,
                 vertex_buffers: [],
             ),
             sample_count: 1,
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -245,6 +245,7 @@ impl RenderBundleEncoder {
                 }
                 RenderCommand::SetIndexBuffer {
                     buffer_id,
+                    index_format,
                     offset,
                     size,
                 } => {
@@ -261,6 +262,7 @@ impl RenderBundleEncoder {
                         Some(s) => offset + s.get(),
                         None => buffer.size,
                     };
+                    state.index.set_format(index_format);
                     state.index.set_buffer(buffer_id, offset..end);
                 }
                 RenderCommand::SetVertexBuffer {
@@ -520,14 +522,17 @@ impl RenderBundle {
                 RenderCommand::SetPipeline(pipeline_id) => {
                     let pipeline = pipeline_guard.get(pipeline_id).unwrap();
                     cmd_buf.bind_graphics_pipeline(&pipeline.raw);
-                    index_type = conv::map_index_format(pipeline.index_format);
+
                     pipeline_layout_id = Some(pipeline.layout_id.value);
                 }
                 RenderCommand::SetIndexBuffer {
                     buffer_id,
+                    index_format,
                     offset,
                     size,
                 } => {
+                    index_type = conv::map_index_format(index_format);
+
                     let &(ref buffer, _) = buffer_guard
                         .get(buffer_id)
                         .unwrap()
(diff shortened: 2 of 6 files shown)


player/tests/data/quad.ron
```diff
@@ -86,7 +86,7 @@
             ],
             depth_stencil_state: None,
             vertex_state: (
-                index_format: Uint16,
+                index_format: Some(Uint16),
```
kvark, line 89: this should actually be `None`. We only need to specify this for strip topologies

wgpu-core/src/command/render.rs
```diff
@@ -1172,6 +1176,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                     state.index.bound_buffer_view = Some((id::Valid(buffer_id), offset..end));
                     state.index.update_limit();
 
+                    state.index.format = index_format;
```
kvark, line 1179: We actually need to track both the `Option<IndexFormat>` from the current pipeline and the current index format of the active `SetIndexBuffer`. Before issuing a draw call there is an `is_ready` function that needs to check for:
  - if the pipeline index format is set, it has to be the same as the current index format, otherwise an error is generated

wgpu-core/src/command/bundle.rs
```diff
@@ -197,14 +197,20 @@ impl RenderBundle {
...
+
                     pipeline_layout_id = Some(pipeline.layout_id.value);
                 }
                 RenderCommand::SetIndexBuffer {
                     buffer_id,
+                    index_format,
                     offset,
                     size,
                 } => {
+                    index_type = conv::map_index_format(index_format);
```
kvark, line 213: we also need to call `state.index.update_limit();` again if this format is changed here

wgpu-types/src/lib.rs
```diff
@@ -944,8 +944,7 @@ impl DepthStencilStateDescriptor {
 /// Format of indices used with pipeline.
 #[repr(C)]
 #[derive(Copy, Clone, Debug, Hash, Eq, PartialEq)]
-#[cfg_attr(feature = "trace", derive(Serialize))]
-#[cfg_attr(feature = "replay", derive(Deserialize))]
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
```
kvark, line 947: Curious why this change was needed?

kvark (commented):
Thank you for taking a stab at this! There is a few things we need to fix before landing. Also, please squash the commits as our merger bot doesn't know how to do it.

kvark:
We are in a situation where the group may back-track from the decision to have the index format here at all. Let's hold this off until the group figures out what to do. Sorry about putting you in the middle of this!

kvark:
I did some experiments in https://github.com/gpuweb/gpuweb/issues/1220#issuecomment-732357591, discussed it with the group, and it sounds like we can proceed with the API as it today. This PR is therefore unlocked. Please rebase!

kvark:
@kejor would you be interested in finishing this? Sorry about blocking progress here previously!

kejor:
> @kejor would you be interested in finishing this? Sorry about blocking progress here previously!

Yeah, no problemo. I was about to get started haha

wgpu-types/src/lib.rs
kejor: There was one or two checks complaining about it. I don't quite remember which ones.

wgpu-core/src/command/bundle.rs
kejor: `execute` in `RenderBundle` doesn't really have a state, should it?

kvark (approved):
I think it's all good now, thank you for finishing this!
bors r+

kvark:
@kejor would you be able to make a `wgpu-rs` update that adjusts for these API changes?

kejor:
> @kejor would you be able to make a `wgpu-rs` update that adjusts for these API changes?

Yes, I'll be making a PR with the update.

PR #1025, opened 2020-11-15
Don't install Android NDK

kvark:
Follows https://github.com/gfx-rs/gfx/pull/3474


kvark:
@msiglreith you said Android NDK is installed. I wonder why it's not in the PATH then?

msiglreith:
see https://github.com/gfx-rs/wgpu/pull/1029

kvark:
bors r=msiglreith

PR #1031, opened 2020-11-18
Update gpu-alloc and naga

kvark:
**Connections**
includes critical https://github.com/zakarumych/gpu-alloc/pull/24

**Description**
Updates Naga and gpu-alloc deps

**Testing**
Tested on wgpu-rs


kvark:
bors r=straightforward

PR #1036, opened 2020-11-21
Stop using Borrow<RefCount>

kvark:
**Connections**
Follow-up to #931

**Description**
Takes advantage of the new `trait Resource` in the state tracker, reducing the code.

**Testing**
Not tested outside of `cargo test`


scoopr:
r+

kvark:
Bors r=scoopr

PR #1037, opened 2020-11-21
Add Ids to PassErrorScope

scoopr:
Also makes it pub, and the places it is used.
Tweak some error messages.

### DIFF
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -680,7 +680,7 @@ where
 #[derive(Clone, Debug, Error)]
 #[error("{scope}")]
 pub struct RenderBundleError {
-    scope: PassErrorScope,
+    pub scope: PassErrorScope,
     #[source]
     inner: RenderBundleErrorInner,
 }
@@ -745,7 +745,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                         num_dynamic_offsets,
                         bind_group_id,
                     } => {
-                        let scope = PassErrorScope::SetBindGroup;
+                        let scope = PassErrorScope::SetBindGroup(bind_group_id);
 
                         let max_bind_groups = device.limits.max_bind_groups;
                         if (index as u32) >= max_bind_groups {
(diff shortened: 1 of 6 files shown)


wgpu-core/src/command/bundle.rs
```diff
@@ -745,7 +745,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                         num_dynamic_offsets,
                         bind_group_id,
                     } => {
-                        let scope = PassErrorScope::SetBindGroup;
+                        let scope = PassErrorScope::SetBindGroup(bind_group_id);
```
kvark, line 748: if we go this route, wouldn't it be easier to just have the full `command` cloned into the error?

wgpu-core/src/command/bundle.rs
```diff
@@ -680,7 +680,7 @@ where
 #[derive(Clone, Debug, Error)]
 #[error("{scope}")]
 pub struct RenderBundleError {
-    scope: PassErrorScope,
+    pub scope: PassErrorScope,
```
kvark, line 683: why does it need to be public?

wgpu-core/src/command/bundle.rs
scoopr: Maybe. I don’t currently see a reason though, what else useful information would be had, that isn’t then also displayed in the more specific error?

I don’t see the implementation to be any simpler the either.

wgpu-core/src/command/bundle.rs
scoopr: I need to access it from the wgpu-rs side, as it is information kind of in the side, with the Ids that need to be resolved

wgpu-core/src/command/bundle.rs
kvark: The main benefit is not having a separate `PassErrorScope` enum at all. If you don't think that's a big one, I'm fine with proceeding with your plan :)

kvark (approved):
bors r+

PR #1047, opened 2020-11-26
Update bind group layout API to match upstream

kvark:
**Connections**
Follows https://github.com/gpuweb/gpuweb/pull/1076, https://github.com/gpuweb/gpuweb/pull/1223 (https://github.com/gpuweb/gpuweb/issues/1164), https://github.com/gpuweb/gpuweb/pull/1255, and https://github.com/gpuweb/gpuweb/pull/1256

**Description**
Aligns our API closer to the latest changes in WebGPU upstream. We technically don't have to do this, but I believe in the end it would be best if our API gets close to upstream.

Note: this is a sensitive change for the users, everybody will get their code broken. So please take a look at the API and see if something is missing or needs improvement, so that we don't have to go through the changes again afterwards.

**Testing**
Doesn't really need testing. Partially covered by the existing playtest.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -447,6 +447,12 @@ pub enum TextureViewDimension {
     D3,
 }
 
+impl Default for TextureViewDimension {
+    fn default() -> Self {
+        Self::D2
+    }
+}
+
 impl TextureViewDimension {
     /// Get the texture dimension required fo this texture view dimension.
     pub fn compatible_texture_dimension(self) -> TextureDimension {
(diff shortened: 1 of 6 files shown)


wgpu-types/src/lib.rs
```diff
@@ -1873,9 +1965,10 @@ pub struct BindGroupLayoutEntry {
     pub ty: BindingType,
     /// If this value is Some, indicates this entry is an array. Array size must be 1 or greater.
     ///
-    /// If this value is Some and `ty` is `BindingType::SampledTexture`, [`Capabilities::SAMPLED_TEXTURE_BINDING_ARRAY`] must be supported.
+    /// If this value is Some and `ty` is `BindingType::Texture`, [`Capabilities::SAMPLED_TEXTURE_BINDING_ARRAY`] must be supported.
```
cwfitzgerald, line 1968: nit, holdover from before:
```suggestion
    /// If this value is Some and `ty` is `BindingType::Texture`, [`Features::SAMPLED_TEXTURE_BINDING_ARRAY`] must be supported.
```

cwfitzgerald (approved):
Codewise this looks fine outside of a nit. If possible, I would like to see one of the examples converted to this new style to see how it looks. This looks like how gfx-hal does it, so it should be fine, just want to check on it.

kvark:
This is how it looks like (WIP) - https://github.com/gfx-rs/wgpu-rs/pull/649
I also noticed that the `TextureComponentType` needed to be removed, so I did it and ported over the conversion from formats to sample types.

kvark:
To minimize disruption, let's block this on https://github.com/gpuweb/gpuweb/pull/1256 for now

cwfitzgerald:
Sounds good to wait. The api looks good from the user perspective.

kvark:
bors r=cwfitzgerald

Issue #836, opened 2020-07-24
Make VertexFormat::size() a const fn when rust 1.46 drops


bootra-dev:
Original issue: https://github.com/gfx-rs/wgpu/issues/801 

When I implemented https://github.com/gfx-rs/wgpu/pull/802 I did not consider that the vertex_attr_array macro could not longer be used in a const environment. Rust 1.46 will allow match in const functions, so this can be resolved then.

cwfitzgerald:
Related #780

cwfitzgerald:
Solved by #1048

PR #1048, opened 2020-11-26
Make VertexFormat::size a const fn

Imberflur:
**Connections**
Addresses #836 

**Description**
The `vertex_attr_array!` macro uses this function. This change enables the macro to be used in `const` contexts and for instance allows returning a `VertexBufferDescriptor` constructed in function without having the difficulties of ensuring the array from `vertex_attr_array!` lives long enough.

**Testing**
I don't believe testing is needed for this.


kvark (approved):
thank you!
bors r+

Issue #1027, opened 2020-11-15
Switch to gfx-rs master


kvark:
**Is your feature request related to a problem? Please describe.**
Patching gfx-rs issues for wgpu involves too many steps today:
  1. making a change
  2. making PR to gfx-rs `master`
  3. making PR to gfx-rs `hal-0.6` branch
  4. merging PRs
  5. publishing the crate updates
  6. updating dependencies

**Describe the solution you'd like**
Switch to master, so that patching gfx-rs is just: 1-2-4-6, without the need to duplicate/rebase the PR and publish it.

**Describe alternatives you've considered**
Keep doing what we are doing.

**Additional context**
We have to go through gfx-rs on crates because we are using `gfx-descriptor` (and used to depend on `gfx-memory`) as intermediate crates. Replacing `gfx-descriptor` by something that's not linked directly to gfx-rs would unblock this issue.

PR #1056, opened 2020-11-30
Replace gfx-descriptor by gpu-descriptor

kvark:
**Connections**
Unblocks #1027
cc @zakarumych 

**Description**
Removes the last link we had between us and gfx-rs.

**Testing**
Tested on wgpu-rs

### DIFF
--- a/wgpu-core/src/device/descriptor.rs
+++ b/wgpu-core/src/device/descriptor.rs
@@ -1,0 +1,168 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use super::DeviceError;
+use arrayvec::ArrayVec;
+
+pub use gpu_descriptor::DescriptorTotalCount;
+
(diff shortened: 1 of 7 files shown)


wgpu-core/src/device/descriptor.rs
```diff
@@ -0,0 +1,168 @@
...
+            },
+            count: descriptor_count.storage_buffer_dynamic as _,
+        });
+        ranges.retain(|rd| rd.count != 0);
+
+        match hal::device::Device::create_descriptor_pool(
+            self.0,
+            max_sets as usize,
+            ranges,
+            hal::pso::DescriptorPoolCreateFlags::from_bits_truncate(flags.bits() as u32),
```
zakarumych, line 124: Even if bits probably will be the same as both gfx and gpu-descriptor tend to use same bits as Vulkan,
it is not guaranteed.

kvark: Can we guarantee that? I mean, at least specify in documentation of gpu-alloc and gpu-descriptor that the bits match Vulkan.

zakarumych: Done

cwfitzgerald (approved):
Nothing looks out of place! Very excited to get this going!

kvark:
bors r=cwfitzgerald

PR #1068, opened 2020-12-02
Fix Display for pass errors

kvark:
**Connections**
Currently printing an error only outputs the scope.

**Description**
This PR makes it print the whole error.

**Testing**
Untested

### DIFF
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -500,7 +500,6 @@ impl RenderBundle {
         use hal::command::CommandBuffer as _;
 
         let mut offsets = self.base.dynamic_offsets.as_slice();
-        let mut index_type = hal::IndexType::U16;
         let mut pipeline_layout_id = None::<id::Valid<id::PipelineLayoutId>>;
 
         for command in self.base.commands.iter() {
@@ -531,7 +530,7 @@ impl RenderBundle {
                     offset,
                     size,
                 } => {
-                    index_type = conv::map_index_format(index_format);
+                    let index_type = conv::map_index_format(index_format);
 
                     let &(ref buffer, _) = buffer_guard
                         .get(buffer_id)
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -1007,7 +1007,7 @@ where
 
 /// Error encountered when finishing recording a render bundle.
 #[derive(Clone, Debug, Error)]
-#[error("{scope}")]
+#[error("Render bundle error {scope}: {inner}")]
 pub struct RenderBundleError {
     pub scope: PassErrorScope,
     #[source]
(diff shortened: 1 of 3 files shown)


kvark:
@scoopr I was doing `format!("{}", err)`. Am I missing anything? Could you review this PR?

scoopr:
I think @Kimundi did that originally, did I mess it up while doing the other changes? 

Shouldn’t the `inner` be a source error, hence shouldn’t need to be repeated in the outer error message?

kvark:
Sorry, I might have confused the authorship.
> Shouldn’t the inner be a source error, hence shouldn’t need to be repeated in the outer error message?

Well, that sounds logical, but how do I get the *full* error then?

scoopr:
> Well, that sounds logical, but how do I get the _full_ error then?

That’s what `format_error` in wgpu-rs does, are you looking at this from another consumers perspective?

kvark:
Yeah

kvark:
I guess another option here is to have `format_error` at `wgpu` level, so that non-wgpu-rs clients can take advantage of it

scoopr:
> I guess another option here is to have `format_error` at `wgpu` level, so that non-wgpu-rs clients can take advantage of it

Well yes, you kind of pushed the error reporting to be moved to wgpu-rs level, this includes the id to label resolution, formatting and also some of the top level errors are injected there.

I don't think there is anything in the way of moving all (or some) of that back to wgpu.

Maybe `format_error` implementation could be moved just by making it accept the `Global`, or implement it directly on `Global`, so it can do the label resolution. But it is also doing the label resolution for the `ContextError` that is defined in `wgpu-rs`, so either that needs to be moved too or some method of "injecting" more supported types.

kvark:
> Well yes, you kind of pushed the error reporting to be moved to wgpu-rs level, this includes the id to label resolution, formatting and also some of the top level errors are injected there.

Yes, I remember. I did a similar thing to errors in general at first, wanting `wgpu-core` to just return `Result` and the clients to deal with the "error IDs". Then I had to rewrite `wgpu-core` to handle that more transparently for all the clients.

So for `format_error`, we need something like this for Servo and Gecko, so we might as well move it in wgpu-core now. Sorry about making the wrong call previously!

kvark:
Is there any harm/downside to landing this? Would it make wgpu-rs errors wrong? cc @Kimundi

Kimundi:
It would change the wgpu-rs error output from something like
```
Caused by:
    In a RenderPass
    In a set_bind_group command
      note: bind group = `<BindGroup-(0, 1, Vulkan)>`    
    bind group index 9 is greater than the device's requested `max_bind_group` limit 4
```

To something like this:

```
Caused by:
    In a RenderPass
    Render pass error In a set_bind_group command: bind group index 9 is greater than the device's requested `max_bind_group` limit 4
      note: bind group = `<BindGroup-(0, 1, Vulkan)>`    
    bind group index 9 is greater than the device's requested `max_bind_group` limit 4
```

...which does not look quite right :D

kvark:
Ok, it sounds like we need to get back to the board and try to integrate the pretty errors into wgpu itself.
Is this something @Kimundi feels generous to help with?

Issue #1069, opened 2020-12-03
Tracking Issue for Wgpu on Raspberry Pi 4


cwfitzgerald:
**Description**

Summary of all current issues and todos for a seamless raspbery pi 4 experience. Most of these will be driver issues and will also be found on https://github.com/gfx-rs/wgpu/wiki/Known-Driver-Issues. Links to upstream issues will be added as they are filed.

Unless listed otherwise, this was tested on mesa commit `a1976e1cb2d817a4d5229994b83c87d82485af9b`.

## Shadow

There is sporatic corruption of the shadows that pops in weird ways. Video here: https://cwfitz.com/zFpfee.mp4.

Upstream bug filed: https://gitlab.freedesktop.org/mesa/mesa/-/issues/3916

## Water

Water works fine until you resize it, then the water looks to have some sort of block-based corruption. If you get lucky when you resize it'll go back to normal. This might be caused by https://github.com/gfx-rs/wgpu-rs/issues/665 as the issue doesn't show up on a pre gfx-master version of wgpu.

## Cube

Cube works correctly most of the time, but random frames will have the vertex order reversed so the lines go the other way through the square faces.

andrewdavidmackenzie:
I am currently (master, Dec 15 2022) on Pi4 seeing:

'No suitable GPU adapters found on the system!'

Am I right in assuming this is the correct open issue to track?

cwfitzgerald:
It is, but if you have a vulkan implementation it should at least find an adapter.

Lokathor:
`sudo apt install vulkan-tools mesa-vulkan-drivers` will now install the vulkan drivers on raspberry pi. If you run a vulkan using program it'll print a message as it loads that technically the driver isn't fully conformant and it's for testing only, but the program will render.

I didn't test any of the examples from this repo specifically (i was checking out the `minimal-winit` example from `pixels`, which uses `wgpu`). If you want I can possibly check more soon.

anselanza:
Indeed, `pixels` does work on the Raspberry Pi 4 as mentioned, after installing the Vulkan V3D driver, but I cannot get the built-in examples from WGPU to run.

The following error is typical:
```
WARNING: v3dv is neither a complete nor a conformant Vulkan implementation. Testing use only.
Using V3D 4.2 (Vulkan)
[2023-05-25T13:39:35Z ERROR wgpu::backend::direct] Error in Adapter::request_device: Not enough memory left
[...]

PR #1071, opened 2020-12-03
Add pervasive Naga support to shader module loading

kvark:
**Connections**
Depends on https://github.com/gfx-rs/naga/pull/299 and https://github.com/gfx-rs/gfx/pull/3503
Closes #1072

**Description**
    The purpose of the PR is to support Naga modules everywhere.
    As a requirement, it updates the gfx-rs version used.
    Most of the logic is dedicated towards building a shader interface,
    where previously we just used naga's IR. Now we have our own mini-IR.

**Testing**
Tested in https://github.com/gfx-rs/wgpu-rs/pull/663 (without the experimental feature).
Enabling the experimental feature doesn't work even on the play test that we have - https://github.com/gfx-rs/naga/issues/300

PR #1074, opened 2020-12-04
Implement ETC2 and ASTC textures

### LINKED ISSUE #1070: ETC and ASTC compression

Compressed textures are very important for running on mobile chips due to their very low memory bandwidth. BCn is generally only supported on desktop gpus due to patent issues. Modern android chips support both ETC, ETC2, EAC and ASTC. Additionally the rpi4 only supports ETC2.

Support for these texture formats.

None

**Additional context**

Intel is a very neat test bed as it supports BCn, ETC2, and ASTC so it's easy to test them all on a non-mobile platform.

cwfitzgerald:
**Connections**

Closes #1070. Makes progress towards #1069.

**Description**

This PR has multiple functions:
 - Adds ETC and ASTC compressed textures behind features.
 - Adds three helper functions on `Extent3d` that help in calculating mip sizes: `at_mip_level`, `max_mips`, and `physical_size`.
 - Refactors various conversions into a public `TextureFormat::describe` function. I have used a decl macro to ease in the declaration and modifcation of an otherwise horribly verbose function.

I have tried to use clever multi-select based data copying to reduce the possibility for errors, but there's a _lot_ of data moving around here.

**Testing**

Upcoming wgpu-rs pr adding wider compressed texture support to the skybox example. Helper functions were tested with doctests.

Marked as draft until I can get the wgpu-rs pr done and prove it works, it is, however, ready for review.

### DIFF
--- a/wgpu-core/src/conv.rs
+++ b/wgpu-core/src/conv.rs
@@ -396,124 +396,48 @@ pub(crate) fn map_texture_format(
         Tf::Bc6hRgbUfloat => H::Bc6hUfloat,
         Tf::Bc7RgbaUnorm => H::Bc7Unorm,
         Tf::Bc7RgbaUnormSrgb => H::Bc7Srgb,
-    }
-}
 
-pub fn texture_block_size(format: wgt::TextureFormat) -> (u32, u32) {
-    use wgt::TextureFormat as Tf;
-    match format {
-        Tf::R8Unorm
-        | Tf::R8Snorm
-        | Tf::R8Uint
-        | Tf::R8Sint
-        | Tf::R16Uint
-        | Tf::R16Sint
-        | Tf::R16Float
-        | Tf::Rg8Unorm
-        | Tf::Rg8Snorm
-        | Tf::Rg8Uint
-        | Tf::Rg8Sint
-        | Tf::R32Uint
-        | Tf::R32Sint
-        | Tf::R32Float
-        | Tf::Rg16Uint
-        | Tf::Rg16Sint
-        | Tf::Rg16Float
-        | Tf::Rgba8Unorm
-        | Tf::Rgba8UnormSrgb
-        | Tf::Rgba8Snorm
-        | Tf::Rgba8Uint
-        | Tf::Rgba8Sint
-        | Tf::Bgra8Unorm
-        | Tf::Bgra8UnormSrgb
-        | Tf::Rgb10a2Unorm
-        | Tf::Rg11b10Float
-        | Tf::Rg32Uint
-        | Tf::Rg32Sint
-        | Tf::Rg32Float
-        | Tf::Rgba16Uint
-        | Tf::Rgba16Sint
-        | Tf::Rgba16Float
-        | Tf::Rgba32Uint
-        | Tf::Rgba32Sint
-        | Tf::Rgba32Float
-        | Tf::Depth32Float
-        | Tf::Depth24Plus
-        | Tf::Depth24PlusStencil8 => (1, 1),
+        // ETC compressed formats
+        Tf::Etc2RgbUnorm => H::Etc2R8g8b8Unorm,
+        Tf::Etc2RgbUnormSrgb => H::Etc2R8g8b8Srgb,
+        Tf::Etc2RgbA1Unorm => H::Etc2R8g8b8a1Unorm,
+        Tf::Etc2RgbA1UnormSrgb => H::Etc2R8g8b8a1Srgb,
+        Tf::Etc2RgbA8Unorm => H::Etc2R8g8b8a8Unorm,
+        Tf::Etc2RgbA8UnormSrgb => H::Etc2R8g8b8a8Unorm,
+        Tf::EacRUnorm => H::EacR11Unorm,
+        Tf::EacRSnorm => H::EacR11Snorm,
+        Tf::EtcRgUnorm => H::EacR11g11Unorm,
+        Tf::EtcRgSnorm => H::EacR11g11Snorm,
 
-        Tf::Bc1RgbaUnorm
-        | Tf::Bc1RgbaUnormSrgb
-        | Tf::Bc2RgbaUnorm
-        | Tf::Bc2RgbaUnormSrgb
-        | Tf::Bc3RgbaUnorm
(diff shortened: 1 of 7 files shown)


wgpu-core/src/instance.rs
```diff
@@ -144,6 +144,14 @@ impl<B: GfxBackend> Adapter<B> {
             wgt::Features::TEXTURE_COMPRESSION_BC,
             adapter_features.contains(hal::Features::FORMAT_BC),
         );
+        features.set(
```
kvark, line 147: This code know when to expose the features. Where is the code that enables them for device creation, if requested?

wgpu-core/src/validation.rs
```diff
@@ -729,32 +729,71 @@ fn map_texture_format(format: wgt::TextureFormat) -> naga::TypeInner {
         Tf::Depth32Float | Tf::Depth24Plus | Tf::Depth24PlusStencil8 => {
             panic!("Unexpected depth format")
         }
+
+        // Rgba Compressed Textures
```
kvark, line 733: would be great to have my stuff landing first, since there is a conflict in this bit of code

wgpu-types/src/lib.rs
```diff
@@ -837,46 +876,371 @@ pub enum TextureFormat {
...
+        let Uint = TextureSampleType::Uint;
+        let Sint = TextureSampleType::Sint;
+        let Nearest = TextureSampleType::Float { filterable: false };
+        let Float = TextureSampleType::Float { filterable: true };
+        let Depth = TextureSampleType::Depth;
+
+        let Linear = false;
+        let Srgb = true;
+
+        texture_format_info_listing! {
```
kvark, line 1146: do we actually need a macro here? it seems to me that we could just write it plain as:
```rust
use Self::*;
let (features, sample_type, block_dimensions, block_size, srgb) = match *self {
  R8Unorm => (NATIVE, Float, (1, 1), 1, Linear),
  ...
};
```

macros introduce another level of complexity, and they confuse both readers and tools (like `rustfmt`).

wgpu-types/src/lib.rs
```diff
@@ -1477,6 +1841,95 @@ impl Default for Extent3d {
     }
 }
 
+impl Extent3d {
+    /// Calculates the [physical size] is backing an texture of the given format and extent.
+    /// This includes padding to the block width and height of the format.
+    ///
+    /// This is the size that you must upload at when uploading to _mipmaps_ of compressed textures.
```
kvark, line 1848: worth noting that you can only use this size for `write_texture`. If you are using `copy_buffer_to_texture`, you'd have to also respect the bytes-per-row alignment.

wgpu-types/src/lib.rs
```diff
@@ -1477,6 +1841,95 @@ impl Default for Extent3d {
...
+    }
+
+    /// Calculates the maximum possible count of mipmaps.
+    ///
+    /// Treats the depth as part of the mipmaps. If calculating
+    /// for a 2DArray texture, which does not mipmap depth, set depth to 1.
+    ///
+    /// ```rust
+    /// # use wgpu_types as wgpu;
+    /// assert_eq!(wgpu::Extent3d { width: 1, height: 1, depth: 1 }.max_mips(), 1);
```
kvark, line 1887: fancy doctests :)

wgpu-types/src/lib.rs
```diff
@@ -1767,64 +2220,7 @@ impl Default for TextureSampleType {
 
 impl From<TextureFormat> for TextureSampleType {
```
kvark, line 2221: let's just remove this `impl` now

kvark (commented):
Great work here! Looking forward to have all the compressed formats supported :)

wgpu-core/src/validation.rs
cwfitzgerald: Sounds good

wgpu-types/src/lib.rs
cwfitzgerald: Yeah, that's fine, the macro was needed before but overstayed its welcome.

wgpu-types/src/lib.rs
cwfitzgerald: Clarified that this was only for the texture extent.

kvark (approved):
Wonderful! Please rebase on #1071 and land this when ready.

cwfitzgerald:
I fixed up the validation to better deal with the difference between virtual and physical sizes in compressed texture mips. Could you review the latest commit as that's entirely new code.

wgpu-core/src/conv.rs
```diff
@@ -755,3 +755,16 @@ pub fn map_index_format(index_format: wgt::IndexFormat) -> hal::IndexType {
         wgt::IndexFormat::Uint32 => hal::IndexType::U32,
     }
 }
+
+/// Take `value` and round it up to the nearest alignment `alignment`.
+///
+/// ```text
+/// (0, 3) -> 0
```
kvark, line 762: should this be written as Rust assertions?

wgpu-types/src/lib.rs
```diff
@@ -1881,8 +1881,8 @@ impl Extent3d {
         let block_width = block_width as u32;
         let block_height = block_height as u32;
 
-        let width = ((self.width + block_width - 1) / block_width) * block_width;
-        let height = ((self.height + block_height - 1) / block_height) * block_height;
+        let width = (((self.width + block_width) - 1) / block_width) * block_width;
```
kvark, line 1884: what happened here?

wgpu-core/src/conv.rs
cwfitzgerald: I tried, but because the conv module isn't visible from the outside, and doctests are technically completely different crates, I couldn't actually call the function in the doctest.

wgpu-types/src/lib.rs
cwfitzgerald: ... I thought I was fixing precidence but I didn't change it at all, let me give this another look

kvark:
Please rebase and land this!

cwfitzgerald:
Thanks for the review!

bors r=kvark

PR #1077, opened 2020-12-06
Remove extern "C" from set_index_buffer

DevOrc:
**Connections**
Fixes an issue that is blocking gfx-rs/wgpu-native#61

**Description**
Moves the ffi definition of wgpu_render_bundle_set_index_buffer / wgpu_render_pass_set_index_buffer to wgpu-native. This is needed because wgpu-native has its own version of IndexFormat that is different than the wgpu_types version. 

**Testing**
My projects with wgpu-native work

### DIFF
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -436,6 +436,22 @@ impl RenderBundleEncoder {
             life_guard: LifeGuard::new(desc.label.borrow_or_default()),
         })
     }
+
+    pub fn set_index_buffer(
+        &mut self,
+        buffer_id: id::BufferId,
+        index_format: wgt::IndexFormat,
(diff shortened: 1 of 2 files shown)


wgpu-core/src/command/bundle.rs
```diff
@@ -1076,8 +1076,7 @@ pub mod bundle_ffi {
             .push(RenderCommand::SetPipeline(pipeline_id));
     }
 
-    #[no_mangle]
-    pub extern "C" fn wgpu_render_bundle_set_index_buffer(
+    pub fn wgpu_render_bundle_set_index_buffer(
```
kvark, line 1079: We should probably just make it a method on `RenderBundleEncoder` then. No need for a standalone function, it's just confusing the reader.

kvark (approved):
thank you!
bors r+

PR #1078, opened 2020-12-06
Switch paytests to WGSL

### LINKED ISSUE #1073: Port playtest shaders to WGSL
WGSL is the future for WebGPU, and the CTS uses it. Our current setup has SPIR-V shaders for playtesting, and we have to keep the original GLSL source around, establish the compilation pipeline, etc. All of that stuff can go away once we just use WGSL shaders.

Use only WGSL sources in "player/tests/data", combine multiple entry points into larger modules.

Keep using SPIR-V would be an alternative.

**Additional context**
Blocked by Naga issues:
  - https://github.com/gfx-rs/naga/issues/303
  - ~~https://github.com/gfx-rs/naga/pull/302~~
  - ~~https://github.com/gfx-rs/naga/pull/304~~
  - ~~https://github.com/gfx-rs/naga/issues/305~~
  - https://github.com/gfx-rs/naga/pull/309
  - https://github.com/gfx-rs/naga/pull/310
  - https://github.com/gfx-rs/naga/issues/311

This becomes more feasible with #1071, since we are able to use the naga modules properly.

kvark:
**Connections**
Closes #1073
Updates naga to "gfx-4" tag.
Updates gfx to https://github.com/gfx-rs/gfx/pull/3518

**Description**
This is the first use of WGSL in our codebase! Much nicer to have WGSL for tests than keeping GLSL+SPV pairs.
Interestingly, the tests go through Naga without ever touching SPIR-V now, unless running on Vulkan.

**Testing**
Tests itself - this change applies to testing infra.


player/tests/test.rs
```diff
@@ -144,7 +144,9 @@ impl Test<'_> {
                 }
             };
 
-            assert_eq!(&expected_data[..], contents);
+            if &expected_data[..] != contents {
+                panic!("Test expectation is not met!");
```
grovesNL, line 148: Could we use `assert_eq!(&expected_data[..], contents, "Test expectation is not met!");` here?

grovesNL (approved):
There's a typo in "playtests" in the commit message, but otherwise looks great, thanks!

player/tests/test.rs
kvark: I think that would still try to print out the comparison to the console? We need to either compare less data, or introduce some metric to output (i.e. percent of difference). In any way, the current state of things is not the best

kvark:
Fixed the commit message now.
Excited to land this at last! Has been busy getting Naga in shape for this :)
bors r=grovesNL

kvark:
Time to 🚀 
bors r=grovesNL

PR #1080, opened 2020-12-07
Fix the validation of vertex buffer sizes

JCapucho:
**Connections**
None that i know of

**Description**
~~The vertex buffer size (in vertices) was being divided by stride causing the limit to be lower than it was supposed to be.~~
This bug wasn't triggered earlier because if the stride was 0 it wouldn't perform any calculation and the stride was only set when a set pipeline command was received and the `VertexState` `inputs` were already created so the following commands would work:
```
SetPipeline with 1 vertex buffer
SetVertexBuffer with only 4 vertices
Draw 6 vertices
```
This would have passed validation while this wouldn't
```
SetPipeline with 1 vertex buffer of stride 8
SetVertexBuffer with 4 vertices
SetPipeline with 1 vertex buffer of stride 8
SetVertexBuffer with 4 vertices
Draw 3 vertices
```

Now all draw calls have proper vertex validation and not only after the `inputs` are populated

**Testing**
This change was tested after debugging an issue with the draw calls failing in a specific order in [veloren](https://gitlab.com/veloren/veloren/-/tree/imbris/wgpu-master-rebased)

### DIFF
--- a/wgpu-core/src/command/render.rs
+++ b/wgpu-core/src/command/render.rs
@@ -1164,6 +1164,10 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
 
                     state.index.pipeline_format = pipeline.index_format;
 
+                    let vertex_strides_len = pipeline.vertex_strides.len();
+                    while state.vertex.inputs.len() < vertex_strides_len {
+                        state.vertex.inputs.push(VertexBufferState::EMPTY);
+                    }
                     // Update vertex buffer limits
                     for (vbs, &(stride, rate)) in
                         state.vertex.inputs.iter_mut().zip(&pipeline.vertex_strides)
@@ -1171,7 +1175,6 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                         vbs.stride = stride;
                         vbs.rate = rate;
                     }
-                    let vertex_strides_len = pipeline.vertex_strides.len();
                     for vbs in state.vertex.inputs.iter_mut().skip(vertex_strides_len) {
                         vbs.stride = 0;
                         vbs.rate = InputStepMode::Vertex;


JCapucho:
Found out another issue with validation if when setting another pipeline it has more vertex buffers and they are instances the `inputs` won't have the correct `rate`

wgpu-core/src/command/render.rs
```diff
@@ -259,10 +259,7 @@ impl VertexState {
         self.vertex_limit = !0;
         self.instance_limit = !0;
         for vbs in &self.inputs {
-            if vbs.stride == 0 {
-                continue;
-            }
-            let limit = (vbs.total_size / vbs.stride) as u32;
+            let limit = vbs.total_size as u32;
```
kvark, line 262: why is this change done? the limit is supposed to be on the number of vertices. If we ignore the stride, we are counting the bytes now.

wgpu-core/src/command/render.rs
```diff
@@ -1164,17 +1161,23 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
 
                     state.index.pipeline_format = pipeline.index_format;
 
+                    let vertex_strides_len = pipeline.vertex_strides.len();
```
kvark, line 1164: could we make it less fancy and just do a plain loop
```rust
while state.vertex.inputs.len() < pipeline.vertex_strides.len() {
  state.vertex.inputs.push(VertexBufferState::EMPTY);
}
```

wgpu-core/src/command/render.rs
```diff
@@ -1164,17 +1161,23 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
                         state.vertex.inputs.iter_mut().zip(&pipeline.vertex_strides)
                     {
                         vbs.stride = stride;
                         vbs.rate = rate;
                     }
-                    let vertex_strides_len = pipeline.vertex_strides.len();
                     for vbs in state.vertex.inputs.iter_mut().skip(vertex_strides_len) {
                         vbs.stride = 0;
                         vbs.rate = InputStepMode::Vertex;
+                        vbs.total_size = !0;
```
kvark, line 1180: why are we changing the total size here?

wgpu-core/src/command/render.rs
JCapucho: I also have some doubts about this but some time total_size contains the number of vertices instead of bytes

wgpu-core/src/command/render.rs
JCapucho: So that unused buffers don't cause validation to fail

wgpu-core/src/command/render.rs
kvark: there is only one place where it's assigned, and it's in bytes:
```rust
state.vertex.inputs[slot as usize].total_size = match size {
                        Some(s) => s.get(),
                        None => buffer.size - offset,
                    };
```

wgpu-core/src/command/render.rs
kvark: but we are resetting the `stride`, so the total size wouldn't be taken into account, anyway

wgpu-core/src/command/render.rs
JCapucho: You're right I was confusing things

kvark (approved):
Thank you!
mind squashing it before we proceed?

JCapucho:
Done

kvark (approved):
Thank you!
bors r+

Issue #949, opened 2020-09-29
Read-write buffers should include SHADER_WRITE_BIT and SHADER_READ_BIT


kvark:
Discovered by @cwfitzgerald in #948

PR #1084, opened 2020-12-10
Add STORAGE_LOAD usage to non-readonly buffers

infinitesnow:
Change-Id: I72261913780b5f1540f88f8eb527efd47b0872fa

**Connections**
https://github.com/gfx-rs/wgpu/issues/949

**Description**
Non-readonly buffers are created only with the STORAGE_WRITE usage flag. Add STORAGE_LOAD usage flag.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -1167,7 +1167,7 @@ impl<B: GfxBackend> Device<B> {
                             if read_only {
                                 resource::BufferUse::STORAGE_LOAD
                             } else {
-                                resource::BufferUse::STORAGE_STORE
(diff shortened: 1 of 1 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -1167,7 +1167,7 @@ impl<B: GfxBackend> Device<B> {
                             if read_only {
                                 resource::BufferUse::STORAGE_LOAD
                             } else {
-                                resource::BufferUse::STORAGE_STORE
+                                resource::BufferUse::STORAGE_LOAD | resource::BufferUse::STORAGE_STORE
```
kvark, line 1170: I believe that would make us throw a validation error because of conflicting usage.
The fix needs to be at a lower level.

kvark (changes requested):

PR #1086, opened 2020-12-10
Add SHADER_READ usage bit to non-readonly buffers

### LINKED ISSUE #949: Read-write buffers should include SHADER_WRITE_BIT and SHADER_READ_BIT

infinitesnow:
Change-Id: I99d189e7ef295a58c2179635f95656eb9a2f5fe6

**Connections**
Fixes #949

**Description**
 Non-readonly buffers are created with the STORAGE_WRITE usage flag, but that implies SHADER_READ too.

### DIFF
--- a/wgpu-core/src/conv.rs
+++ b/wgpu-core/src/conv.rs
@@ -594,7 +594,7 @@ pub(crate) fn map_buffer_state(usage: resource::BufferUse) -> hal::buffer::State
         access |= A::SHADER_READ;
     }
     if usage.contains(W::STORAGE_STORE) {
-        access |= A::SHADER_WRITE;
+        access |= A::SHADER_READ | A::SHADER_WRITE;
     }
     if usage.contains(W::INDIRECT) {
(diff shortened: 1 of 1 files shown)


wgpu-core/src/conv.rs
```diff
@@ -594,6 +594,7 @@ pub(crate) fn map_buffer_state(usage: resource::BufferUse) -> hal::buffer::State
         access |= A::SHADER_READ;
     }
     if usage.contains(W::STORAGE_STORE) {
+        access |= A::SHADER_READ;
```
kvark, line 597: let's write it as `A::SHADER_READ | A::SHADER_WRITE`, consistently with the other arm

kvark (commented):
That should do! Just one nit

wgpu-core/src/conv.rs
infinitesnow: Missed it. Fixed!

kvark (approved):
thank you!
bors r+

PR #1088, opened 2020-12-15
Fix and improve object labels and command markers

kvark:
**Connections**
Fixes a part of #1089 

**Description**
PR contains a bunch of small but important things:
  - reset the command buffer label after submission
  - add labels to compute and pass descriptors
  - actually push/pop markers for the scope of render bundle, compute passes
  - set the label to render pass command buffers
  - set labels on pipelines

**Testing**
Not tested much


cwfitzgerald (approved):
LGTM!

bors r+

PR #1094, opened 2020-12-17
Try installing LavaPipe

kvark:


kvark:
We'll need to come back to this once mesa-20.3 is available...

PR #1100, opened 2020-12-21
Validate swapchain extent as non-zero

### LINKED ISSUE #1026: Add Validation for Swapchains of Width or Height 0
**Description**

Creating a swapchain with height or width of 0 is UB under vulkan.

**Extra materials**

See also https://github.com/hecrj/iced/issues/603.

**Platform**

Intel/Windows is the only one who seems to care, but UB is UB.

cwfitzgerald:
**Connections**

Fixes #1026. 

**Description**

Zero area swapchains are UB under vulkan, and should be verboten.

**Testing**

Not tested

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -3847,7 +3847,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         fn validate_swap_chain_descriptor(
             config: &mut hal::window::SwapchainConfig,
             caps: &hal::window::SurfaceCapabilities,
-        ) {
+        ) -> Result<(), swap_chain::CreateSwapChainError> {
             let width = config.extent.width;
             let height = config.extent.height;
             if width < caps.extents.start().width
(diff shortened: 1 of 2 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -3870,6 +3870,10 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 );
                 config.present_mode = hal::window::PresentMode::FIFO;
             }
+            if width == 0 || height == 0 {
+                return Err(swap_chain::CreateSwapChainError::ZeroAreaSwapchain);
```
kvark, line 3874: We don't need the `Swapchain` suffix of the error variant of `CreateSwapChainError`

wgpu-core/src/device/mod.rs
```diff
@@ -3870,6 +3870,10 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+            }
+            return Ok(());
```
kvark, line 3876: this shouldn't use `return`

kvark (approved):
thank you!
bors r+

PR #1103, opened 2020-12-21
Fix Assorted Low-Hanging Validation and Error Message Issues

### LINKED ISSUE #393: Validate Buffer and Texture Usages are Non-Empty
> VALIDATION [VUID-VkBufferCreateInfo-usage-requiredbitmask (0)] : vkCreateBuffer: value of pCreateInfo->usage must not be 0. The Vulkan spec states: usage must not be 0 (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkBufferCreateInfo-usage-requiredbitmask)
object info: (type: UNKNOWN, hndl: 0)

If we see that a resource doesn't request any usage, we should artificially add one, e.g. `TRANSFER_SRC`.

### LINKED ISSUE #1053: Missing Validation for Bound Vertex Buffer
**Description**

We currently validate that all vertex buffers are within range, but not that they are existent. 

**Repro steps**

Replace https://github.com/sotrh/learn-wgpu/blob/master/code/beginner/tutorial7-instancing/src/main.rs#L587 with index 0 so vertex buffer 1 isn't set.

We need to error in this case.

**Additional Information**

We should be 99% of the way there to validating this, we have all the data, we just don't check for it when state is ready.

### LINKED ISSUE #1085: Expose if Texture Format is Renderable

Applications should have a way of figuring out if they can render to a particular format of image.

Add a field to what `format.describe()` returns that tells you if it is renderable.

This would need to be a massive match on the users side anyway, so we should try to dedupe.

cwfitzgerald:
**Connections**

Closes #1085
Closes #393
Closes #1053

**Description**

These commits are independent and should be reviewed individually. Combined into a single PR to reduce noise.

Overview of what was done:
- Add validation for empty texture and buffer usage flags. (#393)
- Add allowed texture usage flags to `format.describe()` (#1085)
  Validate new textures follow the allowed usage flags.
- Properly validates vertex and buffers are bound. (#1053)
  Improves error messages when no vertex buffer is bound. (Before it said the limit was 0, now it says something is unbound)
- Improve the vertex buffer overrun messages by keeping track of which slot has the smallest index.

**Testing**

Tested on examples by artificially creating the situation I am trying to validate, as well as running clean examples to make sure they pass validation.

### DIFF
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -313,20 +313,22 @@ impl RenderBundleEncoder {
                     first_instance,
                 } => {
                     let scope = PassErrorScope::Draw;
-                    let (vertex_limit, instance_limit) = state.vertex_limits();
+                    let vertex_limits = state.vertex_limits();
                     let last_vertex = first_vertex + vertex_count;
-                    if last_vertex > vertex_limit {
+                    if last_vertex > vertex_limits.vertex_limit {
                         return Err(DrawError::VertexBeyondLimit {
                             last_vertex,
-                            vertex_limit,
+                            vertex_limit: vertex_limits.vertex_limit,
+                            slot: vertex_limits.vertex_limit_slot,
                         })
                         .map_pass_err(scope);
                     }
                     let last_instance = first_instance + instance_count;
-                    if last_instance > instance_limit {
+                    if last_instance > vertex_limits.instance_limit {
                         return Err(DrawError::InstanceBeyondLimit {
                             last_instance,
-                            instance_limit,
+                            instance_limit: vertex_limits.instance_limit,
+                            slot: vertex_limits.instance_limit_slot,
                         })
                         .map_pass_err(scope);
                     }
@@ -343,7 +345,7 @@ impl RenderBundleEncoder {
                 } => {
                     let scope = PassErrorScope::DrawIndexed;
                     //TODO: validate that base_vertex + max_index() is within the provided range
-                    let (_, instance_limit) = state.vertex_limits();
+                    let vertex_limits = state.vertex_limits();
                     let index_limit = state.index.limit();
                     let last_index = first_index + index_count;
                     if last_index > index_limit {
@@ -354,10 +356,11 @@ impl RenderBundleEncoder {
                         .map_pass_err(scope);
                     }
                     let last_instance = first_instance + instance_count;
-                    if last_instance > instance_limit {
+                    if last_instance > vertex_limits.instance_limit {
                         return Err(DrawError::InstanceBeyondLimit {
                             last_instance,
-                            instance_limit,
+                            instance_limit: vertex_limits.instance_limit,
+                            slot: vertex_limits.instance_limit_slot,
                         })
                         .map_pass_err(scope);
                     }
@@ -856,6 +859,18 @@ impl PushConstantState {
 }
 
 #[derive(Debug)]
+struct VertexLimitState {
+    /// Length of the shortest vertex rate vertex buffer
+    vertex_limit: u32,
+    /// Buffer slot which the shortest vertex rate vertex buffer is bound to
+    vertex_limit_slot: u32,
+    /// Length of the shortest instance rate vertex buffer
+    instance_limit: u32,
+    /// Buffer slot which the shortest instance rate vertex buffer is bound to
+    instance_limit_slot: u32,
+}
+
+#[derive(Debug)]
 struct State {
     trackers: TrackerSet,
     index: IndexState,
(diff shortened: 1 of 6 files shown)


wgpu-core/src/command/bundle.rs
```diff
@@ -869,20 +884,34 @@ struct State {
...
             if vbs.stride == 0 {
                 continue;
             }
             let limit = ((vbs.range.end - vbs.range.start) / vbs.stride) as u32;
             match vbs.rate {
-                wgt::InputStepMode::Vertex => vertex_limit = vertex_limit.min(limit),
-                wgt::InputStepMode::Instance => instance_limit = instance_limit.min(limit),
+                wgt::InputStepMode::Vertex => {
+                    vert_state.vertex_limit = vert_state.vertex_limit.min(limit);
+                    if vert_state.vertex_limit == limit {
```
kvark, line 902: nit: it seems like this would be cleaner to just do with a single `if` instead of `if` + `min`:
```
if limit < vert_state.vertex_limit {
  vert_state.vertex_limit = limit;
  vert_state.vertex_limit_slot = idx as _;
}
```

wgpu-core/src/command/bundle.rs
```diff
@@ -869,20 +884,34 @@ struct State {
...
+                        vert_state.vertex_limit_slot = idx as _;
+                    }
+                }
+                wgt::InputStepMode::Instance => {
+                    vert_state.instance_limit = vert_state.instance_limit.min(limit);
+                    if vert_state.instance_limit == limit {
```
kvark, line 908: same here

wgpu-core/src/command/render.rs
```diff
@@ -324,9 +355,19 @@ impl State {
         if self.blend_color == OptionalState::Required {
             return Err(DrawError::MissingBlendColor);
         }
+        // Pipeline expects an index buffer
         if let Some(pipeline_index_format) = self.index.pipeline_format {
-            if pipeline_index_format != self.index.format {
-                return Err(DrawError::UnmatchedIndexFormats);
+            // We have a buffer bound
+            if let Some(buffer_index_format) = self.index.format {
```
kvark, line 361: nit: you can also do
```rust
let buffer_index_format = self.index.format.ok_or(DrawError::MissingIndexBuffer)?;
```

wgpu-types/src/lib.rs
```diff
@@ -1177,116 +1179,123 @@ impl TextureFormat {
         let Linear = false;
         let Srgb = true;
 
-        let (features, sample_type, srgb, block_dimensions, block_size) = match self {
+        // Flags
+        let Basic = TextureUsage::COPY_SRC | TextureUsage::COPY_DST | TextureUsage::SAMPLED;
```
kvark, line 1183: why doing the weird naming scheme? locals should start with lower case

wgpu-types/src/lib.rs
```diff
@@ -1177,116 +1179,123 @@ impl TextureFormat {
         let Linear = false;
         let Srgb = true;
 
-        let (features, sample_type, srgb, block_dimensions, block_size) = match self {
+        // Flags
```
kvark, line 1182: perhaps we could link to https://gpuweb.github.io/gpuweb/#texture-format-caps in a comment?

wgpu-types/src/lib.rs
```diff
@@ -1177,116 +1179,123 @@ impl TextureFormat {
...
-            Self::Rg8Unorm => (NATIVE, Float, Linear, (1, 1), 2),
-            Self::Rg8Snorm => (NATIVE, Float, Linear, (1, 1), 2),
-            Self::Rg8Uint => (NATIVE, Uint, Linear, (1, 1), 2),
-            Self::Rg8Sint => (NATIVE, Sint, Linear, (1, 1), 2),
+            Self::R16Uint => (NATIVE, Uint, Linear, (1, 1), 2, Attachment),
+            Self::R16Sint => (NATIVE, Sint, Linear, (1, 1), 2, Attachment),
+            Self::R16Float => (NATIVE, Float, Linear, (1, 1), 2, Attachment),
+            Self::Rg8Unorm => (NATIVE, Float, Linear, (1, 1), 2, Attachment),
+            Self::Rg8Snorm => (NATIVE, Float, Linear, (1, 1), 2, Attachment),
+            Self::Rg8Uint => (NATIVE, Uint, Linear, (1, 1), 2, Basic),
```
kvark, line 1202: why is this non-renderable?

wgpu-types/src/lib.rs
```diff
@@ -1177,116 +1179,123 @@ impl TextureFormat {
...
-            Self::Rgba8UnormSrgb => (NATIVE, Float, Srgb, (1, 1), 4),
-            Self::Rgba8Snorm => (NATIVE, Float, Linear, (1, 1), 4),
-            Self::Rgba8Uint => (NATIVE, Uint, Linear, (1, 1), 4),
-            Self::Rgba8Sint => (NATIVE, Sint, Linear, (1, 1), 4),
-            Self::Bgra8Unorm => (NATIVE, Float, Linear, (1, 1), 4),
-            Self::Bgra8UnormSrgb => (NATIVE, Float, Srgb, (1, 1), 4),
+            Self::R32Uint => (NATIVE, Uint, Linear, (1, 1), 4, AllFlags),
+            Self::R32Sint => (NATIVE, Sint, Linear, (1, 1), 4, AllFlags),
+            Self::R32Float => (NATIVE, Nearest, Linear, (1, 1), 4, AllFlags),
+            Self::Rg16Uint => (NATIVE, Uint, Linear, (1, 1), 4, AllFlags),
```
kvark, line 1209: these can't be STORAGE, please double-check with the table in the spec

wgpu-types/src/lib.rs
```diff
@@ -1177,116 +1179,123 @@ impl TextureFormat {
...
-            Self::Rgba32Float => (NATIVE, Nearest, Linear, (1, 1), 16),
+            Self::Rgba32Uint => (NATIVE, Uint, Linear, (1, 1), 16, AllFlags),
+            Self::Rgba32Sint => (NATIVE, Sint, Linear, (1, 1), 16, AllFlags),
+            Self::Rgba32Float => (NATIVE, Nearest, Linear, (1, 1), 16, AllFlags),
 
             // Depth-stencil textures
-            Self::Depth32Float => (NATIVE, Depth, Linear, (1, 1), 4),
-            Self::Depth24Plus => (NATIVE, Depth, Linear, (1, 1), 4),
-            Self::Depth24PlusStencil8 => (NATIVE, Depth, Linear, (1, 1), 4),
+            Self::Depth32Float => (NATIVE, Depth, Linear, (1, 1), 4, Attachment),
```
kvark, line 1238: we can't have COPY_DST in these formats

kvark (commented):
Amazing work here!
I got a couple of requests, but overall excited to have this landing 🚀

wgpu-types/src/lib.rs
cwfitzgerald: This had justification at one point, but I don't remember it, and either way on rethink I don't agree with it, will change to conventional names.

wgpu-types/src/lib.rs
cwfitzgerald: Bleh, typo

kvark (approved):
please proceed when ready :)

cwfitzgerald:
bors r=kvark

PR #1106, opened 2020-12-22
Encode render passes while command buffer is only read-locked

### LINKED ISSUE #1104: Allow Parallel Recording of RenderPasses

Currently, the best way to multithreaded the command recording of wgpu applications is to spread the job of recording onto multiple threads. However, due to a singular write lock we take on the command buffer list [here](https://github.com/gfx-rs/wgpu/blob/master/wgpu-core/src/command/render.rs#L522), it enforces that only one RenderPass will be running at once.

Demote this write lock to a read lock. The simplest way is to make the registry hold `Registry<RwLock<CommandBuffer<B>>, CommandBufferId, F>`. There may be better ideas.

None. This is a pretty vital feature and one of the big appeals of modern apis, so we want to make sure this works well.

**Additional context**

This came up when someone was trying to improve encoding time of a program with tens of thousands of direct draw calls. Parallelizing the job made it perform no better.

kvark:
**Connections**
Could fix #1104 to some extent

**Description**
The heaviest part of the render pass encoding doesn't need a write lock. Only in the end we need it.

**Testing**
Untested.

### DIFF
--- a/wgpu-core/src/command/mod.rs
+++ b/wgpu-core/src/command/mod.rs
@@ -64,6 +64,17 @@ impl<B: GfxBackend> CommandBuffer<B> {
         }
     }
 
+    fn get_encoder_read_only(
+        storage: &Storage<Self, id::CommandEncoderId>,
+        id: id::CommandEncoderId,
+    ) -> Result<&Self, CommandEncoderError> {
+        match storage.get(id) {
+            Ok(cmd_buf) if cmd_buf.is_recording => Ok(cmd_buf),
+            Ok(_) => Err(CommandEncoderError::NotRecording),
+            Err(_) => Err(CommandEncoderError::Invalid),
+        }
+    }
+
     pub(crate) fn insert_barriers(
         raw: &mut B::CommandBuffer,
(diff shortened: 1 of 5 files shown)


kvark:
Note: this would benefit from a good refactor to reduce nesting. I don't want to invest in the refactor until we know if this is the way to go.

kvark:
@cwfitzgerald I took a stab at making the code nicer in the last commit. Please check it out.
The PR should be ready for landing otherwise.

wgpu-core/src/command/mod.rs
```diff
@@ -64,6 +64,17 @@ impl<B: GfxBackend> CommandBuffer<B> {
         }
     }
 
+    fn get_encoder_read_only(
```
cwfitzgerald, line 67: Nit: I think it would be better to have this follow thee `get_encoder` `get_encoder_mut` pattern

cwfitzgerald (approved):
Beautiful, thank you for tackling this one.

One very minor nit which you can take or leave.

kvark:
bors r=cwfitzgerald

PR #1109, opened 2020-12-26
Add STORAGE_TEXTURE_ACCESS_READ_WRITE native extension

Wumpf:
**Description**
Adds STORAGE_TEXTURE_ACCESS_READ_WRITE extension and the related StorageTextureAccess::ReadWrite.

There's not per format validation for now (as far as I can tell we're also missing necessary hooks/queries for that in hal), I think this something we want/need in the long run.

**Testing**
Quick & dirty manual testing only so far.

### DIFF
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -132,7 +132,8 @@ impl<B: GfxBackend> Adapter<B> {
 
         let mut features = wgt::Features::default()
             | wgt::Features::MAPPABLE_PRIMARY_BUFFERS
-            | wgt::Features::PUSH_CONSTANTS;
+            | wgt::Features::PUSH_CONSTANTS
+            | wgt::Features::STORAGE_TEXTURE_ACCESS_READ_WRITE; // TODO: Should check for any support at all (applies only to Metal) and expose fine grained format support somehow
         features.set(
             wgt::Features::DEPTH_CLAMPING,
             adapter_features.contains(hal::Features::DEPTH_CLAMP),
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -1310,6 +1310,13 @@ impl<B: GfxBackend> Device<B> {
                                 wgt::StorageTextureAccess::WriteOnly => {
                                     resource::TextureUse::STORAGE_STORE
                                 }
+                                wgt::StorageTextureAccess::ReadWrite => {
+                                    if !self.features.contains(wgt::Features::STORAGE_TEXTURE_ACCESS_READ_WRITE) {
+                                        return Err(Error::MissingFeatures(wgt::Features::STORAGE_TEXTURE_ACCESS_READ_WRITE));
+                                    }
+                                    resource::TextureUse::STORAGE_STORE
+                                        | resource::TextureUse::STORAGE_LOAD
+                                }
                             },
                         ),
                         _ => return Err(Error::WrongBindingType {
--- a/wgpu-core/src/validation.rs
+++ b/wgpu-core/src/validation.rs
@@ -418,6 +418,7 @@ impl Resource {
                         let usage = match access {
                             wgt::StorageTextureAccess::ReadOnly => naga::GlobalUse::LOAD,
                             wgt::StorageTextureAccess::WriteOnly => naga::GlobalUse::STORE,
+                            wgt::StorageTextureAccess::ReadWrite => naga::GlobalUse::all(),
                         };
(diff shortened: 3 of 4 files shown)


wgpu-core/src/instance.rs
```diff
@@ -132,7 +132,8 @@ impl<B: GfxBackend> Adapter<B> {
 
         let mut features = wgt::Features::default()
             | wgt::Features::MAPPABLE_PRIMARY_BUFFERS
-            | wgt::Features::PUSH_CONSTANTS;
+            | wgt::Features::PUSH_CONSTANTS
+            | wgt::Features::STORAGE_TEXTURE_ACCESS_READ_WRITE; // TODO: Should check for any support at all (applies only to Metal) and expose fine grained format support somehow
```
kvark, line 136: yeah, gfx-hal changes are needed to get that

kvark (commented):
The main problem here is - how do we let the users discover which formats support this? There must be some kind of API here, or otherwise this isn't really portable, even across platforms that support the new feature (i.e. all of them in the PR).

Wumpf:
I think we could get away without extra api if we want to: Unlike what I wrote in the code make `STORAGE_TEXTURE_ACCESS_READ_WRITE` require that *all* storage textures are read/writable.

Pros:
* no api changes, texture format query can stay as it is!
* format queries only at device selection time - none at all for Vulkan devices, single simple query for Metal (MTLReadWriteTextureTier2)
Con:
* future users may want to extend the extension so they can use a specific format for r/w which is supported on their device while another one on the set is not (making the extension non-available)
* this effectively introduces capability tiers which is not a good mapping over the texture format landscape [citation needed]


To get some clarity it might help here to have a look at what comes inevitably next: Opt-in storage support for formats that are not supported in webgpu.
Again there is the easy way out, by simply taking over Vulkan's simplified approach and expose [`shaderStorageImageExtendedFormats`](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#features-shaderStorageImageExtendedFormats) which requires a whole group of textures. I prototyped what this could look like in wgpu [here](https://github.com/Wumpf/wgpu/commit/1a19d9413e863353387ac9a115b67df8bbf2fd7b). If such a flag were to be exposed, combining it with `STORAGE_TEXTURE_ACCESS_READ_WRITE` would ofc mean that all of these extra formats then would need also to be r/w  - again "free" in Vulkan, fine-grained queries in DX12, not supported at all with Metal. Arguably that's where the "just a bit" story starts to crumble.


I'm nevertheless leaning towards this simplified approach since otherwise the device query gets imho more painful for implementer and users. Boils down to how flexible we want to be!

So what would the alternative look like api-wise? One straight forward and explicit potential approach I'm thinking of:
* no change in texture format query (I think we want to avoid needing the device for this)
* pass in all requirements via the Limits struct on device creation
   * list of required read/write storage formats
   * list of additional write storage formats
   * list of additional read storage formats
   
-> ignore overlaps of lists / don't try to draw conclusions, just validate lists as is
-> texture creation with STORAGE checks if it format is in webgpu set or any of the above lists
-> texture binding checks is usage+format is in webgpu set or in the right list above (read/write list is enough for either read or write only access)



What do you think? :)
Happy to discuss on the Matrix chat!

kvark:
Thank you for the rich analysis of this problem!

> Again there is the easy way out, by simply taking over Vulkan's simplified approach and expose shaderStorageImageExtendedFormats which requires a whole group of textures

I strongly think that this would be a wrong way to go. Different backends have different capabilities for the storage formats, and trying to fit them all under a single boolean (borrowed from Vulkan) is just too rough...

Getting this API in a portable shape for the Web appears to be a challenge. For the native-only, however, we could just have a query like `Adapter::get_texture_format_properties()` returning bitflags (whether storage is supported, whether read-write storage is supported, and other stuff).

kvark:
> require that all storage textures are read/writable.

On D3D12 it's (unfortunately) not exposed at the device granularity. You can see [this table](https://docs.microsoft.com/en-us/windows/win32/direct3ddxgi/hardware-support-for-direct3d-12-1-formats), look for "UAV Typed Load".

As described in https://docs.microsoft.com/en-us/windows/win32/direct3d12/typed-unordered-access-view-loads, there is a per-format query for this operation support...

Note: read-only textures in D3D12 are not UAV at all (at least in wgpu/gfx), they are SRV. So "typed UAV load" is precisely what tells us if we can both read and write to it.

Wumpf:
Started over in https://github.com/gfx-rs/wgpu/pull/1112

PR #1112, opened 2020-12-30
Expose adapter specific formats features via Extension, support for storage read+write

Wumpf:
**Connections**
Draft until https://github.com/gfx-rs/gfx/pull/3559 landed (then update and comment two lines in)
New & improved replacement for #1109

PR for wgpu-rs to be opened once there's some confidence that this is that it should look like. [Preview](https://github.com/gfx-rs/wgpu-rs/compare/master...Wumpf:texture-format-feature-query?expand=1)

**Description**
Adapter exposes now per texture format specific features. Feature struct defined as allowed usages + flags. Flags describe only storage texture usage so far.

Query not used for validation unless `TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES` feature is enabled on a device. Otherwise uses hardcoded feature set guaranteed by webgpu.
Storage read/write binding can then be used as a consequence (storage atomic has no effect yet).

**Testing**
Manual testing with my fluid sim project through wgpu-rs which covers storage r/w for a few different formats and storage in general for formats that don't support this usage by default at all.
Ran tests and a couple of samples to see if anything broke with the feature disabled.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -516,15 +516,16 @@ impl<B: GfxBackend> Device<B> {
     fn create_texture(
         &self,
         self_id: id::DeviceId,
+        adapter: &crate::instance::Adapter<B>,
         desc: &resource::TextureDescriptor,
     ) -> Result<resource::Texture<B>, resource::CreateTextureError> {
         debug_assert_eq!(self_id.backend(), B::VARIANT);
 
         let format_desc = desc.format.describe();
-        let features = format_desc.features;
-        if !self.features.contains(features) {
+        let required_features = format_desc.features;
+        if !self.features.contains(required_features) {
             return Err(resource::CreateTextureError::MissingFeature(
-                features,
+                required_features,
                 desc.format,
             ));
         }
@@ -546,20 +547,21 @@ impl<B: GfxBackend> Device<B> {
             return Err(resource::CreateTextureError::EmptyUsage);
         }
 
-        if self
+        let format_features = if self
             .features
             .contains(wgt::Features::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES)
         {
-            // To check if the format usage is not available, we'd need to check with the adapter of which we only have an id here.
-            // Assuming that the underlying device fails to create the texture, this _should_ not be a big problem.
+            adapter.get_texture_format_features(desc.format)
         } else {
-            let missing_allowed_usages = desc.usage - desc.format.features_webgpu().allowed_usages;
-            if !missing_allowed_usages.is_empty() {
-                return Err(resource::CreateTextureError::InvalidUsages(
-                    missing_allowed_usages,
-                    desc.format,
-                ));
-            }
+            desc.format.features_webgpu()
+        };
+
+        let missing_allowed_usages = desc.usage - format_features.allowed_usages;
+        if !missing_allowed_usages.is_empty() {
+            return Err(resource::CreateTextureError::InvalidUsages(
+                missing_allowed_usages,
+                desc.format,
+            ));
         }
 
         let kind = conv::map_texture_dimension_size(desc.dimension, desc.size, desc.sample_count)?;
@@ -626,6 +628,7 @@ impl<B: GfxBackend> Device<B> {
             dimension: desc.dimension,
             kind,
             format: desc.format,
+            format_features,
             full_range: TextureSelector {
                 levels: 0..desc.mip_level_count as hal::image::Level,
                 layers: 0..kind.num_layers(),
@@ -782,6 +785,7 @@ impl<B: GfxBackend> Device<B> {
             },
             aspects,
             format: texture.format,
+            format_features: texture.format_features,
             extent: texture.kind.extent().at_level(desc.base_mip_level as _),
             samples: texture.kind.num_samples(),
             selector,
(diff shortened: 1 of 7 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -1301,17 +1313,34 @@ impl<B: GfxBackend> Device<B> {
...
-                                wgt::StorageTextureAccess::WriteOnly => {
-                                    resource::TextureUse::STORAGE_STORE
-                                }
-                            },
-                        ),
+                        wgt::BindingType::StorageTexture { access, .. } => {
+                            (
+                                wgt::TextureUsage::STORAGE,
+                                match access {
+                                    wgt::StorageTextureAccess::ReadOnly => {
```
kvark, line 1320: let's reduce the indentation here:
```
wgt::BindingType::StorageTexture { access, .. } => {
  let internal_use = match access {
  };
  (wgt::TextureUsage::STORAGE, internal_use)
}
```

wgpu-core/src/device/mod.rs
```diff
@@ -1301,17 +1313,34 @@ impl<B: GfxBackend> Device<B> {
...
+                                        resource::TextureUse::STORAGE_LOAD
+                                    }
+                                    wgt::StorageTextureAccess::WriteOnly => {
+                                        resource::TextureUse::STORAGE_STORE
+                                    }
+                                    wgt::StorageTextureAccess::ReadWrite => {
+                                        if !view.format_features.flags.contains(
+                                            wgt::TextureFormatFeatureFlags::STORAGE_READ_WRITE,
+                                        ) {
+                                            if !self.features.contains(
```
kvark, line 1330: nit: can we move `return Err()` out of the if/else?

wgpu-core/src/instance.rs
```diff
@@ -132,7 +133,8 @@ impl<B: GfxBackend> Adapter<B> {
 
         let mut features = wgt::Features::default()
             | wgt::Features::MAPPABLE_PRIMARY_BUFFERS
-            | wgt::Features::PUSH_CONSTANTS;
+            | wgt::Features::PUSH_CONSTANTS
+            | wgt::Features::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES;
```
kvark, line 137: I wonder if we actually need this feature. I.e. can we pretend it's always supported? On the Web, we can just return the statically known features (until WebGPU gets a proper extension).

wgpu-core/src/instance.rs
```diff
@@ -228,11 +244,60 @@ impl<B: GfxBackend> Adapter<B> {
...
+        if texture_format_properties.contains(hal::format::ImageFeature::BLIT_DST) {
+            allowed_usages |= wgt::TextureUsage::COPY_DST;
+        }
+
+        let mut flags = wgt::TextureFormatFeatureFlags::empty();
+        if texture_format_properties.contains(hal::format::ImageFeature::STORAGE_ATOMIC) {
+            flags |= wgt::TextureFormatFeatureFlags::STORAGE_ATOMICS;
+        }
+        // TODO: hal update
+        //if texture_format_properties.contains(hal::format::ImageFeature::STORAGE_READ_WRITE) {
```
kvark, line 291: can now be implemented thanks to you!

wgpu-types/src/lib.rs
```diff
@@ -754,6 +775,28 @@ pub struct RasterizationStateDescriptor {
     pub depth_bias_clamp: f32,
 }
 
+bitflags::bitflags! {
+    /// Feature flags for a texture format.
+    #[repr(transparent)]
+    #[cfg_attr(feature = "trace", derive(Serialize))]
+    #[cfg_attr(feature = "replay", derive(Deserialize))]
+    pub struct TextureFormatFeatureFlags: u32 {
```
kvark, line 783: nice one!

wgpu-types/src/lib.rs
```diff
@@ -1304,7 +1338,133 @@ impl TextureFormat {
             block_dimensions,
             block_size,
             srgb,
+        }
+    }
+
+    /// Lists features supported by WebGPU, not using any native extensions.
+    pub fn features_webgpu(&self) -> TextureFormatFeatures {
```
kvark, line 1345: why would we want to move it out into a separate function? we could still have it all in the `TextureFormatInfo` and have a single `match` over the formats.

wgpu-core/src/instance.rs
Wumpf: That thought crossed my mind as well but what kept me from it is

* (most important) one might develop a cross platform application (e.g. web + desktop) and unknowingly break its portability if this is not an explicit opt-in. Imho a huge selling point of wgpu(-rs) is that it formulates a feature set that will work on a very wide variety of devices
* other things in the limits struct need to be requested explicitly as well and default to the webgpu specification (e.g. how many storage bindings are supported)
* if and when WebGPU gets an extension this may fit more nicely as a "free-for-all tier"

wgpu-core/src/instance.rs
Wumpf: 😊 
once we managed to upgrade to newest gfx that is :)

wgpu-types/src/lib.rs
Wumpf: TextureFormatInfo describes static information about each format. The "true" feature set of a texture format is now determined at runtime (unless WebGPU). Having a "features iff you don't enable the beyond-webgpu-spec-thing"-field felt a bit clumsy and confusing to me - a pure non-web application may never be interested in the value and it could confuse users.

No strong feelings about it though, so if you dislike the second gargantum-match I can fuse them together, just gotta find a good name for the TextureFormatFeatures field 🤷

kvark: we can have this static information be more explicit, i.e. `guaranteed_features` instead of `features` as the field name would nail it

wgpu-core/src/instance.rs
kvark: That's a very good point! Ideally I'd love to make it so you can't use a feature that you haven't either discovered or requested. This should apply to format features as well. So what you are saying makes total sense, and is much appreciated!

We can also go a slightly different route and consider the format features to be "basic" unless the user asks for them. This would make the same code portable across the web and non-web.

Wumpf: > consider the format features to be "basic" unless the user asks for them

You're thinking that the user can request particular texture features ahead of time? That would make sense in regard to how the limit struct works and keeps all the missing-feature surprises to the device selection rather than some random texture binding.

Or did you mean something else? :)
Otherwise, the way it's done here would already be a manual opt-in by the user, only that it's a bit vague what one ends up with :/

kvark: No, I wasn't think about "ahead of time" kind of thing as we do for limits. I was thinking more simple: if the user didn't call `get_texture_features` (or whatever it's called), then they can only assume the base guaranteed feature set for this format specifically (and we validate that).

Wumpf:
Support for read+write textures is all operational but will always fail validation right now (see commented out code with TODO note  in `get_texture_format_features`) until gfx/hal is updated (necessary changes there have landed!). Update is blocked right now by https://github.com/zakarumych/gpu-alloc/issues/33

kvark (approved):
Very nice, thank you!
bors r+

kvark:
I was going back and forth on some pieces, thinking if they could be written better... but no, it's all really good 🕶️ .

PR #1114, opened 2021-01-02
Fixed error message for BindingTypeMaxCountError

Wumpf:
Previously, `BindingTypeMaxCountError` would incorrectly report the binding count as the violated limit. Now it reports both the limit and the count that violated it.

Example error message before:
```
wgpu error: Validation Error

Caused by:
    In Device::create_bind_group_layout
      note: label = `BindGroupLayout write scalar`
    too many bindings of type StorageTextures in stage COMPUTE, limit is 5
```

Example error message after:
```
wgpu error: Validation Error

Caused by:
    In Device::create_bind_group_layout
      note: label = `BindGroupLayout write scalar`
    too many bindings of type StorageTextures in stage COMPUTE, limit is 4, count was 5
```

### DIFF
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -107,10 +107,11 @@ pub enum BindingZone {
 }
 
 #[derive(Clone, Debug, Error)]
-#[error("too many bindings of type {kind:?} in {zone}, limit is {count}")]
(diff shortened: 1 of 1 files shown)


cwfitzgerald (approved):
Thanks!

bors r+

PR #1115, opened 2021-01-02
Request buffer FAST_DEVICE_ACCESS only with some non MAP or COPY usage

### LINKED ISSUE #1113: Buffer download on AMD Linux is extremely slow
**Description**
Any time I try to read from a download buffer (i.e. usage `MAP_READ`|`COPY_DST`) the accesses are unreasonably slow.

**Repro steps**
The _capture_ example demonstrates this bug, but I have a modified version [here](https://github.com/fintelia/wgpu-rs/tree/bench-download-buffer/examples/capture) which specifically times how long it takes to copy from the download buffer to a heap allocated Vec.

Running on my machine...
```
$ cargo run --example capture --release -- 1024 1024
Copy time = 4.2049 seconds / 4096 KB
```
The observed time is over 4 **seconds** to copy a mere 4MB of data, whereas I would expect the time to be many orders of magnitude faster.

**Platform**
Ubuntu 20.10
AMD RX 580
Driver: RADV or AMDVLK (happens on both)

fintelia:
**Connections**
Fixes #1113 

**Description**
On Linux AMD (and likely elsewhere) buffers with FAST_DEVICE_ACCESS are very slow to access when mapped. This makes sure that upload/download buffers don't request it unless those buffers are leveraging MAPPABLE_PRIMARY_BUFFERS for some non-map/copy usage.

**Testing**
I reran the failing reproduction case in the linked issue.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -448,7 +448,8 @@ impl<B: GfxBackend> Device<B> {
 
             let mut flags = Uf::empty();
             let map_flags = desc.usage & (Bu::MAP_READ | Bu::MAP_WRITE);
-            if !(desc.usage - map_flags).is_empty() {
+            let map_copy_flags = desc.usage & (Bu::MAP_READ | Bu::MAP_WRITE | Bu::COPY_SRC | Bu::COPY_DST);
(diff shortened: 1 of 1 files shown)


kvark (approved):
Nicely done!
bors r+

PR #1117, opened 2021-01-02
Free associated staging buffers

kvark:
**Connections**
Reported on the Matrix. Having a loop that creates mapped buffers and drops them results in OOM.

**Description**
We are now checking for the map state when destroying buffers.

**Testing**
Ran through an ad-hoc example.


cwfitzgerald (approved):
LGTM! Nice catch.

bors r+

PR #1126, opened 2021-01-04
Check swapchain frame to still be in use

### LINKED ISSUE #1123: assertion failed: wgpu-core/src/track/mod.rs:373:21
**Description**
Assertion failure when running wgpu with commit 35b163ebc861cae808ca168cdd5edf4ec91456f9 or later.

**Repro steps**

The easiest thing is probably to use the attached trace (below). But you should also be able to reproduce the bug running terra directly.

Clone...
```bash
git clone https://github.com/fintelia/terra && cd terra
git checkout 9ac9d598f23065543df14a8b1051faf918065c41
```
patch Cargo.toml...
```toml
[replace]
"https://github.com/gfx-rs/wgpu#wgpu-core:0.6.0" = { git = "https://github.com/gfx-rs/wgpu", rev = "b82482f31c6fa30fcb27188b561a3eb6ee056580" }
"https://github.com/gfx-rs/wgpu#wgpu-types:0.6.0" = { git = "https://github.com/gfx-rs/wgpu", rev = "b82482f31c6fa30fcb27188b561a3eb6ee056580" }
```
Run...
```bash
RUST_BACKTRACE=1 cargo run --release
```

Uninstall...
```bash
rm -r ~/.cache/terra
```

I would expect the program to run normally, as it does without the patched wgpu-core version. Instead it crashes with an assertion failure:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `2`', /home/jonathan/.cargo/git/checkouts/wgpu-53e70f8674b08dd4/b82482f/wgpu-core/src/track/mod.rs:373:21
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7f9c43cf98cfe1c369045399929cb098155b8374/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/7f9c43cf98cfe1c369045399929cb098155b8374/library/core/src/panicking.rs:92:14
   2: wgpu_core::track::ResourceTracker<S>::merge_extend
   3: wgpu_core::command::CommandBuffer<B>::insert_barriers
             at /home/jonathan/.cargo/git/checkouts/wgpu-53e70f8674b08dd4/b82482f/wgpu-core/src/command/mod.rs:98:9
   4: wgpu_core::device::queue::<impl wgpu_core::hub::Global<G>>::queue_submit
             at /home/jonathan/.cargo/git/checkouts/wgpu-53e70f8674b08dd4/b82482f/wgpu-core/src/device/queue.rs:603:25
   5: <wgpu::backend::direct::Context as wgpu::Context>::queue_submit
             at /home/jonathan/.cargo/git/checkouts/wgpu-rs-40ea39809c03c5d8/3d60546/src/backend/direct.rs:1698:15
   6: wgpu::Queue::submit
             at /home/jonathan/.cargo/git/checkouts/wgpu-rs-40ea39809c03c5d8/3d60546/src/lib.rs:2605:9
   7: terra::Terrain::render
             at ./src/lib.rs:476:9
```

**Extra materials**

[trace.zip](https://github.com/gfx-rs/wgpu/files/5766077/trace.zip)

**Platform**
Ubuntu 20.10
AMD RX 580
wgpu 35b163ebc861cae808ca168cdd5edf4ec91456f9

kvark:
**Connections**
Fixes #1123

**Description**
What happened there is - the same command buffer was used to record work into multiple swapchain frames.
The ID management of those is a bit special, so the error wasn't very descriptive, internal logic caught inconsistency.
Now there is a proper error, and the old error is better.
Note that ideally we'd want "present" to work like "destroy" on textures, but essentially the users would have to fix it anyway. It's not going to make their life easier (in fact - only harder), so we can defer this to a follow-up.

**Testing**
Tested on wgpu-rs examples as well as on Terra.

### DIFF
--- a/wgpu-core/src/swap_chain.rs
+++ b/wgpu-core/src/swap_chain.rs
@@ -81,6 +81,8 @@ pub enum SwapChainError {
     Device(#[from] DeviceError),
     #[error("swap chain image is already acquired")]
     AlreadyAcquired,
+    #[error("acquired frame is still referenced")]
+    StillReferenced,
 }
 
 #[derive(Clone, Debug, Error)]
@@ -250,12 +252,17 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             trace.lock().add(Action::PresentSwapChain(swap_chain_id));
         }
 
-        let view_id = sc
-            .acquired_view_id
-            .take()
-            .ok_or(SwapChainError::AlreadyAcquired)?;
-        let (view_maybe, _) = hub.texture_views.unregister(view_id.value.0, &mut token);
-        let view = view_maybe.ok_or(SwapChainError::Invalid)?;
+        let view = {
+            let view_id = sc
+                .acquired_view_id
+                .take()
+                .ok_or(SwapChainError::AlreadyAcquired)?;
+            let (view_maybe, _) = hub.texture_views.unregister(view_id.value.0, &mut token);
+            view_maybe.ok_or(SwapChainError::Invalid)?
+        };
+        if view.life_guard.ref_count.unwrap().load() != 1 {
+            return Err(SwapChainError::StillReferenced);
+        }
         let image = match view.inner {
             resource::TextureViewInner::Native { .. } => unreachable!(),
             resource::TextureViewInner::SwapChain { image, .. } => image,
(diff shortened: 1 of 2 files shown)


kvark:
bors r=harmless

PR #1127, opened 2021-01-04
Fix API traced dynamic offsets and markers

kvark:
**Connections**
This was a regression from #1106, discovered while investigating #1123

**Description**
When running a pass, it was mutably used for "convenience". Once we moved the tracing of that pass from the beginning of the function to the end, the "convenience" eliminated our dynamic offsets and markers.
The solution is to stop mutating the passes inside our big functions.

**Testing**
Re-traced Terra


kvark:
This fixes API tracing, which is critical for our workflow.
bors r=critical

PR #1128, opened 2021-01-05
Add pipeline statistics and timeline queries

### LINKED ISSUE #721: WebGPU "Queries" implementation (for profiling purposes)
_I am strictly a hobbyist when it comes to graphics programming, so please forgive any misinformed assumptions that I make :)_

I am trying to benchmark individual render passes in an application built on `wgpu-rs`. In my research to resolve this, I stumbled upon the `"timestamp"` [`GPUQueryType`](https://gpuweb.github.io/gpuweb/#querytype). This API _seems_ like it exposes data which could be used to benchmark any particular command, render passes among them (a very useful feature indeed!).

However, I could not find any APIs for interfacing with these Queries in `wgpu-rs`. I followed the trail and it seems that `gfx-hal` [does provide this interface](https://github.com/gfx-rs/gfx/blob/b589a4542d0d1fcd955377ca16f34fc155ca1edf/src/hal/src/query.rs), which (in theory) means it can be propagated to my application code with the only necessary changes being contained within `wgpu-rs` and `wgpu-core`. I have two theories as to why this feature is not already implemented:

1) Software projects must allocate their limited resources judiciously; profiling isn't very useful if there is nothing to profile!
2) [This open issue](https://github.com/gfx-rs/gfx/issues/3145) on `gfx-hal` precludes it from being spec-compliant with WebGPU, namely [this restriction](https://gpuweb.github.io/gpuweb/#timestamp):

> Timestamp query requires `timestamp-query` is available on the device.

I am willing to put in some legwork to make this happen, but I want to make sure my understanding of the situation is correct before I invest too much time. Any guidance would also be appreciated. Thanks for working on `wgpu-rs`!

cwfitzgerald:
**Connections**

Closes #721 

**Description**

This adds Pipeline Statistics queries and Timeline queries. Shoutout to @z2oh for doing a good chunk of the work to enable this feature.

Currently blocked on both https://github.com/gfx-rs/gfx/pull/3563 and rollup of gfx-hal mutability changes, but is ready to review.

**Testing**

Tested in various ways on the hello-compute and hello-triangle example, tested validation extensively. Queries will be permanently added to the mipmap example once the wgpu-rs PR is finished.

### DIFF
--- a/wgpu-core/src/command/query.rs
+++ b/wgpu-core/src/command/query.rs
@@ -1,0 +1,424 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use hal::command::CommandBuffer as _;
+
+#[cfg(feature = "trace")]
+use crate::device::trace::Command as TraceCommand;
+use crate::{
+    command::{CommandBuffer, CommandEncoderError},
+    device::all_buffer_stages,
+    hub::{GfxBackend, Global, GlobalIdentityHandlerFactory, Storage, Token},
+    id::{self, Id, TypedId},
+    resource::{BufferUse, QuerySet},
+    track::UseExtendError,
+    Epoch, FastHashMap, Index,
+};
+use std::{iter, marker::PhantomData};
+use thiserror::Error;
+use wgt::BufferAddress;
+
+#[derive(Debug)]
+pub(super) struct QueryResetMap<B: hal::Backend> {
+    map: FastHashMap<Index, (Vec<bool>, Epoch)>,
+    _phantom: PhantomData<B>,
+}
+impl<B: hal::Backend> QueryResetMap<B> {
+    pub fn new() -> Self {
+        Self {
+            map: FastHashMap::default(),
+            _phantom: PhantomData,
+        }
+    }
+
+    pub fn use_query_set(
+        &mut self,
+        id: id::QuerySetId,
+        query_set: &QuerySet<B>,
+        query: u32,
+    ) -> bool {
+        let (index, epoch, _) = id.unzip();
+        let (vec, _) = self
+            .map
+            .entry(index)
+            .or_insert_with(|| (vec![false; query_set.desc.count as usize], epoch));
+
+        std::mem::replace(&mut vec[query as usize], true)
+    }
+
+    pub fn reset_queries(
+        self,
+        cmd_buf_raw: &mut B::CommandBuffer,
+        query_set_storage: &Storage<QuerySet<B>, id::QuerySetId>,
+        backend: wgt::Backend,
+    ) -> Result<(), id::QuerySetId> {
+        for (query_set_id, (state, epoch)) in self.map.into_iter() {
+            let id = Id::zip(query_set_id, epoch, backend);
+            let query_set = query_set_storage.get(id).map_err(|_| id)?;
+
+            debug_assert_eq!(state.len(), query_set.desc.count as usize);
+
+            // Need to find all "runs" of values which need resets. If the state vector is:
+            // [false, true, true, false, true], we want to reset [1..3, 4..5]. This minimizes
+            // the amount of resets needed.
+            let mut state_iter = state.into_iter().chain(iter::once(false)).enumerate();
+            let mut run_start: Option<u32> = None;
+            while let Some((idx, value)) = state_iter.next() {
+                match (run_start, value) {
+                    // We're inside of a run, do nothing
+                    (Some(..), true) => {}
+                    // We've hit the end of a run, dispatch a reset
+                    (Some(start), false) => {
+                        run_start = None;
+                        unsafe { cmd_buf_raw.reset_query_pool(&query_set.raw, start..idx as u32) };
+                    }
+                    // We're starting a run
+                    (None, true) => {
+                        run_start = Some(idx as u32);
+                    }
+                    // We're in a run of falses, do nothing.
+                    (None, false) => {}
+                }
+            }
+        }
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum SimplifiedQueryType {
+    Timestamp,
+    PipelineStatistics,
+}
+impl From<wgt::QueryType> for SimplifiedQueryType {
+    fn from(q: wgt::QueryType) -> Self {
+        match q {
+            wgt::QueryType::Timestamp => SimplifiedQueryType::Timestamp,
+            wgt::QueryType::PipelineStatistics(..) => SimplifiedQueryType::PipelineStatistics,
+        }
+    }
+}
+
+/// Error encountered when dealing with queries
+#[derive(Clone, Debug, Error)]
+pub enum QueryError {
+    #[error(transparent)]
+    Encoder(#[from] CommandEncoderError),
+    #[error("Error encountered while trying to use queries")]
+    Use(#[from] QueryUseError),
+    #[error("Error encountered while trying to resolve a query")]
+    Resolve(#[from] ResolveError),
+    #[error("Buffer {0:?} is invalid or destroyed")]
+    InvalidBuffer(id::BufferId),
+    #[error("QuerySet {0:?} is invalid or destroyed")]
+    InvalidQuerySet(id::QuerySetId),
+}
+
+/// Error encountered while trying to use queries
+#[derive(Clone, Debug, Error)]
+pub enum QueryUseError {
+    #[error("Query {query_index} is out of bounds for a query set of size {query_set_size}")]
+    OutOfBounds {
+        query_index: u32,
+        query_set_size: u32,
+    },
+    #[error("Query {query_index} has already been used within the same renderpass. Queries must only be used once per renderpass")]
+    UsedTwiceInsideRenderpass { query_index: u32 },
+    #[error("Query {new_query_index} was started while query {active_query_index} was already active. No more than one statistic or occlusion query may be active at once")]
+    AlreadyStarted {
+        active_query_index: u32,
+        new_query_index: u32,
+    },
+    #[error("Query was stopped while there was no active query")]
+    AlreadyStopped,
+    #[error("A query of type {query_type:?} was started using a query set of type {set_type:?}")]
+    IncompatibleType {
+        set_type: SimplifiedQueryType,
+        query_type: SimplifiedQueryType,
+    },
+}
+
+/// Error encountered while trying to resolve a query.
+#[derive(Clone, Debug, Error)]
+pub enum ResolveError {
+    #[error("Queries can only be resolved to buffers that contain the COPY_DST usage")]
+    MissingBufferUsage,
+    #[error("Resolving queries {start_query}..{end_query} would overrun the query set of size {query_set_size}")]
+    QueryOverrun {
+        start_query: u32,
+        end_query: u32,
+        query_set_size: u32,
+    },
+    #[error("Resolving queries {start_query}..{end_query} ({stride} byte queries) will end up overruning the bounds of the destination buffer of size {buffer_size} using offsets {buffer_start_offset}..{buffer_end_offset}")]
+    BufferOverrun {
+        start_query: u32,
+        end_query: u32,
+        stride: u32,
+        buffer_size: BufferAddress,
+        buffer_start_offset: BufferAddress,
+        buffer_end_offset: BufferAddress,
+    },
+}
+
+impl<B: GfxBackend> QuerySet<B> {
+    fn validate_query(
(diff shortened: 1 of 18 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -3593,6 +3599,141 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+
+        let (device_guard, mut token) = hub.devices.read(&mut token);
+
+        let error = loop {
+            let device = match device_guard.get(device_id) {
+                Ok(device) => device,
+                Err(_) => break DeviceError::Invalid.into(),
+            };
+
+            if matches!(desc.ty, wgt::QueryType::Timestamp)
```
kvark, line 3621: please, let's not overuse the fancy stuff. It's clear that you just need one good `match desc.ty` here, not a couple of match macro invocations

wgpu-core/src/hub.rs
```diff
@@ -264,6 +266,12 @@ impl<B: hal::Backend> Access<ComputePipeline<B>> for BindGroup<B> {}
 impl<B: hal::Backend> Access<RenderPipeline<B>> for Device<B> {}
 impl<B: hal::Backend> Access<RenderPipeline<B>> for BindGroup<B> {}
 impl<B: hal::Backend> Access<RenderPipeline<B>> for ComputePipeline<B> {}
+impl<B: hal::Backend> Access<QuerySet<B>> for Root {}
+impl<B: hal::Backend> Access<QuerySet<B>> for Device<B> {}
+impl<B: hal::Backend> Access<QuerySet<B>> for Sampler<B> {}
+impl<B: hal::Backend> Access<QuerySet<B>> for CommandBuffer<B> {}
+impl<B: hal::Backend> Access<QuerySet<B>> for RenderPipeline<B> {}
+impl<B: hal::Backend> Access<QuerySet<B>> for ComputePipeline<B> {}
```
kvark, line 274: uh, that's quite a bit! can we reduce this set? I.e. can we make sure the query set is always locked after command buffers only?

wgpu-types/src/lib.rs
```diff
@@ -2458,3 +2462,71 @@ pub enum SamplerBorderColor {
...
+/// Type of query contained in a QuerySet.
+#[derive(Copy, Clone, Debug)]
+pub enum QueryType {
+    /// Query returns up to 5 64-bit numbers based on the given flags.
+    ///
+    /// See [`PipelineStatisticsTypes`]'s documentation for more information
+    /// on how they get resolved.
+    ///
+    /// [`Features::PIPELINE_STATISTICS_QUERY`] must be enabled to use this query type.
+    PipelineStatistics(PipelineStatisticsTypes),
```
kvark, line 2485: We must be careful here. The upstream API works with a list of types, and it makes it so the implementation has to write down the results according to the list.
In this PR it's a set, however, so what order do the values come in?

wgpu-core/src/device/mod.rs
```diff
@@ -3593,6 +3599,141 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+
+        let device = match device_guard.get(device_id) {
+            Ok(device) => device,
+            Err(_) => return Err(DeviceError::Invalid),
+        };
+
+        Ok(device.private_features.timestamp_period)
+    }
+
+    pub fn query_set_drop<B: GfxBackend>(&self, query_set_id: id::QuerySetId) {
```
kvark, line 3716: I think a few cleanup stages might be missing. You can follow the bind groups for example to see where to put the logic:
  - updating the submission index of the query sets upon submission
  - when triaging suspected resources, checking for the refcount and moving to the free resources, or associated submissions
  - when considering a submission finished, moving the resources in to free resources
  - when going through the free resources, actually deleting the sets
  - when encoding any operations involving the sets, add them to the tracked list of the command buffer (I commented about it separately)

wgpu-types/src/lib.rs
cwfitzgerald: So this is one place that webgpu is a bit underspecified. While upstream is a list of types, all the underlying apis seem to do this according to the order in the bitflags from LSB to MSB. This implementation also follows that behavior. If webgpu wants us to also swizzle the outputs, this is beyond the scope of the initial implementation as it requires a shader.

I have documented all this on the PipelineStatisticsTypes bitflag documentation.

kvark: Yeah, the assumption was that a shader is needed regardless. If this no longer holds, we should reconsider the list there.

wgpu-core/src/command/query.rs
```diff
@@ -0,0 +1,284 @@
...
+        query_set_storage: &Storage<QuerySet<B>, id::QuerySetId>,
+    ) {
+        for (query_set_id, state) in self.map.into_iter() {
+            let query_set = query_set_storage.get(query_set_id).unwrap();
+
+            debug_assert_eq!(state.len(), query_set.desc.count as usize);
+            let count = state.len();
+
+            // Need to find all "runs" of values which need resets. If the state vector is:
+            // [false, true, true, false, true], we want to reset [1..3, 4..5]. This minimizes
```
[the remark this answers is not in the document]
cwfitzgerald: We reset before the renderpass as we cannot execute a reset inside a renderpass. By having this reset here, you can use the same query multiple times outside of a renderpass. This is another area webgpu is underspecified, so I went the most forgiving route.

wgpu-core/src/command/query.rs
```diff
@@ -0,0 +1,284 @@
...
+    ) -> bool {
+        let vec = self
+            .map
+            .entry(id)
+            .or_insert_with(|| vec![false; query_set.desc.count as usize]);
+
+        std::mem::replace(&mut vec[query as usize], true)
+    }
+
+    pub unsafe fn reset_queries<B: hal::Backend>(
```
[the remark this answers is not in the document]
cwfitzgerald: I'm just not confident enough that you couldn't safely cause this function to perform nonsense, especially when dealing directly with recording commands.

wgpu-core/src/command/query.rs
```diff
@@ -0,0 +1,284 @@
...
+        let (buffer_guard, _) = hub.buffers.read(&mut token);
+
+        let cmb_buf = CommandBuffer::get_encoder_mut(&mut cmb_buf_guard, command_encoder_id)?;
+        let cmb_buf_raw = cmb_buf.raw.last_mut().unwrap();
+        let query_set = query_set_guard.get(query_set_id).unwrap();
+
+        let (dst_buffer, dst_pending) = cmb_buf
+            .trackers
+            .buffers
+            .use_replace(&*buffer_guard, destination, (), BufferUse::COPY_DST)
```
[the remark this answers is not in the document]
cwfitzgerald: Afaik, a query resolve is, to the underlying apis, just a transfer.

wgpu-core/src/hub.rs
cwfitzgerald: I tried and managed to get rid of one of them, but doing that would just trade edges before the query set for edges after.

kvark: that's ok, I can try to follow-up and try to reduce this

cwfitzgerald:
Alright, this is still blocked on gfx updates and I need to add queries to traces, but the core changes are ready for second pass review.

cwfitzgerald:
Traces are added

wgpu-core/src/command/query.rs
```diff
@@ -0,0 +1,441 @@
...
+    reset_state: &mut Option<&mut QueryResetMap>,
+    invalid_query_set_fn: impl FnOnce(id::QuerySetId) -> E,
+) -> Result<(&'a QuerySet<B>, hal::query::Query<'a, B>), E>
+where
+    E: From<QueryUseError>,
+{
+    let query_set = tracker
+        .use_extend(storage, query_set_id, (), ())
+        .map_err(|e| match e {
+            UseExtendError::InvalidResource => invalid_query_set_fn(query_set_id),
```
kvark, line 174: I find the need to pass `invalid_query_set_fn` around to be unfortunate, and I'm thinking if we can get around it.
What if, instead of having `tracker`, `storage`, `query_set_id`, and `invalid_query_set_fn` parameters on a free function like this, we would do:
```rust
impl QuerySet {
  fn validate_query(&self, query_type, query_index, reset_state) -> Result<hal::query::Query<'a, B>,  QueryUseError> {...}
  fn write_timestamp(&self, ...) ...
  fn begin_pipeline_statistics_query(&self, ...) ...
  fn end_pipeline_statistics_query() ...
}
```

We'd then use the `tracker.use_extend` call in all the spots where this function is called, explicitly.

Benefits are:
  - less argument complexity on the functions, especially the 
  - easier to import `use crate::resource::QuerySet` instead of `use crate::resource::{fun_a, fun_b, fun_c};`
  - the calls to `use_extend` are more visible and consistent in the code (arguable, I guess?)

kvark (commented):
Please read all the comments before starting to address them :)
Thank you for your patience!

wgpu-core/src/command/query.rs
```diff
@@ -0,0 +1,441 @@
...
+    },
+}
+
+fn validate_query<'a, B: GfxBackend, E>(
+    tracker: &mut ResourceTracker<PhantomData<id::QuerySetId>>,
+    storage: &'a Storage<QuerySet<B>, id::QuerySetId>,
+    query_type: SimplifiedQueryType,
+    query_set_id: id::QuerySetId,
+    query_index: u32,
+    reset_state: &mut Option<&mut QueryResetMap>,
```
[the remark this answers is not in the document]
cwfitzgerald: Because I need to use the option later on as well. If I pass it by value I can't use it later in the function. I need to get (from the option) a `&mut &mut QueryResetMap`, which is only possible when the reference to the option is also mut.

wgpu-core/src/command/query.rs
```diff
@@ -0,0 +1,441 @@
...
+pub(super) fn validate_and_begin_pipeline_statistics_query<B: GfxBackend, E>(
+    cmd_buf_raw: &mut B::CommandBuffer,
+    tracker: &mut ResourceTracker<PhantomData<id::QuerySetId>>,
+    storage: &Storage<QuerySet<B>, id::QuerySetId>,
+    query_set_id: id::QuerySetId,
+    query_index: u32,
+    mut reset_state: Option<&mut QueryResetMap>,
+    active_query: &mut Option<(id::QuerySetId, u32)>,
+    invalid_query_set_fn: impl FnOnce(id::QuerySetId) -> E,
+) -> Result<(), E>
```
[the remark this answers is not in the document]
cwfitzgerald: Because QueryUseError doesn't have the invalid query set values, those are always dealt with at a layer above this one with all the other invalid errors.

cwfitzgerald:
Everything seems to work fine, so we're good to go :)

wgpu-core/src/command/query.rs
```diff
@@ -0,0 +1,424 @@
...
+    },
+}
+
+impl<B: GfxBackend> QuerySet<B> {
+    fn validate_query<'a>(
+        &'a self,
+        query_set_id: id::QuerySetId,
+        query_type: SimplifiedQueryType,
+        query_index: u32,
+        reset_state: &mut Option<&mut QueryResetMap<B>>,
```
kvark, line 173: so this isn't meant to change the `Option` itself, right?
In this case, we should get away with just `Option<&mut QueryResetMap<B>>` parameter. The only use by clients of this function is `if reset_state.is_none() {` condition, so instead of passing `&mut reset_state` here, they could do:
```rust
let need_reset = reset_state.is_none();
validate_query(..., reset_state);
if need reset {
   ...
}
```

wgpu-core/src/command/query.rs
```diff
@@ -0,0 +1,424 @@
...
+        if end_query > query_set.desc.count {
+            return Err(ResolveError::QueryOverrun {
+                start_query,
+                end_query,
+                query_set_size: query_set.desc.count,
+            }
+            .into());
+        }
+
+        let stride = query_set.elements * 8;
```
kvark, line 388: nit: can we have this as a constant somewhere?

wgpu-core/src/device/mod.rs
```diff
@@ -3673,6 +3679,132 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+                        );
+                    }
+                }
+            }
+
+            if desc.count == 0 {
+                break resource::CreateQuerySetError::ZeroCount;
+            }
+
+            if desc.count == wgt::QUERY_SET_MAX_QUERIES {
```
kvark, line 3725: this should probably be `>=` instead of `==`

kvark (commented):
Amazing work, the code looks great!
Just a few small nits, one of them (about the query size check) needs to be addressed, others up to your consideration.

cwfitzgerald:
Addressed; Thanks for the detailed reviews :)

bors r=kvark

PR #1132, opened 2021-01-05
Include pipeline id in some pass error scopes

scoopr:


scoopr:
Motivation was going from 
```
Caused by:
    In a RenderPass
      note: encoder = `frame encoder`
    In a draw_indexed command
    pipeline index format (Uint16) and buffer index format (Uint32) do not match
```
to
```
Caused by:
    In a RenderPass
      note: encoder = `frame encoder`
    In a draw_indexed command
      note: render pipeline = `render pipe`
    pipeline index format (Uint16) and buffer index format (Uint32) do not match
```

wgpu-core/src/command/mod.rs
```diff
@@ -361,19 +361,19 @@ pub enum PassErrorScope {
     #[error("In a set_scissor_rect command")]
     SetScissorRect,
     #[error("In a draw command")]
-    Draw,
+    Draw(Option<id::RenderPipelineId>),
```
kvark, line 364: Can we reduce the duplication here? Something like:
```rust
Draw {
  indexed: bool,
  indirect: bool,
  pipeline: Option<id::RenderPipelineId>,
},
```

kvark (commented):
Thank you, it's great to see errors improving!

wgpu-core/src/command/mod.rs
scoopr: hmm, that would make the type ever so slightly larger ;) 
I'd prefer to punt it to later

kvark: Does it matter how large the error type is? It's not like you'd be juggling it in perf-sensitive contexts

scoopr: No, that was in jest :)

scoopr:
Merged the variants,
now the error looks like:
```
Caused by:
    In a RenderPass
      note: encoder = `frame encoder`
      note: render pipeline = `render pipe`
    pipeline index format (Uint16) and buffer index format (Uint32) do not match
```

wgpu-core/src/command/compute.rs
```diff
@@ -459,7 +459,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                     }
                 }
                 ComputeCommand::DispatchIndirect { buffer_id, offset } => {
-                    let scope = PassErrorScope::DispatchIndirect;
+                    let scope = PassErrorScope::Dispatch{indirect:true, pipeline:state.pipeline.last_state};
```
kvark, line 462: did you run `cargo fmt` on this?

wgpu-core/src/command/render.rs
```diff
@@ -1619,9 +1627,18 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                         indexed,
                     } => {
                         let scope = if indexed {
-                            PassErrorScope::DrawIndexedIndirect
+                            PassErrorScope::Draw {
+                                indexed: true,
```
kvark, line 1631: now we can merge those two branches into one:
```rust
let scope = PassErrorScope::Draw {
  indexed,
  indirect: true,
  pipeline: state.pipeline.last_state,
};
```

kvark (commented):
A few nits, otherwise looks great!

wgpu-core/src/command/compute.rs
scoopr: I did, but I guess I noticed the dispatch stuff after that, oops.

wgpu-core/src/command/render.rs
scoopr: oh duh, of course :)

kvark (approved):
Great!
bors r+

PR #1134, opened 2021-01-06
Never prefer CPU devices over GPUs

myfreeweb:
**Connections**
Rebase of #969 onto master

**Description**
See #969


kvark (approved):
Thank you!
bors r+

PR #1135, opened 2021-01-08
Make all textures support usage COPY_SRC and COPY_DST

fintelia:
**Connections**
N/A

**Description**
In WebGPU, all textures should be allowed to have usages COPY_SRC and COPY_DST. However, right now when Features::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES is enabled any features that the backend says can't be blitted to/from will not gain these capabilities. In particular, compressed textures may not support blitting while still allowing access via copies to/from buffers.

You can also see the defaults used when TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES is not requested:
https://github.com/gfx-rs/wgpu/blob/8c00cbcc8e8b955477f2df29b4b0a20712fdef78/wgpu-types/src/lib.rs#L1224-L1342

**Testing**
I ran a [patched version of the wgpu-rs cube example](https://github.com/fintelia/wgpu-rs/tree/cube-bc4/examples/cube).

### DIFF
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -263,7 +263,7 @@ impl<B: GfxBackend> Adapter<B> {
             )))
             .optimal_tiling;
 
-        let mut allowed_usages = wgt::TextureUsage::empty();
+        let mut allowed_usages = format.describe().guaranteed_format_features.allowed_usages;
         if texture_format_properties.contains(hal::format::ImageFeature::SAMPLED) {
             allowed_usages |= wgt::TextureUsage::SAMPLED;
         }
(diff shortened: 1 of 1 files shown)


wgpu-core/src/instance.rs
```diff
@@ -263,7 +263,7 @@ impl<B: GfxBackend> Adapter<B> {
             )))
             .optimal_tiling;
 
-        let mut allowed_usages = wgt::TextureUsage::empty();
+        let mut allowed_usages = wgt::TextureUsage::COPY_SRC | wgt::TextureUsage::COPY_DST;
```
kvark, line 266: Can we initialize with the base texture usage that is guaranteed (i.e. `guaranteed_format_features.allowed_usages`), instead of this hard-coded value?

kvark (commented):
Thank you for the PR!

kvark (approved):
We'll need to come back to this and make COPY vs BLIT distinction more clear... I filed https://github.com/gfx-rs/gfx/issues/3575 to follow-up on that.

Wumpf:
oh, got that wrong when doing `TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES`. Nice catch @fintelia :)

PR #1136, opened 2021-01-09
Changed Arc<B::Memory> to Arc<Mutex<B::Memory>> in GpuAllocator

ElArtista:
Work around for blocker https://github.com/zakarumych/gpu-alloc/issues/33, makes wgpu compatible with latest gfx


kvark (commented):
Very nice, thank you!
bors r+

PR #1142, opened 2021-01-11
Move get_swap_chain_preferred_format from device to adapter and check formats

niklaskorz:
**Connections**

https://github.com/gfx-rs/wgpu-rs/issues/123#issuecomment-757831458

**Description**

Not all devices actually support `Bgra8UnormSrgb`. If this is the case, a similar format like `Rgba8UnormSrgb` can be used.

**Testing**

https://github.com/niklaskorz/wgpu/blob/e5ebfa70a4416c1b579d32566b5c1c8190dac189/learn-wgpu/src/lib.rs#L137

- Output on MacBook Pro 13" Intel 2020 macOS Metal: `Format: Bgra8UnormSrgb`
- Output on OnePlus 7 Pro Android Vulkan: `Format: Rgba8UnormSrgb`

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -8,7 +8,7 @@ use crate::{
     hub::{
         GfxBackend, Global, GlobalIdentityHandlerFactory, Hub, Input, InvalidId, Storage, Token,
     },
-    id, pipeline, resource, span, swap_chain,
+    id, instance, pipeline, resource, span, swap_chain,
     track::{BufferState, TextureSelector, TextureState, TrackerSet},
     validation::{self, check_buffer_usage, check_texture_usage},
     FastHashMap, FastHashSet, Label, LabelHelpers, LifeGuard, MultiRefCount, PrivateFeatures,
@@ -2379,6 +2379,28 @@ pub struct ImplicitPipelineIds<'a, G: GlobalIdentityHandlerFactory> {
 }
 
 impl<G: GlobalIdentityHandlerFactory> Global<G> {
+    pub fn adapter_get_swap_chain_preferred_format<B: GfxBackend>(
+        &self,
+        adapter_id: id::AdapterId,
+        surface_id: id::SurfaceId,
+    ) -> Result<TextureFormat, instance::GetSwapChainPreferredFormatError> {
+        span!(_guard, INFO, "Adapter::get_swap_chain_preferred_format");
+
+        let hub = B::hub(self);
+        let mut token = Token::root();
+
+        let (mut surface_guard, mut token) = self.surfaces.write(&mut token);
+        let (adapter_guard, mut _token) = hub.adapters.read(&mut token);
+        let adapter = adapter_guard
+            .get(adapter_id)
+            .map_err(|_| instance::GetSwapChainPreferredFormatError::InvalidAdapter)?;
+        let surface = surface_guard
+            .get_mut(surface_id)
+            .map_err(|_| instance::GetSwapChainPreferredFormatError::InvalidSurface)?;
+
+        adapter.get_swap_chain_preferred_format(surface)
+    }
+
     pub fn device_features<B: GfxBackend>(
         &self,
         device_id: id::DeviceId,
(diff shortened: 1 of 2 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -3887,12 +3897,46 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+            }
+            // If not available, use Rgba8UnormSrgb
+            if formats.contains(&conv::map_texture_format(TextureFormat::Rgba8UnormSrgb, device.private_features)) {
+                return Ok(TextureFormat::Rgba8UnormSrgb);
+            }
+            // TODO: are there devices that support neither of these?
+        }
+
+        // If no formats were returned, use Bgra8UnormSrgb
         Ok(TextureFormat::Bgra8UnormSrgb)
```
kvark, line 3940: we'd also hit this path if the formats are returned, but neither of the branches was taken

wgpu-core/src/device/mod.rs
```diff
@@ -3887,12 +3897,46 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
 
     pub fn device_get_swap_chain_preferred_format<B: GfxBackend>(
```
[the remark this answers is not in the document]
niklaskorz: Sure!

wgpu-core/src/device/mod.rs
niklaskorz: Do you think it'd be better to return an error in this case? Or just take the first format in the list, even if it's a single-color format?

kvark: let's take the first format, still better than an error :)

niklaskorz: I had another look at the definition of `getSwapChainPreferredFormat(adapter)` in the WebGPU spec, where it says:

> Return an optimal GPUTextureFormat to use when creating a GPUSwapChain with the given adapter. Must be one of the supported swap chain formats.
> The supported swap chain formats are a set of GPUTextureFormats that must be supported when specified as a GPUSwapChainDescriptor.format regardless of the given GPUSwapChainDescriptor.device, initially set to: «"bgra8unorm", "bgra8unorm-srgb", "rgba8unorm", "rgba8unorm-srgb"».

So I guess it is sufficient to check for these four, and then fallback to the first format in the list, in the unlikely case that none of these is supported.

niklaskorz: On the other hand, there are more formats described in HAL than there are texture formats defined in wgpu. So for now, I think checking these four mentioned in the spec and returning an error instead might be the easier choice until someone comes up with a use case where these four don't work.

wgpu-core/src/instance.rs
```diff
@@ -250,6 +250,44 @@ impl<B: GfxBackend> Adapter<B> {
         }
     }
 
+    pub fn get_swap_chain_preferred_format(
+        &self,
+        surface: &mut Surface,
```
kvark, line 255: does this need to be mutable?

wgpu-core/src/instance.rs
```diff
@@ -250,6 +250,44 @@ impl<B: GfxBackend> Adapter<B> {
         }
     }
 
+    pub fn get_swap_chain_preferred_format(
```
kvark, line 253: we'd also need a method on the `Global` object that takes an `AdapterId` and `SurfaceId` and returns the same result

wgpu-core/src/instance.rs
niklaskorz: Not necessarily, but I only found the `get_surface_mut` function for getting the gfx-hal surface of a wgpu surface. If there's an immutable alternative, I'd be glad to use it.

kvark: that function doesn't force you to accept `&mut` though, you can still have a `&` argument

niklaskorz: Not sure if I'm missing something, but get_surface_mut only appears to take &mut as input. How would I take an immutable reference and pass it as a mutable reference if the type is not copyable?

kvark: oh, sorry, I see now :)

kvark (approved):

kvark:
Please squash and make this a non-draft. I think we are good to go!

niklaskorz:
Checking first if it actually works as intended on Android Vulkan, should be ready in a moment :)

kvark:
thank you!
bors r+

PR #1144, opened 2021-01-11
Fix and simplify attachment dimension mismatch check

DasEtwas:
This also improves error messages with detailed dimension and attachment type name info which developers can hopefully recognize.

**Connections**
None

**Description**
Before this commit, depth/stencil attachments are not considered when evaluating the renderpasses' dimension, which will result in a "NoAttachment" error with a confusing display message.

**Testing**
Tested by running an application making use of color-only and depth/stencil-only attachment renderpasses.

Tested by running boids, hello-triangle and shadow examples.

### DIFF
--- a/wgpu-core/src/command/render.rs
+++ b/wgpu-core/src/command/render.rs
@@ -387,10 +387,22 @@ pub enum RenderPassErrorInner {
     Encoder(#[from] CommandEncoderError),
     #[error("attachment texture view {0:?} is invalid")]
     InvalidAttachment(id::TextureViewId),
-    #[error("there are attachments")]
-    NoAttachments,
-    #[error("attachments have different sizes")]
-    MismatchAttachments,
+    #[error("necessary attachments are missing")]
+    MissingAttachments,
+    #[error("color and/or depth attachments have differing sizes: a {mismatching_attachment_type_name} attachment of size {mismatching_dimensions_width}x{mismatching_dimensions_height} is not compatible with a {previous_attachment_type_name} attachment's size ({previous_dimensions_height}x{previous_dimensions_width})",
+        previous_dimensions_width = previous_dimensions.0,
+        previous_dimensions_height = previous_dimensions.1,
+        mismatching_dimensions_width =  mismatching_dimensions.0,
+        mismatching_dimensions_height = mismatching_dimensions.1,
+    )]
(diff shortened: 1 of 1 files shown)


wgpu-core/src/command/render.rs
```diff
@@ -387,10 +387,22 @@ pub enum RenderPassErrorInner {
     Encoder(#[from] CommandEncoderError),
     #[error("attachment texture view {0:?} is invalid")]
     InvalidAttachment(id::TextureViewId),
-    #[error("there are attachments")]
-    NoAttachments,
-    #[error("attachments have different sizes")]
-    MismatchAttachments,
+    #[error("necessary attachments are missing")]
+    MissingAttachments,
+    #[error("color and/or depth attachments have differing sizes: a {mismatching_attachment_type_name} attachment of size {mismatching_dimensions_width}x{mismatching_dimensions_height} is not compatible with a {previous_attachment_type_name} attachment's size ({previous_dimensions_height}x{previous_dimensions_width})",
```
kvark, line 392: just printing the tuple in the error would be nearly as clear, and would save us a few keystrokes here. Perhaps, we can avoid those `previous_dimensions_width = ...` defines, and just print the whole tuples?

wgpu-core/src/command/render.rs
```diff
@@ -526,9 +538,12 @@ impl<'a, B: GfxBackend> RenderPassInfo<'a, B> {
 
         let mut render_attachments = AttachmentDataVec::<RenderAttachment>::new();
 
+        // need to be initialized to None!, check `AttachmentsDimensionMismatch` error
         let mut attachment_width = None;
         let mut attachment_height = None;
-        let mut valid_attachment = true;
+        let mut attachment_type_name = None;
```
kvark, line 544: could we just initialize it to `""`? the option isn't helping here, I think

kvark (commented):
Very nice! just a few small suggestions here

DasEtwas:
Applied the suggested changes

kvark (approved):
Thank you!
bors r+

kvark:
Note: I'm refactoring this code a bit in https://github.com/gfx-rs/wgpu/pull/1147/commits/a4f897b5d78268cfab1250a24088415baf69dad1

PR #1147, opened 2021-01-13
Update render pass and framebuffers logic to the latest gfx-hal

kvark:
**Connections**
Updates gfx-hal with the latest major changes of https://github.com/gfx-rs/gfx/pull/3571

**Description**
There is no need to track framebuffers at all any more! Since they don't contain images, we can keep them permanently.
There is no need to create and delete framebuffers for the swapchain either. Lots of non-trivial code is gone 🎉
The PR also removes one of the mutexes that the device holds. 

**Testing**
Will be tested on wgpu-rs


scoopr:
Out of curiosity, tried this, and in my project I get part of the viewport cut off after resizing the window to "extreme-ish" aspect ratio. This doesn't happen with the examples.

I did a xcode frame cap, and when the view is cut off, I see there is a `setScissorRect:` call, that doesn't happen with "normal-ish" aspect ratio or master, and the resolution doesn't match what it sets the viewport to.

I wonder if it is the `clamp_scissor` in gfx-hal clamping to against wrong extent or something?

The behaviour seems to be if width is less than some fixed size, then height is cut to that fixed size, and vice versa, if height is less than some fixed size, width is cut at some fixed size.

kvark:
Wow, @scoopr that's amazing feedback! I'll check it out.

kvark:
Everything is fixed, and even better than before!
@cwfitzgerald and/or @grovesNL - please have a look. This is a bit of a complex change.

Cargo.toml
```diff
@@ -6,7 +6,7 @@ members = [
     "wgpu-types",
 ]
 
-#[patch."https://github.com/gfx-rs/gfx"]
+[patch."https://github.com/gfx-rs/gfx"]
```
cwfitzgerald, line 9: I think this slipped in

cwfitzgerald (approved):
Minor nit, but otherwise LGTM!

Cargo.toml
kvark: that was intentional, actually. Having a patch section without contents isn't doing anything, but enabling/disabling overrides is easier

kvark:
Thank you for reviewing!
bors r=cwfitzgerald

Cargo.toml
cwfitzgerald: Oh I didn't realize that was valid, nice

Issue #1143, opened 2021-01-11
Add SHADER_FLOAT64 native feature


fintelia:
**Is your feature request related to a problem? Please describe.**
For most computer graphics use cases 32-bit floats work well enough, but there are specific situations where they don't. For example, the earth has a radius of about 6000km so if the origin were placed at the center of the planet, positions on the surface can only be expressed with 50cm precision. There are many ways around these sorts of limitations, but the simplest all rely on being able to do at least a small amount of 64-bit float computations on the GPU. 

Currently when using SPIR-V shaders, wgpu does allow usage of 64-bits floats ("doubles") provided that the underlying driver also does. Unfortunately there is no way to probe whether this will work and this may stop being supported once shader validation is fully implemented.

**Describe the solution you'd like**
Add a SHADER_FLOAT64 native feature to mirror the one exposed by gfx-hal.

**Describe alternatives you've considered**
A software float64 implementation would enable this feature to be implemented everywhere (which sounds somewhat ridiculous, but is actually the strategy that Mesa uses).

DasEtwas:
Regarding planet rendering precision issues, one would rather resort to using a sparse representation of terrain (chunks) whose individual transforms are calculated using double precision relative to the camera. Then this transform is converted to single floats. Thus, a view matrix with large translations can be omitted and precision is just fine for small enough chunks (<10km). Double floats can have significant performance impacts on many cards (testing needed).

fintelia:
You are correct that double precision is often much slower than single precision. Looking at the GFLOPS numbers from [here](https://www.techpowerup.com/gpu-specs/) the factor seems to be 16x slower for AMD and 32-64x for NVIDIA. However, that still translates to several hundred billion floating point operations per second. You can fit a lot of work into that budget: it is enough to do the chunk generation on the GPU (my preferred solution) or even just the vertex transformations themselves.

PR #1154, opened 2021-01-18
Add SHADER_FLOAT64 feature

fintelia:
**Connections**
Fixes: #1143 

**Description**
This patch enables using 64-bit floats in shaders by requesting the relevant feature from gfx-hal.

**Testing**
I verified that the Adapter claims support when running on Mesa and not when running with swiftshader.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -1939,6 +1939,19 @@ impl<B: GfxBackend> Device<B> {
                         },
                     );
                 }
+
+                if let wgt::VertexFormat::Double
+                | wgt::VertexFormat::Double2
+                | wgt::VertexFormat::Double3
+                | wgt::VertexFormat::Double4 = attribute.format
+                {
+                    if !self.features.contains(wgt::Features::VERTEX_ATTRIBUTE_64BIT) {
+                        return Err(pipeline::CreateRenderPipelineError::MissingFeature(
+                            wgt::Features::VERTEX_ATTRIBUTE_64BIT,
+                        ));
+                    }
+                }
+
                 attributes.alloc().init(hal::pso::AttributeDesc {
                     location: attribute.shader_location,
                     binding: i as u32,
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -439,6 +439,11 @@ impl<B: GfxBackend> Adapter<B> {
             desc.features
                 .contains(wgt::Features::PIPELINE_STATISTICS_QUERY),
         );
+        enabled_features.set(
+            hal::Features::SHADER_FLOAT64,
+            desc.features
+                .contains(wgt::Features::SHADER_FLOAT64),
+        );
 
         let family = self
             .raw
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -441,8 +441,7 @@ impl<B: GfxBackend> Adapter<B> {
         );
         enabled_features.set(
             hal::Features::SHADER_FLOAT64,
-            desc.features
-                .contains(wgt::Features::SHADER_FLOAT64),
+            desc.features.contains(wgt::Features::SHADER_FLOAT64),
         );
 
         let family = self
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -188,6 +188,10 @@ impl<B: GfxBackend> Adapter<B> {
             wgt::Features::PIPELINE_STATISTICS_QUERY,
             adapter_features.contains(hal::Features::PIPELINE_STATISTICS_QUERY),
         );
+        features.set(
+            wgt::Features::SHADER_FLOAT64,
+            adapter_features.contains(hal::Features::SHADER_FLOAT64),
+        );
         #[cfg(not(target_os = "ios"))]
         //TODO: https://github.com/gfx-rs/gfx/issues/3346
         features.set(wgt::Features::ADDRESS_MODE_CLAMP_TO_BORDER, true);
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -390,6 +390,9 @@ bitflags::bitflags! {
         const TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES = 0x0000_0000_1000_0000;
         /// Enables 64-bit floating point types in SPIR-V shaders.
         ///
+        /// Note: even when supported by GPU hardware, 64-bit floating point operations are
+        /// frequently between 16 and 64 _times_ slower than equivelent operations on 32-bit floats.
+        ///
         /// Supported Platforms:
         /// - Vulkan
         ///
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -388,6 +388,13 @@ bitflags::bitflags! {
         ///
         /// This is a native-only feature.
         const TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES = 0x0000_0000_1000_0000;
+        /// Enables 64-bit floating point types in SPIR-V shaders.
+        ///
+        /// Supported Platforms:
+        /// - Vulkan
+        ///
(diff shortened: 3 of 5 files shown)


wgpu-core/src/instance.rs
```diff
@@ -188,6 +188,10 @@ impl<B: GfxBackend> Adapter<B> {
             wgt::Features::PIPELINE_STATISTICS_QUERY,
             adapter_features.contains(hal::Features::PIPELINE_STATISTICS_QUERY),
         );
+        features.set(
+            wgt::Features::SHADER_FLOAT64,
```
cwfitzgerald, line 192: This also needs code to enable the hal feature when creating the device.

wgpu-types/src/lib.rs
```diff
@@ -388,6 +388,13 @@ bitflags::bitflags! {
         ///
         /// This is a native-only feature.
         const TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES = 0x0000_0000_1000_0000;
+        /// Enables 64-bit floating point types in SPIR-V shaders.
+        ///
+        /// Supported Platforms:
+        /// - Vulkan
```
cwfitzgerald, line 394: What is causing this to be limited to only vulkan? Dx12/metal should support it i believe.

cwfitzgerald (changes requested):

wgpu-types/src/lib.rs
```diff
@@ -388,6 +388,13 @@ bitflags::bitflags! {
         ///
         /// This is a native-only feature.
         const TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES = 0x0000_0000_1000_0000;
+        /// Enables 64-bit floating point types in SPIR-V shaders.
```
cwfitzgerald, line 391: This really should carry a warning about the anemic throughout our doubles in shaders. If people are used to the float perf, they're going to get burned here.

wgpu-core/src/instance.rs
cwfitzgerald: https://github.com/gfx-rs/wgpu/blob/master/wgpu-core/src/instance.rs#L381

kvark:
Thank you for the PR!
My main concern is how we are going to deal with vertex attributes that are `f64`. Do we want to make it a part of this feature, or a separate feature, and if so, why?

wgpu-types/src/lib.rs
fintelia: This is just based on which hal backends expose support for the feature. There might also be limitations from the shader translation, but that I'm less sure of

fintelia: Though looking at the Metal Shading Language Specification:
> Metal does not support the double, long long, unsigned long long, and long double data types

fintelia:
> My main concern is how we are going to deal with vertex attributes that are f64. Do we want to make it a part of this feature, or a separate feature, and if so, why?

I don't feel strongly either way. Is there hardware that allows f64 computations in shaders but doesn't allow passing them as vertex attributes? I would think that as long as f64's are allowed in uniform/storage buffers, then it doesn't make much difference whether they can be used as vertex attributes as well.

kvark:
> > My main concern is how we are going to deal with vertex attributes that are f64. Do we want to make it a part of this feature, or a separate feature, and if so, why?
> 
> I don't feel strongly either way. Is there hardware that allows f64 computations in shaders but doesn't allow passing them as vertex attributes? I would think that as long as f64's are allowed in uniform/storage buffers, then it doesn't make much difference whether they can be used as vertex attributes as well.

In this case, let's also add those vertex formats in this PR, gated at runtime by the feature. Unless you want to leave this for later work?

kvark (approved):
Great!
@cwfitzgerald are you concerns addressed?
Let's get this in when ready.

cwfitzgerald:
From a code standpoint this looks fine. My one concern is that DX11/12 don't support f64 vertex attributes but _do_ support f64 in shaders. This may work for vk, but we'd be in a weird situation for DX11/12.

kvark:
@cwfitzgerald interesting! So in the worst case, we'd have gfx-rs emitting the code in the shaders to load these as `u32` and unpack in code. But we'd still be able to support them.

wgpu-core/src/device/mod.rs
```diff
@@ -1939,6 +1939,19 @@ impl<B: GfxBackend> Device<B> {
                         },
                     );
                 }
+
+                if let wgt::VertexFormat::Double
```
kvark, line 1943: Given the feedback from @cwfitzgerald , we can't have these formats exposed behind the feature. We'll either need to remove them, or to have a separate issue for them specifically (something like `VERTEX_ATTRIBUTE_64BIT`), that potentially aggregates all of u64, i64, and f64 that we'll expose.
How do you feel about this? Sorry for turning back here, my initial suggestion was wrong.

fintelia: Either way seems fine to me

cwfitzgerald: Then splitting into two features seems like the way to go, that way we can have both things possible in wgpu and won't have to revisit this later.

fintelia: How would we detect whether the backend supports 64-bit vertex attributes?

kvark: We'd need to call [format_properties](https://github.com/gfx-rs/gfx/blob/00285817af676e091ac886b927e9e5e292f22a03/src/hal/src/adapter.rs#L103) on each of the vertex formats exposed by this potential feature, and ensure that all of them have [BufferFeature::VERTEX](https://github.com/gfx-rs/gfx/blob/00285817af676e091ac886b927e9e5e292f22a03/src/hal/src/format.rs#L194). Saying that, this will be some new code, and it's totally fine if you just make it report as unsupported in this PR, instead. Up to you :)

fintelia: I think I'll just add the feature but leave it unsupported then

wgpu-core/src/device/mod.rs
```diff
@@ -1939,6 +1939,24 @@ impl<B: GfxBackend> Device<B> {
...
+                | wgt::VertexFormat::Double2
+                | wgt::VertexFormat::Double3
+                | wgt::VertexFormat::Double4 = attribute.format
+                {
+                    if !self.features.contains(wgt::Features::VERTEX_ATTRIBUTE_64BIT) {
+                        return Err(pipeline::CreateRenderPipelineError::MissingFeature(
+                            wgt::Features::VERTEX_ATTRIBUTE_64BIT,
+                        ));
+                    }
+                    if !self.features.contains(wgt::Features::SHADER_FLOAT64) {
```
kvark, line 1953: not necessary, the `VERTEX_ATTRIBUTE_64BIT` feature should require this one, which is to be checked at device creation, not here

wgpu-types/src/lib.rs
```diff
@@ -1680,6 +1696,14 @@ pub enum VertexFormat {
     Int3 = 28,
     /// Four signed ints (i32). `ivec4` in shaders.
     Int4 = 29,
+    /// One double-precision float (f64). `double` in shaders. Requires SHADER_FLOAT64 and VERTEX_ATTRIBUTE_64BIT features.
```
kvark, line 1699: similarly, let's adjust the comment to only mention `VERTEX_ATTRIBUTE_64BIT` here

wgpu-types/src/lib.rs
```diff
@@ -388,6 +388,22 @@ bitflags::bitflags! {
...
+        /// Supported Platforms:
+        /// - Vulkan
+        ///
+        /// This is a native-only feature.
+        const SHADER_FLOAT64 = 0x0000_0000_2000_0000;
+        /// Enables using 64-bit types for vertex attributes.
+        ///
+        /// Supported Platforms: N/A
+        ///
+        /// This is a native-only feature.
```
kvark, line 405: here we can mention that it requires `SHADER_FLOAT64`

kvark (commented):
One last thing!

kvark (approved):

PR #1155, opened 2021-01-18
fix the size checks in transfer buffer to image commands

Vengarioth:
**Connections**
_none_

**Description**
It's fixing a trivial bug in which Extend3d was checked against a with of 0 two times, presumably instread checking depth as well.

**Testing**
Ran the tests and ran it with bevy - still worked.


cwfitzgerald:
Oh no, thanks for catching that!

bors r+

PR #1157, opened 2021-01-19
Handle read-only storage + sampled combination

### LINKED ISSUE #724: Image layout consistency between STORAGE and SAMPLED
I just merged an updated RODS image layout scheme in #723 and I realized we may have a larger problem with layouts. Imagine that a texture is used as both `SAMPLED` and `STORAGE_READ` in a render pass (i.e. inside a sync scope). The only layout compatible with with both is `General`, and that's what our state-mapping logic figures out today. HOWEVER:
  1. we don't know that when creating bind groups. We need to always pick `General` for sampled layouts if the texture *can* be a storage, even if user never uses it as storage
  2. we don't get that when the user only uses something as `SAMPLED`. Again, we'd have to make the usage tracking smarter here, and understand that some views/textures are always `General` layout.

**Repro steps**
In a render pass, bind 2 different views of the same texture subresource:
  - as `StorageTexture { readonly: true }`
  - as `SampledTexture`

Expected - `General` image layout is used for this texture.
Observerd - a mess...

kvark:
**Connections**
Fixes #724

**Description**
We now keep the internal use flags for sampled textures inside the view.

**Testing**
Untested

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -834,6 +834,12 @@ impl<B: GfxBackend> Device<B> {
             },
             samples: texture.kind.num_samples(),
             framebuffer_attachment: texture.framebuffer_attachment.clone(),
+            // once a storage - forever a storage
+            sampled_internal_use: if texture.usage.contains(wgt::TextureUsage::STORAGE) {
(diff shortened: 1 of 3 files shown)


cwfitzgerald (approved):
LGTM

bors r+

Issue #563, opened 2020-04-08
Clear or validate resources to be initialized


kvark:
The upstream spec requires us to zero-initialize resources: buffers and textures. I'm not a fan of that personally, but we need to validate the initialization regardless. Currently we aren't validating that. Example complaint:
> [2020-04-08T16:09:02Z TRACE wgpu_core::command::render] Encoding render pass begin in command buffer (0, 1, Vulkan)
[2020-04-08T16:09:02Z WARN  gfx_backend_vulkan] 
    VALIDATION [UNASSIGNED-CoreValidation-DrawState-InvalidRenderpass (0)] : Render pass has an attachment with loadOp == VK_ATTACHMENT_LOAD_OP_LOAD and initialLayout == VK_IMAGE_LAYOUT_UNDEFINED.  This is probably not what you intended.  Consider using VK_ATTACHMENT_LOAD_OP_DONT_CARE instead if the image truely is undefined at the start of the render pass.

The desired behavior here is to lazy-initialize resources that the user doesn't properly initialize, and issue a warning otherwise.

Wumpf:
I might be interested in solving this :)

> The desired behavior here is to lazy-initialize resources that the user doesn't properly initialize, and issue a warning otherwise.

So the idea is that the zero initialize is explicitly discouraged and logs a warning? I was actually hoping to rely on in it for a few things - I have some awkward "if this is first frame then read zero instead" things so far since there's no general clearbuffer/texture exposed :/

Besides this sentiment, there's a couple of cases where I'd argue it's too hard to avoid this warning since I believe there is only two ways to transparently clear a texture/buffer:
* fill it entirely on creation
* clear on start of a renderpass

(binding for write doesn't help because shader may not actually write any data)

This is a bit problematic for a wide range of usecases. E.g. a large voxel volume that is to be filled with mesh data at runtime has no way of clearing to zero other than creating a large zero sized array and "uploading" that on creation. This is unfortunate since inside the wgpu implementation we could call `clear_image`/`fill_buffer` which are much cheaper & easier but sadly not exposed in WebGPU.

kvark:
Upstream requires this to be the desired behavior, so if you are interested in implementing it this way, without any warnings, we'd be totally cool, and much appreciated! This is one of the core bits of the API that we haven't progressed with, and it's getting needed more and more.

[...]

PR #1159, opened 2021-01-19
Zero initialize buffers

Wumpf:
**Connections**
First half of  #563, focusing solely on buffers and ignoring same issue for textures

**Description**
Tracks for each buffer which parts are initialized (i.e. set to zero). Identified three interaction points for this:
* buffer mapping: Zeroing out ad-hoc and marking as initialized
* queue write to buffer: Marks target buffer regions as initialized (i.e. optimizes away buffer init down the line)
* use in binding or copy operation in a command buffer:
   * fine grained tracking of regions that may require init (or will be initialized implicitly) on each command buffer
   * set in motion on queue submit, init is exclusively with `fill_buffer`

Todo list for Ready-to-Review

- [x] memory barriers for `fill_buffer` calls
- [x] better data structure for `memory_init_tracker`
- [x] coarse filtering on command-buffer buffer init requirements (the list should almost always be empty whereas now it pushes any buffer use)
- [x] improve naming of things
- [x] at least pretend this is adequately tested

Todo list beyond this PR

* make data structures usable for textures
  * and.. well.. implement all this for textures!
* explore reusing barrier tracker for memory init tracking?

**Testing**

* Some basic testing by doing some changes to wgpu-rs samples and watching them in in the debugger.
* Added a ron test file for the player (love those!) to poke the feature a little bit
* MemoryInitTracker comes with simple unit tests

Overall this is a bit shallow but as so often in this area accurate testing is hard because the outcomes are quite indirect

### DIFF
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -206,8 +206,7 @@ impl RenderBundleEncoder {
                         .map_pass_err(scope);
                     }
 
-                    buffer_memory_init_actions
-                        .extend(bind_group.used_buffer_ranges.iter().map(|x| x.clone()));
+                    buffer_memory_init_actions.extend_from_slice(&bind_group.used_buffer_ranges);
 
                     state.set_bind_group(index, bind_group_id, bind_group.layout_id, offsets);
                     state
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -49,7 +49,7 @@ use crate::{
     },
     hub::{GfxBackend, GlobalIdentityHandlerFactory, Hub, Resource, Storage, Token},
     id,
-    memory_init_tracker::{MemoryInitTrackerAction, ResourceMemoryInitTrackerAction},
+    memory_init_tracker::{MemoryInitKind, ResourceMemoryInitTrackerAction},
     resource::BufferUse,
     span,
     track::{TrackerSet, UsageConflict},
@@ -269,7 +269,8 @@ impl RenderBundleEncoder {
                     };
                     used_buffer_ranges.push(ResourceMemoryInitTrackerAction {
                         id: buffer_id,
-                        action: MemoryInitTrackerAction::NeedsInitializedMemory(offset..end),
+                        range: offset..end,
+                        kind: MemoryInitKind::NeedsInitializedMemory,
                     });
                     state.index.set_format(index_format);
                     state.index.set_buffer(buffer_id, offset..end);
@@ -295,7 +296,8 @@ impl RenderBundleEncoder {
                     };
                     used_buffer_ranges.push(ResourceMemoryInitTrackerAction {
                         id: buffer_id,
-                        action: MemoryInitTrackerAction::NeedsInitializedMemory(offset..end),
+                        range: offset..end,
+                        kind: MemoryInitKind::NeedsInitializedMemory,
                     });
                     state.vertex[slot as usize].set_buffer(buffer_id, offset..end);
                 }
@@ -411,7 +413,8 @@ impl RenderBundleEncoder {
 
                     used_buffer_ranges.push(ResourceMemoryInitTrackerAction {
                         id: buffer_id,
-                        action: MemoryInitTrackerAction::NeedsInitializedMemory(0..buffer.size),
+                        range: 0..buffer.size,
+                        kind: MemoryInitKind::NeedsInitializedMemory,
                     });
 
                     commands.extend(state.flush_vertices());
@@ -441,7 +444,8 @@ impl RenderBundleEncoder {
                     let stride = 4 * 4; // 4 integers, vertexCount, instanceCount, firstVertex, firstInstance
                     used_buffer_ranges.push(ResourceMemoryInitTrackerAction {
                         id: buffer_id,
-                        action: MemoryInitTrackerAction::NeedsInitializedMemory(0..stride),
+                        range: 0..stride,
+                        kind: MemoryInitKind::NeedsInitializedMemory,
                     });
 
                     commands.extend(state.index.flush());
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -57,7 +57,7 @@ use crate::{
     Label, LabelHelpers, LifeGuard, Stored, MAX_BIND_GROUPS,
 };
 use arrayvec::ArrayVec;
-use std::{borrow::Cow, iter, ops::Range};
+use std::{borrow::Cow, iter, mem, ops::Range};
 use thiserror::Error;
 
 /// Describes a [`RenderBundleEncoder`].
@@ -393,7 +393,7 @@ impl RenderBundleEncoder {
                 }
                 RenderCommand::MultiDrawIndirect {
                     buffer_id,
-                    offset: _,
+                    offset,
                     count: None,
                     indexed: false,
                 } => {
@@ -410,11 +410,18 @@ impl RenderBundleEncoder {
                     check_buffer_usage(buffer.usage, wgt::BufferUsage::INDIRECT)
                         .map_pass_err(scope)?;
 
-                    buffer_memory_init_actions.push(MemoryInitTrackerAction {
-                        id: buffer_id,
-                        range: 0..buffer.size,
-                        kind: MemoryInitKind::NeedsInitializedMemory,
-                    });
+                    buffer_memory_init_actions.extend(
+                        buffer
+                            .initialization_status
+                            .check(
+                                offset..(offset + mem::size_of::<wgt::DrawIndirectArgs>() as u64),
+                            )
+                            .map(|range| MemoryInitTrackerAction {
+                                id: buffer_id,
+                                range,
+                                kind: MemoryInitKind::NeedsInitializedMemory,
+                            }),
+                    );
 
                     commands.extend(state.flush_vertices());
                     commands.extend(state.flush_binds());
@@ -422,7 +429,7 @@ impl RenderBundleEncoder {
                 }
                 RenderCommand::MultiDrawIndirect {
                     buffer_id,
-                    offset: _,
+                    offset,
                     count: None,
                     indexed: true,
                 } => {
@@ -440,12 +447,18 @@ impl RenderBundleEncoder {
                     check_buffer_usage(buffer.usage, wgt::BufferUsage::INDIRECT)
                         .map_pass_err(scope)?;
 
-                    let stride = 4 * 4; // 4 integers, vertexCount, instanceCount, firstVertex, firstInstance
-                    buffer_memory_init_actions.push(MemoryInitTrackerAction {
-                        id: buffer_id,
-                        range: 0..stride,
-                        kind: MemoryInitKind::NeedsInitializedMemory,
-                    });
+                    buffer_memory_init_actions.extend(
+                        buffer
+                            .initialization_status
+                            .check(
+                                offset..(offset + mem::size_of::<wgt::DrawIndirectArgs>() as u64),
+                            )
+                            .map(|range| MemoryInitTrackerAction {
+                                id: buffer_id,
+                                range,
+                                kind: MemoryInitKind::NeedsInitializedMemory,
+                            }),
+                    );
 
                     commands.extend(state.index.flush());
                     commands.extend(state.flush_vertices());
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -49,6 +49,7 @@ use crate::{
     },
     hub::{GfxBackend, GlobalIdentityHandlerFactory, Hub, Resource, Storage, Token},
     id,
+    memory_init_tracker::{MemoryInitTrackerAction, ResourceMemoryInitTrackerAction},
     resource::BufferUse,
     span,
     track::{TrackerSet, UsageConflict},
@@ -159,6 +160,7 @@ impl RenderBundleEncoder {
         let mut commands = Vec::new();
         let mut base = self.base.as_ref();
         let mut pipeline_layout_id = None::<id::Valid<id::PipelineLayoutId>>;
+        let mut used_buffer_ranges = Vec::new();
 
         for &command in base.commands {
             match command {
@@ -203,6 +205,9 @@ impl RenderBundleEncoder {
                         })
                         .map_pass_err(scope);
                     }
+
+                    used_buffer_ranges
+                        .extend(bind_group.used_buffer_ranges.iter().map(|x| x.clone()));
 
                     state.set_bind_group(index, bind_group_id, bind_group.layout_id, offsets);
                     state
@@ -262,6 +267,10 @@ impl RenderBundleEncoder {
                         Some(s) => offset + s.get(),
                         None => buffer.size,
                     };
+                    used_buffer_ranges.push(ResourceMemoryInitTrackerAction {
+                        id: buffer_id,
+                        action: MemoryInitTrackerAction::NeedsInitializedMemory(offset..end),
+                    });
                     state.index.set_format(index_format);
                     state.index.set_buffer(buffer_id, offset..end);
                 }
@@ -284,6 +293,10 @@ impl RenderBundleEncoder {
                         Some(s) => offset + s.get(),
                         None => buffer.size,
                     };
+                    used_buffer_ranges.push(ResourceMemoryInitTrackerAction {
+                        id: buffer_id,
+                        action: MemoryInitTrackerAction::NeedsInitializedMemory(offset..end),
+                    });
                     state.vertex[slot as usize].set_buffer(buffer_id, offset..end);
                 }
                 RenderCommand::SetPushConstant {
@@ -396,6 +409,11 @@ impl RenderBundleEncoder {
                     check_buffer_usage(buffer.usage, wgt::BufferUsage::INDIRECT)
                         .map_pass_err(scope)?;
 
+                    used_buffer_ranges.push(ResourceMemoryInitTrackerAction {
+                        id: buffer_id,
+                        action: MemoryInitTrackerAction::NeedsInitializedMemory(0..buffer.size),
+                    });
+
                     commands.extend(state.flush_vertices());
                     commands.extend(state.flush_binds());
                     commands.push(command);
@@ -419,6 +437,12 @@ impl RenderBundleEncoder {
                         .map_pass_err(scope)?;
                     check_buffer_usage(buffer.usage, wgt::BufferUsage::INDIRECT)
                         .map_pass_err(scope)?;
+
+                    let stride = 4 * 4; // 4 integers, vertexCount, instanceCount, firstVertex, firstInstance
+                    used_buffer_ranges.push(ResourceMemoryInitTrackerAction {
+                        id: buffer_id,
+                        action: MemoryInitTrackerAction::NeedsInitializedMemory(0..stride),
+                    });
 
                     commands.extend(state.index.flush());
                     commands.extend(state.flush_vertices());
@@ -454,6 +478,7 @@ impl RenderBundleEncoder {
                 ref_count: device.life_guard.add_ref(),
             },
             used: state.trackers,
+            used_buffer_ranges,
             context: self.context,
             life_guard: LifeGuard::new(desc.label.borrow_or_default()),
         })
@@ -502,6 +527,7 @@ pub struct RenderBundle {
     base: BasePass<RenderCommand>,
     pub(crate) device_id: Stored<id::DeviceId>,
     pub(crate) used: TrackerSet,
+    pub(crate) used_buffer_ranges: Vec<ResourceMemoryInitTrackerAction<id::BufferId>>,
     pub(crate) context: RenderPassContext,
     pub(crate) life_guard: LifeGuard,
 }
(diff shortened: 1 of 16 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -196,6 +196,30 @@ fn map_buffer<B: hal::Backend>(
...
+    for range in uninitialized_ranges {
+        unsafe {
+            ptr::write_bytes(
+                ptr.as_ptr().offset(range.start as isize),
+                0,
+                (range.end - range.start) as usize,
+            )
+        };
+        // (technically the buffer is only initialized when we're done unmapping but we know it can't be used otherwise meanwhile)
+        if writes_are_synced {
```
kvark, line 218: why are we checking for `writes_are_synced`? what does this variable mean, exactly?

wgpu-core/src/device/mod.rs
```diff
@@ -534,6 +558,13 @@ impl<B: GfxBackend> Device<B> {
             .allocate(&self.raw, requirements, mem_usage)?;
         block.bind_buffer(&self.raw, &mut buffer)?;
 
+        let mut uninitialized_ranges = range_alloc::RangeAllocator::new(Range {
+            start: 0,
+            end: desc.size,
+        });
+        // Mark buffer as uninitialized
+        let _ = uninitialized_ranges.allocate_range(desc.size);
```
kvark, line 566: would reversing this works? i.e. storing `initialized_ranges` instead of uninitialized

wgpu-core/src/device/queue.rs
```diff
@@ -271,6 +271,21 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         device.pending_writes.consume(stage);
         device.pending_writes.dst_buffers.insert(buffer_id);
 
+        // Ensure the overwritten bytes are marked as initialized so they don't need to be nulled prior to mapping or binding.
+        {
+            let dst = buffer_guard.get_mut(buffer_id).unwrap();
+            for range in dst
+                .uninitialized_ranges_in_range::<Vec<std::ops::Range<wgt::BufferAddress>>>(
```
kvark, line 278: in all the places where `uninitialized_ranges_in_range` is called, it would be good to fix two things:
  1. it should be an iterator, not a vector, to avoid the heap
  2. whatever it iterates on would just do `mark_initialized` internally

I.e. we could build an iterator that returns elements as it marks them initialized. Somewhat similar to `drain(..)` that returns elements and removes them.

wgpu-core/src/resource.rs
```diff
@@ -160,12 +160,47 @@ pub struct Buffer<B: hal::Backend> {
...
+    // An allocated range in this allocator means that the range in question is NOT yet initialized.
+    pub(crate) uninitialized_ranges: range_alloc::RangeAllocator<wgt::BufferAddress>,
     pub(crate) sync_mapped_writes: Option<hal::memory::Segment>,
     pub(crate) life_guard: LifeGuard,
     pub(crate) map_state: BufferMapState<B>,
 }
 
+impl<B: hal::Backend> Buffer<B> {
+    pub(crate) fn uninitialized_ranges_in_range<
+        R: std::iter::FromIterator<Range<wgt::BufferAddress>>,
```
kvark, line 172: could return `impl Iterator<...>` instead of this generic magic (minus notes that I wrote in an earlier comment)

wgpu-core/src/resource.rs
```diff
@@ -160,12 +160,47 @@ pub struct Buffer<B: hal::Backend> {
     pub(crate) device_id: Stored<DeviceId>,
     pub(crate) usage: wgt::BufferUsage,
     pub(crate) size: wgt::BufferAddress,
-    pub(crate) full_range: (),
+    // An allocated range in this allocator means that the range in question is NOT yet initialized.
+    pub(crate) uninitialized_ranges: range_alloc::RangeAllocator<wgt::BufferAddress>,
```
kvark, line 164: I feel that `RangeAllocator` is a wrong choice here. The goal for the allocator is to automatically allocate and deallocate ranges (e.g. sub-allocation from a pool). But here instead we need quick intersection search, and we never use automatic allocation, since we know all the ranges...

One possible replacement would be `BTreeMap`, where the key is the start of a range. Since it's ordered by keys, it should be able to quickly find intersections. I'm just a little concerned about how we are going to be updating the range intersections there, merging, etc... 
Another possibility is to look for something like an "interval map" on crates.

kvark (commented):
Thank you for this amazing work!
> Done brute force during queue submit, any uninitialized regions are zeroed out. 

I think that's the only way to go, so no worries about this being too brutal :)

> Does not take into account that only parts of a buffer may be in use. Easiest way to fix this more this is
probably just a list of bufferid + max-read-range in every command buffer. (agree / better idea?)
> Does not take into account if buffer is a copy target earlier on and needs no init,

Here is how I imagine this could work. A command buffer would store a list of actions for each buffer used in it, like: read 0..2, write 0..5, write 5..6, etc. We are probably going to need to mask these upon the uninitialized sections of the buffer *at the moment of recording* (which are easy to get, since we access the buffers immutably during encoding). Upon submission, the queue goes through this list and update the buffer's map of what is initialized and what not, inserting appropriate `fill_buffer` calls if necessary.

As for the "only parts of a buffer", all of the things you can do with a buffer are ranged: creating a bind group gets a range, binding vertex buffers gets a range, etc. So it sounds like we are good here?

Wumpf:
> Thank you for this amazing work!

Thanks for the kind words. Really, I think the way you treat any incoming contributions on gfx-rs sets an example to live by for any (open source) project.

Guess my description of what the init tracking on command buffer doesn't/should do wasn't phrased very clearly, but from what you write I can tell we mean the same thing now and you clarified a few things for me :). Was reluctant to do full "independent" tracking per command buffer which works independently of the existing tracker but I suppose that's the only thing that makes sense!

wgpu-core/src/device/mod.rs
Wumpf: When we zero out the user facing memory we naturally can mark the buffer as initialized if it is a write mapping. If it is a read mapping however, this only works if the memory block has the coherent flag. Otherwise, the writes done here are not flushed on unmap and thus may get lost / not be visible to the GPU.
That's what this this boolean captures (I'd check on the fly but borrows don't allow for that).

.. so that's at least how I understood the (un)mapping code around here. Is that accurate?

wgpu-core/src/device/mod.rs
Wumpf: see "RangeAllocator not a great fit" comment further down

wgpu-core/src/resource.rs
Wumpf: Agreed! I went with it because it was an idea that came up on the Matrix chat and I wasn't too keen on rewriting the merge logic. With RangeAllocator we also necessarily have this weird list of uninitialized ranges instead of the other way round because only free takes ranges on free.

Since you, as implied, don't mind adding another crate for this (if it doesn't have a log of dependencies I reckon :)) I'll have a look around for that. Otherwise, I may bite the bullet after all and write a dedicated thing either using BTreeMap/similar or from scratch which then can be used both here and on the command buffers.

kvark: Sorry about misleading you into `RangeAllocator`! This isn't one of the tasks that's obvious, it requires a bit of R&D, and I believe you are fully capable of doing it (assuming you have the time/fuel to invest).

kvark (commented):
If you want to make it simpler, and just consider a single range per buffer per command buffer, that's also fine by me! I don't want to blow this task out of proportion comparing to what you signed up for :)

wgpu-core/src/device/mod.rs
Wumpf: Yes, for the write mapping they should definitely be considered initialized. But *additionally* also for all read mappings on memory that has `HOST_COHERENT`-flag set, because then any changes will be visible to the GPU by definition.
The fact that nothing can use the buffer during the mapping is already used here by doing the mark-as-initialized right away and not in unmap (mostly for convenience).

For reference, `HOST_COHERENT` (== `block.is_coherent()`) in gpu-alloc is speced out as:
```
        /// This flag is set for host-coherent memory types.
        /// Host-coherent memory does not requires manual invalidation for
        /// modifications on GPU to become visible on host;
        /// nor flush for modification on host to become visible on GPU.
        /// Access synchronization is still required.
```

So that's why I put it down as if "either the memory is flushed on unmap (`buffer.sync_mapped_writes.is_some()`) or it is coherent (`|| block.is_coherent()`)", then we can assume that zeroing out the memory is permanent. Otherwise this host sided "change" may not be visible to the GPU so we're not allowed to mark it as initialized.

Does that make sense now or does my argumentation have holes?

kvark: For all the read mapping, it only comes back to us with a pointer *after* we invalidated the ranges. So regardless of coherent/non-coherent, the data in the buffer that comes in as "mapped for reading" is initialized by that point.

kvark: > Otherwise this host sided "change" may not be visible to the GPU so we're not allowed to mark it as initialized.

To clarify, we don't need to zero out memory that is read from GPU. The logic you are writing ensures that the source data on GPU is initialized properly, so whatever is being read from GPU to CPU must already contain valid to use data.

Wumpf: oh thanks didn't fully digest that mappings are permanent (mentally still stuck a bit in d3d9 style mapping it seems).

> so whatever is being read from GPU to CPU must already contain valid to use data

Hm last mental puzzle piece I'm missing for that is if `block.invalidate_range` won't really be able to drag garbage data from gpu. If I create an uninitialized buffer, map it, get hand-zeroed memory, unmap it, map it again, won't the call to `block.invalidate_range` mean that I could get bad GPU originated memory? After all, the zeros from the first mapping never made it to GPU and now I invalidated my host memory..?

Wumpf: can continue this on Matrix if more convenient

Wumpf:
small status update:
I think I figured out a workable scheme for fine grained buffer init requirement tracking by now.
Biggest pieces missing / still WIP:

* distill and apply needed buffer inits on queue submit
* get rid of range-alloc - looked a bit at what's out there in regards to segment trees or similar. Quite sure I'll roll my own by now since what we need here is a very limited usecase

wgpu-core/src/device/queue.rs
```diff
@@ -466,9 +478,96 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+                        .as_ref()
+                        .ok_or(QueueSubmitError::DestroyedBuffer(buffer_id))?
+                        .0;
+
+                    // TODO: Currently, we implement zero init only via fill_buffer. Should we implement this also via map & shader/storage?
+                    // Note that this requires TRANSFER usage on the buffer - see also device.create_buffer
+
+                    // TODO: Collapse ranges
+                    for range in ranges {
+                        // TODO SHIPSTOPPER: Insert necessary barriers!!
```
Wumpf, line 554: @kvark Not ready for full review again yet, but here I'd need some pointing into the right direction:

So what's happening here is that we just figured out which ranges need initialization via `fill_buffer`. `device.pending_writes` seems to be the right place to put it. But ofc I also need to create memory barriers and inform the tracker about this.. and this is where I'm struggling a bit. In other places this seems to be done without the tracker? Ideally I'd like to treat this just like another command in a wgpu command-buffer and don't operate on the lowlevel...

(note that this is the _only_ way initialization that I allow right now .. which has the awful implication that the TRANSFER bit needs to be always available :(( see also https://github.com/gfx-rs/wgpu/pull/1159/files#diff-90e54105789cf7268e0acbfd914035fccb5967895c81d21ac60f94af6b439edaR505)

Wumpf:
(Keeping main description up to date now with progress)

wgpu-core/src/device/queue.rs
kvark: > In other places this seems to be done without the tracker? 

Check out the `queue_write_buffer`. It needs exactly the same barriers (which are produced from the tracker):
```rust
let (dst, transition) = trackers
            .buffers
            .use_replace(&*buffer_guard, buffer_id, (), BufferUse::COPY_DST)
            .map_err(TransferError::InvalidBuffer)?;
        unsafe {
            stage.cmdbuf.pipeline_barrier(
                super::all_buffer_stages()..hal::pso::PipelineStage::TRANSFER,
                hal::memory::Dependencies::empty(),
                iter::once(hal::memory::Barrier::Buffer {
                    states: hal::buffer::Access::HOST_WRITE..hal::buffer::Access::TRANSFER_READ,
                    target: &stage.buffer,
                    range: hal::buffer::SubRange::WHOLE,
                    families: None,
                })
                .chain(transition.map(|pending| pending.into_hal(dst))),
            );
            stage
                .cmdbuf
                .copy_buffer(&stage.buffer, dst_raw, iter::once(region));
        }
```

> which has the awful implication that the TRANSFER bit needs to be always available

This is totally fine. For buffers, the set of allowed usages doesn't really make a difference for driver/HW. There is no downside of just internally adding `hal::buffer::Usage::TRANSFER_DST`.

Wumpf: perfect that's exactly the hint I needed

> For buffers, the set of allowed usages doesn't really make a difference for driver/HW
would have guess so that usages are only _really_ an issue for textures (because it may change what layout/compression it can do), but I'm very relieved you confirm this :). Then I can remove the TODOs around that aspect

Wumpf:
Ready for proper review now! Updated description accordingly.
Let me know if you want me to clean up the history, I tend to be on the verbose side with git commits ;)

wgpu-core/src/device/mod.rs
```diff
@@ -2559,6 +2600,19 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                         break e.into();
                     }
                 };
+
+                // Zero initialize memory and then mark both staging and buffer as initialized
+                // (it's guaranteed that this is the case by the time the buffer is usable)
+                unsafe { ptr::write_bytes(ptr.as_ptr(), 0, buffer.size as usize) };
+                buffer
+                    .initialization_status
```
kvark, line 2608: should we have a helper `clear()` method on this?

wgpu-core/src/device/queue.rs
```diff
@@ -191,7 +192,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         let device = device_guard
             .get_mut(queue_id)
             .map_err(|_| DeviceError::Invalid)?;
-        let (buffer_guard, _) = hub.buffers.read(&mut token);
+        let (mut buffer_guard, _) = hub.buffers.write(&mut token);
```
kvark, line 195: It's really unfortunate to lock it here for writing. Our locking is imperfect, and we should try to steer away from making it kill our perf.
Here, how do you feel about the following: we do everything as usual read-only, but in the end, if we see that some ranges need to be filled, we `drop(buffer_guard)` and then obtain a new writing guard, to deal with the init ranges specifically?

wgpu-core/src/device/queue.rs
```diff
@@ -458,17 +467,133 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+        if command_buffer_ids.is_empty() {
+            return Ok(());
+        }
+
+        let hub = B::hub(self);
+        let mut token = Token::root();
+
+        let mut required_buffer_inits = {
+            let (command_buffer_guard, mut token) = hub.command_buffers.read(&mut token);
+            let (mut buffer_guard, _) = hub.buffers.write(&mut token);
```
kvark, line 486: Could we perhaps move this check down into the inner loop, and only after checking that `cmdbuf.buffer_memory_init_actions` is non-empty?

wgpu-core/src/device/queue.rs
```diff
@@ -458,17 +467,133 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+        let mut trackers = device.trackers.lock();
+
+        for (buffer_id, mut ranges) in required_buffer_inits.drain() {
+            // Collapse touching ranges. We can't do this any earlier since we only now gathered ranges from several different command buffers!
+            ranges.sort_by(|a, b| a.start.cmp(&b.start));
+            for i in (1..ranges.len()).rev() {
+                assert!(ranges[i - 1].end <= ranges[i].start); // The memory init tracker made sure of this!
+                if ranges[i].start == ranges[i - 1].end {
+                    ranges[i - 1].end = ranges[i].end;
+                    ranges.remove(i);
```
kvark, line 548: uuh, that's going to make it quadratic complexity O(N^2)
Let's find a better way to collapse them?

wgpu-core/src/device/queue.rs
```diff
@@ -458,17 +467,133 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+                    transition.map(|pending| pending.into_hal(buffer)),
+                );
+            }
+            for range in ranges {
+                unsafe {
+                    pending_writes_cmd_buf.fill_buffer(
+                        buffer_raw,
+                        hal::buffer::SubRange {
+                            offset: range.start,
+                            size: Some(range.end - range.start),
```
kvark, line 573: I think `fill_buffer` requires that the offset and size are both aligned to 4. Are we taking care of this anywhere? Can we assert on that here, if we do?

wgpu-core/src/memory_init_tracker.rs
```diff
@@ -0,0 +1,196 @@
...
+            if uninitialized_range.start < self.drain_range.start {
+                let result = self.drain_range.start..uninitialized_range.end;
+                uninitialized_range.end = self.drain_range.start;
+                self.next_index = self.next_index + 1;
+                Some(result)
+            }
+            // fully contained.
+            else {
+                let result = uninitialized_range.clone();
+                self.uninitialized_ranges.remove(self.next_index);
```
kvark, line 69: O(N^2) again here. We need to think a bit more about the data structures, although we can land this variant for a baseline to improve upon. It's also unfortunate to hit this when all the user is doing being `.for_each(drop)`.

Strictly speaking, I think the whole drain should result in only 1 copy of the data. Instead of moving things one by one, we can do the following:
  - once given a range in `drain_uninitialized_ranges`, we need to find the range of indices in the `uninitialized_ranges` array that are affected.
  - we go through the indices and simply clamp each range to the drain range as we iterate, immutably
  - when the iteration is done, we can then do a single mutation of the `uninitialized_ranges`, where we remove everything from the middle and adjust the border ranges. This is where we may need to insert or remove at most 1 thing.

wgpu-core/src/memory_init_tracker.rs
```diff
@@ -0,0 +1,196 @@
...
+        assert!(tracker.is_initialized(&(0..3)));
+        assert!(tracker.is_initialized(&(3..4)));
+        assert!(tracker.is_initialized(&(4..10)));
+    }
+
+    #[test]
+    fn is_initialized_for_partially_filled_tracker() {
+        let mut tracker = MemoryInitTracker::new(10);
+        tracker.drain_uninitialized_ranges(4..6).for_each(drop);
+        assert!(!tracker.is_initialized(&(0..10))); // entire range
```
kvark, line 141: great to have these!

kvark (commented):
This looks amazing! Very well done, I'm now confident that you got the logic pieces right.
The only things left are details: minimizing the extra overhead with better algorithms, data structures, and less locking. And even this isn't mandatory - we can land it in a state close to what it is now and improve in follow-ups.

Please check out the comments I made, see which ones you feel motivated to address right away, and which ones can be postponed.

Another note: since this is a major change in logic, I'd fee a bit more secure if we roll it out in `master` but not get it into 0.7 release, which we are preparing now. So I'll probably branch out the release before merging.

Wumpf:
@kvark thank you for the review. I think I'll actually try to address all the comments over the weekend :). The O(N^2) totally slipped past by me, incorrectly thought it was still O(N).

> I'd fee a bit more secure if we roll it out in master but not get it into 0.7 release, 

ah yes, actually wanted to suggest that!

wgpu-core/src/command/bundle.rs
```diff
@@ -420,6 +441,13 @@ impl RenderBundleEncoder {
                     check_buffer_usage(buffer.usage, wgt::BufferUsage::INDIRECT)
                         .map_pass_err(scope)?;
 
+                    let stride = 4 * 4; // 4 integers, vertexCount, instanceCount, firstVertex, firstInstance
```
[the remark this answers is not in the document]
Wumpf: oh yes I'd like that a lot. I think I'll put that particular thing into a separate PR though

wgpu-core/src/device/queue.rs
Wumpf: offset and size being aligned to 4 should be the case for any operation, right? So I think we can just assume this is true here and assert for it (yes that is definitely in order)

Wumpf:
All done (again 😄)!
Only thing not address should be now the `struct IndirectDraw` for which I opened a separate PR (so once that lands this needs to be "integrated")

As a bonus I made the MemoryInitTracker checks O(log), but now I won't be able to see Ranges for a while 😄
MemoryInitTracker overall a bit more elegant now but also a bit more complex imho ⚖️

Wumpf:
Rebased and integrated the `IndirectDrawArgs`

kvark (approved):
Amazing!
bors r+

Issue #1158, opened 2021-01-19
Bad performance and memory usage on DX12


DasEtwas:
**Description**
The traced application runs poorly on a GTX 970 but it performs fine using the vulkan abckend.

**Repro steps**
Play the trace.

**Expected vs observed behavior**
Attached is a trace of a wgpu application running via DirectX 12 on a GTX 970.
Frame times should be lower than 0.02s. Application memory usage should be nearly static, levelling off at ~300-400MB, with only new planet chunk textures being allocated causing a deviation of at most 50 MB in CPU and GPU memory.
When not tracing, the application actually hangs after rendering two to three frames.

Memory usage is very high, with a constant rise of about 100MB per second or more in CPU memory.
Even when not tracing wgpu, frame times are about ~0.3s compared to ~0.008s via Vulkan.

**Platform**
`wgpu = { git = "https://github.com/gfx-rs/wgpu-rs", rev = "baf6c433bb8daa230c350c777e5f26fc95382495", features = ["trace"] }`

profile:
```
opt-level = 2
debug = true
debug-assertions = false
overflow-checks = true
lto = false
incremental = true
codegen-units = 64
````

trace: https://www.dropbox.com/s/bp1e9btyf9beeaw/wgpu_trace.zip

**Comment by baptistemanson** [comment]
Hi @DasEtwas , 
I ran your trace with wgpu c788 (he one used by wgpu-rs baf6c), and it crashes after opening a window and painting a black rectangle on the right hand side of the window:
```
thread 'main' panicked at 'Index 19 is already occupied', C:\...\wgpu-core\src\hub.rs:168:18
```
Maybe related to your previous finding https://github.com/gfx-rs/wgpu/issues/1146 ?

Hope that helps anyone in diagnosing something.

DasEtwas:
> Hi @DasEtwas , 
> I ran your trace with wgpu c788 (he one used by wgpu-rs baf6c), and it crashes after opening a window and painting a black rectangle on the right hand side of the window:
> ```
> thread 'main' panicked at 'Index 19 is already occupied', C:\...\wgpu-core\src\hub.rs:168:18
> ```
> Maybe related to your previous finding https://github.com/gfx-rs/wgpu/issues/1146 ?
> 
> Hope that helps anyone in diagnosing something.

[...]

PR #1161, opened 2021-01-20
Fix replaying issues with dropped texture views

kvark:
**Connections**
Fixes replaying of https://github.com/gfx-rs/wgpu/issues/1158#issuecomment-763272796

**Description**
The main fix is calling `maintain` in the player before getting the new view from the swapchain. That allows the device to properly remove the ID that is just about to be used.

**Testing**
Tested on that API trace.


kvark:
bors r=urgent

PR #1180, opened 2021-01-30
Add dummy structs for indirect command buffers

Wumpf:
**Connections**
Review of #1159 which could make more use of this expressiveness patch of having the buffers spelled out.

**Description**
Add dummy structs for indirect command buffers and (bonus!) fix missing IndirectBufferOverrun error for IndirectDispatch.

**Testing**
squinting at it only 👁️👁️

### DIFF
--- a/wgpu-core/src/command/compute.rs
+++ b/wgpu-core/src/command/compute.rs
@@ -23,7 +23,7 @@ use thiserror::Error;
 use wgt::{BufferAddress, BufferUsage, ShaderStage};
 
 use crate::track::UseExtendError;
-use std::{fmt, iter, str};
+use std::{fmt, iter, mem, str};
 
 #[doc(hidden)]
 #[derive(Clone, Copy, Debug)]
@@ -140,6 +140,12 @@ pub enum ComputePassErrorInner {
     InvalidQuerySet(id::QuerySetId),
     #[error("indirect buffer {0:?} is invalid or destroyed")]
     InvalidIndirectBuffer(id::BufferId),
+    #[error("indirect buffer uses bytes {offset}..{end_offset} which overruns indirect buffer of size {buffer_size}")]
+    IndirectBufferOverrun {
+        offset: u64,
+        end_offset: u64,
+        buffer_size: u64,
+    },
     #[error(transparent)]
     ResourceUsageConflict(#[from] UsageConflict),
     #[error(transparent)]
@@ -490,6 +496,17 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                         .map_pass_err(scope)?;
                     check_buffer_usage(indirect_buffer.usage, BufferUsage::INDIRECT)
                         .map_pass_err(scope)?;
+
+                    let end_offset = offset + mem::size_of::<wgt::DispatchIndirectArgs>() as u64;
+                    if end_offset > indirect_buffer.size {
+                        return Err(ComputePassErrorInner::IndirectBufferOverrun {
+                            offset,
+                            end_offset,
+                            buffer_size: indirect_buffer.size,
+                        })
+                        .map_pass_err(scope);
+                    }
+
                     let &(ref buf_raw, _) = indirect_buffer
                         .raw
                         .as_ref()
--- a/wgpu-core/src/command/render.rs
+++ b/wgpu-core/src/command/render.rs
@@ -45,6 +45,7 @@ use std::{
     collections::hash_map::Entry,
     fmt, iter,
     marker::PhantomData,
+    mem,
     num::NonZeroU32,
     ops::Range,
     str,
@@ -417,11 +418,10 @@ pub enum RenderPassErrorInner {
     InvalidValuesOffset,
     #[error("required device features not enabled: {0:?}")]
     MissingDeviceFeatures(wgt::Features),
-    #[error("indirect draw with offset {offset}{} uses bytes {begin_offset}..{end_offset} which overruns indirect buffer of size {buffer_size}", count.map_or_else(String::new, |v| format!(" and count {}", v)))]
+    #[error("indirect draw uses bytes {offset}..{end_offset} {} which overruns indirect buffer of size {buffer_size}", count.map_or_else(String::new, |v| format!("(using count {})", v)))]
     IndirectBufferOverrun {
-        offset: u64,
         count: Option<NonZeroU32>,
-        begin_offset: u64,
+        offset: u64,
         end_offset: u64,
         buffer_size: u64,
     },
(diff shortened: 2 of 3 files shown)


wgpu-core/src/command/compute.rs
```diff
@@ -140,6 +140,13 @@ pub enum ComputePassErrorInner {
     InvalidQuerySet(id::QuerySetId),
     #[error("indirect buffer {0:?} is invalid or destroyed")]
     InvalidIndirectBuffer(id::BufferId),
+    #[error("indirect buffer with offset {offset}{} uses bytes {begin_offset}..{end_offset} which overruns indirect buffer of size {buffer_size}")]
```
cwfitzgerald, line 143: `{offset}{}` what happened here?

wgpu-core/src/command/compute.rs
```diff
@@ -228,6 +235,14 @@ impl State {
     }
 }
 
+#[repr(packed)]
+#[allow(dead_code)]
+struct DispatchIndirectBuffer {
```
cwfitzgerald, line 240: These should be in wgpu-types so that wgpu users can use them too if they want to fill an indirect buffer on the cpu (useful for testing)

wgpu-core/src/command/render.rs
```diff
@@ -508,6 +508,25 @@ struct RenderPassInfo<'a, B: hal::Backend> {
...
+    first_instance: u32,
+}
+
+#[repr(packed)]
+#[allow(dead_code)]
+struct DrawIndexedIndirectBuffer {
+    index_count: u32,
+    instance_count: u32,
+    first_index: u32,
+    base_vertex: u32,
```
cwfitzgerald, line 526: One of these members (I think it is this one) is a i32 not a u32. Check some spec (likely DX12 docs)

wgpu-core/src/command/render.rs
```diff
@@ -1549,9 +1568,9 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                         state.is_ready().map_pass_err(scope)?;
 
                         let stride = match indexed {
-                            false => 16,
-                            true => 20,
-                        };
+                            false => std::mem::size_of::<DrawIndirectBuffer>(),
```
cwfitzgerald, line 1571: nit: use `mem::size_of` if possible.

cwfitzgerald (changes requested):
Some notes, curious to see what @kvark thinks.

wgpu-core/src/command/render.rs
Wumpf: Well shouldn't matter what DX12 does but you're right, Webbgpu specs baseVertex as signed!

cwfitzgerald: Only mentioned DX as I knew they had good docs on it, all indirect args are the same cross api.

wgpu-core/src/command/compute.rs
```diff
@@ -228,6 +235,14 @@ impl State {
     }
 }
 
+#[repr(packed)]
```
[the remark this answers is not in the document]
Wumpf: I mean none of these make practically a difference here. But I think repr(c) would technically be incorrect. Let's say I have some esoteric architecture where all these 32 bit values have 64bit padding. Then this would violate the WebGPU spec which says the values are all tightly packed, right? Problem is that we're not really after having a usable struct but after a description of the spec

wgpu-core/src/command/render.rs
Wumpf: you mean as in importing `std::mem` so I can do `mem::size_of::<DrawIndirectBuffer>()` or is there some other Rust thing I don't know? :)

cwfitzgerald: Yeah that's it :)

wgpu-core/src/command/compute.rs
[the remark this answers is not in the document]
cwfitzgerald: I mean an arch where primatives had size != alignment would cause so many issues that it honestly isn't worth thinking about. Either way I think rust guarantees the alignment of primatives. The reason I don't like packed is that it's super easy to get off into _real fuck your shit up_ ub land if you end up with misaligned references and it's use should be discouraged unless there's no way around it.

Wumpf:
@cwfitzgerald thanks for quick replies. Addressed all comments, including the repr(c) although I still disagree there ;) 🤷

cwfitzgerald (approved):

wgpu-core/src/command/compute.rs
```diff
@@ -490,6 +497,19 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                         .map_pass_err(scope)?;
                     check_buffer_usage(indirect_buffer.usage, BufferUsage::INDIRECT)
                         .map_pass_err(scope)?;
+
+                    let begin_offset = offset;
+                    let end_offset = offset + mem::size_of::<wgt::DispatchIndirectArgs>() as u64;
+                    if end_offset > indirect_buffer.size {
+                        return Err(ComputePassErrorInner::IndirectBufferOverrun {
+                            offset,
+                            begin_offset,
```
kvark, line 506: why are we having separate `begin_offset` from `offset`? They appear to bbe the same thing

wgpu-types/src/lib.rs
```diff
@@ -2811,3 +2811,33 @@ bitflags::bitflags! {
         const COMPUTE_SHADER_INVOCATIONS = 0x10;
     }
 }
+
+/// Argument buffer layout for draw_indirect commands.
+#[repr(C)]
+#[derive(Clone, Debug)]
+pub struct DrawIndirectArgs {
+    vertex_count: u32,
```
kvark, line 2819: I wonder if these should be public? I.e. imagine someone using rust-gpu with wgpu-rs, they'd just `use` this structure from their shader modules and fill it up.

kvark (commented):
Great stuff! And thanks @cwfitzgerald for reviewing, I agree with all the notes discussed previously. Just adding a couple of minor ones before we land.

wgpu-core/src/command/compute.rs
Wumpf: that was caused by ~mindless copying~ style preserving from render.rs where this is done in two different places 😄
I'll clean all of them up then 🤷

wgpu-types/src/lib.rs
Wumpf: dunno, had them private initially and in render.rs/compute.rs respectively. @cwfitzgerald suggested to have them in types which I took as making them public. Actually, I think it can be useful for them to be public for e.g. prefilling on host side a `Buffer` that is then adjusted in a shader. Also - more trivially - the same usecase we have here, to have something to use `mem::size_of` on

kvark: Sorry, you misunderstood. I mean, the fields should be public as well, not just the struct types themselves.

Wumpf: ah ofc, my fault. Fixed and took over the docs for the fields from WebGPU spec.

Also briefly looked at the doc in wgpu-rs. Slight differences there in field naming and one mistake there:
https://github.com/gfx-rs/wgpu-rs/blob/master/src/util/encoder.rs#L77
`DrawIndexedIndirect` is described there to have `vertex_count` instead of an `index_count` (also quickly checked with DX12 and Vulkan docs; yes it's an **index** count)
We can probably update that doc to point to these arg types now anyways :)

Wumpf:
squashed the commits. That was too commit history per LOC for even for me 😅

kvark:
Thank you!
Bors r=cwfitzgerald,kvark

PR #1190, opened 2021-02-03
Fix crash on zero init of buffer with no more ref count

Wumpf:
**Description**
Previously, if a buffer would no longer have any reference, the zero init step crashed when trying to unwrap its ref_count.

**Testing**
Tested successfully on https://github.com/Wumpf/blub/commit/54a0f4ff13fcd969d87e4469067c6403999f79e2 where this problem would pop up

### DIFF
--- a/wgpu-core/src/device/queue.rs
+++ b/wgpu-core/src/device/queue.rs
@@ -559,9 +559,15 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 }
             }
 
-            let (buffer, transition) = trackers
-                .buffers
-                .use_replace(&*buffer_guard, buffer_id, (), BufferUse::COPY_DST)
+            // Don't do use_replace since the buffer may already no longer have a ref_count.
+            // However, we *know* that it is currently in use, so the tracker must already know about it.
+            let transition = trackers.buffers.change_replace_tracked(
+                id::Valid(buffer_id),
+                (),
+                BufferUse::COPY_DST,
+            );
+            let buffer = buffer_guard
+                .get(buffer_id)
                 .map_err(|_| QueueSubmitError::DestroyedBuffer(buffer_id))?;
             let &(ref buffer_raw, _) = buffer
                 .raw
--- a/wgpu-core/src/track/mod.rs
+++ b/wgpu-core/src/track/mod.rs
@@ -315,6 +315,18 @@ impl<S: ResourceState> ResourceTracker<S> {
         }
     }
 
+    fn get<'a>(
+        self_backend: wgt::Backend,
+        map: &'a mut FastHashMap<Index, Resource<S>>,
+        id: Valid<S::Id>,
+    ) -> &'a mut Resource<S> {
+        let (index, epoch, backend) = id.0.unzip();
+        debug_assert_eq!(self_backend, backend);
+        let e = map.get_mut(&index).unwrap();
+        assert_eq!(e.epoch, epoch);
+        e
+    }
+
     /// Extend the usage of a specified resource.
     ///
     /// Returns conflicting transition as an error.
@@ -342,6 +354,21 @@ impl<S: ResourceState> ResourceTracker<S> {
         res.state
             .change(id, selector, usage, Some(&mut self.temp))
             .ok(); //TODO: unwrap?
+        self.temp.drain(..)
+    }
+
+    /// Replace the usage of a specified already tracked resource.
+    /// (panics if the resource is not yet tracked)
+    pub(crate) fn change_replace_tracked(
+        &mut self,
+        id: Valid<S::Id>,
+        selector: S::Selector,
+        usage: S::Usage,
+    ) -> Drain<PendingTransition<S>> {
+        let res = Self::get(self.backend, &mut self.map, id);
+        res.state
+            .change(id, selector, usage, Some(&mut self.temp))
+            .ok();
         self.temp.drain(..)
     }
 


wgpu-core/src/device/queue.rs
```diff
@@ -548,6 +553,10 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         let pending_writes_cmd_buf = device.pending_writes.command_buffer.as_mut().unwrap();
         let mut trackers = device.trackers.lock();
 
+        let tmp_ref_count = ptr::NonNull::new(Box::into_raw(Box::new(AtomicUsize::new(1))))
```
kvark, line 556: oh no, that's definitely not what we want here :)
Here it's too late to handle the deleted buffer. Instead, we should grab a refcount from it where it's used.
So `buffer_memory_init_actions` needs to carry that refcount to us.

kvark (changes requested):

Wumpf:
@kvark Let's pretend we never saw the first version of this PR :). Stuff actually makes sense now I think! But also it doesn't quite work, seems this keeps frames now alive for too long. Getting the following crash:
```
 0: std::panicking::begin_panic_handler
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4\/library\std\src\panicking.rs:483
   1: std::panicking::begin_panic_fmt
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4\/library\std\src\panicking.rs:437
   2: wgpu::backend::direct::Context::handle_error_fatal<wgpu_core::swap_chain::SwapChainError>
             at C:\Users\andre\.cargo\git\checkouts\wgpu-rs-40ea39809c03c5d8\f891e86\src\backend\direct.rs:129
   3: wgpu::backend::direct::{{impl}}::swap_chain_present
             at C:\Users\andre\.cargo\git\checkouts\wgpu-rs-40ea39809c03c5d8\f891e86\src\backend\direct.rs:1371
   4: core::ptr::drop_in_place<wgpu::SwapChainTexture>
             at C:\Users\andre\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\ptr\mod.rs:175
   5: core::mem::drop<wgpu::SwapChainTexture>
             at C:\Users\andre\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\mem\mod.rs:881
   6: blub::render_output::screen::Screen::end_frame
             at .\src\render_output\screen.rs:202
   7: blub::Application::draw
   ```
   
So what I naturally tried is to kill all the `buffer_memory_init_actions` from the command buffers by draining them in `initialize_used_uninitialized_memory`, but that didn't have the desired effect. Won't get to more digging today, but maybe you already know where to look? :)

**Comment by Wumpf** [comment]
also it's a bit unfortunate that every buffer use in any binding will now bump the ref_count. Well, but it's ofc not really wrong & we can live with that I believe

**Comment by kvark** [comment]
My apologies, I may have led you to a sub-optimal path. We don't want to do another  `add_ref()` there (although we can!) during recording because all the buffers are already referenced by the command buffer tracker! So we should add the following function to `impl<> ResourceTracker<>`:
```rust
    pub fn get_ref_count(&self, id: Valid<S::Id>) -> Option<&RefCount> {
        let (index, epoch, backend) = id.0.unzip();
        debug_assert_eq!(self.backend, backend);
        self.map.get(&index).map(|e| {
            assert_eq!(e.epoch, epoch);
            &e.ref_count
        })
    }
```
That would allows us to do `let ref_count = cmd_buf.tracker.get_ref_count(my_buffer_id).unwrap()` to then use for the state changes.

Wumpf:
oh. Yeah that's much better and easier. Thanks for writing it out as working code already. 
So when I tried to wire it up I ran into a relatively obvious borrow checker problem:
```
error[E0502]: cannot borrow `trackers` as mutable because it is also borrowed as immutable
   --> wgpu-core\src\device\queue.rs:568:30
    |
564 |             let buffer_ref_count = trackers
    |                                    -------- immutable borrow occurs here
...
568 |             let transition = trackers.buffers.change_replace(
    |                              ^^^^^^^^ mutable borrow occurs here
569 |                 id::Valid(buffer_id),
570 |                 buffer_ref_count,
    |                 ---------------- immutable borrow later used here
```
Pondering it a bit I decided to put it all into a `change_replace_tracked` method. The good thing about this is that it saves a hashmap lookup (as opposed to doing one lookup for `get_ref_count` and one lookup in `get_or_insert`), bad thing is that it adds an awkward "you know what you're doing?"-api to `ResourceTracker`. You be the judge :)
Tested it with Blub, works.


> already referenced by the command buffer tracker
I find it still unfortunate that the MemoryInitTracker and the the Tracker are separate. Feels like it should be possible to have a universal tracker of sorts.

wgpu-core/src/track/mod.rs
```diff
@@ -315,6 +315,18 @@ impl<S: ResourceState> ResourceTracker<S> {
         }
     }
 
+    fn get<'a>(
```
kvark, line 318: How do you feel about making `ref_count: Option<&RefCount>` argument in the existing `get_or_insert`, instead of adding a new method?

kvark (commented):
I see what you did here. This is a good way to approach this, better than what I suggested:)

kvark:
@Wumpf actually, hold on. I think I tried that before, and it becomes a little unfortunate that we wrap it in `Some` when we know the check is not needed. Let's just proceed.
bors r+

PR #1191, opened 2021-02-03
Buffer zero init test for binding (and use in a compute shader)

Wumpf:
**Connections**
Buffer zero init #1159 
Buffer zero crash on unused buffer fix #1190

**Description**
Tried to create a player ron file repro case for #1190 but didn't work out quite well (test keeps buffers alive even when I tried to go through an indirection via buffer copies). But with a bit of cleanup the test still felt very useful since it gives more coverage to buffer zero init.

Generally zero init is hard to test since the way to get the result is to map it which is one particular case of requiring zero init. But there's a few ways like in this test here where we look at what modifications a compute shader did to a buffer - since the buffer was not explicitly initialized, the outcome would be different if the compute shader would have seen a buffer without zero init.

Given how few tests there are right now this ofc implicitly tests a bunch of other things ;)


player/tests/data/buffer-zero-init-for-binding.ron
```diff
@@ -0,0 +1,99 @@
...
+            buffer: (index: 0, epoch: 1),
+            offset: 0,
+            data: Raw([0x00, 0x00, 0x00, 0x00,
+                       0x01, 0x00, 0x00, 0x00,
+                       0x02, 0x00, 0x00, 0x00,
+                       0x03, 0x00, 0x00, 0x00]),
+        )
+    ],
+    actions: [
+        CreateShaderModule(
```
kvark, line 15: If you are trying to ensure that bindings get zeroed out, you don't actually need shaders. Any bindings done inside a render pass are considered a part of the sync scope. So you can just start a dummy render pass, bind something, and end the pass, and you should be able to test the effect of these bindings (without any draws).

kvark:
Also filed https://github.com/gfx-rs/wgpu/issues/1195 to make our RON captures nicer.

player/tests/data/buffer-zero-init-for-binding.ron
Wumpf: Note quite true I believe. Let's say zero out on mapping works perfectly fine but our zero out on binding completely broke. Then just doing the binding will not zero the buffer out but later when we check the result we still get a zeroed out buffer. Thus only testing what the existing test already tested.

By having a shader that produces a different result depending on was in the buffer at the time of the shader invocation, we can reason about the zeroing that should have occurred because of the binding.

kvark: That's good thinking, you are right!

player/tests/data/all.ron
```diff
@@ -2,7 +2,8 @@
 	backends: (bits: 0xF),
 	tests: [
 		"buffer-copy.ron",
-		"buffer-zero-init.ron",
```
kvark, line 5: have you considered leaving this inside a single test? The test would have 2 expectations: one for the mapping, one of the binding. It can use different buffers for these.

Wumpf: sure, can do! Personally, I prefer more atomic unit tests but hard to get anything like that anyways. I'll bundle everying buffer-zero-init related then

kvark (approved):
Thank you!
bors r+

PR #1194, opened 2021-02-04
Isolate binding compatibility logic into a separate module

### LINKED ISSUE #1188: Binding misbehaves when switching between incompatible pipelines
**Description**
I recently changed the first binding group of one of our pipelines to have an incompatible layout with the others by adding a few entries to the bind group specific to that pipeline. Now when switching to that pipeline and binding the alternative bind group at `0` a vulkan validation error that set `#1` isn't bound occurs along with a segfault.
```
VALIDATION [UNASSIGNED-CoreValidation-DrawState-DescriptorSetNotBound (-840888189)] : Validation Error: [ UNASSIGNED-CoreValidation-DrawState-DescriptorSetNotBound ] Object 0: handle = 0x5639ca32d5f0, name = first pass, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xcde11083 | VkPipeline 0x390000000039[] uses set #1 but that set is not bound.
object info: (type: COMMAND_BUFFER, hndl: 94806205453808, name: first pass)
```
Interestingly, keeping the call to switch the pipeline but omitting `self.render_pass.set_bind_group(0, &globals.bind_group, &[]);` which switches to the bind group that's incompatible with the other pipelines makes the issue go away (the draw call for this pipeline also has to be omitted since the bind groups are incompatible without the switch but only omitting the draw call does not eliminate the issue). 

The bind group set at `1` is set at the beginning of the pass and isn't changed.

Inserting numerous calls to `set_bind_group` for the sets at `0` and `1` around all the pipeline switches does not resolve the issue.

**Repro steps**
This isn't a minimal example:

- clone https://gitlab.com/Imbris/veloren (git-lfs needed)
- checkout `wgpu-bind-error`
- for quickest startup, in a separate terminal or in the background, run a server with `cargo test-server --no-auth`
- start voxygen with `cargo r`
- type in random username, set the server to `localhost`, and click the multiplayer button
- create a character and enter the world
- observe bug

Expected: Bind groups remain bound after switching pipelines, rebinding with `set_bind_group` works, and a wgpu level error occurs when bind groups aren't configured properly.
Observed: Vulkan validation error and segfault

**Extra materials**

[wgpu-binding-error-trace.zip](https://github.com/gfx-rs/wgpu/files/5916033/wgpu-binding-error-trace.zip)

**Platform**
Currently on `ab8b0e3766558d541206da2790dfd63f15b13bc4` using the vulkan backend

kvark:
**Connections**
Fixes #1188 

**Description**
The actual problem was the following: when a pipeline changed, we used to go through the binding entries and re-bind only those that *became* compatible. However, we were missing ones that were already compatible, but were not activated because they were behind some previously incompatible ones.

So it's not a very complex fix. However, working with the binding code had me concerned that it wasn't isolated enough. It got quite a few issues discovered in it over the time, which only proven the point that it was way too fragile. So I took this opportunity to bring us closer to the world I want wgpu to be in: the world of semi-independent testable components.

This PR makes the binding compatibility checker one of these components. It's fully isolated, has unit tests, fairly straightforward to reason about, and integrates nicely with the render pass logic (without too many generics!). It's more profitable in the long run, I think, to build this kind of architecture versus integration-testing the thing (even at `wgpu` playtest level).

There is less magic overall now: no binding iterator, no fancy chaining, no complex matching. The logic is straightforward, hopefully. And the 40 LOC saved is a good indication of it.

**Testing**
Tested on wgpu-rs examples.

### DIFF
--- a/wgpu-core/src/command/bind.rs
+++ b/wgpu-core/src/command/bind.rs
@@ -11,236 +11,215 @@ use crate::{
 };
 
 use arrayvec::ArrayVec;
-use std::slice;
-use wgt::DynamicOffset;
 
 type BindGroupMask = u8;
 
-#[derive(Clone, Debug)]
-pub(super) struct BindGroupPair {
-    layout_id: Valid<BindGroupLayoutId>,
-    group_id: Stored<BindGroupId>,
-}
+mod compat {
+    use std::ops::Range;
 
-#[derive(Debug)]
-pub(super) enum LayoutChange<'a> {
-    Unchanged,
-    Match(Valid<BindGroupId>, &'a [DynamicOffset]),
-    Mismatch,
-}
-
-#[derive(Debug)]
-pub enum Provision {
-    Unchanged,
-    Changed { was_compatible: bool },
-}
-
-#[derive(Clone)]
-pub(super) struct FollowUpIter<'a> {
-    iter: slice::Iter<'a, BindGroupEntry>,
-}
-impl<'a> Iterator for FollowUpIter<'a> {
-    type Item = (Valid<BindGroupId>, &'a [DynamicOffset]);
-    fn next(&mut self) -> Option<Self::Item> {
-        self.iter
-            .next()
-            .and_then(|entry| Some((entry.actual_value()?, entry.dynamic_offsets.as_slice())))
+    #[derive(Debug)]
+    struct Entry<T> {
+        assigned: Option<T>,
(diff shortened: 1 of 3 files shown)


cwfitzgerald (approved):
Code looks so much cleaner now! LGTM!

kvark:
bors r=cwfitzgerald

PR #1201, opened 2021-02-06
Don't check the index format for non-indexed calls

### LINKED ISSUE #1200: "index buffer must be set", but I'm not using draw_indexed
I'm using a simple shader to render a rectangle onto the screen. The vertices are hardcoded into the vertex shader, and it uses the vertex index to find out what vertex to emit.

```glsl
#version 450

// triangle strip
const vec2 fb_vert[4] = vec2[4](
	vec2(1, 1),  // top right
	vec2(-1, 1), // top left
	vec2(1, -1), // bottom right
	vec2(-1, -1) // bottom left
);

void main() {
	gl_Position = vec4(fb_vert[gl_VertexIndex], 0, 1);
}
```

Well, on updating from wgpu 0.6.2 to 0.7.0, it started throwing "index buffer must be set" errors, even though it didn't before.

Is this a bug? I don't know what an index buffer is or why I would need one for this use case. I'm not even using `draw_indexed`:

```rust
let mut render_pass = encoder.begin_render_pass(&wgpu::RenderPassDescriptor {
	label: Some("<...> primary render pass descriptor"),
	color_attachments: &[
		wgpu::RenderPassColorAttachmentDescriptor {
			attachment: &frame.output.view,
			resolve_target: None,
			ops: wgpu::Operations {
				load: wgpu::LoadOp::Clear(self.background),
				store: true
			}
		}
	],
	depth_stencil_attachment: None
});

render_pass.set_pipeline(&self.pipeline);
render_pass.set_bind_group(0, &self.bind_group, &[]);
render_pass.draw(0..4, 0..1);
```

kvark:
**Connections**
Fixes #1200

**Description**
Avoid the check that shouldn't be done.

**Testing**
Not tested, but should work.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -867,7 +867,8 @@ impl Default for PolygonMode {
 pub struct PrimitiveState {
     /// The primitive topology used to interpret vertices.
     pub topology: PrimitiveTopology,
-    /// The format of index buffers for strip topologies. Should be left `None` for non-strip.
+    /// When drawing strip topologies with indices, this is the required format for the index buffer.
+    /// This has no effect on non-indexed or non-strip draws.
     #[cfg_attr(any(feature = "trace", feature = "replay"), serde(default))]
     pub strip_index_format: Option<IndexFormat>,
     /// The face to consider the front for the purpose of culling and stencil operations.
--- a/wgpu-core/src/command/render.rs
+++ b/wgpu-core/src/command/render.rs
@@ -336,7 +336,7 @@ struct State {
 }
 
 impl State {
-    fn is_ready(&self) -> Result<(), DrawError> {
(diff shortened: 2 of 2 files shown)


kvark:
improved the documentation a bit
bors r+

wgpu-types/src/lib.rs
```diff
@@ -867,7 +867,8 @@ impl Default for PolygonMode {
 pub struct PrimitiveState {
     /// The primitive topology used to interpret vertices.
     pub topology: PrimitiveTopology,
-    /// The format of index buffers for strip topologies. Should be left `None` for non-strip.
+    /// When drawing stip topologies with indices, this is the required format for the index buffers.
```
LoganDark, line 870: stip

kvark: really need to wake up...

PR #1210, opened 2021-02-11
Reset the bindings on the push constant change

### LINKED ISSUE #1207: Vulkan validation error around descriptor sets not being bound
**Description**
```
[2021-02-10T22:45:37Z ERROR gfx_backend_vulkan] 
    VALIDATION [VUID-vkCmdDraw-None-02697 (684884398)] : Validation Error: [ VUID-vkCmdDraw-None-02697 ] Object 0: handle = 0x7270fd0000000078, type = VK_OBJECT_TYPE_PIPELINE; Object 1: handle = 0x4bad500000000075, type = VK_OBJECT_TYPE_PIPELINE_LAYOUT; Object 2: handle = 0x51053f000000007a, type = VK_OBJECT_TYPE_PIPELINE_LAYOUT; | MessageID = 0x28d281ae | vkCmdDraw(): VkPipeline 0x7270fd0000000078[] defined with VkPipelineLayout 0x4bad500000000075[] is not compatible for maximum set statically used 0 with 
bound descriptor sets, last bound with VkPipelineLayout 0x51053f000000007a[] The Vulkan spec states: For each set n that is statically used by the VkPipeline bound to the pipeline bind point used by this command, a descriptor set must have been bound to n 
at the same pipeline bind point, with a VkPipelineLayout that is compatible for set n, with the VkPipelineLayout used to create the current VkPipeline, as described in Pipeline Layout Compatibility (https://vulkan.lunarg.com/doc/view/1.2.141.0/windows/1.2-extensions/vkspec.html#VUID-vkCmdDraw-None-02697)
    object info: (type: PIPELINE, hndl: 8246369094157205624), (type: PIPELINE_LAYOUT, hndl: 5453102684745760885), (type: PIPELINE_LAYOUT, hndl: 5838141761188266106)

[2021-02-10T22:45:37Z ERROR gfx_backend_vulkan] 
    VALIDATION [UNASSIGNED-CoreValidation-DrawState-DescriptorSetNotBound (-840888189)] : Validation Error: [ UNASSIGNED-CoreValidation-DrawState-DescriptorSetNotBound ] Object 0: handle = 0x27c3a5062d8, name = opaque, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xcde11083 | VkPipeline 0x2f99810000000044[] uses set #0 but that set is not bound.
    cmd buf info: CubemapRenderer.draw
    object info: (type: COMMAND_BUFFER, hndl: 2732577546968, name: opaque)
```

Tried before and after #1194, confirming that it introduced the issue.

The two validation errors seem to be related but distinct - depending on which pipelines I disable I can get either none of them or only the first. Second one causes an actual application crash.
Still investigating the details.

**Repro steps**
Unknown

No warning and crash

**Extra materials**

[trace.zip](https://github.com/gfx-rs/wgpu/files/5962135/trace.zip)

**Platform**
Win @ Vulkan (surprise)

kvark:
**Connections**
Fixes #1207
(unconfirmed!)

**Description**
I was able to replay the trace in #1207 without validation errors, so I looked deeper into what might have gone wrong.
The code isolated in #1194 looks great. However, the calling code's handling of push constants might have changed a bit.
And I see that the trace uses push constants, so here is what happened, I think:
  - layout A was set, expecting bind group layouts X and Z
  - bind groups X and Y where bound
  - layout B was set, expecting bind group layout X and Y. We see that Y group can now be bound, so we do this. The old logic in this case wasn't considering the push constants in any way, it would only consider them if bind group layouts didn't change, erroneously.
  - layout B has different push constant ranges...

So it's a one-line fix now, which I'm hoping is correct.

**Testing**
Not tested
@Wumpf would you be able to check this?

### DIFF
--- a/wgpu-core/src/command/bind.rs
+++ b/wgpu-core/src/command/bind.rs
@@ -179,9 +179,7 @@ impl Binder {
         if let Some(old_id) = old_id_opt {
             let old = &guard[old_id];
             // root constants are the base compatibility property
-            if old.push_constant_ranges != new.push_constant_ranges
-                && bind_range.end != bind_range.start
-            {
+            if old.push_constant_ranges != new.push_constant_ranges {
                 bind_range.start = 0;
             }
         }


Wumpf (approved):
Thank you so much! Confirmed working. Code change as such makes also sense
🚢

cwfitzgerald:
bors r=wumpf

PR #1219, opened 2021-02-16
Validate texture bindings

kvark:
**Connections**
It was mentioned/requested somewhere, but I can't find the place...

**Description**
Fairly straightforward - just check the view properties against the bind group layout.

**Testing**
Tested on wgpu-rs examples


cwfitzgerald (approved):
Whoop!

bors r+

PR #1220, opened 2021-02-16
Update naga to gfx-11, make spirv-cross optional

kvark:
**Connections**
https://github.com/gfx-rs/gfx/pull/3642
https://github.com/gfx-rs/gfx/pull/3641

**Description**
Gets us the latest Naga updates, plus the ability to make spirv-cross optional (on Linux and Windows only for now).

**Testing**
Will be tested on wgpu-rs


kvark:
Seems to work on the examples!
bors r+

PR #1229, opened 2021-02-22
const in ShaderStage that represents vertex and fragment access

runemopar:
We can store `ShaderStage`s as const values or members of const structs in all individual cases, but for example `ShaderStage::VERTEX | ShaderStage::FRAGMENT` cannot be stored in a const. 

The suggested change is to add another member of ShaderStage which provides a const-friendly way of storing `ShaderStage`s intended for both vertex and fragment shaders.

This is not very elegant, but would be convenient.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -566,6 +566,8 @@ bitflags::bitflags! {
         const VERTEX = 1;
         /// Binding is visible from the fragment shader of a render pipeline.
         const FRAGMENT = 2;
+        /// Binding is visible from the vertex and fragment shaders of a render pipeline.
+        const VERTEX_FRAGMENT = 3;
         /// Binding is visible from the compute shader of a compute pipeline.
         const COMPUTE = 4;
     }
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -567,7 +567,7 @@ bitflags::bitflags! {
         /// Binding is visible from the fragment shader of a render pipeline.
         const FRAGMENT = 2;
         /// Binding is visible from the vertex and fragment shaders of a render pipeline.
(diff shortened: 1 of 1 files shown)


wgpu-types/src/lib.rs
```diff
@@ -566,6 +566,8 @@ bitflags::bitflags! {
         const VERTEX = 1;
         /// Binding is visible from the fragment shader of a render pipeline.
         const FRAGMENT = 2;
+        /// Binding is visible from the vertex and fragment shaders of a render pipeline.
+        const VERTEX_FRAGMENT = 3;
```
kvark, line 570: let's assign the value as `Self::VERTEX.bits | Self::FRAGMENT.bits`

kvark (commented):
I think having this value is fine.
You can also do a workaround like:
```rust
const VERTEX_FRAGMENT: wgt::ShaderStage = wgt::ShaderStage::from_bits_truncate(wgt::ShaderStage::VERTEX.bits() | wgt::ShaderStage::FRAGMENT.bits());
```

wgpu-types/src/lib.rs
runemopar: Good call, I have committed this change

runemopar:
> You can also do a workaround like:
> 
> ```rust
> const VERTEX_FRAGMENT: wgt::ShaderStage = wgt::ShaderStage::from_bits_truncate(wgt::ShaderStage::VERTEX.bits() | wgt::ShaderStage::FRAGMENT.bits());
> ```

I employed a much less _safe_ workaround in my code, so it's nice to know that this is also possible.

That said, I think since it is not at all uncommon to have a value shared between the vertex and fragment shaders, it would not hurt to have this more intuitive option as long as it does not conflict with anything. Glad you're not opposed.

kvark (approved):

PR #1231, opened 2021-02-23
Added TextureFormatFeatures::filterable which may overwrite TextureSampleType::Float.filterable

Wumpf:
**Description**
The expectation is that with `wgt::Features::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES` enabled, any texture format which is sampled as float may be (linearly) filtered if so supported by the adapter, not only those that the webgpu specification denotes as filterable.

This arguably causes a bit of a mess in the way we distinguish `sample_type` and `(guaranteed_)format_features`: The float sample type (via spec) essentially integrates a feature, namely if it is filterable or not. Unlike the actual format _type_ which is just a property of the format, this property may or may not be available depending on the device.
So the type of `sample_type` stays statically defined whereas the Float.filterable property suddenly becomes overwritten by a format_feature which may or may not be present. Couldn't come up with a cleaner solution so far.

**Testing**
Tested in project (blub) depending on R32F filtering - passing now what was previously causing
```
wgpu error: Validation Error

Caused by:
    In Device::create_bind_group
      note: label = `BindGroup: Narrow Range filter 1`
    texture binding 1 expects sample type = Float { filterable: true }, but given a view with format = R32Float
```
(Blub still can't run on latest wgpu because of other reported issues)

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -1440,21 +1440,26 @@ impl<B: GfxBackend> Device<B> {
                                     view_samples: view.samples as u32,
                                 });
                             }
-                            match (sample_type, format_info.sample_type) {
-                                (Tst::Uint, Tst::Uint) |
-                                (Tst::Sint, Tst::Sint) |
-                                (Tst::Depth, Tst::Depth) |
-                                // if we expect non-fiterable, accept anything float
-                                (Tst::Float { filterable: false }, Tst::Float { .. }) |
-                                // if we expect fiterable, require it
-                                (Tst::Float { filterable: true }, Tst::Float { filterable: true }) |
+                            match (sample_type, format_info.sample_type, view.format_features.filterable ) {
+                                (Tst::Uint, Tst::Uint, ..) |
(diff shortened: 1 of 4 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -1440,21 +1440,26 @@ impl<B: GfxBackend> Device<B> {
...
                             }
-                            match (sample_type, format_info.sample_type) {
-                                (Tst::Uint, Tst::Uint) |
-                                (Tst::Sint, Tst::Sint) |
-                                (Tst::Depth, Tst::Depth) |
-                                // if we expect non-fiterable, accept anything float
-                                (Tst::Float { filterable: false }, Tst::Float { .. }) |
-                                // if we expect fiterable, require it
-                                (Tst::Float { filterable: true }, Tst::Float { filterable: true }) |
+                            match (sample_type, format_info.sample_type, view.format_features.filterable ) {
```
kvark, line 1443: I think this part can be refactored a bit, but I don't want to hold it. Your fix looks good!

kvark (approved):
Thank you!
bors r+

PR #1235, opened 2021-02-25
Update VertexFormat variants

kvark:
**Connections**
Matches https://github.com/gpuweb/gpuweb/pull/1322 and https://github.com/gpuweb/gpuweb/pull/1469

**Description**
Just updates us to recent upstream API.
Also breaks lots of user code!

**Testing**
Not functional


grovesNL (approved):
bors r+

Issue #1238, opened 2021-02-27
Unmapping a buffer before the mapping completes causes a panic


Imberflur:
**Description**
When attempting to map buffers in `LifetimeTracker::handle_mapping` there is a [panic](https://github.com/gfx-rs/wgpu/blob/29058e03b07ed933b108bc4e482119ce6d1cd989/wgpu-core/src/device/life.rs#L701) if the buffer is not in the `Waiting` state.
```rust
                let mapping = match std::mem::replace(
                    &mut buffer.map_state,
                    resource::BufferMapState::Idle,
                ) {
                    resource::BufferMapState::Waiting(pending_mapping) => pending_mapping,
                    _ => panic!("No pending mapping."),
                };
```
However, calling `buffer_unmap` [ignores this state](https://github.com/gfx-rs/wgpu/blob/c3a18c3d8bf2064b489c5b84e4552c54e807e1e4/wgpu-core/src/device/mod.rs#L4648) and replaces it with the `Idle` state without any type of cancellation mechanism being invoked. 
```rust
            resource::BufferMapState::Waiting(_) => {}
```
This leads to a panic the next time `handle_mapping` is called.

**Repro steps**
In the `wgpu-rs` capture example apply this diff and run:
```diff
diff --git a/examples/capture/main.rs b/examples/capture/main.rs
index 71157c2..522989f 100644
--- a/examples/capture/main.rs
+++ b/examples/capture/main.rs
@@ -126,6 +126,7 @@ async fn create_png(
     // Note that we're not calling `.await` here.
     let buffer_slice = output_buffer.slice(..);
     let buffer_future = buffer_slice.map_async(wgpu::MapMode::Read);
+    output_buffer.unmap();
 
     // Poll the device in a blocking manner so that our future resolves.
     // In an actual application, `device.poll(...)` should
```

**Expected vs observed behavior**
[The spec states](https://gpuweb.github.io/gpuweb/#dom-gpubuffer-unmap)
![image](https://user-images.githubusercontent.com/2002109/109401326-097a7e80-791c-11eb-89dc-b97db338d786.png)

Thus, I expect the `Future` resolves to a mapping aborted error rather than a panic when polling the device

Imberflur:
I believe this is fixed with #1239 and #1240

PR #1239, opened 2021-02-27
Avoid panic when requesting to unmap a buffer that is pending mapping

Imberflur:
**Connections**
Fix for #1238 

**Description**
This might not be the cleanest fix but it is quite minimal. I'm not sure what the exact intent is with the organization of things and with the async status enum, would be happy to hear critiques.

**Testing**
Tested against example in #1238


kvark (approved):
Doesn't look scary to me, thank you for digging into this!
bors r+

PR #1240, opened 2021-02-28
Fix issue with multiple queued mappings where the last one is not cancelled

Imberflur:
**Connections**
Follow up from #1239

**Description**
In #1239 I missed the case where there are multiple rounds of mapping/unmapping with the last map not being cancelled. In practice it doesn't occur very often, but I did encounter it with settings/a scene that created a heavy GPU load.

Sorry for not catching this earlier!

**Testing**
Using `wgpu-rs` capture example, after the first `map_async` I added in an `unmap` followed by another `map_async`.


kvark (approved):
Thank you!
bors r+

PR #1245, opened 2021-03-03
Trace pipeline creation before it actually happens

### LINKED ISSUE #1174: Refactor API tracing order
When reproducing a crash occurring at, say, pipeline creation, we'd want the relevant trace command to be recorded prior to the crash.

Revisit all the places like [this one](https://github.com/gfx-rs/wgpu/blob/60cd0f7e56f22a7b9d37f36559073cc844adace1/wgpu-core/src/device/mod.rs#L2602) and make it so the trace is updated before the operation takes place.

nothing comes to mind

**Additional context**
Would help investigating #1171

kvark:
**Connections**
Fixes #1174

**Description**
We want to add descriptors to the trace before panicking if anything goes wacky.

**Testing**
tested on wgpu-rs examples

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -956,7 +956,6 @@ dependencies = [
  "log",
  "num-traits",
  "petgraph",
- "serde",
  "spirv_headers",
  "thiserror",
 ]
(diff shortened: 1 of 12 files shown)


kvark:
This is done now. Notably, the refactoring saved us 40 lines 🎉 
It's much less repetitive now, and more robust. I'm looking forward to get traces from crashing users :)
bors r+

PR #1258, opened 2021-03-12
Lint all the things

### LINKED ISSUE #1250: Clean up clippy warnings
**Description**
The code could be cleaned up in a few places, according to clippy.

**Repro steps**
Just run `cargo clippy`

We should be clean from warnings. That doesn't mean we should fix everything.
For some of the warnings, like `clippy::suspicious_operation_groupings`, we'd allow them in specific places where they make sense to ignore.

**Extra materials**
Naga enforces clippy green status on CI. Maybe we should follow.

**Platform**
Any

kvark:
**Connections**
Fixes #1250

**Description**
Refactors the code to satisfy the linter, configures it, and enables on CI.

**Testing**
`cargo clippy` saves the dau

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,3 +113,19 @@ jobs:
         run: cargo clippy
       - if: matrix.channel == 'nightly'
         run: cargo test -- --nocapture
+
+  lint:
+    name: Clippy
+    runs-on: ubuntu-latest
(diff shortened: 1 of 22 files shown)


kvark:
For now, I think we'll not block on failed Clippy tests, but we'll see it right here in Github UI.
bors r+

PR #1263, opened 2021-03-14
Add CONSERVATIVE_RASTERIZATION native feature

Wumpf:
**Connections**
https://github.com/gfx-rs/gfx/pull/3672
https://github.com/gfx-rs/wgpu-rs/pull/791

**Description**
Exposes conservative rasterization as native feature + bool on `PrimitiveState`

**Testing**
New sample in wgpu-rs. Tested with D3D12 and Vulkan

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -422,6 +422,17 @@ bitflags::bitflags! {
         ///
         /// This is a native-only feature.
         const VERTEX_ATTRIBUTE_64BIT = 0x0000_0000_4000_0000;
+        /// Allows the user to set a overestimation-conservative-rasterization in [`PrimitiveState::conservative`]
+        ///
+        /// Processing of degenerate triangles/lines is hardware specific.
+        /// Only triangles are supported.
+        ///
+        /// Supported platforms:
+        /// - DX12
+        /// - Vulkan
+        ///
+        /// This is a native only feature.
+        const CONSERVATIVE_RASTERIZATION = 0x0000_0000_8000_0000;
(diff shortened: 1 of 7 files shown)


wgpu-types/src/lib.rs
```diff
@@ -922,6 +933,11 @@ pub struct PrimitiveState {
     /// Setting this to something other than `Fill` requires `Features::NON_FILL_POLYGON_MODE` to be enabled.
     #[cfg_attr(any(feature = "trace", feature = "replay"), serde(default))]
     pub polygon_mode: PolygonMode,
+    /// If set to true, the primitives are rendered with conservative overestimation. I.e. any rastered pixel touched by it is filled.
+    /// Only valid for PolygonMode::Fill!
```
kvark, line 937: is this going to be true forever? We may consider doing the `Fill { conservative: bool }` in this case.

kvark (approved):
Looking good, thank you!

wgpu-types/src/lib.rs
Wumpf: Thought of `Fill { conservative: bool }` as well, less surface to do the wrong thing. But that would break the api in a more uncomfortable way :/

Conservative point and line rendering is today possible on Vulkan only. Checking in Vulkan.gpuinfo.org looks at first like it's covered by most (extension support at 28.9%, the support for lines/points at 20.6% of all reports). But a closer look shows that it's basically an "is this an Nvidia card" check. https://vulkan.gpuinfo.org/displayextensionproperty.php?name=conservativePointAndLineRasterization&platform=windows

PR #1268, opened 2021-03-16
Move from `tracing` to `log` + `profiling`

### LINKED ISSUE #1189: Tracing should be optional (especially in dependencies)
`tracing` features are enabled unconditionally for some dependencies which can add performance overhead. It's also often used as debug feature and release builds might want to avoid it.

e.g. a crate downstream noticed a lot of overhead when tracing is enabled in gpu-alloc, but wgpu provides no way to opt out of this feature because it's always enabled: https://github.com/gfx-rs/wgpu/blob/8ed543f26a379a2d165f1592710e7f882a15b862/wgpu-core/Cargo.toml#L36-L37

Whether tracing is enabled or disabled can be controlled from the tracing crate's API, but downstream crates not interested in tracing probably shouldn't need to be aware of this.

`tracing` probably shouldn't be unconditionally enabled. Maybe wgpu could forward the existing `tracing` feature, or possibly not enable it for dependencies at all.

We could also feature-gate tracing inside of wgpu's span macro or other places, e.g. use `#[cfg(feature = "tracing")]` to avoid creating tracing spans in https://github.com/gfx-rs/wgpu/blob/780637eec029097e37dff9238abca7248354b465/wgpu-core/src/lib.rs#L251

Downstream crates opting out of tracing by using the API in the `tracing` crate

**Additional context**
It was mentioned on Discord that significant amounts of time are spent in gpu-alloc's tracing and places like https://github.com/gfx-rs/wgpu/blob/f0556d7d8b05295956248d6384cffb76bdb12707/wgpu-core/src/command/render.rs#L2072

We might also want to consider making tracing optional in wgpu-rs for some of the same reasons above.

### LINKED ISSUE #1211: Convert `tracing` to `log` + `profiling`
`tracing` has some performance implications and also mixes profiling spans and with logging. Due to a discussion on the matrix, we've decided to move all spans to the `profiling` crate (with no features enabled) and all logging to the `log` crate.

cwfitzgerald:
**Connections**

Fixes #1211. Fixes #1189.

**Description**

Moves from `tracing` to `log` + `profiling`. I also removed some of the offending spans that were only for pushing commands onto a vector of commands.

**Testing**

Tested on wgpu-rs

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -37,15 +37,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
-dependencies = [
- "winapi 0.3.9",
-]
-
-[[package]]
 name = "arrayvec"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -161,19 +152,6 @@ name = "cfg_aliases"
 version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fd16c4719339c4530435d38e511904438d07cce7950afa3718a84ac36c10e89e"
-
-[[package]]
-name = "chrono"
-version = "0.4.19"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "670ad68c9088c2a963aaa298cb369688cf3f9465ce5e2d4ca10e6e0098a1ce73"
-dependencies = [
- "libc",
- "num-integer",
- "num-traits",
- "time",
- "winapi 0.3.9",
-]
 
 [[package]]
 name = "cloudabi"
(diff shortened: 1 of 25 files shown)


Cargo.lock
```diff
@@ -36,15 +36,6 @@ dependencies = [
  "xml-rs",
 ]
 
-[[package]]
```
kvark, line 39: whoa, 190 LOC cleared up from our Cargo lock?
I'm eager to recount our dependencies after this :)

wgpu-core/Cargo.toml
```diff
@@ -27,16 +27,17 @@ arrayvec = "0.5"
 bitflags = "1.0"
 copyless = "0.1"
 fxhash = "0.2"
+log = { version = "0.4" }
```
kvark, line 30: is it intentionally not just `log = "0.4"`?

kvark (approved):
LGTM, thank you!
please proceed when ready

Cargo.lock
cwfitzgerald: 88 build jobs -> 84 building just wgpu-core

wgpu-core/Cargo.toml
cwfitzgerald: Nah, just expected to need to turn off features, but didn't need to.

cwfitzgerald:
bors r=kvark

Got https://github.com/aclysma/profiling/pull/13 in there too.

PR #1275, opened 2021-03-20
rows_per_image & rows_per_image are now optional

### LINKED ISSUE #988: Convert bytes_per_row and rows_per_image to be optional
These fields are now optional in the spec. The idiomatic way to represent this would be turning `u32` into `Option<NonZeroU32>`.

Wumpf:
Fixes #988
Was surprisingly hard, hope it's correct; testing could be more extensive...
Ideas for more elegant handling very welcome.

**Testing**
Tested against wgpu-rs samples

### DIFF
--- a/wgpu-core/src/command/transfer.rs
+++ b/wgpu-core/src/command/transfer.rs
@@ -181,7 +181,7 @@ pub(crate) fn validate_linear_texture_data(
     let bytes_per_row = if let Some(bytes_per_row) = layout.bytes_per_row {
         bytes_per_row.get() as BufferAddress
     } else {
-        if copy_depth > 1 || block_height > 1 {
+        if copy_depth > 1 || height_in_blocks > 1 {
             return Err(TransferError::UnspecifiedBytesPerRow);
         }
         bytes_per_block * width_in_blocks
--- a/wgpu-core/src/command/transfer.rs
+++ b/wgpu-core/src/command/transfer.rs
@@ -76,6 +76,10 @@ pub enum TransferError {
     UnalignedBytesPerRow,
     #[error("number of rows per image is not a multiple of block height")]
     UnalignedRowsPerImage,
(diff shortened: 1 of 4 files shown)


wgpu-core/src/command/transfer.rs
```diff
@@ -157,21 +161,40 @@ pub(crate) fn validate_linear_texture_data(
...
     let block_height = block_height as BufferAddress;
     let block_size = bytes_per_block;
 
+    let width_in_blocks = copy_width / block_width;
+    let height_in_blocks = copy_height / block_height;
+
+    let bytes_per_row = if let Some(bytes_per_row) = layout.bytes_per_row {
+        bytes_per_row.get() as BufferAddress
+    } else {
+        if copy_depth > 1 || block_height > 1 {
```
kvark, line 184: this should be
```
if copy_depth > 1 || height_in_blocks > 1 {
```

wgpu-core/src/device/queue.rs
```diff
@@ -397,21 +403,25 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
 
         let ptr = stage.memory.map(&device.raw, 0, stage_size)?;
         unsafe {
+            let bytes_per_row = if let Some(bytes_per_row) = data_layout.bytes_per_row {
```
kvark, line 406: can this be moved out of `unsafe {}`?

kvark (commented):
Amazing work, thank you for making this real!
I see how it got more complicated than we hoped, but everything looks great so far. There is only one thing I noticed that needs addressing.

wgpu-core/src/command/transfer.rs
Wumpf: very nice catch! That would have broken all compressed single row texture updates 😄

PR #1276, opened 2021-03-22
Align create_swap_chain function signature

kvark:
**Connections**
Follow-up to #1034

**Description**
Swapchains are a bit special. Browsers don't use wgpu-core's swapchains, so we never bothered to convert them into the error model. But we still need this for Deno and webgpu-headers.

**Testing**
Untested, but also harmless

### DIFF
--- a/player/src/bin/play.rs
+++ b/player/src/bin/play.rs
@@ -141,7 +141,10 @@ fn main() {
                                 resize_desc = Some(desc);
                                 break;
                             } else {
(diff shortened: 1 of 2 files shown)


cwfitzgerald (approved):
bors r+

cwfitzgerald:
Seems we have some compile errors on CI

kvark:
bors r=cwfitzgerald

PR #1278, opened 2021-03-23
Rename copy views and layout to match spec

grovesNL:
**Description**
Match the specification naming changes from https://github.com/gpuweb/gpuweb/pull/1375

**Testing**
None because the changes downstream to wgpu-rs and wgpu-native should be trivial (renaming only)


kvark (approved):
Very nice, thank you!
bors r+

PR #1284, opened 2021-03-25
Return Rgba8UnormSrgb instead of Bgra8UnormSrgb as default swapchain format for WebGL target

Gordon-F:
WebGL doesn't supports BGRA texture formats

### DIFF
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -317,8 +317,14 @@ impl<B: GfxBackend> Adapter<B> {
             return Err(GetSwapChainPreferredFormatError::NotFound);
         }
 
-        // If no formats were returned, use Bgra8UnormSrgb
-        Ok(wgt::TextureFormat::Bgra8UnormSrgb)
+        let is_webgl = cfg!(gl) && cfg!(target_arch = "wasm32");
+
+        // If no formats were returned, use Bgra8UnormSrgb (native/WebGPU) and Rgba8UnormSrgb (WebGL)
+        if is_webgl {
+            Ok(wgt::TextureFormat::Rgba8UnormSrgb)
+        } else {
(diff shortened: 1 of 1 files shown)


wgpu-core/src/instance.rs
```diff
@@ -317,8 +317,14 @@ impl<B: GfxBackend> Adapter<B> {
             return Err(GetSwapChainPreferredFormatError::NotFound);
         }
 
-        // If no formats were returned, use Bgra8UnormSrgb
-        Ok(wgt::TextureFormat::Bgra8UnormSrgb)
+        let is_webgl = cfg!(gl) && cfg!(target_arch = "wasm32");
```
kvark, line 320: I don't think we should be fixing this at the wgpu-core level.
The semantics of [supported_formats](https://docs.rs/gfx-hal/0.7.0/gfx_hal/window/trait.Surface.html#tymethod.supported_formats) is:
> If None is returned then the surface has no preferred format and the application may use any desired format.

Obviously, this isn't the case for WebGL. So our OpenGL backend has to return a reasonable set of formats here.

kvark (changes requested):

wgpu-core/src/instance.rs
Gordon-F: gfx-rs fix - https://github.com/gfx-rs/gfx/pull/3699

> Obviously, this isn't the case for WebGL. So our OpenGL backend has to return a reasonable set of formats here

Why we return any formats here if `supported_formats` is none instead of return `None` to users?

kvark: Here? Because in `wgpu` the `get_swap_chain_preferred_format` has to return one format.

Gordon-F: Thank you for explain! gfx-rs fix should be enough.

PR #1289, opened 2021-03-27
Add color blend modes

cwfitzgerald:
**Connections**

None

**Description**

Blend modes are extremely confusing, so I think it is beneficial to have defaults for common color blending solutions.

**Testing**

Stared at it. I _think_ they're right, but I hate blending.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -797,6 +797,34 @@ pub struct BlendState {
     pub alpha: BlendComponent,
 }
 
+impl BlendState {
+    /// Blend mode that does no color blending, just overwrites the output with the contents of the shader.
+    pub const REPLACE: Self = Self {
+        color: BlendComponent::REPLACE,
+        alpha: BlendComponent::REPLACE,
+    };
+
+    /// Blend mode that does standard alpha blending with non-premultiplied alpha.
+    pub const ALPHA_BLENDING: Self = Self {
+        color: BlendComponent {
+            src_factor: BlendFactor::SrcAlpha,
+            dst_factor: BlendFactor::OneMinusSrcAlpha,
+            operation: BlendOperation::Add,
+        },
+        alpha: BlendComponent::REPLACE,
+    };
+
+    /// Blend mode that does standard alpha blending with premultiplied alpha.
+    pub const PREMULTIPLIEDALPHA_BLENDING: Self = Self {
+        color: BlendComponent {
+            src_factor: BlendFactor::One,
+            dst_factor: BlendFactor::OneMinusSrcAlpha,
+            operation: BlendOperation::Add,
+        },
+        alpha: BlendComponent::REPLACE,
+    };
+}
+
 /// Describes the color state of a render pipeline.
 #[repr(C)]
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]


wgpu-types/src/lib.rs
```diff
@@ -764,6 +764,22 @@ impl BlendComponent {
...
+    /// The alpha blend mode often used is REPLACE.
+    pub const COLOR_BLEND: Self = BlendComponent {
+        src_factor: BlendFactor::SrcAlpha,
+        dst_factor: BlendFactor::OneMinusSrcAlpha,
+        operation: BlendOperation::Add,
+    };
+
+    /// Color blend mode that does standard alpha blending with premultiplied alpha.
+    /// The alpha blend mode often used is REPLACE.
+    pub const PREMULTIPLIED_COLOR_BLEND: Self = BlendComponent {
```
aloucks, line 777: What do you think of `COLOR_BLEND_PREMULTIPLIED_ALPHA` instead?

cwfitzgerald: sure

wgpu-types/src/lib.rs
```diff
@@ -764,6 +764,22 @@ impl BlendComponent {
         operation: BlendOperation::Add,
     };
 
+    /// Color blend mode that does standard alpha blending with non-premultiplied alpha.
+    /// The alpha blend mode often used is REPLACE.
+    pub const COLOR_BLEND: Self = BlendComponent {
```
kvark, line 769: Isn't this generally called "alpha blending"? I.e. maybe I'm thinking about this the wrong way, but the model I have in the head is that "blending" term by itself assumes the color, and "alpha blending" means mix colors based on the alpha. From that perspective, "color blend" does not make much sense.

wgpu-types/src/lib.rs
```diff
@@ -764,6 +764,22 @@ impl BlendComponent {
...
+        src_factor: BlendFactor::SrcAlpha,
+        dst_factor: BlendFactor::OneMinusSrcAlpha,
+        operation: BlendOperation::Add,
+    };
+
+    /// Color blend mode that does standard alpha blending with premultiplied alpha.
+    /// The alpha blend mode often used is REPLACE.
+    pub const COLOR_BLEND_PREMULTIPLIED_ALPHA: Self = BlendComponent {
```
kvark, line 777: if the first one is `ALPHA_BLEND`, then this one would be `PREMULTIPLIED_ALPHA_BLEND`

kvark (commented):
Sorry for name bikeshed!

wgpu-types/src/lib.rs
cwfitzgerald: I'm just worried that people will put this in the alpha with REPLACE in color if it's called alpha. Idk if there's a pretty solution to this.

kvark: ah I see. I think one solution could be to define these constants on `impl BlendState` instead of the blend component. Did you consider this? It would be even more ergonomic this way.

cwfitzgerald: Alternative proposal: `ALPHA_BLEND_COLOR_COMPONENT` and `PREMULTIPLIED_ALPHA_BLEND_COLOR_COMPONENT`, which I think is a bit more explicit.

kvark: do you see the need for this (versus just `BlendState` constants)?

cwfitzgerald: I'd be happy with blend state constants, which would eliminate this issue entirely.

kvark: let's just do it then

PR #1291, opened 2021-03-29
Add color blend modes

cwfitzgerald:
Continuation of #1289, because apparently you can't re-open an accidentally closed pull request if you force pushed...

Should be GTG.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -764,6 +764,13 @@ impl BlendComponent {
         operation: BlendOperation::Add,
     };
 
+    /// Blend state of (1 * src) + ((1 - src_alpha) * dst)
+    pub const OVER: Self = BlendComponent {
+        src_factor: BlendFactor::One,
+        dst_factor: BlendFactor::OneMinusSrcAlpha,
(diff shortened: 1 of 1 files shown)


wgpu-types/src/lib.rs
```diff
@@ -797,6 +797,34 @@ pub struct BlendState {
...
+    };
+
+    /// Blend mode that does standard alpha blending with non-premultiplied alpha.
+    pub const ALPHA_BLENDING: Self = Self {
+        color: BlendComponent {
+            src_factor: BlendFactor::SrcAlpha,
+            dst_factor: BlendFactor::OneMinusSrcAlpha,
+            operation: BlendOperation::Add,
+        },
+        alpha: BlendComponent::REPLACE,
```
kvark, line 814: Let's talk about this. I don't think `REPLACE` is what a typical blend equation has. Maybe `ADD`?

kvark (approved):
May be tricky, but we agreed to proceed and try this out
bors r+

kvark:
Oh wait, this code uses REPLACE?
bors r-

kvark (approved):
Alright, let's proceed!
bors r+

Issue #1244, opened 2021-03-03
Tracking Issue and API Proposal for wgpu on "Downlevel" APIs and Hardware


cwfitzgerald:
# Definitions

**Downlevel**: An API or piece of hardware that does not support the feature level required to allow full webgpu API usage. Originally used for DX11's support for DX9 and DX10 hardware, now used in the context of wgpu.
**Platform**: A combination of the hardware, API, and driver in use. These come together to create the features available to the user.

# Intro

We all agree that we want wgpu to eventually be able to run on basically any hardware we throw at it. This means we will add support for APIs and hardware that cannot offer the full webgpu ~~experience~~ spec. When running on such APIs or devices, wgpu needs to be able to throw validation errors when the user uses more features than the backend supports, avoiding potential UB in the backend. In addition, this information should be provided to the user so they can smoothly fall back to implementations that don't use the unsupported features, or smoothly exit if they have no fallback path.

This issue contains two sections. First an API proposal on how this should be exposed to the user, then a list of downlevel platforms that are considered valuable to support.

# API Proposal

wgpu as an API should assume that the backend is fully featured by default. These efforts are for the people who want to go out of their way to support older platform. As such, the proposed API will not require the user to enable features, all features and limits available will be used to their maximum values.

This API will parallel the Features/Limits API, adding a DownlevelFeatures and DownlevelLimits bitflags and struct.

Two methods would be added:

```rust
fn Adapter::webgpu_compatability(&self) -> Compatibility;
fn Device::webgpu_compatability(&self) -> Compatibility;
```

The `Compatability` struct would be defined as the following.

```rust
pub struct Compatibility {
    pub features: DownlevelFeatures,
    pub limits: DownlevelLimits,
}

impl Compatibility {
    pub fn is_conformant(&self) -> bool,
}
```

When targeting WebGPU through wasm, this would always return "Conformant".

To give an example of things that would be in the feature and limits struct, here are some possibilities.

```rust
bitflags! {
    pub struct DownlevelFeatures: u128 {
        // FL10 and below hardware (and webgl) doesn't necessarily support compute.
        const COMPUTE_SHADERS = 0x0000_0000_0000_0001; 
		// SM2 hardware doesn't support non-constant control flow.
        const NON_CONSTANT_CONTROL_FLOW = 0x0000_0000_0000_0002;
    }
}

pub struct DownlevelLimits: u128 {
	// DX11 limits UAV + RTV to 8, DX11_1 to 64.
[...]

PR #1297, opened 2021-03-31
Downlevel Infrastructure and Validate Compute Shader Presence

cwfitzgerald:
**Connections**

Provides the core infrastructure for #1244.

**Description**

Nothing too crazy in here. Includes the infrastructure and the beginnings of validation. Some of the properties can't easily be verified, but I will follow up with the rest of the low-hanging fruit in another PR.

**Testing**

Testing on https://github.com/gfx-rs/wgpu-rs/pull/829.

### DIFF
--- a/wgpu-core/src/lib.rs
+++ b/wgpu-core/src/lib.rs
@@ -237,6 +237,17 @@ struct PrivateFeatures {
     texture_d24_s8: bool,
 }
 
+const DOWNLEVEL_WARNING_MESSAGE: &str = "The underlying API or device in use does not \
+support enough features to be a fully compliant implementation of WebGPU. A subset of the features can still be used. \
+If you are running this program on native and not in a browser and wish to limit the features you use to the supported subset, \
+call Adapter::downlevel_properties or Device::downlevel_properties to get a listing of the features the current \
+platform is supports.";
+const DOWNLEVEL_ERROR_WARNING_MESSAGE: &str = "This is not an invalid use of WebGPU: the underlying API or device does not \
+support enough features to be a fully compliant implementation. A subset of the features can still be used. \
+If you are running this program on native and not in a browser and wish to work around this issue, call \
+Adapter::downlevel_properties or Device::downlevel_properties to get a listing of the features the current \
+platform is supports.";
+
 #[macro_export]
 macro_rules! gfx_select {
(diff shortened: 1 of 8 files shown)


wgpu-types/src/lib.rs
```diff
@@ -538,6 +538,65 @@ impl Default for Limits {
     }
 }
 
+/// Lists various ways the underlying platform does not conform to the WebGPU standard.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct DownlevelProperties {
+    /// The device supports compiling and using compute shaders.
+    pub compute_shaders: bool,
```
kvark, line 545: should these be flags?

cwfitzgerald: Up to you. The latest thing we discussed was because they are passive they should be boolean members, but I don't really have strong feelings.

kvark: oh, sorry, that's not what my reasoning was last time.
I thought the decision was to whether to include them in `Features` or not, it wasn't about the form of inclusion.
They shouldn't be in features, but they can be in their own `DownlevelFlags` thing, which would live inside `DownlevelProperties`.
Anyway, I don't feel strongly either way.

kvark (approved):
please proceed when feel ready!

wgpu-core/src/lib.rs
```diff
@@ -237,6 +237,17 @@ struct PrivateFeatures {
     texture_d24_s8: bool,
 }
 
+const DOWNLEVEL_WARNING_MESSAGE: &str = "The underlying API or device in use does not \
+support enough features to be a fully compliant implementation of WebGPU. A subset of the features can still be used. \
+If you are running this program on native and not in a browser and wish to limit the features you use to the supported subset, \
+call Adapter::downlevel_properties or Device::downlevel_properties to get a listing of the features the current \
+platform is supports.";
```
cwfitzgerald, line 244:
```suggestion
platform supports.";
```
typo

wgpu-types/src/lib.rs
cwfitzgerald: Ahh I get you. I think that's a reasonable balance, I'll fix it in the morn.

cwfitzgerald:
Adjusted based on feedback, we should be good to go!

bors r=kvark

PR #1308, opened 2021-04-05
[blocked] Add a pipeline cache to the device

expenses:
**Connections**
As discussed on Element, it would be good to have a native wgpu feature that allows for serializing and deserializing the pipeline cache. This would allow for a potential speed up for applications that build a lot of pipelines.

**Description**
Before that, we actually need to have a pipeline cache to begin with. This adds that struct to the `Device` and passes it to the backend when creating pipelines.

A few points of interest:

* There isn't any other `unsafe` code in `Device::new` so I'm not sure if that's the right way to do things. Creating a pipeline cache seems very safe though, all things considered.
* We probably want a feature like `wgpu::SERIALIZABLE_PIPELINE_CACHE` or something for this. In the case that this feature isn't enabled, does it still make sense to create a pipeline cache or should we wrap it in an `Option` instead?

**Testing**

I've ran some of the wgpu-rs examples with this branch set on vulkan (a backend with the pipeline cache implemented) and there haven't been any problems.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -286,6 +286,7 @@ pub struct Device<B: hal::Backend> {
     pub(crate) features: wgt::Features,
     pub(crate) downlevel: wgt::DownlevelProperties,
     spv_options: naga::back::spv::Options,
+    pipeline_cache: B::PipelineCache,
     //TODO: move this behind another mutex. This would allow several methods to switch
     // to borrow Device immutably, such as `write_buffer`, `write_texture`, and `buffer_unmap`.
     pending_writes: queue::PendingWrites<B>,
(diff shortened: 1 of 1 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -286,6 +286,7 @@ pub struct Device<B: hal::Backend> {
     pub(crate) features: wgt::Features,
     pub(crate) downlevel: wgt::DownlevelProperties,
     spv_options: naga::back::spv::Options,
+    pipeline_cache: B::PipelineCache,
```
kvark, line 289: we'll need to think more about this before proceeding

expenses:
I'll mark this as a draft for now.

cwfitzgerald:
I'm going to close this for now as we're now on wgpu-hal and this code is mostly for gfx-hal. Please re-open if this gets forward ported!

PR #1312, opened 2021-04-07
Fixes vulkan-portability feature panicking on MacOS

rybertm:
**Connections**
Fixes [#840 (wgpu-rs)](https://github.com/gfx-rs/wgpu-rs/issues/840)

**Description**
`vulkan-portability` feature panics on MacOS since it doesn't have the `gfx-backend-vulkan` feature needed by wgpu-core to enable the Vulkan backend.
This PR simply checks for the `vulkan-portability` feature instead, as suggested in the issue linked above.

**Testing**
Not really tested since `wgpu` won't build locally, but it is just a straight forward replacement of a feature check.

### DIFF
--- a/wgpu-core/src/lib.rs
+++ b/wgpu-core/src/lib.rs
@@ -254,7 +254,7 @@ macro_rules! gfx_select {
         // Note: For some reason the cfg aliases defined in build.rs don't succesfully apply in this
         // macro so we must specify their equivalents manually
         match $id.backend() {
-            #[cfg(all(not(target_arch = "wasm32"), any(not(any(target_os = "ios", target_os = "macos")), feature = "gfx-backend-vulkan")))]
+            #[cfg(all(not(target_arch = "wasm32"), any(not(any(target_os = "ios", target_os = "macos")), feature = "vulkan-portability")))]
             wgt::Backend::Vulkan => $global.$method::<$crate::backend::Vulkan>( $($param),* ),
             #[cfg(all(not(target_arch = "wasm32"), any(target_os = "ios", target_os = "macos")))]
             wgt::Backend::Metal => $global.$method::<$crate::backend::Metal>( $($param),* ),


kvark:
I'm curious why "wgpu won't build locally". What kind of problem are you facing there?

wgpu-core/src/lib.rs
```diff
@@ -254,7 +254,7 @@ macro_rules! gfx_select {
         // Note: For some reason the cfg aliases defined in build.rs don't succesfully apply in this
         // macro so we must specify their equivalents manually
         match $id.backend() {
-            #[cfg(all(not(target_arch = "wasm32"), any(not(any(target_os = "ios", target_os = "macos")), feature = "gfx-backend-vulkan")))]
+            #[cfg(all(not(target_arch = "wasm32"), any(not(any(target_os = "ios", target_os = "macos")), feature = "vulkan-portability")))]
```
kvark, line 257: let's leave a comment saying why the check if for a feature that this crate doesn't have

rybertm: Is 
```
"This macro gets mostly used in the `wgpu-rs` crate, which defines the `vulkan-portability` feature, thus the need for a check to it here"
```
sufficient? I want approval before committing so I don't commit many times for a simple change.

rybertm:
> I'm curious why "wgpu won't build locally". What kind of problem are you facing there?

Last time I tried It showed some erros in some macros if I remember correctly, but now it compiled no problem, not sure what happened, but nevermind that comment then.

wgpu-core/src/lib.rs
kvark: Come to think of it again, we shouldn't go this route at all. wgpu-core doesn't have this feature, so adding checks for it makes it a very pool abstraction. Instead, it should be written in a way that self-contained within wgpu-core.
I think there is such a way: instead of checking for the feature inside the macro, how about we try checking it outside?
```rust
#[cfg(feature = "gfx-backend-vulkan")]
#[macro_export]
macro_rules! gfx_select {
}
#[cfg(not(feature = "gfx-backend-vulkan"))]
#[macro_export]
macro_rules! gfx_select {
}
```
Would you want to try out this approach?

rybertm: Sure, I can try that.
If I understand correctly, in the case that `gfx-backend-vulkan` isn't enabled, should the Vulkan backend not be a option at all?
And if it is enabled, should I assume only Vulkan or GL backend depending on wasm to be the architecture or not?

kvark: > If I understand correctly, in the case that gfx-backend-vulkan isn't enabled, should the Vulkan backend not be a option at all?

My pseudo-code was obviously incomplete.
We need to double check how `feature = xxx` works if `xxx` is non-optional dependency.
The Vulkan backend is always either unconditionally enabled, or behind a feature.

> should I assume only Vulkan or GL backend depending on wasm to be the architecture or not?

Only GL backend makes sense for WASM here

rybertm: I didn't have time to work on this with uni exams and work, but when I can, is it still beneficial for me to work on this since you removed support for it on Apple Platforms?

kvark: It depends on whether you need those Vulkan Validation layers, or just looking for things to help us with.
if you are just looking to help, then no need to resume this work. There are plenty of issues open!

PR #1313, opened 2021-04-08
Don't panic on change_extend errors

kvark:
**Connections**
???

**Description**
We used to have leftover `unwrap()` for extending the usages of render targets.
Now these are all routed via `UsageConflicts` into actual errors.

**Testing**
Untested


cwfitzgerald (approved):
Nice clean up
bors r+

PR #1322, opened 2021-04-12
Add Map Alignment type

Dispersia:
**Description**
 In wgpu-rs, StagingBelt is aligning to 4 bytes instead of 8, failing the new validations. Adds a new type to specify mapping alignment size

**Testing**
 No testing needed afaik, only added new.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -28,7 +28,7 @@ use parking_lot::{Mutex, MutexGuard};
 use thiserror::Error;
 use wgt::{
     BufferAddress, BufferSize, InputStepMode, TextureDimension,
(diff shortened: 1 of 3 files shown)


kvark (commented):
Thank you!
Please also fix the `buffer_map_async`. The `range.start` has to be aligned to `wgt::MAP_ALIGNMENT`.

Dispersia:
thank you for the comment, does it also have to do the alignment for the remainder, or is that meant to be just for the staging belt? I don't fully understand the usage of that method, thanks!

kvark:
The mapping needs to have the offset aligned to 8 but the size - only to 4. So we don't need to change the size (or the remainder) alignment checks, only the offsets.

kvark (approved):
Beautiful, thank you!

PR #1324, opened 2021-04-13
Allow non-storage textures to be able to be queried in validation

expenses:
**Description**
I have a glsl shader that uses `textureSize` which appears to requires the `GlobalUse::QUERY` flag but we don't set it for non-storage textures. This PR changes that and also improves the `WrongUsage` error.

**Testing**
My pipeline now validates correctly :^)

### DIFF
--- a/wgpu-core/src/validation.rs
+++ b/wgpu-core/src/validation.rs
@@ -117,10 +117,10 @@ pub enum BindingError {
     Missing,
     #[error("visibility flags don't include the shader stage")]
     Invisible,
-    #[error("load/store access flags {actual:?} don't match the shader flags {expected:?}")]
+    #[error("The shader requires the load/store access flags {required:?} but only {allowed:?} is allowed")]
     WrongUsage {
-        actual: GlobalUse,
-        expected: GlobalUse,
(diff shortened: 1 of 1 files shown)


expenses:
`error matching global binding at index 1 in group 0 against the pipeline layout: The shader requires the load/store access flags READ | QUERY but only READ is allowed`
It's still a bit hard to understand if you don't get what naga validation is doing, but I think its a lot better.

wgpu-core/src/validation.rs
```diff
@@ -346,7 +349,7 @@ impl Resource {
                             },
                             wgt::TextureSampleType::Depth => naga::ImageClass::Depth,
                         };
-                        (class, GlobalUse::READ)
+                        (class, GlobalUse::READ | GlobalUse::QUERY)
```
kvark, line 352: let's also enable this for uniform buffers, read-only storage buffers, and all of the storage textures

kvark (changes requested):
Thank you for the PR!
I think we need to flip a few more bits to make it complete

kvark (approved):
Thank you!

PR #1325, opened 2021-04-14
Relax and imporve vertex attribute validation

kvark:
**Connections**
None

**Description**
Improves our validation of vertex attributes in the following ways:
  1. Relaxes the size matching, so that we allow the shaders to default some of the values we don't provide
  2. Require interpolation to match
  3. Make the error messages more detailed:

>     In Device::create_render_pipeline
    error in stage VERTEX
    error matching the stage input at 1 (Uint32x2 interpolated as None) against the previous stage outputs
    input type is not compatible with the provided Float32x2

**Testing**
Tested on wgpu-rs examples.


cwfitzgerald (approved):
bors r+

PR #1330, opened 2021-04-14
Fix clippy by removing `.clone()` for types that implement Copy

adamnemecek:


kvark:
Got already implemented

PR #1332, opened 2021-04-15
Validate min/max blending factors

kvark:
**Connections**
Based on #1325 
Matches https://github.com/gpuweb/gpuweb/pull/1635

**Description**
This PR refactors the color state errors, giving them more detailed context, and having a dedicated enum for them.
It also adds the check for the blend factors to be (1, 1) for the min/max operations.

**Testing**
Untested


kvark:
Rebased now

PR #1334, opened 2021-04-15
Don't assume a pipeline is always bound

kvark:
**Connections**
Was looking at https://bugzilla.mozilla.org/show_bug.cgi?id=1705043 trying to figure out what's going on there.

**Description**
This should fix a scenario where one starts with binding groups, and then binds the pipeline.

**Testing**
Tested on wgpu-rs examples


kvark:
Turns out this isn't really a problem. The lack of pipeline layout is handled at the lower level.

PR #1335, opened 2021-04-17
Native Extension: Expose fill buffer & clear image

Wumpf:
**Description**

Adds a native extension that allows access to fill buffer and clear image methods of the backend. Gfx already emulates this where necessary. Available directly on the encoder, very similar to buffer/texture copy operations.

At first I was thinking to name it `fill_buffer` and `fill_texture` for consistency, but figured I stick with Vulkan's namings as they are more "mainstream".
For simplicity, the `clear_image` only allows one subresource-range at a time.
Wanted to keep the whole thing as separate as possible from everything which causes a bit duplication especially with error states, but I think that's better than polluting anything else in the process.

**Testing**
Added two new player tests.


wgpu-core/src/device/trace.rs
```diff
@@ -142,6 +142,17 @@ pub enum Command {
         dst: crate::command::ImageCopyTexture,
         size: wgt::Extent3d,
     },
+    FillBuffer {
+        dst: id::BufferId,
+        offset: wgt::BufferAddress,
+        size: Option<wgt::BufferAddress>,
```
kvark, line 148: should use `wgt::BufferSize`

wgpu-core/src/command/fill_clear.rs
```diff
@@ -0,0 +1,297 @@
...
+    track::TextureSelector,
+};
+
+use hal::command::CommandBuffer as _;
+use thiserror::Error;
+use wgt::{BufferAddress, BufferUsage, Color, ImageSubresourceRange, TextureAspect, TextureUsage};
+
+/// Error encountered while attempting a clear.
+#[derive(Clone, Debug, Error)]
+pub enum FillOrClearError {
```
kvark, line 26: it would be great to come up with a single name for this operation. clear_buffer and clear_image sound good to me. We'd then name this module `clear.rs`, and have the error enum called `ClearError`.

wgpu-core/src/command/fill_clear.rs
```diff
@@ -0,0 +1,297 @@
...
+        subresource_layer_count: Option<hal::image::Layer>,
+    },
+}
+
+impl<G: GlobalIdentityHandlerFactory> Global<G> {
+    pub fn command_encoder_fill_buffer<B: GfxBackend>(
+        &self,
+        command_encoder_id: CommandEncoderId,
+        dst: BufferId,
+        offset: BufferAddress,
```
kvark, line 75: what are the requirements for this offset alignment?

kvark (commented):
This is very impressive, thank you for writing it!

I think we need to talk about thus a little bit before proceeding. In particular, I'm not confident that just exposing the gfx-hal features in a safe way is the best route for us to follow.

Philosophically, we wanted to make gfx-hal to polyfill less. That is, we'd want to steer it into direction where it just communicates higher up which functions are available, and doesn't expose anything it can't natively implement. DX11 for example doesn't have `fill_buffer` at all, and Metal has it in a way that only accepts bytes... We have to work around it, today, in gfx-hal, but for wgpu we'd rather not rely on such workarounds, if possible.

There needs to be a small investigation happening about which of these functions are available in the backends. I could roughly see the following spectrum of decisions:
  1. expose what we have in gfx today (what this PR does)
  2. find an intersection of at least the primary backends and expose it
  3. expose only clearing to zero, which is already implicitly a thing in WebGPU, since it has to initialize resource to zero. This would likely be a subset of (2).
  4. don't expose anything

Wumpf:
Ah not the reaction I was hoping for but the way you lay it out makes a lot of sense to me!

I'll do a short survey of what the main APIs offer and what restrictions they have in similar style to the surveys that are on the WebGPU repo! Then this also has more of a chance to become part of future WebGPU revisions :)

I reckon `3) expose only clearing to zero` is probably a thing that would make sense regardless, because as you say the spec already says that wgpu *must* implement this. Ergo even if gfx takes a different turn and puts less effort into its clear/fill emulation, wgpu will still have at least this very limited set and could expose it to native users 🤔

wgpu-core/src/command/fill_clear.rs
Wumpf: agreed! Very messy right now because I sticked so much with the underlying gfx/vulkan api names

kvark:
Yeah, (3) is definitely the safest to land right now.

Wumpf:
Haven't forgotten about this, just busy with other things atm. I still plan on seeing this through :)

Wumpf:
@kvark I was about to create an API overview on the topic, but quickly ended up again at the report you already did a couple of years ago: https://github.com/gpuweb/gpuweb/issues/28
(which I even commented on not all that long .. oof been away from this stuff for too long :))

This however doesn't mention Metal's [buffer fill](https://developer.apple.com/documentation/metal/mtlblitcommandencoder/2928191-fill) which you mentioned earlier here which indeed is a bytewise buffer fill making this the smallest common denominator.

a "least code" backend implementation of a fill buffer would then rely on
* Vulkan: vkCmdFillBuffer
* DX12: ClearUnorderedAccessViewUint
* Metal: fill(buffer:range:value:)

We can still take a uint8 value, but everything beyond would require emulation (or rather would not allow us to simplify this in the backend - today gfx fill_buffer is always a compute shader on Metal). Debatable if that's useful, but nothing preventing it.
Buffer range would be fine but need to be aligned to 4 bytes.


Texture fill is an entirely different story. Vulkan is again great in this regard but DX12 already poses a problem since it only allows UAV and RenderTargets/DepthStencil to be cleared. If that were an actual limitation of a user facing clear image method, one might as well drop it since a render pass with clear would do the job as well.
This makes it compelling to drop the feature altogether. Then again we still need to implement texture clearing, but this was identified as a slow pass, so better not to rush this

Therefore, I'd suggest to only have a `fill_buffer` ("fill" is more common than clear it seems) which takes a uint8 values and a buffer memory range that is aligned to 4.
Thoughts, counterarguments?

kvark:
> We can still take a uint8 value, but everything beyond would require emulation (or rather would not allow us to simplify this in the backend - today gfx fill_buffer is always a compute shader on Metal). Debatable if that's useful, but nothing preventing it.
Buffer range would be fine but need to be aligned to 4 bytes.

I think it's useful as an extension for the current "uninitialized" semantics. I.e. if the user already knows that there is logic zero-initializing buffers, it would be good to also expose this, maybe they'll find it useful.
Note: gfx-backend-metal would add a fast-path to avoid the compute shader in this scenario

kvark:
For textures, let's only consider clearing to zero, as a way to (once again) expose the logic that an implementation has to have, anyway. Once we think about it this way, it doesn't matter what gfx backends are going to be doing, since they have to do this regardless (of whether or not a native feature is exposed, since we need to handle uninitialized data).

I just realized something probably important. The safest and easiest way for us to expose this is via piggy-backing on the existing "uninitialized" semantics. But this semantics works at the device timeline, so the methods should also be on the device timeline. It would be something like this:
```
buffer.reset_range(4..16);
texture.reset_range(0..1, 10..20); // mipmaps, then layers
```

The implementation would just mark these regions as uninitialized, and do nothing else on them. So it's really tiny.

Wumpf:
That's a cool idea but operating on the device timeline kinda defeats the purpose I was going for again.
Then again I'm not entirely clear why this would need to be on the device timeline since marking as initialized is not - it will need to work similar to the current buffer clear implementation where we react on the queue timeline to initialization states

kvark:
I may be getting it totally wrong. I'm concerned about the case where there is a command encoder, and it repeatedly resets a part of the buffer, then fills it, then resets it, then fills it, etc. This is the case that doesn't fall into "just expose what you already have", because the "uninitialized" handling is one way: command encoding can only reduce the amount of uninitialized data, not increase it. From this perspective, I think, putting it on the device timeline makes sense. When we react to uninitialized state on the queue submit, we are basically checking the device timeline state of initialization. That's why resetting it on the device timeline makes it trivial.

What kind of purpose did you envision for this, again?

Wumpf:
> where there is a command encoder, and it repeatedly resets a part of the buffer, then fills it, then resets it, then fills it, etc.

ah yeah that is pretty much what I want to do though :). I really would like to be able to flush out buffers at any time just like one clears rendertargets

kvark:
Ok, it sounds like we are talking about 2 different things:
  1. exposing the mechanism of zero-initialization. This would be on `device` timeline, can apply trivially to both buffers and textures
  2. clearing buffers

I would be happy with both of these, but they probably need to be different features.

Wumpf:
"simple" clear to zeros now

wgpu-types/src/lib.rs
```diff
@@ -2992,6 +3000,30 @@ pub struct ImageCopyTexture<T> {
...
+/// Subresource range within an image, used for clear_image
+#[repr(C)]
+#[derive(Clone, Debug)]
+#[cfg_attr(feature = "trace", derive(serde::Serialize))]
+#[cfg_attr(feature = "replay", derive(serde::Deserialize))]
+pub struct ImageSubresourceRange {
+    /// Included aspects: color/depth/stencil
+    pub aspects: TextureAspect,
+    /// First mipmap level in this subresource
+    pub level_start: u8,
```
kvark, line 3012: Let's follow the naming/comments of `struct TextureViewDescriptor` here. We should go even further and use this structure as a member of `TextureViewDescriptor`.

kvark (commented):
Great! Just a couple of notes

wgpu-core/src/command/clear.rs
```diff
@@ -0,0 +1,295 @@
...
+        let &(ref dst_raw, _) = dst_buffer
+            .raw
+            .as_ref()
+            .ok_or(ClearError::InvalidBuffer(dst))?;
+        if !dst_buffer.usage.contains(BufferUsage::COPY_DST) {
+            return Err(ClearError::MissingCopyDstUsageFlag(Some(dst), None));
+        }
+
+        // Check if offset & size are valid.
+        if offset % 4 != 0 {
```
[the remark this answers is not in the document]
Wumpf: copy it is then because that's also an required usage :)

Wumpf:
good suggestions. All done!

kvark (approved):

Issue #1350, opened 2021-04-22
Shader expects at least 8 descriptors for binding 1.0 but only 1 provided


kvark:
Happens on `noclip.website` ran on Gecko with WebGPU support.

**Description**
Errors:
> Shader expects at least 8 descriptors for binding 1.0 but only 1 provided 
> Vertex shader consumes input at location 1 but not provided

**Repro steps**
[wgpu-trace.zip](https://github.com/gfx-rs/wgpu/files/6360921/wgpu-trace.zip)

**Expected vs observed behavior**
No validation errors are expected

**Extra materials**
```
    VALIDATION [VUID-VkGraphicsPipelineCreateInfo-layout-00756 (0x45717876)] : Validation Error: [ VUID-VkGraphicsPipelineCreateInfo-layout-00756 ] Object 0: handle = 0x55e659ab73b8, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x45717876 | Shader expects at least 8 descriptors for binding 1.0 but only 1 provided The Vulkan spec states: layout must be consistent with all shaders specified in pStages (https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-VkGraphicsPipelineCreateInfo-layout-00756)
    object info: (type: DEVICE, hndl: 0x55e659ab73b8)
    
[2021-04-22T20:01:57Z ERROR gfx_backend_vulkan] 
    VALIDATION [VUID-VkGraphicsPipelineCreateInfo-layout-00756 (0x45717876)] : Validation Error: [ VUID-VkGraphicsPipelineCreateInfo-layout-00756 ] Object 0: handle = 0x55e659ab73b8, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x45717876 | Shader expects at least 8 descriptors for binding 1.1 but only 1 provided The Vulkan spec states: layout must be consistent with all shaders specified in pStages (https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-VkGraphicsPipelineCreateInfo-layout-00756)
    object info: (type: DEVICE, hndl: 0x55e659ab73b8)
    
[2021-04-22T20:01:57Z ERROR gfx_backend_vulkan] 
[...]

PR #1351, opened 2021-04-22
Hard error on SPIRV parsing failure when VALIDATION is requested

kvark:
**Connections**
Follow-up from https://github.com/gfx-rs/wgpu/issues/1350#issuecomment-825186159

**Description**
If the user provided SPIR-V, and they requested VALIDATION, and we fail to parse the file - there is no point to proceed.
VALIDATION flag should be a gaurantee of safe operation, so we can't just throw the same SPIR-V at SPIRV-Cross and pretend that it's ok.

**Testing**
Tested on noclip shaders

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -1005,10 +1005,11 @@ impl<B: GfxBackend> Device<B> {
                 let module = match parser.parse() {
                     Ok(module) => Some(module),
                     Err(err) => {
-                        // TODO: eventually, when Naga gets support for all features,
-                        // we want to convert these to a hard error,
                         log::warn!("Failed to parse shader SPIR-V code: {:?}", err);
-                        log::warn!("Shader module will not be validated or reflected");
+                        if desc.flags.contains(wgt::ShaderFlags::VALIDATION) {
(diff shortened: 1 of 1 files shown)


cwfitzgerald (approved):
LGTM, reasoning is sound, go on your call

kvark:
bors r=cwfitzgerald
with this in, it makes no sense to keep shipping SPIRV-Cross in Gecko on macOS, we'll drop it with the next wgpu update

cwfitzgerald:
Wait but what if parsing fails and we need to fall back on spirv cross? I'm particularly worried about this cause I think naga still can't parse rend3's crazier shaders.

kvark:
> Wait but what if parsing fails and we need to fall back on spirv cross? I'm particularly worried about this cause I think naga still can't parse rend3's crazier shaders.

The thing is, we can't run anything in Gecko that is not validated. Since without validation there is no safety. So if we get SPIR-V, and we fail to parse it, we can't trust it to go anywhere.

cwfitzgerald:
Ohhhhhhh, I missed the "in Gecko" part. Yes gecko should absolutely drop it.

PR #1352, opened 2021-04-26
Fake present in the player

kvark:
**Connections**
RenderDoc integration turns out to be hard - https://github.com/ebkalderon/renderdoc-rs/issues/113
Talking to RenderDoc author, there are things that can go wrong easily, like the RD API loading a different RD than the one currently registered with the layers...

**Description**
This PR sidesteps the capturing problem of present-less workloads, including the Gecko API traces and compute-only apps.
If we see there was no presentation done, we'll create a new swapchain and present it.

**Testing**
Tested on macOS briefly, still need to see if it actually helps RD.

### DIFF
--- a/player/src/bin/play.rs
+++ b/player/src/bin/play.rs
@@ -120,6 +120,7 @@ fn main() {
 
         let mut resize_desc = None;
         let mut frame_count = 0;
+        let mut has_swap_chain = false;
         let mut done = false;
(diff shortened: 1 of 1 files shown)


grovesNL (approved):

kvark:
Sadly, this doesn't appear to be working. RenderDoc doesn't consider this presentation to be a signal.

PR #1358, opened 2021-04-29
Big refactor of the profiling usage

kvark:
**Connections**

**Description**
Improves the profiling markers in many ways:
  1. remove unnecessary leftovers from the `tracing` era, such as querying device/adapter features. These are one-time and light-weight. It's unlikely we are going to be interested in them.
  2. add markers to where it really matters, such as `maintain` pieces, `submit` phases, render pass preparation, etc.
  3. shorten the markers to be strictly the function names, passing the class name as a tag

**Testing**
Tested on vange-rs


kvark:
Screen Shot 2021-04-29 at 02 07 30

PR #1393, opened 2021-05-17
Fix dynamic stencil reference for Replace ops

kvark:
**Connections**
Fixes an issue reported on Matrix, where we'd not set the stencil reference.

**Description**
Our conditions didn't check the Replace ops.

**Testing**
Untested, YOLO


cwfitzgerald (approved):
Something something needs a regression test

bors r+

PR #1394, opened 2021-05-17
Remove erroneous depth+stencil validation

ashpil:
**Connections**
Reported on Matrix by me.

**Description**
There was a validation check to make sure no depth+stencil textures have both a depth and stencil aspect. In reality it wasn't necessary for reasons I don't quite understand - it's gone now. 

**Testing**
Compiles. Should work.

### DIFF
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -152,8 +152,6 @@ pub enum CreateBindGroupError {
         layout_flt: bool,
         sampler_flt: bool,
     },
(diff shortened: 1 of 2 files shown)


kvark (approved):
Thank you!

kvark (changes requested):
Ok, now I remember

kvark:
I got confused for a moment thinking that this is about view creation.
But this is bind group creation, and we can't create bind group entries for both depth and stencil.

PR #1403, opened 2021-05-20
Trace render passes before they are ran

kvark:
**Connections**
Follow-up to #1245 

**Description**
We want the API traces to reproduce issues. Previously, we moved all the resource creation to happen after the trace is written.
This PR follow-up by also recording the render pass commands even in the case where one of them triggers a validation error.

Another big part of the PR now is handling the command buffers that failed for some reason. The submission is skipping them now instead of erroring the whole thing.

**Testing**
Tested manually on wgpu-rs examples


cwfitzgerald (approved):
bors r+

Issue #1413, opened 2021-05-28
Resource tracking overhead


kvark:
Related to #1411
Firefox profile - https://share.firefox.dev/3fvytEz
If you look at `run_render_pass_impl` in the flamegraph view, the Vulkan driver is taking less than half of that time.
The rest is spent on barriers and tracking (everything related to `TrackerSet`), so clearly this is a hot spot.

Ideas:
  1. Most buffers and textures are immutable after they got the initial contents transferred into them. It would be wonderful to find a way to avoid tracking them entirely, or at least up to the point where they do get mutated.
  2. Today the tracker has `(init, current)` states per subresource. We could extend it to `(init, current, next)`, so that the sync scope (of, say, a render pass) will be accumulated in the `next`. This would mean - no need to allocate/free tracker sets per pass. Everything would be done right in the command buffer tracker. Related to #443

kvark:
Assembly for the following functions need to be looked at under a microscope:
  - `gfx_backend_vulkan::command::CommandBuffer::bind_descriptor_sets`: roughly 1/9 of it is actually spent in `ash::vk::features::DeviceFnV1_0::cmd_bind_descriptor_sets`. The rest includes boilerplate an inlined pipeline compatibility logic.
  - `wgpu_core::command::CommandBuffer::insert_barriers` seems quite heavy, also does some refcount stuff (unexpectedly?)
  - `wgpu_core::command::bind::Binder::assign_group` does refcounting. I don't think it should?
  - `wgpu_core::track::TrackerSet::merge_extend` and `wgpu_core::track::ResourceTracker::use_extend`: maybe we can make them cache friendlier?
  - `wgpu_core::hub::Storage::get` shows up a bit everywhere, but a lot in `submit()`

kvark:
Another idea (3) - a merge of 1) and 2) but without any heavy changes. If we know something doesn't have a state (but just needs to be added to the lifetime tracker), then adding it to the render pass tracker is a waste. We could make it so only buffers and textures are tracked per pass (or per usage scope), and everything else goes straight to the command buffer tracker. For the animometer benchmark, it would cut the costs by almost the factor of 2.

kvark:
New profile after (3) is merged - https://share.firefox.dev/3pjq2Qa
There are still things to address here. Also there is a gap not annotated by anything.

cwfitzgerald:
Closing after #2662

Issue #1397, opened 2021-05-18
Incremental presentation


kvark:
**Is your feature request related to a problem? Please describe.**
Presenting full screen is wasteful if the user knows only a portion of the screen is changed.

**Describe the solution you'd like**
A native-only feature that allows the user to specify a rectangle to present.
This isn't a security concern, but a portability one.

**Describe alternatives you've considered**
No incremental present

**Additional context**
It's not clear if this is possible, given that the user of a swapchain doesn't know about the number of textures there. So there is no good way for the user to know the baseline, from which the changes were made.

cwfitzgerald:
I highly suspect we will have to expose to the user the amount of images in the swapchain. It also implies letting people render to the swapchain without a clear, something that is currently disallowed.

cwfitzgerald:
Closing in favor of #2869, this isn't as widespread as it seems.

PR #1417, opened 2021-05-31
Split the tracker into stateful/stateless to reduce the overhead

kvark:
**Connections**
Implements https://github.com/gfx-rs/wgpu/issues/1413#issuecomment-851604344
Reduces the overhead for resource tracking in the Animometer benchmark by up to 50%.

**Description**
We used to use the full tracker set on the usage scopes associated with compute/render passes. A resource tracker has 2 responsibilities: ensuring the resource is held alive, and validating and recording the state transitions. This PR exploits the fact that the latter responsibility is only applicable for buffers and textures. So doing all the lifetime tracking for a pass is a waste: we can instead just attach the lifetimes to the parent command buffer, straight.

In the Animometer benchmark, there is one large buffer, and thousands of bind groups pointing to different offsets into it. The old code would fill up the pass tracker with those bind groups, and then merge it into the command buffer tracker. The new code would just fill up the command buffer tracker instead. Since there is only one buffer, the pass tracking becomes much lighter.

**Testing**
Untested. It would be nice to have some benchmarks here, possibly after #1397 ?

### DIFF
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -214,7 +214,7 @@ impl RenderBundleEncoder {
                     state.set_bind_group(index, bind_group_id, bind_group.layout_id, offsets);
                     state
                         .trackers
-                        .merge_extend(&bind_group.used)
+                        .merge_extend_all(&bind_group.used)
                         .map_pass_err(scope)?;
                 }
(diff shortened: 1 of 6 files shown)


cwfitzgerald (approved):
bors r+

PR #1419, opened 2021-06-01
Skip the cube compatibility on a downlevel flag

### LINKED ISSUE #1418: texture array interpreted as cube texture array
Short info header:
It seems that repeated growing of a texture 2d array can be reinterpreted as a cube texture array, which can cause failures on systems that don't support cube texture arrays (e.g., MTFGPUFamilyApple3).
For example, repeated uploads of https://github.com/hecrj/iced/blob/e292821c3780fc5a57c835815efc7a232ce4dc32/wgpu/src/shader/image.wgsl#L8 can exhibit this behavior.
- GFX version: wgpu-rs 0.8
- OS: iOS/macOS
- GPU: a10x

kvark:
**Connections**
Fixes #1418

**Description**
Adds a downlevel flag for cube arrays, checks for it.

**Testing**
Untested, but should work.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -681,7 +681,14 @@ impl<B: GfxBackend> Device<B> {
         let mut view_caps = hal::image::ViewCapabilities::empty();
         // 2D textures with array layer counts that are multiples of 6 could be cubemaps
         // Following gpuweb/gpuweb#68 always add the hint in that case
-        if desc.dimension == TextureDimension::D2 && desc.size.depth_or_array_layers % 6 == 0 {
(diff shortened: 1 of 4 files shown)


cwfitzgerald (changes requested):
We should keep the downlevel flag. I know that the webgpu standard does not say that ANISO has to do anything, but people should still know when it's going to be ignored.

kvark:
Ok, the compatibility mask was wrong then, it took aniso into it. I'll add the flag back, but outside of the compat mask.

kvark:
The review note is addressed, let's proceed.
bors r=cwfitzgerald

PR #1469, opened 2021-06-03
Handle winit exception in web to avoid breaking async executor

Frizi:
This is a continuation of https://github.com/gfx-rs/wgpu-rs/pull/923.

I'd still like to fix this on our end, even if the real underlying issue is in winit. The fix is small, and we can migrate to the proper winit solution once it exists. In the mean time, we can continue working on improving webgl support without a major blocker.


grovesNL (approved):
Thank you! Let's proceed with this for now and later plan to remove it once we can fix it upstream (https://github.com/rust-windowing/winit/issues/1956)

bors r+

PR #1474, opened 2021-06-04
Add clear buffer/image to wgpu-rs api

Wumpf:
Follow up to  #1335

Started using this in Blub fluid sim a little bit on volume textures, works nicely https://github.com/Wumpf/blub/commit/432fb75a6865565d29503432f79c60f29ec4c637, more to come

### DIFF
--- a/wgpu/src/backend/web.rs
+++ b/wgpu/src/backend/web.rs
@@ -1899,6 +1899,23 @@ impl crate::Context for Context {
         })
     }
 
+    fn command_encoder_clear_image(
+        &self,
+        _encoder: &Self::CommandEncoderId,
+        _texture: &crate::Texture,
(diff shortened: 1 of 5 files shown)


wgpu/src/backend/web.rs
```diff
@@ -1920,7 +1937,7 @@ impl crate::Context for Context {
         _query_set: &Self::QuerySetId,
         _query_index: u32,
     ) {
-        // Not available in gecko yet
+        unimplemented!();
```
Gordon-F, line 1940: How about `log::error` instead of panic ?

kvark (approved):
Thank you!
bors r+

kvark:
bors retry
I wonder what those failures are about

PR #1502, opened 2021-06-18
Refactor halmark example to properly recycle command pools

kvark:
**Connections**

**Description**
It didn't reset the command encoder until the end, which made it non-suitable for Vulkan benchmarking.

**Testing**
manually


kvark:
With this, memory stays stable on Vulkan (for halmark, not bunnymark), and I'm seeing roughly 20ms frames drawing 190K sprites per frame on Linux/Nvidia.

PR #1505, opened 2021-06-18
Change the order to suspect temporary resources earlier

### LINKED ISSUE #1504: Texture view epochs mismatch
**Description**
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `2`,
 right: `1`: Epoch mismatch for Valid((3, 1, Metal))', /Users/dmalyshau/Code/wgpu/wgpu-core/src/track/mod.rs:253:21
```

**Repro steps**
Closed source, unfortunately

No panics are expected in the tracking code

**Extra materials**
@scoopr project

**Platform**
any

kvark:
**Connections**
Fixes #1504

**Description**
I don't fully understand why this works.

**Testing**
tested on the privately given case, also our examples

### DIFF
--- a/wgpu-core/src/device/life.rs
+++ b/wgpu-core/src/device/life.rs
@@ -223,7 +223,6 @@ impl<A: hal::Api> LifetimeTracker<A> {
     pub fn track_submission(
         &mut self,
         index: SubmissionIndex,
-        new_suspects: &SuspectedResources,
         temp_resources: impl Iterator<Item = TempResource<A>>,
         encoders: Vec<EncoderInFlight<A>>,
     ) {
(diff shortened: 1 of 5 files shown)


kvark:
I thought about this a bit, and concluded that this is the right fix. Here is what's going on:
  1. Resource A is created and used in a command buffer
  2. A is dropped, but the command buffer B keeps it alive
  3. command buffer B is submitted. A is added to the list of resources suspected for removal from the Hub.
  4. this is where the old code diverge: B is dropped, suspected resources get analyzed, and then A is added to the list. During the swipe, we see that A has no outside references, and we remove it.
  5. New resource is created at the same Hub slot as A
  6. The next swipe of suspected resources sees the old reference to A with the epoch behind the current epoch

So basically, if we see a command buffer carrying the last reference to something, it should be analyzed right away. We can't postpone this to the next swipe.

bors r+

PR #1518, opened 2021-06-19
Make spirv an optional feature

kvark:
**Connections**
https://github.com/gfx-rs/naga/issues/940 shows how much SPIR-V parsing can be a pain.

**Description**
Keep it supported natively, but put it behind a feature flag. This allows to skip compilation of parts of Naga as well as dependencies like `petgraph`.
On my machine, compiling `wgpu-core` time is reduced from 40.87s to 35.36s, which is about 13% improvement.

**Testing**
Just compiling

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -838,28 +838,6 @@ impl<A: HalApi> Device<A> {
         source: pipeline::ShaderModuleSource<'a>,
     ) -> Result<pipeline::ShaderModule<A>, pipeline::CreateShaderModuleError> {
         let module = match source {
-            #[cfg(feature = "spirv")]
-            pipeline::ShaderModuleSource::SpirV(spv) => {
-                profiling::scope!("naga::spv::parse");
-                // Parse the given shader code and store its representation.
-                let options = naga::front::spv::Options {
-                    adjust_coordinate_space: false, // we require NDC_Y_UP feature
-                    strict_capabilities: true,
-                    flow_graph_dump_prefix: None,
-                };
-                let parser = naga::front::spv::Parser::new(spv.iter().cloned(), &options);
-                match parser.parse() {
-                    Ok(module) => module,
-                    Err(err) => {
-                        log::warn!(
-                            "Failed to parse shader SPIR-V code for {:?}: {:?}",
-                            desc.label,
-                            err
-                        );
-                        return Err(pipeline::CreateShaderModuleError::Parsing);
-                    }
-                }
-            }
             pipeline::ShaderModuleSource::Wgsl(code) => {
                 profiling::scope!("naga::wgsl::parse_str");
                 // TODO: refactor the corresponding Naga error to be owned, and then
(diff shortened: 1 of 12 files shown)


wgpu/README.md
```diff
@@ -43,6 +43,12 @@ The following environment variables can be used to configure how the framework e
 
 See [wiki article](https://github.com/gfx-rs/wgpu-rs/wiki/Running-on-the-Web-with-WebGPU-and-WebGL).
 
+## Shaders
```
Gordon-F, line 46: Thank you! Compile time improvements are always good. Should we add a note about `wgsl-out` here? So, people can convert their `GLSL` or `SPV` shaders to `WGSL` at compile time, or in runtime (maybe we need an example for this). Or `naga` can help translate their shaders and they can finish shaders up by hand.

kvark: love the idea! doing this now

ElectronicRU:
Regarding #1520 -- should that still be enabled without spirv? I'd say yes because this feature is related to *parsing* priv, not *using* spirv, but the name could get confusing.

kvark:
@ElectronicRU Yes, I was just thinking about the same. Perhaps, we should adjust the (run-time) feature name to spell it more clearly.
Something like: `SPIRV_SHADER_PASSTHROUGH`? If it's just `SPIRV_SHADERS` then it's easy to confuse with the effect of the compile-time feature in this PR.

cwfitzgerald (approved):
I think we should go ahead with this, the compile time improvement is nothing to shake a stick at

grovesNL:
Interestingly if we did allow naga modules to be provided as mentioned on Matrix recently, we wouldn't need SPIR-V specific options here. i.e. the application/library could opt-in to the SPIR-V frontend in naga

cwfitzgerald:
It does predicate that naga can actually parse the spirv, which is one of the things passing raw spirv is trying to get around in #1525

kvark:
@grovesNL indeed. It would have to be passed by move though, because wgpu wants to own Naga's module, but these modules aren't cloneable (at least today).

kvark:
@grovesNL @cwfitzgerald I added a commit that removes SPIR-V support entirely from `wgc`. The support on `wgpu-rs` side is unchanged.

wgpu/README.md
```diff
@@ -43,6 +43,18 @@ The following environment variables can be used to configure how the framework e
...
+
+[WGSL](https://gpuweb.github.io/gpuweb/wgsl/) is the main shading language of WebGPU.
+
+Users can run the [naga](https://github.com/gfx-rs/naga) binary in the following way to convert their SPIR-V shaders to WGSL:
+```bash
+cargo run -- <input.spv> <output.wgsl>
+
```
cwfitzgerald, line 53: s p a c e

cwfitzgerald:
I'd be _really_ hesitant to merge removing spirv entirely _right now_. I do think this is eventually where we want to be, but having things be predicated on naga understanding every shader means that a good chunk of our features are unusable because naga is yet unaware of them. spirv was basically our last way around any current limitations of naga. I think in some time we should reconsider.

wgpu-core/src/device/mod.rs
```diff
@@ -3484,9 +3464,11 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                     pipeline::ShaderModuleSource::Wgsl(ref code) => {
                         trace.make_binary("wgsl", code.as_bytes())
                     }
-                    pipeline::ShaderModuleSource::Naga(_) => {
-                        // we don't want to enable Naga serialization just for this alone
-                        trace.make_binary("ron", &[])
+                    pipeline::ShaderModuleSource::Naga(ref module) => {
```
cwfitzgerald, line 3467: Missed a SpirV above this branch

kvark:
> spirv was basically our last way around any current limitations of naga. I think in some time we should reconsider.

This functionality is here, it stays, and it's no directly related to this PR. See #1525

Also, the current state of the PR is still keeping SPIRV parsing support, just at wgpu-rs level instead of wgpu-core level.

cwfitzgerald:
Oh I see, so we're just removing spirv ingestion from wgpu-core in the safe path. That sounds reasonable to me. I've got some nits, but otherwise LGTM.

cwfitzgerald (approved):

wgpu-core/src/device/mod.rs
kvark: There isn't any SpirV in there any more. Could you clarify, please?

cwfitzgerald: Apparently not, nvm then :)

kvark:
bors r=cwfitzgerald

PR #1524, opened 2021-06-19
Remove MPL 2.0 header in files

Gordon-F:
This is no more required since the repository is licensed under `MIT` and `Apache 2.0`.


kvark (approved):
I still need to figure out how to go about this in mozilla-central
cc @jimblandy
bors r+

PR #1533, opened 2021-06-21
web: add empty `device_create_shader_module_spirv`

grovesNL:
**Description**
We added a SPIR-V passthrough function to the `Context` trait shared between web/native backends, but the web backend also needs an implementation of this function to fully implement the trait.

**Testing**
Compiling the web backend for WebGPU.


kvark (approved):
Thank you!
bors r+

PR #1534, opened 2021-06-21
web: Restore WebGPU CI

grovesNL:
**Connections**
Depends on/currently includes #1533

**Description**
Even though the WebGL backend needs to be temporarily disabled, we should be able to continue checking WebGPU on wasm.

**Testing**
Manually ran the CI command `cargo check --package wgpu --examples --target=wasm32-unknown-unknown` locally which seems to catch the mistake in #1533 without building wgpu-core/wgpu-hal.


kvark (approved):
bors r+

Issue #1532, opened 2021-06-20
Texture Array Examples Shaders Don't Parse


cwfitzgerald:
**Description**

Texture array examples have their spirv rejected by naga.

**Repro steps**

Run 0.9's texture array examples, they all fail in some way.

<details>
<summary>Errors</summary>

```
test texture_arrays_constant ... [2021-06-20T05:48:41Z WARN  wgpu_core::instance] The underlying API or device in use does not support enough features to be a fully compliant implementation of WebGPU. A subset of the features can still be used. If you are running this program on native and not in a browser and wish to limit the features you use to the supported subset, call Adapter::downlevel_properties or Device::downlevel_properties to get a listing of the features the current platform supports.
[2021-06-20T05:48:41Z ERROR wgpu_core::validation] Unexpected varying type: Array { base: [1], size: Constant([5]), stride: 4 }
[2021-06-20T05:48:41Z ERROR wgpu_core::validation] Unexpected varying type: Array { base: [1], size: Constant([5]), stride: 4 }
[2021-06-20T05:48:41Z WARN  wgpu_core::device] Failed to parse shader SPIR-V code for Some("constant.frag.spv"): InvalidGlobalVar(Access { base: [2], index: [12] })        
[2021-06-20T05:48:41Z ERROR wgpu::backend::direct] wgpu error: Validation Error

    Caused by:
        In Device::create_shader_module
          note: label = `constant.frag.spv`
        Failed to parse WGSL


thread 'main' panicked at 'Handling wgpu errors as fatal by default', wgpu\src\backend\direct.rs:1978:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'UNEXPECTED TEST FAILURE', wgpu\examples/texture-arrays\..\..\tests\common\mod.rs:255:13
FAILED
test texture_arrays_non_uniform ... [2021-06-20T05:48:41Z WARN  wgpu_core::instance] The underlying API or device in use does not support enough features to be a fully compliant implementation of WebGPU. A subset of the features can still be used. If you are running this program on native and not in a browser and wish to limit the features you use to the supported subset, call Adapter::downlevel_properties or Device::downlevel_properties to get a listing of the features the current platform supports.
[2021-06-20T05:48:41Z ERROR wgpu_core::validation] Unexpected varying type: Array { base: [1], size: Constant([5]), stride: 4 }
[...]

PR #1538, opened 2021-06-22
Implement Integration and Example Tests

### LINKED ISSUE #1379: [meta] Plan for Writing Convenient and Distributable Tests
## The Problems

1. We've witnessed a vast increase in bug reports as a result of the 0.8 release. While most of these were patched quickly due to @kvark and other's effort and diligence, it indicates that our current testing process is not sufficient to catch all the bugs we have.

2. In addition, the backends that are less manually tested tend to have a large number of uncaught bugs. I've heard from many people that they can only run their project on Vulkan or maybe Vulkan/Metal but no others. This also makes it painfully clear that we need a way to reliably test these other backends.

3. Our current wgpu-level tests are extremely incomplete. We primarily use the wgpu-rs examples as a testing platform, so they are polluted with many extraneous features, and running them all on all backends is an extremely tedious job. An additional step in our release testing process involves porting forward many applications and testing them for bugs on the platforms that we have available. This is a very spot-check-based process and has many holes. It is also dependant on the maintainers of the projects to have time to do the porting process.

4. Writing play-tests are a very tedious process because there are many different things that must all align (such as buffer IDs) and generally can't utilize things like helper functions which can simplify the writing of lots of simple tests easily.

5. There are numerous platform-specific issues that are caused by either driver bugs or wgpu bugs. We have no way of catching these as we only regularly test a small set of platforms and APIs.

## The Solutions

There are quite a large set of problems that we need to solve. Thankfully there is an incremental path to a solution.

### Create Easy-To-Write Easy-To-Modify Testing Infrastructure

While play-tests have their use and I do not recommend getting rid of them, their inconvenience to write makes it difficult to find the time and energy to write new play-tests. As such I propose a new system of testing that programs directly against wgpu-core. While using wgpu-core directly can be a pain, a lot of the operations the tests are doing are repetitive and can easily abstract into helper functions. Once a collection of these functions are built up, adding new tests becomes extremely easy. I will mock up some simple tests in the coming weeks.

Tests can be filtered based on the presence of features/limits/downlevel capabilities. If a test is filtered it will "pass" but print to stdout why it was rejected.

Once we have the infrastructure in place, we should recruit help in writing the tests detailed in #807.

### Run Tests on All Backends and GPUs

With the above in place, it should be easy to test all different backends and cards on the system. The ideal testing flow should look something like this.

```
WGPU_BACKEND=vulkan cargo test
WGPU_BACKEND=dx12 cargo test
..
WGPU_BACKEND=vulkan WGPU_ADAPTER='RX 580' cargo test
WGPU_BACKEND=dx12 WGPU_ADAPTER='RX 580' cargo test
```

This likely could be automated with a simple helper program which would also be necessary for the final step.

### Run Tests in CI

The work set out by #1365 should allow us to test vulkan, dx12, dx11, and opengl in CI. We need to make sure this happens using the above infrastructure to test all the things.

### Split Play Tests into Individual Tests

This might be a bit weird, but if we are trying to programmatically extract data from tests, it would be significantly easier if each playtest was its own rust test. We can make each test a call to the play function with a path to the trace. This would be no more difficult than the current solution with listing everything in a ron file.

### Build Distributed Testing Network

Once we have a reliable and functional test suite, we should build out a testing network using hardware volunteered by their owners, and given to contributors by the [gfx-rs hardware fund](https://opencollective.com/gfx-rs). This involves developing a server to manage and view all the tests and the data and clients who run the tests and report the results.

The details of this are out of scope for this issue but will be expanded on when it is time to build these systems.

### LINKED ISSUE #1465: Let Examples Choose GPU
We already have ways to choose backend, for devs like me with many gpus in one system, it would be helpful to have a way in the framework to override the gpu based on env. Obviously `enumerate_adapters` isn't available on wasm, so this would have to be ignored on WASM.

cwfitzgerald:
**Connections**

Closes #1379! Closes #1465.

**Description**

This adds a testing framework for wgpu. This includes testing infrastructure for integration and example based reftests and other tests.

I have added env-based adapter and backend choices as a common utility in wgpu-rs. This is how the framework based examples choose their adapters, so WGPU_ADAPTER_NAME is how you can choose adapters based on a substring of the name.

I added explicit seeding to all examples that use random numbers so reftesting was deterministic. conservative-raster, mipmap, and texture-arrays (#1532) are broken and are marked as such. This testing framework will make CI fails if this is fixed, so tests don't fall out of date.

Note: as we get more integration tests, we can factor out more of the `pulling_common` to be common between all the tests. 

**Testing**

It is testing :)


wgpu-core/src/lib.rs
```diff
@@ -43,9 +43,6 @@ mod validation;
 
 pub use hal::api;
 
-#[cfg(test)]
-use loom::sync::atomic;
-#[cfg(not(test))]
 use std::sync::atomic;
```
kvark, line 46: let's merge this into the other `use std::`

wgpu/Cargo.toml
```diff
@@ -76,11 +82,61 @@ git = "https://github.com/gfx-rs/naga"
...
+test = true
+
+[[example]]
+name="texture-arrays"
+path="examples/texture-arrays/main.rs"
+test = true
+
+[[example]]
+name="water"
+path="examples/water/main.rs"
```
kvark, line 137: are all these paths necessary? I'd expect these to match the default search paths

wgpu/examples/bunnymark/main.rs
```diff
@@ -349,3 +349,19 @@ impl framework::Example for Example {
 fn main() {
     framework::run::<Example>("bunnymark");
 }
+
+#[test]
+fn bunnymark() {
+    framework::test::<Example>(
+        concat!(
+            env!("CARGO_MANIFEST_DIR"),
+            "/examples/bunnymark/screenshot.png"
```
kvark, line 358: would it make sense to pass only this part? i.e. can the `fn test()` implementation look up the `CARGO_MANIFEST_DIR` and do the path formation?

wgpu/examples/conservative-raster/main.rs
```diff
@@ -285,7 +285,7 @@ impl framework::Example for Example {
             let mut rpass = encoder.begin_render_pass(&wgpu::RenderPassDescriptor {
                 label: Some("full resolution"),
                 color_attachments: &[wgpu::RenderPassColorAttachment {
-                    view: &frame.view,
+                    view: &view,
```
kvark, line 288: shouldn't this be just `view,`? here and in all other places

wgpu/tests/example_wgsl.rs
```diff
@@ -1,6 +1,7 @@
 use naga::{front::wgsl, valid::Validator};
```
kvark, line 1: why are we renaming this example? is there anything wrong with a dash there?

wgpu/tests/vertex_indexes/draw.vert.wgsl
```diff
@@ -0,0 +1,19 @@
+[[block]]
```
kvark, line 1: nit: I've seen and used another spelling "vertex_indices"

wgpu/examples/wgpu-info/main.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    println!("Adapter {}:", idx);
+    println!("\tBackend:   {:?}", info.backend);
+    println!("\tName:      {:?}", info.name);
+    println!("\tVendorID:  {:?}", info.vendor);
+    println!("\tDeviceID:  {:?}", info.device);
+    println!("\tType:      {:?}", info.device_type);
+    println!("\tCompliant: {:?}", downlevel.is_webgpu_compliant());
+    println!("\tFeatures:");
+    println!("\t\tDepth Clamping:                             {}", features.contains(wgpu::Features::DEPTH_CLAMPING));
```
kvark, line 22: is it really not good to just debug-print the features? it would say something like `DEBUG_CLAMPING | TIMESTAMP_QUERY`.
I guess one thing that it gives you is a quick lookup for a question like "is something supported?"

wgpu/src/util/init.rs
```diff
@@ -0,0 +1,91 @@
...
+            .map(str::to_lowercase)
+            .as_deref()
+        {
+            Ok("vulkan") => BackendBit::VULKAN,
+            Ok("dx12") => BackendBit::DX12,
+            Ok("dx11") => BackendBit::DX11,
+            Ok("metal") => BackendBit::METAL,
+            Ok("gl") => BackendBit::GL,
+            Ok("webgpu") => BackendBit::BROWSER_WEBGPU,
+            _ => None?,
```
kvark, line 19: is clippy not yelling at you for this sin? it should :)

wgpu/examples/framework.rs
```diff
@@ -1,11 +1,14 @@
-use std::future::Future;
 #[cfg(not(target_arch = "wasm32"))]
 use std::time::{Duration, Instant};
+use std::{future::Future};
 use winit::{
     event::{self, WindowEvent},
     event_loop::{ControlFlow, EventLoop},
 };
 
+#[path = "../tests/common/mod.rs"]
```
kvark, line 9: can we add `#[cfg(test)]` on top?

wgpu/examples/bunnymark/main.rs
```diff
@@ -349,3 +349,19 @@ impl framework::Example for Example {
...
+        ),
+        1024,
+        768,
+        wgpu::Features::default(),
+        framework::test_common::TestParameters::default(),
+        1,
```
kvark, line 364: It's a bit counter-intuitive to provide a bunch of numbers in the parameters. We can live with this, but if you want to consider having another struct with these numbers in the fields, that would be a nice improvement.

kvark (commented):
This looks quite fancy and very useful. Thank you for taking care of our testing story!
I don't think any of the comments below are blockers.

README.md
```diff
@@ -32,3 +32,31 @@ If you are looking for the native implementation or bindings to the API in other
   GLes3  |                    |                    |                    |
 
 :white_check_mark: = Primary support — :ok: = Secondary support — :construction: = Unsupported, but support in progress
+
+## Testing Infrastructure
+
+wgpu features a set of unit, integration, and example based tests. All framework based examples are automatically reftested against the screenshot in the example directory. The `wgpu-info` example contains the logic which can automatically run the tests multiple times for all the adapters present on the system. These tests are also run on CI on windows and linux over Vulkan/DX12/DX11/GL on software adapters.
```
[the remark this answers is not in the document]
cwfitzgerald: I wanted the tool to have a similar name to vulkaninfo as it is a useful tool on it's own.

wgpu/Cargo.toml
cwfitzgerald: They do, but we have to manually enable tests on every single one of them...

wgpu/examples/bunnymark/main.rs
cwfitzgerald: Sure

wgpu/tests/example_wgsl.rs
cwfitzgerald: We're renaming this because it's now a submodule of the root integration test and dash is an invalid module name.

wgpu/examples/wgpu-info/main.rs
cwfitzgerald: The string on high-supported cards gets _really_ long and wraps 3-4 times, so it's very hard to figure out which are aren't. This also makes it easier to see which _aren't_ supported, which is often more relevant.

wgpu/examples/framework.rs
cwfitzgerald: Oh yes we should, that should also fix wasm

wgpu/examples/bunnymark/main.rs
cwfitzgerald: I agree, especially if we need more tuning in the future.

README.md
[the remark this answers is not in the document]
kvark: Right. But this one is not a standalone tool, it's an example. If we want to have a binary by this name in the repo, that would be totally cool. The old "hello" example actually serves this goal - prints out some info.

wgpu/Cargo.toml
kvark: Yes, I understand. So we can at least remove the `path = ` lines, it seems.

wgpu/examples/wgpu-info/main.rs
kvark: We can print this in a more automated way: we can iterate the bit, and for each bit print a separate line like you did here. Something like:
```rust
let count = wgt::Features::all().bits().trailing_ones();
for i in 0 .. count {
  let bit = wgt::Features::from_bits(1 << i);
  println!("\t\t{:?}: {}", bit, features.contains(bit));
}
```

Same for downlevel flags. Would save quite a bit of repetition here.
But yes, there is a trade-off. Your hand-written description is better.

Maybe we can have a hybrid approach?  We'd implement `Display` for the bitflags, putting your nice descriptions in there, but then just have a loop in here like I suggested, using that `Display` implementation?

cwfitzgerald: That seems reasonable, that also prevents this code from bitrotting as more features get added but no one remembers this exists. I think I may also explicitly destructure the limits structs so that adding new limits errors out here.

wgpu/src/util/init.rs
cwfitzgerald: But think about all 6 characters it saved 😆

wgpu-info/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+authors = ["wgpu developers"]
+edition = "2018"
+description = "Adapter information and per-adapter test program"
+homepage = "https://github.com/gfx-rs/wgpu"
+repository = "https://github.com/gfx-rs/wgpu"
+keywords = ["graphics"]
+license = "MIT OR Apache-2.0"
+
+[dependencies]
+wgpu = { version = "0.8", path = "../wgpu" }
```
kvark, line 13: we should probably bump the versions to 0.9, just to avoid confusion

kvark (approved):
Looks great!
Please rebase :)
Also, we have a list of crates in README, lets add wgpu-info there.

cwfitzgerald:
bors r=kvark

PR #1539, opened 2021-06-22
Fix spirv feature by removing profiling::scope bit

scoopr:
**Description**
Compilation failed with `spirv` featue enabled

**Testing**
Compiled with `spirv` feature enabled


scoopr:
I suppose CI is not enabling spirv in any of the jobs?

kvark:
sorry, we definitely need to enable `spirv` feature coverage on CI!
Would you be open to adding it? just pick any of the CI jobs in the list and add to the "core features" list to test

kvark (approved):

kvark:
Actually, this isn't a core feature...
bors r+

PR #1547, opened 2021-06-23
Don't merge stateless trackers from bind groups

kvark:
**Connections**
Hinted by @pythonesque on the matrix
Follow-up to #1417

**Description**
This is an optimization only.

When a bind group is used, we need to ensure two things:
  1. stateful usages are merged in (buffers and textures)
  2. other resources are help alive by the command buffer

We used to merge all of the states into the local trackers. However, the local tracker already
keeps a strong reference to the whole bind group, so the resources in it are not going away.
Therefore, we can skip having them separately in the local trackers.

**Testing**
Untested.


cwfitzgerald (approved):
LGTM!

bors r+

PR #1549, opened 2021-06-24
Fix cargo-tarpaulin

cwfitzgerald:
I have no way of actually testing this is tarpaulin is linux only, but this should work


kvark (approved):
bors r+

PR #1550, opened 2021-06-24
Update submission indices on all stateless resources

kvark:
**Connections**
This is partially a regression by #1547, but also fixing more things.

**Description**
We stopped tracking the stateless resources included in bind groups. But we still need to update their submission indices, so that they don't get removed too early.
The PR also does the same for render bundles, but with a twist: bind groups used in a bundle are merged early.

**Testing**
untested


kvark:
bors r=pythonesque

PR #1561, opened 2021-06-27
[v0.9] fix buffer inits being delayed by a frame

### LINKED ISSUE #1555: Buffer gets cleared after it has already been used
**Description**
A Buffer gets cleared in the subsequent frame after its creation even though it has already been used in the previous frame.

**Repro steps**
The provided example creates a buffer every other frame, fills it using a compute shader and then uses this buffer to draw a colored triangle (the buffer is the source for the color).

The triangle should always be colored since the buffer gets filled and then not touched anymore for 2 frames. After these 2 frames the buffer gets replaced with a new one.
However the triangle flickers in color and black. Running it in RenderDoc shows that a vkCmdFillBuffer commands gets emitted in the subsequent frame after the buffers creation and first use.

**Example**
[wgpu-buffer-clear-bug.zip](https://github.com/gfx-rs/wgpu/files/6718642/wgpu-buffer-clear-bug.zip)

**Platform**
Windows 10
wgpu-0.9 vulkan backend
AMD RX570 driver version 21.4.1

Wumpf:
**Connections**
Fixes #1555

**Description**
Fix buffer inits being inserted into the next submit call
Haven't checked how long this issue exists already, but this pretty much kills all user data that is generated by rendering/compute

**Testing**
Tested on provided repro project in ticket. Pretty bad bug so would like to see a regression test but can't come up right now with anything easier than the taking over the given example :/

### DIFF
--- a/wgpu-core/src/device/queue.rs
+++ b/wgpu-core/src/device/queue.rs
@@ -601,12 +601,11 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             let device = device_guard
                 .get_mut(queue_id)
                 .map_err(|_| DeviceError::Invalid)?;
-            let pending_write_command_buffer = device.pending_writes.finish();
             device.temp_suspected.clear();
             device.active_submission_index += 1;
             let submit_index = device.active_submission_index;
 
-            let fence = {
+            let (fence, pending_write_command_buffer) = {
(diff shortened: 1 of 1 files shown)


kvark (approved):
Nice fix, thank you!
bors r+

PR #1562, opened 2021-06-27
Fix check for QUERY_SET_MAX_QUERIES was using >= instead of >

### LINKED ISSUE #1560: QuerySet size limit off-by-one
**Description**
According to the documentation, the number of queries in a single QuerySet object is limited to at most 8192. However, the code will currently throw an error if the number of queries is *equal or higher* to 8192. So it is my understanding that having exactly 8192 queries should be allowed, but currently this throws an error.

Relevant code in wgpu:
image

**Repro steps**

```
device.create_query_set(&wgpu::QuerySetDescriptor {
     ty: wgpu::QueryType::Timestamp,
     count: 8192,
})
````

Expected: no error
Observed: `Validation error: 8192 is too many queries for a single QuerySet. QuerySets cannot be made more than 8192 queries`.

If everyone agrees that this is indeed a bug, I can make a PR.

HalfVoxel:
Fixes #1560

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## Unreleased
+  - Fixed:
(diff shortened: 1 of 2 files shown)


cwfitzgerald (approved):
Thanks!

bors r+

PR #1563, opened 2021-06-27
Fix stray header in changelog

HalfVoxel:
It's definitely released by now. I'm guessing it was part of 0.8.1.


cwfitzgerald:
Lmao good catch

bors r+

cwfitzgerald (approved):

PR #1568, opened 2021-06-28
fix conservative rasterization for vulkan

### LINKED ISSUE #1553: Conservative Raster Example Reftests Only Rasters Normally
**Description**

There are two different triangles that should be drawn but only the blue non-conservative one is rasterized. This causes reftest failure. This was known of before, just recording it for tracking purposes.

Wumpf:
**Connections**
Fixes #1553

**Testing**
image comparison test passes now


cwfitzgerald:
Thanks!

bors r+

kvark (approved):
Nice one!
bors r+

PR #1471, opened 2021-06-04
Hardware Abstraction Layer

kvark:
**Connections**
Closes #1423

**Description**
Prototyping the new hardware abstraction layer for wgpu, which replaces gfx-rs dependencies.
The central piece is "wgpu-hal/src/lib.rs".

### Metrics
| what                     | wgpu-core, LOC | graphics, LOC | Total | Build time |
| ----------------- | ----------------- | -------------- | ---- | ---------- |
| Old: _gfx-hal_      | 23K                       | 28K                 | **51K**  | 62s   |
| New: _wgpu-hal_ | 20.5K                    | 10.5K               | **31K** | 48s   |

### TODO:
- [x] wgpu-hal API
- [x] internal example/bench
- [x] Metal
- [x] Vulkan

**Testing**
examples

PR #1576, opened 2021-06-30
Switch all bitflag names to plural

kvark:
**Connections**
#1471 added a lot of singular bitflag names. Previously, we weren't really consistent.

**Description**
This PR changes all of the names to be plural, except for a few that don't imply singularity, and are matching WebGPU upstream. Maybe we need to convert them too?

**Testing**
self-tested

### DIFF
--- a/player/src/bin/play.rs
+++ b/player/src/bin/play.rs
@@ -41,11 +41,8 @@ fn main() {
         .build(&event_loop)
         .unwrap();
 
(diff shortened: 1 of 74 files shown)


cwfitzgerald (approved):

kvark:
Got green light from Veloren and Bevy folks.
bors r=cwfitzgerald

PR #1586, opened 2021-07-02
Assorted Fixes of Testing-Infra

cwfitzgerald:
**Connections**

Couple of low hanging fruit issues solved here. The biggest thing done is moving the masks to associated functions instead of being members of the bitflags, this is done so `all()` only includes the actual members.

Follows up on 579de425e5b2aa861531db8b7f3d12e4fc3fb2ba to mark vk/mipmap as passing

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -480,11 +480,18 @@ bitflags::bitflags! {
         ///
         /// This is a native only feature.
         const SPIRV_SHADER_PASSTHROUGH = 0x0000_0800_0000_0000;
+    }
+}
 
-        /// Features which are part of the upstream WebGPU standard.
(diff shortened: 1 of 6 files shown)


wgpu-types/src/lib.rs
```diff
@@ -657,8 +664,13 @@ bitflags::bitflags! {
...
         /// Supports samplers with anisotropic filtering
         const ANISOTROPIC_FILTERING = 0x0001_0000;
-        /// All flags are in their compliant state.
-        const COMPLIANT = 0x0000_13FF;
+    }
+}
+
+impl DownlevelFlags {
+    /// All flags that indicate if the backend is WebGPU compliant
+    pub fn compliant() -> Self {
```
kvark, line 672: can this be const fn?

kvark (approved):

cwfitzgerald:
bors r=kvark

PR #1589, opened 2021-07-02
[GL] Fix min_storage_buffer_offset_alignment param

Gordon-F:
**Connections**
Android Emulator (11) error:
```
2021-07-02 14:49:12.500 8292-8317/rust.example.cube E/emuglGLESv2_enc: device/generic/goldfish-opengl/system/GLESv2_enc/GL2Encoder.cpp:s_glGetIntegerv:876 GL error 0x500
```

**Description**
As a workaround set this parameter to default value.

### DIFF
--- a/wgpu-hal/src/gles/adapter.rs
+++ b/wgpu-hal/src/gles/adapter.rs
@@ -208,10 +208,10 @@ impl super::Adapter {
 
         let min_uniform_buffer_offset_alignment =
             gl.get_parameter_i32(glow::UNIFORM_BUFFER_OFFSET_ALIGNMENT);
-        let min_storage_buffer_offset_alignment = if cfg!(target_arch = "wasm32") {
-            256
+        let min_storage_buffer_offset_alignment = if ver >= (3, 1) {
+            gl.get_parameter_i32(glow::SHADER_STORAGE_BUFFER_OFFSET_ALIGNMENT) as u32
         } else {
-            gl.get_parameter_i32(glow::SHADER_STORAGE_BUFFER_OFFSET_ALIGNMENT)
+            256
         };
         let max_uniform_buffers_per_shader_stage =
             gl.get_parameter_i32(glow::MAX_VERTEX_UNIFORM_BLOCKS)


wgpu-hal/src/gles/adapter.rs
```diff
@@ -208,11 +208,9 @@ impl super::Adapter {
...
         let min_uniform_buffer_offset_alignment =
             gl.get_parameter_i32(glow::UNIFORM_BUFFER_OFFSET_ALIGNMENT);
-        let min_storage_buffer_offset_alignment = if cfg!(target_arch = "wasm32") {
-            256
-        } else {
-            gl.get_parameter_i32(glow::SHADER_STORAGE_BUFFER_OFFSET_ALIGNMENT)
-        };
+        let min_storage_buffer_offset_alignment = gl
+            .get_parameter_i32(glow::SHADER_STORAGE_BUFFER_OFFSET_ALIGNMENT)
+            .max(256);
```
kvark, line 213: We shouldn't hard-code the default here without referencing to it directly. I think storage buffers can only be queried on GLES 3.1, so here we just need to check for the version, and fall back to some trivial default instead of the actual limits default.
```rust
let min_storage_buffer_offset_alignment = if ver >= (3, 1) {
  gl.get_parameter_i32(glow::SHADER_STORAGE_BUFFER_OFFSET_ALIGNMENT)
} else {
  1
};
```

wgpu-hal/src/gles/adapter.rs
```diff
@@ -298,7 +296,11 @@ impl super::Adapter {
                     .unwrap(),
                 },
             },
-        })
+        };
+
+        log::debug!("{:?}", adapter);
```
kvark, line 301: If we want to print out the adapter, I think wgpu-core is a better place than here.

kvark (changes requested):
Thank you!

wgpu-hal/src/gles/adapter.rs
Gordon-F: Hmm, I think `Capabilities` is enough. No needed to print whole adapter. Can we print it here?

kvark: Capabilities are also not GL specific, so any printing of them should be in wgpu-core

Gordon-F: But wgpu-core will print only for selected adapter? (Vulkan or GL, for example). Should we print this info for all backend specific adapters?

kvark: I don't think so. Is this useful to print this on non-selected adapters?

Gordon-F: This might help with issue like this one #1590

kvark: This one happens on device request, where we are selecting the exact adapter. So printing out the info at that point is perfectly fine.

Gordon-F: I removed this part from PR, will look at how we can implement this in wgpu-core.

kvark (approved):
thank you!
bors r+

Issue #1590, opened 2021-07-02
examples panic when using GL backend (software render)


Fenex:
**Description**
I clone the repo and try to run [examples](https://github.com/gfx-rs/wgpu/tree/1fb77c6/wgpu/examples) `cube` , `hello-triangle`, and others with `WGPU_BACKEND=gl` env. All examples panic right after execute.

**Repro steps**
```
$ git clone https://github.com/gfx-rs/wgpu && cd wgpu
$ RUST_LOG=trace WGPU_BACKEND=gl cargo run --example hello-triangle
```

**Expected vs observed behavior**
Expected: Show window with example.
Observed: panic at start time. 

**Extra materials:**
<details>
  <summary> bash exe log</summary>

```
fenex@lubuntu-HP:~/repos/github/wgpu$ RUST_LOG=trace WGPU_BACKEND=gl cargo run --example hello-triangle
    Finished dev [unoptimized + debuginfo] target(s) in 0.12s
     Running `target/debug/examples/hello-triangle`
[2021-07-02T13:09:06Z INFO  wgpu_hal::vulkan::instance] Unable to find extension: VK_KHR_surface
[2021-07-02T13:09:06Z INFO  wgpu_hal::vulkan::instance] Unable to find extension: VK_KHR_xlib_surface
[2021-07-02T13:09:06Z INFO  wgpu_hal::vulkan::instance] Unable to find extension: VK_KHR_xcb_surface
[2021-07-02T13:09:06Z INFO  wgpu_hal::vulkan::instance] Unable to find extension: VK_KHR_wayland_surface
[2021-07-02T13:09:06Z INFO  wgpu_hal::vulkan::instance] Unable to find extension: VK_KHR_get_physical_device_properties2
[2021-07-02T13:09:06Z WARN  wgpu_hal::vulkan::instance] Unable to find layer: VK_LAYER_KHRONOS_validation
[2021-07-02T13:09:06Z WARN  wgpu_hal::vulkan::instance] create_instance: VkError(ERROR_INCOMPATIBLE_DRIVER)
[2021-07-02T13:09:06Z DEBUG wgpu_hal::gles::egl] Client extensions: [
        "EGL_EXT_device_base",
        "EGL_EXT_device_enumeration",
        "EGL_EXT_device_query",
        "EGL_EXT_platform_base",
        "EGL_KHR_client_get_all_proc_addresses",
        "EGL_EXT_client_extensions",
        "EGL_KHR_debug",
[...]

PR #1592, opened 2021-07-03
Detailed limits check and error

kvark:
**Connections**
Related to #1590

**Description**
The old comparison "A < B" was simply wrong, because it would do every field in order and could bail out mid-way without reaching for all limits.

**Testing**
Untested


cwfitzgerald (approved):
bors r+

Issue #1501, opened 2021-06-17
Vulkan leaks around 25mb/s on bunnymark example


kvark:
**Description**
Vulkan backend is leaking memory. Valgrind doesn't show much.

**Repro steps**
```bash
cargo run --release --example bunnymark
# reach >50k bunnies
# observe the memory growth
```

**Expected vs observed behavior**
No leaks are expected, memory should stabilize

**Extra materials**
- c73ce25f426f18e61c34f2930b27e2ce9137a952 helped this to not crash, but hasn't fixed the leak
- `halmark` doesn't show the leak

**Platform**
Linux/Nv/proprietary

mitchmindtree:
I think we're also running into this after updating from a nannou version that uses wgpu 0.6 or 0.7 to one that uses wgpu 0.9 in one of our recent installations.

Here's an example of one of the `panic!`s that can occur, though it seems it can occur in a few different places where wgpu allocates memory:

```
Running `/home/mindtree/programming/rust/morph/sw/target/release/morph-te`
MESA-INTEL: warning: Performance support disabled, consider sysctl dev.i915.perf_stream_paranoid=0

thread 'main' panicked at 'Handling wgpu errors as fatal by default', /home/mindtree/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.9.0/src/backend/direct.rs:1978:5
stack backtrace:
   0:     0x5573f95aa8d0 - std::backtrace_rs::backtrace::libunwind::trace::h63b7a90188ab5fb3
                               at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
   1:     0x5573f95aa8d0 - std::backtrace_rs::backtrace::trace_unsynchronized::h80aefbf9b851eca7
                               at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x5573f95aa8d0 - std::sys_common::backtrace::_print_fmt::hbef05ae4237a4d72
                               at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/sys_common/backtrace.rs:67:5
   3:     0x5573f95aa8d0 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h28abce2fdb9884c2
[...]

PR #1594, opened 2021-07-03
Generation of memory reports for Hubs

kvark:
**Connections**
Part of #1501 

**Description**
Tried to see if we aren't accidentally growing the hubs.

**Testing**
Examples...


cwfitzgerald (approved):
bors r+

PR #1595, opened 2021-07-03
Fix wgpu::Limits::default()

scoopr:
While using `wgpu::Limits::default()` for device limits, the result was
```
thread 'main' panicked at 'Error in Adapter::request_device: Limit 'max_storage_textures_per_shader_stage' value 8 is better than allowed 4', /Users/scoopr/Code/ext/wgpu/wgpu/src/backend/direct.rs:114:9
```
and
```
thread 'main' panicked at 'Error in Adapter::request_device: Limit 'max_storage_buffers_per_shader_stage' value 8 is better than allowed 4', /Users/scoopr/Code/ext/wgpu/wgpu/src/backend/direct.rs:114:9
```

This PR fixes the default to be the stated allowed value.


scoopr:
I guess the examples doing things with the downlevel limits made the CI miss this one?

cwfitzgerald:
We currently aren't actually testing limits on CI right now. However, this change won't work because WebGPU requires the minimum limits for both those limits be 8 and this is what the default represents. Anything below that is downlevel hardware. I'm working on a PR right now that adds a "downlevel" default limit set that should work on most hardware.

scoopr:
So the issue is that it is checking the given limits against some downlevel value instead of what the real normal value is? I’m not doing any downlevel stuff

cwfitzgerald:
What device is this running on (you can find out with adapter.get_info()), as it seems like this device only supports 4 storage_textures_per_shader_stage not 8. We assume 8 is available for compliance with WebGPU, which is why requesting the device fails.

scoopr:
I get the same on both of my adapters on my mbp16, I very much doubt that this is an actual limitation.. this panic started just started, last wgpu pull was maybe few days ago. 

```
AdapterInfo { name: "Intel(R) UHD Graphics 630", vendor: 0, device: 0, device_type: IntegratedGpu, backend: Metal }
```
```
AdapterInfo { name: "AMD Radeon Pro 5500M", vendor: 0, device: 0, device_type: DiscreteGpu, backend: Metal }
```

scoopr:
You can repro with the examples by uncommenthing this line from framework.rs:
```limits = test_common::lowest_reasonable_limits();```

cwfitzgerald:
This sounds it might be an issue with the metal backend not exposing limits as high as they should be. Re-opening as such.

cwfitzgerald:
Err... forgot this was a PR. Going to convert to an issue.

PR #1599, opened 2021-07-04
Fix Limits for lava/llvmpipe and re-enable and re-work CI

### LINKED ISSUE #1551: eglInitialize Fails on Headless llvmpipe System

With #1538 merged, we have the ability to test our graphics code in CI. Once dx12 and dx11 are implemented, we can use WARP to test these backeds on the windows build.

The two main software implementation of vulkan are lavapipe and swiftshader. Both have had issues. We need to evaluate them or any other options.

Relying solely on WARP and local testing of our test suite. We may also be able to rely on our todo network of testing machines.

cwfitzgerald:
**Connections**

No longer fixes #1551.

**Description**

This PR has a couple things going on at once.

- Fixes limits for llvmpipe, lavapipe, and rpi4.
- Added a downlevel limit to express that RPI4 does not allow storage buffers in vertex shaders on GL.
- Added a `Limits::downlevel_default()` that takes minimum limits from GLES and adds some more documentation on how to choose limits.
- Moved all examples to the new downlevel limits
- Reworks CI to re-enable software testing and unify instructions.

**Testing**

It is


cwfitzgerald:
CI is still mad at me, but the code should be otherwise good to look at.

.github/workflows/ci.yml
```diff
@@ -8,166 +8,160 @@ on:
...
+          
+          # Android
+          - name: Android aarch64
+            os: ubuntu-20.04
+            channel: nightly
+            target: aarch64-linux-android
+            kind: compile
+
+          
+          # WebGL
```
grovesNL, line 63: Is wasm/WebGPU missing here?

.github/workflows/ci.yml
```diff
@@ -8,166 +8,160 @@ on:
...
-          - name: MacOS Nightly
+            channel: stable
+            target: x86_64-apple-darwin
+            kind: compile
+
+          
+          # IOS
+          - name: IOS Nightly x86_64
             os: macos-10.15
             channel: nightly
```
grovesNL, line 36: We'd probably want to include iOS stable instead of nightly unless we can't do that for some reason

.github/workflows/ci.yml
```diff
@@ -8,166 +8,160 @@ on:
...
-          - name: Windows Nightly
-            os: windows-2019
+            target: x86_64-unknown-linux-gnu
+            kind: compile
+
+          
+          # Android
+          - name: Android aarch64
+            os: ubuntu-20.04
+            channel: nightly
```
grovesNL, line 58: Similarly I think we'd want to target Android and wasm on stable instead of nightly

.github/workflows/ci.yml
cwfitzgerald: Yeah this includes webgpu because all web builds unconditionally include it. In practice right now I have webgl disabled as wgc doesn't compile on wasm rn.

.github/workflows/ci.yml
cwfitzgerald: Ack, yeah this should be stable. The old ci script had nightly meaning not running tests, so I forgot to change it back to stable.

wgpu-core/src/device/mod.rs
```diff
@@ -1029,6 +1030,9 @@ impl<A: HalApi> Device<A> {
                         error,
                     })?;
             }
+            if entry.visibility.contains(wgt::ShaderStages::VERTEX) {
+                required_downlevel_flags |= wgt::DownlevelFlags::VERTEX_ACCESSABLE_STORAGE_BUFFERS;
```
kvark, line 1034: would it make sense to call it `VERTEX_READABLE_STORAGE_BUFFERS`?

wgpu-types/src/lib.rs
```diff
@@ -593,6 +607,32 @@ impl Default for Limits {
     }
 }
 
+impl Limits {
+    /// These default limits are guarenteed to be compatible with GLES3, WebGL, and D3D11
+    pub fn downlevel_limits() -> Self {
```
kvark, line 612: can we rename this method a bit? `Limits::downlevel_limits` is a bit repeat-y.
Maybe `Limits::downlevel_default` or something like this?

wgpu-types/src/lib.rs
```diff
@@ -593,6 +607,32 @@ impl Default for Limits {
...
+        Self {
```
kvark, line 613: would be good to add a test in wgpu-core comparing this to the default limit and ensuring the order, but this isn't a blocker for landing

wgpu-types/src/lib.rs
```diff
@@ -662,8 +702,10 @@ bitflags::bitflags! {
         const COMPARISON_SAMPLERS = 0x0000_0100;
         /// Supports different blending modes per color target.
         const INDEPENDENT_BLENDING = 0x0000_0200;
-        /// Supports samplers with anisotropic filtering
-        const ANISOTROPIC_FILTERING = 0x0001_0000;
+        /// Supports samplers with anisotropic filtering.
+        const ANISOTROPIC_FILTERING = 0x0000_0400;
```
kvark, line 706: can we keep the special status for anisotropic? it's the only feature that's not required.
So maybe have it always at the end, and with a value distinct from all others?

kvark (approved):
This is great! A few notes, otherwise land when ready (and CI is happy).

wgpu-core/src/device/mod.rs
cwfitzgerald: Imo no, cause that would imply that they can be something else (i.e. write only)

wgpu-types/src/lib.rs
cwfitzgerald: Amusingly that's the name I put in the PR description, will change :)

wgpu-types/src/lib.rs
cwfitzgerald: We totally could, but I don't see the real point considering flags are supposed to be opaque.

Wumpf: some note that it is not required to be webgpu conform would be nice though - given that the flags are documented as `Binary flags listing various ways the underlying platform does not conform to the WebGPU standard.` it looks now like it would be required for conforimity

.github/workflows/ci.yml
grovesNL: Could we have separate checks for WebGPU and WebGL, or name the check “Web Assembly” if one check does both? i.e. the check is named “WebGL” here but it seems to check both WebGL and WebGPU

cwfitzgerald:
Alright, all reviews addressed, CI is passing, lets goo!

bors r=kvark,groves,wumpf

cwfitzgerald:
bors r-

github's UI lied to me and said all 8 passed when only 2 did.

cwfitzgerald:
bors r=kvark,groves,wumpf

cwfitzgerald:
bors retry

PR #1602, opened 2021-07-05
New DX12 backend (unfinished)

kvark:
**Connections**
Continuation of #1471

**Description**
Mostly fresh implementation on DX12 backend for wgpu-hal
Core changes:
  - texture usage is filtered before passing to `create_texture_view`, so that wgpu-hal can assume every bit is important.
  - all the copies involving textures are limited to one array layer at a time
  - remove render pass boundary usages

**Testing**
Examples (not running yet)


niklaskorz:
Out of interest, what benefits does the `winapi` crate have over Microsoft's official `windows-rs` crate which also includes bindings for DX12?

Gordon-F:
> Out of interest, what benefits does the `winapi` crate have over Microsoft's official `windows-rs` crate which also includes bindings for DX12

Related discussion in winit - https://github.com/rust-windowing/winit/pull/1938

kvark:
I think it's at the point where we can merge and follow-up. I'm not able to test it thoroughly because of Naga's HLSL backend, which needs work. I'll be collaborating with @Gordon-F on this one, and in the meantime DX12 is removed from the `PRIMARY` set of backends.

Another thing missing is rebinding all the bindings. Currently, the only lazy state is vertex buffers, because of https://github.com/gpuweb/gpuweb/issues/1921. However, DX12 prescribes all the bindings to be reset on any layout change. This is unfortunate, since wgpu-core's handling of Vulkan pipeline compatibility is wasted, and we have to have more lazy state in DX12 now... See also - https://github.com/gpuweb/gpuweb/issues/1926

PR #1604, opened 2021-07-05
Upgrade `arrayvec` to 0.7

a1phyr:
Upgrade `arrayvec` to version 0.7 (using const generics).


cwfitzgerald (approved):
Good to go after CI is happy

a1phyr:
I forgot to `cargo fmt`

kvark:
Let's not rush this, we need to make sure the MSRV is reasonable here for the clients.

a1phyr:
I thought it would be fine because I could not find any MSRV policy and only `stable` is enforced in CI.

kvark:
@a1phyr you are right, we don't have a policy in place. Especially once teams start forking the project to run on consoles, they'd be very hesitant to update Rust.
So for now, we only have common sense to rely on.
Looks like this PR bumps the MSRV to 1.51. Let's add a badge for it as a follow-up.
bors r+

PR #1606, opened 2021-07-05
Fix macos build

scoopr:
Building on macos failed with
```
   Compiling wgpu-hal v0.9.0 (/Users/scoopr/Code/ext/wgpu/wgpu-hal)
error[E0609]: no field `entry` on type `&vulkan::Instance`
   --> /Users/scoopr/Code/ext/wgpu/wgpu-hal/src/vulkan/instance.rs:267:61
    |
267 |             let metal_loader = ext::MetalSurface::new(&self.entry, &self.shared.raw);
    |                                                             ^^^^^ unknown field
    |
    = note: available fields are: `shared`, `extensions`

error: aborting due to previous error
```

The PR fixes the issue.

### DIFF
--- a/wgpu-hal/src/vulkan/instance.rs
+++ b/wgpu-hal/src/vulkan/instance.rs
@@ -264,7 +264,7 @@ impl super::Instance {
         };
 
         let surface = {
-            let metal_loader = ext::MetalSurface::new(&self.entry, &self.shared.raw);
+            let metal_loader = ext::MetalSurface::new(&self.shared.entry, &self.shared.raw);
             let vk_info = vk::MetalSurfaceCreateInfoEXT::builder()
                 .flags(vk::MetalSurfaceCreateFlagsEXT::empty())
                 .layer(layer as *mut _)


scoopr:
I'm again a bit perplexed how the CI didn't catch this one.

As I'm looking at the CI steps
```
 cargo clippy --target x86_64-apple-darwin -p wgpu -p wgpu-core -p wgpu-info -p player -- -D warnings
  
  # check with all features
  # explicitly don't mention wgpu-hal so that --all-features don't apply to it
  cargo clippy --target x86_64-apple-darwin -p wgpu -p wgpu-core -p wgpu-info -p player --examples --tests --all-features -- -D warnings 
  ```
Why does it skip the wgpu-hal? Testing locally, it seems if it had been listed it would of caught it, and with the fix it works fine

scoopr:
I suppose enabling metal on other platforms might error out.. does that mean that the CI matrix needs to list out hal features for some of the jobs?

cwfitzgerald:
> Why does it skip the wgpu-hal?

Yeah I skip wgpu-hal because wgpu should automatically add the needed features that are needed and we can't enable the wrong backend on the wrong platform. I'm also perplexed by this....

cwfitzgerald:
Ohhh, I see, it's the vulkan-portability path that isn't getting tested. Idk why vulkan would be enabled at all by default on mac.

Either way the change is fine.

bors r+

Issue #1583, opened 2021-07-02
test_compute_N fails on V4D on GL (rpi4)


cwfitzgerald:
**Description**

The output of the hello-compute tests is partially garbage.

**Repro steps**

Run `WGPU_BACKEND=gl cargo test --example hello-compute -- --test-threads=1 --nocapture` on a raspeberry pi 4.

**Expected vs observed behavior**

Expected: Correct tests
Observed:
```
running 4 tests
test tests::test_compute_1 ... thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `[0, 2, 3, 4]`,
 right: `[0, 1, 7, 2]`', wgpu/examples/hello-compute/tests.rs:89:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
GOT EXPECTED TEST FAILURE: BACKEND | VENDOR | ADAPTER
ok
test tests::test_compute_2 ... thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `[5, 23, 10, 9]`,
 right: `[5, 15, 6, 19]`', wgpu/examples/hello-compute/tests.rs:89:9
GOT EXPECTED TEST FAILURE: BACKEND | VENDOR | ADAPTER
ok
test tests::test_compute_overflow ... thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `[350, 837799, 8400511, 63728127]`,
 right: `[350, 524, 4294967295, 4294967295]`', wgpu/examples/hello-compute/tests.rs:89:9
GOT EXPECTED TEST FAILURE: BACKEND | VENDOR | ADAPTER
```

**Extra materials**
I _expect_ this to be a driver bug, but we shouldn't rule out us messing up, or us being able to compensate in some way.

**Platform**
raspbian buster running a compiled-from-source version of mesa dating like 4-6 months back.

kvark:
The test works on AMD 3500U fwiw.

cwfitzgerald:
And on an intel 620, so this is almost definitely a driver bug of some kind.

kvark:
Barriers, perhaps?..

cwfitzgerald:
Turns out they are spurious failures on both VK and GL. Both possibly driver bugs.

cwfitzgerald:
These are hard skipped on v3d https://github.com/gfx-rs/wgpu/blob/master/wgpu/examples/hello-compute/tests.rs#L18

cwfitzgerald:
[...]

PR #1607, opened 2021-07-06
Fix Downlevel Vertex Stage Storage Buffer Check

cwfitzgerald:
**Connections**

Fixes a bug in #1599. Also follows up on #1583 marking all of hello-compute as a failure.

**Testing**

Tests now pass on rpi4.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -976,22 +976,33 @@ impl<A: HalApi> Device<A> {
         label: Option<&str>,
         entry_map: binding_model::BindEntryMap,
     ) -> Result<binding_model::BindGroupLayout<A>, binding_model::CreateBindGroupLayoutError> {
+        #[derive(PartialEq)]
+        enum WritableStorage {
+            Yes,
+            No,
+        }
+
         for entry in entry_map.values() {
             use wgt::BindingType as Bt;
 
             let mut required_features = wgt::Features::empty();
-            let mut required_downlevel_flags = wgt::DownlevelFlags::empty();
-            let (array_feature, is_writable_storage) = match entry.ty {
+            let (array_feature, writable_storage) = match entry.ty {
                 Bt::Buffer {
                     ty: wgt::BufferBindingType::Uniform,
                     has_dynamic_offset: false,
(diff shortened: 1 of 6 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -981,17 +981,17 @@ impl<A: HalApi> Device<A> {
...
             let mut required_features = wgt::Features::empty();
             let mut required_downlevel_flags = wgt::DownlevelFlags::empty();
-            let (array_feature, is_writable_storage) = match entry.ty {
+            let (array_feature, storage_buffer, is_writable_storage) = match entry.ty {
                 Bt::Buffer {
                     ty: wgt::BufferBindingType::Uniform,
                     has_dynamic_offset: false,
                     min_binding_size: _,
-                } => (Some(wgt::Features::BUFFER_BINDING_ARRAY), false),
+                } => (Some(wgt::Features::BUFFER_BINDING_ARRAY), false, false),
```
kvark, line 989: Would you mind turning this into an enum? booleans in such position are bad, and having two of them is twice as bad:)

kvark:
The SSBO in VS support is pretty sad story - https://www.raspberrypi.org/forums/viewtopic.php?t=271863
Here is my reading of this:
  1. Tiling GPUs don't want *writable* storage resources that can be tested.
  2. Regular writable SSBOs and images can *not* be tested, because accessing them within the same draw call is technically UB.
  3. Only atomic access to SSBO can be tested, and that's what the drivers do *not* want us to do.
  4. GLES doesn't have a proper limit to express this, but Vulkan and WebGPU do. We have `VERTEX_WRITABLE_STORAGE` that captures this much better.
  5. Perhaps we can detect if 0 is returned for `MAX_VERTEX_SHADER_STORAGE_BLOCKS` and ignore it (while not exposing `VERTEX_WRITABLE_STORAGE`)? Unclear what to do if there is a GL error. May need to hack a workaround in the error callback for this specifically :(

cwfitzgerald:
Alright, this is good to go on my machine

kvark (commented):
Great, thank you!
bors r+

PR #1609, opened 2021-07-06
[Vulkan] Initialize wgpu objects from raw handles

zarik5:
**Connections**
This PR is a successor of https://github.com/gfx-rs/gfx/pull/3762

**Description**
The `handle_is_external` flag mechanism was not enough to ensure safety, when for example the Instance is wrapped in `Arc<>` and a library cannot keep track of all its clones. This is the case with the bevy render rewrite. The solution was to let wgpu be in charge of destroying the handles, but it also has to keep a reference to a "drop guard" which is always dropped after the handle is destroyed. For the OpenXR integration, this drop guard will be `xr::Instance`.

For now this is just a proof of concept. Only instance creation is handled, and there is even type error in `wgc::Instance::from_hal()`.

I have a few concerns:
* Is it ok to expose `hal::Instance` from the wgpu crate? Or should the user pass all the parameters so `hal::Instance` can be constructed internally? This second options is more disruptive, since the wgpu-types crate would need to import all platform-specific crates to define the structure/enum to hold the raw handles.
* Without counting the call to create the `hal::Instance`, there are 4 indirection calls to save the raw instance. Can this be optimized in any way?

Do you think it is possible to merge wgpu-hal into wgpu-core? This could help with reducing the distance from the surface level API to the platform specific APIs even more.

**Testing**
Vulkan/Android (Oculus Quest) using [this sample](https://github.com/zarik5/openxrs/blob/wgpu-test/openxr/examples/vulkan.rs).

### DIFF
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -69,9 +69,10 @@ fn downlevel_default_limits_less_than_default_limits() {
     )
 }
 
+#[derive(Default)]
 pub struct Instance {
     #[allow(dead_code)]
-    name: String,
+    pub name: String,
     #[cfg(vulkan)]
     pub vulkan: Option<HalInstance<hal::api::Vulkan>>,
     #[cfg(metal)]
@@ -248,9 +249,10 @@ impl<A: HalApi> Adapter<A> {
         }
     }
 
-    fn create_device(
+    fn create_device_from_hal(
         &self,
         self_id: AdapterId,
+        open: hal::OpenDevice<A>,
         desc: &DeviceDescriptor,
         trace_path: Option<&std::path::Path>,
     ) -> Result<Device<A>, RequestDeviceError> {
@@ -281,11 +283,6 @@ impl<A: HalApi> Adapter<A> {
             log::warn!("Feature MAPPABLE_PRIMARY_BUFFERS enabled on a discrete gpu. This is a massive performance footgun and likely not what you wanted");
         }
 
-        let gpu = unsafe { self.raw.adapter.open(desc.features) }.map_err(|err| match err {
-            hal::DeviceError::Lost => RequestDeviceError::DeviceLost,
-            hal::DeviceError::OutOfMemory => RequestDeviceError::OutOfMemory,
-        })?;
-
         if let Some(_) = desc.label {
             //TODO
         }
@@ -305,7 +302,7 @@ impl<A: HalApi> Adapter<A> {
         }
 
         Device::new(
-            gpu,
+            open,
             Stored {
                 value: Valid(self_id),
                 ref_count: self.life_guard.add_ref(),
@@ -316,6 +313,20 @@ impl<A: HalApi> Adapter<A> {
             trace_path,
         )
         .or(Err(RequestDeviceError::OutOfMemory))
+    }
+
+    fn create_device(
+        &self,
+        self_id: AdapterId,
+        desc: &DeviceDescriptor,
+        trace_path: Option<&std::path::Path>,
+    ) -> Result<Device<A>, RequestDeviceError> {
+        let open = unsafe { self.raw.adapter.open(desc.features) }.map_err(|err| match err {
+            hal::DeviceError::Lost => RequestDeviceError::DeviceLost,
+            hal::DeviceError::OutOfMemory => RequestDeviceError::OutOfMemory,
+        })?;
+
+        self.create_device_from_hal(self_id, open, desc, trace_path)
     }
 }
 
@@ -649,6 +660,34 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         Err(RequestAdapterError::NotFound)
     }
 
+    /// # Safety
+    ///
+    /// `hal_adapter` must be created from this global internal instance handle.
+    pub unsafe fn create_adapter_from_hal<A: HalApi>(
+        &self,
+        hal_adapter: hal::ExposedAdapter<A>,
+        input: Input<G, AdapterId>,
+    ) -> AdapterId {
+        profiling::scope!("create_adapter_from_hal", "Instance");
+
+        let mut token = Token::root();
+        let fid = A::hub(&self).adapters.prepare(input);
+
+        match A::VARIANT {
+            #[cfg(vulkan)]
+            Backend::Vulkan => fid.assign(Adapter::new(hal_adapter), &mut token).0,
+            #[cfg(metal)]
+            Backend::Metal => fid.assign(Adapter::new(hal_adapter), &mut token).0,
+            #[cfg(dx12)]
+            Backend::Dx12 => fid.assign(Adapter::new(hal_adapter), &mut token).0,
+            #[cfg(dx11)]
+            Backend::Dx11 => fid.assign(Adapter::new(hal_adapter), &mut token).0,
+            #[cfg(gl)]
+            Backend::Gl => fid.assign(Adapter::new(hal_adapter), &mut token).0,
+            _ => unreachable!(),
+        }
+    }
+
     pub fn adapter_get_info<A: HalApi>(
         &self,
         adapter_id: AdapterId,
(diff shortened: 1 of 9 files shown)


kvark:
> Is it ok to expose hal::Instance from the wgpu crate?

I think it's OK as long as it's native-only, and it's `unsafe`.

> Without counting the call to create the hal::Instance, there are 4 indirection calls to save the raw instance. Can this be optimized in any way?

This should be nicer once @pythonesque work is finished.

> Do you think it is possible to merge wgpu-hal into wgpu-core? This could help with reducing the distance from the surface level API to the platform specific APIs even more.

No. Merging them together would make a big mess. It is a very nice boundary between platform-agnostic and platform-specific code right now.

wgpu-core/src/instance.rs
```diff
@@ -118,6 +118,46 @@ impl Instance {
         }
     }
 
+    pub fn from_hal<A: HalApi>(name: &str, raw_instance: A::Instance) -> Self {
+        #[cfg(vulkan)]
+        let mut vulkan_instance = None;
```
kvark, line 123: can probably start with `let mut this = Self::default()` and mutate it, to save a ton of lines

zarik5:
> This should be nicer once @pythonesque work is finished.

That is good news 🙂

wgpu-core/src/instance.rs
zarik5: I'm not sure how I would use this. The problem I was having here, is that `<A as hal::api::Api>::Instance` cannot be converted to `hal::api::Vulkan::Instance`. I'm confused since `wgc::Instance::new()` seems so be doing just that.

kvark: What I'm thinking is:
```rust
let mut this = Self::default();
this.name = name.to_string();
match A::VARIANT {
...
}
Ok(this)
```

zarik5: it seems that the problem is that matching by `VARIANT` doesn't actually constrain the type `A`. Maybe this could be fixed with full const generics but we don't have that yet. For now I can wrap hal::Instance into an enum.

kvark: right, we can't have this method as generic over A

kvark: hmm, or maybe we can. There is `get_surface_mut`, which works with generic A. Something similar can be done for instances

zarik5: I tried a few more type manipulations but still no luck. The enum solution works.

zarik5: I did it, at least I think I did what you suggested. I used the `HalApi` trait to convert from hal to wgc instance (while `get_surface_mut` converts a wgc swapchain to hal).

wgpu-hal/src/lib.rs
```diff
@@ -1083,6 +1083,19 @@ pub struct ComputePassDescriptor<'a> {
     pub label: Label<'a>,
 }
 
+pub enum RawInstance {
```
kvark: if it's only needed by wgpu-core, it should be defined there, not here

wgpu-core/src/instance.rs
```diff
@@ -118,44 +119,30 @@ impl Instance {
...
-        let mut dx12_instance = None;
-        #[cfg(dx11)]
-        let mut dx11_instance = None;
-        #[cfg(gl)]
-        let mut gl_instance = None;
-        match A::VARIANT {
+    /// # Safety
+    ///
+    /// Refer to the creation of wgpu-hal Instance for every backend.
+    pub unsafe fn from_hal(name: &str, raw_instance: hal::RawInstance) -> Self {
```
kvark: Does this need to be overly generic? Are you planning on passing anything but a Vulkan instance in here?
Perhaps this can be `from_vulkan` instead

wgpu-hal/src/lib.rs
zarik5: This code is already outdated.

wgpu-core/src/instance.rs
zarik5: This code is already outdated.

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -114,6 +114,125 @@ impl super::Swapchain {
 }
 
 impl super::Instance {
+    pub fn required_extensions(
+        entry: &ash::Entry,
+        driver_api_version: u32,
+        with_debug: bool,
```
kvark, line 120: Let's just pass `hal::InstanceFlags` here. Since the external user would pass in a `hal::Instance`, they already depend on wgpu-hal

kvark (commented):
So this is needed for VR, right? Would using VR in any other backend require to pass in a raw `hal::Instance` to wgpu?
It seems to me that this way of initialization is Vulkan-only, and so we shouldn't try to generalize it (right now).

zarik5:
> So this is needed for VR, right? Would using VR in any other backend require to pass in a raw hal::Instance to wgpu?
It seems to me that this way of initialization is Vulkan-only, and so we shouldn't try to generalize it (right now).

The DirectX path for OpenXR has similar requirements to Vulkan. In both cases the OpenXR runtime provides handles about the adapter. Vulkan has actually two initialization paths in OpenXR (extension v1 and v2) and both are needed to support all platforms. Either the handles have to be externally created using a list of extensions, or the list of extensions is sent to OpenXR which creates the handles. Both cases can be covered in this PR using the same code (when it is finished).

kvark:
Interesting!
> In both cases the OpenXR runtime provides handles about the adapter

But the current PR passes in `hal::Instance`, not `hal::Adapter`, so that's what I'm wondering about. Does DX integration need the adapter? the instance? or you want to pass the instance so that adapter configuration is easier?

zarik5:
> But the current PR passes in hal::Instance, not hal::Adapter

I just started writing this PR. It still needs support for creating the Adapter, Device, and Texture. I opened this draft PR to get early feedback.

zarik5:
I added support for creating the adapter and device. For creating the device there is some code duplication in wgpu and wgpu-core because I wasn't sure how to rearrange the code.

`Global::adapter_from_hal()` and `Global::device_from_hal()` take an `Input<>` like other `Global` methods. This parameter is useless since I only pass `PhantomData` in wgpu, but I was not able to remove it. If I use `PhantomData` inside these methods I get various type errors. Also I don't know what `Input<>`s are for and I don't understand what the majority of the code does. @kwark Can you explain it to me briefly? Or can you point me to some documentation?

kvark:
>  don't know what Input<>s are for and I don't understand what the majority of the code does

It's a generic abstraction over the fact that some `wgpu-core` clients need to generate and provide the IDs explicitly, while others (like wgpu-rs) are cool with `wgpu-core` generating the IDs.
For example, Gecko generates the ID in the content process, then sends it together with the descriptor into the parent process, where `wgpu-core` actually does the object creation. And it's given the ID.
See https://searchfox.org/mozilla-central/rev/3ba42a92cea3e61163091e7fe4d22bcda74a1e5a/gfx/wgpu_bindings/src/identity.rs#33

zarik5:
Thanks!

zarik5:
The code is more or less done, it still needs testing (I don't have an ETA, it depends on various factors). I'm not happy with the `Global` methods, what I did is just copy-pasting similar functions with redundant code. I would be more confident polishing this after Hub-less and Input-less but I know this is far down the road.

zarik5:
The PR is ready for review. I tested this on the Oculus Quest using [this modified sample](https://github.com/zarik5/openxrs/blob/wgpu-test/openxr/examples/vulkan.rs). The drop-safety mechanism seems to work as intended. In the example I linked, there is still the need for dropping some objects manually, but this is because the openxrs crate also needs some sort of drop-safety mechanism. Basically the drop order that must be enforced is: (wgpu textures) --(1)--> (xr swapchain, xr session) --(2)--> (wgpu device, wgpu instance) --(3)--> (xr instance). (1) and (3) are enforced automatically by this PR, (2) is enforced manually in the example.

I haven't cleaned up the Global methods but I think that once hubs and inputs are removed they will slim down by their own.

wgpu-core/src/instance.rs
```diff
@@ -248,9 +249,10 @@ impl<A: HalApi> Adapter<A> {
         }
     }
 
-    fn create_device(
+    fn device_from_open(
```
kvark, line 252: should be `create_device_from_hal`

wgpu-hal/src/vulkan/device.rs
```diff
@@ -535,6 +535,40 @@ impl super::Device {
             config: config.clone(),
         })
     }
+
+    pub fn image_raw_flags(desc: &crate::TextureDescriptor) -> vk::ImageCreateFlags {
+        let mut raw_flags = vk::ImageCreateFlags::empty();
+        if desc.dimension == wgt::TextureDimension::D2 && desc.size.depth_or_array_layers % 6 == 0 {
```
kvark, line 541: this is a dirty hack, but we still need to do our best
let's also check for width == height here

wgpu-hal/src/vulkan/device.rs
```diff
@@ -535,6 +535,40 @@ impl super::Device {
             config: config.clone(),
         })
     }
+
+    pub fn image_raw_flags(desc: &crate::TextureDescriptor) -> vk::ImageCreateFlags {
```
kvark, line 539: actually, how about not exposing this at all? I don't think XR cares about cube views, anyway. This function is a hack, and hacks should ideally stay private

wgpu-hal/src/vulkan/mod.rs
```diff
@@ -257,6 +260,8 @@ pub struct Buffer {
 #[derive(Debug)]
 pub struct Texture {
     raw: vk::Image,
+    handle_is_external: bool,
+    _drop_guard: DropGuard,
```
kvark, line 264: can we combine these two? something like `external_drop: Option<DropGuard>`

kvark (commented):
We chatted about this on Bevy's discord and agreed that generally this is the only way to go. Just some details to polish are left.

zarik5:
I implemented your suggestions. I didn't have the time to throughly test the last commit. If you need anything else I will be available in a couple of days.

kvark (approved):
Thank you! Please squash the commits

zarik5:
Rebased

kvark (approved):
bors r+

PR #1612, opened 2021-07-07
Update `web-sys`

a1phyr:
`wgpu` uses unstable APIs so the version has to be pinned and manually updated.

No breaking changes happened in `0.3.51` so this update is easy.


grovesNL (approved):
Thanks!
bors r+

PR #1618, opened 2021-07-09
webgpu: pass wgsl directly

grovesNL:
**Connections**
Matrix conversation about passing WGSL to Gecko

**Description**
Gecko accepts WGSL now, so we can pass it directly instead of going through naga.

We could also consider converting SPIR-V to WGSL soon, assuming nobody is relying on the SPIR-V path with older versions of Gecko. We could do that in a separate PR though.

**Testing**
Tried hello-compute locally, using the examples HTML file and creating wasm with something like
```
RUSTFLAGS=--cfg=web_sys_unstable_apis cargo build --release --example hello-compute --target wasm32-unknown-unknown
wasm-bindgen target/wasm32-unknown-unknown/release/examples/hello-compute.wasm --out-dir target/generated --web
```

### DIFF
--- a/wgpu/Cargo.toml
+++ b/wgpu/Cargo.toml
@@ -82,12 +82,6 @@ git = "https://github.com/gfx-rs/naga"
 rev = "458db0b"
 features = ["wgsl-in"]
 
-# used to generate SPIR-V for the Web target
-[target.'cfg(target_arch = "wasm32")'.dependencies.naga]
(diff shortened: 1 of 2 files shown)


kvark (commented):
Sweet, thank you!
Let's also remove the `naga/spv-out` dependency from wasm32 now (in Cargo toml)

kvark (approved):
Thank you!
bors r+

Issue #1621, opened 2021-07-09
`device.create_texture_with_data` implicitly requires `COPY_DST`


HalfVoxel:
**Description**
The documentation for `device.create_texture_with_data` doesn't mention that the usage must contain `COPY_DST`, but it also doesn't implicitly add that flag. I think it should either mention that `COPY_DST` is required or add it implicitly. Not sure which is the desired way to go though.

Right now it fails with:
```
thread 'main' panicked at 'Error in Queue::write_texture: destination buffer/texture is missing the `COPY_DST` usage
```

cwfitzgerald:
We should just implicitly add this as there is no performance downside to having COPY_DST.

PR #1622, opened 2021-07-09
Implicitly add `COPY_DST` when using `create_texture_with_data`

HalfVoxel:
**Connections**
Fixes: #1621

**Description**
> The documentation for device.create_texture_with_data doesn't mention that the usage must contain COPY_DST, but it also doesn't implicitly add that flag. I think it should either mention that COPY_DST is required or add it implicitly. Not sure which is the desired way to go though.

It feels kinda ugly to have to clone the descriptor. It's not a big deal performance-wise, but still.
An alternative to this would be a nicer validation error message.

### DIFF
--- a/wgpu/src/util/device.rs
+++ b/wgpu/src/util/device.rs
@@ -82,14 +82,10 @@ impl DeviceExt for crate::Device {
         desc: &crate::TextureDescriptor,
         data: &[u8],
     ) -> crate::Texture {
-        let texture = if desc.usage.contains(crate::TextureUsages::COPY_DST) {
-            self.create_texture(desc)
-        } else {
-            // Implicitly add the COPY_DST usage
-            let mut desc = desc.to_owned();
-            desc.usage |= crate::TextureUsages::COPY_DST;
-            self.create_texture(&desc)
-        };
+        // Implicitly add the COPY_DST usage
+        let mut desc = desc.to_owned();
+        desc.usage |= crate::TextureUsages::COPY_DST;
+        let texture = self.create_texture(&desc);
 
         let format_info = desc.format.describe();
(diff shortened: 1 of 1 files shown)


cwfitzgerald:
Hmmm. I wonder if we should just take the descriptor by value if we're just going to clone it anyway.

HalfVoxel:
But that would be a breaking change, and it would go against the current wgpu convention, wouldn't it? I don't think there's any other place where a descriptor is passed by value?

kocsis1david:
Wouldn't it be possible to only copy the descriptor? like `let mut desc = desc;`

cwfitzgerald:
@HalfVoxel it would be a breaking change, but we don't really mind that here.

@kocsis1david it's not `Copy`.

wgpu/src/util/device.rs
```diff
@@ -79,7 +82,14 @@ impl DeviceExt for crate::Device {
         desc: &crate::TextureDescriptor,
         data: &[u8],
     ) -> crate::Texture {
-        let texture = self.create_texture(desc);
+        let texture = if desc.usage.contains(crate::TextureUsages::COPY_DST) {
+            self.create_texture(desc)
+        } else {
+            // Implicitly add the COPY_DST usage
```
kvark, line 88: let's just always do this

kvark (commented):
Thank you!

wgpu/src/util/device.rs
HalfVoxel: Changed

HalfVoxel:
I simplified the code to always do the clone. However, I left the parameter as a reference to keep the api following the wgpu conventions.

kvark (approved):
Thank you!

PR #1623, opened 2021-07-10
Enable All Backends for Trace Player

zicklag:
**Connections**
None

**Description**
I just spent a bit of time trying to figure out how to get the trace player to play an OpenGL trace before realizing that only the primary backends were enabled at instance creation. This just enables all backends in the trace player. I believe that this should be harmless as it shouldn't change anything for traces on the primary supported backends.

**Testing**
It's not tested specifically, but it's such a minor change it shouldn't matter, I think.

### DIFF
--- a/player/src/bin/play.rs
+++ b/player/src/bin/play.rs
@@ -41,8 +41,7 @@ fn main() {
         .build(&event_loop)
         .unwrap();
 
-    let global =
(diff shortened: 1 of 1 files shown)


cwfitzgerald (approved):
Thanks!

bors r+

PR #1628, opened 2021-07-10
Fix Color and Clear Problems on OpenGL Backend

### LINKED ISSUE #1627: Fastclear Bug With Intel Mesa Adapters on the GL Backend
**Description**
When using the OpenGL backend on Linux the clear color seems to behaving strangely. For one, the color itself is lighter than it should be. For two, the borders around any objects drawn on top of the clear color have an even lighter, pixelated version of the clear color.

I'm currently troubleshooting but I opened the issue to start the discussion. With my experimentation so far it seems completely related to the act of clearing the draw framebuffer. All other rendering and examples seem to work fine, and if you have an example where you can't see the clear color, such as the skybox example, everything looks great.

Also I've noticed some weird related behavior in my Renderdoc captures:

When I launch the example with Renderdoc it has the same problem as running it without Renderdoc ( which makes sense ):

image

And when I view the renderbuffer contents after the initial clear in renderdoc, it shows the clear color like it shows in the render, lighter than it should be:

image

The clear color shows the same in the draw step, all the way until the final framebuffer blit, where it is dark enough ( but still with pixelated edges ):

image

Yet, when I hover over the pixels in the image, the little thumbnail at the bottom shows the wrong ligher color:

image

Very strange.

I found that I could get rid of the pixelated edges by forcing the renderbuffer pixel format to be `RGBA8`, but the color was still off. I think thats the closest lead I have and I'm going to look into how different pixel formats efffect it, and maybe try binding the framebuffer storage to a texture instead of a renderbuffer and see if that makes any difference.

---

**PS:** Very excited that the new GL backend on `wgpu-hal` is working for all the examples! This is the first time I've tried it that the `shadow`, `boids`,  and `skybox` examples have worked. I might try to tackle #1617, but I figured I'd try to get this one out of the way first. :)

**Repro steps**
Run the `cube` or `shadow` example with the OpenGL backend.

There should be no pixelated edges around objects and the clear color should be darker.

**Expected:**
image

**Actual ( I think it's fine if the lines for the trangles aren't there, the issue is the background color ) :**
image

**Extra materials**

**Shadow example:**
image

**Platform**
Running WGPU `cube` or `shadow` examples on Linux Pop!_OS ( Ubuntu ) 20.04. Adapter info:

```
[2021-07-10T16:54:11Z INFO  wgpu_hal::gles::adapter] Vendor: Intel
[2021-07-10T16:54:11Z INFO  wgpu_hal::gles::adapter] Renderer: Mesa Intel(R) UHD Graphics (CML GT2)
[2021-07-10T16:54:11Z INFO  wgpu_hal::gles::adapter] Version: OpenGL ES 3.2 Mesa 20.0.8
[2021-07-10T16:54:11Z INFO  wgpu_hal::gles::adapter] SL version: OpenGL ES GLSL ES 3.20
```

zicklag:
- Disabled writing to the alpha channel of the draw framebuffer. This
   fixes pixelated edges around the borders of the clear color.
- Disable sRGB mode when creating the EGL surface. This was causing the
  color of the OpenGL backend to appear "washed-out" compared to the
  other backends.

**Connections**
Fixes #1627

**Description**
Fixes some rendering differences between OpenGL and Vulkan backends on Linux:

**Before:**

image

**After:**

**Testing**
Tested on my Linux Pop!_OS ( Ubuntu ) 20.04 laptop using my "Mesa Intel(R) UHD Graphics (CML GT2)" adapter.

---

**Extra info:**

- I'm not 100% sure it's the right move to disable the alpha write to the draw framebuffer. It might disable the ability to use the GL backend for transparent windows, but I don't really know anything about transparent windows or if there are other problems/setup needed to make that work.
- I'm not sure _why_ disabling sRGB fixed the color issue, or whether or not there needs to be more work done to make that actually do what it's intented to do.

### DIFF
--- a/wgpu-hal/src/gles/egl.rs
+++ b/wgpu-hal/src/gles/egl.rs
@@ -541,11 +541,14 @@ impl crate::Instance<super::Api> for Instance {
                 egl::SINGLE_BUFFER as usize
             },
         ];
-        if inner.version >= (1, 5) {
-            // Always enable sRGB in EGL 1.5
-            attributes.push(egl::GL_COLORSPACE as usize);
-            attributes.push(egl::GL_COLORSPACE_SRGB as usize);
-        }
+        // sRGB caused color problems, making the render lighter colored than the vulkan version
+        // so we comment this out for now.
+        //
+        // if inner.version >= (1, 5) {
+        //     // Always enable sRGB in EGL 1.5
+        //     attributes.push(egl::GL_COLORSPACE as usize);
+        //     attributes.push(egl::GL_COLORSPACE_SRGB as usize);
+        // }
         attributes.push(egl::ATTRIB_NONE);
 
         let raw = if let Some(egl) = inner.egl.upcast::<egl::EGL1_5>() {
--- a/wgpu-hal/src/gles/queue.rs
+++ b/wgpu-hal/src/gles/queue.rs
@@ -533,7 +533,7 @@ impl super::Queue {
                         0,
                     );
                 }
-                gl.color_mask(true, true, true, true);
+                gl.color_mask(true, true, true, false);
                 gl.depth_mask(true);
                 gl.stencil_mask(!0);
                 gl.disable(glow::DEPTH_TEST);


wgpu-hal/src/gles/egl.rs
```diff
@@ -541,11 +541,14 @@ impl crate::Instance<super::Api> for Instance {
                 egl::SINGLE_BUFFER as usize
             },
         ];
-        if inner.version >= (1, 5) {
-            // Always enable sRGB in EGL 1.5
-            attributes.push(egl::GL_COLORSPACE as usize);
-            attributes.push(egl::GL_COLORSPACE_SRGB as usize);
-        }
+        // sRGB caused color problems, making the render lighter colored than the vulkan version
```
kvark, line 544: let's just remove this code entirely

wgpu-hal/src/gles/queue.rs
```diff
@@ -533,7 +533,7 @@ impl super::Queue {
                         0,
                     );
                 }
-                gl.color_mask(true, true, true, true);
+                gl.color_mask(true, true, true, false);
```
kvark, line 536: ok, this is certainly not the fix we want.
Isn't the shadow example outputting all alpha = 1 anyway?

This sounds like one of the following possibilities:
  - alpha compositing mode is messed up, and we produce alpha that isn't 1.0
  - driver bug

zicklag:
After investigation by @cwfitzgerald and testing summarized in my comment ( https://github.com/gfx-rs/wgpu/issues/1627#issuecomment-878539679 ) it seems like neither of these changes should actually be made and we need to find different ways around the issues.

PR #1634, opened 2021-07-12
Improve documentation of ImageDataLayout

### LINKED ISSUE #1626: Update ImageDataLayout's rows_per_image documentation to document what units are expected.
The wording currently is:
> Rows that make up a single "image". Each "image" is one layer in the z direction of a 3D image. May be larger than copy_size.y. Required if there are multiple images (i.e. the depth is more than one)

Which is a little confusing because I assumed the value should be a row count in the image. Instead it seems to be a Texel size value of how large each row is? If the documentation could be updated to better explain the units expected here it would help out a ton!

cwfitzgerald:
**Connections**

Fixes #1626.

**Description**

Just improves the docs.

**Testing**

Docs. @StarArawn take a look if that clears things up.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -2738,6 +2738,15 @@ impl<T> Default for RenderBundleDescriptor<Option<T>> {
 }
 
 /// Layout of a texture in a buffer's memory.
+///
+/// The bytes per row and rows per image can be hard to figure out so here are some examples:
+///
+/// | Resolution | Format | Bytes per block | Pixels per block | Bytes per row                          | Rows per image               |
+/// |------------|--------|-----------------|------------------|----------------------------------------|------------------------------|
+/// | 256x256    | RGBA8  | 4               | 1 * 1 * 1        | 256 * 4 = Some(1024)                   | None                         |
+/// | 32x16x8    | RGBA8  | 4               | 1 * 1 * 1        | 32 * 4 = 128 padded to 256 = Some(256) | None                         |
+/// | 256x256    | BC3    | 16              | 4 * 4 * 1        | 16 * (256 / 4) = 1024 = Some(1024)     | None                         |
+/// | 64x64x8    | BC3    | 16              | 4 * 4 * 1        | 16 * (64 / 4) = 256 = Some(256)        | 64 / 4 = 16 = Some(16)       |
 #[repr(C)]
 #[derive(Clone, Copy, Debug, Default)]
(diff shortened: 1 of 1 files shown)


wgpu-types/src/lib.rs
```diff
@@ -2738,6 +2738,26 @@ impl<T> Default for RenderBundleDescriptor<Option<T>> {
 }
 
 /// Layout of a texture in a buffer's memory.
+///
+/// The bytes per row and rows per image can be hard to figure out so here are some examples:
+///
+/// 256x256 RGBA8 image
```
kvark, line 2744: none of the examples shows that rows per image are in blocks

wgpu-types/src/lib.rs
```diff
@@ -2738,6 +2738,26 @@ impl<T> Default for RenderBundleDescriptor<Option<T>> {
...
+///  - Bytes per block = 4 
+///  - Pixels per block = 1*1*1
+///  - Bytes per row = 256 * 4 = Some(1024)
+///  - Rows per image = None
+///
+/// 32x16x8 RGBA8 image
```
kvark, line 2750: would it be possible to put these into a table?

wgpu-types/src/lib.rs
```diff
@@ -2746,18 +2766,27 @@ pub struct ImageDataLayout {
...
-    /// Rows that make up a single "image". Each "image" is one layer in the z direction of a 3D image. May be larger
-    /// than `copy_size.y`.
-    /// Required if there are multiple images (i.e. the depth is more than one)
+    /// "Rows" that make up a single "image". 
+    ///
+    /// See documentation of [`bytes_per_row`] for how a row is defined.
+    ///
+    /// An image is one layer in the z direction of a 3D image or 2DArray texture. 
+    ///
+    /// The amount of rows per image may be larger than the actual amount of rows of data.
```
kvark, line 2787: need to talk about texel blocks as well

cwfitzgerald: What do you mean by this? I defined what definition of "row" I was using a line or two above.

kvark: Oh, you mean this?
> A row is one row of pixels or of compressed blocks in the x direction.

Since it's documenting a different field, the docs do not obviously apply here. Let's be more explicit, it's ok to be redundant for this.

kvark (approved):
Thanks!
bors r+

StarArawn:
> Docs. @StarArawn take a look if that clears things up.

Looks good to me! Thank you! 👍

PR #1643, opened 2021-07-13
Check docs in CI

### LINKED ISSUE #1637: CI Needs to Check for Broken Doc-Links

We've got a bunch of broken doc links. We need to check these.

cwfitzgerald:
**Connections**

Closes #1637.

**Description**

Doc links will now fail. `cargo doc` doesn't let you pass through flags, so you have to use RUSTDOCFLAGS.

**Testing**

It is.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,7 @@ jobs:
     env:
       PKG_CONFIG_ALLOW_CROSS: 1 # allow android to work
       RUSTFLAGS: --cfg=web_sys_unstable_apis
+      RUSTDOCFLAGS: -Dwarnings
 
     steps:
       - name: checkout repo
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -128,6 +131,9 @@ jobs:
           # explicitly don't mention wgpu-hal so that --all-features don't apply to it
           cargo clippy --target ${{ matrix.target }} -p wgpu -p wgpu-core -p wgpu-info -p player --examples --tests --all-features -- -D warnings 
 
+          # build docs
+          RUSTDOCFLAGS='-Dwarnings' cargo doc --target ${{ matrix.target }} -p wgpu -p wgpu-core -p wgpu-info -p player -all-features --no-deps
```
kvark, line 135: can we ask it to build docs for all crates (instead of listing them individually)?

kvark (approved):

.github/workflows/ci.yml
cwfitzgerald: Oh, so the reason I'm doing this is because of wgpu-hal's `--all-features` not working, same reason as above. I need to also check wgpu-hal, so I'll add that.

cwfitzgerald:
bors r=kvark

PR #1644, opened 2021-07-13
Fix Example Resolution Limits

### LINKED ISSUE #1616: Examples panic with validation error in Device::create_texture() when maximized on large displays
**Description**

The examples request a device with `wgpu::Limits::downlevel_defaults()`, which IIUC requires a _maximum_ 2D texture size of `2096x2096`: https://github.com/gfx-rs/wgpu/blob/e27fbb821bb3397069d0a82ef8da8e151feb0647/wgpu-types/src/lib.rs#L615

(BTW this is a weird number, it's not a power of 2! Should it be 2048 or 4096?)

This function was added in #1599 (cc @cwfitzgerald). The error does not occur on the main tree before this PR was merged.

Many examples like `shadow` and `water` create a texture using the window dimensions, which I believe to be quite common. When the window size is larger than 2,096 pixels in either dimension, the app panics.

**Repro steps**

Run the `shadow` example and resize the window to be larger than 2,096 pixels in either dimension. It will report a validation error then panic.

Ideally, don't ever panic.

**Extra materials**

```text
$ RUST_BACKTRACE=1 cargo run --release --example shadow
    Finished release [optimized] target(s) in 0.16s
     Running `C:\Users\jay\other-projects\wgpu\target\release\examples\shadow.exe`
Using NVIDIA GeForce RTX 3090 (Vulkan)
[2021-07-08T14:08:09Z ERROR wgpu::backend::direct] wgpu error: Validation Error

    Caused by:
        In Device::create_texture
        Dimension 3440 value 3440 exceeds the limit of 2096


thread 'main' panicked at 'Handling wgpu errors as fatal by default', wgpu\src\backend\direct.rs:2015:5
stack backtrace:
   0: std::panicking::begin_panic
   1: <wgpu::backend::direct::Context as wgpu::Context>::queue_write_buffer
   2: wgpu::util::init::initialize_adapter_from_env_or_default
   3: <wgpu::backend::direct::Context as wgpu::Context>::device_create_texture
   4: wgpu::Device::create_texture
   5: <shadow::Example as shadow::framework::Example>::resize
   6: winit::platform_impl::platform::event_loop::EventLoop<T>::run_return::{{closure}}
   7: <std::panic::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
   8: winit::platform_impl::platform::event_loop::runner::EventLoopRunner<T>::register_window
   9: winit::platform_impl::platform::event_loop::runner::EventLoopRunner<T>::send_event
  10: winit::platform_impl::platform::event_loop::public_window_callback::{{closure}}
  11: winit::platform_impl::platform::event_loop::runner::EventLoopRunner<T>::catch_unwind
  12: winit::platform_impl::platform::event_loop::public_window_callback
[... snip ...]
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
error: process didn't exit successfully: `C:\Users\jay\other-projects\wgpu\target\release\examples\shadow.exe` (exit code: 101)
```

**Platform**

Issue is platform-independent.

cwfitzgerald:
**Connections**

Closes #1616.

**Description**

This change automatically expands resolution to the adapter's limits so examples always work.

**Testing**

Ran locally, but couldn't make it bigger than the screen

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -631,6 +631,20 @@ impl Limits {
             max_push_constant_size: 0,
         }
     }
+
+    /// Modify the current limits to use the resolution limits of the other.
+    ///
+    /// This is useful because the swapchain might need to be larger than any other image in the application.
+    ///
+    /// If your application only needs 512x512, you might be running on a 4k display and need extremely high resolution limits.
+    pub fn using_resolution(self, other: Self) -> Self {
+        Self {
+            max_texture_dimension_1d: other.max_texture_dimension_1d,
+            max_texture_dimension_2d: other.max_texture_dimension_2d,
+            max_texture_dimension_3d: other.max_texture_dimension_3d,
+            ..self
+        }
+    }
 }
 
 /// Represents the sets of additional limits on an adapter,
--- a/wgpu/examples/framework.rs
+++ b/wgpu/examples/framework.rs
@@ -137,7 +137,8 @@ async fn setup<E: Example>(title: &str) -> Setup {
         required_features - adapter_features
     );
 
-    let needed_limits = E::required_limits();
+    // Make sure we use the texture resolution limits from the adapter, so we can support images the size of the swapchain.
+    let needed_limits = E::required_limits().using_resolution(adapter.limits());
 
     let trace_dir = std::env::var("WGPU_TRACE");
     let (device, queue) = adapter
@@ -396,10 +397,7 @@ pub fn test<E: Example>(mut params: FrameworkRefTest) {
     assert_eq!(params.width % 64, 0, "width needs to be aligned 64");
 
     let features = E::required_features() | params.optional_features;
-    let mut limits = E::required_limits();
-    if limits == wgpu::Limits::default() {
-        limits = test_common::lowest_reasonable_limits();
-    }
+    let limits = E::required_limits();
 
     test_common::initialize_test(
         mem::take(&mut params.base_test_parameters)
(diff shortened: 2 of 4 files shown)


wgpu-types/src/lib.rs
```diff
@@ -631,6 +631,19 @@ impl Limits {
             max_push_constant_size: 0,
         }
     }
+
+    /// Modify the current limits to use the 2D resolution limits of the other.
+    ///
+    /// This is useful because the swapchain might need to be larger than any other image in the application.
+    ///
+    /// If your application only needs 512x512, you might be running on a 4k display and need extremely high resolution limits.
+    pub fn using_resolution_2d(self, other: Self) -> Self {
```
kvark, line 640: why don't we use the `default()` implementation limits here?

wgpu-types/src/lib.rs
```diff
@@ -631,6 +631,19 @@ impl Limits {
...
+        Self {
+            max_texture_dimension_1d: other.max_texture_dimension_1d,
+            max_texture_dimension_2d: other.max_texture_dimension_2d,
```
kvark, line 643: it feels rather arbitrary that we are overriding these 2 limits but not the other texture dimensions

wgpu-types/src/lib.rs
cwfitzgerald: Because if we use the default, rpi4 and such won't be able to run (as rpi has a limit of 4096x4096), but if we use 4k limits the guy with two 4k monitors will complain... so what we actually want is _just_ the adapter's limits.

wgpu-types/src/lib.rs
cwfitzgerald: I can add 3d as well, but I only did 1 and 2 because the fundamental problem is the swapchain and swapchain sized things.

kvark: let's grab all the texture dimensions for ... consistency

kvark (approved):
bors r+

PR #1647, opened 2021-07-13
Check for blendable flag in filtering

kvark:
**Connections**
Discussion on the matrix.

**Description**
The validation is written in a way that considers "filterable" to be responsible for both linear sampling and rendering.
But we didn't take blending into account when determining this filtering flag.

**Testing**
Untested


cwfitzgerald (approved):
Logic looks sound

bors r+

PR #1652, opened 2021-07-14
Switch read-only storage textures to be exclusive and behind a feature

kvark:
**Connections**
See https://github.com/gpuweb/gpuweb/pull/1864 and https://github.com/gpuweb/gpuweb/pull/1794

**Description**
WebGPU API doesn't have RO storage textures right now. However, since `wgpu` already had a way to expose read-write storage textures, I figured we can just use it to gate read-only storage.

**Testing**
No examples currently exercise this path.


cwfitzgerald (approved):
bors r+

PR #1654, opened 2021-07-14
hal: renderdoc integration

kvark:
**Connections**
Hooks up renderdoc API

**Description**

**Testing**
Tested manually on Vulkan and DX12


kvark:
Alright, this is https://github.com/rust-lang/rust-clippy/issues/7423
We don't need to wait for it. Merging.

PR #1658, opened 2021-07-14
Removes nightly from bors

cwfitzgerald:
**Connections**

Following up on the trouble #1654 had merging. We actually shouldn't force bors to wait for nightly to work.


kvark (approved):
bors r+

PR #1665, opened 2021-07-15
(macOS/iOS) readWriteTextureSupport is support on macOS 10.13+ / iOS 11.0

xiaopengli89:
```
[MTLIGAccelDevice readWriteTextureSupport]: unrecognized selector sent to instance 0x7f97d1070a00
```
https://developer.apple.com/documentation/metal/mtldevice/2887289-readwritetexturesupport?language=objc

### DIFF
--- a/wgpu-hal/src/metal/adapter.rs
+++ b/wgpu-hal/src/metal/adapter.rs
@@ -573,7 +573,17 @@ impl super::PrivateCapabilities {
                 MTLLanguageVersion::V1_0
             },
             exposed_queues: 1,
-            read_write_texture_tier: device.read_write_texture_support(),
+            read_write_texture_tier: if os_is_mac {
+                if Self::version_at_least(major, minor, 10, 13) {
+                    device.read_write_texture_support()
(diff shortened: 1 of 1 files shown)


wgpu-hal/src/metal/adapter.rs
```diff
@@ -543,6 +543,11 @@ impl super::PrivateCapabilities {
             sample_count_mask |= 8;
         }
 
+        let mut read_write_texture_tier = mtl::MTLReadWriteTextureTier::TierNone;
+        if os_is_mac && Self::version_at_least(major, minor, 10, 13) {
```
kvark, line 547: We also need to check iOS, where this is available starting with 11

kvark (commented):
Thank you for the fix! Just one more thing.
And please squash the commits (our bot doesn't know how to do it).

xiaopengli89:
I squash the commits, also add check for iOS.

kvark (approved):
thank you!
bors r+

PR #1666, opened 2021-07-15
Special mode for rebinding only dynamic offsets

kvark:
**Connections**

**Description**
I noticed that Metal backend does a lot of wasted resource rebinding on Halmark example, since the only thing that changes is the dynamic offset. So this is an attempt to handle this case more carefully across the stack.

As I was finishing this work, I realized that it's better to just ask users to only rebind stuff that truly changes. I.e. in this case the test should either assume that textures can also be different between entities, or it should move the shared textures into a lower bind group, so that the most frequent one only has our dynamic offset buffers.

**Testing**
Just the examples.

### DIFF
--- a/wgpu-core/src/command/bind.rs
+++ b/wgpu-core/src/command/bind.rs
@@ -183,6 +183,17 @@ impl Binder {
         (bind_range.start, &self.payloads[bind_range])
     }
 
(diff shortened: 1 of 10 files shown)


kvark:
For the reasons outlined in the description, I'm going to close this.

PR #1671, opened 2021-07-16
Use EGL_KHR_gl_colorspace for srgb if available

### LINKED ISSUE #1667: sRGB Conversion Not Done for EGL < 1.5
**Description**
I think sRGB converison is not currently working for the GL backend with EGL < 1.6 on Linux. The issue was originally found in #1627.

( https://github.com/gfx-rs/wgpu/issues/1627#issuecomment-878602534 )

@cwfitzgerald said:
> Yeah my haswell rig has srgb issues on both vk and gl.

We have fixed the vulkan issue, but haven't done anything about GL yet.

**Repro steps**

All of the examples appear darker in color than the vulkan ones. ( after we fixed vulkan in #1651 )

**Extra materials**

**Platform**
I was unable to reproduce this on my machine using `Mesa Intel(R) UHD Graphics (CML GT2)`, but @cwfitzgerald mentioned having issues on his haswell rig.

cwfitzgerald:
**Connections**

Fixes #1667 

**Description**

Uses the extension to provide SRGB if needed relevant.

I also promoted the log message about missing robustness to a warning because it is a safety issue.

**Testing**

Tested on rpi and haswell.

### DIFF
--- a/wgpu-hal/src/gles/egl.rs
+++ b/wgpu-hal/src/gles/egl.rs
@@ -8,6 +8,8 @@ const EGL_CONTEXT_OPENGL_DEBUG_BIT_KHR: i32 = 0x0001;
 const EGL_CONTEXT_OPENGL_ROBUST_ACCESS_EXT: i32 = 0x30BF;
 const EGL_PLATFORM_WAYLAND_KHR: u32 = 0x31D8;
 const EGL_PLATFORM_X11_KHR: u32 = 0x31D5;
+const EGL_GL_COLORSPACE_KHR: u32 = 0x309D;
+const EGL_GL_COLORSPACE_SRGB_KHR: u32 = 0x3089;
 
 type XOpenDisplayFun =
     unsafe extern "system" fn(display_name: *const raw::c_char) -> *mut raw::c_void;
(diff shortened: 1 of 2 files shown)


wgpu-hal/src/gles/egl.rs
```diff
@@ -543,11 +569,21 @@ impl crate::Instance<super::Api> for Instance {
...
+            SrgbFrameBufferKind::Core => {
+                attributes.push(egl::GL_COLORSPACE as usize);
+                attributes.push(egl::GL_COLORSPACE_SRGB as usize);
+            }
+            SrgbFrameBufferKind::Khr => {
+                attributes.push(EGL_GL_COLORSPACE_KHR as usize);
+                attributes.push(EGL_GL_COLORSPACE_SRGB_KHR as usize);
+            }
+            SrgbFrameBufferKind::None => {
+                // TODO: expose ot the user that srgb isn't supported
```
kvark, line 583: do we need this TODO? let's just implement it here in the PR

wgpu-hal/src/gles/egl.rs
```diff
@@ -543,11 +569,21 @@ impl crate::Instance<super::Api> for Instance {
...
+            }
         }
+        if inner.version >= (1, 5) {}
```
kvark, line 586: leftover?

wgpu-hal/src/gles/egl.rs
cwfitzgerald: I looked but couldn't find where I would change the list of supported backbuffer formats, so if you point me in the right direction I can do it.

wgpu-hal/src/gles/egl.rs
cwfitzgerald: Oops!

wgpu-hal/src/gles/egl.rs
kvark: Sorry, I should have just posted the link right away - https://github.com/gfx-rs/wgpu/blob/22a81de6a99b25223f2fd9d6233f10fa16908ab6/wgpu-hal/src/gles/adapter.rs#L465

kvark (approved):
bors r+

PR #1673, opened 2021-07-16
hal/vk: relay semaphore

kvark:
**Connections**
Somewhere on the Matrix, by @pythonesque

**Description**
Vulkan (somewhat confusingly) requires semaphore synchronization between `vkQueueSubmit` and `vkQueuePresent` *even if* it's done on the same queue.
This PR exercises a dumb idea of just keeping one binary semaphore as a "relay" between all queue commands, all concealed within hal/vk backend.

**Testing**
Running Halmark shows that all the things break here. I'm not seeing anything particularly wrong with the code though. My humble guess is that AMD driver (at least) isn't happy about the mix of binary and timeline semaphores used together in the same signal operation.

### DIFF
--- a/wgpu-hal/src/vulkan/mod.rs
+++ b/wgpu-hal/src/vulkan/mod.rs
@@ -240,6 +240,12 @@ pub struct Queue {
     swapchain_fn: khr::Swapchain,
     device: Arc<DeviceShared>,
     family_index: u32,
+    /// This special semaphore is used to synchronize GPU work of
+    /// everything on a queue with... itself. Yikes!
+    /// It's required by the confusing portion of the spec to be signalled
+    /// by last submission and waited by the present.
+    relay_semaphore: vk::Semaphore,
+    relay_active: bool,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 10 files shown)


kvark:
wgpu-rs examples appear to work fine with this PR, but Halmark fails miserably...
May be related to https://www.reddit.com/r/vulkan/comments/ihfg82/combining_timeline_semaphore_signals_and_binary/?ref=share&ref_source=link

kvark:
Looks like VVL-1.2.176.0 aren't complaining, and everything works.
Pinging @Ralith to peek at this before we merge.

wgpu-hal/src/vulkan/mod.rs
```diff
@@ -240,6 +240,10 @@ pub struct Queue {
     swapchain_fn: khr::Swapchain,
     device: Arc<DeviceShared>,
     family_index: u32,
+    /// This special semaphore is used to synchronize GPU work of
+    /// everything on a queue with... itself. Yes, thanks Vulkan!
```
Ralith, line 244: I think this comment could be clearer: it looks like the semaphore is only actually being used to synchronize presentation, whereas other work never waits on it in practice (due to TOP_OF_PIPE).

Ralith (approved):
Vulkan stuff looks reasonable to me. If an older validation layer produced errors but a newer one doesn't, that's probably a fixed bug.

wgpu-hal/src/vulkan/mod.rs
kvark: ok, will fix, thank you!

kvark:
bors r=Ralith

Issue #1617, opened 2021-07-08
WebGL2 backend


kvark:
**Is your feature request related to a problem? Please describe.**
The new GLES3 backend landed in #1510
Now we need to enable the Web target for WebGL2.

**Describe the solution you'd like**
On "wasm32-unknown-unknown" target the GLES backend should target WebGL2, just like the old "gfx-backend-gl" did.

**Describe alternatives you've considered**
Having a separate feature for this, may still be a good idea in order to help the `cargo-vendor` volume.

**Additional context**
Required for Bevy

zicklag:
I'm going to start working on this as I have the time. So far I got it compiling, but haven't tested it yet.

zicklag:
Got the hello-triangle example working on web. So far it's the only example running, and I may have to figure out how to calculate certain hardware limits that I am just guessing at so far, but it's a good sign. :smile:

![image](https://user-images.githubusercontent.com/25393315/125984998-555bea73-6f9c-45fd-abf1-870b6273c990.png)

kvark:
It would be very useful to get in touch with you on more immediate channels than Github. Please consider joining Matrix or Discord!

Gordon-F:
@zicklag Hi! A couple of notes, that, maybe, helps you a little bit with your journey :)

1. Bring `WebGL` related fixes, that are presented in `gfx-backend-gl`, but not present in `wgpu-hal`:
- https://github.com/gfx-rs/gfx/pull/3699
- https://github.com/gfx-rs/gfx/pull/3682
- https://github.com/gfx-rs/gfx/pull/3773
2. 2 debugging tools (idk, maybe these are the only 2 existing useful debugging tools for webgl :cry:)
- [`SpectorJS`](https://spector.babylonjs.com/)
- [`webgl-lint`](https://github.com/greggman/webgl-lint). I recommend disabling `throwOnError` first. And you can safely ignore all errors like `glAnyFunction(None)` because `wasmbindgen` send `None` as `undefined`, but `webgl-lint` required `null`.

zicklag:
Got some help on Matrix ( thanks guys! ), and with https://github.com/gfx-rs/naga/pull/1100 the shader problem I was stuck on was fixed.

After that I got to a blank screen for the cube example:

![image](https://user-images.githubusercontent.com/25393315/126040898-b43a25d8-ee18-4f75-a056-53c667abec2f.png)

It's got an invalid operation I've got to investigate, but we got past the shader problems!

[...]

PR #1686, opened 2021-07-17
Implement WebGL2 Backend

### LINKED ISSUE #1740: [Android] Error:glMapBufferRange::<access> is not an accepted value
**Description**
Error:glMapBufferRange::<access> is not an accepted value

**Repro steps**
Run `cube` on Android.

**Extra materials**
https://github.com/gfx-rs/wgpu/blob/451cd217bbf77d94130ef13816bc655341dd558f/wgpu-hal/src/gles/device.rs#L321-L326

`MAP_PERSISTENT_BIT` and `MAP_COHERENT_BIT` are part of [`ARB_buffer_storage`](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_buffer_storage.txt) or [`EXT_buffer_storage`](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_buffer_storage.txt) not [ES 3.0 spec](https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/glMapBufferRange.xhtml).

**Platform**
Android 8.0, wgpu master.

zicklag:
**Connections**
Works towards #1617 .

Closes #1740.

**Description**
Migrates the WebGL2 backend from `gfx-hal-backend` to `wgpu-hal`.

**Testing**
Tested WebGL (  Brave Browser ( Chrome fork ), and Firefox ) and GL backends on Ubuntu 20.04 in.

---

Currently only the cube, triangle, and msaa-line examples work. This is still work-in-progress, with some things still to be done, but I wanted to get the code up here, and there isn't a lot that I think will changing drastically.

image

## Remaining Work

- [x] A little script for building and serving the examples for web
- [x] Fix fastclear bug workaround
- [x] Update shadows example to not use SSBOs on WebGL
- [x] Fix lack of sRGB conversion
- [ ] Properly determining the web hardware limit values ( see `adapter.rs:183` )
- [ ] Fix webgl-lint errors
- [ ] Fix issue with water rendering on water shader


zicklag:
@cwfitzgerald Sorry I had to drop off of chat, but I've got another issue.

When Naga compiles the shadow shader for GLSL ES, it includes certain features that aren't supported such as an array without a length. While that array isn't used, because we were going to use a different entrypoint, it doesn't matter because the GLSL won't compile.

I'm not seeing a way around needing a shader preprocessor. Something like `#ifdef STORAGE` checks.

I could trivially implement one with whatever syntax we want if you don't mind pulling in 1 dependency ( not counting `quote` and `proc_macro2` ) from the [`peg`](https://docs.rs/peg) crate. The implementation would probably cost < 100 lines of code and could even be dropped reasonably in `framework.rs` for the examples.

What do you think?

kvark:
> When Naga compiles the shadow shader for GLSL ES, it includes certain features that aren't supported such as an array without a length. While that array isn't used, because we were going to use a different entrypoint, it doesn't matter because the GLSL won't compile.

Where is this failing and how? Shadow example works on GLES backend generally on native.

When Naga's GLSL backend is invoked, it's given the exact entry point. And it already knows from the IR what global variables are used by this entry point. So technically it could be smart enough to avoid emitting the definition of this structure  entirely if a different entry point is chosen.

zicklag:
> Where is this failing and how? Shadow example works on GLES backend generally on native.

It fails on WebGL when uncommenting this line in the shader:

```glsl
[[block]]
struct Lights {
    data: [[stride(96)]] array<Light>;
}
```

image

It works fine on Native OpenGL.

> So technically it could be smart enough to avoid emitting the definition of this structure entirely if a different entry point is chosen.

Ah, that would be great. I could look into doing a PR for that, then.

zicklag:
Now we've got bunnymark and shadow exeamples working. The water example is almost working, but the water is not appearing, and I can narrow the issue down to a pixel depth calculation failing. If I take out the step in the shader that uses the that depth, it appears fine minus the slight bluring around the edges of the water that isn't there:

image

Here's the portion of the shader in question:

**water.wgsl:**

```glsl
[[stage(fragment)]]
fn fs_main(in: VertexOutput) -> [[location(0)]] vec4<f32> {
    let reflection_colour = textureSample(reflection, colour_sampler, in.f_WaterScreenPos.xy).xyz;

    let pixel_depth = to_linear_depth(in.position.z);
    let normalized_coords = in.position.xy / vec2<f32>(uniforms.time_size_width.w, uniforms.viewport_height);
    let terrain_depth = to_linear_depth(textureSample(terrain_depth_tex, colour_sampler, normalized_coords).r);

    // This pixel depth here is the issue. If set to zero the water is visible
    let dist = terrain_depth - pixel_depth;
    let clamped = pow(smoothStep(0.0, 1.5, dist), 4.8);

    let final_colour = in.f_Light + reflection_colour;
    let t = smoothStep(1.0, 5.0, dist) * 0.2; //TODO: splat for mix()?
    let depth_colour = mix(final_colour, water_colour, vec3<f32>(t, t, t));

    return vec4<f32>(depth_colour, clamped * (1.0 - in.f_Fresnel));
}
```

Getting close!

cwfitzgerald:
For platforms that don't support Read-only depth stencil like OpenGL/WebGL (we should have a downlevel limit for this) we can do an explicit copy-texture-to-texture to another depth buffer so we don't get a cycle of textures.

kvark:
In addition to @cwfitzgerald , this appears to be UB according to WebGL spec. I see that it works at least on some OpenGL ES devices, so I thought we may get away without another downlevel flag. Looks like WebGL is more strict about this, although I'd expect to see an actual error.

wgpu-types/src/lib.rs
```diff
@@ -702,37 +718,51 @@ impl DownlevelCapabilities {
...
+    ///
+    /// ```
+    /// # let my_flags = DownlevelFlags::all();
+    /// # use wgpu_types::DownlevelFlags;
+    /// assert_eq!(my_flags & DownlevelFlags::compliant(), DownlevelFlags::compliant());
+    /// ```
     pub struct DownlevelFlags: u32 {
         /// The device supports compiling and using compute shaders.
-        const COMPUTE_SHADERS = 0x0000_0001;
+        const COMPUTE_SHADERS = 1 << 0;
```
kvark, line 736: Let's not change more than necessary in this PR. It shouldn't affect wgpu-types if possible, we can discuss these changes separately.
Also, it would be great to land smaller patches gradually. This PR is growing to be big, and already colliding with other incoming changes like #1717

wgpu/examples/framework.rs
```diff
@@ -1,6 +1,7 @@
 use std::future::Future;
 #[cfg(not(target_arch = "wasm32"))]
 use std::time::{Duration, Instant};
+use wgpu::DownlevelFlags;
```
kvark, line 4: the rest of the code here refers to `wgpu` namespace explicitly, let's have this following

wgpu/examples/shadow/main.rs
```diff
@@ -540,18 +567,31 @@ impl framework::Example for Example {
...
+                                ty: wgpu::BindingType::Buffer {
+                                    ty: wgpu::BufferBindingType::Storage { read_only: true },
+                                    has_dynamic_offset: false,
+                                    min_binding_size: wgpu::BufferSize::new(light_uniform_size),
+                                },
+                                count: None,
+                            }
+                        } else {
+                            wgpu::BindGroupLayoutEntry {
+                                binding: 2, // lights
```
kvark, line 583: we shouldn't use a separate binding here. Just re-use `1`

wgpu/examples/shadow/main.rs
```diff
@@ -597,16 +637,23 @@ impl framework::Example for Example {
...
-                        binding: 1,
-                        resource: light_storage_buf.as_entire_binding(),
+                    if supports_storage_resources {
+                        wgpu::BindGroupEntry {
+                            binding: 1,
+                            resource: light_storage_buf.as_entire_binding(),
+                        }
+                    } else {
+                        wgpu::BindGroupEntry {
+                            binding: 2,
```
kvark, line 647: same here

wgpu-hal/src/gles/queue.rs
```diff
@@ -25,17 +27,112 @@ fn is_3d_target(target: super::BindTarget) -> bool {
...
+
 impl super::Queue {
-    unsafe fn reset_state(&self) {
+    // Add helper functions for enabling and disabling certain GL features so that we can keep track
+    // of the current enabled/disabled state
+    unsafe fn set_current_program(&self, program: Option<glow::Program>) {
+        self.shared.context.use_program(program);
+        self.state.lock().unwrap().current_program = program;
+    }
+    gl_modes_set_enabled_fn!(
```
kvark, line 51: Are these changes necessary for the purpose of the PR? I think they aren't necessary.
Let's please focus on the goal here, otherwise it's much more difficult to land.

kvark (changes requested):

wgpu-types/src/lib.rs
kvark: Sorry, I just realized #1717 is by you, and it's doing precisely what I'm asking about - moving things out of the giant PR and landing them separately. Thank you for doing this!

zicklag: No problem. :) I do have a habit of shoving a bunch of stuff into one PR/commit.

There probably is still quite a bit of splitting up that I could do so I'll be trying to clean this up a bit and opening separate PRs when possible.

zicklag:
Still don't bother reviewing because this is all work-in-progress, but I got sRGB conversion working!

image

image

wgpu-hal/src/gles/adapter.rs
```diff
@@ -293,13 +339,39 @@ impl super::Adapter {
...
+                "Detected skylake derivative running on mesa i915. Clears to srgb textures will \
+                use manual shader clears."
+            );
+            workarounds.set(super::Workarounds::MESA_I915_SRGB_SHADER_CLEAR, true);
+        }
+
+        #[cfg(not(target_arch = "wasm32"))]
+        {
+            private_caps.set(super::PrivateCapabilities::INDEX_BUFFER_ROLE_CHANGE, true);
+            private_caps.set(super::PrivateCapabilities::BUFFER_STORAGE, true);
```
Gordon-F, line 361: I guess we need move `gfx-hal` buffer storage support check here:
https://github.com/gfx-rs/gfx/blob/3ee1ca9ba486b166a52765024d8d149cbb28d486/src/backend/gl/src/info.rs#L765-L769

cwfitzgerald:
👀

zicklag:
Whoops! That was an accident. I actually just pushed an update. :)

wgpu-hal/src/gles/adapter.rs
zicklag: Updated. :+1:

zicklag:
Just got CI working for the first time on this branch! I just fnished kind of pruning the PR so it only includes what it needs, and unfortunately its still massive. :/ But I hope we're getting to the end of it and I don't think the remaining tasks will make it much bigger.

Tomorrow I'll probably try to get on chat and get some help with the direction on the remaining issues.

agausmann:
I'm trying to run the `hello-triangle` example using your `run-wasm-example.sh`, but I keep encountering the same error. Happens on both Chrome and Firefox, and across multiple devices & OSes (Windows & Linux).

```
Error in Adapter::request_device: Limit 'max_vertex_buffer_array_stride' value 2048 is better than allowed 255
```

zicklag:
Ah, that's my bad, this line needs to be changed:

```rust
                limits: wgpu::Limits::downlevel_defaults().using_resolution(adapter.limits()),
```

It should be `wgpu::Limits::downlevel_webgl2_defaults()` to make it specify limits that will be compatible with the browser. I'll fix that probably tomorrow, but if you want to test it you just have to tweak that line.

agausmann:
Yep, I had manually changed the array stride and storage buffer/texture values in `downlevel_defaults`, which is basically the same as what your function does. Thanks

anlumo:
Anybody else succeeded in using this fork with an external project (not an example)? I had to disable

https://github.com/gfx-rs/wgpu/blob/39e8921fe8ebda314edf5c4787cdbb82281a50de/wgpu-core/Cargo.toml#L53-L61

to get it to compile. Otherwise, wgpu-hal had *all* backends enabled: metal, dx12, vulkan, etc. They don't build for wasm for obvious reasons. As far as I can tell, these lines should be disabled for the `wasm32` target anyways, so I don't know what's going on there.

agausmann:
@anlumo do you have [`resolver = "2"` set in `Cargo.toml`](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions)? That became a requirement recently, since [wgpu 0.10](https://github.com/gfx-rs/wgpu/blob/master/CHANGELOG.md#v010-2021-08-18)

anlumo:
> @anlumo do you have [`resolver = "2"` set in `Cargo.toml`](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions)?

That fixed it, thank you!

wgpu-hal/src/gles/adapter.rs
```diff
@@ -184,33 +184,50 @@ impl super::Adapter {
...
             let (sl_major, sl_minor) = Self::parse_version(&sl_version).ok()?;
             let value = sl_major as u16 * 100 + sl_minor as u16 * 10;
             naga::back::glsl::Version::Embedded(value)
         };
 
-        let vertex_shader_storage_blocks =
-            gl.get_parameter_i32(glow::MAX_VERTEX_SHADER_STORAGE_BLOCKS) as u32;
-        let fragment_shader_storage_blocks =
-            gl.get_parameter_i32(glow::MAX_FRAGMENT_SHADER_STORAGE_BLOCKS) as u32;
+        let vertex_shader_storage_blocks = if cfg!(target_arch = "wasm32") {
```
kvark, line 201: this may need to be rewritten after rebase

wgpu-hal/src/gles/adapter.rs
```diff
@@ -228,6 +245,19 @@ impl super::Adapter {
             vertex_shader_storage_textures.min(fragment_shader_storage_textures)
         };
 
+        let max_vertex_buffer_array_stride = if cfg!(target_arch = "wasm32") {
+            255 // https://www.khronos.org/registry/webgl/specs/latest/1.0/index.html#VERTEX_STRIDE
+        } else {
+            gl.get_parameter_i32(glow::MAX_VERTEX_ATTRIB_STRIDE) as u32
+        };
+
+        let max_vertex_buffers = if cfg!(target_arch = "wasm32") {
```
kvark, line 254: same here, I think the latest master already gets this right

wgpu-hal/src/gles/adapter.rs
```diff
@@ -261,6 +291,14 @@ impl super::Adapter {
             ver >= (3, 2) || extensions.contains("GL_EXT_draw_buffers_indexed"),
         );
 
+        if cfg!(not(target_arch = "wasm32"))
```
kvark, line 294: I think we already have downlevel flags covering this

wgpu-hal/src/gles/adapter.rs
```diff
@@ -318,6 +358,12 @@ impl super::Adapter {
             ver >= (3, 1),
         );
 
+        #[cfg(not(target_arch = "wasm32"))]
+        private_caps.set(super::PrivateCapabilities::INDEX_BUFFER_ROLE_CHANGE, true);
+
+        #[cfg(target_arch = "wasm32")]
+        private_caps.set(super::PrivateCapabilities::EMULATE_BUFFER_MAP, true);
```
kvark, line 365: This is not a capability. It either needs to be reversed (e.g. `NATIVE_MAPPING`), or moved to `workarounds`

wgpu-hal/src/gles/command.rs
```diff
@@ -251,14 +251,12 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
     ) where
         T: Iterator<Item = crate::BufferCopy>,
     {
-        //TODO: preserve `src.target` and `dst.target`
-        // at least for the buffers that require it.
         for copy in regions {
             self.cmd_buffer.commands.push(C::CopyBufferToBuffer {
                 src: src.raw,
-                src_target: glow::COPY_READ_BUFFER,
+                src_target: src.target,
```
kvark, line 257: This is problematic. If I create 2 vertex buffers and copy between them, they'll have the same target

wgpu-hal/src/gles/device.rs
```diff
@@ -87,6 +87,7 @@ impl super::Device {
         };
 
         let raw = gl.create_shader(target).unwrap();
+        #[cfg(not(target_arch = "wasm32"))]
         if gl.supports_debug() {
```
kvark, line 91: is this method not exposed to wasm at all?

wgpu-hal/src/gles/device.rs
```diff
@@ -391,7 +443,26 @@ impl crate::Device<super::Api> for super::Device {
     unsafe fn unmap_buffer(&self, buffer: &super::Buffer) -> Result<(), crate::DeviceError> {
         let gl = &self.shared.context.lock();
         gl.bind_buffer(buffer.target, Some(buffer.raw));
-        gl.unmap_buffer(buffer.target);
+
+        if self
```
kvark, line 447: may just check `emulate_map_allocation.take()` and go from there instead of checking the private caps

wgpu-hal/src/gles/device.rs
```diff
@@ -402,11 +473,35 @@ impl crate::Device<super::Api> for super::Device {
         let gl = &self.shared.context.lock();
         gl.bind_buffer(buffer.target, Some(buffer.raw));
         for range in ranges {
-            gl.flush_mapped_buffer_range(
-                buffer.target,
-                range.start as i32,
-                (range.end - range.start) as i32,
-            );
+            if self
```
kvark, line 476: similarly, let's just check `emulate_map_allocation` since we have to check it anyway

wgpu-hal/src/gles/mod.rs
```diff
@@ -201,8 +213,16 @@ pub struct Buffer {
     target: BindTarget,
     size: wgt::BufferAddress,
     map_flags: u32,
+    #[cfg(target_arch = "wasm32")]
+    emulate_map_allocation: std::cell::Cell<Option<*mut u8>>,
```
kvark, line 217: if the buffer is mappable, why not just create this as `Vec<>` at start and keep around?

wgpu-hal/src/gles/mod.rs
```diff
@@ -217,7 +237,14 @@ enum TextureInner {
 impl TextureInner {
     fn as_native(&self) -> (glow::Texture, BindTarget) {
         match *self {
-            Self::Renderbuffer { raw, .. } => panic!("Unexpected renderbuffer {}", raw),
+            #[cfg(not(target_arch = "wasm32"))] // Wasm renderbuffers don't impl Debug
```
kvark, line 240: let's not split this just because of `Debug`. It's a panic anyway, no need to increase code complexity for it

wgpu-hal/src/gles/mod.rs
```diff
@@ -659,7 +698,7 @@ pub struct CommandBuffer {
     label: Option<String>,
     commands: Vec<Command>,
     data_bytes: Vec<u8>,
-    data_words: Vec<u32>,
+    data_words: Vec<glow::Query>,
```
kvark, line 701: I think this is already addressed in master as well

wgpu-hal/src/gles/queue.rs
```diff
@@ -214,21 +217,58 @@ impl super::Queue {
...
+                    && dst_target == glow::ELEMENT_ARRAY_BUFFER;
 
-                gl.copy_buffer_sub_data(
-                    src_target,
-                    dst_target,
-                    copy.src_offset as i32,
-                    copy.dst_offset as i32,
-                    copy.size.get() as i32,
-                );
+                let copy_src_target = glow::COPY_READ_BUFFER;
```
kvark, line 231: what if we are copying *from* an index buffer. I think `COPY_READ_BUFFER` would not be allowed there too

wgpu-hal/src/gles/queue.rs
```diff
@@ -214,21 +217,58 @@ impl super::Queue {
...
+                } else {
+                    glow::COPY_WRITE_BUFFER
+                };
+
+                gl.bind_buffer(copy_src_target, Some(src));
+                gl.bind_buffer(copy_dst_target, Some(dst));
+
+                if is_index_buffer_only_element_dst {
+                    let mut buffer_data = vec![0; copy.size.get() as usize];
+                    gl.get_buffer_sub_data(
```
kvark, line 245: is this necessary? copying buffers should be GPU-local operation, but this code reads the data to CPU

wgpu-hal/src/gles/web.rs
```diff
@@ -0,0 +1,296 @@
...
+        })
+    }
+
+    unsafe fn enumerate_adapters(&self) -> Vec<crate::ExposedAdapter<super::Api>> {
+        let canvas_guard = self.canvas.lock();
+        let gl = match *canvas_guard {
+            Some(ref canvas) => {
+                let window = web_sys::window().expect("Could not get browser window");
+
+                if canvas.get_attribute("width").is_none() {
```
kvark, line 44: this seems like it should be done at `Surface::configure()` time rather than here

wgpu/Cargo.toml
```diff
@@ -42,6 +42,11 @@ optional = true
 package = "wgpu-types"
 path = "../wgpu-types"
 
+[target.'cfg(target_arch = "wasm32")'.dependencies.hal]
```
kvark, line 45: this shouldn't be required. If there is anything wgpu needs from hal, it needs to be re-exported by wgpu-core

kvark (changes requested):

zicklag:
Thanks for the review @kvark! I'm going to try to get to addressing those soon. For now I've done the rebase, and while the web examples don't run yet after the rebase, they do compile, and I'll be working on getting them back working as soon as I can.

Also, eventually here I'll try to get on chat to work through the last of the unresolved WebGL issues, but I'm not sure if those are 100% necessary before merging this PR.

wgpu-hal/src/gles/web.rs
```diff
@@ -0,0 +1,296 @@
...
+            gl.delete_texture(renderbuffer);
+        }
+    }
+
+    unsafe fn acquire_texture(
+        &mut self,
+        _timeout_ms: u32,
+    ) -> Result<Option<crate::AcquiredSurfaceTexture<super::Api>>, crate::SurfaceError> {
+        let sc = self.swapchain.as_ref().unwrap();
+        let texture = super::Texture {
```
MTRNord, line 279: Hi not sure if I used the PR incorrect but with the code until https://sotrh.github.io/learn-wgpu/beginner/tutorial3-pipeline/#what-s-a-pipeline I get this compiler error for this line of code:

```
error[E0063]: missing field `copy_size` in initializer of `gles::Texture`
   --> C:\Users\marce\.cargo\git\checkouts\wgpu-56857823b01a1582\9c69a96\wgpu-hal\src\gles\web.rs:279:23
    |
279 |         let texture = super::Texture {
    |                       ^^^^^^^^^^^^^^ missing `copy_size`

For more information about this error, try `rustc --explain E0063`.
error: could not compile `wgpu-hal` due to previous error
```

zicklag: After my most recent push this branch isn't currently working. I've still got to fix issues that appeared after rebasing on top of the latest master. I'll try to get to fixing this soon, but I've been pretty busy.

wgpu-hal/src/gles/device.rs
zicklag: The `supports_debug()` function technically works and always returns `false`, as it should, but the issue is actually the `mem::transmut(raw)` won't compile on WASM because the `raw` type is something like `WebShaderKey`, which can't be transmuted to a `u32` because of the size difference.

wgpu-hal/src/gles/queue.rs
zicklag: I'm not certain. I just grabbed the code from gfx. From the comment above I'm guessing it's copying to the CPU because it's not allowed to copy unmatching buffer types without copying to the CPU first. :man_shrugging: Your guess is better than mine.

wgpu-hal/src/gles/command.rs
zicklag: :thinking: Do you think this will work? Not positive I'm understanding this right.

```rust
       let (src_target, dst_target) = if src.target == dst.target {
            (glow::COPY_READ_BUFFER, glow::COPY_WRITE_BUFFER)
        } else {
            (src.target, dst.target)
        };
        for copy in regions {
            self.cmd_buffer.commands.push(C::CopyBufferToBuffer {
                src: src.raw,
                src_target,
                dst: dst.raw,
                dst_target,
                copy,
            })
        }
```

The examples work on web with that, and I think it would fix your 2 vertex buffer problem, but I'm not sure if it introduces other problems.

wgpu-hal/src/gles/web.rs
zicklag: Oh yeah, that canvas is actually purely for enumerating adapters and doesn't actually even need a size, so we can just take that whole section out. Also, it appears that the canvas's size is already properly set by winit or something so I don't think we need to do it in `Surface::configure()` either.

zicklag:
OK, I think I've addressed @kvarks review points so far, and the rest of the web demos are working again! Now I've got to fix CI and the fact that somehow I broke the cube and shadow examples on desktop GL. :frowning:

zicklag:
OK, everything's working again!

There are still WebGL lint errors and the water shader to fix, but I think we can open issues for those and do that in separate PRs.

cwfitzgerald:
So exciting! Going to give this a onceover myself.

Cargo.lock
```diff
@@ -0,0 +1,1979 @@
+# This file is automatically @generated by Cargo.
```
cwfitzgerald, line 1: We removed the Cargo.lock

.github/workflows/ci.yml
```diff
@@ -130,6 +130,9 @@ jobs:
         run: |
           cargo clippy --target ${{ matrix.target }} -p wgpu
 
+          # Build for WebGPU+WebGL
```
cwfitzgerald, line 133: nit, this builds for _just_ webgl, the above command builds for webgpu

wgpu-core/Cargo.toml
```diff
@@ -19,6 +19,7 @@ trace = ["ron", "serde", "wgt/trace", "arrayvec/serde", "naga/serialize"]
 replay = ["serde", "wgt/replay", "arrayvec/serde", "naga/deserialize"]
 # Enable serializable compute/render passes, and bundle encoders.
 serial-pass = ["serde", "wgt/serde", "arrayvec/serde"]
+webgl = ["hal/gles"]
```
cwfitzgerald, line 22: Why do we need this at all, can't we just assume gl is on wasm all the time as wgc isn't built on a webgpu build?

anlumo:
@zicklag Great! Is there any chance that you might be able to take on #1888 as well? It's very much related to this PR, but I suspect that it crosses more layers of wgpu, unfortunately.

zicklag:
I don't want to commit to anything extra right now because I have a lot of stuff on my plate at the moment, but if it happens to be easy and along the way when I get back to working on the remaining WebGL issues I might take a look at it.

wgpu/tests/common/mod.rs
```diff
@@ -178,8 +170,11 @@ impl TestParameters {
         self
     }
 }
-
-pub fn initialize_test(parameters: TestParameters, test_function: impl FnOnce(TestingContext)) {
+pub fn initialize_test(
+    parameters: TestParameters,
+    limits_function: impl FnOnce(&Adapter) -> wgpu::Limits,
```
cwfitzgerald, line 175: Why is this needed, the test should require the very minimum limits it requires and behave accordingly. If there are multiple "modes" to test, there should be multiple tests.

cwfitzgerald (commented):
Some small stuff, didn't look at the higher level things.

zicklag:
Whoops, sorry @cwfitzgerald I accidentally re-requested your review, but it isn't working yet. :/

zicklag:
OK, fixed it and addressed the latest reviews.

wgpu/examples/shadow/shader.wgsl
```diff
@@ -54,11 +54,19 @@ struct Lights {
...
+};
+
 [[group(0), binding(1)]]
 var<storage, read> s_lights: Lights;
-[[group(0), binding(2)]]
-var t_shadow: texture_depth_2d_array;
+[[group(0), binding(1)]]
+var<uniform> u_lights: LightsWithoutStorage;
 [[group(0), binding(3)]]
+var t_shadow: texture_depth_2d_array;
```
kvark, line 68: why is this moved to `binding(3)`?

wgpu/examples/shadow/main.rs
```diff
@@ -213,12 +213,36 @@ impl framework::Example for Example {
         wgpu::Features::DEPTH_CLAMPING
     }
 
+    fn required_limits(adapter: &wgpu::Adapter) -> wgpu::Limits {
```
kvark, line 216: I don't think this function should need an adapter.
According to the spec, you can ask for lower than baseline limits, and you'll still get the baseline. So this function should just ask for 0 storage buffers unconditionally. If there happen to be storage buffers supported, it can discover this by asking for device limits.

kvark (commented):
Happy to see this shaping up!
A couple of last notes from me :)

wgpu/examples/shadow/main.rs
kvark: So today, I think it should just check for `wasm` target. If it's wasm, request 0 storage buffers.

mrk-its:
@zicklag could you take a look on https://github.com/katharostech/wgpu/pull/1/commits/62edf95286bd9f56b7b8ca6dca97f1491eb35589 ? Especially unmap_buffer change. Some time ago I had to apply these changes to be able to run some bevy examples.

wgpu/examples/shadow/main.rs
zicklag: According to the WGPU docs on the `Limits` struct:

> Once a device is requested, you may only use resources up to the limits requested even if the adapter supports “better” limits.

So that would mean that if we request a limit of 0, we can't use 1 storage buffer, even if the device supports it.

Does that mean we should just leave it how it is?

kvark: It's a bit more nuanced. WebGPU spec requires implementations to return limits not worse than the baseline.
So if you request 0, it will still have to return the baseline limit.

kvark:
@zicklag it looks like the PR is mostly ready. We are preparing 0.11 release and wondering if we could/should include this PR.
What is the status? I see a question from @mrk-its which is seemingly still open.

kvark:
Filed #2031 to track the last unresolved thing

zicklag:
Perfect. I also opened #2032 to track the water example issue. Do we want a separate issue for WebGL lint issues?

allsey87:
@zicklag this doesn't support [`OffscreenCanvas`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.OffscreenCanvas.html), right? Is there already an issue tracking that?

anlumo:
> @zicklag this doesn't support [`OffscreenCanvas`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.OffscreenCanvas.html), right? Is there already an issue tracking that?

#1837

haraldreingruber:
@zicklag @anlumo We might consider working on a PR for supporting WebGL2 OffscreenCanvas.

Do you have an idea how difficult this might be or how to get started?
I would assume the internals of wgpu-hal/gles don't need to be touched as it is only the WebGL2Context that has to be created from a different source. 
I guess the main challenge is connecting the wgpu/backend/web.rs `instance_create_surface_from_offscreen_canvas` fn with the GLES backend?

zicklag:
It might have already been implemented here: https://github.com/gfx-rs/wgpu/pull/1932.

haraldreingruber:
> It might have already been implemented here: #1932.

Okay, I thought it might have been disabled here: https://github.com/gfx-rs/wgpu/blob/df50028e49e1448b56ce71f7904921a95c15594c/wgpu/src/lib.rs#L1550

Will give it a try 👍

zicklag:
Oh, I'm not sure then. You might be right.

I don't remember what the state was in, and I haven't been following progress so I'm not sure where it's at now, but my gut reaction is that your initial comment above seemed correct about how you might go about it, and that it probably wouldn't be super difficult to get working.

haraldreingruber-dedalus:
> I guess the main challenge is connecting the wgpu/backend/web.rs `instance_create_surface_from_offscreen_canvas` fn with the GLES backend?

It looks like this shouldn't be very difficult, but one challenge is that the WebGPU function `instance_create_surface_from_offscreen_canvas` is exposed directly, and is not part of the HAL API. Which makes sense, because besides WebGL it's the only WASM backend.
It is not yet clear to me, how to expose a similar function in an elegant manner from GLES. 
I think I would duplicate this function in [lib.rs](https://github.com/gfx-rs/wgpu/blob/988990c24d0933c82cdd661708a3f5bcb4533ea3/wgpu/src/lib.rs#L1550-L1561) with `cfg(feature = webgl)`.

Is it possible to check if the current backend is GLES and directly access a function that is not part of the HAL API?

Any ideas welcome 😊

anlumo:
> Does WebGL have a separate concept of an "offscreen canvas" or is it just a canvas with the CSS set to `display: none`?

The type definition is:

```c
typedef ([OffscreenCanvasRenderingContext2D](https://html.spec.whatwg.org/multipage/canvas.html#offscreencanvasrenderingcontext2d) or [ImageBitmapRenderingContext](https://html.spec.whatwg.org/multipage/canvas.html#imagebitmaprenderingcontext) or [WebGLRenderingContext](https://www.khronos.org/registry/webgl/specs/latest/1.0/#WebGLRenderingContext) or [WebGL2RenderingContext](https://www.khronos.org/registry/webgl/specs/latest/2.0/#WebGL2RenderingContext) or [GPUCanvasContext](https://gpuweb.github.io/gpuweb/#canvas-context)) OffscreenRenderingContext;
```

So the type itself is the same, but the biggest difference is that OffscreenCanvas is available in web workers. You don't have a DOM in web workers, so you can't create a regular canvas there.

haraldreingruber-dedalus:
Thanks a lot @zicklag for the great outline how start with the OffscreenCanvas feature. 🙏
Let's continue the discussion in #2603 

> ( Does WebGL have a separate concept of an "offscreen canvas" or is it just a canvas with the CSS set to `display: none`? )

Yes, it's separate and is not part of the HTML DOM.
It basically only exists in the JS/WASM code, but after rendering an ImageBitmap can be transferred to a canvas.
We use it as a workaround to share resources between multiple canvases, which is not possible for WebGL.

anlumo:
> It basically only exists in the JS/WASM code, but after rendering an ImageBitmap can be transferred to a canvas.

There's also [transferControlToOffscreen](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/transferControlToOffscreen), which allows you to move the renderer to a web worker, but still render directly to the screen. This avoids blocking the UI thread.

haraldreingruber:
Cool, thanks for sharing this approach @anlumo!

PR #1688, opened 2021-07-18
[draft] Zero init textures

Wumpf:
Early preview of texture zero init implementation.

Todo:
* [ ] (hard) implement discarding textures
* [ ] (hard) need much better testing strategy
  * may need to fill textures/buffers with garbage, do operations, disable/circumvent auto init to check result without sideffects
  * many cornercases. reduce? Split out more unit-testable code!
* [ ] init swapchain textures
* [ ] range collapsing across mips & layers
* [ ] insert extra clear for "divergent, discarded stencil" after RenderPass (see comment in `wgpu-core/src/command/render.rs` for how I decided to deal with depth & stencil init state divergence)
* [ ] Implement `clear_texture` for all backends
  * [ ] (Vulkan/all) Figure out how to deal with different texture usages (e.g. can't set copy target on D24 [...])
  * [ ] DX12
  * [ ] Metal
  * [ ] Gles


player/tests/data/all.ron
```diff
@@ -4,8 +4,11 @@
 		"bind-group.ron",
 		"buffer-copy.ron",
 		"clear-buffer-image.ron",
-		"buffer-zero-init.ron",
 		"pipeline-statistics-query.ron",
 		"quad.ron",
+		"zero-init-buffer.ron",
```
kvark, line 9: can we combined these into one test?

Issue #1610, opened 2021-07-07
Reduce Example Dep Set


cwfitzgerald:
**Is your feature request related to a problem? Please describe.**

Wgpu's examples have a lot of deps, which can likely be culled down.

**Additional context**

When testing on weak systems (like the rpi) building the examples an be a 30 minute affair. While a lot of this is needed, we definitely can improve this.

We might also look at the wgpu non-dev dep tree to see if there are any gains to be had.

Craig-Macomber:
I took a quick look at this. I don't have a workflow for measuring the impact, but I made a PR that should help a little: #1692

cwfitzgerald:
Not really a priority right now.

PR #1692, opened 2021-07-19
Avoid dev-dep on image

Craig-Macomber:
**Connections**

#1610

**Description**

This removes image from the dev-deps, reducing build time/size.

Looks like noise was pulling in image with all formats enabled:
├── noise v0.7.0
│   ├── image v0.23.14

This was fixed in https://github.com/Razaekel/noise-rs/commit/1af9e1522236b2c584fb9a02150c9c67a5e6bb04#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542 which has not been released yet.

For now, we can workaround this with:
noise = { version = "0.7", default-features = false }

**Testing**

Ran `cargo build` as well as `cargo run --example water`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,12 +9,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d9fe5e32de01730eb1f6b7f5b51c17e03e2325bf40a74f754f04f130043affff"
 
 [[package]]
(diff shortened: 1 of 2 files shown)


cwfitzgerald (approved):
-212 on our cargo.lock, love to see it

bors r+

PR #1701, opened 2021-07-21
hal/vk: use actual view usage for imageless framebuffers

### LINKED ISSUE #1700: Vulkan validation errors when rendering to a texture with `COPY_SRC` enabled
**Description**
Every time a texture view created from a texture where `usage` includes `COPY_SRC | RENDER_ATTACHMENT` is attached to a render pass, the validation layer throws an error:
```
Validation Error: [ VUID-VkRenderPassBeginInfo-framebuffer-04627 ] Object 0: handle = 0xd572c4000000000b, type = VK_OBJECT_TYPE_RENDER_PASS; | MessageID = 0x45125641 | vkCmdBeginRenderPass(): Image view #0 created from an image with usage set as 0x11 but using VkImageViewUsageCreateInfo the inherited usage is the subset 0x10 and the image info #0 used to create the framebuffer had usage set as 0x11
```
This was produced when running `examples/capture`.  The texture is created with only those two usages.  `0x10` = `VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT` and `0x11` = `0x10 | VK_IMAGE_USAGE_TRANSFER_SRC_BIT`.  So the `COPY_SRC` usage is somehow lost when beginning the render pass.  The example itself produces the expected output.  In my application, I additionally use `SAMPLED`, which causes the reported bits to be `0x14` and `0x15` respectively (due to `0x4` = `VK_IMAGE_USAGE_SAMPLED_BIT`).

**Repro steps**
Running `examples/capture` is enough on my system.

**Platform**
Windows 10
`wgpu` on latest `master`
RX 590 on Radeon Software Adrenalin 21.5.1
Vulkan SDK 1.2.176.1

kvark:
**Connections**
Fixes #1700

**Description**
Vulkan spec is confusing. See [attachment info](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkFramebufferAttachmentImageInfo.html):
> usage is a bitmask of VkImageUsageFlagBits, matching the value of VkImageCreateInfo::usage used to create an image used with this framebuffer.

It doesn't talk about interaction with `ImageViewUsageCreateInfo`, which allows you to constrain the usage on a view.

**Testing**
Unable to test right now. @initial-algebra could you give it a spin?

### DIFF
--- a/wgpu-hal/src/vulkan/device.rs
+++ b/wgpu-hal/src/vulkan/device.rs
@@ -183,7 +183,7 @@ impl super::DeviceShared {
                     .enumerate()
                     .map(|(i, at)| {
                         vk::FramebufferAttachmentImageInfo::builder()
-                            .usage(conv::map_texture_usage(at.texture_usage))
+                            .usage(conv::map_texture_usage(at.view_usage))
                             .flags(at.raw_image_flags)
                             .width(e.key().extent.width)
                             .height(e.key().extent.height)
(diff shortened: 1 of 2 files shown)


initial-algebra:
Yep, checked out your branch and there's no more validation errors when running `examples/capture`.

Issue #967, opened 2020-10-08
Missing TextureFormats


crowlKats:
Was going through the spec and noticed some formats that are missing from wgpu:
`rgb9e5ufloat`
`stencil8`
`depth16unorm`
`depth24unorm-stencil8`
`depth32float-stencil8`

also in the spec `Bc6hRgbSfloat` is called `bc6h-rgb-float` (without the `s`). Is this intentional?

crowlKats:
> > also in the spec Bc6hRgbSfloat is called bc6h-rgb-float (without the s). Is this intentional?
> 
> Please file and issue to https://github.com/gpuweb/gpuweb about this!

https://github.com/gpuweb/gpuweb/issues/1141, so apparently it is intentional

crowlKats:
> `depth16unorm` needs to be added (want to help?).

I would, but I have very little knowledge about anything graphics related, so I cant really help there :/

DasEtwas:
I'm interested in having depth32float-stencil8. It enables using super high-res reversed z projection depth buffers with stencils, for e.g. deferred lighting in big scenes.

cwfitzgerald:
Any and all help appreciated :)

teoxoy:
Current status:
- `rgb9e5ufloat` (#1703)
- `stencil8` (track #3196)
- `depth16unorm` (#2954)
- `depth24unorm-stencil8` (added in #2689, removed in #2954)
- `depth32float-stencil8` (#2664)

Will close this issue since the only texture format remaining has its own issue now.

PR #955, opened 2020-10-04
Add Stencil8 and Rgb9e5uFloat texture formats

lachlansneff:
**Connections**
[matrix](https://matrix.to/#/!XFRnMvAfptAHthwBCx:matrix.org/$9fBtjkRSMQO_CJkfNin9I8rvUDRZ1gzsUgAoIygpiN4?via=matrix.org&via=kyju.org&via=jboi.nl)

**Description**
This adds Stencil8 and Rgb9e5uFloat texture formats.

**Testing**
No testing required, afaik.

PR #1703, opened 2021-07-22
Add Rgb9e5Ufloat Format

cwfitzgerald:
**Connections**

Addresses part of #967, a wgpu-hal followup to #955.

**Description**

This adds only `Rgb9e5Ufloat`. I removed the explicit numbering from the texture format so we don't have to constantly re-number things. The default numbering is identical to what we were doing.

**Testing**

All backends unconditionally support it, so little testing is needed.


kvark (approved):
Looks great, thank you!
bors r+

PR #1717, opened 2021-07-23
Work Around Fastclear Bug for Web and Native GL

### LINKED ISSUE #1627: Fastclear Bug With Intel Mesa Adapters on the GL Backend

zicklag:
**Connections**
Resolves #1627 

**Description**
Works around Mesa fastclear bug by doing a manual shader clear on effected platforms

**Testing**
Tested on Mesa Intel(R) UHD Graphics (CML GT2) (Gl)

### DIFF
--- a/wgpu-hal/src/gles/adapter.rs
+++ b/wgpu-hal/src/gles/adapter.rs
@@ -3,6 +3,9 @@ use std::sync::Arc;
 
 // https://webgl2fundamentals.org/webgl/lessons/webgl-data-textures.html
 
+const GL_UNMASKED_VENDOR_WEBGL: u32 = 0x9245;
+const GL_UNMASKED_RENDERER_WEBGL: u32 = 0x9246;
+
 impl super::Adapter {
     /// According to the OpenGL specification, the version information is
     /// expected to follow the following syntax:
@@ -161,16 +164,28 @@ impl super::Adapter {
     }
 
     pub(super) unsafe fn expose(gl: glow::Context) -> Option<crate::ExposedAdapter<super::Api>> {
-        let vendor = gl.get_parameter_string(glow::VENDOR);
-        let renderer = gl.get_parameter_string(glow::RENDERER);
+        let extensions = gl.supported_extensions();
+
+        let (vendor_const, renderer_const) = if extensions.contains("WEBGL_debug_renderer_info") {
+            (GL_UNMASKED_VENDOR_WEBGL, GL_UNMASKED_RENDERER_WEBGL)
+        } else {
+            (glow::VENDOR, glow::RENDERER)
+        };
+        let (vendor, renderer) = {
+            let vendor = gl.get_parameter_string(vendor_const);
+            let renderer = gl.get_parameter_string(renderer_const);
+
+            (vendor, renderer)
+        };
         let version = gl.get_parameter_string(glow::VERSION);
+
         log::info!("Vendor: {}", vendor);
         log::info!("Renderer: {}", renderer);
         log::info!("Version: {}", version);
 
-        let ver = Self::parse_version(&version).ok()?;
-        let extensions = gl.supported_extensions();
         log::debug!("Extensions: {:#?}", extensions);
+
+        let ver = Self::parse_version(&version).ok()?;
 
         let shading_language_version = {
             let sl_version = gl.get_parameter_string(glow::SHADING_LANGUAGE_VERSION);
@@ -293,6 +308,22 @@ impl super::Adapter {
             ver >= (3, 1),
         );
 
+        let mut workarounds = super::Workarounds::empty();
+        let r = renderer.to_lowercase();
+        // Check for Mesa sRGB clear bug. See
+        // [`super::PrivateCapabilities::MESA_I915_SRGB_SHADER_CLEAR`].
+        if r.contains("mesa")
+            && r.contains("intel")
+            && r.split(&[' ', '(', ')'][..])
+                .any(|substr| substr.len() == 3 && substr.chars().nth(2) == Some('l'))
+        {
+            log::warn!(
+                "Detected skylake derivative running on mesa i915. Clears to srgb textures will \
+                use manual shader clears."
+            );
+            workarounds.set(super::Workarounds::MESA_I915_SRGB_SHADER_CLEAR, true);
+        }
+
         let downlevel_defaults = wgt::DownlevelLimits {};
 
         Some(crate::ExposedAdapter {
@@ -300,6 +331,7 @@ impl super::Adapter {
                 shared: Arc::new(super::AdapterShared {
                     context: gl,
                     private_caps,
+                    workarounds,
                     shading_language_version,
                 }),
             },
@@ -327,6 +359,34 @@ impl super::Adapter {
             },
         })
     }
+
+    unsafe fn create_shader_clear_program(
+        gl: &glow::Context,
+    ) -> (glow::Program, glow::UniformLocation) {
+        let program = gl
+            .create_program()
+            .expect("Could not create shader program");
+        let vertex = gl
+            .create_shader(glow::VERTEX_SHADER)
+            .expect("Could not create shader");
+        gl.shader_source(vertex, include_str!("./shader_clear.vert"));
+        gl.compile_shader(vertex);
+        let fragment = gl
+            .create_shader(glow::FRAGMENT_SHADER)
+            .expect("Could not create shader");
+        gl.shader_source(fragment, include_str!("./shader_clear.frag"));
+        gl.compile_shader(fragment);
+        gl.attach_shader(program, vertex);
+        gl.attach_shader(program, fragment);
+        gl.link_program(program);
+        let color_uniform_location = gl
+            .get_uniform_location(program, "color")
+            .expect("Could not find color uniform in shader clear shader");
+        gl.delete_shader(vertex);
+        gl.delete_shader(fragment);
+
+        (program, color_uniform_location)
+    }
 }
 
 impl crate::Adapter<super::Api> for super::Adapter {
@@ -349,6 +409,11 @@ impl crate::Adapter<super::Api> for super::Adapter {
         let zeroes = vec![0u8; super::ZERO_BUFFER_SIZE];
         gl.buffer_data_u8_slice(glow::COPY_READ_BUFFER, &zeroes, glow::STATIC_DRAW);
 
+        // Compile the shader program we use for doing manual clears to work around Mesa fastclear
+        // bug.
+        let (shader_clear_program, shader_clear_program_color_uniform_location) =
+            Self::create_shader_clear_program(gl);
+
         Ok(crate::OpenDevice {
             device: super::Device {
                 shared: Arc::clone(&self.shared),
@@ -365,8 +430,11 @@ impl crate::Adapter<super::Api> for super::Adapter {
                 copy_fbo: gl
                     .create_framebuffer()
                     .map_err(|_| crate::DeviceError::OutOfMemory)?,
+                shader_clear_program,
+                shader_clear_program_color_uniform_location,
                 zero_buffer,
                 temp_query_results: Vec::new(),
+                draw_buffer_count: 1,
             },
         })
     }
--- a/wgpu-hal/src/gles/command.rs
+++ b/wgpu-hal/src/gles/command.rs
@@ -444,10 +444,11 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
                 self.cmd_buffer
                     .commands
                     .push(match cat.target.view.sample_type {
-                        wgt::TextureSampleType::Float { .. } => C::ClearColorF(
-                            i as u32,
-                            [c.r as f32, c.g as f32, c.b as f32, c.a as f32],
-                        ),
+                        wgt::TextureSampleType::Float { .. } => C::ClearColorF {
+                            draw_buffer: i as u32,
+                            color: [c.r as f32, c.g as f32, c.b as f32, c.a as f32],
+                            is_srgb: cat.target.view.format.describe().srgb,
+                        },
                         wgt::TextureSampleType::Depth => unimplemented!(),
                         wgt::TextureSampleType::Uint => C::ClearColorU(
                             i as u32,
--- a/wgpu-hal/src/gles/device.rs
+++ b/wgpu-hal/src/gles/device.rs
@@ -248,8 +248,10 @@ impl super::Device {
                         return Err(crate::DeviceError::Lost.into());
                     }
                     super::BindingRegister::Textures | super::BindingRegister::Images => {
-                        let loc = gl.get_uniform_location(program, name).unwrap();
-                        gl.uniform_1_i32(Some(&loc), slot as _);
+                        gl.uniform_1_i32(
+                            gl.get_uniform_location(program, name).as_ref(),
+                            slot as _,
+                        );
                     }
                 }
             }
@@ -589,6 +591,7 @@ impl crate::Device<super::Api> for super::Device {
                 & crate::FormatAspects::from(desc.range.aspect),
             mip_levels: desc.range.base_mip_level..end_mip_level,
             array_layers: desc.range.base_array_layer..end_array_layer,
+            format: texture.format,
         })
     }
     unsafe fn destroy_texture_view(&self, _view: super::TextureView) {}
--- a/wgpu-hal/src/gles/mod.rs
+++ b/wgpu-hal/src/gles/mod.rs
@@ -68,6 +68,8 @@ mod queue;
 #[cfg(not(target_arch = "wasm32"))]
 use self::egl::{Instance, Surface};
 
+use arrayvec::ArrayVec;
+
 use glow::HasContext;
 
 use std::{ops::Range, sync::Arc};
@@ -113,13 +115,26 @@ bitflags::bitflags! {
     /// change the exposed feature set.
     struct PrivateCapabilities: u32 {
         /// Support explicit layouts in shader.
-        const SHADER_BINDING_LAYOUT = 0x0001;
+        const SHADER_BINDING_LAYOUT = 1 << 0;
         /// Support extended shadow sampling instructions.
-        const SHADER_TEXTURE_SHADOW_LOD = 0x0002;
+        const SHADER_TEXTURE_SHADOW_LOD = 1 << 1;
         /// Support memory barriers.
-        const MEMORY_BARRIERS = 0x0004;
+        const MEMORY_BARRIERS = 1 << 2;
         /// Vertex buffer layouts separate from the data.
-        const VERTEX_BUFFER_LAYOUT = 0x0008;
+        const VERTEX_BUFFER_LAYOUT = 1 << 3;
+    }
+}
+
+bitflags::bitflags! {
+    /// Flags that indicate necessary workarounds for specific devices or driver bugs
+    struct Workarounds: u32 {
+        // Needs workaround for Intel Mesa bug:
+        // https://gitlab.freedesktop.org/mesa/mesa/-/issues/2565.
+        //
+        // This comment
+        // (https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/4972/diffs?diff_id=75888#22f5d1004713c9bbf857988c7efb81631ab88f99_323_327)
+        // seems to indicate all skylake models are effected.
+        const MESA_I915_SRGB_SHADER_CLEAR = 1 << 0;
     }
 }
 
@@ -148,6 +163,7 @@ struct TextureFormatDesc {
 struct AdapterShared {
     context: glow::Context,
     private_caps: PrivateCapabilities,
+    workarounds: Workarounds,
     shading_language_version: naga::back::glsl::Version,
 }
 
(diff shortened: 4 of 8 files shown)


wgpu-hal/src/gles/queue.rs
```diff
@@ -560,7 +655,15 @@ impl super::Queue {
                 }
             }
             C::ClearColorF(draw_buffer, ref color) => {
-                gl.clear_buffer_f32_slice(glow::COLOR, draw_buffer, color);
+                if self
+                    .shared
+                    .private_caps
+                    .contains(super::PrivateCapabilities::REQUIRES_SHADER_CLEAR)
```
cwfitzgerald, line 661: This check is too general, we should only be doing this if we are clearing to an SRGB buffer.

wgpu-hal/src/gles/queue.rs
```diff
@@ -25,17 +25,112 @@ fn is_3d_target(target: super::BindTarget) -> bool {
...
+        current_stencil_test_enabled,
+        STENCIL_TEST
+    );
+    gl_modes_set_enabled_fn!(
+        set_scissor_test_enabled,
+        current_scissor_test_enabled,
+        SCISSOR_TEST
+    );
+    gl_modes_set_enabled_fn!(set_blend_enabled, current_blend_enabled, BLEND);
+    gl_modes_set_enabled_fn!(set_cull_face_enabled, current_cull_face_enabled, CULL_FACE);
```
cwfitzgerald, line 65: If we're keeping all this state around anyway, we should be using it to deduplicate state changes.

wgpu-hal/src/gles/adapter.rs
```diff
@@ -349,6 +393,29 @@ impl crate::Adapter<super::Api> for super::Adapter {
...
+            gl.compile_shader(fragment);
+            gl.attach_shader(program, vertex);
+            gl.attach_shader(program, fragment);
+            gl.link_program(program);
+            let color_uniform_loc = gl
+                .get_uniform_location(program, "color")
+                .expect("Could not find color uniform in shader clear shader");
+
+            Ok((program, color_uniform_loc))
+        })()
```
cwfitzgerald, line 416: Should probably factor this out into a function.

wgpu-hal/src/gles/device.rs
```diff
@@ -248,8 +248,10 @@ impl super::Device {
                         return Err(crate::DeviceError::Lost.into());
                     }
                     super::BindingRegister::Textures | super::BindingRegister::Images => {
-                        let loc = gl.get_uniform_location(program, name).unwrap();
-                        gl.uniform_1_i32(Some(&loc), slot as _);
+                        // The location could be none if the uniform isn't used in the shader
```
cwfitzgerald, line 251: `gl.uniform_1_i32` can accept a None. In addition, glow is already performing this check for us, so we shouldn't bother.

cwfitzgerald (changes requested):
Some comments, but in general looks good

wgpu-hal/src/gles/adapter.rs
```diff
@@ -293,6 +321,22 @@ impl super::Adapter {
...
+        // (https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/4972/diffs?diff_id=75888#22f5d1004713c9bbf857988c7efb81631ab88f99_323_327)
+        // seems to indicate all skylake models are effected, so we check for renderer models that
+        // have their 3 letter code ending with `L`
+        let r = renderer.to_lowercase();
+        if r.contains("mesa")
+            && r.contains("intel")
+            && r.split(&[' ', '(', ')'][..])
+                .any(|substr| substr.len() == 3 && substr.chars().nth(2) == Some('l'))
+        {
+            private_caps.set(super::PrivateCapabilities::REQUIRES_SHADER_CLEAR, true);
```
cwfitzgerald, line 337: We should also issue a log::warn when we enable this. Something like:

"Detected skylake derivative running on mesa i915. Clears to srgb framebuffers will use manual shader clears."

wgpu-hal/src/gles/adapter.rs
```diff
@@ -293,6 +321,22 @@ impl super::Adapter {
...
+        // Check platforms effected by mesa bug:
+        // https://gitlab.freedesktop.org/mesa/mesa/-/issues/2565.
+        //
+        // This comment
+        // (https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/4972/diffs?diff_id=75888#22f5d1004713c9bbf857988c7efb81631ab88f99_323_327)
+        // seems to indicate all skylake models are effected, so we check for renderer models that
+        // have their 3 letter code ending with `L`
+        let r = renderer.to_lowercase();
+        if r.contains("mesa")
+            && r.contains("intel")
```
kvark, line 333: is it possible to check for Vendor ID here instead of a string? or would WebGL not give us the vendor id?

wgpu-hal/src/gles/adapter.rs
```diff
@@ -349,6 +393,29 @@ impl crate::Adapter<super::Api> for super::Adapter {
         let zeroes = vec![0u8; super::ZERO_BUFFER_SIZE];
         gl.buffer_data_u8_slice(glow::COPY_READ_BUFFER, &zeroes, glow::STATIC_DRAW);
 
+        // Compile the shader program we use for doing manual clears to work around Mesa fastclear
+        // bug.
+        //
+        // Compile in a closure so we can use `?` to unwrap the errors and handle them all once
+        let (shader_clear_program, shader_clear_program_color_uniform_location) = (|| {
```
kvark, line 400: honestly, I'd rather see `unwrap()` here without any closures. Debugging closures is inconvenient, and handling failures here specifically isn't very actionable for the user anyway

wgpu-hal/src/gles/adapter.rs
```diff
@@ -349,6 +393,29 @@ impl crate::Adapter<super::Api> for super::Adapter {
...
+            gl.shader_source(fragment, include_str!("./shader_clear.frag"));
+            gl.compile_shader(fragment);
+            gl.attach_shader(program, vertex);
+            gl.attach_shader(program, fragment);
+            gl.link_program(program);
+            let color_uniform_loc = gl
+                .get_uniform_location(program, "color")
+                .expect("Could not find color uniform in shader clear shader");
+
+            Ok((program, color_uniform_loc))
```
kvark, line 415: we should also delete the shaders here

wgpu-hal/src/gles/adapter.rs
```diff
@@ -365,8 +432,19 @@ impl crate::Adapter<super::Api> for super::Adapter {
                 copy_fbo: gl
                     .create_framebuffer()
                     .map_err(|_| crate::DeviceError::OutOfMemory)?,
+                shader_clear_program,
+                shader_clear_program_color_uniform_location,
                 zero_buffer,
                 temp_query_results: Vec::new(),
+                state: Mutex::new(super::QueueState {
```
kvark, line 439: is this necessary to be a part of the fastclear workaround? I'd like to discuss this separately

kvark (changes requested):

wgpu-hal/src/gles/adapter.rs
cwfitzgerald: Unfortunately it seems to have no concept of a vendor ID :(

wgpu-hal/src/gles/queue.rs
zicklag: How would I know whether or not the buffer is sRGB?

wgpu-hal/src/gles/adapter.rs
zicklag: The mutex made it possible to modify the state ( mutable access ) while also having a reference to `self.shared.context` ( immutable access ), which rust didn't like without the mutex. There might be another way to get around that. I'll see if there's an alternative I could get working. We might just need to not do `let gl = &self.shared.context` at the top of everything so that we can still mutably access our internal state.

kvark: Why does the queue state need to be shared with anything?

zicklag: Ah, nevermind, it doesn't, I had just written the helper methods on `Queue` which took `&mut self` to the entire queue, when it only needed mutable access to the queue state. Moving the helper methods to the `QueueState` struct and adding an argument for the gl context fixed the borrowing issues without needing a mutex.

I pushed an update without the mutex.

wgpu-hal/src/gles/queue.rs
zicklag: When trying to deduplicate the changes, interestingly things stop working. Like the depth or blending not getting set correctly. I think that means our tracked state is somehow getting out of sync with the actual GL state. :/

There's not a way to query the actual GL state is there? That would be better than us tracking it outselves probably.

kvark:
I think we don't actually need to track the GL state here. Correct me if that's wrong.
We may want to do it anyway at some point later, but right now it's not the time and place to do. We need correctness first, complexity later.

The reason why I don't think it's needed is because the clears are only done at the beginning of a render pass. This is where all the state is reset from WebGPU point of view, so the GL backend doesn't need to worry about restoring any states if it does the draw.

zicklag:
OK, I think I addressed all of the review points with the latest push.

wgpu-hal/src/gles/adapter.rs
```diff
@@ -349,6 +394,30 @@ impl crate::Adapter<super::Api> for super::Adapter {
...
+        gl.shader_source(fragment, include_str!("./shader_clear.frag"));
+        gl.compile_shader(fragment);
+        gl.attach_shader(shader_clear_program, vertex);
+        gl.attach_shader(shader_clear_program, fragment);
+        gl.link_program(shader_clear_program);
+        let shader_clear_program_color_uniform_location = gl
+            .get_uniform_location(shader_clear_program, "color")
+            .expect("Could not find color uniform in shader clear shader");
+        gl.delete_shader(vertex);
+        gl.delete_shader(fragment);
```
cwfitzgerald, line 419: This should still be its own function.

wgpu-hal/src/gles/mod.rs
```diff
@@ -562,7 +572,7 @@ enum Command {
     },
     InvalidateAttachments(InvalidatedAttachments),
     SetDrawColorBuffers(u8),
-    ClearColorF(u32, [f32; 4]),
+    ClearColorF(u32, [f32; 4], bool /* is sRGB */),
```
cwfitzgerald, line 575: Could we give these names instead of being a tuple

wgpu-hal/src/gles/mod.rs
```diff
@@ -113,13 +115,16 @@ bitflags::bitflags! {
...
+        const SHADER_TEXTURE_SHADOW_LOD = 1 << 1;
         /// Support memory barriers.
-        const MEMORY_BARRIERS = 0x0004;
+        const MEMORY_BARRIERS = 1 << 2;
         /// Vertex buffer layouts separate from the data.
-        const VERTEX_BUFFER_LAYOUT = 0x0008;
+        const VERTEX_BUFFER_LAYOUT = 1 << 3;
+        /// Indicates that we must do a manual shader clear instead of using `gl.clear_buffer` to
+        /// workaround Mesa fastclear bugs.
+        const REQUIRES_SHADER_CLEAR = 1 << 4;
```
cwfitzgerald, line 127:
```suggestion
        const DRIVER_WORKAROUND_SRGB_SHADER_CLEAR = 1 << 4;
```

open to bikesheding (@kvark)

I want to see which of these are a problem with the driver and which are a device limitation.

wgpu-hal/src/gles/mod.rs
```diff
@@ -113,13 +115,16 @@ bitflags::bitflags! {
...
         /// Support extended shadow sampling instructions.
-        const SHADER_TEXTURE_SHADOW_LOD = 0x0002;
+        const SHADER_TEXTURE_SHADOW_LOD = 1 << 1;
         /// Support memory barriers.
-        const MEMORY_BARRIERS = 0x0004;
+        const MEMORY_BARRIERS = 1 << 2;
         /// Vertex buffer layouts separate from the data.
-        const VERTEX_BUFFER_LAYOUT = 0x0008;
+        const VERTEX_BUFFER_LAYOUT = 1 << 3;
+        /// Indicates that we must do a manual shader clear instead of using `gl.clear_buffer` to
```
cwfitzgerald, line 125: Could we get all the information about the bug here. Then the documentation for the workaround can point to the documentation of this flag.

This way we can have all driver bugs shown off in one place.

wgpu-hal/src/gles/shader_clear.vert
```diff
@@ -0,0 +1,11 @@
+#version 300 es
+precision lowp float;
+// A triangle that fills the whole screen
+const vec2[3] TRIANGLE_POS = vec2[](
+  vec2( 0., -3.),
```
cwfitzgerald, line 5: I have a small personal vendetta against omitting the trailing `0` from floats. Could you add them back :)

wgpu-hal/src/gles/adapter.rs
```diff
@@ -161,17 +161,42 @@ impl super::Adapter {
...
+        let (vendor, renderer) = {
+            let vendor = gl.get_parameter_string(glow::VENDOR);
+            let renderer = gl.get_parameter_string(glow::RENDERER);
+
+            (vendor, renderer)
+        };
+        #[cfg(target_arch = "wasm32")]
+        let (vendor, renderer) = {
+            let vendor_const = if extensions.contains("WEBGL_debug_renderer_info") {
+                0x9245 // UNMASKED_VENDOR_WEBGL
```
cwfitzgerald, line 176: Could we get these magic numbers as constants

cwfitzgerald (changes requested):
A few more minor things, but this looks good.

wgpu-hal/src/gles/mod.rs
cwfitzgerald: A few other options: `WORKAROUND_MESA_I915_SRGB_SHADER_CLEAR`, `WORKAROUND_MESA_SRGB_SHADER_CLEAR`

zicklag:
There you go! I went with the most specific `PrivateCapabilities` flag name: `WORKAROUND_MESA_I915_SRGB_SHADER_CLEAR`. I don't care what we go with final, just let me know if you want to change it.

zicklag:
Oh, wait! The shadow example is only clearing half the screen for some reason. Got to check that out.

zicklag:
OK, fixed it, had to set the viewport and scissor before clearing.

wgpu-hal/src/gles/adapter.rs
```diff
@@ -293,6 +323,24 @@ impl super::Adapter {
             ver >= (3, 1),
         );
 
+        let r = renderer.to_lowercase();
+        // Check for Mesa sRGB clear bug. See
+        // [`super::PrivateCapabilities::WORKAROUND_MESA_SRGB_SHADER_CLEAR`].
```
cwfitzgerald, line 328: This link is wrong, and it's concerning that this wasn't caught.

cwfitzgerald (approved):

wgpu-hal/src/gles/adapter.rs
zicklag: Ah, yeah, it's not actually a doc comment, so it's not checked by the doc tools or anything. I just used the same style. I'll fix it.

wgpu-hal/src/gles/command.rs
```diff
@@ -421,10 +421,7 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
             }
         }
 
-        // set the draw buffers and states
-        self.cmd_buffer
```
kvark, line 425: I don't think we can do this, since the draw targets are required for the clears on the normal path

wgpu-hal/src/gles/queue.rs
```diff
@@ -559,8 +581,21 @@ impl super::Queue {
...
+                ref color,
+                is_srgb,
+            } => {
+                if self
+                    .shared
+                    .private_caps
+                    .contains(super::PrivateCapabilities::WORKAROUND_MESA_I915_SRGB_SHADER_CLEAR)
+                    && is_srgb
+                {
+                    self.perform_shader_clear(draw_buffer, *color);
```
kvark, line 595: I mentioned that we can't postpone `SetDrawBuffers`. My suggestion is the following:
The queue would have the "draw_buffer_count: u32" field as a part of the state, setup by `SetDrawBuffers` in addition to what it does today. Then `fn perform_shader_clear` can reset to this mask after it's done.

zicklag:
Pushed another update. :+1:

wgpu-hal/src/gles/adapter.rs
```diff
@@ -161,17 +166,42 @@ impl super::Adapter {
...
+        #[cfg(not(target_arch = "wasm32"))]
+        let (vendor, renderer) = {
+            let vendor = gl.get_parameter_string(glow::VENDOR);
+            let renderer = gl.get_parameter_string(glow::RENDERER);
+
+            (vendor, renderer)
+        };
+        #[cfg(target_arch = "wasm32")]
+        let (vendor, renderer) = {
+            let vendor_const = if extensions.contains("WEBGL_debug_renderer_info") {
```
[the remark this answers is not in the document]
cwfitzgerald: I think what he meant was

```
let (vendor, renderer) = if extensions.contains("WEBGL_debug_renderer_info") {
     (GL_UNMASKED_VENDOR_WEBGL, GL_UNMASKED_RENDERER_WEBGL)
else {
     (glow::VENDOR, glow::RENDERER)
}
```

wgpu-hal/src/gles/mod.rs
```diff
@@ -113,13 +115,26 @@ bitflags::bitflags! {
...
+bitflags::bitflags! {
+    /// Flags that indicate necessary workarounds for specific devices or driver bugs
+    struct Workarounds: u32 {
+        // Needs workaround for Intel Mesa bug:
+        // https://gitlab.freedesktop.org/mesa/mesa/-/issues/2565.
+        //
+        // This comment
+        // (https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/4972/diffs?diff_id=75888#22f5d1004713c9bbf857988c7efb81631ab88f99_323_327)
+        // seems to indicate all skylake models are effected.
+        const WORKAROUND_MESA_I915_SRGB_SHADER_CLEAR = 1 << 0;
```
cwfitzgerald, line 137: Now that this is in a workarounds struct, the individual const shouldn't be prefixed with WORKAROUND

cwfitzgerald (changes requested):
This has had a lot of iterations, thanks for sticking with us! We're getting close.

wgpu-hal/src/gles/adapter.rs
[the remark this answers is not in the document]
zicklag: Ah, that makes more sense. :) Pushed an update.

wgpu-hal/src/gles/mod.rs
zicklag: That makes sense. Fixed!

cwfitzgerald (approved):
Looks good from my end. Just want kvark to take one last look. Does this make all the GL tests pass on your machine?

zicklag:
One of the skybox tests is the only failing one, and that one fails on master so I think it's unrelated.

cwfitzgerald:
Alright, sounds fine.

kvark (approved):

kvark:
bors r=cwfitzgerald,kvark

PR #1718, opened 2021-07-23
Update naga to bbfa9a0, tweak DX12 debug output

kvark:
**Connections**
Picks up #1132

**Description**
Also reduces the spam output of DX12 backend

**Testing**
Got skybox and water examples running!


kvark:
wgpu-dx12-water

PR #1720, opened 2021-07-23
Gitignore *gitignore*

zicklag:
Ignore everything with `gitignore` in the name, except for the
.gitignore file.

**Connections**
None

**Description**
This is something I personally put in all my gitignore files, but it's totally fine if you don't want it :). It helps me when I have temporary files that I have laying around inside of the git repository that I don't want to accidentally commit.

**Testing**
Just a gitignore change, but tested on Linux.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,8 @@ wgpu/red.png
 # Output from invalid comparison tests
 **/*-actual.png
 **/*-difference.png
+
+# Ignore everything with `gitignore` in the name, except for
+# the .gitignore file
+*gitignore*
+!.gitignore


kvark:
Interesting. For my workflow, such change wouldn't have any effect. There are often files lying around, but they are easy to spot when adding files to the commit.
@cwfitzgerald or @grovesNL would you be (positively) affected by the change?

cwfitzgerald:
Completely indifferent :)

kvark:
@zicklag what Git tooling do you use? Does it not allow you to select what you include in the commits and what not?

Gordon-F:
As an alternative, you can set up global gitignore - https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

zicklag:
It does. I use VSCode. Or the git CLI.

Because I operate largely out of my VSCode workspace, sometimes I'll put a whole `gitignore` folder in the workspace with backup files or samples or stuff that I don't want to commit but that I want to be able to easily find from VSCode. In my workspace right now I've got forks of the `index.html` file used for the WebGL samples called `index.gitignore.html`,  `webgl-lint.gitignore.js`, and `webgl-lint-web-example.sh` script that I have for running the WebGL examples with the webgl-lint script enabled, but those aren't things we'd want in the repo, at least not yet.

> As an alternative, you can set up global gitignore

I didn't know that was a thing! Cool. If I'd have known that I wouldn't have bothered wasting your guys's time with this PR. :) Since it sounds like I'm the only one using this right now, that's a much better option I think. Thanks!

PR #1726, opened 2021-07-26
Adding pipeline caches for wgpu-hal (only works on Vulkan so far)

Lonsdaleiter:
**Connections**
(none)

**Description**
I notice gfx-hal had support for pipeline caches, but wgpu-hal still lacks support. They're not in the webgpu spec, but I think it would at least make sense to expose them for some native backends. This pull request adds support for them in wgpu-hal, but does not yet expose this functionality to wgpu-rs. It only works on the Vulkan backend so far - the rest of them just log a warning and do nothing, though I expect it should be possible to implement it for Metal using binary archives (given that gfx-hal 0.8 did just this).

**Testing**
I have tested that wgpu-hal and wgpu-rs build on my 64-bit x86 Linux running on Vulkan/OpenGL. The wgpu-rs examples and (the slightly modified) wgpu-hal halmark run properly, though none use the pipeline caches.

### DIFF
--- a/wgpu-hal/examples/halmark/main.rs
+++ b/wgpu-hal/examples/halmark/main.rs
@@ -232,7 +232,7 @@ impl<A: hal::Api> Example<A> {
                 write_mask: wgt::ColorWrites::default(),
             }],
         };
-        let pipeline = unsafe { device.create_render_pipeline(&pipeline_desc).unwrap() };
+        let pipeline = unsafe { device.create_render_pipeline(&pipeline_desc, None).unwrap() };
 
         let texture_data = vec![0xFFu8; 4];
 
--- a/wgpu-hal/src/dx12/device.rs
+++ b/wgpu-hal/src/dx12/device.rs
@@ -1106,6 +1106,23 @@ impl crate::Device<super::Api> for super::Device {
         // just drop
     }
 
+    unsafe fn create_empty_pipeline_cache(&self) -> super::PipelineCache {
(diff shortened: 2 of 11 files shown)


kvark:
Thank you for the contribution!
I wish we could chat about this before you did all the work though.

The design process for wgpu-hal is different from gfx-hal. We don't just blindly expose anything that's in Vulkan (or gfx-hal). Instead, wgpu-hal tries very hard to be minimal and portable.
Therefore, before anything related to pipeline caches is introduced, there needs to be a proper investigation on how the backends are going to implement it, and a proposal for the API. Just like for stuff in WebGPU itself.

Lonsdaleiter:
I see - yeah, I probably should have done a more thorough investigation into how wgpu-hal works beforehand. Good thing is that it wasn't very much work, what was in this request.

I suppose it would be out of the scope of a noob contributor to see about starting an investigation into pipeline caches? It seems like it would be possible to make pipeline caches do some things and provide some benefit on some backends (Vk/Mtl/Dx12) and just do nothing on others (Gl/Dx11/WebGPU); where could such an investigation take place or be started for wgpu-hal?

kvark:
If you want to investigate this, we'd very much appreciate it! But I expect the area to be quite complicated. APIs may have multiple ways to speed up pipeline creation, e.g. Vulkan has both pipeline caches and derivatives. And we need to see all of them, weight them properly by how much benefit they provide, to find a sound intersection.

Issue #1082, opened 2020-12-08
Moving error formatting from wgpu-rs to wgpu-core


scoopr:
As discussed in #1068 it may be that some of the error formatting that is now in wgpu-rs should in fact be moved to wgpu-core.

As I don't have time to do help with it right now, I wanted to write down the steps I was thinking about it so that I don't forget. Also maybe it helps if anyone else is willing to help with this one.

The `format_error` in wgpu-rs `error.rs` is the one that should end up either in directly to `Global`, or as a separate helper that takes `Global` as argument (It is basically doing that through `Context` now). Trivial in itself, but it depend on `fmt_pretty_any`

`fmt_pretty_any` is what is doing the downcasting of errors, and mostly could just be moved to wgpu-core as-is, but it also downcasts the `ContextError` that is more specific to wgpu-rs.

The `ContextError` is used to decorate the api calls in wgpu-rs for which api call the error came from. This was initially done because it had convenient access to the descriptor label, even when nothing was created on errors, as the error `Id`s were created on wgpu-rs error side, but since this was changed in #1034, it does not need to be in wgpu-rs side anymore.

So working backwards,
• Move/add `ContextError` to wgpu-core
• Decorate the api call sites in wgpu-core
• Remove the `ContextError` decoration in wgpu-rs `handle_error`
• Now all of error.rs could be moved in wgpu-core (and slight tweak to `handle_error` to call the `format_error` in its new place)

Alternatively, if just having some working `format_error` is deemed more important and waffing about the `ContextError` is boring busywork to be done later,
• Move `PrettyError` trait and its impls (except for the ContextError)
• Add some hook (callback?) to `format_error`/`fmt_pretty_any` that allows wgpu-rs to inject its error to be downcasted
• Move `format_error` to core. That should leave only the `ContextError` related stuff in error.rs

That was quite a lot of text, what is ultimately fairly simple set of changes.

scoopr:
Done in #1728

PR #1728, opened 2021-07-26
Move error formatting functionality 

scoopr:
**Connections**
This is an attempt to implement #1082 

**Description**
This is a fairly straightforward move of the error formatting that were in wgpu-rs in to wgpu-code.

The enriching of some of the errors with the additional `ContextError` were still left with the wgpu-rs code, as I realised the current error messages refer to the API names as they appear in rust code and might not make sense for all the consumers of wgpu-core. It could be left as is and other consumers might enrich in their own ways, or the messages could be reformatted to be more conversational

**Testing**
Limited testing with the examples


scoopr:
It may be a bit rough around the edges still.

As it is, I don't it really bring much progress/clarity/etc to the code, but I think the biggest reason was to get the gecko error formatting to use the same formatting mechanism eventually.

Aside, "bringing the family back together" lowered the bar significantly for me to even start with this :)

wgpu-core/src/error.rs
```diff
@@ -0,0 +1,321 @@
...
+                let name = crate::gfx_select!(id => global.sampler_label(id));
+                ret.push_str(&format_label_line("sampler", &name));
+            }
+            _ => {}
+        };
+        ret
+    }
+}
+
+impl PrettyError for crate::binding_model::CreatePipelineLayoutError {
```
kvark, line 68: Would it make sense to move each implementation into the module that own the type (i.e. `CreatePipelineLayoutError` in this case)? This way, if the error gets more variants, it's easier to see the associated code.

wgpu-core/src/error.rs
```diff
@@ -0,0 +1,321 @@
...
+
+pub fn format_error_line(err: &dyn fmt::Display) -> String {
+    format!("    {}\n", err)
+}
+
+pub fn format_note_line(note: &dyn fmt::Display) -> String {
+    format!("      note: {}\n", note)
+}
+
+pub fn format_label_line(label_key: &str, label_value: &str) -> String {
```
kvark, line 250: It would be elegant to make it accept an `fmt::Formatter`, so that it doesn't allocate anything.
Yes, I know the performance doesn't matter here, so this isn't required to be elegant :)

wgpu-core/src/error.rs
```diff
@@ -0,0 +1,321 @@
+use core::fmt;
+use std::error::Error;
+
+use crate::{
+    gfx_select,
+    hub::{Global, IdentityManagerFactory},
+};
+
+pub trait AsDisplay {
```
kvark, line 9: Looks like this is only needed for just a few cases. Could it be worked around?
Something like https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ced1f8a2ddc2c0b986421b1acd195588

wgpu-core/src/error.rs
```diff
@@ -0,0 +1,321 @@
...
+    }
+}
+
+impl PrettyError for crate::command::ComputePassErrorInner {
+    fn fmt_pretty(&self, global: &Global<IdentityManagerFactory>) -> String {
+        let mut ret = format_error_line(self);
+        match *self {
+            Self::InvalidBindGroup(id) => {
+                let name = crate::gfx_select!(id => global.bind_group_label(id));
+                ret.push_str(&format_label_line("bind group", &name));
```
kvark, line 133: this should probably look like `format_label_line(&mut ret, "bind group", &name)`, so that it writes to the string directly instead of first creating a new string and then taking a reference of it

kvark (commented):
Very nice!

wgpu-core/src/error.rs
scoopr: Ah yeah, true, now that that is possible and makes sense.
Does the `PrettyError` name still make sense if it is used more widely?

I'm somehow still hoping for a way to implement the formatting with the standard `Display` trait, but I guess there is still no way of wiggling the Global in there to resolve the labels. Does pythonesques rewrite affect this, would the label be accessible just from the Id? That would simplify the whole thing, if we could just format the errors like any other errors.

wgpu-core/src/error.rs
scoopr: I think I attempted that while back when I was initially implementing this, and I couldn't figure it out quite right.
The problem is that you can't really construct a `fmt::Formatter`, so it is only really useful in a context where you already have it, like within `Display`/`Debug`. 

Current call sites want the result as a string, and as I understood it, I would have to wrap the stuff to something that implements Display to get the result, and I didn't want to deal with that.

I suppose it could return `fmt::Arguments`, but I guess it would do some allocation anyway? 

If we happen to get rid of the whole `PrettyError` in favour of the `Display`, then the issue is moot and `fmt::Formatter` is the easiest choice.

wgpu-core/src/error.rs
scoopr: Yeah, I don't remember the whole thought process why it originally was like that, but that looks like workable solution.

wgpu-core/src/error.rs
scoopr: Hmm yeah, if the `fmt::Formatter` etc. changes don't pan out, then that seems like a worthwhile change.

wgpu-core/src/error.rs
scoopr: Hmm, couldn't get it quite to work.
```
pub fn format_pretty_any(
    global: &Global<IdentityManagerFactory>,
    error: &(dyn Error + 'static),
) -> String
```
In here, even though `Error´ requires Display, somehow it won't allow casting to it. Adding the trait to the trait object requirement doesn't work and suggests making a wrapper trait that requires the two things, but `Error` already requires `Display`, so that seems like a dead-end.

```
pub trait PrettyError: Error {
    fn fmt_pretty(&self, _global: &Global<IdentityManagerFactory>) -> String {
        let self_display = self as &dyn fmt::Display;
        format_error_line(self_display)
    }
}
```
In here it says self doesn't have compile time size, though I'm not entirely sure why it requires that, isn't trait object a compile time known size?

Either I'm missing some syntax that I'm forgetting, or its not really possible to do it like that.

wgpu-core/src/error.rs
kvark: sorry, it would be generic over `fmt::Write`, not `fmt::Formatter`

scoopr: Oh derp. I should of realized that myself, I implemented this now, it somehow makes a bit more sense now :)

scoopr:
In the latest wip commit, I attempted to move the PrettyError impls closer to the error itself.

As I was trying it out in the naive way of just moving the code, I hated how all kinds of types started to leak out everywhere (PrettyError, Global, gfx_select, etc.), so I thought it could use some abstraction over it.

So in here, I tried a `PrettyError2`, just for `RenderCommandError`, to feel it out. I end up throwing the Write and Global refs in to this wrapper struct, that has convenience methods for formatting the lines, I even ended up making a convenience method per label type, which nicely wraps over the gfx_select stuff. The using code doesn't really need to know more than the `PrettyError2` type itself to be useful.

Do you think this is an approach worth pursuing, or is it too heavy handed?

wgpu-core/src/error.rs
kvark: cc @pythonesque

> Does the PrettyError name still make sense if it is used more widely?

seems fine, unless you have ideas for a better name ;)

kvark (commented):
Not seeing anything wrong with this approach. If you have the set of trade-offs in mind for this, please list them up. Maybe that will ring a bell for me :)

pythonesque:
Seems fine to me :)  All the methods that take Hubs/Global (that are currently needed to extract labels) are just static methods with hubs removed, so this change should extend logically to a Hubs-free world.

scoopr:
If the Global is truly going away with the refactor, I’m somewhat tempted to hold off until that lands, as this could just be a “simply” a Display impl (the error contains everything it needs to display itself). 

Thou I suppose it would be odd to have the Display show the notes/labels, especially if it assumes some form of formatting/indent etc.

What I’m kind of aiming at, is for the user code to be able to format error messages with the standard Error mechanism as it sees fit, but with enough of granular access to the labels etc. to be able to do some rudimentary colouring of the message. Some more advanced cases might try to even explain the source data where the label came from, if it can access it when available. Buuut I might be overthinking it :)

Maybe I’ll just continue with the trait, but impl the Display with it, and see where it takes me to.

scoopr:
Alright, here is the current result. PrettyError impls are now next to the error enums, the formatting is done through the ErrorFormatter thing introduced there.

I also worked the AsDisplay hack, so it gets removed too. I suppose if you have a trait object already, you can't pass it as another trait object, even if its a traits required trait, as far as I can see. Also some magic Sized in there

wgpu-core/src/error.rs
```diff
@@ -0,0 +1,175 @@
...
+    global: &'a Global<IdentityManagerFactory>,
+}
+
+impl<'a> ErrorFormatter<'a> {
+    pub fn error(&mut self, err: &dyn Error) {
+        writeln!(self.writer, "    {}", err).expect("Error formatting error");
+    }
+
+    pub fn note(&mut self, note: &dyn fmt::Display) {
+        writeln!(self.writer, "      note: {}", note).expect("Error formatting error");
```
kvark, line 20: any reason to not use "\t" here?

wgpu-core/src/error.rs
```diff
@@ -0,0 +1,175 @@
...
+    writer: &'a mut dyn fmt::Write,
+    global: &'a Global<IdentityManagerFactory>,
+}
+
+impl<'a> ErrorFormatter<'a> {
+    pub fn error(&mut self, err: &dyn Error) {
+        writeln!(self.writer, "    {}", err).expect("Error formatting error");
+    }
+
+    pub fn note(&mut self, note: &dyn fmt::Display) {
```
kvark, line 19: can this work with `FormatArgs`?

kvark (commented):
Nice work!

wgpu-core/src/error.rs
scoopr: Are we going to have _that_ discussion, tabs vs spaces? ;)
I didn't really intend functional changes here, and that's how it was. I feel the tabs would indent a bit too much though.

For reference, `anyhow` seems to be doing [indenting with spaces](https://github.com/dtolnay/anyhow/blob/master/src/fmt.rs#L73)

wgpu-core/src/error.rs
scoopr: Do you mean `fmt::Arguments`? And that the note/error/label stuff would return it instead of writing directly to the `Writer`? I don't really see the point, the users (PrettyError impls) would just pass it on directly to a `Writer` anyway?

Or do you mean instead of taking a `&dyn fmt::Display` argument, take a `&fmt::Arguments`? I suppose thats ok, though the caller would need to do some extra `format_args!`, that is not needed now?

scoopr: Ooh, do you mean to avoid the subsequent formatting, collapsing them to a single one, by passing fmt::Arguments until the note thing that writes it out?

kvark: yeah, but you are right, it doesn't matter too much

kvark (approved):

PR #1729, opened 2021-07-27
Handle Multi-threaded EGL Context Access

zicklag:
**Connections**
https://github.com/gfx-rs/wgpu/discussions/1630, https://github.com/bevyengine/bevy/issues/841

**Description**
Implements the synchronization necessary to use the GL backend from multiple threads. Accomplishes this by using a mutex around the GL context with extra wrapping to bind and unbind the EGL context when locking and unlocking.

**Testing**
Tested on Ubunty 20.04 with a fork of the Bevy game engine and the WGPU examples ( not that the examples test the multi-threading ).

## Remaining Issues

There is only one Bevy example I cannot get to run yet and it's the `load_gltf` example. It fails with a shader translation error:

```
Jul 26 20:36:50.949 ERROR naga::back::glsl: Conflicting samplers for _group_3_binding_10    
Jul 26 20:36:50.950  WARN wgpu::backend::direct: Shader translation error for stage FRAGMENT: A image was used with multiple samplers    
Jul 26 20:36:50.950  WARN wgpu::backend::direct: Please report it to https://github.com/gfx-rs/naga    
Jul 26 20:36:50.950 ERROR wgpu::backend::direct: wgpu error: Validation Error

Caused by:
    Internal error in FRAGMENT shader: A image was used with multiple samplers
```

Interestingly, I think the shader in question doesn't have a `group(3), binding(10)` anywhere that I know of so I'm going to have to drill down a bit more and find out exactly which shader translation is failing more.

This could potentially be fixed in a separate PR. I think the rest of this PR is rather straight-forward and the fix for the error above is probably mostly unrelated to the primary changes made in this PR.

### DIFF
--- a/wgpu-hal/src/gles/device.rs
+++ b/wgpu-hal/src/gles/device.rs
@@ -73,12 +73,11 @@ impl CompilationContext<'_> {
 
 impl super::Device {
     unsafe fn compile_shader(
-        &self,
+        gl: &glow::Context,
         shader: &str,
         naga_stage: naga::ShaderStage,
         label: Option<&str>,
     ) -> Result<glow::Shader, crate::PipelineError> {
-        let gl = &self.shared.context;
         let target = match naga_stage {
             naga::ShaderStage::Vertex => glow::VERTEX_SHADER,
             naga::ShaderStage::Fragment => glow::FRAGMENT_SHADER,
@@ -111,7 +110,7 @@ impl super::Device {
     }
 
     fn create_shader(
-        &self,
+        gl: &glow::Context,
         naga_stage: naga::ShaderStage,
         stage: &crate::ProgrammableStage<super::Api>,
         context: CompilationContext,
@@ -156,16 +155,16 @@ impl super::Device {
             reflection_info,
         );
 
-        unsafe { self.compile_shader(&output, naga_stage, stage.module.label.as_deref()) }
+        unsafe { Self::compile_shader(gl, &output, naga_stage, stage.module.label.as_deref()) }
     }
 
     unsafe fn create_pipeline<'a, I: Iterator<Item = ShaderStage<'a>>>(
         &self,
+        gl: &glow::Context,
         shaders: I,
         layout: &super::PipelineLayout,
         label: crate::Label,
     ) -> Result<super::PipelineInner, crate::PipelineError> {
-        let gl = &self.shared.context;
         let program = gl.create_program().unwrap();
         if let Some(label) = label {
             if gl.supports_debug() {
@@ -186,7 +185,7 @@ impl super::Device {
                 name_binding_map: &mut name_binding_map,
             };
 
-            let shader = self.create_shader(naga_stage, stage, context)?;
+            let shader = Self::create_shader(gl, naga_stage, stage, context)?;
             shaders_to_delete.push(shader);
         }
 
@@ -199,7 +198,7 @@ impl super::Device {
             let shader_src = format!("#version {} es \n void main(void) {{}}", version,);
             log::info!("Only vertex shader is present. Creating an empty fragment shader",);
             let shader =
-                self.compile_shader(&shader_src, naga::ShaderStage::Fragment, Some("_dummy"))?;
+                Self::compile_shader(gl, &shader_src, naga::ShaderStage::Fragment, Some("_dummy"))?;
             shaders_to_delete.push(shader);
         }
 
@@ -292,7 +291,7 @@ impl super::Device {
 
 impl crate::Device<super::Api> for super::Device {
     unsafe fn exit(self, queue: super::Queue) {
-        let gl = &self.shared.context;
+        let gl = &self.shared.context.lock();
         gl.delete_vertex_array(self.main_vao);
         gl.delete_framebuffer(queue.draw_fbo);
         gl.delete_framebuffer(queue.copy_fbo);
@@ -303,7 +302,7 @@ impl crate::Device<super::Api> for super::Device {
         &self,
         desc: &crate::BufferDescriptor,
     ) -> Result<super::Buffer, crate::DeviceError> {
-        let gl = &self.shared.context;
+        let gl = &self.shared.context.lock();
 
         let target = if desc.usage.contains(crate::BufferUses::INDEX) {
             glow::ELEMENT_ARRAY_BUFFER
@@ -360,7 +359,7 @@ impl crate::Device<super::Api> for super::Device {
         })
     }
     unsafe fn destroy_buffer(&self, buffer: super::Buffer) {
-        let gl = &self.shared.context;
+        let gl = &self.shared.context.lock();
         gl.delete_buffer(buffer.raw);
     }
 
@@ -369,7 +368,7 @@ impl crate::Device<super::Api> for super::Device {
         buffer: &super::Buffer,
         range: crate::MemoryRange,
     ) -> Result<crate::BufferMapping, crate::DeviceError> {
-        let gl = &self.shared.context;
+        let gl = &self.shared.context.lock();
 
         let is_coherent = buffer.map_flags & glow::MAP_COHERENT_BIT != 0;
 
@@ -388,7 +387,7 @@ impl crate::Device<super::Api> for super::Device {
         })
     }
     unsafe fn unmap_buffer(&self, buffer: &super::Buffer) -> Result<(), crate::DeviceError> {
-        let gl = &self.shared.context;
+        let gl = &self.shared.context.lock();
         gl.bind_buffer(buffer.target, Some(buffer.raw));
         gl.unmap_buffer(buffer.target);
         gl.bind_buffer(buffer.target, None);
@@ -398,7 +397,7 @@ impl crate::Device<super::Api> for super::Device {
     where
         I: Iterator<Item = crate::MemoryRange>,
     {
-        let gl = &self.shared.context;
+        let gl = &self.shared.context.lock();
         gl.bind_buffer(buffer.target, Some(buffer.raw));
         for range in ranges {
             gl.flush_mapped_buffer_range(
@@ -416,7 +415,7 @@ impl crate::Device<super::Api> for super::Device {
         &self,
         desc: &crate::TextureDescriptor,
     ) -> Result<super::Texture, crate::DeviceError> {
-        let gl = &self.shared.context;
+        let gl = &self.shared.context.lock();
 
         let render_usage = crate::TextureUses::COLOR_TARGET
             | crate::TextureUses::DEPTH_STENCIL_WRITE
@@ -559,7 +558,7 @@ impl crate::Device<super::Api> for super::Device {
         })
     }
     unsafe fn destroy_texture(&self, texture: super::Texture) {
-        let gl = &self.shared.context;
+        let gl = &self.shared.context.lock();
         match texture.inner {
             super::TextureInner::Renderbuffer { raw, .. } => {
                 gl.delete_renderbuffer(raw);
@@ -600,7 +599,7 @@ impl crate::Device<super::Api> for super::Device {
         &self,
         desc: &crate::SamplerDescriptor,
     ) -> Result<super::Sampler, crate::DeviceError> {
-        let gl = &self.shared.context;
+        let gl = &self.shared.context.lock();
 
         let raw = gl.create_sampler().unwrap();
 
@@ -667,7 +666,7 @@ impl crate::Device<super::Api> for super::Device {
         Ok(super::Sampler { raw })
     }
     unsafe fn destroy_sampler(&self, sampler: super::Sampler) {
-        let gl = &self.shared.context;
+        let gl = &self.shared.context.lock();
         gl.delete_sampler(sampler.raw);
     }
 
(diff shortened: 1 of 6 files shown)


wgpu-hal/src/gles/queue.rs
```diff
@@ -936,7 +939,7 @@ impl crate::Queue<super::Api> for super::Queue {
         self.reset_state();
         for cmd_buf in command_buffers.iter() {
             if let Some(ref label) = cmd_buf.label {
-                self.shared.context.push_debug_group(
+                self.shared.context.lock().push_debug_group(
```
cwfitzgerald, line 942: We should probably be favoring locking once at the top of the function and then releasing at the bottom, I can't imagine context switches are terribly cheap.

cwfitzgerald (commented):
I mean, I hate it, but seems mostly reasonable. I want to test this on stuff like rpi and android to see if it works.

wgpu-hal/src/gles/queue.rs
zicklag: Yeah, I tried that, but the issue was that `self.process` takes `&mut self` and borrowing the `gl` lock at the top takes a reference to `&self.shared.context` so Rust didn't like having the mutable and immutable references to `self` at the same time. I'm not sure a good way around it other than making `process` not take self and moving everything that it nees into function arguments.

Is that what we want to do or is there a better way?

zicklag:
> I mean, I hate it, but seems mostly reasonable.

Yeah, according to the EGL spec we can't access the context in multiple threads at a time so essentially we're stuck. :man_shrugging: 

---

Oh, something I forgot. I don't know how to test it, but according to the EGL spec I think the dummy `pbuffer` that we use as a workaround when surfaceless rendering isn't available won't work in a multi-threaded environment. You're not allowed to have multiple threads that bind the same current `pbuffer` I don't think, and that's what would happen if we didn't have surfaceless support and you tried to access the context from multiple threads.

We may need to try to detect that and maybe add a downlevel flag or an error message or something.

cwfitzgerald:
We already do have a multithreaded example, it's `cargo test --example hello-compute test_multithreaded_compute`. Additionally, the tests are multi-threaded unless `--test-threads=1` is passed.

Would having thread-local pbuffers work?

I tested this PR on my intel/haswell machine using the GL tests, and got tons of errors:

```
[2021-07-27T06:36:21Z ERROR wgpu_hal::gles::egl] EGL 'eglMakeCurrent' code 0x3009: Got an EGLSurface but no EGLContext
```

jakobhellermann:
The shader translation error is probably due to https://github.com/gfx-rs/naga/issues/1137.

wgpu-hal/src/gles/queue.rs
kvark: since `shared` is an arc, we can do:
```rust
let shared = Arc::clone(&self.shared);
let context = shared.context.lock();
```

wgpu-hal/src/gles/device.rs
```diff
@@ -74,11 +74,11 @@ impl CompilationContext<'_> {
 impl super::Device {
     unsafe fn compile_shader(
         &self,
+        gl: &glow::Context,
```
kvark, line 77: this function doesn't need `self` any more

wgpu-hal/src/gles/device.rs
```diff
@@ -112,6 +112,7 @@ impl super::Device {
 
     fn create_shader(
         &self,
+        gl: &glow::Context,
```
kvark, line 115: same here

wgpu-hal/src/gles/egl.rs
```diff
@@ -225,6 +225,77 @@ enum SrgbFrameBufferKind {
     Khr,
 }
 
+pub struct AdapterContext {
+    glow_context: glow::Context,
+    egl: Arc<egl::DynamicInstance<egl::EGL1_4>>,
+    egl_display: egl::Display,
+    egl_context: egl::Context,
+    egl_pbuffer: Option<egl::Surface>,
+    mutex: Mutex<()>,
```
kvark, line 234: why not `Mutex<glow::Context>`?

wgpu-hal/src/gles/egl.rs
```diff
@@ -225,6 +225,77 @@ enum SrgbFrameBufferKind {
...
+    fn drop(&mut self) {
+        // Make the EGL context *not* current on this thread
+        self.egl
+            .make_current(self.egl_display, self.egl_pbuffer, self.egl_pbuffer, None)
+            .expect("Cannot make EGL context ");
+    }
+}
+
+impl AdapterContext {
+    pub fn get_without_lock<'a>(&'a self) -> &'a glow::Context {
```
kvark, line 266: these kind of lifetimes can be elided

kvark, line 266: as we don't want this to be used widely even within the backend, let's make it `unsafe`

wgpu-hal/src/gles/egl.rs
```diff
@@ -225,6 +225,77 @@ enum SrgbFrameBufferKind {
...
+        &self.glow_context
+    }
+
+    #[track_caller]
+    pub fn lock<'a>(&'a self) -> AdapterContextLock<'a> {
+        let _mutex_guard = self
+            .mutex
+            // Don't lock forever. If it takes longer than 1 second to get the lock we've got a
+            // deadlock and should panic to show where we got stuck
+            .try_lock_for(std::time::Duration::from_secs(1))
```
kvark, line 276: let's declare a constant at the top of the module

kvark (commented):
I like this approach quite a bit.
Previously in gfx, we had a wrapper around GL content that would just make it current without any locks. It was quite painful.
Here, we are locking the context and guaranteeing safe access.

wgpu-hal/src/gles/egl.rs
zicklag: Ah, that allows us to still lock the `glow::Context` even when calling `get_without_lock()` which will help a little more to protect against accidental concurrent access. Cool. :)

zicklag:
> Would having thread-local pbuffers work?

I _think_ so. 

> I tested this PR on my intel/haswell machine using the GL tests, and got tons of errors:

Does that machine support surfaceless EGL? If it doesn't, then it's probably the pbfuffer problem. I'll try out thread local pbuffers and then you could test it to see if that fixes it.

> cargo test --example hello-compute test_multithreaded_compute

That one fails on master with the expected "DeviceLost" that you get when creating a buffer on a different thread. It still fails with this PR, but with a different message. Not sure why:

```
---- tests::test_multithreaded_compute stdout ----
thread 'tests::test_multithreaded_compute' panicked at 'UNEXPECTED TEST PASS: BACKEND', wgpu/examples/hello-compute/../../tests/common/mod.rs:299:9
```

cwfitzgerald:
> Does that machine support surfaceless EGL?

It definitely should considering it's running on mesa. Is there a particular extension I should be looking out for? 

> That one fails on master with the expected "DeviceLost" that you get when creating a buffer on a different thread. It still fails with this PR, but with a different message. Not sure why:

We had a test that used to fail on vulkan but now succeeds but we never marked the test to pass. #1731 fixed it.

zicklag:
> Is there a particular extension I should be looking out for?

Yeah, in the info log you should see `EGL context: +surfaceless`:

image

zicklag:
> We had a test that used to fail on vulkan but now succeeds but we never marked the test to pass. #1731 fixed it.

For some reason the test still fails after rebasing.

```
$ WGPU_BACKEND=gl cargo test --example hello-compute test_multithreaded_compute
running 1 test
test tests::test_multithreaded_compute ... FAILED

failures:

---- tests::test_multithreaded_compute stdout ----
thread 'tests::test_multithreaded_compute' panicked at 'UNEXPECTED TEST PASS: BACKEND', wgpu/examples/hello-compute/../../tests/common/mod.rs:299:9


failures:
    tests::test_multithreaded_compute
```

cwfitzgerald:
Oh I misread what was causing it. This is because it now passes on your machine but is still marked to fail on OpenGL. You should got to the test definition `wgpu/example/hello-compute/tests.rs` and un-mark it as failing.

> EGL context: +surfaceless

Yeah this doesn't support that, it only supports EGL 1.4. In fact all my GL devices are only EGL 1.4. That being said, it does support `EGL_KHR_surfaceless_context`, so we might need to use that.

zicklag:
> Oh I misread what was causing it. This is because it now passes on your machine but is still marked to fail on OpenGL. You should got to the test definition wgpu/example/hello-compute/tests.rs and un-mark it as failing.

Ah, I get it now. :+1:

> Yeah this doesn't support that, it only supports EGL 1.4. In fact all my GL devices are only EGL 1.4. That being said, it does support EGL_KHR_surfaceless_context, so we might need to use that.

Well if it supports the extension then we shouldn't have a problem. Are you not seeing the `+surfaceless` in the logs? Maybe there's something wrong with our check for the extension?

https://github.com/gfx-rs/wgpu/blob/9ce884c384b3df024a78a458f5e659ff4f718580/wgpu-hal/src/gles/egl.rs#L331-L343

cwfitzgerald:
Yeah the check is wrong. that should be && not ||.

zicklag:
Oh, duh. :man_facepalming: Well, it looks like we found a way to test the pbuffer workaround. :)

I'll fix that check and push an update, then it should work for you on your EGL 1.4 devices. 

Then I'll also experiment to see if I can get the pbuffer workaround working by adding `|| true` to that check so I can pretend I don't have the required version/extensions on my machine.

zicklag:
OK I fixed the surfaceless check and I also fixed the error you were getting earlier with "Got an EGLSurface but no EGLContext". When I made the context not current I was not supposed to bind the pbuffer as the current surface.

Also the pbuffer workaround works totally fine with multi-threading so we won't need thread-local pbuffers or anything. I was misinterpreting when we would need to bind the pbuffer. Since we only bind the pbuffer when we bind the context, everything is fine and it is only ever bound by one thread.

If this works on your hardware @cwfitzgerald, I think this should be ready ( pending re-review ).

wgpu-hal/src/gles/egl.rs
```diff
@@ -1,8 +1,11 @@
 use glow::HasContext;
-use parking_lot::Mutex;
+use parking_lot::{Mutex, MutexGuard};
 
 use std::{ffi::CStr, os::raw, ptr, sync::Arc};
 
+/// The amount of time to wait while trying to obtain a lock to the adapter context
+const CONTEXT_LOCK_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(1);
```
kvark, line 7: let's just define this in seconds?

kvark, line 7: let's be consistent with other `std` includes by including either `std::time` or `std::time::Duration`

wgpu-hal/src/gles/egl.rs
```diff
@@ -329,7 +416,7 @@ impl Inner {
         // Testing if context can be binded without surface
         // and creating dummy pbuffer surface if not.
         let pbuffer =
-            if version < (1, 5) || !display_extensions.contains("EGL_KHR_surfaceless_context") {
+            if version < (1, 5) && !display_extensions.contains("EGL_KHR_surfaceless_context") {
```
kvark, line 419: uh oh, I think this would look cleaner by reversing the condition entirely
nice that you are fixing this!

wgpu-hal/src/gles/queue.rs
```diff
@@ -933,25 +940,22 @@ impl crate::Queue<super::Api> for super::Queue {
         command_buffers: &[&super::CommandBuffer],
         signal_fence: Option<(&mut super::Fence, crate::FenceValue)>,
     ) -> Result<(), crate::DeviceError> {
-        self.reset_state();
+        let shared = self.shared.clone();
```
kvark, line 943: nit: generally speaking, `Arc::clone(&xx)` is preferable over the regular `xx.clone()`

zicklag:
Pushed an update!

kvark (approved):
just pending on @cwfitzgerald

cwfitzgerald:
Alright, so all tests pass as long as I pass in `--test-threads=1`. I think the problem may be stemming from having multiple adapters open at once. I need to test my raspberry pi, but I think that is an issue we can deal with in a follow up PR.

zicklag:
> Alright, so all tests pass as long as I pass in --test-threads=1. 

Interesting. All the tests pass for me, even without the `--test-threads=1`. I wonder what the difference is.

> I need to test my raspberry pi, but I think that is an issue we can deal with in a follow up PR.

:+1:

kvark:
uh, isn't this a bit worrying? The whole point of the PR is to support more than one thread.
Would you have energy to investigate? My time is getting low now since the move is nigh...

cwfitzgerald:
> The whole point of the PR is to support more than one thread.

It does, the multi-threaded example works perfectly fine, it's juggling multiple adapters which is the issue (which is what the test harness is doing without --test-threads=1). So I'm comfortable trying to deal with this in a follow up.

> Would you have energy to investigate?

Yeah ofc!

kvark:
ok, sounds reasonable
bors r=cwfitzgerald,kvark

zicklag:
Awesome! Thanks guys! I'm _super_ excited to actually have gotten Bevy running on OpenGL. :smiley:

cwfitzgerald:
Thank you for all the help getting it working!

PR #1737, opened 2021-07-27
Use Bitshift Style for All Bitflag Consts

zicklag:
**Connections**
None exactly, but somewhat related to https://github.com/gfx-rs/wgpu/pull/1686 as I was going to put some of this change into that PR, but am splitting it out for ease of review.

**Description**
This is purely code-style related, but it makes it easier to read bitflag constants and much eaiser to add bitflags in the future.

**Testing**
Tested on Ubuntu 20.04

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -175,7 +175,7 @@ bitflags::bitflags! {
         /// - some mobile chips
         ///
         /// This is a web and native feature.
-        const DEPTH_CLAMPING = 0x0000_0000_0000_0001;
+        const DEPTH_CLAMPING = 1 << 0;
         /// Enables BCn family of compressed textures. All BCn textures use 4x4 pixel blocks
         /// with 8 or 16 bytes per block.
         ///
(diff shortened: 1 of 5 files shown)


wgpu-types/src/lib.rs
```diff
@@ -234,7 +234,7 @@ bitflags::bitflags! {
         /// - All
         ///
         /// This is a native only feature.
-        const MAPPABLE_PRIMARY_BUFFERS = 0x0000_0000_0001_0000;
```
cwfitzgerald, line 237: Watch these, all native features are in a different space. Native ones are bits 16-63, webgpu ones are 0-15.

zicklag: Ah, good catch. Updated.

wgpu-types/src/lib.rs
```diff
@@ -459,14 +459,14 @@ bitflags::bitflags! {
...
-        const VERTEX_WRITABLE_STORAGE = 0x0000_0002_0000_0000;
+        const VERTEX_WRITABLE_STORAGE = 1 << 33;
         /// Enables clear to zero for buffers & images.
         ///
         /// Supported platforms:
         /// - All
         ///
         /// This is a native only feature.
-        const CLEAR_COMMANDS = 0x0000_0400_0000_0000;
+        const CLEAR_COMMANDS = 1 << 42;
```
cwfitzgerald, line 469: Jump from 33 to 42?

cwfitzgerald (approved):
Minor comment, but otherwise good

wgpu-types/src/lib.rs
zicklag: I was just copying what the existing flags were, I wasn't sure if there was a reason for the jump or not. If it's native only are these values not tied to the WebGPU spec?

cwfitzgerald: Yeah they aren't. I don't know why we would jump here. Likely just due to how difficult maintaining the old way was :)

zicklag: Ok, pushed an update to fix.

cwfitzgerald:
Thanks!

bors r+

PR #1745, opened 2021-07-29
Fix New Clippy Lints From Rust 1.54.0

zicklag:
**Connections**
None

**Description**
Fixes a number of new lints introduce in the updated clippy that CI is now using. Noticed that CI was failing because of the lints on another PR after the release of Rust 1.54.0.

**Testing**


cwfitzgerald (approved):
CI mad, but looks fine

zicklag:
Yeah, still catching up with it apparently. I'll keep pushing updates 'till it's happy. :)

PR #1749, opened 2021-08-01
update link to WebGPU wiki

euclio:
**Description**

The current link links to the wiki page in the archived repository, not the new one.

There is another link in this README that links to the old "Applications and Libraries" wiki page, but this repository's wiki doesn't have an equivalent, so I left it unchanged.


cwfitzgerald:
Thanks!

cwfitzgerald (approved):

cwfitzgerald:
The new one is called https://github.com/gfx-rs/wgpu/wiki/Users if you want to follow up with that.

PR #1753, opened 2021-08-01
update link to wgpu users wiki

euclio:
**Description**
Use the new wiki page instead of linking to the archived repository wiki.

Followup to #1749.


cwfitzgerald (approved):
Thanks!

PR #1754, opened 2021-08-01
Use Bytemuck to Cast Hello Compute Output

zicklag:
**Connections**

**Description**
Just a slight tweak to the codestyle of the hello-compute example. This makes the cast of the output data in the hello-compute example easier to read and makes it match the way we cast the input data too.

**Testing**
Tested compute example on Linux vulkan.


cwfitzgerald (approved):
Thanks!

PR #1758, opened 2021-08-03
Fix error with resolver=1 on non-Windows

kvark:
**Connections**

**Description**
Many users stumble upon this.

**Testing**
Manually tested

### DIFF
--- a/wgpu-hal/Cargo.toml
+++ b/wgpu-hal/Cargo.toml
@@ -47,7 +47,6 @@ glow = { git = "https://github.com/grovesNL/glow", rev = "0864897a28bbdd43f89f4f
 
 # backend: Dx12
 bit-set = { version = "0.5", optional = true }
-native = { package = "d3d12", git = "https://github.com/gfx-rs/d3d12-rs", rev = "79f29c8", features = ["libloading"], optional = true }
 range-alloc = { version = "0.1", optional = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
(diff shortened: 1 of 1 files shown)


kvark:
The new error:
```
error: DX12 API enabled on non-Windows OS. If your project is not using resolver="2" in Cargo.toml, it should.
  --> /Users/dmalyshau/Code/wgpu/wgpu-hal/src/lib.rs:47:1
   |
47 | compile_error!("DX12 API enabled on non-Windows OS. If your project is not using resolver=\"2\" in Cargo.toml, it should.");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0432]: unresolved import `super::dx12`
  --> /Users/dmalyshau/Code/wgpu/wgpu-hal/src/lib.rs:62:20
   |
62 |     pub use super::dx12::Api as Dx12;
   |                    ^^^^ could not find `dx12` in the crate root

error: aborting due to 2 previous errors

```

cwfitzgerald:
This pr is fine, but it won't actually be hit, d3d12-rs fails to compile first

kvark:
> This pr is fine, but it won't actually be hit, d3d12-rs fails to compile first

the subject of the issue was wrong. It's fixing resolver=1 on *non* windows.
d3d12-rs isn't compiled now because the PR moves it to windows-only dependencies

PR #1424, opened 2021-06-02
Reproduce Naga's nicer error reporting for WGSL

almarklein:
Closes https://github.com/gfx-rs/wgpu-native/issues/104

Basically, I find myself copying WGSL code to a file and running Naga from the command line on it, because that way I get prettier errors than when pushing the (faulty) WGSL via wgpu-native's  `create_shader_module`. Would be nice if the error reports were equally good.

WIP: does not seem to help yet, except duplicating part of the error message.

PR #1760, opened 2021-08-03
Include naga parse error in the error chain

scoopr:
**Connections**
Probably makes #1424 moot

**Description**
Wraps the Naga `ParseError` with `NagaParseError` that shows the full Naga diagnostic in its `Display impl` and adding it to the `CreateShaderModuleError::Parsing` variant, allows for the wgsl diagnostics to be visible in the displayed validation error.

**Testing**
With modifying examples

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -895,7 +895,12 @@ impl<A: HalApi> Device<A> {
                     Ok(module) => module,
                     Err(err) => {
                         log::error!("Failed to parse WGSL code for {:?}: {}", desc.label, err);
-                        return Err(pipeline::CreateShaderModuleError::Parsing);
+                        return Err(pipeline::CreateShaderModuleError::Parsing(
+                            pipeline::NagaParseError {
+                                shader_source: code.to_string(),
+                                error: err,
+                            },
+                        ));
                     }
                 }
             }
--- a/wgpu-core/src/pipeline.rs
+++ b/wgpu-core/src/pipeline.rs
@@ -45,9 +45,24 @@ impl<A: hal::Api> Resource for ShaderModule<A> {
 }
 
 #[derive(Clone, Debug, Error)]
+pub struct NagaParseError {
+    pub shader_source: String,
+    pub error: naga::front::wgsl::ParseError,
+}
+impl std::fmt::Display for NagaParseError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "\nShader error:\n{}",
+            self.error.emit_to_string(&self.shader_source)
+        )
+    }
+}
+
+#[derive(Clone, Debug, Error)]
 pub enum CreateShaderModuleError {
     #[error("Failed to parse a shader")]
-    Parsing,
+    Parsing(#[from] NagaParseError),
     #[error("Failed to generate the backend-specific code")]
     Generation,
     #[error(transparent)]


kvark (approved):
Nice!

almarklein:
I'm really grateful for this change, but one issue I'm facing is that newlines are escaped in the result. An example:

``` 
λ python examples\geometry_cube.py
Failed to parse WGSL code for Some(""): no definition in scope for identifier: 'atomicExchange'
thread '<unnamed>' panicked at 'Parsing(NagaParseError { shader_source: "\n        [[block]]\n        struct Struct_u_stdinfo {\n            cam_transform: mat4x4<f32>;\n            cam_transform_inv: mat4x4<f32>;\n            projection_transform: mat4x4<f32>;\n            projection_transform_inv: mat4x4<f32>;\n            physical_size: vec2<f32>;\n            logical_size: vec2<f32>;\n        };\n\n        [[group(0), binding(0)]]\n        var u_stdinfo: Struct_u_stdinfo;\n        \n\n        [[block]]\n        struct Struct_u_wobject {\n            world_transform: mat4x4<f32>;\n            id: i32;\n        };\n\n        [[group(0), binding(1)]]\n        var u_wobject: Struct_u_wobject;\n        \n\n        [[block]]\n        struct [... etc.]
```

Is this intentional or could we change something to fix this? I reckon it should be a minor change, but I don't know enough about Rust strings to know what that change should be.

kvark:
Interesting. This is definitely not what we see. Where does this `unwrap` happens? It's not in `wgpu-core`, it must be in wgpu-native itself. Should look like this - https://github.com/gfx-rs/wgpu/blob/0e65b9323a2e872367dace0685a7b3278d7a9678/wgpu/src/backend/direct.rs#L2156

almarklein:
Would that be where [create_shader_module is called](https://github.com/gfx-rs/wgpu-native/blob/b10496e7eed9349f0fd541e6dfe5029cb436de74/src/device.rs#L111-L114)? (and in [check_error here](https://github.com/gfx-rs/wgpu-native/blob/c85b882208c64d656d68f36c3b0f710139ab9b44/src/lib.rs#L50-L57)

kvark:
Right. It needs to be `panic!("{}", error);` instead of `panic!("{:?}", error);`

almarklein:
Then I get:
```
λ cargo build
   Compiling wgpu-native v0.6.0 (C:\dev\rust\wgpu-native)
error[E0277]: `E` doesn't implement `std::fmt::Display`
  --> src\lib.rs:53:22
   |
53 |         panic!("{}", error);
   |                      ^^^^^ `E` cannot be formatted with the default formatter
   |
   = note: in format strings you may be able to use `{:?}` (or {:#?} for pretty-print) instead
   = note: required by `std::fmt::Display::fmt`
   = note: this error originates in the macro `$crate::format_args` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider further restricting this bound
   |
51 | pub fn check_error<I, E: std::fmt::Debug + std::fmt::Display>(input: (I, Option<E>)) -> I {
   |                                          ^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.
error: could not compile `wgpu-native`
```

kvark:
Could you try `E: std::error::Error` instead?

almarklein:
Then I can use `panic!("{}", error);` but then also the text written is reduced to only the error title (what we had before this pr),

kvark:
Oh, I see now. In wgpu-rs, between the wgpu-core call and the user error report, there is also this piece happening - https://github.com/gfx-rs/wgpu/blob/0e65b9323a2e872367dace0685a7b3278d7a9678/wgpu/src/backend/direct.rs#L146-L169
And that's not a part of wgpu-native atm. Would you mind filing an issue to wgpu-native to follow-up on this?

PR #1764, opened 2021-08-04
Implement onSubmittedWorkDone

### LINKED ISSUE #1498: Create an API for managing fences
It's currently not possible to determine when the GPU has finished submitted work. This can pose a problem when one needs to, for example, resize the swapchain without triggering a panic. It can also pose a problem when rendering to a Texture, as there is no way to determine when the work is done.

Similar to how Textures or Buffers are managed, fences ought to be a managed resource too.

It may be possible to use a `Future` returned from the submission of the command buffer. However I don't think this will have the desired performance characteristics. A future being returned would require wgpu to do one of two things

1. Hold a collection of blocked threads internally which would be responsible for notifying the waker.
2. Frequently check the internal fences and notify the corresponding waker when a fence is found in a signaled state.

The first solution would result in more threads than the application strictly needs, and the second solution runs the risk of introducing a significant delay between the fence signal and the future being resolved. So it seems the best solution would be to pass this choice onto the user. They can either block a thread, or frequently check the fence state, depending on what performance characteristics they want.

**Additional context**
I am using this information in the construction of a multi-stage pipeline with support for multiple frames in-flight. I must build the feature described here. The decision for wgpu is only if they wish to accept my PR.

kvark:
**Connections**
Closes #1498

**Description**
Implements the upstream entry point that replaced fences.

**Testing**
manually tested
It looks like testing this properly requires either us implementing `Future` for something temporary, or hooking up `futures`. Neither is exciting.

### DIFF
--- a/wgpu-core/src/device/life.rs
+++ b/wgpu-core/src/device/life.rs
@@ -2,7 +2,7 @@
 use crate::device::trace;
 use crate::{
     device::{
-        queue::{EncoderInFlight, TempResource},
+        queue::{EncoderInFlight, SubmittedWorkDoneClosure, TempResource},
         DeviceError,
     },
     hub::{GlobalIdentityHandlerFactory, HalApi, Hub, Token},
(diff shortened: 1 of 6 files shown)


cwfitzgerald (approved):

cwfitzgerald:
Old habits die hard eh?

kvark:
ouch

PR #1766, opened 2021-08-04
Update Naga with new storage classes API

zicklag:
**Connections**
None

**Description**
Updates Naga. Needed by me specifically to get https://github.com/gfx-rs/naga/pull/1144 pulled in to fix OpenGL backend with Bevy.

I had to remove the `[[access(read)]]` the shader for the shadow example or else I would get this error:

```
Caused by:
        Failed to parse a shader
        
    Shader error:
    error: expected type attribute ('access' or 'stride') or an end of the attribute list (']]'), found 'access'
       ┌─ wgsl:58:26
       │
    58 │ var<storage> s_lights: [[access(read)]] Lights;
       │                          ^^^^^^ expected type attribute ('access' or 'stride') or an end of the attribute list (']]')
```

Is this a naga bug, maybe? Seems a little off to me.

Also, I'm not exactly sure how the new struct element in the Naga `StorageClass` enum is meant to be used so let me know if any of that is wrong!

**Testing**
Tested shadow example on linux with GL and vulkan.

### DIFF
--- a/wgpu-hal/src/gles/device.rs
+++ b/wgpu-hal/src/gles/device.rs
@@ -28,7 +28,7 @@ impl CompilationContext<'_> {
             }
             let register = match var.class {
                 naga::StorageClass::Uniform => super::BindingRegister::UniformBuffers,
-                naga::StorageClass::Storage => super::BindingRegister::StorageBuffers,
+                naga::StorageClass::Storage { .. } => super::BindingRegister::StorageBuffers,
                 _ => continue,
             };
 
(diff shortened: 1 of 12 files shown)


zicklag:
I found the new syntax for storage access and update the PR to use that. It wasn't a Naga bug, but I did open a PR to make the naga error message less confusing: https://github.com/gfx-rs/naga/pull/1167.

wgpu-hal/src/gles/device.rs
```diff
@@ -48,7 +48,12 @@ impl CompilationContext<'_> {
 
         for (name, mapping) in reflection_info.texture_mapping {
             let var = &module.global_variables[mapping.texture];
-            let register = if var.storage_access.is_empty() {
+            let empty_storage_access = if let naga::StorageClass::Storage { access } = var.class {
```
kvark, line 51: This needs to be rewritten now. We are checking whether this image is a texture or a storage image. The storage class is never going to be `StorageClass::Storage` - this is used for buffers, not textures. So we need to do something like this:
```rust
let register = match module.types[var.ty].inner {
  naga::TypeInner::Image { class: naga::ImageClass::Storage { .. } } => super::BindingRegister::Images,
  _ => super::BindingRegister::Textures,
};
```

kvark (commented):
Thank you!
Just one small fix is needed

zicklag:
Updated with your code sample, fixed a broken test, and updated the Naga commit hash to the latest master.

zicklag:
Looks like CI is failing on windows for lack of disk space.

kvark (approved):

PR #1770, opened 2021-08-06
Do not error on zero-sized scissor commands.

### LINKED ISSUE #1750: Valid ScissorRect Values are Rejected by set_scissor_rect
**Description**
ScissorRect commands with width or height values of 0 are rejected resulting in an `RenderCommandError::InvalidScissorRect` error.

**Repro steps**
Insert a `set_scissor_rect` call into the `cube.rs` demo with zero width or height, e.g.:

```rust
rpass.push_debug_group("Prepare data for draw.");
rpass.set_pipeline(&self.pipeline);
rpass.set_bind_group(0, &self.bind_group, &[]);
rpass.set_index_buffer(self.index_buf.slice(..), wgpu::IndexFormat::Uint16);
rpass.set_vertex_buffer(0, self.vertex_buf.slice(..));
rpass.pop_debug_group();
rpass.insert_debug_marker("Draw!");
rpass.set_scissor_rect(0, 0, 0, 0); // <---- ADDED
rpass.draw_indexed(0..self.index_count as u32, 0, 0..1);

```

The error is caused by this check in `render.rs`:
```rust
if rect.w == 0 // <---
    || rect.h == 0 // <---
    || rect.x + rect.w > info.extent.width
    || rect.y + rect.h > info.extent.height
{
    return Err(RenderCommandError::InvalidScissorRect).map_pass_err(scope);
}
```

wgpu should not reject these values and should not crash. The Vulkan specification [1] says only that the `offset + extent` cannot overflow, and the OpenGL documentation [2] only says the width/height cannot be negative.

**Extra materials**

**Platform**
- Ubuntu 20.04
- wgpu 0.9
- Tested on Vulkan backend

[1]: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCmdSetScissor.html
[2]: https://www.khronos.org/registry/OpenGL-Refpages/es2.0/xhtml/glScissor.xml

kylc:
**Connections**
Fixes #1750.

**Description**
wgpu currently throws an error if either dimension of the scissor rectangle is zero. It seems that the WebGPU spec was the holdout that caused this requirement, but the spec has since been revised to allow for zero size scissor rectangles (https://github.com/gpuweb/gpuweb/pull/1138/files).

The spec now just says:
>    * x+width is less than or equal to this.[[attachment_size]].width.
>    * y+height is less than or equal to this.[[attachment_size]].height.

This PR removes the failure condition.

**Testing**
Tested on Ubuntu 20.04 with imgui-rs/imgui-wgpu-rs projects with the Vulkan backend and verified that crashes with zero sized scissor rectangles no longer occur.

### DIFF
--- a/wgpu-core/src/command/render.rs
+++ b/wgpu-core/src/command/render.rs
@@ -1279,9 +1279,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                     }
                     RenderCommand::SetScissor(ref rect) => {
                         let scope = PassErrorScope::SetScissorRect;
-                        if rect.w == 0
-                            || rect.h == 0
-                            || rect.x + rect.w > info.extent.width
+                        if rect.x + rect.w > info.extent.width
                             || rect.y + rect.h > info.extent.height
                         {
                             return Err(RenderCommandError::InvalidScissorRect).map_pass_err(scope);


wgpu-core/src/command/render.rs
```diff
@@ -1279,9 +1279,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                     }
                     RenderCommand::SetScissor(ref rect) => {
                         let scope = PassErrorScope::SetScissorRect;
-                        if rect.w == 0
-                            || rect.h == 0
-                            || rect.x + rect.w > info.extent.width
+                        if rect.x + rect.w > info.extent.width
```
kvark, line 1282: Unfortunately, this would make the Metal backend to fail. In order to remove this condition, we need to implement a small workaround in Metal backend to set the viewport to something dummy, as a replacement for zero scissor.

Issue #1771, opened 2021-08-06
Leaking scissor test when presenting gles backend surface


ElArtista:
**Description**
Scissor test leaking into the final blit call executed when presenting the surface in the gles backend here:
https://github.com/gfx-rs/wgpu/blob/master/wgpu-hal/src/gles/egl.rs#L848

**Extra materials**
Behavior noticed when running the egui example  (modified for wgpu master) from here:
https://github.com/hasenbanck/egui_example
![image](https://user-images.githubusercontent.com/4622346/128465543-cf17c83e-57c2-4149-8d67-88abafe38144.png)

PR #1772, opened 2021-08-06
Fix scissor test leak when presenting gles backend

ElArtista:
**Connections**
Fixes: #1771

### DIFF
--- a/wgpu-hal/src/gles/egl.rs
+++ b/wgpu-hal/src/gles/egl.rs
@@ -840,6 +840,8 @@ impl Surface {
                 crate::SurfaceError::Lost
             })?;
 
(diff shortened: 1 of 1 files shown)


wgpu-hal/src/gles/egl.rs
```diff
@@ -840,6 +840,8 @@ impl Surface {
                 crate::SurfaceError::Lost
             })?;
 
+        gl.disable(glow::SCISSOR_TEST);
```
kvark, line 843: let's also reset the color mask, which could show the same issue

ElArtista: Added!

kvark (approved):

PR #1773, opened 2021-08-06
Fix dynamic offset handling for render bundles

pythonesque:
I found this while testing a "bundlemark" version of bunnymark.

**Connections**

**Description**

**Testing**


cwfitzgerald (approved):
Thanks!

kvark (commented):
oh wow

PR #1774, opened 2021-08-06
Add downlevel flag for storage buffers in vertex shaders

ElArtista:
**Description**
This will enable compatibility on systems that report support for storage buffers on fragment shaders but not on  vertex shaders.
This includes all Radeon Terrascale architecture GPUs.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -1104,10 +1104,22 @@ impl<A: HalApi> Device<A> {
                         error,
                     })?;
             }
-            if writable_storage == WritableStorage::Yes
-                && entry.visibility.contains(wgt::ShaderStages::VERTEX)
-            {
-                required_features |= wgt::Features::VERTEX_WRITABLE_STORAGE;
+            if entry.visibility.contains(wgt::ShaderStages::VERTEX) {
+                if writable_storage == WritableStorage::Yes {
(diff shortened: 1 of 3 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -1104,10 +1104,22 @@ impl<A: HalApi> Device<A> {
...
+            if entry.visibility.contains(wgt::ShaderStages::VERTEX) {
+                if writable_storage == WritableStorage::Yes {
+                    required_features |= wgt::Features::VERTEX_WRITABLE_STORAGE;
+                }
+                if let Bt::Buffer {
+                    ty: wgt::BufferBindingType::Storage { .. },
+                    ..
+                } = entry.ty
+                {
+                    self.require_downlevel_flags(wgt::DownlevelFlags::VERTEX_STORAGE)
```
cwfitzgerald, line 1116: Could we do something similar as the required_features, enumerate them then require them once at the bottom.

cwfitzgerald (changes requested):
One issue, nut otherwise lgtm

wgpu-core/src/device/mod.rs
ElArtista: Done as requested!

cwfitzgerald (approved):
Thanks!

PR #1780, opened 2021-08-09
Merge swap chain into surface

### LINKED ISSUE #1499: Merge Surface and Swapchain
See https://github.com/gpuweb/gpuweb/pull/1810

kvark:
**Connections**
Closes #1499

**Description**
Note: this breaks the Web backend. Hopefully temporarily.

**Testing**
Traced/Replayed the examples.


wgpu-core/src/present.rs
```diff
@@ -103,17 +73,17 @@ pub enum CreateSwapChainError {
...
 #[derive(Debug)]
-pub struct SwapChainOutput {
+pub struct SurfaceOutput {
     pub status: Status,
     pub view_id: Option<TextureViewId>,
 }
 
 impl<G: GlobalIdentityHandlerFactory> Global<G> {
-    pub fn swap_chain_get_current_texture_view<A: HalApi>(
+    pub fn surface_get_current_texture_view<A: HalApi>(
```
pythonesque, line 82: Shouldn't this return a texture, not a texture view, as we discussed (per https://www.w3.org/TR/webgpu/#dom-gpucanvascontext-getcurrenttexture)?  Then you can get rid of the `TextureViewSource::Surface` variant (and `TextureViewSource` more generally)--and consequently also get rid of the rogue texture view destruction stuff.  I think we also determined there's no actual reason the texture needs to know it came from a swapchain, so this doesn't have to get replaced with something on `Texture`.

kvark: Yes, it's just a different piece of work, fairly big one, so I didn't want to mix them in the same change. I'm working on it.

cwfitzgerald (approved):
By proxy of #1786 being good, this one is good too :)(

kvark:
Closed by #1786

PR #1785, opened 2021-08-09
Split non-uniform indexing into two different features

cwfitzgerald:
**Connections**

Follows up to #1522 with a combination that was missed.

**Description**

Intel/Windows supports SampledTexture and StorageBuffer non-uniform indexing, but not the rest.

**Testing**

wgpu-info and rend3.

### DIFF
--- a/wgpu-hal/src/metal/adapter.rs
+++ b/wgpu-hal/src/metal/adapter.rs
@@ -864,7 +864,9 @@ impl super::PrivateCapabilities {
             | F::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES;
 
         features.set(
-            F::TEXTURE_BINDING_ARRAY | F::RESOURCE_BINDING_ARRAY_NON_UNIFORM_INDEXING,
+            F::TEXTURE_BINDING_ARRAY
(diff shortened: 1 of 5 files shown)


kvark:
Do we need something prettier and more generic here?
```rust
struct ResourceIndexing {
  Forbidden,
  Uniform,
  NonUniform,
}

struct ResourceIndexingFeatures {
  uniform_buffers: ResourceIndexing,
  storage_buffers: ResourceIndexing,
  textures: ResourceIndexing,
  storage_textures: ResourceIndexing,
}
```

cwfitzgerald:
That definitely would be a lot cleaner, we just currently don't have an endpoint for this. Should we make the bitflags `FeatureFlags` and have the top level Features be a struct similar to the downlevel features?

kvark:
I want to say yes, but I'm also a bit worried that this would make us diverge from the upstream API in the future where the upstream gets similar features :/

kvark:
Actually, it's probably going to be a year or so till the upstream gets this, so we should just pick what we think is reasonable.
Do you feel positive about such a change? If yes, let's try it. If no, we can land the PR.

cwfitzgerald:
We're bors less now

kvark (approved):

PR #1786, opened 2021-08-10
Surface returning textures instead of views

kvark:
**Connections**
Follow-up to #1780
https://github.com/webgpu-native/webgpu-headers/issues/89

**Description**
Aligns to both wgpu-hal and WebGPU upstream API better.

**Testing**
Examples


kvark:
One blocker here is figuring out when/how to properly encode the pipeline barrier of a surface texture that transitions it into `PRESENT` state. Previous approach was limited and essentially a hack (derive the barriers from load/store ops).
I'm thinking that the surface texture isn't going to be globally tracked by the device, and instead it will be considered to always be in PRESENT state at the end of *any* command buffer.

kvark:
Should be ready now. Tested on Metal and Vulkan, no validation issues.

cwfitzgerald (approved):
Nothing in here stands out to me, as weird!

LGTM!

PR #1795, opened 2021-08-13
Remove caching from the main action

kvark:
**Connections**
Seeing how much time is spent in writing caches for #1794

**Description**
Optimize our build times. Maybe the caching isn't worth it?

**Testing**
Let's see the CI log!


kvark:
CI times appear to be around the same but without really bad peaks:
- Windows: 19m -> 11m
- MacOS: 3m -> 6m
- iOS: 6m -> 4m
- Linux: 10m -> 10m
- Android: 2m -> 3m

PR #1810, opened 2021-08-17
Run tests in the CI for all workspace members

Imberflur:
**Connections**
None known

**Description**
The CI was not running unit tests on some of the crates. This PR expands this by replacing `-p player` with `--workspace --exclude wgpu`.

Note: I also added `--no-fail-fast` for these.

**Testing**
I ran the new test command locally and the CI should check to make sure it works.


Imberflur:
It looks like one of the doc tests may be failing to compile but this isn't triggering the CI to fail. https://github.com/gfx-rs/wgpu/pull/1810/checks?check_run_id=3347308424#step:10:477

kvark (approved):
thank you!

PR #1814, opened 2021-08-17
write_texture data offset

kocsis1david:
**Connections**
None

**Description**
When using the `write_texture` function, the `offset` parameter in `ImageDataLayout` was ignored.

**Testing**
I've run the cube example.


cwfitzgerald (changes requested):
Code looks fine. I'm not sure our validation is properly adjusting for the data offset though. Can you double check that the validation will properly fail if you go out of bounds on the buffer? We're using ptr::copy_nonoverlapping which is very unsafe if we miss.

kocsis1david:
The validation is good, it panics if you specify a wrong offset.

cwfitzgerald (approved):
Thanks!

PR #1819, opened 2021-08-18
Implement compressed textures in DX12

cwfitzgerald:
**Connections**

This simply implements compressed textures for DX12

**Description**

Just needed to properly upround all of sizes to the block size.

**Testing**

Skybox BC1 test now passes

### DIFF
--- a/wgpu-hal/src/dx12/command.rs
+++ b/wgpu-hal/src/dx12/command.rs
@@ -13,6 +13,18 @@ fn make_box(origin: &wgt::Origin3d, size: &crate::CopyExtent) -> d3d12::D3D12_BO
     }
 }
 
+fn upround_extent(size: crate::CopyExtent, block_size: u32) -> crate::CopyExtent {
+    debug_assert!(block_size.is_power_of_two());
+
+    let block_mask = block_size - 1;
+
+    crate::CopyExtent {
+        width: (size.width + block_mask) & !(block_mask),
+        height: (size.height + block_mask) & !(block_mask),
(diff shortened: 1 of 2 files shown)


wgpu-hal/src/dx12/command.rs
```diff
@@ -418,18 +433,21 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
         };
         let raw_format = conv::map_texture_format(dst.format);
 
+        let block_size = dst.format.describe().block_dimensions.0 as u32;
         for r in regions {
-            let src_box = make_box(&wgt::Origin3d::ZERO, &r.size);
+            let uprounded_size = upround_extent(r.size, block_size);
```
kvark, line 438: Something is not right here. WebGPU spec [requires](https://gpuweb.github.io/gpuweb/#valid-texture-copy-range) that copy size is already a multiple of the block size:
> copySize.width must be a multiple of blockWidth.
> copySize.height must be a multiple of blockHeight.

So if it's already a multiple, we shouldn't be uprounding it.

kvark (changes requested):
This is great to see! There is a bit of a question here

wgpu-hal/src/dx12/command.rs
kvark: Discussed on the matrix, we'll need to rehaul this for all backends as a follow-up

kvark (approved):

kvark:
Published in wgpu-hal-0.10.3

PR #1822, opened 2021-08-18
Fix repo size and crates.io logo

cwfitzgerald:
**Connections**

Follow up to #1820 

**Description**

Shrink!

**Testing**

Eyes

### DIFF
--- a/wgpu/Cargo.toml
+++ b/wgpu/Cargo.toml
@@ -5,10 +5,20 @@ authors = ["wgpu developers"]
 edition = "2018"
 description = "Rusty WebGPU API wrapper"
 homepage = "https://wgpu.rs/"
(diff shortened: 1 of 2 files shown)


wgpu/Cargo.toml
```diff
@@ -8,7 +8,18 @@ homepage = "https://wgpu.rs/"
 repository = "https://github.com/gfx-rs/wgpu"
```
kvark, line 8: can we also change this to point to v0.10 branch (which will be created)?
people often get confused about this, so it would be nice to have

kvark (commented):
One small wish!

kvark (approved):

kvark:
thank you!

PR #1831, opened 2021-08-19
[metal] set threadgroup memory sizes reflected from the shader

### LINKED ISSUE #1823: [metal] Using workgroup variable causes what is probably a race condition in compute pipelines.
Run the `hello-compute` example on macOS on Intel HD630 with:
```diff
--- a/wgpu/examples/hello-compute/shader.wgsl
+++ b/wgpu/examples/hello-compute/shader.wgsl
@@ -5,6 +5,7 @@ struct PrimeIndices {
 
 [[group(0), binding(0)]]
 var<storage, read_write> v_indices: PrimeIndices;
+var<workgroup> foo: u32;
 
 // The Collatz Conjecture states that for any integer n:
 // If n is even, n = n/2
@@ -13,6 +14,7 @@ var<storage, read_write> v_indices: PrimeIndices;
 // Though the conjecture has not been proven, no counterexample has ever been found.
 // This function returns how many times this recurrence needs to be applied to reach 1.
 fn collatz_iterations(n_base: u32) -> u32{
+    foo = 0u;
     var n: u32 = n_base;
     var i: u32 = 0u;
     loop {
```

Most of the times it will run correctly and output `[0, 1, 7, 2]`, but, if you run it in a loop (e.g. `for i in {1..1000}; do cargo run --example hello-compute; done`) it will eventually output `[1, 2, 3, 4]`. This behavior seems to be easier to cause with big inputs.

Also, running it without the `workgroup` variable seems to work fine.

kvark:
**Connections**
Fixes #1823

**Description**
See https://github.com/gfx-rs/wgpu/issues/1823#issuecomment-902011447

**Testing**
Tested on a modified hello-compute sample, manually

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## TBD
+  - All:
+    - expose more formats via adapter-specific feature
+  - Metal:
+    - fix usage of work group memory
+
 ## v0.10 (2021-08-18)
   - Infrastructure:
     - `gfx-hal` is replaced by the in-house graphics abstraction `wgpu-hal`. Backends: Vulkan, Metal, D3D-12, and OpenGL ES-3.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -909,9 +909,9 @@ dependencies = [
 
 [[package]]
 name = "metal"
-version = "0.23.0"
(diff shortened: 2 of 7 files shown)


kvark:
Included in wgpu-hal-0.10.2

PR #1836, opened 2021-08-20
Fix the check for depth+stencil aspect of views

### LINKED ISSUE #1835: Read-only Depth/Stencil (RODS) in the water example fails on using Depth24PlusStencil8 instead of Depth32Float
**Description**
In a project, I used a Depth24PlusStencil8 buffer as depth/stencil-buffer target in a first pass and read from that same texturebuffer in a second pass. With the transition from Wgpu 0.9.0 to 0.10.1, it stopped working, mentioning: "bound texture views can not have both depth and stencil aspects enabled", I found that the "water" example in this wgpu repository does the same by approximation and that uses a Depth32Float buffer instead and it works. But I need to write to the stencil part in the first pass because I use its output in a third pass. So in a local wgpu clone, I replaced all Depth32Float by Depth24PlusStencil8 in the water example *) and modified the TextureViewDescriptor on main.rs::228 to have aspect: wgpu::TextureAspect::DepthOnly and presto: "bound texture views can not have both depth and stencil aspects enabled". I have added the example main.rs as a zip in this report below.

*) I also removed the wgpu::TextureUsages::COPY_DST on line 214 because it interferes and the example-with-Depth32Float runs fine without it.

**Repro steps**
See the description

I get the message "bound texture views can not have both depth and stencil aspects enabled" in "device.create_bind_group" from water::Example::initialize_resources at .\examples\water\main.rs:231. I expect the example to run fine because the stencil bits from the depth/stencil buffer are not used anywhere and I have set the texture view aspect to DepthOnly. It seems that wgpu internally uses Depth+stencil anyhow.

**Extra materials**
[main.zip](https://github.com/gfx-rs/wgpu/files/7021308/main.zip)

**Platform**
Information about your OS, version of `wgpu`, your tech stack, etc.
win10 / Vulkan / wgpu 0.10.1

kvark:
**Connections**
Fixes #1835

**Description**
The check didn't take into account `wgt::TextureAspect` enum of the view.

**Testing**
Tested on the modified water example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## TBD
   - All:
     - expose more formats via adapter-specific feature
+    - fix creation of depth+stencil views
   - Metal:
     - fix usage of work group memory
 
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -1568,7 +1568,9 @@ impl<A: HalApi> Device<A> {
         expected: &'static str,
     ) -> Result<(wgt::TextureUsages, hal::TextureUses), binding_model::CreateBindGroupError> {
         use crate::binding_model::CreateBindGroupError as Error;
-        if hal::FormatAspects::from(view.desc.format)
+        if view
+            .desc
+            .aspects()
             .contains(hal::FormatAspects::DEPTH | hal::FormatAspects::STENCIL)
         {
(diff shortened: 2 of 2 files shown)


cwfitzgerald (approved):

kvark:
Included in wgpu-core-0.10.1

PR #1840, opened 2021-08-21
fix buffer collapse in initialize_buffer_memory

Wumpf:
**Connections**
Other memory init work #1839  #1688 

**Description**

Due to previous changes there was  only a no-op buffer collapse
The ranges it tried collapsing on were already guranteed to be merged by the init-trackers drain method. However, accross several init actions there can be opportunity for reducing fill_buffer calls.

I'll probably follow up with making `buffer_memory_init_actions` a hashmap to begin with, so that in `initialize_buffer_memory` we don't need to do the bucketing and can look at the problem one buffer at a time. This will have the nice side effect being able to drop the id from BufferInitAction!
But for that #1839 needs to land first!

### DIFF
--- a/wgpu-core/src/command/mod.rs
+++ b/wgpu-core/src/command/mod.rs
@@ -7,6 +7,9 @@ mod query;
 mod render;
 mod transfer;
 
+use std::collections::hash_map::Entry;
+use std::ops::Range;
+
 pub use self::bundle::*;
 pub use self::compute::*;
 pub use self::draw::*;
@@ -15,6 +18,7 @@ pub use self::render::*;
 pub use self::transfer::*;
 
 use crate::error::{ErrorFormatter, PrettyError};
+use crate::FastHashMap;
 use crate::{
     hub::{Global, GlobalIdentityHandlerFactory, HalApi, Storage, Token},
     id,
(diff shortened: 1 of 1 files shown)


kvark:
So this isn't actually fixing a correctness issue, but rather fixing the internal logic, right? I.e. it's an optimization PR?

wgpu-core/src/command/mod.rs
```diff
@@ -90,43 +96,55 @@ impl<A: hal::Api> BakedCommands<A> {
...
+                .map_err(|_| DestroyedBufferError(buffer_id))?;
+            let raw_buf = buffer.raw.as_ref().ok_or(DestroyedBufferError(buffer_id))?;
+
+            unsafe {
+                self.encoder
+                    .transition_buffers(transition.map(|pending| pending.into_hal(buffer)));
+            }
+
+            for range in ranges.iter() {
+                assert!(range.start % 4 == 0, "Buffer {:?} has an uninitialized range with a start not aligned to 4 (start was {})", raw_buf, range.start);
```
kvark, line 143: we should use a constant here, like the `COPY_ALIGNMENT`

wgpu-core/src/command/mod.rs
```diff
@@ -90,43 +96,55 @@ impl<A: hal::Api> BakedCommands<A> {
...
+                .get_mut(buffer_id)
+                .map_err(|_| DestroyedBufferError(buffer_id))?;
+            let raw_buf = buffer.raw.as_ref().ok_or(DestroyedBufferError(buffer_id))?;
+
+            unsafe {
+                self.encoder
+                    .transition_buffers(transition.map(|pending| pending.into_hal(buffer)));
+            }
+
+            for range in ranges.iter() {
```
kvark, line 142: I suppose this could do `in ranges` with moving semantics

kvark (approved):

Wumpf:
@kvark yes, title/description isn't quite accurate. Really this is replacing a  optimization that didn't have any effect with one that does

PR #1844, opened 2021-08-22
Update web-sys 0.3.51 to 0.3.53

### LINKED ISSUE #1843: References to removed APIs in Firefox Nightly with web-sys 0.3.51
**Description**
Error thrown when attempting to get prefered Swapchain Format - 
Uncaught (in promise) TypeError: getObject(...).getSwapChainPreferredFormat is not a function

Some investigation indicates that this is due to the version of web-sys (=0.3.51) used in WGPU v0.9 and v0.10.0 not matching the current state of FF nightly API-wise.

**Repro steps**
Usage of any of the js APIs that have recently been removed from FF nightly and web-sys (SwapChain related things mostly).
Can be seen here: https://orange-sand-08e3a2703.azurestaticapps.net/ which is populated from this repo/branch https://github.com/hwoodiwiss/learn-wgpu/tree/wasm-broken-example

I get a non-working application, I had no expectation, I'm aware that I'm in un-stable API teritory.

**Platform**
Windows 10 Pro x64, wgpu v0.9 or v0.10.1 (Tried with both) on Firefox Nightly, notably stopped working between 16 - 20 of August (Not 100% which day sadly).

hwoodiwiss:
**Connections**
fixes #1843

Also implements surface_get_current_texture for the web backend.

A working copy using this is hosted here: https://agreeable-dune-08facd403.azurestaticapps.net/
Which refers to https://github.com/hwoodiwiss/wgpu/tree/update-web-sys-0.3.53 directly

**Description**
This PR updates web-sys and it's usages within the Web backend to a version better reflected by the current nightly builds of Firefox


wgpu/src/backend/web.rs
```diff
@@ -1255,7 +1252,6 @@ impl crate::Context for Context {
                             }
                         };
                         let mut storage_texture = web_sys::GpuStorageTextureBindingLayout::new(
-                            mapped_access,
```
grovesNL, line 1258: [`access`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.GpuStorageTextureBindingLayout.html#method.access) should still be set even if there's only one variant at the moment (this might change again in the short- to mid-term)

wgpu/src/backend/web.rs
```diff
@@ -1811,8 +1807,15 @@ impl crate::Context for Context {
                     crate::LoadOp::Load => wasm_bindgen::JsValue::from(web_sys::GpuLoadOp::Load),
                 };
 
-                let mut mapped_color_attachment =
-                    web_sys::GpuRenderPassColorAttachment::new(&load_value, &ca.view.id.0);
+                let mut mapped_color_attachment = web_sys::GpuRenderPassColorAttachment::new(
+                    &load_value,
+                    if ca.ops.store {
```
grovesNL, line 1812: We can use `map_store_op` for this

wgpu/src/backend/web.rs
```diff
@@ -1169,9 +1168,7 @@ impl crate::Context for Context {
             crate::ShaderSource::SpirV(ref spv) => {
```
grovesNL, line 1168: Now that `&str` is accepted, the `SpirV` variant won't compile anymore. Instead this should use naga to convert SPIR-V to WGSL, then use `WGSL` instead

grovesNL (commented):
Awesome, thanks for doing this! :+1:

wgpu/src/backend/web.rs
hwoodiwiss: I've added the call to `access` with the `mapped_access`.

wgpu/src/backend/web.rs
hwoodiwiss: I've added the call to `map_store_op`, good catch, didn't see that it existed.

wgpu/src/backend/web.rs
hwoodiwiss: I've taken a first go at converting the spir-v data into a WGSL shader module desc. I'm currently unsure of what to do in regards to error handling, as `device_create_shader_module` doesn't return a result, however, just throwing around `unwrap` feels wrong.

grovesNL: I forgot that we should be able to follow the same logic as `direct` here:
https://github.com/gfx-rs/wgpu/blob/master/wgpu/src/backend/direct.rs#L971-L978
The only difference is that at the end we would do `back::wgsl::write_string` instead of returning the naga module

The `unwrap` here is fine for now, we'll properly include this in error logging later

hwoodiwiss: Ah, that makes sense, and avoids the `unsafe` I had to use for converting [u32] to [u8]. Also gives me the preferred parse options, rather than `..Default::default()`, thanks.

grovesNL (approved):
Looks great, thank you!

wgpu/Cargo.toml
```diff
@@ -32,7 +32,7 @@ path = "tests/root.rs"
 
 [features]
 default = []
-spirv = ["naga/spv-in"]
+spirv = ["naga/spv-in", "naga/wgsl-out"]
```
grovesNL, line 35: @kvark is this fine, or should we separate this feature between web and native (we need the full SPIR-V->WGSL path on web but not native)?

kvark: Oh right, hmm. We shouldn't impose the cost to all of the spirv users. Looks like we'll need to move the feature out, like `spirv-web`?

grovesNL: @kvark ok sounds good.

@hwoodiwiss would you mind making this change too? I guess we could do something like `spirv-web = ["spirv", "naga/wgsl-out"]` or similar. If this isn't clear, please let me know and I'd be happy to help out

hwoodiwiss: Cool, this seems sensible, I'll also update the readme  to include reference to `spirv-web` for wasm targets

kvark (approved):

PR #1846, opened 2021-08-23
Validate against cube storage textures

kvark:
**Connections**
matrix chat

**Description**
Cube textures can't be bound as storage.

**Testing**
untested


kvark:
Included in wgpu-core-0.10.1

PR #1847, opened 2021-08-23
fixed mip level count check for storage textures

kocsis1david:
**Connections**
matrix chat

**Description**
The texture view usage was missing the storage flag, because of the wrong mip level count check. This caused a device lost error.

**Testing**
Writing a cubemap as a storage texture with a 2d array texture view now works.

### DIFF
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -136,6 +136,8 @@ pub enum CreateBindGroupError {
         layout_format: wgt::TextureFormat,
         view_format: wgt::TextureFormat,
     },
+    #[error("storage texture bindings must have a single mip level, but given a view with mip_level_count = {mip_level_count:?} at binding {binding}")]
+    InvalidStorageTextureMipLevelCount { binding: u32, mip_level_count: u32 },
     #[error("sampler binding {binding} expects comparison = {layout_cmp}, but given a sampler with comparison = {sampler_cmp}")]
     WrongSamplerComparison {
         binding: u32,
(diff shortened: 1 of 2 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -1657,6 +1657,15 @@ impl<A: HalApi> Device<A> {
                         view_dimension: view.desc.dimension,
                     });
                 }
+
+                let mip_level_count = view.selector.levels.end - view.selector.levels.start;
```
cwfitzgerald, line 1661: If this is a range, this could be `view.selector.levels.len()`.

wgpu-core/src/binding_model.rs
```diff
@@ -136,6 +136,8 @@ pub enum CreateBindGroupError {
         layout_format: wgt::TextureFormat,
         view_format: wgt::TextureFormat,
     },
+    #[error("storage texture binding {binding} should be a single mip level, but given a view with mip_level_count = {mip_level_count:?}")]
```
cwfitzgerald, line 139:
```suggestion
    #[error("storage texture bindings must have a single mip level, but given a view with mip_level_count = {mip_level_count:?} at binding {binding}")]
```

cwfitzgerald (commented):
Small comments, but otherwise looks good.

wgpu-core/src/device/mod.rs
kocsis1david: It is a range, but `len` is from `Iterator` and it returns an `usize` instead of `u32`. I'm not sure if we need that type conversion.

cwfitzgerald: Man why does Range have to be so weird. Honestly it's a wash, lets just merge this.

cwfitzgerald (approved):

kvark (approved):
Thank you!

kvark:
Included in wgpu-core-0.10.1

PR #1850, opened 2021-08-23
[Vulkan] Initialize wgpu objects from raw handles, part 2

zarik5:
**Connections**

#1609

**Description**

While #1609 was needed to support OpenXR applications, this PR aims to support OpenXR runtime/SteamVR drivers implementations. `Instance::from_raw()`, `Adapter::device_from_raw()` and `Device::texture_from_raw()` are made more flexible to accomodate both use-cases.

Recap:

(With ownership I mean the responsibility of destroying the object)  
In a OpenXR game, the instance and device are created by the runtime and owned by the application. The session is created by the runtime using the instance and device and owned by the application. The swapchain textures are created by the runtime and owned by the runtime. Wgpu textures used to access the swapchain are created by the application and the handles are owned by the runtime.  
The wgpu textures must go out of scope before destroying the swapchain and session, which must be destroyed before destroying the device and instance. Drop guards are used as barriers to prevent the drop of objects on which `Instance` and `Texture` depends on, by keeping a reference to it. The presence or absence of drop guards is used also as flag for deciding if to destroy the underlying Vulkan object (`drop_guard == Some` for `Instance::from_raw()`, `drop_guard == None` for `Device::texture_from_raw()`).

This table summarizes the argument values that should be used for each call:

|                Call                 |    OpenXR application     |       OpenXR runtime       |
| :---------------------------------: | :-----------------------: | :------------------------: |
|       `Instance::from_raw()`        |   `drop_guard == Some`    |    `drop_guard == None`    |
|    `Adapter::device_from_raw()`     | `handle_is_owned == true` | `handle_is_owned == false` |
| `Device::texture_from_raw()` |   `drop_guard == Some`    |    `drop_guard == None`    |

The OpenXR runtime does not need drop guards since it's the client application that is resposible for managing the lifetime and drop order of the objects. The device does not use the `drop_guard` because not needed to enforce drop safety in the context of OpenXR.

**Testing**
None for now. I need https://github.com/MaikKlein/ash/pull/457 to be merged (and then wgpu to be updated with the new ash verison), which is orthogonal to this PR.

### DIFF
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -774,6 +774,7 @@ impl super::Adapter {
     pub unsafe fn device_from_raw(
         &self,
         raw_device: ash::Device,
+        handle_is_owned: bool,
         enabled_extensions: &[&'static CStr],
         family_index: u32,
         queue_index: u32,
@@ -845,6 +846,7 @@ impl super::Adapter {
 
         let shared = Arc::new(super::DeviceShared {
             raw: raw_device,
+            handle_is_owned,
             instance: Arc::clone(&self.instance),
             extension_fns: super::DeviceExtensionFunctions {
                 draw_indirect_count: indirect_count_fn,
@@ -944,6 +946,7 @@ impl crate::Adapter<super::Api> for super::Adapter {
 
         self.device_from_raw(
             raw_device,
+            true,
             &enabled_extensions,
             family_info.queue_family_index,
             0,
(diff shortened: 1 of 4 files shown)


zarik5:
There is still one thing that does not look right. When calling `Device::texture_from_raw()` with `drop_guard == None` (that is, the image handle is owned and destroyed by wgpu) wgpu does not create a memory block for it. Creating a memory block requires access to `self`, but the `wgpu_hal::vulkan::Device` is not clonable (and I was not able to make it clonable) and once it has been moved to `wgpu::Device` there is no easy way to get it back, not even a reference to it. For now I'll create and bind the image memory block externally, with the help of the `ash::Device` handle (that can be cloned).

kvark (approved):

zarik5:
[This](https://github.com/alvr-org/ALVR/blob/master/alvr/experiments/graphics_tests/src/compositor/convert.rs) is the reference code for the runtime I'm working on. It's not tested yet. The code for creating the textures is needlessly complex and I would put this on pause until it is possible to access hal Texture from wgpu::Texture, which should be feasible after Hubs are removed from wgpu-core.

PR #1855, opened 2021-08-23
Bump the vertex buffer sizes by 1

kvark:
**Connections**
Matches https://github.com/gpuweb/gpuweb/issues/2044#issuecomment-904077569

**Description**
Vulkan doesn't like vertex buffers bound at the end of the range.
This change adds 1 to all vertex buffers.

Impl note: it would be better to only do this for Vulkan, but then it would have a weird interaction with buffer zeroing. Since zeroing is handled on wgpu-core side, it needs to know the actual size of the buffer. Maybe we can revisit this later?

**Testing**
Untested


cwfitzgerald (approved):
What a bizarre problem

PR #1857, opened 2021-08-24
hal/dx12: do not reverse the pipeline layout

kvark:
**Connections**
Fixes a validation error in DX12 that I found in one of the vange-rs code paths

**Description**
Since the idea of reversed layout came to the code, I fixed it twice (last time in #1744), and now it turns out there is not enough foundation for it to settle at all. Imagine the following scenario:
- BGLs: A, B, C, D
- PL[1]: A, B, C
- PL[2]: A, D

When switching from PL[1] to PL[2] wgpu-core doesn't try to re-bind `A`, since it's already bound, and it's preceeding any entries that changed. With reversed layout, however, `A` in these two pipeline layouts would occupy different slots. So it would have to be re-bound completely (as opposed to just marking the root signature elements as dirty).

So TL;DR: we can't do this right now because wgpu-core follows Vulkan pipeline compatibility rules. There is no principal reason for this: we could change the behavior based on the backend, but that's just the way it works now.

**Testing**
vange-rs

### DIFF
--- a/wgpu-hal/src/dx12/command.rs
+++ b/wgpu-hal/src/dx12/command.rs
@@ -162,6 +162,7 @@ impl super::CommandEncoder {
     }
 
     fn reset_signature(&mut self, layout: &super::PipelineLayoutShared) {
+        log::trace!("Reset signature {:?}", layout.signature);
         if let Some(root_index) = layout.special_constants_root_index {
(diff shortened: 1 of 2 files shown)


kvark:
Included in wgpu-hal-0.10.2

PR #1866, opened 2021-08-26
Fix implicit storage texture access

### LINKED ISSUE #1865: Auto-generated layout missmatch when using textureDimensions and texture_storage_2d
**Description**
I'm getting the following error when creating a pipeline with `layout: None` (i.e. it should auto-generate):
```
Caused by:
    error matching shader requirements against the pipeline
    shader global ResourceBinding { group: 0, binding: 0 } is not available in the layout pipeline layout
    storage texture usage WRITE | QUERY doesn't have a matching supported `StorageTextureAccess`
```
I suspect it's because I'm using textureDimensions in the shader, so it thinks it should be read, but it's actually write. (Just a hypothesis though)

**Repro steps**
The wgsl binding:
```
[[group(0), binding(0)]]
var heightmap: texture_storage_2d<r32float, write>;
```
And then in the shader I'm using `textureDimensions` and `textureStore` on `heightmap`.

Expected the generated layout to be valid.

**Extra materials**
If I explicitly give it a layout with the following it works:
```
BindGroupLayoutEntry {
    binding: 0,
    visibility: ShaderStages::COMPUTE,
    ty: BindingType::StorageTexture {
        access: wgpu::StorageTextureAccess::WriteOnly,
        format: TextureFormat::R32Float,
        view_dimension: TextureViewDimension::D2,
    },
    count: None
},
```

**Platform**
OSX. wgpu 0.10.1

kvark:
**Connections**
Fixes  #1865

**Description**
The checks where written in assumption that there is only READ and WRITE flags. But then QUERY was added.

**Testing**
Manually tested

### DIFF
--- a/wgpu-core/src/validation.rs
+++ b/wgpu-core/src/validation.rs
@@ -563,7 +563,7 @@ impl Resource {
                         multisampled: multi,
                     },
                     naga::ImageClass::Storage { format, .. } => BindingType::StorageTexture {
-                        access: if shader_usage == GlobalUse::WRITE || shader_usage.is_empty() {
+                        access: if !shader_usage.contains(GlobalUse::READ) {
                             wgt::StorageTextureAccess::WriteOnly
                         } else if !features
                             .contains(wgt::Features::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES)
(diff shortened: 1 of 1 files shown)


kvark:
Published in wgpu-core-0.10.2

Issue #1854, opened 2021-08-23
Executable cross-compiled from Linux to Windows will not run on Wine


John-Nagle:
**Description**
Executable cross-compiled from Linux to Windows will not run on Wine.
Memory types offered by Wine are not compatible with what WGPU wants.

fixme:vkd3d_allocate_device_memory: Failed to find suitable memory type (allowed types 0x82).

Here's the [Vulkan spec for those bits.](https://khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkBufferUsageFlagBits.html) 

I realize this is likely to result in finger-pointing between the WGPU side ("Wine doesn't support the memory type we want") and the Wine side ("WGPU should be more flexible in the types of memory it accepts.") The NVidia position is stated here: 
[Discussion by NVidia](https://developer.nvidia.com/what%E2%80%99s-your-vulkan-memory-type): "A memory allocator that follows the rules and guidance of the Vulkan specification should be able to handle all these memory types gracefully by properly interpreting the VkMemoryRequirements::memoryTypeBits member when selecting an allocation for a specific resource."

**Repro steps**
1. Install current Rust on Ubuntu Linux (20.04 LTS used).
2. Install current Wine (6.0.1) or use Wine 5 that ships with 20.04 LTS.
3. Install Rust target for x86_64-pc-windows-gnu
4. Build WGPU examples for that target, debug mode.
5. wine cube.exe

**Expected vs observed behavior**
Expected cube to appear in a window. Got errors. See attachment.

**Extra materials**
Screenshots to help explain your problem.
Validation logs can be attached in case there are warnings and errors.
Zip-compressed API traces and GPU captures can also land here.

**Platform**
Information about your OS, version of `wgpu`, your tech stack, etc.
Ubuntu 20.04 LTS. Tried two different machines, one with Wine 6.0.1 (current stable) and one with the Wine 5 that ships with Ubuntu.

[wgpubug01.txt](https://github.com/gfx-rs/wgpu/files/7033918/wgpubug01.txt)

kvark:
Thank you for filing!

> Here's the Vulkan spec for those bits.

This is a wrong link. The mask isn't for buffer usages, it's for memory type indices.

> fixme:vkd3d_allocate_device_memory: Failed to find suitable memory type (allowed types 0x82).

This reads to me like a bug in `vkd3d`. "fixme" means that it's a signal to vkd3d developers, not wine users.

Also the log has this:
[...]

PR #1867, opened 2021-08-26
hal/dx12: don't assume D3DCompile error is some

kvark:
**Connections**
Related to https://github.com/gfx-rs/wgpu/issues/1854#issuecomment-906569153

**Description**
There is no guarantee that the `error` is anything

**Testing**
Untested


kvark:
Published in wgpu-hal-0.10.3

PR #1874, opened 2021-08-27
hal/metal: inherit scaling from NSView

### LINKED ISSUE #1872: Render attachment not displaying properly when moving window between monitors with different scaling factors
**Description**
A `wgpu::Surface` connected a window which is moved from one monitor to another with a different scaling factor (higher/lower DPI) does not display it's render attachment correctly.

**Repro steps**
- Compile and run the `cube` example: `cargo run --example cube`
- Move the window from one monitor to another with a different scaling factor (changing the scaling factor of the current monitor may also work, I haven't figured out how to test this though...)
- Then resize the window, the output should look clipped. (We resize the window to force a surface resize, since the `WindowEvent::ScaleFactorChanged` is not handled in `framework.rs`)

After resizing the window, the contents of the surface's render attachment is either scaled up or down (see screenshots below) depending on if moving to/from a higher DPI screen. I would expect the render attachment to be properly scaled to fill the entire window 

**Extra materials**
Output before moving window to another screen (this is the expected output):
Screenshot 2021-08-27 at 14 57 32

Output after having moved from a low DPI screen to high DPI screen:
Screenshot 2021-08-27 at 14 58 56

Output after having moved from a high DPI screen to low DPI screen:
Screenshot 2021-08-27 at 14 59 11

**Platform**
macOS Big Sur 11.2.1
wgpu version: 0.10.1
winit version: 0.25.0

kvark:
**Connections**
Fixes #1872

**Description**
Ports the layer delegate from gfx-backend-metal

**Testing**
Manually tested

### DIFF
--- a/wgpu-hal/src/metal/mod.rs
+++ b/wgpu-hal/src/metal/mod.rs
@@ -61,7 +61,9 @@ impl crate::Api for Api {
     type ComputePipeline = ComputePipeline;
 }
 
-pub struct Instance {}
+pub struct Instance {
+    managed_metal_layer_delegate: surface::HalManagedMetalLayerDelegate,
+}
 
 impl Instance {
     pub fn create_surface_from_layer(&self, layer: &mtl::MetalLayerRef) -> Surface {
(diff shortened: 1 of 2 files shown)


kvark:
Published in wgpu-hal-0.10.3

PR #1876, opened 2021-08-27
hal/gles: fix the VB stride limit

kvark:
**Connections**
Reported on matrix, the limit returns 0 on RPI3

**Description**
We should only ask for the limit if we are going to use the relevant API.

**Testing**
Untested


cwfitzgerald (approved):

kvark:
Published in wgpu-hal-0.10.3

PR #1880, opened 2021-08-29
Add Texture::as_hal()

zarik5:
**Connections**
#1850

**Description**
As proposed in https://github.com/gfx-rs/wgpu/pull/1850#issuecomment-903812856, in this PR I try to expose a way to access a hal Texture (and `vk::Image`) from `wgpu::Texture`. To work around lifetimes I had to use two hacks: `Registry::unsafe_read()` and `HalTextureGuard`. `HalTextureGuard` is needed to keep around a reference to a `RwLockReadGuard` while the hal texture is being read. `Registry::unsafe_read()` allows to skip the need for a `Token`. I'm not sure what invariants I'm breaking with `Registry::unsafe_read()`, but having access to the `vk::Image` would break all invariants anyway.

For `Instance` and `Device` I do not need to expose the inner hal object, as this can more easily worked around with the existing `create_*_from_hal()` methods.

This PR doesn't need to be merged just yet. I want to know if there is something that I can do better.

**Testing**
None yet, but I tested the API ergonomics [here](https://github.com/alvr-org/ALVR/blob/master/alvr/experiments/graphics_tests/src/convert.rs).

### DIFF
--- a/wgpu-core/src/resource.rs
+++ b/wgpu-core/src/resource.rs
@@ -1,6 +1,6 @@
 use crate::{
     device::{DeviceError, HostMap, MissingFeatures},
-    hub::Resource,
+    hub::{Global, GlobalIdentityHandlerFactory, HalApi, Resource, Token},
     id::{DeviceId, SurfaceId, TextureId, Valid},
     init_tracker::BufferInitTracker,
     track::{TextureSelector, DUMMY_SELECTOR},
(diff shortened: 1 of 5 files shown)


wgpu-core/src/resource.rs
```diff
@@ -191,6 +192,33 @@ pub struct Texture<A: hal::Api> {
...
+
+        texture.inner.as_raw()
+    }
+}
+
+impl<G: GlobalIdentityHandlerFactory> Global<G> {
+    /// # Safety
+    ///
+    /// - The raw texture handle must not be manually destroyed
+    pub unsafe fn texture_as_hal<'a, A: HalApi>(&'a self, id: TextureId) -> HalTextureGuard<'a, A> {
```
kvark, line 212: can this be shaped with a closure instead? i.e. function gets a closure that's ran over `&Texture`

kvark (commented):
The `unsafe_read` is worrying me here

wgpu-core/src/resource.rs
zarik5: Thank you for the suggestion! This would remove the need for both `HalTextureGuard` and `Registry::unsafe_read()`.

zarik5:
I updated the PR with your suggestion. I also updated the sample with the usage.

kvark (approved):

kvark:
I generally prefer to avoid closures unless necessary. In this case, avoiding it is really painful. Hopefully we can reconsider after the dehubization is done by @pythonesque

PR #1885, opened 2021-08-30
hal/metal: support Features::NON_FILL_POLYGON_MODE

nolanderc:
**Connections**
None

**Description**
Allows usage of `wgpu::PolygonMode::Line` and `wgpu::PolygonMode::Point` on the Metal backend. This is done by setting `MTLTriangleFillMode` to `Lines`, in case of lines, and changing the topology to `PointList` if drawing points. I believe changing the topology should be sound (see comment in commit, though please correct me if I'm wrong).

**Testing**
Manual testing by changing `polygon_mode` in the `shadow` example.

With `wgpu::PolygonMode::Line`:
Screenshot 2021-08-30 at 12 00 05

With `wgpu::PolygonMode::Point`:
Screenshot 2021-08-30 at 12 01 09

(okay, you can barely see them, but you get the point... :)

### DIFF
--- a/wgpu-hal/src/dx12/adapter.rs
+++ b/wgpu-hal/src/dx12/adapter.rs
@@ -174,8 +174,8 @@ impl super::Adapter {
             | wgt::Features::MULTI_DRAW_INDIRECT
             | wgt::Features::MULTI_DRAW_INDIRECT_COUNT
             | wgt::Features::ADDRESS_MODE_CLAMP_TO_BORDER
-            | wgt::Features::LINE_POLYGON_MODE
-            | wgt::Features::POINT_POLYGON_MODE
+            | wgt::Features::POLYGON_MODE_LINE
+            | wgt::Features::POLYGON_MODE_POINT
             | wgt::Features::VERTEX_WRITABLE_STORAGE
             | wgt::Features::TIMESTAMP_QUERY;
         //TODO: in order to expose this, we need to run a compute shader
--- a/wgpu-hal/src/dx12/adapter.rs
+++ b/wgpu-hal/src/dx12/adapter.rs
@@ -174,7 +174,8 @@ impl super::Adapter {
             | wgt::Features::MULTI_DRAW_INDIRECT
             | wgt::Features::MULTI_DRAW_INDIRECT_COUNT
             | wgt::Features::ADDRESS_MODE_CLAMP_TO_BORDER
-            | wgt::Features::NON_FILL_POLYGON_MODE
+            | wgt::Features::LINE_POLYGON_MODE
+            | wgt::Features::POINT_POLYGON_MODE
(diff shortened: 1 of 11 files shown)


nolanderc:
Though, thinking about it some more, changing the topology when using `PolygonMode::Point` does mean we lose information about triangle orientations, and thus lose the ability to do face culling. Not sure how to fix that though...

kvark:
We did consider using Points back in the day, and it cuts too many corners, unfortunately:
  - face culling is skipped
  - doesn't respect primitive restart index
  - doesn't respect depth offset
  - points are repeated many times, which will fail if they have side effects (like storage writes)
  - potentially other things...

Splitting the feature into 2 would be good.

nolanderc:
@kvark I added two new features to replace the old `NON_FILL_POLYGON_MODE`. I had to adjust the bitflags for the other features in order to make space for the new features (I want to keep them close together if possible). Although, I'm not sure what kind of stability guarantees you have around keeping these bitflags stable between versions?

wgpu-hal/src/dx12/adapter.rs
```diff
@@ -174,7 +174,8 @@ impl super::Adapter {
             | wgt::Features::MULTI_DRAW_INDIRECT
             | wgt::Features::MULTI_DRAW_INDIRECT_COUNT
             | wgt::Features::ADDRESS_MODE_CLAMP_TO_BORDER
-            | wgt::Features::NON_FILL_POLYGON_MODE
+            | wgt::Features::LINE_POLYGON_MODE
```
kvark, line 177: It would be more consistent (e.g. with "ADDRESS_MODE_CLAMP_TO_BORDER") to call it the other way around: "POLYGON_MODE_LINE" and "POLYGON_MODE_POINT"

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -71,7 +71,8 @@ impl PhysicalDeviceFeatures {
                 )
                 .depth_clamp(requested_features.contains(wgt::Features::DEPTH_CLAMPING))
                 .fill_mode_non_solid(
-                    requested_features.contains(wgt::Features::NON_FILL_POLYGON_MODE),
+                    requested_features.contains(wgt::Features::LINE_POLYGON_MODE)
```
kvark, line 74: `requested_features.intersects(wgt::Features::LINE_POLYGON_MODE | wgt::Features::POINT_POLYGON_MODE)`

kvark:
> Although, I'm not sure what kind of stability guarantees you have around keeping these bitflags stable between versions?

No stability guarantees between breaking versions. We are still pre-1.0

kvark (approved):

kvark:
Please rebase

PR #1889, opened 2021-08-30
Align wgpu_types serde serializations to spec

### LINKED ISSUE #1887: Implement `serde::Deserialize` on wgpu_types types, following names from WebGPU JS API
The deno_webgpu crate from https://github.com/gfx-rs/wgpu/pull/1859#discussion_r696039014 could be significantly simplified if `wgpu_types` implemented `serde::Deserialize` on it's types (mainly the enums). The reason for this is that a lot of duplicate boilerplate code needs to be written to deserialize string versions of for example [`GPUTextureFormat`](https://gpuweb.github.io/gpuweb/#enumdef-gputextureformat) into a [`wgpu_types::TextureFormat`](https://docs.rs/wgpu-types/0.10.0/wgpu_types/enum.TextureFormat.html).

Can we just implement `serde::Deserialize` on all enums with the same string representations as the WebGPU spec if an optional `serde` feature is provided to the crate?

lucacasonato:
The representation of most wgpu_types objects now aligns to how the
WebGPU spec represents them. Enums are serialized as kebab-case strings,
and structs are serialized as camelCase objects.

Closes #1887


wgpu-types/src/lib.rs
```diff
@@ -1407,95 +1465,109 @@ pub enum TextureFormat {
...
     Bc7RgbaUnorm,
     /// 4x4 block compressed texture. 16 bytes per block (8 bit/px). Variable sized pallet. 8 bit integer RGBA.
     /// Srgb-color [0, 255] converted to/from linear-color float [0, 1] in shader.
     ///
     /// Also known as BPTC (unorm).
     ///
     /// [`Features::TEXTURE_COMPRESSION_BC`] must be enabled to use this texture format.
+    #[cfg_attr(feature = "serde", serde(rename = "bc7-rgba-unorm-srgb"))]
     Bc7RgbaUnormSrgb,
     /// 4x4 block compressed texture. 8 bytes per block (4 bit/px). Complex pallet. 8 bit integer RGB.
```
lucacasonato, line 1572: Non standard variants below do not get proper names.

wgpu-types/src/lib.rs
```diff
@@ -921,6 +929,7 @@ impl TextureViewDimension {
 #[derive(Copy, Clone, Debug, Hash, Eq, PartialEq)]
 #[cfg_attr(feature = "trace", derive(Serialize))]
 #[cfg_attr(feature = "replay", derive(Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "kebab-case"))]
```
kvark, line 932: wow, this is really neat!

kvark (approved):

kvark:
Thank you, this looks great! Let's make CI happy and proceed

PR #1890, opened 2021-08-31
Fix some comments.

dev-chee:
**Connections**
Fix some comments.

**Description**
Fix some comments.

**Testing**
Unnecessary.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -1409,7 +1409,7 @@ pub enum TextureFormat {
     /// [`Features::TEXTURE_COMPRESSION_BC`] must be enabled to use this texture format.
     Bc1RgbaUnorm,
     /// 4x4 block compressed texture. 8 bytes per block (4 bit/px). 4 color + alpha pallet. 5 bit R + 6 bit G + 5 bit B + 1 bit alpha.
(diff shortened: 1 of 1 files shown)


wgpu-types/src/lib.rs
```diff
@@ -1309,7 +1309,7 @@ pub enum TextureFormat {
     // Normal 8 bit formats
     /// Red channel only. 8 bit integer per channel. [0, 255] converted to/from float [0, 1] in shader.
     R8Unorm,
-    /// Red channel only. 8 bit integer per channel. [-127, 127] converted to/from float [-1, 1] in shader.
+    /// Red channel only. 8 bit integer per channel. [-128, 127] converted to/from float [-1, 1] in shader.
```
kvark, line 1312: I think the old comment is correct. -127 maps to -1.0

dev-chee: sorry, i think you are right. i found the specification for vulkan. https://www.khronos.org/registry/vulkan/specs/1.1/html/vkspec.html#fundamentals-fixedfpconv

dev-chee: i have removed those of the modification.

kvark (approved):
Thanks!

PR #1893, opened 2021-09-01
Implement render pass labels in Vulkan

cwfitzgerald:
**Connections**

None

**Description**

This allows renderpass labels to show up in renderdoc. Just one of those small creature comforts.

**Testing**

Renderdoc'd code.

### DIFF
--- a/wgpu-hal/src/vulkan/command.rs
+++ b/wgpu-hal/src/vulkan/command.rs
@@ -62,6 +62,9 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
             label.unwrap_or_default(),
         );
 
+        // Reset this in case the last renderpass was never ended.
+        self.rpass_debug_marker_active = false;
+
         let vk_info = vk::CommandBufferBeginInfo::builder()
             .flags(vk::CommandBufferUsageFlags::ONE_TIME_SUBMIT)
             .build();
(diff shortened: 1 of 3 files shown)


wgpu-hal/src/vulkan/command.rs
```diff
@@ -427,6 +427,11 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
             vk_info = vk_info.push_next(&mut vk_attachment_info);
         }
 
+        if let Some(label) = desc.label {
+            self.begin_debug_marker(label);
+            self.rpass_debug_marker_active = true;
```
kvark, line 432: we need to reset this properly somewhere. Otherwise, we can be in a situation where the recording is discarded, but it would think that the marker is active on the next command buffer

kvark (changes requested):
Once small fix is needed

wgpu-hal/src/vulkan/command.rs
cwfitzgerald: That work?

wgpu-hal/src/vulkan/command.rs
```diff
@@ -81,6 +81,7 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
     unsafe fn discard_encoding(&mut self) {
         self.discarded.push(self.active);
         self.active = vk::CommandBuffer::null();
+        self.rpass_debug_marker_active = false;
```
kvark, line 84: it would be better to reset it in `begin_encoding`

kvark (approved):

kvark:
Published in wgpu-hal-0.10.3

PR #1894, opened 2021-09-01
hal/vk: always request StorageImageExtendedFormats in SPIR-V

kvark:
**Connections**
Compensates https://github.com/gfx-rs/naga/pull/1306
Related to https://github.com/KhronosGroup/Vulkan-Docs/issues/395

**Description**
SPIR-V requires us to enable this capability, but Vulkan device doesn't require it. So we need to at least not fail on it internally...

**Testing**
TODO


kvark:
Appears to work for me :)

kvark:
Published in wgpu-hal-0.10.3

Issue #1877, opened 2021-08-28
Using two encoders in hello-triangle example causes hang 


fintelia:
**Description**
Wgpu hangs when rendering a frame that involved multiple consecutive command encoders. [This patch](https://github.com/fintelia/wgpu/commit/1568d8577cb4e4e10b4df45fbf62e2f7dd87e619) to the hello-triangle example is enough to trigger the bug.  

**Repro steps**

```
$ git clone https://github.com/fintelia/wgpu && cd wgpu
$ git checkout double-encoder-hang
$ cd wgpu
$ cargo run --example hello-triangle
```

**Expected vs observed behavior**
Expected hello-triangle example to run normally. Instead it displays all black and the window cannot be closed.

**Platform**
* Arch Linux
* Intel integrated GPU
* wgpu commit 39e8921fe8ebda314edf5c4787cdbb82281a50de

fintelia:
Via `git bisect` the this is caused by a regression in 5419947095853ee3b91c04f74f0e4193d0d061a8:

```
$ git checkout 39e8921
$ git bisect start
$ git bisect bad
$ git bisect good 91437ea7
$ git bisect test.sh
```



<details><summary>test.sh</summary>

```bash
#!/bin/bash

cat >test.patch << EOF
diff --git a/wgpu/examples/hello-triangle/main.rs b/wgpu/examples/hello-triangle/main.rs
index 4d27af83..25cfef21 100644
--- a/wgpu/examples/hello-triangle/main.rs
+++ b/wgpu/examples/hello-triangle/main.rs
@@ -95,6 +95,11 @@ async fn run(event_loop: EventLoop<()>, window: Window) {
                     .get_current_frame()
                     .expect("Failed to acquire next swap chain texture")
                     .output;
+
+                *control_flow = ControlFlow::Exit;
+                let e2 = device.create_command_encoder(&wgpu::CommandEncoderDescriptor { label: None });
+                queue.submit(Some(e2.finish()));
+
                 let mut encoder =
                     device.create_command_encoder(&wgpu::CommandEncoderDescriptor { label: None });
                 {
EOF

cd wgpu/examples
if git apply test.patch && cargo build --example hello-triangle
then
	# run project specific test and report its status
    timeout 15 cargo run --example hello-triangle
	status=$?
else
	# tell the caller this is untestable
	status=125
fi
cd ../..

# undo the tweak to allow clean flipping to the next commit
git reset --hard

# return control
exit $status
```

</details>

[...]

Issue #1672, opened 2021-07-16
wgpu may lock up in ioctl on Linux/Vulkan/Intel


kvark:
**Description**
Under certain scenarios, we may see a hang in `ioctl`.

**Repro steps**
Unknown.

**Expected vs observed behavior**
No hangs.

**Extra materials**
Looks like this is described in https://www.reddit.com/r/vulkan/comments/b37762/command_queue_grows_indefinitely_on_intel_gpus/
Edit: actually, no, we aren't expecting `vkAcquireNextImageKHR` to block. We are explicitly blocking on the fence, which was passed to it, instead.

**Platform**
wgpu master

Bobo1239:
Not sure whether this is the same issue but I've recently tracked down something similar in the same environment. (Linux (Wayland)/ Vulkan/Intel) My code is based on the [Vulkan tutorial](https://vulkan-tutorial.com/) so you can just take e.g. this [repo](https://github.com/adrien-ben/vulkan-tutorial-rs) (compile in `release` mode due to asset loading).

In my case (and also with the linked repo) the issue is always quickly reproducible by having a Youtube running in the background (in another [Sway](https://github.com/swaywm/sway) workspace) and switching window focus/workspaces a couple of times. The crux is though that the hang only happens if I set [`FRAMES_IN_FLIGHT`](https://github.com/adrien-ben/vulkan-tutorial-rs/blob/master/src/main.rs#L32) to `1`. The issue never happens with `2`. Since https://github.com/gfx-rs/wgpu/issues/932 is open I'll assume that wgpu doesn't have multiple frames in flight so that may be the same root cause. Unfortunately I have no idea where to go from here since that's probably a driver bug...

Backtrace on my machine (hang is inside `vkQueuePresentKHR()`):
```
[#0] 0x7f6943be559b → ioctl()
[#1] 0x7f6942ead924 → cmp eax, 0xffffffff
[#2] 0x56359223412d → ash::extensions::khr::swapchain::Swapchain::queue_present()
[#3] 0x563591feeb92 → vulkan_tutorial_ash::VulkanApp::draw_frame()
[#4] 0x563591ffd7c0 → _ZN19vulkan_tutorial_ash4main28_$u7b$$u7b$closure$u7d$$u7d$17ha17eb207c13ce696E.llvm.10835609363145980811()
[#5] 0x563592050dc8 → winit::platform_impl::platform::wayland::event_loop::EventLoop<T>::run()
[...]

PR #1898, opened 2021-09-02
hal/linux: Work around Intel+Nvidia presentation conflict

kvark:
**Connections**
This periodically shows up and is very annoying: https://github.com/gfx-rs/wgpu/issues/1877#issuecomment-911504808 and potentially https://github.com/gfx-rs/wgpu/issues/1672
Also fixed upstream around 4 month ago - https://gitlab.freedesktop.org/mesa/mesa/-/issues/4688
I think there is still value in handling that internally, at least until we can detect that the Intel driver is fixed.

**Description**
Vulkan on Linux: if we detect Intel iGPU + NVidia dGPU, we disable presentation on Intel.
The PR also fixes the `compatible_surface` argument in the util and framework.

**Testing**
Manually tested on my laptop!


cwfitzgerald (approved):

kvark:
published in wgpu-hal-0.10.4

PR #1899, opened 2021-09-02
Fix read-write buffer barrier

kvark:
**Connections**
Reported on Matrix by @cwfitzgerald 

**Description**
We were missing an internal usage for reading RW buffers.

**Testing**
VVL doesn't complain!


cwfitzgerald (approved):
Fixes my problem on NV

kvark:
published in wgpu-core-0.10.3

PR #1900, opened 2021-09-02
Derive Clone for ShaderSource

fintelia:
**Connections**
N/A

**Description**
Implements `Clone` for `ShaderSource` enum.

**Testing**
I ran `cargo test`

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -734,6 +734,7 @@ impl Drop for ShaderModule {
 }
 
 /// Source of a shader module.
+#[derive(Clone)]
 pub enum ShaderSource<'a> {
     /// SPIR-V module represented as a slice of words.
     ///


kvark (approved):

wgpu/src/lib.rs
```diff
@@ -734,6 +734,7 @@ impl Drop for ShaderModule {
 }
 
 /// Source of a shader module.
+#[derive(Clone)]
```
kvark, line 737: I don't think you ever want the `Clone` derive here.
For `Cow` type, it just clones the contents instead of re-borrowing them - https://doc.rust-lang.org/stable/src/alloc/borrow.rs.html#194-211

If you want `Clone`, I suggest just implementing it manually, using `Cow::Borrowed`

kvark:
Closing due to inactivity

PR #1905, opened 2021-09-04
Reintroduce clear_texture Metal/Vulkan/DX12

Wumpf:
**Connections**

Major puzzle piece for getting texture zero init in, see #1688
The idea is to implement the lazy texture inits with render target clears _wherever possible_, but for everything else we need to be able to perform clears using merely the COPY_DST usage, wgpu_hal's `clear_texture` is the answer for that.

**Description**

Main piece here is the reintroduction & implementation of clear_texture, but a few related things made it was well in here:

* reenable CLEAR_COMMANDS feature accross apis
* fill_buffer -> clear_buffer (makes things easier and simpler to find)
* improved test for buffer & texture clearing (previously it would just pass since apis give you zero initialized most or even all of the time)
* remove unused player test

Note that clear_texture is more restricted than originally:
* requires COPY_DST usage 
* does not support depth/stencil formats
* does not support multisampled formats
* clears to zero only (same for clear_buffer)

For implementing zero init we will internally need to enforce either COPY_DST or RENDER_ATTACHMENT. Conveniently, the formats that clear_texture is not supported almost always come with RENDER_ATTACHMENT anyways.

Did *not* implemented `clear_texture` for GLES, I simply ran out of energy to do the tricky row copying again. Left as an exercise for the reader.

**Testing**

Only tested using the rather simplistic player test. We'll need to set up tests that iterate through more texture types/sizes/formats in the future (not only for clearing, but for all sort of things).
Ran it with Vulkan/DX12 (@ Win10, Nvidia) and Metal (@ MacOS, Intel)


wgpu/src/lib.rs
```diff
@@ -2260,13 +2260,15 @@ impl CommandEncoder {
 
     /// Clears texture to zero.
     ///
+    /// Where possible it may be significantly more efficient to perform clears via render targets!
```
Wumpf, line 2263: should probably write `render pass`, not target

wgpu-hal/src/dx12/command.rs
```diff
@@ -380,6 +379,101 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
...
+            texture.size.depth_or_array_layers
+        } else {
+            1
+        };
+
+        let mip_range = subresource_range.base_mip_level
+            ..(subresource_range.base_mip_level
+                + subresource_range
+                    .mip_level_count
+                    .map_or(texture.mip_level_count, |c| c.get()));
```
kvark, line 415: this is incorrect, it should be something like
```rust
match subresource_range.mip_level_count {
  Some(c) => subresource_range.base_mip_level + c.get(),
  None => texture.mip_level_count,
}
```

Otherwise, the non-zero `base_mip_level` with no `mip_level_count` would lead to an invalid range

wgpu-hal/src/dx12/command.rs
```diff
@@ -380,6 +379,101 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
...
+                / d3d12::D3D12_TEXTURE_DATA_PITCH_ALIGNMENT
+                * d3d12::D3D12_TEXTURE_DATA_PITCH_ALIGNMENT;
+
+            let max_rows_per_copy = super::ZERO_BUFFER_SIZE as u32 / bytes_per_row;
+            // round down to a multiple of rows needed by the texture format
+            let max_rows_per_copy = max_rows_per_copy / format_desc.block_dimensions.1 as u32
+                * format_desc.block_dimensions.1 as u32;
+            assert!(max_rows_per_copy > 0, "Zero buffer size is too small to fill a single row of a texture with dimension {:?}, size {:?} and format {:?}", texture.dimension, texture.size, texture.format);
+
+            for array_layer in array_range.clone() {
```
kvark, line 437: is `clone` needed here?

wgpu-hal/src/metal/adapter.rs
```diff
@@ -862,7 +862,8 @@ impl super::PrivateCapabilities {
             | F::MAPPABLE_PRIMARY_BUFFERS
             | F::VERTEX_WRITABLE_STORAGE
             | F::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES
-            | F::POLYGON_MODE_LINE;
+            | F::POLYGON_MODE_LINE
+            | wgt::Features::CLEAR_COMMANDS;
```
kvark, line 866: should be `F::CLEAR_COMMANDS`

wgpu-hal/src/metal/command.rs
```diff
@@ -126,9 +126,83 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
...
+        let encoder = self.enter_blit();
+
+        let format_desc = texture.format.describe();
+        let depth = if texture.raw_type == mtl::MTLTextureType::D3 {
+            texture.size.depth_or_array_layers as u64
+        } else {
+            1
+        };
+
+        let mip_range = subresource_range.base_mip_level
```
kvark, line 149: same concern here

wgpu-hal/src/metal/mod.rs
```diff
@@ -399,15 +415,30 @@ impl Buffer {
...
     array_layers: u32,
     mip_levels: u32,
+    size: wgt::Extent3d,
 }
 
 unsafe impl Send for Texture {}
 unsafe impl Sync for Texture {}
 
+impl Texture {
+    fn mip_level_size(&self, level: u32) -> wgt::Extent3d {
```
kvark, line 430: it looks like this method should be shared between the backends?

kvark (changes requested):
What is the semantics of CLEAR_COMMANDS now? Is it purely for textures? If so, we may need to indicate that in the name.

Wumpf:
`CLEAR_COMMANDS` enables the user sided use of `clear_texture` and `clear_buffer`. How `CLEAR_COMMANDS` is now documented:
```
        /// Enables clear to zero for buffers & textures.
        ///
        /// Supported platforms:
        /// - All
        ///
        const CLEAR_COMMANDS = 1 << 36;
```
By renaming & simplifying `clear_buffer` I was hoping that the semantics are now more obvious :)

Arguably now that I explicitly state for the `clear_texture` that it might be inefficient, the main use is internal testing. Although arguably still convenient for various usecases :)

wgpu-hal/src/dx12/command.rs
Wumpf: yeah oddly enough it is. Compiler complains that this is an implicit into_iter() call and as such it is moved from the previous loop run. And range doesn't have an iter() call as alternative

kvark:
> By renaming & simplifying clear_buffer I was hoping that the semantics are now more obvious :)

Ok, that's the actual concern.
Before this PR, wgpu-hal was *required* to implement buffer clears (since it's needed by wgpu-core anyway). And now it's no longer required, and is hidden behind a feature? That seems like a wrong direction.

Wumpf:
Ah okay this is a bit messed up right now, but already was before I reckon. wgpu-hal still is required to implement both clear_buffer and clear_texture because otherwise wgpu won't be able to implement zero-init which it is required to do. However, wgpu won't expose clear_buffer/texture unless asked so because it is not a feature of WebGPU. Allowing that would lead to non-portable applications.

So what I should do in this pr is that it's not wgpu-hal advertising the availability of the feature (it's always there), but wgpu itself?

kvark (approved):
Discussed this on matrix. The short story is: clearing textures is a prerequisite for zero-initialization in wgpu-core, so separating it out from buffers wouldn't make much sense.

scoopr:
It seems that on macos, all the examples now hang on waiting a mutex in `acquire_surface` for me, and I hand bisected it to this change.

PR #1907, opened 2021-09-05
Properly implement timestamp periods

cwfitzgerald:
**Connections**

None

**Description**

We weren't bubbling this up, which caused timestamp results to be ~80 times smaller than they should be on intel.

**Testing**

Manually tested using rend3 and wgpu-profiler.

**Backporting Notes**

I would appreciate this be backported, which is possible by adding a default implementation of the get_timestamp_period method on the queue trait. This makes it backwards compatible


kvark (approved):

kvark:
Timestamp API is about to change drastically soon

kvark:
published in wgpu-hal-0.10.4 and wgpu-core-0.10.3

PR #1908, opened 2021-09-06
hal: switch texture copies to expect physical sizes

### LINKED ISSUE #1895: wgpu-hal Uses Virtual Sizes for Compressed Texture Copies Instead of Physical Sizes

Follow up to unify https://github.com/gfx-rs/wgpu/pull/1819/

We should be using physical sizes for texture copies like WebGPU [specifies](https://gpuweb.github.io/gpuweb/#valid-texture-copy-range):

> copySize.width must be a multiple of blockWidth.
> copySize.height must be a multiple of blockHeight.

For example, when uploading the second mip of a 4x4 compressed texture, the physical size is 4x4 and the virtual is 2x2. We pass 2x2 to wgpu-hal currently, but we should stay consistent with WebGPU. This will make the vulkan backend's job a bit harder, but the dx12 backend easier.

Not doing this. Imo this is a non-answer as we need wgpu-hal to stay fairly consistent with WebGPU, just unsafe.

kvark:
**Connections**
Fixes #1895

**Description**
Currently, wgpu-core expects physical sizes for texture copies, but wgpu-hal expects virtual.
Moreover, Metal texture-texture copies already expect physical (but not buffer-related copies), which wasn't handled properly.

This PR fixes Metal properly and aligns wgpu-hal copy API with wgpu-core.

**Testing**
Tested the examples on metal. Can confirm the "skybox" example no longer throws a validation error on Metal, at least.

### DIFF
--- a/wgpu-hal/src/gles/command.rs
+++ b/wgpu-hal/src/gles/command.rs
@@ -290,7 +290,8 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
     {
         let (src_raw, src_target) = src.inner.as_native();
         let (dst_raw, dst_target) = dst.inner.as_native();
-        for copy in regions {
+        for mut copy in regions {
+            copy.clamp_size_to_virtual(&src.copy_size, &dst.copy_size);
             self.cmd_buffer.commands.push(C::CopyTextureToTexture {
                 src: src_raw,
                 src_target,
@@ -310,7 +311,8 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
         T: Iterator<Item = crate::BufferTextureCopy>,
     {
         let (dst_raw, dst_target) = dst.inner.as_native();
-        for copy in regions {
+        for mut copy in regions {
+            copy.clamp_size_to_virtual(&dst.copy_size);
             self.cmd_buffer.commands.push(C::CopyBufferToTexture {
                 src: src.raw,
                 src_target: src.target,
@@ -332,7 +334,8 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
         T: Iterator<Item = crate::BufferTextureCopy>,
     {
         let (src_raw, src_target) = src.inner.as_native();
-        for copy in regions {
+        for mut copy in regions {
+            copy.clamp_size_to_virtual(&src.copy_size);
             self.cmd_buffer.commands.push(C::CopyTextureToBuffer {
                 src: src_raw,
                 src_target,
--- a/wgpu-hal/src/vulkan/command.rs
+++ b/wgpu-hal/src/vulkan/command.rs
@@ -16,7 +16,9 @@ impl super::Texture {
     {
         let aspects = self.aspects;
         let fi = self.format_info;
+        let copy_size = self.copy_size;
         regions.map(move |r| {
+            let extent = r.texture_base.max_copy_size(&copy_size).min(&r.size);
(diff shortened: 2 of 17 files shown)


wgpu-hal/src/gles/command.rs
```diff
@@ -30,6 +30,46 @@ pub(super) struct State {
     dirty_vbuf_mask: usize,
 }
 
+impl crate::CopyExtent {
```
cwfitzgerald, line 33: All these non-local impls seem misleading -- it exists at the top level, but only if the gles feature is enabled. I think these should just be implemented at the top level and the other backends can just not use them.

wgpu-hal/src/metal/conv.rs
```diff
@@ -261,7 +261,23 @@ pub fn map_range(range: &crate::MemoryRange) -> mtl::NSRange {
     }
 }
 
-pub fn map_copy_extent(extent: &crate::CopyExtent) -> mtl::MTLSize {
+pub fn map_buffer_copy_extent(
```
cwfitzgerald, line 264: This is another place that we might be able to reuse the gles code.

wgpu-hal/src/vulkan/command.rs
```diff
@@ -261,12 +262,18 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
...
                 conv::map_subresource_layers(&r.dst_base, dst.aspects);
+            let src_extent = conv::map_copy_extent(&r.size, &r.src_base, &src.copy_size);
+            let dst_extent = conv::map_copy_extent(&r.size, &r.dst_base, &dst.copy_size);
             vk::ImageCopy {
                 src_subresource,
                 src_offset,
                 dst_subresource,
                 dst_offset,
-                extent: conv::map_copy_extent(&r.size),
+                extent: vk::Extent3D {
```
cwfitzgerald, line 272: And here.

wgpu-hal/src/vulkan/conv.rs
```diff
@@ -504,11 +504,33 @@ pub fn map_view_dimension(dim: wgt::TextureViewDimension) -> vk::ImageViewType {
     }
 }
 
-pub fn map_copy_extent(extent: &crate::CopyExtent) -> vk::Extent3D {
+pub fn map_copy_extent(
```
cwfitzgerald, line 507: And here

cwfitzgerald (approved):
Code looks fine -- not as bad as I had feared. I have a code reuse comment, but other than that we're good.

wgpu-hal/src/gles/command.rs
kvark: Actually, I really liked how they allow us to structure the code. The `impl` itself doesn't have visibility, but the function inside the `impl` does. So if this function is `pub(super)`, it means it's only visible in `gles` sub-module, and not the top-level like you assumed.

kvark: Now I see what you mean, the code isn't backend-specific and should be moved higher up. I agree. Was implementing GLES last and didn't want to re-touch all the previous code.

kvark:
Cleaned up now, and it looks a bit better. Thanks for the review!

wgpu-hal/src/gles/command.rs
cwfitzgerald: That is an interesting side effect, I had no idea you could do that at all.

kvark:
published in wgpu-hal-0.10.4

PR #1912, opened 2021-09-06
Expose raw command buffer

Neo-Zhixing:
This PR exposes the raw command buffer to the user.

Usage:

```rs
unsafe {
  command_encoder.run_raw_command::<wgpu_hal::api::Vulkan, _>(|command_buffer| {
      let command_buffer = command_buffer.active;
      assert_ne!(command_buffer, vk::CommandBuffer::null());
      device.cmd_bind_pipeline(
          command_buffer,
          vk::PipelineBindPoint::RAY_TRACING_KHR,
          ray_shaders.pipeline,
      );
      raytracing_pipeline_loader.cmd_trace_rays(
          command_buffer,
          &ray_shaders.raygen_shader_binding_tables,
          &ray_shaders.miss_shader_binding_tables,
          &ray_shaders.hit_shader_binding_tables,
          &ray_shaders.callable_shader_binding_tables,
          1920, //render_state.extent.width,
          1080, //render_state.extent.height,
          1,
      );
  });
}

```

Note that this is only tested on Vulkan for now - I would like to get some feedbacks on this before proceeding.

### DIFF
--- a/wgpu-core/src/command/mod.rs
+++ b/wgpu-core/src/command/mod.rs
@@ -337,6 +337,27 @@ pub enum CommandEncoderError {
 }
 
 impl<G: GlobalIdentityHandlerFactory> Global<G> {
+    pub unsafe fn command_encoder_run_raw_commands<A, F>(
+        &self,
+        command_encoder_id: id::CommandEncoderId,
+        runner: F,
+    ) -> Result<(), CommandEncoderError>
+    where
+        A: HalApi,
+        F: FnOnce(&mut <A as hal::Api>::CommandEncoder),
+    {
+        profiling::scope!("run_raw_command", "CommandEncoder");
+        let hub = A::hub(self);
+        let mut token = Token::root();
+
+        let (mut cmd_buf_guard, _) = hub.command_buffers.write(&mut token);
+        let cmd_buf = CommandBuffer::get_encoder_mut(&mut *cmd_buf_guard, command_encoder_id)?;
+
+        let cmd_buf_raw = cmd_buf.encoder.open();
+        runner(cmd_buf_raw);
+        Ok(())
+    }
+
     pub fn command_encoder_finish<A: HalApi>(
         &self,
         encoder_id: id::CommandEncoderId,
--- a/wgpu-hal/src/vulkan/mod.rs
+++ b/wgpu-hal/src/vulkan/mod.rs
@@ -352,9 +352,9 @@ impl Temp {
 }
 
 pub struct CommandEncoder {
-    raw: vk::CommandPool,
+    pub raw: vk::CommandPool,
     device: Arc<DeviceShared>,
-    active: vk::CommandBuffer,
+    pub active: vk::CommandBuffer,
     bind_point: vk::PipelineBindPoint,
     temp: Temp,
     free: Vec<vk::CommandBuffer>,
--- a/wgpu/src/backend/direct.rs
+++ b/wgpu/src/backend/direct.rs
@@ -1950,6 +1950,25 @@ impl crate::Context for Context {
         id
     }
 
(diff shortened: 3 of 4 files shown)


kvark:
What is the end goal here? Are you trying to use Vulkan ray tracing within a WebGPU application? I don't think it's feasible to bolt onto wgpu from user space. wgpu's main promise is safety, and it can't provide this if you are messing around the bindings used in command buffers. It's not clear to me what benefit `wgpu` offers to your application if you want to use RT stuff. If we are to call `cmd_trace_rays` on Vulkan, we need to expose this as a real feature.

Neo-Zhixing:
I think the idea here is similar to #1850 and and #1609.

What I'm trying to do here is building a hybrid rendering pipeline. The main scene and terrain are rendered using Vulkan RTX, and the UI and a few other things rendered with a traditional rasterization pipeline in bevy. I would like to take advantage of the ecosystem built upon WGPU, but at the same time I need some customizability to interact with the low-level Vulkan API.

```
wgpu's main promise is safety, and it can't provide this if you are messing around the bindings used in command buffers.
```
Yes this is true, but that's why these APIs are marked as unsafe. If I call an unsafe API into wgpu and things break, it's on me.


```
 If we are to call cmd_trace_rays on Vulkan, we need to expose this as a real feature.
```
Yes sure, having ray tracing support in wgpu would be awesome. However I'm not sure how much of a priority this is right now, and in the short term it might be less of a maintenance burden to directly expose the related low-level APIs to the users.

kvark:
I see, thank you for explaining!
Did you consider an alternative approach that's more similar to the texture creation? Something like `CommandBuffer::from_raw`?

Neo-Zhixing:
This is going in the other direction, from wgpu `CommandEncoder` to raw command buffer.

This is more similar to https://github.com/gfx-rs/wgpu/blob/df2a686c29720679c786d0e4973738379c344fa9/wgpu/src/lib.rs#L2085 and https://github.com/gfx-rs/wgpu/blob/53d262c8ee743c5d567f6ae760ffb57f551917c4/wgpu-core/src/resource.rs#L196

kvark:
Right, this is going in a different direction. Is this direction better?

Neo-Zhixing:
I don't think there's a better direction. Sometimes you have a raw command buffer and you want a wgpu command encoder. Some other time you have a wgpu command buffer and you want the raw one. In my use case, CommandEncoder to raw command buffer is a lot more common, so this PR only addresses one of them.  The other can be added when needed.

kvark:
Is there really a case where you have to use an existing wgpu encoder? I.e. is the work you are going to encode contained within a render pass, or is it outside of a render pass?

Neo-Zhixing:
It is not exactly contained within a render pass since it's a raytracing command. It's taking the color attachment from previous render passes and use it as a storage texture for the raypass.

Render Pass                      <--- Encoded outside of my project, by bevy
Render Pass                      <--- Encoded outside of my project, by bevy
Render Pass                      <--- Encoded outside of my project, by bevy
cmd_pipeline_barrier        <--- Encoded in my project
cmd_trace_rays                 <--- Encoded in my project

So these commands still need to be encoded in the existing command encoder alongside other render passes.

kvark:
Ok, that makes sense. So it sounds like your commands do *not* have to be encoded in the same encoder. You can have an entirely separate encoder for the RT stuff:
```
cmd_pipeline_barrier <--- Encoded in my project
cmd_trace_rays <--- Encoded in my project
```
Then you'd submit them together:
```rust
queue.submit(vec![bevys_encoder.finish(), your_encoder.finish()])`
```

What is not clear to me is how you are going to insert the pipeline barriers. Wgpu assumes full control over resources. If you insert any barriers, you need to very precisely understand how this would interact with any barriers that wgpu itself inserts.

PR #1921, opened 2021-09-08
hal/dx12: fix offsets into storage buffer bindings

encounter:
data.offset is in bytes, and format is in DXGI_FORMAT_R32_TYPELESS


kvark (approved):
Thank you!

kvark:
published in wgpu-hal-0.10.4

PR #1923, opened 2021-09-08
Implement min_{uniform,storage}_buffer_offset_alignment limits

encounter:
**Description**
This implements the `minUniformBufferOffsetAlignment` and `minStorageBufferOffsetAlignment` limits from the WebGPU spec.
https://www.w3.org/TR/webgpu/#dom-supported-limits-minuniformbufferoffsetalignment

This also adds a convenience method `Limits::using_alignment`, similarly to `Limits::using_resolution`.

`wgpu::BIND_BUFFER_ALIGNMENT` was _removed_ in favor of the limits, therefore this is a breaking change.

**Testing**
Manually tested on Vulkan, other backends TBD.

(Open to any and all feedback!)

### DIFF
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -196,15 +196,6 @@ impl RenderBundleEncoder {
 
                     let offsets = &base.dynamic_offsets[..num_dynamic_offsets as usize];
                     base.dynamic_offsets = &base.dynamic_offsets[num_dynamic_offsets as usize..];
-                    // Check for misaligned offsets.
-                    if let Some(offset) = offsets
-                        .iter()
-                        .map(|offset| *offset as wgt::BufferAddress)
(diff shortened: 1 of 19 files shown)


wgpu-core/src/command/bundle.rs
```diff
@@ -220,6 +211,30 @@ impl RenderBundleEncoder {
                         .map_pass_err(scope);
                     }
 
+                    // Check for misaligned offsets.
+                    for (offset, info) in offsets
+                        .iter()
+                        .map(|offset| *offset as wgt::BufferAddress)
+                        .zip(bind_group.dynamic_binding_info.iter())
+                    {
+                        let (alignment, limit_name) = match info.binding_type {
```
kvark, line 220: could we avoid repeating this `match` 3 times in the code?

encounter: Updated in 15cfcc3ab5055ca387ed4c0ce037fb9ffaff1baa, does that look good?

kvark (approved):

PR #1926, opened 2021-09-08
Work around Adreno's OpName bug

kvark:
**Connections**
Based on https://github.com/gfx-rs/naga/pull/1334

**Description**
Works around Adreno bug with `OpName`

**Testing**
Manually tested


kvark:
Published in wgpu-hal-0.10.4

PR #1928, opened 2021-09-09
Expose Metal surface creation

kvark:
**Connections**
Matrix chat

**Description**
This is the optimal way of surface initialization on Metal.

**Testing**
Untested


kvark:
Published in wgpu-0.10.2

PR #1929, opened 2021-09-09
Still close the encoder on error

kvark:
**Connections**
Fixes a crash in Gecko.

**Description**
If a command buffer encountered some issue, we still want to stop encoding properly.

**Testing**
TODO


kvark:
Published in wgpu-core-0.10.4

PR #1932, opened 2021-09-09
web: Add Instance::create_surface_from_canvas (fix #1837)

### LINKED ISSUE #1837: Support for OffscreenCanvas in Web Backend

I've looked into the way the web backend creates the rendering context:

https://github.com/gfx-rs/wgpu/blob/62f26367cd99dde081afdde16eda5ad9077c53d0/wgpu/src/backend/web.rs#L934-L957

This seems like a hack, since I can't pass in a canvas directly, I have to create it elsewhere, insert it into the DOM, assign a random id and then pass in the id.

This is fine for demo applications, but when you have an existing web application you want to integrate into, this doesn't seem like a great idea, since the developer has to be careful not to re-use the same id multiple times if there are multiple canvases in the DOM.

Also, eventually it'll be interesting to use wgpu-rs to render in a Web Worker using OffscreenCanvas. Right now, OffscreenCanvas doesn't support webgpu anyways, but as far as my research has unearthed, this will change at some point.

Not passing in a `raw_handle`, but a `HtmlCanvasElement` or `OffscreenCanvas` directly. Both supply the `get_context` API needed for the backend. Alternatively, the backend could also receive the context directly, but that would require the application to set it up with the right parameters, which might not be trivial (though it looks like it is right now).

I've looked into whether I can skip the function shown above and create this context directly, but all of the types needed aren't `pub`.

**Additional context**

This is more relevant for the WebGL2 Backend #1686 right now, where it would really work.

Herschel:
Add specialized methods on web for creating a wgpu Surface directly
from a `web_sys::HTMLCanvasElement` and `web_sys::OffscreenCanvas`.

 * `Instance::create_surface_from_canvas`
 * `Instance::create_surface_from_offscreen_canvas`

**Connections**
Fixes #1837.

**Description**
Currently [`Instance::create_surface`](https://github.com/gfx-rs/wgpu/blob/ca6fd97c6250c084962a8b8ab153d2f6975d8277/wgpu/src/lib.rs#L1460) requires a `HasRawWindowHandle`. The web backend uses this handle to search for a canvas with a [`data-raw-handle` attribute](https://github.com/gfx-rs/wgpu/blob/master/wgpu/src/backend/web.rs#L945). 

This is handy for full-screen demos or games, or quickly compiling winit apps to both desktop and web. However, this isn't ideal for apps that are part of a larger site, or usable as a JS library. For example, the user may want to instantiate wgpu without attaching the canvas to the DOM, or the canvas may be buried inside a shadow DOM. Additionally, it'd be nice to use `OffscreenCanvas`. See #1837 for more info.

This adds web-specific `Instance::create_surface_from_canvas` and `Instance::create_surface_from_offscreen_canvas` methods that directly accept `web_sys::HtmlCanvasElement` and `web_sys::OffscreenCanvas`. These methods are only compiled for the `wasm32` architecture. This was modeled after the `Instance::create_surface_from_core_animation_layer` method for macOS.

This does add `web_sys` types to the externally visible API. However, `web_sys` is already a dependency of the project, and as WebGPU is a web-based API, wgpu should work directly with standard web types.

### DIFF
--- a/wgpu/Cargo.toml
+++ b/wgpu/Cargo.toml
@@ -269,6 +269,7 @@ web-sys = { version = "0.3.53", features = [
     "GpuVertexState",
     "GpuVertexStepMode",
     "HtmlCanvasElement",
+    "OffscreenCanvas",
     "Window",
 ]}
 js-sys = "0.3.50"
--- a/wgpu/src/backend/web.rs
+++ b/wgpu/src/backend/web.rs
@@ -887,6 +887,30 @@ fn future_map_async(result: JsFutureResult) -> Result<(), crate::BufferAsyncErro
     result.map(|_| ()).map_err(|_| crate::BufferAsyncError)
 }
 
+impl Context {
+    pub fn instance_create_surface_from_canvas(
+        &self,
+        canvas: &web_sys::HtmlCanvasElement,
+    ) -> <Self as crate::Context>::SurfaceId {
+        let context: wasm_bindgen::JsValue = match canvas.get_context("webgpu") {
+            Ok(Some(ctx)) => ctx.into(),
+            _ => panic!("expected to get context from canvas"),
+        };
+        Sendable(context.into())
+    }
+
+    pub fn instance_create_surface_from_offscreen_canvas(
+        &self,
+        canvas: &web_sys::OffscreenCanvas,
+    ) -> <Self as crate::Context>::SurfaceId {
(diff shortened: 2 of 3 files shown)


grovesNL (approved):
Nice, thank you!

This seems good to me assuming we're ok with having web-sys types/version as part of the public API. I'm not sure of a way around this besides having a common canvas trait that web-sys implements (like we do with raw-window-handle)

PR #1933, opened 2021-09-10
Do MTL feature check for Depth Clamping

geertbleyen:
**Connections**
New discovered issue

**Description**
the DEPTH_CLAMPING feature was being enabled for all metal devices, while on iOS GPUFamily4 or newer is needed.
This caused the iOS simulator to assert on calls to to set_depth_clip_mode while Metal API validation is enabled.

**Testing**
Build internal application for both iOS simulator and iOS physical device.
On physical device (iPad Pro 2020 12.9") depth clamping feature is correctly enabled. On iOS simulator for the same device, the feature is correctly not enabled.

### DIFF
--- a/wgpu-hal/src/metal/adapter.rs
+++ b/wgpu-hal/src/metal/adapter.rs
@@ -852,7 +852,7 @@ impl super::PrivateCapabilities {
             },
             //Depth clipping is supported on all macOS GPU families and iOS family 4 and later
             supports_depth_clamping: device.supports_feature_set(MTLFeatureSet::iOS_GPUFamily4_v1)
-            || device.supports_feature_set(MTLFeatureSet::macOS_GPUFamily1_v1)
+                        || os_is_mac
         }
     }
 
--- a/wgpu-hal/src/metal/adapter.rs
+++ b/wgpu-hal/src/metal/adapter.rs
@@ -852,7 +852,7 @@ impl super::PrivateCapabilities {
             },
             //Depth clipping is supported on all macOS GPU families and iOS family 4 and later
             supports_depth_clamping: device.supports_feature_set(MTLFeatureSet::iOS_GPUFamily4_v1)
-                        || os_is_mac
+                || os_is_mac,
         }
     }
 
@@ -866,7 +866,7 @@ impl super::PrivateCapabilities {
             | F::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES
             | F::POLYGON_MODE_LINE
             | F::CLEAR_COMMANDS;
-        
+
         features.set(F::DEPTH_CLAMPING, self.supports_depth_clamping);
 
         features.set(
--- a/wgpu-hal/src/metal/adapter.rs
+++ b/wgpu-hal/src/metal/adapter.rs
@@ -853,17 +853,22 @@ impl super::PrivateCapabilities {
         }
     }
 
-    pub fn features(&self) -> wgt::Features {
+    pub fn features(&self, device: &mtl::Device) -> wgt::Features {
         use wgt::Features as F;
 
         let mut features = F::empty()
-            | F::DEPTH_CLAMPING
             | F::TEXTURE_COMPRESSION_BC
             | F::MAPPABLE_PRIMARY_BUFFERS
             | F::VERTEX_WRITABLE_STORAGE
             | F::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES
             | F::POLYGON_MODE_LINE
             | F::CLEAR_COMMANDS;
+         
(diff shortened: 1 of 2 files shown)


wgpu-hal/src/metal/adapter.rs
```diff
@@ -853,17 +853,22 @@ impl super::PrivateCapabilities {
         }
     }
 
-    pub fn features(&self) -> wgt::Features {
+    pub fn features(&self, device: &mtl::Device) -> wgt::Features {
```
kvark, line 856: instead of passing the device here, it should have a field like `supports_depth_clamping`, which is initialized earlier, like every other thing in here

kvark (changes requested):
Thank you for fixing this!
Just one small adjustment is needed

wgpu-hal/src/metal/adapter.rs
```diff
@@ -850,20 +850,26 @@ impl super::PrivateCapabilities {
...
-            | F::DEPTH_CLAMPING
             | F::TEXTURE_COMPRESSION_BC
             | F::MAPPABLE_PRIMARY_BUFFERS
             | F::VERTEX_WRITABLE_STORAGE
             | F::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES
             | F::POLYGON_MODE_LINE
             | F::CLEAR_COMMANDS;
+        
+        if self.supports_depth_clamping {
+            features |= F::DEPTH_CLAMPING;
```
kvark, line 871: nit: see how this is done for other flags, i.e.
```rust
features.set(..., self.supports_depth_clamping);
```

wgpu-hal/src/metal/adapter.rs
```diff
@@ -850,20 +850,26 @@ impl super::PrivateCapabilities {
             } else {
                 Self::version_at_least(major, minor, 11, 0)
             },
+            //Depth clipping is supported on all macOS GPU families and iOS family 4 and later
+            supports_depth_clamping: device.supports_feature_set(MTLFeatureSet::iOS_GPUFamily4_v1)
+            || device.supports_feature_set(MTLFeatureSet::macOS_GPUFamily1_v1)
```
kvark, line 855: this is really nitty, but we should use `os_is_mac` for consistency

wgpu-hal/src/metal/adapter.rs
geertbleyen: (facepalm).
Will do

kvark (approved):

wgpu-hal/src/metal/adapter.rs
geertbleyen: you mean instead of checking MTLFeatureSet::macOS_GPUFamily1_v1 featureset support?
```
supports_depth_clamping: device.supports_feature_set(MTLFeatureSet::iOS_GPUFamily4_v1)
             || os_is_mac
```
You're right, may be a bit confusing to do the same featureset check multiple times.

kvark: yeah

geertbleyen: Done

wgpu-hal/src/metal/adapter.rs
geertbleyen: Done

wgpu-hal/src/metal/adapter.rs
```diff
@@ -850,20 +850,24 @@ impl super::PrivateCapabilities {
             } else {
                 Self::version_at_least(major, minor, 11, 0)
             },
+            //Depth clipping is supported on all macOS GPU families and iOS family 4 and later
+            supports_depth_clamping: device.supports_feature_set(MTLFeatureSet::iOS_GPUFamily4_v1)
+                        || os_is_mac
```
kvark, line 855: just double-checking if `cargo fmt` agrees :)

wgpu-hal/src/metal/adapter.rs
geertbleyen: Done

wgpu-hal/src/metal/adapter.rs
geertbleyen: Unless I"m doing it wrong, it's a'ok :D

kvark:
Looks like you aren't formatting it, or using a much different version of rust formatter?

geertbleyen:
> Looks like you aren't formatting it, or using a much different version of rust formatter?
Nah, I was having a brainfart. It did format it, I completely missed it :|

kvark:
@geertbleyen are you fine using the master branch, or do you need it on crates instead?

kvark:
Published in wgpu-hal-0.10.6

PR #1937, opened 2021-09-10
web: Return None if GPU.requestAdapter returns null

Herschel:
**Description**
The [WebGPU spec](https://www.w3.org/TR/webgpu/#dom-gpu-requestadapter) says that `GPU.requestAdapter` may return `null`. For example, this happens if hardware acceleration is disabled in Chrome.

Currently `From::from` is used to cast the `JsValue` into `web_sys::GpuAdapter`. However, this will cause an unhandled TypeError in JS if the value is `null` and the cast fails.

Instead, use `JsCast::dyn_into` to properly check the cast and return `None` if it fails. Another option is to manually check for `JsValue::NULL`, but `dyn_into` feels more robust in case of  an unexpected object type.


grovesNL (approved):
Thank you! `dyn_into` sounds fine here

Issue #1936, opened 2021-09-10
wgpu exhausts NSDrawable resources from CAMetalLayer when window isn't visible


xacrimon:
**Description**
wgpu exhausts NSDrawable resources from CAMetalLayer when window isn't visible. Every frame, wgpu calls `nextDrawable` on `CAMetalLayer` in the Metal API to fetch a backbuffer to render onto. When the window isn't in focus or visible, the pool gets exhausted and wgpu panics.

**Repro steps**
- Create any example that unconditionally renders on `WindowEvent::MainEventsCleared`.
- Run the program.
- Hide it or put another window over it and remove focus.

**Expected vs observed behavior**
I expected wgpu stop rendering when the window isn't visible (when `NSWindowOccludedState`) is false. This is because one of the power saving optimizations on M1 is to stop some of Metal's housekeeping tasks such as object reclaiming when the window isn't in focus. The observed behaviour is that wgpu doesn't not respect this or expose an option to find out when the window is visible or not. This quickly exhausts the pool and causes a panic.

**Extra materials**
![Screen Shot 2021-09-10 at 19 42 21](https://user-images.githubusercontent.com/21025159/132895353-fed798b5-246f-48c2-b9b8-edd844e4ccda.png)

**Platform**
M1 MacBook Air with Big Sur

xacrimon:
Filed https://github.com/rust-windowing/winit/issues/2022 so the user can conditionally render.

kvark:
#1938 was merged, but now I realize the solution needed to be a little different. @xacrimon would you be able to change it?

kvark:
closed by #1946

PR #1938, opened 2021-09-10
Return appropriate error when failing to allocate drawable on metal

xacrimon:
**Connections**
Supplies an appropriate error when hitting behaviour described in #1936 

**Description**
Returns an appropriate error instead of causing an uncatchable panic and putting the library in an unusable state.

### DIFF
--- a/wgpu-hal/src/metal/surface.rs
+++ b/wgpu-hal/src/metal/surface.rs
@@ -268,9 +268,12 @@ impl crate::Surface<super::Api> for super::Surface {
     ) -> Result<Option<crate::AcquiredSurfaceTexture<super::Api>>, crate::SurfaceError> {
         let render_layer = self.render_layer.lock();
         let (drawable, texture) = autoreleasepool(|| {
-            let drawable = render_layer.next_drawable().unwrap();
-            (drawable.to_owned(), drawable.texture().to_owned())
-        });
+            if let Some(drawable) = render_layer.next_drawable() {
+                Ok((drawable.to_owned(), drawable.texture().to_owned()))
+            } else {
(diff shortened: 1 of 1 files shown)


kvark (approved):
Thank you!

kvark (approved):

kvark:
@xacrimon hi! Sorry, I missed something very important. This case shouldn't return an error at all, since it wouldn't be actionable. Instead, `acquire_texture` should return `Ok(None)`.

xacrimon:
@kvark Sure! #1946  to fix this

wgpu-hal/src/metal/surface.rs
```diff
@@ -268,9 +268,14 @@ impl crate::Surface<super::Api> for super::Surface {
     ) -> Result<Option<crate::AcquiredSurfaceTexture<super::Api>>, crate::SurfaceError> {
         let render_layer = self.render_layer.lock();
         let (drawable, texture) = autoreleasepool(|| {
-            let drawable = render_layer.next_drawable().unwrap();
```
kvark, line 271: ok, I'm confused. How does this work?
This unwrap didn't panic, so what difference does it make to check if it's `Some` now?

xacrimon: That unwrap did indeed panic?

kvark: Hmm. You are right. I thought we are only exhausting memory in this state.

xacrimon: That happens because `MTLDrawable`'s are limited to 3 active at maximum as per Metal documentation. The job that cleans these is the same one that cleans other resources.

PR #1944, opened 2021-09-12
[hal/metal] Fix stencil ops

### LINKED ISSUE #1943: Stencil texture causing weird behavior when updating from 0.9 to 0.10
**Description**
When I updated my project from wgpu 0.9 to wgpu 0.10, I started seeing some glitchy rendering behavior on MacOS. This seems to happen when drawing to and reading from the stencil texture. I have below included traces and screenshots of the behavior.

**Repro steps**
I have not created a minimum example, but have included the traces both using wgpu 0.9 and 0.10. Hopefully this is enough :)

I expect to see the same rendered output for both versions.
https://youtu.be/eXKrGu2HvM0 (The correctly working version using wgpu 0.9)
vs
https://youtu.be/GFsP1rLP2_g (The version I see strange behavior using wgpu 0.10)

**Extra materials**
I tried doing some digging using your debugging with xcode tutorial (quite nice tutorial), but I am not sure what exactly I should be looking for (I have not really used the debugger before). I seem to not have metal validation available, maybe because I am using an older mac.

From what I did find it seems like it is rendering to 1 in 4 pixels, and that might cause everything to show up strangely. 
The first screenshot is what I see when running with wgpu 0.9:
Skærmbillede 2021-09-12 kl  11 15 36
It looks as expected having nice white pixels where I have my stencil texture.
The second screenshot is what I see when running with wgpu 0.10 and the texture looks like it has been rendered wrong.
Skærmbillede 2021-09-12 kl  11 04 51
An export of the stencil texture can be seen below:
Stencil: Depth texture descriptor

There are some warnings that I am binding the same texture multiple times but these errors exist in both versions.

Below I have included traces for both versions.
[trace0_9.zip](https://github.com/gfx-rs/wgpu/files/7149353/trace0_9.zip)
and
[trace0_10.zip](https://github.com/gfx-rs/wgpu/files/7149355/trace0_10.zip)

**Platform**
Wgpu v0.10
MacOS 10.15 Catalina
Intel Iris Pro 1536 MB
Rust 1.51

kvark:
**Connections**
Fixes #1943

**Description**
It was a copy/paste error in the code.

**Testing**
Tested on the API trace.

### DIFF
--- a/wgpu-hal/src/metal/command.rs
+++ b/wgpu-hal/src/metal/command.rs
@@ -440,13 +440,13 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
                 let at_descriptor = descriptor.stencil_attachment().unwrap();
                 at_descriptor.set_texture(Some(&at.target.view.raw));
 
-                let load_action = if at.depth_ops.contains(crate::AttachmentOps::LOAD) {
+                let load_action = if at.stencil_ops.contains(crate::AttachmentOps::LOAD) {
                     mtl::MTLLoadAction::Load
                 } else {
                     at_descriptor.set_clear_stencil(at.clear_value.1);
                     mtl::MTLLoadAction::Clear
                 };
-                let store_action = if at.depth_ops.contains(crate::AttachmentOps::STORE) {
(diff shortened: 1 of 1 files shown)


kvark:
published in wgpu-hal-0.10.5

PR #1950, opened 2021-09-12
hal/metal: fix storage binding size units

### LINKED ISSUE #1830: Mostly black frames on Metal backend
**Description**
The output frame is mostly black or completely black.

**Repro steps**
Clone and run https://github.com/niklaskorz/linon on macOS.

Expected (wgpu 0.9, commit `8f8f4ed815e3a7d6fb78c13cb0a23a5f67555c54` on my repository):

grafik

Observed one of these two, different with each run (wgpu 0.10, commit `5c1ea66fb88916f1d5eea1ce03a1e44cfbbd5862` on my repository):

1. grafik
2. Bildschirmfoto 2021-08-19 um 16 04 13

To be precise, sometimes the whole screen is just black, sometimes only the view on the lower left is drawn.
The final render pass happens in egui, the two views on the bottom left and on the right are drawn into textures that are then rendered by egui.

**Platform**
- macOS 11.5.2

kvark:
**Connections**
Fixes #1830

**Description**
There was a messup between units. wgpu-hal passed u64 buffer sizes where MTL sources expected u32.
So the second size always got 0, and the third was wrong, etc.

**Testing**
Manually tested

### DIFF
--- a/wgpu-hal/src/metal/command.rs
+++ b/wgpu-hal/src/metal/command.rs
@@ -1,6 +1,7 @@
 use super::{conv, AsNative};
 use std::{mem, ops::Range};
 
+// has to match `Temp::binding_sizes`
 const WORD_SIZE: usize = 4;
 
 impl Default for super::CommandState {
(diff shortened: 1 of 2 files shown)


kvark:
Included in wgpu-hal-0.10.6

PR #1952, opened 2021-09-13
hal/metal: fix the back-face stencil state

kvark:
**Connections**
Fixes https://github.com/gfx-rs/wgpu/issues/1943#issuecomment-918192751
Follow-up to  #1944

**Description**
Another copy-paste error

**Testing**
Untested


kvark:
published in wgpu-hal-0.10.7

PR #1956, opened 2021-09-14
Improve CI testing code

cwfitzgerald:
**Connections**

Follow up to #1954. 

**Description**

This makes our tests in CI a bit clearer. 

Currently draft to test CI.

**Testing**

It is!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
     tags: [v0.*]
   pull_request:
 
+env:
+  RUST_BACKTRACE: 1
+
 jobs:
   build:
     strategy:
@@ -18,6 +21,13 @@ jobs:
             channel: stable
             target: x86_64-pc-windows-msvc
             kind: test
+            backends: dx12 # dx11
+
+          - name: Windows Nightly x86_64
+            os: windows-2019
+            channel: nightly
+            target: x86_64-pc-windows-msvc
+            kind: compile
           
           # MacOS
 
@@ -42,7 +52,8 @@ jobs:
             os: ubuntu-20.04
             channel: stable
             target: x86_64-unknown-linux-gnu
-            kind: test
+            kind: compile
+            backends: # vulkan gl
 
           - name: Linux Nightly x86_64
             os: ubuntu-20.04
@@ -71,7 +82,7 @@ jobs:
 
     env:
       PKG_CONFIG_ALLOW_CROSS: 1 # allow android to work
-      RUSTFLAGS: --cfg=web_sys_unstable_apis
+      RUSTFLAGS: --cfg=web_sys_unstable_apis -D warnings
       RUSTDOCFLAGS: -Dwarnings
 
     steps:
@@ -87,6 +98,11 @@ jobs:
           profile: minimal
           components: clippy
 
+      - name: caching
+        uses: Swatinem/rust-cache@v1
+        with:
+          key: ${{ matrix.target }}-a # suffix for cache busting
+
       - name: add android apk to path
         if: matrix.os == 'ubuntu-20.04' && matrix.target == 'aarch64-linux-android'
         run: |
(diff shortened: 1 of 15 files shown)


cwfitzgerald:
Alright, this is good to go. The required checks need to be adjusted to include windows and exclude the nightlies, but I'm quite happy with this result.

.github/workflows/ci.yml
```diff
@@ -95,80 +111,77 @@ jobs:
...
+          for backend in ${{ matrix.backends }}; do
+            echo "======= NATIVE TESTS $backend ======";
+            # run player tests
+            WGPU_BACKEND=$backend cargo test --target ${{ matrix.target }} -p wgpu-types -p wgpu-hal -p wgpu-core --no-fail-fast -- --nocapture --test-threads=1
+            # run coretests
+            WGPU_BACKEND=$backend cargo test --target ${{ matrix.target }} -p wgpu --no-fail-fast -- --nocapture --test-threads=1
+          done
 
-  test:
+  ctx:
```
lucacasonato, line 165:
```suggestion
  cts:
```

deno_webgpu/src/lib.rs
```diff
@@ -242,13 +242,14 @@ pub async fn op_webgpu_request_adapter(
 ) -> Result<GpuAdapterDeviceOrErr, AnyError> {
     let mut state = state.borrow_mut();
     check_unstable(&state, "navigator.gpu.requestAdapter");
+    let backends = std::env::var("WGPU_BACKEND").ok().and_then(|s| wgpu_types::Backends::from_comma_list(&s)).unwrap_or_else(wgpu_types::Backends::all);
```
lucacasonato, line 245:
```suggestion
    let backends = std::env::var("DENO_WEBGPU_BACKEND").ok().and_then(|s| wgpu_types::Backends::from_comma_list(&s)).unwrap_or_else(wgpu_types::Backends::all);
```

All the environment variables Deno cares about should be prefixed by `DENO_`

lucacasonato (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -183,16 +196,36 @@ jobs:
...
           profile: minimal
 
+      - name: caching
+        uses: Swatinem/rust-cache@v1
+        with:
+          key: cts-a # suffix for cache busting
+          working-directory: wgpu/cts_runner
+          target-dir: wgpu/cts_runner/target
+
+      # We enable line numbers for panics, but that's it
```
kvark, line 206: is this to reduce the cache size, or for something else?

kvark (commented):
LGTM, minus the parts noted by Luca

.github/workflows/ci.yml
cwfitzgerald: Yeah it is, plus reducing IO load a bit, hopefully making things a bit faster.

wgpu-types/src/lib.rs
```diff
@@ -128,6 +123,39 @@ impl From<Backend> for Backends {
...
+    /// Whitespace is stripped, so both 'gl, dx12' and 'gl,dx12' are valid.
+    //
+    /// Names:
+    /// - vulkan = "vulkan" or "vk"
+    /// - dx12   = "dx12" or "d3d12"
+    /// - dx11   = "dx11" or "d3d11"
+    /// - metal  = "metal" or "mtl"
+    /// - gles   = "opengl" or "gles" or "gl"
+    /// - webgpu = "webgpu"
+    pub fn from_comma_list(string: &str) -> Option<Self> {
```
kvark, line 138: can this live in wgpu-core instead of wgpu-types?

wgpu-types/src/lib.rs
```diff
@@ -128,6 +123,39 @@ impl From<Backend> for Backends {
...
+        let mut backends = Self::empty();
+        for backend in string.to_lowercase().split(',') {
+            backends |= match backend.trim() {
+                "vulkan" | "vk" => Self::VULKAN,
+                "dx12" | "d3d12" => Self::DX12,
+                "dx11" | "d3d11" => Self::DX11,
+                "metal" | "mtl" => Self::METAL,
+                "opengl" | "gles" | "gl" => Self::GL,
+                "webgpu" => Self::BROWSER_WEBGPU,
+                _ => continue,
```
kvark, line 148: let's issue a `log::warn!` about unknown backends

wgpu-types/src/lib.rs
```diff
@@ -128,6 +123,39 @@ impl From<Backend> for Backends {
...
+            }
+        }
+        if backends == Self::empty() {
```
kvark, line 151: `if backends.is empty()`?

wgpu-types/src/lib.rs
```diff
@@ -128,6 +123,39 @@ impl From<Backend> for Backends {
...
+            None
+        } else {
+            Some(backends)
```
kvark, line 154: actually, why are we doing this? If the user has a string, we should just assume they know what they are doing. Let's return the bitflags instead of an option

kvark (changes requested):

wgpu-types/src/lib.rs
cwfitzgerald: Addressed.

wgpu-types/src/lib.rs
cwfitzgerald: I changed this in favor of logging when it's empty, as it's _probably_ not intended.

kvark (approved):

PR #1963, opened 2021-09-14
Trigger CTS run by ReadyForReview

kvark:
**Connections**

**Description**
Our time spent in CTS is growing rapidly, and we are just getting started. Most PRs don't really need the run.
What this PR does is trying to make the run to trigger manually. We could do various things, but the approach I found to be most suitable is by doing "ready for review" conversion.

**Testing**
Untested


kvark:
Closing in favor of #1965

PR #1965, opened 2021-09-14
Only run CTS check if 'needs testing' label is added

kvark:
**Connections**
Alternative to #1963

**Description**
Reduces the CI stress and makes us move faster when needed

**Testing**
will be testing here


cwfitzgerald (approved):
Code's fine, just has a conflict

PR #1967, opened 2021-09-15
Add a note that copying within the same buffer panics

jamen:
**Description**
Add a note that `CommandEncoder::copy_buffer_to_buffer` panics if copying within the same buffer.

I couldn't tell if `CommandEncoder::copy_texture_to_texture` is the same, but it seems like you can copy within the same texture.


cwfitzgerald:
Thanks for the fix!

We might not validate it correctly, but the rule is similar for textures:

> [The set of subresources for texture copy(source, copySize) and the set of subresources for texture copy(destination, copySize) is disjoint.](https://gpuweb.github.io/gpuweb/#dom-gpucommandencoder-copytexturetotexture)

kvark (approved):

PR #1975, opened 2021-09-17
Panic with an error by default

kvark:
**Connections**
Too many times we see people missing the actual error message.

**Description**
Flips the error reporting. The notification of "Handling wgpu errors as fatal by default" is informative but not useful.

**Testing**
Should work!


kvark:
Published in wgpu-0.10.2

PR #1978, opened 2021-09-18
Add method to create shader modules without runtime checks

JCapucho:
**Connections**
None

**Description**
Currently wgpu instructs naga to add additional runtime checks when outputting spirv, this checks can be quite heavy. This PR introduces an option to `ShaderModuleDescriptor` to disable this, this option only affects native since only there code generation occurs.

**Testing**
Ran the examples and confirmed that no additional checks were present

### DIFF
--- a/wgpu-core/src/pipeline.rs
+++ b/wgpu-core/src/pipeline.rs
@@ -18,7 +18,7 @@ pub enum ShaderModuleSource<'a> {
 #[cfg_attr(feature = "replay", derive(serde::Deserialize))]
 pub struct ShaderModuleDescriptor<'a> {
     pub label: Label<'a>,
-    pub runtime_checks: bool,
+    pub shader_bound_checks: wgt::ShaderBoundChecks,
 }
 
 #[derive(Debug)]
--- a/wgpu-core/src/pipeline.rs
+++ b/wgpu-core/src/pipeline.rs
@@ -18,6 +18,7 @@ pub enum ShaderModuleSource<'a> {
 #[cfg_attr(feature = "replay", derive(serde::Deserialize))]
 pub struct ShaderModuleDescriptor<'a> {
     pub label: Label<'a>,
+    pub runtime_checks: bool,
 }
 
 #[derive(Debug)]
--- a/wgpu-hal/src/vulkan/device.rs
+++ b/wgpu-hal/src/vulkan/device.rs
@@ -582,15 +582,26 @@ impl super::Device {
         let stage_flags = crate::auxil::map_naga_stage(naga_stage);
         let vk_module = match *stage.module {
             super::ShaderModule::Raw(raw) => raw,
-            super::ShaderModule::Intermediate(ref naga_shader) => {
+            super::ShaderModule::Intermediate {
+                ref naga_shader,
+                runtime_checks,
+            } => {
                 let pipeline_options = naga::back::spv::PipelineOptions {
                     entry_point: stage.entry_point.to_string(),
                     shader_stage: naga_stage,
                 };
+                let mut naga_options = self.naga_options.clone();
+                if !runtime_checks {
+                    naga_options.bounds_check_policies = naga::back::BoundsCheckPolicies {
+                        index: naga::back::BoundsCheckPolicy::Unchecked,
+                        buffer: naga::back::BoundsCheckPolicy::Unchecked,
+                        image: naga::back::BoundsCheckPolicy::Unchecked,
+                    };
+                }
                 let spv = naga::back::spv::write_vec(
                     &naga_shader.module,
                     &naga_shader.info,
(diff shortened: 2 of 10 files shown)


cwfitzgerald:
Unfortunately this setting would have to be unsafe in some way. Naga already shouldn't be generating the relevant checks if RBA is enabled, but if we go beyond that, we need it to be unsafe.

JCapucho:
That's really unfortunate. RBA doesn't solve the entire issue since indexing and image lookups are still restricted. I'll change it to be a different `unsafe` function.

kvark:
I admit this is something native users would need. And it has to be explicitly unsafe.

Currently, we omit checks for buffers if the system supports RBA:
```rust
                 bounds_check_policies: naga::back::BoundsCheckPolicies {
                    index: naga::back::BoundsCheckPolicy::Restrict,
                    image: naga::back::BoundsCheckPolicy::Restrict,
                    buffer: if self.private_caps.robust_buffer_access {
                        naga::back::BoundsCheckPolicy::Unchecked
                    } else {
                        naga::back::BoundsCheckPolicy::Restrict
                    },
                },
```

We'll also make it so image checks are omitted if the system supports RBA2. This is fairly straightforward, and RBA2 coverage is decent. I'll get this rolling shortly.

As for indexing arrays not in buffer memory, there are no plans to omit them.

JCapucho:
Will RBA2 be always enabled if possible or will there be a setting to disable (of course only would work in native) it even if it's available? The extension itself says that it can be quite impactful in performance.

JCapucho:
I've updated the PR with an unsafe method called `create_shader_module_unchecked`

kvark:
> Will RBA2 be always enabled if possible or will there be a setting to disable (of course only would work in native) it even if it's available? The extension itself says that it can be quite impactful in performance.

We have a TODO to make it possible to disable both RBA and runtime checks, but I don't know how this will be seen from the point of view of wgpu user.

wgpu-core/src/pipeline.rs
```diff
@@ -18,6 +18,7 @@ pub enum ShaderModuleSource<'a> {
 #[cfg_attr(feature = "replay", derive(serde::Deserialize))]
 pub struct ShaderModuleDescriptor<'a> {
     pub label: Label<'a>,
+    pub runtime_checks: bool,
```
kvark, line 21: wgpu-core also has `unsafe`-sensitive API. This needs to be unsafe at this level, too.
Maybe wgpu-types can have `struct ShaderBoundChecks` as an opaque type, and we'll add `unsafe fn unchecked() -> Self` constructor?

JCapucho: And how would the backends query if it's unchecked, trough a method?

kvark: when wgpu-core gets this info and passes to wgpu-hal, it can do it as a `bool`, since wgpu-hal is unsafe.

JCapucho: I've updated the PR with your suggestion could you check if it's how you'd expect

wgpu-hal/src/vulkan/device.rs
```diff
@@ -582,15 +582,26 @@ impl super::Device {
...
+            super::ShaderModule::Intermediate {
+                ref naga_shader,
+                runtime_checks,
+            } => {
                 let pipeline_options = naga::back::spv::PipelineOptions {
                     entry_point: stage.entry_point.to_string(),
                     shader_stage: naga_stage,
                 };
+                let mut naga_options = self.naga_options.clone();
+                if !runtime_checks {
```
kvark, line 594: Can we avoid cloning the options if this flag is true? Options contain a hashmap. This is not hugely important at this stage, but fixing seems easy still.

kvark (approved):
Looks great! Just a few nits

wgpu-hal/src/vulkan/device.rs
```diff
@@ -582,18 +582,38 @@ impl super::Device {
...
                     shader_stage: naga_stage,
                 };
-                let spv = naga::back::spv::write_vec(
-                    &naga_shader.module,
-                    &naga_shader.info,
-                    &self.naga_options,
-                    Some(&pipeline_options),
-                )
-                .map_err(|e| crate::PipelineError::Linkage(stage_flags, format!("{}", e)))?;
+                let spv = if !runtime_checks {
```
kvark, line 593: we can do better, actually:
```rust
let temp_options;
let options = if runtime_checks {
  &self.naga_options
} else {
  temp_options = spv::Options {
    bounds_check_policies: ...,
    .. self.naga_options.clone()
  };
  &temp_options // the trick here  <-
};
...
naga::back::spv::write_vec(...)
```

kvark (approved):

PR #1979, opened 2021-09-19
hal/vk: Enable image robustness where available

kvark:
**Connections**
Related to #1978

**Description**
We detect if image robustness is exposed, and enable it in order to save on runtime checks.

**Testing**
Untested


kvark:
@JCapucho noted that the LOD still needs to be checked even with image robustness. Let's follow-up with this.

I tested this on a few platforms:
  - Windows/AMD: not detecting any of the relevant extensions, robust images are disabled
  - Linux/NV: detecting both extensions, enabled robust images

PR #1987, opened 2021-09-21
Fix VK and GL renderdoc captures

cwfitzgerald:
**Description**

Fixes VK and GL renderdoc captures didn't work right in headless, multi-api scenarios. This fixes that.

**Testing**

Manually.


wgpu-hal/src/gles/egl.rs
```diff
@@ -236,7 +236,7 @@ pub struct AdapterContext {
     glow_context: Mutex<glow::Context>,
     egl: Arc<egl::DynamicInstance<egl::EGL1_4>>,
     egl_display: egl::Display,
-    egl_context: egl::Context,
+    pub(crate) egl_context: egl::Context,
```
kvark, line 239: should be `pub(super)` instead

kvark (approved):
Just one nit

PR #1988, opened 2021-09-21
[v0.10] Fix VK and GL renderdoc captures

cwfitzgerald:
**Connections**

Backport of #1987 

**Testing**

This version was actually tested.


kvark:
Could you add a small changelog entry, so that we don't miss it when we publish?

cwfitzgerald:
Added.

kvark (approved):

PR #1989, opened 2021-09-22
Support zero-sized mappings

kvark:
**Connections**
CTS

**Description**
Mapping an empty buffer should succeed still, and it should assume the mapped state.
That second part was missing.

**Testing**
Unlocks CTS

### DIFF
--- a/wgpu-core/src/device/life.rs
+++ b/wgpu-core/src/device/life.rs
@@ -17,7 +17,7 @@ use hal::Device as _;
 use parking_lot::Mutex;
 use thiserror::Error;
 
-use std::mem;
+use std::{mem, ptr};
 
 /// A struct that keeps lists of resources that are no longer needed by the user.
 #[derive(Debug, Default)]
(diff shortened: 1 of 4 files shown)


kvark:
The workaround for the empty mapping is rather hacky. Open to better ideas :)

wgpu-core/src/device/life.rs
```diff
@@ -674,28 +676,34 @@ impl<A: HalApi> LifetimeTracker<A> {
...
+                        Ok(ptr) => Some(ptr),
                         Err(e) => {
                             log::error!("Mapping failed {:?}", e);
-                            resource::BufferMapAsyncStatus::Error
+                            None
                         }
                     }
                 } else {
-                    resource::BufferMapAsyncStatus::Success
+                    ptr::NonNull::new(self.empty_mapping.as_mut_ptr())
```
cwfitzgerald, line 692: Couldn't we just do `ptr::NonNull::new((&mut []).as_mut_ptr())`? We'd be recreating a zero sized slice on the other side, unless for some reason gecko needs to deref this pointer.

kvark: I don't think this is valid. A `*mut u8` isn't holding `[]` alive, so the `[]` will die instantly, and the pointer becomes invalid.

cwfitzgerald: That's the trick though, a zero size slice is pointing to nothing, which is why `&mut []` is of lifetime `'static`. The way that you can show this is that it's valid to create a zero-size slice to an empty vector which hasn't allocated on the heap at all.

kvark: that's tricky indeed :/

kvark:
Weird, CTS fails on Windows but passes on macOS.

cwfitzgerald:
Closing this as out of date.

PR #1991, opened 2021-09-22
Fix write_texture for array textures

### LINKED ISSUE #1990: `Queue::write_texture` appears to require row padding for `Texture`s with depth > 1 on Linux/Vulkan
**Description**

Partly to workaround #1672 and partly to achieve better efficiency, I rewrote nannou's `Texture` constructor short-hand functions to use `Queue::write_texture`, rather than creating and submitting a command buffer each time. One of these updated functions loads a 3D texture from a sequence of image buffers with equal dimensions (see [here](https://github.com/nannou-org/nannou/blob/4f797869fca282f3dac756227c091c768a8e7886/nannou_wgpu/src/texture/image.rs#L590-L664)).

After changing this function to use `Queue::write_texture`, I noticed that our `wgpu_image_sequence.rs` example began to show strange behaviour. The example loads a sequence of small PNGs into a single texture to demonstrate how a single 3D texture can be used for animation. When creating an texture view from any layer other than the first, the view appears vertically offset, which suggested to me that the texture's image data layout might be incorrect. Here's example of how the offset looks:

Screenshot from 2021-09-22 10-52-00

Curiously, this only occurs on Linux + Vulkan and doesn't seem to occur on the default macOS and Windows backends.

I noticed that if I manually pad the rows to meet the multiple-of-256-bytes requirement mentioned [in the docs of the `bytes_per_row` field](https://docs.rs/wgpu/0.10.1/wgpu/struct.ImageDataLayout.html#structfield.bytes_per_row), the problem goes away, however the same docs also mention that the requirement should not apply when using `Queue::write_texture`. I can confirm that the requirement doesn't apply when creating single layer textures (these appear to work fine across platforms) - it only appears to be Linux/Vulkan that's the issue.

**Repro steps**

Fetch the branch of [this PR](https://github.com/nannou-org/nannou/pull/795) and `cargo run --example wgpu_image_sequence` from the workspace root on Linux+Vulkan (haven't yet tried Windows+Vulkan).

**Platform**

Linux, GNOME, Wayland, Intel iGPU, Vulkan, wpgu 0.10.1 ([this PR](https://github.com/nannou-org/nannou/pull/795)).

kvark:
**Connections**
Fixes #1990

**Description**
The logic used the original bytes-per-image for offseting the wgpu-hal copy, while it needed to use the staging one.

**Testing**
Not really

### DIFF
--- a/wgpu-core/src/device/queue.rs
+++ b/wgpu-core/src/device/queue.rs
@@ -371,7 +371,9 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         let (selector, dst_base, texture_format) =
             extract_texture_selector(destination, size, &*texture_guard)?;
         let format_desc = texture_format.describe();
-        let (_, bytes_per_array_layer) = validate_linear_texture_data(
+        //Note: `_source_bytes_per_array_layer` is ignored since we have a staging copy,
+        // and it can have a different value.
+        let (_, _source_bytes_per_array_layer) = validate_linear_texture_data(
             data_layout,
             texture_format,
             data.len() as wgt::BufferAddress,
@@ -444,6 +446,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             .map_err(DeviceError::from)?;
         unsafe {
             profiling::scope!("copy");
+
             if stage_bytes_per_row == bytes_per_row {
                 // Fast path if the data is already being aligned optimally.
                 ptr::copy_nonoverlapping(
@@ -488,7 +491,9 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             texture_base.array_layer += rel_array_layer;
             hal::BufferTextureCopy {
                 buffer_layout: wgt::ImageDataLayout {
-                    offset: rel_array_layer as u64 * bytes_per_array_layer,
+                    offset: rel_array_layer as u64
+                        * block_rows_per_image as u64
+                        * stage_bytes_per_row as u64,
                     bytes_per_row: NonZeroU32::new(stage_bytes_per_row),
                     rows_per_image: NonZeroU32::new(block_rows_per_image),
                 },


wgpu-core/src/device/queue.rs
```diff
@@ -444,6 +446,10 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             .map_err(DeviceError::from)?;
         unsafe {
             profiling::scope!("copy");
+            println!(
+                "Copy stage bpr = {}, original = {}",
+                stage_bytes_per_row, bytes_per_row
+            );
```
cwfitzgerald, line 452: Errant print

wgpu-core/src/device/queue.rs
```diff
@@ -371,7 +371,9 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         let (selector, dst_base, texture_format) =
             extract_texture_selector(destination, size, &*texture_guard)?;
         let format_desc = texture_format.describe();
-        let (_, bytes_per_array_layer) = validate_linear_texture_data(
+        //Note: `_source_bytes_per_array_layer` is ignored since we have a staging copy,
+        // and it can have a different value.
+        let (_, _source_bytes_per_array_layer) = validate_linear_texture_data(
```
cwfitzgerald, line 376: Why aren't we just wholesale ignoring this value (we can leave the comment).

cwfitzgerald (approved):
Good catch. I tried to take a look at this but couldn't catch the issue.

Nits but algo fine.

wgpu-core/src/device/queue.rs
kvark: We could do this. I felt that referring to a field in the comment makes it more clear if the field exists in the code.

wgpu-core/src/device/queue.rs
kvark: ooops

PR #1995, opened 2021-09-24
Support descriptor arrays which are shorter than the BGL

### LINKED ISSUE #1111: Use VK_DESCRIPTOR_BINDING_UPDATE_AFTER_BIND_BIT_EXT when DI is enabled
**Description**

Intel doesn't support high counts of bindings in one bind group (250 being the limit) without using UPDATE_AFTER_BIND_BIT on the descriptors. It's my impression that there are no extra requirements for it, so we should enable it on the descriptors, sets, and pools in the vulkan backend when DI is enabled. 

I need to double-check things before we implement it.

### LINKED ISSUE #1635: Support BindGroups with fewer elements in Binding Array than in BindGroupLayoutEntry
Current wgpu could support unsized binding array via `Features::UNSIZED_BINDING_ARRAY`. However, users still have to manually set the array size of the binding in `BindGroupLayoutEntry`. Is it possible for wgpu to fully support unsized binding array?

Maybe we could provide an API to let users specify a maximum array size

**Additional context**

cwfitzgerald:
**Connections**

Closes #1635
Closes #1111

**Description**

This allows BG arrays to have fewer elements than are in the BGLs. As long as that element count is non-zero.

**Testing**

Rend3

### DIFF
--- a/wgpu-hal/src/lib.rs
+++ b/wgpu-hal/src/lib.rs
@@ -188,7 +188,11 @@ pub trait Surface<A: Api>: Send + Sync {
 }
 
 pub trait Adapter<A: Api>: Send + Sync {
-    unsafe fn open(&self, features: wgt::Features) -> Result<OpenDevice<A>, DeviceError>;
+    unsafe fn open(
+        &self,
+        features: wgt::Features,
+        limits: &wgt::Limits,
+    ) -> Result<OpenDevice<A>, DeviceError>;
 
     /// Return the set of supported capabilities for a texture format.
     unsafe fn texture_format_capabilities(
@@ -525,6 +529,14 @@ bitflags!(
 );
 
 bitflags!(
+    /// Pipeline layout creation flags.
+    pub struct BindGroupLayoutFlags: u32 {
+        /// Allows for bind group binding arrays to be shorter than the array in the BGL.
+        const PARTIALLY_BOUND = 1 << 0;
+    }
+);
+
+bitflags!(
     /// Texture format capability flags.
     pub struct TextureFormatCapabilities: u32 {
         /// Format can be sampled.
@@ -798,6 +810,7 @@ pub struct SamplerDescriptor<'a> {
 #[derive(Clone, Debug)]
 pub struct BindGroupLayoutDescriptor<'a> {
     pub label: Label<'a>,
+    pub flags: BindGroupLayoutFlags,
     pub entries: &'a [wgt::BindGroupLayoutEntry],
 }
 
@@ -847,6 +860,7 @@ impl<A: Api> Clone for TextureBinding<'_, A> {
 pub struct BindGroupEntry {
     pub binding: u32,
     pub resource_index: u32,
+    pub count: u32,
 }
 
 /// BindGroup descriptor.
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -7,15 +7,17 @@ use std::{ffi::CStr, mem, ptr, sync::Arc};
 
 //TODO: const fn?
 fn indexing_features() -> wgt::Features {
-    wgt::Features::BUFFER_BINDING_ARRAY | wgt::Features::TEXTURE_BINDING_ARRAY
+    wgt::Features::SAMPLED_TEXTURE_AND_STORAGE_BUFFER_ARRAY_NON_UNIFORM_INDEXING
+        | wgt::Features::UNIFORM_BUFFER_AND_STORAGE_TEXTURE_ARRAY_NON_UNIFORM_INDEXING
+        | wgt::Features::UNSIZED_BINDING_ARRAY
 }
 
 /// Aggregate of the `vk::PhysicalDevice*Features` structs used by `gfx`.
 #[derive(Debug, Default)]
 pub struct PhysicalDeviceFeatures {
     core: vk::PhysicalDeviceFeatures,
-    vulkan_1_2: Option<vk::PhysicalDeviceVulkan12Features>,
-    descriptor_indexing: Option<vk::PhysicalDeviceDescriptorIndexingFeaturesEXT>,
+    pub(super) vulkan_1_2: Option<vk::PhysicalDeviceVulkan12Features>,
+    pub(super) descriptor_indexing: Option<vk::PhysicalDeviceDescriptorIndexingFeaturesEXT>,
     imageless_framebuffer: Option<vk::PhysicalDeviceImagelessFramebufferFeaturesKHR>,
     timeline_semaphore: Option<vk::PhysicalDeviceTimelineSemaphoreFeaturesKHR>,
     image_robustness: Option<vk::PhysicalDeviceImageRobustnessFeaturesEXT>,
@@ -54,7 +56,29 @@ impl PhysicalDeviceFeatures {
         requested_features: wgt::Features,
         downlevel_flags: wgt::DownlevelFlags,
         private_caps: &super::PrivateCapabilities,
+        uab_types: super::UpdateAfterBindTypes,
     ) -> Self {
+        let needs_sampled_image_non_uniform = requested_features.contains(
+            wgt::Features::TEXTURE_BINDING_ARRAY
+                | wgt::Features::SAMPLED_TEXTURE_AND_STORAGE_BUFFER_ARRAY_NON_UNIFORM_INDEXING,
+        );
+        let needs_storage_buffer_non_uniform = requested_features.contains(
+            wgt::Features::BUFFER_BINDING_ARRAY
+                | wgt::Features::STORAGE_RESOURCE_BINDING_ARRAY
+                | wgt::Features::SAMPLED_TEXTURE_AND_STORAGE_BUFFER_ARRAY_NON_UNIFORM_INDEXING,
+        );
+        let needs_uniform_buffer_non_uniform = requested_features.contains(
+            wgt::Features::TEXTURE_BINDING_ARRAY
+                | wgt::Features::UNIFORM_BUFFER_AND_STORAGE_TEXTURE_ARRAY_NON_UNIFORM_INDEXING,
+        );
+        let needs_storage_image_non_uniform = requested_features.contains(
+            wgt::Features::TEXTURE_BINDING_ARRAY
+                | wgt::Features::STORAGE_RESOURCE_BINDING_ARRAY
+                | wgt::Features::UNIFORM_BUFFER_AND_STORAGE_TEXTURE_ARRAY_NON_UNIFORM_INDEXING,
+        );
+        let needs_partially_bound =
+            requested_features.intersects(wgt::Features::PARTIALLY_BOUND_BINDING_ARRAY);
+
         Self {
             // vk::PhysicalDeviceFeatures is a struct composed of Bool32's while
             // Features is a bitfield so we need to map everything manually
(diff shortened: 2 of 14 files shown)


cwfitzgerald:
Do not merge. There are vulkan issues, specifically we need to turn on VK_DESCRIPTOR_BINDING_PARTIALLY_BOUND_BIT.

cwfitzgerald:
Alright this is ready to go. Needs a thorough comb through as it blew up in scope quite a bit.

wgpu-hal/src/vulkan/device.rs
```diff
@@ -1012,10 +1016,53 @@ impl crate::Device<super::Api> for super::Device {
...
+                    let mut flags = vk::DescriptorBindingFlags::empty();
+
+                    if partially_bound && entry.count.is_some() {
+                        flags |= vk::DescriptorBindingFlags::PARTIALLY_BOUND;
+                    }
+
+                    if self.shared.private_caps.update_after_bind
+                        && !matches!(entry.ty, wgt::BindingType::Sampler { .. })
+                    {
+                        flags |= vk::DescriptorBindingFlags::UPDATE_AFTER_BIND;
```
kvark, line 1051: do we really need to set this flag?
discussed on the matrix: we can be more deliberate for enabling this flag

wgpu-hal/src/vulkan/device.rs
```diff
@@ -377,7 +377,11 @@ impl
             })
             .collect::<ArrayVec<_, 8>>();
 
-        let mut vk_flags = vk::DescriptorPoolCreateFlags::empty();
+        let mut vk_flags = if self.private_caps.update_after_bind {
+            vk::DescriptorPoolCreateFlags::UPDATE_AFTER_BIND
```
kvark, line 381: we shouldn't use `private_caps` here. Instead, the incoming `flags` should be checked for this bit.

wgpu-hal/src/vulkan/device.rs
```diff
@@ -1012,10 +1016,53 @@ impl crate::Device<super::Api> for super::Device {
             })
             .collect::<Vec<_>>();
 
+        let dsl_create_flags = if self.shared.private_caps.update_after_bind {
+            vk::DescriptorSetLayoutCreateFlags::UPDATE_AFTER_BIND_POOL_EXT
```
kvark, line 1020: we should be more deliberate here, and only enable it if any of the BGL entries exceed the base limits

wgpu-hal/src/vulkan/device.rs
```diff
@@ -1012,10 +1016,53 @@ impl crate::Device<super::Api> for super::Device {
...
+
+        let vk_info = if self.shared.private_caps.update_after_bind || partially_bound {
+            binding_flag_vec = desc
+                .entries
+                .iter()
+                .map(|entry| {
+                    let mut flags = vk::DescriptorBindingFlags::empty();
+
+                    if partially_bound && entry.count.is_some() {
+                        flags |= vk::DescriptorBindingFlags::PARTIALLY_BOUND;
```
kvark, line 1045: We need to leave a note somewhere in wgpu-hal API to consider lifting this. This wouldn't make any difference to the end wgpu-rs users tho.

wgpu-hal/src/vulkan/device.rs
```diff
@@ -1085,7 +1132,11 @@ impl crate::Device<super::Api> for super::Device {
         let mut vk_sets = self.desc_allocator.lock().allocate(
             &*self.shared,
             &desc.layout.raw,
-            gpu_descriptor::DescriptorSetLayoutCreateFlags::empty(),
+            if self.shared.private_caps.update_after_bind {
+                gpu_descriptor::DescriptorSetLayoutCreateFlags::UPDATE_AFTER_BIND
```
kvark, line 1136: here the same, we should ideally only pass this flag in if the BGL requires it

wgpu-hal/src/vulkan/mod.rs
```diff
@@ -234,6 +236,7 @@ struct DeviceShared {
     extension_fns: DeviceExtensionFunctions,
     vendor_id: u32,
     timestamp_period: f32,
+    features: wgt::Features,
```
kvark, line 239: something rubs me the wrong way to see features here, only so that "partially bound" can be checked, but I haven't found a sound way that would be significantly better

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -5,9 +5,14 @@ use parking_lot::Mutex;
 
 use std::{ffi::CStr, mem, ptr, sync::Arc};
 
+/// Update After Bind is used on intel only, and all intel devices have at least 524288
+const MAX_UPDATE_AFTER_BIND_DESCRIPTORS_IN_ALL_POOLS: u32 = 524288;
```
kvark, line 9: so if we make it non-Intel, we should grab this number from the adapter instead of hard-coding it here

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -132,32 +158,30 @@ impl PhysicalDeviceFeatures {
...
                         .shader_storage_buffer_array_non_uniform_indexing(
-                            requested_features.contains(
-                                wgt::Features::BUFFER_BINDING_ARRAY
-                                    | wgt::Features::STORAGE_RESOURCE_BINDING_ARRAY
-                                    | wgt::Features::SAMPLED_TEXTURE_AND_STORAGE_BUFFER_ARRAY_NON_UNIFORM_INDEXING,
-                            ),
+                            needs_storage_buffer_non_uniform,
+                        )
+                        .descriptor_binding_sampled_image_update_after_bind(
+                            private_caps.update_after_bind,
```
kvark, line 173: An ideal implementation would have something like a mask:
```rust
bitflags {
bitflags::bitflags!(
    pub struct UpdateAfterBindDescriptorTypes: u32 {
        const SAMPLED_IMAGE = 0x1;
        const STORAGE_IMAGE = 0x2;
        const UNIFORM_BUFFER = 0x4;
        const STORAGE_BUFFER = 0x8;
    }
);
```
We can afford to be more detailed here because it's not user-visible directly. As we discussed, this can be figure out by the backend based on the requested limits.

So if we aren't implementing it this way, let's write down a note somewhere with a TODO.

wgpu-hal/src/vulkan/adapter.rs
cwfitzgerald: This is actually the lowest non zero value for _all_ vulkan devices.

kvark: sure! Still, if we can grab an actual value from a physical device, no need to hard-code the constant in

wgpu-hal/src/vulkan/device.rs
cwfitzgerald: wym by this?

cwfitzgerald:
Alright, take another look at this, updated with comments and a new UAB strategy.

wgpu-hal/src/vulkan/device.rs
kvark: like, whether "partially bound" feature also means that wgpu-hal users specify a flag at BGL creation

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -751,8 +871,26 @@ impl super::Instance {
             },
         };
 
+        let update_after_bind = if phd_capabilities.properties.vendor_id == db::intel::VENDOR {
```
kvark, line 874: do we still need to limit this to Intel? If we are automatically enabling it only for limits exceeds, we should just do it for all vendors

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -751,8 +871,26 @@ impl super::Instance {
...
+            if let Some(ref di) = phd_features.descriptor_indexing {
+                di.descriptor_binding_sampled_image_update_after_bind != 0
```
kvark, line 876: why are we `&&` the conditions together?

wgpu-hal/src/vulkan/device.rs
kvark: Still an issue? We should check for `flags` here, not `self.uab_types`

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -498,8 +534,56 @@ impl PhysicalDeviceCapabilities {
         extensions
     }
 
-    fn to_wgpu_limits(&self) -> wgt::Limits {
+    fn to_wgpu_limits(&self, update_after_bind: bool) -> wgt::Limits {
         let limits = &self.properties.limits;
+        let max_sampled_textures = if update_after_bind {
```
kvark, line 539: I suppose we could check for `descriptor_binding_sampled_image_update_after_bind` specifically here instead of the catch-all boolean. Maybe we can gather a bitmask for the adapter, and that would be different bitmask (of UAB types) than we use for the device (since it depends on the requested limits).

cwfitzgerald:
Alright, third time's the charm, will test this in a minute.

wgpu-hal/src/lib.rs
```diff
@@ -799,6 +803,7 @@ pub struct SamplerDescriptor<'a> {
 pub struct BindGroupLayoutDescriptor<'a> {
     pub label: Label<'a>,
     pub entries: &'a [wgt::BindGroupLayoutEntry],
+    pub partially_bound: bool,
```
kvark, line 806: this should be in flags, like we have `InstanceFlags`, `PipelineLayoutFlags`, etc

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -7,15 +7,17 @@ use std::{ffi::CStr, mem, ptr, sync::Arc};
...
+        | wgt::Features::UNSIZED_BINDING_ARRAY
 }
 
 /// Aggregate of the `vk::PhysicalDevice*Features` structs used by `gfx`.
 #[derive(Debug, Default)]
 pub struct PhysicalDeviceFeatures {
     core: vk::PhysicalDeviceFeatures,
-    vulkan_1_2: Option<vk::PhysicalDeviceVulkan12Features>,
-    descriptor_indexing: Option<vk::PhysicalDeviceDescriptorIndexingFeaturesEXT>,
+    pub(crate) vulkan_1_2: Option<vk::PhysicalDeviceVulkan12Features>,
```
kvark, line 19: should be `pub(super)` instead

wgpu-hal/src/vulkan/device.rs
```diff
@@ -1012,9 +1017,73 @@ impl crate::Device<super::Api> for super::Device {
...
+                .entries
+                .iter()
+                .map(|entry| {
+                    let mut flags = vk::DescriptorBindingFlags::empty();
+
+                    if desc.partially_bound && entry.count.is_some() {
+                        flags |= vk::DescriptorBindingFlags::PARTIALLY_BOUND;
+                    }
+
+                    let binding_requres_update_after_bind = match entry.ty {
```
kvark, line 1037: let's return bitflags from here instead of a bool, we can check for `uab_types` containing the flags after the match

kvark (commented):
This looks great! Just a few nits left

cwfitzgerald:
Nightly fail is unrelated.

kvark (approved):

PR #2001, opened 2021-09-26
Move validation from `Device::create_texture_from_hal()` to `Device::create_texture()`

dwbrite:
**Connections**
It seems the mipmap validation was originally fixed in #893, and then moved to `fn create_texture_from_hal` in #1609?

**Description**
Users encountered a segfault upon creating a texture when mip_level_count was set to 0 with (at least) Vulkan.
This is due to wgpu's validation taking place after the API's texture is created.
This change moves wgpu's validation to before the API's texture is created, introducing a friendlier error instead of a segfault.

**Testing**
I ran the cube example with mip_level_count as 1, and again on 0. 
With mip_level_count set to 0, wgpu now produces the error: 

```

Caused by:
    In Device::create_texture
    texture descriptor mip level count (0) is invalid
    
'
```
    
----
notes: 
`format_features` is being created twice (probably not worth an api change), and validation will no longer happen in the unsafe hal code-paths.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -3122,12 +3122,14 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
 
             let format_features = match device
                 .describe_format_features(adapter, desc.format)
-                .map_err(|error| resource::CreateTextureError::MissingFeatures(desc.format, error)) {
+                .map_err(|error| resource::CreateTextureError::MissingFeatures(desc.format, error))
+            {
                 Ok(features) => features,
                 Err(error) => break error,
             };
 
-            let texture = device.create_texture_from_hal(hal_texture, device_id, desc, format_features);
+            let texture =
+                device.create_texture_from_hal(hal_texture, device_id, desc, format_features);
             let num_levels = texture.full_range.levels.end;
             let num_layers = texture.full_range.layers.end;
             let ref_count = texture.life_guard.add_ref();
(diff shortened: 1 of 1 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -556,6 +556,47 @@ impl<A: HalApi> Device<A> {
 
         let hal_usage = conv::map_texture_usage(desc.usage, desc.format.into());
 
+        let format_features = self
```
kvark, line 559: can this check be moved out as well? This way, we can:
  1. make the function return `resource::Texture<A>` unconditionally
  2. don't pass `adapter` in

kvark (commented):
Nice! Just one small thing

wgpu-core/src/device/mod.rs
dwbrite: yeah, we can move the check into `Global::create_texture_from_hal()`, and then replace `adapter` with `format_features` so it's not checked twice. see d38a400

If we're going to do that we could _also_ pass in `hal_usage` as a parameter instead of mapping from `desc.usage` twice.

---

But something tells me all this could be untangled more. I'd have to dig more into the types to see. 

Something that strikes me is that both `resource::Texture` and `hal::TextureDescriptor` need `hal_usage` (derived from `desc.usage` and `desc.format`). Especially when `resource::Texture` _already_ has access to `desc`.

kvark (approved):

PR #2002, opened 2021-09-27
Adds white background to main logo for better visibility on black

mattlacey:
**Description**
Logo had black text with a transparent background, when using GH in dark mode this wasn't easily visible, to the point that I totally missed the text was there when addressing a similar issue with other images previously:

image

**Testing**

View the Repo! 

image


grovesNL (approved):
Thank you!

PR #2006, opened 2021-09-28
Check Deno build on CI

kvark:
**Connections**
#1978 broke `deno_webgpu` without us noticing.

**Description**
We may not want to run everything on each change, but we can at least make sure Deno still builds.

**Testing**
Self-tested


kvark:
It takes 3 minutes. I think this is useful to have.

Issue #1506, opened 2021-06-18
Leftover mentions of gfx-hal


memoryruins:
Currently, there are a few leftover mentions of `gfx-hal` after the switch to `wgpu-hal` in https://github.com/gfx-rs/wgpu/pull/1471. Originally I was going to mention the "About" section in matrix, but then I noticed some other places so I'm opening this in case it's easier to track.

---

The "About" section in github's sidebar mentions gfx-hal. The use of "native" might be worth removing to avoid confusing with `wgpu-native` which is in another repo.

> Native WebGPU implementation based on gfx-hal 

![image](https://user-images.githubusercontent.com/6868531/122549962-11f8e100-d001-11eb-8624-40e0ad0d7711.png)

---

The "Development" section in https://github.com/gfx-rs/wgpu/blob/master/wgpu/README.md has instructions for patching gfx-hal. The entire section seems safe to remove.

> If you need to test local fixes to gfx or other dependencies, the simplest way is to add a Cargo patch. For example, when working on DX12 backend on Windows, you can check out the latest release branch in the gfx-hal repository (e.g. currently hal-0.8) and add this patch to the end of Cargo.toml

The description in wgpu-core refers to gfx-hal.

https://github.com/gfx-rs/wgpu/blob/0c76819ef76fd25a7faa619059290cd8b2687980/wgpu-core/Cargo.toml#L6

A doc comment in wgpu-core refers to a gfx-hal limitation.

https://github.com/gfx-rs/wgpu/blob/0c76819ef76fd25a7faa619059290cd8b2687980/wgpu-types/src/lib.rs#L205-L209

PR #2012, opened 2021-10-01
Update mentions of gfx-hal to wgpu-hal

memoryruins:
#1506 closed, but some of the mentions still existed. Updated "gfx-hal" to "wgpu-hal" where it made sense to me, leaving alone old changelog mentions etc.


cwfitzgerald (approved):
Nice! I thought they were fixed when I closed it.

PR #2014, opened 2021-10-03
Add spans to provide more information profile gaps

cwfitzgerald:
**Description**

Trying to fill in some information gaps in profiles.
- If there is a long running api call, mark it specifically to indicate what is the actual root performance loss.
- Add spans in places where there there was a decent amount of work happening, but there was no explanation as to why.

Did this for dx12 and vulkan as these are the platforms I have access to.

**Testing**

Inspected logs in tracy

### DIFF
--- a/wgpu-core/src/device/life.rs
+++ b/wgpu-core/src/device/life.rs
@@ -128,35 +128,65 @@ impl<A: hal::Api> NonReferencedResources<A> {
     }
 
     unsafe fn clean(&mut self, device: &A::Device) {
-        for raw in self.buffers.drain(..) {
-            device.destroy_buffer(raw);
+        if !self.buffers.is_empty() {
+            profiling::scope!("destroy_buffers");
+            for raw in self.buffers.drain(..) {
+                device.destroy_buffer(raw);
+            }
         }
-        for raw in self.textures.drain(..) {
-            device.destroy_texture(raw);
+        if !self.textures.is_empty() {
+            profiling::scope!("destroy_textures");
+            for raw in self.textures.drain(..) {
+                device.destroy_texture(raw);
+            }
         }
-        for (_, raw) in self.texture_views.drain(..) {
-            device.destroy_texture_view(raw);
+        if !self.texture_views.is_empty() {
+            profiling::scope!("destroy_texture_views");
+            for (_, raw) in self.texture_views.drain(..) {
+                device.destroy_texture_view(raw);
+            }
         }
-        for raw in self.samplers.drain(..) {
-            device.destroy_sampler(raw);
+        if !self.samplers.is_empty() {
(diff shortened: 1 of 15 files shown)


wgpu-core/src/command/render.rs
```diff
@@ -729,6 +729,7 @@ impl<'a, A: HalApi> RenderPassInfo<'a, A> {
     ) -> Result<StatefulTrackerSubset, RenderPassErrorInner> {
         profiling::scope!("finish", "RenderPassInfo");
         unsafe {
+            profiling::scope!("end_render_pass");
```
kvark, line 732: there is already a scope for this, why do we need the nested one?

wgpu-core/src/device/life.rs
```diff
@@ -128,35 +128,65 @@ impl<A: hal::Api> NonReferencedResources<A> {
     }
 
     unsafe fn clean(&mut self, device: &A::Device) {
-        for raw in self.buffers.drain(..) {
-            device.destroy_buffer(raw);
+        {
+            profiling::scope!("destroy_buffers");
```
kvark, line 132: let's guard this with `if !self.buffers.is_empty()` and everything below the same way

wgpu-hal/src/dx12/adapter.rs
```diff
@@ -44,21 +44,27 @@ impl super::Adapter {
         library: &Arc<native::D3D12Lib>,
         instance_flags: crate::InstanceFlags,
     ) -> Option<crate::ExposedAdapter<super::Api>> {
+        profiling::scope!("expose adapter");
```
kvark, line 47: this thing should live in `wgpu-core` if it doesn't already? No point in adding this scope to each backend

wgpu-hal/src/dx12/command.rs
```diff
@@ -767,9 +767,11 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
                 };
                 self.temp.barriers.push(barrier);
             }
+            profiling::scope!("ID3D12GraphicsCommandList::ResourceBarrier");
```
kvark, line 770: This scope would cover more things than just the next line. If you want this scope, let's do `if !self.temp.barriers.is_empty() { ... }`

kvark (commented):
This looks useful, thank you!
Just a few things that we may fix

wgpu-core/src/command/render.rs
cwfitzgerald: We don't, must have left that in while investigating.

wgpu-hal/src/dx12/adapter.rs
cwfitzgerald: Another investigation artifact.

kvark (approved):

Issue #1638, opened 2021-07-12
Repository Tags for Crates.io Releases


zicklag:
It would be convenient to have GitHub tags corresponding to the `wgpu` crates.io releases. I just ran into the need while wanting to checkout the git commit that corresponded to wgpu 0.9.0. Since `wgpu-rs` is not the only thing in this repo, we would probably want  to differentiate the tags with a prefix or something like that. So the tag could be `wgpu-rs-v0.9.0` or something.

cwfitzgerald:
We do have branches for each release, so you should be able to use those, though if tags would also be useful, that's a possibility too.

zicklag:
The branches are great, but they also receive updates after releases that will go into later bugfix releases ( if I understand correctly ) so it's harder to know that you've gotten a certain release exactly. Like with the 0.9 branch being 12 commits ahead of master makes me think that it's gotten updates that haven't been released yet that might go into an 0.9.1 release or something?

It's not really that big of a deal, but if it wasn't hard to do it's a nice-to-have.

cwfitzgerald:
Alright, I can go through and add tags for our releases.

Gordon-F:
Maybe it would be better to publish new versions by CI ?

https://github.com/rust-windowing/winit/blob/master/.github/workflows/publish.yml

cwfitzgerald:
Now that the family reunion has happened, fully automated publishing is definitely possible. I have a script that works really well for other projects, it would just need to be adapted here. I have held off on it as there were more important things to do and we don't actually release that often.

kvark:
I still don't understand what would tags give you over branches for breaking revisions. At what point do you need to know the exact changes between 0.9.1 and 0.9.2? You can look for them based on the commits in 0.9 branch that correspond to version bumps, for example.

zicklag:
Part of it was just intuition. I tend to look in the releases/tags list for the crates releases, and I didn't initially realize there was a branch there.

It's honestly probably not that big of a deal being that we already have branches.

My motivation is when I have the git source repo because I'm working on `master`, and I've got a project like bevy hooked up with a `patch` so that it uses my local `wgpu`, and then I want to quickly test out how what I'm working on compares to the previous released version that Bevy was already using from crates.io. But if I checkout the `v0.9` branch, then I'll potentially get unreleased commits, so it might not be exactly the same.

kvark:
[...]

PR #2017, opened 2021-10-04
Bring back the root lockfile

kvark:
**Connections**
Related to https://github.com/gfx-rs/wgpu/issues/1638#issuecomment-933370771

**Description**
Lockfile was removed when we added Deno, but afterwards we moved it out of workspace. So the root lockfile should still be fine to keep, especially since we have more binary targets: wgpu-info, in addition to the player and the examples.

**Testing**
No need


cwfitzgerald:
My only worry with this is about dependencies getting stale, but we likely can fix that by setting up dependabot.

cwfitzgerald (approved):

PR #2019, opened 2021-10-05
Disable partially bound arrays on intel/windows

cwfitzgerald:
**Connections**

Follow up to #1995.

**Description**

Intel/Windows get terminally mad with partially bound arrays. This force disables it.

**Testing**

Not needed.


kvark (approved):
We'll probably want something more scaleable here, like "Workarounds". But it's details...

PR #2020, opened 2021-10-05
Add component count data to TextureFormat::describe

cwfitzgerald:
**Connections**

None

**Description**

This is a useful bit of information to know about a texture format. In order to make decisions about what channels are valid, we need to know the component count, but we don't actually care about the format.

**Testing**

👀

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -1360,6 +1360,8 @@ pub struct TextureFormatInfo {
     pub block_dimensions: (u8, u8),
     /// Size in bytes of a "block" of texels. This is the size per pixel on uncompressed textures.
     pub block_size: u8,
+    /// Count of components in the texture. This determines which components there will be actual data in the shader for.
+    pub components: u8,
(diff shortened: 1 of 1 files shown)


kvark:
Can you tell more? What part of userspace actually need to know which channels are valid?
The shader can sampler or load any channels, regardless of the format.

cwfitzgerald:
"Valid" was a bad choice of words on my end. More "contains useful data". 

The use case that spawned this PR was the following two situations:
- When loading a ktx2 normal map, I get the format, then need to decide if to treat it as a bicomponent (r + g with derived b), or tricomponent (r + g + b). This comes from the amount of components that are valid inside the format.
- When loading a ktx2 metallic roughness map in the context of gtlf, I get the format, then need to decide if treat it as (metallic in g, roughness in b) or as (metallic in r, roughness in g). This also comes from the component count.

These are some very specific situations, but component count is a useful cross-format property to know. I could build this up myself, but this is low cost to implement inside wgpu, and gives out this information if other people need it.

kvark (approved):
Makes sense!

PR #2068, opened 2021-10-14
run-wasm-example.sh installs wasm-bindgen-cli if needed

rukai:
**Description**
Finding out how to install wasm-bindgen-cli is tricky!
The actual binary is called wasm_bindgen so you look up wasm_bindgen and end up with a library and get very confused.

**Testing**
I ran the script and it works! Both with and without wasm-bindgen already installed. If its already installed it doesnt try to reinstall it.

### DIFF
--- a/run-wasm-example.sh
+++ b/run-wasm-example.sh
@@ -2,6 +2,8 @@
 
 set -e
 
+cargo install wasm-bindgen-cli
+
 echo "Compiling..."
 cargo build --example $1 --target wasm32-unknown-unknown --features webgl
 


rukai:
Oh I see, it is in the wiki but only under the webgpu section.
I moved it up to a common section: https://github.com/gfx-rs/wgpu/wiki/Running-on-the-Web-with-WebGPU-and-WebGL
Please feel free to rearrange or reword it.

This seems to be the way to go instead of modifying the script.

Issue #1940, opened 2021-09-11
Can't run cube example on OpenGL ES 3.0


dlunch:
**Description**
Unable to run cube example with gl backend. vulkan backend works fine.
```
env WGPU_BACKEND=gl cargo run --example cube
...
[2021-09-11T06:19:41Z INFO  wgpu_hal::gles::adapter] SL version: OpenGL ES GLSL ES 3.00
[2021-09-11T06:19:41Z ERROR wgpu_hal::gles::egl] GLES: [API/Error] ID 1 : GL_INVALID_ENUM in glGetIntegerv(pname=GL_MAX_VERTEX_SHADER_STORAGE_BLOCKS)
```

**Repro steps**
Run cube example on [WSLg](https://github.com/microsoft/wslg/)

**Expected vs observed behavior**
cube example should run. 

**Extra materials**
WSLg supports GLES 3.0.
```
OpenGL ES profile version string: OpenGL ES 3.0 Mesa 21.2.1
OpenGL ES profile shading language version string: OpenGL ES GLSL ES 3.00
```
[`GL_MAX_VERTEX_SHADER_STORAGE_BLOCKS` is not supported on OpenGL ES 3.0](https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/glGet.xhtml), but is available on [OpenGL ES 3.1](https://www.khronos.org/registry/OpenGL-Refpages/es3.1/html/glGet.xhtml)
**Platform**
WSL on Windows 11, Arch Linux
current wgpu master (69e86453d91e2ee55781bc2e2b6cdb63656f67fd)

kvark:
@dlunch thank you for filing this!
The code should look like:
```rust
let supports_storage = ver >= (3, 1);
let vertex_shader_storage_blocks = if supports_storage {
            gl.get_parameter_i32(glow::MAX_VERTEX_SHADER_STORAGE_BLOCKS) as u32
} else { 0 };
```

The problem is that it's hard to experiment with this on platforms that support GLES-3.1
Would you be able to try writing this patch on your side? That would be most helpful

dlunch:
Thanks!

I tried to fix storage block `get_parameter_i32`  and `downlevel_limits`, but it fails later on bufferstorage code.

It seems https://github.com/gfx-rs/wgpu/pull/1686 fixes some problems here(webgl2 is based on gles3.0), so it will be better to work on this after webgl2 code merge.

PR #2078, opened 2021-10-15
GLES mapping fix

dlunch:
**Connections**
Fixes https://github.com/gfx-rs/wgpu/issues/1940

**Description**
Fix running on OpenGL ES 3.0 environment

ES 3.0 actually has glMapBufferRange, but wgpu buffer implementation requires MAP_PERSISTENT_BIT, which is 3.1 extension(https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_buffer_storage.txt), so EMULATE_BUFFER_MAP workaround is required.

### DIFF
--- a/wgpu-hal/src/gles/device.rs
+++ b/wgpu-hal/src/gles/device.rs
@@ -410,6 +410,10 @@ impl crate::Device<super::Api> for super::Device {
             .shared
             .workarounds
             .contains(super::Workarounds::EMULATE_BUFFER_MAP)
+            || !self
+                .shared
+                .private_caps
+                .contains(super::PrivateCapabilities::BUFFER_ALLOCATION)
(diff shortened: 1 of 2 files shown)


kvark:
I'm sorry, I jumped on #2083 without realizing there is a fix coming here!
Could you test and see if the problem is solved now?

dlunch:
`[2021-10-16T22:47:13Z ERROR wgpu_hal::gles::egl] GLES: [API/Error] ID 1 : GL_INVALID_OPERATION in glCopyBufferSubData(readBuffer is mapped)`

dlunch:
It seems `MAP_PERSISTENT_BIT` is always required for host visible buffer, as current wgpu buffer maps buffer at creation

Issue #1572, opened 2021-06-30
Support SamplerArray


igaryhe:
**Is your feature request related to a problem? Please describe.**
Currently wgpu supports `BufferArray` and `TextureViewArray`. It would be great if the API could also have `SamplerArray`.

**Describe the solution you'd like**
I'm not sure if it could work by simply adding `SamplerArray` at places where `BufferArray` appears.

**Describe alternatives you've considered**
I don't have an alternative solution yet.

**Additional context**

cwfitzgerald:
Done

PR #2113, opened 2021-10-24
Add new BindingResource type SamplerArray

igaryhe:
**Connections**
Linked issue: https://github.com/gfx-rs/wgpu/issues/1572

**Description**
`SamplerArray` is guarded with the same flag as `TextureViewArray`, since in both [Vulkan](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPhysicalDeviceFeatures.html) and [Metal](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) they share the identical feature set.

**Testing**
`texture-arrays` example in `wgpu` is modified and tested.


igaryhe:
hi, just would like to know if @cwfitzgerald are still interested in giving some comments on this PR

cwfitzgerald:
Ack, I'm so sorry for abandoning this review, will take a look once I get to my computer. Thanks for the poke!

kvark:
@cwfitzgerald please have another look!

cwfitzgerald (changes requested):
So sorry for the delay

This looks fine to me code wise on the wgpu side, do note that nothing is actually turning this feature on in one of the HAL backends, so you wouldn't be able to actually use it.

I think it's okay to roll into the textures for now, and investigate how it plays with DX12 as we get DI up and running on 12.

igaryhe:
Hi, I didn't find any related code in d3d12 backend, and the feature `TEXTURE_BINDING_ARRAY` is shown supported in the doc, so I was assuming it's it's enabled by default. I'm not quite sure what kind of changes I need to make at the moment, and it would be a huge help if you could specify it.

cwfitzgerald (approved):
Sorry been totally off my game with this issue. You're absolutely right. If you can resolve the conflict we can get this in.

PR #2131, opened 2021-10-30
[gl] Clear existing vertex attributes on gles set_render_pipeline

dlunch:
This PR fixes incorrect unset of vertex_attributes state on gles set_render_pipeline if `PrivateCapabilities::VERTEX_BUFFER_LAYOUT` is not supported. 

The problem can be observed by calling set_pipeline many times in single render pass on gles 3.0 context.


kvark (approved):
This looks reasonable, thank you!

kvark:
Published in wgpu-hal-0.11.5

Issue #2130, opened 2021-10-30
Unable to handle shader compilation errors


Herschel:
`wgpu` currently has no API to inspect errors when compiling a shader module. On desktop, `Device::create_shader_module` will panic on error. On web, the different browsers will print out error info to the console, but execution continues, and errors will usually occur later during pipeline creation with the errored `ShaderModule` objects.

The [WebGPU spec](https://www.w3.org/TR/webgpu/#dom-gpushadermodule-compilationinfo) requires a `GPUShaderModule.compilationInfo()` method that should return any errors generated during compilation.

Ideally, either:
 * change `Device::create_shader_module` to idiomatically return a `Result`,
 * or match the spec and add a `ShaderModule::compilation_info` method. The user can then inspect this info and return an error as appropriate.

kvark:
`create_shader_module` follows the same logic as all the other methods that could fail - it triggers the error on the device. The default handler panics, but you can set your own handler - https://github.com/gfx-rs/wgpu/blob/5f1b3e56bc00eb3f805ffa8ae8ecde4a1bf09949/wgpu/src/lib.rs#L314

`compilation_info` isn't implemented yet, and we definitely need this.

As for `Result`, we need to land @pythonesque work first and then see.

Herschel:
Thanks for the tip about `onUncapturedError`! Unfortunately looks like it's a [TODO on wgpu web backend](https://github.com/gfx-rs/wgpu/blob/master/wgpu/src/backend/web.rs#L1624-L1630).

michael-brennan2005:
Sorry if this isn't the right etiquette, but is there any current work being done to implement `compilation_info`? This is presenting an obstacle to the renderer I'm working on currently.

Wumpf:
not that I know of no. But there's also some movement on the Naga side on how errors should be handled https://github.com/gfx-rs/wgpu/issues/4429

stefnotch:
I would love to take on this issue.

stefnotch:
The PR for this is up at https://github.com/gfx-rs/wgpu/issues/5410, and am happily awaiting reviews

nnmm:
What I'm using, based on the comments in this thread, is

```rust
    // Set up error handling. By using `on_uncaptured_error()`, create_shader_module()
    // will return instead of exiting.
    let shader_err: Arc<Mutex<Option<wgpu::Error>>> = Arc::new(Mutex::new(None));
    {
[...]

PR #2132, opened 2021-10-30
web: Implement Device::on_uncaptured_error

Herschel:
**Connections**
#2130

**Description**
This PR is a first pass at wiring up `Device::on_uncaptured_error` on web, mapping the WebGPU errors into `wgpu::Error`.

Possible improvements:
 * JS does not provide a standardized stack trace on error types, so the `Error::source` value is bogus.
 * Currently the error handler will not be called if an unknown error type is encountered (say, an error is added to the spec).
 * Due to using `wasm_bindgen::Closure::forget`, this will leak the memory associated with the closure. This seemed okay for a one-off diagnostic function, and also can be avoided by enabling [weak references](https://rustwasm.github.io/wasm-bindgen/reference/weak-references.html), available on all major browsers now. Alternatively, a map would need to be maintained between devices and their error handler.

**Testing**
Tested it in [Ruffle](https://ruffle.rs/demo/webgpu.html) to catch errors thrown by Chrome on certain platforms.

### DIFF
--- a/wgpu/src/backend/web.rs
+++ b/wgpu/src/backend/web.rs
@@ -7,7 +7,7 @@ use std::{
     pin::Pin,
     task::{self, Poll},
 };
-use wasm_bindgen::prelude::*;
+use wasm_bindgen::{prelude::*, JsCast};
 
 // We need to make a wrapper for some of the handle types returned by the web backend to make them
 // implement `Send` and `Sync` to match native.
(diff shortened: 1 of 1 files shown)


grovesNL (approved):
Thank you!

Ideally we could have a catch-all and avoid leaking this (using weak references seems reasonable), but we can do that in follow-up PRs

Issue #1808, opened 2021-08-16
Check max_workgroup_size limit


kvark:
See https://github.com/gfx-rs/naga/issues/1205#issuecomment-899743787

teoxoy:
I think we can close this since it has been fixed by https://github.com/gfx-rs/wgpu/pull/2142.

PR #2133, opened 2021-10-30
Compute max workgroup size limits

UpsettingBoy:
**Connections**
#1808

**Description**
Added Limits for compute operations (workgroup sizes) and a check during a _normal_ dispatch operation.

I cannot currently add these new Limits to Metal and GLES because no autocompletion under Windows for them (OpenGL drivers should be installed but there is not even a compatible device ¯\_(ツ)_/¯ ).

Also, I think, since now we have max_workgroup_size_(x,y,z) a validation on compute shaders about their workgroup_size should be possible.

### DIFF
--- a/wgpu-core/src/command/compute.rs
+++ b/wgpu-core/src/command/compute.rs
@@ -121,6 +121,8 @@ pub enum DispatchError {
         //expected: BindGroupLayoutId,
         //provided: Option<(BindGroupLayoutId, BindGroupId)>,
     },
+    #[error("each current dispatch group size dimmension ({current:?}) must be less or equal to {limit}")]
+    InvalidGroupSize { current: [u32; 3], limit: u32 },
 }
 
 /// Error encountered when performing a compute pass.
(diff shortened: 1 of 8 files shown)


kvark (commented):
Nice work!

kvark:
I pushed GLES implementation into your branch, please pick it up locally for further changes.

UpsettingBoy:
Thanks!! I pick your commit locally. Only Metal reamining then.

Also, I have a question . As I understood the WebGPU specification, the [dispatch ](https://gpuweb.github.io/gpuweb/#dom-gpucomputepassencoder-dispatch)operation can only be executed if its dimensions are <= than [maxComputeWorkgroupsPerDimension ](https://gpuweb.github.io/gpuweb/#dom-supported-limits-maxcomputeworkgroupsperdimension); but when testing, I can execute a dispatch with values greater than the limit without any problem. Am I missing something?

kvark:
Also pushed Metal now. Wasn't able to find the relevant limit in the tables, so just put defaults.

kvark:
@UpsettingBoy what are you testing exactly?
Your PR is adding this validation, so obviously nothing will error on wgpu side until your PR is merged.

UpsettingBoy:
I was testing if a program would crash if the dispatch dims are greater than the limits reported, but it didn't.

kvark:
Doing this is invalid usage. This is literally off-limits territory, it's undefined behavior - anything can happen, including not crashing.

UpsettingBoy:
Alright, thanks for the clarification!! I think this PR is ready then.

kvark (approved):

PR #2142, opened 2021-11-01
Compute shader workgroup size validation

UpsettingBoy:
**Connections**
#1808 

**Description**
Validate the a compute shader's current workgroup size with the device limits.

I created a different PR for the same issue because I think this validation is conceptually different to the previous PR on the issue. If you prefer everything related to it in the same PR, let me know so I can do it from now on.

### DIFF
--- a/wgpu-core/src/validation.rs
+++ b/wgpu-core/src/validation.rs
@@ -106,11 +106,13 @@ struct EntryPoint {
     #[allow(unused)]
     spec_constants: Vec<SpecializationConstant>,
     sampling_pairs: FastHashSet<(naga::Handle<Resource>, naga::Handle<Resource>)>,
+    workgroup_size: [u32; 3],
 }
 
 #[derive(Debug)]
 pub struct Interface {
     features: wgt::Features,
+    limits: wgt::Limits,
     resources: naga::Arena<Resource>,
     entry_points: FastHashMap<(naga::ShaderStage, String), EntryPoint>,
 }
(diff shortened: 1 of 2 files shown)


wgpu-core/src/validation.rs
```diff
@@ -1047,6 +1057,22 @@ impl Interface {
...
 
+        // check workgroup size limits
+        if shader_stage == naga::ShaderStage::Compute {
+            let max_workgroup_size_limits = [
+                self.limits.max_compute_workgroup_size_x,
+                self.limits.max_compute_workgroup_size_y,
+                self.limits.max_compute_workgroup_size_z,
+            ];
+
+            if entry_point.workgroup_size > max_workgroup_size_limits {
```
kvark, line 1068: I think this is a wrong way to check. You'd get `[1, 10000, 1] > [16,16,16]` failing.

kvark (changes requested):
Thank you for taking a stab at this!

wgpu-core/src/validation.rs
UpsettingBoy: I changed to the _standard_ way of checking. I don't know why I thougth it could be that way lol

UpsettingBoy:
It may be possible to close #1808 now I think

wgpu-core/src/validation.rs
```diff
@@ -1047,6 +1057,25 @@ impl Interface {
             }
         }
 
+        // check workgroup size limits
+        if shader_stage == naga::ShaderStage::Compute {
+            let max_workgroup_size_limits = [
```
kvark, line 1062: I don't think this value gives us much now, let's use the limits directly

UpsettingBoy: Would't it make sense to keep it since that array has to be created for the error msg anyways?

kvark: oh yes, I missed this, thank you!

kvark (approved):

kvark:
Please rebase

kvark (approved):

PR #2139, opened 2021-10-31
gles: Fix rendering to cubemap faces

magcius:
**Connections**
_Link to the issues addressed by this PR, or dependent PRs in other repositories_

One part of #2138

**Description**
_Describe what problem this is solving, and how it's solved._

This is a cleaned up version of the scratch patch I gave on Matrix to fix cubemap renders.

**Testing**
_Explain how this change is tested._

I didn't test it yet.

Issue #2138, opened 2021-10-31
Shadow map comparison sampler always produces shadow on WebGL2


superdump:
**Description**

In bevy we are reworking the renderer and have a fairly simple test scene we use for demonstrating basic 3D rendering with 3 boxes for walls, 3 spheres representing the positions of 3 coloured point lights (red, green, blue), and a cube and another sphere as just objects in the scene to be lit and cast shadows. There is also a directional light which is pointing toward the ground plane at a 45 degree angle, but rotating about the y axis over time.

**Repro steps**

See the attached zip file. It contains the wasm, js, html necessary to run the bevy wgpu WebGL2 version of this example application. Extract it, serve it using some local http server (I've been using the rust simple-http-server crate with --nocache to make sure it's loading fresh across rebuilds) and open it in your browser of choice that supports WebGL2.

**Expected vs observed behavior**

Expected result from a native build of the exact same bevy wgpu code:
<img width="641" alt="Screenshot 2021-10-31 163459" src="https://user-images.githubusercontent.com/302146/139591605-5ccb7748-a09f-45f5-84c1-d1d0d9d13cf3.png">

Result from Firefox:
<img width="644" alt="Screenshot 2021-10-31 163532" src="https://user-images.githubusercontent.com/302146/139591616-16aff1d1-0543-4ff4-bad7-8bfcb29dffc9.png">
Note that there is no red light. The red light is the first of the three point lights and it should cast shadows. WebGL2 does not support cube array textures, so I have had to drop the code down to support a single point light shadow map. Also note that the white directional light is not casting any shadows.

Result from Chrome:
<img width="640" alt="Screenshot 2021-10-31 165138" src="https://user-images.githubusercontent.com/302146/139591707-fcad19fc-1dcb-472f-895d-72251898b309.png">
Note that the lighting looks very similar to the native version, but there are no shadows from the red light (which is presumably just never occluded in this case, compared to always occluded in the Firefox case) nor are there shadows from the white directional light.

**Extra materials**
Zip file of the web build of the code: [web.zip](https://github.com/gfx-rs/wgpu/files/7449188/web.zip) Unpack it, serve it, open it.

**Platform**
OS: Windows 10
wgpu 0.11.0, naga 0.7.1, bevy running something very close to the pipelined-rendering branch which represents the current state of the renderer rework. Using the Vulkan backend for native. Mobile RTX 3080.

kvark:
[...]

PR #2143, opened 2021-11-01
queue: Another attempt at cube map rendering

magcius:
**Connections**
Issue #2139

**Description**

The last OR was wrong, because I misread a (confusing) line in the GL ES spec.

**Testing**

It wasn't, this is a blind PR. Please don't merge until @superdump has been given an opportunity to test.

### DIFF
--- a/wgpu-hal/src/gles/queue.rs
+++ b/wgpu-hal/src/gles/queue.rs
@@ -22,10 +22,7 @@ fn extract_marker<'a>(data: &'a [u8], range: &std::ops::Range<u32>) -> &'a str {
 
 fn is_layered_target(target: super::BindTarget) -> bool {
     match target {
-        glow::TEXTURE_2D_ARRAY
-        | glow::TEXTURE_3D
-        | glow::TEXTURE_CUBE_MAP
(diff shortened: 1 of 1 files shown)


superdump:
Confirmed working for the point light cube map shadow map texture as in #2138 in both Chrome and Firefox. Thanks!

wgpu-hal/src/gles/queue.rs
```diff
@@ -94,6 +91,14 @@ impl super::Queue {
                         view.mip_levels.start as i32,
                         view.array_layers.start as i32,
                     );
+                } else if target == glow::TEXTURE_CUBE_MAP {
+                    gl.framebuffer_texture_2d(
```
kvark, line 95: nit: could be merged more closely with the other `framebuffer_texture_2d`

magcius: i figured it was clearer this way than doing `if target == glow::TEXTURE_CUBE_MAP { texture = CUBE_MAP_FACES[...]; }` but feel free to make your own judgment

kvark:
Published in wgpu-hal-0.11.5

PR #2152, opened 2021-11-03
Changed repository from tree v0.10 to master

CatCode79:
crates.io wrongly points to version 10 because the repository field value, I was tempted to change the repository to version 11 but ultimately opted for the master, I think it makes more sense


kvark:
That was kinda intentional. Since we use `master` for all development and land breaking changes, a lot of people are confused as to why the `master` examples don't build with latest released `wgpu`. So the solution I went with was to specify this exact release branch in the "repository" link of cargo manifest, so that people jump to the proper example code.
The downside, obviously, is that this link needs to be maintained. I think we'll need to make a TODO list of things for the release and just write it down as one of the actions, instead of changing it back to master.

CatCode79:
The best would be to have an automated system or some kind of script.
However this PR no longer makes sense, I close it.

Issue #1941, opened 2021-09-11
dvec3 division broken


timokoesters:
**Description**
Dividing a dvec3 by a number seems to make components NaN:

```
#version 450

layout(location = 0) in vec3 fragColor;
layout(location = 0) out vec4 outColor;

void main() {
    dvec3 diff = dvec3(0, 0, 1);
    if (isnan((diff / 1.0).x)) { outColor = vec4(1.0, 0.0, 0.0, 1.0); }
    else { outColor = vec4(0.0, 1.0, 0.0, 1.0); }
}
```

This will mark all points red. If you try the same with vec3 or inline it like this: `isnan((dvec3(0, 0, 1) / 1.0).x)` it shows them as green.

**Extra materials**
I compiled the shader using `glsl_to_spirv::compile(fs, glsl_to_spirv::ShaderType::Fragment).unwrap();` and am using these features: SHADER_FLOAT64 | BUFFER_BINDING_ARRAY | STORAGE_RESOURCE_BINDING_ARRAY | VERTEX_WRITABLE_STORAGE,

**Platform**
Information about your OS, version of `wgpu`, your tech stack, etc.

AdapterInfo { name: "Intel(R) HD Graphics 620 (KBL GT2)", vendor: 32902, device: 22806, device_type: IntegratedGpu, backend: Vulkan }
wgpu 0.10.1
naga 0.6.3

danwilhelm:
The [documentation](https://github.com/gfx-rs/wgpu/blob/master/wgpu-types/src/lib.rs#L2314) suggests it might also require the feature `VERTEX_ATTRIBUTE_64BIT`?

cwfitzgerald:
@danwilhelm That's only needed if you have `dvec*` in the `in/out` to a shader. This example uses normal floats.

timokoesters:
You should be able to test this on https://github.com/timokoesters/nbodysim/tree/wgpu-bug (the wgpu-bug branch)

kvark:
@timokoesters could you run the shader through Naga cli to see what the output looks like? Using `spirv-dis` on the output should show us interesting things.

timokoesters:
If you are still interested in this I can run `spirv-dis` on the output tomorrow.

kvark:
Tried to run it on TigerLake GT2 and got an initialization error:
```
Using default config.
MESA-INTEL: warning: Performance support disabled, consider sysctl dev.i915.perf_stream_paranoid=0

[...]

PR #2159, opened 2021-11-06
Validate device descriptor before opening

kvark:
**Connections**
Solves https://github.com/gfx-rs/wgpu/issues/1941#issuecomment-960413188

**Description**
We were opening the device before validating.

**Testing**


kvark:
Published in wgpu-core-0.11.3

PR #2166, opened 2021-11-08
do not query work group params on gles < 3.1

mrk-its:
**Description**

gles backend queries `MAX_COMPUTE_WORK_GROUP_COUNT` and `MAX_COMPUTE_WORK_GROUP_SIZE` parameters generating `getIndexedParameter: Bad `pname`: Invalid enum value` warnings. This PR disables these queries on wasm32 and sets `max_compute_workgroup_size_*` and `max_compute_workgroups_per_dimension` to 
0

**Testing**
It can be tested on wgpu `cube` example

### DIFF
--- a/wgpu-hal/src/gles/adapter.rs
+++ b/wgpu-hal/src/gles/adapter.rs
@@ -191,6 +191,8 @@ impl super::Adapter {
         let ver = Self::parse_version(&version).ok()?;
 
         let supports_storage = ver >= (3, 1);
+        let supports_work_group_params = ver >= (3, 1);
+
         let shading_language_version = {
             let sl_version = gl.get_parameter_string(glow::SHADING_LANGUAGE_VERSION);
             log::info!("SL version: {}", &sl_version);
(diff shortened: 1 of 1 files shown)


wgpu-hal/src/gles/adapter.rs
```diff
@@ -328,11 +328,14 @@ impl super::Adapter {
             gl.get_parameter_i32(glow::MAX_VERTEX_UNIFORM_BLOCKS)
                 .min(gl.get_parameter_i32(glow::MAX_FRAGMENT_UNIFORM_BLOCKS)) as u32;
 
-        let max_compute_workgroups_per_dimension = gl
-            .get_parameter_indexed_i32(glow::MAX_COMPUTE_WORK_GROUP_COUNT, 0)
-            .min(gl.get_parameter_indexed_i32(glow::MAX_COMPUTE_WORK_GROUP_COUNT, 1))
-            .min(gl.get_parameter_indexed_i32(glow::MAX_COMPUTE_WORK_GROUP_COUNT, 2))
-            as u32;
+        let max_compute_workgroups_per_dimension = if cfg!(not(target_arch = "wasm32")) {
```
kvark, line 331: I think this check should be for GLES version not WASM32.
Something like this:
```rust
let supports_work_group_params = ver >= (3, 1);
let max_compute_workgroups_per_dimension = if supports_work_group_params {...
...
} else {...};
```

kvark (changes requested):

wgpu-hal/src/gles/adapter.rs
mrk-its: done!

kvark (approved):
Thank you!

kvark:
Published in wgpu-hal-0.11.5

Issue #2170, opened 2021-11-10
Further spec compliance


crowlKats:
- [x] introduce the `DEPTH_CLIP_CONTROL` feature (edit: #2171)
- [x] introduce the `INDIRECT_FIRST_INSTANCE` feature (edit: #2206)
- [x] implement `GPUCommandEncoder::fillBuffer`
- [x] `GPURenderPassDescriptor::colorAttachments` is now a list of optional elements (edit: #2778)
- [x] `GPURenderPassLayout::colorFormats` is now a list of optional elements (edit: #2778)
- [x] `GPUFragmentState::targets` is now a list of optional elements (edit: #2778)
- [x] `GPUPipelineDescriptorBase::layout` now accepts an enum as well as the `GPUPipelineLayout` (#2895)
- [x] remove `GPURenderPassEncoder::writeTimestamp` and `GPUComputePassEncoder::writeTimestamp` in favour of Descriptor properties
- [x] `GPURenderPassEncoder::beginOcclusionQuery` missing
- [x] `GPURenderPassEncoder::endOcclusionQuery` missing
- [x] `GPURenderPassDescriptor::occlusionQuerySet` missing
- [x] `GPUTextureDescriptor::viewFormats` missing (#3237)
- [ ] `GPURenderPassDescriptor::maxDrawCount` missing: https://github.com/gfx-rs/wgpu/issues/8737
- [x] `GPURenderBundleEncoder::drawIndexedIndirect` missing
- [ ] #8831
- [x] `GPUShaderModule::getCompilationInfo` missing
- [x] `GPURenderPassColorAttachment::depthSlice` missing: https://github.com/gfx-rs/wgpu/issues/4764
- [ ] `AdapterInfo::architecture` missing: https://github.com/gfx-rs/wgpu/issues/8649
- [ ] `GPUDeviceDescriptor::defaultQueue` missing: https://github.com/gfx-rs/wgpu/issues/4014
- [x] `GPUMipmapFilterMode` missing
- [ ] `GPUDepthStencilState` `depthWriteEnabled` and `depthCompare` should be optional: https://github.com/gfx-rs/wgpu/issues/8830
- [ ] `setBindGroup`'s `bindGroup` argument should be optional: https://github.com/gfx-rs/wgpu/issues/4738
- [ ] `setVertexBuffer`'s `buffer` argument should be optional: https://github.com/gfx-rs/wgpu/issues/7354
[...]

PR #2171, opened 2021-11-10
Evolve depth clamping into clip control

kvark:
**Connections**
Upstream https://github.com/gpuweb/gpuweb/issues/2100
Reported in #2170

**Description**
The new semantics is subtly different. We can't portably enable/disable *all* clamping. The final clamp needs to happen unconditionally. But we can disable the early stage clipping. So now this requires Vulkan extension for depth clip control.

**Testing**
Running shadow.

### DIFF
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -22,6 +22,7 @@ pub struct PhysicalDeviceFeatures {
     timeline_semaphore: Option<vk::PhysicalDeviceTimelineSemaphoreFeaturesKHR>,
     image_robustness: Option<vk::PhysicalDeviceImageRobustnessFeaturesEXT>,
     robustness2: Option<vk::PhysicalDeviceRobustness2FeaturesEXT>,
+    depth_clip_enable: Option<vk::PhysicalDeviceDepthClipEnableFeaturesEXT>,
 }
 
 // This is safe because the structs have `p_next: *mut c_void`, which we null out/never read.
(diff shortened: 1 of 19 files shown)


wgpu-types/src/lib.rs
```diff
@@ -163,18 +163,18 @@ bitflags::bitflags! {
     #[repr(transparent)]
     #[derive(Default)]
     pub struct Features: u64 {
-        /// By default, polygon depth is clipped to 0-1 range. Anything outside of that range
-        /// is rejected, and respective fragments are not touched.
+        /// By default, polygon depth is clipped to 0-1 range before/during rasterizatiokn.
```
msiglreith, line 166:
```suggestion
        /// By default, polygon depth is clipped to 0-1 range before/during rasterization.
```

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -281,6 +284,17 @@ impl PhysicalDeviceFeatures {
             } else {
                 None
             },
+            depth_clip_enable: if enabled_extensions.contains(&vk::ExtRobustness2Fn::name()) {
```
msiglreith, line 287: why does it depend on the `VK_EXT_robustness2` extension?

kvark: copy-paste error :sweat:

wgpu-hal/src/vulkan/device.rs
```diff
@@ -1396,6 +1395,13 @@ impl crate::Device<super::Api> for super::Device {
         if desc.primitive.conservative {
             vk_rasterization = vk_rasterization.push_next(&mut vk_rasterization_conservative_state);
         }
+        let mut vk_depth_clip_state =
+            vk::PipelineRasterizationDepthClipStateCreateInfoEXT::builder()
+                .depth_clip_enable(true)
```
msiglreith, line 1400: shouldnt this be `false` here?

kvark: oh, that's a great find!

kvark:
Excellent review of Vulkan code, thanks @msiglreith !

PR #2174, opened 2021-11-12
Renaming CommandEncoder::clearBuffer to CommandEncoder::fillBuffer

UpsettingBoy:
**Connections**
#2170 

**Description**
~~This PR adds `fill_buffer` to CommandEncoder for Vulkan and D3D12. Metal and GLES yet to be done (I cannot implement them 😢).~~

~~There are probably some changes to be made, mainly in the d3d12 implementation. e.g There are no deallocations over the CPU only descriptor heap. Also, I think the changes I made for d3d12 FixedSizeHeap might be problematic on multi-threading 
scenarios.~~

EDIT: For now is just a rename to keep up with the spec.

**Testing**
I only tested this PR with a small change to the `hello-compute` example. No other testing is done :(

### DIFF
--- a/wgpu-core/src/command/clear.rs
+++ b/wgpu-core/src/command/clear.rs
@@ -87,7 +87,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
 
         #[cfg(feature = "trace")]
         if let Some(ref mut list) = cmd_buf.commands {
-            list.push(TraceCommand::ClearBuffer { dst, offset, size });
+            list.push(TraceCommand::FillBuffer { dst, offset, size });
         }
 
         let (dst_buffer, dst_pending) = cmd_buf
--- a/wgpu-core/src/command/clear.rs
+++ b/wgpu-core/src/command/clear.rs
@@ -69,14 +69,14 @@ whereas subesource range specified start {subresource_base_array_layer} and coun
 }
 
 impl<G: GlobalIdentityHandlerFactory> Global<G> {
-    pub fn command_encoder_clear_buffer<A: HalApi>(
+    pub fn command_encoder_fill_buffer<A: HalApi>(
         &self,
         command_encoder_id: CommandEncoderId,
         dst: BufferId,
         offset: BufferAddress,
         size: Option<BufferSize>,
     ) -> Result<(), ClearError> {
-        profiling::scope!("CommandEncoder::clear_buffer");
+        profiling::scope!("CommandEncoder::fill_buffer");
 
         let hub = A::hub(self);
         let mut token = Token::root();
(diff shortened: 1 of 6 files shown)


kvark:
Thank you for taking a stab at this!
We actually don't need this new `fill_buffer` method in wgpu-hal, there is already `clear_buffer` for exactly the same purpose. The API difference between them is the `value: u8`, but if you look at WebGPU spec, there isn't a value there either (because it would be more difficult to implement) - https://gpuweb.github.io/gpuweb/

So all that needs to be done here is exposing the method to wgpu-rs

UpsettingBoy:
Ah, alright, I misread part of the discussions over gpuweb haha.
So this would be more of a renaming from clear_buffer to fill_buffer, wouldn't it?

kvark:
We had it called `fill_buffer` at some point, then renamed to `clear_buffer`. I guess it's time to switch it back now :)
cc @Wumpf who did most of this work.

kvark:
Another thing is - this API should now be available without any features, previously it was behind a feature.

Wumpf:
oh cool didn't know that this is now in the spec. Nice!
Wanted to change the meaning of the extension a little bit towards "this is what zero init does when it hits", but that would change that again then (:
https://github.com/Wumpf/wgpu/commit/127cbab6bdbb53e966c312adfe6667567e040658

UpsettingBoy:
Alright, but does clear_texture remain under CLEAR_COMMANDS?

Also, by gpuweb/gpuweb#2208 comment, `value` may be added in the future (is on the workings I think: gpuweb/gpuweb#2290) so, do you think is better to stall this PR until then and create another one to keep up with the spec in the meantime?

Wumpf:
> Alright, but does clear_texture remain under CLEAR_COMMANDS?

yes, and we should rename it to CLEAR_TEXTURES, but I can do that later; going to make some changes to it anyways.
I'd say let's add `value` when it enters the spec. With zero we can implement it on all backends today, but adding a value means that we need to solve this with shaders on some backends

UpsettingBoy:
Okay, so I removed all the previous commits from this branch. They will still remain on my branch (https://github.com/UpsettingBoy/wgpu/tree/cmd-encoder-value) so when the spec adds `value` there'll be some work done :)

@Wumpf I left CLEAR_COMMAND untouched for you to change.

Wumpf (approved):

wgpu-core/src/command/clear.rs
```diff
@@ -87,11 +87,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
 
         #[cfg(feature = "trace")]
         if let Some(ref mut list) = cmd_buf.commands {
-            list.push(TraceCommand::ClearBuffer { dst, offset, size });
-        }
-
-        if !cmd_buf.support_clear_buffer_texture {
```
kvark, line 93: this boolean should also be renamed now

kvark (approved):

PR #2180, opened 2021-11-15
Add Device::as_hal and always assume texture from hal is initialized

xiaopengli89:
**Description**

`Device::create_texture_from_hal` method will assume that the texture is not initialized and filled with 0, so I add a `Device::create_texture_from_hal_initialized` method to help users import an initialized texture from outside, which is very useful in building applications such as video players.

**Testing**
1. Get a metal texture from CoreVideo
2. Create a wgpu texture with `create_texture_from_hal_initialized` from the metal texture
3. Render the texture

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -3199,13 +3199,13 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
     ///
     /// - `hal_texture` must be created from `device_id` corresponding raw handle.
     /// - `hal_texture` must be created respecting `desc`
+    /// - `hal_texture` must be initialized
     pub unsafe fn create_texture_from_hal<A: HalApi>(
         &self,
         hal_texture: A::Texture,
         device_id: id::DeviceId,
         desc: &resource::TextureDescriptor,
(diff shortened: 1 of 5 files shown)


kvark:
Is there ever a use case where a texture is imported and `wgpu` initializes it? I find this unexpected. The lazy initialization only makes sense when we fully control how something is used. I believe external resources like this should never be attempted at filling with zeroes.

xiaopengli89:
When I was developing a video playback app, I wanted to avoid copying the decoded frames to the memory and then writing them to the wgpu, and instead imported them directly into the wgpu through the metal texture. Initially I used `Device::create_texture_from_hal`, but I couldn't render the correct video frame. It seemed that the texture data was filled with 0. I found that it was caused by the `initialization_status` flag of `wgpu_core::resource::Texture`. To avoid breaking changes, I added the `Device::create_texture_from_hal_initialized` method.
You mean `Device::create_texture_from_hal` should always assume that the texture has been initialized?

kvark:
> You mean Device::create_texture_from_hal should always assume that the texture has been initialized?

Yes. I don't think we need the "initialization" be even configurable for external textures like this one.

kvark (approved):

PR #2181, opened 2021-11-15
Introduce SamplerBindingType enum and fix bug in validation

Bobo1239:
**Description**
Was investigating why the [shadow](https://wgpu.rs/examples/?example=shadow) example didn't work in Firefox Nightly and dug around until finding the validation bug described in the first commit. Thought it made more sense to switch to an enum so there's no mapping necessary between the `(bool, bool)` tuple (filtering  & comparison) and the `SamplerBindingType` from the spec. Though it's possible that there was a good reason to use the previous representation in which case I can submit a simpler patch which just fixes the validation.

**Testing**
Tested the examples natively (Vulkan; Linux; AMD). Also tested a custom build of firefox with an updated wgpu and [this patch](https://github.com/gfx-rs/wgpu/files/7541248/patch.txt) to confirm that the shadow example now works properly.

### DIFF
--- a/deno_webgpu/src/binding.rs
+++ b/deno_webgpu/src/binding.rs
@@ -56,34 +56,7 @@ impl From<GpuBufferBindingType> for wgpu_types::BufferBindingType {
 #[derive(Deserialize)]
 #[serde(rename_all = "camelCase")]
 struct GpuSamplerBindingLayout {
-    r#type: GpuSamplerBindingType,
-}
-
-#[derive(Deserialize)]
-#[serde(rename_all = "kebab-case")]
-enum GpuSamplerBindingType {
-    Filtering,
-    NonFiltering,
-    Comparison,
-}
-
-impl From<GpuSamplerBindingType> for wgpu_types::BindingType {
-    fn from(binding_type: GpuSamplerBindingType) -> Self {
-        match binding_type {
-            GpuSamplerBindingType::Filtering => wgpu_types::BindingType::Sampler {
-                filtering: true,
-                comparison: false,
-            },
-            GpuSamplerBindingType::NonFiltering => wgpu_types::BindingType::Sampler {
-                filtering: false,
-                comparison: false,
-            },
-            GpuSamplerBindingType::Comparison => wgpu_types::BindingType::Sampler {
-                filtering: true,
-                comparison: true,
-            },
-        }
-    }
+    r#type: wgpu_types::SamplerBindingType,
 }
 
 #[derive(Deserialize)]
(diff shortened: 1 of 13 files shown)


deno_webgpu/src/binding.rs
```diff
@@ -70,18 +70,15 @@ enum GpuSamplerBindingType {
...
-            },
-            GpuSamplerBindingType::NonFiltering => wgpu_types::BindingType::Sampler {
-                filtering: false,
-                comparison: false,
-            },
-            GpuSamplerBindingType::Comparison => wgpu_types::BindingType::Sampler {
-                filtering: true,
-                comparison: true,
-            },
+            GpuSamplerBindingType::Filtering => {
```
kvark, line 73: with this PR, we shouldn't need `GpuSamplerBindingType` at all. Deno should just reuse `SamplerBindingType` instead directly

wgpu-core/src/device/mod.rs
```diff
@@ -1521,30 +1521,37 @@ impl<A: HalApi> Device<A> {
...
 
-                            // Check the actual sampler to also (not) be a comparison sampler
-                            if sampler.comparison != comparison {
-                                return Err(Error::WrongSamplerComparison {
-                                    binding,
-                                    layout_cmp: comparison,
-                                    sampler_cmp: sampler.comparison,
-                                });
+                            // Allowed sampler values for filtering and comparison
+                            let (allowed_filtering, allowed_comparison) = match ty {
```
kvark, line 1531: since `allowed_comparison` is always `Some`, why does it need to be an `Option`?

wgpu-core/src/validation.rs
```diff
@@ -536,10 +533,13 @@ impl Resource {
                 has_dynamic_offset: false,
                 min_binding_size: Some(size),
             },
-            ResourceType::Sampler { comparison } => BindingType::Sampler {
-                filtering: true,
-                comparison,
-            },
+            ResourceType::Sampler { comparison } => {
+                if comparison {
+                    BindingType::Sampler(wgt::SamplerBindingType::Comparison)
```
kvark, line 538: nit: let's move the `if` inside, e.g.
```rust
BindingType::Sampler(if comparison {..} else {..})
```

kvark (changes requested):
Thank you for doing this!
Aligning with WebGPU spec is appreciated.

wgpu-core/src/device/mod.rs
Bobo1239: Good point, pulled out the `Option`.

kvark (approved):

PR #2183, opened 2021-11-15
Don't use duplicate function name in boids shader

Bobo1239:
**Description**
From the spec: "A declaration must not introduce a name when that
identifier is already in scope with the same end scope as another
instance of that name."

naga doesn't yet check this (gfx-rs/naga#1480) but Chrome/Tint rightly complains with:
error: redefinition of 'main'

**Testing**
Tested the boids example natively and in both Firefox Nightly and Chrome Canary.


kvark (approved):
Thank you!

Issue #2186, opened 2021-11-16
Multi-view rendering 


LaylBongers:
## Is your feature request related to a problem? Please describe.

With the recent move to wgpu-hal, and now being able to access the raw backing types, it has become possible to use OpenXR with the current released version of WGPU. (It works pretty well too!)

However, for efficient VR stereo rendering, multi-view has become the standard in most engines. Multi-view allows rendering to two different targets (array layers on the same texture), using different matrices. Currently WGPU does not provide multi-view support.

## Describe the solution you'd like

Preferably, multi-view support at least for the Vulkan target would be nice, as almost all platforms supporting VR currently support the Vulkan extension. (With the exception of WMR, which only supports DX11 and DX12, but SteamVR can act as a bridge for that to provide support)

Multi-view has become core in Vulkan 1.1, and in 1.0 is broadly supported through an identical `VK_KHR_multiview` extension.

## Describe alternatives you've considered

Other alternative options are mostly impractical. Older solutions such as using geometry shaders are no longer necessary because of multi-view's broad support. Multi-view could be implemented using raw handle access but because of the simplicity of adding support this isn't really necessary.

Edit: Multi-view-like functionality can also be accomplished through a compute shader, though I haven't measured the difference in performance and I don't know how this affects support for mobile targets such as the Quest.

## Additional context

[VK_KHR_multiview](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_KHR_multiview.html)
[VkRenderPassMultiviewCreateInfo](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkRenderPassMultiviewCreateInfo.html)

schell:
@LaylBongers in your PR description you said that OpenXR and wgpu was "working pretty well" - which sounds great! Do you have any pointers or links to getting the two working together?

philpax:
> @LaylBongers in your PR description you said that OpenXR and wgpu was "working pretty well" - which sounds great! Do you have any pointers or links to getting the two working together?

I just publicised my example of integrating OpenXR and wgpu (I came across this issue while looking for example code) - you can find more details here: https://github.com/gfx-rs/wgpu/issues/602#issuecomment-1288217600

PR #2187, opened 2021-11-16
Add multiview support

LaylBongers:
## Connections
Fixes: https://github.com/gfx-rs/wgpu/issues/2186

## Description
Adds basic multiview support.

## Testing
You can now create a render pass with a target view with two array layers.

### DIFF
--- a/wgpu-core/src/command/render.rs
+++ b/wgpu-core/src/command/render.rs
@@ -9,8 +9,8 @@ use crate::{
         RenderCommandError, StateChange,
     },
     device::{
-        AttachmentData, MissingDownlevelFlags, MissingFeatures, RenderPassCompatibilityError,
-        RenderPassContext,
+        AttachmentData, Device, MissingDownlevelFlags, MissingFeatures,
+        RenderPassCompatibilityError, RenderPassContext,
     },
     error::{ErrorFormatter, PrettyError},
     hub::{Global, GlobalIdentityHandlerFactory, HalApi, Storage, Token},
@@ -29,7 +29,8 @@ use arrayvec::ArrayVec;
 use hal::CommandEncoder as _;
 use thiserror::Error;
 use wgt::{
-    BufferAddress, BufferSize, BufferUsages, Color, IndexFormat, TextureUsages, VertexStepMode,
+    BufferAddress, BufferSize, BufferUsages, Color, IndexFormat, TextureUsages,
+    TextureViewDimension, VertexStepMode,
 };
 
 #[cfg(any(feature = "serial-pass", feature = "replay"))]
@@ -461,6 +462,12 @@ pub enum RenderPassErrorInner {
     Bind(#[from] BindError),
     #[error(transparent)]
     QueryUse(#[from] QueryUseError),
+    #[error("multiview layer count must match")]
+    MultiViewMismatch,
+    #[error(
+        "multiview pass texture views with more than one array layer must have D2Array dimension"
+    )]
+    MultiViewDimensionMismatch,
 }
 
 impl PrettyError for RenderPassErrorInner {
@@ -542,6 +549,7 @@ struct RenderPassInfo<'a, A: hal::Api> {
 
     pending_discard_init_fixups: SurfacesInDiscardState,
     divergent_discarded_depth_stencil_aspect: Option<(wgt::TextureAspect, &'a TextureView<A>)>,
+    multiview: Option<NonZeroU32>,
 }
 
 impl<'a, A: HalApi> RenderPassInfo<'a, A> {
@@ -582,6 +590,7 @@ impl<'a, A: HalApi> RenderPassInfo<'a, A> {
     }
 
     fn start(
+        device: &Device<A>,
         label: Option<&str>,
         color_attachments: &[RenderPassColorAttachment],
         depth_stencil_attachment: Option<&RenderPassDepthStencilAttachment>,
@@ -605,6 +614,39 @@ impl<'a, A: HalApi> RenderPassInfo<'a, A> {
         let mut extent = None;
         let mut sample_count = 0;
 
+        let mut detected_multiview: Option<Option<NonZeroU32>> = None;
+
+        let mut check_multiview = |view: &TextureView<A>| {
+            // Get the multiview configuration for this texture view
+            let layers = view.selector.layers.end - view.selector.layers.start;
+            let this_multiview = if layers >= 2 {
+                // Trivially proven by the if above
+                Some(unsafe { NonZeroU32::new_unchecked(layers) })
+            } else {
+                None
+            };
+
+            // Make sure that if this view is a multiview, it is set to be an array
+            if this_multiview.is_some() && view.desc.dimension != TextureViewDimension::D2Array {
+                return Err(RenderPassErrorInner::MultiViewDimensionMismatch);
+            }
+
+            // Validate matching first, or store the first one
+            if let Some(multiview) = detected_multiview {
+                if multiview != this_multiview {
+                    return Err(RenderPassErrorInner::MultiViewMismatch);
+                }
+            } else {
+                // Multiview is only supported if the feature is enabled
+                if this_multiview.is_some() {
+                    device.require_features(wgt::Features::MULTIVIEW)?;
+                }
+
+                detected_multiview = Some(this_multiview);
+            }
+
+            Ok(())
+        };
         let mut add_view = |view: &TextureView<A>, type_name| {
             if let Some(ex) = extent {
                 if ex != view.extent {
@@ -637,6 +679,7 @@ impl<'a, A: HalApi> RenderPassInfo<'a, A> {
                 .views
                 .use_extend(&*view_guard, at.view, (), ())
                 .map_err(|_| RenderPassErrorInner::InvalidAttachment(at.view))?;
+            check_multiview(view)?;
             add_view(view, "depth")?;
 
             let ds_aspects = view.desc.aspects();
@@ -745,6 +788,7 @@ impl<'a, A: HalApi> RenderPassInfo<'a, A> {
                 .views
                 .use_extend(&*view_guard, at.view, (), ())
                 .map_err(|_| RenderPassErrorInner::InvalidAttachment(at.view))?;
+            check_multiview(color_view)?;
             add_view(color_view, "color")?;
 
             if !color_view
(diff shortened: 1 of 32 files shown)


wgpu-core/src/command/render.rs
```diff
@@ -154,6 +154,8 @@ pub struct RenderPassDescriptor<'a> {
     pub color_attachments: Cow<'a, [RenderPassColorAttachment]>,
     /// The depth and stencil attachment of the render pass, if any.
     pub depth_stencil_attachment: Option<&'a RenderPassDepthStencilAttachment>,
+    /// If this render pass will be rendering to multiple array layers.
+    pub multiview: bool,
```
kvark, line 158: is this actually needed? If I understand correctly, we expect all of the attachments to be layered in this case, and with the same number of layers in them. So this flag becomes redundant - one could just check the actual texture views in the attachments.

wgpu-core/src/command/render.rs
```diff
@@ -162,6 +164,7 @@ pub struct RenderPass {
     parent_id: id::CommandEncoderId,
     color_targets: ArrayVec<RenderPassColorAttachment, { hal::MAX_COLOR_TARGETS }>,
     depth_stencil_target: Option<RenderPassDepthStencilAttachment>,
+    multiview: bool,
```
kvark, line 167: same here?

wgpu-core/src/pipeline.rs
```diff
@@ -243,6 +243,8 @@ pub struct RenderPipelineDescriptor<'a> {
     pub multisample: wgt::MultisampleState,
     /// The fragment processing state for this pipeline.
     pub fragment: Option<FragmentState<'a>>,
+    /// If the pipeline will be used with a multiview render pass.
+    pub multiview: bool,
```
kvark, line 247: would this need to know the number of layers?

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -1052,6 +1052,7 @@ impl super::Adapter {
                 spv::Capability::ImageQuery,
                 spv::Capability::DerivativeControl,
                 spv::Capability::SampledCubeArray,
+                spv::Capability::MultiView,
```
kvark, line 1055: multi-view needs to be an optional feature exposed by the adapter and then activated by the user on device request.
The logic here, therefore, should be to only include this multi-view capability if the feature is activated.

kvark (changes requested):
Looks great for a prototype!
In order to land this, we need a bit more polished look.
In addition to review notes, we'll also need to validate that multi-view is not used when not supported.

wgpu-core/src/pipeline.rs
LaylBongers: Tested with the mask set to all 1s, resulted in really bad driver crash on AMD so I added the layer count here.

wgpu-core/src/command/render.rs
```diff
@@ -605,6 +614,41 @@ impl<'a, A: HalApi> RenderPassInfo<'a, A> {
...
 
+        let mut detected_mutiview: Option<Option<NonZeroU32>> = None;
+
+        let mut check_multiview = |view: &TextureView<A>| {
+            // Get the multiview configuration for this texture view
+            let layers = view.selector.layers.end - view.selector.layers.start;
+            let this_multiview = if layers == 1 {
+                None
+            } else {
+                let layers = NonZeroU32::new(layers).expect("Non-zero layers expected");
```
kvark, line 625: this doesn't need to check for view layers being zero. View creation already does.
So this code could be just `NonZeroU32::new(layers)`

wgpu-core/src/command/render.rs
```diff
@@ -605,6 +614,41 @@ impl<'a, A: HalApi> RenderPassInfo<'a, A> {
...
+                }
+            } else {
+                // Multiview is only supported if the feature is enabled
+                if this_multiview.is_some() && !device.features.contains(wgt::Features::MULTIVIEW) {
+                    return Err(RenderPassErrorInner::MissingFeatures(MissingFeatures(
+                        wgt::Features::MULTIVIEW,
+                    )));
+                }
+
+                detected_mutiview = Some(this_multiview);
```
kvark, line 647: I think this logic isn't going to work right. It scans several attachments and only errors if we are switching from multiview to non-multiview. Imagine if the first analyzed attachment is non-layered but the second is layered. We should error, but this could wouldn't.

wgpu-core/src/command/render.rs
```diff
@@ -745,6 +790,7 @@ impl<'a, A: HalApi> RenderPassInfo<'a, A> {
                 .views
                 .use_extend(&*view_guard, at.view, (), ())
                 .map_err(|_| RenderPassErrorInner::InvalidAttachment(at.view))?;
+            check_multiview(color_view)?;
             add_view(color_view, "color")?;
```
kvark, line 794: we should do this logic in `add_view`, which is already called

wgpu-hal/src/vulkan/device.rs
```diff
@@ -140,10 +142,30 @@ impl super::DeviceShared {
...
                     .subpasses(&vk_subpasses);
 
+                let mut multiview_info;
+                let mask;
+                if let Some(multiview) = e.key().multiview {
+                    // Sanity checks, better to panic here than cause a driver crash
+                    assert!(multiview.get() <= 8);
+                    assert!(multiview.get() > 1);
+
+                    let mut maskvalue = 0;
```
kvark, line 156: could be just `(1 << multiview.get())- 1`

wgpu-types/src/lib.rs
```diff
@@ -526,6 +526,13 @@ bitflags::bitflags! {
         ///
         /// This is a native only feature.
         const SHADER_PRIMITIVE_INDEX = 1 << 39;
+        /// Enables multiview render passes and `builtin(view_index)` in vertex shaders.
+        ///
+        /// Supported platforms:
+        /// - Vulkan
+        ///
+        /// This is a native only feature.
+        const MULTIVIEW = 1 << 40;
```
kvark, line 535: should this really be called `MULTIVIEW`? we don't have separate viewport and scissors for them (yet). It looks like this is more of `LAYERED_RENDERING` thing.

kvark (commented):
Vulkan still needs the proper feature detection

wgpu-core/src/command/render.rs
LaylBongers: The implemented logic here on every view is:
- check if multiview on or off has already been detected
- if it has: check if the multiview on/off and layer count matches
- if it hasn't: use the current view's value as the multiview detection

Subsequent checks will be checked against the first one's detected value. This is what the state of `detected_multiview` means:
- None: Not detected
- Some(None): Detected, no multiview
- Some(Some(2)): Detected, multiview with two layers

wgpu-core/src/command/render.rs
LaylBongers: Did you mean `NonZeroU32::new_unchecked(layers)`?

wgpu-core/src/command/render.rs
LaylBongers: `check_multiview` is called below on the resolve too, as it has to match, which does not have `add_view`. The function takes a mutable reference to `detected_multiview` and thus can't be mutable borrowed by `add_view` in addition to being used there.

wgpu-types/src/lib.rs
LaylBongers: This was discussed and resolved on the matrix chat, so I'll resolve this

wgpu-core/src/device/mod.rs
```diff
@@ -2449,6 +2458,13 @@ impl<A: HalApi> Device<A> {
             .get(pipeline_layout_id)
             .map_err(|_| pipeline::CreateRenderPipelineError::InvalidLayout)?;
 
+        // Multiview is only supported if the feature is enabled
+        if desc.multiview.is_some() && !self.features.contains(wgt::Features::MULTIVIEW) {
```
kvark, line 2462: we should use `self.require_features(..)` here

wgpu-core/src/command/render.rs
```diff
@@ -605,6 +614,41 @@ impl<'a, A: HalApi> RenderPassInfo<'a, A> {
...
+            }
+
+            // Validate matching first, or store the first one
+            if let Some(multiview) = detected_multiview {
+                if multiview != this_multiview {
+                    return Err(RenderPassErrorInner::MultiViewMismatch);
+                }
+            } else {
+                // Multiview is only supported if the feature is enabled
+                if this_multiview.is_some() && !device.features.contains(wgt::Features::MULTIVIEW) {
```
kvark, line 641: should use `device.require_features(..)` here

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -23,6 +23,7 @@ pub struct PhysicalDeviceFeatures {
     image_robustness: Option<vk::PhysicalDeviceImageRobustnessFeaturesEXT>,
     robustness2: Option<vk::PhysicalDeviceRobustness2FeaturesEXT>,
     depth_clip_enable: Option<vk::PhysicalDeviceDepthClipEnableFeaturesEXT>,
+    multiview: Option<vk::PhysicalDeviceMultiviewFeatures>,
```
kvark, line 26: I think this should be `vulkan_1_1: vk::PhysicalDeviceVulkan11Features` and come right between `core` and `vulkan_1_2`.
In addition, I think we'd need `multiview: VkPhysicalDeviceMultiviewFeaturesKHR`, so that we can work with either Vk-1.1 or Vk-1.0 with extensions.

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -1078,17 +1109,17 @@ impl super::Adapter {
                 lang_version: (1, 0),
                 flags,
                 capabilities: Some(capabilities.iter().cloned().collect()),
-                bounds_check_policies: naga::back::BoundsCheckPolicies {
-                    index: naga::back::BoundsCheckPolicy::Restrict,
+                bounds_check_policies: naga::proc::BoundsCheckPolicies {
+                    index: naga::proc::BoundsCheckPolicy::Restrict,
```
kvark, line 1113: ideally, we'd land this in a separate PR, but this isn't blocking

kvark (commented):
This is shaping up well! I think it's almost ready, but we'll need one more iteration.
It would be nice to separate the Naga update from the rest, and we'll need to make Vulkan adapter logic a bit more powerful. I.e. it needs to support both extension and Vk_1_1 path for this, ideally.

wgpu-hal/src/vulkan/adapter.rs
LaylBongers: Decided to keep this in if it's not blocking, as the multiview builtin update is necessary for testing this functionality for me personally.

kvark (approved):
Great!

PR #2188, opened 2021-11-17
Populate sampling pairs for validation

### LINKED ISSUE #2182: `sampling_pairs` validation doesn't actually get populated and thus doesn't work
**Description**
As far as I can tell the following validation exists in code but doesn't actually run since `sampling_pairs` is never populated.
https://github.com/gfx-rs/wgpu/blob/c824c8fa53d938aa835400251512acb27ac50d72/wgpu-core/src/validation.rs#L1021-L1057
The repro case below is a project where I've noticed the missing validation.

**Repro steps**
Check out https://github.com/hecrj/wgpu_glyph/tree/215f3be7c5ee6f9788d5eefe600ca2271bc5373a which uses a [filtering sampler with an unfilterable texture](https://github.com/hecrj/wgpu_glyph/blob/215f3be7c5ee6f9788d5eefe600ca2271bc5373a/src/pipeline.rs#L239-L250). Run `cargo run --example hello`.

The example should fail validation but doesn't.

Note that Chrome/Tint does report an validation error with the WASM version:
> Texture binding (group:0, binding:2) is TextureSampleType::UnfilterableFloat but used statically with a sampler (group:0, binding:1) that's SamplerBindingType::Filtering
> - While validating fragment stage (module: [ShaderModule "Glyph Shader"], entryPoint: fs_main).
> - While validating fragment state.
> - While calling [Device].CreateRenderPipeline([RenderPipelineDescriptor]).

kvark:
**Connections**
Fixes #2182

**Description**
We didn't populate the set

**Testing**
Modified the skybox example:
```
Caused by:
      note: label = `Sky`
    error matching FRAGMENT shader requirements against the pipeline
    unable to filter the texture (ResourceBinding { group: 0, binding: 1 }) by the sampler (ResourceBinding { group: 0, binding: 2 })
    non-filterable float texture

```

### DIFF
--- a/wgpu-core/src/validation.rs
+++ b/wgpu-core/src/validation.rs
@@ -934,6 +934,11 @@ impl Interface {
                 }
             }
 
+            for key in info.sampling_set.iter() {
+                ep.sampling_pairs
+                    .insert((resource_mapping[&key.image], resource_mapping[&key.sampler]));
+            }
+
             ep.workgroup_size = entry_point.workgroup_size;
 
             entry_points.insert((entry_point.stage, entry_point.name.clone()), ep);


Jerrody:
Hey, in Rust was added some syntax sugar with that you can iter over an array without explicitly writing `iter()`.
Like:
```rs
            for key in info.sampling_set.iter() {
                ep.sampling_pairs
                    .insert((resource_mapping[&key.image], resource_mapping[&key.sampler]));
            }
```

kvark:
Oh yes, I'm well aware, just like my iterators explicit.

kvark:
Published in wgpu-core-0.11.3

PR #2195, opened 2021-11-19
Validate min_binding_size at draw time

### LINKED ISSUE #2194: Validate min_binding_size late if not provided
Currently, we validate the binding size at BindGroup creation if `min_binding_size` was specified. However, if it's not specified, we still need to validate it. We just have to do it later, at draw time, instead.

kvark:
**Connections**
Fixes #2194

**Description**
I think validating bindings is very important, because users can waste hours investigating driver crashes otherwise.
The main sell of `wgpu` is that they should trust us.

Implementation-wise, it's complicated because we can't just go through all the bindings at draw time - this could be too slow. So there is a compaction scheme in place.

**Testing**
Tweaking the cube example gives me:
```
Caused by:
    In a RenderPass
      note: render pipeline = `<RenderPipeline-(0, 1, Vulkan)>`
    Buffer is bound with size 64 where the shader expects 80 in group[0] compact index 0
```

Obviously, the error could be nicer, but I don't want to invest into that just yet.


cwfitzgerald (approved):
LGTM! Some very clever code to help with performance, well done

PR #2201, opened 2021-11-21
use native webgl2 buffers for MAP_READ usage

mrk-its:
**Description**
While memory-only buffers work great for MAP_WRITE usage, it seems it is better to create regular webgl2 buffers for MAP_READ. They may be created with _READ usage hint - it signals intent of reading back data from buffer and allows webgl2 to prefetch data to host memory and makes non-blocking read possible (take a look on bottom note of https://www.khronos.org/registry/webgl/specs/latest/2.0/#3.7.3)

**Testing**
I'm testing this approach in my application reading texture data each frame (`copy_texture_to_buffer`, then reading buffer data in next frame) and it works, I was able to significantly reduce time spent on reading buffer data back (comparing to direct reading texture data)

### DIFF
--- a/wgpu-hal/src/gles/device.rs
+++ b/wgpu-hal/src/gles/device.rs
@@ -1,7 +1,7 @@
-use super::{conv, BufferInner};
+use super::conv;
 use crate::auxil::map_naga_stage;
 use glow::HasContext;
-use std::{convert::TryInto, iter, ptr, sync::Arc};
+use std::{convert::TryInto, iter, ptr, sync::{Arc, Mutex}};
 
 #[cfg(not(target_arch = "wasm32"))]
 use std::mem;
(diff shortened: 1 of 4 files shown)


wgpu-hal/src/gles/device.rs
```diff
@@ -412,11 +413,18 @@ impl crate::Device<super::Api> for super::Device {
             }
         }
 
+        let map_read_allocation = if emulate_map && desc.usage.contains(crate::BufferUses::MAP_READ) {
+            Some(std::sync::Mutex::new(vec![0; desc.size as usize]))
```
kvark, line 417: It's a bit unfortunate to have this data buffer present in either `BufferInner::Data` or here. Can we refactor `struct Buffer` to avoid this duplication? Maybe
```
struct Buffer {
    raw: Option<glow::Buffer>,
    target: BindTarget,
    size: wgt::BufferAddress,
    map_flags: u32,
    map_data: Option<Mutex<..>>,
}
```

mrk-its: good point, will fix

kvark (approved):

PR #2209, opened 2021-11-24
Unique HalManagedMetalLayerDelegate class

xiaopengli89:
**Description**

When there are multiple versions of wgpu in the project, creating multiple versions of Instance on macOS will cause a panic, because the `HalManagedMetalLayerDelegate` class can only be created once.

### DIFF
--- a/wgpu-hal/src/metal/surface.rs
+++ b/wgpu-hal/src/metal/surface.rs
@@ -1,4 +1,4 @@
-use std::{mem, os::raw::c_void, ptr::NonNull, thread};
+use std::{mem, os::raw::c_void, ptr::NonNull, sync::Once, thread};
 
 use core_graphics_types::{
     base::CGFloat,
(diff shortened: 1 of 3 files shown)


wgpu-hal/src/metal/surface.rs
```diff
@@ -31,27 +32,29 @@ extern "C" fn layer_should_inherit_contents_scale_from_window(
     YES
 }
 
-const CAML_DELEGATE_CLASS: &str = "HalManagedMetalLayerDelegate";
-static CAML_DELEGATE_REGISTER: Once = Once::new();
+static CAML_DELEGATE_REGISTER: OnceCell<&'static Class> = OnceCell::new();
```
kvark, line 35: do I understand correctly that we just need a pointer, and that's why we use `OnceCell` here?
In this case, we should be able to avoid `once-cell` dependency and instead just declare a dummy static thing, to get a pointer from.

kvark (changes requested):

kvark (approved):

kvark:
Published in wgpu-hal-0.11.5

PR #2212, opened 2021-11-25
hal/vulkan: use multiple semaphores in a relay

kvark:
**Connections**
Attempting to work around #1672, which is currently blocking gecko.
https://phabricator.services.mozilla.com/D132058

**Description**
Intel apparently doesn't like that the same semaphore is both waited upon and signaled.
This PR makes it 2 semaphores, which we use in ping-pong fashion.

**Testing**
On Gecko


kvark:
Confirmed this to make vange-rs work.

kvark:
Published in wgpu-hal-0.11.5

PR #2218, opened 2021-11-27
hal/vulkan: Enable depth clip if available

JCapucho:
**Connections**
None that I know of

**Description**
The `DEPTH_CLIP_CONTROL` feature was failing to be requested when using
the vulkan backend because the functionality wasn't being activated.

This wasn't being done even if the VK_EXT_depth_clip_enable was present
causing validation errors when requesting DEPTH_CLIP_CONTROL in the
device.

~~Strangely this was being done for raw calls to `wgpu_hal::Adapter::open`
because it calls `physical_device_features` and then
`from_extensions_and_requested_features`.~~

**Testing**
Modified the cube test to require this feature, without this patch it failed to start.
Furthermore I checked that my gpu really did support the extension trough `vulkaninfo`.

### DIFF
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -821,6 +821,13 @@ impl super::InstanceShared {
                 let mut_ref = features.robustness2.as_mut().unwrap();
                 mut_ref.p_next = mem::replace(&mut features2.p_next, mut_ref as *mut _ as *mut _);
             }
+            if capabilities.supports_extension(vk::ExtDepthClipEnableFn::name()) {
+                features.depth_clip_enable =
+                    Some(vk::PhysicalDeviceDepthClipEnableFeaturesEXT::builder().build());
+
+                let mut_ref = features.depth_clip_enable.as_mut().unwrap();
+                mut_ref.p_next = mem::replace(&mut features2.p_next, mut_ref as *mut _ as *mut _);
+            }
 
             unsafe {
                 get_device_properties.get_physical_device_features2_khr(phd, &mut features2);
@@ -847,6 +854,7 @@ impl super::InstanceShared {
             null_p_next(&mut features.timeline_semaphore);
             null_p_next(&mut features.image_robustness);
             null_p_next(&mut features.robustness2);
+            null_p_next(&mut features.depth_clip_enable);
         }
 
         (capabilities, features)


kvark:
Interesting, I ran the shadow example and didn't see any validation errors on my end. Although, maybe my validation layers are inactive on that new machine...

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -821,6 +821,16 @@ impl super::InstanceShared {
                 let mut_ref = features.robustness2.as_mut().unwrap();
                 mut_ref.p_next = mem::replace(&mut features2.p_next, mut_ref as *mut _ as *mut _);
             }
+            if capabilities.supports_extension(vk::ExtDepthClipEnableFn::name()) {
+                features.depth_clip_enable = Some(
+                    vk::PhysicalDeviceDepthClipEnableFeaturesEXT::builder()
+                        .depth_clip_enable(true)
```
kvark, line 827: this is not a place where features are enabled. This is where they are queried, so this line needs to be removed (see like everything else is done in this block)

kvark (changes requested):

JCapucho:
to be clear the validation error is from wgpu not the vulkan validation layers

JCapucho:
For example applying this patch 
```patch
diff --git a/wgpu/examples/cube/main.rs b/wgpu/examples/cube/main.rs
index f06c5643..9f6a3887 100644
--- a/wgpu/examples/cube/main.rs
+++ b/wgpu/examples/cube/main.rs
@@ -111,6 +111,10 @@ impl framework::Example for Example {
         wgt::Features::POLYGON_MODE_LINE
     }
 
+    fn required_features() -> wgpu::Features {
+        wgt::Features::DEPTH_CLIP_CONTROL
+    }
+
     fn init(
         config: &wgpu::SurfaceConfiguration,
         _adapter: &wgpu::Adapter,

```
Gets me
```
Using NVIDIA GeForce GTX 1060 with Max-Q Design (Vulkan)
thread 'main' panicked at 'Adapter does not support required features for this example: DEPTH_CLIP_CONTROL', wgpu/examples/cube/../framework.rs:147:5
```
But `vulkaninfo` reports the contrary
```sh
$ vulkaninfo | rg "VK_EXT_depth_clip_enable"
	VK_EXT_depth_clip_enable                  : extension revision 1
```

wgpu-hal/src/vulkan/adapter.rs
JCapucho: Do I just remove that line? I can't seem to find where the features are enabled

kvark: Yes. It's enabled here - https://github.com/gfx-rs/wgpu/blob/943145b4df9dd4c0bc236a76ef96acd13c46fead/wgpu-hal/src/vulkan/adapter.rs#L305

JCapucho: Oh okay, thanks for pointing me to the right place.

kvark: still unresolved

JCapucho: classic forgot to push

kvark (approved):

PR #2220, opened 2021-11-28
hal/gles: fix setting filtering on integer textures

kvark:
**Connections**
[_Link to the issues addressed by this PR, or dependent PRs in other repositories_](https://github.com/kvark/vange-rs/issues/153#issuecomment-975163806)

**Description**
Silly, we bound the texture after setting the filter modes...

**Testing**
Tested on vange-rs


kvark:
oh, I need to get my thoughts fixed

PR #2222, opened 2021-11-29
hal/gles: fix setting filtering on integer textures

kvark:
**Connections**
Addresses an error reported by RenderDoc upon capturing.

**Description**
RenderDoc seems to expect that we first set the filtering modes and then initialize the texture.

**Testing**
Tested on vange-rs


kvark:
Published in wgpu-hal-0.11.5

PR #2226, opened 2021-11-29
hal/gles: force unbind samplers

kvark:
**Connections**
Discovered in vange-rs

**Description**
It turns out that we can't just leave the sampler object in place upon changing a texture binding. The sampler has to be compatible with the texture, or GL screams in pain, even if we don't actually sample anything.
I.e. for UINT textures, the sampler can't be linear.

**Testing**
Tested on vange-rs


kvark:
vangers-gles-2

kvark:
Published in wgpu-hal-0.11.5

PR #2227, opened 2021-11-29
Relax depth texture views creation

IcanDivideBy0:
**Connections**
n/a

**Description**
Currently impossible to sample a Depth24PlusStencil texture, this solves the problem by allowing to create a Depth24Plus view on a Depth24PlusStencil texture.

**Testing**
Tested on Debian Bullseye, Nvidia + VK backend

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -822,11 +822,15 @@ impl<A: HalApi> Device<A> {
             extent.depth_or_array_layers = view_layer_count;
         }
         let format = desc.format.unwrap_or(texture.desc.format);
-        if format != texture.desc.format {
-            return Err(resource::CreateTextureViewError::FormatReinterpretation {
-                texture: texture.desc.format,
-                view: format,
-            });
+        match (texture.desc.format, format) {
+            (a, b) if a == b => {}
+            (wgt::TextureFormat::Depth24PlusStencil8, wgt::TextureFormat::Depth24Plus) => {}
+            _ => {
+                return Err(resource::CreateTextureViewError::FormatReinterpretation {
+                    texture: texture.desc.format,
+                    view: format,
+                });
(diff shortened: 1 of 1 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -822,11 +822,15 @@ impl<A: HalApi> Device<A> {
...
         }
         let format = desc.format.unwrap_or(texture.desc.format);
-        if format != texture.desc.format {
-            return Err(resource::CreateTextureViewError::FormatReinterpretation {
-                texture: texture.desc.format,
-                view: format,
-            });
+        match (texture.desc.format, format) {
+            (a, b) if a == b => {}
+            (wgt::TextureFormat::Depth24PlusStencil8, wgt::TextureFormat::Depth24Plus) => {}
```
kvark, line 827: unfortunately, we can't offer these copies, see https://github.com/gpuweb/gpuweb/issues/2322#issuecomment-972344119
> 2021-11-17 investigation #2322 (comment) showed T2T copies must always copy all aspects and have exactly the same format. I must have gotten confused between T2T and B2T/T2B copies while writing this up.

kvark (changes requested):

wgpu-core/src/device/mod.rs
IcanDivideBy0: I'm not sure to understand, this is not about texture copies, but texture views creation. Does the same rules applies here?

IcanDivideBy0: ah ok, I got it now, creating a view with `aspect: wgpu::TextureAspect::DepthOnly` on a Depth24PlusStencil is enouth, no need to change the format

kvark: right, sorry I mistakenly thought this is about copies

Issue #2215, opened 2021-11-26
Examples crash with "Unable to load get_semaphore_counter_value"


NiklasEi:
**Description**
All wgpu examples crash with `Unable to load get_semaphore_counter_value`.

**Expected vs observed behavior**
I expect the example to run, but it crashes instead.

**Extra materials**
Output from running `boids` example:
```
➜  wgpu git:(master) ✗ cargo run --example boids
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/examples/boids`
Using Intel(R) UHD Graphics 630 (CFL GT2) (Vulkan)
thread 'main' panicked at 'Unable to load get_semaphore_counter_value', /home/niklas/.cargo/registry/src/github.com-1ecc6299db9ec823/ash-0.33.3+1.2.191/src/vk/features.rs:6307:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'Unable to load wait_semaphores', /home/niklas/.cargo/registry/src/github.com-1ecc6299db9ec823/ash-0.33.3+1.2.191/src/vk/features.rs:6328:21
stack backtrace:
   0:     0x562c702c278c - std::backtrace_rs::backtrace::libunwind::trace::h3fea1eb2e0ba2ac9
                               at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
   1:     0x562c702c278c - std::backtrace_rs::backtrace::trace_unsynchronized::h849d83492cbc0d59
                               at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x562c702c278c - std::sys_common::backtrace::_print_fmt::he3179d37290f23d3
                               at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/sys_common/backtrace.rs:67:5
   3:     0x562c702c278c - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h140f6925cad14324
                               at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/sys_common/backtrace.rs:46:22
   4:     0x562c702e562c - core::fmt::write::h31b9cd1bedd7ea38
[...]

PR #2228, opened 2021-11-29
hal/vk: fix support for vk-1.0 and refactor the adapter code

kvark:
**Connections**
Investigating #2215

**Description**
Previously, we were considering the instance version and extensions for checking `KhrStorageBufferStorageClassFn`, but it's a device extension... Effectively means we couldn't run on VkInstance 1.0 at all.

This PR raises MSRV to 1.53 because of `Option::insert()`.

**Testing**
Tested on examples locally.

### DIFF
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -751,8 +751,7 @@ impl super::InstanceShared {
                 }
 
                 unsafe {
-                    get_device_properties
-                        .get_physical_device_properties2_khr(phd, &mut properites2);
+                    get_device_properties.get_physical_device_properties2(phd, &mut properites2);
                 }
                 properites2.properties
             } else {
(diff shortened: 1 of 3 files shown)


kvark:
Published in wgpu-hal-0.11.5

parasyte:
Heads up! I don't know if you have a minimum supported Rust version policy (I couldn't find one) but `wgpu-hal` 0.11.5 broke my CI because this PR uses `Option::insert` which was stabilized in Rust 1.53.0. (My CI was building with Rust 1.52.0, which was our MSRV).

kvark:
Rrright, sorry about that! Are you fine updating MSRV to 1.53, or is it a showstopper? This version is close to 6 month old.

parasyte:
I’m good with just updating the compiler on my end for CI.

It’s a bit concerning though that this changes the policy of versions I released some months ago, too. That’s why I have a heads up. We should be careful about making these kinds of changes in minor releases.

kvark:
Interesting idea - to only change MSRV in breaking versions. Thanks for the hint, I'll try to stick with it.

PR #2230, opened 2021-11-29
Allows for depth texture copies

IcanDivideBy0:
**Connections**
n/a

**Description**
Currently, depth texture cannot use the COPY_SRC or COPY_DST usage flag, this is now possible. The copy checks the whole texture is copied.

**Testing**
Debian bullseye, Nvidia + VK backend


kvark (approved):
Thank you!
We'll also need a check to prevent buffer -> depth copies, but this can be done separately (or maybe it's already in).

PR #2231, opened 2021-11-29
hal/vk: fix adapter vulkan version check when requesting 1.2 capabilities

kvark:
**Connections**
#2215 (unlikely fixes it just yet)

**Description**
We were using the adapter version in order to form a capability request that would tell us... the adapter version, among other things. This means, we were never requesting Vulkan-1.2 capabilities.
This fix should affect our exposed resource limits on Vulkan (cc @cwfitzgerald )

Also has a commit to follow-up to #1898

**Testing**
If @NiklasEi could give it a shot, that would be wonderful!


kvark:
Published in wgpu-hal-0.11.5

PR #2232, opened 2021-11-29
core: Use a indexmap for bindings

JCapucho:
**Connections**
None that I know of

**Description**
This was being done by a standard `HashMap` but in
`make_late_sized_buffer_groups` the values were being
iterated on and since the std hashmap doesn't guaranteed
ordering the buffer might get the wrong size making for
a very nice debugging experience.

`indexmap` is already used in naga and it seemed like the best solution
for the problem, but I'm open to better suggestions.

**Testing**
This is very hard to test, all the examples ran fine but larger projects like
[veloren](https://gitlab.com/veloren/veloren) which I was using for testing hit this.

In this specific case, the fourth and ninth bindings were coming before the third when
iterating.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1839,6 +1839,7 @@ dependencies = [
  "codespan-reporting",
  "copyless",
  "fxhash",
+ "indexmap",
  "log",
  "naga",
  "parking_lot",
--- a/wgpu-core/Cargo.toml
+++ b/wgpu-core/Cargo.toml
@@ -34,6 +34,7 @@ ron = { version = "0.7", optional = true }
 serde = { version = "1.0", features = ["serde_derive"], optional = true }
 smallvec = "1"
 thiserror = "1"
+indexmap = "1.6" # 1.7 has MSRV 1.49
 
 [dependencies.naga]
 git = "https://github.com/gfx-rs/naga"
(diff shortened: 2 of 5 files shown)


kvark:
Could you please describe a problematic scenario?

Yes, hashmap iteration isn't stable between modifications. But an immutable hashmap will always be iterated the same way. So my understanding is that the code should be correct as it stands today.

JCapucho:
> Could you please describe a problematic scenario?

Sure, in case the hashmap iteration doesn't correspond to the shader bindings order, `make_late_sized_buffer_groups` in `wgpu_core::Device` will output a unordered array, for example let's say we have 3 bindings from 0 to 2 in a group, and the hashmap iterates as 1,0,2 the result of `make_late_sized_buffer_groups` will be `[size1, size0, size2]`.

https://github.com/gfx-rs/wgpu/blob/28ba9d86ebb30173c910303b1ceb38e101a8a4a1/wgpu-core/src/device/mod.rs#L1122-L1140

Then this is `zip`ed in `make_late_sized_buffer_groups` on the `Binder` with the bindings but they will have a different order, so the binding will have the wrong size attached.

https://github.com/gfx-rs/wgpu/blob/28ba9d86ebb30173c910303b1ceb38e101a8a4a1/wgpu-core/src/command/bind.rs#L198-L204

kvark:
> Then this is ziped in make_late_sized_buffer_groups on the Binder with the bindings but they will have a different order, so the binding will have the wrong size attached.

The `late_buffer_bindings` are indexed by "compact index", which is pretty much defined by the iteration order here and nothing else. The idea is that it keeps the pipeline side of information on the binder, and it is expected that by the draw() time the pipeline side will match the binding side.

Would you mind filing a bug for your original issue, attaching an API trace or a repro case, please?

JCapucho:
I've opened #2236 and tested that this PR fixes it.

PR #2233, opened 2021-11-30
Support for --target wasm32-unknown-emscripten

caiiiycuk:
**Description**
I found that adding support for `wasm32-unknown-emscripten` is not so hard. Even more implementation is not depended on 3rd party libraries like winit or stdweb. It's because emscripten it self provides complete implementation for creating OpenGL 3.0 context. So crates like glow, khronos-egl works as is. Only visible differences between emscripten and native platform is that emscripten does not support pbuffer (which is not needed) and window will created automatically.

I think supporting `wasm32-unknown-emscripten` will be a great unique feature of wgpu.

**Testing**

You can test this changes by calling `./run-wasm-emscripten-example.sh`. However you need to install [emsdk](https://github.com/emscripten-core/emsdk) first, and `emcc` should be in your path.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,6 +94,14 @@ jobs:
             tool: clippy
             kind: web
 
+          # Emscripten WebGL
+          - name: WebAssembly emscripten
+            os: ubuntu-20.04
+            channel: stable
+            target: wasm32-unknown-emscripten
+            tool: clippy
+            kind: webgl-em
+
     name: Check ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
 
@@ -163,6 +171,15 @@ jobs:
 
           # build docs
           cargo doc --target ${{ matrix.target }} -p wgpu --no-deps
+
+      - name: check web-em
+        if: matrix.kind == 'webgl-em'
+        run: |
+          # check with no features
+          cargo ${{matrix.tool}} --target ${{ matrix.target }} -p wgpu -p wgpu-core
+
+          # build docs
+          cargo doc --target ${{ matrix.target }} -p wgpu -p wgpu-core --no-deps
 
       - name: check native
         if: matrix.kind == 'local' || matrix.kind == 'other'
--- a/wasm-resources/index.em.template.html
+++ b/wasm-resources/index.em.template.html
@@ -1,0 +1,15 @@
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  </head>
+  <body>
+    <canvas id="canvas" width="640" height="400"></canvas>
+    <script>
+        var Module = {
+            canvas: document.getElementById("canvas"),
+        };
+    </script>
+    <script src="{{example}}.js"></script>
+  </body>
+</html>
--- a/wgpu-hal/src/gles/egl.rs
+++ b/wgpu-hal/src/gles/egl.rs
@@ -10,10 +10,15 @@ const CONTEXT_LOCK_TIMEOUT_SECS: u64 = 1;
 const EGL_CONTEXT_FLAGS_KHR: i32 = 0x30FC;
 const EGL_CONTEXT_OPENGL_DEBUG_BIT_KHR: i32 = 0x0001;
 const EGL_CONTEXT_OPENGL_ROBUST_ACCESS_EXT: i32 = 0x30BF;
+#[cfg_attr(target_os = "emscripten", allow(dead_code))]
 const EGL_PLATFORM_WAYLAND_KHR: u32 = 0x31D8;
+#[cfg_attr(target_os = "emscripten", allow(dead_code))]
 const EGL_PLATFORM_X11_KHR: u32 = 0x31D5;
+#[cfg_attr(target_os = "emscripten", allow(dead_code))]
 const EGL_PLATFORM_ANGLE_ANGLE: u32 = 0x3202;
+#[cfg_attr(target_os = "emscripten", allow(dead_code))]
 const EGL_PLATFORM_ANGLE_NATIVE_PLATFORM_TYPE_ANGLE: u32 = 0x348F;
+#[cfg_attr(target_os = "emscripten", allow(dead_code))]
 const EGL_PLATFORM_ANGLE_DEBUG_LAYERS_ENABLED: u32 = 0x3451;
 const EGL_GL_COLORSPACE_KHR: u32 = 0x309D;
 const EGL_GL_COLORSPACE_SRGB_KHR: u32 = 0x3089;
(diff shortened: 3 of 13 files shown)


caiiiycuk:
Can I just apply `cargo fmt` to fix formatting errors?

kvark:
> Can I just apply cargo fmt to fix formatting errors?

yes, that's what we do.
Unless the rustfmt version on CI is very different from yours

Cargo.toml
```diff
@@ -24,7 +24,7 @@ default-members = ["wgpu", "wgpu-hal", "wgpu-info"]
 [patch."https://github.com/zakarumych/gpu-alloc"]
 #gpu-alloc = { path = "../gpu-alloc/gpu-alloc" }
 
-[patch."https://github.com/grovesNL/glow"]
+#[patch."https://github.com/grovesNL/glow"]
```
kvark, line 27: accidental change?

wgpu/examples/hello-triangle-em/main.rs
```diff
@@ -0,0 +1,134 @@
+use env_logger::Builder;
```
kvark, line 1: does this need to be a separate example? Can we instead just have `#[cfg(target_os = "emscripten")]` within the original hello-triangle?

wasm-resources/index.em.template.html
```diff
@@ -0,0 +1,1300 @@
+<!doctype html>
```
kvark, line 1: can this template be reduced, or is this a standard one?

kvark (commented):
Thank you for experimenting with this!
I agree it doesn't look too bad, and we can have this support.
In addition to the notes in the review, we need a CI job to test this configuration.

wasm-resources/index.em.template.html
caiiiycuk: This is standard one, however I can reduce it to simple one.

caiiiycuk:
Blocked by grovesNL/glow#195

wasm-resources/index.em.template.html
kvark: would be great!

caiiiycuk:
Done

.github/workflows/ci.yml
```diff
@@ -77,6 +77,13 @@ jobs:
             target: wasm32-unknown-unknown
             kind: webgl
 
+          # Emscripten WebGL
```
kvark, line 80: sorry, I'm doing a moderate refactor in #2249
we'll need to rebase on top.

Cargo.toml
```diff
@@ -24,7 +24,7 @@ default-members = ["wgpu", "wgpu-hal", "wgpu-info"]
 [patch."https://github.com/zakarumych/gpu-alloc"]
 #gpu-alloc = { path = "../gpu-alloc/gpu-alloc" }
 
-[patch."https://github.com/grovesNL/glow"]
+[patch."https://github.com/caiiiycuk/glow"]
```
kvark, line 27: we are blocked on this currently?

wasm-resources/index.em.template.html
```diff
@@ -0,0 +1,15 @@
+<html>
```
kvark, line 1: nice!

Cargo.toml
caiiiycuk: I don't know why maintainer of glow is not responding to my PR. I think it's better to wait until he merges it...

caiiiycuk:
Updated. Blocker is gone, can be merged now if it's ok

.github/workflows/ci.yml
```diff
@@ -94,6 +94,14 @@ jobs:
             tool: clippy
             kind: web
 
+          # Emscripten WebGL
+          - name: WebAssembly emscripten
+            os: ubuntu-20.04
+            channel: stable
```
kvark, line 100: line no longer applicable

.github/workflows/ci.yml
```diff
@@ -164,6 +172,15 @@ jobs:
           # build docs
           cargo doc --target ${{ matrix.target }} -p wgpu --no-deps
 
+      - name: check web-em
+        if: matrix.kind == 'webgl-em'
+        run: |
+          # check with no features
+          cargo ${{matrix.tool}} --target ${{ matrix.target }} -p wgpu -p wgpu-core
```
kvark, line 179: I think we need to do the same as regulalr webgl - require `webgl` feature to be passed in. Otherwise, this will be confusing when we get emscripten webgpu target.

wgpu-hal/src/gles/egl.rs
```diff
@@ -519,15 +538,13 @@ unsafe impl Send for Instance {}
 unsafe impl Sync for Instance {}
 
 impl crate::Instance<super::Api> for Instance {
+    #[cfg_attr(target_os = "emscripten", allow(unused_variables, unused_mut))]
     unsafe fn init(desc: &crate::InstanceDescriptor) -> Result<Self, crate::InstanceError> {
-        let egl_result = if cfg!(windows) {
```
kvark, line 523: why is this removed?

wgpu-hal/src/gles/egl.rs
```diff
@@ -613,6 +637,9 @@ impl crate::Instance<super::Api> for Instance {
             (display, None, WindowKind::Unknown)
         };
 
+        #[cfg(target_os = "emscripten")]
+        let display = egl.get_display(egl::DEFAULT_DISPLAY).unwrap();
```
kvark, line 641: wouldn't it fall under this path anyway in the previous `if/else` chain?

wgpu-hal/src/gles/egl.rs
```diff
@@ -946,6 +979,8 @@ impl crate::Surface<super::Api> for Surface {
                         wl_window = Some(window);
                         window
                     }
+                    #[cfg(target_os = "emscripten")]
```
kvark, line 982: no need to `cfg` this line

wgpu-hal/src/gles/egl.rs
```diff
@@ -986,27 +1021,35 @@ impl crate::Surface<super::Api> for Surface {
...
-                            .collect::<Vec<_>>();
-                        egl.create_platform_window_surface(
-                            self.display,
-                            self.config,
-                            native_window_ptr,
-                            &attributes_usize,
-                        )
-                    }
-                    _ => self.egl.create_window_surface(
+                #[cfg(not(target_os = "emscripten"))]
```
kvark, line 1024: no need for this cfg, it will go the regular `create_window_surface` way naturally, since `self.wsi.kind == WindowKind::Unknown`

wgpu/Cargo.toml
```diff
@@ -84,13 +84,13 @@ replay = ["serde", "wgc/replay"]
 angle = ["wgc/angle"]
 webgl = ["wgc"]
 
-[target.'cfg(not(target_arch = "wasm32"))'.dependencies.wgc]
+[target.'cfg(any(not(target_arch = "wasm32"), target_os = "emscripten"))'.dependencies.wgc]
```
kvark, line 87: I don't think this configuration needs to change. Instead, it should be aligned with the current webgl path.

wgpu/src/backend/mod.rs
```diff
@@ -1,12 +1,32 @@
-#[cfg(all(target_arch = "wasm32", not(feature = "webgl")))]
+#[cfg(all(
+    target_arch = "wasm32",
+    not(target_os = "emscripten"),
```
kvark, line 3: similarly, this configuration doesn't need to change. If emscripten path requires "webgl" feature, it mostly follows the same steps.

kvark (changes requested):

caiiiycuk:
Will do new one based on "feature"

PR #2237, opened 2021-11-30
Small Vulkan/GLES init tweaks

kvark:
**Connections**
https://github.com/kvark/vange-rs/issues/160
#2215
Follow-up to #2231

**Description**
The last check was still wrong, d'oh!

**Testing**
Locally


kvark:
Published in wgpu-hal-0.11.5

Issue #2247, opened 2021-12-02
copy_texture_to_texture does not check to see if format is compatible.


StarArawn:
**Description**
I attempted to copy a `Rgba8Unorm` texture to a `Rgba16Float`. wgpu doesn't complain but I get the following vulkan error:
```
Validation Error: [ VUID-vkCmdCopyImage-srcImage-01548 ] Object 0: handle = 0x26270e78b30, name = create_sdf_from_atlas_encoder, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x1c1c4e09 | vkCmdCopyImage(): Unmatched 
image format sizes. The src format VK_FORMAT_R8G8B8A8_SRGB has size of 4 and dst format VK_FORMAT_R16G16B16A16_SFLOAT has size of 8. The Vulkan spec states: If the VkFormat of each of srcImage and dstImage is not a multi-planar fo
```

**Repro steps**
1. Create a `Rgba8Unorm` texture.
2. Create a `Rgba16Float` texture.
3. Attempt to copy the `Rgba8Unorm` into the `Rgba16Float` using: `command_encoder.copy_texture_to_texture(..)`

**Expected vs observed behavior**
I think wgpu should complain with an error message here if the format sizes don't match.

**Extra materials**
N/A

**Platform**
Vulkan and Windows with wgpu version 0.11.0

kvark:
Fixed by #2248

PR #2248, opened 2021-12-03
copy_texture_to_texture block size validation

UpsettingBoy:
**Connections**
#2247

**Description**
Simple checks if both the source and destination tessel block size are equal (ex. from `Rgba8Uint` to `R32Float` or even `Rgba10a2Unorm`).  I'm not sure if we should check the format instead.


kvark:
We could get away with this check on some platforms but not the others.
Please refactor this to strictly check for the formats to match.

UpsettingBoy:
Alright, this should check src and dst texture formats are the same.

kvark (approved):

PR #2249, opened 2021-12-03
Enforce MSRV with a large CI refactor

kvark:
**Connections**
https://github.com/gfx-rs/wgpu/pull/2228#issuecomment-985197944

**Description**
Main assumption I'm making here - is that we shouldn't test with "stable" ever.
We either care about MSRV, which we definitely need to run CI on. Or we care about the future breaking, so we pick an arbitrary fresh nightly.
If something works on MSRV and doesn't work on latest stable - it's worrying, but it should be a problem for Rust Core developers more than us, and presumably we'll see it for a while with nightly before it goes into stable.

It also addresses the perpetual problem we had with "clippy" - that new rules are introduced that break our CI. Now clippy is only tested with Rust MSRV. So we only get the breakage when MSRV is pushed forward (in a breaking release).

Finally, looks like we weren't running the player tests for a long time! So they are all fixed and enabled.

**Testing**
CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,7 @@ on:
 
 env:
   RUST_BACKTRACE: 1
+  RUST_VERSION: 1.53
 
 jobs:
   build:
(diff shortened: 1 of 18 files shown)


cwfitzgerald (approved):
CI is a real bother to get right!

LGTM!

Issue #2216, opened 2021-11-27
Does Queue::write_buffer do bounds checks?


danielkeller:
What happens if the data given to Queue::write_buffer goes past the end of the buffer? The docs should mention this.

kvark:
You'll get a validation error, just like with any other copy operation. If this isn't documented, we need to extend the docs there.

teoxoy:
Was addressed by #2250.

PR #2250, opened 2021-12-03
Extended Queue::write_* docs to reflect overrun error

UpsettingBoy:
**Connections**
#2216 

**Description**
Extended documentatio with a bit about the failability of `Queue::write_buffer`.

Should I add something similar to `Queue::write_texture`?


kvark:
> Should I add something similar to Queue::write_texture

yes, that would be welcome!

kvark (approved):

kvark (approved):

PR #2251, opened 2021-12-03
Add SampleRateShading spv capability

IcanDivideBy0:
**Connections**
n/a

**Description**
Using `[[builtin(sample_index)]]` throw an error:
```
Internal error in FRAGMENT shader: using `sample_index` built-in requires at least one of the capabilities [SampleRateShading], but none are available
```
while the feature is always enabled (https://github.com/gfx-rs/wgpu/blob/master/wgpu-hal/src/vulkan/adapter.rs#L103)

**Testing**
Locally tested, everything looks fine


kvark (approved):
Thanks!

PR #2257, opened 2021-12-05
Angle support on Windows

kvark:
**Connections**
Follow-up to #2256
Partially addresses the need to target DX11

**Description**
Makes our GLES/EGL backend be compatible with Angle, thus running on more platforms (e.g. anything not supporting Vulkan and DX12).

TODO:
- [x] make it work
- [x] remove `backends_map!` entirely
- [x] find a way to make the GLES backend optional

**Testing**
Tested on vange-rs


README.md
```diff
@@ -66,15 +66,25 @@ We have a [wiki](https://github.com/gfx-rs/wgpu/wiki) that serves as a knowledge
...
-  DX11   | :construction:                |                    |                    |
-  GLES3  |                               | :ok:               |                    |
-
-:white_check_mark: = First Class Support — :ok: = Best Effort Support — :construction: = Unsupported, but support in progress
+   API   |    Windows                    |  Linux & Android          |    macOS & iOS     |
+  -----  | ----------------------------- | ------------------------- | ------------------ |
+  Vulkan | :white_check_mark:            | :white_check_mark:        |                    |
+  Metal  |                               |                           | :white_check_mark: |
+  DX12   | :white_check_mark: (W10 only) |                           |                    |
+  DX11   | :hammer_and_wrench:           |                           |                    |
```
cwfitzgerald, line 74: Should probably be both the hammer and wrench and angle.

cwfitzgerald (approved):
LGTM!

README.md
kvark: it's tricky. Currently the table talks about wgpu backends, not the *actual* backends. They used to match, but with Angle they don't

cwfitzgerald: I think people end up using this table as "what wgpu can run on", so if we get there via angle or via a native backend it only matters to them in that they need to ship angle or not. So maybe let's have it be just the angle marker for now and once dx11 is back in place we can figure out what to do then.

kvark: I moved Angle to a separate row

PR #2258, opened 2021-12-05
support push constants on metal

### LINKED ISSUE #1825: Missing Features::PUSH_CONSTANTS for Metal
It looks like PUSH_CONSTANTS support accidentally got dropped in all the recent changes for the Metal driver.

I'll open a pull request adding it back. It seems it's working fine, and just needed the device to return that it was a supported feature.

molikto:
**Connections**

fix #1825
**Description**

support push constants on metal. needed to run https://github.com/googlefonts/compute-shader-101/tree/main/rust-gpu-toy on macOS

I have no experience working with graphics API, so I don't know if the code in the PR is actually correct :-(

**Testing**

That repo above seems to run correctly.

### DIFF
--- a/wgpu-hal/src/metal/command.rs
+++ b/wgpu-hal/src/metal/command.rs
@@ -592,7 +592,27 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
         _offset: u32,
         _data: &[u32],
     ) {
-        //TODO
+        if _stages.contains(wgt::ShaderStages::COMPUTE) {
+            self.state.compute.as_ref().unwrap().set_bytes(
+                _offset as _,
+                (_data.len() * WORD_SIZE) as u64,
+                _data.as_ptr() as _,
+            )
+        }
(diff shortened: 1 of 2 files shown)


wgpu-hal/src/metal/command.rs
```diff
@@ -592,7 +592,27 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
         _offset: u32,
         _data: &[u32],
     ) {
-        //TODO
+        if _stages.contains(wgt::ShaderStages::COMPUTE) {
```
kvark, line 595: This needs to be a tiny bit more complex.
The way it's supposed to work is if you have `[u8; max_push_constants]` array on the command encoder. `set_push_constants` would just overwrite some of these bytes, but the shaders would see as much as `PipelineLayout` have declared.

E.g. user could 0..4 with one `set_push_constants` and 4..8 with another call, and the shader needs to see all of them.

kvark (changes requested):

Issue #350, opened 2019-10-17
Setting vsync on Intel integrated has validation error


m4b:
when i switch to NoVsync i get this validation warning (it does appear however to not vsync anymore):

```
VALIDATION [VUID-VkSwapchainCreateInfoKHR-presentMode-01281 (0)] : vkCreateSwapchainKHR() called with a non-supported presentMode (i.e. VK_PRESENT_MODE_IMMEDIATE_KHR). The Vulkan spec states: presentMode must be one of the VkPresentModeKHR values returned by vkGetPhysicalDeviceSurfacePresentModesKHR for the surface (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkSwapchainCreateInfoKHR-presentMode-01281)
object info: (type: DEVICE, hndl: 94235621371968)
```

This is on archlinux with intel integrated card

m4b:
And this is on wgpu 0.3.3 crates io version

kvark:
Basically, we need to consider the available presentation modes when choosing one. In the worst case, we keep Vsync ON.

m4b:
So what is strange to me is how could it not be vsyncing now if my card is reporting it cannot be put in a NON vsync mode ?

kvark:
Found the relevant TODO: https://github.com/gfx-rs/wgpu/blob/78fbbba5e928e4a6f15a03a1c04d06916a947673/wgpu-native/src/swap_chain.rs#L98

kvark:
The supported present are obtained here: https://github.com/gfx-rs/wgpu/blob/78fbbba5e928e4a6f15a03a1c04d06916a947673/wgpu-native/src/device.rs#L1987

You can add a `info!` log to print these out and see.

paulkernfeld:
I think this was fixed by #424.

PR #2272, opened 2021-12-09
[WIP] Arcanization

kvark:
This is what's left from @pythonesque work.


John-Nagle:
Any progress on this? This is blocking [Rend3 #350, ](https://github.com/BVE-Reborn/rend3/issues/350), which is really slowing down my Second Life / Open Simulator renderer. That, being a modern multiverse system where content comes from the network, is constantly loading content in other threads while displaying the current scene. I need minimal slowdown in the refresh thread as this takes place.

kvark:
This is stale, in need of a champion.

John-Nagle:
A rename might help. Like "Fix concurrency performance bug".

John-Nagle:
Still a problem. As I said above, 20 FPS. Updating the scene from another thread kills rendering performance. 
Is there going to be action on this in the near future? I know it's hard, but the whole point of all this is to do high-performance rendering in Rust. I'm one of the few people doing really complex scenes, and I need all the performance I can get.

Watch this video. This is the kind of thing I need to render fast while new content is coming in from the server.

https://video.hardlimit.com/w/peBesyAgtzfRWS5FnDQQtn

jimblandy:
@John-Nagle This is absolutely the kind of thing that `wgpu` should be aiming to do well - no disagreement there. It's why we exist.

One thing you might be in a good position to do that would help enormously to move this forward is to pull together some benchmarks that exhibit the slowdowns you're concerned about. For performance work, it is almost always futile to stare at the code and guess how to make your users' code run faster. Having a realistic load to benchmark and instrument makes it possible to direct one's work effectively. These benchmarks would probably need to be freely redistributable, so that wgpu contributors could download them and work on them.

Is that something you might be able to help with?

jimblandy:
For example - we know that the hub RwLocks are contended - that's the point of this bug. But which ones are worth fixing first? I'll bet cleaning up `Adapter` ownership won't make any difference at all. Are there others that just don't matter?

And what if it's not just the `Hub` `RwLocks`? I'll bet my hat there are other bottlenecks. Having realistic loads will help us find all of them.

John-Nagle:
I have a big system that does a lot in parallel, but not a micro-benchmark. We need 1) something self-contained refreshing a scene with a lot of different objects, and 2) other threads busily adding and deleting objects, textures, and materials from the scene. I think Connor Fitzgerald might have a test. Something to mod. It's not a standard Rend3 example, though.

Does anyone have something I can work from? Something that generates test meshes, materials, and textures? Thanks.

kvark:
I suppose a potential benchmark can just be that: how many draw calls can be recorded in fixed time. Supposing 2 pipelines, 2 bind groups, and each draw call is just alternating between them, in order to avoid the state change being optimized out. We could write this down as a function of N threads and see if the number of draw calls scales up accordingly. Literally, as a criterion benchmark. We don't care about execution, even - just recording. The hypothesis is that it will not scale, because of the render pass end locks.

jimblandy:
> We could write this down as a function of N threads and see if the number of draw calls scales up accordingly.

It makes sense to check this specific scaling property. That's a good litmus test for "hub rwlocks are fixed".

But we also need to know how that one aspect of performance fits into everything else that goes on in a realistic load. It seems unlikely to me that the hub rwlocks are the only thing we need to know about, if we really want to support cases like the one shown in John-Nagle's video. This is why I want more realistic loads to work with - my guess is that *full* arcanization isn't necessary, and that arcanization alone isn't sufficient, to bring wgpu to the performance we want.

cwfitzgerald:
There are some tracy logs in [this rend3 issue](https://github.com/BVE-Reborn/rend3/issues/350#issuecomment-1032196846) of @John-Nagle's program specifically. I don't believe resource contention is the direct cause of the slowdown in his case (it is caused by tracking performance) but once that problem is solved it will start to show its face again.

It's a bit hard to tell a single culprit for locking from the trace, but it's mainly thing fightings against create_texture or write_textures on other threads.

As for arcinzation itself, we definitely should write a benchmark command lists encoding against each other, as it should be able to operate fully in parallel. This is a pretty common one people hit and has been the direct cause of us losing users.

Imberflur:
I may be able to find some time to contribute to this, although a good portion of that will be reviewing the previous discussions and the current diff to get an understanding of all the details.

John-Nagle:
Plugging away on my render-bench. Right now, one big brick cube appears. Soon, something more reasonable with content stats similar to the real scenes I've shown. This should make the complex scene case more testable.

John-Nagle:
As requested, I have constructed a benchmark/test for this situation. See

https://github.com/John-Nagle/render-bench

This creates and removes a large number of non-shared meshes and materials. Currently, 
the textures do not change. This is a first cut at simulating metaverse-type content, where, due to a large number of independent creators, meshes are rarely shared and there is little instancing.

The frame rate here is at 60FPS on the static scene, but drops to 13 FPS during mesh loading. 

slowupdate

This is built on Rend3, and uses code from the Rend3 examples.

cwfitzgerald:
@John-Nagle Thank you so much for this! The issues reproduce very clearly on my machine and gives me a really useful tracy trace. I will put this all together into a larget scale tracking issue detailing the various bottlenecks we have right now.

John-Nagle:
Thanks. It's good to hear that. Now this is an easily reproducible problem.

cwfitzgerald:
Considering our change of plans, and the formalization into an issue, I'm going to close this.

PR #2279, opened 2021-12-11
feat(deno_webgpu): Implement Surface API ops

DjDeveloperr:
**Description**
Implements only the ops (for now?) required to operate wgpu Surface API. This enables the projects which embed `deno_runtime` to add support for rendering through WebGPU on a window. Currently want to use in my [project](https://github.com/DjDeveloperr/deno_desktop), which at the moment implements these ops in there itself,
using a fork of Deno which is not fun to maintain.

These new ops are not exposed by default, instead, there is a new function exported from `deno_webgpu` called `init_surface_ext` which returns an `Extension` to register these ops.

**TODO**
- JS bindings: without JS bindings, using these ops is not possible without hacks including patching private properties of certain objects. [Look at this](https://github.com/DjDeveloperr/deno_desktop/blob/main/src/core.js). But how to expose these bindings in such a way that it is spec compliant is a challenge in itself because WebGPU API is very much tied to HTML canvas.

### DIFF
--- a/deno_webgpu/01_webgpu.js
+++ b/deno_webgpu/01_webgpu.js
@@ -1,6 +1,6 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
-// @ts-check
+// @ts-dont-check
 /// <reference path="../../core/lib.deno_core.d.ts" />
 /// <reference path="../web/internal.d.ts" />
 /// <reference path="../web/lib.deno_web.d.ts" />
@@ -291,7 +291,7 @@
         "op_webgpu_request_device",
         {
           adapterRid: this[_adapter].rid,
-          labe: descriptor.label,
+          label: descriptor.label,
           requiredFeatures,
           requiredLimits,
         },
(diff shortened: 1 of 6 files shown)


crowlKats:
Will take a look at this once i can.

DjDeveloperr:
I'm looking into possible ways to implement the JS bindings for this, as I mentioned in the PR description, it's not easy to use these ops without actual bindings in `deno_webgpu`. 

For exposing it, the spec says `GPUCanvasContext` is exposed in global scope. We do that and add a "private" method to it (non-spec-compliant) for creating this context without an actual canvas: something like `static [Symbol("[[create]]")](...): GPUCanvasContext | null`. This brings some more problems with it though... like what would `canvas` in `GPUCanvasContext` be? How would default size of configuration be determined?

Another solution would be to make a spec-compliant `OffscreenCanvas` for Deno but not implement any context for it for now, except the `webgpu` one. For making `OffscreenCanvas` to actually render on-screen, we could set a symbol `Symbol.for("[[windowRid]]")`, with rid for the DynHasRawWindowHandle resource (non-standard). What do you think? I would prefer this solution, as it also enables off-screen rendering use-cases for WebGPU.

I have some questions regarding implementing this API according to the WebGPU spec:

1. In `wgpu`, you have to call `instance.surface_present` in order to present the current texture on the window. But on browser WebGPU, there's no such method. It is presented on canvas automatically. How will we handle this?

2. From what I know, it seems `wgpu`'s `Surface` is actually `GPUCanvasContext`. In browser WebGPU, there is a `unconfigure` method on `GPUCanvasContext`. But in `wgpu` API I think there's no way to unconfigure a surface, but to only configure it using `instance.surface_configure`. How should this be handled?

3. `wgpu` lets you specify a `PresentMode` when configuring context. What value should I set for this? Since browser WebGPU does not have any such configuration.

Side question, how do I add new IDL types?

kvark:
On (2) we need to expose `unconfigure` here as well.

DjDeveloperr:
Further on (1), WebGPU canvas rendering is much tied to browser canvas. I'm confused how would we go about implementing this according to the spec. [`getCurrentTexture`](https://www.w3.org/TR/webgpu/#dom-gpucanvascontext-getcurrenttexture) returns the texture depending on current frame and it is presented every frame automatically, when browsers [update the rendering](https://html.spec.whatwg.org/multipage/webappapis.html#update-the-rendering). How would we identify that?

I also noticed that unlike WebGPU, wgpu does not support `colorSpace` and `compositingAlphaMode` in surface configuration.

lucacasonato:
@DjDeveloperr Do you have a little example of how this would be integrated into your deno_desktop project?

DjDeveloperr:
@lucacasonato I've just made a WIP PR that makes use of this. Currently I have upgraded deno_webgpu on my fork so the PR requires it. 

You should be able to try it out now:

```
git clone https://github.com/DjDeveloperr/deno_desktop --branch use-deno-webgpu-pr
cd deno_desktop
git clone https://github.com/DjDeveloperr/deno --branch upgrade-webgpu
cargo build
./target/debug/deno_desktop ./examples/hello_camera.js
```

DjDeveloperr:
Closing as stale...

crowlKats:
Not exactly stale, but as I have stated to you quite a few times, I currently don't have the time to look into this. I might have soon, but I cant promise anything.

PR #2283, opened 2021-12-12
Add back non-optional flags for wasm compilation

JackYoustra:
**Connections**

(None)

**Description**

Creating a project with wasm-pack and adding [gpgpu-rs](https://github.com/UpsettingBoy/gpgpu-rs/) as a dependency leads to a failed compilation due to a missing WriterFlags argument. 

**Testing**

It builds. I don't have much experience, so I'm not sure if it makes more sense to add a default argument or if hti sbreaks the non-wasm components, but it lets me build.

### DIFF
--- a/wgpu/src/backend/web.rs
+++ b/wgpu/src/backend/web.rs
@@ -1243,7 +1243,7 @@ impl crate::Context for Context {
                 );
                 let spv_module_info = validator.validate(&spv_module).unwrap();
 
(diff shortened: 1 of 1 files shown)


kvark:
Something is strange here. We do build `wgpu` on CI for WASM - https://github.com/gfx-rs/wgpu/runs/4494957516?check_suite_focus=true#step:9:1
and it's green. So why would it be failing for you?

JackYoustra:
Oops. Was failing for non-head. Removing

PR #2305, opened 2021-12-19
hal/dx12: Fix UMA check

### LINKED ISSUE #2304: AMD iGPUs are detected as DiscreteGpus for the DX12 backend.
**Description**
On systems running AMD iGPUs, the iGPU's DX12 backend is labelled as a DiscreteGpu, while the Vulkan backend is labelled as an IntegratedGpu. This can cause issues when requesting a high power adapter as it will choose the DX12 backend over the Vulkan backend on systems that only have an AMD iGPU.

**Repro steps**
Run any wgpu application that requests a high power adapter on a laptop with an AMD iGPU and no discrete GPU on Windows.

Expected: You get a Vulkan backend.
Observed: You get a DX12 backend.

**Extra materials**
https://gist.github.com/Elabajaba/0a4f6cd1202791f1a4ccfc544d130481
kvark's aha: https://matrix.to/#/!XFRnMvAfptAHthwBCx:matrix.org/$i9mLMKUQ4BP1YCSOM8kcOTlt55YxOIHCHAUOy6g_QRE

> so it's not reporting to be cache-coherent... interesting

**Platform**
Windows 10, tested on wgpu 0.11 and wgpu 0.12. Laptop with a Vega8 AMD iGPU and no discrete GPU.

kvark:
**Connections**
Fixes #2304

**Description**
Dunno why exactly, but possible AMD APU doesn't report to be coherent.

**Testing**
TODO

### DIFF
--- a/wgpu-hal/src/dx12/adapter.rs
+++ b/wgpu-hal/src/dx12/adapter.rs
@@ -98,7 +98,7 @@ impl super::Adapter {
             device_type: if (desc.Flags & dxgi::DXGI_ADAPTER_FLAG_SOFTWARE) != 0 {
                 workarounds.avoid_cpu_descriptor_overwrites = true;
                 wgt::DeviceType::Cpu
-            } else if features_architecture.CacheCoherentUMA != 0 {
+            } else if features_architecture.UMA != 0 {
                 wgt::DeviceType::IntegratedGpu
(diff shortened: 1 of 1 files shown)


kvark:
published in wgpu-hal-0.12.1

PR #2307, opened 2021-12-19
Do texture init via clear passes when possible

Wumpf:
**Connections**
Should fix #2149 - haven't confirmed since I don't a repro at hand.
Changes clear extension semantics a bit as discussed on #2174

**Description**
Previously, render target/depth textures couldn't be initialized at all if they not also were COPY_DST, violating the spec.
Now, we always will use render passes when possible since it's expected to be faster.

All this comes with an overhaul of the texture clearing mechanisms as we tried to optimize things based on the assumption of texture-init == copy zero_buffer:
* Textures now know their `TextureClearMode` which may bring with it pre-created render views for clearing
* I consolidated all texture clears to go through more similar code paths. In some cases this trades performance against maintainability of the code, but imho it looks like a better starting point for optimizations now

Other smaller but important things in this PR:
* Rename CLEAR_COMMANDS to CLEAR_TEXTURE and change semantics a bit (see doc changes)
* creation of 3D depth texture will cause an error now, see also https://github.com/gpuweb/gpuweb/issues/2304

**Testing**
Tests in tests/clear_texture.rs clear now depth/stencil formats as well (this was not possible previously)
Ran `cargo test` locally (win10)

PR #2308, opened 2021-12-19
[dx12] Fix partial texture barrier not affecting stencil aspect

Wumpf:
**Connections**
Issues accidentally covered by tests in #2307 (i.e. this change is required to land it)

**Description**
Fix partial texture barrier (i.e. when not the entire texture is transitioned) not affecting stencil aspect.
Fix clearing of D24Plus also clearing "hidden" stencil.

**Testing**
Tested in #2307, a dedicated test would be nice to be fair but this will have to do for now

### DIFF
--- a/wgpu-hal/src/dx12/command.rs
+++ b/wgpu-hal/src/dx12/command.rs
@@ -1,3 +1,5 @@
+use crate::FormatAspects;
+
 use super::{conv, HResult as _};
 use std::{mem, ops::Range, ptr};
 use winapi::um::d3d12;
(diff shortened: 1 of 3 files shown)


wgpu-hal/src/dx12/command.rs
```diff
@@ -319,15 +321,29 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
                     // Only one barrier if it affects the whole image.
                     self.temp.barriers.push(raw);
                 } else {
-                    // Generate barrier for each layer/level combination.
+                    let has_stencil = FormatAspects::from(barrier.texture.format)
+                        .contains(FormatAspects::STENCIL);
+                    let planes = if has_stencil {
+                        match barrier.range.aspect {
+                            wgt::TextureAspect::All => 0..2,
+                            wgt::TextureAspect::StencilOnly => 1..2,
```
kvark, line 329: what if it's a stencil-only format? would it still be 1..2?

wgpu-hal/src/dx12/command.rs
```diff
@@ -1,3 +1,5 @@
+use crate::FormatAspects;
```
kvark, line 1: let's not diverge from the current style where `crate::xxx` are either used directly, or included close to the place they are used

wgpu-hal/src/dx12/command.rs
```diff
@@ -610,7 +626,9 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
             if !ds.depth_ops.contains(crate::AttachmentOps::LOAD) {
                 flags |= native::ClearFlags::DEPTH;
             }
-            if !ds.stencil_ops.contains(crate::AttachmentOps::LOAD) {
+            if !ds.stencil_ops.contains(crate::AttachmentOps::LOAD)
+                && ds.target.view.has_stencil_aspect
```
kvark, line 630: similarly, if we are doing this check for stencil, we need to do a similar one for depth

kvark (changes requested):

wgpu-hal/src/dx12/command.rs
Wumpf: `use` and me really don't get along well 😄

wgpu-hal/src/dx12/command.rs
Wumpf: huh, yeah that should be 0..1 as well I suppose! Fixed!

kvark (approved):

kvark:
published in wgpu-hal-0.12.1

PR #2313, opened 2021-12-23
Update ash to 0.34

a1phyr:
**Connections**
`ash` changelog: https://github.com/MaikKlein/ash/releases/tag/0.34.0

**Description**
Update `ash` from `0.33` to `v0.34`.
`ash` can now link to vulkan at compile time (and does so by default), but this PR keeps the old behavior for now.

**Testing**
Ran tests and some examples with Vulkan backend.

### DIFF
--- a/wgpu-hal/Cargo.toml
+++ b/wgpu-hal/Cargo.toml
@@ -37,7 +37,7 @@ block = { version = "0.1", optional = true }
 foreign-types = { version = "0.3", optional = true }
 
 # backend: Vulkan
-ash = { version = "0.33", optional = true }
+ash = { version = "0.34", optional = true, default-features = false, features = ["debug", "loaded"] }
 gpu-alloc = { version = "0.5", optional = true }
 gpu-descriptor = { version = "0.2", optional = true }
 inplace_it = { version ="0.3.3", optional = true }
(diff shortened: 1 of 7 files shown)


wgpu-hal/Cargo.toml
```diff
@@ -37,7 +37,7 @@ block = { version = "0.1", optional = true }
 foreign-types = { version = "0.3", optional = true }
 
 # backend: Vulkan
-ash = { version = "0.33", optional = true }
+ash = { version = "0.34", optional = true, default-features = false, features = ["debug", "loaded"] }
```
cwfitzgerald, line 40: Are we able to get away without debug?

a1phyr: It would be difficult for little benefit:
- `wgpu_hal::Api` requires many associated types to implement `Debug`
- Because `ash` types wouldn't implement `Debug` we wouldn't be able to derive it so we would have to write manual impl
- `Debug` impls are used here and there for logging
- We would lose quite a bit of information

cwfitzgerald: Alright, that seems fine to me, if we're using them, we should keep them.

cwfitzgerald (approved):

kvark:
published in wgpu-hal-0.12.1

PR #2325, opened 2021-12-28
Clamp surface extent to within the supported range

### LINKED ISSUE #2286: Most of the samples have Vulkan swapchain validation errors
**Description**
Most of the samples have validation errors.

**Repro steps**
All of the samples that open a window and render show this swapchain error:

```
$ WGPU_POWER_PREF=high cargo run --example boids
    Finished dev [unoptimized + debuginfo] target(s) in 0.13s
     Running `target\debug\examples\boids.exe`
Using NVIDIA GeForce RTX 2060 (Vulkan)
[2021-12-13T14:33:24Z ERROR wgpu_hal::vulkan::instance] VALIDATION [VUID-VkSwapchainCreateInfoKHR-imageExtent-01274 (0x7cd0911d)]
        Validation Error: [ VUID-VkSwapchainCreateInfoKHR-imageExtent-01274 ] Object 0: handle = 0x1a999117898, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x7cd0911d | vkCreateSwapchainKHR() called with imageExtent = (1904,990), which is outside the bounds returned by vkGetPhysicalDeviceSurfaceCapabilitiesKHR(): currentExtent = (800,600), minImageExtent = (800,600), maxImageExtent = (800,600). The Vulkan spec states: imageExtent must be between minImageExtent and maxImageExtent, inclusive, where minImageExtent and maxImageExtent are members of the VkSurfaceCapabilitiesKHR structure returned by vkGetPhysicalDeviceSurfaceCapabilitiesKHR for the surface (https://vulkan.lunarg.com/doc/view/1.2.198.1/windows/1.2-extensions/vkspec.html#VUID-VkSwapchainCreateInfoKHR-imageExtent-01274)
[2021-12-13T14:33:24Z ERROR wgpu_hal::vulkan::instance]         objects: (type: DEVICE, hndl: 0x1a999117898, name: ?)
```

No validation errors

**Platform**
Windows 10

TheSpydog:
**Connections**
Fixes #2286

**Description**
When running any of the wgpu samples with the Vulkan backend (at least on Windows), they trip a validation error:
```
[2021-12-28T14:25:22Z ERROR wgpu_hal::vulkan::instance] VALIDATION [VUID-VkSwapchainCreateInfoKHR-imageExtent-01274 (0x7cd0911d)]
        Validation Error: [ VUID-VkSwapchainCreateInfoKHR-imageExtent-01274 ] Object 0: handle = 0x24223725ad0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x7cd0911d | vkCreateSwapchainKHR() called with imageExtent = (2858,1494), which is outside the bounds returned by vkGetPhysicalDeviceSurfaceCapabilitiesKHR(): currentExtent = (800,600), minImageExtent = (800,600), maxImageExtent = (800,600). The Vulkan spec states: imageExtent must be between minImageExtent and maxImageExtent, inclusive, where minImageExtent and maxImageExtent are members of the VkSurfaceCapabilitiesKHR structure returned by vkGetPhysicalDeviceSurfaceCapabilitiesKHR for the surface (https://vulkan.lunarg.com/doc/view/1.2.198.1/windows/1.2-extensions/vkspec.html#VUID-VkSwapchainCreateInfoKHR-imageExtent-01274)
```

This bizarre surface size request seems to be caused by a winit bug, but wgpu should handle this gracefully without violating the Vulkan spec.

To address this, I added a fallback path in `validate_surface_configuration` so that if the requested surface extents are outside of the valid range, then they will be clamped into the valid range.

**Testing**
The examples should now appear correctly without the Vulkan validation error above.
I also ran `cargo test` to be safe, although it did not trigger any validation errors in the first place.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -4558,19 +4558,25 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             config: &mut hal::SurfaceConfiguration,
             caps: &hal::SurfaceCapabilities,
         ) -> Result<(), E> {
-            let width = config.extent.width;
-            let height = config.extent.height;
-            if width < caps.extents.start().width
-                || width > caps.extents.end().width
-                || height < caps.extents.start().height
-                || height > caps.extents.end().height
+            let config_width = config.extent.width;
+            let config_height = config.extent.height;
+            let min_width = caps.extents.start().width;
+            let max_width = caps.extents.end().width;
+            let min_height = caps.extents.start().height;
+            let max_height = caps.extents.end().height;
+            if config_width < min_width
+                || config_width > max_width
+                || config_height < min_height
+                || config_height > max_height
             {
                 log::warn!(
-                    "Requested size {}x{} is outside of the supported range: {:?}",
-                    width,
-                    height,
+                    "Requested size {}x{} is outside of the supported range: {:?}, clamping size into supported range",
+                    config_width,
+                    config_height,
                     caps.extents
                 );
+                config.extent.width = config_width.clamp(min_width, max_width);
+                config.extent.height = config_height.clamp(min_height, max_height);
             }
             if !caps.present_modes.contains(&config.present_mode) {
                 log::warn!(
@@ -4588,7 +4594,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             if !caps.usage.contains(config.usage) {
                 return Err(E::UnsupportedUsage);
             }
-            if width == 0 || height == 0 {
+            if config.extent.width == 0 || config.extent.height == 0 {
                 return Err(E::ZeroArea);
             }
             Ok(())


kvark:
Thank you for the fix!
Unfortunately, we can't land it. The problem is not just the lack of validation.
If we start messing with the surface size without user knowing about it, any operations that depend on the size will fail.

For example, user may want to render to a surface with some depth testing. They initialize the surface and create a depth texture of the same size, then they record a render pass that uses both. Render pass will fail because the targets have different sizes, out of a sudden.

TheSpydog:
What should be done instead then? Should it return an error if the requested size is out of bounds? I considered that but that would mean all of the examples would panic on startup until the winit resizing bug is fixed.

kvark:
I think the only sane way is to have some mechanism of wgpu telling the user what the range of sizes is, and then validating if the requested size doesn't fit in.

TheSpydog:
> I think the only sane way is to have some mechanism of wgpu telling the user what the range of sizes is, and then validating if the requested size doesn't fit in.

Isn't that what it currently does? The code that I modified in this PR logs a debug message telling the user the acceptable range and that the new surface size is outside of those bounds, but then still resizes anyway. I'm not sure what you mean by "validating" here.

kvark:
No, it's not what's happening today, and not what this PR does. There is no mechanism for wgpu to tell the user what the min/max size of surfaces are. So no way to write correct code for the user that wouldn't fail.

TheSpydog:
> No, it's not what's happening today, and not what this PR does.

To be clear I'm referring to [this log statement here](https://github.com/gfx-rs/wgpu/blob/master/wgpu-core/src/device/mod.rs#L4678), which existed before this PR.

Regardless, it seems this is not an appropriate fix so I'll close the PR.

PR #2331, opened 2021-12-30
Improve detection and validation of cubemap views

### LINKED ISSUE #2323: Creating TextureDimension::D1 depth/stencil formats makes metal panic (and does not fail)
**Description**
Creating a 1 dimensional depth(/stencil) format does not cause a failure but makes metal panic:
```
-[MTLTextureDescriptorInternal validateWithDevice:]:1248: failed assertion `Texture Descriptor Validation
MTLTextureDescriptor: depth/stencil pixel formats require texture type of MTLTextureType2D[Array] or MTLTextureTypeCube[Array].
```

Similar to #2045, #2057

**Repro steps**
As described. Seen when running `clear_texture` test on Metal with a depth format after depth formats got enabled as in #2307

Should give a specialized error.
This is also lacking in the WebGPU spec I believe.

**Extra materials**
n/a

**Platform**
MacOS

kvark:
**Connections**
Loosely connected to https://github.com/gpuweb/gpuweb/pull/2463
Fixes  #2323

**Description**
Vulkan doesn't allow cubemaps to be multisampled. So here we are doing 2 things:
  1. adding relevant validation to the view creation
  2. reducing the set of cases where we tell Vulkan to support cubemaps

**Testing**
Untested

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -637,11 +637,14 @@ impl<A: HalApi> Device<A> {
     ) -> Result<resource::Texture<A>, resource::CreateTextureError> {
         let format_desc = desc.format.describe();
 
-        // Depth volume textures can't be written to - depth forbids COPY_DST and volume textures can't be rendered to - therefore they aren't allowed.
+        // Depth textures can only be 2D
         if format_desc.sample_type == wgt::TextureSampleType::Depth
-            && desc.dimension == wgt::TextureDimension::D3
+            && desc.dimension != wgt::TextureDimension::D2
         {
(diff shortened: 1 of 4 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -639,9 +639,12 @@ impl<A: HalApi> Device<A> {
 
         // Depth volume textures can't be written to - depth forbids COPY_DST and volume textures can't be rendered to - therefore they aren't allowed.
```
cwfitzgerald, line 640: Comment is out of date

cwfitzgerald (approved):
Nit but otherwise g2g

PR #2332, opened 2021-12-30
[v0.12] Fix profiling compile error under tracy and tracing

cwfitzgerald:
**Description**

FIxes a compile error under tracy and tracing that was uncaught by CI.

**Testing**

rend3


kvark:
published in wgpu-core-0.12.2

PR #2333, opened 2021-12-30
Fix trac(y/ing) compile issue

cwfitzgerald:
**Connections**

#2332 but for master and with a CI fix

**Description**

Added a CI flag to turn on `tracing`, profiling which seems to be a bit more restrictive of actual errors we will hit.

**Testing**

rend3


cwfitzgerald:
Why does this work on windows but not any other platform. I am confuse

cwfitzgerald:
Removed CI change, it just wasn't working. Very confusing.

kvark (approved):

PR #2336, opened 2021-12-30
Search for different versions of libwayland

### LINKED ISSUE #2335: Failed to initialize window when libwayland is not symllinked properly
**Description**
GL backend does not work on Wayland.

**Repro steps**
0. wgpu repository
1. `WGPU_BACKEND=gl cargo run --example cube` panics
2. `WAYLAND_DISPLAY= WGPU_BACKEND=gl cargo run --example cube` works

Expected: Both Wayland and X11 (Xwayland) work with both Vulkan and GLES
Got: Wayland only works with Vulkan, X11 is fine with both

Error: `[2021-12-30T12:23:13Z ERROR wgpu_core::device] surface configuration failed: incompatible window kind`

**Extra materials**
I'm going to investigate the issue on my own currently and submit a PR in case of successful fix.

**Platform**
GNOME Shell/Mutter 41.2 on Fedora 35
AMDGPU (Ryzen 4500U)
wgpu from `master` branch

sh7dm:
**Connections**
Fixes #2335

**Description**
Workaround for distributions that do not have symlinks for Wayland libraries.
If Wayland/EGL libraries are not found but system is running Wayland, we get a Wayland window handle from Winit, but the WSI that has been found is X11. This leads to error on systems that do not have versioned Wayland libraries symlinked to plain `.so` extension.

**Testing**
Tested on Fedora machine which I previously could not use `WGPU_BACKEND=gl`, now GL window is created successfully.

### DIFF
--- a/wgpu-hal/src/gles/egl.rs
+++ b/wgpu-hal/src/gles/egl.rs
@@ -114,20 +114,30 @@ fn open_x_display() -> Option<(ptr::NonNull<raw::c_void>, libloading::Library)> 
     }
 }
 
+unsafe fn find_library(paths: &[&str]) -> Option<libloading::Library> {
+    for path in paths {
+        match libloading::Library::new(path) {
+            Ok(lib) => return Some(lib),
+            _ => continue,
+        };
+    }
+    None
+}
+
 fn test_wayland_display() -> Option<libloading::Library> {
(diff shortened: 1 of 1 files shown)


kvark (approved):
Thank you!

kvark:
published in wgpu-hal-0.12.2

PR #2346, opened 2022-01-02
Replace run-wasm-example.sh with aliased rust crate (xtask)

rukai:
To use the new method simply do:
`cargo run-wasm --example cube --features webgl`

**Advantages over existing solution**
* cross platform
* 0 external dependencies - the existing solution depends on bash, a webserver and wasm_bindgen.
* better UX + more robust than anything hacked together with bash/powershell/bat
* wasm-bindgen-cli version is always in sync with wasm-bindgen version because `cargo update` updates both of them at the same time thanks to being in the same workspace

Currently uses a git dependency, but i'll release to crates.io when we are happy with the API.

**Testing**
I ran `cargo run-wasm --example cube --features webgl` etc for every example on linux.
I confirmed at least the cube example works on mac os and windows.

The documentation at https://github.com/gfx-rs/wgpu/wiki/Running-on-the-Web-with-WebGPU-and-WebGL will need to be changed after merging.

There is a PR in progress to use this same crate for running winit's web examples https://github.com/rust-windowing/winit/pull/2115


.cargo/config
```diff
@@ -0,0 +1,7 @@
+[alias]
+run-wasm = "run --release --package run-wasm --"
+
+[build]
+rustflags = [
+"--cfg=web_sys_unstable_apis"
```
kvark, line 6: does this affect all of the builds, even unrelated to WASM?

Cargo.lock
```diff
@@ -1280,6 +1396,15 @@ version = "0.6.25"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f497285884f3fcff424ffc933e56d7cbca511def0c9831a7f9b5f6153e3cc89b"
 
+[[package]]
```
kvark, line 1399: the dependencies for `run-wasm` largely are different from the rest of the project, so we should exclude it from the workspace, like we do with `deno_webgpu`

.cargo/config
```diff
@@ -0,0 +1,7 @@
+[alias]
+run-wasm = "run --release --package run-wasm --"
```
kvark, line 2: does this support `wasm32-unknown-unknown` only or the emscripten stuff as well?

rukai: just wasm32-unknown-unknown.
The existing script only supports wasm32-unknown-unknown as well.

Cargo.lock
rukai: We can do that.
However the downside would be that wasm-bindgen and wasm-bindgen-cli versions could get out of sync.
Because we have the Cargo.lock checked in this would be less likely but still relies on a manual review to catch this problem.

Possibly we could add a --no-serve flag to cargo-run-wasm to allow running run-wasm in CI to ensure the versions are compatible.

.cargo/config
rukai: It is set for all builds but I cant see a reason the flag would affect non wasm targets.
And this is actually good because having all targets use the same rustflags prevents cargo from trashing our build cache.

Cargo.lock
kvark: ok, that's one of the selling points of this PR, so let's keep it here

.cargo/config
kvark: #2233 could use essentially the same script but for emscripten target. I'm still unsure how much we really need emscripten support though.

kvark:
please rebase

.cargo/config
rukai: Alright, well let me know if I should implement emscripten support. I'm happy to do so if its needed by wgpu.

rukai:
I just noticed a potential problem with using wasm-bindgen-cli-support: https://github.com/rustwasm/wasm-bindgen/issues/2757
If it turns out to be a problem I'll figure out some solution to avoid cargo-run-wasm constantly breaking.

Issue #2297, opened 2021-12-16
halmark freezes


caiiiycuk:
The halmark test starts but show broken window and I can't close it (looks like it freezes). Maybe related to #2196 

```sh
cargo run --example halmark 
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/examples/halmark`
MESA-INTEL: warning: Performance support disabled, consider sysctl dev.i915.perf_stream_paranoid=0

Surface caps: Some(SurfaceCapabilities { formats: [Bgra8Unorm, Bgra8UnormSrgb], swap_chain_sizes: 4..=4294967295, current_extent: None, extents: Extent3d { width: 1, height: 1, depth_or_array_layers: 1 }..=Extent3d { width: 16384, height: 16384, depth_or_array_layers: 1 }, usage: COPY_SRC | COPY_DST | RESOURCE | COLOR_TARGET | STORAGE_READ | STORAGE_WRITE, present_modes: [Mailbox, Fifo], composite_alpha_modes: [Opaque, PreMultiplied] })
[2021-12-16T18:49:58Z ERROR wgpu_hal::vulkan::instance] VALIDATION [VUID-VkSwapchainCreateInfoKHR-minImageCount-01271 (0xe18476f3)]
        Validation Error: [ VUID-VkSwapchainCreateInfoKHR-minImageCount-01271 ] Object 0: handle = 0x55cf4cc6c280, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0xe18476f3 | vkCreateSwapchainKHR() called with minImageCount = 3, which is outside the bounds returned by vkGetPhysicalDeviceSurfaceCapabilitiesKHR() (i.e. minImageCount = 4, maxImageCount = 0). The Vulkan spec states: minImageCount must be greater than or equal to the value returned in the minImageCount member of the VkSurfaceCapabilitiesKHR structure returned by vkGetPhysicalDeviceSurfaceCapabilitiesKHR for the surface (https://github.com/KhronosGroup/Vulkan-Docs/search?q=)VUID-VkSwapchainCreateInfoKHR-minImageCount-01271)
[2021-12-16T18:49:58Z ERROR wgpu_hal::vulkan::instance]         objects: (type: DEVICE, hndl: 0x55cf4cc6c280, name: ?)
```

![изображение](https://user-images.githubusercontent.com/1727152/146431006-fa70a1ab-7906-4c27-89d0-49599d844336.png)

[...]

PR #2350, opened 2022-01-03
hal/gles: support externally initialized contexts

### LINKED ISSUE #2344: Pure GL context
Sometimes we want to integrate wgpu into a system that already created the window and GL context, and we want to draw something using wgpu's GLES context in addition to other stuff. In this case, EGL isn't involved.

Some sort of constructor for GLES adapter, used in conjunction with `create_adapter_from_raw`.
This also means working without any Instance, and without any EGL involved.
Would be great to make it not a compile time feature.

Separate cargo feature for "raw" mode.

**Additional context**
required for https://github.com/kvark/vange-rs/issues/172

kvark:
**Connections**
Closes #2344
cc @lpenguin @caiiiycuk

**Description**
This PR attempts to make minimal set of changes to be able to use GLES in a context we aren't creating.
Overall, I'm happy with the way it turned out, especially `struct EglContext` part.

**Testing**
Comes with a whole example!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,6 +24,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "android_glue"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "000444226fcff248f2bc4c7625be32c63caccfecc2723a2b9f78a7487a49c407"
+
+[[package]]
 name = "approx"
 version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


caiiiycuk:
Should I see something if I run halmark test? I only saw this:
изображение

caiiiycuk:
Have the same error as in #2297 , so problem is not related to winit.

https://user-images.githubusercontent.com/1727152/147933733-13935be2-25f3-4a22-9bbd-c77225ef1ce4.mp4

caiiiycuk:
glutin does not support wasm taget (https://github.com/rust-windowing/glutin/issues/1167),  so for wasm need to  create separate example in future.

kvark:
> Have the same error as in #2297

The examples in wgpu-hal aren't handling window resize, and it's not their goal to do so. This is irrelevant of the PR here.

> glutin does not support wasm taget (rust-windowing/glutin#1167), so for wasm need to create separate example in future.

we could put stuff in the same example, just not using glutin, via a bunch of `#[cfg()]`

cwfitzgerald (approved):

PR #2363, opened 2022-01-06
hal: limit binding sizes to i32

### LINKED ISSUE #2361: Incorrect max_storage_buffer_binding_size with silent failure
**Description**
On a device with a reported `max_storage_buffer_binding_size` of 4294967295 (2³² - 1), I should be able to allocate, bind, and access a storage buffer containing 1073741823 (⌊(2³² - 1) / 4⌋) `f32` values in a compute shader, without silent failures.

**Repro steps**
See attached trace files below.

*Expected*:

Allocation, binding, and access either succeed for buffers smaller than `device.limits().max_storage_buffer_binding_size` or fail with an error if an out-of-memory condition occurs.

*Observed*:

Allocation and binding are successful, however access fails silently for any storage buffer containing more than 536870911 (⌊(2³¹ - 1) / 4⌋) `f32` values. On the Vulkan backend, no entries are accessed successfully anymore for larger arrays (reduction kernel produces a value of -0.0). On the DX12 backend, entries beyond index 536870911 are accessed as zero (the reduction produces a value of 536870900 due to floating point rounding). No suspicious log outputs (except the `ReportLiveObjects()` warnings on termination that I always get in debug mode) are produced. Same if Vulkan validation layers are enforced via the Vulkan configurator.

Due to the point at which failure occurs, it seems an `i32` is involved at some point where it shouldn't. It is unclear if the issue is driver- or wgpu-related. Alternatively, it may be that the `max_storage_buffer_binding_size` is being misreported.

**Extra materials**
[trace_dx12.zip.zip](https://github.com/gfx-rs/wgpu/files/7823149/trace_dx12.zip.zip)
[trace_vulkan_2.zip.zip](https://github.com/gfx-rs/wgpu/files/7823150/trace_vulkan_2.zip.zip)
(Double zipped to reduce size, slightly less than 4 GiB each when uncompressed)

**Platform**
Windows 10 21H2
rustc 1.57.0
NVIDIA GeForce 3070 Laptop
NVIDIA GeForce 510.06 driver (WSLg preview release)
wgpu 9aac77883b40b6fb6bd2c6f8047b56d05f6d678a (`gecko`-branch)

kvark:
**Connections**
Closes #2361

**Description**
Driver compiler toolchain produce an offset as i32 internally, we can't expose bindings larger than that until we start doing 64-bit indexing.

**Testing**
Untested

### DIFF
--- a/wgpu-hal/src/auxil/mod.rs
+++ b/wgpu-hal/src/auxil/mod.rs
@@ -15,6 +15,12 @@ pub mod db {
     }
 }
 
+/// Maximum binding size for the shaders that only support `i32` indexing.
+/// Interestingly, the index itself can't reach that high, because the minimum
+/// element size is 4 bytes, but the compiler toolchain still computes the
+/// offset at some intermediate point, internally, as i32.
+pub const MAX_I32_BINDING_SIZE: u32 = 1 << 31;
+
 pub fn map_naga_stage(stage: naga::ShaderStage) -> wgt::ShaderStages {
     match stage {
         naga::ShaderStage::Vertex => wgt::ShaderStages::VERTEX,
--- a/wgpu-hal/src/dx12/adapter.rs
+++ b/wgpu-hal/src/dx12/adapter.rs
@@ -240,7 +240,7 @@ impl super::Adapter {
                     max_uniform_buffers_per_shader_stage: full_heap_count,
                     max_uniform_buffer_binding_size: d3d12::D3D12_REQ_CONSTANT_BUFFER_ELEMENT_COUNT
                         * 16,
-                    max_storage_buffer_binding_size: !0,
+                    max_storage_buffer_binding_size: crate::auxil::MAX_I32_BINDING_SIZE,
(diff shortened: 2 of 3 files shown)


wgpu-hal/src/auxil/mod.rs
```diff
@@ -15,6 +15,12 @@ pub mod db {
     }
 }
 
+/// Maximum binding size for the shaders that only support `i32` indexing.
+/// Interestingly, the index itself can't reach that high, because the minimum
+/// element size is 4 bytes, but the compiler toolchain still computes the
+/// offset at some intermediate point, internally, as i32.
+pub const MAX_I32_BINDING_SIZE: u32 = 1 << 31;
```
a1phyr, line 22: I may miss something, but `1 << 31` does not fit in an `i32`, as the maximum value for it is `(1 << 31) - 1`.

I think that using `i32::MAX as u32` instead could be less error-prone.

kvark: The size doesn't need to fit into i32. It's the byte offsets that are internally converted into i32, and they are strictly less than the full binding size.

PR #2364, opened 2022-01-06
add AddressMode::ClampToZero

laptou:
**Description**

On iOS, the "clamp to border" sampler address mode is not supported, however, the ["clamp to zero"](https://developer.apple.com/documentation/metal/mtlsampleraddressmode/mtlsampleraddressmodeclamptozero?language=objc) address mode does seem to be supported, so I would like to use that instead. I have added a new corresponding sampler address mode `wgpu::AddressMode::ClampToZero`. On the Metal backend I implement this using the native "clamp to zero" address mode, while I implement it using the "clamp to border" address mode on every other backend.

**Testing**
Test this change by verifying that:
- `wgpu::AddressMode::ClampToZero` works as expected on all platforms

### DIFF
--- a/wgpu-hal/src/dx12/device.rs
+++ b/wgpu-hal/src/dx12/device.rs
@@ -620,15 +620,7 @@ impl crate::Device<super::Api> for super::Device {
                 .anisotropy_clamp
                 .map_or(0, |_| d3d12::D3D12_FILTER_ANISOTROPIC);
 
-        let border_color = if desc
-            .address_modes
-            .iter()
-            .any(|am| am == &wgt::AddressMode::ClampToZero)
-        {
-            [0.0; 4]
-        } else {
-            conv::map_border_color(desc.border_color)
-        };
+        let border_color = conv::map_border_color(desc.border_color);
 
         self.raw.create_sampler(
             handle.raw,
--- a/wgpu-hal/src/dx12/device.rs
+++ b/wgpu-hal/src/dx12/device.rs
@@ -620,17 +620,15 @@ impl crate::Device<super::Api> for super::Device {
                 .anisotropy_clamp
                 .map_or(0, |_| d3d12::D3D12_FILTER_ANISOTROPIC);
 
-        let border_color = conv::map_border_color(desc.border_color).or_else(|| {
-            if desc
-                .address_modes
-                .iter()
-                .any(|am| am == &wgt::AddressMode::ClampToBorder)
-            {
-                Some([0.0; 4])
-            } else {
-                None
-            }
-        });
+        let border_color = if desc
+            .address_modes
+            .iter()
+            .any(|am| am == &wgt::AddressMode::ClampToZero)
+        {
+            [0.0; 4]
+        } else {
+            conv::map_border_color(desc.border_color)
+        };
 
         self.raw.create_sampler(
             handle.raw,
--- a/wgpu-hal/src/dx12/device.rs
+++ b/wgpu-hal/src/dx12/device.rs
@@ -620,6 +620,18 @@ impl crate::Device<super::Api> for super::Device {
                 .anisotropy_clamp
                 .map_or(0, |_| d3d12::D3D12_FILTER_ANISOTROPIC);
 
+        let border_color = conv::map_border_color(desc.border_color).or_else(|| {
+            if desc
+                .address_modes
+                .iter()
+                .any(|am| am == &wgt::AddressMode::ClampToBorder)
+            {
+                Some([0.0; 4])
+            } else {
+                None
+            }
+        });
+
         self.raw.create_sampler(
             handle.raw,
             filter,
@@ -631,7 +643,7 @@ impl crate::Device<super::Api> for super::Device {
             0.0,
             desc.anisotropy_clamp.map_or(0, |aniso| aniso.get() as u32),
             conv::map_comparison(desc.compare.unwrap_or(wgt::CompareFunction::Always)),
-            conv::map_border_color(desc.border_color),
+            border_color,
             desc.lod_clamp.clone().unwrap_or(0.0..16.0),
         );
 
--- a/wgpu-hal/src/gles/adapter.rs
+++ b/wgpu-hal/src/gles/adapter.rs
@@ -292,7 +292,7 @@ impl super::Adapter {
             extensions.contains("GL_EXT_texture_border_clamp"),
         );
         features.set(
-            wgt::Features::SAMPLER_BORDER_COLOR_AUTO_BLACK,
+            wgt::Features::ADDRESS_MODE_CLAMP_TO_ZERO,
             extensions.contains("GL_EXT_texture_border_clamp"),
         );
         features.set(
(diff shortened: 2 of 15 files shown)


laptou:
I could also have implemented this by trying to use ClampToBorder on the Metal backend, and then falling back to ClampToZero if it is not supported and if the border colour is transparent black. However, if the border colour is not transparent black, then it seems like we would have to panic, so I'm not sure that that design would be better.

kvark:
Thank you for filing this! I didn't know about clamp-to-zero being a separate mode on Metal.
Changes like this should go through https://github.com/gpuweb/gpuweb first, unless you are adding a native-only extension.

There is a problem with your proposal though. Clamp-to-zero returns something that depends on the texture format:
> Out-of-range texture coordinates return transparent zero (0,0,0,0) for images with an alpha channel and return opaque zero (0,0,0,1) for images without an alpha channel.

This is different from the clamp-to-border semantics that other backends expose. So it doesn't look like we can implement this portably.

laptou:
What should I do? I would love to be able to use a clamping address mode on all platforms that I'm targeting (which includes iOS), and if I can't then my shaders will have to become more complicated (and probably slower) because I would have to introduce bounds checks for my texture coordinates.

> Changes like this should go through gpuweb/gpuweb first, unless you are adding a native-only extension.

Since I am not targeting the web with my application, I'm going to say that this is a native-only extension 😅 someone else can deal with getting this past the W3C if they need it on the web

kvark:
Ok, let's make it a native-only feature then. We'll have to document the behavior guarantees of Metal, in a similar way you suggested.
It's supported 100% on native, but it has to be a feature because the Web doesn't expose this, and we want applications on wgpu-rs to be targeting the Web seamlessly.

laptou:
> it has to be a feature

What do you mean by this? Do you mean that I should put `AddressMode::ClampToZero` behind a Cargo feature flag? Or what do I need to do for this change to move forward?

kvark:
I mean a feature in WebGPU sense, it needs to be a flag in `wgt::Features`, similar to how clamp-to-border is exposed

laptou:
Oh, I already implemented that ([see here](https://github.com/gfx-rs/wgpu/pull/2364/files#diff-755ab512580335f8c40050a2825d5f44e875385eeab83a9d5f02543f53d714b9R554-R564)). The feature is enabled for any adapter where clamp-to-border is enabled, except for the Metal adapter where it is always enabled.

wgpu-hal/src/dx12/device.rs
```diff
@@ -620,6 +620,16 @@ impl crate::Device<super::Api> for super::Device {
...
                 .map_or(0, |_| d3d12::D3D12_FILTER_ANISOTROPIC);
 
+        let border_color = if desc
+            .address_modes
+            .iter()
+            .any(|am| am == &wgt::AddressMode::ClampToZero)
+        {
+            [0.0; 4]
+        } else {
+            conv::map_border_color(desc.border_color)
```
kvark, line 630: oh snap! Actually we have a problem here.
What happens if a sampler has both `ClampToZero` and `ClampToBorder` modes? We can't set the border color to zero in this case. It's a conceptual problem, and I'm not seeing an easy solution

kvark (changes requested):
Unfortunately, we are hitting another problem here.
I think a solution would require this new behavior to be a part of `enum SamplerBorderColor` enum (e.g. `AutoBlack`) instead of the sampler addressing mode enum. This would ensure there is never a conflict between clamping to zero and clamping to other borders.

wgpu-hal/src/dx12/device.rs
laptou: I think the only solution is to declare that this is not allowed and panic

laptou: > I think a solution would require this new behavior to be a part of enum SamplerBorderColor enum (e.g. AutoBlack)

Actually, I think this would work, but it's not too different from just writing a workaround in the Metal backend where it falls back to using ClampToZero if ClampToBorder is not supported. What do you think of this approach vs the one you descibed?

kvark: Please clarify your suggestion. Are you saying, let's expose this feature on 100% metal, and just document that the border color may be ignored on platforms that don't support it? I'm concerned that this would make it problematic for people who need a specific color, and they'd lose an ability to discover if it's supported. Frankly, I don't know right off the bat who would need that.

laptou: You're right that not having the ability to check whether it's supported is not ideal. So then we'd have to go back to my initial suggestion, which is to declare that using `ClampToZero` and `ClampToBorder` at the same time is a validation error.

kvark: I still think it's better to have this as an enum value for `SamplerBorderColor`, which would encode this validation rule by construction.

laptou: Alright, I'll do that.

laptou:
Okay, I implemented it the way you suggested, but I just ran into a problem with this design when trying to test it on iOS: AddressMode::ClampToBorder is not allowed on iOS with the current settings, so we would need to put in a special case where it checks the border color and panics if we're on iOS and the border colour is not `AutoBlack`.

kvark:
> Okay, I implemented it the way you suggested, but I just ran into a problem with this design when trying to test it on iOS: AddressMode::ClampToBorder is not allowed on iOS with the current settings, so we would need to put in a special case where it checks the border color and panics if we're on iOS and the border colour is not AutoBlack.

The way validation would work then is the following:
  - if `ADDRESS_MODE_CLAMP_TO_ZERO` is enabled (your new feature), then it's allowed to use `border_color = Some(Zero)` (or `AutoBlack`, but I think `Zero` is closer to the feature name)
  - if `ADDRESS_MODE_CLAMP_TO_BORDER` is enabled, then it's allowed to use `border_color = Some(other border colors)`.
  - otherwise, `border_color` has to be `None`.

laptou:
done @kvark

wgpu-hal/src/metal/conv.rs
```diff
@@ -73,6 +73,7 @@ pub fn map_border_color(border_color: wgt::SamplerBorderColor) -> mtl::MTLSample
         wgt::SamplerBorderColor::TransparentBlack => TransparentBlack,
         wgt::SamplerBorderColor::OpaqueBlack => OpaqueBlack,
         wgt::SamplerBorderColor::OpaqueWhite => OpaqueWhite,
+        _ => unreachable!()
```
kvark, line 76: let's list this variant explicitly

wgpu-hal/src/gles/adapter.rs
```diff
@@ -291,6 +291,10 @@ impl super::Adapter {
             wgt::Features::ADDRESS_MODE_CLAMP_TO_BORDER,
             extensions.contains("GL_EXT_texture_border_clamp"),
         );
+        features.set(
+            wgt::Features::ADDRESS_MODE_CLAMP_TO_ZERO,
```
kvark, line 295: let's combine it with the case above? Just put the `wgt::Features::ADDRESS_MODE_CLAMP_TO_BORDER | wgt::Features::ADDRESS_MODE_CLAMP_TO_ZERO`

kvark (approved):
Very nice!

laptou:
done

kvark (approved):

PR #2365, opened 2022-01-07
Add `util::indirect::*` helper structs

IcanDivideBy0:
**Connections**
n/a

**Description**
I think most people using indirect draw calls will end up rewriting those structs just as they are described in the doc, make it simpler to have them implemented right here

**Testing**
tested locally, works as expected

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -2605,18 +2605,7 @@ impl<'a> RenderPass<'a> {
     ///
     /// The active vertex buffers can be set with [`RenderPass::set_vertex_buffer`].
     ///
-    /// The structure expected in `indirect_buffer` is the following:
-    ///
-    /// ```rust
-    /// #[repr(C)]
-    /// struct DrawIndirect {
-    ///     vertex_count: u32, // The number of vertices to draw.
-    ///     instance_count: u32, // The number of instances to draw.
-    ///     first_vertex: u32, // The Index of the first vertex to draw.
-    ///     first_instance: u32, // The instance ID of the first instance to draw.
-    ///     // has to be 0, unless [`Features::INDIRECT_FIRST_INSTANCE`] is enabled.
-    /// }
-    /// ```
+    /// The structure expected in `indirect_buffer` must conform to [`DrawIndirect`](crate::util::DrawIndirect).
     pub fn draw_indirect(&mut self, indirect_buffer: &'a Buffer, indirect_offset: BufferAddress) {
         self.id.draw_indirect(&indirect_buffer.id, indirect_offset);
     }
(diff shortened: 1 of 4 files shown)


wgpu/src/util/indirect.rs
```diff
@@ -0,0 +1,55 @@
...
+    /// Has to be 0, unless [`Features::INDIRECT_FIRST_INSTANCE`](crate::Features::INDIRECT_FIRST_INSTANCE) is enabled.
+    pub base_instance: u32,
+}
+
+impl DrawIndexedIndirect {
+    /// Returns the bytes representation of the struct, ready to be written in a [`Buffer`](crate::Buffer).
+    pub fn as_bytes(&self) -> &[u8] {
+        unsafe {
+            std::mem::transmute(std::slice::from_raw_parts(
+                self as *const _ as *const u8,
```
cwfitzgerald, line 50: may I suggest: `<*const u8>::cast(self)`

wgpu/src/lib.rs
```diff
@@ -2605,18 +2605,7 @@ impl<'a> RenderPass<'a> {
...
-    /// #[repr(C)]
-    /// struct DrawIndirect {
-    ///     vertex_count: u32, // The number of vertices to draw.
-    ///     instance_count: u32, // The number of instances to draw.
-    ///     first_vertex: u32, // The Index of the first vertex to draw.
-    ///     first_instance: u32, // The instance ID of the first instance to draw.
-    ///     // has to be 0, unless [`Features::INDIRECT_FIRST_INSTANCE`] is enabled.
-    /// }
-    /// ```
+    /// The structure expected in `indirect_buffer` is a [`DrawIndirect`](crate::util::DrawIndirect).
```
cwfitzgerald, line 2608: I'm a bit worried about this pedagogically that it might make it harder to figure out what is going on.

Otoh, people using indirect rendering generally should know what's going on.

cwfitzgerald (approved):

wgpu/src/lib.rs
IcanDivideBy0: What do you think of:
> The structure expected in indirect_buffer must conform to DrawIndirect.
?

wgpu/src/util/indirect.rs
IcanDivideBy0: I don't like it mush either, but `<*const u8>::cast(self)` gives me:
```
mismatched types

expected `u8`, found struct `util::indirect::DrawIndirect`

note: expected raw pointer `*const u8`
           found reference `&util::indirect::DrawIndirect`
```

kvark (approved):
Looks great. Would you mind adding `DispatchIndirect` as well?

IcanDivideBy0:
> Looks great. Would you mind adding `DispatchIndirect` as well?

sure, i've added it

kvark (approved):
thank you!

wgpu/src/lib.rs
kvark: yes, they should probably know :) I think it's ok, we can come back to this if needed. Resolving to make CI happy

PR #2377, opened 2022-01-11
Texture format MSAA capabilities

kvark:
**Connections**
Matches https://github.com/gpuweb/gpuweb/pull/2463

**Description**
Turned out a bit more involved, but hopefully right. Some of the lines are just `rustfmt` deciding to split.

MSAA and resolves are not guaranteed for all formats. This PR exposes it via our `TextureFormatFeatureFlags`, and also converts "filterable" to such flag for consistency.

**Testing**
MSAA example.


wgpu-types/src/lib.rs
```diff
@@ -1984,9 +1996,18 @@ impl TextureFormat {
         let float = TextureSampleType::Float { filterable: true };
         let depth = TextureSampleType::Depth;
 
-        // Color spaces
-        let linear = false;
-        let srgb = true;
+        enum ColorSpace {
```
cwfitzgerald, line 1999: Why not expose this, when we get support for certain hdr color spaces, it will be useful

cwfitzgerald (approved):
Small comment, but otherwise lgtm

wgpu-types/src/lib.rs
kvark: I haven't worked with HDR yet, so didn't consider the way it's to be exported. If it's to be added to this enum - then totally let's expose this as an enum. Maybe in a follow-up?

kvark:
Thanks for the quick review!

PR #2382, opened 2022-01-11
hal/vk: use stencil read/write masks

### LINKED ISSUE #2380: Stencil Acts Incorrectly On Some Systems
**Description**
I've been having some trouble with stencilling. Things seem to work fine on MacOS, but not so much on Win10 and Ubuntu. The stencil itself works fine, but what's drawn afterwards is muddled somehow. It's almost as if the stencil carries on with subsequent draw calls, denying future calls the ability to draw in previously used space.

**Repro steps**
I've created a single file program for this which you can find in the wgpu/examples folder of [this fork](https://github.com/metasophiea/wgpu).
You can run the [example](https://github.com/metasophiea/wgpu/tree/master/wgpu/examples/hello-stencil) with "cargo -r --example hello-stencil

Here we see three triangles being rendered to the screen
no-stencil
Now a fourth stencilling triangle is added, which works correctly
screenshot
On Ubuntu however, we see the incorrect behaviour
Screenshot 2022-01-11 at 16 01 09

kvark:
**Connections**
Fixes #2380

**Description**
We didn't use the stencil masks.

**Testing**
hello-stencil example

### DIFF
--- a/wgpu-hal/src/vulkan/conv.rs
+++ b/wgpu-hal/src/vulkan/conv.rs
@@ -698,14 +698,18 @@ pub fn map_stencil_op(op: wgt::StencilOperation) -> vk::StencilOp {
     }
 }
 
-pub fn map_stencil_face(face: &wgt::StencilFaceState) -> vk::StencilOpState {
+pub fn map_stencil_face(
+    face: &wgt::StencilFaceState,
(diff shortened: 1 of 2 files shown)


kvark:
published in wgpu-hal-0.12.3

PR #2400, opened 2022-01-17
hal/gl: Allow push constants trough emulation

JCapucho:
**Connections**
Depends on https://github.com/gfx-rs/naga/pull/1672 so bumps `naga` revision to pick it up

**Description**
Push constants are a great way to pass small amounts of data to the gpu between draw calls.

But maintaining two code paths for backends that support push constants and those who don't can be prohibitive to picking them.

So a good middle ground is to emulate push constants in the secondary backends, this is done by using a uniform buffer for push constants.

*Implementation notes*:

The buffer is always kept mapped and when a `set_push_constants` call is issued the mapping is written to.

Push constants stages are ignored.

The uniform buffer uses the binding 0 in the shader.

**TODOS**:

- [ ] The code currently uses the pointer returned from `glMapBufferRange` as a `*mut u32` for the slice, but according to https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glMapBufferRange.xhtml alignment is not guaranteed so it might need to be changed to a `*mut u8` and `set_push_constants` will need to convert the slice from `u32` to `u8`
- [ ] Another problem is synchronization, `glMapBufferRange` doesn't give much information but has an `GL_MAP_UNSYNCHRONIZED_BIT` so it leads me to believe that synchronization is implicit, but if someone who knows opengl better could confirm it would be helpful

**Testing**
I tested on a modified cube example
```patch
diff --git a/wgpu/examples/cube/main.rs b/wgpu/examples/cube/main.rs
index 60a49542..556ec979 100644
--- a/wgpu/examples/cube/main.rs
+++ b/wgpu/examples/cube/main.rs
@@ -124,10 +124,21 @@ impl Example {
 }
 
 impl framework::Example for Example {
+    fn required_features() -> wgt::Features {
+        wgt::Features::PUSH_CONSTANTS
+    }
+
     fn optional_features() -> wgt::Features {
         wgt::Features::POLYGON_MODE_LINE
     }
 
+    fn required_limits() -> wgpu::Limits {
+        wgpu::Limits {
+            max_push_constant_size: 64,
+            ..wgpu::Limits::downlevel_webgl2_defaults()
+        }
+    }
+
     fn init(
         config: &wgpu::SurfaceConfiguration,
         _adapter: &wgpu::Adapter,
@@ -179,7 +190,10 @@ impl framework::Example for Example {
         let pipeline_layout = device.create_pipeline_layout(&wgpu::PipelineLayoutDescriptor {
             label: None,
             bind_group_layouts: &[&bind_group_layout],
-            push_constant_ranges: &[],
+            push_constant_ranges: &[wgpu::PushConstantRange {
+                stages: wgpu::ShaderStages::VERTEX,
+                range: 0..4,
+            }],
         });
 
         // Create the texture
@@ -380,6 +394,7 @@ impl framework::Example for Example {
             rpass.set_bind_group(0, &self.bind_group, &[]);
             rpass.set_index_buffer(self.index_buf.slice(..), wgpu::IndexFormat::Uint16);
             rpass.set_vertex_buffer(0, self.vertex_buf.slice(..));
+            rpass.set_push_constants(wgpu::ShaderStages::VERTEX, 0, &(0.2_f32).to_ne_bytes());
             rpass.pop_debug_group();
             rpass.insert_debug_marker("Draw!");
             rpass.draw_indexed(0..self.index_count as u32, 0, 0..1);
diff --git a/wgpu/examples/cube/shader.wgsl b/wgpu/examples/cube/shader.wgsl
index 4bd4305a..ba8423b4 100644
--- a/wgpu/examples/cube/shader.wgsl
+++ b/wgpu/examples/cube/shader.wgsl
@@ -23,11 +23,16 @@ fn vs_main(
 [[group(0), binding(1)]]
 var r_color: texture_2d<u32>;
 
+struct PushConstants {
+    mul: f32;
+};
+var<push_constant> pc: PushConstants;
+
 [[stage(fragment)]]
 fn fs_main(in: VertexOutput) -> [[location(0)]] vec4<f32> {
     let tex = textureLoad(r_color, vec2<i32>(in.tex_coord * 256.0), 0);
     let v = f32(tex.x) / 255.0;
-    return vec4<f32>(1.0 - (v * 5.0), 1.0 - (v * 15.0), 1.0 - (v * 50.0), 1.0);
+    return pc.mul * vec4<f32>(1.0 - (v * 5.0), 1.0 - (v * 15.0), 1.0 - (v * 50.0), 1.0);
 }
 
 [[stage(fragment)]]
```

As well as a modified branch of https://gitlab.com/veloren/veloren

**Note**: I don't have much experience with openGL so the code might not be the best.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,6 @@ default-members = ["wgpu", "wgpu-hal", "wgpu-info"]
 
 [patch."https://github.com/gfx-rs/naga"]
 #naga = { path = "../naga" }
-naga = { git = "https://github.com/JCapucho/naga", branch = "glsl-out-push-constants-v2" }
 
 [patch."https://github.com/zakarumych/gpu-descriptor"]
 #gpu-descriptor = { path = "../gpu-descriptor/gpu-descriptor" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ default-members = ["wgpu", "wgpu-hal", "wgpu-info"]
 
 [patch."https://github.com/gfx-rs/naga"]
 #naga = { path = "../naga" }
+naga = { git = "https://github.com/JCapucho/naga", branch = "glsl-out-push-constants-v2" }
 
 [patch."https://github.com/zakarumych/gpu-descriptor"]
 #gpu-descriptor = { path = "../gpu-descriptor/gpu-descriptor" }
--- a/wgpu-hal/src/gles/command.rs
+++ b/wgpu-hal/src/gles/command.rs
@@ -28,6 +28,7 @@ pub(super) struct State {
     has_pass_label: bool,
     instance_vbuf_mask: usize,
     dirty_vbuf_mask: usize,
+    push_offset_to_uniform: ArrayVec<super::UniformDesc, { super::MAX_PUSH_CONSTANTS }>,
 }
 
 impl super::CommandBuffer {
@@ -42,6 +43,21 @@ impl super::CommandBuffer {
         let start = self.data_bytes.len() as u32;
         self.data_bytes.extend(marker.as_bytes());
         start..self.data_bytes.len() as u32
+    }
+
+    fn add_push_constant_data(&mut self, data: &[u32]) -> Range<u32> {
+        let data_raw = unsafe {
+            std::slice::from_raw_parts(
+                data.as_ptr() as *const _,
+                data.len() * mem::size_of::<u32>(),
+            )
+        };
+        let start = self.data_bytes.len();
+        assert!(start < u32::MAX as usize);
+        self.data_bytes.extend_from_slice(data_raw);
+        let end = self.data_bytes.len();
+        assert!(end < u32::MAX as usize);
+        (start as u32)..(end as u32)
     }
 }
 
@@ -148,8 +164,10 @@ impl super::CommandEncoder {
     fn set_pipeline_inner(&mut self, inner: &super::PipelineInner) {
         self.cmd_buffer.commands.push(C::SetProgram(inner.program));
 
-        //TODO: push constants
-        let _ = &inner.uniforms;
+        self.state.push_offset_to_uniform.clear();
+        self.state
+            .push_offset_to_uniform
+            .extend(inner.uniforms.iter().cloned());
 
         // rebind textures, if needed
         let mut dirty_textures = 0u32;
@@ -603,10 +621,25 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
         &mut self,
         _layout: &super::PipelineLayout,
         _stages: wgt::ShaderStages,
-        _offset: u32,
-        _data: &[u32],
+        start_offset: u32,
+        data: &[u32],
     ) {
-        unimplemented!()
+        let range = self.cmd_buffer.add_push_constant_data(data);
+
+        let end = start_offset + data.len() as u32 * 4;
+        let mut offset = start_offset;
+        while offset < end {
+            let uniform = self.state.push_offset_to_uniform[offset as usize / 4].clone();
+            let size = uniform.size;
+            if uniform.location.is_none() {
+                panic!("No uniform for push constant");
+            }
+            self.cmd_buffer.commands.push(C::SetPushConstants {
+                uniform,
+                offset: range.start + offset,
+            });
+            offset += size;
+        }
     }
 
     unsafe fn insert_debug_marker(&mut self, label: &str) {
--- a/wgpu-hal/src/gles/device.rs
+++ b/wgpu-hal/src/gles/device.rs
@@ -272,30 +272,35 @@ impl super::Device {
             }
         }
 
-        let uniforms = {
-            let count = gl.get_active_uniforms(program);
-            let mut offset = 0;
-            let mut uniforms = Vec::new();
+        let mut uniforms: [super::UniformDesc; super::MAX_PUSH_CONSTANTS] = Default::default();
+        let count = gl.get_active_uniforms(program);
+        let mut offset = 0;
 
-            for uniform in 0..count {
-                let glow::ActiveUniform { size, utype, name } =
-                    gl.get_active_uniform(program, uniform).unwrap();
+        for uniform in 0..count {
+            let glow::ActiveUniform { utype, name, .. } =
+                gl.get_active_uniform(program, uniform).unwrap();
 
-                if let Some(location) = gl.get_uniform_location(program, &name) {
-                    // Sampler2D won't show up in UniformLocation and the only other uniforms
-                    // should be push constants
-                    uniforms.push(super::UniformDesc {
-                        location,
-                        offset,
-                        utype,
-                    });
-
-                    offset += size as u32;
-                }
+            if conv::is_sampler(utype) {
+                continue;
             }
 
-            uniforms.into_boxed_slice()
-        };
+            if let Some(location) = gl.get_uniform_location(program, &name) {
+                if uniforms[offset / 4].location.is_some() {
+                    panic!("Offset already occupied")
+                }
+
+                // `size` will always be 1 so we need to guess the real size from the type
+                let uniform_size = conv::uniform_byte_size(utype);
+
+                uniforms[offset / 4] = super::UniformDesc {
+                    location: Some(location),
+                    size: uniform_size,
+                    utype,
+                };
+
+                offset += uniform_size as usize;
+            }
+        }
 
         Ok(super::PipelineInner {
             program,
@@ -858,7 +863,6 @@ impl crate::Device<super::Api> for super::Device {
                 version: self.shared.shading_language_version,
                 writer_flags,
                 binding_map,
-                push_constant_binding: 0, //TODO?
             },
         })
     }
--- a/wgpu-hal/src/gles/mod.rs
+++ b/wgpu-hal/src/gles/mod.rs
@@ -88,6 +88,7 @@ const MAX_TEXTURE_SLOTS: usize = 16;
 const MAX_SAMPLERS: usize = 16;
 const MAX_VERTEX_ATTRIBUTES: usize = 16;
 const ZERO_BUFFER_SIZE: usize = 256 << 10;
+const MAX_PUSH_CONSTANTS: usize = 16;
 
 impl crate::Api for Api {
     type Instance = Instance;
@@ -392,13 +393,18 @@ struct VertexBufferDesc {
     stride: u32,
 }
 
-#[allow(unused)]
-#[derive(Clone)]
+#[derive(Clone, Debug, Default)]
 struct UniformDesc {
-    location: glow::UniformLocation,
-    offset: u32,
+    location: Option<glow::UniformLocation>,
+    size: u32,
     utype: u32,
 }
+
+// Safe: WASM doesn't have threads
+#[cfg(target_arch = "wasm32")]
+unsafe impl Sync for UniformDesc {}
+#[cfg(target_arch = "wasm32")]
+unsafe impl Send for UniformDesc {}
 
 /// For each texture in the pipeline layout, store the index of the only
 /// sampler (in this layout) that the texture is used with.
@@ -407,7 +413,7 @@ type SamplerBindMap = [Option<u8>; MAX_TEXTURE_SLOTS];
 struct PipelineInner {
     program: glow::Program,
     sampler_map: SamplerBindMap,
-    uniforms: Box<[UniformDesc]>,
+    uniforms: [UniformDesc; MAX_PUSH_CONSTANTS],
 }
 
 #[derive(Clone, Debug)]
@@ -716,6 +722,11 @@ enum Command {
     InsertDebugMarker(Range<u32>),
     PushDebugGroup(Range<u32>),
     PopDebugGroup,
+    SetPushConstants {
+        uniform: UniformDesc,
+        /// Offset from the start of the `data_bytes`
+        offset: u32,
+    },
 }
 
 #[derive(Default)]
--- a/wgpu-hal/src/gles/queue.rs
+++ b/wgpu-hal/src/gles/queue.rs
@@ -1067,6 +1067,70 @@ impl super::Queue {
                 #[cfg(not(target_arch = "wasm32"))]
                 gl.pop_debug_group();
             }
+            C::SetPushConstants {
+                ref uniform,
+                offset,
+            } => {
+                fn get_data<T>(data: &[u8], offset: u32) -> &[T] {
+                    let raw = &data[(offset as usize)..];
+                    unsafe {
+                        slice::from_raw_parts(
+                            raw.as_ptr() as *const _,
+                            raw.len() / mem::size_of::<T>(),
+                        )
+                    }
+                }
+
+                let location = uniform.location.as_ref();
+
+                match uniform.utype {
+                    glow::FLOAT => {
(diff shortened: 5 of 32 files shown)


kvark (changes requested):
This is a bit unexpected. What I thought we would be doing is - using freestanding uniforms for push constants. That's what gfx-hal did, IIRC. But instead this PR tries to do this via a uniform buffer.

The current approach isn't going to work from the point of view of synchronization between GPU and CPU. Within the same submission, multiple different draw calls would want to overwrite the mapped buffer contents, bind it, and use it on GPU. This means 100% data race between CPU and GPU.

JCapucho:
> This is a bit unexpected. What I thought we would be doing is - using freestanding uniforms for push constants. That's what gfx-hal did, IIRC. But instead this PR tries to do this via a uniform buffer.

Should have probably checked if `gfx-hal` already had this implemented

> The current approach isn't going to work from the point of view of synchronization between GPU and CPU. Within the same submission, multiple different draw calls would want to overwrite the mapped buffer contents, bind it, and use it on GPU. This means 100% data race between CPU and GPU.

Could you elaborate from what I read from the OpenGL wiki about synchronization, https://www.khronos.org/opengl/wiki/Synchronization#Implicit_synchronization

> Most attempts to write to a buffer object, either with glBufferSubData or mapping, will halt until all rendering commands using that part of the buffer object have completed

So unless I'm reading wrong (which is definitely possible), writing to a mapped buffer is implicitly synchronized, even between draw calls

kvark:
First of all, this "will halt" is a terrible thing, and we should not expose a feature that results in that happening potentially many times per frame.
Secondly, it talks about us doing `glMapBuffer`. This is not the same as just writing data into a buffer that's already mapped. With a persistently mapped buffer, the driver can't detect when you are writing to it.

JCapucho:
> First of all, this "will halt" is a terrible thing, and we should not expose a feature that results in that happening potentially many times per frame. Secondly, it talks about us doing `glMapBuffer`. This is not the same as just writing data into a buffer that's already mapped. With a persistently mapped buffer, the driver can't detect when you are writing to it.

Okay, then I'll go with the `gfx-hal` approach, from my understanding the push constants data is copied to a global memory area and then after that it's made into a uniform with [`glUniform`](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glUniform.xhtml) is that correct?

Since `naga` currently only supports one push constant buffer it would mean that it would only support 64 bytes for now

kvark:
> Okay, then I'll go with the gfx-hal approach, from my understanding the push constants data is copied to a global memory area and then after that it's made into a uniform with glUniform is that correct?

Yes. It's not far from what this PR is doing already.

JCapucho:
@kvark I've hit a problem, to use the `glUniform` functions a uniform location needs to be passed, but because `naga` is generating a uniform block it doesn't get a location but a binding, so either `naga` starts outputting non block uniforms for push constants or we can't use this method.

JCapucho:
I've tried fiddling around with the shaders and glsl only allows to use directly the struct without the wrapping block if it also drops the layout qualifier, which isn't that great since then we get implementation defined layouts.

kvark:
Right, naga should not generate them in a uniform block.

JCapucho:
@kvark I've update it to use freestanding uniforms, it's basically a copy of what gfx-hal does.

wgpu-hal/src/gles/device.rs
```diff
@@ -281,9 +281,48 @@ impl super::Device {
...
+                    | glow::SAMPLER_CUBE
+                    | glow::SAMPLER_CUBE_MAP_ARRAY
+                    | glow::SAMPLER_CUBE_MAP_ARRAY_SHADOW
+                    | glow::SAMPLER_CUBE_SHADOW => continue,
+                    _ => {}
+                }
+
                 if let Some(location) = gl.get_uniform_location(program, &name) {
-                    // Sampler2D won't show up in UniformLocation and the only other uniforms
-                    // should be push constants
```
JCapucho, line 286: In my modified buffer example, while not a `sampler2d`, a `usampler2d` triggered this code path, that's why there is a match now to guard against sampler types, because they were being included and causing problems later.

Cargo.toml
```diff
@@ -18,6 +18,7 @@ default-members = ["wgpu", "wgpu-hal", "wgpu-info"]
 
 [patch."https://github.com/gfx-rs/naga"]
 #naga = { path = "../naga" }
+naga = { git = "https://github.com/JCapucho/naga", branch = "glsl-out-push-constants-v2" }
```
JCapucho, line 21: This is using a patched `naga` for now to allow for iterating, I'll merge it upstream after we converge on a desing and update `naga`

wgpu-hal/src/gles/device.rs
kvark: I wish there was an easier way to detect if this uniform is for a texture. Maybe by name?

wgpu-hal/src/gles/mod.rs
```diff
@@ -716,6 +716,11 @@ enum Command {
     InsertDebugMarker(Range<u32>),
     PushDebugGroup(Range<u32>),
     PopDebugGroup,
+    SetPushConstants {
+        uniform: UniformDesc,
+        /// Offset from the start of the `data_bytes`
+        range: Range<usize>,
```
kvark, line 722: let's use `Range<u32>` for consistency with the other things here?

wgpu-hal/src/gles/queue.rs
```diff
@@ -1067,6 +1067,84 @@ impl super::Queue {
...
+                            raw.as_ptr() as *const _,
+                            raw.len() / mem::size_of::<T>(),
+                        )
+                    }
+                }
+
+                match uniform.utype {
+                    glow::FLOAT => {
+                        let data = get_data::<f32>(data_bytes, range)[0];
+                        gl.uniform_1_f32(Some(&uniform.location.clone()), data);
```
kvark, line 1091: looks like we could save `Some(&uniform.location.clone())` somewhere for DRY

wgpu-hal/src/gles/command.rs
```diff
@@ -603,10 +616,31 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
...
+        if uniforms.is_empty() {
+            unimplemented!()
+        }
+
+        let uniform = if offset == 0 {
+            // If offset is zero, we can just return the first item
+            // in our uniform list
+            uniforms.get(0).unwrap()
+        } else {
+            match uniforms.binary_search_by(|uniform| uniform.offset.cmp(&offset)) {
```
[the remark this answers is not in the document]
JCapucho: I copied this straight from gfx-hal but I can see the problem and triggering it is really easy, we could store the size of the uniform and then issue multiple `SetPushConstants` as needed.

wgpu-hal/src/gles/device.rs
JCapucho: By the name would be hard since naga names are like _group_0_binding_0, only option would be changing naga's naming scheme

wgpu-hal/src/gles/command.rs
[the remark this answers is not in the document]
kvark: yeah, gfx-hal wasn't ideal by any means :)

wgpu-hal/src/gles/command.rs
```diff
@@ -148,8 +164,7 @@ impl super::CommandEncoder {
     fn set_pipeline_inner(&mut self, inner: &super::PipelineInner) {
         self.cmd_buffer.commands.push(C::SetProgram(inner.program));
 
-        //TODO: push constants
-        let _ = &inner.uniforms;
+        self.state.uniforms = inner.uniforms.clone();
```
kvark, line 167: We really don't want to clone a heap-allocated container here.
Instead, it should do something like this:
```rust
self.state.uniforms.clear();
self.state.uniforms.extend_from_slice(&inner.uniforms);
```

wgpu-hal/src/gles/command.rs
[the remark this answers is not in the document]
kvark: I think what we can do here is simply have an array like this:
```rust
push_offset_to_uniform: [Uniform; MAX_PUSH_CONSTANTS];
```
You'd then have some special value (e.g. `!0`) for offsets that don't have a corresponding uniform.
So here, you have a range, and you need to figure out all of the uniforms affected by it. You'd do something simple, like this:
```rust
let mut offset = start_offset;
while offset < end {
  let uniform = &self.state.push_offset_to_uniform[offset as u8];
  self.cmd_buffer
            .commands
            .push(C::SetPushConstants { *uniform, offset..offset+uniform.size });
  offset += uniform.size;
}
```

That's a very straightforward linear scan, it should perform well.

wgpu-hal/src/gles/device.rs
```diff
@@ -281,9 +281,48 @@ impl super::Device {
...
+                    | glow::SAMPLER_2D_ARRAY_SHADOW
+                    | glow::SAMPLER_2D_MULTISAMPLE
+                    | glow::SAMPLER_2D_MULTISAMPLE_ARRAY
+                    | glow::SAMPLER_2D_RECT
+                    | glow::SAMPLER_2D_RECT_SHADOW
+                    | glow::SAMPLER_3D
+                    | glow::SAMPLER_CUBE
+                    | glow::SAMPLER_CUBE_MAP_ARRAY
+                    | glow::SAMPLER_CUBE_MAP_ARRAY_SHADOW
+                    | glow::SAMPLER_CUBE_SHADOW => continue,
```
kvark, line 321: oh, can we move this match into `conv.rs` so that it doesn't clutter this module too much?

wgpu-hal/src/gles/command.rs
JCapucho: This is using a `Box` so it can't be done this way but I agree with you that cloning a `Box` isn't ideal

kvark: Can't you switch to a `Vec<>`?

wgpu-hal/src/gles/command.rs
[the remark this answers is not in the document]
JCapucho: I'm sorry there are some parts that confuse me, first what should be `MAX_PUSH_CONSTANTS` from what I could find there's nothing like it in wgpu yet.

Second how would you assign `0!` if the type of the array is `Uniform`.

Finally, how would this work
```rust
&self.state.push_offset_to_uniform[offset as u8]
```
indexing by `u8` and the index isn't increasing by one but by the size of the previous uniform, which would mean the array should have as many element as the address space of the push constant

JCapucho:
@kvark I've updated the PR with your recommendations, I think the annoying `Vec` copy could be removed if instead we used a `Arc<[UniformDesc; MAX_PUSH_CONSTANTS]>` but this has the downside that it can't derive `Default` so `State` would need a custom impl.

wgpu-hal/src/gles/command.rs
```diff
@@ -28,6 +28,7 @@ pub(super) struct State {
     has_pass_label: bool,
     instance_vbuf_mask: usize,
     dirty_vbuf_mask: usize,
+    push_offset_to_uniform: Vec<super::UniformDesc>,
```
kvark, line 31: looks like this is the only part that's heap allocated. This isn't a problem in itself, but we could use `ArrayVec` to solve it

wgpu-hal/src/gles/command.rs
```diff
@@ -603,10 +621,25 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
...
+        start_offset: u32,
+        data: &[u32],
     ) {
-        unimplemented!()
+        let range = self.cmd_buffer.add_push_constant_data(data);
+
+        let end = start_offset + data.len() as u32 * 4;
+        let mut offset = start_offset;
+        while offset < end {
+            let uniform = self.state.push_offset_to_uniform[offset as usize % 4].clone();
```
kvark, line 632: this should be `offset >> 2` (or `offset / 4`) instead of `offset % 4`

wgpu-hal/src/gles/command.rs
```diff
@@ -603,10 +621,25 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
...
+            let size = uniform.size;
+            if uniform.location.is_none() {
+                panic!("No uniform for push constant");
+            }
+            self.cmd_buffer.commands.push(C::SetPushConstants {
+                uniform,
+                range: (range.start + offset)..range.end,
```
kvark, line 639: should this end with `range.start + offset + size` instead of `range.end`?

wgpu-hal/src/gles/device.rs
```diff
@@ -272,30 +272,44 @@ impl super::Device {
...
 
-            uniforms.into_boxed_slice()
-        };
+            if let Some(location) = gl.get_uniform_location(program, &name) {
+                if uniforms[offset % 4].location.is_some() {
+                    panic!("Offset already occupied")
+                }
+
+                // `size` will always be 1 so we need to guess the real size from the type
+                let uniform_size = match utype {
```
kvark, line 293: similarly, let's move this match into `conv.rs`

kvark (commented):
Looks good overall, thank you for doing this!
Just a few smaller corrections are left to do

wgpu-hal/src/gles/command.rs
JCapucho: `ArrayVec` needs for `T` to be `Copy` when using `try_extend_from_slice` and `UniformDesc`doesn't implement it nor can it, since `glow::UniformLocation` isn't `Copy`

wgpu-hal/src/gles/command.rs
JCapucho: Yes, but since the range only specifies where in `data_bytes` is our push constant data, the end functions more like a bound check, so if we leave the end as the end of the allocation it serves the same purpose.

Technically we only need the start offset

wgpu-hal/src/gles/command.rs
kvark: `extend_from_slice` is just a convenience method, which btw isn't requiring `Copy` in std. So let's just `extend()` instead.

wgpu-hal/src/gles/command.rs
kvark: let's remove the end entirely then, no need to carry it around for validation

JCapucho:
@kvark I've finished applying your suggestions, only one problem remains, `UniformLocation` for webgl is not `Sync` which causes `CommandEncoder` to also not be.

kvark:
@JCapucho let's `unsafe impl Sync for UniformDesc`, at least for now

JCapucho:
@kvark everything seems to be working now, deno is failing because it's not picking up the patch for naga but once everything is merged upstream it should be fixed. Are we good to go?

kvark (approved):
Looking good!
So the last thing here is to change the naga dependency back to upstream.

kvark:
Weird, deno stuff fails but the rest is fine? https://github.com/gfx-rs/wgpu/runs/4900108750?check_suite_focus=true#step:4:255
@crowlKats @lucacasonato

JCapucho:
@kvark I've updated to latest naga, it includes quite some shader changes beacuse of the attribute changes so I've made it into another PR but it can be all squashed

crowlKats:
@kvark seems to me like it was just ci being flaky due to caching

kvark (approved):
Thank you so much for updating the shaders to the new syntax!

PR #2401, opened 2022-01-17
chore: bump naga to a1840be

AaronO:
So tip no longer indirectly pins `indexmap`, etc...

### DIFF
--- a/wgpu-hal/src/gles/device.rs
+++ b/wgpu-hal/src/gles/device.rs
@@ -858,6 +858,7 @@ impl crate::Device<super::Api> for super::Device {
                 version: self.shared.shading_language_version,
                 writer_flags,
                 binding_map,
+                push_constant_binding: 0, //TODO?
             },
         })
     }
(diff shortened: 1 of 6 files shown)


wgpu-hal/src/metal/device.rs
```diff
@@ -117,7 +117,10 @@ impl super::Device {
         let mut immutable_buffer_mask = 0;
         for (var_handle, var) in module.global_variables.iter() {
             if var.class == naga::StorageClass::WorkGroup {
-                let size = module.types[var.ty].inner.span(&module.constants);
+                let size = module.types[var.ty]
+                    .inner
+                    .try_size(&module.constants)
```
kvark, line 122: should just call `size()` here

kvark (commented):
One small thing

wgpu-hal/src/metal/device.rs
AaronO: Fixed

kvark (approved):

wgpu-hal/src/gles/device.rs
```diff
@@ -858,6 +858,7 @@ impl crate::Device<super::Api> for super::Device {
                 version: self.shared.shading_language_version,
                 writer_flags,
                 binding_map,
+                push_constant_binding: 0, //TODO?
```
kvark, line 861: Yes, I think doing this should work. The backend still isn't exposing PUSH_CONSTANTS feature, so it should fail the validation before using this `0`.

kvark (approved):

PR #2402, opened 2022-01-17
fix(deno_webgpu): add TS type declarations

AaronO:
This restores the TS types to `deno_webgpu` and the `deno_webgpu::get_declaration()` func, which are necessary for `deno_webgpu` to be consumed by deno CLI


AaronO:
Closing since after talking to @kvark, we concluded that Deno will maintain the TS type declarations in https://github.com/denoland/deno

PR #2403, opened 2022-01-18
lint: deno_webgpu & wgpu-core

AaronO:
Mostly unnecessary `.into()`s


deno_webgpu/src/pipeline.rs
```diff
@@ -322,7 +322,7 @@ pub fn op_webgpu_create_render_pipeline(
         let mut targets = Vec::with_capacity(fragment.targets.len());
 
         for target in fragment.targets {
-            targets.push(target.try_into()?);
+            targets.push(target);
         }
```
crowlKats, line 326: This whole thing is useless then if try_into isnt required anymore. Could you remove it (and adjust line 333 accordingly then)?

AaronO: Sure, there's more code that could potentially be cleaned up / refactored. Just wanted to scope it to lints so clippy no longer complains (since our CI will fail)

PR #2409, opened 2022-01-19
hal/dx12: improve RowPitch computation

kvark:
**Connections**
Fixes https://github.com/kvark/vange-rs/issues/183

**Description**
For 1D copies, we set RowPitch==0, and DX12 doesn't like this.
PR includes a few nice refactoring bits.

**Testing**
vange-rs


kvark:
published in wgpu-hal-0.12.3

PR #2422, opened 2022-01-25
feature: emscripten

caiiiycuk:
This is reimplementation of #2233 

--

Added new feature "emscripten", this feature can be used in combination with --target wasm32-unknown-emscripten to build wgpu for emscripten. 

**raw-gles.rs** example now works also with emscripten, to run it:
1. install emsdk
2. build this example with cargo:
```
EMMAKEN_CFLAGS="-g -s ERROR_ON_UNDEFINED_SYMBOLS=0 --no-entry -s FULL_ES3=1" cargo build --example raw-gles --target wasm32-unknown-emscripten --features gles,emscripten,webgl
```
3. copy raw-gles.em.html into target/wasm32-unknown-emscripten/debug/examples/ and open it in browser

This PR also adds a workflow to check emscripten build. I think that raw-gles example is only needed example for emscripten integration it also can be a test, but it hard to do it automatically because need to install emcc. I think not worth to doing it.

Typical usage of wgpu and emscripten will be through external gles integration like in raw-gles.rs. Because  windows management (like winit) crate not exists for emscripten, so it will be used with some native library like SDL/2+wgpu. SDL will initialize context and controls window/events and wgpu will be used as renderer library. 

--

There is 2 identical comments in 2233 that I can't solve:
```rust
        #[cfg(feature = "emscripten")]
        let wsi_library = None;

        #[cfg(feature = "emscripten")]
        let wsi_kind = WindowKind::Unknown;

        let (display, wsi_library, wsi_kind) = if let (Some(library), Some(egl)) =
            (wayland_library, egl.upcast::<egl::EGL1_5>())
        {
            log::info!("Using Wayland platform");
...
        } else {
            log::info!("Using default platform");
            let display = egl.get_display(egl::DEFAULT_DISPLAY).unwrap();
            (display, None, WindowKind::Unknown)
        };

        #[cfg(feature = "emscripten")]
        let display = egl.get_display(egl::DEFAULT_DISPLAY).unwrap();
```

Ideally we can remove this `#[cfg` because last `else` is what we actually need. But I can't find good way to jump into last **else branch** when feature emscripten is set. And egl::EGL1_5 is not defined when feature is emscripten.

raw-gles.rs in browser (tested in Chrome an FF):
изображение

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ env:
 # - local: build for the same target as we compile on, and do local tests
 # - other: build without testing, e.g. cross-build
 # - web: build for the Web
+# - em: build for the Emscripten
 
 jobs:
   build:
@@ -94,6 +95,12 @@ jobs:
             tool: clippy
             kind: web
 
+          - name: Emscripten
+            os: ubuntu-20.04
+            target: wasm32-unknown-emscripten
+            tool: clippy
+            kind: em
+
     name: Check ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
 
@@ -167,6 +174,15 @@ jobs:
 
           # build docs
           cargo doc --target ${{ matrix.target }} -p wgpu --no-deps
+
+      - name: check em
+        if: matrix.kind == 'em'
+        run: |
+          # build for Emscripten/WebGL
+          cargo ${{matrix.tool}} --target ${{ matrix.target }} -p wgpu -p wgpu-hal --features webgl,emscripten
+
+          # build raw-gles example
+          cargo ${{matrix.tool}} --target ${{ matrix.target }} --example raw-gles --features webgl,emscripten
 
       - name: check native
         if: matrix.kind == 'local' || matrix.kind == 'other'
--- a/wgpu-hal/Cargo.toml
+++ b/wgpu-hal/Cargo.toml
@@ -18,6 +18,7 @@ vulkan = ["naga/spv-out", "ash", "gpu-alloc", "gpu-descriptor", "libloading", "i
 gles = ["naga/glsl-out", "glow", "egl", "libloading"]
 dx12 = ["naga/hlsl-out", "native", "bit-set", "range-alloc", "winapi/d3d12", "winapi/d3d12shader", "winapi/d3d12sdklayers", "winapi/dxgi1_6"]
 renderdoc = ["libloading", "renderdoc-sys"]
+emscripten = ["gles"]
 
 [[example]]
 name = "halmark"
@@ -66,6 +67,11 @@ egl = { package = "khronos-egl", version = "4.1", features = ["dynamic"], option
 #Note: it's only unused on Apple platforms
 libloading = { version = "0.7", optional = true }
 
+[target.'cfg(target_os = "emscripten")'.dependencies]
+egl = { package = "khronos-egl", version = "4.1", features = ["static", "no-pkg-config"] }
+#Note: it's unused by emscripten, but we keep it to have single code base in egl.rs
+libloading = { version = "0.7", optional = true }
+
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3", features = ["libloaderapi", "windef", "winuser"] }
 native = { package = "d3d12", version = "0.4.1", features = ["libloading"], optional = true }
@@ -75,7 +81,7 @@ mtl = { package = "metal", git = "https://github.com/gfx-rs/metal-rs", rev = "14
 objc = "0.2.5"
 core-graphics-types = "0.1"
 
-[target.'cfg(target_arch = "wasm32")'.dependencies]
+[target.'cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))'.dependencies]
 wasm-bindgen = { version = "0.2" }
 web-sys = { version = "0.3", features = ["Window", "HtmlCanvasElement", "WebGl2RenderingContext"] }
 js-sys = { version = "0.3" }
--- a/wgpu-hal/examples/raw-gles.rs
+++ b/wgpu-hal/examples/raw-gles.rs
@@ -1,15 +1,16 @@
 //! This example shows interop with raw GLES contexts -
 //! the ability to hook up wgpu-hal to an existing context and draw into it.
+//!
+//! Emscripten build:
+//! 1. install emsdk
+//! 2. build this example with cargo:
+//!    EMMAKEN_CFLAGS="-g -s ERROR_ON_UNDEFINED_SYMBOLS=0 --no-entry -s FULL_ES3=1" cargo build --example raw-gles --target wasm32-unknown-emscripten --features emscripten,webgl
+//! 3. copy raw-gles.em.html into target/wasm32-unknown-emscripten/debug/examples/ and open it in browser
 
 extern crate wgpu_hal as hal;
 
-#[cfg(target_arch = "wasm32")]
-fn main() {}
-
 #[cfg(not(target_arch = "wasm32"))]
 fn main() {
-    use hal::{Adapter as _, CommandEncoder as _, Device as _, Queue as _};
-
     env_logger::init();
     println!("Initializing external GL context");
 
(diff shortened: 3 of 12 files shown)


wgpu-hal/Cargo.toml
```diff
@@ -66,6 +67,11 @@ egl = { package = "khronos-egl", version = "4.1", features = ["dynamic"], option
 #Note: it's only unused on Apple platforms
 libloading = { version = "0.7", optional = true }
 
+[target.'cfg(target_os = "emscripten")'.dependencies]
+egl = { package = "khronos-egl", version = "4.1", features = ["static", "no-pkg-config"] }
```
kvark, line 71: that's a bit weird: "static" actually implies "pkg-config":
https://github.com/timothee-haudebourg/khronos-egl/blob/76d7d917f1c5b40061596653a16f0340683b5e0b/Cargo.toml#L15

wgpu-hal/src/gles/egl.rs
```diff
@@ -1015,6 +1049,7 @@ impl crate::Surface<super::Api> for Surface {
                 attributes.push(egl::ATTRIB_NONE as i32);
 
                 // Careful, we can still be in 1.4 version even if `upcast` succeeds
+                #[cfg(not(feature = "emscripten"))]
                 let raw_result = match self.egl.instance.upcast::<egl::EGL1_5>() {
```
kvark, line 1053: this is done because `egl::EGL1_5` isn't on emscripten? any idea why this is the case?
I see "1_5" feature enabled by default - https://github.com/timothee-haudebourg/khronos-egl/blob/76d7d917f1c5b40061596653a16f0340683b5e0b/Cargo.toml#L14

wgpu-hal/Cargo.toml
caiiiycuk: This is from [official documentation](https://github.com/timothee-haudebourg/khronos-egl)
### Static linking

You must enable static linking using the `static` feature in your `Cargo.toml`:
```toml
khronos-egl = { version = ..., features = ["static"] }
```

This will add a dependency to the [`pkg-config`](https://crates.io/crates/pkg-config) crate,
necessary to find the EGL library at compile time.

If you wish to disable linking EGL in this crate, and provide linking in
your crate instead, enable the `no-pkg-config` feature.
```toml
khronos-egl = {version = ..., features = ["static", "no-pkg-config"]}
```

--

We use "no-pkg-config" because emscripten provides own implementation of EGL

wgpu-hal/src/gles/egl.rs
caiiiycuk: Not 100% sure how features works, but we use "static" feature that not implies 1_5 or any.

wgpu-hal/Cargo.toml
```diff
@@ -66,6 +67,11 @@ egl = { package = "khronos-egl", version = "4.1", features = ["dynamic"], option
...
+#Note: it's only unused on Apple platforms
```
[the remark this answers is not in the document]
caiiiycuk: It's sort of compromise, if I remove it then I need rewrite more code in egl.rs for emscripten. I think keeping it same as much as possible is better then removing this crate from Cargo.toml. What do you think about this?

.github/workflows/ci.yml
```diff
@@ -168,6 +175,15 @@ jobs:
           # build docs
           cargo doc --target ${{ matrix.target }} -p wgpu --no-deps
 
+      - name: check em
+        if: matrix.kind == 'em'
+        run: |
+          # build for Emscripten/WebGL
+          cargo ${{matrix.tool}} --target ${{ matrix.target }} -p wgpu -p wgpu-hal --features webgl,emscripten,gles
```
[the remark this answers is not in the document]
caiiiycuk: :+1: will do

wgpu-hal/examples/raw-gles.rs
```diff
@@ -1,15 +1,16 @@
 //! This example shows interop with raw GLES contexts -
 //! the ability to hook up wgpu-hal to an existing context and draw into it.
+//!
+//! Emscripten build:
+//! 1. install emsdk
```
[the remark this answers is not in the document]
caiiiycuk: It's not difficult, e.g. we can install emsdk and build raw-gles. Then we can run test in browser take screenshot, and compare it with expected, or something similar. I think you already should have last part for wasm32-unknown-unknown, and I can reuse it. So if you think it's worth to doing I will do.
 
Let me explain why I think it's not worth to doing. The emscripten build is testing same path that works on linux pc, so I use only khronos-egl crate in raw-gles code, probably we can use the same code for native integration. The only one difference is in window creation:

```rust
    let surface = unsafe {
        let window = std::ptr::null_mut::<std::ffi::c_void>();
        egl.create_window_surface(display, config, window, None)
    }
    .expect("unable to create surface");
```

We used null_ptr because it's allowed for em, if we put actual code of window creation for linux this example also should works in linux. Moreover this code path is same that egl.rs did internally. I mean there is no browser related stuff in emscripten integration.

The whole project is compiled by rust with emscripten triple, and when code is compiled emcc linker is used to produce WebAssemlby module. I want to say while code compiles it's not an wgpu/rust problem if something not working, probably it's emscripten bug, however egl implementation of em is very solid it didnt' change for years. I mean that if we do this in CI it will be more like integration test, that test compatibility of emscripten version against khronos-egl crate/rust/wgpu. 
Regarding native path - it's already covered by other tests. There are small amount of code  related to emscripten that are not covered, but I think that compilation check is enough here.

So, maybe is better to just add to README.md how to build example with emscripten, and who wants will check if it works.

Anyway, if you think it worth to add in CI, lets do it.

wgpu-hal/Cargo.toml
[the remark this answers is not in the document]
kvark: ok, reducing the changes to minimum makes sense to me. Then let's just change the comment to say that this isn't actually used.

wgpu-hal/examples/raw-gles.rs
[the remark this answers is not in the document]
kvark: Ok, that sounds rather convincing. Let's proceed with your original plan:)

wgpu-hal/Cargo.toml
kvark: thanks! that's some perverted stuff here, but we are complying. Your code is good then

wgpu-hal/src/gles/egl.rs
kvark: Ok, here is what I see. It's not that EGL1_5 isn't available, but rather that a static instance can't be `upcast` at all...

caiiiycuk: How I can use the same default branch (_ =>) and not failing on unresolved EGL1_5?

caiiiycuk: Ok, I have an idea

caiiiycuk:
Mark it as draft until fully test it on vaners + librusty_vangers

caiiiycuk:
Tested, works fine

kvark (approved):

kvark:
We'll probably need a separate section in README about all the ways this can be compiled for the Web

caiiiycuk:
@kvark Regarding to emscripten, I think it's will be very handy for projects that use wgpu-native. Since this projects are on C++ and use Rust to render, they can't be ported to web until this PR. Because you can't merge 2 Modules one is from emscripten, second if from wasm32-unknown-unknown. But now, it's super simple compile both with emscripten and link wgpu-native statically or dynamically and it will work.

So for emscripten, I think we can try to compile examples from wgpu-native. Or I think it's better we can create simple SDL1/2+wgpu-native project and write tutorial how to compile it. Because I think SDL1/2 is widely used and have first class support from emscripten.

caiiiycuk:
Also I can do announcement in emscripten channels about it this template (SDL+wgpu)

kvark:
Yes, I'm totally with you on this. C-based projects would be able to target WebGL via webgpu-headers, it's a very cool feature. The example code should live in https://github.com/gfx-rs/wgpu-native/ for this.

PR #2424, opened 2022-01-27
IdentityManager: `from_index` method is unneeded.

jimblandy:
The `min_index` parameter has no useful effect. Because the range 0..n is pushed
on the free list, index values less than `min_index` do get allocated.

Document `IdentityManager` and `TypedId`.

### DIFF
--- a/wgpu-core/src/hub.rs
+++ b/wgpu-core/src/hub.rs
@@ -16,21 +16,52 @@ use wgt::Backend;
 use std::cell::Cell;
 use std::{fmt::Debug, marker::PhantomData, mem, ops};
 
-/// A simple structure to manage identities of objects.
+/// A simple structure to allocate [`Id`] identifiers.
+///
+/// Calling [`alloc`] returns a fresh, never-before-seen id. Calling [`free`]
+/// marks an id as dead; it will never be returned again by `alloc`.
+///
+/// Use `IdentityManager::default` to construct new instances.
+///
+/// `IdentityManager` returns `Id`s whose index values are suitable for use as
+/// indices into a vector of information (which you maintain elsewhere) about
+/// those ids' referents:
+///
+/// - Every live id has a distinct index value. Each live id's index selects a
+///   distinct element in the vector.
+///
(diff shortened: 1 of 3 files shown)


wgpu-core/src/hub.rs
```diff
@@ -16,21 +16,42 @@ use wgt::Backend;
...
+/// A simple structure to allocate [`Id`] identifiers.
+///
+/// Calling [`alloc`] returns a fresh, never-before-seen id. Calling [`free`]
+/// marks an id as dead; it will never be returned again by `alloc`.
+///
+/// Use `IdentityManager::default` to construct new instances.
+///
+/// The index values of `Id`s allocated by an `IdentityManager` are suitable for
+/// use as indices into a vector of information about the ids' referents. The
+/// `alloc` method returns an id with a given index only if there is some extant
```
kvark, line 28: I don't understand this sentence. The method returns a given index only if there is some higher index in use?
Also, "ll lower non-negative index values are currently in use" isn't entirely correct. There is a freelist inside, and it gives you the last index that was freed, which isn't necessarily the lowest free index.

wgpu-core/src/hub.rs
```diff
@@ -16,21 +16,42 @@ use wgt::Backend;
...
 
 impl IdentityManager {
-    pub fn from_index(min_index: u32) -> Self {
-        Self {
-            free: (0..min_index).collect(),
-            epochs: vec![1; min_index as usize],
-        }
-    }
-
+    /// Allocate a fresh, never-before-seen id with the given `backend`.
```
kvark, line 54: Let's add a note saying that the backend becomes a part of the ID, so that there isn't any confusion about backend being some namespace from the outside.

kvark (changes requested):
Thank you! Just one note

wgpu-core/src/hub.rs
jimblandy: Right, I was trying to capture the behavior of the freelist there. It's too dense, I'll rephrase.

kvark (approved):

Issue #2419, opened 2022-01-25
`request_adapter()` failed at Android 8.1


jinleili:
Test device: Google Nexus 5x(Android 8.1.0)
Backend: VULKAN

```sh
2022-01-25 10:55:31.018 27110-27110/name.jinleili.wgpu W/wgpu_hal::vulkan::ada..: sample_rate_shading feature is not supported, hiding adapter: Adreno (TM) 418
2022-01-25 10:55:31.019 27110-27110/name.jinleili.wgpu A/libc: Fatal signal 6 (SIGABRT), code -6 in tid 27110 (e.jinleili.wgpu), pid 27110 (e.jinleili.wgpu)
```
<details>
	<summary>Logcat full output:</summary>
	<pre><code>
2022-01-25 11:16:12.163 28652-28652/? I/zygote64: Late-enabling -Xcheck:jni
2022-01-25 11:16:12.163 28652-28652/? I/zygote64: Late-enabling -Xcheck:jni
2022-01-25 11:16:12.163 28652-28652/? I/zygote64: Late-enabling -Xcheck:jni
2022-01-25 11:16:12.697 28652-28688/name.jinleili.wgpu D/OpenGLRenderer: HWUI GL Pipeline
2022-01-25 11:16:12.909 28652-28652/name.jinleili.wgpu W/MaterialButtonToggleGroup: Child order wasn't updated
2022-01-25 11:16:12.909 28652-28652/name.jinleili.wgpu I/chatty: uid=10106(name.jinleili.wgpu) identical 2 lines
2022-01-25 11:16:12.910 28652-28652/name.jinleili.wgpu W/MaterialButtonToggleGroup: Child order wasn't updated
2022-01-25 11:16:13.299 28652-28657/name.jinleili.wgpu I/zygote64: Do partial code cache collection, code=30KB, data=26KB
2022-01-25 11:16:13.299 28652-28657/name.jinleili.wgpu I/zygote64: After code cache collection, code=30KB, data=26KB
2022-01-25 11:16:13.299 28652-28657/name.jinleili.wgpu I/zygote64: Increasing code cache capacity to 128KB
2022-01-25 11:16:47.502 28652-28688/name.jinleili.wgpu I/Adreno: QUALCOMM build                   : 8e59954, I0be83d0d26
    Build Date                       : 09/22/17
    OpenGL ES Shader Compiler Version: EV031.21.02.00
    Local Branch                     : O17A
    Remote Branch                    : 
[...]

PR #2425, opened 2022-01-27
Add MULTISAMPLED_SHADING downlevel flag

jinleili:
**Connections**
About https://github.com/gfx-rs/wgpu/issues/2419

### DIFF
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -100,7 +100,9 @@ impl PhysicalDeviceFeatures {
             core: vk::PhysicalDeviceFeatures::builder()
                 .robust_buffer_access(private_caps.robust_buffer_access)
                 .independent_blend(true)
-                .sample_rate_shading(true)
+                .sample_rate_shading(
+                    downlevel_flags.contains(wgt::DownlevelFlags::MULTISAMPLED_SHADING),
+                )
                 .image_cube_array(
                     downlevel_flags.contains(wgt::DownlevelFlags::CUBE_ARRAY_TEXTURES),
                 )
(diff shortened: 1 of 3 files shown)


wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -929,13 +932,13 @@ impl super::Instance {
...
-            );
-            return None;
-        }
+        // if phd_features.core.sample_rate_shading == 0 {
+        //     log::warn!(
+        //         "sample_rate_shading feature is not supported, hiding adapter: {}",
+        //         info.name
+        //     );
+        //     return None;
+        // }
```
cwfitzgerald, line 941: If this isn't needed, we should just remove it.

cwfitzgerald (changes requested):
One nit, but otherwise LGTM

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -342,6 +344,7 @@ impl PhysicalDeviceFeatures {
             Df::FRAGMENT_WRITABLE_STORAGE,
             self.core.fragment_stores_and_atomics != 0,
         );
+        dl_flags.set(Df::MULTISAMPLED_SHADING, self.core.sample_rate_shading != 0);
```
cwfitzgerald, line 347: The other backends should also set this appropriately

cwfitzgerald (changes requested):
Sorry missed one thing

wgpu-hal/src/vulkan/adapter.rs
jinleili: But I didn't find sample_rate_shading logic at other backends

cwfitzgerald: If there is no flag determining if it is supporting, they should always enable it.

jinleili: Checked other backends, all flags are enabled by default like Vulkan backends：
https://github.com/gfx-rs/wgpu/blob/b19000367c15ebc3f4aa51871e6d6b9b670126f8/wgpu-hal/src/metal/adapter.rs#L1036

cwfitzgerald: Ahhh I didn't catch that. Cool

cwfitzgerald (approved):
Thanks!

Issue #2426, opened 2022-01-27
MAPPABLE_PRIMARY_BUFFERS on M1 Pro


BGR360:
Came over here from https://github.com/bevyengine/bevy/issues/3686#issuecomment-1020820103.

Mostly just starting a conversation, as I have little to no knowledge on the subject matter.

It was suggested in that discussion I linked that perhaps wgpu shouldn't warn MAPPABLE_PRIMARY_BUFFERS on platforms like Apple M1 that have powerful integrated GPUs. Thoughts?

cwfitzgerald:
This is just from us erroneously detecting it as a dedicated, not an integrated, gpu. Once this is fixed, the warning will go away.

BGR360:
I'd be willing to take a stab at fixing that. Could you point me in the right direction?

cwfitzgerald:
https://github.com/gfx-rs/wgpu/blob/master/wgpu-hal/src/metal/mod.rs#L117-L121

Here is the offending check. I'm not sure there's a clean way in the metal api to determine uma-ness, but I haven't checked. If not we can likely use the adapter family to determine which is which.

kvark:
See also - https://github.com/KhronosGroup/MoltenVK/issues/1201
We could do a similar fix.

superdump:
I was just looking into this and I noticed that it's currently based on 'power' where high is mapped to discrete and low is mapped to integrated. But then I saw this in mtl::Device: https://github.com/gfx-rs/metal-rs/blob/master/src/device.rs#L1580 This seems like a good deciding factor perhaps?

cwfitzgerald:
Yup, that's what MVK does

superdump:
So, to clarify, we would need to fix this for the Metal backend, and the GL backend at least for WebGL2. The Vulkan backend using MoltenVK with the fix should be correct already. Right?

superdump:
https://github.com/gfx-rs/wgpu/pull/2429 how's this?

PR #2429, opened 2022-01-28
Identify Apple M1 GPU as integrated

superdump:
**Connections**
* Closes https://github.com/gfx-rs/wgpu/issues/2426

**Description**
The Apple M1 line of GPUs use a unified memory architecture that implies they are integrated but wgpu currently uses whether they are low power (Metal backend), or name-based heuristics (GL backend) which classify them as discrete, which seems incorrect.

**Testing**
I ran a few native examples (so only tested the Metal backend) and checked the adapter info to see that it had changed to `IntegratedGpu`.

### DIFF
--- a/wgpu-hal/src/metal/mod.rs
+++ b/wgpu-hal/src/metal/mod.rs
@@ -108,18 +108,13 @@ impl crate::Instance<Api> for Instance {
             .into_iter()
             .map(|dev| {
                 let name = dev.name().into();
-                let has_unified_memory = dev.has_unified_memory();
                 let shared = AdapterShared::new(dev);
                 crate::ExposedAdapter {
                     info: wgt::AdapterInfo {
                         name,
                         vendor: 0,
                         device: 0,
-                        device_type: if has_unified_memory {
-                            wgt::DeviceType::IntegratedGpu
-                        } else {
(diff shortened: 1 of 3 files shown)


wgpu-hal/src/metal/mod.rs
```diff
@@ -108,13 +108,14 @@ impl crate::Instance<Api> for Instance {
             .into_iter()
             .map(|dev| {
                 let name = dev.name().into();
+                let has_unified_memory = dev.has_unified_memory();
```
kvark, line 111: We can't call it unconditionally, this method is only available on macOS 10.15 and iOS 13.0.
Generally what we do for this is adding a boolean into `PrivateCapabilities` and checking it here.

superdump: I added an `Option<bool>` so that I could tell whether it was set or not. That way, when the OS versions are new enough, I can set and use `has_unified_memory` to indicate whether the GPU is integrated or not, and for older OSes where the API is unavailable, I can fall back to using the low power property to indicate an integrated GPU.

kvark (approved):

kvark:
This isn't how I saw this being done, but it's good. Maybe even better than I wanted :)

PR #2433, opened 2022-01-29
Make shadow example runnable on iOS Android devices

jinleili:
**Description**
On iOS and Android devices, the `min_uniform_buffer_offset_alignment` is equal 64 less than the `EntityUniforms` size, running on these devices failed at `assert!(entity_uniform_size <= uniform_alignment)`:
```sh
thread '<unnamed>' panicked at 'assertion failed: entity_uniform_size <= uniform_alignment', src/examples/shadow.rs:289:9
```

**Testing**
Tested on macOS, iPhone 12 Pro (iOS 15.3) and Honor V20 (Android 10.0)
| IMG_0998 | Screenshot_20220129_185831 |
|--|--|

### DIFF
--- a/wgpu/examples/shadow/main.rs
+++ b/wgpu/examples/shadow/main.rs
@@ -292,9 +292,17 @@ impl framework::Example for Example {
 
         let entity_uniform_size = mem::size_of::<EntityUniforms>() as wgpu::BufferAddress;
         let num_entities = 1 + cube_descs.len() as wgpu::BufferAddress;
-        let uniform_alignment =
-            device.limits().min_uniform_buffer_offset_alignment as wgpu::BufferAddress;
-        assert!(entity_uniform_size <= uniform_alignment);
+        // Make the `uniform_alignment` >= `entity_uniform_size` and aligned to `min_uniform_buffer_offset_alignment`.
+        let uniform_alignment = {
+            let alignment =
+                device.limits().min_uniform_buffer_offset_alignment as wgpu::BufferAddress;
+            let rem = entity_uniform_size % alignment;
+            if rem != 0 {
+                entity_uniform_size + alignment - rem
+            } else {
+                entity_uniform_size
+            }
+        };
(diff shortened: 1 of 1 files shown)


wgpu/examples/shadow/main.rs
```diff
@@ -292,9 +292,11 @@ impl framework::Example for Example {
 
         let entity_uniform_size = mem::size_of::<EntityUniforms>() as wgpu::BufferAddress;
         let num_entities = 1 + cube_descs.len() as wgpu::BufferAddress;
-        let uniform_alignment =
+        let mut uniform_alignment =
             device.limits().min_uniform_buffer_offset_alignment as wgpu::BufferAddress;
-        assert!(entity_uniform_size <= uniform_alignment);
+        // Make the `uniform_alignment` >= `entity_uniform_size` and aligned to `min_uniform_buffer_offset_alignment`.
+        uniform_alignment *=
+            (entity_uniform_size as f32 / uniform_alignment as f32).ceil() as wgpu::BufferAddress;
```
kvark, line 299: Let's not involve floating-point arithmetic here. Instead:
```rust
let uniform_stride = {
  let alignment = device.limits().min_uniform_buffer_offset_alignment;
  let rem = entity_uniform_size % alignment;
  if rem != 0 {
    entity_uniform_size + alignment - rem
  } else {
    entity_uniform_size
  }
};
```

kvark (dismissed):
Thank you for addressing this!

wgpu/examples/shadow/main.rs
jinleili: If I'm not mistaken, `entity_uniform_size` is an integer multiple of `alignment` and greater than 1, `uniform_stride` result will be wrong?

kvark: > If I'm not mistaken, when entity_uniform_size is an integer multiple of alignment and greater than 1, uniform_stride result will be wrong?

I don't think so. In this case `rem == 0`, which means `uniform_stride == entity_uniform_size`, which is what we want

cwfitzgerald (approved):

PR #2434, opened 2022-01-29
Support to create surface from visual on Windows

xiaopengli89:
**Description**
1. Support to create surface from `DirectComposition`  visual
2. Add MPO feature to AdapterInfo

**Testing**
Render content into `DirectComposition` visual to compose in a window.

Run `water` example with `DX12` backend, `AdapterInfo.mpo` will show wether hardware supports multiplane overlays or not, if hardware supports MPO, surface will present independent with dwm, which means app's fps may be higher than dwm.

mpo
fps

### DIFF
--- a/wgpu-hal/src/dx12/adapter.rs
+++ b/wgpu-hal/src/dx12/adapter.rs
@@ -1,10 +1,7 @@
-use super::{conv, HResult as _};
-use crate::dx12::SurfaceTarget;
+use super::{conv, HResult as _, SurfaceTarget};
 use std::{mem, sync::Arc, thread};
-use winapi::shared::dxgi1_3;
-use winapi::shared::minwindef::TRUE;
 use winapi::{
-    shared::{dxgi, dxgi1_2, dxgi1_5, minwindef, windef, winerror},
+    shared::{dxgi, dxgi1_2, dxgi1_3, dxgi1_5, minwindef, windef, winerror},
     um::{d3d12, d3d12sdklayers, winuser},
 };
 
@@ -121,7 +118,7 @@ impl super::Adapter {
                             if let Ok(output2) =
                                 output.cast::<dxgi1_3::IDXGIOutput2>().into_result()
                             {
-                                info.mpo = output2.SupportsOverlays() == TRUE;
+                                info.mpo = output2.SupportsOverlays() == minwindef::TRUE;
                                 output2.destroy();
                             }
                             output.destroy();
--- a/wgpu-hal/src/dx12/adapter.rs
+++ b/wgpu-hal/src/dx12/adapter.rs
@@ -1,7 +1,7 @@
 use super::{conv, HResult as _, SurfaceTarget};
 use std::{mem, sync::Arc, thread};
 use winapi::{
-    shared::{dxgi, dxgi1_2, dxgi1_3, dxgi1_5, minwindef, windef, winerror},
+    shared::{dxgi, dxgi1_2, dxgi1_5, minwindef, windef, winerror},
     um::{d3d12, d3d12sdklayers, winuser},
 };
 
@@ -90,7 +90,7 @@ impl super::Adapter {
 
         let mut workarounds = super::Workarounds::default();
 
-        let mut info = wgt::AdapterInfo {
+        let info = wgt::AdapterInfo {
             backend: wgt::Backend::Dx12,
             name: device_name,
             vendor: desc.VendorId as usize,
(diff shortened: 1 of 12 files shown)


wgpu-hal/src/dx12/adapter.rs
```diff
@@ -1,5 +1,8 @@
 use super::{conv, HResult as _};
+use crate::dx12::SurfaceTarget;
 use std::{mem, sync::Arc, thread};
+use winapi::shared::dxgi1_3;
+use winapi::shared::minwindef::TRUE;
```
kvark, line 5: please add these to the existing `use winapi:{}` to keep the same style

wgpu-hal/src/dx12/adapter.rs
```diff
@@ -1,5 +1,8 @@
 use super::{conv, HResult as _};
+use crate::dx12::SurfaceTarget;
```
kvark, line 2: please add this to the one above, this type is in `super::{}`

wgpu-types/src/lib.rs
```diff
@@ -975,6 +975,8 @@ pub struct AdapterInfo {
     pub device_type: DeviceType,
     /// Backend used for device
     pub backend: Backend,
+    /// MPO(Multiple Plane Overlay) supported
+    pub mpo: bool,
```
kvark, line 979: why does this flag need to be on adapterinfo? Is it applicable to other backends to some way? Is it actionable by the user?

wgpu-hal/src/dx12/mod.rs
```diff
@@ -44,6 +44,7 @@ mod view;
 use arrayvec::ArrayVec;
 use parking_lot::Mutex;
 use std::{borrow::Cow, ffi, mem, num::NonZeroU32, ptr, sync::Arc};
+use winapi::um::dcomp::IDCompositionVisual;
```
kvark, line 47: similarly, please add to the existing use statement

wgpu-hal/src/dx12/instance.rs
```diff
@@ -1,4 +1,5 @@
 use super::HResult as _;
+use crate::dx12::SurfaceTarget;
```
kvark, line 2: please add to `super::{}`

wgpu-hal/src/dx12/adapter.rs
```diff
@@ -103,8 +106,36 @@ impl super::Adapter {
...
             },
+            mpo: false,
         };
 
+        unsafe {
+            if let Ok(adapter1) = adapter.cast::<dxgi::IDXGIAdapter1>().into_result() {
+                for i in 0.. {
+                    let mut output = native::WeakPtr::<dxgi::IDXGIOutput>::null();
+                    match adapter1
+                        .EnumOutputs(i, output.mut_void() as _)
```
kvark, line 117: based on this code, `mpo` would be a property of the pair (adapter, surface), not just the adapter alone (i.e. surface is presented via one of these outputs). So perhaps we should move `mpo` flag to the `SurfaceCapabilities`?

kvark (changes requested):
I think this is a good direction, thank you for hacking on this!
Just a couple of questions and a few stylistic notes.

wgpu-hal/src/dx12/adapter.rs
xiaopengli89: [IDXGIAdapter::EnumOutputs](https://docs.microsoft.com/en-us/windows/win32/api/dxgi/nf-dxgi-idxgiadapter-enumoutputs) says that the `EnumOutputs` method enumerate adapter (video card) outputs, which I don't think has anything to do with the surface

wgpu-types/src/lib.rs
xiaopengli89: MPO is a hardware feature and not actionable by the user, if hardware supports MPO and the surface meets certain conditions, window compositor will display the surface as an MPO automatically.

xiaopengli89:
This image describes how MPO works.

wgpu-hal/src/dx12/adapter.rs
xiaopengli89: I'm confused, [vkCreateDisplayPlaneSurfaceKHR](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkCreateDisplayPlaneSurfaceKHR.html) and [DXGI_SWAP_CHAIN_FLAG_FOREGROUND_LAYER](https://docs.microsoft.com/en-us/windows/win32/api/dxgi/ne-dxgi-dxgi_swap_chain_flag) seem to explicitly create surfaces on MPOs, but even without them the window compositor will present surfaces as MPOs when possible.

wgpu-types/src/lib.rs
xiaopengli89: We can use [vkGetPhysicalDeviceDisplayPlanePropertiesKHR](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkGetPhysicalDeviceDisplayPlanePropertiesKHR.html) to get the number of display planes available on `vulkan` backend.

kvark: I started reading up on it and got lost a bit. So this is KHR_display: we have planes and displays, in addition to adapters and surfaces. wgpu currently doesn't have a way to expose planes or displays.

What value does this `mpo` flag bring to users? It seems like the actual API landscape is much more complicated than just a single boolean.

xiaopengli89: In my scenario, I can suggest users to use the app in full screen mode based on this flag, which can effectively increase frame rate and reduce latency.

kvark: ok, right. But even if the adapter itself supports MPO, it doesn't necessarily mean that the particular window/surface/adapter combination is going to support MPO, right?

xiaopengli89: Yes, I thought about it, I can integrate this detection function in other tools, not in wgpu, so this PR will only include the create_surface_from_visual feature.

kvark: that's a good way forward, thank you!
I'm totally positive about exposing some capability, it's just not clear what shape this would be, and we can follow-up on it.

kvark (approved):

PR #2443, opened 2022-01-30
Dx11 Backend

cwfitzgerald:
**Description**

At long last here is the PR that begins the Dx11 api. I'm making this PR early to make progress transparent and encourage comments as I build this out.

This first PR will aim to get us a working hello-triangle example, follow up PRs will build out more features. I'm trying to keep PRs as small as reasonable due to the hard lesseons we've learned in the past.

**Testing**

Doesn't work... yet :)


cwfitzgerald:
Alright, this is ready for review, though dependent on https://github.com/gfx-rs/d3d12-rs/pull/37.

This factors out a lot of the DXGI code into common code, as well as implementing DX11 adapter enumeration. The DX11 backend is behind an environment flag so that it will not show up unless you externally as for it.

The commits are self contained, but contain a dependency override on a fork, so it's up to you if to squash.

wgpu-hal/Cargo.toml
```diff
@@ -75,6 +76,7 @@ libloading = { version = "0.7", optional = true }
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3", features = ["libloaderapi", "windef", "winuser", "dcomp"] }
 native = { package = "d3d12", version = "0.4.1", features = ["libloading"], optional = true }
+libloading = { version = "0.7", optional = true }
```
kvark, line 79: shouldn't be needed - it's already a dependency for non-wasm32

wgpu-hal/src/auxil/dxgi/conv.rs
```diff
@@ -0,0 +1,175 @@
+use winapi::shared::dxgiformat;
+
+pub fn map_texture_format(format: wgt::TextureFormat) -> dxgiformat::DXGI_FORMAT {
```
kvark, line 3: nice to see code sharing here!

wgpu-hal/src/dx11/adapter.rs
```diff
@@ -0,0 +1,274 @@
...
+    }
+}
+
+impl super::Adapter {
+    pub(super) fn expose(
+        instance: &super::library::D3D11Lib,
+        adapter: native::DxgiAdapter,
+    ) -> Option<crate::ExposedAdapter<super::Api>> {
+        use d3dcommon::{
+            D3D_FEATURE_LEVEL_10_0 as FL10_0, D3D_FEATURE_LEVEL_10_1 as FL10_1,
```
kvark, line 35: this is understandable, but a bit unfortunate
perhaps we'll be better off just using the full names?

wgpu-hal/src/dx11/adapter.rs
```diff
@@ -0,0 +1,274 @@
...
+            | wgt::Features::CLEAR_TEXTURE
+            | wgt::Features::TEXTURE_FORMAT_16BIT_NORM
+            | wgt::Features::ADDRESS_MODE_CLAMP_TO_ZERO;
+        let mut downlevel =
+            wgt::DownlevelFlags::BASE_VERTEX | wgt::DownlevelFlags::READ_ONLY_DEPTH_STENCIL;
+
+        // Features from queries
+        downlevel.set(
+            wgt::DownlevelFlags::NON_POWER_OF_TWO_MIPMAPPED_TEXTURES,
+            d3d9_features.FullNonPow2TextureSupported == 1,
```
kvark, line 100: whaaat, dx9 feature checks?

wgpu-hal/src/dx11/adapter.rs
```diff
@@ -0,0 +1,274 @@
...
+
+        //
+        // Fill out limits and alignments
+        //
+
+        let max_texture_dimension_2d = match feature_level {
+            FL9_1 | FL9_2 => 2048,
+            FL9_3 => 4096,
+            FL10_0 | FL10_1 => 8192,
+            _ => 16384,
```
kvark, line 148: we should be able to use the constants here, like `D3D11_REQ_TEXTURE2D_U_OR_V_DIMENSION`

wgpu-hal/src/dx11/adapter.rs
```diff
@@ -0,0 +1,274 @@
...
+            FL9_1..=FL9_3 => wgt::ShaderModel::Sm2,
+            FL10_0 | FL10_1 => wgt::ShaderModel::Sm4,
+            _ => wgt::ShaderModel::Sm5,
+        };
+
+        let device_info = wgt::AdapterInfo {
+            name: String::new(),
+            vendor: 0,
+            device: 0,
+            device_type: match d3d11_features2.UnifiedMemoryArchitecture {
```
kvark, line 238: we should be able to detect WARP as well

kvark (approved):
Very nice!
It looks mergable, I left a few notes in case you are going to do another iteration.

wgpu-hal/src/dx11/adapter.rs
cwfitzgerald: I'll add that in a follow up that does name/type detection correctly

wgpu-hal/src/dx11/adapter.rs
cwfitzgerald: I almost removed the aliases when I just had the feature code as it wasn't saving us much, but with limits being able to do FL9_1..=FL10_0 tersely was way too useful so I ended up keeping them

wgpu-hal/src/dx11/adapter.rs
cwfitzgerald: Especially new hal, being able to run on dx9 hardware might actually be a possibility :tada:

PR #2451, opened 2022-02-04
Allow non struct buffers in wgsl

IcanDivideBy0:
Thanks to https://github.com/gfx-rs/naga/commit/75692d3, it is now possible to use arrays top level components of a shader, this PR integrates the latest changes from naga to wgpu.

So far, it seems to work fine, but i'm still having an issue: when using the `arrayLength` builtin function on such an array (runtime sized), i have the following error:
```

Caused by:
      note: label = `Particles pipeline`
    Internal error: module is not validated properly: array length expression

', /home/samuel/Workspace/wgpu/wgpu/src/backend/direct.rs:2273:5
```

I've tried to inspect the generated spirv but i'm not really familiar with this, any help appreciated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1041,7 +1041,7 @@ dependencies = [
 [[package]]
 name = "naga"
 version = "0.8.0"
-source = "git+https://github.com/gfx-rs/naga?rev=81dc674#81dc67402a743b4dc6b2d24c0d306cd18799238b"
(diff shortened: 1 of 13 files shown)


kvark (approved):
Nice!

IcanDivideBy0:
https://github.com/gfx-rs/naga/pull/1717 definitely fixed the issue with `arrayLength`!

kvark (approved):

Issue #2452, opened 2022-02-04
M1 in macOS incorrectly reports supported compressed texture formats


superdump:
**Description**
Apple M1 GPUs support ASTC, BC, EAC/ETC2, and even PVRTC compressed texture formats but on BC is reported as being supported.

**Repro steps**
Inspect the `Features` returned from `Adapter::features()` for an M1 GPU in macOS.

**Expected vs observed behavior**
- Expected
  - `TEXTURE_COMPRESSION_BC`, `TEXTURE_COMPRESSION_ASTC_LDR`, and `TEXTURE_COMPRESSION_ETC2` are all reported as being supported.
- Observed
  - Only `TEXTURE_COMPRESSION_BC` is reported as being supported.

**Platform**
Tested with wgpu 0.8 and master on an M1 Max running macOS Monterey 12.2

superdump:
I dug into this a bit and with some hacking to report all of ASTC, BC, EAC/ETC2 as being supported, I have been able to make use of all on M1. I did this because at first it was unclear from Apple documentation exactly what M1 should support.

It looks like the checks in `metal-rs` probably need to be updated, but also the way the features are constructed in `wgpu` look like they could be improved as they seem to always report BC: https://github.com/gfx-rs/wgpu/blob/766c6cda1917c1265fa8a5f610ef7516d4314d1b/wgpu-hal/src/metal/adapter.rs#L1013

superdump:
I didn't understand the relation between `MTLGPUFamily` and `MTLFeatureSet` and was trying to understand where the Apple7 family fit into the feature sets, but the table at the beginning of: https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf clarifies that there is not yet any corresponding feature set for Apple7 (which includes A14 and M1 GPUs.)

However, I see elsewhere in the code that `device.supports_family(MTLGPUFamily::Apple7)` is used so I think I can just make use of that.

superdump:
I've made a PR that fixes both the issues detailed above: https://github.com/gfx-rs/wgpu/pull/2453

In metal-rs, it seems that it is not currently possible to correctly detect compressed texture format support using the following methods:
```rust
    pub fn supports_pvrtc_pixel_formats(&self) -> bool {
        self.os() != OS::macOS
    }

    pub fn supports_eac_etc_pixel_formats(&self) -> bool {
        self.os() != OS::macOS
    }

    pub fn supports_astc_pixel_formats(&self) -> bool {
[...]

PR #2453, opened 2022-02-04
Fix M1 in macOS incorrectly reports supported compressed texture formats

superdump:
**Connections**
[M1 in macOS incorrectly reports supported compressed texture formats](https://github.com/gfx-rs/wgpu/issues/2452)

**Description**
The metal backend in wgpu-hal was incorrectly detecting support for ASTC and EAC/ETC2 on macOS as the M1 line of GPUs supports these formats as well as BC. It was also misrepresenting the support in the values returned from `Adapter::features()` as always support BC and never ASTC nor ETC2.

**Testing**
On an M1 running macOS, check whether the `Adapter::features()` includes all three of `TEXTURE_COMPRESSION_ASTC`, `TEXTURE_COMPRESSION_BC`, and `TEXTURE_COMPRESSION_ETC2`. I have tested this locally on an M1 Max running macOS.

### DIFF
--- a/wgpu-hal/src/metal/adapter.rs
+++ b/wgpu-hal/src/metal/adapter.rs
@@ -1014,7 +1014,6 @@ impl super::PrivateCapabilities {
         use wgt::Features as F;
 
         let mut features = F::empty()
-            | F::TEXTURE_COMPRESSION_BC
             | F::INDIRECT_FIRST_INSTANCE
             | F::MAPPABLE_PRIMARY_BUFFERS
             | F::VERTEX_WRITABLE_STORAGE
@@ -1023,6 +1022,10 @@ impl super::PrivateCapabilities {
             | F::POLYGON_MODE_LINE
             | F::CLEAR_TEXTURE
             | F::TEXTURE_FORMAT_16BIT_NORM;
+
+        features.set(F::TEXTURE_COMPRESSION_ASTC_LDR, self.format_astc);
+        features.set(F::TEXTURE_COMPRESSION_BC, self.format_bc);
+        features.set(F::TEXTURE_COMPRESSION_ETC2, self.format_eac_etc);
 
         features.set(F::DEPTH_CLIP_CONTROL, self.supports_depth_clip_control);
 
--- a/wgpu-hal/src/metal/adapter.rs
+++ b/wgpu-hal/src/metal/adapter.rs
@@ -720,10 +720,10 @@ impl super::PrivateCapabilities {
             format_bc: os_is_mac,
             format_eac_etc: !os_is_mac
                 // M1 in macOS supports EAC/ETC2
-                || device.supports_family(MTLGPUFamily::Apple7),
+                || (family_check && device.supports_family(MTLGPUFamily::Apple7)),
             format_astc: Self::supports_any(device, ASTC_PIXEL_FORMAT_FEATURES)
-                // A14/M1 always support ASTC
-                || device.supports_family(MTLGPUFamily::Apple7),
+                // A13/A14/M1 and later always support ASTC
+                || (family_check && device.supports_family(MTLGPUFamily::Apple6)),
             format_any8_unorm_srgb_all: Self::supports_any(device, ANY8_UNORM_SRGB_ALL),
             format_any8_unorm_srgb_no_write: !Self::supports_any(device, ANY8_UNORM_SRGB_ALL)
                 && !os_is_mac,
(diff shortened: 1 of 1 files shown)


wgpu-hal/src/metal/adapter.rs
```diff
@@ -718,8 +718,12 @@ impl super::PrivateCapabilities {
...
             format_b5: !os_is_mac,
             format_bc: os_is_mac,
-            format_eac_etc: !os_is_mac,
-            format_astc: Self::supports_any(device, ASTC_PIXEL_FORMAT_FEATURES),
+            format_eac_etc: !os_is_mac
+                // M1 in macOS supports EAC/ETC2
+                || device.supports_family(MTLGPUFamily::Apple7),
+            format_astc: Self::supports_any(device, ASTC_PIXEL_FORMAT_FEATURES)
+                // A14/M1 always support ASTC
+                || device.supports_family(MTLGPUFamily::Apple7),
```
cwfitzgerald, line 726: Shouldn't this be family 6? If ASTC is supported on ios family 2, then all gpu family 2+ should support this.

What family does the M1 report?

wgpu-hal/src/metal/adapter.rs
```diff
@@ -718,8 +718,12 @@ impl super::PrivateCapabilities {
             format_min_srgb_channels: if os_is_mac { 4 } else { 1 },
             format_b5: !os_is_mac,
             format_bc: os_is_mac,
-            format_eac_etc: !os_is_mac,
-            format_astc: Self::supports_any(device, ASTC_PIXEL_FORMAT_FEATURES),
+            format_eac_etc: !os_is_mac
+                // M1 in macOS supports EAC/ETC2
+                || device.supports_family(MTLGPUFamily::Apple7),
```
kvark, line 723: we can only call `supports_family` if `family_check` is true

wgpu-hal/src/metal/adapter.rs
kvark, line 726: same here

kvark (changes requested):
very nice!

wgpu-hal/src/metal/adapter.rs
superdump: > Shouldn't this be family 6? If ASTC is supported on ios family 2, then all gpu family 2+ should support this.

The `Self::supports_any(device, ASTC_PIXEL_FORMAT_FEATURES)` works for all `MTLGPUFamily` that have 'OS GPU families' defined in `MTLFeatureSet` but it misses those that don't. It looks like Apple6 doesn't have an 'OS GPU family' defined in `MTLFeatureSet` either. I think I saw that the feature sets are guaranteed to be supersets as the numbers get higher, but are the MTLGPUFamily necessarily supersets (i.e. Apple7 can only have features that are a superset of Apple6?) So this should either be:

- Apple7 is a superset of Apple6
```suggestion
                || device.supports_family(MTLGPUFamily::Apple6),
```
- Apple7 is not a superset of Apple6:
```suggestion
                || device.supports_family(MTLGPUFamily::Apple6)
                || device.supports_family(MTLGPUFamily::Apple7),
```

> What family does the M1 report?

M1 Max is Apple7. If someone has a 'plain' M1 or otherwise knows, it would be good to verify that all M1 are Apple7.

superdump: > More recent GPUs have higher version numbers and support larger feature sets. A higher GPU version is always a superset of an earlier version in the same GPU family. For details on the individual families and version numbers, see [MTLGPUFamily](https://developer.apple.com/documentation/metal/mtlgpufamily).

So yeah, Apple6 is correct. Good catch!

wgpu-hal/src/metal/adapter.rs
superdump: Done.

wgpu-hal/src/metal/adapter.rs
superdump: Done.

superdump: Fixed. Note that I left the ETC2 check as Apple7 because that explicit checks for not macOS or Apple7. Apple6 can't be used in macOS so it is covered by the first case.

superdump: This should be resolved but I'll leave it unresolved so you all can see the discussion and disagree if you think I made a mistake.

kvark (approved):

PR #2457, opened 2022-02-06
Add a :star: to the feature matrix of examples README

yutannihilation:
On searching for the example of `StagingBelt`, I found the skybox example uses it while the README doesn't say so. I think we can add a ⭐ here.


cwfitzgerald (approved):
Thanks!

cwfitzgerald:
Something is up with CI...

PR #2462, opened 2022-02-07
Don't recycle indices that reach EOL

kvark:
**Connections**
Related to https://github.com/gfx-rs/wgpu-native/pull/171
But also affects 64-bit systems.

**Description**
Once the epoch reaches its max value, we should never try to recycle it.

**Testing**
small unit test inside


cwfitzgerald (approved):
Huh, didn't realize we were using a freelist before for IDs

PR #2464, opened 2022-02-08
added id32 feature

caiiiycuk:
**Connections**
Related to https://github.com/gfx-rs/wgpu-native/pull/171. Added feature to enable 32 bit ids.

**Description**
If you activate --features id32bits, then wgpu will generate 32 bits ids, and so wgpu-native can be used on 32bit systems like emscripten.

**Testing**
Added unit test to check corner cases.

### DIFF
--- a/wgpu-core/src/id.rs
+++ b/wgpu-core/src/id.rs
@@ -1,9 +1,25 @@
 use crate::{Epoch, Index};
-use std::{cmp::Ordering, fmt, marker::PhantomData, num::NonZeroU64};
+use std::{cmp::Ordering, fmt, marker::PhantomData};
 use wgt::Backend;
 
+#[cfg(feature = "id32")]
+type IdType = u32;
+#[cfg(not(feature = "id32"))]
+type IdType = u64;
+#[cfg(feature = "id32")]
+type NonZeroId = std::num::NonZeroU32;
+#[cfg(not(feature = "id32"))]
+type NonZeroId = std::num::NonZeroU64;
+#[cfg(feature = "id32")]
+type ZippedIndex = u16;
+#[cfg(not(feature = "id32"))]
+type ZippedIndex = Index;
+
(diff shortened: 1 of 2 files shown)


caiiiycuk:
WIth current code max values are:
id 65535 
epoch 8191 
backend 7

caiiiycuk:
@kvark can you take a look? I didn't understand why rust complains if I don't specify repr(C) in 64bit case, and why deserialize fails...

wgpu-core/src/id.rs
```diff
@@ -1,9 +1,27 @@
...
+#[cfg(not(feature = "id32"))]
+type IdType = u64;
+#[cfg(feature = "id32")]
+type ZippedIndexType = u16;
+#[cfg(not(feature = "id32"))]
+type ZippedIndexType = Index;
+#[cfg(feature = "id32")]
+const INDEX_BITS: usize = 16;
+#[cfg(not(feature = "id32"))]
+const INDEX_BITS: usize = 32;
```
kvark, line 20: we can use (size_of)[https://doc.rust-lang.org/stable/std/mem/fn.size_of.html] for all the places instead of INDEX_BITS: `size_of::<IdType>()`. It's a const fn, so it can be used everywhere.

wgpu-core/src/id.rs
```diff
@@ -1,9 +1,27 @@
 use crate::{Epoch, Index};
-use std::{cmp::Ordering, fmt, marker::PhantomData, num::NonZeroU64};
+#[cfg(feature = "id32")]
+use std::num::NonZeroU32;
+#[cfg(not(feature = "id32"))]
+use std::num::NonZeroU64;
```
kvark, line 5: let's import it as `NonZeroIndex` in both cases, to avoid branching where we are using it

caiiiycuk:
Fixed

kvark (approved):
Nice work!

PR #2470, opened 2022-02-10
Fix `Surface::frome_uiview` can not guarantee set correct `contentScaleFactor`

jinleili:
**Description**
 On iOS, `Surface::from_uiview` may be called before the application initialization is complete (when using winit to manage windows and events),  `msg_send![view, window]` or `msg_send![window, screen]` will get null, then, view's `contentScaleFactor` remain default 1. The correct value should be between [2.0, 3.0], depending on physical device.

`from_uiview` is the same as `from_nsview` more than half of code, so, merged together.

**Testing**
Tested wgpu + winit on iOS, follow the implementation of [bevy's iOS demo](https://github.com/bevyengine/bevy/tree/latest/examples#setup-1).
Tested winit-free mode on iOS, use [wgpu-on-app](https://github.com/jinleili/wgpu-on-app).
Tested all Wgpu's examples on M1 Mac(macOS 12.2)

### DIFF
--- a/wgpu-hal/src/metal/surface.rs
+++ b/wgpu-hal/src/metal/surface.rs
@@ -74,10 +74,12 @@ impl super::Surface {
         }
     }
 
-    #[cfg(target_os = "ios")]
     #[allow(clippy::transmute_ptr_to_ref)]
-    pub unsafe fn from_uiview(uiview: *mut c_void) -> Self {
-        let view = uiview as *mut Object;
+    pub unsafe fn from_view(
+        view: *mut c_void,
+        delegate: Option<&HalManagedMetalLayerDelegate>,
+    ) -> Self {
+        let view = view as *mut Object;
         if view.is_null() {
             panic!("window does not have a valid contentView");
         }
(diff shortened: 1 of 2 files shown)


wgpu-hal/src/metal/surface.rs
```diff
@@ -74,88 +74,54 @@ impl super::Surface {
...
-            let bounds: CGRect = msg_send![view, bounds];
-            let () = msg_send![layer.as_ref(), setBounds: bounds];
-
-            let window: *mut Object = msg_send![view, window];
-            if !window.is_null() {
-                let scale_factor: CGFloat = msg_send![window, backingScaleFactor];
-                let () = msg_send![layer, setContentsScale: scale_factor];
+            let frame: CGRect = msg_send![main_layer, bounds];
+            let () = msg_send![new_layer.as_ref(), setFrame: frame];
+            #[cfg(target_os = "ios")]
```
kvark, line 98: can we use more type checked variant, i.e. `if cfg!(target_os = ...) {...} else if ... {...}`

kvark (commented):
Nice work!

wgpu-hal/src/metal/surface.rs
jinleili: `cfg!` is a runtime condition and `kCAGravityTopLeft` only available on macOS target, compiling to iOS target cause error:
```sh
error[E0425]: cannot find value `kCAGravityTopLeft` in this scope
   --> /Users/LiJinlei/Rust/forks/wgpu/wgpu-hal/src/metal/surface.rs:109:76
    |
109 |                 let () = msg_send![new_layer.as_ref(), setContentsGravity: kCAGravityTopLeft];
    |                                                                            ^^^^^^^^^^^^^^^^^ not found in this scope

For more information about this error, try `rustc --explain E0425`.
```

PR #2474, opened 2022-02-11
fix: blit should only draw one triangle

CurryPseudo:
**Description**
Original shader:
https://github.com/gfx-rs/wgpu/blob/48f8811f0f6643f112baa3a2a9e2c287017a022d/wgpu/examples/mipmap/blit.wgsl#L7-L22
`x` in [0, 1]
`y` in [0, 1]
`out.tex_coord` in [(0, 0), (2, 2)]
`out.position` in [(-1, -3), (3, 1)]
Clearly it is wrong, but it actually covers the whole render area([(-1, -1), (1, 1)]), and `tex_coord` happened to be right, so it worked.
image

**Testing**
Run the mipmap example, result is the same.
master branch:
image
this branch:
image

### DIFF
--- a/wgpu/examples/mipmap/main.rs
+++ b/wgpu/examples/mipmap/main.rs
@@ -98,7 +98,7 @@ impl Example {
                 targets: &[TEXTURE_FORMAT.into()],
             }),
             primitive: wgpu::PrimitiveState {
-                topology: wgpu::PrimitiveTopology::TriangleStrip,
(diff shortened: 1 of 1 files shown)


CurryPseudo:
That makes sense.
I changed to triangle list and use only 3 vertices, it works.

kvark (approved):
Thank you!

Issue #2371, opened 2022-01-09
Support ASTC HDR formats


cwfitzgerald:
**Is your feature request related to a problem? Please describe.**

Gives access to hdr compressed textures on mobile devices. 

**Describe the solution you'd like**

Implement it on top of metal and on top of VK_EXT_texture_compression_astc_hdr

**Describe alternatives you've considered**

Not doing this, there's no alternative formats.

cwfitzgerald:
D3D only supports BCn

PR #2477, opened 2022-02-13
Metal backend ASTC HDR formats support

jinleili:
**Connections**

**Description**
- Add all `metal.framework` defined HDR ASTC formats;
- Add a new `TEXTURE_COMPRESSION_ASTC_HDR` feature;
- Properly change metal surface `configure` fn;

**Testing**

Tested on MacBook Pro(M1 Max), iPhone 12 Pro(A14), using [wgpu-on-app](https://github.com/jinleili/wgpu-on-app).

Verified Intel cpu Mac and iOS simulator can not support HDR ASTC formats.

`.astc` test file generated by [ARM astc encoder](https://github.com/ARM-software/astc-encoder)

Loading the `.astc` file to create textures does not require any external crates:
```rust
let astc_data = &include_bytes!("../../assets/6x6.astc")[..];
let size = wgpu::Extent3d {width: 4096,height: 2048,depth_or_array_layers: 1};
let texture = device.create_texture_with_data(queue, &wgpu::TextureDescriptor {
		size, mip_level_count: 1, sample_count: 1,
		format: TextureFormat::Astc { block: AstcBlock::B6x6,  channel: AstcChannel::Hdr },
		usage: wgpu::TextureUsages::TEXTURE_BINDING | wgpu::TextureUsages::COPY_DST,
		label: None,
	},
	// need offset header size, witch is equal to one block size
	&astc_data[16..],
);
```

https://user-images.githubusercontent.com/1001342/153751508-04dc9cdd-f1cf-48dd-bf9a-b200808545d4.MOV


wgpu-hal/src/metal/surface.rs
```diff
@@ -208,10 +208,13 @@ impl crate::Surface<super::Api> for super::Surface {
         render_layer.set_pixel_format(self.raw_swapchain_format);
         render_layer.set_framebuffer_only(framebuffer_only);
         render_layer.set_presents_with_transaction(self.present_with_transaction);
+        // opt-in to Metal EDR
+        // EDR potentially more power used in display and more bandwidth, memory footprint.
+        render_layer.set_wants_extended_dynamic_range_content(
+            self.raw_swapchain_format == mtl::MTLPixelFormat::RGBA16Float,
+        );
```
cwfitzgerald, line 215: Is the only thing I'm not sure about, will defer to the mac experts here.

cwfitzgerald (approved):
LGTM besides one line which I'm unsure about

wgpu-hal/src/metal/surface.rs
jinleili: `wantsExtendedDynamicRangeContent` might only available on macOS 10.11+, but tested on iOS and didn't report any error , however setting this true was necessary to be able to display HDR contents

wgpu-types/src/lib.rs
```diff
@@ -1993,6 +1998,90 @@ pub enum TextureFormat {
     /// [`Features::TEXTURE_COMPRESSION_ASTC_LDR`] must be enabled to use this texture format.
     #[cfg_attr(feature = "serde", serde(rename = "astc-12x12-unorm-srgb"))]
     Astc12x12RgbaUnormSrgb,
+    /// 4x4 block compressed HDR texture. 16 bytes per block (8 bit/px). Complex pallet. floating-point RGBA.
```
kvark, line 2001: It seems to me that we should just encode it more combinatorically:
```rust
Astc { block: AstcBlock, channel: AstcChannel },
...
enum AstcBlock {
  B4x4,
  B5x4,
  ...
}
enum AstcChannel {
  Unorm,
  UnormSrgb,
  Hdr,
}
```
This would allow backends to easily skip Astc, and us to manage the API surface more efficiently

wgpu-hal/src/metal/surface.rs
```diff
@@ -208,10 +208,13 @@ impl crate::Surface<super::Api> for super::Surface {
...
 
-        // this gets ignored on iOS for certain OS/device combinations (iphone5s iOS 10.3)
```
kvark, line 212: let's keep the comment

wgpu-hal/src/metal/surface.rs
kvark: There is a concern of calling something that we aren't supposed to call.
Perhaps, we could only call this if the requested format was ever `mtl::MTLPixelFormat::RGBA16Float` for this surface?

jinleili: During the lifetime of a application,  user can switch surface `configure`, and since Metal EDR may consume more power, non-RGBA16Float needs to be set to false

kvark: Yes, that's why I wrote "if the requested format was ever ". I.e. the surface could keep a flag "HDR property is active", which we'd set once the format is configured to be `RGBA16Float` and keep on. Based on the flag we'd know when to call `set_wants_extended_dynamic_range_content`.

Alternatively, the `PrivateCapabilities` could have a flag for this, saying whether or not we can call `set_wants_extended_dynamic_range_content` at all.

jinleili: I'll modify the code according to your suggestions

wgpu-types/src/lib.rs
jinleili: This suggestion is so great, now the removed LOC are larger than the added.

PR #2478, opened 2022-02-14
Fix get_timestamp_period type in docs

superdump:
**Description**
Minor documentation error where `Device::get_timestamp_period` is referenced but that method is on `Queue`.


cwfitzgerald (approved):
Thanks! That method migrated

PR #2481, opened 2022-02-15
Enable support for 3D textures

ryanw:
Currently when attempting to create a 3D texture it panics with: `internal error: entered unreachable code`

But simply adding D3 to the `match` seems to work just fine. I've tested as far as creating a 3D texture and inspecting it using RenderDoc.

Here's an example of creating a texture that is fixed by this change:

```
device.create_texture(&wgpu::TextureDescriptor {
        label: Some("Voxel texture"),
        size: wgpu::Extent3d {
                width: 128,
                height: 128,
                depth_or_array_layers: 128,
        },
        mip_level_count: 1,
        sample_count: 1,
        dimension: wgpu::TextureDimension::D3,
        format: wgpu::TextureFormat::Rgba8UnormSrgb,
        usage: wgpu::TextureUsages::TEXTURE_BINDING
                | wgpu::TextureUsages::COPY_SRC
                | wgpu::TextureUsages::COPY_DST
                | wgpu::TextureUsages::RENDER_ATTACHMENT,
});
```

A 3D texture inside RenderDoc:
image

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -730,7 +730,7 @@ impl<A: HalApi> Device<A> {
             let dimension = match desc.dimension {
                 wgt::TextureDimension::D1 => wgt::TextureViewDimension::D1,
                 wgt::TextureDimension::D2 => wgt::TextureViewDimension::D2,
-                wgt::TextureDimension::D3 => unreachable!(),
+                wgt::TextureDimension::D3 => wgt::TextureViewDimension::D3,
             };
 
             let mut clear_views = SmallVec::new();


wgpu-core/src/device/mod.rs
```diff
@@ -730,7 +730,7 @@ impl<A: HalApi> Device<A> {
             let dimension = match desc.dimension {
                 wgt::TextureDimension::D1 => wgt::TextureViewDimension::D1,
                 wgt::TextureDimension::D2 => wgt::TextureViewDimension::D2,
-                wgt::TextureDimension::D3 => unreachable!(),
+                wgt::TextureDimension::D3 => wgt::TextureViewDimension::D3,
```
kvark, line 733: Unfortunately, it's not that simple. The matter of fact is that 3D textures aren't really renderable in baseline WebGPU. Creating 2D views of them is not portable. See https://github.com/gpuweb/gpuweb/issues/2303 for a starting point.

In this case, we are missing a validation check that would see that the `RENDER_ATTACHMENT` usage is requested for a 3D texture and complain. This code path here is truly unreachable therefore.

kvark (changes requested):

wgpu-core/src/device/mod.rs
kvark: I filed https://github.com/gpuweb/gpuweb/pull/2603 to address this in the spec. We'll still need to add the relevant validation code on our side.

FYI, your workaround should be making it a 2D array texture instead of 3D.

PR #2493, opened 2022-02-19
Update dev deps

rukai:
**Description**
Updated deps in your examples makes it easier for users to use as a base.
Looks like some of these are also used for CI infrastructure (png)

**Testing**
I think CI should catch any issues with the png changes as that seems to be the main user of png images.
I think the bunnymark example is broken on the master branch on my machine so that didnt help with testing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "adler"
+version = "1.0.2"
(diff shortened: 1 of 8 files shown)


rukai:
I reverted the rand changes as it caused a chain of suffering that should be resolved once we bump the msrv.

cwfitzgerald (approved):
Party in the dep tree!

PR #2495, opened 2022-02-19
Use cargo nextest for running our tests

cwfitzgerald:
**Connections**

None

**Description**

This uses cargo-nextest to run our tests. 

This provides a couple of benefits:
- Runs each test in its own process, preventing issues with EGL's single threadedness.
- Runs all tests individually and in parallel.

**Testing**

It is.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,7 @@ jobs:
             tool: clippy
             kind: local
             backends: dx12 # dx11
+            nextest_url: https://get.nexte.st/latest/windows-tar
 
           - name: Windows Nightly x86_64
             os: windows-2019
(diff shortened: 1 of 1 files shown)


cwfitzgerald:
See https://github.com/gfx-rs/wgpu/runs/5257083717?check_suite_focus=true for a good example of how clear this new test framework makes errors extremely clear, even if we didn't see them before.

.github/workflows/ci.yml
```diff
@@ -139,6 +141,12 @@ jobs:
         with:
           key: ${{ matrix.target }}-a # suffix for cache busting
 
+      - name: download nextest
+        if: matrix.kind == 'local'
+        shell: bash
+        run: |
+          curl -LsSf ${{ matrix.nextest_url }} | tar xzf - -C ${CARGO_HOME:-~/.cargo}/bin
```
kvark, line 148: this is a bit scary :/

kvark (approved):

.github/workflows/ci.yml
cwfitzgerald: Yeah not the couthest thing in the world, but not really any better options.

PR #2496, opened 2022-02-20
vulkan: HDR ASTC formats support

jinleili:
**Connections**

**Description**
- Add `astc_hdr` field to PhysicalDeviceFeatures to hold `PhysicalDeviceTextureCompressionASTCHDRFeaturesEXT` when this feature is supported;
- Set TEXTURE_COMPRESSION_ASTC_HDR feature if `astc_hdr` is not none, then use this feature  to enable extension;
- Enable `ExtSwapchainColorspaceFn` during create `Instance`  to provide wide color gamut support;
- Choose `EXTENDED_SRGB_LINEAR_EXT`  color space when creating the swapchain;
	- HDR ASTC formarts need linear color space to display, and vulkan swapchain for Android only supports `DISPLAY_P3_NONLINEAR_EXT` and `EXTENDED_SRGB_LINEAR_EXT`
	- Maybe better way is using `get_physical_device_surface_formats` detect color space to choose, but need `PhysicalDevice` as a parameter that `fn create_swapchain` cannot provide, that needs to change a lot of code.

**Testing**
Running on maOS 12.2.1 (M1 Max MacBook Pro, vulkan-portability) can perfectly display HDR ASTC texture via [wgpu-on-app demo](https://github.com/jinleili/wgpu-on-app), but on Ubundu 20.04, currently vulkan driver cannot support WCG color space:
```sh
Using llvmpipe (LLVM 13.0.0, 128 bits) (Vulkan)
WARNING: lavapipe is not a conformant vulkan implementation, testing use only.
thread 'main' panicked at 'Error in Surface::configure: requested format Rgba16Float is not in list of supported formats: [Bgra8Unorm, Bgra8UnormSrgb]', /home/parallels/Downloads/forks/wgpu/wgpu/src/backend/direct.rs:231:9
```

The vk backend on Android has weak support for HDR ASTC formats,  I'm tested a lot of Android devices(real devices + cloud devices), all cannot find `VK_EXT_texture_compression_astc_hdr` extension, screenshots below are of tested cloud Android devices with GL Extensions Viewer installed to detect the vulkan extensions:
cloud android device
|vulkan extensions (android 9)| vulkan extensions (android 10) |
|---|---|

### DIFF
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -197,6 +197,7 @@ impl Surface {
             wgt::TextureFormat::Rgba8UnormSrgb,
             wgt::TextureFormat::Bgra8Unorm,
             wgt::TextureFormat::Rgba8Unorm,
+            wgt::TextureFormat::Rgba16Float,
         ];
 
         let suf = A::get_surface(self);
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -25,6 +25,7 @@ pub struct PhysicalDeviceFeatures {
     robustness2: Option<vk::PhysicalDeviceRobustness2FeaturesEXT>,
     depth_clip_enable: Option<vk::PhysicalDeviceDepthClipEnableFeaturesEXT>,
     multiview: Option<vk::PhysicalDeviceMultiviewFeaturesKHR>,
+    astc_hdr: Option<vk::PhysicalDeviceTextureCompressionASTCHDRFeaturesEXT>,
 }
 
 // This is safe because the structs have `p_next: *mut c_void`, which we null out/never read.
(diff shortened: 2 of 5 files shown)


kvark:
Thank you for the PR!
Worth noting that while WebGPU upstream requires Rgba16float, we can't really require it on native. It has to be optional.

wgpu-core/src/instance.rs
```diff
@@ -197,6 +197,7 @@ impl Surface {
             wgt::TextureFormat::Rgba8UnormSrgb,
             wgt::TextureFormat::Bgra8Unorm,
             wgt::TextureFormat::Rgba8Unorm,
+            wgt::TextureFormat::Rgba16Float,
```
kvark, line 200: so this. We can't do this on native. Not all platforms support this format for presentation.

kvark (approved):

jinleili:
Supplementing Android test info, I changed to another remote Android devices platform, Snapdragon 865+ CPU devices basically all support HDR ASTC formats.

PR #2501, opened 2022-02-23
Improve shader support documentation

### LINKED ISSUE #2500: Clarify wgpu's shader support in docs

A common source of confusion I've seen is about `wgpu`'s shader support and their target platform support, e.g. some believe GLSL/SPIR-V shader support will be removed, others believe SPIR-V/GLSL translation is only available on native but not web targets, etc. I might be carrying a set of misconceptions as well. Currently the [Learn Wgpu](https://sotrh.github.io/learn-wgpu/) tutorial [states](https://github.com/sotrh/learn-wgpu/blob/36658596393913f5d780e6e026af2ac353544603/docs/beginner/tutorial3-pipeline/README.md?plain=1#L29),

> Note that, at the time of writing this, some WebGPU implementations also support SPIR-V, but it's just a temporary measure during the transition period to WGSL and will be removed

While this is can be true when describing WebGPU implementations in general, `wgpu` can continue to support more than what is in the WebGPU spec. As I understand it, WGSL, GLSL, and SPIR-V are all supported by `wgpu`, both on native and web targets, as `naga` can handle these translations as needed.

Additionally, [`wgpu::ShaderSource`](https://docs.rs/wgpu/0.12.0/wgpu/enum.ShaderSource.html) rustdocs could also benefit from mentioning which features need to be enabled. SPIR-V and GLSL require `features = ["spirv", "glsl"]` respectively. After linking that page, I noticed the type's docs were out of date on the current release, but that appears to have already been fixed by <https://github.com/gfx-rs/wgpu/pull/2315/> on master.

Add a section in the readme (and rustdocs?) that describes shader support such as the targets targets each work on (native/web), features required, etc. After clarifying, I will open an issue on the Learn Wgpu tutorial to expand on `wgpu`'s shader support specifically.

**Additional context**

Some cases of potential confusion I've seen in chats in the past month:

> wgpu never "supported" glsl, it accepts wgsl and spir-v (on native, if you enable that extension).

> My primary concern with sticking to glsl is that learn wgpu (It's a decent reference to remember details) claims that glsl is going to be removed from wgpu soon

> im just happy im still on wgpu old enough to still just use GLSL

cwfitzgerald:
**Connections**

Closes #2500 

**Description**

Improves the public facing documentation about our shader support.

**Testing**

@memoryruins should see if it's good enough :)

kvark I think this is uncontroversial, but wanted to get your double check on the cfg stuff. This seems to be the standard way of getting the "this is only available with X feature" and tokio and friends all use it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,21 @@ We have a [wiki](https://github.com/gfx-rs/wgpu/wiki) that serves as a knowledge
 
 :white_check_mark: = First Class Support — :ok: = Best Effort Support — :hammer_and_wrench: = Unsupported, but support in progress
 
+### Shader Support
+
+wgpu supports shaders in [WGSL](https://gpuweb.github.io/gpuweb/wgsl/), SPIR-V, and GLSL. 
+Both [HLSL](https://github.com/Microsoft/DirectXShaderCompiler) and [GLSL](https://github.com/KhronosGroup/glslang) 
+have compilers to target SPIR-V. All of these shader languages can be used with any backend, we
+will handle all of the conversion. Additionally, support for these shader inputs is not going away.
+
+While WebGPU does not support any shader language other than WGSL, we will automatically convert your
+non-WGSL shaders if you're running on WebGPU.
+
+WGSL is always supported by default, but GLSL and SPIR-V need features enabled to compile in support.
+
+To enable SPIR-V shaders, enable the `spirv` feature of wgpu.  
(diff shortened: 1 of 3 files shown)


kvark (approved):
Reads well, thank you!

memoryruins:
@cwfitzgerald you succinctly covered it all, even the rustdocs! the comment about HLSL->SPIR-V compilers is a great addition. ❤️

PR #2502, opened 2022-02-23
metal: set `max_buffer_size`  by the correct physical device restriction

jinleili:
**Connections**
The wgpu matrix room mentions the 256MB restriction on macOS several times:
> ...
>  [wgpu won't stop you, but mac doens't like buffers above 256mb or so](vector://vector/webapp/#/room/!FZyQrssSlHEZqrYcOb:matrix.org/$-BQNuKgMznywXC9zaBW0wes7uFiZl84RRi1COnd-dag)
>  [they don't like buffers over 256Mb](vector://vector/webapp/#/room/!FZyQrssSlHEZqrYcOb:matrix.org/$pnvyZQEyAUFsF9KJzRLA19t8o5ZCu7MqKEOknjpT8bA)
>  [Consider it to be 256mb to be safe](vector://vector/webapp/#/room/!XFRnMvAfptAHthwBCx:matrix.org/$ReHNqwdVupEsa0oUGUffAP0OXRAnZX_-JWxyCHlZL38)
>  ...

**Description**
But starting from iOS 12 and macOS 10.14, there is a [ `maxBufferLength` property ](https://developer.apple.com/documentation/metal/mtldevice/2966563-maxbufferlength?language=objc) can used to query max buffer size.

In addition, since `[view convertRect: toCoordinateSpace:]` only affects x, y coordinates, deleting it from surface `fn dimentions` does not affect correctness, getting `bounds` and `contentsScale` from metalLayer is guaranteed.

**Testing**
Tested on these devices:
- M1 max MacBook Pro (64G Memory), max_buffer_size = **36864** MB = **36** GB
- Intel MacBook Pro (2018, 32G Memorry), max_buffer_size = **3072** MB = **3** GB
- iPhone 6 Plus, max_buffer_size = **256** MB
- iPad Mini 4, max_buffer_size = **497** MB
- iPhone 8 plus, max_buffer_size = **747** MB
- iPad Pro 2018, max_buffer_size = **947** MB
- iPhone 12 Pro, max_buffer_size = **1433** MB

### DIFF
--- a/wgpu-hal/src/metal/adapter.rs
+++ b/wgpu-hal/src/metal/adapter.rs
@@ -1,5 +1,9 @@
 use mtl::{MTLFeatureSet, MTLGPUFamily, MTLLanguageVersion};
-use objc::{class, msg_send, sel, sel_impl};
+use objc::{
+    class, msg_send,
+    runtime::{BOOL, YES},
+    sel, sel_impl,
+};
 use parking_lot::Mutex;
 use wgt::{AstcBlock, AstcChannel};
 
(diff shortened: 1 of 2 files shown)


kvark (approved):
Looks clean, thank you!

PR #2504, opened 2022-02-23
Many Steps Towards GL Testing Working

cwfitzgerald:
**Connections**

Multiple issues have caused GL testing to not work in CI. This is the next step towards that.

**Description**

So first, I apologize, this PR got _way_ out of control way quickly. I started with the goal of "fix up GL testing" but it turns out there were quite a few blockers here, so I had to resolve those... and suddenly I have 12 commits in the PR. I have more coming, but needed to cut it off before it got crazy.

Every commit is freestanding and should be reviewed individually, and I will make whatever changes using fixups.

**Testing**

Vk and Dx12 tests pass, GL still don't pass, but a lot more of them do than did before.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -2735,10 +2735,13 @@ impl<A: HalApi> Device<A> {
         let format_desc = format.describe();
         self.require_features(format_desc.required_features)?;
 
-        if self
+        let using_device_features = self
             .features
-            .contains(wgt::Features::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES)
-        {
+            .contains(wgt::Features::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES);
+        // If we're running downlevel, we need to manually ask the backend what we can use as we can't trust WebGPU.
+        let downlevel = !self.downlevel.is_webgpu_compliant();
+
+        if using_device_features || downlevel {
             Ok(adapter.get_texture_format_features(format))
         } else {
             Ok(format_desc.guaranteed_format_features)
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -28,49 +28,16 @@ pub struct FailedLimit {
 }
 
 fn check_limits(requested: &wgt::Limits, allowed: &wgt::Limits) -> Vec<FailedLimit> {
-    use std::cmp::Ordering;
     let mut failed = Vec::new();
 
-    macro_rules! compare {
-        ($name:ident, $ordering:ident) => {
-            match requested.$name.cmp(&allowed.$name) {
-                Ordering::$ordering | Ordering::Equal => (),
-                _ => failed.push(FailedLimit {
-                    name: stringify!($name),
-                    requested: requested.$name,
-                    allowed: allowed.$name,
-                }),
-            }
-        };
-    }
+    requested.check_limits_with_fail_fn(allowed, false, |name, requested, allowed| {
+        failed.push(FailedLimit {
+            name,
+            requested,
+            allowed,
+        })
+    });
 
-    compare!(max_texture_dimension_1d, Less);
-    compare!(max_texture_dimension_2d, Less);
-    compare!(max_texture_dimension_3d, Less);
-    compare!(max_texture_array_layers, Less);
-    compare!(max_bind_groups, Less);
-    compare!(max_dynamic_uniform_buffers_per_pipeline_layout, Less);
-    compare!(max_dynamic_storage_buffers_per_pipeline_layout, Less);
-    compare!(max_sampled_textures_per_shader_stage, Less);
-    compare!(max_samplers_per_shader_stage, Less);
-    compare!(max_storage_buffers_per_shader_stage, Less);
-    compare!(max_storage_textures_per_shader_stage, Less);
-    compare!(max_uniform_buffers_per_shader_stage, Less);
-    compare!(max_uniform_buffer_binding_size, Less);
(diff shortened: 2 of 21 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -2735,10 +2748,13 @@ impl<A: HalApi> Device<A> {
         let format_desc = format.describe();
         self.require_features(format_desc.required_features)?;
 
-        if self
+        let using_device_features = self
             .features
-            .contains(wgt::Features::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES)
-        {
+            .contains(wgt::Features::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES);
+        // If we're running downlevel, we need to manually ask the backend what we can use as we can't trust WebGPU.
```
kvark, line 2754: That's a bit concerning. If we can't respect WebGPU guarantees on texture formats, then we'll need to set downlevel baseline for them. This code implies that the downlevel baseline is basically zero capabilities, and it would make peoples lives really miserable

wgpu-core/src/instance.rs
```diff
@@ -28,49 +28,16 @@ pub struct FailedLimit {
...
-                Ordering::$ordering | Ordering::Equal => (),
-                _ => failed.push(FailedLimit {
-                    name: stringify!($name),
-                    requested: requested.$name,
-                    allowed: allowed.$name,
-                }),
-            }
-        };
-    }
+    requested.check_limits_with_fail_fn(allowed, false, |name, requested, allowed| {
```
kvark, line 33: how does this work with some limits having different types from the others? i.e. `u32` vs `u64`

wgpu-core/src/instance.rs
```diff
@@ -878,7 +845,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             .map_err(|_| InvalidAdapter)
     }
 
-    pub fn adapter_downlevel_properties<A: HalApi>(
+    pub fn adapter_downlevel_capabilities<A: HalApi>(
```
kvark, line 848: nice rename!

wgpu-hal/src/gles/conv.rs
```diff
@@ -8,7 +8,7 @@ impl super::AdapterShared {
 
         let (internal, external, data_type) = match texture_format {
             Tf::R8Unorm => (glow::R8, glow::RED, glow::UNSIGNED_BYTE),
-            Tf::R8Snorm => (glow::R8, glow::RED, glow::BYTE),
+            Tf::R8Snorm => (glow::R8_SNORM, glow::RED, glow::BYTE),
```
kvark, line 11: holy cow, I didn't know this is a thing in GL

wgpu-hal/src/gles/conv.rs
```diff
@@ -74,14 +74,14 @@ impl super::AdapterShared {
...
-            | Tf::Bc6hRgbSfloat
-            | Tf::Bc7RgbaUnorm
-            | Tf::Bc7RgbaUnormSrgb => unimplemented!(),
+            Tf::Bc4RUnorm => (glow::COMPRESSED_RED_RGTC1, glow::RED, 0),
+            Tf::Bc4RSnorm => (glow::COMPRESSED_SIGNED_RED_RGTC1, glow::RED, 0),
+            Tf::Bc5RgUnorm => (glow::COMPRESSED_RG_RGTC2, glow::RG, 0),
+            Tf::Bc5RgSnorm => (glow::COMPRESSED_SIGNED_RG_RGTC2, glow::RG, 0),
+            Tf::Bc6hRgbUfloat => (glow::COMPRESSED_RGB_BPTC_UNSIGNED_FLOAT, glow::RGB, 0),
+            Tf::Bc6hRgbSfloat => (glow::COMPRESSED_RGB_BPTC_SIGNED_FLOAT, glow::RGB, 0),
+            Tf::Bc7RgbaUnorm => (glow::COMPRESSED_RGBA_BPTC_UNORM, glow::RGB, 0),
```
kvark, line 83: this (and the one blow) should have `RGBA` instead of `RGB`

wgpu-hal/src/gles/queue.rs
```diff
@@ -452,9 +452,16 @@ impl super::Queue {
                         _ => unreachable!(),
                     }
                 } else {
-                    let bytes_per_image =
-                        copy.buffer_layout.rows_per_image.map_or(1, |rpi| rpi.get())
-                            * copy.buffer_layout.bytes_per_row.map_or(1, |bpr| bpr.get());
+                    let bytes_per_row = copy.buffer_layout.bytes_per_row.map_or(
+                        copy.size.width / format_info.block_dimensions.0 as u32,
```
kvark, line 456: this seems wrong? the equation here would return the block count per row, not bytes per row

kvark (changes requested):
Looks very nice, and thank you for expanding the test coverage!
Just a few things I spotted that we need to fix

wgpu-hal/src/gles/queue.rs
cwfitzgerald: Indeed

wgpu-core/src/instance.rs
cwfitzgerald: It doesn't, but the existing code didn't either. I think we will just migrate all callbacks to be u64 and just cast.

wgpu-core/src/device/mod.rs
cwfitzgerald: I agree, though I need to more fully establish the exact features allowed by various backends before I start to build that out. Let's leave this as is and I'll address it in a follow up -- I'll file an issue.

kvark (approved):

PR #2508, opened 2022-02-24
Re-enable GL testing in CI

cwfitzgerald:
**Connections**

This is a follow up to #2504, currently in draft state to get CI to run. Will rebase once it is merged in.

**Description**
GL testing back on CI!

**Testing**
It is.

### DIFF
--- a/wgpu/tests/clear_texture.rs
+++ b/wgpu/tests/clear_texture.rs
@@ -315,7 +315,7 @@ fn clear_texture_2d_bc() {
     initialize_test(
         TestParameters::default()
             .features(wgpu::Features::CLEAR_TEXTURE | wgpu::Features::TEXTURE_COMPRESSION_BC)
-            .specific_failure(Some(wgpu::Backends::GL), None, None, true),
+            .specific_failure(Some(wgpu::Backends::GL), None, Some("ANGLE"), true), // https://bugs.chromium.org/p/angleproject/issues/detail?id=7056
         |ctx| {
             clear_texture_tests(&ctx, TEXTURE_FORMATS_BC, false, true);
         },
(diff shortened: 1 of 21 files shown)


kvark (approved):

cwfitzgerald:
Re-requesting review on this, I've rebased against merged master. Same deal commits are free-standing.

Most should be pretty uncontroversial, but I added a downlevel flag, and added a validation canary to wgpu-hal, which you should take a look at.

kvark (approved):

wgpu/tests/clear_texture.rs
```diff
@@ -239,20 +239,28 @@ fn single_texture_clear_test(
...
 fn clear_texture_tests(
     ctx: &TestingContext,
     formats: &[wgpu::TextureFormat],
     supports_1d: bool,
     supports_3d: bool,
 ) {
     for &format in formats {
+        let desc = format.describe();
+        let rounded_width = round_up(64, desc.block_dimensions.0 as u32);
+        let rounded_height = round_up(64, desc.block_dimensions.1 as u32);
```
kvark, line 255: why are we rounding?
WebGPU allows creating any texture sizes regardless of the block size

kvark (changes requested):

wgpu/tests/clear_texture.rs
cwfitzgerald: What how does that work with vk requires textures a multiple if the block size?

cwfitzgerald: Per https://gpuweb.github.io/gpuweb/#dom-gpudevice-createtexture at least currently, images need to be the block size.

> descriptor.[size](https://gpuweb.github.io/gpuweb/#dom-gputexturedescriptor-size).[width](https://gpuweb.github.io/gpuweb/#extent3d-width) must be multiple of [texel block width](https://gpuweb.github.io/gpuweb/#texel-block-width).

> descriptor.[size](https://gpuweb.github.io/gpuweb/#dom-gputexturedescriptor-size).[height](https://gpuweb.github.io/gpuweb/#extent3d-height) must be multiple of [texel block height](https://gpuweb.github.io/gpuweb/#texel-block-height).

kvark: oh right, hmm

kvark (approved):

PR #2509, opened 2022-02-25
metal: refactor `PrivateCapabilities` creation 

jinleili:
**Description**

- **Follow Apple's API doc:**
> If your app is running on an operating system that supports the [`supportsFamily:`](https://developer.apple.com/documentation/metal/mtldevice/3143473-supportsfamily?language=objc) method, use that method instead.

Some original `family_check && supports_family`  conditons，if MTLGPUFamily lower than `Apple6`,  replaced to: 
```rust
if family_check { supports_family(xx) } else { supports_feature_set(xx) }
``` 
Some original `supports_feature_set || (family_check && supports_family)` conditons，changed  `family_check && supports_family` to front.

- **Follow Apple's [Detecting GPU Features](https://developer.apple.com/documentation/metal/mtldevice/detecting_gpu_features_and_metal_software_versions?language=objc) doc**:
> More recent GPUs have higher version numbers and support larger feature sets. A higher GPU version is always a superset of an earlier version in the same GPU family.

`supports_any` does not need to enumerate all feature sets in the same family，It is enough to keep only the first one。

The OS is updated every year, Maybe some day, we can remove all `supports_feature_set` code.

- `format_r32_xx` detecting is wrong, all GPU Families supports write, r32_float support filer only on macOS, and M1 Mac also support `iOS_GPUFamily_xx` feature:
截屏2022-02-23 15 38 10

Similarly,  `format_rg32_xx` , `format_rgba32_xx`,  `format_depth16unorm_xx`, `format_bgr10a2_xx` need to be fixed too.

- `supports_feature_set(iOS_GPUFamily1_v1)` is equal to `!os_is_mac`  
- `supports_feature_set(macOS_GPUFamily1_v1)` is equal to `os_is_mac`
- `supports_any(&[iOS_GPUFamily1_v1, macOS_GPUFamily1_v1])` is equal to `true` 
- `!supports_any(&[iOS_GPUFamily1_v1, macOS_GPUFamily1_v1])` is equal to `false`

**Testing**
Tested wgpu examples on M1 Mac, Intel Mac.
Tested [wgpu-on-app demos](https://github.com/jinleili/wgpu-on-app) on M1 Mac, Intel Mac, iPhone 6 Plus, iPad Mini4, iPhone 12 Pro.
Tested my private app [字习 Pro](https://apps.apple.com/cn/app/字习-pro/id1507339788) on M1 Mac,  iPhone 6 Plus, iPad Mini4, iPhone 12 Pro，iPad Pro 2018


wgpu-hal/src/metal/adapter.rs
```diff
@@ -395,161 +391,95 @@ impl crate::Adapter<super::Api> for super::Adapter {
 
 const RESOURCE_HEAP_SUPPORT: &[MTLFeatureSet] = &[
     MTLFeatureSet::iOS_GPUFamily1_v3,
-    MTLFeatureSet::iOS_GPUFamily2_v3,
```
kvark, line 398: This looks wrong to me. Is the change based on this quote from their docs?
> More recent GPUs have higher version numbers and support larger feature sets. A higher GPU version is always a superset of an earlier version in the same GPU family.

It says `iOS_GPUFamily1_v3` includes `iOS_GPUFamily1_v2` and `iOS_GPUFamily1_v1`.
It doesn't say that `iOS_GPUFamily2_v3` includes `iOS_GPUFamily1_v3`, which is what the change seems to imply here.

wgpu-hal/src/metal/adapter.rs
```diff
@@ -560,28 +490,82 @@ const LAYERED_RENDERING_SUPPORT: &[MTLFeatureSet] = &[
...
+}
+
+const OS_NOT_SUPPORT: (usize, usize) = (10000, 0);
+// Metal was first introduced on iOS 8.0
+const IOS_MIN_VERSION: (usize, usize) = (8, 0);
+// Metal was first introduced on macOS 10.11
+const MAC_MIN_VERSION: (usize, usize) = (10, 11);
+
+impl NSOperatingSystemVersion {
+    fn at_least(
```
kvark, line 519: this seems a bit strange to me. Shouldn't `self` already know if it's for mac or not? Why are we providing more information about `self` in an argument?

wgpu-hal/src/metal/adapter.rs
```diff
@@ -560,28 +490,82 @@ const LAYERED_RENDERING_SUPPORT: &[MTLFeatureSet] = &[
...
+        &self,
+        os_is_mac: bool,
+        mac_version: NSOperatingSystemVersion,
+        ios_version: NSOperatingSystemVersion,
+    ) -> bool {
+        if os_is_mac {
+            self.major > mac_version.major
```
kvark, line 526: why not just compare tuples or `derive(PartialOrd)`?

wgpu-hal/src/metal/adapter.rs
```diff
@@ -622,54 +591,50 @@ impl super::PrivateCapabilities {
             sample_count_mask |= 8;
         }
 
+        let rw_texture_tier = if version.at_least(os_is_mac, (10, 13).into(), 11.into()) {
```
kvark, line 594: looks like we could make `at_least` accept generic argument(s) converting into the version

wgpu-hal/src/metal/adapter.rs
```diff
@@ -702,44 +666,20 @@ impl super::PrivateCapabilities {
...
-                ],
-            ) && !os_is_mac,
-            format_r32float_no_filter: !Self::supports_any(
-                device,
-                &[
-                    MTLFeatureSet::iOS_GPUFamily1_v1,
-                    MTLFeatureSet::iOS_GPUFamily2_v1,
-                ],
-            ) && !os_is_mac,
+            format_r32_all: true,
```
kvark, line 679: I'm always getting confused about what this means

wgpu-hal/src/metal/adapter.rs
```diff
@@ -845,45 +749,33 @@ impl super::PrivateCapabilities {
...
-                device,
-                &[
-                    MTLFeatureSet::iOS_GPUFamily1_v2,
-                    MTLFeatureSet::iOS_GPUFamily2_v2,
-                    MTLFeatureSet::tvOS_GPUFamily1_v1,
-                ],
-            ) {
-                8192
             } else {
-                4096
```
kvark, line 858: why are we never getting 4096 any more?

kvark (changes requested):

wgpu-hal/src/metal/adapter.rs
jinleili: > Metal organizes GPUs into four major families:
Use the Common family to create apps that target a range of GPUs on multiple platforms.
Use the Apple family to create apps that target Apple GPUs.
Use the Mac family to create apps that target GPUs on macOS systems.
Use the Mac Catalyst family when building an iPadOS app to run on macOS.

`iOS_GPUFamilyX_vX` all the same Apple family.

kvark: Interesting, thank you for the links and quotes!

kvark (approved):

PR #2511, opened 2022-02-25
Wrap all validation logs with catch_unwinds

### LINKED ISSUE #2487: Inadequate unwind protection for log calls under DX12
**Description**
The `log` macros and functions are internally calls out to "whatever zany stuff the Rust ecosystem has dreamt up" so it's nearly impossible to prove they won't panic. Currently `wgpu` is using `log` while behind FFI in [at least one instance](https://github.com/gfx-rs/wgpu/blob/0545e36aa82709cca78c14eb3813f10eea7a9275/wgpu-hal/src/dx12/instance.rs#L61) and this is causing dead locks when used with `log4rs`. `log4rs` will spawn a thread to roll its file over as part of [its rolling_file functionality](https://docs.rs/log4rs/latest/log4rs/append/rolling_file/index.html). After the Rust runtime has called `ExitProcess` from Win32 the aforementioned wgpu FFI code is ***still running***.  However, Windows does not permit you to spawn a thread after `ExitProcess` has been called, so this log statement panics and hangs indefinitely.

**Repro steps**
Add `log4rs` logging using a rolling file appender to a `wgpu` program using only the DX12 backend, and set the size at which it rolls over to a very small size such that it is rolling over very often. Do not filter `warn` level logs from `wgpu`. Use the log_panics crate to log panics as well. Start and stop the program repeatedly. Eventually it should hang.
Attaching a debugger to it will reveal that it has hung due to a panic on a `log` statement behind FFI.

I expected that `wgpu` would [catch_unwind](https://doc.rust-lang.org/stable/std/panic/fn.catch_unwind.html) to catch any potential `log` panic, and then either do nothing, or abort.

**Platform**
Windows 11, wgpu 0.11, rustc 1.58.1

cwfitzgerald:
**Connections**

Closes #2487

**Description**

Wraps all log calls with catch_unwind on VK, GL, and DX12

**Testing**
Unneeded

### DIFF
--- a/wgpu-hal/src/dx12/instance.rs
+++ b/wgpu-hal/src/dx12/instance.rs
@@ -58,7 +58,9 @@ unsafe extern "system" fn output_debug_string_handler(
         return excpt::EXCEPTION_CONTINUE_SEARCH;
     }
 
-    log::log!(level, "{}", message);
+    let _ = std::panic::catch_unwind(|| {
+        log::log!(level, "{}", message);
+    });
 
     if cfg!(debug_assertions) && level == log::Level::Error {
         // Panicking behind FFI is UB, so we just exit.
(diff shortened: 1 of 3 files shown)


cwfitzgerald:
Merging, uncontroversial

PR #2520, opened 2022-03-01
gles: Fix texture description for bgra formats

JCapucho:
**Connections**
None that I'm aware

**Description**
Bgra formats were previously using `glow::RGBA` as the format which caused
texture copies to produce weird colors.

This PR changes them to `glow::BGRA`

**Testing**
This change was tested on a modified fork of [veloren](https://gitlab.com/veloren/dev/veloren/-/tree/capucho/wgpu-0.10) which when using gles was producing weird colors when doing a screenshot, applying this PR fixes it.


cwfitzgerald (approved):
Ooops!

PR #2524, opened 2022-03-05
Fix wasm examples failing to compile

Liamolucko:
**Description**
The web examples were broken by 476b6a1, because apparently enabling a feature in any package enables it for the whole workspace, and the wasm examples fail to compile with Vulkan enabled.

It wasn't failing in CI because CI specifies `-p wgpu`, which seems to stop `wgpu-info`'s enabling of features having an effect.

**Testing**
If you try to run a wasm example on master, it'll fail to compile due to `libloading` being included as a part of vulkan support, which doesn't support wasm. With this change, it works again.

### DIFF
--- a/wgpu-info/Cargo.toml
+++ b/wgpu-info/Cargo.toml
@@ -9,6 +9,10 @@ repository = "https://github.com/gfx-rs/wgpu"
 keywords = ["graphics"]
 license = "MIT OR Apache-2.0"
 
-[dependencies]
+# Even if you aren't actually building `wgpu-info`,
+# enabling the `angle` and `vulkan-portability` features of wgpu enables them for everything built in this workspace.
+# This includes the web examples, which fail to compile when Vulkan is enabled,
+# so disable that on web (by just disabling everything, since this won't compile for wasm anyway).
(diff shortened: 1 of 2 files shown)


cwfitzgerald:
That's some wild stuff, my worry though now is that `cargo build --target wasm32-unknown-unknown` will now fail to compile as wgpu-info won't compile. Maybe the solution is to have the features enabled on non-wasm only (so have two wgpu deps, one unconditionally with no features, and one with these two features)

kvark:
@cwfitzgerald do we need wgpu-info on the Web at all? If not, we could add a top-level guard on wgpu-info's code to only work on native, and then it will happily compile (as an empty crate) for the web.

cwfitzgerald:
No we don't, so that would be a perfectly viable solution too.

kvark:
@Liamolucko would you be open to making this change to wgpu-info, making it an empty crate for wasm targets?

Liamolucko:
> @Liamolucko would you be open to making this change to wgpu-info, making it an empty crate for wasm targets?

Okay, done.

cwfitzgerald (approved):
LG TM

PR #2525, opened 2022-03-06
Fix incorrect filtering used in mipmap generation

LaylBongers:
The min filtering setting is used to down-sample, this is what is used for mipmap generation in this example. It was set to `Nearest` which will pick one of the pixels, which is incorrect behavior for down-sampling. `Linear` correctly averages the 4 pixels when down-sampling.


LaylBongers:
Difference in behavior in my application between these sampling modes making use of this example's code.

Nearest:
magecraft-player_UACdEUSPsr

Linear:
magecraft-player_NSv6HhjkOg

cwfitzgerald:
Doh! That's a big one, good catch. You need to regenerate the rest screenshot to get ci to pass. Go to wgpu/examples/mipmap/screenshot.png and delete it, then run `cargo test --examples mipmap` which will regenerate it.

kvark:
Interesting how this went on unnoticed. Good catch!

cwfitzgerald (approved):

PR #2527, opened 2022-03-07
Prefer dx12 adapter

xiaopengli89:
**Description**
Since DX12 is already first-class support, it should be preferred on Windows.


cwfitzgerald:
Thanks for the PR!

Unfortunately, while we do have the end goal of dx12 being the default adapter, there are blocking issuws with dx12, mainly around performance and native feature support that cause us to stick with vulkan as default for the time being.

PR #2537, opened 2022-03-13
metal: improve `push | pop_debug_marker`

jinleili:
**Description**
Invoking `push | pop_debug_marker`, `entry_any` may create empty blit command encoder, which results in warnings and incorrect `debug_group` hierarchy relationships:
```sh
`Blit Encoder "Oxxxx" has no work encoded to it` (Issue) Your app created an encoder that had no work encoded to it. Creating an empty encoder causes unnecessary CPU overhead and potentiallv bandwidth usage.
```

boids_push_debug_marker

**Testing**
Tested boids example on macOS and iOS:
boids_push_debug_marker-1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -995,7 +995,7 @@ dependencies = [
 [[package]]
 name = "metal"
 version = "0.23.1"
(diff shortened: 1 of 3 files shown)


kvark (approved):

kvark:
Great work, thank you!

Issue #2144, opened 2021-11-01
hello-triangle resize behavior


prideout:
Resizing the window in `wgpu/examples/hello-triangle` works fine on my Windows NVIDIA machine but not on my M1 MacBook. It doesn't seem to refresh even though the `WindowEvent::Resized` arm is definitely getting hit in the event loop.  Screenshot attached after widening the window. (no error messages).

<img width="1485" alt="Screen Shot 2021-11-01 at 3 35 23 PM" src="https://user-images.githubusercontent.com/1288904/139750973-30931b4e-53bc-4b38-9a15-92f4964c21a7.png">

kvark:
What about other examples like `cube`?

prideout:
`cube` works great, there is no resize issue there.

xiaopengli89:
This problem can be solved by calling `window.request redraw();` during the resize event.

xiaopengli89:
The `cube` example uses `framework.rs`, which will redraw every 1/60 second.

PR #2543, opened 2022-03-20
Fix resizing behaviour of hello-triangle example

FrankenApps:
On macos the window is not updated until requested explicitly.

**Connections**
Issue was first raised in #2144.

**Description**
The window extended and the surface was resized accordingly, but the window was never updated with the new render.

**Testing**
Tested on Intel macbook pro.

### DIFF
--- a/wgpu/examples/hello-triangle/main.rs
+++ b/wgpu/examples/hello-triangle/main.rs
@@ -93,6 +93,7 @@ async fn run(event_loop: EventLoop<()>, window: Window) {
                 config.width = size.width;
                 config.height = size.height;
                 surface.configure(&device, &config);
+                window.request_redraw();
             }
             Event::RedrawRequested(_) => {
                 let frame = surface
--- a/wgpu/examples/hello-triangle/main.rs
+++ b/wgpu/examples/hello-triangle/main.rs
@@ -93,6 +93,7 @@ async fn run(event_loop: EventLoop<()>, window: Window) {
                 config.width = size.width;
                 config.height = size.height;
                 surface.configure(&device, &config);
+                // On macos the window needs to be redrawn manually after resizing
                 window.request_redraw();
             }
             Event::RedrawRequested(_) => {
--- a/wgpu/examples/hello-windows/main.rs
+++ b/wgpu/examples/hello-windows/main.rs
@@ -108,6 +108,8 @@ async fn run(event_loop: EventLoop<()>, viewports: Vec<(Window, wgpu::Color)>) {
                 // Recreate the swap chain with the new size
                 if let Some(viewport) = viewports.get_mut(&window_id) {
                     viewport.resize(&device, size);
+                    // On macos the window needs to be redrawn manually after resizing
+                    viewport.desc.window.request_redraw();
                 }
             }
             Event::RedrawRequested(window_id) => {


cwfitzgerald:
Do the other examples need to be fixed too? In a way this feels like a winit bug, but I'm happy to accept this workaround.

wgpu/examples/hello-triangle/main.rs
```diff
@@ -93,6 +93,7 @@ async fn run(event_loop: EventLoop<()>, window: Window) {
                 config.width = size.width;
                 config.height = size.height;
                 surface.configure(&device, &config);
+                window.request_redraw();
```
cwfitzgerald, line 96: Could you just add a comment why we need this?

FrankenApps:
@cwfitzgerald I am not sure if it is a `winit` Issue. They do recommend calling [`window.request_redraw()`](https://docs.rs/winit/latest/winit/window/struct.Window.html#method.request_redraw) and also claim that it integrates with OS issued redraw requests, so I do not think that this should lead to additional redraws on other platforms.

There was one more example `hello-windows` that needed to be fixed, so I did this too.

The other examples with a `winit` window use the `framework` which renders at a given target framerate nevertheless (see [here](https://github.com/gfx-rs/wgpu/blob/326af60df8623e93b47a0de090e6cb449c8507f5/wgpu/examples/framework.rs#L249)) so it does not really matter.

cwfitzgerald:
My thinking this is a winit issue is because a resize should always retrigger a redraw without the need to manually do it.

cwfitzgerald (approved):

cwfitzgerald:
Alright, I poked them on their matrix to see what they say.

FrankenApps:
@cwfitzgerald I was in the process of opening an Issue on their repo, but now I think we might be on to something weird here and I am not sure if the bug is on their side (I would say most like yes, but I am not sure). 

So I created a minimal example for filing a bug report for `winit`, but I discovered, that everything behaved as expected on my macos machine:
```rust
use winit::{
    event::{Event, WindowEvent},
    event_loop::{ControlFlow, EventLoop},
    window::Window,
};

fn main() {
    let event_loop = EventLoop::new();
    let _window = Window::new(&event_loop).unwrap();

    event_loop.run(move |event, _, control_flow| {
        *control_flow = ControlFlow::Wait;
        match event {
            Event::WindowEvent {
                event: WindowEvent::Resized(_),
                ..
            } => {
                println!("Resized event fired.")
            }
            Event::RedrawRequested(_) => {
                println!("Redraw requested.")
            }
            Event::WindowEvent {
                event: WindowEvent::CloseRequested,
                ..
            } => *control_flow = ControlFlow::Exit,
            _ => {}
        }
    });
}
```

results in an output like this:
```
Redraw requested.
Redraw requested.
Resized event fired.
Resized event fired.
Redraw requested.
Resized event fired.
Redraw requested.
Resized event fired.
Redraw requested.
...
```

So I compared this minimal version to the `hello-triangle` example and noted some small differences. At first I thought the problem might be related to `pollster` or that we are running inside an `async` function and pass the window around. But it turns out that is not the problem. Instead the `window` stops firing redraw requests on resize automatically after `let _surface = unsafe { instance.create_surface(&window) };` is executed. 

Here is a minimal version, with which I can reproduce the Issue on macos:
```rust
use winit::{
    event::{Event, WindowEvent},
    event_loop::{ControlFlow, EventLoop},
    window::Window,
};

fn main() {
    let event_loop = EventLoop::new();
    let window = Window::new(&event_loop).unwrap();

    let instance = wgpu::Instance::new(wgpu::Backends::all());

    // Comment out the next line to see, that it works without creating
    // a surface first.
    let _surface = unsafe { instance.create_surface(&window) };

    event_loop.run(move |event, _, control_flow| {
        *control_flow = ControlFlow::Wait;
        match event {
            Event::WindowEvent {
                event: WindowEvent::Resized(_),
                ..
            } => {
                println!("Resized event fired.")
            }
            Event::RedrawRequested(_) => {
                println!("Redraw requested.")
            }
            Event::WindowEvent {
                event: WindowEvent::CloseRequested,
                ..
            } => *control_flow = ControlFlow::Exit,
            _ => {}
        }
    });
}
```

it produces the following output when run:
```
Redraw requested.
Resized event fired.
Resized event fired.
Resized event fired.
Resized event fired.
...
```

I am unfamiliar with what you are doing with the window in `create_surface()` and if there might be any Issue there, so I decided to report back here first.

Liamolucko:
I wonder if this is related to rust-windowing/winit#2189? Apparently using Gl/Metal in a view causes the AppKit equivalent of `request_redraw` not to work, so that might be preventing resizes from triggering redraws too.

cwfitzgerald:
Yeah I definitely would consider this a bug because of what @Liamolucko said, they likely need to automatically add a request redraw on their own for the resize events. The impression I got from chatting with them is they definitely want it to fire every time no matter what.

PR #2548, opened 2022-03-21
Simplify implementation of RefCount and MultiRefCount.

jimblandy:
`RefCount::rich_drop_inner` is no longer used by anything other than `RefCount::drop`. It's simpler to just handle it directly in `drop`.

`MultiRefCount` has no need to heap-allocate the count, since it's never cloned.


kvark (approved):
llove this, thank you!

Issue #1904, opened 2021-09-04
Skip buffer zero initialization for Metal backend


Wumpf:
We create buffers with new_buffer which is guaranteed to return a zero initialized buffer
https://developer.apple.com/documentation/metal/mtldevice/1433375-newbufferwithlength?language=objc

-> wgpu_hal backends should communicate their init guarantees to wgpu which then can set the init status accordingly

Wumpf:
For DX12 we rely on zero init by api here
https://github.com/gfx-rs/wgpu/blob/master/wgpu-hal/src/dx12/device.rs#L74

Are we sure this is sufficient?
```
D3D12_HEAP_FLAG_CREATE_NOT_ZEROED
Allows the OS to not zero the heap created. By default, committed resources and heaps are almost always zeroed upon creation.
```
(https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_heap_flags)
It says "almost always zeroed" 😕

kvark:
Ugh, I wish I knew what "almost always" means here :(

Onirik79:
I found this note about it:
"Actual zeroing depending on resource size and CPU visibility, with details of what properties changing from one OS release to another, and from one CPU architecture to another."

Source:
https://devblogs.microsoft.com/directx/coming-to-directx-12-more-control-over-memory-allocation/

kvark:
I think we'll need to zero-initialize all CPU-visible buffers in D3D12, given that:
  - we need to manage them without 1:1 committed resources, via `gpu-alloc` or by hand.
  - linked blog mentions them specifically
  - zeroing them out is purely a CPU operation

PR #2550, opened 2022-03-23
Skip buffer zero initialization for Metal backend

jinleili:
**Connections**
https://github.com/gfx-rs/wgpu/issues/1904

**Testing**
Tested wgpu examples on M1 Mac, Intel Mac and iPhone 12.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -569,6 +569,12 @@ bitflags::bitflags! {
         ///
         /// This is a native-only feature.
         const TEXTURE_COMPRESSION_ASTC_HDR = 1 << 43;
+        /// Skip buffer zero initialization
+        /// Supported Platforms:
+        /// - Metal
+        ///
+        /// This is a native-only feature.
+        const ALLOCATED_BUFFER_ALREADY_ZERO_FILLED = 1 << 44;
     }
 }
 
--- a/wgpu-core/src/device/queue.rs
+++ b/wgpu-core/src/device/queue.rs
@@ -758,9 +758,14 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                                 .map_err(DeviceError::from)?
                         };
                         log::trace!("Stitching command buffer {:?} before submission", cmb_id);
-                        baked
-                            .initialize_buffer_memory(&mut *trackers, &mut *buffer_guard)
-                            .map_err(|err| QueueSubmitError::DestroyedBuffer(err.0))?;
+                        if !device
+                            .features
+                            .contains(wgt::Features::ALLOCATED_BUFFER_ALREADY_ZERO_FILLED)
+                        {
+                            baked
+                                .initialize_buffer_memory(&mut *trackers, &mut *buffer_guard)
+                                .map_err(|err| QueueSubmitError::DestroyedBuffer(err.0))?;
+                        }
                         baked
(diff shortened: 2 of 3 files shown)


wgpu-core/src/command/memory_init.rs
```diff
@@ -148,11 +148,15 @@ pub(crate) fn fixup_discarded_surfaces<
 
 impl<A: hal::Api> BakedCommands<A> {
     // inserts all buffer initializations that are going to be needed for executing the commands and updates resource init states accordingly
+    #[allow(unused_variables, unreachable_code)]
     pub(crate) fn initialize_buffer_memory(
         &mut self,
         device_tracker: &mut TrackerSet,
         buffer_guard: &mut Storage<Buffer<A>, id::BufferId>,
     ) -> Result<(), DestroyedBufferError> {
+        #[cfg(metal)]
```
kvark, line 157: Let's implement it properly. wgpu-core doesn't need any platform-specific hacks outside of the backend selection process.
A proper way would be for the wgpu-hal backend to offer a feature "my allocated buffers are always initialized to zero".  And then wgpu-core would check for this feature and activate it.

kvark (changes requested):

wgpu-types/src/lib.rs
```diff
@@ -569,6 +569,12 @@ bitflags::bitflags! {
         ///
         /// This is a native-only feature.
         const TEXTURE_COMPRESSION_ASTC_HDR = 1 << 43;
+        /// Skip buffer zero initialization
+        /// Supported Platforms:
+        /// - Metal
+        ///
+        /// This is a native-only feature.
+        const ALLOCATED_BUFFER_ALREADY_ZERO_FILLED = 1 << 44;
```
cwfitzgerald, line 577: I don't think this should end up in the public facing feature set here as this is something that the user doesn't need to know about at all. @kvark thoughts on where it should go instead?

kvark: I think it's fine for some of the internal (i.e. needed on the hal - core boundary only) flags to be in this public bitflags list, as long as wgpu-core never reports them back to the user.
We could improve this by moving those internal flags into a separate bitflags type that would live in wgpu-hal.

jinleili: @kvark Not sure if I understand correctly, is the internal flags something like this?
```rust
bitflags::bitflags! {
	#[repr(transparent)]
	#[derive(Default)]
	pub struct InternalFeatures: u64 {
		const ALLOCATED_BUFFER_ALREADY_ZERO_FILLED = 1 << 0;
	}
}
// wgpu_hal/lib.rs
pub trait Device<A: Api>: Send + Sync {
	...
	fn contains_internal_feature(&self, _feature: wgt::InternalFeatures) -> bool;
}
// wgpu_hal/metal/mod.rs
pub struct Device {
	...
	internal_features: wgt::InternalFeatures,
}

// wgpu_hal/metal/device.rs
impl crate::Device<super::Api> for super::Device {
	...
	fn contains_internal_feature(&self, feature: wgt::InternalFeatures) -> bool {
		self.internal_features.contains(feature)
	}
}
```

kvark: Let's discuss it as a follow-up, specifically aimed at the flags. Current PR should go in by just using the `Features` and having one there.

PR #2557, opened 2022-03-24
Update `ash` to `0.37`

a1phyr:
This raises MSRV to 1.59, not sure if this is acceptable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on:
 
 env:
   RUST_BACKTRACE: 1
-  RUST_VERSION: 1.53
+  RUST_VERSION: 1.59
 
 # We distinguish the following kinds of builds:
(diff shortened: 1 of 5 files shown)


cwfitzgerald (approved):
I'm good with this, but MSRV bump needs a look by @kvark - who are the stakeholders who might have an issue with this?

kvark:
We should give a heads up to Bevy, Dotrix, Rend3, etc.
Not as much worried about Gecko, since it's very close to the latest stable at all times.

a1phyr:
I did a little investigation:
- Bevy uses latest stable
- Rend3 uses 1.57
- I could not find the information for Dotrix and Vange-rs
- Veloren uses nightly but is still using an old `wgpu` version anyway

kvark:
Thank you!
So let's go with it.

kvark (approved):

PR #2558, opened 2022-03-27
Document some aspects of resource tracking.

jimblandy:
Please read carefully - these are just my inferences from reading the code, so I could have misunderstood something important. Hopefully it's at least a step forward.

### DIFF
--- a/wgpu-core/src/lib.rs
+++ b/wgpu-core/src/lib.rs
@@ -50,7 +50,11 @@ use atomic::{AtomicUsize, Ordering};
 
 use std::{borrow::Cow, os::raw::c_char, ptr, sync::atomic};
 
+/// The index of a queue submission.
+///
+/// These are the values stored in `Device::fence`.
 type SubmissionIndex = hal::FenceValue;
+
 type Index = u32;
 type Epoch = u32;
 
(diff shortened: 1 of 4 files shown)


wgpu-core/src/lib.rs
```diff
@@ -122,10 +135,48 @@ impl MultiRefCount {
...
+    /// When the user first creates a `wgpu-core` resource, this `RefCount` is
+    /// created along with the resource's `LifeGuard`. When the user drops the
+    /// resource, we swap this out for `None`. Note that the resource may
+    /// still be held alive by other resources.
+    ///
+    /// Any `Stored<T>` value holds a clone of this `RefCount` along with the id
+    /// of a `T` resource.
     ref_count: Option<RefCount>,
+
+    /// A timestamp of the last queue submission in which the resource
```
kvark, line 170: It's not a timestamp, technically, it's just an index.

kvark (approved):
Good docs, thank you!

kvark:
formatting is angry a bit

wgpu-core/src/lib.rs
jimblandy: 'Serial number' would be an exact fit for this, but the code and comments all use 'index', so I'll go with that.

jimblandy:
I don't think the remaining failure has anything to do with this PR.

cwfitzgerald:
We have a spurious failure on GL multithreaded

PR #2567, opened 2022-03-31
Free the raw device when `wgpu::Device` is dropped.

### LINKED ISSUE #2563: `Device::drop` doesn't actually free the device when using `backend::direct::Context`
`wgpu::Device::drop` calls `Context::device_drop`. On `backend::direct::Context`, the implementation doesn't actually call `wgpu_core::hub::Global::device_drop`, so the backend device resources are never freed.

The body of [`backend::direct::Context::device_drop`](https://github.com/gfx-rs/wgpu/blob/58f318550dedaec6ed901c83131e424cf39ee648/wgpu/src/backend/direct.rs#L1556) has a `TODO` for this:

```
//TODO: make this work in general
#[cfg(any(not(target_arch = "wasm32"), feature = "emscripten"))]
#[cfg(feature = "metal-auto-capture")]
{
    let global = &self.0;
    wgc::gfx_select!(device.id => global.device_drop(device.id));
}
```

So I guess I'm filing this issue to promote that TODO to an actual bug.

jimblandy:
Go ahead and call `global.device_drop` from `direct::Context::device_drop`.

Let `Global::device_drop` simply drop the life guard's `RefCount`, and put off everything else entailed in freeing a device until `Device::maintain` says its queue is empty and there are no more references to it. (The user can reach that function, even after dropping their `Device`, by calling `wgpu::Instance::poll_all`.)

Fixes #2563.

This won't have any effect until #2565 is landed as well.

### DIFF
--- a/wgpu-core/src/device/life.rs
+++ b/wgpu-core/src/device/life.rs
@@ -309,6 +309,11 @@ impl<A: hal::Api> LifetimeTracker<A> {
         }
     }
 
+    /// Return true if there are no queue submissions still in flight.
+    pub fn queue_empty(&self) -> bool {
+        self.active.is_empty()
+    }
+
     /// Start tracking resources associated with a new queue submission.
     pub fn track_submission(
         &mut self,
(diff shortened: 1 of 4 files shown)


jimblandy:
This is just a "see if this works" patch; totally open to suggestions for different approaches.

kvark (approved):

PR #2572, opened 2022-04-02
impl Debug for StagingBelt

kpreid:
**Description**

This PR adds an explicit `impl Debug for StagingBelt`. Most `wgpu` types implement `Debug`, but `StagingBelt` doesn't, so I figured it was reasonable to add. The individual `Chunk`s are not included but their counts are.

**Testing**

I didn't see any unit tests for other `Debug` impls, so I didn't add one, but I manually tested that the output was reasonable by adding debug printing to a program using `wgpu`.


cwfitzgerald (approved):
Nice!

PR #2576, opened 2022-04-03
wgpu-hal: Work around cbindgen bug: ignore `gles::egl` module.

### LINKED ISSUE #2575: cbindgen stumbles over egl type aliases
The [following code](https://github.com/gfx-rs/wgpu/blob/fab11dcc76e160271f69c800896ff260f15af220/wgpu-hal/src/gles/egl.rs#L30-L34) in `wgpu-hal/src/gles/egl.rs` triggers a bug in cbindgen:

```
type EglInstance = egl::DynamicInstance<egl::EGL1_4>;

#[cfg(feature = "emscripten")]
type EglInstance = egl::Instance<egl::Static>;
```

As described in eqrion/cbindgen#286, cbindgen stumbles when the same name refers to a generic type in one module and a non-generic type in another, and then a type alias is introduced for one in the scope of the other. (As the docs warn, cbindgen [doesn't really understand Rust modules](https://github.com/eqrion/cbindgen/blob/master/docs.md#writing-your-c-api)).

To reproduce, run `cbindgen --lang C` in the `wgpu-hal` directory (using cbindgen version 0.19.0). For me, this panics:

```
thread 'main' panicked at 'Instance is not generic', src/bindgen/ir/opaque.rs:105:9
```

This is obviously a cbindgen bug, not wgpu's problem. But comments in the cbindgen issue linked above ominously suggest that this is a consequence of eqrion/cbindgen#7, which seems like a pretty difficult architectural issue to fix. So I think it makes sense for wgpu to work around it.

jimblandy:
The definition of `EglInstance` trips eqrion/cbindgen#286. Since the `wgpu_hal::gles::egl` module is an implementation detail anyway, it should be harmless to ask cbindgen to just ignore it.

Fixes #2575.

### DIFF
--- a/wgpu-hal/src/gles/mod.rs
+++ b/wgpu-hal/src/gles/mod.rs
@@ -56,6 +56,7 @@ To address this, we invalidate the vertex buffers based on:
 
 */
 
+///cbindgen:ignore
 #[cfg(any(not(target_arch = "wasm32"), feature = "emscripten"))]
 mod egl;
 #[cfg(all(target_arch = "wasm32", not(feature = "emscripten")))]


cwfitzgerald (approved):
Can't get mad at that lol

PR #2590, opened 2022-04-12
Prefix every wgpu-generated label with `(wgpu)`.

kpreid:
This is intended to help developers new to wgpu or to graphics debugging quickly recognize in a debugging tool which items are wgpu-generated, as opposed to either part of their program or part of the platform graphics system.

I also removed existing marker-like text such as leading underscores and angle brackets. I'm not sure if this was the right decision, but they seemed arbitrary.

**Testing**

* The wgpu test suite gets as far as it does on my machine without this change.
* Running with a GPU debugger (Xcode/Metal) shows the changed labels as I expect.

### DIFF
--- a/wgpu-core/src/command/clear.rs
+++ b/wgpu-core/src/command/clear.rs
@@ -450,7 +450,7 @@ fn clear_texture_via_render_passes<A: hal::Api>(
             };
             unsafe {
                 encoder.begin_render_pass(&hal::RenderPassDescriptor {
-                    label: Some("clear_texture clear pass"),
+                    label: Some("(wgpu internal) clear_texture clear pass"),
                     extent,
                     sample_count,
                     color_attachments,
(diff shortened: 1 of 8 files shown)


cwfitzgerald (approved):
I think this is a good change, but we might want to say like `(wgpu internal)` to be clear that it's an internal resource, not just one made with wgpu

kpreid:
> … but we might want to say like `(wgpu internal)` …

Done.

kpreid:
Another idea I had while working on this: some of the objects could be usefully labeled after the other object they're being used to operate on. However, that would require allocating strings, so it would make more sense as an optional feature for debugging.

kpreid:
The check failure log doesn't give any information other than 

```
        FAIL [   0.485s]              wgpu::example/shadow shadow
```

and it doesn't look like the `shadow` test should care about labels. Should I be doing something here?

kvark (approved):

kvark:
@cwfitzgerald same lavapipe issue?

cwfitzgerald:
Just needs to be rebased on top of master.

kvark:
easy peasy

PR #2602, opened 2022-04-15
Update to Naga rev 1720725

Gordon-F:
Update `naga` with new entry point stage attributes: https://github.com/gfx-rs/naga/pull/1833


jimblandy (approved):
Looks good - thank you!

PR #2604, opened 2022-04-17
Add symlink LICENSE files into crates.

### LINKED ISSUE #2561: LICENSE files in package subdirs
### Background:
Hello, the Fuchsia project vendors crates from crates.io, and in order to do so we require explicit license files alongside the source code. Here is the policy: https://fuchsia.dev/fuchsia-src/contribute/governance/policy/open-source-licensing-policies?hl=en#licenses_and_tracking .  In particular, reading the SPDX `package/license` field from the crate's `Cargo.toml` is not good enough.

### Request:
Could you please add LICENSE-* files to the package subdirectories?  We care particularly about `wgpu`, `wgpu-core`, `wgpu-hal`, and `wgpu-types` but you might as well do the other subdirs while you're at it, e.g. `wgpu-info`.

Regarding the versions that we need updated, we're just about to start using `wgpu` (as soon as we can work through all these licensing issues, including for downstream dependencies), so bumping the latest versions of all crates would work for us.

I'd appreciate if you're able to do this, since I have to file similar issues for many other crate dependencies.  But if you don't have the bandwidth to address this issue, please let me know and I'll find time to submit a pull request (although of course I won't be able to upload anything to crates.io)

dskkato:
Symlink the top-level LICENSE into each published crate.

This will resolve #2561

### DIFF
--- a/deno_webgpu/LICENSE.APACHE
+++ b/deno_webgpu/LICENSE.APACHE
@@ -1,1 +1,0 @@
-../LICENSE.APACHE
--- a/deno_webgpu/LICENSE.APACHE
+++ b/deno_webgpu/LICENSE.APACHE
@@ -1,0 +1,1 @@
+../LICENSE.APACHE
--- a/deno_webgpu/LICENSE.MIT
+++ b/deno_webgpu/LICENSE.MIT
@@ -1,0 +1,1 @@
+../LICENSE.MIT
--- a/deno_webgpu/LICENSE.MIT
+++ b/deno_webgpu/LICENSE.MIT
@@ -1,1 +1,0 @@
-../LICENSE.MIT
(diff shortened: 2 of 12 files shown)


dskkato:
I don't think it is appropriate for me to send this PR since I am not a party to this issue, so I will close this. If this correction looks good, please reopen.

kvark:
We should only put it in the sub-crates where it's required (and which are published). As such, I don't think we want to modify deno_webgpu and wgpu-info with these.

cwfitzgerald:
I don't think the symlink thing ended up working considering the dffs

dskkato:
> I don't think the symlink thing ended up working considering the dffs

Is it better to put actual copies to those subdirectory? Currently, I'm just naively trying to follow [the case of prost](https://github.com/tokio-rs/prost/pull/619).

cwfitzgerald (approved):
Alright, if that works for them, that's fine

PR #2606, opened 2022-04-18
Avoid new WGSL reserved words in wgpu examples.

jimblandy:
I don't like WGSL's mass reservation strategy, but it's what we have to live with for now.


cwfitzgerald (approved):
Fun

jimblandy:
oops, should have rebased, not merged

PR #2609, opened 2022-04-18
[WebGL] Allow creating a texture from an external `web_sys::WebGlFramebuffer` and writing to it

expenses:
**Description**

The [WebXR DeviceAPI](https://developer.mozilla.org/en-US/docs/Web/API/WebXR_Device_API) works a little different from HTML canvases as it gives you an opaque `WebGlFramebuffer` that you are intended to bind and write to: https://developer.mozilla.org/en-US/docs/Web/API/XRWebGLLayer/framebuffer

To be able to do this in wgpu we need to:

1. Be able to construct a `wgpu::Texture` from the `WebGlFramebuffer`.
2. Bind it in a renderpass and write to it successfully.

which is what this PR does.

~~It currently depends on https://github.com/grovesNL/glow/pull/218 and should not be merged until that is.~~

**Now depends on the glow branch that wgpu is using**

~~I've got a demo up at https://expenses.github.io/wgpu-vr/ (if you don't have a VR device then you can use the google cardboard stuff in chrome on android to try it out). Source: https://github.com/expenses/kiss-engine/blob/d877a28c0bf86e3b65fb438446647e2998820be4/ludum-dare/src/main.rs~~

**Outdated**

**Things to check**

There are a couple of things in this PR that I'm unsure about.

1. Is it possible to bind WebGL drawbuffers and framebuffers at the same time, so we could (in wgpu speak) bind two or more render attachments, one being the external framebuffer and the others being internal textures?
2. Given that the [opaque framebuffer can have both a color buffer attachment and a depth buffer attachment](https://www.w3.org/TR/webxr/#opaque-framebuffer), how should we be handling these? In my code I just create a texture for both: https://github.com/expenses/kiss-engine/blob/d877a28c0bf86e3b65fb438446647e2998820be4/ludum-dare/src/main.rs#L704-L772

### DIFF
--- a/wgpu-hal/src/gles/command.rs
+++ b/wgpu-hal/src/gles/command.rs
@@ -427,6 +427,20 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
             self.state.has_pass_label = true;
         }
 
+        let rendering_to_external_framebuffer = desc
+            .color_attachments
+            .iter()
+            .filter_map(|at| at.as_ref())
+            .any(|at| match at.target.view.inner {
+                #[cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))]
+                super::TextureInner::ExternalFramebuffer { .. } => true,
+                _ => false,
+            });
+
+        if rendering_to_external_framebuffer && desc.color_attachments.len() != 1 {
+            panic!("Multiple render attachments with external framebuffers are not supported.");
+        }
+
         match desc
             .color_attachments
             .first()
@@ -489,10 +503,12 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
                     }
                 }
 
-                // set the draw buffers and states
-                self.cmd_buffer
-                    .commands
-                    .push(C::SetDrawColorBuffers(desc.color_attachments.len() as u8));
+                if !rendering_to_external_framebuffer {
+                    // set the draw buffers and states
+                    self.cmd_buffer
+                        .commands
+                        .push(C::SetDrawColorBuffers(desc.color_attachments.len() as u8));
+                }
             }
         }
 
--- a/wgpu-hal/src/gles/mod.rs
+++ b/wgpu-hal/src/gles/mod.rs
@@ -184,10 +184,10 @@ impl Default for VertexAttribKind {
 }
 
 #[derive(Clone, Debug)]
-struct TextureFormatDesc {
-    internal: u32,
-    external: u32,
-    data_type: u32,
+pub struct TextureFormatDesc {
+    pub internal: u32,
+    pub external: u32,
+    pub data_type: u32,
 }
 
 struct AdapterShared {
@@ -244,7 +244,7 @@ unsafe impl Sync for Buffer {}
 unsafe impl Send for Buffer {}
 
 #[derive(Clone, Debug)]
-enum TextureInner {
+pub enum TextureInner {
     Renderbuffer {
         raw: glow::Renderbuffer,
     },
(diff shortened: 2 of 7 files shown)


kvark:
This is the same situation as we have with regular OpenGL, but the PR's solution goes a very different route.
What our OpenGL backend does is handling this at the surface level. On present, the surface can take the result of rendering and blit it into the internal framebuffer (see `Surface::present` in `wgpu-hal/src/gles/web.rs` for example).

Would it be possible to follow the same route for WebXR?

expenses:
> Would it be possible to follow the same route for WebXR?

It's possible, but IMO I think that'd feel a bit hacky. You need to be able to change which framebuffer the surface presents to every frame. I'll experiment with that approach

expenses:
> It's possible, but IMO I think that'd feel a bit hacky. You need to be able to change which framebuffer the surface presents to every frame. I'll experiment with that approach

Oh, so this would work for the 'ImmersiveVR' mode but not the 'ImmersiveAR' one, where the framebuffer that you're give contains the image from the camera and you draw on top of that without clearing. We'd have to allow alpha blending when presenting to the surface. 

I think sticking with this PR as-is is the way to go.

wgpu-hal/src/gles/mod.rs
```diff
@@ -240,28 +240,40 @@ enum TextureInner {
         raw: glow::Texture,
         target: BindTarget,
     },
+    #[cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))]
+    ExternalFramebuffer {
+        inner: web_sys::WebGlFramebuffer,
+    },
 }
 
+// SAFE: WASM doesn't have threads
```
bjorn3, line 249: WASM can have threads.

expenses: Well, technically yes but we're ignoring that for now :P. There are some other comments in this file with the same assumption and when threads in wasm are more readily used, we'll have to change things.

expenses:
Note: waiting for a new verison of glow (https://github.com/grovesNL/glow) to be published before merging.

wgpu-hal/src/lib.rs
```diff
@@ -58,7 +58,7 @@ mod dx11;
 mod dx12;
 mod empty;
 #[cfg(all(feature = "gles"))]
-mod gles;
+pub mod gles;
```
cwfitzgerald, line 61: This should probably be an all or nothing deal - if we want to make the backends public, they should all be public.

cwfitzgerald (changes requested):

wgpu-hal/src/lib.rs
expenses: Cool, done in [9bd69ee](https://github.com/gfx-rs/wgpu/pull/2609/commits/9bd69ee5de8223cf54be719d0a37af11ad763154).

wgpu-hal/src/gles/command.rs
```diff
@@ -489,10 +503,12 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
...
 
-                // set the draw buffers and states
-                self.cmd_buffer
-                    .commands
-                    .push(C::SetDrawColorBuffers(desc.color_attachments.len() as u8));
+                if !rendering_to_external_framebuffer {
+                    // set the draw buffers and states
+                    self.cmd_buffer
+                        .commands
+                        .push(C::SetDrawColorBuffers(desc.color_attachments.len() as u8));
```
cwfitzgerald, line 510: How does the framebuffer get set when rendering to an external framebuffer?

expenses: That happens down in `set_attachment`: https://github.com/gfx-rs/wgpu/pull/2609/files#diff-c91d45156b742345a9fe20a6742d5fbc22e2c546a80291a227b1f3bf852052e3R131 which is called by the `C::BindAttachment` command I believe

expenses:
I'm still using this branch and just squashed all the commits together in order to clean things up. It'd be great to get at least some part of this merged soon - even if it's just exposing some of the internals with `pub`.

cwfitzgerald (approved):
We will really need to think about unifiying the raw interfaces betwen backends. In the mean time, this looks fine.

cwfitzgerald:
Checks are failing.

expenses:
Should be good to go now after the tests run

PR #2616, opened 2022-04-19
Do a downlevel check for anisotrophy and enable it in the webgl backend

expenses:
**Description**

Removes a TODO item

**Testing**

Tested on my machine


cwfitzgerald (changes requested):

cwfitzgerald (approved):
Thanks!

PR #2621, opened 2022-04-21
Update texture_create_view logic to  match spec

jinleili:
**Connections**
Spec PR: https://github.com/gpuweb/gpuweb/pull/2755

**Description**
Also updated TextureDescriptor `fn array_layer_count` 
https://github.com/gpuweb/gpuweb/blob/7863716ea2233a5642c123b4c2b30acc193fb26e/spec/index.bs#L3121-L3131

**Testing**
Only the wgpu examples were used for testing.


cwfitzgerald (approved):

kvark:
That sucks, I don't like the spec change.

PR #2622, opened 2022-04-21
metal: Don't skip incomplete binding resources.

dragostis:
The old way of generating buffer size updates would only produce updates when all bindings were used by the entry point. If one happened to be missing in only one entry point, the updates wouldn't happen.

### DIFF
--- a/wgpu-hal/src/metal/command.rs
+++ b/wgpu-hal/src/metal/command.rs
@@ -74,13 +74,17 @@ impl super::CommandState {
     ) -> Option<(u32, &'a [u32])> {
         let stage_info = &self.stage_infos[stage];
         let slot = stage_info.sizes_slot?;
+
         result_sizes.clear();
-        for br in stage_info.sized_bindings.iter() {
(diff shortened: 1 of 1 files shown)


wgpu-hal/src/metal/command.rs
```diff
@@ -74,13 +74,17 @@ impl super::CommandState {
...
-        Some((slot as _, result_sizes))
+        result_sizes.extend(
+            stage_info
+                .sized_bindings
+                .iter()
+                .filter_map(|br| self.storage_buffer_length_map.get(br))
+                .map(|size| size.get().min(!0u32 as u64) as u32),
+        );
+
+        (!result_sizes.is_empty()).then(move || (slot as _, result_sizes.as_slice()))
```
cwfitzgerald, line 87: Let's use a normal if here, then reads a bit confusing imo

cwfitzgerald (changes requested):

cwfitzgerald (approved):
Thanks!

PR #2623, opened 2022-04-21
Bind group deduplication

cwfitzgerald:
**Connections**

One component of https://github.com/BVE-Reborn/rend3/issues/350

**Description**

This adds immediate bind group deduplication and makes pipeline deduplication immediate. This happens before the commands are even recorded to get them as early as possible. It intentionally goes out of its way to _not_ deduplicate bind group calls with dynamic offsets.

I haven't done any other kinds of bindings because I don't know how much they show up as performance issues in practice. Bind groups are expensive enough that deduplicating them is easy enough.

**Testing**

Test suite


kvark (approved):
Looks reasonably light.

PR #2630, opened 2022-04-25
Remove a `log::error!` debugging statement from the gles queue

expenses:
@cwfitzgerald I'm guessing we can remove this now? Or optionally change to to a `log::trace!`.


cwfitzgerald (approved):
Big oops

PR #2637, opened 2022-05-02
Update push constant docs to reflect the API

Noxime:
Alignment rules for push constant data are no longer enforced by using &[u32] as the input type, but the docs don't reflect that. This is a simple documentation fix to address it.


cwfitzgerald (approved):
Thanks!

cwfitzgerald (changes requested):
Oh forgot, we actually need to change the docs on the compute passes too

cwfitzgerald (approved):

PR #2639, opened 2022-05-03
Update parking_lot.

emilio:
There's this pending glutin PR:

  https://github.com/rust-windowing/glutin/pull/1402

And winit, which has already updated but hasn't done a new release yet.
But I'm interested in updating these in Firefox, where we don't use
these, so this patch as-is would still be useful.

**Testing**
Shouldn't change behavior.

### DIFF
--- a/wgpu/Cargo.toml
+++ b/wgpu/Cargo.toml
@@ -112,7 +112,7 @@ version = "0.12"
 [dependencies]
 arrayvec = "0.7"
 log = "0.4"
-parking_lot = "0.11"
+parking_lot = "0.12"
 raw-window-handle = "0.4"
 serde = { version = "1", features = ["derive"], optional = true }
 smallvec = "1"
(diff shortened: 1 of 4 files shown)


wgpu/Cargo.toml
```diff
@@ -283,8 +283,7 @@ web-sys = { version = "0.3.53", features = [
 ]}
 js-sys = "0.3.50"
 wasm-bindgen-futures = "0.4.23"
-# enable parking_lot's wasm-bindgen feature so that it, in turn, enables that of crate `instant`
-parking_lot = { version = "0.11", features = ["wasm-bindgen"] }
+parking_lot = { version = "0.12" }
```
cwfitzgerald, line 286: Could we just move this back to 

```suggestion
parking_lot = "0.12"
```

cwfitzgerald (changes requested):
One comment otherwise LGTM

wgpu/Cargo.toml
emilio: Indeed, done!

cwfitzgerald (approved):
Thanks!

PR #2642, opened 2022-05-04
Exclude dependencies from documents

yutannihilation:
Currently, the document contains dependency crates. If this isn't intentional, is it okay to add `--no-deps` to `cargo doc`?

e.g. https://wgpu.rs/doc/wgpu/struct.RenderPass.html?search=clear


cwfitzgerald (approved):
That's normally how build docs anyway, so LGTM

yutannihilation:
Thanks!

PR #2645, opened 2022-05-07
Rename `timeout_us` to `timeout_ns`, to match actual units.

jimblandy:


jimblandy:
Fortunately, the actual conversion factor applied to the argument was the right one.

cwfitzgerald (approved):
Whoops

PR #2646, opened 2022-05-07
Add `SHADER_FLOAT16` feature

jinleili:
**Connections**
[WebGPU spec](https://gpuweb.github.io/gpuweb/#shader-f16)
https://github.com/gfx-rs/naga/issues/1884

**Description**
Dx12 need to check `D3D12_FEATURE_DATA_D3D12_OPTIONS4::Native16BitShaderOpsSupported` and `D3D_SHADER_MODEL_6_2` to comform whether  supports true FP16 which `winapi` currently don't defined, maybe one more reason to replace `winapi` with `windows-rs`?

**Testing**
Tested on metal and vk backend.

### DIFF
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -25,6 +25,10 @@ pub struct PhysicalDeviceFeatures {
     depth_clip_enable: Option<vk::PhysicalDeviceDepthClipEnableFeaturesEXT>,
     multiview: Option<vk::PhysicalDeviceMultiviewFeaturesKHR>,
     astc_hdr: Option<vk::PhysicalDeviceTextureCompressionASTCHDRFeaturesEXT>,
+    shader_float16: Option<(
+        vk::PhysicalDeviceShaderFloat16Int8Features,
+        vk::PhysicalDevice16BitStorageFeatures,
+    )>,
 }
 
 // This is safe because the structs have `p_next: *mut c_void`, which we null out/never read.
@@ -61,6 +65,10 @@ impl PhysicalDeviceFeatures {
         }
         if let Some(ref mut feature) = self.astc_hdr {
             info = info.push_next(feature);
+        }
+        if let Some((ref mut f16_i8_feature, ref mut _16bit_feature)) = self.shader_float16 {
+            info = info.push_next(f16_i8_feature);
+            info = info.push_next(_16bit_feature);
         }
         info
     }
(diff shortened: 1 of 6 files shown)


wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -326,6 +334,20 @@ impl PhysicalDeviceFeatures {
             } else {
                 None
             },
+            shader_float16: if requested_features.contains(wgt::Features::SHADER_FLOAT16) {
+                Some((
+                    vk::PhysicalDeviceShaderFloat16Int8Features::builder()
+                        .shader_float16(true)
+                        .shader_int8(true)
```
kvark, line 341: why are we enabling int8?

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -531,6 +553,16 @@ impl PhysicalDeviceFeatures {
             );
         }
 
+        if let Some((ref f16_i8, ref bit16)) = self.shader_float16 {
+            features.set(
+                F::SHADER_FLOAT16,
+                f16_i8.shader_float16 != 0
+                    && f16_i8.shader_int8 != 0
```
kvark, line 560: similarly, why are we checking for int8?

kvark (commented):
Nice! Just one small thing to fix here

wgpu-hal/src/vulkan/adapter.rs
jinleili: Oh, my mistake.

wgpu-types/src/lib.rs
```diff
@@ -560,6 +560,13 @@ bitflags::bitflags! {
         ///
         /// This is a native-only feature.
         const TEXTURE_COMPRESSION_ASTC_HDR = 1 << 42;
+        /// Allows shaders to acquire the FP16 ability
+        /// Supported Platforms:
+        /// - Vulkan
+        /// - Metal
+        ///
+        /// This is a web and native feature.
+        const SHADER_FLOAT16 = 1 << 43;
```
cwfitzgerald, line 569: This should be in the web section, not in the native section

wgpu-types/src/lib.rs
```diff
@@ -560,6 +560,13 @@ bitflags::bitflags! {
         ///
         /// This is a native-only feature.
         const TEXTURE_COMPRESSION_ASTC_HDR = 1 << 42;
+        /// Allows shaders to acquire the FP16 ability
```
cwfitzgerald, line 563: Could we also note that this isn't supported in naga, only through spir-v passthrough right now.

wgpu-types/src/lib.rs
jinleili: Thanks for pointing out, I was totally unaware of this.

kvark (approved):

PR #2647, opened 2022-05-08
Make GLES DeviceType unknown by default

PolyMeilex:
I just found out that my integrated GPU is not on `wgp-hal` list.

Thanks to that I realized that when I request `HighPerformance` adapter GLES one will always be selected instead of Vulkan one, because Vulkan reports `IntegratedGpu` properly, so it gets filtered out in favor of GLES `DiscreteGpu`(that in reality is not discrete).

I could just add it to the list, but I don't think that's the right solution, it's an endless fight, and we will never have all integrated gpus on the list no matter what.

So I think the right solution is to just admit that we don't know what is the device type. So `DeviceType::Other` should be used, maybe the name does not fit the case well, but it directly maps to `WGPUAdapterType_Unknown` in `webgpu-native` so I think it's right to use it for this.

### DIFF
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -735,7 +735,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
 
         let preferred_gpu = match desc.power_preference {
             PowerPreference::LowPower => integrated.or(other).or(discrete).or(virt).or(cpu),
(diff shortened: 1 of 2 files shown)


kvark (approved):

PolyMeilex:
Thanks :+1:

PR #2653, opened 2022-05-13
Add cross-references to each wgpu type's documentation.

kpreid:
**Description**

Added:
* WebGPU specification links.
* How to create an instance, if it is not all-public-fields.
* What function or enclosing type this type is used for.

(The latter two are not present for `wgpu-types` items due to the dependency running the wrong direction for that — I saw existing `../wgpu/` links, but those do not work on https://docs.rs/wgpu/.)

This commit specifically does not:
* substantially revise existing language, except where necessary to fit the new information
* add similar information to the documentation of types not directly corresponding to the WebGPU specification

**Testing**

I ran `cargo doc`, saw no broken intra-doc link warnings, and inspected each type's page for correct appearance.

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -507,6 +507,8 @@ trait Context: Debug + Send + Sized + Sync {
 /// Its primary use is to create [`Adapter`]s and [`Surface`]s.
 ///
 /// Does not have to be kept alive.
+///
+/// Corresponds to [WebGPU `GPU`](https://gpuweb.github.io/gpuweb/#gpu-interface).
 #[derive(Debug)]
 pub struct Instance {
     context: Arc<C>,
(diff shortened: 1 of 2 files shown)


wgpu/src/lib.rs
```diff
@@ -1114,25 +1197,65 @@ pub struct RenderPassDepthStencilAttachment<'a> {
...
+///
+/// Corresponds to [WebGPU `GPURenderBundleDescriptor`](
+/// https://gpuweb.github.io/gpuweb/#dictdef-gpurenderbundledescriptor).
 pub type RenderBundleDescriptor<'a> = wgt::RenderBundleDescriptor<Label<'a>>;
 /// Describes a [`Texture`].
+///
+/// For use with [`Device::create_texture`].
+///
+/// Corresponds to [WebGPU `GPURenderBundleDescriptor`](
+/// https://gpuweb.github.io/gpuweb/#dictdef-gpurenderbundledescriptor).
```
jimblandy, line 1243: This isn't the correct link.

jimblandy (changes requested):
Thank you very much for doing this! Some of the links seem to be wrong; see below. If you could fix these and re-submit, that would be great.

wgpu/src/lib.rs
kpreid: Fixed. I reviewed the full diff and did not find any more errors of this type.

jimblandy (approved):
Looks good now. I fixed two more minor things.

PR #2655, opened 2022-05-13
Use fully qualified syntax for some calls.

jimblandy:
Without this change, rust-analyzer cannot determine the type of
cmd_buf.

The rust-analyzer bug is reported as rust-lang/rust-analyzer#12247.

### DIFF
--- a/wgpu-core/src/command/compute.rs
+++ b/wgpu-core/src/command/compute.rs
@@ -309,8 +309,11 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         let (device_guard, mut token) = hub.devices.read(&mut token);
 
         let (mut cmd_buf_guard, mut token) = hub.command_buffers.write(&mut token);
-        let cmd_buf = CommandBuffer::get_encoder_mut(&mut *cmd_buf_guard, encoder_id)
-            .map_pass_err(init_scope)?;
+        // Spell out the type, to placate rust-analyzer.
+        // https://github.com/rust-lang/rust-analyzer/issues/12247
+        let cmd_buf: &mut CommandBuffer<A> =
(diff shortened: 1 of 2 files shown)


wgpu-core/src/command/compute.rs
```diff
@@ -309,8 +309,10 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         let (device_guard, mut token) = hub.devices.read(&mut token);
 
         let (mut cmd_buf_guard, mut token) = hub.command_buffers.write(&mut token);
-        let cmd_buf = CommandBuffer::get_encoder_mut(&mut *cmd_buf_guard, encoder_id)
-            .map_pass_err(init_scope)?;
+        let result = CommandBuffer::get_encoder_mut(&mut *cmd_buf_guard, encoder_id);
+        // Call using fully qualified syntax, to placate rust-analyzer.
+        // https://github.com/rust-lang/rust-analyzer/issues/12247
+        let cmd_buf = MapPassErr::<&mut CommandBuffer<A>, _>::map_pass_err(result, init_scope)?;
```
cwfitzgerald, line 315: How about?

I've actually done this for a lot of other things in my tracker refactor, as I was touching all that code

```suggestion
        let cmd_buf: &mut CommandBuffer<A> = CommandBuffer::get_encoder_mut(&mut *cmd_buf_guard, encoder_id)
```

cwfitzgerald (changes requested):

wgpu-core/src/command/compute.rs
jimblandy: That works too.

cwfitzgerald (approved):
Format mad, but I'm not

jimblandy:
d'oh

PR #2657, opened 2022-05-14
fix: panic in `Storage::get`

SparkyPotato:
**Description**
When an `Id` with `Index` out of bounds of `self.map` is given to `Storage::contains`, `Storage::get`, or `Storage::get_mut`, it panics due to directly indexing the Vec. This PR fixes that, and makes `contains` return false, and `get*` return `Err(InvalidId)` like they should.

This also indirectly fixes a panic when `Device::as_hal` is invoked with an API that is *not* being used at the time.

**Testing**
This was tested by creating a device with `Backends::Vulkan`, and trying to get a `Dx12` backend (`Device::as_hal::<Dx12, _, _>`) from it. It correctly gives the closure a None, instead of panicking as it did earlier.

### DIFF
--- a/wgpu-core/src/hub.rs
+++ b/wgpu-core/src/hub.rs
@@ -150,12 +150,8 @@ impl<T, I: id::TypedId> Storage<T, I> {
     pub(crate) fn contains(&self, id: I) -> bool {
         let (index, epoch, _) = id.unzip();
         match self.map.get(index as usize) {
-            Some(element) => match *element {
-                Element::Vacant => false,
-                Element::Occupied(_, storage_epoch) | Element::Error(storage_epoch, ..) => {
-                    epoch == storage_epoch
-                }
-            },
+            Some(&Element::Vacant) => false,
+            Some(&Element::Occupied(_, storage_epoch) | &Element::Error(storage_epoch, _)) => storage_epoch == epoch,
             None => false,
         }
     }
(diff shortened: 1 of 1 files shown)


wgpu-core/src/hub.rs
```diff
@@ -149,46 +149,59 @@ impl<T, I: id::TypedId> ops::IndexMut<id::Valid<I>> for Storage<T, I> {
 impl<T, I: id::TypedId> Storage<T, I> {
     pub(crate) fn contains(&self, id: I) -> bool {
         let (index, epoch, _) = id.unzip();
-        match self.map[index as usize] {
-            Element::Vacant => false,
-            Element::Occupied(_, storage_epoch) | Element::Error(storage_epoch, ..) => {
-                epoch == storage_epoch
-            }
+        match self.map.get(index as usize) {
```
cwfitzgerald, line 152: Could we avoid the double match here (and elsewhere and then have arms for Some(Occupied) etc

wgpu-core/src/hub.rs
```diff
@@ -149,46 +149,59 @@ impl<T, I: id::TypedId> ops::IndexMut<id::Valid<I>> for Storage<T, I> {
...
-        };
-        assert_eq!(
-            epoch, storage_epoch,
-            "{}[{}] is no longer alive",
-            self.kind, index
-        );
-        result
+        match self.map.get(index as usize) {
+            Some(element) => {
+                let (result, storage_epoch) = match *element {
```
cwfitzgerald, line 169: Here you can explicitly return Err in the None branch to avoid the double match here

cwfitzgerald (changes requested):

cwfitzgerald (approved):
Lgtm

PR #2660, opened 2022-05-15
Accept both parking-lot 0.11 and 0.12, to avoid windows-rs.

jimblandy:
In parking_lot 0.12 and parking_lot_core 0.9.0, those crates switched from the winapi crate to the official Microsoft windows-sys crate. This is fine, except that windows-sys and its dependencies are even larger than winapi. Some users may wish to stick with winapi for the time being; this change allows wgpu to accommodate them.

### DIFF
--- a/wgpu-core/Cargo.toml
+++ b/wgpu-core/Cargo.toml
@@ -30,7 +30,8 @@ codespan-reporting = "0.11"
 copyless = "0.1"
 fxhash = "0.2"
 log = "0.4"
-parking_lot = "0.12"
+# parking_lot 0.12 switches from `winapi` to `windows`; permit either
+parking_lot = ">=0.11,<0.13"
 profiling = { version = "1", default-features = false }
 raw-window-handle = { version = "0.4", optional = true }
 ron = { version = "0.7", optional = true }
--- a/wgpu/Cargo.toml
+++ b/wgpu/Cargo.toml
@@ -112,7 +112,8 @@ version = "0.12"
 [dependencies]
 arrayvec = "0.7"
 log = "0.4"
-parking_lot = "0.12"
+# parking_lot 0.12 switches from `winapi` to `windows`; permit either
+parking_lot = ">=0.11,<0.13"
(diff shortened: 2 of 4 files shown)


jimblandy:
Firefox is trying to put off taking the `windows` crate plunge, so this is needed to allow us to update `wgpu` without local changes.

wgpu-core/Cargo.toml
```diff
@@ -30,7 +30,8 @@ codespan-reporting = "0.11"
 copyless = "0.1"
 fxhash = "0.2"
 log = "0.4"
-parking_lot = "0.12"
+# parking_lot 0.12 switches from `winapi` to `windows`; permit either
+parking_lot = ">=0.11,<=0.12"
```
cwfitzgerald, line 34: I might be wrong here, but I think we need this to be <0.13 as 0.12.1 is technically greater than 0.12

cwfitzgerald (changes requested):
One note, but otherwise lgtm. We should also have a discussion about the timing about when we should take the windows-rs plunge as it is inevitable, but obviously stakeholders have opinions outside of our immediate needs.

wgpu-core/Cargo.toml
jimblandy: Yes - thanks.

wgpu/Cargo.toml
```diff
@@ -283,7 +284,7 @@ web-sys = { version = "0.3.53", features = [
 ]}
 js-sys = "0.3.50"
 wasm-bindgen-futures = "0.4.23"
-parking_lot = "0.12"
+parking_lot = ">=0.11,<=0.12"
```
cwfitzgerald, line 287: Missed this one?

cwfitzgerald (changes requested):

wgpu/Cargo.toml
jimblandy: Thanks.

cwfitzgerald (approved):
Party

PR #2661, opened 2022-05-15
Move set_index_buffer FFI functions back into wgpu.

jimblandy:
It's very odd to have almost all the render pass and compute pass ffi functions in `wgpu` except for the `set_index_buffer` functions, which live in Firefox. I'd like to remove these from Firefox and put them back next to their companions.

These functions were originally removed from wgpu in #1077, because wgpu-native has its own incompatible version of IndexFormat (see that PR for details). However, with wgpu-native#85, that code was removed, so having these functions in `wgpu` should be no longer be a problem for wgpu-native.


cwfitzgerald (approved):
LGTM

PR #2663, opened 2022-05-16
Adapter and Instance as_hal functions

i509VCB:
**Description**
At the moment the inner hal type for Instance and Adapter are not possible to access. This adds `as_hal` functions to Instance and Adapter.

HalApi does gain a new function (`instance_as_hal`) to implement `as_hal` for Instance without resorting to a `Instance::as_(backend)` function for every backend.

### DIFF
--- a/wgpu-core/src/hub.rs
+++ b/wgpu-core/src/hub.rs
@@ -907,6 +907,17 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         }
     }
 
+    /// # Safety
+    ///
+    /// - The raw handle obtained from the hal Instance must not be manually destroyed
+    pub unsafe fn instance_as_hal<A: HalApi, F: FnOnce(Option<&A::Instance>) -> R, R>(
+        &self,
+        hal_instance_callback: F,
+    ) -> R {
+        let hal_instance = A::instance_as_hal(&self.instance);
+        hal_instance_callback(hal_instance)
+    }
+
     pub fn clear_backend<A: HalApi>(&self, _dummy: ()) {
         let mut surface_guard = self.surfaces.data.write();
         let hub = A::hub(self);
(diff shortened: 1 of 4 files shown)


wgpu/src/backend/direct.rs
```diff
@@ -67,6 +75,16 @@ impl Context {
         self.0.create_adapter_from_hal(hal_adapter, PhantomData)
     }
 
+    #[cfg(any(not(target_arch = "wasm32"), feature = "emscripten"))]
```
cwfitzgerald, line 78: This isn't your fault, but this cfg block isn't right as it will fail on normal webgl2. It really should be `any(not(target_arch = "wasm32"), feature = "webgl2")`

wgpu-core/src/hub.rs
```diff
@@ -907,6 +907,15 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         }
     }
 
+    pub unsafe fn instance_as_hal<A: HalApi, F: FnOnce(Option<&A::Instance>) -> R, R>(
+        &self,
+        hal_instance_callback: F,
+    ) -> R {
+        profiling::scope!("as_hal", "Instance");
```
cwfitzgerald, line 914: Don't really need a scope here

cwfitzgerald (changes requested):
Two comments, both fairly minor, but otherwise LGTM

wgpu/src/lib.rs
```diff
@@ -1632,6 +1632,21 @@ impl Instance {
         }
     }
 
+    /// Returns the inner hal Instance using a callback. The hal instance will be `None` if the
+    /// backend type argument does not match with this wgpu Instance
+    ///
+    /// # Safety
+    ///
+    /// - The raw handle obtained from the hal Instance must not be manually destroyed
+    #[cfg(any(not(target_arch = "wasm32"), feature = "emscripten"))]
```
cwfitzgerald, line 1641: These ones aren't changed

cwfitzgerald (changes requested):

cwfitzgerald (approved):
Thanks!

PR #2664, opened 2022-05-16
Add DEPTH32FLOAT_STENCIL8 featue

jinleili:
WebGPU spec: https://gpuweb.github.io/gpuweb/#depth32float-stencil8

**Testing**
Modified skybox locally and tested on metal, vulkan(--features=vulkan-portability) and dx12 backends:
```rust
const DEPTH_FORMAT = wgpu::TextureFormat::Depth32FloatStencil8;
fn required_features() -> wgpu::Features {
    wgpu::Features::DEPTH32FLOAT_STENCIL8
}

### CODE REVIEW

File: wgpu/tests/clear_texture.rs [hunk 1]
```diff
@@ -308,7 +309,8 @@ fn clear_texture_tests(
 #[test]
 fn clear_texture_2d_uncompressed() {
     initialize_test(
-        TestParameters::default().features(wgpu::Features::CLEAR_TEXTURE),
+        TestParameters::default()
+            .features(wgpu::Features::CLEAR_TEXTURE | wgpu::Features::DEPTH32FLOAT_STENCIL8),
```
cwfitzgerald, line 313: This needs to separately ask for depth32float as an optional feature and then only test it if it is available. As it stands, we can't run this test on any platform without depth32float.

cwfitzgerald (changes requested):
One comment, otherwise lgtm.

wgpu/tests/clear_texture.rs
jinleili: Sorry, I don't know how to set optional feature in the tests module, is like this enough?
```rust
#[test]
fn clear_texture_d32_s8() {
    initialize_test(
        TestParameters::default()
        .features(wgpu::Features::CLEAR_TEXTURE | wgpu::Features::DEPTH32FLOAT_STENCIL8)
        .specific_failure(Some(wgpu::Backends::GL), None, None, false),
        |ctx| {
            clear_texture_tests(
		&ctx, &[wgpu::TextureFormat::Depth32FloatStencil8],
		false, false,
            );
        },
    )
}
```

cwfitzgerald: Hmm I forgot that the test doesn't have the ability to have optional features. That works fine, though you don't need the fail on GL there.

cwfitzgerald (approved):
Thanks!

PR #2675, opened 2022-05-19
[Gles] Fix clearing depth and stencil at the same time

expenses:
**Description**

Currently in the gles backend, if a depth+stencil buffer with a format like `TextureFormat::Depth24FloatPlusStencil8` is attached to a render pass with both the depth and stencil set to clear, only the depth is cleared. I'm not sure if this happens with native gl, but it certainly does with webgl.

**Testing**

This fixed a problem I had in my project. Here's a before and after of using a stencil buffer for a mirror:

mirror_before_and_after

### DIFF
--- a/wgpu-hal/src/gles/mod.rs
+++ b/wgpu-hal/src/gles/mod.rs
@@ -667,6 +667,7 @@ enum Command {
     ClearColorI(u32, [i32; 4]),
     ClearDepth(f32),
     ClearStencil(u32),
+    ClearDepthAndStencil(f32, u32),
     BufferBarrier(glow::Buffer, crate::BufferUses),
     TextureBarrier(crate::TextureUses),
     SetViewport {
--- a/wgpu-hal/src/gles/mod.rs
+++ b/wgpu-hal/src/gles/mod.rs
@@ -667,6 +667,10 @@ enum Command {
     ClearColorI(u32, [i32; 4]),
     ClearDepth(f32),
     ClearStencil(u32),
+    // Clearing both the depth and stencil buffer individually appears to
+    // result in the stencil buffer failing to clear, atleast in WebGL.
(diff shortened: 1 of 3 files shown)


wgpu-hal/src/gles/mod.rs
```diff
@@ -667,6 +667,7 @@ enum Command {
     ClearColorI(u32, [i32; 4]),
     ClearDepth(f32),
     ClearStencil(u32),
+    ClearDepthAndStencil(f32, u32),
```
cwfitzgerald, line 670: Could you just write a note why we need this?

cwfitzgerald (approved):
Looks good, kinda weird we need it, but this is also an optimization, so totally fine with me

expenses:
>  kinda weird we need it

Wait.... Could it be that the problem is actually that the depth buffer doesn't get cleared in the area under the stencil, because we're still stencil testing or something? I need to test what happens if we clear the stencil buffer first. Regardless though, this PR is still good because we shouldn't have to think about this stuff

cwfitzgerald:
That's very plausible, not sure how stencil interacts with clears on GL

expenses:
> Wait.... Could it be that the problem is actually that the depth buffer doesn't get cleared in the area under the stencil, because we're still stencil testing or something? I need to test what happens if we clear the stencil buffer first. Regardless though, this PR is still good because we shouldn't have to think about this stuff

Nope, switching the two around doesn't seem to make a difference. Odd. Oh well, I've added some documentation

wgpu-hal/src/gles/mod.rs
expenses: Done

cwfitzgerald (approved):
Thanks!

PR #2677, opened 2022-05-20
Vulkan: Allow creating hal Instance and OpenDevice with additional extensions

i509VCB:
**Description**
Vulkan being a very explicit graphics API gates functionality behind extensions that must be enabled when an instance or device is created.

Out of tree wgpu extension libraries may want to enable extensions without duplicating a bunch of wgpu-hal code (which would have to be done with creating a raw device). This adds two functions to Vulkan's wgpu-hal types, `Instance::init_with_extensions` and `Adapter::open_with_extensions` as alternatives to wgpu-hal's `Instance::init` and `Adapter::open` that allow specifying additional Vulkan extensions to enable when creating an Instance or Device.

### DIFF
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -1360,18 +1360,18 @@ impl super::Adapter {
 
         Ok(crate::OpenDevice { device, queue })
     }
-}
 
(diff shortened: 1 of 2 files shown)


i509VCB:
This isn't going to work with how Vulkan is designed

PR #2678, opened 2022-05-21
Actually use RenderBundleEncoder::set_bind_group in tests.

jimblandy:


cwfitzgerald (approved):
Nice use of render bundles too

PR #2679, opened 2022-05-21
Clean up render bundle bind group tracking.

jimblandy:
In `wgpu_core::command::bundle::BindState`, the `dynamic_offsets` and `is_dirty` flags only make sense when the slot is occupied, so they should be inside the `Option`. Doing so makes `State::bind` into an `ArrayVec<Option<BindState>>`, and cleans up various other bits.

Document some parts of wgpu_core's render bundle code.

### DIFF
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -167,6 +167,16 @@ impl RenderBundleEncoder {
         self.parent_id
     }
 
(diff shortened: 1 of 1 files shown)


jimblandy:
Err, with this change `bind` becomes an `ArrayVec<Option<BindState>>`. That was kind of the whole point. PR description and commit message fixed.

Note, by the way, that a doc commit is split out from the code changes, for easier review.

kvark:
Merged in #2684

PR #2684, opened 2022-05-21
Eliminate wgpu_core::commands::bundle::State::raw_dynamic_offsets.

jimblandy:
This depends on #2679.

This vector's contents always ended up identical to the `RenderBundleEncoder`'s `BasePass`'s `dynamic_offsets` vector, so we can just take values from there instead of copying them.


kvark (approved):
Wanted to scream about why a PR named "eliminate XXX" adds 3x lines more than it removes, but then realized it's classy Jim's comments :)

PR #2689, opened 2022-05-23
Add DEPTH24UNORM_STENCIL8 feature

jinleili:
WebGPU spec: https://gpuweb.github.io/gpuweb/#depth24unorm-stencil8

**Testing**
Tested on metal vk(--features=vulkan-portability) an dx12 backends

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -185,6 +185,15 @@ bitflags::bitflags! {
         ///
         /// This is a web and native feature.
         const DEPTH_CLIP_CONTROL = 1 << 0;
+        /// Allows for explicit creation of textures of format [`TextureFormat::Depth24UnormStencil8`]
+        ///
+        /// Supported platforms:
+        /// - Vulkan (some)
+        /// - DX12
+        /// - Metal (Macs with amd GPUs)
+        ///
+        /// This is a web and native feature.
+        const DEPTH24UNORM_STENCIL8 = 1 << 1;
         /// Allows for explicit creation of textures of format [`TextureFormat::Depth32FloatStencil8`]
         ///
         /// Supported platforms:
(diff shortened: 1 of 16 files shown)


wgpu-types/src/lib.rs
```diff
@@ -185,6 +185,15 @@ bitflags::bitflags! {
         ///
         /// This is a web and native feature.
         const DEPTH_CLIP_CONTROL = 1 << 0;
+        /// Allows for explicit creation of textures of format [`TextureFormat::Depth24UnormStencil8`]
+        ///
+        /// Supported platforms:
+        /// - Vulkan (some)
+        /// - DX12
+        /// - Metal (Intel Mac with discrete GPU)
```
cwfitzgerald, line 193:
```suggestion
        /// - Metal (Macs with intel or amd GPUs)
```

cwfitzgerald (changes requested):
One nit, otherwise lgtm

wgpu-types/src/lib.rs
jinleili: Tested on MacBook Pro 2017, only amd GPU supports this feature, so change to "Macs with and GPUs"?

cwfitzgerald: The mac feature tables say this is supported on Intel macs too, which is where I was getting my wording from. So they not actually support it?

jinleili: This additional text is marked in the feature tables:
> Support for this pixel format varies by device. Query the depth24Stencil8PixelFormatSupported property of MTLDevice to determine support. 

And `depth24Stencil8PixelFormatSupported` property is only available on [macOS](https://developer.apple.com/documentation/metal/mtldevice/1433371-depth24stencil8pixelformatsuppor)

截屏2022-05-30 16 57 43

cwfitzgerald: Woah, I totally missed that footnote!

cwfitzgerald (approved):

PR #2690, opened 2022-05-23
Perform "valid to use with" checks when recording render bundles.

jimblandy:
Also, update module docs for `wgpu_core::commands::bundle`.

### DIFF
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -263,6 +263,15 @@ impl RenderBundleEncoder {
                 } => {
                     let scope = PassErrorScope::SetBindGroup(bind_group_id);
 
+                    let bind_group: &binding_model::BindGroup<A> = state
+                        .trackers
+                        .bind_groups
+                        .add_single(&*bind_group_guard, bind_group_id)
+                        .ok_or(RenderCommandError::InvalidBindGroup(bind_group_id))
+                        .map_pass_err(scope)?;
+                    self.check_valid_to_use(bind_group.device_id.value)
+                        .map_pass_err(scope)?;
+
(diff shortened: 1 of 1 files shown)


wgpu-core/src/validation.rs
```diff
@@ -1186,3 +1186,7 @@ impl Interface {
         Ok(outputs)
     }
 }
+
+#[derive(Clone, Debug, Error)]
+#[error("GPUObject is not valid to use with method target")]
```
cwfitzgerald, line 1191: Could we make this error a bit clearer, it's very vague

cwfitzgerald (changes requested):
One small comment

jimblandy:
(just rebased, comments not addressed)

jimblandy:
okay, comment addressed

wgpu-core/src/command/bundle.rs
```diff
@@ -1238,6 +1298,8 @@ impl<A: HalApi> State<A> {
 /// Error encountered when finishing recording a render bundle.
 #[derive(Clone, Debug, Error)]
 pub(super) enum RenderBundleErrorInner {
+    #[error("resource is not valid to use with this render bundle")]
```
cwfitzgerald, line 1301:
```suggestion
    #[error("resource is not valid to use with this render bundle because the resource and the bundle come from different devices")]
```

In an ideal world we would also say which resource, and the identities of both devices, but I'm not sure we have all the infrastructure for that right now

cwfitzgerald (changes requested):
One more error comment, this time with a suggestion.

cwfitzgerald (approved):
Thanks!

PR #2691, opened 2022-05-23
Update deno

crowlKats:
Discrepancies added to #2170

### DIFF
--- a/deno_webgpu/src/01_webgpu.js
+++ b/deno_webgpu/src/01_webgpu.js
@@ -154,15 +154,35 @@
     }
   }
 
-  class GPUOutOfMemoryError extends Error {
-    name = "GPUOutOfMemoryError";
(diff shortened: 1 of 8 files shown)


deno_webgpu/src/01_webgpu.js
```diff
@@ -154,15 +154,35 @@
     }
   }
 
-  class GPUOutOfMemoryError extends Error {
-    name = "GPUOutOfMemoryError";
+  class GPUError extends Error {
```
crowlKats, line 157: unsure about correctness

kvark (approved):

deno_webgpu/src/01_webgpu.js
kvark: I don't have an answer, but I also don't think this would be a blocker for the PR

PR #2697, opened 2022-05-28
Add links for SpirV documents.

huandzh:
**Connections**

None

**Description**

- **Problem:** It is not obvious for users to create spriv shader modules when consulting docs of `wgpu::ShaderSource::SpirV`. The signature of `SpirV(Cow<'a, [u32]>)` alone doesn't given enough information. And the return type of `include_spirv!` is also not documented. 
- See:
   - https://docs.rs/wgpu/0.12.0/wgpu/enum.ShaderSource.html#variant.SpirV 
   - https://docs.rs/wgpu/0.12.0/wgpu/macro.include_spirv.html

- **Proposed documents enhancement:**
   - Add links of `include_spirv!` and `util::make_spirv`
   - Add return type of `include_spirv!`

**Testing**

Has built documents and checked added links with the following commands:

```shell
 cargo doc --features spirv --open
```

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -788,6 +788,8 @@ impl Drop for ShaderModule {
 #[non_exhaustive]
 pub enum ShaderSource<'a> {
     /// SPIR-V module represented as a slice of words.
+    ///
+    /// See also: [`util::make_spirv`], [`include_spirv`]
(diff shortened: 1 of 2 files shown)


cwfitzgerald (approved):
Thanks!

PR #2699, opened 2022-05-29
Try not dirtying a bundle's vertex buffer when the stride/step mode c…

jimblandy:
Just want CI


jimblandy:
(I now know that I can get CI just by editing the .yaml file - apologies for the noise)

PR #2700, opened 2022-05-29
Implement submission indexes

cwfitzgerald:
**Connections**

Connected with #2698 and our whole backpressure story.

**Description**

This implements submission indexes so that compute workloads can properly implement backpressure.

I also found issues with our lifetime management, which is why every unit test I wrote is skipped. We are dropping the buffer immediately after the submission instead of properly waiting for it to finish gpu side execution. I tested these unit tests back before the tracker rewrite, so the tracker rewrite isn't at fault.

**Testing**

It has tests!


jimblandy (approved):
Pretty straightforward. I do have a suggestion.

PR #2704, opened 2022-05-29
Check that all vertex outputs are consumed by the fragment shader

cwfitzgerald:
**Connections**

https://github.com/gfx-rs/naga/issues/1945 only happens when this rule is violated. 

**Description**

We never checked that all vertex outputs are consumed by the fragment shader.

**Testing**

Ran tests, it actually caught a violation in our tests.

### DIFF
--- a/wgpu-core/src/validation.rs
+++ b/wgpu-core/src/validation.rs
@@ -263,6 +263,8 @@ pub enum StageError {
         #[source]
         error: InputError,
     },
+    #[error("location[{location}] is provided by the previous stage output but is not consumed as input by this stage.")]
+    InputNotConsumed { location: wgt::ShaderLocation },
 }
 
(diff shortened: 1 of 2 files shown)


cwfitzgerald:
Uncontroversial and works, merging

wgpu-core/src/validation.rs
```diff
@@ -263,6 +263,8 @@ pub enum StageError {
         #[source]
         error: InputError,
     },
+    #[error("location[{location}] is provided by the previous stage output but is not sonsumed as input by this stage.")]
```
teoxoy, line 266: Just noticed a typo `sonsumed`

cwfitzgerald: Name a more iconic duo than cwfitzgerald and typos.

Thanks!

PR #2706, opened 2022-05-31
gles: fix surface supported formats

jinleili:
**Testing**
Tested on M1 Mac via `--features=angle`

### DIFF
--- a/wgpu-hal/src/gles/adapter.rs
+++ b/wgpu-hal/src/gles/adapter.rs
@@ -706,18 +706,20 @@ impl crate::Adapter<super::Api> for super::Adapter {
     ) -> Option<crate::SurfaceCapabilities> {
         if surface.presentable {
             Some(crate::SurfaceCapabilities {
-                formats: if surface.supports_srgb() {
-                    vec![
-                        wgt::TextureFormat::Rgba8UnormSrgb,
-                        #[cfg(not(target_arch = "wasm32"))]
-                        wgt::TextureFormat::Bgra8UnormSrgb,
-                    ]
-                } else {
-                    vec![
+                formats: {
+                    let mut formats = vec![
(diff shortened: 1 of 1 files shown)


cwfitzgerald (commented):
Why the switch in order?

jinleili:
Not about the order, code before:
https://github.com/gfx-rs/wgpu/blob/8063edc6482cfc828895ec0feaa446767fecc510/wgpu-hal/src/gles/adapter.rs#L709-L721
change to:
```rust
formats: {
	let mut formats = vec![
		wgt::TextureFormat::Bgra8Unorm,
	];
	
	if surface.supports_srgb() {
		formats.append(&mut vec![
			wgt::TextureFormat::Rgba8UnormSrgb,
			wgt::TextureFormat::Bgra8UnormSrgb,
		]);
	}
	formats
},
```

cwfitzgerald:
I guess more my question is that this now makes the preferred format the non-srgb formats on GL and I don't think that's what we want.

jinleili:
Maybe I misunderstood [WebGPU spec](https://gpuweb.github.io/gpuweb/#dom-gpu-getpreferredcanvasformat)，that says:
> Must only return ["rgba8unorm"](https://gpuweb.github.io/gpuweb/#dom-gputextureformat-rgba8unorm) or ["bgra8unorm"](https://gpuweb.github.io/gpuweb/#dom-gputextureformat-bgra8unorm).

cwfitzgerald:
While you understand correctly, native swapchains don't need to play by webgpu's rules (and we don't, almost all backends return srgb formats), and the lack of texture views on gles means we can't implement what webgpu wants (make a unorm texture, then create an srgb view) so we should expose srgb by default as the preferred format.

Issue #2676, opened 2022-05-19
Improve Invalid Scissor Rect Error Message


cwfitzgerald:
The error with an invalid scissor rect doesn't actually say what is invalid and what the possible valid values are.

PR #2713, opened 2022-06-02
Improve InvalidScissorRect error message

jinleili:
**Connections**
https://github.com/gfx-rs/wgpu/issues/2676

**Testing**
Calling RenderPass `set_scissor_rect` with illegal arguments will output the following error message:
```sh

Caused by:
    In a RenderPass
      note: encoder = `<CommandBuffer-(0, 1, Metal)>`
    In a set_scissor_rect command
    Scissor Rect { x: 1000, y: 1000, w: 601, h: 200 } is not contained in the render target Extent3d { width: 1600, height: 1200, depth_or_array_layers: 1 }

```


jinleili:
Okay, I can do that.

PR #2716, opened 2022-06-02
Fixed builtin(primitive_index) for vulkan backend

kwillemsen:
**Description**

The `primitive_index` builtin in wgsl was not usable on the vulkan backend, because the adapter did not properly advertise the spirv `Geometry` capability. This has been fixed by exposing `Geometry` when the adapter supports `SHADER_PRIMITIVE_INDEX`.

**Testing**

Two new unit tests select their fragment shader output based on `primitive_index`.

### DIFF
--- a/wgpu/tests/shader_primitive_index/mod.rs
+++ b/wgpu/tests/shader_primitive_index/mod.rs
@@ -1,0 +1,254 @@
+use std::num::NonZeroU32;
+
+use wgpu::util::DeviceExt;
+
+use crate::common::{initialize_test, TestParameters, TestingContext};
+
+//
+// These tests render two triangles to a 2x2 render target. The first triangle
+// in the vertex buffer covers the bottom-left pixel, the second triangle
+// covers the top-right pixel.
+// XY layout of the render target, with two triangles:
+//
+//     (-1,1)   (0,1)   (1,1)
+//        +-------+-------+
+//        |       |   o   |
+//        |       |  / \  |
+//        |       | /   \ |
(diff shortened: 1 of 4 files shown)


wgpu/tests/shader_primitive_index/mod.rs
```diff
@@ -0,0 +1,253 @@
...
+    drop(rpass);
+
+    ctx.queue.submit(Some(encoder.finish()));
+
+    let data = capture_rgba_u8_texture(ctx, color_texture, texture_size);
+
+    assert_eq!(data, expected);
+}
+
+fn align_to(n: u32, alignment: u32) -> u32 {
```
cwfitzgerald, line 201: Is there anywhere else we could share this code with? I feel like we have it in a couple other places in the codebase.

cwfitzgerald (changes requested):
This is really great stuff, the test is especially great! I have one very minor comment and there are CI issue

wgpu/tests/shader_primitive_index/mod.rs
kwillemsen: I've added changes that move this function to the `wgpu::util` namespace and replaced the occurences I could find. I've also changed the implementation to a modulo operation with an explicit check for zero. That seems to be the more popular version in `wgpu.rs` :-)

cwfitzgerald (approved):
Great stuff!

cwfitzgerald:
Just need conflict resolution and we're good

PR #2724, opened 2022-06-03
Acquire texture:  `Option<std::time::Duration>` timeouts

rib:
**Description**

The Vulkan backend wasn't correctly interpreting the u32 timeout value for Surface `acquire_texture` considering the special meaning for `u32::MAX` that should mean 'no timeout'.

The Vulkan `vkAcquireNextImageKHR` API takes a 64bit timeout in nanoseconds and `u32::MAX` timeout in milliseconds should have been converted to a `u64::MAX`.

Prior to Android 11 then Android's `vkAcquireNextImageKHR` implementation is also non-conformant and doesn't support timeouts (and will also output verbose log messages if timeouts are given).

To make the intent clearer, this updates the `acquire_texture` API to take a `std::time::Duration` that allows the caller to use whatever units are convenient. Additionally the duration is wrapped in an `Option` so it's clearer to see when no timeout is wanted without needing to reserve a special value to indicate this.

On Android < 11 then any given timeout is now ignored (as with the egl / web / metal backends) and `u64::MAX` is always passed to  `vkAcquireNextImageKHR` 

For reference this version of AcquireNextImageKHR doesn't support timeouts: https://android.googlesource.com/platform/frameworks/native/+/refs/tags/android-mainline-10.0.0_r13/vulkan/libvulkan/swapchain.cpp#1426 and this version does: https://android.googlesource.com/platform/frameworks/native/+/refs/tags/android-mainline-11.0.0_r45/vulkan/libvulkan/swapchain.cpp#1438 (i.e. timeout support was added in Android 11)

**Testing**
I have tested this via an egui Android application I have been working on recently, and this can also be tested with this minimal winit + wgpu Android application: https://github.com/rib/agdk-rust/tree/main/examples/na-winit-wgpu and these same patches based on the 0.12 branch of wgpu: https://github.com/rib/wgpu/tree/acquire_texture_duration_timeouts_v0.12

### DIFF
--- a/wgpu-core/src/present.rs
+++ b/wgpu-core/src/present.rs
@@ -123,7 +123,12 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         let _ = device;
 
         let suf = A::get_surface_mut(surface);
-        let (texture_id, status) = match unsafe { suf.raw.acquire_texture(FRAME_TIMEOUT_MS) } {
+        let (texture_id, status) = match unsafe {
+            suf.raw
+                .acquire_texture(Some(std::time::Duration::from_millis(
+                    FRAME_TIMEOUT_MS as u64,
+                )))
+        } {
             Ok(Some(ast)) => {
                 let clear_view_desc = hal::TextureViewDescriptor {
                     label: Some("(wgpu internal) clear surface texture view"),
--- a/wgpu-hal/src/lib.rs
+++ b/wgpu-hal/src/lib.rs
@@ -193,10 +193,19 @@ pub trait Surface<A: Api>: Send + Sync {
 
     unsafe fn unconfigure(&mut self, device: &A::Device);
 
+    /// Returns the next texture to be presented by the swapchain for drawing
+    ///
+    /// A `timeout` of `None` means to wait indefinitely, with no timeout.
+    ///
+    /// # Portability
+    ///
+    /// Some backends can't support a timeout when acquiring a texture and
+    /// the timeout will be ignored.
+    ///
     /// Returns `None` on timing out.
     unsafe fn acquire_texture(
         &mut self,
-        timeout_ms: u32,
+        timeout: Option<std::time::Duration>,
     ) -> Result<Option<AcquiredSurfaceTexture<A>>, SurfaceError>;
     unsafe fn discard_texture(&mut self, texture: A::SurfaceTexture);
 }
--- a/wgpu-hal/src/vulkan/instance.rs
+++ b/wgpu-hal/src/vulkan/instance.rs
@@ -214,10 +214,13 @@ impl super::Instance {
     /// - `raw_instance` must be created respecting `driver_api_version`, `extensions` and `flags`
     /// - `extensions` must be a superset of `required_extensions()` and must be created from the
     ///   same entry, driver_api_version and flags.
+    /// - `android_sdk_version` is ignored and can be `0` for all platforms besides Android
+    #[allow(clippy::too_many_arguments)]
     pub unsafe fn from_raw(
         entry: ash::Entry,
         raw_instance: ash::Instance,
         driver_api_version: u32,
+        android_sdk_version: u32,
         extensions: Vec<&'static CStr>,
         flags: crate::InstanceFlags,
         has_nv_optimus: bool,
@@ -283,6 +286,7 @@ impl super::Instance {
                 entry,
                 has_nv_optimus,
                 driver_api_version,
+                android_sdk_version,
             }),
             extensions,
         })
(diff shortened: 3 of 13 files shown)


wgpu-hal/src/vulkan/instance.rs
```diff
@@ -557,6 +559,18 @@ impl crate::Instance<super::Api> for super::Instance {
             layers
         };
 
+        #[cfg(target_os = "android")]
+        let platform_version = {
```
cwfitzgerald, line 563: How about android_platform_version just to make it clear that this value is meaningless on other platforms.

cwfitzgerald (changes requested):
This is great! One nit, but otherwise LGTM

wgpu-hal/src/vulkan/instance.rs
rib: yeah, sounds good to me; maybe `android_sdk_version` to help clarify that it won't have a value like 10 for Android 10 but rather will be based on these SDK version codes: https://developer.android.com/reference/android/os/Build.VERSION_CODES

wgpu-core/src/present.rs
```diff
@@ -10,6 +10,7 @@ extract it from the hub.
 !*/
 
 use std::borrow::Borrow;
+use std::time::Duration;
```
cwfitzgerald, line 13: Could you just use this directly vs using it

wgpu-hal/src/lib.rs
```diff
@@ -196,7 +196,7 @@ pub trait Surface<A: Api>: Send + Sync {
     /// Returns `None` on timing out.
     unsafe fn acquire_texture(
         &mut self,
-        timeout_ms: u32,
+        timeout: Option<std::time::Duration>,
```
cwfitzgerald, line 199: Should add to the docs what None means

cwfitzgerald (changes requested):
Sorry two more nits I missed on the first pass

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -557,6 +560,20 @@ impl crate::Instance<super::Api> for super::Instance {
             layers
         };
 
+        #[cfg(target_os = "android")]
+        let android_sdk_version = {
+            // See: https://developer.android.com/reference/android/os/Build.VERSION_CODES
+            let mut prop = android_properties::getprop("ro.build.version.sdk");
```
Gordon-F, line 566: Can we avoid panics in this code? For example, if something goes wrong, just log error and return 0.

rib: can just fallback to a version to zero and hope for the best yeah, in this case it'd also have the intended effect if disabling the use of timeouts.

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -214,10 +214,12 @@ impl super::Instance {
     /// - `raw_instance` must be created respecting `driver_api_version`, `extensions` and `flags`
     /// - `extensions` must be a superset of `required_extensions()` and must be created from the
     ///   same entry, driver_api_version and flags.
+    #[allow(clippy::too_many_arguments)]
     pub unsafe fn from_raw(
         entry: ash::Entry,
         raw_instance: ash::Instance,
         driver_api_version: u32,
+        android_sdk_version: u32,
```
cwfitzgerald, line 222: Could you add to the from_raw documentation that this can be set to zero on all other platforms

wgpu-hal/src/lib.rs
```diff
@@ -193,10 +193,19 @@ pub trait Surface<A: Api>: Send + Sync {
 
     unsafe fn unconfigure(&mut self, device: &A::Device);
 
+    /// Returns the next texture to be presented by the swapchain for drawing
+    ///
+    /// A `timeout` of `None` mean to wait indefinitely, with no timeout.
```
cwfitzgerald, line 198:
```suggestion
    /// A `timeout` of `None` means to wait indefinitely, with no timeout.
```
Typo

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -557,6 +560,21 @@ impl crate::Instance<super::Api> for super::Instance {
             layers
         };
 
+        #[cfg(target_os = "android")]
+        let android_sdk_version = {
+            // See: https://developer.android.com/reference/android/os/Build.VERSION_CODES
+            let mut prop = android_properties::getprop("ro.build.version.sdk");
+            if let Some(val) = prop.value() {
+                u32::from_str_radix(&val, 10)
+                    .expect("Failed to parse ro.build.version.sdk property")
```
cwfitzgerald, line 569: Could this also log on failure, printing the value it tried and failed to parse?

cwfitzgerald (changes requested):
Third times the charm

rib:
ah, oops I also missed removing the the `expect()` from parsing the sdk version. I'll also switch to using `val.parse::<u32>()` as suggested by clippy / CI.

wgpu-hal/src/vulkan/instance.rs
rib: right, yeah I just realized I missed this, by bad - I'll update and also use `val.parse::<u32>()`  as suggested by clippy

cwfitzgerald (approved):
Thanks!

rib:
btw just pushed one last second tweak to include the value of the property in case of parse failure, as suggested

cwfitzgerald (approved):
Thanks!

PR #2725, opened 2022-06-03
[gles] Handle cubemap copies

expenses:
**Description**

Currently you can't copy a texture to a cubemap in the gles backend because it wrongly uses the `glow::TEXTURE_2D` target. I've pretty hackily set things up so we check how many array layers the target has, and if it has 6 then we write to the correct cubemap face.

**Testing**

Tested in Chrome. ~~I get a `WebGL: INVALID_OPERATION: bindTexture: textures can not be used with multiple targets` error but it functions at least.~~ Fixed that now.

### DIFF
--- a/wgpu-hal/src/gles/queue.rs
+++ b/wgpu-hal/src/gles/queue.rs
@@ -332,20 +332,10 @@ impl super::Queue {
                     );
                 }
 
+                gl.bind_texture(dst_target, Some(dst));
                 if dst_array_layer_count == 6 {
-                    let cube_map_target = match copy.dst_base.array_layer {
-                        0 => glow::TEXTURE_CUBE_MAP_POSITIVE_X,
-                        1 => glow::TEXTURE_CUBE_MAP_NEGATIVE_X,
-                        2 => glow::TEXTURE_CUBE_MAP_POSITIVE_Y,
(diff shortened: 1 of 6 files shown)


wgpu-hal/src/gles/queue.rs
```diff
@@ -333,7 +333,18 @@ impl super::Queue {
                 }
 
                 gl.bind_texture(dst_target, Some(dst));
-                if is_layered_target(dst_target) {
+                if dst_array_layer_count == 6 {
```
cwfitzgerald, line 336: Do we have other such assumptions in the GLES backend? I'm just worrieed that this is a pretty big footgun.

expenses: We have: 

https://github.com/gfx-rs/wgpu/blob/46420212d71edb6daa1337284f6d94efa7c4ceb3/wgpu-hal/src/gles/device.rs#L568

we could copy an `is_cubemap` bool into the `Texture` and then into this command.

cwfitzgerald: Yeah that would probably be the most consistent

expenses: Done

cwfitzgerald (approved):
TYSM!

PR #2729, opened 2022-06-03
[WebGL] Add a downlevel capability for rendering to floating point textures

expenses:
**Description**

The [`EXT_color_buffer_float`](https://www.khronos.org/registry/webgl/extensions/EXT_color_buffer_float/) extension let's you rendering to floating point textures.

Not sure if passing the entire capabilities struct to the adapter is the way to go.


wgpu-hal/src/dx11/adapter.rs
```diff
@@ -14,6 +14,7 @@ impl crate::Adapter<super::Api> for super::Adapter {
     unsafe fn texture_format_capabilities(
         &self,
         format: wgt::TextureFormat,
+        capabilities: &crate::Capabilities,
```
cwfitzgerald, line 17: I don't think this is the best way to go about this. We should pass along what we need internal to the backends.

cwfitzgerald (changes requested):
I agree with the problem you identified, but otherwise looks okay

wgpu-hal/src/dx11/adapter.rs
expenses: Done

wgpu-types/src/lib.rs
```diff
@@ -1026,6 +1026,9 @@ bitflags::bitflags! {
         ///
         /// GLES/WebGL don't support this.
         const DEPTH_TEXTURE_AND_BUFFER_COPIES = 1 << 13;
+
+        /// Supports rendering to floating-point textures.
+        const COLOR_ATTACHMENT_FLOAT = 1 << 14;
```
cwfitzgerald, line 1031: This needs to be added to the DX11 backend too. The two downlevel backends start from no flags and add them, the other backends start with all flags and remove them.

cwfitzgerald (changes requested):
One comment, otherwise LGTM.

wgpu-types/src/lib.rs
cwfitzgerald, line 1031: I realize now, I don't think this is a scalable way to do this. Long term we want one or two sets of formats and usages for downlevel backends. In the short term though, the feature should just say "webgpu_format_features" or something and when it's not there, you need to call https://docs.rs/wgpu/latest/wgpu/struct.Adapter.html#method.get_texture_format_features to determine if you can render to that format at all. This way we can support the diverse set of texture formats that real downlevel devices support.

wgpu-hal/src/gles/mod.rs
```diff
@@ -188,6 +188,7 @@ struct AdapterShared {
 
 pub struct Adapter {
     shared: Arc<AdapterShared>,
+    downlevel_flags: wgt::DownlevelFlags,
```
cwfitzgerald, line 191: Because of my other comment, I feel like we should store "is_ext_color_buffer_float_supported" as a boolean, and it should live inside the shared adapter.

cwfitzgerald (changes requested):
Backend works LGTM, comments on how we expose it

wgpu-hal/src/gles/mod.rs
expenses: Done

wgpu-types/src/lib.rs
expenses: In this PR what should I be doing for this?

cwfitzgerald: In this PR, lets make a feature called `WEBGPU_TEXTURE_FORMAT_SUPPORT` or something, with the description of "Supports all the texture usages described in WebGPU. If this isn't supported, you should call get_texture_format_features to get how you can use textures of a given format" and turn it on for dx11 FL11+ and off for GLES.

wgpu-hal/src/dx11/adapter.rs
```diff
@@ -129,7 +130,8 @@ impl super::Adapter {
         }
 
         if feature_level >= FL11_0 {
-            downlevel |= wgt::DownlevelFlags::INDIRECT_EXECUTION;
+            downlevel |= wgt::DownlevelFlags::INDIRECT_EXECUTION
+                | wgt::DownlevelFlags::WEBGPU_TEXTURE_FORMAT_SUPPORT;
```
cwfitzgerald, line 134:
```suggestion
            downlevel |= wgt::DownlevelFlags::INDIRECT_EXECUTION;
            downlevel |= wgt::DownlevelFlags::WEBGPU_TEXTURE_FORMAT_SUPPORT;
```
nit to bring in line with rest of the code.

wgpu-types/src/lib.rs
```diff
@@ -1026,6 +1026,13 @@ bitflags::bitflags! {
...
```
cwfitzgerald, line 1031: We should have this feature at all, as you can just query support for the texture you need when WEBGPU_TEXTURE_FORMAT_SUPPORT isn't available.

cwfitzgerald (changes requested):
N-th time's the charm

cwfitzgerald (commented):
Great!

cwfitzgerald (approved):

Issue #2540, opened 2022-03-14
Panic: called glDebugMessageCallback but it was not loaded


davidar:
**Description**
The examples are failing to run in recent master (on Windows). I bisected it down to 476b6a117713592cb3fc49b3ad112618fa7238d8

**Repro steps**
`cargo run --example cube`

**Expected vs observed behavior**
Should run normally, instead it panics with `thread 'main' panicked at 'called glDebugMessageCallback but it was not loaded.'`

It runs if I set explicitly set `WGPU_BACKEND` to vulkan or dx12, but looks like it's now defaulting to gl for some reason

**Extra materials**
```
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a\/library\std\src\panicking.rs:498
   1: core::panicking::panic_fmt
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a\/library\core\src\panicking.rs:116
   2: glow::gl46::go_panic_because_fn_not_loaded
             at C:\Users\d\.cargo\registry\src\github.com-1ecc6299db9ec823\glow-0.11.1\src\gl46.rs:4232
   3: glow::gl46::call_atomic_ptr_2arg
             at C:\Users\d\.cargo\registry\src\github.com-1ecc6299db9ec823\glow-0.11.1\src\gl46.rs:4314
   4: glow::gl46::struct_commands::GlFns::DebugMessageCallback
             at C:\Users\d\.cargo\registry\src\github.com-1ecc6299db9ec823\glow-0.11.1\src\gl46.rs:10199
   5: glow::native::impl$2::debug_message_callback<void (*)(u32,u32,u32,u32,str)>
             at C:\Users\d\.cargo\registry\src\github.com-1ecc6299db9ec823\glow-0.11.1\src\native.rs:2299
   6: wgpu_hal::gles::egl::impl$11::enumerate_adapters
             at .\wgpu-hal\src\gles\egl.rs:840
   7: wgpu_core::instance::impl$6::request_adapter::gather<wgpu_hal::gles::Api,core::marker::PhantomData<wgpu_core::id::Id<wgpu_core::instance::Adapter<wgpu_hal::empty::Api> > > >
             at .\wgpu-core\src\instance.rs:663
   8: wgpu_core::hub::Global<wgpu_core::hub::IdentityManagerFactory>::request_adapter<wgpu_core::hub::IdentityManagerFactory>
[...]

PR #2730, opened 2022-06-04
Fix default backends selection when running examples at the workspace root

jinleili:
**Connections**
https://github.com/gfx-rs/wgpu/issues/2540

**Description**
Since wgpu-info as a workspace  `default-members`  item depends on wgpu and enabled `angle`  `vulkan-portability` features, it causes  `#[cfg(vulkan)]` `#[cfg(gl)]`  compile conditions alway pass when  run examples at the workspace root .
https://github.com/gfx-rs/wgpu/blob/9e3cd08e5913bbb22e4d5b1f3fd30dccbf4ba9c7/wgpu-core/build.rs#L10-L19

**Testing**
Tested examples at the workspace root and `wgpu` package directory

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -218,7 +218,7 @@ jobs:
         shell: bash
         run: |
           # run wgpu-info
-          cargo run --bin wgpu-info
+          cargo run -p wgpu-info
           # run unit and player tests
           cargo nextest run -p wgpu-types -p wgpu-hal -p wgpu-core -p player --no-fail-fast
           # run native tests
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,13 +8,13 @@ members = [
     "wgpu-hal",
     "wgpu-info",
     "wgpu-types",
-    "run-wasm",
+    "run-wasm"
 ]
 exclude = [
     "cts_runner",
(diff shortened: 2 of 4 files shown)


cwfitzgerald:
I don't think this actually solves the base problem, just cures the symptoms. The problem here isn't that we are enabling the GL backend on windows, we should always be allowed to do that, the issue is that we're crashing when trying to call a function that we shouldn't be calling.

jinleili:
@cwfitzgerald Although not solving the crash, I think what this commit solves is the inconsistent results of running the examples in the workspace root and wgpu package directory without specifying `WGPU_BACKEND`. The same issue exists not only on Windows, but also on macOS.

Executing `cargo run --example xx` from wgpu package directory on macOS, always using the metal backend

The result of executing `cargo run --example xx` from the workspace root on macOS.
- When no "angle" is installed, an alert window pops up:
no_angle

- When vulkan is installed,   always using the vk backend

cwfitzgerald:
Mac why do you have to be like this.

Alright, I think the best option is to add a message to wgpu-info saying "for gl in non Linux add the angle feature, for vk on Mac add the vulkan-portability feature" , then remove the default features from wgpu-info, then make sure CI covers the angle and vk portability cases.

Issue #2032, opened 2021-10-07
Water Example Not Working on WebGL


zicklag:
**Description**

**Repro steps**
Run `./run-wasm-example.sh water` in the WGPU repo.

**Expected vs observed behavior**

The water is invisible, when it should be visible.

**Extra materials**

Observed:

![image](https://user-images.githubusercontent.com/25393315/136456630-e2141f7e-c2bd-422c-b3de-fa255bd6f8c3.png)

Expected:

![image](https://user-images.githubusercontent.com/25393315/136456713-2cc9ce04-0767-4684-9330-59b80c914b7b.png)

**Platform**

WebGL2

---

The issue with the water rendering has something to do with the water shader. Context on exactly what causes the shading issue is here: https://github.com/gfx-rs/wgpu/pull/1686#issuecomment-884596050.

I don't currently understand the fix necessary.

cwfitzgerald:
This is expected behavior due to rods

zicklag:
Should we update the example shader to use a different entrypoint for WebGL, so that we can get it displaying, just without the extra blurring around the water border ( like this https://github.com/gfx-rs/wgpu/pull/1686#issuecomment-884596050 )?

PR #2732, opened 2022-06-04
Make Water Visible in Water Example on WebGL

zicklag:
Uses a separate entrypoint for the water fragment shader on WebGL to
avoid the unsupported depth calculation.

**Connections**
Addresses #2032

**Description**
Uses an alternate fragment shader entrypoint on for the water example on WebGL. Looks slightly worse, but actually renders the water.

This should mean we could put this example on the website with the others.

**Testing**
Tested in my browser: `Brave Browser 97.1.34.80 unknown` ( a Chrome derivative )

image

Tested on Ubuntu 20.04 Vulkan that the example still works normally after the change.

---

BTW, love the new `cargo run-wasm` utility, and the cleaner `@binding` syntax for WGSL. :)

### DIFF
--- a/wgpu-hal/src/gles/adapter.rs
+++ b/wgpu-hal/src/gles/adapter.rs
@@ -289,6 +289,9 @@ impl super::Adapter {
             wgt::DownlevelFlags::ANISOTROPIC_FILTERING,
             extensions.contains("EXT_texture_filter_anisotropic"),
         );
+        if cfg!(not(target_arch = "wasm32")) {
+            downlevel_flags.set(wgt::DownlevelFlags::READ_ONLY_DEPTH_STENCIL, true);
+        }
 
         let mut features = wgt::Features::empty()
             | wgt::Features::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES
--- a/wgpu/examples/water/main.rs
+++ b/wgpu/examples/water/main.rs
@@ -269,7 +269,7 @@ impl Example {
 impl framework::Example for Example {
     fn init(
         config: &wgpu::SurfaceConfiguration,
-        _adapter: &wgpu::Adapter,
+        adapter: &wgpu::Adapter,
         device: &wgpu::Device,
         queue: &wgpu::Queue,
     ) -> Self {
@@ -504,6 +504,13 @@ impl framework::Example for Example {
             source: wgpu::ShaderSource::Wgsl(Cow::Borrowed(include_str!("water.wgsl"))),
         });
 
+        // Check for read-only depth stencil support so we can use a different fragment shader
+        // entrypoint if it is unsupported on the current adapter.
+        let read_only_depth_stencil = adapter
+            .get_downlevel_capabilities()
+            .flags
+            .contains(wgpu::DownlevelFlags::READ_ONLY_DEPTH_STENCIL);
+
         // Create the render pipelines. These describe how the data will flow through the GPU, and what
         // constraints and modifiers it will have.
         let water_pipeline = device.create_render_pipeline(&wgpu::RenderPipelineDescriptor {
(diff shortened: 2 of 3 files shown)


wgpu/examples/water/main.rs
```diff
@@ -528,7 +528,10 @@ impl framework::Example for Example {
             // Fragment shader and output targets
             fragment: Some(wgpu::FragmentState {
                 module: &water_module,
+                #[cfg(not(feature = "webgl"))]
                 entry_point: "fs_main",
```
kvark, line 532: if you can use run-time conditions, it's always preferred over compile time, since both branches are checked at compilation: `entry_point: if cfg!(something) { ... } else { ... },`

kvark (approved):

wgpu/examples/water/main.rs
zicklag: Ah, good point. I hadn't thought of that before. I pushed an update.

wgpu/examples/water/main.rs
```diff
@@ -528,7 +528,11 @@ impl framework::Example for Example {
             // Fragment shader and output targets
             fragment: Some(wgpu::FragmentState {
                 module: &water_module,
-                entry_point: "fs_main",
+                entry_point: if cfg!(feature = "webgl") {
```
cwfitzgerald, line 531: The split here is about RODS not webgl. We should be checking for the RODS downlevel flag, not based on platform: https://docs.rs/wgpu/0.12.0/wgpu/struct.DownlevelFlags.html#associatedconstant.READ_ONLY_DEPTH_STENCIL

cwfitzgerald (changes requested):

wgpu/examples/water/main.rs
zicklag: Got it, I'll update it.

zicklag: Fixed!

zicklag:
I'm not sure if there is a way to check for depth stencil support on OpenGL, but I also had to add the downlevel flag in the GL backend, and I just set the flag if the target wasn't wasm.

Let me know if there's a way I should detect RODS for OpenGL.

wgpu-hal/src/gles/adapter.rs
```diff
@@ -289,6 +289,9 @@ impl super::Adapter {
             wgt::DownlevelFlags::ANISOTROPIC_FILTERING,
             extensions.contains("EXT_texture_filter_anisotropic"),
         );
+        if cfg!(not(target_arch = "wasm32")) {
+            downlevel_flags.set(wgt::DownlevelFlags::READ_ONLY_DEPTH_STENCIL, true);
+        }
```
cwfitzgerald, line 294: This isn't true, iirc according to the gl spec RODS is always UB. It happens to work because it's not a real race condition but I don't think you can technically do it.

wgpu/examples/water/main.rs
```diff
@@ -528,7 +535,11 @@ impl framework::Example for Example {
             // Fragment shader and output targets
             fragment: Some(wgpu::FragmentState {
                 module: &water_module,
-                entry_point: "fs_main",
+                entry_point: if read_only_depth_stencil {
```
cwfitzgerald, line 538: I realize now this won't be enough, we should have RODS validation against the downlevel limit, and that validation will still cause this to fail even if we're using the new shader - we need to completely remove the bindings from the bgl

cwfitzgerald (changes requested):
More considerations

cwfitzgerald:
Poke from triage: what's the status of this PR? Does the problem still exist?

zicklag:
I haven't tested recently, but as far as I know the problem still exists, but I ran out of time to work on it.

cwfitzgerald:

Issue #2727, opened 2022-06-03
Nothing will be presented if a Surface is created by a non-ui thread (Metal backend)


shi-yan:
I'm trying to create a multi-window wgpu program.

My current design is a single producer multiple consumer architecture. 

The main event loop is the producer, it accepts events and then dispatches them to corresponding windows. 

Each window has its own event loop on a dedicated thread listening to the messages from the main thread. From its dedicated thread, a window also creates its own wgpu command buffer and presents it. 

This design seems to work on Linux with the vulkan/xcb backend.

But it doesn't work on Mac with the Metal/Appkit backend.

I wonder if this is expected? because I can't find confirmation from anywhere.

To illustrate the problem, I have created a testing project using winit (I have also tried another windowing system. and The behavior is the same.) (See Archive.zip)

the sample app has 2 variants,  in `RUST_LOG=trace cargo run --bin multi`,  I follow the above design. As you can see, the window is empty. What's rendered cannot be seen.

the singlethread variant `RUST_LOG=trace cargo run --bin single` simply creates command buffers and presents them from the main thread. Things can be rendered properly. 

If we check the logs printed by wgpu from both variants, we won't find any issues. Both seem to be able to successfully submit buffers. 

I wonder if this is a bug? or if this is expected behavior on Mac? If this is not a bug, I hope this can be mentioned in the document.

Thanks.

[Archive.zip](https://github.com/gfx-rs/wgpu/files/8834227/Archive.zip)

shi-yan:
so I did more investigation. I read the wgpu-hal code and noticed that Surface has a field "main_thread_id".

In the above code, I created the surface in a non-ui thread. So the "main_thread_id" isn't really the main thread's id.

I adjusted my code to create the Surface on the main thread, and things could render now.

shi-yan:
Yes, I double-checked, with raw Metal, a CAMetalLayer can be created in a non-ui thread. 

```objc

// the following can be in a non-ui thread
  auto swap_chain = [CAMetalLayer layer];
  swap_chain.device = device;
  swap_chain.pixelFormat = MTLPixelFormatBGRA8Unorm_sRGB;
  swap_chain.framebufferOnly = YES;
  swap_chain.frame = native_window.frame;
    native_window.contentView.wantsLayer = YES;
  native_window.contentView.layer = swap_chain;
  native_window.opaque = YES;
  native_window.backgroundColor = nil;
```

jinleili:
[...]

PR #2736, opened 2022-06-06
metal: check if in the main thread when calling `create_surface`

jinleili:
**Connections**
https://github.com/gfx-rs/wgpu/issues/2727

**Description**
On macOS and iOS platforms, calling UI APIs in non-ui thread will cause unexpected results. Tell the user exactly what the problem is by checking if it is in the main thread.

Also tried using Objective-C GCD to specify the main thread to execute the code, but `view: *mut c_void` is not a thread-safe object, and I don't think actively jumping threads is something wgpu should do.

**Testing**
Tested on macOS and iOS.

### DIFF
--- a/wgpu-hal/src/metal/surface.rs
+++ b/wgpu-hal/src/metal/surface.rs
@@ -9,7 +9,7 @@ use objc::{
     declare::ClassDecl,
     msg_send,
     rc::autoreleasepool,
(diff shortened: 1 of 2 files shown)


cwfitzgerald (changes requested):
We should add documentation to create_surface and this function it will panic if not called on the main thread.

cwfitzgerald (approved):
Thanks!

PR #2740, opened 2022-06-06
gles: allow clearing index buffers

grovesNL:
**Connections**
Noticed while testing WebGL manually, noticed it was referenced by some other projects `too,` e.g. https://github.com/iced-rs/iced/issues/1241

**Description**
On some browsers like Chrome, it seems like `ELEMENT_ARRAY_BUFFER` can't be used as a destination for `copy_buffer_sub_data` if the source is `COPY_READ_BUFFER`, so we can't copy from `zero_buffer`. Instead we workaround this by uploading zeroes.

We could also consider caching a `Vec` filled with zeroes if we'd like to avoid an allocation, but it might result in several uploads depending on the size of the cached `Vec`.

**Testing**
Tested manually on my own project and the warning disappears. We should probably check it against all wgpu WebGL examples though.

### DIFF
--- a/wgpu-hal/src/gles/queue.rs
+++ b/wgpu-hal/src/gles/queue.rs
@@ -213,19 +213,39 @@ impl super::Queue {
                 ref range,
             } => match dst.raw {
                 Some(buffer) => {
-                    gl.bind_buffer(glow::COPY_READ_BUFFER, Some(self.zero_buffer));
-                    gl.bind_buffer(dst_target, Some(buffer));
-                    let mut dst_offset = range.start;
-                    while dst_offset < range.end {
-                        let size = (range.end - dst_offset).min(super::ZERO_BUFFER_SIZE as u64);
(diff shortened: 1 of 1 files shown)


grovesNL:
I've tested the existing WebGL examples and they seem ok.

cwfitzgerald (approved):
We've committed worse sins

expenses:
Just saw this! I've been dealing with trying to get index buffer to index buffer copies to work. I suppose you could implement that by mapping the source buffer memory and doing `buffer_sub_data_u8_slice` from that right? It'd be pretty (very?) slow, but would at least be implemented.

expenses:
> Just saw this! I've been dealing with trying to get index buffer to index buffer copies to work. I suppose you could implement that by mapping the source buffer memory and doing `buffer_sub_data_u8_slice` from that right? It'd be pretty (very?) slow, but would at least be implemented.

Hmm, it seems I don't actually need to do this. NVM!

PR #2743, opened 2022-06-06
Clean up render bundle index buffer tracking.

jimblandy:
Put all state associated with an established index buffer within an `Option`, so that the Rust types accurately represent value liveness.

Generate `MissingIndexBuffer` errors as needed for `DrawIndexed` and indexed `MultiDrawIndirect` commands.

Remove unused field `bundle::IndexState::pipeline_format`.

This may be easier to review if you look at the commits separately.

### DIFF
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -355,7 +355,6 @@ impl RenderBundleEncoder {
                     pipeline_layout_id = Some(pipeline.layout_id.value);
 
                     state.set_pipeline(
-                        pipeline.strip_index_format,
                         &pipeline.vertex_strides,
                         &layout.bind_group_layout_ids,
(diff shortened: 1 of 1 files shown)


wgpu-core/src/command/bundle.rs
```diff
@@ -1231,6 +1203,38 @@ impl<A: HalApi> State<A> {
...
 
+    /// Set the bundle's current index buffer and its associated parameters.
+    fn set_index_buffer(
+        &mut self,
+        buffer: id::BufferId,
+        format: wgt::IndexFormat,
+        range: Range<wgt::BufferAddress>,
+    ) {
+        match self.index {
+            Some(ref current)
```
cwfitzgerald, line 1214: God can't wait for if let chains

cwfitzgerald (approved):
Better validation is better!

PR #2744, opened 2022-06-06
Don't dirty the vertex buffer for stride/rate changes on bundles.

jimblandy:
`wgpu_core::command::bundle::State::set_pipeline` marks a vertex buffer slot dirty if the pipeline's stride or step mode for that vertex buffer slot differs from what had been previously established. The effect of marking the slot dirty is to ensure that a new `SetVertexBuffer` command is inserted before the next draw command that uses that vertex buffer. However, this is unnecessary: `wgpu_hal::CommandEncoder::set_vertex_buffer` does not need to be called simply because the stride or rate has changed.


cwfitzgerald (approved):
GLTM

Issue #1557, opened 2021-06-26
Bunnymark Screenshot is Empty Screen


Wumpf:
**Description**
Bunnymark shows an empty screen. The official screenshot is empty as well, indicating this happens on ci just the same (i.e. cementing whatever underlying bug there is into the new (?) image comparison tests)

**Repro steps**
cargo run --example bunnymark

**Expected vs observed behavior**
expected: something
observed: nothing

**Platform**
Windows, Vulkan

**TODO:**
- [ ] Update test framework to automatically tick to get things on the screen.
- [ ] Add readme with instructions to hit spacebar to see bunnies.

kvark:
It requires you to hit "space" to spawn bunnies.
@cwfitzgerald I think we should exclude it from snapshot testing

Wumpf:
I see :). Needs a readme then I guess.
Also it doesn't spawn bunny, but flipped wgpu logos. I feel betrayed!

cwfitzgerald:
@kvark i was planning on adding a way to have the snapshot automatically tick X times so that there is something visible, just didn't do it yet.

PR #2746, opened 2022-06-07
Fix bunnymark test screenshot and replace rand with nanorand

stevenhuyn:
**Connections**
Closes [Bunnymark Screenshot is Empty Screen #1557](https://github.com/gfx-rs/wgpu/issues/1557)

**Description**
Bunnymark screenshot right now is an empty screen because that example requires you to press spacebar. 

Also replacing rand dependency with smaller nanorand dependency

unsafe seems to be required because that seems to be how it's meant to be used to generate a DeviceID. https://docs.rs/winit/latest/winit/event/struct.DeviceId.html#impl

KeyboardInput: https://docs.rs/winit/latest/winit/event/enum.WindowEvent.html#variant.KeyboardInput

Adds a check if it is the bunnymark test and then gives the example a spacebar event and steps a couple of frames.

**Testing**
It is the change


wgpu/examples/framework.rs
```diff
@@ -488,6 +488,26 @@ pub fn test<E: Example>(mut params: FrameworkRefTest) {
 
             example.render(&dst_view, &ctx.device, &ctx.queue, &spawner);
 
+            // Handle specific case for bunnymark
+            if (params.image_path == "/examples/bunnymark/screenshot.png") {
```
cwfitzgerald, line 492: This feels weird, but I don't currently have a better idea, so thank you!

cwfitzgerald (approved):
Looks good to me, thank you for the PR and the thorough description!

wgpu/examples/framework.rs
stevenhuyn: Yep I tried doing a const from bunnymark.rs, but some scariness with the way imports/crates work won't allow it :/

stevenhuyn:
Failing because of:
```Rust
error: use of deprecated field `winit::event::KeyboardInput::modifiers`: Deprecated in favor of WindowEvent::ModifiersChanged
   --> wgpu\examples\conservative-raster\..\framework.rs:585:25
    |
585 |                         modifiers: winit::event::ModifiersState::empty(),
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `-D deprecated` implied by `-D warnings`
```

I'm not sure what you're meant to use instead of it, the type rqeuires ModifiersState, not ModifiersChanged.

cwfitzgerald:
I would just silence the warning for that struct init. We don't have a choice it seems.

stevenhuyn:
Also just checked, the color of the bunny changes per run, checking it out now.

cwfitzgerald (approved):
Good stuff

stevenhuyn:
Wait I think StdRng is chosen depending on the platform.

stevenhuyn:
I have followed their suggestion and am now using ChaChaRng for reproducibility.

cwfitzgerald (changes requested):
Looks good outside of the dependency, let's nix the dep on `rand` at all, and just use `nanorand`'s WyRand

stevenhuyn:
> Looks good outside of the dependency, let's nix the dep on `rand` at all, and just use `nanorand`'s WyRand

Hmm looks like a lot of the other packages use rng to an extent, I will expand the scope of the PR to just switch out this dependency?

wgpu/Cargo.toml
```diff
@@ -131,7 +130,7 @@ log = "0.4"
 noise = { version = "0.7", default-features = false }
 obj = "0.10"
 png = "0.17"
-rand = "0.7.2"
+nanorand = "0.7"
```
cwfitzgerald, line 133:
```suggestion
nanorand = { version = "0.7", default-features = false, features = ["wyrand"] }
```

Just because people get worried about dependency size.

cwfitzgerald (approved):
One touch, but looks good after that

wgpu/examples/framework.rs
```diff
@@ -488,6 +488,27 @@ pub fn test<E: Example>(mut params: FrameworkRefTest) {
...
+                // Press spacebar to spawn bunnies
+                example.update(winit::event::WindowEvent::KeyboardInput {
+                    input: winit::event::KeyboardInput {
+                        scancode: 0,
+                        state: winit::event::ElementState::Pressed,
+                        virtual_keycode: Some(winit::event::VirtualKeyCode::Space),
+                        modifiers: winit::event::ModifiersState::empty(),
+                    },
+                    device_id: unsafe { winit::event::DeviceId::dummy() },
+                    is_synthetic: false,
```
stevenhuyn, line 503: Maybe this?

```suggestion
                    is_synthetic: false,
```
```suggestion
                    is_synthetic: true,
```

stevenhuyn, line 503: Maybe this?

```suggestion
                    is_synthetic: false,
```
```suggestion
                    is_synthetic: true,
```

cwfitzgerald: Nah, this is never actually getting to winit, we're just unpacking the struct on the other end.

cwfitzgerald:
Hmm, I'll take a look on my windows machine in a bit.

stevenhuyn:
> Hmm, I'll take a look on my windows machine in a bit.

I'm actually using x86_64 Windows and it passes on my machine so no idea.

cwfitzgerald:
Try running with `$env:WGPU_ADAPTER_NAME='basic'` to run on the same software adapter that CI is running

cwfitzgerald:
lmao, mark the test as failing on that device `.specific_failure(None, None, Some("Basic"), false)` added to the test parameters and lets get this out the door.

stevenhuyn:
Okay now water panics:

```rust
running 1 test
test water ... FAILED

failures:

---- water stdout ----
117 outliers over max difference 166
thread 'water' panicked at 'UNEXPECTED TEST PASS: BACKEND | ADAPTER', wgpu\examples\water\..\..\tests\common\mod.rs:305:9
```

cwfitzgerald:
That's hilarious, you don't get a screenshot because it expects to fail but it passes (CI fails when tests unexpectedly pass so the tests stay updated).

Thanks for this PR and sorry for the hassle, I'll sort this PR out in the morning, need to double check some stuff.

stevenhuyn:
nws, thank you for your work!

cwfitzgerald:
@stevenhuyn what OS do you have? Windows 11?

stevenhuyn:
ye win 11

cwfitzgerald:
Yeah, that'd do it, so the water example is fixed on windows 11's warp, but not on windows 10's warp. We currently have no way of representing that in the code right now....

stevenhuyn:
Woo!

cwfitzgerald:
Filed https://github.com/gfx-rs/wgpu/issues/2760 as a follow up.

PR #2747, opened 2022-06-07
Introduce `VertexStep`: a stride and a step mode.

jimblandy:
This is used in various places around render pipelines, passes, and
bundles.

The public `wgpu_core::pipeline::VertexBufferLayout` could use
`VertexStep` as well, but I think it's best to let that continue to
resemble `GPUVertexBufferLayout`.


jimblandy:
premature

PR #2752, opened 2022-06-08
Expose naga span location helpers

nical:
**Description**

This exposes the `location` method from naga errors. The idea is to let the user obtain an `Option<naga::SourceLocation>` out of `CreateShaderModuleError` without matching on the internal error types.

This PR also bumps the naga dependency to a more recent revision since the `location` methods were added recently. Note that the glsl backend recently got bounds check for image loads in https://github.com/gfx-rs/naga/pull/1889, I left the extra parameter disabled in this PR to get the code to build.

**Testing**

None (the location method has a test in naga).

### DIFF
--- a/wgpu-hal/src/gles/device.rs
+++ b/wgpu-hal/src/gles/device.rs
@@ -144,6 +144,23 @@ impl super::Device {
             .position(|ep| ep.name.as_str() == stage.entry_point)
             .ok_or(crate::PipelineError::EntryPoint(naga_stage))?;
 
+        use naga::proc::BoundsCheckPolicy;
+        // The image bounds checks require the TEXTURE_LEVELS feature available in GL core 1.3+.
+        let version = gl.version();
+        let image_check = if !version.is_embedded && (version.major, version.minor) >= (1, 3) {
+            BoundsCheckPolicy::ReadZeroSkipWrite
+        } else {
+            BoundsCheckPolicy::Unchecked
+        };
+
+        // Other bounds check are either provided by glsl or not implemented yet.
+        let policies = naga::proc::BoundsCheckPolicies {
+            index: BoundsCheckPolicy::Unchecked,
+            buffer: BoundsCheckPolicy::Unchecked,
+            image: image_check,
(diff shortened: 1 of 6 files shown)


wgpu-hal/src/gles/device.rs
```diff
@@ -144,13 +144,23 @@ impl super::Device {
             .position(|ep| ep.name.as_str() == stage.entry_point)
             .ok_or(crate::PipelineError::EntryPoint(naga_stage))?;
 
+        // TODO: expose these parameters.
+        use naga::proc::BoundsCheckPolicy;
+        let policies = naga::proc::BoundsCheckPolicies {
+            index: BoundsCheckPolicy::Unchecked,
+            buffer: BoundsCheckPolicy::Unchecked,
+            image: BoundsCheckPolicy::Unchecked,
+            binding_array: BoundsCheckPolicy::Unchecked,
```
cwfitzgerald, line 153: These aren't the right settings for GLSL, unless only some of them work.

nical: I don't know what the right settings are, only that this parameter is new and that it was added in a recent commit implementing checks for image loads.

I'm happy to change it for whatever settings make more sense (you tell me!). There was no setting until recently so I assumed leaving them to unchecked would avoid changing the behavior. I'm only attempting to pull a recent naga revision but I suspect that someone will want to make a PR to wire these parameters up properly in wgpu.

cwfitzgerald: @JCapucho have any insight here?

jimblandy: In the wgpu chat, we decided that `wgpu_hal::gles::device::Device::create_shader` should check `gl.version()` to decide whether or not to request `ReadZeroSkipWrite` or `Unchecked`. See [Naga's GLSL back end](https://github.com/gfx-rs/naga/blob/53aa3e2df525dd772e6e9508e8e4a9758e9f5f98/src/back/glsl/features.rs#L114) for which versions have the needed functions.

cwfitzgerald (approved):
LGTM

PR #2754, opened 2022-06-09
Expect shader_primitive_index tests to fail on AMD RADV POLARIS12.

### LINKED ISSUE #2751: New `shader_primitive_index::draw` and `shader_primitive_index::draw_indexed` tests fail on Linux Vulkan back end
#2716 seems to break the following tests on Linux on a Vulkan back end:

    shader_primitive_index::draw
    shader_primitive_index::draw_indexed

```
---- shader_primitive_index::draw_indexed stdout ----
thread 'shader_primitive_index::draw_indexed' panicked at 'assertion failed: `(left == right)`
  left: `[255, 255, 255, 255, 255, 0, 0, 255, 255, 0, 0, 255, 255, 255, 255, 255]`,
 right: `[255, 255, 255, 255, 255, 0, 0, 255, 0, 0, 255, 255, 255, 255, 255, 255]`', wgpu/tests/shader_primitive_index/mod.rs:196:5
thread 'shader_primitive_index::draw_indexed' panicked at 'UNEXPECTED TEST FAILURE DUE TO PANIC', wgpu/tests/common/mod.rs:307:9

---- shader_primitive_index::draw stdout ----
thread 'shader_primitive_index::draw' panicked at 'assertion failed: `(left == right)`
  left: `[255, 255, 255, 255, 255, 0, 0, 255, 255, 0, 0, 255, 255, 255, 255, 255]`,
 right: `[255, 255, 255, 255, 0, 0, 255, 255, 255, 0, 0, 255, 255, 255, 255, 255]`', wgpu/tests/shader_primitive_index/mod.rs:196:5
thread 'shader_primitive_index::draw' panicked at 'UNEXPECTED TEST FAILURE DUE TO PANIC', wgpu/tests/common/mod.rs:307:9

```

jimblandy:
On Fedora, mesa-vulkan-drivers 22.0.3 AMD RADV POLARIS12 just colors both the NE and SW corners red. There's no blue. (NW and SE are white, as expected.) I suspect this means that the primitive ID is always being passed as zero.

This is probably not just due to a difference in rasterization: I tried enlarging the texture to 4x4 and shifting the triangles to ensure that there was a pixel in each quadrant that was fully covered.

With Vulkan "llvmpipe (LLVM 14.0.0, 256 bits)", the unmodified test passes.

This is probably limited to Linux, but `wgpu_tests::common::TestParameters` doesn't have a slot for that.

Fixes #2751.

### DIFF
--- a/wgpu/tests/shader_primitive_index/mod.rs
+++ b/wgpu/tests/shader_primitive_index/mod.rs
@@ -51,7 +51,14 @@ fn draw() {
     initialize_test(
         TestParameters::default()
             .test_features_limits()
-            .features(wgpu::Features::SHADER_PRIMITIVE_INDEX),
+            .features(wgpu::Features::SHADER_PRIMITIVE_INDEX)
+            // https://github.com/gfx-rs/wgpu/issues/2751
+            .specific_failure(
+                Some(wgt::Backends::VULKAN),
+                Some(0x1002), // AMD
+                Some("AMD RADV"),
+                false,
+            ),
         |ctx| {
             pulling_common(ctx, &expected, |rpass| {
                 rpass.draw(0..6, 0..1);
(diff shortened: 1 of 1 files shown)


wgpu/tests/shader_primitive_index/mod.rs
```diff
@@ -51,7 +51,14 @@ fn draw() {
     initialize_test(
         TestParameters::default()
             .test_features_limits()
-            .features(wgpu::Features::SHADER_PRIMITIVE_INDEX),
+            .features(wgpu::Features::SHADER_PRIMITIVE_INDEX)
+            // https://github.com/gfx-rs/wgpu/issues/2751
+            .specific_failure(
+                Some(wgt::Backends::VULKAN),
+                Some(0x1002), // AMD
+                Some("AMD RADV POLARIS12"),
```
cwfitzgerald, line 59: Let's just match against radv as a whole, it's probably a radv issue more than a Polaris specific issue

cwfitzgerald (changes requested):

cwfitzgerald (approved):

PR #2755, opened 2022-06-09
Move RenderBundleEncoder::finish's pipeline layout id into the state.

jimblandy:
All the rest of the tracking state is in the `State` struct, so it would be consistent for the pipeline layout id to be stored there as well.


cwfitzgerald (approved):
Scans

PR #2757, opened 2022-06-09
Don't panic on naga parsing error

### LINKED ISSUE #2545: ShaderSource::Glsl panics if shader has syntax error
**Description**
I'm trying to port an existing application using GLSL to `wgpu`, so I have existing complex shaders I want to modify to be compatible. While trying to get them working, I have found that if the shader has (something which `naga` considers) a syntax error, `wgpu` will panic via `.unwrap()`:

https://github.com/gfx-rs/wgpu/blob/326af60df8623e93b47a0de090e6cb449c8507f5/wgpu/src/backend/direct.rs#L1054

This has the following problems:

1. The application cannot recover from the error (e.g. in case it is allowing the shaders to be edited and reloaded) without using `catch_unwind()`.
2. The error message is presented in `Debug` formatting, which is not ideal for readability.
3. The error message provides only character (byte?), not line, positions (let alone line positions respecting GLSL `#line` directives), so it is difficult to correlate with a line in one's editor.

The last is presumably `naga`'s problem, but I propose that `wgpu` should report shader parsing errors in a `Result::Err()` instead of a panic.

(The `ShaderSource::SpirV` path also has an `.unwrap()` which could be improved the same way, but presumably that's less of a routine occurrence.)

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: [Error { kind: NotImplemented("variable qualifier"), meta: Span { start: 571, end: 580 } }]', /Users/kpreid/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.12.0/src/backend/direct.rs:1037:61
stack backtrace:
   0: rust_begin_unwind
             at /rustc/3a06854129d1ff7cb1e8a7fb62671782fe2513ba/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/3a06854129d1ff7cb1e8a7fb62671782fe2513ba/library/core/src/panicking.rs:116:14
             at /rustc/3a06854129d1ff7cb1e8a7fb62671782fe2513ba/library/core/src/result.rs:1690:5
             at /rustc/3a06854129d1ff7cb1e8a7fb62671782fe2513ba/library/core/src/result.rs:1018:23
   4: <wgpu::backend::direct::Context as wgpu::Context>::device_create_shader_module
             at /Users/kpreid/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.12.0/src/backend/direct.rs:1037:30
   5: wgpu::Device::create_shader_module
             at /Users/kpreid/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.12.0/src/lib.rs:1705:17
   6: all_is_cubes_gpu::in_wgpu::EverythingRenderer::new
             at ./all-is-cubes-gpu/src/in_wgpu.rs:144:29
```

**Platform**
macOS 12.3, `wgpu` 0.12

Uriopass:
**Connections**
Fixes #2545 

**Description**

Previously, the preprocess step of naga parsing to a module while
calling create_shader_module would panic and led to not so good error
messages. Now this is up to the user to make something out of it or
unwrap() knowingly.

**Testing**

Tested that an invalid glsl shader doesn't panic and simply returns an error


wgpu/src/lib.rs
```diff
@@ -2188,6 +2191,20 @@ impl Drop for Device {
     }
 }
 
+/// Error when parsing shader source that isn't wgsl, thus converting to a naga module beforehand.
+#[derive(Debug)]
+pub enum CreateShaderModuleError {
```
cwfitzgerald, line 2196: I just want to note that this is the first time naga becomes a public dependency.

wgpu/src/lib.rs
```diff
@@ -1954,16 +1954,19 @@ impl Device {
     }
 
     /// Creates a shader module from either SPIR-V or WGSL source code.
```
cwfitzgerald, line 1956: We should add a comment about what the result catches. We don't return an error on any validation problem, just parsing issues.

cwfitzgerald (changes requested):
I think this change is perfectly fine, but I want @kvark to look at this before we merge it

wgpu/src/lib.rs
jimblandy: Would it make sense to do a `Box<dyn Error + Send + Sync + 'static>` here, to avoid that?

Uriopass: I'm not sure how to impl Error on this error since glsl's parsers return multiple errors. For example what should source() return ?

wgpu/src/lib.rs
Uriopass: I have trouble understanding how the error scopes and everything works... Why does `handle_error` exists and pushes to a some sink instead of just returning the error directly in the function ?
I understand that we want to be able to disable validation errors, but we could just return Ok in that case ?

wgpu/src/lib.rs
kpreid: Perhaps `naga` could provide an error type which wraps multiple errors where they can occur, so that the `source()` of this error could be that error type. Then, if this enum (or its variants' fields) are hidden in a struct with private fields (or, I suppose, a `Box<dyn Error + ...`), the `naga` types become invisible except when the caller chooses to attempt downcasting to them.

kpreid: P.S. It would be good if, whatever the type structure turns out to be, the result of printing the error (chain) generically is a readable formatted report of the kind naga can produce. (I mentioned this already in #2545 but it seems worth reiterating as a design goal here.)

wgpu/examples/boids/main.rs
```diff
@@ -49,14 +49,18 @@ impl framework::Example for Example {
...
+                label: None,
+                source: wgpu::ShaderSource::Wgsl(Cow::Borrowed(include_str!("compute.wgsl"))),
+            })
+            .unwrap();
+        let draw_shader = device
+            .create_shader_module(&wgpu::ShaderModuleDescriptor {
+                label: None,
+                source: wgpu::ShaderSource::Wgsl(Cow::Borrowed(include_str!("draw.wgsl"))),
+            })
+            .unwrap();
```
kvark, line 63: I'm fine with generally exposing the fact GLSL and SPIRV inputs are translated on the client side, and therefore returning a Result() for that part of the shader translation.
However, I don't think it's sound to return a Result in case the user provides WGSL. This is confusing, since the signature of the function implies some error to be returned.
Therefore, perhaps, we should have a separate method? `create_shader_module_foreign` or something like this? That one could work with GLSL/SPIR-V inputs and return a result.

Uriopass: what would `device_create_shader_module` take in argument then ? What happens if it takes a ShaderModuleDescriptor that contains glsl data ?

wgpu/src/lib.rs
jimblandy: The root difficulty here is that Rust's `Error` trait simply is not designed to represent a list of errors like the kind a compiler front end would produce. This is why questions like "What should `source()` return?" have no good answer. Also, note that `naga::front::glsl::Parser::parse` returns `Result<Module, Vec<Error>>` - and `Vec<Error>` does not implement `Error`. So Naga's API is stretching Rust's `Result` paradigm a bit too far.

In the WebGPU API for content JavaScript, [`createShaderModule`](https://www.w3.org/TR/webgpu/#dom-gpudevice-createshadermodule) always succeeds, but the `GPUShaderModule` it returns may be marked "invalid". Valid or not, one can always call [`compilationInfo`](https://www.w3.org/TR/webgpu/#dom-gpushadermodule-compilationinfo) on a `GPUShaderModule` to retrieve a [`GPUCompilationInfo`](https://www.w3.org/TR/webgpu/#shader-module-compilation-information) structure, which has a list of error messages. Since the `GPUCompilationInfo` struct is available even when compilation succeeds, it can also carry warning messages.

Perhaps the thorough approach here would be to make `wgpu` match WebGPU more closely:
- add a `CompilationInfo` struct to `wgpu`, modeled after WebGPU's `GPUCompilationInfo`
- add an `Option<Box<CompilationInfo>>` to `wgpu::ShaderModule`
- add a method to `ShaderModule` to retrieve it
- add a method to `wgpu_core::hub::Global` akin to [`Global::create_buffer_error`](https://github.com/gfx-rs/wgpu/blob/f4c01052ef709e010f4e228d7d7d668f0166b3f5/wgpu-core/src/device/mod.rs#L3289-L3324) for allocating a dead-on-arrival shader module id, so that you can actually return a `ShaderModule` even when the Naga functions return an error.

-

jimblandy: I agree that Naga's error messages with spans and labels are really nice - nicer than WebGPU's `GPUCompilationMessage`, for sure. I guess `wgpu`'s `CompilationInfo` should follow Naga, and then it'd be Firefox's problem to convert that into what WebGPU requires.

Uriopass: I don't think this is a good first issue anymore :D I'll be closing the PR since it seems like there's too much at stake here and I'm not comfortable breaking new ground on wgpu. I think you summed it up well though.

PR #2759, opened 2022-06-10
Align the validation of Device::create_texture with the WebGPU spec

nical:
**Connections**

The [WebGPU specification for createTexture](https://www.w3.org/TR/webgpu/#texture-creation).

**Description**

This PR aligns the validation of create_texture with the steps described in the aforementioned specification. I tried to make sure the validation is at least as strict as the specification, but didn't attempt to remove any existing validation that isn't described in the specification.

There is a tiny bit of reordering so that the validation in the code goes in roughly the same order as the specification (for ease of comparison).

I also found in the process that the method is probably missing a view_formats parameter but if so, it will come in a followup. 

**Testing**

None.

### DIFF
--- a/wgpu-core/src/conv.rs
+++ b/wgpu-core/src/conv.rs
@@ -116,14 +116,7 @@ pub fn check_texture_dimension_size(
     use wgt::TextureDimension::*;
 
     let (extent_limits, sample_limit) = match dimension {
-        D1 => (
-            [
-                limits.max_texture_dimension_1d,
-                1,
-                limits.max_texture_array_layers,
-            ],
-            1,
-        ),
+        D1 => ([limits.max_texture_dimension_1d, 1, 1], 1),
         D2 => (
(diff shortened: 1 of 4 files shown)


nical:
Note I didn't quite get the difference between:
```
descriptor.[usage](https://www.w3.org/TR/webgpu/#dom-gputexturedescriptor-usage) must contain only bits present in this’s [allowed texture usages](https://www.w3.org/TR/webgpu/#allowed-texture-usages).
```
and
```
descriptor.[usage](https://www.w3.org/TR/webgpu/#dom-gputexturedescriptor-usage) must be a combination of [GPUTextureUsage](https://www.w3.org/TR/webgpu/#namespacedef-gputextureusage) values.
```

The latter I think is enforced by the type itself, but I'm not sure whether the former is provisioning for a future with more usages, some of which are not supported by all devices.

PR #2763, opened 2022-06-11
allow creating wgpu::Instance from wgpu_core::Instance

i509VCB:
**Description**
When creating custom hal instances, you may want to do so for multiple backends. Since a wgpu_core::Instance may have multiple instance members constructed, that is probably the best spot to deal with multiple custom wgpu_hal backends.

I do this specifically in my wgpu fork to integrate with some Linux specific APIs and want to be able to use EGL and Vulkan adapters.

### DIFF
--- a/wgpu/src/backend/direct.rs
+++ b/wgpu/src/backend/direct.rs
@@ -47,12 +47,20 @@ impl Context {
         ))
     }
 
-    #[cfg(any(not(target_arch = "wasm32"), feature = "webgl2"))]
+    #[cfg(any(not(target_arch = "wasm32"), feature = "webgl"))]
     pub unsafe fn instance_as_hal<A: wgc::hub::HalApi, F: FnOnce(Option<&A::Instance>) -> R, R>(
         &self,
         hal_instance_callback: F,
     ) -> R {
         self.0.instance_as_hal::<A, F, R>(hal_instance_callback)
+    }
+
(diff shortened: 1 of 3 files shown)


wgpu/src/lib.rs
```diff
@@ -1647,6 +1647,22 @@ impl Instance {
...
+    /// Create an new instance of wgpu from a wgpu-core instance.
+    ///
+    /// # Arguments
+    ///
+    /// - `core_instance` - wgpu-core instance.
+    ///
+    /// # Safety
+    ///
+    /// Refer to the creation of wgpu-core Instance.
+    #[cfg(any(not(target_arch = "wasm32"), feature = "webgl2"))]
```
jimblandy, line 1659: This ought to match the condition used to select `backend/direct.rs`, shouldn't it? That's `"webgl"`,not `"webgl2"`.

wgpu/src/backend/direct.rs
```diff
@@ -55,6 +55,14 @@ impl Context {
         self.0.instance_as_hal::<A, F, R>(hal_instance_callback)
     }
 
+    #[cfg(any(not(target_arch = "wasm32"), feature = "webgl2"))]
```
jimblandy, line 58: `wgpu` doesn't have a `"webgl2"` feature, does it? Should this be `"webgl"`?

jimblandy (changes requested):

jimblandy (approved):

PR #2766, opened 2022-06-12
Clarify docs for `wgpu_core`'s `Id` and `gfx_select!`.

jimblandy:
Clarify that `gfx_select!` is not specific to `hub::Global`.

Clarify that the `T` in `id<T>` is not a real resource type.


cwfitzgerald (approved):
Very good stuff

PR #2770, opened 2022-06-14
Update the Supported Platforms table in README

jinleili:
**Description**
vk backend was already supported on macOS and iOS via `vulkan-portability` feature, gl backend was already supported on macOS via `angle` feature, fixed `WGPU_POWER_PREF` environment variable name.


cwfitzgerald (approved):
Party

PR #2773, opened 2022-06-14
Add type annotation to render pass code, for rust-analyzer.

jimblandy:
This is a workaround for a limitation in rust-analyzer:
https://github.com/rust-lang/rust-analyzer/issues/12247


jimblandy:
Without this, rust-analyzer doesn't know the types of important variables, and things like find-references don't work, which is extremely confusing.

cwfitzgerald (approved):

Issue #2526, opened 2022-03-07
Misleading create_render_pipeline error message when hitting bind group limit


Oberdiah:
**Description**
When calling create_render_pipeline with a shader that goes over the allowed maximum bind groups as set by `DeviceDescriptor::limits::max_bind_groups` (by default 4), the returned error message is quite misleading:

```
error matching FRAGMENT shader requirements against the pipeline
shader global ResourceBinding { group: 4, binding: 0 } is not available in the layout pipeline layout
binding is missing from the pipeline layout
```

The error should mention hitting the limit and possibly suggest raising it (if the new value would be less than the adapter limit)

PR #2775, opened 2022-06-15
Add InvalidGroupIndex validation at create_shader_module

jinleili:
**Connections**
https://github.com/gfx-rs/wgpu/issues/2526

**Testing**
Tested locally, now, if an invalid group index is used, the following validation message will be output:
```sh
Caused by:
    shader global ResourceBinding { group: 4, binding: 0 } uses a group index 4 that exceeds the max_bind_groups limit of 4.
```

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -1189,6 +1189,18 @@ impl<A: HalApi> Device<A> {
             }
             pipeline::ShaderModuleSource::Naga(module) => (module, String::new()),
         };
+        for (_, var) in module.global_variables.iter() {
+            match var.binding {
+                Some(ref br) if br.group >= self.limits.max_bind_groups => {
+                    return Err(pipeline::CreateShaderModuleError::InvalidGroupIndex {
+                        bind: br.clone(),
+                        group: br.group,
+                        limit: self.limits.max_bind_groups,
+                    });
+                }
+                _ => continue,
+            };
+        }
 
         use naga::valid::Capabilities as Caps;
         profiling::scope!("naga::validate");
--- a/wgpu-core/src/pipeline.rs
+++ b/wgpu-core/src/pipeline.rs
@@ -124,6 +124,14 @@ pub enum CreateShaderModuleError {
     Validation(#[from] ShaderError<naga::WithSpan<naga::valid::ValidationError>>),
     #[error(transparent)]
     MissingFeatures(#[from] MissingFeatures),
+    #[error(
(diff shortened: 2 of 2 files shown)


jimblandy (changes requested):
I think this error should be detected in `wgpu_core::command::render::State::is_ready`, not in `create_shader_module`. It looks like render bundle encoding doesn't detect this problem either.

The WebGPU spec says that these errors should be reported when a `draw` command is encoded, by the ["Validate encoder bind groups"](https://www.w3.org/TR/webgpu/#abstract-opdef-validate-encoder-bind-groups) algorithm. Since [`setBindGroup`](https://www.w3.org/TR/webgpu/#dom-gpubindingcommandsmixin-setbindgroup) refuses to set bind groups at indices greater than `this.[[device]].[[limits]].maxBindGroups`, the "Validate encoder bind groups" algorithm will fail if the module uses bind group numbers that are out of range.

It's important that `wgpu` follow the WebGPU specification carefully regarding when errors are reported. The WebGPU Conformance Test Suite will treat variant behaviors as failures - appropriately, because they mean that WebGPU behavior won't be portable across different environments.

jinleili:
If detect in `is_ready`，will need to repeat the detection code in `render::State::is_ready` and `compute::State::is_ready`.
Is it acceptable to move detection to `wgpu_core::validation::Interface::check_stage` ?

In addition, looking at google `dawn` is also done in the initialization phase of ShaderModule:
https://dawn.googlesource.com/dawn/+/refs/heads/main/src/dawn/native/ShaderModule.cpp#871

jimblandy:
I'm sorry - I've looked at the WebGPU spec more, and it looks to me like this is an unfinished area of the spec, so it's not important for us to comply exactly at this point. In particular, in the [default pipeline layout](https://www.w3.org/TR/webgpu/#abstract-opdef-default-pipeline-layout) algorithm, I noticed:

- "ISSUE 35" says that the details of how a shader module's bindings are exposed are not yet settled.
- Step 3.12.1 says, "Append entry to `groupDescs[group]`." There's no validation of `group` specified before this point, so this would be an out-of-bounds access if the resource's group attribute were out of range.

So my inference that the error was meant to be reported at draw call time isn't supported - it needs to be validated before that, at least at pipeline creation time.

Furthermore, "ISSUE 31" says that the rules for module validation are unsettled. So clearly this area is still being developed.

Doing the check at shader module creation time seems friendliest, so let's just go with that.

jimblandy (approved):
This seems like the right approach, until the spec gives us more specific guidance.

PR #2777, opened 2022-06-15
Implement Queue::write_buffer_with

### LINKED ISSUE #1670: Queue::write_buffer_with

Currently I'm using the Queue to write an array of Std140-convertible data to a Buffer (using Queue::write_buffer).

This takes the following form:

1. Allocate a Vec of Std140 convertible data
2. Allocate a `Vec<u8>` as "scratch space" to write the final Std140-formatted bytes to
3. Pass the `Vec<u8>` as a reference to Queue::write_buffer (which accepts `&[u8]`)

Under the hood, I know that Queue::write_buffer is writing to a memory-mapped staging buffer internally. It occurred to me that the intermediate Vec<u8> allocation shouldn't be necessary. Instead I could just write the "converted std140 bytes" directly to the memory-mapped staging buffer.

It would be nice to have something like:

```rust
fn write_buffer_with(&self, buffer: &Buffer, size: BufferAddress, func: impl FnMut(&mut [u8])) {};
```

This would reserve the requested amount of space in the staging buffer and expose a slice that I could then pass directly into the `std140::Writer`.

Use a staging buffer manually (this is what i used to do). This requires more machinery because we need manage the staging buffer, call copy_buffer_to_buffer on a CommandEncoder, and then manually submit the encoder. I personally like the ergonomics of Queue::write_buffer.

teoxoy:
**Connections**
closes #1670

**Description**
The new `Queue::write_buffer_with` method allows users to write their data directly in a staging buffer; avoiding an extra copy in more complex projects where you'd have to first write the content to a temporary `Vec` and then pass it to `Queue::write_buffer`.

**Testing**
I converted a few examples locally to test the new method.
Let me know if I should push some of them.

### DIFF
--- a/wgpu-core/src/device/queue.rs
+++ b/wgpu-core/src/device/queue.rs
@@ -8,10 +8,10 @@ use crate::{
     conv,
     device::{DeviceError, WaitIdleError},
     get_lowest_common_denom,
-    hub::{Global, GlobalIdentityHandlerFactory, HalApi, Token},
+    hub::{Global, GlobalIdentityHandlerFactory, HalApi, Input, Token},
     id,
     init_tracker::{has_copy_partial_init_tracker_coverage, TextureInitRange},
-    resource::{BufferAccessError, BufferMapState, TextureInner},
+    resource::{BufferAccessError, BufferMapState, StagingBuffer, TextureInner},
     track, FastHashSet, SubmissionIndex,
 };
 
@@ -86,28 +86,6 @@ pub struct WrappedSubmissionIndex {
     pub index: SubmissionIndex,
 }
 
-struct StagingData<A: hal::Api> {
-    buffer: A::Buffer,
-}
-
-impl<A: hal::Api> StagingData<A> {
-    unsafe fn write(
-        &self,
-        device: &A::Device,
-        offset: wgt::BufferAddress,
-        data: &[u8],
-    ) -> Result<(), hal::DeviceError> {
-        let mapping = device.map_buffer(&self.buffer, offset..offset + data.len() as u64)?;
-        ptr::copy_nonoverlapping(data.as_ptr(), mapping.ptr.as_ptr(), data.len());
-        if !mapping.is_coherent {
-            device
-                .flush_mapped_ranges(&self.buffer, iter::once(offset..offset + data.len() as u64));
-        }
-        device.unmap_buffer(&self.buffer)?;
-        Ok(())
-    }
-}
-
 #[derive(Debug)]
 pub enum TempResource<A: hal::Api> {
     Buffer(A::Buffer),
@@ -178,8 +156,8 @@ impl<A: hal::Api> PendingWrites<A> {
         self.temp_resources.push(resource);
     }
 
-    fn consume(&mut self, stage: StagingData<A>) {
-        self.temp_resources.push(TempResource::Buffer(stage.buffer));
+    fn consume(&mut self, buffer: StagingBuffer<A>) {
+        self.temp_resources.push(TempResource::Buffer(buffer.raw));
     }
 
     #[must_use]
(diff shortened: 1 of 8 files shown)


wgpu/src/lib.rs
```diff
@@ -3355,6 +3367,40 @@ impl<'a> RenderBundleEncoder<'a> {
...
+    buffer: &'a Buffer,
+    offset: BufferAddress,
+    inner: QueueWriteBuffer,
+}
+
+impl<'a> std::ops::Deref for QueueWriteBufferView<'a> {
+    type Target = [u8];
+
+    fn deref(&self) -> &Self::Target {
+        panic!("QueueWriteBufferView is write-only!");
```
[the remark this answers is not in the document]
teoxoy: I don't think you can. See https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8507973a9c6ed94ee89eb6670980314f

As far as I understand, rust reborrows through the mutable reference which in turn panics in this case.

`BufferViewMut` also has the same implementation https://github.com/gfx-rs/wgpu/blob/0d4d3f4ceb0c07de38064533c3409a2f8c1d8d95/wgpu/src/lib.rs#L2297-L2308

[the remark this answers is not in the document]
a1phyr: You can still do the following casts: `&mut QueueWriteBufferView` -> `&mut [u8]` -> `&[u8]. With your example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d473956694eea896d1e8de9aa5ab9273.

Anyway, I guess it is fine to do this as it may prevent errors, as long as the buffer is properly initialized to prevent UB if someone reads from it anyway ; especially if the same pattern is used elsewhere in `wgpu`.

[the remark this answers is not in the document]
cwfitzgerald: That's only there to try and discourage reading - it will work and it's not UB - but reading from a write combined buffer is just comically slow.

wgpu/src/lib.rs
```diff
@@ -3367,6 +3413,33 @@ impl Queue {
         Context::queue_write_buffer(&*self.context, &self.id, &buffer.id, offset, data)
     }
 
+    /// Schedule a data write into `buffer` starting at `offset` via the returned [QueueWriteBufferView].
+    ///
+    /// The returned value can be dereferenced to a `&mut [u8]`.
+    ///
+    /// Dropping the returned value fails if `size` is greater than the size of `buffer` starting at `offset`.
```
cwfitzgerald, line 3420: Can this check be done immediately so we don't get a validation error inside drop?

wgpu-core/src/device/queue.rs
```diff
@@ -305,30 +305,141 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+
+            let dst = buffer_guard.get_mut(buffer_id).unwrap();
+            dst.initialization_status
+                .drain(buffer_offset..(buffer_offset + data_size));
+        }
+
+        Ok(())
+    }
+
+    pub fn queue_create_staging_buffer<A: HalApi>(
```
cwfitzgerald, line 397: My only real concern is if we can get these two methods of writing buffers to share mode code.

cwfitzgerald (commented):
This is damn good, I really don't have many concerns about it at all besides the small things I mentioned

wgpu/src/lib.rs
teoxoy: I added a new fn `queue_validate_write_buffer` that gets called by `write_buffer_with`.

I only implemented it for the direct backend since the web one requires [this change](https://github.com/rustwasm/wasm-bindgen/commit/b63dd8b0d3688b8464ccafb662c4f2ee39cedfc5#diff-99b4e1aaad73aae9f74eef93ac039f36a85e634b88efed7f83258fcbdaf42ee2) to implement.

Not sure how this impacts the perf (and if the impact is worth it) since we are now locking the device and buffers for a short period of time just to get the buffer size and its usages. What are your thoughts on this?

cwfitzgerald: This will go away with the new storages, and shouldn't be that slow in the mean time, so should be fine

wgpu-core/src/device/queue.rs
```diff
@@ -312,81 +314,197 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
-        let data_size = data.len() as wgt::BufferAddress;
         if data_size == 0 {
             log::trace!("Ignoring write_buffer of size 0");
             return Ok(());
         }
 
-        let stage = device.prepare_stage(data_size)?;
+        let (staging_buffer, staging_buffer_ptr) = device.prepare_staging_buffer(data_size)?;
+
+        profiling::scope!("copy");
```
cwfitzgerald, line 329: This scope will end too late, it should be inside the unsafe block

wgpu/src/lib.rs
```diff
@@ -3367,6 +3420,32 @@ impl Queue {
...
+    ///
+    /// This method fails if `size` is greater than the size of `buffer` starting at `offset`.
+    #[must_use]
+    pub fn write_buffer_with<'a>(
+        &'a self,
+        buffer: &'a Buffer,
+        offset: BufferAddress,
+        size: BufferSize,
+    ) -> QueueWriteBufferView<'a> {
+        Context::queue_validate_write_buffer(&*self.context, &self.id, &buffer.id, offset, size);
```
cwfitzgerald, line 3439: Why is this a separate call through the context and not just part of queue_create_staging_buffer?

teoxoy: I thought it would be appropriate to call it here in `write_buffer_with` since `queue_validate_write_buffer` and `queue_create_staging_buffer` don't share any parameters/functionality.

cwfitzgerald: I'd prefer this to be a single call so we don't need to pay the cost of grabbing locks etc again

teoxoy: The only lock we could reuse is the device one (one being read; for `queue_validate_write_buffer`, the other write; for `queue_create_staging_buffer`) but you mentioned previously that we'll be able to avoid the locks for the validation part (where we only need to read the size and usage of the dst buffer).

cwfitzgerald: Alright, this is good enough for now, I need to revisit write_buffer at some point in the future anyway, so I'll look at it more in depth then.

cwfitzgerald (approved):
Thanks for taking this on!

Issue #2469, opened 2022-02-09
Relax color attachment validation


kvark:
See https://github.com/gpuweb/gpuweb/pull/2562, which covers 2 of the issues linked from there:

1. We need to make the color attachments `Option`-al in render pipelines, render passes, and render bundles
2. We need to allow shaders to have more outputs than the pipeline has. (maybe we already allow that?)

teoxoy:
This has been done in https://github.com/gfx-rs/wgpu/pull/2778 and https://github.com/gfx-rs/wgpu/pull/2822.

PR #2778, opened 2022-06-16
Relax render pass color_attachments validation

jinleili:
**Connections**
https://github.com/gfx-rs/wgpu/issues/2469

**Description**

> 1.  We need to make the color attachments `Option`-al in render pipelines, render passes, and render bundles

I tried `Option`-al all of these, but that makes the code changes look not so good.
Since `&[]` can express `None`, is it still necessary to use `Option`?

**Testing**
Tested locally

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -20,7 +20,7 @@ use hal::{CommandEncoder as _, Device as _};
 use parking_lot::{Mutex, MutexGuard};
 use smallvec::SmallVec;
 use thiserror::Error;
-use wgt::{BufferAddress, TextureFormat, TextureViewDimension};
+use wgt::{BufferAddress, ColorTargetState, TextureFormat, TextureViewDimension};
 
 use std::{borrow::Cow, iter, mem, num::NonZeroU32, ops::Range, ptr};
 
@@ -52,7 +52,7 @@ pub enum HostMap {
 #[derive(Clone, Debug, Hash, PartialEq)]
 #[cfg_attr(feature = "serial-pass", derive(serde::Deserialize, serde::Serialize))]
 pub(crate) struct AttachmentData<T> {
-    pub colors: ArrayVec<T, { hal::MAX_COLOR_ATTACHMENTS }>,
+    pub colors: ArrayVec<Option<T>, { hal::MAX_COLOR_ATTACHMENTS }>,
     pub resolves: ArrayVec<T, { hal::MAX_COLOR_ATTACHMENTS }>,
     pub depth_stencil: Option<T>,
 }
@@ -60,7 +60,7 @@ impl<T: PartialEq> Eq for AttachmentData<T> {}
 impl<T> AttachmentData<T> {
     pub(crate) fn map<U, F: Fn(&T) -> U>(&self, fun: F) -> AttachmentData<U> {
         AttachmentData {
-            colors: self.colors.iter().map(&fun).collect(),
+            colors: self.colors.iter().map(|c| c.as_ref().map(&fun)).collect(),
             resolves: self.resolves.iter().map(&fun).collect(),
             depth_stencil: self.depth_stencil.as_ref().map(&fun),
         }
@@ -78,8 +78,8 @@ pub(crate) struct RenderPassContext {
 pub enum RenderPassCompatibilityError {
     #[error("Incompatible color attachment: the renderpass expected {0:?} but was given {1:?}")]
     IncompatibleColorAttachment(
-        ArrayVec<TextureFormat, { hal::MAX_COLOR_ATTACHMENTS }>,
-        ArrayVec<TextureFormat, { hal::MAX_COLOR_ATTACHMENTS }>,
+        ArrayVec<Option<TextureFormat>, { hal::MAX_COLOR_ATTACHMENTS }>,
+        ArrayVec<Option<TextureFormat>, { hal::MAX_COLOR_ATTACHMENTS }>,
     ),
     #[error(
         "Incompatible depth-stencil attachment: the renderpass expected {0:?} but was given {1:?}"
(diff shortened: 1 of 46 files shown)


jimblandy (changes requested):
The rename looks good - could we split that out into its own PR?

However, the WebIDL and the spec language make it clear that `colorAttachments` can be an array with holes, like `[x, null, y]`, so it really does need to be an `Cow<'a, [Option<RenderPassColorAttachment>]>` to match the spec.

jinleili:
Tested color_attachments: `[Some]`, `[Some, None]`, `[None, Some]`, `[Some, None, Some]` on metal, dx12, vulkan backends,  wasm target and gl backend only support non-hole attachments currently. Since a new version of the aligned WebGPU spec `web-sys` was recently released, sparse attachments support can be put into the update of `web-sys`.

wgpu-core/src/device/mod.rs
```diff
@@ -2754,29 +2759,24 @@ impl<A: HalApi> Device<A> {
...
+                    validation::check_texture_format(format, &output.ty).map_err(|pipeline| {
+                        pipeline::CreateRenderPipelineError::ColorState(
+                            *i as u8,
+                            pipeline::ColorStateError::IncompatibleFormat {
+                                pipeline,
+                                shader: output.ty,
                             },
-                        )?;
-                    }
-                    None if state.write_mask.is_empty() => {}
```
jimblandy, line 2772: I don't quite understand what this code is meant to do, but, this test seems to have disappeared. Why is that?

wgpu-core/src/device/mod.rs
```diff
@@ -2754,29 +2759,24 @@ impl<A: HalApi> Device<A> {
...
-                    None => {
-                        log::warn!("Missing fragment output[{}], expected {:?}", i, state,);
-                        return Err(pipeline::CreateRenderPipelineError::ColorState(
-                            i as u8,
-                            pipeline::ColorStateError::Missing,
```
jimblandy, line 2777: Nothing raises this error any more. Shouldn't it be in an `else` clause of the new `if let Some(state) = color_targets.get(...)` conditional?

wgpu-core/src/device/mod.rs
```diff
@@ -2754,29 +2759,24 @@ impl<A: HalApi> Device<A> {
...
-                        ));
-                    }
+                        )
+                    })?;
```
jimblandy, line 2779: Suggestion: it might be a bit nicer to just flatten this into a `match` statement:

```
                match color_targets.get(*i as usize) {
                    Some(&Some(ref state)) => {
                        let format = state.format;
                        validation::check_texture_format(format, &output.ty).map_err(|pipeline| {
                            pipeline::CreateRenderPipelineError::ColorState(
                                *i as u8,
                                pipeline::ColorStateError::IncompatibleFormat {
                                    pipeline,
                                    shader: output.ty,
                                },
                            )
                        })?;
                    }
                    Some(&None) => {
                        return Err(
                            pipeline::CreateRenderPipelineError::InvalidFragmentOutputLocation(*i),
                        );
                    }
                    None => {
                        /* possibly missing error check here */
                    },
                }
```

jimblandy (changes requested):
Thank you so much for undertaking this! That turned out to be a massive change.

wgpu-core/src/device/mod.rs
jinleili: Since I changed the order of the loop and used shader output `@location` index to fetch the color_target, ` None if state.write_mask.is_empty()` is no longer needed.

wgpu-core/src/device/mod.rs
jinleili: My mistake

wgpu-hal/src/vulkan/command.rs
```diff
@@ -362,31 +362,37 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
...
+                        target.make_attachment_key(crate::AttachmentOps::STORE, caps)
+                    }),
+                };
+
+                rp_key.colors.push(Some(color));
+                fb_key.attachments.push(cat.target.view.attachment.clone());
+                if let Some(ref at) = cat.resolve_target {
+                    vk_clear_values.push(mem::zeroed());
+                    vk_image_views.push(at.view.raw);
+                    println!("fb_key.attachments: {:?}", &at.view.attachment.raw);
```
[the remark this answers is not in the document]
jinleili: Shame on myself for committing so much temporary code :(

wgpu-hal/src/dx12/mod.rs
```diff
@@ -221,7 +221,7 @@ pub struct Device {
     private_caps: PrivateCapabilities,
     shared: Arc<DeviceShared>,
     // CPU only pools
-    rtv_pool: Mutex<descriptor::CpuPool>,
+    rtv_pool: Arc<Mutex<descriptor::CpuPool>>,
```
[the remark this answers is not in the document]
jinleili: I don't know if `Arc::try_unwrap(self.shared).ok().map(|shared| shared.rtv_pool.into_inner().destroy())` is the right way to destroy rtv_pool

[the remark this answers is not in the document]
jimblandy: I brought this up [in Matrix](https://matrix.to/#/!FZyQrssSlHEZqrYcOb:matrix.org/$CP4QMBU4CGgAbv6bigj6XBohhf1Uv9N_3pm6JuxhENU?via=matrix.org&via=mozilla.org&via=kde.org) and the suggestion there was that we create the null descriptor when we create the Device, let it get destroyed when the Device is destroyed, and then just give every CommandEncoder a copy of the handle. That way, there's no need to allocate the null descriptor in `begin_render_pass`, so there's no need for `rtv_pool` to be shared.

In chat, @magcius (jasperrlz) said:
> RTV/DSV descriptors are weird, because they don't have to persist in a pool. As long as the descriptors are alive at the time of bind, it's fine.

wgpu-hal/src/dx12/mod.rs
```diff
@@ -211,6 +212,7 @@ impl DeviceShared {
         self.cmd_signatures.destroy();
         self.heap_views.raw.destroy();
         self.heap_samplers.raw.destroy();
+        // self.rtv_pool.into_inner().destroy();
```
jimblandy, line 215: This probably should come out (but see the overall review comments, maybe it'll go away entirely).

jimblandy (changes requested):
This looks great - thank you!

The only thing that still concerns me is the handling of `rtv_pool`. Based on discussion in Matrix, I think we can avoid having to share `rtv_pool` altogether, by allocating the null handle we need earlier, and just passing it in to command buffer creation. See [this comment](https://github.com/gfx-rs/wgpu/pull/2778#discussion_r907627272).

jimblandy (approved):
superb

PR #2783, opened 2022-06-17
Change get_preferred_format to get_supported_formats

### LINKED ISSUE #2718: Change `get_preferred_format` to `get_supported_formats`

There's currently no way to actually get the full list of supported surface textures.

Turn 

```rust
Surface::get_preferred_format(&self, adapter: &Adapter) -> TextureFormat
```

into

```rust
Surface::get_supported_formats(&self, adapter: &Adapter) -> Vec<TextureFormat>
```

which will always return the "preferred" format first.

Having both calls be independent.

stevenhuyn:
**Connections**
Closes #2718

**Description**
Converted everything related to `get_preferred_format -> Option<TextureFormat>` to `get_supported_formats -> Option<Vec<TextureFormat>>`

Converted the web version to return the set of texture formats defined by https://gpuweb.github.io/gpuweb/#supported-context-formats

**Testing**
I ran `cargo nextest run --no-fail-fast`

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -3013,12 +3013,12 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             .map_err(|_| instance::IsSurfaceSupportedError::InvalidSurface)?;
         Ok(adapter.is_surface_supported(surface))
     }
-    pub fn surface_get_preferred_format<A: HalApi>(
+    pub fn surface_get_supported_formats<A: HalApi>(
         &self,
         surface_id: id::SurfaceId,
         adapter_id: id::AdapterId,
-    ) -> Result<TextureFormat, instance::GetSurfacePreferredFormatError> {
-        profiling::scope!("surface_get_preferred_format");
+    ) -> Result<Vec<TextureFormat>, instance::GetSurfacePreferredFormatError> {
+        profiling::scope!("surface_get_supported_formats");
         let hub = A::hub(self);
         let mut token = Token::root();
 
@@ -3031,7 +3031,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             .get(surface_id)
             .map_err(|_| instance::GetSurfacePreferredFormatError::InvalidSurface)?;
 
-        surface.get_preferred_format(adapter)
+        surface.get_supported_formats(adapter)
     }
 
     pub fn device_features<A: HalApi>(
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -3018,7 +3018,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         surface_id: id::SurfaceId,
         adapter_id: id::AdapterId,
     ) -> Result<Vec<TextureFormat>, instance::GetSurfacePreferredFormatError> {
-        profiling::scope!("surface_get_supported_formats");
+        profiling::scope!("Surface::get_supported_formats");
         let hub = A::hub(self);
         let mut token = Token::root();
 
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -156,17 +156,6 @@ impl Surface {
         &self,
         adapter: &Adapter<A>,
     ) -> Result<Vec<wgt::TextureFormat>, GetSurfacePreferredFormatError> {
-        // Check the four formats mentioned in the WebGPU spec.
-        // Also, prefer sRGB over linear as it is better in
-        // representing perceived colors.
-        let candidate_formats = [
-            wgt::TextureFormat::Bgra8UnormSrgb,
-            wgt::TextureFormat::Rgba8UnormSrgb,
-            wgt::TextureFormat::Bgra8Unorm,
-            wgt::TextureFormat::Rgba8Unorm,
-            wgt::TextureFormat::Rgba16Float,
-        ];
-
         let suf = A::get_surface(self);
         let caps = unsafe {
             profiling::scope!("surface_capabilities");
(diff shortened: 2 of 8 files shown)


stevenhuyn:
Trying to understand this error, when `get_preferred_format` doesn't exist in either this branch nor the `web_sys` crate.

```rust
error[E0599]: no method named `get_supported_formats` found for struct `GpuCanvasContext` in the current scope
    --> wgpu\src\backend\web.rs:1198:14
     |
1198 |             .get_supported_formats(&adapter.0)
     |              ^^^^^^^^^^^^^^^^^^^^^ help: there is an associated function with a similar name: `get_preferred_format`

For more information about this error, try `rustc --explain E0599`.
```

cwfitzgerald:
All that stuff is behind the web_sys_unstable_apis cfg, so doesn't show up on docs.rs.

wgpu/src/backend/web.rs
```diff
@@ -553,7 +553,7 @@ fn map_texture_format(texture_format: wgt::TextureFormat) -> web_sys::GpuTexture
     }
 }
 
-fn map_texture_format_from_web_sys(
+fn _map_texture_format_from_web_sys(
```
stevenhuyn, line 556: Should this function be removed now that it's unused?

wgpu/examples/framework.rs
```diff
@@ -198,7 +198,11 @@ fn start<E: Example>(
     let spawner = Spawner::new();
     let mut config = wgpu::SurfaceConfiguration {
         usage: wgpu::TextureUsages::RENDER_ATTACHMENT,
-        format: surface.get_preferred_format(&adapter).unwrap(),
+        format: surface
```
stevenhuyn, line 201: pop() to own the last element. Is there any implications of the change from previous behavior where the first compatible texture format was chosen?

wgpu-core/src/instance.rs
```diff
@@ -177,10 +177,16 @@ impl Surface {
                 .ok_or(GetSurfacePreferredFormatError::UnsupportedQueueFamily)?
         };
 
-        preferred_formats
+        let supported_formats = candidate_formats
             .iter()
             .cloned()
-            .find(|preferred| caps.formats.contains(preferred))
+            .filter(|candidate| caps.formats.contains(candidate))
```
cwfitzgerald, line 183: Instead of filtering the candidates we should just read the capabilities and return the formats the device says its capable of

wgpu/examples/framework.rs
cwfitzgerald: We should be able to do .first().unwrap()

wgpu/src/backend/web.rs
cwfitzgerald: Yes

wgpu/src/lib.rs
```diff
@@ -3438,8 +3438,8 @@ impl Surface {
     /// Returns an optimal texture format to use for the [`Surface`] with this adapter.
     ///
     /// Returns None if the surface is incompatible with the adapter.
```
cwfitzgerald, line 3440: Should make a note that the first format is the preferred.

cwfitzgerald (changes requested):
Looks good, thanks for tackling this, some comments

wgpu/examples/framework.rs
stevenhuyn: It returns a reference and not the owned value.

stevenhuyn: Wait the compiler was angry at me before about it

stevenhuyn: Just clone the enum?

image

cwfitzgerald: TextureFormat should be Copy (if it isn't, make it so) so you can just dereference it.

wgpu-core/src/device/mod.rs
```diff
@@ -3013,12 +3013,12 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
     }
-    pub fn surface_get_preferred_format<A: HalApi>(
+    pub fn surface_get_supported_formats<A: HalApi>(
         &self,
         surface_id: id::SurfaceId,
         adapter_id: id::AdapterId,
-    ) -> Result<TextureFormat, instance::GetSurfacePreferredFormatError> {
-        profiling::scope!("surface_get_preferred_format");
+    ) -> Result<Vec<TextureFormat>, instance::GetSurfacePreferredFormatError> {
+        profiling::scope!("surface_get_supported_formats");
```
cwfitzgerald, line 3021:
```suggestion
        profiling::scope!("surface::get_supported_formats");
```

wgpu-core/src/instance.rs
```diff
@@ -177,10 +166,9 @@ impl Surface {
                 .ok_or(GetSurfacePreferredFormatError::UnsupportedQueueFamily)?
         };
 
-        preferred_formats
-            .iter()
-            .cloned()
-            .find(|preferred| caps.formats.contains(preferred))
+        caps.formats
+            .is_empty()
```
cwfitzgerald, line 170: This is the inverse condition from what you want - let's just do an early return if statement for the error case

cwfitzgerald (changes requested):
One more change then we're good

wgpu-core/src/instance.rs
stevenhuyn: Good catch

wgpu-core/src/device/mod.rs
stevenhuyn: Are you sure? The function itself is `surface_get_supported_formats`

cwfitzgerald: Yeah, we should be naming the functions the wgpu (not wgpu-core) name

cwfitzgerald (approved):
Thanks!

PR #2785, opened 2022-06-18
Add back get_preferred_format and sRGB preference

victorvde:
**Connections**
Partially reverts the changes in #2783 .

**Description**
Adds back the get_preferred_format function, to avoid breaking people's code and get back the preference for sRGB.

**Testing**
Ran some examples.


wgpu/src/lib.rs
```diff
@@ -3447,13 +3447,32 @@ impl Drop for SurfaceTexture {
...
     /// Returns a vec of supported texture formats to use for the [`Surface`] with this adapter.
-    /// Note: The first format in the vector is preferred
     ///
     /// Returns None if the surface is incompatible with the adapter.
     pub fn get_supported_formats(&self, adapter: &Adapter) -> Option<Vec<TextureFormat>> {
         Context::surface_get_supported_formats(&*self.context, &self.id, &adapter.id)
     }
 
+    /// Returns an optimal texture format to use for the [`Surface`] with this adapter.
+    pub fn get_preferred_format(&self, adapter: &Adapter) -> Option<wgt::TextureFormat> {
```
cwfitzgerald, line 3457: Instead of doing this separately, both methods of "preferred" format should return the same value. How about inside of get_supported_formats we just bring the srgb formats to the front - `caps.format.sort_by_key(|f| !f.describe().srgb `. Once we support texture view changing srgb-ness we can remove the sort.

wgpu/examples/framework.rs
```diff
@@ -269,11 +269,7 @@ fn start<E: Example>(
     let spawner = Spawner::new();
     let mut config = wgpu::SurfaceConfiguration {
         usage: wgpu::TextureUsages::RENDER_ATTACHMENT,
-        format: *surface
```
cwfitzgerald, line 272: I'm not sold on needing a separate api here - especially if we remove the outer option.

cwfitzgerald (changes requested):

PR #2796, opened 2022-06-21
Add a vulkan workaround for large buffers.

nical:
**Description**

Some drivers run into fairly risky integer overflows with buffer sizes and buffer range parameters that don't fit in a 32 bits integer. Mesa/lavapipe is one of them.

This PR adds a driver workaround flag which, when set, limits buffer allocation sizes to what fits an i32.

For some context about mesa, it basically boils down to internal commands passing buffer ranges and sizes via `pipe_box`, see https://github.com/mesa3d/mesa/blob/81d6ae31d6f18d6fd2894a8b6dfe4323eea797f9/src/gallium/include/pipe/p_state.h#L529

I am not sure how to be more precise about enabling the workaround on linux. I suspect that it does not affect the nvidia proprietary driver, Iet me know if I can improve this.

Also I have no doubt that plenty of other drivers make the same mistake. If you want to be on the safe side, I don't think it would be unreasonable to set this workaround on all android devices (these tend to have more limited available memory anyway).

**Testing**

None.

### DIFF
--- a/wgpu-hal/src/dx11/adapter.rs
+++ b/wgpu-hal/src/dx11/adapter.rs
@@ -220,6 +220,8 @@ impl super::Adapter {
             max_compute_workgroup_size_y: max_workgroup_size_xy,
             max_compute_workgroup_size_z: max_workgroup_size_z,
             max_compute_workgroups_per_dimension,
+            // D3D11_BUFFER_DESC represents the buffer size as a 32 bit int.
+            max_buffer_size: u32::MAX as u64,
         };
 
         //
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -768,6 +768,15 @@ impl PhysicalDeviceCapabilities {
             .min(limits.max_compute_work_group_count[1])
             .min(limits.max_compute_work_group_count[2]);
 
+        // Prevent very large buffers on mesa and most android devices.
+        let is_nvidia = self.properties.vendor_id == crate::auxil::db::nvidia::VENDOR;
+        let max_buffer_size =
+            if (cfg!(target_os = "linux") || cfg!(target_os = "android")) && !is_nvidia {
+                i32::MAX as u64
+            } else {
+                u64::MAX
+            };
+
         wgt::Limits {
             max_texture_dimension_1d: limits.max_image_dimension1_d,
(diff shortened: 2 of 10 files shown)


nical:
Added [an entry](https://github.com/gfx-rs/wgpu/wiki/Known-Driver-Issues#mesa--vulkan) in the wiki.

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -1012,6 +1012,12 @@ impl super::Instance {
             );
         };
 
+        let is_mesa = cfg!(linux);
```
cwfitzgerald, line 1015: Lets say `(linux || android) && !nvidia` and you can identify nvidia by the PCI vendor ID. If it isn't in auxil::db, add it.

wgpu-hal/src/vulkan/device.rs
```diff
@@ -725,6 +725,15 @@ impl crate::Device<super::Api> for super::Device {
         &self,
         desc: &crate::BufferDescriptor,
     ) -> Result<super::Buffer, crate::DeviceError> {
+        if self
+            .shared
+            .workarounds
+            .contains(super::Workarounds::BUFFER_EXTENTS_I32)
+            && desc.size > std::i32::MAX as wgt::BufferAddress
+        {
+            return Err(crate::DeviceError::OutOfMemory);
```
cwfitzgerald, line 734: We should still reduce the maximum buffer size limit to 2GB so that we don't need this check here and the user is aware of the hard limit

cwfitzgerald (changes requested):

wgpu-hal/src/dx11/adapter.rs
```diff
@@ -220,6 +220,8 @@ impl super::Adapter {
             max_compute_workgroup_size_y: max_workgroup_size_xy,
             max_compute_workgroup_size_z: max_workgroup_size_z,
             max_compute_workgroups_per_dimension,
+            // D3D11_BUFFER_DESC represents the buffer size as a 32 bit int.
+            max_buffer_size: std::u32::MAX as u64,
```
cwfitzgerald, line 224: Let's use the newer style `u32::MAX` here (and elsewhere)

cwfitzgerald (changes requested):
One nit, otherwise LGTM

cwfitzgerald (approved):
Thanks!

PR #2800, opened 2022-06-22
Add create_texture_error

nical:
**Description**

Firefox needs it to implement WebGPU's error model (see create_buffer_error). The WebGPU specification allows texture descriptors that are not allowed by wgpu's TextureDescriptor type such as the `usage`  bitfield. Firefox takes care of validating these special cases, however the invalid textures have to be registered which is done via this method. See also `create_buffer_error`. 

**Testing**

None.


cwfitzgerald (approved):
Nice

Issue #2788, opened 2022-06-19
Allow create_shader_module to receive a Naga module as ShaderSource


rttad:
**Is your feature request related to a problem? Please describe.**
Ultimately, when creating a shader module on a native backend, the shader source is parsed into a `naga::Module` so that wgpu can serve different backends at once. This means no matter what shading language you pass as input, there will be an inevitable parsing step to an intermediate representation. This is wasting runtime performance when we could just be embedding the intermediate representation itself.

**Describe the solution you'd like**
To allow that, the `ShaderSource` API could be extended to allow a new variant: `ShaderSource::Naga(naga::Module)`. The module provided by the user would be passed down as-is in `wgc::pipeline::ShaderModuleSource::Naga(module)` on native, where things like validation and runtime checks are already handled. On the web, it would be validated and written back to WGSL.

**Describe alternatives you've considered**
Embedding IR has the advantage of being platform-agnostic while reducing binary size, but it will still require a translation step to native shaders. It seems to be a fair trade-off with precompiled shaders when you need to retain the cross-platform aspect but still want the performance gains.

**Additional context**
`naga::Module` can be [serialized](https://docs.rs/naga/latest/naga/struct.Module.html#impl-Serialize), indicating it can be used as a shader data format like any other.

jimblandy:
It's true that `naga::Module` can be serialized, but we change the IR pretty frequently. It would not be a stable storage format.

rttad:
The main use case for this feature is to parse and use shaders in the same project, where compatibility issues are unlikely to occur unless different versions of `naga` are being used. In that case, it would be up to the library user to check compatibility with whichever version they are using until the IR is stabilized.

PR #2801, opened 2022-06-22
Add Naga variant to ShaderSource

rttad:
**Connections**
Implemention of #2788

**Description**
Adds a new `Naga(Cell<naga::Module>)` variant to ShaderSource. The module has to be wrapped in `std::cell::Cell` because `naga::Module` does not implement Copy.

**Testing**
`cargo check --features naga` passed on native target and WebAssembly.

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -257,7 +257,7 @@ trait Context: Debug + Send + Sized + Sync {
     fn device_create_shader_module(
         &self,
         device: &Self::DeviceId,
-        desc: &ShaderModuleDescriptor,
+        desc: ShaderModuleDescriptor,
         shader_bound_checks: wgt::ShaderBoundChecks,
     ) -> Self::ShaderModuleId;
     unsafe fn device_create_shader_module_spirv(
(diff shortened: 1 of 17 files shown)


wgpu/src/lib.rs
```diff
@@ -812,6 +812,10 @@ pub enum ShaderSource<'a> {
     },
     /// WGSL module as a string slice.
     Wgsl(Cow<'a, str>),
+    /// Naga module.
+    #[cfg(feature = "naga")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "naga")))]
+    Naga(std::cell::Cell<naga::Module>),
```
cwfitzgerald, line 818: Not too happy about this, but can't think of a better solution.

jimblandy:
If we use `Cell` in `wgpu::ShaderSource`, then that makes it not implement `Sync`, which can be a pain. I think it should be fine if we simply pass in the `naga::Module` by value:

- give `ShaderSource` a variant `Naga(naga::Module)`, that owns the module,
- change `create_shader_module` to simply take the `ShaderModuleDescriptor` by value, not by reference, and
- plumb that through `wgpu::Context::device_create_shader_module`.

Then the `direct` back end can just move the `naga::Module` into the `wgpu_hal::ShaderInput` it creates.

This is a breaking API change, but now's the time to make such changes. I've pushed changes to pass `ShaderModuleDescriptor` by value here, in case you want to rebase your change on that: https://github.com/jimblandy/wgpu/tree/pass-ShaderModuleDescriptor-by-value
It's a big patch, but it's mechanical.

rttad:
Had to force push to fix a git manipulation. The change should now be rebased correctly.

cwfitzgerald (approved):
Cool beans

PR #2802, opened 2022-06-23
Restrict WriteTimestamp Inside Passes

### LINKED ISSUE #2738: Feature Gate Timestamp Queries Inside Renderpasses/ComputePasses

Follows upstream changes while still letting people who are on dx12 and vulkan get their timestamps.

Getting rid of them entirely. 

**Additional context**

https://github.com/gpuweb/gpuweb/pull/2190 is the most recent modification of timestamp queries

cwfitzgerald:
**Connections**

Closes #2738 

**Description**

This brings us more in line with upstream and rips of the bandaid before we get timestamp query support on metal.

Also cleaned up a bit, commits separately reviewable.

**Testing**

Stared at it a good while.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -173,6 +173,12 @@ bitflags::bitflags! {
     #[repr(transparent)]
     #[derive(Default)]
     pub struct Features: u64 {
+        //
+        // ---- Start numbering at 1 << 0, each one defined in terms of the last ----
+        //
+        // WebGPU features:
+        //
+
         /// By default, polygon depth is clipped to 0-1 range before/during rasterization.
         /// Anything outside of that range is rejected, and respective fragments are not touched.
         ///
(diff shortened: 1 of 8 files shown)


wgpu-types/src/lib.rs
```diff
@@ -173,6 +173,12 @@ bitflags::bitflags! {
     #[repr(transparent)]
     #[derive(Default)]
     pub struct Features: u64 {
+        //
+        // ---- Start numbering at 1 << 0, each one defined in terms of the last ----
```
jimblandy, line 177: This is an interesting idea. The idea is to let us insert new values wherever we like, without having to renumber everything, right?

I just hope we don't find ourselves needing to understand hex `Features` values, because it'll be a bear to figure out anything's actual numeric value. (Note that this has no `Debug` impl.)

Thank you for the new docs!

jimblandy (approved):

wgpu-types/src/lib.rs
cwfitzgerald: All bitflags have debug impls, we use them all over for printing features.

But yes, this way we're just inserting without needing to do that error prone and menial work

PR #2809, opened 2022-06-25
gles: Fix buffer-texture copy for 2d arrays

tuchs:
**Connections**
None

**Description**
For buffer-to-texture copies into 2D array textures the GLES HAL incorrectly used origin.z (which is always 0 here) for the depth argument, this change corrects this by using the array_layer field instead.

**Testing**
Use write_texture on a 2D_ARRAY with origin.z != 0.


cwfitzgerald (approved):
Thanks!

PR #2811, opened 2022-06-26
Remove redundant `#[cfg]` conditions from `backend/direct.rs`.

jimblandy:
The entire `wgpu::backend::direct` module is included only under the
condition:

    #[cfg(any(not(target_arch = "wasm32"), feature = "webgl"))]

So there is no need to apply this condition to any item within that
module.


cwfitzgerald (approved):
Dope!

PR #2814, opened 2022-06-27
egl: recreate Wayland context if X surface is supplied

### LINKED ISSUE #2762: wgpu wrongly assumes Wayland as the default platform.
**Description**
wgpu when libwayland is installed will try to use the Wayland platform but fail to present to an X11 window.

**Repro steps**
With libwayland-client installed, run `WINIT_UNIX_BACKEND=x11 cargo run --example=hello-triangle`

`WINIT_UNIX_BACKEND=x11` will force winit to create X11 windows.

wgpu should present using X11.

**Extra materials**
```
[2022-06-10T22:54:35Z ERROR wgpu_core::device] surface configuration failed: incompatible window kind
thread 'main' panicked at 'Error in Surface::configure: invalid surface', wgpu/src/backend/direct.rs:249:9
```

**Platform**
Information about your OS, version of `wgpu`, your tech stack, etc.

Linux, master branch, running Wayland under KDE.

RX 6700XT w/ Mesa.

sh7dm:
**Connections**
Fixes #2762

**Description**
When user has libwayland and WAYLAND_DISPLAY, wgpu creates EGLDisplay using EGL_PLATFORM_WAYLAND_KHR, which is incompatible with X windows. However, application might use X in Wayland environment, and so its surface will have incompatible type thus leading to panic. In this case reinitialize the EGLDisplay with EGL_PLATFORM_X11_KHR.

**Testing**
WINIT_UNIX_BACKEND=x11

### DIFF
--- a/wgpu-hal/src/gles/egl.rs
+++ b/wgpu-hal/src/gles/egl.rs
@@ -764,6 +764,7 @@ impl crate::Instance<super::Api> for Instance {
         let mut inner = self.inner.lock();
 
         match raw_window_handle {
+            #[cfg(not(feature = "emscripten"))]
             Rwh::Xlib(handle) => {
                 // If Wayland EGLDisplay was created, but we got an X window, recreate the context
                 // https://github.com/gfx-rs/wgpu/issues/2762
--- a/wgpu-hal/src/gles/egl.rs
+++ b/wgpu-hal/src/gles/egl.rs
@@ -407,6 +407,7 @@ struct Inner {
     wl_display: Option<*mut raw::c_void>,
     /// Method by which the framebuffer should support srgb
     srgb_kind: SrgbFrameBufferKind,
+    wsi: WindowSystemInterface,
 }
 
 impl Inner {
@@ -414,6 +415,7 @@ impl Inner {
         flags: crate::InstanceFlags,
         egl: Arc<EglInstance>,
         display: egl::Display,
+        wsi: WindowSystemInterface,
     ) -> Result<Self, crate::InstanceError> {
         let version = egl.initialize(display).map_err(|_| crate::InstanceError)?;
         let vendor = egl.query_string(Some(display), egl::VENDOR).unwrap();
@@ -541,6 +543,7 @@ impl Inner {
             config,
             wl_display: None,
             srgb_kind,
+            wsi,
         })
     }
 }
@@ -575,7 +578,6 @@ struct WindowSystemInterface {
 }
 
 pub struct Instance {
-    wsi: WindowSystemInterface,
     flags: crate::InstanceFlags,
     inner: Mutex<Inner>,
 }
@@ -733,13 +735,17 @@ impl crate::Instance<super::Api> for Instance {
             (function)(Some(egl_debug_proc), attributes.as_ptr());
         }
 
-        let inner = Inner::create(desc.flags, egl, display)?;
-
-        Ok(Instance {
-            wsi: WindowSystemInterface {
+        let inner = Inner::create(
+            desc.flags,
+            egl,
+            display,
+            WindowSystemInterface {
                 library: wsi_library,
                 kind: wsi_kind,
             },
+        )?;
+
+        Ok(Instance {
             flags: desc.flags,
             inner: Mutex::new(inner),
         })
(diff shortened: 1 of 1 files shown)


wgpu-hal/src/gles/egl.rs
```diff
@@ -758,8 +764,47 @@ impl crate::Instance<super::Api> for Instance {
...
 
         match raw_window_handle {
-            Rwh::Xlib(_) => {}
-            Rwh::Xcb(_) => {}
+            #[cfg(not(feature = "emscripten"))]
+            Rwh::Xlib(handle) => {
+                // If Wayland EGLDisplay was created, but we got an X window, recreate the context
+                // https://github.com/gfx-rs/wgpu/issues/2762
+                if inner.wl_display.is_some() || inner.wsi.kind == WindowKind::Wayland {
+                    log::warn!("Re-initializing Gles context due to Xlib window");
```
i509VCB, line 772: One interesting case to consider is that wgpu takes a RawWindowHandle. Nothing actually prevents you from creating a surface using x11 and another with Wayland in the same application.

Vulkan handles that properly but not egl.

wgpu-hal/src/gles/egl.rs
```diff
@@ -758,8 +764,47 @@ impl crate::Instance<super::Api> for Instance {
...
+                    .map_err(|_| crate::InstanceError)?;
+
+                    let old_inner = std::mem::replace(inner.deref_mut(), new_inner);
+                    inner.wl_display = None;
+                    drop(old_inner);
+                }
+            }
+            Rwh::Xcb(_) => {
+                // TODO: support Xcb context reinit
+                // https://www.khronos.org/registry/EGL/extensions/EXT/EGL_EXT_platform_xcb.txt
```
i509VCB, line 806: Last time I recall Mesa wasn't advertising this extension? Probably need to check again.

sh7dm: I left a TODO just because didn't find any necessity to implement this

wgpu-hal/src/gles/egl.rs
```diff
@@ -758,8 +764,47 @@ impl crate::Instance<super::Api> for Instance {
         let mut inner = self.inner.lock();
 
         match raw_window_handle {
-            Rwh::Xlib(_) => {}
-            Rwh::Xcb(_) => {}
+            #[cfg(not(feature = "emscripten"))]
+            Rwh::Xlib(handle) => {
+                // If Wayland EGLDisplay was created, but we got an X window, recreate the context
+                // https://github.com/gfx-rs/wgpu/issues/2762
+                if inner.wl_display.is_some() || inner.wsi.kind == WindowKind::Wayland {
```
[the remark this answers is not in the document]
sh7dm: But the same happens a bit further in the code (same match), Wayland context is replaced by another one. 🤷

wgpu-hal/src/gles/egl.rs
i509VCB: Maybe a solution could be to add the following to create_surface:

> All created surfaces must have the same type of RawWindowHandle

sh7dm: Hmm, exactly! Should it be just a comment or a warning emitted on second change of context type?

sh7dm: I guess there is the need to create something like The Book of Wgpu to document such edge cases and the overall design of an application.

wez:
I've been looking at this from over in https://github.com/gfx-rs/wgpu/issues/3922

I'm not as familiar as the rest of you with the wgpu code, but my perception of this issue is that deferring the fixup to create_surface time makes resolving this issue more awkward.

What do you think about passing a `RawDisplayHandle` into `Instance::new` and using that to set things up appropriately?
I'm not sure if that would work for every target.

i509VCB:
> What do you think about passing a `RawDisplayHandle` into `Instance::new` and using that to set things up appropriately? I'm not sure if that would work for every target.

If the parameter is optional or we provide a seperate with_display that would work.

There is also the nuclear option like always which involves advertising an egl adapter per supported display platform (so a device for x11 and a device for Wayland)

wez:
> There is also the nuclear option like always which involves advertising an egl adapter per supported display platform (so a device for x11 and a device for Wayland)

Sounds kinda appealing, but I think there's still (at least for me) a conceptual issue with not having the application-provided display handle in the egl context.  There's some existing fixup logic that inspects and compares wl_display that is a specialization of this more general problem.

Maybe the way to go is to embrace the craziness and change Instance so that it has an Inner for enumerate_adapters and separate deferred-initialized Inner that is setup based on the parameters passed to the first create_surface call.

Hmm, I think there's another problem with this, because AdapterContext::raw_display could return a different display from the one used by the application.  I don't know the full extent of the consequences of that!

cwfitzgerald:
As this MR is currently stale, I'm going to close it for triage purposes. Please feel-free to re-open there is renewed interest in landing these changes, or other ones that serve the same purpose.

PR #2821, opened 2022-06-29
Split clear_value from LoadOp::Clear(Color) to match spec

jinleili:
**Connections**
From WebGPU Spec: https://gpuweb.github.io/gpuweb/#color-attachments

**Testing**
Tested on metal backend and wasm target

### DIFF
--- a/wgpu/examples/boids/main.rs
+++ b/wgpu/examples/boids/main.rs
@@ -276,6 +276,7 @@ impl framework::Example for Example {
             view,
             resolve_target: None,
             ops: wgpu::Operations {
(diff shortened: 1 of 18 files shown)


cwfitzgerald (changes requested):
I don't think we should actually adopt this change from the spec -- it doesn't actually buy us any expressiveness and exposes more nonsense configurations -- It makes sense to me to lean on a Color::default() being defined as black versus adding another layer of option just for the default value. Thoughts?

jinleili:
Totally agree with what you think.

PR #2822, opened 2022-06-29
Allow fragment stage to have more outputs than the pipeline has

jinleili:
**Connections**
Follow https://github.com/gfx-rs/wgpu/issues/2469
WebGPU Spec: https://gpuweb.github.io/gpuweb/#fragment-state

**Testing**
Tested on metal, gl, vulkan, dx12 backends.
wasm target can not running on Chrome Canary because Dawn's validation failure:
```sh
Color target has no corresponding fragment stage output but writeMask (ColorWriteMask::(Red|Green|Blue|Alpha)) is not zero.
 - While validating targets[0].
 - While validating fragment state.
 - While calling [Device].CreateRenderPipeline([RenderPipelineDescriptor]).
localhost/:1 [Invalid RenderPipeline] is invalid.
 - While encoding [RenderPassEncoder].SetPipeline([Invalid RenderPipeline]).
```

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -2774,16 +2774,14 @@ impl<A: HalApi> Device<A> {
                             },
                         )?;
                     }
-                    Some(&None) => {
-                        return Err(
-                            pipeline::CreateRenderPipelineError::InvalidFragmentOutputLocation(*i),
-                        );
-                    }
                     _ => {
-                        return Err(pipeline::CreateRenderPipelineError::ColorState(
-                            *i as u8,
-                            pipeline::ColorStateError::Missing,
-                        ));
+                        log::info!(
+                            "The fragment stage {:?} output @location({}) values are ignored",
+                            fragment_stage
+                                .as_ref()
+                                .map_or("", |stage| stage.entry_point),
+                            i
+                        );
                     }
                 }
             }
--- a/wgpu-core/src/pipeline.rs
+++ b/wgpu-core/src/pipeline.rs
@@ -280,8 +280,6 @@ pub struct RenderPipelineDescriptor<'a> {
 
 #[derive(Clone, Debug, Error)]
 pub enum ColorStateError {
-    #[error("output is missing")]
-    Missing,
     #[error("format {0:?} is not renderable")]
     FormatNotRenderable(wgt::TextureFormat),
     #[error("format {0:?} is not blendable")]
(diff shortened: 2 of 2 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -2774,16 +2774,11 @@ impl<A: HalApi> Device<A> {
...
-                        return Err(
-                            pipeline::CreateRenderPipelineError::InvalidFragmentOutputLocation(*i),
-                        );
-                    }
                     _ => {
-                        return Err(pipeline::CreateRenderPipelineError::ColorState(
-                            *i as u8,
-                            pipeline::ColorStateError::Missing,
-                        ));
+                        log::warn!(
```
cwfitzgerald, line 2778: I don't think we should warn on this - it's going to be really noisy if people intentionally are doing this.

cwfitzgerald (changes requested):
One comment, otherwise lgtm

wgpu-core/src/device/mod.rs
jinleili: Is it okay to change it to log::info!()? which is helpful for users when setting up `RUST_LOG=info` to debug the program.

cwfitzgerald: Yeah, let's do info. We should probably re-evaluate logging levels at some point.

wgpu-core/src/device/mod.rs
```diff
@@ -2774,16 +2774,11 @@ impl<A: HalApi> Device<A> {
...
-                            pipeline::CreateRenderPipelineError::InvalidFragmentOutputLocation(*i),
-                        );
-                    }
                     _ => {
-                        return Err(pipeline::CreateRenderPipelineError::ColorState(
-                            *i as u8,
-                            pipeline::ColorStateError::Missing,
-                        ));
+                        log::info!(
+                            "The fragment shader output @location({}) values are ignored",
```
cwfitzgerald, line 2779: Could we drop in shader ID/label in here, do we have it available?

cwfitzgerald (approved):
Non-blocking comment

wgpu-core/src/device/mod.rs
jinleili: ShaderModuleId output is like `(0, 1, Metal)`, which is not helpful. entry_point is available but shader label is not.

cwfitzgerald: Good 'nuff

cwfitzgerald (approved):

jinleili:
Another rare CI failure again :)

cwfitzgerald:
Tfw rare CI failures are common - maybe if I just close my eyes and go lalala I won't consider the fact it might be a bug in wgpu - but then I would have to have even the faintest idea how to fix it....

Issue #687, opened 2020-05-31
Window alpha transparency support


valpackett:
**Is your feature request related to a problem? Please describe.**
Currently, windows are hardcoded to be opaque:

```rust
config.composite_alpha_mode = hal::window::CompositeAlphaMode::OPAQUE;
```

so the `a` in `wgpu::RenderPassColorAttachmentDescriptor` `clear_color` doesn't do anything.

**Describe the solution you'd like**
There should be support for the transparent modes. In particular, premultiplied mode should be handled, as it's the only one supported at least on my machine (using Mesa RADV driver for Radeon).

**Additional context**
Hardcoding `::PREMULTIPLIED` makes transparency work, but I actually have to multiply the colors by the alpha:

![Screen_2020-05-31-14:44:07-fs8](https://user-images.githubusercontent.com/208340/83351592-fb2c1180-a34d-11ea-90bd-185a23e2ef69.png)

kvark:
I would be fine with exposing the composite alpha mode

myfreeweb:
ah good old "let the user handle it" :)

Swapchain creation is public API so this affects `webgpu-headers`…

kvark:
Is there an alternative to telling users what modes are available, and letting them select one?
I'd be happy to consider one!

As for webgpu-headers - you are right, we'll need to expose it there as well, just like we expose the present modes. cc @Kangz

myfreeweb:
I was thinking "pick the most common mode and handle conversions" but yeah, exposing everything to the user would be much easier.

Kangz:
Can we have a brief investigation of what the compositing modes are and how they translate to the various APIs? I worry for example that on macOS the compositing mode might be a part of the CALayer so we'd be mutating that.

Also are there any guarantees that we can support all compositing modes on all platforms, or does for example Vulkan give little guarantees so we need to have queries in webgpu.h for it?

kvark:
> Can we have a brief investigation of what the compositing modes are and how they translate to the various APIs?

Agreed, would be great to have it!

>I worry for example that on macOS the compositing mode might be a part of the CALayer so we'd be mutating that.

Pretty sure we are already mutating `CALayer` properties when you are re-configuring the swapchain, since we set the size and the vsync, etc

> Also are there any guarantees that we can support all compositing modes on all platforms

No

atsuzaki:
Any updates on this?

nyxtom:
[...]

PR #2836, opened 2022-07-02
Expose alpha_mode, support non-opaque mode on metal and vk backends

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
https://github.com/gfx-rs/wgpu/issues/687

**Description**

##### To get a transparent background `Canvas`, there are different cases on the user side:
- If use  `winit`, have to set  `window_builder.with_transparent(true)` ;
- If use [custom `SurfaceView`](https://github.com/jinleili/wgpu-on-app/blob/master/Android/app/src/main/java/name/jinleili/wgpu/WGPUSurfaceView.kt) on Android, the only way to set SurfaceView background color to transparent is:
```kotlin
val sv = WGPUSurfaceView(ctx)
sv.setZOrderOnTop(true)  
sv.holder.setFormat(PixelFormat.TRANSPARENT)
```
- If call `create_surface_from_core_animation_layer` on metal to create surface, have to ensure that the background color of the layer is nil or clearColor

##### Known issues:
- wasm target cannot update alpha_mode, `web-sys` documentation states that `GpuCanvasAlphaMode` is supported, but it is not possible to configure this feature in `.toml`. nor to call the `alpha_mode()`
- dx12 setting non-opaque alpha_mode on Windows 11 causes crash(running on Parallels Desktop), haven't found a solution yet, checked out the source code for `Dawn` and it is hard coded `swapChainDesc.AlphaMode = DXGI_ALPHA_MODE_IGNORE`:
https://dawn.googlesource.com/dawn/+/24239fcc4786bc231eacc065eaa49fd484546984/src/dawn/native/d3d12/SwapChainD3D12.cpp#243

**Testing**
Tested on macOS(metal, vulkan-portability, angle), iOS, Nexus 5x(Android 8.1, vk and gl backends).
| macOS|  android|
-----  |-----  |


wgpu-hal/src/metal/surface.rs
```diff
@@ -186,9 +186,21 @@ impl crate::Surface<super::Api> for super::Surface {
         let drawable_size = CGSize::new(config.extent.width as f64, config.extent.height as f64);
 
         match config.composite_alpha_mode {
-            crate::CompositeAlphaMode::Opaque => render_layer.set_opaque(true),
-            crate::CompositeAlphaMode::PostMultiplied => render_layer.set_opaque(false),
-            crate::CompositeAlphaMode::PreMultiplied => (),
+            wgt::CompositeAlphaMode::Opaque => render_layer.set_opaque(true),
```
geertbleyen, line 189: Does this mean that when composite_alpha_mode is not Opaque, the background will always be clear color? Even when the background color used to be some other color with non opaque alpha?

jinleili: The metalLayer's background color is like the color of a canvas itself, while ColorAttachment's wgpu::LoadOp::Clear is the color to fill the specified area of the canvas. 
```
metal_layer.set_opaque(false); 
msg_send![metal_layer, setBackgroundColor: clear_color];
```
these code means that the canvas itself is transparent, and any colors drawn on this canvas with transparency will be composited by the layer behind it, which is exactly the purpose of using non-opaque AlphaMode.

geertbleyen: I understand, but I'm only wondering if it is ok to override the user-set background color (which could or could not be opaque). Would the user expect that when he sets the background to (1.0, 0.0, 0.0, 0.5) that (s)he gets a half transparent red background?
I thought the CALayer's isOpaque flag indicates if the opacity component of the content when drawing into it should be ignored or not (for layer composition efficiency reasons) -> https://developer.apple.com/documentation/quartzcore/calayer/1410763-isopaque
It doesn't mention that the background color should be fully transparent (clear color).
We currently use a fork of wgpu and patch ``` crate::CompositeAlphaMode::Opaque => render_layer.set_opaque(true) ``` to call ```set_opaque(false)``` instead. Without any changes to the metal layer's background, we are getting the result we need (we never explicitly set the MetalLayer's background anywhere). So I'm assuming you were setting the layer's background somewhere else (can't find where wgpu's metal backend would be changing it)?

jinleili: As my code comment mentions the reason：
```
// CALayer may have a background color by default,
// if the alpha_mode of the surface is set to non-opaque, unexpected colors are obtained.
```
It's also possible to remove `setBackgroundColor`  line of code, which is only about how we think about: the `metal_layer` may have been created by a window lib and not directly accessible to the user.

geertbleyen: With unexpected colors, you mean incorrect colors? Or do you mean that because the background color is not nil (default for a MetalLayer) and was set (by the application or external window lib) to a not 100% transparent color, that you get blended colors with that background from the MetalLayer (because it's not clear)?

jinleili: I mean the second case

geertbleyen: Isn't it then up to that window lib to allow for controller the CAMetalLayer's backgroundColor property instead of overriding the one set by users here in wgpu? Somewhere you are passing your UIView to wpgu, I assume you can use that integration point to alter the view's main layer's backgroundColor.
We are working on an iOS app, so in our case wpgu always will create the CAMetalLayer, as the view's main layer can't be changed on iOS, and that way the metal_layer's background is always nil.
I'd prefer wpgu not to alter background colors of layer implicitely without a way to opt-in/out.

cwfitzgerald:
No, just my highly split attention. I'll take a look soon. Thank you for reviewing!

wgpu-hal/src/metal/surface.rs
geertbleyen: Thanks @jinleili for the changes!

wgpu-hal/src/dx12/adapter.rs
```diff
@@ -489,9 +489,9 @@ impl crate::Adapter<super::Api> for super::Adapter {
                 | crate::TextureUses::COPY_DST,
             present_modes,
             composite_alpha_modes: vec![
-                crate::CompositeAlphaMode::Opaque,
-                crate::CompositeAlphaMode::PreMultiplied,
-                crate::CompositeAlphaMode::PostMultiplied,
+                wgt::CompositeAlphaMode::Opaque,
```
cwfitzgerald, line 492: Why are we exposing modes other than Opaque if that's all flip model swapchains support?

cwfitzgerald (changes requested):
Code generally looks good - one comment. We should also add a function to query what modes are actually available and enforce that the user only uses an available mode (similar to how we do present modes now).

Hoverbear:
I had the opportunity to test this patch atop `v0.13.1` in Bevy as part of exploring the cause of https://github.com/bevyengine/bevy/issues/5779. It seemed functional and working in x86_64 Linux on a Vulkan backend with a `AMD RADV NAVI10` adapter running inside Wayland (Gnome).

geertbleyen:
@jinleili , do you intend to continue work on this PR?

jinleili:
> @jinleili , do you intend to continue work on this PR?

Yes, I still want to get it done.

jinleili:
> Code generally looks good - one comment. We should also add a function to query what modes are actually available and enforce that the user only uses an available mode (similar to how we do present modes now).

Is it acceptable to simply match the [WebGPU Spec](https://gpuweb.github.io/gpuweb/#GPUCanvasAlphaMode) and provide only normal `Opaque` and `PreMultiplied` options?

jimblandy (approved):
It looks to me like @jinleili has implemented all of @cwfitzgerald's requested changes.

cwfitzgerald (changes requested):
Still need a way to query which transparency modes are supported and validation that the requested mode is supported.

jinleili:
@cwfitzgerald Also renamed `get_supported_modes` to `get_supported_present_modes`.

jinleili:
Since `alpha_mode`  is now strictly validated, `Opaque` is no longer guaranteed to be supported, because the vk backend on Android devices tends to only support `Inherit`

geertbleyen:
@cwfitzgerald , these changes are now ok to merge?

cwfitzgerald (approved):
Great stuff - sorry for the delay on merging!

codeart1st:
Great work @jinleili 
Looking forward this change in the next release👍

PR #2842, opened 2022-07-03
Fixes #2840

### LINKED ISSUE #2840: v0.13 - BrowserWebGpu - AutoNoVSync does not properly fallback
**Description**
AutoNoVSync appears to not fall-back correctly.

**Repro steps**
https://github.com/simbleau/nbody-wasm-sim/tree/4cd3b2fadb0bed6b71ef3253f3989d78fd5d2020
`wasm-pack build --target web`
`python -m http.server`
[`http://localhost:8000/`](http://localhost:8000/)

Error stack forwarded to the console via WASM:
`panicked at 'Error in Surface::configure: requested present mode AutoNoVsync is not in the list of supported present modes: [Fifo]'`

This seems like a bug because the [AutoNoVSync documentation](https://docs.rs/wgpu/latest/wgpu/enum.PresentMode.html#variant.AutoNoVsync) says it will fallback to Fifo, but clear it is not, and detects Fifo as a supported present mode.

**Extra materials**

**Platform**
wgpu 0.13
linux, ubuntu 20.04 LTS

simbleau:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
This fixes #2840 

**Description**
AutoNoVSync should fall back on Fifo as a last resort. This is important for Web, since Web only has Fifo. [Documentation already shows this should use Fifo](https://docs.rs/wgpu/latest/wgpu/enum.PresentMode.html#variant.AutoNoVsync), but it was missing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -289,6 +289,7 @@ DeviceDescriptor {
 
 #### General
 
+- AutoNoVSync now correctly falls back to Fifo @simbleau in [#2842](https://github.com/gfx-rs/wgpu/pull/2842)
 - Fix trac(y/ing) compile issue by @cwfitzgerald in [#2333](https://github.com/gfx-rs/wgpu/pull/2333)
 - Improve detection and validation of cubemap views by @kvark in [#2331](https://github.com/gfx-rs/wgpu/pull/2331)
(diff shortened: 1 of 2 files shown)


cwfitzgerald (approved):
Thanks!

PR #2848, opened 2022-07-04
Ensure the BufferMapCallback is always called.

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

[WebGPU specification](https://gpuweb.github.io/gpuweb/#buffer-mapping)

**Description**

This PR enforces that the mapping callback is always called (be it with an error). It does so by calling it manually in more places (with corresponding error codes) and by catching the rest in the callback's destructor.

This solves two issues on the Gecko side:
 - The callback cleans up after itself (the user data is deleted at the end of the callback), so dropping the callback without calling it is a memory leak. I can't think of a better way to implement this on the C++ side since there can be any number of callbacks at any time living for an unspecified amount of time.
 - This makes it easier to implement the error reporting of the WebGPU spec.

Note that when the callback is fired by its `Drop` impl, there is no guarantee that it happens outside of the locks. If that's a big issue, we could instead panic in the drop impl and be very careful about the control flow.

Also, the WebGPU spec specifies what should happen when destroying a buffer that is mapped or pending. The easiest way to conform to that is to just call unmap. I suspect that it will also avoid issues with wgpu itself?

**Testing**

None.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -7,7 +7,9 @@ use crate::{
         BufferInitTracker, BufferInitTrackerAction, MemoryInitKind, TextureInitRange,
         TextureInitTracker, TextureInitTrackerAction,
     },
-    instance, pipeline, present, resource,
+    instance, pipeline, present,
+    resource::{self, BufferMapState},
+    resource::{BufferAccessError, BufferMapAsyncStatus, BufferMapOperation},
     track::{BindGroupStates, TextureSelector, Tracker},
     validation::{self, check_buffer_usage, check_texture_usage},
     FastHashMap, Label, LabelHelpers as _, LifeGuard, MultiRefCount, RefCount, Stored,
@@ -127,7 +129,7 @@ impl RenderPassContext {
     }
 }
 
-pub type BufferMapPendingClosure = (resource::BufferMapOperation, resource::BufferMapAsyncStatus);
+pub type BufferMapPendingClosure = (BufferMapOperation, BufferMapAsyncStatus);
 
 #[derive(Default)]
 pub struct UserClosures {
@@ -159,7 +161,7 @@ fn map_buffer<A: hal::Api>(
     offset: BufferAddress,
     size: BufferAddress,
     kind: HostMap,
-) -> Result<ptr::NonNull<u8>, resource::BufferAccessError> {
+) -> Result<ptr::NonNull<u8>, BufferAccessError> {
     let mapping = unsafe {
         raw.map_buffer(buffer.raw.as_ref().unwrap(), offset..offset + size)
             .map_err(DeviceError::from)?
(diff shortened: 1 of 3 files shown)


nical:
I added this to the changelog because the PR template says so, but I wouldn't be offended if this type of plumbing isn't changelog-worthy!

nical:
Hm, I just realized that `map_async` was already firing the callback inside of the locks and I added more of that. Let me know how important that is and in what cases. I'll update the PR so that it does whatever the right thing is.

nical:
I updated the PR to be more principled about invoking the callback outside of the lock in buffer_map_async.

jimblandy (changes requested):
This all makes sense - thank you!

There are a few things that I think need to be addressed.

wgpu-core/src/device/mod.rs
```diff
@@ -5331,10 +5353,54 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+        &self,
+        buffer_id: id::BufferId,
+        range: Range<BufferAddress>,
+        op: BufferMapOperation,
+    ) -> Result<(), (BufferMapOperation, BufferAccessError)> {
         profiling::scope!("Buffer::map_async");
 
+        // Note: We need the callback to be called in all error cases, however it must be called
+        // outside of buffer_map_async_inner, so that we don't hold any locks. Ensure that any
+        // early path that doesn't lead to a pending map state sets closure_to_resolve accordingly.
```
jimblandy, line 5402: This comment is out of date. I don't think we need it at all anymore, since the types essentially require us to hand back `op` on failure. Perhaps the best place for the comment is on the function itself, explaining why it has an interesting return type.

wgpu-core/src/device/mod.rs
```diff
@@ -5437,41 +5512,31 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             }
             resource::BufferMapState::Active { ptr, ref range, .. } => {
                 if offset < range.start {
-                    return Err(resource::BufferAccessError::OutOfBoundsUnderrun {
+                    return Err(BufferAccessError::OutOfBoundsUnderrun {
```
jimblandy, line 5515: This is fine for this PR, but for the future, it helps me review more quickly and accurately if mechanical changes like this are separated out from substantive ones. Putting it in either a separate PR or just a separate commit within the PR is fine.

wgpu-core/src/resource.rs
```diff
@@ -13,14 +13,35 @@ use thiserror::Error;
 
 use std::{borrow::Borrow, num::NonZeroU8, ops::Range, ptr::NonNull};
 
+/// The status code provided to the buffer mapping callback.
```
jimblandy, line 16: It might be nice to explain,

> This is very similar to `Result<(), BufferAccessError>`, except that this is FFI-friendly.

wgpu-core/src/resource.rs
```diff
@@ -71,33 +92,45 @@ enum BufferMapCallbackInner {
...
     }
 
-    pub(crate) fn call(self, status: BufferMapAsyncStatus) {
-        match self.inner {
-            BufferMapCallbackInner::Rust { callback } => callback(status),
+    pub(crate) fn call(mut self, status: BufferMapAsyncStatus) {
+        self.call_impl(status);
+    }
+
+    fn call_impl(&mut self, status: BufferMapAsyncStatus) {
```
jimblandy, line 113: This can be folded back into `call` now, I think.

jimblandy (approved):
Looks good, thank you! Just a few requests.

jimblandy (approved):
Thanks!

PR #2853, opened 2022-07-04
Prevent OpenGL from taking preference over Vulkan

Craig-Macomber:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Issue is worked around in https://github.com/Craig-Macomber/rusty-flame/commit/8a20c7f5efb35e437c462ed5d8bc1781109ea1c7#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fcR115

**Description**

Prefer `DeviceType::DiscreteGpu` over `DeviceType::Other` for `PowerPreference::LowPower` so Vulkan is preferred over OpenGL again.

This change: https://github.com/gfx-rs/wgpu/commit/bfcf5fa513d51f9f4de4318f9627b39b528c8977 caused default configurations (low power, all back-ends) such as the "hello" example to prefer OpenGL over Vulkan for DiscreteGpu devices.

This happens because OpenGL lists the device_type as "Other" when its actually "DiscreteGpu" (but OpenGL does not know this).

Since the OpenGL backend currently never reports any device as DiscreteGpu, it seems like a good idea to make sure "Other" is not preferred over DiscreteGpu: without this GL will get preferred over Vulkan when both are present for the DiscreteGpu.

The low power configuration (which is the default) used to violate this (after the above linked change) which resulted in regressions where GL would get selected over Vulkan, causing reduced feature sets.

This fix is not ideal: it could regress some cases including multi-gpu systems where one of them gets device type Other, and thats actually lower power than the one listed as DiscreteGpu.

It also seems bad to be using "Other" when we mean "Unknown", so I have updated the comment on that device type to clarify its actual usage. 

Another possible fix would be to change the GL backend back to reporting DiscreteGpu instead of Other, or add actual detection support for DiscreteGpu (ex: maybe consider all devices from amd, nvidia or intel not detected as one of the other types as DiscreteGpu?).

**Testing**

My system, which outputs the adapters now listed in the updated readme, now selects the Vulkan version like it did back in wgpu 0.12.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ Bottom level categories:
 ## Unreleased
 
 ### Bug Fixes
+- Prefer `DeviceType::DiscreteGpu` over `DeviceType::Other` for `PowerPreference::LowPower` so Vulkan is preferred over OpenGL again by @Craig-Macomber in [#2853](https://github.com/gfx-rs/wgpu/pull/2853)
 
 #### DX12
 - `DownlevelCapabilities::default()` now returns the `ANISOTROPIC_FILTERING` flag set to true so DX12 lists `ANISOTROPIC_FILTERING` as true again by @cwfitzgerald in [#2851](https://github.com/gfx-rs/wgpu/pull/2851)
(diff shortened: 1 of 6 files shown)


cwfitzgerald (approved):
Good stuff

Issue #2726, opened 2022-06-03
WebGL: format `Depth24PlusStencil8` can't be multisampled


maxammann:
**Description**
I recently updated from the wgpu-hal 0.12.5 to the latest master (as of yesterday). As a result my demo failed for WebGL: https://maplibre-rs-demos.pages.dev/webgl/

```
panicked at 'wgpu error: Validation Error

Caused by:
    In Device::create_render_pipeline
    depth/stencil state is invalid
    format Depth24PlusStencil8 can't be multisampled

', /home/runner/.cargo/git/checkouts/wgpu-53e70f8674b08dd4/9e3cd08/wgpu/src/backend/direct.rs:2286:5
```

**Repro steps**
Run the demo of: https://github.com/maplibre/maplibre-rs

**Expected vs observed behavior**
I think this multisampling should be allowed.

**Extra materials**
-
**Platform**
Firefox

maxammann:
Got he same error when running headlessly on egl:

```
error: XDG_RUNTIME_DIR not set in the environment.
thread 'main' panicked at 'wgpu error: Validation Error
Caused by:
    In Device::create_render_pipeline
    depth/stencil state is invalid
    format Depth24PlusStencil8 can't be multisampled
', /home/runner/.cargo/git/checkouts/wgpu-53e70f8674b08dd4/9e3cd08/wgpu/src/backend/direct.rs:2286:5
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: wgpu::backend::direct::default_error_handler
   3: core::ops::function::Fn::call
   4: wgpu::backend::direct::ErrorSinkRaw::handle_error
   5: <wgpu::backend::direct::Context as wgpu::Context>::device_create_render_pipeline
   6: maplibre::render::resource::pipeline::RenderPipelineDescriptor::initialize
   7: <maplibre::render::stages::PrepareStage as maplibre::schedule::Stage>::run
   8: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   9: criterion::bencher::AsyncBencher<A,M>::iter
  10: criterion::benchmark_group::BenchmarkGroup<M>::bench_function::{{closure}}
  11: <criterion::routine::Function<M,F,T> as criterion::routine::Routine<M,T>>::warm_up
  12: criterion::analysis::common
  13: render::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

PR #2854, opened 2022-07-05
gl: fix depth stencil texture format capability

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Closes https://github.com/gfx-rs/wgpu/issues/2726

**Testing**
Tested on locally and Chrome Canary(v105.0.5160.0)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,9 @@ Bottom level categories:
 #### DX12
 - `DownlevelCapabilities::default()` now returns the `ANISOTROPIC_FILTERING` flag set to true so DX12 lists `ANISOTROPIC_FILTERING` as true again by @cwfitzgerald in [#2851](https://github.com/gfx-rs/wgpu/pull/2851)
 
+#### GLES
+- gl: fix depth stencil texture format capability by @jinleili in [#2854](https://github.com/gfx-rs/wgpu/pull/2854)
+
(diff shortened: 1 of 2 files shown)


maxammann:
This fix works for https://github.com/gfx-rs/wgpu/issues/2726

But are the changes actually correct? can we assure that depth buffers support multisampling?

jinleili:
[Google Dawn](https://github.com/gpuweb/gpuweb/issues/1924#issuecomment-880650003) says this capability works on all backends, and [OpenGL Image Format wiki](https://www.khronos.org/opengl/wiki/Image_Format) also says:
> There are three basic kinds of image formats: color, depth, and depth/stencil. Unless otherwise specified, all formats can be used for textures and renderbuffers equally. Also, unless otherwise specified, all formats can be [multisampled](https://www.khronos.org/opengl/wiki/Multisampling "Multisampling")equally.

cwfitzgerald (approved):

jinleili:
Also searched for [OpenGL ES 3.2 Spec](https://www.khronos.org/registry/OpenGL/specs/es/3.2/es_spec_3.2.pdf)
> #### 8.8 Multisample Textures
> In addition to the texture types described in previous sections, two additional types of texture are supported. Multisample textures are similar to two-dimensional or two-dimensional array textures, except that they contain multiple samples per texel.

Issue #2652, opened 2022-05-11
WebGL and wasm multithreading with web workers


expenses:
*So this is something that seems pretty difficult to implement and the best bet is almost definitely to just wait until WebGPU becomes more widely available (I'm using WebXR which doesn't work with WebGPU yet).*

In a couple of places we lie about wasm not having threads:

https://github.com/gfx-rs/wgpu/blob/9be974aa6bff1170e95f6668fe79291a7d59e55c/wgpu-hal/src/gles/device.rs#L1146-L1150

This is not true, as wasm has thread support via web workers, see https://web.dev/webassembly-threads/ https://rustwasm.github.io/docs/wasm-bindgen/examples/raytrace.html https://github.com/wngr/wasm-futures-executor etc.

There are a couple of limitations that make trying to use wgpu in a multthreaded way with the WebGL backend very difficult at the moment:

1. You can't block on the main thread at all, and wgpu has internal Mutexes that try to block when shared across threads. An error for this looks like:
  ```
VM879:1284 RuntimeError: Atomics.wait cannot be called in this context
    at parking_lot::raw_mutex::RawMutex::lock_slow::h6c0cf413016f1347 (webxr_pbr_bg.wasm:0x1e9d17)
    at wgpu_core::device::<impl wgpu_core::hub::Global<G>>::texture_drop::h189d38f7cd4d16b8 (webxr_pbr_bg.wasm:0x1ff48e)
    at <wgpu::Texture as core::ops::drop::Drop>::drop::h80e5c76af149e025 (webxr_pbr_bg.wasm:0x302957)
```

2. You can't access a lot of global properties in web workers. See https://developer.mozilla.org/en-US/docs/Web/API/DedicatedWorkerGlobalScope. This means that you can't run your main function in a web worker.

3. Any `wasm_bindgen` javascript values that are passed across the thread boundary such as the `WebGl2RenderingContext` become `undefined`. 

I did look into creating a more threadsafe device api that involved pushing tasks to a device in the main thread: https://gist.github.com/expenses/64a467b5f09462ad2441b9ceefe55b6f but I was not able to make this work fully.

It's possible that there are good ways around these issues and that it would end up being worthwhile. The idea of having a separate thread for WebGL rendering has been around for a while: 

https://research.mozilla.org/2014/07/22/webgl-in-web-workers-today-and-faster-than-expected/

haraldreingruber-dedalus:
[...]

Issue #1986, opened 2021-09-21
Web backend crashes when initialized within a worker


strax:
**Description**
The web backend [uses `web_sys::window`](https://github.com/gfx-rs/wgpu/blob/master/wgpu/src/backend/web.rs#L954) to obtain a handle to the global `Navigator` object. This does not work when called from a worker scope since there is no `window` global variable nor a `Window` interface.

**Repro steps**
Create a `Worker` that executes a Rust program compiled to WebAssembly calling `wgpu::Instance::new`. Unfortunately I don't have a minimal repro available at the moment.

**Expected vs observed behavior**
Observed behavior: the web backend tries to obtain a handle to the `window` object and performs an `instanceof` check against the `Window` interface, which does not exist in the worker scope. This results in an error:
```
Uncaught (in promise) ReferenceError: Window is not defined
```

Expected behavior: 
In a worker scope, a [`web_sys::WorkerGlobalScope`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.WorkerGlobalScope.html) could be used instead to obtain the `Navigator` handle, but this might be tricky in practice (https://github.com/rustwasm/wasm-bindgen/issues/1046). Since the program will crash and not return `None` when `window` is not available, it is not enough to use `Option::or` to fall back to, say, `web_sys::global()`. Based on https://github.com/rustwasm/wasm-bindgen/issues/1982#issuecomment-579131489, this is a problem with other `web_sys` consumers too and might require some glue code like [this](https://github.com/rustwasm/gloo/blob/f9b1727481125f1f2e3d052137f92d1e6de9af93/crates/timers/src/callback.rs#L19-L38).

**Platform**
* `wgpu` version: master

codeart1st:
Can confirm this for me. Tried an experiment with OffscreenCanvas and run into the same problem.

teoxoy:
I believe this was fixed by https://github.com/gfx-rs/wgpu/pull/2858, closing.

codeart1st:
`self.Window = WorkerGlobalScope` is still needed for me.

https://github.com/codeart1st/wgpu-layers/blob/main/demo/src/worker.js#L9

Issue #1466, opened 2021-06-03
Worker/offscreen canvas example


lwansbrough:
Would be nice if there was a (probably broken) demo that utilized a worker for offscreen canvas rendering, sort of as a target to hit. I imagine the APIs will change, but having the demo there could serve as way for those of us interested to check on the status of that specific aspect of the spec.

cwfitzgerald:
Just as a point of clarity: this is about rendering to an offscreen canvas not render-to-texture?

lwansbrough:
> Just as a point of clarity: this is about rendering to an offscreen canvas not render-to-texture?

Correct, a canvas whose control has been transferred to a web worker.

PR #2858, opened 2022-07-06
Create instance from worker

JolifantoBambla:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
https://github.com/gfx-rs/wgpu/issues/1986
https://github.com/gfx-rs/wgpu/issues/1466
https://github.com/gfx-rs/wgpu/issues/2652

**Description**
On the WebGPU backend, `Context::init()` uses `web_sys::window()` to access `navigator.gpu`.
Since `window` is not available on web workers this function panics when called from a web worker (#1986).
This is solved by introducing a `Global` type the result of `js_sys::global()` is cast to.
This instance of `Global` is then used to check whether or not `window()` is defined to determine if the function was called in a worker context or not. If it's in a worker context, it will use `WorkerGlobalScope` instead of `Window` to access the `gpu` property.
I'm relatively new to rust and took the solution from [here](https://github.com/rustwasm/gloo/blob/2c9e776701ecb90c53e62dec1abd19c2b70e47c7/crates/timers/src/callback.rs#L8-L40) after stumbling upon [this issue in wasm-bindgen](https://github.com/rustwasm/wasm-bindgen/issues/2148), so it's maybe not the prettiest solution.
Let me know if you have a better solution in mind.

**Testing**
I added logging of `adapter.get_info()` to the `hello` example and am about to submit another PR over at https://github.com/gfx-rs/wgpu-rs.github.io that adds an example running the `hello` example in a webworker.
Note though, that this example doesn't work in Firefox since it doesn't support loading modules in workers. I tested it in Chromium on Linux.


JolifantoBambla:
the example I used to test this is here: https://github.com/gfx-rs/wgpu-rs.github.io/pull/24

JolifantoBambla:
Hey, @cwfitzgerald, I fixed the formatting. Could you maybe try running the checks again? Thanks :)

wgpu/examples/hello/main.rs
```diff
@@ -16,6 +16,9 @@ async fn run() {
             .unwrap()
     };
 
+    #[cfg(target_arch = "wasm32")]
+    log::info!("{:?}", adapter.get_info());
```
cwfitzgerald, line 20: I feel like we should either always use log - or explain why we're only sometimes using it.

cwfitzgerald (changes requested):
One comment, but otherwise looks good.

wgpu/examples/hello/main.rs
JolifantoBambla: Good point! I changed it to use log now in all cases. Or is there a reason log wasn't used in the non-wasm case before?

cwfitzgerald (approved):
Nice!

PR #2866, opened 2022-07-10
dx12: Add support for push constants

### LINKED ISSUE #2829: Implement Push Constants for DX12
We currently cannot combine user defined push constants with the existing root constants that naga sends forward. This is partly a wgpu and partly a naga task.

JCapucho:
**Checklist**

- [x] Run `cargo clippy` (shows some warning unrelated to my work).
- [x] Run `RUSTFLAGS=--cfg=web_sys_unstable_apis cargo clippy --target wasm32-unknown-unknown` if applicable (not applicable).

**Connections**
~~Requires https://github.com/gfx-rs/naga/pull/2005~~ (merged)

Closes #2829

**Description**
Adds support for push constants on the dx12 backend by using a slot on the root signature for a constant buffer.

**Testing**
Tested with a very basic cube program, needs more testing

### DIFF
--- a/wgpu-hal/src/dx12/adapter.rs
+++ b/wgpu-hal/src/dx12/adapter.rs
@@ -205,7 +205,8 @@ impl super::Adapter {
             | wgt::Features::WRITE_TIMESTAMP_INSIDE_PASSES
             | wgt::Features::TEXTURE_COMPRESSION_BC
             | wgt::Features::CLEAR_TEXTURE
-            | wgt::Features::TEXTURE_FORMAT_16BIT_NORM;
+            | wgt::Features::TEXTURE_FORMAT_16BIT_NORM
+            | wgt::Features::PUSH_CONSTANTS;
         //TODO: in order to expose this, we need to run a compute shader
         // that extract the necessary statistics out of the D3D12 result.
         // Alternatively, we could allocate a buffer for the query set,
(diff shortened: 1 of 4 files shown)


wgpu-hal/src/dx12/adapter.rs
```diff
@@ -270,7 +271,7 @@ impl super::Adapter {
                         .min(crate::MAX_VERTEX_BUFFERS as u32),
                     max_vertex_attributes: d3d12::D3D12_IA_VERTEX_INPUT_RESOURCE_SLOT_COUNT,
                     max_vertex_buffer_array_stride: d3d12::D3D12_SO_BUFFER_MAX_STRIDE_IN_BYTES,
-                    max_push_constant_size: 0,
+                    max_push_constant_size: 128,
```
cwfitzgerald, line 274: wouldn't this technically be 256 - 12?

wgpu-hal/src/dx12/command.rs
```diff
@@ -784,11 +796,25 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
...
         &mut self,
-        _layout: &super::PipelineLayout,
+        layout: &super::PipelineLayout,
         _stages: wgt::ShaderStages,
-        _offset: u32,
-        _data: &[u32],
+        offset: u32,
+        data: &[u32],
     ) {
+        log::trace!("Set push constants");
```
cwfitzgerald, line 804: I don't think we should log this.

cwfitzgerald (commented):
Small comments, but code looks good.

wgpu-hal/src/dx12/adapter.rs
JCapucho: I'll be honest, I copied this from the `wgpu-hal` code

JCapucho: Actually, I've took some time to read about `dx12` root signatures and it's limits and we can't make the `max_push_constant_size` that big.

That's because the root signature has a limit of `64 DWORDS` (`256 bytes`):
- push constants consume a `DWORD` for each `4 bytes` of data (1:1)
- If a bind group has buffers it will consume a `DWORD` for the descriptor table
- If a bind group has samplers it will consume a `DWORD` for the descriptor table
- Each dynamic buffer will consume `2 DWORDs` for the root descriptor
- The special constants count as constants and occupy `3 DWORDS`

Source: https://learn.microsoft.com/en-us/windows/win32/direct3d12/root-signature-limits#memory-limits-and-costs

Since we can't know beforehand all root signatures that will be created, we should set a more reasonable `max_push_constant_size` and I think that half (`32 DWORDS = 128 bytes`) like `gfx-hal` had is a reasonable value.

cwfitzgerald: That seems fair enough - we should probably document this in the code

cwfitzgerald (approved):
LGTM after added documentation on the push constant size

wgpu-hal/src/dx12/adapter.rs
JCapucho: done

JCapucho:
CI is failing on vulkan tests... on dx12 changes 🙃

cwfitzgerald:
Hmm, seems to be spurious

PR #2867, opened 2022-07-10
fix deduplication not taking into account render bundles

### LINKED ISSUE #2865: Bind group / pipeline deduplication does not take into account RenderBundles
**Description**
#2623 introduced not pushing a pipeline/bind group set command if it was already set
Render bundles reset the pipeline and bind groups after being executed though, and this isn't taken into account by this check

**Repro steps**
Edited the hello-triangle example: 
https://github.com/shoebe/wgpu/tree/115ee0fa8cc8b03fd61fc015827afc30f84821ca/wgpu/examples/deduplication-bug
In short:
```
render_pass.set_pipeline(&pipeline); // 1
render_pass.execute_bundles(bundles);
render_pass.set_pipeline(&pipeline); // This pipeline will not be set because it was set before the bundle execution
render_pass.draw(0..3, 0..1);
drop(render_pass);
self.queue.submit(std::iter::once(encoder.finish())); // crash here: 'render pipeline must be set'
// if line 1 is commented out, no crash
```

**Extra materials**
Crash log:
```
[2022-07-10T17:30:11Z ERROR wgpu::backend::direct] Handling wgpu errors as fatal by default

Caused by:
    In a RenderPass
    In a draw command, indexed:false indirect:false
    render pipeline must be set

', wgpu/src/backend/direct.rs:2391:5
```
**Possible fix**
See https://github.com/shoebe/wgpu/commit/54c330aa275ffed712a2dee6bf567c733f151886

shoebe:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
Fixes #2865


cwfitzgerald (approved):

cwfitzgerald:
Thanks!

PR #2868, opened 2022-07-10
Fix instruction to run example test in README.

PrototypeNM1:
**Description**
The provided command to run example tests in the README returns `error: Found argument '--example' which wasn't expected, or isn't valid in this context`. This corrects the documented command.

**Testing**
Run prior and new commands (e.g. `cargo nextest run --example water --no-fail-fast` and `cargo nextest run water --no-fail-fast`), confirm the first fails to run and the latter succeeds.


cwfitzgerald (approved):
Thanks!

PR #2877, opened 2022-07-13
Record that the buffer is mapped when its size is zero.

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Description**

This makes buffers behave properly when the user maps and unmaps an empty buffer range. Without this PR, 'unmap' will return an error saying that the buffer is not mapped even though the mapping closure reproted success. 

It's a bit of a ridiculous test cases but fuzzers love this type of input.

The `NonNull::dangling` pointer might be intimidating, thankfully it's used in very few places so it's easy to check that we don't do dangerous things things with it. It can be summed up to the following snipped which I ran under miri  without complaints of undefined behavior:

```rust
    #[test]
    fn it_works() {
        unsafe {
            let nonnul_ptr: std::ptr::NonNull<u8> = std::ptr::NonNull::dangling();
            let actual_ptr = nonnul_ptr.as_ptr();
            let offset = actual_ptr.offset(0);
            let slice = std::slice::from_raw_parts(actual_ptr, 0);
            println!("ok!");
        }
    }
```

**Testing**

No automated tests.


cwfitzgerald (approved):
Ahh edge cases. Thank you!

cwfitzgerald:
Format is mad

nical:
> Format is mad

Oops! Fixed.

Issue #2316, opened 2021-12-25
Labels, debug markers, and validation layers should be available on release builds


aloucks:
**Is your feature request related to a problem? Please describe.**
I was trying to debug something in Renderdoc and none of the labels or debug groups were showing up. I switched to a debug build and my labels and groups started showing up. The issue is that debug builds can be painfully slow in some cases. I often just use release builds.

**Describe the solution you'd like**
I'd like to be able to set two flags during instance creation. One to enable debug markers, groups, and labels and a separate flag to enable validation. It looks like these already exist as flags on the `hal` Instance:

https://github.com/gfx-rs/wgpu/blob/61fa8a3b631e5f07509bd7387e2c0c841649bfba/wgpu-core/src/instance.rs#L107-L110

**Describe alternatives you've considered**
1. I could enable `debug_assertions` in release builds or change debug builds to be more optimized. I'd still prefer a runtime flag during instance creation.
2. I could use the `hal` API to create the instance.

**Additional context**
The labels and debug groups have a cost associated with the string allocation but I think there's opportunity for improvement here by using string interning or other options to minimize the allocation.

PR #2881, opened 2022-07-13
Add WGPU_BACKEND_VALIDATION environment variable

cwfitzgerald:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Part of #2316 

**Description**

This adds an env var which turns on backend validation no matter if we're in debug or release. This is useful for both forcing it on in release and forcing it off in debug.

**Testing**
None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,12 @@ Bottom level categories:
 
 ## Unreleased
 
+### Added/New Features
+
+#### General
+
+- Add `WGPU_BACKEND_VALIDATION` env var to control validation by @cwfitzgerald in [#2881](https://github.com/gfx-rs/wgpu/pull/2881)
+
 ### Bug Fixes
 
 #### General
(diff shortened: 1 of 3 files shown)


jimblandy (commented):
Is an environment variable really how people want to work around this?

Would it make more sense to check the environment variable from something analogous to `initialize_adapter_from_env`, so that the app gets to opt into sensitivity to the environment? Usually we haven't hacked this kind of thing directly into Instance::new. It almost seems like this should be some sort of blocklist in the back end, rather than a magic var users need to know about to make things work for them.

Wearing my Mozilla hat, Firefox doesn't object, but we do have a general preference for less environmental sensitivity, because it makes problems that much harder to reproduce.

For all the above reasons, something more targeted would be better, like this one, which applies only to a specific back end, to address a specific bug: https://github.com/gfx-rs/wgpu/blob/6058676dd20de9f4fc37dfccf09e9e14524bcfbb/wgpu-hal/src/dx11/instance.rs#L5

PR #2887, opened 2022-07-15
Add warning when using CompareFunction::*Equal without an invariant Attribute

cwfitzgerald:
**Checklist**

- [x] Run `cargo clippy`.

**Description**

This is a common issue that only actually manifests on specific sets of hardware (apple based gpus) and I know of no cases where this will be a false positive, so having a global warning for this is useful.

**Testing**

Tested by changing an example to use Equal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,12 +40,31 @@ Bottom level categories:
 
 ## Unreleased
 
+### Major Changes
+
+#### @invariant Warning
+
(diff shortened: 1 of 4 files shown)


nical:
Nice! would it make sense for the warning to be only emitted in debug builds?

cwfitzgerald:
Yeah I'm generally unconvinced that wgpu itself should be making any decisions (besides literal debug assertions) based off of `debug_assertions` - we should let the user decide what is best for them. In this case logging frameworks let you filter based on module as you please, so we should be okay here.

PR #2888, opened 2022-07-15
Add WGSL examples to documentation 

### LINKED ISSUE #2849: Use WGSL instead of GLSL in library documentation
There are a number of items in `wgpu` whose documentation contain examples using GLSL syntax or other references to GLSL elements. Since WGSL is now the standard shading language for WebGPU, it would be beneficial to readers if these examples were presented first in WGSL. (Keeping the GLSL would still be helpful for new users arriving from WebGL.)

Relevant occurrences of the text "GLSL" in documentation as of commit c36eb9f44bd9462fe176fb859839592131428feb (just after 0.13.1):

* `Features::SAMPLED_TEXTURE_AND_STORAGE_BUFFER_ARRAY_NON_UNIFORM_INDEXING`
* `Features::UNIFORM_BUFFER_AND_STORAGE_TEXTURE_ARRAY_NON_UNIFORM_INDEXING`
* `BufferBindingType`
* `BindingType`
* `StorageTextureAccess`
* `TextureSampleType`
* `TextureViewDimension`

Incomplete list (since it's hard to search for) of locations where the documentation refers to GLSL entities without even mentioning GLSL:

* `PipelineLayoutDescriptor` refers to "set = 0" which is GLSL syntax.
* `BindingType::Texture::multisampled` refers to GLSL texture sampling functions.
* `Features::TEXTURE_BINDING_ARRAY` and `BUFFER_BINDING_ARRAY` have GLSL examples.

Also possibly of interest: the following items mention something like “See the OpenGL specification” but are not about shading language per se and are mostly just declining to explain a topic:

* `BlendFactor`
* `BlendState`
* `BindingType::StorageTexture`
* `SamplerBindingType`

norepimorphism:
**Checklist**

- [x] Run `cargo clippy`.
- [x] Add `Features` example(s).
- [x] Add `BufferBindingType` example(s).
- [x] Add `BindingType` example(s).
- [x] Add `StorageTextureAccess` example(s).
- [x] Add `TextureSampleType` example(s).
- [x] Add `TextureViewDimension` example(s).

**Connections**
Partly addresses #2849.

**Description**
Adds WGSL examples to *wgpu* documentation to complement the existing examples written in GLSL.

**Testing**
Run `cargo doc` to observe that the documentation is updated, and compile the WGSL examples (e.g. with *naga*) to confirm they are valid.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -326,15 +326,16 @@ bitflags::bitflags! {
         const MAPPABLE_PRIMARY_BUFFERS = 1 << 16;
         /// Allows the user to create uniform arrays of textures in shaders:
         ///
-        /// ex. 
+        /// ex.  
         /// `var textures: binding_array<texture_2d<f32>, 10>` (WGSL)  
         /// `uniform texture2D textures[10]` (GLSL)
         ///
         /// If [`Features::STORAGE_RESOURCE_BINDING_ARRAY`] is supported as well as this, the user
         /// may also create uniform arrays of storage textures.
         ///
-        /// eg. `var textures: array<texture_storage_2d<f32, write>, 10>` (WGSL),
-        /// `uniform image2D textures[10]` (GLSL).
+        /// ex.  
+        /// `var textures: array<texture_storage_2d<f32, write>, 10>` (WGSL)  
+        /// `uniform image2D textures[10]` (GLSL)
         ///
         /// This capability allows them to exist and to be indexed by dynamically uniform
         /// values.
@@ -348,8 +349,9 @@ bitflags::bitflags! {
         const TEXTURE_BINDING_ARRAY = 1 << 17;
         /// Allows the user to create arrays of buffers in shaders:
         ///
-        /// eg. `var<uniform> buffer_array: array<MyBuffer, 10>` (WGSL),
-        /// `uniform myBuffer { ... } buffer_array[10]` (GLSL).
+        /// ex.  
+        /// `var<uniform> buffer_array: array<MyBuffer, 10>` (WGSL)  
+        /// `uniform myBuffer { ... } buffer_array[10]` (GLSL)
         ///
         /// This capability allows them to exist and to be indexed by dynamically uniform
         /// values.
@@ -357,8 +359,9 @@ bitflags::bitflags! {
         /// If [`Features::STORAGE_RESOURCE_BINDING_ARRAY`] is supported as well as this, the user
         /// may also create arrays of storage buffers.
         ///
-        /// eg. `var<storage> buffer_array: array<MyBuffer, 10>` (WGSL),
-        /// `buffer myBuffer { ... } buffer_array[10]` (GLSL).
+        /// ex.  
+        /// `var<storage> buffer_array: array<MyBuffer, 10>` (WGSL)  
+        /// `buffer myBuffer { ... } buffer_array[10]` (GLSL)
         ///
         /// Supported platforms:
         /// - DX12
@@ -381,7 +384,7 @@ bitflags::bitflags! {
         const STORAGE_RESOURCE_BINDING_ARRAY = 1 << 19;
         /// Allows shaders to index sampled texture and storage buffer resource arrays with dynamically non-uniform values:
         ///
-        /// eg. `texture_array[vertex_data]`
+        /// ex. `texture_array[vertex_data]`
         ///
         /// In order to use this capability, the corresponding GLSL extension must be enabled like so:
         ///
@@ -389,11 +392,11 @@ bitflags::bitflags! {
         ///
         /// and then used either as `nonuniformEXT` qualifier in variable declaration:
         ///
-        /// eg. `layout(location = 0) nonuniformEXT flat in int vertex_data;`
+        /// ex. `layout(location = 0) nonuniformEXT flat in int vertex_data;`
         ///
         /// or as `nonuniformEXT` constructor:
         ///
-        /// eg. `texture_array[nonuniformEXT(vertex_data)]`
+        /// ex. `texture_array[nonuniformEXT(vertex_data)]`
         ///
         /// WGSL and HLSL do not need any extension.
         ///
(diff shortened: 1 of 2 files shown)


norepimorphism:
Hm. Comparing the [WGSL](https://gpuweb.github.io/gpuweb/wgsl) and [GLSL](https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.60.html) specs, it seems that WGSL is far more restrictive than GLSL in terms of access modes. For example, according to [this table](https://gpuweb.github.io/gpuweb/wgsl/#address-space), I don't think it's possible to define a read-write or write-only uniform, texture, or sampler. However, I see no such restrictions in [this rather minimal explanation](https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.60.html#memory-qualifiers) of access modes offered by the GLSL spec.

If my understanding is correct, then some documentation items may be necessarily devoid of a WGSL example due to the limitations of the language. In particular, I think this is the case with `StorageTextureAccess::WriteOnly` and `StorageTextureAccess::ReadWrite`.

norepimorphism:
> GLSL _may_ allow you to put the keywords there, but uniform buffers, textures, and samplers are always read-only, there is no way to actually write to them. In those examples, I think it's perfectly fine to do normal wgsl declarations.

Ah. I will remove the explicit `readonly` annotations then because it *should* be read-only for both WGSL and GLSL.

> We offer read-write and read storage texture support as a native extension, so using a wgsl example (using the extension) is totally fine.

Hm. *naga* accepts that? I don't think the WGSL syntax technically permits that, even if it is possible hardware-wise. Should it be noted somewhere that that's not an official WGSL extension (the only one I see is `f16` support) but is supported by *wgpu*/*naga*?

norepimorphism:
As an aside: I think some of my confusion arose from storage textures, as defined by the WGSL spec, being write-only, but their *handles* being read-only. Apparently there is an issue to clarify this already: https://github.com/gpuweb/gpuweb/issues/3047.

cwfitzgerald:
> Should it be noted somewhere that that's not an official WGSL extension (the only one I see is f16 support) but is supported by wgpu/naga?

Wouldn't hurt, though we already do this to an extent by separating "native" and "webgpu" extensions.

wgpu-types/src/lib.rs
```diff
@@ -326,12 +326,14 @@ bitflags::bitflags! {
         const MAPPABLE_PRIMARY_BUFFERS = 1 << 16;
         /// Allows the user to create uniform arrays of textures in shaders:
         ///
-        /// eg. `uniform texture2D textures[10]`.
+        /// eg. `var textures: array<texture_2d<f32>, 10>` (WGPU), `uniform texture2D textures[10]`
+        /// (GLSL).
```
cwfitzgerald, line 330: Four nits (sorry for the amount on this one lol), first lets put these on separate lines, second /s/wgpu/wgsl/, third /s/eg/ex/, fourth /s/array/binding_array/ - all binding arrays need to use the binding_array type

```suggestion
        /// ex. 
        /// `var textures: binding_array<texture_2d<f32>, 10>` (WGSL)
        /// `uniform texture2D textures[10]` (GLSL)
```

wgpu-types/src/lib.rs
```diff
@@ -3706,6 +3710,16 @@ pub struct ImageDataLayout {
 pub enum BufferBindingType {
     /// A buffer for uniform values.
     ///
+    /// Example WGSL syntax:
+    /// ```rust,ignore
+    /// struct Globals {
+    ///     a_uniform: vec2<f32>,
+    ///     another_uniform: vec2<f32>
```
cwfitzgerald, line 3717:
```suggestion
    ///     another_uniform: vec2<f32>,
```

for consistency

wgpu-types/src/lib.rs
```diff
@@ -3806,23 +3858,49 @@ impl Default for TextureSampleType {
...
     /// layout(set=0, binding=0, r32f) writeonly uniform image2D myStorageImage;
     /// ```
     WriteOnly,
-    /// The texture can only be read in the shader and it must be annotated with `readonly`.
-    /// [`Features::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES`] must be enabled to use this access mode,
+    /// The texture can only be read in the shader and it must be annotated with `read` (WGSL) or
+    /// `readonly` (GLSL).
+    ///
+    /// [`Features::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES`] must be enabled to use this access
+    /// mode. This is a nonstandard, native-only extension.
```
cwfitzgerald, line 3880:
```suggestion
    /// mode. This is a native-only extension.
```

We don't call it non-standard elsewhere, so let's just stick to native-only.

cwfitzgerald (changes requested):
By and large looks great, thank you! Some nits

wgpu-types/src/lib.rs
norepimorphism: > second /s/wgpu/wgsl

Oops. Acronyms are hard, haha.

> third /s/eg/ex

Should I update the other occurrences of 'eg.' in the `Features` documentation? I see eight others.

> fourth /s/array/binding_array

Interesting; good to know.

wgpu-types/src/lib.rs
norepimorphism: Looks good. I think I did that because I wasn't sure if *naga* (and the WGSL spec?) supported trailing commas in struct declarations (I previously ran into issues with that, I think). But I see it in the spec now: https://gpuweb.github.io/gpuweb/wgsl/#syntax-struct_body_decl.

wgpu-types/src/lib.rs
cwfitzgerald: > Should I update the other occurrences of 'eg.' in the Features documentation? I see eight others

Yeah that'd be great!

cwfitzgerald: Oh another subtle detail here, the first example should have two spaces after it so that markdown keeps it the same paragraph, but on different lines.

norepimorphism: Should there be a double-space line break after 'ex.' as well, or no?

cwfitzgerald (approved):
Thanks!

cwfitzgerald:
Yeah that would work - sorry for the delay in responding.

PR #2889, opened 2022-07-15
use smallvec instead of inplace_it

mockersf:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

**Description**
Remove `inplace_it` to use `smallvec` instead. While `inlace_it` tries to be clever with the size of the array used, I just set it to 32 in `smallvec`
On latest nightly, `inplace_it` fails to compile. This will probably be fixed on Rust (https://github.com/rust-lang/rust/pull/99288) or `inplace_it` (https://github.com/NotIntMan/inplace_it/pull/11), but `inplace_it` hasn't been updated in a while.

**Testing**
This is Vulkan only change and I don't have a Vulkan capable machine, so not tested a lot...


notgull:
Could we use `tinyvec` instead and reduce the unsafe code surface?

mockersf:
I went with `smallvec` as it's already in tree

cwfitzgerald (approved):
I'm just going to say this is fine for now - we really should revisit this stuff later if we have a more coherent allocation solution.

PR #2872, opened 2022-07-12
Implement "strict_asserts" feature in wgpu-core.

jimblandy:
Since `wgpu-core`'s public functions are supposed to validate their
parameters, the internal `track` module skips many of Rust's usual
run-time checks in release builds. However, some `wgpu-core` users
are happy to pay the performance cost in exchange for more safety.
The `"tracker_asserts"` feature causes the `track` module to perform
the same checks in release builds as it does in debug builds.

**Checklist**

- [X] Run `cargo clippy`.
- [X] Run `RUSTFLAGS=--cfg=web_sys_unstable_apis cargo clippy --target wasm32-unknown-unknown` if applicable.
- [x] Add change to CHANGELOG.md. See simple instructions inside file.

**Connections**
_Link to the issues addressed by this PR, or dependent PRs in other repositories_

**Description**
_Describe what problem this is solving, and how it's solved._

**Testing**
_Explain how this change is tested._

PR #2890, opened 2022-07-17
[BREAKING] Make `Shader(Module)Source::Wgsl` optional

daxpedda:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Discussed in Matrix.

**Description**
This is totally a breaking change! It also might turn out entirely unnecessary if the compile-time savings aren't worth the added complexity.

At the moment the `naga` `wgsl-in` crate feature is active even when completely unnecessary. This addresses it by gating `Shader(Module)Source::Wgsl` behind a new `wgsl` crate feature, like with `ShaderSource::Glsl` for example. The added `wgsl` crate feature is also active by default.

Code quality is definitely a bit worse with all of these `#[cfg(...)]` now. CI time is increased because `--no-default-features` has to be tested.

I also didn't know what to do about `Shader(Module)Source`s lifetime, at the moment the lifetime is not present when the only active variant is `Naga`, which I consider very user unfriendly and could only be fixed by either stop using `Cow` or adding a dummy lifetime to the `Naga` variant. ~~This could also be solved by #2903, so maybe we should merge that first.~~ (can't bind `'a` to the `'static` in `Cow`)

I also had to increase the MSRV to 1.60 to not unnecessarily pull in `wgpu-core` on WASM by using the weak dependency feature.

**Testing**
Added additional runs with `--no-default-features` in the CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,6 +178,9 @@ jobs:
       - name: check web
         if: matrix.kind == 'web'
         run: |
+          # build with no features
+          cargo ${{matrix.tool}} --target ${{ matrix.target }} -p wgpu --no-default-features
+
           # build examples
           cargo ${{matrix.tool}} --target ${{ matrix.target }} -p wgpu --examples
 
@@ -194,7 +197,7 @@ jobs:
         if: matrix.kind == 'em'
         run: |
           # build for Emscripten/WebGL
-          cargo ${{matrix.tool}} --target ${{ matrix.target }} -p wgpu -p wgpu-hal --features webgl,emscripten
+          cargo ${{matrix.tool}} --target ${{ matrix.target }} -p wgpu -p wgpu-hal --no-default-features --features webgl,emscripten
 
           # build raw-gles example
           cargo ${{matrix.tool}} --target ${{ matrix.target }} --example raw-gles --features webgl,emscripten
@@ -203,7 +206,7 @@ jobs:
         if: matrix.kind == 'local' || matrix.kind == 'other'
         run: |
           # check with no features
-          cargo ${{matrix.tool}} --target ${{ matrix.target }} -p wgpu -p wgpu-core -p wgpu-info -p player
+          cargo ${{matrix.tool}} --target ${{ matrix.target }} -p wgpu -p wgpu-core -p wgpu-info -p player --no-default-features
 
           # check with all features
           # explicitly don't mention wgpu-hal so that --all-features don't apply to it
--- a/wgpu/Cargo.toml
+++ b/wgpu/Cargo.toml
@@ -76,9 +76,10 @@ name = "water"
 test = true
 
 [features]
-default = []
+default = ["wgsl"]
 spirv = ["naga/spv-in"]
 glsl = ["naga/glsl-in"]
+wgsl = ["wgc?/wgsl"]
 trace = ["serde", "wgc/trace"]
 replay = ["serde", "wgc/replay"]
 angle = ["wgc/angle"]
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -858,11 +858,17 @@ pub enum ShaderSource<'a> {
         defines: naga::FastHashMap<String, String>,
     },
     /// WGSL module as a string slice.
+    #[cfg(feature = "wgsl")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "wgsl")))]
     Wgsl(Cow<'a, str>),
     /// Naga module.
     #[cfg(feature = "naga")]
     #[cfg_attr(docsrs, doc(cfg(feature = "naga")))]
     Naga(Cow<'static, naga::Module>),
+    /// Dummy variant because `Naga` doesn't have a lifetime and without enough active features it
+    /// could be the last one active.
+    #[doc(hidden)]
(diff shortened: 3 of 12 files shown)


daxpedda:
Okay, CI just has issues installing Rust, I will try again later.

.github/workflows/ci.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 env:
   RUST_BACKTRACE: 1
-  RUST_VERSION: 1.59
+  RUST_VERSION: 1.60
```
rajveermalviya, line 11:
```suggestion
  RUST_VERSION: '1.60'
```
ci failure is because of this, just yaml being yaml

daxpedda: Wow, thanks.

daxpedda:
Rebased after #2872.

wgpu/src/lib.rs
```diff
@@ -816,7 +816,7 @@ impl Drop for ShaderModule {
 /// Any necessary shader translation (e.g. from WGSL to SPIR-V or vice versa)
 /// will be done internally by wgpu.
 #[non_exhaustive]
-pub enum ShaderSource<'a> {
+pub enum ShaderSource<#[cfg(any(feature = "spirv", feature = "glsl", feature = "wgsl"))] 'a> {
```
cwfitzgerald, line 819: Instead of this - let's have a dummy variant that is only enabled when all features are off. Right now - this technically means that adding features can cause code to break, which differs from the advice that features should only be additive

wgpu/Cargo.toml
```diff
@@ -76,9 +76,10 @@ name = "water"
 test = true
 
 [features]
-default = []
+default = ["wgsl"]
 spirv = ["naga/spv-in"]
```
cwfitzgerald, line 80: I'm skeptical of default features here - imo default features are impossible to turn off once you have multiple crates in your tree that depend on the crate - as someone at some point has forgotten to turn them off. This is something we should chat about though, as I'm sure some people have opinions on it.

cwfitzgerald (changes requested):

wgpu/Cargo.toml
daxpedda: I agree with you, I only added it to keep the behavior the same and any breaking changes to a minimum, will remove it.

grovesNL: I think it would be confusing to require a feature to use WGSL because it's the shading language for WebGPU, and almost all crates will enable this feature anyway.

wgpu/src/lib.rs
daxpedda: Good idea, I decided to just always have the `Dummy` variant present and just slap a `#[doc(hidden)]` on it. I hope that's okay!

wgpu/Cargo.toml
daxpedda: I removed it for now, happy to change it to whatever you guys end up deciding on. Also want to note that if we decide removing this as a default crate feature, we might want to add more tests with different combinations in CI?

daxpedda:
Rebased again.
Too bad it didn't make the cut for 0.14. @cwfitzgerald can we get this in before it gets forgotten again?

wgpu/Cargo.toml
cwfitzgerald: I think I agree with groves here on second thought. Let's keep wgsl as a default feature - it's going to be a footgun if you can't use wgpu without any features.

wgpu/src/lib.rs
```diff
@@ -858,11 +858,17 @@ pub enum ShaderSource<'a> {
...
+    #[cfg(feature = "wgsl")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "wgsl")))]
     Wgsl(Cow<'a, str>),
     /// Naga module.
     #[cfg(feature = "naga")]
     #[cfg_attr(docsrs, doc(cfg(feature = "naga")))]
     Naga(Cow<'static, naga::Module>),
+    /// Dummy variant because `Naga` doesn't have a lifetime and without enough active features it
+    /// could be the last one active.
+    #[doc(hidden)]
```
cwfitzgerald, line 870: Any reason this can't be conditional?

cwfitzgerald (changes requested):
Ack! So sorry about this! I didn't realize the "changes requested" text doesn't get reset when the changes get addressed.

Small comments but then g2g

wgpu/src/lib.rs
daxpedda: It could, but it didn't seem to me to give any advantage, I would have to put this conditional on every `match` statement too now. Should I still add the conditional?

wgpu/Cargo.toml
daxpedda: Done.

daxpedda: I just noticed, should I add the `wgsl` feature as a default to `wgpu-core` too?

wgpu/src/lib.rs
cwfitzgerald: Oh, true - eh whatever

wgpu/Cargo.toml
cwfitzgerald: No, that's fine - wgpu-core is essentially an implementation detail.

daxpedda:
Done.

PR #2892, opened 2022-07-18
Validate against the maximum binding index

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

WebGPU spec: https://gpuweb.github.io/gpuweb/#dom-gpudevice-createbindgrouplayout

**Description**

This PR validates that binding indices in `create_bind_group_layout` aren't larger than the maximum binding index from teh WebGPU specification, so it's another baby step towards WebGPU compliance.

This also sort of mitigates driver issues with very large binding indices on vulkan. The vulkan spec is made with the expectation that the driver allocates a contiguous buffer of descriptors at least as large as the maximum binding index. Binding indices are expected to be small, and some drivers don't expect that allocation to fail (which is fairly easy to trigger by repeatedly creating bind group layouts with enormous binding indices.

Note that Dawn internally renames all binding indices to compact them, including in their vulkan backend. That's a much stronger mitigation against that source of driver issues.

**Testing**

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,7 @@ the same every time it is rendered, we now warn if it is missing.
 - Add some validation in map_async by @nical in [#2876](https://github.com/gfx-rs/wgpu/pull/2876)
 - Fix bugs when mapping/unmapping zero-sized buffers and ranges by @nical in [#2877](https://github.com/gfx-rs/wgpu/pull/2877)
 - Validate the number of color attachments in `create_render_pipeline` by @nical in [#2913](https://github.com/gfx-rs/wgpu/pull/2913)
+- Validate against the maximum binding index in `create_bind_group_layout` by @nical in [#2892]
 
 #### DX12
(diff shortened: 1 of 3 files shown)


wgpu-hal/src/vulkan/device.rs
```diff
@@ -742,6 +742,15 @@ impl crate::Device<super::Api> for super::Device {
         &self,
         desc: &crate::BufferDescriptor,
     ) -> Result<super::Buffer, crate::DeviceError> {
+        if self
+            .shared
+            .workarounds
+            .contains(super::Workarounds::BUFFER_EXTENTS_I32)
```
cwfitzgerald, line 748: Do we already advertise this in the max buffer binding size?

cwfitzgerald (approved):
LGTM, just the one question

wgpu-hal/src/vulkan/device.rs
nical: Hm, that commit wasn't supposed to show up in ths PR.

nical:
Ok removed the unwanted commit. Sorry about that!

PR #2903, opened 2022-07-21
[BREAKING] Wrap `ShaderSource::Naga` in `Cow<'static>`

daxpedda:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Depends on https://github.com/gfx-rs/naga/pull/2013.

**Description**
I'm currently pre-compiling shaders and storing them as `const`, but `ShaderSource::Naga` wants to own `naga::Module`s, as far as I understand, for good reasons, because it stores them, so not owning them would be problematic. Using `Cow<'static, Module>`, solves this, even if it looks strange.

**Testing**
None.

### DIFF
--- a/wgpu/src/backend/direct.rs
+++ b/wgpu/src/backend/direct.rs
@@ -1097,7 +1097,7 @@ impl crate::Context for Context {
                 };
                 let parser = naga::front::spv::Parser::new(spv.iter().cloned(), &options);
                 let module = parser.parse().unwrap();
-                wgc::pipeline::ShaderModuleSource::Naga(module)
+                wgc::pipeline::ShaderModuleSource::Naga(std::borrow::Cow::Owned(module))
             }
             #[cfg(feature = "glsl")]
             ShaderSource::Glsl {
@@ -1113,7 +1113,7 @@ impl crate::Context for Context {
                 let mut parser = naga::front::glsl::Parser::default();
                 let module = parser.parse(&options, shader).unwrap();
 
-                wgc::pipeline::ShaderModuleSource::Naga(module)
+                wgc::pipeline::ShaderModuleSource::Naga(std::borrow::Cow::Owned(module))
             }
             ShaderSource::Wgsl(ref code) => wgc::pipeline::ShaderModuleSource::Wgsl(Borrowed(code)),
             #[cfg(feature = "naga")]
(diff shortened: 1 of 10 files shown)


grovesNL:
Could you provide an example that this change fixes? Is there another way to handle it without changing this type?

daxpedda:
```rust
// I'm hiding some stuff here, but I've done this with a proc-macro
// and using `serde` to deserialize the `Module` in a `Lazy`.
const MODULE: &'static Module = ...;

// With the PR.
let source = ShaderModuleDescriptor {
    label: None,
    source: ShaderSource::Naga(Cow::Borrowed(MODULE)),
};
// Without the PR.
let source = ShaderModuleDescriptor {
    label: None,
    source: ShaderSource::Naga(MODULE.clone()),
};

// With the PR, I can now keep re-using the `Module` without having to clone it unnecessarily.
device.create_shader_module(source);
```
I hope this explains it.

On a side note, this change would also solve the lifetime issue in #2890.

daxpedda:
Probably related: #2801.

daxpedda:
Now that https://github.com/gfx-rs/naga/pull/2013 was merged, I update the relevant parts to depend on it.

daxpedda:
I think CI isn't failing because of this PR, it fails with the same errors in the latest commit on master too.

wgpu/src/backend/direct.rs
```diff
@@ -10,6 +10,8 @@ use crate::{
 use arrayvec::ArrayVec;
 use parking_lot::Mutex;
 use smallvec::SmallVec;
+#[cfg(any(feature = "glsl", feature = "spirv"))]
```
cwfitzgerald, line 13: Lets fully qualify all uses instead of Cow here.

cwfitzgerald (approved):
I'm game, small comment

wgpu/src/backend/direct.rs
daxpedda: Done!

daxpedda:
Added CHANGELOG entry and rebased, not sure why CI fails now, but it doesn't look like it's because of this PR.

daxpedda:
Rebased after #3008.

cwfitzgerald (approved):
Hopefully this doesn't start another cowpocolypse 🐄 😆

dubrowgn:
Sorry to revive this PR, but does the `'static` in `Cow<'static, Module>` mean that we can't use references to runtime generated naga modules? Why couldn't we use `'a` instead, like the other `ShaderSource`'s?

ErichDonGubler:
@dubrowgn: I'm not sure if this is an issue that couldn't have been resolved by specifying a `'static` lifetime in calling code (CC @cwfitzgerald, @daxpedda). If it's interesting to you, could you please investigate and file a PR if you discover that's true?

cwfitzgerald-nominal:
I think the trick here is that we need it to have a 'static lifetime, so either it does already, or if you don't clone it, we will have to. This just exposes that requirement.

PR #2908, opened 2022-07-22
use custom ser/de impl for TextureFormat

crowlKats:


wgpu-types/src/lib.rs
```diff
@@ -2109,6 +2037,467 @@ pub enum TextureFormat {
...
+                };
+
+                let channel = match channel {
+                    AstcChannel::Unorm => "unorm",
+                    AstcChannel::UnormSrgb => "unorm-srgb",
+                    AstcChannel::Hdr => "hdr",
+                };
+
+                let name = format!("astc-{block}-{channel}");
+                serializer.serialize_str(&name)
```
crowlKats, line 2495: sadly `serialize_unit_variant` wants `&'static str`, so had to do this

wgpu-types/src/lib.rs
```diff
@@ -2109,6 +2037,467 @@ pub enum TextureFormat {
...
+                    AstcBlock::B10x10 => "10x10",
+                    AstcBlock::B12x10 => "12x10",
+                    AstcBlock::B12x12 => "12x12",
+                };
+
+                let channel = match channel {
+                    AstcChannel::Unorm => "unorm",
+                    AstcChannel::UnormSrgb => "unorm-srgb",
+                    AstcChannel::Hdr => "hdr",
+                };
```
crowlKats, line 2492: I was going to call `AstcChannel`'s ser/de impl, but unsure if thats even possible

cwfitzgerald:
What's the status of this, I know there was a bunch of discussion about it.

jimblandy (changes requested):
This looks great - thank you for taking care of it!

I think all we need here are some tests. Please add two `#[test]` functions to `wgpu-types/src/lib.rs`, just after the the `impl TextureFormat` block, that compare the serializations and deserializations of `wgpu_types::TextureFormat` against the strings given in the WebGPU spec's definition of [`GPUTextureFormat`](https://www.w3.org/TR/webgpu/#enumdef-gputextureformat).

crowlKats:
ok, will do. will wait for #2954 to land first.
Honestly, i am not too happy about this PR as it feels extremely hacky, but then again, unsure how else this could be achieved

wgpu-types/src/lib.rs
```diff
@@ -2109,6 +2037,467 @@ pub enum TextureFormat {
...
+                            return Err(E::invalid_value(Unexpected::Str(s), &self));
+                        }
+                    }
+                };
+
+                Ok(format)
+            }
+        }
+
+        deserializer.deserialize_any(TextureFormatVisitor)
```
a1phyr, line 2175:
```suggestion
        deserializer.deserialize_str(TextureFormatVisitor)
```

I would rather not use `deserialize_any` here, as it is only supported by serializers for self-describing formats. The only implemented `visit_*` implemented is for `str` anyway, so there is not downsides doing that

wgpu-types/src/lib.rs
```diff
@@ -2109,6 +2037,467 @@ pub enum TextureFormat {
...
+                    "eac-rg11unorm" => TextureFormat::EacRg11Unorm,
+                    "eac-rg11snorm" => TextureFormat::EacRg11Snorm,
+                    other => {
+                        if other.starts_with("astc") {
+                            let mut parts = other.split('-');
+                            parts.next(); // discard "astc"
+                            let block = parts
+                                .next()
+                                .ok_or_else(|| E::invalid_value(Unexpected::Str(s), &self))?;
+                            let channel = parts.collect::<Vec<&str>>().join("-");
```
a1phyr, line 2137:
```suggestion
                        if let Some(parts) = other.strip_prefix("astc-") {
                            let (block, channel) = parts
                                .split_once('-')
                                .ok_or_else(|| E::invalid_value(Unexpected::Str(s), &self))?;
```

I think using `strip_prefix("astc-")` and then `split_once('-')` would be clearer here. (It could be `split` and `Split::as_str` but the latter is unstable :'( )

Additionally, it would avoid two allocations and would discard invalid formats such as "astcXXX-4x4-unorm"

wgpu-types/src/lib.rs
```diff
@@ -2109,6 +2037,467 @@ pub enum TextureFormat {
...
+#[cfg(feature = "serde")]
+impl Serialize for TextureFormat {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        match *self {
+            TextureFormat::R8Unorm => {
+                let name = "r8unorm";
+                serializer.serialize_unit_variant(name, 0, name)
```
a1phyr, line 2188: This seems wrong to me. The `Deserialize` impl expects a string, so the `Serialize` impl should give one. If not, `deserialize(serialize(format))` won't give the same format back but an error.

Morover, serializing everything as a string better matches the spec, is clearer and more resilient to changes to this enum.

wgpu-types/src/lib.rs
```diff
@@ -2109,6 +2037,467 @@ pub enum TextureFormat {
...
+    {
+        use serde::de::{self, Error, Unexpected};
+
+        struct TextureFormatVisitor;
+
+        impl<'de> de::Visitor<'de> for TextureFormatVisitor {
+            type Value = TextureFormat;
+
+            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+                formatter.write_str("Invalid TextureFormat")
```
a1phyr, line 2054: Small nit: https://docs.rs/serde/latest/serde/de/trait.Visitor.html#tymethod.expecting

```suggestion
                formatter.write_str("a valid texture format")
```

a1phyr (changes requested):

wgpu-types/src/lib.rs
crowlKats: @a1phyr this has been addressed now, can you please take a look again?

cwfitzgerald (approved):
Once CI isn't mad, G2G

jimblandy (approved):
Thanks!

crowlKats:
I am unsure whats going on with the errors

jimblandy:
Perhaps try updating your clippy?

crowlKats:
@jimblandy clippy should be fixed now, but i was rather asking about the windows test that is failing

cwfitzgerald:
Has the ron serialization of traces changed? It's a parsing failure of the .ron traces in the test.

a1phyr:
> Has the ron serialization of traces changed?

Yes, texture formats were previously (de)serialized as enum variants and they are now (de)serialized as strings.

a1phyr (approved):
LGTM !

PR #2911, opened 2022-07-24
Fix documentation claiming shader entry points must return void

e00E:
While following https://sotrh.github.io/learn-wgpu/beginner/tutorial3-pipeline/ I was confused that the documentation for the shader entry point states that the function must return void when this was [not the case](https://github.com/sotrh/learn-wgpu/blob/1220dfa70ba9c85f2a3d078914ae811a6a63c5ad/code/beginner/tutorial3-pipeline/src/shader.wgsl) in the article.

As far as I can tell no such restriction exists.

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -1468,8 +1468,8 @@ pub struct VertexBufferLayout<'a> {
 pub struct VertexState<'a> {
     /// The compiled shader module for this stage.
     pub module: &'a ShaderModule,
-    /// The name of the entry point in the compiled shader. There must be a function that returns
-    /// void with this name in the shader.
+    /// The name of the entry point in the compiled shader. There must be a function with this name
+    /// in the shader.
     pub entry_point: &'a str,
     /// The format of any vertex buffers used with this pipeline.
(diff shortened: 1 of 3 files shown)


wgpu/src/lib.rs
```diff
@@ -1545,8 +1545,8 @@ pub struct ComputePipelineDescriptor<'a> {
     pub layout: Option<&'a PipelineLayout>,
     /// The compiled shader module for this stage.
     pub module: &'a ShaderModule,
-    /// The name of the entry point in the compiled shader. There must be a function that returns
-    /// void with this name in the shader.
+    /// The name of the entry point in the compiled shader. There must be a function with this name
+    /// in the shader.
     pub entry_point: &'a str,
```
e00E, line 1550: I am not sure about this change because I did not test a compute shader.

cwfitzgerald: Compute shaders actually do need to return void.

e00E: Fixed.

cwfitzgerald (approved):
Thanks!

PR #2916, opened 2022-07-26
Fix buffer zeroing with offset in map_buffer

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

None.

**Description**

The buffer zeroing code in `map_buffer` currently writes out of bounds if a non zero offset is passed.

This is because the map_buffer code passes the `mapped_pointer + uninitialized.start`  to `write_bytes`, however the uninitialized range is relative to the start of the buffer while our mapped pointer is offset.

The offset parameter has to be subtracted from the uninitialized range so that the new range is relative to the mapped range and avoid (very scary) out of bounds writes.

**Testing**

None.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -192,21 +192,14 @@ fn map_buffer<A: hal::Api>(
     let zero_init_needs_flush_now = mapping.is_coherent && buffer.sync_mapped_writes.is_none(); // No need to flush if it is flushed later anyways.
     for uninitialized_range in buffer.initialization_status.drain(offset..(size + offset)) {
         let num_bytes = uninitialized_range.end - uninitialized_range.start;
-        unsafe {
-            ptr::write_bytes(
-                mapping
-                    .ptr
-                    .as_ptr()
-                    .offset(uninitialized_range.start as isize),
-                0,
-                num_bytes as usize,
-            )
-        };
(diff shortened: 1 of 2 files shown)


jimblandy (approved):
Wow.

wgpu-core/src/device/mod.rs
```diff
@@ -192,21 +192,14 @@ fn map_buffer<A: hal::Api>(
...
-                    .ptr
-                    .as_ptr()
-                    .offset(uninitialized_range.start as isize),
-                0,
-                num_bytes as usize,
-            )
-        };
+        // The mapping's pointer is already offset, however we track the uninitialized range relative to the buffer's start.
+        let start = (uninitialized_range.start - offset) as isize;
+        unsafe { ptr::write_bytes(mapping.ptr.as_ptr().offset(start), 0, num_bytes as usize) };
```
cwfitzgerald, line 197: Could set just remove the unsafe here? I don't think this is giving us meaningful performance boosts.

nical: Good point, I updated the patch. We can't entirely get rid of the unsafe (since we strat from an unsafe pointer), but the part unsafe piece is now pretty trivial and the part that writes the range is now safe and bounds checked. I also checked with godbolt that `slice.fill(0)` desugars into an memset (at least with u8 slices).

cwfitzgerald (approved):
Very nice stuff

PR #2918, opened 2022-07-27
winit 0.27 Update and raw-window-handle 0.5

### LINKED ISSUE #2928: Support for raw-window-handle 0.5
WGPU does not work with the latest winit due to updates in raw-window-handle crate.

WGPU too build with winit 0.27 and raw-window-handle 0.5

wyatt-herkamp:
**Checklist**

- [x] Run `cargo clippy`.

Closes #2928 

**Description**
Winit released 0.27 today. With this decoupling Display and Window Handles https://github.com/rust-windowing/raw-window-handle/issues/92

If I am within the Window Handle match block I get the Display handle via an if let statement returning with an Error if the type does not match. 
**Testing**
Ran the examples on a Windows 11 environment x86 AMD/Nvidia setup

### DIFF
--- a/wgpu-hal/Cargo.toml
+++ b/wgpu-hal/Cargo.toml
@@ -109,7 +109,7 @@ features = ["wgsl-in"]
 
 [dev-dependencies]
 env_logger = "0.9"
-winit = "0.27"     # for "halmark" example
+winit = "~0.27.1"     # for "halmark" example
 
 [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
 glutin = "0.28.0" # for "gles" example
--- a/wgpu-hal/Cargo.toml
+++ b/wgpu-hal/Cargo.toml
@@ -109,7 +109,7 @@ features = ["wgsl-in"]
 
 [dev-dependencies]
 env_logger = "0.9"
-winit = "~0.27.1"     # for "halmark" example
+winit = "0.27.1"     # for "halmark" example
 
 [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
 glutin = "0.28.0" # for "gles" example
--- a/wgpu-hal/Cargo.toml
+++ b/wgpu-hal/Cargo.toml
@@ -8,7 +8,7 @@ homepage = "https://github.com/gfx-rs/wgpu"
 repository = "https://github.com/gfx-rs/wgpu"
 keywords = ["graphics"]
 license = "MIT OR Apache-2.0"
-rust-version = "1.59"
+rust-version = "1.63"
 
 [lib]
 
--- a/wgpu-hal/Cargo.toml
+++ b/wgpu-hal/Cargo.toml
@@ -8,7 +8,7 @@ homepage = "https://github.com/gfx-rs/wgpu"
 repository = "https://github.com/gfx-rs/wgpu"
 keywords = ["graphics"]
 license = "MIT OR Apache-2.0"
-rust-version = "1.63"
+rust-version = "1.59"
 
 [lib]
 
--- a/wgpu-hal/Cargo.toml
+++ b/wgpu-hal/Cargo.toml
@@ -34,7 +34,7 @@ bitflags = "1.0"
 # parking_lot 0.12 switches from `winapi` to `windows`; permit either
 parking_lot = ">=0.11,<0.13"
 profiling = { version = "1", default-features = false }
-raw-window-handle = "0.4"
+raw-window-handle = "0.5"
 thiserror = "1"
 
 # backends common
@@ -109,7 +109,7 @@ features = ["wgsl-in"]
 
 [dev-dependencies]
 env_logger = "0.9"
-winit = "0.26"     # for "halmark" example
+winit = "0.27"     # for "halmark" example
 
 [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
-glutin = "0.28" # for "gles" example
+glutin = "0.28.0" # for "gles" example
--- a/wgpu-hal/src/gles/egl.rs
+++ b/wgpu-hal/src/gles/egl.rs
@@ -773,7 +773,6 @@ impl crate::Instance<super::Api> for Instance {
         has_handle: &(impl HasRawWindowHandle + HasRawDisplayHandle),
     ) -> Result<Surface, crate::InstanceError> {
         use raw_window_handle::RawWindowHandle as Rwh;
-        use raw_window_handle::RawDisplayHandle as Rdh;
 
         let raw_window_handle = has_handle.raw_window_handle();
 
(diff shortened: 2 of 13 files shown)


dhardy:
Tested on Fedora / `5.18.10-100.fc35.x86_64` / RX570: examples work except

- `mipmap` which requires feature `wgpu::Features::WRITE_TIMESTAMP_INSIDE_PASSES` to be enabled
- `conservative-raster` (doesn't work on my GPU)

wyatt-herkamp:
This bug will be fixed in https://github.com/rust-lang/rust/pull/97791

cwfitzgerald:
So this would bump our MSRV to 1.63?

wyatt-herkamp:
> So this would bump our MSRV to 1.63?

Apparently so.

cwfitzgerald:
Could you have this PR bump the MSRV?

emilk:
winit 0.27.1 has been released which lowers the MSRV to `1.57.0`: https://github.com/rust-windowing/winit/blob/master/CHANGELOG.md#0271-2022-07-30

wyatt-herkamp:
> winit 0.27.1 has been released which lowers the MSRV to `1.57.0`: [rust-windowing/winit@`master`/CHANGELOG.md#0271-2022-07-30](https://github.com/rust-windowing/winit/blob/master/CHANGELOG.md#0271-2022-07-30)

Alright all builds were able to run https://github.com/wyatt-herkamp/wgpu/actions/runs/2774303424

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -602,17 +604,29 @@ impl crate::Instance<super::Api> for super::Instance {
...
+                    Ok(self.create_surface_from_xlib(v.display as *mut _, handle.window))
+                } else {
+                    Err(crate::InstanceError)
+                }
             }
             RawWindowHandle::Xcb(handle)
                 if self.shared.extensions.contains(&khr::XcbSurface::name()) =>
             {
-                Ok(self.create_surface_from_xcb(handle.connection, handle.window))
+                if let RawDisplayHandle::Xlib(v) = has_handle.raw_display_handle() {
```
emilk, line 625: should this be `RawDisplayHandle::Xcb(v)` ?

wyatt-herkamp: Yes, it has be corrected

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -591,8 +591,10 @@ impl crate::Instance<super::Api> for super::Instance {
...
     unsafe fn create_surface(
         &self,
-        has_handle: &impl raw_window_handle::HasRawWindowHandle,
+        has_handle: &(impl raw_window_handle::HasRawWindowHandle
+              + raw_window_handle::HasRawDisplayHandle),
     ) -> Result<super::Surface, crate::InstanceError> {
+        use raw_window_handle::RawDisplayHandle;
         use raw_window_handle::RawWindowHandle;
 
         match has_handle.raw_window_handle() {
```
cwfitzgerald, line 600: If we match on `(has_handle.raw_window_handle(), has_handle.raw_display_handle())` we can unpack both halves at once, preventing needing a separate if let inside each branch.

wgpu-hal/Cargo.toml
```diff
@@ -109,7 +109,7 @@ features = ["wgsl-in"]
 
 [dev-dependencies]
 env_logger = "0.9"
-winit = "0.26"     # for "halmark" example
+winit = "~0.27.1"     # for "halmark" example
```
cwfitzgerald, line 112: This doesn't need to be a tilde dep, it can just be `0.27.1`

wgpu/Cargo.toml
```diff
@@ -131,7 +131,7 @@ noise = { version = "0.7", default-features = false }
 obj = "0.10"
 png = "0.17"
 nanorand = { version = "0.7", default-features = false, features = ["wyrand"] }
-winit = "0.26"
+winit = "~0.27.1"     # for "halmark" example
```
cwfitzgerald, line 134: same here

cwfitzgerald (changes requested):
Thanks for the PR, some comments

wgpu-hal/Cargo.toml
wyatt-herkamp: Wouldn't we want to let it support 0.27.1 up till 0.28?

wgpu-hal/src/vulkan/instance.rs
wyatt-herkamp: This is a concept new to me? How would you match on two enums at the same time?

wgpu-hal/Cargo.toml
cwfitzgerald: Yeah, the standard dependency is an implicit carret dependency, so it would be the same as `>=0.27.1, <0.28.0`

wgpu-hal/src/vulkan/instance.rs
cwfitzgerald: Because you're matching on the tuple of the enums, so you can then in each arm do `(Rwh::Xcb(..), Rdh::Xcb(..))` and it will only match that branch if they're both Xcb (now we know they will always be the same, so this isn't a problem).

wyatt-herkamp: That is awesome. Did not know you could match tuples like that. I am updating the code RN

wgpu-hal/src/gles/egl.rs
```diff
@@ -800,24 +800,17 @@ impl crate::Instance<super::Api> for Instance {
                 }
             }
             #[cfg(not(feature = "emscripten"))]
-            Rwh::Wayland(_) => {
+            (Rwh::Wayland(_), raw_window_handle::RawDisplayHandle::Wayland(display_handle)) => {
```
cwfitzgerald, line 803: Let's add a second using for this as well, so we can do Rdh::Wayland(display_handle) and keep it all on one line.

So the using at the top of the function would be:

```rust
use raw_window_handle::{RawWindowHandle as Rwh, RawDisplayHandle as Rdh};
```

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -591,52 +591,52 @@ impl crate::Instance<super::Api> for super::Instance {
...
-                Ok(self.create_surface_from_wayland(handle.display, handle.surface))
+        match (
+            has_handle.raw_window_handle(),
+            has_handle.raw_display_handle(),
+        ) {
+            (RawWindowHandle::Wayland(handle), RawDisplayHandle::Wayland(display)) => {
+                Ok(self.create_surface_from_wayland(display.display, handle.surface))
             }
-            RawWindowHandle::Xlib(handle)
+            (RawWindowHandle::Xlib(handle), RawDisplayHandle::Xlib(display))
```
cwfitzgerald, line 607: One last comment that I missed last time - sorry.

We should give this code the same treatment as the other code with Rdh and Rwh with a sinlge-line use statement.

wyatt-herkamp: Should of listened to my brain. I was thinking that shouldn't the code be matching.

cwfitzgerald (approved):
Thanks so much!

cwfitzgerald:
Enscripten CI is mad

wyatt-herkamp:
> Enscripten CI is mad

It doesn't like importing `RawDisplayHandle as Rdh` because when the one feature is not enabled the import is not used.

emilk:
🥳 

Question: what is the release schedule for wgpu? i.e. when can I expect this to be released?

cwfitzgerald:
Currently we're on a loose 90 day schedule, without making up for lost time so next release will be late September - the calculated date is listed on the milestone for 0.14

Issue #2894, opened 2022-07-19
wgpu-hal and wgpu-core should enable web-sys features when necessary


grovesNL:
**Description**
We expose some `web_sys` types in wgpu-hal and wgpu-core that should require the relevant feature on web-sys (e.g. `OffscreenCanvas` in https://github.com/gfx-rs/wgpu/blob/25b16d5cc8ca84b82574bb1d02d2c51d3ea264c8/wgpu-hal/src/gles/web.rs#L55)

**Repro steps**
Try to build wgpu-native with the wasm32-unknown-unknown target

**Expected vs observed behavior**
Compilation can fail if a crate attempts to use wgpu-hal or wgpu-core without enabling the web-sys features manually, but this shouldn't be required - we should handle it in wgpu-hal and wgpu-core

**Extra materials**
https://github.com/gfx-rs/wgpu-native/issues/207 is related

**Platform**
Latest wgpu

teoxoy:
Was fixed by #2922.

PR #2922, opened 2022-07-30
changed websys features in core and hal to align with used features

Seamooo:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
addresses https://github.com/gfx-rs/wgpu/issues/2894 

**Description**
if using wgpu-core or wgpu-hal as a dependency with target `wasm32-unknown-unknown` without specifying the web-sys feature: `OffscreenCanvas` in the dependencies, both wgpu-core and wgpu-hal would fail to compile (wgpu-hal only when using the gles feature which is effectively implied by the target). The trivial fix to this is adding `OffscreenCanvas` to the web-sys features in wgpu-hal, as wgpu-core also depends on it whenever needing to use web-sys, however, for semantics sake, the feature is in use in wgpu-core as well, so has been added, however behind a more restrictive target as that's the only time web-sys is in use.

**Testing**
inside wgpu/wgpu-core:
- `cargo build --target wasm32-unknown-unknown`
- `cargo build --target wasm32-unknown-emscripten`

inside wgpu/wgpu-hal
- `cargo build --target wasm32-unknown-unknown --features gles`
- `cargo build --target wasm32-unknown-emscripten --features emscripten`

all builds succeed, whereas previously the wasm32-unknown-unknown target builds would fail

### DIFF
--- a/wgpu-core/Cargo.toml
+++ b/wgpu-core/Cargo.toml
@@ -56,9 +56,11 @@ path = "../wgpu-hal"
 package = "wgpu-hal"
 version = "0.13"
 
+[target.'cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))'.dependencies]
+web-sys = { version = "0.3", features = ["HtmlCanvasElement", "OffscreenCanvas"] }
+
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 hal = { path = "../wgpu-hal", package = "wgpu-hal", version = "0.13", features = ["gles"] }
-web-sys = { version = "0.3", features = ["HtmlCanvasElement"] }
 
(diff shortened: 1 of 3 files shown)


wgpu-core/Cargo.toml
```diff
@@ -56,9 +56,11 @@ path = "../wgpu-hal"
 package = "wgpu-hal"
 version = "0.13"
 
+[target.'cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))'.dependencies]
```
cwfitzgerald, line 59: What a mess lol

cwfitzgerald (approved):
Thanks!

PR #2923, opened 2022-07-30
Add `Buffer::size()` and `Buffer::usage()`.

### LINKED ISSUE #2904: Expose readonly size/usage on wgpu::Buffer

Per the draft spec, it appears that GPUBuffer is supposed to expose it's size and usage via readonly properties:
https://gpuweb.github.io/gpuweb/#buffer-interface

As of wgpu-rs 0.13.1, on the rust side neither of these properties are currently visible to application code. Also the wgpu::Buffer type doesn't appear to be caching the size at all.

Keep a cached copy of the size on wgpu::Buffer, and expose member functions to provide readonly access to both the size and usage fields.

The application can always track these attributes externally, since they are immutable after construction, but I'd hate to permanently incur that added burden on application code.

kpreid:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
Fixes #2904

**Description**
Implements `size` and `usage` from https://gpuweb.github.io/gpuweb/#buffer-interface

**Testing**
Added a unit test and ran the test suite. Since this does not interact with the backend at all, there should be no platform-specific testing needed.

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -659,6 +659,7 @@ pub struct Buffer {
     context: Arc<C>,
     id: <C as Context>::BufferId,
     map_context: Mutex<MapContext>,
+    size: wgt::BufferAddress,
     usage: BufferUsages,
 }
 
@@ -2129,6 +2130,7 @@ impl Device {
             context: Arc::clone(&self.context),
             id: Context::device_create_buffer(&*self.context, &self.id, desc),
             map_context: Mutex::new(map_context),
+            size: desc.size,
             usage: desc.usage,
         }
     }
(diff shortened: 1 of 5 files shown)


wgpu/tests/buffer.rs
```diff
@@ -0,0 +1,19 @@
+use crate::common::{initialize_test, TestParameters};
```
cwfitzgerald, line 1: Great use of a test, all I would say is that this should probably be a file for all pieces of attached resource data.

wgpu/src/lib.rs
```diff
@@ -2129,6 +2130,7 @@ impl Device {
             context: Arc::clone(&self.context),
             id: Context::device_create_buffer(&*self.context, &self.id, desc),
             map_context: Mutex::new(map_context),
+            size: desc.size,
```
cwfitzgerald, line 2133: Long term we should think of we should be storing this stuff at the surface, but for now this is fine.

cwfitzgerald (changes requested):

wgpu/tests/buffer.rs
kpreid: I'm sorry, I don't entirely understand what change you're proposing. Do you mean renaming this module to something more general, like `resource.rs`?

cwfitzgerald: Yeah or `resource_descriptor_accessor` or something that captures that vibe.

kpreid: I'm skeptical that this is a good way to organize tests, but okay.

cwfitzgerald (approved):
Tysm!

PR #2930, opened 2022-07-31
Fix opening of renderdoc lib

### LINKED ISSUE #2793: Loading Renderdoc Wrong
**Description**

See #2789 for original issue.

We aren't properly loading renderdoc. We are missing some flags that needs to be passed to dlopen to make sure we never actually load renderdoc's ourselves we only pull the existing copy.

See the documentation here for more info: https://renderdoc.org/docs/in_application_api.html

The important bit:

> To do this you’ll use your platforms dynamic library functions to see if the library is open already - e.g. GetModuleHandle on Windows, or dlopen with the RTLD_NOW | RTLD_NOLOAD flags if available on *nix systems. On most platforms you can just search for the module name - renderdoc.dll on Windows, or librenderdoc.so on Linux, or libVkLayer_GLES_RenderDoc.so on Android should be sufficient here, so you don’t need to know the path to where RenderDoc is running from. 

In libloading terms:

On windows this should call https://docs.rs/libloading/latest/libloading/os/windows/struct.Library.html#method.open_already_loaded which is GetModuleHandle.

On linux this should call https://docs.rs/libloading/latest/libloading/os/unix/struct.Library.html#method.open with `RTLD_NOW | RTLD_NOLOAD` - the latter needs to be added to lib loading, but it's just an integer so we can just specify the value until lib loading adds it.

ABuffSeagull:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
Same issue in ebkalderon/renderdoc-rs#128
Fixes #2793

**Description**
Renderdoc needs to not be opened by us, but instead open the existing copy.
This currently requires OS specific flags for opening and `libloading` doesn't have full API coverage.

**Testing**
The examples didn't work before, now they work fine.
Although I was just able to test Linux only, so anyone who wants to make a quick test with RenderDoc on Windows and Mac would be great.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ the same every time it is rendered, we now warn if it is missing.
 
 #### General
 - Improve the validation and error reporting of buffer mappings by @nical in [#2848](https://github.com/gfx-rs/wgpu/pull/2848)
+- Fixed opening of RenderDoc library by @abuffseagull in [#2930](https://github.com/gfx-rs/wgpu/pull/2930)
 
 ### Changes
 
--- a/wgpu-hal/src/auxil/renderdoc.rs
+++ b/wgpu-hal/src/auxil/renderdoc.rs
@@ -29,6 +29,7 @@ pub enum RenderDoc {
 }
 
 // TODO: replace with libloading API once supported
+#[cfg(unix)]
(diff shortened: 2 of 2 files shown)


cwfitzgerald (approved):
Thanks!

Issue #2925, opened 2022-07-30
Validation Errors With pNext Chain on Vulkan


Rigidity:
**Description**
I get a long error about the pNext chain including structures that can't co-exist.

<details open>
<summary>Here is the complete log.</summary>
<br>
2022-07-30T19:14:40.951440Z  INFO bevy_render::renderer: AdapterInfo { name: "NVIDIA GeForce RTX 3070 Ti", vendor: 4318, device: 9346, device_type: DiscreteGpu, backend: Vulkan }
2022-07-30T19:14:40.954007Z ERROR wgpu_hal::vulkan::instance: VALIDATION [VUID-VkDeviceCreateInfo-pNext-02830 (0x211e533b)]
[...]

PR #2934, opened 2022-08-01
vulkan: fix issues querying multiview support

i509VCB:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
This starts part of #2925 by removing no longer using `VkPhysicalDeviceVulkan11Features`. VkPhysicalDeviceVulkan11Features may imply availability in Vulkan 1.1, but confusingly it is only first available in Vulkan 1.2. Vulkan 1.1 supports chaining the `VkPhysicalDeviceMultiviewFeatures` type, which is aliased by the KHR name to get this information instead. It is cleaner for the implementation to chain specific feature/properties structures instead of switching between the specific and Vulkan1X feature/properties structures.

**Description**
The Vulkan hal did not chain `VkPhysicalDeviceMultiviewFeatures` to the `VkDeviceCreateInfo`. This meant that with drivers that only support Vulkan 1.0 and VK_KHR_multiview would false report no support for the `MULTIVIEW` feature.


cwfitzgerald (approved):
Thanks!

PR #2938, opened 2022-08-03
Map bounds check

### LINKED ISSUE #2935: `wgpu_core::Device::buffer_map_async` doesn't bounds-check `range`
The `wgpu_core::Device::buffer_map_async` function doesn't bounds-check `range`. As far as I can tell, an invalid range will get stashed in `wgpu_core::resource::Buffer::map_state`, yoinked out in `LifetimeTracker::handle_mapping`, and make it all the way to the `wgpu_hal::Device::map_buffer` call in `wgpu_core::device::map_buffer`, which is much farther than it should get.

Granted, `wgpu` doesn't need that validation, since it makes `map_async` a method on `BufferSlice`, whose length is already validated. But I think the checks are pretty cheap compared to everything else that's going on there, and having them in `wgpu_core` would benefit all the different language bindings, and Firefox.

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Fixes #2935.

**Description**

The range is not entirely bounds checked in buffer_map_async. It uses unsigned integers guaranteeing that the range can't underflow the buffer, and the range's end is also checked against the size of the buffer, but a user calling into wgpu_core directly (such as Firefox), could pass a negative range which is invalid and also would allow the start offset to be outside of the bounds of the buffer.

The fix is simply to validate that the range is not negative. Since the end of the range is already validated against the buffer size, it means the buffer start isn't allowed to overflow the buffer.

**Testing**

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,7 +104,8 @@ the same every time it is rendered, we now warn if it is missing.
 - Fix bugs when mapping/unmapping zero-sized buffers and ranges by @nical in [#2877](https://github.com/gfx-rs/wgpu/pull/2877)
 - Fix out-of-bound write in `map_buffer` with non-zero offset by @nical in [#2916](https://github.com/gfx-rs/wgpu/pull/2916)
 - Validate the number of color attachments in `create_render_pipeline` by @nical in [#2913](https://github.com/gfx-rs/wgpu/pull/2913)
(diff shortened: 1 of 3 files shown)


jimblandy (approved):

jimblandy:
The CI failures here are due to the macOS 10.15 deprecation. I've filed #2940 for that.

jimblandy:
... and it would be good to rebase this PR on that once it lands, so we can get a clean run before merging.

PR #2950, opened 2022-08-07
Expose the vk::Queue in the Vulkan hal

anlumo:
**Connections**

Followup to #2667.

**Description**

Allows users of wgpu to access `vk::Queue` using a simple accessor in the Vulkan backend.

### DIFF
--- a/wgpu-hal/src/vulkan/device.rs
+++ b/wgpu-hal/src/vulkan/device.rs
@@ -714,12 +714,20 @@ impl super::Device {
         self.shared.family_index
     }
 
+    pub fn queue_index(&self) -> u32 {
+        self.shared.queue_index
+    }
+
     pub fn raw_device(&self) -> &ash::Device {
         &self.shared.raw
     }
 
     pub fn raw_physical_device(&self) -> ash::vk::PhysicalDevice {
         self.shared.physical_device
+    }
+
+    pub fn raw_queue(&self) -> ash::vk::Queue {
+        self.shared.raw_queue
     }
 
     pub fn enabled_device_extensions(&self) -> &[&'static CStr] {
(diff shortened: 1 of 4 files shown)


i509VCB:
Would it be useful to also expose the vk::Queue on the Vulkan hal's `Queue` type?

anlumo:
> Would it be useful to also expose the vk::Queue on the Vulkan hal's `Queue` type?

I have looked into that, but I couldn't find a way to access the `wgpu_hal::Queue` from the `wgpu::Queue`, so it was kinda pointless.

wgpu-hal/src/vulkan/device.rs
```diff
@@ -719,6 +719,14 @@ impl super::Device {
         self.shared.physical_device
     }
 
+    pub fn raw_queue(&self) -> ash::vk::Queue {
+        unsafe {
+            self.shared
+                .raw
+                .get_device_queue(self.queue_family_index(), self.shared.queue_index)
```
cwfitzgerald, line 726: Since we already have the queue, can we avoid the extra call here in some way?

cwfitzgerald (changes requested):

wgpu-hal/src/vulkan/device.rs
anlumo: I couldn't find a way, but that doesn't mean that there isn't any. The code isn't exactly easy to read and there are no comments and no developer documentation.

i509VCB: If you look at where DeviceShared is initialized, you'll see a raw_queue variable.

Although I still think the index should be exposed as well

anlumo: I added this to the pull request, hope that this is ok. Since `ash::vk::Queue` is `Copy`, this shouldn't be an issue.

i509VCB: Yep that's fine.

cwfitzgerald (approved):
Thanks!

PR #2955, opened 2022-08-11
Fix calculation/validation of layer/mip ranges in create_texture_view

nical:
Back with some more validation fixes.

**Checklist**

- [x] Run `cargo clippy`.

**Connections**

https://bugzilla.mozilla.org/show_bug.cgi?id=1780057

**Description**

The previous code was doing "selector.layers = base_layer .. count" instead of "base_layer .. base_layer + count" which looked dubious. Same thing for the mip range. It was also not taking the base layer count when in requested_layer_count when the texture was not a texture array which led to invalid ranges and underflow when asking for a non-zero base layer with a (non-array) 2d texture.

The fix consists in:
 - writing the ranges as `base .. base + count`
 - always taking the provided base layer into account in the requested layer count even where it doesn't make sense (since it will ensure validation catches the error).

**Testing**

None


nical:
Added a commit that checks for integer overflow because because the fuzzer told me I had to.

cwfitzgerald (approved):

PR #2954, opened 2022-08-11
Updating TextureFormat matching specs

gents83:
Fixing compilation error - when doing a full clean:

error[E0599]: no variant or associated item named `Depth24unormStencil8` found for enum `GpuTextureFormat` in the current scope
   --> D:\a\INOX\INOX\extern\wgpu\wgpu\src\backend\web.rs:553:52
    |
553 |         TextureFormat::Depth24UnormStencil8 => tf::Depth24unormStencil8,
    |                                                    ^^^^^^^^^^^^^^^^^^^^
    |                                                    |
    |                                                    variant or associated item not found in `GpuTextureFormat`
    |                                                    help: there is a variant with a similar name: `Depth24plusStencil8`

PR #2957, opened 2022-08-12
Shrink the default max binding index to 640 and expose it in Limits.

nical:
Following the corresponding changes happening in WebGPU.

**Checklist**

- [x] Run `cargo clippy`.
- [x] `web-sys` updated

**Connections**

- https://github.com/gpuweb/gpuweb/issues/3279
- https://github.com/gpuweb/gpuweb/pull/3318

**Description**

The max binding index is reduced to 640 to avoid issues with wgpu's vulkan backend, for which binding indices map directly to the indices in vkCreateDescriptorSetLayout.
640 is the new default limit advertised by the spec, but technically non-vukan backends rename all of the bindings so they aren't affected by the same issue and can continue using 65535 (the previous limit) if requested.

This PR also exposes the limit in `Limits`.

We don't have to merge this right away since the change has not made it into the WebGPU spec yet.

**Testing**

None


cwfitzgerald (approved):
G2G once spec is clear and web_sys is ready.

nical:
Updated the PR to align with the terminology that made it into the spec. I will get to the web-sys update soon.

jimblandy:
@nical @cwfitzgerald I think web-sys 0.3.60 should include Nical's update.

nical:
This needs https://github.com/gfx-rs/wgpu/pull/2954

nical:
This should be good to go.

jimblandy (approved):

PR #2967, opened 2022-08-16
v0.8 Fix build failure due to rust-lang/rust PR 99413

steffahn:
v0.8 branch: Fixes build failure on current `beta` (1.64).

PR #2966, opened 2022-08-16
v0.7 Fix build failure due to rust-lang/rust PR 99413

steffahn:
v0.7 branch: Fixes build failure on current `beta` (1.64).

PR #2968, opened 2022-08-16
v0.9 Fix build failure due to rust-lang/rust PR 99413

steffahn:
v0.9 branch: Fixes build failure on current `beta` (1.64).

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -1483,6 +1483,7 @@ impl<B: GfxBackend> Device<B> {
 
         // `BTreeMap` has ordered bindings as keys, which allows us to coalesce
         // the descriptor writes into a single transaction.
+        let mut desc_set; // early declaration so it's dropped after write_map
         let mut write_map = BTreeMap::new();
         let mut used_buffer_ranges = Vec::new();
         for entry in desc.entries.iter() {
@@ -1776,7 +1777,7 @@ impl<B: GfxBackend> Device<B> {
             self.desc_allocator
                 .lock()
                 .allocate(&self.raw, &layout.raw, &layout.desc_count, 1)?;
-        let mut desc_set = desc_sets.pop().unwrap();
+        desc_set = desc_sets.pop().unwrap();
 
         // Set the descriptor set's label for easier debugging.
         if let Some(label) = desc.label.as_ref() {
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -80,7 +80,7 @@ pub fn all_image_stages() -> hal::pso::PipelineStage {
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, Debug, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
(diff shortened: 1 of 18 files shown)


steffahn:
It would make sense to review this together with #2966 and #2967.

I’m unfamiliar with this code, and since there’s some `unsafe` blocks involved, someone needs to confirm that the change in drop order isn’t a soundness issue.

These PRs modify the `wgpu-core` crate, which would need new point releases 0.7.2, 0.8.2, and 0.9.3.

wgpu-core/src/device/mod.rs
```diff
@@ -1483,6 +1483,7 @@ impl<B: GfxBackend> Device<B> {
 
         // `BTreeMap` has ordered bindings as keys, which allows us to coalesce
         // the descriptor writes into a single transaction.
+        let mut desc_set; // early declaration so it's dropped before write_map
```
cwfitzgerald, line 1486: This comment is inverted - it gets dropped after write_map when unwinding.

What was the error in the first place?

cwfitzgerald (changes requested):
Normally I wouldn't do patches for these old crates, but a build failure seems a good enough reason, so once we get these sorted I'll patch them.

wgpu-core/src/device/mod.rs
steffahn: Eh, right, the comment is the wrong way around 😅 

Build error is
```
error[E0597]: `desc_set` does not live long enough
    --> wgpu-core/src/device/mod.rs:1792:26
     |
1792 |                     set: desc_set.raw_mut(),
     |                          ^^^^^^^^^^^^^^^^^^ borrowed value does not live long enough
...
1813 |     }
     |     -
     |     |
     |     `desc_set` dropped here while still borrowed
     |     borrow might be used here, when `write_map` is dropped and runs the `Drop` code for type `BTreeMap`
     |
     = note: values in a scope are dropped in the opposite order they are defined
```

The reason is the PR mentioned in the title, a soundness fix for `BTreeMap` which previously behaved (to dropck) as if no `K` and `V` values would be dropped when a `BTreeMap<K, B>` is dropped.

steffahn:
I fixed the typo/inversion of the comment in the PRs :-)

steffahn:
FYI, I found this breakage in this crater run report: https://crater-reports.s3.amazonaws.com/pr-99413/index.html

_(click on “regressed: dependencies (29 root crates, 322 regressed crates in total)”, then search for “wgpu-core”)_

cwfitzgerald (approved):
Alright, this looks like it was a legit soundness issue before, and shouldn't have any major issues unless I'm totally mis-remembering how gfx-hal works. So I'll merge these and get these out later today. Thanks for the proactiveness!

cwfitzgerald:
@jimblandy why are there extra PRs on here?

jimblandy:
I added them to get it to build with current Rust.

jimblandy:
The remaining CI problems are test runs, not build failures. Since this is an old release, I think we should merge the PR anyway.

m4b:
@cwfitzgerald can we get a crates release for this when you get the chance? i've tested locally and compiles fine for me now on 1.64 + 1.65, thanks!

PR #2972, opened 2022-08-19
Use () instead of PhantomData as IdentityManager's Input type.

jimblandy:
PhantomData suggests that there's some sort of persuasion required for lifetime variance inference or other sorts of arcana, but it doesn't seem to be necessary at all. `()` works just fine.

I'm a little worried about this on [Chesterton's Fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence) grounds, but there's only one way to find out!

**Checklist**

- [X] Run `cargo clippy`.


cwfitzgerald (approved):
I likee it

Issue #2898, opened 2022-07-20
max_sampled_texture_limit on metal is too low 


Darksecond:
**Description**
According to https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf the limit for "Maximum number of textures you can access, per stage, from an argument buffer" is 1M on my device (M1 Pro, so Apple7 in the pdf). However wgpu has a hardcoded limit (on metal) of 16 max_sampled_textures_per_shader_stage.
I would like to make a TextureViewArray binding with more than 16 items.

**Expected vs observed behavior**
I get limits that are lower than possible according to the Apple's specs.

**Platform**
wgpu master on metal on m1 pro

Darksecond:
Actually, since it looks like wgpu and naga do not use argument buffers, the limit is 128 (which is still much higher than the 16 currently enforced).

yzsolt:
Ran into the same issue: I can create a texture array with 256 layers without any issues, but I can't use it because the `BindGroupLayoutEntry` count must be set to texture layer count and then it runs into the aforementioned limit. Reproducable with a tweaked (count > 16) texture-arrays example.

I'm not sure if WGPU is checking the correct limit here... Isn't texture array binding counts as a single texture binding, regardless of how many layers it contains? Otherwise texture arrays wouldn't make much sense.

jinleili:
> Reproducable with a tweaked (count > 16) texture-arrays example.

The default limits used by wgpu examples are `Limits::downlevel_webgl2_defaults()`,   set `limits: adapter.limits()` can use the actual hardware limits.

PR #2978, opened 2022-08-21
metal: fix `max_buffer`  `max_texture` and `max_vertex_buffers` limits

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
https://github.com/gfx-rs/wgpu/issues/2898

**Description**
The modification of the limit values is based on [WebKit/WebGPU/HardwareCapabilities.mm](https://github.com/WebKit/WebKit/blob/71835d21e936177ccbd469196fa52cba44fdf6c9/Source/WebGPU/WebGPU/HardwareCapabilities.mm#L94) and the latest [Metal Feature Set Tables](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf)

**Testing**
Tested on macOS and iOS via [wgpu-on-app](https://github.com/jinleili/wgpu-on-app)

### DIFF
--- a/wgpu-hal/src/metal/adapter.rs
+++ b/wgpu-hal/src/metal/adapter.rs
@@ -616,21 +616,17 @@ impl super::PrivateCapabilities {
             format_bgr10a2_all: Self::supports_any(device, BGR10A2_ALL),
             format_bgr10a2_no_write: !Self::supports_any(device, BGR10A2_ALL),
             max_buffers_per_stage: 31,
-            max_textures_per_stage: if os_is_mac {
-                128 // On macOS, minimun value is 128
-            } else if device.supports_feature_set(MTLFeatureSet::iOS_GPUFamily4_v1) {
+            max_vertex_buffers: 31,
+            max_textures_per_stage: if os_is_mac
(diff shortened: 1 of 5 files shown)


wgpu-hal/src/metal/adapter.rs
```diff
@@ -501,6 +501,19 @@ impl super::PrivateCapabilities {
             MTLReadWriteTextureTier::TierNone
         };
 
+        let (max_buffers_per_stage, max_textures_per_stage, max_samplers_per_stage) = if (os_is_mac
+            && rw_texture_tier == MTLReadWriteTextureTier::Tier2)
+            || (family_check && device.supports_family(MTLGPUFamily::Apple6))
+        {
+            (500000 / 2, 500000 / 2, 1024)
```
cwfitzgerald, line 508: Don't we have the problem of us not using argument buffers?

jinleili: Oh, my mistake, I was only thinking about the maximum value.

cwfitzgerald (approved):

PR #2980, opened 2022-08-23
vk: find the supported wide color gamut color space

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Description**
Prefer `BT2020_LINEAR_EXT` color space, followed by `EXTENDED_SRGB_LINEAR_EXT`,  linear color space maybe easier to implement tone mapping.

Referenced document [Enhance graphics with wide color content](https://developer.android.com/training/wide-color-gamut)

**Testing**
Tested on macOS(--features=vulkan-portability) and Android 12(OnePlus Ace Pro)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,9 @@ the same every time it is rendered, we now warn if it is missing.
 #### Metal
 - Extract the generic code into `get_metal_layer` by @jinleili in [#2826](https://github.com/gfx-rs/wgpu/pull/2826)
 
+#### Vulkan
+- Find the supported wide color gamut color space by @jinleili in [#2980](https://github.com/gfx-rs/wgpu/pull/2980)
+
 ### Documentation
 
 #### General
--- a/wgpu-hal/src/vulkan/device.rs
+++ b/wgpu-hal/src/vulkan/device.rs
@@ -528,10 +528,34 @@ impl super::Device {
             None => vk::SwapchainKHR::null(),
         };
 
-        let color_space = if config.format == wgt::TextureFormat::Rgba16Float {
-            // Enable wide color gamut mode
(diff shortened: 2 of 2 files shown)


kocsis1david:
If you build a renderer on top of wgpu, shouldn't it know what color space to output the image in?  
For example if you output a red of 1.0 in srgb, the same red color could be less than 1.0 in a different color space.

jinleili:
Different formats have different WCG color space support on different devices，wide color gamut means that color values have the ability to go beyond the [0, 1] range.

kocsis1david:
https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkColorSpaceKHR.html  
"The color components of non-linear color space swap chain images must have had the appropriate transfer function applied."

If I understand this sentence correctly, you have to apply the transfer function at the end of the shader (or post processing). But you can only apply it if you know which transfer function wgpu selects.

Before the transfer function is applied, it can go beyond the [0, 1] range, but after you apply it, it should be within the [0, 1] range. Except the extended sRGB color space, where it can go beyound [0, 1] even after the transfer function is applied.

Maybe I misunderstand things, I haven't worked with wide color gamut.

jinleili:
> If I understand this sentence correctly, you have to apply the transfer function at the end of the shader (or post processing). But you can only apply it if you know which transfer function wgpu selects.

transfer function is applied through the Vulkan driver.

> Before the transfer function is applied, it can go beyond the [0, 1] range, but after you apply it, it should be within the [0, 1] range.

The vulkan link you mentioned above has this description: _The color space selected for the swap chain image will not affect the processing of data written into the image by the implementation._

cwfitzgerald:
As mentioned, we can't change the color space out from under the users, we need to expose the available color spaces and have the users choose. Rec2020 uses different definitions of red, green, and blue than sRGB does. The user outputting the same value would result in a different color on different platforms. 

Note that the implementation applies the transfer function to be already applied when it receives the data from the swapchain.

Issue #2979, opened 2022-08-22
Depth clear value not used in web backend


metacean:
**Description**
When using the web backend, the depth clear value is not used - depth is cleared to 0.0, which is likely to break many applications that rely on the depth buffer

**Repro steps**
Run any application that clears the depth buffer to 1.0, and writes to it with a 'Less' or 'LessEqual' compare op

**Expected vs observed behavior**
What i expect to see is a depth buffer initialised to white, and samples passing the test. What i observe is no samples passing the depth test, and an all-black buffer in gpu debug tools such as XCode. 
**Extra materials**
In this section of code: https://github.com/gfx-rs/wgpu/blob/f918ac1ed139d4d34aa4775f8ea10d4e2b8213fb/wgpu/src/backend/web.rs#L2090

if I add 

mapped_depth_stencil_attachment.depth_clear_value(1.0);

then my application displays corrrectly, though obviously the specified clear value should be passed through - i have only been using rust for a couple of days so am not sure how best to patch this issue myself. hopefully the correct solution is clear from this description.
 
**Platform**
OS X, running webgpu code on Chrome + dawn

PR #2984, opened 2022-08-26
Depth clear value for web target

gents83:
When using the web backend, the depth clear value is not used - depth is cleared to 0.0, which is likely to break many applications that rely on the depth buffer

### DIFF
--- a/wgpu/src/backend/web.rs
+++ b/wgpu/src/backend/web.rs
@@ -2066,10 +2066,14 @@ impl crate::Context for Context {
         }
 
         if let Some(dsa) = &desc.depth_stencil_attachment {
+            let mut clear_value = 1.0;
             let (depth_load_op, depth_store_op) = match dsa.depth_ops {
                 Some(ref ops) => {
                     let load_op = match ops.load {
-                        crate::LoadOp::Clear(_) => web_sys::GpuLoadOp::Clear,
+                        crate::LoadOp::Clear(v) => {
+                            clear_value = v;
+                            web_sys::GpuLoadOp::Clear
+                        }
                         crate::LoadOp::Load => web_sys::GpuLoadOp::Load,
                     };
                     (load_op, map_store_op(ops.store))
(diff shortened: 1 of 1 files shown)


gents83:
#2979

wgpu/src/backend/web.rs
```diff
@@ -2066,10 +2066,14 @@ impl crate::Context for Context {
         }
 
         if let Some(dsa) = &desc.depth_stencil_attachment {
+            let mut clear_value = 1.0;
```
expenses, line 2069: It doesn't matter, but given that the default value is `0.0` we might as well use that.

```suggestion
            let mut clear_value = 0.0;
```

Could even set it to `None` by default and only call `mapped_depth_stencil_attachment.depth_clear_value` if it's `Some`, but I don't think that's necessary.

gents83: Make definitely sense - default is 0.0

stakka:
adding this fixes e.g. the boids example in the web backend - webgpu in chrome canary. the triangles (boids) are being rendered again

expenses:
Oh, we need to remember to do the same with the stencil value as well. Doesn't need to be done here but it might be nice.

wgpu/src/backend/web.rs
```diff
@@ -2066,10 +2066,14 @@ impl crate::Context for Context {
         }
 
         if let Some(dsa) = &desc.depth_stencil_attachment {
+            let mut clear_value = 0.;
```
cwfitzgerald, line 2069:
```suggestion
            let mut clear_value = 0.0;
```

wgpu/src/backend/web.rs
```diff
@@ -2088,6 +2092,7 @@ impl crate::Context for Context {
             };
             let mut mapped_depth_stencil_attachment =
                 web_sys::GpuRenderPassDepthStencilAttachment::new(&dsa.view.id.0);
+            mapped_depth_stencil_attachment.depth_clear_value(clear_value);
```
cwfitzgerald, line 2095: We really should do stencil while we're at it

cwfitzgerald (changes requested):

wgpu/src/backend/web.rs
gents83: Will do

cwfitzgerald (approved):
Nice!

gents83:
Can be merged even if  **Check Linux x86_64** is failing?
(It seems to be broken on all PRs actually, no?)

cwfitzgerald:
Yeah, we're holding all PRs until that's fixed - it's org-wide highest priority rn

jimblandy:
Rebased on master, since those CI failures should be fixed now.

PR #2986, opened 2022-08-26
Implement `device_features` and `adapter_features` for the web backend

expenses:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

**Description**

These weren't implemented and I needed in order to be able to check for compressed texture support. I had to do an unchecked cast of the [GPUSupportedFeatures](https://gpuweb.github.io/gpuweb/#gpusupportedfeatures) class to a `js_sys::Set` because it doesn't seem like wasm-bindgen implements `setlike` and thus doesn't have a `has` method.

I also added some BC formats to the texture format mapping function because my code needed them. We of-course should implement all of these in time, but I didn't want to do that here.

**Testing**

Tested by running a program that checks for compressed texture support and renders a model with compressed textures.


expenses:
>I had to do an unchecked cast of the [GPUSupportedFeatures](https://gpuweb.github.io/gpuweb/#gpusupportedfeatures) class to a `js_sys::Set` because it doesn't seem like wasm-bindgen implements `setlike`

We should probably find a solution to this before merging.

jimblandy:
Just marking this as "Draft", based on comments and title.

wgpu/src/backend/web.rs
```diff
@@ -1263,9 +1278,22 @@ impl crate::Context for Context {
         // Can't really discard this on the Web
     }
 
-    fn device_features(&self, _device: &Self::DeviceId) -> wgt::Features {
-        // TODO
-        wgt::Features::empty()
+    fn device_features(&self, device: &Self::DeviceId) -> wgt::Features {
+        let features = device.0.features();
+
+        let features_set: js_sys::Set = features.unchecked_into();
```
cwfitzgerald, line 1284: Is a runtime checked into a thing? That would solve the problem and not block us on web-sys

expenses: Yep, https://docs.rs/wasm-bindgen/0.2.83/wasm_bindgen/trait.JsCast.html#method.dyn_into should work.

cwfitzgerald: Awesome!

expenses:
Think this is about ready - I had some problems running the demos on chrome though so it'd be great if someone could check them before merging.

wgpu/src/backend/web.rs
```diff
@@ -16,6 +16,7 @@ use wasm_bindgen::{
     prelude::*,
     JsCast,
 };
+use web_sys::GpuFeatureName as Gfn;
```
cwfitzgerald, line 19: We shouldn't rename in a global import like this

cwfitzgerald (approved):
Looks good after nit

gents83:
Trying to implement it directly in wasm-bindgen: https://github.com/rustwasm/wasm-bindgen/pull/3268

cwfitzgerald:
@gents83 thank you so much for that work!

PR #3006, opened 2022-09-01
Document that `write_buffer_with` is sound to read from.

kpreid:
**Checklist**

- [X] Run `cargo clippy`.

**Description**

To a reader informed about Rust's memory model, the existing claim that

> dereferencing it to a `&[u8]` panics

without further context sounds awfully like someone thinks they've invented a write-only reference, and that the API might actually be exposing undefined behavior via an uninitialized `&mut [u8]`. Therefore, let's specify what happens if you *do* read through the mutable reference.  The text added in this commit is based on what was said in [the review when `write_buffer_with` was added](https://github.com/gfx-rs/wgpu/pull/2777/files#r901392551).

This is also relevant information to someone considering using `write_buffer_with()` for performance gains: for example, it suggests that it might be a bad idea to write data into the view and then sort it in-place. (Or is that not a bad idea? Is it not slow if the CPU already wrote over all the memory contiguously? I don't know.)

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -3468,7 +3468,7 @@ impl Queue {
     /// The returned value can be dereferenced to a `&mut [u8]`; dereferencing it to a
     /// `&[u8]` panics!
     /// (It is not unsound to read through the `&mut [u8]` anyway, but doing so will not
-    /// yield the existing contents of `buffer`, and it is likely to be slow.)
+    /// yield the existing contents of `buffer` from the GPU, and it is likely to be slow.)
     ///
     /// This method is intended to have low performance costs.
     /// As such, the write is not immediately submitted, and instead enqueued
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -3462,9 +3462,13 @@ impl Queue {
         Context::queue_write_buffer(&*self.context, &self.id, &buffer.id, offset, data)
     }
 
-    /// Schedule a data write into `buffer` starting at `offset` via the returned [QueueWriteBufferView].
+    /// Schedule a data write into `buffer` starting at `offset` via the returned
+    /// [QueueWriteBufferView].
     ///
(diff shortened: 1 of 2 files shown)


cwfitzgerald (approved):
Thanks!

cwfitzgerald:
As for:

> it suggests that it might be a bad idea to write data into the view and then sort it in-place

Currently it will be fast, but we reserve the right (and we might do it) to make it terribly slow

kpreid:
> Currently it will be fast, but we reserve the right (and we might do it) to make it terribly slow

Another question that comes to mind is whether this also applies to buffers created as `mapped_at_creation`. If so, that is worth documenting too. In my own thinking about perf optimization, I've thought about doing (as I mentioned in the description) sorting or other read-write operations inside of newly allocated buffers or StagingBelt views.

wgpu/src/lib.rs
```diff
@@ -3464,9 +3464,13 @@ impl Queue {
...
 
-    /// Schedule a data write into `buffer` starting at `offset` via the returned [QueueWriteBufferView].
+    /// Schedule a data write into `buffer` starting at `offset` via the returned
+    /// [QueueWriteBufferView].
     ///
-    /// The returned value can be dereferenced to a `&mut [u8]`; dereferencing it to a `&[u8]` panics!
+    /// The returned value can be dereferenced to a `&mut [u8]`; dereferencing it to a
+    /// `&[u8]` panics!
+    /// (It is not unsound to read through the `&mut [u8]` anyway, but doing so will not
+    /// yield the existing contents of `buffer`, and it is likely to be slow.)
```
a1phyr, line 3473: This is oddly phrased, it feels like you're not guaranteed to read the same value that you've just written, which would definitely be UB.

kpreid: I'm specifying that it _will not_ give the contents of `buffer`, the preexisting GPU buffer, and leaving unspecified what the initial value of this `[u8]` is. I agree this isn't the clearest wording; ideas welcome.

jimblandy: I added a minor qualification to clarify this.

PR #3011, opened 2022-09-05
vulkan: use effective api version for determining device features

i509VCB:
**Checklist**

- [x] Run `cargo clippy`.

**Description**
wgpu wrongly assumes that the instance version will always be the maximum version supported. This assumption can be broken by using Instance::from_raw. This pull request fixes that issue by calculating the "effective" api version the device supports, which is the lower of the instance version and physical device version. More information can be found in the Vulkan specification's documentation for VkPhysicalDeviceProperties.

**Testing**
Tests succeed locally.

Testing could be improved by introducing CI runs in the future which artificially lower the instance version to test the behavior is correct. Possibly via an environment variable such as `WGPU_VULKAN_INSTANCE_VERSION`.


cwfitzgerald (approved):
Slipped through the cracks, sorry!

PR #3022, opened 2022-09-13
pass around RawDisplayHandle and RawWindowHandle in implementation

i509VCB:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
There has been discussion on Matrix to allow swapping the "Context" type inside wgpu at runtime. 

**Description**
This change is to help with an attempt to allow the Context type in wgpu to be swappable at runtime in the future.

In order to do that, the functions provided by a dynamic blanket implemented Context and it's associated types need to be object safe. For surface creation, instead of passing a impl trait that combines both HasRawWindowHandle and HasRawDisplayHandle, we separate the types into their RawDisplayHandle and RawWindowHandle parts internally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,8 @@ the same every time it is rendered, we now warn if it is missing.
 ### Changes
 
 #### General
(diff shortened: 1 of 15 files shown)


jimblandy (approved):
Looks good to me. A nice cleanup, actually.

PR #3023, opened 2022-09-14
Buffer usages mismatch check and documentation for mapped_at_creation size requirement.

Imberflur:
**Checklist**

- [x] Run `cargo clippy`.
- [ ] ~~Run `RUSTFLAGS=--cfg=web_sys_unstable_apis cargo clippy --target wasm32-unknown-unknown` if applicable.~~

**Connections**
I based this on the documented validation behavior:
* <https://docs.rs/wgpu/0.13.1/wgpu/struct.BufferUsages.html>
* <https://gpuweb.github.io/gpuweb/#dom-gpudevice-createbuffer>

**Description**
Buffer usage mismatch validation was lost to a refactor. Specifically, this is the check needed when one of the `MAP_*` usages is enabled. This re-adds the check.

I also added some documentation on the the validation of the buffer size when `mapped_at_creation` is `true`.

**Testing**
A test was added at `wgpu/tests/buffer_usage.rs`. I confirmed that improper combinations were being allowed via this test and that with the changes they now fail validation.


cwfitzgerald (approved):
I love the tests!

Imberflur:
I was looking over this again and realized there is a small issue with the way the tests are set up. Specifically, I tried to reduce the amount of times we need to perform the setup so that it doesn't need to be done before running each case. However, this means with the failing cases we only test that the first item fails! :sob: 

Maybe I could instead use error scopes to catch the errors so each failure case doesn't need to redo the whole setup?

PR #3034, opened 2022-09-18
Add `get_default_configure` to simplify user creation of `SurfaceConfiguration`

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Description**
Most users don't really need to care about the values of `usage`, `format`, `present_mode`, and possibly `alpha_mode`, `color_space`, etc. when configuring `SurfaceConfiguration`. It is only necessary to understand and specify these fields if the user has a clear need for them. `impl Default`  simplifies the use of users.

`Bgra8UnormSrgb` is set as the default format value because it is the only format supported by all platforms. Although Bgra8Unorm and Rgba8Unorm are defined in the WebGPU spec, metal actually only supports Bgra8Unorm as the texture format for the SwapChain

**Testing**
Tested examples locally

### DIFF
--- a/wgpu/examples/framework.rs
+++ b/wgpu/examples/framework.rs
@@ -267,14 +267,7 @@ fn start<E: Example>(
     }: Setup,
 ) {
     let spawner = Spawner::new();
-    let mut config = wgpu::SurfaceConfiguration {
-        usage: wgpu::TextureUsages::RENDER_ATTACHMENT,
-        format: surface.get_supported_formats(&adapter)[0],
-        width: size.width,
-        height: size.height,
-        present_mode: wgpu::PresentMode::Fifo,
-        alpha_mode: surface.get_supported_alpha_modes(&adapter)[0],
-    };
+    let mut config = surface.get_default_config(&adapter, size.width, size.height);
     surface.configure(&device, &config);
 
     log::info!("Initializing the example...");
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ Bottom level categories:
 - Implement `Hash` for `DepthStencilState` and `DepthBiasState`
 - Add the `"wgsl"` feature, to enable WGSL shaders in `wgpu-core` and `wgpu`. Enabled by default in `wgpu`. By @daxpedda in [#2890](https://github.com/gfx-rs/wgpu/pull/2890).
 - Implement `Clone` for `ShaderSource` and `ShaderModuleDescriptor` in `wgpu`. By @daxpedda in [#3086](https://github.com/gfx-rs/wgpu/pull/3086).
+- Add `get_default_config` for `Surface` to simplify user creation of `SurfaceConfiguration`. By @jinleili in [#3034](https://github.com/gfx-rs/wgpu/pull/3034)
 
 #### GLES
 
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -3713,6 +3713,23 @@ impl Surface {
         Context::surface_get_supported_alpha_modes(&*self.context, &self.id, &adapter.id)
     }
 
+    /// Return a default `SurfaceConfiguration` from width and height to use for the [`Surface`] with this adapter.
+    pub fn get_default_config(
+        &self,
+        adapter: &Adapter,
+        width: u32,
+        height: u32,
+    ) -> wgt::SurfaceConfiguration {
+        wgt::SurfaceConfiguration {
+            usage: wgt::TextureUsages::RENDER_ATTACHMENT,
+            format: self.get_supported_formats(adapter)[0],
+            width,
+            height,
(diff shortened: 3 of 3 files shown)


kpreid:
Adding `Default` seems like it might be a future compatibility risk. What if some new backend results in there being no single always-supported format? How about instead making it a method on `Surface` (or `Device`, or a function in `wgpu::util`) that accepts a width and height, and returns a configuration?

jinleili:
All field values within SurfaceConfiguration should have a valid default value, it's just that this value may be reinterpreted in different backends, as in the case of `Auto` for [alpha_mode](https://github.com/gfx-rs/wgpu/pull/2836#).

If used as an util function, additional `surface` and `adapter` are also required to obtain values that can actually be supported

cwfitzgerald:
I think the biggest problem here is that FIFO isn't actually available everywhere - this is a problem I discovered after starting to tell people that FIFO works everywhere 😬  - wayland doesn't actually have any presentation mode other than MAILBOX.

jinleili:
@cwfitzgerald Is it acceptable to use `AutoNoVsync` instead of `Fifo` here?

wgpu-types/src/lib.rs
```diff
@@ -3122,6 +3122,19 @@ pub struct SurfaceConfiguration {
     pub alpha_mode: CompositeAlphaMode,
 }
 
+impl Default for SurfaceConfiguration {
+    fn default() -> Self {
```
cwfitzgerald, line 3126: I think I'd be more happy with a builtin `from_size` function that takes width and height and fills in the rest.

wgpu-types/src/lib.rs
```diff
@@ -3122,6 +3122,19 @@ pub struct SurfaceConfiguration {
...
+        Self {
+            usage: crate::TextureUsages::RENDER_ATTACHMENT,
+            format: TextureFormat::Bgra8UnormSrgb,
+            width: 0,
+            height: 0,
+            present_mode: PresentMode::AutoNoVsync,
```
cwfitzgerald, line 3132: Lets do AutoVsync here

cwfitzgerald (changes requested):

wgpu/examples/framework.rs
```diff
@@ -267,14 +267,7 @@ fn start<E: Example>(
...
     let spawner = Spawner::new();
-    let mut config = wgpu::SurfaceConfiguration {
-        usage: wgpu::TextureUsages::RENDER_ATTACHMENT,
-        format: surface.get_supported_formats(&adapter)[0],
-        width: size.width,
-        height: size.height,
-        present_mode: wgpu::PresentMode::Fifo,
-        alpha_mode: surface.get_supported_alpha_modes(&adapter)[0],
-    };
+    let mut config = wgpu::SurfaceConfiguration::from_size(size.width, size.height);
```
cwfitzgerald, line 270: This won't actually work as Srgb isn't available on the web.

jinleili: Added a wasm32 target compile condition.

cwfitzgerald: I think we need to add the adapter so we can call get_surface_preferred_formats - these kinds of blanket defaults make me uneasy with the general complexity of WSI

wgpu-core/src/instance.rs
```diff
@@ -195,6 +195,26 @@ impl Surface {
 
         Ok(caps)
     }
+
+    pub fn get_default_configure<A: HalApi>(
```
cwfitzgerald, line 199: Can't we implement this completely in user space by using the other functions? This doesn't feel like it needs any special magic.

cwfitzgerald:
Sorry, this one has had a lot of revisions

wgpu-core/src/instance.rs
jinleili: You're right, after leaving `get_default_configure` only in the wgpu crate, the code is now simpler.

wgpu/examples/framework.rs
```diff
@@ -267,14 +267,7 @@ fn start<E: Example>(
...
     let spawner = Spawner::new();
-    let mut config = wgpu::SurfaceConfiguration {
-        usage: wgpu::TextureUsages::RENDER_ATTACHMENT,
-        format: surface.get_supported_formats(&adapter)[0],
-        width: size.width,
-        height: size.height,
-        present_mode: wgpu::PresentMode::Fifo,
-        alpha_mode: surface.get_supported_alpha_modes(&adapter)[0],
-    };
+    let mut config = surface.get_default_configure(&adapter, size.width, size.height);
```
cwfitzgerald, line 270:
```suggestion
    let mut config = surface.get_default_config(&adapter, size.width, size.height);
```

nit: this should be get_default_config or get_default_configuration

cwfitzgerald (approved):
LGTM after one nit

PR #3037, opened 2022-09-20
expose driver info in AdapterInfo

i509VCB:
**Checklist**

- [X] Run `cargo clippy`.

**Description**
Exposes driver info to users via AdapterInfo.

This is a native only feature and has been implemented for Vulkan so far. For Vulkan, this may also be useful for future workaround code.


wgpu-hal/src/gles/adapter.rs
```diff
@@ -169,6 +169,8 @@ impl super::Adapter {
             vendor: vendor_id as usize,
             device: 0,
             device_type: inferred_device_type,
+            driver: String::from("?"),
```
cwfitzgerald, line 172: Let's just keep these blank to signal for "unknown"

cwfitzgerald (changes requested):
One nit

cwfitzgerald (approved):

PR #3042, opened 2022-09-21
gl: `ASTC_HDR` feature detection

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

**Testing**
Tested locally on Android phone (OnePlus Ace Pro) via [wgpu-on-app](https://github.com/jinleili/wgpu-on-app)

### DIFF
--- a/wgpu-hal/src/gles/conv.rs
+++ b/wgpu-hal/src/gles/conv.rs
@@ -115,7 +115,7 @@ impl super::AdapterShared {
             Tf::EacRg11Unorm => (glow::COMPRESSED_RG11_EAC, glow::RG, 0),
             Tf::EacRg11Snorm => (glow::COMPRESSED_SIGNED_RG11_EAC, glow::RG, 0),
             Tf::Astc { block, channel } => match channel {
-                AstcChannel::Unorm => match block {
+                AstcChannel::Unorm | AstcChannel::Hdr => match block {
                     AstcBlock::B4x4 => (glow::COMPRESSED_RGBA_ASTC_4x4_KHR, glow::RGBA, 0),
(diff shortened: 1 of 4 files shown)


wgpu-hal/src/gles/conv.rs
```diff
@@ -115,7 +115,7 @@ impl super::AdapterShared {
             Tf::EacRg11Unorm => (glow::COMPRESSED_RG11_EAC, glow::RG, 0),
             Tf::EacRg11Snorm => (glow::COMPRESSED_SIGNED_RG11_EAC, glow::RG, 0),
             Tf::Astc { block, channel } => match channel {
-                AstcChannel::Unorm => match block {
+                AstcChannel::Unorm | AstcChannel::Hdr => match block {
```
cwfitzgerald, line 118: Shouldn't the formats be different for hdr?

cwfitzgerald (changes requested):
Looks good outside one comment

wgpu-hal/src/gles/conv.rs
jinleili: From [GL spec](https://registry.khronos.org/OpenGL/extensions/KHR/KHR_texture_compression_astc_hdr.txt), astc_ldr decode to vec4\<f16\> or 8 bit sRGB, astc_hdr only decode to vec4\<f16\>, only two types of formats are available for astc: `_RGBA_ASTC_XXX_KHR ` and `_SRGB8_ALPHA8_ASTC_XXX_KHR`.   `_SRGB8_ALPHA8_ASTC_XXX_KHR` cannot apply to astc_hdr
Screen Shot 2022-09-22 at 05 13 30

cwfitzgerald: This is the most bizarre thing to me.

GL is always full of surprises

cwfitzgerald (approved):

Issue #2320, opened 2021-12-26
Texture memory import API


zarik5:
**Is your feature request related to a problem? Please describe.**
In my personal project I need to import texture memory into wgpu (Vulkan backend) from file descriptors on Linux, `HANDLE`s on Windows and `AHardwareBuffer` on Android. fd's and `HANDLE`s are for cross-process interop, `AHardwareBuffer` is for decoder interop.

**Describe the solution you'd like**
I have not designed the API yet.

**Describe alternatives you've considered**
Use the raw handles Vulkan API (initially authored by me). Having a texture memory import API integrated directly into wgpu would remove a lot of boilerplate.

i509VCB:
I'd be willing to help implement this.

My expertise here is on Linux, I don't know how the situation would look on Android or Windows. On Linux the texture memory import/export object is a dmabuf.

I help develop [Smithay](https://github.com/Smithay/smithay), a library for writing Wayland compositors. Smithay heavily relies on texture memory import from dmabufs to implement fast hardware accelerated rendering. I would like to use wgpu as a renderer, but this is one of the big blockers that would make a wgpu renderer something that could be useful in a high performance setting.

Regarding this API, I think it would also make sense to support texture memory export as well.

## API additions

In my opinion the texture memory import/export functionality is generally going to be platform dependent. I don't see any nice way to avoid this. This probably means each platform will need it's own codepath to support these features.

If there are any conventions for platform specific apis, please do mention these.

## Technical details

Dmabuf import/export can be implemented in both EGL and Vulkan given that the requisite extensions are available:

For Vulkan the following extensions are required:
* `VK_EXT_external_memory_dma_buf`
* `VK_EXT_image_drm_format_modifier` (to get a list of valid DRM modifiers)

These Vulkan extensions are supported in the latest NVIDIA beta driver (515 at the time of writing) and generally in Mesa (anv, radv, v3dv (Vulkan 1.1)).

For EGL the following extensions are required:
* `GL_OES_EGL_image` (this would make import/export for an EGLImage possible if anyone would desire such behavior)
* `EGL_EXT_image_dma_buf_import`
* `EGL_EXT_image_dma_buf_import_modifiers`
* `EGL_MESA_image_dma_buf_export`

I believe NVIDIA implements these extensions from what I've heard. (I will edit the message when get a response from someone to run eglinfo for me).

For Linux with Dmabufs I would expose an API something like the following (specifics pending bikeshed and testing the api):

```rust
[...]

PR #3046, opened 2022-09-23
allow creation of wgpu_hal textures for gles backend

i509VCB:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
Relating to #2320, this should expose enough to allow external textures to work on Linux. I will write a further explanation on the linked issue about Android and what is covered so far by that issue.

**Description**
Add functions to create gles textures from external objects. Specifically, from a GL texture and GL renderbuffer.


expenses:
It'd be great to be able to roll the functionality in https://github.com/gfx-rs/wgpu/pull/2609 into this! I think it'd require something like this:

```rust
pub unsafe fn texture_from_raw(
        &self,
        raw_texture: RawTexture,
        ...
)
```
```rust
enum RawTexture {
    Name(&str),
    WebGl(web_sys::WebGlFramebuffer)
}
```

wgpu-hal/src/gles/device.rs
```diff
@@ -84,6 +84,77 @@ impl CompilationContext<'_> {
...
+    /// - If `drop_guard` is `Some`, the application must manually destroy the texture id. This
+    ///   can be done inside the `Drop` impl of `drop_guard`.
+    #[cfg(any(not(target_arch = "wasm32"), feature = "emscripten"))] // create_texture_from_gl_name is only available on native.
+    pub unsafe fn texture_from_raw(
+        &self,
+        name: u32,
+        desc: &crate::TextureDescriptor,
+        drop_guard: Option<crate::DropGuard>,
+    ) -> super::Texture {
+        let name = std::num::NonZeroU32::new(name).expect("Invalid GL name");
```
cwfitzgerald, line 100: Can't we just bubble up the NonZeroU32 to the caller?

wgpu-hal/src/gles/device.rs
```diff
@@ -676,14 +724,16 @@ impl crate::Device<super::Api> for super::Device {
...
     unsafe fn destroy_texture(&self, texture: super::Texture) {
-        let gl = &self.shared.context.lock();
-        match texture.inner {
-            super::TextureInner::Renderbuffer { raw, .. } => {
-                gl.delete_renderbuffer(raw);
-            }
-            super::TextureInner::DefaultRenderbuffer => {}
-            super::TextureInner::Texture { raw, .. } => {
-                gl.delete_texture(raw);
+        if texture.drop_guard.is_none() {
```
cwfitzgerald, line 727: Where do we actually call the drop guard?

cwfitzgerald (changes requested):
Small comments, otherwise LGTM

wgpu-hal/src/gles/device.rs
i509VCB: Isnt the drop guard's Drop implementation called at the end of the scope since this function takes ownership of the hal texture?

cwfitzgerald: Ah yes, could we just call that out either with a explicit call to `drop(texture)` or a comment?

wgpu-hal/src/gles/device.rs
```diff
@@ -10,6 +10,9 @@ use std::{
 #[cfg(not(target_arch = "wasm32"))]
 use std::mem;
 
+#[cfg(any(not(target_arch = "wasm32"), feature = "emscripten"))]
+use std::num::NonZeroU32;
```
cwfitzgerald, line 14: Let's name this explicitly to avoid the cfg'd use

cwfitzgerald (changes requested):
One more nit, then G2G

cwfitzgerald (approved):

PR #3047, opened 2022-09-23
Fixed buffer sizes encoding on Metal.

dragostis:
**Checklist**

- [x] Run `cargo clippy`.

**Description**
Metal validation layers were complaining that the Naga-generated `_buffer_sizes` argument was not properly encoded in the cases where only a part of the buffers declared in the shader file was used by one entry. Thus, the size buffer would only be partially written.

This patch writes 0s in the case of unused buffer's sizes.

**Testing**
Tested locally. Validation reports no errors.


cwfitzgerald (approved):
Good catch!

PR #3052, opened 2022-09-27
Return an error instead of panicking when canvas context is unavailable

### LINKED ISSUE #3017: WebGL and WebGPU backends should not panic if canvas context is unavailable
The WebGL and WebGPU backends have code which `expect()`s success in obtaining a canvas:

https://github.com/gfx-rs/wgpu/blob/94ce76391b560a66e36df1300bd684321e57511a/wgpu/src/backend/web.rs#L956-L959

https://github.com/gfx-rs/wgpu/blob/94ce76391b560a66e36df1300bd684321e57511a/wgpu-hal/src/gles/web.rs#L36-L41

These will panic if the browser declines to provide a context, which may happen for reasons outside of the application programmer's control (browser does not support WebGPU or WebGL2, insufficient GPU memory, browser glitches) as well as ones which are (such as trying to use the same canvas for `"2d"` drawing). Therefore, the application may have reason to handle the error (if nothing else, to give the user an error message visible on the web page), which is best done by returning an `Err` rather than panicking — especially as Rust on `wasm32` does not support unwinding, so every panic is fatal to the Rust module and can at best be handled from JavaScript.

This is not a solely theoretical scenario — I noticed the problem due to some [Bevy Jam #2](https://itch.io/jam/bevy-jam-2) games failing to start.

**Platform**
web, wgpu 0.13

kpreid:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
Fixes #3017 "WebGL and WebGPU backends should not panic if canvas context is unavailable".

**Description**
- Add a `Result` return value to `wgpu::Instance::create_surface()` and related functions.
- Add `wgpu::CreateSurfaceError`.
- Expand documentation of `create_surface()`.

**Testing**
There is no current way to run automated tests under WASM, but I ran `cargo check` with and without the `webgl` feature; `cargo nextest run`; and some `run-wasm` examples.

**Further notes and concerns**

* This is a breaking change (changing return types to be `Result<Surface>` instead of `Surface`) — is there anything to improve about the new API?
* It would be really nice to have more details from the error, so that it is diagnosable rather than “Sorry, something wrong happened”. The main obstacle to that is the non-specificity of `wgpu_hal::InstanceError`. However, this PR doesn't strictly need that, and it can be enhanced later without `wgpu` API changes.
* The docs for `create_surface()` say more things than they used to. I think they're better, but since this is an `unsafe fn`, everything it says is important. Particularly, I moved the remark about panicking on Metal from Safety to Panics on the assumption that if it's a panic it's not unsound (and therefore should not be in the Safety section), but perhaps that's wrong.

* ~~**This is a draft because some changes are missing for `features = ["webgl"]`.** I discovered that `wgpu_hal` is not depended on directly from `wgpu` (when the target is wasm32) even though it is indirectly depended on, so `wgpu_hal::InstanceError` isn't usable for communicating errors from `wgpu_core` to `wgpu` in the WebGL backend's context creation. So, a different error type will be needed, or that dependency added, or a reexport. I'm posting this PR anyway (as a draft) because it's mostly complete from an API and docs perspective.~~ Resolved.

### DIFF
--- a/wgpu/Cargo.toml
+++ b/wgpu/Cargo.toml
@@ -83,7 +83,7 @@ wgsl = ["wgc?/wgsl"]
 trace = ["serde", "wgc/trace"]
 replay = ["serde", "wgc/replay"]
 angle = ["wgc/angle"]
-webgl = ["wgc"]
+webgl = ["hal", "wgc"]
 emscripten = ["webgl"]
 vulkan-portability = ["wgc/vulkan-portability"]
 expose-ids = []
@@ -100,8 +100,12 @@ optional = true
 [dependencies.wgt]
 workspace = true
 
-[target.'cfg(any(not(target_arch = "wasm32"), target_os = "emscripten"))'.dependencies.hal]
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies.hal]
 workspace = true
+
+[target.'cfg(target_arch = "wasm32")'.dependencies.hal]
+workspace = true
+optional = true
 
 [dependencies]
 arrayvec.workspace = true
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -204,7 +204,7 @@ trait Context: Debug + Send + Sized + Sync {
         &self,
         display_handle: raw_window_handle::RawDisplayHandle,
         window_handle: raw_window_handle::RawWindowHandle,
-    ) -> Self::SurfaceId;
+    ) -> Result<Self::SurfaceId, crate::CreateSurfaceError>;
     fn instance_request_adapter(
         &self,
         options: &RequestAdapterOptions<'_>,
@@ -1816,23 +1816,34 @@ impl Instance {
     ///
     /// # Safety
     ///
-    /// - Raw Window Handle must be a valid object to create a surface upon and
-    ///   must remain valid for the lifetime of the returned surface.
-    /// - If not called on the main thread, metal backend will panic.
+    /// - `raw_window_handle` must be a valid object to create a surface upon.
+    /// - `raw_window_handle` must remain valid until after the returned [`Surface`] is
+    ///   dropped.
+    ///
+    /// # Errors
+    ///
+    /// - On WebGL2: Will return an error if the browser does not support WebGL2,
+    ///   or declines to provide GPU access (such as due to a resource shortage).
+    ///
+    /// # Panics
+    ///
+    /// - On macOS/Metal: will panic if not called on the main thread.
+    /// - On web: will panic if the `raw_window_handle` does not properly refer to a
+    ///   canvas element.
     pub unsafe fn create_surface<
         W: raw_window_handle::HasRawWindowHandle + raw_window_handle::HasRawDisplayHandle,
     >(
         &self,
         window: &W,
-    ) -> Surface {
-        Surface {
+    ) -> Result<Surface, CreateSurfaceError> {
+        Ok(Surface {
             context: Arc::clone(&self.context),
             id: Context::instance_create_surface(
                 &*self.context,
                 raw_window_handle::HasRawDisplayHandle::raw_display_handle(window),
                 raw_window_handle::HasRawWindowHandle::raw_window_handle(window),
-            ),
-        }
+            )?,
+        })
     }
 
     /// Creates a surface from `CoreAnimationLayer`.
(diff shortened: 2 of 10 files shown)


cwfitzgerald:
> Is there a feasible way to run automated tests under WASM? 

Yeah, through wasm-bindgen/wasm-pack, it's on my list to implement soon.

> So, a different error type will be needed, or that dependency added,

We can take a dep on hal, that's fine.

cwfitzgerald (commented):
Nothing looks too bad, please re-request review once this is fully ready!

wgpu/Cargo.toml
```diff
@@ -100,7 +100,7 @@ optional = true
 [dependencies.wgt]
 workspace = true
 
-[target.'cfg(any(not(target_arch = "wasm32"), target_os = "emscripten"))'.dependencies.hal]
```
cwfitzgerald, line 103: What's up with this?

wgpu/src/lib.rs
```diff
@@ -1826,23 +1826,34 @@ impl Instance {
...
-    /// - Raw Window Handle must be a valid object to create a surface upon and
-    ///   must remain valid for the lifetime of the returned surface.
-    /// - If not called on the main thread, metal backend will panic.
+    /// - `raw_window_handle` must be a valid object to create a surface upon.
+    /// - `raw_window_handle` must remain valid until after the returned [`Surface`] is
+    ///   dropped.
+    ///
+    /// # Errors
+    ///
+    /// - On WebGL 2: Will return an error if the browser does not support WebGL 2,
```
cwfitzgerald, line 1835: Minor nit: `WebGL2` is how we generally style it.

wgpu/src/lib.rs
```diff
@@ -2363,6 +2387,22 @@ impl Display for RequestDeviceError {
 
 impl error::Error for RequestDeviceError {}
 
+/// [`Instance::create_surface()`] or a related function failed.
+#[derive(Clone, PartialEq, Eq, Debug)]
+#[non_exhaustive]
+pub struct CreateSurfaceError {
```
cwfitzgerald, line 2393: Let's make this an enum

cwfitzgerald (changes requested):

wgpu/src/lib.rs
kpreid: If we make it an enum, it needs at least one variant, which will have public fields if it has any fields. What should that variant be called, and what fields should it have, given that currently no specific information is available?

Something like this as an explicitly “some platform-specific situation” variant?

```rust
#[non_exhaustive]
pub struct CreateSurfaceError {
    Platform(String),
}
```

I was hoping that the totally-private struct would be adequate as a placeholder until we could fix the lack of information.

wgpu/Cargo.toml
cwfitzgerald: Oh I see the problem. This change makes wgpu-hal a dependency even when targetting WebGPU. We need to follow the dependency model of wgpu-core here, where it's optional on wasm, only enabled if the webgl/emscripten.

wgpu/src/lib.rs
cwfitzgerald: I just mean internally consistent with ourselves like https://docs.rs/wgpu-types/latest/wgpu_types/struct.DownlevelFlags.html#associatedconstant.COMPUTE_SHADERS. Ultimately doesn't matter.

wgpu/src/lib.rs
cwfitzgerald: Reasonable enough.

wgpu/src/lib.rs
kpreid: … done.

wgpu/Cargo.toml
kpreid: Done, I think?

It'd be nice if there was big-picture documentation explaining the intended crate relationships and API/abstraction layering. That would have helped me write even the beginnings of this PR, since there are so many layers that I had to adjust to match each other and test on multiple feature sets.

cwfitzgerald: Agreed.

cwfitzgerald (approved):

cwfitzgerald:
Needs conflict resolution, then g2g

cwfitzgerald (approved):

cwfitzgerald:
> By the way, it would be nice if there was a CONTRIBUTING.md or other documentation that specified project preferences for commit/PR management, such as whether I ought to have just fully squashed unconditionally and not thought about it.)

That would be great to have.

> I wasn't sure from the beginning whether this is really “major” (and now, deserving of its own entire section heading), but that does seem to be how breaking changes are reported, so perhaps it's just right.

Yeah any breaking changes we should call out explicitly even if they aren't "major" in scope.

PR #3054, opened 2022-09-28
Handle unsupported surfaces more gracefully

i509VCB:
**Checklist**

- [X] Run `cargo clippy`.

**Description**
The Vulkan backend panicked when creating a surface when using a supported window and display handle but the dependent extensions are not available.

Per discussion on Matrix, the surface should instead not be usable on that device.

Getting the hal surface in wgpu-core may now return `None` to correctly represent the expected semantics.

**Testing**
Testing with DX12 and Metal is needed.


cwfitzgerald (approved):
Good stuff.

Issue #1783, opened 2021-08-09
Memory leak on MacOS | M1 OSX


sotrh:
**Description**
Some users of my tutorial have been experiencing memory issues on my buffer tutorial. It seems to only be an issue on the new M1 chips.

**Repro steps**
You'll need a mac with and M1 chip, then run the tutorial code here https://github.com/sotrh/learn-wgpu/tree/master/code/beginner/tutorial4-buffer. If you have the repo already downloaded you can just run `cargo run --bin tutorial4-buffer`.

**Expected vs observed behavior**
The expected behaviour is no memory leaks on M1

**Extra materials**
I don't have a Mac, so I can't provide hardware specifics other that it's occurring on M1, but the I'll link the issue from my repo here. https://github.com/sotrh/learn-wgpu/issues/207

**Platform**
OSX with M1 chip, wgpu 0.9

kvark:
Similar report in #1537

JunkuiZhang:
On my Intel MacBook Pro it works as expected, I think its a M1 SoC related issue.

kvark:
I don't have an M1. We need help investigating this.

jbatez:
I've been debugging a similar (same?) issue the last couple days. It has something to do with `begin_render_pass` and high framerates. You can reproduce with `wgpu/examples/cube` with only a slight modification.

In `framework.rs`, set `target_framework` way above 60hz. On my system, I don't see the issue at 200hz, but I deifnitely see it at 300hz.

jbatez:
With the high framerate modification above, if I remove the `begin_render_pass` call and all the `rpass.*` method calls, the problem goes away. If I keep the `begin_render_pass` call but remove all the method calls, the problem persists. Hence the conclusion that it has something to do with `begin_render_pass`.

jbatez:
Also, note: I first noticed this bug with `v0.9` but I still see it in `master`.

jbatez:
Also maybe noteworthy: with the `target_framerate` modification above plus a switch to `PresentMode::Fifo`, the fullscreen framerate reported to stdout reaches well above my display's framerate. Is that expected? If not, maybe the command queue's getting filled faster than the GPU can drain it.

jbatez:
It's not just when the window' fullscreen, but when it's completely obscured. Adding a `NSWindowOcclusionStateVisible` check before frame acquisition helps, but memory usage still seems to shoot up on occluded -> unoccluded transitions.

[...]

PR #3056, opened 2022-09-30
Fix memory leak on macOS

xiaopengli89:
**Checklist**

- [x] Run `cargo clippy`.

**Description**
Fix remaining leaks for metal backend.

### DIFF
--- a/wgpu-hal/src/metal/command.rs
+++ b/wgpu-hal/src/metal/command.rs
@@ -900,11 +900,13 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
         self.begin_pass();
 
         let raw = self.raw_cmd_buf.as_ref().unwrap();
-        let encoder = raw.new_compute_command_encoder();
-        if let Some(label) = desc.label {
-            encoder.set_label(label);
-        }
-        self.state.compute = Some(encoder.to_owned());
+        objc::rc::autoreleasepool(|| {
+            let encoder = raw.new_compute_command_encoder();
+            if let Some(label) = desc.label {
+                encoder.set_label(label);
+            }
+            self.state.compute = Some(encoder.to_owned());
+        });
     }
     unsafe fn end_compute_pass(&mut self) {
         self.state.compute.take().unwrap().end_encoding();
--- a/wgpu-hal/src/metal/device.rs
+++ b/wgpu-hal/src/metal/device.rs
@@ -235,13 +235,15 @@ impl crate::Device<super::Api> for super::Device {
 
         //TODO: HazardTrackingModeUntracked
 
-        let raw = self.shared.device.lock().new_buffer(desc.size, options);
-        if let Some(label) = desc.label {
-            raw.set_label(label);
-        }
-        Ok(super::Buffer {
-            raw,
-            size: desc.size,
+        objc::rc::autoreleasepool(|| {
+            let raw = self.shared.device.lock().new_buffer(desc.size, options);
+            if let Some(label) = desc.label {
+                raw.set_label(label);
+            }
+            Ok(super::Buffer {
+                raw,
+                size: desc.size,
+            })
         })
     }
     unsafe fn destroy_buffer(&self, _buffer: super::Buffer) {}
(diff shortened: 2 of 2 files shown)


jinleili:
From Apple's [autoreleasepool](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/mmAutoreleasePools.html) doc:
> ... Therefore you typically do not have to create an autorelease pool block yourself, or even see the code that is used to create one. There are, however, three occasions when you might use your own autorelease pool blocks:
> - If you are writing a program that is not based on a UI framework, such as a command-line tool.
> - If you write a loop that creates many temporary objects.
	You may use an autorelease pool block inside the loop to dispose of those objects before the next iteration. Using an autorelease pool block in the loop helps to reduce the maximum memory footprint of the application.
> -  If you spawn a secondary thread.
	You must create your own autorelease pool block as soon as the thread begins executing; otherwise, your application will leak objects.

I have also checked several official Metal demos,  all Metal instructions  put into @autoreleasepool is not necessary. Also, no resulting memory leaks were observed on e Intel and M1 Macs.

xiaopengli89:
If the user does not interact with the program for a long time, leakage will occur, such as a video playback software on mac, but the user only operates remotely through the mobile phone, related #2092.

xiaopengli89:
Here is the memory leak diagnosed with Xcode:

jinleili:
Indeed, after the window is completely occluded, the occupied memory will suddenly increase to a higher level, and then it can be observed that it will slowly decrease, and then there will be repeats, but it will always be below that highest value.

I also tested Apple's official metal demo and didn't observe this phenomenon.

This phenomenon is not like a memory leak caused by the wgpu itself, but a combination of wgpu + winit, and my personal projects that do not use winit have not observed this problem.

jinleili:
I'm sorry, but after testing your PR, the memory growth phenomenon is still the same as described above.

截屏2022-10-08 12 09 25

xiaopengli89:
The problem with the occluded window should be related to this issue #1783.

jinleili:
Thanks for the link, I didn't realize this issue has been discussed for so long. Basically I agree with the statement on [this beby issue](https://github.com/bevyengine/bevy/issues/5856): 
> It was concluded that this is impractical to solve entirely at the `wgpu` level, and that there are two primary workarounds for this issue:
>1. Detecting window occlusion using `winit`'s `WindowEvent::Occluded` and avoiding any rendering at all when the window is occluded.
>2. Driving rendering through `CVDisplayLink` (macOS) / `CADisplayLink` (iOS) frame callbacks.

xiaopengli89:
In my scenario, it was not caused by the concluded window, here is the memory growth of the wgpu process:

jinleili:
[WebKit's metal-cpp doc](https://github.com/WebKit/WebKit/blob/main/Source/WebGPU/WebGPU/metal-cpp/README.md) mentions a way to debug autoreleasepool:
> Use the Environment Variable `OBJC_DEBUG_MISSING_POOLS=YES` to print a runtime warning when an autoreleased object is leaked because no enclosing `AutoreleasePool` is available for its thread.

It doesn't seem to eliminate the missing autoreleasepool warning. And since autoreleasepool isn't zero-overhead, maybe we can dig a little deeper into the cause of the problem to make the yy backend perform better.
截屏2022-10-08 17 31 58

xiaopengli89:
I tested this pr with `OBJC_DEBUG_MISSING_POOLS=YES` and it reduces the warnings from 101 to 45, maybe there are other leaks, also outside of wgpu.

cwfitzgerald:
I'm quite unfamiliar with all this autorelease pool stuff, so I'll defer to general consensus if this is something we need. Less warnings sounds generally good though.

kpreid:
2¢ from someone who once wrote a lot of Objective-C:

_If_ adding an autorelease pool reduces leaks, then there _definitely should_ be an autorelease pool added. There might be a more ideal scope for the pool — wider to catch more things, or narrower to reduce the maximum accumulation of not-yet-autoreleased objects — but there should definitely be a pool.

The only way adding an autorelease pool can cause harm (besides the cost of constructing it) is if code tries to `autorelease` an object pointer inside the pool scope and then use the pointer after the pool scope ends. You can think of the required discipline as vaguely like a Rust lifetime: `autorelease`ing an object pointer is sort of like converting `Rc<T>` into `&'a T`, and the end of the pool's scope is dropping the `Rc<T>`, thus invalidating the “reference” you were using.

Skimming, I don't see any risks of that in this PR, because the pools all end at the same time the function scope ends (and if the WGPU objects being returned were containing non-`retain`ed references to ObjC objects, then that'd *already* be unsound).

cwfitzgerald (approved):
Cool then, lets land this!

kpreid:
FWIW, I just tested my application with wgpu from git `master` vs stable 0.14.0, and there seems to be no longer a casually visible memory leak (stable showed a constantly growing size in Activity Monitor, around 0.01 MB/s). So, this change seems to have made a substantial improvement.

PR #3063, opened 2022-10-05
Set the image stride to zero when updating a single layer on metal.

### LINKED ISSUE #3062: [metal] destinationBytesPerImage should probably be 0 when copying to/from a 2d texture with
**Description**

When using `queue.write_texture` to update a subset of a simple 2d texture (no arrays, no mipmaps, etc), metal validation fails with an error of the form:

```
-[MTLDebugBlitCommandEncoder validateCopyFromBuffer:sourceOffset:sourceBytesPerRow:sourceBytesPerImage:sourceSize:toTexture:destinationSlice:destinationLevel:destinationOrigin:options:]:804: failed assertion `Copy From Buffer Validation
totalBytesUsed(1048576) must be <= [sourceBuffer length](2048).
```

where `totalBytesUsed` is equal to the total size of the texture (+ offset if there is one) and `sourceBuffer length` is the amount of data being passed for that subset of the image that we are trying to update.

The program runs on its own, but validation fails for example when running in xcode which prevents using the associated gpu tooling.

Relevant metal doc: https://developer.apple.com/documentation/metal/mtlblitcommandencoder/1400752-copy

The problem seems to come from `destinationBytesPerImage` which (somewhat confusingly) expresses the stride between layers of the image (for array textures and 3d textures) rather than the image size (same quantity but not same purpose).

The closest I could find online of someone running into a similar issue is: https://stackoverflow.com/questions/43069257/how-to-use-mtlblitcommandencoder-for-copying-interlaced-video-fields-into-a-mtlb

The answer suggests passing 0 as the `destinationBytesPerImage`, I think that it sounds reasonable. 

Note that letting the user pass `None` in `rows_per_image` does not work around the problem because of https://github.com/gfx-rs/wgpu/blob/26f223964e7bf57b7f207cf09efa82b732c470cf/wgpu-core/src/device/queue.rs#L613

**Repro steps**

```rust
        let tex = device.create_texture(&wgpu::TextureDescriptor {
            label: None,
            size: wgpu::Extent3d {
                width: 1024,
                height: 1024,
                depth_or_array_layers: 1,
            },
            format: wgpu::TextureFormat::R8Uint,
            usage: wgpu::TextureUsages::COPY_DST | wgpu::TextureUsages::TEXTURE_BINDING,
            mip_level_count: 1,
            sample_count: 1,
        });
        let data = vec![1u8; 1024 * 2];
        // Write the first two rows
        queue.write_texture(
            wgpu::ImageCopyTexture {
                texture: &tex,
                mip_level: 0,
                origin: wgpu::Origin3d::ZERO,
                aspect: wgpu::TextureAspect::All,
            },
            bytemuck::cast_slice(&data),
            wgpu::ImageDataLayout {
                offset: 0,
                bytes_per_row: std::num::NonZeroU32::new(1024),
                rows_per_image: std::num::NonZeroU32::new(1024),
            },
            wgpu::Extent3d {
                width: 1024,
                height: 2,
                depth_or_array_layers: 1,
            }
        );
```

Expect the code to run fine when validation is enabled (for example when running in xcode).

**Platform**

MacOS, wgpu 0.13.1

nical:
Probably just a tad late to make the cut to the next version, but perhaps it could squeeze in a dot release?

**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Fixes #3062 

**Description**

When updating a single layer of a texture on metal, set the stride to zero to avoid angering the validatio layer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,7 @@ SurfaceConfiguration {
 - Add the missing `msg_send![view, retain]` call within `from_view` by @jinleili in [#2976](https://github.com/gfx-rs/wgpu/pull/2976)
 - Fix `max_buffer` `max_texture` and `max_vertex_buffers` limits by @jinleili in [#2978](https://github.com/gfx-rs/wgpu/pull/2978)
 - Remove PrivateCapabilities's `format_rgb10a2_unorm_surface` field by @jinleili in [#2981](https://github.com/gfx-rs/wgpu/pull/2981)
+- Fix validation error when copying into a subset of a single-layer texture by @nical in [#3063](https://github.com/gfx-rs/wgpu/pull/3063)
 
 #### Vulkan
 - Fix `astc_hdr` formats support by @jinleili in [#2971]](https://github.com/gfx-rs/wgpu/pull/2971)
--- a/wgpu-hal/src/metal/command.rs
+++ b/wgpu-hal/src/metal/command.rs
@@ -228,15 +228,21 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
                 .buffer_layout
                 .bytes_per_row
                 .map_or(0, |v| v.get() as u64);
-            let bytes_per_image = copy
-                .buffer_layout
-                .rows_per_image
-                .map_or(0, |v| v.get() as u64 * bytes_per_row);
+            let image_byte_stride = if extent.depth > 1 {
+                copy.buffer_layout
+                    .rows_per_image
+                    .map_or(0, |v| v.get() as u64 * bytes_per_row)
+            } else {
+                // Don't pass a stride when updating a single layer, otherwise metal validation
+                // fails when updating a subset of the image due to the stride being larger than
+                // the amount of data to copy.
+                0
+            };
             encoder.copy_from_buffer_to_texture(
                 &src.raw,
                 copy.buffer_layout.offset,
                 bytes_per_row,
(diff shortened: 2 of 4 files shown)


cwfitzgerald (approved):
You and your amazing tests 😍 

Got into the latest :)

cwfitzgerald:
Seems like this code is broken on windows too 😆

teoxoy:
Should we also do the same for `copy_texture_to_buffer`?

> The number of bytes between each 2D image of a 3D texture. This value needs to be a multiple of the source texture’s pixel size, in bytes.
> 
> Set this value to 0 for 2D textures, which means sourceSize.depth is equal to 1.

The docs of `destinationBytesPerImage` (arg of [`copyFromTexture`](https://developer.apple.com/documentation/metal/mtlblitcommandencoder/1400756-copyfromtexture)) are the same as `sourceBytesPerImage` (arg of [`copyFromBuffer`](https://developer.apple.com/documentation/metal/mtlblitcommandencoder/1400752-copyfrombuffer)).

The [stackoverflow post](https://stackoverflow.com/questions/43069257/how-to-use-mtlblitcommandencoder-for-copying-interlaced-video-fields-into-a-mtlb) in the issue https://github.com/gfx-rs/wgpu/issues/3062#issue-1397658278 was also referring to `destinationBytesPerImage`.

I assume this hasn't been an issue since the test copies back the full texture and not a subregion.

PR #3065, opened 2022-10-05
Vulkan Command pool optimization

gents83:
Looking at Vulkan Samples and other Vulkan materials even on mobile it seems that the right way to manage command pools in a transient way as they are used by wgpu should be enabling the right flag and resetting it without any flags.

Vulkan Samples:
https://github.com/KhronosGroup/Vulkan-Samples/blob/master/framework/core/command_pool.cpp

Arm Mobile developer:
https://arm-software.github.io/vulkan_best_practice_for_mobile_developers/samples/performance/command_buffer_usage/command_buffer_usage_tutorial.html

In my local profilings in my proto engine we're passing from around 9ms to around 800μs (more than 10x speedup)
No warning on leaks from validation layers is appearing in RenderDoc on my side too.

### DIFF
--- a/wgpu-hal/src/vulkan/command.rs
+++ b/wgpu-hal/src/vulkan/command.rs
@@ -98,7 +98,7 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
         let _ = self
             .device
             .raw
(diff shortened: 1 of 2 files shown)


cwfitzgerald (approved):
LGTM

PR #3066, opened 2022-10-05
Implement missing PrettyError impls

### LINKED ISSUE #3061: implement PrettyError for DrawError
**Description**
Given for example a `LateMinBufferBindingSizeMismatch` error:
```
In a RenderPass
      note: encoder = `egui_webpainter_paint_and_update_textures`
    In a draw command, indexed:false indirect:false
      note: render pipeline = `<RenderPipeline-(1, 1, Gl)>`
    Buffer is bound with size 16 where the shader expects 32 in group[0] compact index 0
```
.. then the inner `DrawError` is not pretty printed

scoopr:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Closes #3061 

**Description**
Some of the Error impls that refer to the resource ids were not formatting the labels for the resources

**Testing**
I must confess that I only compile-tested. Inducing or synthesizing the error conditions are a bit difficult.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,7 @@ SurfaceConfiguration {
   enabled. By @imberflur in [#3023](https://github.com/gfx-rs/wgpu/pull/3023)
 - Fixed `CommandEncoder` not being `Send` and `Sync` on web by @i509VCB in [#3025](https://github.com/gfx-rs/wgpu/pull/3025)
 - Document meaning of `vendor` in `AdapterInfo` if the vendor has no PCI id.
+- Fix missing resource labels from some Errors by @scoopr in [#3066](https://github.com/gfx-rs/wgpu/pull/3066)
 
 #### Metal
 - Add the missing `msg_send![view, retain]` call within `from_view` by @jinleili in [#2976](https://github.com/gfx-rs/wgpu/pull/2976)
--- a/wgpu-core/src/command/query.rs
+++ b/wgpu-core/src/command/query.rs
@@ -112,6 +112,18 @@ pub enum QueryError {
     InvalidQuerySet(id::QuerySetId),
 }
 
+impl crate::error::PrettyError for QueryError {
+    fn fmt_pretty(&self, fmt: &mut crate::error::ErrorFormatter) {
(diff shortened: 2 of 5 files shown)


scoopr:
I notice that these are possibly missing label formatting:
* RequestAdapterError
* ClearError
* WaitIdleError
* QueueSubmitError
Those seemed less important though.
Didn't double check the already existing PrettyError impls that they still handle all the correct labels

scoopr:
I also promote this to a point release if merged :)

cwfitzgerald (approved):
In the nick of time; changelog pls!

Wumpf:
@scoopr thank you so much for going after this! But wasn't the bug I encountered a missing pretty print for `DrawError` which is still not covered?

cwfitzgerald:
(if closing the bug was not intended, that is my fault not scoopr's - I edited the body to close)

Wumpf:
yeah from my understanding needs to be reopened. From my understanding the issue is still there, but scoopr's changes are great regardless :)

scoopr:
This should fix #3061, I believe I misattributed the missing bits initially, when talking about it on Matrix. The #3061 error that is missing label is actually `PassErrorScope::Draw`, `PassErrorScope` already implemented `PrettyError`,  but it had been forgotten from `format_pretty_any` downcasts.

Neither `DrawError` nor `LateMinBufferBindingSizeMismatch` actually referenced any `Id`s to resolve.

It would be great if you tested with the original error condition though, to make sure!

Wumpf:
ohhh I see that makes sense, double missunderstanding then :). I'll test it today and get back to you!

Wumpf:
the fact that encoders keep their label regardless of debug_assertions enabled could be regardless a bug - but it seems it makes the code a lot easier since it needs to pass on that label to commandbuffer, so the tradeoffs are imho in favor of status quo

scoopr:
Oh alright, that makes more sense, `PassErrorScope` didn't really need to be handled in `format_pretty_any` after all, and now I fear that its actually printing the labels doubly, as it gets printed through the `MapPassError` impl.

PR #3068, opened 2022-10-05
Anger CI

cwfitzgerald:
DO NOT MERGE


crowlKats:
@cwfitzgerald hm thats weird that its happening out of nowhere; i'll investigate

crowlKats:
@cwfitzgerald cts runner is now fixed

crowlKats:
running is now fixed, though fairly quickly tests fail. do you want me to look into them?

cwfitzgerald:
That would be great - seems to be some CTS/deno mismatch or something - there's no great rush though.

crowlKats:
yea you are right; the runner is set to use a CTS commit from over a year ago. however updating it causes errors with the `run CTS` bash script

cwfitzgerald:
Going to rebase this on top of #3097 soon-ish

crowlKats:
just as a heads-up: requiredLimits option is currently broken due to a problem in deno, so it potentially might take a while until i continue here

cwfitzgerald:
@crowlKats If I pull the CI parts of this, would this be landable - I realize we really shouldn't be having this just sitting around.

crowlKats:
@cwfitzgerald I think so

cwfitzgerald:
Discussed offline that @crowlKats will bring the non-CI changes into a upcoming deno update branch - going to close, but keep the branch alive.

PR #3070, opened 2022-10-06
webgl backend supports now non-srgb framebuffer

### LINKED ISSUE #3059: [gles] surface from canvas supports only srgb
**Description**
Surfaces created from a canvas in gles only support RGBA8srgb. Would expect non-srgb to work as well

Haven't checked how prevalent this is, might happen in other configurations as well and just be a property of the gles backend.

**Platform**
gles / wasm / browser

Wumpf:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
fixes #3059 

**Description**
enable non-srgb surfaces for gles
Still can't get around the texture copy due to the opengl coordinate flipping issue. But at least it's a "simple blit" which egl.rs already does (special care needed to be taken in the webgl case)

**Testing**
tested against egui in the browser which prefers non-srgb targets
image
patching wgpu in egui with this version (and building with the right feature flags) is sufficient :)

*Tried* testing egl.rs as well, but I'm failing to build angle for my apple silicon machine. I **think** it should work as well but if not this introduces a new bug (tit for tat!)


wgpu-hal/src/gles/adapter.rs
```diff
@@ -806,6 +806,19 @@ impl crate::Adapter<super::Api> for super::Adapter {
             formats.push(wgt::TextureFormat::Rgba16Float)
         }
         if surface.presentable {
+            let mut formats = vec![
```
cwfitzgerald, line 809: This shadowing the outer format vec seems very strange.

cwfitzgerald (changes requested):
LGTM outside of one comment. Would be good to get this resolved in the egl case too.

wgpu-hal/src/gles/adapter.rs
Wumpf: oops, that must have happend when rebasing because of https://github.com/gfx-rs/wgpu/pull/3045
Fixed and disabled the new f16 path - I think we'd need a special path for that since right now it would get the srgb treatment and blit wouldn't work. cc: @jinleili

Wumpf:
fairly sure it should "just work" with egl as we - as far as I've seen - resolve srgb there only via blit, never via shader. As writen in description, I failed to test it :/

wgpu-hal/src/gles/adapter.rs
cwfitzgerald: Could you add that path in this PR? Hopefully it shouldn't be that difficult - the #[cfg] on the f16 path feels kinda dirty

Wumpf: you were right, the blit actually supports it as is. Went through the motion to test it and renders just fine :)

Wumpf: f16 works as is

cwfitzgerald: Great!

cwfitzgerald (approved):

Wumpf:
fixed changelog conflict in rebase (also added missing `General` section there)

cwfitzgerald (approved):

PR #3071, opened 2022-10-06
Fix egl only enabled with emscripten feature flag

Wumpf:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
encountered during test attempt for #3070 

**Description**
fixes needing to enable `emscripten` feature flag to use egl (i.e. gles for non-web!)

**Testing**
builds but I don't have a libegl.dylib and I failed building angle so far, so compile only :(


Wumpf:
ignore me, I can't read :)

PR #3076, opened 2022-10-06
vulkan: Don't use pointer to dropped PhysicalDeviceDriverProperties.

jimblandy:
Detected by ASAN, but seems impossible to exploit.

Since `ash::vk::definitions::PhysicalDeviceDriverProperties` is `Copy`, the statement

    if let Some(driver) = phd_capabilities.driver { ... }

actually makes `driver` a local copy of the struct. The code uses `as_ptr` to create a pointer to the `driver_name` field of this local copy, and then tries to use that pointer outside the `if let`, when the local copy has gone out of scope. This is UB.

Taking a reference to the properties struct is correct and more efficient.

**Checklist**

- [X] Run `cargo clippy`.

**Connections**

**Description**

**Testing**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,7 +59,7 @@ Bottom level categories:
 
 #### Vulkan
 
-- Don't use a pointer to a local copy of a `PhysicalDeviceDriverProperties` struct after it has gone out of scope. In fact, don't make a local copy at all. By @jimblandy in [#3076](https://github.com/gfx-rs/wgpu/pull/3076).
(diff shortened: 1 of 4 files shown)


jimblandy:
We could avoid those `unsafe` blocks altogether if `CStr::from_bytes_until_nul` were stable.

cwfitzgerald (approved):
This is good as is, but it might be useful to have our own utility function that implements from_bytes_until_nul so that we get the lifetime and don't need the unsafe anymore

cwfitzgerald (approved):
Music to my eyes!

cwfitzgerald:
CI is mad

PR #3080, opened 2022-10-07
Example for `copyExternalImageToTexture`

ybiletskyi:
**Description**
Example for `copyExternalImageToTexture` function.


ybiletskyi:
@cwfitzgerald I have created a separated PR for `copyExternalImageToTexture` example. So we could continue a discussion here.

gilescope:
We could kill 2 birds with one stone. Would be great if this example could also show how to use webgl anti aliasing turned on for textures?

cwfitzgerald:
Anti-aliasing is no different on webgl than other targets

wgpu/examples/hello-copy-source/main.rs
```diff
@@ -0,0 +1,369 @@
...
+    Vertex {
+        position: [0.44147372, 0.2347359, 0.0],
+        tex_coords: [0.9414737, 0.2652641],
+    }, // E
+];
+
+const INDICES: &[u16] = &[0, 1, 4, 1, 2, 4, 2, 3, 4, /* padding */ 0];
+
+#[cfg(target_arch = "wasm32")]
+static mut imageData: Option<ImageBitmap> = None;
```
cwfitzgerald, line 71: `static mut` is basically guarenteed unsound city - we should use OnceCell plus a normal static

wgpu/examples/hello-copy-source/main.rs
```diff
@@ -0,0 +1,369 @@
...
+                mip_level_count: 1,
+                sample_count: 1,
+                dimension: wgpu::TextureDimension::D2,
+                format: wgpu::TextureFormat::Rgba8UnormSrgb,
+                usage: wgpu::TextureUsages::COPY_DST
+                    | wgpu::TextureUsages::RENDER_ATTACHMENT
+                    | wgpu::TextureUsages::TEXTURE_BINDING,
+            });
+
+            queue.copy_external_image_to_texture(image, texture.as_image_copy(), texture_size);
```
cwfitzgerald, line 132: tbh I'm not sure of the value of a whole example for something that is ultimately just this one line which isn't very complex

ybiletskyi: Probably You are right. The example doesn't contains enough value. So I'll close PR.

PR #3081, opened 2022-10-07
Fixed missing device feature in mipmap example

Olaroll:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
None

**Description**
I'm currently learning wgpu and noticed that the mipmap example didn't work. I added the feature indicated in the error to the appropriate place and now it works. The feature in question is `WRITE_TIMESTAMP_INSIDE_PASSES`.

**Testing**
Ran the example and it worked.


cwfitzgerald (changes requested):
Good stuff but we also need the list of features to check for when creating the query set to be extended with this feature

Olaroll:
Not quite sure what a query set does, but hopefully that fixes it 😅

cwfitzgerald (approved):
Thanks!

Query sets qre ways of asking the GPU information (either timing or statistics) and managing the queries while they are being processed - the features involved are required for their creation.

PR #3083, opened 2022-10-07
Convert All Default Implementations on Enums to derive(Default)

### LINKED ISSUE #3077: Convert All Default Implementations on Enums to derive(Default) + #[default]
You can now tag enum variants as #[default] to be able to derive default, so we should just use that instead of manual impls

killianc3:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Closes #3077.

**Description**
Since https://github.com/rust-lang/rust/pull/101040, we can now mark enums as `#[default]`, this replaces manual implementation where possible.

**Testing**
I ran the cargo test before and after the changes and nothing changed, so I guess nothing was broken because these are very simple changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on:
 
 env:
   RUST_BACKTRACE: 1
-  RUST_VERSION: 1.59
+  RUST_VERSION: 1.62
 
 # We distinguish the following kinds of builds:
 # - local: build for the same target as we compile on, and do local tests
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -145,6 +145,7 @@ both `raw_window_handle::HasRawWindowHandle` and `raw_window_handle::HasRawDispl
 
 #### General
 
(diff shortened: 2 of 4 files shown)


cwfitzgerald (approved):
Code looks good, just need to change the msrv in the Readme and ci to be the latest needed for #[default]

killianc3:
It seems that the minimum supported version is now `1.62.0`. (https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1620-2022-06-30)

cwfitzgerald:
Nice, not nearly as bad as I thought

killianc3:
Yes, it's not so bad and it seems to have passed the test now. Should we add an entry to the changelog to say that the msvr has changed ? If so, I don't really know how and what to add, any ideas ?

cwfitzgerald:
Look at last releases big announcements - I'd add a thing like that that just says what msrv we bumped to.

killianc3:
Thank you ! I think it's pretty understandable as is, what do you think ?

cwfitzgerald (approved):
Thanks!

PR #3091, opened 2022-10-11
web: call `create_query_set` and  `write_timestamp` of web-sys

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Testing**
Tested locally via Chrome Canary

### DIFF
--- a/wgpu/src/backend/web.rs
+++ b/wgpu/src/backend/web.rs
@@ -144,11 +144,15 @@ impl crate::ComputePassInner<Context> for ComputePass {
             .dispatch_workgroups_indirect_with_f64(&indirect_buffer.0, indirect_offset as f64);
     }
 
-    fn write_timestamp(&mut self, _query_set: &(), _query_index: u32) {
+    fn write_timestamp(&mut self, _query_set: &Sendable<web_sys::GpuQuerySet>, _query_index: u32) {
         panic!("WRITE_TIMESTAMP_INSIDE_PASSES feature must be enabled to call write_timestamp in a compute pass")
     }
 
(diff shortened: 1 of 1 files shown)


wgpu/src/backend/web.rs
```diff
@@ -144,11 +144,15 @@ impl crate::ComputePassInner<Context> for ComputePass {
             .dispatch_workgroups_indirect_with_f64(&indirect_buffer.0, indirect_offset as f64);
     }
 
-    fn write_timestamp(&mut self, _query_set: &(), _query_index: u32) {
-        panic!("WRITE_TIMESTAMP_INSIDE_PASSES feature must be enabled to call write_timestamp in a compute pass")
```
cwfitzgerald, line 148: I think we should keep the wording about the lacking feature as the reason for failure.

cwfitzgerald (changes requested):
Do we also need to change any feature detection code?

wgpu/src/backend/web.rs
jinleili: There should be no need to change feature detection, as it has already been done in wgpu-hal. `WRITE_TIMESTAMP_INSIDE_PASSES` is a native-only feature, web-sys does not implement `write_timestamp` for compute and render pass.

cwfitzgerald: I agree, but the story we should be presenting to the user is that WebGPU is just another device, except this device never supports `WRITE_TIMESTAMP_INSIDE_PASSES` so the error itself is not unique to wasm, just a property of the feature set of the device.

cwfitzgerald (approved):

PR #3092, opened 2022-10-11
Implement Hash for DepthStencilState and DepthBiasState

### LINKED ISSUE #3087: derive(Hash) for DepthStencilState and DepthBiasState

With the number of pipelines in use in a reasonably complicated program, it's useful to be able to cache wgpu::RenderPipeline objects. Most of the data involved in creating a RenderPipelineDescriptor is already hashable, but notably neither wgpu::DepthStencilState nor wgpu::DepthBiasState have a derive(Hash).

Add #[derive(Hash)] to the remaining data-only structs involved in resource creation.

Users can manually implement hashing for these types in their own code, but they can't actually implement the Hash trait directly (3rd-party-impl rules), nor do they have the ability to auto-derive the implementation.

killianc3:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Hopefully, a simple solution for this issue : closes #3087 

**Description**
As described in the question `DepthStencilState` and `DepthBiasState` doesn't have a Hash implementation, I tried to manage the implementation manually. I've never done this before, so it may not be perfect, but I'm interested in your opinion on this.

**Testing**
Cargo fmt / test / clippy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,7 @@ Bottom level categories:
 
 #### General
 
-- Implement `Hash` for `DepthStencilState` and `DepthBiasState` in `wgpu-types`.
+- Implement `Hash` for `DepthStencilState` and `DepthBiasState`
 - Add the `"wgsl"` feature, to enable WGSL shaders in `wgpu-core` and `wgpu`. Enabled by default in `wgpu`. By @daxpedda in [#2890](https://github.com/gfx-rs/wgpu/pull/2890).
 - Implement `Clone` for `ShaderSource` and `ShaderModuleDescriptor` in `wgpu`. By @daxpedda in [#3086](https://github.com/gfx-rs/wgpu/pull/3086).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ Bottom level categories:
 
 #### General
 
+- Implement `Hash` for `DepthStencilState` and `DepthBiasState` in `wgpu-types`.
 - Add the `"wgsl"` feature, to enable WGSL shaders in `wgpu-core` and `wgpu`. Enabled by default in `wgpu`. By @daxpedda in [#2890](https://github.com/gfx-rs/wgpu/pull/2890).
 - Implement `Clone` for `ShaderSource` and `ShaderModuleDescriptor` in `wgpu`. By @daxpedda in [#3086](https://github.com/gfx-rs/wgpu/pull/3086).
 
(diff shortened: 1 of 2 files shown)


wgpu-types/src/lib.rs
```diff
@@ -3171,12 +3172,28 @@ impl DepthBiasState {
...
+        self.constant.hash(state);
+        self.slope_scale.to_bits().hash(state);
+        self.clamp.to_bits().hash(state);
+    }
+}
+
+impl PartialEq for DepthBiasState {
+    fn eq(&self, other: &Self) -> bool {
+        (self.constant == other.constant)
+            & (self.slope_scale == other.slope_scale)
```
cwfitzgerald, line 3186: This is a subtle one: if we use to_bits() for the hash implementation, we also need to use it for the eq implementation. One of the rules of `Hash` is that if `a == b` then `hash(a) == hash(b)` and for negative zero and other such values that is broken.

wgpu-types/src/lib.rs
```diff
@@ -3171,12 +3172,28 @@ impl DepthBiasState {
...
+            & (self.clamp == other.clamp)
```
cwfitzgerald, line 3187: Nit: use && not &

CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ Bottom level categories:
 
 #### General
 
+- Implement `Hash` for `DepthStencilState` and `DepthBiasState` in `wgpu-types`.
```
cwfitzgerald, line 54:
```suggestion
- Implement `Hash` for `DepthStencilState` and `DepthBiasState`.
```

We don't need to mention wgt as it's not part of our public interface, we re-export everything under the `wgpu` banner.

cwfitzgerald (changes requested):
Good stuff, some small comments

wgpu-types/src/lib.rs
killianc3: I see, so in this case what should we do? use to_bits() assuming the value never reaches zero and below or handle the problem case manually ?

wgpu-types/src/lib.rs
killianc3: Yes, I will do that.

CHANGELOG.md
killianc3: Okay, now I know.

wgpu-types/src/lib.rs
cwfitzgerald: We should just compare the bits as is and not worry about the weird extra cases.

cwfitzgerald (approved):
Great!

killianc3:
Thanks again !

Wumpf:
🙇 thank you. I was just *now* needing this 😄

PR #3093, opened 2022-10-11
Fix webgl srgb blit not setting the viewport correctly

Wumpf:
**Checklist**

- [x] Run `cargo clippy`.

**Description**
Srgb conversion on webgl assumed the viewport was correctly set. But what if it isn't? This can happen:
image

**Testing**
manual testing with an egui application

### DIFF
--- a/wgpu-hal/src/gles/web.rs
+++ b/wgpu-hal/src/gles/web.rs
@@ -171,6 +171,13 @@ impl Surface {
         ))?;
 
         if swapchain.format.describe().srgb {
+            // Important to set the viewport since we don't know in what state the user left it.
+            gl.viewport(
+                0,
+                0,
+                swapchain.extent.width as _,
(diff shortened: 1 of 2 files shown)


cwfitzgerald (approved):
Heh

wgpu-hal/src/gles/web.rs
```diff
@@ -171,6 +171,13 @@ impl Surface {
         ))?;
 
         if swapchain.format.describe().srgb {
+            // Important to set the viewport since we don't know in what state the user left it.
+            gl.viewport(
```
cwfitzgerald, line 175: Actually is scissor rect also a problem?

Wumpf: nope, because there is a `gl.disable(glow::SCISSOR_TEST);` a few lines below 😄

PR #3094, opened 2022-10-11
Make some errors not fatal

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

https://bugzilla.mozilla.org/show_bug.cgi?id=1791881

**Description**

Objects can be created with invalid parameters, putting them in an error state. When interacting with these objects, in the majority of cases validation should fail but wgpu should not panic.
`buffer_get_mapped_range` is one of these exceptions because the API user must not call it without getting consent from the status provided in the `map_async` callback.

Currently wgpu does panic when interacting with invalid resources in a few places and this PR starts fixing that. I did not go over all of them to make sure there is consensus before producing a large-ish patch, and to get the few changes needed to unblock fuzzing as soon as possible.

There two important changes in this patch:
 - in `hub.rs`, `Some(vacant)` and `None` are the same (it indicates the resource was not created or was already dropped) so they should both be fatal errors.
 - with the non-existent-or-freed error case out of the way, most places that call `handle_error_fatal` in `backend/direct.rs` should actually push the error into the error sink intsead of panicking.

Only the first change really affects Firefox, but I think that the second one better matches the spec and how Firefox uses wgpu-core.

**Testing**

Tests! I did it again.

### DIFF
--- a/wgpu/tests/resource_error.rs
+++ b/wgpu/tests/resource_error.rs
@@ -1,0 +1,53 @@
+use crate::common::{fail, initialize_test, valid, TestParameters};
+
+#[test]
+fn bad_buffer() {
+    // Create a buffer with bad parameters and call a few methods.
+    // Validation should fail but there should be not panic.
+    initialize_test(TestParameters::default(), |ctx| {
+        let buffer = fail(&ctx.device, || {
+            ctx.device.create_buffer(&wgpu::BufferDescriptor {
+                label: None,
+                size: 99999999,
+                usage: wgpu::BufferUsages::MAP_READ | wgpu::BufferUsages::STORAGE,
+                mapped_at_creation: false,
+            })
+        });
+
+        fail(&ctx.device, || {
+            buffer.slice(..).map_async(wgpu::MapMode::Write, |_| {})
+        });
+        fail(&ctx.device, || buffer.unmap());
+        valid(&ctx.device, || buffer.destroy());
+        valid(&ctx.device, || buffer.destroy());
+    });
+}
+
+#[test]
+fn bad_texture() {
+    // Create a texture with bad parameters and call a few methods.
+    // Validation should fail but there should be not panic.
+    initialize_test(TestParameters::default(), |ctx| {
+        let texture = fail(&ctx.device, || {
+            ctx.device.create_texture(&wgpu::TextureDescriptor {
+                label: None,
+                size: wgpu::Extent3d {
+                    width: 0,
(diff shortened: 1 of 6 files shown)


wgpu/tests/resource_error.rs
```diff
@@ -0,0 +1,61 @@
+use crate::common::{initialize_test, TestParameters};
+
+#[cfg(test)]
+// Run some code in an error scope and assert that validation failed.s
+fn fail(device: &wgpu::Device, callback: &mut dyn FnMut()) {
```
cwfitzgerald, line 5: This is great! Lets hoist this into common.

wgpu/tests/resource_error.rs
```diff
@@ -0,0 +1,61 @@
...
+        fail(&ctx.device, &mut || buffer.destroy());
+    });
+}
+
+#[test]
+fn bad_texture() {
+    // Create a texture with bad parameters and call a few methods.
+    // Validation should fail but there should be not panic.
+    initialize_test(TestParameters::default(), |ctx| {
+        ctx.device.push_error_scope(wgpu::ErrorFilter::Validation);
```
cwfitzgerald, line 39: Why can't this block also use `fail`

wgpu/tests/resource_error.rs
```diff
@@ -0,0 +1,61 @@
+use crate::common::{initialize_test, TestParameters};
+
+#[cfg(test)]
```
cwfitzgerald, line 3: Unneeded

wgpu/tests/resource_error.rs
cwfitzgerald: We can use `impl FnOnce()` and the callsite won't need a borrow.

cwfitzgerald (changes requested):
Tests make me very happy

wgpu/tests/resource_error.rs
nical: I did not bother with making it possible for fail() to return a value, but if this is to be hoisted, I should probably make an effort.

nical:
I updated the patch. `fail` can return values and is hoisted in common, there's a `valid` equivalent that asserts validation succeeded (useful when you want to assert that specific/individual calls pass validation).

I made a notable change to the error handling after check the spec: `buffer`/`texture.destroy` does not actually raise any error even when called multiple times.

cwfitzgerald (approved):
Format mad, but otherwise LGTM

Issue #2832, opened 2022-07-02
Helpful information for `BUFFER_BINDINGS_NOT_16_BYTE_ALIGNED`


kpreid:
**Is your feature request related to a problem? Please describe.**

1. Write a shader, uniform buffers, etc.
2. Try to run the program with the WebGL backend.
3. Get an error that tells you `BUFFER_BINDINGS_NOT_16_BYTE_ALIGNED` is unsupported, and the pipeline name, and nothing else.

**Describe the solution you'd like**
Provide diagnostic information to help fix the alignment problem, such as:

* The fact that this error is provoked by the _shader's_ code, not the size of CPU-provided buffers.
* The name or number of the binding that was the wrong size.
* The field layout of the struct type of that binding, if it is a struct.

All but the first of these improvements would require making the `MissingDownlevelFlags` error able to carry additional context, or some other kind of refactoring of the error types.

rparrett:
I am a bevy user experiencing this error while migrating to a version of bevy that uses wgpu 0.13, and I would definitely love to have any sort of idea what's going on.

cwfitzgerald:
The problem is that the size of your shader side struct must be aligned to size 16 - if it isn't, it needs to be padded out to be size 16

gilescope:
One of your shader structs isn't aligned. Not going to tell you which one. You have to find the needle in the haystack. Sad face.

cwfitzgerald:
@gilescope We should have the binding information available in the key of the hash map we're iterating through here, so a PR adding this information shouldn't be that difficult. https://github.com/gfx-rs/wgpu/blob/master/wgpu-core/src/device/mod.rs#L2857

PR #3099, opened 2022-10-13
Better error message

gilescope:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Closes https://github.com/gfx-rs/wgpu/issues/2832

**Description**
Previous error message was the generic downflags error message for `BUFFER_BINDINGS_NOT_16_BYTE_ALIGNED` that didn't highlight that if it was 16 bit aligned then this *is* supported. But it wasn't clear which shader group / bind was not aligned.

New error message:
```
      note: label = `cuboids_pipeline`
    shader group 1 binding 1 must be 16bit aligned as device does not support DownlevelFlags `BUFFER_BINDINGS_NOT_16_BYTE_ALIGNED`
```

Also have switched the code around to check the flag once as if it's set then we can skip the check entirely.

**Testing**
I've tested that I get the updated error locally in the browser.

### DIFF
--- a/wgpu-core/src/pipeline.rs
+++ b/wgpu-core/src/pipeline.rs
@@ -371,6 +371,8 @@ pub enum CreateRenderPipelineError {
         stage: wgt::ShaderStages,
         error: String,
     },
+    #[error("shader group {group} binding {binding} must be 16bit aligned as device does not support DownlevelFlags `BUFFER_BINDINGS_NOT_16_BYTE_ALIGNED`")]
+    UnalignedShader { group: u32, binding: u32 },
 }
 
 bitflags::bitflags! {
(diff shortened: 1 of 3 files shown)


gilescope:
Could state what size we found?

cwfitzgerald:
The more information the merrier!

kpreid:
Could the shader's `label` or entry point name be included too?

wgpu-core/src/pipeline.rs
```diff
@@ -371,6 +371,8 @@ pub enum CreateRenderPipelineError {
         stage: wgt::ShaderStages,
         error: String,
     },
+    #[error("shader group {group} binding {binding} must be 16bit aligned as device does not support DownlevelFlags `BUFFER_BINDINGS_NOT_16_BYTE_ALIGNED`")]
```
grovesNL, line 374: This should be bytes instead of bits

cwfitzgerald:
Poke from triage

PR #3051, opened 2022-09-27
Context dynamic dispatch

i509VCB:
**Checklist**

- [X] Run `cargo clippy`.
- [X] Run `RUSTFLAGS=--cfg=web_sys_unstable_apis cargo clippy --target wasm32-unknown-unknown` if applicable.
- [X] Add change to CHANGELOG.md. See simple instructions inside file.

**Description**
This pull request does most of the work required to allow wgpu backends to be set at runtime. In the future this would allow selection between webgl2 and webgpu on web and on native other backends (including my use case which crosses process boundaries).

What would need to be done after this is to solidify the internal Context trait and types so those can be publicly exported.

**Testing**
This pull request needs testing with crates across the ecosystem. The dynamic dispatch in the internals does pass the tests in wgpu, but best to be safe.

There hasn't been any testing for the web backend. There could be errors that may have been added during the dynamic dispatch internal mess. However the testing situation for the web has been grim for a while.

PR #3106, opened 2022-10-14
Lower some data into the wgpu-core storages

i509VCB:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
Related to #3051 and the work to move that pull request towards using Ids instead of Boxes.

**Description**
In order for all the assoicated types in the native backend to be ids we can convert directly, we need to lower some of the data associated with some types into the tracker instead of living next to the ids.

However this is a challenge to do and this pull request is here to explore the idea or find a better alternative. There are still some types which may need data lowered into the tracker or living in a type erased form along side ids. Namely RenderPass, ComputePass and RenderBundleEncoder carry a large amount of data along.

**Testing**
Local tests appear to run fine.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -4583,12 +4583,29 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         A::hub(self).command_buffers.label_for_resource(id)
     }
 
-    pub fn command_encoder_drop<A: HalApi>(&self, command_encoder_id: id::CommandEncoderId) {
+    pub fn command_encoder_drop<A: HalApi>(
+        &self,
+        command_encoder_id: id::CommandEncoderId,
+    ) -> Result<(), resource::DestroyError> {
         profiling::scope!("CommandEncoder::drop");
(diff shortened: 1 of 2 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -3598,10 +3598,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         A::hub(self).buffers.label_for_resource(id)
     }
 
-    pub fn buffer_destroy<A: HalApi>(
-        &self,
-        buffer_id: id::BufferId,
-    ) -> Result<(), resource::DestroyError> {
```
ErichDonGubler, line 3604: nitpick: It'd be nice to have a separate commit for moving `DestroyError` to a direct `use`. This would separate some mechanical changes from the more substantive changes in the `remove open field from CommandBuffer` commit.

i509VCB: I'll probably undo the import since every other type in that module has the qualified error type.

i509VCB:
Going to delay this temporarily since I have a solution which is working in #3051

cwfitzgerald:

PR #3107, opened 2022-10-14
Use cargo 1.64 workspace inheritance feature

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Description**
Now, dependencies used by more than one package use inheritance instead.

In the `[package]` table, three fields of `name`, `description`, and `version` remain independent. Although each crate `version` is now synchronized, they should be able to upgrade the version separately, in dummy, player two crates, the authors value is not the same, so it also remains independent.
    - In `dummy`, `player` crates, the `authors` field value are not the same, so they remain independent.
    - The value of  `homepage` field is "https://wgpu.rs/" in  wgpu crate and "https://github.com/gfx-rs/wgpu" in  other crate, I'm not sure if this is intentional or not, but now it is inherited as  "https://wgpu.rs/" value

`deno_webgpu` and `cts_runner` are not workspace members, so inheritance cannot be used.

**Testing**
Tested locally on macOS, Windows11 and Android 12.

### DIFF
--- a/dummy/Cargo.toml
+++ b/dummy/Cargo.toml
@@ -1,16 +1,13 @@
 [package]
 name = "dummy"
-version = "0.1.0"
-authors = [
-    "Dzmitry Malyshau <kvark@mozilla.com>",
-]
-edition = "2021"
-license = "MIT OR Apache-2.0"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
 publish = false
 
 [features]
 
 [dependencies.wgc]
-path = "../wgpu-core"
-package = "wgpu-core"
+workspace = true
 features = ["serial-pass", "trace"]
--- a/wgpu-core/Cargo.toml
+++ b/wgpu-core/Cargo.toml
@@ -1,13 +1,13 @@
 [package]
 name = "wgpu-core"
-version = "0.14.0"
-authors = ["wgpu developers"]
-edition = "2021"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
 description = "WebGPU core logic on wgpu-hal"
-homepage = "https://github.com/gfx-rs/wgpu"
-repository = "https://github.com/gfx-rs/wgpu"
-keywords = ["graphics"]
-license = "MIT OR Apache-2.0"
+homepage.workspace = true
+repository.workspace = true
+keywords.workspace = true
+license.workspace = true
 
 [lib]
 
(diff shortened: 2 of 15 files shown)


dummy/Cargo.toml
```diff
@@ -2,15 +2,14 @@
 name = "dummy"
 version = "0.1.0"
 authors = [
-    "Dzmitry Malyshau <kvark@mozilla.com>",
+    "Dzmitry Malyshau <kvark@mozilla.com>"
```
cwfitzgerald, line 5: We can make these inherent both version and authors

wgpu-core/Cargo.toml
```diff
@@ -29,55 +29,48 @@ wgsl = ["naga/wgsl-in"]
 vulkan-portability = ["hal/vulkan"]
 
 [dependencies]
-arrayvec = "0.7"
-bitflags = "1.0"
+arrayvec.workspace = true
+bitflags.workspace = true
 bit-vec = "0.6"
```
cwfitzgerald, line 34: Any reason we wouldn't just bring all dependencies up to the workspace level and inherit them all?

cwfitzgerald (changes requested):
Cool!

dummy/Cargo.toml
jinleili: Does the `version` of deno_webgpu also inherit the same field values?

cwfitzgerald: No, deno-webgpu is kinda a separate project but resides in repo so we can test with it

wgpu-core/Cargo.toml
jinleili: I'm just not sure that if should put dependencies that are only used in a certain crate into the workspace. There is no other reason, I can do it.

ErichDonGubler:
Not a blocker, but I wanted to note that this PR would be a breaking change, because it would change the [repo's current MSRV](https://github.com/gfx-rs/wgpu/commit/d5f1941fea019b4061459bf465afcb06696b2608#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R34-R36) to Rust 1.64, AIUI:

> Minimum Supported Rust Version is **1.62**. It is enforced on CI (in "/.github/workflows/ci.yml") with `RUST_VERSION` variable. This version can only be upgraded in breaking releases.

cwfitzgerald:
@ErichDonGubler any changes in either wgpu or naga going into master can break - we don't really do patch releases outside of major bugs, and when we patch we go into the specific version (eg. `v0.14`) branch.

cwfitzgerald (approved):
This is totally wacky wild, but I love it!

PR #3111, opened 2022-10-15
Update present_with_transaction

xiaopengli89:
**Checklist**

- [x] Run `cargo clippy`.

**Description**
Set `present_with_transaction` according to `PresentMode`.

### DIFF
--- a/wgpu-hal/src/metal/surface.rs
+++ b/wgpu-hal/src/metal/surface.rs
@@ -186,6 +186,7 @@ impl crate::Surface<super::Api> for super::Surface {
             wgt::PresentMode::Immediate => false,
             m => unreachable!("Unsupported present mode: {m:?}"),
         };
+        self.present_with_transaction = display_sync;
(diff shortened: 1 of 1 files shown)


cwfitzgerald:
I don't think this is what we actually want. Presenting with transaction is a much more intense form of synchronization which synchronizes with CoreAnimation.

xiaopengli89:
Is there currently a way to configure wgpu to use CoreAnimation? Because I found that there is a memory leak problem here: https://github.com/gfx-rs/wgpu/blob/ce081796c80675b56b3874089238789e2f98893a/wgpu-hal/src/metal/mod.rs#L381 it is likely to be a metal driver's bug. When there is no better solution, I hope to avoid the use of this api

image

grovesNL:
@xiaopengli89 does this leak happen in regular Metal Objective-C or Swift examples? It would be great to verify if this was a actually a Metal bug or if we're accidentally holding an extra reference somewhere (i.e., we have an auto release pool missing somewhere other than `present_drawable`)

PR #3113, opened 2022-10-16
enable doc_auto_cfg for docs.rs

i509VCB:
**Checklist**

- [X] Run `cargo clippy`.
- [ ] Add change to CHANGELOG.md. See simple instructions inside file. (Not really applicable?)

**Description**
This should expose more feature labels in the generated documentation and removes the needs for the manually labeling the features for a type, function or enum variants.

wgpu-hal cannot enable every feature because the dx11, dx12 and metal backends do not cross compile on x86_64-unknown-linux-gnu. For metal, the objc_exception dependency does not compile because it invokes clang. The metal-rs objc2 migration appears to fix that bug. I have not looked into dx11 and dx12 yet.

**Testing**
Documentation labels appear correct when running the nightly rustdoc (like docs.rs does) with the cfg feature enabled.

### DIFF
--- a/wgpu-hal/Cargo.toml
+++ b/wgpu-hal/Cargo.toml
@@ -10,6 +10,15 @@ keywords = ["graphics"]
 license = "MIT OR Apache-2.0"
 rust-version = "1.60"
 
+[package.metadata.docs.rs]
+# Ideally we would enable all the features.
+#
+# However the metal features fails to be documented because the docs.rs runner cross compiling under
+# x86_64-unknown-linux-gnu and metal-rs cannot compile in that environment at the moment. The same applies
+# with the dx11 and dx12 features.
+features = ["vulkan", "gles", "renderdoc"]
+rustdoc-args = ["--cfg", "docsrs"]
+
 [lib]
 
 [features]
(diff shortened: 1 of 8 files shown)


wgpu-hal/Cargo.toml
```diff
@@ -61,6 +70,8 @@ glow = { git = "https://github.com/grovesNL/glow/", rev = "c8a011fcd57a5c68cc917
 # backend: Dx12
 bit-set = { version = "0.5", optional = true }
 range-alloc = { version = "0.1", optional = true }
+winapi = { version = "0.3", features = ["libloaderapi", "windef", "winuser", "dcomp"] }
+native = { package = "d3d12", version = "0.5.0", features = ["libloading"], optional = true }
```
cwfitzgerald, line 74: Why do these need to be added to general dependencies? We only need them on windows.

cwfitzgerald (changes requested):
Good stuff, have a comment.

We should also change the documentation CI to set the docsrs cfg and to build on nightly so our master docs get this too.

wgpu-hal/Cargo.toml
i509VCB: I guess I forgot to move these back, I was testing the documentation and tried to see if I could get these to build on Linux where docs.rs generates docs

cwfitzgerald (approved):

Issue #2985, opened 2022-08-26
Work Group Size Error is Confusing


mwalczyk:
I am working on a project involving a compute shader on MacOS and have set the local work group size in the shader like so:

`@workgroup_size(8, 8, 8)`

However, this results in an error that reads: 

`shader entry point current workgroup size [8, 8, 8] must be less or equal to [256, 256, 64] of total 256`

After talking to some folks on Matrix, it seems that what the error is trying to say is: the _total number of invocations in a work group_ cannot exceed 256. In my case, the total # of invocations is 8x8x8 = 512, which is greater than 256, so it fails to compile. I'm not sure how to best word the error message, but I think it's a little confusing as-is, since [8, 8, 8] is clearly less than [256, 256, 64] across all dimensions.

kpreid:
Idea for improvement: add "of total 512" to the first half — show the value derived from the input — so that it is immediately obvious what is too big because it parallels the second half specifying the limits.

"Total" doesn't seem like a great description but I don't have any suggestions I like very much. "Product" would specify how the number was gotten better? What if the dimensions were printed with "×"es between?

haraldreingruber:
I would like to fix this.

Would a message like `shader entry point current workgroup size [8, 8, 8] of total 512 must be less or equal to [256, 256, 64] and of total 256` be an improvement? @mwalczyk

By the way, I would be motivated to work on a few PRs for the Hacktoberfest challenge. 
@kpreid would you be open to adding the “hacktoberfest” topic to the repo? https://hacktoberfest.com/participation/#maintainers

haraldreingruber:
I've went ahead and proposed #3139.

The idea from @kpreid is definitely an improvement to the current message. If anybody shares new ideas, we can improve further easily....

haraldreingruber:
This issue can be closed because #3139 was merged to master. @cwfitzgerald

PR #3139, opened 2022-10-25
Improve compute shader validation error message

haraldreingruber:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Potential improvement for #2985 

**Description**
The previous error message was easily misunderstood, as it didn't say explicitly that each dimension and the total number must be below the bounds.

For example, instead of:

    shader entry point current workgroup size [8, 8, 8] must be less or equal to [256, 256, 64] of total 256

after applying this PR the error is:

    shader entry point current workgroup size [8, 8, 8] of total 512 must be less or equal to [256, 256, 64] and of total 256

Thanks to @mwalczyk for the original improvement idea 🙏

**Testing**
In `shader.wgsl` of the `hello-compute` example, I've changed the `@workgroup_size(1)` to `@workgroup_size(8, 8, 8)`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@ Bottom level categories:
 
 - Convert all `Default` Implementations on Enums to `derive(Default)`
 - Implement `Default` for `CompositeAlphaMode`
-- Improve compute shader validation error message
+- Improve compute shader validation error message. By @haraldreingruber in [#3139](https://github.com/gfx-rs/wgpu/pull/3139)
 
 #### WebGPU
 - Implement `queue_validate_write_buffer` by @jinleili in [#3098](https://github.com/gfx-rs/wgpu/pull/3098)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ Bottom level categories:
 - DX11
 - GLES
 - WebGPU
-- Enscripten
+- Emscripten
 - Hal
 -->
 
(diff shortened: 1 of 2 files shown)


wgpu-core/src/validation.rs
```diff
@@ -236,10 +236,11 @@ pub enum StageError {
     #[error("shader module is invalid")]
     InvalidModule,
     #[error(
-        "shader entry point current workgroup size {current:?} must be less or equal to {limit:?} of total {total}"
+        "shader entry point current workgroup size {current:?} of total {current_total} must be less or equal to {limit:?} and of total {total}"
```
cwfitzgerald, line 239:
```suggestion
        "shader entry point's workgroup size {current:?} ({current_total} total invocations) must be less or equal to the per-dimension limits {limit:?} and the total invocation limit {total}"
```

CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ Bottom level categories:
 
 - Convert all `Default` Implementations on Enums to `derive(Default)`
 - Implement `Default` for `CompositeAlphaMode`
+- Improve compute shader validation error message
```
cwfitzgerald, line 49: Don't forget to credit yourself and the PR

cwfitzgerald (changes requested):
Thank you! Suggested some alternative wording to help make the error message even more clear.

wgpu-core/src/validation.rs
haraldreingruber: Thanks for the great improvement!

cwfitzgerald (approved):
Thanks!

Issue #2910, opened 2022-07-23
Can't set MSAA samples beyond 4


swiftcoder:
**Description**

The Vulkan capabilities viewer says that my hardware/driver supports MSAA samples up to 8 for all targets:

![image](https://user-images.githubusercontent.com/4706210/180598869-998f22cb-8faa-4216-ba99-914e77102499.png)

Setting MSAA samples to 1 or 4 works correctly, however increasing the samples to 8 causes the following error:

```
     Running `C:\...\msaa-line.exe`
Using NVIDIA GeForce GTX 1080 (Vulkan)
[2022-07-23T09:22:55Z ERROR wgpu::backend::direct] Handling wgpu errors as fatal by default
thread 'main' panicked at 'wgpu error: Validation Error

Caused by:
    In a RenderPass
      note: encoder = `<CommandBuffer-(0, 1, Vulkan)>`
    In a pass parameter
      note: command buffer = `<CommandBuffer-(0, 1, Vulkan)>`
    attachment's sample count 8 is invalid

', wgpu\src\backend\direct.rs:2391:5
```

**Repro steps**
Modify the `msaa-line` sample to set the number of samples to 8, and run it.

**Expected vs observed behavior**
Clearly describe what you get, and how it goes across your expectations.

I appreciate that the current web spec only supports sample counts of 1 or 4, but I don't think this limitation should be enforced as a strict validation (especially for desktop builds, where the driver likely supports at least 8, and one can always query against the underlying API directly).

**Extra materials**
This appears to be failing a [hardcoded validation](https://github.com/gfx-rs/wgpu/blob/9f82504d8a3b6e234599b19cc35c299bcec8b138/wgpu-core/src/command/render.rs#L725-L727).

All other hardcoded validations I can find only check that the number of samples are less than 32.
 
**Platform**
Information about your OS, version of `wgpu`, your tech stack, etc.

Windows 10, Vulkan, v0.13.1 of wgpu

cwfitzgerald:
Thanks for the issue! We definitely want to add this, and the way I think that this should manifest is by extending the TextureFormatFeatutes and splitting multisampling into three flags, 2, 4, and 8 sample flags for each texture.

PR #3140, opened 2022-10-26
Added support for MSAA(x2,x8) beyond WEBGPU restrictions for native a…

39ali:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Closes [Can't set MSAA samples beyond 4](https://github.com/gfx-rs/wgpu/issues/2910) .

**Description**
Native adapters can now use MSAA x2 and x8 if it's supported , previously only x1 and x4 were supported .

**Testing**
Tested locally on chrome canary , macOS , win10


wgpu-types/src/lib.rs
```diff
@@ -1656,6 +1674,8 @@ pub struct TextureFormatFeatures {
     pub allowed_usages: TextureUsages,
     /// Additional property flags for the format.
     pub flags: TextureFormatFeatureFlags,
+    /// supported sample count
+    pub sample_count: TextureFormatSampleCountFlags,
```
cwfitzgerald, line 1678: We don't need another bitflags for this - we can use the existing `TextureFormatFeatureFlags::MULTISAMPLE` and split it into 3 flags: `MULTISAMPLE_X2` `MULTISAMPLE_X4` and `MULTISAMPLE_X8` (as x1 multisampling is guarenteed for all formats that you can use at all)

cwfitzgerald (changes requested):
PR on the whole looks good, but I think there's a cleaner strategy for the flags which will re-arrange things a bit

wgpu-types/src/lib.rs
39ali: That does look cleaner , I'll refactor code to reflect the suggested changes .

39ali:
Hi @cwfitzgerald , let me know if it needs any further changes

39ali:
Hi @cwfitzgerald , let me know what you think!

wgpu-core/src/resource.rs
```diff
@@ -515,6 +515,8 @@ pub enum CreateTextureError {
     MultisampledNotRenderAttachment,
     #[error("Texture format {0:?} can't be used due to missing features.")]
     MissingFeatures(wgt::TextureFormat, #[source] MissingFeatures),
+    #[error("Sample count {0} is invalid")]
```
cwfitzgerald, line 518: Let's be a bit more specific here:

```suggestion
    #[error("Sample count {0} is not supported by format {1} on this device. It may be supported by your adapter through the TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES feature")]
```

wgpu-hal/src/dx12/adapter.rs
```diff
@@ -447,12 +447,37 @@ impl crate::Adapter<super::Api> for super::Adapter {
...
+            Flags: d3d12::D3D12_MULTISAMPLE_QUALITY_LEVELS_FLAG_NONE,
+            NumQualityLevels: 0,
+        };
+
+        let mut set_sample_count = |sc: u32, tfc: Tfc| {
+            ms_levels.SampleCount = sc;
+
+            if self.device.CheckFeatureSupport(
+                d3d12::D3D12_FEATURE_MULTISAMPLE_QUALITY_LEVELS,
+                &mut ms_levels as *mut _ as *mut _,
```
cwfitzgerald, line 468: (This isn't your fault, you're just copying the pattern of surrounding code)

We should use `<*mut c_void>::cast(&mut ms_levels)` this is cleaner and makes it clearer what's going on.

wgpu-hal/src/gles/adapter.rs
```diff
@@ -666,10 +665,10 @@ impl crate::Adapter<super::Api> for super::Adapter {
         let empty = Tfc::empty();
         let base = Tfc::COPY_SRC | Tfc::COPY_DST;
         let unfilterable = base | Tfc::SAMPLED;
-        let depth = base | Tfc::SAMPLED | Tfc::MULTISAMPLE | Tfc::DEPTH_STENCIL_ATTACHMENT;
+        let depth = base | Tfc::SAMPLED | Tfc::MULTISAMPLE_X4 | Tfc::DEPTH_STENCIL_ATTACHMENT;
```
cwfitzgerald, line 668: We need to get glow::MAX_SAMPLES with `gl.get_parameter_i32` and then use the value to populate if 2, 4, or 8 is available.

wgpu-hal/src/metal/adapter.rs
```diff
@@ -57,12 +57,12 @@ impl crate::Adapter<super::Api> for super::Adapter {
             }
         };
         let msaa_desktop_if = if pc.msaa_desktop {
-            Tfc::MULTISAMPLE
+            Tfc::MULTISAMPLE_X4
```
cwfitzgerald, line 60: We need to call https://developer.apple.com/documentation/metal/mtldevice/1433355-supportstexturesamplecount and store it in our private capabilities and use that to determine the multisample levels supported.

wgpu/examples/msaa-line/main.rs
```diff
@@ -32,6 +32,7 @@ struct Example {
     sample_count: u32,
     rebuild_bundle: bool,
     config: wgpu::SurfaceConfiguration,
+    max_sample_count: u32,
```
cwfitzgerald, line 35: I think we just need to store the flags, some devices support 4 but not 2, etc.

cwfitzgerald (changes requested):
This is great stuff, have some comments, mostly around feature querying.

Looking forward to having this in!

wgpu-hal/src/metal/adapter.rs
39ali: Do we still need to check for pc.msaa_desktop and pc.msaa_apple7 or can we just use the flag like so : 
`    let msaa_desktop_if = pc.sample_count_mask;`
     `let msaa_apple7x_if =pc.sample_count_mask;
  `?

wgpu/examples/msaa-line/main.rs
39ali: I'm not sure im getting it , we get the value of `max_sample_count` from `adapter.get_texture_format_features()` so we know for a fact that it's supported

wgpu-hal/src/gles/adapter.rs
39ali: I read through the GL doc but I still have a question :
does gl.get_parameter_i32() value means that smaller sampling count is supported? , for example : 
if it returns 32 , does that means that 16,8 etc is supported ? .

cwfitzgerald: Correct

wgpu-hal/src/metal/adapter.rs
cwfitzgerald: Probably calling `supportstexturesamplecount` is probably good enough and we don't need a general flag.

wgpu/examples/msaa-line/main.rs
cwfitzgerald: We don't get the max sample count, we get the array of supported samples. If x4 is supported, but x2 isn't, this won't catch that case.

39ali: I'm storing the maximum sample count in `max_sample_count` which I'm getting from the array I get from `adapter.get_texture_format_features()` , and then I'm just flipping between `1` and `max_sample_count` using the arrow keys , this is not used to figure out the supported sample count , I think you're thinking of a different use for this variable ? , if you think it's more suitable , i could rename it to `supproted_sample_counts` which will store the supported sample flags , and then use it in the arrow keys if statements to figure out the max sample count .

cwfitzgerald: Ohh, I misunderstood. The msaa example _used_ to cycle through all MSAA settings supported, but that functionality apparently changed9 was changed once 4x only was being enforced.

39ali: Oh , thanks for the clarification !

39ali:
hey @cwfitzgerald, please take a look !.

wgpu-hal/src/metal/adapter.rs
```diff
@@ -110,7 +102,7 @@ impl crate::Adapter<super::Api> for super::Adapter {
             | Tf::Rgba8Sint
             | Tf::Rgba16Uint
             | Tf::Rgba16Sint => {
-                read_write_tier2_if | Tfc::STORAGE | Tfc::COLOR_ATTACHMENT | Tfc::MULTISAMPLE
+                read_write_tier2_if | Tfc::STORAGE | Tfc::COLOR_ATTACHMENT | Tfc::MULTISAMPLE_X4
```
cwfitzgerald, line 105: All X4s should should be `msaa_count`

wgpu-hal/src/metal/adapter.rs
```diff
@@ -276,7 +267,11 @@ impl crate::Adapter<super::Api> for super::Adapter {
             }
         };
 
-        Tfc::COPY_SRC | Tfc::COPY_DST | Tfc::SAMPLED | extra
+        Tfc::COPY_SRC
+            | Tfc::COPY_DST
+            | Tfc::SAMPLED
+            | extra
+            | self.shared.private_caps.sample_count_mask
```
cwfitzgerald, line 274: This should only be added as part of the `extra` flags as only some formats support MSAA at all

wgpu-types/src/lib.rs
```diff
@@ -1629,19 +1629,41 @@ bitflags::bitflags! {
...
+        use TextureFormatFeatureFlags as tfsc;
+
+        if count == 1
+            || (count == 2 && self.contains(tfsc::MULTISAMPLE_X2))
+            || (count == 4 && self.contains(tfsc::MULTISAMPLE_X4))
+            || (count == 8 && self.contains(tfsc::MULTISAMPLE_X8))
+        {
+            return true;
+        }
+        false
```
cwfitzgerald, line 1666: This can be restructured to be a  match on count, then inside the branches for 1, 2, 4, and 8, is the `self.contains` call. The fallthrough branch can return false.

cwfitzgerald (changes requested):
Very close, some small stuff

cwfitzgerald (approved):
Thanks!

PR #3146, opened 2022-10-28
Avoid integer overflow on multiplication in write_texture.

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Found by 

https://bugzilla.mozilla.org/show_bug.cgi?id=1791809

**Description**

With large depth values in `copy_size`, `validate_linear_texture_data` can run into integer overflows.
This is avoided by validating the copy depth before calling `validate_linear_texture_data` in `queue_write_texture`.
The other two `validate_linear_texture_data` call sites already have the copy size validated beforehand.

**Testing**

I can add a test when I come back from vacation a week from now.

### DIFF
--- a/wgpu-core/src/device/queue.rs
+++ b/wgpu-core/src/device/queue.rs
@@ -600,6 +600,19 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         let (selector, dst_base, texture_format) =
             extract_texture_selector(destination, size, &*texture_guard)?;
         let format_desc = texture_format.describe();
+
+        let dst = texture_guard.get_mut(destination.texture).unwrap();
+        if !dst.desc.usage.contains(wgt::TextureUsages::COPY_DST) {
+            return Err(
+                TransferError::MissingCopyDstUsageFlag(None, Some(destination.texture)).into(),
+            );
+        }
+
+        // Note: Doing the copy range validation early is important because ensures that the
+        // dimensions are not going to cause overflow in other parts of the validation.
+        let (hal_copy_size, array_layer_count) =
+            validate_texture_copy_range(destination, &dst.desc, CopySide::Destination, size)?;
+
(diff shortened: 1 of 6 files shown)


nical:
I started writing the test and that was a good thing as it showed that fixing the cause of the panic lets us panic in backend_direct.rs instead which treats validation errors as fatal in queue_write_texture so that has to be fixed as well, I'll do it in a week or so.

wgpu-core/src/device/queue.rs
```diff
@@ -600,6 +600,17 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
         let format_desc = texture_format.describe();
+
+        let dst = texture_guard.get_mut(destination.texture).unwrap();
+        if !dst.desc.usage.contains(wgt::TextureUsages::COPY_DST) {
+            return Err(
+                TransferError::MissingCopyDstUsageFlag(None, Some(destination.texture)).into(),
+            );
+        }
+
+        let (hal_copy_size, array_layer_count) =
```
cwfitzgerald, line 611: Could you add a note here about how the order is important here to avoid overflows?

cwfitzgerald:
I also think we may have the same problem in copy_texture_*? Or do those use the right order?

nical:
`copy_texture_to_buffer` already validates the size by calling `validate_texture_copy_range` before `validate_linear_texture_data`. The other place is `copy_texture_to_texture` however it doesn't call `validate_linear_texture_data`. I don't see risky arithmetic before the `validate_copy_range` calls but these are easy to miss. The fuzzers will find issues if any.

nical:
I added a commit that adds a ref to the error sink in the queue id, the same way it is done for device and similar handles, in order for relevant queue methods to be able handle errors gracefully and get the new test to pass.

cwfitzgerald (approved):
Great stuff, as always!

Issue #2722, opened 2022-06-02
DXC Integration in DX12 Backend


cwfitzgerald:
**Is your feature request related to a problem? Please describe.**

We currently use fxc (aka d3dcompiler) to compile our HLSL to DXBC. It has a couple issues:
- Completely unmaintained, with many issues
- Extremely slow
- DXBC is limited to SM5.1 which limits the features we can offer.
  - Subgroup Ops
  - Native float16

While our end goal is to have naga have a DXIL backend, that is a lot of work and a long way away, so an intermediate step is to integrate with DXC.

Because DXC is a runtime dependency, we still need to be able to fall back to FXC if DXC isn't available, and to get these features users will need to ship dxcompiler.dll with their application. 

**Describe the solution you'd like**

Integration with DXC, possibly through the [hassle-rs](https://crates.io/crates/hassle-rs) crate.

**Describe alternatives you've considered**

Waiting for a DXIL backend for naga. No C code but will take a long while.

**Additional context**

See this naga issue for some of the problems we've needed to work around to get FXC to work https://github.com/gfx-rs/naga/pull/1918.

PR #3147, opened 2022-10-29
DXC Integration in DX12 Backend

39ali:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
[https://github.com/gfx-rs/wgpu/issues/2722](https://github.com/gfx-rs/wgpu/issues/2722)

**Description**
DXC will be the default shader compiler for the d3d12 backend , I'm not sure about the addition of `enum ShaderDXIL` , the other solution would be to convert the `ID3DBlob` to `Vec<u8>` which would require a malloc+copy and freeing the blob .  

**Testing**

### DIFF
--- a/wgpu-hal/Cargo.toml
+++ b/wgpu-hal/Cargo.toml
@@ -27,7 +27,7 @@ metal = ["naga/msl-out", "block", "foreign-types"]
 vulkan = ["naga/spv-out", "ash", "gpu-alloc", "gpu-descriptor", "libloading", "smallvec"]
 gles = ["naga/glsl-out", "glow", "egl", "libloading"]
 dx11 = ["naga/hlsl-out", "native", "libloading", "winapi/d3d11", "winapi/d3d11_1", "winapi/d3d11_2", "winapi/d3d11sdklayers", "winapi/dxgi1_6"]
-dx12 = ["naga/hlsl-out", "native", "bit-set", "range-alloc", "winapi/d3d12", "winapi/d3d12shader", "winapi/d3d12sdklayers", "winapi/dxgi1_6"]
+dx12 = ["naga/hlsl-out", "native", "bit-set", "range-alloc", "winapi/d3d12", "winapi/d3d12shader", "winapi/d3d12sdklayers", "winapi/dxgi1_6", "hassle-rs"]
 renderdoc = ["libloading", "renderdoc-sys"]
 emscripten = ["gles"]
 
@@ -67,6 +67,7 @@ glow = { workspace = true, optional = true }
 # backend: Dx12
 bit-set = { workspace = true, optional = true }
 range-alloc = { workspace = true, optional = true }
+hassle-rs = { workspace = true, optional = true } 
 
 [dependencies.wgt]
 workspace = true
--- a/wgpu-hal/src/dx12/device.rs
+++ b/wgpu-hal/src/dx12/device.rs
@@ -239,6 +239,9 @@ impl super::Device {
             compile_flags.push("-Od"); /* d3dcompiler::D3DCOMPILE_DEBUG */
         }
 
+        // This closure gets called if `hassle_rs::compile_hlsl() fails due to a `hassle_rs::HassleError::LoadLibraryError`,
+        //
+        //  This usually means that `dxcompiler.dll` and/or `dxil.dll` could not be found in the current execution environment.
         let load_shader_fxc = || {
             let mut shader_data = native::Blob::null();
             let mut compile_flags = d3dcompiler::D3DCOMPILE_ENABLE_STRICTNESS;
@@ -262,22 +265,22 @@ impl super::Device {
             let hr = unsafe {
                 profiling::scope!("d3dcompiler::D3DCompile");
                 d3dcompiler::D3DCompile(
-                    source.as_ptr() as *const _,
+                    source.as_ptr().cast(),
                     source.len(),
-                    source_name.as_ptr() as *const i8,
+                    source_name.as_ptr().cast(),
                     ptr::null(),
                     ptr::null_mut(),
                     raw_ep.as_ptr(),
-                    full_stage.as_ptr() as *const i8,
+                    full_stage.as_ptr().cast(),
                     compile_flags,
                     0,
-                    shader_data.mut_void() as *mut *mut _,
-                    error.mut_void() as *mut *mut _,
+                    shader_data.mut_void().cast(),
+                    error.mut_void().cast(),
                 )
             };
 
             match hr.into_result() {
-                Ok(()) => (Ok(super::ShaderDXIL::FXC(shader_data)), log::Level::Info),
+                Ok(()) => (Ok(super::ShaderDXIL::Fxc(shader_data)), log::Level::Info),
                 Err(e) => {
                     let mut full_msg = format!("D3DCompile error ({})", e);
                     if !error.is_null() {
@@ -311,10 +314,10 @@ impl super::Device {
                 raw_ep.to_str().unwrap(),
                 &full_stage,
                 &compile_flags,
-                &vec![],
+                &[],
             )
         }) {
-            Ok(dxil) => (Ok(super::ShaderDXIL::DXC(dxil)), log::Level::Info),
+            Ok(dxil) => (Ok(super::ShaderDXIL::Dxc(dxil)), log::Level::Info),
             Err(err) => match err {
                 hassle_rs::HassleError::Win32Error(err) => {
                     let err_str = i32::into_result(err.0).unwrap_err();
@@ -1385,7 +1388,7 @@ impl crate::Device<super::Api> for super::Device {
                 shader_stages |= wgt::ShaderStages::FRAGMENT;
                 self.load_shader(stage, desc.layout, naga::ShaderStage::Fragment)?
             }
-            None => crate::dx12::ShaderDXIL::DXC(vec![]),
+            None => crate::dx12::ShaderDXIL::Dxc(vec![]),
         };
 
         let mut vertex_strides = [None; crate::MAX_VERTEX_BUFFERS];
@@ -1459,12 +1462,12 @@ impl crate::Device<super::Api> for super::Device {
         let raw_desc = d3d12::D3D12_GRAPHICS_PIPELINE_STATE_DESC {
             pRootSignature: desc.layout.shared.signature.as_mut_ptr(),
             VS: match blob_vs {
-                super::ShaderDXIL::DXC(ref blob_vs) => *native::Shader::from_raw(&blob_vs),
-                super::ShaderDXIL::FXC(blob_vs) => *native::Shader::from_blob(blob_vs),
+                super::ShaderDXIL::Dxc(ref blob_vs) => *native::Shader::from_raw(blob_vs),
+                super::ShaderDXIL::Fxc(blob_vs) => *native::Shader::from_blob(blob_vs),
             },
             PS: match blob_fs {
-                super::ShaderDXIL::DXC(ref blob_fs) => *native::Shader::from_raw(&blob_fs),
-                super::ShaderDXIL::FXC(blob_fs) => {
+                super::ShaderDXIL::Dxc(ref blob_fs) => *native::Shader::from_raw(blob_fs),
+                super::ShaderDXIL::Fxc(blob_fs) => {
                     if blob_fs.is_null() {
                         *native::Shader::null()
                     } else {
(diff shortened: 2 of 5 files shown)


wgpu-hal/src/dx12/device.rs
```diff
@@ -221,64 +222,147 @@ impl super::Device {
             .map(|name| ffi::CString::new(name.as_str()).unwrap())
             .map_err(|e| crate::PipelineError::Linkage(stage_bit, format!("{}", e)))?;
 
-        let mut shader_data = native::Blob::null();
-        let mut error = native::Blob::null();
-        let mut compile_flags = d3dcompiler::D3DCOMPILE_ENABLE_STRICTNESS;
+        let empty_str = "";
```
notgull, line 225: Is it necessary to make `empty_str` its own variable here? In fact, you could probably make this entire block:

```rust
let source_name = stage.module.raw_name.as_ref().and_then(|cstr| cstr.to_str()).unwrap_or("");
```

wgpu-hal/src/dx12/device.rs
```diff
@@ -221,64 +222,147 @@ impl super::Device {
...
-                (
-                    Err(crate::PipelineError::Linkage(stage_bit, full_msg)),
-                    log::Level::Warn,
-                )
             }
         };
 
+        // hassle_rs doesn't check for `dxil.dll` if you only call `hassle_rs::compile_hlsl` but calling hassle_rs::Dxil::new will .
+        profiling::scope!("hassle_rs::compile_hlsl");
+        let (result, log_level) = match hassle_rs::Dxil::new(None).and(hassle_rs::compile_hlsl(
```
notgull, line 305: You may want to make this `and_then` so that `hassle_rs::compile_hlsl` isn't called if `Dxil` fails.

39ali: nice catch !

wgpu-hal/src/dx12/device.rs
```diff
@@ -1092,7 +1179,7 @@ impl crate::Device<super::Api> for super::Device {
             },
             bind_group_infos,
             naga_options: hlsl::Options {
-                shader_model: hlsl::ShaderModel::V5_1,
```
Elabajaba, line 1095: Does bumping this to 6.0 remove support for cards that support shadermodel 5.1?

According to techpowerup's gpu database that would mean removing support for Nvidia's Kepler (700 series, and some laptop 900 series) and earlier, as well as AMD's 1st gen GCN (old 7000 series, which had some cards rebranded and rereleased though the 200 and 300 series, and even some OEM only models in the 400 series and a single OEM only card in the 500 series).

Afaik all of those cards have had their driver support dropped at this point.

Elabajaba: Looked into this a bit, DXC only supports SM6.0 and higher, and FXC only supports up to SM5.1. DXC will upgrade anything it compiles below SM6.0 to SM6.0.

Do you know how other engines handle this, or what happens if you try running SM6.0 code on a SM5.1 gpu?
https://github.com/microsoft/DirectXShaderCompiler/issues/4038 seems like it's probably fine?

39ali: If I'm understanding this correctly according to the (docs)[https://learn.microsoft.com/en-us/windows/win32/direct3d12/hardware-feature-levels], if a device supports `D3D_FEATURE_LEVEL_11_0` it should support shader model `6.0` and `5.1`, if this is incorrect, we can call CheckFeatureSupport(D3D12_FEATURE_SHADER_MODEL..) to check for shader support and choose FXC or DXC accordingly. 
unfortunately i don't have an old GPU to test this .

39ali: It seems that only DXC CLI will upgrade SM5.1 to SM6.0, DXC lib would return an error, this won't be an issue if we take the solution i previously suggested.

Elabajaba: Looks like I was wrong about hardware support. GPU companies seemed to have added support for higher shadermodels in driver updates according to the DXC repo.

```
NVIDIA's r396 drivers (r397.64 and later) provide release mode support for DXIL 1.1 and Shader Model 6.1 on Win10 1709 and later, and experimental mode support for DXIL 1.2 and Shader Model 6.2 on Win10 1803 and later. These drivers also support DXR in experimental mode.

AMD’s driver (Radeon Software Adrenalin Edition 18.4.1 or later) provides release mode support for DXIL 1.1 and Shader Model 6.1.

Intel's 15.60 drivers (15.60.0.4849 and later) support release mode for DXIL 1.0 and Shader Model 6.0 as well as release mode for DXIL 1.1 and Shader Model 6.1 (View Instancing support only).
```

cwfitzgerald: We can assume that all DX12 devices support DXIL - DXBC is already being converted to DXIL by the d3d12 runtime behind the scenes.

wgpu-hal/src/dx12/device.rs
```diff
@@ -221,62 +222,148 @@ impl super::Device {
...
+                "{}_{}\0",
+                naga_stage.to_hlsl_str(),
+                hlsl::ShaderModel::V5_1.to_str()
+            );
+            let mut error = native::Blob::null();
+
+            let hr = unsafe {
+                profiling::scope!("d3dcompiler::D3DCompile");
+                d3dcompiler::D3DCompile(
+                    source.as_ptr() as *const _,
```
Elabajaba, line 265: Preferable to use .cast() here instead of as casts for all of these pointer casts.

Elabajaba (changes requested):
Clippy errors also need to be fixed, and imo this needs visible docs for users about why they should use dxc, and how they can use it.

wgpu-hal/src/dx12/device.rs
39ali: @cwfitzgerald `winapi` only exposes `SM6.0` and `SM5.1` , once wgpu starts using `windows-rs` we can start using the highest supported SM , supporting higher SM is a must so we can support other features like float16,Subgroup etc... ,
Using DXC will still be a big win even if we only support SM6.0 for now .

wgpu-hal/src/dx12/device.rs
```diff
@@ -1482,14 +1579,19 @@ impl crate::Device<super::Api> for super::Device {
...
+                },
                 0,
                 native::CachedPSO::null(),
                 native::PipelineStateFlags::empty(),
             )
         };
 
-        blob_cs.destroy();
+        if let super::ShaderDXIL::FXC(cs_fxc) = blob_cs {
+            cs_fxc.destroy();
```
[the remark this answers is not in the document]
39ali: not sure im getting it , Blob will only be created if FXC was used , when we use DXC no blob will be allocated .

wgpu-hal/src/dx12/device.rs
```diff
@@ -221,62 +222,148 @@ impl super::Device {
...
+                }
+            }
         };
 
-        let hr = unsafe {
-            profiling::scope!("d3dcompiler::D3DCompile");
-            d3dcompiler::D3DCompile(
-                source.as_ptr() as *const _,
-                source.len(),
+        // hassle_rs doesn't check for `dxil.dll` if you only call `hassle_rs::compile_hlsl` but calling hassle_rs::Dxil::new will .
```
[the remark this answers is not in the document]
39ali: Good point , as of now `dx12` mod is only used on windows and looking at CI i don't think we validate shaders, let alone DXCIL .

wgpu-hal/src/dx12/device.rs
cwfitzgerald: The SM numbers are just enums, we can locally define the higher SMs

wgpu-hal/src/dx12/mod.rs
```diff
@@ -527,6 +527,11 @@ pub struct ShaderModule {
     raw_name: Option<ffi::CString>,
 }
 
+enum ShaderDXIL {
```
cwfitzgerald, line 530: This has a misleading name, this is a compiled shader, not DXIL

wgpu-hal/Cargo.toml
```diff
@@ -27,7 +27,7 @@ metal = ["naga/msl-out", "block", "foreign-types"]
 vulkan = ["naga/spv-out", "ash", "gpu-alloc", "gpu-descriptor", "libloading", "smallvec"]
 gles = ["naga/glsl-out", "glow", "egl", "libloading"]
 dx11 = ["naga/hlsl-out", "native", "libloading", "winapi/d3d11", "winapi/d3d11_1", "winapi/d3d11_2", "winapi/d3d11sdklayers", "winapi/dxgi1_6"]
-dx12 = ["naga/hlsl-out", "native", "bit-set", "range-alloc", "winapi/d3d12", "winapi/d3d12shader", "winapi/d3d12sdklayers", "winapi/dxgi1_6"]
+dx12 = ["naga/hlsl-out", "native", "bit-set", "range-alloc", "winapi/d3d12", "winapi/d3d12shader", "winapi/d3d12sdklayers", "winapi/dxgi1_6", "hassle-rs"]
```
cwfitzgerald, line 30: We probably shouldn't be depending on this unconditionally. It depends on the needs of firefox.

@jimblandy tl;dr: hassle-rs are bindings to DXC, they're not terribly thick bindings, but they exist. If FF is going to ship on FXC, do you need a way to disable support DXC entirely?

wgpu-hal/src/dx12/device.rs
```diff
@@ -221,62 +222,151 @@ impl super::Device {
...
-                            error.GetBufferPointer() as *const u8,
-                            error.GetBufferSize(),
-                        )
-                    };
-                    let _ = write!(full_msg, ": {}", String::from_utf8_lossy(message));
-                    unsafe {
-                        error.destroy();
-                    }
+                hassle_rs::HassleError::LoadLibraryError { filename, inner: _ } => {
+                    log::warn!("{:?} was not found ,falling back to FXC ,in order to use the new hlsl compiler `DXC`,`dxcompiler.dll` and `dxil.dll` should be included in the current execution environment" ,filename );
```
cwfitzgerald, line 344: We should only send this warning once per program execution.

wgpu-hal/src/dx12/device.rs
```diff
@@ -221,62 +222,151 @@ impl super::Device {
...
-        let mut error = native::Blob::null();
-        let mut compile_flags = d3dcompiler::D3DCOMPILE_ENABLE_STRICTNESS;
+        let source_name = stage
+            .module
+            .raw_name
+            .as_ref()
+            .and_then(|cstr| cstr.to_str().ok())
+            .unwrap_or_default();
+
+        let mut compile_flags = vec!["-Ges"]/* d3dcompiler::D3DCOMPILE_ENABLE_STRICTNESS */;
```
cwfitzgerald, line 232: Lets use an ArrayVec<_, 3> here

wgpu-hal/src/dx12/device.rs
```diff
@@ -221,62 +222,151 @@ impl super::Device {
...
+            compile_flags.push("-Od"); /* d3dcompiler::D3DCOMPILE_DEBUG */
         }
 
-        let source_name = match stage.module.raw_name {
-            Some(ref cstr) => cstr.as_c_str().as_ptr(),
-            None => ptr::null(),
+        // This closure gets called if `hassle_rs::compile_hlsl() fails due to a `hassle_rs::HassleError::LoadLibraryError`,
+        //
+        //  This usually means that `dxcompiler.dll` and/or `dxil.dll` could not be found in the current execution environment.
+        let load_shader_fxc = || {
```
cwfitzgerald, line 245: Having this much code in a closure is a bit of a code smell. It should likely be its own function.

wgpu-hal/src/dx12/device.rs
```diff
@@ -221,62 +222,151 @@ impl super::Device {
...
 
-        let hr = unsafe {
-            profiling::scope!("d3dcompiler::D3DCompile");
-            d3dcompiler::D3DCompile(
-                source.as_ptr() as *const _,
-                source.len(),
+        // hassle_rs doesn't check for `dxil.dll` if you only call `hassle_rs::compile_hlsl` but calling hassle_rs::Dxil::new will .
+        profiling::scope!("hassle_rs::compile_hlsl");
+        let (result, log_level) = match hassle_rs::Dxil::new(None).and_then(|_| {
+            hassle_rs::compile_hlsl(
```
cwfitzgerald, line 310: We shouldn't be using `compile_hlsl` directly as it needs to load the library every single time you call the function. Ideally you'll load a DXCCompiler and DXCLibrary once on dx12 instance creation, then re-use it for every time we need to compile stuff.

cwfitzgerald (changes requested):

wgpu-hal/src/dx12/device.rs
MarijnS95: And what's up with the commas?

cwfitzgerald:
Superceded by #3356

Issue #3112, opened 2022-10-15
`TextureFormat::Depth24PlusStencil8` no longer supported in 0.14.0


relrelb:
**Description**
Before 0.14.0, `wgpu::TextureFormat::Depth24PlusStencil8` was supported without a need to request any `wgpu::Features`. Since 0.14.0, it requires `wgpu::Features::DEPTH24PLUS_STENCIL8`, which is not supported on all platforms.
Together with the fact that `wgpu::Depth32FloatStencil8` requires a feature as well (albeit it has slightly wider support), this means stencil textures cannot be used reliably anymore.

**Repro steps**
Not an ideal repro, but while upgrading to `wgpu` 0.14.0 in [Ruffle](https://github.com/ruffle-rs/ruffle), the following error happened:
```
[2022-10-15T08:01:27Z ERROR wgpu::backend::direct] Handling wgpu errors as fatal by default
thread 'main' panicked at 'wgpu error: Validation Error

Caused by:
    In Device::create_texture
    Texture format Depth24PlusStencil8 can't be used due to missing features.
    Features DEPTH24PLUS_STENCIL8 are required but not enabled on the device
```
As a solution, `wgpu::Features::DEPTH24PLUS_STENCIL8` was added to the list of required features: https://github.com/ruffle-rs/ruffle/pull/8226/files#diff-d2d6e71ebcd52c28fa4e4c89c8249dfe5b9deb974415e86181856aef3331eb76R594.
However, it turned out that `wgpu::Features::DEPTH24PLUS_STENCIL8` is unsupported on some platforms, where the following error happens:
```
[2022-10-15T07:40:03Z ERROR wgpu::backend::direct] Error in Adapter::request_device: unsupported features were requested: DEPTH24PLUS_STENCIL8
Error: Couldn't create wgpu rendering backend
```
See https://github.com/ruffle-rs/ruffle/issues/8295 for more context.

Note that before upgrading to 0.14.0, `wgpu::TextureFormat::Depth24PlusStencil8` did work on those now-unsupported machines, arising the question whether the requirement of `wgpu::Features::DEPTH24PLUS_STENCIL8` is precise.

**Expected vs observed behavior**
I'd expect one of the following scenarios:
1. `wgpu::TextureFormat::Depth24PlusStencil8` will work without `wgpu::Features::DEPTH24PLUS_STENCIL8`, as it used to be before 0.14.0.
[...]

PR #3151, opened 2022-10-30
Remove `DEPTH24PLUS_STENCIL8` feature

Healthire:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
[ #3112](https://github.com/gfx-rs/wgpu/issues/3112)

**Description**
Removes the `DEPTH24PLUS_STENCIL8` feature. The feature is not present in the WebGPU spec and was blocking the use of the `Depth24PlusStencil8` texture format on several platforms that support it. Tests have been adjusted to include `Depth24PlusStencil8` in  `clear_texture_2d_uncompressed` instead of having its own test.

**Testing**
The change passes tests, and has been briefly tested locally on a Windows machine with an RTX 2080, using the Vulkan and WebGL backends.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ Bottom level categories:
 - Bother to free the `hal::Api::CommandBuffer` when a `wgpu_core::command::CommandEncoder` is dropped. By @jimblandy in [#3069](https://github.com/gfx-rs/wgpu/pull/3069).
 - Fixed the mipmap example by adding the missing WRITE_TIMESTAMP_INSIDE_PASSES feature. By @Olaroll in [#3081](https://github.com/gfx-rs/wgpu/pull/3081).
 - Avoid panicking in some interactions with invalid resources by @nical in (#3094)[https://github.com/gfx-rs/wgpu/pull/3094]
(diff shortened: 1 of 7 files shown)


maxammann:
I just tested this on my WebGL build of maplibre-rs and it still errors with:

```
ERROR /home/max/.cargo/git/checkouts/wgpu-746059844e0eb641/fe5b48c/wgpu/src/backend/direct.rs:908 
Error in Adapter::request_device: unsupported features were requested: DEPTH32FLOAT_STENCIL8
```

Tested on Firefox and Chrome with WebGL.

maxammann:
> The feature is not present in the WebGPU spec and was blocking the use of the `Depth24PlusStencil8` texture format on several platforms that support it.

I think you meant here Depth32FloatStencil8 right?

Healthire:
> I think you meant here Depth32FloatStencil8 right?

No, this change addresses the `DEPTH24PLUS_STENCIL8` feature, which wasn't marked as supported on some backends, making the `Depth24PlusStencil8` texture format unavailable even though it was actually supported.

`Depth32FloatStencil8` is gated by the `DEPTH32FLOAT_STENCIL8` feature, which is in the [WebGPU spec](https://gpuweb.github.io/gpuweb/#depth32float-stencil8) and so should probably stay? I don't know what support looks like for `Depth32FloatStencil8` on the different backends but it's unrelated to this change.

cwfitzgerald (approved):
Thanks! Will backport shortly

PR #3157, opened 2022-11-01
Added new  UNRESTRICTED_INDEX_BUFFER downlevel flag.

Wumpf:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
n/a

**Description**

WebGL doesn't support buffers that are are used both as vertex and index buffer. This introduces a new downlevel feature to check for this.

**Testing**
Tested same application in browser and desktop (Metal). Browser now correctly logs an error:

```
panicked at 'wgpu error: Validation Error

Caused by:
    In Device::create_buffer
      note: label = `rerun logo`
    Downlevel flags BUFFER_USAGE_COMBINE_VERTEX_INDEX are required but not supported on the device.
This is not an invalid use of WebGPU: the underlying API or device does not support enough features to be a fully compliant implementation. A subset of the features can still be used. If you are running this program on native and not in a browser and wish to work around this issue, call Adapter::downlevel_properties or Device::downlevel_properties to get a listing of the features the current platform supports.
```
not super expressive but good enough I reckon

### DIFF
--- a/wgpu-core/src/command/transfer.rs
+++ b/wgpu-core/src/command/transfer.rs
@@ -541,9 +541,12 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         let hub = A::hub(self);
         let mut token = Token::root();
 
+        let (device_guard, mut token) = hub.devices.read(&mut token);
         let (mut cmd_buf_guard, mut token) = hub.command_buffers.write(&mut token);
         let cmd_buf = CommandBuffer::get_encoder_mut(&mut *cmd_buf_guard, command_encoder_id)?;
         let (buffer_guard, _) = hub.buffers.read(&mut token);
+
+        let device = &device_guard[cmd_buf.device_id.value];
 
         #[cfg(feature = "trace")]
         if let Some(ref mut list) = cmd_buf.commands {
@@ -593,6 +596,26 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         }
         if destination_offset % wgt::COPY_BUFFER_ALIGNMENT != 0 {
             return Err(TransferError::UnalignedBufferOffset(destination_offset).into());
+        }
+        if !device
+            .downlevel
+            .flags
+            .contains(wgt::DownlevelFlags::UNRESTRICTED_INDEX_BUFFER)
+            && (src_buffer.usage.contains(BufferUsages::INDEX)
+                || dst_buffer.usage.contains(BufferUsages::INDEX))
+        {
+            let non_index_non_copy_buffer_usages = wgt::BufferUsages::VERTEX
+                | wgt::BufferUsages::UNIFORM
+                | wgt::BufferUsages::INDIRECT
+                | wgt::BufferUsages::STORAGE;
+            if src_buffer.usage.contains(non_index_non_copy_buffer_usages)
+                || dst_buffer.usage.contains(non_index_non_copy_buffer_usages)
+            {
+                return Err(TransferError::MissingDownlevelFlags(MissingDownlevelFlags(
+                    wgt::DownlevelFlags::UNRESTRICTED_INDEX_BUFFER,
+                ))
+                .into());
+            }
         }
 
         let source_end_offset = source_offset + size;
--- a/wgpu-core/src/command/transfer.rs
+++ b/wgpu-core/src/command/transfer.rs
@@ -601,15 +601,15 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             .downlevel
             .flags
             .contains(wgt::DownlevelFlags::UNRESTRICTED_INDEX_BUFFER)
-            && (src_buffer.usage.contains(BufferUsages::INDEX)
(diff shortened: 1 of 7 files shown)


cwfitzgerald:
Maybe `MIXED_USAGE_INDEX_BUFFER` or something.

Wumpf:
Thanks for digging that up! Went ahead and looked into doing transfer restrictions as well which is why I called it `UNRESTRICTED_INDEX_BUFFER`

Only open question would be if we should restrict UNIFORM interaction with other buffer as well, but I haven't looked into how that emulation is done exactly

wgpu-types/src/lib.rs
```diff
@@ -1087,6 +1087,13 @@ bitflags::bitflags! {
         ///
         /// WebGL doesn't support this.
         const BUFFER_BINDINGS_NOT_16_BYTE_ALIGNED = 1 << 15;
+
+        /// Supports buffers to combine [`BufferUsages::INDEX`] with usages other than [`BufferUsages::COPY_DST`] and [`BufferUsages::COPY_SRC`].
+        /// Furthermore, in absence of this feature it is not allowed to copy index buffers from/to buffers with usages other than
+        /// [`BufferUsages::INDEX`]/[`BufferUsages::COPY_DST`]/[`BufferUsages::COPY_SRC`].
```
cwfitzgerald, line 1093: This needs to include MAP_* as it seems like you can't copy to and from mapped buffers.

wgpu-core/src/command/transfer.rs
```diff
@@ -594,6 +597,26 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+            .flags
+            .contains(wgt::DownlevelFlags::UNRESTRICTED_INDEX_BUFFER)
+            && (src_buffer.usage.contains(BufferUsages::INDEX)
+                || dst_buffer.usage.contains(BufferUsages::INDEX))
+        {
+            let non_index_non_copy_buffer_usages = wgt::BufferUsages::VERTEX
+                | wgt::BufferUsages::UNIFORM
+                | wgt::BufferUsages::INDIRECT
+                | wgt::BufferUsages::STORAGE;
+            if src_buffer.usage.contains(non_index_non_copy_buffer_usages)
```
cwfitzgerald, line 611: Contains is the wrong operation here, we want intersects (as we want any, not all)

wgpu-core/src/command/transfer.rs
```diff
@@ -594,6 +597,26 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         if destination_offset % wgt::COPY_BUFFER_ALIGNMENT != 0 {
             return Err(TransferError::UnalignedBufferOffset(destination_offset).into());
         }
+        if !device
+            .downlevel
+            .flags
+            .contains(wgt::DownlevelFlags::UNRESTRICTED_INDEX_BUFFER)
+            && (src_buffer.usage.contains(BufferUsages::INDEX)
```
cwfitzgerald, line 604: Mixed imported and fully qualified usage of `BufferUsages` - prefer `wgt::BufferUsages`

cwfitzgerald (changes requested):

wgpu-types/src/lib.rs
Wumpf: the doc here is correct then - it says the only things *allowed* to copy from/to are INDEX/COPY_DST/COPY_SRC. But the code in transfer.rs isn't correct and doesn't mind the MAP_WRITE/READ currently.

Wumpf: oh, or are you saying there is the additional restriction that index buffers can't be mapped?

cwfitzgerald: I phrased this poorly - it's _valid_ to copy to and from mapped buffers (and this is a very common operation) but the docs make it seem like it's invalid.

wgpu-types/src/lib.rs
```diff
@@ -1063,7 +1063,7 @@ bitflags::bitflags! {
 
         /// Supports samplers with anisotropic filtering. Note this isn't actually required by
         /// WebGPU, the implementation is allowed to completely ignore aniso clamp. This flag is
-        /// here for native backends so they can comunicate to the user of aniso is enabled.
+        /// here for native backends so they can communi    cate to the user of aniso is enabled.
```
cwfitzgerald, line 1066:
```suggestion
        /// here for native backends so they can communicate to the user of aniso is enabled.
```

cwfitzgerald (approved):

wgpu-types/src/lib.rs
Wumpf: I keep on messing up this pr Oo

cwfitzgerald: My cat has tried to... contribute to PRs before :)

Wumpf: ... wow. okay. Yeah is late here and I shouldn't be let on a computer either. Read the diff the other way round xD

PR #3163, opened 2022-11-02
Suballocate DX12 buffer creation

### LINKED ISSUE #2720: Suballocate Buffers in DX12
We currently don't suballocate buffers or textures in DX12, we directly call `CreateCommittedResource` which creates a memory heap for each individual object. This is very slow, we should be suballocating larger heaps and using `CreatePlacedResource`. We would use gpu-alloc for this.

Elabajaba:
**Checklist**

- ~~[ ] Blocked on https://github.com/gfx-rs/wgpu/issues/3207~~ Worked around by feature gating it behind the `windows_rs` feature
- [x] Run `cargo clippy`.
- [x] Fix the buffer and texture usage flags.
- [ ] Don't panic if deallocation fails when destroying a texture or resource.
  - Is panic() on deallocation failure even a real issue?
  - What should happen on deallocation failure, considering it can't (currently) return an error?
- [ ] Figure out what to do with heap_created_not_zeroed.
- [x] Fix the error handling so the gpu-allocator -> wgpu error implementation is not just a bunch of todo!().
- [ ] Consider waiting until https://github.com/Traverse-Research/gpu-allocator/pull/138 lands, and see if migrating to [presser](https://github.com/EmbarkStudios/presser) would be needed
  - More safety!
  - Might break a lot of user code...
  - That user code is probably UB anyways?
- [ ] Investigate if changing gpu-alloc to support dx12 is feasible, and would be accepted
  - [ ] https://github.com/zakarumych/gpu-alloc/issues/66
  - imo probably not, as gpu-alloc seems tightly coupled to the vulkan way, and dx12's memory management is different enough from vulkan's to make it painful to try and fit into the vulkan way

**Connections**
~~ Blocked on #3207 ~~ Worked around by feature gating it behind the `windows_rs` feature
closes  #2720

**Description**
DX12 is currently quite slow in wgpu. This uses gpu-allocator to batch together allocations into heaps and uses CreatePlacedResource instead of CreateCommittedResource to create buffers and textures, which leads to large performance gains (~30-50% in "normal" scenarios, with significantly larger gains in write_buffer heavy scenarios (~250x+ in an unrealistic scenario where it calls write_buffer 1000x in a loop, going from ~1fps to ~250fps)), and in my testing no performance decreases.

**Testing**
Tested the examples, ran cargo test, backported it to 0.14 and tested against bevy+bistro, and tested against a modified water example where it loops the render write_buffer 1000x times on the main thread, 500x each on 2 scoped threads, or 100x each on 10 scoped threads to make sure multithreading wouldn't panic.

It was quite a bit faster in all of these scenarios, except for bevy+bistro at 4k where it was heavily gpu limited and ran about the same.

**Potential Future Improvements**
- [ ] Make these into issues
- [ ] Consider if adding a way to get the vulkan/dx12/etc allocator is worthwhile
- [ ] Consider actually suballocating resources instead of just suballocating heaps. DX12 allows for multiple bindings to the same resource through subresources.
  - This would probably require changing the unmap_buffer trait to pass in the subresource id for unmapping a buffer
- [ ] Figure out why dx12 is still slower than Vulkan when calling write_buffer a lot
  - Basic profiling seems to point to windows syscalls taking the majority of the time, with ntdll`ZwAllocateLocallyUniqueId` taking almost 40% of the time in the 1000x looped write_buffer test
- [ ] Most dx12 things are already thread safe, see if it's possible to avoid wrapping them in rust sync primitives to avoid multiple levels of locks and/or reference counting


Jasper-Bekkers:
Nice to see these changes @Elabajaba, if there are any features that wgpu would like to see in our allocator please just file an issue on the repo.

Elabajaba:
Blocked on #3207 until Mozilla gets around to vendoring windows-rs.

MarijnS95:
Just stumbling upon this PR, I'll see to accelerating https://github.com/Traverse-Research/gpu-allocator/pull/138 so that you're unblocked on that regard!

cwfitzgerald:
As #3207 is basically perma-blocked until further notice, I think we should work around the situation by having a feature flag and falling back to the old behavior when it's disabled. This will let us continue to innovate, and also not force the issue with moz.

wgpu-hal/src/dx12/device.rs
```diff
@@ -328,6 +356,7 @@ impl crate::Device<super::Api> for super::Device {
         unsafe { queue.raw.destroy() };
     }
 
+    #[cfg(feature = "windows_rs")]
```
cwfitzgerald, line 359: This all looks good, but could we put all the allocator stuff in it's own module such that this code doesn't have to be cfg guarded. For example the non-cfg can have a memory allocator type that is just (). That kind of thing.

wgpu-hal/src/dx12/device.rs
```diff
@@ -3,19 +3,25 @@ use crate::{
...
+        create_allocator_wrapper, free_buffer_allocation, free_texture_allocation,
+    },
+};
 use parking_lot::Mutex;
 use std::{ffi, mem, num::NonZeroU32, ptr, slice, sync::Arc};
 use winapi::{
     shared::{dxgiformat, dxgitype, minwindef::BOOL, winerror},
-    um::{d3d12, d3dcompiler, synchapi, winbase},
+    um::{
+        d3d12::{self, D3D12_RESOURCE_DESC},
```
cwfitzgerald, line 17: Let's always refer to D3D12_RESOURCE_DESC as qualified from d3d12

wgpu-hal/src/dx12/mod.rs
```diff
@@ -40,6 +40,7 @@ mod descriptor;
 mod device;
 mod instance;
 mod view;
+mod windows_rs_suballocation;
```
cwfitzgerald, line 43: I would name it something like `suballocation`

wgpu-hal/src/dx12/device.rs
```diff
@@ -355,84 +369,64 @@ impl crate::Device<super::Api> for super::Device {
...
-                    d3d12::D3D12_HEAP_FLAG_NONE
-                },
-                &raw_desc,
-                d3d12::D3D12_RESOURCE_STATE_COMMON,
-                ptr::null(),
-                &d3d12::ID3D12Resource::uuidof(),
-                resource.mut_void(),
-            )
-        };
+        let (hr, allocation) = my_buffer_hr(self, desc, raw_desc, &mut resource)?;
```
cwfitzgerald, line 372: hr?

wgpu-hal/src/dx12/device.rs
```diff
@@ -312,34 +321,39 @@ impl super::Device {
...
-    unsafe fn exit(self, queue: super::Queue) {
+    unsafe fn exit(#[allow(unused_mut)] mut self, queue: super::Queue) {
         self.rtv_pool.lock().free_handle(self.null_rtv_handle);
         unsafe { self.rtv_pool.into_inner().destroy() };
         unsafe { self.dsv_pool.into_inner().destroy() };
         unsafe { self.srv_uav_pool.into_inner().destroy() };
         unsafe { self.sampler_pool.into_inner().destroy() };
         unsafe { self.shared.destroy() };
         unsafe { self.idler.destroy() };
+        #[cfg(feature = "windows_rs")]
```
cwfitzgerald, line 338: This shouldn't need to be conditional?

wgpu-hal/src/dx12/device.rs
```diff
@@ -312,34 +321,39 @@ impl super::Device {
...
+        drop(self.mem_allocator.take());
         unsafe { queue.raw.destroy() };
     }
 
     unsafe fn create_buffer(
         &self,
         desc: &crate::BufferDescriptor,
     ) -> Result<super::Buffer, crate::DeviceError> {
+        use super::windows_rs_suballocation::my_buffer_hr;
```
cwfitzgerald, line 347: let's just fully qualify

wgpu-hal/src/dx12/windows_rs_suballocation.rs
```diff
@@ -0,0 +1,303 @@
...
+        Ok(_) => (),
+        // TODO: Don't panic here
+        Err(e) => panic!("failed to destroy dx12 texture, {}", e),
+    };
+}
+
+#[cfg(feature = "windows_rs")]
+impl From<gpu_allocator::AllocationError> for crate::DeviceError {
+    fn from(result: gpu_allocator::AllocationError) -> Self {
+        use log::error;
```
cwfitzgerald, line 274: Let's just qualify

wgpu-hal/src/dx12/device.rs
Elabajaba: Clippy complains if it isn't

wgpu-hal/src/dx12/device.rs
Elabajaba: HRESULT, I just copied what the name was before. Looks like I forgot to rename the buffer and texture resource creation functions.

wgpu-hal/src/dx12/device.rs
cwfitzgerald: fwiw, you could just write `self.mem_allocator = None`

Elabajaba: I just copied their example there, which says `drop(allocator); // Explicitly drop before destruction of device.`.

Does setting it to None ensure that it'll be dropped before the device is destroyed, or does it only drop at the end of the block where it was set to `None`?

cwfitzgerald: It drops immediately.

MarijnS95: I personally prefer `.take()` over `= None`, for these kind of things, fwiw. It reads a bit more natural IMO.

wgpu-hal/src/dx12/suballocation.rs
```diff
@@ -0,0 +1,307 @@
...
+pub(crate) const D3D12_HEAP_FLAG_CREATE_NOT_ZEROED: u32 = d3d12::D3D12_HEAP_FLAG_NONE;
+
+#[derive(Debug)]
+pub(crate) struct GpuAllocatorWrapper {
+    #[cfg(feature = "windows_rs")]
+    pub(crate) allocator: gpu_allocator::d3d12::Allocator,
+    #[cfg(not(feature = "windows_rs"))]
+    #[allow(unused)]
+    pub(crate) allocator: (),
+}
```
cwfitzgerald, line 34: Lets split this into two large blocks, one for with-windows-rs and one for without, either large cfg blocks or two inline modules, then just expose the same interface from both modules. As it stands you have a bunch of unrelated code mashed together which makes it very hard to follow for someone who doesn't know what's going on.

wgpu-hal/src/dx12/suballocation.rs
```diff
@@ -0,0 +1,307 @@
...
+        (true, true) => MemoryLocation::CpuToGpu,
+        (true, false) => MemoryLocation::GpuToCpu,
+        (false, true) => MemoryLocation::CpuToGpu,
+        (false, false) => MemoryLocation::GpuOnly,
+    };
+
+    let name = desc.label.unwrap_or("Unlabeled buffer");
+
+    // SAFETY: allocator exists when the windows_rs feature is enabled
+    let mut allocator = unsafe { device.mem_allocator.as_ref().unwrap_unchecked().lock() };
```
cwfitzgerald, line 92: Now that #3051 is in, let's make an extension trait in assertions.rs like:

```rust
trait StrictAssertUnwrapExt<T> {
    fn strict_unwrap_unchecked(self) -> T;
}

impl<T> StrictAssertUnwrapExt<T> for Option<T> {
    fn strict_unwrap_unchecked(self) -> T {
        strict_assert!(self.is_some(), "Called strict_unwrap_unchecked on None");
        unsafe { self.unwrap_unchecked() }
    }
}

impl<T, E> StrictAssertUnwrapExt<T> for Result<T, E> {
    fn strict_unwrap_unchecked(self) -> T {
        strict_assert!(self.is_ok(), "Called strict_unwrap_unchecked on Err");
        unsafe { self.unwrap_unchecked() }
    }
}
```

wgpu-hal/src/dx12/suballocation.rs
```diff
@@ -0,0 +1,307 @@
...
+// Currently this will work the older, slower way if the windows_rs feature is disabled,
+// and will suballocate buffers using gpu_allocator if the windows_rs feature is enabled.
+
+#[cfg(feature = "windows_rs")]
+use gpu_allocator::{
+    d3d12::{AllocationCreateDesc, ToWinapi, ToWindows},
+    MemoryLocation,
+};
+
+// TODO: find the exact value
```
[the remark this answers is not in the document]
Elabajaba: Copied from the old implementation. I wasn't sure what exactly kvark meant by that so I just left it.

wgpu-hal/src/dx12/suballocation.rs
Elabajaba: I wasn't sure if you meant the assertions.rs in `wgpu-types`, or the one in `wgpu-core`, but to avoid a circular dependency I implemented this in the wgpu-types one and changed its `mod assert` to public.

wgpu-hal/src/dx12/suballocation.rs
Elabajaba: I went with #[cfg] blocked inline modules, seemed cleaner to me.

cwfitzgerald: I think having them inline is fine for when there's a decent overlap in the code paths, but when they're basically completely different, it helps to separate them

wgpu-hal/src/dx12/suballocation.rs
cwfitzgerald: Ack, the one in core should have gotten removed. Types was the right decision - I'll sort the assertions out.

wgpu-hal/src/dx12/suballocation.rs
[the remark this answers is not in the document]
cwfitzgerald: This needs to be 0x1000  https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_heap_flags

wgpu-hal/Cargo.toml
```diff
@@ -33,6 +33,8 @@ vulkan = ["naga/spv-out", "ash", "gpu-alloc", "gpu-descriptor", "libloading", "s
 gles = ["naga/glsl-out", "glow", "egl", "libloading"]
 dx11 = ["naga/hlsl-out", "native", "libloading", "winapi/d3d11", "winapi/d3d11_1", "winapi/d3d11_2", "winapi/d3d11sdklayers", "winapi/dxgi1_6"]
 dx12 = ["naga/hlsl-out", "native", "bit-set", "range-alloc", "winapi/d3d12", "winapi/d3d12shader", "winapi/d3d12sdklayers", "winapi/dxgi1_6"]
+# TODO: This is a separate feature until Mozilla okays windows-rs, see https://github.com/gfx-rs/wgpu/issues/3207 for the tracking issue.
+windows_rs = ["gpu-allocator"]
```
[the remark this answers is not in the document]
Elabajaba: Done.

wgpu-hal/src/dx12/suballocation.rs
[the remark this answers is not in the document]
Elabajaba: According to https://devblogs.microsoft.com/directx/coming-to-directx-12-more-control-over-memory-allocation/ this isn't always available with DX12.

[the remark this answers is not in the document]
cwfitzgerald: I had suspected this but was surprised it wasn't mentioned in the docs. Fine to leave

cwfitzgerald (approved):
Thank you so much for all this work! Looks great!

wgpu-hal/src/dx12/suballocation.rs
```diff
@@ -0,0 +1,307 @@
...
+// TODO: figure out if this is even needed?
```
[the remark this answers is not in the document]
MarijnS95: We don't zero memory as that's really unneeded overhead in the vast majority of cases where an allocation will be written to first anyway.

PR #3165, opened 2022-11-02
Backport #3151 Without Breaking Changes

### LINKED ISSUE #3112: `TextureFormat::Depth24PlusStencil8` no longer supported in 0.14.0

cwfitzgerald:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Fixes #3112 

**Description**

Makes `wgpu::TextureFormat::Depth24PlusStencil8` available on all backends by making the feature unconditionally available and the feature unneeded to use the format.

**Testing**

Local tests were run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,12 @@ Bottom level categories:
 
 ## Unreleased
 
+## wgpu-0.14.1 (2022-11-02)
+
+### Bug Fixes
+
+- Make `wgpu::TextureFormat::Depth24PlusStencil8` available on all backends by making the feature unconditionally available and the feature unneeded to use the format. By @Healthire and @cwfitzgerald in [#3151](https://github.com/gfx-rs/wgpu/pull/3151)
+
 ## wgpu-0.14.0 (2022-10-05)
 
 ### Major Changes
--- a/wgpu-hal/src/dx11/adapter.rs
+++ b/wgpu-hal/src/dx11/adapter.rs
@@ -86,6 +86,7 @@ impl super::Adapter {
         // TODO(cwfitzgerald): Needed downlevel features: 3D dispatch
 
         let mut features = wgt::Features::DEPTH_CLIP_CONTROL
+            | wgt::Features::DEPTH24PLUS_STENCIL8 // workaround #3112
             | wgt::Features::PUSH_CONSTANTS
             | wgt::Features::POLYGON_MODE_LINE
             | wgt::Features::CLEAR_TEXTURE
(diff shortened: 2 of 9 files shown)


cwfitzgerald:
The test failed, but the test failed via segfault on lavapipe, so I'm going to ignore that failure and push on.

Issue #3159, opened 2022-11-01
Test Notable Special Shader Functions


cwfitzgerald:
For wgsl functions that have special meanings (clamp), weird edge cases, or particularly difficult implementations (like polyfills), we should have shader tests based on the work in #3125 which makes sure they behave within contract.

Filing for transparency, this is what I'm working on next.

PR #3167, opened 2022-11-02
Test `clamp` and Generalize Shader Tests

cwfitzgerald:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Works on #3159 

**Description**

First pass which generalizes the existing shader test suite to be able to handle a wider range of cases.

**Testing**

Tested on NV 1080ti, AMD 580, Intel 630, and Apple M1.


cwfitzgerald:
These are tests and shouldn't be controversial.

PR #3173, opened 2022-11-03
Expand documentation of `Queue::write_texture`.

kpreid:
* Explain which of the source and destination each parameter applies to.
* Clarify that just because this “schedules” a write doesn't mean that the slice has to be kept alive (this is “obvious” from the lifetimes but I have heard of it being confusing).

Inspired by <https://github.com/gfx-rs/wgpu/discussions/3172>

----

P.S. I have intentionally not bothered to update the changelog because this is a trivial documentation update and I would really like to make contributing this as simple as possible. If it is project policy that all PRs must also include changelog entries, … well, please reconsider that, but if so, I will.

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -3596,13 +3596,23 @@ impl Queue {
         }
     }
 
-    /// Schedule a data write into `texture`.
+    /// Schedule a write of some data into a texture.
+    ///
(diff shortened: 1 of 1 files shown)


cwfitzgerald:
wrt the changelog - not a project requirement at all, just there to make sure people give themselves credit for the work they do

Issue #2285, opened 2021-12-12
ID3D12CommandAllocator Error for heavy computation pipeline


haixuanTao:
**Description**
I have written a library that transform Deep Learning models into WGPU compute pipeline, called wonnx: https://github.com/haixuanTao/wonnx

The library works on a mnist model on Windows DX12 but larger model like squeezenet fail to run on DX12. 
Both mnist and squeezenet works on Linux VULKAN on github action and local with NVIDIA card.

The error I get is:
```
ID3D12CommandAllocator::Reset: A command allocator 0x000001C4C43FE4C0:'Unnamed ID3D12CommandAllocator Object' is being reset before previous executions associated with the allocator have completed. [ EXECUTION ERROR #552: COMMAND_ALLOCATOR_SYNC]
```

I have scoped the error to the line:
```
device.poll(wgpu::Maintain::Wait);
```

From my research, I think this error has to do with the high number of compute pipeline as squeezenet is 10x larger than mnist.

I have gotten this error on a vagrant VM and github action VM and it might be caused by the virtualisation.

**Repro steps**
To reproduce the error, you can clone my repo:

```
SETX RUST_LOG debug
git clone https://github.com/haixuanTao/wonnx
git checkout 71e25a47f5ed831fa96499b77084424188b2e35d
cargo run --example squeeze
```
You can also run the test that should be passing
```
cargo test
```

You can also check my github action here: https://github.com/haixuanTao/wonnx/actions/runs/1569686479 that has test check for both linux x86 and windows x86.

**Expected vs observed behavior**
I would expect Windows to either fail both MNIST and SQUEEZENET if it was an implementation problem.

**Extra materials**

```
time: pre_run: 24.2054ms                                                                                                
[2021-12-12T18:56:15Z INFO  wgpu_core::device] Created buffer Valid((53, 2, Dx12)) with BufferDescriptor { label: Some("staging_squeezenet0_flatten0_reshape0"), size: 4000, usage: MAP_READ | COPY_DST, mapped_at_creation: false }            
time: run: 159.6157ms                                                                                                   
time: run: 200.1022ms                                                                                                   
[...]

Issue #552, opened 2020-04-05
Check BindGroup compatibility in compute pass


kunalmohan:
We check the compatibility of BindGroup before issuing a draw call in render pass. Similar check needs to be done before any dispatch in compute pass.

cwfitzgerald:
Seems to have been solved in the mean time by https://github.com/gfx-rs/wgpu/blob/master/wgpu-core/src/command/compute.rs#L438

PR #3174, opened 2022-11-03
Zero-initialize workgroup memory

### LINKED ISSUE #2430: Zero out workgroup memory
For WebGPU it's important that the workgroup memory doesn't contain garbage. Otherwise one page could see the side products of another page. We need to find a way to zero-initialize it, and we need to figure out when exactly we have to do this.

teoxoy:
Depends on https://github.com/gfx-rs/naga/pull/2111

**Checklist**

- [x] Run `cargo clippy`.

**Connections**
fixes #2430

**Description**
Zero-initialize workgroup memory mostly via polyfills (injected by naga) with extra support for [`VK_KHR_zero_initialize_workgroup_memory`](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VK_KHR_zero_initialize_workgroup_memory.html) (when available).

**Testing**
Added a test and tested manually on all backends to make sure everything is working properly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ path = "./wgpu-hal"
 
 [workspace.dependencies.naga]
 git = "https://github.com/gfx-rs/naga"
-rev = "1be8024"
+rev = "c7d02151f08d6285683795289b5725b827d836d1"
 version = "0.10"
 
 [workspace.dependencies]
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -31,6 +31,8 @@ pub struct PhysicalDeviceFeatures {
         vk::PhysicalDeviceShaderFloat16Int8Features,
         vk::PhysicalDevice16BitStorageFeatures,
     )>,
+    zero_initialize_workgroup_memory:
+        Option<vk::PhysicalDeviceZeroInitializeWorkgroupMemoryFeatures>,
 }
 
 // This is safe because the structs have `p_next: *mut c_void`, which we null out/never read.
@@ -68,6 +70,9 @@ impl PhysicalDeviceFeatures {
         if let Some((ref mut f16_i8_feature, ref mut _16bit_feature)) = self.shader_float16 {
             info = info.push_next(f16_i8_feature);
             info = info.push_next(_16bit_feature);
+        }
+        if let Some(ref mut feature) = self.zero_initialize_workgroup_memory {
+            info = info.push_next(feature);
         }
         info
     }
@@ -283,6 +288,19 @@ impl PhysicalDeviceFeatures {
                         .uniform_and_storage_buffer16_bit_access(true)
                         .build(),
                 ))
+            } else {
+                None
+            },
+            zero_initialize_workgroup_memory: if effective_api_version >= vk::API_VERSION_1_3
+                || enabled_extensions.contains(&vk::KhrZeroInitializeWorkgroupMemoryFn::name())
+            {
+                Some(
+                    vk::PhysicalDeviceZeroInitializeWorkgroupMemoryFeatures::builder()
+                        .shader_zero_initialize_workgroup_memory(
+                            private_caps.zero_initialize_workgroup_memory,
+                        )
+                        .build(),
+                )
             } else {
                 None
             },
@@ -876,6 +894,16 @@ impl super::InstanceShared {
                 builder = builder.push_next(&mut next.1);
             }
 
+            // `VK_KHR_zero_initialize_workgroup_memory` is promoted to 1.3
+            if capabilities.effective_api_version >= vk::API_VERSION_1_3
+                || capabilities.supports_extension(vk::KhrZeroInitializeWorkgroupMemoryFn::name())
+            {
+                let next = features
+                    .zero_initialize_workgroup_memory
+                    .insert(vk::PhysicalDeviceZeroInitializeWorkgroupMemoryFeatures::default());
+                builder = builder.push_next(next);
+            }
+
             let mut features2 = builder.build();
             unsafe {
                 get_device_properties.get_physical_device_features2(phd, &mut features2);
@@ -1035,6 +1063,11 @@ impl super::Instance {
                     .image_robustness
                     .map_or(false, |ext| ext.robust_image_access != 0),
             },
+            zero_initialize_workgroup_memory: phd_features
+                .zero_initialize_workgroup_memory
+                .map_or(false, |ext| {
+                    ext.shader_zero_initialize_workgroup_memory == vk::TRUE
+                }),
         };
         let capabilities = crate::Capabilities {
             limits: phd_capabilities.to_wgpu_limits(),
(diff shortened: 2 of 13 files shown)


wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -295,6 +300,19 @@ impl PhysicalDeviceFeatures {
...
+            {
+                Some(
+                    vk::PhysicalDeviceZeroInitializeWorkgroupMemoryFeatures::builder()
+                        .shader_zero_initialize_workgroup_memory(
+                            private_caps.zero_initialize_workgroup_memory,
+                        )
+                        .build(),
+                )
+            } else {
+                None
```
ErichDonGubler, line 314: question: would a diagnostic via `tracing` be interesting to acknowledge that zero-initialization was requested but unsupported?

ErichDonGubler (commented):
I don't think I know enough about workgroups yet to be able to review this meaningfully. I've left what I could, though!

Cargo.toml
```diff
@@ -38,8 +38,8 @@ package = "wgpu-hal"
 path = "./wgpu-hal"
 
 [workspace.dependencies.naga]
-git = "https://github.com/gfx-rs/naga"
-rev = "e7fc8e6"
+git = "https://github.com/teoxoy/naga"
+rev = "6bc913d"
```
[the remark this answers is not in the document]
teoxoy: Oh yeah, sorry forgot to add a link to the naga PR here. In cases like these we usually wait for the naga PR to land then update the git dep to the correct rev.

wgpu-hal/src/vulkan/adapter.rs
teoxoy: Doesn't seem we do this for the other extensions. Not sure how useful it is.

teoxoy:
Converting to draft so that https://github.com/gfx-rs/naga/pull/2111 can land first.

wgpu/tests/shader/zero_init_workgroup_mem.rs
```diff
@@ -0,0 +1,150 @@
...
+    initialize_test(
+        TestParameters::default()
+            .downlevel_flags(DownlevelFlags::COMPUTE_SHADERS)
+            .limits(Limits::downlevel_defaults()),
+        zero_init_workgroup_mem_impl,
+    );
+}
+
+/// Increases iterations and writes random data to workgroup memory before reading it each iteration.
+const TRY_TO_FAIL: bool = false;
```
cwfitzgerald, line 21: I feel like this should just always be on

teoxoy: This is more for testing for failures specifically since I noticed that I was getting quite a few false positives when testing for failures. Turning it on still doesn't guarantee that it will fail since some drivers seem to zero initialize workgroup memory. Maybe the name doesn't really help convey this.

Thoughts?

cwfitzgerald: Fundamentally what we care about is that it gets zero initialized somehow either by us or the driver.

So if we try our best to make it fail, then it succeeds, we know we're getting what we need done. Without trying our best to fail we might just be getting extra lucky.

teoxoy: Hmm, I see what you mean but it will increase the running time of the test quite a bit. Maybe we should lower the iterations to 10 or 20? I just picked 100 randomly for my manual testing.

cwfitzgerald: Would dispatching 64x64 read workgroups (and having them all write their value to the buffer we check to be zero) make it such that we hit the non-zero-init the first time? Shouldn't increase execution time a lot but should make it more reproducable?

cwfitzgerald: Actually now that I look at it, there's no actual synchronization dependency between the two workgroups - that might also exacerbate the problem (if they both had the bg binding)

teoxoy: I used to have the workgroup size for the read pipeline be 64x64 but FXC was complaining about the race condition and I just set it to 1.

teoxoy: I think there is no way to 100% make sure this test passes because we are initializing the memory since the driver is free to do so (and some do; I encountered it on my dedicated gpu on the DX12 backend) and cause it to succeed.

cwfitzgerald: What I mean for reading is a workgroup size of 1, but doing a 64x64 dispatch, writing out the contents of every workgroup's memory separately. Then compare all of them to see if they're all zero as expected.

We can't make sure 100% that the test fails every time if our zero-init breaks, but we can raise that chance has high as possible, which is what spamming values into workgroup memory does. We want to try to write to and read from every workgroup's memory.

When our zero-init is working, it should always pass. It would false positive on a zero-init-ing driver/gpu, but should be as close to reliable as it can be on a non-zero-init-ing gpu because of how many workgroups we're spamming.

Hope that makes sense

teoxoy: Ah, I see what you mean. Increasing the "read" dispatch makes sense. Will update the test tomorrow.

Regarding the nr of iterations, should we still do like 10 or do you think having a read dispatch of 64x64x64 be enough?

cwfitzgerald: The fewest amount of iterations that consistently results in failures on drivers without zero init.

teoxoy:
Looks like the new improved test is causing some unrelated issues to surface.

- DX12 is failing with
  ```
  [2022-11-07T12:39:55Z ERROR wgpu_hal::auxil::dxgi::exception] ID3D12CommandAllocator::Reset: A command allocator 0x000002882019FFD0:'Unnamed ID3D12CommandAllocator Object' is being reset before previous executions associated with the allocator have completed. [ EXECUTION ERROR #552: COMMAND_ALLOCATOR_SYNC]
  [2022-11-07T12:39:55Z ERROR wgpu_hal::auxil::dxgi::exception] ID3D12Resource2::<final-release>: CORRUPTION: An ID3D12Resource object (0x000002882019FB60:'(wgpu internal) Staging') is referenced by GPU operations in-flight on Command Queue (0x000002882004E570:'Unnamed ID3D12CommandQueue Object').  It is not safe to final-release objects that may have GPU operations pending.  This can result in application instability. [ EXECUTION ERROR #921: OBJECT_DELETED_WHILE_STILL_IN_USE]
  ```

  which was also reported in #3031 and possibly #2285


- Vulkan is failing with no errors?

I can't reproduce these locally through `llvmpipe (LLVM 14.0.6, 256 bits)`/`win11 warp` drivers.

@cwfitzgerald any ideas?

cwfitzgerald:
> Vulkan is failing with no errors?

Vulkan is segfaulting

> DX12 is failing with

That's weird, I wonder if the problems are fundamentally related. Especially as swiftshader's cpu based execution will be sensitive to various memory corruption.

We have quite a few issues with freeing too early unfortunately.

teoxoy:
I managed to reproduce the DX12 errors locally and filed #3193.

I'll test swiftshader next.

> We have quite a few issues with freeing too early unfortunately.

Can you point me to some?

cwfitzgerald:
I re-ran it, but it still segfaults

teoxoy:
```
[2023-01-25T17:46:25Z ERROR wgpu_hal::gles::egl] EGL 'eglInitialize' code 0x3001: DRI2: failed to load driver
thread 'shader::zero_init_workgroup_mem::zero_init_workgroup_mem' panicked at 'Zero-initialization of workgroup memory failed (12% of disptaches failed).', wgpu/tests/shader/zero_init_workgroup_mem.rs:173:5
```
I'm confused about the new CI error. Did it fail to load the driver or did it run and 12% of dispatches failed to be 0 initialized?

jimblandy (approved):
Looks good to me.

Issue #3135, opened 2022-10-22
Feature request: disable uniformity analysis on shader creation


raphlinus:
Uniformity analysis of WGSL is too strict and fails to accept some valid shaders, including the coarse rasterization stage of [piet-gpu](https://github.com/linebender/piet-gpu). This is captured in an [upstream WGSL bug](https://github.com/gpuweb/gpuweb/issues/3479) and mitigations are being discussed. One such mitigation is to opt-out of the analysis.

I'd like a flag to disable uniformity analysis at shader creation time. That flag already exists in the naga API as `ValidationFlags`, but that's currently hardcoded to `all()` in wgpu-core's `Device::create_shader_module()` method, so basically what I'm asking is for that to be plumbed through.

The feature can be a temporary stopgap, and should probably be marked as such, until the issue is resolved upstream.

ErichDonGubler:
Related/perhaps a dupe: https://github.com/gfx-rs/wgpu/issues/5320

ErichDonGubler:
@raphlinus: Would implementing the WGSL diagnostic filter management per https://github.com/gfx-rs/wgpu/issues/5320 satisfy this particular request? That's not the same as your request to expose this on the API side, but if you're willing to add `diagnostic` directives in shader source, this might be a standards-supported way to do this.

PR #3175, opened 2022-11-03
Allow unchecked shader modules to also skip (naga's) uniformity checks

DJMcNab:
**Checklist**

- [x] Run `cargo clippy` 
   Ran but couldn't find any non-pre-existing errors. Didn't do a complete diff as that felt like a waste of time.
   I think not applicable, but ran anyway - no warnings. Please clarify what "if applicable" means here.

**Connections**
Modelled after https://github.com/gfx-rs/wgpu/pull/1978, using the same struct this already created.
Fixes https://github.com/gfx-rs/wgpu/issues/3135
The changes might be removable after https://github.com/gpuweb/gpuweb/issues/3479 is resolved and filters through

**Description**
When using shaders with `wgpu`, it may be necessary to work around https://github.com/gpuweb/gpuweb/issues/3479.
To allow this, `Device::create_shader_module_unchecked` has gained a parameter for configuration. This allows This disabling [`ValidationFlags::CONTROL_FLOW_UNIFORMITY`](https://docs.rs/naga/latest/naga/valid/struct.ValidationFlags.html#associatedconstant.CONTROL_FLOW_UNIFORMITY) when used, but may allow maintaining bounds checks.

**Testing**
I don't believe that we should expect much use of this API; it should primarily be used for exploratory work.

TODO, awaiting results from @raphlinus

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -34,12 +34,12 @@ pub use wgt::{
     Face, Features, FilterMode, FrontFace, ImageDataLayout, ImageSubresourceRange, IndexFormat,
     Limits, MultisampleState, Origin3d, PipelineStatisticsTypes, PolygonMode, PowerPreference,
     PresentMode, PrimitiveState, PrimitiveTopology, PushConstantRange, QueryType,
-    RenderBundleDepthStencil, SamplerBindingType, SamplerBorderColor, ShaderLocation, ShaderModel,
-    ShaderStages, StencilFaceState, StencilOperation, StencilState, StorageTextureAccess,
-    SurfaceConfiguration, SurfaceStatus, TextureAspect, TextureDimension, TextureFormat,
-    TextureFormatFeatureFlags, TextureFormatFeatures, TextureSampleType, TextureUsages,
-    TextureViewDimension, VertexAttribute, VertexFormat, VertexStepMode, COPY_BUFFER_ALIGNMENT,
-    COPY_BYTES_PER_ROW_ALIGNMENT, MAP_ALIGNMENT, PUSH_CONSTANT_ALIGNMENT,
+    RenderBundleDepthStencil, SamplerBindingType, SamplerBorderColor, ShaderBoundChecks,
+    ShaderLocation, ShaderModel, ShaderStages, StencilFaceState, StencilOperation, StencilState,
+    StorageTextureAccess, SurfaceConfiguration, SurfaceStatus, TextureAspect, TextureDimension,
+    TextureFormat, TextureFormatFeatureFlags, TextureFormatFeatures, TextureSampleType,
+    TextureUsages, TextureViewDimension, VertexAttribute, VertexFormat, VertexStepMode,
+    COPY_BUFFER_ALIGNMENT, COPY_BYTES_PER_ROW_ALIGNMENT, MAP_ALIGNMENT, PUSH_CONSTANT_ALIGNMENT,
     QUERY_RESOLVE_BUFFER_ALIGNMENT, QUERY_SET_MAX_QUERIES, QUERY_SIZE, VERTEX_STRIDE_ALIGNMENT,
 };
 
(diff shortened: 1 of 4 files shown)


wgpu/src/lib.rs
```diff
@@ -2124,6 +2124,31 @@ impl Device {
...
+    ///
+    /// # Safety
+    /// In contrast with [`create_shader_module`](Self::create_shader_module) this function
+    /// creates a shader module without unifromity checks which allows shaders to perform
+    /// operations which can lead to undefined behavior like data races, thus it's
+    /// the caller responsibility to pass a shader which doesn't perform any of these
+    /// operations.
+    ///
+    /// This is equivalent to `create_shader_module` effect on web.
+    pub unsafe fn create_shader_module_non_uniform(
```
cwfitzgerald, line 2137: I think we should combine this function with the `create_shader_module_unchecked` into a function that just takes an explicilt `wgt::ShaderBoundChecks` argument.

wgpu/src/lib.rs
```diff
@@ -2124,6 +2124,31 @@ impl Device {
...
+    /// Creates a shader module from either SPIR-V or WGSL source code without uniformity checks.
+    ///
+    /// # Safety
+    /// In contrast with [`create_shader_module`](Self::create_shader_module) this function
+    /// creates a shader module without unifromity checks which allows shaders to perform
+    /// operations which can lead to undefined behavior like data races, thus it's
+    /// the caller responsibility to pass a shader which doesn't perform any of these
+    /// operations.
+    ///
+    /// This is equivalent to `create_shader_module` effect on web.
```
cwfitzgerald, line 2136: This isn't actually equivalent. With disabling bounds checks, it doesn't actually change the validity of the shader. This method makes otherwise invalid shaders valid. I think we need a proper feature for this. Maybe `UNCHECKED_SHADERS` to cover any future validation restrictions too.

cwfitzgerald (changes requested):
Thanks for this, some comments

wgpu/src/lib.rs
DJMcNab: I'd avoided doing this because I couldn't find a case where `wgpu_types` types were exposed in the public API of wgpu.

If that's permissible then I'll add that.

wgpu/src/lib.rs
DJMcNab: Ah, I was coming at this from 'web=webgpu', which should have uniformity analysis forgetting the 'web=webgl' case. Thanks for the reminder

wgpu/src/lib.rs
DJMcNab: Oh, it's called `wgt`, and so searching for `wgpu_types` in the `wgpu` folder gives no useful results.

That's not very helpful

wgpu/src/lib.rs
cwfitzgerald: My point is more that, when calling *_unchecked, it doesn't matter if there are bounds checks in the shader or not, so the user doesn't need to know if eliding bounds checks are supported.

When calling `*_non_uniform`, it _does_ matter if eliding uniformity analysis is supported, because some shaders won't compile. Because of this, we need some way to inform the user if eliding uniformity analysis is supported on the device.

DJMcNab: Possibly? I'd hope use of this API would be relatively rare. 

More precisely, I think that the non-uniform part of this API is kind of inherently tied to working around naga, with knowledge of the code it emits to the backend. It's unclear to me what cases are valid wgsl if they don't pass uniformity analysis. That is, if you have a version of your code which uses the fact that uniformity analysis can be disabled, and another one which doesn't, you should always use the one which doesn't.

raphlinus: If the workgroupBroadcast proposal is accepted (and implemented in naga), then that would be my preferred way to avoid triggering uniformity analysis errors or warnings, and then I would not use this API. However, it might still make sense as a short term solution.

DJMcNab:
Closing this for now.

We will revisit if and when naga's uniformity check is improved such that we need the workaround, but at the moment we can just make use of the fact naga doesn't actually check uniformity for the scenario we need.

Issue #274, opened 2019-08-08
Shared Textures with Native Backend


jinleili:
On camera, video or AR apps, pixel data updated every frame, currently, we use ```wgpu_command_buffer_copy_buffer_to_texture``` to fill a texture,  this command will cause pixel copy.

On these scenario, we need more efficient way to fill a wgpu texture, maybe shared memory with app client?

grenlight:
On iOS and macOS, camera or video decode generated data is called ```CVPixelBuffer```,on Android, camera’s frame data is called ```byte[]``` , but video decode data is called ```ByteBuffer```, I understand that it may be  hard to provide a general ```fn``` wrap a  texture.

kvark:
This is a much needed issue, which also happens to be very hard to bring in. "wgpu-native" attempts to be safe and portable. Having externally managed texture contents means it's not directly in control of our implementation, and thus it's difficult to make it safe. Portability is also thrown out because these video subsystems are all different in different platforms.

First step - we need some way to provide external texture data to gfx-hal backends. Second step would be to make our tracking aware of external resources, investigate the needs for state transitions there, etc. It's a whole ton of work.

bes:
Sorry for bumping this, but I am looking to start experimenting on converting our video pipeline (ios/web/android/desktop/headless) which has implementation copied and slightly different for each platform into wgpu.

I am trying to get a feel for how far into the future this support would be, if there has been any movement?

As stated each platform is different, and I guess it has overlap with https://github.com/gfx-rs/wgpu/issues/274 and maybe other issues as well.

Sorry for bumping an old issue :(

kvark:
There's been a ton of changes since this issue was updated. The last state of things is: we still want this.
One way to get there is via #1609 on Vulkan only. Another way would be to rely on texture sharing (between backends, and backend-users), which we haven't implemented yet.

bes:
Thanks for the update @kvark, I would be happy with an OpenGL-only solution for now, but it seems I am out of luck. Would be happy to help with testing or other simliar things if needed. I really like this project 👍

Xaeroxe:
I'm interested in making shared textures happen and might be able to provide some labor towards it. I have a specific focus on Windows, which means I'd like `wgpu` to be able to interact with shared resources from DirectX 11 and DirectX 12. Is there any work towards this? Do we have any specifications or prior art for how we'd like this feature to work?

kvark:
[...]

PR #3177, opened 2022-11-04
vk: add external memory extensions for Android

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
https://github.com/gfx-rs/wgpu/issues/274

**Description**
To implement zero-copy external memory access on Android devices, need to use [AHardwareBuffer](https://developer.android.com/ndk/reference/group/a-hardware-buffer) to create `vk::Image`, and to use AHardwareBuffer, `VK_ANDROID_external_memory_android_hardware_buffer` and related extensions are required.

I looked at some official Google examples, and these extensions should be supported by all Android devices:
https://github.com/android/renderscript-samples/blob/08d5130faeae4d21005b0f043959243b85c73d89/RenderScriptMigrationSample/app/src/main/cpp/VulkanContext.cpp#L157-L163

**Testing**
Tested on Android 10.0 and Android 12.1.1

### DIFF
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -600,6 +600,17 @@ impl PhysicalDeviceCapabilities {
         #[cfg(any(target_os = "macos", target_os = "ios"))]
         extensions.push(vk::KhrPortabilitySubsetFn::name());
 
+        #[cfg(target_os = "android")]
+        {
+            extensions.append(&mut vec![
+                vk::KhrBindMemory2Fn::name(),
+                vk::KhrExternalMemoryFn::name(),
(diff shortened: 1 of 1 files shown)


Gordon-F:
> I looked at some official Google examples, and these extensions should be supported by all Android devices

Based on [gpuinfo](http://www.vulkan.gpuinfo.org/listdevicescoverage.php?extension=VK_ANDROID_external_memory_android_hardware_buffer&platform=android&option=not), not all devices support these extensions. I think we should check, that the device support that extension before using it.

jinleili:
Based on the discussion on #wgpu:matrix.org:
> there was discussion around whether wgpu-hal should support the memory import inside itself or not.
> ...
> we need a proper mechanism for enabling extensions that avoids the various issues

 Maybe I should close this.

PR #3184, opened 2022-11-07
Migrate to Rust 1.65

ErichDonGubler:
**Checklist**

- [x] Run `cargo clippy`.
- [x] ~Add change to CHANGELOG.md. See simple instructions inside file.~ Opting not to do this, unless somebody thinks it's actually interesting for downstream.

**Connections**

None that I know of.

**Description**

The Rust 1.65 toolchain has introduced several new lint warnings for this repository. Keep it clean by taking the most straightforward path for resolving lints. The only change that I think is controversial is the final change to appease [`clippy::result_large_err`](https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err), which adds a `.clippy.toml` to set the lint threshold to right about our current largest case.

Recommended review experience: commit-by-commit.

**Testing**

It still compiles, so we should be good to go!


wgpu-hal/src/dx12/device.rs
```diff
@@ -1358,8 +1358,8 @@ impl crate::Device<super::Api> for super::Device {
             DepthBias: bias.constant,
             DepthBiasClamp: bias.clamp,
             SlopeScaledDepthBias: bias.slope_scale,
-            DepthClipEnable: if desc.primitive.unclipped_depth { 0 } else { 1 },
-            MultisampleEnable: if desc.multisample.count > 1 { 1 } else { 0 },
+            DepthClipEnable: desc.primitive.unclipped_depth.into(),
```
teoxoy, line 1361: This seems to now be inverted.

ErichDonGubler: D'oh, good catch. Fixed with d210a2b13779df16a581097e6adccfe3169adb21.

ErichDonGubler:
Mmmm, looks like CI is failing because the `large-error-threshold` configuration I added to the new `.clippy.toml` is rejected in the Rust 1.64 toolchain. That means if we want to go with that solution, we'd effectively have to do it in lockstep with an upgrade to Rust 1.65. Would it be a problem to upgrade that here, too (changing the PR title, etc., OFC)? 🤔

wgpu-hal/src/dx12/conv.rs
```diff
@@ -329,14 +329,14 @@ fn map_stencil_face(face: &wgt::StencilFaceState) -> d3d12::D3D12_DEPTH_STENCILO
 
 pub fn map_depth_stencil(ds: &wgt::DepthStencilState) -> d3d12::D3D12_DEPTH_STENCIL_DESC {
     d3d12::D3D12_DEPTH_STENCIL_DESC {
-        DepthEnable: if ds.is_depth_enabled() { 1 } else { 0 },
+        DepthEnable: ds.is_depth_enabled().into(),
```
cwfitzgerald, line 332: I'm pretty biased against into - I would either use as or do the explicit if.

cwfitzgerald (commented):
Small comment on one thing

wgpu-hal/src/dx12/conv.rs
ErichDonGubler: Why is that? 🤔 The opinion at [`clippy::bool_to_int_with_if`](https://rust-lang.github.io/rust-clippy/master/index.html#bool_to_int_with_if) is that `into` is actually very idiomatic. Are you worried about problems like a foot-gun here?

Whatever the case, I don't mind reverting; this would require an alternate strategy, though, like `#![allow(clippy::bool_to_int_with_if)]` on crates, since it's now a warn-by-default lint.

teoxoy: I personally use `T::from(v)` more often than `.into()`

cwfitzgerald: I agree with the lint that the if statement is unnecessary, I don't like into because it hides the destination type from the source, so unless you have an ide, you don't know the type. From would be fine, as would an explicit as.

We use `as _` around a bit and we probably shouldn't be either, but at least you know it's a primative cast.

README.md
```diff
@@ -31,8 +31,8 @@ For an overview of all the components in the gfx-rs ecosystem, see [the big pict
 
 ### MSRV policy
 
-Minimum Supported Rust Version is **1.64**.
-It is enforced on CI (in "/.github/workflows/ci.yml") with `RUST_VERSION` variable.
+Minimum Supported Rust Version is **1.65**.
+It is enforced on CI (in "/.github/workflows/ci.yml", "/.github/workflows/cts.yml") with `RUST_VERSION` variable.
```
cwfitzgerald, line 35: I wouldn't bother mentioning this as the cts file is likely going to go away soon.

ErichDonGubler: Reverted this hunk with c14908b5c9c53479a2c6433d3ac5dadd7c85250a.

wgpu-hal/src/dx12/conv.rs
ErichDonGubler: Move to use `From` with 342b36e5.

PR #3186, opened 2022-11-07
fix(wgpu-info): s/driver/&_info for `DriverInfo` line

ErichDonGubler:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

**Description**

Using the `driver` member instead of the `driver_info` member looks like a mistake here; we already print `driver` on the line right above, and the current line differs with a `DriverInfo` prepend.

**Testing**


grovesNL (approved):
Good catch, thank you!

PR #3183, opened 2022-11-07
allow gles float textures to be multisampled if available

Wumpf:
**Checklist**

- [x] Run `cargo clippy`.
- [x] Run `RUSTFLAGS=--cfg=web_sys_unstable_apis cargo clippy --target wasm32-unknown-unknown` if applicable.
- [x] Add change to CHANGELOG.md. See simple instructions inside file.

**Connections**
n/a

**Description**
Couldn't find any restrictions on render target use for enabled float textures on GLES. It seems MSAA use works just fine iff the respective float extensions are enabled.

**Testing**
Tested rendering to a RGBA32f and RGBA16f target on WebGl. Did not check on resolve

PR #3187, opened 2022-11-07
Alpha to coverage support for GLES

Wumpf:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Fixed float multisample target support in #3183 

**Description**
MSAA alpha to coverage support for GLES!

**Testing**
Tested on WebGL:

after/before
image

(these are sphere faking quads with analytic pixel coverage in their alpha channel the same way as performed here https://www.shadertoy.com/view/MsSSWV)


cwfitzgerald (approved):
Dope

Issue #3190, opened 2022-11-08
Metal hal seems to accumulate CAMetalLayers in view on iOS


jimblandy:
On iOS, this code seems to be appending a new sublayer to the UIView's main layer each time the function is called:
https://github.com/gfx-rs/wgpu/blob/a377ae2b7fe6c1c9412751166f0917e617164e49/wgpu-hal/src/metal/surface.rs#L125-L134

If the `view`'s `layer` is not a `CAMetalLayer`, we create one, and add to the end of `view`'s `layer`'s list of sublayers. But the `view`'s layer itself doesn't change, so if we call this function again, we're just going to create and add a new `CAMetalLayer`, not return the one we created last time, which seems to be what the function is trying to do.

I'm not sure this is actually a bug; I just noticed it while looking at the code. Putting it here just in case.

jimblandy:
Cc @jinleili

jinleili:
get_metal_layer is actually only called when `create_surface()`, so is it enough to change visibility to pub(crate)?

jimblandy:
If that's the case, then is the code that checks for an existing `CAMetalLayer` needed? Will people ever pass in views that already have a CAMetalLayer?

jimblandy:
Yes, making it only callable from the crate would make sense.

It confused me because it looked like it was designed to be idempotent, but then it wasn't.

jinleili:
> If that's the case, then is the code that checks for an existing `CAMetalLayer` needed? Will people ever pass in views that already have a CAMetalLayer?

It is indeed possible for a user to pass a UIView with CAMetalLayer, and I need to think about how to better check CAMetalLayer.

PR #3192, opened 2022-11-08
Change `get_metal_layer()` visiability to pub(crate)

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
https://github.com/gfx-rs/wgpu/issues/3190

**Description**

I think we need to keep things simple, because on both Mac and iOS, layer can add sublayers  and keep nesting that way, and it is possible for a CAMetalLayer to be added as a sublayer's sublayer  of a layer by the user, but that is not a best practice.  
  
Also checked the approach of detecting CAMetalLayer within `winit`, although the implementation is a bit different, the code is equivalent: it only detects if the main layer is a CAMetalLayer.  
https://github.com/rust-windowing/winit/blob/08f9e374e0111168c282d0cfffbcd469dd8fdcb0/src/platform_impl/ios/window.rs#L432-L434
  
The reason why CAMetalLayer is added via sub layer on the iOS side when does not exist is that the main layer in iOS UIView can only be overridden by class method, and the main layer of the UIView instance cannot be replaced at runtime.  

**Testing**

Tested locally on Metal and Vulkan(vulkan-portability)

### DIFF
--- a/wgpu-hal/src/metal/surface.rs
+++ b/wgpu-hal/src/metal/surface.rs
@@ -98,7 +98,7 @@ impl super::Surface {
     }
 
     /// If not called on the main thread, this will panic.
-    pub unsafe fn get_metal_layer(
+    pub(crate) unsafe fn get_metal_layer(
         view: *mut Object,
         delegate: Option<&HalManagedMetalLayerDelegate>,
     ) -> *mut Object {


jimblandy (approved):
Sweet and simple.

PR #3194, opened 2022-11-08
refactor(dx12): remove `unsafe` ops in `Adapter::texture_format_capabilities`

ErichDonGubler:
**Checklist**

- [x] Run `cargo clippy`.
- [x] ~Add change to CHANGELOG.md. See simple instructions inside file.~ I don't think that this is necessary.

**Connections**

**Description**

The code being changed here used `unsafe` and `as` casting unnecessarily. Both are unusually fraught parts of writing in Rust. So, where possible, we'd like to avoid them. In this case…we can!

**Testing**

It still compiles, and I feel confident that the changes are correct, but I haven't tested locally. I can test this locally, though, if you wish.

### DIFF
--- a/wgpu-hal/src/dx12/adapter.rs
+++ b/wgpu-hal/src/dx12/adapter.rs
@@ -2,9 +2,9 @@ use crate::{
     auxil::{self, dxgi::result::HResult as _},
     dx12::SurfaceTarget,
 };
-use std::{mem, sync::Arc, thread};
+use std::{mem, ptr, sync::Arc, thread};
 use winapi::{
-    shared::{dxgi, dxgi1_2, windef, winerror},
+    shared::{dxgi, dxgi1_2, minwindef::DWORD, windef, winerror},
     um::{d3d12, d3d12sdklayers, winuser},
 };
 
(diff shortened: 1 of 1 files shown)


wgpu-hal/src/dx12/adapter.rs
```diff
@@ -381,17 +381,18 @@ impl crate::Adapter<super::Api> for super::Adapter {
...
         if raw_format != no_depth_format {
             // Only-recheck if we're using a different format
             assert_eq!(
                 winerror::S_OK,
                 self.device.CheckFeatureSupport(
                     d3d12::D3D12_FEATURE_FORMAT_SUPPORT,
-                    &mut data_no_depth as *mut _ as *mut _,
-                    mem::size_of::<d3d12::D3D12_FEATURE_DATA_FORMAT_SUPPORT>() as _,
+                    ptr::addr_of_mut!(data_no_depth).cast::<c_void>(),
+                    DWORD::try_from(mem::size_of::<d3d12::D3D12_FEATURE_DATA_FORMAT_SUPPORT>())
```
cwfitzgerald, line 394: My only hesitation is the verbosity of this

ErichDonGubler: Coming back to this with a fresh brain:

1. RE: pointer slinging:
	1. We can definitely omit the explicit `::<c_void>`. I've done so with 7c46b4b824e69b06df1d60b0b48c563b14251986.
	1. We could fully qualify the `addr_of_mut` symbol, which would make this shorter and, IMO, easier to follow. In general, I think a _lot_ of this code would be smaller with no loss of clarity (and therefore more friendly to the mental cache) without path qualifications for most symbols.
1. I've seen folks like @retep998 make utility crates for Windows specifically to convert `usize`s to `DWORD`s. Maybe adding something like an extension trait for this in a separate commit would be good?

Thoughts?

cwfitzgerald: Honestly now that I look at this, this seems fine.

ErichDonGubler:
The `wgpu::example/hello-compute tests::test_multithreaded_compute` test in Linux CI is currently failing with:

```
[2022-11-28T19:23:18Z ERROR wgpu_hal::gles::egl] EGL 'eglInitialize' code 0x3001: DRI2: failed to load driver
corrupted double-linked list
```

...where `0x3001 == EGL_NOT_INITIALIZED`.

I don't think that this failure is related to this change, but I'm not positive. Do you know, @cwfitzgerald?

cwfitzgerald:
Yeah, llvmpipe has been crapping the bed on the multi-threaded test of late.

cwfitzgerald (approved):

PR #3217, opened 2022-11-18
Revert to rust 1.63

jimblandy:
Firefox is still using an MSRV of 1.63, and wgpu currently uses 1.65.
The 1.64->1.65 jump is from [#3184](https://github.com/gfx-rs/wgpu/pull/3184), which I merged. Egg on my face.
The 1.63->1.64 jump is from [#3107](https://github.com/gfx-rs/wgpu/pull/3107).
We should have a comment in the top-level Cargo.toml about this, to catch this problem in the future.

**Checklist**

- [ ] Run `cargo clippy`.

**Connections**

**Description**

**Testing**


jimblandy:
This depends on #3226.

jimblandy:
We don't need this much.

PR #3227, opened 2022-11-22
vk: eliminate duplicate values in SurfaceCapabilities's formats

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Description**

`Vec<vk::SurfaceFormatKHR>` obtained by `get_physical_device_surface_formats` contains all supported color spaces and matching formats. Typically, a format appears in multiple color spaces. So currently in the vk backend, `get_supported_formats` will return duplicate `TextureFormat`'s : 
```sh
get_supported_formats: [Bgra8UnormSrgb, Bgra8UnormSrgb, Bgra8UnormSrgb, Bgra8UnormSrgb, Bgra8UnormSrgb, Bgra8UnormSrgb, Bgra8UnormSrgb, Bgra8UnormSrgb, Bgra8UnormSrgb, Bgra8UnormSrgb, Bgra8UnormSrgb, Bgra8UnormSrgb, 
Bgra8Unorm, Rgba16Float, Rgb10a2Unorm, Bgra8Unorm, Rgba16Float, Rgb10a2Unorm, Bgra8Unorm, Rgba16Float, Rgb10a2Unorm, Bgra8Unorm, Rgba16Float, Rgb10a2Unorm, Bgra8Unorm, Rgba16Float, Rgb10a2Unorm, Bgra8Unorm, Rgba16Float, Rgb10a2Unorm, Bgra8Unorm, Rgba16Float, 
Rgb10a2Unorm, Bgra8Unorm, Rgba16Float, Rgb10a2Unorm, Bgra8Unorm, Rgba16Float, Rgb10a2Unorm, Bgra8Unorm, Rgba16Float, Rgb10a2Unorm, Bgra8Unorm, Rgba16Float, Rgb10a2Unorm, Bgra8Unorm, Rgba16Float, Rgb10a2Unorm]
```

Since `TextureFormat` does not derive `Ord` tait,   cannot directly call `sort_xx()` and then `dedup()`, a HashSet is used to help detect uniqueness.

**Testing**
Tested locally

### DIFF
--- a/wgpu-hal/src/vulkan/conv.rs
+++ b/wgpu-hal/src/vulkan/conv.rs
@@ -149,16 +149,22 @@ pub fn map_vk_surface_formats(sf: vk::SurfaceFormatKHR) -> Option<wgt::TextureFo
     use ash::vk::Format as F;
     use wgt::TextureFormat as Tf;
     // List we care about pulled from https://vulkan.gpuinfo.org/listsurfaceformats.php
-    Some(match sf.format {
-        F::B8G8R8A8_UNORM => Tf::Bgra8Unorm,
-        F::B8G8R8A8_SRGB => Tf::Bgra8UnormSrgb,
-        F::R8G8B8A8_SNORM => Tf::Rgba8Snorm,
-        F::R8G8B8A8_UNORM => Tf::Rgba8Unorm,
-        F::R8G8B8A8_SRGB => Tf::Rgba8UnormSrgb,
-        F::R16G16B16A16_SFLOAT => Tf::Rgba16Float,
-        F::R16G16B16A16_SNORM => Tf::Rgba16Snorm,
-        F::R16G16B16A16_UNORM => Tf::Rgba16Unorm,
-        F::A2B10G10R10_UNORM_PACK32 => Tf::Rgb10a2Unorm,
+    Some(match sf.color_space {
+        vk::ColorSpaceKHR::SRGB_NONLINEAR => match sf.format {
+            F::B8G8R8A8_UNORM => Tf::Bgra8Unorm,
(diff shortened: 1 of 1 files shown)


wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -1565,9 +1565,11 @@ impl crate::Adapter<super::Api> for super::Adapter {
             }
         };
 
+        let mut uniques = std::collections::HashSet::new();
```
cwfitzgerald, line 1568: This is a kinda weird pattern, could you add a comment that explains what's going on.

cwfitzgerald (changes requested):

cwfitzgerald (changes requested):
I realize this solution is actually wrong, you need to filter out the formats that have a differing color space than we want to use (see https://github.com/gfx-rs/wgpu/blob/master/wgpu-hal/src/vulkan/device.rs#L542-L548). It's possible for a format to only be supported with a different color space than we want, and would cause UB on surface creation.

jinleili:
This commit also wrong, I'll recommit later.

cwfitzgerald (approved):
Nit then G2G

MarijnS95:
Note that this PR was found to regress 10-bit (and 16-bit non-SFLOAT...) UNORM/SNORM formats that can now no longer be used with non-linear sRGB. Despite the code linked from @cwfitzgerald (false location turned back into a permalink: https://github.com/gfx-rs/wgpu/blob/f41a1c294bf09c4ab7e523f6c9aff5119044a097/wgpu-hal/src/vulkan/device.rs#L542-L548) only selecting extended scRGB for fp16 texture/surface format and the regular non-linear sRGB for everything else.

I've moved those back to the correct match arm in https://github.com/gfx-rs/wgpu/pull/8226.

PR #3229, opened 2022-11-23
Allow unspecified bits when deserializing the API's bitflags

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

https://bugzilla.mozilla.org/show_bug.cgi?id=1771246 and a host of similar fuzz bugs where web content serves us unspecified bits in some of the API's bitflags.

**Description**

The bitflags_serde_shim crate treats unspecified bits as errors during deserialization. To make the browser integration easier, we want the unsepcified bits (which can be produced by web content) to turn into validation errors instead of panics during deserialization.

Thankfully the bitflags_serde_shim crate is quite trivial, so this PR inlines the macro with the required modification to not check the bits. In addition, bitflags_serde_shim feature is removed, the Serialize and Deserialize implementations are always the same and only their presence depend on the trace and replay feature flags.

**Testing**

A test for invalid buffer usage bits in create_buffer.


nical:
I'll follow this up with extra checks in wgpu-core.

wgpu-core/src/binding_model.rs
```diff
@@ -50,6 +50,8 @@ pub enum CreateBindGroupLayoutError {
     TooManyBindings(BindingTypeMaxCountError),
     #[error("Binding index {binding} is greater than the maximum index {maximum}")]
     InvalidBindingIndex { binding: u32, maximum: u32 },
+    #[error("Invalid visibility {0:?}")]
```
cwfitzgerald, line 53: This will likely print nothing if made up of fully invalid bits. We should probably print the raw u32 or something.

wgpu-types/src/lib.rs
```diff
@@ -14,6 +14,48 @@ use serde::{Deserialize, Serialize};
 use std::hash::{Hash, Hasher};
 use std::{num::NonZeroU32, ops::Range};
 
+// Use this macro instead of the one provided by the bitflags_serde_shim crate
+// because the latter produces an error when deserializing bits that are not
+// specified in the bitflags, while we want deserialization to succeed and
+// and unspecified bits to lead to errors handled in wgpu-core.
+// Note that plainly deriving Serialize and Deserialized would have a similar
```
cwfitzgerald, line 21: Any reason we can't just do that?

wgpu-types/src/lib.rs
```diff
@@ -14,6 +14,48 @@ use serde::{Deserialize, Serialize};
...
+                Ok(unsafe { $name::from_bits_unchecked(value) })
+            }
+        }
+
+        impl $name {
+            /// Returns true if the bitflags contains bits that are not part of
+            /// the bitflags definition.
+            pub fn contains_invalid_bits(&self) -> bool {
+                let all = Self::all().bits();
+                self.bits() | all != all
```
cwfitzgerald, line 53: This has _very_ confusing order of operations, I stared at this for like a minute to get what it's doing :)

cwfitzgerald (changes requested):

wgpu-core/src/binding_model.rs
nical: bitflags handles it surprisingly well, for example:

```rust
let flags = unsafe { wgpu::BufferUsages::from_bits_unchecked(0b1100000001) };
println!("{:?}", flags);
```

prints `MAP_READ | INDIRECT | 0x200`

wgpu-types/src/lib.rs
nical: I am not sure. [The commit that introduced the bitflags serde shim](https://github.com/gfx-rs/wgpu/commit/d5ba0b439d5722980ab958842fa0fe23112548b0) was about deno/cts and doesn't explain why. From a few questions on the chat, my understanding is that it gives you nicer looking serialization by avoiding to wrap all bitflags in single-member structs.

I don't mind removing the custom serialization code, although I'd keep the macro in order to have `contains_invalid_bits` so the cost of having the manual serialize/deserialize written once in the macro doesn't seem particularly bad to me. Let me know what you prefer.

wgpu-types/src/lib.rs
nical: Is it confusing because the negation? I read it as "the union of the real bits and the set of know bits is different from the set of known bits" or in other words "the set of real bits is not contained in the set of known bits".

I could rewrite this as "There are bits in the set of unknown bits", aka:  `self.bits() & !all != 0` which felt a bit more complicated to me because there are two negations instead of one but but everyone has their own way oflaying out this kind of logic in their head. Let me know what you prefer (or if you have a different formulation in mind).

cwfitzgerald: All I found confusing was the literal precidence of the operation. I kept trying to read it as `self.bits() | (all != all)` which obviously makes no sense. So just adding parens (or making the left side a variable) would clear it up for me.

wgpu-types/src/lib.rs
cwfitzgerald: The current way SGTM then.

wgpu-types/src/lib.rs
nical: Oh, I see, makes sense. Done.

cwfitzgerald (approved):

cwfitzgerald:
Got some merge conflicts

PR #3233, opened 2022-11-24
Fix incorrect offset in get_mapped_range

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Found the issue while fuzzing Gecko's WebGPU integration.

**Description**

This one is kind of serious.

`get_mapped_range` ([spec](https://gpuweb.github.io/gpuweb/#dom-gpubuffer-getmappedrange)) takes a range that is relative to the beginning of the buffer. The implementation in webgpu seems to want to agree judging by how the validation checks that the range is contained in the mapped one. However internally the `ptr` stored in  `BufferMapState::Active` is actually relative to the beginning of the mapped range (rather than beginning of the buffer). This means that we have to subtract the mapped offset to the one provided in `get_mapped_range`, otherwise seriously scary things happen.

**Testing**

Added a simple test.

Note that in the process of adding the test I found out that tests/buffer.rs was not run (oops!). I hadn't finished it when it landed a little while ago, it was not quite working yet because of how wgpu does not accept empty slices, I had put it on hold for a bit and when it landed anyway without breaking the build I didn't think too much of it but the version that landed was not part of the build. So this commit puts the file back in the build with a #[ignore] slapped on it, and of course it got reformatted in the process.

### DIFF
--- a/wgpu/tests/buffer.rs
+++ b/wgpu/tests/buffer.rs
@@ -1,9 +1,6 @@
 use crate::common::{initialize_test, TestParameters, TestingContext};
-use std::sync::{Arc, atomic::{AtomicBool, Ordering}};
 
 fn test_empty_buffer_range(ctx: &TestingContext, buffer_size: u64, label: &str) {
-    let status = Arc::new(AtomicBool::new(false));
-
     let r = wgpu::BufferUsages::MAP_READ;
     let rw = wgpu::BufferUsages::MAP_READ | wgpu::BufferUsages::MAP_WRITE;
     for usage in [r, rw] {
@@ -14,15 +11,10 @@ fn test_empty_buffer_range(ctx: &TestingContext, buffer_size: u64, label: &str) 
             mapped_at_creation: false,
         });
 
-        let done = status.clone();
-        b0.slice(0..0).map_async(wgpu::MapMode::Read, move |result| {
-            assert!(result.is_ok());
-            done.store(true, Ordering::SeqCst);
-        });
+        b0.slice(0..0)
+            .map_async(wgpu::MapMode::Read, Result::unwrap);
 
-        while !status.load(Ordering::SeqCst) {
-            ctx.device.poll(wgpu::MaintainBase::Poll);
-        }
+        ctx.device.poll(wgpu::MaintainBase::Wait);
 
         {
             let view = b0.slice(0..0).get_mapped_range();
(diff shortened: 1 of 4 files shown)


nical:
@cwfitzgerald Perhaps this one fixes a sufficiently bad issue to warrant being rebased on top of the current release and publishing it as `0.14.1`? It allows you to do arbitrary reads and writes past the buffer mapping without unsafe code.

wgpu/tests/buffer.rs
```diff
@@ -91,18 +97,100 @@ fn test_empty_buffer_range(ctx: &TestingContext, buffer_size: u64, label: &str)
...
+        write_buf
+            .slice(32..)
+            .map_async(wgpu::MapMode::Write, move |result| {
+                assert!(result.is_ok());
+                done.store(true, Ordering::SeqCst);
+            });
+
+        while !status.load(Ordering::SeqCst) {
+            ctx.device.poll(wgpu::MaintainBase::Poll);
+        }
```
cwfitzgerald, line 151: What's the point of this vs just waiting?

wgpu/tests/buffer.rs
```diff
@@ -91,18 +97,100 @@ fn test_empty_buffer_range(ctx: &TestingContext, buffer_size: u64, label: &str)
...
+            size: 256,
+            usage: wgpu::BufferUsages::MAP_READ | wgpu::BufferUsages::COPY_DST,
+            mapped_at_creation: false,
+        });
+
+        let done = status.clone();
+        write_buf
+            .slice(32..)
+            .map_async(wgpu::MapMode::Write, move |result| {
+                assert!(result.is_ok());
```
cwfitzgerald, line 145: lets do result.unwrap() as it gives a better message

wgpu/tests/buffer.rs
```diff
@@ -91,18 +97,100 @@ fn test_empty_buffer_range(ctx: &TestingContext, buffer_size: u64, label: &str)
 
     b1.unmap();
 
-    for _ in  0..10 {
+    for _ in 0..10 {
         ctx.device.poll(wgpu::MaintainBase::Poll);
     }
```
cwfitzgerald, line 102: This is simiarly weird

cwfitzgerald (changes requested):
Yeah that feels backport worthy.

Are you able to backport the change, and I'll take the release to the finish line?

wgpu/tests/buffer.rs
nical: Are you referring to using `MaintainBase::Wait` instead of `Poll` ? I don't know, both appear to work, changing it to Wait.

wgpu/tests/buffer.rs
nical: That's in the old unfinished stuff, I'm not sure what I meant by "poll ten times". My guess is that I probably wanted the test to get a chance to run unfinished async code in case there is anything queued up while not have anything in particular to wait for, if that makes any sense.

cwfitzgerald (changes requested):

nical:
The linux failure looks like an unrelated driver sneeze.

cwfitzgerald (approved):
Party

PR #3235, opened 2022-11-25
Document when types are not WebGPU types.

kpreid:
**Checklist**

- [X] Run `cargo clippy`.
- [ ] ~~Add change to CHANGELOG.md. See simple instructions inside file.~~ Skipping this

**Connections**
This is a probably-less-interesting side branch from #2653 where I added documentation linking to the WebGPU specification to many types.

**Description**
This change documents types that *don't* correspond to the WebGPU specification, so readers can know they are unique to `wgpu`.

I also changed the wording in `BufferSlice` to be more consistent with other documentation.

**Testing**
Only documentation was changed. I manually examined the rustdoc output.

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -590,6 +590,9 @@ static_assertions::assert_impl_all!(Device: Send, Sync);
 /// Identifier for a particular call to [`Queue::submit`]. Can be used
 /// as part of an argument to [`Device::poll`] to block for a particular
 /// submission to finish.
+///
+/// This type is unique to the Rust API of `wgpu`.
+/// There is no analogue in the WebGPU specification.
 #[derive(Debug, Copy, Clone)]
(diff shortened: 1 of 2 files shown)


cwfitzgerald:
The only thing I worry about is that people might interpret this as them being native features, not just that they're unique to wgpu.

kpreid:
Hmm. For that concern, what do you think of something like

> This type is unique to the Rust API of `wgpu`. In the WebGPU specification, {explain the corresponding thing} instead.

The idea being to emphasize that this is a “Rust bindings” thing.

cwfitzgerald:
That looks great

cwfitzgerald (changes requested):
For above change.

kpreid:
Done. For example, `Operations`' documentation now contains

> This type is unique to the Rust API of `wgpu`. In the WebGPU specification, separate `loadOp` and `storeOp` fields are used instead.

cwfitzgerald (approved):

PR #3237, opened 2022-11-26
Add `view_formats` in TextureDescriptor

### LINKED ISSUE #3030: Question: Can we create a linear-space views of SRGB textures or SRGB views of linear-space textures?
I have a pipeline that is SRGB aware, but after a certain point I'd like to treat it as `Rgba8Unorm`, because I composite UI on top, and there's colors there that users set in code, so their eye does the correction already.

Tried to achieve this with a texture that is `Rgba8Unorm`, but creating an `Rgba8UnormSrgb` write view and a `Rgba8Unorm` read view. Unfortunately, the validator doesn't like it much:

```

Caused by:
    In Texture::create_view
    Unable to view texture Rgba8Unorm as Rgba8UnormSrgb
```

Is there a reason this shouldn't be working?

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
WebGPU spec: https://www.w3.org/TR/webgpu/#dom-gputexturedescriptor-viewformats

**Description**

Closes #3030 

**Testing**
Tested on Metal, Vulkan(--features vulkan-portability) and Dx12 backend.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -852,14 +852,12 @@ impl<A: HalApi> Device<A> {
             ));
         }
 
-        if let Some(ref view_formats) = desc.view_formats {
-            for format in view_formats {
-                if desc.format == *format {
-                    continue;
-                }
-                if desc.format.remove_srgb_suffix() != format.remove_srgb_suffix() {
-                    return Err(CreateTextureError::InvalidViewFormat(*format, desc.format));
-                }
+        for format in desc.view_formats.iter() {
+            if desc.format == *format {
+                continue;
+            }
+            if desc.format.remove_srgb_suffix() != format.remove_srgb_suffix() {
+                return Err(CreateTextureError::InvalidViewFormat(*format, desc.format));
             }
         }
 
@@ -1094,17 +1092,11 @@ impl<A: HalApi> Device<A> {
             extent.depth_or_array_layers = view_layer_count;
         }
         let format = desc.format.unwrap_or(texture.desc.format);
-        if format != texture.desc.format {
-            let compatible = match texture.desc.view_formats {
-                Some(ref view_formats) => view_formats.contains(&format),
-                None => false,
-            };
-            if !compatible {
-                return Err(resource::CreateTextureViewError::FormatReinterpretation {
-                    texture: texture.desc.format,
-                    view: format,
-                });
-            }
+        if format != texture.desc.format && !texture.desc.view_formats.contains(&format) {
+            return Err(resource::CreateTextureViewError::FormatReinterpretation {
+                texture: texture.desc.format,
+                view: format,
+            });
         }
 
         // filter the usages based on the other criteria
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -1092,11 +1092,15 @@ impl<A: HalApi> Device<A> {
             extent.depth_or_array_layers = view_layer_count;
         }
         let format = desc.format.unwrap_or(texture.desc.format);
-        if format != texture.desc.format && !texture.desc.view_formats.contains(&format) {
-            return Err(resource::CreateTextureViewError::FormatReinterpretation {
-                texture: texture.desc.format,
-                view: format,
-            });
+        if format != texture.desc.format {
+            if texture.desc.view_formats.contains(&format) {
+                self.require_downlevel_flags(wgt::DownlevelFlags::VIEW_FORMATS)?;
+            } else {
+                return Err(resource::CreateTextureViewError::FormatReinterpretation {
+                    texture: texture.desc.format,
+                    view: format,
+                });
+            }
         }
 
         // filter the usages based on the other criteria
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -1093,14 +1093,13 @@ impl<A: HalApi> Device<A> {
         }
         let format = desc.format.unwrap_or(texture.desc.format);
         if format != texture.desc.format {
-            if texture.desc.view_formats.contains(&format) {
-                self.require_downlevel_flags(wgt::DownlevelFlags::VIEW_FORMATS)?;
-            } else {
+            if !texture.desc.view_formats.contains(&format) {
                 return Err(resource::CreateTextureViewError::FormatReinterpretation {
                     texture: texture.desc.format,
                     view: format,
                 });
             }
+            self.require_downlevel_flags(wgt::DownlevelFlags::VIEW_FORMATS)?;
         }
 
         // filter the usages based on the other criteria
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -852,7 +852,16 @@ impl<A: HalApi> Device<A> {
             ));
         }
 
-        // TODO: validate missing TextureDescriptor::view_formats.
+        if let Some(ref view_formats) = desc.view_formats {
+            for format in view_formats {
+                if desc.format == *format {
+                    continue;
+                }
+                if desc.format.remove_srgb_suffix() != format.remove_srgb_suffix() {
+                    return Err(CreateTextureError::InvalidViewFormat(*format, desc.format));
+                }
+            }
+        }
 
         // Enforce having COPY_DST/DEPTH_STENCIL_WRIT/COLOR_TARGET otherwise we
         // wouldn't be able to initialize the texture.
@@ -1086,10 +1095,16 @@ impl<A: HalApi> Device<A> {
         }
         let format = desc.format.unwrap_or(texture.desc.format);
         if format != texture.desc.format {
-            return Err(resource::CreateTextureViewError::FormatReinterpretation {
-                texture: texture.desc.format,
-                view: format,
-            });
+            let compatible = match texture.desc.view_formats {
+                Some(ref view_formats) => view_formats.contains(&format),
+                None => false,
+            };
+            if !compatible {
+                return Err(resource::CreateTextureViewError::FormatReinterpretation {
+                    texture: texture.desc.format,
+                    view: format,
+                });
+            }
         }
 
         // filter the usages based on the other criteria
--- a/wgpu-core/src/resource.rs
+++ b/wgpu-core/src/resource.rs
@@ -628,6 +628,8 @@ pub enum CreateTextureViewError {
         texture: wgt::TextureFormat,
         view: wgt::TextureFormat,
     },
+    #[error(transparent)]
+    MissingDownlevelFlags(#[from] MissingDownlevelFlags),
 }
 
 #[derive(Clone, Debug, Error)]
--- a/wgpu-core/src/resource.rs
+++ b/wgpu-core/src/resource.rs
@@ -507,7 +507,7 @@ pub enum CreateTextureError {
         if *.2 { " due to downlevel restrictions" } else { "" }
     )]
     InvalidFormatUsages(wgt::TextureUsages, wgt::TextureFormat, bool),
-    #[error("The view format {0:?} is not compatible with texture format {1:?}.")]
+    #[error("The view format {0:?} is not compatible with texture format {1:?}, only changing srgb-ness is allowed.")]
     InvalidViewFormat(wgt::TextureFormat, wgt::TextureFormat),
     #[error("Texture usages {0:?} are not allowed on a texture of dimensions {1:?}")]
     InvalidDimensionUsages(wgt::TextureUsages, wgt::TextureDimension),
--- a/wgpu-core/src/resource.rs
+++ b/wgpu-core/src/resource.rs
@@ -507,6 +507,8 @@ pub enum CreateTextureError {
         if *.2 { " due to downlevel restrictions" } else { "" }
     )]
     InvalidFormatUsages(wgt::TextureUsages, wgt::TextureFormat, bool),
+    #[error("The view format {0:?} is not compatible with texture format {1:?}.")]
+    InvalidViewFormat(wgt::TextureFormat, wgt::TextureFormat),
     #[error("Texture usages {0:?} are not allowed on a texture of dimensions {1:?}")]
     InvalidDimensionUsages(wgt::TextureUsages, wgt::TextureDimension),
     #[error("Texture usage STORAGE_BINDING is not allowed for multisampled textures")]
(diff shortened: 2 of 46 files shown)


crowlKats:
Could the deno webgpu part be implemented fully/properly?

wgpu-types/src/lib.rs
```diff
@@ -4248,7 +4271,10 @@ pub struct TextureDescriptor<L> {
     pub format: TextureFormat,
     /// Allowed usages of the texture. If used in other ways, the operation will panic.
     pub usage: TextureUsages,
-    // TODO: missing view_formats https://www.w3.org/TR/webgpu/#dom-gputexturedescriptor-viewformats
+    /// Specifies what view format values will be allowed when calling create_view() on this texture.
+    /// Note: Adding a format to this list may have a significant performance impact,
```
cwfitzgerald, line 4275: I'd causion against this verbiage, as afaik srgb-ness doesn't have significant cost.

wgpu-types/src/lib.rs
```diff
@@ -2522,6 +2522,29 @@ impl TextureFormat {
             },
         }
     }
+
+    /// Get texture view format compatible none `Srgb` suffix format.
```
cwfitzgerald, line 2526:
```suggestion
    /// Strips the `Srgb` suffix from the given texture format.
```

wgpu-core/src/resource.rs
```diff
@@ -507,6 +507,8 @@ pub enum CreateTextureError {
         if *.2 { " due to downlevel restrictions" } else { "" }
     )]
     InvalidFormatUsages(wgt::TextureUsages, wgt::TextureFormat, bool),
+    #[error("The view format {0:?} is not compatible with texture format {1:?}.")]
```
cwfitzgerald, line 510: Could we add to this error message that only changing srgb-ness is allowed.

wgpu-core/src/device/mod.rs
```diff
@@ -852,7 +852,16 @@ impl<A: HalApi> Device<A> {
             ));
         }
 
-        // TODO: validate missing TextureDescriptor::view_formats.
+        if let Some(ref view_formats) = desc.view_formats {
+            for format in view_formats {
+                if desc.format == *format {
```
cwfitzgerald, line 857: Does the spec require that the format list is unique?

cwfitzgerald (changes requested):
The code is fine, but there are two extra parts we need (besides the deno stuff).

First can you add a test to make sure the feature works (read a unorm texture as srgb and validate it does the conversion - and vise versa)

We also need a downlevel feature for this, as gles doesn't support texture views. And afaik we can't polyfill this feature in any way.

wgpu-core/src/device/mod.rs
jinleili: If I understand correctly, the spec does not require uniqueness of format list values.

jinleili:
@crowlKats I'm sorry, I don't know how  to properly implement view_formats in deno_webgpu. deno_webgpu's `webgpu.idl` doesn't have `viewFormats` field, is it enough to add `viewFormats` field to `deno_webgpu::texture::CreateTextureArgs`?

crowlKats:
no, its more than just that. but no worries, i will do it then

wgpu-types/src/lib.rs
```diff
@@ -4248,7 +4271,10 @@ pub struct TextureDescriptor<L> {
     pub format: TextureFormat,
     /// Allowed usages of the texture. If used in other ways, the operation will panic.
     pub usage: TextureUsages,
-    // TODO: missing view_formats https://www.w3.org/TR/webgpu/#dom-gputexturedescriptor-viewformats
+    /// Specifies what view format values will be allowed when calling create_view() on this texture.
```
cwfitzgerald, line 4274: This should note that it's _in addition_ to the real format of the texture.

wgpu-types/src/lib.rs
```diff
@@ -4248,7 +4271,10 @@ pub struct TextureDescriptor<L> {
...
+    /// Note: Adding a format to this list may have a significant performance impact,
+    /// so it is best to avoid adding formats unnecessarily.
+    pub view_formats: Option<Vec<TextureFormat>>,
```
cwfitzgerald, line 4277: I don't think the `Option` is actually necessary here.

jinleili: That means users need to use `vec![]` to set empty value, would this be inconsistent with the style of using other nullable values in wgpu?

cwfitzgerald: It should be - actually the issue is that we want it to be `&'a [TextureFormat]` so people can do `&[]` for the empty value.

jinleili: Setting the view_formats type in `wgt::TextureDescriptor` to `&'a [TextureFormat]` results in an error like this:
```
the trait bound `&[TextureFormat]: Deserialize<'_>` is not satisfied  
the following other types implement trait `Deserialize<'de>`:  
&'a [u8]  
[T; 0]  
[T; 10]  
[T; 11]  
[T; 12]  
[T; 13]  
[T; 14]  
[T; 15]  
and 26 othersrustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message?3#file%3A%2F%2F%2FUsers%2Flijinlei%2FRust%2Fforks%2Fwgpu%2Fwgpu-types%2Fsrc%2Flib.rs)
```

The current implementation seems a bit complicated, not sure if it's appropriate?

cwfitzgerald: I think it's just because TextureFormat doesn't impl deserialize/serialize

jinleili: TextureFormat was already implemented deserialize/serialize.

Setting to `&'a [TextureFormat]`  also causes another problem, having to add a lifetime parameter for `Texture` causes `pub type TextureId = Id<crate::resource::Texture<Dummy>>`  also have a lifetime parameter added, and `TextureId` should not have a lifetime parameter.

My current implementation is to redefine a `TextureDescriptor` in wgpu just like  `TextureViewDescriptor`.
but in `wgpu::util::device::create_texture_with_data` need to call some `wgt::TextureDescriptor` 's functions, Causes the code to be a bit weird.

khyperia:
Hi! I'm just a random user of wgpu, so take this whole comment with a grain of salt. Proper sRGB support in the webgpu backend of wgpu is blocked on specifying the view format of the Surface's current_texture (the swapchain image). In webgpu, this is done by setting GPUCanvasConfiguration.viewFormats to the formats you'd like to view the current_texture as, then specifying that format in create_view. (I guess this GPUCanvasConfiguration.viewFormats field is a sort of backdoor to pass in view_formats into the system's internal create_texture for the current_texture?)

https://www.w3.org/TR/webgpu/#canvas-configuration

In wgpu, the GPUCanvasConfiguration is created here, and I suppose `mapped.view_formats()` should be called, similar to what this PR is already doing in the device_create_texture function in web.rs https://github.com/gfx-rs/wgpu/blob/aa46e82f4087209b6a331220e37dca142d4ffed5/wgpu/src/backend/web.rs#L1037-L1041

I suppose SurfaceConfiguration should get a new `view_formats: Vec<TextureFormat>` field, similar to how `TextureDescriptor` got that field in this current PR? I haven't looked into how other backends other than webgpu handle the view format of the surface current_texture, and if they support alternate views at all, and how they're configured/enabled.

If you could implement this in this PR or a followup, that would be amazing, but please ping me if you don't plan on doing so and I can try looking into implementing it!

jinleili:
The implementation of this PR is not yet accepted and may need to be modified. I have plans to implement `view_formats` in `SurfaceConfiguration` after this PR.

wgpu-hal/src/gles/adapter.rs
```diff
@@ -283,7 +283,8 @@ impl super::Adapter {
         let mut downlevel_flags = wgt::DownlevelFlags::empty()
             | wgt::DownlevelFlags::NON_POWER_OF_TWO_MIPMAPPED_TEXTURES
             | wgt::DownlevelFlags::CUBE_ARRAY_TEXTURES
-            | wgt::DownlevelFlags::COMPARISON_SAMPLERS;
+            | wgt::DownlevelFlags::COMPARISON_SAMPLERS
+            | wgt::DownlevelFlags::VIEW_FORMATS;
```
cwfitzgerald, line 287: This is the one backend where this isn't supported

wgpu-core/src/device/mod.rs
```diff
@@ -1086,10 +1093,14 @@ impl<A: HalApi> Device<A> {
...
-                view: format,
-            });
+            if texture.desc.view_formats.contains(&format) {
+                self.require_downlevel_flags(wgt::DownlevelFlags::VIEW_FORMATS)?;
+            } else {
+                return Err(resource::CreateTextureViewError::FormatReinterpretation {
+                    texture: texture.desc.format,
+                    view: format,
+                });
+            }
```
cwfitzgerald, line 1103: This is a hyper-nit:

This would read better inverted:

```suggestion
            if !texture.desc.view_formats.contains(&format) {
                return Err(resource::CreateTextureViewError::FormatReinterpretation {
                    texture: texture.desc.format,
                    view: format,
                });
            }
            self.require_downlevel_flags(wgt::DownlevelFlags::VIEW_FORMATS)?;
```

As the requiring of the downlevel flags is actually unrelated to if the view format is in the supported list.

wgpu-types/src/lib.rs
cwfitzgerald: This didn't seem to be addressed?

wgpu/src/lib.rs
```diff
@@ -887,7 +887,28 @@ static_assertions::assert_impl_all!(RenderBundleDescriptor: Send, Sync);
...
+    pub sample_count: u32,
+    /// Dimensions of the texture.
+    pub dimension: TextureDimension,
+    /// Format of the texture.
+    pub format: TextureFormat,
+    /// Allowed usages of the texture. If used in other ways, the operation will panic.
+    pub usage: TextureUsages,
+    /// Specifies what view format values will be allowed when calling create_view() on this texture.
+    /// Note: currenly, only srgb-ness is allowed.
+    pub view_formats: &'a [TextureFormat],
```
cwfitzgerald, line 910: Instead of completely redefining this - can we use the generic trick like we're already using for labels? That was wgc could deal Vec (or Cow) while the surface api can deal with &[]

wgpu/tests/shader_view_format/mod.rs
```diff
@@ -0,0 +1,208 @@
+use crate::common::{image::calc_difference, initialize_test, TestParameters, TestingContext};
+use std::num::NonZeroU32;
+use wgpu::{util::DeviceExt, DownlevelFlags, Limits, TextureFormat};
+
+#[test]
+fn reinterpret_srgb_ness() {
```
cwfitzgerald, line 6: This is an amazingly thorough test <3

cwfitzgerald (changes requested):
Another round, we're very close!

wgpu/src/lib.rs
jinleili: I'm sorry I still haven't figured out how to do this. I looked at structes like `RenderBundleEncoderDescriptor` `RenderPassDescriptor` `VertexBufferLayout` etc that use the `Cow<'a [AType]>` form and they are all redefined in wgpu/lib.rs. And the `Label` generic would make passing `XXDescriptor` always require using `map_label<K>(&self, fun: impl FnOnce(&L) -> K)-> XXDescriptor<K>`, view_formats also defined as a generic would complicate the code.

wgpu-types/src/lib.rs
```diff
@@ -4336,12 +4364,14 @@ pub struct TextureDescriptor<L> {
     pub format: TextureFormat,
     /// Allowed usages of the texture. If used in other ways, the operation will panic.
     pub usage: TextureUsages,
-    // TODO: missing view_formats https://www.w3.org/TR/webgpu/#dom-gputexturedescriptor-viewformats
+    /// Specifies what view format values will be allowed when calling create_view() on this texture.
+    /// Note: currenly, only srgb-ness is allowed.
+    pub view_formats: V,
```
cwfitzgerald, line 4369:
```suggestion
    /// Specifies what view formats will be allowed when calling create_view() on this texture.
    ///
    /// View formats of the same format as the texture are always allowed.
    ///
    /// Note: currently, only the srgb-ness is allowed to change. (ex: Rgba8Unorm texture + Rgba8UnormSrgb view)
    pub view_formats: V,
```

wgpu-types/src/lib.rs
```diff
@@ -4362,14 +4393,15 @@ impl<L> TextureDescriptor<L> {
     ///
     /// ```rust
     /// # use wgpu_types as wgpu;
-    /// let desc = wgpu::TextureDescriptor {
+    /// let desc: wgpu::TextureDescriptor<(), Vec<wgpu::TextureFormat>> = wgpu::TextureDescriptor {
```
cwfitzgerald, line 4396:
```suggestion
    /// # type TextureDescriptor = wgpu::TextureDescriptor<(), &[wgpu::TextureFormat]>
    /// let desc  = TextureDescriptor {
```

To try to make the code look as similar as possible to actual user code

wgpu/src/backend/direct.rs
```diff
@@ -135,15 +135,19 @@ impl Context {
...
-            )
+        let descriptor = wgt::TextureDescriptor {
+            label: desc.label.map(Borrowed),
+            size: desc.size,
+            mip_level_count: desc.mip_level_count,
+            sample_count: desc.sample_count,
+            format: desc.format,
+            dimension: desc.dimension,
+            usage: desc.usage,
+            view_formats: desc.view_formats.to_vec(),
```
cwfitzgerald, line 146: A `map` function on TextureDescriptor would help this out a lot.

wgpu/src/util/device.rs
```diff
@@ -82,18 +82,29 @@ impl DeviceExt for crate::Device {
         desc: &crate::TextureDescriptor,
         data: &[u8],
     ) -> crate::Texture {
+        let wgt_desc: wgt::TextureDescriptor<Option<&str>, Vec<wgt::TextureFormat>> =
+            wgt::TextureDescriptor {
```
cwfitzgerald, line 86: This can just be crate::TextureDescriptor and get rid of the explicit type

cwfitzgerald (changes requested):
Really good stuff, we're almost ready to go!

cwfitzgerald (approved):

cwfitzgerald:
Thank you for all the hard work!

Issue #3162, opened 2022-11-01
wasm Testing Infrastructure 


cwfitzgerald:
We currently are only able to run our tests on native. We should implement support for running our tests on top of WebGL/WebGPU in a headless browser.

This would consist of a couple changes:

- Exposing tests to wasm-bindgen-tests
- On wasm, the test harness needs to create a canvas, resize it, and make a surface from that, before we request an adapter (as webgl requires a surface to find an adapter at all)
- Add a CI job that runs the tests using wasm-pack test using a headless chrome instance.

Our tests are designed to be run from separate processes, but running single threaded is okay, so we should be fine.

We may need to embed the test images into the binary because we don't have file IO.

haraldreingruber-dedalus:
This sounds interesting for me.
I would like to work on a first "dummy test" proof-of-concept.

I guess running something like the hello-triangle example as a wasm-bindgen test in a headless chrome via GH actions would be a good scope for a first version? 
Or did you have something other than rendering a triangle in mind?

cwfitzgerald:
`hello-triangle` doesn't actually have a corresponding test, so that wouldn't be useful for testing.

Getting https://github.com/gfx-rs/wgpu/blob/master/wgpu/tests/device.rs#L3 working would be a good first shot to make sure the `initialize_test` function is fully up to snuff to create a context in the web. Once that is going and working in the web, the other tests should be as simple as making them accessable to the wasm-bindgen harness.

haraldreingruber:
@cwfitzgerald I started a quick&dirty prototype of what I had in mind: https://github.com/gfx-rs/wgpu/pull/3238

Maybe it's easier now for exchanging more specific ideas.
It's more of an integration test, verifying the expected result of 2 pixels in the rendering result.

Could [fn init()](https://github.com/gfx-rs/wgpu/blob/242a1a77a238de54be0aaed23dccc8efc34d6270/wgpu/tests/wasm.rs#L166) be reused for the mentioned test in `device.rs`? What would you try to verify after initializing the device? Or would not panicking be sufficient?

cwfitzgerald:
I added a couple comments to the PR. 

> What would you try to verify after initializing the device?

Nothing, that test is just checking that the device is made correctly.

> Could fn init() be reused for the mentioned test in device.rs? 

[...]

PR #3238, opened 2022-11-26
Test setup for WebAssembly+WebGL

haraldreingruber:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
https://github.com/gfx-rs/wgpu/issues/3162

**Description**
This PR runs wasm-bindgen-tests in a headless-chrome (WebGL backend only).


Cargo.toml
```diff
@@ -112,6 +112,7 @@ console_log = "0.2"
 js-sys = "0.3.60"
 wasm-bindgen = "0.2.83"
 wasm-bindgen-futures = "0.4.33"
+wasm-bindgen-test = "0.3" # dev-dependency
```
cwfitzgerald, line 115: Personally wouldn't bother specifying it's a dev-dep

wgpu/tests/wasm.rs
```diff
@@ -0,0 +1,210 @@
...
+            web_sys::WebGl2RenderingContext::UNSIGNED_BYTE,
+            &mut result,
+            0,
+        )
+        .unwrap();
+    result
+}
+
+async fn init() -> (Window, EventLoop<()>, Surface, Adapter, Device, Queue) {
+    let event_loop = EventLoop::new();
```
cwfitzgerald, line 167: I wouldn't bother with winit here - all of our tests run headless, so all we need is the canvas and surface to be created, we don't actually need to interact with the canvas otherwose.

wgpu/tests/wasm.rs
```diff
@@ -0,0 +1,210 @@
...
+    console_log::init().expect("could not initialize logger");
+    // On wasm, append the canvas to the document body
+    web_sys::window()
+        .and_then(|win| win.document())
+        .and_then(|doc| doc.body())
+        .and_then(|body| {
+            body.append_child(&web_sys::Element::from(window.canvas()))
+                .ok()
+        })
+        .expect("couldn't append canvas to document body");
```
cwfitzgerald, line 180: The end goal here is to get this logic into initialize_test in tests/common. This way all our existing tests can run on wasm without modification of the test itself.

cwfitzgerald (changes requested):

wgpu/tests/clear_texture.rs
```diff
@@ -1,6 +1,9 @@
 use crate::common::{initialize_test, TestParameters, TestingContext};
+use wasm_bindgen_test::*;
 use wgpu::util::align_to;
 
+wasm_bindgen_test_configure!(run_in_browser);
```
cwfitzgerald, line 5: This only needs to happen once at tests/root.rs

wgpu/tests/common/mod.rs
```diff
@@ -187,18 +187,28 @@ pub fn initialize_test(parameters: TestParameters, test_function: impl FnOnce(Te
...
+        delete_html_canvas();
+
+        // TODO: we probably should use log crate here for logging also to wasm console
         println!("TEST SKIPPED: MISSING FEATURES {:?}", missing_features);
         return;
     }
 
     if !parameters.required_limits.check_limits(&adapter_limits) {
+        #[cfg(all(target_arch = "wasm32", feature = "webgl"))]
+        delete_html_canvas();
```
cwfitzgerald, line 200: Could we use a surface guard to do this such we don't need to remember it for every branch?

cwfitzgerald (approved):
(I know you're probably not done yet, but some small comments)

Looking great!

cwfitzgerald (changes requested):
Whoops wrong button

codeart1st:
wasm-bindgen-test with headless chrome under linux is broken for now.

https://github.com/codeart1st/wgpu-layers/actions/runs/3558060356/jobs/5976466514
https://bugs.chromium.org/p/chromium/issues/detail?id=1346838

I only get it working with a real physical GPU, but software vulkan rendering is broken in chrome with llvmpipe or swiftshader.

Maybe you have more luck to get it working.

cwfitzgerald:
This PR will only be testing webgl - so we should be okay. With how unstable webgpu is anyway, I wouldn't want to commit to testing it yet.

wgpu/tests/common/mod.rs
haraldreingruber: Yeah, that would be great. Do you mean something like the C++ RAII pattern? How would you do that?

wgpu/tests/common/mod.rs
```diff
@@ -187,18 +187,28 @@ pub fn initialize_test(parameters: TestParameters, test_function: impl FnOnce(Te
 
     let missing_features = parameters.required_features - adapter_features;
     if !missing_features.is_empty() {
+        #[cfg(all(target_arch = "wasm32", feature = "webgl"))]
+        delete_html_canvas();
+
+        // TODO: we probably should use log crate here for logging also to wasm console
         println!("TEST SKIPPED: MISSING FEATURES {:?}", missing_features);
```
haraldreingruber, line 194: @cwfitzgerald How do you feel about adding and using [console_log](https://docs.rs/console_log/latest/console_log/) as a dev-dependency here? 
It looks like it's already used in the examples.

cwfitzgerald: That's fine

wgpu/tests/common/mod.rs
cwfitzgerald: I'd have create_html_canvas return a type which impls Drop that destroys the canvas

wgpu/tests/common/mod.rs
```diff
@@ -314,6 +330,60 @@ pub fn initialize_test(parameters: TestParameters, test_function: impl FnOnce(Te
...
+    pollster::block_on(wgpu::util::initialize_adapter_from_env_or_default(
+        &instance,
+        backend_bits,
+        None,
+    ))
+    .expect("could not find suitable adapter on the system")
+}
+
+#[cfg(all(target_arch = "wasm32", feature = "webgl"))]
+fn initialize_adapter() -> Adapter {
```
cwfitzgerald, line 347: We should be able to unify these two functions by having a few conditional lines of code. initialize_adapter_from_env_or_default and friends should work on web, will just always fallback to the default.

haraldreingruber:
One thing I was not able yet to wrap my head around is, why clippy fails for the PR in the official repo, but it works (for the same commit) on my fork: https://github.com/haraldreingruber/wgpu/actions/runs/3604776801/jobs/6074498868

Here is the same job in upstream: https://github.com/gfx-rs/wgpu/actions/runs/3604776894/jobs/6074499138

Any ideas? Maybe a different cache content?

cwfitzgerald:
This is because your fork is running your code, and the PR is running against the result of merging your code into master. There are changes on master that cause the code to break. Merging in master to this branch will let you fix the problems.

.github/workflows/ci.yml
```diff
@@ -161,6 +161,33 @@ jobs:
           # build docs
           cargo doc --target ${{ matrix.target }} -p wgpu -p wgpu-core -p wgpu-info -p player --all-features --no-deps
 
+  wasm-test:
+    name: Test Wasm
```
cwfitzgerald, line 165:
```suggestion
    name: Test Webassembly
```

For consistency

.github/workflows/ci.yml
```diff
@@ -161,6 +161,33 @@ jobs:
...
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+# doesn't work for this project because of https://github.com/rustwasm/wasm-pack/issues/1180
+#      - name: install wasm-pack
+#        uses: jetli/wasm-pack-action@v0.4.0
+#
+# install from fork until this is merged: https://github.com/rustwasm/wasm-pack/pull/1185
```
cwfitzgerald, line 181:
```suggestion
					# doesn't work for this project because of https://github.com/rustwasm/wasm-pack/issues/1180
					#      - name: install wasm-pack
					#        uses: jetli/wasm-pack-action@v0.4.0
					#
					# install from fork until this is merged: https://github.com/rustwasm/wasm-pack/pull/1185
```

cwfitzgerald (commented):
Nits

wgpu/tests/common/mod.rs
haraldreingruber: It works. Thanks for suggesting the idea!

haraldreingruber:
I think this PR is close to a final review.

There are tests in 4 files, which don't work yet. I'd like to have a look if they are easily fixable.
grafik

And adding the changelog entry, if that makes sense as it's not a change that impacts wgpu users?

cwfitzgerald:
> There are tests in 4 files, which don't work yet. I'd like to have a look if they are easily fixable.

Those look like tests that all rely on panics being catchable to work (they aren't on wasm). If we replace the panics with wrapping the panicing code with `fail`, it should prevent the validation errors from turning into panics.

> And adding the changelog entry, if that makes sense as it's not a change that impacts wgpu users?

Not directly, but it's a massive indirect change that will help stability going forward!

haraldreingruber:
> Those look like tests that all rely on panics being catchable to work (they aren't on wasm). 

Yes, that's correct. At least it is the case for `buffer_copy.rs` and `texture_bounds.rs`.

`example_wgsl.rs` is searching for *.wgsl files in the examples source folders, which would need to be done completely differently for wasm. Not sure if this test is actually interesting to test in the browser... Is it mainly testing the WGSL validator? Maybe it is safe to assume that the code should work equally for wasm if the test works for non-wasm?

`clear_texture.rs` tests raise `texSubImage2D: invalid format` warnings in the Chrome dev-console. Maybe we need to improve the checks (and/or validations) to catch these errors?

> If we replace the panics with wrapping the panicing code with `fail`, it should prevent the validation errors from turning into panics.

I am not sure how to do that. I guess we only want to modify test code and don't want to change the wgpu itself?
If you could point me to some starting position I would be happy to give it a try.
But maybe we should consider to finish up this PR and improve the wasm testing in a separate one, unless it could be a simple change?

wgpu/tests/common/mod.rs
```diff
@@ -3,12 +3,17 @@
 
 use std::panic::{catch_unwind, AssertUnwindSafe};
 
+#[cfg(all(target_arch = "wasm32", feature = "webgl"))]
+use wasm_bindgen::JsCast;
```
cwfitzgerald, line 7: Let's just fully qualify these instead of using conditional imports

wgpu/tests/common/mod.rs
```diff
@@ -187,19 +188,20 @@ pub fn initialize_test(parameters: TestParameters, test_function: impl FnOnce(Te
 
     let missing_features = parameters.required_features - adapter_features;
     if !missing_features.is_empty() {
-        println!("TEST SKIPPED: MISSING FEATURES {:?}", missing_features);
+        // TODO: we probably should use log crate here for logging also to wasm console
```
cwfitzgerald, line 191: TODO is already implemented

wgpu/tests/common/mod.rs
```diff
@@ -314,6 +317,77 @@ pub fn initialize_test(parameters: TestParameters, test_function: impl FnOnce(Te
...
 
+fn initialize_adapter() -> (Adapter, SurfaceGuard) {
+    let instance;
+    let backend_bits;
+    let compatible_surface;
+
+    #[cfg(not(all(target_arch = "wasm32", feature = "webgl")))]
+    {
+        backend_bits = wgpu::util::backend_bits_from_env().unwrap_or_else(Backends::all);
+        instance = Instance::new(backend_bits);
```
cwfitzgerald, line 328: This can also be shared - webgl is the only backend on web

cwfitzgerald (changes requested):
This, plus the minor failure stuff, then we're good to go! Thanks for bringing this to completion!

cwfitzgerald:
Curious why headless firefox didn't work?

wgpu/tests/common/mod.rs
haraldreingruber: Thanks for catching this!

haraldreingruber:
> Will try again in the CI script, just in case we fixed it with the other improvements...

All tests fail with `panicked at 'could not create surface from canvas: CreateSurfaceError', wgpu/tests/common/mod.rs:331:14`

One guess would be that it's related to FF not working with software rendering in GH CI? On the other hand, I did also try running it with the "gfx.prefer-mesa-llvmpipe" config flag.
If this is not the issue, I could try to create another job with a windows runner?

cwfitzgerald:
Interesting, let's leave FF to a followup, it's not terribly important anyway.

haraldreingruber:
> Wrapping https://github.com/gfx-rs/wgpu/blob/master/wgpu/tests/buffer_copy.rs#L18 in a call to `fail_if(should_panic, || ...)` and removing the if statement before should fix the problem.

Thanks! It worked for `texture_bounds.rs`.
For `buffer_copy.rs` it still fails with `panicked at 'Error in Queue::write_buffer: copy size 17 does not respect COPY_BUFFER_ALIGNMENT', wgpu\src\backend\direct.rs:311:9`

Any idea why this workaround doesn't work in this case but does for the other?
Could it have to do with how the panics and error_scopes are implemented in the production code?

cwfitzgerald:
Yeah that means it's not getting redirected to the error scope correctly - lets just skip that one and follow up later.

cwfitzgerald (approved):
Amazing work! I'll add a changelog entry, then I'll merge it!

haraldreingruber:
Thanks for all the support!
I've created https://github.com/gfx-rs/wgpu/issues/3282 for the remaining tests. Please let me know if anything is missing in the description.

Issue #2869, opened 2022-07-10
Extended Presentation API Investigation


cwfitzgerald:
## Context

I'm working on frame pacing and we need some help from the api. The difficulty of designing this api is each WSI has different pieces of information and give it to us in different ways.

Superceeds #682 
Superceeds #2650

## Investigation

We have the following major WSIs to think about:

- IDXGISwapchain (Windows 7+ - D3D)
- IPresentationManager (Windows 11+ - D3D)
- CAMetalLayer (Mac - Metal)
- VK_GOOGLE_display_timing (Vulkan - Android)
- VK_KHR_present_wait (Vulkan - Nvidia)
- VK_KHR_incremental_present (Mainly Mesa/Android)
- VK_KHR_swapchain (All Vulkan)

And we have the following primatives:
- Get Present Start/End Time
- Wait for Present Finish
- Present with Damage
- Schedule Present Time
- Primary Monitor Frequency

<table>
<tr>
	<td>
	<td> Present Time
	<td> Wait for Present
	<td> Present with Damage
	<td> Scheduled Present
    <td> Monitor Frequency
<tr>
	<td> IDXGISwapchain
	<td> 🆗 (1a)
	<td> 🆗 (2)
	<td> ✅ (3)
	<td> 🆗 (4)
	<td> ❌ 
<tr>
	<td> IPresentationManager
	<td> ✅ (1b)
	<td> ✅ 
	<td> ❌ 
	<td> ✅ 
	<td> ❌ 
<tr>
	<td> CAMetalLayer
	<td> ✅ 
	<td> ✅ 
	<td> ❌ 
	<td> ✅ 
	<td> ✅ (5)
<tr>
	<td> VK_GOOGLE_display_timing
	<td> ✅ 
	<td> ❌ 
	<td> ❌ 
	<td> ✅ 
	<td> ✅
<tr>
	<td> VK_KHR_present_wait
	<td> ❌ 
	<td> ✅ 
	<td> ❌ 
	<td> ❌ 
	<td> ❌ 
<tr>
	<td> VK_KHR_incremental_present
	<td> ❌ 
	<td> ❌ 
	<td> ✅ 
	<td> ❌ 
	<td> ❌ 
<tr>
	<td> VK_KHR_swapchain
	<td> ❌ 
	<td> ❌ 
	<td> ❌
	<td> ❌ 
	<td> ❌ 
</table>

Notes:
1a. Presentation times need to be queried actively, it doesn't get told to us.
[...]

PR #3240, opened 2022-11-28
Implement Presentation Timestamp Correlation

cwfitzgerald:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

One step of #2869.

**Description**

Adds a way to get the appropriate presentation timestamps and correlate it with a user timestamp.

**Testing**

Manually tested using example.


wgpu-hal/src/metal/adapter.rs
```diff
@@ -319,6 +319,37 @@ impl crate::Adapter<super::Api> for super::Adapter {
...
+        #[repr(C)]
+        #[derive(Debug)]
+        struct mach_timebase_info {
+            numerator: u32,
+            denominator: u32,
+        }
+        extern "C" {
+            fn mach_timebase_info(out: *mut mach_timebase_info) -> u32;
+            fn mach_absolute_time() -> u64;
+        }
```
cwfitzgerald, line 336: Extract this into a time module.

wgpu-hal/src/metal/adapter.rs
```diff
@@ -319,6 +319,16 @@ impl crate::Adapter<super::Api> for super::Adapter {
             usage: crate::TextureUses::COLOR_TARGET | crate::TextureUses::COPY_DST, //TODO: expose more
         })
     }
+
+    unsafe fn correlate_presentation_timestamp(
+        &self,
+        user_tiemstamp_function: &mut dyn FnMut(),
+    ) -> wgt::PresentationTimestamp {
+        user_tiemstamp_function();
```
jimblandy, line 327: Ahh, the glories of autocompleiton!

```suggestion
        user_timestamp_function: &mut dyn FnMut(),
    ) -> wgt::PresentationTimestamp {
        user_timestamp_function();
```

wgpu-hal/src/lib.rs
```diff
@@ -233,6 +233,16 @@ pub trait Adapter<A: Api>: Send + Sync {
     ///
     /// `None` means presentation is not supported for it.
     unsafe fn surface_capabilities(&self, surface: &A::Surface) -> Option<SurfaceCapabilities>;
+
+    /// Creates a [PresentationTimestamp](wgt::PresentationTimestamp) then immediately calls the user_timestamp_function,
+    /// allowing the user to correlate the presentation timestamp with another set of timestamps (like `Instant`s).
+    ///
+    /// While user_timestamp_function is an FnMut, this is because of higher level internal details. It will only ever be called once
+    /// and it will unconditionally be called.
```
jimblandy, line 241:
```suggestion
    /// The `user_timestamp_function` will always be called exactly once to
    /// produce the user's timestamp, so you might expect it to be `FnOnce() ->
    /// T`. However, we want to call it by dynamic dispatch, which supports
    /// neither `FnOnce` nor generic return types. Instead, we take a `FnMut()`
    /// (no return value) which can store the user's timestamp in some captured
    /// mutable variable.
```

wgpu/src/lib.rs
```diff
@@ -2078,6 +2086,40 @@ impl Adapter {
     pub fn get_texture_format_features(&self, format: TextureFormat) -> TextureFormatFeatures {
         Context::adapter_get_texture_format_features(&*self.context, &self.id, format)
     }
+
+    /// Correleates presentation timestamps with a given user timestamp by generating a presentation timestamp
+    /// immediately after calling the user's timestamp function.
+    ///
+    /// When comparing completely opaque timestamp systems, we need a way of generating timestamps that signal
+    /// the exact same time. This function allows you to generate a timestamp (such as an [Instant]) at the
+    /// exact same time as a `PresentationTimestamp` is generated, allowing you to translate from one to another.
```
jimblandy, line 2095: This seems pretty complicated. I understand that we're trying to avoid blocking on acquiring wgpu's read locks for the adapter, but wouldn't it be simpler to just return a `PresentationTimestamp` and tell the user to call their own timestamp function immediately *after* this function returns?

> Note that this function may block waiting to acquire locks internally. To prevent this from introducing a delay between your timestamp and the WSI's timestamp, generate your timestamp immediately *after* this function returns. (Releasing locks never blocks.)

jimblandy (changes requested):
This looks correct, but it seems more complicated than necessary. See comments.

wgpu/src/lib.rs
cwfitzgerald: Did some testing and this seems viable.

The difference between the current scheme and doing it after the timestamp is 0 to 300ns in release and 300 to 1500ns in debug. This is perfectly acceptable, especially considering the precision of timestamps may be way less than that anyway

wgpu-hal/src/metal/adapter.rs
cwfitzgerald: Heh

wgpu-types/src/lib.rs
```diff
@@ -3953,6 +3953,43 @@ pub enum SurfaceStatus {
...
+///     <td>CAMetalLayer
+///     <td><a href="https://developer.apple.com/documentation/kernel/1462446-mach_absolute_time">mach_absolute_time</a>
+/// <tr>
+///     <td>VK_GOOGLE_display_timing
+///     <td><a href="https://linux.die.net/man/3/clock_gettime">clock_gettime(CLOCK_MONOTONIC)</a>
+/// </table>
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct PresentationTimestamp(
+    /// Timestamp in nanoseconds.
+    pub u128,
```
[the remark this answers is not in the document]
cwfitzgerald: I think you need to in order to robustly deal with nanosecond level ticks with a 0 point who knows when in the past.

Either way division in practice seems to be no more than a couple hundred nanoseconds. See other discussion.

CHANGELOG.md
```diff
@@ -132,6 +132,7 @@ Additionally `Surface::get_default_config` now returns an Option and returns Non
 - Implement `Clone` for `ShaderSource` and `ShaderModuleDescriptor` in `wgpu`. By @daxpedda in [#3086](https://github.com/gfx-rs/wgpu/pull/3086).
 - Add `get_default_config` for `Surface` to simplify user creation of `SurfaceConfiguration`. By @jinleili in [#3034](https://github.com/gfx-rs/wgpu/pull/3034)
 - Native adapters can now use MSAA x2 and x8 if it's supported , previously only x1 and x4 were supported . By @39ali in [3140](https://github.com/gfx-rs/wgpu/pull/3140)
+- Implemented correleation between user timestamps and platform specific presentation timestamps via [`Adapter::correlate_presentation_timestamp`]. By @cwfitzgerald in [#3240](https://github.com/gfx-rs/wgpu/pull/3240)
```
jimblandy, line 135:
```suggestion
- Implemented correleation between user timestamps and platform specific presentation timestamps via [`Adapter::get_presentation_timestamp`]. By @cwfitzgerald in [#3240](https://github.com/gfx-rs/wgpu/pull/3240)
```

jimblandy (approved):
Looks good to me! One doc fix, and one question:

PR #3241, opened 2022-11-28
V0.14

nical:
This is https://github.com/gfx-rs/wgpu/pull/3233#pullrequestreview-1194829610 backported to the v0.14 branch.

The fix applied cleanly and builds.


cwfitzgerald (approved):
Tysm!

PR #3254, opened 2022-12-02
Make `wgpu-core` users responsible for choosing back ends.

jimblandy:
Before this change, backend selection is done by target dependencies in
`wgpu-core/Cargo.toml`, giving `wgpu-core` users no way to override those
choices. (Firefox doesn't want the gles back end, for example.)

There doesn't seem to be any way to have a crate select backends based on target
architecture and OS that users of that crate can still override. The default
features can't be selected based on the target, for example. That implies that
we should do the selection as late in the dependency DAG as feasible. Having
`wgpu` (and `wgpu-core`'s other dependents) choose backends seems good enough.

Draft, because there are still some issues around the renderdoc feature. But comments are welcome!

**Checklist**

- [ ] Run `cargo clippy`.

**Connections**

**Description**

**Testing**


cwfitzgerald (approved):
Looks good! 

I have ideas to build on this on how we can expand this to the users and have non-platform specific features.

PR #3262, opened 2022-12-05
Force all adapters to report min buffer alignment of 32

Wumpf:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

-

**Description**

This follows up on the WebGPU spec which doesn't allow an api to report less than 32 bytes buffer alignment.
(brought up on Matrix https://matrix.to/#/!XFRnMvAfptAHthwBCx:matrix.org/$YbQEfpuY-nNlZWNGexna4w0pHa_ItGlm0oHlPvrC5Ds?via=matrix.org&via=mozilla.org&via=kyju.org)

**Testing**

Ran some sample applications with this


wgpu-hal/src/lib.rs
```diff
@@ -103,6 +103,10 @@ pub const MAX_COLOR_ATTACHMENTS: usize = 8;
 pub const MAX_MIP_LEVELS: u32 = 16;
 /// Size of a single occlusion/timestamp query, when copied into a buffer, in bytes.
 pub const QUERY_SIZE: wgt::BufferAddress = 8;
+/// WebGPU spec doesn't allow uniform buffer alignment to go under this value.
+pub const MIN_UNIFORM_BUFFER_OFFSET_ALIGNMENT_LOWER_BOUND: u32 = 32;
```
cwfitzgerald, line 107: This can be the same const for both. I never expect these to be different.

wgpu-hal/src/dx11/adapter.rs
```diff
@@ -213,8 +213,10 @@ impl super::Adapter {
             max_vertex_attributes: max_vertex_buffers,
             max_vertex_buffer_array_stride: u32::MAX,
             max_push_constant_size: 1 << 16,
-            min_uniform_buffer_offset_alignment: 256,
-            min_storage_buffer_offset_alignment: 1,
+            min_uniform_buffer_offset_alignment: 256
```
cwfitzgerald, line 216: Could we handle these at the core level instead of the hal level.

cwfitzgerald (changes requested):

Wumpf:
@cwfitzgerald apologies for the hiatus!
Solved it completely differently now - the modified callsite seems to be a chokepoint through which adapter limits from hal will go through, so seems like a better place to enforce this the limit

cwfitzgerald (approved):
Thanks!

cwfitzgerald:
Merge conflicts

Wumpf:
resolved

wgpu-core/src/instance.rs
```diff
@@ -176,7 +176,19 @@ pub struct Adapter<A: hal::Api> {
...
-    fn new(raw: hal::ExposedAdapter<A>) -> Self {
+    fn new(mut raw: hal::ExposedAdapter<A>) -> Self {
+        // WebGPU requires this offset alignment as lower bound on all adapters.
+        const MIN_BUFFER_OFFSET_ALIGNMENT_LOWER_BOUND: u32 = 32;
+
+        let limits = &mut raw.capabilities.limits;
+
+        limits.min_uniform_buffer_offset_alignment = limits
+            .min_uniform_buffer_offset_alignment
+            .min(MIN_BUFFER_OFFSET_ALIGNMENT_LOWER_BOUND);
```
nical, line 187: Did you mean `max`? will make it so the alignment is at most 32.

cwfitzgerald: Gah, I also got fooled by that. Petition for min/max to be "no_more_than" and "no_less_than"

Wumpf: egui does that actually because of that :D

Wumpf: excellent catch btw. Fixed now

cwfitzgerald (approved):

PR #3265, opened 2022-12-06
feat(deno): surface support

crowlKats:
WGPU has various incompatibilities with the spec with no clear solutions:

- [x] no equivalent of https://gpuweb.github.io/gpuweb/#ref-for-dom-gpucanvascontext-unconfigure
- [x] output of `surface_get_current_texture` doesnt match the equivalent web API: the wgpu function returns a struct with 2 fields, of which one is an `Option<TextureFormat>`, whereas the spec only returns a `TextureFormat`, always.


deno_webgpu/src/03_surface.js
```diff
@@ -0,0 +1,125 @@
...
+
+      if (this[_configuration] === null) {
+        throw new DOMException("context is not configured.", "InvalidStateError");
+      }
+
+      const device = assertDevice(this, { prefix, context: "this" });
+
+      const { rid } = ops.op_webgpu_surface_get_current_texture(device.rid, this[_surfaceRid]);
+
+      const texture = createGPUTexture(
```
DjDeveloperr, line 83: Shouldn’t currentTexture be *cached* until its presented?

Ref: https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-getcurrenttexture

deno_webgpu/src/03_surface.js
```diff
@@ -0,0 +1,125 @@
...
+      device.trackResource(texture);
+      return texture;
+    }
+
+    // Extended from spec. Required to present the texture; browser don't need this.
+    present() {
+      webidl.assertBranded(this, GPUCanvasContextPrototype);
+      const prefix = "Failed to execute 'present' on 'GPUCanvasContext'";
+      const device = assertDevice(this, { prefix, context: "this" });
+      ops.op_webgpu_surface_present(device.rid, this[_surfaceRid]);
```
DjDeveloperr, line 108: And after present, current texture should be destroyed and cleared (unsure if the destroy part is already done in Rust).

img

deno_webgpu/src/03_surface.js
crowlKats: Done

deno_webgpu/src/03_surface.js
crowlKats: done

cwfitzgerald (approved):
This isn't really my code to check, so as long as deno is happy with everything, consider this g2g

littledivy (approved):
LGTM

jbatez:
@crowlKats What's the use-case you have in-mind for this? Is there a current Deno window system support effort? Last week I decided to try tackling it myself (see https://github.com/denoland/deno/compare/v1.29...jbatez:deno_wsi:wsi) and didn't realize your surface extension existed since I forked off [denoland/deno@v1.29](https://github.com/denoland/deno/tree/v1.29).

My version's still very much a work-in-progress. I decided to call the WebGPU type `GPUSurface` instead of `GPUCanvasContext` because, as you noted, it's hard to get the interfaces to match exactly, and it's not clear if we'd even want to. I'm happy to use whatever you decide, though; I'm more interested in working on the window creation / configuration / event handling APIs.

Also, to any Deno devs who see this: if you have time, please take a look at my fork and let me know if it's a direction upstream Deno would be interested in pursuing. I think it's something that needs to be built into the Deno CLI itself because of how window event loops like hijack the main thread.

You can run a tiny example of my work-in-progress with `deno run --unstable --wsi example.js`:
https://github.com/jbatez/deno_wsi/blob/wsi/ext/wsi/example.js

crowlKats:
@jbatez this is a low priority side project of mine, which wouldnt be part of the CLI (we dont want a windowing manager built into the CLI). We have an internal prototype already, which does window creation, configuration & event handling.

jbatez:
@crowlKats Got it. Is your prototype based on Winit? If so, what do you do about `EventLoop::run` wanting to take over the main thread? If not, what are you doing instead?

I find the idea of an all-in-one tool for JavaScript gamedev without having to resort to native code extremely appealing. Built-in window system integration for Deno doesn't seem like much of a stretch since WebGPU's already there, but I understand if you consider it out-of-scope.

crowlKats:
Yes, we are using winit. spawning separate thread for deno.

The CLI is meant to be headless, which is why we dont have any APIs like notification or clipboard.

jbatez:
Oh, you're probably creating a whole new binary and using `deno_runtime` as a library. I was stuck wondering how to hijack the main thread of a `deno run --allow-ffi` process. Using `deno_runtime` as a library makes more sense.

The Deno CLI still has a lot of functionality I'd want in an all-in-one tool that isn't available in `deno_runtime`, but I suppose such a tool could shell out to the Deno CLI under-the-hood for most of that.

joubertnel:
> This quarter I am quite busy, but potentially next quarter I will have time to continue on this project and potentially have something to release.

This was exciting to read. Now that Deno supports WebGPU again, I'm curious to hear whether there's any update on Windowing support? It would be pretty awesome to use Deno for graphics apps!

quinton-ashley:
Is windowing support still planned for WebGPU? Could be a cool way to make game apps without Electron.

Issue #3287, opened 2022-12-11
[MacOS] shadow example not working correctly anymore on MacOS


perludum:
Here a screenshot of the shadow example on latest master:

<img width="801" alt="Bildschirm­foto 2022-12-11 um 15 41 10" src="https://user-images.githubusercontent.com/8280779/206910373-19671a71-6e72-4ea2-8e6b-19d45ba69e3d.png">

MacOS 13.0
Using AMD Radeon Pro 570 (Vulkan)

This clearly appeared with commit a50836e0cbdb9a658213c7cee5d4d290e48397ed (https://github.com/gfx-rs/wgpu/pull/3254), and with the Vulkan backend.

Before that commit, the Metal backend was selected, so I suspect it has to do with the selection of Vulkan on Mac. (Shouldn't Metal be always selected on MacOS?).

PR #3292, opened 2022-12-14
vulkan feature exclude macOS and iOS by default

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Closes https://github.com/gfx-rs/wgpu/issues/3287

**Description**
- `vulkan` feature exclude macOS and iOS by default, users can use `vulkan-portability` feature to enable vk backend.
- player can be re-runnable locally.

Run `RUSTFLAGS=--cfg=web_sys_unstable_apis cargo clippy --target wasm32-unknown-unknown` will cause error, but this is another issue:
```log
error: No back ends enabled in `wgpu-hal`. Enable at least one backend feature.
  --> wgpu-hal/src/lib.rs:59:1
   |
59 | compile_error!("No back ends enabled in `wgpu-hal`. Enable at least one backend feature.");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `wgpu-hal` due to previous error
```

**Testing**
Tested locally

### DIFF
--- a/wgpu/Cargo.toml
+++ b/wgpu/Cargo.toml
@@ -85,7 +85,7 @@ replay = ["serde", "wgc/replay"]
 angle = ["wgc/angle"]
 webgl = ["hal", "wgc"]
 emscripten = ["webgl"]
-vulkan-portability = ["wgc/vulkan-portability"]
+vulkan-portability = ["wgc/vulkan"]
 expose-ids = []
 
 # wgpu-core is always available as an optional dependency, "wgc".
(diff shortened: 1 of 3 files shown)


cwfitzgerald:
Ah, so this is because the command is explicitly trying to build wgpu-hal on wasm, which doesn't make any sense right now. We should specify to run with `-p wgpu`.

wgpu/Cargo.toml
```diff
@@ -164,7 +164,7 @@ pollster.workspace = true
 png.workspace = true
 nanorand = { workspace = true, features = ["wyrand"] }
 wasm-bindgen-test.workspace = true
-winit.workspace = true # for "halmark" example
+winit.workspace = true                                 # for "halmark" example
```
cwfitzgerald, line 167:
```suggestion
winit.workspace = true # for "halmark" example
```

cwfitzgerald (approved):

Issue #3282, opened 2022-12-10
Fix remaining tests for WebAssembly


haraldreingruber:
After adding support for running the current tests via WebAssembly in the browser (#3238), there were 2 tests that were not easily fixable for the WebAssembly target.

1. `clear_texture.rs` pass, but there are `texSubImage2D: invalid format` warnings in the browser dev-console. Which probably means that there are some false-positives.
2. `buffer_copy.rs` is executing code that is expected to panic. As catching an unwinding panic is not working for WASM, a [workaround needs to be used](https://github.com/gfx-rs/wgpu/pull/3238#issuecomment-1340290733). For this test, the workaround doesn't work because apparently the panic is not getting redirected to the error scope of the device.

In order to make tests to be executed by the WASM browser test, the `#[wasm_bindgen_test]` attribute needs to be added (next to the usual `#[test]`)

cwfitzgerald:
I think it would end up as wasm specific code which accesses the webgl2 context directly. All other apis have callback systems (including native gl https://github.com/gfx-rs/wgpu/blob/master/wgpu-hal/src/gles/egl.rs#L232) which we already hook.

cwfitzgerald:
Closing as old, see #3678 for more info about current state

PR #3296, opened 2022-12-15
Check for webgl2 validation errors and catch issues

cwfitzgerald:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Helps with https://github.com/gfx-rs/wgpu/issues/3282

**Description**

**Testing**

### DIFF
--- a/wgpu/examples/boids/main.rs
+++ b/wgpu/examples/boids/main.rs
@@ -329,7 +329,10 @@ fn main() {
     framework::run::<Example>("boids");
 }
 
+wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
+
 #[test]
+#[wasm_bindgen_test::wasm_bindgen_test]
 fn boids() {
     framework::test::<Example>(framework::FrameworkRefTest {
         image_path: "/examples/boids/screenshot.png",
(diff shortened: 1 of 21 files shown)


haraldreingruber:
Hey, nice work 🤙

I also spent a bit of time experimenting with catching the WebGL errors...
Your implementation looks much cleaner, and seems to fit better with the wgpu (test) architecture.
Still, I wanted to share my [current implementation](https://github.com/gfx-rs/wgpu/compare/master...haraldreingruber:wgpu:check-for-webgl-errors#diff-278a646ac9d51713102adcd9bca5a964ab9b4a50ff897ff3c89bb6617700d4d0R373-R382), for example, the error logging might be something you might find useful to cherry-pick? 

Just an idea...

nical (approved):

cwfitzgerald:
@haraldreingruber Normally I would be totally game for expanded error messages - but in this particular case it can be a bit misleading. Because there are potentially so many errors registered during execution, combined with how _horrid_ the GL error scheme is, adding more information when we don't really know much could leave people going down the wrong path. GL errors will already come through through the console, so we'll get some level of error messages.

haraldreingruber:
@cwfitzgerald makes totally sense. Thanks for the explanation.

PR #3310, opened 2022-12-19
Fix documentation comments for Rust 1.66.

jimblandy:
Somewhere after 1.64 but by 1.66, `rustdoc` started checking for unmatched HTML tags in doc strings, meaning that text like

    /// Convenience function turning Option<Selector> into this enum.

causes problems, since `rustdoc` will pass through `<Selector>` as HTML tag, which browsers displaying the output will misunderstand.

**Checklist**

- [X] Run `cargo clippy`.
- (unnecessary) Add change to CHANGELOG.md. See simple instructions inside file.


jimblandy:
Documentation only, non-substantive, merging without review.

PR #3316, opened 2022-12-20
Add missing `DEPTH_BIAS_CLAMP` and `FULL_DRAW_INDEX_UINT32` downlevel flags

teoxoy:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

**Description**
Added missing Vulkan/GLES downlevel flags.
Do note that we don't validate `FULL_DRAW_INDEX_UINT32` because we'd have to go through all values in the index buffers and check them.

**Testing**

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -2738,6 +2738,15 @@ impl<A: HalApi> Device<A> {
             if let Some(e) = error {
                 return Err(pipeline::CreateRenderPipelineError::DepthStencilState(e));
             }
+
+            if ds.bias.clamp != 0.0
+                && !self
+                    .shared
+                    .downlevel_flags
+                    .contains(wgt::DownlevelFlags::DEPTH_BIAS_CLAMP)
+            {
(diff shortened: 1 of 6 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -2738,6 +2738,15 @@ impl<A: HalApi> Device<A> {
             if let Some(e) = error {
                 return Err(pipeline::CreateRenderPipelineError::DepthStencilState(e));
             }
+
+            if ds.bias.clamp != 0.0
+                && !self
+                    .shared
+                    .downlevel_flags
+                    .contains(wgt::DownlevelFlags::DEPTH_BIAS_CLAMP)
```
cwfitzgerald, line 2746: This could be self.require_downlevel_flags(wgt::DownlevelFlags::DEPTH_BIAS_CLAMP) I believe

teoxoy: Right, copy-pasted this from the vulkan file - not sure why rust analyzer didn't complain.

PR #3319, opened 2022-12-20
Re-add indirect execution to vulkan downlevel flags

### LINKED ISSUE #3318: INDIRECT_EXECUTION downlevel flag not set on 2080SUPER, driver 527.56, vulkan backend
After updating today, my desktop apps no longer run because the `INDIRECT_EXECUTION` downlevel flag is not set. Latest NVidia studio driver on a 2080 SUPER using the vulkan backend, windows 10. A call to `downlevel_capabilities.is_webgpu_compliant()` returns `false`, as expected.

All other downlevel flags are set (`get_downlevel_capabilities()` returns 0x7FFFB) but the indirect bit is 0 for some reason. This adapter obviously supports indirect drawing/calls, and the flag is set if I use the DX12 backend.

Any suggestions as to what could be causing this on vulkan? Thanks...

cwfitzgerald:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Closes #3318

Indirect execution is guaranteed.

### DIFF
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -327,7 +327,8 @@ impl PhysicalDeviceFeatures {
             | Df::DEPTH_TEXTURE_AND_BUFFER_COPIES
             | Df::WEBGPU_TEXTURE_FORMAT_SUPPORT
             | Df::BUFFER_BINDINGS_NOT_16_BYTE_ALIGNED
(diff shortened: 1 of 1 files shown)


teoxoy (approved):
My bad, I double checked and this is the only one that I missed.
Thanks for adding it back!

cwfitzgerald:
Thanks for double checking!

PR #3327, opened 2022-12-22
Document and improve extension detection

teoxoy:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

**Description**
We used to require `VK_EXT_sampler_filter_minmax` and never used it.
We used to require `VK_KHR_maintenance2` and `VK_KHR_timeline_semaphore` unconditionally.
We never required `VK_EXT_image_robustness` and `VK_EXT_robustness2` (if available).
We weren't building `PhysicalDeviceDescriptorIndexingFeaturesEXT` for Vulkan 1.2+.
We weren't building `GetPhysicalDeviceProperties2` for Vulkan 1.0, this in turn making some features unusable.

**Testing**

### DIFF
--- a/wgpu-hal/src/vulkan/instance.rs
+++ b/wgpu-hal/src/vulkan/instance.rs
@@ -276,17 +276,16 @@ impl super::Instance {
             None
         };
 
-        let get_physical_device_properties = if driver_api_version >= vk::API_VERSION_1_1
-            || extensions.contains(&khr::GetPhysicalDeviceProperties2::name())
-        {
-            log::info!("Enabling device properties2");
-            Some(khr::GetPhysicalDeviceProperties2::new(
-                &entry,
-                &raw_instance,
-            ))
-        } else {
-            None
-        };
+        let get_physical_device_properties =
+            if extensions.contains(&khr::GetPhysicalDeviceProperties2::name()) {
+                log::info!("Enabling device properties2");
+                Some(khr::GetPhysicalDeviceProperties2::new(
+                    &entry,
+                    &raw_instance,
+                ))
+            } else {
+                None
+            };
 
         Ok(Self {
             shared: Arc::new(super::InstanceShared {
--- a/wgpu-hal/src/vulkan/instance.rs
+++ b/wgpu-hal/src/vulkan/instance.rs
@@ -207,9 +207,7 @@ impl super::Instance {
         // VK_KHR_get_physical_device_properties2
         // Even though the extension was promoted to Vulkan 1.1, we still require the extension
         // so that we don't have to conditionally use the functions provided by the 1.1 instance
-        // if driver_api_version < vk::API_VERSION_1_1 {
         extensions.push(vk::KhrGetPhysicalDeviceProperties2Fn::name());
-        // }
 
         // Only keep available extensions.
         extensions.retain(|&ext| {
--- a/wgpu-hal/src/vulkan/instance.rs
+++ b/wgpu-hal/src/vulkan/instance.rs
@@ -153,7 +153,7 @@ impl super::Instance {
 
     pub fn required_extensions(
         entry: &ash::Entry,
-        driver_api_version: u32,
+        _driver_api_version: u32,
         flags: crate::InstanceFlags,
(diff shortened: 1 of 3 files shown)


wgpu-hal/src/vulkan/instance.rs
```diff
@@ -172,29 +175,40 @@ impl super::Instance {
...
         }
 
-        extensions.push(vk::KhrGetPhysicalDeviceProperties2Fn::name());
-
+        // VK_EXT_swapchain_colorspace
         // Provid wide color gamut
         extensions.push(vk::ExtSwapchainColorspaceFn::name());
 
+        // VK_KHR_get_physical_device_properties2
+        if driver_api_version < vk::API_VERSION_1_1 {
```
cwfitzgerald, line 208: Because we only enable this on vulkan 1 - we end up failing to load the function pointers on vk1.1+ (at least on my machine). If we unconditionally add the extension, everything works fine.

teoxoy: So, this is not working properly?

https://github.com/gfx-rs/wgpu/blob/61d375ed0db2e6df18963537b59e478dcba0e2a4/wgpu-hal/src/vulkan/instance.rs#L279-L280

cwfitzgerald: I'm not sure what you mean - the link is to the same code I pointed at.

cwfitzgerald: Sorry read that first thing in the morning on my phone where they don't give me highlighted line numbers 😆 

It does work - the problem is that https://github.com/gfx-rs/wgpu/blob/61d375ed0db2e6df18963537b59e478dcba0e2a4/wgpu-hal/src/vulkan/instance.rs#L283-L286 only gets the proper pointer if the instance extension is enabled, which it isn't on VK1.1+

teoxoy: Ah, so even though the [`VK_KHR_get_physical_device_properties2`](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_get_physical_device_properties2.html) extension became part of core Vulkan 1.1 with no changes, the function pointers are still different between the extension and core.

I now see why we we still require `VK_KHR_draw_indirect_count` even for vk 1.2+; so that we can avoid having to load different function pointers.

Are there any downsides to this approach?

cwfitzgerald: > the function pointers are still different between the extension and core.

The function pointers you get are likely the same, the loader is just being correct and only giving you the pointer by the KHR name if you've enabled the extension. 

AFAIK there are no downsides for requiring the extension - I believe extensions are always required to be supported if the functionality is supported through core. Even if it's not required, I don't think I've ever seen it not be the case.

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -172,29 +175,42 @@ impl super::Instance {
...
-        extensions.push(vk::KhrGetPhysicalDeviceProperties2Fn::name());
-
+        // VK_EXT_swapchain_colorspace
         // Provid wide color gamut
         extensions.push(vk::ExtSwapchainColorspaceFn::name());
 
+        // VK_KHR_get_physical_device_properties2
+        // Even though the extension was promoted to Vulkan 1.1, we still require the extension
+        // so that we don't have to conditionally use the functions provided by the 1.1 instance
+        // if driver_api_version < vk::API_VERSION_1_1 {
```
cwfitzgerald, line 210: I would remove the code in the comment - it doesn't really add anything

cwfitzgerald (approved):
Works on my machine - g2g after nit

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -172,29 +175,42 @@ impl super::Instance {
...
+        extensions.push(vk::KhrGetPhysicalDeviceProperties2Fn::name());
+        // }
```
teoxoy, line 212:
```suggestion
        extensions.push(vk::KhrGetPhysicalDeviceProperties2Fn::name());
```

PR #3336, opened 2022-12-27
Remove panics in `Deref` impl of `QueueWriteBufferView` and `BufferViewMut`

### LINKED ISSUE #3134: `QueueWriteBufferView` is strange.
According to the documentation for [`Queue::write_buffer_with`](https://docs.rs/wgpu/0.14.0/wgpu/struct.Queue.html#method.write_buffer_with):

> The returned value can be dereferenced to a `&mut [u8]`; dereferencing it to a `&[u8]` panics!

According to the documentation for [`Deref`](https://doc.rust-lang.org/nightly/core/ops/trait.Deref.html):

>  **this trait should never fail**. Failure during dereferencing can be extremely confusing when `Deref` is invoked implicitly.

botahamec:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
closes #3134 

**Description**

According to the documentation for [`Queue::write_buffer_with`](https://docs.rs/wgpu/0.14.0/wgpu/struct.Queue.html#method.write_buffer_with):

> The returned value can be dereferenced to a `&mut [u8]`; dereferencing it to a `&[u8]` panics!

According to the documentation for [`Deref`](https://doc.rust-lang.org/nightly/core/ops/trait.Deref.html):

>  **this trait should never fail**. Failure during dereferencing can be extremely confusing when `Deref` is invoked implicitly.

A similar problem appears in `BufferViewMut`.

The way I solved this problem was to get rid of the implementations of `Deref` and `DerefMut` for each, implement `AsMut` for `QueueWriteBufferView`, and add the `read` method to `BufferViewMut`

**Testing**
I ran `cargo nextest run`.

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -2304,14 +2304,6 @@ pub struct BufferViewMut<'a> {
     readable: bool,
 }
 
-impl<'a> BufferViewMut<'a> {
-    /// Gives a read-only view into the buffer if this is a readable view.
-    /// Otherwise, this will return `None`.
-    pub fn read(&self) -> Option<&[u8]> {
-        self.readable.then(|| self.data.slice())
-    }
-}
-
 impl std::ops::Deref for BufferView<'_> {
     type Target = [u8];
 
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -3813,7 +3813,7 @@ impl Queue {
     ///
     /// The returned value can be converted into a `&mut [u8]` with the `as_mut` method.
     /// However, `as_ref` is not implemented for [`QueueWriteBufferView`].
-    /// (It is notunsound to read through the `&mut [u8]` anyway, but doing so will not
+    /// (It is not unsound to read through the `&mut [u8]` anyway, but doing so will not
     /// yield the existing contents of `buffer` from the GPU, and it is likely to be slow.)
     ///
     /// This method is intended to have low performance costs.
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -20,7 +20,7 @@ use std::{
     future::Future,
     marker::PhantomData,
     num::{NonZeroU32, NonZeroU8},
-    ops::{Bound, Range, RangeBounds},
+    ops::{Bound, Deref, DerefMut, Range, RangeBounds},
     sync::Arc,
     thread,
 };
@@ -2335,6 +2335,20 @@ impl AsMut<[u8]> for BufferViewMut<'_> {
     }
 }
 
+impl Deref for BufferViewMut<'_> {
+    type Target = [u8];
+
+    fn deref(&self) -> &Self::Target {
+        self.data.slice()
+    }
+}
+
+impl DerefMut for BufferViewMut<'_> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.data.slice_mut()
+    }
+}
+
 impl Drop for BufferView<'_> {
     fn drop(&mut self) {
         self.slice
@@ -3767,6 +3781,20 @@ pub struct QueueWriteBufferView<'a> {
     inner: Box<dyn context::QueueWriteBuffer>,
 }
 static_assertions::assert_impl_all!(QueueWriteBufferView: Send, Sync);
+
+impl Deref for QueueWriteBufferView<'_> {
+    type Target = [u8];
+
+    fn deref(&self) -> &Self::Target {
+        self.inner.slice()
+    }
+}
+
+impl DerefMut for QueueWriteBufferView<'_> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.inner.slice_mut()
+    }
+}
 
 impl<'a> std::convert::AsMut<[u8]> for QueueWriteBufferView<'a> {
     fn as_mut(&mut self) -> &mut [u8] {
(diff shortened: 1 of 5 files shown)


nical:
Sorry I'm chiming in a bit late, didn't see https://github.com/gfx-rs/wgpu/issues/3134.

Considering we can read from the `&mut` deref, I think that we should just keep things simple, remove the panic and simply make sure to document the behavior well (reading the bytes may be slow and won't yield the real content of the buffer).
I'm not fond of the panic in the first place since it solves at best half of the problem it tries to solve.

Thoughts @teoxoy @cwfitzgerald ?

teoxoy:
> Considering we can read from the `&mut` deref, I think that we should just keep things simple, remove the panic and simply make sure to document the behavior well (reading the bytes may be slow and won't yield the real content of the buffer).

I'd say that this is quite difficult to document since `Deref`/`DerefMut` happen implicitly.
This PR looks good to me, makes things more idiomatic.

@nical are you concerned about the ergonomics being slightly worse?

cwfitzgerald:
I am having trouble coming to a conclusion of which option I actually like more. Deref* makes it really easy to use - lets it slot into various apis that take anything that impls Deref (notably Encase). And it's not like we don't have other giant performance footguns that you can easily write. At the end of the day, as long as the panic goes away (as that is notably _unidomatic_. I think we'll be fine.

nical:
In my opinion it's irrelevant whether someone ends up calling Deref or DerefMut (explicitly or implicitly), what matters is that they don't read form the (& or &mut) reference. AsMut has the same problem, it doesn't prevent or discourage reading, it just forces you to type "as_mut()" before doing so. As a result we still need to document that reading has likely undesired effects (potentially slow uncached memory reads and not providing access to the real buffer data).

Deref could be useful, here. We should not panic because it is not idiomatic, but we can log a performance warning. If we remove that hook, it will be as easy for accidental reads to happen but we won'be able to put a red flag on some of them. 

> @nical are you concerned about the ergonomics being slightly worse?

Yes-ish. My primary concern is that I don't think the way we are trying to paper over the issue helps (I don't think Rust can let us express something that would be helpful for this). The ergonomics cost is not something I'll lose sleep over but at the same time since I'm unconvinced that it helps in a meaningful way, my preference would be to keep things simple.


The problem isn't important enough to press further so if my arguments have not swayed you, don't feel bad about merging this as is.

teoxoy:
I'm in the same boat as @cwfitzgerald (don't think there is a satisfactory solution).

> We should not panic because it is not idiomatic, but we can log a performance warning.

This sounds better to me if we keep the `Deref`/`DerefMut`; log a warning instead of panicking.

@botahamec how do you feel about keeping the current API and logging a warning instead of panicking?

botahamec:
@teoxoy That works for me. I assume that's referring to `QueueWriteBufferView`. Should `BufferViewMut` stay the way I currently have it in this PR?

nical:
I'd say let them both have the familiar `Deref`/`DerefMut` API.

botahamec:
@nical Does that make sense for `BufferViewMut`? Currently `Deref` panics only if `self.readable` is false. Is it similar to `QueueWriteBufferView` where it will just give nonsense data if `self.readable` is false?

cwfitzgerald:
I _believe_ it will give something resembling the correct data, just may be horrendously slow to read from.

nical:
Yep, like @cwfitzgerald said. It makes sense for BufferViewMut` for the same reason it makes sense for `QueueWriteBufferView` (For some definition of "makes sense" since it's hard to express what we really want to express in rust).

botahamec:
I added back `Deref` and `DerefMut`. Let me know if anyone has any other comments

botahamec:
Oh I guess `readable` isn't being used anymore on `BufferViewMut`. Should I remove it?

wgpu/src/lib.rs
```diff
@@ -3777,21 +3777,26 @@ pub struct QueueWriteBufferView<'a> {
...
 
-impl<'a> std::ops::DerefMut for QueueWriteBufferView<'a> {
-    #[inline]
+impl DerefMut for QueueWriteBufferView<'_> {
     fn deref_mut(&mut self) -> &mut Self::Target {
         self.inner.slice_mut()
     }
 }
 
+impl<'a> std::convert::AsMut<[u8]> for QueueWriteBufferView<'a> {
```
cwfitzgerald, line 3794: Nit:
```suggestion
impl<'a> AsMut<[u8]> for QueueWriteBufferView<'a> {
```

cwfitzgerald (approved):
One nit, then g2g

teoxoy:
What about logging out warnings? We don't seem to do that yet but was what we previously talked about.

botahamec:
@teoxoy Oh yes! You're right, I completely forgot. I'll get to that.

wgpu/src/lib.rs
botahamec: Should be good now

teoxoy (approved):
LGTM, thanks!

teoxoy:
@botahamec looks like CI is unhappy

botahamec:
Hmm, it's weird that my editor didn't show any problems. Maybe that's a sign to put less faith in rust-analyzer. Hopefully it works now.

botahamec:
Ok, coincidentally, `cargo deny` generated an error because of a recent vulnerability in tokio #3368

Issue #2710, opened 2022-06-02
Remove Locking From Hot Paths


cwfitzgerald:
**Is your feature request related to a problem? Please describe.**

This formalizes #2272 into a issue.

Our current system takes global locks per resource. This means that if one method needs mutable access to a particular type, it needs to lock _all_ resources of that type as mutable. Because every method needs read or write access to some resource, it's very easy to write multithreaded code that is essentially serialized.

The worst case of this is renderpass encoding. In an ideal world, this would be fully parallelizable, but currently, because of us needing a write lock, is serial.

This is one of the ways that wgpu has major disadvantages over the underlying api.

**Describe the solution you'd like**

We should switch to a wait free datastructure on the hot path with per-object locking or more fine grained locking. 

**Describe alternatives you've considered**
The arcinization was another alternative, but we have shifted away from that style;

Imberflur:
related issues https://github.com/gfx-rs/wgpu/issues/1485, https://github.com/gfx-rs/wgpu/issues/2023, https://github.com/gfx-rs/wgpu/issues/2394, https://github.com/gfx-rs/wgpu/issues/2395

John-Nagle:
That identifies the problem better. See my test case for this here: https://github.com/John-Nagle/render-bench

John-Nagle:
It's an ongoing problem. The overall effect is that WGPU is S-L-O-W for anything which loads its content while rendering. 
A major claimed advantage of Vulkan over OpenGL is that you can have one CPU loading content into the GPU while another manages the rendering. WGPU loses that advantage.

John-Nagle:
Discussion on Reddit: https://www.reddit.com/r/rust_gamedev/comments/11b0brr/were_not_really_game_yet/

If someone needs **render-bench** updated to use later versions of Rend3 and WGPU, please let me know. That's a good test for this problem.

gents83:
@cwfitzgerald if someone would like to take this is it suggested to start from your work or to restart from scratch (you mentioned that you had new ideas, maybe it could be worth to write down the intentions)?
I could try to take this, but I didn't want to do something in a way that is not what you thought 😁

PS: this is how I tackle the same issue in my engine, with Arc<RwLock<T>> on single Resource and keeping a simple Storage.
- Resource: https://github.com/gents83/INOX/blob/master/crates/resources/src/resource.rs
- Storage:
https://github.com/gents83/INOX/blob/master/crates/resources/src/storage.rs

[...]

PR #3346, opened 2023-01-03
added texture metadata

kurtkuehnert:
**Description**

This PR exposes the metadata/attributes of a `Texture`, according to the WebGpu spec: https://www.w3.org/TR/webgpu/#gputexture.

This is a draft because I am unsure how to set those attributes for the `SurfaceTexture`s (see the Todo).
I would appreciate any guidance you can give me.

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -228,14 +228,7 @@ pub struct Texture {
     id: ObjectId,
     data: Box<Data>,
     owned: bool,
-    width: u32,
-    height: u32,
-    depth_or_array_layers: u32,
-    mip_level_count: u32,
-    sample_count: u32,
-    dimension: TextureDimension,
-    format: TextureFormat,
-    usage: TextureUsages,
+    descriptor: TextureDescriptor<'static>,
 }
 static_assertions::assert_impl_all!(Texture: Send, Sync);
 
@@ -291,6 +284,7 @@ pub struct Surface {
     context: Arc<C>,
     id: ObjectId,
     data: Box<Data>,
+    // config: SurfaceConfiguration,
 }
 static_assertions::assert_impl_all!(Surface: Send, Sync);
 
@@ -2047,14 +2041,10 @@ impl Device {
             id,
             data,
             owned: true,
-            width: desc.size.width,
-            height: desc.size.height,
-            depth_or_array_layers: desc.size.depth_or_array_layers,
-            mip_level_count: desc.mip_level_count,
-            sample_count: desc.sample_count,
-            dimension: desc.dimension,
-            format: desc.format,
-            usage: desc.usage,
+            descriptor: TextureDescriptor {
+                label: None,
+                ..desc.clone()
+            },
         }
     }
 
@@ -2087,14 +2077,10 @@ impl Device {
             id: ObjectId::from(texture.id()),
             data: Box::new(texture),
             owned: true,
-            width: desc.size.width,
-            height: desc.size.height,
-            depth_or_array_layers: desc.size.depth_or_array_layers,
-            mip_level_count: desc.mip_level_count,
-            sample_count: desc.sample_count,
-            dimension: desc.dimension,
-            format: desc.format,
-            usage: desc.usage,
+            descriptor: TextureDescriptor {
+                label: None,
+                ..desc.clone()
+            },
         }
     }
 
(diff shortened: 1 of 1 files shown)


wgpu/src/lib.rs
```diff
@@ -227,6 +228,14 @@ pub struct Texture {
...
     data: Box<Data>,
     owned: bool,
+    width: u32,
+    height: u32,
+    depth_or_array_layers: u32,
+    mip_level_count: u32,
+    sample_count: u32,
+    dimension: TextureDimension,
+    format: TextureFormat,
+    usage: TextureUsages,
```
cwfitzgerald, line 238: Should we just store the descriptor?

wgpu/src/lib.rs
```diff
@@ -2522,6 +2547,62 @@ impl Texture {
...
+    pub fn height(&self) -> u32 {
+        self.height
+    }
+
+    /// Returns the depth or layer count of this `Texture`.
+    ///
+    /// This is always equal to the `size.depth_or_array_layers` that was specified when creating the texture.
+    pub fn depth_or_array_layers(&self) -> u32 {
+        self.depth_or_array_layers
+    }
```
cwfitzgerald, line 2570: Could we also add a "size" function that returns an Extent3D

wgpu/src/lib.rs
```diff
@@ -3990,6 +4071,18 @@ impl Surface {
                     id,
                     data,
                     owned: false,
+
+                    // Todo: how do we specify these values correctly?
```
cwfitzgerald, line 4075: We should store the surface configuration in the surface.

kurtkuehnert: Yes, I think so too, but unfortunately, the configure method (of the surface) has an immutable reference to self. We could use interior mutability but this feels like a hack. Do you have any advice on how to resolve this?

wgpu/src/lib.rs
kurtkuehnert: Sure thing. 👍

wgpu/src/lib.rs
kurtkuehnert: This is possible, but the lifetime of the label might be an issue. I can just set it to None.

wgpu/src/lib.rs
cwfitzgerald: The way we're doing it now is entirely a hack to begin with - we should be actually reaching into the underlying resources - but because we can't do that efficiently currently (see #2710) we're doing this at the surface level. I would just throw it in a mutex.

kurtkuehnert: Ok, this should work.

wgpu/src/lib.rs
```diff
@@ -3982,6 +4066,25 @@ impl Surface {
             SurfaceStatus::Lost => return Err(SurfaceError::Lost),
         };
 
+        let config = self
+            .config
+            .lock()
+            .expect("This surface has not been configured yet.");
```
kurtkuehnert, line 4072: What should I do in case the `SurfaceConfiguration` is not available? Does a panic suffice?

wgpu/src/lib.rs
```diff
@@ -1563,6 +1571,8 @@ impl Instance {
             id: ObjectId::from(surface),
             #[cfg(all(target_arch = "wasm32", not(feature = "webgl")))]
             data: Box::new(()),
+            #[cfg(all(target_arch = "wasm32", not(feature = "webgl")))]
```
cwfitzgerald, line 1574: This should be unconditional

wgpu/src/lib.rs
```diff
@@ -282,6 +284,7 @@ pub struct Surface {
     context: Arc<C>,
     id: ObjectId,
     data: Box<Data>,
+    config: Mutex<Option<SurfaceConfiguration>>,
```
cwfitzgerald, line 287: Could you add a note about why we need this?

cwfitzgerald (changes requested):
G2G after some small things

wgpu/src/lib.rs
kurtkuehnert: Sure. 👍

wgpu/src/lib.rs
kurtkuehnert: I changed this on lines 1543 and 1579.

cwfitzgerald (approved):
Thanks!

Issue #3349, opened 2023-01-04
GLES 3.0 Binding Model Broken with Shared Vertex/Fragment Uniform Buffers


cwfitzgerald:
**Description**

Due to GLES 3.0 not having binding indexes, we need to rely completely on reflection to bind all the buffers in the right place. Naga (correctly) generates unique names and structures and such for the vertex and fragment shaders. Because there are no binding indexes, GLES matches up bindings by name of the struct, and because they are different, creates different bindings for each stage.

Our choices are to:
- Change our binding model to bind buffers separately for vertex and fragment on GLES 3.0.
- Try to share the binding across states. This unfortunately also means that the structs in vertex and fragment must be _identical_, and this is not a requirement of WebGPU at all.

In order to not add this giant requirement, we should change our binding model.

**Repro steps**

Bind a buffer to both vertex and fragment shaders in GLES 3.0.

**Expected vs observed behavior**

Only one of the buffers will be correctly bound, the other won't be. This will cause all sorts of nonsense.

ryankaplan:
Hi again @cwfitzgerald! Can you help me figure out if I'm running into this issue? The repro steps are `Bind a buffer to both vertex and fragment shaders in GLES 3.0` but I'm not sure what this means from the perspective of the wgpu API. Is it sufficient to draw using a uniform buffer that is bound to uniforms with both `wgpu::ShaderStages::VERTEX` and `wgpu::ShaderStages::FRAGMENT` visibility?

cwfitzgerald:
Yeah - binding a single buffer with VERTEX_FRAGMENT doesn't work. It will only end up binding to one stage.

ryankaplan:
So I'm extra sure, you're saying that this is no good...

```
wgpu::BindGroupLayoutEntry {
  // Problem: we will bind to only one of these...
  visibility: wgpu::ShaderStages::VERTEX | wgpu::ShaderStages::FRAGMENT,
  ...
}
```

But I could, for example, split the above layout entry into two layout entries that bind to the same buffer that I bound to before. Does that sound right?

cwfitzgerald:
Correct!

cwfitzgerald:
This bug doesn't exist.

I do not have the original reproduction for this issue and the test I wrote for it passes every time.

Notably we use glUniformBlockBinding to bind the blocks on the CPU side, which allows us to call bind once to multiple blocks, and everything is happy.

[...]

PR #3352, opened 2023-01-05
Update naga to  e98bd92

nical:
**Description**

A routine naga update update.
I don't think there is anything to add in the changelog since naga has its own.


cwfitzgerald (approved):

JMS55 (approved):

teoxoy:
This test now passes
https://github.com/gfx-rs/wgpu/blob/a06ef71fd773914f7bd68773054a1a2222d70f98/wgpu/examples/shadow/main.rs#L856
not sure why though (new version of angle?)

cc @cwfitzgerald (since it was added quite recently in https://github.com/gfx-rs/wgpu/pull/3296)

cwfitzgerald:
I don't know why either - it was failing due to validation errors stemming from #3349 and it's not like that has been fixed.

teoxoy:
What should we do for now to get this merged? I reran the workflow and it still succeeds when it shouldn't.
Comment out the failure and go from there?

cwfitzgerald (approved):
Alright, this is fine to keep things moving, it doesn't make any sense though - will look into it tomorrow.

Issue #3145, opened 2022-10-27
RFC: Introduce `ExternalTexture`


i509VCB:
Related to the discussion around a texture [memory import api](https://github.com/gfx-rs/wgpu/issues/2320), it may be desirable to import memory from external sources, such as video.

The current state of externally created textures in wgpu restricts the types of textures that can be imported. In particular, textures right now must meet the following requirements:
1. Be RGB or some permutation of this
2. Must be a format known by wgpu. The graphics apis and platforms that wgpu using may support more formats that are common in for example video.
3. Not multi-planar. Multi-planar formats are common in video, such as decoding from a video codec or the memory objects from something like vaapi.

The solution to this would be to introduce an `ExternalTexture` type which represents an opaque, read-only, 2d, implementation dependent view over some texture memory that can be copied and or sampled from.

This type would be analogous to the existing `GPUExternalTexture` from webgpu. In order to keep the maintenance burden down, creation of `ExternalTexture`s on native from platform memory handles is considered to be out of scope of wgpu's public api (except for an `external_texture_from_hal` function). Hal implementations inside wgpu-hal will be responsible for creating any `ExternalTexture` instances and should expose mechanisms to allow this. The web backend will provide a function to create an `ExternalTexture` from a `GPUExternalTexture`.

Along with the new type, `ExternalTexture` like `TextureView` would become a valid resource to bind in a bind group.

Naga would need to grow infrastructure to support the `texture_external` sampler.

## Implementation

One thing that would differ between webgpu and the native backends would be how external textures are managed. On webgpu, external textures are only valid for the lifetime of the current frame. Essentially you need to reimport the texture every time you render. On native, the external texture is valid for the lifetime of the instance of the external texture. One solution to keep uniform behavior would be to have wgpu on web reimport the GPUExternalTexture every time it is used. This would likely involve owning a copy of a `HTMLVideoElement` object used the import the texture.

For native, any format which Vulkan supports with our current infrastructure should work. YUV and multi-planar formats may require internal support for VK_KHR_ycbcr_sampler_conversion. On Android, any AHardwareBufferFormat should work with Vulkan, as the extensions which allow importing an AHB in Vulkan allow using formats not known by Vulkan, but are still valid AHB formats.

EGL I am not so familiar with, but there could be some integration with the `TEXTURE_EXTERNAL` texture type.

I am not familiar with how DX12 and Metal would work here. I would like some advice on how DX12 and Metal should be integrated into this.

[...]

PR #3019, opened 2022-09-12
tracking for EXTERNAL texture use

i509VCB:
**Checklist**

- [X] Run `cargo clippy`.
- [X] Run `RUSTFLAGS=--cfg=web_sys_unstable_apis cargo clippy --target wasm32-unknown-unknown` if applicable.
- [ ] Add change to CHANGELOG.md. See simple instructions inside file.

**Connections**
Resolves #2948

**Description**
This pull request adds tracking for a new `EXTERNAL` texture use. In the Vulkan hal, a new parameter for `texture_from_raw` to specify the owning foreign queue of the texture is added and logic changes to transition external textures to and from the external queue.

**Testing**
An example needs to be created to use an image that would have an EXTERNAL texture use.

I will write an example to test this. It will probably be an example using kernel mode setting (KMS).

Regarding testing on CI, I am not certain how we could approach that.

PR #3355, opened 2023-01-06
Add support for importing external buffers

AdrianEddy:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Related to https://github.com/gfx-rs/wgpu/issues/3145 and https://github.com/gfx-rs/wgpu/issues/2320

**Description**
This PR introduces `create_buffer_from_hal` in addition to the existing `create_texture_from_hal`, and also adds `buffer_from_raw` to Dx12 and Metal backends.

These new functions are mostly copy-paste from `create_texture_from_hal`, with minor adjustments for buffers instead of textures.

I also wanted to add `buffer_from_raw` to Vulkan, but I'm not sure what should be used in `block` variable ( `Mutex<gpu_alloc::MemoryBlock<vk::DeviceMemory>>`). How does the MemoryBlock work here if we provide external memory? Should we also get `vk::DeviceMemory` in `buffer_from_raw` argument and somehow fit that to the MemoryBlock struct?

**Testing**
I tested the external Dx12 buffer with rendering over normal `wgpu::Texture` and then using _encoder.copy_texture_to_buffer_ and _copy_buffer_to_texture_ and rendering that buffer in an external application. Everything worked as expected.


cwfitzgerald:
> I also wanted to add buffer_from_raw to Vulkan, but I'm not sure what should be used in block variable ( Mutex<gpu_alloc::MemoryBlock<vk::DeviceMemory>>). How does the MemoryBlock work here if we provide external memory? Should we also get vk::DeviceMemory in buffer_from_raw argument and somehow fit that to the MemoryBlock struct?

I think this just needs to be optional as it isn't our responsibility to free it or manage its memory at all.

This PR generally looks fine - but there doesn't seem to be a reasonable way of actually _using_ the buffer you import. Specifically the first time it's used in wgpu-land we'll issue a transition from UNDEFINED and the contents of the buffer are now moot. Additionally there's no way to specify barriers from transitioning control to and from wgpu-land.

AdrianEddy:
> Specifically the first time it's used in wgpu-land we'll issue a transition from UNDEFINED and the contents of the buffer are now moot.

right, so we should assume that externally imported buffers are already initialized and ready to use. I now changed that by using `BufferInitTracker::new(0)` which _I believe_ means "There's 0 bytes left to initialize"

I also added `buffer_from_raw` to the Vulkan backend and tested it. I made `block` optional, and this works, but it makes these buffers unable to be mapped by `wgpu`. This may be fine for some use cases, but may not be for other cases. We could made it work by adding `Option<vk::DeviceMemory>` as a second argument, but I was unable to figure out how to fit that with `gpu_alloc` and it's private `MemoryBlock::new` function. This is fine for my case, as I'm only interested in GPU-only interoperability, and if I wanted GPU-CPU interop, I would just use normal `wgpu.create_buffer`

I tested it using Metal, Dx12 and Vulkan backends and it all worked as expected.


> Additionally there's no way to specify barriers from transitioning control to and from wgpu-land.

In which cases this is needed? Do we need to have a way to specify barriers for this feature to be merged?

wgpu-hal/src/vulkan/device.rs
```diff
@@ -835,65 +846,75 @@ impl crate::Device<super::Api> for super::Device {
...
+        if let Some(block) = &buffer.block {
+            let size = range.end - range.start;
+            let mut block = block.lock();
+            let ptr = unsafe { block.map(&*self.shared, range.start, size as usize)? };
+            let is_coherent = block
+                .props()
+                .contains(gpu_alloc::MemoryPropertyFlags::HOST_COHERENT);
+            Ok(crate::BufferMapping { ptr, is_coherent })
+        } else {
+            Err(crate::DeviceError::OutOfMemory)
```
AdrianEddy, line 877: I guess this is not the best error for this case. I think the best one would be `BufferAccessError::Invalid`, but that is in `wgpu_core` crate. Should I add `DeviceError::InvalidBufferAccess` to the `crate::DeviceError` enum?

cwfitzgerald: I think we need to add validation at the wgpu-core level that makes sure that imported buffers don't have the MAP* usage. Then this can be unreachable

cwfitzgerald:
Have you run sync validation against your example? Unless I'm missing some component of this, which is possible, we should end up issuing a weirdly meaningless barrier. I think in practice it might work, but it is likely invalid. 

After https://github.com/gfx-rs/wgpu/pull/3355/files#diff-90e54105789cf7268e0acbfd914035fccb5967895c81d21ac60f94af6b439edaR4020 the buffer is stored with no state in the tracker. This is how new buffers are made in wgpu as well, so that's fine. But on first usage, we'll issue a barrier for {empty}..UNIFORM or whatever and I'm not positive that won't invalidate the contents of the buffer (or is generally invalid, as the previous usage was something else, not empty).  See https://github.com/gfx-rs/wgpu/blob/master/wgpu-hal/src/vulkan/conv.rs#L516-L567 for how we map our usages to barrier usage

AdrianEddy:
I have to dig deeper into the tracking and barrier stuff, I don't have the full picture of that process and something weird might be going on here. 

It works correctly for my case with Metal/Vulkan/Dx12, but on Metal and Dx12 I'm getting much worse performance with buffers than textures (same data, buffer ~20ms/frame vs texture ~10ms/frame). Vulkan is fast (3ms) with both buffers and textures. I'm not sure if it has to do with these barriers and tracking, but it might, so I need to investigate that. 

If you have any ideas or hints please let me know

i509VCB:
My comment does not need to be addressed in this pull request, I believe it would also probably be useful to replicate something like https://github.com/gfx-rs/wgpu/pull/3019 for external buffers. In particular on Linux a `VkBuffer` can be shared between processes using external memory handles.

cwfitzgerald (approved):
Looks good, thanks for the patience!

PR #3356, opened 2023-01-06
Updated Dxc integration for DX12 backend

### LINKED ISSUE #2722: DXC Integration in DX12 Backend

Elabajaba:
**Checklist**

- [x] Run `cargo clippy`.
- [x] Figure out how we should enable and disable DXC, and whether that's at runtime/compiletime/startup/etc
- [x] Figure out what the fallback should be if the user enables DXC, but the .dlls aren't available (~~crash with an error message?~~ Fall back to fxc but log the error)
- [x] Decide if we're going to allow users to set an explicit path to the .dll location
- [x] ~~Figure out why it's working on my machine if I only have dxil.dll, and not dxcompiler.dll in the folder~~ I have the Windows SDK installed
- [x] Documentation about why users would/should use DXC, and how to get the .dlls
- [x] ~~Fix all the tests in github actions being broken because the .dlls are missing~~ Currently defaults to using FXC.

**Connections**
Closes #2722 
Builds off of, and closes #3147 

**Description**
wgpu currently uses the old, slow, unmaintained FXC shader compiler for DX12. This allows use of Microsoft's newer faster shader compiler and also allows use of ShaderModel 6.0 instead of being limited to SM 5.1.

**Testing**
Tested both the FXC and DXC paths by running `cargo nextest`, as well as running the examples.


cwfitzgerald (commented):
Some comments:

> Figure out why it's working on my machine if I only have dxil.dll, and not dxcompiler.dll in the folder

You probably have the windows sdk, which includes dxc. 

> Figure out how we should enable and disable DXC

I think we should have a feature at the wgpu-hal level, but _not_ at the wgpu level. `wgpu` the crate unconditionally depends on it.

> Figure out what the fallback should be if the user enables DXC, but the .dlls aren't available (crash with an error message? Fall back to dxc but log the error?)

We should warn in the case the user wants dxc, but the path isn't found, then fall back to FXC.

> Figure out how we should enable and disable DXC, and whether that's at runtime/compiletime/startup/etc

I'm thinking of this enum sitting in an InstanceDescriptor:

```rust
enum Dx12Compiler {
    #[default]
    Fxc,
    Dxc { dxil_path: Option<String>, dxc_path: Option<String> },
}
```

> Fix all the tests in github actions being broken because the .dlls are missing

We likely should default our tests to be fxc only, but with a flag they can enable DXC support. FF etc will be using fxc, so it's the most important path.

wgpu-hal/src/dx12/adapter.rs
```diff
@@ -243,6 +244,7 @@ impl super::Adapter {
                 private_caps,
                 presentation_timer,
                 workarounds,
+                dx12_shader_compiler: dx12_shader_compiler.clone(),
```
Elabajaba, line 247: I just cloned this down the initialization pipeline from Instance to Device as it was simple, and I wasn't sure if Instance lasted longer than Device.

wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -0,0 +1,260 @@
...
+            Err(e) => return (Err(e), log::Level::Error),
+        };
+
+        let compiled = dxc_container.compiler.compile(
+            &blob,
+            source_name,
+            raw_ep,
+            &full_stage,
+            &compile_flags,
+            None, // TODO: idk what the include handler does
```
Elabajaba, line 163: I still don't know what the include handler does. @MarijnS95, could you clarify what the include handler does when compiling with dxc?

wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -0,0 +1,260 @@
+use std::ptr;
```
Elabajaba, line 1: I split it up in a similar way to the gpu_allocator pr, to avoid a mess of cfgs in device.rs.

wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -0,0 +1,260 @@
...
+        _dxc_path: Option<PathBuf>,
+        _dxil_path: Option<PathBuf>,
+    ) -> Result<Option<DxcContainer>, crate::DeviceError> {
+        // Falls back to Fxc and logs an error.
+        log::error!("DXC shader compiler was requested on Instance creation, but the DXC feature is disabled. Enable the `dxc_shader_compiler` feature on wgpu_hal to use DXC.");
+        Ok(None)
+    }
+
+    // It shouldn't be possible that this gets called with the `dxc_shader_compiler` feature disabled, but just in case...
+    pub(crate) fn compile_dxc(
```
Elabajaba, line 238: This only ever gets called when the DxcContainer field on `Device` is `Some<DxcContainer>`, but it will/should always be `None` if the dxc_shader_compiler feature is disabled. Not sure if it should be `unimplemented!()`, or leave it stubbed in as a safety net.

wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -0,0 +1,260 @@
...
+        // Has to be held onto for the lifetime of the device otherwise shaders will fail to compile
+        _dxc: hassle_rs::Dxc,
+    }
+
+    pub(crate) fn get_dxc_container(
+        dxc_path: Option<PathBuf>,
+        dxil_path: Option<PathBuf>,
+    ) -> Result<Option<DxcContainer>, crate::DeviceError> {
+        // Make sure that dxil.dll exists.
+        match hassle_rs::Dxil::new(dxil_path) {
```
Elabajaba, line 97: `dxil.dll` never gets used directly, it just gets sneakily called by `dxcompiler.dll` to validate the shaders during compilation so I didn't bother storing it.

wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -0,0 +1,260 @@
...
+        let blob = match dxc_container
+            .library
+            .create_blob_with_encoding_from_str(source)
+            .map_err(|e| crate::PipelineError::Linkage(stage_bit, format!("DXC blob error: {}", e)))
+        {
+            Ok(blob) => blob,
+            Err(e) => return (Err(e), log::Level::Error),
+        };
+
+        let compiled = dxc_container.compiler.compile(
```
Elabajaba, line 157: TODO: add a comment that points out that this also validates the shaders.

wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -0,0 +1,260 @@
...
+                    // TODO: This returns an HRESULT, should we try and use the associated Windows error message?
+                    log::error!("Win32 error: {e:?}");
+                    crate::DeviceError::Lost
+                }
+                hassle_rs::HassleError::LoadLibraryError { filename, inner } => {
+                    log::error!("Failed to load dxc library {filename:?}. Inner error: {inner:?}");
+                    crate::DeviceError::Lost
+                }
+                hassle_rs::HassleError::LibLoadingError(e) => {
+                    log::error!("Failed to load dxc library. {e:?}");
```
Elabajaba, line 206: My additional text on these errors is probably redundant, as I'm also displaying the hassle error and whatever their text for it is.

wgpu/src/util/init.rs
```diff
@@ -86,3 +86,21 @@ pub async fn initialize_adapter_from_env_or_default(
         }
     }
 }
+
+/// Choose which DX12 shader compiler to use from the environment variable WGPU_DX12_COMPILER
+pub fn dx12_shader_compiler_from_env() -> Option<wgt::Dx12Compiler> {
```
Elabajaba, line 91: This is nice for testing, and could probably be used in the future for running tests on FXC and DXC. I didn't bother trying to figure out how to parse paths from the environment variable so it only checks locally. (or wherever `libloading` checks, I'm not 100% sure how it works).

CHANGELOG.md
```diff
@@ -107,10 +107,15 @@ Additionally `Surface::get_default_config` now returns an Option and returns Non
...
 
+#### DXC Shader Compiler Support for DX12
+
+You can now choose to use the DXC compiler for DX12 instead of FXC. The DXC compiler is faster, less buggy, and allows for new features compared to the old, unmaintained FXC compiler. You can choose which compiler to use at `Instance` creation using the `Dx12Compiler` field in the `InstanceOptions` struct. Note that DXC requires both `dxcompiler.dll` and `dxil.dll`, which can be downloaded from https://github.com/microsoft/DirectXShaderCompiler/releases. Both .dlls need to be shipped with your application when targeting DX12 and using the `DXC` compiler. If the .dlls can't be loaded, then it will fall back to the FXC compiler. By @39ali and @Elabajaba in [#3356](https://github.com/gfx-rs/wgpu/pull/3356)
+
 ### Changes
 
 #### General
 
+- `Instance` creation now takes an `InstanceOptions` struct which cointains both the existing `Backends` selection as well as a new `Dx12Compiler` field for selecting which Dx12 shader compiler to use. By @Elabajaba in [#3356](https://github.com/gfx-rs/wgpu/pull/3356)
```
Elabajaba, line 118: TODO: This is a major change.

player/src/bin/play.rs
```diff
@@ -43,7 +43,11 @@ fn main() {
         .build(&event_loop)
         .unwrap();
 
-    let global = wgc::hub::Global::new("player", IdentityPassThroughFactory, wgt::Backends::all());
+    let global = wgc::hub::Global::new(
+        "player",
+        IdentityPassThroughFactory,
+        wgt::InstanceOptions::new(wgt::Backends::all(), wgt::Dx12Compiler::Fxc),
```
Elabajaba, line 49: TODO: This could be InstanceOptions::default()

Elabajaba (commented):
A few changes, and a reminder for myself that I need to retest this on my laptop.

wgpu-hal/src/dx12/shader_compilation.rs
cwfitzgerald: Awesome!

wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -0,0 +1,253 @@
...
+                hassle_rs::HassleError::ValidationError(_e) => unimplemented!(),
+                hassle_rs::HassleError::CompileError(_e) => unimplemented!(),
+            }
+        }
+    }
+}
+
+// These are stubs for when the `dxc_shader_compiler` feature is disabled.
+#[cfg(not(feature = "dxc_shader_compiler"))]
+mod shader {
```
cwfitzgerald, line 224: Maybe `dxc` or something for module name

wgpu-types/src/lib.rs
```diff
@@ -5293,3 +5294,55 @@ impl Default for ShaderBoundChecks {
...
+    /// The Dxc compiler is new, fast and maintained.
+    /// However, it requires both `dxcompiler.dll` and `dxil.dll` to be shipped with the application.
+    /// These files can be downloaded from https://github.com/microsoft/DirectXShaderCompiler/releases
+    Dxc {
+        /// Path to the `dxcompiler.dll` file. Passing `None` will check the local scope for the file.
+        dxil_path: Option<PathBuf>,
+        /// Path to the `dxil.dll` file. Passing `None` will check the local scope for the file.
+        dxc_path: Option<PathBuf>,
+    },
+}
```
cwfitzgerald, line 5321:
```suggestion
/// Selects which DX12 shader compiler to use.
///
/// If the `wgpu-hal/dx12-shader-compiler` isn't enabled,
/// then this will fall back to the Fxc compiler at runtime and log an error.
/// This feature is always enabled when using `wgpu`.
///
/// If the `Dxc` option is selected, but the `dxcompiler.dll` and `dxil.dll` files aren't found,
/// then this will fall back to the Fxc compiler at runtime and log an error.
///
/// `wgpu::utils::init::dx12_shader_compiler_from_env` can be used to set the compiler
/// from the `WGPU_DX12_SHADER_COMPILER` environment variable,
///
/// but this should only be used for testing.
#[derive(Clone, Debug, Default)]
pub enum Dx12Compiler {
    /// The Fxc compiler (default) is old, slow and unmaintained. However, it doesn't require
    /// any additional .dlls to be shipped with the application.
    #[default]
    Fxc,
    /// The Dxc compiler is new, fast and maintained.
    /// 
    /// However, it requires both `dxcompiler.dll` and `dxil.dll` to be shipped with the application.
    /// These files can be downloaded from https://github.com/microsoft/DirectXShaderCompiler/releases
    Dxc {
        /// Path to the `dxil.dll` file. Passing `None` will use standard platform specific dll loading rules.
        dxil_path: Option<PathBuf>,
        /// Path to the `dxcompiler.dll` file. Passing `None` will use standard platform specific dll loading rules.
        dxc_path: Option<PathBuf>,
    },
}
```

Just some documentation refinements.

wgpu-types/src/lib.rs
```diff
@@ -5293,3 +5294,55 @@ impl Default for ShaderBoundChecks {
...
+        Self {
+            backends: Backends::all(),
+            dx12_shader_compiler: Dx12Compiler::default(),
+        }
+    }
+}
+
+impl InstanceOptions {
+    /// Creates a new `InstanceOptions` with the given `Backends` and `Dx12Compiler`.
+    pub fn new(backends: Backends, dx12_shader_compiler: Dx12Compiler) -> Self {
```
cwfitzgerald, line 5342: IMO there's no reason to bother with a constructor, especially when convention in wgpu is to just fill out the struct.

wgpu/src/lib.rs
```diff
@@ -1277,16 +1277,24 @@ impl Display for SurfaceError {
 
 impl error::Error for SurfaceError {}
 
+impl Default for Instance {
+    /// Creates a new instance of wgpu with default options.
+    /// Backends are set to `Backends::all()`, and FXC is chosen as the `dx12_shader_compiler`.
```
cwfitzgerald, line 1282:
```suggestion
    /// Creates a new instance of wgpu with default options.
    ///
    /// Backends are set to `Backends::all()`, and FXC is chosen as the `dx12_shader_compiler`.
```

The break will make the first line the summary shown on the module page

wgpu/src/util/init.rs
```diff
@@ -86,3 +86,21 @@ pub async fn initialize_adapter_from_env_or_default(
         }
     }
 }
+
+/// Choose which DX12 shader compiler to use from the environment variable WGPU_DX12_COMPILER
```
cwfitzgerald, line 90:
```suggestion
/// Choose which DX12 shader compiler to use from the environment variable `WGPU_DX12_COMPILER`. Possible values are `dxc` and `fxc`. Case insensitive.
```

wgpu/src/util/init.rs
cwfitzgerald: It just defers to the system style - for this kind of thing "just put it in your working dir" is perfectly reasonable - so you made the right call not bothering. The existance of this env var should be documented in the readme though.

wgpu-types/src/lib.rs
```diff
@@ -5293,3 +5294,55 @@ impl Default for ShaderBoundChecks {
...
+    Dxc {
+        /// Path to the `dxcompiler.dll` file. Passing `None` will check the local scope for the file.
+        dxil_path: Option<PathBuf>,
+        /// Path to the `dxil.dll` file. Passing `None` will check the local scope for the file.
+        dxc_path: Option<PathBuf>,
+    },
+}
+
+/// Options for creating an instance.
+pub struct InstanceOptions {
```
cwfitzgerald, line 5324: Nit: for consistency this should be `InstanceDescriptor`

wgpu-hal/src/dx12/shader_compilation.rs
cwfitzgerald: The include handler allows you to specify `#include`s in your hlsl and the handler will provide the actual text of the include.

cwfitzgerald (changes requested):
This is really shaping up well. Got some comments, but nothing show stopping!

wgpu-hal/src/dx12/adapter.rs
cwfitzgerald: It does, but cloning here is fine, this is slow code.

wgpu-types/src/lib.rs
Elabajaba: I found the constructor to be a bit less verbose, but I guess it's better to follow convention here.

wgpu-types/src/lib.rs
Elabajaba: `InstanceDescriptor` is already used, so I opted to avoid the name clash.

wgpu-types/src/lib.rs
cwfitzgerald: It is, but I anticipate this struct will have quite a few fields at the end of the day.

wgpu-core/src/instance.rs
```diff
@@ -67,8 +67,12 @@ pub struct Instance {
 }
 
 impl Instance {
-    pub fn new(name: &str, backends: Backends) -> Self {
-        fn init<A: HalApi>(_: A, mask: Backends) -> Option<A::Instance> {
+    pub fn new(name: &str, instance_options: wgt::InstanceOptions) -> Self {
+        fn init<A: HalApi>(
+            _: A,
+            mask: Backends,
+            dx12_shader_compiler: Option<wgt::Dx12Compiler>,
```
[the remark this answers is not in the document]
Elabajaba: I feel like it's cleaner to pass `None` to all the non dx12 backends here, rather than passing it in and just having them ignore it (though I guess they still have the field on their `InstanceDescriptor`).

deno_webgpu/src/lib.rs
```diff
@@ -240,7 +240,10 @@ pub async fn op_webgpu_request_adapter(
         state.put(wgpu_core::hub::Global::new(
             "webgpu",
             wgpu_core::hub::IdentityManagerFactory,
-            backends,
+            wgpu_types::InstanceOptions {
+                backends,
+                dx12_shader_compiler: wgpu_types::Dx12Compiler::Fxc,
```
crowlKats, line 245: I am not sure what exactly this PR does as I am not familiar with this, however I see that in wgpu its going to be configurable via an env var; would it make sense for this to be configurable for deno, or is it not necessary enough of a feature?

Elabajaba: It's letting users use a better DX12 shader compiler (newer, faster, maintained, allows for newer feature support), but it comes at the cost of requiring users to ship `dxil.dll` and `dxcompiler.dll`, so I just defaulted to using Fxc (the old unmaintained one) and having users opt into [Dxc](https://github.com/microsoft/DirectXShaderCompiler) (the new one). FXC only support up to ShaderModel 5.1, while DXC supports ShaderModel 6.0 and higher.

An example of why you might want to use DXC is if you wanted to add [FSR2](https://github.com/GPUOpen-Effects/FidelityFX-FSR2) to a game engine, as it requires ShaderModel 6.2 (or it can use 6.6 for 64 wide wavefronts), or if you wanted to use DirectX Raytracing (DXR/RTX/etc).

You can look at the [DXC wiki](https://github.com/microsoft/DirectXShaderCompiler/wiki/Shader-Model-6.0) for what features ShaderModel 6.0+ adds.

I wasn't sure if deno even used DX12 on Windows, or if the deno maintainers considered having potentially backwards incompatible shaders (that work with DXC, but not FXC) a worthwhile tradeoff for getting support for 64bit ints, native 16bit types, DXR support, etc.

crowlKats: > but it comes at the cost of requiring users to ship `dxil.dll` and `dxcompiler.dll`

so a user of wgpu has to manually link to these, or is this taken care of?

If this is taken care of, and its only a matter of specifying the right enum value, I would say it seems favourable to support via a `DENO_WEBGPU_DX12_COMPILER` env var

Elabajaba: The wgpu user has to do it, but linking them just requires dropping the .dlls in the folder, and making sure they're included in what's sent to the end user (that eg. plays the game that uses wgpu and the dxc compiler backend).

wgpu-core/src/instance.rs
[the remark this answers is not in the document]
Elabajaba: So it might make sense to pass in `InstanceOptions` here if the `InstanceOptions` struct is going to get bigger, to avoid the `init()` call here from getting too verbose, but you'll still have to destructure it in the `init()` function here to transform it into a `hal::InstanceDescriptor`. IMO passing in None for backends where the feature being passed in isn't used is also clearer, but I guess it doesn't matter that much.

deno_webgpu/src/lib.rs
crowlKats: hm, i guess for now lets leave it as-is

wgpu-types/src/lib.rs
Elabajaba: Renamed to `InstanceDescriptor`

wgpu-core/src/instance.rs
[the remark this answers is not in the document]
Elabajaba: This passes in `wgt::InstanceDescriptor` now, but changes `hal::InstanceDescriptor` to use `Dx12Compiler` instead of `Option<Dx12Compiler>`.

Elabajaba:
I think everything should be cleaned up now.

cwfitzgerald (approved):
Amazing stuff!

PR #3364, opened 2023-01-09
fix(core): don't panic on mapped buffer in queue_submit

crowlKats:
Reported in https://github.com/denoland/deno/issues/10098


nical (approved):
LGTM

nical:
This is just missing a changelog entry.

PR #3374, opened 2023-01-12
Increase GL MAX_PUSH_CONSTANTS from 16 to 64

Dinnerbone:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
None.

**Description**
I've found that using the emulated push constants is significantly faster than buffer juggling on our own on the webgl backend. However, it's limited to 16*4 bytes right now, which is too small for our needs.

I can't find any rationale behind it behind just 16*4, it's possible it was a naga limitation but it doesn't seem to be the case anymore? The docs for `Limits::max_push_constant_size` suggest it should be 256 on GLES so this increases it to that, which can hold a couple of matrices nicely.

If I'm wrong and there's a limitation I missed somewhere and this *doesn't* actually work somehow, then okay, nevermind! But I've tested this out on Ruffle and it seems to work perfect and performant for us.

**Testing**
Manually tested and it works fine, not sure it deserves its own unit test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -133,6 +133,7 @@ Additionally `Surface::get_default_config` now returns an Option and returns Non
 #### GLES
 
 - Browsers that support `OVR_multiview2` now report the `MULTIVIEW` feature by @expenses in [#3121](https://github.com/gfx-rs/wgpu/pull/3121).
+- `Limits::max_push_constant_size` on GLES is now 256 by @Dinnerbone in [#3374](https://github.com/gfx-rs/wgpu/pull/3374).
 
 #### Vulkan
 
--- a/wgpu-hal/src/gles/device.rs
+++ b/wgpu-hal/src/gles/device.rs
@@ -372,7 +372,8 @@ impl super::Device {
             }
         }
 
-        let mut uniforms: [super::UniformDesc; super::MAX_PUSH_CONSTANTS] = Default::default();
+        let mut uniforms: [super::UniformDesc; super::MAX_PUSH_CONSTANTS] =
+            [None; super::MAX_PUSH_CONSTANTS].map(|_: Option<()>| Default::default());
         let count = unsafe { gl.get_active_uniforms(program) };
         let mut offset = 0;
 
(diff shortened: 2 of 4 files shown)


cwfitzgerald (approved):
LGTM, 256 is the standard push constant limit anyway

cwfitzgerald:
You have a test failing - but also we might have issues with the number of uniforms - we probably should at least be consulting the limits set by GL

Dinnerbone:
Darn, I couldn't test that here trivially. It looks like the test was set up assuming a size of 16 though, if I'm reading this right. Will need to find a way to adjust that.

Dinnerbone:
Actually, am I right in reading that this test wouldn't run if the push constants limit was less than 128? If so, this test was implicitly skipped until now and has never run for GLES. I can still find a way to make it work though!

https://github.com/gfx-rs/wgpu/blob/f0f700c2d93e08334752e4ee109b48416e60690c/wgpu/tests/shader/struct_layout.rs#L221-L224

cwfitzgerald:
Heh, you are. If you can make it work great! Otherwise you can mark it as a failure and file a followup issue.

Dinnerbone:
Looked into this a bit, I'll mark it as failing and open an issue. In short:
The test makes a `PushConstantRange(0..128)` whilst only allocating (for example) `vec2<f32>` in the shader. When the render pass tries to use the push constants, in GL that means uploading uniforms and since there's only actually a few bytes for uniforms... that's an error.

It definitely didn't work before this PR, but it was ignored until now. I'm not sure how this is supposed to work though, how do other backends handle `var<push_constant> input: CustomStruct;` where `struct CustomStruct { member: vec2<f32>, }` whilst also allowing the user to say "I have 128 bytes", and not error out?

cwfitzgerald:
> where struct CustomStruct { member: vec2<f32>, } whilst also allowing the user to say "I have 128 bytes", and not error out?

The shader is just the view into the push constants - you're allowed to push up more and only view part of it.

cwfitzgerald (approved):

PR #3380, opened 2023-01-14
GLES: Cache and reuse programs between similar pipelines

Dinnerbone:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
None

**Description**
It's a common practice to create many, many pipelines due to a lot of state being "per pipeline" and not adjustable on the fly, such as blend modes. This turns them into an often giant explosion of permutations for shaders vs blend vs depth operations (and other things). As the GL pipeline unfortunately needs to block on the creation of each program, and we create one program per pipeline, this can lead to stalls over 10 seconds long as we create the pipelines needed for our application.

This change decouples a Pipeline from a GL program, and allows sharing of the same program across many pipelines. As long as the inputs are the same, the program created will always be the same and thus can be shared safely.

For ruffle with webgl, this reduces a startup time from about 10 seconds to almost instantaneous due to the amount of programs we create, for a very few combinations of shaders.

**Testing**
All the existing tests pass and I've monitored that programs do indeed get destroyed when every pipeline that uses them is deleted.

I've manually tested Ruffle with this change to confirm the speedup, and it's fast and stable.

I haven't created any new tests, I'm not sure how to test this automatically.

### DIFF
--- a/wgpu-hal/src/gles/device.rs
+++ b/wgpu-hal/src/gles/device.rs
@@ -3,12 +3,14 @@ use crate::auxil::map_naga_stage;
 use glow::HasContext;
 use std::{
     convert::TryInto,
-    iter, ptr,
+    ptr,
     sync::{Arc, Mutex},
 };
 
+use arrayvec::ArrayVec;
 #[cfg(not(target_arch = "wasm32"))]
 use std::mem;
+use std::sync::atomic::{AtomicU32, Ordering};
 
 type ShaderStage<'a> = (
     naga::ShaderStage,
@@ -265,14 +267,60 @@ impl super::Device {
         unsafe { Self::compile_shader(gl, &output, naga_stage, stage.module.label.as_deref()) }
     }
 
-    unsafe fn create_pipeline<'a, I: Iterator<Item = ShaderStage<'a>>>(
+    unsafe fn create_pipeline<'a>(
         &self,
         gl: &glow::Context,
-        shaders: I,
+        shaders: ArrayVec<ShaderStage<'a>, 3>,
         layout: &super::PipelineLayout,
         #[cfg_attr(target_arch = "wasm32", allow(unused))] label: Option<&str>,
         multiview: Option<std::num::NonZeroU32>,
-    ) -> Result<super::PipelineInner, crate::PipelineError> {
+    ) -> Result<Arc<super::PipelineInner>, crate::PipelineError> {
+        let mut program_stages = ArrayVec::new();
+        let mut group_to_binding_to_slot = Vec::with_capacity(layout.group_infos.len());
+        for group in &*layout.group_infos {
+            group_to_binding_to_slot.push(group.binding_to_slot.clone());
+        }
+        for &(naga_stage, stage) in &shaders {
+            program_stages.push(super::ProgramStage {
+                naga_stage: naga_stage.to_owned(),
+                shader_id: stage.module.id,
+                entry_point: stage.entry_point.to_owned(),
+            });
+        }
+        let glsl_version = match self.shared.shading_language_version {
+            naga::back::glsl::Version::Embedded { version, .. } => version,
+            naga::back::glsl::Version::Desktop(_) => unreachable!(),
+        };
+        let mut guard = self.shared.program_cache.lock();
+        let program = guard
+            .entry((program_stages, group_to_binding_to_slot.into_boxed_slice()))
+            .or_insert_with(|| unsafe {
+                Self::create_program(
+                    gl,
+                    shaders,
+                    layout,
+                    label,
+                    multiview,
+                    glsl_version,
+                    self.shared.private_caps,
+                )
+            })
+            .to_owned()?;
+        program.ref_count.fetch_add(1, Ordering::Relaxed);
+        drop(guard);
+
+        Ok(program)
+    }
+
+    unsafe fn create_program<'a>(
+        gl: &glow::Context,
+        shaders: ArrayVec<ShaderStage<'a>, 3>,
+        layout: &super::PipelineLayout,
+        #[cfg_attr(target_arch = "wasm32", allow(unused))] label: Option<&str>,
+        multiview: Option<std::num::NonZeroU32>,
+        glsl_version: u16,
+        private_caps: super::PrivateCapabilities,
+    ) -> Result<Arc<super::PipelineInner>, crate::PipelineError> {
         let program = unsafe { gl.create_program() }.unwrap();
         #[cfg(not(target_arch = "wasm32"))]
         if let Some(label) = label {
@@ -302,11 +350,7 @@ impl super::Device {
 
         // Create empty fragment shader if only vertex shader is present
         if has_stages == wgt::ShaderStages::VERTEX {
-            let version = match self.shared.shading_language_version {
-                naga::back::glsl::Version::Embedded { version, .. } => version,
-                naga::back::glsl::Version::Desktop(_) => unreachable!(),
-            };
-            let shader_src = format!("#version {} es \n void main(void) {{}}", version,);
+            let shader_src = format!("#version {} es \n void main(void) {{}}", glsl_version,);
             log::info!("Only vertex shader is present. Creating an empty fragment shader",);
             let shader = unsafe {
                 Self::compile_shader(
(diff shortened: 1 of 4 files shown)


wgpu-hal/src/gles/device.rs
```diff
@@ -265,14 +267,60 @@ impl super::Device {
...
+                naga_stage: naga_stage.to_owned(),
+                shader_id: stage.module.id,
+                entry_point: stage.entry_point.to_owned(),
+            });
+        }
+        let glsl_version = match self.shared.shading_language_version {
+            naga::back::glsl::Version::Embedded { version, .. } => version,
+            naga::back::glsl::Version::Desktop(_) => unreachable!(),
+        };
+        let mut guard = self.shared.program_cache.lock();
```
cwfitzgerald, line 294: This should be try_lock() - there is no situation where this is called where we aren't rlready behind the gl context lock

wgpu-hal/src/gles/device.rs
```diff
@@ -302,11 +350,7 @@ impl super::Device {
 
         // Create empty fragment shader if only vertex shader is present
         if has_stages == wgt::ShaderStages::VERTEX {
-            let version = match self.shared.shading_language_version {
-                naga::back::glsl::Version::Embedded { version, .. } => version,
-                naga::back::glsl::Version::Desktop(_) => unreachable!(),
-            };
-            let shader_src = format!("#version {} es \n void main(void) {{}}", version,);
+            let shader_src = format!("#version {} es \n void main(void) {{}}", glsl_version,);
```
cwfitzgerald, line 353: We do eventually want to support desktop glsl - panicing here I think is more useful than silently generating bad glsl

wgpu-hal/src/gles/device.rs
```diff
@@ -403,11 +443,12 @@ impl super::Device {
             }
         }
 
-        Ok(super::PipelineInner {
+        Ok(Arc::new(super::PipelineInner {
             program,
             sampler_map,
             uniforms,
-        })
+            ref_count: AtomicU32::new(0),
```
cwfitzgerald, line 450: Why a separate refcount vs using the Arc's refcount?

wgpu-hal/src/gles/mod.rs
```diff
@@ -495,8 +504,20 @@ struct ColorTargetDesc {
...
+struct ProgramStage {
+    naga_stage: naga::ShaderStage,
+    shader_id: ShaderId,
+    entry_point: String,
+}
+
+type ProgramCache = FastHashMap<
+    (ArrayVec<ProgramStage, 3>, Box<[Box<[u8]>]>),
+    Result<Arc<PipelineInner>, crate::PipelineError>,
+>;
```
cwfitzgerald, line 517: Holy type batman!

wgpu-hal/src/gles/device.rs
```diff
@@ -1133,23 +1175,38 @@ impl crate::Device<super::Api> for super::Device {
...
     }
     unsafe fn destroy_compute_pipeline(&self, pipeline: super::ComputePipeline) {
-        let gl = &self.shared.context.lock();
-        unsafe { gl.delete_program(pipeline.inner.program) };
+        let mut program_cache = self.shared.program_cache.lock();
+        if pipeline.inner.ref_count.fetch_sub(1, Ordering::Relaxed) == 1 {
+            program_cache.retain(|_, v| match *v {
+                Ok(ref p) => p.program != pipeline.inner.program,
+                Err(_) => false,
+            });
```
cwfitzgerald, line 1206: Why does this need to be a retain call vs a remove?

cwfitzgerald (changes requested):
Great idea, good execution, have some questions/comments but looking forward to getting this in.

wgpu-hal/src/gles/device.rs
Dinnerbone: I agree, this was old code that was just split apart and the panic still happens above.

wgpu-hal/src/gles/device.rs
Dinnerbone: We can't do anything with the program cache if the lock fails to acquire, so would you have it silently fail or panic on None?

wgpu-hal/src/gles/device.rs
Dinnerbone: Good point!

wgpu-hal/src/gles/device.rs
Dinnerbone: Remove requires the key which we don't have. We can either calculate the key at cost of more cycles and storing more info inside the program, or store a copy of the key entirely at cost of more memory, but this felt right for now. With no performance metrics I have no idea which one would be best, so if you have a preference I'm okay switching!

wgpu-hal/src/gles/device.rs
cwfitzgerald: I would panic - it should be unreachable.

cwfitzgerald: We're basically using a lock as an atomic refcell

Dinnerbone: But then isn't `try_lock().expect()` the same as `lock()`? Or do you specifically want an `unreachable!()` panic?

cwfitzgerald: Yeah - if we get it wrong with lock, we deadlock, if we get it wrong with try_lock, we panic which is much more discoverable.

Dinnerbone: Ahh of course, I think I'm mixing up my locks. I'll switch it over!

cwfitzgerald:
You re-requested review but didn't push anything 😄

Dinnerbone:
That's... huh. Where the hell am I pushing things?!
Sorry about that, one moment!

Dinnerbone:
I had a bunch of commits since I made the PR and apparently I've been pushing them to the wrong place. I'm juggling too many things 😅

wgpu-hal/src/gles/device.rs
```diff
@@ -1133,23 +1180,40 @@ impl crate::Device<super::Api> for super::Device {
         })
     }
     unsafe fn destroy_render_pipeline(&self, pipeline: super::RenderPipeline) {
-        let gl = &self.shared.context.lock();
-        unsafe { gl.delete_program(pipeline.inner.program) };
+        let mut program_cache = self.shared.program_cache.lock();
+        // 1 means we're the last thing holding this, so we can safely clean it up
+        if Arc::strong_count(&pipeline.inner) == 1 {
```
cwfitzgerald, line 1185: This is probably correct due to the program cache lock, but if that wasn't there, this would be ever so subtly wrong. Arc::try_unwrap would get rid of that issue by doing the decrement and the extraction in the same step. It's a bit messier syntactically, but much more clearly correct.

cwfitzgerald (approved):
G2G after nit

cwfitzgerald:
Thanks for all your help on the GL backend!

wgpu-hal/src/gles/device.rs
Dinnerbone: If the lock weren't there it wouldn't compile 😅
I can try that but you're right that it's going to be a little messy.

The actual last holder is inside the cache, but we'll need to drop the parameter `pipeline` to ensure it'll be the final final one. `Arc::try_unwrap` requires passing in an owned arc, not a ref, so we'd need to take it out of the cache but we can't/shouldn't take it out of the cache until we know it's the last one.

(I just noticed with the change to Arc refcount instead of the owned refcount, behaviour slightly changed: now it's counting the users handle of a `RenderPipeline`, whereas before I didn't care about that. That means the `1` should actually be `2` here, or I explicitly drop the passed-in pipeline first.)

I can only think of taking out of the cache, doing the test, and adding it back in - but that's even more error prone. What do you think?

cwfitzgerald: Alright, yeah that seems quite a bit more complicated 😅. Sticking with strong count is fine then, but could you add comments to this line (and things that increase the refcount) why it's correct?

danielhjacobs: I believe one of you needs to mark this conversation as resolved for the auto-merge to go through.

cwfitzgerald: Thanks! Didn't realize it wasn't resolved

yudshj:
Hello, may I ask if it’s possible to **cache and reuse similar pipelines** based on a similar approach, rather than just caching the program? This would help save time in assembling WebGL pipelines.

Wumpf:
It sure could be done, but the idea here is that only the program creation is the actual slow part. Everything else is mostly just ephemeral state that needs to be set during rendering, so I don't think there is any speedup to be gained.

PR #3384, opened 2023-01-15
deno fixes

crowlKats:
Changes mainly taken from changes i did in #3068

~sadly basically have to duplicate the `wgpu_types::Limits` struct in js; only other solution would be adding serde default attributes to all members of the struct, which doesnt seem too great either...~

Also fixed the default values & order to match the spec & adds serde renames for the texture dimension members.

~If adding serde default attributes would be acceptable, then i would much rather do that than having to need to have similar things spread around~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -549,8 +549,6 @@ dependencies = [
 [[package]]
 name = "deno_console"
 version = "0.84.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


cwfitzgerald (approved):
I don't see a problem with serde defaulting - at least it'll be in the same area of code. Want to get a +1 from a ff person first though before merging.

nical (approved):
LGTM!

PR #3386, opened 2023-01-15
In IntelliJ Rust plugin, Declarative macro expansion of bitflags! fails when a backslash is present in rustdoc.

823984418:
**Checklist**
No check was performed, this only changed the comment.
- [ ] Run `cargo clippy`.

**Connections**
This is because of the problem of intellij-rust. Until this problem is solved, avoiding it may temporarily improve the user experience.
https://github.com/intellij-rust/intellij-rust/issues/9843#issue-1475060423

**Description**
Due to a problem with intellij-rust, deleting these backslashes will enable the macro to expand correctly, thus improving the code experience.

**Testing**
No test was performed, this only changed the comment.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -360,14 +360,14 @@ bitflags::bitflags! {
         /// Allows the user to create uniform arrays of textures in shaders:
         ///
         /// ex.
-        /// `var textures: binding_array<texture_2d<f32>, 10>` (WGSL)\
+        /// `var textures: binding_array<texture_2d<f32>, 10>` (WGSL)
         /// `uniform texture2D textures[10]` (GLSL)
         ///
         /// If [`Features::STORAGE_RESOURCE_BINDING_ARRAY`] is supported as well as this, the user
         /// may also create uniform arrays of storage textures.
         ///
         /// ex.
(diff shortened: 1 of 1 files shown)


wgpu-types/src/lib.rs
```diff
@@ -360,14 +360,14 @@ bitflags::bitflags! {
         /// Allows the user to create uniform arrays of textures in shaders:
         ///
         /// ex.
-        /// `var textures: binding_array<texture_2d<f32>, 10>` (WGSL)\
+        /// `var textures: binding_array<texture_2d<f32>, 10>` (WGSL)
```
cwfitzgerald, line 363: If we're removing the slash - let's add two spaces so that they show up on separate lines. Hopefully ijr doesn't have an issue with that.

823984418: Yes, two spaces can also work properly.

cwfitzgerald (approved):
tysm!

wgpu-types/src/lib.rs
823984418: But rustfmt seems to have a problem with that.

cwfitzgerald: Bleh, let's just add a new paragraph then

cwfitzgerald:
CI is still mad :|

cwfitzgerald:
Converted to a bullet point list

PR #3390, opened 2023-01-17
Handle naga's 16-bit unorm/snorm formats

fintelia:
**Checklist**

- [ ] Run `cargo clippy`.

**Connections**
Depends on gfx-rs/naga/pull/2210

**Description**
Add mappings to/from naga's 16-bit unorm and snorm storage formats.

**Testing**
None yet.


teoxoy (changes requested):
- we should add the formats to `map_storage_format_to_naga` as well
- https://github.com/gfx-rs/naga/pull/2210 introduced a new capability that needs to be set if we have `Features::TEXTURE_FORMAT_16BIT_NORM`

teoxoy:
We ended up doing it in https://github.com/gfx-rs/wgpu/pull/3405.

PR #3396, opened 2023-01-18
Don't panic when gl.create_buffer() fails on webgl

Dinnerbone:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
n/a

**Description**
In webgl `gl.create_buffer()` is fallible, and can return null (or Err, thanks to glow) when the device is unable to allocate any more buffers. We unwrap and crash instead of treating it as we do other OOM cases.

Side note: Has there been a discussion already about potentially disallowing `unwrap()` going forwards? So many unwraps are quite fragile :( `expect` forces developers to give a reason to why the panic should never happen.

**Testing**
Passes existing tests.


cwfitzgerald (approved):
> Side note: Has there been a discussion already about potentially disallowing unwrap() going forwards? So many unwraps are quite fragile :( expect forces developers to give a reason to why the panic should never happen.

We haven't - my gut feeling is that the verbosity of expect might get very annoying for trivial stuff.

PR #3401, opened 2023-01-19
Add `SHADERINT_16` feature to allow 16bit ints in Vulkan shaders

Elabajaba:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Stops Vulkan Validation errors for FSR2 when using https://github.com/JMS55/fsr2_wgpu (currently very much a WIP).

**Description**
This just enables the Vulkan `shader_int16` device feature if the feature is requested. I copied how it seems to be done for `SHADER_FLOAT64`.

According to [the docs](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceFeatures.html), this isn't an extension and is part of vulkan 1.0 `VkPhysicalDeviceFeatures`. ~~I don't know if anything doesn't support this feature, so I'm wondering if it should be enabled by default?~~ Did some more digging on https://vulkan.gpuinfo.org/ and it isn't nearly as universal as I thought.

**Testing**
Tested using bevy+fsr2, it stopped the validation errors about 16bit ints.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -674,6 +674,13 @@ bitflags::bitflags! {
         /// - DX12
         /// - Metal (Intel and AMD GPUs)
         const WRITE_TIMESTAMP_INSIDE_PASSES = 1 << 41;
+        /// Allows shaders to acquire the INT16 ability
+        ///
+        /// Supported platforms:
+        /// - Vulkan
+        ///
+        /// This is a native-only feature.
+        const SHADER_INT16 = 1 << 42;
     }
 }
 
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -674,7 +674,7 @@ bitflags::bitflags! {
         /// - DX12
         /// - Metal (Intel and AMD GPUs)
         const WRITE_TIMESTAMP_INSIDE_PASSES = 1 << 41;
-        /// Allows shaders to use i16. Not currently supported in naga, only available through spirv-passthrough.```
+        /// Allows shaders to use i16. Not currently supported in naga, only available through `spirv-passthrough`.
         ///
         /// Supported platforms:
         /// - Vulkan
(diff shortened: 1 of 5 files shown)


crowlKats (changes requested):
don't forget to update `deserialize_features` in deno_webgpu/src/lib.rs, and update the `GPUFeatureName` enum in deno_webgpu/src/02_idl_types.js

teoxoy:
Wanted to mention that I filed https://github.com/gfx-rs/wgpu/issues/3406; not a blocker for this PR but something to put on our radar.

Elabajaba:
> don't forget to update `deserialize_features` in deno_webgpu/src/lib.rs, and update the `GPUFeatureName` enum in deno_webgpu/src/02_idl_types.js

Does the `GPUFeatureName` enum in deno_wgpu/webgpu.idl also have to be updated?

crowlKats:
> Does deno_webgpu want what are potentially native only features?

yes; if my memory serves right, the spec states that the implementer can extend the standard set of features. we just allow all features that wgpu has, because why not.

Elabajaba:
I think this should be done now.

crowlKats:
thanks!

crowlKats (commented):
Deno part looks good to me.

wgpu-types/src/lib.rs
```diff
@@ -674,6 +674,13 @@ bitflags::bitflags! {
         /// - DX12
         /// - Metal (Intel and AMD GPUs)
         const WRITE_TIMESTAMP_INSIDE_PASSES = 1 << 41;
+        /// Allows shaders to acquire the INT16 ability
```
cwfitzgerald, line 677:
```suggestion
        /// Allows shaders to use i16. Not currently supported in naga, only available through spirv-passthrough.```

cwfitzgerald (changes requested):

cwfitzgerald (approved):

PR #3402, opened 2023-01-19
Add support for occlusion queries

valaphee:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

**Description**
At the moment its possible to create an occlusion query set, but it can't be used as there is no way to start and end a occlusion
query.

This pr implements the occlusion query set and adds the occlusion_query_set to the render pass descriptor which will be used by subsequent beginOcclusionQuery and endOcclusionQuery commands.

For finding purposes, here is a good overview of the differences for queries for different backends https://github.com/gpuweb/gpuweb/issues/614

**Testing**
I modified the mipmap example, and replaced pipeline statistics with occlusion query, as the used value (FRAGMENT_SHADER_INVOCATIONS) should be the same as the value returned by occlusion query in this case (might differ on DirectX as binary mode is used, either 0 if occluded or 1, but can be changed)


valaphee:
I also want to add, as this is my first contribution, and a rather large, to check if I did the Features Flag implementation correctly,

as for my understanding OCCLUSION_QUERY, is not a feature flag in WebGPU, and is always supported there, but atm. there is no support for this in DirectX11 FL 9.1, and Metal (Queries are generally not implemented yet)

deno_webgpu/src/01_webgpu.js
```diff
@@ -3598,12 +3598,62 @@
       );
     }
 
-    beginOcclusionQuery(_queryIndex) {
-      throw new Error("Not yet implemented");
+    /**
+     * @param {GPUQuerySet} querySet
+     * @param {number} queryIndex
+     */
+    beginOcclusionQuery(querySet, queryIndex) {
```
crowlKats, line 3605: beginOcclusionQuery doesnt accept a querySet

teoxoy:
> I also want to add, as this is my first contribution, and a rather large, to check if I did the Features Flag implementation correctly, as for my understanding OCCLUSION_QUERY, is not a feature flag in WebGPU, and is always supported there, but atm. there is no support for this in DirectX11 FL 9.1, and Metal (Queries are generally not implemented yet)

It should be a new flag in the `DownlevelFlags` (not a flag in `Features`).

(Downlevel flags are features that WebGPU must support but not all backends/adapters do.)

> If we want to also add query support for Metal.

Ideally, we should add core features to all primary backends (Vulkan, DX12 and Metal); not a blocker though.

valaphee:
Changed OCCLUSION_QUERY to downlevel feature, and made everything spec conform again,
but occlusion queries will be a breaking change, as it requires to add occlusion query set to the render pass descriptor

Still has to be implement in web backend, auxil, gles, and metal

deno_webgpu/src/01_webgpu.js
```diff
@@ -3598,12 +3598,52 @@
...
+      assertResource(this[_encoder], {
+        prefix,
+        context: "encoder referenced by this",
+      });
+      const renderPassRid = assertResource(this, { prefix, context: "this" });
+      assertDeviceMatch(device, querySet, {
+        prefix,
+        resourceContext: "Argument 1",
+        selfContext: "this",
+      });
```
crowlKats, line 3626:
```suggestion
```

deno_webgpu/src/01_webgpu.js
```diff
@@ -3598,12 +3598,52 @@
...
+     * @param {number} queryIndex
+     */
+    beginOcclusionQuery(queryIndex) {
+      webidl.assertBranded(this, GPURenderPassEncoderPrototype);
+      const prefix =
+        "Failed to execute 'beginOcclusionQuery' on 'GPURenderPassEncoder'";
+      webidl.requiredArguments(arguments.length, 2, { prefix });
+      queryIndex = webidl.converters.GPUSize32(queryIndex, {
+        prefix,
+        context: "Argument 2",
```
crowlKats, line 3611:
```suggestion
        context: "Argument 1",
```

deno_webgpu/src/01_webgpu.js
```diff
@@ -3598,12 +3598,52 @@
...
-    beginOcclusionQuery(_queryIndex) {
-      throw new Error("Not yet implemented");
+    /**
+     * @param {number} queryIndex
+     */
+    beginOcclusionQuery(queryIndex) {
+      webidl.assertBranded(this, GPURenderPassEncoderPrototype);
+      const prefix =
+        "Failed to execute 'beginOcclusionQuery' on 'GPURenderPassEncoder'";
+      webidl.requiredArguments(arguments.length, 2, { prefix });
```
crowlKats, line 3608:
```suggestion
      webidl.requiredArguments(arguments.length, 1, { prefix });
```

crowlKats (dismissed):

cwfitzgerald:
This is really good stuff! The code looks good - though CI and merging is mad.

Would it be possible to write a simple test (a new file under wgpu/tests/root.rs) that draws two fullscreen triangles with a depth buffer and makes sure the occlusion queries return the correct values on both?

valaphee:
Added a simple test for occlusion queries, and added occlusion queries to deno (never worked with deno before)

Theoretically there is still a check missing if the provided query set is an occlusion query set, but an panic should occur anyway when starting the query

crowlKats:
i see some problems with the deno part; will give a proper review tomorrow, but overall its good

cwfitzgerald:
@crowlKats poke from triage

wgpu/tests/occlusion_query/mod.rs
```diff
@@ -0,0 +1,136 @@
+use crate::common::{initialize_test, TestParameters};
+use std::borrow::Cow;
+use wasm_bindgen_test::wasm_bindgen_test;
+
+#[test]
+#[wasm_bindgen_test]
```
cwfitzgerald, line 6: Will do a more thorough review after the deno part is reviewed, but this was exactly what I wanted - thank you!

crowlKats (commented):
Deno part LGTM. not sure what i was thinking when i said something didnt look right 😓

valaphee:
Updated the PR, but I noticed that the https://github.com/gfx-rs/wgpu/pull/3489 PR broke the mipmap example, it runs because query_set is None for some reason when running the example test, but running the example main results in an error:
```
Caused by:
    In Device::create_buffer
    `MAP` usage can only be combined with the opposite `COPY`, requested BufferUsages(MAP_READ | QUERY_RESOLVE)
```

The examples use a staging buffer now (Don't know if this is mandatory, but seems to be intended according to WebGPU spec)

And WebGL doesn't work because queries are not immediately available.

valaphee:
GL doesn't work because of `GLES: [API/Error] ID 2 : GL_INVALID_OPERATION in glBufferSubData`

but have absolutely no idea why, checked all listed reasons, and validation

cwfitzgerald:
@valaphee Sorry this got caught up in the backlog. Could you merge in main and we'll get this reviewed and in

valaphee:
Merged with upstream

cwfitzgerald (approved):
Absolutely phenomenal work!

PR #3404, opened 2023-01-19
validate the number of color_attachments in `begin_render_pass`

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Description**
Fix a crash when the number of color_attachments in `begin_render_pass` exceeds the limit of `MAX_COLOR_ATTACHMENTS`
```log
thread 'main' panicked at 'ArrayVec: capacity exceeded in extend/from_iter', /Users/lijinlei/.cargo/registry/src/github.com-1ecc6299db9ec823/arrayvec-0.7.2/src/arrayvec.rs:1025:5
```
In the [WebGPU spec](https://gpuweb.github.io/gpuweb/#render-pass-encoder-creation), this validation is also required.

**Testing**
Tested locally

### DIFF
--- a/wgpu-core/src/command/render.rs
+++ b/wgpu-core/src/command/render.rs
@@ -436,15 +436,23 @@ impl State {
     }
 }
 
+#[derive(Clone, Debug, Error)]
+pub enum ColorAttachmentError {
+    #[error("attachment format {0:?} is not a color format")]
+    InvalidFormat(wgt::TextureFormat),
+    #[error("the number of color attachments {given} exceeds the limit {limit}")]
+    TooMany { given: usize, limit: usize },
+}
+
 /// Error encountered when performing a render pass.
 #[derive(Clone, Debug, Error)]
 pub enum RenderPassErrorInner {
     #[error(transparent)]
+    ColorAttachment(#[from] ColorAttachmentError),
+    #[error(transparent)]
     Encoder(#[from] CommandEncoderError),
     #[error("attachment texture view {0:?} is invalid")]
     InvalidAttachment(id::TextureViewId),
-    #[error("attachment format {0:?} is not a color format")]
-    InvalidColorAttachmentFormat(wgt::TextureFormat),
     #[error("attachment format {0:?} is not a depth-stencil format")]
     InvalidDepthStencilAttachmentFormat(wgt::TextureFormat),
     #[error("attachment format {0:?} can not be resolved")]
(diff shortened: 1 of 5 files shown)


wgpu-core/src/command/render.rs
```diff
@@ -189,7 +189,7 @@ pub struct RenderPassDescriptor<'a> {
 pub struct RenderPass {
     base: BasePass<RenderCommand>,
     parent_id: id::CommandEncoderId,
-    color_targets: ArrayVec<Option<RenderPassColorAttachment>, { hal::MAX_COLOR_ATTACHMENTS }>,
+    color_targets: Vec<Option<RenderPassColorAttachment>>,
```
cwfitzgerald, line 192: If we check the condition, we should be able to keep it an arrayvec

cwfitzgerald (changes requested):

wgpu-core/src/command/render.rs
jinleili: Currently RenderPass's all validation occurs within `command_encoder_run_render_pass_impl`, unless this validation is placed separately in `wgc::command::RenderPass::new()`

cwfitzgerald: How about this: we keep it an ArrayVec, silently fail to push if we run over, then keep a side tally of how many it actually tried to enqueue (with a note as to why we had to do it this way)

jinleili: I'm sorry, I don't quite understand how to do "keep a side tally", add a field to `BasePass` to record it? (only `BasePass` can be accessed in `trace::Command::RunRenderPass`)

cwfitzgerald: Yeah we sould need to keep a count - why can't we put it in the count in the renderpass directly?

wgpu/src/backend/direct.rs
```diff
@@ -1854,7 +1854,7 @@ impl crate::Context for Context {
                         channel: map_pass_channel(Some(&at.ops)),
                     })
             })
-            .collect::<ArrayVec<_, { wgc::MAX_COLOR_ATTACHMENTS }>>();
```
cwfitzgerald, line 1857: Why does this need to be a vec too? I'm not sure of the exact data flow, but we should be able to preserve the arrayvecs everywhere

jinleili: Is it acceptable to validate directly within `command_encoder_begin_render_pass`?

cwfitzgerald (approved):

PR #3405, opened 2023-01-19
Update naga to 1be8024.

nical:
**Description**

Routine naga update.


teoxoy:
Looks like CI is failing because we no longer support top level `let`, instead we should change those to `const`.

I saw you also added the 16bit normalized formats, ideally we'd also address https://github.com/gfx-rs/wgpu/pull/3390#pullrequestreview-1261732456 - up to you.

teoxoy:
Could you also add the following code block somewhere here?

https://github.com/gfx-rs/wgpu/blob/2c3f9fabb74025b7ba35c32d6ff99eab2b2af418/wgpu-core/src/device/mod.rs#L1288

```rust
caps.set(
    Caps::STORAGE_TEXTURE_16BIT_NORM_FORMATS,
    self.features.contains(wgt::Features::TEXTURE_FORMAT_16BIT_NORM),
);
```

and we can then close https://github.com/gfx-rs/wgpu/pull/3390.

Thanks!

teoxoy (approved):

PR #3409, opened 2023-01-20
Implement `view_formats` for `SurfaceConfiguration`

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Follow up to https://github.com/gfx-rs/wgpu/pull/3237

**Testing**
Tested locally except deno part

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -5354,6 +5354,14 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 hal_view_formats.push(*format);
             }
 
+            if !hal_view_formats.is_empty() {
+                if let Err(missing_flag) = device
+                    .require_downlevel_flags(wgt::DownlevelFlags::SURFACE_CONFIGURE_VIEW_FORMATS)
+                {
+                    break 'outter E::MissingDownlevelFlags(missing_flag);
+                }
+            }
+
             let num_frames = present::DESIRED_NUM_FRAMES
                 .clamp(*caps.swap_chain_sizes.start(), *caps.swap_chain_sizes.end());
             let mut hal_config = hal::SurfaceConfiguration {
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -5337,7 +5337,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 }
             };
 
-            let mut hal_view_formats = vec![config.format];
+            let mut hal_view_formats = vec![];
             for format in config.view_formats.iter() {
                 if *format == config.format {
                     continue;
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -5355,8 +5355,8 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             }
 
             if !hal_view_formats.is_empty() {
-                if let Err(missing_flag) = device
-                    .require_downlevel_flags(wgt::DownlevelFlags::SURFACE_CONFIGURE_VIEW_FORMATS)
+                if let Err(missing_flag) =
+                    device.require_downlevel_flags(wgt::DownlevelFlags::SURFACE_VIEW_FORMATS)
                 {
                     break 'outter E::MissingDownlevelFlags(missing_flag);
                 }
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -5188,7 +5188,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         &self,
         surface_id: id::SurfaceId,
         device_id: id::DeviceId,
-        config: &wgt::SurfaceConfiguration,
+        config: &wgt::SurfaceConfiguration<Vec<TextureFormat>>,
     ) -> Option<present::ConfigureSurfaceError> {
         use hal::{Adapter as _, Surface as _};
         use present::ConfigureSurfaceError as E;
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -5311,7 +5311,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         let (adapter_guard, mut token) = hub.adapters.read(&mut token);
         let (device_guard, _token) = hub.devices.read(&mut token);
 
-        let error = 'outter: loop {
+        let error = 'outer: loop {
             let device = match device_guard.get(device_id) {
                 Ok(device) => device,
                 Err(_) => break DeviceError::Invalid.into(),
@@ -5343,13 +5343,13 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                     continue;
                 }
                 if !caps.formats.contains(&config.format) {
-                    break 'outter E::UnsupportedFormat {
+                    break 'outer E::UnsupportedFormat {
                         requested: config.format,
                         available: caps.formats.clone(),
                     };
                 }
                 if config.format.remove_srgb_suffix() != format.remove_srgb_suffix() {
-                    break 'outter E::InvalidViewFormat(*format, config.format);
+                    break 'outer E::InvalidViewFormat(*format, config.format);
                 }
                 hal_view_formats.push(*format);
             }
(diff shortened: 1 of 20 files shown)


crowlKats (commented):
Deno part looks good to me.

If you want, i can test the deno change, but am unsure how viewformats are used. Though, given its a simple & straight-forward change, testing it doesnt seem too necessary

jinleili:
Looking at the #3399 implementation, I'm not sure if the Vulkan backend needs to be handled separately here as well.

teoxoy:
Thanks for updating the validation but there is one more rule:
> [Validate texture format required features](https://gpuweb.github.io/gpuweb/#abstract-opdef-validate-texture-format-required-features) of each element of configuration.[viewFormats](https://gpuweb.github.io/gpuweb/#dom-gputexturedescriptor-viewformats) with device.[[[device]]](https://gpuweb.github.io/gpuweb/#dom-gpuobjectbase-device-slot).

Ideally, we'd share some validation for the `TextureDescriptor` and the `configure` fn... but not a blocker.

> Looking at the https://github.com/gfx-rs/wgpu/pull/3399 implementation, I'm not sure if the Vulkan backend needs to be handled separately here as well.

I don't think we need to do anything extra for Vulkan here. @cwfitzgerald can maybe confirm?

cwfitzgerald:
Yeah, I think we need mutable format on the swapchain via https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSwapchainCreateFlagBitsKHR.html and VK_KHR_swapchain_mutable_format. This is fine on desktop, but has zero support on android.

jinleili:
> Yeah, I think we need mutable format on the swapchain via https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSwapchainCreateFlagBitsKHR.html and VK_KHR_swapchain_mutable_format. This is fine on desktop, but has zero support on android.

The addition of the corresponding extension `KhrImageFormatListFn` was in #3412

teoxoy:
`VK_KHR_image_format_list` can be optional but it seems to be required if we require `VK_KHR_swapchain_mutable_format`.
I think we should land https://github.com/gfx-rs/wgpu/pull/3412 first and make it required in this PR.

jinleili:
I checked the `VK_KHR_swapchain_mutable_format` extension on Android 13 devices, and sure enough, it's not supported on Android devices.

> VK_KHR_image_format_list can be optional but it seems to be required if we require VK_KHR_swapchain_mutable_format?

Yes, after re-reading the documentation, they should all be required. Since `VK_KHR_imageless_framebuffer` also requires `VK_KHR_image_format_list`, the permission of this extension has to be determined separately.

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -643,6 +643,15 @@ impl PhysicalDeviceCapabilities {
...
         }
 
+        if self.effective_api_version >= vk::API_VERSION_1_2
+            || self.supports_extension(vk::KhrImageFormatListFn::name())
+        {
+            // Optional `VK_KHR_swapchain_mutable_format`
+            if self.supports_extension(vk::KhrSwapchainMutableFormatFn::name()) {
+                extensions.push(vk::KhrSwapchainMutableFormatFn::name());
+            }
+        }
```
teoxoy, line 653:
```suggestion
        // Optional `VK_KHR_swapchain_mutable_format`
        if self.supports_extension(vk::KhrSwapchainMutableFormatFn::name()) {
            extensions.push(vk::KhrSwapchainMutableFormatFn::name());
        }
```

I'm quite certain that `supports_extension` returns `false` if the required extensions for this one are not supported.

wgpu-hal/src/vulkan/device.rs
```diff
@@ -565,6 +583,12 @@ impl super::Device {
             .clipped(true)
             .old_swapchain(old_swapchain);
 
+        let mut format_list_info = vk::ImageFormatListCreateInfo::builder();
+        if !raw_view_formats.is_empty() {
+            format_list_info = format_list_info.view_formats(&raw_view_formats);
+            info = info.push_next(&mut format_list_info);
+        }
```
teoxoy, line 590: Can we move the builder inside the if?

wgpu-hal/src/vulkan/device.rs
```diff
@@ -546,11 +546,29 @@ impl super::Device {
...
-        let info = vk::SwapchainCreateInfoKHR::builder()
-            .flags(vk::SwapchainCreateFlagsKHR::empty())
+
+        let original_format = self.shared.private_caps.map_texture_format(config.format);
+        let mut raw_flags = vk::SwapchainCreateFlagsKHR::empty();
+        let mut raw_view_formats: Vec<vk::Format> = vec![];
+        if !config.view_formats.is_empty()
+            && self
+                .enabled_device_extensions()
+                .contains(&vk::KhrSwapchainMutableFormatFn::name())
```
teoxoy, line 556: This looks like we'd silently ignore view formats.
Considering the extension has almost 0 support on android, we don't want to require it either.

I think we need a new downlevel flag for this capability.

wgpu-hal/src/vulkan/device.rs
jinleili: Moving into `if` causes `format_list_info` to have a shorter lifetime than `info`

teoxoy: Hmm, I see.

wgpu-hal/src/vulkan/device.rs
jinleili: Now there is a problem I don't know how to handle properly: we already have a `VIEW_FORMATS` downlevel flag, and inside the `frame.texture.create_view()` we no longer know if the texture is from `SurfaceTexture` or not.
https://github.com/gfx-rs/wgpu/blob/964c94a02dd13fdde4faf8ffb4a1b6756aa84fd7/wgpu-core/src/device/mod.rs#L1031-L1037

teoxoy: I think as long as we require the new downlevel flag in the configure fn then we know it should not be possible to get a surface texture with view formats set on it.

jinleili: It looks like also need to move `require_downlevel_flags(VIEW_FORMATS)?` from `create_texture_view()` to `create_texture()`?

teoxoy: Oh yeah, ideally we'd check that as soon as possible. @cwfitzgerald is there any reason why we have it in `create_texture_view()` right now?

Regarding the new downlevel flag name, maybe we should shorten it to just `SURFACE_VIEW_FORMATS`?

jinleili: OK, renamed.

wgpu-core/src/device/mod.rs
```diff
@@ -5303,7 +5303,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         let (adapter_guard, mut token) = hub.adapters.read(&mut token);
         let (device_guard, _token) = hub.devices.read(&mut token);
 
-        let error = loop {
+        let error = 'outter: loop {
```
cwfitzgerald, line 5306: Nit: 
```suggestion
        let error = 'outer: loop {
```

wgpu-hal/src/vulkan/device.rs
cwfitzgerald: @teoxoy not that I know of - it should be texture creation

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -320,7 +320,8 @@ impl PhysicalDeviceFeatures {
             | Df::UNRESTRICTED_INDEX_BUFFER
             | Df::INDIRECT_EXECUTION
             | Df::VIEW_FORMATS
-            | Df::UNRESTRICTED_EXTERNAL_TEXTURE_COPIES;
+            | Df::UNRESTRICTED_EXTERNAL_TEXTURE_COPIES
+            | Df::SURFACE_VIEW_FORMATS;
```
teoxoy, line 324: We need to set this downlevel flag based on support for `VK_KHR_swapchain_mutable_format`.

```rust
dl_flags.set(
    Df::SURFACE_VIEW_FORMATS,
    caps.supports_extension(vk::KhrSwapchainMutableFormatFn::name()),
);
```

wgpu-hal/src/vulkan/device.rs
teoxoy, line 556: Now that we have the downlevel flag, we can remove this check.

teoxoy (changes requested):
2 remaining things and we should be good to go.
Thanks for addressing all the comments!

cwfitzgerald:
Adopting this PR so we can get it in this release

wgpu-hal/src/vulkan/adapter.rs
teoxoy: @cwfitzgerald the flag still gets set here, even though it gets set/unset later, I think we should remove it for clarity.

cwfitzgerald: Oh whoops

PR #3410, opened 2023-01-20
Update `TextureView` validation

teoxoy:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

**Description**
Update `TextureView` validation according to the WebGPU spec.
I have a follow-up PR to allow creation of depth-only and stencil-only views (according to the spec).

**Testing**

### DIFF
--- a/wgpu-core/src/command/clear.rs
+++ b/wgpu-core/src/command/clear.rs
@@ -408,9 +408,8 @@ fn clear_texture_via_render_passes<A: hal::Api>(
     };
 
     let sample_count = dst_texture.desc.sample_count;
-    let is_3d_texture = dst_texture.desc.dimension == wgt::TextureDimension::D3;
     for mip_level in range.mip_range {
-        let extent = extent_base.mip_level_size(mip_level, is_3d_texture);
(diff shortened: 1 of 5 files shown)


teoxoy:
I'd also like some opinions on changing the type of `mip_level_count` and `array_layer_count` from `Option<NonZeroU32>` to `Option<u32>` (members of `ImageSubresourceRange`) since the spec does allow you to pass a 0 but validation will fail.

I imagine most people use `NonZeroU32::new` to create the value for those members, the issue being that the `new` fn will return `None` if you pass it 0; effectively changing the meaning of 0 (`None` means "use the rest" whereas 0 should just fail validation).

It would also be nice to have a resolved and validated version of `ImageSubresourceRange` that we use internally with the type of those members being a `NonZeroU32`.

nical:
> I'd also like some opinions on changing the type of `mip_level_count` and `array_layer_count` from `Option<NonZeroU32>` to `Option<u32>` (members of `ImageSubresourceRange`) since the spec does allow you to pass a 0 but validation will fail.

I agree, let's match the spec.

nical (approved):

cwfitzgerald:
NonZero values are also insanely annoying to deal with as a user, so I support this.

PR #3411, opened 2023-01-20
Make `make_spirv_raw` and `make_spirv` handle big-endian binaries

### LINKED ISSUE #3408: `make_spirv_raw` should handle big-endian SPIR-V binaries
**Description**
We used to allow both little- and big-endian SPIR-V binaries in gfx/wgpu, but now it looks like the host endianness is assumed. See https://github.com/gfx-rs/gfx/blob/3da35c4339513de6a69e7794d42a63203f5820e5/src/auxil/auxil/src/lib.rs#L84 or https://github.com/ash-rs/ash/blob/8b4575086e7a6eff4d81eb5c2cf88714e56c4128/ash/src/util.rs#L127 for reference.

**Repro steps**
n/a

Big-endian words would be passed to the Vulkan driver but this should be handled at the application level.

**Extra materials**
n/a

**Platform**
n/a

1e1001:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Fixes #3408.

**Description**
Swaps the words if that makes the first word match the magic. Also adds a check that the input isn't empty, not sure if it's needed but I think the better error message would be useful.

**Testing**
Locally tested by flipping a spir-v binary and comparing the result, if I need to add an actual test I will.

### DIFF
--- a/wgpu/src/util/mod.rs
+++ b/wgpu/src/util/mod.rs
@@ -66,7 +66,8 @@ pub fn make_spirv_raw(data: &[u8]) -> Cow<[u32]> {
         Cow::from(words)
     };
 
-    // swap if that corrects the magic
+    // Before checking if the data starts with the magic, check if it starts
+    // with the magic in non-native endianness, own & swap the data if so.
     if words[0] == MAGIC_NUMBER.swap_bytes() {
         for word in Cow::to_mut(&mut words) {
             *word = word.swap_bytes();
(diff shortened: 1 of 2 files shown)


wgpu/src/util/mod.rs
```diff
@@ -64,6 +66,13 @@ pub fn make_spirv_raw(data: &[u8]) -> Cow<[u32]> {
         Cow::from(words)
     };
 
+    // swap if that corrects the magic
```
cwfitzgerald, line 69: Could this be a bit more verbose

cwfitzgerald (changes requested):
Good minus a nit!

wgpu/src/util/mod.rs
1e1001: fair point, i don't really know how to phrase things well :)

cwfitzgerald (approved):
Tyvm!

PR #3412, opened 2023-01-21
vk: improve view_formats setting

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Follow up to https://github.com/gfx-rs/wgpu/pull/3237

**Description**
More efficient view_formats support is described in the [VK_KHR_image_format_list extension](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_image_format_list.html)

**Testing**
Only tested via `--features=vulkan-portability`

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -852,7 +852,7 @@ impl<A: HalApi> Device<A> {
             ));
         }
 
-        let mut hal_view_formats = vec![desc.format];
+        let mut hal_view_formats = vec![];
         for format in desc.view_formats.iter() {
             if desc.format == *format {
                 continue;
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -852,7 +852,7 @@ impl<A: HalApi> Device<A> {
             ));
         }
 
-        let mut allow_different_view_format = false;
+        let mut hal_view_formats = vec![desc.format];
         for format in desc.view_formats.iter() {
             if desc.format == *format {
                 continue;
@@ -860,7 +860,7 @@ impl<A: HalApi> Device<A> {
             if desc.format.remove_srgb_suffix() != format.remove_srgb_suffix() {
                 return Err(CreateTextureError::InvalidViewFormat(*format, desc.format));
             }
-            allow_different_view_format = true;
+            hal_view_formats.push(*format);
         }
 
         // Enforce having COPY_DST/DEPTH_STENCIL_WRIT/COLOR_TARGET otherwise we
@@ -893,7 +893,7 @@ impl<A: HalApi> Device<A> {
             format: desc.format,
             usage: hal_usage,
             memory_flags: hal::MemoryFlags::empty(),
-            allow_different_view_format,
+            view_formats: hal_view_formats,
         };
 
         let raw_texture = unsafe {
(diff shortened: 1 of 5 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -852,15 +852,15 @@ impl<A: HalApi> Device<A> {
             ));
         }
 
-        let mut allow_different_view_format = false;
+        let mut hal_view_formats = vec![desc.format];
```
teoxoy, line 855: Looking at the Vulkan spec, it shouldn't be necessary to pass the original format to the list.

wgpu-hal/src/vulkan/device.rs
```diff
@@ -895,11 +895,23 @@ impl crate::Device<super::Api> for super::Device {
             raw_flags |= vk::ImageCreateFlags::CUBE_COMPATIBLE;
         }
 
-        if desc.allow_different_view_format {
+        let mut hal_view_formats: Vec<vk::Format> = vec![];
+        if desc.view_formats.len() > 1 {
```
teoxoy, line 899:
```suggestion
        if !desc.view_formats.is_empty() {
```

wgpu-hal/src/vulkan/device.rs
```diff
@@ -916,6 +928,15 @@ impl crate::Device<super::Api> for super::Device {
...
 
+        let mut format_list_info = if !hal_view_formats.is_empty() {
+            Some(vk::ImageFormatListCreateInfo::builder().view_formats(&hal_view_formats))
+        } else {
+            None
+        };
+        if let Some(list_info) = format_list_info.as_mut() {
+            vk_info = vk_info.push_next(list_info);
+        }
+
```
teoxoy, line 939: Might be nicer to have just one `if` block here.

teoxoy (changes requested):
Thanks for this PR! Left a few comments, looks good otherwise.

wgpu-core/src/device/mod.rs
jinleili: I get these validation messages when I remove the original format from `hal_view_formats`:

```log
[2023-01-21T10:32:16Z ERROR wgpu_hal::vulkan::instance] VALIDATION [VUID-VkImageViewCreateInfo-pNext-01585 (0xdd1d0240)]
        Validation Error: [ VUID-VkImageViewCreateInfo-pNext-01585 ] Object 0: handle = 0xdd3a8a0000000015, name = red, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0xdd1d0240 | vkCreateImageView(): image was created with a VkImageFormatListCreateInfo in pNext of vkImageCreateInfo, but none of the formats match the VkImageViewCreateInfo::format (VK_FORMAT_R8G8B8A8_SRGB). The Vulkan spec states: If a VkImageFormatListCreateInfo structure was included in the pNext chain of the VkImageCreateInfo structure used when creating image and VkImageFormatListCreateInfo::viewFormatCount is not zero then format must be one of the formats in VkImageFormatListCreateInfo::pViewFormats (https://vulkan.lunarg.com/doc/view/1.3.204.1/mac/1.3-extensions/vkspec.html#VUID-VkImageViewCreateInfo-pNext-01585)
[2023-01-21T10:32:16Z ERROR wgpu_hal::vulkan::instance]         objects: (type: IMAGE, hndl: 0xdd3a8a0000000015, name: red)
```

teoxoy: Hmm, I see. I'd prefer if we push the original format in hal though (if view formats is not empty) since hal users might not know they need to pass in the original format as well.

teoxoy (approved):

PR #3414, opened 2023-01-21
Better error message for 16 byte alignment error

IceSentry:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
(Based on) Closes https://github.com/gfx-rs/wgpu/pull/3099
Closes https://github.com/gfx-rs/wgpu/issues/2832

**Description**
Previous error message was the generic downflags error message for BUFFER_BINDINGS_NOT_16_BYTE_ALIGNED that didn't highlight that if it was 16 bit aligned then this is supported. But it wasn't clear which shader group / bind was not aligned.

New error message:
```
      note: label = `cuboids_pipeline`
    shader group 1 binding 1 must be 16 bytes aligned as device does not support DownlevelFlags `BUFFER_BINDINGS_NOT_16_BYTE_ALIGNED`
```

Also have switched the code around to check the flag once as if it's set then we can skip the check entirely.

**Testing**
I didn't test it myself, but the original author tested and I simply updated the message to address the comment on the PR


wgpu-core/src/pipeline.rs
```diff
@@ -373,6 +373,8 @@ pub enum CreateRenderPipelineError {
         stage: wgt::ShaderStages,
         error: String,
     },
+    #[error("shader group {group} binding {binding} must be 16 bytes aligned as device does not support DownlevelFlags `BUFFER_BINDINGS_NOT_16_BYTE_ALIGNED`")]
```
cwfitzgerald, line 376:
```suggestion
    #[error("In the provided shader, the type given for group {group} binding {binding} has an alignment of {alignment}. As the device does not support `DownlevelFlags::BUFFER_BINDINGS_NOT_16_BYTE_ALIGNED` the type must have an alignment of at least 16.")]
```

Sorry, one more round on this error message - note the extra argument alignment. Adding the eixsting alignment really helps make the message clear.

IceSentry: @cwfitzgerald I'm trying to get the alignment, but I'm not sure how to go about this. Do I just get the size of the binding?

IceSentry:
@cwfitzgerald btw, I replied to the code suggestion, but I know github notifications are a bit weird with resolved discussions so I'll ping you here.

cwfitzgerald:
@IceSentry Good shout on the ping - I didn't see this. I just realized the requirement is that the _size_, not the alignment, is a multiple of 16. So my suggestion is actually slightly wrong. /s/alignment/size should take care of it. The requirement should be "a multiple of 16".

cwfitzgerald (approved):
Party party party

PR #3418, opened 2023-01-24
Add test for writing a 3D texture

almarklein:
While search for [a problem related to uploading 3D tex data](https://github.com/gfx-rs/wgpu-native/pull/238) I wrote a test to see whether the problem was perhaps in wgpu-core. It was not. But the test might still be useful 😄.

### DIFF
--- a/wgpu/tests/write_texture.rs
+++ b/wgpu/tests/write_texture.rs
@@ -7,7 +7,7 @@ use wasm_bindgen_test::*;
 
 #[test]
 #[wasm_bindgen_test]
-fn write_texture_subset() {
+fn write_texture_subset_2d() {
     let size = 256;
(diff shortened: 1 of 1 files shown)


cwfitzgerald:
Thanks for this! The test does seem to work on dx12 so you don't need the expected failure on dx12

almarklein:
Interesting, so the test failed because it passed while it was marked to fail 😄 

Anyway, how come my new tests passes on dx12 while the original fails?

cwfitzgerald:
> Interesting, so the test failed because it passed while it was marked to fail

Yup! It's how we try to keep those listings up-to-date.

> Anyway, how come my new tests passes on dx12 while the original fails?

No idea without seeing the logs, but that is kinda spooky.

cwfitzgerald (approved):

PR #3426, opened 2023-01-26
Implement `clear_buffer` on web

raphlinus:
This was "TODO" but would be very nice to have.

**Checklist**

- [x] Run `cargo clippy`.

**Connections**
There's a workaround for this in https://github.com/linebender/vello/issues/267

**Description**
The ` clear_buffer()` method is an empty TODO on the web back-end. This patch fills in the obvious implementation.

**Testing**
It's not carefully tested just yet, but we do plan to soon. Necessary steps will be landing vello#267, updating vello to wgpu 0.15 (vello#263 is the PR in progress for that), removing the hacky workaround, and verifying that it works in Chrome Canary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,14 @@ Bottom level categories:
 - Hal
 -->
 
+## Unreleased
+
+### Changes
+
+#### WebGPU
+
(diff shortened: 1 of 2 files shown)


cwfitzgerald (approved):
Thanks!

PR #3436, opened 2023-01-30
Support stencil-only views and copying to/from combined depth-stencil textures

teoxoy:
General changes:

- break down `describe` into `block_dimensions`, `required_features`, `guaranteed_format_features`, `sample_type` and `block_size`
- `sample_type` and `block_size` now take an optional `TextureAspect`

To enable copying to/from combined depth-stencil textures:
- [validation] copy aspect must refer to a single aspect of format
- [validation] update `validate_linear_texture_data` according to the WebGPU spec
- [validation] require missing `DEPTH_TEXTURE_AND_BUFFER_COPIES` downlevel flag for `queue_write_texture` and `command_encoder_copy_buffer_to_texture`
- [metal] set the right `MTLBlitOption` for texture/buffer copies
- [dx12] use specific copy texture formats and set the right texture plane

To create stencil-only views from combined depth-stencil textures:
- [metal] set the right format `X24_Stencil8`/`X32_Stencil8` for the view and `MTLTextureUsage::PixelFormatView` for the texture
- [gles] set the right `DEPTH_STENCIL_TEXTURE_MODE` to either `DEPTH_COMPONENT` or `STENCIL_INDEX`
- [dx12] set the right formats `DXGI_FORMAT_X32_TYPELESS_G8X24_UINT` and `DXGI_FORMAT_X24_TYPELESS_G8_UINT`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,34 @@ Bottom level categories:
 
 ### Major changes
 
+#### TextureFormat info API
+
+The `describe` fn was removed in favor of separate functions: `block_dimensions`, `is_compressed`, `is_srgb`, `required_features`, `guaranteed_format_features`, `sample_type` and `block_size`.
+
+```diff
+- let block_dimensions = format.describe().block_dimensions;
++ let block_dimensions = format.block_dimensions();
+- let is_compressed = format.describe().is_compressed();
++ let is_compressed = format.is_compressed();
+- let is_srgb = format.describe().srgb;
++ let is_srgb = format.is_srgb();
+- let required_features = format.describe().required_features;
++ let required_features = format.required_features();
+- let guaranteed_format_features = format.describe().guaranteed_format_features;
++ let guaranteed_format_features = format.guaranteed_format_features();
+```
+
+Additionally `sample_type` and `block_size` now take an optional `TextureAspect` and return `Option`s.
+
+```diff
+- let sample_type = format.describe().sample_type;
++ let sample_type = format.sample_type(None).expect("combined depth-stencil format requires specifying a TextureAspect");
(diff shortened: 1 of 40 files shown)


CHANGELOG.md
```diff
@@ -42,6 +42,10 @@ Bottom level categories:
 
 ### Changes
 
+#### General
```
cwfitzgerald, line 45: We should also make a major changes note about the api of `describe()` changing.

cwfitzgerald (changes requested):
This is amazing stuff - my only main ask is if there could be a test for it - with this amount of code it's hard to judge if it's actually correct

teoxoy:
I created a `ReadbackBuffers` abstraction which allows us to copy textures to buffers regardless if they support `COPY_SRC` and updated `clear_texture` and `zero_init_texture_after_discard` to use it.

It should test depth view creation indirectly since it's needed for the copy via shader and to test stencil view creation, there is some commented out code to do so.

One thing worth noting is that once we get the CTS up and running, it should also test copying to/from combined depth-stencil textures and creation of stencil-only views.

teoxoy:
WARP is failing with...

```
[2023-02-11T00:11:10Z ERROR wgpu_hal::auxil::dxgi::result] Map buffer failed: 0x887A0005
[2023-02-11T00:11:10Z ERROR wgpu_core::device::life] Mapping failed Device(Lost)
```

My Nvidia dGPU and Intel iGPU run the test without issues.

teoxoy:
The `clear_texture` GLES implementation is also failing with "texture was not fully cleared".

teoxoy:
Should be good to go now.

cwfitzgerald (approved):
Absolutely stunning work all over, as always!

Issue #3407, opened 2023-01-19
Reintroduce workspace inheritance


jimblandy:
In #3295, we had to remove the use of workspace inheritance from wgpu, because of rust-lang/cargo#11192. It seems like rust-lang/cargo#11414 has fixed the problem, so we should try re-introducing inheritance and see how that goes.

jimblandy:
Mozilla is tracking this in [Bugzilla 1804530](https://bugzilla.mozilla.org/show_bug.cgi?id=1804530).

teoxoy:
The version of cargo with the fix [will be part of rust 1.68](https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-168-2023-03-09), scheduled for 9th March. We also need to wait for mozilla-central to upgrade.

jimblandy:
Confirmed that Cargo 1.68 does indeed fix this.

At the moment, Firefox 115's MSRV is Cargo 1.66, so we still need to wait a little longer.

cwfitzgerald:
@jimblandy could you link again where we can tell ff's msrv? Would like to add it to our own documentation.

jimblandy:
https://firefox-source-docs.mozilla.org/writing-rust-code/update-policy.html

jimblandy:
It seems like the MSRV in Firefox is now 1.70, so maybe we can fix this:
https://searchfox.org/mozilla-central/source/python/mozboot/mozboot/util.py#14

cwfitzgerald:
Done in #6020. Double done in #6980

PR #3439, opened 2023-01-31
Remove some workspace inheritence from wgpu-types/Cargo.toml.

nical:
**Connections**

https://bugzilla.mozilla.org/show_bug.cgi?id=1813547

**Description**

Unfortunately that breaks the cargo vendor stuff in mozilla-central even though it is part of a build configuration that m-c does not use.


teoxoy (approved):
Thanks for the fix, I left a comment here with more info on when we should reintroduce it https://github.com/gfx-rs/wgpu/issues/3407#issuecomment-1410189137.

Issue #3179, opened 2022-11-05
PrimitiveTopology::PointList does not render on WebGL


Oberdiah:
**Description**
Any draw commands using `PrimitiveTopology::Pointlist` do not render anything when using the WebGL backend. All other backends I've tested render as expected.

**Repro steps**
`hello-triangle/shader.wgsl` (So you can see the triangle points)
```diff
@vertex
fn vs_main(@builtin(vertex_index) in_vertex_index: u32) -> @builtin(position) vec4<f32> {
    let x = f32(i32(in_vertex_index) - 1);
    let y = f32(i32(in_vertex_index & 1u) * 2 - 1);
-   return vec4<f32>(x, y, 0.0, 1.0);
+   return vec4<f32>(x * 0.9, y * 0.9, 0.0, 1.0);
}
```
`hello-triangle/main.rs`
```diff
             entry_point: "fs_main",
             targets: &[Some(swapchain_format.into())],
         }),
-        primitive: wgpu::PrimitiveState::default(),
+        primitive: wgpu::PrimitiveState {
+            topology: wgpu::PrimitiveTopology::PointList,
+            ..wgpu::PrimitiveState::default()
+        },
         depth_stencil: None,
         multisample: wgpu::MultisampleState::default(),
         multiview: None,
```
Then running `cargo run-wasm --example hello-triangle --features webgl`

**Expected vs observed behavior**
*Expected behaviour:*
Three pixel-sized red dots on screen.
This can also be seen working by running the above code with `cargo run --example hello-triangle`
*Observed behaviour:*
No dots are rendered

**Platform**
Windows 10, wgpu 0.14.0, tested on both Firefox and Chrome.

PR #3440, opened 2023-01-31
[gles] fix: Set `FORCE_POINT_SIZE` if it is vertex shader with mesh consist of point list

REASY:
# fix: Set `FORCE_POINT_SIZE` if it is vertex shader with mesh consist of point list
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Fixes https://github.com/gfx-rs/wgpu/issues/3179
Blocked by https://github.com/gfx-rs/naga/pull/2223

**Description**
PrimitiveTopology::PointList does not render on WebGL on Windows OS.

It is because gl_PointSize is not set. According to https://registry.khronos.org/OpenGL/specs/es/3.2/GLSL_ES_Specification_3.20.html#built-in-language-variables
> The variable gl_PointSize is intended for a shader to write the size of the point to be rasterized. It is measured in pixels. If gl_PointSize is not written to, its value is _undefined_ in subsequent pipe stages.

**Testing**
I'm pointing WGPU to use fixed version of naga from https://github.com/gfx-rs/naga/pull/2223. Then I run hello-triangle 
```bash
cargo run-wasm --example hello-triangle --features webgl --port 18000
```
And expect to see 64 red pixels as in the screenshot below:
image

### DIFF
--- a/wgpu-hal/src/gles/device.rs
+++ b/wgpu-hal/src/gles/device.rs
@@ -331,6 +331,7 @@ impl super::Device {
         Ok(program)
     }
 
+    #[allow(clippy::too_many_arguments)]
     unsafe fn create_program<'a>(
         gl: &glow::Context,
         shaders: ArrayVec<ShaderStage<'a>, 3>,
--- a/wgpu-hal/src/gles/device.rs
+++ b/wgpu-hal/src/gles/device.rs
@@ -192,8 +192,10 @@ impl super::Device {
         naga_stage: naga::ShaderStage,
         stage: &crate::ProgrammableStage<super::Api>,
         context: CompilationContext,
+        primitive_topology: Option<wgt::PrimitiveTopology>,
     ) -> Result<glow::Shader, crate::PipelineError> {
         use naga::back::glsl;
+
         let pipeline_options = glsl::PipelineOptions {
             shader_stage: naga_stage,
             entry_point: stage.entry_point.to_string(),
@@ -225,12 +227,31 @@ impl super::Device {
             binding_array: BoundsCheckPolicy::Unchecked,
         };
 
+        let update_naga_options = primitive_topology == Some(wgt::PrimitiveTopology::PointList)
+            && naga_stage == naga::ShaderStage::Vertex;
+        // Update naga options if mesh consist of point list and it is vertex shader
+        let naga_options = if update_naga_options {
+            let mut wrt_flags = context.layout.naga_options.writer_flags;
+            wrt_flags.set(glsl::WriterFlags::FORCE_POINT_SIZE, true);
+            glsl::Options {
+                version: context.layout.naga_options.version,
+                writer_flags: wrt_flags,
+                binding_map: context.layout.naga_options.binding_map.clone(),
+                zero_initialize_workgroup_memory: context
+                    .layout
+                    .naga_options
+                    .zero_initialize_workgroup_memory,
+            }
+        } else {
+            context.layout.naga_options.clone()
+        };
+
         let mut output = String::new();
         let mut writer = glsl::Writer::new(
             &mut output,
             &shader.module,
             &shader.info,
-            &context.layout.naga_options,
+            &naga_options,
             &pipeline_options,
             policies,
         )
@@ -262,6 +283,7 @@ impl super::Device {
         layout: &super::PipelineLayout,
         #[cfg_attr(target_arch = "wasm32", allow(unused))] label: Option<&str>,
         multiview: Option<std::num::NonZeroU32>,
+        primitive_topology: Option<wgt::PrimitiveTopology>,
     ) -> Result<Arc<super::PipelineInner>, crate::PipelineError> {
         let mut program_stages = ArrayVec::new();
         let mut group_to_binding_to_slot = Vec::with_capacity(layout.group_infos.len());
@@ -300,6 +322,7 @@ impl super::Device {
                     multiview,
                     glsl_version,
                     self.shared.private_caps,
+                    primitive_topology,
                 )
             })
             .to_owned()?;
@@ -316,6 +339,7 @@ impl super::Device {
         multiview: Option<std::num::NonZeroU32>,
         glsl_version: u16,
         private_caps: super::PrivateCapabilities,
+        primitive_topology: Option<wgt::PrimitiveTopology>,
     ) -> Result<Arc<super::PipelineInner>, crate::PipelineError> {
         let program = unsafe { gl.create_program() }.unwrap();
         #[cfg(not(target_arch = "wasm32"))]
@@ -340,7 +364,7 @@ impl super::Device {
                 multiview,
             };
 
-            let shader = Self::create_shader(gl, naga_stage, stage, context)?;
+            let shader = Self::create_shader(gl, naga_stage, stage, context, primitive_topology)?;
             shaders_to_delete.push(shader);
         }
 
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -41,9 +41,8 @@ path = "./wgpu-hal"
 version = "0.15"
 
 [workspace.dependencies.naga]
-git = "https://github.com/gfx-rs/naga"
-rev = "f0edae8"
-version = "0.11"
```
cwfitzgerald, line 46: When you update naga after the naga PR is merged, leave the version number in place

wgpu-hal/src/gles/device.rs
```diff
@@ -237,12 +239,32 @@ impl super::Device {
             binding_array: BoundsCheckPolicy::Unchecked,
         };
 
+        // Update naga options if mesh consist of point list and it is vertex shader
+        let maybe_update_naga_options = if Some(wgt::PrimitiveTopology::PointList)
+            == primitive_topology
+            && naga_stage == naga::ShaderStage::Vertex
```
cwfitzgerald, line 245: Let's move the conditional to it's own boolean variable called `update_naga_options` and the actual options to just `naga_options`

wgpu/examples/hello-triangle/main.rs
```diff
@@ -64,7 +64,10 @@ async fn run(event_loop: EventLoop<()>, window: Window) {
             entry_point: "fs_main",
             targets: &[Some(swapchain_format.into())],
         }),
-        primitive: wgpu::PrimitiveState::default(),
+        primitive: wgpu::PrimitiveState {
+            topology: wgpu::PrimitiveTopology::PointList,
+            ..wgpu::PrimitiveState::default()
+        },
```
cwfitzgerald, line 70: Obviously for testing, just flagging so it isn't forgotten

cwfitzgerald (dismissed):

Cargo.toml
teoxoy: We landed https://github.com/gfx-rs/wgpu/pull/3451, if you rebase you won't need to change the `toml`s.

wgpu-hal/src/gles/device.rs
```diff
@@ -237,12 +239,31 @@ impl super::Device {
...
+                writer_flags: wrt_flags,
+                binding_map: context.layout.naga_options.binding_map.clone(),
+                zero_initialize_workgroup_memory: context
+                    .layout
+                    .naga_options
+                    .zero_initialize_workgroup_memory,
+            }
+        } else {
+            context.layout.naga_options.clone()
+        };
```
teoxoy, line 259: Looking at the Vulkan backend,

https://github.com/gfx-rs/wgpu/blob/48d8666de002c7bfddbe85a2596f7748b4cc18b4/wgpu-hal/src/vulkan/adapter.rs#L1256-L1262

I'm not sure what the best solution is but cloning the options here doesn't sound too good to me.

Are there any issues if we always set `FORCE_POINT_SIZE` to true?

REASY: @teoxoy I don't think I understand the domain well enough to comment.

@cwfitzgerald what do you think?

REASY: I could rewrite to something like this (extracting to reusable method):
```rust
let update_naga_options = primitive_topology == Some(wgt::PrimitiveTopology::PointList)
    && naga_stage == naga::ShaderStage::Vertex;
// Update naga options if mesh consist of point list and it is vertex shader
let reflection_info = if update_naga_options {
    let mut wrt_flags = context.layout.naga_options.writer_flags;
    wrt_flags.set(glsl::WriterFlags::FORCE_POINT_SIZE, true);
    let options = glsl::Options {
        version: context.layout.naga_options.version,
        writer_flags: wrt_flags,
        binding_map: context.layout.naga_options.binding_map.clone(),
        zero_initialize_workgroup_memory: context
            .layout
            .naga_options
            .zero_initialize_workgroup_memory,
    };
    let mut writer = glsl::Writer::new(
        &mut output,
        &shader.module,
        &shader.info,
        &options,
        &pipeline_options,
        policies,
    )
    .map_err(|e| {
        let msg = format!("{e}");
        crate::PipelineError::Linkage(map_naga_stage(naga_stage), msg)
    })?;
    writer.write().map_err(|e| {
        let msg = format!("{e}");
        crate::PipelineError::Linkage(map_naga_stage(naga_stage), msg)
    })
} else {
    let mut writer = glsl::Writer::new(
        &mut output,
        &shader.module,
        &shader.info,
        &context.layout.naga_options,
        &pipeline_options,
        policies,
    )
    .map_err(|e| {
        let msg = format!("{e}");
        crate::PipelineError::Linkage(map_naga_stage(naga_stage), msg)
    })?;
    writer.write().map_err(|e| {
        let msg = format!("{e}");
        crate::PipelineError::Linkage(map_naga_stage(naga_stage), msg)
    })
}?;
```

REASY: @teoxoy both `Options` and `PipelineOptions` are from naga. It requires larger refactoring. 

Originally I implemented the parameter not as part of `WriterFlags`, but `PipelineOptions` (https://github.com/gfx-rs/naga/commit/4df4161d8d00c82910cb9cbdb8599abd1f82af46#diff-cc33bd3c8c18000d3ee7b6f5578e8fa07ffd08d87b0aa3d9525a107937fb3d3fR264). I followed how it was implemented for Metal [metal/device.rs#L71](https://github.com/gfx-rs/wgpu/blob/master/wgpu-hal/src/metal/device.rs#L71) [back/msl/mod.rs#L238 ](https://github.com/gfx-rs/naga/blob/master/src/back/msl/mod.rs#L238)

The same functionality is implemented in two ways.

teoxoy: I don't think it's the same functionality, for metal it works like a filter and is needed because you'll get an error otherwise.

For Vulkan, we can always force naga to write it and will be ignored by the driver if it's not a point list primitive.

I think we can have gles behave the same as Vulkan here. Let's see what @cwfitzgerald thinks as well.

teoxoy: I think why we need to have `allow_point_size` for metal is due to:

> glslang generates gl_PerVertex struct with gl_CullDistance, gl_ClipDistance and gl_PointSize builtin inside by default

(from https://github.com/gfx-rs/naga/blob/37be4df2b8e7c56b0390445f4457955211040433/src/back/wgsl/writer.rs#L1161-L1169)

And if we always set `FORCE_POINT_SIZE` to true, we'll essentially behave the same as glslang.

wgpu-hal/src/gles/device.rs
```diff
@@ -308,6 +331,7 @@ impl super::Device {
         Ok(program)
     }
 
+    #[allow(clippy::too_many_arguments)]
```
REASY, line 334: Otherwise getting
```
error: this function has too many arguments (8/7)
```
https://github.com/gfx-rs/wgpu/actions/runs/4083942757/jobs/7147934141

wgpu-hal/src/gles/device.rs
REASY: As per discussion, I'm setting `FORCE_POINT_SIZE` to make it consistent with the Vulkan backend.

teoxoy (approved):
LGTM, thank you!

PR #3447, opened 2023-02-02
Update gpu allocator to 0.22

Elabajaba:
**Checklist**

- [x] Run `cargo clippy`.

**Description**
Minimal update to gpu allocator 0.22.

**Testing**
Tested water, shadow, and compute example. Also ran `cargo nextest` and everything (except the `zero_init_workgroup_mem` test that was already broken) passed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1150,9 +1150,9 @@ dependencies = [
 
 [[package]]
 name = "gpu-allocator"
-version = "0.21.0"
+version = "0.22.0"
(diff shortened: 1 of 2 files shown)


Elabajaba:
We could potentially use gpu-allocator's new `allocator.create_resource(...)` function, but I'm not sure what the benefit of using it over what we're already doing would be.

cwfitzgerald (approved):

PR #3454, opened 2023-02-03
Add MULTISAMPLE_X16 texture format feature flag

Dinnerbone:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
None

**Description**
MSAA x16 is at the limit of what most hardware & backends support, and it's not nearly as common as x4 or x8, but it *is* sometimes supported. We'd like to be able to see and use it when possible, as some of our content unfortunately requests it. This PR adds the flag so we can see when x16 is supported.

**Testing**
Manually tested, and all existing tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,10 @@ Bottom level categories:
 
 ### Changes
 
+#### General
+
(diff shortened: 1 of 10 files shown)


teoxoy (approved):
Solid work!

PR #3462, opened 2023-02-09
Build for WASM on docs.rs

daxpedda:
I noticed that on docs.rs the WASM target isn't available, which made it hard to search for some functions.

This PR explicitly defines which targets to build, I threw out the 32-bit targets and added `wasm32-unknown-unknown`.
It might not be useful to build all three main targets, Linux, MacOS and Windows for all crates, any feedback is welcome.

@cwfitzgerald it would be nice to get this in before v0.15.1 (https://github.com/gfx-rs/wgpu/pull/3461).


cwfitzgerald (approved):
Yeet, will get this in the patch

PR #3466, opened 2023-02-10
Make crate features no-op on incompatible targets

### LINKED ISSUE #3463: WASM and MacOS target not built on docs.rs
I believed that #3462 would add the `wasm32-unknown-unknown` successfully to docs.rs builds, but it didn't.

See https://docs.rs/wgpu/0.15.1/wgpu/.

Trying to dig into it further, I noticed that previous versions didn't succeed in building `x86_64-apple-darwin` either.
Unfortunately https://docs.rs/crate/wgpu/0.15.1/builds/740775 doesn't provide any further help.

### LINKED ISSUE #3465: Make `wgpu-hal` crate features target aware
I noticed that the way `wgpu-hal`s crate features are designed is target independent.

The problem is of course that using `--all-features` or any other non-compatible crate feature will break compilation. See #3463 for example.

I propose that activating crate features on `wgpu-hal` that only work on specific targets should be no-op on non-compatible targets.
For example activating the `vulkan` crate feature on `wasm32-unknown-unknown` should do nothing.

Additionally I think removing https://github.com/gfx-rs/wgpu/blob/5b8c55c4510fc58aa2b6baf03e31ed61d640b866/wgpu-hal/src/lib.rs#L54-L61 would also remove some unnecessary barriers.

I think the only downside this change would introduce is that you could have activated crate features that you don't really need for your target, which shouldn't really have any downside if they are no-op.

daxpedda:
This changes crate features to be no-op on incompatible targets instead of not compiling.

- Vulkan is no-op on WASM
- DirectX12 and DirectX11 is no-op on everything but Windows
- Metal is no-op on everything but MacOS and iOS

As far as I'm aware Vulkan is needed on MacOS for Vulkan Portability.
I saw a feature called `portable_features` that is used for testing purposes. I believe I can actually remove this now?

~~Depends on #3467.~~

Fixes #3463 (only WASM).
Fixes #3465.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,12 +125,16 @@ jobs:
 
           # build for WebGPU
           cargo clippy --target ${{ matrix.target }} -p wgpu --tests --features glsl,spirv
+          cargo doc --target ${{ matrix.target }} -p wgpu --no-deps --features glsl,spirv
 
           # build for WebGL
           cargo clippy --target ${{ matrix.target }} -p wgpu --tests --features webgl,glsl,spirv
+          cargo doc --target ${{ matrix.target }} -p wgpu --no-deps --features webgl,glsl,spirv
 
-          # build docs
-          cargo doc --target ${{ matrix.target }} -p wgpu --no-deps --features glsl,spirv
+          # all features
+          cargo clippy --target ${{ matrix.target }} -p wgpu-hal --tests --all-features
+          cargo clippy --target ${{ matrix.target }} -p wgpu --tests --all-features
+          cargo doc --target ${{ matrix.target }} -p wgpu --no-deps --all-features
 
       - name: check em
         if: matrix.kind == 'em'
@@ -138,7 +142,7 @@ jobs:
         run: |
           set -e
 
-          # build for Emscripten/WebGL
+          # build for Emscripten
           cargo clippy --target ${{ matrix.target }} -p wgpu -p wgpu-hal --no-default-features
 
           # build cube example
@@ -146,6 +150,10 @@ jobs:
 
           # build raw-gles example
           cargo clippy --target ${{ matrix.target }} --example raw-gles
+
+          # all features
+          cargo clippy --target ${{ matrix.target }} -p wgpu-hal --all-features
+          cargo clippy --target ${{ matrix.target }} -p wgpu --all-features
 
       - name: check native
         if: matrix.kind == 'native'
@@ -159,8 +167,9 @@ jobs:
           # Check with all features.
           # (But watch out for backend-selection features in wgpu-core; some of
           # those only build on the right platforms.)
-          cargo clippy --target ${{ matrix.target }} -p wgpu -p wgpu-info -p player --tests --all-features
+          cargo clippy --target ${{ matrix.target }} -p wgpu-hal --tests --all-features
           cargo clippy --target ${{ matrix.target }} -p wgpu-core --tests --all-features
+          cargo clippy --target ${{ matrix.target }} -p wgpu -p wgpu-info -p player --tests --all-features
 
           # build docs
           # (Watch out for backend-selection features in wgpu-core; some of
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,13 +127,8 @@ jobs:
           cargo clippy --target ${{ matrix.target }} -p wgpu --tests --features glsl,spirv
           cargo doc --target ${{ matrix.target }} -p wgpu --no-deps --features glsl,spirv
 
-          # build for WebGL
-          cargo clippy --target ${{ matrix.target }} -p wgpu --tests --features webgl,glsl,spirv
-          cargo doc --target ${{ matrix.target }} -p wgpu --no-deps --features webgl,glsl,spirv
-
           # all features
-          cargo clippy --target ${{ matrix.target }} -p wgpu-hal --tests --all-features
-          cargo clippy --target ${{ matrix.target }} -p wgpu --tests --all-features
+          cargo clippy --target ${{ matrix.target }} -p wgpu -p wgpu-hal --tests --all-features
           cargo doc --target ${{ matrix.target }} -p wgpu --no-deps --all-features
 
       - name: check em
@@ -165,17 +160,10 @@ jobs:
           cargo clippy --target ${{ matrix.target }} -p wgpu -p wgpu-core -p wgpu-info -p player --no-default-features
 
           # Check with all features.
-          # (But watch out for backend-selection features in wgpu-core; some of
-          # those only build on the right platforms.)
-          cargo clippy --target ${{ matrix.target }} -p wgpu-hal --tests --all-features
-          cargo clippy --target ${{ matrix.target }} -p wgpu-core --tests --all-features
-          cargo clippy --target ${{ matrix.target }} -p wgpu -p wgpu-info -p player --tests --all-features
+          cargo clippy --target ${{ matrix.target }} -p wgpu -p wgpu-info -p player -p wgpu-core -p wgpu-hal --tests --all-features
 
           # build docs
-          # (Watch out for backend-selection features in wgpu-core; some of
-          # those only build on the right platforms.)
-          cargo doc --target ${{ matrix.target }} -p wgpu -p wgpu-info -p player --all-features --no-deps
-          cargo doc --target ${{ matrix.target }} -p wgpu-core --no-deps --all-features
+          cargo doc --target ${{ matrix.target }} -p wgpu -p wgpu-info -p player -p wgpu-core -p wgpu-hal --all-features --no-deps
 
   wasm-test:
     name: Test WebAssembly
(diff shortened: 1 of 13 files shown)


daxpedda:
@grovesNL if you can take a look and see if and how you are affected by this.
Any feedback is welcome!

daxpedda:
What exactly is the point of the `emscripten` crate feature? Couldn't it be completely replaced by `target_os = "emscripten"`?
This would also make it not necessary to activate the feature for downstream users.

grovesNL:
Nice! I thought there was some reason we weren't able to set it up this way originally (maybe @kvark remembers?). Obviously it makes the `cfg`s slightly messier but it doesn't seem like a big deal since there shouldn't be very many of them anyway.

> What exactly is the point of the emscripten crate feature? Couldn't it be completely replaced by target_os = "emscripten"?
This would also make it not necessary to activate the feature for downstream users.

That sounds right - if we don't need a separate feature to conditionally enable it then I think could be replaced by `target_os = "emscripten"`

daxpedda:
> That sounds right - if we don't need a separate feature to conditionally enable it then I think could be replaced by `target_os = "emscripten"`

I will do this in a separate PR.

teoxoy:
> Fixes https://github.com/gfx-rs/wgpu/issues/3463 (only WASM).

Shouldn't it also fix metal?

> I saw a feature called `portable_features` that is used for testing purposes. I believe I can actually remove this now?

I think so.

teoxoy (approved):
LGTM, thanks for the PR!

We seemed to be moving in this direction already, some `cfg`s had the `target_os` (ex `windows` for `dx11`/`dx12`) and also had some dependencies gated by `target_os`.

This is also more in line with the notion that feature flags should be additive and not cause compilation errors.

teoxoy:
Could you remove `portable_features` in this PR as well?

daxpedda:
> > Fixes #3463 (only WASM).
> 
> Shouldn't it also fix metal?

As far as I am aware it will not, currently `metal` can't be built on Linux.
> Could you remove `portable_features` in this PR as well?

Will do.

daxpedda:
I can't remove `portable_features` right now because I have to put in something instead in the CI, considering this PR I thought it would be appropriate to put in `--all-features` instead, but that still doesn't work because of `emscripten`, which should be fixed in #3467.

So either we remove `portable_features` in a follow-up PR, or I just make this PR depend on #3467.

I tested my assumption in #3488, looks like it works.

cwfitzgerald:
Let's make the PRs dependent, emscripten should go in soon.

daxpedda:
Done!

daxpedda:
This is ready now!

cwfitzgerald (approved):
This is awesome to finally have!

Does this mean https://github.com/gfx-rs/wgpu/blob/master/wgpu/Cargo.toml#L106-L109 we can enable all features on all platforms?

daxpedda:
> Does this mean https://github.com/gfx-rs/wgpu/blob/master/wgpu/Cargo.toml#L106-L109 we can enable all features on all platforms?

Yes!
And as far as I checked the CI again now this is tested for native.

daxpedda:
I was just going through the CI again and I think we might want to test for `--all-features` a bit more.
Specifically `web` isn't tested with `--all-features` and `wgpu-hal` should maybe be tested separately just to make sure everything works as intended. Should I add this here or should I make a new PR?

cwfitzgerald:
Let's fix CI in this PR - it's all entangled

.github/workflows/ci.yml
```diff
@@ -159,14 +167,15 @@ jobs:
           # Check with all features.
           # (But watch out for backend-selection features in wgpu-core; some of
           # those only build on the right platforms.)
```
cwfitzgerald, line 169: Comment out of date

.github/workflows/ci.yml
```diff
@@ -159,14 +167,15 @@ jobs:
...
+          cargo clippy --target ${{ matrix.target }} -p wgpu-hal --tests --all-features
+          cargo clippy --target ${{ matrix.target }} -p wgpu-core --tests --all-features
           cargo clippy --target ${{ matrix.target }} -p wgpu -p wgpu-info -p player --tests --all-features
```
cwfitzgerald, line 172: I think this can now be one call to clippy with all the crates, as --all-features works - that's the only reason we didn't do it all in one command before

.github/workflows/ci.yml
```diff
@@ -159,14 +167,15 @@ jobs:
...
-          cargo clippy --target ${{ matrix.target }} -p wgpu-core --tests --features="portable_features"
 
           # build docs
           # (Watch out for backend-selection features in wgpu-core; some of
           # those only build on the right platforms.)
           cargo doc --target ${{ matrix.target }} -p wgpu -p wgpu-info -p player --all-features --no-deps
-          cargo doc --target ${{ matrix.target }} -p wgpu-core --no-deps --features="portable_features"
+          cargo doc --target ${{ matrix.target }} -p wgpu-core --no-deps --all-features
```
cwfitzgerald, line 178: Same here

cwfitzgerald (commented):
Some comments, but looking good

.github/workflows/ci.yml
daxpedda: Done.

.github/workflows/ci.yml
daxpedda: Done.

daxpedda: Now that i think about it, I could also just use `--workspace`. Wdyt?

cwfitzgerald: Now that it should work, let's do it

daxpedda: `--workspace` fails to compile something with the `v8` dependency, which is something in Deno I assume, for Android and iOS.
https://github.com/gfx-rs/wgpu/actions/runs/4218266653/jobs/7322645792
https://github.com/gfx-rs/wgpu/actions/runs/4218266653/jobs/7322645995

Reverting again.

daxpedda: Done.

wgpu-hal/src/dx12/mod.rs
```diff
@@ -27,7 +27,7 @@ and [`crate::CommandEncoder::set_render_pipeline`] with
 [`crate::CommandEncoder::set_compute_pipeline`].
 
 For this reason, in order avoid repeating the binding code,
-we are binding everything in [`CommandEncoder::update_root_elements`].
+we are binding everything in `CommandEncoder::update_root_elements`.
```
daxpedda, line 30: This type is private, so rustdoc failed on this.

.github/workflows/ci.yml
```diff
@@ -125,20 +125,24 @@ jobs:
...
 
           # build for WebGL
           cargo clippy --target ${{ matrix.target }} -p wgpu --tests --features webgl,glsl,spirv
+          cargo doc --target ${{ matrix.target }} -p wgpu --no-deps --features webgl,glsl,spirv
 
-          # build docs
-          cargo doc --target ${{ matrix.target }} -p wgpu --no-deps --features glsl,spirv
+          # all features
+          cargo clippy --target ${{ matrix.target }} -p wgpu-hal --tests --all-features
+          cargo clippy --target ${{ matrix.target }} -p wgpu --tests --all-features
```
cwfitzgerald, line 136: Why do these need to be separate

.github/workflows/ci.yml
```diff
@@ -125,20 +125,24 @@ jobs:
 
           # build for WebGPU
           cargo clippy --target ${{ matrix.target }} -p wgpu --tests --features glsl,spirv
+          cargo doc --target ${{ matrix.target }} -p wgpu --no-deps --features glsl,spirv
 
           # build for WebGL
           cargo clippy --target ${{ matrix.target }} -p wgpu --tests --features webgl,glsl,spirv
+          cargo doc --target ${{ matrix.target }} -p wgpu --no-deps --features webgl,glsl,spirv
```
cwfitzgerald, line 132: This is redundant with the all features

.github/workflows/ci.yml
```diff
@@ -157,16 +165,12 @@ jobs:
...
 
           # Check with all features.
-          # (But watch out for backend-selection features in wgpu-core; some of
-          # those only build on the right platforms.)
-          cargo clippy --target ${{ matrix.target }} -p wgpu -p wgpu-info -p player --tests --all-features
-          cargo clippy --target ${{ matrix.target }} -p wgpu-core --tests --features="portable_features"
+          cargo clippy --target ${{ matrix.target }} -p wgpu -p wgpu-info -p player -p wgpu-core -p wgpu-hal --tests --all-features
 
           # build docs
           # (Watch out for backend-selection features in wgpu-core; some of
```
cwfitzgerald, line 171: Similarly redundant comment

cwfitzgerald (dismissed):

.github/workflows/ci.yml
daxpedda: Removed.

.github/workflows/ci.yml
daxpedda: Merged.

.github/workflows/ci.yml
daxpedda: Comment removed.

daxpedda:
Just rebased on top of #3505 and #3511.

PR #3468, opened 2023-02-10
Add Dependabot

daxpedda:
I noticed a bunch of dependencies are out-of-date.

Also noticed that `Cargo.lock` was part of the repo, not entirely sure how Dependabot will behave in this case, [the documentation](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy) wasn't entirely clear to me.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+
+updates:
+  - package-ecosystem: cargo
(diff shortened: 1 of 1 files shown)


cwfitzgerald (approved):
Dope! keep forgetting to set this up!

Issue #3430, opened 2023-01-27
Dynamic dispatch layer breaks on web


i509VCB:
**Description**
The web backend currently has an error like this when running the `hello-triangle` example using `cargo run-wasm --example=hello-triangle` in the browser console output:
```
panicked at 'assertion failed: wasm.is_instance_of::<T>()', wgpu/src/backend/web.rs:53:9

Stack:

getImports/imports.wbg.__wbg_new_abda76e883ba8a5f@http://localhost:8000/hello-triangle.js:340:21
console_error_panic_hook::Error::new::h09e78aee7fd991ae@http://localhost:8000/hello-triangle_bg.wasm:wasm-function[4182]:0xf15ec
console_error_panic_hook::hook_impl::h81c8bf5494034336@http://localhost:8000/hello-triangle_bg.wasm:wasm-function[717]:0x66f7d
console_error_panic_hook::hook::hc9dd7d1d176aa725@http://localhost:8000/hello-triangle_bg.wasm:wasm-function[4719]:0xf8b1e
core::ops::function::Fn::call::hc3d05c66054772b6@http://localhost:8000/hello-triangle_bg.wasm:wasm-function[3865]:0xec751
std::panicking::rust_panic_with_hook::hb42d415afcc11f2f@http://localhost:8000/hello-triangle_bg.wasm:wasm-function[1721]:0xaeb05
std::panicking::begin_panic_handler::{{closure}}::h57dd84c078404aa2@http://localhost:8000/hello-triangle_bg.wasm:wasm-function[2168]:0xc3482
std::sys_common::backtrace::__rust_end_short_backtrace::hef94dcc71d3b37f5@http://localhost:8000/hello-triangle_bg.wasm:wasm-function[5052]:0xfcef3
rust_begin_unwind@http://localhost:8000/hello-triangle_bg.wasm:wasm-function[3238]:0xe0f9f
core::panicking::panic_fmt::h32de9c76c9d5eb0c@http://localhost:8000/hello-triangle_bg.wasm:wasm-function[4581]:0xf6f1d
core::panicking::panic::hff08c7cf90edd9fe@http://localhost:8000/hello-triangle_bg.wasm:wasm-function[3862]:0xec692
<wgpu::backend::web::Identified<T> as core::convert::From<wgpu::context::ObjectId>>::from::h09e99484006351c1@http://localhost:8000/hello-triangle_bg.wasm:wasm-function[1664]:0xaba02
[...]

PR #3489, opened 2023-02-15
Implement resolve query set (#3489

JolifantoBambla:
**Checklist**

- [x] Run `cargo clippy`.

**Description**
I've added the missing `BufferUsages::QUERY_RESOLVE` (https://www.w3.org/TR/webgpu/#buffer-usage) and implemented `command_encoder_resolve_query_set` for the WebGPU backend.

**Testing**
Unfortunately, the [WebGPU backend is broken in v0.15.1](https://github.com/gfx-rs/wgpu/issues/3430), so I only tested this with a personal project and [my own patch for v0.14.2](https://github.com/JolifantoBambla/wgpu/tree/implement-timestamp-query-on-v0.14.2).


cwfitzgerald:
This is a good start, but I can merge this just yet - because this brings us in line with the upstream api, we also need the native api to behave the same. Once this merges, you'd need two different sets of flags to do query resolves.

This should also change the validation of the resolve query to check for the RESOLVE_QUERY usage. If anything is confusing about that transformation, please feel free to ask questions over on the matrix.

cwfitzgerald (changes requested):
As this is a breaking change, could you add an entry in the "major changes" explaining what changed? After that, this looks g2g

JolifantoBambla:
done :)
I also fixed the mipmap example - missed that one before
I don't know why the shader test (`shader::zero_init_workgroup_mem::zero_init_workgroup_mem`) fails though

cwfitzgerald (approved):
Thanks!

PR #3467, opened 2023-02-10
Remove `emscripten` crate features

daxpedda:
This removes the `emscripten` crate feature from all crates and replaces necessary guards with `target_os = "emscripten"`.
`wgpu` will activate `wgpu-hal/gles` by default for the emscripten target and guards that were in place for `feature = "webgl"` for Emscripten were adjusted.

Requires careful review, it was a lot of search & replace and copy & paste.

See #3466.

PR #3490, opened 2023-02-15
Enable more `wgpu-hal` functions for `webgl`

daxpedda:
As discussed in https://github.com/gfx-rs/wgpu/pull/3467#discussion_r1107748845.

Apparently there were some functions related to `wgpu-hal` that were not exposed in WebGL but should. I'm surprised that `enumerate_adapters` is available in `webgl` for example, so please review this carefully.

Wasn't sure about `map_texture_tagged_copy_view`, so I left it alone, happy to fix it too if somebody assures me it's correct.

### DIFF
--- a/wgpu/src/backend/direct.rs
+++ b/wgpu/src/backend/direct.rs
@@ -68,13 +68,11 @@ impl Context {
         &self.0
     }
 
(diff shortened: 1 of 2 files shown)


cwfitzgerald (approved):
Yup, that's all right!

`enumerate_adapters` exists, it just always returns a single adapter from the GL backend.

PR #3494, opened 2023-02-17
Set the new naga capabilities

teoxoy:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

**Description**
- set the right naga capabilities
- add `Features::SHADER_EARLY_DEPTH_TEST`
- set the `MULTISAMPLED_SHADING` downlevel flag for gles and dx11

**Testing**

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -681,6 +681,13 @@ bitflags::bitflags! {
         ///
         /// This is a native-only feature.
         const SHADER_INT16 = 1 << 42;
+        /// Allows shaders to use the `early_depth_test` attribute.
+        ///
+        /// Supported platforms:
+        /// - GLES 3.1+
+        ///
+        /// This is a native-only feature.
+        const SHADER_EARLY_DEPTH_TEST = 1 << 43;
     }
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,7 @@ By @teoxoy in [#3436](https://github.com/gfx-rs/wgpu/pull/3436)
 
 - Added `TextureFormatFeatureFlags::MULTISAMPLE_X16`. By @Dinnerbone in [#3454](https://github.com/gfx-rs/wgpu/pull/3454)
 - Support stencil-only views and copying to/from combined depth-stencil textures. By @teoxoy in [#3436](https://github.com/gfx-rs/wgpu/pull/3436)
+- Added `Features::SHADER_EARLY_DEPTH_TEST`. By @teoxoy in [#3494](https://github.com/gfx-rs/wgpu/pull/3494)
 
 #### WebGPU
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1515,7 +1515,7 @@ dependencies = [
 [[package]]
 name = "naga"
 version = "0.11.0"
-source = "git+https://github.com/gfx-rs/naga?rev=568d7c4c136dada369ef7f59ee8414a263d6c7b2#568d7c4c136dada369ef7f59ee8414a263d6c7b2"
+source = "git+https://github.com/gfx-rs/naga?rev=7422ace934b7ed51c1e061eb1d36de8259b5f619#7422ace934b7ed51c1e061eb1d36de8259b5f619"
 dependencies = [
  "bit-set",
  "bitflags",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ version = "0.15"
 
 [workspace.dependencies.naga]
 git = "https://github.com/gfx-rs/naga"
-rev = "568d7c4c136dada369ef7f59ee8414a263d6c7b2"
+rev = "7422ace934b7ed51c1e061eb1d36de8259b5f619"
 version = "0.11.0"
 
 [workspace.dependencies]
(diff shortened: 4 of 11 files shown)


wgpu-types/src/lib.rs
```diff
@@ -681,6 +681,13 @@ bitflags::bitflags! {
         ///
         /// This is a native-only feature.
         const SHADER_INT16 = 1 << 42;
+        /// Allows shaders to use the `early_depth_test` attribute.
+        ///
+        /// Supported platforms:
+        /// - GLES 3.1+
```
cwfitzgerald, line 687: This support table should be significantly wider

cwfitzgerald: And vk/d3d11 need this feature added

cwfitzgerald (changes requested):

wgpu-types/src/lib.rs
teoxoy: Indeed, it's just that we only have it implemented for the gles backend in naga.

cwfitzgerald: Wait how does the water example work then?

teoxoy: No idea 😅

cwfitzgerald: As this is going to make the water example not run anymore on most platforms, let's remove the early depth test attribute from it - as it was likely ignored anyway

cwfitzgerald (approved):

maaku:
I'm going through the process of upgrading to wgpu 0.16, and this PR has broken my application.  The following excerpt of my fragment shader for imposter-based drawing of atoms (spheres) has worked in all previous version of wgpu, on all platforms I have access to test (Metal on macOS, Vulkan on Linux, DX12 on Windows, WebGL2 for web):

```
    const float dist = length(uv);
    if (dist > element.radius) {
        discard;
    }

    const float z = sqrt(element.radius * element.radius - dist * dist);
    const vec4 in_pos_clipspace = position_clip_space + camera.projection[2] * z;

    gl_FragDepth = in_pos_clipspace.z / in_pos_clipspace.w;
```

Why is this feature gated for GLES 3.1+ only? As noted in the code review above, it works perfectly fine on other platforms.

Wumpf:
@maaku what error are you getting? The snippet seems unrelated to both multisampled shading and early depth test

maaku:
```
[2023-06-29T10:32:00Z ERROR wgpu::backend::direct] Handling wgpu errors as fatal by default

Caused by:
      note: label = `/Users/mark/Development/atomCAD/target/debug/build/atomcad-render-30b7e31c716879fe/out/shaders/billboard.frag`
    
Shader validation error: 


    Entry point main at Fragment is invalid
    Capability EARLY_DEPTH_TEST is not supported
```

Here is the whole fragment shader:

```
#version 450

struct Element {
    vec3 color;
    float radius;
};

layout(set = 0, binding = 0) uniform Camera {
    vec4 projection[4];
    vec4 view[4];
    vec4 projection_view[4];
} camera;

layout(location = 0) in vec2 uv;
layout(location = 1) in vec4 position_clip_space;
layout(location = 2) flat in vec4 element_vec;
layout(location = 4) flat in vec4 center_view_space;
layout(location = 5) in vec4 position_view_space;

layout(depth_greater) out float gl_FragDepth;
layout(location = 0) out vec4 color;
layout(location = 1) out vec4 normal;

float map(float value, float low1, float high1, float low2, float high2) {
    return low2 + (value - low1) * (high2 - low2) / (high1 - low1);
}

vec4 linear_to_srgb(vec4 input_color) {
    bvec3 cutoff = lessThan(input_color.rgb, vec3(0.0031308));
    vec3 higher = vec3(1.005) * pow(input_color.rgb, vec3(1.0 / 2.4)) - vec3(0.055);
    vec3 lower = input_color.rgb * vec3(12.92);

    return vec4(mix(higher, lower, cutoff), input_color.a);
}

void main(void) {
    Element element;
    element.color = element_vec.xyz;
    element.radius = element_vec.w;
    const float dist = length(uv);
    if (dist > element.radius) {
        discard;
    }

    const float z = sqrt(element.radius * element.radius - dist * dist);
    const vec4 in_pos_clipspace = position_clip_space + camera.projection[2] * z;

    gl_FragDepth = in_pos_clipspace.z / in_pos_clipspace.w;

    color = vec4(
        element.color * map(z, 0.0, element.radius, 0.25, 1.0),
        1.0
    );
    normal = vec4(normalize(position_view_space.xyz - center_view_space.xyz), 0.0);

#ifdef TARGET_WASM
    // Currently, firefox webgpu doesn't automatically convert linear rgb outputs to srgb
    // so we do it manually.
    color = linear_to_srgb(color);
#endif
}
```

teoxoy:
That shader doesn't seem to contain the `early_fragment_tests` qualifier, so I don't see how the validation would get triggered in this case.

Also, trying to validate the shader via naga cli, I'm getting:

```
error: Unexpected qualifier
   ┌─ test.frag:20:8
   │
20 │ layout(depth_greater) out float gl_FragDepth;
   │        ^^^^^^^^^^^^^
```

Which sounds right since we don't have conservative depth implemented in the GLSL frontend.

maaku:
Thank you. I think I misunderstood how early fragment tests worked. I thought that setting was required for what I was doing. Removing it fixes the problem and I was able to upgrade to 0.16 without any other major issues. Thank you.

PR #3502, opened 2023-02-20
Replace fxhash with rustc-hash

### LINKED ISSUE #3369: Both fxhash and rustc-hash are in the dependency tree
`naga` uses `rustc-hash` and `wgpu-hal` uses `fxhash`.  Both do the same thing, but both are being compiled separately. `rustc-hash` has hit 1.0, has been more recently updated, and is maintained by the rustlang devs themselves. It may be worth changing the `fxhash` dependency on `wgpu-hal` to `rustc-hash`.

james7132:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Fixes #3369.

**Description**
Changes all dependencies that currently depend on `fxhash` to use `rustc-hash` instead, which should be a drop-in replacement that is both maintained and hit 1.0. Brings it into line with `naga`, which is using `rustc-hash`.

**Testing**
All of the crates in the workspace build properly.


cwfitzgerald (approved):
Nice!

nical (approved):

cwfitzgerald:
Format mad

brody2consult:
Quoting a note I added to #6999 for the record on this update (apologies in advance for all the ~~notes~~ __spam__):

> * [Both fxhash and rustc-hash are in the dependency tree #3369](https://github.com/gfx-rs/wgpu/issues/3369) / PR [Replace fxhash with rustc-hash #3502](https://github.com/gfx-rs/wgpu/pull/3502) - it looks to me like the old `fxhash` crate was some kind of copy of the hasher from `rustc-hash`. I think it was ~100%~ **200%** correct to drop use of the old `fxhash` in favor of `rustc-hash`. My one issue with `rustc-hash` is that it seems to contain both `FxHasher`, which can be used with `hashbrown`, and `FxHashMap` / `FxHashSet` typedefs which are hard-coded to using `std::collections`. So moving forward, we should use `rustc_hash::FxHasher` but _not_ `rustc_hash::FxHashMap` or `rustc_hash::FxHashSet`.

PR #3504, opened 2023-02-20
Allow tests to be captured and replayed in xcode

nical:
**Checklist**

- [ ] Run `cargo clippy`. (I'd love to but I need https://github.com/gfx-rs/wgpu/pull/3505 for that to work)

**Description**

I recently had to debug a failing test on mac and being able to capture/replay a test run in xcode was very useful. It took some fiddling to get it to work, here is the patch that allowed me to do it.

Obviously calling into metal directly from the test framework isn't great. We would benefit a lot from this being lifted into wgpu proper with solutions for other backends. However I am going to be unavailable for a while starting soon so I wanted to put this here in a less than ideal but working state. It could either be merged as is or just exist here as a reference for anyone in need of troubleshooting a test failure on mac or to build a better solution upon.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3028,6 +3028,7 @@ dependencies = [
  "image",
  "js-sys",
  "log",
+ "metal",
  "naga",
  "nanorand",
  "noise",
--- a/wgpu/Cargo.toml
+++ b/wgpu/Cargo.toml
@@ -176,6 +176,9 @@ nanorand = { workspace = true, features = ["wyrand"] }
 wasm-bindgen-test.workspace = true
 winit.workspace = true # for "halmark" example
 
+[target.'cfg(any(target_os="macos", target_os="ios"))'.dev-dependencies]
+mtl = { package = "metal", version = "0.24.0" }
+
 [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
 async-executor.workspace = true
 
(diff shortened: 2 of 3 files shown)


nical:
If there is any prior discussion about a portable API for wgpu capture hints, I'd love to know, I'm sure I'll find myself needing this for vk/renderdoc soon enough.

teoxoy:
I haven't looked into it too much but we have a `renderdoc` feature already.
Maybe we can add support for `xcode` via the same api and rename the feature to something that would enable either `renderdoc` or `xcode` based on platform?

@cwfitzgerald thoughts on this?

nical:
Looking at renderdoc.rs, I think that we can do something along these lines:

```Rust
// Returns None when not supported
let mut capture: Option<wgpu::CaptureContext> = device.capture_context();
// ...

if let Some(ctx) = capture {
    ctx.start_capture();
}

// ...

if let Some(ctx) = capture {
    ctx.end_capture();
}
```

Looks like the equivalent APIs via DXGI or PIX would fit in as well https://devblogs.microsoft.com/pix/programmatic-capture/

cwfitzgerald:
We already do support xcode - https://github.com/gfx-rs/wgpu/blob/master/wgpu-hal/src/metal/device.rs#L1137-L1155

PIX support would be nice.

nical:
Oh I totally missed that this was already implemented and exposed 🤦🏻

PR #3505, opened 2023-02-20
Work around a clippy error

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Description**

I was hitting this error on mac when running clippy:

```
error: this public function might dereference a raw pointer but is not marked `unsafe`
   --> wgpu-core/src/instance.rs:502:81
    |
502 |                     inst.create_surface_from_layer(unsafe { std::mem::transmute(layer) })
    |                                                                                 ^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#not_unsafe_ptr_arg_deref
    = note: `#[deny(clippy::not_unsafe_ptr_arg_deref)]` on by default
```

The simplest fix is to mark the function as unsafe. Clippy was then warning that its only caller `create_surface_from_core_animation_layer`, while being unsafe itself, should still have unsafe blocks around whatever unsafe functions it calls, and that a public unsafe function should have a `# Safety` section in its doc, so I continued obeying the machine.

Clippy is quiet now.

### DIFF
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -485,8 +485,11 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         id.0
     }
 
+    /// # Safety
+    ///
+    /// `layer` must be a valid pointer.
     #[cfg(feature = "metal")]
-    pub fn instance_create_surface_metal(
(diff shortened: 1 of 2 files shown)


teoxoy (approved):
LGTM

PR #3512, opened 2023-02-21
Update naga to 0.11.0@git:9742f1616c3e3dd2cc9a5880616fc886c391bb9f

nical:
Another naga update  shortly after the previous one because we'd like to pick up https://github.com/gfx-rs/naga/pull/2259 in Gecko soon.

Issue #3492, opened 2023-02-16
zero_init_workgroup_mem fails in some configurations


nical:
**Description**

On some platforms `cargo nextest run fails` in `zero_init_workgroup_mem`.

For example on a Mac M1 max:

```
thread 'shader::zero_init_workgroup_mem::zero_init_workgroup_mem' panicked at 'Zero-initialization of workgroup memory failed (81% of disptaches failed).', wgpu/tests/shader/zero_init_workgroup_mem.rs:175:5
```

**Repro steps**

run `cargo nextest run`.

**Platform**

 - Mac M1 max (Moac OS 12.5.1 "Metal GPUFamily Apple 7") on current wgpu (rev 719ba83726343c0afd3f7327e0150a9b76e8d1b3)
 - Windows + DXC according to https://github.com/gfx-rs/wgpu/pull/3447#issue-1568690578

teoxoy:
@Elabajaba we think we fixed this issue in https://github.com/gfx-rs/naga/pull/2259 (https://github.com/gfx-rs/wgpu/pull/3512).
Can you confirm that on your Windows + DXC setup it now succeeds?

nical:
The mac M1 issue is fixed.

Elabajaba:
> Can you confirm that on your Windows + DXC setup it now succeeds?

DX12 now passes all the tests with both FXC and DXC. (apart from the 1 skipped test)

cwfitzgerald:
Sounds like this has been fixed!

PR #3515, opened 2023-02-21
Remove SwiftShader failure from zero init test

teoxoy:
We fixed this in naga. See https://github.com/gfx-rs/wgpu/pull/3512 and https://github.com/gfx-rs/wgpu/issues/3492.

### DIFF
--- a/wgpu/tests/shader/zero_init_workgroup_mem.rs
+++ b/wgpu/tests/shader/zero_init_workgroup_mem.rs
@@ -23,13 +23,6 @@ fn zero_init_workgroup_mem() {
                 Some(5140),
                 Some("Microsoft Basic Render Driver"),
                 true,
-            )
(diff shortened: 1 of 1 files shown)


wgpu/tests/shader/zero_init_workgroup_mem.rs
```diff
@@ -23,13 +23,6 @@ fn zero_init_workgroup_mem() {
                 Some(5140),
                 Some("Microsoft Basic Render Driver"),
                 true,
```
cwfitzgerald, line 25: This should be false unless the test segfaults

cwfitzgerald: I guess if the test is flaky, that would count too

cwfitzgerald (approved):

Issue #3245, opened 2022-11-29
Emscripten `wgpu-hal` example is broken


jimblandy:
When I build `wgpu-hal/examples/raw-gles.rs` according to the instructions in #2422, I get the following error in the page's console:

```
This page is in Quirks Mode. Page layout may be impacted. For Standards Mode use “<!DOCTYPE html>”.
[raw-gles.em.html](http://localhost:8000/raw-gles.em.html)
GEThttp://localhost:8000/favicon.ico
[HTTP/1.0 404 File not found 1ms]

Initializing external GL context [raw-gles.js:1799:16](http://localhost:8000/raw-gles.js)
Hooking up to wgpu-hal [raw-gles.js:1799:16](http://localhost:8000/raw-gles.js)
thread 'main' panicked at 'Get parameter string 0x9245 failed. Maybe your GL context version is too outdated.', /home/jimb/.cargo/git/checkouts/glow-0323f9242b6190b5/c8a011f/src/native.rs:1181:13 [raw-gles.js:1811:16](http://localhost:8000/raw-gles.js)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace [raw-gles.js:1811:16](http://localhost:8000/raw-gles.js)
WebGL warning: getParameter: Bad `pname`: 0x9245
```

I'm building and serving the page as follows:

```
$  cargo clean
$  EMCC_CFLAGS="-g -s ERROR_ON_UNDEFINED_SYMBOLS=0 --no-entry -s FULL_ES3=1" cargo build --example raw-gles --target wasm32-unknown-emscripten --features gles,emscripten,webgl
$  cp wgpu-hal/examples/raw-gles.em.html target/wasm32-unknown-emscripten/debug/examples/
$  cd target/wasm32-unknown-emscripten/debug/examples/
$  python -m http.server
```

jimblandy:
FWIW, the GLenum `0x9245` is `UNMASKED_VENDOR_WEBGL` in the [`WEBGL_debug_renderer_info`](https://registry.khronos.org/webgl/extensions/WEBGL_debug_renderer_info/) extension.

jimblandy:
[...]

PR #3519, opened 2023-02-21
enable debug extension on emscripten

coderedart:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
https://github.com/gfx-rs/wgpu/issues/3245

**Description**
we check if `WEBGL_debug_renderer_info` extension is available on emscripten and if it is, we query for unmasked vendor/renderer strings. 
but on webgl, we need to enable the extension before using it. but emscripten avoids enabling this particular extension by default. 
so, we use emscripten's extern functions defined in https://github.com/emscripten-core/emscripten/blob/4d864df0a57024d667e8db171aab2c3385d04c30/system/include/emscripten/html5_webgl.h#L61 to manually enable the extension if it is found on emscripten target. or fallback to  vendor/renderer constants instead of the unmasked variants.

**Testing**
wgpu-hal's gles html example used for testing. 

I believe the wiki needs to be updated https://github.com/gfx-rs/wgpu/wiki/Running-on-the-web-with-Emscripten

### DIFF
--- a/wgpu-hal/src/gles/adapter.rs
+++ b/wgpu-hal/src/gles/adapter.rs
@@ -186,14 +186,15 @@ impl super::Adapter {
         let extensions = gl.supported_extensions();
 
         let (vendor_const, renderer_const) = if extensions.contains("WEBGL_debug_renderer_info") {
-            // context must be current
+            // emscripten doesn't enable "WEBGL_debug_renderer_info" extension by default. so, we do it manually.
+            // See https://github.com/gfx-rs/wgpu/issues/3245 for context
             #[cfg(target_os = "emscripten")]
             if unsafe { super::emscripten::enable_extension("WEBGL_debug_renderer_info\0") } {
                 (GL_UNMASKED_VENDOR_WEBGL, GL_UNMASKED_RENDERER_WEBGL)
             } else {
                 (glow::VENDOR, glow::RENDERER)
             }
-            // glow enables WEBGL_debug_renderer_info on w-u-u target.
+            // glow already enables WEBGL_debug_renderer_info on wasm32-unknown-unknown target by default.
             #[cfg(not(target_os = "emscripten"))]
             (GL_UNMASKED_VENDOR_WEBGL, GL_UNMASKED_RENDERER_WEBGL)
         } else {
--- a/wgpu-hal/src/gles/adapter.rs
+++ b/wgpu-hal/src/gles/adapter.rs
@@ -8,7 +8,14 @@ use crate::auxil::db;
 
 const GL_UNMASKED_VENDOR_WEBGL: u32 = 0x9245;
 const GL_UNMASKED_RENDERER_WEBGL: u32 = 0x9246;
-
+extern "C" {
+    /// returns 1 if success. 0 if failure. extension name must be null terminated
+    pub fn emscripten_webgl_enable_extension(
+        context: std::ffi::c_int,
+        extension: *const std::ffi::c_char,
+    ) -> std::ffi::c_int;
+    pub fn emscripten_webgl_get_current_context() -> std::ffi::c_int;
+}
 impl super::Adapter {
     /// According to the OpenGL specification, the version information is
     /// expected to follow the following syntax:
@@ -184,8 +191,18 @@ impl super::Adapter {
     ) -> Option<crate::ExposedAdapter<super::Api>> {
         let gl = context.lock();
         let extensions = gl.supported_extensions();
-
-        let (vendor_const, renderer_const) = if extensions.contains("WEBGL_debug_renderer_info") {
+        // if emscripten, has debug extension and if we can enable it -> use unmasked vendor/renderer
+        // https://github.com/gfx-rs/wgpu/issues/3245
+        let (vendor_const, renderer_const) = if cfg!(target_os = "emscripten")
+            && extensions.contains("WEBGL_debug_renderer_info")
+            && unsafe {
+                emscripten_webgl_enable_extension(
+                    emscripten_webgl_get_current_context(),
+                    "WEBGL_debug_renderer_info\0".as_ptr() as _,
+                )
+            } == 1
+        {
+            // if we try querying unmasked constants without enabling extension, we crash on emscripten.
             (GL_UNMASKED_VENDOR_WEBGL, GL_UNMASKED_RENDERER_WEBGL)
         } else {
             (glow::VENDOR, glow::RENDERER)
--- a/wgpu-hal/src/gles/adapter.rs
+++ b/wgpu-hal/src/gles/adapter.rs
@@ -6,16 +6,6 @@ use crate::auxil::db;
 
 // https://webgl2fundamentals.org/webgl/lessons/webgl-data-textures.html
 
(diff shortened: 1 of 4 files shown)


wgpu-hal/src/gles/adapter.rs
```diff
@@ -8,7 +8,14 @@ use crate::auxil::db;
...
 const GL_UNMASKED_RENDERER_WEBGL: u32 = 0x9246;
-
+extern "C" {
+    /// returns 1 if success. 0 if failure. extension name must be null terminated
+    pub fn emscripten_webgl_enable_extension(
+        context: std::ffi::c_int,
+        extension: *const std::ffi::c_char,
+    ) -> std::ffi::c_int;
+    pub fn emscripten_webgl_get_current_context() -> std::ffi::c_int;
+}
```
cwfitzgerald, line 18: Could we move this to a emscripten module, and expose from that module a simple enable_extension that uses them

wgpu-hal/src/gles/adapter.rs
```diff
@@ -184,8 +191,18 @@ impl super::Adapter {
...
+        // https://github.com/gfx-rs/wgpu/issues/3245
+        let (vendor_const, renderer_const) = if cfg!(target_os = "emscripten")
+            && extensions.contains("WEBGL_debug_renderer_info")
+            && unsafe {
+                emscripten_webgl_enable_extension(
+                    emscripten_webgl_get_current_context(),
+                    "WEBGL_debug_renderer_info\0".as_ptr() as _,
+                )
+            } == 1
+        {
```
cwfitzgerald, line 204: The logic isn't right here.

It should be:

- if extension:
  - if emscripten:
    - activate extension
  - use extension

We don't need to check if activation fails, as it shouldn't fail

cwfitzgerald (changes requested):

wgpu-hal/src/gles/adapter.rs
coderedart: what if it fails though? better to fallback to vendor/renderer instead of crashing.

cwfitzgerald: Fair enough

wgpu-hal/src/gles/adapter.rs
coderedart: done

wgpu-hal/src/gles/adapter.rs
cwfitzgerald: The logic still needs to be changed here to what I listed, just with an extra conditional at the end

cwfitzgerald: The logic as it stands, excludes regular webgl wasm

coderedart: oh! you are right. I was looking at the web.rs file in the same directory and smh assumed that all w-u-u logic is separate from gles. thanks for pointing this out

coderedart: done. and tested on ff/chrome linux.

wgpu-hal/src/gles/adapter.rs
```diff
@@ -186,10 +186,20 @@ impl super::Adapter {
...
 
         let (vendor_const, renderer_const) = if extensions.contains("WEBGL_debug_renderer_info") {
+            // context must be current
+            #[cfg(target_os = "emscripten")]
+            if unsafe { super::emscripten::enable_extension("WEBGL_debug_renderer_info\0") } {
+                (GL_UNMASKED_VENDOR_WEBGL, GL_UNMASKED_RENDERER_WEBGL)
+            } else {
+                (glow::VENDOR, glow::RENDERER)
+            }
+            // glow enables WEBGL_debug_renderer_info on w-u-u target.
```
cwfitzgerald, line 196: W-u-u is great shorthand, but unreadable if you don't know what it means  - maybe just call it the "standard wasm target" or "non-emscripten wasm"

wgpu-hal/src/gles/adapter.rs
```diff
@@ -186,10 +186,20 @@ impl super::Adapter {
         let extensions = gl.supported_extensions();
 
         let (vendor_const, renderer_const) = if extensions.contains("WEBGL_debug_renderer_info") {
+            // context must be current
```
cwfitzgerald, line 189: Comment in the wrong place

wgpu-hal/src/gles/emscripten.rs
```diff
@@ -0,0 +1,22 @@
+extern "C" {
+    /// returns 1 if success. 0 if failure. extension name must be null terminated
+    fn emscripten_webgl_enable_extension(
+        context: std::ffi::c_int,
+        extension: *const std::ffi::c_char,
+    ) -> std::ffi::c_int;
+    fn emscripten_webgl_get_current_context() -> std::ffi::c_int;
+}
+/// if we have debug extension and if we can enable it -> use unmasked vendor/renderer
```
cwfitzgerald, line 9:
```suggestion
/// If we have the debug extension and if we can enable it, enable it. 
```

wgpu-hal/src/gles/emscripten.rs
```diff
@@ -0,0 +1,22 @@
...
+/// https://github.com/gfx-rs/wgpu/issues/3245
+/// returns true on success
+/// # Safety:
+/// opengl context MUST BE current
+/// extension_name_null_terminated argument must be a valid string with null terminator.
```
cwfitzgerald, line 14:
```suggestion
/// # Safety:
///
/// - Context must be current
/// - `extension_name_null_terminated` argument must be a valid string with null terminator.
```

wgpu-hal/src/gles/emscripten.rs
```diff
@@ -0,0 +1,22 @@
+extern "C" {
+    /// returns 1 if success. 0 if failure. extension name must be null terminated
+    fn emscripten_webgl_enable_extension(
+        context: std::ffi::c_int,
+        extension: *const std::ffi::c_char,
+    ) -> std::ffi::c_int;
+    fn emscripten_webgl_get_current_context() -> std::ffi::c_int;
+}
+/// if we have debug extension and if we can enable it -> use unmasked vendor/renderer
+/// https://github.com/gfx-rs/wgpu/issues/3245
```
cwfitzgerald, line 10:
```suggestion
///
/// See https://github.com/gfx-rs/wgpu/issues/3245 for context
///
```

cwfitzgerald (changes requested):
Some notes

wgpu-hal/src/gles/adapter.rs
coderedart: where should i place it? i thought being closest the unsafe usage would be good.

wgpu-hal/src/gles/emscripten.rs
coderedart: moved this comment to adapter file because that's where we actually make this decision.

cwfitzgerald (approved):
Amazing!

coderedart:
nice. don't forget to update the wiki too :+1:

cwfitzgerald:
The wiki is publicly editable :)

PR #3521, opened 2023-02-22
Replace "Intel/Vulkan" w/ "Vulkan on Intel"

### LINKED ISSUE #3520: Docs suggestion: be more explicit about supported platforms in the `Features` documentation
For example:

```
        /// Enables ASTC family of compressed textures. ASTC textures use pixel blocks varying from 4x4 to 12x12.
        /// Blocks are always 16 bytes.
        ///
        /// Compressed textures sacrifice some quality in exchange for significantly reduced
        /// bandwidth usage.
        ///
        /// Support for this feature guarantees availability of [`TextureUsages::COPY_SRC | TextureUsages::COPY_DST | TextureUsages::TEXTURE_BINDING`] for ASTC formats.
        /// [`Features::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES`] may enable additional usages.
        ///
        /// Supported Platforms:
        /// - Intel/Vulkan
        /// - Mobile (some)
        ///
        /// This is a web and native feature.
        const TEXTURE_COMPRESSION_ASTC_LDR = 1 << 5;
```

> Intel/Vulkan

This part confused me into thinking that any Vulkan driver would support this format. I realize in hindsight (after asking @cwfitzgerald) that this is wrong and probably doesn't make sense to interpret this way.

But I think it would be better to explicitly say "Vulkan on Intel" to avoid any confusion.

bonsairobo:
**Checklist**

- [x] Run `cargo clippy`.

Fixes #3520

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -265,7 +265,7 @@ bitflags::bitflags! {
         /// [`Features::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES`] may enable additional usages.
         ///
         /// Supported Platforms:
-        /// - Intel/Vulkan
+        /// - Vulkan on Intel
         /// - Mobile (some)
         ///
         /// This is a web and native feature.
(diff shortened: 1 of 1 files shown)


teoxoy (approved):
LGTM, thanks!

Issue #3345, opened 2023-01-03
"Suboptimal present of frame i" warn log spam on Android


James2022-rgb:
## Description

`log::warn!("Suboptimal present of frame {}", texture.index);` warn log may be emitted on every call to `vkQueuePresentKHR` via `wgpu::SurfaceTexture::present`, depending on the device and its current orientation.

Log output example from a session on my Galaxy S20 Ultra 5G (Adreno 650), Android 10:
```
01-03 11:25:00.523  3115  5742 W wgpu_hal::vulkan: [C:\Users\James\.cargo\registry\src\github.com-1ecc6299db9ec823\wgpu-hal-0.14.1\src\vulkan\mod.rs:659] Suboptimal present of frame 0
01-03 11:25:00.539  3115  5742 W wgpu_hal::vulkan: [C:\Users\James\.cargo\registry\src\github.com-1ecc6299db9ec823\wgpu-hal-0.14.1\src\vulkan\mod.rs:659] Suboptimal present of frame 1
01-03 11:25:00.556  3115  5742 W wgpu_hal::vulkan: [C:\Users\James\.cargo\registry\src\github.com-1ecc6299db9ec823\wgpu-hal-0.14.1\src\vulkan\mod.rs:659] Suboptimal present of frame 2
01-03 11:25:00.572  3115  5742 W wgpu_hal::vulkan: [C:\Users\James\.cargo\registry\src\github.com-1ecc6299db9ec823\wgpu-hal-0.14.1\src\vulkan\mod.rs:659] Suboptimal present of frame 0
01-03 11:25:00.590  3115  5742 W wgpu_hal::vulkan: [C:\Users\James\.cargo\registry\src\github.com-1ecc6299db9ec823\wgpu-hal-0.14.1\src\vulkan\mod.rs:659] Suboptimal present of frame 1
01-03 11:25:00.606  3115  5742 W wgpu_hal::vulkan: [C:\Users\James\.cargo\registry\src\github.com-1ecc6299db9ec823\wgpu-hal-0.14.1\src\vulkan\mod.rs:659] Suboptimal present of frame 2
...
...
```

Relevant piece of code in _wgpu-hal_:
https://github.com/gfx-rs/wgpu/blob/186a29c34d54d8628bbebb998a537210ac565b71/wgpu-hal/src/vulkan/mod.rs#L650-L662

[...]

PR #3525, opened 2023-02-23
Treat `VK_SUBOPTIMAL_KHR` as `VK_SUCCESS` on Android.

James2022-rgb:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Addresses issue #3345. 

**Description**
This PR makes is so that `vkQueuePresentKHR`'s `VK_SUBOPTIMAL_KHR` is treated as `VK_SUCCESS` (i.e ignored and not logged) on Android.

On Android 10+, `vkQueuePresentKHR` returns `VK_SUBOPTIMAL_KHR` if the swapchain's `VkSwapchainCreateInfoKHR::preTransform` does not match the device's current orientation.
This is always the case in _wgpu_ when the device orientation is anything other than the identity one,
because we unconditionally use `VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR` when creating the swapchain.

The assumption is that on Android, as discussed in the issue, `VK_SUBOPTIMAL_KHR` is only returned when the device's current orientation doesn't match `VkSwapchainCreateInfoKHR::preTransform`.

https://developer.android.com/games/optimize/vulkan-prerotation#detect_device_orientation_changes
> The most reliable way to detect an orientation change in your application is to verify whether the `vkQueuePresentKHR()` function returns `VK_SUBOPTIMAL_KHR`.

The only code path that returns `VK_SUBOPTIMAL_KHR` in Android libvulkan's `vkQueuePresentKHR` implementation:
https://android.googlesource.com/platform/frameworks/native/+/master/vulkan/libvulkan/swapchain.cpp#2021

More in-detail description of the problem in the connected issue.

**Testing**
Tested on my Galaxy S20 Ultra 5G (Adreno 650), Android 10, that the log spam is no longer exhibited.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,6 +109,10 @@ By @teoxoy in [#3436](https://github.com/gfx-rs/wgpu/pull/3436)
 - `copyTextureToTexture` src/dst aspects must both refer to all aspects of src/dst format. By @teoxoy in [#3431](https://github.com/gfx-rs/wgpu/pull/3431)
 - Validate before extracting texture selectors. By @teoxoy in [#3487](https://github.com/gfx-rs/wgpu/pull/3487)
 
+#### Vulkan
+
+- Treat `VK_SUBOPTIMAL_KHR` as `VK_SUCCESS` on Android. By @James2022-rgb in [#3525](https://github.com/gfx-rs/wgpu/pull/3525)
+
 ## wgpu-0.15.0 (2023-01-25)
 
 ### Major Changes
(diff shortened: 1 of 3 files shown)


teoxoy (changes requested):
Thanks for the PR!

Looks good but there is one more spot where we should ignore this.
Search for `acquire_next_image` in the codebase.

James2022-rgb:
Added a commit to ignore `VK_SUBOPTIMAL_KHR` for `vkAcquireNextImageKHR` as well!

teoxoy (approved):
LGTM

PR #3527, opened 2023-02-24
Don't rename imports

teoxoy:
I think renaming imports is not a great idea; it took me a bit to find out that `native` = `d3d12`.


cwfitzgerald (approved):
Full agree!

PR #3530, opened 2023-02-25
fix(deno): cleanup resources when closed

crowlKats:
This implements resource cleanup similar to the drop handlers used in https://github.com/gfx-rs/wgpu/blob/master/wgpu/src/lib.rs, as currently none of resources clean up and causes massive memory leaks.

Please review carefully as I potentially replaced some unrelated changes


cwfitzgerald (approved):
Nothing looks wrong here, but I could have also missed something

PR #3538, opened 2023-02-28
[metal] Fix metal erroring on an `array_stride` of 0

teoxoy:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
-

**Description**
Setting `array_stride` to 0 resulted in the following error:
```
Vertex Descriptor Validation
Attribute at index X references a buffer at index Y that has no stride.
```
Fixed it by calculating the stride and using metal's Constant step function + a rate of 0.

see https://developer.apple.com/documentation/metal/mtlvertexbufferlayoutdescriptor/1515341-stepfunction?language=objc

**Testing**
Should be covered by CTS; tested locally by setting `array_stride` to 0 in one of the tests.


wgpu-hal/src/auxil/mod.rs
```diff
@@ -65,6 +65,17 @@ pub fn align_to(value: u32, alignment: u32) -> u32 {
...
+pub fn align_to_u64(value: u64, alignment: u64) -> u64 {
+    if alignment.is_power_of_two() {
+        (value + alignment - 1) & !(alignment - 1)
+    } else {
+        match value % alignment {
+            0 => value,
+            other => value - other + alignment,
+        }
+    }
+}
```
cwfitzgerald, line 77: This should probably go in a `math` module in wgpu-types, we need it in `wgpu` and `wgpu-core` too

wgpu-hal/src/metal/device.rs
```diff
@@ -918,8 +918,20 @@ impl crate::Device<super::Api> for super::Device {
                         self.shared.private_caps.max_vertex_buffers as u64 - 1 - i as u64;
                     let buffer_desc = vertex_descriptor.layouts().object_at(buffer_index).unwrap();
 
-                    buffer_desc.set_stride(vb.array_stride);
-                    buffer_desc.set_step_function(conv::map_step_mode(vb.step_mode));
+                    if vb.array_stride == 0 {
```
cwfitzgerald, line 921: Could you just add a note about what we're working around here

cwfitzgerald (changes requested):

wgpu-hal/src/auxil/mod.rs
teoxoy: On a similar note, it might be nice to have an `Alignment` type similar to [naga's](https://github.com/gfx-rs/naga/blob/a7b52b8663a09506dc417e9149d64abcf637f9f4/src/proc/layouter.rs#L10).

wgpu/src/util/mod.rs
```diff
@@ -141,31 +140,3 @@ impl std::ops::Deref for DownloadBuffer {
         self.1.slice()
     }
 }
-
-///
```
cwfitzgerald, line 145: We should re-export `wgt::math` here

wgpu-types/src/math.rs
```diff
@@ -0,0 +1,57 @@
...
+/// # Examples
+///
+/// ```
+/// # use wgpu_types::math::align_to;
+/// assert_eq!(align_to(253, 16), 256);
+/// assert_eq!(align_to(256, 16), 256);
+/// assert_eq!(align_to(0, 16), 0);
+/// ```
+///
+pub fn align_to(value: u32, alignment: u32) -> u32 {
```
cwfitzgerald, line 19: Any reason this can't remain generic? I know it wouldn't get the po2 optimization, but this isn't likely to be a very hot function

wgpu/examples/shadow/main.rs
```diff
@@ -292,7 +292,7 @@ impl framework::Example for Example {
         let uniform_alignment = {
             let alignment =
                 device.limits().min_uniform_buffer_offset_alignment as wgpu::BufferAddress;
-            align_to(entity_uniform_size, alignment)
+            wgt::math::align_to_u64(entity_uniform_size, alignment)
```
cwfitzgerald, line 295: Examples shouldn't use the `wgt` alias - see re-export comment

cwfitzgerald (changes requested):

wgpu-hal/src/metal/device.rs
cwfitzgerald: This didn't seem to have happened?

cwfitzgerald (approved):
LGTM minus the one thing

wgpu-hal/src/metal/device.rs
teoxoy: Odd, I wrote a comment but it seems when I squashed a few commits I somehow lost this one

PR #3544, opened 2023-03-01
Add suballocation feature

xiaopengli89:
**Connections**
#3163 

**Description**
Suballocation sometimes seems to work incorrectly(uploading transparent pixel data).
图片

Disable `suballocation` feature to fallback to old behavior.

### DIFF
--- a/wgpu/Cargo.toml
+++ b/wgpu/Cargo.toml
@@ -86,7 +86,7 @@ name = "stencil-triangles"
 test = true
 
 [features]
-default = ["wgsl"]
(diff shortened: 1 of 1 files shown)


cwfitzgerald:
I agree this isn't the correct solution, I'm going to close this for now - please file an issue describing the problem

Elabajaba:
> I think we should fix the bug instead of making this optional. @Elabajaba got any ideas why this might be happening?

I have no idea. Iirc `gpu-allocator` doesn't zero the memory it gives you, but I thought we handled that at the wgpu level?

PR #3547, opened 2023-03-01
Add a DropGuard to Vulkan's `DeviceShared` implementation

JesseTG:
**Checklist**

- [x] Run `cargo clippy`.
- [ ] Run `RUSTFLAGS=--cfg=web_sys_unstable_apis cargo clippy --target wasm32-unknown-unknown` if applicable. **I don't know if this is applicable.**

**Connections**

None known.

**Description**

I need to use `wgpu` to manage an externally-owned `VkInstance` and `VkDevice`, in order to maintain compatibility with [ruffle](https://github.com/ruffle-rs/ruffle)'s `wgpu` renderer. I do so by creating a [`wgpu_hal::vulkan::Instance`](https://docs.rs/wgpu-hal/latest/wgpu_hal/vulkan/struct.Instance.html) and [`wgpu_hal::vulkan::Device`](https://docs.rs/wgpu-hal/latest/wgpu_hal/vulkan/struct.Device.html), respectively. I need to be able to clean up the `wgpu_hal` objects without cleaning up the underlying resources, since they're disposed of elsewhere.

I can do this with `VkInstance` through the `drop_guard` argument of [`Instance::from_raw`](https://docs.rs/wgpu-hal/latest/wgpu_hal/vulkan/struct.Instance.html#method.from_raw). No equivalent exists for `VkDevice`; this PR addresses that need.

**Testing**

It hasn't been tested yet, but I intend to try it out in [a project I'm working on](https://github.com/JesseTG/ruffle_libretro). For reasons beyond the scope of this post, I need time to integrate it.


JesseTG:
Hm, it seems that I have to fix that failing test first.

JesseTG:
Actually, wait, this PR might not be necessary at all. Let me double-check something before I close it.

JesseTG:
Nope, don't need this.

PR #3562, opened 2023-03-06
Update deno

teoxoy:
update deno to avoid https://rustsec.org/advisories/RUSTSEC-2023-0005

also updates tempfile to avoid https://rustsec.org/advisories/RUSTSEC-2023-0018


crowlKats:
This wont work; deno had a large rework in terms of js files, and a larger set of changes need to be applied. I'll work on it later today

crowlKats:
@teoxoy should be good now.

cwfitzgerald (approved):

PR #3563, opened 2023-03-07
In `Context::handle_error_fatal`, print cause chain.

### LINKED ISSUE #3551: `handle_error_fatal` does not print the complete error
**Description**
I tried using a render bundle and got this error:

`thread 'main' panicked at 'Error in RenderBundleEncoder::finish: In a set_pipeline command', /Users/kpreid/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.15.0/src/backend/direct.rs:316:9`

The error is incomplete: it doesn't say _what_ was wrong with the operation. Based on the line number, this appears to be because

```rust
    fn handle_error_fatal(
        &self,
        cause: impl Error + Send + Sync + 'static,
        string: &'static str,
    ) -> ! {
        panic!("Error in {string}: {cause}");
    }
```

prints `cause` but not `cause`'s chain (like non-fatal error handling does).

kpreid:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
Fixes #3551.

**Description**
In `Context::handle_error_fatal`, print cause chain using the same formatting as for non-fatal errors.

Also add `#[track_caller]` so that the panic location is specific to the error.

**Testing**
Manually tested using an application with a deliberately introduced error. Example result:

```
thread 'main' panicked at 'Error in RenderBundleEncoder::finish: Validation Error

Caused by:
    In a set_pipeline command
      note: render pipeline = `BloomPipelines::downsample_pipeline`
    Render pipeline targets are incompatible with render pass
    Incompatible color attachment: the renderpass expected [] but was given [Some(Rgba16Float)]
', /Users/kpreid/Projects/rust/wgpu/wgpu/src/backend/direct.rs:2105:18
```

Previously, the message would have ended at "In a set_pipeline command".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -125,6 +125,7 @@ By @teoxoy in [#3534](https://github.com/gfx-rs/wgpu/pull/3534)
 
 - `copyTextureToTexture` src/dst aspects must both refer to all aspects of src/dst format. By @teoxoy in [#3431](https://github.com/gfx-rs/wgpu/pull/3431)
 - Validate before extracting texture selectors. By @teoxoy in [#3487](https://github.com/gfx-rs/wgpu/pull/3487)
+- Fix fatal errors (those which panic even if an error handler is set) not including all of the details. By @kpreid in [#3563](https://github.com/gfx-rs/wgpu/pull/3563)
 
 #### Vulkan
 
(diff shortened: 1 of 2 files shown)


kpreid:
One possible flaw in this PR is that some fatal errors probably should not be described as "Validation Error"s (which `format_error` currently does unconditionally). I don't know exactly which ones, though.

cwfitzgerald:
Fatal errors are almost all not supposed to be fatal, we just haven't implemented non-fatality yet.

cwfitzgerald (approved):

PR #3564, opened 2023-03-07
DX12: Discard uninitialized buffers

### LINKED ISSUE #3552: Suballocation works incorrectly
**Description**
Related to #3544 

**Repro steps**
Uploading transparent pixel data

Transparent pixels should be rendered correctly

**Extra materials**
图片

**Platform**
os: Win10
gpu: Intel iris xe
wgpu: 0.15 with dx12 backend

xiaopengli89:
**Connections**
Close #3552 

**Description**
[Initializing DX12 Textures After Allocation and Aliasing](https://asawicki.info/news_1724_initializing_dx12_textures_after_allocation_and_aliasing)

### DIFF
--- a/wgpu-hal/src/dx12/command.rs
+++ b/wgpu-hal/src/dx12/command.rs
@@ -1,7 +1,7 @@
 use crate::auxil::{self, dxgi::result::HResult as _};
 
 use super::conv;
-use std::{mem, ops::Range, ptr};
+use std::{mem, ops::Range, ptr, sync::atomic::Ordering};
 use winapi::um::d3d12 as d3d12_ty;
 
 fn make_box(origin: &wgt::Origin3d, size: &crate::CopyExtent) -> d3d12_ty::D3D12_BOX {
@@ -312,11 +312,8 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
                 barrier.usage.start,
                 barrier.usage.end
             );
-            if barrier
-                .buffer
-                .memory_flags
-                .contains(crate::MemoryFlags::TRANSIENT)
-            {
+
+            if !barrier.buffer.initialized.swap(true, Ordering::SeqCst) {
                 unsafe {
                     self.list
                         .unwrap()
--- a/wgpu-hal/src/dx12/command.rs
+++ b/wgpu-hal/src/dx12/command.rs
@@ -466,14 +466,6 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
         }
     }
 
-    unsafe fn discard_buffer(&mut self, buffer: &super::Buffer) {
-        unsafe {
-            self.list
-                .unwrap()
-                .DiscardResource(buffer.resource.as_mut_ptr(), ptr::null());
-        }
-    }
-
     unsafe fn copy_buffer_to_buffer<T>(
         &mut self,
         src: &super::Buffer,
--- a/wgpu-hal/src/dx12/command.rs
+++ b/wgpu-hal/src/dx12/command.rs
@@ -312,6 +312,18 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
                 barrier.usage.start,
                 barrier.usage.end
             );
+            if barrier
+                .buffer
+                .memory_flags
+                .contains(crate::MemoryFlags::TRANSIENT)
+            {
+                unsafe {
+                    self.list
+                        .unwrap()
+                        .DiscardResource(barrier.buffer.resource.as_mut_ptr(), ptr::null());
+                }
+            }
+
             let s0 = conv::map_buffer_usage_to_state(barrier.usage.start);
             let s1 = conv::map_buffer_usage_to_state(barrier.usage.end);
             if s0 != s1 {
--- a/wgpu-hal/src/dx12/command.rs
+++ b/wgpu-hal/src/dx12/command.rs
@@ -466,6 +466,14 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
         }
     }
 
+    unsafe fn discard_buffer(&mut self, buffer: &super::Buffer) {
+        unsafe {
+            self.list
+                .unwrap()
+                .DiscardResource(buffer.resource.as_mut_ptr(), ptr::null());
+        }
+    }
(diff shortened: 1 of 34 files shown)


teoxoy:
I think it would make sense to instead call `DiscardResource` in `transition_buffers` for all buffers that were created with the `hal::MemoryFlags::TRANSIENT` flag.

@cwfitzgerald thoughts?

Elabajaba:
I get that it's named after the DX12 API call, but I feel like we could probably come up with a better name than `discard_buffer()`.

Afaik we don't currently allow memory aliasing (creating multiple resources in the same or overlapping region of an `ID3D12Heap`), and I'm not sure it's possible to allow it with the current API so we can probably ignore that case.

cwfitzgerald:
I agree, if we need to call this at all, it should be hidden inside the transitions.

I don't fully understand why we would need this though, as we should always clear resources using an approved method before use as part of our zero-init process.

xiaopengli89:
> I agree, if we need to call this at all, it should be hidden inside the transitions.
> 
> I don't fully understand why we would need this though, as we should always clear resources using an approved method before use as part of our zero-init process.

We don't do any initialization before writing to staging buffers, this may an undefined behavior on some hardware or driver.

wgpu-hal/src/dx12/command.rs
```diff
@@ -312,6 +312,18 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
...
+                .buffer
+                .memory_flags
+                .contains(crate::MemoryFlags::TRANSIENT)
+            {
+                unsafe {
+                    self.list
+                        .unwrap()
+                        .DiscardResource(barrier.buffer.resource.as_mut_ptr(), ptr::null());
+                }
+            }
```
cwfitzgerald, line 325: This is overzealous - we should only issue a discard if the transition is from undefined, otherwise we'll be blowing away perfectly good data.

cwfitzgerald (changes requested):

cwfitzgerald:
I still don't understand why this is needed:
- Transient buffers come from write_*.
- None of those buffers are render target or depth state.

> Placed and reserved resources with either the render target or depth stencil flags must be initialized with one of the following operations before other operations are supported.

wgpu-hal/src/dx12/command.rs
teoxoy: I was thinking that this might be ok since we only call `transition_buffers` once with the temp buffer to transition it to be the copy source but adding a check wouldn't hurt.

One thing to note is that the transition is always from `MAP_WRITE` to `COPY_SRC` (and this should be what we check for).

https://github.com/gfx-rs/wgpu/blob/110e62aed255777ac635a504ec81029483065be8/wgpu-core/src/device/queue.rs#L801-L804

cwfitzgerald: Actually, that won't work either. If we discard the resource _after_ writing the bytes to it, we'll get a potentially munged data.

teoxoy: Hmm, right

xiaopengli89: So we should add a `UNDEFINED` flag, right?

teoxoy: Does this PR actually fix the issue you encountered? After rereading the docs, I'm also unsure why this would be needed.

xiaopengli89: > Does this PR actually fix the issue you encountered? After rereading the docs, I'm also unsure why this would be needed.

Yes, at the beginning I changed the fully transparent data to rgba (0x00, 0x00, 0x00, 0x01), and the problem no longer appeared.   

Later, I searched for the commit history of `wgpu` to locate the `suballocation` feature, and then I disable `suballocation` to solve this problem, because the `CommittedResource` will be initialized automatically.   

I suspect that some hardware or drivers will assume that the buffer has been initialized with 0 value, so some write operations may be skipped, the documentation of msdn does only mention render target or depth stencil need to be initialized, but from my actual test results, `DiscardResource` can solve this problem.

wgpu-hal/src/dx12/command.rs
```diff
@@ -312,6 +312,15 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
                 barrier.usage.start,
                 barrier.usage.end
             );
+
+            if !barrier.buffer.initialized.swap(true, Ordering::SeqCst) {
+                unsafe {
+                    self.list
+                        .unwrap()
+                        .DiscardResource(barrier.buffer.resource.as_mut_ptr(), ptr::null());
+                }
```
cwfitzgerald, line 321: I think the only reason we're getting away with this is that discard resource is a no-op on buffers on most implementations. By way of the spec, this will clear all the data we just wrote into it.

wgpu-hal/src/dx12/device.rs
```diff
@@ -349,6 +354,7 @@ impl crate::Device<super::Api> for super::Device {
         Ok(super::Buffer {
             resource,
             size,
+            initialized: AtomicBool::new(allocation.is_none()),
```
cwfitzgerald, line 357: Under the belief this is a bug in the Xe drivers, we should only be calling discard resource on the Xe drivers, and we should be doing it unconditionally for transiant buffers on their creation.

cwfitzgerald (changes requested):

wgpu-hal/src/dx12/command.rs
xiaopengli89: In my original commit, discard before writing data, would this be feasible?

wgpu-hal/src/dx12/device.rs
xiaopengli89: I prefer this to be the difference between `PlacedResource` and `CommittedResource`, because even on Xe, there is no problem using `CommittedResource`.

cwfitzgerald: Conditional on both xe and suballocation is fine

xiaopengli89:
Sorry, I found out today that discard does not completely solve the problem, although the probability of occurrence is much lower, I will seek help from intel technical support.

PR #3575, opened 2023-03-10
Add a generic parameter to `wgpu::Id`

TheOnlyMrCat:
**Checklist**

- [x] Run `cargo clippy`.

**Description**
The documentation I wrote in `wgpu::*::global_id()` didn't match the behaviour in `wgpu-core`. Specifically, `wgpu::Id`s are only unique among objects of the same type, while the documentation promised them to be unique among all objects spawned from the same `wgpu::Instance`.

This PR changes the documentation of all previous public `global_id` methods, and adds a type parameter to `wgpu::Id` to prevent the caller from trying to compare `Id`s from different resource types.

**Testing**
Not tested, only typechecked

### DIFF
--- a/wgpu/src/backend/web.rs
+++ b/wgpu/src/backend/web.rs
@@ -27,7 +27,7 @@ fn create_identified<T>(value: T) -> Identified<T> {
         if #[cfg(feature = "expose-ids")] {
             static NEXT_ID: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(1);
             let id = NEXT_ID.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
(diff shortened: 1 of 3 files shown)


cwfitzgerald (approved):
Seems reasonable enough, I just worry about a use case where you want to store multiple different kinds of Ids in the same structure, but I think that is enough of an edge case to not worry about

TheOnlyMrCat:
In any such use-case, you'd want to differentiate based on Id type anyway, since if you lose the information about what type of Id it is, you lose the ability to meaningfully compare them. It's easy enough to just create a bespoke enum that keeps that information around.

On an unrelated note, I forgot to consider how the `#[derive(Clone, Copy, ...)]` macros would handle the type parameter, so I need to submit another fix for that,

Issue #3578, opened 2023-03-11
WebGL Internal error: 0x00000502: Vertex buffer is not big enough for the draw call


madmaxio:
This happens on WebGL when recreating a buffer:

`[.WebGL-0000554C38753500] GL_INVALID_OPERATION: Error: 0x00000502, in ..\..\third_party\angle\src\libANGLE\renderer\d3d\VertexDataManager.cpp, reserveSpaceForAttrib:520. Internal error: 0x00000502: Vertex buffer is not big enough for the draw call.`

At the moment it is hard to provide small example to reproduce, sadly.

madmaxio:
And this error doesn't appear on wgpu 0.14.0, so the problem appeared somewhere on the way from wgpu 0.14.0 to 0.15.2, and looks like it is not libANGLE bug.

grovesNL:
Still having trouble simplifying a case to reproduce this. I'm not seeing much that would've changed between 0.14.0 and 0.15.2.

Is it possible it's the `STATIC_DRAW` to `DYNAMIC_DRAW` change in #3391 ? You could try reverting that locally and seeing if it helps.

madmaxio:
> Still having trouble simplifying a case to reproduce this. I'm not seeing much that would've changed between 0.14.0 and 0.15.2.
> 
> Is it possible it's the `STATIC_DRAW` to `DYNAMIC_DRAW` change in #3391 ? You could try reverting that locally and seeing if it helps.

That's exactly it. Reverting to glow::STATIC_DRAW from glow::DYNAMIC_DRAW fixes the problem. So this is about #3391 and #3371.

grovesNL:
Very weird. We can't back that fix out either because that breaks some older devices. `STATIC_DRAW` and `DYNAMIC_DRAW` is just supposed to be a usage hint so this whole situation is a bit silly...

> usage is provided as a performance hint only. The specified usage value does not constrain the actual usage pattern of the data store.

Is it possible that the buffer sizes are different between the buffer at creation vs. the one we use during draw? I don't think so because we I believe we're supposed to be pre-allocating the buffer fully and copying into that original buffer.

I wonder if we can easily trace this code path through ANGLE to see how we can workaround it, or possibly file a bug upstream to ANGLE if the buffer size is correct.

kdashg:
[...]

PR #3597, opened 2023-03-16
WebGl workaround for drawing single instance 

Wumpf:
**Checklist**
- [x] Run `cargo clippy`.

**Connections**
fixes https://github.com/gfx-rs/wgpu/issues/3578
backport: https://github.com/gfx-rs/wgpu/pull/3596

**Description**
Whenever there is any instance attribute, we need to need use an instanced draw call on ANGLE. Otherwise, the following error will be (potentially spuriously) generated and may render incorrectly:
```
[.WebGL-0000554C38753500] GL_INVALID_OPERATION: Error: 0x00000502, in ..\..\third_party\angle\src\libANGLE\renderer\d3d\VertexDataManager.cpp, reserveSpaceForAttrib:520. Internal error: 0x00000502: Vertex buffer is not big enough for the draw call.
```

**Testing**
Tested the backport to v0.15 on https://github.com/rerun-io/rerun

### DIFF
--- a/wgpu-hal/src/gles/queue.rs
+++ b/wgpu-hal/src/gles/queue.rs
@@ -164,18 +164,17 @@ impl super::Queue {
                 vertex_count,
                 instance_count,
             } => {
-                if instance_count == 1 {
-                    unsafe { gl.draw_arrays(topology, start_vertex as i32, vertex_count as i32) };
-                } else {
-                    unsafe {
-                        gl.draw_arrays_instanced(
-                            topology,
-                            start_vertex as i32,
-                            vertex_count as i32,
-                            instance_count as i32,
-                        )
-                    };
-                }
+                // Don't use `gl.draw_arrays` for `instance_count == 1`.
+                // Angle has a bug where it doesn't consider the instance divisor when `DYNAMIC_DRAW` is used in `draw_arrays`.
+                // See https://github.com/gfx-rs/wgpu/issues/3578
+                unsafe {
+                    gl.draw_arrays_instanced(
+                        topology,
+                        start_vertex as i32,
+                        vertex_count as i32,
+                        instance_count as i32,
+                    )
+                };
             }
             C::DrawIndexed {
                 topology,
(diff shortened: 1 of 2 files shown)


wgpu-hal/src/gles/queue.rs
```diff
@@ -184,44 +184,54 @@ impl super::Queue {
...
-                        index_type,
-                        index_offset as i32,
-                        instance_count as i32,
-                        base_vertex,
-                    )
-                },
-            },
+                any_per_instance_attributes_active,
+            } => {
+                let instanced = instance_count > 1
```
grovesNL, line 189: I think it's still an issue for `instance_count == 0`

wgpu-hal/src/gles/adapter.rs
```diff
@@ -564,6 +564,10 @@ impl super::Adapter {
             super::Workarounds::EMULATE_BUFFER_MAP,
             cfg!(target_arch = "wasm32"),
         );
+        workarounds.set(
+            super::Workarounds::INSTANCE_ATTRIBUTE_IMPLY_DRAW_INSTANCED,
+            cfg!(target_arch = "wasm32") || is_angle, // It would be nice to narrow down which browsers are affected (i.e., use ANGLE).
```
grovesNL, line 569: Firefox appears to be ok on ANGLE as long as `DrawArraysInstanced` is supported at the driver-level (see the conversation in Matrix), but maybe we could be conservative for now and remove this workaround on `wasm32` when it's fixed in Chrome stable. We could still use it for old ANGLE versions though.

wgpu-hal/src/gles/mod.rs
```diff
@@ -168,8 +168,19 @@ bitflags::bitflags! {
         // (https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/4972/diffs?diff_id=75888#22f5d1004713c9bbf857988c7efb81631ab88f99_323_327)
         // seems to indicate all skylake models are effected.
         const MESA_I915_SRGB_SHADER_CLEAR = 1 << 0;
-        /// Buffer map must emulated becuase it is not supported natively
+
+        /// Buffer map must emulated because it is not supported natively
         const EMULATE_BUFFER_MAP = 1 << 1;
+
+        /// Whenever there is any instance attribute, we need to need use an instanced draw call.
```
grovesNL, line 175: nit: extra `need`

grovesNL (commented):
I think we need this for both `C::Draw` and C::DrawIndexed`, so we could probably remove the `instance_count == 1`/`draw_arrays` branch in `C::Draw` entirely

Wumpf:
@grovesNL Given that Firefox just uses `DrawArraysInstanced` anyways in all circumstances (it's enabled for WebGL2 support which is a requirement for us), maybe we should simplify this entire PR down to just that and leave a comment? Is there any drawback to not always use the instanced version(s) in our hal layer?

Otherwise, only checking on `is_angle` sounds good! I thought this wouldn't work on the web correctly, but seems we get that information just fine.

grovesNL:
> maybe we should simplify this entire PR down to just that and leave a comment

That sounds good to me. I think originally we were considering falling back when we don't have instancing support, but our minimum requirements have been raised a bit since then.

Wumpf:
Done! Much simpler now, applied the same to the v0.15 backport :)

grovesNL (approved):
Thank you!

Issue #3483, opened 2023-02-13
Shader module unchecked doesn't disable bounds checking codegen


unfolding18:
**Description**
Naga adds additional bounds checking code to WGSL shaders prior to compilation. It is configurable via Naga, but does not appear to work in `wgpu-rs` itself. In particular, `create_shader_module_unchecked()` does not disable bounds checking. Note that I have only checked this on the Metal backend, as that is the hardware I am using/targeting.

**Repro steps**
Create a WGSL shader that indexes into any array, buffer, matrix, or vec. Set `wgpu` to use the Metal backend. Run the executable through Xcode's shader debugger. You should see that, regardless of whether you use `create_shader_module` or `create_shader_module_unchecked`, the resulting Metal shader code contains various conditionals and bounds checking routines.

**Expected vs observed behavior**
Ideally, `create_shader_module_unchecked` would disable bounds checking (or this would be configurable somehow via the API).

**Platform**
MacOS Ventura 13.2, `wgpu` version 0.14.0, Metal backend.

cwfitzgerald:
This appears to be a two sided problem. First is that we just are never changing the bounds check policies we pass to naga. See https://github.com/gfx-rs/wgpu/blob/e4445205c2df23a629760a1fccb8e46c574d2587/wgpu-hal/src/metal/device.rs#L692-L699 for where they are hard coded. Second we are setting them _in the pipline layout_, when the unchecked-ness comes on a per-module basis.

PR #3603, opened 2023-03-17
feat: correct bounds policy checking on MSL

FL33TW00D:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
Fixes https://github.com/gfx-rs/wgpu/issues/3483

**Description**
Previously on metal, using `create_shader_module_unchecked` would not set the bounds check policy to unchecked.

**Testing**
Check the produced metal shader.

### DIFF
--- a/wgpu-hal/src/metal/device.rs
+++ b/wgpu-hal/src/metal/device.rs
@@ -75,11 +75,27 @@ impl super::Device {
             },
         };
 
+        let mut temp_options;
+        let options = if !stage.module.runtime_checks {
+            temp_options = layout.naga_options.clone();
+            if !stage.module.runtime_checks {
+                temp_options.bounds_check_policies = naga::proc::BoundsCheckPolicies {
+                    index: naga::proc::BoundsCheckPolicy::Unchecked,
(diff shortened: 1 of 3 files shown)


wgpu-hal/src/metal/device.rs
```diff
@@ -75,11 +75,27 @@ impl super::Device {
             },
         };
 
+        let mut temp_options;
+        let options = if !stage.module.runtime_checks {
+            temp_options = layout.naga_options.clone();
+            if !stage.module.runtime_checks {
```
teoxoy, line 81: We already checked that `!stage.module.runtime_checks`

teoxoy (changes requested):
Thanks for the PR; it looks good besides the comment!

wgpu-hal/src/metal/device.rs
FL33TW00D: Thanks @teoxoy my mistake

FL33TW00D (commented):
Should be all good!

teoxoy (approved):
LGTM

Issue #3275, opened 2022-12-08
Relax `max_anisotropy` validation


teoxoy:
https://github.com/gfx-rs/wgpu/blob/a50836e0cbdb9a658213c7cee5d4d290e48397ed/wgpu-core/src/device/mod.rs#L1181-L1185

> Note: Most implementations support [maxAnisotropy](https://gpuweb.github.io/gpuweb/#dom-gpusamplerdescriptor-maxanisotropy) values in range between 1 and 16, inclusive. The used value of [maxAnisotropy](https://gpuweb.github.io/gpuweb/#dom-gpusamplerdescriptor-maxanisotropy) will be clamped to the maximum value that the platform supports.

_from https://gpuweb.github.io/gpuweb/#dom-gpusamplerdescriptor-maxanisotropy_

Filed https://github.com/gpuweb/gpuweb/issues/3671 to check for non power of two behavior.

PR #3610, opened 2023-03-20
Fix Metal Mipmap Behvior

cwfitzgerald:
**Checklist**

- [x] Run `cargo clippy`.

**Description**

Couple issues I found when debugging a problem: 

- The mipmap filter was set to `NotMipmapped` (which unconditionally samples mip 0) when the LodClamp was None. This is incorrect, LodClamp of None just means there's no restrictions on the Lod.
- Mipmap Lod averaging was turned on. This has ramifications when neighboring pixels are `texelFetch`-ing texels of different mipmaps. This would potentially blend the mipmap levels in an unintended way.

Also resolves https://github.com/gfx-rs/wgpu/issues/3275.

**Testing**

Tested using rend3.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,21 @@ The following `Features` have been renamed.
 
 By @teoxoy in [#3534](https://github.com/gfx-rs/wgpu/pull/3534)
 
+#### Anisotropic Filtering
+
+Anisotropic filtering has been brought in line with the spec. The anisotropic clamp is now a f32 which must be between 1.0 and 16.0 inclusive.
+
+If the anisotropy clamp is not 1.0, all the filters in a sampler must be `Linear`.
+
+```diff
+SamplerDescriptor {
+-    anisotropic_clamp: None,
++    anisotropic_clamp: 1.0,
+}
+```
+
+By @cwfitzgerald in [#3610](https://github.com/gfx-rs/wgpu/pull/3610).
+
 #### General
 
 - Change type of `mip_level_count` and `array_layer_count` (members of `TextureViewDescriptor` and `ImageSubresourceRange`) from `Option<NonZeroU32>` to `Option<u32>`. By @teoxoy in [#3445](https://github.com/gfx-rs/wgpu/pull/3445)
@@ -99,6 +114,7 @@ By @teoxoy in [#3534](https://github.com/gfx-rs/wgpu/pull/3534)
 - Improve attachment related errors. By @cwfitzgerald in [#3549](https://github.com/gfx-rs/wgpu/pull/3549)
 - Make error descriptions all upper case. By @cwfitzgerald in [#3549](https://github.com/gfx-rs/wgpu/pull/3549)
 - Don't include ANSI terminal color escape sequences in shader module validation error messages. By @jimblandy in [#3591](https://github.com/gfx-rs/wgpu/pull/3591)
+- Bring anisotropic filtering in line with the spec. 
 
 #### WebGPU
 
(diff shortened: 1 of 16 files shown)


CHANGELOG.md
```diff
@@ -113,6 +114,9 @@ By @teoxoy in [#3534](https://github.com/gfx-rs/wgpu/pull/3534)
 
 ### Bug Fixes
 
+#### Metal
+- Fix incorrect mipmap being sampled when using `MinLod <= 0.0` and `MaxLod >= 32.0` or when the fragment shaders in the same squad sample different Lods. By @cwfitzgerald in [#3610](https://github.com/gfx-rs/wgpu/pull/3610).
```
teoxoy, line 118:
```suggestion
- Fix incorrect mipmap being sampled when using `MinLod <= 0.0` and `MaxLod >= 32.0` or when the fragment shader samples different Lods in the same quad. By @cwfitzgerald in [#3610](https://github.com/gfx-rs/wgpu/pull/3610).
```

wgpu-hal/src/metal/device.rs
```diff
@@ -409,16 +409,12 @@ impl crate::Device<super::Api> for super::Device {
...
-        let caps = &self.shared.private_caps;
         objc::rc::autoreleasepool(|| {
             let descriptor = metal::SamplerDescriptor::new();
 
             descriptor.set_min_filter(conv::map_filter_mode(desc.min_filter));
             descriptor.set_mag_filter(conv::map_filter_mode(desc.mag_filter));
             descriptor.set_mip_filter(match desc.mipmap_filter {
-                wgt::FilterMode::Nearest if desc.lod_clamp.is_none() => {
-                    metal::MTLSamplerMipFilter::NotMipmapped
-                }
```
teoxoy, line 421: I think we can still check if min and max are both 0 and take this branch.

teoxoy (changes requested):
There are a few other things that don't look right with `lod_clamp` but could be addressed by a different PR.

- `lod_max_clamp` needs to default to `32` (right now it defaults to `f32::MAX`)
- `lod_clamp` is set to `None` if `lod_max_clamp > 32` which doesn't seem right
- `dx12` and `vulkan` then map a `lod_clamp: None` to `0.0..16.0` which also doesn't seem right; I think we should make `lod_clamp` required at the hal level

cwfitzgerald:
K, went a little hog wild fixing up sampler stuff. I also brought aniso behavior inline with the spec and made the errors better.

wgpu/src/lib.rs
```diff
@@ -1008,8 +1008,8 @@ pub struct SamplerDescriptor<'a> {
     pub lod_max_clamp: f32,
     /// If this is enabled, this is a comparison sampler using the given comparison function.
     pub compare: Option<CompareFunction>,
-    /// Valid values: 1, 2, 4, 8, and 16.
-    pub anisotropy_clamp: Option<NonZeroU8>,
+    /// Valid values between 1 and 16 inclusive. If this is not 1.0, all filter modes must be linear.
+    pub anisotropy_clamp: f32,
```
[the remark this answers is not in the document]
cwfitzgerald: Ah, misinterpreted `most implementations support` as a validation rule, will fix.

teoxoy:
Ah, actually one remaining thing regarding the LOD changes, could you also update the Default impl of the sampler descriptors? `lod_max_clamp` needs to default to `32` but it currently defaults to `f32::MAX`.

I'm not sure why we have 2 types of `SamplerDescriptor` (one in wgpu-core and another in wgpu), they seem to have the same-ish data.

wgpu-hal/src/vulkan/device.rs
```diff
@@ -1120,14 +1120,12 @@ impl crate::Device<super::Api> for super::Device {
                 .compare_op(conv::map_comparison(fun));
         }
 
-        if self
-            .shared
-            .downlevel_flags
-            .contains(wgt::DownlevelFlags::ANISOTROPIC_FILTERING)
-        {
+        if desc.anisotropy_clamp != 1 {
+            // We only enable the downlevel flag if supports 16x anisotropy,
```
teoxoy, line 1124:
```suggestion
            // We only enable anisotropy if it is supported,
```

Not sure if the wording is better (just a suggestion).

teoxoy (changes requested):

teoxoy (approved):
Looks great!

> Someday I'll stop pushing commits to this damn PR

I think you can now :)

PR #3630, opened 2023-03-31
Manually implement traits on `wgpu::Id`

TheOnlyMrCat:
**Checklist**

- [x] Run `cargo clippy`.

**Description**
In #3575, I neglected the fact that `#[derive(Trait)]` for built-in traits puts a `T: Trait` bound on `wgpu::Id`. This is certainly unwanted, so I've replaced the derive macros with the equivalent manual implementations.

**Testing**
Tested on my own code which relies upon these traits being implemented for concrete `T`


cwfitzgerald (approved):
Thanks!

PR #3632, opened 2023-04-01
add dxc error message formatting

Davidster:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
n/a

**Description**
Currently, error messages from dxc are not included in the error messages reported by wgpu. Before this change, error messages would look like this:

```
[2023-04-01T19:51:14Z ERROR wgpu::backend::direct] Handling wgpu errors as fatal by default

Caused by:
      note: label = `iced_wgpu::quad pipeline`
    Internal error in VERTEX shader: DXC compile error: DxcOperationResult { inner: ComPtr { ptr: 0x1c754cd23c0 } }

', wgpu\src\backend\direct.rs:3016:5
```

After this change, they look like this:

```
[2023-04-01T19:49:53Z ERROR wgpu::backend::direct] Handling wgpu errors as fatal by default

Caused by:
      note: label = `iced_wgpu::quad pipeline`
    Internal error in VERTEX shader: DXC compile error: "error: error reading 'iced_wgpu::quad::shader'\n\0"

', wgpu\src\backend\direct.rs:3016:5
```

**Testing**
The change can be tested by triggering a dxc compile error, which can be triggered by adding a colon in the name of the shader module:

```rust
wgpu::ShaderModuleDescriptor {
    label: Some(":"),
    ..
}
```

### DIFF
--- a/wgpu-hal/src/dx12/shader_compilation.rs
+++ b/wgpu-hal/src/dx12/shader_compilation.rs
@@ -83,8 +83,6 @@ pub(super) fn compile_fxc(
 mod dxc {
     use std::path::PathBuf;
 
-    use hassle_rs::{DxcBlob, DxcOperationResult};
-
     // Destructor order should be fine since _dxil and _dxc don't rely on each other.
     pub(crate) struct DxcContainer {
         compiler: hassle_rs::DxcCompiler,
@@ -178,12 +176,6 @@ mod dxc {
             &[],
         );
 
-        let format_error_message = |e: DxcOperationResult| {
-            e.get_error_buffer()
-                .and_then(|e| dxc_container.library.get_blob_as_string(&DxcBlob::from(e)))
-                .unwrap_or_default()
-        };
-
         let (result, log_level) = match compiled {
             Ok(dxc_result) => match dxc_result.get_result() {
                 Ok(dxc_blob) => {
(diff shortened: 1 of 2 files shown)


wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -176,6 +178,12 @@ mod dxc {
             &[],
         );
 
+        let format_error_message = |e: DxcOperationResult| {
```
Elabajaba, line 181: I generally prefer to make stuff like this a function, but it isn't used anywhere else and it's short enough that it isn't a big deal.

Elabajaba:
LGTM.

I don't love the closure, but that's just nitpicking over style.

wgpu-hal/src/dx12/shader_compilation.rs
Davidster: meh, I'd rather not infect the project with a change of style if possible. How about a private function in the dxc module, like `get_error_string_from_dxc_result`?

cwfitzgerald: We generally stay away from lambdas, yeah. A private function would be fine.

cwfitzgerald (approved):

PR #3641, opened 2023-04-03
[gles] Fix `tex_sub_image_*` calls not always adjusting to target texture

### LINKED ISSUE #3197: copy_texture_to_texture does not work correct in webGL2 with array textures.
**Description**
When using `copy_texture_to_texture` with the webGL2 backend and array textures I get invalid data displayed. It's hard to tell exactly what is going wrong since I can't really debug the code. I tried SpectorJS but it does not show individual textures.

**Repro steps**
Clone:
https://github.com/StarArawn/bevy_ecs_tilemap

Run(wasm/webGL2/array texture):
```
cargo build --target wasm32-unknown-unknown --example iso_diamond --release
wasm-server-runner .\target\wasm32-unknown-unknown\release\examples\iso_diamond.wasm
```

Run(wasm/webGL2/atlas):
```
cargo build --target wasm32-unknown-unknown --example iso_diamond --release --features atlas
wasm-server-runner .\target\wasm32-unknown-unknown\release\examples\iso_diamond.wasm
```

Run(desktop)
```
cargo run --example iso_diamond
```

Compare with the desktop or atlas versions and you can see the issue.

**Extra materials**
None

**Platform**
Chrome/WebGL2

daxpedda:
**Connections**
Fixes #3197.

**Description**
In GLES different arguments have to be passed into `tex_sub_image_*` calls depending on the target texture. Specifically cube maps, 2D texture arrays and cube map arrays. This was done correctly in some calls, specifically `write_texture`, but not in others.

This fixes it.

**Testing**
There was a test actually testing `copy_external_image_to_texture` for texture array, which was expected to fail, it succeeds now.

---

**Details**

The following functions now take the array index into account (only listing changes):
- `copy_external_image_to_texture` when targeting cube maps, 2D texture arrays and cube map arrays.
- `copy_texture_to_texture` when targeting 2D texture arrays and cube map arrays.
- `copy_buffer_to_texture` when targeting uncompressed cube map arrays.
- `copy_buffer_to_texture` when targeting compressed 2D texture arrays and cube map arrays.

Basically the change is to make sure that all calls to `tex_sub_image_*` take into account the following:
- If the target texture is a 2D texture array or cube map array, pass the array layer in the `offset_z` parameter.
- If the target texture is a cube map, pass the array layer in the `target` parameter.

### DIFF
--- a/wgpu/tests/external_texture.rs
+++ b/wgpu/tests/external_texture.rs
@@ -244,7 +244,7 @@ async fn image_bitmap_import() {
                         dest_layers = 2;
                     }
                     TestCase::SecondSliceCopy => {
-                        correct = false; // TODO: what?
+                        correct = true;
                         dest_origin.z = 1;
                         dest_data_layer = 1;
                         dest_layers = 2;
--- a/wgpu/tests/external_texture.rs
+++ b/wgpu/tests/external_texture.rs
@@ -162,8 +162,6 @@ async fn image_bitmap_import() {
 
                 // If the test is suppoed to be valid call to copyExternal.
                 let mut valid = true;
-                // If the result is incorrect
-                let mut correct = true;
                 match case {
                     TestCase::Normal => {}
                     TestCase::FlipY => {
@@ -244,7 +242,6 @@ async fn image_bitmap_import() {
                         dest_layers = 2;
                     }
                     TestCase::SecondSliceCopy => {
-                        correct = true;
                         dest_origin.z = 1;
                         dest_data_layer = 1;
                         dest_layers = 2;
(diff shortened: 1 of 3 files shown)


wgpu/tests/external_texture.rs
```diff
@@ -244,7 +244,7 @@ async fn image_bitmap_import() {
                         dest_layers = 2;
                     }
                     TestCase::SecondSliceCopy => {
-                        correct = false; // TODO: what?
+                        correct = true;
```
teoxoy, line 247: I think we can remove `correct` altogether now.

teoxoy (commented):
Looks good but I think we should also handle `TEXTURE_CUBE_MAP_ARRAY`.

wgpu/tests/external_texture.rs
daxpedda: Done.

daxpedda:
> Looks good but I think we should also handle `TEXTURE_CUBE_MAP_ARRAY`.

Done!

I updated OP with some details, would be great if you could confirm my assumptions to be right.

teoxoy (approved):
This looks great; thank you!

PR #3647, opened 2023-04-04
Properly drop surfaces

### LINKED ISSUE #3646: Surfaces leak, never being destroyed
**Description**
When making multiple surfaces the surfaces get leaked until program exit.

Incidentally this bug hides a different bug where destroying a surface leads to a vulkan validation error, as surfaces are not unconfigured before being destroyed.

**Repro steps**
1. Add logging to `wgpu-hal/src/vulkan/instance.rs#create_surface` and `wgpu-hal/src/vulkan/instance.rs#destroy_surface`
2. Create multiple windows, each with a surface
3. Close one window, observe no surface being destroyed
4. Close remaining windows, observe all surfaces being destroyed

Example project: https://gist.github.com/BenjaminSchaaf/46773ca8f1dc7217372c42d06922fd33

Surfaces should be destroyed when the `Surface` is dropped, instead of leaking.

**Platform**
This happens on master with vulkan.

BenjaminSchaaf:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Fixes #3646

**Description**
Surfaces weren't being dropped until exit due to the code commented out in Context::surface_drop. This hid a bug where surfaces dropped before exit weren't being unconfigured.

**Testing**
I ran some of the examples as well as the example from the issue and they worked as expected.


cwfitzgerald (approved):
Seems perfectly reasonable

PR #3652, opened 2023-04-05
Make wgpu-core errors non-exhaustive

cwfitzgerald:
**Checklist**

- [x] Run `cargo clippy`.

This lets us backport changes that change error messages. The error message is otherwise opaque to the user, but would be technically be a breaking change.


teoxoy (approved):
Sounds good!

PR #3657, opened 2023-04-08
Avoid using `WasmAbi` functions on WebGPU backend

### LINKED ISSUE #3430: Dynamic dispatch layer breaks on web

grovesNL:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
Fixes #3430

**Description**
`IntoWasmAbi` and `FromWasmAbi` can be problematic for our use case because `IntoWasmAbi` takes ownership of our type. We could workaround this in different ways, but the most straightforward path right now seems to be to avoid using these entirely and instead not tracking IDs for the WebGPU backend.

We could look at doing this in a few different ways once we internally make some changes to better support multi-threading (e.g., many types will be `Arc` internally so we could consider some other approaches).

**Testing**
Tested the WebGPU examples that are known to work (hello-triangle, boids, msaa-line, etc.) in Canary.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,8 +1,8 @@
 [workspace]
 resolver = "2"
 members = [
-    # "cts_runner",
-    # "deno_webgpu",
+    "cts_runner",
+    "deno_webgpu",
     "dummy",
     "player",
     "wgpu",
@@ -10,7 +10,7 @@ members = [
     "wgpu-hal",
     "wgpu-info",
     "wgpu-types",
-    # "run-wasm"
+    "run-wasm"
 ]
 exclude = []
 default-members = ["wgpu", "wgpu-hal", "wgpu-info", "wgpu-types"]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -170,6 +170,10 @@ By @cwfitzgerald in [#3610](https://github.com/gfx-rs/wgpu/pull/3610).
 
 - Fix metal erroring on an `array_stride` of 0. By @teoxoy in [#3538](https://github.com/gfx-rs/wgpu/pull/3538)
 
+#### WebGPU
+
+- Avoid using `WasmAbi` functions for WebGPU backend. By @grovesNL in [#3657](https://github.com/gfx-rs/wgpu/pull/3657)
+
 #### General
 
 - `copyTextureToTexture` src/dst aspects must both refer to all aspects of src/dst format. By @teoxoy in [#3431](https://github.com/gfx-rs/wgpu/pull/3431)
--- a/wgpu/src/backend/direct.rs
+++ b/wgpu/src/backend/direct.rs
@@ -2235,13 +2235,15 @@ impl crate::Context for Context {
         }
     }
 
-    fn queue_submit<I: Iterator<Item = Self::CommandBufferId>>(
+    fn queue_submit<I: Iterator<Item = (Self::CommandBufferId, Self::CommandBufferData)>>(
         &self,
         queue: &Self::QueueId,
         _queue_data: &Self::QueueData,
         command_buffers: I,
     ) -> (Self::SubmissionIndex, Self::SubmissionIndexData) {
-        let temp_command_buffers = command_buffers.collect::<SmallVec<[_; 4]>>();
+        let temp_command_buffers = command_buffers
+            .map(|(i, _)| i)
+            .collect::<SmallVec<[_; 4]>>();
 
         let global = &self.0;
         let index = match wgc::gfx_select!(*queue => global.queue_submit(*queue, &temp_command_buffers))
@@ -2924,9 +2926,11 @@ impl crate::Context for Context {
         &self,
         _pass: &mut Self::RenderPassId,
         pass_data: &mut Self::RenderPassData,
-        render_bundles: Box<dyn Iterator<Item = Self::RenderBundleId> + 'a>,
+        render_bundles: Box<
+            dyn Iterator<Item = (Self::RenderBundleId, &'a Self::RenderBundleData)> + 'a,
+        >,
     ) {
-        let temp_render_bundles = render_bundles.collect::<SmallVec<[_; 4]>>();
+        let temp_render_bundles = render_bundles.map(|(i, _)| i).collect::<SmallVec<[_; 4]>>();
         unsafe {
             wgpu_render_pass_execute_bundles(
                 pass_data,
--- a/wgpu/src/backend/web.rs
+++ b/wgpu/src/backend/web.rs
@@ -6,30 +6,27 @@ use std::{
     cell::RefCell,
     fmt,
     future::Future,
+    marker::PhantomData,
     ops::Range,
     pin::Pin,
     rc::Rc,
     task::{self, Poll},
 };
-use wasm_bindgen::{
-    convert::{FromWasmAbi, IntoWasmAbi},
-    prelude::*,
-    JsCast,
-};
+use wasm_bindgen::{prelude::*, JsCast};
 
 use crate::{
-    context::{ObjectId, QueueWriteBuffer, Unused},
+    context::{downcast_ref, ObjectId, QueueWriteBuffer, Unused},
     UncapturedErrorHandler,
 };
 
-fn create_identified<T>(value: T) -> Identified<T> {
+fn create_identified<T>(value: T) -> (Identified<T>, Sendable<T>) {
     cfg_if::cfg_if! {
         if #[cfg(feature = "expose-ids")] {
             static NEXT_ID: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(1);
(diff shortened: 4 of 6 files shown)


niklaskorz:
Tried it on https://github.com/niklaskorz/linon/tree/dynamic-dispatch-workaround and it works (deployed on https://niklaskorz.github.io/linon/):

grafik

A note though (that might be unrelated to this PR, not sure): using depth textures without a stencil part in the format does not seem to work in Chrome, i.e. using `Depth32Float` instead of `Depth24PlusStencil8` gives this error, even though no stencil ops are actually set in the render pass:

```
stencilLoadOp is (LoadOp::Load) and stencilStoreOp is (StoreOp::Store) when stencilReadOnly (0) or the attachment ([TextureView of Texture "depth_texture"]) has no stencil aspect.
 - While validating depthStencilAttachment.
 - While encoding [CommandEncoder "encoder"].BeginRenderPass([RenderPassDescriptor "rpass"]).

[Invalid CommandBuffer "encoder"] is invalid.
    at ValidateObject (..<URL>)
    at ValidateSubmit (..<URL>)

localhost/:1 WebGPU: too many warnings, no more warnings will be reported to the console for this GPUDevice.
```

Edit: ok this is indeed unrelated to this PR but it's an existing bug in wgpu, I'll try to fix it :)

cwfitzgerald (approved):
All seems reasonable, thank you for taking the time to fix this!

Cargo.toml
```diff
@@ -1,16 +1,16 @@
...
+    # "deno_webgpu",
     "dummy",
     "player",
     "wgpu",
     "wgpu-core",
     "wgpu-hal",
     "wgpu-info",
     "wgpu-types",
-    "run-wasm"
+    # "run-wasm"
```
SludgePhD: These should probably be reverted before merging

cwfitzgerald: Fudge

cwfitzgerald: Fudge

cwfitzgerald: Thankfully CI actually caught this.

PR #3668, opened 2023-04-12
Disable suballocation on Iris(R) Xe gpu

### LINKED ISSUE #3552: Suballocation works incorrectly

xiaopengli89:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Close #3552


Elabajaba:
This technically invalidates the safety comments for the suballocated path, but I don't think that really matters.

Should probably also not create the allocator (just add an early return to the top of `create_allocator_wrapper()` in suballocation.rs if suballocation isn't supported).

Also should probably add a comment above each of the `if !device.private_caps.suballocation_supported {` blocks saying that it's a workaround for Intel Xe drivers.

xiaopengli89:
Ok.

wgpu-hal/src/dx12/suballocation.rs
```diff
@@ -111,6 +157,40 @@ mod allocation {
         raw_desc: d3d12_ty::D3D12_RESOURCE_DESC,
         resource: &mut WeakPtr<ID3D12Resource>,
     ) -> Result<(HRESULT, Option<AllocationWrapper>), crate::DeviceError> {
+        // It's a workaround for Intel Xe drivers.
+        if !device.private_caps.suballocation_supported {
```
cwfitzgerald, line 127: Can we re-configure the code in some way such that we can just call the other function instead of copy and pasting the body?

I would:
- Rename modules to `placed` and `committed`.
- Make `placed` conditional on windows_rs, but `committed` unconditional 
- Re-export the module based on the windows_rs cfg.
- If suballocation isn't supported, call the committed version of the function.

wgpu-hal/src/dx12/device.rs
```diff
@@ -20,7 +20,10 @@ impl super::Device {
         library: &Arc<d3d12::D3D12Lib>,
         dx12_shader_compiler: wgt::Dx12Compiler,
     ) -> Result<Self, crate::DeviceError> {
-        let mem_allocator = super::suballocation::create_allocator_wrapper(&raw)?;
+        let mut mem_allocator = None;
+        if private_caps.suballocation_supported {
+            mem_allocator = super::suballocation::create_allocator_wrapper(&raw)?;
+        }
```
cwfitzgerald:
```suggestion
        let mem_allocator = if private_caps.suballocation_supported {
            Some(super::suballocation::create_allocator_wrapper(&raw)?)
        } else {
            None
        }
```

cwfitzgerald (changes requested):

wgpu-hal/src/dx12/suballocation.rs
Elabajaba: I originally did it this way to avoid having cfg blocks scattered through `device.rs`. We could move the inner part of the committed stuff out of the module and have the if blocks and the non-windows_rs path call those functions?

cwfitzgerald: I still thing that the suballocation module should automatically switch to the appropriate call, we just should _also_ have the committed functions available, just not re-exported.

Elabajaba: So having something like this?

```
#[cfg(windows_rs)]
pub(crate) use placed as allocation;
#[cfg(not(windows_rs))]
pub(crate) use committed as allocation;

#[cfg(windows_rs)]
mod placed {
    ...
}

pub(crate) mod committed {
    ...
}
```

wgpu-hal/src/dx12/suballocation.rs
```diff
@@ -63,6 +68,13 @@ mod allocation {
     ) -> Result<(HRESULT, Option<AllocationWrapper>), crate::DeviceError> {
         let is_cpu_read = desc.usage.contains(crate::BufferUses::MAP_READ);
         let is_cpu_write = desc.usage.contains(crate::BufferUses::MAP_WRITE);
+
+        // It's a workaround for Intel Xe drivers.
+        if !device.private_caps.suballocation_supported {
+            return super::committed::create_buffer_resource(device, desc, raw_desc, resource)
+                .map(|(hr, _)| (hr, None));
```
Elabajaba, line 75: Is mapping this necessary? Committed creates will never return an allocation.

wgpu-hal/src/dx12/suballocation.rs
```diff
@@ -111,6 +123,12 @@ mod allocation {
         raw_desc: d3d12_ty::D3D12_RESOURCE_DESC,
         resource: &mut WeakPtr<ID3D12Resource>,
     ) -> Result<(HRESULT, Option<AllocationWrapper>), crate::DeviceError> {
+        // It's a workaround for Intel Xe drivers.
+        if !device.private_caps.suballocation_supported {
+            return super::committed::create_texture_resource(device, desc, raw_desc, resource)
+                .map(|(hr, _)| (hr, None));
```
Elabajaba, line 129: Is mapping this necessary? Committed creates will never return an allocation.

wgpu-hal/src/dx12/suballocation.rs
xiaopengli89: Because the allocation type is different.

Elabajaba: Makes sense. I don't think we can fix that until Mozilla finally gets around to allowing `windows-rs`.

wgpu-hal/src/dx12/suballocation.rs
xiaopengli89: Done.

Elabajaba:
I'm getting a test failure when I run this on my AMD 6800xt when I force it to use suballocation on the `wgpu::wgpu-tests zero_init_texture_after_discard::discarding_depth_target_resets_texture_init_state_check_visible_on_copy_in_same_encoder` test. It passes when suballocation is enabled, but it times out if it's disabled.

Does it pass that test on an Xe iGPU?

xiaopengli89:
> I'm getting a test failure when I run this on my AMD 6800xt when I force it to use suballocation on the `wgpu::wgpu-tests zero_init_texture_after_discard::discarding_depth_target_resets_texture_init_state_check_visible_on_copy_in_same_encoder` test. It passes when suballocation is enabled, but it times out if it's disabled.
> 
> Does it pass that test on an Xe iGPU?

I ran the test on AMD Radeon(TM) Graphics with suballoction disabled and it pass `wgpu::wgpu-tests zero_init_texture_after_discard::discarding_depth_target_ resets_texture_init_state_check_visible_on_copy_in_same_encoder`.

But `texture_arrays_non_uniform` does not pass, the error is `Image data mismatch! Outlier count 768 over limit 0. Max difference 255`, even if suballoction is enabled, this should have nothing to do with this PR.

xiaopengli89:
All tests passed on my Intel(R) Iris(R) Xe Graphics.

cwfitzgerald (approved):
Thanks so much!

PR #3689, opened 2023-04-14
add feature RG11B10UFLOAT_RENDERABLE on webgpu

mockersf:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

**Description**

Add feature `RG11B10UFLOAT_RENDERABLE` on webgpu. I don't think this needs anything else added as, at least for Bevy, it was already working for metal/dx/vulkan

**Testing**

Rendered bloom examples from Bevy on webgpu and metal that use this texture format
webgpu:
Screenshot 2023-04-14 at 12 13 34

metal:
Screenshot 2023-04-14 at 12 31 11

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -296,7 +296,16 @@ bitflags::bitflags! {
         // ? const FLOAT32_BLENDABLE = 1 << 20; (https://github.com/gpuweb/gpuweb/issues/3556)
         // ? const 32BIT_FORMAT_MULTISAMPLE = 1 << 21; (https://github.com/gpuweb/gpuweb/issues/3844)
         // ? const 32BIT_FORMAT_RESOLVE = 1 << 22; (https://github.com/gpuweb/gpuweb/issues/3844)
-        // TODO const RG11B10UFLOAT_RENDERABLE = 1 << 23;
+
+        /// Allows for usage of textures of format [`TextureFormat::Rg11b10Float`] as a render target
+        ///
+        /// Supported platforms:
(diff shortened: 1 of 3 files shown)


wgpu-types/src/lib.rs
```diff
@@ -296,7 +296,16 @@ bitflags::bitflags! {
...
+
+        /// Allows for usage of textures of format [`TextureFormat::Rg11b10Float`] as a render target
+        ///
+        /// Supported platforms:
+        /// - Vulkan
+        /// - DX12
+        /// - Metal
+        ///
+        /// This is a web and native feature.
+        const RG11B10UFLOAT_RENDERABLE = 1 << 23;
```
cwfitzgerald, line 308: This is already working because bevy uses adapter specific texture format features. If a user isn't, it needs to both advertise this feature, and change how we check the flags for texture features

teoxoy: @mockersf you can use https://github.com/gfx-rs/wgpu/pull/3634 as an example for this

jinleili: As wgpu 0.16 is going to be released in the next few days, if there is a need, I am happy to complete this part of the code.

mockersf: I started looking into this, I will open the PR now but I'm trying to find if there are conditions to check before adding it to DX12/Vulkan/Metal

mockersf: https://github.com/gfx-rs/wgpu/pull/3702

PR #3690, opened 2023-04-14
Adjust internal canvas size on `Surface::configure()`

### LINKED ISSUE #3620: `Surface::configure()` should resize canvas
The problem is https://github.com/rust-windowing/winit/issues/2733:
> I am using [`transferControlToOffscreen()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/transferControlToOffscreen) to draw in a different thread, which unfortunately invalidates any calls to [`width()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/width) and [`height()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/height).

While trying to figure out how this could be fixed in `winit`, I got the idea that maybe this shouldn't be `winit`s problem after all. Setting the `width` and `height` of the canvas doesn't actually resize the canvas visually, it only resizes the surface the canvas is drawing on (the browser stretches or compresses the surface to fit the visual canvas size), which sounds like it would be the job of the renderer, not the window handler. Currently `winit` does both, changing the visual size of the canvas **and** it's surface.

`Surface::configure()` gets passed the size anyway, it would then use `width()` and `height()` to resize the canvas as well.
This is what `softbuffer` does as well ([src](https://github.com/rust-windowing/softbuffer/blob/77f9f68f45fc0926590b03cabcbf5fb652665353/src/web.rs#L80-L81)).

The alternative is to fix it in `winit`, which would require additional awareness by `winit` about what is done to the canvas to render to it.

daxpedda:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Fixes https://github.com/gfx-rs/wgpu/issues/3620.
Related to https://github.com/rust-windowing/winit/issues/2733.

**Description**
According to the [spec](https://html.spec.whatwg.org/multipage/canvas.html#output-bitmap), a canvas has two sizes, a rendered size, let's call it canvas size, and an output bitmap size. In WebGPU and WebGL a third size is added, the surface size.

The surface is rendered onto -> the output bitmap size is rendered onto -> the canvas. Each time the image will be resized to fit the size of what it's rendered to.

The canvas size is the only one actually affecting the visual size of the canvas (as long as it is set) on the screen and therefore the layout of the web page.

Currently `wgpu` only sets the surface size, when using `Surface::configure()`, this PR changes that to set the output bitmap size as well.

The main motivation was that when using [`HTMLCanvasElement.transferControlToOffscreen()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/transferControlToOffscreen) this can become difficult to handle otherwise. As setting the output bitmap size on the `HTMLCanvasElement` will fail. It is difficult to design an API around that in e.g. Winit (see https://github.com/rust-windowing/winit/issues/2733), so this PR alleviates this issue and shifts the responsibility of resizing the output bitmap size to the renderer.

After this PR, users of `wgpu` don't have to set the output bitmap size anymore, but they will need to continue to set the canvas size.

**Testing**
It's not, testing has to be done manually. Maybe an example can be adjusted to help test this manually?

### DIFF
--- a/wgpu/tests/common/mod.rs
+++ b/wgpu/tests/common/mod.rs
@@ -358,7 +358,7 @@ fn initialize_adapter() -> (Adapter, SurfaceGuard) {
         let canvas = create_html_canvas();
 
         let surface = instance
-            .create_surface_from_canvas(&canvas)
+            .create_surface_from_canvas(canvas.clone())
             .expect("could not create surface from canvas");
 
         surface_guard = SurfaceGuard { canvas };
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,6 +129,7 @@ By @cwfitzgerald in [#3671](https://github.com/gfx-rs/wgpu/pull/3671).
 - Change type of `mip_level_count` and `array_layer_count` (members of `TextureViewDescriptor` and `ImageSubresourceRange`) from `Option<NonZeroU32>` to `Option<u32>`. By @teoxoy in [#3445](https://github.com/gfx-rs/wgpu/pull/3445)
 - All `fxhash` dependencies have been replaced with `rustc-hash`. By @james7132 in [#3502](https://github.com/gfx-rs/wgpu/pull/3502)
 - Change type of `bytes_per_row` and `rows_per_image` (members of `ImageDataLayout`) from `Option<NonZeroU32>` to `Option<u32>`. By @teoxoy in [#3529](https://github.com/gfx-rs/wgpu/pull/3529)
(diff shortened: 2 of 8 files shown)


wgpu/tests/common/mod.rs
```diff
@@ -358,7 +358,7 @@ fn initialize_adapter() -> (Adapter, SurfaceGuard) {
         let canvas = create_html_canvas();
 
         let surface = instance
-            .create_surface_from_canvas(&canvas)
```
cwfitzgerald, line 361: This is a breaking change, so should be highlighted in the major changes section of the changelog

daxpedda: Done, I (hopefully) improved the changelog a bit.

cwfitzgerald (approved):

Issue #3026, opened 2022-09-15
Web and native backends disagree on some types being `Send` and `Sync`


i509VCB:
In #3025 when implementing static assertions to ensure types properly implement (or don't) `Send` and `Sync`, I noticed a few types do not agree depending whether you are compiling for web or native.

The following types differ on native and web:

| Type                 | Native            | Web               |
|---------------------|-------------------|--------------------|
| RenderPass     | Send + Sync | !Send + !Sync |
| ComputePass  | Send + Sync | !Send + !Sync |
| BufferView       | Send + Sync | !Send + !Sync |
| BufferViewMut | Send + Sync | !Send + !Sync |

teoxoy:
Related: https://github.com/gfx-rs/wgpu/issues/2652 & https://github.com/gfx-rs/wgpu/issues/2884

teoxoy:
Fixed by https://github.com/gfx-rs/wgpu/pull/3691

PR #3691, opened 2023-04-14
Don't implement `Send` or `Sync` on Wasm

### LINKED ISSUE #2884: wgpu calls `createBuffer` on an `undefined` from a worker thread on Wasm
**Description**
*wgpu* attempts to execute the `createBuffer` method on an `undefined` when running in a worker thread on the target `wasm32-unknown-unknown`. The instance, adapter, surface, and device were created on the main thread.

**Repro steps**
A runnable example is [here](https://norepimorphism.github.io), the source code for which is [here](https://github.com/norepimorphism/noctane); see [`wasm.rs`](https://github.com/norepimorphism/noctane/blob/a9f7e684785377acb230d39ef6b251f8c24c9589/core/examples/wasm.rs) and [`core/src/lib.rs`](https://github.com/norepimorphism/noctane/blob/a9f7e684785377acb230d39ef6b251f8c24c9589/core/src/lib.rs). More generic instructions are described below.

1. Create a new Rust crate targeting `wasm32-unknown-unknown`.
2. Export a `main` function using `#[wasm_bindgen]`.
3. In `main`, create a *wgpu* instance, adapter, device, and surface.
4. Create a new thread (e.g., using `wasm_thread`) in which a vertex buffer is created and initialized with the device returned by *wgpu*.
5. Observe the error.

I expect the vertex buffer creation to be successful, after which render passes would occur, the command queue would be submitted, and the surface frame would ultimately be presented.

Instead, I observe that attempting to create the vertex buffer from a worker thread crashes the Wasm process.

**Extra materials**
On Microsoft Edge (Chromium-based), I receive the following error:
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'createBuffer')

On Firefox, I receive the following error:
TypeError: can't access property "createBuffer", getObject(...) is undefined

**Platform**
Windows 11. *wgpu* 0.13 (latest minor release). *naga* 0.9.

daxpedda:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Fixes #2884.
Addresses #3026.

**Description**
Currently all types unsafely implement `Send` and `Sync` on the Wasm target under the assumption that Wasm doesn't support multi-threading. Which it actually does (on nightly).

This PR removes all these implementations and introduces `MaybeSend` and `MaybeSync` supertraits to continue implementing `Send` and `Sync` on non-Wasm targets.

**Testing**
It's not, maybe in the future we could add some Wasm multithreading examples/tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -138,6 +138,7 @@ jobs:
           set -e
 
           # build for WebGPU
+          cargo clippy --target ${{ matrix.target }} --tests --features glsl,spirv,fragile-send-sync-non-atomic-wasm
           cargo clippy --target ${{ matrix.target }} --tests --features glsl,spirv
           cargo doc --target ${{ matrix.target }} --no-deps --features glsl,spirv
 
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -746,7 +746,9 @@ pub struct RenderBundle<A: HalApi> {
     pub(crate) life_guard: LifeGuard,
 }
 
+#[cfg(not(target_arch = "wasm32"))]
 unsafe impl<A: HalApi> Send for RenderBundle<A> {}
+#[cfg(not(target_arch = "wasm32"))]
 unsafe impl<A: HalApi> Sync for RenderBundle<A> {}
 
 impl<A: HalApi> RenderBundle<A> {
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -746,9 +746,21 @@ pub struct RenderBundle<A: HalApi> {
     pub(crate) life_guard: LifeGuard,
 }
 
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(any(
+    not(target_arch = "wasm32"),
+    all(
+        feature = "fragile-send-sync-non-atomic-wasm",
+        not(target_feature = "atomics")
+    )
+))]
 unsafe impl<A: HalApi> Send for RenderBundle<A> {}
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(any(
+    not(target_arch = "wasm32"),
+    all(
+        feature = "fragile-send-sync-non-atomic-wasm",
+        not(target_feature = "atomics")
+    )
+))]
 unsafe impl<A: HalApi> Sync for RenderBundle<A> {}
 
 impl<A: HalApi> RenderBundle<A> {
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -6312,3 +6312,26 @@ impl Default for InstanceDescriptor {
         }
     }
 }
+
+pub use send_sync::*;
+
+#[doc(hidden)]
+mod send_sync {
+    #[cfg(not(target_arch = "wasm32"))]
+    pub trait WasmNotSend: Send {}
+    #[cfg(not(target_arch = "wasm32"))]
+    impl<T: Send> WasmNotSend for T {}
+    #[cfg(target_arch = "wasm32")]
+    pub trait WasmNotSend {}
+    #[cfg(target_arch = "wasm32")]
+    impl<T> WasmNotSend for T {}
+
+    #[cfg(not(target_arch = "wasm32"))]
+    pub trait WasmNotSync: Sync {}
+    #[cfg(not(target_arch = "wasm32"))]
+    impl<T: Sync> WasmNotSync for T {}
+    #[cfg(target_arch = "wasm32")]
+    pub trait WasmNotSync {}
+    #[cfg(target_arch = "wasm32")]
+    impl<T> WasmNotSync for T {}
+}
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -5917,6 +5917,19 @@ impl std::ops::Deref for ExternalImageSource {
     }
 }
 
+#[cfg(all(
+    target_arch = "wasm32",
+    feature = "fragile-send-sync-non-atomic-wasm",
+    not(target_feature = "atomics")
+))]
+unsafe impl Send for ExternalImageSource {}
+#[cfg(all(
+    target_arch = "wasm32",
+    feature = "fragile-send-sync-non-atomic-wasm",
+    not(target_feature = "atomics")
+))]
+unsafe impl Sync for ExternalImageSource {}
+
 /// Color spaces supported on the web.
 ///
 /// Corresponds to [HTML Canvas `PredefinedColorSpace`](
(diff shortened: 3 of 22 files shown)


wgpu/src/lib.rs
```diff
@@ -4531,3 +4581,51 @@ impl Display for Error {
         }
     }
 }
+
+use send_sync::*;
+
+mod send_sync {
```
cwfitzgerald, line 4977: Could we just put this as a two modules in `wgpu-types` so that we don't to duplicate it across crates?

wgpu/src/lib.rs
```diff
@@ -4531,3 +4581,51 @@ impl Display for Error {
...
+    use std::any::Any;
+    use std::fmt;
+
+    #[cfg(not(target_arch = "wasm32"))]
+    pub trait MaybeSend: Send {}
```
cwfitzgerald: We probably should call it something other than `MaybeSend`, as users will need to use it too.`WasmNotSend` or so? Suggestions welcome

wgpu/src/lib.rs
```diff
@@ -4502,16 +4543,25 @@ pub enum Error {
     /// Out of memory error
     OutOfMemory {
         /// Lower level source of the error.
+        #[cfg(not(target_arch = "wasm32"))]
         source: Box<dyn error::Error + Send + 'static>,
```
cwfitzgerald, line 4913: These can't use `MaybeSend`?

cwfitzgerald (changes requested):

wgpu/src/lib.rs
daxpedda: Unfortunately not, `dyn` has some serious limitations around that.
I could use a `Box<dyn ...>` type definition, but I thought it's a bit overkill in this case.

wgpu/src/lib.rs
daxpedda: Right, I keep forgetting about this crate.

wgpu/src/lib.rs
daxpedda: I thought about this but honestly can't come up with a good name.
Also played around with re-exporting `Send` and `NoSend` under the same name and hoping that `rustdoc` would show it under it's actual name, but it didn't.

I will just use `WasmNotSend`.

wgpu/src/lib.rs
daxpedda: Done.

wgpu/src/lib.rs
daxpedda: Done.

wgpu/src/lib.rs
daxpedda: Done.

wgpu-types/src/lib.rs
```diff
@@ -6165,3 +6160,26 @@ impl Default for InstanceDescriptor {
         }
     }
 }
+
+pub use send_sync::*;
+
+#[doc(hidden)]
```
daxpedda, line 6331: I hope that's appropriate.

wgpu/src/lib.rs
```diff
@@ -4531,3 +4582,35 @@ impl Display for Error {
...
+
+use send_sync::*;
+
+mod send_sync {
+    use std::any::Any;
+    use std::fmt;
+
+    use wgt::{WasmNotSend, WasmNotSync};
+
+    pub trait AnyWasmNotSendSync: Any + WasmNotSend + WasmNotSync {
```
daxpedda, line 4983: I left this trait here because it's only used in this crate.
I hope the name isn't too awful.

wgpu-types/src/lib.rs
cwfitzgerald: It won't show up in the docs as it's not a public module

daxpedda:
@cwfitzgerald happy to rebase when this is ready to be merged, just hit me up.
Don't want to continuously rebase unnecessarily.

cwfitzgerald (changes requested):
I've had a comment queued up for the last 2 months....

The main change this needs is to have a feature to enable Send/Sync on wasm with a scary name like `unsound_send_sync_threadless_wasm32` or something

wgpu-types/src/lib.rs
daxpedda: Right ... removed.

daxpedda:
> The main change this needs is to have a feature to enable Send/Sync on wasm with a scary name like `unsound_send_sync_threadless_wasm32` or something

It's quite a big change actually, I removed a lot of those `Sendable`, I might almost rewrite the PR from scratch.
May I ask why this is necessary?

wgpu-types/src/lib.rs
daxpedda: Right, I forgot the context here.
The reason why I had `#[doc(hidden)]` here was because this is a public module.
See https://github.com/gfx-rs/wgpu/pull/3691#discussion_r1171754933.

daxpedda:
> > The main change this needs is to have a feature to enable Send/Sync on wasm with a scary name like `unsound_send_sync_threadless_wasm32` or something
> 
> It's quite a big change actually, I removed a lot of those `Sendable`, I might almost rewrite the PR from scratch. May I ask why this is necessary?

Just a suggestion, not less work, instead of a crate feature, we could just let `Send` and `Sync` be implemented if `not(feature = "atomics")`.

Another one, we could do that this feature only work on `not(feature = "atomics")` targets, which makes the most sense actually.

Though the disadvantage is that some libraries might rely on this and wouldn't be compatible with atomics then.

daxpedda:
I added the crate feature. It's all very verbose and I should probably adjust the CI to cover it.
Some documentation of this feature would also be nice.

But before I continue, WDYT @cwfitzgerald?

cwfitzgerald:
I think this situation is fine, though I'm not sure the feature is technically unsound in this case, as you require atomics to have threads at all. 

I'm torn as I think we probably should make it unconditional on atomics, as if someone enables this feature, you get in that same "downstream someone breaks it", but maybe that's a good thing if it would lead to unsoundness.

daxpedda:
> I think this situation is fine, though I'm not sure the feature is technically unsound in this case, as you require atomics to have threads at all.

True, this isn't unsound then. Do you want me to rename the crate feature?

EDIT: I'm gonna deal with CI and docs tmrw.

cwfitzgerald:
Maybe rename it fragile instead of unsound. Thoughts?

daxpedda:
Yeah, that sounds fine!

daxpedda:
I renamed the crate feature to `fragile-send-sync-non-atomic-wasm`.

.github/workflows/ci.yml
```diff
@@ -138,6 +138,7 @@ jobs:
           set -e
 
           # build for WebGPU
+          cargo clippy --target ${{ matrix.target }} --tests --features glsl,spirv,fragile-send-sync-non-atomic-wasm
```
daxpedda, line 141: I would like to add testing with `target-feature=+atomics` as well, but that would require Rust nightly.
Should I add another matrix entry for that?

daxpedda:
I couldn't find a central place where all crate features were documented, would like some guidance on where to document this new crate feature.

.github/workflows/ci.yml
cwfitzgerald: I honestly wouldn't bother on it right now, we need a central place to document this, likely in the readme.

wgpu-core/src/command/bundle.rs
```diff
@@ -746,7 +746,21 @@ pub struct RenderBundle<A: HalApi> {
     pub(crate) life_guard: LifeGuard,
 }
 
+#[cfg(any(
```
cwfitzgerald, line 749: Do we actually need to do this for wgpu-core and hal? Or would forcing things at the `wgpu` layer work?

cwfitzgerald (changes requested):

.github/workflows/ci.yml
daxpedda: Alright.

wgpu-core/src/command/bundle.rs
daxpedda: Maybe? I assume it was implemented initially here for a reason.
I can try and play around with that idea.

daxpedda: Well, after scrolling through some of the places where this is done, there might be some types where this isn't necessary, but there are plenty of other implementations that require this, like callback types.

So I would say maybe, but we couldn't remove the crate feature from `core` or `hal` completely anyway.

Basically what I did in this PR is just gate the already existing `Send` and `Sync` implementations, I assumed they were there for a good reason. If they aren't then I would argue we can figure this out in a different PR, it takes me quite a bit of time to understand the internals enough to figure out if this was unnecessary or not.

cwfitzgerald: Alright, let's keep it as is then, was just trying to reduce conflicts with arcinization.

daxpedda:
Rebased.

cwfitzgerald (approved):
Thanks so much for the persistence on this!

expenses:
Thanks for adding the `fragile-send-sync-non-atomic-wasm` feature! I just tried this and it broke a ton of stuff as bevy components need to be send + sync: https://docs.rs/bevy_ecs/latest/bevy_ecs/component/trait.Component.html.

daxpedda:
I guess bevy would need https://github.com/bevyengine/bevy/issues/6689 to support this properly.

FYI `fragile-send-sync-non-atomic-wasm` doesn't work with `target-feature = "+atomics"`, which requires nightly anyway.

PR #3704, opened 2023-04-19
Change AdapterInfo::vendor to string and add AdapterInfo::vendor_id

i509VCB:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**

**Description**
Start to make AdapterInfo match more closely to GPUAdapterInfo. The old `vendor` field was renamed to `vendor_id`. 

Note that looking up the vendor name is not implemented in all backends yet (Web, DX12 and Vulkan).

### DIFF
--- a/wgpu-hal/src/metal/mod.rs
+++ b/wgpu-hal/src/metal/mod.rs
@@ -109,11 +109,13 @@ impl crate::Instance<Api> for Instance {
             .into_iter()
             .map(|dev| {
                 let name = dev.name().into();
+                let vendor = dev.vendor().into();
                 let shared = AdapterShared::new(dev);
                 crate::ExposedAdapter {
                     info: wgt::AdapterInfo {
                         name,
-                        vendor: 0,
+                        vendor,
+                        vendor_id: 0,
                         device: 0,
                         device_type: shared.private_caps.device_type(),
                         driver: String::new(),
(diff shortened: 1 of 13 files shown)


wgpu-hal/src/metal/mod.rs
```diff
@@ -109,11 +109,13 @@ impl crate::Instance<Api> for Instance {
             .into_iter()
             .map(|dev| {
                 let name = dev.name().into();
+                let vendor = dev.vendor().into();
```
i509VCB, line 112: I guess this is a private API?

wgpu-types/src/lib.rs
```diff
@@ -1303,11 +1303,17 @@ pub enum DeviceType {
...
+    /// This may be empty if not available.
+    pub vendor: String,
+    /// Vendor PCI id of the adapter.
+    ///
+    /// This value is not available on web.
     ///
     /// If the vendor has no PCI id, then this value will be the backend's vendor id equivalent. On Vulkan,
     /// Mesa would have a vendor id equivalent to it's `VkVendorId` value.
-    pub vendor: usize,
+    pub vendor_id: usize,
```
cwfitzgerald, line 1316: I think we might consider:

```rust
struct VendorInfo {
    id: Option<u32>,
    string: Option<String>,
}
```

Then use option to be clear which information is actually there and which isn't

cwfitzgerald:

Issue #3457, opened 2023-02-06
GL Vertex Buffers Not Reset After Ending Renderpass


cwfitzgerald:
**Description**

If you end a renderpass with N vertex buffers bound, then try to start one with N-1 bound, you'll get GL errors as you have extra buffers bound.

Working on a fix + regression test. Just getting an issue number for the regression test.

PR #3706, opened 2023-04-19
Fix unsetting vertex attributes in gles backend

Azorlogh:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
https://github.com/bevyengine/bevy/issues/5732 (and all linked issues)

**Description**
Currently, the code responsible for un-setting vertex attributes uses the attribute's index in the state's array.
Instead we need to use the location of the vertex attribute.
This was causing issues with apps running in wasm on the webgl backend, such as bevy apps.

**Testing**
I tweaked [the test from issue #3457](https://github.com/gfx-rs/wgpu/blob/trunk/wgpu/tests/regression/issue_3457.rs) to use non-consecutive vertex attribute locations. This covers the code path that unsets attribs at the end of a render pass, but does not cover the code path that unsets them upon switching to a new pipeline.
Unfortunately I didn't manage to test that one (unsetting upon pipeline switching), because the webgl context fails silently in that case.


cwfitzgerald (approved):
Great!

PR #3739, opened 2023-04-30
Add a link to the WebGPU C++ programming guide

eliemichel:
This is just a suggestion, my apologies if it feels to much like advertisement!

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,9 @@ If you are looking for a wgpu tutorial, look at the following:
 
 To use wgpu in C/C++, you need [wgpu-native](https://github.com/gfx-rs/wgpu-native).
 
+If you are looking for a wgpu C++ tutorial, look at the following:
+- https://eliemichel.github.io/LearnWebGPU/
+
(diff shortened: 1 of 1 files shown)


Wumpf:
Given that there's already a similar link under Rust this seems to be a good addition. But also would be treat to put it into https://github.com/gfx-rs/wgpu/wiki/Learning-wgpu
Personally, I think the readme should just link to the wiki which then details all the different resources and don't refer to either this or sotrh directly.

cwfitzgerald:
> Personally, I think the readme should just link to the wiki which then details all the different resources and don't refer to either this or sotrh directly.

This is a larger discussion but every time we require someone to jump through the hoop of a link, we will lose people, and makes finding the right place to be just a little bit harder. I think in general it's better to duplicate information more than to aggressively follow DRY.

All that said, not a discussion for this issue :)

cwfitzgerald:
I don't know why checks didn't trigger, going to just merge as it's a docs only change.

PR #3744, opened 2023-05-03
web: use `get_preferred_canvas_format()` to fill `SurfaceCapabilities`

jinleili:
**Checklist**

- [X] Run `cargo clippy`.

**Description**
On the web, wgpu always prioritizes returning the `Bgra8Unorm` format in the `formats` field of `SurfaceCapabilities`, but this format is not supported on Android devices:
截屏2023-05-03 19 44 17

**Testing**
Tested on Android Chrome Canary 115.0.5747.0 via https://jinleili.github.io/learn-wgpu-zh/simuverse
截屏2023-05-03 19 58 50

### DIFF
--- a/wgpu/src/backend/web.rs
+++ b/wgpu/src/backend/web.rs
@@ -1114,14 +1114,20 @@ impl crate::context::Context for Context {
         _adapter_data: &Self::AdapterData,
     ) -> wgt::SurfaceCapabilities {
         let format = self.0.get_preferred_canvas_format();
-        let preferred_format = if map_texture_format(wgt::TextureFormat::Bgra8Unorm) == format {
-            wgt::TextureFormat::Bgra8Unorm
+        let (preferred, second) = if map_texture_format(wgt::TextureFormat::Bgra8Unorm) == format {
+            (
+                wgt::TextureFormat::Bgra8Unorm,
+                wgt::TextureFormat::Rgba8Unorm,
+            )
         } else {
-            wgt::TextureFormat::Rgba8Unorm
+            (
+                wgt::TextureFormat::Rgba8Unorm,
+                wgt::TextureFormat::Bgra8Unorm,
+            )
         };
         wgt::SurfaceCapabilities {
             // https://gpuweb.github.io/gpuweb/#supported-context-formats
-            formats: vec![preferred_format, wgt::TextureFormat::Rgba16Float],
+            formats: vec![preferred, second, wgt::TextureFormat::Rgba16Float],
             // Doesn't really have meaning on the web.
             present_modes: vec![wgt::PresentMode::Fifo],
             alpha_modes: vec![wgt::CompositeAlphaMode::Opaque],
--- a/wgpu/src/backend/web.rs
+++ b/wgpu/src/backend/web.rs
@@ -1118,17 +1118,11 @@ impl crate::context::Context for Context {
             wgt::TextureFormat::Bgra8Unorm,
             wgt::TextureFormat::Rgba16Float,
         ];
-        let mapped_formats: Vec<_> = formats
-            .iter()
-            .map(|format| map_texture_format(*format))
-            .collect();
+        let mut mapped_formats = formats.iter().map(|format| map_texture_format(*format));
         // Preferred canvas format will only be either "rgba8unorm" or "bgra8unorm".
         // https://www.w3.org/TR/webgpu/#dom-gpu-getpreferredcanvasformat
         let preferred_format = self.0.get_preferred_canvas_format();
-        if let Some(index) = mapped_formats
-            .iter()
-            .position(|format| *format == preferred_format)
-        {
+        if let Some(index) = mapped_formats.position(|format| format == preferred_format) {
             formats.swap(0, index);
         }
 
--- a/wgpu/src/backend/web.rs
+++ b/wgpu/src/backend/web.rs
@@ -1113,13 +1113,15 @@ impl crate::context::Context for Context {
         _adapter: &Self::AdapterId,
         _adapter_data: &Self::AdapterData,
     ) -> wgt::SurfaceCapabilities {
+        let format = self.0.get_preferred_canvas_format();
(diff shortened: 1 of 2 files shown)


wgpu/src/backend/web.rs
```diff
@@ -1113,13 +1113,15 @@ impl crate::context::Context for Context {
...
+            wgt::TextureFormat::Rgba8Unorm
+        };
         wgt::SurfaceCapabilities {
             // https://gpuweb.github.io/gpuweb/#supported-context-formats
-            formats: vec![
-                wgt::TextureFormat::Bgra8Unorm,
-                wgt::TextureFormat::Rgba8Unorm,
-                wgt::TextureFormat::Rgba16Float,
-            ],
+            formats: vec![preferred_format, wgt::TextureFormat::Rgba16Float],
```
Wumpf: that link says though that _by spec_ those formats need to be in there, so leaving one out is actually not allowed 🤔 
Arguably this means that the browser implementation is broken, something that should be logged and linked here.
In any case it would be nice to at least leave a comment on what happened here and why we don't stick to the spec

Wumpf (approved):
ah well that will do I guess 😄 👍

wgpu/src/backend/web.rs
```diff
@@ -1113,13 +1113,21 @@ impl crate::context::Context for Context {
...
                 wgt::TextureFormat::Rgba8Unorm,
-                wgt::TextureFormat::Rgba16Float,
-            ],
+            )
+        } else {
+            (
+                wgt::TextureFormat::Rgba8Unorm,
+                wgt::TextureFormat::Bgra8Unorm,
+            )
+        };
```
cwfitzgerald: This code won't respect get_preferred_canvas_format if it's `Rgba16Float` or something else entirely. We should create the vector with the preferred canvas format, then insert-if-not-present the other 3 formats.

cwfitzgerald (changes requested):

wgpu/src/backend/web.rs
```diff
@@ -1113,13 +1113,19 @@ impl crate::context::Context for Context {
         _adapter: &Self::AdapterId,
         _adapter_data: &Self::AdapterData,
     ) -> wgt::SurfaceCapabilities {
+        use wgt::TextureFormat::{Bgra8Unorm, Rgba16Float, Rgba8Unorm};
+        let preferred_format = self.0.get_preferred_canvas_format();
+        let mut formats = vec![Rgba8Unorm, Bgra8Unorm, Rgba16Float];
+        if let Some(index) = formats
+            .iter()
+            .position(|format| map_texture_format(*format) == preferred_format)
```
cwfitzgerald: Loop invariant code motion _should_ fix this, but let's put the map call outside of the position loop.

wgpu/src/backend/web.rs
```diff
@@ -1113,13 +1113,19 @@ impl crate::context::Context for Context {
...
+        {
+            formats.swap(0, index);
+        }
```
cwfitzgerald, line 1127: If it's not found, it needs to be inserted.

wgpu/src/backend/web.rs
```diff
@@ -1113,13 +1113,19 @@ impl crate::context::Context for Context {
         _adapter: &Self::AdapterId,
         _adapter_data: &Self::AdapterData,
     ) -> wgt::SurfaceCapabilities {
+        use wgt::TextureFormat::{Bgra8Unorm, Rgba16Float, Rgba8Unorm};
```
cwfitzgerald: Let's just use these directly.

cwfitzgerald (changes requested):
Nits, then good!

wgpu/src/backend/web.rs
cwfitzgerald: Nevermind, totally misread it.

wgpu/src/backend/web.rs
jinleili: `preferred_format` is of type `web_sys::GpuTextureFormat`, and there is no existing fn to map it back to `wgt::TextureFormat`. Do I need to add a fn similar to `map_texture_format` in web.rs to handle this mapping?

cwfitzgerald: Ah yeah, I think you do need to add that unfortunately

jinleili: From the [WebGPU Spec](https://www.w3.org/TR/webgpu/#dom-gpu-getpreferredcanvasformat):
> getPreferredCanvasFormat()
Returns an optimal [GPUTextureFormat](https://www.w3.org/TR/webgpu/#enumdef-gputextureformat) for displaying 8-bit depth, standard dynamic range content on this system. Must only return ["rgba8unorm"](https://www.w3.org/TR/webgpu/#dom-gputextureformat-rgba8unorm) or ["bgra8unorm"](https://www.w3.org/TR/webgpu/#dom-gputextureformat-bgra8unorm).

Do we currently have the option to not handle the "not found" case?

cwfitzgerald: Oh well, nevermind then, could you just note this requirement in a comment in the code

jinleili: Sure, it has been added.

wgpu/src/backend/web.rs
```diff
@@ -1113,13 +1113,28 @@ impl crate::context::Context for Context {
...
     ) -> wgt::SurfaceCapabilities {
+        let mut formats = vec![
+            wgt::TextureFormat::Rgba8Unorm,
+            wgt::TextureFormat::Bgra8Unorm,
+            wgt::TextureFormat::Rgba16Float,
+        ];
+        let mapped_formats: Vec<_> = formats
+            .iter()
+            .map(|format| map_texture_format(*format))
+            .collect();
```
cwfitzgerald: Sorry one more nit: this collect is unnecessary, avoiding an allocation.

cwfitzgerald (approved):

cwfitzgerald (approved):
Nice, thanks!

PR #3750, opened 2023-05-05
Publish examples to wgpu.rs on updates to trunk branch instead of gecko

### LINKED ISSUE #3684: Update CI Publishing Jobs for WebGPU Release
We have traditionally only updated the examples on the website when the gecko branch is updated. But because that branch has fallen in to disrepair and WebGPU is about to hit stable on chrome, we should have the builds happen off trunk.

paul-hansen:
**Checklist**

- [x] Run `cargo clippy`. 

**Connections**
Fixes #3684

**Description**
Switched the branch the publish action is using to `trunk` to allow the examples to be kept up to date as requested in #3684 

When testing I ran into the error `rustup: command not found` in the `Main Install Rust WASM target` step, I fixed this by using [dtolnay/rust-toolchain](https://github.com/marketplace/actions/rustup-toolchain-install) which is what the [Bevy CI template](https://github.com/bevyengine/bevy_github_ci_template/blob/main/.github/workflows/release.yaml) uses for installing Rust in an action. Lmk if you would prefer I revert that.

**Testing**
I tested this by installing https://github.com/nektos/act and running this command to mimic a github action runner using a minimal image that still contained the Rust stuff needed.
```shell
act -P ubuntu-latest=ghcr.io/catthehacker/ubuntu:rust-latest -j publish
``` 
Note that it fails at uploading to github pages, which is expected since running locally you wouldn't have access to the secrets needed to do so.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -21,9 +21,7 @@ jobs:
           persist-credentials: false
 
       - name: Install Rust WASM target
-        uses: dtolnay/rust-toolchain@stable
-        with:
-          targets: wasm32-unknown-unknown
+        run: rustup target add wasm32-unknown-unknown
 
       - name: Install wasm-bindgen-cli
         run: cargo install wasm-bindgen-cli --version=0.2.84
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -3,7 +3,7 @@ name: Publish
 on:
   push:
     branches:
-      - gecko
+      - trunk
 
 env:
   CARGO_INCREMENTAL: false
@@ -21,10 +21,12 @@ jobs:
           persist-credentials: false
 
       - name: Install Rust WASM target
-        run: rustup target add wasm32-unknown-unknown
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: wasm32-unknown-unknown
 
       - name: Install wasm-bindgen-cli
-        run: cargo install wasm-bindgen-cli --version=0.2.83
+        run: cargo install wasm-bindgen-cli --version=0.2.84
 
       - name: Build WebGPU examples
         run: cargo build --package wgpu --release --target wasm32-unknown-unknown --examples
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -20,6 +20,11 @@ jobs:
         with:
           persist-credentials: false
 
+      - name: Remove hardcoded MSRV
+        # Okay to build examples using the latest Rust.
+        # Fixes issues building wasm-bindgen-cli
+        run: rm -f rust-toolchain
+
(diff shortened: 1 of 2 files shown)


paul-hansen:
CI failure is not related to this PR. The [time crate](https://crates.io/crates/time) updated since our last CI run and [raised it's MSRV to 1.65](https://github.com/time-rs/time/blob/main/CHANGELOG.md#changed) which broke the version of wasm-pack we are using since we compile for `1.64`.

paul-hansen:
Opened a different PR that should fix the CI issue: https://github.com/gfx-rs/wgpu/pull/3754

paul-hansen:
Rebased on the latest trunk to get the CI fix

paul-hansen:
Looks like running the publish job locally is failing due to the updated time dependency too. Working on a fix.

cwfitzgerald:
We can absolutely run the publishing job on latest stable. Only normal CI needs to check our MSRV

paul-hansen:
Ah cool, I was thinking of using `taiki-e/install-action` to install cargo-binstall and use that to install wasm-bindgen-cli. But I was getting a bunch of too many requests errors using binstall so changing it to not use the msrv might be simpler.

.github/workflows/publish.yml
```diff
@@ -20,11 +20,16 @@ jobs:
...
+        # Okay to build examples using the latest Rust.
+        # Fixes issues building wasm-bindgen-cli
+        run: rm -f rust-toolchain
+
       - name: Install Rust WASM target
         run: rustup target add wasm32-unknown-unknown
 
       - name: Install wasm-bindgen-cli
-        run: cargo install wasm-bindgen-cli --version=0.2.83
+        run: cargo install wasm-bindgen-cli --version=0.2.84
```
paul-hansen: Note that this was also updated to match our wasm-bindgen version:
https://github.com/gfx-rs/wgpu/blob/3432aabab158b2183d4eb2b6f58b74639acad2f3/Cargo.toml#L120

.github/workflows/publish.yml
```diff
@@ -20,11 +20,16 @@ jobs:
         with:
           persist-credentials: false
 
+      - name: Remove hardcoded MSRV
+        # Okay to build examples using the latest Rust.
+        # Fixes issues building wasm-bindgen-cli
+        run: rm -f rust-toolchain
```
cwfitzgerald: You can do `cargo +stable install wasm-bindgen-cli` to override it for a specific command

paul-hansen: Nice! I've used +nightly for things before but I hadn't connected the dots that I could use it to override the rust-toolchain file (that file is also new to me). Learned some things, thanks!

cwfitzgerald (approved):
Thanks so much!

PR #3743, opened 2023-05-02
fix(wgpu-core)!: use `u32` indices for bind group layouts everywhere

ErichDonGubler:
**Checklist**

- [x] Run `cargo clippy`.
- [x] ~Run `RUSTFLAGS=--cfg=web_sys_unstable_apis cargo clippy --target wasm32-unknown-unknown` if applicable.~
- [x] Add change to CHANGELOG.md. See simple instructions inside file.

**Connections**
_Link to the issues addressed by this PR, or dependent PRs in other repositories_

See `Description`.

**Description**
_Describe what problem this is solving, and how it's solved._

See discussions in [Mozilla's Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1813705#c8) and [`wgpu`'s Matrix chat](https://matrix.to/#/!FZyQrssSlHEZqrYcOb:matrix.org/$0wv-PJgyTTZ7LHtEhzFmAouNYa4VMY0MzVzW0n5J394?via=matrix.org&via=mozilla.org&via=kde.org).

**Testing**
_Explain how this change is tested._

`cargo nextest run -- --workspace` WFM!

PR #3726, opened 2023-04-24
Fix crash on command buffer drop

Wumpf:
**Checklist**

- [x] Run `cargo clippy`.
- [ ] Run `RUSTFLAGS=--cfg=web_sys_unstable_apis cargo clippy --target wasm32-unknown-unknown` if applicable.
- [x] Add change to CHANGELOG.md. See simple instructions inside file.

**Connections**

Fixes #3724

**Description**

Oversight in: #3657
Passed the wrong object was passed for `data` to `command_buffer_drop`, causing type conversion assertion to fail at runtime 😱 

**Testing**

Add minimal repro
```rust
let encoder = device.create_command_encoder(&wgpu::CommandEncoderDescriptor { label: None });
drop(encoder.finish());
```
to hello triangle. Before: `thread 'main' panicked at 'assertion failed: data.is::<T>()', wgpu/src/context.rs:1045:5`

PR #3718, opened 2023-04-21
Pass correct surface for existing canvases

grovesNL:
**Checklist**

- [ ] Run `cargo clippy`.
- [ ] Run `RUSTFLAGS=--cfg=web_sys_unstable_apis cargo clippy --target wasm32-unknown-unknown` if applicable.
- [ ] Add change to CHANGELOG.md. See simple instructions inside file.

**Connections**
Fixes #3710

**Description**
We were passing the `(id, data)` tuple instead of `data` which was expected. This was hard to spot because we use `Any` for surface types, so it eventually results in a `downcast_ref` much later. We should try to improve type safety here after we figure out how `Arc` will be integrated with the WebGPU backend.

**Testing**
Tested by modifying hello-triangle to create the surface from a canvas instead of a handle.

PR #3774, opened 2023-05-14
Backport bugfixes to v0.16

Wumpf:
**Checklist**

- [x] Run `cargo clippy`.

**Description**

Backports

* #3718 
* #3726
* #3743

all of which are crash fixes without any apparent api breackage.

**Testing**

It compiles and the only thing conflicting was the changelog

### DIFF
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -697,7 +697,7 @@ pub enum BindError {
         s1 = if *.actual >= 2 { "s" } else { "" },
     )]
     MismatchedDynamicOffsetCount {
-        group: u8,
+        group: u32,
         actual: usize,
         expected: usize,
     },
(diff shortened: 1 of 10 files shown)


Wumpf:
I think this one should also go in here https://github.com/gfx-rs/wgpu/pull/3744
There might be more :)

emilk:
I also merged in this fix:
* https://github.com/gfx-rs/wgpu/pull/3780

wgpu-core/src/binding_model.rs
```diff
@@ -697,7 +697,7 @@ pub enum BindError {
         s1 = if *.actual >= 2 { "s" } else { "" },
     )]
     MismatchedDynamicOffsetCount {
-        group: u8,
+        group: u32,
```
cwfitzgerald, line 700: Oi, this is, technically, a semver breaking change. 

I think we could validly argue that while it is breaking, no one is going to be looking at it, only calling Display::fmt.

Let's go ahead with it.

cwfitzgerald (approved):

PR #3779, opened 2023-05-17
Fix multiview rendering

### LINKED ISSUE #3778: Add a feature gate to enable XR/Multiview rendering
As mentioned here https://github.com/gpuweb/gpuweb/issues/4109 the webgpu spec specifically does not allow for multiview rendering because of it's validation of 2D texture and 1 array layer of the render target texture. In order to be able to use WGPU to make XR applications I forked WGPU here with an example of how to resolve this issue. https://github.com/MalekiRe/wgpu and used it here in an openxr-wgpu example https://github.com/MalekiRe/wgpu-openxr-example

If we could have some form of disable for the validation as part of wgpu, because it seems like multiview *is* something eventually intended, that would be awesome.

MalekiRe:
**Checklist**

- [ x] Run `cargo clippy`.
- [ x] Add change to CHANGELOG.md. See simple instructions inside file.

**Connections**
fixes https://github.com/gfx-rs/wgpu/issues/3778

**Description**
WGPU Multiview works again by disabling some validation of TextureViewDimension and ResolvedArrayLayerCount

**Testing**
Running WGPU with a multiview enabled and trying it with multiple views.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -1189,7 +1189,10 @@ impl<A: HalApi> Device<A> {
                 break 'b Err(TextureViewNotRenderableReason::Usage(texture.desc.usage));
             }
 
-            if resolved_dimension != TextureViewDimension::D2 {
+            if !(resolved_dimension == TextureViewDimension::D2
+                || (self.features.contains(wgt::Features::MULTIVIEW)
+                    && resolved_dimension == TextureViewDimension::D2Array))
+            {
                 break 'b Err(TextureViewNotRenderableReason::Dimension(
                     resolved_dimension,
                 ));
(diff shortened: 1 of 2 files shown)


MalekiRe:
@cwfitzgerald This is a PR to do the fix for the thing in the issue

Wumpf (approved):
lgtm!

wgpu-core/src/device/mod.rs
```diff
@@ -1189,7 +1189,9 @@ impl<A: HalApi> Device<A> {
                 break 'b Err(TextureViewNotRenderableReason::Usage(texture.desc.usage));
             }
 
-            if resolved_dimension != TextureViewDimension::D2 {
+
+            if resolved_dimension != TextureViewDimension::D2
+                && !(self.features.contains(wgt::Features::MULTIVIEW)){
```
teoxoy: I think this should still check that the view dimension is `D2Array` if the feature is available instead of allowing all through.

teoxoy (dismissed):
Thanks for the PR! Besides the comment it looks good!

MalekiRe:
@teoxoy Made the change

MalekiRe:
Do not merge this yet! I messed up the logic with the 2DTextureView to be inverted with the fix

MalekiRe:
Fixed, feel free to do checks, resolve the issue and merge

cwfitzgerald (approved):

teoxoy (approved):

PR #3783, opened 2023-05-18
Fix order of arguments to glPolygonOffset.

komadori:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
While investigating a Z-fighting issue shown in https://github.com/komadori/bevy_mod_outline/issues/19 which is apparently specific to WebGL, I noticed that wgpu-hal's GL ES back-end appears to be passing the wrong arguments to glPolygonOffset. However, this bug is ultimately largely incidental to that issue.

**Description**
[glPolygonOffset](https://registry.khronos.org/OpenGL-Refpages/es3.0/html/glPolygonOffset.xhtml) takes two arguments, `factor` and `units`. The terminology used is slightly different to wgpu, but I believe `factor` ("_specifies a scale factor_") should correspond to `slope_scale` and `units` ("_create a constant depth offset_") to `constant`. However, wgpu passes these arguments in the reverse order and this PR corrects that.

**Testing**
In practice, both the `slope_scale` and `constant` are often set to around 1 so wgpu swapping them would go unnoticed. I don't have a meaningful test case for this and am judging it correct by inspection of the code and OpenGL documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,8 @@ Bottom level categories:
 
 ### Bug Fixes
 
+- Fix order of arguments to glPolygonOffset by @komadori in [#3783](https://github.com/gfx-rs/wgpu/pull/3783).
+
 #### General
 
 - Fix crash on dropping `wgpu::CommandBuffer`. By @wumpf in [#3726](https://github.com/gfx-rs/wgpu/pull/3726).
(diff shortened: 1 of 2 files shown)


cwfitzgerald (commented):
Looks great! Thank you

cwfitzgerald (approved):

PR #3626, opened 2023-03-30
Arcanization of wgpu core resources

gents83:
**Connections**

Closes https://github.com/gfx-rs/wgpu/issues/2710
Closes https://github.com/gfx-rs/wgpu/issues/4139
Closes https://github.com/gfx-rs/wgpu/issues/3193
Closes https://github.com/gfx-rs/wgpu/issues/4686

**Steps done:**

- Removed Token and LifeTime related management
- Removed RefCount and MultiRefCount in favour of using only Arc internal reference count
- Removing mut from resources and added instead internal members locks on demand or atomics operations
- Resources now implement Drop and destroy stuff when last Arc resources is released
- Resources hold an Arc<Device> in order to be able to implement Drop
- Resources have an utility to retrieve the id of the resource itself
- Remove all guards and just retrive the Arc needed on-demand to unlock registry of resources asap
- Verify correct resources release when unused or not needed
- Check Web and Metal compliation (thanks to @niklaskorz)
- Fix tests on all platforms
- Test a multithreaded scenario
- Storage is now holding only user-land resources, but Arc is keeping refcount for resources
- When user unregister a resource, it's not dropped if still in use due to refcount inside wgpu
- IdentityManager is now unique and free is called on resource drop instead of storage unregister
- Identity changes due to Arcanization and Registry being just the user reference
- Added MemLeaks test and fixing mem leaks

STATUS:
- Ready to be REVIEWED and INTEGRATED 👍

PR #3796, opened 2023-05-23
Pull out code motion from arcanization

jimblandy:
The arcanization PR #3626 includes a bunch of code motion, in particular breaking `wgpu-core/src/hub.rs` into several different modules. This introduces a lot of noise around imports and paths, and makes it hard to see what has been done to the contents of those modules.

This PR splits out the code motion. It's easy to 'merge' this into the real arcanization branch simply by:
1) merging `trunk` into `gents83/master` as has been done several times before
2) rebasing this PR onto the same `trunk` commit that was just merged into `gents83/master` and committing that as the new `trunk`
3) 'merging' that `trunk` into `gents83/master` in a way that accepts no changes from `trunk`

The resulting diff against trunk is substantially easier to review.

This PR is marked as a draft at the moment, since I'd still like to do `registry.rs` and `storage.rs`, and maybe the two new modules in `device`.


gents83:
Thanks for this! It's a very smart way to handle that big CL

jimblandy:
I believe all the commits in this PR are individually CI-safe, so we probably don't need to squash.

jimblandy:
and I think they're all just code motion, path adjustments and `use` tweaks.

teoxoy (approved):
It's unfortunate that git can't detect that part of the code from `device/mod.rs` was moved to `device/global.rs` and blame won't work past the commit in this PR but I don't see a solution for it and that file was getting increasingly large.

Otherwise looks good!

Issue #3333, opened 2022-12-26
Audit D3D12 Features and Limits wrt Resource Binding Tiers


cwfitzgerald:
We should go through the limits and feature code for d3d12 - specifically looking at the implications of resource binding tiers, partial binding, etc

https://learn.microsoft.com/en-us/windows/win32/direct3d12/hardware-support

- We can assume FL11_1 for WebGPU purposes. So this can be set to `64 / 2 shader stages / 2 types of UAVs` = 16 minimum for both `max_storage_textures_per_shader_stage` and `max_storage_buffers_per_shader_stage`

PR #3798, opened 2023-05-23
DX12 increase `max_storage_buffers_per_shader_stage` and `max_storage_textures_per_shader_stage`

Elabajaba:
**Checklist**

- [ ] Run `cargo clippy`.

**Connections**
#3333 

**Description**
DX12's `max_storage_buffers_per_shader_stage` and `max_storage_textures_per_shader_stage` were set too low. This changes it to use the limits laid out in https://learn.microsoft.com/en-us/windows/win32/direct3d12/hardware-support

WebGPU minimum is FeatureLevel(FL) 11.0 + resource binding tier 2, or FL11.1 which is 64 UAVs (https://github.com/gpuweb/gpuweb/issues/1069), so I'm not sure if this should error if we're FL11.0 + tier 1.

**Testing**
It didn't crash when I ran the tests or the water example with WGPU_BACKEND="dx12".

### DIFF
--- a/wgpu-hal/src/dx12/adapter.rs
+++ b/wgpu-hal/src/dx12/adapter.rs
@@ -316,8 +316,8 @@ impl super::Adapter {
                     // these both account towards `uav_count`, but we can't express the limit as as sum
                     // of the two, so we divide it by 4 to account for the worst case scenario
                     // (2 shader stages, with both using 16 storage textures and 16 storage buffers)
-                    max_storage_buffers_per_shader_stage: (uav_count / 4).min(8),
-                    max_storage_textures_per_shader_stage: (uav_count / 4).min(8),
+                    max_storage_buffers_per_shader_stage: uav_count / 4,
+                    max_storage_textures_per_shader_stage: uav_count / 4,
                     max_uniform_buffers_per_shader_stage: full_heap_count,
                     max_uniform_buffer_binding_size:
                         d3d12_ty::D3D12_REQ_CONSTANT_BUFFER_ELEMENT_COUNT * 16,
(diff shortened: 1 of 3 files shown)


Elabajaba:
~~Feature level detection seems to be broken, don't merge as is.~~

Edit: Fixed

wgpu-hal/src/dx12/adapter.rs
```diff
@@ -284,9 +314,10 @@ impl super::Adapter {
                         _ => d3d12_ty::D3D12_MAX_SHADER_VISIBLE_SAMPLER_HEAP_SIZE,
                     },
                     // these both account towards `uav_count`, but we can't express the limit as as sum
-                    max_storage_buffers_per_shader_stage: base.max_storage_buffers_per_shader_stage,
-                    max_storage_textures_per_shader_stage: base
-                        .max_storage_textures_per_shader_stage,
+                    // of the two, so we divide it by 4 to account for the worst case scenario
+                    // (2 shader stages, with both using 16 storage textures and 16 storage buffers)
+                    max_storage_buffers_per_shader_stage: (uav_count / 4).min(8),
+                    max_storage_textures_per_shader_stage: (uav_count / 4).min(8),
```
teoxoy: I think we shouldn't force the minimum to be 8 here.

> so I'm not sure if this should error if we're FL11.0 + tier 1

We will if the limits passed at device creation are not satisfied (`Limits::default()` correspond to WebGPU's minimum required limits).

https://github.com/gfx-rs/wgpu/blob/4478c52debcab1b88b80756b197dc10ece90dec9/wgpu-core/src/instance.rs#L347-L349

teoxoy (changes requested):

wgpu-hal/src/dx12/adapter.rs
Elabajaba: So I should just remove the min and let it fail for devices that only support 8 UAVs?

JMS55 (changes requested):
The doc comment on `max_storage_textures_per_shader_stage` says the default is 8, but according to the spec and wgpu's code it should be 4.

wgpu-hal/src/dx12/adapter.rs
teoxoy: Yeah, as a user you can always lower the limits you pass to device creation if you really want the app to run on older hardware.

teoxoy (approved):
LGTM

Issue #3206, opened 2022-11-14
Standardize our usage of logging levels


ErichDonGubler:
**Description**

While @emilk was investigating a (separate) crashing issue with `wgpu`, they [reported](https://matrix.to/#/!FZyQrssSlHEZqrYcOb:matrix.org/$ZEPKl2e324q_kMIF3Y2F8nDMlvmCIpQ1-pRUKgwTUI4?via=matrix.org&via=mozilla.org&via=kde.org) that `wgpu-hal` and `wgpu-core`'s logging at the `INFO` level were so verbose as to hinder debugging.

**Repro steps**

See [this comment](https://github.com/gfx-rs/wgpu/issues/3206#issuecomment-1314258786).

**Expected vs observed behavior**

In [@emilk's words](https://matrix.to/#/!FZyQrssSlHEZqrYcOb:matrix.org/$MbcDUlQxxR3EPXvP-SaomLl-NRC-zTHoFiVwwwPIR18?via=matrix.org&via=mozilla.org&via=kde.org):

> I think INFO should be used sparingly to inform users of important and rare events. Currently I cannot even resize my window without four log lines for each resize event:
> 
> ```
> 2022-11-11T13:18:11.813326Z  INFO wgpu_core::device: configuring surface with SurfaceConfiguration { usage: RENDER_ATTACHMENT, format: Bgra8Unorm, width: 1889, height: 2135, present_mode: AutoVsync, alpha_mode: Auto }    
> 2022-11-11T13:18:11.813387Z  INFO wgpu_core::device: Automatically choosing presentation mode by rule AutoVsync. Chose Fifo    
> 2022-11-11T13:18:11.813391Z  INFO wgpu_core::device: Automatically choosing alpha mode by rule Auto. Chose Opaque    
> 2022-11-11T13:18:11.813397Z  INFO wgpu_hal::metal::surface: build swapchain SurfaceConfiguration { swap_chain_size: 3, present_mode: Fifo, composite_alpha_mode: Opaque, format: Bgra8Unorm, extent: Extent3d { width: 1889, height: 2135, depth_or_array_layers: 1 }, usage: COLOR_TARGET }
> ```
> 
> This is not interesting to me. This should be at the TRACE or DEBUG level imho.

Out of the box, we'd like `INFO`-level and higher severities of logging to have a high ratio of useful information.

**Extra materials**

\-

**Platform**

\-

**Remaining action items**

[...]

PR #3800, opened 2023-05-24
Adjust log levels per #3626

jimblandy:
#3626 makes a bunch of minor changes to `log::foo!` macros. I thought it might make it easier to review to pull them out, but it's kind of marginal.

In any case: this PR depends on #3796.

**Checklist**

- [X] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -241,7 +241,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             };
 
             let id = fid.assign(buffer, &mut token);
-            log::info!("Created buffer {:?} with {:?}", id, desc);
+            log::info!("Created Buffer {:?} with {:?}", id, desc);
 
             device
                 .trackers
@@ -461,7 +461,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             //TODO: lock pending writes separately, keep the device read-only
             let (mut device_guard, mut token) = hub.devices.write(&mut token);
 
-            log::info!("Buffer {:?} is destroyed", buffer_id);
+            log::debug!("Buffer {:?} is destroyed", buffer_id);
             let (mut buffer_guard, _) = hub.buffers.write(&mut token);
             let buffer = buffer_guard
                 .get_mut(buffer_id)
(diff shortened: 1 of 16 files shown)


jimblandy:
This is just extracting the changes to log calls made in #3626. I'm not sure I have an opinion on whether the changes are good or not.

@gents83, how did you go about choosing the new log levels? Is there a pattern or a philosophy? Did it just make things more useful for you?

jimblandy:
fixed up authorship of the commit

gents83:
> This is just extracting the changes to log calls made in #3626. I'm not sure I have an opinion on whether the changes are good or not.
> 
> @gents83, how did you go about choosing the new log levels? Is there a pattern or a philosophy? Did it just make things more useful for you?
> 

You are right, it definitely need and explanation.
I defined that "info" should be everything related to creation/destruction of resources and/or anything that can be useful when running unit tests and fail (that's why I added the log info level in CI too)
"warn" instead is everything that should be a warning for the user if something bad is happening and it should not, and I would say it should be the "normal user" use case.
"debug" instead is the most generic one where you can spam whatever, getting also the previous ones included

As you already pointed out anyway I probably missed some changes or messed them up in a couple of cases, but they should follow the above phylosophy

cwfitzgerald:
Thank you for splitting these out. All the wording changes are fine and any demotions are okay. I am quite apprehensive to _promote_ any warnings, especially when https://github.com/gfx-rs/wgpu/issues/3206 exists and is something we should be striving for.

gents83:
I also fixed the one that I was messing up in #3626 👍

wgpu-core/src/device/global.rs
```diff
@@ -241,7 +241,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             };
 
             let id = fid.assign(buffer, &mut token);
-            log::info!("Created buffer {:?} with {:?}", id, desc);
+            log::info!("Created Buffer {:?} with {:?}", id, desc);
```
ErichDonGubler, line 244: nitpick(non-blocking): This change and others like it, where the casing of a phrase otherwise present before is adjusted, seem insubstantial. I think an adjustment to match type names could be desirable, but it's not consistently implemented here. For example, `compute pipeline` and `command {buffer,encoder}` were missed.

Thoughts on dropping these changes, and optionally bringing them back with more polish would be better?

nical:
Should this be closed and re-extracted from the arcanization PR?

cwfitzgerald:
Most likely - I'm going to close and we can discuss plans for arcinization next maintainers call.

PR #3801, opened 2023-05-24
Implement Send and Sync unconditionally for exposed Id type.

### LINKED ISSUE #3730: #3575 makes `Id` types `!Send` and `!Sync`
There's no meaningful reason I can see for them not to be `Send` and `Sync`, but it's a bit gross to use unsafe impls? Maybe that's the only sensible way to ensure that these auto traits don't depend on the `T`.

Currently, the `Id` type is defined as:

```
pub struct Id<T>(core::num::NonZeroU64, std::marker::PhantomData<*mut T>);
```

which makes it unconditionally `!Send` and `!Sync` because of the `*mut` part.

kyren:
**Checklist**

- [x] Run `cargo clippy`.

Unintentionally changed by #3575, Fixes #3730

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -4263,6 +4263,15 @@ impl Surface {
 #[repr(transparent)]
 pub struct Id<T>(core::num::NonZeroU64, std::marker::PhantomData<*mut T>);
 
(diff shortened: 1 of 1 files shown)


teoxoy (approved):
LGTM, thanks!

PR #3803, opened 2023-05-25
Simplify `ResourceMetadata` and `ResourceMetadataProvider` after arcanization 

jimblandy:
Relative to #3626.

Because #3626 lets `ResourceMetadata` and `ResourceMetadataProvider` hold an `Arc` pointing directly at the resource, there's no need to track the epoch separately. This means that `ResourceMetadataProvider::Resource` is no longer necessary.


gents83:
I've integrated directly the change in #3626 then 👍

cwfitzgerald:
Closing as this has been adopted in arcinization

Issue #3207, opened 2022-11-14
Migrate to Windows-rs from winapi


Elabajaba:
**Is your feature request related to a problem? Please describe.**
winapi-rs should probably be considered dead at this point, and Microsoft seems to have settled on [windows-rs](https://github.com/microsoft/windows-rs/) as the officially supported Windows api bindings. (and have deprecated all the other ones they were experimenting with).

**Describe the solution you'd like**
Migrate all the winapi code and dependencies to windows-rs.

**Additional context**
I believe this was previously blocked on a combination of windows-rs being immature, and Mozilla having to get windows-rs approved internally. Windows-rs seems to have settled down a lot, and I'm not sure whether or not Mozilla has okayed it, or if that's even still considered a blocker these days.

jrmuizel:
It seems like there's still a lot of churn happening: https://github.com/microsoft/windows-rs/issues/2156. We don't really want to end up with a bunch of different versions of windows-rs vendored into the mozilla tree so I think we'll need to wait longer for things to settle more.

cwfitzgerald:
Copying what I said on matrix last night for long term transparency on the reasoning for the switch:

> wgpu is going to need more modern dx12 features at some point (including access to new windows 11 apis). winapi, as it stands, doesn't offer bindings to any of these because development died about 2 years ago. winapi is extensable though as they do export their macros. So we can add bindings to these things to d3d12-rs, it just means we have to manually write them.

> Before we commit to that, if it would be possible to upgrade to the new, "en vogue" windows bindings it would be good to do so. If we can't due to one of a variety of reasons, that's fine, we just need to expand our bindings when the time comes for us to have newer features

> From the comments we've gotten from @jimblandy, @kdashg, and jrmuizel it seems like it's not possible at this point - which is fine, we just should commit to our stance while windows-rs, moz, etc all figure out what's going on wrt windows bindings

cwfitzgerald:
There's also a standalone PR https://github.com/retep998/winapi-rs/pull/812 that never got merged that we could yank.

Elabajaba:
Windows-rs maintainers have stated they're going to slow down their cadence and release new `windows` versions "as necessary", instead of a regular monthly release (but still at a faster rate than the every ~6 months of `windows-sys`).

[...]

PR #3807, opened 2023-05-27
Bump `khronos-egl` to v6

daxpedda:
Nothing changed except the `libloading` dependency: https://github.com/timothee-haudebourg/khronos-egl/blob/5.0.0/CHANGELOG.md#500.

This is done in the same spirit as https://github.com/gfx-rs/wgpu/pull/3711 because `khronos-egl` v5 also depends on `libloading` v0.8, which in turn depends on `windows-sys`, which `wgpu` doesn't want to commit to yet: https://github.com/gfx-rs/wgpu/issues/3207.

Requires https://github.com/timothee-haudebourg/khronos-egl/pull/23.
Replaces https://github.com/gfx-rs/wgpu/pull/3797.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -115,7 +115,7 @@ winapi = "0.3"
 hassle-rs = "0.10.0"
 
 # Gles dependencies
-khronos-egl = "4.1"
+khronos-egl = "6"
 glow = "0.12.3"
 glutin = "0.29.1"
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1466,12 +1466,12 @@ dependencies = [
 
 [[package]]
 name = "khronos-egl"
-version = "4.1.0"
+version = "6.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8c2352bd1d0bceb871cb9d40f24360c8133c11d7486b68b5381c1dd1a32015e3"
+checksum = "6aae1df220ece3c0ada96b8153459b67eebe9ae9212258bb0134ae60416fdf76"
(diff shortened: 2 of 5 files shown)


Snowiiii:
Why this taking so long...

cwfitzgerald:
We're waiting on https://github.com/timothee-haudebourg/khronos-egl/pull/23

daxpedda:
Rebased on trunk.

This can be merged now with the `khronos-egl` v6 release. But actually v5 won't compile on Windows.
Should we ask `khronos-egl` to backport the fix (https://github.com/timothee-haudebourg/khronos-egl/pull/23) to a v5.1 release?

cwfitzgerald:
I don't think so, we never shipped v5.

Cargo.toml
```diff
@@ -115,7 +115,7 @@ winapi = "0.3"
 hassle-rs = "0.10.0"
 
 # Gles dependencies
-khronos-egl = "4.1"
+khronos-egl = ">=4.1,<7"
```
cwfitzgerald: This can now be just `7`. Firefox has committed to windows-sys (but not windows) 0.48.

cwfitzgerald (changes requested):

Cargo.toml
daxpedda: v6 you mean.
Done!

daxpedda:
> I don't think so, we never shipped v5.

I meant because this PR does. Now that we just rely on v6 it's not an issue though.

cwfitzgerald (approved):
Amazing!

PR #3808, opened 2023-05-27
Update deno

crowlKats:
Updated #2170.

Had to bump minimum required rust version as deno requires 1.65


jimblandy (approved):
Looks fine to me.

crowlKats:
@jimblandy any clue whats going on with CI? cant replicate these errors locally

cwfitzgerald:
> note: the `aarch64-apple-darwin` target may not be installed

This is the error that happens when the rustup target is not properly installed.

We very explicitly install 1.64 in https://github.com/gfx-rs/wgpu/blob/trunk/.github/workflows/ci.yml#L99, so updating the toolchain file makes it download 1.65 just-in-time, but not including the proper toolchain.

PR #3809, opened 2023-05-28
Don't report violations of VUID-vkCmdEndDebugUtilsLabelEXT-commandBuffer-01912

### LINKED ISSUE #3733: Performing a `RenderPass` inside a debug group results in Vulkan validation errors
**Description**
When a `RenderPass` is performed inside of a `push_debug_group`/`pop_debug_group` pair, the following Vulkan validation error is emitted:

```
[2023-04-27T20:32:37Z ERROR wgpu_hal::vulkan::instance] VALIDATION [VUID-vkCmdEndDebugUtilsLabelEXT-commandBuffer-01912 (0x56146426)]
        Validation Error: [ VUID-vkCmdEndDebugUtilsLabelEXT-commandBuffer-01912 ] Object 0: handle = 0x563ee6e45520, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x56146426 | vkCmdEndDebugUtilsLabelEXT() called without a corresponding vkCmdBeginDebugUtilsLabelEXT first The Vulkan spec states: There must be an outstanding vkCmdBeginDebugUtilsLabelEXT command prior to the vkCmdEndDebugUtilsLabelEXT on the queue that commandBuffer is submitted to (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdEndDebugUtilsLabelEXT-commandBuffer-01912)
[2023-04-27T20:32:37Z ERROR wgpu_hal::vulkan::instance]         objects: (type: COMMAND_BUFFER, hndl: 0x563ee6e45520, name: ?)
```

**Repro steps**
Modify the `hello-triangle` example to add `encoder.push_debug_group("bla")` before and `encoder.pop_debug_group()` after the render pass.

There should not be any validation errors.

**Extra materials**
I'm using the `vulkan-validation-layers` Arch Linux package at version 1.3.243.0-1 (recent versions add more validation, so this might not reproduce with a lower version, although I haven't tried).

**Platform**
Arch Linux, Vulkan, Wayland/X11 both repro.

jimblandy:
As described in [Vulkan-ValidationLayers#5671], the validation layers don't understand debug ranges paired across different command buffers on the same queue, even though the Vulkan spec says:

> An application may open a debug label region in one command buffer and close it in another, or otherwise split debug label regions across multiple command buffers or multiple queue submissions. When viewed from the linear series of submissions to a single queue, the calls to vkCmdBeginDebugUtilsLabelEXT and vkCmdEndDebugUtilsLabelEXT must be matched and balanced.

Until this is fixed, wgpu should ignore this validation error to reduce noise in test runs.

Fixes #3733.

[Vulkan-ValidationLayers#5671]: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5671

**Checklist**

- [X] Run `cargo clippy`.

### DIFF
--- a/wgpu-hal/src/vulkan/instance.rs
+++ b/wgpu-hal/src/vulkan/instance.rs
@@ -14,12 +14,36 @@ unsafe extern "system" fn debug_utils_messenger_callback(
     message_severity: vk::DebugUtilsMessageSeverityFlagsEXT,
     message_type: vk::DebugUtilsMessageTypeFlagsEXT,
     callback_data_ptr: *const vk::DebugUtilsMessengerCallbackDataEXT,
-    _user_data: *mut c_void,
+    user_data: *mut c_void,
 ) -> vk::Bool32 {
-    const VUID_VKSWAPCHAINCREATEINFOKHR_IMAGEEXTENT_01274: i32 = 0x7cd0911d;
     use std::borrow::Cow;
 
     if thread::panicking() {
+        return vk::FALSE;
+    }
+
+    let cd = unsafe { &*callback_data_ptr };
+    let user_data = unsafe { &*(user_data as *mut super::DebugUtilsMessengerUserData) };
+
+    const VUID_VKCMDENDDEBUGUTILSLABELEXT_COMMANDBUFFER_01912: i32 = 0x56146426;
+    if cd.message_id_number == VUID_VKCMDENDDEBUGUTILSLABELEXT_COMMANDBUFFER_01912 {
+        // https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5671
+        // Versions 1.3.240 through 1.3.250 return a spurious error here if
+        // the debug range start and end appear in different command buffers.
+        let khronos_validation_layer =
+            std::ffi::CStr::from_bytes_with_nul(b"Khronos Validation Layer\0").unwrap();
+        if user_data.validation_layer_description.as_ref() == khronos_validation_layer
+            && user_data.validation_layer_spec_version >= vk::make_api_version(0, 1, 3, 240)
+            && user_data.validation_layer_spec_version <= vk::make_api_version(0, 1, 3, 250)
+        {
+            return vk::FALSE;
+        }
+    }
+
+    // Silence Vulkan Validation error "VUID-VkSwapchainCreateInfoKHR-imageExtent-01274"
+    // - it's a false positive due to the inherent racy-ness of surface resizing
+    const VUID_VKSWAPCHAINCREATEINFOKHR_IMAGEEXTENT_01274: i32 = 0x7cd0911d;
+    if cd.message_id_number == VUID_VKSWAPCHAINCREATEINFOKHR_IMAGEEXTENT_01274 {
         return vk::FALSE;
     }
 
@@ -31,8 +55,6 @@ unsafe extern "system" fn debug_utils_messenger_callback(
         _ => log::Level::Warn,
     };
 
-    let cd = unsafe { &*callback_data_ptr };
-
     let message_id_name = if cd.p_message_id_name.is_null() {
         Cow::from("")
     } else {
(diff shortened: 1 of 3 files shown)


FrankenApps:
Here is the correct link to the Issue https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5671 in case anyone is interested.

jimblandy:
> Here is the correct link to the Issue [KhronosGroup/Vulkan-ValidationLayers#5671](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5671) in case anyone is interested.

Fixed, thanks.

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -16,13 +16,30 @@ unsafe extern "system" fn debug_utils_messenger_callback(
...
+
+    const VUID_VKSWAPCHAINCREATEINFOKHR_IMAGEEXTENT_01274: i32 = 0x7cd0911d;
+    const VUID_VKCMDENDDEBUGUTILSLABELEXT_COMMANDBUFFER_01912: i32 = 0x56146426;
+
+    // Silence Vulkan Validation error " VUID-vkCmdEndDebugUtilsLabelEXT-commandBuffer-01912"
+    // This is a bug in the validation layer:
+    // https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5671
+    if cd.message_id_number == VUID_VKCMDENDDEBUGUTILSLABELEXT_COMMANDBUFFER_01912 {
+        return vk::FALSE;
+    }
```
teoxoy: If we can get the version of the validation layers, it would be better to conditionally ignore it.

jimblandy: To narrow down the affected Mesa versions: it seems like the validation was introduced in KhronosGroup/Vulkan-ValidationLayers#5277, merged 2023-2-16.

jimblandy:
Layer properties in KhronosGroup/Vulkan-ValidationLayers@a90644606 (after introduction of 01912 test):

```
LayerProperties {
    layer_name: "VK_LAYER_KHRONOS_validation",
    spec_version: 4206832, // 1.3.240
    implementation_version: 1,
    description: "Khronos Validation Layer",
}
```

jimblandy:
Layer properties in https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/a54744d67 (current main):

```
JIMB: instance layer: LayerProperties {
    layer_name: "VK_LAYER_KHRONOS_validation",
    spec_version: 4206842, // 1.3.250
    implementation_version: 1,
    description: "Khronos Validation Layer",
}
```

jimblandy:
Okay, this version of the PR supplies the validation layer description and version to the debug utils messenger callback, and checks them to ensure that we're only filtering out errors from affected versions of `Vulkan-ValidationLayers`.

But I think it does violence to the API of `vulkan::Instance::from_raw`, so I want to look at it again tomorrow.

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -236,48 +252,67 @@ impl super::Instance {
...
-            let mut severity = vk::DebugUtilsMessageSeverityFlagsEXT::ERROR;
-            if log::max_level() >= log::LevelFilter::Debug {
-                severity |= vk::DebugUtilsMessageSeverityFlagsEXT::VERBOSE;
-            }
-            if log::max_level() >= log::LevelFilter::Info {
-                severity |= vk::DebugUtilsMessageSeverityFlagsEXT::INFO;
-            }
-            if log::max_level() >= log::LevelFilter::Warn {
-                severity |= vk::DebugUtilsMessageSeverityFlagsEXT::WARNING;
+        eprintln!(
```
cwfitzgerald: Just putting a note on this println

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -236,48 +252,67 @@ impl super::Instance {
...
+            "JIMB: Instance::from_raw(debug_utils_user_data = {:?})",
+            debug_utils_user_data
+        );
+        let debug_utils = if let Some(debug_callback_user_data) = debug_utils_user_data {
+            if extensions.contains(&ext::DebugUtils::name()) {
```
cwfitzgerald, line 273: Would be nice to have an info level print about why debug callbacks are disabled.

cwfitzgerald (changes requested):
Would be great to get this in as #3830 is now hitting it.

wgpu-hal/src/vulkan/instance.rs
jimblandy: Yep, that's why it's a draft PR. :)

cwfitzgerald (approved):
Nice!

PR #3828, opened 2023-06-02
Fix trying to use a renderpass inside a compute pass

cwfitzgerald:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Found by #3827 

**Description**

Modifies how compute passes are encoded. This switches to a two-command-buffer setup where fixups and barriers are dealt with in a separate command buffer that is inserted before.

**Testing**

Found by an assertion and error.

### DIFF
--- a/player/tests/data/pipeline-statistics-query.ron
+++ b/player/tests/data/pipeline-statistics-query.ron
@@ -1,5 +1,5 @@
 (
-    features: 0x0000_0000_0000_0100, // PIPELINE_STATISTICS_QUERY
+    features: 0x0000_0001_0000_0000, // PIPELINE_STATISTICS_QUERY
     expectations: [
         (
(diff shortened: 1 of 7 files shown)


player/tests/data/pipeline-statistics-query.ron
```diff
@@ -1,5 +1,5 @@
 (
-    features: 0x0000_0000_0000_0100, // PIPELINE_STATISTICS_QUERY
+    features: 0x0000_0001_0000_0000, // PIPELINE_STATISTICS_QUERY
```
teoxoy, line 2: Maybe we can use the new `Serialize`/`Deserialize` derives that use a textual representation (in a separate PR).

teoxoy (approved):
LGTM

Issue #3622, opened 2023-03-27
Reconsider lavapipe


jimblandy:
The wgpu sources have various comments here and there remarking on how lavapipe doesn't implement Vulkan well enough to fully support WebGPU, but I wonder if that's true any more. Apparently Linux Mesa lavapipe 22.1.2 [passes](https://www.khronos.org/conformance/adopters/conformant-products#submission_696) the Vulkan 1.3 CTS.

We should go through wgpu and see what works now and which caveats still apply. It would sure be nice to have lavapipe for software-based testing.

teoxoy:
The `zero_init_workgroup_mem` test failing in CI (running via swiftshader currently) is becoming more and more prevalent. I wonder how lavapipe compares.

cwfitzgerald:
The `zero_init_workgroup_mem` test failing in CI is our fault not swiftshader's, it's the same thing affecting WARP: https://github.com/gfx-rs/wgpu/issues/3193

This wasn't obvious before as the vulkan validation layers were off. Afaict, there have been no spurious failures from swiftshader (that weren't our fault), but there have been quite a few from lavapipe. We should definitely try it again, but I have a lot more confidence in swiftshader than lavapipe.

PR #3830, opened 2023-06-04
Migrate to nv-flip for image comparison

### LINKED ISSUE #2761: Better Metric for Image Comparison Testing Failure

Due to the highly variable nature of graphics cards, we cannot use exact comparison to compare the images rendered from a test with the sample image. Every vendor does sampling, rendering, etc ever so slightly differently so we need to fuzzy match them.

Currently we just take the difference between each pixel and the expected pixels and count how many pixels are over a certain tolerance. This unfortunately is rather crude and once you have it tuned for real differences in rendering it can ignore real breakages. For example I can totally ruin bunnymark's sampling so the logo is just a blank color and its image comparison tests still match.

Some kind of better image comparison. PSNR is a possibility but I don't really know the possibilities out there.

Keeping the same system. Given the example I gave, this isn't great.

cwfitzgerald:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Resolves #2761 

**Description**

Integrates NVlab's [ꟻLIP](https://github.com/NVlabs/flip) for image comparison, using their various weighted histogram methods. This feels soooo much more stable, thresholds aren't insanity, and new devices seem to consistently fall under threshold.

I also fixed up some other issues with the tests.

**Testing**

I ran the tests on all of my machines' cards and they all pass.

### DIFF
--- a/wgpu/tests/common/image.rs
+++ b/wgpu/tests/common/image.rs
@@ -134,6 +134,7 @@ impl ComparisonType {
 
 pub fn compare_image_output(
     path: impl AsRef<Path> + AsRef<OsStr>,
+    backend: wgpu::Backend,
     width: u32,
     height: u32,
     test_with_alpha: &[u8],
@@ -187,17 +188,17 @@ pub fn compare_image_output(
             all_passed &= check.check(&mut pool);
         }
 
+        let file_stem = reference_path.file_stem().unwrap().to_string_lossy();
         // Determine the paths to write out the various intermediate files
         let actual_path = Path::new(&path).with_file_name(
-            OsString::from_str(
-                &(reference_path.file_stem().unwrap().to_string_lossy() + "-actual.png"),
-            )
-            .unwrap(),
+            OsString::from_str(&format!("{}-{}-actual.png", file_stem, backend.to_str(),)).unwrap(),
         );
         let difference_path = Path::new(&path).with_file_name(
-            OsString::from_str(
-                &(reference_path.file_stem().unwrap().to_string_lossy() + "-difference.png"),
-            )
+            OsString::from_str(&format!(
+                "{}-{}-difference.png",
+                file_stem,
+                backend.to_str(),
+            ))
             .unwrap(),
         );
 
--- a/wgpu/tests/common/image.rs
+++ b/wgpu/tests/common/image.rs
@@ -230,7 +230,7 @@ pub fn compare_image_output(
 
     #[cfg(target_arch = "wasm32")]
     {
-        let _ = (path, width, height, test_with_alpha, checks);
+        let _ = (path, backend, width, height, test_with_alpha, checks);
     }
 }
 
--- a/wgpu/tests/common/image.rs
+++ b/wgpu/tests/common/image.rs
@@ -1,10 +1,4 @@
-use std::{
-    borrow::Cow,
-    ffi::{OsStr, OsString},
-    io,
-    path::Path,
-    str::FromStr,
-};
+use std::{borrow::Cow, ffi::OsStr, io, path::Path};
 use wgpu::util::DeviceExt;
 use wgpu::*;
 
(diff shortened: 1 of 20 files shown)


wgpu/tests/common/image.rs
```diff
@@ -71,80 +65,172 @@ pub fn calc_difference(lhs: u8, rhs: u8) -> u8 {
...
+    }
+    println!("\tMax Value: {:.6}", pool.max_value());
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum ComparisonType {
+    // If the mean error is greater than the given value, the test will fail.
+    Mean(f32),
+    // If the error at the given percentile [0, 1] is greater than the given value, the test will fail.
+    Percentile(f32, f32),
```
nical: nitpick: I'd rather the two fields be named (I don't think I'll remember which is which in a couple of days).

Also I'm not sure I parse the comment: does `[0, 1]` refer to the range of the error?

wgpu/tests/common/image.rs
```diff
@@ -71,80 +65,172 @@ pub fn calc_difference(lhs: u8, rhs: u8) -> u8 {
...
+        println!(
+            "\t      {percentile}%: {:.6}",
+            pool.get_percentile(percentile as f32 / 100.0, true)
+        );
+    }
+    println!("\tMax Value: {:.6}", pool.max_value());
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum ComparisonType {
```
nical, line 107: Here (and/or in the nv-flip doc), it would be useful to explain what the values represent in more details, for example is the error metric a normalized value in the [0, 1] range ? If you know what a reasonable default might be for the mean error, it would be useful to document it here as well.

nical (commented):
Looks good overall. What color space is nvflip expecting the data to be in? It should be documented in the nv-flip crate and we should make sure the tests actually produce pixels in the right color space otherwise we are only giving the illusion of having a principled error metric.

cwfitzgerald:
Filed the color space info here: https://github.com/gfx-rs/nv-flip-rs/issues/1

wgpu/tests/common/image.rs
cwfitzgerald: Yeah - I'll clarify the comments and separate them.

wgpu/tests/common/image.rs
cwfitzgerald: Yeah I'll document - it's hard to describe what it is, it's basically just "from [0, 1] how different these images".

nical (approved):
Thanks!

wgpu/tests/shader/zero_init_workgroup_mem.rs
```diff
@@ -16,14 +16,15 @@ fn zero_init_workgroup_mem() {
...
+            // remove both of these once we get to https://github.com/gfx-rs/wgpu/issues/3193 or
             // https://github.com/gfx-rs/wgpu/issues/3160
             .specific_failure(
                 Some(Backends::DX12),
                 Some(5140),
                 Some("Microsoft Basic Render Driver"),
                 true,
-            ),
+            )
+            .specific_failure(Some(Backends::VULKAN), None, Some("swiftshader"), true),
```
teoxoy, line 27: I just left this comment here https://github.com/gfx-rs/wgpu/issues/3622#issuecomment-1582501473. I guess this works for now though 👍

teoxoy (approved):
LGTM

teoxoy:
@cwfitzgerald there seems to be an unexpected test pass 👀

wgpu/tests/shader/zero_init_workgroup_mem.rs
cwfitzgerald: Addressed that on that issue

cwfitzgerald:
Yeah, that was https://github.com/gfx-rs/wgpu/pull/3809 landing - will remove the failures.

PR #3840, opened 2023-06-06
Allow workspace to be built on wasm

cwfitzgerald:
**Connections**

Fixes long standing issues with RA building to wasm.

**Description**

deno/v8 doesn't build for wasm, so get the crates to conditionally ignore it.


crowlKats (approved):
LGTM

PR #3841, opened 2023-06-06
Move Examples and Tests to Their Own Crates

cwfitzgerald:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Depends on #3830 

**Description**

This does the following transforms to the codebase:

- Split out examples into their own crates, under `examples/*`
- Move example only common code into a crate in `examples/common`
- Move generic tests and common testing code into crate in `tests/`
- Made wgpu-core compile on wasm, so can include `wgpu-core` in the workspace.

These changes result in code that I believe is not only clearer to a new contributor what is happening, but will be much easier for contributors too, as they don't need to worry about permutations of tests so much.

**Testing**

CI is doing most testing, manual testing supplemented


examples/boids/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+name = "wgpu-boids-example"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+description = "wgpu boids example"
+homepage.workspace = true
+repository.workspace = true
+keywords.workspace = true
+license.workspace = true
+publish = false
```
teoxoy: Do we need all these fields? since we won't publish those

teoxoy (approved):
Makes things much clearer 👍

examples/boids/Cargo.toml
cwfitzgerald: Good shout, will trim them down

cwfitzgerald: Had to keep license, version, and edition, the rest (of the .workspace ones) I could rid.

Issue #3680, opened 2023-04-13
Hard to Tell if Test is Skipped


cwfitzgerald:
We have a rather complex set of situations where a test is either skipped or expected to fail. As such we can't use the standard `#[should_panic]` or `#[ignored]` as we need to make that decision at runtime based on the device we're running on.

This can cause problems where the developer is trying to fix a particular problem, runs the tests, sees that they all pass, not realizing that the test they're trying to measure against isn't actually running or has an expected fail on it.

We should find some way to express if something is skipped.

One possibility is to use [`libtest-mimic`](https://docs.rs/libtest-mimic/latest/libtest_mimic/) to allow us to add a "kind" to tests which is then visible to cargo-nextest or cargo test. We likely would need our own macro for tests in that case and would have interesting ramifications for wasm testing via wasm-pack.

cwfitzgerald:
Going to copy here some thinking I had regarding this issue:

> Specifically that we actually can't use wasm-pack test to do what we really want - we do image comparison testing, which requires sending images back and forth from the filesystem and on the web that requires at minimum a webserver. I think the solution is to lean into xtask, make a test harness for the web that communicates back with the webserver.

> for native when something like `cargo xtask test` is run:
> - our xtask tool will enumerate all gpus on the system and get all of their information and serialize this to a file
> - the test harness (using libtest-mimic) will read this file and enumerate which tests are skipped, expected fail, etc. This also gives us a chance to run permutations on tests and make sure we test all devices on the system.
> - as all these decisions are made from pre-prepared data, it can be done extremely quickly, keeping the harness runtime short

PR #53, opened 2019-02-05
Depenency gfx update to 0.1.1

kvark:
It appears that we still can't run on DX12, because it has `update_buffer` as `unimplemented!`. But at least we have the benefit of the empty backend updates.

PR #3844, opened 2023-06-06
Add `xtask` for project workflows, but start small with only `run-wasm`

ErichDonGubler:
Recommended review experience is commit-by-commit.

---

**Checklist**

- [x] Run `cargo clippy`.
- [x] ~Run `RUSTFLAGS=--cfg=web_sys_unstable_apis cargo clippy --target wasm32-unknown-unknown` if applicable.~ N/A

**Connections**

**Description**

[`cargo-xtask`](https://github.com/matklad/cargo-xtask) is a convention for project workflows using a Rust crate that is separate from the main body of code. We've identified some problems that we'd like to solve with `xtask`[^1][^2]. This PR doesn't try to address those problems specifically, but it _does_ take on setting up `xtask` for further work. It also [absorbs `run-wasm`](https://matrix.to/#/!FZyQrssSlHEZqrYcOb:matrix.org/$66P5BiynlkVtvMfQkZGYzFLHrqi5JpN6tzWlAFZj6RA?via=matrix.org&via=mozilla.org&via=kde.org) as an easy first bit of scope.

[^1]: See [this conversation](https://matrix.to/#/!FZyQrssSlHEZqrYcOb:matrix.org/$yYDCpNlw-ayuwLi0D-MC8afZzN_U5HPWUMoo_OHlUjI?via=matrix.org&via=mozilla.org&via=kde.org) about eventually needing to manage test expectations for WebGPU CTS.
[^2]: See this pain point issue: https://github.com/gfx-rs/wgpu/issues/3680

- Add a separate binary crate called `xtask` that is aliased in Cargo with `cargo xtask`.
	- Migrate the functionality of `run-wasm` to `xtask`. Notably, this moves the some pieces of the main workspace's dependency graph to `xtask`.
	- [Current rationale](https://github.com/gfx-rs/wgpu/pull/3844#issuecomment-1579422557) for why using a crate outside of the existing main workspace is okay.
- [x] Hard dependency on https://github.com/rukai/cargo-run-wasm/pull/37.

**Testing**

- [ ] Have not yet tested that `cargo xtask run-wasm` yields the same running results, but it should.


xtask/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "xtask"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+anyhow = "1.0.71"
+# cargo-run-wasm = "0.3.2"
+cargo-run-wasm = { version = "0.3.2", git = "https://github.com/ErichDonGubler/cargo-run-wasm", branch = "expose-args" }
```
ErichDonGubler: - [x] This needs to be resolved upstream before merging, _or_ we need to commit to using my Git branch for now. I don't particularly care which we do right now.

Cargo.toml
```diff
@@ -1,16 +1,15 @@
...
-    "run-wasm"
+	"cts_runner",
+	"deno_webgpu",
+	"dummy",
+	"player",
+	"wgpu",
+	"wgpu-core",
+	"wgpu-hal",
+	"wgpu-info",
+	"wgpu-types",
```
ErichDonGubler: - [x] issue: Shouldn't have changed indentation here.

xtask/Cargo.toml
cwfitzgerald: We can accept a git dep here, it's fine - this is never getting published anyway

.cargo/config.toml
```diff
@@ -1,5 +1,6 @@
 [alias]
 run-wasm = "run --release --package run-wasm --"
```
cwfitzgerald: Need to axe this

ErichDonGubler:
Gonna copy in some discussion from [Matrix](https://matrix.to/#/!FZyQrssSlHEZqrYcOb:matrix.org/$GV94zmCbZJ_xLN56TGgrWIQ070uoK6LmwIpnQeVhubQ?via=matrix.org&via=mozilla.org&via=kde.org) to here, since this is a better place for topical discussion.

---

@cwfitzgerald asked:

> @ErichDonGubler: remind me the benefits/drawbacks of leaving the xtask out of the workspace

I responded:

> Pros:
> 
> -   People who don't use `xtask` don't ever pay the cost of compiling it.
> -   Toolchain-related questions for developer experience (esp. `Cargo.lock`, i.e., using newer versions of things) can be separate lanes from the `wgpu` workspace. Case in point, basically everybody using `wgpu` in production was compiling `run-wasm` whenever they ran `cargo build --workspace`.
> 
> Cons:
> 
> -   Another compile step for people who aren't expecting it can be weird.
>     
>     -   This usually isn't a problem in practice, because it's rare for `xtask` to need recompilation when working on the main body of `wgpu`'s code.
> -   A separate `xtask` crate/workspace can complicate CI a bit with pressure on caching for compute-bound environments like cloud runners. Naga's compiles of `xtask`, for instance, would/will be a handful of minutes more until something is merged [`ErichDonGubler/naga`#53](https://github.com/ErichDonGubler/naga/pull/53).
>     
>     -   Then again, the compiles are also unconditionally paid for in CI anyway currently, so this isn't as bad as it sounds.

xtask/Cargo.toml
ErichDonGubler: Okay! 😁

ErichDonGubler:
@cwfitzgerald: I've updated the OP with more notes explaining why we're doing this. The "further work" mentioned in the OP doesn't seem to have any issues I can link to for WebGPU CTS. Should I file one?

cwfitzgerald (commented):
Yeah, filling an issue about CTS test expectations sounds good, plus putting it in the testing tracking issue

xtask/src/cli.rs
```diff
@@ -0,0 +1,60 @@
...
+        let mut args = Arguments::from_env();
+        log::debug!("parsing args: {args:?}");
+        if args.contains("--help") {
+            eprint!("{HELP}");
+            exit(101);
+        }
+        match (|| -> anyhow::Result<_> {
+            let subcommand = Subcommand::parse(args)?;
+            Ok(Self { subcommand })
+        })() {
```
cwfitzgerald: Immediately invoked lambdas look really weird - what's the purpose for them in this case? Maybe a named function would work better.

xtask/src/cli.rs
```diff
@@ -0,0 +1,60 @@
...
+    pub(crate) subcommand: Subcommand,
+}
+
+impl Args {
+    pub fn parse() -> Self {
+        let mut args = Arguments::from_env();
+        log::debug!("parsing args: {args:?}");
+        if args.contains("--help") {
+            eprint!("{HELP}");
+            exit(101);
```
cwfitzgerald: Remind me what 101 is for?

xtask/src/cli.rs
```diff
@@ -0,0 +1,60 @@
...
+                eprintln!("{:?}", anyhow!(e));
+                exit(1)
+            }
+        }
+    }
+}
+
+pub(crate) enum Subcommand {
+    RunWasm {
+        // TODO: I'd really like for this to not leak. `Vec<OsString>` plz?
```
cwfitzgerald: How does this leak?

xtask/src/cli.rs
```diff
@@ -0,0 +1,60 @@
...
+        args: Arguments,
+    },
+}
+
+impl Subcommand {
+    fn parse(mut args: Arguments) -> anyhow::Result<Subcommand> {
+        args.subcommand()
+            .context("failed to parse subcommand")
+            .and_then(|parsed| match parsed.as_deref() {
```
cwfitzgerald: Could we use regular control flow + ? here (or let-else), it'd be a lot clearer.

cwfitzgerald (changes requested):

xtask/src/cli.rs
ErichDonGubler: Ah, sorry, this is pure muscle memory that I'm glad you called out. [Cargo uses the 101 exit status to indicate that Cargo has failed](https://doc.rust-lang.org/cargo/commands/cargo.html#exit-status). It's become an (informal, AFAIK) exit status for "the Cargo plugin you called failed, not something underneath that it may have been subprocess'ing to".

The thing is, I don't think this habit makes sense here. It's actually debatable whether a `--help` string should return non-zero (`cargo`, for example, returns 0).

I'm happy to change this to whatever, with whatever docs you want. :)

xtask/src/cli.rs
ErichDonGubler: Oh, derp, I never intend `TODO`s to actually ship. 😅 I'm annoyed here that there's an _abstraction_ leak with `pico_args::Arguments`, not a resource leak.

I think that the abstraction leak is fine for now while upstream figures things out. I'm inclined to just delete the `TODO` comment, unless you have a better idea.

cwfitzgerald: Yeah, that's fine - I don't really see much of an issue with this design.

xtask/src/cli.rs
ErichDonGubler: I have replaced it with e9617531.

xtask/src/cli.rs
ErichDonGubler: Changed with f97e6c24—how's it look?

.cargo/config.toml
ErichDonGubler: D'oh, good catch. Fixed with 4428c462.

xtask/src/cli.rs
cwfitzgerald: This is interesting, I called a bunch of things and they seem to disagree.

cargo-related things all return 0 on finish.
unix tools (`find` returns 1, `grep` returns 2??) return non-zero

I kinda prefer returning non-zero as returning `-h` in something like a script/CI is almost always should fail.

xtask/src/cli.rs
cwfitzgerald: I would additionally move the argument to the `match` into it's own variable, but beyond that it looks fine.

xtask/src/cli.rs
cwfitzgerald: Tldr: 101 is fine

cwfitzgerald: Though let's add a bit of documentation about why 101 vs any other number, maybe a const?

cwfitzgerald (approved):
G2G after discussed nits

xtask/src/cli.rs
ErichDonGubler: Moved with 9e38358a.

xtask/src/cli.rs
ErichDonGubler: Removed commentary with 0127ed10.

xtask/src/cli.rs
ErichDonGubler: I avoided `const` so I didn't have to specify a type, but I extracted a variable + doc link with 32dec0f2.

cwfitzgerald (commented):
Yay!

cwfitzgerald (approved):
Yay!

PR #3854, opened 2023-06-10
Bumps wasm-bindgen version

cwfitzgerald:
**Checklist**

- [x] Run `cargo clippy`.

**Description**

Fixes failure in wasm-bindgen job as well as ensure it can't happen again by running it as part of the PR process.


Wumpf (approved):
ah shoot, didn't pay attention to this when merging in the wasm-bindgen update PR.
change makes sense

There is however a (still somewhat annoying) alternative that I recently learned: Instead of running `wasm-bindgen` command, add a build.rs script that uses `wasm-bindgen-cli-support`. That way dependabot can pick it up as well

cwfitzgerald:
Yeah, manging the wasm-bindgen stuff is annoyingly manual.

We still need to bump the version in the wiki every time, so I think this is probably the least annoying solution, as at least CI will catch it.

Issue #3609, opened 2023-03-20
wgpu-types error when using `serde` feature without tracing feature


paul-hansen:
**Description**
The wgpu-types crate returns an error if compiled with the "serde" feature and not the trace feature.

**Repro steps**
```
cargo test --package wgpu-types --features serde
```

**Expected vs observed behavior**
I was expecting it to compile with serde serialization and deserialization of common structs.
Instead it returned many instances of this compile time error :
```
error: cannot find attribute `serde` in this scope                                                                                                                                                
   --> wgpu-types\src\lib.rs:123:31
    |
123 | #[cfg_attr(feature = "serde", serde(rename_all = "kebab-case"))]
    |                               ^^^^^
    |
    = note: `serde` is in scope, but it is a crate, not an attribute
```

**Platform**
- Windows 11
- wgpu master branch, commit `7495646d`

**Addition Info**
My use case was trying to load image sampler settings from a file so I can set the texture to be repeating in Bevy. Some of wgpu's types are exposed by Bevy so I thought I would try turning on serde support to be able to load those values from a file directly. This isn't really a big deal for me, but I thought you might want to know about the inconsistency in the features.

paul-hansen:
This is happening because `Serialize` and `Deserialize` are only derived when the `trace` and `replay` features are enabled. 
https://github.com/gfx-rs/wgpu/blob/7495646d5fec0711e31b56619e94daffc03e5a77/wgpu-types/src/lib.rs#L1608-L1610

Without the Serialize derive, the `serde(rename_all = "kebab-case")` is not valid. Which means the serde feature is fairly useless as it can't be used without one of the other features enabled.

I see a few different options so far:
1. Remove the serde feature, continue to use the trace and replay features to enable Derive and Serialize.
  Things like `#[cfg_attr(feature = "serde", ... ` would need changed to `#[cfg_attr(any(feature = "trace", feature = "replay"), ... `
2. Leave as is, maybe document somewhere that the serde feature shouldn't be used alone. Not ideal but it's not a big problem once you know.
3. Split the serde feature into `serde_serialize` and `serde_deserialize` so the trace and replay features can require them separately, then we can require those features for trace and replay respectively.
  `#[cfg_attr(feature = "trace", derive(Serialize))]` would be changed to `#[cfg_attr(feature = "serde_serialize", derive(Serialize))]` etc.
  
I'd be happy to work up a PR if desired and there's a direction you want to try.

teoxoy:
[...]

PR #3856, opened 2023-06-11
Make `wgpu-info` into a proper CLI Tool

cwfitzgerald:
**Checklist**

- [x] Run `cargo clippy`.

**Description**

One pre-component for https://github.com/gfx-rs/wgpu/issues/3680 is being able to generate a machine-readable report of the user's GPU so that tests can be filtered very quickly based on it.

This reworks wgpu-info into a proper CLI with options and the ability to dump information to json. Additionally, preparations are in place to publish it to crates, so other people can use it for testing purposes.

CCing @ErichDonGubler as the resident CLI-goto :)

**Testing**

Manually tested, will be part of the test flow soon.


wgpu-info/Cargo.toml
```diff
@@ -3,14 +3,18 @@ name = "wgpu-info"
...
 
 [dependencies]
+anyhow.workspace = true
 bitflags.workspace = true
 env_logger.workspace = true
+pico-args.workspace = true
+serde.workspace = true
+serde_json.workspace = true
 wgpu.workspace = true
+wgpu-types = { workspace = true, features = ["serde"] }
```
Wumpf: we have a bug in wgpu-types that causes `cargo run -p wgpu-info` not to work standalone: The `serde` feature is on, but because neither `trace` nor `replay` are enabled, neither `serde::Serialize` nor `serde::Deserialize` are active causing errors like this for me:

```
error: cannot find attribute `serde` in this scope
    --> wgpu-types/src/lib.rs:5932:31
     |
5932 | #[cfg_attr(feature = "serde", serde(rename_all = "kebab-case"))]
     |                               ^^^^^
     |
     = note: `serde` is in scope, but it is a crate, not an attribute
```

Not sure if this is a rust version thing (but the toolchain file should enforce same version for all of us) or why you didn't hit this 🤔 
To work around this I had to enable `trace` here locally

Wumpf (dismissed):
tried it on my Mac, works nicely and looks pretty :). Code looks good to me as well.
But we need to fix that serde-feature build issue I pointed out. If you have a way to build the tool as is we can also handle that in a separate PR

wgpu-info/Cargo.toml
teoxoy: For reference, this is https://github.com/gfx-rs/wgpu/issues/3609

cwfitzgerald: Weirdly I only hit this in my follow up work... I'll backport the changes I made for that.

Issue #3852, opened 2023-06-09
Update RenderPass draw Calls with better Documentation and examples.


genusistimelord:
So the issue is Some special draw calls like [draw_indexed](https://docs.rs/wgpu/0.16.1/wgpu/struct.RenderPass.html#method.draw_indexed) does not really give full details about what parameters do what exactly. 

So for this case draw_indexed Would Set the index Array range via indices and then the base_vertex which I had to find out in the chat which increments via a offset for the indices and then instances which is a range of the instance being used for the vertex's. It also leaves out important information. For example the indices are used to determine what vertex to load so if we set 0,1,2,0,2,3 and then 4,5,6,4,5,7
this would get the appropriate vertex's from the buffer based on their Count or Byte's / stride to get the count. 

Also with the base_vertex being used as a offset this would increment the index's for me automatically so for 2 objects this could be
 0,1,2,0,2,3 and  0,1,2,0,2,3 for the 2nd object. then I set base_vertex to 0 for the first indices and set it to 4 for the 2nd indices which should if i understood them correctly  update them internally to 0,1,2,0,2,3 and then 4,5,6,4,5,7.
 
Going through and adding this specific type of information to draw calls and other functions could be helpful.

PR #3860, opened 2023-06-12
Update RenderPass draw, draw_indexed, set_viewport  and set_scissor_rect documentation

genusistimelord:
This should hopefully make the documentation a bit better and more understandable for these functions. Please let me know what you like or dislike about this and anything I can fix or improve upon if i was wrong in some way.

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -3159,7 +3159,7 @@ impl<'a> RenderPass<'a> {
     /// vertices: The range of vertex's being used from the Vertex Buffer.
     /// instances: Change the vertex data used for any instance-rate vertex attributes
     ///     from a set instance buffer. The instance index is passed to the vertex
-    ///     shader through the instance_index builtin. Default to 0..1 if instance
+    ///     shader through the instance_index builtin. Use 0..1 if instance
     ///     buffers are not used.
     pub fn draw(&mut self, vertices: Range<u32>, instances: Range<u32>) {
         DynContext::render_pass_draw(
@@ -3212,7 +3212,7 @@ impl<'a> RenderPass<'a> {
     ///     becomes 4,5,6,4,6,7 if base_vertex is set to 4 on the next index offset.
     /// instances: Change the vertex data used for any instance-rate vertex attributes
     ///     from a set instance buffer. The instance index is passed to the vertex
-    ///     shader through the instance_index builtin. Default to 0..1 if instance
+    ///     shader through the instance_index builtin. use 0..1 if instance
     ///     buffers are not used.
     pub fn draw_indexed(&mut self, indices: Range<u32>, base_vertex: i32, instances: Range<u32>) {
         DynContext::render_pass_draw_indexed(
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -3807,6 +3807,15 @@ impl<'a> RenderBundleEncoder<'a> {
     /// Draws primitives from the active vertex buffer(s).
     ///
     /// The active vertex buffers can be set with [`RenderBundleEncoder::set_vertex_buffer`].
+    /// Does not use an Index Buffer. If you need this see [`RenderBundleEncoder::draw_indexed`]
+    ///
+    /// Panics if vertices Range is not within 0..BufferLen / BufferStride.
+    ///
+    /// vertices: The range of vertices being used from the Vertex Buffer.
+    /// instances: Change the vertex data used for any instance-rate vertex attributes
+    ///     from a set instance buffer. The instance index is passed to the vertex
+    ///     shader through the instance_index builtin. Use 0..1 if instance
+    ///     buffers are not used.
     pub fn draw(&mut self, vertices: Range<u32>, instances: Range<u32>) {
         DynContext::render_bundle_encoder_draw(
             &*self.parent.context,
@@ -3819,8 +3828,18 @@ impl<'a> RenderBundleEncoder<'a> {
 
     /// Draws indexed primitives using the active index buffer and the active vertex buffers.
     ///
-    /// The active index buffer can be set with [`RenderBundleEncoder::set_index_buffer`], while the active
-    /// vertex buffers can be set with [`RenderBundleEncoder::set_vertex_buffer`].
+    /// The active index buffer can be set with [`RenderBundleEncoder::set_index_buffer`].
+    /// The active vertex buffers can be set with [`RenderBundleEncoder::set_vertex_buffer`].
+    ///
+    /// Panics if indices Range is not within 0..BufferLen / BufferStride.
+    ///
+    /// indices: The indices range within the index buffer.
+    /// base_vertex: Addition upon the next set of Index's. example: 0,1,2,0,2,3 on next instance
+    ///     becomes 4,5,6,4,6,7 if base_vertex is set to 4 on the next index offset.
+    /// instances: Change the vertex data used for any instance-rate vertex attributes
+    ///     from a set instance buffer. The instance index is passed to the vertex
+    ///     shader through the instance_index builtin. use 0..1 if instance
+    ///     buffers are not used.
     pub fn draw_indexed(&mut self, indices: Range<u32>, base_vertex: i32, instances: Range<u32>) {
         DynContext::render_bundle_encoder_draw_indexed(
             &*self.parent.context,
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -3104,10 +3104,10 @@ impl<'a> RenderPass<'a> {
         )
     }
 
-    /// Sets the scissor region.
+    /// Sets the scissor rectangle used during the rasterization stage.
+    /// After transformation into [viewport coordinates](https://www.w3.org/TR/webgpu/#viewport-coordinates).
     ///
-    /// Subsequent draw calls will discard any fragments that fall outside this region
-    /// as framebuffer space.
+    /// Subsequent draw calls will discard any fragments which fall outside the scissor rectangle.
     pub fn set_scissor_rect(&mut self, x: u32, y: u32, width: u32, height: u32) {
         DynContext::render_pass_set_scissor_rect(
             &*self.parent.context,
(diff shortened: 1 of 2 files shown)


genusistimelord:
This does not fully fix #3852 but it is a good start at least.

wgpu/src/lib.rs
```diff
@@ -3106,7 +3106,8 @@ impl<'a> RenderPass<'a> {
 
     /// Sets the scissor region.
     ///
-    /// Subsequent draw calls will discard any fragments that fall outside this region.
+    /// Subsequent draw calls will discard any fragments that fall outside this region
+    /// as framebuffer space.
```
```suggestion
    /// Subsequent draw calls will discard any fragments that fall outside this region.
```

`as framebuffer space`?

wgpu/src/lib.rs
```diff
@@ -3119,7 +3120,7 @@ impl<'a> RenderPass<'a> {
         );
     }
 
-    /// Sets the viewport region.
+    /// Sets the viewport region of framebuffer space.
```
Wumpf: I don't think "framebuffer" is all that well defined? Not even sure if the spec ended up needing a color target now on renderpass or not..

Could just go with what WebGPU spec writes: `Sets the viewport used during the rasterization stage to linearly map from [normalized device coordinates](https://www.w3.org/TR/webgpu/#ndc) to [viewport coordinates](https://www.w3.org/TR/webgpu/#viewport-coordinates).`

wgpu/src/lib.rs
```diff
@@ -3150,7 +3151,16 @@ impl<'a> RenderPass<'a> {
 
     /// Draws primitives from the active vertex buffer(s).
     ///
-    /// The active vertex buffers can be set with [`RenderPass::set_vertex_buffer`].
+    /// The active vertex buffer(s) can be set with [`RenderPass::set_vertex_buffer`].
+    /// Does not use an Index Buffer. If you need this see [`RenderPass::draw_indexed`]
+    ///
+    /// Panics if vertices Range is not within 0..BufferLen / BufferStride.
```
Wumpf: wgpu doesn't panic if you set an error handler.
Vertex buffers are completely optional to drawing and the validation there is more complicated than that

maybe something like that
```suggestion
    /// Fails if the vertex range is outside of the range of the vertex range of any set vertex buffer.
```

wgpu/src/lib.rs
```diff
@@ -3150,7 +3151,16 @@ impl<'a> RenderPass<'a> {
...
+    ///
+    /// vertices: The range of vertex's being used from the Vertex Buffer.
```
Wumpf: need to reformulate to account for not using vertex buffer and using many vertex buffers

wgpu/src/lib.rs
```diff
@@ -3150,7 +3151,16 @@ impl<'a> RenderPass<'a> {
...
+    /// instances: Change the vertex data used for any instance-rate vertex attributes
+    ///     from a set instance buffer. The instance index is passed to the vertex
```
Wumpf: again, this needs some love for not actually using vertex buffers. Something along the line that this is what advances the stepping on instance stepped vertex buffer if any set. Note on `0..1` is really good, newcomers are bound to trip on that otherwise.
Haven't checked what WebGPU specs here, should be able to find some nice text blocks there

wgpu/src/lib.rs
```diff
@@ -3192,8 +3202,18 @@ impl<'a> RenderPass<'a> {
...
     ///
-    /// The active index buffer can be set with [`RenderPass::set_index_buffer`], while the active
-    /// vertex buffers can be set with [`RenderPass::set_vertex_buffer`].
+    /// The active index buffer can be set with [`RenderPass::set_index_buffer`]
+    /// The active vertex buffers can be set with [`RenderPass::set_vertex_buffer`].
+    ///
+    /// Panics if indices Range is not within 0..BufferLen / BufferStride.
+    ///
+    /// indices: The index range within the index buffer.
+    /// base_vertex: Addition upon the next set of Index's. example: 0,1,2,0,2,3 on next instance
```
```suggestion
    /// base_vertex: Added upon the next set of Index's. E.g. 0,1,2,0,2,3 on next instance
```
but don't know what you mean with "next set" here

wgpu/src/lib.rs
```diff
@@ -3192,8 +3202,18 @@ impl<'a> RenderPass<'a> {
...
+    ///     becomes 4,5,6,4,6,7 if base_vertex is set to 4 on the next index offset.
+    /// instances: Change the vertex data used for any instance-rate vertex attributes
+    ///     from a set instance buffer. The instance index is passed to the vertex
+    ///     shader through the instance_index builtin. use 0..1 if instance
+    ///     buffers are not used.
```
Wumpf: as above :)

wgpu/src/lib.rs
```diff
@@ -3192,8 +3202,18 @@ impl<'a> RenderPass<'a> {
 
     /// Draws indexed primitives using the active index buffer and the active vertex buffers.
     ///
-    /// The active index buffer can be set with [`RenderPass::set_index_buffer`], while the active
-    /// vertex buffers can be set with [`RenderPass::set_vertex_buffer`].
+    /// The active index buffer can be set with [`RenderPass::set_index_buffer`]
+    /// The active vertex buffers can be set with [`RenderPass::set_vertex_buffer`].
+    ///
+    /// Panics if indices Range is not within 0..BufferLen / BufferStride.
```
ideally also contains again a note on vertex range requirements (although I don't even know right now what is supposed to happen when an index is out of bounds for a vertex buffer - I know this is detected, but not sure if it's an error or defined as zeroed data)

Wumpf (changes requested):
Thank you so much for contributing and getting the docs to a better place!

Really like the cross references!

I think though it's super important that the descriptions here are razor sharp and correct for all cases, so there's some more work needed here.
Generally I'd encourage to copy paste from the webgpu spec!

wgpu/src/lib.rs
genusistimelord: this was a comment in the WGPU user chat about this left at https://matrix.to/#/!XFRnMvAfptAHthwBCx:matrix.org/$jYU5UBDRL6d0yqihEv6l2eKCEBTeBtvge_G4WQdTd1Y?via=matrix.org&via=mozilla.org&via=kyju.org  which was also Thumbed up by @cwfitzgerald. We are trying to find a way to show that the Clip region is based on the device coords and not world coords. As this was a confusing thing for some. 

We probably could change this in another way to make it more apparent though as i do agree that FrameBuffer space is not well defined as to what it really is as you said in your other comment.

wgpu/src/lib.rs
genusistimelord: /// Panics if the vertex range is outside of the range of the vertex range of any set vertex buffer unless [on_uncaptured_error](https://docs.rs/wgpu/0.16.1/wgpu/struct.Device.html#method.on_uncaptured_error) is set then will Fail instead.

Or something like this would be good since in most cases people wont know to set this as I did not even know it existed till now.

wgpu/src/lib.rs
genusistimelord: not sure what or how for this?

wgpu/src/lib.rs
genusistimelord: Any idea's on how this could be rewritten to be better? and yeah the 0..1 got me in the beginning too as I thought this originally was for choosing the next vertex. Since this had no documentation on what it does. 

Also I am not sure if the WebGPU specs cover all these for everything or not. by this i mean for Metal, opengl, Vulkan etc. If it does then we could probably be safe in using whatever they give. If not maybe need to phrase it to include all instances?

wgpu/src/lib.rs
genusistimelord: this was taken from jasperriz in the wgpu_user chat. Im guessing he meant that upon the next Call based on the next range of index's given that it would get added to them. Though I am still unsure how this really works fully in the backend other than what @cwfitzgerald  said in the chat of 
```rust
for instance_id in instance_range {
    for index_index in index_range {
        let vertex_id = index_buffer[index_index];
        let adjusted_vertex_id = vertex_id + base_vertex;
        let vertex = vertex[adjusted_vertex_id];
        vertex_shader(vertex, adjusted_vertex_id, instance_id);
    }
}
```

from this example I see this is only useful as a per snippet call of the function and not useful for looping as it does not Adjust the base_vertex by the base vertex on every instance or every set of index's. So yeah this needs to be fixed to remove the next set and more of a Increments the current call or something like this.

wgpu/src/lib.rs
genusistimelord: The WGPU Spec actually cover's this by linking to the viewport coordinates inside the wgpu specs.

wgpu/src/lib.rs
genusistimelord: a valid program should also never use vertex indices with [GPUVertexStepMode."vertex"](https://www.w3.org/TR/webgpu/#dom-gpuvertexstepmode-vertex) that are out of bounds. WebGPU implementations have different ways of handling this, and therefore a range of behaviors is allowed. Either the whole draw call is discarded, or the access to those attributes out of bounds is described by WGSL’s [invalid memory reference](https://gpuweb.github.io/gpuweb/wgsl/#invalid-memory-reference).

is how they describe this issue. It could be a error or a discarded draw.

wgpu/src/lib.rs
cwfitzgerald: This still needs to be changed to "viewport coordinates" as that's the spec name for it.

wgpu/src/lib.rs
cwfitzgerald: I think it would be clearer if we just included this code in the text.

wgpu/src/lib.rs
genusistimelord: sorry this was updated i just did not get a chance to push it XD. please check it again.

wgpu/src/lib.rs
genusistimelord: I agree to this and we can do this!

wgpu/src/lib.rs
Wumpf: sounds good!
Scrolling around a bit, I looks like we write "will panic" etc. in a whole bunch of places right now and don't be that super-accurate on "uhm actually you can do on_uncaptured_error", so I guess saying it's paniking is just fine and easier to read

wgpu/src/lib.rs
Wumpf: new version works for me :)

wgpu/src/lib.rs
Wumpf: oh but what you've written now sounds quite good I think and describes what it does with applicable vertex buffers :)
Tbh I haven't read everything in the spec about it, but I think it covers this well enough and there should be no backend differences.

genusistimelord: I changed this to be an actual usage Example of how its done as this shows Way more than a description would in most cases.

Wumpf (approved):
great stuff! Thanks again :)

PR #3863, opened 2023-06-13
Allow empty scissor rects

PJB3005:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
*none*

**Description**
This used to be disallowed in older versions of the WebGPU spec, but it's allowed now: https://github.com/gpuweb/gpuweb/commit/eb96ea65229d99027f85b280e5ddf73a27541948

That said, I'm not sure about the Metal side of it all. According to the original discussion, Metal doesn't support empty scissor rects. I tested just now however, and at least on Ventura 13.4 it's absolutely allowed. I cross-checked Dawn's source code and there doesn't seem to be any special trickery of using empty viewports there either (as suggested in the spec commit).

I just left the Metal backend untouched and, well, it seems to work.

**Testing**
I modified `hello-triangle` to draw with an empty scissor rectangle, and confirmed it did what I want.


cwfitzgerald (approved):
LGTM: Filed https://github.com/gfx-rs/wgpu/issues/3871 as a follow up

PR #3873, opened 2023-06-16
GPU Test Framework

### LINKED ISSUE #3622: Reconsider lavapipe

### LINKED ISSUE #3680: Hard to Tell if Test is Skipped

### LINKED ISSUE #4097: Examples which exclude lavapipe wrongly exclude other drivers with mesa vendor id
**Description**
Examples and tests which exclude lavapipe use the vendor id to identify the device. Community drivers in Mesa use the same vendor id. To be more precise, any case where lavapipe is excluded should also consider the adapter name.

**Repro steps**
Well... I'm not going to encourage you to run Vulkan drivers (agxv) which are still in development.

Tests and examples which skip/explicitly fail when using lavapipe should not cause other drivers with the mesa vendor id to fail

cwfitzgerald:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Closes https://github.com/gfx-rs/wgpu/issues/3680 
Closes #3622 
Closes https://github.com/gfx-rs/wgpu/issues/4097

**Description**

This revamps our testing framework in multiple different ways.
- We now automatically test against all adapters in a given system.
- Supports async and sync code in our tests.
- Improves the visibility to skipped tests.

**How to Review**

While there are a ton of changes in this MR, a lot of it is very boring structural changes in the code. Please focus on `tests/src` `examples/common`, `wgpu-macros` and `xtask`.

**Testing**

It is lol


cwfitzgerald:
Hey me! Rebase this!

.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ env:
   CARGO_TERM_COLOR: always
   RUST_LOG: info
   RUST_BACKTRACE: full
-  MSRV: 1.65
+  MSRV: 1.70
```
cwfitzgerald: Figure this out

wgpu-macros/src/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+
+            ::wgpu_test::infra::TEST_LIST.lock().push(
+                // Allow any type that can be converted to a GpuTestConfiguration
+                ::wgpu_test::infra::GpuTestConfiguration::from(#expr).name_from_init_function_typename::<S>(#ident_lower)
+            )
+        }
+
+        #[cfg(target_arch = "wasm32")]
+        fn #test_name_webgl() {
+            // todo
```
cwfitzgerald: Figure this out

tests/tests/shader/zero_init_workgroup_mem.rs
```diff
@@ -27,151 +26,153 @@ fn zero_init_workgroup_mem() {
...
-    MAX_DEVICE_WORKGROUP_MEMORY / (SHADER_WORKGROUP_MEMORY * TOTAL_WORK_GROUPS) + 1; // TODO: use div_ceil once stabilized
+            ))
+            .skip(FailureCase::backend_adapter(
+                Backends::VULKAN,
+                "llvmpipe",
+            ))
+            .limits(Limits::downlevel_defaults()),
+    )
+    .run_sync(|ctx| {
+        let bgl = ctx
```
cwfitzgerald, line 29: These changes look suspisious

tests/tests/device.rs
```diff
@@ -1,45 +1,32 @@
...
+                layout: &bind_group_layout,
+                entries: &[],
+            });
+        }
+
+        ctx.device.poll(wgpu::Maintain::Poll);
+    });
 
 #[cfg(not(all(target_arch = "wasm32", not(target_os = "emscripten"))))]
 #[test]
```
cwfitzgerald, line 30: This needs to be wired up.

nical:
It looks like metal is not tested anymore.

nical:
> It looks like metal is not tested anymore.

Nevermind that.

cwfitzgerald:
Alright, I think this MR is finally good to land whenever anyone can find time to review it - thankfully it's not actually as large as it seems:

`git diff origin/trunk --stat tests/src examples/common wgpu-macros xtask`

is

`16 files changed, 1280 insertions(+), 782 deletions(-)`

teoxoy (approved):
Took a brief look, nothing looks off.

The new features look great!
Even automatic retry and reporting of flaky tests! 🎉 

Shippit! 🚀

cwfitzgerald:
WOOOOOOOOO

PR #3874, opened 2023-06-17
Expose TextureUsages in SurfaceCapabilities

### LINKED ISSUE #3869: `wgpu::SurfaceCapabilities` does not expose supported usages
This came up over at egui where the screenshot mechanism wants to use COPY_DST to copy the screenshot target to the output surface which is supported on some backends - see https://github.com/emilk/egui/issues/3068

We already have this information on `hal::SurfaceCapabilities` but don't forward it all the way!

amfaber:
**Checklist**

- [ ] Run `cargo clippy`.

**Connections**
Closes https://github.com/gfx-rs/wgpu/issues/3869.

**Description**
As mentioned in the issue, on native the texture usages of the surface are already known on `hal::SurfaceCapabilites`. This is in the form of hal::TextureUses rather than the frontfacing TextureUsages however. I couldn't find a function mapping in this direction, but tried to construct the correct logic based on the inverse mapping of `map_texture_usage` in wgpu-core/src/conv.rs.

This is very much a draft though, as I have no idea how to access the TextureUs[es|ages] from the wasm side.

**Testing**
None yet.

### DIFF
--- a/wgpu-core/src/conv.rs
+++ b/wgpu-core/src/conv.rs
@@ -151,10 +151,7 @@ pub fn map_texture_usage_from_hal(uses: hal::TextureUses) -> wgt::TextureUsages 
     );
     u.set(
         wgt::TextureUsages::RENDER_ATTACHMENT,
-        uses.contains(hal::TextureUses::COLOR_TARGET)
-            | uses.contains(
-                hal::TextureUses::DEPTH_STENCIL_READ | hal::TextureUses::DEPTH_STENCIL_WRITE,
-            ),
+        uses.contains(hal::TextureUses::COLOR_TARGET),
     );
     u
 }
--- a/wgpu-core/src/conv.rs
+++ b/wgpu-core/src/conv.rs
@@ -131,9 +131,7 @@ pub fn map_texture_usage(
     u
 }
 
-pub fn map_texture_usage_from_hal(
-    uses: hal::TextureUses,
-) -> wgt::TextureUsages{
+pub fn map_texture_usage_from_hal(uses: hal::TextureUses) -> wgt::TextureUsages {
     let mut u = wgt::TextureUsages::empty();
     u.set(
         wgt::TextureUsages::COPY_SRC,
@@ -149,21 +147,17 @@ pub fn map_texture_usage_from_hal(
     );
     u.set(
         wgt::TextureUsages::STORAGE_BINDING,
-        uses.contains(
-            hal::TextureUses::STORAGE_READ | hal::TextureUses::STORAGE_READ_WRITE,
-        ),
+        uses.contains(hal::TextureUses::STORAGE_READ | hal::TextureUses::STORAGE_READ_WRITE),
     );
     u.set(
         wgt::TextureUsages::RENDER_ATTACHMENT,
         uses.contains(hal::TextureUses::COLOR_TARGET)
             | uses.contains(
-                hal::TextureUses::DEPTH_STENCIL_READ
-                    | hal::TextureUses::DEPTH_STENCIL_WRITE,
+                hal::TextureUses::DEPTH_STENCIL_READ | hal::TextureUses::DEPTH_STENCIL_WRITE,
             ),
     );
     u
 }
-
 
 pub fn check_texture_dimension_size(
     dimension: wgt::TextureDimension,
--- a/wgpu-core/src/conv.rs
+++ b/wgpu-core/src/conv.rs
@@ -131,6 +131,40 @@ pub fn map_texture_usage(
     u
 }
 
+pub fn map_texture_usage_from_hal(
+    uses: hal::TextureUses,
+) -> wgt::TextureUsages{
+    let mut u = wgt::TextureUsages::empty();
+    u.set(
+        wgt::TextureUsages::COPY_SRC,
+        uses.contains(hal::TextureUses::COPY_SRC),
+    );
+    u.set(
+        wgt::TextureUsages::COPY_DST,
+        uses.contains(hal::TextureUses::COPY_DST),
+    );
+    u.set(
+        wgt::TextureUsages::TEXTURE_BINDING,
+        uses.contains(hal::TextureUses::RESOURCE),
+    );
+    u.set(
+        wgt::TextureUsages::STORAGE_BINDING,
+        uses.contains(
+            hal::TextureUses::STORAGE_READ | hal::TextureUses::STORAGE_READ_WRITE,
+        ),
+    );
+    u.set(
+        wgt::TextureUsages::RENDER_ATTACHMENT,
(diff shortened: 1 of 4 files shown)


wgpu-types/src/lib.rs
```diff
@@ -4584,6 +4584,10 @@ pub struct SurfaceCapabilities {
     ///
     /// Will return at least one element, CompositeAlphaMode::Opaque or CompositeAlphaMode::Inherit.
     pub alpha_modes: Vec<CompositeAlphaMode>,
+    /// List of supported texture usages for the surface to use with the given adapter.
+    ///
+    /// What happens if the surface is incompatible with the adapter?
```
Wumpf: It should panic or give an error in the error handler if one is installed. We should add that to the list of error cases on `wgpu::Adapter::configure`

Wumpf: comment here should also mention that render target usage is guaranteed (just like the corresponding HAL field does :))

wgpu/src/backend/web.rs
```diff
@@ -1123,6 +1124,8 @@ impl crate::context::Context for Context {
             // Doesn't really have meaning on the web.
             present_modes: vec![wgt::PresentMode::Fifo],
             alpha_modes: vec![wgt::CompositeAlphaMode::Opaque],
+            
+            // Something has to go here
```
Wumpf: surface handling on web is a bit shacky - the spec doesn't have this api as-is as it works with the canvas directly, so it's more of a pure wgpu construct.
Only the RENDER_ATTACHMENT usage should go here for now - see also https://www.w3.org/TR/webgpu/#canvas-configuration

wgpu-core/src/device/global.rs
```diff
@@ -58,10 +58,39 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+                ),
+            );
+            usages.set(
+                wgt::TextureUsages::RENDER_ATTACHMENT,
+                hal_caps.usage.contains(hal::TextureUses::COLOR_TARGET)
+                    | hal_caps.usage.contains(
+                        hal::TextureUses::DEPTH_STENCIL_READ
+                            | hal::TextureUses::DEPTH_STENCIL_WRITE,
+                    ),
+            );
```
Wumpf: there's a `map_texture_usage` on `conv.rs` that you should be able to use for this

amfaber: That one is the inverse mapping unfortunately. It goes TextureUsages -> TextureUses, but we need TextureUses -> TextureUsages here. I based this logic on inverting the logic I found there. We could extract all of this to its own function in conv.rs for going in this direction though. Personally I think this would probably be more elegant as From<T> implementations in both directions.

wgpu-types/src/lib.rs
amfaber: Does this mean that the default implementation should be RENDER_ATTACHMENT rather than empty()?

I assumed that the error reporting is already handled elsewhere. The comments mention what the fields are in the case of incompatibility though. I assume that that is the purpose of the Default implementation, since it matches the comments?

wgpu/src/backend/web.rs
amfaber: Well that makes my job easy atleast ^^
Perhaps in the future this can provide the usages https://www.w3.org/TR/webgpu/#dom-gpucanvascontext-getcurrenttexture. I saw it was unstable for now though

wgpu-core/src/device/global.rs
Wumpf: oh, you're right ofc, didn't pay enough attention
Yeah let's put that into conv.rs, surprised we didn't have it yet :)

wgpu-types/src/lib.rs
Wumpf: > Does this mean that the default implementation should be RENDER_ATTACHMENT rather than empty()?

I think so, yes!

wgpu-core/src/device/global.rs
amfaber: Great! Just as a plain ol' function, or what do you think about From implementations?

fralonra:
Hi, any progress?

wgpu/src/backend/web.rs
```diff
@@ -1160,6 +1160,8 @@ impl crate::context::Context for Context {
             // Doesn't really have meaning on the web.
             present_modes: vec![wgt::PresentMode::Fifo],
             alpha_modes: vec![wgt::CompositeAlphaMode::Opaque],
+            // Statically set to RENDER_ATTACHMENT for now. See https://github.com/gfx-rs/wgpu/pull/3874
+            usages: wgt::TextureUsages::RENDER_ATTACHMENT,
```
cwfitzgerald, line 1164: Currently this _must_ be RENDER_ATTACHMENT on WebGPU, so this is fine. https://gpuweb.github.io/gpuweb/#dom-gpucanvasconfiguration-usage

wgpu-core/src/conv.rs
```diff
@@ -131,6 +131,34 @@ pub fn map_texture_usage(
...
+    u.set(
+        wgt::TextureUsages::STORAGE_BINDING,
+        uses.contains(hal::TextureUses::STORAGE_READ | hal::TextureUses::STORAGE_READ_WRITE),
+    );
+    u.set(
+        wgt::TextureUsages::RENDER_ATTACHMENT,
+        uses.contains(hal::TextureUses::COLOR_TARGET)
+            | uses.contains(
+                hal::TextureUses::DEPTH_STENCIL_READ | hal::TextureUses::DEPTH_STENCIL_WRITE,
+            ),
```
cwfitzgerald:
```suggestion
        uses.contains(hal::TextureUses::COLOR_TARGET),
```

Surfaces can never be depth/stencil

wgpu/src/backend/web.rs
```diff
@@ -1160,6 +1160,8 @@ impl crate::context::Context for Context {
             // Doesn't really have meaning on the web.
             present_modes: vec![wgt::PresentMode::Fifo],
             alpha_modes: vec![wgt::CompositeAlphaMode::Opaque],
+            // Statically set to RENDER_ATTACHMENT for now. See https://github.com/gfx-rs/wgpu/pull/3874
```
cwfitzgerald:
```suggestion
            // Statically set to RENDER_ATTACHMENT for now. See https://gpuweb.github.io/gpuweb/#dom-gpucanvasconfiguration-usage
```

cwfitzgerald (approved):
Great! Thank you for the wait, got caught up in the chaos.

wgpu-core/src/conv.rs
amfaber: Now that this function is in the more general conv.rs, what if someone down the road uses it on something that isn't a surface?

PR #3875, opened 2023-06-18
Fix the examples link

jembishop:
**Description**
The examples link pointed to the wrong path and gave a 404.


Wumpf (approved):
Thank you! That go missed in the recent move

PR #3877, opened 2023-06-19
Add tests for scissors

### LINKED ISSUE #3871: Tests for Scissors
Particularly we need tests for zero size, but general tests would be good too.

artyomd:
**Checklist**

- [x] Run `cargo clippy`.
- [ ] ~~Add change to CHANGELOG.md. See simple instructions inside file.~~ (no need)

**Description**
Added some tests for scissors. Resolves #3871

### DIFF
--- a/tests/Cargo.toml
+++ b/tests/Cargo.toml
@@ -20,14 +20,14 @@ webgl = ["wgpu/webgl"]
 
 [dependencies]
 bitflags.workspace = true
+bytemuck.workspace = true
 cfg-if.workspace = true
 env_logger.workspace = true
 log.workspace = true
-pollster.workspace = true
 png.workspace = true
+pollster.workspace = true
 wgpu.workspace = true
 wgt.workspace = true
-bytemuck.workspace = true
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 nv-flip.workspace = true
@@ -43,8 +43,8 @@ naga = { workspace = true, features = ["wgsl-in"] }
 wasm-bindgen-test.workspace = true
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
-js-sys.workspace = true
 image.workspace = true
-wasm-bindgen.workspace = true
+js-sys.workspace = true
 wasm-bindgen-futures.workspace = true
+wasm-bindgen.workspace = true
 web-sys = { workspace = true, features = ["CanvasRenderingContext2d", "Blob"] }
--- a/tests/Cargo.toml
+++ b/tests/Cargo.toml
@@ -27,6 +27,7 @@ pollster.workspace = true
 png.workspace = true
 wgpu.workspace = true
 wgt.workspace = true
+bytemuck.workspace = true
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 nv-flip.workspace = true
@@ -38,7 +39,6 @@ wasm-bindgen.workspace = true
 web-sys = { workspace = true }
 
 [dev-dependencies]
-bytemuck.workspace = true
 naga = { workspace = true, features = ["wgsl-in"] }
 wasm-bindgen-test.workspace = true
 
--- a/tests/src/image.rs
+++ b/tests/src/image.rs
@@ -1,7 +1,7 @@
+use crate::TestingContext;
 use std::{borrow::Cow, ffi::OsStr, io, path::Path};
 use wgpu::util::DeviceExt;
 use wgpu::*;
-use crate::TestingContext;
 
 fn read_png(path: impl AsRef<Path>, width: u32, height: u32) -> Option<Vec<u8>> {
     let data = match std::fs::read(&path) {
@@ -207,7 +207,7 @@ pub fn compare_image_output(
         let file_stem = reference_path.file_stem().unwrap().to_string_lossy();
         // Determine the paths to write out the various intermediate files
         let actual_path = Path::new(&path).with_file_name(
-            OsString::from_str(&format!("{}-{}-actual.png", file_stem, backend.to_str(), )).unwrap(),
+            OsString::from_str(&format!("{}-{}-actual.png", file_stem, backend.to_str(),)).unwrap(),
         );
         let difference_path = Path::new(&path).with_file_name(
             OsString::from_str(&format!(
@@ -215,7 +215,7 @@ pub fn compare_image_output(
                 file_stem,
                 backend.to_str(),
             ))
-                .unwrap(),
+            .unwrap(),
         );
 
         // Convert the error values to a false color reprensentation
@@ -465,14 +465,14 @@ impl ReadbackBuffers {
         if texture.format().is_combined_depth_stencil_format() {
             let buffer_size = base_size
                 * texture
-                .format()
-                .block_size(Some(TextureAspect::DepthOnly))
-                .unwrap_or(4);
+                    .format()
+                    .block_size(Some(TextureAspect::DepthOnly))
+                    .unwrap_or(4);
             let buffer_stencil_size = base_size
                 * texture
-                .format()
-                .block_size(Some(TextureAspect::StencilOnly))
-                .unwrap();
+                    .format()
+                    .block_size(Some(TextureAspect::StencilOnly))
+                    .unwrap();
             let buffer = device.create_buffer_init(&util::BufferInitDescriptor {
                 label: Some("Texture Readback"),
                 usage: BufferUsages::MAP_READ | BufferUsages::COPY_DST,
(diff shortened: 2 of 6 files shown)


tests/Cargo.toml
```diff
@@ -27,6 +27,7 @@ pollster.workspace = true
 png.workspace = true
 wgpu.workspace = true
 wgt.workspace = true
+bytemuck.workspace = true
```
cwfitzgerald: Mind alphabetical order please

tests/src/image.rs
```diff
@@ -526,3 +527,51 @@ impl ReadbackBuffers {
                 .unwrap_or(true)
     }
 }
+
+pub fn capture_rgba_u8_texture(
```
cwfitzgerald: Couldn't you use the already existing ReadbackBuffers to do this (and the SPI tests should use them too?)

tests/tests/scissor_tests/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+    width: u32,
+    height: u32,
+}
+
+fn scissor_test_impl(ctx: &TestingContext, scissor_rect: Rect, expected_data: [u8; 16]) {
+    let vertex_data: [f32; 12] = [
+        -1.0_f32, -1.0_f32, 0.0_f32, 1.0_f32, -1.0_f32, 0.0_f32, 1.0_f32, 1.0_f32, 0.0_f32,
+        -1.0_f32, 1.0_f32, 0.0_f32,
+    ];
+    let index_data: [i16; 6] = [0, 1, 3, 1, 3, 2];
```
cwfitzgerald: If you're just trying to make fullscreen geometry, you can use the same VS as mipmap does here: https://github.com/gfx-rs/wgpu/blob/trunk/examples/mipmap/src/blit.wgsl

cwfitzgerald (changes requested):
Looking good minus comments

tests/Cargo.toml
artyomd: 👍 I fixed the order of all the dependencies in this file

tests/src/image.rs
artyomd: This is a bit tricky. The entire problem is the COPY_BYTES_PER_ROW_ALIGNMENT requirement. Both in the scissors test and in the spi, the tests render the 2x2 texture. At first, I thought that it will be a good idea to support copying textures that do not meet the requirement in the readback buffers implementation, but when I started implementing that, at some point it felt that readback buffers needed their own tests as the implementation was getting too complicated (multiple types of textures, multiple ways to copy them, etc..) Personally, I like to keep tests as simple as possible, so I decided the best approach to this will be just to rewrite tests to meet the alignment requirement. What do you think about this?

tests/tests/scissor_tests/mod.rs
artyomd: Thank you for mentioning that. I wasn't familiar with that technique. Utilizing it allowed me to simplify the test a lot.

tests/src/image.rs
cwfitzgerald: Shouldn't you be able to manage the COPY_BYTES_PER_ROW_ALIGNMENT by purely byte manipulation?

When computing the buffer size and executing the copy round up the `bytes_per_row` to `COPY_BYTES_PER_ROW_ALIGNMENT`. Then using the real bytes_per_row information, copy only the chunks that have real data in them into the vector.

I would prefer to keep the tests small, as it helps with understanding what exactly is getting tested.

artyomd: "I would prefer to keep the tests small, as it helps with understanding what exactly is getting tested."
I agree with this.

"Shouldn't you be able to manage the COPY_BYTES_PER_ROW_ALIGNMENT by purely byte manipulation?"
Yes, and that was done in the original `capture_rgba_u8_texture` function that was written in the SPI test.

I think that I may have overanalyzed your initial comment. Just to make it clear, you want to keep the `capture_rgba_u8_texture` function in the image.rs, but use ReadbackBuffers in the implementation, am I right?

cwfitzgerald: Ideally ReadbackBuffers can just deal with the COPY_BYTES_PER_ROW requirement internally. Then all the tests can just use ReadbackBuffers without thinking about it.

artyomd: Done.

tests/src/image.rs
```diff
@@ -1,6 +1,8 @@
 use std::{borrow::Cow, ffi::OsStr, io, path::Path};
+
 use wgpu::util::DeviceExt;
 use wgpu::*;
+use wgt::math::align_to;
```
cwfitzgerald: We shouldn't dep on wgt if we can help it, this is re-exported in wgpu.

tests/src/image.rs
```diff
@@ -458,20 +472,37 @@ pub struct ReadbackBuffers {
 impl ReadbackBuffers {
     pub fn new(device: &Device, texture: &Texture) -> Self {
         let (block_width, block_height) = texture.format().block_dimensions();
-        let base_size = (texture.width() / block_width)
-            * (texture.height() / block_height)
-            * texture.depth_or_array_layers();
+        let should_align_buffer_size = !([
+            TextureFormat::Depth24Plus,
+            TextureFormat::Depth24PlusStencil8,
```
cwfitzgerald, line 476: I don't understand why we're only aligning to the buffer size on these types. No matter the texture format, we need to align the copies to the bytes_per_row alignment.

tests/src/image.rs
```diff
@@ -505,24 +550,70 @@ impl ReadbackBuffers {
...
+            * self.texture_format.block_size(aspect).unwrap_or(4);
+        let expected_buffer_size = expected_bytes_per_row
+            * (self.texture_height / block_height)
+            * self.texture_depth_or_array_layers;
+        let data: BufferView = buffer_slice.get_mapped_range();
+        if expected_buffer_size as usize == data.len() {
+            data.to_vec()
+        } else {
+            bytemuck::cast_slice(&data)
+                .to_vec()
```
cwfitzgerald: This to_vec is unneeded

tests/src/image.rs
```diff
@@ -505,24 +550,70 @@ impl ReadbackBuffers {
...
 
-        is_zero(device, &self.buffer)
+        is_zero(device, &self.buffer, self.buffer_aspect())
             && self
                 .buffer_stencil
                 .as_ref()
-                .map(|buffer_stencil| is_zero(device, buffer_stencil))
+                .map(|buffer_stencil| {
+                    is_zero(device, buffer_stencil, Some(TextureAspect::StencilOnly))
+                })
```
cwfitzgerald: Could we split this into two bools which are then and'ed

tests/tests/shader_primitive_index/mod.rs
```diff
@@ -168,83 +169,31 @@ fn pulling_common(
...
-                rows_per_image: None,
-            },
-        },
-        texture_size,
-    );
-
+        rpass.set_pipeline(&pipeline);
+        rpass.set_index_buffer(index_buffer.slice(..), wgpu::IndexFormat::Uint32);
+        rpass.set_vertex_buffer(0, vertex_buffer.slice(..));
+        function(&mut rpass);
```
cwfitzgerald: What is this call?

cwfitzgerald (changes requested):
Much better direction, some comments

tests/src/image.rs
artyomd: fixed

tests/src/image.rs
artyomd: It is the opposite, we are **not aligning** if the texture format is one of these. The reason why we are not checking the alignment is that we are using a custom compute pipeline to copy the data if the depth format is Depth24Plus. And as far as I know, that is because as of [webgpu spec](https://gpuweb.github.io/gpuweb/#depth-formats), depth aspects of textures with those formats are an invalid copy source.

tests/src/image.rs
artyomd: removed

tests/src/image.rs
artyomd: Honestly, I prefer the way that it was written, but I agree that at first, it is a little bit hard to grasp this logic.
refactored this one, too

tests/tests/shader_primitive_index/mod.rs
artyomd: So this is the original name of the parameter that was written. It was very descriptive,  but as I was refactoring this function, somehow I didn't notice that.

tests/src/image.rs
artyomd: Improved the readability of these lines too

cwfitzgerald (approved):
Amazing

PR #3880, opened 2023-06-20
Make examples in main workspace visible to run-wasm xtask

Gobbel2000:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

The run-wasm xtask was introduced with #3844.

**Description**

Running examples through run-wasm (as in `cargo xtask run-wasm --bin hello-triangle`) would fail with the error message:
```
error: no bin target named `hello-triangle`.
Available bin targets:
    xtask
```
This is because the xtask package was separate and not part of the wgpu workspace (`xtask` is the only known binary target). This change makes xtask a workspace member so that the examples are visible, causing the above command to succeed as expected.

**Testing**

Running `cargo xtask run-wasm --bin hello-triangle` in the top directory of the repository now correctly builds and serves the example. It can correctly be viewed in Firefox Nightly.

I was also going to add a test in the github CI workflow that simply calls `cargo xtask run-wasm --bin hello-triangle --build-only` but it turns out that `cargo_run_wasm` doesn't report any errors so the test would never fail in the intended cases.

It was also necessary to adjust `.deny.toml` so that the cargo-deny checks pass.

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -2,6 +2,7 @@ use std::process::ExitCode;
 
 use anyhow::Context;
 use cli::{Args, Subcommand};
+use pico_args::Arguments;
 
 mod cli;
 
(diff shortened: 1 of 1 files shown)


ErichDonGubler:
`xtask` was deliberately left out of the main workspace and its `Cargo.lock`. Would it not be possible to, say, change the CWD before invoking `cargo_run_wasm` logic, and achieve the same effect?

Gobbel2000:
cargo_run_wasm uses the `CARGO_MANIFEST_DIR` environment variable for figuring out which workspace to run in, so the working directory shouldn't have any effect. But it is possible to point to the main workspace with the --manifest-path option to cargo_run_wasm, as in `cargo xtask run-wasm --bin hello-triangle --manifest-path ../Cargo.toml`. Adding that to the args inside the xtask might work as well, I'll see if I can't get that to work.

Gobbel2000:
This new approach leaves the workspace structure as it was and injects `--manifest-dir ../Cargo.toml` into the arguments for `cargo_run_wasm`. `cargo build` is called inside the xtask directory, so this points to the Cargo.toml in the main workspace. The builds are still placed into `xtask/target`. Otherwise the result is the same and running the examples works as expected.

cwfitzgerald (approved):
Thanks!

PR #3883, opened 2023-06-22
metal: support (simulated) visionOS

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Description**
Apple announced the new **visionOS** at WWDC 2023. Now, visionOS 1.0 can be run on Xcode 15 Beta 2 through the simulator. This pull request addresses a bug that occurs when running wgpu on visionOS.

The MSL version was determined based on the iOS and macOS versions. The following error occurs when running the `boids` example: 
```log
Using Apple xrOS simulator GPU (Metal)
[2023-06-22T07:41:49Z ERROR wgpu::backend::direct] Handling wgpu errors as fatal by default
thread '<unnamed>' panicked at 'wgpu error: Validation Error

Caused by:
      note: label = `Compute pipeline`
    Internal error: MSL: UnsupportedAddressSpace(Storage { access: StorageAccess(LOAD) })

', /Users/lijinlei/Rust/forks/wgpu/wgpu/src/backend/direct.rs:3023:5
```

In addition,  the `family_check` field was also set based on the iOS and macOS versions, which causes feature detection to fail on visionOS. The following error occurs when running the `shadow` example: 
```log
-[MTLDebugRenderCommandEncoder validateCommonDrawErrors:]:5780: failed assertion `Draw Errors Validation

Vertex Function(vs_bake): the offset into the buffer u_entity that is bound at buffer index 1 must be a multiple of 256 but was set to 128.
'
```

**Testing**
Tested via https://github.com/jinleili/wgpu-in-app

https://github.com/gfx-rs/wgpu/assets/1001342/d6926b17-8707-4cdf-815b-659d6f2e3049

### DIFF
--- a/wgpu-hal/src/metal/adapter.rs
+++ b/wgpu-hal/src/metal/adapter.rs
@@ -478,7 +478,12 @@ impl super::PrivateCapabilities {
         };
 
         let os_is_mac = device.supports_feature_set(MTLFeatureSet::macOS_GPUFamily1_v1);
-        let family_check = version.at_least((10, 15), (13, 0), os_is_mac);
+        // Metal was first introduced in OS X 10.11 and iOS 8. The current version number of visionOS is 1.0.0. Additionally,
+        // on the Simulator, Apple only provides the Apple2 GPU capability, and the Apple2+ GPU capability covers the capabilities of Apple2.
+        // Therefore, the following conditions can be used to determine if it is visionOS.
+        // https://developer.apple.com/documentation/metal/developing_metal_apps_that_run_in_simulator
+        let os_is_xr = version.major < 8 && device.supports_family(MTLGPUFamily::Apple2);
+        let family_check = os_is_xr || version.at_least((10, 15), (13, 0), os_is_mac);
 
         let mut sample_count_mask = crate::TextureFormatCapabilities::MULTISAMPLE_X4; // 1 and 4 samples are supported on all devices
         if device.supports_texture_sample_count(2) {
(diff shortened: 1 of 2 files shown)


wgpu-hal/src/metal/adapter.rs
```diff
@@ -478,7 +478,12 @@ impl super::PrivateCapabilities {
         };
 
         let os_is_mac = device.supports_feature_set(MTLFeatureSet::macOS_GPUFamily1_v1);
-        let family_check = version.at_least((10, 15), (13, 0), os_is_mac);
+        // Metal was first introduced in OS X 10.11 and iOS 8. The current version number of visionOS is 1.0.0. Additionally,
+        // on the Simulator, Apple only provides the Apple2 GPU capability, and the Apple2+ GPU capability covers the capabilities of Apple2.
+        // Therefore, the following conditions can be used to determine if it is visionOS.
+        // https://developer.apple.com/documentation/metal/developing_metal_apps_that_run_in_simulator
+        let os_is_xr = version.major < 8 && device.supports_family(MTLGPUFamily::Apple2);
```
teoxoy, line 485: I'd have expected this to be `Apple8` or `Metal3` since it's running an M2.

teoxoy (commented):
I can't yet find any info on visionOS in the Metal Feature Set Tables. I feel like it might be too early for this but if people want to play with it I'm not opposed to it.

wgpu-hal/src/metal/adapter.rs
jinleili: This is the frustrating part about Apple. visionOS only supports apple2 on the simulator (M1 Mac) and requires a real device to support apple8. I also can't understand why Apple would impose such limitations on the simulator.

jinleili: Starting from Apple7, Apple no longer specifically indicates the corresponding OS for the GPU in their  [Metal Feature Set Tables](https://developer.apple.com/documentation/metal/gpu_devices_and_work_submission/detecting_gpu_features_and_metal_software_versions/). This means that the feature is solely determined by the GPU hardware and is independent of the OS.
截屏2023-06-22 20 10 32

teoxoy: I think we should rename this PR and changelog entry to "support (simulated) visionOS" to not give people the wrong impression.

Once Apple updates the Metal docs we can improve the support.

cwfitzgerald (approved):

PR #3885, opened 2023-06-23
Add Extended Examples

JustAnotherCodemonkey:
**Checklist**

- [X] Run `cargo clippy`.

**Description**
Add examples, especially regarding wgpu for GPGPU.

**Testing**
(Most of these I don't really know how to test. I guess the testing method is running it and seeing if what goes to screen / fs is the same as what was advertised.)


examples/repeated-compute/src/main.rs
```diff
@@ -0,0 +1,238 @@
...
+    buffer_slice.map_async(wgpu::MapMode::Read, move |r| sender.send(r).unwrap());
+    /* In order for the mapping to be completed, one of three things must happen.
+    One of those can be calling `Device::poll`. This isn't nessecary on the web as devices
+    are polled automatically but natively, we need to make sure this happens manually. */
+    // `Maintain::Wait` will cause the thread to wait on native but not the web.
+    context.device.poll(wgpu::Maintain::Wait);
+    log::info!("Device polled.");
+    // Now we await the recieving and panic if anything went wrong because we're lazy.
+    reciever.receive().await.unwrap().unwrap();
+    log::info!("Result recieved.");
```
a1phyr:
```suggestion
    let (sender, receiver) = futures_intrusive::channel::shared::oneshot_channel();
    buffer_slice.map_async(wgpu::MapMode::Read, move |r| sender.send(r).unwrap());
    /* In order for the mapping to be completed, one of three things must happen.
    One of those can be calling `Device::poll`. This isn't necessary on the web as devices
    are polled automatically but natively, we need to make sure this happens manually. */
    // `Maintain::Wait` will cause the thread to wait on native but not the web.
    context.device.poll(wgpu::Maintain::Wait);
    log::info!("Device polled.");
    // Now we await the receiving and panic if anything went wrong because we're lazy.
    receiver.receive().await.unwrap().unwrap();
    log::info!("Result received.");
```

A few typos

JustAnotherCodemonkey:
Re-drafting since I decided to add at least one more example for compute workgroups.

cwfitzgerald:
(Will take a bit to review due to the sheer volume of code, likely will get to it next week)

examples/hello-synchronization/src/main.rs
```diff
@@ -0,0 +1,205 @@
...
+    } else {
+        log::info!("hasty_main got lucky.");
+    }
+}
+
+async fn execute(
+    device: &wgpu::Device,
+    queue: &wgpu::Queue,
+    local_patient_workgroup_results: &mut [u32],
+    local_hasty_workgroup_results: &mut [u32],
```
cwfitzgerald: Can these be returned by value instead of being out params

examples/hello-workgroups/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+This example is an extremely bare-bones and arguably somewhat unreasonable demonstration of what workgroup sizes mean in an attempt to explain workgroups in general.
+
+The example starts with two arrays of numbers. One where `a[i] = i` and the other where `b[i] = 2i`. Both are bound to the shader. The program dispatches a workgroup for each index, each workgroup representing both elements at that index in both arrays. Each invocation in each workgroup works on its respective array and adds 1 to the element there.
+
+## What are Workgroups?
+
+### TLDR / Key Takeaways
+
+- Workgroups are 3d grids of invocations that are guaranteed to execute in parallel.
```
cwfitzgerald: This actually isn't true, workgroups are executed one wave at a time by the same SM.

Pedagogically we shouldn't mention this as it's too much information, but we should be careful how to word this.

examples/hello-workgroups/README.md
```diff
@@ -0,0 +1,59 @@
...
...your shader to consider what workgroup the current invocation is in, add a function argument with type `vec3<u32>` and with the attribute `@builtin(workgroup_id)`. Note here that in this example, the term "dispatch grid" is used throughout even though it's not a proper term within WGSL. The specification refers to "workgroup grids" where cells / points are invocations within a workgroup and "compute shader grids" where cells / points are also invocations but for the entire dispatch. Nowhere in the WGSL specification does it ever refer to "dispatch grids." It only ever, in poor grammar, refers to the `workgroup_id` as ["the position of the workgroup in overall compute shader grid"](https://www.w3.org/TR/2023/WD-WGSL-20230629/#workgroup-id-builtin-value) although it does provide a formula to get the workgroup id (["`⌊ CSi ÷ workgroup_size_x ⌋, ⌊ CSj ÷ workgroup_size_y ⌋, ⌊ CSk ÷ workgroup_size_z ⌋`"](https://www.w3.org/TR/2023/WD-WGSL-20230629/#workgroup-id) where CSi, j, and k refer the x, y, and z of a invocation's global invocation id) which is all we need to know what it actually measures. All in all, this example will use the idea of "dispatch grids" because they are convenient.
```
cwfitzgerald: The explanation about the spec should likely be moved to a note as it doesn't _really_ matter to someone just learning, but is good for a concrete reference.

examples/repeated-compute/src/main.rs
```diff
@@ -0,0 +1,248 @@
...
+    flipping ownership of the buffer over to the CPU and making access legal. We do this
+    with `BufferSlice::map_async`. The problem is that map_async is not an async function
+    and we can't await it. What we need to do instead is pass in a closure that will
+    be executed when the slice is either mapped or the mapping has failed. The problem
+    with this is that we don't have a reliable way to wait in the main code for the buffer
+    to be mapped and even worse, calling get_mapped_range or get_mapped_range_mut prematurely
+    will cause a panic, not return an error. Using async channels solves this as awaiting
+    the recieving of a message from the passed closure will force the outside code to wait.
+    It also doesn't hurt if the closure finishes before the outside code catches up as
+    the message is buffered and recieving will just pick that up. */
```
cwfitzgerald: Really minor nit: we should use // comments everywhere and maybe break this up into smaller paragraphs.

Just feels a bit "wall of text"-ey

examples/storage-texture/src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+        .unwrap();
+    image_element.set_src(&data_url);
+    body.append_child(&image_element).unwrap();
+    log::info!("Created image element with data url.");
+    body.set_inner_html(
+        &(body.inner_html()
+            + r#"<p>The above image is for you!
+        You can drag it to your desktop to download.</p>"#),
+    );
+}
```
cwfitzgerald: This seems to be common between examples, so you can move that to `examples/common` so that we don't need to repeat it.

examples/uniform-values/screenshot1.png
cwfitzgerald, line 1: This should be trimmed to not show the window boarder. This image ideally can be used for an example test to make sure the output image is the same.

cwfitzgerald (changes requested):
This is mind blowingly amazing! I have some small, mostly pedagogical nits, but I'm so happy to have this!

examples/hello-workgroups/README.md
JustAnotherCodemonkey: Wait are you talking about the invocations or the workgroups themselves? What I meant is really kind of a copy of line 15 (this was accidental redundancy) but either way, it should probably be removed.

examples/hello-synchronization/src/main.rs
JustAnotherCodemonkey: Oh yeah. Kind of used to this thing from a C background and I've seen this before in Rust. I thought it was appropriate seeing as how we're all rather low level programmers around here with WGPU but if it's not to your liking, I can certainly change it. I can take in size params but what would be the best way to return new `Vec`'s? It might get a little confusing which is which with a tuple? Idk. I'll try that in the next push and see what you think.

examples/hello-workgroups/README.md
JustAnotherCodemonkey: You're totally right. I think I'm going to just trim it way down. You're right, the average person doesn't need literal citations of _the_ tech spec. I do want to include the visual aid though and hit the terminology as it's important to make sure everyone's all on the same page when terms come up. I may later add technical bonus MD's as I do want to give people aids for navigating the spec but that's probably for a later date.

examples/repeated-compute/src/main.rs
JustAnotherCodemonkey: Sure. See if you like the way I broke it up.

examples/uniform-values/screenshot1.png
JustAnotherCodemonkey: I have some ideas but it's a significant change to the example and I'll do that after I do all the minor stuff first.

examples/storage-texture/src/main.rs
JustAnotherCodemonkey: Done in https://github.com/gfx-rs/wgpu/pull/3885/commits/6945d99f3e3499484ab4e1272ad8898ad0320603.

examples/hello-workgroups/README.md
JustAnotherCodemonkey: Ok removed the point in https://github.com/gfx-rs/wgpu/pull/3885/commits/f246ac53ed65bae2b4873fa8a6c1c7800b9365ea but there is also the other point I mentioned. Is that one ok to keep?

examples/hello-workgroups/README.md
JustAnotherCodemonkey: Ok I made some modifications in https://github.com/gfx-rs/wgpu/pull/3885/commits/5a0b591e16564c910a939d01fbc937bd0f7183d8. How do you like it now?

examples/repeated-compute/src/main.rs
JustAnotherCodemonkey: Oh this is in https://github.com/gfx-rs/wgpu/pull/3885/commits/42586901e9efcab52d465e3c40b557a6fd17fe5c btw just so it's easy to find.

JustAnotherCodemonkey:
Hey I just realized that examples/capture is a thing. I don't know what we should do with render-to-texture. I kind of really like render-to-texture, especially because I personally feel it supports wasm better but idk.

examples/uniform-values/screenshot1.png
JustAnotherCodemonkey: Ok so my big idea was to add a capture system to the code; you press space, it renders an image to the FS or on wasm, creates an image like render-to-texture. This would make the example quite testable. The problem, I'm realizing, is that the code for that particular example is so large, it's already a bit hard to follow and find the specifics the example was made for in the first place. Adding this feature would mean that much of the code for render-to-texture would wind up getting copied in, massively increasing the size of the code. I don't know how you feel about what I initially thought. I'm starting to lean towards doing what you initially suggested and just cropping the images.

cwfitzgerald:
General comment: `examples/readme.md` likely needs reworking

examples/hello-synchronization/src/main.rs
cwfitzgerald: You can put them in a named struct that you return

cwfitzgerald:
Things are looking good here, a few comments:

- Check for opportunities to share code between examples, it feels like there's a bit that could be deduplicated
- `exmaples/readme.md` likely needs some kind of "look at this in this order" kind of things, maybe separated by interested subject? Something to guide people through the examples in a sane order.

JustAnotherCodemonkey:
> General comment: `examples/readme.md` likely needs reworking

Definitely. Can you open this as a request for changes?

>Check for opportunities to share code between examples, it feels like there's a bit that could be deduplicated

That's definitely something I'm going to prioritize going forward, starting with rust-struct-to-wgsl.

cwfitzgerald (dismissed):
> Definitely. Can you open this as a request for changes?

Not besides generally doing it for the whole PR - you can't comment on unmodified code in a PR :)

JustAnotherCodemonkey:
Ah well I just need something to remind me to come back and do that because in continuing with everything else, I'm likely to forget.

examples/hello-synchronization/src/main.rs
JustAnotherCodemonkey: How does [30075a6](https://github.com/gfx-rs/wgpu/pull/3885/commits/30075a6f059bdae076b098e13c5a9af8b3a2ab08) look?

examples/uniform-values/screenshot1.png
JustAnotherCodemonkey: Ok tbh I don't really see how trimming it would make it testable. You would have to trim it pixel perfect every time, the window size would always have to be the same, and even then, it's possible different capture technologies could get slightly different pixel colors, especially depending on what image encoding they output. I honestly think the most reasonable test can only be done with the human eye.

cwfitzgerald:
@jimblandy @teoxoy @nical could someone from moz-land take a look at this, I've done a solid first pass on it and the PR is strong, but am really struggling to find the bandwidth to do a second pass.

examples/hello-workgroups/src/shader.wgsl
```diff
@@ -0,0 +1,23 @@
+// This is useful because we can't use, say, vec2<array<i32>> because
+// of array<T> being unsized. Normally we would interweave them or use
+// and array of structus but this is just for the sake of demonstration.
```
nical: typo: "structus"

examples/render-to-texture/src/shader.wgsl
```diff
@@ -0,0 +1,11 @@
+@vertex
+fn vs_main(@builtin(vertex_index) in_vertex_index: u32) -> @builtin(position) vec4<f32> {
+    let x = f32(i32(in_vertex_index) - 1);
+    let y = f32(i32(in_vertex_index & 1u) * 2 - 1);
```
nical: nit: maybe for the sake of the example something like:

```wgsl
    var vertices = array<vec2f, 3>(
        vec2f(-1.0, 0.0),
        vec2f(0.0, -1.0),
        vec2f(1.0, 1.0),
    );

    var position = vertices[in_vertex_index];
```

While a bit less compatc, will actually give the reader less to chew.

examples/repeated-compute/src/main.rs
```diff
@@ -0,0 +1,249 @@
...
+    //
+    // The problem is that map_async is not an async function so we can't await it. What
+    // we need to do instead is pass in a closure that will be executed when the slice is
+    // either mapped or the mapping has failed.
+    //
+    // The problem with this is that we don't have a reliable way to wait in the main
+    // code for the buffer to be mapped and even worse, calling get_mapped_range or
+    // get_mapped_range_mut prematurely will cause a panic, not return an error.
+    //
+    // Using async channels solves this as awaiting the receiving of a message from
```
nical: I don't think that using async channels here is different from using a regular channel, because wgpu isn't integrated with the event loop so if we `await` the channel without polling the device, then the map callback will not be called.

So the expectation here is that the callback will be fired within `context.device.poll(...)` below. and therefore using a standard channel would work the same way.

I think that for these examples it would be better to use standard channels so that it can be used as a reference for folks needing to do something similar in a non-async context, and also to avoid making it look like we need the channel to be async.

The other thing I want to double check before merging is the assumption that it is enough to poll the device once after having submitted and called `map_async` (for example if there is already something in the queue before `submit`, is `poll` guaranteed to process everything including the followups from submitting the frame).

Reading some data back after submission is a common enough pattern that we should be 100% sure about the correct way to do it, and it probably even deserves its own example (no need add it now).

examples/repeated-compute/src/main.rs
```diff
@@ -0,0 +1,249 @@
...
+        // This is where the GPU will read from and write to.
+        let storage_buffer = device.create_buffer(&wgpu::BufferDescriptor {
+            label: None,
+            size: buffer_size as wgpu::BufferAddress,
+            usage: wgpu::BufferUsages::STORAGE
+                | wgpu::BufferUsages::COPY_DST
+                | wgpu::BufferUsages::COPY_SRC,
+            mapped_at_creation: false,
+        });
+        // In WebGPU (unlike many graphics libraries), you aren't allowed to map
```
nical: nit: WebGPU isn't very different from vulkan and other modern APIs. The main difference is that for portability reasons it pretends that there is no memory heaps that are  both CPU and GPU-visible while with vulkan you may or may not have access to such memory heaps depending on the GPU.

I'll propose this formulation:

In WebGPU, for portability reason, there is a strong distinction between GPU-visible and CPU-visible buffers. The former can be used by shaders and the latter can be mapped for the CPU to read and write into. The way to communicate data from the CPU to shaders and back is to copy between the CPU-visible and GPU-visible buffers using `copy_buffer_to_buffer`.

examples/uniform-values/src/main.rs
```diff
@@ -0,0 +1,338 @@
...
+                let view = frame
+                    .texture
+                    .create_view(&wgpu::TextureViewDescriptor::default());
+
+                // (8)
+                wgpu_context_ref.queue.write_buffer(
+                    &wgpu_context_ref.uniform_buffer,
+                    0,
+                    // Ew gross I know. We have to do this to circumvent bytemuck's requirement
+                    // that AppState implement POD (Plain Old Data).
```
nical: Maybe worth expanding what "this" is. Reading this I focused on `bytemuck::bytes_of` and missed the fact that the members are extracted into an array of floats. Also it's something every wgpu user have to deal with and can easily lead to UB so I think it's worth expanding into more words.
How about:

We need to copy our app state into an array of bytes. This is typically done by copying the structure directly using `bytemuck::cast_slice` which, to avoid undefined behavior, requires that the struct implement the `POD` trait. Here, we instead first extract the data into an array of floats which fulfills `bytemuck`'s requirements and cast it into a byte slice using `bytemuck::bytes_of`.

nical (commented):
Great work, thanks a lot! A few minor adjustments before merging this. In particular let's first clarify the parts around `map_async` and double check that it's a reliable pattern to use in general.

examples/repeated-compute/src/main.rs
nical: Connor confirmed that calling poll once after submit and map_async is guaranteed to work :+1: 
Let's still convert the channels to non-async standard ones.

examples/hello-workgroups/src/shader.wgsl
JustAnotherCodemonkey: No clue how that slipped past my spell checker lsp but fixed now.

examples/render-to-texture/src/shader.wgsl
JustAnotherCodemonkey: I agree. I yanked this from hello-triangle so it has the same issue and I feel like it's not as important for users to be able to read this bit, especially when it's the same as hello-triangle (which should probably be changed as well). I will change it though.

examples/repeated-compute/src/main.rs
JustAnotherCodemonkey: How does [ffd1525](https://github.com/gfx-rs/wgpu/pull/3885/commits/ffd1525e9d90c7d61b5198f67e01ec68bb13c4d7) look on this front?

examples/repeated-compute/src/main.rs
JustAnotherCodemonkey: What do you recommend for the channels? The standard Rust channels use mutexes which are not allowed on WASM. I used futures-intrusive because I knew that worked (because hello-compute does it).

cwfitzgerald: I would use flume with the spin feature enabled, but this needs to be an async receive anyway.

@nical on WebGPU, poll is a no-op, so the wait will actually happen in the async_recv.

JustAnotherCodemonkey: Ok cool flume but what is the reason why flume should be used over futures-intrusive which is declared as a dependency of wgpu anyways?

examples/uniform-values/src/main.rs
JustAnotherCodemonkey: Does [291f11c](https://github.com/gfx-rs/wgpu/pull/3885/commits/291f11cb223ac5e9f441f1e26bac220efedb5010) work well or is it too much?

examples/repeated-compute/src/main.rs
nical: > What do you recommend for the channels? The standard Rust channels use mutexes which are not allowed on WASM. I used futures-intrusive because I knew that worked (because hello-compute does it).

I did not have the wasm restrictions in mind, bummer. I think that this context is valuable to learn so I suggest you keep these channels or use another one if you want but explain in a comment that:
 - For native, we could use standard rust channel in principle as the work happens during poll and the "async-ness" of the channel does not matter.
 - But since standard channels use muteces they wouldn't work with the wasm target. In addition, for the wasm target with the WebGPU backend, poll() is a no-op and the wait actually happens in the `async_receive`.
 - As a result the most portable solution involved using async channels.

Sorry for the roundabout, it turned out to be a learning occasion for me.

examples/repeated-compute/src/main.rs
nical: LGTM

examples/uniform-values/src/main.rs
nical: I personally don't mind long comments, however:

> The problem is that there are a lot of things you need to be able to guarantee about how _Rust_ lays out the struct in memory

You can't rely on rust's struct layout at all so I think it would be more fair to say instead:

> The danger is that the struct _must_ be `#[repr(C)]`, its layout _must_ carefully match that of the structure in the shader, and _must_ not contain any padding.

And then,

> Generally, translating structs to bytes using bytemuck and POD should be avoided.

It's a strong stance, I don't entirely disagree but I don't fully agree either and I don't expect everyone in the project shares the opinion.

How about:

> Mistakes are easily made and not caught at build time, so translating structs to bytes using bytemuck and POD should be done very carefully.

examples/repeated-compute/src/main.rs
JustAnotherCodemonkey: No you're fine. I'll definitely include a notice though so that people don't understandably get confused.

examples/uniform-values/src/main.rs
JustAnotherCodemonkey: >You can't rely on rust's struct layout at all so I think it would be more fair to say instead...

I changed it to be more strongly-worded about the difficulty of predicting Rust's structure memory layout. It can be predicted but the rules are complex.

As for your being more soft on people implementing POD, I really strongly believe that it should be emphasized that 90% of users don't know nearly enough about memory internals to be able to reasonably judge when POD can be safely implemented. As you mention, mistakes are easily made and are not caught at build time and in addition, bugs resulting from guessing wrong can be absolutely infuriating to diagnose. I really believe that newer users should just skip learning how to implement POD and instead spend their time learning how to serialize structs in a WGSL-compatible way as even if you know there will be a padding incongruity, the best you can do it just not implement POD but then what? How do you get it to WGSL safely if you don't know how to serialize anyways?

I'm on and off writing a WGPU book and have a whole chapter planned for how to do that serialization.

See [0a1f2eb](https://github.com/gfx-rs/wgpu/pull/3885/commits/0a1f2ebc8a5582b10a0e2d2fd863c3b9fc28f9e6).

JustAnotherCodemonkey:
@cwfitzgerald Since I updated the README, do you think you could lift the change requested that was regarding the README?

examples/uniform-values/src/main.rs
nical: I brought this up at the wgpu maintainer's call and the resolution is that if we want to encourage an opinionated way to transfer data to the gpu it should be via the encase crate.

JustAnotherCodemonkey: Cool! What exactly would this mean for this PR? Should I add a mention of encase or ought I reword significant parts of that comment?

nical: For this PR this would mean replace `bytemuck::bytes_of` and the float slice with something that uses encase (I haven't used encase so I hope what I'm saying makes sense, I'm relaying what was said during the call).

> I really believe that newer users should just skip learning how to implement POD 

I see where you are coming from but you take this stance to a degree that I and other maintainers don't entirely agree with. It's best if the docs that are associated in this repository keep things at a less opinionated level (a level that reflects wgpu's maintainers stance). That level is to explain the requirements and risks, mention encase as a solution for avoiding UB hazards and let users decide what is right for their level of expertise with that information.

For the wording:

> Rust representations of structs are not only incompatible but also complicated to predict

In the general case you simply can't predict it. Let's just say:

> Rust makes no guarantees about the layout of structs unless they are annotated with a specific representation such as #[repr(C)]. 


> It's not only safer but often necessary to implement translations to bytes manually

This is debatable (I at least don't think that the necessary part is true). Let's say something along the lines of: 

> Since this is error prone, we suggest using the encase crate for writing structures into buffers.

JustAnotherCodemonkey: >For this PR this would mean replace bytemuck::bytes_of and the float slice with something that uses encase (I haven't used encase so I hope what I'm saying makes sense, I'm relaying what was said during the call).

Sounds good. I'll start work on this when I can.

>I see where you are coming from but you take this stance to a degree that I and other maintainers don't entirely agree with. It's best if the docs that are associated in this repository keep things at a less opinionated level (a level that reflects wgpu's maintainers stance). That level is to explain the requirements and risks, mention encase as a solution for avoiding UB hazards and let users decide what is right for their level of expertise with that information.

I think this is very fair, especially the part about it being too opinionated. And also since it seems to be the position of the maintainers in general (since you brought this up at the maintainer's call), I'm absolutely going to honor that.

>In the general case you simply can't predict it. Let's just say:

Although Rust's layout of structs is (to my understanding) theoretically predictable, I can definitely get behind that it's not reasonable to try due to the lack of official guarantees and I'm down with the wording you provide.

Also, I think it definitely needed clarifying anyways that by "manually implementing WGSL serialization", I meant to say 'something that wasn't trying to blindly hammer a square, C-compatible peg into a star-shaped WGPU hole' and didn't mean to exclude things like the encase crate. I can 100% see how my poor wording confused though.

nical: No worries we are definitely going deep in a very nuanced topic, and you did a good job of explaining your position. Thanks for your understanding.

JustAnotherCodemonkey: Ok what do you think of the commenting on and implementation of [c5e95f5](https://github.com/gfx-rs/wgpu/pull/3885/commits/c5e95f5ed7749ffeeeb181a151b3fb3047b77359)?

nical (approved):
Thanks a lot for your patience!

examples/hello-synchronization/src/main.rs
JustAnotherCodemonkey: Since I haven't gotten anything back, I'm going to assume that this is resolved.

examples/uniform-values/screenshot1.png
JustAnotherCodemonkey: @nical While you're here, do you think you could weigh in on this one? I don't really want to leave it hanging like this.

cwfitzgerald:
Just need conflicts fixed and we can land!

examples/uniform-values/screenshot1.png
JustAnotherCodemonkey: Resolving as stale

PR #3886, opened 2023-06-23
Update to Rust 1.66

Snowiiii:
**Checklist**

- [X] Run `cargo clippy`.

**Description**
Hello, I have updated our Rust version from 1.65 to 1.66. Why? Firstly, I encountered issues getting Rust to work with this repository. Whenever I used any Cargo command, I received the following error: `the 'cargo' binary, normally provided by the 'cargo' component, is not applicable to the '1.65-x86_64-unknown-linux-gnu' toolchain.` After updating to 1.66, everything worked fine, and I no longer encountered the error. Additionally, Rust 1.66 doesn't have many breaking changes; instead, it has stabilized many APIs. One cool change is that Linux builds now optimize the rustc frontend and LLVM backend with LTO and BOLT, respectively, resulting in improved runtime performance and memory usage. You can find the full changelog [here](https://blog.rust-lang.org/2022/12/15/Rust-1.66.0.html).

**Testing**
I used Clippy, and it ran successfully.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ Bottom level categories:
 #### Misc Breaking Changes
 
 - Change `AdapterInfo::{device,vendor}` to be `u32` instead of `usize`. By @ameknite in [#3760](https://github.com/gfx-rs/wgpu/pull/3760)
+- Move from Rust 1.65 to 1.66. by Snowii [#3886](https://github.com/gfx-rs/wgpu/pull/3886)
 
 #### DX12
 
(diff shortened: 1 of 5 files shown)


nical:
Firefox is built against rustc 1.66, so LGTM

cwfitzgerald:
I'm going to close this for now as we don't have a reason to bump MSRV and we should keep msrv as low as we can. 

As for your error, that seems to be a problem with your local install that should be resolved by removing and adding the toolchain with rustup should get your toolchain back in wack.

PR #3892, opened 2023-06-28
Implement depth-clip-control using depthClamp

AlbinBernhardssonARM:
**Checklist**

- [x] Run `cargo clippy`.

**Description**
In Vulkan, use depthClamp instead of VK_EXT_depth_clip_enable to implement depth-clip-control (unclipped_depth). depthClamp implicitly disables clipping in addition to enabling clamping ([27.4. Primitive Clipping](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#vertexpostproc-clipping) in Vulkan spec). Enabling clamping is fine as clamping depth should always be done according to WebGPU spec ([23.3.6. Fragment Processing](https://www.w3.org/TR/webgpu/#fragment-processing)).

depthClamp is more widely supported than VK_EXT_depth_clip_enable (85.94% vs. 38.42% according to [vulkan.gpuinfo.org](http://vulkan.gpuinfo.org/)) so this enables depth-clip-control on more devices.

**Testing**
Ran shadow example.
Ran DepthClamping unit test in https://github.com/Chainsawkitten/HymnToBeauty/ (requires https://github.com/gfx-rs/wgpu-native/pull/270).


cwfitzgerald (approved):
Good stuff!

Issue #2897, opened 2022-07-20
write-only storage buffer cannot match a Pipeline Layout


cwfitzgerald:
**Description**

When you provide a storage buffer pipeline layout with read-only set to false, if the shader marks the buffer is `write`, it will complain about a mismatched PLL and shader.

```
[ERROR wgpu::backend::direct] Handling wgpu errors as fatal by default
    error matching shader requirements against the pipeline
    shader global ResourceBinding { group: 0, binding: 2 } is not available in the layout pipeline layout
    storage class Storage { access: LOAD | STORE } doesn't match the shader Storage { access: STORE }
```

teoxoy:
Fixed by https://github.com/gfx-rs/wgpu/pull/3893

PR #3893, opened 2023-06-28
Make shader write&read storage buffers match non readonly layouts

fornwall:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
See [#2897, write-only storage buffer cannot match a Pipeline Layout](https://github.com/gfx-rs/wgpu/issues/2897).

**Description**
From the linked issue: When you provide a storage buffer pipeline layout with read-only set to false, if the shader marks the buffer is write, it will complain about a mismatched PLL and shader

**Testing**
Mark a storage storage buffer as `write` in a shader, and use it with a buffer pipeline layout with read-only set to false. There is also a unit test of the `address_space_matches` function.

### DIFF
--- a/wgpu-core/src/validation.rs
+++ b/wgpu-core/src/validation.rs
@@ -409,7 +409,7 @@ impl Resource {
                                 )
                             }
                         };
-                        if self.class != class {
+                        if !address_space_matches(self.class, class) {
                             return Err(BindingError::WrongAddressSpace {
                                 binding: class,
                                 shader: self.class,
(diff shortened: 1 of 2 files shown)


fornwall:
After doing this, I realised that perhaps write-only storage buffers are not allowed in WGSL? From https://www.w3.org/TR/WGSL/#storage-buffer:

> A variable in the [storage](https://www.w3.org/TR/WGSL/#address-spaces-storage) address space is a storage buffer variable. Its [store type](https://www.w3.org/TR/WGSL/#store-type) must be a [host-shareable](https://www.w3.org/TR/WGSL/#host-shareable) type and must satisfy the [address space layout constraints](https://www.w3.org/TR/WGSL/#address-space-layout-constraints). The variable may be declared with a [read](https://www.w3.org/TR/WGSL/#access-read) or [read_write](https://www.w3.org/TR/WGSL/#access-read_write) access mode; the default is read.

If so, should #2897 be closed instead?

cwfitzgerald:
Welp, lets go ahead and merge this as it fixes the current behavior, thanks! But apparently they aren't anymore. There was never really a point, but whatever. Removing that is a separate issue.

wgpu-core/src/validation.rs
```diff
@@ -1237,3 +1237,67 @@ impl Interface {
...
+                naga::StorageAccess::STORE,
+                naga::StorageAccess::LOAD | naga::StorageAccess::STORE,
+                true,
+            ),
+            (
+                naga::StorageAccess::LOAD,
+                naga::StorageAccess::LOAD | naga::StorageAccess::STORE,
+                true,
+            ),
+        ] {
```
cwfitzgerald: Minor nit, could you move the list of tests into it's own variable, to make indenting a bit less complicated

wgpu-core/src/validation.rs
```diff
@@ -1237,3 +1237,67 @@ impl Interface {
...
+fn address_space_matches(shader: naga::AddressSpace, binding: naga::AddressSpace) -> bool {
+    match (shader, binding) {
+        (
+            naga::AddressSpace::Storage {
+                access: access_shader,
+            },
+            naga::AddressSpace::Storage {
+                access: access_pipeline,
+            },
+        ) => access_shader & access_pipeline == access_shader,
```
cwfitzgerald: Could you add a comment about why we need this, and what it does (as it may not be immediately obvious to an observer.

```suggestion
        ) => (access_shader & access_pipeline) == access_shader,
```

I always confuse the order of operations here, so let's be extra clear

cwfitzgerald (changes requested):
Some small nits, but code looks good

wgpu-core/src/validation.rs
fornwall: > Could you add a comment about why we need this, and what it does (as it may not be immediately obvious to an observer.

I added `// Allow read- and write-only usages to match read-write layouts:` as a comment - looks ok or do you want to clarify/rephrase?

> I always confuse the order of operations here, so let's be extra clear

Absolutely, done now.

wgpu-core/src/validation.rs
fornwall: Done.

fornwall:
@cwfitzgerald Thanks for the feedback! I addressed your comments, you could re-review?

cwfitzgerald (approved):
Great, thanks!

Issue #3792, opened 2023-05-21
`initialize_adapter_from_env_or_default` ignores `backend_bits` if `WGPU_ADAPTER_NAME` is not set


kurtschelfthout:
**Description**
`initialize_adapter_from_env_or_default` ignores `WGPU_BACKEND` if `WGPU_ADAPTER_NAME` is not set.

**Repro steps**

```rust
        let instance = wgpu::Instance::default();

        let backends = wgpu::util::backend_bits_from_env().unwrap_or_else(wgpu::Backends::all);
        println!("backends: {backends:?}");
        let adapters = instance.enumerate_adapters(backends);

        for adapter in adapters {
            let info = adapter.get_info();
            println!("adapter: {:?}", info);
        }

        let adapter = wgpu::util::initialize_adapter_from_env_or_default(&instance, backends, None)
            .await
            .expect("No suitable GPU adapters found on the system!");
        let info = adapter.get_info();
        println!(
            "Using {:#?} {} with {:#?} backend",
            info.device_type, info.name, info.backend
        );
```

I have an Intel and a discrete Nvidia GPU. When I run this with env variables like:

```
WGPU_POWER_PREF="high"
WGPU_BACKEND = "DX12"
```

It prints:

```
backends: Backends(DX12)
adapter: AdapterInfo { name: "NVIDIA GeForce RTX 2070 with Max-Q Design", vendor: 4318, device: 7956, device_type: DiscreteGpu, driver: "", driver_info: "", backend: Dx12 }
adapter: AdapterInfo { name: "Intel(R) UHD Graphics", vendor: 32902, device: 39876, device_type: IntegratedGpu, driver: "", driver_info: "", backend: Dx12 }
adapter: AdapterInfo { name: "Microsoft Basic Render Driver", vendor: 5140, device: 140, device_type: Cpu, driver: "", driver_info: "", backend: Dx12 }     
Using DiscreteGpu NVIDIA GeForce RTX 2070 with Max-Q Design with Vulkan backend (Env vars: WGPU_POWER_PREF (low|high), WGPU_ADAPTER_NAME, WGPU_BACKEND).  
```

**Expected vs observed behavior**
Despite setting the backend to DX12, wgpu uses Vulkan.

**Extra materials**
This happens because `initialize_adapter_from_env_or_default` calls `initialize_adapter_from_env`, which returns `None` if WGPU_ADAPTER_NAME is not set:

```rust
[...]

PR #3904, opened 2023-07-03
Remove backend_bits from initialize_adapter_from_env

fornwall:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
[#3792, initialize_adapter_from_env_or_default ignores backend_bits if WGPU_ADAPTER_NAME is not set](https://github.com/gfx-rs/wgpu/issues/3792)

**Description**
As described in #3792, the `initialize_adapter_from_env_or_default` function currently ignores `backend_bits` if `WGPU_ADAPTER_NAME` is not set.

Instead of going towards the path of making these utility functions more complex (or adding e.g. `backend_bits` to `RequestAdapterOptions`), let's perhaps remove `backend_bits` from these functions completely, in favour of using `InstanceDescriptor::backends`? So instead of

```rust
let instance = wgpu::Instance::default();
let backends = wgpu::util::backend_bits_from_env().unwrap_or_else(wgpu::Backends::all);
let adapter = wgpu::util::initialize_adapter_from_env_or_default(&instance, backends, None)
```

the recommended way would be:

```rust
let backends = wgpu::util::backend_bits_from_env().unwrap_or_else(wgpu::Backends::all);
let instance = wgpu::Instance::new(wgpu::InstanceDescriptor { backends, ..Default::default() });
let adapter = wgpu::util::initialize_adapter_from_env_or_default(&instance, None)
```

What do you think?

**Testing**
By existing tests, or a utility program such as listed above.

### DIFF
--- a/wgpu/src/util/init.rs
+++ b/wgpu/src/util/init.rs
@@ -37,13 +37,13 @@ pub fn power_preference_from_env() -> Option<PowerPreference> {
 
 /// Initialize the adapter obeying the WGPU_ADAPTER_NAME environment variable.
 #[cfg(not(target_arch = "wasm32"))]
-pub fn initialize_adapter_from_env(instance: &Instance, backend_bits: Backends) -> Option<Adapter> {
+pub fn initialize_adapter_from_env(instance: &Instance) -> Option<Adapter> {
     let desired_adapter_name = std::env::var("WGPU_ADAPTER_NAME")
         .as_deref()
         .map(str::to_lowercase)
         .ok()?;
 
-    let adapters = instance.enumerate_adapters(backend_bits);
+    let adapters = instance.enumerate_adapters(Backends::all());
 
     let mut chosen_adapter = None;
     for adapter in adapters {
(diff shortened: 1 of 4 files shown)


wgpu/src/util/init.rs
```diff
@@ -37,13 +37,13 @@ pub fn power_preference_from_env() -> Option<PowerPreference> {
...
 #[cfg(not(target_arch = "wasm32"))]
-pub fn initialize_adapter_from_env(instance: &Instance, backend_bits: Backends) -> Option<Adapter> {
+pub fn initialize_adapter_from_env(instance: &Instance) -> Option<Adapter> {
     let desired_adapter_name = std::env::var("WGPU_ADAPTER_NAME")
         .as_deref()
         .map(str::to_lowercase)
         .ok()?;
 
-    let adapters = instance.enumerate_adapters(backend_bits);
+    let adapters = instance.enumerate_adapters(Backends::all());
```
Wumpf, line 46: it took me really long to understand that `enumerate_adapter` does in fact not list backends that it didn't know about at creation. Should really be explicit on on its doc string I suppose.

Wumpf (approved):
I really like this idea, makes it much simpler. Thanks!

Issue #3915, opened 2023-07-07
The exceptional value of the resizing event in winit leads to crashes in examples


jinleili:
**Description**
`winit` generates a sudden exception resize event with a value of u32::MAX during the initial execution, causing it to crash when running examples such as `shadow` and `water`:

```log
Using Apple M1 Max (Metal)
2023-07-08 04:34:15.147 boids[14432:158270] CAMetalLayer ignoring invalid setDrawableSize width=4294967295.000000 height=4294967295.000000
```

```log
[2023-07-07T20:35:54Z ERROR wgpu::backend::direct] Handling wgpu errors as fatal by default
thread 'main' panicked at 'wgpu error: Validation Error

Caused by:
    In Device::create_texture
    Dimension X value 4294967295 exceeds the limit of 16384

', wgpu/src/backend/direct.rs:3024:5
```

**Platform**
macOS Sonoma Beta 3

PR #3916, opened 2023-07-07
Ignore the exception values generated by the winit resize event

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
https://github.com/gfx-rs/wgpu/issues/3915

**Testing**
Tested on macOS 14 Beta 3 and macOS 13.


examples/common/src/framework.rs
```diff
@@ -307,11 +307,20 @@ fn start<E: Example>(
                     },
                 ..
             } => {
-                log::info!("Resizing to {:?}", size);
-                config.width = size.width.max(1);
-                config.height = size.height.max(1);
-                example.resize(&config, &device, &queue);
-                surface.configure(&device, &config);
+                let max_dimension = adapter.limits().max_texture_dimension_2d;
+                if size.width > max_dimension || size.height > max_dimension {
```
cwfitzgerald, line 313: I don't get how this is happening in the first place. The limits for max texture dimension should be raised to the maximum supported by the adapter for this very reason.

jinleili: This seems to be a bug in winit itself. I have also encountered the same crash when running examples of bevy on my system.

cwfitzgerald: What is the dimension given by winit?

jinleili: u32::MAX, only happened on macOS Sonoma.

cwfitzgerald: The fuck? Can you file a bug with winit, then note that that bug is the reason for this in a comment in the code.

jinleili: Sure.

jinleili: Someone already filed this bug two weeks ago: https://github.com/rust-windowing/winit/issues/2876

cwfitzgerald: Alright, if you can just add the reasoning to the code and then this should be good.

cwfitzgerald (approved):

PR #3925, opened 2023-07-12
Bind group layout dedup

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Description**

Currently wgpu-core implement bind group layout deduplication only when it creates its own resource IDs. In other words it works for wgpu but not in Firefox.

This PR bridges the gap by allowing an optional indirection in bind group layouts: each BGL may store an ID referring to its "deduplicated" BGL. The indirection ID is never set in wgpu (when the IDs are generated by wgpu core), so in this configuration the behavior is unchanged (the indirection ID is always `None`).

When referring to a BGL the rest of the code must make sure to follow the indirection. The exception is command buffer processing which is considered hot code and where we first validate against the provided BGL ID and only follow the indirection if the initial check failed.

The main pain point with this approach is the various places where wgpu-core manually updates reference counts: we have to be careful about following the indirection to track the right BGL. I suspect that arcanization makes this better but I haven't tried to rebase this on top of it yet.

On the plus side, the plumbing to enable the validation to follow the indirection will make it very easy to decorate validation errors with labels in a followup PR.

This is the main thing preventing bevy apps from running in Firefox.

**Alternatives**

Deduplication could be re-implemented in Firefox's DOM glue, however it has a few disadvantages:
 - The deduplication logic would be duplicated.
 - More importantly, Firefox's WebGPU DOM/IPC glue is currently very simple. It is mostly a bunch of proxies for the various objects created and manipulated by wgpu-core. If BGL deduplication was implemented there, these proxies would have to reimplement much of the lifetime tracking logic that is present in wgpu-core: bind groups and pipeline layouts must keep their layouts alive, pipelines must keep their pipeline layouts alive, etc. We would prefer to preserve the simplicity of these proxies.

**Can we do better in the long run?**

I hope so. In a potential future where wgpu-core generates its own IDs even when run in Firefox, everything could be simpler and most of what's in this PR would be unnecessary. We have been discussing this with @jimblandy our plans aren't very well formed and it is rather far down the priority list but I'll try to summarize that in another issue.

**Testing**

I added a test which actually only covers wgpu's default configuration and not the new code. It's a bit tedious to test that (Need to write the test against wgpu-core, add an identity factory, etc), however that code path is covered by the WebGPU CTS so regressions will be caught in Firefox's CI (with a delay unfortunately).

### DIFF
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -436,6 +436,8 @@ pub struct BindGroupLayoutDescriptor<'a> {
 
 pub(crate) type BindEntryMap = FastHashMap<u32, wgt::BindGroupLayoutEntry>;
 
+pub type BindGroupLayouts<A> = crate::storage::Storage<BindGroupLayout<A>, BindGroupLayoutId>;
+
 /// Bind group layout.
 ///
 /// The lifetime of BGLs is a bit special. They are only referenced on CPU
@@ -450,6 +452,12 @@ pub struct BindGroupLayout<A: hal::Api> {
     pub(crate) device_id: Stored<DeviceId>,
     pub(crate) multi_ref_count: MultiRefCount,
     pub(crate) entries: BindEntryMap,
+    // When a layout created and there already exists a compatible layout the new layout
+    // keeps a reference to the older compatible one. In some places we substitute the
+    // bind group layout id with its compatible sibling.
+    // Since this substitution can come at a cost, it is skipped when wgpu-core generates
+    // its own resource IDs.
+    pub(crate) compatible_layout: Option<Valid<BindGroupLayoutId>>,
     #[allow(unused)]
     pub(crate) dynamic_count: usize,
     pub(crate) count_validator: BindingTypeMaxCountValidator,
@@ -470,6 +478,30 @@ impl<A: hal::Api> Resource for BindGroupLayout<A> {
         #[cfg(not(debug_assertions))]
         return "";
     }
+}
+
+// If a bindgroup needs to be substitued with its compatible equivalent, return the latter.
+pub(crate) fn try_get_bind_group_layout<A: HalApi>(
+    layouts: &BindGroupLayouts<A>,
+    id: BindGroupLayoutId,
+) -> Option<&BindGroupLayout<A>> {
+    let layout = layouts.get(id).ok()?;
+    if let Some(compat) = layout.compatible_layout {
+        return Some(&layouts[compat]);
+    }
+
+    Some(layout)
+}
+
+pub(crate) fn get_bind_group_layout<A: HalApi>(
+    layouts: &BindGroupLayouts<A>,
+    id: Valid<BindGroupLayoutId>,
+) -> (Valid<BindGroupLayoutId>, &BindGroupLayout<A>) {
+    let layout = &layouts[id];
+    layout
+        .compatible_layout
+        .map(|compat| (compat, &layouts[compat]))
+        .unwrap_or((id, layout))
 }
 
 #[derive(Clone, Debug, Error)]
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -1078,19 +1078,30 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 }
             }
 
-            // If there is an equivalent BGL, just bump the refcount and return it.
-            // This is only applicable for identity filters that are generating new IDs,
-            // so their inputs are `PhantomData` of size 0.
-            if mem::size_of::<Input<G, id::BindGroupLayoutId>>() == 0 {
+            let mut compatible_layout = None;
+            {
                 let (bgl_guard, _) = hub.bind_group_layouts.read(&mut token);
                 if let Some(id) =
                     Device::deduplicate_bind_group_layout(device_id, &entry_map, &*bgl_guard)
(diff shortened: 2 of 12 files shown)


wgpu-core/src/binding_model.rs
```diff
@@ -450,6 +452,12 @@ pub struct BindGroupLayout<A: hal::Api> {
     pub(crate) device_id: Stored<DeviceId>,
     pub(crate) multi_ref_count: MultiRefCount,
     pub(crate) entries: BindEntryMap,
+    // When a layout created and there already exists a compatible layout the new layout
+    // keeps a reference to the older compatible one. In some places we the substitute
+    // the bind group layout id with its compatible sibbling.
```
valaphee: Just a little typo, *we substitute", and "sibling"

kdashg:
Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1611425

wgpu-core/src/device/global.rs
```diff
@@ -1069,19 +1069,30 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+                    // If there is an equivalent BGL, just bump the refcount and return it.
+                    // This is only applicable to identity filters that generate their IDs,
+                    // which use PhantomData as their identity.
+                    // In practice this means:
+                    //  - wgpu users take this branch and return the existing
+                    //    id without using the indirection layer in BindGrouoLayout.
+                    //  - Other users like gecko or the replay tool use don't take
+                    //    the branch and instead rely on the indirection to use the
+                    //    proper bind group layout id.
+                    if std::mem::size_of::<Input<G, id::BindGroupLayoutId>>() == 0 {
```
cwfitzgerald: Could we somehow add a flag to the identity factory trait as being completely sizeless is a very easy thing to accidentally break in the future.

cwfitzgerald (dismissed):
Code looks fine minus nit

wgpu-core/src/device/global.rs
```diff
@@ -1078,19 +1078,28 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+            {
                 let (bgl_guard, _) = hub.bind_group_layouts.read(&mut token);
                 if let Some(id) =
                     Device::deduplicate_bind_group_layout(device_id, &entry_map, &*bgl_guard)
                 {
-                    return (id, None);
+                    // If there is an equivalent BGL, just bump the refcount and return it.
+                    // This is only applicable if ids are generated in wgpu. In practice:
+                    //  - wgpu users take this branch and return the existing
+                    //    id without using the indirection layer in BindGrouoLayout.
```
teoxoy:
```suggestion
                    //    id without using the indirection layer in BindGroupLayout.
```

wgpu-core/src/device/resource.rs
```diff
@@ -1368,6 +1369,11 @@ impl<A: HalApi> Device<A> {
             .iter(self_id.backend())
             .find(|&(_, bgl)| bgl.device_id.value.0 == self_id && bgl.entries == *entry_map)
             .map(|(id, value)| {
+                if let Some(comaptible) = value.compatible_layout {
+                    guard[comaptible].multi_ref_count.inc();
+                    return comaptible.0;
+                }
+
```
teoxoy: I think we should instead filter out BGL's if they have a `compatible_layout`. (making it an invariant that the chain of BGL's can be at most 2; or put another way, the referenced `compatible_layout` can't itself have a `compatible_layout`).

wgpu-core/src/device/life.rs
```diff
@@ -770,14 +770,24 @@ impl<A: HalApi> LifetimeTracker<A> {
...
-                if guard[id].multi_ref_count.dec_and_check_empty() {
-                    log::debug!("Bind group layout {:?} will be destroyed", id);
-                    #[cfg(feature = "trace")]
-                    if let Some(t) = trace {
-                        t.lock().add(trace::Action::DestroyBindGroupLayout(id.0));
-                    }
-                    if let Some(lay) = hub.bind_group_layouts.unregister_locked(id.0, &mut *guard) {
-                        self.free_resources.bind_group_layouts.push(lay.raw);
+                let mut bgl_to_check = Some(id);
+                while let Some(id) = bgl_to_check.take() {
```
teoxoy, line 774: With the new invariant, I think we should replace the loop with a fn/closure and call it once for the BGL and another for its `compatible_layout` if present.

teoxoy (changes requested):

wgpu-core/src/device/life.rs
nical: I tried but the way we access various things in here made using a function or closure very convoluted. I'm leaving this as an exercise to you if you want to make a followup.

wgpu-core/src/device/resource.rs
nical: You are right I, made the change and it looks simpler.

teoxoy (approved):

gents83:
@nical correct me if I'm wrong - what you're trying to do here is to get a BGL that already exist if descriptor is the same or if you already have the id ?
It's not compatible with Arcanization so I'll have to reimplement it and I would like to be sure to understand it 👍

nical:
@gents83 yes, BGLs may now conceptually hold a strong reference to another compatible BGL.  They need to keep the BGL they point to alive so in terms of arcanization they probably need to hold an arc.

PR #3936, opened 2023-07-18
Port to d3d12 0.7 with ownership

jrmuizel:
This replaces the WeakPtr usage with ComPtr which is no longer Copy. This means we need to do a bunch of clone()s and take some things as references.

**Checklist**

- [ ] Run `cargo clippy`.

**Connections**

**Description**

**Testing**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -481,10 +481,11 @@ checksum = "b365fabc795046672053e29c954733ec3b05e4be654ab130fe8f1f94d7051f35"
 
 [[package]]
 name = "d3d12"
-version = "0.6.0"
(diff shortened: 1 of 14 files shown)


ErichDonGubler (approved):
Noting that I have no merge privileges, I've reviewed these changes in some depth for the last couple of days. I feel satisfied that no meaningful changes have been made WRT the actual `Release` of COM objects where intended, which IMO is the major concern here. This _does_ extend the lifetime of some objects based on our testing of issues that Firefox users have encountered, but in a way that's needed to avoid issues. We've already tested consuming this change in Firefox, and it fixes some important bugs, a few CTS test cases among them.

jimblandy (approved):
This all looks reasonable to me. Together with Erich's review I think this should land.

PR #3940, opened 2023-07-19
fix(limits): properly calculate `max_bindings_per_bind_group`

ErichDonGubler:
Depends on #3942.

---

**Checklist**

-   [x] Run `cargo clippy`.
-   [x] ~Run `cargo clippy --target wasm32-unknown-unknown` if applicable.~
-   [ ] Add change to CHANGELOG.md. See simple instructions inside file.

**Connections**

Currently tracked in the Mozilla bug tracker at [bug 1844230](https://bugzilla.mozilla.org/show_bug.cgi?id=1844230).

**Description**

`AdapterInfo::max_bindings_per_bind_group` is incorrectly calculated, according to the [WebGPU spec.]:

> **_max shader stages per pipeline_** is `2`, because a `GPURenderPipeline` supports both a vertex and fragment shader.

[WebGPU spec.]: https://gpuweb.github.io/gpuweb/#max-shader-stages-per-pipeline

**Testing**

Tested by experimentally consuming this fix in Firefox, and demonstrating that the `webgpu:api,operation,adapter,requestDevice:limit,worse_than_default:limit="maxBindingsPerBindGroup"` case works correctly (TODO: link).

### DIFF
--- a/wgpu-hal/src/auxil/mod.rs
+++ b/wgpu-hal/src/auxil/mod.rs
@@ -1,3 +1,5 @@
+use wgt::Limits;
+
 #[cfg(all(any(feature = "dx11", feature = "dx12"), windows))]
 pub(super) mod dxgi;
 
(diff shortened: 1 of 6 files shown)


wgpu-hal/src/auxil/mod.rs
```diff
@@ -46,6 +46,15 @@ pub mod db {
 /// offset at some intermediate point, internally, as i32.
 pub const MAX_I32_BINDING_SIZE: u32 = 1 << 31;
 
+/// Per the [WebGPU spec.]:
+///
+/// > **_max shader stages per pipeline_** is `2`, because
+/// > a `[GPURenderPipeline](https://gpuweb.github.io/gpuweb/#gpurenderpipeline)` supports both
```
ErichDonGubler: issue: The embedded link here is superfluous, let's remove it.

ErichDonGubler:
This PR is basically ready to go, minus a `CHANGELOG` entry and fixing the `Clippy WebAssembly` CI job that's failing because of an unused symbol error from `rustc`. Not sure what the right `cfg` for that is, yet, but this should otherwise be runnable and more-or-less the shape of code I was thinking of.

wgpu-hal/src/gles/adapter.rs
```diff
@@ -499,6 +499,16 @@ impl super::Adapter {
...
+        let max_sampled_textures_per_shader_stage = super::MAX_TEXTURE_SLOTS as u32;
+        let max_samplers_per_shader_stage = super::MAX_SAMPLERS as u32;
+
+        let max_bindings_per_bind_group = (max_sampled_textures_per_shader_stage
+            + max_samplers_per_shader_stage
+            + max_storage_buffers_per_shader_stage
+            + max_storage_textures_per_shader_stage
+            + max_uniform_buffers_per_shader_stage)
+            * crate::auxil::MAX_SHADER_STAGES_PER_PIPELINE;
+
```
ErichDonGubler: suggestion: This calculation is currently duplicated across all back ends. We should make this shared code, too.

wgpu-hal/src/dx11/adapter.rs
```diff
@@ -203,6 +205,19 @@ impl super::Adapter {
...
+        let max_storage_buffers_per_shader_stage = max_uavs;
+        let max_storage_textures_per_shader_stage = max_uavs;
+        let max_uniform_buffers_per_shader_stage = max_constant_buffers;
+
+        let max_bindings_per_bind_group = (max_sampled_textures_per_shader_stage
+            + max_samplers_per_shader_stage
+            + max_storage_buffers_per_shader_stage
+            + max_storage_textures_per_shader_stage
+            + max_uniform_buffers_per_shader_stage)
+            * MAX_SHADER_STAGES_PER_PIPELINE;
```
nical: I don't think this is correct. `maxBindingsPerBindGroup`'s purpose is to limit the maximum index used for bind group bindings, because in some implementations (wgpu with the vulkan backend for example) binding descriptors end up being allocated in a dense array, the size of which is the biggest index in the group, and with some vulkan drivers that was causing some stability/safety issues with large indices.

The limit was first introduced in the spec to be 640, and it has apparently been changed to 1000 since.

So the fix should involve updating the limit to 1000 and making it work consistently over the various backends.

nical (changes requested):

wgpu-hal/src/dx11/adapter.rs
ErichDonGubler: Hmm...sounds like we need to talk about it. @nical and I resolved in Firefox team chat to discuss with the WebGPU team. I'll set this PR back as draft for now.

ErichDonGubler:
Status: currently blocked on pending discussion with the WebGPU standards body (see https://github.com/gfx-rs/wgpu/pull/3940#discussion_r1268323779).

nical:
> Status: currently blocked on pending discussion with the WebGPU standards body (see [#3940 (comment)](https://github.com/gfx-rs/wgpu/pull/3940#discussion_r1268323779)).

No need to wait, this is superseded by https://github.com/gfx-rs/wgpu/pull/3942. There is a bit of confusion in the spec around that formula but it is the spec that needs change/clarification in this case, not wgpu.

ErichDonGubler:
> > Status: currently blocked on pending discussion with the WebGPU standards body (see [#3940 (comment)](https://github.com/gfx-rs/wgpu/pull/3940#discussion_r1268323779)).
> 
> No need to wait, this is superseded by #3942. There is a bit of confusion in the spec around that formula but it is the spec that needs change/clarification in this case, not wgpu.

I agree that this PR cannot progress for now. I _don't_ agree that it should be closed right now, given that we still need to discuss your conclusion with the wider team and the standards committee first.

cwfitzgerald:
I'm going to close this for triage purposes as I don't really want unmergable PRs hanging out - please feel free to keep the branch alive, and re-open once a decision has been made upstream.

PR #3943, opened 2023-07-19
Update `max_bindings_per_bind_group` default to 1000

ErichDonGubler:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

* [Firefox downstream](https://bugzilla.mozilla.org/show_bug.cgi?id=1843250),
    which is testing against WebGPU CTS.

**Description**

Both default `Limits` and the Vulkan backend use 640 as a hard-coded limit value, but the spec. was [updated](https://github.com/gpuweb/gpuweb/issues/3864) to use 1000 instead.  

**Testing**

* TODO: Link to Firefox CI builds and tests consuming this change.


cwfitzgerald (approved):
Nice

PR #3958, opened 2023-07-21
fix: macOS - added portability bit and extension when creating vulkan instance

schell:
**Checklist**

- [x] Run `cargo clippy`.
- [x] Run `cargo clippy --target wasm32-unknown-unknown` if applicable.
- [ ] Add change to CHANGELOG.md. See simple instructions inside file.

**Description**
`wgpu` can't select the vulkan backend on macos. This should at least get us a bit further down the path. Here's the relevant discussion on the [wgpu users matrix channel](https://app.element.io/#/room/#wgpu-users:matrix.org/$Zd3HHMSCFmNu79FM0XH6D07JP4Cjgiq67jwp3WaU1V4)

**Testing**
Run an example (on macos) with `RUST_LOG=trace WGPU_BACKEND=vk VK_LOADER_DEBUG=error cargo run --bin skybox --features=vulkan-portability`

PR #3962, opened 2023-07-22
fix: vulkan only uses robust_*_access2 if the driver supports it

schell:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
This finishes up the work started in #3958 and allows the vulkan backend to be selected on macOS.

**Description**
Added `robust_[buffer|image]_access2 parameters to `PrivateCapabilities` to specify if those caps are supported by the driver, then bubbled those up to `PhysicalDeviceFeatures`.

**Testing**
On macOS, run `WGPU_BACKEND=vk VK_LOADER_DEBUG=error cargo run --bin skybox --features=vulkan-portability` and it should succeed.


cwfitzgerald (approved):
Great! Amazing stuff, going to get this working in CI

PR #3963, opened 2023-07-23
Fix the error when running the example with `cargo xtask run-wasm --bin`

jinleili:
**Checklist**

- [x] Run `cargo clippy`.

**Description**
Due to the inconsistency between the wasm-bindgen version in `xtask/Cargo.lock` and the version in `Cargo.toml` within the workspace, an error occurred when running `examples` using `cargo xtask run-wasm --bin xxx`:

```log
 Compiling wgpu-example v0.17.0 (/Users/lijinlei/Rust/forks/wgpu/examples/common)
   Compiling wgpu-skybox-example v0.17.0 (/Users/lijinlei/Rust/forks/wgpu/examples/skybox)
    Finished dev [unoptimized + debuginfo] target(s) in 1.64s
[2023-07-23T01:42:31Z WARN  walrus::module] failed to parse `name` custom section Invalid name type (at offset 28708263)
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: 

it looks like the Rust project used to create this wasm file was linked against
version of wasm-bindgen that uses a different bindgen format than this binary:

  rust wasm file schema version: 0.2.87
     this binary schema version: 0.2.86

Currently the bindgen format is unstable enough that these two schema versions
must exactly match. You can accomplish this by either updating this binary or 
the wasm-bindgen dependency in the Rust project.

You should be able to update the wasm-bindgen dependency with:

    cargo update -p wasm-bindgen --precise 0.2.86

don't forget to recompile your wasm file! Alternatively, you can update the 
binary with:

    cargo install -f wasm-bindgen-cli --version 0.2.87

if this warning fails to go away though and you're not sure what to do feel free
to open an issue at https://github.com/rustwasm/wasm-bindgen/issues!
', /Users/lijinlei/.cargo/git/checkouts/cargo-run-wasm-7fc75a2384c70976/e9b502e/src/lib.rs:267:10
fatal runtime error: failed to initiate panic, error 5
zsh: abort      cargo xtask run-wasm --bin skybox
```

**Testing**
Tested `examples` using `cargo xtask run-wasm --bin xxx`

### DIFF
--- a/xtask/Cargo.lock
+++ b/xtask/Cargo.lock
@@ -378,9 +378,9 @@ dependencies = [
 
 [[package]]
 name = "wasm-bindgen-cli-support"
-version = "0.2.86"
+version = "0.2.87"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8315d6503415e5d44ff64f1ba34aefd8264c561df17e0f1c8eb8c96bde79c45e"
+checksum = "d21c60239a09bf9bab8dfa752be4e6c637db22296b9ded493800090448692da9"
 dependencies = [
  "anyhow",
  "base64",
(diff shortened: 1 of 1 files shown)


cwfitzgerald:
Gah, wasm-bindgen is a tough one to stay in sync, thanks!

Issue #3160, opened 2022-11-01
Early Resource Deletion in Poll Tests


cwfitzgerald:
Poll tests (https://github.com/gfx-rs/wgpu/blob/master/wgpu/tests/poll.rs) are currently completely disabled due to bugs in resource tracking. I believe the problem stems from the resources being completely dropped by the time the device is maintained.

- [x] These tests should be modified to not hit the bug (to let the tests pass)
- [ ] Separate tests should be added which exploit the bug.
- [ ] The bug should be fixed.

teoxoy:
I ran the poll tests locally and this is what I'm seeing:

- DX12 on Windows 11 with Intel iGPU, Nvidia dGPU and WARP errors with
  ```
  [2022-11-10T11:04:34Z ERROR wgpu_hal::auxil::dxgi::exception] ID3D12Resource2::<final-release>: CORRUPTION: An ID3D12Resource object (0x00000204A3036F50:'Unnamed Object') is referenced by GPU operations in-flight on Command Queue (0x000002049A790500:'Unnamed ID3D12CommandQueue Object').  It is not safe to final-release objects that may have GPU operations pending.  This can result in application instability. [ EXECUTION ERROR #921: OBJECT_DELETED_WHILE_STILL_IN_USE]
  ```

- Vulkan on Windows 11 with Intel iGPU and Nvidia dGPU passes
- Vulkan on Linux with llvmpipe and swiftshader segfaults
- OpenGL on Linux with llvmpipe and Intel iGPU passes

---

In #3174 the CI seems to have the same issues (segfaults in linux/vulkan and errors with OBJECT_DELETED_WHILE_STILL_IN_USE on windows/dx12). So, finding and fixing the underlying bug might also fix that test case.

jimblandy:
It would be nice to see if this still occurs after arcanization.

cwfitzgerald:
I tested it and it does.

teoxoy:
https://github.com/gfx-rs/wgpu/pull/3873 seems to have removed all the `.skip(FailureCase::always())` from the poll tests.

I tried running the poll tests on the parent commit and they failed.

I can't tell why they are passing with https://github.com/gfx-rs/wgpu/pull/3873, I don't see any changes in `wgpu-core`.

@cwfitzgerald do you know why they are working now?

teoxoy:
[...]

PR #3964, opened 2023-07-23
Vulkan portability in CI

cwfitzgerald:
**Checklist**

- [x] Run `cargo clippy`.

Followup to fixes letting vulkan-portability run #3962

### DIFF
--- a/examples/boids/src/main.rs
+++ b/examples/boids/src/main.rs
@@ -340,6 +340,7 @@ fn boids() {
         base_test_parameters: wgpu_test::TestParameters::default()
             .downlevel_flags(wgpu::DownlevelFlags::COMPUTE_SHADERS)
             .limits(wgpu::Limits::downlevel_defaults())
+            // Lots of validation errors, maybe related to https://github.com/gfx-rs/wgpu/issues/3160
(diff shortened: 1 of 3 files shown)


examples/boids/src/main.rs
```diff
@@ -339,7 +339,8 @@ fn boids() {
         optional_features: wgpu::Features::default(),
         base_test_parameters: wgpu_test::TestParameters::default()
             .downlevel_flags(wgpu::DownlevelFlags::COMPUTE_SHADERS)
-            .limits(wgpu::Limits::downlevel_defaults()),
+            .limits(wgpu::Limits::downlevel_defaults())
```
Wumpf, line 342: do we know why boids doesn't work for moltenvk on mac?

Wumpf (approved):
if you like a feature put a test on it 💍 😄

examples/boids/src/main.rs
cwfitzgerald: There's a slew of validation errors related to early deletion of data, I believe it's related to https://github.com/gfx-rs/wgpu/issues/3160

PR #3966, opened 2023-07-23
More as_hal() methods

### LINKED ISSUE #3517: Allow hooking into command buffers on a backend-specific basis
I'm trying to hook into FidelityFX Super Resolution 2, a Vulkan/DirectX12 library. This library handles all the dispatches, but requires that we pass it a `VkCommandBuffer`.

Add a method to get a backend-specific `VkCommandBuffer` (for Vulkan, or the equivalent for DirectX12) from a `&mut wgpu::CommandEncoder`. I believe this is not exactly straightforward, as it seems like wgpu's CommandEncoder wraps many command buffers or something. 

The user should also be able to set a debug label that will show up in debuggers like RenderDoc for subsequent commands issued on the command buffer.

JMS55:
**Checklist**

- [x] Run `cargo clippy`.

**Description**
* Change `Texture::as_hal()` callback to return a value, matching other as_hal callbacks
* Add `TextureView::as_hal()`
* Add `CommandEncoder::as_hal_mut()` (closes https://github.com/gfx-rs/wgpu/issues/3517)
* Add `Adapter::texture_format_as_hal()`


JMS55:
Open question: How to ensure textures are kept alive while native/non-wgpu render work takes place?

wgpu-core/Cargo.toml
```diff
@@ -71,8 +71,8 @@ smallvec = "1"
 thiserror = "1"
 
 [dependencies.naga]
-git = "https://github.com/gfx-rs/naga"
-rev = "bac2d82a430fbfcf100ee22b7c3bc12f3d593079"
+# git = "https://github.com/gfx-rs/naga"
+# rev = "bac2d82a430fbfcf100ee22b7c3bc12f3d593079"
```
cwfitzgerald: We should keep these.

wgpu/src/lib.rs
```diff
@@ -2942,17 +2956,17 @@ impl Texture {
         target_os = "emscripten",
         feature = "webgl"
     ))]
-    pub unsafe fn as_hal<A: wgc::hal_api::HalApi, F: FnOnce(Option<&A::Texture>)>(
+    pub unsafe fn as_hal<A: wgc::hub::HalApi, F: FnOnce(Option<&A::Texture>) -> R, R>(
```
cwfitzgerald: We already have a direct dep on wgpu-hal, so we can use that here.

cwfitzgerald (approved):
Minus nits, lets just get this in, we can improve the apis later.

wgpu-core/Cargo.toml
JMS55: Ah oops, I modified that to get it working with bevy, and forgot that pushing my changes would affect this PR...

wgpu-core/src/resource.rs
```diff
@@ -499,6 +538,28 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
 
         hal_surface_callback(hal_surface)
     }
+
+    /// # Safety
+    /// - The raw command encoder handle must not be manually destroyed
+    pub unsafe fn command_encoder_as_hal_mut<
```
JMS55, line 556: Do I need to add the not wasm #cfg's here, and in a few other places?

wgpu-core/src/command/mod.rs
```diff
@@ -79,7 +79,7 @@ impl<A: hal::Api> CommandEncoder<A> {
         }
     }
 
-    fn open(&mut self) -> &mut A::CommandEncoder {
+    pub fn open(&mut self) -> &mut A::CommandEncoder {
```
JMS55, line 82: Did I need to make this method public, and CommandBuffer.encoder pub(crate)? Not sure.

wgpu-core/src/resource.rs
cwfitzgerald: We should

cwfitzgerald:
Poke - need to rebase and fix CI.

JMS55:
I've rebased, but not quite sure what cfg's to add where. It seems inconsistent, not all existing as_hal methods have them, and I'm not sure which crates need it or not. Also, not sure what cfg to use - it seems like some of them allow webgl or emscripten? Those seem wrong, surely it's supposed to be `not(any(wasm32, webgl, emscripten))`?

cwfitzgerald:
Both emscripten and webgl are `wgpu-core`/`wgpu-hal` based, so they should be allowed through. The intent is for it to be consistent, but I wouldn't be surprised if there were mistakes

cwfitzgerald (commented):
Look ma, a review

cwfitzgerald (changes requested):
Seem to have some CI issues.

xiaopengli89:
Can we add 2 more methods?
- `hal::Texture::raw_texture` return the raw texture
- `hal::CommandEncoder::raw_command_buffer` return the raw command buffer

JMS55:
I don't have the time or motivation to work on this anymore (this was originally intended for use with DLSS, but I gave up on that idea). Ideally someone will take up https://github.com/gfx-rs/wgpu/issues/4067.

PR #3975, opened 2023-07-25
Omit texture store bound checks since they are no-ops if out of bounds on all APIs

### LINKED ISSUE #3894: Texture writes don't need bounds checks on Vulkan, Metal, or Direct3D
In light of this: https://github.com/gpuweb/gpuweb/issues/3893#issuecomment-1453847182
Once naga#2265 lands,  it seems like `wgpu-hal` should change its use of Naga to always request `BoundsCheckPolicy::Unchecked` for texture stores.

cc: @teoxoy

teoxoy:
resolves https://github.com/gfx-rs/wgpu/issues/3894

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,10 @@ Bottom level categories:
 
 ## Unreleased
 
+### Changes
+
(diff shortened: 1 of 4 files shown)


ErichDonGubler (approved):
LGTM!

PR #3977, opened 2023-07-25
Satisfy `clippy` lints up to Rust 1.71.0

ErichDonGubler:
Note that `clippy::drop_refs` was renamed to `dropping_references` (N.B., upstreamed from the `clippy` namespace to `rustc`). This is a backwards-incompatible change, and the migration needs to wait until we actually have Rust 1.71 as our MSRV.

---

**Checklist**

- [x] Run `cargo clippy`.
- [ ] ~~Add change to CHANGELOG.md. See simple instructions inside file.~~

**Connections**

N/A

**Description**

Changing to MSRV of Rust 1.71 may run into issues without resolving these, and
they're backwards-compatible fixes with the current MSRV. 🤘

**Testing**

It still compiles, so this should be good.

### DIFF
--- a/examples/common/src/framework.rs
+++ b/examples/common/src/framework.rs
@@ -453,7 +453,7 @@ pub fn run<E: Example>(title: &str) {
 
 #[cfg(target_arch = "wasm32")]
 pub fn run<E: Example>(title: &str) {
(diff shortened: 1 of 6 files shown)


cwfitzgerald (approved):
LGTM

Issue #3983, opened 2023-07-09
Shader translation from SPIR-V fails targeting DX12


radgeRayden:
I have previously filed an issue about this for wgpu-native: https://github.com/gfx-rs/wgpu-native/issues/274
I also have an even smaller trace, it's just a very basic shader that indexes into a Storage Buffer. I've extracted the [HLSL output](https://gist.github.com/radgeRayden/a5de943501c964c979bbdfb6672decb1) for convenience, but the complete wgpu trace is below.
[wgpu-trace.txt](https://github.com/gfx-rs/naga/files/11994567/wgpu-trace.txt)

Also the original shader (in SPIR-V form):
[shader.zip](https://github.com/gfx-rs/naga/files/11994569/shader.zip)

teoxoy:
`CreateGraphicsPipelineState` is returning [`E_INVALIDARG`](https://learn.microsoft.com/en-us/windows/win32/direct3d12/d3d12-graphics-reference-returnvalues) which doesn't tell us much.

Could you post the trace of a debug build? We should get more info that way since the D3D12 debug layers are only enabled if `cfg!(debug_assertions)`.

radgeRayden:
> `CreateGraphicsPipelineState` is returning [`E_INVALIDARG`](https://learn.microsoft.com/en-us/windows/win32/direct3d12/d3d12-graphics-reference-returnvalues) which doesn't tell us much.
> 
> Could you post the trace of a debug build? We should get more info that way since the D3D12 debug layers are only enabled if `cfg!(debug_assertions)`.

[wgpu-trace.txt](https://github.com/gfx-rs/naga/files/12167665/wgpu-trace.txt)

(edited with non cutoff log)

teoxoy:
From the log, these are the expected bindings for the Vertex shader:

```
var GlobalVariable { name: Some("input-data"), space: Storage { access: LOAD | STORE }, binding: Some(ResourceBinding { group: 0, binding: 1 }), ty: [6], init: None }
var GlobalVariable { name: Some("uniforms"), space: Uniform, binding: Some(ResourceBinding { group: 0, binding: 0 }), ty: [13], init: None }
```

What are the bindings you are passing to wgpu?

radgeRayden:
> From the log, these are the expected bindings for the Vertex shader:
> 
> ```
[...]

PR #3984, opened 2023-07-27
Revert "Make shader write&read storage buffers match non readonly layouts"

teoxoy:
This reverts commit 17143c149c13ea5af36909ef6033e776cecad28c (#3893) as it changed validation in a non-spec compliant way.

Relevant Spec Text:

image

Even if we want to relax the validation, this won't work for dx12 as we use UAVs for read-write storage buffers and SRVs for read-only storage buffers.

I think we should remove write-only storage buffers as they are not part of the spec anymore (avoiding https://github.com/gfx-rs/wgpu/issues/2897 altogether). Related: https://github.com/gfx-rs/wgpu/issues/4411

Found while investigating https://github.com/gfx-rs/wgpu/issues/3983.

### DIFF
--- a/wgpu-core/src/validation.rs
+++ b/wgpu-core/src/validation.rs
@@ -409,7 +409,7 @@ impl Resource {
                                 )
                             }
                         };
-                        if !address_space_matches(self.class, class) {
+                        if self.class != class {
                             return Err(BindingError::WrongAddressSpace {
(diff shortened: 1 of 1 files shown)


cwfitzgerald (approved):
Makes sense to me. Thanks for catching this!

PR #3985, opened 2023-07-27
Derive storage bindings via `naga::StorageAccess` instead of `naga::GlobalUse`

### LINKED ISSUE #3983: Shader translation from SPIR-V fails targeting DX12

teoxoy:
Resolves https://github.com/gfx-rs/wgpu/issues/3983.

**Changes**
- Derive storage bindings via `naga::StorageAccess` instead of `naga::GlobalUse`.
- Removes the feature check for read-only and read-write storage textures as it's later checked by `create_bind_group_layout`.
  https://github.com/gfx-rs/wgpu/blob/335f40f85a492a5d9a69eda6835a30173f2e6b03/wgpu-core/src/device/resource.rs#L1505-L1518
- Also removes the `GlobalUse` checks from `check_binding_use` as naga is already checking those.
  https://github.com/gfx-rs/naga/blob/535701f6b2f38f7e0088701ee4cf6d8024f74a21/src/valid/interface.rs#L639-L669


ErichDonGubler:
I'll take this as a reviewer! I might not be able to do it until my tomorrow morning, but I'll strive to do it before next office hours on the Mozilla side.

CHANGELOG.md
```diff
@@ -42,6 +42,10 @@ Bottom level categories:
 
 ### Bug Fixes
 
+#### General
+
+- Derive storage bindings via `naga::StorageAccess` instead of `naga::GlobalUse`. By @teoxoy in [#3985](https://github.com/gfx-rs/wgpu/pull/3985).
```
ErichDonGubler, line 51: nitpick: Might it be worth mentioning what kinds of validation errors would now be unblocked for users?

ErichDonGubler (approved):
LGTM (noting that my review is impotent still, according to GH)! It took me some reading, but after getting to https://github.com/gfx-rs/wgpu/issues/3983#issuecomment-1653248054, the changes here clicked for me. It's great that this makes for simpler code, too.

CHANGELOG.md
teoxoy: I think it's quite niche the situation you'd have to be in where you'd get a validation error from dx12. You'd have to use an `auto` layout, declare the storage buffer in the shader with `read_write` access but only `read` from it and run on dx12 (which is not the default).

PR #3987, opened 2023-07-28
Bring `d3d12` crate in-tree

ErichDonGubler:
I've [noted](https://matrix.to/#/!FZyQrssSlHEZqrYcOb:matrix.org/$Fv-F0YkeGQcfcLE_cpuwmrqnjDs2qVIBANHHD8mU2BY?via=matrix.org&via=mozilla.org&via=kde.org) in Matrix chat that `d3d12`, as a crate, sits in an awkward place. It does not realize the potential it was apparently envisioned to have of being a robust abstraction that others can use, like with [`metal`](https://github.com/gfx-rs/metal-rs) or [`ash`](https://github.com/ash-rs/ash). It remains totally coupled to WGPU's needs. And yet, `d3d12` also reaps no benefits of its coupling. Many D3D12 APIs used WGPU's D3D backends are bound only in `wgpu-hal`, which defeats the purpose of having a dedicated layer for D3D12 interop. For those APIs that _are_ captured by the `d3d12` crate, effective development requires two compilation feedback loops _across repositories_. ~~The crate barrier also inhibits Clippy and `rustc` lints that would surely be useful for code that could otherwise be a module tree in `wgpu-hal`.~~

@cwfitzgerald [noted](https://matrix.to/#/!FZyQrssSlHEZqrYcOb:matrix.org/$2zTYNx4FmVJr_GDexw5QP3mCGZgG5UIHNB84fHl_9FM?via=matrix.org&via=mozilla.org&via=kde.org) that he'd be open to inlining `d3d12` code into `wgpu-hal`. I also took time to get some Mozilla consensus; @jimblandy laughed in apparent catharsis when I declared that `d3d12` is, proverbially, a child who either needs to get a job and pay its rent, or move back in with its parents (i.e., `wgpu-hal`). So, here's the PR making moving the entirety of the `d3d12` crate from there to here!

**This PR copies the subtree-merges `d3d12` repo into a new `d3d12` top-level folder.** ~~It also fixes up some integration issues that appear unique to the current `HEAD` of the `d3d12` repo.~~

Note that while this is inlining `d3d12` _for now_, there's no reason that we can't or shouldn't eventually try to grow the in-tree crate into a well-formed separate repository again. No plans for that are established here, but it's a definite future possibility.

---

~~I've taken pains to ensure that actual changes to code are easy to distinguish as small commits, while keeping mechanical movement of code to larger ones. Therefore, **recommended review experience is by commit**.~~

Post-merge checklist:

- [ ] Archive the `d3d12` crate?

---

**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Linked in prose above.

**Description**

See my prose above.

**Testing**

No behavior should change because of this PR. I consider successful compilation to be a sufficient test for it.


wgpu-hal/Cargo.toml
```diff
@@ -37,8 +37,40 @@ default = ["link"]
 metal = ["naga/msl-out", "block"]
 vulkan = ["naga/spv-out", "ash", "gpu-alloc", "gpu-descriptor", "libloading", "smallvec"]
 gles = ["naga/glsl-out", "glow", "khronos-egl", "libloading"]
-dx11 = ["naga/hlsl-out", "d3d12", "libloading", "winapi/d3d11", "winapi/std", "winapi/d3d11_1", "winapi/d3d11_2", "winapi/d3d11sdklayers", "winapi/dxgi1_6"]
-dx12 = ["naga/hlsl-out", "d3d12", "bit-set", "libloading", "range-alloc", "winapi/std", "winapi/winbase", "winapi/d3d12", "winapi/d3d12shader", "winapi/d3d12sdklayers", "winapi/dxgi1_6"]
+dx11 = [
```
ErichDonGubler: TODO: Split out putting implied features into their own lines in a preparatory commit.

wgpu-hal/src/auxil/dxgi/bindings/mod.rs
ErichDonGubler, line 1: TODO: There's several `clippy` lints that need to be resolved here before integration. They don't look too hard, but they should probably be separate commits prior to the one moving `d3d12` code into `wgpu-hal`.

teoxoy:
I think there is still value in keeping it as a separate crate. I thought the plan was to only move it in this repository so that we can iterate on it faster. Once we are happy with it, we can publish it as well.

ErichDonGubler:
@teoxoy:

> I think there is still value in keeping it as a separate crate. I thought the plan was to only move it in this repository so that we can iterate on it faster. Once we are happy with it, we can publish it as well.

Agreed. I think that eventually publishing well-formed D3D* bindings are well within the purview of this org. Once we have the resources to truly create it, we should do it. I've added this to the OP:

> Note that while this is inlining `d3d12` _for now_, there's no reason that we can't or shouldn't eventually try to grow the inlined code into a well-formed separate crate again. No plans for that are established here, but it's a definite future possibility.

cwfitzgerald:
Following up on this, what's the current deal?

cwfitzgerald:
Whoops wrong button.

ErichDonGubler:
Short answer: will get to this when there's some more impetus from the `wgpu` community or Firefox's side. Firefox has a few bugs in our triage queue that may quickly turn this into a priority, but it's not currently.

ErichDonGubler:
Just to ensure we have consensus: @cwfitzgerald, @jimblandy, and I have been thinking that `d3d12` should be inlined as a set of modules within `wgpu-hal`. @teoxoy, are you saying that this is acceptable iff we plan on eventually breaking it back out again into a fully formed crate? Or did you not want to go past the point of inlining `d3d12`-the-crate into the same Git repository?

cwfitzgerald:
The plan has always ended with a vague "once things are settled evaluate if it should be split back out" at the end

ErichDonGubler:
Actually, since it's relevant to the question I've posed of consensus, _and_ I think it'll be helpful to have multiple stages for this inlining, I'll limit scope in this PR. Step two would be to do the inlining anyway; there are a lot of lints to cover when we actually convert `d3d12` into modules, assuming we reach consensus with @teoxoy.

wgpu-hal/Cargo.toml
ErichDonGubler: Outdated, gonna move to a different PR.

wgpu-hal/src/auxil/dxgi/bindings/mod.rs
ErichDonGubler: Outdated, gonna move to a different PR.

ErichDonGubler:
I'm getting failures for `wasm32` tests, since CI is attempting to compile `d3d12` for the WASM target. Not sure how to fix this yet, but I'll drill down into it once Windows failures are cleared up (we're getting several, ATM).

cwfitzgerald:
We probably need a `#![cfg(not(target_arch = "wasm32"))]` at the top level

ErichDonGubler:
@cwfitzgerald: Figured it out! We want to `exclude` `d3d12` in the workspace `Cargo.toml`, because [`path` dependencies are counted as `members` by default](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields):

> All [`path` dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies) residing in the workspace directory automatically become members. …
>
> …
>
> The `exclude` key can be used to prevent paths from being included in a workspace. This can be useful if some path dependencies aren’t desired to be in the workspace at all, or using a glob pattern and you want to remove a directory.
>
> …

Fixed with a419a50e.

cwfitzgerald:
Hmm interesting - just verify clippy, tests, etc still work with it.

teoxoy:
My main concern at the time was that we'll be "inlining" all `d3d12-rs`'s crate code into our d3d12 backend. The PR was still WIP and looking at what the PR is doing wasn't enough to gauge its end state. My interpretation of  "inlining" being the traditional ["replaces a function call site with the body of the called function"](https://en.wikipedia.org/wiki/Inline_expansion).

After talking about it in a meeting with @ErichDonGubler he mentioned that wasn't the case. We would just keep the contents of the crate in a module.

This then became a question of should this be a module or a crate?

I felt like the answer was: keep it as a crate since it's already one and having to convert it to a module to then convert it back once the status quo improves felt unnecessary.

This is however not a strong opinion.

cwfitzgerald:
> I felt like the answer was: keep it as a crate since it's already one and having to convert it to a module to then convert it back once the status quo improves felt unnecessary.

Ah yeah, leaving it a crate seems fine. It being in-tree is the most important part.

ErichDonGublerUnprivileged:
@gfx-rs/wgpu

ErichDonGubler:
@cwfitzgerald (CC @teoxoy, @jimblandy): Interestingly, I kept running into crashes and errors regarding certain COM objects apparently being null after naively switching to the in-tree copy of `d3d12` ([1](https://github.com/gfx-rs/wgpu/actions/runs/6253114430/job/16977821472?pr=3987), [2](https://github.com/gfx-rs/wgpu/actions/runs/6254174102/job/16981148682?pr=3987)). I resolved this by removing the nullability of `ComPtr`, but I don't think the current set of internal APIs is great. There's a _lot_ of places where code apparently tries to be a "thin" FFI into Windows, but ends up missing opportunities to eliminate accidental complexity in our usage of DX.

I think it'd be much better to transition to using stronger `Result` types, perhaps as a follow-up PR. I've gotten everything compiling and passing again here; I'm planning on splitting commits with the current end result of code. Would you like to see any further major end-state changes before I go to the effort?

ErichDonGubler:
@cwfitzgerald, @teoxoy: I have updated the OP to indicate the current plan to just leave the in-tree copy of `d3d12` at that, and not inline any code into modules in `wgpu-hal`. Thanks!

teoxoy (dismissed):
Good stuff!

> I think it'd be much better to transition to using stronger `Result` types, perhaps as a follow-up PR. Would you like to see any further major end-state changes before I go to the effort?

Sounds great; anything getting us closer to more idiomatic rust I'm rooting for!

ErichDonGubler:
@teoxoy: Marked as `Ready for review`, with only these changes (`git diff 652ad50..d31c3ab0 -- d3d12/`):

```diff
diff --git a/d3d12/src/com.rs b/d3d12/src/com.rs
index d137f120..db790e8a 100644
--- a/d3d12/src/com.rs
+++ b/d3d12/src/com.rs
@@ -10,14 +10,12 @@ use winapi::{um::unknwnbase::IUnknown, Interface};
 pub struct ComPtr<T: Interface>(*mut T);
 
 impl<T: Interface> ComPtr<T> {
-    // TODO: remove `mut_void`, null`/`is_null`
-
     /// Create a ComPtr from a raw pointer. This will _not_ call AddRef on the pointer, assuming
     /// that it has already been called.
     ///
     /// # Safety
     ///
-    /// - if `raw` is not null, it must be a valid pointer to a COM object that implements T.
+    /// - `raw` must be a valid pointer to a COM object that implements T.
     pub unsafe fn from_reffed(raw: *mut T) -> Self {
         debug_assert!(!raw.is_null());
         ComPtr(raw)
@@ -34,29 +32,15 @@ impl<T: Interface> ComPtr<T> {
         ComPtr(raw)
     }
 
-    /// Returns the raw inner pointer. May be null.
+    /// Returns the raw inner pointer.
     pub fn as_ptr(&self) -> *const T {
         self.0
     }
 
-    /// Returns the raw inner pointer as mutable. May be null.
+    /// Returns the raw inner pointer as mutable.
     pub fn as_mut_ptr(&self) -> *mut T {
         self.0
     }
-
-    /// Returns a mutable reference to the inner pointer.
-    ///
-    /// This is useful when D3D functions initialize objects by filling in a pointer to pointer
-    /// by taking `T**` as an argument.
-    ///
-    /// # Safety
-    ///
-    /// - Any modifications done to this pointer must result in the pointer either:
-    ///   - being set to null
-    ///   - being set to a valid pointer to a COM object that implements T
-    pub fn mut_self(&mut self) -> &mut *mut T {
-        &mut self.0
-    }
 }
 
 impl<T: Interface> ComPtr<T> {
```

If I don't see an update in the next day or so, I'll assume that you're okay with what is effectively doc-comment clean-up, and just merge.

EDIT: Wrong commit, sorry! 😖

ErichDonGubler:
TIL GitHub added the `Compare` control in-line with force push events show in this feed. 👀

cwfitzgerald:
It was always there, just completely non-obvious. If you click on "force-pushed" you'll see a diff too.

d3d12/.github/workflows/ci.yml
```diff
@@ -0,0 +1,42 @@
+name: CI
```
cwfitzgerald: This shouldn't be included

d3d12/.gitignore
```diff
@@ -0,0 +1,12 @@
+# Generated by Cargo
```
cwfitzgerald, line 1: Same hjere

d3d12/rustfmt.toml
cwfitzgerald, line 1: This file shouldn't be here, especially empty

cwfitzgerald (dismissed):

d3d12/.github/workflows/ci.yml
ErichDonGubler: I think cribbing CI from `wgpu` would be an overall improvement, so I see not reason not to do this for now. 🙂

Done with 6d355881.

d3d12/rustfmt.toml
ErichDonGubler: Removed with 0614445a.

d3d12/.gitignore
ErichDonGubler: Removing this actually seems undesirable, since this directory is its technically own crate outside the main workspace. We happen to still ignore it because of how the `.gitignore` matching rules work (i.e., `target/` in our top-level `.gitignore` matches a `target` folder in _all_ subdirectories 😶), but this `.gitignore` still makes sense, IMO.

I'll still delete it if you want, but I don't think this is the same class of redundancy as the `rustfmt.toml` or GH actions configuration.

ErichDonGubler:
@teoxoy, @cwfitzgerald: I have reverted to draft and marked this PR as ready again, re-requesting your reviews. At @kdashg's encouragement, I decided to actually try to diagnose apparent issues with `ComPtr::mut_void`. This has resulted in significant changes to this current PR, making the diff. overall much less complex.

The issues with testing appear to have been a regression introduced in https://github.com/gfx-rs/d3d12-rs/pull/50 (authored by @cwfitzgerald), which was not released or consumed here before. A summary of the problems we encountered can be found [here](https://github.com/gfx-rs/d3d12-rs/pull/50#discussion_r1337512140). I've fixed in this PR with 1ad40cf8.

d3d12/src/com.rs
```diff
@@ -0,0 +1,322 @@
...
+    /// # Safety
+    ///
+    /// This method is not `unsafe` by itself. However, readers should remember that it is unsound
+    /// to assign to the pointee returned here unless [invariants for this data structure][self]
+    /// are maintained.
+    pub fn mut_void(&mut self) -> *mut *mut c_void {
+        let refer: *mut *mut T = &mut self.0;
+        let void: *mut *mut c_void = refer.cast();
+        void
+    }
```
kdashg: Why use cast? I think we shouldn't.
It is risky to change anything vs what we know worked before.

kdashg (changes requested):

cwfitzgerald:
I think we should split this change - one MR  to bring in exactly what we depended on previously. Then follow up MRs to improve/iterate on the apis.

kdashg:
> I think we should split this change - one MR to bring in exactly what we depended on previously. Then follow up MRs to improve/iterate on the apis.

Thank you, this has been my pitch as well.

ErichDonGubler:
I will plan on changing this PR to “just” bring 0.7 (or a fixed `d3d12` upstream) in-tree.

kdashg:
Specifically:
1. revert gfx-rs/d3d12-rs#50
2. Optional: Re-fix the parts of gfx-rs/d3d12-rs#50 that don't touch ComPtr.
3. Required: Leave ComPtr untouched
4. Required: Test this new d3d12-rs with Firefox.
  4a. Ensure there are no changes to https://searchfox.org/mozilla-central/rev/57f6fbd39c0b5957e11b27b4db58b821d8e1607d/third_party/rust/d3d12/src/com.rs#37
6. Inline d3d12-rs into wgpu
7. Required: Test this new wgpu with Firefox.
6. Optional: Further improvements

cwfitzgerald:
I think we should leave d3d12 completely untouched from the commit we were depending on. That way we can land the organizational change without testing on Firefox, and reducing the pressure to land all of the subsequent changes

kdashg:
I'm also happy to do that, what I've been thinking of as "the most safe option".

ErichDonGubler:
FTR: the older discussion about inlining code from `d3d12` into `wgpu-hal` can be found in Matrix [here](https://matrix.to/#/!FZyQrssSlHEZqrYcOb:matrix.org/$2zTYNx4FmVJr_GDexw5QP3mCGZgG5UIHNB84fHl_9FM?via=matrix.org&via=mozilla.org&via=kde.org).

cwfitzgerald (dismissed):
Dismissing my own review for bookkeeping.

Cargo.toml
```diff
@@ -17,7 +17,9 @@ members = [
     "wgpu-types",
     "wgpu",
 ]
-exclude = []
+exclude = [
+    "d3d12",
```
cwfitzgerald, line 21: Why are we excluding d3d, this should be a default member of the repo?

d3d12/bors.toml
```diff
@@ -0,0 +1,5 @@
+status = [
```
cwfitzgerald, line 1: This file can definitely be nuked

d3d12/rustfmt.toml
cwfitzgerald: Seems to still be here

wgpu-hal/Cargo.toml
```diff
@@ -138,7 +138,7 @@ winapi = { version = "0.3", features = [
     "winuser",
     "dcomp",
 ] }
-d3d12 = { version = "0.7", features = ["libloading"], optional = true }
+d3d12 = { path = "../d3d12/", features = ["libloading"], optional = true }
```
cwfitzgerald, line 141: This should both be a path and a version I think, to let us publish the crate without changing this.

cwfitzgerald (changes requested):

Cargo.toml
ErichDonGubler: Per previous conversation [here](https://github.com/gfx-rs/wgpu/pull/3987#issuecomment-1728471456), this is to avoid breakage with WASM builds (since all members of the workspace are attempted to be built, but we don't actually want that in WASM builds).

wgpu-hal/Cargo.toml
ErichDonGubler: Fixed up with 22d01e452.

d3d12/bors.toml
ErichDonGubler: This sounds like an adjustment to make after merge, yes? As you [stated earlier](https://github.com/gfx-rs/wgpu/pull/3987#issuecomment-1736213961), our intent here is to bring in the original source as-is, unless it's breaking something, with refinements in follow-up PRs.

Cargo.toml
```diff
@@ -175,7 +177,6 @@ termcolor = "1.3.0"
 
 [patch.crates-io]
 #glow = { path = "../glow" }
-#d3d12 = { path = "../d3d12-rs" }
```
ErichDonGubler, line 178: @cwfitzgerald: FYI, added this in with 887868927.

d3d12/bors.toml
cwfitzgerald: Ah alright, If I said that, do that :)

Cargo.toml
cwfitzgerald: Okay lets fix this in a follow up

cwfitzgerald (approved):

PR #3988, opened 2023-07-28
List limits for Limits::downlevel_defaults and downlevel_webgl2_defaults in docs.

JustAnotherCodemonkey:
**Checklist**

- [X] Run `cargo clippy`.

**Description**
When looking at the docs for Limits, the actual limits returned by the above default functions are not listed. This PR, among a few other small changes to make things more clear, lists the values for the Limits returned by each of those functions within the docs. This PR makes the following changes:
1. Lists the values for `Limits::downlevel_defaults` and `Limits::downlevel_webgl2_defaults` in the documentation for `Limits`.
2. Changes the memory size units used the the docs for `Limits` to be their binary unit counterparts to avoid confusion (MiB instead of MB as a true megabyte is 1000 KB instead of 1024 even though it is often used to mean the latter, as is done in the docs as they stand).
3. Refactors the value of `max_buffer_size` used for both `<Limits as Default>::default` and `downlevel_defaults` from `1 << 28` to `256 << 20` for clarity and readability. This follows the pattern of both `max_uniform_buffer_binding_size` and `max_storage_buffer_binding_size` which use a similar notation.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -847,7 +847,7 @@ pub struct Limits {
     pub max_texture_array_layers: u32,
     /// Amount of bind groups that can be attached to a pipeline at the same time. Defaults to 4. Higher is "better".
     pub max_bind_groups: u32,
-    /// Maximum binding index allowed in `create_bind_group_layout`. Defaults to 1000.
+    /// Maximum binding index allowed in `create_bind_group_layout`. Defaults to 1000. Higher is "better".
     pub max_bindings_per_bind_group: u32,
     /// Amount of uniform buffer bindings that can be dynamic in a single pipeline. Defaults to 8. Higher is "better".
     pub max_dynamic_uniform_buffers_per_pipeline_layout: u32,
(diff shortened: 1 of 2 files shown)


wgpu-types/src/lib.rs
```diff
@@ -961,6 +963,37 @@ impl Default for Limits {
 
 impl Limits {
     /// These default limits are guaranteed to be compatible with GLES-3.1, and D3D11
+    /// 
+    /// Those limits are as follows:
```
Wumpf: I think it would be nice to list those in terms of deviating fromi the regular default - same for the actual implementation

wgpu-types/src/lib.rs
```diff
@@ -991,11 +1024,42 @@ impl Limits {
...
             max_compute_workgroup_size_z: 64,
             max_compute_workgroups_per_dimension: 65535,
-            max_buffer_size: 1 << 28,
+            max_buffer_size: 256 << 20,
         }
     }
 
     /// These default limits are guaranteed to be compatible with GLES-3.0, and D3D11, and WebGL2
+    /// 
+    /// Those limits are as follows:
```
Wumpf: as above

Wumpf (approved):
Thanks for the improvements!

personally I'm not a huge fan of duplicating the constant values from the code into the docs (in particular those that aren't dicated by the spec), but I get that it's quite annoying when browsing e.g. docs.rs.

wgpu-types/src/lib.rs
JustAnotherCodemonkey: 1. In terms of specifying based on deviation in the implementation, that's already done for downlevel_webgl2_defaults but not downlevel_defaults. 2. I personally much prefer seeing all the limits listed like I did as you don't have to scroll back and forth if needed; it's all right there. But if you really feel like it would be better to list them in deviations from the default, I can change it.

cwfitzgerald: Maybe bold the limits that have changed from the previous step down?

cwfitzgerald:
To make sure these stay in sync, if you change the list to a doctest, you can assert the documented list is the same as that returned by the function.

wgpu-types/src/lib.rs
JustAnotherCodemonkey: That's something I thought about just an hour ago but totally forgot about. I really want to do that as soon as I can. How should I mark it? * at the beginning of the line sound good?

Edit: oops. Yeah bold works too but may be hard to see. I can't see the original thing I'm responding to on mobile without deleting the message as I'm working on it.

JustAnotherCodemonkey:
What do you mean? How do I make a doctest without changing how the docs look in the end result? I would love to do that.

cwfitzgerald:
The documentation will end up looking different - it would be in a code block, and we'd need something other than bold to determine which ones have changed.

JustAnotherCodemonkey:
I'll try that. I'll add another commit and see what you think.

JustAnotherCodemonkey:
Ok so I marked those changed from defaults to downlevel defaults are marked with a `*` and those changed from downlevel defaults in downlevel webgl2 defaults are marked with a `+` with carryover `*`'s. I do worry about the accessibility of that it might be hard to distinguish. I've thought of some other symbol combinations but they don't make as much sense. What do you guys think?

Wumpf (approved):
nice! This is a much better solution!

PR #3992, opened 2023-07-30
Make `Queue::write_texture` work Correctly on DX12 + Add new test `write_texture_subset_2d_mips`

### LINKED ISSUE #3072: Get test write_texture_subset to pass with dx12
**Description**

In https://github.com/gfx-rs/wgpu/pull/3063 a test was added which appears to fail on dx12. The test creates a texture and uploads to a subset of the texture using `queue.write_texture`.

**Repro steps**

in `wgpu/tests/write_texture.rs`, replace

```rust
let parameters = TestParameters::default().backend_failure(wgpu::Backends::DX12);
```
 with

```rust
let parameters = TestParameters::default();
```

and run the test on Windows suing the dx12 backend.

**Platform**

Windows

39ali:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
fixes https://github.com/gfx-rs/wgpu/issues/3072

**Description**
Fixes `Queue::write_texture` failing on the DX12 backend and added a new test to test writing/copying from different mipmap levels .   
**Testing**
Run tests for write_texture.rs

### DIFF
--- a/tests/tests/write_texture.rs
+++ b/tests/tests/write_texture.rs
@@ -4,55 +4,82 @@ use wgpu_test::{initialize_test, TestParameters};
 
 use wasm_bindgen_test::*;
 
-#[test]
-#[wasm_bindgen_test]
-fn write_texture_subset_2d() {
-    let size = 256;
-    let parameters = TestParameters::default().backend_failure(wgpu::Backends::DX12);
+fn total_bytes_in_copy(
+    texture_format: wgpu::TextureFormat,
+    bytes_per_row: u32,
+    rows_per_image: u32,
+    copy_extent: wgpu::Extent3d,
+) -> u32 {
+    let block_size = texture_format.block_size(None).unwrap_or(1);
+
+    let block_dim = texture_format.block_dimensions();
+
+    let block_width = copy_extent.width / block_dim.0;
+    let block_height = copy_extent.height / block_dim.1;
+
+    let bytes_per_image = bytes_per_row * rows_per_image;
+    let mut total_bytes = bytes_per_image * (copy_extent.depth_or_array_layers - 1);
+
+    if block_height != 0 {
+        let last_row_bytes = block_width * block_size;
+        let last_image_bytes = bytes_per_row * (block_height - 1) + last_row_bytes;
+        total_bytes += last_image_bytes;
+    }
+
+    total_bytes
+}
+
+fn test(
+    format: wgpu::TextureFormat,
+    bytes_per_row: u32,
+    rows_per_image: u32,
+    total_size_in_bytes: u32,
+    mip_level: u32,
+    mips_count: u32,
+    tex_size: wgpu::Extent3d,
+    write_size: wgpu::Extent3d,
+    copy_size: wgpu::Extent3d,
+) {
+    let parameters = TestParameters::default();
     initialize_test(parameters, |ctx| {
         let tex = ctx.device.create_texture(&wgpu::TextureDescriptor {
             label: None,
             dimension: wgpu::TextureDimension::D2,
-            size: wgpu::Extent3d {
-                width: size,
-                height: size,
-                depth_or_array_layers: 1,
-            },
-            format: wgpu::TextureFormat::R8Uint,
+            size: tex_size,
+            format,
             usage: wgpu::TextureUsages::COPY_DST
                 | wgpu::TextureUsages::COPY_SRC
                 | wgpu::TextureUsages::TEXTURE_BINDING,
-            mip_level_count: 1,
+            mip_level_count: mips_count,
             sample_count: 1,
             view_formats: &[],
         });
-        let data = vec![1u8; size as usize * 2];
+
+        let val = (mip_level + 1) as u8;
+        let data = vec![val; total_size_in_bytes as usize];
+
         // Write the first two rows
         ctx.queue.write_texture(
             wgpu::ImageCopyTexture {
                 texture: &tex,
-                mip_level: 0,
+                mip_level,
                 origin: wgpu::Origin3d::ZERO,
                 aspect: wgpu::TextureAspect::All,
             },
             bytemuck::cast_slice(&data),
             wgpu::ImageDataLayout {
                 offset: 0,
-                bytes_per_row: Some(size),
-                rows_per_image: Some(size),
+                bytes_per_row: Some(bytes_per_row),
+                rows_per_image: Some(rows_per_image),
             },
-            wgpu::Extent3d {
-                width: size,
-                height: 2,
-                depth_or_array_layers: 1,
-            },
+            write_size,
         );
 
         ctx.queue.submit(None);
 
         let read_buffer = ctx.device.create_buffer(&wgpu::BufferDescriptor {
             label: None,
-            size: (size * size) as u64,
+            size: (total_size_in_bytes) as u64,
             usage: wgpu::BufferUsages::MAP_READ | wgpu::BufferUsages::COPY_DST,
             mapped_at_creation: false,
         });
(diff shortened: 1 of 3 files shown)


wgpu-hal/src/dx12/command.rs
```diff
@@ -552,7 +552,24 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
         for r in regions {
             let src_box = make_box(&wgt::Origin3d::ZERO, &r.size);
             unsafe {
-                *src_location.u.PlacedFootprint_mut() = r.to_subresource_footprint(dst.format)
+                let mut desc = dst.resource.GetDesc();
```
cwfitzgerald: texture_to_buffer seems to use "calc_subresource_for_copy", is this a thing this can use too?

cwfitzgerald (commented):
Great stuff, one minor nit

wgpu-hal/src/dx12/command.rs
```diff
@@ -552,7 +552,24 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
...
+                desc.Width = r.size.width as u64;
+                desc.Height = r.size.height;
+                desc.DepthOrArraySize = r.size.depth as u16;
+                let mut fp = d3d12_ty::D3D12_PLACED_SUBRESOURCE_FOOTPRINT::default();
+
+                self.device.GetCopyableFootprints(
```
teoxoy: Instead of calling this fn we have `to_subresource_footprint` which is also used by `copy_texture_to_buffer`. I think we should instead see what's wrong with the calculations in `to_subresource_footprint` or use `GetCopyableFootprints` there.

39ali: good point, I can't think of a reason as to not use `GetCopyableFootprints `, I'll refactor `to_subresource_footprint ` so it can use it.

wgpu-hal/src/dx12/command.rs
39ali: `calc_subresource_for_copy` is already being used for `dst_location.u.SubresourceIndex_mut()`, we use it to calc `SubresourceIndex` not `PlacedFootprint`

cwfitzgerald (dismissed):
Nice!

cwfitzgerald:
Seems to not pass tests

teoxoy:
I haven't looked at the details but we are now ignoring

```
self.buffer_layout.rows_per_image
self.buffer_layout.bytes_per_row
self.texture_base.aspect
```

39ali:
@teoxoy  `self.texture_base.aspect` is now being used , correct me if I'm wrong but I believe `rows_per_image` and `bytes_per_row`  are redundant since they're being calculated by  `GetCopyableFootprints`.

teoxoy:
`rows_per_image` and `bytes_per_row` are passed by the user and they can be higher than the minimum required to perform the copy, so we can't assume they will be the minimum.

Related: https://gpuweb.github.io/gpuweb/#abstract-opdef-validating-linear-texture-data

teoxoy:
I doubt we can use `GetCopyableFootprints` as it seems to calculate the minimums.

But maybe we can use it to figure out why our calculations were wrong previously by comparing their outputs.

39ali:
@teoxoy looks like it, I'll look into it.

39ali:
ran all tests and it's working as expected

tests/tests/write_texture.rs
```diff
@@ -90,107 +122,119 @@ fn write_texture_subset_2d() {
...
+        height: 256,
+        depth_or_array_layers: 2,
+    };
+
+    let bytes_per_row = tex_size.width;
+    let rows_per_image = tex_size.height;
+
+    let total_bytes_in_copy = total_bytes_in_copy(format, bytes_per_row, rows_per_image, copy_size);
+
+    test(
```
teoxoy, line 227: The dimension is hardcoded to be `wgpu::TextureDimension::D2` but should be 3d for this test.

tests/tests/write_texture.rs
```diff
@@ -90,107 +122,119 @@ fn write_texture_subset_2d() {
...
+        LayoutDesc {
+            bytes_per_row,
+            rows_per_image,
+            total_size_in_bytes: total_bytes_in_copy,
+        },
+        0,
+        mips_count,
+        tex_size,
+        copy_size,
+        copy_size,
```
teoxoy, line 238:
```suggestion
        tex_size,
```

We should copy back the whole texture.

tests/tests/write_texture.rs
```diff
@@ -4,55 +4,91 @@ use wgpu_test::{initialize_test, TestParameters};
...
+            mip_level_count: mips_count,
             sample_count: 1,
             view_formats: &[],
         });
-        let data = vec![1u8; size as usize * 2];
+
+        let val = (mip_level + 1) as u8;
+        let data = vec![val; total_size_in_bytes as usize];
+
         // Write the first two rows
```
teoxoy, line 70: Comment is outdated.

tests/tests/write_texture.rs
```diff
@@ -4,55 +4,91 @@ use wgpu_test::{initialize_test, TestParameters};
...
             },
+            write_size,
         );
 
         ctx.queue.submit(None);
 
         let read_buffer = ctx.device.create_buffer(&wgpu::BufferDescriptor {
             label: None,
-            size: (size * size) as u64,
+            size: (total_size_in_bytes) as u64,
```
teoxoy, line 91: Should be large enough to copy back the whole texture.

tests/tests/write_texture.rs
```diff
@@ -90,107 +122,119 @@ fn write_texture_subset_2d() {
...
+        LayoutDesc {
+            bytes_per_row,
+            rows_per_image,
+            total_size_in_bytes: total_bytes_in_copy,
+        },
+        0,
+        mips_count,
+        tex_size,
+        tex_size,
+        tex_size,
```
teoxoy, line 161: This test seems to now copy the whole texture instead of a subset (prev. `height: 2`).

wgpu-hal/src/dx12/command.rs
```diff
@@ -30,10 +30,7 @@ impl crate::BufferTextureCopy {
                 )
                 .unwrap(),
                 Width: self.size.width,
-                Height: self
-                    .buffer_layout
-                    .rows_per_image
-                    .map_or(self.size.height, |count| count * block_height),
+                Height: self.size.height,
```
teoxoy, line 33: This looks correct but I'm not sure what happens if you copy a 3D texture with a non-default `rows_per_image`.
We have `RowPitch` but no `ColPitch`.
Vulkan has [`bufferImageHeight`](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkBufferImageCopy.html) and Metal has [`sourceBytesPerImage`](https://developer.apple.com/documentation/metal/mtlblitcommandencoder/1400752-copyfrombuffer).

@cwfitzgerald any ideas?

cwfitzgerald: Dawn appears to issue one copy per layer, though I'm not 100% sure on this, as dawn has a lot of layers to their d3d12 copy code.

It seems like we should issue one copy if it's dense, otherwise one copy per subresource if it's not

teoxoy: That's my conclusion as well, Vulkan and Metal support specifying `rows_per_image` for 3D textures but D3D12 doesn't.
So, for a non-default `rows_per_image` we should set `depth: 1` and issue multiple copies by advancing `D3D12_PLACED_SUBRESOURCE_FOOTPRINT.Offset` by `rows_per_image * depth_index`.

This is a pre-existing issue, will open an issue to track it.

teoxoy (changes requested):
Marking as "Request changes"

PR #3998, opened 2023-08-01
Allow specifying minor GLES3 version

### LINKED ISSUE #3978: Ability to force GLES-3.1 on Angle D3D11
D3D11 can easily support GLES-3.1. However, Angle is not officially conformant on this configuration.

Some sort of a knob on our side that would force GLES-3.1 on Angle.
Specifically, it would add this to the context attributes:
```rust
        let mut context_attributes = vec![
            khronos_egl::CONTEXT_MAJOR_VERSION, 3,
            khronos_egl::CONTEXT_MINOR_VERSION, 1,
        ];
```

None

**Additional context**
This would make it easier to run on Windows machines that don't have proper Vulkan support. Current limitation of GLES-3.0 is way too strict there.

PJB3005:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
Fixes #3978

**Description**
Adds a `force_angle_gles31` flag to `wgpu::InstanceDescriptor`. This flag will force us to request a GLES3.1 context, which makes ANGLE provide as such on D3D11 (even though it's non-conformant).

**Testing**
I ran a bunch of the samples with this forced on. There were some issues I found, but I didn't diagnose them too much:
* `boids` doesn't seem to present correctly, but in RenderDoc it does seem to work.
* `water` doesn't render the water.
* sRGB when presenting is broken, so colors were messed up. Checking through renderdoc, it seems correct during internal rendering though.


PJB3005:
Is there a better place to put it than `InstanceDescriptor`? The breaking change from this kinda sucks.

cwfitzgerald:
Not really... this is where we need it. We probably should consider the ergonomics of InstanceDescriptor as this kind of thing likely will happen quite a bit.

Let's leave it as is for now.

wgpu-types/src/lib.rs
```diff
@@ -6322,13 +6322,16 @@ pub struct InstanceDescriptor {
     pub backends: Backends,
     /// Which DX12 shader compiler to use.
     pub dx12_shader_compiler: Dx12Compiler,
+    /// Force ANGLE to support GLES 3.1, even on nonconformant implementations like D3D11.
+    pub force_angle_gles31: bool,
```
cwfitzgerald: Could we generalize this to "force gles minor version", with valid values being 0, 1, or 2.

PJB3005: I decided to remove the `force` naming, and make it a fully general-purpose thing like the DX12 compiler. I also made it its own enum to enforce the values.

cwfitzgerald (approved):
I have some concerns about how people will interpret and fill out the instance descriptor, but this is a larger question than this PR, so lets get this in.

Issue #3855, opened 2023-06-10
Imageless Swapchain Vulkan Validation Error


LPGhatguy:
**Description**
On all wgpu programs that create a swapchain, I receive this Vulkan validation layer on my machine:

> Validation Error: [ VUID-VkRenderPassBeginInfo-framebuffer-04627 ] Object 0: handle = 0x3a6cbb0000000025, type = `VK_OBJECT_TYPE_RENDER_PASS`; | MessageID = 0x45125641 | vkCmdBeginRenderPass: Image view #0 created from an image with usage set as (`VK_IMAGE_USAGE_TRANSFER_SRC_BIT|VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT`) but using VkImageViewUsageCreateInfo the inherited usage is the subset (`VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT`) and the image info #0 used to create the framebuffer had usage set as (`VK_IMAGE_USAGE_TRANSFER_SRC_BIT|VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT`). The Vulkan spec states: If framebuffer was created with a VkFramebufferCreateInfo::flags value that included `VK_FRAMEBUFFER_CREATE_IMAGELESS_BIT`, each element of the pAttachments member of a VkRenderPassAttachmentBeginInfo structure included in the pNext chain must be a VkImageView with an inherited usage equal to the usage member of the corresponding element of VkFramebufferAttachmentsCreateInfo::pAttachmentImageInfos used to create framebuffer (https://vulkan.lunarg.com/doc/view/1.3.250.0/windows/1.3-extensions/vkspec.html#VUID-VkRenderPassBeginInfo-framebuffer-04627)

On some wgpu applications I have, this error is logged every frame. I think that's because I'm recreating the image view every frame, but I haven't looked into it.

**Repro steps**
Run the `hello-triangle` example from wgpu with Vulkan validation layers enabled.

**Platform**
* Windows 10 22H2
* Nvidia Driver 535.98
* Vulkan SDK 1.3.250.0
* Nvidia RTX 3070

JMS55:
Some bevy users have been reporting the same thing.

dphfox:
I've been experiencing the same issue on Windows 11 22H2, Nvidia driver 531.79, Vulkan SDK 1.3.246.1, Nvidia RTX 4090. The message is logged every frame.

cwfitzgerald:
[...]

PR #4000, opened 2023-08-02
Allow to use NVIDIA Nsight Graphics with wgpu applications on Vulkan 

gents83:
Add check to avoid new issue with IMAGE_STATE in PostCallRecordCreateImageView
linked to this: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6122

@kvark it would be nice if you could check if it invalidate somehow your imageless framebuffer impl of https://github.com/gfx-rs/wgpu/pull/1701 or if everything is still consistent

### DIFF
--- a/wgpu-core/src/present.rs
+++ b/wgpu-core/src/present.rs
@@ -154,7 +154,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                     label: Some("(wgpu internal) clear surface texture view"),
                     format: config.format,
                     dimension: wgt::TextureViewDimension::D2,
-                    usage: hal::TextureUses::COLOR_TARGET,
+                    usage: hal::TextureUses::empty(),
                     range: wgt::ImageSubresourceRange::default(),
                 };
                 let mut clear_views = smallvec::SmallVec::new();
--- a/wgpu-hal/src/vulkan/mod.rs
+++ b/wgpu-hal/src/vulkan/mod.rs
@@ -596,6 +596,9 @@ impl crate::Queue<Api> for Queue {
         texture: SurfaceTexture,
     ) -> Result<(), crate::SurfaceError> {
         let ssc = surface.swapchain.as_ref().unwrap();
+        if ssc.device.raw.handle() != self.device.raw.handle() {
+            return Err(crate::SurfaceError::Outdated);
+        }
 
         let swapchains = [ssc.raw];
         let image_indices = [texture.index];


gents83:
Well - yes - it seems there is a validation error then:

"Validation Error: [ VUID-VkRenderPassBeginInfo-framebuffer-04627 ] Object 0: handle = 0x2070b42d580, type = VK_OBJECT_TYPE_RENDER_PASS; | MessageID = 0x45125641 | vkCmdBeginRenderPass: Image view #0 created from an image with usage set as (VK_IMAGE_USAGE_TRANSFER_SRC_BIT|VK_IMAGE_USAGE_TRANSFER_DST_BIT|VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT) but using VkImageViewUsageCreateInfo the inherited usage is the subset (VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT) and the image info #0 used to create the framebuffer had usage set as (VK_IMAGE_USAGE_TRANSFER_SRC_BIT|VK_IMAGE_USAGE_TRANSFER_DST_BIT|VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT). The Vulkan spec states: If framebuffer was created with a VkFramebufferCreateInfo::flags value that included VK_FRAMEBUFFER_CREATE_IMAGELESS_BIT, each element of the pAttachments member of a VkRenderPassAttachmentBeginInfo structure included in the pNext chain must be a VkImageView with an inherited usage equal to the usage member of the corresponding element of VkFramebufferAttachmentsCreateInfo::pAttachmentImageInfos used to create framebuffer (https://vulkan.lunarg.com/doc/view/1.3.250.1/windows/1.3-extensions/vkspec.html#VUID-VkRenderPassBeginInfo-framebuffer-04627)"

cwfitzgerald:
This validation error exists on trunk right now.

gents83:
Oh - ok then :)

wgpu-core/src/present.rs
```diff
@@ -154,7 +154,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                     label: Some("(wgpu internal) clear surface texture view"),
                     format: config.format,
                     dimension: wgt::TextureViewDimension::D2,
-                    usage: hal::TextureUses::COLOR_TARGET,
+                    usage: hal::TextureUses::empty(),
```
cwfitzgerald, line 157: This... seems unsound

aloucks:
> This validation error exists on trunk right now.

It would be nice if the validation errors were treated as hard errors that failed the CI build and prevent the merge. It seems like the validation errors have been whack-a-mole over the years.

cwfitzgerald:
@aloucks they are CI fail. This particular validation error only comes up when you are presenting to an actual surface. Because we're running everything on headless boxes, we can't test surface integrations right now.

It's something we want to have at some point, but would require a lot of effort to implement.

gents83:
> This validation error exists on trunk right now.

https://github.com/gfx-rs/wgpu/issues/3855

PR #4003, opened 2023-08-03
Move timestamp writes to unreleased in CHANGELOG

valaphee:
**Checklist**

- [ ] Run `cargo clippy`.

**Description**
Just a follow-up as I noticed that the per-pass timestamp queries also landed, but the changelog wasn't updated. (there was also a non-breaking space character)


cwfitzgerald (approved):
Nice catch

PR #4019, opened 2023-08-09
Add request_device_error

nical:
Will be used by gecko for some bits of firefox-sepcific validation

**Connections**

Bugzilla bug https://bugzilla.mozilla.org/show_bug.cgi?id=1846661

**Description**

In Firefox we want to be able to impose stricter limits than what the device advertises (in particular for things related to large allocations, like `max_buffer_size`) and be stricter than what would make sense for native wgpu.
This means a bit of validation is done outside of wgpu and we need the entry point for creating a device in the error state.

**Testing**

This is covered by the CTS in firefox.

### DIFF
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -25,9 +25,9 @@ pub struct HalSurface<A: hal::Api> {
 #[derive(Clone, Debug, Error)]
 #[error("Limit '{name}' value {requested} is better than allowed {allowed}")]
 pub struct FailedLimit {
-    name: &'static str,
-    requested: u64,
-    allowed: u64,
+    pub name: &'static str,
+    pub requested: u64,
+    pub allowed: u64,
 }
 
 fn check_limits(requested: &wgt::Limits, allowed: &wgt::Limits) -> Vec<FailedLimit> {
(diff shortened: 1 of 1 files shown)


teoxoy:
We should be validating the `requiredLimits` in wgpu but we are only cloning them.

https://github.com/gfx-rs/wgpu/blob/e36c560748ece319a57e0c4a8fef1d1681ba20c5/wgpu-core/src/device/resource.rs#L209

But we seem to do the validation in our bindings.

https://searchfox.org/mozilla-central/rev/503938c13ef2dd174705dc0f6d0683ae43074ccc/dom/webgpu/Adapter.cpp#389

If we move the limits validation in wgpu, I'd propose that we instead add an additional parameter to `requestAdapter` allowing us to override the supported limits advertised by the adapter (since we'd have to do something similar in our bindings anyway).

teoxoy:
Actually when we get to do limits bucketing, we'll need to request an adapter from wgpu, do the bucketing and then inform wgpu of the changed adapter limits. So, we could use the same mechanism to override any limit.

nical:
This can be closed if we merge #4053.

nical:
Not needed anymore.

Issue #3326, opened 2022-12-22
Make use of `maxMemoryAllocationSize` for the `max_buffer_size` limit


teoxoy:
Use `VkPhysicalDeviceMaintenance3Properties::maxMemoryAllocationSize` (part of `VK_KHR_maintenance3`) to query for the `max_buffer_size` limit.

teoxoy:
Mistakenly closed this, we only set `max_memory_allocation_size` for the allocator.

teoxoy:
There is also `maxBufferSize`, see https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceMaintenance4PropertiesKHR.html.

We should probably use the minimum of the 2.

PR #4020, opened 2023-08-09
Use i32::MAX as the max buffer size for Dx12

nical:
This is also the limit used in Dawn.

**Checklist**

- [x] Run `cargo clippy`.

**Description**

Orthogonal but somewhat related to #4019 . We currently expose a max buffer size of u64::MAX in Dx12 which is not realistic. 

I propose using i32::MAX has the advantage of being more bad-driver-proof as some implementation may use signed integers to store sizes. We've already run into this situation with vulkan and were able to diagnose it thanks to the driver in question being open-source but it's trickier to investigate this type of things with closed source windows drivers.

Dawn uses a similar limit for its dx12 backend.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -166,6 +166,7 @@ By @fornwall in [#3904](https://github.com/gfx-rs/wgpu/pull/3904) and [#3905](ht
 #### DX12
 
 - Disable suballocation on Intel Iris(R) Xe. By @xiaopengli89 in [#3668](https://github.com/gfx-rs/wgpu/pull/3668)
+- Change the `max_buffer_size` limit from `u64::MAX` to `i32::MAX`. By @nical in [#4020](https://github.com/gfx-rs/wgpu/pull/4020)
 
 #### WebGPU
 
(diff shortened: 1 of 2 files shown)


teoxoy:
> I propose using i32::MAX has the advantage of being more bad-driver-proof as some implementation may use signed integers to store sizes. We've already run into this situation with vulkan and were able to diagnose it thanks to the driver in question being open-source but it's trickier to investigate this type of things with closed source windows drivers.

I can't find any reports where a dx12 driver behaves improperly with a buffer size > `i32::MAX`. `CreatePlacedResource` will return an OOM error if the allocation failed (and so will we).

We are also still using `u64::MAX` on Vulkan unless we are [on linux/android and the gpu is not from nvidia](https://github.com/gfx-rs/wgpu/blob/e36c560748ece319a57e0c4a8fef1d1681ba20c5/wgpu-hal/src/vulkan/adapter.rs#L709-L714). Ideally we'd use `VkPhysicalDeviceMaintenance3Properties::maxMemoryAllocationSize` (https://github.com/gfx-rs/wgpu/issues/3326) but since it's optional, we still need a default.

Specific driver bugs aside, my impression is that it's ok to treat this limit as if it didn't exist (i.e. `u64::MAX`) on backends that don't have a corresponding limit.

nical:
You won't see reports because people don't usually do silly things like that in their apps (and expect things to work). 
In the browser we have to be careful we expose people's computer to arbitrary scripts so I naturally err on the side of better safe that sorry. It's certainly less harmful for wgpu to take risks outside of the browser and we do have stricter limits in Gecko on top so I won't press with one if it's controversial.

cwfitzgerald:
Will this MR be superceeded by #4053?

nical:
It can be as far as I'm concerned although it's not unreasonable to also have a more realistic default here if it is something we want.

cwfitzgerald:
SGTM

PR #4022, opened 2023-08-09
Support dual source blending

### LINKED ISSUE #1804: Dual Source Blending
I'd like to be able to use "Dual Source Blending" with wgpu, so that I can perform sub-pixel-AA with alpha blending.

I can see some mention of whether the Metal backend supports this in the code, but I don't see anything more than this so far.
I think that what I'd like to see for this is for [BlendFactor](https://docs.rs/wgpu/0.9.0/wgpu/enum.BlendFactor.html) to have enum variants representing the second source, and a way to reference that in the shader program.

Is this a lot of work to plumb in and get implemented?

For my purposes, I'm most interested in using wgpu as a way to run against Metal and DX backends as I already have an OpenGL implementation of my renderer.

freqmod:
Support writing shaders with alt blender mode.
See https://github.com/freqmod/helicoid/tree/master/helicoid-wgpu for usage example.

**Checklist**
- [x] Run `cargo clippy`.
- [ ] Add change to CHANGELOG.md. See simple instructions inside file. TBD: This is a PR to get initial feedback.

**Connections**
Resolves https://github.com/gfx-rs/wgpu/issues/1804.
Depends on https://github.com/gfx-rs/naga/pull/2427.

**Description**
See issue 1804 (linked above)

**Testing**
Tested on linux with OpenGL, and Vulkan, and on Mac using metal. I have made some changes for Direct X but since I don't have a windows build environment I haven't tested that.
Tested in the helicoid-wgpu crate in external repo, I guess a test or example could be made if we deem it useful.

### DIFF
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -2705,7 +2705,6 @@ impl<A: HalApi> Device<A> {
                     {
                         break Some(pipeline::ColorStateError::FormatNotMultisampled(cs.format));
                     }
-
                     if let Some(blend_mode) = cs.blend {
                         for factor in [
                             blend_mode.color.src_factor,
@@ -2889,12 +2888,25 @@ impl<A: HalApi> Device<A> {
                     }
                 }
 
+                if let Some(ref interface) = shader_module.interface {
+                    // Is it here we should set shader_expects_dual_source_blending. Where should we read it?
+                    if pipeline_expects_dual_source_blending {
+                        interface.is_fragment_entry_dual_source(fragment)
+                                .expect("Internal error: Fragment entrypoint should not be set in function if not present in shader interface");
+                    }
+                }
+
                 Some(hal::ProgrammableStage {
                     module: &shader_module.raw,
                     entry_point: fragment.stage.entry_point.as_ref(),
                 })
             }
-            None => None,
+            None => {
+                if pipeline_expects_dual_source_blending {
+                    return Err(pipeline::CreateRenderPipelineError::ShaderExpectsPipelineToUseDualSourceBlendingNoFragmentStage);
+                }
+                None
+            }
         };
 
         if validated_stages.contains(wgt::ShaderStages::FRAGMENT) {
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -2889,11 +2889,7 @@ impl<A: HalApi> Device<A> {
                 }
 
                 if let Some(ref interface) = shader_module.interface {
-                    // Is it here we should set shader_expects_dual_source_blending. Where should we read it?
-                    if pipeline_expects_dual_source_blending {
-                        interface.is_fragment_entry_dual_source(fragment)
-                                .expect("Internal error: Fragment entrypoint should not be set in function if not present in shader interface");
-                    }
+                    shader_expects_dual_source_blending = interface.is_fragment_entry_dual_source(fragment);
                 }
 
                 Some(hal::ProgrammableStage {
@@ -2901,13 +2897,15 @@ impl<A: HalApi> Device<A> {
                     entry_point: fragment.stage.entry_point.as_ref(),
                 })
             }
-            None => {
-                if pipeline_expects_dual_source_blending {
-                    return Err(pipeline::CreateRenderPipelineError::ShaderExpectsPipelineToUseDualSourceBlendingNoFragmentStage);
-                }
-                None
-            }
+            None => None,
         };
+
+        if !pipeline_expects_dual_source_blending && shader_expects_dual_source_blending {
+            return Err(pipeline::CreateRenderPipelineError::ShaderExpectsPipelineToUseDualSourceBlending);
+        }
+        if pipeline_expects_dual_source_blending && !shader_expects_dual_source_blending {
+            return Err(pipeline::CreateRenderPipelineError::PipelineExpectsShaderToUseDualSourceBlending);
+        }
 
         if validated_stages.contains(wgt::ShaderStages::FRAGMENT) {
             for (i, output) in io.iter() {
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -2564,6 +2564,7 @@ impl<A: HalApi> Device<A> {
         let mut vertex_steps = Vec::with_capacity(desc.vertex.buffers.len());
         let mut vertex_buffers = Vec::with_capacity(desc.vertex.buffers.len());
         let mut total_attributes = 0;
+        let mut shader_expects_dual_source_blending = false;
         let mut pipeline_expects_dual_source_blending = false;
         for (i, vb_state) in desc.vertex.buffers.iter().enumerate() {
             vertex_steps.push(pipeline::VertexStep {
@@ -2889,7 +2890,8 @@ impl<A: HalApi> Device<A> {
                 }
 
                 if let Some(ref interface) = shader_module.interface {
-                    shader_expects_dual_source_blending = interface.is_fragment_entry_dual_source(fragment);
+                    shader_expects_dual_source_blending =
+                        interface.is_fragment_entry_dual_source(fragment).expect("Internal error: Fragment entrypoint should not be set in function if not present in shader interface");
                 }
 
                 Some(hal::ProgrammableStage {
@@ -2901,10 +2903,14 @@ impl<A: HalApi> Device<A> {
         };
 
         if !pipeline_expects_dual_source_blending && shader_expects_dual_source_blending {
-            return Err(pipeline::CreateRenderPipelineError::ShaderExpectsPipelineToUseDualSourceBlending);
+            return Err(
+                pipeline::CreateRenderPipelineError::ShaderExpectsPipelineToUseDualSourceBlending,
+            );
         }
         if pipeline_expects_dual_source_blending && !shader_expects_dual_source_blending {
-            return Err(pipeline::CreateRenderPipelineError::PipelineExpectsShaderToUseDualSourceBlending);
+            return Err(
+                pipeline::CreateRenderPipelineError::PipelineExpectsShaderToUseDualSourceBlending,
+            );
         }
 
         if validated_stages.contains(wgt::ShaderStages::FRAGMENT) {
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -1276,6 +1276,10 @@ impl<A: HalApi> Device<A> {
                 .flags
                 .contains(wgt::DownlevelFlags::MULTISAMPLED_SHADING),
         );
+        caps.set(
+            Caps::DUAL_SOURCE_BLENDING,
+            self.features.contains(wgt::Features::DUAL_SOURCE_BLENDING),
+        );
 
         let info = naga::valid::Validator::new(naga::valid::ValidationFlags::all(), caps)
             .validate(&module)
@@ -2560,6 +2564,7 @@ impl<A: HalApi> Device<A> {
         let mut vertex_steps = Vec::with_capacity(desc.vertex.buffers.len());
         let mut vertex_buffers = Vec::with_capacity(desc.vertex.buffers.len());
         let mut total_attributes = 0;
+        let mut pipeline_expects_dual_source_blending = false;
         for (i, vb_state) in desc.vertex.buffers.iter().enumerate() {
             vertex_steps.push(pipeline::VertexStep {
                 stride: vb_state.array_stride,
@@ -2701,6 +2706,25 @@ impl<A: HalApi> Device<A> {
                         break Some(pipeline::ColorStateError::FormatNotMultisampled(cs.format));
                     }
 
+                    if let Some(blend_mode) = cs.blend {
+                        for factor in [
+                            blend_mode.color.src_factor,
+                            blend_mode.color.dst_factor,
+                            blend_mode.alpha.src_factor,
+                            blend_mode.alpha.dst_factor,
+                        ] {
+                            if factor.ref_second_blend_source() {
+                                if i == 0 {
+                                    self.require_features(wgt::Features::DUAL_SOURCE_BLENDING)?;
+                                    pipeline_expects_dual_source_blending = true;
+                                    break;
+                                } else {
+                                    return Err(crate::pipeline::CreateRenderPipelineError
+                            ::BlendFactorOnUnsupportedTarget { factor, target: i as u32 });
+                                }
+                            }
+                        }
+                    }
                     break None;
                 };
                 if let Some(e) = error {
@@ -2807,6 +2831,14 @@ impl<A: HalApi> Device<A> {
                         error,
                     })?;
                 validated_stages |= flag;
+
+                let dual_source = interface.has_dual_source_blending_entry_point();
+                if !pipeline_expects_dual_source_blending && dual_source {
+                    return Err(pipeline::CreateRenderPipelineError::ShaderExpectsPipelineToUseDualSourceBlending);
+                }
+                if pipeline_expects_dual_source_blending && !dual_source {
+                    return Err(pipeline::CreateRenderPipelineError::PipelineExpectsShaderToUseDualSourceBlending);
+                }
             }
 
             hal::ProgrammableStage {
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -2714,8 +2714,8 @@ impl<A: HalApi> Device<A> {
                             blend_mode.alpha.dst_factor,
                         ] {
                             if factor.ref_second_blend_source() {
+                                self.require_features(wgt::Features::DUAL_SOURCE_BLENDING)?;
                                 if i == 0 {
-                                    self.require_features(wgt::Features::DUAL_SOURCE_BLENDING)?;
                                     pipeline_expects_dual_source_blending = true;
                                     break;
                                 } else {
@@ -2831,14 +2831,6 @@ impl<A: HalApi> Device<A> {
                         error,
                     })?;
                 validated_stages |= flag;
-
-                let dual_source = interface.has_dual_source_blending_entry_point();
-                if !pipeline_expects_dual_source_blending && dual_source {
-                    return Err(pipeline::CreateRenderPipelineError::ShaderExpectsPipelineToUseDualSourceBlending);
-                }
-                if pipeline_expects_dual_source_blending && !dual_source {
-                    return Err(pipeline::CreateRenderPipelineError::PipelineExpectsShaderToUseDualSourceBlending);
-                }
             }
 
             hal::ProgrammableStage {
@@ -2890,8 +2882,12 @@ impl<A: HalApi> Device<A> {
                 }
 
                 if let Some(ref interface) = shader_module.interface {
-                    shader_expects_dual_source_blending =
-                        interface.is_fragment_entry_dual_source(fragment).expect("Internal error: Fragment entrypoint should not be set in function if not present in shader interface");
+                    shader_expects_dual_source_blending = interface
+                        .fragment_uses_dual_source_blending(&fragment.stage.entry_point)
+                        .map_err(|error| pipeline::CreateRenderPipelineError::Stage {
+                            stage: flag,
+                            error,
+                        })?;
                 }
 
                 Some(hal::ProgrammableStage {
--- a/wgpu-core/src/pipeline.rs
+++ b/wgpu-core/src/pipeline.rs
@@ -384,6 +384,15 @@ pub enum CreateRenderPipelineError {
     },
     #[error("In the provided shader, the type given for group {group} binding {binding} has a size of {size}. As the device does not support `DownlevelFlags::BUFFER_BINDINGS_NOT_16_BYTE_ALIGNED`, the type must have a size that is a multiple of 16 bytes.")]
     UnalignedShader { group: u32, binding: u32, size: u64 },
+    #[error("Using the blend factor {factor:?} for render target {target} is not possible. Only the first render target may be used when dual-source blending.")]
+    BlendFactorOnUnsupportedTarget {
+        factor: wgt::BlendFactor,
+        target: u32,
+    },
+    #[error("Pipeline expects the shader entry point to make use of dual-source blending.")]
+    PipelineExpectsShaderToUseDualSourceBlending,
+    #[error("Shader entry point expects the pipeline to make use of dual-source blending.")]
+    ShaderExpectsPipelineToUseDualSourceBlending,
 }
 
 bitflags::bitflags! {
--- a/wgpu-core/src/pipeline.rs
+++ b/wgpu-core/src/pipeline.rs
@@ -393,6 +393,8 @@ pub enum CreateRenderPipelineError {
     PipelineExpectsShaderToUseDualSourceBlending,
     #[error("Shader entry point expects the pipeline to make use of dual-source blending.")]
     ShaderExpectsPipelineToUseDualSourceBlending,
+    #[error("Shader entry point expects the pipeline to make use of dual-source blending, but pipeline contains no fragment stage.")]
+    ShaderExpectsPipelineToUseDualSourceBlendingNoFragmentStage,
 }
 
 bitflags::bitflags! {
--- a/wgpu-core/src/pipeline.rs
+++ b/wgpu-core/src/pipeline.rs
@@ -393,8 +393,6 @@ pub enum CreateRenderPipelineError {
     PipelineExpectsShaderToUseDualSourceBlending,
     #[error("Shader entry point expects the pipeline to make use of dual-source blending.")]
     ShaderExpectsPipelineToUseDualSourceBlending,
-    #[error("Shader entry point expects the pipeline to make use of dual-source blending, but pipeline contains no fragment stage.")]
-    ShaderExpectsPipelineToUseDualSourceBlendingNoFragmentStage,
 }
 
 bitflags::bitflags! {
--- a/wgpu-core/src/validation.rs
+++ b/wgpu-core/src/validation.rs
@@ -1184,4 +1184,19 @@ impl Interface {
             .values()
             .any(|point| point.dual_source_blending)
     }
+    pub fn is_fragment_entry_dual_source<'a>(
+        &self,
+        fragment: &crate::pipeline::FragmentState<'a>,
+    ) -> Result<bool, StageError> {
+        if let Some(entry_point) = self.entry_points.get(&(
+            naga::ShaderStage::Fragment,
+            String::from(fragment.stage.entry_point.as_ref()),
+        )) {
+            Ok(entry_point.dual_source_blending)
+        } else {
+            Err(StageError::MissingEntryPoint(String::from(
+                fragment.stage.entry_point.as_ref(),
+            )))
+        }
+    }
 }
--- a/wgpu-core/src/validation.rs
+++ b/wgpu-core/src/validation.rs
@@ -116,6 +116,7 @@ struct EntryPoint {
     spec_constants: Vec<SpecializationConstant>,
     sampling_pairs: FastHashSet<(naga::Handle<Resource>, naga::Handle<Resource>)>,
     workgroup_size: [u32; 3],
+    dual_source_blending: bool,
 }
 
 #[derive(Debug)]
@@ -903,7 +904,7 @@ impl Interface {
                 ep.sampling_pairs
                     .insert((resource_mapping[&key.image], resource_mapping[&key.sampler]));
             }
-
+            ep.dual_source_blending = info.dual_source_blending;
             ep.workgroup_size = entry_point.workgroup_size;
 
             entry_points.insert((entry_point.stage, entry_point.name.clone()), ep);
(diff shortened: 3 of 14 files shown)


wgpu-types/src/lib.rs
```diff
@@ -783,7 +783,20 @@ bitflags::bitflags! {
...
+        ///
+        /// Supported platforms:
+        /// - OpenGL (with ARB_blend_func_extended)
+        /// - Metal
+        /// - Vulkan (with ARB_blend_func_extended)
+        /// TBD:
+        /// - DX11 / DX 12
+        ///
+        /// This is a native only feature.
+        const BLEND_FUNC_EXTENDED = 1 << 63;
```
```suggestion
        const DUAL_SOURCE_BLENDING = 1 << 63;
```

I think calling this dual-source blending would be preferred.

wgpu-types/src/lib.rs
```diff
@@ -783,7 +783,20 @@ bitflags::bitflags! {
...
+
+        // 63..64 available
```
```suggestion
```
no more space left

wgpu-types/src/lib.rs
```diff
@@ -1572,6 +1587,14 @@ pub enum BlendFactor {
     Constant = 11,
     /// 1.0 - Constant
     OneMinusConstant = 12,
+    /// Extension: Dual buffer alternate buffer, 1.0 - S.component
+    Src1 = 13,
+    /// Extension: Dual buffer alternate buffer, S.Component 
+    OneMinusSrc1 = 14,
```
teoxoy: The docs for these are reversed.

wgpu-types/src/lib.rs
```diff
@@ -1572,6 +1587,14 @@ pub enum BlendFactor {
...
+    /// Extension: Dual buffer alternate buffer, S.alpha
+    Src1Alpha = 15,
+    /// Extension: Dual buffer alternate buffer, 1.0 - S.alpha
+    OneMinusSrc1Alpha = 16,
```
teoxoy, line 1603: General suggestions for the docs:

- use `S1` instead of `S` (since it's source 1 instead of the implied 0)
- mention "Requires [\`Features::DUAL_SOURCE_BLENDING\`]." instead of "Extension: Dual buffer alternate buffer" (since it makes it easier to search for those blend modes)

wgpu/src/backend/web.rs
```diff
@@ -421,6 +421,12 @@ fn map_blend_factor(factor: wgt::BlendFactor) -> web_sys::GpuBlendFactor {
         BlendFactor::SrcAlphaSaturated => bf::SrcAlphaSaturated,
         BlendFactor::Constant => bf::Constant,
         BlendFactor::OneMinusConstant => bf::OneMinusConstant,
+        // Extended blend types are converted to their normal counterparts on web where they are not supported.
+        // TODO: Consider panicking here
+        BlendFactor::Src1 => bf::Src,
+        BlendFactor::OneMinusSrc1 => bf::OneMinusSrc,
+        BlendFactor::Src1Alpha => bf::SrcAlpha,
+        BlendFactor::OneMinusSrc1Alpha => bf::OneMinusSrcAlpha,
```
teoxoy: Since we won't advertise the `DUAL_SOURCE_BLENDING` feature for this backend, we should panic here.

Example where we do something similar:

https://github.com/gfx-rs/wgpu/blob/e36c560748ece319a57e0c4a8fef1d1681ba20c5/wgpu-hal/src/metal/device.rs#L807-L810

wgpu-hal/src/dx12/conv.rs
```diff
@@ -242,12 +242,12 @@ fn map_blend_factor(factor: wgt::BlendFactor, is_alpha: bool) -> d3d12_ty::D3D12
...
-        //Bf::Src1Color if is_alpha => d3d12_ty::D3D12_BLEND_SRC1_ALPHA,
-        //Bf::Src1Color => d3d12_ty::D3D12_BLEND_SRC1_COLOR,
-        //Bf::OneMinusSrc1Color if is_alpha => d3d12_ty::D3D12_BLEND_INV_SRC1_ALPHA,
-        //Bf::OneMinusSrc1Color => d3d12_ty::D3D12_BLEND_INV_SRC1_COLOR,
-        //Bf::Src1Alpha => d3d12_ty::D3D12_BLEND_SRC1_ALPHA,
-        //Bf::OneMinusSrc1Alpha => d3d12_ty::D3D12_BLEND_INV_SRC1_ALPHA,
+        Bf::Src1Color if is_alpha => d3d12_ty::D3D12_BLEND_SRC1_ALPHA,
+        Bf::Src1Color => d3d12_ty::D3D12_BLEND_SRC1_COLOR,
+        Bf::OneMinusSrc1Color if is_alpha => d3d12_ty::D3D12_BLEND_INV_SRC1_ALPHA,
+        Bf::OneMinusSrc1Color => d3d12_ty::D3D12_BLEND_INV_SRC1_COLOR,
```
teoxoy:
```suggestion
        Bf::Src1 if is_alpha => d3d12_ty::D3D12_BLEND_SRC1_ALPHA,
        Bf::Src1 => d3d12_ty::D3D12_BLEND_SRC1_COLOR,
        Bf::OneMinusSrc1 if is_alpha => d3d12_ty::D3D12_BLEND_INV_SRC1_ALPHA,
        Bf::OneMinusSrc1 => d3d12_ty::D3D12_BLEND_INV_SRC1_COLOR,
```
These seem to have an extra `Color` suffix.

wgpu-hal/src/metal/adapter.rs
```diff
@@ -773,9 +773,10 @@ impl super::PrivateCapabilities {
             } else {
                 None
             },
-            support_timestamp_query: version.at_least((11, 0), (14, 0), os_is_mac)
+            support_timestamp_query: version.at_least((12, 0), (11, 0), os_is_mac)
```
teoxoy: This seems to have been correct previously. It's guarding the fn call (see [supportsCounterSampling](https://developer.apple.com/documentation/metal/mtldevice/3564459-supportscountersampling?language=objc)).

wgpu-hal/src/metal/adapter.rs
```diff
@@ -773,9 +773,10 @@ impl super::PrivateCapabilities {
...
                 && device
                     .supports_counter_sampling(metal::MTLCounterSamplingPoint::AtStageBoundary),
+            blend_func_extended: version.at_least((11, 0), (14, 0), os_is_mac),
```
teoxoy: Dual source blending should be universally supported (see https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf).

However, the `index` attribute seems to require MSL v1.2 (see suggestion below).

wgpu-hal/src/metal/adapter.rs
```diff
@@ -807,6 +808,7 @@ impl super::PrivateCapabilities {
             | F::DEPTH32FLOAT_STENCIL8
             | F::MULTI_DRAW_INDIRECT;
 
+        features.set(F::BLEND_FUNC_EXTENDED , self.blend_func_extended);
```
```suggestion
        features.set(F::DUAL_SOURCE_BLENDING, self.msl_version >= MTLLanguageVersion::V1_2);
```

wgpu-hal/src/dx12/conv.rs
teoxoy, line 1: We don't seem to be setting the feature flag for DX12.

wgpu-types/src/lib.rs
```diff
@@ -783,7 +783,20 @@ bitflags::bitflags! {
...
-        // 62..64 available
+        /// Allows two outputs from a shader to be used by blending
+        /// For more info see the OpenGL extension ARB_blend_func_extended
+        ///
+        /// Supported platforms:
+        /// - OpenGL (with ARB_blend_func_extended)
+        /// - Metal
+        /// - Vulkan (with ARB_blend_func_extended)
+        /// TBD:
+        /// - DX11 / DX 12
```
```suggestion
        /// Allows two outputs from a shader to be used for blending.
        /// Note that dual-source blending doesn't support multiple render targets.
        ///
        /// For more info see the OpenGL ES extension GL_EXT_blend_func_extended.
        ///
        /// Supported platforms:
        /// - OpenGL ES (with GL_EXT_blend_func_extended)
        /// - Metal (with MSL 1.2+)
        /// - Vulkan (with dualSrcBlend)
        /// - DX12
```

Doc suggestions.

wgpu-types/src/lib.rs
```diff
@@ -1540,6 +1553,8 @@ impl TextureViewDimension {
 ///
 /// Corresponds to [WebGPU `GPUBlendFactor`](
 /// https://gpuweb.github.io/gpuweb/#enumdef-gpublendfactor).
+/// For the extended values see the OpenGL extension
+/// [ARB blend func extended][https://registry.khronos.org/OpenGL/extensions/ARB/ARB_blend_func_extended.txt]
```
```suggestion
```
Since we'll document this on the feature flag which we reference on the enum variants we probably don't need the reference here.

wgpu-core/src/validation.rs
teoxoy, line 1: We should validate that:
- src1 factors are only used if the new feature flag is set ([somewhere in this block](https://github.com/gfx-rs/wgpu/blob/e36c560748ece319a57e0c4a8fef1d1681ba20c5/wgpu-core/src/device/resource.rs#L2661), [example](https://github.com/gfx-rs/wgpu/blob/e36c560748ece319a57e0c4a8fef1d1681ba20c5/wgpu-core/src/device/resource.rs#L1091-L1093))
- src1 factors are only used on render target 0 (since dual-source blending doesn't support multiple render targets; see references below)
- if src1 factors are used, the fragment entry point has the (new) naga `FunctionInfo::dual_source_blending` flag set

References
- The [remarks section of `D3D12_BLEND_OP`](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_blend_op#remarks) links to [Output-Merger Stage](https://learn.microsoft.com/en-us/windows/win32/direct3d11/d3d10-graphics-programming-guide-output-merger-stage) which has a section on "Dual-Source Color Blending" indicating that it's not possible to bind multiple render attachments.
- Vulkan's `maxFragmentDualSrcAttachments` limit [is always 1 in practice](https://github.com/teoxoy/gpuinfo-vulkan-query/blob/088565a0e0e0c8fcad680ea85075dd62da167bb1/maxFragmentDualSrcAttachments%20%3E%3D%202.txt).
- Metal also doesn't support multiple render targets when using dual-source blending ([ref](https://developer.apple.com/library/archive/documentation/Miscellaneous/Conceptual/MetalProgrammingGuide/WhatsNewiniOS10tvOS10andOSX1012/WhatsNewiniOS10tvOS10andOSX1012.html)).
- OpenGL's `GL_MAX_DUAL_SOURCE_DRAW_BUFFERS` [is also 1 in practice](https://www.khronos.org/opengl/wiki/Blending#Dual_Source_Blending).

teoxoy (changes requested):
Thanks for the PR!

wgpu-hal/src/dx12/conv.rs
teoxoy: On another note, @cwfitzgerald do you know why the DX12 backend is choosing different blend factors based on `if is_alpha` (which is set if we are setting blend factors for the alpha channel)? I don't see any reason for that + none of the other backends do this.

freqmod:
Thanks for the feedback, they make sense. I think i would like to finish the naga review first, when that is done i will look into these comments.

freqmod:
I hope I have addressed the issues you pointed out. 

I tested briefly with my simple application. It works when everything is set up correctly. It stops when the feature is not defined, and it complains when blend is enabled and the shader does not have an output with it.

I also ran the unittests and they seem to pass except one which also fails in trunk. 

Cargo.toml conflicts as i have updated the hash to contain the naga changes.

wgpu-hal/src/metal/mod.rs
```diff
@@ -241,6 +241,7 @@ struct PrivateCapabilities {
     has_unified_memory: Option<bool>,
     support_timestamp_query: bool,
     support_timestamp_query_in_passes: bool,
+    blend_func_extended: bool,
```
teoxoy: Looks like this is now unused, I think we can remove it.

wgpu-hal/src/metal/adapter.rs
teoxoy: Can you revert this change?

wgpu-core/src/validation.rs
```diff
@@ -762,6 +764,67 @@ pub fn check_texture_format(
...
+                        return Err(crate::pipeline::CreateRenderPipelineError
+                            ::NonExistingSecondBlendSourceReferenced { factor });
+                    }
+                }
+                Ok(())
+            };
+        check_target(blend_mode.color.src_factor)?;
+        check_target(blend_mode.color.dst_factor)?;
+        check_target(blend_mode.alpha.src_factor)?;
+        check_target(blend_mode.alpha.dst_factor)?;
```
teoxoy: This fn is only called for all shader outputs but there might be more render targets bound.

We should check if the factors reference the 2nd blend source [here](https://github.com/gfx-rs/wgpu/blob/e36c560748ece319a57e0c4a8fef1d1681ba20c5/wgpu-core/src/device/resource.rs#L2661). We'll end up with a bool `pipeline_expects_dual_source_blending` which we can later check against naga's `FunctionInfo::dual_source_blending` (avoiding the logic above this block since we already checked those invariants in naga).

wgpu-core/src/pipeline.rs
```diff
@@ -384,6 +384,15 @@ pub enum CreateRenderPipelineError {
     },
     #[error("In the provided shader, the type given for group {group} binding {binding} has a size of {size}. As the device does not support `DownlevelFlags::BUFFER_BINDINGS_NOT_16_BYTE_ALIGNED`, the type must have a size that is a multiple of 16 bytes.")]
     UnalignedShader { group: u32, binding: u32, size: u64 },
+    #[error("Blend factor {factor:?} using S1 target used for invalid render target {target}.")]
+    AlternateBlendSourceOnUnsupportedTarget {
```
teoxoy:
```suggestion
    #[error("Using the blend factor {factor:?} for render target {target} is not possible. Only the first render target may be used when dual-source blending.")]
    BlendFactorOnUnsupportedTarget {
```

wgpu-core/src/pipeline.rs
```diff
@@ -384,6 +384,15 @@ pub enum CreateRenderPipelineError {
...
+        factor: wgt::BlendFactor,
+        target: u32,
+    },
+    #[error(
+        "Blend factor {factor:?} refer second blend source, but the shader does not provide it."
+    )]
+    NonExistingSecondBlendSourceReferenced { factor: wgt::BlendFactor },
```
teoxoy:
```suggestion
    #[error("Pipeline expects the shader entry point to make use of dual-source blending.")]
    PipelineExpectsShaderToUseDualSourceBlending,
    #[error("Shader entry point expects the pipeline to make use of dual-source blending.")]
    ShaderExpectsPipelineToUseDualSourceBlending,
```

wgpu-types/src/lib.rs
```diff
@@ -1582,6 +1596,28 @@ pub enum BlendFactor {
...
+    /// Dual buffer alternate buffer, S1.alpha
+    Src1Alpha = 15,
+    /// Dual buffer alternate buffer, 1.0 - S1.alpha
+    OneMinusSrc1Alpha = 16,
+}
+
+impl BlendFactor {
+    /** Distinguish if the blend source is refering to an alternate buffer (S1)
+    and require [`Features::DUAL_SOURCE_BLENDING`]. */
+    pub fn is_second_blend_source(&self) -> bool {
```
```suggestion
    /// Returns `true` if the blend factor references the second blend source.
    ///
    /// Note that the usage of those blend factors require [`Features::DUAL_SOURCE_BLENDING`].
    pub fn ref_second_blend_source(&self) -> bool {
```

wgpu-types/src/lib.rs
```diff
@@ -1582,6 +1596,28 @@ pub enum BlendFactor {
...
     /// 1.0 - Constant
     OneMinusConstant = 12,
+    /// Dual buffer alternate buffer, S1.Component
+    Src1 = 13,
+    /// Dual buffer alternate buffer, 1.0 - S1.component
+    OneMinusSrc1 = 14,
+    /// Dual buffer alternate buffer, S1.alpha
+    Src1Alpha = 15,
+    /// Dual buffer alternate buffer, 1.0 - S1.alpha
+    OneMinusSrc1Alpha = 16,
```
```suggestion
    /// S1.component
    Src1 = 13,
    /// 1.0 - S1.component
    OneMinusSrc1 = 14,
    /// S1.alpha
    Src1Alpha = 15,
    /// 1.0 - S1.alpha
    OneMinusSrc1Alpha = 16,
```

teoxoy (commented):
Thanks for making the changes!
Besides a few naming/phrasing suggestions, I think the validation can be improved a bit more.

wgpu-hal/src/metal/adapter.rs
freqmod: yes

wgpu-hal/src/metal/mod.rs
freqmod: removed

wgpu-core/src/validation.rs
freqmod: thanks. This looks like it is a bit more involved. I'll see if i can look at this tomorrow or next week.

wgpu-hal/src/dx12/conv.rs
cwfitzgerald: Yeah I suspect this is wrong and we always need the alpha blending. Someone needs to cross reference the webgpu spec with https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_blend

wgpu-core/src/validation.rs
cwfitzgerald: Is this resolved?

freqmod: I did upload a version with better validation logic  in 3deac94 , but I guess teoxoy should check if he is happy about it.

wgpu-core/src/device/resource.rs
```diff
@@ -2701,6 +2706,29 @@ impl<A: HalApi> Device<A> {
                         break Some(pipeline::ColorStateError::FormatNotMultisampled(cs.format));
                     }
 
+                    if let Some(blend_mode) = cs.blend {
+                        if blend_mode.color.src_factor.ref_second_blend_source()
+                            | blend_mode.color.dst_factor.ref_second_blend_source()
+                            | blend_mode.alpha.src_factor.ref_second_blend_source()
+                            | blend_mode.alpha.dst_factor.ref_second_blend_source()
+                        {
```
teoxoy:
```suggestion
                        {
                            self.require_features(wgt::Features::DUAL_SOURCE_BLENDING)?;
```

wgpu-core/src/device/resource.rs
```diff
@@ -2701,6 +2706,29 @@ impl<A: HalApi> Device<A> {
                         break Some(pipeline::ColorStateError::FormatNotMultisampled(cs.format));
                     }
 
+                    if let Some(blend_mode) = cs.blend {
+                        if blend_mode.color.src_factor.ref_second_blend_source()
+                            | blend_mode.color.dst_factor.ref_second_blend_source()
+                            | blend_mode.alpha.src_factor.ref_second_blend_source()
+                            | blend_mode.alpha.dst_factor.ref_second_blend_source()
```
teoxoy: Can we replace this with the for loop + if that we have below? That way we shouldn't have to check `ref_second_blend_source` twice.

wgpu-core/src/device/resource.rs
```diff
@@ -2892,6 +2920,36 @@ impl<A: HalApi> Device<A> {
...
+                                        .as_ref()
+                                        .map(|fragment| fragment.stage.entry_point.clone())
+                                )),
+                                })?;
+                        if !pipeline_expects_dual_source_blending && dual_source {
+                            return Err(pipeline::CreateRenderPipelineError::PipelineExpectsShaderToUseDualSourceBlending);
+                        }
+                        if pipeline_expects_dual_source_blending && !dual_source {
+                            return Err(pipeline::CreateRenderPipelineError::ShaderExpectsPipelineToUseDualSourceBlending);
+                        }
```
teoxoy: Errors seem to be reversed.

wgpu-core/src/device/resource.rs
```diff
@@ -2892,6 +2920,36 @@ impl<A: HalApi> Device<A> {
...
             }
+
+            if let Some(inteface) = desc
+                .fragment
+                .as_ref()
+                .and_then(|fragment| shader_module_guard.get(fragment.stage.module).ok())
+                .and_then(|shader_module| shader_module.interface.as_ref())
+            {
+                {
+                    if let Some(fragment) = desc.fragment.as_ref() {
```
teoxoy: Making the validation code dependent on the fragment stage will not report the case where the pipeline expects dual source blending and the fragment stage doesn't (because it's missing).

Note that we already have the [`shader_module`](https://github.com/gfx-rs/wgpu/blob/471229a48f184d9c095c52f7390816f4b4d30e92/wgpu-core/src/device/resource.rs#L2822) 2 code blocks above. So, we can call `has_dual_source_blending_entry_point` on its `interface`, set the bool result on a flag `shader_expects_dual_source_blending` and later compare the expectations.

wgpu-hal/src/metal/mod.rs
teoxoy: This seems to have come back.

teoxoy (changes requested):
This should be the last round of review.
Note that the PR needs to be rebased.

wgpu-hal/src/dx12/conv.rs
teoxoy: It seems that for alpha blending: "Blend options that end in _COLOR are not allowed." ([ref](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_render_target_blend_desc)), so we are doing the right thing here. It doesn't actually matter that this restriction exists since the alpha component is the same between the blend factors that end with `_COLOR` and the ones that end in `_ALPHA`. Vulkan and Metal don't have this restriction since it functionally doesn't matter.

wgpu-core/src/device/resource.rs
freqmod: I am not sure if i understand properly what you want here. Moving the check up to above is OK, but how do i know which entry point to validate if i do not look up the name of the entry point in the fragment state. Do you want me to  iterate over all entry points in the interface and see if any of them expects dual source blending?

freqmod: I am not sure if i properly understand how we validate in relation to the fragment stage here. 

If we are not depending on the fragment stage it means that we cannot know which entry point that require dual source blending. Therefore i am now looking at all the entry points and assuming that if any of them refers to dual source blending that is sufficient for a dual source pipeline, i hope this is the right assumption.

wgpu-core/src/device/resource.rs
freqmod: i have reversed them

wgpu-core/src/device/resource.rs
freqmod: i guess, i would think doing the for loop only if there is actually a second blend source is more efficient, but i guess it is less DRY to maintain.

wgpu-core/src/device/resource.rs
freqmod: ok

freqmod:
I have addressed the issues in this round and performed a rebase. If the assumptions in the comment to resource.rs lines 2924 to 2931 l are correct i hope indeed it is the last chages, if not i guess we may need some more rounds. All unittests except  wgpu-test::wgpu-tests shader::struct_layout::uniform_input pass on linux, but that test also fails on trunk.

freqmod:
seems like i have forgot clippy. ~Will fix,~ Fixed

wgpu-core/src/device/resource.rs
teoxoy: > If we are not depending on the fragment stage it means that we cannot know which entry point that require dual source blending.

This is correct but not my concern. My concern was a situation where the pipeline expects to do dual source blending but there is no fragment stage. I would expect a validation error.

I proposed that we add a `shader_expects_dual_source_blending` flag (in addition to the existing `pipeline_expects_dual_source_blending`) which we can then set based on the right entry point somewhere around [here](https://github.com/gfx-rs/wgpu/blob/471229a48f184d9c095c52f7390816f4b4d30e92/wgpu-core/src/device/resource.rs#L2822). And then later cross-checking the expectations.

Sorry for the confusion. Let me know what you think!

freqmod: i guess that could work, i'll see if i figure out how to code it later tonight.

freqmod: i had a look. I still am a bit confused. As far as i can see, we validate the entry point after all the other validations, and at that point pipeline_expects_dual_source_blending is true if we should check the entry point designated in the fragment, or false if not. I don't really see where the bool fits in, in particular i am unsure about where we should read it and what to  cross check it against. I did push up a version without a boolean that returns the error message as soon as it detects a problem as a starting point for further discussion.

wgpu-core/src/device/resource.rs
```diff
@@ -2857,12 +2888,25 @@ impl<A: HalApi> Device<A> {
                     }
                 }
 
+                if let Some(ref interface) = shader_module.interface {
+                    // Is it here we should set shader_expects_dual_source_blending. Where should we read it?
+                    if pipeline_expects_dual_source_blending {
+                        interface.is_fragment_entry_dual_source(fragment)
+                                .expect("Internal error: Fragment entrypoint should not be set in function if not present in shader interface");
+                    }
```
teoxoy:
```suggestion
                    shader_expects_dual_source_blending = interface.is_fragment_entry_dual_source(fragment);
```

wgpu-core/src/device/resource.rs
```diff
@@ -2857,12 +2888,25 @@ impl<A: HalApi> Device<A> {
...
                 })
             }
-            None => None,
+            None => {
+                if pipeline_expects_dual_source_blending {
+                    return Err(pipeline::CreateRenderPipelineError::ShaderExpectsPipelineToUseDualSourceBlendingNoFragmentStage);
+                }
+                None
+            }
         };
```
teoxoy:
```suggestion
            None => None,
        };

        if !pipeline_expects_dual_source_blending && shader_expects_dual_source_blending {
            return Err(pipeline::CreateRenderPipelineError::ShaderExpectsPipelineToUseDualSourceBlending);
        }
        if pipeline_expects_dual_source_blending && !shader_expects_dual_source_blending {
            return Err(pipeline::CreateRenderPipelineError::PipelineExpectsShaderToUseDualSourceBlending);
        }
```

This is what I had in mind.

wgpu-core/src/device/resource.rs
teoxoy: Thanks! I left some suggestions with what I had in mind. The direction you went with achieves the same goal though. Feel free to pick whichever one you think is best!

wgpu-hal/src/dx12/conv.rs
cwfitzgerald: What I'm worried about is the other way around - the `COLOR` ones are actually using the color as the blend factor, instead of the alpha - but the documentation isn't terribly clear on what' going on.

wgpu-core/src/device/resource.rs
freqmod: Thanks, they seem good, then i realised where you wanted to check the booleans. I added them to the PR. I will fix up the PR to make stuff compile and remove the unneeded error type.

wgpu-core/src/device/resource.rs
```diff
@@ -2889,7 +2890,8 @@ impl<A: HalApi> Device<A> {
                 }
 
                 if let Some(ref interface) = shader_module.interface {
-                    shader_expects_dual_source_blending = interface.is_fragment_entry_dual_source(fragment);
+                    shader_expects_dual_source_blending =
+                        interface.is_fragment_entry_dual_source(fragment).expect("Internal error: Fragment entrypoint should not be set in function if not present in shader interface");
```
freqmod: It's fine by me if you don't want this expect (or want to reword it), but currently it has to be here as the function it calls can fail if the entry point for some reason should not exist.

wgpu-hal/src/dx12/conv.rs
teoxoy: I'm not sure I understand, could you elaborate?

teoxoy:
I pushed a few more tweaks. Let me know if they look good and we can merge this.

Thank you for implementing this new functionality in naga & wgpu!

wgpu-core/src/validation.rs
```diff
@@ -1179,24 +1179,14 @@ impl Interface {
         Ok(outputs)
     }
 
-    pub fn has_dual_source_blending_entry_point(&self) -> bool {
```
freqmod: Removing this means that we will not validate any entry points that are not referenced by the pipeline. I thought you wanted that, but it is ok by me to not do it.

freqmod:
The changes looks good to me, thanks for you patience with the multiple review rounds.

wgpu-core/src/validation.rs
teoxoy: Since the entry point won't be used by the pipeline, we shouldn't take it into account for validation.

teoxoy (approved):

Issue #2077, opened 2021-10-15
Command encoder released without endEncoding


kvark:
**Description**
Still getting these errors from Gecko sometimes.

**Repro steps**
None

**Expected vs observed behavior**
No Metal validation error is expected

**Extra materials**
https://bugzilla.mozilla.org/show_bug.cgi?id=1729648#c5

kpreid:
Here's a repro (of at least one way to get this error; I don't know if Gecko is hitting the same case I am, but I'm mentioning this here to hopefully avoid filing duplicate issues):

0. At wgpu commit 0b61a191244da0f0d987d53614a6698097a7622f
1. Take `wgpu/examples/hello-triangle`, perhaps.
2. Comment out `rpass.set_pipeline(&render_pipeline);` — i.e. submit a render pass without having specified the pipeline.

The results I get are:

```
[2022-04-23T22:20:09Z ERROR wgpu::backend::direct] Handling wgpu errors as fatal by default
thread 'main' panicked at 'wgpu error: Validation Error

Caused by:
    In a RenderPass
      note: encoder = `<CommandBuffer-(0, 1, Metal)>`
    In a draw command, indexed:false indirect:false
    render pipeline must be set

', wgpu/src/backend/direct.rs:2286:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
-[_MTLCommandEncoder dealloc]:131: failed assertion `Command encoder released without endEncoding'
Abort trap: 6
```

By inserting logging, the assertion appears to occur while Rust is unwinding, unsurprisingly when `wgpu_hal::metal::CommandState` is dropping its `render` field of type `Option<mtl::RenderCommandEncoder>`. Experimentally adding the code

```rust
impl Drop for CommandEncoder {
    fn drop(&mut self) {
        if let Some(render) = self.state.render.take() {
            render.end_encoding();
        }
    }
}
```

to `wgpu_hal::metal::CommandEncoder` causes the program to finish panicking and exit without the assertion, but I haven't studied the code architecture and certainly can't say this is a correct general fix. (Presumably it at least should treat `blit` and `compute` similarly?)

RayMarch:
i encountered a very similar error today which might be related:
- i created a render pipeline with a `PrimitiveTopology::TriangleStrip` and `strip_index_format: Some(IndexFormat::Uint32)`.
- created a render pass, bound the pipeline
- bound an index buffer with mismatching index format `IndexFormat::Uint16`
[...]

Issue #543, opened 2020-03-30
supports_queue_family assertion fails, possibly because of two different gpus


Nejat:
I first ran across this while running the provided examples in the [coffee](https://github.com/hecrj/coffee) project using the `vulkan` feature ...

`thread 'main' panicked at 'assertion failed: suf.supports_queue_family(&adapter.raw.queue_families[0])', /home/nejat/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-native-0.4.3/src/device.rs:2059:9`

> the examples work while using the `opengl` feature

Seeing that the assertion failure was in the [wgpu-native](https://crates.io/crates/wgpu-native) crate, I went back to test a program I wrote using the [Pixels](https://github.com/parasyte/pixels) project, which also depends on the same crate, and the project that was working before now has the same exact assertion failure as above.

There is one change in the system I'm using that might help point to what is causing this issue, but I have not been able to debug it.

Originally I had a single `RTX 2080 ti` when my Pixels based application worked. I recently added a second GPU, a `Radeon Pro WX 5100`. 

Is it possible the existence of two GPUs might be the cause of the assertion failure? 

This is all on `Pop!_OS 19.10`

kvark:
Thank you for filing! That's a tough issue here, and we need a good plan for that together with @Kangz. When you are requesting an adapter, we don't know what surface you'd want to use it with, so it may end up incompatible with the surface.
We could provide a way to find out if a surface can be used with this adapter, but we also need a way to let you pick the right one. So we may need something like a list of surfaces passed in (or at least a single surface), to which an adapter has to be able to present to, at the time an adapter is requested...
Filed https://github.com/webgpu-native/webgpu-headers/pull/39 to discuss an upstream solution.

Nejat:
Excuse me if I'm over simplifying it, I do not mean to trivialize the complexity of the issue, especially knowning that Vulkan provides a much lower level interface, however could the answer be found in how OpenGL or DX12 handle the issue? 

Not only does the OpenGL version initialize on this machine with two different GPUs, I can drag the window from one GPU display to the other. 

Just a thought.

kvark:
@Nejat I can tell you how :)
-  OpenGL doesn't allow you to create a device without a surface to begin with
- DX12 and Vulkan have an explicit check for a surface being compatible with an adapter, and they expose all adapters at once

[...]

PR #4023, opened 2023-08-09
Prevent Metal from crashing when a still-open encoder is deallocated, resolve issue #2077.

bradwerth:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
None

**Description**
Metal requires that a MTLCommandEncoder must be closed before it is deallocated, otherwise Metal will assert that endEncoding was not called. This manifests as a crash in CTS tests that might otherwise be failures or timeouts.

**Testing**
CTS tests that are expected fail on Metal will no longer crash.


CHANGELOG.md
```diff
@@ -87,7 +103,7 @@ This release was fairly minor as breaking changes go.
 
 #### `wgpu` types now `!Send` `!Sync` on wasm
 
-Up until this point, wgpu has made the assumption that threads do not exist on wasm. With the rise of libraries like [`wasm_thread`](https://crates.io/crates/wasm_thread) making it easier and easier to do wasm multithreading this assumption is no longer sound. As all wgpu objects contain references into the JS heap, they cannot leave the thread they started on. 
+Up until this point, wgpu has made the assumption that threads do not exist on wasm. With the rise of libraries like [`wasm_thread`](https://crates.io/crates/wasm_thread) making it easier and easier to do wasm multithreading this assumption is no longer sound. As all wgpu objects contain references into the JS heap, they cannot leave the thread they started on.
```
ErichDonGubler, line 120: trivia: olol, trailing space. CC @valaphee.

wgpu-hal/src/metal/command.rs
ErichDonGubler, line 1: nitpick(non-blocking): Can we please cite sources in the commit message changing this file? The need for this wasn't obvious before, and I haven't found any place that documents this assertion directly yet. It's not necessary, but it'd be nice.

---

In the OP: 

> Metal requires that a MTLCommandEncoder must be closed before it is deallocated, otherwise Metal will assert that endEncoding was not called. This manifests as a crash in CTS tests that might otherwise be failures or timeouts.

Context from the [Mozilla bug we have for this](https://bugzilla.mozilla.org/show_bug.cgi?id=1846366) that seems relevant:

> In the macOS implementation, this leaves the encoder in an unexpected state that causes an assertion when the encoder is deallocated. It possibly could cause problems for other platforms.
> 
> This avoids a Metal assertion 'Command encoder released without endEncoding'.

[Relevant Apple documentation](https://developer.apple.com/documentation/metal/mtlcommandencoder) that…doesn't directly imply that this is necessary (ugh, Apple, y u no document this?), but which might be good to at least link to:

> While a command encoder is active, it has the exclusive right to append commands to its command buffer. Once you finish encoding commands, call the [`endEncoding()`](https://developer.apple.com/documentation/metal/mtlcommandencoder/1458038-endencoding) method to finish encoding the commands. To write further commands into the same command buffer, create a new command encoder.

CHANGELOG.md
```diff
@@ -38,6 +38,14 @@ Bottom level categories:
...
 -->
 
+By @bradwerth in [#4203](https://github.com/gfx-rs/wgpu/pull/4023)
+
+### Bug Fixes
+
+#### Metal
+
+- Ensure that MTLCommandEncoder calls endEncoding before it is deallocated.
+
```
ErichDonGubler: nitpick(blocking): Whoa, this `CHANGELOG` entry is unexpectedly formatted. Right now, you have outside of the `Unreleased` section:

```md
$credits

$section_headers

- $desc
```

Looks like you need to do at least:

1. Merge with the already existing `Bug Fixes` section below, inside of `Unreleased`.
2. Rearrange your content to be:

  ```
  $section_headers

  - $desc $credits 
  ```

@teoxoy's has an example of this below, if it helps.

wgpu-hal/src/metal/command.rs
ErichDonGubler, line 1: question: In Bugzilla, you mention:

> In the macOS implementation, [this is a problem]. It possibly could cause problems for other platforms.

Is trying to solve this problem for other platforms something you intend to resolve or track in another place?

CHANGELOG.md
```diff
@@ -38,6 +38,14 @@ Bottom level categories:
 - Hal
 -->
 
+By @bradwerth in [#4203](https://github.com/gfx-rs/wgpu/pull/4023)
```
kpreid: This line should be at the end of the section, not before it, and the whole section should go in the Bug Fixes section under `## Unreleased`, not above it.

ErichDonGubler (changes requested):
This change makes sense to me; I think the shape of it might change a bit, but using `Drop` to implment this seems like an elegant way to solve the problem.

If it weren't for the `CHANGELOG` entry, I'd "only" be leaving a `Comment` review, not a `Request changes` review.

kpreid (commented):
Does this PR fix #2077 and particularly the repro case I described in <https://github.com/gfx-rs/wgpu/issues/2077#issuecomment-1107660896> ?

If so, it should be mentioned in the PR description.

CHANGELOG.md
ErichDonGubler: Mid-air collision! 😆 I opened a similar conversation about this: https://github.com/gfx-rs/wgpu/pull/4023#discussion_r1289303065

wgpu-hal/src/metal/command.rs
cwfitzgerald: I would much rather these sources be in the source code itself, otherwise they'll get lost to time

bradwerth: No. For now, there's only evidence that Metal has problems with the failure to call end_encoding. I'm not sure it's necessary to move this to the superclass, and if it is necessary, it's easy enough to do when the need is identified.

jimblandy (changes requested):
This needs a test case, probably in `tests/tests/encoder.rs`.

jimblandy:
@ErichDonGubler [asked](https://github.com/gfx-rs/wgpu/pull/4023#discussion_r1289308235): 
> Is trying to solve this problem for other platforms something you intend to resolve or track in another place?

As long as the fix is in `wgpu-core`, the encoders will get closed on all backends, as `wgpu-core` is simply driving whatever `wgpu_hal::Api` implementation you gave it.

Perhaps the right place to document this requirement is in `wgpu-hal/src/lib.rs`. I would like that file to start recording the invariants that `wgpu-core` works so hard to ensure. @cwfitzgerald, does that sound right to you?

ErichDonGubler:
@jimblandy: I'm confused. Why are we talking about `wgpu-core`? This fix _isn't_ in `wgpu-core`; this is a `wgpu-hal` change specific to the `metal` module family.

wgpu-hal/src/metal/command.rs
```diff
@@ -21,6 +21,26 @@ impl Default for super::CommandState {
...
+        // encoders we are tracking. Redundant calls appear to be harmless.
+        if let Some(ref encoder) = self.blit {
+            encoder.end_encoding();
+        }
+        if let Some(ref encoder) = self.render {
+            encoder.end_encoding();
+        }
+        if let Some(ref encoder) = self.compute {
+            encoder.end_encoding();
+        }
```
ErichDonGubler: question(blocking): Is there a reason that this `Drop` isn't implemented for the individual `{Blit,Render,Compute}Encoder` types themselves?

jimblandy:
> @jimblandy: I'm confused. Why are we talking about `wgpu-core`? This fix _isn't_ in `wgpu-core`; this is a `wgpu-hal` change specific to the `metal` module family.

Sorry - Brad had shown me an earlier version of this patch that addressed the problem in generic code. This PR is completely different from that. This is indeed a Metal-only fix.

wgpu-hal/src/metal/command.rs
```diff
@@ -21,6 +21,26 @@ impl Default for super::CommandState {
     }
 }
 
+impl Drop for super::CommandState {
+    fn drop(&mut self) {
+        // Metal asserts when a MTLCommandEncoder is deallocated without a call
+        // to endEncoding. This isn't documented at
+        // https://developer.apple.com/documentation/metal/mtlcommandencoder?language=objc
```
jimblandy: Oh, I found the rule in the docs here:
https://developer.apple.com/documentation/metal/mtlcomputecommandencoder#overview

> You must always call [endEncoding()](https://developer.apple.com/documentation/metal/mtlcommandencoder/1458038-endencoding) before the encoder is released or before creating another encoder.

jimblandy: That's just for `MTLComputeCommandEncoder`. There does not seem to be a similar rule mentioned in the corresponding sections for `MTLRenderCommandEncoder` or `MTLBlitCommandEncoder`, but our crashes certainly establish that a similar rule holds for them as well.

wgpu-hal/src/metal/command.rs
jimblandy: Those types are defined by the `metal` crate. I think the usage rule that encoders must have `endEncoding` called before they are released falls under the category of things that `metal`'s users should be responsible for, not things that `metal` itself should track and take care of internally.

jimblandy: This looks an awful lot like the `discard_encoding` method. Could we just call that?

jimblandy (changes requested):

jimblandy:
It seems like wgpu-hal's Vulkan backend doesn't have any similar requirement: `wgpu_hal::vulkan::Device::destroy_command_encoder` calls [`vkDestroyCommandPool`](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#vkDestroyCommandPool), which only seems to require that none of the `vkCommandBuffer`s in the pool be actively in use by the GPU (the ["pending"](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#commandbuffers-lifecycle) state). 

But it's a little hard to follow, which is one reason I'd like to have a test case in this PR that runs on all platforms but reproduces the problem on Metal.

wgpu-hal/src/metal/command.rs
bradwerth: It's a bit obfuscated, but yes `discard_encoding` does these steps, and then also sets `raw_cmd_buf` to `None`. That seems fine, and I'll change the `Drop` implementation to call that function instead.

bradwerth:
From the failures in the attempted merge, it appears that DX12 is failing the new test and therefore needs a similar intervention. The DX12 error is `ID3D12CommandAllocator::Reset: The command allocator cannot be reset because a command list is currently being recorded with the allocator. [ EXECUTION ERROR #543: COMMAND_ALLOCATOR_CANNOT_RESET]`. That sounds very much like the Metal issue with failing to call endEncoding. I'll move the Drop implementation further up the class hierarchy so it covers all adapters.

wgpu-hal/src/metal/command.rs
```diff
@@ -1054,3 +1055,17 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
...
 }
+
+impl Drop for super::CommandEncoder {
+    fn drop(&mut self) {
+        // Metal asserts when a MTLCommandEncoder is deallocated without a call
+        // to endEncoding. This isn't documented at
+        // https://developer.apple.com/documentation/metal/mtlcommandencoder?language=objc
+        // but it manifests as a crash with the message 'Command encoder released without
+        // endEncoding'. To prevent this, we explicitiy call discard_encoding, which
+        // calls end_encoding on any still-held metal::CommandEncoders.
```
cwfitzgerald: I understand your use of the verb `asserts` here means "to raise an assert", but reads initially as "metal guarantees" 

```suggestion
        // Metal requires that a MTLCommandEncoder is ended before being deallocated.
        // This isn't documented at
        // https://developer.apple.com/documentation/metal/mtlcommandencoder?language=objc
        // but it manifests as a crash with the message 'Command encoder released without
        // endEncoding'. To prevent this, we explicitiy call discard_encoding, which
        // calls end_encoding on any still-held `metal::CommandEncoders`.
```

cwfitzgerald (commented):
LGTM after nit

cwfitzgerald:
D3D12 isn't happy with this change, unfortunately:

> [2023-08-16T13:41:19Z ERROR wgpu_hal::auxil::dxgi::exception] ID3D12CommandAllocator::Reset: The command allocator cannot be reset because a command list is currently being recorded with the allocator. [ EXECUTION ERROR #543: COMMAND_ALLOCATOR_CANNOT_RESET]

bradwerth:
> I'll move the Drop implementation further up the class hierarchy so it covers all adapters.

Since all the superclass implementations were empty, I just re-implemented it for DX12.

wgpu-hal/src/dx12/command.rs
```diff
@@ -1192,3 +1193,13 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
...
     }
 }
+
+impl Drop for super::CommandEncoder {
+    fn drop(&mut self) {
+        // DX12 throws an exception when an encoder is dropped without being closed.
+        // To prevent this, we explicitiy call discard_encoding.
+        unsafe {
+            self.discard_encoding();
+        }
```
ErichDonGubler: 🎤Registering @cwfitzgerald's [concern](https://github.com/gfx-rs/wgpu/pull/4023#issuecomment-1680915302) as a spanned conversation.

ErichDonGubler (commented):
LGTM, minus Connor's [concern](https://github.com/gfx-rs/wgpu/pull/4023#issuecomment-1680915302).

wgpu-hal/src/dx12/command.rs
ErichDonGubler: Hmm, this `Drop` implementation appears to cause at least one compile regression in CI: https://github.com/gfx-rs/wgpu/actions/runs/5882808070/job/15954619239?pr=4023

wgpu-hal/src/metal/command.rs
jimblandy: We still need the link I found mentioned in the comment.

bradwerth:
I'm not confident I'll be able to resolve the DX12 test failure with my setup in a reasonable time. I think I will disable the new test for DX12 in this PR, and treat it as a follow-up issue for somebody who is in a better position to fix it.

jimblandy (approved):
Looks fantastic.

Please file a follow-up issue for fixing the test on DX12. You can just leave it for others to triage.

jimblandy:
There's an unrelated timeout on mach aarch64. I'm going to try re-running, and if it goes away we'll rebase and merge this.

jimblandy:
I filed #4096 for the aarch64 timeout.

Issue #3896, opened 2023-06-29
Shader Debug Flag


wicast:
**Is your feature request related to a problem? Please describe.**
I want to have the ability to debug spv output, since it is very useful feature in Renderdoc

**Describe the solution you'd like**
add a debug symbol generation flag to `ShaderModuleDescriptor` to control it, also the `include_wgsl!` macro

**Additional context**
the inital spv debug feature is merged recently: [2379](https://github.com/gfx-rs/naga/pull/2379) and Renderdoc is able to use wgsl as source code(see that PR's attachments).

PR #4028, opened 2023-08-10
feat(spv): shader debug option

wicast:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
https://github.com/gfx-rs/wgpu/issues/3896

**Description**
Generate shader debug info for spv

**Testing**
There are some spv level tests in naga already. Testing with frame debugger is hard to make an autotest. 

So the following step needs done by hands:
add debug flag to the shader macro `wgpu::include_wgsl!("shader.wgsl", true));` or use struct
```rust
wgpu::ShaderModuleDescriptor {
        label: Some("shader.wgsl"),
        source: wgpu::ShaderSource::Wgsl(Cow::Borrowed(include_str!("shader.wgsl"))),
        debug: true,
    }
```
Make sure wgpu is using Vulkan backend.
Launch compiled exe from Renderdoc and capture a frame and trigger the shader debugger via vertex or pixel.
press F10 to step over and F10+shift to step back.
图片

Test on Win10 19045.3324, Nvidia.

I also tried other debugger like PIX, Nsight, but only Renderdoc support vulkan+spv debug.

I'll test Linux+Renderdoc later.


wgpu-core/src/device/resource.rs
```diff
@@ -1276,6 +1276,19 @@ impl<A: HalApi> Device<A> {
                 .contains(wgt::DownlevelFlags::MULTISAMPLED_SHADING),
         );
 
+        let debug_source = if desc.debug {
+            Some(hal::DebugSource {
+                file_name: Cow::Owned(
+                    desc.label
+                        .as_ref()
+                        .map_or("shader".to_string(), |l| l.to_string()),
```
Wumpf, line 1325: a bit of a stretch to use this as a filename 🤔

Wumpf (commented):
Given that this flag only applies to spirv-out I'm wondering we could find another way to specify this instead of putting it on every `ShaderModuleDescriptor`

That said, maybe this is something we could expose for more inputs - @teoxoy can you comment on that? If we can make more use of that it's surely a useful pattern, otherwise I'm a bit on the fence, but still pondering what might be better there. We should definitely expose this _somehow_

wgpu-core/src/device/resource.rs
wicast: I can't determ whether the origin source is really a wgsl or glsl, or we just use "shader.wgsl" for default?

wicast:
> Given that this flag only applies to spirv-out I'm wondering we could find another way to specify this instead of putting it on every `ShaderModuleDescriptor`
> 
> That said, maybe this is something we could expose for more inputs - @teoxoy can you comment on that? If we can make more use of that it's surely a useful pattern, otherwise I'm a bit on the fence, but still pondering what might be better there. We should definitely expose this _somehow_

Honestly I have the same thought on it while writing, but I didn't want to changing the `create_shader_module` api too much at the time.

Now I have two idea:

1. add another option struct for `create_shader_module`.
2. add another api on device like `set_shader_compile_option`, every call on `create_shader_module` will query this option.

i vote for the second plan, because this will leave `create_shader_module` unchanged and you don't need to set a debug flag on every shader creation.

cwfitzgerald:
This kind of switch shows up in multiple places and is likely okay being a instance level switch for all shaders from the instance. 

We have a couple things that we want to be able to adjust on the instance level, including turning on/off underlying-api validation, debug labels, and graphics debugger integrations. I'm going to make a instance flags in a PR shortly and this can use that.

cwfitzgerald:
Ping from triage

teoxoy:
Sorry for taking a while to respond on this but I think what @cwfitzgerald has in mind sounds good to me.

cwfitzgerald:
To clarify, I will implement the flags in general very soon, then you can merge on top of that work

jimblandy:
Firefox wants this. We want to be able to enable the backend's debugging layers in non-debug Firefox builds, for fuzzing and things like that.

cwfitzgerald:
https://github.com/gfx-rs/wgpu/issues/2316 is the main issue for the flags

cwfitzgerald:
#4230 has now landed, giving you a flag to latch onto!

wicast:
> #4230 has now landed, giving you a flag to latch onto!

sure! I will take some time to integrate

nical:
https://github.com/gfx-rs/wgpu/pull/4246 adds the plumbing to access the instance flags from the instance or the device.

deno_webgpu/shader.rs
```diff
@@ -43,6 +43,7 @@ pub fn op_webgpu_create_shader_module(
     let descriptor = wgpu_core::pipeline::ShaderModuleDescriptor {
         label: label.map(Cow::from),
         shader_bound_checks: wgpu_types::ShaderBoundChecks::default(),
+        debug: false,
```
crowlKats: could an env var be added for this? Something along the lines of `DENO_WEBGPU_DEBUG_SHADERS` (and documented in the readme).

wicast: I've done integration with instance flags and this field is removed. But I can't find how to control the flags after creating the instance. Or could I add a debug env to control shader compile?

cwfitzgerald: Yeah, an env var handled by a utility function in wgpu::util seems like the way to go.

cwfitzgerald: The expectation is that you make the decision once on instance creation - these things aren't harmful once on, and you know ahead of time if you might need debug information.

wicast: > Yeah, an env var handled by a utility function in wgpu::util seems like the way to go.

There is a `WGPU_DEBUG` env i found which can control shader compile. What can i do to deal with deno?

cwfitzgerald: DENO is the WGPU env but with DENO on the front

wicast: > DENO is the WGPU env but with DENO on the front

What I mean is deno test is using the `state` to get an instance
```rust
let instance = state.borrow::<super::Instance>();
```
How can I affect this by a DENO_WEBGPU_DEBUG_SHADERS env?

crowlKats: given that there are InstanceFlags now and it handles multiple usecases, I'll take care of this in the coming days myself, so this isn't something this PR needs to deal with

cwfitzgerald (approved):

PR #4033, opened 2023-08-11
DX12 set `Features::VERTEX_WRITABLE_STORAGE` based on the right FL

teoxoy:


teoxoy:
It's probably better to wait for https://github.com/gfx-rs/wgpu/pull/3987 and derive `PartialEq`.

wgpu-hal/src/dx12/adapter.rs
```diff
@@ -259,6 +258,10 @@ impl super::Adapter {
         // write the results there, and issue a bunch of copy commands.
         //| wgt::Features::PIPELINE_STATISTICS_QUERY
 
+        if max_feature_level >= d3d12::FeatureLevel::L11_1 {
```
cwfitzgerald: Instead of waiting, could we just cast both to u32 with a note on the d3d12 PR?

PR #4035, opened 2023-08-11
use WGSL for VertexFormat example types

### LINKED ISSUE #3990: VertexFormat documentation uses GLSL instead of WGSL types
The current documentation uses GLSL types like `float` and `ivec4` instead of WGSL types like `f32` or `vec4<i32>`. This is related to #2849. This should be an easy fix, but I'm not sure how to handle double precision types. It doesn't look like double precision is mentioned anywhere in the [WGSL spec](https://www.w3.org/TR/WGSL/) other than literals.

https://www.w3.org/TR/webgpu/#enumdef-gpuvertexformat

ScanMountGoat:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
closes #3990 

**Description**
Most users will write WGSL shaders instead of GLSL. This updates the example types to use WGSL syntax.

**Testing**
Comparing cargo doc output with https://gpuweb.github.io/gpuweb/#enumdef-gpuvertexformat

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -4334,9 +4334,9 @@ pub enum VertexFormat {
     Snorm16x2 = 14,
     /// Four signed shorts (i16). [-32767, 32767] converted to float [-1, 1] `vec4<f32>` in shaders.
     Snorm16x4 = 15,
-    /// Two half-precision floats (no Rust equiv). `vec2<f16>` in shaders.
+    /// Two half-precision floats (no Rust equiv). `vec2<f32>` in shaders.
     Float16x2 = 16,
-    /// Four half-precision floats (no Rust equiv). `vec4<f16>` in shaders.
+    /// Four half-precision floats (no Rust equiv). `vec4<f32>` in shaders.
     Float16x4 = 17,
     /// One single-precision float (f32). `f32` in shaders.
     Float32 = 18,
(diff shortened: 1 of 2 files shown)


wgpu-types/src/lib.rs
```diff
@@ -4302,73 +4302,73 @@ pub struct VertexAttribute {
...
+    /// Two signed shorts (i16). [-32767, 32767] converted to float [-1, 1] `vec2<f32>` in shaders.
     Snorm16x2 = 14,
-    /// Four signed shorts (i16). [-32767, 32767] converted to float [-1, 1] `vec4` in shaders.
+    /// Four signed shorts (i16). [-32767, 32767] converted to float [-1, 1] `vec4<f32>` in shaders.
     Snorm16x4 = 15,
-    /// Two half-precision floats (no Rust equiv). `vec2` in shaders.
+    /// Two half-precision floats (no Rust equiv). `vec2<f16>` in shaders.
     Float16x2 = 16,
-    /// Four half-precision floats (no Rust equiv). `vec4` in shaders.
+    /// Four half-precision floats (no Rust equiv). `vec4<f16>` in shaders.
```
teoxoy: The spec lists that `unorm`/`snorm`/`float` can all be accessed as `f16`/`f32` on the shader side.
Not sure what's best to document. All `f32`? or mention that both `f16`/`f32` work?

ref: https://gpuweb.github.io/gpuweb/#vertex-formats

ScanMountGoat: Is `f16` usable currently from WGSL? The two approaches I can think of are to match the spec and list both `f16` and `f32` for the appropriate types or match what's supported by naga. As a user, I would prefer to see a WGSL type in the docs that currently works, but I'm happy to make changes to this PR to match whatever people decide.

teoxoy: `f16` is not yet implemented in naga (see https://github.com/gfx-rs/naga/issues/1884).

I don't have a strong opinion either way.

cwfitzgerald: Maybe lets use `vec4<{float}>` or something similar.

ScanMountGoat: Do vertex formats like `Float64x3` also require the WGSL type to be`vec3<f32>` on the current version? One approach is to use `f32` for all the docs for all the float formats and make a separate issue to update the documentation once `f16` and `f64` are supported in shaders.

teoxoy: > Do vertex formats like `Float64x3` also require the WGSL type to be`vec3<f32>` on the current version?

The `VERTEX_ATTRIBUTE_64BIT` feature is apparently not currently advertised by any backend.

> One approach is to use `f32` for all the docs for all the float formats and make a separate issue to update the documentation once `f16` and `f64` are supported in shaders.

I think we can either do this or go with a generic solution like @cwfitzgerald was proposing. Thoughts?

cwfitzgerald: Either solution work for me

ScanMountGoat: I've added another commit to use f32 for format types represented as float in the shaders.

PR #4038, opened 2023-08-11
[vk] Don't request portability enumeration without the extension.

jimblandy:
VUID-VkInstanceCreateInfo-flags-06559 forbids us from passing `vk::InstanceCreateFlags::ENUMERATE_PORTABILITY_KHR` unless we actually found and requested the "VK_KHR_portability_enumeration" layer.

Fixes #4004.

**Checklist**

- [X] Run `cargo clippy`.
- [X] Run `cargo clippy --target wasm32-unknown-unknown` if applicable.
- [x] Add change to CHANGELOG.md. See simple instructions inside file.

**Testing**
This is covered by `wgpu-test::wgpu-tests encoder::drop_encoder`.

PR #4044, opened 2023-08-13
Better way to handle vk khr portability enumeration extension

hybcloud:
Better way to handle vk khr portability enumeration extension of hard-coded c_str.

**Checklist**

- [x] Run `cargo clippy`.

**Connections**
#4038
#3958
https://github.com/ash-rs/ash/blob/f5587619979b791c8d37110393c4a9b4c50e0a34/examples/src/lib.rs#L235

**Description**
The PR enhances the solution of #4038, which solves the bug introduced in #3958 by using ash's KHR definition (semantically, it's the funtion's name, not extension's name, but they share the same name, and ash uses it as extension's name in its own
example code 
(https://github.com/ash-rs/ash/blob/f5587619979b791c8d37110393c4a9b4c50e0a34/examples/src/lib.rs#L235)
(In ash's example code just mentioned, they requires another extension called KhrGetPhysicalDeviceProperties2. The extension is a KHR in vk1.0 but soon get into core in vk 1.1, and I don't have a mac to test if it matters, so I choose to keep it right the same as before)
)

**Testing**
Thanks to #4038,  validation layers on windows stop reporting error about VK_KHR_portability_subset. It's a simple enhancement of #4038, and wgpu example (hello-triangle) has been shown to work fine when backends flag is VULKAN.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,8 @@ By @Valaphee in [#3402](https://github.com/gfx-rs/wgpu/pull/3402)
 
 - Don't pass `vk::InstanceCreateFlags::ENUMERATE_PORTABILITY_KHR` unless the `VK_KHR_portability_enumeration` extension is available. By @jimblandy in[#4038](https://github.com/gfx-rs/wgpu/pull/4038).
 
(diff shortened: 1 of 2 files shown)


Wumpf (approved):
nice catch, thank you for contributing!

PR #4053, opened 2023-08-14
Extra limits

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

**Description**

In gecko we want to be able to impose limits on the devices we expose to web content that are potentially more conservative than what the device may advertise.

This PR adds allows optionally specifying extra limits at the instance level. If so, enumerated adapters will pretend to be at least as limited as the provided extra limits.

For example if the extra limits have `max_texture_size` set to `4096` and the device exposes adapters with max texture sizes of `8192`, then the user will see `4096` when looking in the limits of enumerated adapters, and more importantly, the `4096` limit will be enforced during validation.

To specify some limits and leave others untouched, the rule of thumb is to use the maximum value for limits that start with `max` and zero for limits that start with  `min`.

**Testing**

Yep.


nical:
Discussed with @teoxoy who pointed out that we'll need a similar API for bucketing but at the adapter level, so I'm going to change this to be an extra methods on adapters to apply further restrictions.

nical:
Nevermind, I'm going to do it all on the gekco side for now.

PR #4055, opened 2023-08-14
Add an overview of `RenderPass` and how render state works.

kpreid:
**Checklist**

- [X] Run `cargo clippy`.

**Description**
Add an overview of `RenderPass` and how render state works.

* Explain the general relationship of `set` and `draw` methods.
* Explain render state changes and that you can have lots of them.
* Specify default values of render state.

This will be useful to users who are not familiar with this style of graphics API (or any).


Wumpf (approved):
excellent, thank you!

PR #4058, opened 2023-08-15
Add validation in accordance with WebGPU `setViewport` valid usage fo…

James2022-rgb:
…r `x`, `y` and `this.[[attachment_size]]`.

`x` and `y` must not be negative, and the rect must be contained in the render target.

**Checklist**

- [x] Run `cargo clippy`.
   Done just in case, with --feature gles.

**Description**
For `wgpu::RenderPass::set_viewport`, it was possible to use values such that:
- `x` and/or `y` is negative
- the rect is not contained in the render target.

both of which are in violation of the valid usage defined for `setViewport` in the WebGPU spec:
> x ≥ 0
> y ≥ 0
> x + width ≤ this.[[[attachment_size]]].width
> y + height ≤ this.[[[attachment_size]]].height

This PR renames `RenderCommandError::InvalidViewportDimension` to `InvalidViewportRect` and adds render target extent to it, in the same spirit as `InvalidScissorRect` (which has similar valid usage regarding render target extent).

**Testing**
Tested locally.


Wumpf (approved):
nice and clean, thank you!

PR #4059, opened 2023-08-15
[vulkan] Don't bother to free command buffers before destroying the pool.

jimblandy:
Calling `vkDestroyCommandPool` automatically frees all command buffers allocated from that pool, so there is no need for `Device::destroy_command_encoder` to explicitly call `vkFreeCommandBuffers` on the `CommandEncoder`'s `free` and `discarded` lists.

**Checklist**

- [X] Run `cargo clippy`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,10 @@ Bottom level categories:
 
 - Omit texture store bound checks since they are no-ops if out of bounds on all APIs. By @teoxoy in [#3975](https://github.com/gfx-rs/wgpu/pull/3975)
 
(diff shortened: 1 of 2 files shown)


jimblandy:
Please check my understanding of Vulkan, but the spec does say:

> vkDestroyCommandPool
> ...
> When a pool is destroyed, all command buffers allocated from the pool are [freed](https://registry.khronos.org/vulkan/specs/1.2-extensions/html/vkspec.html#vkFreeCommandBuffers).

teoxoy (approved):
LGTM!

PR #4066, opened 2023-08-16
Add details to `InstanceError` and `CreateSurfaceError`.

kpreid:
**Checklist**

- [x] Update all backends
- [x] Add tests where feasible, such as WebGL and WebGPU canvas errors
- [X] Run `cargo clippy`.
- [x] Add change to CHANGELOG.md.

**Connections**
Followup to #3052.

**Description**
This will allow developers, particularly those in communication with users with hardware they don't own, to diagnose “why won't this work on my machine” better, without needing to obtain logs in addition to the error value.

Note that this is a semver-breaking change because `InstanceError` and `CreateSurfaceError` no longer implement `PartialEq`. This could be avoided by pointer comparison of the internal `Arc`, but I think it is more typical Rust to not implement `PartialEq` for error types unless there is a strong motivation to and their contents are tightly controlled.

**Testing**
Ran `cargo nextest run`; CI will check other platforms.


kpreid:
This PR is in principle ready to merge, but I am hoping to get around to writing tests for the `canvas.getContext()` errors before I mark it ready.

kpreid:
I've added a test for `create_surface_from_canvas()` under WebGL and WebGPU. However, a certain amount of persuasion was required for `tests/src/lib.rs` to cooperate with the custom canvas creation. Please review carefully (and CI will hopefully tell us if the cfgs are wrong).

kpreid:
That `clippy::redundant_clone` lint that failed in CI is completely bogus — the code wouldn't compile (and would be wrong) without it, since `SurfaceGuard` has a `Drop`. I've `allow`ed it.

kpreid:
Ping? This should be ready to go now.

Also, I have belatedly realized that I want to do exactly the same thing to `RequestDeviceError` — would you prefer that be folded into this PR or separate?

cwfitzgerald (approved):
This is beautiful stuff!

PR #4068, opened 2023-08-17
Faster set_bind_group for WebGPU backend

### LINKED ISSUE #4047: Using simpler set bindgroup function call on web improves performance
First, thanks for creating this great library.

We observed a strange performance issue when using wgpu's webgpu backend directly on the web.

The performance test is roughly drawing 10000 cubes using 10000 drawcall. Each one using a unique bindgroup. From the browser's devtools profiler, we see the setBindgroup is the biggest cost(23% on Chrome, and even more on Firefox). We haven't seen this on the native side.

After some investigation, we found that at least on Firefox, the browser will do expensive checks on third
parameter. This code is from firefox:

code

If we directly use the https://docs.rs/web-sys/latest/web_sys/struct.GpuRenderPassEncoder.html#method.set_bind_group instead of using https://docs.rs/web-sys/latest/web_sys/struct.GpuRenderPassEncoder.html#method.set_bind_group_with_u32_array_and_f64_and_dynamic_offsets_data_length , the setbindgroup cost on chrome reduced to 15% and the issue totally fixed on firefox.

I suggest using simpler calls on the web if the dynamic offset feature is not actively used. The setBindgroup is the hottest path in rendering. Even if the browser fixed this performance issue in the future, pass fewer parameters across the wasm js boundary is good for performance.

We tested on chrome 115, firefox nightly 2023-08-10, using wgpu 0.17

JMS55:
Closes https://github.com/gfx-rs/wgpu/issues/4047

(I didn't do any testing of this)

### DIFF
--- a/wgpu/src/backend/web.rs
+++ b/wgpu/src/backend/web.rs
@@ -2604,15 +2604,19 @@ impl crate::context::Context for Context {
         bind_group_data: &Self::BindGroupData,
         offsets: &[wgt::DynamicOffset],
     ) {
-        pass_data
-            .0
-            .set_bind_group_with_u32_array_and_f64_and_dynamic_offsets_data_length(
-                index,
-                &bind_group_data.0,
-                offsets,
-                0f64,
-                offsets.len() as u32,
-            );
+        if offsets.is_empty() {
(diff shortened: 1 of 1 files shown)


cwfitzgerald (approved):
LGTM

PR #4071, opened 2023-08-17
Add a way to further restrict limits at the adaper level in wgpu-core.

nical:
**Connections**

Alternative to #4053 

**Description**

This version adds a function allowing Firefox to restrict the limits at the adapter level (instead of the instance) in wgpu-core. The upside is that it does not have to be exposed to the wgpu API where it's probably not super useful, and it makes for a smaller diff.

**Testing**

The downside is that it's tedious to test wgpu-core APIs so I didn't bother.


nical:
Nevermind, I'm going to do it all on the gekco side for now.

PR #4076, opened 2023-08-18
Make power_preference_from_env() respect WGPU_POWER_PREF=none

fornwall:
Also update wgpu/README.md now that default() is None.

**Checklist**

- [x] Run `cargo clippy`.

**Description**
Respecting `WGPU_POWER_PREF=none` makes no difference to wgpu testing and examples, as the fallback is already `default()` (which is `PowerPreference::None`), but makes a difference to other projects which for instance uses the logic "check for WGPU_POWER_PREF in the env, otherwise use high power".

Also updates `wgpu/README.md` to reflect that the default is now `None`.

**Testing**
Manually testing.


teoxoy (approved):
LGTM, thanks!

PR #4077, opened 2023-08-18
Parametrize the size of the non-sampler descriptor heap in dx12

nical:
**Checklist**

- [ ] Run `cargo clippy`.

**Connections**

https://bugzilla.mozilla.org/show_bug.cgi?id=1848937

**Description**

This is the first / short term step to mitigate OOMs after creating a number of devices. The browser can have to create a lot of devices so it can't afford to provision for a million descriptors per device. Firefox will set the limit to something lower. This shouldn't change the behavior of wgpu users if they keep the default limit.

**Testing**

In Firefox CTS runs this should remove or reduce a fair amount of intermittent OOMs.

### DIFF
--- a/wgpu-info/src/human.rs
+++ b/wgpu-info/src/human.rs
@@ -124,6 +124,7 @@ fn print_adapter(output: &mut impl io::Write, report: &AdapterReport, idx: usize
         max_compute_workgroup_size_y,
         max_compute_workgroup_size_z,
         max_compute_workgroups_per_dimension,
+        max_non_sampler_bindings,
     } = limits;
     writeln!(output, "\t\t                        Max Texture Dimension 1d: {max_texture_dimension_1d}")?;
     writeln!(output, "\t\t                        Max Texture Dimension 2d: {max_texture_dimension_2d}")?;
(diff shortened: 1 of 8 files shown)


nical:
I can see that it's a bit awkward to put this workaround in `Limits`. I'm fine with getting the same results via some other API if we feel strongly about it.

wgpu-info/src/human.rs
```diff
@@ -155,6 +156,10 @@ fn print_adapter(output: &mut impl io::Write, report: &AdapterReport, idx: usize
     writeln!(output, "\t\t                    Max Compute Workgroup Size Z: {max_compute_workgroup_size_z}")?;
     writeln!(output, "\t\t            Max Compute Workgroups Per Dimension: {max_compute_workgroups_per_dimension}")?;
 
+    // This one reflects more of a wgpu implementation limitations than a hardware limit
+    // so don't hsow it here.
```
ErichDonGubler: typo:

```suggestion
    // so don't show it here.
```

teoxoy (approved):
LGTM

I was wondering if we need extra validation for this limit but it will already return an OOM error if exceeded.

https://github.com/gfx-rs/wgpu/blob/c7e262c5d6fc52e672e5f796d73215e34f683189/wgpu-hal/src/dx12/descriptor.rs#L94

PR #4096, opened 2023-08-29
Skip `test_multithreaded_compute` on MoltenVK.

jimblandy:
Skip `tests::test_multithreaded_compute` in `examples/hello-compute/src/tests.rs` on MoltenVK for timeouts.

All commits should pass CI, but may be fine-grained enough that squashing makes sense. Up to you.

Various cleanups to expected failure code:

- Give test image snapshots distinctive names.

  Include the adapter name and driver name in the filenames generated to store the `-actual.png` and `-difference.png` files holding screenshots of test program output.

  Previously, the filename only incorporated the backend, but since we test the Vulkan backend on both Mac and Windows, one was overwriting the other in the artifacts.

- Upload GitHub workflow artifacts even when the test fails.

  When a test in `.github/workflows/ci.yml`'s `gpu-test` fails, ensure that it uploads the artifacts anyway. These are screenshots and comparison images, so they're valuable in debugging the failure.

- Allow expected test failures to match the driver name.

  Add a new field `FailureCase::driver`, which must match `AdapterInfo::driver` if given.

  Use this in `TestParameters::molten_vk_failure`.

- Implement `Default` for `FailureCase`, and use it where appropriate.

  This makes it easier to add new fields to `FailureCase` in the future, without having to fix every use.

- Make `TestParameters::specific_failure` take a `FailureCase`.

  Make `TestParameters::specific_failure` take a `FailureCase` struct as its sole argument, thus requiring callers to label each parameter with its meaning. This is clearer than using four positional parameters.

- Introduce TestParameters::backend_adapter_failure.

  Introduce a new `TestParameters` builder method, `backend_adapter_failure`, for marking an expected failure on a specific adapter and backend. Use where appropriate.

- Introduce TestParameters::adapter_failure_skip.

  Introduce a new `TestParameters` builder method for skipping a test on a specific adapter. Use where appropriate.

- Update doc comments for some `TestParameters` methods.

  Replace references to `specific_failures` with outdated parameter lists with plain English explaining what each function actually does.

- Use TestParameters::backend_failure method where appropriate.

**Checklist**

- [X] Run `cargo clippy`.

**Connections**

**Description**

**Testing**

### DIFF
--- a/tests/src/lib.rs
+++ b/tests/src/lib.rs
@@ -115,7 +115,7 @@ impl TestParameters {
         self
     }
 
-    /// Mark the test as always failing, equivalent to specific_failure(None, None, None)
+    /// Mark the test as always failing, but not to be skipped.
     pub fn failure(mut self) -> Self {
         self.failures.push(FailureCase {
             backends: None,
@@ -126,7 +126,7 @@ impl TestParameters {
         self
     }
 
-    /// Mark the test as always failing and needing to be skipped, equivalent to specific_failure(None, None, None)
+    /// Mark the test as always failing, and needing to be skipped.
     pub fn skip(mut self) -> Self {
         self.failures.push(FailureCase {
             backends: None,
@@ -137,7 +137,7 @@ impl TestParameters {
         self
     }
 
-    /// Mark the test as always failing on a specific backend, equivalent to specific_failure(backend, None, None)
+    /// Mark the test as always failing on `backends`, but not to be skipped.
     pub fn backend_failure(mut self, backends: wgpu::Backends) -> Self {
         self.failures.push(FailureCase {
             backends: Some(backends),
@@ -148,8 +148,10 @@ impl TestParameters {
         self
     }
 
-    /// Mark the test as always failing on WebGL. Because limited ability of wasm to recover from errors, we need to wholesale
-    /// skip the test if it's not supported.
+    /// Mark the test as always failing on WebGL, and needing to be skipped.
+    ///
+    /// Because limited ability of wasm to recover from errors, we
+    /// need to wholesale skip the test if it's not supported.
     pub fn webgl2_failure(mut self) -> Self {
         let _ = &mut self;
         #[cfg(target_arch = "wasm32")]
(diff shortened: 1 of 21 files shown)


tests/src/lib.rs
```diff
@@ -53,11 +53,80 @@ fn lowest_downlevel_properties() -> DownlevelCapabilities {
...
+    /// this is `None`, the driver name isn't considered.
+    ///
+    /// [`AdapterInfo::driver`]: wgt::AdapterInfo::driver
+    pub driver: Option<S>,
+
+    /// If `true`, skip the test altogether.
+    ///
+    /// If `false`, the test is run, and an unexpected pass is
+    /// reported if it doesn't fail.
+    pub skip: bool,
```
cwfitzgerald: It would improve readability signifigantly to have this be an enum between "ExpectFailure" and "skip" or something like that, I've already gotten tripped up reading the bools in this MR.

`.molten_vk_failure(false)` sounds like it shouldn't fail on moltenvk, but molten_vk_failure(FailureMode::ExpectFailure) or whatever is much better

tests/src/lib.rs
```diff
@@ -53,11 +53,80 @@ fn lowest_downlevel_properties() -> DownlevelCapabilities {
...
+/// # ;
+/// ```
+///
+/// This applies to all cards with `"RTX'` in their name on either
+/// Direct3D backend, no matter the vendor ID or driver name.
+///
+/// The default value of `FailureCase` applies to any test case (that
+/// is, there are no criteria to constrain the match), and the test is
+/// not skipped.
+pub struct FailureCase<S> {
```
cwfitzgerald: We should be able to mark this as `#[non_exhaustive]` to really enforce the "you must use default"

cwfitzgerald (approved):
This is great stuff, have some nits, but good after

jimblandy:
Requesting a new review, since some of the new commits are non-trivial. None of the older commits are affected.

cwfitzgerald (approved):
Very nice!

PR #4110, opened 2023-09-02
Enable vulkan presentation on Intel Mesa >= v21.2

### LINKED ISSUE #4101: Forced GPU selection prevents wgpu running on hybrid nvidia laptops
The block at https://github.com/gfx-rs/wgpu/blob/625afc3b4299c92c79e0b3881e22a0bb780678ed/wgpu-hal/src/vulkan/instance.rs#L771 is an issue. On hybrid nvidia laptops it is preventing running wgpu based applications completely.

So what is happening for every Nvidia hybrid user is:
1. Intel GPU is blocked so it's impossible to select it
2. Nvidia will never work either due to https://github.com/NVIDIA/egl-wayland/issues/72
3. wgpu based apps just will not work (such as https://github.com/lapce/floem/issues/100 and all wgpu examples)

Given the above, this commit should be reverted, and be done before the next release of wgpu.

Related: https://github.com/gfx-rs/wgpu/issues/1215 (nvidia provides ways to force nvidia gpu selection)
Related: https://github.com/gfx-rs/wgpu/pull/1898 (the commit)

flukejones:
Due to an issue with Mesa versions less than 21.2 presentation on Vulkan was forced to Nvidia only. This in itself brought new issues around the Nvidia driver specfic format modifers.

As of Mesa 21.2 the Intel vulkan issue is fixed. This commit enables presentation on versions 21.2 and above for Intel.

References:
- https://github.com/NVIDIA/egl-wayland/issues/72

Closes [#4101](https://github.com/gfx-rs/wgpu/issues/4101)

### DIFF
--- a/wgpu-hal/src/vulkan/instance.rs
+++ b/wgpu-hal/src/vulkan/instance.rs
@@ -786,13 +786,22 @@ impl crate::Instance<super::Api> for super::Instance {
                 if exposed.info.device_type == wgt::DeviceType::IntegratedGpu
                     && exposed.info.vendor == db::intel::VENDOR
                 {
-                    // See https://gitlab.freedesktop.org/mesa/mesa/-/issues/4688
-                    log::warn!(
-                        "Disabling presentation on '{}' (id {:?}) because of NV Optimus (on Linux)",
-                        exposed.info.name,
-                        exposed.adapter.raw
-                    );
-                    exposed.adapter.private_caps.can_present = false;
+                    // Check if mesa driver and version less than 21.2
+                    if let Some(version) = exposed.info.driver_info.split_once("Mesa ").map(|s| {
+                        s.1.rsplit_once('.')
+                            .map(|v| v.0.parse::<f32>().unwrap_or_default())
+                            .unwrap_or_default()
+                    }) {
+                        if version < 21.2 {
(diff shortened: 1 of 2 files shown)


wgpu-hal/src/vulkan/instance.rs
```diff
@@ -786,13 +786,22 @@ impl crate::Instance<super::Api> for super::Instance {
...
-                        exposed.adapter.raw
-                    );
-                    exposed.adapter.private_caps.can_present = false;
+                    // Check if mesa driver and version less than 21.2
+                    if let Some(version) = exposed.info.driver_info.split_once("Mesa ").map(|s| {
+                        s.1.rsplit_once('.')
+                            .map(|v| v.0.parse::<f32>().unwrap_or_default())
+                            .unwrap_or_default()
+                    }) {
+                        if version < 21.2 {
```
Wumpf, line 795: imho the logic should be flipped: Once we detected it's Mesa + Intel, we should required >= 21.2 to _enable_ it, this way this becomes more robust against parsing error (e.g. if any of the `unwrap_or_default` yields the default, we should not enable the adapter)

Wumpf (approved):
was thinking at first the logic should be split so that we opt in the adapater, but that might too easily break future mesa version strings I reckon.
unfortunately I don't have a way to test this, trusting the version parsing works as intended

PR #4114, opened 2023-09-04
wgpu_core: Add logging to Instance::new.

jimblandy:
For each backend `blah`, log::debug whether we were able to populate `Instance::blah`.

**Checklist**

- [X] Run `cargo clippy`.
- (unnecessary) Add change to CHANGELOG.md. See simple instructions inside file.

### DIFF
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -84,8 +84,22 @@ impl Instance {
                     dx12_shader_compiler: instance_desc.dx12_shader_compiler.clone(),
                     gles_minor_version: instance_desc.gles_minor_version,
                 };
-                unsafe { hal::Instance::init(&hal_desc).ok() }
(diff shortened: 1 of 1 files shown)


wgpu-core/src/instance.rs
```diff
@@ -84,8 +84,22 @@ impl Instance {
...
+                        log::debug!(
+                            "Instance::new: failed to create {:?} backend: {:?}",
+                            A::VARIANT,
+                            err
+                        );
+                        None
+                    }
+                }
             } else {
+                log::debug!("Instance::new: backend {:?} not requested", A::VARIANT);
```
Wumpf: that's more of a `log::trace` I'd say. On the fence with the others since they _may_ actually contain error information so don't feel like banning that into trace

cwfitzgerald (approved):

cwfitzgerald:
I think the debug levels are good now

PR #4143, opened 2023-09-15
update deno and spec

crowlKats:
Updated #2170


mmastrac (approved):
LGTM. I would recommend avoiding the #[serde] return value for WebGpuResult by reworking the APIs to make use of Result instead which will perform much better. For now, looks great!

rust-toolchain
```diff
@@ -5,6 +5,6 @@
 # to the user in the error, instead of "error: invalid channel name '[toolchain]'".
 
 [toolchain]
-channel = "1.65"
+channel = "1.70"
```
cwfitzgerald: Oof this is tough - we did this so that people would be using the proper MSRV for wgpu (and not get errant clippy warnings), but we still can't actually bump our MSRV appreciably.

crowlKats: Yes; it annoyed me s well, however some of the deno deps use features from 1.70 :/

crowlKats: does it maybe work to have toolchains for the 2 deno folders? not sure if thats possible

cwfitzgerald: I don't think so. Let's bump this for now, but keep MSRV testing in CI. There's probably a better solution out there.

cwfitzgerald:
So thinking about it, dealing with the MSRV bump should be done the following way:

- Up the rust toolchain file to what deno needs (already done)
- Add a CI job for windows/mac/linux which only builds wgpu-core (and dependencies) which is built with 1.65. This will ensure we don't break firefox. It also doesn't need to run clippy, just `cargo check`.
- Document mismatched MSRV.

cwfitzgerald (changes requested):
Marking as request changes for bookkeeping purposes.

.github/workflows/ci.yml
```diff
@@ -351,8 +352,8 @@ jobs:
 
       - name: Install MSRV toolchain
         run: |
-          rustup toolchain install ${{ env.MSRV }} --no-self-update --profile=minimal --component clippy
-          rustup default ${{ env.MSRV }}
+          rustup toolchain install ${{ env.DENO_MSRV }} --no-self-update --profile=minimal --component clippy
```
cwfitzgerald: I don't think this is going to be sufficient.

The PR for testing frameworks also needs an msrv bump, I'm going to raise an PR which bumps MSRV for everything except core/hal/types.

cwfitzgerald (changes requested):
Everything else looks fine, just need to address this first.

.github/workflows/ci.yml
cwfitzgerald: #4204 is landed, will rebase on top of this.

cwfitzgerald:
Seems to have broken the player.

crowlKats:
@cwfitzgerald I fixed it; problem was the RON for one of the tests because I added rename_all=camelCase, though now for one struct, the RON looks off in files since its the only thing with camelCase. Is this a problem? Else, I can just rename the serialize and leave deserialize to normal if preffered

cwfitzgerald:
That's fine, those Ron aren't a big deal.

cwfitzgerald (approved):
Yeet

PR #4145, opened 2023-09-15
Add details to `RequestDeviceError`.

kpreid:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Basically just like #4066 — giving the same treatment to `RequestDeviceError` as `InstanceError` and `CreateSurfaceError` got.

**Description**
This will allow developers, particularly those in communication with users with hardware they don't own, to diagnose “why won't this work on my machine” better, without needing to obtain logs in addition to the error value.

Note that this is a semver-breaking change because `RequestDeviceError` no longer implements `PartialEq`, and cannot be constructed outside the crate. (Just like #4066.)

**Testing**
Added a test case and ran it in CI.


cwfitzgerald (approved):
Very nice!

PR #4147, opened 2023-09-16
Make `StoreOp` an enum instead of a bool

### LINKED ISSUE #4146: Depth target gets cleared before compute pass when previously used as attachement without storing values
**Description**
Occurs with the following sequence

* render pass that clears targets and stores depth values
* render pass that does not clear targets and does not store depth values
* (wgpu clears depth target internally)
* depth target gets used as texture_depth_2d in compute pass (and has now all 0 values) -> I would have expected to still have the depth values from the first render pass available here

**Repro steps**
see the attached example

I would expect that the depth target does not get cleared when running the compute pass.

**Extra materials**
The attached zip contains a minimal example of the issue.

**Platform**
Windows 10
wgpu 0.17 and master on Vulkan backend
RX 570 driver version 21.50.21.11

[wgpu-depth-clear-bug.zip](https://github.com/gfx-rs/wgpu/files/12640986/wgpu-depth-clear-bug.zip)

Wumpf:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
* fixes #4146
   * (description is actually not indicative of the actual error but of the common misconception this avoids)

**Description**

`wgpu::Operations::store` used to be an underdocumented boolean value,
causing missunderstandings of the effect of setting it to `false`.

The API now resembles WebGPU which distinguishes `store` and `discard`,
see [WebGPU spec on GPUStoreOp](https://gpuweb.github.io/gpuweb/#enumdef-gpustoreop).

```diff
// ...
depth_ops: Some(wgpu::Operations {
    load: wgpu::LoadOp::Clear(1.0),
-   store: false,
+   store: wgpu::StoreOp::Discard,
}),
// ...
```

Opted to put this text into the changelog under `Major Changes` since it breaks like every single use ;-)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,7 +68,7 @@ By @Valaphee in [#3402](https://github.com/gfx-rs/wgpu/pull/3402)
 #### Render pass store operation is now an enum
 
 `wgpu::Operations::store` used to be an underdocumented boolean value,
-causing missunderstandings of the effect of setting it to `false`.
+causing misunderstandings of the effect of setting it to `false`.
 
 The API now resembles WebGPU which distinguishes `store` and `discard`,
 see [WebGPU spec on GPUStoreOp](https://gpuweb.github.io/gpuweb/#enumdef-gpustoreop).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,7 +70,7 @@ By @Valaphee in [#3402](https://github.com/gfx-rs/wgpu/pull/3402)
 `wgpu::Operations::store` used to be an underdocumented boolean value,
 causing misunderstandings of the effect of setting it to `false`.
 
-The API now resembles WebGPU which distinguishes `store` and `discard`,
+The API now more closely resembles WebGPU which distinguishes between `store` and `discard`,
 see [WebGPU spec on GPUStoreOp](https://gpuweb.github.io/gpuweb/#enumdef-gpustoreop).
 
 ```diff
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,28 @@ let render_pass = encoder.begin_render_pass(&wgpu::RenderPassDescriptor {
 
 By @Valaphee in [#3402](https://github.com/gfx-rs/wgpu/pull/3402)
 
+
+#### Render pass store operation is now an enum
+
+`wgpu::Operations::store` used to be an underdocumented boolean value,
+causing missunderstandings of the effect of setting it to `false`.
+
+The API now resembles WebGPU which distinguishes `store` and `discard`,
+see [WebGPU spec on GPUStoreOp](https://gpuweb.github.io/gpuweb/#enumdef-gpustoreop).
+
+```diff
+// ...
+depth_ops: Some(wgpu::Operations {
+    load: wgpu::LoadOp::Clear(1.0),
+-   store: false,
++   store: wgpu::StoreOp::Discard,
+}),
+// ...
+```
+
+By @wumpf in [#4147](https://github.com/gfx-rs/wgpu/pull/4147)
+
+
 ### Added/New Features
 
 - Add `gles_minor_version` field to `wgpu::InstanceDescriptor`. By @PJB3005 in [#3998](https://github.com/gfx-rs/wgpu/pull/3998)
--- a/wgpu/src/backend/direct.rs
+++ b/wgpu/src/backend/direct.rs
@@ -4,7 +4,7 @@ use crate::{
     BufferDescriptor, CommandEncoderDescriptor, ComputePassDescriptor, ComputePipelineDescriptor,
     DownlevelCapabilities, Features, Label, Limits, LoadOp, MapMode, Operations,
     PipelineLayoutDescriptor, RenderBundleEncoderDescriptor, RenderPipelineDescriptor,
-    SamplerDescriptor, ShaderModuleDescriptor, ShaderModuleDescriptorSpirV, ShaderSource,
+    SamplerDescriptor, ShaderModuleDescriptor, ShaderModuleDescriptorSpirV, ShaderSource, StoreOp,
     SurfaceStatus, TextureDescriptor, TextureViewDescriptor, UncapturedErrorHandler,
 };
 
@@ -392,6 +392,13 @@ fn map_texture_tagged_copy_view(
     }
 }
 
+fn map_store_op(op: StoreOp) -> wgc::command::StoreOp {
+    match op {
+        StoreOp::Store => wgc::command::StoreOp::Store,
+        StoreOp::Discard => wgc::command::StoreOp::Discard,
+    }
+}
+
 fn map_pass_channel<V: Copy + Default>(
     ops: Option<&Operations<V>>,
 ) -> wgc::command::PassChannel<V> {
@@ -401,11 +408,7 @@ fn map_pass_channel<V: Copy + Default>(
             store,
         }) => wgc::command::PassChannel {
             load_op: wgc::command::LoadOp::Clear,
-            store_op: if store {
-                wgc::command::StoreOp::Store
-            } else {
-                wgc::command::StoreOp::Discard
-            },
+            store_op: map_store_op(store),
             clear_value,
             read_only: false,
         },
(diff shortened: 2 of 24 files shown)


CHANGELOG.md
```diff
@@ -64,6 +64,28 @@ let render_pass = encoder.begin_render_pass(&wgpu::RenderPassDescriptor {
 
 By @Valaphee in [#3402](https://github.com/gfx-rs/wgpu/pull/3402)
 
+
+#### Render pass store operation is now an enum
+
+`wgpu::Operations::store` used to be an underdocumented boolean value,
+causing missunderstandings of the effect of setting it to `false`.
```
cwfitzgerald:
```suggestion
causing misunderstandings of the effect of setting it to `false`.
```

CHANGELOG.md
```diff
@@ -64,6 +64,28 @@ let render_pass = encoder.begin_render_pass(&wgpu::RenderPassDescriptor {
...
+
+The API now resembles WebGPU which distinguishes `store` and `discard`,
+see [WebGPU spec on GPUStoreOp](https://gpuweb.github.io/gpuweb/#enumdef-gpustoreop).
```
cwfitzgerald:
```suggestion
The API now more closely resembles WebGPU which distinguishes between `store` and `discard`,
see [WebGPU spec on GPUStoreOp](https://gpuweb.github.io/gpuweb/#enumdef-gpustoreop).
```

wgpu/src/lib.rs
```diff
@@ -1006,16 +1006,22 @@ static_assertions::assert_impl_all!(BufferBinding: Send, Sync);
...
 pub enum LoadOp<V> {
-    /// Clear with a specified value.
+    /// Loads the specified value for this attachment into the render pass.
+    ///
+    /// On some GPU hardware (primarily mobile), "clear" is significantly cheaper
+    /// because it avoids loading data from main memory into tile-local memory.
+    /// On other GPU hardware, there isn’t a significant difference.
+    /// As a result, it is recommended to use "clear" rather than "load" in cases
+    /// where the initial value doesn’t matter
+    /// (e.g. the render target will be cleared using a skybox).
```
cwfitzgerald:
```suggestion
    /// Loads the specified value for this attachment into the render pass.
    ///
    /// On some GPU hardware (primarily mobile), "clear" is significantly cheaper
    /// because it avoids loading data from main memory into tile-local memory.
    ///
    /// On other GPU hardware, there isn’t a significant difference.
    ///
    /// As a result, it is recommended to use "clear" rather than "load" in cases
    /// where the initial value doesn’t matter
    /// (e.g. the render target will be cleared using a skybox).
```

Just spacing to make paragraph breaks

wgpu/src/lib.rs
```diff
@@ -1025,6 +1031,25 @@ impl<V: Default> Default for LoadOp<V> {
...
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, Default)]
+#[cfg_attr(feature = "trace", derive(serde::Serialize))]
+#[cfg_attr(feature = "replay", derive(serde::Deserialize))]
+pub enum StoreOp {
+    /// Stores the resulting value of the render pass for this attachment.
+    #[default]
+    Store,
+    /// Discards the resulting value of the render pass for this attachment.
+    ///
+    /// The attachment/affected texture aspect(s) will be treated as uninitialized afterwards.
```
cwfitzgerald:
```suggestion
    /// The attachment's affected texture aspect(s) will be treated as uninitialized afterwards.
```

This.. might be clearer? Not a lot of people understand texture aspects, so I'm hesitant to write about it without something explaining what it is. Maybe a link back to the texture aspect type?

wgpu/src/lib.rs
```diff
@@ -1025,6 +1031,25 @@ impl<V: Default> Default for LoadOp<V> {
...
+    ///
+    /// This can be significantly faster on tile-based render hardware.
+    /// Prefer this if the attachment is not read by subsequent passes.
```
cwfitzgerald:
```suggestion
    /// This can be significantly faster on tile-based render hardware.
    ///
    /// Prefer this if the attachment is not read by subsequent passes.
```

wgpu/src/lib.rs
```diff
@@ -1036,14 +1061,15 @@ pub struct Operations<V> {
     /// How data should be read through this attachment.
     pub load: LoadOp<V>,
     /// Whether data will be written to through this attachment.
```
cwfitzgerald, line 1068: I think we also need to change this comment/the other one to mention the fact that the resolve texture is _always_ stored to, no matter what this is set to.

wgpu/src/backend/direct.rs
```diff
@@ -392,6 +392,13 @@ fn map_texture_tagged_copy_view(
     }
 }
 
+fn map_store_op(op: StoreOp) -> wgc::command::StoreOp {
+    match op {
+        StoreOp::Store => wgc::command::StoreOp::Store,
+        StoreOp::Discard => wgc::command::StoreOp::Discard,
+    }
+}
```
cwfitzgerald, line 400: We totally should lift this into wgt and use the same type everywhere

cwfitzgerald (changes requested):
Comments, like the api change!

wgpu/src/lib.rs
Wumpf: we could also drop it from here, it's a super strange implementation detail about how different aspects of a texture can be in an uninitialized state (I hate it ;))

wgpu/src/backend/direct.rs
Wumpf: I was about to, but then figured since the LoadOp can't go into wgt (it has the clear value with it for improved Rust ergonomics) I didn't want to put them into different places so it's easier to find stuff 🤔

Wumpf:
Thanks for all the comment improvements and keeping the bar for documentation high!

wgpu/src/lib.rs
Wumpf: good idea

wgpu/src/backend/direct.rs
cwfitzgerald: Fair

wgpu/src/lib.rs
cwfitzgerald: This doesnt' seem to have been addressed.

wgpu/src/lib.rs
cwfitzgerald: It's not really an implementation detail, as it results in a visible clear happening - maybe we should document that behavior.

cwfitzgerald (changes requested):
Did you miss a push, a couple resolved things are still unaddressed

Wumpf:
> Did you miss a push

eh, yeah I managed to push to the wrong remote 🤦

wgpu/src/lib.rs
Wumpf: Hm maybe just best to talk about depth/stencil aspects inline here, shouldn't be too hard, let me try.

teoxoy (approved):
LGTM!

wgpu/src/lib.rs
```diff
@@ -1036,14 +1066,18 @@ pub struct Operations<V> {
...
-    pub store: bool,
+    ///
+    /// Note that resolve textures (if specified) are always written to,
+    /// regardless of this setting.
+    pub store: StoreOp,
 }
 
 impl<V: Default> Default for Operations<V> {
+    #[inline]
```
cwfitzgerald, line 1076: Why was this added?

cwfitzgerald (changes requested):

wgpu/src/lib.rs
Wumpf: counter question, why was it missing - I added this in passing since trivial default like this should imho be inlinable and Rust's compilation model doesn't allow for inlining accros crates unless either this attribute is present or link time optimizations are at play.
Unrelated to this PR ofc, but figured can't hurt. You object?

cwfitzgerald: I'm honestly not sure of the full costs/benefits of this - I know it can benefit when you have LTO off (but also why do you have LTO off if you care about performance) but I think there is costs to just slapping it on all over the place, even for trivial stuff.

cwfitzgerald: Not a discussion for this PR - let's merge.

cwfitzgerald (approved):

PR #4154, opened 2023-09-19
[metal] set the image stride to zero when updating a single layer of a 3D texture

teoxoy:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Follow up to https://github.com/gfx-rs/wgpu/pull/3063.

**Description**
See https://github.com/gfx-rs/wgpu/pull/3063#issuecomment-1662472629.

**Testing**
Test has been modified.

### DIFF
--- a/tests/tests/write_texture.rs
+++ b/tests/tests/write_texture.rs
@@ -51,9 +51,16 @@ fn write_texture_subset_2d() {
 
         ctx.queue.submit(None);
 
-        let read_buffer = ctx.device.create_buffer(&wgpu::BufferDescriptor {
+        let read_buffer_0 = ctx.device.create_buffer(&wgpu::BufferDescriptor {
             label: None,
-            size: (size * size) as u64,
+            size: (size * 2) as u64,
+            usage: wgpu::BufferUsages::MAP_READ | wgpu::BufferUsages::COPY_DST,
+            mapped_at_creation: false,
+        });
+
+        let read_buffer_1 = ctx.device.create_buffer(&wgpu::BufferDescriptor {
+            label: None,
+            size: (size * (size - 2)) as u64,
             usage: wgpu::BufferUsages::MAP_READ | wgpu::BufferUsages::COPY_DST,
             mapped_at_creation: false,
         });
@@ -70,7 +77,7 @@ fn write_texture_subset_2d() {
                 aspect: wgpu::TextureAspect::All,
             },
             wgpu::ImageCopyBuffer {
-                buffer: &read_buffer,
+                buffer: &read_buffer_0,
                 layout: wgpu::ImageDataLayout {
                     offset: 0,
                     bytes_per_row: Some(size),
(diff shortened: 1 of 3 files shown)


teoxoy:
It seems that Metal's validation errors don't surface in CI.
I removed the fix added in the previous PR and the test still succeeded (a68dc1077700ca18f1e4be45f7fd3a8720225d63).

teoxoy:
Ah, I think I see why. Is there a way to run the tests under xcode (in CI)? I think we should do that.

cc @nical @cwfitzgerald

cwfitzgerald:
We need to set the env var METAL_DEVICE_WRAPPER_TYPE=1 - we should do it in the test harness before wgpu is initialized so they get enabled locally too.

tests/tests/write_texture.rs
```diff
@@ -51,9 +51,16 @@ fn write_texture_subset_2d() {
 
         ctx.queue.submit(None);
 
-        let read_buffer = ctx.device.create_buffer(&wgpu::BufferDescriptor {
+        let read_buffer_0 = ctx.device.create_buffer(&wgpu::BufferDescriptor {
```
Wumpf, line 54: would imho be nice to parameterize this test to have it  both do a full update and two piecewise.
The second one can then point to this PR, pointing out that it protects against regressing this

teoxoy:
> We need to set the env var METAL_DEVICE_WRAPPER_TYPE=1 - we should do it in the test harness before wgpu is initialized so they get enabled locally too.

Wouldn't that require https://github.com/gfx-rs/wgpu/pull/3873? I can add it in CI for now though.

tests/tests/write_texture.rs
teoxoy: I can probably trigger the validation error with a simpler test case.
Will try to do that instead as opposed to modifying this one.

jimblandy (approved):
I like the idea of separating out the test case. The code changes look good.

teoxoy:
Besides some other validation errors that we get when I enabled metal's validation, I can't seem to get metal's validation to trigger for this specific case.

Maybe newer versions don't care but older do?
Should we just land the code change to make things consistent? (and revert the test & CI changes for now?)

cwfitzgerald:
> Wouldn't that require https://github.com/gfx-rs/wgpu/pull/3873? I can add it in CI for now though
 
@teoxoy No, you can just add the env var around here https://github.com/gfx-rs/wgpu/blob/trunk/tests/src/lib.rs#L328

As for what to do - split this however you feel is best, as long as at the end we end up:
- With the change landed
- With the env var set
- Test case written

We're where we need to be :)

teoxoy:
@cwfitzgerald question about the env variable:
I can't find any reference to `METAL_DEVICE_WRAPPER_TYPE` (even though it does seem to work).
I did however find [`MTL_DEBUG_LAYER`](https://developer.apple.com/documentation/xcode/validating-your-apps-metal-api-usage) & [`MTL_SHADER_VALIDATION`](https://developer.apple.com/documentation/xcode/validating-your-apps-metal-shader-usage).
I feel like we should use those instead. Thoughts?

cwfitzgerald:
I have no idea why we used this, I think this might be what xcode uses? I'm not particularly strongly either way, as long as it does what we want.

cwfitzgerald:
#3873 now includes both `MTL_DEBUG_LAYER` and `MTL_SHADER_VALIDATION`

cwfitzgerald:
@teoxoy what's the status on this?

teoxoy:
From my testing I did at the time, I couldn't produce the validation error I was expecting by not setting `image_byte_stride` to `0` if `extent.depth > 1`.

We ran into the validation error for `copy_buffer_to_texture` in https://github.com/gfx-rs/wgpu/pull/3063 but not yet for `copy_texture_to_buffer`. The metal docs have the same requirements for `image_byte_stride` in both those cases. So, I don't know why I could not produce the validation error.

If someone more familiar with macOS/Metal could help on this front that would be great but it seems that this issue isn't a high priority right now.

I mainly left this open and as draft to remind me to look at it at some later time but closing it and opening an issue sounds like the better approach.

PR #4159, opened 2023-09-20
Added wgpu.ico

dcerisano:
Multipart favicon with resolutions from 16x16 to 256x256


cwfitzgerald:
I mean, this is neat, but why?

dcerisano:
Well, WebGPU is for the web, and websites have favicons.

As I commented, I created this for personal use, and new websites that are
specifically promoting WebGPU features (demos, wgsl editors, etc).

eg:

   - https://webgpulab.xbdev.net/
   - https://s-macke.github.io/WebGPU-Lab/build/index.html/
   - https://www.wgsl.dev/
   - etc


There are many more that could use an official webgpu favicon.  Actually
the number of these seems to be exploding.
I imagine a major site will emerge for WebGPU demos, like ShaderToy
<https://www.shadertoy.com/> is for WebGL.

Wumpf:
aren't 32x32 pngs how favicons are done these days? Overall leaning against adding this kind of resource to the repo if there's not a good reason: I wouldn't even encouraging the wgpu logo as a favicon like ever (except maybe for [wgpu.rs](wgpu.rs)): It's neat but not great at that size and why would you use that instead of what-ever else your website is about.
Like, thank god there's now webgl icon on shadertoy.com! 😄

cwfitzgerald:
Yeah, I'm going to say that, while an ico is useful in the general sense, it's doesn't really belong here. The wgpu logo is already in tree and can be used as a source for whatever uses people need.

dcerisano:
Favicons are a common requirement, public facing and non-trival to produce
since they have to meet Material Design requirements.
I rather think such a thing does belong here, where it can easily be found.

PR #4165, opened 2023-09-22
Specify the metal dependency in hal instead of the workspace

nical:
**Connections**

https://bugzilla.mozilla.org/show_bug.cgi?id=1854615

**Description**

Specifying the dependency at the workspace level does not appear to work with cargo vendor which Firefox uses.


fornwall:
It (dependency patching at workspace level) also doesn't work with cargo projects depending on git (#4164):

```toml
wgpu = { git = "https://github.com/gfx-rs/wgpu", branch = "trunk" }
```

The above pulls in version 0.26 of metal, so wgpu fails to build.

nical:
Fixed in #4164

Issue #4167, opened 2023-09-22
`ComputePipeline::get_bind_group_layout` panics if the pipeline's module is invalid


jimblandy:
The following test case fails:

```
use wasm_bindgen_test::*;
use wgpu_test::{fail, initialize_test, FailureCase, TestParameters};

#[test]
#[wasm_bindgen_test]
fn pipeline_default_layout_bad_module() {
    // Create an invalid shader and a compute pipeline using it with a
    // default bindgroup layout, and then ask for its layout.
    // Validation should fail but there should be not panic.
    let parameters = TestParameters::default()
        .expect_fail(FailureCase::always());
    initialize_test(parameters, |ctx| {
        ctx.device.push_error_scope(wgpu::ErrorFilter::Validation);

        fail(&ctx.device, || {
            let module = ctx.device.create_shader_module(wgpu::ShaderModuleDescriptor {
                label: None,
                source: wgpu::ShaderSource::Wgsl("not valid wgsl".into()),
            });

            let pipeline = ctx.device.create_compute_pipeline(&wgpu::ComputePipelineDescriptor {
                label: Some("mandelbrot compute pipeline"),
                layout: None,
                module: &module,
                entry_point: "doesn't exist",
            });

            pipeline.get_bind_group_layout(0);
        });
    });
}
```

To see the error:
- Adding the above as a new module of `tests/tests/src/root.rs` named `pipeline.rs`
- Remove the `.expect_fail` call and the `FailureCase` import.

We see the error as follows:

```
$ RUST_BACKTRACE=1 cargo nextest run --test wgpu-tests pipeline::pipeline_default_layout_bad_module
    Finished test [unoptimized + debuginfo] target(s) in 0.14s
    Starting 1 test across 1 binary (60 skipped)
        FAIL [   0.149s] wgpu-test::wgpu-tests pipeline::pipeline_default_layout_bad_module

--- STDOUT:              wgpu-test::wgpu-tests pipeline::pipeline_default_layout_bad_module ---

running 1 test
test pipeline::pipeline_default_layout_bad_module ... FAILED

failures:

failures:
    pipeline::pipeline_default_layout_bad_module

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 60 filtered out; finished in 0.13s


--- STDERR:              wgpu-test::wgpu-tests pipeline::pipeline_default_layout_bad_module ---
[...]

PR #4168, opened 2023-09-22
Add test case for #4167, marked as an expected failure.

jimblandy:
This adds a test case for #4167, marked as an expected failure with a comment referring to the issue.

**Checklist**

- [X] Run `cargo clippy`.

### DIFF
--- a/tests/tests/pipeline.rs
+++ b/tests/tests/pipeline.rs
@@ -1,0 +1,37 @@
+use wasm_bindgen_test::*;
+use wgpu_test::{fail, initialize_test, FailureCase, TestParameters};
+
+#[test]
+#[wasm_bindgen_test]
+fn pipeline_default_layout_bad_module() {
+    // Create an invalid shader and a compute pipeline that uses it
+    // with a default bindgroup layout, and then ask for that layout.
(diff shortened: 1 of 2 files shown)


tests/tests/pipeline.rs
```diff
@@ -0,0 +1,36 @@
...
+
+#[test]
+#[wasm_bindgen_test]
+fn pipeline_default_layout_bad_module() {
+    // Create an invalid shader and a compute pipeline using it with a
+    // default bindgroup layout, and then ask for its layout.
+    // Validation should fail but there should be not panic.
+    let parameters = TestParameters::default()
+        // https://github.com/gfx-rs/wgpu/issues/4167
+        .expect_fail(FailureCase::always());
```
cwfitzgerald, line 13: We need a webgl skip here for webassembly.

cwfitzgerald (approved):

jimblandy:
I think this is what you asked for??

Do we really have to skip every test that might panic on wasm?

cwfitzgerald:
Yeah we do - there's not a lot of good ways around this as we aren't using our own test harness on wasm.

PR #4172, opened 2023-09-25
Re-export Naga

### LINKED ISSUE #3559: Reexport items needed to create a ShaderSource::Glsl

I recently removed `shaderc` from my game's engine and started relying on naga to compile glsl in runtime (for both hot-reloading and not). All shaders work great, but the experience of doing this was somewhat sketchy.

`ShaderSource::Glsl` requires you to provide a `naga::ShaderStage` and a `rustc_hash::FxHashMap`, neither of which is exported from `wgpu`, unless I am mistaken. So I had to add top-level dependencies on these that version-match with what the gfx ecosystem uses.

I'd like to import these directly from wgpu, something like:

```rust
let foo = wgpu::ShaderSource::Glsl {
    shader: Cow::from(FOO_VS),
    stage: wgpu::naga::ShaderStage::Vertex,
    defines: wgpu::naga::FastHashMap::default(),
}
```

exrook:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Fixes #3559

**Description**
Allow users to use the same version of Naga wgpu is using.

Allows the `wgpu::ShaderSource::Naga` and `wgpu::ShaderSource::Glsl` apis to be used without the user having to manually match their own `naga` dependency version to that of `wgpu`.

**Testing**
n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,6 +89,7 @@ By @wumpf in [#4147](https://github.com/gfx-rs/wgpu/pull/4147)
 ### Added/New Features
 
 - Add `gles_minor_version` field to `wgpu::InstanceDescriptor`. By @PJB3005 in [#3998](https://github.com/gfx-rs/wgpu/pull/3998)
+- Re-export Naga. By @exrook in [#4172](https://github.com/gfx-rs/wgpu/pull/4172)
 
 ### Changes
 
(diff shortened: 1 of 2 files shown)


jimblandy (approved):
This looks good. At first I thought it was overkill for just the shader stage and the hash table type, but people also need the right Naga if they want to use `ShaderSource::Naga`, so this is probably inevitable.

jimblandy:
@exrook I've enabled CI for this PR - could you take a look at the failures?

exrook:
@jimblandy Just rustfmt wanting the `use` statements in alphabetical order, fixed :+1:

PR #4183, opened 2023-09-28
Add trace level logging to most API entry points

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Description**

This PR adds trace-level logs to all or most of the entry points in wgpu-core. A fair amount of debug level logs were converted to trace in the process.

### DIFF
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -262,7 +262,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             };
 
             let id = fid.assign(buffer, &mut token);
-            log::info!("Created buffer {:?} with {:?}", id, desc);
+            log::trace!("Device::create_buffer -> {:?}", id);
 
             device
                 .trackers
@@ -482,7 +482,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             //TODO: lock pending writes separately, keep the device read-only
             let (mut device_guard, mut token) = hub.devices.write(&mut token);
 
-            log::info!("Buffer {:?} is destroyed", buffer_id);
+            log::trace!("Buffer::destroy {buffer_id:?}");
             let (mut buffer_guard, _) = hub.buffers.write(&mut token);
             let buffer = buffer_guard
                 .get_mut(buffer_id)
@@ -533,7 +533,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
 
     pub fn buffer_drop<A: HalApi>(&self, buffer_id: id::BufferId, wait: bool) {
         profiling::scope!("Buffer::drop");
-        log::debug!("buffer {:?} is dropped", buffer_id);
+        log::trace!("Buffer::drop {buffer_id:?}");
 
         let hub = A::hub(self);
         let mut token = Token::root();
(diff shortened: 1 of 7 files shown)


nical:
I didn't format many of the arguments in the log (mostly the ID of the object in question) in this PR, We should probably   format more arguments (the descriptors may be a bit much but there's a fair amount of other useful info) Let me know what you think. I also didn't add much logging to the error paths, I can add them in this PR or a followup.

wgpu-core/src/device/global.rs
```diff
@@ -2043,7 +2081,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
 
     pub fn compute_pipeline_drop<A: HalApi>(&self, compute_pipeline_id: id::ComputePipelineId) {
         profiling::scope!("ComputePipeline::drop");
-        log::debug!("compute pipeline {:?} is dropped", compute_pipeline_id);
+        log::debug!("ComputePipeline::drop {:?}", compute_pipeline_id);
```
teoxoy:
```suggestion
        log::trace!("ComputePipeline::drop {:?}", compute_pipeline_id);
```

wgpu-core/src/device/global.rs
```diff
@@ -1669,7 +1700,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
 
     pub fn render_bundle_drop<A: HalApi>(&self, render_bundle_id: id::RenderBundleId) {
         profiling::scope!("RenderBundle::drop");
-        log::debug!("render bundle {:?} is dropped", render_bundle_id);
+        log::debug!("RenderBundle::drop {:?}", render_bundle_id);
```
teoxoy:
```suggestion
        log::trace!("RenderBundle::drop {:?}", render_bundle_id);
```

wgpu-core/src/device/global.rs
```diff
@@ -780,14 +780,14 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         texture_id: id::TextureId,
     ) -> Result<(), resource::DestroyError> {
         profiling::scope!("Texture::destroy");
+        log::info!("Texture::destroy {texture_id:?}");
```
teoxoy:
```suggestion
        log::trace!("Texture::destroy {texture_id:?}");
```

teoxoy (commented):
Should these be `trace` as well?

nical:
indeed!

teoxoy (approved):
LGTM

ErichDonGubler (approved):
LGTM!

The use case I can imagine for the additions specifically appears to be adding context for other log messages. This sort of context-keeping is very similar in domain to `tracing`'s "span"-based model. I wonder if this might be worth exploring for `wgpu` later; it might obviate things like this PR later, and avoid pitfalls when multiple threads are involved (which I think we do for `wgpu` right now?).

EDIT: olol, @teoxoy beat me to approval. 😆

nical:
My main motivation right now is to easily inspect what failing webgpu pages do in firefox's about:logging. If we can get that with spans then :+1:

JMS55:
Did wgpu ever compile-out its logs with a cargo feature? In the past, Bevy has noted that a significant amount of time is wasted in wgpu checking if logs are enabled or not at runtime: https://github.com/bevyengine/bevy/pull/7223. I can't find an issue or PR about it in this repo though.

nical:
If you depend on the log crate with a feature flag like `release_max_level_info`, the `trace` level logs should be discarded at build time. You can do it in bevy directly and that should affect all of the dependencies that use the same version of `log`.

PR #4184, opened 2023-09-28
Fix x11 hang when resizing on vulkan

### LINKED ISSUE #4180: Examples crash when resizing on x11+vulkan (known commit)
**Description**
Running the `cube` example, my window manager (on x11) hangs when resizing the window.
I have pinpointed the bug to the following commit: https://github.com/gfx-rs/wgpu/commit/1d0a22b1a8f38d9c9fe3418c543ee2bfaa046cb2

**Repro steps**
- Linux w/ X11
- Run the `cube` example
- Resize the window

Expected: should resize and run smoothly
Observed: My WM hangs for ~5 seconds. Then it becomes responsive again, but the window has not been redrawn. Attempting to resize again at this point will *not* make the WM hang a second time, but the window never gets redrawn again.

**Extra materials**
Trace:
[wgpu_resize_bug_trace.zip](https://github.com/gfx-rs/wgpu/files/12744032/wgpu_resize_bug_trace.zip)

**Platform**
Archlinux, X11, no compositor
WGPU version: Any commit since https://github.com/gfx-rs/wgpu/commit/1d0a22b1a8f38d9c9fe3418c543ee2bfaa046cb2
Adapter: NVIDIA GeForce RTX 2060 (Vulkan)

Azorlogh:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Fixes https://github.com/gfx-rs/wgpu/issues/4180

**Description**
From @cwfitzgerald : The problem is that we actually don't have access to wait until presentation is finished. I suspect what is happening is that without the device.wait_idle() call whatever presentation things are happening are still in progress and calling device.wait_idle() is the only way to actually do that

**Testing**
Reproduced the steps in https://github.com/gfx-rs/wgpu/issues/4180, and everything resizes smoothly again

### DIFF
--- a/wgpu-hal/src/vulkan/instance.rs
+++ b/wgpu-hal/src/vulkan/instance.rs
@@ -159,6 +159,8 @@ impl super::Swapchain {
         profiling::scope!("Swapchain::release_resources");
         {
             profiling::scope!("vkDeviceWaitIdle");
+            // We need to also wait until all presentation work is done. Because there is no way to portably wait until
+            // the presentation work is done, we are forced to wait until the device is idle.
             let _ = unsafe { device.device_wait_idle() };
         };
         unsafe { device.destroy_fence(self.fence, None) };
--- a/wgpu-hal/src/vulkan/instance.rs
+++ b/wgpu-hal/src/vulkan/instance.rs
@@ -157,6 +157,10 @@ impl super::Swapchain {
     /// - The device must have been made idle before calling this function.
     unsafe fn release_resources(self, device: &ash::Device) -> Self {
         profiling::scope!("Swapchain::release_resources");
+        {
+            profiling::scope!("vkDeviceWaitIdle");
+            let _ = unsafe { device.device_wait_idle() };
+        };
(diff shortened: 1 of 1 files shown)


wgpu-hal/src/vulkan/instance.rs
```diff
@@ -157,6 +157,10 @@ impl super::Swapchain {
     /// - The device must have been made idle before calling this function.
     unsafe fn release_resources(self, device: &ash::Device) -> Self {
         profiling::scope!("Swapchain::release_resources");
+        {
+            profiling::scope!("vkDeviceWaitIdle");
+            let _ = unsafe { device.device_wait_idle() };
```
cwfitzgerald, line 164:
```suggestion
            // We need to also wait until all presentation work is done. Because there is no way to portably wait until
            // the presentation work is done, we are forced to wait until the device is idle.
            let _ = unsafe { device.device_wait_idle() };
```

cwfitzgerald (approved):
Adding a comment

Issue #4109, opened 2023-09-01
Example hello fails on MS Windows


tigerwings:
**Description**
Run hello on Microsoft Windows 10
```
$ set WGPU_BACKEND=dx12
$ cargo run --example hello
    Finished dev [unoptimized + debuginfo] target(s) in 0.27s
     Running `target\debug\examples\hello.exe`
Error loading EGL entry points.
error: process didn't exit successfully: `target\debug\examples\hello.exe` (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
```

while `cargo run --example cube` works.

**Repro steps**
After tracing down the issue, it seems [this line](https://github.com/gfx-rs/wgpu/blob/trunk/wgpu-hal/src/gles/egl.rs#L756) is the root cause. And the stack trace is the following:

```
<unknown> 0x0000000000000000
[Inlined] khronos_egl::impl$98::eglQueryString(*mut khronos_egl::Dynamic<libloading::safe::Library,khronos_egl::EGL1_4>,*mut core::ffi::c_void,i32) lib.rs:2032
khronos_egl::Instance<khronos_egl::Dynamic<libloading::safe::Library,khronos_egl::EGL1_4> >::query_string<khronos_egl::Dynamic<libloading::safe::Library,khronos_egl::EGL1_4> >(enum2$<core::option::Option<khronos_egl::egl1_0::Display> >,i32) lib.rs:987
wgpu_hal::gles::egl::impl$14::init(*mut wgpu_hal::InstanceDescriptor) egl.rs:756
wgpu_core::instance::impl$0::new::init<wgpu_hal::gles::Api>(wgpu_hal::gles::Api,*mut wgpu_types::InstanceDescriptor) instance.rs:87
wgpu_core::instance::Instance::new(str,wgpu_types::InstanceDescriptor) instance.rs:104
wgpu_core::global::Global<wgpu_core::identity::IdentityManagerFactory>::new<wgpu_core::identity::IdentityManagerFactory>(str,wgpu_core::identity::IdentityManagerFactory,wgpu_types::InstanceDescriptor) global.rs:36
wgpu::backend::direct::impl$7::init(wgpu_types::InstanceDescriptor) direct.rs:570
wgpu::Instance::new(wgpu_types::InstanceDescriptor) lib.rs:1733
wgpu::impl$21::default() lib.rs:1720
[...]

PR #4185, opened 2023-09-28
Make GLES optional on Windows & macOS

teoxoy:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
https://github.com/gfx-rs/wgpu/issues/4109#issuecomment-1739651614

**Description**
Users of GLES on Windows and macOS should have to opt into using the GLES backend via Angle since they'd have to package it with the app. Otherwise we might pick up a random version of Angle on their system.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,12 @@ depth_ops: Some(wgpu::Operations {
 
 By @wumpf in [#4147](https://github.com/gfx-rs/wgpu/pull/4147)
 
+#### The GLES backend is now optional on Windows & macOS
+
+The `angle` feature flag has to be set for the GLES backend to be enabled on Windows & macOS.
+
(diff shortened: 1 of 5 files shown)


cwfitzgerald:
I'm hesitant to go forward with this - this seems to be moving in the opposite direction of what we want. We were trying to make features work on all platforms so that the features can be enabled everywhere (also removing the need for the second resolver). I do agree we probably shouldn't have angle enabled by default, but this may not be the right way. I'm not sure what the right way is though.

teoxoy:
I haven't added any restrictions though, only made the GLES backend optional by default on the platforms that need Angle. This is similar to `vulkan-portability` needing MoltenVK.

cwfitzgerald (approved):
Yeah, totally misread the PR when I looked at it last, half asleep. Let's do it

PR #4196, opened 2023-10-02
[WebGPU] validate polygon mode

### LINKED ISSUE #4194: [WebGPU] Error if any required feature is not available
**Description**
When requesting POLYGON_MODE_LINE in webgpu it doesn't give an error despite it not working. Webgl correctly gives an error.

A similar issue also happen when using `PolygonMode::Line` without requesting the feature. In webgl it will give an error while in webgpu it gives no error but doesn't render correctly.

**Repro steps**
I discovered this with bevy. I tried making a simple reproduction but I couldn't figure out how to build a simple repro and build it for wasm, I kept getting unrelated errors, I was probably doing something wrong. Is there a simple template somewhere that I could use?

Expected: All platform that do not support `POLYGON_MODE_LINE` to raise an error if the feature is requested
Observed: Some platform that do not support `POLYGON_MODE_LINE`  do not raise an error if the feature is requested

**Platform**
I was using windows, but I don't think this is OS related.
This was found with bevy 0.12-dev which is still on wgpu 0.16

teoxoy:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
fixes https://github.com/gfx-rs/wgpu/issues/4194

### DIFF
--- a/wgpu/src/backend/web.rs
+++ b/wgpu/src/backend/web.rs
@@ -340,6 +340,18 @@ fn map_primitive_state(primitive: &wgt::PrimitiveState) -> web_sys::GpuPrimitive
     //TODO:
     //mapped.unclipped_depth(primitive.unclipped_depth);
 
+    match primitive.polygon_mode {
+        wgt::PolygonMode::Fill => {}
+        wgt::PolygonMode::Line => panic!(
+            "{:?} is not enabled for this backend",
+            wgt::Features::POLYGON_MODE_LINE
+        ),
+        wgt::PolygonMode::Point => panic!(
+            "{:?} is not enabled for this backend",
+            wgt::Features::POLYGON_MODE_POINT
+        ),
+    }
+
     mapped
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -150,6 +150,7 @@ By @wumpf in [#4147](https://github.com/gfx-rs/wgpu/pull/4147)
 #### WebGPU
 
 - Ensure that limit requests and reporting is done correctly. By @OptimisticPeach in [#4107](https://github.com/gfx-rs/wgpu/pull/4107)
+- Validate usage of polygon mode. By @teoxoy in [#4196](https://github.com/gfx-rs/wgpu/pull/4196)
 
 #### Testing
 


wgpu/src/backend/web.rs
```diff
@@ -340,6 +340,18 @@ fn map_primitive_state(primitive: &wgt::PrimitiveState) -> web_sys::GpuPrimitive
     //TODO:
     //mapped.unclipped_depth(primitive.unclipped_depth);
 
+    match primitive.polygon_mode {
+        wgt::PolygonMode::Fill => {}
```
ErichDonGubler, line 344: nitpick: I think `()` is generally preferable over `{}`; it's more direct.

Pushed as 48f701bb.

wgpu/src/backend/web.rs
```diff
@@ -340,6 +340,18 @@ fn map_primitive_state(primitive: &wgt::PrimitiveState) -> web_sys::GpuPrimitive
...
+        wgt::PolygonMode::Line => panic!(
+            "{:?} is not enabled for this backend",
+            wgt::Features::POLYGON_MODE_LINE
+        ),
+        wgt::PolygonMode::Point => panic!(
+            "{:?} is not enabled for this backend",
+            wgt::Features::POLYGON_MODE_POINT
+        ),
```
ErichDonGubler, line 352: nitpick: Pushed this suggestion as 16ff4df1:

```suggestion
        mode @ (wgt::PolygonMode::Line | wgt::PolygonMode::Point) => {
            panic!("{mode:?} is not enabled for this backend",)
        }
```

ErichDonGubler (approved):
Pushed up my nitpicks as `fixup!` commits, so you can autosquash. If you disagree, feel free to drop the relevant commits instead. 🙂

Non-blocking, but: Is there a way we can meaningfully test for this? Not sure what appetite we have for backend-specific tests, but it'd be nice to not regress later.

wgpu/src/backend/web.rs
teoxoy: I think the panic should mention the feature rather than the mode itself. There is precedent for this, if we want to use the mode instead we should discuss about it and do it everywhere.

wgpu/src/backend/web.rs
teoxoy: It's not very often I see `()` over `{}` in this case. I personally prefer `{}` but don't care much as long as we consistently use one of them, the only reliable way of doing that being through rustfmt. However, there seems to be no such rule we could enable (https://rust-lang.github.io/rustfmt).

cwfitzgerald: I also prefer {}.  It's standard for "empty block" == do nothing, as opposed "returns the empty tuple" == do something by returning a nothing.

wgpu/src/backend/web.rs
cwfitzgerald: Yeah, the feature is the error you get elsewhere as well. We should keep this consistant.

cwfitzgerald (commented):
Going to respectfully veto @ErichDonGubler's changes, and this is good to go once they're reverted.

ErichDonGubler:
> Going to respectfully veto @ErichDonGubler's changes, and this is good to go once they're reverted.

🤔 Huh, I didn't expect these to be controversial. Why's that?

cwfitzgerald:
I responded individually in each subthread :)

wgpu/src/backend/web.rs
ErichDonGubler: Ah, I see how this refactor was incorrect. Sorry about that!

PR #4199, opened 2023-10-03
Add `Rgb10a2Uint` format

teoxoy:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
https://github.com/gpuweb/gpuweb/issues/3841
https://github.com/gfx-rs/naga/pull/2525


wgpu-hal/src/vulkan/conv.rs
```diff
@@ -34,6 +34,7 @@ impl super::PrivateCapabilities {
             Tf::Bgra8Unorm => F::B8G8R8A8_UNORM,
             Tf::Rgba8Uint => F::R8G8B8A8_UINT,
             Tf::Rgba8Sint => F::R8G8B8A8_SINT,
+            Tf::Rgb10a2Uint => F::A2B10G10R10_UINT_PACK32,
             Tf::Rgb10a2Unorm => F::A2B10G10R10_UNORM_PACK32,
```
Wumpf, line 38: bit confused why these and not `A2R10G10B10_UINT_PACK32`/`A2R10G10B10_UNORM_PACK32`
but it's consistent what was talked about over at WebGPU repo

Wumpf (approved):
what a beautiful format 😄

wgpu-hal/src/vulkan/conv.rs
teoxoy: Vulkan's packed formats are written in reverse (relative to the other ones) for some reason.
See https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap49.html#texel-block-size

Wumpf: huh fascinating. Well guess it makes sense when one thinks about where the bits are (red bits are low bits)

PR #4200, opened 2023-10-03
Implement a missing part of bind group layout deduplication

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

- Blocking https://bugzilla.mozilla.org/show_bug.cgi?id=1856371
- Original BGL de-duplication PR: https://github.com/gfx-rs/wgpu/pull/3925

**Description**

`render_pipeline_get_bind_group_layout` should do bind group layout de-duplication, but it currently does not.

It will probably help the review to consider the two commits separately:
 - The first one changes the way duplicate bind group layouts are stored: most of the data is only ever stored in the case of the "leaf" layouts (not duplicates), while duplicates mostly contain the ID to forward to. Before this commit we would create a real raw bind group layout for duplicates (but then using it would be a bug since we always want to use the one the duplicate points to). In retrospect I should have done that in the initial BGL de-duplication PR, as it avoids creating resources unnecessarily and removes the possibility for making mistakes. This makes the second commit easier to write:
 - The second commit implements the missing bit of bind group layout de-duplication in `render_pipeline_get_bind_group_layout`. The latter must create a new (duplicate) BGL if the ID is provided externally but was missing that part before this commit.

I wrote this with the assumption that the `MultiRefCount` should be zero at the creation of a bind group layout, I believe it is correct but it would not hurt to look at it with a critical eye during the review.

@gents83 This PR will probably collide a bit with arcanization as most do I assume. The main thing that matters for arcanization is the change in the second commit where instead of creating a new BGL, a duplicate that references an existing one is created, similarly to what was done in https://github.com/gfx-rs/wgpu/pull/3925

**Testing**

That's covered by the CTS in Firefox, but unfortunately tedious to test here because wgpu does not exercise this code path.

### DIFF
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -473,6 +473,7 @@ pub struct BindGroupLayoutInner<A: hal::Api> {
 }
 
 impl<A: hal::Api> BindGroupLayout<A> {
+    #[track_caller]
     pub(crate) fn assume_deduplicated(&self) -> &BindGroupLayoutInner<A> {
         self.as_inner().unwrap()
     }
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -446,23 +446,57 @@ pub type BindGroupLayouts<A> = crate::storage::Storage<BindGroupLayout<A>, BindG
 ///  - produced bind groups
 ///  - produced pipeline layouts
 ///  - pipelines with implicit layouts
-#[derive(Debug)]
 pub struct BindGroupLayout<A: hal::Api> {
-    pub(crate) raw: A::BindGroupLayout,
     pub(crate) device_id: Stored<DeviceId>,
     pub(crate) multi_ref_count: MultiRefCount,
-    pub(crate) entries: BindEntryMap,
     // When a layout created and there already exists a compatible layout the new layout
     // keeps a reference to the older compatible one. In some places we substitute the
     // bind group layout id with its compatible sibling.
     // Since this substitution can come at a cost, it is skipped when wgpu-core generates
     // its own resource IDs.
-    pub(crate) compatible_layout: Option<Valid<BindGroupLayoutId>>,
+    pub(crate) inner: BglOrDuplicate<A>,
+}
+
+pub(crate) enum BglOrDuplicate<A: hal::Api> {
+    Inner(BindGroupLayoutInner<A>),
+    Duplicate(Valid<BindGroupLayoutId>),
+}
+
+pub struct BindGroupLayoutInner<A: hal::Api> {
+    pub(crate) raw: A::BindGroupLayout,
+    pub(crate) entries: BindEntryMap,
     #[allow(unused)]
     pub(crate) dynamic_count: usize,
     pub(crate) count_validator: BindingTypeMaxCountValidator,
     #[cfg(debug_assertions)]
     pub(crate) label: String,
+}
+
+impl<A: hal::Api> BindGroupLayout<A> {
+    pub(crate) fn assume_deduplicated(&self) -> &BindGroupLayoutInner<A> {
+        self.as_inner().unwrap()
+    }
+
+    pub(crate) fn as_inner(&self) -> Option<&BindGroupLayoutInner<A>> {
+        match self.inner {
+            BglOrDuplicate::Inner(ref inner) => Some(inner),
+            BglOrDuplicate::Duplicate(_) => None,
+        }
+    }
+
+    pub(crate) fn into_inner(self) -> Option<BindGroupLayoutInner<A>> {
+        match self.inner {
+            BglOrDuplicate::Inner(inner) => Some(inner),
+            BglOrDuplicate::Duplicate(_) => None,
+        }
+    }
+
+    pub(crate) fn as_duplicate(&self) -> Option<Valid<BindGroupLayoutId>> {
+        match self.inner {
+            BglOrDuplicate::Duplicate(id) => Some(id),
+            BglOrDuplicate::Inner(_) => None,
+        }
+    }
 }
 
 impl<A: hal::Api> Resource for BindGroupLayout<A> {
@@ -474,7 +508,7 @@ impl<A: hal::Api> Resource for BindGroupLayout<A> {
 
     fn label(&self) -> &str {
         #[cfg(debug_assertions)]
-        return &self.label;
+        return self.as_inner().map_or("", |inner| &inner.label);
         #[cfg(not(debug_assertions))]
         return "";
     }
@@ -486,8 +520,8 @@ pub(crate) fn try_get_bind_group_layout<A: HalApi>(
     id: BindGroupLayoutId,
 ) -> Option<&BindGroupLayout<A>> {
     let layout = layouts.get(id).ok()?;
-    if let Some(compat) = layout.compatible_layout {
-        return Some(&layouts[compat]);
+    if let BglOrDuplicate::Duplicate(original_id) = layout.inner {
+        return Some(&layouts[original_id]);
     }
 
     Some(layout)
@@ -499,9 +533,8 @@ pub(crate) fn get_bind_group_layout<A: HalApi>(
 ) -> (Valid<BindGroupLayoutId>, &BindGroupLayout<A>) {
     let layout = &layouts[id];
     layout
-        .compatible_layout
-        .map(|compat| (compat, &layouts[compat]))
-        .unwrap_or((id, layout))
+        .as_duplicate()
+        .map_or((id, layout), |deduped| (deduped, &layouts[deduped]))
 }
 
 #[derive(Clone, Debug, Error)]
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -1,7 +1,8 @@
 #[cfg(feature = "trace")]
 use crate::device::trace;
 use crate::{
-    binding_model, command, conv,
+    binding_model::{self, BindGroupLayout},
+    command, conv,
     device::{life::WaitIdleError, map_buffer, queue, Device, DeviceError, HostMap},
     global::Global,
     hal_api::HalApi,
@@ -1115,7 +1116,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             }
 
             let mut compatible_layout = None;
-            {
+            let layout = {
                 let (bgl_guard, _) = hub.bind_group_layouts.read(&mut token);
                 if let Some(id) =
                     Device::deduplicate_bind_group_layout(device_id, &entry_map, &*bgl_guard)
@@ -1134,18 +1135,25 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
 
                     compatible_layout = Some(id::Valid(id));
                 }
-            }
 
-            let mut layout = match device.create_bind_group_layout(
-                device_id,
-                desc.label.borrow_option(),
-                entry_map,
-            ) {
-                Ok(layout) => layout,
-                Err(e) => break e,
+                if let Some(original_id) = compatible_layout {
+                    let original = &bgl_guard[original_id];
+                    BindGroupLayout {
+                        device_id: original.device_id.clone(),
+                        inner: crate::binding_model::BglOrDuplicate::Duplicate(original_id),
+                        multi_ref_count: crate::MultiRefCount::new(),
+                    }
+                } else {
+                    match device.create_bind_group_layout(
+                        device_id,
+                        desc.label.borrow_option(),
+                        entry_map,
+                    ) {
+                        Ok(layout) => layout,
+                        Err(e) => break e,
+                    }
+                }
             };
-
-            layout.compatible_layout = compatible_layout;
 
             let id = fid.assign(layout, &mut token);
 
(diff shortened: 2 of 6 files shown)


gents83:
Do you think we can wait to do this after Arcanization or do you prefer me to re-implement/re-check it after your submit?

nical:
> Do you think we can wait to do this after Arcanization or do you prefer me to re-implement/re-check it after your submit?

It blocks some important stuff in firefox and our deadlines are getting pretty tight, so we can wait for a few days but not much longer (we actually held off #3925 for about 6 weeks because we thought arcanization was about to land, but as you know it's a difficult beast to push over the finish line!).

gents83:
Oki, this week I'll be away for work, so I'll be able to be back on this during the weekend.

If you need to go on, no problem, I'll catch up 👍🏼

wgpu-core/src/binding_model.rs
```diff
@@ -446,25 +446,64 @@ pub type BindGroupLayouts<A> = crate::storage::Storage<BindGroupLayout<A>, BindG
...
     // Since this substitution can come at a cost, it is skipped when wgpu-core generates
     // its own resource IDs.
-    pub(crate) compatible_layout: Option<Valid<BindGroupLayoutId>>,
+    pub(crate) inner: BglOrDuplicate<A>,
+}
+
+pub(crate) enum BglOrDuplicate<A: hal::Api> {
+    Inner(BindGroupLayoutInner<A>),
+    Duplicate(Valid<BindGroupLayoutId>),
+}
```
ErichDonGubler, line 463: nitpick: I think this naming might be clearer:

```suggestion
pub(crate) enum BindGroupLayoutDedupedInner<A: hal::Api> {
    Original(BindGroupLayoutInner<A>),
    Duplicate(Valid<BindGroupLayoutId>),
}
```

…with according renames for methods, of course.

wgpu-core/src/binding_model.rs
```diff
@@ -446,25 +446,64 @@ pub type BindGroupLayouts<A> = crate::storage::Storage<BindGroupLayout<A>, BindG
...
     #[cfg(debug_assertions)]
     pub(crate) label: String,
 }
 
+impl<A: hal::Api> BindGroupLayout<A> {
+    pub(crate) fn assume_deduplicated(&self) -> &BindGroupLayoutInner<A> {
+        match self.inner {
+            BglOrDuplicate::Inner(ref inner) => inner,
+            BglOrDuplicate::Duplicate(_) => {
+                panic!()
```
ErichDonGubler: - [x] issue: Having an empty panic message like this is usually an indicator that we don't know enough about what we're doing to provide a good diagnostic. We should know better, IMO.

    Is there a reason we don't follow the same pattern as `{as,into}_duplicate` here by returning an `Option<&BindGorupLayoutInner<A>>`, and let consuming code call `Option::unwrap` instead?

wgpu-core/src/binding_model.rs
```diff
@@ -474,7 +513,11 @@ impl<A: hal::Api> Resource for BindGroupLayout<A> {
 
     fn label(&self) -> &str {
         #[cfg(debug_assertions)]
-        return &self.label;
+        return if let BglOrDuplicate::Inner(ref inner) = self.inner {
+            &inner.label
+        } else {
+            ""
+        };
```
ErichDonGubler: nitpick:

```suggestion
        return self.as_inner().map_or("", |inner| &inner.label);
```

Pushed as a `fixup!` with 466309dc.

wgpu-core/src/binding_model.rs
ErichDonGubler, line 463: praise: Having a bespoke `Option`-alike `enum` actually feels nicely self-documenting. I feel more confident that I know what “real” bind group layout data is used for and when, with `raw` and `entries` getting moved behind a discriminant. Thanks!

wgpu-core/src/binding_model.rs
```diff
@@ -474,7 +513,11 @@ impl<A: hal::Api> Resource for BindGroupLayout<A> {
 
     fn label(&self) -> &str {
         #[cfg(debug_assertions)]
```
ErichDonGubler, line 511: question(non-blocking): Shouldn't we always preserve the `label`, according to the WebGPU spec.? 🤔

wgpu-core/src/binding_model.rs
```diff
@@ -498,10 +541,10 @@ pub(crate) fn get_bind_group_layout<A: HalApi>(
...
 ) -> (Valid<BindGroupLayoutId>, &BindGroupLayout<A>) {
     let layout = &layouts[id];
-    layout
-        .compatible_layout
-        .map(|compat| (compat, &layouts[compat]))
-        .unwrap_or((id, layout))
+    if let BglOrDuplicate::Duplicate(original_id) = layout.inner {
+        return (original_id, &layouts[original_id]);
+    }
+    (id, layout)
```
ErichDonGubler: nitpick: This diverges less from previous code, and I think it's easier to follow:

```suggestion
    layout
        .as_duplicate()
        .map_or((id, layout), |deduped| (deduped, &layouts[deduped]))
```

Pushed as a `fixup!` with 03a51250.

wgpu-core/src/device/global.rs
```diff
@@ -1134,19 +1135,26 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+                    ) {
+                        Ok(layout) => layout,
+                        Err(e) => break e,
+                    }
+                }
             };
 
-            layout.compatible_layout = compatible_layout;
-
             let id = fid.assign(layout, &mut token);
```
ErichDonGubler, line 1158: praise: This feels much less error-prone, nice.

wgpu-core/src/device/resource.rs
```diff
@@ -1391,9 +1394,15 @@ impl<A: HalApi> Device<A> {
...
-                    && bgl.entries == *entry_map
+                if bgl.device_id.value.0 != self_id {
+                    return false;
+                }
+
+                if let Some(inner) = bgl.as_inner() {
+                    return inner.entries == *entry_map;
+                }
+
+                false
```
ErichDonGubler: nitpick: We can use `Option::map_or` to diverge less from previous code, and be just as clear (IMO):

```suggestion
                bgl.device_id.value.0 == self_id
                    && bgl
                        .as_inner()
                        .map_or(false, |inner| inner.entries == *entry_map)
```

Pushed as a `fixup!` with f99fb845.

wgpu-core/src/device/global.rs
```diff
@@ -1924,7 +1924,24 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
+            // with the given ID as a duplicate of the existing one.
+            let new_layout = BindGroupLayout {
+                device_id: layout.device_id.clone(),
+                inner: crate::binding_model::BglOrDuplicate::<A>::Duplicate(*id),
+                multi_ref_count: crate::MultiRefCount::new(),
+            };
+
+            let fid = hub.bind_group_layouts.prepare(id_in);
+            let id = fid.assign(new_layout, &mut Token::root());
+
```
ErichDonGubler, line 1944: note: This fix matches mentally with what you told me we needed to do yesterday. Yay!

wgpu-core/src/device/resource.rs
```diff
@@ -2316,7 +2341,12 @@ impl<A: HalApi> Device<A> {
         let bgl_vec = desc
             .bind_group_layouts
             .iter()
-            .map(|&id| &try_get_bind_group_layout(bgl_guard, id).unwrap().raw)
+            .map(|&id| {
+                &try_get_bind_group_layout(bgl_guard, id)
+                    .unwrap()
+                    .assume_deduplicated()
+                    .raw
+            })
```
ErichDonGubler, line 2341: question(non-blocking): Only tangentially relevant, but: Are there any other places we should be using `{try_,}get_bind_group_layout` (as follow-up work)?

This is a lazy question; I can try to answer this myself (TODO), independent of this PR.

wgpu-core/src/device/life.rs
```diff
@@ -786,7 +787,9 @@ impl<A: HalApi> LifetimeTracker<A> {
                         if let Some(lay) =
                             hub.bind_group_layouts.unregister_locked(id.0, &mut *guard)
                         {
-                            self.free_resources.bind_group_layouts.push(lay.raw);
+                            if let BglOrDuplicate::Inner(inner) = lay.inner {
+                                self.free_resources.bind_group_layouts.push(inner.raw);
+                            }
```
ErichDonGubler: nitpick:

```suggestion
                            if let Some(inner) = lay.into_inner() {
                                self.free_resources.bind_group_layouts.push(inner.raw);
                            }
```

Pushed as a `fixup!` with 66dcca3d.

ErichDonGubler: note: This is the first time I've seen this chain-following code for potentially duplicated BGLs. Seems sound to me, after taking some time with it. 👍🏻

wgpu-core/src/device/resource.rs
```diff
@@ -2304,10 +2326,13 @@ impl<A: HalApi> Device<A> {
...
-            let bind_group_layout = bgl_guard
-                .get(id)
-                .map_err(|_| Error::InvalidBindGroupLayout(id))?;
-            count_validator.merge(&bind_group_layout.count_validator);
+            let bind_group_layout = match try_get_bind_group_layout(bgl_guard, id) {
+                Some(layout) => layout,
+                None => {
+                    return Err(Error::InvalidBindGroupLayout(id));
+                }
+            };
```
ErichDonGubler: nitpick: We could use `let-else` here:

```suggestion
            let Some(bind_group_layout) = try_get_bind_group_layout(bgl_guard, id) else {
                return Err(Error::InvalidBindGroupLayout(id));
            };
```

Pushed as a `fixup!` with 77b37c98.

wgpu-core/src/device/resource.rs
ErichDonGubler: nitpick: This diverges less from previous code, and is still just as clear (IMO):

```suggestion
                bgl.device_id.value.0 == self_id
                    && bgl
                        .as_inner()
                        .map_or(false, |inner| inner.entries == *entry_map)
```

ErichDonGubler:
@gents83: Does this actually look like a significant amount of effort to adjust arcanization for? I don't think there are any significant changes to how tracking with `BindGroupLayout::multi_ref_count` was used here. This is mostly a refactor to use an `enum` representation for the guts of BGLs.

wgpu-core/src/device/resource.rs
```diff
@@ -1823,6 +1840,8 @@ impl<A: HalApi> Device<A> {
         Ok(())
     }
 
+    // This function expects the provided bind group layout to be resolved
+    // (not passing a duplicate) beforehand.
```
ErichDonGubler, line 1839: - [x] question(blocking): I'm not sure how to validate that these should always be original/deduplicated. How can I do that?

ErichDonGubler (approved):
Crazily enough (given my newness to `wgpu`), I actually feel confident in approving on this. @nical onboarded me to his solution synchronously yesterday, which helped a great deal. 🙂

Left mostly non-blocking feedback. For most code style nitpicks, I left `fixup!` commits. Gonna trust @nical to address feedback before merging, and leave a green checkmark here for now.

wgpu-core/src/binding_model.rs
ErichDonGubler: If we _do_ persist with this API (the only callers I see are places where calling code _should_ guarantee that this is a successful check), I would recommend adding [`#[track_caller]`](https://doc.rust-lang.org/reference/attributes/codegen.html#the-track_caller-attribute) on here, for a better default stacktrace.

wgpu-core/src/device/resource.rs
nical: It is called once and the de-duplication is done just before calling it.

wgpu-core/src/binding_model.rs
nical: I don't know how strict the spec is about this but Connor and I talked a while back about keeping labels for everything and decided we should (but nobody has done the work yet).

wgpu-core/src/binding_model.rs
nical: It's rather that the function is self-explanatory enough that we don't need a string to paraphrase what the problem is about.
Calling `assume_deduplicated` better expresses what invariant we have upheld for things to work (this will never panic if we did the de-duplication beforehand), than calling `as_inner().unwrap()` which only says "I'm confident this contains inner bind group layout data".

wgpu-core/src/binding_model.rs
nical: I prefer the current name. It's sandwitched between `BindGroupLayout` and `BindGroupLayoutInner` so in every context where it is used there is zero ambiguity about what "Bgl" means. The enum represents either a real BGL or a duplicate indirection, so the "A or B" formulation seems pretty clear to me (I'm not sure how to interpret `DedupedInner` in your proposal).

wgpu-core/src/device/resource.rs
nical: I guess I should note: calling `get_bind_group_layout` or `try_get_bind_group_layout`  is what resolves the indirection (what I call de-duplication in this context) in various pieces of the code, that's generally what you want to call to make sure you have the real BGL.

wgpu-core/src/device/resource.rs
nical: This refactoring makes it hard to do the wrong thing because in most places where we interact with the BGL, we want to reach into some data that is now stored in the inner struct so you pretty much have to check whether it's a duplicate and follow the indirection to achieve that.

wgpu-core/src/binding_model.rs
ErichDonGubler: 🤷🏻‍♂️Not concerned either way. Onward!

cwfitzgerald (approved):
Looks good :)

cwfitzgerald:
@nical merge when ready, don't want to merge early if you're not done adjusting.

PR #4204, opened 2023-10-03
Split MSRV in Two Parts

cwfitzgerald:
This PR splits our MSRV into two sections:

- Infrastructure and wgpu-rs itself.
- wgpu-core/hal/types for so firefox remains happy.

We need this as both #4143 and #3873 require 1.70 for various non-core parts due to incidental dependencies.


cwfitzgerald (commented):
I'm sending this through without review as it should be strictly additive and requires some administrative changes after it lands. If there are any issues, we can touch it up later.

PR #4205, opened 2023-10-04
fix clippy warnings and errors

TornaxO7:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
None

**Description**
All warnings and errors from `cargo clippy` are resolved.

**Testing**
By just executing `cargo clippy.

### DIFF
--- a/wgpu-hal/src/gles/egl.rs
+++ b/wgpu-hal/src/gles/egl.rs
@@ -796,12 +796,9 @@ impl crate::Instance<super::Api> for Instance {
                         khronos_egl::DEFAULT_DISPLAY,
                         &display_attributes,
                     )
-                    .unwrap();
-                (
-                    display,
-                    Some(Rc::new(library)),
-                    WindowKind::Wayland,
-                )
+                }
+                .unwrap();
+                (display, Some(Rc::new(library)), WindowKind::Wayland)
             } else if let (Some(display_owner), Some(egl)) = (x11_display_library, egl1_5) {
                 log::info!("Using X11 platform");
                 let display_attributes = [khronos_egl::ATTRIB_NONE];
@@ -811,12 +808,9 @@ impl crate::Instance<super::Api> for Instance {
                         display_owner.display.as_ptr(),
                         &display_attributes,
                     )
-                    .unwrap();
-                (
-                    display,
-                    Some(Rc::new(display_owner)),
-                    WindowKind::X11,
-                )
+                }
+                .unwrap();
+                (display, Some(Rc::new(display_owner)), WindowKind::X11)
             } else if let (Some(display_owner), Some(egl)) = (angle_x11_display_library, egl1_5) {
                 log::info!("Using Angle platform with X11");
                 let display_attributes = [
@@ -832,12 +826,9 @@ impl crate::Instance<super::Api> for Instance {
                         display_owner.display.as_ptr(),
                         &display_attributes,
                     )
-                    .unwrap();
-                (
-                    display,
-                    Some(Rc::new(display_owner)),
-                    WindowKind::AngleX11,
-                )
+                }
+                .unwrap();
+                (display, Some(Rc::new(display_owner)), WindowKind::AngleX11)
             } else if client_ext_str.contains("EGL_MESA_platform_surfaceless") {
                 log::info!("No windowing system present. Using surfaceless platform");
                 let egl = egl1_5.expect("Failed to get EGL 1.5 for surfaceless");
@@ -849,6 +840,7 @@ impl crate::Instance<super::Api> for Instance {
                     )
                 }
                 .unwrap();
+
                 (display, None, WindowKind::Unknown)
             } else {
                 log::info!("EGL_MESA_platform_surfaceless not available. Using default platform");
@@ -1320,7 +1312,7 @@ impl crate::Surface<super::Api> for Surface {
         };
 
         if let Some(window) = wl_window {
-            let library = &self.wsi.display_owner.as_ref().unwrap().library ;
+            let library = &self.wsi.display_owner.as_ref().unwrap().library;
             let wl_egl_window_resize: libloading::Symbol<WlEglWindowResizeFun> =
                 unsafe { library.get(b"wl_egl_window_resize") }.unwrap();
             unsafe {
--- a/wgpu-hal/src/gles/egl.rs
+++ b/wgpu-hal/src/gles/egl.rs
@@ -1203,8 +1203,7 @@ impl crate::Surface<super::Api> for Surface {
                         let library = &self.wsi.display_owner.as_ref().unwrap().library;
                         let wl_egl_window_create: libloading::Symbol<WlEglWindowCreateFun> =
                             unsafe { library.get(b"wl_egl_window_create") }.unwrap();
-                        let window = unsafe { wl_egl_window_create(handle.surface, 640, 480) }
-                            as *mut _ as *mut std::ffi::c_void;
+                        let window = unsafe { wl_egl_window_create(handle.surface, 640, 480) };
                         wl_window = Some(window);
                         window
                     }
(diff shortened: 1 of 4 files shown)


cwfitzgerald:
Outside of the one unnecessary cast, how did you get these clippy errors? Were you running with a newer compiler (we by default only use 1.65 (now 1.70))?

wgpu-hal/src/gles/egl.rs
```diff
@@ -683,7 +688,7 @@ enum WindowKind {
 
 #[derive(Clone, Debug)]
 struct WindowSystemInterface {
-    display_owner: Option<Arc<DisplayOwner>>,
+    display_owner: Option<Arc<Mutex<DisplayOwner>>>,
```
cwfitzgerald: Why is this a mutex all of a sudden?

This could use an in-code explanation.

cwfitzgerald (changes requested):

TornaxO7:
> Outside of the one unnecessary cast, how did you get these clippy errors? Were you running with a newer compiler (we by default only use 1.65 (now 1.70))?

that might be the reason! That's my output of `cargo clippy --version`:
```rust
clippy 0.1.72 (d5c2e9c 2023-09-13)
```

wgpu-hal/src/gles/egl.rs
TornaxO7: > Why is this a mutex all of a sudden?

`cargo clippy` gives me the following warning:

<details>

```rust
warning: usage of an `Arc` that is not `Send` or `Sync`
   --> wgpu-hal/src/gles/egl.rs:805:32
    |
805 |                 (display, Some(Arc::new(library)), WindowKind::Wayland)
    |                                ^^^^^^^^^^^^^^^^^
    |
    = note: the trait `Sync` is not implemented for `DisplayOwner`
    = note: required for `Arc<DisplayOwner>` to implement `Send` and `Sync`
    = help: consider using an `Rc` instead or wrapping the inner type with a `Mutex`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#arc_with_non_send_sync
    = note: `#[warn(clippy::arc_with_non_send_sync)]` on by default

warning: usage of an `Arc` that is not `Send` or `Sync`
   --> wgpu-hal/src/gles/egl.rs:816:32
    |
816 |                 (display, Some(Arc::new(display_owner)), WindowKind::X11)
    |                                ^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: the trait `Sync` is not implemented for `DisplayOwner`
    = note: required for `Arc<DisplayOwner>` to implement `Send` and `Sync`
    = help: consider using an `Rc` instead or wrapping the inner type with a `Mutex`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#arc_with_non_send_sync

warning: usage of an `Arc` that is not `Send` or `Sync`
   --> wgpu-hal/src/gles/egl.rs:833:32
    |
833 |                 (display, Some(Arc::new(display_owner)), WindowKind::AngleX11)
    |                                ^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: the trait `Sync` is not implemented for `DisplayOwner`
    = note: required for `Arc<DisplayOwner>` to implement `Send` and `Sync`
    = help: consider using an `Rc` instead or wrapping the inner type with a `Mutex`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#arc_with_non_send_sync
```

</details>

Should I remove it?

> This could use an in-code explanation.

Aye, I can add it.

kpreid: There certainly shouldn't be a `Mutex` that isn't actually being used for interior mutability. Either

* the `Arc` can simply be downgraded to an `Rc` (most likely),
* this is a situation clippy doesn't understand and the lint should be `allow`, or
* there should be a (justified!) `unsafe impl Sync for DisplayOwner` to go along with the `unsafe impl Send`.

Since this is supposed to be a lint fix PR, not add new capabilities, the third option is certainly wrong.

(I'm not familiar with this part of the `wgpu` code, so I don't know which is right; I just saw a situation that could use further explanation.)

wgpu-hal/src/gles/egl.rs
```diff
@@ -135,7 +135,7 @@ impl DisplayRef {
 }
 
 /// DisplayOwner ties the lifetime of the system display handle
-/// to that of the loaded library.
+/// tk that of the loaded library.
```
kpreid: Typo introduced.

wgpu-hal/src/gles/egl.rs
```diff
@@ -145,6 +145,11 @@ struct DisplayOwner {
     display: DisplayRef,
 }
 
+/// # Safety
+/// `library` implements `Send` and `display` is just an enum which gets passed
+/// around.
+unsafe impl Send for DisplayOwner {}
```
kpreid: But is it safe to run the `Drop` code on arbitrary threads? That's what matters here, not just the field contents. In any case, lint fixes shouldn't require adding new impls — see my other comment.

kpreid (changes requested):

wgpu-hal/src/gles/egl.rs
TornaxO7: > the Arc can simply be downgraded to an Rc,

Hm... I assumed that there's an `Arc` in order to be able to let multiple threads use the wgpu-api.

> there should be a (justified!) unsafe impl Sync for DisplayOwner to go along with the unsafe impl Send.

`Sync` shouldn't be possible, if I'm right because two threads could call the `get_ptr` function on `DisplayOwner::X11` and then they can write/read/modify this pointer (through `unsafe` though) without "notifying" the other thread(s), but maybe I'm  wrong here.

wgpu-hal/src/gles/egl.rs
TornaxO7: fixed

wgpu-hal/src/gles/egl.rs
TornaxO7: oh damn. That's a good point! It looks like that [it's not documented](https://www.x.org/releases/X11R7.6/doc/libX11/specs/libX11/libX11.html#Closing_the_Display) what happens, if you use the same pointer and call `XCloseDisplay` on it multiple times. I'll remove the `Send` implementation. Thank you.

wgpu-hal/src/gles/egl.rs
kpreid: > I assumed that there's an `Arc` in order to be able to let multiple threads use the wgpu-api.

`Arc` doesn't *add* any thread-safety. The only thing it does is add multiple-ownership, like `Rc`. The difference is that using `Rc` loses `Send + Sync` if you already had either, but using `Arc` does not. But (in the existing state of the code) you don't have `Send` *or* `Sync`, so there is no benefit to `Arc`. That's what Clippy's trying to tell you: “this `Arc` does you no good”.

wgpu-hal/src/gles/egl.rs
kpreid: “Multiple times” isn't relevant — `Drop` is always executed at most once. The relevant question, to understand whether it's OK to implement `Send`, is whether it's okay for `Drop` — or other operations on the type — to be run *on a different thread* than the object was created on.

wgpu-hal/src/gles/egl.rs
TornaxO7: > > I assumed that there's an `Arc` in order to be able to let multiple threads use the wgpu-api.
> 
> `Arc` doesn't _add_ any thread-safety. The only thing it does is add multiple-ownership, like `Rc`. The difference is that using `Rc` loses `Send + Sync` if you already had either, but using `Arc` does not. But (in the existing state of the code) you don't have `Send` _or_ `Sync`, so there is no benefit to `Arc`. That's what Clippy's trying to tell you: “this `Arc` does you no good”.

oh ok, that sounds reasonable. I replaced `Arc` with `Rc` now.

cwfitzgerald (approved):
Noice

JMS55 (approved):

PR #4207, opened 2023-10-04
Validate that resources belong to the right device.

### LINKED ISSUE #3927: Validate that inputs are created from the right device
**Description**

IDs currently do not contain a way to identify (and therefore validate) the associated device. 

**Repro steps**

See the test added in https://github.com/gfx-rs/wgpu/pull/3926

When using a resource created via the wrong device, the index of the resource ends up pointing to either another resource in the current device or some out of bounds non-existent resource.

As a result, the wrong resource is used or the code panics when trying to access it.

**Platform**

All platforms and configurations.

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Fixes #3927

**Description**

A pretty boring and mechanical change set although fairly important  for the browser use case. Each entry point that takes resources as input must check that these resources are associated with the same device.

I'm sure that I have forgotten some, thankfully it looks like the CTS is rather thorough about testing this so we'll see what's missing when this makes its way into firefox.

**Testing**

I enabled the the `device_msimatch` test which I added when filing the issue. It only covers a tiny portion of the cases but the CTS has good coverage of this so adding exhaustive tests here is not as important.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,7 @@ By @teoxoy in [#4185](https://github.com/gfx-rs/wgpu/pull/4185)
 - Add stub support for device destroy and device validity. By @bradwerth in [4163](https://github.com/gfx-rs/wgpu/pull/4163)
 - Add trace-level logging for most entry points in wgpu-core By @nical in [4183](https://github.com/gfx-rs/wgpu/pull/4183)
 - Add `Rgb10a2Uint` format. By @teoxoy in [4199](https://github.com/gfx-rs/wgpu/pull/4199)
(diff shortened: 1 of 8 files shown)


ErichDonGubler (approved):
These changes LGTM. I'm paranoid and want to make sure we cover _all_ of the API that's necessary, but I guess that need not block. We can keep iterating and covering more cases with follow-up work.

cwfitzgerald (approved):
In the future, I think it would be nice to have a trait like

```trait
trait DeviceChild {
    fn get_device_id(&self) -> ....;
    
    fn is_child_of(&self, device: Id<Device>) -> Result<(), DeviceError> {
        ....
    }
}
```

or something to streamline this a bit and prevent mistakes, but for now this is fine.

PR #4211, opened 2023-10-05
Expose and validate the bgra8unorm-storage feature

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1829895

**Description**

This adds initial support for the `bgra8unorm-storage` feature by only proposing it where supported out of the box. I've seen some discussions about emulating it on some configurations so I'll probably look into that next.

I haven't looked into whether it can work with the GL backend.

**Testing**

I only checked that the feature is supported via wgpu-info on a linux and a windows laptop. I suspect that the CTS has coverage for this but I have not checked.

### DIFF
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -558,6 +558,14 @@ impl<A: HalApi> Device<A> {
             return Err(CreateTextureError::InvalidUsage(desc.usage));
         }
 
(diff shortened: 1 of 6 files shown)


nical:
Oops I just found out about #3634

PR #4212, opened 2023-10-06
Make buffer_map and buffer_unmap check for device validity, add tests.

bradwerth:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**

**Description**
This patch makes buffer_unmap check for a valid device, and corrects buffer_map to return the appropriate error for an invalid device. It also adds device validity checks to device_maintain_ids and to the functions that get and set buffer sub data.

**Testing**
Tests are added in device.rs checking buffer map and unmap for device validity errors.


Wumpf (approved):
thanks for keeping chipping away on this one :)

PR #4219, opened 2023-10-09
Argument Buffers on Metal

### LINKED ISSUE #3334: Use Argument Buffers for `binding_array` on Metal
This is a major change for our binding model on Mac, but would allow us to support bindless in it's true form by letting us bind up to a million textures and buffers on modern chips.

~~Argument buffers are unconditionally supported for our target, so we can just universally change the API usage.~~

This isn't true, it requires Mac 10.13 and WebGPU is minimum 10.12. There are also issues with write resources on tier 1 hardware. I suspect we should just use argument buffer for bindless arrays.

kanerogers:
# Overview
This PR attempts to close #3334 

## Progress
When used in concert with [this naga PR](https://github.com/gfx-rs/naga/pull/2547), significantly simplifies wgpu's Metal backend by, in large part, replacing the existing resource binding model with argument buffers.

In many ways, a large part of the work done by the existing Metal backend is attempting to resolve the impedance mismatch between WebGPU (and, by extension, Vulkan's) descriptor set based binding model and Metal's pre-argument buffer binding model. This required very careful tracking and populating of a function argument table and passing knowledge of those bindings down to naga at shader generation time.

This PR instead tries to simplify that work by instead generating a Metal shader that uses argument buffers and employs a "reflection" style architecture that allows the generated shader to provide a simple 1:1 mapping between the members of a given bind group and the resources used in the shader.

## Further work
While I believe the bulk of the work has been done, I have sadly run out of available time to fully test and resolve all the edge cases required to finish this issue.

In particular, there are, unfortunately, *at least* two wrinkles that prevent the wholesale adoption of argument buffers:

- Writable textures on devices below tier 1
- Buffers with dynamic offsets

These two use cases would require passing the resource by argument into the entry point, as per the existing binding model. It should be reasonably straightforward to check for the presence of these two cases, and any others that arise, and handle them accordingly.

I have also not yet integrated the new binding model with compute shaders as I had hoped to resolve the issues with fragment/vertex shaders first.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1603,7 +1603,6 @@ dependencies = [
 [[package]]
 name = "naga"
 version = "0.13.0"
-source = "git+https://github.com/gfx-rs/naga?rev=cc87b8f9eb30bb55d0735b89d3df3e099e1a6e7c#cc87b8f9eb30bb55d0735b89d3df3e099e1a6e7c"
 dependencies = [
  "bit-set",
  "bitflags 2.4.0",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ default-members = [
     "wgpu-hal",
     "wgpu-info",
     "wgpu-types",
-    "tests"
+    "tests",
 ]
 
 [workspace.package]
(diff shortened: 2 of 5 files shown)


cwfitzgerald:
As this PR is likely not going to be merged in this form, and the design for bindless will be different in the future, I'm going to close this.

Really appreciate all the effort you put into this, however :)

PR #4220, opened 2023-10-09
Handle user callbacks in surface_configure

### LINKED ISSUE #4214: `Surface::configure` can panic with "map callback was leaked"
**Description**
When `Surface::configure` is called, it's possible for the panic at:
https://github.com/gfx-rs/wgpu/blob/c85cbea39ef9133f52625b473bfffcced6b44a1e/wgpu-core/src/resource.rs#L204-L210
to trigger, specifically when:
https://github.com/gfx-rs/wgpu/blob/c85cbea39ef9133f52625b473bfffcced6b44a1e/wgpu-core/src/device/global.rs#L2343-L2345
returned a non-empty `Ok`

**Repro steps**
Running [vello](https://github.com/linebender/vello)'s examples (since https://github.com/linebender/vello/pull/350, at the time of writing https://github.com/linebender/vello/commit/34483eb787c5e3d43a940ac999b0ba98653f6a0e).
Pressing `v` to toggle vsync (and force a reconfigure) may also be required, although sometimes for reasons unknown this isn't needed

I expected for a panic not to occur

I observed that a panic occured

**Platform**
Wayland Gnome on Fedora 38 on an intel i5-1240p
Reproduction also confirmed on ["on Sway and with an amdgpu"](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Map.20Callbacks.20Leak/near/395311371)

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Fixes #4214.

**Description**

We have assertions in place to check that we never forget to fire a map callback (otherwise that can cause memory leaks for some users of the API). `surface_configure` was not handling the user callbacks returned by maintain and this PR adds that.

**Testing**

:cricket:

### DIFF
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -2256,15 +2256,19 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             Ok(())
         }
 
-        log::info!("configuring surface with {:?}", config);
-        let hub = A::hub(self);
-        let mut token = Token::root();
+        // User callbacks must not be called while we are holding locks.
+        let mut user_callbacks = None;
 
-        let (mut surface_guard, mut token) = self.surfaces.write(&mut token);
-        let (adapter_guard, mut token) = hub.adapters.read(&mut token);
-        let (device_guard, mut token) = hub.devices.read(&mut token);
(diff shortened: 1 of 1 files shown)


cwfitzgerald (approved):
LGTM, thanks for handling this!

PR #4227, opened 2023-10-10
Properly handle user callbacks in surface_configure

### LINKED ISSUE #4214: `Surface::configure` can panic with "map callback was leaked"

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

I did a sloppy job of fixing this in #4220.
Fixes #4214.

**Description**

Fix the fix. The initial patch was handling user callbacks only in some cases, missing that there was a `return None;` at the end of the block. This nests most of the loop into an additional block to reduce the scope of the locks while still handling the user callbacks in the loop as the portion outside of the loop is only taken when an error occurs.


dhardy:
Works for me :+1:

PR #4230, opened 2023-10-11
Expose Instance flags

### LINKED ISSUE #2316: Labels, debug markers, and validation layers should be available on release builds

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Blocks #4028
Closes https://github.com/gfx-rs/wgpu/issues/2316

**Description**

- Move instance flags from wgpu-hal to wgpu-types
- expose them in the instance descriptor
- Add a few helpers for picking the flags:
  - `InstanceFlags::debugging()` picks all flags related to debugging and validation,
  - `InstanceFlags::from_build_type()` picks debugging flags if `debug_assertions` is enabled (matches current behavior)
  - DEBUG and VALIDATION flags can be overridden via environment variables using `InstanceFlags::with_env`.

### DIFF
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -73,14 +73,9 @@ impl Instance {
     pub fn new(name: &str, instance_desc: wgt::InstanceDescriptor) -> Self {
         fn init<A: HalApi>(_: A, instance_desc: &wgt::InstanceDescriptor) -> Option<A::Instance> {
             if instance_desc.backends.contains(A::VARIANT.into()) {
-                let mut flags = wgt::InstanceFlags::empty();
-                if cfg!(debug_assertions) {
-                    flags |= wgt::InstanceFlags::VALIDATION;
-                    flags |= wgt::InstanceFlags::DEBUG;
-                }
                 let hal_desc = hal::InstanceDescriptor {
                     name: "wgpu",
-                    flags,
+                    flags: instance_desc.flags,
                     dx12_shader_compiler: instance_desc.dx12_shader_compiler.clone(),
                     gles_minor_version: instance_desc.gles_minor_version,
                 };
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -73,10 +73,10 @@ impl Instance {
     pub fn new(name: &str, instance_desc: wgt::InstanceDescriptor) -> Self {
         fn init<A: HalApi>(_: A, instance_desc: &wgt::InstanceDescriptor) -> Option<A::Instance> {
             if instance_desc.backends.contains(A::VARIANT.into()) {
-                let mut flags = hal::InstanceFlags::empty();
+                let mut flags = wgt::InstanceFlags::empty();
                 if cfg!(debug_assertions) {
-                    flags |= hal::InstanceFlags::VALIDATION;
-                    flags |= hal::InstanceFlags::DEBUG;
+                    flags |= wgt::InstanceFlags::VALIDATION;
+                    flags |= wgt::InstanceFlags::DEBUG;
                 }
                 let hal_desc = hal::InstanceDescriptor {
                     name: "wgpu",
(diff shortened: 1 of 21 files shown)


wgpu-types/src/lib.rs
```diff
@@ -6520,6 +6540,7 @@ impl Default for InstanceDescriptor {
     fn default() -> Self {
         Self {
             backends: Backends::all(),
+            flags: InstanceFlags::default(),
```
teoxoy, line 6581: I feel like we should use `from_build_type` here to keep the current behavior.

wgpu-core/src/instance.rs
```diff
@@ -71,8 +71,24 @@ pub struct Instance {
...
         fn init<A: HalApi>(_: A, instance_desc: &wgt::InstanceDescriptor) -> Option<A::Instance> {
             if instance_desc.backends.contains(A::VARIANT.into()) {
+                let mut flags = instance_desc.flags;
+                if let Some(bit) = env("WGPU_VALIDATION") {
+                    flags.set(wgt::InstanceFlags::VALIDATION, bit);
+                }
+                if let Some(bit) = env("WGPU_DEBUG") {
+                    flags.set(wgt::InstanceFlags::DEBUG, bit);
+                }
+
```
teoxoy: I don't think this is a good idea, wgpu is a library and we shouldn't become sensitive to env variables.
We can however add an util function similar to the ones we already have that read and set these based on env variables.

teoxoy (changes requested):

wgpu-core/src/instance.rs
nical: I was hesitating between this and adding `InstanceFlags::from_env()`, I'll switch over to that.

wgpu-types/src/lib.rs
nical: Having default return a different thing depending on the build configuration sort of rubs me the wrong way but I don't feel very strongly about it so I changed `InstanceFlag`'s default impl to do `from_build_config` instead.

teoxoy: > Having default return a different thing depending on the build configuration sort of rubs me the wrong way but I don't feel very strongly about it

That's a good point. I was mainly worried that we won't get debug info and validation errors by default in debug builds (which was the default behavior before).

Not sure what's best here since now wgpu users will have to pass the flags.

teoxoy (approved):

wgpu-types/src/lib.rs
```diff
@@ -818,6 +818,76 @@ impl Features {
...
+        InstanceFlags::empty()
+    }
+
+    /// Returns this set of flags, affected by environment variables.
+    ///
+    /// The presence of an environment variable implies that the corresponding flag should be set
+    /// unless the value is "0" in which case the flag is unset. If the environment variable is
+    /// not present, then the flag is unaffected.
+    ///
+    /// For example `let flags = InstanceFlags::debuggin().with_env();` with `WGPU_VALIDATION=0`
```
```suggestion
    /// For example `let flags = InstanceFlags::debugging().with_env();` with `WGPU_VALIDATION=0`
```

wgpu-types/src/lib.rs
cwfitzgerald: DIscussed in call - keeping behavior the same.

cwfitzgerald (approved):
Amazing!

PR #4237, opened 2023-10-13
GLES refactors

teoxoy:
Commits are self-contained and can land without squashing.


cwfitzgerald (approved):
All very nice cleanups, thanks!

PR #4238, opened 2023-10-13
Use DeviceError::Lost to represent device loss.

bradwerth:
The device.valid boolean, regardless of its name, represents a state that is only activated through "lose the device". This patch changes error reporting of an invalid/lost device to use DeviceError::Lost. The value DeviceError::Invalid is used when a device id is not found.

**Checklist**

- [X] Run `cargo clippy`.

**Connections**
None

**Description**
This separates the meaning of DeviceError::Invalid (unknown device id) from DeviceError::Lost (device known and lost).

**Testing**
There is no real testing of this. The test `device_destroy_then_more` is making calls that result in a `DeviceError::Lost` value, but the test infrastructure checks for *any* failure, not a specific error type.


cwfitzgerald (approved):
Noice

PR #4246, opened 2023-10-16
Make it possible to filter labels out ahead of wgpu-hal

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Description**

This is intended as a hardening feature. If we don't trust the author of labels (for example web content) and don't trust drivers to handle long or non-ascii strings, an instance flags can be set to filter out hal labels. When the flag is set, wgpu-core can still hold on to labels but they are not passed down to GPU drivers.

Most people should not enable this flag. Labels are pretty useful for debugging with renderdoc and similar tools.

### DIFF
--- a/wgpu-core/src/command/compute.rs
+++ b/wgpu-core/src/command/compute.rs
@@ -776,11 +776,11 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 }
                 ComputeCommand::PushDebugGroup { color: _, len } => {
                     state.debug_scope_depth += 1;
+                    let label =
+                        str::from_utf8(&base.string_data[string_offset..string_offset + len])
+                            .unwrap();
+                    string_offset += len;
                     if !discard_hal_labels {
-                        let label =
-                            str::from_utf8(&base.string_data[string_offset..string_offset + len])
-                                .unwrap();
-                        string_offset += len;
                         unsafe {
                             raw.begin_debug_marker(label);
                         }
@@ -801,11 +801,11 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                     }
                 }
                 ComputeCommand::InsertDebugMarker { color: _, len } => {
+                    let label =
+                        str::from_utf8(&base.string_data[string_offset..string_offset + len])
+                            .unwrap();
+                    string_offset += len;
                     if !discard_hal_labels {
-                        let label =
-                            str::from_utf8(&base.string_data[string_offset..string_offset + len])
-                                .unwrap();
-                        string_offset += len;
                         unsafe { raw.insert_debug_marker(label) }
                     }
                 }
--- a/wgpu-core/src/command/compute.rs
+++ b/wgpu-core/src/command/compute.rs
@@ -14,6 +14,7 @@ use crate::{
     error::{ErrorFormatter, PrettyError},
     global::Global,
     hal_api::HalApi,
+    hal_label,
     hub::Token,
     id,
     id::DeviceId,
@@ -476,8 +477,12 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
             Some(&*query_set_guard),
         );
 
+        let discard_hal_labels = self
+            .instance
+            .flags
+            .contains(wgt::InstanceFlags::DISCARD_HAL_LABELS);
         let hal_desc = hal::ComputePassDescriptor {
-            label: base.label,
+            label: hal_label(base.label, self.instance.flags),
             timestamp_writes,
         };
 
@@ -771,12 +776,14 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 }
                 ComputeCommand::PushDebugGroup { color: _, len } => {
                     state.debug_scope_depth += 1;
-                    let label =
-                        str::from_utf8(&base.string_data[string_offset..string_offset + len])
-                            .unwrap();
-                    string_offset += len;
-                    unsafe {
-                        raw.begin_debug_marker(label);
+                    if !discard_hal_labels {
+                        let label =
+                            str::from_utf8(&base.string_data[string_offset..string_offset + len])
+                                .unwrap();
+                        string_offset += len;
+                        unsafe {
+                            raw.begin_debug_marker(label);
+                        }
                     }
                 }
                 ComputeCommand::PopDebugGroup => {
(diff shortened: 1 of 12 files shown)


nical:
There are still a couple of internal labels that are missed (because the flag was not easily reachable). They are less important for hardening since they short and ascii-only, but we can look into filtering them as well for consistency.

wgpu-core/src/command/compute.rs
```diff
@@ -476,8 +476,18 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
 
+        let label = if self
+            .instance
+            .flags
+            .contains(wgt::InstanceFlags::DISCARD_HAL_LABELS)
+        {
+            None
+        } else {
+            base.label
+        };
```
jimblandy: Could we use `bool::then_some` here, and elsewhere?

jimblandy (changes requested):
I believe this PR is missing fixes for the following labels:
- [x] `wgpu_hal::RenderPassDescriptor::label`, passed in `wgpu_core::render::RenderPassInfo::start`
- [x] `wgpu_types::QuerySetDescriptor::label`, passed in `wgpu_core::Device::create_query_set` - although this is not currently used by Firefox (`GPUDevice.createQuerySet` is commented out in the WebIDL), we probably should address it now
- [x] `wgpu_types::RenderBundleDescriptor::label`, which `RenderBundle::execute` passes to `wgpu_hal::CommandEncoder::begin_debug_marker`
- [x] It seems like `wgpu_hal::CommandEncoder::begin_encoding` gets passed `wgpu_core::command::CommandEncoder::label`, which it seems can be set from `wgpu_types::CommandEncoderDescriptor::label`.
- [x] `Global::command_encoder_run_render_pass_impl` takes a `BasePassRef` as an argument whose `label` field is passed to `wgpu_core::command::CommandEncoder::open_pass`, which passes it to `wgpu_hal::CommandEncoder::begin_encoding`.
- [x] `wgpu_hal::CommandEncoder::insert_debug_marker` seems to take labels from compute pass encoders, and is also called from `Gloal::command_encoder_insert_debug_marker`. I don't know why we have two ways of doing this, but they both seem reachable from Firefox.
- [x] It seems like `wgpu_hal::CommandEncoder::begin_debug_marker` is also reachable.

jimblandy:
I marked this as "changes requested", but the PR as it stands is a step in the right direction. If you'd like to address the cases I found in a follow-up, that's fine with me.

wgpu-types/src/lib.rs
```diff
@@ -841,6 +841,8 @@ bitflags::bitflags! {
         const DEBUG = 1 << 0;
         /// Enable validation, if possible.
         const VALIDATION = 1 << 1;
+        /// Don't pass labels to wgpu-hal.
+        const DISCARD_HAL_LABELS = 1 << 2;
```
cwfitzgerald, line 845: I feel like this should be a positive flag that is on in the debugging configuration and debug mode (by default)

wgpu-core/src/command/compute.rs
nical: The closest thing I could do to what you are looking for is:

```rust
let label = (!self
    .instance
    .flags
    .contains(wgt::InstanceFlags::DISCARD_HAL_LABELS))
.then_some(base.label)
.flatten();
```

Which I find confusing in more ways than one (cargo fmt does a number on it). I'll add a helper function instead.

wgpu-types/src/lib.rs
nical: My expectation is that the majority of people won't use this flag (if perf is what you care about you'll want to filter out label before passing them to wgpu-core). It's formulated this way so that `InstanceFlags::empty()` remains a good default and people opt into filtering labels by adding the bit.

wgpu-core/src/command/compute.rs
jimblandy: Oh, sure, that's not an improvement. How about:

```
let label = base.label.as_ref().filter(|_| !self.instance.flags.contains(wgt::InstanceFlags::DISCARD_HAL_LABELS));
```

JimB's Law: `Option` always has the right method - you just didn't recognize it.

But this is a very minor point, not worth delaying the PR for.

wgpu-types/src/lib.rs
jimblandy: I see @cwfitzgerald's point - positive flags are generally less confusing that negative flags - but I also see Nical's point that all the extant flags are non-default features.

In the interest of progress, I'm going to mark this conversation as resolved. @cwfitzgerald, if you feel strongly about it, I'll write the PR to flip it.

cwfitzgerald: Don't feel particularly strongly, lets land as is.

wgpu-core/src/command/compute.rs
```diff
@@ -771,12 +776,14 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
-                        raw.begin_debug_marker(label);
+                    if !discard_hal_labels {
+                        let label =
+                            str::from_utf8(&base.string_data[string_offset..string_offset + len])
+                                .unwrap();
+                        string_offset += len;
+                        unsafe {
+                            raw.begin_debug_marker(label);
+                        }
                     }
```
jimblandy: The update to `string_offset` needs to be unconditional, unless you're going to make putting the labels in the `string_data` buffer conditional in the first place.

wgpu-core/src/command/render.rs
```diff
@@ -2155,14 +2163,17 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
-                        log::trace!("RenderPass::push_debug_group {label:?}");
-                        string_offset += len;
-                        unsafe {
-                            raw.begin_debug_marker(label);
+                            log::trace!("RenderPass::push_debug_group {label:?}");
+                            string_offset += len;
+                            unsafe {
+                                raw.begin_debug_marker(label);
+                            }
                         }
```
jimblandy: Similarly

jimblandy (approved):
I came across a few missed labels, and some problems with string processing in render passes and compute passes. I've added a fixup commit; please look it over.

wgpu-core/src/command/compute.rs
jimblandy: This is addressed in the fixup commit; please look it over and resolve the conversation if you think my fix is okay.

wgpu-core/src/command/render.rs
jimblandy: This is addressed in the fixup commit; please look it over and resolve the conversation if you think my fix is okay.

wgpu-core/src/command/compute.rs
nical: I pushed another commit that updates string_offset unconditionally but keeps the validation of utf8 strings in the branch.

jimblandy: > validation of utf8 strings

Oh, good point!

PR #4248, opened 2023-10-16
Add WGL context creation and OpenGL support

Zoxc:
This adds WGL context creation and desktop OpenGL support to the `gles` backend. It's used instead of `EGL` on Windows. Only WGL requests desktop OpenGL contexts. The behavior of `EGL` is left unchanged.

To enable construction of an OpenGL context without a surface, a memory device is used. Each surface gets their own OpenGL context which share object with the memory device OpenGL context. Rendering to surfaces happens indirectly via renderbuffers as framebuffers cannot be shared across OpenGL contexts.

Currently a lot of `gles` backend features are not enabled on desktop OpenGL.

I've tested it with an HD 5870 using the examples and [vello](https://github.com/linebender/vello).

`bunnymark` is plain blue. `skybox` crashes in the driver. `mipmap` is too white, Vsync locks to 30 fps instead of 60 fps on a 60 Hz display. These may be issues with the driver or `gles` backend rather than this PR however.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,8 +76,7 @@ We have a [wiki](https://github.com/gfx-rs/wgpu/wiki) that serves as a knowledge
 | Metal       |                                |                    | :white_check_mark:        |                           |
 | DX12        | :white_check_mark: (W10+ only) |                    |                           |                           |
 | DX11        | :hammer_and_wrench:            |                    |                           |                           |
-| GLES3       |                                | :ok:               | :ok: (angle; macOS only)  | :ok: (WebGL2 Only)        |
-| OpenGL 3.3+ | :ok:                           |                    |                           |                           |
+| OpenGL      |  :ok: (Desktop GL 3.3+)        | :ok: (GL ES 3.0+)  | :ok: (angle; GL ES 3.0+)  | :ok: (WebGL2)             |
 | WebGPU      |                                |                    |                           | :white_check_mark:        |
 
 :white_check_mark: = First Class Support — :ok: = Best Effort Support — :hammer_and_wrench: = Unsupported, but support in progress
--- a/README.md
+++ b/README.md
@@ -70,14 +70,15 @@ We have a [wiki](https://github.com/gfx-rs/wgpu/wiki) that serves as a knowledge
 
 ## Supported Platforms
 
-| API       | Windows                        | Linux & Android    | macOS & iOS               | Web (wasm)                |
-| --------- | ------------------------------ | ------------------ | ------------------------- | ------------------------- |
-| Vulkan    | :white_check_mark:             | :white_check_mark: | :ok: (vulkan-portability) |                           |
-| Metal     |                                |                    | :white_check_mark:        |                           |
-| DX12      | :white_check_mark: (W10+ only) |                    |                           |                           |
-| DX11      | :hammer_and_wrench:            |                    |                           |                           |
-| GLES3     | :ok: (angle)                   | :ok:               | :ok: (angle; macOS only)  | :ok: (WebGL2 Only)        |
-| WebGPU    |                                |                    |                           | :white_check_mark:        |
+| API         | Windows                        | Linux & Android    | macOS & iOS               | Web (wasm)                |
+| ----------- | ------------------------------ | ------------------ | ------------------------- | ------------------------- |
+| Vulkan      | :white_check_mark:             | :white_check_mark: | :ok: (vulkan-portability) |                           |
+| Metal       |                                |                    | :white_check_mark:        |                           |
+| DX12        | :white_check_mark: (W10+ only) |                    |                           |                           |
+| DX11        | :hammer_and_wrench:            |                    |                           |                           |
+| GLES3       |                                | :ok:               | :ok: (angle; macOS only)  | :ok: (WebGL2 Only)        |
+| OpenGL 3.3+ | :ok:                           |                    |                           |                           |
+| WebGPU      |                                |                    |                           | :white_check_mark:        |
 
 :white_check_mark: = First Class Support — :ok: = Best Effort Support — :hammer_and_wrench: = Unsupported, but support in progress
 
@@ -148,6 +149,7 @@ We have multiple methods of testing, each of which tests different qualities abo
 | DX11/Windows 10  | :construction:     | —                  | using WARP                            |
 | Metal/MacOS      | :heavy_check_mark: | —                  | using hardware runner                 |
 | Vulkan/Linux     | :heavy_check_mark: | -                  | using swiftshader                     |
+| GL/Windows       |                    | —                  |                                       |
 | GLES/Linux       | :heavy_check_mark: | —                  | using llvmpipe                        |
 | WebGL/Chrome     | :heavy_check_mark: | —                  | using swiftshader                     |
 
--- a/wgpu-hal/src/gles/adapter.rs
+++ b/wgpu-hal/src/gles/adapter.rs
@@ -244,9 +244,9 @@ impl super::Adapter {
         }
 
         if let Some(full_ver) = full_ver {
-            if full_ver < (3, 2) {
+            if full_ver < (3, 3) {
                 log::warn!(
-                    "Returned GL context is {}.{}, when 3.2+ was requested",
+                    "Returned GL context is {}.{}, when 3.3+ is needed",
                     full_ver.0,
                     full_ver.1
                 );
@@ -722,6 +722,7 @@ impl super::Adapter {
                     max_texture_size,
                     next_shader_id: Default::default(),
                     program_cache: Default::default(),
+                    es: es_ver.is_some(),
                 }),
             },
             info: Self::make_info(vendor, renderer),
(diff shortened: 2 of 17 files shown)


Zoxc:
I added some more feature detection and driver workarounds. That makes pass tests for [vger-rs](https://github.com/audulus/vger-rs).

teoxoy:
Is it possible to test this in CI?

cwfitzgerald:
@teoxoy not yet. Unless we manage to set up gl-on-12 over warp, this will need to wait until we have a physical box with a real gpu running our windows tests.

cwfitzgerald:
I am very happy to be wrong! GLon12 isn't terribly hard to set up, though definitely shouldn't be part of this PR. Will file a follow up with more information.

Zoxc:
I've changed the PR to use a dummy window instead of a memory device, as that makes OpenGL context creation work on Mesa and my Sandy Bridge laptop.

cwfitzgerald:
Pushed a commit that enables renderdoc with wgl.

Outside of presentation being all black on nvidia, this is looking good.

cwfitzgerald:
Just tried this on an AMD machine, and it indeed works correctly, so this really seems like an nv-specific issue.

wgpu-hal/src/gles/adapter.rs
```diff
@@ -212,37 +221,95 @@ impl super::Adapter {
...
+                .map(|full_ver| full_ver >= (req_full_major, req_full_minor))
+                .unwrap_or_default();
+
+            es_supported || full_supported
+        };
+
+        let supports_storage =
+            supported((3, 1), (4, 3)) || extensions.contains("GL_ARB_shader_storage_buffer_object");
+        let supports_compute =
+            supported((3, 1), (4, 6)) || extensions.contains("GL_ARB_compute_shader");
```
cwfitzgerald: Isn't compute `(4, 3)` as well?

wgpu-hal/src/gles/adapter.rs
```diff
@@ -369,9 +439,14 @@ impl super::Adapter {
         );
         features.set(
             wgt::Features::SHADER_PRIMITIVE_INDEX,
-            ver >= (3, 2) || extensions.contains("OES_geometry_shader"),
+            supported((3, 2), (4, 6))
```
cwfitzgerald: 4.6 seems really high for this, gl wiki says 3.2 for geometry shaders

wgpu-hal/src/gles/adapter.rs
```diff
@@ -369,9 +439,14 @@ impl super::Adapter {
...
+                || extensions.contains("OES_geometry_shader")
+                || extensions.contains("GL_ARB_geometry_shader4"),
+        );
+        features.set(
+            wgt::Features::SHADER_EARLY_DEPTH_TEST,
+            supported((3, 1), (4, 6)) || extensions.contains("GL_ARB_shader_image_load_store"),
```
cwfitzgerald: This likely should be 4.2

wgpu-hal/src/gles/adapter.rs
```diff
@@ -521,7 +599,9 @@ impl super::Adapter {
             max_uniform_buffer_binding_size: unsafe {
                 gl.get_parameter_i32(glow::MAX_UNIFORM_BLOCK_SIZE)
             } as u32,
-            max_storage_buffer_binding_size: if ver >= (3, 1) {
+            max_storage_buffer_binding_size: if supported((3, 1), (4, 3))
+                || extensions.contains("GL_ARB_shader_storage_buffer_object")
```
cwfitzgerald: Couldn't this be `if supports_storage`

wgpu-hal/src/gles/adapter.rs
```diff
@@ -909,7 +1061,11 @@ impl crate::Adapter<super::Api> for super::Adapter {
 
             Some(crate::SurfaceCapabilities {
                 formats,
-                present_modes: vec![wgt::PresentMode::Fifo], //TODO
+                present_modes: if cfg!(windows) {
+                    vec![wgt::PresentMode::Fifo, wgt::PresentMode::Mailbox]
```
cwfitzgerald, line 1063: If you turn off vsync, do you get tearing? If so, this should be Fifo + Immediate; not mailbox.

```suggestion
                    vec![wgt::PresentMode::Fifo, wgt::PresentMode::Mailbox]
```

wgpu-hal/src/gles/wgl.rs
```diff
@@ -0,0 +1,750 @@
...
+    let name = CString::from_vec_with_nul(name.into_bytes()).unwrap();
+
+    unsafe extern "system" fn wnd_proc(
+        window: HWND,
+        msg: UINT,
+        wparam: WPARAM,
+        lparam: LPARAM,
+    ) -> LRESULT {
+        unsafe { DefWindowProcA(window, msg, wparam, lparam) }
+    }
```
cwfitzgerald, line 247: Why can't we pass `DefWindowProcA` directly to `WNDCLASSEXA`? We probably should have a comment describing why.

wgpu-hal/src/gles/wgl.rs
```diff
@@ -0,0 +1,750 @@
...
+            0,
+            0,
+            1,
+            1,
+            ptr::null_mut(),
+            ptr::null_mut(),
+            instance,
+            ptr::null_mut(),
+        )
+    };
```
cwfitzgerald, line 289: Could you add a comment why this ends up creating a hidden window?

wgpu-hal/src/gles/wgl.rs
```diff
@@ -0,0 +1,750 @@
...
+    unsafe impl Sync for SendDc {}
+    unsafe impl Send for SendDc {}
+
+    static GLOBAL: Mutex<Option<Result<SendDc, crate::InstanceError>>> = Mutex::new(None);
+    let mut guard = GLOBAL.lock();
+    if guard.is_none() {
+        *guard = Some(create_global_device_context().map(SendDc));
+    }
+    guard.clone().unwrap().map(|dc| dc.0)
+}
```
cwfitzgerald: Isn't this basically just a `once_cell::Lazy<Result<...>>`?

wgpu-hal/src/gles/wgl.rs
```diff
@@ -0,0 +1,750 @@
...
+                        CONTEXT_PROFILE_MASK_ARB as c_int,
+                        CONTEXT_CORE_PROFILE_BIT_ARB as c_int,
+                        CONTEXT_FLAGS_ARB as c_int,
+                        if desc.flags.contains(InstanceFlags::DEBUG) {
+                            CONTEXT_DEBUG_BIT_ARB as c_int
+                        } else {
+                            0
+                        },
+                        0, // End of list
+                    ]
```
cwfitzgerald: I know this is correct, but this is super spooky from a lifetime perspective, could you lift this array out to a binding to make it very clear it lives long enough.

wgpu-hal/src/gles/wgl.rs
```diff
@@ -0,0 +1,750 @@
...
+    /// Extent because the window lies
+    extent: wgt::Extent3d,
+    format: wgt::TextureFormat,
+    format_desc: super::TextureFormatDesc,
+    #[allow(unused)]
+    sample_type: wgt::TextureSampleType,
+}
+
+pub struct Surface {
+    // FIXME: Find a way to call ReleaseDC on the same thread as GetDC.
```
cwfitzgerald: Maybe when we drop the surface, clear it if it's being dropped from the same thread, otherwise log.

wgpu-hal/src/gles/wgl.rs
```diff
@@ -0,0 +1,750 @@
...
+        device: &super::Device,
+        config: &crate::SurfaceConfiguration,
+    ) -> Result<(), crate::SurfaceError> {
+        let format_desc = device.shared.describe_texture_format(config.format);
+        let inner = &device.shared.context.inner.lock();
+
+        if let Err(e) = inner.context.make_current() {
+            log::error!("unable to make the shared OpengL context current: {e}",);
+            return Err(crate::SurfaceError::Other(
+                "unable to make the shared OpengL context current",
```
cwfitzgerald:
```suggestion
            log::error!("unable to make the shared OpenGL context current: {e}",);
            return Err(crate::SurfaceError::Other(
                "unable to make the shared OpenGL context current",
```

wgpu-hal/src/gles/wgl.rs
```diff
@@ -0,0 +1,750 @@
...
+        let extentions = extensions(&extra, self.dc);
+        if !(extentions.contains("WGL_EXT_swap_control") && extra.SwapIntervalEXT.is_loaded()) {
+            log::error!("WGL_EXT_swap_control is unsupported");
+            return Err(crate::SurfaceError::Other(
+                "WGL_EXT_swap_control is unsupported",
+            ));
+        }
+
+        let vsync = match config.present_mode {
+            wgt::PresentMode::Mailbox => false,
```
cwfitzgerald, line 663: See note about mailbox

wgpu/tests/multi-instance.rs
```diff
@@ -0,0 +1,33 @@
...
+
+        // Sequential instances.
+        for _ in 0..3 {
+            pollster::block_on(get());
+        }
+
+        // Concurrent instances
+        let _instances: Vec<_> = (0..3).map(|_| pollster::block_on(get())).collect();
+    }
+}
```
cwfitzgerald: This should go in `tests/` not `wgpu/tests`

cwfitzgerald (changes requested):
This is super amazing stuff. Tested successfully on all of my devices!

Everything is fairly minor

teoxoy:
Could you also add an "OpenGL" row to the "Supported Platforms" table in the readme (and remove the GLES3 Angle entry for windows)?

wgpu-hal/src/gles/adapter.rs
Zoxc: I interpreted the version numbers from the OpenGL Wiki wrong.

wgpu-hal/src/gles/adapter.rs
Zoxc: If you turn off the compositor on older version of Windows, you might manage some tearing.

wgpu-hal/src/gles/wgl.rs
Zoxc: You can't with `windows-sys` as it doesn't use `extern "system"`.

wgpu-hal/src/gles/wgl.rs
Zoxc: Probably. I didn't want to add a dependency.

wgpu-hal/src/gles/adapter.rs
cwfitzgerald: Ah, then lets leave it as mailbox.

wgpu-hal/src/gles/wgl.rs
cwfitzgerald: `once_cell` is fine to add due to its ubiquity

wgpu-hal/src/gles/wgl.rs
cwfitzgerald: Ah gottcha, could you just make a note of that

wgpu-hal/src/gles/wgl.rs
```diff
@@ -0,0 +1,748 @@
...
+        Ok(())
+    }
+
+    pub fn supports_srgb(&self) -> bool {
+        self.srgb_capable
+    }
+}
+
+impl crate::Surface<super::Api> for Surface {
+    unsafe fn configure(
```
cwfitzgerald, line 567: Missed a thing here:

This needs to be documented in the wgpu-hal api, but `configure` is expected to clean up after the previous configuration if there was one. `unconfigure` is only called when the surface is destroyed.

wgpu-hal/src/gles/wgl.rs
Zoxc: I found a way to use it on-demand on the same thread.

README.md
```diff
@@ -70,14 +70,15 @@ We have a [wiki](https://github.com/gfx-rs/wgpu/wiki) that serves as a knowledge
...
-| WebGPU    |                                |                    |                           | :white_check_mark:        |
+| API         | Windows                        | Linux & Android    | macOS & iOS               | Web (wasm)                |
+| ----------- | ------------------------------ | ------------------ | ------------------------- | ------------------------- |
+| Vulkan      | :white_check_mark:             | :white_check_mark: | :ok: (vulkan-portability) |                           |
+| Metal       |                                |                    | :white_check_mark:        |                           |
+| DX12        | :white_check_mark: (W10+ only) |                    |                           |                           |
+| DX11        | :hammer_and_wrench:            |                    |                           |                           |
+| GLES3       |                                | :ok:               | :ok: (angle; macOS only)  | :ok: (WebGL2 Only)        |
+| OpenGL 3.3+ | :ok:                           |                    |                           |                           |
```
cwfitzgerald:
```suggestion
| OpenGL      |  :ok: (Desktop GL 3.3+)        | :ok: (GL ES 3.0+)  | :ok: (angle; GL ES 3.0+)  | :ok: (WebGL2)             |
```

cwfitzgerald (approved):
Stunning work! Thank you for getting this all through!

PR #4250, opened 2023-10-17
wgpu-hal: vulkan: ensure entries to PhysicalDeviceCapabilities use effective api version

i509VCB:
This was the cause of an error while testing wgpu-hal on agxv. Specifically I was getting an error about running out of memory while creating the "zero init buffer". This was occuring because PhysicalDeviceCapabilities was reading the instance api version and not the device version. But furthermore this is wrong because the actual api version that can be used is the lower of the instance's api version and device's api version.

agxv does not implement maintenance3 yet, so what was being returned for a max memory allocation size was 0, because ash initialized the value to 0. Then when the gpu allocator is created with a maximum allocation size of 0, all gpu memory allocations fail.

**Checklist**

- [x] Run `cargo clippy`.

Although when running clippy this error shows up. Definitely not a fault of my changes:
```
error: incorrect implementation of `partial_cmp` on an `Ord` type
   --> wgpu-core/src/id.rs:155:1
    |
155 | /  impl<T> PartialOrd for Id<T> {
156 | |      fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
    | | _____________________________________________________________-
157 | ||         self.0.partial_cmp(&other.0)
158 | ||     }
    | ||_____- help: change this to: `{ Some(self.cmp(other)) }`
159 | |  }
    | |__^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#incorrect_partial_ord_impl_on_ord_type
    = note: `#[deny(clippy::incorrect_partial_ord_impl_on_ord_type)]` on by default
    ```

### DIFF
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -788,17 +788,25 @@ impl super::InstanceShared {
             let mut capabilities = PhysicalDeviceCapabilities::default();
             capabilities.supported_extensions =
                 unsafe { self.raw.enumerate_device_extension_properties(phd).unwrap() };
+
+            // Set the effective api version before we test for extension support
+            capabilities.effective_api_version = self
+                .driver_api_version
+                .min(capabilities.properties.api_version);
+
             capabilities.properties = if let Some(ref get_device_properties) =
                 self.get_physical_device_properties
             {
                 // Get these now to avoid borrowing conflicts later
-                let supports_descriptor_indexing = self.driver_api_version >= vk::API_VERSION_1_2
+                let supports_descriptor_indexing = capabilities.effective_api_version
+                    >= vk::API_VERSION_1_2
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by nical — APPROVED** [review]

File: wgpu-hal/src/vulkan/adapter.rs [hunk 1]
```diff
@@ -788,17 +788,25 @@ impl super::InstanceShared {
             let mut capabilities = PhysicalDeviceCapabilities::default();
             capabilities.supported_extensions =
                 unsafe { self.raw.enumerate_device_extension_properties(phd).unwrap() };
+
+            // Set the effective api version before we test for extension support
+            capabilities.effective_api_version = self
+                .driver_api_version
+                .min(capabilities.properties.api_version);
```
teoxoy, line 795: Moving this up here won't work as `capabilities.properties.api_version` will be 0.

I opened https://github.com/gfx-rs/wgpu/pull/4252 as I also wanted to decouple the instance and device API versions for some time now.

PR #4252, opened 2023-10-17
Vulkan API version refactors

teoxoy:
- with this patch we get a clear distinction between instance and device API versions as described by the spec.
- fixes incorrect usage of instance API version where the device API version was expected (resolves https://github.com/gfx-rs/wgpu/pull/4250)
- changes max application API version to be VK1.3 instead of using `VK_HEADER_VERSION_COMPLETE` which [is not recommended for this usecase](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_HEADER_VERSION_COMPLETE.html)

Relevant links:

https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkEnumerateInstanceVersion.html
https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceProperties.html
https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkApplicationInfo.html

Commits are self-contained and shouldn't need to be squashed.

### DIFF
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -792,13 +792,13 @@ impl super::InstanceShared {
                 self.get_physical_device_properties
             {
                 // Get these now to avoid borrowing conflicts later
-                let supports_descriptor_indexing = self.driver_api_version >= vk::API_VERSION_1_2
+                let supports_descriptor_indexing = self.instance_api_version >= vk::API_VERSION_1_2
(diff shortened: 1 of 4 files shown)


cwfitzgerald (approved):
Very nice! VK init continues to be a total cluster 😆

cwfitzgerald:
Ack, squashed by accident, sorry :)

teoxoy:
np, the PR wasn't terribly big

PR #4253, opened 2023-10-17
Fix locking order in render_pipeline_get_bind_group_layout

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Blocks https://bugzilla.mozilla.org/show_bug.cgi?id=1856371

**Description**

The code path for when IDs are provided externally was creating a new root token while the previous one still existed. The right thing to do is to ensure the other guards are closed before assigning the newly created bind group which this patch does by scoping the them.

Also add the equivalent code for `compute_pipeline_get_bind_group_layout`.

**Testing**

This is covered by the CTS in Firefox.

### DIFF
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -2084,26 +2084,51 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
     ) {
         let hub = A::hub(self);
         let mut token = Token::root();
-        let (pipeline_layout_guard, mut token) = hub.pipeline_layouts.read(&mut token);
 
         let error = loop {
-            let (bgl_guard, mut token) = hub.bind_group_layouts.read(&mut token);
-            let (_, mut token) = hub.bind_groups.read(&mut token);
-            let (pipeline_guard, _) = hub.compute_pipelines.read(&mut token);
+            let device_id;
+            let id;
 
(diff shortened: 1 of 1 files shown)


jimblandy:
@nical There are [checks](https://github.com/gfx-rs/wgpu/blob/a842019b348618e4b67e87710fb143b58cf64511/wgpu-core/src/hub.rs#L339-L342) in `Token::root` that are supposed to detect attempts to create multiple roots simultaneously on the same thread. Did those debug assertions trigger in this case?

cwfitzgerald:
(These checks probably should be bounded by "strict-asserts" rather than debug_assertions)

jimblandy (approved):
Looks good to me.

jimblandy:
> (These checks probably should be bounded by "strict-asserts" rather than debug_assertions)

Oooh, good point.

jimblandy:
> (These checks probably should be bounded by "strict-asserts" rather than debug_assertions)

Filed as #4258.

nical:
> @nical There are [checks](https://github.com/gfx-rs/wgpu/blob/a842019b348618e4b67e87710fb143b58cf64511/wgpu-core/src/hub.rs#L339-L342) in `Token::root` that are supposed to detect attempts to create multiple roots simultaneously on the same thread. Did those debug assertions trigger in this case?

Yes that's how I found the issue.

PR #4258, opened 2023-10-18
Let the `"strict_asserts"` feature enable `Token::root` assertions.

jimblandy:
Perform run-time checks that only one root `Token` exists per thread when the `"strict_asserts"` feature is enabled, not only in debug builds.

**Checklist**

- [X] Run `cargo clippy`.


cwfitzgerald (approved):
Not terribly happy about the extra cfg, but not sure of a way around it.

PR #4259, opened 2023-10-18
Skip Integrated WARP

cwfitzgerald:
This is a problem that #3873 brings to light very clearly. As described in the code comment, windows will add a second instance of warp that lies about being an integrated card so that programs still run even if they filter CPU devices.

This is rather annoying behavior, so we now filter out those adapters.


teoxoy (approved):
LGTM!

teoxoy:
Those commit messages cracked me up :))

PR #4267, opened 2023-10-19
Add timestamp support to OpenGL

Zoxc:
This adds support for timestamp to the OpenGL backend using timer queries and query buffer objects.

This depends on an unreleased version of `glow` for the `query_counter` and `get_query_parameter_u64_with_offset` functions.

### DIFF
--- a/wgpu-hal/src/gles/adapter.rs
+++ b/wgpu-hal/src/gles/adapter.rs
@@ -228,6 +228,28 @@ impl super::Adapter {
             return None;
         }
 
+        if let Some(es_ver) = es_ver {
+            if es_ver < (3, 0) {
+                log::warn!(
+                    "Returned GLES context is {}.{}, when 3.0+ was requested",
+                    es_ver.0,
+                    es_ver.1
+                );
+                return None;
+            }
+        }
+
+        if let Some(full_ver) = full_ver {
+            if full_ver < (3, 3) {
+                log::warn!(
+                    "Returned GL context is {}.{}, when 3.3+ is needed",
+                    full_ver.0,
+                    full_ver.1
+                );
+                return None;
+            }
+        }
+
         let shading_language_version = {
             let sl_version = unsafe { gl.get_parameter_string(glow::SHADING_LANGUAGE_VERSION) };
             log::info!("SL version: {}", &sl_version);
@@ -250,28 +272,6 @@ impl super::Adapter {
         };
 
         log::trace!("Supported GL Extensions: {:#?}", extensions);
-
-        if let Some(es_ver) = es_ver {
-            if es_ver < (3, 0) {
-                log::warn!(
-                    "Returned GLES context is {}.{}, when 3.0+ was requested",
-                    es_ver.0,
-                    es_ver.1
-                );
-                return None;
-            }
-        }
-
-        if let Some(full_ver) = full_ver {
-            if full_ver < (3, 3) {
-                log::warn!(
-                    "Returned GL context is {}.{}, when 3.3+ is needed",
-                    full_ver.0,
-                    full_ver.1
-                );
-                return None;
-            }
-        }
 
         let supported = |(req_es_major, req_es_minor), (req_full_major, req_full_minor)| {
             let es_supported = es_ver
--- a/wgpu-hal/src/gles/adapter.rs
+++ b/wgpu-hal/src/gles/adapter.rs
@@ -450,6 +450,12 @@ impl super::Adapter {
             supported((3, 1), (4, 2)) || extensions.contains("GL_ARB_shader_image_load_store"),
         );
         features.set(wgt::Features::SHADER_UNUSED_VERTEX_OUTPUT, true);
+        let query_buffers = extensions.contains("GL_ARB_query_buffer_object")
+            || extensions.contains("GL_AMD_query_buffer_object");
+        if extensions.contains("GL_ARB_timer_query") && query_buffers {
+            features.set(wgt::Features::TIMESTAMP_QUERY, true);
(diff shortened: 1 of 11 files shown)


cwfitzgerald:
Want to land #4270 first so this gets test coverage. Should be able to do this tomorrow.

wgpu-hal/src/gles/adapter.rs
```diff
@@ -538,7 +545,10 @@ impl super::Adapter {
         );
         private_caps.set(
             super::PrivateCapabilities::GET_BUFFER_SUB_DATA,
-            cfg!(target_arch = "wasm32"),
+            cfg!(target_arch = "wasm32")
+                || full_ver
+                    .map(|full_ver| full_ver >= (2, 0))
+                    .unwrap_or_default(),
```
cwfitzgerald: I doubt we'll ever support GL1, so we can leave this as true

wgpu-hal/src/gles/queue.rs
```diff
@@ -808,34 +808,95 @@ impl super::Queue {
...
-                        self.temp_query_results.as_ptr() as *const u8,
-                        self.temp_query_results.len() * mem::size_of::<u64>(),
-                    )
-                };
-                match dst.raw {
-                    Some(buffer) => {
-                        unsafe { gl.bind_buffer(dst_target, Some(buffer)) };
+                if self
+                    .features
+                    .contains(wgt::Features::TIMESTAMP_QUERY_INSIDE_PASSES)
```
cwfitzgerald: I don't understand how this code works - we potentially need to deal with both occlusion and timestamp queries at all times?

cwfitzgerald: Similarly, both timestamps and occlusion queries should be able to use GL_QUERY_BUFFER if it's available, right?

cwfitzgerald (changes requested):

wgpu-hal/src/gles/queue.rs
Zoxc: It might be better to make a private feature for query buffers.

wgpu-hal/src/gles/adapter.rs
```diff
@@ -273,6 +252,27 @@ impl super::Adapter {
             }
         }
 
+        let shading_language_version = {
```
cwfitzgerald, line 253: Moving this back down is fine - could we just move the "supported GL extensions" printout below this as well?

I was trying to keep all the version numbers together before this massive extension list is printed.

wgpu-hal/src/gles/adapter.rs
```diff
@@ -450,6 +450,12 @@ impl super::Adapter {
             supported((3, 1), (4, 2)) || extensions.contains("GL_ARB_shader_image_load_store"),
         );
         features.set(wgt::Features::SHADER_UNUSED_VERTEX_OUTPUT, true);
+        let query_buffers = extensions.contains("GL_ARB_query_buffer_object")
+            || extensions.contains("GL_AMD_query_buffer_object");
+        if extensions.contains("GL_ARB_timer_query") && query_buffers {
```
cwfitzgerald:
```suggestion
        if extensions.contains("GL_ARB_timer_query") {
```

As we have the infrastructure to resolve without query_buffers, we should expose it.

We should however expose the blocking-ness to the user, as the user potentially can work around it. If we add a downlevel flag with a name like `NONBLOCKING_QUERY_RESOLVE` with a description like:

```rust
/// If this is true, calls to `CommandEncoder::resolve_query_set` will be performed on the queue timeline.
///
/// If this is false, calls to `CommandEncoder::resolve_query_set` will be performed on the device (i.e. cpu) timeline
/// and will block that timeline until the query has data. You may work around this limitation by waiting until the submit
/// whose queries you are resolving is fully finished (through use of `queue.on_submitted_work_done`) and only
/// then submitting the resolve_query_set command. The queries will be guarenteed finished, so will not block.
///
/// Supported by:
/// - Vulkan,
/// - DX12
/// - Metal
/// - OpenGL 4.2+
///
/// Not Supported by:
/// - GL ES / WebGL
```

cwfitzgerald: Does the code as written work with GLES's `EXT_disjoint_timer_query`/`EXT_disjoint_timer_query_webgl2`? I suspect it might?

cwfitzgerald (changes requested):
Looking good, I think we need one more iteration and we should be good

wgpu-hal/src/gles/adapter.rs
Zoxc: Probably. They'd need `glow` support. Not sure if I can test those either.

cwfitzgerald: Alright, we can leave those out.

wgpu-types/src/lib.rs
```diff
@@ -1463,6 +1463,23 @@ bitflags::bitflags! {
...
+        ///
+        /// Supported by:
+        /// - Vulkan,
+        /// - DX12
+        /// - Metal
+        /// - OpenGL 4.4+
+        ///
+        /// Not Supported by:
+        /// - GL ES / WebGL
+        const NONBLOCKING_QUERY_RESOLVE = 1 << 22;
```
cwfitzgerald, line 1482: Check the dx12/vk/metal backends and make sure this is enabled there. Some assume all downlevel, some don't.

cwfitzgerald (approved):
G2G after comment

wgpu-types/src/lib.rs
Zoxc: Vulkan was missing it.

PR #4268, opened 2023-10-19
Allow xtasks to have more dependencies

cwfitzgerald:
**Checklist**

- [x] Run `cargo clippy`.

As #3873 paves the way for a proper wasm implementation, including a webserver, we want the ability for certain xtask commands to have many more dependencies than the base xtask. This facilitates that. The base xtask has exactly 4 dependencies.

When you ask for a command which needs a certain set of dependencies, it will recurse, adding those dependencies via a feature.


teoxoy (approved):
LGTM

PR #4270, opened 2023-10-20
Update WARP and Install Mesa on Windows CI

### LINKED ISSUE #4255: Enable GLon12 on Windows CI
With #4248 on the horizon, we want to be able to test it, which requires GL setup on windows on the headless machines. We can use GLon12

Here is the dialog I had with Jesse on the DX12 discord about how to set this up. We should look at the mesa build option.

```text
cwfitzgerald: Before I go down this rabbit hole of possibly setting this up - is it possible to use GLon12 on top of warp to get a gl device when you don't have a gpu (trying to test gl on windows in CI)
Jessie: Yes
Jessie: That's how we test GLOn12 in Mesa CI
cwfitzgerald: Oh nice! Is there any docs on how to actually set up GLon12?
Jessie: The easiest way is just install the GL/CL compat pack from the Microsoft store. On a server SKU for CI though that might be tricky. There is a version available that should be installable on server from https://github.com/microsoft/OpenCLOn12/releases/tag/v1.2112.2.0 but that's a bit stale - I need to put up a new one at some point
Jessie: The other option is to build it or grab a Mesa build from https://github.com/pal1000/mesa-dist-win/releases and drop the opengl32.dll and other DLLs from that package next to the .exe you're looking to test to override the OS's OpenGL ICD loader
cwfitzgerald: This solution looks likely the easiest to pull off in CI at least - with the gl/cl compat package, if there a way to force its use for local testing purposes?
Jessie: It kicks in if there's no GL driver. You can do that by disabling your GPU (including the GL driver), or by modifying the registry to unregister the GL driver
cwfitzgerald: Alright, it sounds like it would be easier to do the Mesa dll thing locally as well anyway
cwfitzgerald: Thanks for all the info! 
Jessie: Happy to have more people using this stuff
```

cwfitzgerald:
**Checklist**

- [ ] Run `cargo clippy`.

**Connections**

Closes #4255 

**Description**

**Testing**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -296,7 +296,20 @@ jobs:
           rustup toolchain install ${{ env.REPO_MSRV }} --no-self-update --profile=minimal
           cargo -V
 
-      - name: install llvmpipe, vulkan sdk
+      - name: install warp
+        if: matrix.os == 'windows-2022'
+        shell: powershell
+        run: |
+          Invoke-WebRequest -Uri https://www.nuget.org/api/v2/package/Microsoft.Direct3D.WARP/1.0.7.1 -OutFile warp.zip
+          Expand-Archive warp.zip -DestinationPath warp
+
+          New-Item -ItemType Directory target/debug -Force
+          New-Item -ItemType Directory target/debug/deps -Force
+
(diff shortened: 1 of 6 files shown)


teoxoy:
Great stuff!

PR #4271, opened 2023-10-20
Use the device OpenGL context for WGL surfaces

Zoxc:
This uses a single OpenGL context per device instead of creating an OpenGL context per surface and sharing it using `wglShareLists`. This is allowed by setting up the pixel format on the surfaces before trying to use the OpenGL context. This is cleaner and makes presentation work on Intel Sandy Bridge.

Tested on i5-1035G1, i5-6200U and HD 5870.


cwfitzgerald (approved):
Confirmed that this also works with multiple windows and on my 1080ti!

PR #4275, opened 2023-10-22
Add sRGB and linear blending example

Zoxc:
This tests if sRGB and linear framebuffers are working as expected.


examples/srgb-blend/src/main.rs
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    fn render(
+        &mut self,
+        view: &wgpu::TextureView,
+        device: &wgpu::Device,
+        queue: &wgpu::Queue,
+        spawner: &wgpu_example::framework::Spawner,
+    ) {
+        device.push_error_scope(wgpu::ErrorFilter::Validation);
```
cwfitzgerald, line 177: We don't actually need any manual testing of the error scope - if you don't add an error scope, the default uncaptured error handler panics.

examples/common/src/framework.rs
```diff
@@ -29,6 +29,7 @@ pub enum ShaderStage {
 }
 
 pub trait Example: 'static + Sized {
+    const SRGB: bool = true;
```
cwfitzgerald, line 33: If we make this a function, there doesn't need to be const generics in the example

cwfitzgerald (changes requested):
My only real question is if this should be an example or just a test. I think this is fine as is if you feel strongly, just throwing that possibility out there.

Zoxc:
It's kind of a test, but you need to be able to run it outside the test framework to observe proper surface / presentation behavior.

cwfitzgerald:
Yeah fair enough

examples/common/src/framework.rs
Zoxc: You'd still need const generics as `framework::run` doesn't take arguments.

cwfitzgerald (approved):

PR #4278, opened 2023-10-22
Implement mipmap views on the OpenGL backend

Zoxc:
This makes the `mipmap` example work on AMD HD 5870 / llvmpipe on Windows.

### DIFF
--- a/examples/mipmap/src/main.rs
+++ b/examples/mipmap/src/main.rs
@@ -514,8 +514,7 @@ static TEST: wgpu_example::framework::ExampleTestParams =
         width: 1024,
         height: 768,
         optional_features: wgpu::Features::default(),
-        base_test_parameters: wgpu_test::TestParameters::default()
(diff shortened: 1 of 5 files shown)


cwfitzgerald (commented):
The code looks fine, and we can leave this to a follow up, but we need to add a downlevel flag for this, and add validation at the wgpu-core level to make sure that any bound texture is only used with one range of mipmaps, so users get proper errors, not corruption.

teoxoy:
We recently talked about this restriction in the committee (https://docs.google.com/document/d/1ERkTrtClVa3JSIGUUaTT3eYR9T8TtM06XNOqrkAdI94/view) and also decided on:

Validate (at draw call time) that all bound textures (not storage textures) share the same mip range.

cwfitzgerald (changes requested):
Hitting this button for admin purposes

Zoxc:
I'm not very familiar with the wgpu-core level so I'll leave that as follow up work.

cwfitzgerald (approved):

PR #4287, opened 2023-10-23
Fix missing feature checks for dx11 and dx12

valaphee:
Add missing feature checks for dx11 and dx12 in surface creation, to allow for removing dx11 and dx12 feature.


cwfitzgerald (approved):
Heh, cfgs are hard

PR #4294, opened 2023-10-24
Improve error message: "Format Bgra8UnormSrgb can't be multisampled"

### LINKED ISSUE #4288: Confusing error, "Format Bgra8UnormSrgb can't be multisampled", needs better wording
I've been experimenting with MSAA multisampling for the first time.

It appears that, by default, WGPU (and also WebGPU) only support x4 multisampling for the Bgra8UnormSrgb texture format. I believe this because guaranteed_format_features says so:

```
wgpu::TextureFormat::Bgra8UnormSrgb.guaranteed_format_features(adapter.features()) = TextureFormatFeatures {
    allowed_usages: TextureUsages(
        COPY_SRC | COPY_DST | TEXTURE_BINDING | RENDER_ATTACHMENT,
    ),
    flags: TextureFormatFeatureFlags(
        FILTERABLE | MULTISAMPLE_X4 | MULTISAMPLE_RESOLVE | BLENDABLE,
    ),
}
```

This is all good so far; no complaints so far.

My complaint is that when I tried to use x8 multisampling I get the error "**Format Bgra8UnormSrgb can't be multisampled**" which is misleading. **The statement is simply false**, Bgra8UnormSrgb *can* be multisampled. The error message makes a simple statement that is false.

Fortunately, I chose to start by using x4 multisampling, so I got the multisampling to work without encountering this error. If I had encountered this error earlier I would have been very confused and would have probably started refactoring my code to use a different texture format.

I propose changing the wording of this error to something like "Format Bgra8UnormSrgb could not be multisampled". A more open-ended error message like this makes it clear that something else, besides the texture format, might be to blame for the multisampling failure. This would be a trivial change, simply changing one string to another.

If we want to go further, the error could say something like "Format Bgra8UnormSrgb does not support x8 multisampling". This would require more than changing just a couple of strings though.

I intend to investigate these changes myself, although anyone else is welcome to as well, no worries. I'm happy for any input or suggestions.

DevJac:
A misleading error message was stating that "Format can't be multisampled" when using an unsupported sample count. This is misleading because often these texture formats can be multisample by using a supported sample count.

Fixes #4288

### DIFF
--- a/wgpu-core/src/pipeline.rs
+++ b/wgpu-core/src/pipeline.rs
@@ -299,8 +299,8 @@ pub enum ColorStateError {
     FormatNotBlendable(wgt::TextureFormat),
     #[error("Format {0:?} does not have a color aspect")]
     FormatNotColor(wgt::TextureFormat),
-    #[error("Format {0:?} can't be multisampled")]
-    FormatNotMultisampled(wgt::TextureFormat),
+    #[error("Sample count {0} is not supported by format {1:?} on this device. It may be supported by your adapter through the TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES feature.")]
+    InvalidSampleCount(u32, wgt::TextureFormat),
     #[error("Output format {pipeline} is incompatible with the shader {shader}")]
     IncompatibleFormat {
         pipeline: validation::NumericType,
@@ -321,8 +321,8 @@ pub enum DepthStencilStateError {
     FormatNotDepth(wgt::TextureFormat),
     #[error("Format {0:?} does not have a stencil aspect, but stencil test/write is enabled")]
     FormatNotStencil(wgt::TextureFormat),
-    #[error("Format {0:?} can't be multisampled")]
-    FormatNotMultisampled(wgt::TextureFormat),
+    #[error("Sample count {0} is not supported by format {1:?} on this device. It may be supported by your adapter through the TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES feature.")]
+    InvalidSampleCount(u32, wgt::TextureFormat),
 }
 
 #[derive(Clone, Debug, Error)]
--- a/wgpu-core/src/pipeline.rs
+++ b/wgpu-core/src/pipeline.rs
@@ -299,8 +299,8 @@ pub enum ColorStateError {
     FormatNotBlendable(wgt::TextureFormat),
     #[error("Format {0:?} does not have a color aspect")]
     FormatNotColor(wgt::TextureFormat),
-    #[error("Sample count {0} is not supported by format {1:?} on this device. It may be supported by your adapter through the TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES feature.")]
-    InvalidSampleCount(u32, wgt::TextureFormat),
+    #[error("Sample count {0} is not supported by format {1:?} on this device. The WebGPU spec guarentees {2:?} samples are supported by this format. With the TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES feature your device supports {3:?}.")]
+    InvalidSampleCount(u32, wgt::TextureFormat, Vec<u32>, Vec<u32>),
     #[error("Output format {pipeline} is incompatible with the shader {shader}")]
     IncompatibleFormat {
         pipeline: validation::NumericType,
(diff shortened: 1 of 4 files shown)


wgpu-core/src/pipeline.rs
```diff
@@ -299,8 +299,8 @@ pub enum ColorStateError {
     FormatNotBlendable(wgt::TextureFormat),
     #[error("Format {0:?} does not have a color aspect")]
     FormatNotColor(wgt::TextureFormat),
-    #[error("Format {0:?} can't be multisampled")]
-    FormatNotMultisampled(wgt::TextureFormat),
+    #[error("Sample count {0} is not supported by format {1:?} on this device. It may be supported by your adapter through the TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES feature.")]
```
cwfitzgerald:
```suggestion
    #[error("Sample count {0} is not supported by format {1:?} on this device. The WebGPU spec guarentees {2:?} samples are supported on this format. With the TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES feature, your device supports {3:?}.")]
```

Adding the lists of supported sample counts will make this a lot more actionable. The error case is a slow path, so don't worry about not writing the most efficient struct.

cwfitzgerald (changes requested):
Good progress! A bit of possible improvement.

wgpu-core/src/pipeline.rs
DevJac: I think we would have to instantiate a new adapter with the TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES feature in order to know what sample counts are supported by that feature. Instantiating a new adapter seems like too much for an error message.

I did improve the error message further though, take a look.

DevJac:
I added `TextureFormatFeatureFlags.supported_sample_counts` and included the supported sample counts in the error messages. I also updated a 3rd error message which I was using as a pattern for my previous error message changes.

wgpu-core/src/pipeline.rs
cwfitzgerald: The feature is just an internal logic gate, you should be able to (where these errors are made) call `adapter.get_texture_format_features` even with the feature off.

DevJac: I don't know how to get guaranteed format features? I've seen the function `guaranteed_format_features`, but it requires the device features which I don't have access to at the points in the code where these errors are created.

DevJac: I did manage to produce error messages such as:


Sample count 8 is not supported by format Bgra8UnormSrgb on this device (supported sample counts are: [1, 4]). The TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES feature enables support for sample counts [1, 2, 4, 8].


and


Sample count 3 is not supported by format Bgra8UnormSrgb on this device (supported sample counts are: [1, 2, 4, 8]). The TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES feature enables support for sample counts [1, 2, 4, 8].


That last one is a little confusing though.

cwfitzgerald: The features are available from the device - in this case `self.features`.

DevJac:
I included information about spec supported sample counts and device supported sample counts, as suggested. @cwfitzgerald

DevJac:
Squashed the most recent commits.

wgpu-core/src/pipeline.rs
DevJac: I believe this is implemented the way you wanted with the wording you wanted. Thanks for the help.

wgpu-types/src/lib.rs
```diff
@@ -2097,6 +2097,16 @@ impl TextureFormatFeatureFlags {
...
+
+    /// A `Vec` of supported sample counts.
+    pub fn supported_sample_counts(&self) -> Vec<u32> {
+        let all_possible_sample_counts: Vec<u32> = vec![1, 2, 4, 8, 16];
+        all_possible_sample_counts
+            .iter()
+            .filter(|sc| self.sample_count_supported(**sc))
+            .cloned()
+            .collect()
+    }
```
```suggestion
    pub fn supported_sample_counts(&self) -> Vec<u32> {
        let all_possible_sample_counts: [u32; 5] = [1, 2, 4, 8, 16];
        all_possible_sample_counts
            .into_iter()
            .filter(|&sc| self.sample_count_supported(sc))
            .collect()
    }
```

Small nit here, will auto-apply

cwfitzgerald (approved):
Thanks!

PR #4202, opened 2023-10-03
Bump raw window handle to 0.6.0

TornaxO7:
**Checklist**

- [x] Run `cargo clippy`.
- [ ] Run `cargo clippy --target wasm32-unknown-unknown` if applicable.
- [ ] Add change to CHANGELOG.md. See simple instructions inside file.

**Connections**
This PR tries to fix the code of https://github.com/gfx-rs/wgpu/pull/4193

**Description**
- Bumping up the dependency `raw-window-handle` from `0.5` -> `0.6`
- New error-kind added to `CreateSurfaceErrorKind`: `CreateSurfaceErrorKind::RawHandle(raw_window_handle::HandleError)` which holds the error when trying to get the an handler for a `Window` or a `Display`.

**Testing**
TODO

**Status**
We have to wait for https://github.com/rust-windowing/winit/pull/3126 because `winit` hasn't bumped `raw-window-handle` yet, hence I'm getting the following error messages for example:
```rust
error[E0277]: the trait bound `Window: raw_window_handle::borrowed::HasDisplayHandle` is not satisfied
    --> examples/hello-windows/src/main.rs:23:56
     |
23   |         let surface = unsafe { instance.create_surface(&window) }.unwrap();
     |                                         -------------- ^^^^^^^ the trait `raw_window_handle::borrowed::HasDisplayHandle` is not implemented for `Window`
     |                                         |
     |                                         required by a bound introduced by this call
     |
     = help: the following other types implement trait `raw_window_handle::borrowed::HasDisplayHandle`:
               Box<H>
               raw_window_handle::borrowed::DisplayHandle<'a>
               Rc<H>
               Arc<H>
               &H
               &mut H
note: required by a bound in `wgpu::Instance::create_surface`
    --> /home/tornax/Programming/pull/wgpu/wgpu/src/lib.rs:1940:49
     |
1939 |     pub unsafe fn create_surface<
     |                   -------------- required by a bound in this associated function
1940 |         W: raw_window_handle::HasWindowHandle + raw_window_handle::HasDisplayHandle,
     |                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `Instance::create_surface`
                                                                                                                                                                                                       
```

PR #4295, opened 2023-10-25
Release v0.18.0

cwfitzgerald:


daxpedda:
Could we not wait for https://github.com/gfx-rs/wgpu/pull/4202?

cwfitzgerald:
Not really, there's a decent amount of internal stuff depending on the wgpu release going out today.

More generally, we've taken to not waiting for particular things to land before cutting releases - they go out every 12 weeks like clockwork. We've found if we allow waiting, releases can easily "wait for just one more thing" until releases are multiple months late.

RWH is a big enough compatibility thing in the ecosystem, we might consider doing something special once it lands, but we shouldn't hold up existing releases for anything.

daxpedda:
Alright, thanks for the explanation!

PR #4299, opened 2023-10-25
More complete implementation of "lose the device".

bradwerth:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
N/A

**Description**
This provides a way for wgpu-core to specify a callback on "lose the device". It ensures this callback is called at the appropriate times: either after `device.destroy` has empty queues, or on demand from `device.lose`.

**Testing**
A test has been added to device.rs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,13 @@ Bottom level categories:
 - Hal
 -->
 
+## Unreleased ##
+
+### Changes ###
+
+#### General ####
+* `LoseDeviceClosure` callback mechanism provided so user agents can resolve `GPUDevice.lost` Promises at the appropriate time by @bradwerth in [#4299](https://github.com/gfx-rs/wgpu/pull/4299)
+
 ## v0.18.0 (2023-10-25)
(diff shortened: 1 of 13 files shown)


nical (commented):
I would prefer the terminology to match the spec more for API entry points:
- `LoseDeviceClosure` -> `DeviceLostClosure`
- `set_lose_device_callback` -> `set_device_lost_callback`
- etc.

The spec uses "lose the device" for descriptions of the steps to take but APIs use "lost".

With the exception of `fn device_lose(...)` which if I understand correctly is the internal entry point for marking that device was lost and isn't part of the public API so we can be a bit more free with the terminology, (or you could rename it into `device_mark_lost`).

bradwerth:
I'm in some kind of merge hell here, trying to rebase back onto trunk, and I'm really tempted to resubmit as a new PR. If no objections in the next hours, I'll do that.

bradwerth:
Abandoned in favor of https://github.com/gfx-rs/wgpu/pull/4645.

PR #4573, opened 2023-10-26
Support nv12 texture format

xiaopengli89:
**Checklist**

- [x] Run `cargo clippy`.

**Description**
With this pr, we can create/import nv12 textures, which helps with video rendering.

### DIFF
--- a/tests/tests/nv12_texture/mod.rs
+++ b/tests/tests/nv12_texture/mod.rs
@@ -1,6 +1,6 @@
 //! Tests for nv12 texture creation and sampling.
 
-use wgpu_test::{gpu_test, GpuTestConfiguration, TestParameters};
+use wgpu_test::{gpu_test, FailureCase, GpuTestConfiguration, TestParameters};
 
 #[gpu_test]
 static NV12_TEXTURE_CREATION_SAMPLING: GpuTestConfiguration = GpuTestConfiguration::new()
@@ -116,4 +116,122 @@ static NV12_TEXTURE_CREATION_SAMPLING: GpuTestConfiguration = GpuTestConfigurati
         rpass.draw(0..4, 0..1);
         drop(rpass);
         ctx.queue.submit(Some(encoder.finish()));
+    });
+
+#[gpu_test]
+static NV12_TEXTURE_CREATION_BAD_VIEW_FORMATS: GpuTestConfiguration = GpuTestConfiguration::new()
+    .parameters(
+        TestParameters::default()
+            .features(wgpu::Features::TEXTURE_FORMAT_NV12)
+            .expect_fail(FailureCase::always()),
+    )
+    .run_sync(|ctx| {
+        let size = wgpu::Extent3d {
+            width: 256,
+            height: 256,
+            depth_or_array_layers: 1,
+        };
+        let _ = ctx.device.create_texture(&wgpu::TextureDescriptor {
+            label: None,
+            dimension: wgpu::TextureDimension::D2,
+            size,
+            format: wgpu::TextureFormat::NV12,
+            usage: wgpu::TextureUsages::TEXTURE_BINDING,
+            mip_level_count: 1,
+            sample_count: 1,
+            view_formats: &[wgpu::TextureFormat::Rgba8Unorm],
+        });
+    });
+
+/// For D3D12 backend, textures always have plane 0
+#[gpu_test]
+static NV12_TEXTURE_VIEW_PLANE_ON_NON_PLANAR_FORMAT: GpuTestConfiguration =
+    GpuTestConfiguration::new()
+        .parameters(
+            TestParameters::default()
+                .features(wgpu::Features::TEXTURE_FORMAT_NV12)
+                .expect_fail(FailureCase::backend(
+                    wgpu::Backends::all().remove(wgpu::Backends::DX12),
+                )),
+        )
+        .run_sync(|ctx| {
+            let size = wgpu::Extent3d {
+                width: 256,
+                height: 256,
+                depth_or_array_layers: 1,
+            };
+            let tex = ctx.device.create_texture(&wgpu::TextureDescriptor {
+                label: None,
+                dimension: wgpu::TextureDimension::D2,
+                size,
+                format: wgpu::TextureFormat::R8Unorm,
+                usage: wgpu::TextureUsages::TEXTURE_BINDING,
+                mip_level_count: 1,
+                sample_count: 1,
+                view_formats: &[],
+            });
+            let _ = tex.create_view(&wgpu::TextureViewDescriptor {
+                plane: Some(0),
+                ..Default::default()
+            });
+        });
+
+#[gpu_test]
+static NV12_TEXTURE_VIEW_PLANE_OUT_OF_BOUNDS: GpuTestConfiguration = GpuTestConfiguration::new()
+    .parameters(
+        TestParameters::default()
+            .features(wgpu::Features::TEXTURE_FORMAT_NV12)
+            .expect_fail(FailureCase::always()),
+    )
+    .run_sync(|ctx| {
+        let size = wgpu::Extent3d {
+            width: 256,
+            height: 256,
+            depth_or_array_layers: 1,
+        };
+        let tex = ctx.device.create_texture(&wgpu::TextureDescriptor {
+            label: None,
+            dimension: wgpu::TextureDimension::D2,
+            size,
+            format: wgpu::TextureFormat::NV12,
+            usage: wgpu::TextureUsages::TEXTURE_BINDING,
+            mip_level_count: 1,
+            sample_count: 1,
+            view_formats: &[wgpu::TextureFormat::R8Unorm, wgpu::TextureFormat::Rg8Unorm],
+        });
+        let _ = tex.create_view(&wgpu::TextureViewDescriptor {
+            format: Some(wgpu::TextureFormat::R8Unorm),
+            plane: Some(2),
+            ..Default::default()
+        });
+    });
+
+#[gpu_test]
+static NV12_TEXTURE_BAD_FORMAT_VIEW_PLANE: GpuTestConfiguration = GpuTestConfiguration::new()
+    .parameters(
+        TestParameters::default()
+            .features(wgpu::Features::TEXTURE_FORMAT_NV12)
+            .expect_fail(FailureCase::always()),
+    )
+    .run_sync(|ctx| {
+        let size = wgpu::Extent3d {
+            width: 256,
+            height: 256,
+            depth_or_array_layers: 1,
+        };
+        let tex = ctx.device.create_texture(&wgpu::TextureDescriptor {
+            label: None,
+            dimension: wgpu::TextureDimension::D2,
+            size,
+            format: wgpu::TextureFormat::NV12,
+            usage: wgpu::TextureUsages::TEXTURE_BINDING,
+            mip_level_count: 1,
+            sample_count: 1,
+            view_formats: &[wgpu::TextureFormat::R8Unorm, wgpu::TextureFormat::Rg8Unorm],
+        });
+        let _ = tex.create_view(&wgpu::TextureViewDescriptor {
+            format: Some(wgpu::TextureFormat::R8gUnorm),
+            plane: Some(0),
+            ..Default::default()
+        });
     });
--- a/tests/tests/nv12_texture/mod.rs
+++ b/tests/tests/nv12_texture/mod.rs
@@ -1,0 +1,119 @@
+//! Tests for nv12 texture creation and sampling.
+
+use wgpu_test::{gpu_test, GpuTestConfiguration, TestParameters};
+
+#[gpu_test]
+static NV12_TEXTURE_CREATION_SAMPLING: GpuTestConfiguration = GpuTestConfiguration::new()
+    .parameters(TestParameters::default().features(wgpu::Features::TEXTURE_FORMAT_NV12))
+    .run_sync(|ctx| {
+        let size = wgpu::Extent3d {
+            width: 256,
+            height: 256,
+            depth_or_array_layers: 1,
+        };
+        let target_format = wgpu::TextureFormat::Bgra8UnormSrgb;
+
+        let shader = ctx
+            .device
+            .create_shader_module(wgpu::include_wgsl!("nv12_texture.wgsl"));
+        let pipeline = ctx
+            .device
+            .create_render_pipeline(&wgpu::RenderPipelineDescriptor {
+                label: Some("nv12 pipeline"),
+                layout: None,
+                vertex: wgpu::VertexState {
+                    module: &shader,
+                    entry_point: "vs_main",
+                    buffers: &[],
+                },
+                fragment: Some(wgpu::FragmentState {
+                    module: &shader,
+                    entry_point: "fs_main",
+                    targets: &[Some(target_format.into())],
+                }),
+                primitive: wgpu::PrimitiveState {
+                    topology: wgpu::PrimitiveTopology::TriangleStrip,
+                    strip_index_format: Some(wgpu::IndexFormat::Uint32),
+                    ..Default::default()
+                },
+                depth_stencil: None,
+                multisample: wgpu::MultisampleState::default(),
+                multiview: None,
+            });
+
+        let tex = ctx.device.create_texture(&wgpu::TextureDescriptor {
+            label: None,
+            dimension: wgpu::TextureDimension::D2,
+            size,
+            format: wgpu::TextureFormat::NV12,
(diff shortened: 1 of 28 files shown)


wgpu-hal/src/vulkan/conv.rs
```diff
@@ -74,6 +74,7 @@ impl super::PrivateCapabilities {
                 }
             }
             Tf::Depth16Unorm => F::D16_UNORM,
+            Tf::NV12 => unreachable!(),
```
AdrianEddy: I believe this is `F::G8_B8R8_2PLANE_420_UNORM` according to [ffmpeg](https://github.com/FFmpeg/FFmpeg/blob/ba6a5e7a3d4c372a98f910dda7e4d65847eda817/libavutil/hwcontext_vulkan.c#L194).

To properly support Vulkan, this needs to be implemented in the Vulkan view [subresource aspect](https://github.com/gfx-rs/wgpu/blob/af4a97fe063e34f14a57751ddde9da5935175c4d/wgpu-hal/src/vulkan/conv.rs#L592) as `ImageAspectFlags::PLANE_0`, `ImageAspectFlags::PLANE_1` or `ImageAspectFlags::PLANE_2`

More info: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkImageSubresourceRange.html#:~:text=be%20VK_IMAGE_ASPECT_COLOR_BIT.-,When%20creating%20a%20VkImageView%2C%20if%20sampler%20Y%E2%80%B2CBCR,%E2%80%99s%20subresourceRange%20must%20be%20VK_IMAGE_ASPECT_PLANE_0_BIT%2C%20VK_IMAGE_ASPECT_PLANE_1_BIT%20or%20VK_IMAGE_ASPECT_PLANE_2_BIT.,-Valid%20Usage

xiaopengli89: Currently [ash](https://github.com/ash-rs/ash/blob/f5e7606c039c5a054544231509e1daf436820da5/ash/src/vk/bitflags.rs#L387C6-L387C22) lacks the definition of `VK_IMAGE_ASPECT_PLANE_*_BIT`, I think we can support vulkan at some time later.

cwfitzgerald: We can still support it later if you want to, but you can define your own flags: https://docs.rs/ash/latest/ash/vk/struct.ImageAspectFlags.html#method.from_raw

xiaopengli89: Done.

AdrianEddy: > Currently [ash](https://github.com/ash-rs/ash/blob/f5e7606c039c5a054544231509e1daf436820da5/ash/src/vk/bitflags.rs#L387C6-L387C22) lacks the definition of `VK_IMAGE_ASPECT_PLANE_*_BIT`, I think we can support vulkan at some time later.

I thought so too, but they are apparently defined here: https://github.com/ash-rs/ash/blob/master/ash/src/vk/feature_extensions.rs#L85-L87

cwfitzgerald:
There's CI failures here, but we really need some kind of tests for this new kind of texture before we land support for this. Things like texture/buffer copies, etc. Otherwise we need to make sure the usages block those. Additionally, sampling etc needs tests.

xiaopengli89:
> There's CI failures here, but we really need some kind of tests for this new kind of texture before we land support for this. Things like texture/buffer copies, etc. Otherwise we need to make sure the usages block those. Additionally, sampling etc needs tests.

Currently we can create nv12 textures, but cannot copy from/to planes other than plane0, as this requires changing many related apis. Users can use [hal::Device::texture_from_raw](https://github.com/gfx-rs/wgpu/blob/ad3094c126e191d879a6d8684d254e13d4f8cb09/wgpu-hal/src/dx12/device.rs#L283) to import external nv12 textures for rendering, which is the most common use case. So currently I can only add a test that contains creation and sampling.

cwfitzgerald:
We should then either prevent these textures from being copied, or document that they currently can only copy from plane one (and test it to make sure it doesn't invoke ub - the main thing I'm worried about).

cwfitzgerald (changes requested):
Waiting on tests

xiaopengli89:
I don't know why the test fails on macOS:
> Validation Error: [ VUID-VkImageCreateInfo-pNext-06722 ] Object 0: handle = 0x13b83f218, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x1a04b807 | vkCreateImage(): VkImageFormatListCreateInfo::pViewFormats[0] (VK_FORMAT_R8_UNORM) and VkImageCreateInfo::format (VK_FORMAT_G8_B8R8_2PLANE_420_UNORM) are not compatible. The Vulkan spec states: If a VkImageFormatListCreateInfo structure was included in the pNext chain and VkImageFormatListCreateInfo::viewFormatCount is not zero, then each format in VkImageFormatListCreateInfo::pViewFormats must either be compatible with the format as described in the compatibility table or, if flags contains VK_IMAGE_CREATE_BLOCK_TEXEL_VIEW_COMPATIBLE_BIT, be an uncompressed format that is size-compatible with format (https://vulkan.lunarg.com/doc/view/1.3.250.1/mac/1.3-extensions/vkspec.html#VUID-VkImageCreateInfo-pNext-06722)

image

xiaopengli89:
https://github.com/KhronosGroup/MoltenVK/issues/440

It seems that MoltenVK does not support multi-planar format, so I disabled the feature on macOS/iOS, also prevent copying from/to nv12 textures.

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -530,6 +530,21 @@ impl PhysicalDeviceFeatures {
             supports_bgra8unorm_storage(instance, phd, caps.device_api_version),
         );
 
+        // https://github.com/KhronosGroup/MoltenVK/issues/440
+        #[cfg(not(any(target_os = "macos", target_os = "ios")))]
+        features.set(
+            F::TEXTURE_FORMAT_NV12,
+            supports_format(
```
i509VCB: I don't think this is enough.

VK_FORMAT_G8_B8R8_2PLANE_420_UNORM(_KHR) was added in VK_KHR_sampler_ycbcr_conversion or Vulkan 1.1. This should check for the extension or Vulkan version and reject the format if both are false.

wgpu-hal/src/vulkan/conv.rs
```diff
@@ -1,5 +1,10 @@
 use ash::vk;
 
+// https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkImageAspectFlagBits.html
+const VK_IMAGE_ASPECT_PLANE_0_BIT: vk::Flags = 0x00000010;
+const VK_IMAGE_ASPECT_PLANE_1_BIT: vk::Flags = 0x00000020;
+const VK_IMAGE_ASPECT_PLANE_2_BIT: vk::Flags = 0x00000040;
```
i509VCB: These constants are available in ash: https://docs.rs/ash/0.37.3+1.3.251/ash/vk/struct.ImageAspectFlags.html#associatedconstant.PLANE_0_KHR

xiaopengli89: Thanks.

wgpu-hal/src/vulkan/adapter.rs
xiaopengli89: > Validation Error: [ VUID-VkImageCreateInfo-pNext-06722 ] Object 0: handle = 0x1400bbe18, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x1a04b807 | vkCreateImage(): VkImageFormatListCreateInfo::pViewFormats[0] (VK_FORMAT_R8_UNORM) and VkImageCreateInfo::format (VK_FORMAT_G8_B8R8_2PLANE_420_UNORM) are not compatible. The Vulkan spec states: If a VkImageFormatListCreateInfo structure was included in the pNext chain and VkImageFormatListCreateInfo::viewFormatCount is not zero, then each format in VkImageFormatListCreateInfo::pViewFormats must either be compatible with the format as described in the compatibility table or, if flags contains VK_IMAGE_CREATE_BLOCK_TEXEL_VIEW_COMPATIBLE_BIT, be an uncompressed format that is size-compatible with format (https://vulkan.lunarg.com/doc/view/1.3.250.1/mac/1.3-extensions/vkspec.html#VUID-VkImageCreateInfo-pNext-06722)

Ci always failed on macOS(MoltenVK), but I can create nv12 texture properly on my windows pc with vulkan backend.

i509VCB: Probably wasn't clear, you need to do the extension check/version check AND supports_format

Technically the format could be unsupported in Vulkan 1.1 but it is a known format.

xiaopengli89: > Probably wasn't clear, you need to do the extension check/version check AND supports_format
> 
> Technically the format could be unsupported in Vulkan 1.1 but it is a known format.

I did the extension check/version check and supports_format, but got the same error, I think it's a bug of MoltenVK.

wgpu-core/src/device/resource.rs
```diff
@@ -718,7 +718,18 @@ impl<A: HalApi> Device<A> {
...
+            if !matches!(
+                (desc.format, *format),
+                (
+                    TextureFormat::NV12,
+                    TextureFormat::R8Unorm
+                        | TextureFormat::R8Uint
+                        | TextureFormat::Rg8Unorm
+                        | TextureFormat::Rg8Uint,
+                )
+            ) && desc.format.remove_srgb_suffix() != format.remove_srgb_suffix()
```
cwfitzgerald: Could we move this logic into a function on format that determines if a given view format is compatible with the texture format.

tests/tests/nv12_texture/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+        let y_view = tex.create_view(&wgpu::TextureViewDescriptor {
+            format: Some(wgpu::TextureFormat::R8Unorm),
+            plane: Some(0),
+            ..Default::default()
+        });
+        let uv_view = tex.create_view(&wgpu::TextureViewDescriptor {
+            format: Some(wgpu::TextureFormat::Rg8Unorm),
+            plane: Some(1),
+            ..Default::default()
+        });
```
cwfitzgerald, line 63: There doesn't seem to be anything validating that the plane and the view format agree.

cwfitzgerald: More generally, there doesn't seem to be anything validating that the plane is valid for the given format.

Would it be possible to write a short test that just tries to create various invalid view configurations and makes sure it errors.

Specifically I'm thinking
- Using planes on non-planar texture
- Using out of bounds planes
- Using the wrong plane with the wrong format

cwfitzgerald (changes requested):
Some more testing and validation needed, but looking good.

wgpu-core/src/device/resource.rs
xiaopengli89: Done.

wgpu-hal/src/metal/adapter.rs
```diff
@@ -232,6 +232,7 @@ impl crate::Adapter<super::Api> for super::Adapter {
                 }
                 flags
             }
+            Tf::NV12 => unreachable!(),
```
nical: This doesn't look unreachable to me. Would it make sense to return empty caps if we don't handle this format in metal?

nical:
This looks almost ready to land, let's wait for #3626 (arcanization) which will hopefully be merged next week though.

tests/tests/nv12_texture/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+        let _ = ctx.device.create_texture(&wgpu::TextureDescriptor {
+            label: None,
+            dimension: wgpu::TextureDimension::D2,
+            size,
+            format: wgpu::TextureFormat::NV12,
+            usage: wgpu::TextureUsages::TEXTURE_BINDING,
+            mip_level_count: 1,
+            sample_count: 1,
+            view_formats: &[wgpu::TextureFormat::Rgba8Unorm],
+        });
```
cwfitzgerald: Sorry I should have been more clear here. We should be checking very specifically for a wgpu validation error - these errors can't come from vulkan validation or a panic - we need to raise the validation error ourselves. There's a function in wgpu-test called [fail](https://wgpu.rs/doc/wgpu_test/fn.fail.html) which can assert that there is a validation failure.

Once that is in place, you can remove the expect_fail from the test parameters.

(This goes for all of the tests)

tests/tests/nv12_texture/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+    });
+
+/// For DX11/DX12 backend, textures always have plane 0
```
cwfitzgerald: The behavior should be consistent across backends. We should treat all non-NV12 textures as "not-planar" and raise a validation error if used in a planar way.

cwfitzgerald (changes requested):

wgpu-types/src/lib.rs
```diff
@@ -2318,6 +2327,14 @@ pub enum TextureFormat {
     /// [`Features::DEPTH32FLOAT_STENCIL8`] must be enabled to use this texture format.
     Depth32FloatStencil8,
 
+    /// YUV 4:2:0 video resource format.
+    /// Valid luminance data view formats for this video resource format are [`TextureFormat::R8Unorm`] and [`TextureFormat::R8Uint`].
+    /// Valid chrominance data view formats (width and height are each 1/2 of luminance view) for this video resource format are [`TextureFormat::Rg8Unorm`] and [`TextureFormat::Rg8Uint`].
+    /// Width and height must be even.
```
```suggestion
    /// YUV 4:2:0 chroma subsampled format.
    ///
    /// Contains two planes:
    /// - 0: Single 8 bit channel luminance.
    /// - 1: Dual 8 bit channel chrominance at half width and half height.
    ///
    /// Valid view formats for luminance are [`TextureFormat::R8Unorm`] and [`TextureFormat::R8Uint`].
    /// 
    /// Valid view formats for chrominance are [`TextureFormat::Rg8Unorm`] and [`TextureFormat::Rg8Uint`].
    ///
    /// Width and height must be even.
```

Making docs a bit more verbose

cwfitzgerald: There doesn't appear to be any validation of this, unless I'm missing something.

We should add a test for this (sorry I didn't catch this the first round).

One way this could be implemented is if we treat NV12's block size as (2, 2), then it will use the compressed texture block size validation.

cwfitzgerald (changes requested):
Looking really good, one more round, thanks for sticking with it!

wgpu-core/src/command/clear.rs
```diff
@@ -338,6 +338,11 @@ fn clear_texture_via_buffer_copies<A: HalApi>(
         hal::FormatAspects::COLOR
     );
 
+    if texture_desc.format == wgt::TextureFormat::NV12 {
+        // TODO: Currently COPY_DST for NV12 textures is unsupported.
```
i509VCB, line 341: Should this have a validation error? Or is this something you'd need ycbcr samplers for in Vulkan and that hasn't been implemented yet.

cwfitzgerald: This code is the runtime clear stuff, which we unconditionally invoke right now. So basically we can't properly zero them out right now.

cwfitzgerald (approved):
Thanks for the work!

teoxoy:
CI is now running into:

```
Validation Error: [ VUID-VkImageCreateInfo-pNext-06722 ] | MessageID = 0x1a04b807 | vkCreateImage(): pCreateInfo->pNext<VkImageFormatListCreateInfo>.pViewFormats[0] (VK_FORMAT_R8_UNORM) and VkImageCreateInfo::format (VK_FORMAT_G8_B8R8_2PLANE_420_UNORM) are not compatible. The Vulkan spec states: If a VkImageFormatListCreateInfo structure was included in the pNext chain and VkImageFormatListCreateInfo::viewFormatCount is not zero, then each format in VkImageFormatListCreateInfo::pViewFormats must either be compatible with the format as described in the compatibility table or, if flags contains VK_IMAGE_CREATE_BLOCK_TEXEL_VIEW_COMPATIBLE_BIT, be an uncompressed format that is size-compatible with format 
```

Looking at [this table](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap49.html#formats-compatibility-classes), I don't think it's valid to pass the plane formats as `view_formats`. The plane functionality is similar to the existing aspect functionality used for combined depth-stencil textures where the format specified for the view is the format of one of the planes/aspects.

I think it would have made sense to add variants for the planes to `TextureAspect` rather than the new `plane` param.

PR #706, opened 2020-06-07
Return errors on create_render_pipeline

kvark:
This is a follow-up to #705 that switches `create_render_pipeline` to return `Result`.

PR #175, opened 2019-05-14
API update with naming, formats, and extra fields 

kvark:

PR #4575, opened 2023-10-26
Avoid a division by zero and improve the validation of vertex attributes

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Found a CTS run in firefox: https://treeherder.mozilla.org/logviewer?job_id=433879486&repo=try&lineNumber=1796

**Description**

The array stride can be zero which causes an invalid division in `bundle::State::vertex_limits`. That validation is implemented differently from the spec which makes it difficult to compare. I think that it isn't quite correct.
 
In the process of investigating that I found some differences between the vertex attribute validation as we implement it and the spec so I fixed these as well.

**Testing**

Covered by the CTS.

### DIFF
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -2682,7 +2682,6 @@ impl<A: HalApi> Device<A> {
 
         let mut vertex_steps = Vec::with_capacity(desc.vertex.buffers.len());
         let mut vertex_buffers = Vec::with_capacity(desc.vertex.buffers.len());
-        let mut total_attributes = 0;
         let mut shader_expects_dual_source_blending = false;
         let mut pipeline_expects_dual_source_blending = false;
         for (i, vb_state) in desc.vertex.buffers.iter().enumerate() {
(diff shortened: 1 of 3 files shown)


nical:
While submitting this I realized that my fix was too simplistic. That particular validation needs to be redone in a more invasive way I think.

wgpu-core/src/pipeline.rs
```diff
@@ -344,8 +344,8 @@ pub enum CreateRenderPipelineError {
     InvalidSampleCount(u32),
     #[error("The number of vertex buffers {given} exceeds the limit {limit}")]
     TooManyVertexBuffers { given: u32, limit: u32 },
-    #[error("The total number of vertex attributes {given} exceeds the limit {limit}")]
-    TooManyVertexAttributes { given: u32, limit: u32 },
+    #[error("The a vertex attribute location {given} exceeds the limit {limit}")]
```
cwfitzgerald:
```suggestion
    #[error("A vertex attribute location {given} exceeds the limit {limit}")]
```

wgpu-core/src/device/resource.rs
```diff
@@ -2713,7 +2713,16 @@ impl<A: HalApi> Device<A> {
...
             for attribute in vb_state.attributes.iter() {
-                if attribute.offset >= 0x10000000 {
+                let format_size = attribute.format.size();
+                let attrib_end_offset = attribute.offset + format_size;
+                let offset_limit = if vb_state.array_stride == 0 {
+                    max_array_stride as u64
+                } else {
+                    vb_state.array_stride
+                }
+                .min(0x10000000);
```
cwfitzgerald, line 2723: What does this magic number signify?

cwfitzgerald (changes requested):

wgpu-core/src/device/resource.rs
nical: I don't know, it was already there

cwfitzgerald: That's a spooky kind of magic number

fanninpm: If it helps, this magic number probably first appeared in #706, where it replaced another magic number that probably first appeared in #175.

cwfitzgerald:
@nical any chance of this getting revived? Otherwise we probably should close it.

PR #4577, opened 2023-10-26
wgsl-errors: Reverse "old" and "new" sides of diff output.

jimblandy:
When an error snapshot test fails and we generate a diff comparing the expected output with the actual output, treat the expected output as the diff "from", and the actual output as the diff "to" - not the reverse.


ErichDonGubler (approved):
ezpz 😉

PR #4578, opened 2023-10-26
[wgsl-in] Replace `TypedExpression` with a generic enum, `Typed`.

jimblandy:
Replace the `TypedExpression` struct, used to distinguish between WGSL pointers and references since Naga has only `Pointer`, with an enum, `Typed`, with variants for references and plain types. This cleans up a bunch of code, since the struct's `is_reference` field basically served as a detached enum discriminant. This also prepares the code for adding abstract types.


teoxoy (approved):
Great stuff!

PR #4585, opened 2023-10-27
Remove `flume` dependency of `wgpu`

a1phyr:
**Checklist**

- [x] Run `cargo clippy`.
- [ ] Add change to CHANGELOG.md. See simple instructions inside .


cwfitzgerald (approved):
Nice, I personally prefer flume, but less dependencies are good, and you solved the big problem with mpsc channels

PR #4586, opened 2023-10-27
vulkan: Log validation layer messages during instance creation/destruction

exrook:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
n/a

**Description**
We currently miss some potentially useful validation layer messages by only registering our message callback after the instance has been created, by additionally passing the callback to VkCreateInstance we can log messages that occur during instance creation and destruction.

**Testing**
covered by existing vulkan tests


cwfitzgerald (approved):
Nice

PR #4590, opened 2023-10-27
Fix typo in documentation of QueueWriteBufferView

### LINKED ISSUE #4587: `QueueWriteBufferView` has confusing documentation
The description for `wgpu::QueueWriteBufferView` reads:

> A read-only view into a staging buffer.
> 
> Reading into this buffer won’t yield the contents of the buffer from the GPU and is likely to be slow. Because of this, although [AsMut](https://doc.rust-lang.org/nightly/core/convert/trait.AsMut.html) is implemented for this type, [AsRef](https://doc.rust-lang.org/nightly/core/convert/trait.AsRef.html) is not.

This is contradictory. A read-only view shouldn't implement `AsMut`. Should the documentation say "write-only"?

dthusian:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Fixes #4587

**Description**
Change "read-only" to "write-only" in documentation of "QueueWriteBufferView".

**Testing**
View the generated docs.

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -4649,7 +4649,7 @@ impl<'a> RenderBundleEncoder<'a> {
     }
 }
 
-/// A read-only view into a staging buffer.
+/// A write-only view into a staging buffer.
 ///
 /// Reading into this buffer won't yield the contents of the buffer from the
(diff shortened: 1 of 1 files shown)


cwfitzgerald (approved):
Thanks!

PR #4594, opened 2023-10-27
Naga(msl): Clear named expressions after writing each function

DJMcNab:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
This bug started in https://github.com/gfx-rs/naga/pull/2309

Found in linebender/vello#398

**Description**
When writing the local variables for a shader, the existance of previous functions could cause expressions to be incorrectly treated as if they were named.

This change appears to match other backends, and fixes the case where expressions which were named in earlier functions are used in local variable declarations

**Testing**
I made a reproducing wgsl file:
```wgsl
fn func() -> f32 {
    return 1.0;
}

fn blend_mix() {
    func();
}

fn blend_compose(
) {
    var fa = 0.0;
}
```
This used to have the problem, but doesn't after this PR. Additionally, we ~~will validate~~ have validated that Vello works as expected ~~(once someone with a Mac can check)~~.

I'm not sure that a regression test is very useful here.


raphlinus:
Confirm this works with Vello. Thanks for the fix!

cwfitzgerald (dismissed):
Dismissing erroneous wgpu review request

teoxoy (approved):
Makes sense, thanks!

PR #4588, opened 2023-10-27
fix example-framework

TornaxO7:
**Checklist**

- [ ] Run `cargo clippy`.
- [ ] Run `cargo clippy --target wasm32-unknown-unknown` if applicable.
- [ ] Add change to CHANGELOG.md. See simple instructions inside file.

**Connections**
Closes https://github.com/gfx-rs/wgpu/issues/4584

**Description**
Adding a missing attribute to the testing framework which should fix the issue.

**Testing**
`cargo run` in the examples

PR #3932, opened 2023-07-16
Port to raw-window-handle v0.6

notgull:
**Checklist**

- [ ] Run `cargo clippy`.
- [ ] Run `cargo clippy --target wasm32-unknown-unknown` if applicable.
- [ ] Add change to CHANGELOG.md. See simple instructions inside file.

**Connections**

Testing out rust-windowing/raw-window-handle#125 so I can get a good idea of how it impacts the ecosystem.

**Description**

I implemented a system similar to how `softbuffer` is set up now. However it requires dynamic dispatch due to `wgpu_core` storing the `Surface`s in `Global`.

**Testing**

None for now, hence the draft status. Metal is also broken.

PR #4597, opened 2023-10-27
Safe surface creation

### LINKED ISSUE #1463: Let's make surface creation safe!
There has been some discussion in the past over how to make create_surface safe:

https://github.com/gfx-rs/wgpu-rs/issues/78

https://github.com/gfx-rs/wgpu-rs/pull/281#discussion_r415495633

From these comments, my impression was that the only issue was not being able to trust the pointers provided by the type implementing https://docs.rs/raw-window-handle/0.3.3/raw_window_handle/trait.HasRawWindowHandle.html.  However, this is an `unsafe trait` and the requirements it puts on its implementors are actually quite strong:

> Safety guarantees

> **Users can safely assume that non-null/0 fields are valid handles**, and it is up to the implementer of this trait to ensure that condition is upheld. However, *It is entirely valid behavior for fields within each platform-specific RawWindowHandle variant to be null or 0, and appropriate checking should be done before the handle is used.**

> Despite that qualification, implementers should still make a best-effort attempt to fill in all available fields. If an implementation doesn't, and a downstream user needs the field, it should try to derive the field from other fields the implementer does provide via whatever methods the platform provides.

> **The exact handles returned by raw_window_handle must remain consistent between multiple calls to raw_window_handle, and must be valid for at least the lifetime of the HasRawWindowHandle implementer.**

(emphasis mine).

We can instantly see that just by taking `&W` where `W: HasRawWindowHandle`, we do not need to worry about arbitrary pointers or integers--we can rely on the guarantees of `HasRawWindowHandle` alone, which ensures that as long as the underlying `W` instance is alive:

* The handles won't change.
* All handles will either be valid, or null/0 (depending on platform).

Since wgpu already needs to deal with handles in a backend-sensitive way, it makes perfect sense for it to just check all the individual handles.  Problem solved, right?  Unfortunately, it turns out that while this is the only *documented* source of unsafety, it is not the only source of unsafety when using a surface created by `create_surface`.  @cwfitzgerald gave the following code as an example, when performing rendering using a task system off the main thread:

```rust
frame = None;
loop {
    MainEventsCleared => {
        wait_for_frame(frame);

        frame = Some(render());
    }
}
```

When trying to close a window, this event loop would segfault on some Linux systems, but moving `wait_for_frame` below the `frame = _` portion (at the end of the `MainEventsCleaared` arm) would prevent the segfault.  Here, `wait_for_frame` waits for all commands to be submitted from the CPU to the GPU (but does not wait for the GPU).  This suggests that after exiting the event loop (dropping the window), the window surface may be freed on the CPU on some platforms while there are still in-flight commands being sent to the surface from another thread; presumably, someone (the window manager or graphics driver) is able to make sure device resources are not reused until rendering is complete, but the same is not true of host resources.

The implication here is clear: in general, we can't treat a surface as valid unless the underlying window from which it was created is valid.  Here, "window" doesn't necessarily have to be a window (although in practice it usually will be), it just has to be the object that owns whatever resource is referred to by the handle it passes to wgpu.  Note that in a wgpu context, we only care about "real" surfaces with valid handles.

While on some platforms (e.g. Android), it seems that a window is always kept alive as long as the surface is, on others (e.g. X11 on Linux), it seems like the surface may be deallocated together with the Window (we could not inspect the actual implementation of https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCreateXlibSurfaceKHR.html so we could not verify this for certain, but either way the Vulkan documentation doesn't indicate that there's any requirement that the surface hold onto a reference to the window).  So our first conclusion is that if we want to be safe, we should make sure--within Rust--that a window always outlives the surface from which it was created.

We discussed several approaches here, but the most obvious is this one:

1. Surface is parameterized by its underlying window: `Surface<W>`.
2. The safe version of `create_surface` (in wgpu) takes `W: HasRawWindowHandle` *by value* rather than by reference.
3. The safe version of `create_surface` (in wgpu) validates any handles it uses are non-0 / non-null (depending on platform), runs all the checks and resource creation steps it currently does, and then stores `W` *within* the `Surface` structure.
4. `Surface<W>` exposes a few methods to get back the window data: `fn window(&self) -> &W`, and `fn into_window(self) -> W`.  Note that we do *not* provide `fn window_mut(&mut self) -> &mut W`, for reasons explained below.

First, safety arguments for why this makes it okay to use the window handle associated with the surface:

- From the definition of `HasRawWindowHandle`, as long as the underlying `W` is alive, the window handles will not change after repeated calls, so we know our checks from (3) will continue to be valid.
- From the definition of `HasRawWindowHandle`, as long as the underlying `W` is alive, the handle returned from a call will refer to a live window, so we know the window is valid as long as the instance of `W` is; which is at least as long as the `Surface` itself, *provided* that we don't provide any means of partially moving out of the `Window`.
- And, because we don't provide `&mut W` access or other partial window move methods, there is no way to alter the stored window without dropping the whole `Surface` (this is why we can't implement `deref_mut()`).

These may seem like uncomfortably strong restrictions on the window used for gfx-rs.  We address these concerns in two ways:

- Technically: We propose adding several implementations of `HasRawWindowHandle` upstream in the `raw-window-handle` crate (all of which we believe are justified by the requirements on `HasRawWindowHandle`):

```rust

impl<'a, T: HasRawWindowHandle> HasRawWindowHandle for &'a T { /* .. */ }
impl<T: HasRawWindowHandle> HasRawWindowHandle for Rc<T> { /* .. */ }
impl<T: HasRawWindowHandle> HasRawWindowHandle for Arc<T> { /* .. */ }

```

This means that in order to share access to the `Window` with the rest of an application, one can simply pass `Arc<W>`, `Rc<W>`, or `&W` (and possibly other implementations, should they prove useful) to `create_surface`.  Due to the nature of how windows work (all meaningful operations on them must work in the context of a separate device being able to query them through a duplicable handle), we know that the important, shared component of the window (the part that needs to be alive when the surface is) must have interior mutability of some sort if it can be mutated at all; for actually existing window APIs, this interior mutability is made safe dynamically rather than within the type system, so we don't lose anything by preventing `&mut` access to the window while wgpu may have references to it.  In exchange, we get a powerful safety argument that relies solely on the correct implementation of `HasRawWindowHandle` and `create_surface`, rather than requiring detailed investigation of every possible window management API.

We also examine practical, large wgpu projects to see if these restrictons would be onerous.  From my own work on Veloren I know that these restrictions would be no problem for us.  We also looked at Bevy, Dotrix, and Iced, and manually verified that they could move to this API with minimal changes:

- The Bevy API has a global hashmap from window ids to winit windows and could pretty clearly be turned into a map from window ids to `Arc<winit::Window>`.
  It requires `&mut` access only to its own internal `bevy::Window`, which is distinct from the one to which you have a handle, and even this is only needed
  during window creation right before it's inserted into the hash map.  So there would be no issue with switching to `Arc<winit::Window>`.
- For Dotrix, the event loop gets an owned `winit::Window` and it could easily have `Renderer` own it: https://github.com/lowenware/dotrix/blob/fix/wgpu-pipeline-creation/src/application.rs#L67.  `Arc<winit::Window>` would make it even easier, but is not required; the code that needs access to the window can always get at the underlying `Renderer` instance by accessing it through the global services structure, so if the renderer owned the `winit::Window` (indirectly through the surface), it would work fine (`Renderer::new` constructs and stores the surface).
- For iced, it will require a minor change to their compositor API (currently, it basically just propagates up the requirement for `&W` instead of `W`, but as that API is unsafe it will have to be changed regardless).  In the actual examples I could find, the window is easily available by value, however, e.g. https://github.com/hecrj/iced/blob/ff15ebc54778dee2a7262469c8a2bcd5abecd4d1/examples/integration/src/main.rs and https://github.com/hecrj/iced/blob/1f7e8b7f3d1804c39c8e0934b25f3ef178de269c/wgpu/src/window/compositor.rs.

So we can see that our theoretical expectations translate well to practice.  We can also see that in all known big examples, `winit::Window` is what's providing the `RawWindowHandle`, which makes sense since it works to abstract window creation across as many platforms as possible.  As such, verifying that `winit::Window`'s implementation of `HasRawWindowHandle` satisfies the trait requirements, together with correctness of `wgpu`'s implementation of functions on `Surface`, will in practice be sufficient to verify the safety of real world code using this API (which is important because it means that we are not simply hiding an added unsafety burden by moving it somewhere else in the stack).

However, this is *still* not enough!  This is because we still haven't addressed the thing you actually *do* with surfaces, which is use them with swap chains.  Swap chains are currently produced from the combination of a surface and a compatible device.  When `get_next_frame()` is called on a swap chain, we acquire a texture resource on the device; the resource is compatible with the target surface.  If successful, we can then render to the texture resource from our device, and then finally (on dropping the swap frame) "release" the texture, and it is queued for presentation to the surface.  Clearly, the safety of all of this depends crucially on both the device and surface remaining alive for the duration of the swap chain, but this is not guaranteed by the current API!

As discussed on Matrix, the swap chain API is still very much in flux, with no major players having made a firm commitment to any particular API.  However, wgpu itself has been learning from how people use it (together with constraints imposed by cross-platform requirements), and the current thought is that *surfaces* should own the queue currently associated with swap chains; a surface will own one swap chain at a time, and (from our discussion) wgpu has no interest in surfaces that are not part of a swap chain.  This takes care of one safety issue: if a surface replaces the functionality of the swap chain, then clearly the swap chain cannot outlive the surface!

Since the swap chain needs to allocate device memory, this also implies that the swap chain should maintain a strong reference to the backing `Device`.  It is possible that given that `Window` is alive, the existing gfx-rs `Context` mechanism will be sufficient to prevent `Device` host memory from being deallocated, since unlike with `Window` we started out with ownership over the device data; however, it's possible that the current complex dependency graph around swap chains, swap chain frames, devices, surfaces, etc. could be greatly simplified by maintaining a direct `Arc<Device>`.  Either of these mechanisms should be sufficient to ensure safety of *creating* a swap frame while being reasonably flexible.

To make sure that a swapchain *frame* is still alive, the most straightforward method would be to add a lifetime to `SwapChainFrame`; `get_next_frame(&'a self) -> SwapChainFrame<'a>` would be the new signature defined on `Surface`, and `SwapChainFrame` would maintain a reference to the `Surface` that instantiated it.  If this is too inflexible for some purposes (although I suspect most could be made to fit into this pattern), we could define on the method as `get_next_frame<S: Borrow<Self>>(surface: S) -> SwapChainFrame<S>`, which would allow using `Rc<Surface>`, `Arc<Surface>`, `Surface`, `&Surface`, etc. as desired (we woul also want to provide an `into_surface()` and `surface()` method as usual); though we might not be able to get away with this safely for arbitrary `Borrow<Self>`, we could at least hardcode the stuff we know works.  Alternately, we could again rely on reference counting / pointer following in the shared context, and avoid any explicit references (as `Surface` would already take care of them).

---

I believe this plan closes all the potential soundness holes in the current API around windows and surfaces.  There may still be implementation issues, of course, but the safety argument should hold as long as `HasRawWindowHandle` is implemented properly.

---

So, to recap:

- `create_surface` needs to check for null and zero handles as appropriate (in a cross-platform way).
- The `Surface` will own the window `W` and provide `&W` access and `into_window(self)`, but not `&mut W`.
- `Surface` will own the resources needed for the actual swap chain (queue etc.) and those resources in turn should hold onto the Device.
- Swap chain frames will hold references to the `Surface` (and therefore `Device`) that produced them when they are acquired (exact mechanism TBD).
- From manual inspection, all known major users of wgpu can easily accommodate this API change, as long as it comes together with `HasRawWindowHandle` for `Arc<W: AsRawWindowHandle>` and related instances implemented upstream (technically I think all of them can actually own the `Window`, but the `Arc` instance makes it easier).
- Jointly, this all makes creating a surface and acquiring a swap chain (which are the two main operations on surfaces) safe regardless of what safe code does with the window; all the safety requirements rely solely on a correct implementation of unsafe trait `AsRawWindowHandle` (needs to satisfy the requirements it already has).

daxpedda:
Now that we use `raw-window-handle` v0.6 (#4202), safe `Surface` creation should be possible. This is done by adding a lifetime to `Surface` that is bound to the passed `window`.

Additionally I added `Surface::create_surface_from_raw()` which still takes a `HasWindowHandle + HasDisplayHandle` but remains `unsafe` and returns a `Surface<'static>`. 

Fixes #1463.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,10 @@ Bottom level categories:
 - Log vulkan validation layer messages during instance creation and destruction: By @exrook in [#4586](https://github.com/gfx-rs/wgpu/pull/4586)
 - `TextureFormat::block_size` is deprecated, use `TextureFormat::block_copy_size` instead: By @wumpf in [#4647](https://github.com/gfx-rs/wgpu/pull/4647)
 
+#### Safe `Surface` creation
+
+It is now possible to safely create a `wgpu::Surface` with `Surface::create_surface()` and carries a lifetime to the passed `window`. `Surface::create_surface_from_raw()` can be used to produce a `Surface<'static>`, which remains `unsafe`.
+
 #### Naga
 
 - Introduce a new `Scalar` struct type for use in Naga's IR, and update all frontend, middle, and backend code appropriately. By @jimblandy in [#4673](https://github.com/gfx-rs/wgpu/pull/4673).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,11 @@ Bottom level categories:
 
 #### Safe `Surface` creation
 
-It is now possible to safely create a `wgpu::Surface` with `Surface::create_surface()`. `Surface::create_surface_from_raw()` can be used to produce a `Surface<'static>`, which remains `unsafe`.
+It is now possible to safely create a `wgpu::Surface` with `Surface::create_surface()` by letting `Surface` hold a lifetime to `window`.
+
+Passing an owned value `window` to `Surface` will return a `Surface<'static>`. Shared ownership over `window` can still be achieved with e.g. an `Arc`. Alternatively a reference could be passed, which will return a `Surface<'window>`.
+
+`Surface::create_surface_from_raw()` can be used to continue producing a `Surface<'static>` without any lifetime requirements over `window`, which also remains `unsafe`.
 
 #### Naga
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@ Bottom level categories:
 
 #### Safe `Surface` creation
 
-It is now possible to safely create a `wgpu::Surface` with `Surface::create_surface()` and carries a lifetime to the passed `window`. `Surface::create_surface_from_raw()` can be used to produce a `Surface<'static>`, which remains `unsafe`.
+It is now possible to safely create a `wgpu::Surface` with `Surface::create_surface()`. `Surface::create_surface_from_raw()` can be used to produce a `Surface<'static>`, which remains `unsafe`.
 
 #### Naga
 
--- a/examples/common/src/framework.rs
+++ b/examples/common/src/framework.rs
@@ -1,4 +1,4 @@
-use std::sync::OnceLock;
+use std::sync::Arc;
 
 use wgpu::{Instance, Surface, WasmNotSend, WasmNotSync};
 use wgpu_test::GpuTestConfiguration;
@@ -90,17 +90,17 @@ fn init_logger() {
     }
 }
 
-struct EventLoopWrapper(EventLoop<()>);
+struct EventLoopWrapper {
+    event_loop: EventLoop<()>,
+    window: Arc<Window>,
+}
 
 impl EventLoopWrapper {
-    fn new() -> Self {
-        Self(EventLoop::new().unwrap())
-    }
-
-    fn window(&self, title: &str) -> Window {
+    pub fn new(title: &str) -> Self {
+        let event_loop = EventLoop::new().unwrap();
         let mut builder = winit::window::WindowBuilder::new();
         builder = builder.with_title(title);
-        let window = builder.build(&self.0).unwrap();
+        let window = Arc::new(builder.build(&event_loop).unwrap());
 
         #[cfg(target_arch = "wasm32")]
         {
@@ -115,8 +115,7 @@ impl EventLoopWrapper {
                 .expect("couldn't append canvas to document body");
         }
 
-        #[cfg_attr(not(target_arch = "wasm32"), allow(clippy::let_and_return))]
-        window
+        Self { event_loop, window }
     }
 }
 
@@ -144,7 +143,7 @@ impl SurfaceWrapper {
     ///
     /// We cannot unconditionally create a surface here, as Android requires
     /// us to wait until we recieve the `Resumed` event to do so.
-    fn pre_adapter(&mut self, instance: &Instance, window: &'static Window) {
+    fn pre_adapter(&mut self, instance: &Instance, window: Arc<Window>) {
         if cfg!(target_arch = "wasm32") {
             self.surface = Some(instance.create_surface(window).unwrap());
         }
@@ -166,7 +165,7 @@ impl SurfaceWrapper {
     /// On all native platforms, this is where we create the surface.
     ///
     /// Additionally, we configure the surface based on the (now valid) window size.
-    fn resume(&mut self, context: &ExampleContext, window: &'static Window, srgb: bool) {
+    fn resume(&mut self, context: &ExampleContext, window: Arc<Window>, srgb: bool) {
         // Window size is only actually valid after we enter the event loop.
         let window_size = window.inner_size();
         let width = window_size.width.max(1);
@@ -255,10 +254,7 @@ struct ExampleContext {
 }
 impl ExampleContext {
     /// Initializes the example context.
-    async fn init_async<'a, E: Example>(
-        surface: &mut SurfaceWrapper,
-        window: &'static Window,
-    ) -> Self {
+    async fn init_async<'a, E: Example>(surface: &mut SurfaceWrapper, window: Arc<Window>) -> Self {
         log::info!("Initializing wgpu...");
 
         let backends = wgpu::util::backend_bits_from_env().unwrap_or_default();
@@ -361,12 +357,9 @@ impl FrameCounter {
 
 async fn start<E: Example>(title: &str) {
     init_logger();
-    let window_loop = EventLoopWrapper::new();
-    static WINDOW: OnceLock<Window> = OnceLock::new();
-    let window = WINDOW.get_or_init(|| window_loop.window(title));
+    let window_loop = EventLoopWrapper::new(title);
     let mut surface = SurfaceWrapper::new();
-    let context = ExampleContext::init_async::<E>(&mut surface, window).await;
-
+    let context = ExampleContext::init_async::<E>(&mut surface, window_loop.window.clone()).await;
     let mut frame_counter = FrameCounter::new();
 
     // We wait to create the example until we have a valid surface.
@@ -385,14 +378,14 @@ async fn start<E: Example>(title: &str) {
     // On native this is a result, but on wasm it's a unit type.
     #[allow(clippy::let_unit_value)]
     let _ = (event_loop_function)(
-        window_loop.0,
+        window_loop.event_loop,
         move |event: Event<()>, target: &EventLoopWindowTarget<()>| {
             // We set to refresh as fast as possible.
             target.set_control_flow(ControlFlow::Poll);
 
             match event {
                 ref e if SurfaceWrapper::start_condition(e) => {
-                    surface.resume(&context, window, E::SRGB);
+                    surface.resume(&context, window_loop.window.clone(), E::SRGB);
 
                     // If we haven't created the example yet, do so now.
                     if example.is_none() {
@@ -416,7 +409,7 @@ async fn start<E: Example>(title: &str) {
                             &context.queue,
                         );
 
-                        window.request_redraw();
+                        window_loop.window.request_redraw();
                     }
                     WindowEvent::KeyboardInput {
                         event:
@@ -456,7 +449,7 @@ async fn start<E: Example>(title: &str) {
 
                         frame.present();
 
-                        window.request_redraw();
+                        window_loop.window.request_redraw();
                     }
                     _ => example.as_mut().unwrap().update(event),
                 },
--- a/examples/common/src/framework.rs
+++ b/examples/common/src/framework.rs
@@ -1,3 +1,5 @@
+use std::sync::OnceLock;
+
 use wgpu::{Instance, Surface, WasmNotSend, WasmNotSync};
 use wgpu_test::GpuTestConfiguration;
 use winit::{
@@ -88,17 +90,17 @@ fn init_logger() {
     }
 }
 
-struct EventLoopWrapper {
-    event_loop: EventLoop<()>,
-    window: Window,
-}
+struct EventLoopWrapper(EventLoop<()>);
 
 impl EventLoopWrapper {
-    pub fn new(title: &str) -> Self {
-        let event_loop = EventLoop::new().unwrap();
+    fn new() -> Self {
+        Self(EventLoop::new().unwrap())
+    }
+
+    fn window(&self, title: &str) -> Window {
         let mut builder = winit::window::WindowBuilder::new();
         builder = builder.with_title(title);
-        let window = builder.build(&event_loop).unwrap();
+        let window = builder.build(&self.0).unwrap();
 
         #[cfg(target_arch = "wasm32")]
         {
@@ -113,7 +115,8 @@ impl EventLoopWrapper {
                 .expect("couldn't append canvas to document body");
         }
 
-        Self { event_loop, window }
+        #[cfg_attr(not(target_arch = "wasm32"), allow(clippy::let_and_return))]
+        window
     }
 }
 
@@ -121,7 +124,7 @@ impl EventLoopWrapper {
 ///
 /// As surface usage varies per platform, wrapping this up cleans up the event loop code.
 struct SurfaceWrapper {
-    surface: Option<wgpu::Surface>,
+    surface: Option<wgpu::Surface<'static>>,
     config: Option<wgpu::SurfaceConfiguration>,
 }
 
@@ -141,9 +144,9 @@ impl SurfaceWrapper {
     ///
     /// We cannot unconditionally create a surface here, as Android requires
     /// us to wait until we recieve the `Resumed` event to do so.
-    fn pre_adapter(&mut self, instance: &Instance, window: &Window) {
+    fn pre_adapter(&mut self, instance: &Instance, window: &'static Window) {
         if cfg!(target_arch = "wasm32") {
-            self.surface = Some(unsafe { instance.create_surface(&window).unwrap() });
+            self.surface = Some(instance.create_surface(window).unwrap());
         }
     }
 
@@ -163,7 +166,7 @@ impl SurfaceWrapper {
     /// On all native platforms, this is where we create the surface.
     ///
     /// Additionally, we configure the surface based on the (now valid) window size.
-    fn resume(&mut self, context: &ExampleContext, window: &Window, srgb: bool) {
+    fn resume(&mut self, context: &ExampleContext, window: &'static Window, srgb: bool) {
         // Window size is only actually valid after we enter the event loop.
         let window_size = window.inner_size();
         let width = window_size.width.max(1);
@@ -173,7 +176,7 @@ impl SurfaceWrapper {
 
         // We didn't create the surface in pre_adapter, so we need to do so now.
         if !cfg!(target_arch = "wasm32") {
-            self.surface = Some(unsafe { context.instance.create_surface(&window).unwrap() });
+            self.surface = Some(context.instance.create_surface(window).unwrap());
         }
 
         // From here on, self.surface should be Some.
@@ -252,7 +255,10 @@ struct ExampleContext {
 }
 impl ExampleContext {
     /// Initializes the example context.
-    async fn init_async<E: Example>(surface: &mut SurfaceWrapper, window: &Window) -> Self {
+    async fn init_async<'a, E: Example>(
+        surface: &mut SurfaceWrapper,
+        window: &'static Window,
+    ) -> Self {
         log::info!("Initializing wgpu...");
 
         let backends = wgpu::util::backend_bits_from_env().unwrap_or_default();
@@ -355,9 +361,11 @@ impl FrameCounter {
 
 async fn start<E: Example>(title: &str) {
     init_logger();
-    let window_loop = EventLoopWrapper::new(title);
+    let window_loop = EventLoopWrapper::new();
+    static WINDOW: OnceLock<Window> = OnceLock::new();
+    let window = WINDOW.get_or_init(|| window_loop.window(title));
     let mut surface = SurfaceWrapper::new();
-    let context = ExampleContext::init_async::<E>(&mut surface, &window_loop.window).await;
+    let context = ExampleContext::init_async::<E>(&mut surface, window).await;
 
     let mut frame_counter = FrameCounter::new();
 
@@ -377,14 +385,14 @@ async fn start<E: Example>(title: &str) {
     // On native this is a result, but on wasm it's a unit type.
     #[allow(clippy::let_unit_value)]
     let _ = (event_loop_function)(
-        window_loop.event_loop,
+        window_loop.0,
         move |event: Event<()>, target: &EventLoopWindowTarget<()>| {
             // We set to refresh as fast as possible.
             target.set_control_flow(ControlFlow::Poll);
 
             match event {
                 ref e if SurfaceWrapper::start_condition(e) => {
-                    surface.resume(&context, &window_loop.window, E::SRGB);
+                    surface.resume(&context, window, E::SRGB);
 
                     // If we haven't created the example yet, do so now.
                     if example.is_none() {
@@ -408,7 +416,7 @@ async fn start<E: Example>(title: &str) {
                             &context.queue,
                         );
 
-                        window_loop.window.request_redraw();
+                        window.request_redraw();
                     }
                     WindowEvent::KeyboardInput {
                         event:
@@ -448,7 +456,7 @@ async fn start<E: Example>(title: &str) {
 
                         frame.present();
 
-                        window_loop.window.request_redraw();
+                        window.request_redraw();
                     }
                     _ => example.as_mut().unwrap().update(event),
                 },
--- a/examples/common/src/framework.rs
+++ b/examples/common/src/framework.rs
@@ -254,7 +254,7 @@ struct ExampleContext {
 }
 impl ExampleContext {
     /// Initializes the example context.
-    async fn init_async<'a, E: Example>(surface: &mut SurfaceWrapper, window: Arc<Window>) -> Self {
+    async fn init_async<E: Example>(surface: &mut SurfaceWrapper, window: Arc<Window>) -> Self {
         log::info!("Initializing wgpu...");
 
         let backends = wgpu::util::backend_bits_from_env().unwrap_or_default();
--- a/examples/hello-windows/src/main.rs
+++ b/examples/hello-windows/src/main.rs
@@ -1,26 +1,26 @@
 #![cfg_attr(target_arch = "wasm32", allow(dead_code))]
 
-use std::collections::HashMap;
+use std::{collections::HashMap, sync::Arc};
 use winit::{
     event::{Event, WindowEvent},
     event_loop::EventLoop,
     window::{Window, WindowId},
 };
 
-struct ViewportDesc<'window> {
-    window: &'window Window,
+struct ViewportDesc {
+    window: Arc<Window>,
     background: wgpu::Color,
-    surface: wgpu::Surface<'window>,
+    surface: wgpu::Surface<'static>,
 }
 
-struct Viewport<'window> {
-    desc: ViewportDesc<'window>,
+struct Viewport {
+    desc: ViewportDesc,
     config: wgpu::SurfaceConfiguration,
 }
 
-impl<'window> ViewportDesc<'window> {
-    fn new(window: &'window Window, background: wgpu::Color, instance: &wgpu::Instance) -> Self {
-        let surface = instance.create_surface(window).unwrap();
+impl ViewportDesc {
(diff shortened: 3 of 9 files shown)


examples/common/src/framework.rs
```diff
@@ -100,7 +102,8 @@ async fn setup<E: Example>(title: &str) -> Setup {
         use winit::platform::windows::WindowBuilderExtWindows;
         builder = builder.with_no_redirection_bitmap(true);
     }
-    let window = builder.build(&event_loop).unwrap();
+    static WINDOW: OnceLock<Window> = OnceLock::new();
+    let window = WINDOW.get_or_init(|| builder.build(&event_loop).unwrap());
```
daxpedda: ~~I know this is a bit sketchy, but as we are using `winit::event_loop::EventLoop::spawn()` for Wasm, we actually require a `'static` `Window`.~~
Scratch that, this was a different example I was remembering.

wgpu/src/lib.rs
```diff
@@ -388,8 +388,9 @@ static_assertions::assert_impl_all!(SurfaceConfiguration: Send, Sync);
 /// [`GPUCanvasContext`](https://gpuweb.github.io/gpuweb/#canvas-context)
 /// serves a similar role.
 #[derive(Debug)]
-pub struct Surface {
+pub struct Surface<'window> {
     context: Arc<C>,
+    surface: PhantomData<&'window ()>,
```
daxpedda: My assumption here was that `Surface` is the only type requiring the lifetime, e.g. `SurfaceTexture` doesn't. Is that correct?

wgpu/src/lib.rs
```diff
@@ -1936,12 +1931,29 @@ impl Instance {
...
+
+    /// See [`create_surface()`](Self::create_surface) for more details.
+    ///
+    /// # Safety
+    ///
+    /// - `raw_window_handle` must be a valid object to create a surface upon.
+    /// - `raw_window_handle` must remain valid until after the returned [`Surface`] is
+    ///   dropped.
+    pub unsafe fn create_surface_from_raw<
         W: raw_window_handle::HasWindowHandle + raw_window_handle::HasDisplayHandle,
```
daxpedda: It might actually be more useful to take a [`RawDisplayHandle`](https://docs.rs/raw-window-handle/0.6.0/raw_window_handle/enum.RawDisplayHandle.html) and [`RawWindowHandle`](https://docs.rs/raw-window-handle/0.6.0/raw_window_handle/enum.RawWindowHandle.html) here instead.

notgull (changes requested):
If you parameterize the `Surface` by the actual window type, you can avoid the weird `'static` mutation and instead just pass around the `Surface` wrapping an `Arc<Window>` or something. See the softbuffer version of this PR for more info (https://github.com/rust-windowing/softbuffer/pull/132)

wgpu/src/lib.rs
Zoxc: Trying to render to the surface after it's destroyed does seem like a questionable operation. I was wondering if `Surface` could return a `&'window SurfaceTexture`.

cwfitzgerald: We do have visibility in wgpu-core if there are any active surface textures - so we should error on surface unconfigure, reconfigure, or surface drop.

daxpedda: Actually, now that I think about it, if this is really an issue, it was an issue already before this PR.

daxpedda:
> If you parameterize the `Surface` by the actual window type, you can avoid the weird `'static` mutation and instead just pass around the `Surface` wrapping an `Arc<Window>` or something. See the softbuffer version of this PR for more info ([rust-windowing/softbuffer#132](https://github.com/rust-windowing/softbuffer/pull/132))

I don't think this would work as well in Wgpu as it did in Softbuffer. Wgpu has a bunch of methods that can create a `Surface` from pointers, which I would argue isn't desirable: `Surface<*mut std::ffi::c_void>` (we would also need to solve the `Send + Sync` problem then). We could of course in cases like this return a `Surface<()>`, but at that point I don't see how `Surface<'static>` is worse.

Maybe I'm missing something here, let me know.

wgpu/src/lib.rs
daxpedda: See https://github.com/rust-windowing/raw-window-handle/pull/139#issuecomment-1703884412.
As far as I interpret this isn't an issue.

daxpedda:
Rebased on `trunk`.

examples/common/src/framework.rs
daxpedda: This basically reverts #4588 now, which isn't needed now because we never drop `Window` at this point.

So if we want to *improve* this:
- Use some self-referential shenanigans. E.g. [`self_cell`](https://crates.io/crates/self_cell) or [`oroboros`](https://crates.io/crates/ouroboros).
- Separate `Window` creation from `setup()` so can hold it beyond `run()`, solving the lifetime issue because `Surface` is moved into `run()`, which makes it drop before `Window`.

We don't have to do anything though, this is just an improvement for users trying to learn how to do this from that code.

kpreid:
What about the “`raw_window_handle` must be a valid object to create a surface upon” part of the safety contract? Are we sure that there is no possible state of the window, on all platforms, that will not be surprising to wgpu and thereby cause unsound misbehavior?

daxpedda:
> What about the “`raw_window_handle` must be a valid object to create a surface upon” part of the safety contract? Are we sure that there is no possible state of the window, on all platforms, that will not be surprising to wgpu and thereby cause unsound misbehavior?

My assumption here is that's what `raw-window-handle` intended here, but I have really no knowledge about this and couldn't find anything mentioned in the documentation.

@notgull would you be able to comment on that?

notgull:
> I don't think this would work as well in Wgpu as it did in Softbuffer. Wgpu has a bunch of methods that can create a `Surface` from pointers, which I would argue isn't desirable: `Surface<*mut std::ffi::c_void>` (we would also need to solve the `Send + Sync` problem then). 

This happens in `softbuffer` sometimes too, in this case we return a [`Display<NoDisplayHandle>`](https://github.com/rust-windowing/softbuffer/blob/0bcd2e22a2b89423f7c8852dd5033c6f4aa6f9d0/src/lib.rs#L453-L463) or a [`Surface<NoWindowHandle>`](https://github.com/rust-windowing/softbuffer/blob/0bcd2e22a2b89423f7c8852dd5033c6f4aa6f9d0/src/lib.rs#L465-L475). Something similar should be able to happen in `wgpu`.

> We could of course in cases like this return a `Surface<()>`, but at that point I don't see how `Surface<'static>` is worse.

`Surface<'static>` is worse because it lies about its lifetime requirements, imparting unsoundness on the rest of the program.

> My assumption here is that's what `raw-window-handle` intended here, but I have really no knowledge about this and couldn't find anything mentioned in the documentation.

This is true; while a `WindowHandle` is alive the window is guaranteed to be alive right alongside it, like with `AsFd`.

daxpedda:
> > I don't think this would work as well in Wgpu as it did in Softbuffer. Wgpu has a bunch of methods that can create a `Surface` from pointers, which I would argue isn't desirable: `Surface<*mut std::ffi::c_void>` (we would also need to solve the `Send + Sync` problem then).
> 
> This happens in `softbuffer` sometimes too, in this case we return a [`Display<NoDisplayHandle>`](https://github.com/rust-windowing/softbuffer/blob/0bcd2e22a2b89423f7c8852dd5033c6f4aa6f9d0/src/lib.rs#L453-L463) or a [`Surface<NoWindowHandle>`](https://github.com/rust-windowing/softbuffer/blob/0bcd2e22a2b89423f7c8852dd5033c6f4aa6f9d0/src/lib.rs#L465-L475). Something similar should be able to happen in `wgpu`.

I tried this idea, but again, Wgpu doesn't make this straightforward. I have to pass down the generic now somehow in all the functions that take `Surface`. E.g. `Instance::request_adapter()` takes a `RequestAdapterOptions` which holds `&Surface`. I can't pass this down easily, because `DynContext` is supposed to be object safe, adding any generics to methods breaks that. Adding a generic to `DynContext` itself seems a bit too messy for me. Lifetimes don't have that problem because you can just use anonymous ones.

I could continue to play around with all this, but would like to hear a word from a member if this is desired in the first place or not.

daxpedda:
Rebased on trunk, no changes.

notgull:
I think I used dynamic dispatch and `'static` when I tried this in #3932 to get around this. It reduces functionality but will probably work, at least in the short term.

daxpedda:
Rebased on master.

cwfitzgerald:
So making surface creation safe is something we definitely want, I don't think this is quite the way to do it though.

Notably, we only actually need the very surface `wgpu` layer be safe - wgpu-core can purely address the raw window handle assuming that the layer above it is handling it correctly.

We should have create_surface be:

```rust
trait WgpuSurfaceRequirement: HasWindowHandle + HasDisplayHandle {}

pub fn create_surface<'a, W>(&self, window: W) -> Result<Surface<'a>, CreateSurfaceError>
where
     W: WgpuSurfaceRequirement + 'a,
```

Note that there isn't actually a generic here. We can store `W` in a `Option<Box<dyn WgpuSurfaceRequirement + 'a>>` (minding the send/sync requirements of wasm/non-wasm). For the functions that take a raw pointer to create a surface, we put `None` in that slot and make the lifetime static. 

This prevents the generics from infecting user code, and allows the flexibility to store both references and owned values as `W`, behaving as expected.

wgpu-core and below behave like they did before with raw handles everywhere.

cwfitzgerald (changes requested):

daxpedda:
@cwfitzgerald I'm unsure what exactly your request is.

I looked over my changes again, but I don't believe I introduced any generics. Maybe you were just pointing out that we don't want to introduce a generic based on the conversation I had with [nogull](https://github.com/notgull) before?

In any case, the current implementation seems to be exactly what you are requesting, including not touching anything else then `wgpu`. The only difference I can see is the `WgpuSurfaceRequirement` trait, which wasn't necessary as we don't actually store the given window, but only the surface we extract from it.

Apologies for any misunderstanding.

daxpedda:
Rebased on master.
Only `CHANGELOG.md` changed.

wgpu/src/lib.rs
```diff
@@ -1936,12 +1931,29 @@ impl Instance {
...
-    pub unsafe fn create_surface<
+    pub fn create_surface<
+        'window,
+        W: raw_window_handle::HasWindowHandle + raw_window_handle::HasDisplayHandle,
+    >(
+        &self,
+        window: &'window W,
+    ) -> Result<Surface<'window>, CreateSurfaceError> {
+        unsafe { self.create_surface_from_raw(window) }
+    }
```
cwfitzgerald: So the thing I want to avoid is having surface logically contain a _reference_ to the window. This prevents people from storing both the window and the surface in the same struct (such a struct would be self-referential). The design I proposed stores a W directly. This allows a user to not need to deal with lifetimes in most cases (only if they give us a lifetime-bearing W) and you can easily slot in a Arc<Window>. 

Indeed I was confused from the above conversation about the "going all the way into wgpu" part.

daxpedda: Ah, I see! That totally makes sense.
Will do that!

examples/hello-windows/src/main.rs
```diff
@@ -7,28 +7,28 @@ use winit::{
...
+    desc: ViewportDesc<'window>,
     config: wgpu::SurfaceConfiguration,
 }
 
-impl ViewportDesc {
-    fn new(window: Window, background: wgpu::Color, instance: &wgpu::Instance) -> Self {
-        let surface = unsafe { instance.create_surface(&window) }.unwrap();
+impl<'window> ViewportDesc<'window> {
+    fn new(window: &'window Window, background: wgpu::Color, instance: &wgpu::Instance) -> Self {
+        let surface = instance.create_surface(window).unwrap();
```
daxpedda: Some examples continued to work just fine even after we changed `Surface` to store the `window`, so I left them as is. I guess it's also good to test if this works correctly with borrowing.

Let me know if you want me to change that by just wrapping them in an `Arc`.

wgpu/src/lib.rs
```diff
@@ -400,6 +408,26 @@ pub struct Surface {
...
+                } else {
+                    "None"
+                },
+            )
+            .field("id", &self.id)
+            .field("data", &self.data)
+            .field("config", &self.config)
+            .finish()
+    }
+}
```
daxpedda, line 428: I didn't want to add the `Debug` requirement to `WgpuSurfaceRequirement`, so I implemented it by hand.

wgpu/src/lib.rs
daxpedda: Done in https://github.com/gfx-rs/wgpu/pull/4597/commits/3d57ebadabb1e356eeebed07d0d5634cdffb3cec.

examples/common/src/framework.rs
daxpedda: Resolved in https://github.com/gfx-rs/wgpu/pull/4597/commits/3d57ebadabb1e356eeebed07d0d5634cdffb3cec.

examples/hello-windows/src/main.rs
cwfitzgerald: Let's use Arc as that's the recommended way to deal with this.

CHANGELOG.md
```diff
@@ -47,6 +47,10 @@ Bottom level categories:
 - Log vulkan validation layer messages during instance creation and destruction: By @exrook in [#4586](https://github.com/gfx-rs/wgpu/pull/4586)
 - `TextureFormat::block_size` is deprecated, use `TextureFormat::block_copy_size` instead: By @wumpf in [#4647](https://github.com/gfx-rs/wgpu/pull/4647)
 
+#### Safe `Surface` creation
+
+It is now possible to safely create a `wgpu::Surface` with `Surface::create_surface()`. `Surface::create_surface_from_raw()` can be used to produce a `Surface<'static>`, which remains `unsafe`.
```
cwfitzgerald: I think we should go into more detail on the api change - specifically what actions the user should take to port their code

examples/common/src/framework.rs
```diff
@@ -252,7 +254,7 @@ struct ExampleContext {
 }
 impl ExampleContext {
     /// Initializes the example context.
-    async fn init_async<E: Example>(surface: &mut SurfaceWrapper, window: &Window) -> Self {
+    async fn init_async<'a, E: Example>(surface: &mut SurfaceWrapper, window: Arc<Window>) -> Self {
```
cwfitzgerald: Unused 'a

wgpu/src/lib.rs
```diff
@@ -379,6 +380,13 @@ impl Drop for Sampler {
 pub type SurfaceConfiguration = wgt::SurfaceConfiguration<Vec<TextureFormat>>;
 static_assertions::assert_impl_all!(SurfaceConfiguration: Send, Sync);
 
+trait WgpuSurfaceRequirement: HasWindowHandle + HasDisplayHandle + WasmNotSend + WasmNotSync {}
```
cwfitzgerald: We need to make this public and document this. Otherwise an end user would have no way to know how to get a T which implements this.

wgpu/src/lib.rs
```diff
@@ -387,9 +395,9 @@ static_assertions::assert_impl_all!(SurfaceConfiguration: Send, Sync);
 /// This type is unique to the Rust API of `wgpu`. In the WebGPU specification,
 /// [`GPUCanvasContext`](https://gpuweb.github.io/gpuweb/#canvas-context)
 /// serves a similar role.
-#[derive(Debug)]
-pub struct Surface {
+pub struct Surface<'window> {
     context: Arc<C>,
+    _surface: Option<Box<dyn 'window + WgpuSurfaceRequirement>>,
```
cwfitzgerald:
```suggestion
    _surface: Option<Box<dyn WgpuSurfaceRequirement + 'window>>,
```

Just style preference

wgpu/src/lib.rs
```diff
@@ -400,6 +408,26 @@ pub struct Surface {
     // been created is is additionally wrapped in an option.
     config: Mutex<Option<SurfaceConfiguration>>,
 }
+
+impl<'window> fmt::Debug for Surface<'window> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Surface")
+            .field("context", &self.context)
+            .field(
```
cwfitzgerald, line 415: Could you just note why were doing this manually

wgpu/src/lib.rs
```diff
@@ -1936,12 +1958,29 @@ impl Instance {
     /// - On macOS/Metal: will panic if not called on the main thread.
     /// - On web: will panic if the `raw_window_handle` does not properly refer to a
     ///   canvas element.
-    pub unsafe fn create_surface<
-        W: raw_window_handle::HasWindowHandle + raw_window_handle::HasDisplayHandle,
+    pub fn create_surface<
+        'window,
+        W: 'window + HasWindowHandle + HasDisplayHandle + WasmNotSend + WasmNotSync,
```
cwfitzgerald: Nit: put the lifetime last, use WgpuSurfaceRequirement and move to where clause

wgpu/src/lib.rs
```diff
@@ -1920,12 +1948,6 @@ impl Instance {
     /// If the specified display and window handle are not supported by any of the backends, then the surface
     /// will not be supported by any adapters.
     ///
-    /// # Safety
```
cwfitzgerald, line 1923: We really need to add documentation on how this is used and how what you pass in relates to the output lifetime of the surface.

While the signature and trait definition does technically have the information, this isn't readily grokable, especially to rust newcomers.

wgpu/src/lib.rs
```diff
@@ -1936,12 +1958,29 @@ impl Instance {
...
     >(
+        &self,
+        window: W,
+    ) -> Result<Surface<'window>, CreateSurfaceError> {
+        let mut surface = unsafe { self.create_surface_from_raw(&window) }?;
+        surface._surface = Some(Box::new(window));
+        Ok(surface)
+    }
+
+    /// See [`create_surface()`](Self::create_surface) for more details.
```
cwfitzgerald, line 1981: Could we expand on the differences between the above and this. This should explain why this has a bunch of requirements the other doesn't?

cwfitzgerald (changes requested):
Code is good, just style and docs.

wgpu/src/lib.rs
daxpedda: Why make `WgpuSurfaceRequirement` public if it works with existing types already?
I thought the only reason why `WgpuSurfaceRequirement` exists at all is because it's required to store the window in a `Box<dyn T>`, which doesn't allow multiple non-auto traits.

(in the meantime I will just go ahead and still apply the suggestion)

examples/hello-windows/src/main.rs
daxpedda: I left the `hello-triangle` example as is because it didn't require any lifetime introduction. It's also nice to make sure this actually works correctly as well (but that could easily be done in a test instead).

Let me know if you want me to convert the `hello-triangle` example as well.

wgpu/src/lib.rs
cwfitzgerald: Oh yeah, you're totally right, this can just be internal as it's only needed to physically construct the dyn we want - actually, we don't need to access the window handle at all from the dyn. We can just use Option<Box <dyn Any>> and get rid of this trait entirely.

cwfitzgerald:
LGTM after comment!

cwfitzgerald (approved):
Hit the wrong button!

wgpu/src/lib.rs
daxpedda: > We can just use Option<Box > and get rid of this trait entirely.

Did you have an idea what to store in the `Box`? I couldn't remove the trait entirely because we still need *any* trait **and** `Send + Sync`.

I was thinking of introducing a new `WasmNotSendSync` trait, which might shorten a lot of lines in the codebase as well in a follow-up PR?

cwfitzgerald: Sure, that sounds reasonable.

Note that `Send + Sync` are special in dyn - you can do `dyn Trait + Send + Sync` fine. You can't do `dyn Trait + WasmNotSend + WasmNotSync` as they aren't the "special" ones :)

wgpu/src/lib.rs
```diff
@@ -1936,12 +1961,34 @@ impl Instance {
     /// - On macOS/Metal: will panic if not called on the main thread.
     /// - On web: will panic if the `raw_window_handle` does not properly refer to a
     ///   canvas element.
-    pub unsafe fn create_surface<
-        W: raw_window_handle::HasWindowHandle + raw_window_handle::HasDisplayHandle,
+    pub fn create_surface<
+        'window,
+        W: HasWindowHandle + HasDisplayHandle + WasmNotSend + WasmNotSync + 'window,
```
cwfitzgerald: Still need to move these to where clasues.

daxpedda: Done!
Applied the same to `create_surface_from_raw()` as well.

daxpedda:
Just rebased onto trunk again.
No changes.

cwfitzgerald (approved):
Nice!

PR #4598, opened 2023-10-27
Add more lints

daxpedda:
In #4597 I stumbled on a couple of `Surface`s where I forgot to apply `'static`, which reminded me of `elided-lifetimes-in-paths`. I used it to make sure I didn't miss any others.

I'm not sure if this is desired at all, but thought I would make a PR so maintainers can take a look.
I could also add the whole [`rust-2018-idioms`](https://doc.rust-lang.org/1.73.0/rustc/lints/groups.html) group if that's desirable, it's pretty simply with `cargo fix`.

Requires #4597, but I'm happy to rebase it to `trunk` if it helps.

### DIFF
--- a/wgpu/src/backend/direct.rs
+++ b/wgpu/src/backend/direct.rs
@@ -101,7 +101,7 @@ impl Context {
         &self,
         adapter: &wgc::id::AdapterId,
         hal_device: hal::OpenDevice<A>,
-        desc: &crate::DeviceDescriptor,
+        desc: &crate::DeviceDescriptor<'_>,
         trace_dir: Option<&std::path::Path>,
     ) -> Result<(Device, Queue), crate::RequestDeviceError> {
         let global = &self.0;
@@ -134,7 +134,7 @@ impl Context {
         &self,
         hal_texture: A::Texture,
         device: &Device,
-        desc: &TextureDescriptor,
+        desc: &TextureDescriptor<'_>,
     ) -> Texture {
         let descriptor = desc.map_label_and_view_formats(|l| l.map(Borrowed), |v| v.to_vec());
         let global = &self.0;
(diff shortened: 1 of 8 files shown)


kpreid:
> I could easily add more lints like …

`rust-2018-compatibility` and `rust-2021-compatibility` aren't relevant in code that is already in the 2021 edition.

daxpedda:
> > I could easily add more lints like …
> 
> `rust-2018-compatibility` and `rust-2021-compatibility` aren't relevant in code that is already in the 2021 edition.

Aye, that makes sense.
I'm wondering why [`absolute-paths-not-starting-with-crate`](https://doc.rust-lang.org/1.73.0/rustc/lints/listing/allowed-by-default.html#absolute-paths-not-starting-with-crate) is in `rust-2018-compatibility` then ... that still seems useful to me in any edition. But it's the only relevant lint I could find.

kpreid:
> I'm wondering why [absolute-paths-not-starting-with-crate](https://doc.rust-lang.org/1.73.0/rustc/lints/listing/allowed-by-default.html#absolute-paths-not-starting-with-crate) is in rust-2018-compatibility then ... that still seems useful to me in any edition. But it's the only relevant lint I could find.

That one's not enabled because, past the 2015 edition, such paths don't even work; [the semantics were changed in 2018](https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html) such that absolute paths may only refer to named crates, where they could previously refer to items declared in the current crate root. So, that lint only means anything in 2015 edition code.

**All** compatibility lints are about “this will mean something different, or be rejected, in the next edition”. So there is no reason to enable any of them once you are using later editions.

daxpedda:
Ah! Thanks @kpreid, much appreciated.

daxpedda:
Just to clarify: my suggestion in OP now boils down to adding [`rust-2018-idioms`](https://doc.rust-lang.org/1.73.0/rustc/lints/groups.html) only.

cwfitzgerald:
Yeah adding rust-2018 idioms seems good

cwfitzgerald (changes requested):
Requesting changes per above.

daxpedda:
Done.

cwfitzgerald (approved):
We probably should add these to wgpu-core and hal as well, but let's wait till #3626 lands.

PR #4602, opened 2023-10-28
Use `VK_EXT_robustness2` only when not using an outdated intel iGPU driver

### LINKED ISSUE #4599: Intel iGPUs with outdated graphics drivers struggle to render with `VK_EXT_robustness2` enabled
**Description**
For 0.15 and later, the shadow example doesn't render the cubes nor shadows, only the plane. The problem is expanded on in bevy https://github.com/bevyengine/bevy/issues/8037, where it only shows the clear color and sometimes UI. The shadow example was just what I used to find where the problem was in the update from 0.14.0 to 0.15.0.

**Repro steps**
1. Find an intel laptop with an integrated GPU
2. Make sure the graphics driver is relatively old (one I'm using is from June 2021)
3. Run the shadow example in a version of wgpu 0.15 and later

**Extra materials**
The problem comes from 39f6b03e197830ee9243443bfc47fa673d44a9ad / https://github.com/gfx-rs/wgpu/pull/3327

Removing this snippet of code fixes the bevy issue and the shadow example:
https://github.com/gfx-rs/wgpu/blob/43b74f9892abab4069e653671f75b3ff5850aa62/wgpu-hal/src/vulkan/adapter.rs#L671-L675 I have no idea why, so I wanted to ask if anyone has any ideas on why `VK_EXT_robustness2` causes almost everything to not be rendered in vulkan, yet be fine with a more recent driver or a different GPU. A couple thoughts I got was perhaps:
- "could be that Intel screwed up the implementation of that extension"
- "maybe an alignment issue? 'It is important to query the robustUniformBufferAccessSizeAlignment and robustStorageBufferAccessSizeAlignment from VkPhysicalDeviceRobustness2PropertiesEXT as the alignment of where buffers are bound-checked is different between implementations'"

_I'm sure this extension is useful to have, so I don't want to add a PR removing it for all GPUs, and am currently unaware of how best to add an exception for outdated intel iGPU drivers._

**Platform**
Windows 10
Intel UHD Graphics, version: 27.20.100.9664
wgpu 0.15 and later
Vulkan

### LINKED ISSUE #8037: Broken rendering in examples on intel graphics
### Bevy version

v0.10.0

### [Optional] Relevant system information

The system is a Dell XPS 13 7390 2-in-1 running windows.

```
2023-03-11T05:22:43.409198Z  INFO bevy_winit::system: Creating new window "Bevy App" (0v0)
2023-03-11T05:22:44.121756Z  INFO bevy_render::renderer: AdapterInfo { name: "Intel(R) Iris(R) Plus Graphics", vendor: 32902, device: 35410, device_type: IntegratedGpu, driver: "Intel Corporation", driver_info: "Intel driver", backend: Vulkan }
2023-03-11T05:22:45.181282Z  INFO bevy_diagnostic::system_information_diagnostics_plugin::internal: SystemInfo { os: "Windows 11 Home", kernel: "22621", cpu: "Intel(R) Core(TM) i7-1065G7 CPU @ 1.30GHz", core_count: "4", memory: "7.8 GiB" }
```

### What you did

Ran `cargo run --example breakout`.

Also occurs with other examples, and in release builds, but the breakout one seems to be the most interesting broken output.

### What went wrong

The blocks, paddle, and ball don't render, but the score does. The game plays as normal, with score going up and with correct sound.

A few other examples i've tried only render a grey screen, but some seem to work

image

### Additional Information

Possibly related to #6746 ? Am happy to try and debug on my end by running tools etc.

Additionally, a simple game I was testing/building on my own computer seems to render correctly

### LINKED ISSUE #8395: Unable to render sprites
## Bevy version

v0.10.1

## Relevant system information

SystemInfo { os: "Windows 11 Home", kernel: "22621", cpu: "Intel(R) Core(TM) i5-1035G4 CPU @ 1.10GHz", core_count: "4", memory: "7.6 GiB" }

AdapterInfo { name: "Intel(R) Iris(R) Plus Graphics", vendor: 32902, device: 35418, device_type: IntegratedGpu, driver: "Intel Corporation", driver_info: "Intel driver", backend: Vulkan }

## What you did

I have cloned the bevy repo and try to run some of the examples after realising my game doesn't render after updating version

## What went wrong

If I run the sprite example i get nothing but a gray background
If I run the contributor example i get the names of the contriutors but not the sprites just a gray backgound

I have tried to run both examples with version 0.9.0 and there they work as expeted so it seems that the issue got introduced in version 0.10.0 because it persists on the main branch

I have also tried the [Boids](https://github.com/gfx-rs/wgpu/tree/trunk/wgpu/examples/boids) example in the wgpu repo and it seems to work as intended

TheoDulka:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Fixes #4599, fixes https://github.com/bevyengine/bevy/issues/8037 and fixes https://github.com/bevyengine/bevy/issues/8395

**Description**
Intel iGPUs with outdated drivers can break rendering if `VK_EXT_robustness2` is pushed.
The solution is to not push if its driver version is before [31.0.101.2115](https://www.intel.com/content/www/us/en/download/762755/intel-6th-10th-gen-processor-graphics-windows.html), because this one works. There's probably an earlier functional version as it was just the earliest version I could find and try (due to Intel's End Of Life on 6th gen proc).

**Testing**
Not really something testable beyond getting an Intel iGPU laptop and seeing if it is unable to render. I used the shadow example for my testing.


wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -669,8 +669,20 @@ impl PhysicalDeviceCapabilities {
         }
 
         // Optional `VK_EXT_robustness2`
+        // Intel iGPUs with outdated drivers can break rendering if `VK_EXT_robustness2` is pushed.
+        // Driver version 31.0.101.2115 works, but there's probably an earlier functional version.
```
cwfitzgerald: Could you include a reference to the original issue for context?

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -669,8 +669,20 @@ impl PhysicalDeviceCapabilities {
...
-            extensions.push(vk::ExtRobustness2Fn::name());
+            const DRIVER_VERSION_INTEL_31_0_101: u32 = 0x194000;
+            const DRIVER_VERSION_INTEL_WORKING: u32 = DRIVER_VERSION_INTEL_31_0_101 + 2115;
+
+            let props = self.properties;
+            let is_intel_igpu_outdated = props.vendor_id == crate::auxil::db::intel::VENDOR
+                && props.device_type == vk::PhysicalDeviceType::INTEGRATED_GPU
+                && props.driver_version < DRIVER_VERSION_INTEL_WORKING;
+
+            if !is_intel_igpu_outdated {
```
cwfitzgerald: Could we put a `log::trace` level log when this condition isn't hit - with something like "Disabling robustBufferAccess2: ntegrated Intel Driver found with driver version {} which is less than known good driver 31.0.101.2115"

Also we should make sure this doesn't hit mesa drivers as well, only proprietary.

cwfitzgerald (changes requested):
Unfortunately I don't think this will work directly. You can't just remove the extension from the list.

We should instead just set the feature flag to false - in adapter.rs:236, we can just flip it false on this condition.

wgpu-hal/src/vulkan/adapter.rs
TheoDulka: What do you mean by "this doesn't hit"? The issue potentially affecting mesa drivers or the `is_intel_igpu_outdated` including mesa drivers? If the latter, doesn't mesa have its own vendor id here https://github.com/gfx-rs/wgpu/blob/43b74f9892abab4069e653671f75b3ff5850aa62/wgpu-hal/src/auxil/mod.rs#L28-L34 or is that what the "Mesa does not actually have a PCI vendor id." refers to, where it could be Intel's `0x8086`

TheoDulka:
> Unfortunately I don't think this will work directly. You can't just remove the extension from the list.

Yeah I realize that now, but I don't understand how I would do this:
> We should instead just set the feature flag to false - in adapter.rs:236, we can just flip it false on this condition.

wgpu-hal/src/vulkan/adapter.rs
i509VCB: In Mesa, official drivers are considered to have the vendor id of the hardware vendor. Community drivers such as asahi use Mesa's vendor id.

TheoDulka: Ok I think my newest addition to the conditional accounts for mesa drivers now

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -869,7 +869,9 @@ impl super::InstanceShared {
                     .insert(vk::PhysicalDeviceImageRobustnessFeaturesEXT::default());
                 builder = builder.push_next(next);
             }
-            if capabilities.supports_extension(vk::ExtRobustness2Fn::name()) {
+            if !is_intel_igpu_outdated_for_robustness2(capabilities.properties, capabilities.driver)
+                && capabilities.supports_extension(vk::ExtRobustness2Fn::name())
```
cwfitzgerald: This won't properly tell us that robustness 2 is not in use. We need to make this decision when filling out the capabilities struct.

cwfitzgerald (dismissed):
Test code looks good, just some changes on the enabling code.

wgpu-hal/src/vulkan/adapter.rs
TheoDulka: Okay I think I did what you were looking for

cwfitzgerald (commented):
Nothing wrong with the code, just want teo to have a final look before we merge

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -825,6 +825,23 @@ impl super::InstanceShared {
                 unsafe {
                     get_device_properties.get_physical_device_properties2(phd, &mut properties2);
                 }
+
+                if is_intel_igpu_outdated_for_robustness2(
+                    capabilities.properties,
+                    capabilities.driver,
+                ) {
+                    use crate::auxil::cstr_from_bytes_until_nul;
+                    capabilities.supported_extensions.remove(
```
teoxoy: Using `.retain()` would be best here.

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -1802,3 +1819,34 @@ fn supports_bgra8unorm_storage(
...
+    const DRIVER_VERSION_WORKING: u32 = DRIVER_VERSION_31_0_101 + 2115;
+
+    let is_outdated = props.vendor_id == crate::auxil::db::intel::VENDOR
+        && props.device_type == vk::PhysicalDeviceType::INTEGRATED_GPU
+        && props.driver_version < DRIVER_VERSION_WORKING
+        && cstr_from_bytes_until_nul(&driver.unwrap_or_default().driver_info)
+            .unwrap_or_default()
+            .to_str()
+            .unwrap_or_default()
+            == "Intel driver";
```
```suggestion
        && driver
            .map(|driver| driver.driver_id == vk::DriverId::INTEL_PROPRIETARY_WINDOWS)
            .unwrap_or_default();
```

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -1802,3 +1819,34 @@ fn supports_bgra8unorm_storage(
...
+// Intel iGPUs with outdated drivers can break rendering if `VK_EXT_robustness2` is used.
+// Driver version 31.0.101.2115 works, but there's probably an earlier functional version.
+fn is_intel_igpu_outdated_for_robustness2(
+    props: vk::PhysicalDeviceProperties,
+    driver: Option<vk::PhysicalDeviceDriverPropertiesKHR>,
+) -> bool {
+    use crate::auxil::cstr_from_bytes_until_nul;
+
+    const DRIVER_VERSION_31_0_101: u32 = 0x194000;
+    const DRIVER_VERSION_WORKING: u32 = DRIVER_VERSION_31_0_101 + 2115;
```
```suggestion
    const DRIVER_VERSION_WORKING: u32 = (101 << 14) | 2115; // X.X.101.2115
```

Not sure where the X.X is coming from but it's not reported by Vulkan.
Even Intel on [their page](https://www.intel.com/content/www/us/en/download/762755/intel-6th-10th-gen-processor-graphics-windows.html) only lists it as 101.2115 on the download button and release notes name.

Double checked the scheme [here](https://github.com/SaschaWillems/vulkan.gpuinfo.org/blob/4b05d1dcc4005c591e0e0ea901e1238d9ad0818c/includes/functions.php#L428).

teoxoy (changes requested):

teoxoy (approved):
Good stuff, thanks!

PR #4603, opened 2023-10-28
[wgl] Create a hidden window per instance

Zoxc:
This creates a hidden window per instance instead of of sharing a global one. I was running into issues using multiple OpenGL contexts on a single window in `vger-rs`'s test suite.

### DIFF
--- a/wgpu-hal/src/gles/wgl.rs
+++ b/wgpu-hal/src/gles/wgl.rs
@@ -329,6 +329,7 @@ fn create_instance_device() -> Result<InstanceDevice, crate::InstanceError> {
     // We spawn a thread which owns the hidden window for this instance.
     thread::Builder::new()
         .stack_size(256 * 1024)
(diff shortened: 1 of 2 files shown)


wgpu-hal/src/gles/wgl.rs
```diff
@@ -269,54 +273,131 @@ fn create_global_device_context() -> Result<HDC, crate::InstanceError> {
...
+    }
+
+    let window_class = get_global_window_class()?;
+
+    let (drop_tx, drop_rx) = sync_channel(0);
+    let (setup_tx, setup_rx) = sync_channel(0);
+
+    // We spawn a thread which owns the hidden window for this instance.
+    thread::Builder::new()
+        .stack_size(256 * 1024)
```
cwfitzgerald, line 331: We should probably give this thread a name

cwfitzgerald (approved):
Cursed, I love it

Approved after nit

PR #4605, opened 2023-10-28
bumping `winit` to 0.29.3

TornaxO7:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
None

**Description**
Just bumping `winit` to from `0.29.2` to `0.29.3`

**Testing**
Eh

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -123,7 +123,7 @@ wgpu-example = { version = "0.18.0", path = "./examples/common" }
 wgpu-macros = { version = "0.18.0", path = "./wgpu-macros" }
 wgpu-test = { version = "0.18.0", path = "./tests" }
 wgpu-types = { version = "0.18.0", path = "./wgpu-types" }
-winit = { version = "0.29.3", features = [ "android-native-activity" ] }
+winit = { version = "0.29", features = [ "android-native-activity" ] }
 
 # Metal dependencies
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -123,7 +123,7 @@ wgpu-example = { version = "0.18.0", path = "./examples/common" }
 wgpu-macros = { version = "0.18.0", path = "./wgpu-macros" }
 wgpu-test = { version = "0.18.0", path = "./tests" }
 wgpu-types = { version = "0.18.0", path = "./wgpu-types" }
-winit = { version = "0.29.2", features = [ "android-native-activity" ] }
+winit = { version = "0.29.3", features = [ "android-native-activity" ] }
```
cwfitzgerald: We can leave this at 0.29 - and just update the lock file.

cwfitzgerald (approved):
Nit, but otherwise I'm the future dependabot does this for us, so we don't need to do it manually

Cargo.toml
TornaxO7: so, you mean, I can just remove the third number (here the `3`)?

cwfitzgerald: Yes - it's equivalent to writing 0.29.0

TornaxO7: Done!

PR #4613, opened 2023-10-30
Minspec Devices

### LINKED ISSUE #4582: Expose Ability to Turn Important Private Capabilities Off
Issues like #4581 come up when everyone (including CI) is running on devices that support the latest and greatest. There are issues on older devices which don't support everything. We have multiple paths right now that are completely untested in practice because everyone has new, powerful machines. I only discovered this one as I had an ancient lavapipe installed.

We should expose some way so that the backends can force these features off, so the testing harness can run again with all of it disabled. Additionally we should run with as many _features_ exposed as we can, just private capabilities which cause different code paths to be taken should be disabled.

This might be either a request adapter flag, instance flag, or a method on the device itself.

cwfitzgerald:
**Checklist**

- [ ] Run `cargo clippy`.

**Connections**

Closes https://github.com/gfx-rs/wgpu/issues/4582
This is kinda blocked on https://github.com/taiki-e/cargo-llvm-cov/issues/335

**Description**

**Testing**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -435,6 +435,7 @@ jobs:
 
       - name: upload coverage report to codecov
         uses: codecov/codecov-action@v3
+        if: always() # We want artifacts even if the tests fail.
         continue-on-error: true
         with:
           files: coverage-partition-1.info,coverage-partition-2.info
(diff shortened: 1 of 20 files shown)


cwfitzgerald:
I don't have a ton of motivation to work on this right now and there are other testing priorities. I will revive it if I feel it's relevant.

PR #4624, opened 2023-11-02
Ensure that pipeline creation errors register layouts as errors also

bradwerth:
Since the pipeline id is provided by the caller, the caller may presume that an implicit pipeline layout id is also created, even in error conditions such as with an invalid device. Since our registry system will panic if asked to retrieve a pipeline layout id that has never been registered, it's dangerous to leave that id unregistered. This ensures that the layout ids also get error values when the pipeline creation returns an error.

**Checklist**

- [X] Run `cargo clippy`.

**Connections**

**Description**
Allows user agents to drop the implicit pipeline layouts associated with a pipeline, without worrying about whether or not the registry has created an entry for that layout.

**Testing**
None yet.


bradwerth:
Through some manual testing, found that repeated calls to `insert_error` will panic. Need to use some method of setting values that can tolerate an existing value already being present.

wgpu-core/src/device/global.rs
```diff
@@ -1897,6 +1900,24 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         };
 
         let id = fid.assign_error(desc.label.borrow_or_default(), &mut token);
+
+        // We also need to assign errors to the pipeline layout and the bind
```
nical: nit: "to the *implicit* pipeline layout" it's easy to forget while reading this code that implicit layouts exist.

nical (approved):

Issue #4498, opened 2023-08-30
[hlsl-out] Output is not HLSL 2021 compatible


fornwall:
Naga currently generates hlsl that is not valid [HLSL 2021](https://github.com/microsoft/DirectXShaderCompiler/wiki/HLSL-2021), as there [binary logic operators can only be used with scalars](https://github.com/microsoft/DirectXShaderCompiler/wiki/HLSL-2021#logical-operation-short-circuiting-for-scalars), so we need to change generation:

```diff
-  doit = vec && (ShouldIDoIt(vec) || !ShouldINotDoIt(vec));
+  doit = and(vec, or(ShouldIDoIt(vec), !ShouldINotDoIt(vec)));
```

While possible (changing `countLeadingZeros()` and `select()` writing is enough to pass CI), fxc does not support HLSL 2021 (and the `and()`, `or()` and `select()` intrinsics are only available in HLSL 2021).

In gfx-rs/naga#2447 we worked around the [DX Compiler release for August 2023 ](https://github.com/microsoft/DirectXShaderCompiler/releases/tag/v1.7.2308) making HLSL 2021 the default by specifying `-HV 2018` in CI, [but](https://github.com/gfx-rs/naga/pull/2447#issuecomment-1697897998):

> Ideally we should be forward compatible. [..] I think we are running into this issue only for the select implementation; we can either take the language version as a parameter or inject a polyfill for vector conditions regardless of version.

This issue is tracking that.

teoxoy:
Note that this is only an issue with the `select` built-in as that's the only one getting translated to the ternary op `?:` with a `vecN<bool>` condition.

PR #4629, opened 2023-11-04
Use HLSL 2018

daxpedda:
I found that Naga doesn't support HLSL 2021, which produces some shader errors.
This tells the DXC compiler to use HLSL 2018.

See https://github.com/gfx-rs/wgpu/issues/4498.

### DIFF
--- a/wgpu-hal/src/dx12/shader_compilation.rs
+++ b/wgpu-hal/src/dx12/shader_compilation.rs
@@ -142,9 +142,12 @@ mod dxc {
         log::Level,
     ) {
         profiling::scope!("compile_dxc");
-        let mut compile_flags = arrayvec::ArrayVec::<&str, 4>::new_const();
+        let mut compile_flags = arrayvec::ArrayVec::<&str, 6>::new_const();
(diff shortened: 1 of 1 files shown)


cwfitzgerald (approved):
Thanks!

wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -145,6 +145,9 @@ mod dxc {
         let mut compile_flags = arrayvec::ArrayVec::<&str, 4>::new_const();
```
cwfitzgerald: You need to bump the capacity here too

cwfitzgerald (approved):
Good after nits causing ci failure

wgpu-hal/src/dx12/shader_compilation.rs
daxpedda: Ah, now the CI failure makes sense, thank you!

daxpedda: Done.

PR #4633, opened 2023-11-05
[naga wgsl-in] Introduce `Scalar` type.

jimblandy:
Introduce a new struct `Scalar`, holding a scalar kind and width, and use it as appropriate in the WGSL front end. This consolidates many (kind, width) pairs, and lets us name the two components.

Ideally, `Scalar` would be used throughout Naga, but this would be a large change, touching hundreds of use sites. This patch begins by introducing `Scalar` to the WGSL front end only.

**Checklist**

- [X] Run `cargo clippy`.


teoxoy (approved):
Nice!

PR #4634, opened 2023-11-06
Rewrite Example Event Loop

### LINKED ISSUE #4593: hello-triangle tries to configure with zero size on web
**Description**
When `hello-triangle` starts up under wasm, it configures the surface with `window.inner_size()`, but that size is zero. Furthermore, the WebGPU backend does not catch this validation error.

**Repro steps**

`cargo xtask run-wasm --bin hello-triangle`
`cargo xtask run-wasm --bin hello-triangle --features webgl`

* The example should not configure with zero size. (Is this a winit bug that the canvas size is incorrectly reported?)
* This should be a validation error on all backends, or succeed on all backends. Instead, on WebGPU this does not produce a wgpu validation error, but produces a validation error from the underlying WebGPU, and rendering then succeeds later.

    ```
    The texture size ([Extent3D width:0, height:0, depthOrArrayLayers:1]) or mipLevelCount (1) is empty.
    at ValidateTextureDescriptor (../../third_party/dawn/src/dawn/native/Texture.cpp:478)
    ```

    On WebGL it panics with a validation error inside wgpu.

**Platform**

* Chrome 118.0.5993.88
* macOS `x86_64-apple-darwin`
* wgpu trunk b6fa2c536c75c9da7aecbc5491d346fbc29a336e

### LINKED ISSUE #4627: Winit Updated Caused Redraws Not to Happen
**Description**
request_redraw might not be working, the example only renders one frame.  This also affects my application which is based on the skybox example.  This is probably an Nvidia driver bug because this driver version reworks many Wayland things.

**Repro steps**
`cargo run --bin skybox`

Many frames should be rendered per second, but only one is observed once.

**Extra materials**
None

**Platform**
OS: Archlinux x86-64
Nvidia Driver: 545.29.02
GPU: 3080 Ti
DE: Hyprland

cwfitzgerald:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
Closes https://github.com/gfx-rs/wgpu/issues/4627
Closes https://github.com/gfx-rs/wgpu/issues/4593

**Description**

This just completely redoes our event loop in the examples to:
- Be comprehensible
- Be correct

**Testing**

Manually

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1277,6 +1277,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "fern"
+version = "0.6.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9f0c14694cbd524c8720dd69b0e3179344f04ebb5f90f2e4a440c6ea3b2f1ee"
+dependencies = [
+ "log",
+]
+
+[[package]]
 name = "fixedbitset"
 version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4138,6 +4147,7 @@ dependencies = [
  "console_error_panic_hook",
  "console_log",
  "env_logger",
+ "fern",
  "js-sys",
  "log",
  "png",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4134,6 +4134,7 @@ dependencies = [
 name = "wgpu-example"
 version = "0.18.0"
 dependencies = [
+ "cfg-if",
  "console_error_panic_hook",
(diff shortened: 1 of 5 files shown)


PsichiX:
reminder for myself to test a fix with using canvas client size instead of inner size to set for canvas element size, to make canvas fill entire window area.
related conversation: https://discord.com/channels/676678179678715904/677286494033018924/1170896087020744835
notes to myself: also reproduce canvas setup of how Oxygengine does this, bc this precise combination of style worked in all and edge cases, also test scenario with canvas embedded in other elements.

PR #4635, opened 2023-11-06
Fix Panic in Surface Configure

### LINKED ISSUE #4630: "assertion `left == right` failed: Root token is already active" (wgpu 0.18.0)
After upgrading from wgpu 0.16 to 0.18, my code now prints a panic

```
assertion `left == right` failed: Root token is already active
  left: 0
 right: 1
```

followed by a segmentation fault. Here is a full RUST_BACKTRACE

[PANIC_BACKTRACE.txt](https://github.com/gfx-rs/wgpu/files/13260164/PANIC_BACKTRACE.txt)

### DETAILS/REPRO

I have a project https://github.com/mcclure/webgpu-tutorial-rs . As the name suggests, it is a (close to finished now) WebGPU tutorial application. It is based on the hello-triangle example and attempts to use as many WebGPU features as possible.

As of two days ago, commit [a8e1516](https://github.com/mcclure/webgpu-tutorial-rs/commit/a8e1516a27474275fd6d4364b7d7cf09f79d67f4) was running wgpu 0.16 and winit 0.29-beta (a very early 0.29 beta, so that it had still the 0.28 API surface). Today, I attempted to upgrade. In commit [50c079e](https://github.com/mcclure/webgpu-tutorial-rs/commit/50c079e46f94d22a656cc1b8dc30ee616b1c2a75) I upgraded to winit 0.29 final (while keeping wgpu fixed at 0.16). This had the effect of making the application segfault at close time (although until it quit the application worked). I do not know whether this segfault was caused by wgpu or winit, although the gdb backtrace had symbols for wgpu_hal but *not* winit. Before investigating further, I upgraded to wgpu 0.18 in the most recent commit, [9c1b58](https://github.com/mcclure/webgpu-tutorial-rs/commit/9c1b58b20331abae7e75ead9fbc8b7937c6d1b16). This commit 9c1b58 is the one that sees the "left == right" error. The example is run with simple `cargo run`. **Warning:** When run successfully, this example has audio which currently may be somewhat loud.

I asked about the message in the Matrix and someone named @Wumpf said this was probably bugworthy.

I am running with Rust 1.73.0 on Ubuntu 23.10 with Wayland and an almost entirely stock desktop environment. I am on a "Lenovo ThinkPad T14 Gen 3" with gpu "AMD Ryzen™ 5 PRO 6650U with Radeon™ Graphics × 12". The failure occurs 100% of the time. I have not tested on other platforms.

Because this error occurred after a sloppy upgrade from earlier versions of winit and wgpu, it is likely to me that my code is doing something incorrect. However if so, the error message should be more helpful.

cwfitzgerald:
Closes https://github.com/gfx-rs/wgpu/issues/4630

### DIFF
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -2296,12 +2296,12 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         log::info!("configuring surface with {:?}", config);
 
         let error = 'outer: loop {
-            let hub = A::hub(self);
-            let mut token = Token::root();
-
             // User callbacks must not be called while we are holding locks.
             let user_callbacks;
             {
+                let hub = A::hub(self);
+                let mut token = Token::root();
+
                 let (mut surface_guard, mut token) = self.surfaces.write(&mut token);
                 let (adapter_guard, mut token) = hub.adapters.read(&mut token);
                 let (device_guard, mut token) = hub.devices.read(&mut token);


mcclure:
Can/will this fix get backported to 0.18 branch?

cwfitzgerald:
Yeah

PR #4645, opened 2023-11-06
More complete implementation of "lose the device".

bradwerth:
**Checklist**

- [X] Run `cargo clippy`.

**Connections**
N/A

**Description**
This provides a way for wgpu-core to specify a callback on "lose the device". It ensures this callback is called at the appropriate times: either after device.destroy has empty queues, or on demand from device.lose.

**Testing**
A test has been added to device.rs.


nical (approved):
Nitpick: "callback" instead of "closure", otherwise :+1:

bradwerth:
> Nitpick: "callback" instead of "closure", otherwise 👍

The remaining use of "closure" is in `DeviceLostClosure` which wraps `DeviceLostCallback` and `DeviceLostCallbackC`. I can't rename that to use "callback" without introducing naming redundancy. This naming is also matching the naming pattern established by `SubmittedWorkDoneClosure`. I'm going to leave this as it is.

Issue #4054, opened 2023-08-14
Add `.bits_per_pixel()` to the TextureFormat to simplify buffer size calculation for each texture format.


dtzxporter:
**Is your feature request related to a problem? Please describe.**
It would make it easy to perform some calculations if each format had it's bits per pixel value.

**Describe the solution you'd like**
A method on TextureFormat that returns the bits per pixel for the format: Bc1: 4, rgba8: 32, etc.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

Wumpf:
Since you explicitly mention `Bc1` as 4 bits per pixel, can you outline the usecase a bit?

There's already [block_size](https://docs.rs/wgpu/latest/wgpu/enum.TextureFormat.html#method.block_size) which returns a texel's block size in bytes. So it would be trivial to implement a bit size per pixel based on that (since there's also [block_dimensions](https://docs.rs/wgpu/latest/wgpu/enum.TextureFormat.html#method.block_dimensions), but I fear it could be confusing to put that in for anyone not familiar with the concept of texel blocks 🤔

dtzxporter:
I feel like since wgpu is fairly low level, the point should be to have something like this, especially since you don't have to use it.

Using bits per pixel, its also very easy to calculate the size in bytes of a buffer, for instance, (bpp * width * height) / 8 = size of the buffer required for the texture.

I'm currently using wgpu to perform texture conversion, because there is no rust support for most bc compressed formats, and conversion between other non-compressed formats.

Libraries such as DirectXTex provide this as well.

While I have my own ImageFormat which closely maps DXGI_FORMAT as it has more options, I have to map into wgpu's TextureFormat for instances where it's used as a conversion backend, and it would be useful to have that information at the ready.

JMS55:
You also have to account for both mipmaps and depth/array layers as well. Bevy has some code for this already, but doesn't account for mips. Iirc there's also code that correctly accounts for all of this built into wgpu already, as it uses it to validate create_texture_with_data().

teoxoy:
Bits per pixel can be derived from the existing texel block size and dimensions.

`bits_per_pixel = (block_size * 8) / (block_dimensions.0 * block_dimensions.1)`

Since the texture size has to be a multiple of the texel block dimensions, I'd imagine calculating the size in texel blocks and later multiplying it with the texel block size to get the size of the buffer required for the texture rather than calculating the bits per pixel.

[...]

PR #4647, opened 2023-11-07
rename/deprecate block_size -> block_copy_size, improve docs

Wumpf:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**
* Very loosely related to #4054

**Description**
I keep seeing people getting confused about block_size, so I made its name more explicit and fixed the docs a bit (for instance the link in there was broken)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ For naga changelogs at or before v0.14.0. See [naga's changelog](naga/CHANGELOG.
 #### General
 
 - Log vulkan validation layer messages during instance creation and destruction: By @exrook in [#4586](https://github.com/gfx-rs/wgpu/pull/4586)
+- `TextureFormat::block_size` is deprecated, use `TextureFormat::block_size_in_bytes` instead: By @wumpf in [#4647](https://github.com/gfx-rs/wgpu/pull/4647)
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ For naga changelogs at or before v0.14.0. See [naga's changelog](naga/CHANGELOG.
 #### General
 
 - Log vulkan validation layer messages during instance creation and destruction: By @exrook in [#4586](https://github.com/gfx-rs/wgpu/pull/4586)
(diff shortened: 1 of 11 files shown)


teoxoy:
`texel_block_size` has been renamed to `texel_block_copy_footprint` in the spec which is a lot clearer in what it means. Can we use that instead?

https://github.com/gpuweb/gpuweb/pull/3922

Wumpf:
Oh didn't notice, thank you! Let's go with that then and link to it ofc.
That's a pretty awkward name though imho. Also doesn't imply bytes -.-

Wumpf:
actually.. linking it okay. But do we think anyone will ever find `TextureFormat::block_copy_footprint` if they're looking for a byte size of a texel?

Wumpf:
Technically we are providing the informative value that is described in the spec as

> The texel block memory cost of a [GPUTextureFormat](https://gpuweb.github.io/gpuweb/#enumdef-gputextureformat) is the number of bytes needed to store one [texel block](https://gpuweb.github.io/gpuweb/#texel-block). It is not fully defined for all formats. This value is informative and non-normative.


so from that pov we would need to call it block_memory_cost :/

Wumpf:
moving (potential) discussion to Matrix

teoxoy:
The texel block memory cost is only "informative and non-normative".
The texel block copy footprint is the useful value since it's needed to calculate some of the parameters of `ImageDataLayout`/`ImageCopyBuffer`.

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ For naga changelogs at or before v0.14.0. See [naga's changelog](naga/CHANGELOG.
 #### General
 
 - Log vulkan validation layer messages during instance creation and destruction: By @exrook in [#4586](https://github.com/gfx-rs/wgpu/pull/4586)
+- `TextureFormat::block_size` is deprecated, use `TextureFormat::block_copy_footprint` instead: By @wumpf in [#4647](https://github.com/gfx-rs/wgpu/pull/4647)
```
teoxoy:
```suggestion
- `TextureFormat::block_size` is deprecated, use `TextureFormat::block_copy_size` instead: By @wumpf in [#4647](https://github.com/gfx-rs/wgpu/pull/4647)
```

teoxoy (approved):
LGTM

PR #4648, opened 2023-11-07
[wgsl-in] Remove unused expressions from Module::const_expressions.

jimblandy:
After building a Naga module from the WGSL AST, make a pass over the module to determine which expressions in `Module::const_expressions` are actually used, and compact out all unused expressions from that arena, adjusting references appropriately.

This isn't necessary on trunk, but it will be when we add abstract types to the WGSL front end. The evaluation of expressions containing abstract values will naturally introduce lots of 64-bit float and integer literals to the constant arena, which the validator would rightly reject. However, because all abstract values are converted to concrete types before appearing as part of the module's final code, all such 64-bit expressions should be unused.

**Checklist**

- [X] Run `cargo clippy`.


jimblandy:
@teoxoy Since this isn't actually needed on trunk, should I put this on a separate branch in the repo, as we discussed doing for the override work?

jimblandy:
After talking about this, @teoxoy pointed out that we are going to have to compact function expression arenas as well. Since it's harder to find the roots for those, it's actually probably not possible to be much simpler than the main compaction algorithm in `crate::compact`, and the WGSL front end should just call that.

The optimizations to avoid a stack / recursion I use in this PR could apply to the main compactor, though, and those should probably be applied.

PR #4649, opened 2023-11-07
Fix Panic in Surface Configure (#4635) (0.18 backport)

mcclure:
Am I doing this right
#4635 has a "PR: Needs back-porting" tag
This is #4635 grafted (hg graft, not git cherry-pick, may not work the same way, I don't know, github marks the commit as "unverified" which makes sense because it's not really by @cwfitzgerald) onto the 0.18 branch

**Checklist**

- [X] Run `cargo clippy`.

**Testing**
I tested it with the code that was crashing in #4630 and it did not crash

### DIFF
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -2296,12 +2296,12 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         log::info!("configuring surface with {:?}", config);
 
         let error = 'outer: loop {
-            let hub = A::hub(self);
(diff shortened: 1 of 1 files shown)


Wumpf:
> Am I doing this right

Almost, you targeted the wrong branch in the github ui - this PR shouldn't go to `trunk` (the change is already there!) but instead to https://github.com/gfx-rs/wgpu/tree/v0.18

Wumpf:
(fixed that for you, just waiting for CI now)

Wumpf (approved):

PR #647, opened 2020-05-06
Add a way to destroy a pass by a mutable reference

kvark:
This will make https://github.com/gfx-rs/wgpu-rs/issues/299 much easier.

PR #153, opened 2019-05-07
Update deps + upgrade examples to rust 2018

rukai:
Things to note:
*   I removed the Cargo.lock - I assume it was only accidentally checked in?
*   I dont know what `wgpu-remote` is for, so I haven't tested it beyond compiling it.

PR #4655, opened 2023-11-08
vk: Only support bgra8unorm_storage if STORAGE_READ_WITHOUT_FORMAT is supported

sethdusek:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

**Description**
On Intel 6th generation GPUs, a simple compute shader that reads from a bgra8unorm texture and writes to a bgra8unorm texture does not work. Upon further inspection, writes to the texture were working but reading wasn't. After running it through renderdoc with validation layers I get the following error: https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdDispatch-OpTypeImage-07028

It seems my Intel GPU supports  VK_FORMAT_FEATURE_2_STORAGE_WRITE_WITHOUT_FORMAT but does not support VK_FORMAT_FEATURE_2_STORAGE_READ_WITHOUT_FORMAT. 

This PR only enables BGRA8UNORM_STORAGE if both of these are present.

However I do not understand why wgpu requires both WRITE_WITHOUT_FORMAT and READ_WITHOUT_FORMAT in the first place. I've tried specifying the format of the texture when creating the TextureView that gets passed to the compute shader but to no avail, as I get the same validation errors (and an empty screen since nothing is being read). It seems wgpu is ignoring these view formats somehow and thus requires both of these WITHOUT_FORMAT extensions?

**Testing**
I've tested this patch on an Intel HD Graphics 520. Now it does not support BGRA8UNORM_STORAGE, as it should, instead of spitting out validation errors later. It works as expected on any normal GPU that supports both these features.

> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: RequestDeviceError { inner: Core(UnsupportedFeature(Features(BGRA8UNORM_STORAGE))) }', src/wgpu_renderer.rs:167:14

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,8 @@ For naga changelogs at or before v0.14.0. See [naga's changelog](naga/CHANGELOG.
 
 - Fix issue where local variables were sometimes using variable names from previous functions.
 
+#### Vulkan
+- Only expose support for BGRA8UNORM_STORAGE if STORAGE_READ_WITHOUT_FORMAT is supported: By @sethdusek in [#4655](https://github.com/gfx-rs/wgpu/pull/4655)
 ## v0.18.0 (2023-10-25)
 
 ### Desktop OpenGL 3.3+ Support on Windows
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -1776,5 +1776,6 @@ fn supports_bgra8unorm_storage(
 
         features2.contains(vk::FormatFeatureFlags::STORAGE_IMAGE)
             && features3.contains(vk::FormatFeatureFlags2::STORAGE_WRITE_WITHOUT_FORMAT)
+            && features3.contains(vk::FormatFeatureFlags2::STORAGE_READ_WITHOUT_FORMAT)
     }
 }


Wumpf (dismissed):
> However I do not understand why wgpu requires both WRITE_WITHOUT_FORMAT and READ_WITHOUT_FORMAT in the first place.

The feature flag enabled here corresponds to https://www.w3.org/TR/webgpu/#bgra8unorm-storage which requires both read & write storage access. So I think you fix here makes perfectly sense.
If we want to break this down, we'd have to add a wgpu-native only feature for this.

> I've tried specifying the format of the texture when creating the TextureView that gets passed to the compute shader but to no avail, as I get the same validation errors (and an empty screen since nothing is being read). It seems wgpu is ignoring these view formats somehow and thus requires both of these WITHOUT_FORMAT extensions?

The only "conversion" a texture view can do right now is adding/removing `srgb` flag https://www.w3.org/TR/webgpu/#dom-gputexturedescriptor-viewformats
I'm not 100% sure on this part, but texture views aren't really relevant for storage textures at all, only about non-storage texture access, i.e. when you access via `texture_2d` instead of `texture_storage_2d`
(worth noting that `texture_2d` should always be supported for bgra8)

Wumpf:
Can you add a changelog line? That's still a bugfix worth calling out :)

teoxoy:
Can we hold off for a bit on merging this?

`bgra8unorm-storage` was at first only targeted at `write-only` storage textures.

The PR that added `read-only` and `read-write` textures to the spec updated the table based on the findings in https://github.com/gpuweb/gpuweb/issues/3838#issuecomment-1761399085.

But the situation of `bgra8unorm-storage` is different from all the other formats because we need to check for `STORAGE_WRITE_WITHOUT_FORMAT`/`STORAGE_READ_WITHOUT_FORMAT` rather than `STORAGE_IMAGE`.

I'll look some more into what would be best here and open a spec issue.

Wumpf:
thanks for the context @teoxoy, had no idea ofc 😄

teoxoy:
Based on the data from https://vulkan.gpuinfo.org/ it seems that requiring `STORAGE_READ_WITHOUT_FORMAT` on top of `STORAGE_WRITE_WITHOUT_FORMAT` doesn't lose additional devices.

https://github.com/teoxoy/gpuinfo-vulkan-query/blob/87708bea8170b9fcf9dec0afb5f25c9e5b302f48/bgra8unorm-storage.txt#L1348

Which contradicts what @SethDusek experienced (Intel HD Graphics 520 supporting `STORAGE_WRITE_WITHOUT_FORMAT` but not `STORAGE_READ_WITHOUT_FORMAT`).

Not sure what to make of this.

teoxoy:
> However I do not understand why wgpu requires both WRITE_WITHOUT_FORMAT and READ_WITHOUT_FORMAT in the first place. I've tried specifying the format of the texture when creating the TextureView that gets passed to the compute shader but to no avail, as I get the same validation errors (and an empty screen since nothing is being read). It seems wgpu is ignoring these view formats somehow and thus requires both of these WITHOUT_FORMAT extensions?

We only require `STORAGE_WRITE_WITHOUT_FORMAT`/`STORAGE_READ_WITHOUT_FORMAT` for `bgra8unorm` because we must use SPIR-V's `Unknown` format (since there is no `bgra8` format in SPIR-V).

For more background on this see: https://github.com/KhronosGroup/Vulkan-Docs/issues/2027#issuecomment-1380608011.

sethdusek:
> Which contradicts what @SethDusek experienced (Intel HD Graphics 520 supporting STORAGE_WRITE_WITHOUT_FORMAT but not STORAGE_READ_WITHOUT_FORMAT).
> Not sure what to make of this.

From what I can tell reading ANV source code (Intel driver from Mesa), it only enables support for READ_WITHOUT_FORMAT if `isl_format_supports_typed_reads` is true for said format. `isl_format_typed_reads` and `isl_format_typed_writes` both read from a table which indicates which versions of Intel GPUs support reading/writing from this format

The table says that versions of Intel GPUs > 125 support "typed reads" for B8R8G8A8_UNORM but versions of Intel GPUs > 70 support typed writes (I'm not sure what exactly these version numbers correspond to among generations of Intel GPUs), but this indicates there are certain GPUs that can write to a B8G8R8A8_Unorm image but can't read from them. Which would explain why I don't get validation errors binding the image as write-only but only get it when binding it read-only. 

Relevant files in mesa:
- [src/intel/isl/isl_format.c line #153](https://gitlab.freedesktop.org/mesa/mesa/-/blob/424df6a68d286171d14886d0383685ea65cca1bc/src/intel/isl/isl_format.c#L153)
- [src/intel/vulkan/anv_formats.c line #647](https://gitlab.freedesktop.org/mesa/mesa/-/blob/424df6a68d286171d14886d0383685ea65cca1bc/src/intel/vulkan/anv_formats.c#L647)

Output of `vulkaninfo --show-formats`: https://gist.github.com/SethDusek/a2ac08483670f7019db350ec1e51923d

teoxoy:
That makes sense, thanks for looking into it and sharing those resources!

The `vulkaninfo` CLI also correctly (based on the driver src code) reports it https://gist.github.com/SethDusek/a2ac08483670f7019db350ec1e51923d#file-vulkaninfo-L2370.

The remaining mystery is why does the vulkaninfo DB only contain reports where either both or none of those capabilities show up.

teoxoy:
Could you share the output of `vulkaninfo --json`?

sethdusek:
[VP_VULKANINFO_Intel(R)_HD_Graphics_520_(SKL_GT2)_23_2_1.json](https://github.com/gfx-rs/wgpu/files/13322356/VP_VULKANINFO_Intel.R._HD_Graphics_520_.SKL_GT2._23_2_1.json)

teoxoy:
It turns out that the reports at https://vulkan.gpuinfo.org/ don't contain `VkFormatFeatureFlagBits2` info https://github.com/SaschaWillems/VulkanCapsViewer/issues/194.

I got tripped up by https://github.com/SaschaWillems/vulkan.gpuinfo.org/issues/50#issuecomment-1806517087.

Because bit 31 (`STORAGE_READ_WITHOUT_FORMAT`) was originally set, it was causing all other bits after it to also be set, which included bit 32 (`STORAGE_WRITE_WITHOUT_FORMAT`).

That's why we were seeing either both or none of them set.

That was quite the rabbit hole.

teoxoy:
Given the current requirements for `bgra8unorm` with `STORAGE_BINDING` usage on underlying APIs:

||Vulkan|D3D12 FL11_0|Metal|
|-|-|-|-|
|write-only|`STORAGE_WRITE_WITHOUT_FORMAT`|"UAV Typed Store"|:white_check_mark:|
|read-only|`STORAGE_READ_WITHOUT_FORMAT`|:white_check_mark: (via SRV)|:white_check_mark:|
|read-write|`STORAGE_WRITE_WITHOUT_FORMAT` & `STORAGE_READ_WITHOUT_FORMAT`|"UAV Typed Store" & "UAV Typed Load"|:x:|

I think the `bgra8unorm-storage` feature should only allow write-only access. Since read-write access is not possible on Metal and requiring `STORAGE_READ_WITHOUT_FORMAT` & "UAV Typed Load" might lower the availability of the feature on some hardware/drivers.

I'll open an issue on the spec to see what the others think.

We also need to add some more validation for this.

cwfitzgerald (changes requested):
Requesting changes per above comments.

Please re-request a review from teo once the changes are addressed to make sure they see it!

teoxoy:
`bgra8unorm-storage` will most likely only cover `write-only` access (https://github.com/gpuweb/gpuweb/issues/4377) and the spec needs updating.

I filed https://github.com/gfx-rs/wgpu/issues/4704 to track the new changes to the spec (which should cover the validation I was mentioning before).

PR #4662, opened 2023-11-10
[naga] Improve algorithm for Module compaction.

### LINKED ISSUE #4509: Compaction could be faster
Caution: I don't actually have any benchmarks here. So this would need to be undertaken along with adding a compaction benchmark to our `criterion` benchmark suite.

If compaction took advantage of the fact that expressions, constants, and types only refer to elements *earlier* in the arena than they are, it could trace these by making a single pass from end to start, marking (say) an expression's operands as used only if the expression itself is marked used.

To trace a function, you'd first trace the statement tree, but for each expression the statement tree refers to, you'd simply mark it used - you would not call `trace_expression`.

Then, after the statement tree has been traced, you'd make a single pass over the expression arena from back to front, marking an expression's operands as used only if the expression itself had been marked as used previously. This would require no work list manipulation.

A similar approach could be extended to the constants, constant expressions, and types.

Unlike the current algorithm, this would require traversing both live and dead arena elements. However, that traversal would be guided by the contents of the `HandleSet`s, so that we wouldn't ever actually touch the elements we didn't intend to traverse.

jimblandy:
Identify reachable function expressions, constant expressions, and types using a single pass over each arena, taking advantage of the fact that expressions and types may only refer to other entries that precede them within their arena. Only walking the statement tree still requires a worklist/recursion.

In addition to presumably being faster, this change slightly reduces the number of non-comment lines of code in `src/compact`.

Closes https://github.com/gfx-rs/wgpu/issues/4509.

**Checklist**

- [X] Run `cargo clippy`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,7 +54,11 @@ For naga changelogs at or before v0.14.0. See [naga's changelog](naga/CHANGELOG.
 
 - Create a hidden window per `wgpu::Instance` instead of sharing a global one.
 
-#### Naga MSL-OUT
+#### Naga
+
+- Improve algorithm used by module compaction. By @jimblandy in [#4662](https://github.com/gfx-rs/wgpu/pull/4662).
+
+##### MSL-OUT
 
 - Fix issue where local variables were sometimes using variable names from previous functions.
 
--- a/naga/src/compact/expressions.rs
+++ b/naga/src/compact/expressions.rs
@@ -1,8 +1,7 @@
 use super::{HandleMap, HandleSet, ModuleMap};
-use crate::arena::{Arena, Handle, UniqueArena};
+use crate::arena::{Arena, Handle};
 
 pub struct ExpressionTracer<'tracer> {
-    pub types: &'tracer UniqueArena<crate::Type>,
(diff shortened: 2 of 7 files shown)


teoxoy (approved):
Good stuff!

teoxoy:
I noticed the prior PR [changed the IR tests](https://github.com/gfx-rs/wgpu/pull/4648/files#diff-31b775351e47744ad64b48ddf68294a6857b047a75e3c8bf5ffc5eef42cb534e) but this one doesn't.
Was this expected? I haven't looked too deeply what the differences between the code are.

jimblandy:
> I noticed the prior PR [changed the IR tests](https://github.com/gfx-rs/wgpu/pull/4648/files#diff-31b775351e47744ad64b48ddf68294a6857b047a75e3c8bf5ffc5eef42cb534e) but this one doesn't. Was this expected? I haven't looked too deeply what the differences between the code are.

Yes, this is expected: this PR only improves the algorithm, but it doesn't yet apply it in any new places. The prior PR introduced a new compactor specific to the WGSL front end, and added uses of it.

PR #4678, opened 2023-11-13
Keep the value in its storage after destroy

### LINKED ISSUE #4668: Commit #4657 appears to introduce a race condition
Since syncing yesterday, I started getting consistent panics here:

```
thread 'main' panicked at D:\projects\rust\wgpu\wgpu-core\src\track\buffer.rs:324:17:
Tried to insert buffer already tracked
```

in my code where I shut down wgpu buffers and immediately create new ones. I tracked it down to commit [#4657](https://github.com/gfx-rs/wgpu/pull/4657).

Looking at the changes, my best guess is that calls to `take_and_mark_destroyed` such as this one in `D:\projects\rust\wgpu\wgpu-core\src\device\global.rs`:

```
            //TODO: lock pending writes separately, keep the device read-only
            let (mut device_guard, mut token) = hub.devices.write(&mut token);

            log::trace!("Buffer::destroy {buffer_id:?}");
            let (mut buffer_guard, _) = hub.buffers.write(&mut token);
            let mut buffer = buffer_guard
                .take_and_mark_destroyed(buffer_id)
                .map_err(|_| resource::DestroyError::Invalid)?;

            let device = &mut device_guard[buffer.device_id.value];
```

are allowing ids to be reused before the corresponding async GPU resources are released. This seems to be validated by the fact that I can make the panic disappear by waiting at a breakpoint before the id is reused.

I am reverting to before this commit until the problem can be properly diagnosed by someone "in the know" - please let me know if I can provide any further help!

Thanks!

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Fixes #4668

**Description**

in #4657 the destroy implementation was made to remove the value from the storage and leave an error variant in its place. Unfortunately this causes some issues with the buffer tracking code which expects the ID to be unregistered after the value has been fully destroyed, even if the latter does not need to be in storage anymore. To work around that, this commit adds a `Destroyed` variant in storage which keeps the value so that the previous tracking behavior is preserved while still making sure that most accesses to the destroyed resource lead to validation errors.

... Except for submitted command buffers that need to be consumed right away. These are replaced with `Element::Error` like before this commit.

It's a fair bit unsavory to have the two ways to do the same thing in principle, but a better solution will take more time to bake and would conflict unnecessarily with the arcanization work.

**Testing**

An existing test was altered to cover the regression.

### DIFF
--- a/wgpu-core/src/identity.rs
+++ b/wgpu-core/src/identity.rs
@@ -66,7 +66,6 @@ impl IdentityManager {
 
     /// Free `id`. It will never be returned from `alloc` again.
     pub fn free<I: id::TypedId + Debug>(&mut self, id: I) {
-        println!("free id");
         let (index, epoch, _backend) = id.unzip();
         let pe = &mut self.epochs[index as usize];
         assert_eq!(*pe, epoch);
--- a/wgpu-core/src/identity.rs
+++ b/wgpu-core/src/identity.rs
@@ -53,7 +53,6 @@ impl IdentityManager {
     /// The backend is incorporated into the id, so that ids allocated with
     /// different `backend` values are always distinct.
     pub fn alloc<I: id::TypedId>(&mut self, backend: Backend) -> I {
-        println!("alloc id");
         match self.free.pop() {
             Some(index) => I::zip(index, self.epochs[index as usize], backend),
             None => {
--- a/wgpu-core/src/identity.rs
+++ b/wgpu-core/src/identity.rs
@@ -53,6 +53,7 @@ impl IdentityManager {
     /// The backend is incorporated into the id, so that ids allocated with
     /// different `backend` values are always distinct.
     pub fn alloc<I: id::TypedId>(&mut self, backend: Backend) -> I {
+        println!("alloc id");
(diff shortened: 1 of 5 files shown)


wgpu-core/src/identity.rs
```diff
@@ -53,6 +53,7 @@ impl IdentityManager {
     /// The backend is incorporated into the id, so that ids allocated with
     /// different `backend` values are always distinct.
     pub fn alloc<I: id::TypedId>(&mut self, backend: Backend) -> I {
+        println!("alloc id");
```
teoxoy:
```suggestion
```

wgpu-core/src/identity.rs
```diff
@@ -66,6 +67,7 @@ impl IdentityManager {
 
     /// Free `id`. It will never be returned from `alloc` again.
     pub fn free<I: id::TypedId + Debug>(&mut self, id: I) {
+        println!("free id");
```
teoxoy:
```suggestion
```

teoxoy (approved):
LGTM

Issue #4484, opened 2023-03-13
[meta] const/override-expressions


teoxoy:
const-expressions phase 1 (IR architecture changes) https://github.com/gfx-rs/naga/pull/2266:
- https://github.com/gfx-rs/naga/issues/1956
- https://github.com/gfx-rs/naga/issues/1957

const-expressions phase 2 (const evaluation + support for const-expressions in remaining places (attributes and switch cases)) https://github.com/gfx-rs/naga/pull/2309:
- https://github.com/gfx-rs/naga/issues/97
- https://github.com/gfx-rs/naga/issues/1829
- https://github.com/gfx-rs/naga/issues/733
- https://github.com/gfx-rs/naga/issues/471
- https://github.com/gfx-rs/naga/issues/2430
- https://github.com/gfx-rs/naga/issues/2106
- https://github.com/gfx-rs/naga/issues/2313

override-expressions:
- https://github.com/gfx-rs/wgpu/issues/4366 (+ https://github.com/gfx-rs/wgpu/issues/1762)
- https://github.com/gfx-rs/wgpu/issues/4450
- https://github.com/gfx-rs/wgpu/issues/5315

full abstract types:
- https://github.com/gfx-rs/wgpu/issues/4400
- https://github.com/gfx-rs/wgpu/issues/4492
- https://github.com/gfx-rs/wgpu/issues/4435

misc (future work):
- https://github.com/gfx-rs/wgpu/issues/4493
- https://github.com/gfx-rs/wgpu/issues/4507

jimblandy:
Now that #4400 is closed, shall we close this as well? The only remaining checklist item is #4507, and that's a meta-bug in its own right.

ErichDonGubler:
Yes please! 👍🏻

PR #4677, opened 2023-11-13
add pipeline constants plumbing

teoxoy:
Adds pipeline constants plumbing to the `pipeline-constants` feature branch.

Tracking meta issue: https://github.com/gfx-rs/wgpu/issues/4484.

PR #4683, opened 2023-11-14
add pipeline constants plumbing

teoxoy:
Adds pipeline constants plumbing to the `pipeline-constants` feature branch.

### DIFF
--- a/examples/water/src/main.rs
+++ b/examples/water/src/main.rs
@@ -512,6 +512,7 @@ impl wgpu_example::framework::Example for Example {
             vertex: wgpu::VertexState {
                 module: &water_module,
                 entry_point: "vs_main",
+                constants: &Default::default(),
                 // Layout of our vertices. This should match the structs
                 // which are uploaded to the GPU. This should also be
                 // ensured by tagging on either a `#[repr(C)]` onto a
@@ -527,6 +528,7 @@ impl wgpu_example::framework::Example for Example {
             fragment: Some(wgpu::FragmentState {
                 module: &water_module,
                 entry_point: "fs_main",
+                constants: &Default::default(),
                 // Describes how the colour will be interpolated
                 // and assigned to the output attachment.
                 targets: &[Some(wgpu::ColorTargetState {
@@ -581,6 +583,7 @@ impl wgpu_example::framework::Example for Example {
             vertex: wgpu::VertexState {
                 module: &terrain_module,
                 entry_point: "vs_main",
+                constants: &Default::default(),
                 buffers: &[wgpu::VertexBufferLayout {
                     array_stride: terrain_vertex_size as wgpu::BufferAddress,
                     step_mode: wgpu::VertexStepMode::Vertex,
@@ -590,6 +593,7 @@ impl wgpu_example::framework::Example for Example {
             fragment: Some(wgpu::FragmentState {
                 module: &terrain_module,
                 entry_point: "fs_main",
+                constants: &Default::default(),
                 targets: &[Some(config.view_formats[0].into())],
             }),
             primitive: wgpu::PrimitiveState {
--- a/naga-cli/src/bin/naga.rs
+++ b/naga-cli/src/bin/naga.rs
@@ -529,6 +529,7 @@ fn write_output(
                     pipeline_options_owned = spv::PipelineOptions {
                         entry_point: name.clone(),
                         shader_stage: module.entry_points[ep_index].stage,
+                        constants: naga::back::PipelineConstants::default(),
                     };
                     Some(&pipeline_options_owned)
                 }
@@ -569,6 +570,7 @@ fn write_output(
                     _ => unreachable!(),
                 },
                 multiview: None,
+                constants: naga::back::PipelineConstants::default(),
             };
 
             let mut buffer = String::new();
@@ -604,6 +606,7 @@ fn write_output(
                         "Generating hlsl output requires validation to \
                          succeed, and it failed in a previous step",
                     ))?,
+                    &hlsl::PipelineOptions::default(),
                 )
                 .unwrap_pretty();
             fs::write(output_path, buffer)?;
--- a/naga/src/back/msl/mod.rs
+++ b/naga/src/back/msl/mod.rs
@@ -221,7 +221,7 @@ impl Default for Options {
 }
 
 /// A subset of options that are meant to be changed per pipeline.
-#[derive(Debug, Default, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Default, Clone)]
 #[cfg_attr(feature = "serialize", derive(serde::Serialize))]
 #[cfg_attr(feature = "deserialize", derive(serde::Deserialize))]
 pub struct PipelineOptions {
@@ -232,6 +232,8 @@ pub struct PipelineOptions {
     ///
     /// Enable this for vertex shaders with point primitive topologies.
     pub allow_and_force_point_size: bool,
+    /// Pipeline constants.
+    pub constants: crate::back::PipelineConstants,
 }
 
 impl Options {
--- a/tests/tests/scissor_tests/mod.rs
+++ b/tests/tests/scissor_tests/mod.rs
@@ -38,16 +38,18 @@ fn scissor_test_impl(ctx: &TestingContext, scissor_rect: Rect, expected_data: [u
             label: Some("Pipeline"),
             layout: None,
             vertex: wgpu::VertexState {
-                entry_point: "vs_main",
                 module: &shader,
+                entry_point: "vs_main",
+                constants: &Default::default(),
                 buffers: &[],
             },
             primitive: wgpu::PrimitiveState::default(),
             depth_stencil: None,
             multisample: wgpu::MultisampleState::default(),
             fragment: Some(wgpu::FragmentState {
-                entry_point: "fs_main",
                 module: &shader,
+                entry_point: "fs_main",
+                constants: &Default::default(),
                 targets: &[Some(wgpu::ColorTargetState {
                     format: wgpu::TextureFormat::Rgba8Unorm,
                     blend: None,
--- a/wgpu-core/src/pipeline.rs
+++ b/wgpu-core/src/pipeline.rs
@@ -164,6 +164,12 @@ pub struct ProgrammableStageDescriptor<'a> {
     /// The name of the entry point in the compiled shader. There must be a function with this name
     /// in the shader.
     pub entry_point: Cow<'a, str>,
+    /// Specifies the values of pipeline-overridable constants in the shader module.
+    ///
+    /// The key represents the numeric ID of the constant, if one is specified, and otherwise the constant's identifier name.
+    ///
+    /// The value may represent any of WGSL's concrete scalar types.
+    pub constants: Cow<'a, naga::back::PipelineConstants>,
 }
 
 /// Number of implicit bind groups derived at pipeline creation.
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -15,6 +15,7 @@ mod macros;
 use std::{
     any::Any,
     borrow::Cow,
+    collections::HashMap,
     error, fmt,
     future::Future,
     marker::PhantomData,
@@ -1465,6 +1466,12 @@ pub struct VertexState<'a> {
     /// The name of the entry point in the compiled shader. There must be a function with this name
     /// in the shader.
     pub entry_point: &'a str,
+    /// Specifies the values of pipeline-overridable constants in the shader module.
+    ///
+    /// The key represents the numeric ID of the constant, if one is specified, and otherwise the constant's identifier name.
+    ///
+    /// The value may represent any of WGSL's concrete scalar types.
+    pub constants: &'a HashMap<String, f64>,
     /// The format of any vertex buffers used with this pipeline.
     pub buffers: &'a [VertexBufferLayout<'a>],
 }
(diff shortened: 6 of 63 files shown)


examples/water/src/main.rs
```diff
@@ -580,7 +582,7 @@ impl wgpu_example::framework::Example for Example {
             layout: Some(&terrain_pipeline_layout),
             vertex: wgpu::VertexState {
                 module: &terrain_module,
-                entry_point: "vs_main",
+                entry_point: "vs_main",constants: &Default::default(),
```
ErichDonGubler: issue: The lack of spacing b/w these fields in this list make is seem like `rustfmt` either hasn't been run, or is struggling, for some reason. Both are 🙅🏻 no good!

naga-cli/src/bin/naga.rs
```diff
@@ -529,6 +529,7 @@ fn write_output(
                     pipeline_options_owned = spv::PipelineOptions {
                         entry_point: name.clone(),
                         shader_stage: module.entry_points[ep_index].stage,
+                        constants: naga::back::PipelineConstants::default(),
```
ErichDonGubler, line 532: question: Why not use `Default::default()` like in other parts of the commit? 🤔

This question also applies to changes in `naga/benche/criterion.rs`.

tests/tests/scissor_tests/mod.rs
```diff
@@ -38,16 +38,18 @@ fn scissor_test_impl(ctx: &TestingContext, scissor_rect: Rect, expected_data: [u
             label: Some("Pipeline"),
             layout: None,
             vertex: wgpu::VertexState {
-                entry_point: "vs_main",
                 module: &shader,
+                entry_point: "vs_main",
```
ErichDonGubler, line 42: nitpick: It'd be nice to not move the  `entry_point` field here; that only seems like diff noise. This seems to happen in a few other tests as well.

wgpu-core/src/pipeline.rs
```diff
@@ -164,6 +164,12 @@ pub struct ProgrammableStageDescriptor<'a> {
     /// The name of the entry point in the compiled shader. There must be a function with this name
     /// in the shader.
     pub entry_point: Cow<'a, str>,
+    /// Specifies the values of pipeline-overridable constants in the shader module module.
+    ///
+    /// The key represents the numeric ID of the constant, if one is specified, and otherwise the constant's identifier name.
+    ///
+    /// The value may represent any of WGSL's concrete scalar types.
+    pub constants: Cow<'a, naga::back::PipelineConstants>,
```
ErichDonGubler, line 172: question: Is `pipeline_constants` overly verbose in this context? I'm not too familiar with how a “programmable stage” and a pipeline are linked, so I might not realize that it should be fine to elide the “pipeline” context.

This question applies to other places where this identifer is used for a new field in like manner.

wgpu/src/lib.rs
```diff
@@ -1465,6 +1466,12 @@ pub struct VertexState<'a> {
     /// The name of the entry point in the compiled shader. There must be a function with this name
     /// in the shader.
     pub entry_point: &'a str,
+    /// Specifies the values of pipeline-overridable constants in the shader module module.
```
ErichDonGubler: question: `module module` seems wrong; should this be just `module`?

This also happens in a few other places.

wgpu/src/lib.rs
```diff
@@ -1465,6 +1466,12 @@ pub struct VertexState<'a> {
...
+    ///
+    /// The key represents the numeric ID of the constant, if one is specified, and otherwise the constant's identifier name.
```
ErichDonGubler, line 1471: > The key represents the numeric ID of the constant, if one is specified, and otherwise the constant's identifier name.

I don't think I understand this. Does this mean that the key might be one of (1) `String` containing ASCII numeric digits _or_ (2) a typical C-style identifier like might be written in a shader?

Doing further reading in ``7.2.2 `override` Declaration`` from [the spec.](https://gpuweb.github.io/gpuweb/wgsl/#override-decls), it seems that option (1) corresponds to the `id(…)` attribute:

> If the declaration has an [id](https://gpuweb.github.io/gpuweb/wgsl/#attribute-id) attribute applied, the literal operand is known as the pipeline constant ID.
>
> … The constant is identified by a pipeline-overridable constant identifier string[](https://gpuweb.github.io/gpuweb/wgsl/#pipeline-overridable-constant-identifier-string), which is the base-10 representation of the [pipeline constant ID](https://gpuweb.github.io/gpuweb/wgsl/#pipeline-constant-id) if specified, and otherwise the declared [name](https://gpuweb.github.io/gpuweb/wgsl/#name) of the constant.

However, that seems like `String` would be the wrong type for this, since we'd have to parse the ID as a `u16` anyway for validation, right? Am I missing something?

---

This question applies to other places where this doc has been copy-pasted.

naga/src/back/msl/mod.rs
```diff
@@ -221,7 +221,7 @@ impl Default for Options {
 }
 
 /// A subset of options that are meant to be changed per pipeline.
-#[derive(Debug, Default, Clone, PartialEq, Eq, Hash)]
```
ErichDonGubler, line 224: question: [You shouldn't have to remove `PartialEq, Eq` here on account of a `HashMap`.](https://doc.rust-lang.org/std/collections/struct.HashMap.html#impl-Eq-for-HashMap%3CK,+V,+S%3E) Is the intention of removing this now to be forwards-compatible with an arbitrary choice of map-like data structure down the road, or something?

This question also applies to other places where these `derive`s have been removed in like manner.

wgpu/src/lib.rs
```diff
@@ -1465,6 +1466,12 @@ pub struct VertexState<'a> {
...
+    ///
+    /// The value may represent any of WGSL's concrete scalar types.
+    pub constants: &'a HashMap<String, f64>,
```
ErichDonGubler, line 1474: nitpick: Is there a strong reason to not reexport and use `naga::PipelineConstants` here?

This question applies to other places where this doc has been copy-pasted.

wgpu/src/lib.rs
```diff
@@ -1465,6 +1466,12 @@ pub struct VertexState<'a> {
     /// The name of the entry point in the compiled shader. There must be a function with this name
     /// in the shader.
     pub entry_point: &'a str,
+    /// Specifies the values of pipeline-overridable constants in the shader module module.
+    ///
+    /// The key represents the numeric ID of the constant, if one is specified, and otherwise the constant's identifier name.
+    ///
+    /// The value may represent any of WGSL's concrete scalar types.
```
ErichDonGubler, line 1473: question: I doubt it, but: is it worth linking to `Scalar` docs here?

This question applies to other places where this doc has been copy-pasted.

ErichDonGubler (changes requested):
Only [one hard (potential?) issue I see with `rustfmt`](https://github.com/gfx-rs/wgpu/pull/4683#discussion_r1393543545); otherwise, I have a bunch of nitpicks and questions, but nothing that I think is a hard blocker.

jimblandy (changes requested):
This definitely needs a big honkin' `CHANGELOG.md` entry, since it breaks pretty much everybody.

examples/water/src/main.rs
teoxoy: I was relying on `rustfmt` to take care of the formatting on save but it seems it's not working in that specific code block due to the call to `wgpu::vertex_attr_array!`.

wgpu-core/src/pipeline.rs
teoxoy: I think it would be too verbose, we are always inside a pipeline context where this is defined.

naga-cli/src/bin/naga.rs
teoxoy: I wanted to be a bit more verbose in naga since we have other "constants" but for `wgpu` and friends, `Default::default()` seemed fine.

wgpu/src/lib.rs
teoxoy: I think it would be easier to look at `wgpu`'s docs and see the actual type rather than the alias.

`naga` is also internal to `wgpu-core`, most `wgpu` users shouldn't have to interact with it (only if they want to pass a `wgpu::ShaderSource::Naga`).

naga/src/back/msl/mod.rs
teoxoy: I had to remove `Eq` (due to the `f64`) and `Hash` (due to the `HashMap`). I highly doubt anyone was relying on these anyway (they were most likely copy-pasted from the `Options` structs).

tests/tests/scissor_tests/mod.rs
teoxoy: I thought this is small enough to be fine. These fields are always in that order in the other 80-90% of cases and I fixed the remaining ones.

wgpu/src/lib.rs
teoxoy: > if specified, and otherwise the declared [name](https://gpuweb.github.io/gpuweb/wgsl/#name) of the constant.

They key must be the [pipeline constant ID](https://gpuweb.github.io/gpuweb/wgsl/#pipeline-constant-id) (`u16`) if one was specified via the `@id()` attribute or the [name](https://gpuweb.github.io/gpuweb/wgsl/#name)/[identifier](https://gpuweb.github.io/gpuweb/wgsl/#identifier) of the pipeline-overridable constant.

Since the `key` is a `String` in the IDL, we keep it as such until we reach naga where we will validate the necessary invariants.

If `wgpu` users want more type expressiveness for the `constants` we can update its API later to accommodate that. It could look like:

```rust
enum Key {
  Id(u16),
  Name(&str)
}
enum Value {
  F32(f32),
  U32(u32),
  I32(i32),
  Bool(bool),
}
constants: &HashMap<Key, Value> or &[(Key, Value)]
```

cc @cwfitzgerald

wgpu/src/lib.rs
teoxoy: idk really, maybe we can clarify it once someone runs into this?

teoxoy:
> This definitely needs a big honkin' `CHANGELOG.md` entry, since it breaks pretty much everybody.

I was thinking it would be best to add a changelog entry once we are done with all the changes (this PR is targeting a feature branch not trunk).

teoxoy:
Pushed a new commit that addresses 2 comments.
Diff: https://github.com/gfx-rs/wgpu/compare/c3f93e63ab50b298040654661fef7fae47af7729..e3d4779ec08e3c5dc5f37553f6db9f6753f9f350

naga/src/back/msl/mod.rs
ErichDonGubler: D'oh, right, `f64` doesn't implement `Eq`. 👍🏻

wgpu/src/lib.rs
ErichDonGubler: I consider this conversation resolved, but I'll leave it up in case others might be interested in this question. Feel free to `Resolve`, if you want.

wgpu/src/lib.rs
ErichDonGubler: Ah, okay. I thought, for some reason, we were re-exporting `naga` API, but if that's not the case, then I don't see a strong reason to do as I suggest.

tests/tests/scissor_tests/mod.rs
ErichDonGubler: If this change is to be consistent with how things normally get formatted, then 👍🏻👍🏻. Minor complaint leading to new knowledge, yay!

ErichDonGubler (approved):
All concerns I had during my PR LGTM. I can't speak for @jimblandy's concerns, tho.

wgpu/src/lib.rs
cwfitzgerald: I'm not particularly happy with the untyped api. At the very least, I think we should add the Key type. I'm not sure there's actually any benefit from strongly typing the Value here.

teoxoy:
@crowlKats do the deno changes look good?

wgpu/src/lib.rs
teoxoy: `Value` could implement `From` for all the types which would make it easier to use. But yeah, `f64` already has `From` impls for all of those.

teoxoy: What is your take on?

```rust
constants: &'a HashMap<Key, Value> or &'a [(Key<'a>, Value)]
```

Note that `Key::Name` would most likely have to be a `String` for the `HashMap` but could be a `&str` for the slice of tuples.

cwfitzgerald: I'm definitely leaning array - hash maps are hard to construct inline and we don't really need deduplication behavior at the edge.

teoxoy: Thinking a bit more about this, I think we should come up with a solution that works for all users of this API (Firefox, deno and wgpu) since all have to pass this structure through `wgpu-core` -> `wgpu-hal` -> `naga`.

teoxoy: I'll try to update it to the shape below and report back.

```rust
enum Key<'a> {
  Id(u16),
  Name(&'a str)
}
constants: &'a [(Key<'a>, f64)]
```

crowlKats (commented):
Deno part LGTM

teoxoy:
I managed to do it again... 🤦‍♂️😅 (https://github.com/gfx-rs/wgpu/pull/4677#issuecomment-1810285493)

I need to get used to this workflow.

wgpu/src/lib.rs
teoxoy: @cwfitzgerald I tried to update the API to use ^ everywhere but I'm running into lifetime issues and would like to get to the implementation. Would you be ok with the current API for now?

cwfitzgerald: Yeah, we can leave it for now and iterate on it once it's in.

PR #4684, opened 2023-11-14
[naga] Exercise GLSL double-precision builtin functions, and fix the fallout.

### LINKED ISSUE #4680: [naga glsl-in] double-precision ldexp overload expects a 64-bit integer argument
In Naga's GLSL front end, the `ldexp` overload for `double` arguments expects a 64-bit integer as its second argument, but 64-bit integers are not allowed in Naga.

To reproduce:

```
$ cat ldexp-double.frag
#version 450

void main() {
    double f = 42.0;
    int i = 5;
    double ldexpOut = ldexp(f, i);
}
$ cargo run -p naga-cli -- ldexp-double.frag
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
     Running `/home/jimb/rust/wgpu/target/debug/naga ldexp-double.frag`
error: 

Type [10] '' is invalid: 
	The Sint scalar width 8 is not supported
$ 
```

### LINKED ISSUE #4681: WGSL backend writes `matNxM<f64>` construction expressions incorrectly.
The Naga WGSL backend writes incorrect WGSL for `Compose` expressions building matrices whose elements are `f64` values.

To reproduce:

```
$ cat mat4.frag
#version 450

void main() {
    dvec4 a = dvec4(1.0);
    dvec4 b = dvec4(2.0);
    dmat4 m = dmat4(a, b, a, b);
}
$ cargo run -p naga-cli -- mat4.frag mat4.wgsl
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
     Running `/home/jimb/rust/wgpu/target/debug/naga mat4.frag mat4.wgsl`
$ cat mat4.wgsl 
fn main_1() {
    var a: vec4<f64>;
    var b: vec4<f64>;
    var m: mat4x4<f32>;

    a = vec4(f64(1.0));
    b = vec4(f64(2.0));
    let _e8 = a;
    let _e9 = b;
    let _e10 = a;
    let _e11 = b;
    m = mat4x4<f32>(vec4<f64>(_e8.x, _e8.y, _e8.z, _e8.w), vec4<f64>(_e9.x, _e9.y, _e9.z, _e9.w), vec4<f64>(_e10.x, _e10.y, _e10.z, _e10.w), vec4<f64>(_e11.x, _e11.y, _e11.z, _e11.w));
    return;
}

@fragment 
fn main() {
    main_1();
    return;
}
$ 
```

Note that the construction expression tries to build a `mat4x4<f32>`. The IR's types arena is:

```
    types: {
        [1]: Type {
            name: None,
            inner: Vector {
                size: Quad,
                kind: Float,
                width: 8,
            },
        },
        [2]: Type {
            name: None,
            inner: Matrix {
                columns: Quad,
                rows: Quad,
                width: 8,
            },
        },
    },
```

There is no `mat4x4<f32>` type in that arena for the Naga `Compose` expression to refer to, so the WGSL backend must be writing incorrect code for that expression.

jimblandy:
Exercise GLSL double-precision builtin functions, and fix the fallout.

-   [naga] New tests for GLSL double-precision builtin functions.

-   [naga wgsl-out] Correctly include width in matrix constructions.

    When generating WGSL for an `Expression::Compose` constructing a
    matrix, consult `TypeInner::Matrix::width` when writing the
    type name in the construction expression, rather than just always
    writing `matNxM<f32>`.

    Fixes #4681.

-   [naga glsl-in] Fix type of double ldexp's second arg.

    The second argument of the GLSL `ldexp` builtin is always a 32-bit
    integer or a vector of such, never a 64-bit integer. In
    `inject_common_builtin`, that argument's type should not be influenced
    by `float_width`.

    Fixes #4680.

-   [naga glsl-in] Fix double overload for dot/reflect/distance/ldexp.

    The first argument of the `dot`, `reflect`, `distance`, and `ldexp`
    GLSL builtin functions may be either a float or a double, and thus the
    argument type registered by `inject_common_builtin` must depend on the
    `float_width` argument; it cannot simply be `Scalar::F32`.

    Introduced by #4673.

**Checklist**

- [X] Run `cargo clippy`.


naga/src/back/wgsl/writer.rs
```diff
@@ -524,14 +524,14 @@ impl<W: Write> Writer<W> {
...
+                width,
             } => {
                 write!(
                     self.out,
-                    //TODO: Can matrix be other than f32?
-                    "mat{}x{}<f32>",
+                    "mat{}x{}<{}>",
                     back::vector_size_str(columns),
                     back::vector_size_str(rows),
+                    scalar_kind_str(crate::Scalar::float(width))
```
ErichDonGubler, line 534: thought(non-blocking): RE: #4687: This looks like a good candidate for moving to `to_wgsl`, eh? 😀

naga/tests/in/glsl/double-math-functions.frag
```diff
@@ -0,0 +1,41 @@
...
+    double dotOut = dot(a, b);
+    dvec4 maxOut = max(a, b);
+    dvec4 minOut = min(a, b);
+    dvec4 reflectOut = reflect(a, b);
+    dvec3 crossOut = cross(a.xyz, b.xyz);
+    // TODO: support outerProduct function in wgsl output
+    // dmat4 outerProductOut = outerProduct(a, b);
+    double distanceOut = distance(a, b);
+    dvec4 stepOut = step(a, b);
+    // TODO: support out params in wgsl output
```
ErichDonGubler: issue: We should prefer issues to `TODO`s, right?

naga/tests/in/glsl/double-math-functions.frag
ErichDonGubler, line 1: question: Shouldn't there be more `out` snapshots with other backends for this `in` test? 🤔

ErichDonGubler (approved):
LGTM, minus nits/questions.

naga/src/back/wgsl/writer.rs
jimblandy: We don't really have a good place at present for code that is language-specific, but shared between the front and back end. The `proc` module is all language-generic IR utilities. So the WGSL front and back ends don't share much code. That might not be bad to fix (and this is a good candidate), but I don't think this PR is the right place for it.

naga/tests/in/glsl/double-math-functions.frag
jimblandy: We definitely should! I was just copy-pasting this from `math-functions.frag`, and forgot about those.

naga/tests/in/glsl/double-math-functions.frag
jimblandy: The general principle of Naga snapshot tests is that WGSL inputs go to all outputs, and non-WGSL inputs go only to WGSL outputs. So it's like, fan-in to WGSL, and fan-out from WGSL. We don't actually want the full cartesian product, because it's a lot of churn for not much additional coverage.

PR #4687, opened 2023-11-14
[naga wgsl-in] Clean up type formatting.

jimblandy:
-   Use `to_wgsl` functions everywhere appropriate.

    Replace `ExpressionContext`'s methods `format_typeinner`,
    `format_type`, and `format_type_resolution` with more `to_wgsl`
    methods in the `naga::front::wgsl::to_wgsl` module.

-   Let TypeInner::to_wgsl take the context by reference.

-   Move all `to_wgsl` functions into their own module.

    This is code motion only, with minor doc fixes. There should be no
    changes to the code.

**Checklist**

- [X] Run `cargo clippy`.


naga/src/front/wgsl/to_wgsl.rs
```diff
@@ -0,0 +1,256 @@
...
+            crate::ScalarKind::Uint => "u",
+            crate::ScalarKind::Float => "f",
+            crate::ScalarKind::Bool => return "bool".to_string(),
+        };
+        format!("{}{}", prefix, self.width * 8)
+    }
+}
+
+impl crate::StorageFormat {
+    pub const fn to_wgsl(self) -> &'static str {
```
ErichDonGubler, line 151: nitpick: Should this not be `pub(crate)`? This seems intended as purely internal.

This likely applies to other items that have been similarly moved.

naga/src/front/wgsl/to_wgsl.rs
```diff
@@ -0,0 +1,256 @@
+impl crate::TypeInner {
+    /// Formats the type as it is written in wgsl.
+    ///
+    /// For example `vec3<f32>`.
+    ///
+    /// Note: `TypeInner::Struct` doesn't include the name of the
+    /// struct type. Therefore this method will simply return "struct"
+    /// for them.
```
ErichDonGubler, line 35: funny: Heeey, I was promised _only_ code motion here. 😱🙃

Change still LGTM; I do agree that this is more clear.

naga/src/front/wgsl/lower/mod.rs
```diff
@@ -925,17 +908,13 @@ impl<'source, 'temp> Lowerer<'source, 'temp> {
...
-                                .name
-                                .clone()
-                                .unwrap_or_else(|| ty.inner.to_wgsl(&ctx.module.to_ctx()));
+                            let gctx = ctx.module.to_ctx();
+                            let ty = &gctx.types[explicit];
+                            let expected =
+                                ty.name.clone().unwrap_or_else(|| ty.inner.to_wgsl(&gctx));
+
+                            let ty = &gctx.types[inferred_type];
+                            let got = ty.name.clone().unwrap_or_else(|| ty.inner.to_wgsl(&gctx));
```
ErichDonGubler: This looks like a missed opportunity to use `format_type`/`to_wgsl`! 😀

```diff
diff --git a/naga/src/front/wgsl/lower/mod.rs b/naga/src/front/wgsl/lower/mod.rs
index a770e2134..684ba7ab0 100644
--- a/naga/src/front/wgsl/lower/mod.rs
+++ b/naga/src/front/wgsl/lower/mod.rs
@@ -909,17 +909,10 @@ impl<'source, 'temp> Lowerer<'source, 'temp> {
                     if let Some(explicit) = explicit_ty {
                         if explicit != inferred_type {
                             let gctx = ctx.module.to_ctx();
-                            let ty = &gctx.types[explicit];
-                            let expected =
-                                ty.name.clone().unwrap_or_else(|| ty.inner.to_wgsl(&gctx));
-
-                            let ty = &gctx.types[inferred_type];
-                            let got = ty.name.clone().unwrap_or_else(|| ty.inner.to_wgsl(&gctx));
-
                             return Err(Error::InitializationTypeMismatch {
                                 name: c.name.span,
-                                expected,
-                                got,
+                                expected: explicit.to_wgsl(&gctx),
+                                got: inferred_type.to_wgsl(&gctx),
                             });
                         }
                     }

```

Pushed as cbd4c6fa8 as a `fixup!`, in case it's helpful. Feel free to make it its own standalone commit, if that makes more sense.

ErichDonGubler (approved):
LGTM overall! This feels like a move to something more cohesive, so 👍🏻. I have some nits that should be trivial to address.

nitpick: I think I liked the `Option` combinators from the code you're moving/factoring out better than the explicit match. I pushed 84ac63216 to express this preference, but absolutely feel free to `drop` that commit in a rebase if you want. 🙂 No hurt feelings here.

naga/src/front/wgsl/to_wgsl.rs
jimblandy: I did say, "with minor doc fixes"! :dango:

naga/src/front/wgsl/to_wgsl.rs
jimblandy: I don't think Naga is consistently using `pub(crate)` for items in private modules.

naga/src/front/wgsl/lower/mod.rs
jimblandy: Oh, yes - thanks!

jimblandy:
> nitpick: I think I liked the `Option` combinators from the code you're moving/factoring out better than the explicit match. I pushed [84ac632](https://github.com/gfx-rs/wgpu/commit/84ac63216710b008984565ac592f64b0fb8e9602) to express this preference, but absolutely feel free to `drop` that commit in a rebase if you want. 🙂 No hurt feelings here.

I felt like the `match` was clearer, so I dropped this commit.

PR #4698, opened 2023-11-16
Prevent panic when submitting a destroyed buffer

nical:
**Checklist**

- [x] Run `cargo clippy`.

**Connections**

Another adjustment after #4657.

**Description**

The code in command buffer submission uses the "infallible" storage getter that panics if a texture or buffer has been destroyed, this PR adds the explicit check.  Before #4657, the case of destroyed buffer was somewhat handled by checking whether the raw buffer was `None`. it would not catch that a buffer had been scheduled for destruction but not actually destroyed yet. With this PR we can safely unwrap the option because get only returns the buffer if it is in a valid state.

**Testing**

Covered by the CTS

### DIFF
--- a/wgpu-core/src/device/queue.rs
+++ b/wgpu-core/src/device/queue.rs
@@ -1139,13 +1139,16 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
 
                         // update submission IDs
                         for id in cmdbuf.trackers.buffers.used() {
-                            let buffer = &mut buffer_guard[id];
-                            let raw_buf = match buffer.raw {
-                                Some(ref raw) => raw,
-                                None => {
+                            let buffer = match buffer_guard.get(id.0) {
+                                Ok(buf) => buf,
+                                Err(..) => {
                                     return Err(QueueSubmitError::DestroyedBuffer(id.0));
                                 }
                             };
+                            // get fails if the buffer is invalid or destroyed so we can assume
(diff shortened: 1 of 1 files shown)


wgpu-core/src/device/queue.rs
```diff
@@ -1139,13 +1139,16 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
...
-                                None => {
+                            let buffer = match buffer_guard.get(id.0) {
+                                Ok(buf) => buf,
+                                Err(..) => {
                                     return Err(QueueSubmitError::DestroyedBuffer(id.0));
                                 }
                             };
+                            // get fails if the buffer is invalid or destroyed so we can assume
+                            // the raw buffer is not None.
+                            let raw_buf = buffer.raw.as_ref().unwrap();
```
teoxoy, line 1150: The `Option` around raw hal buffers and textures seems to have the same semantics as `.get()` and `.get_mut()` returning an `Error`.

This is probably the same thing you were saying on the call yesterday that we now have multiple ways to do the same thing and that we should wait for arcanization to land and settle on one way after.

To keep things consistent at least to the extent we can right now, can we either error with `QueueSubmitError::DestroyedBuffer` on a `None` here or use `.unwrap()` below for the `TextureInner::Native` variant instead of erroring there?

teoxoy (approved):
Looks fine for now, left a suggestion.

PR #4701, opened 2023-11-16
[naga msl-out] Emit and init `struct` member padding always

### LINKED ISSUE #4522: Metal shader translation does not align `vec3`s to match std140
***See later updates as this original description is an incorrect diagnosis.***

My application worked OK under the Metal backend but mis-shaded under the WebGL backend. I made the problem disappear by modifying my camera uniform struct to avoid lone `f32`s in favor of `vec4<f32>`s.

My understanding is that this is because WebGL uniforms expect [“std140” layout](https://www.khronos.org/opengl/wiki/Interface_Block_(GLSL)#Memory_layout) which pads every field up to a multiple of 16 bytes (the size of a `vec4<f32>`). (My own knowledge of GL is limited here, as I never used structs in uniforms rather than individual scalar/vector/matrix uniforms before starting to use `wgpu`, so I could be missing some nuance.)

It seems to me that, to support portability between backends, `wgpu` should either:

* do whatever code transformation is necessary to effectively _not_ have std140 padding on the WebGL backend (probably not a good idea), or
* consider it a `MissingDownlevelFlags` error when a shader has a struct used as a uniform which has padding under std140 that would not be present on other backends. (In particular, `DownlevelFlags::BUFFER_BINDINGS_NOT_16_BYTE_ALIGNED` probably applies to the same situations; maybe it should be renamed and expanded to cover this?)

ErichDonGubler:
**Checklist**

- [x] ~~Run `cargo clippy`.~~
- [x] ~~Run `cargo clippy --target wasm32-unknown-unknown` if applicable.~~

**Connections**

Resolves #4522.

**Description**

Previously, implicit padding members of `struct`s were suppressed from structure definitions in Metal output if they had a binding specified for them (i.e., `@location(0)`). This padding is, however, is necessary for correct access of member fields passed into shaders by uniform and storage buffers. Unconditionally emit padding members for `struct`s.

**Testing**

I've added a new `struct-layout` test to `naga`'s `snapshots` test suite to exercise the adjustments to vertex shader coercion of `varyings` structures (i.e., "

### DIFF
--- a/naga/tests/out/msl/quad.msl
+++ b/naga/tests/out/msl/quad.msl
@@ -6,6 +6,7 @@ using metal::uint;
 
 struct VertexOutput {
     metal::float2 uv;
+    char _pad1[8];
     metal::float4 position;
 };
 constant float c_scale = 1.2;
@@ -23,7 +24,7 @@ vertex vert_mainOutput vert_main(
 ) {
     const auto pos = varyings.pos;
     const auto uv = varyings.uv;
-    const auto _tmp = VertexOutput {uv, metal::float4(c_scale * pos, 0.0, 1.0)};
+    const auto _tmp = VertexOutput {uv, {}, metal::float4(c_scale * pos, 0.0, 1.0)};
     return vert_mainOutput { _tmp.uv, _tmp.position };
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,8 @@ Passing an owned value `window` to `Surface` will return a `Surface<'static>`. S
 - Implement WGSL abstract types (by @jimblandy):
   - Add a new `naga::Literal` variant, `I64`, for signed 64-bit literals. [#4711](https://github.com/gfx-rs/wgpu/pull/4711)
 
+- Emit and init `struct` member padding always. By @ErichDonGubler in [#4701](https://github.com/gfx-rs/wgpu/pull/4701).
+
 ### Bug Fixes
 
 #### WGL
--- a/naga/src/back/msl/writer.rs
+++ b/naga/src/back/msl/writer.rs
@@ -3302,7 +3302,6 @@ impl<W: Write> Writer<W> {
                     writeln!(self.out, "struct {name} {{")?;
                     let mut last_offset = 0;
                     for (index, member) in members.iter().enumerate() {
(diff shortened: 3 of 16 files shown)


teoxoy:
@ErichDonGubler and I went over this but ultimately found out that this solution might cause more issues than it solves.

I filed https://github.com/gfx-rs/wgpu/issues/4703 to track the full solution.

ErichDonGubler:
Reopening this, as this fix is _actually_ the right direction for #4522 (which has been reopened, also).

ErichDonGubler:
@teoxoy (who I presume will review this): I've added e70e4ba28, 5a8e4ae13, and a199623c7 to address what I believe is the remaining scope of work here. I'm uncertain if the test I've added needs more cases, so I've marked it as WIP for now. I can reword that commit if you think it's good enough, tho.

naga/CHANGELOG.md
```diff
@@ -79,6 +79,7 @@ For changelogs after v0.14, see [the wgpu changelog](../CHANGELOG.md).
 
 - Add and fix minimum Metal version checks for optional functionality. ([#2486](https://github.com/gfx-rs/naga/pull/2486)) **@teoxoy**
 - Make varyings' struct members unique. ([#2521](https://github.com/gfx-rs/naga/pull/2521)) **@evahop**
+- Emit and init `struct` member padding always ([#4701](https://github.com/gfx-rs/wgpu/pull/4701)) **@ErichDonGubler**
```
teoxoy: This should go in the changelog at the root of the project.

naga/tests/out/msl/quad.msl
```diff
@@ -6,6 +6,7 @@ using metal::uint;
 
 struct VertexOutput {
     metal::float2 uv;
+    char _pad1[8];
```
teoxoy, line 9: Hmm, we don't need to explicitly pad here but I guess this doesn't hurt anything and is actually a preexisting issue.

The padding calculation needs to align the offset before it checks the member's offset.
We can fix it later.

teoxoy (approved):
Great stuff!

naga/CHANGELOG.md
ErichDonGubler: oderp, fixed with a6c176f4f. 😅

PR #4706, opened 2023-11-17
[naga] remove `span` and `validate` features

### LINKED ISSUE #4433: Evaluate cost / benefit of `"span"` feature
The `"span"` feature is essential for producing usable diagnostics. As we improve Naga's diagnostics I suspect it's going to become harder to maintain `"span"` as an optional feature, as opposed to simply providing it by default.

We should measure the impact of the `"span"` feature on speed, memory consumption, and code size, and see what it would cost to simply enable it unconditionally.

### LINKED ISSUE #4490: Do we need to call `VaryingContext::validate` at all when `#[cfg(not(feature = "validate"))]`?
If I understand correctly, when the `"validate"` feature is not enabled, we should only do enough work to produce a `ModuleInfo` the backends can use, and no more. But `naga::valid::Validator::validate_entry_point` calls `VaryingContext::validate` on the entry point's arguments and return value, and I don't see that that contributes anything to the final `ModuleInfo`.

Should those calls be protected by a `#[cfg(not(feature = "validate"))]` condition?

### LINKED ISSUE #4705: Remove naga's `span` and `validate` features
It's unclear why the spans are gated behind the `span` feature. They are essential for diagnostics, take almost no space, and all users of naga that I could find enable it or have it enabled because they depend on `wgpu-core` transitively.

The `validate` feature might have more value than `span` as it gates quite a bit more code than `span` but still gets enabled by `wgpu-core` unconditionally.

The main driving factor behind removing them is maintenance burden (having to always consider the case where those are not enabled).

Related (original PRs):

- https://github.com/gfx-rs/naga/pull/1092
- https://github.com/gfx-rs/naga/pull/1437

teoxoy:
- closes https://github.com/gfx-rs/wgpu/issues/4705
- closes https://github.com/gfx-rs/wgpu/issues/4433
- closes https://github.com/gfx-rs/wgpu/issues/4490

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ Passing an owned value `window` to `Surface` will return a `Surface<'static>`. S
 
 #### Naga
 
+- Remove `span` and `validate` features. Always fully validate shader modules, and always track source positions for use in error messages. By @teoxoy in [#4706](https://github.com/gfx-rs/wgpu/pull/4706)
 - Introduce a new `Scalar` struct type for use in Naga's IR, and update all frontend, middle, and backend code appropriately. By @jimblandy in [#4673](https://github.com/gfx-rs/wgpu/pull/4673).
 
 ### Bug Fixes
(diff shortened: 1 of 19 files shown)


cwfitzgerald (approved):
Consider the one line toml change in wgpu-core approved lol

Also yay!

jimblandy:
Pushed a doc fix (stray reference to `span` feature).

jimblandy (approved):
Looks good to me!

jimblandy:
In CHANGELOG.md, I explained the consequences of this change a bit.

Issue #4555, opened 2023-06-02
Add backend intrinsics/built-in functions to their respective reserved keywords list


teoxoy:
User defined symbols might collide with backend intrinsics, we need to rename them.

- [x] HLSL https://github.com/gfx-rs/naga/pull/2367
- [ ] MSL: https://github.com/gfx-rs/wgpu/pull/4707, might need more
- [ ] WGSL
- [x] GLSL https://github.com/gfx-rs/naga/pull/2410

PR #4707, opened 2023-11-17
Add more metal keywords

### LINKED ISSUE #4545: Declaring something with name 'half' causes broken MSL generated.
It seems that `half` is used for a built-in type in MSL, so when a field with name `half` is declared in wgsl the resulting MSL cannot be compiled.

fornwall:
Fixes #4545.

I'm not totally sure if some name shadowing could be allowed?

Also moved `clamped_lod_e` from keywords to reserved prefixes.


teoxoy:
For the ones that have been removed, I compiled a list that mentions why they were unnecessary:

- `char8_t` - C++20
- `co_await` - C++20
- `co_return` - C++20
- `co_yield` - C++20
- `module` - C++20
- `import` - C++20
- `visible` - metal function attribute
- `vec_step` - clang specific for OpenCL
- `typeof` - specific to gcc
- `non-temporal` - can't find these in C++ nor Metal
- `dereferenceable` - can't find these in C++ nor Metal
- `invariant` - metal attribute
- `compute` - this is `kernel` for metal
- `read_only` - we use `metal::access::read`
- `write_only` - we use `metal::access::write`
- `read_write` - we use `metal::access::read_write`

teoxoy:
> I'm not totally sure if some name shadowing could be allowed?

We can't allow shadowing or else we might run into something like:
https://shader-playground.timjones.io/11df1fe53f589f9a52906a21c8454170

https://github.com/gfx-rs/wgpu/issues/4555 is tracking this issue.

teoxoy (changes requested):
This looks great, but it seems `packed_booln`, `halfnxm` and `floatnxm` are missing.

fornwall:
> This looks great, but it seems packed_booln, halfnxm and floatnxm are missing.

@teoxoy Thanks - added them in https://github.com/gfx-rs/wgpu/pull/4707/commits/52ee9780928cee38c90b45c6806ab6ff85439869

teoxoy (approved):
Looks great, thanks!

PR #4175, opened 2023-09-26
Improve safety of COM API usage in `d3d12` by introducing non-nullable type state

ErichDonGubler:
Hard dependency on #3987.

PR #4716, opened 2023-11-17
Add non-nullable type state `d3d12::ComPtr`

ErichDonGubler:
**Connections**

**Description**

**Testing**

**Checklist**

- [ ] Run `cargo fmt`.

### DIFF
--- a/d3d12/src/com.rs
+++ b/d3d12/src/com.rs
@@ -23,7 +23,7 @@ impl<T: Interface> ComPtr<T> {
         ComPtr(ptr::null_mut())
     }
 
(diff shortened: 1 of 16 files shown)


ErichDonGubler:
@cwfitzgerald: This is _not_ ready to go yet, but I've refreshed commits from https://github.com/gfx-rs/wgpu/pull/4175. I do intend to work on this soon, since @nical and I are discovering some crashes in Firefox's WebGPU CTS runs that indicate we're getting null `TextureView`s. This should make debugging for that and future issues easier to diagnose, since this will force `wgpu-hal::dx12` to check for null much earlier than it currently does.

PR #4717, opened 2023-11-17
[naga]: Removed unused arg to `TypeInner::size`.

jimblandy:
- [X] Run `cargo fmt`.


jimblandy:
... Didn't realize this was a public function. :sweat_smile:

cwfitzgerald:
(it still would be nice to have, there doesn't seem to be a reason for the argument)

PR #4719, opened 2023-11-17
[naga] Let constant evaluation of `As` preserve `Splat` expressions.

jimblandy:
When asked to evaluate an `Expression::As` cast applied to a `Splat` expression, change `ConstantEvaluator::cast` to preserve the `Splat`, rather than expanding it out to a `Compose` expression.

- [X] Run `cargo fmt`.

### DIFF
--- a/naga/src/proc/constant_evaluator.rs
+++ b/naga/src/proc/constant_evaluator.rs
@@ -868,6 +868,22 @@ impl<'a> ConstantEvaluator<'a> {
     /// [`ZeroValue`]: Expression::ZeroValue
     /// [`Literal`]: Expression::Literal
     /// [`Compose`]: Expression::Compose
+    fn eval_zero_value(
+        &mut self,
+        expr: Handle<Expression>,
+        span: Span,
+    ) -> Result<Handle<Expression>, ConstantEvaluatorError> {
+        match self.expressions[expr] {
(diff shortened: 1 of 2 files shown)


naga/src/proc/constant_evaluator.rs
```diff
@@ -1022,6 +1038,14 @@ impl<'a> ConstantEvaluator<'a> {
...
                 Expression::Compose { ty, components }
             }
+            Expression::Splat { size, value } => {
+                let value_span = self.expressions.get_span(value);
+                let cast_value = self.cast(value, target, value_span)?;
+                Expression::Splat {
+                    size,
+                    value: cast_value,
+                }
+            }
```
teoxoy, line 1048: It seems simple enough to also handle `ZeroValue` removing the need for the new `eval_zero_value` fn.

teoxoy (approved):
Looks good, left a suggestion. Feel free to push it to this PR/start another or ignore it :)

naga/src/proc/constant_evaluator.rs
jimblandy: At the moment, `eval_zero_value_impl` handles arrays, including nested arrays. Supporting those cases in `cast` would entail walking all the way into the type, adjusting the leaf scalar, and then rebuilding new types on the way back out. I think we will have to do that eventually to support concretizing arrays, but I would rather not take it on as part of this PR.

PR #4723, opened 2023-11-19
Test And Normalize Vertex Behavior on All Backends

### LINKED ISSUE #4256: Missing test coverage for draw_indirect
This was something noticed by @cwfitzgerald on matrix when I was asking about it

### LINKED ISSUE #4276: `draw_instanced_offset` fails on GL
The `draw_instanced_offset` test is failing in GL, suggesting the workaround translating gl_InstanceIndex into gl_InstanceID + offset isn't functioning properly.

cwfitzgerald:
**Connections**

Closes #4256 
Closes #4276

**Description**

This adds the necessary logic so that `@builtin(instance_index)` works correctly against all supported calls, both direct and indirect.

**Testing**

Completely re-implemented the tests to cover many permutations.

**Checklist**

- [x] Run `cargo fmt`.


wgpu/src/util/mod.rs
```diff
@@ -19,9 +18,8 @@ use std::{
 pub use belt::StagingBelt;
 pub use device::{BufferInitDescriptor, DeviceExt};
 pub use encoder::RenderEncoder;
-pub use indirect::*;
 pub use init::*;
-pub use wgt::math::*;
+pub use wgt::{math::*, DispatchIndirectArgs, DrawIndexedIndirectArgs, DrawIndirectArgs};
```
i509VCB, line 22: Technically a breaking change due to type name changes (`Arg` was added to each type name)

cwfitzgerald: Yeah, I should probably switch it back, was trying not to touch `core` though, but this obviously isn't going to land before arcinization :)

cwfitzgerald: Decided to keep the change, but mention it

tests/tests/vertex_indices/mod.rs
```diff
@@ -1,14 +1,237 @@
...
+            self.instance.start
+        } else {
+            0
+        };
+
+        if let Some(vertex_offset) = self.vertex_offset {
+            bytes.extend_from_slice(
+                wgpu::util::DrawIndexedIndirectArgs {
+                    vertex_count: self.vertex.end - self.vertex.start,
+                    instance_count: self.instance.end - self.instance.start,
```
teoxoy, line 44:
```suggestion
                    instance_count: self.instance.end - first_instance,
```

tests/tests/vertex_indices/mod.rs
```diff
@@ -1,14 +1,237 @@
...
+                    first_index: self.vertex.start,
+                    first_instance,
+                }
+                .as_bytes(),
+            )
+        } else {
+            bytes.extend_from_slice(
+                wgpu::util::DrawIndirectArgs {
+                    vertex_count: self.vertex.end - self.vertex.start,
+                    instance_count: self.instance.end - self.instance.start,
```
teoxoy, line 55:
```suggestion
                    instance_count: self.instance.end - first_instance,
```

tests/tests/vertex_indices/mod.rs
```diff
@@ -1,14 +1,237 @@
...
+
+    /// Add the draw call to the given indirect buffer
+    fn add_to_buffer(&self, bytes: &mut Vec<u8>, features: wgpu::Features) {
+        // The behavior of non-zero first_instance in indirect draw calls in currently undefined if INDIRECT_FIRST_INSTANCE is not supported.
+        let supports_first_instance = features.contains(wgpu::Features::INDIRECT_FIRST_INSTANCE);
+        let first_instance = if supports_first_instance {
+            self.instance.start
+        } else {
+            0
+        };
```
teoxoy, line 38: Or would it make sense to turn this into an assert instead and make sure we don't call this fn with params that are not supported?

tests/tests/vertex_indices/mod.rs
```diff
@@ -1,14 +1,237 @@
...
+                    instance: 0..1,
+                    vertex_offset: None,
+                },
+                Draw {
+                    vertex: 3..6,
+                    instance: 0..1,
+                    vertex_offset: None,
+                },
+            ],
+            TestCase::DrawBaseVertex => &[Draw {
```
teoxoy, line 124: I find the names for `DrawVertexOffset`, `DrawBaseVertex` and `DrawInstancedOffset` to be a bit confusing.

`DrawVertexOffset` doesn't specify a `vertex_offset` but `DrawBaseVertex` does.

It might be better to rename `vertex_offset` to `base_vertex` (that's also what the spec calls it).

wgpu-types/src/lib.rs
```diff
@@ -6507,44 +6553,83 @@ impl_bitflags!(PipelineStatisticsTypes);
...
 pub struct DrawIndexedIndirectArgs {
-    /// The number of indices to draw.
-    pub index_count: u32,
+    /// The number of vertices to draw.
+    pub vertex_count: u32,
     /// The number of instances to draw.
     pub instance_count: u32,
-    /// Offset into the index buffer, in indices, begin drawing from.
+    /// The first index within the index buffer.
     pub first_index: u32,
```
teoxoy, line 6591:
```suggestion
    pub first_vertex: u32,
```

wgpu-types/src/lib.rs
```diff
@@ -6507,44 +6553,83 @@ impl_bitflags!(PipelineStatisticsTypes);
...
-    /// Added to each index value before indexing into the vertex buffers.
-    pub base_vertex: i32,
-    /// First instance to draw.
+    /// The value added to the vertex index before indexing into the vertex buffer.
+    pub vertex_offset: i32,
```
```suggestion
    pub base_vertex: i32,
```

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -991,12 +992,19 @@ impl super::Instance {
         };
 
         if let Some(driver) = phd_capabilities.driver {
+            #[allow(clippy::unnecessary_cast, trivial_numeric_casts)]
+            // False positive on the `as u8` cast on Linux/Android aarch64
```
```suggestion
            #[cfg(not(any(target_os = "macos", target_os = "ios")))]
```

I was thinking of silently ignoring the check but if the message has value, we can go with that.

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -991,12 +992,19 @@ impl super::Instance {
...
             if driver.conformance_version.major == 0 {
-                log::warn!(
-                    "Adapter is not Vulkan compliant, hiding adapter: {}",
-                    info.name
-                );
-                return None;
+                let driver_name_unsigned = driver.driver_name.map(|i| i as u8);
+                if !driver_name_unsigned.starts_with(b"MoltenVK") {
```
```suggestion
                if driver.driver_id != ash::vk::DriverId::MOLTENVK {
```

naga/src/back/glsl/mod.rs
```diff
@@ -4308,29 +4356,32 @@ const fn glsl_scalar(scalar: crate::Scalar) -> Result<ScalarString<'static>, Err
...
         Bi::ViewIndex => "gl_ViewIndex",
         // vertex
         Bi::BaseInstance => "uint(gl_BaseInstance)",
         Bi::BaseVertex => "uint(gl_BaseVertex)",
         Bi::ClipDistance => "gl_ClipDistance",
         Bi::CullDistance => "gl_CullDistance",
-        Bi::InstanceIndex => "uint(gl_InstanceID)",
+        Bi::InstanceIndex => {
+            if options.draw_parameters {
+                "(uint(gl_InstanceID) + uint(gl_BaseInstanceARB))"
```
teoxoy, line 4369: `gl_BaseInstance` seems to also be available in GLSL 4.6

naga/src/back/glsl/mod.rs
```diff
@@ -76,6 +76,9 @@ const CLAMPED_LOD_SUFFIX: &str = "_clamped_lod";
 pub(crate) const MODF_FUNCTION: &str = "naga_modf";
 pub(crate) const FREXP_FUNCTION: &str = "naga_frexp";
 
+// Must match code in glsl_built_in
+pub const BASE_INSTANCE_BINDING: &str = "naga_vs_base_instance";
```
teoxoy:
```suggestion
pub const FIRST_INSTANCE_BINDING: &str = "naga_vs_first_instance";
```

naga/src/back/glsl/mod.rs
```diff
@@ -652,6 +686,16 @@ impl<'a, W: Write> Writer<'a, W> {
             writeln!(self.out)?;
         }
 
+        if self.entry_point.stage == ShaderStage::Vertex
+            && !self
+                .options
+                .writer_flags
+                .contains(WriterFlags::DRAW_PARAMETERS)
+        {
+            writeln!(self.out, "uniform uint {BASE_INSTANCE_BINDING};")?;
```
teoxoy: Can we make this conditional on the presence of the builtin? I'd recommend looking at `Features::MULTI_VIEW`.

wgpu-hal/src/gles/command.rs
```diff
@@ -1004,38 +1019,44 @@ impl crate::CommandEncoder<super::Api> for super::CommandEncoder {
         &mut self,
         start_vertex: u32,
         vertex_count: u32,
-        start_instance: u32,
+        base_instance: u32,
```
teoxoy: Can we keep the name as it was before? Or rename both to `first_vertex`/`first_instance` everywhere?

naga/src/back/glsl/mod.rs
teoxoy: But it seems all 4.6 implementations also support the extension, I guess this is fine then.

https://opengl.gpuinfo.org/listreports.php?extension=GL_ARB_shader_draw_parameters&option=not

teoxoy (changes requested):

teoxoy:
These are all great changes btw!

wgpu-hal/src/gles/command.rs
teoxoy: The naming across APIs looks to be a mess as well but we should at least try to keep it consistent with the WebGPU and Vulkan specs (both using `first_vertex`/`first_instance`).

tests/tests/vertex_indices/mod.rs
cwfitzgerald: This is what's actually doing the "sanitizing" - as some tests are speced to use a non-zero first_instance, and we need them to behave deterministically.

tests/tests/vertex_indices/mod.rs
cwfitzgerald: What it was is actually the indended behavior, we want it to "malfunction"

wgpu-types/src/lib.rs
cwfitzgerald: Spec has this as first_index

wgpu-hal/src/vulkan/adapter.rs
cwfitzgerald: Just so that our adapter choosing behavior is clear from the logs.

tests/tests/vertex_indices/mod.rs
teoxoy: I see, I guess it's fine then.

wgpu-types/src/lib.rs
teoxoy: Ah, right.

wgpu-types/src/lib.rs
```diff
@@ -6507,44 +6553,83 @@ impl_bitflags!(PipelineStatisticsTypes);
...
+
 /// Argument buffer layout for draw_indexed_indirect commands.
 #[repr(C)]
-#[derive(Clone, Copy, Debug)]
+#[derive(Copy, Clone, Debug, Default)]
 pub struct DrawIndexedIndirectArgs {
-    /// The number of indices to draw.
-    pub index_count: u32,
+    /// The number of vertices to draw.
+    pub vertex_count: u32,
```
teoxoy: This was correct previously.

tests/tests/vertex_indices/mod.rs
cwfitzgerald: Alright, tried some new names

teoxoy:
@cwfitzgerald I got a few more renames / small fixes, can I just push a commit? It's easier than pointing out all the places through github's UI.

cwfitzgerald:
@teoxoy Yeah, always just go ahead :)

teoxoy (approved):

teoxoy:
Pushed, if the changes look good, feel free to land it.

On a related note, I noticed we don't seem to validate `DownlevelFlags::BASE_VERTEX`.

cwfitzgerald (commented):
Nice stuff

> On a related note, I noticed we don't seem to validate DownlevelFlags::BASE_VERTEX.

Biiiiig spooky, willfix.

PR #4725, opened 2023-11-19
Corrects typo in examples FrameCounter

### LINKED ISSUE #4724: Typo in FrameCounter from examples
**Description**
There is a typo in the `FrameCounter` `update` function implementation from the examples ([here](https://github.com/gfx-rs/wgpu/blob/fd53ea90e675d94f1d79a1c3c44b2f356cecd9c5/examples/common/src/framework.rs#L345)), there is a variable called `elasped_secs`, which naturally should be `elapsed_secs`. 

**Repro steps**
The code I am referring to is [here](https://github.com/gfx-rs/wgpu/blob/fd53ea90e675d94f1d79a1c3c44b2f356cecd9c5/examples/common/src/framework.rs#L345).

I would expect the variable to be called `elapsed_secs` instead of  `elasped_secs`.

**Extra materials**
N/A

**Platform**
N/A

cantudo:
**Connections**
The issue this PR addreses is #4724 .

**Description**
The issue is that a variable in the `update` method from `FrameCounter` in the examples has a typo ([here](https://github.com/gfx-rs/wgpu/blob/fd53ea90e675d94f1d79a1c3c44b2f356cecd9c5/examples/common/src/framework.rs#L345), `elasped_secs`). This PR just renames that variable to `elapsed_secs`.

**Testing**
Because of the nature of the change no new tests were added.

**Checklist**

- [x] Run `cargo fmt`.

**Comments**
I do not think it is necessary to add this to `CHANGELOG.md`.

### DIFF
--- a/examples/common/src/framework.rs
+++ b/examples/common/src/framework.rs
@@ -342,11 +342,11 @@ impl FrameCounter {
     fn update(&mut self) {
         self.frame_count += 1;
         let new_instant = web_time::Instant::now();
-        let elasped_secs = (new_instant - self.last_printed_instant).as_secs_f32();
-        if elasped_secs > 1.0 {
-            let elapsed_ms = elasped_secs * 1000.0;
+        let elapsed_secs = (new_instant - self.last_printed_instant).as_secs_f32();
+        if elapsed_secs > 1.0 {
+            let elapsed_ms = elapsed_secs * 1000.0;
             let frame_time = elapsed_ms / self.frame_count as f32;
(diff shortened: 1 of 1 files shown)


Wumpf:
thanks! But no need to open an issue for something like that, we don't have a strict process that requires an issue for every pr :)

Wumpf (approved):

cantudo:
thank you, yeah, I am sorry about that, I saw the description template and
thought the issue was required

PR #4726, opened 2023-11-20
Fix Javascript exception on repeated `BufferSlice::get_mapped_range` calls

DouglasDwyer:
**Connections**
N/A

**Description**
Using the BrowserWebGPU backend on latest Chrome, calling `BufferSlice::get_mapped_range` more than once causes a Javascript exception, even though the `BufferView` objects are appropriately dropped Rust-side. This is because each call creates a new JS-side `ArrayBuffer` via [`getMappedRange`](https://developer.mozilla.org/en-US/docs/Web/API/GPUBuffer/getMappedRange), but the specification for `getMappedRange` forbids creating multiple overlapping array buffers. To solve this, this PR creates a single  JS-side `ArrayBuffer` and re-uses it until `unmap` is called.

**Testing**

**Checklist**

- [x] Run `cargo fmt`.


wgpu/src/backend/web.rs
```diff
@@ -3462,6 +3467,57 @@ impl QueueWriteBuffer for WebQueueWriteBuffer {
     }
 }
 
+#[derive(Debug)]
+pub struct WebBuffer {
```
Wumpf, line 3479: This should be documented with the motivation on why this struct is needed instead of just using the web_sys::GpuBuffer (essentially what you wrote in the PR description)

Wumpf (approved):
Looking great, thank you!
Good to go after a tiny bit of additional docs

Wumpf:
perfect, thank you again!

PR #4733, opened 2023-11-21
[naga wgsl-in] Preserve type names in `alias` declarations.

jimblandy:
Given a WGSL `alias` declaration, create a Naga `Type` with the alias's name, rather than dropping the type name on the floor.

**Checklist**

- [X] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,8 @@ Passing an owned value `window` to `Surface` will return a `Surface<'static>`. S
 
 - Preserve the source spans for constants and expressions correctly across module compaction. By @jimblandy in [#4696](https://github.com/gfx-rs/wgpu/pull/4696).
 
+- Record the names of WGSL `alias` declarations in Naga IR `Type`s. By @jimblandy in [#4733](https://github.com/gfx-rs/wgpu/pull/4733).
+
 ### Examples
 
 - remove winit dependency from hello-compute example by @psvri in [#4699](https://github.com/gfx-rs/wgpu/pull/4699)
--- a/naga/src/front/wgsl/lower/mod.rs
+++ b/naga/src/front/wgsl/lower/mod.rs
@@ -98,10 +98,14 @@ impl<'source> GlobalContext<'source, '_, '_> {
         }
     }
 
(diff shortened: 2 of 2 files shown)


teoxoy (changes requested):
`alias`es are just sugar for the full type and putting them in the IR will cause issues with type equivalence (`ensure_type_exists` will now return a different type than the one the `alias` is pointing to).

jimblandy:
> `alias`es are just sugar for the full type and putting them in the IR will cause issues with type equivalence (`ensure_type_exists` will now return a different type than the one the `alias` is pointing to).

I think we should be using `TypeInner::equivalent` for comparisons, not handle comparison, for this reason, and others.

Edit: We should be using `TypeInner::equivalent`, or `<TypeInner as PartialEq>` when we know that the two types being compared cannot be pointers. See, for example, [the code that validates `Statement::Store`](https://github.com/gfx-rs/wgpu/blob/72462267e8318c3fbceaef00f5b80cb9e06c185d/naga/src/valid/function.rs#L653), which compares two `TypeInner`s using `==`.

teoxoy:
`TypeInner::equivalent` is only needed because we have `TypeInner::ValuePointer`.

The existence of `TypeInner::ValuePointer` is questionable to me; it seems to complicate the code and make things harder to reason about given we already have `TypeInner::Pointer`.

Having the invariant that types are only equal if `Type == Type` is a lot easier to reason about and I thought is the main reason our type arena is a `UniqueArena`.

As an example of why I think we should avoid complicating type equality:

```wgsl
struct A {
    a: f32
}

struct B {
    a: f32
}

fn test(a: A) {}

fn test2(b: B) {
    test(b);
}
```

seems to validate just fine 😅 

or https://github.com/gfx-rs/naga/pull/1875 (since we have 2 ways to do the same thing).

teoxoy:
@jimblandy and I talked about this and we decided to go with this PR for now and later remove `TypeInner::ValuePointer`/`TypeInner::equivalent` + achieve the effect this PR will have through a different mechanism.

teoxoy (approved):

PR #4734, opened 2023-11-21
[naga] Make compaction preserve named types, even if unused.

jimblandy:
Have `compact::compact` preserve entries in the `Module::types` arena if they have names.

Future abstract type support will require the WGSL front end to compact the module before validation. Without this change, that will drop `alias` declarations, making it harder to test type validation.

**Checklist**

- [X] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,6 +103,8 @@ Passing an owned value `window` to `Surface` will return a `Surface<'static>`. S
 
 #### Naga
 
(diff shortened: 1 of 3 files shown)


teoxoy (commented):
I suppose this is only needed if we retain `alias`es in the IR (https://github.com/gfx-rs/wgpu/pull/4733) right?

jimblandy:
> I suppose this is only needed if we retain `alias`es in the IR (#4733) right?

Yes, the two PRs are collaborative, but they'll both be required for abstract types.

teoxoy (approved):

PR #4735, opened 2023-11-21
[naga]: Let `TypeInner::Matrix` hold a `Scalar`, not just a width.

jimblandy:
Let `naga::TypeInner::Matrix` hold a full `Scalar`, with a kind and byte width, not merely a byte width, to make it possible to represent matrices of AbstractFloats for WGSL.

This is a prerequisite for implementing abstract types in the WGSL front end, because it allows us to describe matrices whose elements are abstract floats.

**Checklist**

- [X] Run `cargo fmt`.


teoxoy (approved):
Good stuff!

PR #4737, opened 2023-11-21
Make the command_encoder_clear_buffer's size an Option<BufferAddress>

nical:
**Connections**

Gecko side work: https://bugzilla.mozilla.org/show_bug.cgi?id=1865830
CTS test `cts.https.html?q=webgpu:api,operation,command_buffer,clearBuffer:clear:*`

**Description**

The size parameter in WebGPU can be zero so the types have to allow expressing it.
In general I don't think any of the `BufferSize` (which is `NonZero<u32>`) usage in wgpu_core's API is desirable for the same reason (I'd extend that to wgpu's API since it's rather tedious to deal with NonZero there as well).

**Testing**

CTS test `cts.https.html?q=webgpu:api,operation,command_buffer,clearBuffer:clear:*`

**Checklist**

- [x] Run `cargo fmt`.


Wumpf (approved):
lgtm!

PR #4743, opened 2023-11-21
Implement WGSL abstract types for global `const` declarations and constructor calls.

jimblandy:
**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


jimblandy:
This is still dependent on a bunch of other PRs, visible in the merge history. Once those are landed, and I've addressed CI issues, I'll take this out of draft.

jimblandy:
Regarding the mess described in #4747: this PR just deletes `tests/in/glsl/double-math-functions.frag` for the time being, as it can't be translated to WGSL.

jimblandy:
Since #4747 has landed, this is ready for review.

jimblandy:
By the way - I think we should sometime soon rename `Lowerer::expression` to `concretized_expression`, and then rename `expression_for_abstract` to `expression`. Since so many operations in WGSL accept abstract types, the "natural" choice should the one that doesn't concretize yet, and code that does concretize should look "suspicious", as potentially in need of an update.

(In #4755 I was confused for a bit by test failures before I realized that I was still automatically concretizing the initialization expression, so that coercion to the var's declared type didn't work.)

But that can be a separate PR, isolated from any semantic changes.

jimblandy:
Hmm. There is a slightly different way to approach this patch, which would be to make constant evaluation responsible for detecting when conversions are necessary and applying them. The WGSL front end is still going to need to do things like choose types for all those `Compose` expressions in `construction.rs`, for example, but once it's done so, it seems like the actual conversions could be done directly in `constant_evaluator.rs`, rather than having the front end create a bunch of `As` expressions which it trusts the constant evaluator to consume.

@teoxoy If you'd like me to try out that alternative, just let me know. I don't think it'd take long.

naga/src/valid/expression.rs
```diff
@@ -670,22 +670,36 @@ impl super::Validator {
                 let good = match op {
                     Bo::Add | Bo::Subtract => match *left_inner {
                         Ti::Scalar(scalar) | Ti::Vector { scalar, .. } => match scalar.kind {
-                            Sk::Uint | Sk::Sint | Sk::Float => left_inner == right_inner,
+                            Sk::Uint
+                            | Sk::Sint
+                            | Sk::Float
+                            | Sk::AbstractInt
+                            | Sk::AbstractFloat => left_inner == right_inner,
                             Sk::Bool => false,
```
teoxoy: Shouldn't we return `false` for `AbstractInt`/`AbstractFloat` (in all the other places in this file as well)? It would make things clearer when reading the code.

naga/src/front/wgsl/lower/conversion.rs
```diff
@@ -0,0 +1,378 @@
...
+            (Sk::AbstractFloat, Sk::Float) => {}
+            (Sk::AbstractInt, Sk::Sint | Sk::Uint | Sk::AbstractFloat | Sk::Float) => {}
+            _ => return None,
+        }
+
+        log::trace!("      okay: expr {expr_scalar:?}, goal {goal_scalar:?}");
+        Some((expr_scalar, goal_scalar))
+    }
+
+    fn automatically_convertable_scalar(
```
teoxoy:
```suggestion
    fn automatically_convertible_scalar(
```

naga/src/front/wgsl/lower/conversion.rs
```diff
@@ -0,0 +1,378 @@
...
+            self.as_const_evaluator()
+                .cast_array(expr, goal_scalar, span)
+                .map_err(|err| super::Error::ConstantEvaluatorError(err, span))?
+        } else {
+            let cast = crate::Expression::As {
+                expr,
+                kind: goal_scalar.kind,
+                convert: Some(goal_scalar.width),
+            };
+            log::trace!("JIMB: Emitting {cast:?}");
```
teoxoy:
```suggestion
```

naga/src/front/wgsl/lower/conversion.rs
```diff
@@ -0,0 +1,378 @@
...
+    /// return `None`.
+    fn automatically_converts_to(
+        &self,
+        goal: &Self,
+        types: &crate::UniqueArena<crate::Type>,
+    ) -> Option<(crate::Scalar, crate::Scalar)> {
+        use crate::ScalarKind as Sk;
+        use crate::TypeInner as Ti;
+
+        log::trace!("JIMB: automatically_converts_to: {self:?} -> {goal:?}");
```
teoxoy:
```suggestion
```

naga/src/front/wgsl/lower/construction.rs
```diff
@@ -546,3 +591,45 @@ impl<'source, 'temp> Lowerer<'source, 'temp> {
...
+///
+/// This function doesn't fully type-check the arguments - it only
+/// considers their leaf scalar types. This means it may return `Ok`
+/// even when the Naga validator will reject the resulting
+/// construction expression later.
+fn automatic_conversion_consensus(
+    components: &[Handle<crate::Expression>],
+    ctx: &ExpressionContext<'_, '_, '_>,
+) -> Result<Scalar, usize> {
+    log::trace!("JIMB: automatic_conversion_consensus");
```
teoxoy:
```suggestion
```

naga/src/front/wgsl/lower/conversion.rs
```diff
@@ -0,0 +1,378 @@
...
+            _ => return None,
+        }
+
+        match (expr_scalar.kind, goal_scalar.kind) {
+            (Sk::AbstractFloat, Sk::Float) => {}
+            (Sk::AbstractInt, Sk::Sint | Sk::Uint | Sk::AbstractFloat | Sk::Float) => {}
+            _ => return None,
+        }
+
+        log::trace!("      okay: expr {expr_scalar:?}, goal {goal_scalar:?}");
```
teoxoy, line 276:
```suggestion
```

naga/src/front/wgsl/lower/construction.rs
```diff
@@ -546,3 +591,45 @@ impl<'source, 'temp> Lowerer<'source, 'temp> {
...
+    let types = &ctx.module.types;
+    let mut inners = components
+        .iter()
+        .map(|&c| ctx.typifier()[c].inner_with(types));
+    let mut best = inners.next().unwrap().scalar().ok_or(0_usize)?;
+    log::trace!("      start: {best:?}");
```
teoxoy:
```suggestion
```

naga/src/front/wgsl/lower/construction.rs
```diff
@@ -546,3 +591,45 @@ impl<'source, 'temp> Lowerer<'source, 'temp> {
...
+    for (inner, i) in inners.zip(1..) {
+        let scalar = inner.scalar().ok_or(i)?;
+        match best.automatic_conversion_join(scalar) {
+            Some(new_best) => {
+                best = new_best;
+                log::trace!("      new:   {best:?}");
```
teoxoy:
```suggestion
```

naga/src/front/wgsl/lower/conversion.rs
```diff
@@ -0,0 +1,378 @@
...
+
+            // AbstractInt converts to concrete integer or float.
+            (Sk::AbstractInt, kind @ (Sk::Uint | Sk::Sint | Sk::Float)) => Some(Self {
+                kind,
+                width: other.width,
+            }),
+            (kind @ (Sk::Uint | Sk::Sint | Sk::Float), Sk::AbstractInt) => Some(Self {
+                kind,
+                width: self.width,
+            }),
```
teoxoy:
```suggestion
            (Sk::AbstractFloat, Sk::AbstractInt) => Some(other),
            (Sk::AbstractInt, Sk::AbstractFloat) => Some(self),

            // AbstractFloat converts to Float.
            (Sk::AbstractFloat, Sk::Float) => Some(other),
            (Sk::Float, Sk::AbstractFloat) => Some(self),

            // AbstractInt converts to concrete integer or float.
            (Sk::AbstractInt, Sk::Uint | Sk::Sint | Sk::Float) => Some(other),
            (Sk::Uint | Sk::Sint | Sk::Float, Sk::AbstractInt) => Some(self),
```

I find this easier to follow.

naga/src/front/wgsl/lower/conversion.rs
```diff
@@ -0,0 +1,378 @@
...
+}
+
+impl crate::Scalar {
+    /// Find the common type of `self` and `other` under WGSL's
+    /// automatic conversions.
+    ///
+    /// If there are any scalars to which WGSL's automatic conversions
+    /// will convert both `self` and `other`, return the best such
+    /// scalar. Otherwise, return `None`.
+    pub const fn automatic_conversion_join(self, other: Self) -> Option<crate::Scalar> {
```
teoxoy:
```suggestion
    pub const fn best_common_conversion(self, other: Self) -> Option<crate::Scalar> {
```

Is this better?

teoxoy (commented):
Solid stuff!

Left a few suggestions.

teoxoy:
> Hmm. There is a slightly different way to approach this patch, which would be to make constant evaluation responsible for detecting when conversions are necessary and applying them. The WGSL front end is still going to need to do things like choose types for all those `Compose` expressions in `construction.rs`, for example, but once it's done so, it seems like the actual conversions could be done directly in `constant_evaluator.rs`, rather than having the front end create a bunch of `As` expressions which it trusts the constant evaluator to consume.
> 
> @teoxoy If you'd like me to try out that alternative, just let me know. I don't think it'd take long.

If it's possible that would be great as we are now injecting `As` expressions that just end up being evaluated by the constant evaluator later anyway.

naga/src/valid/expression.rs
jimblandy: You're right. Changed as suggested.

naga/src/front/wgsl/lower/conversion.rs
jimblandy: ... I have no excuse ...

naga/src/front/wgsl/lower/conversion.rs
jimblandy: I've removed most of these, and cleaned up the one I wanted to keep.

naga/src/front/wgsl/lower/conversion.rs
jimblandy: Very much so.

naga/src/front/wgsl/lower/conversion.rs
jimblandy: I want to have the phrase "automatic conversion" in there, because this is supposed to be limited the WGSL "automatic conversions".

jimblandy: But I agree that "join" is a bit jargony.

teoxoy (approved):

jimblandy:
@teoxoy Just added one more commit, nothing else changed.

teoxoy:
Looks good, let's land it.

PR #4745, opened 2023-11-22
[naga wgsl-in] Test hex float suffix handling corner case.

jimblandy:
Test Naga's WGSL front end's handling of `h` and `f` suffixes on hexadecimal float literals. WGSL permits these suffixes only if an exponent is present, because otherwise `f` suffixes can be confused with a hexadecimal digit.

**Checklist**

- [X] Run `cargo fmt`.


teoxoy (approved):
Nice!

PR #4747, opened 2023-11-22
[naga wgsl-in] Experimental 64-bit floating-point literals.

jimblandy:
In the WGSL front end, support an `l` suffix on floating-point literals to yield 64-bit integer literals.

Unless `Capabilities::FLOAT64` is selected, these literals will always be validated out, so it doesn't affect standards compliance.

This PR is questionable. I guess it's nice to have `f64` support to play with (edit: and the front and back ends already parse and emit an `f64` WGSL type), but it's hardly urgent. Who knows whether the spec will ever actually have `f64` support, and if it does, what the syntax will be?

But the truth is, I have ulterior motives for this patch. It's a little circuitous:

- On the abstract type branch, all constructor built-in functions know how to consume abstract values. Calls to scalar constructor built-in functions (say, `f32(5)`) are handled by generating `As` expressions and letting the constant evaluator convert abstract values to concrete values. So `f32(5)` constant-evaluates the AbstractInt `5` to the `f32` `5.0`.
- It happens that the test case `tests/out/wgsl/double-math-functions.frag.wgsl` contains expressions like `f64(1.0)`. It's translated from GLSL, which does have a `double` type.
- Naga should either accept `f64(1.0)` or reject it.
    - If Naga rejects it, then we cannot translate `tests/in/glsl/double-math-functions.frag` to WGSL. The GLSL snapshots aren't really set up to generate other output languages; I don't really want to change that. Fine, maybe we back this test out, it's hardly essential. But let's assume that we want to accept it.
    - If Naga accepts it, then the only reasonable semantics for this expression should be that applying the `f64` constructor built-in function to an AbstractFloat gives you an `f64`.
- `ConstantEvaluator::cast` doesn't happen to support casting to `f64`. Since it returns an error, the WGSL front end assumes that the expression is not constant-evaluatable, and leaves it in the IR as an`AbstractFloat` literal, which the validator then rejects.
- Expanding `ConstantEvaluator::cast` to support casting to `f64` allows the `out/wgsl` file to validate.
- However, it also means that `Literal::F64` values start making it to the WGSL backend, which errors out because **WGSL has no `f64` literal syntax.** Hence the present PR.

So it seems to me we have two options:

- Give WGSL syntax for 64-bit literals, and have it generate them in `tests/out/wgsl/double-math-functions.frag.wgsl` and then read and validate them.
- Remove `tests/in/glsl/double-math-functions.frag` and forget all this.

**Checklist**

- [X] Run `cargo fmt`.

### DIFF
--- a/naga/src/front/wgsl/parse/number.rs
+++ b/naga/src/front/wgsl/parse/number.rs
@@ -16,6 +16,8 @@ pub enum Number {
     U32(u32),
     /// Concrete f32
     F32(f32),
+    /// Concrete f64
+    F64(f64),
 }
 
 impl Number {
@@ -61,9 +63,11 @@ enum IntKind {
     U32,
 }
 
+#[derive(Debug)]
 enum FloatKind {
-    F32,
     F16,
+    F32,
+    F64,
 }
 
 // The following regexes (from the WGSL spec) will be matched:
@@ -104,9 +108,9 @@ fn parse(input: &str) -> (Result<Number, NumberError>, &str) {
     /// if one of the given patterns are found at the start of the buffer
     /// returning the corresponding expr for the matched pattern
     macro_rules! consume_map {
-        ($bytes:ident, [$($pattern:pat_param => $to:expr),*]) => {
+        ($bytes:ident, [$( $($pattern:pat_param),* => $to:expr),* $(,)?]) => {
             match $bytes {
-                $( &[$pattern, ref rest @ ..] => { $bytes = rest; Some($to) }, )*
+                $( &[ $($pattern),*, ref rest @ ..] => { $bytes = rest; Some($to) }, )*
                 _ => None,
             }
         };
@@ -134,6 +138,16 @@ fn parse(input: &str) -> (Result<Number, NumberError>, &str) {
             }
             start_len - $bytes.len()
         }};
+    }
+
+    macro_rules! consume_float_suffix {
+        ($bytes:ident) => {
+            consume_map!($bytes, [
+                b'h' => FloatKind::F16,
+                b'f' => FloatKind::F32,
+                b'l', b'f' => FloatKind::F64,
+            ])
+        };
     }
 
     /// maps the given `&[u8]` (tail of the initial `input: &str`) to a `&str`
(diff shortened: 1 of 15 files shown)


jimblandy:
Point in favor of this PR: It seems like the front and back ends already support an `f64` WGSL type. We just don't have a syntax for literals of that type.

If we do feel that adding experimental `f64` support is okay, then it would probably make sense to pair it with backend support for literals as well. There should then be snapshot tests for the literals too.

cwfitzgerald:
Speaking from community side, having support for this is one step closer to end-to-end f64 support, something that is asked for a decent amount, particularly from the scientifically minded folks.

naga/src/front/wgsl/parse/number.rs
```diff
@@ -286,11 +296,14 @@ fn parse(input: &str) -> (Result<Number, NumberError>, &str) {
 
                 let digits_with_sign = general_extract.end(bytes);
 
+                // Don't accept the `l` suffix here; a decimal point
+                // or exponent is required for `f64` literals. We
+                // might want to use `l` for 64-bit integers as well.
```
teoxoy: I think we should keep things consistent with `f`/`h` where it's not required to have a decimal point or exponent if the suffix has been specified.

The only way to do that is to disambiguate the syntax by having the suffixes map to a single type.

I like rust's suffixes but those don't fit in with WGSL's current suffixes.

I suggest using `lf` for `f64`.
Then, we can have `li` and `lu` for `i64`/`u64`.

We can also use the same `SK` (`S` = size, `K` = kind) scheme later for 8-bit and 16-bit types (with the only odd one out being `h` where `hf` would be equivalent).

teoxoy:
> - Give WGSL syntax for 64-bit literals, it generate them in `tests/out/wgsl/double-math-functions.frag.wgsl` and then read and validate them.

I'm also in favor of supporting `f64` 👍

teoxoy (changes requested):

jimblandy:
Okay, here's full front and back end support with an `lf` suffix.

teoxoy (approved):

PR #4748, opened 2023-11-22
Allow Tests to Expect Certain Panic or Validation Messages

cwfitzgerald:
**Connections**

#4746 is an example of a test that is currently hard skipped, while we could be checking "if it errors like this, ignore it". This provides the infrastructure for this.

**Description**

This adds the following to the test harness:
- You can now restrict a failure to a certain set of panic or validation error messages.
- You can now mark a test as "flaky" which will ignore any failures matching the failure case.

It also fixed a few misc issues:
- We were using 64 bit queries on GL when only 32 bit queries were supported.
- Refined failure cases.
- Fixed issue with trying to name occlusion queries on GL.

**Testing**

It is.


nical (approved):
Nice!

PR #4752, opened 2023-11-22
Logging cleanups in device/global.rs

nical:
**Description**

  - Clean up the trace-level logging for devices
  - Log errors in more places.
  - Log the descriptors for create_buffer and create_texture.
  - Make logged ids more concise
     - Logged ids go from `Id { index: 204, epoch: 1, backend: Vulkan }` to `Id(204,1,vk)`.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/id.rs
+++ b/wgpu-core/src/id.rs
@@ -155,12 +155,17 @@ where
 {
     fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
         let (index, epoch, backend) = self.unzip();
-        formatter
-            .debug_struct("Id")
-            .field("index", &index)
-            .field("epoch", &epoch)
-            .field("backend", &backend)
-            .finish()
+        let backend = match backend {
+            Backend::Vulkan => "vk",
+            Backend::Metal => "mtl",
+            Backend::Dx12 => "d3d12",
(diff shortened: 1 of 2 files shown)


wgpu-core/src/id.rs
```diff
@@ -154,13 +154,9 @@ where
...
     fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
-        let (index, epoch, backend) = self.unzip();
-        formatter
-            .debug_struct("Id")
-            .field("index", &index)
-            .field("epoch", &epoch)
-            .field("backend", &backend)
-            .finish()
+        let (index, epoch, _) = self.unzip();
+        write!(formatter, "Id({index},{epoch})")?;
```
cwfitzgerald: Could we keep the backend in here (maybe a shortened version of it [vk, 12, 11, mtl, gl])

cwfitzgerald (approved):

wgpu-core/src/id.rs
nical: Alright, done!

gents83:
@nical just pay attention that whenever we're logging drop now in fact we're just requesting the drop, but the resource will be really dropped only when last reference to it will be removed.

I don't know if this has to be clarified somehow with the tracing

PR #4718, opened 2023-11-17
[vk] check that adapters are Vulkan compliant

teoxoy:
Resolves https://github.com/gfx-rs/wgpu/issues/4715.

This PR is meant to cut down on the number of reports we get due to Vulkan implementations not conforming to the spec.

The only driver/hardware combinations I could find that are not compliant are:

- [Mesa's Amd/RADV on GFX6 & GFX7 GPUs](https://gitlab.freedesktop.org/mesa/mesa/-/blob/bafc27583cf7c2d136fa05764e5e2145fedbc9cc/src/amd/vulkan/radv_physical_device.c#L86-90)
- [Mesa's Intel/HASVK on Gen7 iGPUs (Bay Trail/Haswell/Ivy Bridge)](https://gitlab.freedesktop.org/mesa/mesa/-/blob/bafc27583cf7c2d136fa05764e5e2145fedbc9cc/src/intel/vulkan_hasvk/anv_device.c#L1004-1024)
- [Mesa's Nouveau/NVK](https://gitlab.freedesktop.org/mesa/mesa/-/blob/bafc27583cf7c2d136fa05764e5e2145fedbc9cc/src/nouveau/vulkan/nvk_physical_device.c#L566-571)
- [Mesa's Dozen/DZN](https://gitlab.freedesktop.org/mesa/mesa/-/blob/bafc27583cf7c2d136fa05764e5e2145fedbc9cc/src/microsoft/vulkan/dzn_device.c#L979-984)

Related bugs:

- https://github.com/gfx-rs/wgpu/issues/2027
- https://github.com/gfx-rs/wgpu/issues/2480
- https://github.com/gfx-rs/wgpu/issues/3422
- https://github.com/gfx-rs/wgpu/issues/4670

We can provide an instance flag for people to experiment with uncompliant implementations but it's not part of this PR.

PR #4754, opened 2023-11-22
wgpu-hal(vk): Add WGPU_ALLOW_NONCOMPLIANT_ADAPTER

i509VCB:
**Connections**
#4718

**Description**
This allows a noncompliant adapter per VK_KHR_driver_properties to be used if the WGPU_ALLOW_NONCOMPLIANT_ADAPTER environment variable is added.

**Testing**
This can only really be locally tested as there aren't exactly non-compliant drivers being shipped.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,10 @@ Bottom level categories:
 
 ## Unreleased
 
+### `WGPU_ALLOW_NONCOMPLIANT_ADAPTER` environment variable
+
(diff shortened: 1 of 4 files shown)


cwfitzgerald (approved):
Nice

teoxoy:
Thoughts on renaming this to: `ALLOW_UNDERLYING_NONCOMPLIANT_ADAPTER`?

I think people might get the wrong idea that the wgpu adapter would be noncompliant; it could be, since the underlying one is but the current name could also be interpreted as: all `DownlevelFlags` are present if `ALLOW_NONCOMPLIANT_ADAPTER=0`.

cwfitzgerald:
Good idea

PR #4755, opened 2023-11-23
[naga wgsl-in] Automatic conversions for `var` initializers.

jimblandy:
Draft because it depends on #4743.

**Checklist**

- [X] Run `cargo fmt`.
- [X] Run `cargo clippy`.


jimblandy:
Seems like the global var and const cases have some common code that could be factored out.

jimblandy:
The new tests are kind of all over the place. At first I was trying to be exhaustive, but it started to get out of hand, so I checked the actual code coverage we were getting from `snapshots` and `wgsl_errors` for the changes in this PR, and it was fine. So I left the tests as they stood.

In the future, I think we should try to use coverage more to guide our test authoring.

jimblandy:
I believe the Linux failure is #4825, and this PR is safe to land.

teoxoy (approved):
LGTM

PR #4756, opened 2023-11-23
Move to A Single Example Executable

cwfitzgerald:
**Connections**

This is a continuation of an attempt to clean up our examples and our workflow.

**Description**

We've had multiple issues caused by all of our examples being separate crates (and binaries in general).
- Large memory usage at the end of test building due to redundant linking. This can often exceed the memory a 4gb-16gb machine has, and generally includes a ton of memory pressure.
- All this redundant linking can hurt incremental compilation times, especially on highly parallel machines.
- With rust's static linking, all of those means larger target directory.

To solve this, this is multiple steps to reduce the amount of test binaries.
- Combined the examples into a single crate. This also just feels a lot cleaner.
- Combined naga integration tests into a single root integration test.
- Turned off tests on a few crates that will never have inline tests, removing those links.

M1 Before:
- 60s clean build
- 362 clean build jobs
- 43 total test binaries
- 9.0s incremental build (string change in wgpu-hal)

M1 After:
- 56s clean build (-4s)
- 308 clean build jobs (-54)
- 13 total test binaries (-30)
- 6.5s incremental build (-2.5s)

Now when you run an example you do:

```
cargo run --bin wgpu-examples cube
```

To access an example on the web, we present a menu of the examples to the user where they can choose an example based on query parameter.

**Testing**

Manually tested.

**Checklist**

- [x] Run `cargo fmt`.


nical (approved):

cwfitzgerald:
Naga review unneeded, we previously discussed the tiny changes, and they are accepted.

tests/src/image.rs
```diff
@@ -11,7 +11,7 @@ async fn read_png(path: impl AsRef<Path>, width: u32, height: u32) -> Option<Vec
         Ok(f) => f,
         Err(e) => {
             log::warn!(
-                "image comparison invalid: file io error when comparing {}: {}",
+                "image comparison invalid: file io error when 3comparing {}: {}",
```
teoxoy, line 14: typo?

cwfitzgerald: Heh, change to test recompilation that snuck in

FrankenApps:
It's actually 30 test binaries less.

cwfitzgerald:
Math is hard :)

tests/src/image.rs
cwfitzgerald: Fixed in https://github.com/gfx-rs/wgpu/pull/4765

PR #4758, opened 2023-11-23
[vk] remove (old) unused blocklist for dual source blending

teoxoy:
As far as I could find, the Vulkan driver at the time were still in beta. It should have been fixed by now.


nical:
I'd feel better about check that users have the right driver version. Our experience with Firefox is that users typically don't update their drivers (often because they can't).

teoxoy:
The issue was only with Intel's Vulkan drivers on Windows; we are targeting D3D12 on Windows.

nical (approved):
Ok well we can always add some more fine-grained blocklisting later if we run into issues.

PR #4759, opened 2023-11-23
Add feature float32-filterable

almarklein:
**Connections**

* https://github.com/gfx-rs/wgpu-native/pull/322
* https://github.com/gfx-rs/wgpu/issues/2170
* https://github.com/pygfx/pygfx/issues/618

**Description**

The `float32-filterable` is currently the only remaining webgpu feature not supported by wgpu-core (and wgpu-native).

(BTW: `bgra8unorm-storage` has been added recently and can be ticked in #2170.)

**Testing**

I compiled wgpu-native with this change included, ran a volume rendering example in pygfx (via wgpu-py). The example works when I activate this feature, and does not when I don't. (We've been using `TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES` as a workaround).

Maybe this also needs a unit test?

**Checklist**

- [x] Run `cargo fmt`.
- [x] Activate/check feature on each gfx-hal backend:
    - [x] Metal
    - [x] Vulkan 
    - [x] dx11
    - [x] dx12
    - [x] gles
    - [ ] Need checking.
- [x] Unit test(s)


teoxoy:
> gles. Looks like it's automatically enabled when `TEXTURE_FLOAT_LINEAR` ext is available. Is that correct?

I think it is.

This should always be available on d3d12. Vulkan needs a check (for `SAMPLED_IMAGE_FILTER_LINEAR`).

Background: https://github.com/gpuweb/gpuweb/issues/3601#issuecomment-1432303013 & https://github.com/gpuweb/gpuweb/issues/3556#issuecomment-1431979258

wgpu-hal/src/metal/adapter.rs
```diff
@@ -612,6 +612,7 @@ impl super::PrivateCapabilities {
             function_specialization: Self::supports_any(device, FUNCTION_SPECIALIZATION_SUPPORT),
             depth_clip_mode: Self::supports_any(device, DEPTH_CLIP_MODE),
             texture_cube_array: Self::supports_any(device, TEXTURE_CUBE_ARRAY_SUPPORT),
+            supports_float_filtering: device.supports_32bit_float_filtering(),
```
```suggestion
            supports_float_filtering: os_is_mac
                || (version.at_least((11, 0), (14, 0), os_is_mac)
                    && device.supports_32bit_float_filtering()),
```

This is always available on macOS.
We also need to gate the usage of `supports32BitFloatFiltering` as it's not always available.

wgpu-types/src/lib.rs
```diff
@@ -337,7 +337,17 @@ bitflags::bitflags! {
...
         // ? const NORM16_FILTERABLE = 1 << 17; (https://github.com/gpuweb/gpuweb/issues/3839)
         // ? const NORM16_RESOLVE = 1 << 18; (https://github.com/gpuweb/gpuweb/issues/3839)
-        // TODO const FLOAT32_FILTERABLE = 1 << 19;
+        
+        /// Allows textures with formats "r32float", "rg32float", and "rgba32float" to be filterable.
+        ///
+        /// Supported Platforms:  (TODO is this correct?)
+        /// - Vulkan
+        /// - DX12
+        /// - Metal
```
```suggestion
        /// Supported Platforms:
        /// - Vulkan (mainly on Desktop GPUs)
        /// - DX12
        /// - Metal on macOS or Apple9+ GPUs, optional on iOS/iPadOS with Apple7/8 GPUs
        /// - GL with one of `GL_ARB_color_buffer_float`/`GL_EXT_color_buffer_float`/`OES_texture_float_linear`
```

crowlKats (commented):
Deno part LGTM

wgpu-types/src/lib.rs
```diff
@@ -3172,9 +3182,14 @@ impl TextureFormat {
...
+        
+        let sample_type = self.sample_type(None);
+        let mut is_filterable = 
+            sample_type == Some(TextureSampleType::Float { filterable: true });
+        if device_features.contains(Features::FLOAT32_FILTERABLE) {
+            is_filterable |= sample_type == Some(TextureSampleType::Float { filterable: false });
+        } 
 
-        let is_filterable =
-            self.sample_type(None) == Some(TextureSampleType::Float { filterable: true });
```
teoxoy: This needs to happen in `sample_type` since it's used by validation.

teoxoy (changes requested):

wgpu-types/src/lib.rs
teoxoy: Also, note that the formats are not blendable. Before this feature, it was always the case that filterable formats always supported blending as well but not anymore.

almarklein: > This needs to happen in `sample_type` since it's used by validation.

I need some help moving this logic into `sample_type`. To decided whether a format that is normally `unfilterable` is currently indeed `filterable`, we need the `device_features` object at hand. We cannot simply change the signature since this method is called from several different places in the codebase.

I'm also a bit confused. Since it's here  (in the current method `guaranteed_format_features()`) that we set the filterable and blendable `TextureFormatFeatureFlags`. So we need to make adjustments here too.
 
Also, for my understanding, I've tested and it  *does* work as it's written now. Is that expected or are there cases where it would not.

almarklein:
I implemented setting the adapter feature for all hal backends. These really need double-checking - I have no idea whether this is right (e.g. dx11?).

wgpu/src/backend/web.rs
```diff
@@ -701,6 +701,10 @@ const FEATURES_MAPPING: [(wgt::Features, web_sys::GpuFeatureName); 10] = [
         wgt::Features::BGRA8UNORM_STORAGE,
         web_sys::GpuFeatureName::Bgra8unormStorage,
     ),
+    (
+        wgt::Features::FLOAT32_FILTERABLE,
+        web_sys::GpuFeatureName::Float32Filterable,
+    ),
```
teoxoy, line 709: The array size needs updating a few lines above (CI is also running into this), can't comment there though.

wgpu-hal/src/gles/adapter.rs
```diff
@@ -549,6 +549,13 @@ impl super::Adapter {
             );
         }
 
+        features.set(
+            wgt::Features::FLOAT32_FILTERABLE,
+            extensions.contains("GL_ARB_color_buffer_float")
+                || extensions.contains("GL_EXT_color_buffer_float")
+                || extensions.contains("OES_texture_float_linear"),
+        );
+
```
teoxoy, line 558: This looks good but we can now remove `PrivateCapabilities::TEXTURE_FLOAT_LINEAR` and just check for the presence of `Features::FLOAT32_FILTERABLE`.

teoxoy:
> I implemented setting the adapter feature for all hal backends. These really need double-checking - I have no idea whether this is right (e.g. dx11?).

Thanks! The d3d11 backend is not currently working, and it's unclear if we need it anymore now that OpenGL is also working on Windows without Angle.

@cwfitzgerald is there still value in keeping the d3d11 backend around?

wgpu-types/src/lib.rs
teoxoy: > > This needs to happen in `sample_type` since it's used by validation.
> 
> I need some help moving this logic into `sample_type`. To decided whether a format that is normally `unfilterable` is currently indeed `filterable`, we need the `device_features` object at hand. We cannot simply change the signature since this method is called from several different places in the codebase.
> 
> I'm also a bit confused. Since it's here (in the current method `guaranteed_format_features()`) that we set the filterable and blendable `TextureFormatFeatureFlags`. So we need to make adjustments here too.

Passing `device_features` to `sample_type` is what I had in mind, I also looked around where `sample_type` is used and see no issue passing the set of features to it. There are a few places where it doesn't matter if the sample type is filterable or not, for those cases we can even pass it `Features::empty()`.

> Also, for my understanding, I've tested and it _does_ work as it's written now. Is that expected or are there cases where it would not.

The feature does work but validation won't work properly, I suspect this is why CI is also failing on the new `float32_filterable_without_feature` test.

crowlKats (commented):
Deno part LGTM

wgpu-hal/src/gles/adapter.rs
almarklein: 👍

wgpu-types/src/lib.rs
almarklein: I'm getting stuck at adding the argument to `TextureFormat.sample_type()`. E.g how do I get a reference to the device features here?
https://github.com/gfx-rs/wgpu/blob/f9711831d28e751f463374fa7a172c93e0d21276/wgpu-hal/src/gles/command.rs#L608-L608

teoxoy: This is one of the cases where it doesn't matter whether or not the format is filterable and passing `Features::empty()` is fine (it would be worth adding a comment though).

cwfitzgerald:
@teoxoy https://github.com/gfx-rs/wgpu/issues/4797 no :)

Fancy2209:
> @teoxoy #4797 no :)

You mean wgpu isn't removing DX11?

teoxoy:
> > @teoxoy #4797 no :)
> 
> You mean wgpu isn't removing DX11?

This was in response to:

> @cwfitzgerald is there still value in keeping the d3d11 backend around?

teoxoy:
@almarklein I think the implementation looks good but I think the tests are still failing because we might allow usage of features even if they have not been explicitly enabled by users...

I'll look into it a bit more and put up a PR for it.

almarklein:
That's what I thought too. I am just prepping a Windows machine to investigate this further and see where I can best add a check for this (suggestions are welcome).

teoxoy:
Actually that will only happen in the case of the OpenGL ES backend.

https://github.com/gfx-rs/wgpu/blob/418493239a8b24163587a027bcf4a00da25f91fb/wgpu-core/src/device/resource.rs#L3207-L3215

If you could figure out why Windows and Vulkan are failing that would be great.

almarklein:
So it look like for the GL and DX12 backends `downlevel` is `true`, causing it to use `get_texture_format_features` and thus bypassing the checks that I added in `guaranteed_format_features` 😄 

If I add checks here (in the code you referred to) all tests pass on my Windows machine. Does this mean it makes more sense to just apply the checks here instead? Or do we still need them in `sample_type()`?

almarklein:
> Does this mean it makes more sense to just apply the checks here instead? Or do we still need them in sample_type()?

In retrospect I don't think it makes sense. Instead I made the `device_features` arg optional, so the code that provides the `Features::empty()` can pass `None` instead, which feels less awkward.

CI is green!

teoxoy (approved):
Great stuff!

teoxoy:
CI is running into https://github.com/gfx-rs/wgpu/pull/4573#issuecomment-1838575468, not related to this PR.

PR #4766, opened 2023-11-24
[spv-out] decorate non-uniform image/sampler access chains

### LINKED ISSUE #4762: Texture sampling issues when using a texture binding_array on Windows+AMD gpu with wgpu 0.18
**Description**
Textures look borked when sampling from a `binding_array<texture_2d<f32>>` on Windows+AMD gpu+wgpu 0.18+vulkan.

**Repro steps**
Run bevy's [`texture_binding_array`](https://github.com/bevyengine/bevy/blob/main/examples/shader/texture_binding_array.rs) ([shader code](https://github.com/bevyengine/bevy/blob/main/assets/shaders/texture_binding_array.wgsl)) example on Windows with an AMD gpu.

broken sampling

**Extra materials**

The wgsl: https://gist.github.com/Elabajaba/783e298cff814526a4e9e17031a7346f

A similar bug happened on Linux+AMD (RADV) with wgpu 0.17, but is fixed with wgpu 0.18.
This bug only appeared for me on windows+amd with wgpu 0.18 + vulkan.

**Platform**
Windows 11, AMD 6800xt with the 23.11.1 driver.

teoxoy:
See https://github.com/gfx-rs/wgpu/issues/4762#issuecomment-1825522864.

Fixes https://github.com/gfx-rs/wgpu/issues/4762.
Reverts https://github.com/gfx-rs/wgpu/commit/c3e35df576af8cdd345875b8af1ab541c43defd7#diff-42823a07fbb2862baa61487d0fe17c6ba02e2db32560e884875449114897d975L1459-R1480.


Elabajaba:
Tested this and it fixed the issue on windows+amd+vulkan for me, and didn't regress linux (radv)+amd+vulkan.

jimblandy (changes requested):
I'm confused - the patch doesn't match the commit message or the discussion.

The code looks like it has always been decorating the results of `OpAccessChain`, just not for binding arrays in the `Handle` address space.

The discussion suggests that both the pointer and the result of the load need to be decorated, but we are not decorating the result of the load / sample at all, as far as I can see.

Can't we do better at helping ourselves understand what's going on the next time we have to look at this?

teoxoy:
> The code looks like it has always been decorating the results of `OpAccessChain`, just not for binding arrays in the `Handle` address space.

The code originally decorated all access chains, https://github.com/gfx-rs/wgpu/commit/c3e35df576af8cdd345875b8af1ab541c43defd7 changed it so that it only happened for storage and uniform buffers since I thought that decorating the image/sampler access chains was not necessary. We now know this was necessary and is what this PR is reverting.

> The discussion suggests that both the pointer and the result of the load need to be decorated, but we are not decorating the result of the load / sample at all, as far as I can see.

Since https://github.com/gfx-rs/wgpu/commit/c3e35df576af8cdd345875b8af1ab541c43defd7 we decorate the load result of image/sampler binding arrays (the load result being an image/sampler); this is part of the VUID requirement.

The results of texture load/sampling operations don't have to be decorated.

jimblandy (approved):
Okay, I understand this better now. But please update the commit message.

teoxoy:
How should the commit message be updated?

jimblandy:
@teoxoy Here's my suggestion (summary line and detailed explanation):

> [spv-out] Decorate all non-uniform binding array access chains as `NonUniform`, not just buffer binding arrays.
>
> Apply the `NonUniform` decoration to the results of all access chains rooted in binding arrays that use non-uniform values as indices, regardless of the binding array's element type and address space. Previously, Naga only decorated non-uniform access chains for binding arrays of buffers.

The important difference is that this message doesn't mention load results, which this PR has no effect on (and which Naga has never decorated as far as I can tell).

PR #4769, opened 2023-11-24
Conditionally lift API logging from trace to info level

nical:
Most of this logging used to be info level until I demoted it to trace. Unfortunately this gets in my way because:
 - Most of the logging I currently need is these API entry points, there is is a fair amount of more verbose logging in wgpu at higher levels than trace
 - Firefox disable all trace and debug logging for optimized builds, which means I miss the this API logging where I need most.

This patch lifts the API logging back to info level.
It also adds a few missing logs that I found along the way.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/Cargo.toml
+++ b/wgpu-core/Cargo.toml
@@ -29,6 +29,8 @@ targets = [
 
 [features]
 default = ["link"]
+# Log all API entry points at info instead of trace level.
+api_log_info = []
 
 # Backends, passed through to wgpu-hal
 metal = ["hal/metal"]
--- a/wgpu-core/src/command/clear.rs
+++ b/wgpu-core/src/command/clear.rs
@@ -76,7 +76,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         size: Option<BufferAddress>,
     ) -> Result<(), ClearError> {
         profiling::scope!("CommandEncoder::clear_buffer");
-        log::trace!("CommandEncoder::clear_buffer {dst:?}");
+        log::info!("CommandEncoder::clear_buffer {dst:?}");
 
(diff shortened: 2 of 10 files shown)


nical:
If this is controversial, an alternative is to add a new macro, say `api_log!`, which would desugar into either `log::info` or `log::trace` depending on a build flag.

Wumpf:
If we put these on info we essentially force every user of wgpu(-core) to disable wgpu-core logging other than warnings. Having a "you entered this call" be info instead of trace makes semantically no sense imho, there's no worthwhile info contained, it is literally a trace of what's happening.
With that in mind the `api_log!` solution is still _weird_ imho but much more workable.

Wumpf (changes requested):
also, we still have an open ticket on having _less_ things on info:

* https://github.com/gfx-rs/wgpu/issues/3206

nical:
I find it more useful than most of the info logging in wgpu_core, but I understand where you are coming from. `api_log` macro it is then.

wgpu-core/Cargo.toml
```diff
@@ -29,6 +29,7 @@ targets = [
 
 [features]
 default = ["link"]
+api_log_info = []
```
Wumpf, line 33: I know we have the bad habit in this repo not to, but can we document what the feature does here please?

Wumpf (approved):
thank you!

nical:
Reading the table in https://github.com/gfx-rs/wgpu/issues/3206, It's clear that `debug` is actually most appropriate by default for these logs (although I still need them at `info` or above in gecko).

nical:
This PR is rather uncontroversial now that it doesn't change the current behavior by default. Merging.

PR #4771, opened 2023-11-24
Downgrade some of wgpu_core's logging level from info to trace and debug

nical:
**Connections**

#3206

**Description**

 - Most of the log that was very specific to some wgpu internals was downgraded to `trace`.
 - A few logs that provided more rare contextual information were downgraded to `debug`.
 - The main remaining per-frame `info` log is relative to the internal resource lifetime management (which currently has bugs that I am investigating so I'll handle them separately in another PR, probably similarly to the api log until the issues are resolved).

I didn't audit the existing `debug` logging.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


Wumpf (approved):
yasss!! <3

PR #4772, opened 2023-11-24
Downgrade resource lifetime management log level to trace.

nical:
Allow promoting it back to info via an feature flag.

**Connections**

#3206

**Description**

Wrap the resource-lifecycle related logging in a `resource_log` macro which defaults to `trace` level but can be lifted to `info` via a feature flag.
Being able to lift them to `info` is important for Firefox integration reasons, and that's an area I am currently investigating, but once the lifecycle bugs are fixed, I'll likely remove the feature flag and let them be simple `log::trace!` calls (unless anyone prefers to keep it this way).

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


nical:
With this PR, enabling `info` logs in the water example for `wgpu_core` no longer spams the log with content every frame.

Wumpf (approved):
when you started these logging changes I was concerned as I wrote, but loving where this is going now! Also, having those gaps where didn't log at all filled with trace logs by default is also super useful! <3

PR #4780, opened 2023-11-26
Fixes for Skybox and All Examples

cwfitzgerald:
**Description**

The skybox example on firefox was trying to use uncompressed textures, but webgl doesn't support brga textures.

In an attempt to regenerate the textures to not be brga, I re-stumbled down the rabbit hole that is compressed textures. I've now properly put the generation process into a bash script so that I don't have to re-do the same research next time.

Additionally, the examples no longer use `Poll` control flow on winit, which is no longer recommended in 0.29.

**Testing**

Both manual and automated testing. 

**Checklist**

- [x] Run `cargo fmt`.


teoxoy (approved):
Nice!

PR #4781, opened 2023-11-26
remove_abandoned fix

gents83:
Whenever a resource is not inside a tracker we can allow the triage resource to remove it.
In the example shadow we saw that surface texture was continously not removed due to this wrong check not allowing the texture to be dropped
@scoopr discovered it - tnx a lot 👍

### DIFF
--- a/wgpu-core/src/track/buffer.rs
+++ b/wgpu-core/src/track/buffer.rs
@@ -339,11 +339,11 @@ impl<A: HalApi> ResourceTracker<BufferId, Buffer<A>> for BufferTracker<A> {
                         id,
                         existing_ref_count
                     );
+                    return false;
                 }
             }
         }
-
-        false
+        true
     }
 }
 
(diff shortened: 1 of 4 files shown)


nical (changes requested):
The comment at the top of the function needs to be updated (and should explain what's going on with the special case)

wgpu-core/src/track/buffer.rs
```diff
@@ -308,8 +308,7 @@ impl<A: HalApi> ResourceTracker<BufferId, Buffer<A>> for BufferTracker<A> {
     /// `triage_suspected` will remove 3), leaving 1) as the sole
     /// remaining reference.
     ///
-    /// Return `true` if this tracker contained the buffer `id`. This
-    /// implies that we removed it.
+    /// Returns true if the resource was removed or if not exiting in metadata.
```
nical: nit: existing

nical (approved):
Thanks!

PR #4739, opened 2023-11-21
Revert changes to `zero_init_texture_after_discard`

teoxoy:
See https://github.com/gfx-rs/wgpu/pull/3626#issuecomment-1819169439.

PR #4782, opened 2023-11-26
Remove resources ONLY when needed inside wgpu and not in user land

### LINKED ISSUE #4245: Resource state between subsequent commands
While debugging test **zero_init_texture_after_discard::discarding_either_depth_or_stencil_aspect()** with _Arcanization CL_ #3626

I noticed that the following code was causing in Compute pass to have the resource in a wrong state compared from the one present in the device (it's initialized to RESOURCE instead of DEPTH_STENCIL_WRITE).

```
            case.create_command_encoder();
            case.discard_depth();
            case.submit_command_encoder();

            case.create_command_encoder();
            case.discard_stencil();
            case.submit_command_encoder();

            case.create_command_encoder();
            case.copy_texture_to_buffer();
            case.submit_command_encoder();
```
            
And I was forced to have a unique command buffer with all operations in the same flow in order to avoid the issue, and have not validation error trigger on wrong resource state:

```
            case.create_command_encoder();
            case.discard_depth();
            case.discard_stencil();
            case.copy_texture_to_buffer();
            case.submit_command_encoder();
```

So may be worth to check resource state between different command buffers.
I added following command into the code to keep track of it (other then adding this issue):

```
            //When splitting it in subsequent submits of different command encoders
            //it seems that texture tracker is not able anymore to get that the texture has been left in DEPTH_STENCIL_WRITE
            //and it assume that it could find it uninitialized and set it in RESOURCE as not owning it
            //When using a unique submit instead the tracker is able to follow all resource barriers and everything is smooth
```

### LINKED ISSUE #4740: CI Flakes on VK discarding_either_depth_or_stencil_aspect_test
```
  TRY 3 FAIL [   0.460s] wgpu-test::gpu-tests [Executed] [Vulkan/llvmpipe (LLVM 15.0.7, 256 bits)/0] gpu_tests::zero_init_texture_after_discard::discarding_either_depth_or_stencil_aspect_test

thread '<unnamed>' panicked at 'texture was not fully cleared', tests/tests/zero_init_texture_after_discard.rs:306:9
```

https://github.com/gfx-rs/wgpu/actions/runs/6947331918/job/18900887783?pr=4723

gents83:
After investigations we can definitely revert changes to zero_init_texture_after_discard and define that resources are definitely removed from trackers only if are kept just from tracker and not from user land.

- fixes #4245
- closes #4739 (done as part of this PR)
- fixes #4740 (reverts #4746)

cc @cwfitzgerald & @teoxoy 

I've also done some cleanup of code not needed anymore now that we're back to a device-tracker-only check 👍


cwfitzgerald (approved):
Approving, but want one more review before landing!

This is a much better solution and feels so much cleaner!

teoxoy:
This PR should revert https://github.com/gfx-rs/wgpu/pull/4746 and also undo the wait (https://github.com/gfx-rs/wgpu/pull/4739) as it shouldn't be necessary.

wgpu-core/src/track/buffer.rs
```diff
@@ -327,8 +327,7 @@ impl<A: HalApi> ResourceTracker<BufferId, Buffer<A>> for BufferTracker<A> {
                 let existing_ref_count = self.metadata.get_ref_count_unchecked(index);
                 //2 ref count if only in Device Tracker and suspected resource itself and already released from user
                 //so not appearing in Registry
```
Wumpf: this comment is super important and a bit cryptic to me. If I understand correctly "a resource is exactly then unused if it is solely owned by the device tracker itself and by the caller (originating from the suspected to be destroyed list), i.e. up to two strong references"?

Also, the method comment still talks about three references..?

Wumpf (approved):
>  one more review before landing!

Looks good to me and a loot nicer. After catching up on Matrix chat a bit I think this makes all sense. But was quite hard to catch up there, so not overly confident that I got this all

wgpu-core/src/track/buffer.rs
gents83: I read it and I agree - it's not really clear :)
Would be better something like:
"RefCount 2 means that resource is hold just by DeviceTracker and this suspected resource itself - so it's already been released from user and so it's not inside Registry\Storage"

nical (approved):

wgpu-core/src/track/buffer.rs
nical: I like this version (the one you are proposing) better.

teoxoy (changes requested):
See https://github.com/gfx-rs/wgpu/pull/4782#issuecomment-1827574777.

Should have marked as "request changes", sorry.

gents83:
> See [#4782 (comment)](https://github.com/gfx-rs/wgpu/pull/4782#issuecomment-1827574777).
> 
> Should have marked as "request changes", sorry.

Is it something to be done in this PR (if so, I need more info) or is something to do manually afterwards?

teoxoy:
My request was to:

- revert the changes in https://github.com/gfx-rs/wgpu/pull/4746/commits/b8e73f4106f0a422886dade1ac0d727211987029.
- apply the changes in https://github.com/gfx-rs/wgpu/pull/4739/commits/e97ac96a1ce3f511daaabd60a32cbcb243697b96.

I think they should be part of this PR.

gents83:
@teoxoy should be fine now - tnx

teoxoy (approved):
Looks good, thanks!

PR #4786, opened 2023-11-27
Don't keep a strong ref in storage for destroyed resources

gents83:
Changing Destroyed(Arc<T>, Epoch) into Destroyed(Epoch) to actually release that reference


gents83:
Depending on #4782

wgpu-core/src/storage.rs
```diff
@@ -213,15 +213,12 @@ where
                 std::mem::replace(slot, Element::Vacant)
```
nical: Nit: with this change you can make this even simpler, no need for the borrow check workaround, you can `std::mem::replace(slot, Element::Destroyed(storage_epoch))` directly without the intermediate `Vacant` step.

nical (approved):
Let's merge #4782 first, then rebase this on top of land it.

nical (approved):
Neat!

PR #4793, opened 2023-11-28
[wgsl-in] add support for override declarations

teoxoy:
Adds support for override declarations and splits overrides from constants in the IR.


naga/tests/in/overrides.wgsl
jimblandy, line 1: We also need a test like this:

```
override inferred_f32 = 3.1415;
```

When we merge this branch with the abstract type changes, we will need to make sure that lowering uses the *concretized* initializer value to infer the override's type, not its value in isolation. I think this test will catch that case, so we'll know to fix it when the merge happens.

Also, let's give this file a final newline.

jimblandy (approved):
The compaction can be simplified, and I think one of the tests should be tweaked slightly, but otherwise this looks great. I'll make the changes myself.

PR #4796, opened 2023-11-28
Remove surface extent validation (and thus fix the annoying `Requested size ... is outside of the supported range` warning)

Wumpf:
**Connections**
-

**Description**

Removes surface extent validation since its inherently racy: The user has no way of getting the size that we expect in there. Also our expectation might be outdated already anyways.

Enable `VUID-VkSwapchainCreateInfoKHR-imageExtent-01689` vulkan validation warning - according to spec this is ONLY about zero size surfaces and we correctly validate against this.
Added exception for `VUID-VkSwapchainCreateInfoKHR-pNext-07781` instead which triggers when configuring a surface with the wrong size.

**Testing**

ran hello_triangle, didn't get the validation there to begin with, but it's gone so I hardly gonna see it again in the egui based applications I've been hitting this on Linux and Windows😄 

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-hal/src/gles/adapter.rs
+++ b/wgpu-hal/src/gles/adapter.rs
@@ -798,7 +798,6 @@ impl super::Adapter {
                     workarounds,
                     features,
                     shading_language_version,
-                    max_texture_size,
                     next_shader_id: Default::default(),
                     program_cache: Default::default(),
                     es: es_ver.is_some(),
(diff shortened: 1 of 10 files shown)


wgpu-hal/src/gles/adapter.rs
```diff
@@ -1144,15 +1143,6 @@ impl crate::Adapter<super::Api> for super::Adapter {
...
                 current_extent: None,
-                extents: wgt::Extent3d {
-                    width: 4,
-                    height: 4,
-                    depth_or_array_layers: 1,
-                }..=wgt::Extent3d {
-                    width: self.shared.max_texture_size,
-                    height: self.shared.max_texture_size,
-                    depth_or_array_layers: 1,
-                },
```
cwfitzgerald, line 1155: Could you just make sure there is validation on configure that the texture is within the device texture size.

cwfitzgerald (approved):
Good after question

Wumpf:
@cwfitzgerald I'm using now the max texture device as configured on the device which may be smaller than the max supported size. So this is technically different than what the warning did before, but I think that's the only sane way. You might want to have a look on the last commit before merging

Issue #3103, opened 2022-10-13
Precompiled Shaders


dfrg:
I am working with @raphlinus on [piet-gpu](https://github.com/linebender/piet-gpu) and we're considering replacing our internal HAL with wgpu-hal but we would like to retain the ability to use precompiled shaders to reduce both binary size and startup cost.

After some investigation, it seems the least invasive solution would be to add variants to `ShaderInput` for MSL source (or potentially a `MTLLibrary` instance) and DXIL and feed those through to pipeline creation in the same way that `SpirV` works for Vulkan today. This would require the user to assign bindings that are compatible with those generated by wgpu-hal during pipeline layout creation. For our purposes, this would be an acceptable constraint.

Alternatively, we've also considered a more sophisticated approach that would allow the user to provide a bind mapping during shader module creation. This seems to require deferring native resource binding until the whole pipeline state is known so that discrepancies between the pipeline layout and the expected binding slots for the entry point can be reconciled.

My understanding is that the project is amenable to such a change, and we're willing to do the work, but wanted to provide an opportunity for other potential users who are interested in this feature to weigh in and develop a consensus on the best path forward.

cwfitzgerald:
We definitely would be interested in having this feature!

A couple considerations to keep in mind:

- In the mid-term future it will be possible to write custom wgpu-hal backends out-of-tree. This is particularly important for console API backends where NDAs are involved, but can also be useful for "trying weird stuff" kind of deals. This means we cannot easily just have enums of choices, as some choices may be unknown to us.
- I think we should keep the wgpu-authoritative binding model. While specifying the binding model is possible, it is a lot of information to pipe through and is inherently a very leaky abstraction, and suffers from the above problem.

As such I would propose an api that looks something like this:

```rust
// Replaces ShaderModuleDescriptorSpirv
pub struct ShaderModuleDescriptorRaw<'a> {
    pub label: Label<'a>,
    /// Raw data of whatever type the backend needs - likely Vec<u32> or a string. Maybe something fancier like a MTLLibrary wrapper.
    pub source: Box<dyn Any>, // Maybe a custom trait?
}

// Replaces create_shader_module_spirv
//
// Note the lack of reference.
pub unsafe fn create_shader_module_raw(&self, desc: ShaderModuleDescriptorRaw) -> ShaderModule
```

Then inside each wgpu-hal backend we can properly document how bindings get mapped based on all possible arguments to a pipeline creation.

Wandalen:
[...]

PR #4811, opened 2023-11-30
Expose shader validation

daxpedda:
I was trying to pre-compile shaders while still running the exact same validation that `wgpu` does.
Would it be acceptable to expose this module for that purpose?

### DIFF
--- a/wgpu-core/src/lib.rs
+++ b/wgpu-core/src/lib.rs
@@ -68,7 +68,7 @@ pub mod resource;
 pub mod storage;
 mod track;
 // This is public for users who pre-compile shaders while still wanting to
-// preserve all run-time checks that `wgpu-hal` does.
+// preserve all run-time checks that `wgpu-core` does.
 // See <https://github.com/gfx-rs/wgpu/issues/3103>, after which this can be
 // made private again.
 pub mod validation;
--- a/wgpu-core/src/lib.rs
+++ b/wgpu-core/src/lib.rs
@@ -67,7 +67,11 @@ pub mod registry;
 pub mod resource;
 pub mod storage;
 mod track;
-mod validation;
+// This is public for users who pre-compile shaders while still wanting to
+// preserve all run-time checks that `wgpu-hal` does.
+// See <https://github.com/gfx-rs/wgpu/issues/3103>, after which this can be
+// made private again.
+pub mod validation;
 
 pub use hal::{api, MAX_BIND_GROUPS, MAX_COLOR_ATTACHMENTS, MAX_VERTEX_BUFFERS};
 


Wumpf:
I think the better way to go about this is to call https://docs.rs/naga/0.14.1/naga/valid/struct.Validator.html#method.validate
which is most of what wgpu does. Is there a reason this would miss something in your case?

```
let module =
    naga::front::wgsl::parse_str(&shader.source).expect("should parse :O");

let mut validator = naga::valid::Validator::new(
    naga::valid::ValidationFlags::all(),
    naga::valid::Capabilities::all(),
);

validator.validate(&module).expect("should be valid!");
```

The nice thing about this is that you can then at least skip the parsing on native wgpu since you're allowed to pass in a naga module for creating the shader module https://github.com/gfx-rs/wgpu/blob/trunk/wgpu/src/lib.rs#L576

daxpedda:
What you are describing is a different use-case, one which I've used pretty successfully so far!

I'm talking specifically about passing raw SPIRV modules, which I would generate externally first. Wgpu currently does a bunch of validation, including using the Naga `Validator`, but that's not all it does. E.g. [`Interface::check_stage()`](https://github.com/gfx-rs/wgpu/blob/v0.18.1/wgpu-core/src/validation.rs#L931-L1196) does a lot of checking against the render pipeline, which is completely skipped when providing a raw SPIRV shader.

The goal is to replicate exactly what Wgpu does internally, most of this stuff is exposed and easy to re-use. But the whole validation going on in `wgpu_core::validation` will require me to almost copy the whole file. Which, alternatively, I can do as well.

I was told before that large parts of the `wgpu-core` and `wgpu-hal` API are not exactly considered stable, so I thought exposing this isn't an issue.

Cc https://github.com/gfx-rs/wgpu/issues/3103, which would make this pointless as well.

Wumpf:
ah I see, thanks for elaborating, makes sense! I'd be in favor of exposing then as well, but I'd like to hear what @cwfitzgerald says

wgpu-core/src/lib.rs
```diff
@@ -67,7 +67,7 @@ pub mod registry;
 pub mod resource;
 pub mod storage;
 mod track;
-mod validation;
+pub mod validation;
```
cwfitzgerald, line 74: Could you add a note why this is public and what would allow us to bring it private again?

cwfitzgerald (changes requested):
I don't love it, but can't argue against it.

wgpu-core/src/lib.rs
```diff
@@ -67,7 +67,11 @@ pub mod registry;
 pub mod resource;
 pub mod storage;
 mod track;
-mod validation;
+// This is public for users who pre-compile shaders while still wanting to
+// preserve all run-time checks that `wgpu-hal` does.
+// See <https://github.com/gfx-rs/wgpu/issues/3103>, after which this can be
+// made private again.
```
daxpedda: Let me know if this can be improved, unfortunately English isn't my first language.

wgpu-core/src/lib.rs
daxpedda: Done.
Continued in https://github.com/gfx-rs/wgpu/pull/4811#discussion_r1417940633.

wgpu-core/src/lib.rs
```diff
@@ -67,7 +67,11 @@ pub mod registry;
 pub mod resource;
 pub mod storage;
 mod track;
-mod validation;
+// This is public for users who pre-compile shaders while still wanting to
+// preserve all run-time checks that `wgpu-hal` does.
```
cwfitzgerald:
```suggestion
// preserve all run-time checks that `wgpu-core` does.
```

cwfitzgerald (approved):

wgpu-core/src/lib.rs
cwfitzgerald: One small typo, otherwise good!

Issue #3514, opened 2023-02-21
Add a default feature for wgpu that enables Vulkan & GLES on Windows & Linux


teoxoy:
We currently only enable backends based on the target platform with the only exception being GLES which is always available.

We should expose a new feature flag (enabled by default) that covers the GLES and DX11 backends so that users not interested in those backends can avoid compiling them in their binaries.

Feature name ideas:
- `legacy`
- `compat` (mapping to a future WebGPU Compat; might not be the best choice since we support even lower limits/features than what WebGPU Compat is aiming for; i.e. GLES 3.1+ and DX11 FL10+)

See background for this: https://github.com/gfx-rs/wgpu/issues/1221#issuecomment-1430067962

daxpedda:
I was recently looking into this but the first roadblock I hit is that we can't differentiate between a new `gles` crate feature and the current `angle` crate feature, both enable `wgc/gles`.

A solution could be to introduce target-specific features in `wgpu-core` and `wgpu-hal` that are no-op on certain platforms.
E.g. `gles-native` (no-op on MacOS) and `gles-angle` (no-op on everything but MacOS). This proposal would be similar to https://github.com/gfx-rs/wgpu/pull/3466.

I'd be happy to make a PR!

cwfitzgerald:
I think we're moving in the direction of every backend being feature flagged. We have all the infrastructure in place, we just haven't lifted this to the wgpu level.

Distinguishing between angle and native would be useful!

daxpedda:
So after playing around with it a bit I was so unsatisfied it doesn't make sense to me to proceed like that.
The issue basically boils down to Cargo not being able to connect crate features that activate optional dependencies to specific targets.

Introducing `gles-native` and `gles-angle` we can use `cfg`s to guard code to specific targets, but the dependencies these features activate will be active nevertheless. E.g. if we use `default = ["gles-native"]`, OpenGL won't work on MacOS, but the dependencies will still be present.

So the only good solution I can come up with to this problem is to create a separate crate that adds these defaults.
```toml
[package]
name = "wgpu-default-features"

# We want the wgpu-core Metal backend on macOS and iOS.
[target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies.wgc]
workspace = true
features = ["metal"]

# We want the wgpu-core Direct3D backends and OpenGL (via WGL) on Windows.
[target.'cfg(windows)'.dependencies.wgc]
workspace = true
[...]

PR #4815, opened 2023-12-02
Add `wgpu` crate features for backends

daxpedda:
This adds crate features `dx12` and `metal` to `wgpu`, which are enabled by default.
`dx12` is no-op unless used on Windows.
`metal` is no-op unless used on MacOS/iOS.

Vulkan and OpenGL can currently not guarded by such a crate feature because it would mess up our defaults, e.g. a `gles` crate feature can not be enabled by default without also enabling if for MacOS and Windows, which is currently not what we want. See https://github.com/gfx-rs/wgpu/issues/3514#issuecomment-1836885698 for more details.

Cc #3514.


daxpedda:
I will add some documentation upon some initial feedback if we can proceed like this.

CI already tests `--no-default-features`, but doesn't test all sorts of crate feature combinations. Let me know if you want me to change/add something there.

daxpedda:
I believe CI failure is unrelated.

wgpu/Cargo.toml
```diff
@@ -47,6 +47,10 @@ fragile-send-sync-non-atomic-wasm = [
 ]
 # Log all API entry points at info instead of trace level.
 api_log_info = ["wgc/api_log_info"]
+# Backends
```
Wumpf, line 44: would be nice to have all the backend related features in the same place and give them a doc string, i.e. move `angle` and `webgl` down. I want to soon enable proper feature documentation, it's quite messy atm.

wgpu/Cargo.toml
```diff
@@ -64,12 +68,11 @@ features = ["raw-window-handle"]
 # We want the wgpu-core Metal backend on macOS and iOS.
```
Wumpf: comment here and on dx backends outdated now

Wumpf (changes requested):
I like where this is going (having an explicit feature for every platform backend) but I feel like we should first figure out what to do with Vulkan and Gl, otherwise this gets _even more_ confusing for anyone trying to use wgpu and make a choice of feature flag.
Let's try finishing the discussion on https://github.com/gfx-rs/wgpu/issues/3514 first

wgpu/Cargo.toml
Wumpf: actually the entire block doesn't make sense anymore, no? wgc without a backend feature isn't meaningful  to begin with

wgpu/src/backend/dummy.rs
```diff
@@ -0,0 +1,1519 @@
+use std::future::Pending;
```
Wumpf, line 1: that's part of your experiment of having a dummy package in between I suppose?

Wumpf:
Hmm maybe we could leverage https://docs.rs/cfg_aliases/latest/cfg_aliases/ to detangle things a bit further. I wouldn't mind having something like `gles-linux`, `gles-windows`, `gles-mac` (which implies angle iff we're on mac) and then we use config aliases in the code to distill the real `gles` feature.

What do you think?

wgpu/src/backend/mod.rs
```diff
@@ -10,14 +10,41 @@ mod web;
...
+#[cfg(all(
+    any(target_os = "macos", target_os = "ios"),
+    not(any(feature = "metal", feature = "vulkan-portability", feature = "angle"))
+))]
+mod dummy;
+#[cfg(all(
+    any(target_os = "macos", target_os = "ios"),
+    not(any(feature = "metal", feature = "vulkan-portability", feature = "angle"))
+))]
+pub(crate) use dummy::Context;
```
Wumpf: thinking about this some more I don't quite get why we need this. If you're on mac and no backend is enabled we want to fail compilation right away (_ideally_ with a message)? 🤔
A runtime error seems quite annoying, also this adds yet another thing where we need to duplicate the interface, which makes me very sad...

daxpedda: I have made two separate commits for this so we can easily remove it if we want to.

Wgpu is used by a lot of libraries, so failing compilation when no backend is selected will make it harder for libraries to use `default-features = false`, which is what we should encourage, so end-users can select backends they want instead of libraries doing it for them.

daxpedda: Alternatively to just panicking with `unimplemented!`, we could actually return an error that no device/adapter was found. But that seems worse to me, unless we clarify the error to also explain that there might be devices/adapters available but no backend is enabled for them.

Wumpf: not so picky about how a runtime placeholder backend would fail exactly. Just the fact that we have to duplicate the entire surface for a third time makes me unhappy.

> Wgpu is used by a lot of libraries, so failing compilation when no backend is selected will make it harder for libraries to use default-features = false, which is what we should encourage, so end-users can select backends they want instead of libraries doing it for them.

I mean this totally makes sense, but the question is whether there is any usecase at all for not having a valid backend and I doubt there is. Therefore, why should one wait for compilation to finish and running to find out that the build configuration was broken in the first place.

daxpedda: I don't disagree that there is a tradeoff. This is a problem many libraries have faced and I don't know if any popular consensus was reached.

> I mean this totally makes sense, but the question is whether there is any usecase at all for not having a valid backend and I doubt there is. Therefore, why should one wait for compilation to finish and running to find out that the build configuration was broken in the first place.

The only advantage I see is to allow library developers to depend on Wgpu without having to select a backend.
See a recent discussion in https://github.com/rust-windowing/winit/issues/3174 and a comment by @cwfitzgerald https://github.com/rust-windowing/winit/issues/3174#issuecomment-1775626177.

I have to say that initially I was in favor of supporting compilation with `default-features = false`, but now that I think about it a bit more I'm unsure. It seems simple to me for library developers to enable `default` in `dev-dependencies` to be able to work with it. End-users have to select a backend anyway, so getting the compilation error is even helpful in this case.

Right now at least I'm back to removing this, let me know how you want me to proceed.

wgpu/Cargo.toml
daxpedda: Done.

wgpu/Cargo.toml
daxpedda: This block is necessary to enable `wgpu-core` by default on MacOS & iOS. This lets `metal = ["wgc?/metal"]` enable the Metal backend only when `wgc` is active. E.g. when targeting WebGPU on Wasm, `wgpu-core` is not enabled, therefor the `metal` crate feature will not pull in any dependencies.

I find this horribly complicated, but unfortunately Cargo is very limited and these are the kind of workarounds we have to use to get around these issues.

daxpedda:
> Hmm maybe we could leverage https://docs.rs/cfg_aliases/latest/cfg_aliases/ to detangle things a bit further. I wouldn't mind having something like `gles-linux`, `gles-windows`, `gles-mac` (which implies angle iff we're on mac) and then we use config aliases in the code to distill the real `gles` feature.
> 
> What do you think?

This would be nice if we can solve the crate feature issue. Unfortunately, e.g. `gles-linux = ["wgc/gles"]` would still pull in all OpenGL dependencies on MacOS as well.

I described this problem in more detail in https://github.com/gfx-rs/wgpu/issues/3514.

wgpu/Cargo.toml
daxpedda: I tried to add some explanation to this block, let me know if it makes sense to you.
https://github.com/daxpedda/wgpu/blob/999372c872f5657d082b63cdf056b24f20782ba5/wgpu/Cargo.toml#L72-L73

daxpedda:
I believe the CI failure is unrelated, but it's consistent, so I assume it's broken on `trunk` as well if we rerun it there.
See https://github.com/gfx-rs/wgpu/pull/4813 for the same CI failure.

daxpedda:
Rebased after #4828.
Which also fixed the CI.

Wumpf (commented):
we talked about this PR (and the general feature story) on our weekly wgpu maintainers call. Key takeaways:

* reaffirming that we want to go the route of detailed opt-in features
* this PR is as-is an improvement despite unsolved vulkan/gles stuff
* going forward we want to rely on cfg_alias to reduce the complicated cfg expressions in the code (this also affects other stuff like the notoriously length cfg checks on send/sync on wasm)
* if possible we don't want the dummy.rs way of not having a backend, but the rationale makes sense so if we don't find another way let's go for it
    * maybe it's possible to fail instance creation before even instantiating any instance since everything already goes through `DynContext` 🤔 . I want to experiment with it before we commit to `dummy.rs`
* we don't have a great out-of-the-box solution for the vulkan/gles problematic; handing more things down to wgc is definitely on the table

-> Action points for this PR before landing:
* experiment with default-feature builds that don't rely on dummy.rs @Wumpf 
* changelog update

-> Next steps after this PR
* iterate on GLES/Vulkan issue
* start cfg_alias'ing things to make cfg configs nicer

wgpu/src/backend/mod.rs
daxpedda: Fixed in https://github.com/gfx-rs/wgpu/pull/4815/commits/8a6ca6de06867760c7bdf1368a1831b64586b703.

daxpedda:
> * if possible we don't want the dummy.rs way of not having a backend, but the rationale makes sense so if we don't find another way let's go for it
> 
>   * maybe it's possible to fail instance creation before even instantiating any instance since everything already goes through `DynContext` 🤔 . I want to experiment with it before we commit to `dummy.rs`

So I did remove the `Dummy` backend, instead I'm relying on `hal::api::Empty` to let everything compile. I believe right now it's not going to error on `Instance::new()`, but I'm not entirely sure where to implement returning the error.
1. `Instance::new()`
2. `direct::Context::init()`
3. `wgc::Global::new()`
4. `wgc::Instance::new()`

Let me know how you want me to proceed.

daxpedda:
> -> Action points for this PR before landing:
> 
> * experiment with default-feature builds that don't rely on dummy.rs @Wumpf
> 
> * changelog update

I covered both points now.
Whats still missing is figuring out where exactly we want to error out when no backend is available.

Otherwise the PR should be ready!

Wumpf:
Ah nice, the empty feature makes actually a lot of sense, didn't think of that. What I actually meant is something along those lines
https://github.com/Wumpf/wgpu/commit/9c7f77068

Haven't tried it out yet, but this also gives a suggestion of when to panic in the stack: Right on wgpu::Instance. I.e. compared to the dummy solution we fail before the first call into the context. Also, I sketched this out to be information that's easily accessible to the user at compile time.
What do you think of that direction? (also my attempt to doc status quo in there surely has some mistakes?)

daxpedda:
Panicking at `wgpu::Instance` looks fine to me. I don't think `Instance::any_backend_feature_enabled()` will actually be used by anybody? But it's a nice place to document all that, alternatively we can make the function private and put the documentation on `Instance` itself.

The whole `hal::api::Empty` is still necessary to make things compile, unless we want to put guards in all the functions in `wgpu` that use `gfx_select!`. E.g. your current solution in https://github.com/Wumpf/wgpu/commit/9c7f77068 won't compile with `cargo build --target x86_64-apple-darwin -p wgpu --no-default-features`.

> (also my attempt to doc status quo in there surely has some mistakes?)

I left a comment here: https://github.com/Wumpf/wgpu/commit/9c7f77068#r135074861.
Except the crate features names it's great!

---

Let me know how you want to proceed, I'm happy to pull in your commit here.

EDIT: I made a commit with what I'm proposing. Happy to change it of course.

Wumpf:
ah, thanks for fixing up the feature names and docs! Needing the `hal::api::Empty` makes a lot of sense, by basing off that version I meant to imply that it's included 😄 

I'd really like `any_backend_feature_enabled` to be `pub` - I already can see that some library write wants to run all their tests in `no-default-features` and then wants a convenient runtime check to opt out of creating a wgpu instance, so I think it  would be really useful to expose it. Unless you see some good reason better not to?

But I think that's all I have left to note. I'd would do a final pass over it tomorrow (or maybe later today) and then finally merge it in :)

daxpedda:
> Needing the `hal::api::Empty` makes a lot of sense, by basing off that version I meant to imply that it's included 😄

Ah, my bad! I didn't even check :sweat_smile: 

> I'd really like `any_backend_feature_enabled` to be `pub` - I already can see that some library write wants to run all their tests in `no-default-features` and then wants a convenient runtime check to opt out of creating a wgpu instance, so I think it would be really useful to expose it. Unless you see some good reason better not to?

No, that sounds absolutely fine to me! I've done that now.

Wumpf (approved):
Confirming that all looks good now. Great step forward, thanks @daxpedda for being patient and going through these iterations!

PR #4816, opened 2023-12-02
[naga] Fix crash when range is invalid

gents83:
Often when trying to validate a "wrongly written" wsgl I'm getting a crash.
This allow to output the diagnostic message without crashing


teoxoy:
Just landed https://github.com/gfx-rs/wgpu/pull/4808 which addressed the issue.

Issue #4323, opened 2021-04-08
Combined image samplers


kvark:
This is totally out of scope for WebGPU, but it will be useful for gfx-rs.
For example, gzDoom uses combined image samplers, and we want to run it in gfx-portability without spirv-cross.

PR #4819, opened 2023-12-04
[naga/spv-in] Allow desugaring combined image samplers into split image/samplers at the same bind point

### LINKED ISSUE #4342: [glsl-in] Consider supporting combined image/samplers
My use case for Naga is converting from GLSL ES to WGSL at runtime. GLSL ES only supports combined image/samplers, and it would be quite annoying for me to change my code to do something else. Many typical approaches to this problem are limited by the restrictions of GLSL.

For instance, my current policy is to have texture bindings and samplers take consecutive bindings.

I would be fine if the GLSL is parsed into an IR that can't be output by the WGSL backend, and I would have to convert it to the proper policy myself.

chyyran:
**Connections**
* https://github.com/gfx-rs/wgpu/issues/4342
* https://github.com/gfx-rs/wgpu/issues/4323

**Description**
This adds an option `CombinedImageSamplerDesugaring` to configure the behaviour when spv-in encounters a combined image sampler. 

`CombinedImageSamplerDesugaring::None` preserves the existing behaviour--at the time of this PR's filing, Naga fails on this case due to the `OpSampledImage` instruction not being in the sampled image lookup, but in the future there could be potential IR changes as described in #4342.

`CombinedImageSamplerDesugaring::SplitIntoOverlappedBinding` presents a solution to combined image samplers in spv-in without needing IR changes, and with the potential to convert shaders that use combined image samplers into WGSL-valid Naga IR. When an `OpLoad` of a combined image sampler is encountered, we effectively emit a sampler global variable at the same bind point as the image, and the image sample instruction is rewritten to use the emitted sampler. The resultant IR will not validate until the caller re-assigns the bindings such that they do not overlap. For binding arrays of combined image samplers, an equivalent array of samplers is created, and the image sample instruction is rewritten to use the sampler at the same index of the image.

This does not completely close #4342 but provides a useful backstop: this solution is only implemented for `spv-in` but if something else is used to convert GLSL with combined image samplers to SPIR-V, the resultant SPIR-V will be parseable in Naga, which after manual tweaking of bind points can be used to output valid WGSL.

**Testing**
SPIR-V assemblies with a combined image sampler, and a binding array of combined image sampler, was added to `snapshots.rs` with relevant tests to the behaviour of `CombinedImageSamplerDesugaring::SplitIntoOverlappedBinding`. These tests demonstrate validation after bind points have been adjusted.

**Checklist**

- [x] Run `cargo fmt`.


jimblandy:
In general, I would like for Naga to be able to at least read GLSL ES texture-combined sampler types. WGSL has no support for these, but our other frontends and backends do, and it would be good for Naga to be able to serve those use cases as well.

Unfortunately, I think we can't merge this PR. The functionality it provides is too incomplete, and the IR produced doesn't actually model the input in a way that allows one to build reliable consumers. Taking the PR in its current state will make the SPIR-V front end harder to work with: rather than supporting GLES or Vulkan shaders per spec, it introduces "works for my use case" semantics that are not really maintainable in the long run.

My understanding from our discussion on Matrix is that this PR does not fully support your use case either: you want to accept arbitrary user-provided shader modules, and this PR will still struggle with code your users will expect to work.

In that light, I would encourage you to maintain this work on your own branch, merging from our trunk periodically, until you're able to fully meet your users' expectations. My guess is that, by the time you actually have everything working that your users expect, you will have necessarily evolved this PR in a direction that addresses many of my concerns with the approach.

PR #4821, opened 2023-12-04
[naga/spv-in]: parse combined image samplers as ImageSample expression where image and sampler point to the same handle

chyyran:
**Connections**
* https://github.com/gfx-rs/wgpu/issues/4342
* https://github.com/gfx-rs/wgpu/issues/4323

**Description**
This is a more conservative alternative to https://github.com/gfx-rs/wgpu/pull/4819, and only allows spv-in to "parse" combined image samplers by representing them as `ImageSample` expressions where the `image` and `sampler` are the same. The resulting IR will not validate until the user manually separates the sampler.

**Testing**
todo

**Checklist**

- [ ] Run `cargo fmt`.


naga/src/front/spv/image.rs
```diff
@@ -526,7 +526,20 @@ impl<I: Iterator<Item = u32>> super::Frontend<I> {
             image_ops ^= bit;
         }
 
-        let si_lexp = self.lookup_sampled_image.lookup(sampled_image_id)?;
+        let si_lexp = self.lookup_sampled_image.lookup(sampled_image_id);
+        let si_lexp = match si_lexp {
```
jimblandy: I think you could just say

```
let si_lexp = match self.lookup_sampled_image.get(sampled_image_id) {
    Some(lookup) => ...
    None => ...
}
```

here and skip the whole `InvalidId` error because it's irrelevant

jimblandy:
I don't think we can take this PR, for reasons similar to those explained in #4819.

PR #4826, opened 2023-12-04
validation: More detailed on incompatible BGL

scoopr:
When encountering a incompatible BindGroupLayout on RenderPipelines,
attempt to report the actual difference causing the incompatibility.

**Description**
When encountering a validation error of incompatible layouts on a renderpipeline, the validation error wasn't too helpful.

```
Caused by:
    In a RenderPass
      note: encoder = `<CommandBuffer-(0, 1, Metal)>`
      note: render pipeline = `shadow`
    The pipeline layout, associated with the current render pipeline, contains a bind group layout at index 1 which is incompatible with the bind group layout associated with the bind group at 1
```
I shortened the final error message a little, as it was confusingly wordy, and then added info on the differences:
```
Caused by:
    In a RenderPass
      note: encoder = `<CommandBuffer-(0, 1, Metal)>`
      note: render pipeline = `shadow`
    The pipeline layout, associated with the current render pipeline, contains a incompatible bind group layout at index 1
      note: Entry 1 not found in assigned bindgroup layout
```

At this stage, I'm still asking opinions on how the overall approach, and implementation.

Still things to do:
- [ ] Go through the wording a bit more
- [ ] Show the labels of the BGLs
- [ ] Implement the same for compute pipelines
- [ ] .. and the checklist stuff

**Testing**
Manually munging the examples.

**Checklist**

- [ ] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/command/bind.rs
+++ b/wgpu-core/src/command/bind.rs
@@ -59,24 +59,33 @@ mod compat {
 
             if let Some(expected_bgl) = self.expected.as_ref() {
                 if let Some(assigned_bgl) = self.assigned.as_ref() {
-
                     for (id, e_entry) in &expected_bgl.entries {
                         if let Some(a_entry) = assigned_bgl.entries.get(id) {
-
                             if a_entry.binding != e_entry.binding {
-                                    diff.push(format!("Entry {id} binding expected {}, got {}", e_entry.binding, a_entry.binding));
-                                }
-                                if a_entry.count != e_entry.count {
-                                    diff.push(format!("Entry {id} count expected {:?}, got {:?}", e_entry.count, a_entry.count));
-                                }
-                                if a_entry.ty != e_entry.ty {
-                                    diff.push(format!("Entry {id} type expected {:?}, got {:?}", e_entry.ty, a_entry.ty));
-                                }
-                                if a_entry.visibility != e_entry.visibility {
-                                    diff.push(format!("Entry {id} visibility expected {:?}, got {:?}", e_entry.visibility, a_entry.visibility));
-                                }
-
-                            } else {
+                                diff.push(format!(
+                                    "Entry {id} binding expected {}, got {}",
+                                    e_entry.binding, a_entry.binding
+                                ));
+                            }
+                            if a_entry.count != e_entry.count {
+                                diff.push(format!(
+                                    "Entry {id} count expected {:?}, got {:?}",
+                                    e_entry.count, a_entry.count
+                                ));
+                            }
+                            if a_entry.ty != e_entry.ty {
+                                diff.push(format!(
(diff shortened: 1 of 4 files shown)


wgpu-core/src/command/bind.rs
```diff
@@ -53,6 +53,50 @@ mod compat {
         fn is_incompatible(&self) -> bool {
             self.expected.is_none() || !self.is_valid()
         }
+
+        fn bgl_diff(&self) -> Vec<String> {
+            let mut diff = vec![];
```
cwfitzgerald:
```suggestion
            let mut diff = Vec::new();
```

Just a style thing

wgpu-core/src/command/bind.rs
```diff
@@ -121,6 +165,16 @@ mod compat {
...
         }
+
+        pub fn bgl_diff(&self) -> Vec<String> {
+            for e in &self.entries {
+                if !e.is_valid() {
+                
+                    return e.bgl_diff()
+                }                   
+            }
+            vec![format!("huh?!")]
```
cwfitzgerald:
```suggestion
            vec![String::from("huh?!")]
```

cwfitzgerald (changes requested):
Some small things, then CI, then good

scoopr:
Alright, I addressed the nits, tweaked the error messages, and applied the same for compute side. I think this is now in a bit better shape.

Example of with compute looks like:
```
Caused by:
    In a ComputePass
      note: encoder = `<CommandBuffer-(0, 1, Metal)>`
    In a dispatch command, indirect:false
      note: compute pipeline = `Compute pipeline`
    Incompatible bind group at index 0 in the current compute pipeline
      note: Should be compatible with bind group layout with label = `compute bgl2`
      note: Assigned bind group layout with label = `compute bgl`
      note: Entry 3 not found in assigned bindgroup layout
```

Few notes,
* I couldn't use the same labeling infra that the other errors use as I couldn't get the bgl ids nicely, but probably for the best, as I think the added context in the error messages is useful, though the lines are a bit long, and not sure if equals sign is helpful there.
* As it happens, this type of error message is presented very rarely, as lot of the other validation checks trigger before getting this far, so depending on the case, it might be nice to ease the validation on another layer to make it trigger here, but then again earlier check might have more useful context.
* And my main concern is actually that the validation check itself is different code than the validation error generation, which might cause some deviation. Its also pretty mechanical code, and it almost triggered me to go learn proc-macros to generate the diff and a fast-path validation "easily" from the same source. But that seemed a lot of effort for just this one problem :)

cwfitzgerald (approved):
Amazing!

cwfitzgerald:
I think we should give this a spin, then iterate if there are issues.

PR #4834, opened 2023-12-05
Eagerly release GPU resources when we lose the device.

bradwerth:
**Connections**
N/A

**Description**
This destroys buffer objects whenever we "lose the device", including at the end of the `device.destroy` process. Destruction of the buffers triggers reclamation of GPU memory, rather than waiting for the garbage collector to drop the buffer. It would be ideal to also destroy texture objects, but `texture.destroy` is not implemented, as noted in a TODO.

**Testing**
This is exercised by any bulk test execution, like the CTS, which relies on `device.destroy` (not waiting for the garbage collector) to free up GPU resources. Bulk test execution will have less failures due to full memory conditions.

**Checklist**

- [X] Run `cargo fmt`.


wgpu-core/src/device/resource.rs
```diff
@@ -3317,7 +3321,8 @@ impl<A: HalApi> Device<A> {
         self.valid.store(false, Ordering::Release);
 
         // 1) Resolve the GPUDevice device.lost promise.
-        let closure = self.lock_life().device_lost_closure.take();
+        let mut life_lock = self.lock_life();
```
cwfitzgerald, line 3346: We shouldn't be holding the life lock across external closure invocations - this could potentially deadlock.

cwfitzgerald (changes requested):

bradwerth:
Ugh, I botched the rebase/merge and I'm not sure how to recover. I'll resubmit as a new PR.

bradwerth:
New, non-borked PR is https://github.com/gfx-rs/wgpu/pull/4851.

Issue #4825, opened 2023-12-04
NV12 Vulkan CI Errors


cwfitzgerald:
CI is now running into after a mesa update:

```
Validation Error: [ VUID-VkImageCreateInfo-pNext-06722 ] | MessageID = 0x1a04b807 | vkCreateImage(): pCreateInfo->pNext<VkImageFormatListCreateInfo>.pViewFormats[0] (VK_FORMAT_R8_UNORM) and VkImageCreateInfo::format (VK_FORMAT_G8_B8R8_2PLANE_420_UNORM) are not compatible. The Vulkan spec states: If a VkImageFormatListCreateInfo structure was included in the pNext chain and VkImageFormatListCreateInfo::viewFormatCount is not zero, then each format in VkImageFormatListCreateInfo::pViewFormats must either be compatible with the format as described in the compatibility table or, if flags contains VK_IMAGE_CREATE_BLOCK_TEXEL_VIEW_COMPATIBLE_BIT, be an uncompressed format that is size-compatible with format 
```

Looking at [this table](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap49.html#formats-compatibility-classes), I don't think it's valid to pass the plane formats as `view_formats`. The plane functionality is similar to the existing aspect functionality used for combined depth-stencil textures where the format specified for the view is the format of one of the planes/aspects.

I think it would have made sense to add variants for the planes to `TextureAspect` rather than the new `plane` param.

_Originally posted by @teoxoy in https://github.com/gfx-rs/wgpu/issues/4573#issuecomment-1838575468_

PR #4835, opened 2023-12-05
Device drop lost

bradwerth:
**Connections**
N/A

**Description**
A User agent may reasonably assume that any supplied DeviceLostCallbacks is invoked eventually on a device. For example, the user agent may hold resources waiting for a callback, and if it is never invoked, the user agent will leak that memory. Triggering "lose the device" during device.drop ensures that the callback is called in all cases.

**Testing**
Added new test DEVICE_DROP_THEN_LOST.

**Checklist**

- [X] Run `cargo fmt`.


ErichDonGubler:
NV12 failures are just https://github.com/gfx-rs/wgpu/issues/4825, but the [`SEGV` for the `multithreaded_compute`](https://github.com/gfx-rs/wgpu/actions/runs/7107705627/job/19349658909?pr=4835#step:12:6891) seems to be a regression in these changes.

cwfitzgerald (changes requested):
I don't think this is correct as it stands.

It's perfectly valid, say, to let go of your handle to the device before you're done using it. There's not a ton you can do, but you can still submit to the queue, map buffers, etc all after the user handle to the user device handle has died.

In general, I'm not sure we actually need to call the device loss callback on any kind of device drop (we need to call the C callback so the C callback can clean itself up, but in rust land you don't need that.

bradwerth:
> I don't think this is correct as it stands.
> 
> It's perfectly valid, say, to let go of your handle to the device before you're done using it. There's not a ton you can do, but you can still submit to the queue, map buffers, etc all after the user handle to the user device handle has died.

Good point. I guess this is on embedders to ensure that they don't rely on the callback being called.

jimblandy:
Firefox does need to know when the callback gets dropped without being used. Our C callback owns data that needs to be freed. In discussion this morning, we thought it might serve if `DeviceLostClosure` had a `Drop` implementation that checked if `inner` was a `DeviceLostClosure::C`, and then invoked the callback passing `DeviceLostReason::Destroyed`.

The Rust callbacks can just have the closure own values that implement `Drop`, so they don't need any special handling. Although it might be more consistent to invoke the callback in that case as well.

bradwerth:
I've revived this approach in https://github.com/gfx-rs/wgpu/pull/4862, more narrowly-tailored to only affect the behavior of `DeviceLostClosureC`.

Issue #3762, opened 2023-05-09
Support `POLYGON_MODE_LINE` and `POLYGON_MODE_POINT` for GL


LucaCiucci:
**Is your feature request related to a problem? Please describe.**
I'm trying to use wgpu in a tauri application (compiling for `wasm32`), I need to draw some lines and points so I tried to request [`Features::POLYGON_MODE_LINE`](https://github.com/gfx-rs/wgpu/blob/f173575427b028dde71bdb76dce10d27060b03ba/wgpu-types/src/lib.rs#L651) and [`Features::POLYGON_MODE_POINT`](https://github.com/gfx-rs/wgpu/blob/f173575427b028dde71bdb76dce10d27060b03ba/wgpu-types/src/lib.rs#L651) using a GL backend, but the [`Adapter::request_device`](https://github.com/gfx-rs/wgpu/blob/f173575427b028dde71bdb76dce10d27060b03ba/wgpu/src/lib.rs#L1715) function returns an error if these features are requested. 

**Describe the solution you'd like**
Support these two polygon modes for the GL backend.

**Describe alternatives you've considered**
Use `PrimitiveTopology::LineList` or `PrimitiveTopology::PointList`, but `POLYGON_MODE_*` could simplify the usage in some cases, for example when switching the representation of a mesh to wireframe or points.

**Additional context**
None.

cwfitzgerald:
`glPolygonMode` seems to be completely unsupported on GLES outside of `GL_NV_polygon_mode` which has basically zero support, so this is unfortunately not really feasible to support

PR #4836, opened 2023-12-06
[gl] add support for line and point polygon modes

valaphee:
**Connections**
https://github.com/gfx-rs/wgpu/issues/3762

**Description**
Since Desktop GL is now supported, polygon modes can now be implemented for the GL backend.

Also made the full ver/es ver parsing more readable. (is_none then_some and_then -> map_or_else). 

There is no explicit feature check in the map_primitive_state but this is fine as it seems to be already checked before and polygon_mode is always called as glow ignores it if its fill on web.

**Testing**
Tested by running cube, also with different culling options, to see if the behavior is identical to Vulkan, and running the WebGL demos.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-hal/src/gles/queue.rs
+++ b/wgpu-hal/src/gles/queue.rs
@@ -1330,7 +1330,10 @@ impl super::Queue {
                         unsafe { gl.disable(glow::DEPTH_CLAMP) };
                     }
                 }
-                unsafe { gl.polygon_mode(glow::FRONT_AND_BACK, state.polygon_mode) };
+                // POLYGON_MODE_LINE also implies POLYGON_MODE_POINT
+                if self.features.contains(wgt::Features::POLYGON_MODE_LINE) {
+                    unsafe { gl.polygon_mode(glow::FRONT_AND_BACK, state.polygon_mode) };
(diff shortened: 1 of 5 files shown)


nical (approved):

teoxoy:
It looks like CI is failing on GLES.

wgpu-hal/src/gles/queue.rs
```diff
@@ -1330,6 +1330,7 @@ impl super::Queue {
                         unsafe { gl.disable(glow::DEPTH_CLAMP) };
                     }
                 }
+                unsafe { gl.polygon_mode(glow::FRONT_AND_BACK, state.polygon_mode) };
```
Gordon-F: I think these should be:
```rust
if (!gl.version().is_embedded) {
    unsafe { gl.polygon_mode(glow::FRONT_AND_BACK, state.polygon_mode) };
}
```

valaphee: Yea didn't thought of it, as for WebGL it was just ignored, but calling it on GLES logically results in an invalid op

PR #4837, opened 2023-12-06
Update multi-planar texture API

teoxoy:
**Connections**
https://github.com/gfx-rs/wgpu/issues/4825 & https://github.com/gfx-rs/wgpu/pull/4823#issuecomment-1840612949

**Description**
- remove the need to pass the plane formats to `view_formats`
- move plane selection in `TextureAspect`

cc @xiaopengli89

### DIFF
--- a/wgpu-hal/src/dx12/device.rs
+++ b/wgpu-hal/src/dx12/device.rs
@@ -467,11 +467,7 @@ impl crate::Device<super::Api> for super::Device {
             aspects: view_desc.aspects,
             target_base: (
                 texture.resource.clone(),
-                texture.calc_subresource(
-                    desc.range.base_mip_level,
-                    desc.range.base_array_layer,
-                    desc.plane.unwrap_or(0),
-                ),
+                texture.calc_subresource(desc.range.base_mip_level, desc.range.base_array_layer, 0),
             ),
             handle_srv: if desc.usage.intersects(crate::TextureUses::RESOURCE) {
                 let raw_desc = unsafe { view_desc.to_srv() };
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -2411,9 +2411,9 @@ pub enum TextureFormat {
     /// - 0: Single 8 bit channel luminance.
     /// - 1: Dual 8 bit channel chrominance at half width and half height.
     ///
-    /// Valid view formats for luminance are [`TextureFormat::R8Unorm`] and [`TextureFormat::R8Uint`].
+    /// Valid view formats for luminance are [`TextureFormat::R8Unorm`].
     ///
-    /// Valid view formats for chrominance are [`TextureFormat::Rg8Unorm`] and [`TextureFormat::Rg8Uint`].
+    /// Valid view formats for chrominance are [`TextureFormat::Rg8Unorm`].
     ///
     /// Width and height must be even.
     ///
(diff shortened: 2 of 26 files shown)


crowlKats (commented):
no comment

nical (approved):

wgpu-types/src/lib.rs
```diff
@@ -2850,7 +2862,16 @@ impl TextureFormat {
             ) => Some(Self::Stencil8),
             (Self::Depth24PlusStencil8, TextureAspect::DepthOnly) => Some(Self::Depth24Plus),
             (Self::Depth32FloatStencil8, TextureAspect::DepthOnly) => Some(Self::Depth32Float),
-            (format, TextureAspect::All) => Some(format),
+            (Self::NV12, TextureAspect::Plane0) => Some(Self::R8Unorm),
+            (Self::NV12, TextureAspect::Plane1) => Some(Self::Rg8Unorm),
+            (format, TextureAspect::All) => {
```
```suggestion
            (format, TextureAspect::All) if !format.is_multi_planar_format() => {
```

Then you can just fallthrough to the None

wgpu-types/src/lib.rs
```diff
@@ -2920,6 +2949,14 @@ impl TextureFormat {
         }
     }
 
+    /// Returns the size multiple requirement for a texture using this format.
+    pub fn size_multiple_requirement(&self) -> (u32, u32) {
+        match *self {
+            Self::NV12 => (2, 2),
+            _ => (1, 1),
```
cwfitzgerald: I think this should be the block size for non-nv12. A user might rely on this value without knowing about the whole block thing.

wgpu-types/src/lib.rs
```diff
@@ -3524,11 +3572,15 @@ impl TextureFormat {
...
             Self::Depth24PlusStencil8 | Self::Depth32FloatStencil8 => match aspect {
-                TextureAspect::All => 2,
                 TextureAspect::DepthOnly | TextureAspect::StencilOnly => 1,
+                _ => 2,
             },
 
-            Self::NV12 => 3,
+            Self::NV12 => match aspect {
+                TextureAspect::Plane0 => 1,
+                TextureAspect::Plane2 => 2,
```
cwfitzgerald: Typo? NV12 only has two planes?

cwfitzgerald (changes requested):
Some small stuff, but looks much better

wgpu-types/src/lib.rs
teoxoy: Good catch!

wgpu-hal/src/dx12/device.rs
```diff
@@ -467,11 +467,7 @@ impl crate::Device<super::Api> for super::Device {
             aspects: view_desc.aspects,
             target_base: (
                 texture.resource.clone(),
-                texture.calc_subresource(
-                    desc.range.base_mip_level,
-                    desc.range.base_array_layer,
-                    desc.plane.unwrap_or(0),
-                ),
+                texture.calc_subresource(desc.range.base_mip_level, desc.range.base_array_layer, 0),
```
xiaopengli89, line 470: The value of plane should be obtained from aspect.

teoxoy: `target_base` is only used for resolve (I added a comment on it) which as far as I could find doesn't work with multi-planar nor combined depth-stencil textures.

xiaopengli89: Yeah, since we can now get the plane from the aspect, I think it is better to use the real plane to calculate the subresource.

teoxoy: I meant that in this case the plane is irrelevant since you can never resolve into a multi-planar or combined depth-stencil texture.

cwfitzgerald (approved):

PR #4849, opened 2023-12-08
[wgpu] correct doc for resolve query set

### LINKED ISSUE #4838: Incorrect docs for resolve_query_set
I was reading about occlusion culling and looking at the `wgpu` documentation for relevant features, and noticed something odd in  [the docs for `CommandEncoder::resolve_query_set()` and its `impl`](https://docs.rs/wgpu/0.18.0/wgpu/struct.CommandEncoder.html#impl-CommandEncoder-2) ([source](https://docs.rs/wgpu/0.18.0/src/wgpu/lib.rs.html#3604-3628)).

* The function documentation says "See `PipelineStatisticsTypes` for more information." Surely this should be `QueryType`?
* The `impl` block documentation says “`Features::TIMESTAMP_QUERY` or `Features::PIPELINE_STATISTICS_QUERY` must be enabled on the device in order to call these functions.” But occlusion queries are neither of those, and there's only one function in the `impl` block.

I suspect these are both left over from a previous WebGPU API revision, but I have no experience with queries so I can't be confident that I'm right or what the correct statement is. Happy to send a PR if the intent can be clarified.

valaphee:
**Connections**
Closes https://github.com/gfx-rs/wgpu/issues/4838

**Description**
Moved `resolve_query_set` to the general command encoder impl, as it technically doesn't require any feature, is not meaningful to the end user, and it's not possible to create a query set with an unsupported type anyway.

And added info about how many bytes to expect.

**Testing**

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -3674,7 +3674,7 @@ impl CommandEncoder {
 
     /// Resolves a query set, writing the results into the supplied destination buffer.
     ///
-    /// Occlusion and timestamp queries are 8 bytes each (see [`wgpu::QUERY_SIZE`]). For pipeline statistics queries,
+    /// Occlusion and timestamp queries are 8 bytes each (see [`crate::QUERY_SIZE`]). For pipeline statistics queries,
     /// see [`PipelineStatisticsTypes`] for more information.
     pub fn resolve_query_set(
         &mut self,
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -3671,6 +3671,31 @@ impl CommandEncoder {
         let id = self.id.as_ref().unwrap();
         DynContext::command_encoder_pop_debug_group(&*self.context, id, self.data.as_ref());
     }
+
+    /// Resolves a query set, writing the results into the supplied destination buffer.
+    ///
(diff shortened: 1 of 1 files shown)


wgpu/src/lib.rs
```diff
@@ -3671,34 +3671,11 @@ impl CommandEncoder {
...
-    }
-}
 
-/// [`Features::TIMESTAMP_QUERY`] or [`Features::PIPELINE_STATISTICS_QUERY`] must be enabled on the device in order to call these functions.
-impl CommandEncoder {
-    /// Resolve a query set, writing the results into the supplied destination buffer.
+    /// Resolves a query set, writing the results into the supplied destination buffer.
     ///
-    /// Queries may be between 8 and 40 bytes each. See [`PipelineStatisticsTypes`] for more information.
+    /// Occlusion and timestamp queries are 8 bytes each. For pipeline statistics queries,
```
```suggestion
    /// Occlusion and timestamp queries are 8 bytes each (see [`wgpu::QUERY_SIZE`]). For pipeline statistics queries,
```

Wumpf (approved):
nice, thank you!

cwfitzgerald:
CI is failing

Wumpf:
oops, my fault!

PR #4850, opened 2023-12-08
[naga wgsl-in] Support abstract operands to binary operators.

jimblandy:
Still some missing bits and some invalid output, but roughly working.

Note that commits are split out, only the last one should be interesting.

**Checklist**

- [X] Run `cargo fmt`.
- [X] Run `cargo clippy`.


teoxoy (approved):
Looks good but I think it needs to be rebased since I landed https://github.com/gfx-rs/wgpu/pull/4869.

PR #4851, opened 2023-12-08
Eagerly release GPU resources when we lose the device.

bradwerth:
**Connections**
N/A

**Description**
This destroys buffer objects whenever we "lose the device", including at the end of the device.destroy process. Destruction of the buffers triggers reclamation of GPU memory, rather than waiting for the garbage collector to drop the buffer. It would be ideal to also destroy texture objects, but texture.destroy is not implemented, as noted in a TODO.

**Testing**
This is exercised by any bulk test execution, like the CTS, which relies on device.destroy (not waiting for the garbage collector) to free up GPU resources. Bulk test execution will have less failures due to full memory conditions.

**Checklist**

- [X] Run `cargo fmt`.


bradwerth:
This is a resubmit of https://github.com/gfx-rs/wgpu/pull/4834, fixing the botched merge there, and addressing the issues that had been raised so far on that PR.

cwfitzgerald (approved):
Thanks! Definitely an improvement!

PR #4853, opened 2023-12-09
Format extent and origin types as tuples.

nical:
**Description**

Makes logging a few common types a fair bit less verbose.

```rust
println!("origin: {origin:?}, size: {size:?}");
```
goes from:

```
origin Origin2D { x: 0, y: 0 }, size: Extent3D { width: 128, height: 128, depth_or_array_layers: 1 }
```

to:

```
origin: (0, 0), size: (128, 128, 1)
```

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -5274,7 +5274,7 @@ pub enum TextureDimension {
 /// Corresponds to [WebGPU `GPUOrigin2D`](
 /// https://gpuweb.github.io/gpuweb/#dictdef-gpuorigin2ddict).
 #[repr(C)]
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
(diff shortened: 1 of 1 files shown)


wgpu-types/src/lib.rs
```diff
@@ -5336,12 +5342,18 @@ impl Default for Origin3d {
     }
 }
 
+impl std::fmt::Debug for Origin3d {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        (self.x, self.y, self.x).fmt(f)
```
Wumpf:
```suggestion
        (self.x, self.y, self.z).fmt(f)
```

Wumpf (changes requested):
love it! But you got a pretty consequential typo there!

wgpu-types/src/lib.rs
nical: woops!

Wumpf (approved):

PR #4861, opened 2023-12-11
[naga] Teach the constant evaluator vector/vector operators.

jimblandy:
Allow constant evaluation of binary operators whose left and right
operands are both vectors.

**Checklist**

- [X] Run `cargo fmt`.
- [X] Run `cargo clippy`.


teoxoy (approved):
Nice!

Issue #4787, opened 2023-11-27
Dependencies between resources prevent destroy from doing its job


nical:
We currently don't deallocate resources as long as their reference count (as in the counter in the `Arc` itself) does not get down to a certain number which is semantically equivalent to zero ([See the checks](https://github.com/gfx-rs/wgpu/blob/238c3732a42ca1b5eb20ca3afcb8c8f4078ee620/wgpu-core/src/track/stateless.rs#L104)).

Dependencies between resources, for example a texture view holding on to a texture, contribute a strong reference count, however we need `texture.destroy` to be able to deallocate textures even while it is referenced by texture views. This is important because the role of the `destroy` methods is to ensure resources can predictably be deallocated even in garbage collected environments (such the web).

nical:
# TODO

 - [x] Introduce the device-global snatch guards (#4867)
 - [x] Make buffers use snatch guards (#4867)
 - [x] Make textures use snatch guards
   - [ ] make `surface_present` and `surface_texture_discard` not [mutate a texture's raw handle](https://github.com/gfx-rs/wgpu/blame/6dc9ccab8592645fda3204be1cfb5929fd7f924d/wgpu-core/src/present.rs#L343).
 - [x] accessing bind groups should happen with the snatch lock taken. 
 - [x] snatch buffers in `buffer.destroy()`
 - [x] snatch textures in `texture.destroy()`

PR #4867, opened 2023-12-12
Reintroduce buffer snatching Part 1

nical:
**Connections**

This is the first step towards addressing #4787.

**Description**

We need `destroy()` methods to be able to deallocate gpu resources while the associated handle is still alive. This is difficult to do safely because most of wegpu's API can be used from any thread.One of the steps towards our goal is to re-introduce "buffer snatching". The idea is that when, say, `buffer.destroy()` is called, we snatch the raw buffer out of `wgpu_core`'s buffer struct, and enqueue it to be deallocated after the latest submission the buffer has been used in is complete.
Snatching the raw handle out requires some form of locking. @cwfitzgerald's proposed solution is to protect all raw buffer and texture reads behind the read guard of a device-global `RwLock`, and only ever thake a write lock during snatching. The hope is that the overhead of taking a single read guard in most places will be small enough and that `destroy` will be called concurrently with other wgpu work rarely enough that it won't cause very bad contention.

This PR implements the very first step of Connor's idea:
 - `Device` holds `snatchable_lock`.
 - a `Snatchable<T>` type is introduced. Its content cannot be accessed without an appropriate lock guard.
 - `Buffer.raw` is now a ``Snatchable<A::Buffer>` and all associated reads need the guard.

The snatching is not actually performed in this PR (yet). It will come in a followup.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -34,7 +34,7 @@ use crate::{
 
 use arrayvec::ArrayVec;
 use hal::{CommandEncoder as _, Device as _};
-use parking_lot::{Mutex, MutexGuard, RwLock};
+use parking_lot::{Mutex, MutexGuard, RwLock, RwLockReadGuard, RwLockWriteGuard};
 
 use smallvec::SmallVec;
 use thiserror::Error;
@@ -42,6 +42,7 @@ use wgt::{DeviceLostReason, TextureFormat, TextureSampleType, TextureViewDimensi
 
 use std::{
     borrow::Cow,
+    cell::UnsafeCell,
     iter,
     num::NonZeroU32,
     sync::{
@@ -3414,25 +3415,30 @@ impl<A: HalApi> Resource<DeviceId> for Device<A> {
 }
 
 pub struct SnatchableMarker;
-use parking_lot::RwLockReadGuard;
-
 pub type SnatchGuard<'a> = RwLockReadGuard<'a, SnatchableMarker>;
+pub type ExclusiveSnatchGuard<'a> = RwLockWriteGuard<'a, SnatchableMarker>;
 
 pub struct Snatchable<T> {
-    value: Option<T>,
+    value: UnsafeCell<Option<T>>,
 }
 
 impl<T> Snatchable<T> {
     pub fn new(val: T) -> Self {
-        Snatchable { value: Some(val) }
+        Snatchable {
+            value: UnsafeCell::new(Some(val)),
+        }
     }
 
     pub fn get(&self, _guard: &SnatchGuard) -> Option<&T> {
-        self.value.as_ref()
+        unsafe { (*self.value.get()).as_ref() }
+    }
+
+    pub fn snatch(&self, _guard: &ExclusiveSnatchGuard) -> Option<T> {
+        unsafe { (*self.value.get()).take() }
     }
 
     pub fn take(&mut self) -> Option<T> {
-        self.value.take()
+        self.value.get_mut().take()
     }
 }
 
@@ -3443,3 +3449,5 @@ impl<T> std::fmt::Debug for Snatchable<T> {
         write!(f, "<snatchable>")
     }
 }
+
+unsafe impl<T> Sync for Snatchable<T> {}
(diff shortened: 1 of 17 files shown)


nical:
I don't want this marked as draft, it should get reviewed and hopefully landed. The "WIP" in the title was referring to the fact that it's a part of a larger ongoing refactoring, I'll remove it since it is more confusing than helpful.

wgpu-core/src/device/resource.rs
```diff
@@ -3406,3 +3412,34 @@ impl<A: HalApi> Resource<DeviceId> for Device<A> {
...
+pub struct Snatchable<T> {
+    value: Option<T>,
+}
+
+impl<T> Snatchable<T> {
+    pub fn new(val: T) -> Self {
+        Snatchable { value: Some(val) }
+    }
+
+    pub fn get(&self, _guard: &SnatchGuard) -> Option<&T> {
```
nical: Note: To be perfectly transparent, one could argue that even passing the a lock guard, this method is not strictly safe because one could pass the wrong guard. Now we only have a single lock per device and there is no cross-device interaction allowed so one would really need to go out of their way to get this wrong.

I don't want to sprinkle `unsafe` over the entire code base for this but now is the time to argue if you feel otherwise or have a better solution.

cwfitzgerald: I think it's fine to have internal code not be fully up to unsafe standards, especially when the chance of getting it wrong is low. You could argue the majority of wgpu-core is unsafe, but that's would make the unsafe not really useful.

wgpu-core/src/device/resource.rs
```diff
@@ -3406,3 +3413,55 @@ impl<A: HalApi> Resource<DeviceId> for Device<A> {
...
+/// A guard that allows snatching the snatchable data.
+pub type ExclusiveSnatchGuard<'a> = RwLockWriteGuard<'a, SnatchableMarker>;
+
+/// A value that is mostly immutable but can be "snatched" if we need to destroy
+/// it early.
+///
+/// In order to safely access the underlying data, the device's global snatchable
+/// lock must be taken. To guarentee it, methods take a read or write guard of that
+/// special lock.
+pub struct Snatchable<T> {
```
cwfitzgerald: Could we move the implementation of Snatchable (and the various associated types) into it's own module.

wgpu-core/src/device/resource.rs
```diff
@@ -3406,3 +3413,55 @@ impl<A: HalApi> Resource<DeviceId> for Device<A> {
         &mut self.info
     }
 }
+
+pub struct SnatchableMarker;
+/// A guard that provides read access to snatchable data.
+pub type SnatchGuard<'a> = RwLockReadGuard<'a, SnatchableMarker>;
+/// A guard that allows snatching the snatchable data.
+pub type ExclusiveSnatchGuard<'a> = RwLockWriteGuard<'a, SnatchableMarker>;
```
cwfitzgerald: Once we move this into a module - we should make these opaque wrapper types. Specifically `SnatchLock` (name can be changed) being it's own type would be useful.

That way we can mark SnatchLock::new() as unsafe. It would force the user to have to think about if calling new is actually valid, and help prevent accidental shenanigans.

cwfitzgerald (changes requested):
Looks good! Some organizational notes.

ErichDonGubler:
> I don't want this marked as draft, it should get reviewed and hopefully landed. The "WIP" in the title was referring to the fact that it's a part of a larger ongoing refactoring, I'll remove it since it is more confusing than helpful.

Ah, sorry!

wgpu-core/src/snatch.rs
```diff
@@ -0,0 +1,87 @@
...
+unsafe impl<T> Sync for Snatchable<T> {}
+
+
+/// A Device-global lock for all snatchable data.
+pub struct SnatchLock {
+    lock: RwLock<()>,
+}
+
+impl SnatchLock {
+    /// The safety of `Snatchable::get` and `Snatchable::snatch` rely on their using the
```
cwfitzgerald:
```suggestion
    /// The safety of `Snatchable::get` and `Snatchable::snatch` rely on their using of the
```

wgpu-core/src/snatch.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// Take the value without a guard. This can only be used with exclusive access
+    /// to self, so it does not require locking.
+    ///
+    /// Typically useful in a drop implementation.
+    pub fn take(&mut self) -> Option<T> {
+        self.value.get_mut().take()
+    }
+}
+
+// Can't safely print the contents of a snatchable objetc without holding
```
cwfitzgerald:
```suggestion
// Can't safely print the contents of a snatchable object without holding
```

cwfitzgerald (approved):
Nice!

Issue #4400, opened 2022-11-17
[wgsl-in] Implement automatic type conversions (abstract types)


teoxoy:
https://gpuweb.github.io/gpuweb/wgsl/#conversion-rank

jimblandy:
@teoxoy @ErichDonGubler @cwfitzgerald What do you think about just closing this? With #6833, we don't have anything that we're specifically aware of that needs to be fixed. It's not clear how valuable the metabug is any more.

cwfitzgerald:
Sounds good to me

jimblandy:
Okay, closing this, then.

PR #4870, opened 2023-12-12
[naga wgsl] Let unary operators accept and produce abstract types.

### LINKED ISSUE #4435: [wgsl-in] Parse error for subtraction in array index
This shader:
```rust
var<workgroup> a: array<u32, 64>;

fn foo() {
    let x = 1;
    let y = a[x-1];
}
```
Produces this error:

```
error: expected ']', found '-1'
  ┌─ a.wgsl:6:16
  │
6 │     let y = a[x-1];
  │                ^^ expected ']'

Could not parse WGSL
```

### LINKED ISSUE #4445: Subtracting float from product requires parentheses
It looks like using a minus sign in an "additive_expression" wrapped in a "paren_expression" results in a parser error

```wgsl
// This ok:
let a = (3.0*2.0-(1.0)) * 1.0;
let b = (3.0*2.0+1.0) * 1.0;
// This fails:
let c = (3.0*2.0-1.0) * 1.0;
```

```
Parse Error: expected ')', found '-1.0'
```

This confused me for a while, and from what I understand from the [grammar](https://www.w3.org/TR/WGSL/#syntax-additive_expression) it should be allowed.

To reproduce this you can run this fragment shader with my [slimshader](https://github.com/wrnrlr/slimshader) or any such tool:

```wgsl
[[block]]
struct Uniforms {
  resolution: vec2<f32>; // in pixels
  playtime: f32; // in seconds
};

[[group(0), binding(0)]]
var<uniform> uniforms: Uniforms;

struct VertexOutput {
  [[location(0)]] coord: vec2<f32>;
  [[builtin(position)]] position: vec4<f32>;
};

[[stage(fragment)]]
fn main(in: VertexOutput) -> [[location(0)]] vec4<f32> {
  // This ok
  let a = (3.0*2.0-(1.0)) * 1.0;
  let b = (3.0*2.0+1.0) * 1.0;
  // This fails
  let c = (3.0*2.0-1.0) * 1.0;
  return vec4<f32>(1.0, 0.0, 0.0, 1.0);
}
```

### LINKED ISSUE #4492: [wgsl-in] Incorrect handling of negative integers with hex values
The following isn't accepted right now (note that this is how the spec defines the minimum integer https://www.w3.org/TR/WGSL/#integer-types)
```
const i32min = i32(-0x80000000);
```

Instead this is accepted and probably shouldn't:
```
const i32min = -0x80000000i
```
In fact, Tint will fail this with:
```
Tint WGSL reader failure: :24:17 error: value cannot be represented as 'i32'
const i32min = -0x80000000i; // Tint can't handle this being represented as `-0x80000000i`
```
I thought originally that this is the actual error, but I think the reasoning over at the chrome bug I opened is correct and above should be rejected, see https://bugs.chromium.org/p/chromium/issues/detail?id=1439274

For anyone else running into this, note that `const i32min = -2147483648;` is accepted by both compilers.

jimblandy:
This lets us get rid of sign handling in the lexer, which is nice.

Draft because it depends on #4850 and #4869.

Fixes https://github.com/gfx-rs/wgpu/issues/4445.
Fixes https://github.com/gfx-rs/wgpu/issues/4492.
Fixes https://github.com/gfx-rs/wgpu/issues/4435.

**Checklist**

- [X] Run `cargo fmt`.
- [X] Run `cargo clippy`.

### DIFF
--- a/naga/src/proc/constant_evaluator.rs
+++ b/naga/src/proc/constant_evaluator.rs
@@ -1191,8 +1191,10 @@ impl<'a> ConstantEvaluator<'a> {
         let expr = match self.expressions[expr] {
             Expression::Literal(value) => Expression::Literal(match op {
                 UnaryOperator::Negate => match value {
-                    Literal::I32(v) => Literal::I32(-v),
+                    Literal::I32(v) => Literal::I32(v.wrapping_neg()),
                     Literal::F32(v) => Literal::F32(-v),
+                    Literal::AbstractInt(v) => Literal::AbstractInt(v.wrapping_neg()),
+                    Literal::AbstractFloat(v) => Literal::AbstractFloat(-v),
                     _ => return Err(ConstantEvaluatorError::InvalidUnaryOpArg),
                 },
                 UnaryOperator::LogicalNot => match value {
@@ -1202,6 +1204,7 @@ impl<'a> ConstantEvaluator<'a> {
                 UnaryOperator::BitwiseNot => match value {
                     Literal::I32(v) => Literal::I32(!v),
                     Literal::U32(v) => Literal::U32(!v),
+                    Literal::AbstractInt(v) => Literal::AbstractInt(!v),
                     _ => return Err(ConstantEvaluatorError::InvalidUnaryOpArg),
                 },
             }),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,7 +68,7 @@ This feature allowed you to call `global_id` on any wgpu opaque handle to get a 
 
 #### Naga
 
-- Naga's WGSL front end now allows binary operators to produce values with abstract types, rather than concretizing thir operands. By @jimblandy in [#4850](https://github.com/gfx-rs/wgpu/pull/4850).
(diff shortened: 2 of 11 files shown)


jimblandy:
This may address some items in #4484:

full abstract types:
- https://github.com/gfx-rs/wgpu/issues/4400
- https://github.com/gfx-rs/wgpu/issues/4492
- https://github.com/gfx-rs/wgpu/issues/4435

naga/src/proc/constant_evaluator.rs
```diff
@@ -1193,6 +1193,8 @@ impl<'a> ConstantEvaluator<'a> {
                 UnaryOperator::Negate => match value {
                     Literal::I32(v) => Literal::I32(-v),
                     Literal::F32(v) => Literal::F32(-v),
+                    Literal::AbstractInt(v) => Literal::AbstractInt(-v),
+                    Literal::AbstractFloat(v) => Literal::AbstractFloat(-v),
```
```suggestion
                    Literal::I32(v) => Literal::I32(v.wrapping_neg()),
                    Literal::F32(v) => Literal::F32(-v),
                    Literal::AbstractInt(v) => Literal::AbstractInt(v.wrapping_neg()),
                    Literal::AbstractFloat(v) => Literal::AbstractFloat(-v),
```

This is an existing issue but the spec says:

> If `T` is an integer scalar type and `e` evaluates to the largest negative value, then the result is `e`.

jimblandy: Oh, good catch!

teoxoy (approved):
Good stuff!

PR #4878, opened 2023-12-14
Buffer snatching part 2 - Refactor create_buffer

nical:
**Connections**

Another small step towards #4787 

**Description**

A tiny bit of the ongoing refactoring which I could extract from the rest.

The general idea is to postpone registering the buffer until towards the end of the function to keep a unique reference to it and easily snatch the raw buffer in the error code path.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


Wumpf (approved):
nice, looking a lot better!

Issue #4507, opened 2023-10-05
[meta] Constant evaluator built-in functions support


teoxoy:
- [ ] `bitcast`: https://github.com/gfx-rs/wgpu/issues/7700
- [ ] `select`
- [ ] `all`
- [ ] `any`
- [ ] `quantizeToF16`

comparison

- [x] `abs`: https://github.com/gfx-rs/wgpu/pull/4879
- [x] `min`: https://github.com/gfx-rs/wgpu/pull/5098
- [x] `max`: https://github.com/gfx-rs/wgpu/pull/5098
- [x] `clamp`
- [x] `saturate`: https://github.com/gfx-rs/wgpu/pull/4879

trigonometry

- [x] `cos`: https://github.com/gfx-rs/wgpu/pull/4879
- [x]  `cosh`: https://github.com/gfx-rs/wgpu/pull/4879
- [x]  `sin`: https://github.com/gfx-rs/wgpu/pull/4879
- [x]  `sinh`: https://github.com/gfx-rs/wgpu/pull/4879
- [x]  `tan`: https://github.com/gfx-rs/wgpu/pull/4879
- [x]  `tanh`: https://github.com/gfx-rs/wgpu/pull/4879
- [x]  `acos`: https://github.com/gfx-rs/wgpu/pull/4879
- [x]  `acosh`: https://github.com/gfx-rs/wgpu/pull/4879
- [x]  `asin`: https://github.com/gfx-rs/wgpu/pull/4879
- [x]  `asinh`: https://github.com/gfx-rs/wgpu/pull/4879
- [x]  `atan`: https://github.com/gfx-rs/wgpu/pull/4879
- [x]  `atanh`: https://github.com/gfx-rs/wgpu/pull/4879
- [ ] `atan2`
- [x] `radians`: https://github.com/gfx-rs/wgpu/pull/5098
- [x] `degrees`: https://github.com/gfx-rs/wgpu/pull/5098

decomposition

- [x] `ceil`: https://github.com/gfx-rs/wgpu/pull/5098
- [x] `floor`: https://github.com/gfx-rs/wgpu/pull/5098
- [x]  `round`: https://github.com/gfx-rs/wgpu/pull/4879
- [x] `fract`: https://github.com/gfx-rs/wgpu/pull/5098
- [x] `trunc`: https://github.com/gfx-rs/wgpu/pull/5098
- [ ] `modf`
[...]

PR #4879, opened 2023-12-14
Impl. more component-wise built-ins in const. eval.

ErichDonGubler:
**Connections**

Resolves many numeric built-ins listed under https://github.com/gfx-rs/wgpu/issues/4507, and sets the groundwork for many more.

**Description**

Adds support for evaluating the following built-in functions:

- `abs`
- `saturate`
- `cos`
- `cosh`
- `sin`
- `sinh`
- `tan`
- `tanh`
- `acos`
- `acosh`
- `asin`
- `asinh`
- `atan`
- `atanh`
- `round`
- `step`
- `sqrt`

**Testing**

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/naga/src/lib.rs
+++ b/naga/src/lib.rs
@@ -458,6 +458,10 @@ pub enum VectorSize {
     Quad = 4,
 }
 
+impl VectorSize {
+    const MAX: usize = Self::Quad as u8 as usize;
+}
+
 /// Primitive type for a scalar.
 #[repr(u8)]
 #[derive(Clone, Copy, Debug, Hash, Eq, Ord, PartialEq, PartialOrd)]
--- a/naga/src/proc/constant_evaluator.rs
+++ b/naga/src/proc/constant_evaluator.rs
@@ -869,6 +869,9 @@ impl<'a> ConstantEvaluator<'a> {
             crate::MathFunction::Sinh => {
                 component_wise_float!(self, span, [arg], |e| { Ok([e.sinh()]) })
             }
+            crate::MathFunction::Tan => {
+                component_wise_float!(self, span, [arg], |e| { Ok([e.tan()]) })
+            }
             crate::MathFunction::Sqrt => {
                 component_wise_float!(self, span, [arg], |e| { Ok([e.sqrt()]) })
             }
--- a/naga/src/proc/constant_evaluator.rs
+++ b/naga/src/proc/constant_evaluator.rs
@@ -1,8 +1,208 @@
+use std::iter;
+
+use arrayvec::ArrayVec;
+
 use crate::{
     arena::{Arena, Handle, UniqueArena},
     ArraySize, BinaryOperator, Constant, Expression, Literal, ScalarKind, Span, Type, TypeInner,
     UnaryOperator,
 };
+
+/// A macro that allows dollar signs (`$`) to be emitted by other macros. Useful for generating
+/// `macro_rules!` items that, in turn, emit their own `macro_rules!` items.
+///
+/// Technique stolen directly from
+/// <https://github.com/rust-lang/rust/issues/35853#issuecomment-415993963>.
+macro_rules! with_dollar_sign {
+    ($($body:tt)*) => {
+        macro_rules! __with_dollar_sign { $($body)* }
+        __with_dollar_sign!($);
+    }
+}
+
+macro_rules! gen_component_wise_extractor {
+    (
+        $ident:ident -> $target:ident,
+        literals: [$( $literal:ident => $mapping:ident: $ty:ident ),+ $(,)?],
+        scalar_kinds: [$( $scalar_kind:ident ),* $(,)?],
+    ) => {
+        /// A subset of [`Literal`]s intended to be used for implementing numeric built-ins.
+        enum $target<const N: usize> {
+            $(
+                #[doc = concat!(
+                    "Maps to [`Literal::",
+                    stringify!($mapping),
+                    "`]",
+                )]
+                $mapping([$ty; N]),
+            )+
+        }
+
+        impl From<$target<1>> for Expression {
+            fn from(value: $target<1>) -> Self {
+                match value {
+                    $(
+                        $target::$mapping([value]) => {
+                            Expression::Literal(Literal::$literal(value))
+                        }
+                    )+
+                }
+            }
+        }
+
+        #[doc = concat!(
+            "Attempts to evaluate multiple `exprs` as a combined [`",
+            stringify!($target),
+            "`] to pass to `handler`. ",
+        )]
+        /// If `exprs` are vectors of the same length, `handler` is called for each corresponding
+        /// component of each vector.
+        ///
+        /// `handler`'s output is registered as a new expression. If `exprs` are vectors of the
+        /// same length, a new vector expression is registered, composed of each component emitted
+        /// by `handler`.
+        fn $ident<const N: usize, const M: usize, F>(
+            eval: &mut ConstantEvaluator<'_>,
+            span: Span,
+            exprs: [Handle<Expression>; N],
+            mut handler: F,
+        ) -> Result<Handle<Expression>, ConstantEvaluatorError>
+        where
+            $target<M>: Into<Expression>,
+            F: FnMut($target<N>) -> Result<$target<M>, ConstantEvaluatorError> + Clone,
+        {
+            assert!(N > 0);
+            let err = ConstantEvaluatorError::InvalidMathArg;
+            let mut exprs = exprs.into_iter();
+
+            macro_rules! sanitize {
+                ($expr:expr) => {
+                    eval.eval_zero_value_and_splat($expr, span)
+                        .map(|expr| &eval.expressions[expr])
+                };
+            }
+
+            let new_expr = match sanitize!(exprs.next().unwrap())? {
+                $(
+                    &Expression::Literal(Literal::$literal(x)) => iter::once(Ok(x))
+                        .chain(exprs.map(|expr| {
+                            sanitize!(expr).and_then(|expr| match expr {
+                                &Expression::Literal(Literal::$literal(x)) => Ok(x),
+                                _ => Err(err.clone()),
+                            })
+                        }))
+                        .collect::<Result<ArrayVec<_, N>, _>>()
+                        .map(|a| a.into_inner().unwrap())
+                        .map($target::$mapping)
+                        .and_then(|comps| Ok(handler(comps)?.into())),
+                )+
+                &Expression::Compose { ty, ref components } => match &eval.types[ty].inner {
+                    &TypeInner::Vector { size: _, scalar } => match scalar.kind {
+                        $(ScalarKind::$scalar_kind)|* => {
+                            let first_ty = ty;
+                            let mut component_groups =
+                                ArrayVec::<ArrayVec<_, { crate::VectorSize::MAX }>, N>::new();
+                            component_groups.push(crate::proc::flatten_compose(
+                                first_ty,
+                                components,
+                                eval.expressions,
+                                eval.types,
+                            ).collect());
+                            component_groups.extend(
+                                exprs
+                                    .map(|expr| {
+                                        sanitize!(expr).and_then(|expr| match expr {
+                                            &Expression::Compose { ty, ref components }
+                                                if &eval.types[ty].inner
+                                                    == &eval.types[first_ty].inner =>
+                                            {
+                                                Ok(crate::proc::flatten_compose(
+                                                    ty,
+                                                    components,
+                                                    eval.expressions,
+                                                    eval.types,
+                                                ).collect())
+                                            }
+                                            _ => Err(err.clone()),
+                                        })
+                                    })
+                                    .collect::<Result<ArrayVec<_, { crate::VectorSize::MAX }>, _>>(
+                                    )?,
+                            );
+                            let component_groups = component_groups.into_inner().unwrap();
+                            let mut new_components =
+                                ArrayVec::<_, { crate::VectorSize::MAX }>::new();
+                            for idx in 0..N {
+                                let group = component_groups
+                                    .iter()
+                                    .map(|cs| cs[idx])
+                                    .collect::<ArrayVec<_, N>>()
+                                    .into_inner()
+                                    .unwrap();
+                                new_components.push($ident(
+                                    eval,
+                                    span,
+                                    group,
+                                    handler.clone(),
+                                )?);
+                            }
+                            Ok(Expression::Compose {
+                                ty: first_ty,
+                                components: new_components.into_iter().collect(),
+                            })
+                        }
+                        _ => return Err(err),
+                    },
+                    _ => return Err(err),
+                },
+                _ => return Err(err),
+            }?;
+            eval.register_evaluated_expr(new_expr, span)
+        }
+
+        with_dollar_sign! {
+            ($d:tt) => {
+                #[allow(unused)]
+                #[doc = concat!(
+                    "A convenience macro for using the same RHS for each [`",
+                    stringify!($target),
+                    "`] variant in a call to [`",
+                    stringify!($ident),
+                    "`].",
+                )]
+                macro_rules! $ident {
+                    (
+                        $eval:expr,
+                        $span:expr,
+                        [$d ($d expr:expr),+ $d (,)?],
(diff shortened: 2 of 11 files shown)


naga/src/lib.rs
```diff
@@ -458,6 +458,12 @@ pub enum VectorSize {
     Quad = 4,
 }
 
+impl VectorSize {
+    pub(crate) const fn max() -> usize {
+        Self::Quad as u8 as usize
+    }
+}
```
jimblandy: This could be an associated `const`, right?

naga/src/proc/constant_evaluator.rs
```diff
@@ -2196,3 +2275,46 @@ impl TryFromAbstract<f64> for u32 {
...
+    fn from(value: u32) -> Self {
+        Self::Literal(Literal::U32(value))
+    }
+}
+
+impl From<i32> for Expression {
+    fn from(value: i32) -> Self {
+        Self::Literal(Literal::I32(value))
+    }
+}
```
ErichDonGubler: - [x] issue: I am not inclined to merely assume that these `From` implementations are okay to add; I suspect that they would have been added a long time ago, if they were. However, some of the brevity of the built-ins I'm contributing here depends on these currently. Wanted to ask @jimblandy's/@teoxoy's thoughts about alternatives here, if this isn't acceptable.

    Personally, I actually really like these, minus maybe `From<{u,f}64>`. I don't think it's obvious that these would correspond to `Abstract{Int,Float}`.

ErichDonGubler:
Will fix up CI as I examine changed SPIR-V assembly with `spirv-diff`. Getting that set up locally now!

naga/src/proc/constant_evaluator.rs
ErichDonGubler, line 1: - [x] issue: Actually write a `CHANGELOG` entry.

naga/src/proc/constant_evaluator.rs
```diff
@@ -592,186 +771,82 @@ impl<'a> ConstantEvaluator<'a> {
...
-
-        self.register_evaluated_expr(expr, span)
+            crate::MathFunction::Cos => {
+                component_wise_float!(self, span, [arg], |e| { e.cos().into() })
+            }
+            crate::MathFunction::Cosh => {
+                component_wise_float!(self, span, [arg], |e| { e.cosh().into() })
+            }
+            crate::MathFunction::Round => {
+                component_wise_float!(self, span, [arg], |e| { e.round().into() })
```
teoxoy: According to the WGSL spec this should be `round_ties_even` but the method hasn't been stabilized yet.

naga/src/proc/constant_evaluator.rs
```diff
@@ -592,186 +771,82 @@ impl<'a> ConstantEvaluator<'a> {
...
-
-        let expr = match (&self.expressions[e1], &self.expressions[e2]) {
-            (&Expression::Literal(Literal::F32(a)), &Expression::Literal(Literal::F32(b))) => {
-                Expression::Literal(Literal::F32(a.powf(b)))
+            crate::MathFunction::Abs => {
+                component_wise_scalar(self, span, [arg], |args| match args {
+                    Scalar::AbstractFloat([e]) => e.abs().into(),
+                    Scalar::F32([e]) => e.abs().into(),
+                    Scalar::AbstractInt([e]) => e.abs().into(),
+                    Scalar::I32([e]) => e.abs().into(),
```
```suggestion
                    Scalar::I32([e]) => e.wrapping_abs().into(),
```

We should use `wrapping_abs` per spec.

naga/src/proc/constant_evaluator.rs
```diff
@@ -781,7 +781,14 @@ impl<'a> ConstantEvaluator<'a> {
                 })
             }
             crate::MathFunction::Pow => self.math_pow(arg, arg1.unwrap(), span),
-            crate::MathFunction::Clamp => self.math_clamp(arg, arg1.unwrap(), arg2.unwrap(), span),
+            crate::MathFunction::Clamp => {
+                component_wise_scalar!(
+                    self,
+                    span,
+                    [arg, arg1.unwrap(), arg2.unwrap()],
+                    |e, low, high| { e.clamp(low, high).into() }
```
teoxoy: We were previously erroring with `ConstantEvaluatorError::InvalidClamp` if `low > high` we should still do that since the spec requires it and rust's `clamp` fn will panic in this case.

naga/src/proc/constant_evaluator.rs
teoxoy: > Personally, I actually really like these, minus maybe `From<{u,f}64>`. I don't think it's obvious that these would correspond to `Abstract{Int,Float}`.

I think having these cleans up the code nicely but I think that we shouldn't have them for `{i,f}64` -> `Abstract{Int,Float}` though. Maybe we should introduce 2 wrapper types for `Abstract{Int,Float}` around `{i,f}64`.

naga/src/proc/constant_evaluator.rs
```diff
@@ -1,9 +1,188 @@
...
+                        .map($target::$mapping)
+                        .map(handler),
+                )+
+                &Expression::Compose { ty, ref components } => match &eval.types[ty].inner {
+                    &TypeInner::Vector { size: _, scalar } => match scalar.kind {
+                        $(ScalarKind::$scalar_kind)|* => {
+                            let first_ty = ty;
+                            let mut component_groups =
+                                ArrayVec::<ArrayVec<_, { crate::VectorSize::max() }>, N>::new();
+                            component_groups.push(components.iter().copied().collect());
```
teoxoy: We were previously using `crate::proc::flatten_compose` here and I think we still should or else we might run into trouble.

naga/src/proc/constant_evaluator.rs
```diff
@@ -1,9 +1,188 @@
...
+                            component_groups.extend(
+                                exprs
+                                    .map(|expr| {
+                                        sanitize!(expr).and_then(|expr| match expr {
+                                            &Expression::Compose { ty, ref components }
+                                                if &eval.types[ty].inner
+                                                    == &eval.types[first_ty].inner =>
+                                            {
+                                                Ok(components.iter().copied().collect())
```
teoxoy: Here as well.

teoxoy (changes requested):
Good stuff! Left some comments about handling edge-cases.

naga/src/proc/constant_evaluator.rs
ErichDonGubler: This definitely seems like a valid issue. Given that the soonest that `f{32,64}::round_ties_even` can land is 2024-02-02, I see several options to resolve here (all of which should IMO, be punted to another PR):

* Don't even implement this until Rust 1.77.0 (or perhaps later) stabilizes the method. This seems unnecessary, when compared to other options.
* Use an implementation like [`ndarray-ndimage`'s](https://github.com/imeka/ndarray-ndimage/blob/8b14b4d6ecfbc96a8a052f802e342a7049c68d8f/src/lib.rs#L98), which is linked in [the stabilization report for `round_ties_even`](https://github.com/rust-lang/rust/issues/96710#issuecomment-1875226346); for convenience:

	```rust
	// TODO Use x.round_ties_even() when available on stable
	// https://github.com/rust-lang/rust/issues/96710
	fn round_ties_even(x: f64) -> f64 {
	    let i = x as i32;
	    let f = (x - i as f64).abs();
	    if f == 0.5 {
	        if i & 1 == 1 {
	            // -1.5, 1.5, 3.5, ...
	            (x.abs() + 0.5).copysign(x)
	        } else {
	            (x.abs() - 0.5).copysign(x)
	        }
	    } else {
	        x.round()
	    }
	}
	```
* ~~Do one of the above after we land this. I'd prefer to avoid this, since having incorrect behavior for a brief time can be confusing, but maybe it's acceptable?~~ EDIT: Actually, nevermind; I just thought of the test updates we'd do for this, and _those_ alone sound super not worth it. 😅

naga/src/proc/constant_evaluator.rs
ErichDonGubler: Good catch! Fixed up with e6f60f255.

naga/src/proc/constant_evaluator.rs
ErichDonGubler: Good catch! I've addressed the `low > high` check with 88659b92e, but checking [`f32::clamp`](https://doc.rust-lang.org/std/primitive.f32.html#method.clamp), we also have a mismatch when `low` or `high` are NaN (which would panic on the Rust side). 🤔 I'm not sure what the expected output would be here; do you know?

naga/src/proc/constant_evaluator.rs
ErichDonGubler: And e6589247f2a789c29acab1b8ebf9159bdd90f7ff as well. 🙂

naga/src/proc/constant_evaluator.rs
ErichDonGubler: Fixed up with 3d8bff2eb83e7649a8d6f923868c13d47c19acc8. I _think_ I'm using it correctly?

naga/src/proc/constant_evaluator.rs
teoxoy: The spec says that NaNs and Infinities are not allowed in const contexts, the check below validates those out.

https://github.com/gfx-rs/wgpu/blob/4b82121501a61c2c2e11cb472d70ba54af3aa12d/naga/src/proc/constant_evaluator.rs#L1548-L1550

naga/src/proc/constant_evaluator.rs
teoxoy: Using a polyfill for now is what I had in mind. I think we can just copy-paste `ndarray-ndimage`'s impl with a permalink where we got it from and a TODO to switch it out once it becomes stable and our MSRV reaches that version.

ErichDonGubler: Fixed up as suggested with 691f37a68c21d09ab6c731cfbc1dc81809414c08, minus that I have changed the `let i = x as i32;` to be an `i64` instead, to retain precision. I _think_ this is a good idea, but I'm not positive. I'm not enough of a numerics expert to assert such. 😅

naga/src/proc/constant_evaluator.rs
jimblandy: We may eventually have 64-bit floats, so having something that automatically assumed that Rust f64 values correspond to WGSL abstract floats would be pretty confusing.

jimblandy: Yeah, an `Abstract<T>` wrapper wouldn't be a bad idea.

naga/src/lib.rs
ErichDonGubler: I have just changed this to be a `const MAX: …` with my [latest force-push](https://github.com/gfx-rs/wgpu/compare/99c54f4731522cbbcbf0f485f1c6336da1f406a5..b860233cf7dfbd3be4b15f27401171e2ab40e9ef#diff-ce893561808eb35d7490c99adf1bf104c82a02caed616e98f00766950f46c9d5). Resolving!

naga/src/proc/constant_evaluator.rs
ErichDonGubler: So, I just did a bunch of stuff ([diff](https://github.com/gfx-rs/wgpu/compare/99c54f4731522cbbcbf0f485f1c6336da1f406a5..b860233cf7dfbd3be4b15f27401171e2ab40e9ef#diff-ce893561808eb35d7490c99adf1bf104c82a02caed616e98f00766950f46c9d5), all changes are in `constant_evaluator.rs`) with the `enum`s subsetting `Literal` that have been added by this PR:

* I now use the `gen_component_wise_extractor` macro to generate these, with largely the same input as before. The only thing I had to add to the current macro input is the concrete types that will be stored in the subset `enum`.
* I have removed the `From` implementations on primitive types behind the controversy in this discussion. Instead, I now use `impl From<$target<1>> for Expression` inside the macro (i.e., `Scalar<1>`). I chose to compromise a bit on syntax to do this: you now have to return a `[T; 1]` instead of just a `T`.

    I suspect we may actually want sizes greater than 1 later, but I'm not sure. If not, this shouldn't be hard to change back to requiring "just" `T` later.
* Not super relevant for this convo, but FTR: I have also removed `pub(super)` visibility. It wasn't necessary.

I think this pretty neatly resolves this conversation. The `abs` and `acos` built-ins showcase how code looks now:

```rust
            crate::MathFunction::Abs => {
                component_wise_scalar(self, span, [arg], |args| match args {
                    Scalar::AbstractFloat([e]) => Ok(Scalar::AbstractFloat([e.abs()])),
                    Scalar::F32([e]) => Ok(Scalar::F32([e.abs()])),
                    Scalar::AbstractInt([e]) => Ok(Scalar::AbstractInt([e.abs()])),
                    Scalar::I32([e]) => Ok(Scalar::I32([e.wrapping_abs()])),
                    Scalar::U32([e]) => Ok(Scalar::U32([e])), // TODO: just re-use the expression, ezpz
                })
            }
            crate::MathFunction::Acos => {
                component_wise_float!(self, span, [arg], |e| { Ok([e.acos()]) })
            }
```

Still pretty readable, methinks!

teoxoy: > I suspect we may actually want sizes greater than 1 later, but I'm not sure. If not, this shouldn't be hard to change back to requiring "just" `T` later.

I don't think we will, or I hope we don't as scalars should only hold one value.
It's a bit odd to see the values being wrapped in an array with length 1.

@jimblandy thoughts on this?

ErichDonGubler: @teoxoy: I think we might need to improve the naming of these convenience structures to handle your concern? `Float` and `Scalar`  are essentially a representation of components. They could be a single scalar, or multiple components of a `vec`. Having more than 1 would "just" mean that we want to emi a `vec` and not just a single scalar value.

naga/src/proc/constant_evaluator.rs
```diff
@@ -1,9 +1,226 @@
...
+/// Technique stolen directly from
+/// <https://github.com/rust-lang/rust/issues/35853#issuecomment-415993963>.
+macro_rules! with_dollar_sign {
+    ($($body:tt)*) => {
+        macro_rules! __with_dollar_sign { $($body)* }
+        __with_dollar_sign!($);
+    }
+}
+
+macro_rules! gen_component_wise_extractor {
```
jimblandy, line 23: I don't want to be negative, but the truth is that I cannot understand this at all.

Is there a way we could implement this without macro-generating macros?

jimblandy (dismissed):

naga/src/proc/constant_evaluator.rs
ErichDonGubler: @teoxoy, @jimblandy: Ping on this.

naga/src/proc/constant_evaluator.rs
ErichDonGubler: Using `cargo-expand`, most likely! :^)

I don't have any objection to inlining the output of the current macro. I'm cautious of DRY in the long term, but I don't think the cost will be noticeable until and unless we discover that the current body of this macro is incomplete or wrong. I actually consider this likely, but if you're willing to fix that sort of thing up, I don't see a reason to let perfect be the enemy of good (viz., merging this).

naga/src/proc/constant_evaluator.rs
teoxoy: Ok, I think this is fine for now, we can revisit it later.

teoxoy (approved):

naga/src/proc/constant_evaluator.rs
ErichDonGubler: Spoke with @jimblandy at length on a call here. Tentative conclusion: We'd like to iterate on this, but for now, we can merge this.

naga/src/proc/constant_evaluator.rs
```diff
@@ -1,9 +1,227 @@
...
+        U32 => U32: u32,
+        I32 => I32: i32,
+    ],
+    scalar_kinds: [
+        Float,
+        AbstractFloat,
+        Sint,
+        Uint,
+        AbstractInt,
+    ],
```
ErichDonGubler, line 210: issue: Ah, I was missing `Sint` here. Fixed!

jimblandy:
:tada:

PR #4880, opened 2023-12-14
Simplify `ResourceMaps`

nical:
**Connections**

This PR is rebased on top of #4878 
Another step towards #4787 

**Description**

`ResourceMaps` had a rather convoluted system for erasing types that isn't needed anywhere except in one place in the `triage_resource` function. Everywhere else we are always dealing with specific types so using a member of the resource maps is simpler than going through an abstraction. More importantly there was a constraint that all contents of the resource maps implement the Resource trait which got in the way of some of the ongoing buffer snatching changes.
    
This commit simplifies this by removing the abstraction. Each resource type has its hash map directly in `ResourceMaps` and it is easier to have different requirements and behaviors depending on the type of the each resource.

The main wrinkle is that for `triage_resource` to still work without knowing about the resource type, it takes a callback to go from `ResourceMaps` to the right hash map. it could be done with some trickery in `ResourceMap` but I found that this was the simplest way albeit not the most elegant.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/device/life.rs
+++ b/wgpu-core/src/device/life.rs
@@ -27,18 +27,18 @@ use thiserror::Error;
 /// A struct that keeps lists of resources that are no longer needed by the user.
 #[derive(Default)]
 pub(crate) struct ResourceMaps<A: HalApi> {
-    pub(crate) buffers: FastHashMap<BufferId, Arc<Buffer<A>>>,
-    pub(crate) staging_buffers: FastHashMap<StagingBufferId, Arc<StagingBuffer<A>>>,
-    pub(crate) textures: FastHashMap<TextureId, Arc<Texture<A>>>,
-    pub(crate) texture_views: FastHashMap<TextureViewId, Arc<TextureView<A>>>,
-    pub(crate) samplers: FastHashMap<SamplerId, Arc<Sampler<A>>>,
-    pub(crate) bind_groups: FastHashMap<BindGroupId, Arc<BindGroup<A>>>,
-    pub(crate) bind_group_layouts: FastHashMap<BindGroupLayoutId, Arc<BindGroupLayout<A>>>,
-    pub(crate) render_pipelines: FastHashMap<RenderPipelineId, Arc<RenderPipeline<A>>>,
-    pub(crate) compute_pipelines: FastHashMap<ComputePipelineId, Arc<ComputePipeline<A>>>,
-    pub(crate) pipeline_layouts: FastHashMap<PipelineLayoutId, Arc<PipelineLayout<A>>>,
-    pub(crate) render_bundles: FastHashMap<RenderBundleId, Arc<RenderBundle<A>>>,
-    pub(crate) query_sets: FastHashMap<QuerySetId, Arc<QuerySet<A>>>,
+    pub buffers: FastHashMap<BufferId, Arc<Buffer<A>>>,
+    pub staging_buffers: FastHashMap<StagingBufferId, Arc<StagingBuffer<A>>>,
+    pub textures: FastHashMap<TextureId, Arc<Texture<A>>>,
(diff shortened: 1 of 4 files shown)


wgpu-core/src/device/life.rs
```diff
@@ -429,7 +404,8 @@ impl<A: HalApi> LifetimeTracker<A> {
             if is_removed {
                 on_remove(&id, resource);
                 removed_resources.push(resource.clone());
-                non_referenced_resources.insert(id, resource.clone());
+                get_resource_map(non_referenced_resources).insert(id, resource.clone());
+                // TODO(nical)
```
nical:
```suggestion
```

nical:
After some more poking, I think I won't actually need this to move forward with buffer snatching (but I prefer the way it works after this change so I'm keen on landing it).

wgpu-core/src/device/life.rs
```diff
@@ -14,111 +14,81 @@ use crate::{
...
+    pub(crate) textures: FastHashMap<TextureId, Arc<Texture<A>>>,
+    pub(crate) texture_views: FastHashMap<TextureViewId, Arc<TextureView<A>>>,
+    pub(crate) samplers: FastHashMap<SamplerId, Arc<Sampler<A>>>,
+    pub(crate) bind_groups: FastHashMap<BindGroupId, Arc<BindGroup<A>>>,
+    pub(crate) bind_group_layouts: FastHashMap<BindGroupLayoutId, Arc<BindGroupLayout<A>>>,
+    pub(crate) render_pipelines: FastHashMap<RenderPipelineId, Arc<RenderPipeline<A>>>,
+    pub(crate) compute_pipelines: FastHashMap<ComputePipelineId, Arc<ComputePipeline<A>>>,
+    pub(crate) pipeline_layouts: FastHashMap<PipelineLayoutId, Arc<PipelineLayout<A>>>,
+    pub(crate) render_bundles: FastHashMap<RenderBundleId, Arc<RenderBundle<A>>>,
+    pub(crate) query_sets: FastHashMap<QuerySetId, Arc<QuerySet<A>>>,
```
Wumpf: nit: `pub(crate)` clutter. You _could_ just make all of this regular `pub` since `life` is dragged in as `mod life` (not `pub mod life`)

wgpu-core/src/device/life.rs
```diff
@@ -14,111 +14,81 @@ use crate::{
...
+            compute_pipelines: FastHashMap::default(),
+            pipeline_layouts: FastHashMap::default(),
+            render_bundles: FastHashMap::default(),
+            query_sets: FastHashMap::default(),
+        }
     }
+
     pub(crate) fn clear(&mut self) {
-        self.maps.iter_mut().for_each(|(_t, map)| map.clear_map());
+        self.buffers.clear();
```
Wumpf: It's longer, but both here and in `extend` it would be way better to use destructuring on the struct. This way there's a guarantee that none of the fields is missed. This makes this safer against resource types being added and as as a reviewer I refuse to manually check completelnes given that there's a compiler mechanism to do so that will warn when one of the destructured elements is unused ;)

i.e. this
``` 
let ResourceMaps {
    staging_buffers,
    textures,
    // ...
} = &mut self;
staging_buffers.clear()
// ...
```

Wumpf (approved):
nice, makes sense to use the fact that we always know the type!
only cosmetic comments. I think the callback on `triage_resource` isn't all that bad indeed, seems like the least invasive thing for the moment :)

wgpu-core/src/device/life.rs
nical: The dreadful `clippy::pattern_type_mismatch` lint really gets in the way of that, sadly.

nical: I disabled the lint locally and will petition to get rid of it globally in a followup.

PR #4887, opened 2023-12-17
Allow clippy::pattern_type_mismatch

nical:
**Description**

I hate this lint with passion. 

Its [documentation](https://rust-lang.github.io/rust-clippy/master/index.html#/pattern_type_mismatch) itself admits that it does not fix anything serious:

> It isn’t bad in general. But in some contexts it can be desirable because it increases ownership hints in the code, and will guard against some changes in ownership.

In order to work around it we even have to disable another clippy lint (`needless_borrowed_reference`) which it clashes with.

In all of my years of rust I have never run into an issue that could have been prevented by this lint. Rust's type system does not leave much room for ambiguity between owned and borrowed data.
There is a language keyword, `ref`, which you typically don't see much if at all in rust code, because it was made obsolete by the syntax that the lint aims to prevent, but we have to sprinkle it in places to pass clippy.

Rust's syntax is pretty clear when it comes to pattern matching:

```rust
let something = &mut Some(thing);
match something {
    Some(thing) => {
        // thing is &mut because we matched an &mut Option<Thing>
    }
    None => {}
}

let something_moved = Some(thing);
match something_moved {
    Some(thing) => {
        // thing is by value because we matched an Option<Thing> (by value).
    }
    None => {}
}
```

It could be argued that some code could benefit from being a bit explicit with `ref` keyword (I don't think so but there's always going to be someone who does), like there are cases where spelling out types makes some code easier to understand even if type inference makes it optional.
But enforcing the explicit `ref` for references in all matched patterns makes as much sense as disallowing inferred types entirely.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -3073,7 +3073,7 @@ impl<A: HalApi> Device<A> {
         if validated_stages.contains(wgt::ShaderStages::FRAGMENT) {
             for (i, output) in io.iter() {
                 match color_targets.get(*i as usize) {
-                    Some(&Some(ref state)) => {
+                    Some(Some(state)) => {
                         validation::check_texture_format(state.format, &output.ty).map_err(
                             |pipeline| {
                                 pipeline::CreateRenderPipelineError::ColorState(
--- a/wgpu-core/src/lib.rs
+++ b/wgpu-core/src/lib.rs
@@ -34,17 +34,15 @@
     // For some reason `rustc` can warn about these in const generics even
     // though they are required.
     unused_braces,
-    // Clashes with clippy::pattern_type_mismatch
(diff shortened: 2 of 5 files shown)


cwfitzgerald (approved):
Agreed!

kpreid:
> Its [documentation](https://rust-lang.github.io/rust-clippy/master/index.html#/pattern_type_mismatch) itself admits that it does not fix anything serious:

This is normal for **`restriction`** lints. The docs could definitely be improved for all restriction lints by changing "Why is this bad?" (generic template) to "Why might you want to restrict this?” — I've been thinking about doing that.

But also, because it is a `restriction` lint, you shouldn't need to `allow` it, just not mention it.

PR #4888, opened 2023-12-17
Web: add support for more `RawWindowHandle` variants

daxpedda:
**Connections**
- #4202 
- #4597

**Description**
This implements handling of the new `RawWindowHandle` variants specifically introduced for Web: `WebCanvas` and `WebOffscreenCanvas`.

**Testing**
It isn't, the new version of Winit will use `WebCanvas`.

**Notes**
This didn't seem to justify an additional entry in the changelog, because its part of #4202, which is still in the unreleased section.

### DIFF
--- a/wgpu-hal/src/gles/web.rs
+++ b/wgpu-hal/src/gles/web.rs
@@ -1,6 +1,6 @@
 use glow::HasContext;
 use parking_lot::{Mutex, RwLock};
-use wasm_bindgen::JsCast;
+use wasm_bindgen::{JsCast, JsValue};
 
 use super::TextureFormatDesc;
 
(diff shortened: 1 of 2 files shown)


wgpu-hal/src/gles/web.rs
```diff
@@ -147,22 +147,33 @@ impl crate::Instance<super::Api> for Instance {
         _display_handle: raw_window_handle::RawDisplayHandle,
         window_handle: raw_window_handle::RawWindowHandle,
     ) -> Result<Surface, crate::InstanceError> {
-        if let raw_window_handle::RawWindowHandle::Web(handle) = window_handle {
-            let canvas: web_sys::HtmlCanvasElement = web_sys::window()
+        let canvas: web_sys::HtmlCanvasElement = match window_handle {
```
cwfitzgerald, line 150: Don't we need to do this for WebGPU as well?

cwfitzgerald (changes requested):

wgpu-hal/src/gles/web.rs
daxpedda: Indeed!
Done as well.

cwfitzgerald (approved):
This is great, and I'm happy rwh/winit made this change!

PR #4893, opened 2023-12-18
Avoid allocating memory every time we might log a label

nical:
**Description**

We allocate a String every time we want to get a label for logging. The string is also allocated when logging is disabled. Either way, the allocation is unnecessary. This commit replaces the String with a dyn Debug reference which does not need any allocation.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


Wumpf (approved):
nice, much better

PR #4894, opened 2023-12-18
Remove some locks in BindGroup

nical:
These are only written to clear the vectors when triaging bind groups for destruction, which is not necessary. We can let the reference counts drop when the bind group is dropped.

**Description**

The three locks aren't necessary. They allow us to clear references to the buffers and textures a tad earlier when triaging dropped bind groups, but we can let that happen automatically when the `BindGroup` struct is actually dropped soon after.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


Wumpf (approved):
nice catch! reasoning makes sense and looks straight forward

PR #4895, opened 2023-12-18
Check that raw buffers and raw bind groups are valid

nical:
**Connections**

Another step towards #4787.
Rebased on top of #4894.

**Description**

This addresses a lot of wgpu-core's code which assumes that the raw buffer is always accessible (`Buffer::raw(guard)` was unconditionally unwrapping the buffer), and instead returns an error when this is not the case.

In addition, accessing a raw bind group now requires checking that the underlying raw buffers are also accessible.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


cwfitzgerald (approved):
Great stuff!

PR #4896, opened 2023-12-18
Buffer snatching

nical:
**Connections**

Fixes the buffer part of #4787
Rebased on top of #4895 and #4894

**Description**

Move (snatch) the raw buffer out of the wgpu-core buffer when destroying the latter, putting it into a `DestroyedBuffer<A>` structure which lifecycle is tracked independently. The idea is that the `Buffer<A>` can be held alive by other resource like bind groups in an a destroyed state, which we go ahead ad deallocated the snatched buffer as soon as it is safe to do so.

For garage collected environments such as the web it means we can rely on `destroy()` for releasing resources in a reasonable time.

I think it might even actually work.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


cwfitzgerald (approved):
This is great, end to end!

PR #4899, opened 2023-12-19
Expose maximum_frame_latency

Wumpf:
**Connections**
 * Scratches the surface of https://github.com/gfx-rs/wgpu/issues/2869
 * _Way_ better alternative to setting to `Immediate` as initially proposed here
      * https://github.com/emilk/egui/pull/3714
      * https://github.com/rerun-io/rerun/pull/4575

**Description**
Previously, the swapchain buffer count was hardcoded to be always 3 no matter what is supported. This PR exposes the desired swap chain count as a minimal extension to the wgc presentation setup.

Among other things this allows one now to switch between "Game Mode" and "Classic Game Mode" as classified here
https://www.intel.com/content/www/us/en/developer/articles/code-sample/sample-application-for-direct3d-12-flip-model-swap-chains.html

Prime motivation for this is to reduce perceived lag in egui applications where a buffer count of 2 makes usually more sense (assuming a stable and low workload).

**Testing**
Experimented with buffer count of 2 in egui example on Mac & Windows. Gives a lot less latency.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,7 +67,7 @@ Wgpu now exposes backend feature for the Direct3D 12 (`dx12`) and Metal (`metal`
 - Added support for the float32-filterable feature. By @almarklein in [#4759](https://github.com/gfx-rs/wgpu/pull/4759)
 - GPU buffer memory is released during "lose the device". By @bradwerth in [#4851](https://github.com/gfx-rs/wgpu/pull/4851)
 - wgpu and wgpu-core features are now documented on docs.rs. By @wumpf in [#4886](https://github.com/gfx-rs/wgpu/pull/4886)
-- `SurfaceConfiguration` now exposes `desired_swap_chain_size` which was previously hard-coded to 2. By setting it to 1 you can reduce latency. By @emilk & @wumpf in [#4899](https://github.com/gfx-rs/wgpu/pull/4899)
+- `SurfaceConfiguration` now exposes `desired_maximum_frame_latency` which was previously hard-coded to 2. By setting it to 1 you can reduce latency under the risk of making GPU & CPU work sequential. Currently, on DX12 this affects the `MaximumFrameLatency`, on all other backends except OpenGL the size of the swapchain (on OpenGL this has no effect). By @emilk & @wumpf in [#4899](https://github.com/gfx-rs/wgpu/pull/4899)
 - DeviceLostClosure is guaranteed to be invoked exactly once. By @bradwerth in [#4862](https://github.com/gfx-rs/wgpu/pull/4862)
 
 #### OpenGL
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,7 +67,7 @@ Wgpu now exposes backend feature for the Direct3D 12 (`dx12`) and Metal (`metal`
 - Added support for the float32-filterable feature. By @almarklein in [#4759](https://github.com/gfx-rs/wgpu/pull/4759)
 - GPU buffer memory is released during "lose the device". By @bradwerth in [#4851](https://github.com/gfx-rs/wgpu/pull/4851)
 - wgpu and wgpu-core features are now documented on docs.rs. By @wumpf in [#4886](https://github.com/gfx-rs/wgpu/pull/4886)
-- `SurfaceConfiguration` now exposes `desired_swap_chain_size` which was previously hard-coded to 3. By setting it to 2 you can reduce latency. By picking a large value you can ensure higher throughput. By @emilk & @wumpf in [#4899](https://github.com/gfx-rs/wgpu/pull/4899)
+- `SurfaceConfiguration` now exposes `desired_swap_chain_size` which was previously hard-coded to 2. By setting it to 1 you can reduce latency. By @emilk & @wumpf in [#4899](https://github.com/gfx-rs/wgpu/pull/4899)
 - DeviceLostClosure is guaranteed to be invoked exactly once. By @bradwerth in [#4862](https://github.com/gfx-rs/wgpu/pull/4862)
 
 #### OpenGL
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,7 @@ Wgpu now exposes backend feature for the Direct3D 12 (`dx12`) and Metal (`metal`
 - No longer validate surfaces against their allowed extent range on configure. This caused warnings that were almost impossible to avoid. As before, the resulting behavior depends on the compositor. By @wumpf in [#4796](https://github.com/gfx-rs/wgpu/pull/4796)
 - Added support for the float32-filterable feature. By @almarklein in [#4759](https://github.com/gfx-rs/wgpu/pull/4759)
 - wgpu and wgpu-core features are now documented on docs.rs. By @wumpf in [#4886](https://github.com/gfx-rs/wgpu/pull/4886)
+- `SurfaceConfiguration` now exposes `desired_swap_chain_size` which was previously hard-coded to 3. By setting it to 2 you can reduce latency. By picking a large value you can ensure higher throughput. By @emilk & @wumpf in [#????](https://github.com/gfx-rs/wgpu/pull/????)
 
 #### OpenGL
 - `@builtin(instance_index)` now properly reflects the range provided in the draw call instead of always counting from 0. By @cwfitzgerald in [#4722](https://github.com/gfx-rs/wgpu/pull/4722).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,7 +66,7 @@ Wgpu now exposes backend feature for the Direct3D 12 (`dx12`) and Metal (`metal`
 - No longer validate surfaces against their allowed extent range on configure. This caused warnings that were almost impossible to avoid. As before, the resulting behavior depends on the compositor. By @wumpf in [#4796](https://github.com/gfx-rs/wgpu/pull/4796)
 - Added support for the float32-filterable feature. By @almarklein in [#4759](https://github.com/gfx-rs/wgpu/pull/4759)
 - wgpu and wgpu-core features are now documented on docs.rs. By @wumpf in [#4886](https://github.com/gfx-rs/wgpu/pull/4886)
-- `SurfaceConfiguration` now exposes `desired_swap_chain_size` which was previously hard-coded to 3. By setting it to 2 you can reduce latency. By picking a large value you can ensure higher throughput. By @emilk & @wumpf in [#????](https://github.com/gfx-rs/wgpu/pull/????)
+- `SurfaceConfiguration` now exposes `desired_swap_chain_size` which was previously hard-coded to 3. By setting it to 2 you can reduce latency. By picking a large value you can ensure higher throughput. By @emilk & @wumpf in [#4899](https://github.com/gfx-rs/wgpu/pull/4899)
 
 #### OpenGL
 - `@builtin(instance_index)` now properly reflects the range provided in the draw call instead of always counting from 0. By @cwfitzgerald in [#4722](https://github.com/gfx-rs/wgpu/pull/4722).
--- a/wgpu-hal/src/dx12/mod.rs
+++ b/wgpu-hal/src/dx12/mod.rs
@@ -654,13 +654,18 @@ impl crate::Surface<Api> for Surface {
 
         let non_srgb_format = auxil::dxgi::conv::map_texture_format_nosrgb(config.format);
 
+        // Nvidia recommends to use 1-2 more buffers than the maximum latency
+        // https://developer.nvidia.com/blog/advanced-api-performance-swap-chains/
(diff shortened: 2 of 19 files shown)


emilk:
I just want to emphasize what a huge difference changing going from 3 to 2 is on Mac, at least in windowed mode on a 60 Hz monitor. It takes the application from feeling sluggish to feeling responsive (using `AutoVsync`).

When the CPU is fast (e.g. producing a frame in 1ms), `SwapChain::get_next_texture` will block, saving CPU time and preventing tearing.

Interestingly though, when the CPU is too slow (e.g. taking 20ms to prepare a frame), then `SwapChain::get_next_texture` will _not_ block. I'm not sure why, but I like it.

So: I'm planning on making `2` the default for all `eframe` apps

CHANGELOG.md
```diff
@@ -66,6 +66,7 @@ Wgpu now exposes backend feature for the Direct3D 12 (`dx12`) and Metal (`metal`
 - No longer validate surfaces against their allowed extent range on configure. This caused warnings that were almost impossible to avoid. As before, the resulting behavior depends on the compositor. By @wumpf in [#4796](https://github.com/gfx-rs/wgpu/pull/4796)
 - Added support for the float32-filterable feature. By @almarklein in [#4759](https://github.com/gfx-rs/wgpu/pull/4759)
 - wgpu and wgpu-core features are now documented on docs.rs. By @wumpf in [#4886](https://github.com/gfx-rs/wgpu/pull/4886)
+- `SurfaceConfiguration` now exposes `desired_swap_chain_size` which was previously hard-coded to 3. By setting it to 2 you can reduce latency. By picking a large value you can ensure higher throughput. By @emilk & @wumpf in [#????](https://github.com/gfx-rs/wgpu/pull/????)
```
emilk:
```suggestion
- `SurfaceConfiguration` now exposes `desired_swap_chain_size` which was previously hard-coded to 3. By setting it to 2 you can reduce latency. By picking a large value you can ensure higher throughput. By @emilk & @wumpf in [#4899](https://github.com/gfx-rs/wgpu/pull/4899)
```

emilk (approved):

cwfitzgerald (changes requested):
Marking request review per discussion in call

Wumpf:
code looks mostly now like we want it to be I think! For sure way more expressive of the stated goal and in close alignment with
* #2869

Need to update changelog and test whether `maximum_frame_latency == 1` causes gpu/cpu to run in sync on Vulkan & Metal (hoping to find this answered with 'no, they still run parallel')

wgpu-hal/src/dx12/mod.rs
```diff
@@ -654,13 +654,18 @@ impl crate::Surface<Api> for Surface {
 
         let non_srgb_format = auxil::dxgi::conv::map_texture_format_nosrgb(config.format);
 
+        // Nvidia recommends to use 1-2 more buffers than the maximum latency
+        // https://developer.nvidia.com/blog/advanced-api-performance-swap-chains/
+        // For high latency extra buffers seems excessive, so go with a minimum of 3 and beyond that add 1.
+        let swap_chain_buffer = (config.maximum_frame_latency + 1).min(3);
```
emilk, line 666: won't this make double-buffering impossible?

wgpu-hal/src/dx12/mod.rs
```diff
@@ -687,7 +692,9 @@ impl crate::Surface<Api> for Surface {
                         quality: 0,
                     },
                     buffer_usage: dxgitype::DXGI_USAGE_RENDER_TARGET_OUTPUT,
-                    buffer_count: config.swap_chain_size,
+                    // Nvidia recommends to use 1-2 more buffers than the maximum latency
+                    // https://developer.nvidia.com/blog/advanced-api-performance-swap-chains/
+                    buffer_count: swap_chain_buffer,
```
emilk: duplicate comment

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -1838,7 +1838,12 @@ impl crate::Adapter<super::Api> for super::Adapter {
             .collect();
         Some(crate::SurfaceCapabilities {
             formats,
-            swap_chain_sizes: caps.min_image_count..=max_image_count,
+            // TODO: Right now we're always trunkating the swap chain
+            // (presumably - we're actually setting the min image count which isn't necessarily the swap chain size)
+            // Instead, we should use extensions when available to wait in present.
+            // See https://github.com/gfx-rs/wgpu/issues/2869
+            maximum_frame_latency: caps.min_image_count.saturating_sub(1)
+                ..=max_image_count.saturating_sub(1),
```
emilk: this should also have an `.max(1)` on it, since the docstring for `SurfaceCapabilities::maximum_frame_latency` requires at least `1`

wgpu-types/src/lib.rs
```diff
@@ -5107,13 +5107,16 @@ pub struct SurfaceConfiguration<V> {
...
+    /// Defaults to 2 when created via `wgpu::Surface::get_default_config`.
+    /// Recommended to use 2 (or higher) for high throughput, 1 for low latency.
     ///
-    /// Recommended to use 3 (or higher) for high throughput, 2 for low latency.
     /// This is a hint to the backend implementation and will be clamped to the supported range.
-    pub desired_swap_chain_size: u32,
+    /// As a consequence either the maximum frame latency is set directly on the swap chain,
+    /// or waits on present to avoid exceeding the maximum frame latency,
+    /// or the swap chain size is set to max-latency + 1.
+    pub desired_maximum_frame_latency: u32,
```
emilk, line 5129: we should make a note on what happens if the user sets this to zero.

Do all backends clamp this to a valid range?

CHANGELOG.md
```diff
@@ -66,6 +66,7 @@ Wgpu now exposes backend feature for the Direct3D 12 (`dx12`) and Metal (`metal`
 - No longer validate surfaces against their allowed extent range on configure. This caused warnings that were almost impossible to avoid. As before, the resulting behavior depends on the compositor. By @wumpf in [#4796](https://github.com/gfx-rs/wgpu/pull/4796)
 - Added support for the float32-filterable feature. By @almarklein in [#4759](https://github.com/gfx-rs/wgpu/pull/4759)
 - wgpu and wgpu-core features are now documented on docs.rs. By @wumpf in [#4886](https://github.com/gfx-rs/wgpu/pull/4886)
+- `SurfaceConfiguration` now exposes `desired_swap_chain_size` which was previously hard-coded to 3. By setting it to 2 you can reduce latency. By picking a large value you can ensure higher throughput. By @emilk & @wumpf in [#4899](https://github.com/gfx-rs/wgpu/pull/4899)
```
daxpedda:
```suggestion
- `SurfaceConfiguration` now exposes `desired_swap_chain_size` which was previously hard-coded to 3. By setting it to 1 you can reduce latency. By picking a large value you can ensure higher throughput. By @emilk & @wumpf in [#4899](https://github.com/gfx-rs/wgpu/pull/4899)
```
According to the documentation on `SurfaceConfiguration` its now 2 by default and it's recommended to set it to 1 to reduce latency.

wgpu-types/src/lib.rs
```diff
@@ -5107,6 +5107,16 @@ pub struct SurfaceConfiguration<V> {
     /// AutoNoVsync will gracefully do a designed sets of fallbacks if their primary modes are
     /// unsupported.
     pub present_mode: PresentMode,
+    /// Desired maximum number of frames that the presentation engine should queue in advance.
+    ///
+    /// Defaults to 2 when created via `wgpu::Surface::get_default_config`.
+    /// Recommended to use 2 (or higher) for high throughput, 1 for low latency.
```
daxpedda: Could we clarify exactly what "high throughput" means?
Maybe it just isn't clear to me because I'm clueless on the topic ...

CHANGELOG.md
Wumpf: yep, as mentiond below I haven't updated the changelog again yet

wgpu-types/src/lib.rs
Wumpf: Hmm I'll try to expand a bit on it, but it's intentionally handwavy :/

wgpu-hal/src/dx12/mod.rs
Wumpf: The number of swap chain buffer isn't _actually_ the thing we care about to begin with: we care about the latency from recording to display and DX12 can set this more directly than using [`SetMaximumFrameLatency`](https://learn.microsoft.com/en-us/windows/win32/api/dxgi/nf-dxgi-idxgidevice1-setmaximumframelatency). Essentially it decouples the "wait on queue is full" from the number of buffers that are used. Which is why @cwfitzgerald actually advocated originally to just always use 3. But naturally, the desired frequency isn't entirely decoupled which is why Nvidia suggests to over provision by at least 1.
To my understanding the major perk of this more principled approach is that instead of blocking on `Present`/buffer flip, y we wait at the begin of a frame using a "waitable object" which is signaled **after** a present is done. See [getframelatencywaitableobject](https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_3/nf-dxgi1_3-idxgiswapchain2-getframelatencywaitableobject). This means that if there's (for whatever reason) a longer wait, you have frame preparation including all the input query much much closer to the actual presenting of the frame.

Timing your frame preparation more accurately and consistently gets you into the even more complex real of "frame pacing" which is all about pacing your frame recording to hit the next screen refresh as closely as possible; there's various mutally inconsistent APIs popping up these days (see #2869) so it's that part of the equation isn't that easy to solve.

Unfortunately, on Vulkan this works only with some extensions. To be implemented. (Oh and since you're wondering now: Rerun doesn't use DX12 right now for a bunch or reasons, we always stick with Vulkan)

I'm still learning this stuff, so maybe @cwfitzgerald can correct me if I got something wrong :)

wgpu-hal/src/dx12/mod.rs
Wumpf: oops

wgpu-hal/src/dx12/mod.rs
Wumpf: tl'dr: The trick is that on dx12 `SetMaximumFrameLatency` takes `maximum_frame_latency` as-is.

wgpu-hal/src/vulkan/adapter.rs
Wumpf: good catch. I checked the spec and I'd now say my comment is actually wrong. `maximum_frame_latency`==0 (if supported) to my understanding now should mean that gpu/cpu are in sequence (although @cwfitzgerald was skeptical about this before). I'll have to experiment though to confirm

wgpu-types/src/lib.rs
Wumpf: Yeah right now, they all clamp to the range - this is unchanged from the thing we pair programmed on.
I think in the future we want to go as in Connor's original proposal and expose this just like we expose format and size restrictions. We'll then drop the `desired_` part and check again it as with other properties.

Expanding the comment!

wgpu-types/src/lib.rs
Wumpf: rewrote that section now to be a lot more precise :)

Wumpf:
Got around to test with Vulkan on windows of when GPU/CPU ending up to working in sync.
My test setup was to render the fractal of wgpu-profiler's demo with more iterations (making it easy to control the gpu load) and do articial cpu load by via thread sleep. I then compared frame times with `maximum_frame_latency` 1 and `maximum_frame_latency` 2.
Turns out with a frame latency setting of 1 the gpu won't have any work between `get_current_texture` and the next queue submit. Or in other words: When I put the sleep _after_  `get_current_texture`, the gpu will be fully idle since there's no work to be done at all. Meaning at the very least need to adjust the documentation.

On DX12 I'm hitting some odd bug making frame time expand to 1 full second, so I don't have test results from that so far.

Not the result I was hoping for. Need to repeat the same on Mac to better understand stuff.

wgpu-hal/src/dx12/mod.rs
emilk: …I also misread `.min(3)` as "at minimum 3" 🤦

emilk:
> On DX12 I'm hitting some odd bug making frame time expand to 1 full second, so I don't have test results from that so far.

Sounds like you might be hitting `const FRAME_TIMEOUT_MS: u32 = 1000;`

Wumpf:
Same behavior on Metal.
Thinking about this some more this starts making more sense to me. Already updated docs accordingly again. I suppose setting the latency to 1 still makes sense for fast rendering ui intense applications, so exposing this is still good. But use of it definitely requires some special care to not completely starve the gpu.

The thing I don't get yet is how one `CAMetalLayer` can do a wait on present. #2869 mentioned one can but I didn't find out how.

cwfitzgerald (approved):
Lets party!

PR #4871, opened 2023-12-13
Speed up Naga's `cargo xtask validate wgsl` from 12s to 0.8s

jimblandy:
Implement a `--bulk-validate` option in the Naga CLI, and use it in Naga's `cargo xtask validate wgsl` to validate all WGSL files in a single execution. This reduces real time from 12s to 0.8s.

**Checklist**

- [X] Run `cargo fmt`.
- [X] Run `cargo clippy`.
- [X] Run `cargo xtask test` to run tests.
- [x] Add change to `CHANGELOG.md`. See simple instructions inside file.

PR #4902, opened 2023-12-19
[naga xtask] Run validation jobs in parallel, using jobserver.

jimblandy:
Run all snapshot validation tasks in parallel, using the [jobserver](https://crates.io/crates/jobserver) crate to limit concurrency.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/naga/xtask/src/main.rs
+++ b/naga/xtask/src/main.rs
@@ -13,6 +13,7 @@ use crate::{
 mod cli;
 mod fs;
 mod glob;
+mod jobserver;
 mod path;
 mod process;
 mod result;
@@ -24,6 +25,8 @@ fn main() -> ExitCode {
         .parse_default_env()
         .format_indent(Some(0))
         .init();
+
+    jobserver::init();
 
     let args = Args::parse();
 
--- a/naga/xtask/src/main.rs
+++ b/naga/xtask/src/main.rs
@@ -1,19 +1,13 @@
-use std::{
-    io::{BufRead, BufReader},
-    path::Path,
-    process::{ExitCode, Stdio},
-};
+use std::process::ExitCode;
 
-use anyhow::{bail, Context};
+use anyhow::Context;
 use cli::Args;
 
 use crate::{
-    cli::{Subcommand, ValidateHlslCommand, ValidateSubcommand},
-    fs::{open_file, remove_dir_all},
-    glob::visit_files,
+    cli::Subcommand,
+    fs::remove_dir_all,
     path::join_path,
     process::{which, EasyCommand},
-    result::{ErrorStatus, LogIfError},
 };
 
 mod cli;
@@ -22,6 +16,7 @@ mod glob;
 mod path;
 mod process;
 mod result;
+mod validate;
 
 fn main() -> ExitCode {
     env_logger::builder()
@@ -42,8 +37,6 @@ fn main() -> ExitCode {
 }
 
 fn run(args: Args) -> anyhow::Result<()> {
-    let snapshots_base_out = join_path(["tests", "out"]);
-
     let Args { subcommand } = args;
 
     assert!(which("cargo").is_ok());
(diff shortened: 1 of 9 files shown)


jimblandy:
Combined with #4871, this reduces the time required to validate the Naga snapshot outputs from 21s to 0.6s on my Linux build machine (omitting Metal and HLSL FXC validation, since I can't run those on Linux).

jimblandy:
These commits can be landed without squashing. They should be a bit easier to review individually.

naga/xtask/src/validate.rs
```diff
@@ -9,86 +9,127 @@ use anyhow::{bail, Context};
...
+    for job in jobs {
+        if let Err(error) = job() {
+            all_good = false;
+            eprintln!("{:#}", error);
+        }
+    }
+
+    if !all_good {
+        bail!("failed to validate one or more files, see above output for more details")
+    }
```
ErichDonGubler: nitpick: `ensure!(all_good, …)` is a more concise way to do what you want here.

naga/xtask/src/validate.rs
```diff
@@ -9,86 +9,127 @@ use anyhow::{bail, Context};
...
 
 pub(crate) fn validate(cmd: ValidateSubcommand) -> anyhow::Result<()> {
+    let mut jobs = vec![];
+    collect_validation_jobs(&mut jobs, cmd)?;
+
+    let mut all_good = true;
+    for job in jobs {
+        if let Err(error) = job() {
+            all_good = false;
+            eprintln!("{:#}", error);
```
ErichDonGubler: question: Is there a strong reason here to prefer `eprintln!(…)` over `log::error!(…)`?

naga/xtask/src/validate.rs
```diff
@@ -9,86 +9,127 @@ use anyhow::{bail, Context};
...
-    result::{ErrorStatus, LogIfError},
 };
 
 fn ack_visiting(path: &Path) {
     log::info!("Validating {}", path.display());
 }
 
 pub(crate) fn validate(cmd: ValidateSubcommand) -> anyhow::Result<()> {
+    let mut jobs = vec![];
+    collect_validation_jobs(&mut jobs, cmd)?;
```
ErichDonGubler, line 20: ~~question: I feel like it'd more clearly communicate intent to have `collect_validation_jobs` to return a `Vec<…>` by-value, rather than mutating a `&mut Vec<…>`.~~

Ah, it's nice to keep a single collection to use between all backends. I feel fine with this!

naga/xtask/src/validate.rs
```diff
@@ -13,10 +13,6 @@ use crate::{
     process::{which, EasyCommand},
 };
 
-fn ack_visiting(path: &Path) {
-    log::info!("Validating {}", path.display());
-}
-
```
ErichDonGubler: question: Would it be interesting to restore this messaging? I feel like it's certainly possible, since we're using a thread pool and no child processes.

naga/xtask/src/validate.rs
ErichDonGubler: This question also applies to some other places, too.

naga/xtask/src/main.rs
```diff
@@ -25,6 +26,8 @@ fn main() -> ExitCode {
         .format_indent(Some(0))
         .init();
 
+    jobserver::init();
+
```
ErichDonGubler, line 30: question: Should this init. of the job server not be deferred until we know the `validate` subcommand is actually going to be used? Otherwise, this wastes resources speculatively, right?

In fact, why do we want to have a `static` job server at all? Is there a vision of integrating more subcommands and/or child processes with the job server than can be captured within the model of local scope? If not, I think we might be able to simplify code significantly when handling the `validate` subcommand with a different abstraction.[^1]

[^1]: It's pretty concise to initialize a thread pool with [`rayon::ThreadPool::install`](https://docs.rs/rayon/1.8.0/rayon/struct.ThreadPool.html#method.install) and [`rayon::iter::ParallelIterator`](https://docs.rs/rayon/1.8.0/rayon/iter/trait.ParallelIterator.html), for instance.

naga/xtask/src/validate.rs
```diff
@@ -13,15 +13,28 @@ use crate::{
...
+                tx_results.send(job()).unwrap();
+            })?;
+        }
+        Ok(())
+    });
+
     let mut all_good = true;
-    for job in jobs {
-        if let Err(error) = job() {
+    for result in rx_results {
```
ErichDonGubler, line 45: issue: We don't maintain a strong expectation of "sent X jobs, must get X results back" with this `Iterator` API usage over a single channel. We also make no attempt to `join` on threads that get spun up. Because of this, I'm concerned that this code will miss gaps in the result pipeline that may be introduced by panicking threads. This would be _weird_, admittedly, but it would also a weak link in `xtask`'s ability to report an issue.[^1]

Two directions I can see from here:

1. Accept this limitation, assume things won't break, and understand that we will be handling panicking issues in job threads reactively. AFAICT, we could only introduce problems here _per validation backend_ (which is a small factor for us), and we update our logic (once working) infrequently. It still makes me nervous with the potential for bug infiltration.
2. Iterate on this code in a separate PR, and do something like create a sender and receiver pair per job we submit.[^2] This way, we can force ourselves to acknowledge missing values as errors, too.

[^1]: Given the scale of testing that we're operating at, I don't think that manual inspection of `xtask`'s `stderr` output (which AIUI would record an issue) is likely to happen without a clear signal from CI or the community that something is wrong in the first place.
[^2]: We might pivot to a single-value channel API like [`postage::oneshot`](https://docs.rs/postage/0.5.0/postage/oneshot/index.html), to keep things cleaner.

naga/xtask/src/validate.rs
```diff
@@ -126,6 +126,24 @@ fn collect_validation_jobs(jobs: &mut Vec<Job>, cmd: ValidateSubcommand) -> anyh
...
+            collect_validation_jobs(jobs, ValidateSubcommand::Dot)?;
+            collect_validation_jobs(jobs, ValidateSubcommand::Wgsl)?;
+
+            // The DXC compiler can be built and run on any platform.
+            collect_validation_jobs(jobs, ValidateSubcommand::Hlsl(ValidateHlslCommand::Dxc))?;
+
+            // The FXC compiler is only available on Windows.
+            #[cfg(windows)]
+            collect_validation_jobs(jobs, ValidateSubcommand::Hlsl(ValidateHlslCommand::Fxc))?;
+        }
```
ErichDonGubler: issue(non-blocking): I think it'd be better to iterate over a `derive`d `Iterator` for the full set of back end variants that we support. This will force anyone who adds a new back end to be tested to confront adding this at compile-time. This code would become a `match` over all things instead; `#[cfg(…)]` attributes can otherwise be used in the same way.

ErichDonGubler (changes requested):
This mostly LGTM. There's enough misc. issues that I'd like to force their discussion before actually approving, tho.

jimblandy:
Rebased on current trunk; review comments not addressed yet.

naga/xtask/src/validate.rs
jimblandy: Since we do have an indication of progress, and we do provide full context for jobs that fail, I think it's better to be silent when all its going well. I am not a fan of programs that fill my scrollback with "all is well".

naga/xtask/src/main.rs
jimblandy: The [jobserver docs](https://docs.rs/jobserver/latest/jobserver/struct.Client.html#method.from_env_ext) say:

> It’s recommended to call \[the jobserver initialization function] very early in the lifetime of a program before any other file descriptors are opened. That way you can make sure to take ownership properly of the file descriptors passed down, if any.

That's why we have the call to `init` at the top of main.

The jobserver may take ownership of file descriptors passed down from a parent process, so you're only supposed to create one `Client` per process. For this, the `static` approach makes the most sense.

The jobserver is simply a pipe (on Unix) or a semaphore (on Windows). It's not expensive.

There are certainly plenty of other ways to have a thread pool of an appropriate size run the jobs, many of which would be more Rustic than jobserver, which has the feel of a 1980's Unix hack (which is what it is). The advantage of jobserver is that it will help Naga's `cargo xtask` cooperate properly when used as part of some larger build process, coordinated by `make` or another tool that uses the jobserver protocol. I don't know of anyone who's doing that right now, but when the question is "I want to run as many automation jobs concurrently as possible", it seems like using the defacto standard for whole-task coordination is a good thing.

naga/xtask/src/validate.rs
jimblandy: Capturing panics is an important thing to get right.

naga/xtask/src/validate.rs
jimblandy: This doesn't seem worthwhile to me. The set of backends changes extremely rarely. I think a comment on the enum should suffice. Even granting that mistakes are possible (I double-listed `Spirv` there), they're easy to fix.

This is an area where I think you and I calculate differently. It looks to me like your process is, "I want an iterator over an enum. There's a procmacro crate for that." In my view, there's a small cost to bringing in dependencies like `enum_iterator`: because they are not standard Rust, they're going to be outside some contributors' experience, and now we've given them an additional participant in the situation that they need to go understand when diagnosing problems. Sticking to standard Rust greatly increases the changes that contributors already know what they're looking at (and makes it more valuable for them to learn it if they don't, because they're more likely to see it elsewhere). Obviously, many third-party crates contribute enough that this cost is worthwhile; this is why Naga has sixteen dependencies. But for this, I feel that the likelihoods and costs don't add up.

naga/xtask/src/validate.rs
jimblandy: The new version uses `catch_unwind` while running the job, leaving the `unwrap` after the channel send outside the `catch`, which seems correct. Let me know what you think.

naga/xtask/src/validate.rs
jimblandy: In the final commit I could only find this one use of `eprintln`. Are there others?

jimblandy:
@ErichDonGubler Thanks for the review! I believe all comments have been resolved one way or another. Please take a look.

naga/xtask/src/validate.rs
ErichDonGubler: > This is an area where I think you and I calculate differently. …
>
> … Sticking to standard Rust greatly increases the changes that contributors already know what they're looking at (and makes it more valuable for them to learn it if they don't, because they're more likely to see it elsewhere).

You're right, I actually disagree with this. I suspect this might be a difference in perception of cost as somebody who's pretty accustomed to hundreds of deps. and constant onboarding onto new crates. 😅 I consider the cost of bringing in a small crate like `strum` negligible, in part because I've worked with it quite a bit before.  I consider it worthwhile, even, to introduce other developers to it. It's that good and useful.

However, `strum` does depend on `syn` and `quote`, which, now that you mention it, IMO would go against the current grain of rapid compilation of dependencies for this `xtask` crate. I would accept rejection of `strum` based on that alone. So, cool, let's not use `strum`. However…

I think rejecting the idea to use an `enum` is a separate matter from the idea of how to implement it. A custom `Iterator` implementation for `ValidateSubcommand` is not hard with "standard" Rust, so lemme just do it right here:

```rust
impl ValidateSubcommand {
    pub(crate) fn all() -> impl Iterator<Item = Self> { // using Rust 1.74.0 or higher; see <https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html#projections-in-opaque-return-types>
        [
			Self::Spirv,
			Self::Metal,
			Self::Glsl,
			Self::Dot,
			Self::Wgsl,
			Self::Hlsl(ValidateHlslCommand::Dxc),
			Self::Hlsl(ValidateHlslCommand::Fxc),
		].into_iter()
    }
}
```

…and I'll leave it to you separately accept or reject the idea itself.

naga/xtask/src/main.rs
ErichDonGubler: If there's a suspicion that `naga xtask` may need to integrate with a concrete job server later, then I think this is fine. Doing so speculatively makes me uncomfortable (because simpler code plz; this has been your schtick for perf.), but I'd consider it your decision to make.

I consider this conversation resolved, since I feel I've delivered the point I wanted to make.

naga/xtask/src/validate.rs
ErichDonGubler: > Since we do have an indication of progress

Admittedly, I'd forgotten to consider your usage of `indicatif` here. So, for local usage with a human, LGTM!

I think having `log` output would be preferable for CI, since AFAIK there's not going to be an indication of progress there. I definitely don't want overly verbose output for typical human usage, though, so perhaps we can replace the default `Info`-level `env_logger` initialization with `env_logger::init()`, which would make this `INFO`-level log silent by default instead?

I'll let you decide whether that's something to handle right now, tho. For the purposes of review, I consider this conversation resolved.

naga/xtask/src/validate.rs
ErichDonGubler: I see no others, and I don't think taking the time to track down what I was referring to is worth it here. LGTM now!

naga/xtask/src/validate.rs
ErichDonGubler: LGTM!

naga/xtask/src/validate.rs
ErichDonGubler: Sticking to my assertion that this is not blocking, I'll marked this conversation as resolved, and let @jimblandy decide how to follow up independent of merging.

ErichDonGubler (approved):
@jimblandy:

> @ErichDonGubler Thanks for the review! I believe all comments have been resolved one way or another. Please take a look.

Look taken! I've got some continued discussion in the threads here, but none of it is blocking IMO. Ship it!

naga/xtask/src/validate.rs
ErichDonGubler: Made this diff., just to make sure I could compile stuff. Oops, I implemented it. 😆

<details><summary>Diff</summary>

```diff
diff --git a/naga/xtask/src/cli.rs b/naga/xtask/src/cli.rs
index cd510124c..d41b86ba2 100644
--- a/naga/xtask/src/cli.rs
+++ b/naga/xtask/src/cli.rs
@@ -127,6 +127,19 @@ impl ValidateSubcommand {
             }
         }
     }
+
+    pub(crate) fn all() -> impl Iterator<Item = Self> {
+        [
+            Self::Spirv,
+            Self::Metal,
+            Self::Glsl,
+            Self::Dot,
+            Self::Wgsl,
+            Self::Hlsl(ValidateHlslCommand::Dxc),
+            Self::Hlsl(ValidateHlslCommand::Fxc),
+        ]
+        .into_iter()
+    }
 }
 
 #[derive(Debug)]
diff --git a/naga/xtask/src/validate.rs b/naga/xtask/src/validate.rs
index bed71b05d..394b7b00d 100644
--- a/naga/xtask/src/validate.rs
+++ b/naga/xtask/src/validate.rs
@@ -144,24 +144,27 @@ fn collect_validation_jobs(jobs: &mut Vec<Job>, cmd: ValidateSubcommand) -> anyh
             });
         }
         ValidateSubcommand::All => {
-            collect_validation_jobs(jobs, ValidateSubcommand::Wgsl)?;
-            collect_validation_jobs(jobs, ValidateSubcommand::Spirv)?;
-            collect_validation_jobs(jobs, ValidateSubcommand::Glsl)?;
-            collect_validation_jobs(jobs, ValidateSubcommand::Dot)?;
-
-            #[cfg(any(target_os = "macos", target_os = "ios"))]
-            collect_validation_jobs(jobs, ValidateSubcommand::Metal)?;
-
-            // The FXC compiler is only available on Windows.
-            //
-            // The DXC compiler can be built and run on any platform,
-            // but they don't make Linux releases and it's not clear
-            // what Git commit actually works on Linux, so restrict
-            // that to Windows as well.
-            #[cfg(windows)]
-            {
-                collect_validation_jobs(jobs, ValidateSubcommand::Hlsl(ValidateHlslCommand::Dxc))?;
-                collect_validation_jobs(jobs, ValidateSubcommand::Hlsl(ValidateHlslCommand::Fxc))?;
+            for subcmd in ValidateSubcommand::all() {
+                let should_validate = match &subcmd {
+                    ValidateSubcommand::Wgsl
+                    | ValidateSubcommand::Spirv
+                    | ValidateSubcommand::Glsl
+                    | ValidateSubcommand::Dot => true,
+                    ValidateSubcommand::Metal => cfg!(any(target_os = "macos", target_os = "ios")),
+                    // The FXC compiler is only available on Windows.
+                    //
+                    // The DXC compiler can be built and run on any platform,
+                    // but they don't make Linux releases and it's not clear
+                    // what Git commit actually works on Linux, so restrict
+                    // that to Windows as well.
+                    ValidateSubcommand::Hlsl(
+                        ValidateHlslCommand::Dxc | ValidateHlslCommand::Fxc,
+                    ) => cfg!(os = "windows"),
+                    ValidateSubcommand::All => continue,
+                };
+                if should_validate {
+                    collect_validation_jobs(jobs, subcmd)?;
+                }
             }
         }
     };
```

</details>

jimblandy: I've added this as an additional commit.

naga/xtask/src/validate.rs
jimblandy: Dropping the default logging level and then bumping it up for CI sounds like a good idea. But I think it could be handled in a separate PR.

jimblandy: Filed as #4938.

PR #4918, opened 2023-12-21
Make wgpu_render_pass_set_bind_group clamp the number of dynamic offsets within limits.

bradwerth:
This similarly updates wgpu_compute_pass_set_bind_group. Neither function throws an error when clamping. It's not clear what mechanism there is for error reporting at this entry point.

**Connections**
N/A

**Description**
User agents calling wgpu_compute_pass_set_bind_group directly can provide an arbitrary number of dynamic offsets without passing validation.

**Testing**
Untested. Any test using something like `cpass.set_bind_group` is stopped by validation.

**Checklist**

- [X] Run `cargo fmt`.


nical (approved):

wgpu-core/src/command/compute.rs
```diff
@@ -919,9 +919,17 @@ pub mod compute_ffi {
             return;
         }
 
+        // Clamp the number of dynamic offsets to the maximum value that the
```
cwfitzgerald, line 925: For validation we should include the original length, so that in the body of the pass we can check to make sure the original_length == clamped length, and error if they don't match.

teoxoy (commented):
Briefly looking at the code around `num_dynamic_offsets` it looks to me like it would make more sense to just change its data type to be `usize` since every time we set it we cast it to `u8` and every time we read it we cast it back to `usize`.

cwfitzgerald (changes requested):
For bookkeeping, going to request changes

nical:
Addressing this differently in #5026

PR #4886, opened 2023-12-17
Document wgpu & wgpu-core features

Wumpf:
**Connections**
(none that I'm aware of)

**Description**
Documents our feature flags for wgpu & wgpu-core. Focused mainly on wgpu since this is where most users enter. Uses `document-features` to spill out the whole thing into the api reference (see screenshot).
Re-arranged feature flags for the purpose of structuring

Couldn't help myself and sorted a few crate deps as well alphabetically ;-).

**Testing**
image

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`. If applicable, add:
  - [ ] `--target wasm32-unknown-unknown`
  - [ ] `--target wasm32-unknown-emscripten`
- [ ] Run `cargo xtask test` to run tests.
- [x] Add change to `CHANGELOG.md`. See simple instructions inside file.

PR #4924, opened 2023-12-22
Inline `document-features` usage, remove dep.

ErichDonGubler:
**Connections**

**Description**

Re-implements https://github.com/gfx-rs/wgpu/pull/4886 (CC @Wumpf) without the `document-features` crate, which has issues integrating into Firefox builds after being `cargo vendor`ed into its repository. This issue is being tracked against https://github.com/slint-ui/document-features/issues/20. Once resolved, I expect that we will want to revert this PR in its entirety, since `document-features` is still a good addition to `wgpu`'s documentation story.

Internally, consensus has already been achieved with @cwfitzgerald for this change. Firefox's ability to build unfortunately take priority over this particular convenience. Hopefully, we won't have to compromise shortly!

**Testing**

I tested this by ensuring that the HTML output of our existing  `document_features::document_features!(…)` usage was exactly the same. There should be exactly zero regressions in the current state of documentation for users. For maintainers, I have added a disclaimer that one needs to keep changes in sync. with the relevant `Cargo.toml` manifests.

**Checklist**

- [x] Run `cargo fmt`.
- ~~Add change to `CHANGELOG.md`. See simple instructions inside file.~~ This isn't a user-visible change, so a `CHANGLEOG` entry doesn't seem necessary here.


wgpu-core/src/lib.rs
```diff
@@ -3,8 +3,36 @@
 //! into other language-specific user-friendly libraries.
 //!
 //! ## Feature flags
-#![doc = document_features::document_features!()]
+// NOTE: the above should be kept in sync. with `Cargo.toml`!
```
jimblandy: Should this be "the below"?

jimblandy (approved):
Thank you!

wgpu-core/src/lib.rs
ErichDonGubler: oderp, yes, that should be "below". 😅

Fixed with 3ebec9c13; I amended "the above" to be "feature docs. below".

PR #4935, opened 2023-12-26
Add `cfg_aliases` to `wgpu`

daxpedda:
As discussed in https://github.com/gfx-rs/wgpu/pull/4815#pullrequestreview-1780357351.

### DIFF
--- a/wgpu/build.rs
+++ b/wgpu/build.rs
@@ -1,0 +1,14 @@
+fn main() {
+    cfg_aliases::cfg_aliases! {
+        native: { not(target_arch = "wasm32") },
+        webgl: { all(target_arch = "wasm32", not(target_os = "emscripten"), feature = "webgl") },
+        webgpu: { all(target_arch = "wasm32", not(target_os = "emscripten"), not(feature = "webgl")) },
+        emscripten: { all(target_arch = "wasm32", target_os = "emscripten") },
+        send_sync: { any(
+            not(target_arch = "wasm32"),
+            all(feature = "fragile-send-sync-non-atomic-wasm", not(target_feature = "atomics"))
+        ) },
+        dx12: { all(target_os = "windows", feature = "dx12") },
+        metal: { all(any(target_os = "ios", target_os = "macos"), feature = "metal") }
+    }
+}
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -97,26 +97,18 @@ pub use wgt::{
     QUERY_RESOLVE_BUFFER_ALIGNMENT, QUERY_SET_MAX_QUERIES, QUERY_SIZE, VERTEX_STRIDE_ALIGNMENT,
 };
 
-#[cfg(any(
-    not(target_arch = "wasm32"),
-    feature = "webgl",
-    target_os = "emscripten"
-))]
+#[cfg(not(webgpu))]
 #[doc(hidden)]
 pub use ::hal;
 #[cfg(feature = "naga")]
 pub use ::naga;
-#[cfg(any(
-    not(target_arch = "wasm32"),
-    feature = "webgl",
-    target_os = "emscripten"
-))]
+#[cfg(not(webgpu))]
 #[doc(hidden)]
 pub use ::wgc as core;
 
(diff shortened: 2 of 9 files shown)


wgpu/build.rs
```diff
@@ -0,0 +1,14 @@
...
+        native: { not(target_arch = "wasm32") },
+        webgl: { all(target_arch = "wasm32", not(target_os = "emscripten"), feature = "webgl") },
+        webgpu: { all(target_arch = "wasm32", not(target_os = "emscripten"), not(feature = "webgl")) },
+        emscripten: { all(target_arch = "wasm32", target_os = "emscripten") },
+        send_sync: { any(
+            not(target_arch = "wasm32"),
+            all(feature = "fragile-send-sync-non-atomic-wasm", not(target_feature = "atomics"))
+        ) },
+        dx12: { all(target_os = "windows", feature = "dx12") },
+        metal: { all(any(target_os = "ios", target_os = "macos"), feature = "metal") }
```
daxpedda, line 12: I would recommend to give those nicer names as they are visible in the documentation:
image
We could use "WebGPU", "DX12" and "Metal". This has the drawback that seeing capitalized letters in `cfg` guards is a bit odd.

wgpu/src/util/init.rs
```diff
@@ -2,10 +2,10 @@ use wgt::{Backends, PowerPreference, RequestAdapterOptions};
 
 use crate::{Adapter, Instance, Surface};
 
-#[cfg(any(not(target_arch = "wasm32"), feature = "wgc"))]
+#[cfg(not(webgpu))]
```
daxpedda, line 5: Rust isn't able to figure out that these appear regardless, probably because one is a typedef and the other a function.
image
How about using `#[cfg_attr(docsrs, doc(cfg(all())))]` here?

wgpu/src/lib.rs
```diff
@@ -5428,55 +5035,25 @@ pub enum Error {
...
     OutOfMemory {
         /// Lower level source of the error.
-        #[cfg(any(
-            not(target_arch = "wasm32"),
-            all(
-                feature = "fragile-send-sync-non-atomic-wasm",
-                not(target_feature = "atomics")
-            )
-        ))]
+        #[cfg(send_sync)]
```
daxpedda, line 5038: These are also visible, which from Rustdocs perspective makes sense:
image
Would also recommend to remove that.

wgpu/src/util/init.rs
Wumpf: works for me if it does the trick 👍

wgpu/build.rs
Wumpf: hmm yeah I'm a bit torn, but as you say capitalized cfgs are very unusual, so I'd stick with that regardless. This is still _sooo_ much better anyways 😄

Wumpf (approved):
oh my I love this so much So clean 🤩

wgpu/src/lib.rs
Wumpf: I know of at least one project that uses this, it can be super useful if you also depend on wgpu-core. So removing (or making private) isn't an option.
These small wrinkles in the generated docs are super annoying but at least most of it happens only on fairly advanced stuff, so I'm not _too_ worried

daxpedda: Sorry, I was using the word "remove" in context of what I was proposing in https://github.com/gfx-rs/wgpu/pull/4935#discussion_r1436659586.

So my suggestion is not to remove the variants, it's to remove the "Available on `send_sync` only." documentation. Which isn't correct, it's available on both, just not with the `Send` bound on Wasm.

Wumpf: ah, makes more sense 😄

daxpedda: Done.

wgpu/src/util/init.rs
daxpedda: Done.

daxpedda:
No changes, only rebased.

daxpedda:
I think CI failure is unrelated?

PR #4940, opened 2023-12-28
Align `wgpu_types::CompositeAlphaMode` serde serializations to spec

littledivy:
**Checklist**

- [x] Run `cargo fmt`.

**Overview**

Variants should be serialized as lowercase instead of camelCase to match spec. https://gpuweb.github.io/gpuweb/#gpucanvasalphamode


crowlKats (commented):
This is something needed for Deno, since the serialization does not match to spec.

cwfitzgerald (approved):

PR #4944, opened 2023-12-28
Fix xtask wasm-bindgen install

rukai:
The xtask assumes that a crates binary name is the same as the crate name, which is not neccessarily true.
This resulted in it reccomending I run the command `cargo install wasm-bindgen` which fails as wasm-bindgen is not a binary but a library.
What it should have reccomended was `cargo install wasm-bindgen-cli`.
This PR fixes that.


cwfitzgerald (approved):
Big derp, thanks!

Issue #4919, opened 2023-12-21
Vulkan sometimes present while unfinished


valaphee:
**Description**
Sometimes, during the configuration of the surface, at start or when resizing, the frames get presented even though they are still written to, which leads to flickering.

The frame rate also increases from 1200 FPS to 2000 FPS which lead me to this conclusion.

**Repro steps**
Run wgpu water example in release mode with Vulkan backend.

**Expected vs observed behavior**
The rendering finishes before presenting, even though shouldn't there be triple buffering?

**Extra materials**
https://github.com/gfx-rs/wgpu/assets/32491319/ca41f826-d735-43ee-9d72-cd90935930b1
(epilepsy warning)

**Platform**
Latest commit of wgpu `79e51692`, Windows 11 Build 22631, AMD Radeon RX6900 XT, with latest drivers 23.11.1

valaphee:
It's most likely not fixable in wgpu, as I couldn't reproduce it on my AMD iGPU, and Nvidia GPU. There is also no error reported when enabling validation layers.

It might be also solved by using semaphores instead of fences for presenting, as the fences sometimes just get ignored
https://github.com/gfx-rs/wgpu/issues/4775

ids1024:
> The rendering finishes before presenting, even though shouldn't there be triple buffering?

https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPresentModeKHR.html specifies which presentation modes can result in tearing and which cannot. These map to https://docs.rs/wgpu/0.18.0/wgpu/enum.PresentMode.html, which also mentions which can have tearing.

Looks like the examples are all using `PresentMode::Fifo`, so I believe that means this shouldn't be happening. 

> It's most likely not fixable in wgpu, as I couldn't reproduce it on my AMD iGPU, and Nvidia GPU. There is also no error reported when enabling validation layers.

I don't think validation layers catch everything, and incorrect use of Vulkan may work on some drivers but not others. Though driver bugs are definitely a possibility too.

valaphee:
I assume its also a driver bug, or some race condition when creating the fence, as most of the times the fence just works fine, and its also only occuring in release mode, not debug.

But its definitely not tearing because the frame is not finished and already presented, which is the opposite of tearing.

cwfitzgerald:
Wait does this only happen on the water example?

valaphee:
[...]

Issue #4775, opened 2023-11-25
Applications using `wgpu` hang forever on bleeding edge Linux with Nvidia drivers 545.29.06 on GNOME / Wayland


udoprog:
**Repro steps**
Running anything which tries to use wgpu Vulkan, like:

```
cd examples && cargo run cube
```

The window starts and renders at least one frame, but becomes completely non-interactive (windows can't be interacted with or moved) and you receive a "hanging" prompt from GNOME:

![image](https://github.com/gfx-rs/wgpu/assets/111092/433b9677-339e-433a-abb8-88b6c3e6bd67)

Note that I think this might legitimately be a platform issue, however:

* I am unable to reproduce it with both `vkcube` (X11) and `vkcube-wayland` which reports and runs (see below).
* And [winit examples](https://github.com/rust-windowing/winit/tree/master/examples) run without issues.

```
> sudo dnf install vulkan-tools
> vkcube-wayland
Selected GPU 0: NVIDIA GeForce RTX 2080 Ti, type: DiscreteGpu
```

[Screencast from 2023-11-25 18-23-21.webm](https://github.com/gfx-rs/wgpu/assets/111092/21045238-08d3-4e4e-858a-e7c2454a00f4)

So `wgpu` is currently the lowest level of abstraction I've chased down.

**Platform**

Log output from running the example:

```
wgpu_core::instance] Adapter Vulkan AdapterInfo { name: "NVIDIA GeForce RTX 2080 Ti", vendor: 4318, device: 7687, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "545.29.06", backend: Vulkan }
```

`uname -r`:

```
6.7.0-0.rc2.20231122gitc2d5304e6c64.23.fc40.x86_64
```

udoprog:
This is probably a duplicate of #4689, but I'm just gonna add what I've found so far here:

This is where we hang forever:

https://github.com/gfx-rs/wgpu/blob/ebcfd25b58a2c4f3b442031d22b510576c1b8155/wgpu-hal/src/vulkan/instance.rs#L960

Out of curiosity, I added some instrumentation:

```rust
let fences = &[sc.fence];

unsafe {
    let status = sc.device.raw.get_fence_status(sc.fence)
        .map_err(crate::DeviceError::from)?;
    println!("wait: {}", status);
    sc.device.raw.wait_for_fences(fences, true, !0)
        .map_err(crate::DeviceError::from)?;
[...]

Issue #4689, opened 2023-11-15
Nvidia Driver 545 Breaks Skybox Example on Linux Wayland


krakow10:
Note that this is a duplicate of #4627.  I was not able to comment or reopen the issue, so I've created a new issue.  I built and tested a6aa3cd30d3a3c0efde5d81fb5435d1c281a0ab1 and the issue has not been fixed.  Works great on NVIDIA 535 driver, broken on NVIDIA 545 driver.

**Description**
The example only renders one frame.  This also affects my application which is based on the skybox example.  This is probably an Nvidia driver bug because this driver version reworks many Wayland things.

**Repro steps**
`cargo run --bin skybox`

**Expected vs observed behavior**
Many frames should be rendered per second, but only one is observed once.

**Extra materials**
None

**Platform**
OS: Archlinux x86-64
Nvidia Driver: 545.29.02
GPU: 3080 Ti
DE: Hyprland

teoxoy:
545.29.02 seems to still be on the "New Feature Branch"
535.129.03 is the latest on the "Production Branch"

Marking as a driver-bug for now.

teoxoy:
Is this happening only on Vulkan?

krakow10:
Testing this commit on 8 configurations: 30d67a329a2906a0a54704a8e8297b39f929a987

Nvidia 535.113.01, Vulkan, wayland (Hyprland):
`cargo run --bin skybox`
works ✅

Nvidia 535.113.01, Vulkan, x11 (xfce):
`cargo run --bin skybox`
works ✅

Nvidia 545.29.02, Vulkan, wayland (Hyprland):
`cargo run --bin skybox`
no errors, only one frame is drawn ❌

Nvidia 545.29.02, Vulkan, x11 (xfce):
`cargo run --bin skybox`
works ✅

I was not able to get OpenGL to run on any combination of 535, 545, x11, or wayland.
<details>
   <summary>Show big errors</summary>

   Nvidia 535.113.01, OpenGL, wayland (Hyprland):

   `WGPU_BACKEND=gl cargo run --bin skybox`

   ```
   [quat@quat-desktop wgpu]$ WGPU_BACKEND=gl cargo run --bin skybox
       Finished dev [unoptimized + debuginfo] target(s) in 0.12s
        Running `target/debug/skybox`
   [2023-11-22T05:08:46Z INFO  wgpu_example::framework] Initializing wgpu...
[...]

PR #4946, opened 2023-12-31
Introduce semaphores and wait on submit

udoprog:
**Connections**
#4689, #4775, #4919

bevyengine/bevy#10792

**Description**
This is a WIP PR to gather feedback. I've yet to port all the examples or other backends than Vulkan to use semaphores and fill out the `SemaphoreDescriptor` fully.

This introduces user-facing semaphores and allows one to be used to synchronize when a surface texture becomes available during submission.

It was suggested in #4775 that the fence blocking might be a driver bug. I don't necessarily think that's the case. The fence used is only required to trigger at some point in the future when the surface texture becomes available. Vulkan on Android for example (I can't recall where I read this) seems to hold onto the surface texture until a new frame is available from a call to submit. In that scenario it's understandable why the fence never signals in user facing code prior to that.

This might imply that wait semaphores are the only "correct" way to perform this synchronization when using [`vkAcquireNextImageKHR`](https://stackoverflow.com/questions/52665787/vkacquirenextimagekhr-semantics-what-does-timing-out-mean). So it might be a good idea to make this a user facing feature which is what this PR does.

**Alternative designs**

It might be possible to keep track of the semaphores internally. I'm just not sure how that can be done right now while not limiting functionality.

**TODO**

* Implement changes to other backends.
* Fix examples to use semaphores.
* Should we support more than one semaphore during submit? One might use more than one acquired image in the pipeline.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ members = [
 exclude = []
 default-members = [
     "d3d12",
-    "examples/",
+    # "examples/",
     "naga-cli",
     "naga",
     "naga/fuzz",
--- a/examples/src/framework.rs
+++ b/examples/src/framework.rs
@@ -125,6 +125,7 @@ impl EventLoopWrapper {
 struct SurfaceWrapper {
     surface: Option<wgpu::Surface<'static>>,
     config: Option<wgpu::SurfaceConfiguration>,
+    image_available: Option<wgpu::Semaphore>,
 }
 
 impl SurfaceWrapper {
@@ -133,6 +134,7 @@ impl SurfaceWrapper {
         Self {
             surface: None,
             config: None,
+            image_available: None,
         }
     }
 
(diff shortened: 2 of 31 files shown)


cwfitzgerald:
Before you get too far down this path - we unfortunately can't accept this as it is. The semaphore/fence is a decision completely internal to wgpu-hal's vulkan backend, and it definitely doesn't need to go all the way out to the user.

You definitely can set up semaphores completely internally to the vulkan backend, I had a branch which did exactly that. It made no difference on all the drivers I tested, so it never went anywhere. I believe you need a semaphore per frame in the swapchain. Each submit waits on the previous frame's submission semaphore, and the acquire semaphore for that image. If you're interested in implementing this, let's chat on the matrix.

Fence is definitely _correct_, but it doesn't seem common, which could perhaps cause problems with drivers.

udoprog:
Thanks! And don't worry, this is a draft because I just wanted to garner feedback. I never expected it to be merged like this. Implementing this was a decent excuse to get a feel for how wgpu manages and tracks resources.

I'll poke you about the other parts on Matrix.

PR #4954, opened 2024-01-02
Put raw texture access behind snatch guards

nical:
**Connections**

Another step towards #4787

**Description**

Require that the snatch_guard be held to raw textures.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


teoxoy (approved):
Looks good!

PR #4955, opened 2024-01-02
(Don't merge) fiddle with wgpu-info

nical:
I can't reproduce the wgpu-info CI issue locally so i'm going to try a few things via PRs.


cwfitzgerald:
This is https://gitlab.freedesktop.org/mesa/mesa/-/issues/10371

PR #4958, opened 2024-01-02
Change examples page menu on smaller screens

Dinnerbone:
**Connections**
Crash discovered whilst trying to debug https://bugs.chromium.org/p/chromium/issues/detail?id=1510149

**Description**
This fixes the crash on chrome android because we didn't set initial scaling/sizing info, so the browse zoomed way out and made the canvas size hit the 4k limit on Chrome.

It also makes it kinda actually usable on mobile. :D

## 720p desktop (unchanged)
Screen Shot 2024-01-02 at 19 04 36

## iPad (default)
Screen Shot 2024-01-02 at 19 06 53

## iPad (expanded)
Screen Shot 2024-01-02 at 19 07 17

## Galaxy Note 20 (expanded)
Screen Shot 2024-01-02 at 19 07 30

**Testing**
Manually.

**Checklist**

- [ ] Run `cargo fmt`.
- [ ] Run `cargo clippy`. If applicable, add:
  - [ ] `--target wasm32-unknown-unknown`
  - [ ] `--target wasm32-unknown-emscripten`
- [ ] Run `cargo xtask test` to run tests.
- [x] Add change to `CHANGELOG.md`. See simple instructions inside file.

PR #4960, opened 2024-01-02
Fix incorrect ConfigureSurfaceError::TooLarge message

Dinnerbone:
**Connections**
Seen in error message before the fix by #4958

**Description**
Very very small change, a typo in the error message made it not actually show the value that was wrong.

Before:
>`Surface` width and height must be within the maximum supported texture size. Requested was (2572, height), maximum extent is 4096.
After:
>`Surface` width and height must be within the maximum supported texture size. Requested was (2572, 5110), maximum extent is 4096.

**Testing**
Manually. You can open chrome on android on our webgl demos before #4958, or just manually make a large surface.

**Checklist**

- [x] Run `cargo fmt`.


cwfitzgerald (approved):
Heh

PR #4964, opened 2024-01-03
Backport device limits fix to 0.17

cshenton-work:
**Description**
[This commit](https://github.com/gfx-rs/wgpu/commit/41efabbd886d09163b7e340caee6bca84eae968d) in 0.18 and later fixed an issue where limits passed to `request_device` would be dropped due to an upstream issue with `wasm_bindgen` and `web_sys`.

This PR backports that fix to 0.17 so that dependent packages can enjoy correct device initialisation behaviour on the web.

**Testing**

This has been tested alongside a modified copy of `bevy 0.12.0` (updated to depend on `wgpu 0.17.2`), to confirm that the device limits are now respected. For example device request properly fails on too large a `maxBufferSize` and no longer errors when buffers less than the requested `maxBufferSize`, but greater than the default max size, are created.

**Checklist**

- [X] Run `cargo fmt`.
  - [Upstream Failure in `noise`] `--target wasm32-unknown-emscripten`
- [unrecognized subcommand "test"] Run `cargo xtask test` to run tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,8 +50,8 @@ path = "./wgpu-hal"
 version = "0.17"
 
 [workspace.dependencies.naga]
-git = "https://github.com/gfx-rs/naga"
-rev = "bac2d82a430fbfcf100ee22b7c3bc12f3d593079"
+# git = "https://github.com/gfx-rs/naga"
+# rev = "bac2d82a430fbfcf100ee22b7c3bc12f3d593079"
 version = "0.13.0"
 
 [workspace.dependencies]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ Bottom level categories:
 #### Vulkan
 
 - Fix x11 hang while resizing on vulkan. @Azorlogh in [#4184](https://github.com/gfx-rs/wgpu/pull/4184).
+- Backport fix to ensure that limit requests and reporting is done correctly. (By @OptimisticPeach in [#4107](https://github.com/gfx-rs/wgpu/pull/4107)), @cshenton in  [#4964](https://github.com/gfx-rs/wgpu/pull/4964)
 
 ## v0.17.1 (2023-09-27)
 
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -50,8 +50,8 @@ path = "./wgpu-hal"
 version = "0.17"
 
 [workspace.dependencies.naga]
-git = "https://github.com/gfx-rs/naga"
-rev = "bac2d82a430fbfcf100ee22b7c3bc12f3d593079"
+# git = "https://github.com/gfx-rs/naga"
```
cwfitzgerald, line 53: This got accidentally committed

cwfitzgerald (changes requested):
Looks good after nit.

Please re-request my review after it's addressed!

Cargo.toml
cshenton-work: This was actually not an accident. Without removing this, there are naga version conflicts when building against other packages. Is there a reason this particular commit is targeted rather than the version on crates.io?

cwfitzgerald: The commit here is just the commit for the release of 0.17 - the sources are the same. When locally patching, you should be able to patch both wgpu and naga to the same place and it work fine. Something like:

```rust
[patch.crates-io]
wgpu = { git = "https://github.com/gfx-rs/wgpu.git", rev = "7b7dfa6" }
wgpu-core = { git = "https://github.com/gfx-rs/wgpu.git", rev = "7b7dfa6" }
wgpu-hal = { git = "https://github.com/gfx-rs/wgpu.git", rev = "7b7dfa6" }
wgpu-types = { git = "https://github.com/gfx-rs/wgpu.git", rev = "7b7dfa6" }
naga = { git = "https://github.com/gfx-rs/naga.git", rev = "bac2d82a430fbfcf100ee22b7c3bc12f3d593079" }
```

Wumpf:
@cshenton-work  do you still need this or are you on a newer version by now?

cwfitzgerald:
As this PR has gone stale, I'm going to close it - please refile it if you have hte need for this

PR #4967, opened 2024-01-03
vulkan: Replace fence with semaphore when acquiring surfaces

udoprog:
**Connections**
Different way of solving #4946 which uses internal wait semaphores.

#4689, #4775, #4919

**Description**
This removes the use of a fence in favor of internally using and keeping track of one wait semaphore per swapchain image.

It roughly uses the approach initially attempted at https://github.com/cwfitzgerald/wgpu/tree/vulkan-timing-fixes,

~~I've defined a generic set in `A::SubmitSurfaceTextureSet` which can be used to collect whatever information it needs from a surface texture, this is defined as a dummy implementation for backends which does not use it.~~

**Multiple calls to get_current_texture**

There is a tricky situation which arises if a user were to perform multiple calls to `get_current_texture`. The current implementation naively just rotates `surface_semaphores`. So if we call this function images.len() + 1 times in a row, we will exhaust the semaphores and they will be assigned to a random set of swapchain images by the presentation engine.

This wasn't a problem in #4946 since the user is responsible for ensuring that the number of pending get_current_texture matches the number of semaphores they have (e.g. exactly one per image), but here we have to deal with this scenario somehow.

I think the "correct" way to deal with this is to somehow free the semaphore after a successful call to `acquire_next_image` has returned a previously used index again. But I'm looking for input.

**Testing**
Using the vulkan backend for a project I'm working on and dealing with any issues that show up.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.
- [x] Run `cargo xtask test` to run tests (still doesn't work).

### DIFF
--- a/wgpu-core/src/device/queue.rs
+++ b/wgpu-core/src/device/queue.rs
@@ -22,7 +22,7 @@ use crate::{
     resource_log, track, FastHashMap, SubmissionIndex,
 };
 
-use hal::{CommandEncoder as _, Device as _, Queue as _};
+use hal::{CommandEncoder as _, Device as _, Queue as _, RawSet as _};
 use parking_lot::Mutex;
 
 use std::{
@@ -1115,6 +1115,12 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 .fetch_add(1, Ordering::Relaxed)
                 + 1;
             let mut active_executions = Vec::new();
+
+            // SAFETY: We're constructing this during the submission phase,
+            // where all resources it uses are guaranteed to outlive this
+            // short-lived set.
+            let mut submit_surface_textures = A::SubmitSurfaceTextureSet::new();
+
             let mut used_surface_textures = track::TextureUsageScope::new();
 
             let snatch_guard = device.snatchable_lock.read();
@@ -1217,8 +1223,19 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                                         return Err(QueueSubmitError::DestroyedTexture(id));
                                     }
                                     Some(TextureInner::Native { .. }) => false,
-                                    Some(TextureInner::Surface { ref has_work, .. }) => {
+                                    Some(TextureInner::Surface {
+                                        ref has_work,
+                                        ref raw,
+                                        ..
+                                    }) => {
                                         has_work.store(true, Ordering::Relaxed);
+
+                                        if let Some(raw) = raw {
+                                            unsafe {
+                                                submit_surface_textures.insert(raw);
+                                            }
+                                        }
+
                                         true
                                     }
                                 };
@@ -1409,8 +1426,19 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                             return Err(QueueSubmitError::DestroyedTexture(id));
                         }
                         Some(TextureInner::Native { .. }) => {}
-                        Some(TextureInner::Surface { ref has_work, .. }) => {
+                        Some(TextureInner::Surface {
+                            ref has_work,
+                            ref raw,
+                            ..
+                        }) => {
                             has_work.store(true, Ordering::Relaxed);
+
+                            if let Some(raw) = raw {
+                                unsafe {
+                                    submit_surface_textures.insert(raw);
+                                }
+                            }
+
                             unsafe {
                                 used_surface_textures
                                     .merge_single(texture, None, hal::TextureUses::PRESENT)
@@ -1449,12 +1477,13 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                         .flat_map(|pool_execution| pool_execution.cmd_buffers.iter()),
                 )
                 .collect::<Vec<_>>();
+
             unsafe {
                 queue
                     .raw
                     .as_ref()
                     .unwrap()
-                    .submit(&refs, Some((fence, submit_index)))
+                    .submit(&refs, &submit_surface_textures, Some((fence, submit_index)))
                     .map_err(DeviceError::from)?;
             }
 
--- a/wgpu-core/src/device/queue.rs
+++ b/wgpu-core/src/device/queue.rs
@@ -22,8 +22,9 @@ use crate::{
     resource_log, track, FastHashMap, SubmissionIndex,
 };
 
-use hal::{CommandEncoder as _, Device as _, Queue as _, RawSet as _};
+use hal::{CommandEncoder as _, Device as _, Queue as _};
 use parking_lot::Mutex;
+use smallvec::SmallVec;
 
 use std::{
     iter, mem, ptr,
@@ -1116,14 +1117,11 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 + 1;
             let mut active_executions = Vec::new();
 
-            // SAFETY: We're constructing this during the submission phase,
-            // where all resources it uses are guaranteed to outlive this
-            // short-lived set.
-            let mut submit_surface_textures = A::SubmitSurfaceTextureSet::new();
-
             let mut used_surface_textures = track::TextureUsageScope::new();
 
             let snatch_guard = device.snatchable_lock.read();
+
+            let mut submit_surface_textures_owned = SmallVec::<[_; 2]>::new();
 
             {
                 let mut command_buffer_guard = hub.command_buffers.write();
@@ -1230,10 +1228,8 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                                     }) => {
                                         has_work.store(true, Ordering::Relaxed);
 
-                                        if let Some(raw) = raw {
-                                            unsafe {
-                                                submit_surface_textures.insert(raw);
-                                            }
+                                        if raw.is_some() {
+                                            submit_surface_textures_owned.push(texture.clone());
                                         }
 
                                         true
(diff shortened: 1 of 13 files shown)


ids1024:
I think this should also potentially perform better in some circumstances, by not unnecessarily blocking the CPU on a fence and instead just marking this as a dependency of the queue submission. Right? So this seems better in general.

udoprog:
Looks like a legit validation failure, I'll try to suss it out before unmarking this as a draft:

[2024-01-04T08:20:08Z ERROR wgpu_test::expectations] Unexpected failure due to: ValidationError(Some("Validation Error: [ VUID-VkSubmitInfo-pSignalSemaphores-03242 ] Object 0: handle = 0xf56c9b0000000004, type = VK_OBJECT_TYPE_SEMAPHORE; Object 1: handle = 0x7f09e27c8fd0, type = VK_OBJECT_TYPE_QUEUE; | MessageID = 0xdb30ee87 | vkQueueSubmit(): pSubmits[0].pSignalSemaphores[0] signal value (0xffffffffffffffff) in VkQueue 0x7f09e27c8fd0[] must be greater than pending signal timeline semaphore VkSemaphore 0xf56c9b0000000004[] value (0xffffffffffffffff). The Vulkan spec states: For each element of pSignalSemaphores created with a VkSemaphoreType of VK_SEMAPHORE_TYPE_TIMELINE the corresponding element of VkTimelineSemaphoreSubmitInfo::pSignalSemaphoreValues must have a value greater than the current value of the semaphore when the semaphore signal operation is executed (https://vulkan.lunarg.com/doc/view/1.3.268.0/linux/1.3-extensions/vkspec.html#VUID-VkSubmitInfo-pSignalSemaphores-03242)"))

Seems like I wasn't careful when refactoring the timeline semaphores.

valaphee:
Doesn't fix https://github.com/gfx-rs/wgpu/issues/4919 but the water example runs 600 fps faster (~1500 fps and now with ~2100 fps)

udoprog:
That's both unfortunate that it doesn't address your problem but neat about the fps boost. Thanks for trying it out! Personally I can't compare since the unpatched wgpu with the fence doesn't work at all for me right now.

cwfitzgerald:
@udoprog Haven't had a chance to fully look over the code, but going to send it out to a few people for testing in various environments - could you just resolve the merge conflicts with trunk?

udoprog:
@cwfitzgerald All right, that should be it. Tell me if there's anything else.

wgpu-hal/src/vulkan/mod.rs
```diff
@@ -80,6 +82,23 @@ impl crate::Api for Api {
...
+}
+
+pub struct SubmitSurfaceTextureSet {
+    semaphores: ArrayVec<vk::Semaphore, 1>,
+}
+
+impl RawSet<SurfaceTexture> for SubmitSurfaceTextureSet {
+    fn new() -> Self {
+        Self {
+            semaphores: ArrayVec::new(),
```
cwfitzgerald: This doesn't seem correct - what if the user writes to multiple surfaces in a single submission?

wgpu-core/src/device/queue.rs
```diff
@@ -1133,6 +1133,12 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 .fetch_add(1, Ordering::Relaxed)
                 + 1;
             let mut active_executions = Vec::new();
+
+            // SAFETY: We're constructing this during the submission phase,
+            // where all resources it uses are guaranteed to outlive this
+            // short-lived set.
+            let mut submit_surface_textures = A::SubmitSurfaceTextureSet::new();
```
cwfitzgerald: I don't understand why insert needs to be unsafe, it's just a vector

cwfitzgerald (changes requested):

wgpu-hal/src/vulkan/mod.rs
udoprog: You're right, this is supposed to be a smallvec. Will fix!

wgpu-core/src/device/queue.rs
udoprog: Insertion is unsafe because it can own raw resources related to surfaces (e.g. the wait semaphore). For a variety of reasons this could mean that the resource is no longer valid when the collection goes out of scope and is destructed, so memory safety depends on it being correctly used.

Now, we could skirt this and claim that this is only used internally. I didn't want to do that by default since it is part of a public API.

I would note that you implemented it as a collection of references to the surface which ensures that, but that is harder to do now that surfaces are accessed under a lock.

cwfitzgerald:
Hey! Sorry forgot to follow up on this - sent this to a bunch of people and things look good! Everything seems to work as expected on all platforms. It either has equivalent performance, or significantly higher performance.

Going to do another proper review in the next day or so and we'll get this in!

wgpu-core/src/device/queue.rs
cwfitzgerald: > Insertion is unsafe because it can own raw resources related to surfaces (e.g. the wait semaphore).

I believe this is actually safe. The backends aren't acutally storing references, they're copying in handles which do nothing on drop.

The unsafe-ness comes from the `use` of the raw resources - in this case the queue.submit in which it is passed. Similarly this is why `ash::*Builder::build()` is safe even though it creates a bunch of raw pointers, it's the _use_ of those pointers which is unsafe, and the functions it is passed to are unsafe.

More generally, I think we should just pass a normal slice into `queue.submit` - removing the collection of surface textures seems not really worth the wgpu-hal complexity.

udoprog: > More generally, I think we should just pass a normal slice into queue.submit - removing the collection of surface textures seems not really worth the wgpu-hal complexity.

Resource tracking has changed quite a bit since your branch, this requires breaking up `pending_writes` (since we would borrow from it and it currently would need mutable access to [`pre_submit` here](https://github.com/gfx-rs/wgpu/pull/4967/files#diff-150156a37cf3627465ceb22096ed995ee26ae640007c421e726134bafd499dbeR1492)) and dealing with how texture handles are handed it here in `cmd_buf_trackers`:

```
error[E0597]: `texture` does not live long enough
    --> wgpu-core\src\device\queue.rs:1227:59
     |
1225 | ...                   for texture in cmd_buf_trackers.textures.used_resources() {
     |                           ------- binding `texture` declared here
1226 | ...                       let id = texture.info.id();
1227 | ...                       let should_extend = match texture.inner.get(&snatch_guard) {
     |                                                     ^^^^^^^ borrowed value does not live long enough
...
1240 | ...                                   submit_surface_textures.push(raw);
     |                                       --------------------------------- borrow later used here
...
1261 | ...                   }
     |                       - `texture` dropped here while still borrowed
```
Holding onto `used_resources` in a separate collection would get this down. Alternatively use the unsafe APIs related to accessing tracked textures.

Finally this means that every backend needs to do this work, even though they don't need to access their surface textures.

Does this sound about what you expect and do you want me to move forward with the change?

cwfitzgerald: Hmmmmm

My resistance to this is solely because having a container trait in wgpu-hal feels very wrong - we should be able to do what we need without needing that.

There are potential solutions where we clone the texture, and extract the surface texture _after_ pre_submit, maybe?

As it stands, this is not sound by construction - it happens to be sound because the snatch guard is held open, but it would be nice to have a raw resource attached by lifetime to the actual snatch guard, so it can't be messed up.

I'm not too worried about the cost of doing it "unnecessarily" for other backends, we do that a lot.

udoprog: So this might be a workable solution, we can change `Snatchable<T>` so that `pub fn get()` works like this instead (introducing the `'snatch`) lifetime:

```rust
pub fn get<'snatch>(&self, _guard: &'snatch SnatchGuard) -> Option<&'snatch T> {
    unsafe { (*self.value.get()).as_ref() }
}
```

That would mean that references to the `A::SurfaceTexture` would outlive the loop and live as long as the snatch guard instead of the handle it was taken from, allowing them to be used. This does seem like the intent of the snatch guard.

How does this sound to you?

udoprog: I've reworked it to do this (in a separate commit), tell me what you think.

wgpu-core/src/device/queue.rs
```diff
@@ -1468,8 +1448,18 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                 }
             }
```
udoprog: Note that we essentially have to inline `pre_submit` here, since we're using a reference into dst_textures and pre_submit would need `&mut self` over the whole structure. We clear dst_textures after submit instead.

wgpu-core/src/device/queue.rs
udoprog: So I'm pretty sure now that lifetime is wrong. Since Snatchable types can be destructed independently of the lock, allowing such a lifetime would allow them to be used after they've been freed.

udoprog: I've opted now to do the naive thing so we can ship this: take a clone of the texture and reborrow before passing it in.

cwfitzgerald (approved):
Looks a lot better and performs well based on testing, lets get this out there!

emilk:
@EmbersArc reports this PR causes 100% CPU usage:

* https://github.com/emilk/egui/issues/5092
* OS: Arch Linux 6.10.8-arch1-1 (64-bit)
* X11 with NVIDIA 560.35.03

cwfitzgerald:
This got entirely rewritten in the 0.20.1 update (wgpu-hal 0.21), so it's unlikely to be this PR.

Replying upstream

PR #4969, opened 2024-01-03
Texture snatching

nical:
**Connections**

Most of the remaining work towards #4787
Depends on #4954 

**Description**

This is almost identical to the work that was done to snatch buffers.

- When accessing a raw bind group, check that the underlying raw textures are not destroyed.
- Add a `DestroyedTexture` structure that will hold the snatched texture until it is ready to be deleted (pending writes/submissions are done).
- Snatch the raw texture in `Texture::destroy`.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -847,10 +847,13 @@ impl<A: HalApi> Drop for BindGroup<A> {
 
 impl<A: HalApi> BindGroup<A> {
     pub(crate) fn raw(&self, guard: &SnatchGuard) -> Option<&A::BindGroup> {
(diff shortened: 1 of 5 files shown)


nical:
There appears to be a deadlock that is sometimes reproducible in the `multithreaded_compute`, I'm looking into it.

Edit: Nevermind, this is a preexisting issue.

cwfitzgerald:
This failure is solved by https://github.com/gfx-rs/wgpu/pull/4980

teoxoy (approved):
Consistent with the buffer logic, good stuff!

PR #4977, opened 2024-01-04
Use Custom Mesa for Building

### LINKED ISSUE #4961: Mesa swrast Unconditionally Segfaults on 23.3.2 or Later
Upstream https://gitlab.freedesktop.org/mesa/mesa/-/issues/10371

Disabling Linux testing for now. Will figure out other solutions if need be if this persists.

cwfitzgerald:
Closes https://github.com/gfx-rs/wgpu/issues/4961

We are now pinned to an explicit version of mesa so that updates won't surprised us anymore.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -332,9 +332,8 @@ jobs:
             os: [self-hosted, macOS]
 
           # Linux
-          # https://github.com/gfx-rs/wgpu/issues/4961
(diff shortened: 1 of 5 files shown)


cwfitzgerald (commented):
Pushing this through without review as it re-adds an important part of CI. if anyone has any comments/issue, I will follow up.

PR #4980, opened 2024-01-04
Pin DXC and Vulkan SDK version

cwfitzgerald:
Furthers the goal of reproducable builds by getting rid of a unreliable and unmaintained action in exchange for a simple script.


cwfitzgerald:
Well that was exceedingly annoying and dramatic, but it's ready to go. Thankfully I now actually understand all the problems I had during it.

nical (approved):

Issue #2804, opened 2022-06-23
Dynamic WebGPU support detection


daxpedda:
**Is your feature request related to a problem? Please describe.**
It's not really a problem at the moment, because no stable browser has WebGPU support turned on by default, yet. In the future, it would be great to be able to ship a single binary in WASM that supports both WebGL and WebGPU, depending on browser support.
Currently this is impossible, as selecting the `webgl` crate feature automatically disables the WebGPU backend.

**Describe the solution you'd like**
Detecting WebGPU support in the browser is fairly easy. So depending on it's availability, the correct backend should be chosen automatically by `wgpu`.

**Describe alternatives you've considered**
The only solution I can come up with right now, is to have two binaries. Either detect support in the browser and load the appropriate WASM binary, or ship the appropriate binary by the server depending on the requesting browser.

mikialex:
I don't think wgpu is responsible to handle the compilation process of multiple wasm binaries and browser feature detection. Although it's a feasible way to solve this issue, it should be handled by wgpu users.

Another way is we ship a new backend/implementation by wrapping up two kinds of implementations: the webgl one and webgpu one, pattern matching every api call at runtime. The downside is it actually contains two impls in one binary but we eventually can assume almost everyone uses webgpu impl.

daxpedda:
> I don't think wgpu is responsible to handle the compilation process of multiple wasm binaries and browser feature detection. Although it's a feasible way to solve this issue, it should be handled by wgpu users.

Just to clarify, I totally agree that compiling multiple WASM binaries and how to handle that is not `wgpu`s responsibility.

Feature detection is, in my opinion, `wgpu` already does this with the other backends on the desktop. Though I would already be happy if `wgpu` would allow users to do feature detection themselves.

Currently, as far as I'm aware, you can't even have WebGL and WebGPU support in the same binary, so even if users try to figure out what should be used, it's not possible to do with only one binary.

mikialex:
> Feature detection is, in my opinion,  wgpu already does this with the other backends on the desktop. 

Actually no, the wgpu backend selection is in compile time (through cargo feature). No such runtime detection.

> you can't even have WebGL and WebGPU support in the same binary

Actually yes, that is what I said impl a new backend in wgpu that dispatch api calls to webgl/webgpu 
 backend and does runtime feature detection. but I don't recommend it because from a long time perspective it's meaningless and boring

daxpedda:
[...]

PR #4984, opened 2024-01-04
Unify surface creation by introducing new `SurfaceTarget` enum

Wumpf:
**Connections**
Came up while talking about #2804 since all the different surface creation code paths were a bit in the way when working on this.

**Description**
Unifies all `create_surface_X` methods to just `create_surface` by expecting a new `SurfaceTarget` enum.
To be accurate, we're expecting `Into<SurfaceTarget>` which allows to pass anything that implements raw window handle directly.

Was a bit unsure about the `create_surface_raw` one. I think removing it in the way I did is correct, but I'm wonder if I blocked some usecases or made things less safe.

Future work:
* Trickle this down to wgpu-core?
* consolidate the values of this new enum. Right now it's just blindly ported over from existing methods

**Testing**
Compiles! Also ran examples on native (mac) and webgl.

**Checklist**

- [x] Run `cargo fmt`.


wgpu/src/lib.rs
```diff
@@ -409,6 +409,102 @@ impl Drop for Surface<'_> {
...
+pub trait WindowHandle: HasWindowHandle + HasDisplayHandle + WasmNotSendSync {}
+
+impl<T> WindowHandle for T where T: HasWindowHandle + HasDisplayHandle + WasmNotSendSync {}
+
+/// The window/canvas/surface/swap-chain/etc. a surface is attached to.
+///
+/// This is either a window or an actual web canvas depending on the platform and
+/// enabled features.
+/// Refer to the individual variants for more information.
+pub enum SurfaceTarget<'window> {
```
cwfitzgerald, line 425: Let's make this #[non_exhaustive]

wgpu/src/lib.rs
```diff
@@ -1740,258 +1836,32 @@ impl Instance {
...
     ///
-    /// # Safety
+    /// See [`SurfaceTarget`] for what targets for surface are supported.
     ///
-    /// - visual must be a valid IDCompositionVisual to create a surface upon.
-    #[cfg(dx12)]
-    pub unsafe fn create_surface_from_visual(
+    /// Most commonly used are window handles (or provider of windows handles)
+    /// which can be passed directly as they're automatically converted to [`SurfaceTarget`].
+    pub fn create_surface<'window>(
```
cwfitzgerald, line 1893: This isn't safe, unfortunately. 

I can very easily do

`instance.create_surface(SurfaceTarget::CoreAnimationLayer(std::ptr::null_mut()))` and bing bong boom segfault in safe code.

The "other" ones need to be unsafe somehow. Either by splitting this into two methods, one which accepts a raw-window-handle through ownership, one which accepts a raw variant.

Additionally I think having the raw variant in the unsafe bin is important, in case users have weird requirements where they can't hand us ownership of the window.

cwfitzgerald (dismissed):
> Trickle this down to wgpu-core?

Yes please!

>

Wumpf:
>  Trickle this down to wgpu-core?
> Yes please!

Let's leave this for another time, this PR is quite big & scary as is already

wgpu/src/lib.rs
```diff
@@ -409,6 +409,149 @@ impl Drop for Surface<'_> {
...
+}
+
+impl SurfaceTargetUnsafe {
+    /// Creates a [`SurfaceTargetUnsafe::RawHandle`] from a window.
+    ///
+    /// # Safety
+    ///
+    /// - `window` must outlive the resulting surface target
+    ///   (and subsequently the surface created for this target).
+    pub unsafe fn from_window<T>(window: &T) -> Result<Self, raw_window_handle::HandleError>
```
Wumpf, line 544: Didn't quite realize that the lifetime on `wgpu::Surface` is new in the upcoming release. When updating on egui this was quite a bit in the way, making me go the `unsafe` route. This utility came in handy then and I think others may want to use it as well.
I believe the upcoming raw window handle change of internally having an Arc might make things better again.

cwfitzgerald (approved):
Very good stuff!

PR #4990, opened 2024-01-04
Fix Queue::write_texture, Fix DX12 write_texture_subset_2d and re-enable the test.

### LINKED ISSUE #3072: Get test write_texture_subset to pass with dx12

dtzxporter:
Queue::write_texture calculated the rows_per_image incorrectly for block compressed textures. This causes the DX12 backend to fail when uploading a block compressed image. Enabled the write_texture_subset_2d test now that it passes.

**Connections**
Closes https://github.com/gfx-rs/wgpu/pull/3992  
Closes https://github.com/gfx-rs/wgpu/issues/3072

**Description**
`Queue::write_texture` fails for block compressed textures on DX12.

**Testing**
Enabled the write_texture_subset_2d test now that it passes.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -182,6 +182,10 @@ Passing an owned value `window` to `Surface` will return a `Surface<'static>`. S
 - Align `wgpu_types::CompositeAlphaMode` serde serialization to spec. By @littledivy in [#4940](https://github.com/gfx-rs/wgpu/pull/4940)
 - Fix error message of `ConfigureSurfaceError::TooLarge`. By @Dinnerbone in [#4960](https://github.com/gfx-rs/wgpu/pull/4960)
 
+#### DX12
+
+- Fixed D3D12_SUBRESOURCE_FOOTPRINT calculation for block compressed textures which caused a crash with `Queue::write_texture` on DX12. By @DTZxPorter in [#4990](https://github.com/gfx-rs/wgpu/pull/4990)
+
 #### Vulkan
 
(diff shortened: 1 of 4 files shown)


cwfitzgerald:
We had a PR very similar that was abandoned - does this solve the same problem? https://github.com/gfx-rs/wgpu/pull/3992 (issue https://github.com/gfx-rs/wgpu/issues/3072)

Either way would it be possible to write a small test for this case?

dtzxporter:
I believe that the root cause is a bad calculation here:

(Should also be height_blocks) as `rows_per_image` doc says it's in blocks.
https://github.com/gfx-rs/wgpu/blob/trunk/wgpu-core/src/device/queue.rs#L739C13-L743

It just happens that the DX12 backend used `rows_per_image` and the vulkan backend doesn't seem to, so this leads to the crash. I'm happy to fix whatever here, and or accept a fix to both changes to make the DX12 backend match the other backends.

cwfitzgerald:
Definitely whatever you're willing to do - definitely the more fixed the merrier! 

Seems to be a CI failure on DX12 though.

dtzxporter:
All tests should pass now, and I enabled the write_texture_subset_2d test now that it passes on DX12!

cwfitzgerald (approved):

PR #4998, opened 2024-01-05
[spv/msl/hlsl-out] support pipeline constant value replacements

teoxoy:
Adds support for value replacements of pipeline-overridable constants for our 3 main back-ends.

- [ ] File issues to implement override constants in terms of platform features, like SPIR-V's specialization constants. This would let us move SPIR-V generation from pipeline creation time back to module creation time.


cwfitzgerald (approved):
Just booping to get out of my review queue :)

naga/src/arena.rs
```diff
@@ -297,6 +297,15 @@ impl<T> Arena<T> {
             .map(|(i, v)| unsafe { (Handle::from_usize_unchecked(i), v) })
     }
 
+    /// Drains the arena, returning an iterator over the items stored.
+    pub fn drain(self) -> impl DoubleEndedIterator<Item = (Handle<T>, T, Span)> {
```
jimblandy: We should only use the name `drain` for things that work like the standard library's `drain`: they take a mutable reference and leave `self` empty when iteration is complete. So either this function should be changed to work like `Vec::drain`, etc, or it should be renamed to something like `into_blah`.

It seems like making this a proper `drain` would be as simple as moving the `replace` that the callers are doing now into this function.

naga/src/back/mod.rs
```diff
@@ -16,6 +16,14 @@ pub mod spv;
 #[cfg(feature = "wgsl-out")]
 pub mod wgsl;
 
+#[cfg(any(
+    feature = "hlsl-out",
+    feature = "msl-out",
+    feature = "spv-out",
+    feature = "glsl-out"
+))]
+pub mod pipeline_constants;
```
jimblandy: Does this module need to be `pub`? Should it be part of Naga's public API?

(I tried deleting the `pub` keyword and everything still compiled.)

naga/tests/snapshots.rs
```diff
@@ -331,18 +342,27 @@ fn check_targets(
...
     {
         if targets.contains(Targets::METAL) {
+            let mut pipeline_options = params.msl_pipeline.clone();
+            let old = std::mem::replace(
+                &mut pipeline_options.constants,
+                params.pipeline_constants.clone(),
+            );
+            if !old.is_empty() {
+                panic!("Supply pipeline constants via pipeline_constants instead of msl_pipeline.constants!");
+            }
```
jimblandy: This could be simpler:

```suggestion
            if !params.msl_pipeline.constants.is_empty() {
                panic!("Supply pipeline constants via pipeline_constants instead of msl_pipeline.constants!");
            }
            let mut pipeline_options = params.msl_pipeline.clone();
            pipeline_options.constants = params.pipeline_constants.clone();
```

naga/tests/snapshots.rs
```diff
@@ -331,18 +342,27 @@ fn check_targets(
...
     #[cfg(all(feature = "deserialize", feature = "msl-out"))]
     {
         if targets.contains(Targets::METAL) {
+            let mut pipeline_options = params.msl_pipeline.clone();
+            let old = std::mem::replace(
+                &mut pipeline_options.constants,
+                params.pipeline_constants.clone(),
+            );
+            if !old.is_empty() {
+                panic!("Supply pipeline constants via pipeline_constants instead of msl_pipeline.constants!");
```
jimblandy, line 353: In a follow-up, let's try to make the parameters for all the snapshot backends work in the same way.

naga/src/back/pipeline_constants.rs
```diff
@@ -0,0 +1,118 @@
...
+            ty: override_.ty,
+            init,
+        };
+        module.constants.append(constant, span);
+    }
+
+    Ok(Cow::Owned(module))
+}
+
+fn map_value_to_literal(value: f64, scalar: Scalar) -> Result<Literal, PipelineConstantError> {
```
jimblandy, line 59: This is a perfect candidate for a `#[test]` just below.

naga/src/back/pipeline_constants.rs
```diff
@@ -0,0 +1,118 @@
...
+    match scalar {
+        Scalar::BOOL => {
+            // https://webidl.spec.whatwg.org/#js-boolean
+            let value = value != 0.0 && !value.is_nan();
+            Ok(Literal::Bool(value))
+        }
+        Scalar::I32 => {
+            // https://webidl.spec.whatwg.org/#js-long
+            if value.is_finite() {
+                let value = value.abs().floor() * value.signum();
```
jimblandy: Isn't this just `value.trunc()`?

naga/src/arena.rs
```diff
@@ -297,6 +297,15 @@ impl<T> Arena<T> {
...
+        self.data
+            .into_iter()
+            .zip(self.span_info.into_iter())
+            .enumerate()
+            .map(|(i, (v, span))| unsafe { (Handle::from_usize_unchecked(i), v, span) })
+    }
```
jimblandy: This actually works - with corresponding changes to `process_overrides`. The return value doesn't seem to extend the lifetime of the `&mut self`, so `process_overrides` is free to mutate `module`.

```suggestion
    pub fn drain(&mut self) -> impl DoubleEndedIterator<Item = (Handle<T>, T, Span)> {
        let arena = std::mem::take(self);
        arena.data
            .into_iter()
            .zip(arena.span_info.into_iter())
            .enumerate()
            .map(|(i, (v, span))| unsafe { (Handle::from_usize_unchecked(i), v, span) })
    }
```

jimblandy (approved):
Okay, this looks good to me.

I have various suggestions. I'll push commits for some of them to https://github.com/teoxoy/wgpu/compare/pipeline-constants...jimblandy:wgpu:pipeline-constants

naga/src/arena.rs
jimblandy: I implemented this on my branch. See what you think.

naga/src/back/pipeline_constants.rs
jimblandy, line 59: It would be nicer to use early returns for errors, and reserve top-level control flow for the success path. This is a rule in SpiderMonkey, and the effect is really nice. I've implemented this on my branch; see what you think.

naga/src/back/pipeline_constants.rs
teoxoy: `trunc` seems to be equivalent indeed.

PR #4999, opened 2024-01-05
Propagate errors when openning/closing a command encoder

nical:
**Connections**

Downstream bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1873047

**Description**

There are a few `unwrap`s in `CommandEncoder` (with TODO comment about dealing with the errors), which Firefox's fuzzer is too good at hitting. This PR just does the plumbing to propagate the errors instead of packing.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/command/clear.rs
+++ b/wgpu-core/src/command/clear.rs
@@ -5,6 +5,7 @@ use crate::device::trace::Command as TraceCommand;
 use crate::{
     api_log,
     command::CommandBuffer,
+    device::DeviceError,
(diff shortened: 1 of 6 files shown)


Wumpf (approved):
less unwraps <3 <3
might be worth a note in the changelog, albeit from the looks of it it's kind hard to pinpoint concrete crashes

nical:
Device loss is what our CI seemed to be hitting

Issue #4558, opened 2023-06-06
Mistranslation of `loop` with `continuing` block


Patryk27:
Hi,

Converting this wgsl:

```rust
var<private> global: vec3<u32>;
@group(0) @binding(2) 
var global_1: texture_storage_2d<rgba16float,read_write>;

fn function() {
    var phi_79_: i32;
    var phi_89_: i32;

    let _e17 = global;
    phi_79_ = 0;
    loop {
        let _e22 = phi_79_;
        let _e23 = (_e22 < 8);
        if _e23 {
            phi_89_ = (_e22 + 1);
        } else {
            textureStore(global_1, vec2<u32>(_e17.x, _e17.y), vec4<f32>(0.10000000149011612, 0.20000000298023224, 0.30000001192092896, 1.0));
            phi_89_ = 0;
        }
        let _e26 = phi_89_;
        continue;
        continuing {
            _ = !(select(false, true, _e23));
            phi_79_ = _e26;
            break if !(select(false, true, _e23));
        }
    }
    return;
}

@compute @workgroup_size(8, 8, 1) 
fn main(@builtin(global_invocation_id) param: vec3<u32>) {
    global = param;
    function();
}
```

... yields an invalid msl:

```metal
// language: metal2.0
#include <metal_stdlib>
#include <simd/simd.h>

using metal::uint;

constant metal::float4 const_type_9_ = {0.10000000149011612, 0.20000000298023224, 0.30000001192092896, 1.0};

void function(
    thread metal::uint3& global,
    metal::texture2d<float, metal::access::read_write> global_1
) {
    int phi_79_ = {};
    int phi_89_ = {};
    metal::uint3 _e17 = global;
    phi_79_ = 0;
    bool loop_init = true;
    while(true) {
        if (!loop_init) {
            bool unnamed = !(((phi_79_ < 8) ? true : false));
            phi_79_ = phi_89_;
            if (!(((phi_79_ < 8) ? true : false))) {
                break;
            }
        }
        loop_init = false;
        int _e22 = phi_79_;
        bool _e23 = _e22 < 8;
        if (_e23) {
            phi_89_ = _e22 + 1;
        } else {
            global_1.write(const_type_9_, metal::uint2(metal::uint2(_e17.x, _e17.y)));
            phi_89_ = 0;
        }
        int _e26 = phi_89_;
        continue;
    }
    return;
}

struct main_Input {
};
kernel void main_(
[...]

PR #5001, opened 2024-01-05
[naga wgsl-in] Fix parsing `break if`s

### LINKED ISSUE #4982: naga: Miscompilation with "break if"
Hi, this wgsl:

```wgsl
fn main() -> u32 {
  var counter = 0u;

  loop {
    counter += 1u;

    continuing {
      break if counter == 5u;
    }
  }

  return counter;
}
```

... gets compiled into this Metal:

```metal
// language: metal1.0
#include <metal_stdlib>
#include <simd/simd.h>

using metal::uint;


uint main_(
) {
    uint counter = 0u;
    uint _e5 = counter;
    bool loop_init = true;
    while(true) {
        if (!loop_init) {
            if (_e5 == 5u) {
                break;
            }
        }
        loop_init = false;
        uint _e3 = counter;
        counter = _e3 + 1u;
    }
    uint _e8 = counter;
    return _e8;
}
```

I'm specifically wondering about the `_e5 == 5u` comparison there - shouldn't the code either refer to `counter` or update `_e5` somewhere? Currently it looks like the comparison would always yell `false`.

(I might be also misunderstanding how `break if` is supposed to work, though!)

Patryk27:
Closes https://github.com/gfx-rs/wgpu/issues/4982.

Note that https://github.com/gfx-rs/wgpu/issues/4558#issuecomment-1879019338 does _not_ get solved by this commit - in particular, something like this:

```wgsl
fn foo() {
    var a = 0;

    loop {
        let cmp = a < 8;

        if cmp {
            a += 1;
        }

        continuing {
            break if cmp;
        }
    }
}
```

... still gets miscompiled (when going to glsl, hlsl _and_ msl):

```metal
void foo(
) {
    int a_2 = 0;
    bool loop_init_4 = true;
    while(true) {
        if (!loop_init_4) {
            if (a_2 < 8) { // whoopsie - should read `cmp`, not redo the comparison here!
                break;
            }
        }
        loop_init_4 = false;
        int _e2 = a_2;
        bool cmp = _e2 < 8;
        if (cmp) {
            int _e6 = a_2;
            a_2 = _e6 + 1;
        }
    }
    return;
}
```

... 'cause there's some another bug somewhere, not related to the wgsl frontend, that I can't put my finger on right now (using `var cmp` instead of `let cmp` generates correct output, tho).

### DIFF
--- a/naga/src/front/wgsl/lower/mod.rs
+++ b/naga/src/front/wgsl/lower/mod.rs
@@ -1328,7 +1328,9 @@ impl<'source, 'temp> Lowerer<'source, 'temp> {
                 let mut emitter = Emitter::default();
                 emitter.start(&ctx.function.expressions);
                 let break_if = break_if
-                    .map(|expr| self.expression(expr, &mut ctx.as_expression(block, &mut emitter)))
+                    .map(|expr| {
+                        self.expression(expr, &mut ctx.as_expression(&mut continuing, &mut emitter))
+                    })
                     .transpose()?;
                 continuing.extend(emitter.finish(&ctx.function.expressions));
 
--- a/naga/tests/in/break-if.wgsl
+++ b/naga/tests/in/break-if.wgsl
@@ -30,3 +30,15 @@ fn breakIf(a: bool) {
         }
     }
 }
+
+fn breakIfSeparateVariable() {
+    var counter = 0u;
+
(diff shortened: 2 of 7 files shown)


teoxoy (approved):
Thanks, looks good!

PR #5012, opened 2024-01-08
Properly Deal with Timeouts Inside device::maintain

### LINKED ISSUE #3601: Timeout on `Device::maintain` with `Maintain::WaitForSubmissionIndex` is ignored
Whenever `Device::maintain` with `Maintain::WaitForSubmissionIndex` times out, wgpu assumes that the specified submission index is done and initiates cleanup as-if the frame was done. This can cause crashes down the line as buffers may be freed too early (haven't confirmed this part, but seems shaky!).
There is a `StuckGpu` (on `WaitIdleError` and `QueueSubmitError`) but it is unused currently.

Wgpu needs to react to timeouts by *not* assuming the passed submission index is done and instead query the actual done frame.

---

The situation is particularly weird on WebGL where the max timeout is browser defined (and usually much lower than our hardcoded 5 seconds), see https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/clientWaitSync, `MAX_CLIENT_WAIT_TIMEOUT_WEBGL`. Instead, we pass a timeout of zero, meaning that any `WaitForSubmissionIndex` call will pretend the frame is done. However, since freeing gl buffers that are still in use has no practical repercussions, this does not cause any crashes etc..
To make matters worse, on WebGL we can't actually block on this by calling it repeatedly, since [clientWaitSync](https://registry.khronos.org/webgl/specs/latest/2.0/#3.7.14) will not finish unless control is given back to the browser. Effectively meaning that we can't actually block on any particular event. 

Worth noting here that blocking in this manner isn't possible at all either on WebGPU either where `poll` does not exist. 

---

Given that `poll` is a pure Rust space method and not covered by the WebGPU spec I propose to return an enum for all these different failure modes and avoid any panic on `StuckGpu`, i.e. as far as possible not treating hitting the timeout as a crash but rather as an expected error state.

### LINKED ISSUE #4589: Device::poll with Maintain::Wait returns early and with an incorrect value when a 5-second timeout is reached
We're using wgpu to do some computing tasks where individual compute passes will take more than CLEANUP_WAIT_MS.  In the case of a long running compute pass, waiting on the submission index with a Maintain::Wait or Maintain::WaitForSubmissionIndex will return after the timeout is exceeded, even if the task is not complete.  It will also return true indicating the queue is empty if there's nothing else in the queue.  

This timeout should be configurable.  If that's hard, in the meantime, if this timeout is reached, device::poll call should return false to indicate there is still an item in the queue.  Regardless, this behavior should be documented.

cwfitzgerald:
**Connections**

Closes #4589 
Closes #3601

**Description**

We were not properly dealing with polls which timed out. This exposed the concept of timeouts to the user so that they can properly deal with polls which time out.

This doesn't fix anything with timeouts inside surface.configure.

**Testing**

Added timeout test.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -347,7 +347,6 @@ jobs:
     
     name: Test WebAssembly
     runs-on: ubuntu-latest
-    needs: [check]
 
     steps:
       - name: checkout repo
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1504,6 +1504,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "151665d9be52f9bb40fc7966565d39666f2d1e69233571b71b87791c7e0528b3"
 
 [[package]]
+name = "gloo-timers"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bbb143cf96099802033e0d4f4963b19fd2e0b728bcf076cd9cf7f6634f092994"
+dependencies = [
+ "futures-channel",
+ "futures-core",
+ "js-sys",
+ "wasm-bindgen",
+]
+
+[[package]]
 name = "glow"
 version = "0.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4126,7 +4138,9 @@ dependencies = [
  "console_log",
  "ctor",
  "env_logger",
+ "flume",
  "futures-lite",
+ "gloo-timers",
  "heck",
  "image",
  "js-sys",
@@ -4145,6 +4159,7 @@ dependencies = [
  "wasm-bindgen-futures",
  "wasm-bindgen-test",
  "web-sys",
+ "web-time",
  "wgpu",
  "wgpu-macros",
  "wgpu-types",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -84,6 +84,7 @@ flume = "0.11"
 futures-lite = "2"
 getrandom = "0.2"
 glam = "0.25"
+gloo-timers = "0.3"
 heck = "0.4.0"
 image = { version = "0.24", default-features = false, features = ["png"] }
 ktx2 = "0.3"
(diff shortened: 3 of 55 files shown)


cwfitzgerald:
Pulled this into draft as it breaks all of our WebGL2 tests.

tl;dr: the whole reason device.poll(Wait) worked on WebGL2 is because of a the very logic bug that this fixes. WebGL2 is _always_ treated as a timeout, so it always treats the wait as finished. Which is completely UB on other backends, but on WebGL2 it doesn't matter as there's no way to observe the lack of it actually finishing.

Going to need to:
- [ ] Add a feature to make poll(Wait) fail validation.
- [x] Add an async helper function for a cross platform poll(Wait)
- [x] Switch all existing polls to that function.

cwfitzgerald:
I still want to deal with this, it was just a much larger task than I initially thought.

teoxoy:
This should also include the removal of the expected failure on MoltenVK (https://github.com/gfx-rs/wgpu/pull/5964/commits/f38bdd8e67ce221e3104be89acd7d7e335e37b9c).

thalisvilela:
My first time working with WebGPU today, trying to implement heavy slow compute shader workload, as i set up and tested everything, as soon i started throwing real work at it, stumbled upon this issue immediately... Not my lucky day...

Wumpf:
@thalisvilela fwiw this issue does not affect WebGPU but all native backends + WebGL, but that's probably what you meant :)

thalisvilela:
@Wumpf thank you very much for the tip! I would have probably give a shot at these other methods, you just saved me from a lot of frustration! It's the first time i'm working directly with "shaders", i've mostly used CUDA before. I'm glad this issue surfaced immediately to be honest.

torokati44:
@thalisvilela While this is off-topic, you might also be interested in these projects (not that `wgpu` is any less capable of pure compute stuff):
https://github.com/EmbarkStudios/rust-gpu
https://github.com/tracel-ai/cubecl

PR #5014, opened 2024-01-08
Track the actual size of the buffer for memory init

### LINKED ISSUE #4731: Panic Buffer has an uninitialized range with an end not aligned to 4 (end was 31) if buffer size is 31
**Description**

Running into this assertion in the CTS after landing arcanization. There is a decent chance that this isn't actually caused by arcanization but is that it perhaps fixed something that led us to cover a case that was not run before.
Edit: I have growing confidence that it isn't related to arcanization and we might have had it before it landed. 

Test log: https://treeherder.mozilla.org/logviewer?job_id=436806642&repo=try&lineNumber=2709

The assertion being hit is: https://github.com/gfx-rs/wgpu/blob/16ec5b00fc50c688a987effed9f470db58b2389e/wgpu-core/src/command/memory_init.rs#L244
The end of the range is 31.

At a glance, one might think that the logic that rounds up some range [here](https://github.com/gfx-rs/wgpu/blob/16ec5b00fc50c688a987effed9f470db58b2389e/wgpu-core/src/command/memory_init.rs#L177) would prevent that, but in practice the code that iterates the ranges [here](https://github.com/gfx-rs/wgpu/blob/16ec5b00fc50c688a987effed9f470db58b2389e/wgpu-core/src/init_tracker/mod.rs#L86) pulls from the original (not fixed up) ranges.

There is a lot going on here and my I haven't quite wrapped my head around it but a notable thing is that the buffer size is 31.

**Repro steps**

For firefox devs (or adventurous folks):
```bash
# in the mozilla-central repository
./mach wpt '/_mozilla/webgpu/chunked/63/cts.https.html?q=webgpu:shader,execution,robust_access_vertex:vertex_buffer_access:*'
```

I can upload a wgpu trace of the bug (although it's a bit large).

nical:
**Connections**

Fixes #4731

**Description**

The init tracking code was using the buffer's requested size rather than the actual size which aligned specifically for the purpose of the requirements of initializing/clearing.

**Testing**

Covered by the CTS.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -569,7 +569,7 @@ impl<A: HalApi> Device<A> {
             device: self.clone(),
             usage: desc.usage,
             size: desc.size,
-            initialization_status: RwLock::new(BufferInitTracker::new(desc.size)),
+            initialization_status: RwLock::new(BufferInitTracker::new(aligned_size)),
             sync_mapped_writes: Mutex::new(None),
             map_state: Mutex::new(resource::BufferMapState::Idle),
             info: ResourceInfo::new(desc.label.borrow_or_default()),


Wumpf (approved):
makes sense, good catch

PR #5018, opened 2024-01-08
Prefer `Arc::into_inner` over `Arc::try_unwrap`.

jimblandy:
As explained in its Rust documentation, calling `Arc::try_unwrap(a)`
where the intention is to drop `a` if it is not uniquely owned suffers
from a race condition: if two threads call `try_unwrap` simulaneously
on the last two `Arc`s referring to a value, they will both see a
reference count greater than one, and then both drop the `Err`, with
neither thread getting the value. Because `try_unwrap` returns the
`Arc` in its `Err`, there is a check-versus-use gap. `Arc::into_inner`
returns an `Option` of the result, and guarantees that checking the
count and dropping the `Arc` are indivisible.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ env:
   REPO_MSRV: "1.71"
   # This is the MSRV used by the `wgpu-core`, `wgpu-hal`, and `wgpu-types` crates,
   # to ensure that they can be used with firefox.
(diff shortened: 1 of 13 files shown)


cwfitzgerald (approved):
Seems like a pretty clear benefit!

cwfitzgerald:
Naga changes immaterial, merging.

Issue #5022, opened 2024-01-09
panic in `tracker_assert_in_bounds`


nical:
**Description**

```
assertion failed: index < self.owned.len()) at wgpu-core/src/track/metadata.rs:52
```

Backtrace:

```
10  core::panicking::panic() [panicking.rs:82e1608dfa6e0b5569232559e3d385fea5a93112 : 127 + 0x7]
     rbx = 0x000001cba80782e8    rbp = 0x0000004eba78e6e0
     rsp = 0x0000004eba78e650    r12 = 0x000001cba80e03e8
     r13 = 0x000001cb9fcaac80    r14 = 0x000001cba80e07b8
     r15 = 0x00000000ffffbce0    rip = 0x00007ff98592b4f2
    Found by: call frame info
11  track::metadata::ResourceMetadata<dx12::Api,id::Id<command::bundle::RenderBundle<empty::Api> >,command::bundle::RenderBundle<dx12::Api> >::tracker_assert_in_bounds<dx12::Api,id::Id<command::bundle::RenderBundle<empty::Api> >,command::bundle::RenderBundle<dx12::Api> >(unsigned long long) [metadata.rs : 52 + 0x17]
     rbx = 0x000001cba80782e8    rbp = 0x0000004eba78e6e0
     rsp = 0x0000004eba78e6c0    r12 = 0x000001cba80e03e8
     r13 = 0x000001cb9fcaac80    r14 = 0x000001cba80e07b8
     r15 = 0x00000000ffffbce0    rip = 0x00007ff9838e7353
    Found by: call frame info
12  track::stateless::StatelessTracker<dx12::Api,id::Id<resource::TextureView<empty::Api> >,resource::TextureView<dx12::Api> >::tracker_assert_in_bounds(unsigned long long) [stateless.rs : 132]
    Found by: inlining
13  track::stateless::impl$1::remove_abandoned(track::stateless::StatelessTracker<dx12::Api,id::Id<resource::TextureView<empty::Api> >,resource::TextureView<dx12::Api> >*, id::Id<resource::TextureView<empty::Api> >) [stateless.rs : 97]
    Found by: inlining
[...]

PR #5024, opened 2024-01-09
Fix a bounds check in remove_abandoned

nical:
**Connections**

May or may not be a fix for #5022 (I have not been able to reproduce so far).

**Description**

Fixes what looks like a classic off-by-one error comparing an index with a length.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


cwfitzgerald (approved):
Big whoops

PR #5026, opened 2024-01-09
Don't cast the dynamic offsets count to u8

nical:
**Connections**

Supersedes #4918

**Description**

For some reason we store the dynamic offset count of the SetBindGroup command as a u8 even though it is used as an usize before and after. The conversion to u8 crashes if the number of offsets doesn't fit.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


teoxoy (approved):
Much better without all the conversions!

PR #5027, opened 2024-01-09
Validate that a binding offset fits in the buffer

nical:
**Connections**

Crash found when [running the CTS](https://treeherder.mozilla.org/logviewer?job_id=442315504&repo=try&lineNumber=1605) in Firefox.

**Description**

If the binding size is not specified we didn't validate that the offset was in the bounds of the buffer, causing an overflow panic on debug and UB/nonsense on release. This PR adds the missing validation.

**Testing**

Covered by the CTS

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -1832,7 +1832,16 @@ impl<A: HalApi> Device<A> {
                 }
                 (size.get(), end)
             }
-            None => (buffer.size - bb.offset, buffer.size),
+            None => {
+                if buffer.size < bb.offset {
+                    return Err(Error::BindingRangeTooLarge {
+                        buffer: bb.buffer_id,
+                        range: bb.offset..bb.offset,
+                        size: buffer.size,
+                    });
+                }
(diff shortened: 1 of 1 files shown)


teoxoy (approved):
Nice!

wgpu-core/src/device/resource.rs
```diff
@@ -1832,7 +1832,16 @@ impl<A: HalApi> Device<A> {
                 }
                 (size.get(), end)
             }
-            None => (buffer.size - bb.offset, buffer.size),
+            None => {
+                if buffer.size < bb.offset {
+                    return Err(Error::BindingRangeTooLarge {
+                        buffer: bb.buffer_id,
+                        range: bb.offset..bb.offset,
```
cwfitzgerald, line 1839: My only concern is that this faux range may make the error message very confusing for the user, as this range never actually shows up in their code.

cwfitzgerald (changes requested):

PR #5032, opened 2024-01-10
Make DeviceLostClosure.from_c mutate the closure before dropping it.

bradwerth:
**Description**
Since `DeviceLostClosureC` requires that it be called before it is dropped, and we receive an existing `DeviceLostClosureC` struct as param in `DeviceLostClosure.from_c`, we need to mark the param as called before it drops out of scope.

**Testing**
Tested by embedders using `DeviceLostClosure.from_c`.

**Checklist**

- [X] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -224,13 +224,13 @@ pub type DeviceLostCallback = Box<dyn Fn(DeviceLostReason, String) + 'static>;
 
 pub struct DeviceLostClosureRust {
     pub callback: DeviceLostCallback,
-    called: bool,
+    consumed: bool,
 }
 
 impl Drop for DeviceLostClosureRust {
     fn drop(&mut self) {
-        if !self.called {
-            panic!("DeviceLostClosureRust must be called before it is dropped.");
+        if !self.consumed {
+            panic!("DeviceLostClosureRust must be consumed before it is dropped.");
         }
     }
 }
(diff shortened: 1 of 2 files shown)


wgpu-core/src/device/mod.rs
```diff
@@ -294,14 +294,18 @@ impl DeviceLostClosure {
...
         // Build an inner with the values from closure, ensuring that
         // inner.called is false.
         let inner = DeviceLostClosureC {
             callback: closure.callback,
             user_data: closure.user_data,
             called: false,
         };
+
+        // Mark the original closure as called, so we can safely drop it.
+        closure.called = true;
```
cwfitzgerald: This doesn't actually do anything - `closure` is dropped immediately after this.

cwfitzgerald (dismissed):

wgpu-core/src/device/mod.rs
nical: It prevents the `drop` impl from panicking (it asserts that we don't leak the closure by checking `called`).

nical (approved):
nit: renaming `called` into `consumed` would make this easier to understand.

wgpu-core/src/device/mod.rs
bradwerth: I've changed the name of the member from "called" to "consumed" to clarify its purpose, and why it is mutated in this case.

Issue #5019, opened 2024-01-09
panic in create_shader_module when naga parse Result is unwrapped


waynr:
**Description**
I am building a shader livecoding tool using winit + wgpu where I have an off-main thread watching the shader file for changes as I edit it, recreating the render pipeline, then sending that via the winit EventLoopProxy as a user event to be handled by updating a struct holding the "active" render pipeline.

I have found that when I make shader coding mistake (ie no return value) a naga parse is attempted on the code and the resulting Err unwrapped: https://github.com/gfx-rs/wgpu/blob/trunk/wgpu/src/backend/direct.rs#L930

**Repro steps**
Run `Device.create_shader_module` on an invalid glsl shader.

**Expected vs observed behavior**
My filesystem watcher thread panics during the attempted shader recompile. I would expect that fallible operations on `Device` such as `create_shader_module` should return a `Result` so users like myself can decide how to handle them (in this case I would report the parse failure in my logs and continue).

**Extra materials**
```
thread 'notify-rs debouncer loop' panicked at /home/wayne/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.18.0/src/backend/direct.rs:924:61:
called `Result::unwrap()` on an `Err` value: [Error { kind: UnknownVariable("meow"), meta: Span { start: 2138, end: 2142 } }]
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/std/src/panicking.rs:597:5
   1: core::panicking::panic_fmt
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/panicking.rs:72:14
   2: core::result::unwrap_failed
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/result.rs:1652:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/result.rs:1077:23
   4: <wgpu::backend::direct::Context as wgpu::context::Context>::device_create_shader_module
             at /home/wayne/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.18.0/src/backend/direct.rs:924:30
   5: <T as wgpu::context::DynContext>::device_create_shader_module
[...]

PR #5034, opened 2024-01-10
don't panic if naga parsing of shader source fails

### LINKED ISSUE #2545: ShaderSource::Glsl panics if shader has syntax error

waynr:
**Connections**
https://github.com/gfx-rs/wgpu/issues/5019

fixes https://github.com/gfx-rs/wgpu/issues/2545

**Description**
In my application shaders are meant to be hot-reloaded on filesystem changes.
If there is a syntax issue with the newly-updated shader source `naga` parsing
of the file can fail. In current trunk, the `Result` returned by `naga` parsing
is simply unwrapped.

This PR removes the `naga` parse unwrap and instead bubbles the `Result` up
through the call stack so to be handled by the calling code.

I am opening this PR prior to updating the changelog and getting `cargo xtask
test` to run just to start discussion (should have commits addressing those
shortly after opening it).

**Testing**
I've updated the examples and test cases that were already calling
`Device.create_shader_module`.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/naga/src/front/spv/error.rs
+++ b/naga/src/front/spv/error.rs
@@ -1,5 +1,9 @@
 use super::ModuleState;
 use crate::arena::Handle;
+use codespan_reporting::diagnostic::Diagnostic;
+use codespan_reporting::files::SimpleFile;
+use codespan_reporting::term;
+use termcolor::{NoColor, WriteColor};
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
@@ -126,4 +130,25 @@ pub enum Error {
          come from a binding)"
     )]
     NonBindingArrayOfImageOrSamplers,
+}
+
+impl Error {
+    pub fn emit_to_writer(&self, writer: &mut impl WriteColor, source: &str) {
+        self.emit_to_writer_with_path(writer, source, "glsl");
+    }
+
+    pub fn emit_to_writer_with_path(&self, writer: &mut impl WriteColor, source: &str, path: &str) {
+        let path = path.to_string();
+        let files = SimpleFile::new(path, source);
+        let config = codespan_reporting::term::Config::default();
+        let diagnostic = Diagnostic::error().with_message(format!("{self:?}"));
+
+        term::emit(writer, &config, &files, &diagnostic).expect("cannot write error");
+    }
+
+    pub fn emit_to_string(&self, source: &str) -> String {
+        let mut writer = NoColor::new(Vec::new());
+        self.emit_to_writer(&mut writer, source);
+        String::from_utf8(writer.into_inner()).unwrap()
+    }
 }
--- a/wgpu-core/Cargo.toml
+++ b/wgpu-core/Cargo.toml
@@ -63,7 +63,7 @@ wgsl = ["naga/wgsl-in"]
 glsl = ["naga/glsl-in"]
 
 ## Enable `ShaderModuleSource::SpirV`
-spirv = ["naga/spv-in", "dep:zerocopy"]
+spirv = ["naga/spv-in", "dep:bytemuck"]
 
 ## Implement `Send` and `Sync` on Wasm, but only if atomics are not enabled.
 ##
@@ -99,6 +99,7 @@ dx12 = ["hal/dx12"]
 arrayvec = "0.7"
 bit-vec = "0.6"
 bitflags = "2"
+bytemuck = { version = "1.14", optional = true }
 codespan-reporting = "0.11"
 indexmap = "2"
 log = "0.4"
@@ -112,7 +113,6 @@ rustc-hash = "1.1"
 serde = { version = "1", features = ["serde_derive"], optional = true }
 smallvec = "1"
 thiserror = "1"
-zerocopy = { version = "0.7", optional = true }
 
 [dependencies.naga]
 path = "../naga"
--- a/wgpu-core/Cargo.toml
+++ b/wgpu-core/Cargo.toml
@@ -59,6 +59,9 @@ serial-pass = ["serde", "wgt/serde", "arrayvec/serde"]
 ## Enable `ShaderModuleSource::Wgsl`
 wgsl = ["naga/wgsl-in"]
 
+## Enable `ShaderModuleSource::Glsl`
+glsl = ["naga/glsl-in"]
+
 ## Implement `Send` and `Sync` on Wasm, but only if atomics are not enabled.
 ##
 ## WebGL/WebGPU objects can not be shared between threads.
--- a/wgpu-core/Cargo.toml
+++ b/wgpu-core/Cargo.toml
@@ -62,6 +62,9 @@ wgsl = ["naga/wgsl-in"]
 ## Enable `ShaderModuleSource::Glsl`
 glsl = ["naga/glsl-in"]
 
+## Enable `ShaderModuleSource::SpirV`
+spirv = ["naga/spv-in", "dep:zerocopy"]
+
 ## Implement `Send` and `Sync` on Wasm, but only if atomics are not enabled.
 ##
 ## WebGL/WebGPU objects can not be shared between threads.
@@ -109,6 +112,7 @@ rustc-hash = "1.1"
 serde = { version = "1", features = ["serde_derive"], optional = true }
 smallvec = "1"
 thiserror = "1"
+zerocopy = { version = "0.7", optional = true }
 
 [dependencies.naga]
 path = "../naga"
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -22,6 +22,8 @@ use crate::{
 use arrayvec::ArrayVec;
 use hal::Device as _;
 use parking_lot::RwLock;
+#[cfg(feature = "spirv")]
+use zerocopy::AsBytes;
 
 use wgt::{BufferAddress, TextureFormat};
 
@@ -1192,6 +1194,10 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                     #[cfg(feature = "glsl")]
                     pipeline::ShaderModuleSource::Glsl(ref code, _) => {
                         trace.make_binary("glsl", code.as_bytes())
+                    }
+                    #[cfg(feature = "spirv")]
+                    pipeline::ShaderModuleSource::SpirV(ref code, _) => {
+                        trace.make_binary("spirv", code.as_bytes())
                     }
                     pipeline::ShaderModuleSource::Naga(ref module) => {
                         let string =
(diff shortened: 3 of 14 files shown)


wgpu/src/backend/direct.rs
```diff
@@ -927,7 +927,7 @@ impl crate::Context for Context {
                     defines: defines.clone(),
                 };
                 let mut parser = naga::front::glsl::Frontend::default();
-                let module = parser.parse(&options, shader).unwrap();
```
nical: wgpu's error handling story is a tad unusual because it implements WebGPU's error handling API and has to work in a completely asynchronous environment. So `create_shader_module` cannot return a `Result`.

Instead of unwrapping here, the error should be put into an error scope the same way it is done a few lines belowe with `handle_error`.

nical (changes requested):

wgpu/src/backend/direct.rs
waynr: Oh, I wondered why no `Result`s were being returned anywhere. I'll come back to this later today and make this change!

wgpu/src/backend/direct.rs
```diff
@@ -912,8 +913,19 @@ impl crate::Context for Context {
...
+                    Ok(m) => wgc::pipeline::ShaderModuleSource::Naga(Owned(m)),
+                    Err(cause) => {
+                        self.handle_error(
+                            &device_data.error_sink,
+                            cause,
+                            LABEL,
+                            desc.label,
+                            "Device::create_shader_module",
+                        );
+                        wgc::pipeline::ShaderModuleSource::Dummy(PhantomData)
```
waynr: @nical thanks for the quick review earlier today!

Here is my second attempt. Because this function must always return something but we can't build a `ShaderModuleSource` from invalid shader source, the least disruptive approach here seemed to me like using `ShaderModuleSource::Dummy`, but it also seems less than ideal. It feels to me like the `Dummy` source is never supposed to be used at runtime (maybe it's for tests or something?) given the fact that there were explicit panics set for match arms matching it.

This in combination with `Device.on_uncaptured_error` seems to work for me. I chose not to use `push_error_scope` and `pop_error_scope` because multiple errors can and do get captured but `pop_error_scope` only returns a single error; if `pop_error_scope` could return multiple errors this would be a better approach.

Other possibilities I am considering for solving my own immediate problem:

* Assuming it's not a horrible solution in general for this PR, I could treat `ShaderModuleSource::Dummy` as a kind of error state indicator
* I could just import naga directly, attempt the GLSL parsing before attempting to call `create_shader_module`, and simply handle naga errors at that point

nical: Lets add a `ShaderModuleSource::Invalid` variant instead of using the `Dummy` one here. I think that you are otherwise on the right track.

Error scopes capturing a single error looks like a shortcoming of wgpu. It would be a valuable thing to improve, I filed #5041.

nical (changes requested):

wgpu/src/backend/direct.rs
waynr: Hey! I looked into adding `ShaderModuleSource::Invalid` as you suggested, but realized that this (and my use of `ShaderModuleSource::Dummy`) ends up leading to duplicate errors being reported which seems to be why I ended up with multiple errors on parse failure during `Device.on_uncaptured_error` -- I would get one error from the parse error handled using `handle_error` in `wgpu/src/backend/wgpu_core.rs` then another when handling the `ShaderModuleSource::Dummy` variant in `wgpu-core/src/device/resource.rs` (which previously just resulted in a panic).

So I tried a couple different approaches involving `ShaderModuleSource::Invalid` that didn't feel right, but then I realized why none of the approaches up to this point felt right...because in `wgpu/src/backend/wgpu_core.rs` there is no parsing of `wgsl` and hence no handling of parse errors at that level. This led me to the eventual conclusion that the cleanest way to improve error handling for `glsl` is to add `ShaderModuleSource::Glsl` and `ShaderModuleSource::SpirV` (the latter isn't necessary for me, but I figured i would make the change to both as long as I was working on the glsl variant), return that from `ContextWgpuCore.device_create_shader_module` and do the parsing for `glsl` and `spirv` in the same place it's done for wgsl.

This required adding a couple more variants to `CreateShaderModuleError` specific to `glsl` and `spirv` as well as the updating `naga::front::glsl::Frontend.parse` to return a `Result<_, naga::front::glsl::error::ParseError>` where `naga::front::glsl::error::ParseError` is a type I've introduced to encapsulate the `Vec<naga::front::glsl::error::Error>` that was being returned before (i gave up trying to use a vec in `CreateShaderModuleError::ParsingGlsl` after fighting with trait bounds somewhere).

Let me know if this makes sense!

waynr: Oh yeah, just in case you don't get pinged automatically by new messages: @nical

nical: Since this approach involves larger naga changes I'll deffer to @teoxoy or @jimblandy

teoxoy:
@waynr could you rebase the PR on latest trunk? so the new CI jobs can run.

teoxoy:
Ah, nvm, they didn't run because the clippy jobs are failing.

naga/src/front/spv/error.rs
```diff
@@ -127,3 +131,24 @@ pub enum Error {
     )]
     NonBindingArrayOfImageOrSamplers,
 }
+
+impl Error {
+    pub fn emit_to_writer(&self, writer: &mut impl WriteColor, source: &str) {
+        self.emit_to_writer_with_path(writer, source, "glsl");
```
teoxoy, line 137:
```suggestion
        self.emit_to_writer_with_path(writer, source, "spv");
```

teoxoy (approved):
This PR looks good to me, it's consistent with how we deal with WGSL shaders as well.

teoxoy:
@nical do you want to take another look?

wgpu-core/Cargo.toml
```diff
@@ -106,6 +112,7 @@ rustc-hash = "1.1"
 serde = { version = "1", features = ["serde_derive"], optional = true }
 smallvec = "1"
 thiserror = "1"
+zerocopy = { version = "0.7", optional = true }
```
cwfitzgerald: I would prefer ~~if we didn't take a dependency for this,~~ we used bytemuck.

This code doesn't show up in Mozilla's build, but neither bytemuck or zerocopy are used there.

cwfitzgerald (changes requested):

wgpu-core/src/device/global.rs
```diff
@@ -1197,7 +1197,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                     }
                     #[cfg(feature = "spirv")]
                     pipeline::ShaderModuleSource::SpirV(ref code, _) => {
-                        trace.make_binary("spirv", code.as_bytes())
+                        trace.make_binary("spirv", cast_slice::<u32, u8>(code.borrow()))
```
cwfitzgerald: Nit:

```suggestion
                        trace.make_binary("spirv", bytemuck::cast_slice::<u32, u8>(&code))
```

Should work and remove the need for any imports at all.

waynr: Thanks for catching that! I pushed the change in a different commit to also remove the imports as suggested.

wgpu/src/backend/wgpu_core.rs
```diff
@@ -844,22 +844,19 @@ impl crate::Context for ContextWgpuCore {
...
-                // Parse the given shader code and store its representation.
-                let options = naga::front::glsl::Options { stage, defines };
-                let mut parser = naga::front::glsl::Frontend::default();
-                let module = parser.parse(&options, shader).unwrap();
-
-                wgc::pipeline::ShaderModuleSource::Naga(Owned(module))
+                let options = naga::front::glsl::Options {
+                    stage,
+                    defines: defines.clone(),
+                };
```
a1phyr:
```suggestion
                let options = naga::front::glsl::Options { stage, defines };
```
This probably shouldn't change.

teoxoy:
This should be ready to go. Pinging @nical & @cwfitzgerald for pending reviews.

nical (approved):

cwfitzgerald (approved):

waynr:
Thanks for working through this with me!

Issue #4917, opened 2023-12-21
Resources should need only one submission to be released


gents83:
In mem_leaks.rs we can see that it has been modified to add 2 consecutive poll.
This should not be necessary.
Resources should be released in a specific order to avoid keeping as marked to be used resourced that should not be used anymore.

![image](https://github.com/gfx-rs/wgpu/assets/62186646/16a2fce2-9ed1-4df9-8ba8-c99a974e1303)

**Expected behavior:**
In a single submission all resources even with dependencies should be released

Tagging @nical & @jimblandy for info

nical:
My take is that we should not attempt to release resource in a specific order and just let the `Arc`s do their thing (if there is a particular order to preserve, then model it with strong references). This means that we stop conditionally retaining resources in the `triage_*` functions based on their reference count (I believe that's what prevents some resources from being released in a single polling step of the device.

gents83:
True, but we've to solve this problem.
Since some resources keep some Arc to other resources and we want to be sure that in 1 submit we need to be sure to release before the one holding refs to others.
Otherwise they will not be released in the same frame.
So basically what the triage was doing is:
If A has an Arc reference of B, then we've first to remove the reference and then release A, so B will result as a lonely reference and can be released.
Maybe we could find a smarter alternative?

nical:
I am in the middle of a fair amount of churn in the resource deallocation logic, and there was the winter break but I'm hoping to get to all of this soon.

The alternative I'm hoping for is the simple/dumb one: When A is dropped, it drops the reference of B. If A was the only thing holding B alive, then B naturally gets deallocated immediately after A.

PR #5036, opened 2024-01-10
Move the trace recording of destroy events out of triage_resources

nical:
**Description**

This PR moves all of the `trace::Action::DestroyFoo(id)` out of `triage_resource`. The main motivation is to avoid holding the trace lock for a long time and let the resource `drop` impls take that lock without risking deadlocks if the reference count happens to go to zero during triage.

This will allow us to remove the `free_resource` tracker in a followup which is needed for #4917.

Incidentally I'm happy about how it simplifies the code. It makes sense for this stuff to be handled by the resources themselves instead of the lifecycle triaging code which in an ideal world wouldn't need to do resource-specific logic. `triage_resource` does not need to take an `on_remove` callback anymore.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -1,3 +1,5 @@
+#[cfg(feature = "trace")]
+use crate::device::trace;
 use crate::{
     device::{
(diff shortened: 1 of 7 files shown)


cwfitzgerald (approved):
Truly amazing!

PR #5037, opened 2024-01-10
Remove the free_resources tracker

nical:
**Connections**

Requires #5036 

**Description**

Don't put resources that we want to deallocate in a special tracker that is dropped at the end of poll. Instead, just let the reference count of the resource reach zero wherever it happens (usually while triaging resources).

This is a step in the direction of simplifying the resource lifecycle code. 
It is also a step towards fixing #4917, although it isn't quite enough to fix it. The remaining issue is that `triage_submissions` happens after `triage_suspected`, so if a resource is used in a submission it won't be released on the `poll` invocation where the submission is resolved as done.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/device/life.rs
+++ b/wgpu-core/src/device/life.rs
@@ -142,9 +142,8 @@ struct ActiveSubmission<A: HalApi> {
     /// Resources to be freed once this queue submission has completed.
     ///
     /// When the device is polled, for completed submissions,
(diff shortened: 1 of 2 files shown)


cwfitzgerald (approved):
Very nice! Needs a rebase, then good to go

PR #5040, opened 2024-01-11
Pin rust-cache-version

cwfitzgerald:
rust-cache has started taking 2m30s to finish instead of 11 seconds. They had a release 6 hours ago, this pins to the sha of the previous version.

Will file upstream https://github.com/Swatinem/rust-cache/issues/182

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ env:
   CI_BINARY_BUILD: "build18"
 
   # We sometimes need nightly to use special things in CI.
-  # 
+  #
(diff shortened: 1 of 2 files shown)


jimblandy (approved):

ErichDonGubler:
@cwfitzgerald: Is an issue filed on the speed regression? I wonder if some defaults changes for the bucketing, or is this actually just a slower cache rehydration?

cwfitzgerald:
I linked it in the body, the underlying issue seems to be node20 actions take a long time to be detected as "finished"

PR #5042, opened 2024-01-11
Prevent a few integer overflows

nical:
**Connections**

Found while running the CTS in Firefox.

**Description**

We have a few places with the following pattern:

```rust
let last_foo = first_foo + foo_count;
if last_foo > foo_limit {
   // error ...
}
```

With the addition in the first line vulnerable to integer overflows. This PR addresses it by making sure that the limit is always smaller than `u32::MAX` and using `saturating_add` so that cases that would have overflown are treated as errors.

**Testing**

Covered by the CTS.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -181,6 +181,7 @@ Passing an owned value `window` to `Surface` will return a `Surface<'static>`. S
 - `BufferMappedRange` trait is now `WasmNotSendSync`, i.e. it is `Send`/`Sync` if not on wasm or `fragile-send-sync-non-atomic-wasm` is enabled. By @wumpf in [#4818](https://github.com/gfx-rs/wgpu/pull/4818)
 - Align `wgpu_types::CompositeAlphaMode` serde serialization to spec. By @littledivy in [#4940](https://github.com/gfx-rs/wgpu/pull/4940)
 - Fix error message of `ConfigureSurfaceError::TooLarge`. By @Dinnerbone in [#4960](https://github.com/gfx-rs/wgpu/pull/4960)
+- Fixed a number of panics. by @nical in [#4999](https://github.com/gfx-rs/wgpu/pull/4999), [#5014](https://github.com/gfx-rs/wgpu/pull/5014), [#5024](https://github.com/gfx-rs/wgpu/pull/5024), [#5025](https://github.com/gfx-rs/wgpu/pull/5025), [#5026](https://github.com/gfx-rs/wgpu/pull/5026), [#5027](https://github.com/gfx-rs/wgpu/pull/5027), [#5028](https://github.com/gfx-rs/wgpu/pull/5028) and [#5042](https://github.com/gfx-rs/wgpu/pull/5042).
 
 #### DX12
 
--- a/wgpu-core/src/command/render.rs
+++ b/wgpu-core/src/command/render.rs
@@ -335,7 +335,10 @@ impl IndexState {
                     IndexFormat::Uint16 => 1,
                     IndexFormat::Uint32 => 2,
                 };
-                ((range.end - range.start) >> shift) as u32
+                // Ensure that the limit is always smaller than u32::MAX so that we can
+                // use saturating additions to catch case that would have caused integers
+                // overflow later.
+                ((range.end - range.start) >> shift).min(u32::MAX as u64 - 1) as u32
             }
             None => 0,
         }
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -181,6 +181,7 @@ Passing an owned value `window` to `Surface` will return a `Surface<'static>`. S
 - `BufferMappedRange` trait is now `WasmNotSendSync`, i.e. it is `Send`/`Sync` if not on wasm or `fragile-send-sync-non-atomic-wasm` is enabled. By @wumpf in [#4818](https://github.com/gfx-rs/wgpu/pull/4818)
 - Align `wgpu_types::CompositeAlphaMode` serde serialization to spec. By @littledivy in [#4940](https://github.com/gfx-rs/wgpu/pull/4940)
 - Fix error message of `ConfigureSurfaceError::TooLarge`. By @Dinnerbone in [#4960](https://github.com/gfx-rs/wgpu/pull/4960)
+- Fixed a number of panics. by @nical in [#4999](https://github.com/gfx-rs/wgpu/pull/4999), [#5014](https://github.com/gfx-rs/wgpu/pull/5014), [#5024](https://github.com/gfx-rs/wgpu/pull/5024), [#5025](https://github.com/gfx-rs/wgpu/pull/5025), [#5026](https://github.com/gfx-rs/wgpu/pull/5026), [#5027](https://github.com/gfx-rs/wgpu/pull/5027), [#5028](https://github.com/gfx-rs/wgpu/pull/5028) and [#5042](https://github.com/gfx-rs/wgpu/pull/5042).
```
Wumpf, line 184: <3 <3

wgpu-core/src/command/render.rs
```diff
@@ -382,13 +385,16 @@ struct VertexState {
 
 impl VertexState {
     fn update_limits(&mut self) {
-        self.vertex_limit = u32::MAX;
-        self.instance_limit = u32::MAX;
+        // Ensure that the limits are always smaller than u32::MAX so that
+        // interger overlows can be prevented via saturating additions.
+        let max = u32::MAX - 1;
+        self.vertex_limit = max;
+        self.instance_limit = max;
```
Wumpf: isn't it valid to have a vertex limit of u32::MAX? Wouldn't that mean that the last vertex/instance isn't accounted for?

ErichDonGubler: I was also confused from the OP as to why `saturating_add` was selected, as opposed to `checked_add`. @nical, would that not be more appropriate?

nical: With this patch, having u32::MAX vertices would produce an error. It's a bit theoretical because you'll most likely hit the max buffer size before you manage to stuff u32::MAX vertices into the buffer.

wgpu-core/src/command/render.rs
```diff
@@ -1891,7 +1897,7 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
                         };
                         state.is_ready(indexed).map_pass_err(scope)?;
 
-                        let last_vertex = first_vertex + vertex_count;
+                        let last_vertex = first_vertex.saturating_add(vertex_count);
```
cwfitzgerald: More concretely, using saturating_add will result in the error message being incorrect - I feel like using `u64` for these intermediate calculations (and the vertex_limit) would be a lot cleaner.

cwfitzgerald (changes requested):

wgpu-core/src/command/render.rs
nical: `checked_add` forces you to spell out a separate check and return path for this whereas `saturating_add` was a simple way to just treat the  "range too big" case in one go without having to deal with overflow separately. It's also hot code so keeping the validation short has some value.

wgpu-core/src/command/render.rs
nical: You are worrying about an error message that you would only get if you are actively trying to break the system. I changed it to u64 anyway.

cwfitzgerald (approved):

wgpu-core/src/command/render.rs
cwfitzgerald: People can end up in these seemingly insane error cases when they're learning and doing the math all wrong, leading to passing crazy things to us - letting the user correlate the error message to a value in their code is very important in that case.

PR #5044, opened 2024-01-11
Allow WebGPU &  WebGL in same wasm and detect WebGPU availability

### LINKED ISSUE #2804: Dynamic WebGPU support detection

Wumpf:
**Connections**
* Based on #4984
* Fixes #2804

Shoutout to @daxpedda who did a lot of the groundwork to make this possible!

**Description**
Allows to compile with both webgpu & webgl backends enabled. There's a new `webgpu` feature flag that is enabled by default for this purpose. When additionally enabling `webgl` this will no longer disable `webgpu` and both features can live side by side.
The WebGPU backend remains a special case in that far that an instance decides (via very simple support detection logic) upon creation if it is a WebGPU instance or a WgpuCore instance. I did a bunch of renamings in that area to make this more obvious. Also, all `as_hal` methods now return an `Option` (most already did) to make it possible for them to return `None` whenever both WebGPU & WebGL are enabled (only a WebGL/WgpuCore instance can do hal conversion). Similar patter applies to a variety of things.

**DRAFT until #4984 lands**
(makes the diff also *significantly* smaller)

**Testing**
* clippy etc.
* local web tests to make sure they still do webgl when asked to (workarounds were needed, see code comments)
* there's an egui branch that works with this https://github.com/emilk/egui/tree/wgpu-update

image

**Checklist**

- [x] Run `cargo fmt`.


wgpu/src/backend/mod.rs
```diff
@@ -1,9 +1,9 @@
...
 #[cfg(webgpu)]
-pub(crate) use web::Context;
+pub(crate) use webgpu::{get_browser_gpu_property, ContextWebGpu};
 
-#[cfg(not(webgpu))]
-mod direct;
-#[cfg(not(webgpu))]
-pub(crate) use direct::Context;
+#[cfg(wgpu_core)]
+mod wgpu_core;
```
cwfitzgerald, line 7: This isn't a great name as it's a conflict I wonder if there's a better option

wgpu/src/lib.rs
```diff
@@ -1999,13 +2029,24 @@ impl Instance {
     }
 
     /// Generates memory report.
-    #[cfg(not(webgpu))]
+    ///
+    /// Like `try_generate_report` but panics if the active backend is WebGPU.
```
cwfitzgerald: Couldn't this just be optional?

cwfitzgerald (changes requested):
Need to see it rebased to fully review, but some initial comments

wgpu/src/backend/mod.rs
Wumpf: maybe just `core`? I found the previous `direct` to be super confusing back in the day I figured this is a good opportunity to rename. Arguably we can also take this discussion/change separately :)

wgpu/src/lib.rs
Wumpf: tbh I have the two versions mostly I found it too tedious to patch all existing `generate_report` calls. Can revisit

wgpu/src/lib.rs
```diff
@@ -1809,34 +1833,40 @@ impl Instance {
...
             context: Arc::new(unsafe {
-                crate::backend::Context::from_core_instance(core_instance)
+                crate::backend::ContextWgpuCore::from_core_instance(core_instance)
             }),
         }
     }
 
     /// Retrieves all available [`Adapter`]s that match the given [`Backends`].
     ///
+    /// Always returns an empty iterator if the instance decided upon creation to
```
```suggestion
    /// Always returns an empty vector if the instance decided upon creation to
```

would be nice to still return an iterator, but for that we'd need `itertools::Either`

wgpu/src/lib.rs
```diff
@@ -2987,10 +3035,13 @@ impl<'a> BufferSlice<'a> {
...
-    /// which unconditionally copies mapped data into the wasm heap, this function directly hands you the
-    /// ArrayBuffer that we mapped the data into in js.
-    #[cfg(webgpu)]
+    /// This is useful when targeting WebGPU and you want to pass mapped data directly to js.
+    /// Unlike `get_mapped_range` which unconditionally copies mapped data into the wasm heap,
+    /// this function directly hands you the ArrayBuffer that we mapped the data into in js.
+    ///
+    /// If you are not targeting WebGPU, this function will perform an additional copy from the wasm heap to
+    /// the returned array buffer.
+    #[cfg(all(not(native), not(target_os = "emscripten")))]
```
```suggestion
    #[cfg(any(webgpu, webgl))]
```

wgpu/src/lib.rs
Wumpf: remove the extra report method again, wasn't _that_ bad

wgpu/src/backend/wgpu_core.rs
```diff
@@ -1537,6 +1539,19 @@ impl crate::Context for Context {
...
+        &self,
+        buffer: &Self::BufferId,
+        buffer_data: &Self::BufferData,
+        sub_range: Range<wgt::BufferAddress>,
+    ) -> js_sys::ArrayBuffer {
+        let mapped_range = self.buffer_get_mapped_range(buffer, buffer_data, sub_range);
+        let array_buffer = js_sys::Uint8Array::new_with_length(mapped_range.slice().len() as u32);
+        array_buffer.copy_from(mapped_range.slice());
+        array_buffer.buffer()
+    }
```
cwfitzgerald: I don't think this is correct - the changes to the array buffer won't be reflected in the upstream changes.

cwfitzgerald (changes requested):
Looks good outside of a issue with get_mapped_range

wgpu/src/backend/wgpu_core.rs
Wumpf: ah you're right. We'd need a special unmap operation then! We'll have to make this function webgpu only and thus return an `Option` in the dynamic interface provided by `instance`

wgpu/src/backend/mod.rs
Wumpf: changed your mind and just going with `wgpu_core` as module name despite the clash with the feature? 🤷 fine by me either way

wgpu/src/backend/wgpu_core.rs
Wumpf: addressed in [6c10575](https://github.com/gfx-rs/wgpu/pull/5044/commits/6c105754e92da2997aeb15d461815aeaec94f2da)

cwfitzgerald (approved):
Hell yes!

cwfitzgerald:
@Wumpf please resolve conflicts, then merge

Wumpf:
woo 🥳

Keavon:
This looks very handy! Is there a documented list of things the WebGL version doesn't support? I assume compute shaders is one such obvious missing feature (although I wonder if there's been any effort put towards polyfilling compute shaders with fragment shader magic?).

Wumpf:
@Keavon there's unfortunately no written out list since it's a bit hard to capture - some of the exact capabilities are determined at runtime. But you can check https://docs.rs/wgpu/latest/wgpu/struct.DownlevelFlags.html# for possible lacking features as well as https://docs.rs/wgpu/latest/wgpu/struct.Limits.html#method.downlevel_webgl2_defaults which lists out the reduced limits of webgl directly.
As you guessed there's no compute shaders and no effort to polyfill them - this would be huge undertaking and likely run into issues that would make you resort to cpu based emulation (most prominently there's no random access writes at all which bars the door to emulating the harder stuff like shared memory and don't even get me started about atomics ;))

cwfitzgerald:
It's almost certainly easier to polyfill them _in the user's application_ using application specific knowledge if you need them.

PR #5045, opened 2024-01-11
More eagerly destroy buffers and texture on the web backend

nical:
**Description**

Since the web backend is run in a garbage collected environment, it is important to eagerly destroy resources when we know they won't be used anymore (when the handle is dropped) rather that wait for GC to kick off and get the dropped resource deallocated. Especially since the GC doesn't know about gpu memory pressure.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu/src/backend/web.rs
+++ b/wgpu/src/backend/web.rs
@@ -2098,7 +2098,7 @@ impl crate::context::Context for Context {
     }
 
     fn buffer_drop(&self, _buffer: &Self::BufferId, _buffer_data: &Self::BufferData) {
-        // Dropped automatically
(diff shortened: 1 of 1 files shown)


cwfitzgerald:
This unfortunately doesn't work - you can access buffers after drop through bind groups, command buffers and such. We would need to track the dependencies ourself in the bindings to be able to auto-call destroy. This is something the emscripten might actually do, so its something to consider.

nical:
Ouch, I hadn't thought of that. It's a shame, the current situation is rather bad.

PR #5050, opened 2024-01-12
Triage suspected resources after triaging submissions

### LINKED ISSUE #4917: Resources should need only one submission to be released

nical:
**Connections**

Fixes #4917

**Description**

`triage_submissions` puts some resources in the suspected list. These currently need to wait for the next `Device::maintain` but they could be released right away if the order is swapped which is what this PR basically does.

I woudldn't be surprised if the order is intentional and the story is more complicated (hopefully not), in any case @cwfitzgerald this PR also swaps the order of the temp_suspected and life locks, I know you auditted these recently so let me know if this might cause deadlocks.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/tests/tests/mem_leaks.rs
+++ b/tests/tests/mem_leaks.rs
@@ -243,10 +243,6 @@ fn draw_test_with_reports(
 
     ctx.device
         .poll(wgpu::Maintain::WaitForSubmissionIndex(submit_index));
-    // Because of dependency between resources, A resource being destroyed during poll
-    // can cause another resource to be ready for destruction next time poll is called.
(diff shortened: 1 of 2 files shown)


cwfitzgerald (approved):
There's no reason I know for this to be the way it is - I don't think there are any ordering problems doing this.

PR #5051, opened 2024-01-12
Add Proper Timeouts to CI/nextest

cwfitzgerald:
Part of splitting up #5012 into reviewable parts.

Adding timeouts to CI in case we get jobs that take forever.


Wumpf (approved):
nice! :)

PR #5052, opened 2024-01-12
Fix DX12 Wait with Timeout

cwfitzgerald:
Part of splitting up https://github.com/gfx-rs/wgpu/pull/5012 into reviewable parts.

There was an issue in DX12 where we don't properly deal with the signalling patterns of a windows event if previous Polls called timeout. Further explanation in-code in a comment.

This is tested by #5012.

### DIFF
--- a/wgpu-hal/src/dx12/device.rs
+++ b/wgpu-hal/src/dx12/device.rs
@@ -5,7 +5,13 @@ use crate::{
 
 use super::{conv, descriptor, view};
 use parking_lot::Mutex;
-use std::{ffi, mem, num::NonZeroU32, ptr, sync::Arc};
+use std::{
+    ffi, mem,
+    num::NonZeroU32,
+    ptr,
+    sync::Arc,
+    time::{Duration, Instant},
+};
 use winapi::{
     shared::{dxgiformat, dxgitype, minwindef::BOOL, winerror},
     um::{d3d12 as d3d12_ty, synchapi, winbase},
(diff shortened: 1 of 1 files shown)


wgpu-hal/src/dx12/device.rs
```diff
@@ -1535,19 +1541,80 @@ impl crate::Device<super::Api> for super::Device {
...
+            // We need to explicitly use checked_sub. Overflow with duration panics, and if the
+            // timing works out just right, we can get a negative remaining wait duration.
+            //
+            // This happens when a previous iteration WaitForSingleObject succeeded with a previous fence value,
+            // right before the timeout would have been hit.
+            let remaining_wait_duration = match timeout_duration.checked_sub(elapsed) {
+                Some(remaining) => remaining,
+                None => {
+                    log::trace!("Timeout elapsed inbetween waits!");
+                    break Ok(false);
```
Wumpf, line 1582: I'm a bit uneasy about potentially returning before even attempting to call `WaitForSingleObject`. This means that in theory we could return false despite having hit the fence within the timeout.
This could be remodeled a little bit to wait with a zero timeout in those cases.

Wumpf (approved):
makes a lot of sense with the explanation in the comment! Minor wrinkle pointed out in the comments, leaving that to you

wgpu-hal/src/dx12/device.rs
cwfitzgerald: It's all ultimately just a race, either way you spin it you can end up with the fence being hit without this returning true. In the code of #5012, after we wait we query the fence value, so we're always using the latest fence value no matter what.

PR #5053, opened 2024-01-12
Convert Tests to Use Async Poll

cwfitzgerald:
Part of splitting up https://github.com/gfx-rs/wgpu/pull/5012 into reviewable parts.

In 5012, `poll(Wait)` no longer works on WebGL. As such, we need to switch to an async method of polling the device. This PR stubs out the async poll and polling functions needed, while not actually changing the behavior at all.

This is a fairly major part of 5012, and made the diff hard to read due to how sweeping the changes are. None of this, besides the stubbing, should be controversial.

### DIFF
--- a/examples/src/repeated_compute/mod.rs
+++ b/examples/src/repeated_compute/mod.rs
@@ -108,7 +108,7 @@ async fn compute(local_buffer: &mut [u32], context: &WgpuContext) {
     // In order for the mapping to be completed, one of three things must happen.
     // One of those can be calling `Device::poll`. This isn't necessary on the web as devices
     // are polled automatically but natively, we need to make sure this happens manually.
-    // `Maintain::Wait` will cause the thread to wait on native but not the web.
+    // `Maintain::Wait` will cause the thread to wait on native but not on WebGpu.
     context
         .device
         .poll(wgpu::Maintain::wait())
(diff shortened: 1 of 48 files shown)


wgpu/src/lib.rs
```diff
@@ -2205,16 +2206,16 @@ impl Adapter {
...
     ///
     /// Return `true` if the queue is empty, or `false` if there are more queue
     /// submissions still in flight. (Note that, unless access to the [`Queue`] is
     /// coordinated somehow, this information could be out of date by the time
     /// the caller receives it. `Queue`s can be shared between threads, so
     /// other threads could submit new work at any time.)
     ///
-    /// On the web, this is a no-op. `Device`s are automatically polled.
-    pub fn poll(&self, maintain: Maintain) -> bool {
+    /// When running on WebGPU, this is a no-op. `Device`s are automatically polled.
```
kpreid, line 2217: Should the text of `Maintain`'s variant documentation also be changed to match?

cwfitzgerald: Adjusted those docs.

examples/src/repeated_compute/mod.rs
```diff
@@ -109,7 +109,10 @@ async fn compute(local_buffer: &mut [u32], context: &WgpuContext) {
     // One of those can be calling `Device::poll`. This isn't necessary on the web as devices
     // are polled automatically but natively, we need to make sure this happens manually.
     // `Maintain::Wait` will cause the thread to wait on native but not the web.
```
Wumpf: since you nicely updated it elsewhere

```suggestion
    // `Maintain::Wait` will cause the thread to wait on native but not on WebGpu.
```

Wumpf (approved):
thanks for splitting all of this out. lgtm!

wgpu/src/lib.rs
cyqsimon: The return type was changed too but the doc comments were never updated. It's not obvious to a user like me how the variants of `MaintainResult` maps to `bool` pre-change.

PR #4828, opened 2023-12-04
Remove DX11 backend

valaphee:
**Connections**
https://github.com/gfx-rs/wgpu/issues/4797

**Description**
This backend had no functionality, and is mainly just boilerplate.

**Testing**

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`. If applicable, add:
  - [x] `--target wasm32-unknown-unknown`
  - [x] `--target wasm32-unknown-emscripten`
- [x] Run `cargo xtask test` to run tests.
- [x] Add change to `CHANGELOG.md`. See simple instructions inside file.

Also updated etc/big-picture.png, and all references should be removed except in the changelog, logically.

PR #5055, opened 2024-01-12
Add `cfg_aliases` to `wgpu-core` and `wgpu-hal`

daxpedda:
Follow up to #4935 as requested.

### DIFF
--- a/wgpu-hal/src/auxil/dxgi/factory.rs
+++ b/wgpu-hal/src/auxil/dxgi/factory.rs
@@ -7,8 +7,6 @@ use super::result::HResult as _;
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub enum DxgiFactoryType {
-    #[cfg(feature = "dx11")]
-    Factory1,
     Factory2,
     Factory4,
     Factory6,
--- a/wgpu-hal/src/gles/mod.rs
+++ b/wgpu-hal/src/gles/mod.rs
@@ -82,11 +82,12 @@ we don't bother with that combination.
 */
 
 ///cbindgen:ignore
-#[cfg(not(any(windows, all(target_arch = "wasm32", not(target_os = "emscripten")))))]
+#[cfg(not(any(windows, webgl)))]
 mod egl;
-#[cfg(target_os = "emscripten")]
+#[allow(clippy::mismatched_target_os)]
+#[cfg(emscripten)]
 mod emscripten;
-#[cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))]
+#[cfg(webgl)]
 mod web;
 #[cfg(windows)]
 mod wgl;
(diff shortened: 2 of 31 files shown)


wgpu-hal/src/auxil/dxgi/factory.rs
```diff
@@ -7,8 +7,6 @@ use super::result::HResult as _;
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub enum DxgiFactoryType {
-    #[cfg(feature = "dx11")]
-    Factory1,
```
daxpedda, line 11: I assume this was just missed in #4828?

wgpu-hal/src/gles/mod.rs
```diff
@@ -82,11 +82,12 @@ we don't bother with that combination.
 */
 
 ///cbindgen:ignore
-#[cfg(not(any(windows, all(target_arch = "wasm32", not(target_os = "emscripten")))))]
+#[cfg(not(any(windows, webgl)))]
 mod egl;
-#[cfg(target_os = "emscripten")]
+#[allow(clippy::mismatched_target_os)]
```
daxpedda: Apparently this lint is a bit buggy, putting it in root or on top of the file wouldn't help.

I'm happy to either rename `emscripten` to something else or just use `target_os = "emscripten"` in places where it makes sense (e.g. we are already behind a `feature = "gles"` guard).

wgpu-hal/src/auxil/dxgi/factory.rs
Wumpf: looks like! 👍

wgpu-hal/src/gles/mod.rs
Wumpf: Yeah I hit this before as well. I think it would be best if we could come up with a different name for the emscripten alias since the clippy lint is otherwise quite useful. But I'm not sure to what exactly 🤔 . `emscripten_wasm` or something like that maybe (with an explanation in the build.rs to why)

Wumpf (dismissed):
love it, thank you so much!
Let's try to remove `#![allow(clippy::mismatched_target_os)]` (see comment) and then 🚢

wgpu-hal/src/gles/mod.rs
daxpedda: How about `Emscripten`?
Would also make it more beautiful for the user: in the documentation :smiley:.

Wumpf: didn't we discuss casing already on the previous PR and concluded lowercase 😄. Ah well fine by me :)

daxpedda: Yes, I thought I should try again :sweat_smile:!
Will get it done soon!

For reference: https://github.com/gfx-rs/wgpu/pull/4935#discussion_r1436658697.
I also left a comment at Rustc for a renaming feature: https://github.com/rust-lang/rust/issues/43781#issuecomment-1888878314.

daxpedda: Done.

cwfitzgerald (approved):
code go bye bye!

PR #5056, opened 2024-01-12
Staged CI

cwfitzgerald:
Splits test running to two stages, build and run. So expensive gpu machines only need to run the test, not build it too.

Currently blocked on a few llvm-cov issues:
- https://github.com/taiki-e/cargo-llvm-cov/issues/333
- https://github.com/taiki-e/cargo-llvm-cov/issues/334


cwfitzgerald:
Closing as #5165 means we are no longer paying for CI.

PR #5059, opened 2024-01-14
Fix cases in which instance creation may not create any context

Wumpf:
**Connections**
* Small follow-up to #5044

**Description**
See description:

* When wgpu_core isn't enabled, don't detect lack of webgpu support right away, this would lead to reaching an `unrechable!` whenever webgpu isn't supported by the browser
* `any_backend_feature_enabled` didn't take emscripten into account

**Testing**
Hit this when running an application with various feature combinations off latest trunk, this resolves it!

**Checklist**

- [x] Run `cargo fmt`.
- ~[ ] Run `cargo xtask test` to run tests.~
- ~[ ] Add change to `CHANGELOG.md`. See simple instructions inside file.~

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -1718,6 +1718,7 @@ impl Instance {
     ///
     /// Which feature makes this method return true depends on the target platform:
     /// * MacOS/iOS: `metal`, `vulkan-portability` or `angle`
+    /// * Wasm32: `webgpu`, `webgl` or target emscripten.
     /// * All other: Always returns true
     ///
     /// TODO: Right now it's otherwise not possible yet to opt-out of all features on most platforms.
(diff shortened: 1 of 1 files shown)


wgpu/src/lib.rs
```diff
@@ -1771,8 +1772,10 @@ impl Instance {
         }
 
         #[cfg(webgpu)]
-        if _instance_desc.backends.contains(Backends::BROWSER_WEBGPU)
-            && crate::backend::get_browser_gpu_property().map_or(false, |gpu| !gpu.is_undefined())
+        if !cfg!(wgpu_core) // If wgpu-core isn't enabled we can't use anything else.
+            || (_instance_desc.backends.contains(Backends::BROWSER_WEBGPU)
+                && crate::backend::get_browser_gpu_property()
+                    .map_or(false, |gpu| !gpu.is_undefined()))
```
cwfitzgerald: Could we split these up into 3 named variables (within the webgpu block) to make this a bit less logic soup?

cwfitzgerald (changes requested):

wgpu/src/lib.rs
```diff
@@ -1718,6 +1718,7 @@ impl Instance {
     ///
     /// Which feature makes this method return true depends on the target platform:
     /// * MacOS/iOS: `metal`, `vulkan-portability` or `angle`
+    /// * Wasm32: `webgpu`, `webgl` or target emscripten.
```
daxpedda:
```suggestion
    /// * Wasm32: `webgpu`, `webgl` or Emscripten target.
```

cwfitzgerald (approved):

PR #5063, opened 2024-01-15
Re-export public dependencies

### LINKED ISSUE #5054: Re-Export All Public Dependencies
As raised in https://github.com/gfx-rs/wgpu/issues/4802, it is useful to have all public dependencies as a re-export so people can use the same versions without needing to name them directly in the Cargo.toml

cwfitzgerald:
Closes #5054.

Useful enough to have, and doesn't clutter things too much.

This re-organizes our naga dependency a bit, notably that it's now properly optional on wasm, only enabled if you enable a feature that requires transpilation, or a direct mention in our public interface.

image

### DIFF
--- a/wgpu/Cargo.toml
+++ b/wgpu/Cargo.toml
@@ -45,7 +45,7 @@ angle = ["wgc?/gles"]
 vulkan-portability = ["wgc?/vulkan"]
 
 ## Enables the WebGPU backend on Wasm. Disabled when targeting `emscripten`.
-webgpu = []
+webgpu = ["naga?/wgsl-out"]
 
 ## Enables the GLES backend on Wasm
 ##
@@ -63,6 +63,9 @@ glsl = ["naga/glsl-in"]
 
 ## Enable accepting WGSL shaders as input.
 wgsl = ["wgc?/wgsl"]
+
+## Enable accepting naga IR shaders as input.
+naga-ir = ["naga"]
 
 #! ### Logging & Tracing
 # --------------------------------------------------------------------
@@ -177,10 +180,6 @@ cfg_aliases.workspace = true
 [dev-dependencies.naga]
 workspace = true
 features = ["wgsl-in"]
-
-[target.'cfg(target_arch = "wasm32")'.dependencies.naga]
-workspace = true
-features = ["wgsl-out"]
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 web-sys = { workspace = true, features = [
--- a/wgpu/src/backend/webgpu.rs
+++ b/wgpu/src/backend/webgpu.rs
@@ -1400,7 +1400,7 @@ impl crate::context::Context for ContextWebGpu {
             feature = "spirv",
             feature = "glsl",
             feature = "wgsl",
-            feature = "naga"
+            feature = "naga-ir"
         )),
         allow(unreachable_code, unused_variables)
     )]
@@ -1411,7 +1411,7 @@ impl crate::context::Context for ContextWebGpu {
         desc: crate::ShaderModuleDescriptor<'_>,
         _shader_bound_checks: wgt::ShaderBoundChecks,
     ) -> (Self::ShaderModuleId, Self::ShaderModuleData) {
-        let mut descriptor = match desc.source {
+        let mut descriptor: web_sys::GpuShaderModuleDescriptor = match desc.source {
             #[cfg(feature = "spirv")]
             crate::ShaderSource::SpirV(ref spv) => {
                 use naga::{back, front, valid};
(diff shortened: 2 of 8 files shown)


cwfitzgerald:
@Wumpf @daxpedda could you check my math on the cfgs - I think this is a better arrangement and pass my various checks, but need a double check

wgpu/src/backend/webgpu.rs
```diff
@@ -1400,7 +1400,7 @@ impl crate::context::Context for ContextWebGpu {
             feature = "spirv",
             feature = "glsl",
             feature = "wgsl",
-            feature = "naga"
+            feature = "naga-ir"
```
Wumpf, line 1403: this should check the new alias `naga` you just introduced: `not(any(naga, feature = "wgsl"))`

wgpu/Cargo.toml
```diff
@@ -64,6 +64,9 @@ glsl = ["naga/glsl-in"]
 ## Enable accepting WGSL shaders as input.
```
Wumpf, line 64: while we're here we should probably make a note how this interacts with `webgpu` feature: unless `wgsl` is enabled only the `webgpu` backend can accept wgsl

wgpu/src/lib.rs
```diff
@@ -100,15 +101,51 @@ pub use wgt::{
...
+#[cfg(wgpu_core)]
+#[doc(inline)]
+pub use ::hal;
+
+/// Re-export of our `naga` dependency.
+///
+#[cfg(wgpu_core)]
+#[cfg_attr(
+    docsrs,
+    doc(cfg(any(wgpu_core, feature = "glsl", feature = "spirv", feature = "naga-ir",)))
```
Wumpf: why not use `any(wgpu_core, naga)`

wgpu/src/lib.rs
```diff
@@ -100,15 +101,51 @@ pub use wgt::{
...
+)]
+#[doc(inline)]
+// We re-export wgpu-core's re-export of naga, as we may not have direct access to it.
+pub use ::wgc::naga;
+/// Re-export of our `naga` dependency.
+///
+#[cfg(all(not(wgpu_core), naga))]
+#[cfg_attr(
+    docsrs,
+    doc(cfg(wgpu_core, feature = "glsl", feature = "spirv", feature = "naga-ir"))
```

wgpu/src/lib.rs
Wumpf: not quite following why we don't want doc inline otherwise, but my understanding on what it does is a bit shady

wgpu/src/util/init.rs
```diff
@@ -2,13 +2,13 @@ use wgt::{Backends, PowerPreference, RequestAdapterOptions};
...
+#[cfg(wgpu_core)]
 #[cfg_attr(docsrs, doc(cfg(all())))]
 pub use wgc::instance::parse_backends_from_comma_list;
-/// Always returns WEBGPU on wasm over webgpu.
-#[cfg(webgpu)]
+/// Just return ALL, if wgpu_core is not enabled.
+#[cfg(not(wgpu_core))]
 pub fn parse_backends_from_comma_list(_string: &str) -> Backends {
-    Backends::BROWSER_WEBGPU
+    Backends::all()
```
Wumpf, line 11: nice catch!
But at this point we probably should do the actual parsing here as well. Too unrelated though, for another time

Wumpf (approved):
cfgs all look right to me, just a few cosmetic comments
this will be super useful!

wgpu/src/util/init.rs
cwfitzgerald: Yeah we probably should move the function into wgpu-types or something, as it's used by deno too

wgpu/Cargo.toml
cwfitzgerald: Actually there's no special behavior - if the webgl feature isn't available, you can't construct a wgsl variant at all.

wgpu/src/lib.rs
cwfitzgerald: Basically there's no good intuition on what "naga" represents to the outside user. wgpu_core is also a bit weird, but it's at least more clear.

daxpedda (commented):
`cfg`s seem fine to me.
I think renaming the documented `cfg` with `doc(cfg(...))` like you did is a good idea.

cwfitzgerald:
I wonder if it's worth documenting the feature shorthands in the lib.rs file so the user knows what to expect and we don't need to rename all the shorthands

daxpedda:
> I wonder if it's worth documenting the feature shorthands in the lib.rs file so the user knows what to expect and we don't need to rename all the shorthands

I believe this would be an improvement in any case :+1:.

wgpu/Cargo.toml
Wumpf: but if you enable `webgpu` and `webgl` then you can pass in wgsl but only to the webgpu backend unless you also enable the `wgsl` feature, correct?

cwfitzgerald: No, to able to use wgsl at all, you need the cfg:

https://github.com/gfx-rs/wgpu/blob/trunk/wgpu%2Fsrc%2Flib.rs#L669-L670

Wumpf: ah nice, consistency ✨

wgpu/src/backend/webgpu.rs
cwfitzgerald: Eh, this is basically not the same thing - I'm not predicated on naga, just that none of the following features have been enabled.

PR #5065, opened 2024-01-15
d3d12: Propagate descriptor heap and handle allocation errors

nical:
**Connections**

We are getting some crashes related to that in CTS runs ([example](https://treeherder.mozilla.org/logviewer?job_id=443130268&repo=try)).
This will turn the crashes into failures which is only slightly better, but we'll get more info out of it.

**Description**

Adds some missing error propagation in the d3d12 backend.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


teoxoy (approved):
Good stuff!

PR #5068, opened 2024-01-16
Improve test feedback and implementation

udoprog:
**Description**
I wanted to include the image path to improve test ergonomics (in the excellent test framework) when `compare_image_output` failed, and when doing so an output like this is generated:

```
[2024-01-16T07:32:51Z ERROR wgpu_test::expectations] Unexpected failure due to: Panic(Some("Image data mismatch: D:\\Repo\\wgpu\\examples/..//examples/src/water\\screenshot-dx12-Microsoft_Basic_Render_Driver--difference.png"))
```

While this is mostly fine, most UI terminals allow you to ctrl+click path-looking items to open it. However, formatting the path like this prevents it from working.

I changed the output so that the message is reported in isolation, or `*no message*` if no message is available, giving us this if a path is included in the tail like with the change I included:

```
[2024-01-16T07:35:42Z ERROR wgpu_test::expectations] Panic: Image data mismatch: D:\Repo\wgpu\examples/..//examples/src/water\screenshot-dx12-Microsoft_Basic_Render_Driver--difference.png
```

Allowing the output path to be ctrl-clicked.

In the process of doing so, I simplified how the `FailureResult` / `FailureReason` log diagnostics and are matched against each other since I noticed it.

**Testing**
Ran `cargo xtask test`.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`

Internal change, so did not modify `CHANGELOG.md`. But will do so if it's still appropriate!

### DIFF
--- a/tests/src/expectations.rs
+++ b/tests/src/expectations.rs
@@ -250,23 +250,16 @@ impl FailureCase {
     /// Returns true if the given failure "satisfies" this failure case.
     pub(crate) fn matches_failure(&self, failure: &FailureResult) -> bool {
         for reason in self.reasons() {
-            let mut result = true;
+            let kind_matched = reason.kind.map_or(true, |kind| kind == failure.kind);
 
-            result &= reason.kind.map_or(true, |kind| kind == failure.kind);
+            let message_matched =
+                reason
+                    .message
+                    .map_or(true, |message| matches!(&failure.message, Some(actual) if actual.to_lowercase().contains(&message.to_lowercase())));
 
-            result &= reason
-                .message
-                .map_or(true, |message| match failure.message.as_deref() {
-                    Some(actual) => actual.to_lowercase().contains(&message.to_lowercase()),
-                    None => false,
-                });
-
-            if result {
+            if kind_matched && message_matched {
                 let message = failure.message.as_deref().unwrap_or("*no message*");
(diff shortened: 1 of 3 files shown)


tests/src/expectations.rs
```diff
@@ -247,41 +250,21 @@ impl FailureCase {
...
+            let mut result = true;
+
+            result &= reason.kind.map_or(true, |kind| kind == failure.kind);
+
+            result &= reason
+                .message
+                .map_or(true, |message| match failure.message.as_deref() {
+                    Some(actual) => actual.to_lowercase().contains(&message.to_lowercase()),
+                    None => false,
+                });
```
cwfitzgerald: I find this extremely hard to follow - it's basically just logic soup to me. I definitely agree that what we had previously wasn't good either and this is moving in the right direction, but could we possibly use explicit named conditions which are then combined into the final value.

cwfitzgerald (changes requested):
Like the change in general, have one main concern.

cwfitzgerald:
@udoprog make sure to re-request the review of whoever reviewed the PR to make sure we actually see it!

cwfitzgerald (approved):
Now that is what I'm talking about!

PR #5069, opened 2024-01-16
Improve ergonomics of gfx_select!

udoprog:
**Description**
Improves the ergonomics of `gfx_select!` so that the more common `self.<field>` expression can be used directly with it.

**Testing**
Compile tested for the platforms I have available, but since this only affects code in the hal layer, that should cover most. I'm not sure how to compile test deno, so I just did the best I could and hopefully CI can try it out for me.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

Internal change, so did not modify `CHANGELOG.md`. But will do so if it's still appropriate!

### DIFF
--- a/deno_webgpu/binding.rs
+++ b/deno_webgpu/binding.rs
@@ -21,8 +21,7 @@ impl Resource for WebGpuBindGroupLayout {
     }
 
     fn close(self: Rc<Self>) {
-        let instance = &self.0;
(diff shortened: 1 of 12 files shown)


Wumpf (approved):
looks a lot nicer 🤩 . Thank you!

PR #5070, opened 2024-01-16
CI: Collect basic checks in one place and add concurrency

udoprog:
**Description**

CI improvements:
* Collect basic checks under a single job named `basics`.
* Add `concurrency` setting which ensures that existing jobs are cancelled if a commit related to a pull request is changed.

This continues to reduce resource use on Github in case there's a bit of churn in a PR or if a basic compliance error occurs.

Note that it would also be nice to break up `check` a bit more.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,13 @@ on:
   pull_request:
   merge_group:
 
+# Ensures that only one job is allowed to run at any given time for a pull
+# request (or commit sha). Existing jobs if the PR for example is rebased or
+# force pushed are cancelled.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
 env:
   #
   # Dependency versioning
@@ -74,6 +81,16 @@ env:
 # It adds overhead to the build and another point of failure.
 
 jobs:
+  # List of basic checks that needs to pass before we fire heavier jobs.
+  basics:
+    runs-on: ubuntu-latest
+    needs:
+      - fmt
+      - doctest
+      - check
+    steps:
+      - run: exit 0
+
   check:
     # runtime is normally 2-8 minutes
     timeout-minutes: 15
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -7,6 +7,13 @@ on:
   pull_request:
   merge_group:
 
+# Ensures that only one job is allowed to run at any given time for a pull
+# request (or commit sha). Existing jobs if the PR for example is rebased or
+# force pushed are cancelled.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
```
cwfitzgerald, line 15: Ah! This is nice - I was thinking about adding this, but I wanted multiple merges into main to each get their own job. I like the "PR | sha" form to solve this problem

.github/workflows/ci.yml
```diff
@@ -74,6 +81,16 @@ env:
...
 jobs:
+  # List of basic checks that needs to pass before we fire heavier jobs.
+  basics:
+    runs-on: ubuntu-latest
+    needs:
+      - fmt
+      - doctest
+      - check
+    steps:
+      - run: exit 0
```
cwfitzgerald, line 92: The only jobs that we need to be blocked on this is the `test` build, the rest can fire on failure. All the regular runners we can just fire off aggressively, trying to reduce CI latency.

.github/workflows/ci.yml
```diff
@@ -391,7 +411,7 @@ jobs:
 
     name: Test ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
-    needs: [check]
+    needs: basics
```
cwfitzgerald, line 414: i.e. this can be [check, fmt, doctest] - the rest should be left as is.

cwfitzgerald (changes requested):

cwfitzgerald:
> Note that it would also be nice to break up check a bit more.

#5056 removes the dependency on check for testing (only test build success), so we probably don't need to break it up

udoprog:
I'll close this in favor of #5056? You can just pull in the changes you want manually so we don't have to clobber each other, like with the concurrency setting.

PR #5073, opened 2024-01-16
Introduce error helpers in wgpu-core

udoprog:
**Description**

I noted that wgpu crates rely on two error patterns which can be improved through methods which are fairly common in the Rust ecosystem:

The first one is:

```rust
return Err(SomeLocalError.into());
```

And the second is testing a precondition which if it fails return an error:

```rust
if !condition {
    return Err(SomeLocalError.into());
}
```

This PR introduces two local error helpers in wgpu-core, who's behavior are borrowed from the popular [`anyhow` crate](https://docs.rs/anyhow/latest/anyhow/macro.bail.html):

* `bail!` which takes an expression, converts it through `From` and returns it as an error.
* `ensure!` which takes two arguments, and if the first one is false does the same as `bail!`.

There are two potential ergonomics wins here. First uses do not have to worry about local `.into()` conversions and `ensure!` can be more concise (all though if the line is broken up isn't).

I've introduced a few example usages of each as a proposal here to garner feedback, and will mark this as a draft for now.

### DIFF
--- a/wgpu-core/src/command/clear.rs
+++ b/wgpu-core/src/command/clear.rs
@@ -228,9 +228,10 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
         }
 
         let device = &cmd_buf.device;
-        if !device.is_valid() {
-            return Err(ClearError::InvalidDevice(cmd_buf.device.as_info().id()));
-        }
+        ensure!(
+            device.is_valid(),
+            ClearError::InvalidDevice(cmd_buf.device.as_info().id())
+        );
         let (encoder, tracker) = cmd_buf_data.open_encoder_and_tracker()?;
 
         clear_texture(
--- a/wgpu-core/src/command/query.rs
+++ b/wgpu-core/src/command/query.rs
@@ -438,18 +438,18 @@ impl<G: GlobalIdentityHandlerFactory> Global<G> {
 
         let dst_barrier = dst_pending.map(|pending| pending.into_hal(&dst_buffer, &snatch_guard));
 
-        if !dst_buffer.usage.contains(wgt::BufferUsages::QUERY_RESOLVE) {
-            return Err(ResolveError::MissingBufferUsage.into());
-        }
+        ensure!(
+            dst_buffer.usage.contains(wgt::BufferUsages::QUERY_RESOLVE),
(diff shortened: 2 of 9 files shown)


nical (commented):
Hi, Sorry, this is probably mostly a matter of taste, we'll see how the other maintainers feel about this, but I'm not a fan.

 - I don't like how it hides control flow.
 - I generally don't like relying on macros unless they bring something big to the table, this is not the case.
 - It's just as verbose or marginally smaller as the straightforward "regular rust" version.

udoprog:
> It's just as verbose or marginally smaller as the straightforward "regular rust" version.

Well, [`bail!` does avoid this awkward rustfmt](https://github.com/gfx-rs/wgpu/pull/5073/files#diff-d33e2d791fcf615a0ae2e4c3c22c0977e7c6d0d5681bb1904a50523a57859bbcL968). And [`ensure!` is tighter](https://github.com/gfx-rs/wgpu/pull/5073/files#diff-d33e2d791fcf615a0ae2e4c3c22c0977e7c6d0d5681bb1904a50523a57859bbcR1008) when the line doesn't wrap.

nical:
I discussed this with some of the other maintainers and the general sentiment is that the gains are not worth the extra mental overhead of hiding control flow and using macros over plain rust, so we are not going to pursue this direction.

Thanks for proposing it, though. I can understand why some people might prefer the changes you are proposing. That's not a judgement of whether your approach is right or wrong, just an expression of our preferences and the way we are used to reading code.

udoprog:
I'm OK with that, but I do feel a bit dissatisfied that the discussion was held somewhere private where I couldn't participate rather than here where I raised the question.

Wumpf:
> I'm OK with that, but I do feel a bit dissatisfied that the discussion was held somewhere private where I couldn't participate rather than here where I raised the question.

good point, let's improve that and not make it a bad habit! cc: @cwfitzgerald

nical:
Thanks for your understanding. Apologies about the private discussion. I've had a few bad experiences with "matter of tastes" debates that go out of control in public venues so I wanted get a sense of the maintainers stance before considering opening the floodgates. Since the sentiment was shared I didn't open the question for further debate. I  am sorry for the frustration that caused.

PR #5077, opened 2024-01-17
Remove many uses of Option<T> and unwraps

udoprog:
**Description**
This is another internal refactoring I'd like to suggest.

I noticed that if we make the hal `destroy_*` functions take mutable references instead of owned values, we can avoid many internal uses of `Option<T>` effectively making the hal types more "drop friendly". Most hal types perform this kind of accounting internally where needed so the added `raw: Option<T>` does not seem to achieve anything important. To me this seems like a worthwhile tradeoff. These types are primarily used in destructors, so making this easier should be worthwhile.

The end result of this round is the removal of roughly 100 unwraps. And a handful of instances where we had to move the Option<T> into the backends themselves. This is mainly due to how `gpu-alloc` and the `gpu-descriptor` crates work in that they need to take owned values.

**Testing**
Tested locally and running the test suite.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -449,7 +449,7 @@ pub type BindGroupLayouts<A> = crate::storage::Storage<BindGroupLayout<A>, BindG
 /// Bind group layout.
 #[derive(Debug)]
 pub struct BindGroupLayout<A: HalApi> {
-    pub(crate) raw: Option<A::BindGroupLayout>,
+    pub(crate) raw: A::BindGroupLayout,
     pub(crate) device: Arc<Device<A>>,
     pub(crate) entries: bgl::EntryMap,
     /// It is very important that we know if the bind group comes from the BGL pool.
@@ -470,17 +470,16 @@ impl<A: HalApi> Drop for BindGroupLayout<A> {
         if matches!(self.origin, bgl::Origin::Pool) {
             self.device.bgl_pool.remove(&self.entries);
         }
-        if let Some(raw) = self.raw.take() {
-            #[cfg(feature = "trace")]
-            if let Some(t) = self.device.trace.lock().as_mut() {
-                t.add(trace::Action::DestroyBindGroupLayout(self.info.id()));
-            }
 
-            resource_log!("Destroy raw BindGroupLayout {:?}", self.info.label());
-            unsafe {
-                use hal::Device;
-                self.device.raw().destroy_bind_group_layout(raw);
-            }
+        #[cfg(feature = "trace")]
+        if let Some(t) = self.device.trace.lock().as_mut() {
+            t.add(trace::Action::DestroyBindGroupLayout(self.info.id()));
+        }
+
+        resource_log!("Destroy raw BindGroupLayout {:?}", self.info.label());
+        unsafe {
+            use hal::Device;
+            self.device.raw().destroy_bind_group_layout(&mut self.raw);
         }
     }
 }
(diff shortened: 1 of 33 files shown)


nical:
Cc @cwfitzgerald. it's nice to get rid of the options, although I don't remember whether or not passing the resources by value to the hal destroy functions was part of you grand plan for the resource management stuff.

Wumpf:
The thing that concerns me here is that this is semantically less safe/unsound from HAL's perspective, as it means that a resource may be destroyed twice. (no doubt that removing myriads of `unwrap` is a _good thing_™️)  
(that said, also opting to follow @cwfitzgerald judgement here since he has the best overview on HAL)

udoprog:
@Wumpf Is `wgpu-hal` intended to be used publicly?

nical:
> @Wumpf Is `wgpu-hal` intended to be used publicly?

It doesn't have many users today, but there's a plan to use it directly in Firefox at some point for CSS rendering.

udoprog:
All right. There's so much unsafe and undocumented assumptions that *I'm not sure* how much passing ownership moves the needle in terms of correct use. Whether if it's worth the extra codegen* and complexity in `wgpu-core` and other higher level abstractions using it I'll leave to you to decide.

> *: This change saves me about 20kb of executable when release building wgpu-examples in release mode.

I'll also note that we could still get rid of the unwraps without changing hal's `destroy_*` methods by marking fields as `ManuallyDrop` and using the `unsafe` [`ManuallyDrop::take`](https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html#method.take).

nical:
Let's see what Connor says. I'm personally quite happy with the prospect of getting rid of unwraps and the code bloat. This code has had a lot of changes recently and there's more to come so I just want to make sure it doesn't conflict with the plan.

cwfitzgerald:
I'm going to say I don't think I want this. This doesn't conflict with any grand-plan stuff, but I'm extremely hesitant to introduce non-linear types on the hal level as we've gotten _burned_ by issues with accidental double/early frees and we're still struggling with them.

I just don't think the potential serious security issues is worth the removal of an option.

We could get rid of some of the boilerplate by introducing an "assumed resident option" which impls deref and such and panics if it ever isn't present. As it stands, quite a few assumptions are correct by construction:
- The device must be alive when drop is called, otherwise you could not call it.
- The resource must not have been previously dropped, otherwise it would be None.
- It's impossible to drop something in a weird place, as you require an Arc, and you need to `take` to get it out.

udoprog:
All right

Before closing, there's two points I want to cover: wgpu-hal and safety with linear types

#### wgpu-hal

This is the part I agree with. Correct use of wgpu-hal becomes harder.

I wouldn't classify it as a blocking concern because:
* It's already super duper unsafe it is to use directly.
* Users of wgpu-hal probably also want to organize their code with destructors since that is idiomatic Rust, so focusing on this use case is probably better.

That being said, if this is a blocking concern for you I totally get it!

#### safety with linear types

I do understand your concern about use of linear types when you're dealing with data in code.

But, this PR only touches the *internals* of a handful of types where we should be able to assume that unless there is a bug somewhere these are only destructed once. So the surface area of where we need to be careful is limited to where we implement destructors.

We're not doing something like this, which would be rather dangerous:

```rust
let mut value = Resource;

destroy(&mut value);

// oh no, someone might use value again!
```

Guarding types internally against double destruction isn't something we should generally be worried about. If we actually were, most of wgpu would already be broken.

cwfitzgerald:
> It's already super duper unsafe it is to use directly.

To an extent, yes, I don't really want to make it harder. I understand that there wouldn't be any practical differences, it just feels like moving in the wrong direction for not a ton of gain. If the gain was bigger I would consider it.

udoprog:
> If the gain was bigger I would consider it.

@cwfitzgerald I just want to ask for clarification; do you mean conceptual gains or would your opinion shift if more unwraps were removed and further binary size reduction? This was a preliminary change to see if there was appetite for this kind of change, I guesstimated that around 200 unwraps in total could be removed.

Issue #5080, opened 2024-01-17
AnyDevice and AnySurface possibly makes incorrect layout assumptions about trait objects


udoprog:
Reference: https://github.com/rust-lang/reference/issues/1448

**Description**

[This](https://github.com/gfx-rs/wgpu/blob/b8f27c7284b95809e26069a9ace0282cc5904006/wgpu-core/src/any_surface.rs#L64C51-L64C61) and [this](https://github.com/gfx-rs/wgpu/blob/b8f27c7284b95809e26069a9ace0282cc5904006/wgpu-core/src/device/any_device.rs#L44) cast assumes that the dst cast extracts the (thin) data pointer.

```
*pointer*
*vtable*
```

Which while it should mostly match up in practice, is to my knowledge not currently guaranteed by Rust. A different compiler option or future compilers could legally opt to re-order or change the semantics of this behavior.

[I have a branch](https://github.com/gfx-rs/wgpu/compare/trunk...udoprog:wgpu:type-erased?expand=1) which re-implements these using internal vtables, while also removing HalSurface. I'll be happy to clean it up into a PR once this has been triaged.

udoprog:
Note that [ptr-meta](https://rust-lang.github.io/rfcs/2580-ptr-meta.html) is an unstable feature which provides APIs for decomposing wide pointers, but it is not yet stable nor can I find the proposed casting logic having been accepted into a stable compiler.

PR #5081, opened 2024-01-17
Remove HalSurface and fix layout assumptions in AnySurface and AnyDevice

udoprog:
**Connections**
#5080

**Description**
This avoids figuring out if #5080 is an actual problem by introducing an internal vtable for the types in question instead of using a trait object. This also affords us to stored the `backend` statically in the vtable, which hopefully makes figuring out which backend an `AnySurface` belongs to much nicer.

I've also reworked `AnySurface` to avoid a double-indirection, since AFAICT it only really needs to hold onto an `A::Surface`. So it can be directly boxed.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/any_surface.rs
+++ b/wgpu-core/src/any_surface.rs
@@ -1,92 +1,93 @@
 use wgt::Backend;
 
-/// The `AnySurface` type: a `Arc` of a `HalSurface<A>` for any backend `A`.
+/// The `AnySurface` type: a `Arc` of a `A::Surface` for any backend `A`.
 use crate::hal_api::HalApi;
-use crate::instance::HalSurface;
 
-use std::any::Any;
 use std::fmt;
-use std::sync::Arc;
+use std::mem::ManuallyDrop;
+use std::ptr::NonNull;
 
-/// A `Arc` of a `HalSurface<A>`, for any backend `A`.
+struct AnySurfaceVtable {
+    // We oppurtunistically store the backend here, since we now it will be used
+    // with backend selection and it can be stored in static memory.
+    backend: Backend,
+    // Drop glue which knows how to drop the stored data.
+    drop: unsafe fn(*mut ()),
+}
+
+/// An `A::Surface`, for any backend `A`.
 ///
-/// Any `AnySurface` is just like an `Arc<HalSurface<A>>`, except that the
-/// `A` type parameter is erased. To access the `Surface`, you must
-/// downcast to a particular backend with the \[`downcast_ref`\] or
-/// \[`take`\] methods.
-pub struct AnySurface(Arc<dyn Any + 'static>);
+/// Any `AnySurface` is just like an `A::Surface`, except that the `A` type
+/// parameter is erased. To access the `Surface`, you must downcast to a
+/// particular backend with the \[`downcast_ref`\] or \[`take`\] methods.
+pub struct AnySurface {
+    data: NonNull<()>,
+    vtable: &'static AnySurfaceVtable,
+}
 
 impl AnySurface {
-    /// Return an `AnySurface` that holds an owning `Arc` to `HalSurface`.
-    pub fn new<A: HalApi>(surface: HalSurface<A>) -> AnySurface {
-        AnySurface(Arc::new(surface))
+    /// Construct an `AnySurface` that owns an `A::Surface`.
+    pub fn new<A: HalApi>(surface: A::Surface) -> AnySurface {
+        unsafe fn drop_glue<A: HalApi>(ptr: *mut ()) {
+            unsafe {
+                _ = Box::from_raw(ptr.cast::<A::Surface>());
+            }
+        }
+
+        let data = NonNull::from(Box::leak(Box::new(surface)));
+
+        AnySurface {
+            data: data.cast(),
+            vtable: &AnySurfaceVtable {
+                backend: A::VARIANT,
+                drop: drop_glue::<A>,
+            },
+        }
     }
 
+    /// Get the backend this surface was created through.
     pub fn backend(&self) -> Backend {
-        #[cfg(vulkan)]
-        if self.downcast_ref::<hal::api::Vulkan>().is_some() {
-            return Backend::Vulkan;
-        }
-        #[cfg(metal)]
-        if self.downcast_ref::<hal::api::Metal>().is_some() {
-            return Backend::Metal;
-        }
-        #[cfg(dx12)]
-        if self.downcast_ref::<hal::api::Dx12>().is_some() {
-            return Backend::Dx12;
-        }
-        #[cfg(gles)]
-        if self.downcast_ref::<hal::api::Gles>().is_some() {
-            return Backend::Gl;
-        }
-        Backend::Empty
+        self.vtable.backend
     }
 
-    /// If `self` is an `Arc<HalSurface<A>>`, return a reference to the
-    /// HalSurface.
-    pub fn downcast_ref<A: HalApi>(&self) -> Option<&HalSurface<A>> {
-        self.0.downcast_ref::<HalSurface<A>>()
+    /// If `self` refers to an `A::Surface`, returns a reference to it.
+    pub fn downcast_ref<A: HalApi>(&self) -> Option<&A::Surface> {
+        if A::VARIANT != self.vtable.backend {
+            return None;
+        }
+
+        // SAFETY: We just checked the instance above implicitly by the backend
+        // that it was statically constructed through.
+        Some(unsafe { &*self.data.as_ptr().cast::<A::Surface>() })
     }
 
-    /// If `self` is an `Arc<HalSurface<A>>`, returns that.
-    pub fn take<A: HalApi>(self) -> Option<Arc<HalSurface<A>>> {
-        // `Arc::downcast` returns `Arc<T>`, but requires that `T` be `Sync` and
-        // `Send`, and this is not the case for `HalSurface` in wasm builds.
-        //
-        // But as far as I can see, `Arc::downcast` has no particular reason to
-        // require that `T` be `Sync` and `Send`; the steps used here are sound.
-        if (self.0).is::<HalSurface<A>>() {
-            // Turn the `Arc`, which is a pointer to an `ArcInner` struct, into
-            // a pointer to the `ArcInner`'s `data` field. Carry along the
-            // vtable from the original `Arc`.
-            let raw_erased: *const (dyn Any + 'static) = Arc::into_raw(self.0);
-            // Remove the vtable, and supply the concrete type of the `data`.
-            let raw_typed: *const HalSurface<A> = raw_erased.cast::<HalSurface<A>>();
-            // Convert the pointer to the `data` field back into a pointer to
-            // the `ArcInner`, and restore reference-counting behavior.
-            let arc_typed: Arc<HalSurface<A>> = unsafe {
-                // Safety:
-                // - We checked that the `dyn Any` was indeed a `HalSurface<A>` above.
-                // - We're calling `Arc::from_raw` on the same pointer returned
-                //   by `Arc::into_raw`, except that we stripped off the vtable
-                //   pointer.
-                // - The pointer must still be live, because we've borrowed `self`,
-                //   which holds another reference to it.
-                // - The format of a `ArcInner<dyn Any>` must be the same as
-                //   that of an `ArcInner<HalSurface<A>>`, or else `AnyHalSurface::new`
-                //   wouldn't be possible.
-                Arc::from_raw(raw_typed)
-            };
-            Some(arc_typed)
-        } else {
-            None
+    /// If `self` is an `Arc<A::Surface>`, returns that.
+    pub fn take<A: HalApi>(self) -> Option<A::Surface> {
+        if A::VARIANT != self.vtable.backend {
+            return None;
         }
+
+        // Disable drop glue, since we're returning the owned surface. The
+        // caller will be responsible for dropping it.
+        let this = ManuallyDrop::new(self);
+
+        // SAFETY: We just checked the instance above implicitly by the backend
+        // that it was statically constructed through.
+        Some(unsafe { *Box::from_raw(this.data.as_ptr().cast::<A::Surface>()) })
+    }
+}
+
+impl Drop for AnySurface {
+    fn drop(&mut self) {
+        unsafe { (self.vtable.drop)(self.data.as_ptr()) }
     }
 }
 
 impl fmt::Debug for AnySurface {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str("AnySurface")
+        f.debug_tuple("AnySurface")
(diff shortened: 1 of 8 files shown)


ErichDonGubler (commented):
I've done a slow pass over this, and the broad strokes LGTM. It's always good to avoid "We upgraded `rustc`, and things broke, and it's our fault." 😅 I want to do another review pass focused on the fine-grained logic and types here, since my experience with Rust FFI is that there's often subtle bugs lurking in even the best human's code that aren't obvious. Will hopefully report back in the next day or so.

wgpu-core/src/any_surface.rs
```diff
@@ -1,92 +1,93 @@
...
-                //   which holds another reference to it.
-                // - The format of a `ArcInner<dyn Any>` must be the same as
-                //   that of an `ArcInner<HalSurface<A>>`, or else `AnyHalSurface::new`
-                //   wouldn't be possible.
-                Arc::from_raw(raw_typed)
-            };
-            Some(arc_typed)
-        } else {
-            None
+    /// If `self` is an `Arc<A::Surface>`, returns that.
```
ErichDonGubler, line 64: issue: We're not talking about an `Arc<…>` at all, tho! 🤔

suggestion:

```suggestion
    /// If `self` is an `A::Surface`, returns that.
```

wgpu-core/src/any_surface.rs
```diff
@@ -1,92 +1,93 @@
...
-/// Any `AnySurface` is just like an `Arc<HalSurface<A>>`, except that the
-/// `A` type parameter is erased. To access the `Surface`, you must
-/// downcast to a particular backend with the \[`downcast_ref`\] or
-/// \[`take`\] methods.
-pub struct AnySurface(Arc<dyn Any + 'static>);
+/// Any `AnySurface` is just like an `A::Surface`, except that the `A` type
+/// parameter is erased. To access the `Surface`, you must downcast to a
+/// particular backend with the \[`downcast_ref`\] or \[`take`\] methods.
+pub struct AnySurface {
+    data: NonNull<()>,
```
ErichDonGubler, line 24: suggestion(blocking): It's not _incorrect_, per se, to use `NonNull` here, but it makes me nervous to see in an FFI layer. Two other (seemingly inert, in this case?) benefits-that-can-be-hazards start to come into consideration when we use it with FFI:

1. `NonNull` can be a footgun WRT subtyping, particularly lifetimes. `NonNull`, as its docs state, assume that the pointee is covariant with its `T`. This shouldn't change anything notable when used with the the concrete `()`, but this might cause some cognitive dissonance WRT intent (like it did for me).
2. `()` is a ZST, and _definitely_ not what's behind this pointer. Normally, `std::ffi::c_void` or equivalent is used for this, which has a different representation (and is actually FFI-friendly). ZSTs can change pointer offset operations, which, again, we're not doing here, but it makes me nervous to see in FFI code. Best to avoid problems with assumptions the compiler can make about ZSTs, in general, IMO.

I think it'd be better to have `*mut c_void` here (and to change `AnySurfaceVtable::drop` to `unsafe fn(*mut c_void)`, OFC), to avoid both of these hazards. Yes, `*mut c_void` isn't as "nice" (read: ergonomic, strongly typed), but it's a lot easier to find issues with a typical mis-use of `*mut _` than accidentally going astray with the guarantees we opt into with `NonNull<()>`.

wgpu-core/src/any_surface.rs
```diff
@@ -1,92 +1,93 @@
...
+            return None;
         }
+
+        // Disable drop glue, since we're returning the owned surface. The
+        // caller will be responsible for dropping it.
+        let this = ManuallyDrop::new(self);
+
+        // SAFETY: We just checked the instance above implicitly by the backend
+        // that it was statically constructed through.
+        Some(unsafe { *Box::from_raw(this.data.as_ptr().cast::<A::Surface>()) })
```
ErichDonGubler, line 76: suggestion(blocking): I don't see a good reason to use `ManuallyDrop::new(self)` over destructuring `self`. It's significantly more common to do the latter, and involves less API surface. Therefore, I believe it's easier to understand immediately—I daresay it wouldn't even need a comment:

```suggestion
        // Disable drop glue, since we're returning the owned surface. The
        // caller will be responsible for dropping it.
        let Self { data, .. } = self;

        // SAFETY: We just checked the instance above implicitly by the backend
        // that it was statically constructed through.
        Some(unsafe { *Box::from_raw(data.as_ptr().cast::<A::Surface>()) })
```

wgpu-core/src/any_surface.rs
ErichDonGubler: This feedback also applies to the new usage of `NonNull<()>` in `any_device.rs`.

wgpu-core/src/device/any_device.rs
```diff
@@ -1,72 +1,98 @@
...
+        // invalidate this object.
+        //
+        // SAFETY: We just checked the instance above implicitly by the backend
+        // that it was statically constructed through.
+        let this =
+            ManuallyDrop::new(unsafe { Arc::from_raw(self.data.as_ptr().cast::<Device<A>>()) });
+
+        // Cloning it increases the reference count, and we return a new arc
+        // instance.
+        Some((*this).clone())
```
ErichDonGubler, line 83: nitpick: I think it would be clearer to use [`Arc::increment_strong_count`](https://doc.rust-lang.org/stable/std/sync/struct.Arc.html#method.increment_strong_count), despite technically being yet another `unsafe` call. Ignoring my feedback about changing `NonNull<()>` to be a `*mut c_void`, this would look like:

```suggestion        let ptr = self.data.as_ptr().cast::<Device<A>>();
        // Increment the strong count on this pointer so we can give a "cloned" `Arc` back.
        //
        // SAFETY:
        //
        // * We know that `ptr` points to a `Device<A>` because of the above check.
        // * We also know the strong reference count should be greater than 0, because, in this
        //   API, we've only allowed either (a) references to the pointee, or (b) `Arc`s with their
        //   own corresponding strong reference count to decrement on `Drop`.
        unsafe { Arc::increment_strong_count(ptr) };
        // SAFETY: We already assume we have a `Device<A>` above, so this isn't adding unsoundness.
        Some(unsafe { Arc::from_raw(ptr) })

```

wgpu-core/src/any_surface.rs
```diff
@@ -1,92 +1,93 @@
...
 
-use std::any::Any;
 use std::fmt;
-use std::sync::Arc;
+use std::mem::ManuallyDrop;
+use std::ptr::NonNull;
 
-/// A `Arc` of a `HalSurface<A>`, for any backend `A`.
+struct AnySurfaceVtable {
+    // We oppurtunistically store the backend here, since we now it will be used
```
ErichDonGubler, line 11: typo:

```suggestion
    // We opportunistically store the backend here, since we know it will be used
```

wgpu-core/src/device/any_device.rs
```diff
@@ -1,72 +1,98 @@
...
 
 impl AnyDevice {
     /// Return an `AnyDevice` that holds an owning `Arc` pointer to `device`.
     pub fn new<A: HalApi>(device: Arc<Device<A>>) -> AnyDevice {
-        AnyDevice(device)
+        unsafe fn drop_glue<A: HalApi>(ptr: *mut ()) {
+            // Drop the arc this instance is holding.
+            unsafe {
+                _ = Arc::from_raw(ptr.cast::<A::Surface>());
+            }
```
ErichDonGubler, line 38: nitpick: I think it would be clearer to use [`Arc::decrement_strong_count`](https://doc.rust-lang.org/stable/std/sync/struct.Arc.html#method.decrement_strong_count) here:

```suggestion
            unsafe {
                Arc::decrement_strong_count(ptr.cast::<A::Surface>());
            }
```

wgpu-core/src/device/any_device.rs
```diff
@@ -1,72 +1,98 @@
...
+        }
+
+        // SAFETY: The pointer returned by Arc::into_raw is gauranteed to be
+        // non-null.
```
ErichDonGubler, line 42: nitpick:

```suggestion
        // SAFETY: The pointer returned by `Arc::into_raw` is guaranteed to be non-null.
```

wgpu-core/src/device/any_device.rs
```diff
@@ -1,72 +1,98 @@
...
 use crate::hal_api::HalApi;
 
-use std::any::Any;
 use std::fmt;
+use std::mem::ManuallyDrop;
+use std::ptr::NonNull;
 use std::sync::Arc;
 
+struct AnyDeviceVtable {
+    // We oppurtunistically store the backend here, since we now it will be used
```
ErichDonGubler, line 13: typo:

```suggestion
    // We opportunistically store the backend here, since we know it will be used
```

wgpu-core/src/device/any_device.rs
```diff
@@ -1,72 +1,98 @@
...
 ///
-/// Any `AnyDevice` is just like an `Arc<Device<A>>`, except that the
-/// `A` type parameter is erased. To access the `Device`, you must
-/// downcast to a particular backend with the \[`downcast_ref`\] or
-/// \[`downcast_clone`\] methods.
-pub struct AnyDevice(Arc<dyn Any + 'static>);
+/// Any `AnyDevice` is just like an `Arc<Device<A>>`, except that the `A` type
+/// parameter is erased. To access the `Device`, you must downcast to a
+/// particular backend with the \[`downcast_ref`\] or \[`downcast_clone`\]
+/// methods.
```
ErichDonGubler, line 25: nitpick(non-blocking): It's nice to link to referenced methods:

```suggestion
/// particular backend with the [`Self::downcast_ref`] or
/// [`Self::downcast_clone`] methods.
```

ErichDonGubler: thought: Side note, your editor seems to soft-break lines to be more narrow than my `neovim` configuration does. 🤔 I wonder why?

ErichDonGubler (changes requested):
praise: This is a nice simplification and fix. Thanks! 🙂

Overall, LGTM minus suggestions and nits! I'm marking as `Request changes` so that the issues I've found get handled, but I'm confident in the overall thrust of this PR. On that note, I'm going to prompt for feedback from @Wumpf and/or @cwfitzgerald, since while I've done quite a bit of FFI/pointer-slinging in Rust before, I'm not as familiar with native platform domain/API for surfaces and devices.


---

<sub>I used [Conventional Comments](https://conventionalcomments.org/) in this review! I hope they help with clarity and tone. 🙂</sub>

wgpu-core/src/device/any_device.rs
udoprog: We probably just have different widths for soft breaking.

wgpu-core/src/any_surface.rs
udoprog: Unfortunately this doesn't prevent the destructor from being called, the `data` field is copied out. Note that the drop implementation we want to prevent from running is `self`, not `data`.

[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d87d79f3bd852e80c7db13e27d2d87a7).

wgpu-core/src/any_surface.rs
udoprog: > NonNull can be a footgun WRT subtyping

Covariance should only matter for a concrete `T` and particularly when it [leaks through as a public type parameter](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=98e32deeb95e9389b37ce442205d7d22). The only property we're using here is niche filling (cannot inhabit zero bits). But I don't mind changing it either way though.

> () is a ZST, and definitely not what's behind this pointer

Every raw thin pointer (but not unsized pointers) has the same layout and representation both from a Rust and an FFI perspective. The type parameter should only matter when it's being dereferenced, for which we are careful to cast it to the correct underlying type.

Regarding FFI, this is to my understanding currently just a private implementation detail. To make the whole structure FFI safe to the extent that we'd allow callers to access implementation details of it (e.g. assume that the first field is a pointer), we'd have to mark it `repr(C)`. If we do so, NonNull has been fine to use in FFI since 1.29 (when it received `#[repr(transparent)]`). Today it's guaranteed from a layout perspective to be exactly the same as `*mut T`.

I'd like it if you elaborated on how this is currently used in FFI, and how you want to use it.

udoprog:
Thanks for taking a look @ErichDonGubler!

I'll rebase the change once it has been looked over by all parties who might find it objectionable and the outstanding questions have been resolved.

wgpu-core/src/any_surface.rs
ErichDonGubler: Oh, derp, you're totally right, and I forgot to consider that interaction with `Copy`. 😅 I'll `Resolve` this.

Non-blocking thoughts: I would suggest `std::mem::forget` after the _fully_ destructuring, but I don't consider that _significantly_ better than `ManuallyDrop` in this particular case. I would consider it for anything more complicated, since destructuring can help a programmer remember to handle other fields or otherwise have a signal that other fields might need to be handled.

wgpu-core/src/any_surface.rs
ErichDonGubler: To make my position clear: I don't believe the properties of `NonNull` and `()` that I've brought up are actually problematic in this concrete usage. I'm mostly concerned with avoiding habits that can be problematic in other situations where APIs are being built on top of pointer-casting code. I doubt this code will change to ever expose `data`'s type parameter, so I don't see a latent long-term problem, either.

To me, what I bring up is a question of taste. I feel this conversation has been given due consideration from you as the PR author, so I don't consider the points I've raised to be blocking now. I'll invite @cwfitzgerald and @jimblandy to provide their thoughts, if any, but I'm not invested in a specific outcome.

RE: FFI usage: I think I used the term "FFI" more loosely than I should have, sorry about that. Concerns when authoring FFI overlap with my thinking here, to be sure. However, FFI is only tangentially related to this conversation because we use FFI to interface with the platforms that back surfaces.

The only usage of this specific API that I'm aware of is to consolidate a family of types into one (`AnySurface`) for the sake of an easier-to-write `wgpu_core::instance::Surface` API. In fact, I'm not even sure that exposing this as `pub` to consumers of `wgpu-core` is intended, since I don't see a way for `AnySurface`s to be used in our API at all (CC @cwfitzgerald, @jimblandy; is this right?).

ErichDonGubler (approved):
LGTM now! Planning on waiting to merge this for a day or so, to give @jimblandy or @cwfitzgerald a chance to look over the pings I've given them specifically. I don't see a strong impetus to wait longer than that, though.

udoprog:
All right, I've rebased the change to try and get CI green so you're unblocked once you make a decision.

I've also [included this minor nit](https://github.com/gfx-rs/wgpu/pull/5081/commits/6d3aa52a6758968cfae8ae378f34326ffb009ec0), because it doesn't make much sense to format one as a tuple and omit useful information in the other.

PR #5089, opened 2024-01-18
[msl-out] prevent infinite loops from being optimized out by the metal compiler

### LINKED ISSUE #4972: Metal Shading Language compiler treats infinite loops as UB
The Metal Shading Language compiler sometimes determines that a loop diverges. Since divergence is undefined behavior in C++, the compiler will assume the loop must be unreachable, which can lead it to make other conclusions that could result in erased bounds checks.

The workaround is to place a magic incantation in loops to cause the compiler to believe that every loop has a side effect.

Here's Dawn's solution: https://dawn-review.googlesource.com/c/dawn/+/166860 

A better explanation for it: https://dawn-review.googlesource.com/c/dawn/+/167763/3/src/tint/lang/msl/writer/ast_printer/ast_printer.cc 

```
    // This statement prevents the MSL compiler from erasing a loop during
    // optimizations.  In the AIR dialiect of LLVM IR, WGSL loops should compile
    // to a loop that contains an 'asm' call with a 'sideeffect' annotation.
    //
    // For example, compile a WGSL file with a trivial while(1) loop to 'a.metal',
    // then compile that to AIR (LLVM IR dialect):
    //
    //    xcrun metal a.metal -S -o -
    //
    // The loop in the AIR should look something like this:
    //
    //    1: ...
    //      br label %2
    //
    //    2:                                      ; preds = %1, %2
    //      tail call void asm sideeffect "", ""() #1, !srcloc !27
    //      br label %2, !llvm.loop !28
    //
    // It is important that the 'sideeffect' annotation exist. That tells the
    // optimizer that the instruction has side effects invisible to the
    // optimizer, and therefore the loop should not be eliminated.
    Line() << R"(__asm__("");)";
```

teoxoy:
Fixes https://github.com/gfx-rs/wgpu/issues/4972.

### DIFF
--- a/naga/src/back/msl/writer.rs
+++ b/naga/src/back/msl/writer.rs
@@ -2768,7 +2768,9 @@ impl<W: Write> Writer<W> {
                     if !continuing.is_empty() || break_if.is_some() {
                         let gate_name = self.namer.call("loop_init");
                         writeln!(self.out, "{level}bool {gate_name} = true;")?;
-                        writeln!(self.out, "{level}while(true) {{")?;
+                        // The empty inline assembly "call" will prevent the metal compiler
+                        // from optimizing out these potentially-infinte loops.
+                        writeln!(self.out, "{level}while(true) {{__asm__(\"\");")?;
                         let lif = level.next();
                         let lcontinuing = lif.next();
                         writeln!(self.out, "{lif}if (!{gate_name}) {{")?;
@@ -2783,7 +2785,7 @@ impl<W: Write> Writer<W> {
                         writeln!(self.out, "{lif}}}")?;
                         writeln!(self.out, "{lif}{gate_name} = false;")?;
                     } else {
-                        writeln!(self.out, "{level}while(true) {{")?;
+                        writeln!(self.out, "{level}while(true) {{__asm__(\"\");")?;
                     }
                     self.put_block(level.next(), body, context)?;
                     writeln!(self.out, "{level}}}")?;
--- a/naga/tests/out/msl/boids.msl
+++ b/naga/tests/out/msl/boids.msl
@@ -56,7 +56,7 @@ kernel void main_(
     metal::float2 _e14 = particlesSrc.particles[index].vel;
     vVel = _e14;
     bool loop_init = true;
-    while(true) {
+    while(true) {__asm__("");
         if (!loop_init) {
             uint _e91 = i;
             i = _e91 + 1u;
--- a/naga/tests/out/msl/break-if.msl
+++ b/naga/tests/out/msl/break-if.msl
@@ -8,7 +8,7 @@ using metal::uint;
 void breakIfEmpty(
 ) {
     bool loop_init = true;
-    while(true) {
+    while(true) {__asm__("");
         if (!loop_init) {
             if (true) {
                 break;
@@ -25,7 +25,7 @@ void breakIfEmptyBody(
     bool b = {};
     bool c = {};
     bool loop_init_1 = true;
-    while(true) {
+    while(true) {__asm__("");
         if (!loop_init_1) {
             b = a;
             bool _e2 = b;
@@ -46,7 +46,7 @@ void breakIf(
     bool d = {};
     bool e = {};
     bool loop_init_2 = true;
-    while(true) {
+    while(true) {__asm__("");
         if (!loop_init_2) {
             bool _e5 = e;
             if (a_1 == e) {
@@ -65,7 +65,7 @@ void breakIfSeparateVariable(
 ) {
     uint counter = 0u;
     bool loop_init_3 = true;
-    while(true) {
+    while(true) {__asm__("");
         if (!loop_init_3) {
             uint _e5 = counter;
             if (counter == 5u) {
--- a/naga/tests/out/msl/collatz.msl
+++ b/naga/tests/out/msl/collatz.msl
@@ -19,7 +19,7 @@ uint collatz_iterations(
     uint n = {};
     uint i = 0u;
     n = n_base;
-    while(true) {
+    while(true) {__asm__("");
         uint _e4 = n;
         if (_e4 > 1u) {
         } else {
--- a/naga/tests/out/msl/control-flow.msl
+++ b/naga/tests/out/msl/control-flow.msl
@@ -31,7 +31,7 @@ void switch_case_break(
 void loop_switch_continue(
     int x
 ) {
-    while(true) {
+    while(true) {__asm__("");
         switch(x) {
             case 1: {
                 continue;
--- a/naga/tests/out/msl/do-while.msl
+++ b/naga/tests/out/msl/do-while.msl
@@ -9,7 +9,7 @@ void fb1_(
     thread bool& cond
 ) {
     bool loop_init = true;
-    while(true) {
+    while(true) {__asm__("");
         if (!loop_init) {
             bool _e1 = cond;
             if (!(cond)) {
--- a/naga/tests/out/msl/ray-query.msl
+++ b/naga/tests/out/msl/ray-query.msl
@@ -63,7 +63,7 @@ kernel void main_(
     rq.intersector.force_opacity((_e12.flags & 1) != 0 ? metal::raytracing::forced_opacity::opaque : (_e12.flags & 2) != 0 ? metal::raytracing::forced_opacity::non_opaque : metal::raytracing::forced_opacity::none);
     rq.intersector.accept_any_intersection((_e12.flags & 4) != 0);
     rq.intersection = rq.intersector.intersect(metal::raytracing::ray(_e12.origin, _e12.dir, _e12.tmin, _e12.tmax), acc_struct, _e12.cull_mask);    rq.ready = true;
-    while(true) {
+    while(true) {__asm__("");
         bool _e13 = rq.ready;
         rq.ready = false;
         if (_e13) {
--- a/naga/tests/out/msl/shadow.msl
+++ b/naga/tests/out/msl/shadow.msl
@@ -101,7 +101,7 @@ fragment fs_mainOutput fs_main(
     uint i = 0u;
     metal::float3 normal_1 = metal::normalize(in.world_normal);
     bool loop_init = true;
-    while(true) {
+    while(true) {__asm__("");
         if (!loop_init) {
             uint _e40 = i;
             i = _e40 + 1u;
@@ -151,7 +151,7 @@ fragment fs_main_without_storageOutput fs_main_without_storage(
     uint i_1 = 0u;
     metal::float3 normal_2 = metal::normalize(in_1.world_normal);
     bool loop_init_1 = true;
-    while(true) {
+    while(true) {__asm__("");
         if (!loop_init_1) {
             uint _e40 = i_1;
             i_1 = _e40 + 1u;


teoxoy:
Looks like we are now getting the AIR we expect.

Diff of AIR generated from one of our snapshot files:

```diff
diff --git a/t.air b/t.air
index 76ef40ad7..a4c248870 100644
--- a/t.air
+++ b/t.air
@@ -1,5 +1,5 @@
-; ModuleID = 'C:\local\Temp\7ce36e6a-2906-4e76-b08d-935c32650cf8.metal'
-source_filename = "C:\5Clocal\5CTemp\5C7ce36e6a-2906-4e76-b08d-935c32650cf8.metal"
+; ModuleID = 'C:\local\Temp\20f1b85f-a3b9-4b1e-9061-0e9e90934cb9.metal'
+source_filename = "C:\5Clocal\5CTemp\5C20f1b85f-a3b9-4b1e-9061-0e9e90934cb9.metal"
 target datalayout = "e-p:64:64:64-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-f32:32:32-f64:64:64-v16:16:16-v24:32:32-v32:32:32-v48:64:64-v64:64:64-v96:128:128-v128:128:128-v192:256:256-v256:256:256-v512:512:512-v1024:1024:1024-n8:16:32"
 target triple = "air64-apple-macosx13.0.0"
 
@@ -8,7 +8,7 @@ target triple = "air64-apple-macosx13.0.0"
 %struct.Particle = type { <2 x float>, <2 x float> }
 %struct._mslBufferSizes = type { i32, i32 }
 
-; Function Attrs: nounwind
+; Function Attrs: convergent nounwind
 define void @main_(<3 x i32>, %struct.SimParams addrspace(2)* noalias nocapture readonly dereferenceable(28), %struct.Particles addrspace(1)* noalias nocapture readonly dereferenceable(16), %struct.Particles addrspace(1)* noalias nocapture dereferenceable(16), %struct._mslBufferSizes addrspace(2)* noalias nocapture readnone dereferenceable(8)) local_unnamed_addr #0 {
   %6 = extractelement <3 x i32> %0, i64 0
   %7 = icmp ugt i32 %6, 1499
@@ -36,6 +36,7 @@ define void @main_(<3 x i32>, %struct.SimParams addrspace(2)* noalias nocapture
   %25 = phi <2 x float> [ zeroinitializer, %8 ], [ %59, %55 ]
   %26 = phi <2 x float> [ zeroinitializer, %8 ], [ %60, %55 ]
   %27 = phi <2 x float> [ zeroinitializer, %8 ], [ %61, %55 ]
+  tail call void asm sideeffect "", ""() #2, !srcloc !29
   %28 = add i32 %22, %21
   %29 = icmp ugt i32 %28, 1499
   br i1 %29, label %55, label %30
@@ -105,11 +106,11 @@ define void @main_(<3 x i32>, %struct.SimParams addrspace(2)* noalias nocapture
 ; <label>:78:                                     ; preds = %73, %70
   %79 = phi <2 x float> [ %77, %73 ], [ %60, %70 ]
   %80 = getelementptr inbounds %struct.SimParams, %struct.SimParams addrspace(2)* %1, i64 0, i32 4
-  %81 = load float, float addrspace(2)* %80, align 4, !tbaa !29
+  %81 = load float, float addrspace(2)* %80, align 4, !tbaa !30
   %82 = getelementptr inbounds %struct.SimParams, %struct.SimParams addrspace(2)* %1, i64 0, i32 5
-  %83 = load float, float addrspace(2)* %82, align 4, !tbaa !32
+  %83 = load float, float addrspace(2)* %82, align 4, !tbaa !33
   %84 = getelementptr inbounds %struct.SimParams, %struct.SimParams addrspace(2)* %1, i64 0, i32 6
-  %85 = load float, float addrspace(2)* %84, align 4, !tbaa !33
+  %85 = load float, float addrspace(2)* %84, align 4, !tbaa !34
   %86 = insertelement <2 x float> undef, float %81, i32 0
   %87 = shufflevector <2 x float> %86, <2 x float> undef, <2 x i32> zeroinitializer
   %88 = fmul fast <2 x float> %87, %71
@@ -133,7 +134,7 @@ define void @main_(<3 x i32>, %struct.SimParams addrspace(2)* noalias nocapture
   %106 = shufflevector <2 x float> %105, <2 x float> undef, <2 x i32> zeroinitializer
   %107 = fmul fast <2 x float> %102, %106
   %108 = getelementptr inbounds %struct.SimParams, %struct.SimParams addrspace(2)* %1, i64 0, i32 0
-  %109 = load float, float addrspace(2)* %108, align 4, !tbaa !34
+  %109 = load float, float addrspace(2)* %108, align 4, !tbaa !35
   %110 = insertelement <2 x float> undef, float %109, i32 0
   %111 = shufflevector <2 x float> %110, <2 x float> undef, <2 x i32> zeroinitializer
   %112 = fmul fast <2 x float> %107, %111
@@ -179,8 +180,9 @@ declare float @air.fast_rsqrt.f32(float) local_unnamed_addr #1
 ; Function Attrs: nounwind readnone
 declare float @air.fast_clamp.f32(float, float, float) local_unnamed_addr #1
 
-attributes #0 = { nounwind "correctly-rounded-divide-sqrt-fp-math"="false" "disable-tail-calls"="false" "less-precise-fpmad"="false" "no-frame-pointer-elim"="true" "no-frame-pointer-elim-non-leaf" "no-infs-fp-math"="true" "no-jump-tables"="false" "no-nans-fp-math"="true" "no-signed-zeros-fp-math"="true" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "unsafe-fp-math"="true" "use-soft-float"="false" }
+attributes #0 = { convergent nounwind "correctly-rounded-divide-sqrt-fp-math"="false" "disable-tail-calls"="false" "less-precise-fpmad"="false" "no-frame-pointer-elim"="true" "no-frame-pointer-elim-non-leaf" "no-infs-fp-math"="true" "no-jump-tables"="false" "no-nans-fp-math"="true" "no-signed-zeros-fp-math"="true" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "unsafe-fp-math"="true" "use-soft-float"="false" }
 attributes #1 = { nounwind readnone }
+attributes #2 = { convergent nounwind }
 
 !air.kernel = !{!0}
 !llvm.module.flags = !{!12, !13, !14, !15, !16, !17, !18}
@@ -215,13 +217,14 @@ attributes #1 = { nounwind readnone }
 !22 = !{!"Apple metal version 31001.642 (metalfe-31001.642-windows)"}
 !23 = !{i32 2, i32 5, i32 0}
 !24 = !{!"Metal", i32 3, i32 0, i32 0}
-!25 = !{!"D:\5CDWASFiles\5CSites\5Ctryhlsl\5CTemp\5C7ce36e6a-2906-4e76-b08d-935c32650cf8.metal"}
+!25 = !{!"D:\5CDWASFiles\5CSites\5Ctryhlsl\5CTemp\5C20f1b85f-a3b9-4b1e-9061-0e9e90934cb9.metal"}
 !26 = !{!27, !27, i64 0}
 !27 = !{!"omnipotent char", !28, i64 0}
 !28 = !{!"Simple C++ TBAA"}
-!29 = !{!30, !31, i64 16}
-!30 = !{!"_ZTS9SimParams", !31, i64 0, !31, i64 4, !31, i64 8, !31, i64 12, !31, i64 16, !31, i64 20, !31, i64 24}
-!31 = !{!"float", !27, i64 0}
-!32 = !{!30, !31, i64 20}
-!33 = !{!30, !31, i64 24}
-!34 = !{!30, !31, i64 0}
+!29 = !{i32 1452}
+!30 = !{!31, !32, i64 16}
+!31 = !{!"_ZTS9SimParams", !32, i64 0, !32, i64 4, !32, i64 8, !32, i64 12, !32, i64 16, !32, i64 20, !32, i64 24}
+!32 = !{!"float", !27, i64 0}
+!33 = !{!31, !32, i64 20}
+!34 = !{!31, !32, i64 24}
+!35 = !{!31, !32, i64 0}
```

naga/tests/out/msl/boids.msl
```diff
@@ -56,7 +56,7 @@ kernel void main_(
     metal::float2 _e14 = particlesSrc.particles[index].vel;
     vVel = _e14;
     bool loop_init = true;
-    while(true) {
+    while(true) {__asm__("");
```
ErichDonGubler, line 59: nitpick(style): Do we not want this indented in-line with the rest of the `while` loop's contents?

Won't block either way, but I didn't see anyone call this out yet.

naga/src/back/msl/writer.rs
```diff
@@ -2768,7 +2768,9 @@ impl<W: Write> Writer<W> {
                     if !continuing.is_empty() || break_if.is_some() {
                         let gate_name = self.namer.call("loop_init");
                         writeln!(self.out, "{level}bool {gate_name} = true;")?;
-                        writeln!(self.out, "{level}while(true) {{")?;
+                        // The empty inline assembly "call" will prevent the metal compiler
+                        // from optimizing out these potentially-infinte loops.
+                        writeln!(self.out, "{level}while(true) {{__asm__(\"\");")?;
```
ErichDonGubler, line 2773: idea(non-blocking): I wonder if this might be easier for a Rustacean to understand if this had a helper a la [`std::hint::black_box`](https://doc.rust-lang.org/stable/std/hint/fn.black_box.html)?

ErichDonGubler (approved):

naga/tests/out/msl/boids.msl
teoxoy: At first I thought so too but it seems appropriate that it stands out.

naga/src/back/msl/writer.rs
teoxoy: And the helper would inject `__asm__("");`? That would work but I think that for now it's clearer that only loops use this hack.

jimblandy:
It turns out that this workaround works for some versions of the driver on some platforms, but not others. There is new advice:

https://dawn-review.googlesource.com/c/dawn/+/171440/2/src/tint/lang/msl/writer/ast_printer/ast_printer.cc

When Metal determines that a loop is infinite, it replaces the loop with a marker that essentially means, "This point in the control flow graph is unreachable." The troubles arise because Metal then concludes that everything *following* the loop is thus also unreachable: after all, you'd have to get to the loop first.

Tint's workaround is to declare a volatile `bool`, and then to wrap each loop with an `if` that tests that bool.

They define a preprocessor macro for this:

```c++
#define TINT_ISOLATE_UB(VOLATILE_NAME) \
  volatile bool VOLATILE_NAME = true; \
  if (VOLATILE_NAME)
```

And then they prefix each loop with a test of that macro:

```
  TINT_ISOLATE_UB(tint_volatile_true) for(uint idx = local_invocation_index; (idx < 4u); idx = (idx + 1u)) {
      ...
  }
```

With this done, every loop also has a branch that skips over it, and that branch can't be predicted because it involves a reference to a volatile variable, so the loop no longer dominates the code that follows it.

This workaround could affect performance; the whole point is that the optimizer isn't free to remove these bools, so they will presumably appear in the generated machine code. I'd like to think about this / discuss this a bit more before we adopt this new approach.

teoxoy:
Such un unfortunate situation to be in. I wonder what DXC does with infinite loops since it's also a fork of clang.

jimblandy:
I think we need to go with the volatile bool kludge, unfortunately. inline asm isn't well-supported by Metal drivers.

jimblandy:
Since we can't use this approach, I think we have to abandon this PR.

PR #5091, opened 2024-01-18
Make sure to copy all of the buffers into the resource array for dx12.

### LINKED ISSUE #5088: DX12 crash with wpgu 0.19 when `desired_maximum_frame_latency` < 3
**Description**
The default `desired_maximum_frame_latency` of 2 causes crashes with dx12.

**Repro steps**
Run any of the wgpu examples with dx12. (it might only crash on amd? @cwfitzgerald said on matrix that it worked fine on his nvidia gpu)

Expected: Doesn't crash.
Observed: Crashes immediately. (setting `desired_maximum_frame_latency` to 3 or higher doesn't crash)

**Extra materials**
This line has a few issues https://github.com/gfx-rs/wgpu/blob/adf1e3b6c546098e936f49602c4e39c0e416e1d6/wgpu-hal/src/dx12/mod.rs#L666 (the `+1` seems to be what causes the crash and the `.min(3)` means you can never have more than 3 even if you set `maximum_frame_latency` to eg. 10 because you made a video player that doesn't care about input latency but about smoothness and performance).

`ERROR log: ID3D12CommandQueue::ExecuteCommandLists: A command list, which writes to a swapchain back buffer, may only be executed when that back buffer is the back buffer that will be presented during the next call to Present*. Such a back buffer is also referred to as the "current back buffer". Swap Chain: 0x0000021394DF6D10:'Unnamed Object' - Current Back Buffer Buffer: 0x000002139A18A650:'Unnamed ID3D12Resource Object' - Attempted Write Buffer: 0x000002139A18F180:'Unnamed ID3D12Resource Object' [ STATE_SETTING ERROR #907: EXECUTECOMMANDLISTS_WRONGSWAPCHAINBUFFERREFERENCE]
`

At first I thought it was because we were creating more `swap_chain` buffers than what gets passed to `SetMaximumFrameLatency`, but that was incorrect (you can have a higher number of swap_chain buffers than what you passed to `SetMaximumFrameLatency`). 

The issue appears to be something to do with the `resources` we use when creating the wgpu-hal dx12 `SwapChain` being `maximum_frame_latency` sized, but creating the raw `swap_chain` with `maximum_frame_latency + 1` `swap_chain_buffers`.

https://github.com/gfx-rs/wgpu/blob/adf1e3b6c546098e936f49602c4e39c0e416e1d6/wgpu-hal/src/dx12/mod.rs#L818

**Platform**
wgpu 0.19, windows 11, amd 6800xt

dtzxporter:
Fixes #5088. Even though we're telling DX12 that the maximum frame latency should be our non-padded value, the swap chain may request any of the buffers allocated to it.

**Connections**
Issue #5088.

**Description**
DX will still request any swap chain buffer that was allocated to it.

**Testing**
No longer crashes on my machine, checked swap chain indices requested and it's now correct.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-hal/src/dx12/mod.rs
+++ b/wgpu-hal/src/dx12/mod.rs
@@ -660,10 +660,14 @@ impl crate::Surface<Api> for Surface {
 
         let non_srgb_format = auxil::dxgi::conv::map_texture_format_nosrgb(config.format);
 
+        // The range for `SetMaximumFrameLatency` is 1-16 so the maximum latency requested should be 15 because we add 1.
+        // https://learn.microsoft.com/en-us/windows/win32/api/dxgi/nf-dxgi-idxgidevice1-setmaximumframelatency
+        debug_assert!(config.maximum_frame_latency <= 15);
+
         // Nvidia recommends to use 1-2 more buffers than the maximum latency
         // https://developer.nvidia.com/blog/advanced-api-performance-swap-chains/
         // For high latency extra buffers seems excessive, so go with a minimum of 3 and beyond that add 1.
-        let swap_chain_buffer = (config.maximum_frame_latency + 1).min(3);
+        let swap_chain_buffer = (config.maximum_frame_latency + 1).min(16);
 
         let swap_chain = match self.swap_chain.write().take() {
             //Note: this path doesn't properly re-initialize all of the things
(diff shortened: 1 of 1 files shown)


Elabajaba:
Fixes #5088 for me.

nical (approved):

torokati44:
Can we expect `wgpu-hal` `0.19.1` with this in sometime soon? 👀😶

cwfitzgerald:
Yes - within the next day or so

torokati44:
Well this was quick! c:

PR #5098, opened 2024-01-19
Impl. const. eval. for yet more trivial component-wise numeric built-ins in WGSL

ErichDonGubler:
**Connections**

Resolves many more of the now-trivial-to-implement built-ins, as an iteration over #4879.

- [x] Hard dependency on #5103.

**Description**

See above.

**Testing**

The current form of this PR depends on our snapshot test coverage for `naga`. I don't think it's complete; I'd be open to adding more coverage as part of this PR.

**Checklist**

- [x] Run `cargo fmt`.
  - ~~`--target wasm32-unknown-unknown`~~
  - ~~`--target wasm32-unknown-emscripten`~~


naga/src/proc/constant_evaluator.rs
```diff
@@ -862,6 +888,86 @@ impl<'a> ConstantEvaluator<'a> {
...
+            }
+            crate::MathFunction::Fract => {
+                component_wise_float!(self, span, [arg], |e| { Ok([e.fract()]) })
+            }
+            crate::MathFunction::Fma => {
+                component_wise_float!(
+                    self,
+                    span,
+                    [arg, arg1.unwrap(), arg2.unwrap()],
+                    |e1, e2, e3| { Ok([e1 * e2 + e3]) }
```
```suggestion
                    |e1, e2, e3| { Ok([e1.mul_add(e2, e3)]) }
```
The function could be faster and the result could be more accurate.

naga/src/proc/constant_evaluator.rs
```diff
@@ -862,6 +888,86 @@ impl<'a> ConstantEvaluator<'a> {
...
+                component_wise_float!(self, span, [arg], |e| { Ok([e.exp()]) })
+            }
+            crate::MathFunction::Exp2 => {
+                component_wise_float!(self, span, [arg], |e| { Ok([e.exp2()]) })
+            }
+            crate::MathFunction::Floor => {
+                component_wise_float!(self, span, [arg], |e| { Ok([e.floor()]) })
+            }
+            crate::MathFunction::Fract => {
+                component_wise_float!(self, span, [arg], |e| { Ok([e.fract()]) })
```
```suggestion
                component_wise_float!(self, span, [arg], |e| { Ok([e - e.trunc()]) })
```
Rust's `.fract()` impl seems to be `e - e.floor()` but the WGSL spec says it should be `e - e.trunc()`.

teoxoy (changes requested):
Regarding f1d7cc646713f2dd3c0f79ce80594c2f7d084198, can we instead keep them in the order they are defined? We usually do that for enum variants and struct fields throughout the codebase. I don't know why the WGSL spec went with alphabetic order; I find it makes things harder to find.

naga/src/proc/constant_evaluator.rs
ErichDonGubler, line 1: RE: [alphabetization/order of `match` arms](https://github.com/gfx-rs/wgpu/pull/5098#pullrequestreview-1836080682): I will see about matching variant declaration order.👍🏻

naga/src/proc/constant_evaluator.rs
ErichDonGubler: Derp derp derp, I even knew that this was a thing on the Rust side too, and I failed to connect the dots. Good catch! 😅 Fixed up with 038b0d560.

naga/src/proc/constant_evaluator.rs
ErichDonGubler: uh

The source for [`f32::fract`](https://doc.rust-lang.org/stable/src/std/f32.rs.html#165-167) and [`f64::fract`](https://doc.rust-lang.org/stable/src/std/f64.rs.html#165-167) both contain `self - self.trunc()` as the body. So, your problem statement doesn't seem valid.

~~I feel this is conclusively resolved, but I'll let you hit `Resolve` just in case.~~ EDIT: Oh, you have to `Approve` still, anyways. I'll hit `Resolve` now, I guess.

naga/src/proc/constant_evaluator.rs
ErichDonGubler: I have implemented this in 119a549cf, which includes some commentary that mirrors the “section” groupings of variants in the declaration.

naga/src/proc/constant_evaluator.rs
teoxoy: Sorry, I mistyped the comment. Rust does use the truncated version but the WGSL spec expects the floored version (i.e. the opposite of my previous comment).

ErichDonGubler: uh

I don't think that that's right, either. In the WGSL spec. (and other places AFAIK), `trunc` goes _towards 0_ on the number scale. IOW, that means `floor`ing when positive, and `ceil`ing when negative. This is what the Rust's `trunc` implements, too, i.e., [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7a8316d77509c504ba2d60bb54df428f):

```rust
assert_eq!(1.25f32.trunc(), 1.0);
assert_eq!(-3.14f32.trunc(), -3.0); 
```

From [the spec. text on `truncate`](https://gpuweb.github.io/gpuweb/wgsl/#truncate):

> The truncate function is defined over real numbers x extended with +∞ and −∞:
> 
> -   truncate(+∞) = +∞
>     
> -   truncate(−∞) = −∞
>     
> -   for real number x, computes the nearest whole number whose absolute value is less than or equal to the absolute value of x:
>     
>     -   truncate(x) = ⌊x⌋ if x ≥ 0, and ⌈x⌉ if x < 0.

Meanwhile, if we were to always `floor`, the spec. would say, `truncate(x) = ⌊x⌋` with no condition noted. If written as Rust (rather than WGSL), that would mean this shouldn't panic:

```rust
assert_eq!(1.25f32.trunc(), 1.0);
assert_eq!(-3.14f32.trunc(), -4.0); // RHS changed
```

ErichDonGubler: oderp, I got lost looking at the definition of the `truncate` logic and `fn trunc`, and forgot to connect it back to the actual matter at hand (which is `fract`). You're right, and I'm a ditz. Fixing now!

ErichDonGubler: Fixed up to be `e - e.floor()` with 7e78aca19. Sorry, it took me a lot longer than I care to admit to really process what you were saying! Thanks for catching this. 😀💦

teoxoy: No worries, thanks for addressing all the feedback! :)

teoxoy (approved):
Good stuff!

Issue #5071, opened 2024-01-16
Make error plumbing less public in `wgpu-core`


udoprog:
**Is your feature request related to a problem? Please describe.**

Thoughts while looking over issues like #5066

The internals of every error [is currently part of the public API surface of this project](https://docs.rs/wgpu-core/latest/wgpu_core/?search=Error). I'd propose as a policy making these details private unless an error variant or field propagates communicates something which is publicly actionable (See `SurfaceError` example below).

I believe this is better for at least two reasons:

* The API documentation is more terse, allowing users to easily spot what is and isn't important.
* Errors can be internally refactored to a greater extent without fear of breaking external use, which will be very useful in a post 1.0 world. While I believe all types are marked `#[non_exhaustive]`, this still does not permit variants or fields to be removed.

Note that all errors should continue to support `std::fmt::Display` and `std::error::Error::source` since this allows for a user to print decently rich diagnostics and error chains:

```
An foo error happened.

Caused by: Bar tried to bar
Caused by: Baz tried to baz
```

An example of an error which is publicly actionable is [`SurfaceError::Outdated`](https://docs.rs/wgpu/latest/wgpu/enum.SurfaceError.html#variant.Outdated), which signals that a surface needs to be reconfigured.

**Describe the solution you'd like**

The pattern I usually deploy is something like this:

```
// relevant derives
pub struct SomeError {
     kind: SomeErrorKind,
}

// relevant derives
pub(crate) enum SomeErrorKind {
    /* .. internal details ... */
}
```

For e.g. SurfaceError, I'd probably convert the existing error variants to a simple:

```rust
impl SurfaceError {
    pub fn is_outdated(&self) -> bool {
        /* ... */
    }
}
```

Or, if there is public interest in more than one state:

```rust
impl SurfaceError {
    pub fn kind(&self) -> SurfaceErrorKind {
        /* ... */
    }
}
```

Like with [`io::Error::kind`](https://doc.rust-lang.org/src/std/io/error.rs.html#910), the publicly exported `SurfaceErrorKind` doesn't have to match the internal representation if a representation is available which is more efficient than the publicly exported type.

Wumpf:
> I'd propose as a policy making these details private unless an error variant or field propagates communicates something which is publicly actionable

It's quite hard to tell though when something is actionable and when not. In the "worst case" somebody might want to count errors of a kind and then suddenly all internals are actionable since it's hard to keep statistics otherwise.

[...]

PR #5099, opened 2024-01-19
Add a wgpu-core feature flag for turning device lost errors into panics

nical:
This is intended to help flush out invalid API usage when testing wgpu. For example Firefox's fuzzing infrastructure only detects bugs that cause the process crash.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/Cargo.toml
+++ b/wgpu-core/Cargo.toml
@@ -36,6 +36,9 @@ api_log_info = []
 ## Log resource lifecycle management at info instead of trace level.
 resource_log_info = []
 
+## Panic when the device is lost.
+device_lost_panic = []
+
 ## Use static linking for libraries. Disale to manually link. Enabled by default.
 link = ["hal/link"]
 
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -438,7 +438,13 @@ pub enum DeviceError {
 impl From<hal::DeviceError> for DeviceError {
     fn from(error: hal::DeviceError) -> Self {
         match error {
-            hal::DeviceError::Lost => DeviceError::Lost,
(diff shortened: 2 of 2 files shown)


nical:
If there is appetite for it we could also add the feature to wgpu (this PR only exposes it in wgpu-core).

udoprog:
A cfg seems better suited for this `--cfg wgpu_panic`, since feature flags can be transitively activated by dependencies and this wouldn't be one I'd expect we ever want ending up in user facing code unless explicitly opted for.

Exposing more error details in error handling and optionally capturing backtraces (e.g. [through the `Context` in my error proposal](https://github.com/gfx-rs/wgpu/issues/5071#issuecomment-1894699650)) would probably allow for this kind of instrumentation without the need to introduce ad hoc code like this.

cwfitzgerald:
This is fine as an interim solution, but this feels like it should be able to be accomplished by registering a device loss callback which panics? Have we hooked that up yet?

> A cfg seems better suited for this --cfg wgpu_panic, since feature flags can be transitively activated by dependencies and this wouldn't be one I'd expect we ever want ending up in user facing code unless explicitly opted for.

Very possibly - I'm not too worried about people actually enabling it though. I don't think we should expose this to the users (see the comment about a device loss callback) but this is likely easier for firefox to setup.

cwfitzgerald (approved):
Conditionally approving with consideration of the above points

nical:
Fair point. I was a bit lazy but I might as well hook up the device loss plumbing and do it via the callback. I'll try to get that done first. I'll keep this solution at hand and land it if we need it before I get to do it properly.

cwfitzgerald:
Sounds good, I'm going to demote to draft in the mean time

PR #5100, opened 2024-01-19
[d3d12] use plane 1 for stencil only views

### LINKED ISSUE #5097: d3d12: Device loss due to invalid usage of CreateShaderResourceView
**Description**

**Repro steps**

The following triggers a device loss in Firefox:

```
<script>
document.addEventListener("DOMContentLoaded", async () => {
    const adpt = await self.navigator.gpu.requestAdapter({})
    const dev = await adpt.requestDevice({})
    const tex = dev.createTexture({
        "size": {"width": 88},
        "format": "stencil8",
        "usage": GPUTextureUsage.TEXTURE_BINDING
    })
    tex.createView()
    dev.createSampler({})
})
</script>
```

Log:

```
D3D12 ERROR: ID3D12Device::CreateShaderResourceView: The Plane Slice 0 cannot be used when the resource format is R24G8_TYPELESS and the view format is X24_TYPELESS_G8_UINT.  See documentation for the set of valid view format names for this resource format, determining which how the resource (or part of it) will appear to shader. [ STATE_CREATION ERROR #29: CREATESHADERRESOURCEVIEW_INVALIDVIDEOPLANESLICE]
D3D12: Removing Device.
D3D12 ERROR: ID3D12Device::RemoveDevice: Device removal has been triggered for the following reason (DXGI_ERROR_INVALID_CALL: There is strong evidence that the application has performed an illegal or undefined operation, and such a condition could not be returned to the application cleanly through a return code). [ EXECUTION ERROR #232: DEVICE_REMOVAL_PROCESS_AT_FAULT]
Exception thrown at 0x00007FFA9C07CF19 in firefox.exe: Microsoft C++ exception: _com_error at memory location 0x0000001DE87F9A80.
Exception thrown at 0x00007FFA9C07CF19 in firefox.exe: Microsoft C++ exception: _com_error at memory location 0x0000001DE87F9DF8.
Exception thrown at 0x00007FFA9C07CF19 in firefox.exe: Microsoft C++ exception: _com_error at memory location 0x0000001DE87FAB80.
```

**Platform**

Windows with d3d12

teoxoy:
Fixes https://github.com/gfx-rs/wgpu/issues/5097.

### DIFF
--- a/tests/tests/root.rs
+++ b/tests/tests/root.rs
@@ -34,6 +34,7 @@ mod shader;
 mod shader_primitive_index;
 mod shader_view_format;
 mod texture_bounds;
+mod texture_view_creation;
 mod transfer;
 mod vertex_indices;
 mod write_texture;
--- a/tests/tests/texture_view_creation.rs
+++ b/tests/tests/texture_view_creation.rs
@@ -1,9 +1,14 @@
 use wgpu::*;
-use wgpu_test::{gpu_test, GpuTestConfiguration};
+use wgpu_test::{gpu_test, FailureCase, GpuTestConfiguration, TestParameters};
 
 #[gpu_test]
-static STENCIL_ONLY_VIEW_CREATION: GpuTestConfiguration =
-    GpuTestConfiguration::new().run_async(|ctx| async move {
+static STENCIL_ONLY_VIEW_CREATION: GpuTestConfiguration = GpuTestConfiguration::new()
+    .parameters(
+        TestParameters::default()
+            .skip(FailureCase::webgl2()) // WebGL doesn't have stencil only views
+            .limits(wgpu::Limits::downlevel_defaults()),
+    )
+    .run_async(|ctx| async move {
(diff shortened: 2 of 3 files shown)


cwfitzgerald (approved):
Let's land this as it's, but could we get a test which makes a texture view of every type for each texture - for that matter, why wasn't this caught by a clear test for the stencil aspect?

teoxoy:
> for that matter, why wasn't this caught by a clear test for the stencil aspect?

I think that's because we clear textures with depth/stencil formats with render passes where we use a different format.

I added a test for depth only and stencil only view creation.

ErichDonGubler:
CI for Windows [failed](https://github.com/gfx-rs/wgpu/actions/runs/7589090955/job/20673359740?pr=5100), but only on artifact upload, and not tests running. I think this is good to merge, then!

ErichDonGubler (approved):

PR #5101, opened 2024-01-19
Impl. const. eval. for "first bit" numeric built-ins

ErichDonGubler:
**Connections**

Soft dependency on https://github.com/gfx-rs/wgpu/pull/5098.

**Description**

I think it's self-evident, but I'm happy to write something longer if it's not.

**Testing**

Some testing is evident from changes to Naga's snapshot tests, but there's not full coverage of these. I would accept a request to add tests before merging.

**Checklist**

- [x] Run `cargo fmt`.
  - ~~`--target wasm32-unknown-unknown`~~
  - ~~`--target wasm32-unknown-emscripten`~~


cwfitzgerald:
Any chance of this being revived?

ErichDonGubler:
@cwfitzgerald: It's not dead, just sleeping! 😉 The only thing this PR needs is prioritization from the Firefox side. I'll tag @jimblandy for comment.

ErichDonGubler:
Rebased.

naga/src/proc/constant_evaluator.rs
```diff
@@ -2096,6 +2104,182 @@ impl<'a> ConstantEvaluator<'a> {
...
+    // NOTE: Bit indices for this built-in start at 0 at the "right" (or LSB). For example, a value
+    // of 1 means the least significant bit is set. Therefore, an input of `0x[80 00…]` would
+    // return a right-to-left bit index of 0.
+    let trailing_zeros_to_bit_idx = |e: u32| -> u32 {
+        match e {
+            idx @ 0..=31 => idx,
+            32 => u32::MAX,
+            _ => panic!("invalid `trailing_zeros` values"),
+        }
+    };
```
ErichDonGubler: suggestion: I think this could be used in the body of `first_leading_bit` too, instead of branching on `0 | -1`?

teoxoy (approved):
Nice!

naga/src/proc/constant_evaluator.rs
teoxoy: That would work, we'd only branch once on `e.is_positive() || e == 0`.

ErichDonGubler: Just pushed up e3293f8fcb66 as a simplification, and tests still pass. According to [Godbolt](https://godbolt.org/z/YxTz4T8fv), it reduced instruction count, yay!

teoxoy: Looks good, let's land it!

teoxoy:
Just rebased this via GH's UI since it was running into timeouts that got fixed in https://github.com/gfx-rs/wgpu/pull/6030.

PR #5104, opened 2024-01-20
Error on missing web_sys_unstable_apis

### LINKED ISSUE #5095: `webgpu` feature fails to compile
**Repro steps**
Create a new crate with wgpu 0.19 as a dependency:
```toml
[dependencies]
wgpu = "0.19"
```

Build it with `cargo build --target wasm32-unknown-unknown`

It fails to build due to missing features on web_sys:
e.g.
```
error[E0412]: cannot find type `GpuBuffer` in crate `web_sys`
    --> /home/rukai/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.19.0/src/backend/webgpu.rs:3480:29
     |
3480 |     fn new(buffer: web_sys::GpuBuffer, desc: &crate::BufferDescriptor<'_>) -> Self {
     |                             ^^^^^^^^^ not found in `web_sys`
```

I could probably enable those features in web_sys myself but wgpu should really do that for me.

If I instead do:
```toml
[dependencies]
wgpu = { version = "0.19", default-features = false }
```

It will build successfully.

**Platform**
* Linux
* rust 1.75

rukai:
closes https://github.com/gfx-rs/wgpu/issues/5095

I thought about this some more and I think we can and should just give the user a helpful error when they are missing the unstable cfg.
With this we no longer need to take up prime documentation real estate with it.
I imagine resources like sotrh's learn wgpu will push users towards using `wgpu = { .. default-features = false, features = ["webgl"] }` to make it easy for users to run on the web and then this error will catch the user when they get that slightly wrong.

**Testing**

I created a seperate crate that depends on wgpu and tested the following cases:
This compiles successfully:
```toml
wgpu = {version = "0.19", path = "", features = ["webgl"], default-features = false}
```

This compiles successfully:
```toml
wgpu = {version = "0.19", path = "", default-features = false}
```

This fails to compile with the new error:
```toml
wgpu = {version = "0.19", path = "" }
```

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -5,3 +5,6 @@ xtask = "run --manifest-path xtask/Cargo.toml"
 rustflags = [
 "--cfg=web_sys_unstable_apis"
 ]
+rustdocflags = [
+"--cfg=web_sys_unstable_apis"
+]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ env:
   RUST_BACKTRACE: full
   PKG_CONFIG_ALLOW_CROSS: 1 # allow android to work
   RUSTFLAGS: --cfg=web_sys_unstable_apis -D warnings
-  RUSTDOCFLAGS: -Dwarnings
+  RUSTDOCFLAGS: --cfg=web_sys_unstable_apis -D warnings
   WASM_BINDGEN_TEST_TIMEOUT: 300 # 5 minutes
   CACHE_SUFFIX: c # cache busting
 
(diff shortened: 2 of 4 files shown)


rukai:
I will investigate the CI failure.

rukai:
`cargo doc --target wasm32-unknown-unknown` was failing in CI due to triggering the compile failure.
I resolved it by setting rustdocflags the same as we set for rustflags: https://doc.rust-lang.org/cargo/reference/config.html#buildrustdocflags
This fixed the issue locally.

Then I had to set RUSTDOCFLAGS env var in the CI script since it had a custom value that was overwriting rustdocflags config.
This fixed the issue in CI.

Then finally I changed the compile_error message to include setting rustdocflags since if we need to set it to get docs working then we should recommend that users set it too.

cwfitzgerald (approved):
This is great and a really novel solution I hadn't even considered!

Issue #5105, opened 2024-01-20
High type complexity around identifiers and a plan to reduce it


udoprog:
In `wgpu-core`, identifiers are currently generated through a framework which tries to maximize flexibility.

The way you specify which identifiers to generate is by implementing the `IdentityHandlerFactory` and `GlobalIdentityHandlerFactory` traits. By providing an implementation of `autogenerate_ids`, you can decide whether identifiers should be generated internally or whether they can be provided externally which happens by mapping a generic input through `input_to_id`.

The signature for external identifiers in `Global` is the `Input<G, Identifier>` type alias. This expands to `<G as IdentityHandlerFactory<I>>::Input`.

`G` is an implementation of `GlobalIdentityHandlerFactory`, which has the following signature:

```rust
pub trait GlobalIdentityHandlerFactory:
    IdentityHandlerFactory<id::AdapterId>
    /* ~20 more trait bounds, one for each identifier */
    + IdentityHandlerFactory<id::SurfaceId>
{
}
```

It extends one `IdentityHandlerFactory` trait for every kind of wgpu-core identifier. The intent here is that you theoretically should be able to build an implementation of `GlobalIdentityHandlerFactory` which has full discretion over how any kind of identifier.

In practice, this system can only be fulfilled through something akin to the following implementation (taken from player):

```rust
pub struct IdentityPassThroughFactory;

impl<I: wgc::id::TypedId> wgc::identity::IdentityHandlerFactory<I> for IdentityPassThroughFactory {
    type Input = I;

    fn input_to_id(id_in: Self::Input) -> I {
        id_in
    }

    fn autogenerate_ids() -> bool {
        false
    }
}
impl wgc::identity::GlobalIdentityHandlerFactory for IdentityPassThroughFactory {}
```

This implements `GlobalIdentityHandlerFactory`, because there is an IdentityPassThroughFactory wilcard implementation of all of the above `IdentityHandlerFactory<T>` implementations.

Next let's look at the `TypeId` trait. It restricts which kinds of identifiers can be constructed through this:

```rust
pub trait TypedId: Copy + Debug + Any + 'static + WasmNotSendSync + Eq + Hash {
    fn zip(index: Index, epoch: Epoch, backend: Backend) -> Self;
    fn unzip(self) -> (Index, Epoch, Backend);
    fn into_raw(self) -> NonZeroId;
}
```

[...]

PR #5108, opened 2024-01-21
Increase the consistency of identifiers in wgpu-core

udoprog:
**Connections**
#5105

**Description**
This introduces the following:

**Add the `Marker` associated type to `Resource`**

Instances of the `Marker` trait indicates a unique (non-generic) type for each resource that is used as the type parameter to typed identifiers.

Each distinct resources gets their own instance, for example:
* `wgpu_core::id::markers::Device`
* `wgpu_core::id::markers::Surface`

Note that these *do not* have a `A: HalApi` type argument.

* This gets rid of a large number of generic `Id` parameters and is correct since resources knows which identifier they use.
* This gets rid of [these type aliases which incorporate the `Dummy` parameter](https://github.com/gfx-rs/wgpu/blob/f9509bcf9ec2b63a64eb7fea93f7f44cd5ae4d2e/wgpu-core/src/id.rs#L237).
* We avoid the scenario where `Id<Device<A>>` is not the same type as `Id<Device<Dummy>>`.

**Remove aliased identifiers and introduce explicit transmutes**

This ensures that each resource have a unique identifier which have distinct types. In cases where identifiers are shared they have to be explicitly transmuted (e.g. `DeviceId` <-> `QueueId`). Legal transmutations are controlled through the implementation of a `T: Transmute<U>` trait where `T: Marker, U: Marker`. The intent is that these transmutations should be removed.

**Remove `TypedId` trait**

`wgpu-core` doesn't need to be *fully* generic over identifiers. It's sufficient for it to be generic over `Id<T>` where `T: Marker`. This also means that the methods that used to be associated with `TypeId` (`zip`, `unzip`, `into_raw`) can be used directly on the imported identifiers instead of having to import the trait.

**Incremental introduction of `RawId`**

There is currently a slightly odd scenario. `Id<T>` construction is unsafe, but `Id<T>` also can also implement `Serialize`, `Deserialize`, and be constructed through `zip` / `unzip`. The intent with the introduction of `RawId` is to eventually create a clear distinction between "raw identifiers" which can safely be constructed, manipulated, and serialized and "typed identifiers" which guarantees to the greatest extent possible that the referenced resource exists.

**Future changes**

Once things have been sorted out, I expect that:
* The serialization of Id<T> is removed.
* Id<T> can only be constructed through unsafe constructors. While I don't think this involves any memory safety, this currently seems to be the status quo, so enforcing that is a good thing.
* Transmutes for device and queue are removed (once multiple queues can be constructed and have their own proper resource).

**Testing**

Build, clippy, and `xtask test`.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`. If applicable.


nical:
The change looks good to me. I'll  merge it early next week so that it coincides with me updating gecko's wgpu revision.

udoprog:
Neat! I rebased the changes to latest trunk for the occasion.

If you need it to be a merge commit instead for clarity I can undo the rebase and do that instead, but the changes should be the same.

nical (approved):

Issue #5090, opened 2024-01-18
Crashes on the d3d12 backend when creating a lot of texture views


nical:
Bugzilla entry: https://bugzilla.mozilla.org/show_bug.cgi?id=1863872

jimblandy:
Closed as dup of #5079

PR #5125, opened 2024-01-23
d3d12: Propagate errors when closing command lists

nical:
**Connections**

I run into this fairly consistently while debugging #5090

**Description**

Before this commit, command lists that we failed to close were used anyway during submit, causing device loss.

This commit propagates the error and removes the now unused closed field in CommandBuffer.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


teoxoy (approved):
LGTM

PR #5126, opened 2024-01-23
Remove the lock around TextureView::parent

nical:
**Description**

We don't need to remove the reference during triage anymore, instead the texture's reference count will be decremented automatically when the texture view is dropped. So `parent` can be immutable and no longer needs a lock.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


cwfitzgerald (approved):
Love to see it!

PR #5129, opened 2024-01-24
Join threads in `MULTITHREADED_COMPUTE` example.

### LINKED ISSUE #4285: Multithreaded Compute Segfaults in CI on GL/Linux
Randomly, but consistently within the run, multithreaded compute segfaults in CI. This doesn't happen when I run it locally against llvmpipe. This doesn't happen when running it in CI on windows against llvmpipe.

I suspect this might be an issue with EGL, but I have no idea how to reproduce it.

[ci-log.txt](https://github.com/gfx-rs/wgpu/files/13066029/ci-log.txt)

### LINKED ISSUE #5084: Intermittent crashes in Vulkan tests on Linux
Many wgpu tests crash intermittently at exit on Linux.

Comments copied from #4285, which I inadvertently crashed (as in "crashed a party"):

----

I'm looking at one interesting crash that occurs in exactly the scenario described by [this GLIBC comment](https://github.com/bminor/glibc/blob/db57da5be27a0887ce5131c071e2240f5c17aae1/stdlib/cxa_finalize.c#L50C62-L52C64):

```
	    /* We don't want to run this cleanup more than once.  The Itanium
	       C++ ABI requires that multiple calls to __cxa_finalize not
	       result in calling termination functions more than once.  One
	       potential scenario where that could happen is with a concurrent
	       dlclose and exit, where the running dlclose must at some point
	       release the list lock, an exiting thread may acquire it, and
	       without setting flavor to ef_free, might re-run this destructor
	       which could result in undefined behaviour.  Therefore we must
	       set flavor to ef_free to avoid calling this destructor again.
```

The comment is explaining why `__exit_funcs_lock` needs to be held and why the `flavor` field needs to be set to `ef_free`. The code does do these things, so I don't think that's the bug.

But I bring this up because I have some core dumps in which this is exactly what's happening: one thread has this stack frame:

```
#1  0x00007f26284e85e7 in __do_global_dtors_aux () from /lib64/libVkLayer_khronos_validation.so
```

indicating that the validation layer DSO is being closed down, while another thread has these stack frames:

```
#11 0x00007f268757dfd6 in __run_exit_handlers (status=0, listp=<optimized out>, run_list_atexit=run_list_atexit@entry=true, run_dtors=run_dtors@entry=true) at exit.c:111
#12 0x00007f268757e11e in __GI_exit (status=<optimized out>) at exit.c:141
#13 0x00007f2687565151 in __libc_start_call_main (main=main@entry=0x55dd5c35ef80 <main>, argc=argc@entry=2, argv=argv@entry=0x7fff46fd4c18) at ../sysdeps/nptl/libc_start_call_main.h:74
#14 0x00007f268756520b in __libc_start_main_impl (main=0x55dd5c35ef80 <main>, argc=2, argv=0x7fff46fd4c18, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, 
    stack_end=0x7fff46fd4c08) at ../csu/libc-start.c:360
#15 0x000055dd5c34f655 in _start ()
```

indicating that this is the main thread, which has returned from `main` and is traversing the same list of cleanup handles as the other thread.

-----

In another core dump, we have one thread in the midst of unmapping the Vulkan validation library (based on locals at frame #10):

```
#0  0x00007f50a90350eb in __GI_munmap () at ../sysdeps/unix/syscall-template.S:117
#1  0x00007f50a902840a in _dl_unmap_segments (l=0x7f509c08fb00) at ./dl-unmap-segments.h:32
#2  _dl_unmap (map=map@entry=0x7f509c08fb00) at ../sysdeps/x86_64/tlsdesc.c:31
#3  0x00007f50a901503d in _dl_close_worker (map=<optimized out>, map@entry=0x7f509c08fb00, force=force@entry=false) at dl-close.c:628
#4  0x00007f50a901569b in _dl_close (_map=0x7f509c08fb00) at dl-close.c:793
#5  0x00007f50a9014523 in __GI__dl_catch_exception (exception=exception@entry=0x7f4f5c5f6090, operate=0x7f50a9015660 <_dl_close>, args=0x7f509c08fb00) at dl-catch.c:237
#6  0x00007f50a9014679 in _dl_catch_error (objname=0x7f4f5c5f60f8, errstring=0x7f4f5c5f6100, mallocedp=0x7f4f5c5f60f7, operate=<optimized out>, args=<optimized out>) at dl-catch.c:256
#7  0x00007f50a8aa8143 in _dlerror_run (operate=<optimized out>, args=<optimized out>) at dlerror.c:138
#8  0x00007f50a8aa7e76 in __dlclose (handle=<optimized out>) at dlclose.c:31
#9  0x00007f50a017565c in loader_platform_close_library (library=<optimized out>) at /usr/src/debug/vulkan-loader-1.3.268.0-1.fc39.x86_64/loader/vk_loader_platform.h:381
#10 loader_delete_layer_list_and_properties (inst=inst@entry=0x7f509c09f460, layer_list=layer_list@entry=0x7f509c0a0788)
    at /usr/src/debug/vulkan-loader-1.3.268.0-1.fc39.x86_64/loader/loader.c:493
#11 0x00007f50a017cc2d in loader_delete_layer_list_and_properties (layer_list=<optimized out>, inst=<optimized out>)
    at /usr/src/debug/vulkan-loader-1.3.268.0-1.fc39.x86_64/loader/loader.c:720
#12 vkDestroyInstance (pAllocator=0x0, instance=<optimized out>) at /usr/src/debug/vulkan-loader-1.3.268.0-1.fc39.x86_64/loader/trampoline.c:796
#13 vkDestroyInstance (instance=<optimized out>, pAllocator=0x0) at /usr/include/vulkan/vulkan_core.h:4124
```

while another thread is, again, running the destructor list:

```
#0  std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >::~vector () at /usr/include/c++/13/bits/stl_vector.h:730
#1  0x00007f50a8a5efd6 in __run_exit_handlers (status=0, listp=<optimized out>, run_list_atexit=run_list_atexit@entry=true, run_dtors=run_dtors@entry=true) at exit.c:111
#2  0x00007f50a8a5f11e in __GI_exit (status=<optimized out>) at exit.c:141
#3  0x00007f50a8a46151 in __libc_start_call_main (main=main@entry=0x55ae5a3bb340 <main>, argc=argc@entry=4, argv=argv@entry=0x7ffced56f6c8) at ../sysdeps/nptl/libc_start_call_main.h:74
```

jimblandy:
Join all threads before returning from the test case, to ensure that we don't return from `main` until all open `Device`s have been dropped.

This avoids a race condition in glibc in which a thread calling `dlclose` can unmap a shared library's code even while the main thread is still running its finalization functions. (See #5084 for details.) Joining all threads before returning from the test ensures that the Vulkan loader has finished `dlclose`-ing the Vulkan validation layer shared library before `main` returns.

Remove `skip` for this test on GL/llvmpipe. With this change, that has not been observed to crash. Without it, the test crashes within ten runs or so.

Fixes #5084.
Fixed #4285.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,10 @@ Bottom level categories:
 
 - In Surface::configure and Surface::present, fix the current GL context not being unset when releasing the lock that guards access to making the context current. This was causing other threads to panic when trying to make the context current. By @Imberflur in [#5087](https://github.com/gfx-rs/wgpu/pull/5087).
 
+#### Tests
+
+- Fix intermittent crashes on Linux in the `multithreaded_compute` test. By @jimblandy in [#5129](https://github.com/gfx-rs/wgpu/pull/5129).
+
 ## v0.19.0 (2024-01-17)
 
 This release includes:
--- a/examples/src/hello_compute/tests.rs
+++ b/examples/src/hello_compute/tests.rs
@@ -57,9 +57,7 @@ static MULTITHREADED_COMPUTE: GpuTestConfiguration = GpuTestConfiguration::new()
         TestParameters::default()
             .downlevel_flags(wgpu::DownlevelFlags::COMPUTE_SHADERS)
             .limits(wgpu::Limits::downlevel_defaults())
-            .skip(FailureCase::adapter("V3D"))
-            // Segfaults on linux CI only https://github.com/gfx-rs/wgpu/issues/4285
-            .skip(FailureCase::backend_adapter(wgpu::Backends::GL, "llvmpipe")),
+            .skip(FailureCase::adapter("V3D")),
     )
     .run_sync(|ctx| {
         use std::{sync::mpsc, sync::Arc, thread, time::Duration};
@@ -69,24 +67,30 @@ static MULTITHREADED_COMPUTE: GpuTestConfiguration = GpuTestConfiguration::new()
         let thread_count = 8;
 
         let (tx, rx) = mpsc::channel();
-        for _ in 0..thread_count {
-            let tx = tx.clone();
-            let ctx = Arc::clone(&ctx);
-            thread::spawn(move || {
-                let input = &[100, 100, 100];
-                pollster::block_on(assert_execute_gpu(
-                    &ctx.device,
-                    &ctx.queue,
-                    input,
-                    &[25, 25, 25],
-                ));
-                tx.send(true).unwrap();
-            });
-        }
+        let workers: Vec<_> = (0..thread_count)
+            .map(move |_| {
+                let tx = tx.clone();
+                let ctx = Arc::clone(&ctx);
+                thread::spawn(move || {
+                    let input = &[100, 100, 100];
+                    pollster::block_on(assert_execute_gpu(
+                        &ctx.device,
+                        &ctx.queue,
+                        input,
+                        &[25, 25, 25],
+                    ));
+                    tx.send(true).unwrap();
+                })
+            })
+            .collect();
 
         for _ in 0..thread_count {
             rx.recv_timeout(Duration::from_secs(10))
                 .expect("A thread never completed.");
+        }
+
+        for worker in workers {
+            worker.join().unwrap();
         }
     });
 


cwfitzgerald (approved):
Noice

Issue #5079, opened 2024-01-17
Eagerly destroy texture views and bind groups


nical:
When a texture or buffer is destroyed we need to get rid of the hal resources for associated bind groups and texture views to avoid issues in GC'd environments.

nical:
Bugzilla entry: https://bugzilla.mozilla.org/show_bug.cgi?id=1863872

PR #5131, opened 2024-01-24
Snatch texture views of destroyed textures

nical:
**Connections**

Addresses the TextureView part of #5079.

**Description**

Keep track of the texture views associated with a texture and snatch/destroy them along with the texture.
This prevents garbage collected texture views from accumulating and maintaining resources alive. 

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


ErichDonGubler (approved):
LGTM!

PR #5133, opened 2024-01-24
Differentiate between device loss caused by error/drop

nical:
**Description**

We need this in order to help with detecting internal errors during testing/fuzzing. In other words we'd like fuzzing to detect when the device loss reason is `Unknown` (interpreted as an internal error) without treating dropped devices as an error.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


teoxoy (commented):
Do we need the new `DeviceLostReason::Dropped` variant? Wouldn't reusing `DeviceLostReason::Destroyed` on drop achieve the same result?

nical:
In the browser we can't pass `Destroyed` to the callback unless the script explicitly destroyed the device. Otherwise it makes GC observable.

teoxoy:
Ah, I see - thanks for the clarification!

teoxoy (approved):

PR #5134, opened 2024-01-24
[d3d12] Avoid panic on instance drop

### LINKED ISSUE #4604: panic! when Instance is destroyed
**Description**
When an Instance is created with flags set to wgpu::InstanceFlags::DEBUG or wgpu::InstanceFlags::DISCARD_HAL_LABELS and later goes out of scope and rust is running in Debug mode it panics with the message:

thread 'main' panicked at 'attempt to subtract with overflow', C:\Users\rasmu\.cargo\registry\src\github.com-1ecc6299db9ec823\wgpu-hal-0.18.0\src\auxil\dxgi\exception.rs:34:5

It does not panic with the flag wgpu::InstanceFlags::VALIDATION

If a Surface is created using the Instance it does not panic until the surface is destroyed.

**Repro steps**
Ideally, a runnable example we can check out.

```
fn main() {
        backends: wgpu::Backends::all(),
        flags: wgpu::InstanceFlags::DEBUG,
        dx12_shader_compiler: Default::default(),
        gles_minor_version: wgpu::Gles3MinorVersion::Automatic,
    });
}
```

**Platform**
OS: Windows 11
System Type: x64-based PC
Processor: AMD Ryzen 5 4500U with Radeon Graphics, 2375 Mhz, 6 Core(s), 6 Logical Processor(s)
GPU: AMD Radeon(TM) Graphics
wgpu version 0.18.0

hakolao:
**Connections**
Attempts to fix #4604

**Description**
`register_exception_handler` is only called when `InstanceFlags::VALIDATION` are present. Thus `unregister` should also be only called then.

```rust
    if instance_flags.contains(wgt::InstanceFlags::VALIDATION) {
        //...
        super::exception::register_exception_handler();
    }
```

**Testing**
Tested on one of my sample projects...

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,9 @@ Bottom level categories:
 
 ### Bug Fixes
 
+#### General
+- Fix `panic!` when dropping `Instance` without `InstanceFlags::VALIDATION`. By @hakolao in [#5134](https://github.com/gfx-rs/wgpu/pull/5134)
+
 #### WGL
 
 - In Surface::configure and Surface::present, fix the current GL context not being unset when releasing the lock that guards access to making the context current. This was causing other threads to panic when trying to make the context current. By @Imberflur in [#5087](https://github.com/gfx-rs/wgpu/pull/5087).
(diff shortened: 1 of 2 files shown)


nical:
Nice!

nical (approved):

nical:
Looks like you need to remove that `bitflags::Flags` import.

PR #5137, opened 2024-01-24
Discard draws when the instance/vertex/index count is zero

### LINKED ISSUE #1879: Assertion failure when using Metal GPU capture
**Description**
While trying to take a GPU capture in Xcode I'm encountering an assertion failure:

```objective-c
-[MTLDebugRenderCommandEncoder validateCommonDrawErrors:instanceCount:baseInstance:maxVertexID:]:4185: failed assertion `instanceCount(0) must be non-zero.'
```

**Platform**
macos 11.5.2 (20G95)
wgpu 0.10.1

nical:
**Connections**

Fixes #1879

**Description**

Discard draws when the instance/vertex/index count is zero

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -514,9 +514,12 @@ impl RenderBundleEncoder {
                         })
                         .map_pass_err(scope);
                     }
-                    commands.extend(state.flush_vertices());
-                    commands.extend(state.flush_binds(used_bind_groups, base.dynamic_offsets));
-                    commands.push(command);
+
+                    if instance_count > 0 && vertex_count > 0 {
+                        commands.extend(state.flush_vertices());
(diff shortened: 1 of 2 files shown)


ErichDonGubler (approved):
I _think_ this is still compliant based on my read of [`GPURenderCommandsMixin.draw`](https://gpuweb.github.io/gpuweb/#dom-gpurendercommandsmixin-draw), though I'm surprised that there's no validation/mention of this case for this. LGTM, but I wanted to loop in @teoxoy or @jimblandy to make sure I'm not missing anything with my lesser familiarity with the spec.

nical:
There is no validation simply because it is valid. The reason we have to discard these empty draws is to work around the backend having issues with it.

cwfitzgerald:
Could we add documentation to hal that these are not valid calls to make?

Arguably they should be valid calls and this be handled by the metal backend?

PR #5141, opened 2024-01-26
Avoid leaking submitted command encoders

### LINKED ISSUE #5860: Submitting command encoders caused memory leak
**Description**
When I iteratively create command encoders and submit them to queue , memory leak occurred. This has been occurring since wgpu 0.20. Not occurring until 0.19.

**Repro steps**
1. create Instance, Adapter, Queue.
2. create a encoder and submit in loop.
3. Application uses more cpu memory as iterations progress.

Reproducible example:
`

    #[tokio::main]
    async fn main() {
        let instance = wgpu::Instance::default();

        let adapter = instance
            .request_adapter(&wgpu::RequestAdapterOptions::default())
            .await
            .unwrap();

        let (device, queue) = adapter
            .request_device(
                &wgpu::DeviceDescriptor {
                    label: None,
                    required_features: wgpu::Features::empty(),
                    required_limits: wgpu::Limits::downlevel_defaults(),
                },
                None,
            )
            .await
            .unwrap();

        loop {
            let encoder = device.create_command_encoder(&wgpu::CommandEncoderDescriptor {
                label: Some("debug encoder"),
            });
            let command_buffer = encoder.finish();

            queue.submit(Some(command_buffer));

            device.poll(wgpu::Maintain::Wait);
        }
    }
`

Expected : Memory keeps stable.
observed behavior : Memory usage increases

**Platform**
OS: Windows 11
wgpu: 0.20

nical:
**Connections**

Part of #4660 got [accidentally clobbered by arcanization](https://github.com/gfx-rs/wgpu/pull/3626/files?file-filters%5B%5D=.rs&show-viewed-files=true#diff-620fd073d10cd5d6302171f28dd7db7ddf66d6b78eca56b2fd90cc01d318ed9fL2348) 

**Description**

unlike `wgpu-core`, `wgpu` consumes command buffers during submit so it needs to drop them

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -154,6 +154,7 @@ By @atlv24 in [#5383](https://github.com/gfx-rs/wgpu/pull/5383)
 - Ensure render pipelines have at least 1 target. By @ErichDonGubler in [#5715](https://github.com/gfx-rs/wgpu/pull/5715)
 - `wgpu::ComputePass` now internally takes ownership of `QuerySet` for both `wgpu::ComputePassTimestampWrites` as well as timestamp writes and statistics query, fixing crashes when destroying `QuerySet` before ending the pass. By @wumpf in [#5671](https://github.com/gfx-rs/wgpu/pull/5671)
 - Validate resources passed during compute pass recording for mismatching device. By @wumpf in [#5779](https://github.com/gfx-rs/wgpu/pull/5779)
+- Fix a `CommandBuffer` leak. By @cwfitzgerald and @nical in [#5141](https://github.com/gfx-rs/wgpu/pull/5141)
 
 #### DX12
 
--- a/wgpu/src/backend/wgpu_core.rs
+++ b/wgpu/src/backend/wgpu_core.rs
@@ -2326,10 +2326,16 @@ impl crate::Context for ContextWgpuCore {
             .map(|(i, _)| i)
             .collect::<SmallVec<[_; 4]>>();
 
-        match wgc::gfx_select!(*queue => self.0.queue_submit(*queue, &temp_command_buffers)) {
+        let index = match wgc::gfx_select!(*queue => self.0.queue_submit(*queue, &temp_command_buffers)) {
             Ok(index) => index,
             Err(err) => self.handle_error_fatal(err, "Queue::submit"),
+        };
+
+        for cmdbuf in &temp_command_buffers {
+            wgc::gfx_select!(*queue => self.0.command_buffer_drop(*cmdbuf));
         }
+
+        index
     }
 
     fn queue_get_timestamp_period(
--- a/wgpu/src/backend/wgpu_core.rs
+++ b/wgpu/src/backend/wgpu_core.rs
@@ -2326,7 +2326,8 @@ impl crate::Context for ContextWgpuCore {
             .map(|(i, _)| i)
             .collect::<SmallVec<[_; 4]>>();
 
-        let index = match wgc::gfx_select!(*queue => self.0.queue_submit(*queue, &temp_command_buffers)) {
+        let index = match wgc::gfx_select!(*queue => self.0.queue_submit(*queue, &temp_command_buffers))
+        {
             Ok(index) => index,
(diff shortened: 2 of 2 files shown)


cwfitzgerald (approved):
Nice

nical:
It breaks the multithreaded_compute test.

nical:
I suspect that the error is that multithreaded_compute is racing to create command buffers in the same slot:

```
[2024-01-26T09:34:35Z TRACE wgpu_core::storage] User is inserting CommandBufferId(0,1,gl)  <---
[2024-01-26T09:34:35Z TRACE wgpu_core::storage] User is inserting CommandBufferId(0,2,gl)  <---
[2024-01-26T09:34:35Z TRACE wgpu_core::storage] User is inserting CommandBufferId(1,1,gl)
[2024-01-26T09:34:35Z TRACE wgpu_core::storage] User is inserting CommandBufferId(2,1,gl)
[2024-01-26T09:34:35Z TRACE wgpu_core::storage] User is inserting CommandBufferId(3,1,gl)
[2024-01-26T09:34:35Z TRACE wgpu_core::storage] User is inserting CommandBufferId(4,1,gl)
[2024-01-26T09:34:35Z TRACE wgpu_core::storage] User is inserting CommandBufferId(5,1,gl)
[2024-01-26T09:34:35Z TRACE wgpu_core::storage] User is inserting CommandBufferId(6,1,gl)
[2024-01-26T09:34:35Z TRACE wgpu_core::storage] User is removing CommandBufferId(0,1,gl) <---
[2024-01-26T09:34:35Z TRACE wgpu_core::storage] User is removing CommandBufferId(0,2,gl) <--- Boom.

```

nical:
What's happening is:
 - the command buffer is used in a submission,
   - it snatches it from the registry and leaves its spot with an occupied error state.
   - the command buffer's `ResourceInfo` gets dropped which frees the ID too soon since the registry still has a spot for the command buffer and we haven't dropped the command buffer.
 - After that a variety of things can go wrong depending on scheduling, but the one I'm observing most often is a new command buffer gets placed in the yet-to-be-dropped spot in the registry.


So we have to ensure that the `ResourceInfo` doesn't get dropped before we've had a chance to call `command_buffer_drop` and clear out the spot in the registry.

nical:
Unfortunately, trying to keep an Arc to the resource in the Destroyed variant of the registry causes `CommandBuffer::from_arc_into_baked` to panic because it expects to have the only arc left to the command buffer.  There's a tension between this code using the reference counting to signify actual usage of the resource and the `ResourceInfo` info lifetime (tied to the resource itself), determining how IDs can be reused in the registry.

We could put the `ResourceInfo` into an arc of its own, but that would have to be done for all resources because of all of the generics involved, that's not great. We could also devise another way to decide when to reuse IDs. I would prefer it to stay close to the registry itself instead of being driven by the resource reference counts. Ideally IDs would be freed by the `Gloabal::<resource>_drop` functions, but for it to work well we need to remove the remaining uses of IDs during command recording otherwise one could drop/re-create a resource used by a command encoder before submit.

nical:
Another solution that doesn't require as much refactoring is to make command buffers snatchable. We probably should not use the `Snatchable` type for this since snatching a resource while anything wgpu-related is happening on another thread is guaranteeing lock contention. But a mutex should suffice since when we are extracting the command buffer we already expect to be the only users.

Wumpf:
@nical with the other things you landed by now, has the situation changed much on how to fix this?

Snatchability out of the registry makes the most sense to me since this best represents what should be happening (the queue submissing takes over the lifetime of the actual resource, from a hub perspective it's already dead), but then again I'm not 100% sure I've been following correctly

nical:
I'm note sure, this is completely paged out of my brain. I'll try to rebase it and see if it breaks.

teoxoy:
For tracking purposes: This was unblocked by https://github.com/gfx-rs/wgpu/pull/5244.

PR #5144, opened 2024-01-26
Fix the validation of vertex/index/instance ranges in render bundles

nical:
**Connections**

Spec: https://gpuweb.github.io/gpuweb/#dom-gpurendercommandsmixin-draw

**Description**

The current bundle validation has two problems:
 - division by zero if the vertex stride is zero (we hit this in CTS runs in Firefox)
 - It is stricter than the specification by requesting that the buffer has space for `stride * count` whereas the specified formula is `stride * (count - 1) + last_stride` where `last_stride` is the space required to fit the last element (potentially smaller than the stride.

This PR addresses these issues by implementing the spec pretty much verbatim, but for render bundles only. Regular encoders are unfortunately still not up to spec. One way would be to use the same validation but it will be more expensive since it has to be done for each draw call while the current validation gets away with doing most of the calculation only when switching pipelines. I think that I can adapt that to match the spec but it's going to be more complicated.

I'm eager to get at least the render bundle part fixed so that we get rid of an outstanding CTS crash.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -96,7 +96,7 @@ use crate::{
     hub::Hub,
     id::{self, RenderBundleId},
     init_tracker::{BufferInitTrackerAction, MemoryInitKind, TextureInitTrackerAction},
-    pipeline::{self, PipelineFlags, RenderPipeline},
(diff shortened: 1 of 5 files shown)


teoxoy (approved):
Looks solid!

PR #5148, opened 2024-01-26
feat: make `ProgrammableStage::entry_point` optional

### LINKED ISSUE #5145: `GPUProgrammableStage.entryPoint` should be optional
In https://github.com/gpuweb/gpuweb/pull/4387, the WebGPU API spec was changed such that, if there's only one appropriate entry point in the shader module, pipeline creation will just use that one automatically: you can omit its name from the `GPUProgrammableStage`  when you create the pipeline.

This is a cute readability hack for small (and maybe not-so-small) examples, so folks are excited about using it in the WebGPU examples repo: https://github.com/webgpu/webgpu-samples/pull/334

In our code, this would mean that `wgpu_core::pipeline::ProgrammableStageDescriptor::entry_point` would become an `Option`, consumers would use reflection to find the right entry point, and then callers in `wgpu`/Firefox/etc. would be changed to accommodate.

Since `wgpu_core` takes care of producing a `naga::Module`, it has enough information to choose an entry point before passing the module to `wgpu_hal`, so `wgpu_hal` can continue to require a specific name.

ErichDonGubler:
**Connections**

Resolves #5145.

- [x] Soft (?) dependency on https://github.com/rustwasm/wasm-bindgen/pull/3816.
- [x] Hard dependency on <https://github.com/gfx-rs/wgpu/pull/5306>.

**Description**

**Testing**

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/benches/benches/computepass.rs
+++ b/benches/benches/computepass.rs
@@ -242,7 +242,7 @@ impl ComputepassState {
                     label: Some("Compute Pipeline"),
                     layout: Some(&pipeline_layout),
                     module: &sm,
-                    entry_point: "cs_main",
+                    entry_point: Some("cs_main"),
                     compilation_options: wgpu::PipelineCompilationOptions::default(),
                     cache: None,
                 });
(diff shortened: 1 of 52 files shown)


ErichDonGubler:
Status: I have what I believe to be working changes in `wgpu-core`. I need to propagate this to `wgpu`. After that, this needs some test coverage.

ErichDonGubler:
- [x] TODO: I'm going to split out `wgpu-core` support from this PR, since Firefox likely can use that sooner than `wgpu`. That will also make the set of changes to review here easier, I think!

nical (approved):

PR #619, opened 2020-04-28
Tracing Infrastructure

kvark:
Closes #289
Downstream changes: https://github.com/gfx-rs/wgpu-rs/pull/291 https://github.com/gfx-rs/wgpu-native/pull/17
Note: close to half of the new LOCs belong to `Cargo.lock`, which we are now including since we have a binary target crate `player`.

## Why do we need that?

Have you ever seen a case that doesn't work for somebody? Often, reproducing the complete setup requires elaborate toolchains, external resources, and closed-source code. Even with all of that, the difference between platforms may hide the issue. [RenderDoc](https://renderdoc.org/) or Metal captures do not help, since they are extremely driver/hardware-dependent.

This problem is amplified for WebGPU implementation in Gecko. Some workloads are rendered incorrectly, and debugging them within the browser becomes too much of a challenge.

Temporary limitations (can be fixed later):
  - replaying has to happen on the same backend. This is easy to work around by just mass-replacing the backend in the trace (i.e. `s/Vulkan/Metal/g`)
  - buffer readbacks are not recorded

Tracing allows us to swiftly isolate and reproduce rendering issues. We can now consider a [Warden](https://github.com/gfx-rs/gfx/pull/1589)-like testing and benchmarking framework built around the traces.

## What the hell it is?

This is a state of art tracing infrastructure built into `wgpu-core`. It traces all the commands and resource changes, very similarly to `apitrace`. Tracing code is enabled by a "trace" feature, and the actual support is enabled at the device instantiation, controlled at run-time. There is no run-time overhead for running with tracing feature but in disabled state.

The result of tracing is a directory, containing the following files:
  - "trace.ron" is basically a list of all the actions taken by the app, in [RON](https://github.com/ron-rs/ron) format.
  - each buffer update, or a shader module created, is placed into a separate binary file referenced by the trace.

The new "player" application can be used to replay the traces. It's launched as:
```rust
target/debug/player <trace-dir>
```
When built with "winit" feature, it's able to replay the workloads that operate on a swapchain. It renders each frame consequently, then waits for the user to close the window. When built without "winit", it launches in console mode and can replay any trace that doesn't use swapchains.

## How does it look?

Having the trace in RON makes it easy to inspect visually (by humans!). There is no separate "dump" phase. Each submission is labelled, and the trace is easy to navigate. Moreover, everything is human-editable: you can tweak it to see what changes would be needed in an application in order to fix a rendering issue. For example, you can change the blend modes and re-play.

```ron
[...]

PR #5149, opened 2024-01-27
Feature/serde feature

### LINKED ISSUE #3609: wgpu-types error when using `serde` feature without tracing feature

KirmesBude:
**Connections**
#3609

**Description**
Allows using feature `serde` on `wgpu`, `wgpu-core` and `wgpu-types`.
`wgpu-types` no longer has `trace` and `replay` features, but an implicit `serde` feature.
`wgpu-core` has an explicit `serde` feature and removes the `serial-pass` feature.
`wgpu` now has an explicit `serde` feature.

Almost all occurrences of `cfg_attr(feature = "trace")` and `cfg_attr(feature = "replay")` have been replaced with `cfg_attr(feature = "serialize")` and `cfg_attr(feature = "deserialize")` respectively.
For `wgpu` and `wgpu-core` the usage for `trace` and `replay`  should not have changed. If you are using `serial-pass` you should use `serde` instead. For `wgpu-types` if you have previously enabled the `trace` or `replay` feature you need to migrate to `serde`.

**Testing**
`cargo test --package <package> --features <feature>`
for the following packages: `wgpu-types`, `wgpu-core`, `wgpu`
for the following features: `"None"`, `serde`, (`trace`, `replay`)[if available] [*no combinations]

**Compile timings**
Comparing 273684f1b with f748c43 on debug.
With default features takes about 25 sec.
With serde takes around the same time (38 sec).
With deserialize (on first commit) takes about the same amount of time (38 sec).
With serialize takes about 27 sec.

**Checklist**

- [x] Run `cargo fmt`.
  - [x] `--target wasm32-unknown-emscripten` (fails somewhere in winit)
- [x] Run `cargo xtask test` to run tests. (yes, and it fails in some places, no sure what to do with the results)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,12 +67,15 @@ Bottom level categories:
         - `tan`
         - `tanh`
 - Eager release of GPU resources comes from device.trackers. By @bradwerth in [#5075](https://github.com/gfx-rs/wgpu/pull/5075)
+- `wgpu`, `wgpu-core` and `wgpu-types` now have `serialize` and `deserialize` features. By @KirmesBude in [#5149](https://github.com/gfx-rs/wgpu/pull/5149)
+- `wgpu-types`'s `trace` and `replay` features have been replaced by `serialize` and `deserialize` respectively. By @KirmesBude in [#5149](https://github.com/gfx-rs/wgpu/pull/5149)
 
 
 ### Bug Fixes
 
 #### General
 - Fix `panic!` when dropping `Instance` without `InstanceFlags::VALIDATION`. By @hakolao in [#5134](https://github.com/gfx-rs/wgpu/pull/5134)
+- Fix `serde` feature not compiling for `wgpu-types`. By @KirmesBude in [#5149](https://github.com/gfx-rs/wgpu/pull/5149)
 
 #### WGL
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,8 +67,8 @@ Bottom level categories:
         - `tan`
         - `tanh`
 - Eager release of GPU resources comes from device.trackers. By @bradwerth in [#5075](https://github.com/gfx-rs/wgpu/pull/5075)
-- `wgpu`, `wgpu-core` and `wgpu-types` now have `serialize` and `deserialize` features. By @KirmesBude in [#5149](https://github.com/gfx-rs/wgpu/pull/5149)
 - `wgpu-types`'s `trace` and `replay` features have been replaced by `serialize` and `deserialize` respectively. By @KirmesBude in [#5149](https://github.com/gfx-rs/wgpu/pull/5149)
+- `wgpu-core`'s `serial-pass` feature has been removed. Use `serde` instead. By @KirmesBude in [#5149](https://github.com/gfx-rs/wgpu/pull/5149)
 
 
 ### Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,7 +67,7 @@ Bottom level categories:
         - `tan`
         - `tanh`
 - Eager release of GPU resources comes from device.trackers. By @bradwerth in [#5075](https://github.com/gfx-rs/wgpu/pull/5075)
-- `wgpu-types`'s `trace` and `replay` features have been replaced by `serialize` and `deserialize` respectively. By @KirmesBude in [#5149](https://github.com/gfx-rs/wgpu/pull/5149)
+- `wgpu-types`'s `trace` and `replay` features have been replaced by the `serde` feature. By @KirmesBude in [#5149](https://github.com/gfx-rs/wgpu/pull/5149)
 - `wgpu-core`'s `serial-pass` feature has been removed. Use `serde` instead. By @KirmesBude in [#5149](https://github.com/gfx-rs/wgpu/pull/5149)
 
 
--- a/wgpu-core/Cargo.toml
+++ b/wgpu-core/Cargo.toml
@@ -48,7 +48,7 @@ renderdoc = ["hal/renderdoc"]
 strict_asserts = ["wgt/strict_asserts"]
 
 ## serde
-serde = ["serialize", "deserialize", "wgt/serde"]
+serde = ["serialize", "deserialize", "serial-pass", "wgt/serde"]
 
 ## serialize
 serialize = ["dep:serde", "wgt/serialize", "arrayvec/serde"]
@@ -63,7 +63,7 @@ trace = ["ron", "serialize", "naga/serialize"]
 replay = ["deserialize", "naga/deserialize"]
 
 ## Enable serializable compute/render passes, and bundle encoders.
-serial-pass = ["serde", "wgt/serde"]
+serial-pass = ["dep:serde", "wgt/serde", "arrayvec/serde"]
 
 ## Enable `ShaderModuleSource::Wgsl`
 wgsl = ["naga/wgsl-in"]
--- a/wgpu-core/Cargo.toml
+++ b/wgpu-core/Cargo.toml
@@ -51,10 +51,10 @@ strict_asserts = ["wgt/strict_asserts"]
 serde = ["serialize", "deserialize", "serial-pass", "wgt/serde"]
 
 ## Enables serialize functionality on common wgpu types.
-serialize = ["dep:serde", "wgt/serialize", "arrayvec/serde"]
+serialize = ["dep:serde", "wgt/serde", "arrayvec/serde"]
 
 ## Enables deserialize functionality on common wgpu types.
-deserialize = ["dep:serde", "wgt/deserialize", "arrayvec/serde"]
+deserialize = ["dep:serde", "wgt/serde", "arrayvec/serde"]
 
 ## Enable API tracing.
 trace = ["ron", "serialize", "naga/serialize"]
--- a/wgpu-core/Cargo.toml
+++ b/wgpu-core/Cargo.toml
@@ -47,7 +47,7 @@ renderdoc = ["hal/renderdoc"]
 ## to the validation carried out at public APIs in all builds.
 strict_asserts = ["wgt/strict_asserts"]
 
-## Enables serde functionality on common wgpu types.
+## Enables serialization via `serde` on common wgpu types.
 serde = ["dep:serde", "wgt/serde", "arrayvec/serde"]
 
 ## Enable API tracing.
--- a/wgpu-core/Cargo.toml
+++ b/wgpu-core/Cargo.toml
@@ -51,19 +51,19 @@ strict_asserts = ["wgt/strict_asserts"]
 serde = ["serialize", "deserialize", "wgt/serde"]
 
 ## serialize
-serialize = ["dep:serde", "wgt/serialize"]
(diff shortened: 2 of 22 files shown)


wgpu/Cargo.toml
```diff
@@ -78,13 +78,22 @@ strict_asserts = ["wgc?/strict_asserts", "wgt/strict_asserts"]
...
 api_log_info = ["wgc/api_log_info"]
 
+## serde
+serde = ["serialize", "deserialize", "wgc/serde"]
+
+## serialize
+serialize = ["dep:serde", "wgc/serialize"]
+
+## deserialize
+deserialize = ["dep:serde", "wgc/deserialize"]
```
Wumpf: Needs proper documentation here and in `wgpu/src/lib.rs`. What we'd want here is an explanation how the features related to each other and what they do

wgpu-core/Cargo.toml
```diff
@@ -47,14 +47,23 @@ renderdoc = ["hal/renderdoc"]
...
 strict_asserts = ["wgt/strict_asserts"]
 
+## serde
+serde = ["serialize", "deserialize", "wgt/serde"]
+
+## serialize
+serialize = ["dep:serde", "wgt/serialize", "arrayvec/serde"]
+
+## deserialize
+deserialize = ["dep:serde", "wgt/deserialize", "arrayvec/serde"]
```
Wumpf: see comment on wgpu (toml/lib)

wgpu-core/Cargo.toml
```diff
@@ -47,14 +47,23 @@ renderdoc = ["hal/renderdoc"]
...
-trace = ["ron", "serde", "wgt/trace", "arrayvec/serde", "naga/serialize"]
+trace = ["ron", "serialize", "naga/serialize"]
 
 ## Enable API replaying
-replay = ["serde", "wgt/replay", "arrayvec/serde", "naga/deserialize"]
+replay = ["deserialize", "naga/deserialize"]
 
 ## Enable serializable compute/render passes, and bundle encoders.
-serial-pass = ["serde", "wgt/serde", "arrayvec/serde"]
+serial-pass = ["serde", "wgt/serde"]
```
Wumpf: this seams wrong: it now pulls in serialization/deserialization for all types when it should enable serialization only for compute/render pass.
I think this needs to be instead
`serial-pass = ["dep:serde", "wgt/serde", "arrayvec/serde"]`
and `serde` should imply `serial-pass` for completeness

wgpu-core/src/command/compute.rs
```diff
@@ -40,14 +40,8 @@ use std::{fmt, mem, str};
...
-#[cfg_attr(
-    any(feature = "serial-pass", feature = "trace"),
-    derive(serde::Serialize)
-)]
-#[cfg_attr(
-    any(feature = "serial-pass", feature = "replay"),
-    derive(serde::Deserialize)
-)]
+#[cfg_attr(feature = "serialize", derive(serde::Serialize))]
+#[cfg_attr(feature = "deserialize", derive(serde::Deserialize))]
```
Wumpf: serial pass should stick around since it enables specifically pass (de)serialization
Applies to all removals of `serial-pass` as far as I can tell

Wumpf (changes requested):
Please add a changelog entry, fairly important to highlight this feature config change :)

Handling of `serial-pass` looks broken to me but I'm a bit on the fence whether we still need this one 🤔.
I'm actually overall here with what @cwfitzgerald's [comment](https://github.com/gfx-rs/wgpu/issues/3609#issuecomment-1505804861) on the issue and would suggest joining all of `serde`, `serial-pass`, `serialize` and `deserialize` to one feature, I doubt anyone needs this level of granularity realistically.

Anyways, thanks so much for looking into this! Happy to merge once the `serial-pass` story, docs and changelog is sorted out! :)

KirmesBude:
> Please add a changelog entry, fairly important to highlight this feature config change :)
> 
> Handling of `serial-pass` looks broken to me but I'm a bit on the fence whether we still need this one 🤔. I'm actually overall here with what @cwfitzgerald's [comment](https://github.com/gfx-rs/wgpu/issues/3609#issuecomment-1505804861) on the issue and would suggest joining all of `serde`, `serial-pass`, `serialize` and `deserialize` to one feature, I doubt anyone needs this level of granularity realistically.
> 
> Anyways, thanks so much for looking into this! Happy to merge once the `serial-pass` story, docs and changelog is sorted out! :)

Thanks for the heads up about `serial-pass`. Should be fixed now, I hope.
Wasn't sure where to put stuff for the Changelog. Let me know if the current version works for you.

KirmesBude:
Oh and about merging everything to serde. I am also fine with that. Just let me know and I will try to do the necessary changes.

wgpu/Cargo.toml
```diff
@@ -78,13 +78,22 @@ strict_asserts = ["wgc?/strict_asserts", "wgt/strict_asserts"]
...
 
+## Enables serde functionality on common wgpu types.
+serde = ["serialize", "deserialize", "wgc/serde"]
+
+## Enables serialize functionality on common wgpu types.
+serialize = ["dep:serde", "wgc/serialize"]
+
+## Enables deserialize functionality on common wgpu types.
+deserialize = ["dep:serde", "wgc/deserialize"]
+
```
Wumpf: unfortunately this doc / feature change needs to be copy pasted to here
https://github.com/gfx-rs/wgpu/blob/trunk/wgpu/src/lib.rs#L33
Same is true for the wgpu-core ones. (this should be done by `document-features` crate but we can't use this right now since it breaks vendoring for firefox builds)

Wumpf (approved):
thanks for the fixes! changelog looks great
Yeah if you don't mind changing it again right away I'd say we should go with just `serde`, it makes all the cfgs so much simpler and reduces the number of feature flags and thus possible permutations by a lot.

KirmesBude:
> thanks for the fixes! changelog looks great Yeah if you don't mind changing it again right away I'd say we should go with just `serde`, it makes all the cfgs so much simpler and reduces the number of feature flags and thus possible permutations by a lot.

Yeah, I will try that. On wgpu-types level it is trivial, but I am not sure how to approach it for wgpu-core and wgpu.
`serial-pass` is separate but should be part of `serde`. For that reason `trace`/`replay`can not imply `serde` and I can not use `serde` in the source code.

Edit: I could replace `feature = "serialize"` with `any(feature = "trace", feature = "serde")` and `feature = "deserialize"` with `any(feature = "replay", feature = "serde")`. Is that what you had in mind?

Wumpf:
> Edit: I could replace feature = "serialize" with any(feature = "trace", feature = "serde") and feature = "deserialize" with any(feature = "replay", feature = "serde"). Is that what you had in mind?

I was thinking simpler still that both `replay` and `trace` would just imply the `serde` feature and all checks where we decide whether a thing implements the serde serialization/deserialization would only check for the serde feature

Wumpf:
fyi I just posted about this also on the wgpu Matrix chat room just in case anyone has a good reason that we're not seeing right now on why deserialize & serialize should stay separate :)
https://matrix.to/#/!FZyQrssSlHEZqrYcOb:matrix.org/$DnwtH2xxPKoTq7OjVYEVoyUHzZMR3mVRIAvOhy3_GeI?via=matrix.org&via=mozilla.org&via=envs.net

KirmesBude:
> > Edit: I could replace feature = "serialize" with any(feature = "trace", feature = "serde") and feature = "deserialize" with any(feature = "replay", feature = "serde"). Is that what you had in mind?
> 
> I was thinking simpler still that both `replay` and `trace` would just imply the `serde` feature and all checks where we decide whether a thing implements the serde serialization/deserialization would only check for the serde feature

With your earlier comment about `serial-pass` this means that `trace`/`replay` implies `serde` which implies `serial-pass`. So `trace`/`replay` implies `serial-pass`. Is that fine?

Wumpf:
Ah apologies for the confusion! If we go with `serde` for both serialize & deserialize then `serial-pass` should also go away I figure. Essentially we'd bulldozer everything over to only have `serde`, `trace`->`serde`, `replay`->`serde`.
Apparently, the original concern was compile time performance https://github.com/gfx-rs/wgpu/pull/619#discussion_r417727801 but it looks like we don't have numbers for this and the setup with just `serde` implying both serialization and deserialization is much more common so I'll happily take all the blame (:

CHANGELOG.md
```diff
@@ -67,12 +67,15 @@ Bottom level categories:
         - `tan`
         - `tanh`
 - Eager release of GPU resources comes from device.trackers. By @bradwerth in [#5075](https://github.com/gfx-rs/wgpu/pull/5075)
+- `wgpu-types`'s `trace` and `replay` features have been replaced by `serialize` and `deserialize` respectively. By @KirmesBude in [#5149](https://github.com/gfx-rs/wgpu/pull/5149)
```
Wumpf: we removed `serialize` and `deserialize` now again, let's update :)

wgpu-core/src/command/compute.rs
```diff
@@ -40,14 +40,8 @@ use std::{fmt, mem, str};
...
-#[cfg_attr(
-    any(feature = "serial-pass", feature = "trace"),
-    derive(serde::Serialize)
-)]
-#[cfg_attr(
-    any(feature = "serial-pass", feature = "replay"),
-    derive(serde::Deserialize)
-)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize))]
```
Wumpf: nit, here and elsewhere this could go now on the same line :)

wgpu/Cargo.toml
```diff
@@ -78,6 +78,9 @@ strict_asserts = ["wgc?/strict_asserts", "wgt/strict_asserts"]
 ## Log all API entry points at info instead of trace level.
 api_log_info = ["wgc/api_log_info"]
 
+## Enables serde functionality on common wgpu types.
```
Wumpf: nit: 
```suggestion
## Enables serialization via `serde` on common wgpu types.
```

Wumpf (changes requested):
awesome, yeah this is much cleaner! Will merge after changelog hickup is resolved!

Wumpf (approved):

PR #5123, opened 2024-01-23
Support for I64 and I64 image atomics

JMS55:
**Connections**
Implements https://github.com/gfx-rs/wgpu/issues/5009

**Description**
Add support for U64/I64 types in shaders, and atomic operations on `storage_texture<u64/i64>` from within compute shaders.

**Testing**
TODO

**TODO**
- [ ] Support for R64Uint/R64Sint textures
- [ ] Naga frontend + validation for atomic ops on storage textures
- [ ] Tests

**Checklist**

- [ ] Run `cargo fmt`.
- [ ] Run `cargo clippy`. If applicable, add:
  - [ ] `--target wasm32-unknown-unknown`
  - [ ] `--target wasm32-unknown-emscripten`
- [ ] Run `cargo xtask test` to run tests.
- [ ] Add change to `CHANGELOG.md`. See simple instructions inside file.

Issue #4424, opened 2022-12-27
Support 8- and 16-bit integers


AveryanAlex:
I tried to load SPIRV compute shader compiled by rust-gpu with u64 integers and got validation error:

```
thread 'main' panicked at 'wgpu error: Validation Error

Caused by:
    In Device::create_shader_module
      note: label = `REDACTED/target/spirv-builder/spirv-unknown-vulkan1.1/release/deps/shader.spv`

Shader validation error:
  ┌─ REDACTED/target/spirv-builder/spirv-unknown-vulkan1.1/release/deps/shader.spv:1:1
  │
1 │
  │   naga::Constant [6]


    Constant [6] '' is invalid
    The type doesn't match the constant

', REDACTED/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.14.2/src/backend/direct.rs:2403:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Adding wgpu::Features::SHADER_FLOAT64 to features didn't help.

JMS55:
Would love both u64 support in shaders, as well as R64Uint textures.

tombh:
I think `u8` support is needed to allow `rust-gpu`'s enums to work? I can compile Rust's dataless enums into SPIR-V without issue, I just need to add the `Int8` capability to the `rust-gpu` compiler. But then when I try to load the shader into `wgpu`, I get: `Shader '' parsing error: InvalidTypeWidth(1)`. Does that seem right?

Is there an issue to track for this?

andyleiserson:
Retitled; 64-bit integer support was added in the PRs linked above.

xbwwj:
Why is this closed? It seems that shader-i8 is not yet supported in #9412 @cwfitzgerald

cwfitzgerald:
Oops

PR #5154, opened 2024-01-27
Add shader I64 and U64 support

atlv24:
**Connections**
Others in this series: #5155
Addresses #4424

**Description**
Shaders currently do not support 64 bit integer types. Add support for these gated by a feature flag when the capability is present.

**Testing**
A snapshot test called i64.wgsl has been added to test parsing, type casting, and other basic operations.

**Checklist**

- [x] Run `cargo fmt`.


cwfitzgerald (dismissed):
Howdy! wgpu test fairy here: could you add execution tests for this in wgpu. We need at least one test, but there could be other things:
- struct layout should be tested in struct_layout.rs

wgpu-types/src/lib.rs
```diff
@@ -817,10 +817,16 @@ bitflags::bitflags! {
...
 
-        // 58 available
-
         // Shader:
 
+        /// Allows shaders to use i64 and u64.
+        ///
+        /// Supported platforms:
+        /// - Vulkan
+        /// - DX12
```
JMS55: Probably need to leave a note saying it requires DXC.

naga/src/front/spv/mod.rs
```diff
@@ -4876,6 +4876,11 @@ impl<I: Iterator<Item = u32>> Frontend<I> {
                 let low = self.next()?;
                 match width {
                     4 => crate::Literal::U32(low),
+                    8 => {
+                        inst.expect(5)?;
```
JMS55, line 4880: Why 5 here? What does this do?

wgpu-core/src/device/resource.rs
```diff
@@ -1493,6 +1493,10 @@ impl<A: HalApi> Device<A> {
             self.features
                 .contains(wgt::Features::SHADER_EARLY_DEPTH_TEST),
         );
+        caps.set(
```
JMS55, line 1514: @cwfitzgerald / wgpu maintainers: I'd love to expose this as a public function in wgpu. Bevy already has a very incomplete way to automatically map wgpu features -> naga caps, but we rarely remember to update it.

naga/src/front/spv/mod.rs
atlv24: Makes sure there's at least 4 bytes more to read for the second half of the u64 literal, and that we're not at the file end, if i understand correctly. Its a common pattern in this file

wgpu-core/src/device/resource.rs
atlv24: gonna make a followup for this

CHANGELOG.md
```diff
@@ -95,6 +95,7 @@ Bottom level categories:
   - As with other instance flags, this flag can be changed in calls to `InstanceFlags::with_env` with the new `WGPU_GPU_BASED_VALIDATION` environment variable.
 
   By @ErichDonGubler in [#5146](https://github.com/gfx-rs/wgpu/pull/5146), [#5046](https://github.com/gfx-rs/wgpu/pull/5046).
+- Signed and unsigned 64 bit integer support in shaders. By @rodolphito in [#5154](https://github.com/gfx-rs/wgpu/pull/5154)
```
cwfitzgerald:
```suggestion
- Signed and unsigned 64 bit integer support in shaders. By @rodolphito and @cwfitzgerald in [#5154](https://github.com/gfx-rs/wgpu/pull/5154)
```

😉

naga/src/valid/expression.rs
```diff
@@ -1332,12 +1334,32 @@ impl super::Validator {
...
+                                            scalar.width,
+                                        ));
+                                    }
+                                }
+                                _ => return Err(ExpressionError::InvalidArgumentType(fun, 0, arg)),
+                            },
+                            _ => return Err(ExpressionError::InvalidArgumentType(fun, 0, arg)),
+                        }
+                    }
+                    Mf::CountTrailingZeros => {
```
teoxoy: Considering these work in Metal and HLSL, we should error in the SPIR-V backend instead.

cwfitzgerald: I think that's a massive portability hazard - The plan is to polyfill these on spir-v as well (see #5276) so I think it's cleaner to just ban them everywhere until it is available everywhere we support 64bit.

teoxoy: That's what we usually do with things that don't yet work in specific backends but since this is an optional native feature I guess validating it for all backends is fine.

naga/src/valid/expression.rs
```diff
@@ -1332,12 +1334,32 @@ impl super::Validator {
                             _ => return Err(ExpressionError::InvalidArgumentType(fun, 0, arg)),
                         }
                     }
-                    Mf::CountTrailingZeros
-                    | Mf::CountLeadingZeros
```
teoxoy, line 1336: `CountTrailingZeros` with non 32 bit types should also not be allowed.

naga/src/back/spv/block.rs
```diff
@@ -1069,11 +1086,52 @@ impl<'w> BlockContext<'w> {
...
+                                arg0_id,
+                            ));
+
+                            MathOp::Custom(Instruction::ext_inst(
+                                self.writer.gl450_ext_inst_id,
+                                spirv::GLOp::FindILsb,
+                                result_type_id,
+                                id,
+                                &[rev_id],
+                            ))
```
teoxoy: I don't think we should add non-working code. Same above.

naga/src/valid/expression.rs
atlv24: I did not find any spec saying they are not supported for 64 bit operands, and it does work fine in my tests. Check out the two pages linked in #5276

naga/src/back/spv/block.rs
atlv24: Fair, I'll remove it later tonight.

naga/src/valid/expression.rs
teoxoy: The backend uses `FindILsb` for `CountTrailingZeros` which only works with 32 bit types.

teoxoy (approved):
LGTM

teoxoy:
This needs a rebase or another merge of trunk into it.

PR #5155, opened 2024-01-28
Refactor tests to be feature flag bit-width agnostic

atlv24:
**Connections**
Others in this series: #5154

**Description**
Refactor tests to be feature flag bit-width agnostic. This is to make it possible to easily change Features from u64 to u128 in the future. This will be needed because we are out of feature flag space to add new ones.

**Testing**
This only affects existing tests, and the tests still pass.

**Checklist**

- [x] Run `cargo fmt`.


cwfitzgerald (approved):
Nice, we've needed this for a while!

PR #5156, opened 2024-01-29
Fix the validation for vertex limits for regular render passes

nical:
**Connections**

Similar to #5144 but for regular render passes instead of bundles.

**Description**

The previous validation:
- was skipped when the vertex stride was 0,
- did not take the "last stride" (size of the last vertex) into account.

This test fixes both problems while (unlike #5144) staying in the spirit of how the validation was implemented (presumably for performance reasons) instead of sticking to how it is described in the specification.

**Testing**

Covered by the CTS.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


teoxoy (approved):
Good stuff!

PR #5162, opened 2024-01-30
Add support for f64 in vertex format

valaphee:
**Connections**
GL depends on https://github.com/grovesNL/glow/pull/275
https://github.com/gpuweb/gpuweb/issues/2805

**Description**
At the moment no backend seems to support f64 in vertex formats (even Vulkan). This pr tries to implement f64 for all backends except metal.

**Testing**

**Checklist**

- [ ] Run `cargo fmt`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1558,9 +1558,8 @@ checksum = "151665d9be52f9bb40fc7966565d39666f2d1e69233571b71b87791c7e0528b3"
 
 [[package]]
 name = "glow"
-version = "0.13.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bd348e04c43b32574f2de31c8bb397d96c9fcfa1371bd4ca6d8bdc464ab121b1"
+version = "0.13.0"
+source = "git+https://github.com/grovesNL/glow.git#e4a9a702111c425abdf9febef3ec2f2ce4822c26"
 dependencies = [
  "js-sys",
  "slotmap",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -151,7 +151,7 @@ hassle-rs = "0.11.0"
 
 # Gles dependencies
 khronos-egl = "6"
-glow = "0.13.1"
+glow = { git = "https://github.com/grovesNL/glow.git" }
 glutin = "0.29.1"
 
 # wasm32 dependencies
--- a/wgpu-hal/Cargo.toml
+++ b/wgpu-hal/Cargo.toml
@@ -105,7 +105,7 @@ rustc-hash = "1.1"
 log = "0.4"
 
 # backend: Gles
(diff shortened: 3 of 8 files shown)


valaphee:
On a different note, whats the minimum supported Desktop GL version? According to README its 3.3+ but this is most likely wrong as glVertexAttribFormat is only supported 4.3 upwards. There are also checks for 4.2 in the code.

As general everything before Windows 7 is not worth supporting anyway, as Firefox doesn't even support Windows XP anymore (and 3.3 is in the era of XP and Vista)

examples/src/cube/mod.rs
```diff
@@ -5,13 +5,13 @@ use wgpu::util::DeviceExt;
 #[repr(C)]
 #[derive(Clone, Copy, Pod, Zeroable)]
 struct Vertex {
-    _pos: [f32; 4],
+    _pos: [f64; 4],
```
cwfitzgerald: To be resolved when no longer draft: Instead of hijacking cube, could this be a set of tests in `tests/tests`

cwfitzgerald:
> On a different note, whats the minimum supported Desktop GL version?

It should be 3.3.

> According to README its 3.3+ but this is most likely wrong as glVertexAttribFormat is only supported 4.3 upwards. 

Those are only used if the [VERTEX_BUFFER_LAYOUT private cap is supported](https://github.com//gfx-rs/wgpu/blob/580340f2d3d9410573d2fab4c7090d7669fccefb/wgpu-hal/src/gles/command.rs#L122-L124). Which is [`supported((3, 1), (4, 3)) || extensions.contains("GL_ARB_vertex_attrib_binding")`](https://github.com//gfx-rs/wgpu/blob/b8f27c7284b95809e26069a9ace0282cc5904006/wgpu-hal/src/gles/adapter.rs#L581).

> As general everything before Windows 7 is not worth supporting anyway, as Firefox doesn't even support Windows XP anymore (and 3.3 is in the era of XP and Vista)

We definitely don't support Windows 7 in any meaningful way, but you can still have a 3.3 GPU on a Windows 10 machine. @Zoxc, who initially implemented desktop OpenGL support, ran it all the way back to 3.1 (with some limitations).

valaphee:
Ah thanks, didn't saw the VERTEX_BUFFER_LAYOUT private cap. I'll add a proper test shortly. Didn't meant to push the hijacked cube. 😅

DX12 doesn't seem to support 64 bit in vertex format entirely (no dxgi format), but shaders at least should support it (DoublePrecisionFloatShaderOps).

cwfitzgerald:
Do you still have interest in pursuing this?

valaphee:
Sure, I'll finish it up and add some tests, just waited for the OpenGL Non-ES backend to work on Linux (as I switched to Linux)

valaphee:
I should have rebased, oops...

Issue #4215, opened 2023-10-07
buffer_get_mapped_range different way of passing range than buffer_map_async


sagudev:
**Is your feature request related to a problem? Please describe.**
[`buffer_get_mapped_range`](https://docs.rs/wgpu-core/latest/wgpu_core/global/struct.Global.html#method.buffer_get_mapped_range) has offset and size while [`buffer_map_async`](https://docs.rs/wgpu-core/latest/wgpu_core/global/struct.Global.html#method.buffer_map_async) uses rust's range.

**Describe the solution you'd like**
To make things more concise and more up to specs `buffer_map_async` should have offset and size like in spec: https://gpuweb.github.io/gpuweb/#dom-gpubuffer-mapasync

**Describe alternatives you've considered**
Alternative would be to switch `buffer_get_mapped_range` to use range.

**Additional context**
This is all in wgpu-core, the goal is to have only one way of describing ranges.

cwfitzgerald:
I would personally prefer they both just use `Range`

andristarr:
Is this not already implemented? `buffer_get_mapped_range` in the core wrapper already uses range instead of the offset

sagudev:
> Is this not already implemented?

Not on wgpu-core (latest on crates): https://docs.rs/wgpu-core/latest/wgpu_core/global/struct.Global.html#method.buffer_get_mapped_range also not on trunk: https://github.com/gfx-rs/wgpu/blob/trunk/wgpu-core/src/device/global.rs#L2462

sagudev:
Done in https://github.com/gfx-rs/wgpu/pull/5185

PR #5163, opened 2024-01-30
WIP public api for buffer_get_mapped_range now uses range

andristarr:
**Connections**
https://github.com/gfx-rs/wgpu/issues/4215

**Description**
buffer_get_mapped_range uses a range instead of buffer + size

**Testing**

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/naga/src/proc/constant_evaluator.rs
+++ b/naga/src/proc/constant_evaluator.rs
@@ -837,7 +837,9 @@ impl<'a> ConstantEvaluator<'a> {
             ));
         }
 
+        // NOTE: We try to match the declaration order of `MathFunction` here.
         match fun {
+            // comparison
             crate::MathFunction::Abs => {
                 component_wise_scalar(self, span, [arg], |args| match args {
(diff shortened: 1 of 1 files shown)


nical (changes requested):
This is a departure from the WebGPU specification. If we are to do this, Let's keepthe change in wgpu only not change wgpu-core (which firefox uses in its WebGPU implementation).
~It loses the ability to compute the mapped size automatically, it's fine by me if restricted to wgpu but I wonder what @cwfitzgerald thinks about minor departures from the spec?~

Edit: Nevermind the spec thing, I missed the original issue before writing the comment above. My request about not modifying wgpu-core still stands.

andristarr:
I will reopen this in a few

andristarr:
Sorry, I am not exactly clear on this actually @nical .

Do we want to keep the scope of change only to the wgpu_core wrapper in the backend? If yes, `buffer_get_mapped_range` already consumes a range, unless I am missing something.

PR #5165, opened 2024-01-30
ci: Update to macos-14, xlarge runner is no longer needed

Dinnerbone:
**Connections**
https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

**Description**
The new macos-14 runner seems to be equivalent to macos-13-xlarge, save for less CPU cores. However it's faster than 13 (seemingly), so that might be okay.

It's not documented to have a GPU - but it does, and seems to run tests fine. @cwfitzgerald suggested to try and switch :D

Clippy macos goes from nearly 11m to 4m
GPU test macos goes from 2m to 3m. But it's free now!

**Testing**
n/a

**Checklist**

- [ ] Run `cargo fmt`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,18 +92,18 @@ jobs:
 
           # MacOS
           - name: MacOS x86_64
-            os: macos-12
+            os: macos-14
             target: x86_64-apple-darwin
             kind: native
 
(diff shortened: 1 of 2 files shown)


cwfitzgerald (approved):
Noice

PR #5166, opened 2024-01-30
Fix panic when creating a surface while no backend is available

Wumpf:
**Connections**
connected to 
* https://github.com/gfx-rs/wgpu/pull/4984

at first I thought I introduced the unwrap, but [it was already there](https://github.com/gfx-rs/wgpu/pull/4984/files#diff-1b40237cc7b42c2140c93d8910af3cb74461441d611f1db3f69e81617229eccbL539) (still guilty, why did I not forsee this -.-)

**Description**
Fixes crash via unwrap when trying to create a surface on an instance that has no workable backend availble (e.g. try to create a surface when an instance is limited to dx12 on a metal device)

**Testing**
It compiles and there's an `unwrap` less!

**Checklist**

- [x] Run `cargo fmt`.


nical (approved):

kpreid:
This fix wasn't backported to 0.19.3 which was released at the same time as the "needs back-porting" label was removed. Seems like an error.

Wumpf:
The label was removed because this is a breaking change, thus can't be back ported to 0.19.x

Issue #5142, opened 2024-01-26
Allow checking type of wgpu::Instance's context


Wumpf:
I ran into a situation where WebGPU _adapter_ creation fails and I want to retry with WebGL (Chrome Linux 121 advertises WebGPU support but can't create adapters -.-)

Turns out though it's actually not easily possible to programmatically determine that this was the error: If the instance is created with both WebGPU & WebGL enabled and adapter creation fails subsequently,it might mean that **either** of the context type (webgpu or wgpu-core) was picked and failed. Only in the case of a webgpu-instance do I want to retry instance creation.
One hacky way to check for this is the result of `as_hal`, but this method is only available when wgpu_core compilation is enabled, and in the context I'm in (egui's guts) I don't know this.

Proposing to add this simple method on ` wgpu::Instance` :
`is_webgpu_instance() -> bool`

PR #5167, opened 2024-01-30
Allow wgpu::Instance to report which backends were enabled

Wumpf:
**Connections**
In spirit related to
* #5142

(but does not fix that issue)

**Description**
Replaces `Instance::any_backend_feature_enabled` with `Instance::enabled_backend_features` which reports all available backends instead of just reporting if none is available.

We already chugged through the very same information but didn't expose it. This is useful when a library or application wants to reason about potentially available backends based on its build configuration.

**Testing**
Run an example, didn't explode.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -1826,13 +1826,13 @@ impl Instance {
     /// # Panics
     ///
     /// If no backend feature for the active target platform is enabled,
-    /// this method will panic, see [`Instance::any_backend_feature_enabled()`].
+    /// this method will panic, see [`Instance::enabled_backend_features()`].
     #[allow(unreachable_code)]
     pub fn new(_instance_desc: InstanceDescriptor) -> Self {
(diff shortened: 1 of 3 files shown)


nical (approved):

wgpu/src/lib.rs
```diff
@@ -1748,39 +1748,63 @@ impl Default for Instance {
...
+            if cfg!(metal) {
+                backends = backends.union(Backends::METAL);
+            }
+            if cfg!(dx12) {
+                backends = backends.union(Backends::DX12);
+            }
+
+            // Windows & Linux currently always enable Vulkan and OpenGL.
+            // See <https://github.com/gfx-rs/wgpu/issues/3514>
+            if cfg!(target_os = "windows") || cfg!(target_os = "linux") {
```
Wumpf: last minute noticed that this doesn't check for android, or rather for android we don't add any backend

Issue #4455, opened 2021-12-21
[glsl-out] fix usage of gsamplerCubeArrayShadow


kvark:
from https://github.com/bevyengine/bevy/issues/841#issuecomment-999017895
> Internal error in FRAGMENT shader: gsamplerCubeArrayShadow isn't supported in textureGrad, textureLod or texture with bias

allsey87:
A quick note for those who are trying to get Bevy working under Linux with an Intel-based graphics card, you probably need to install the `vulkan-intel` package for your distribution.

magcius:
I realize immediately below the code that errors out is code that checks for TEXTURE_SHADOW_LOD. Assuming the user has support for TEXTURE_SHADOW_LOD, this shouldn't be impossible to fix.

teoxoy:
https://github.com/gfx-rs/wgpu/pull/5171 should have resolved this, closing.

Issue #4405, opened 2022-12-03
[wgsl-in] Remove some texture types


teoxoy:
- `texture_1d_array`/`texture_storage_1d_array` (https://github.com/gpuweb/gpuweb/pull/1452)
- `texture_multisampled_2d_array` (https://github.com/gpuweb/gpuweb/pull/1741)
- [ ] File a spec issue to get these added back, because WebGPU could support these.

PR #5171, opened 2024-01-31
Fix glsl backend errors regarding samplerCubeArrayShadow 

cmrschwarz:
**Connections**
Offers a solution for https://github.com/gfx-rs/wgpu/issues/4455, and the related downstream issues ( see https://github.com/bevyengine/bevy/issues/7320 )

**Description**
GL's [`textureLod`](https://registry.khronos.org/OpenGL-Refpages/gl4/html/textureLod.xhtml) does not support sampling from a cube depth texture with an explicit LOD.

This is made significantly worse by the fact that the only way to use wgsl's [`textureSampleCompare`](https://gpuweb.github.io/gpuweb/wgsl/#texturesamplecompare) from 
[non uniform control flow](https://gpuweb.github.io/gpuweb/wgsl/#uniform-control-flow) is to use [`textureSampleCompareLevel`](https://gpuweb.github.io/gpuweb/wgsl/#texturesamplecomparelevel), which forces us to specify an LOD (0).

This currently triggers a [backend error](https://github.com/gfx-rs/wgpu/blob/e75fb3c224aed976cfc730dae1bbf1822f9eb65b/naga/src/back/glsl/mod.rs#L2637) that has been observed downstream multiple times,
e.g. breaking bevy's PBR pipeline for the GL backend, and therefore many Intel Graphics Cards.  

This PR mostly solves this problem by changing the behavior for this case from a backend error to a
GL feature requirement on [GL_EXT_texture_shadow_lod](https://registry.khronos.org/OpenGL/extensions/EXT/EXT_texture_shadow_lod.txt) , as was previously suggested by @magcius .
Fortunately, this extension seems to be available on many of the affected devices.

 The existing functionality around [`workaround_lod_array_shadow_as_grad`](https://github.com/gfx-rs/wgpu/blob/e75fb3c224aed976cfc730dae1bbf1822f9eb65b/naga/src/back/glsl/mod.rs#L2648) is preserved, and preferred over requiring the extension where possible.

Some impossible cases still remain ( [`textureGrad`](https://registry.khronos.org/OpenGL-Refpages/gl4/html/textureGrad.xhtml) still doesn't support `samplerCubeArrayShadow`), but those should be far less common.

**Testing**
A testcase showing the extension working was added in b2ea9fe20e76548b36afa1c8e3f3c16312b5015b.
A testcase around overriding the workaround if instructed was added in cd4108f29660edb0955834918128a82ea47bc166.

**Checklist**

- [x] Run `cargo fmt`.


cwfitzgerald (dismissed):
This is all very cool!

I feel like this requires a `wgpu` downlevel flag, so users can know if `GL_EXT_texture_shadow_lod` is allowed and avoid a runtime crash.

cmrschwarz:
Sure, I could do that.

But  I don't fully understand yet why we would need / want that.
Where do you see the difference between this GL extension and others that we are already using (like `GL_EXT_texture_cube_map_array`) that would warrant a feature flag?

We don't have any alternatives to using the extension. 
The logic in `FeaturesManager` makes sure that we only require it if absolutely necessary, just like with any other extensions we use. 
So we could only throw a backend error again if a potential 'allow_gl_ext_texture_shadow_lod' feature flag was not set.  

If we want users to be able to get better error messages (custom ones instead of what their driver ends up reporting for missing extensions), wouldn't that apply to all our used GL extensions, and be out of scope for this PR?

magcius:
The downlevel flag would still be a good idea since there's one case we can't emulate even with EXT_texture_shadow_lod: samplerCubeArrayShadow on textureGrad.

cmrschwarz:
I kept the backend error that we had before this PR [for that case](https://github.com/gfx-rs/wgpu/pull/5171/commits/97930557a9b6ed33b7cbc70da47c3965cad76865#diff-4caf37666347d5ddf75507038b3cf9addea5f396df2b1657c6346bdec65a6686R2623) @magcius .

What is the behavior you would like to see instead? (with / without the downlevel flag set)

naga/src/back/glsl/mod.rs
```diff
@@ -2649,10 +2631,7 @@ impl<'a, W: Write> Writer<'a, W> {
                     || dim == crate::ImageDimension::Cube)
```
teoxoy: We need to tighten the image type check here or it will match things we didn't expect (ex `samplerCubeArrayShadow`).

Also, (this is an existing issue but) this workaround doesn't actually work for `crate::SampleLevel::Exact`, only for `crate::SampleLevel::Zero`.

https://github.com/gfx-rs/wgpu/blob/84ba4e5461fd1f305f8591cb086c40a488fbdb71/naga/src/back/glsl/mod.rs#L2743

It would be better to error instead of continuing.

naga/src/back/glsl/features.rs
```diff
@@ -469,6 +477,48 @@ impl<'a, W> Writer<'a, W> {
...
+                            || (cube && arrayed)
+                        );
+
+                        let mut is_used = false;
+
+                        // float texture(sampler2DArrayShadow sampler, vec4 P [, float bias])
+                        is_used |= array2d && bias && !use_grad_workaround;
+
+                        // float texture(samplerCubeArrayShadow sampler, vec4 P, float compare [, float bias])
+                        is_used |= cube && bias;
```
teoxoy:
```suggestion
                        is_used |= (cube && arrayed) && bias;
```

naga/src/back/glsl/features.rs
```diff
@@ -469,6 +477,48 @@ impl<'a, W> Writer<'a, W> {
...
+                        let array2d = dim == ImageDimension::D2 && arrayed;
+
+                        // We have an workaround for cases other than samplerCubeArrayShadow,
+                        // so we don't *need* this extension for those cases.
+                        // But if we we're explicitly allowed to use it (`WriterFlags::TEXTURE_SHADOW_LOD`),
+                        // we prefer it over the workaround.
+                        let use_grad_workaround = !(
+                            self.options.writer_flags.contains(WriterFlags::TEXTURE_SHADOW_LOD)
+                            || (cube && arrayed)
+                        );
```
teoxoy: I find this a bit hard to wrap my head around when combined with the other operations below.

It would be easier to follow if we do something like:
```rust
let possibly_needs_grad_workaround = (sampler2DArrayShadow || samplerCubeShadow) && crate::SampleLevel::Zero;
is_used |= possibly_needs_grad_workaround && self.options.writer_flags.contains(WriterFlags::TEXTURE_SHADOW_LOD);
```

teoxoy (changes requested):

cmrschwarz:
Thank you for taking a look at this.
I'm glad to see that there's interest in merging this.

You're definitely right that the criteria for using the extension / workaround weren't quite perfect yet.
I updated the code to hopefully fix this, and replaced all the problematic cases with errors, as you suggested. 
I would appreciate somebody taking the time to review this again.

There's a conversation to be had here about the exact GL / GLES version requirements of the sampling functions, which currently aren't enforced properly. This is especially problematic since we currently default to GLES 3.1, which lacks some of the features we're using here. While that's potentially out of scope for this PR, I would be willing to work on that.
In the process of fixing the criteria for the extension I already ended up creating the following table by scouring through the documentations.
I'm reproducing it here hoping it may be useful:

<table>
    <tr>
        <td></td>
        <td colspan=2>2DArray</td>
        <td colspan=2>cube</td>
        <td colspan=2>cubeArray</td>
        <td colspan=2>2DArrayShadow</td>
        <td colspan=2>cubeShadow</td>
        <td colspan=2>cubeArrayShadow</td>
        <td>workaround for lod 0</td>
    </tr>
    <tr>
        <td></td>
        <td>GL</td>
        <td>GLES</td>
        <td>GL</td>
        <td>GLES</td>
        <td>GL</td>
        <td>GLES</td>
        <td>GL</td>
        <td>GLES</td>
        <td>GL</td>
        <td>GLES</td>
        <td>GL</td>
        <td>GLES</td>
        <td></td>
    </tr>
    <tr>
        <td>texture</td>
        <td>1.3</td>
        <td>3.0</td>
        <td>1.3</td>
        <td>3.0</td>
        <td>4.0</td>
        <td>3.2</td>
        <td>1.3</td>
        <td>3.0</td>
        <td>1.3</td>
        <td>3.0</td>
        <td>4.0</td>
        <td>3.2</td>
        <td></td>
    </tr>
    <tr>
        <td>texture+bias</td>
        <td>1.3</td>
        <td>3.0</td>
        <td>1.3</td>
        <td>3.0</td>
        <td>4.0</td>
        <td>3.2</td>
        <td>EXT</td>
        <td>EXT</td>
        <td>EXT</td>
        <td>EXT</td>
        <td>EXT</td>
        <td>EXT</td>
        <td></td>
    </tr>
    <tr>
        <td>textureLod</td>
        <td>1.3</td>
        <td>3.0</td>
        <td>1.3</td>
        <td>3.0</td>
        <td>4.0</td>
        <td>3.2</td>
        <td>EXT</td>
        <td>EXT</td>
        <td>EXT</td>
        <td>EXT</td>
        <td>EXT</td>
        <td>EXT</td>
        <td>textureGrad</td>
    </tr>
    <tr>
        <td>textureGrad</td>
        <td>1.3</td>
        <td>3.0</td>
        <td>1.3</td>
        <td>3.0</td>
        <td>4.0</td>
        <td>3.2</td>
        <td>1.3</td>
        <td>3.0</td>
        <td>1.3</td>
        <td>3.0</td>
        <td>no</td>
        <td>no</td>
        <td></td>
    </tr>
    <tr>
        <td>textureGather</td>
        <td>4.0</td>
        <td>3.1</td>
        <td>4.0</td>
        <td>3.1</td>
        <td>4.0</td>
        <td>3.2</td>
        <td>4.0</td>
        <td>3.1</td>
        <td>4.0</td>
        <td>3.1</td>
        <td>4.0</td>
        <td>3.2</td>
        <td></td>
    </tr>
    <tr>
        <td>textureOffset</td>
        <td>1.3</td>
        <td>3.0</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td>4.3</td>
        <td>EXT</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td></td>
    </tr>
    <tr>
        <td>textureOffset+bias</td>
        <td>1.3</td>
        <td>3.0</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td>EXT</td>
        <td>EXT</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td></td>
    </tr>
    <tr>
        <td>textureLodOffset</td>
        <td>1.3</td>
        <td>3.0</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td>EXT</td>
        <td>EXT</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td>textureGradOffset</td>
    </tr>
    <tr>
        <td>textureGradOffset</td>
        <td>1.3 </td>
        <td>3.0</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td>1.3</td>
        <td>3.0</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td></td>
    </tr>
    <tr>
        <td>textureGatherOffset</td>
        <td>4.0</td>
        <td>3.1</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td>4.0</td>
        <td>3.0</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td>no</td>
        <td></td>
    </tr>
</table>

naga/src/back/glsl/features.rs
cmrschwarz: I hated this code too, clippy forced me to invert the conditon...
Should be more readable now.

naga/src/back/glsl/mod.rs
```diff
@@ -2620,51 +2610,49 @@ impl<'a, W: Write> Writer<'a, W> {
...
-                                 textureLod or texture with bias",
-                            )))
-                        }
-                        crate::SampleLevel::Auto => {}
+                let mut err = None;
+                if dim == crate::ImageDimension::Cube {
+                    if offset.is_some() {
+                        err = Some("gsamplerCube[Array][Shadow] doesn't support texture sampling with offsets");
+                    }
+                    if matches!(level, crate::SampleLevel::Gradient { .. }) {
```
teoxoy:
```suggestion
                    if arrayed && matches!(class, crate::ImageClass::Depth { .. }) && matches!(level, crate::SampleLevel::Gradient { .. }) {
```

naga/src/back/glsl/mod.rs
```diff
@@ -2620,51 +2610,49 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                        err = Some("samplerCubeArrayShadow don't support textureGrad");
                     }
                 }
+                if gather.is_some() && level != crate::SampleLevel::Zero {
+                    err = Some("textureGather doesn't support LOD parameters");
+                }
+                if dim == crate::ImageDimension::D3 && arrayed {
+                    err = Some("3d array textures don't support sampling");
+                }
```
teoxoy: Could you move this check in the validator?

https://github.com/gfx-rs/wgpu/blob/563d9202c33ae97b1dd6f1f6c54c7cd1f57158dc/naga/src/valid/type.rs#L596

The `arrayed` param can only be `true` for `ImageDimension::D2` or `ImageDimension::Cube`.

naga/src/back/glsl/mod.rs
```diff
@@ -2620,51 +2610,49 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                if let Some(err) = err {
+                    return Err(Error::Custom(String::from(err)));
+                }
+
+                // `textureLod[Offset]` on `sampler2DArrayShadow` and `samplerCube[Array]Shadow` does not exist in GLSL,
+                // unless `GL_EXT_texture_shadow_lod` is present.
+                // But if the target LOD is zero, we can emulate that by using `textureGrad[Offset]` with a constant gradient of 0.
+                let workaround_lod_with_grad = (dim == crate::ImageDimension::Cube
```
teoxoy:
```suggestion
                let workaround_lod_with_grad = ((dim == crate::ImageDimension::Cube && !arrayed)
```

`textureGrad` doesn't support `cubeArrayShadow`; it shouldn't matter however due to how the logic is implemented but makes things consistent with `grad_workaround_applicable`.

naga/src/back/glsl/features.rs
```diff
@@ -469,6 +477,46 @@ impl<'a, W> Writer<'a, W> {
...
+
+                        let mut ext_used = false;
+
+                        // float texture(sampler2DArrayShadow sampler, vec4 P [, float bias])
+                        // float texture(samplerCubeArrayShadow sampler, vec4 P, float compare [, float bias])
+                        ext_used |= (array2d || cube && arrayed) && bias;
+
+                        // The non `bias` version of this was standardized in GL 4.3, but never in GLES.
+                        // float textureOffset(sampler2DArrayShadow sampler, vec4 P, ivec2 offset [, float bias])
+                        ext_used |= array2d && (bias || gles) && offset.is_some();
```
teoxoy:
```suggestion
                        ext_used |= array2d && (bias || (auto && gles)) && offset.is_some();
```

We need to check for `SampleLevel::Auto` or else we might also match grad and gather.

teoxoy:
Thanks for sharing the table!

The only discrepancy I found was that `texture`, `texture+bias` and `textureGrad` with `cube[Shadow]` are available since GLES 1.3.

> There's a conversation to be had here about the exact GL / GLES version requirements of the sampling functions, which currently aren't enforced properly.

The only restrictions from our base of GLSL 1.3 & GLSL ES 3.0 seem to be:

- `textureGather[Offset]` available since 4.0 / 3.1
- `cubeArray[Shadow]` available since 4.0 / 3.2

We seem to check for availability of `textureGather[Offset]` and we should.
For cube arrays, we have `Features::CUBE_TEXTURES_ARRAY` which will request the extension. An improvement would be to omit requesting the extension if we are on 3.2 already.

teoxoy (changes requested):

teoxoy:
glslangValidator v13.1.0 (released on 2023-10-13) added support for `GL_EXT_texture_shadow_lod` but CI seems to be on v11.8.0 (released on 2022-01-27).

@cwfitzgerald do you know how we could get v13.1.0 on CI?

cwfitzgerald:
Installing the vulkan sdk is a good way of getting an up to date copy of the glsl tooling

teoxoy:
https://github.com/gfx-rs/wgpu/pull/5202 landed; @cmrschwarz could you rebase on top of it to see if the CI failure goes away?

naga/src/back/glsl/mod.rs
cmrschwarz: We rely on this working for `ImageDimension::D1` aswell, see for example [here](https://github.com/gfx-rs/wgpu/blob/b8a8ff6e44162a41b8d54184a88613794dd3d2e4/naga/tests/out/wgsl/images.frag.wgsl#L8).
GLSL officially supports that, but you seem to have a point about WebGPU [not supporting it?](https://gpuweb.github.io/gpuweb/wgsl/#sampled-texture-type). What to do?

naga/src/back/glsl/mod.rs
teoxoy: It looks like this didn't change.

naga/src/back/glsl/mod.rs
```diff
@@ -2620,51 +2610,46 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                }
+
+                // `textureLod[Offset]` on `sampler2DArrayShadow` and `samplerCubeShadow` does not exist in GLSL,
+                // unless `GL_EXT_texture_shadow_lod` is present.
+                // But if the target LOD is zero, we can emulate that by using `textureGrad[Offset]` with a constant gradient of 0.
+                let workaround_lod_with_grad = (dim == crate::ImageDimension::Cube && !arrayed)
+                    || (dim == crate::ImageDimension::D2 && arrayed)
+                        && level == crate::SampleLevel::Zero
+                        && matches!(class, crate::ImageClass::Depth { .. })
+                        && !self.features.contains(Features::TEXTURE_SHADOW_LOD);
```
teoxoy:
```suggestion
                let workaround_lod_with_grad = ((dim == crate::ImageDimension::Cube && !arrayed)
                    || (dim == crate::ImageDimension::D2 && arrayed))
                    && level == crate::SampleLevel::Zero
                    && matches!(class, crate::ImageClass::Depth { .. })
                    && !self.features.contains(Features::TEXTURE_SHADOW_LOD);
```

I think we need another set of parens here.

naga/src/back/glsl/mod.rs
teoxoy: The validator usually implements the WGSL restrictions and use [`Capabilities`](https://github.com/gfx-rs/wgpu/blob/07e59eb6fc7de3f682f1c401b9cf9f0da9ee4b4a/naga/src/valid/mod.rs#L80) to allow extra things through.

In this case, I'm not sure how prevalent usage of 1d storage arrays is. I think wgpu also won't allow you to create 1d-array texture views. It might be best to validate out 1d arrays (https://github.com/gfx-rs/wgpu/issues/4405) and if people would like to use them in the future we implement proper support for them.

Since https://github.com/gfx-rs/wgpu/issues/4405 is already tracking this, it can be a different PR.

cmrschwarz:
> We seem to check for availability of textureGather[Offset] and we should.

 `textureGather[Offset]`  before GL 4.0 / GLES 3.1 currently compiles without any errors / extensions used.

Demonstration test:

wgsl:
```wgsl
@group(0) @binding(0) var texture: texture_cube<f32>;
@group(0) @binding(1) var texture_sampler: sampler;

@fragment
fn main() -> @location(0) vec4<f32>  {
    return textureGather(0, texture, texture_sampler, vec3<f32>(0.0));
}
```

ron:
```ron
(
	glsl: (
        writer_flags: (""),
        version: Desktop(140),
		binding_map: {},
		zero_initialize_workgroup_memory: true,
	),
)
```

output:
```
#version 140 core
uniform samplerCube _group_0_binding_0_fs;

out vec4 _fs2p_location0;

void main() {
    vec4 _e4 = textureGather(_group_0_binding_0_fs, vec3(vec3(0.0)), vec3(0.0), vec3(0.0), 0);
    _fs2p_location0 = _e4;
    return;
}

```
---

> An improvement would be to omit requesting the extension if we are on 3.2 already.

Things that currently require an extension for but don't have to:
-  any use of `cubeArray` / `cubeArrayShadow` in GL 4.0 / GLES 3.2
- `textureOffset` on `2DArrayShadow` in GL 4.3

--- 

> The only restrictions from our base of GLSL 1.3 & GLSL ES 3.0 seem to be [...]

Maybe not too really relevant, but we currently [don't seem to support GL 1.3](https://github.com/gfx-rs/wgpu/blob/b8a8ff6e44162a41b8d54184a88613794dd3d2e4/naga/src/back/glsl/mod.rs#L67), only GL 1.4. I assume that's intended.

teoxoy:
> > We seem to check for availability of `textureGather[Offset]` and we should.
> 
> `textureGather[Offset]`  before GL 4.0 / GLES 3.1 currently compiles without any errors / extensions used.

Right, I mistyped, should have been: "We **don't** seem to check for availability of `textureGather[Offset]` and we should.".

cmrschwarz:
Damn, sorry for all my mistakes on this.
35dbd9d2a525e3b96e8658396fd5009d77f8e544 should address the issues you pointed out.
Thank you for your patience with me @teoxoy .

Regarding the`textureGather` situation, would you be fine with a separate PR regarding that?

naga/src/back/glsl/mod.rs
```diff
@@ -2620,51 +2610,46 @@ impl<'a, W: Write> Writer<'a, W> {
...
-                            )))
-                        }
-                        crate::SampleLevel::Auto => {}
+                let mut err = None;
+                if dim == crate::ImageDimension::Cube {
+                    if offset.is_some() {
+                        err = Some("gsamplerCube[Array][Shadow] doesn't support texture sampling with offsets");
+                    }
+                    if arrayed && matches!(level, crate::SampleLevel::Gradient { .. }) {
+                        err = Some("samplerCubeArrayShadow don't support textureGrad");
```
teoxoy:
```suggestion
                    if arrayed && matches!(class, crate::ImageClass::Depth { .. }) && matches!(level, crate::SampleLevel::Gradient { .. }) {
                        err = Some("samplerCubeArrayShadow don't support textureGrad");
```

The depth check is still missing 😅

teoxoy:
No worries :)

Thanks for addressing all the feedback!

> Regarding the `textureGather` situation, would you be fine with a separate PR regarding that?

Yeah, that sounds the best.

naga/src/back/glsl/mod.rs
cmrschwarz: Oh god. Fixed 😅

teoxoy (approved):
Good stuff!

teoxoy:
Ah, I forgot about the downlevel flag request. Would you be able to add a downlevel flag for the `GL_EXT_texture_shadow_lod` extension? We could do that in a follow-up PR as well.

The role of downlevel flags is letting users know what functionality is available before they use it. The presence of all dowlevel flags also signifies that the device implements the WebGPU API fully.

cmrschwarz:
Oh! When I initially submitted this, I wasn't too familiar with the library, so I was thinking Rust feature flag, not `wgpu::DownlevelFlag`. 

My initial comments regarding this probably didn't make any sense at all 😅 . Sorry for my ignorance.

Since this comment thread is getting pretty long, and as the downlevel flag sort of addresses an issue that was present before this PR, I would prefer doing a follow up PR, if thats ok for you guys.

billyb2:
hey, would it be possible to cut a new release? Using this PR would be super useful for a personal project of mine. If not, no worries. Hope I'm not bugging yall :)

cwfitzgerald:
Just landed in 0.19.2 :)

billyb2:
> Just landed in 0.19.2 :)

You guys are awesome

PR #5181, opened 2024-02-01
Remove another unused A: HalApi parameter

udoprog:
This removes two more instances that follows from the identifier refactoring (#5108) where we're not actually using the `HalApi` parameter. This changes nothing in terms of codegen, but there's little to no point in keeping unused type complexity around.


nical (approved):
Neat!

PR #5182, opened 2024-02-01
Always serialize and deserialize ids the same way

nical:
**Description**

`RawId`'s serialization/deserialization currently depends on what cargo features are enabled. If `trace` is enabled it is serialized via the `SerialId` type and if `replay` is enabled it is deserialized from `SerialId`.

This means that if an application serializes and deserializes Ids (like gecko does for some of the remoting), then having only one of `trace` and `replay` features enabled is broken. Firefox only uses the `trace` feature.

For example:

```rust
let id = wgc::id::AdapterId::zip(0, 1, wgt::Backend::Vulkan);
let mut tmp = Vec::new();
bincode::serialize_into(&mut tmp, &id).unwrap();
let de = bincode::deserialize::<wgc::id::AdapterId>(tmp.as_slice()).unwrap(); // Boom.
```

It's unclear why it used to work and only started breaking at the latest update.

With this patch PR, ids are always serialized via `SerialId`. It's probably not what we want long term, but we need the fix rather urgently so it will have to do for now.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/id.rs
+++ b/wgpu-core/src/id.rs
@@ -19,10 +19,16 @@ pub const EPOCH_MASK: u32 = (1 << (EPOCH_BITS)) - 1;
 
 /// The raw underlying representation of an identifier.
 #[repr(transparent)]
-#[cfg_attr(any(feature = "serde", feature = "trace"), derive(serde::Serialize))]
-#[cfg_attr(any(feature = "serde", feature = "replay"), derive(serde::Deserialize))]
-#[cfg_attr(feature = "trace", serde(into = "SerialId"))]
-#[cfg_attr(feature = "replay", serde(from = "SerialId"))]
+#[cfg_attr(
+    any(feature = "serde", feature = "trace"),
+    derive(serde::Serialize),
+    serde(into = "SerialId")
+)]
+#[cfg_attr(
+    any(feature = "serde", feature = "replay"),
+    derive(serde::Deserialize),
+    serde(from = "SerialId")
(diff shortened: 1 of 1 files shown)


teoxoy:
> It's unclear why it used to work and only started breaking at the latest update.

https://github.com/gfx-rs/wgpu/pull/5149 changed some of the features but I don't see anything off that could have caused this.

It's worth noting that the `serial-pass` feature was replaced by `serde` but if either `replay` or `trace` are enabled it should also get enabled.

> Firefox only uses the `trace` feature.

I [see](https://searchfox.org/mozilla-central/rev/76cb3efe3b19e649bf675bb6ec5d4af8109b9771/gfx/wgpu_bindings/Cargo.toml#23) we also use the `replay` feature.

This change looks fine to me either way.

teoxoy (approved):

nical:
> I [see](https://searchfox.org/mozilla-central/rev/76cb3efe3b19e649bf675bb6ec5d4af8109b9771/gfx/wgpu_bindings/Cargo.toml#23) we also use the `replay` feature.
> 
> This change looks fine to me either way.

Ah, that explains it then. I'm continuing Jim's work-in-progress wgpu update, I'm on top of https://phabricator.services.mozilla.com/D200266 where `trace` is enabled but not `replay`.

nical:
Let's get rid of the footgun anyway.

Issue #5180, opened 2024-02-01
wgpu version 0.19 performance regression


nathanielsimard:
We recently released a new version of Burn that includes the latest version of wgpu, and we observed a significant performance regression. The original issue is documented here: https://github.com/tracel-ai/burn/issues/1213, and it appears that reverting to wgpu version 0.18.0 resolves the problem (refer to the `fix/slow-wgpu` branch).

It's possible that this regression is related to the new and improved multithreading in wgpu. However, our framework follows a client-server architecture, allowing us to configure the communication channel between them. Whether using a mutex (involving multiple threads) or a channel with communication limited to a single thread, we observed no change in performance behavior.

For short-lived programs, the impact on execution is minimal. It's worth noting that we manually select the adapter, and I did all tests on a discrete GPU with the Vulkan backend on Pop OS.

Elabajaba:
Do you know why the wgpu 0.19 version is taking significantly more memory than the previous version in the screenshots? It looks like it might be hitting a memory leak?

nathanielsimard:
I don't believe memory is the issue. When a memory leak occurs, it only takes a few seconds to fill the GPU memory and freeze the entire system 😅. My intuition suggests that the recent changes with the read-write locks are causing excessive overhead. While this might not be significant for graphics since you typically won't render at more than 240 FPS, for compute tasks, it could be comparable to rendering at 20,000 FPS or even more.

If this is indeed the case (uncertain how we can easily test it), then we could swap those read-write locks with `RefCell` using a feature flag. This way, we can avoid altering the architecture while enabling single-threaded execution with minimal overhead.

hakolao:
I have the same in my project. FPS slowly just keeps dropping and dropping. I have not spent too much time figuring out why, because I didn't have time to spend for debugging, but the drop is significant and makes my game unusable.

I use compute shaders a lot.

The difference is: 0.18: Very playable vs 0.19 completely unplayable.

cryscan:
@hakolao It seems that in v0.19, the program is waiting on CPUs a lot in `wgpu::ComputePipeline::get_bind_group_layout`. See #5196

cwfitzgerald:
I think this an #5196 are the same, going to close in favor of that.

PR #5183, opened 2024-02-01
Suggestion to introduce wgpu-sync for lock profiling

udoprog:
**Connections**
Initially I started working on this to introduce loom tests, but I figured building a face like this would be useful to troubleshoot potential issues and bottlenecks caused by locks.

Could be used to troubleshoot issues which ask whether things have become slower with arcanization and increased use of locks in wgpu, such as #5180.

**Description**
This introduces a compatible layer which conditionally replaces Mutexes and RwLocks with facades which efficiently captures events into a global collection. This is compile-time enabled with the `"trace"` feature. Using this, users of `wgpu` can capture a lock trace, and generate a html report like this:

```rust
    if condition {
        wgpu::sync::capture();
    }

    /* do some wgpu work */

    if condition {
        let events = wgpu::sync::drain();

        let f = std::fs::File::create("trace.html").context("Creating trace.html")?;
        wgpu::sync::html::write(f, &events).context("Writing trace.html")?;
        println!("Wrote trace.html");
    }
```

I've included an example report I've generated from one of my projects here:
https://udoprog.github.io/trace.html

Note that if the `"trace"` feature is disabled, wgpu-sync instead provides the original parking_lot implementations without using facades.

This is a draft for now. Please take a look and tell me if y'all are interested in adopting something like this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ members = [
     "wgpu-info",
     "wgpu-macros",
     "wgpu-types",
+    "wgpu-sync",
     "wgpu",
 ]
 exclude = []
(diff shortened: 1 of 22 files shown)


udoprog:
Closing for now to calm the builds down. I'll keep my branch around for profiling and might suggest it for inclusion later if it pans out.

PR #5185, opened 2024-02-02
using offset + size instead of range

andristarr:
**Connections**
https://github.com/gfx-rs/wgpu/issues/4215

**Description**
buffer_map_async uses offset + size instead of range

**Testing**

**Checklist**

- [ ] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -26,9 +26,7 @@ use wgt::{BufferAddress, TextureFormat};
 
 use std::{
     borrow::Cow,
-    iter,
-    ops::Range,
-    ptr,
+    iter, ptr,
     sync::{atomic::Ordering, Arc},
 };
 
@@ -2336,15 +2334,18 @@ impl Global {
     pub fn buffer_map_async<A: HalApi>(
         &self,
         buffer_id: id::BufferId,
-        range: Range<BufferAddress>,
+        offset: BufferAddress,
+        size: Option<BufferAddress>,
         op: BufferMapOperation,
     ) -> BufferAccessResult {
-        api_log!("Buffer::map_async {buffer_id:?} range {range:?} op: {op:?}");
+        api_log!("Buffer::map_async {buffer_id:?} offset {offset:?} size {size:?} op: {op:?}");
 
         // User callbacks must not be called while holding buffer_map_async_inner's locks, so we
         // defer the error callback if it needs to be called immediately (typically when running
         // into errors).
-        if let Err((mut operation, err)) = self.buffer_map_async_inner::<A>(buffer_id, range, op) {
+        if let Err((mut operation, err)) =
+            self.buffer_map_async_inner::<A>(buffer_id, offset, size, op)
+        {
             if let Some(callback) = operation.callback.take() {
                 callback.call(Err(err.clone()));
             }
@@ -2360,7 +2361,8 @@ impl Global {
     fn buffer_map_async_inner<A: HalApi>(
         &self,
         buffer_id: id::BufferId,
-        range: Range<BufferAddress>,
+        offset: BufferAddress,
+        size: Option<BufferAddress>,
         op: BufferMapOperation,
     ) -> Result<(), (BufferMapOperation, BufferAccessError)> {
         profiling::scope!("Buffer::map_async");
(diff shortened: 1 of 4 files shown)


nical:
Sorry I missed your question in the other PR:

> Sorry, I am not exactly clear on this actually @nical .
> 
> Do we want to keep the scope of change only to the wgpu_core wrapper in the backend? If yes, buffer_get_mapped_range already consumes a range, unless I am missing something.

I'd like `wgpu-core` to stay close to the specification (offset + optional size), while `wgpu` can have a more rusty interface.

nical:
Oh I had not realized your goal wasn't to change `wgpu`'s API which means I'm asking something contrary to what Connor said earlier. Lemme debate it with him and I'll get back to you.

andristarr:
Let me know what was the result of the talk please!

nical:
Sorry! I thought I did but apparently failed to hit the green button. So the conclusion is to do the opposite: `wgpu-core` should use offset + optional size for both functions, while `wgpu` takes ranges. Sorry again for the confusion and the late reply.

andristarr:
Just to be clear: in global.rs in wgpu-core we want it to the offset + size combo, while in wgpu we want it to have range. If I am not mistaken we already have this. 
wgpu_core.rs wrapper in wgpu `buffer_get_mapped_range` already uses range
global.rs in wgpu-core uses offset + size

nical:
Yes, it's just `buffer_map_async` in wgpu-core that should be offset + optional size but is currently a range.

andristarr:
Sorry for the many questions, I think I understood you now.

I am not exactly sure we want the check I added in `buffer_map_async`

Let me know if you are happy with this solution

wgpu-core/src/device/global.rs
```diff
@@ -2336,15 +2336,47 @@ impl Global {
...
         buffer_id: id::BufferId,
-        range: Range<BufferAddress>,
+        offset: BufferAddress,
+        size: Option<BufferAddress>,
         op: BufferMapOperation,
     ) -> BufferAccessResult {
-        api_log!("Buffer::map_async {buffer_id:?} range {range:?} op: {op:?}");
+        api_log!("Buffer::map_async {buffer_id:?} offset {offset:?} size {size:?} op: {op:?}");
+
+        let hub = A::hub(self);
```
nical: Let's put the code below in `buffer_map_async_inner`, to put the validation in once place and so that we don't have to get the buffer from the registry twice.

nical (changes requested):
Thanks! Just a small change to make.

wgpu-core/src/device/global.rs
andristarr: I think we need to change the signature of `buffer_map_async_inner` as well then to pass in the buffer. Otherwise the we cannot get the actual `range_size`

Would that be fine? That way the check can happen in `buffer_map_async` and the extracted buffer will be passed so the wrapped call doesn't have to get it again.

Or we change the signature of the inner call to also have offset + option(size) instead of range.

nical: Let's make `buffer_map_async_inner` take an offset and an optional size.

andristarr: Done! Sorry for the long thread for such a small change.

I am still picking up Rust, so appreciate any comments!

wgpu-core/src/device/global.rs
```diff
@@ -2372,22 +2374,43 @@ impl Global {
...
+                Err(_) => {
+                    return Err((op, BufferAccessError::Invalid));
                 }
             };
+            {
+                let snatch_guard = buffer.device.snatchable_lock.read();
+                if buffer.is_destroyed(&snatch_guard) {
+                    return Err((op, BufferAccessError::Destroyed));
+                }
+            }
```
nical, line 2391: This check is already happening 50-ish lines below, no need to do it twice. It's a good idea to check this early so you can remove the other one,

nical (changes requested):
Almost there

wgpu-core/src/device/global.rs
nical: No worries, I was the one who came with conflicting requirements in the first place.

wgpu-core/src/device/global.rs
andristarr: Done!

nical (approved):

PR #5198, opened 2024-02-05
Improve backends doc

waywardmonkeys:
**Description**
This improves the docs for `Backends` and mentions that DirectX12 support is for Windows versions 10 and later rather than only 10.

**Testing**
Built docs and looked at them.

**Checklist**

- [x] Run `cargo fmt`.

I didn't think this was important enough for a changelog entry. If I'm wrong, let me know and I'll add one.


teoxoy (approved):
Thanks!

PR #5199, opened 2024-02-05
Improve Instance::available_backend_features() on macOS.

waywardmonkeys:
The check for macOS and Angle should add `Backends::GL` not `Backends::VULKAN`. Similarly, the comment gets updated to not refer to Vulkan.

This code was introduced in #5167.

**Checklist**

- [x] Run `cargo fmt`.

The change that this is fixing is already covered in the changeling and hasn't been released yet.

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -1791,9 +1791,9 @@ impl Instance {
                 backends = backends.union(Backends::VULKAN);
             }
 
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
Should this be true for iOS as well and not just macOS?

cwfitzgerald (approved):
Thanks!

No, iirc you can't ship angle on ios

PR #1892, opened 2021-09-01
More validation of the surface texture lifetime

kvark:
**Connections**
Closes #1878

**Description**
Adds `has_work` flag to surface textures. This is temporary, until #1688 takes a better care of it. cc @Wumpf 
Adds a few error conditions in addition to the "no work done":
  - if we finished recording a pass, but the surface texture is gone
  - if we are presenting a wrong frame to a wrong surface

**Testing**
Manually tested

PR #5200, opened 2024-02-05
Remove TextureInner::Surface::has_work.

### LINKED ISSUE #3189: "No work has been submitted for this frame" timeout
**Description**
If you do not submit any work for a surface, and present it, problems occur running wgpuSwapChainGetCurrentTextureView (wgpu-native), it seems to be hitting [this line](https://github.com/gfx-rs/wgpu/blob/master/wgpu-core/src/present.rs#L314) specifically

**Repro steps**
You can just strip out all code that submits work from any rendering example

Nothing should happen, and wgpuSwapChainGetCurrentTextureView should run just fine, but it does not, sometimes failing with a timeout, sometimes working 

**Extra materials**
the 'beginning scene' is where i get the swap chain's current texture view, things marked with 'CallbackError' are errors from wgpuDeviceSetUncapturedErrorCallback
image

**Platform**
Arch Linux, AMD GPU, latest master of wgpu-native

waywardmonkeys:
When no work is submitted for a frame, presenting the surface results in a timeout due to no work having been submitted.

Fixes #3189.

This flag was added in #1892 with a note that it was going to be temporary until #1688 landed.

**Testing**
Tested by commenting out the rendering code of the cube example and running it with and without this patch.

**Checklist**

- [x] Run `cargo fmt`.


nical (approved):

waywardmonkeys:
Thank you for the review and merge!

PR #5201, opened 2024-02-05
Prevent another deadlock.

nical:
**Connections**

Similar to #5197

**Description**

Another case of calling `release_gpu_resources` the life lock is held.

**Testing**

Caught bt Gecko's tests

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


ErichDonGubler (approved):
LGTM!

Issue #5066, opened 2024-01-15
`Device::create_bind_group` unhelpful error message for `MissingBufferUsageError`


jorendorff:
The message should say which entry is causing the problem.

Currently, it doesn't:

```
Validation Error

Caused by:
    In Device::create_bind_group
      note: label = `bind group`
    Buffer usage is BufferUsages(COPY_DST | UNIFORM) which does not contain required usage BufferUsages(STORAGE)
```

reima:
This has been fixed with #5203.

PR #5203, opened 2024-02-05
MissingBufferUsageError now has an id prop

andristarr:
**Connections**
https://github.com/gfx-rs/wgpu/issues/5066

**Description**
MissingBufferUsageError now has an id prop which is displayed when its raised

**Testing**

**Checklist**

- [ ] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/validation.rs
+++ b/wgpu-core/src/validation.rs
@@ -1,4 +1,8 @@
-use crate::{device::bgl, FastHashMap, FastHashSet};
+use crate::{
+    device::bgl,
+    id::{markers::Buffer, Id},
+    FastHashMap, FastHashSet,
+};
 use arrayvec::ArrayVec;
 use std::{collections::hash_map::Entry, fmt};
 use thiserror::Error;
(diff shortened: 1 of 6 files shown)


andristarr:
Let me know if you would prefer another approach, i figured the buffer id would be sufficient

wgpu-core/src/validation.rs
```diff
@@ -134,20 +138,26 @@ pub struct Interface {
 }
 
 #[derive(Clone, Debug, Error)]
-#[error("Buffer usage is {actual:?} which does not contain required usage {expected:?}")]
+#[error("Buffer usage is {actual:?} for {id:?} which does not contain required usage {expected:?}")]
```
nical: Let's rephrase this:

```
"Usage flags {actual:?} for buffer {id:?} do not contain required usage flags {expected:?}"
```

nical (changes requested):
Sorry about the review lag, This one fell under the radar. Just a small change and this is is good to go.

wgpu-core/src/validation.rs
andristarr: Done!

nical (approved):

nical:
Thanks!

PR #5216, opened 2024-02-07
Prevent a deadlock when destroying resources

nical:
It's risky to get write access through the snatchlock from a drop implementation since the snatch lock is typically held for large scopes. This commit makes it so we defer snatching some resources to when the device is polled and we know the snatch lock is not held.

**Testing**
 
We hit this very easily in Firefox.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -376,7 +376,6 @@ impl<A: HalApi> Device<A> {
         maintain: wgt::Maintain<queue::WrappedSubmissionIndex>,
     ) -> Result<(UserClosures, bool), WaitIdleError> {
         profiling::scope!("Device::maintain");
-
         let last_done_index = if maintain.is_wait() {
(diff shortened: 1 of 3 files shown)


wgpu-core/src/device/resource.rs
```diff
@@ -300,6 +307,56 @@ impl<A: HalApi> Device<A> {
         self.life_tracker.lock()
     }
 
+    /// Run some destroy operations that were deferred.
+    ///
+    /// Destroying the resources requires taking a write lock on the device's snatch lock,
+    /// so a good reason for deferring resource destruction is when we dont' know for sure
```
ErichDonGubler: typo:

```suggestion
    /// so a good reason for deferring resource destruction is when we don't know for sure
```

ErichDonGubler (approved):
LGTM. Deadlock whack-a-mole, whee!

PR #5218, opened 2024-02-07
Add max_color_attachments and max_color_attachment_bytes_per_sample

nical:
**Connections**

- Spec discussion: https://github.com/gpuweb/gpuweb/issues/2965
- Spec for color attachment bytes per sample: https://gpuweb.github.io/gpuweb/#abstract-opdef-calculating-color-attachment-bytes-per-sample

**Description**

This PR adds the two limits. `max_color_attachment` was already partially implemented via a constant in `wgpu-hal`. The render bundle code still uses that instead of the device limit so there is followup work to do.

For `max_color_attachment_bytes_per_sample` This does a best effort to pick reasonable defaults since the limit is largely undocumented for most platforms (See the linked gpuweb discussion). The general idea is that in doubt, 32 bytes is a safe assumption, and the limit is mostly to cater to tiled GPUs so on non-tiled ones we could increase the limits (for example d3d12 does not target non-tiled GPUs today).

**Testing**

The CTS covers these.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -260,6 +260,9 @@ impl RenderBundleEncoder {
             None => (true, true),
         };
 
+        // TODO: should be device.limits.max_color_attachments
+        let max_color_attachments = hal::MAX_COLOR_ATTACHMENTS;
+
         //TODO: validate that attachment formats are renderable,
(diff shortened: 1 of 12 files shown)


teoxoy (approved):
LGTM

PR #5222, opened 2024-02-08
[wgpu-core] Compute minimum binding size correctly for arrays.

jimblandy:
In early versions of WGSL, `storage` or `uniform` global variables had
to be either structs or runtime-sized arrays. This rule was relaxed,
and now globals can have any type; Naga automatically wraps such
variables in structs when required by the backend shading language.

Under the old rules, whenever wgpu-core saw a `storage` or `uniform`
global variable with an array type, it could assume it was a
runtime-sized array, and take the stride as the minimum binding size.
Under the new rules, wgpu-core must consider fixed-sized and
runtime-sized arrays separately.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


nical (approved):

ErichDonGubler:
@jimblandy: Looks like Windows CI is failing, and it's relevant? https://github.com/gfx-rs/wgpu/actions/runs/7822956888/job/21349870422?pr=5222#step:13:4146

jimblandy:
Hoho, maybe we can reliably reproduce #3193 now??

jimblandy:
Okay, the change between 3194089 and 7ffe936 made CI stop complaining, which is further info for #3189. I'll publish the former commit somewhere accessible.

PR #5223, opened 2024-02-08
New validation error: struct members with the location attribute...

### LINKED ISSUE #5184: Improve error message for invalid @location types.
**Motivation**
I'm a newbie to writing shaders. I recently tried to use `@location` incorrectly:
```
struct Rotor {
    s: f32,
    xy: f32,
    xz: f32,
    yz: f32
}

struct InstanceInput {
    @location(5) position: vec3<f32>,
    @location(6) scale: f32,
    @location(7) rotation: Rotor, // <-- This is the problem
};

@vertex fn main(in : InstanceInput) -> @builtin(position) vec4f { return vec4f(); }
```
The error message I got confused me:
```
> naga .\x.wgsl
error: 
   ┌─ x.wgsl:8:1
   │
 8 │ ╭ struct InstanceInput {
 9 │ │     @location(5) position: vec3<f32>,
10 │ │     @location(6) scale: f32,
11 │ │     @location(7) rotation: Rotor, // <-- This is the problem
   │ ╰────────────────────────────────^ naga::Type [4]

Entry point main at Vertex is invalid:
        Argument 0 varying error
        The type [2] does not match the varying
```
I had trouble understanding what I was doing wrong. I later learned from the WGSL Spec that [@location can only be used with numeric scalars or vectors](https://www.w3.org/TR/WGSL/#attribute-location).

By contrast, the error from tint was a lot easier to understand:
```
cannot apply @location to declaration of type 'Rotor'
@location must only be applied to declarations of numeric scalar or numeric vector type
```

**Proposed solution**
There is already some [code](https://github.com/gfx-rs/wgpu/blob/87b6513df32e8a9c588962ba8509019c277438e2/naga/src/valid/interface.rs#L46) to test against IO_SHAREABLE. We would just need to change the message.

I'd happily volunteer to make this change. Just let me know if this is an acceptable proposal.

ted-dokos:
…can only be numeric scalars or numeric vectors.

**Connections**
resolves https://github.com/gfx-rs/wgpu/issues/5184

**Description**
Naga used to allow this case through:
```
struct X { x:f32, }
struct Y { @location(0) x: X }
```
However, this is not valid wgsl. [The wgsl spec](https://www.w3.org/TR/WGSL/#attribute-location) states that location "[Must](https://www.w3.org/TR/WGSL/#shader-creation-error) only be applied to declarations of objects with [numeric scalar](https://www.w3.org/TR/WGSL/#numeric-scalar) or [numeric vector](https://www.w3.org/TR/WGSL/#numeric-vector) type."

**Testing**
Added a new test invalid_member_with_location_attribute. The test attempts to use a struct type for a member with the location attribute. This should error.

**Checklist**

- [x] Run `cargo fmt`.


ted-dokos:
Test results were not clean, but appear unrelated to my change:
```
------------
     Summary [ 140.811s] 623 tests run: 621 passed (12 flaky), 2 failed, 0 skipped
   FLAKY 2/3 [   4.927s] wgpu-examples [Executed] [Dx12/AMD Radeon RX 5700 XT/1] srgb-blend-linear
   FLAKY 2/3 [   3.801s] wgpu-examples [Executed] [Dx12/AMD Radeon RX 5700 XT/1] stencil-triangles
   FLAKY 2/3 [   2.777s] wgpu-examples [Executed] [Vulkan/AMD Radeon RX 5700 XT/0] texture-arrays-uniform
   FLAKY 2/3 [   1.319s] wgpu-test::wgpu-test [Executed Failure: ALWAYS] [Dx12/Microsoft Basic Render Driver/2] wgpu_test::buffer::empty_buffer
   FLAKY 2/3 [  10.965s] wgpu-test::wgpu-test [Executed] [Dx12/AMD Radeon RX 5700 XT/1] wgpu_test::clear_texture::clear_texture_uncompressed_gles
   FLAKY 2/3 [  36.677s] wgpu-test::wgpu-test [Executed] [Dx12/AMD Radeon RX 5700 XT/1] wgpu_test::shader::struct_layout::uniform_input
   FLAKY 2/3 [  28.881s] wgpu-test::wgpu-test [Executed] [Dx12/Microsoft Basic Render Driver/2] wgpu_test::shader::struct_layout::storage_input
   FLAKY 2/3 [   0.806s] wgpu-test::wgpu-test [Executed] [Gl/AMD Radeon RX 5700 XT/3] wgpu_test::device::device_lost_replaced_callback
   FLAKY 2/3 [   1.739s] wgpu-test::wgpu-test [Executed] [Gl/AMD Radeon RX 5700 XT/3] wgpu_test::queue_transfer::queue_write_texture_overflow
  TRY 3 FAIL [   7.847s] wgpu-examples [Executed] [Dx12/Microsoft Basic Render Driver/2] water
  TRY 3 FAIL [   2.239s] wgpu-test::wgpu-test [Executed Failure: BACKEND] [Vulkan/AMD Radeon RX 5700 XT/0] wgpu_test::shader::struct_layout::uniform_input
  ```

File: naga/src/valid/type.rs [hunk 1]
```diff
@@ -583,6 +583,26 @@ impl super::Validator {
...
+                                sampling: _,
+                            } => {
+                                let default = std::format!("{}", member.ty.index());
+                                let st = gctx.types[member.ty].name.as_ref().unwrap_or(&default);
+                                return Err(TypeError::LocationMemberUsingInvalidType {
+                                    struct_type: std::format!("[{}]", st),
+                                });
+                            }
+                        }
+                    }
```
teoxoy: I don't think we need this block as we check this here:

https://github.com/gfx-rs/wgpu/blob/07e59eb6fc7de3f682f1c401b9cf9f0da9ee4b4a/naga/src/valid/interface.rs#L268-L274

teoxoy (changes requested):

naga/src/valid/type.rs
ted-dokos: Here's why I think we should have it anyway: the following snippet is wrong according to the spec:
```
struct X { x:f32, }
struct Y { @location(0) x: X }
```
But since we haven't written an entry point function using Y yet, the check in interface.rs won't catch this.

I should probably reuse NotIOShareableType, instead of creating my own error. What do you think?

ted-dokos: @teoxoy just checking in on the above. Do you still feel that this is redundant with the check in interface.rs?

teoxoy (changes requested):
Previous discussion was not resolved.

ted-dokos:
I did respond to your comments, but apparently they were saved by GitHub as drafts and not sent.

At the moment I don't have the bandwidth to revisit this, so I will close it out.

PR #5188, opened 2024-02-02
Separate native-only feature for `wgpu::CommandEncoder::write_timestamp`

Wumpf:
**Connections**
* https://github.com/gpuweb/gpuweb/pull/4370

**Description**
WebGPU no longer supports timestamps on encoder (the only kind supported now are the ones on render/compute passes), meaning we have to split up out `TIMESTAMP_QUERY_INSIDE_ENCODERS` of the existing `TIMESTAMP_QUERY` feature.

Note that this gets our feature number at 64, meaning I had to switch the feature flag datatype to u128 (.. this will solve the problem _once and for all_! ;))

**Testing**

There's a test on timestamp that I ran on Mac. Change overall fairly straight forward, low risk.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`. If applicable, add:
  - [x] `--target wasm32-unknown-unknown`
  - [ ] `--target wasm32-unknown-emscripten`
- [x] Run `cargo xtask test` to run tests.
- [x] Add change to `CHANGELOG.md`. See simple instructions inside file.

PR #5224, opened 2024-02-08
Pin `web-sys` to `0.3.67` since we are using unstable APIs

hecrj:
**Connections**
- https://github.com/rustwasm/wasm-bindgen/issues/3834

**Description**
`wgpu` relies on unstable APIs of the `web-sys` crate and the latest `0.3.68` version breaks types related to WebGPU.

Since semantic versioning isn't enforced for unstable APIs in `web-sys`, we should rely instead on a specific crate version to avoid broken builds in the future.

**Testing**
N/A

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -160,7 +160,7 @@ js-sys = "0.3.67"
 wasm-bindgen = "0.2.87"
 wasm-bindgen-futures = "0.4.40"
 wasm-bindgen-test = "0.3"
-web-sys = "0.3.67"
+web-sys = "=0.3.67"
 web-time = "0.2.4"
 
 # deno dependencies


kpreid:
`=` version requirements are fragile. This will irreparably break builds for any dependent who has another dependency requiring a newer version of `web_sys`, and *isn't even using* the unstable apis part of `wgpu` (which could be `webgl` backend only, or even a non-web build that happens to also have an unused `web_sys` dep somewhere else).

This would be worthwhile if `wgpu` _always_ used `web_sys_unstable_apis`, but it doesn't.

hecrj:
It certainly will break builds for some users, but they will break upfront.

The current approach allows for Wasm builds to break after the fact, in code that has already been released.

In fact, [`0.19` is actually broken right now](https://github.com/iced-rs/iced/actions/runs/7812210438/job/21308782429?pr=2200).

kpreid:
This imposes a very large cost on all _users of only stable features_ ("don't upgrade your other dependencies") in order to benefit users of unstable features.

kpreid:
Hmm, would it be possible to have two `web-sys` dependencies declared, one pinned and one not, and the pinned one only enabled when `webgpu` is enabled? If Cargo accepts that, it would handle both cases well.

ErichDonGubler:
This doesn't solve the immediate problem, but: We currently plan on upgrading `web-sys` to address (see #5148, https://github.com/gfx-rs/wgpu/pull/5188). Notice that I was the one who filed <https://github.com/rustwasm/wasm-bindgen/pull/3816/>.

Cargo.toml
```diff
@@ -160,7 +160,7 @@ js-sys = "0.3.67"
 wasm-bindgen = "0.2.87"
 wasm-bindgen-futures = "0.4.40"
 wasm-bindgen-test = "0.3"
-web-sys = "0.3.67"
+web-sys = "=0.3.67"
```
nical, line 163: It would be good to have a comment here explaining why we are pinning to a specific version.

hecrj:
> Hmm, would it be possible to have two `web-sys` dependencies declared, one pinned and one not, and the pinned one only enabled when `webgpu` is enabled? If Cargo accepts that, it would handle both cases well.

It may be possible with a bit of conditional compilation. I'll look into it.

Issue #5196, opened 2024-02-05
V0.19 seems taking toooo long on preparing `BindGroupLayout`


cryscan:
**Description**
I try to upgrade [`web-rwkv`](https://github.com/cryscan/web-rwkv), an LLM inferencing backend using compute shaders. to v0.19. However after upgrading, I find that when running the model, it **gets slower and slower**, and most of the time, the GPU is **idle**. I suspect that internally, the CPU side is waiting on something.

This does not happen in v0.18.

**Repro steps**
Try to upgrade [`web-rwkv`](https://github.com/cryscan/web-rwkv) to v0.19 without touch anything other than adding `.into_iter()` when selecting adapters and run the model via
```bash
$ cargo run --example -r chat -- -m .\assets\models\rwkv-x060-3b-world-v2-28%trained-20231208-ctx4k.st -t
```

**Expected vs observed behavior**
- Expected: the inference goes at a constant speed, since the model is an RNN.
- Observed: it goes **slower and slower**, with GPU usage going down.

**Extra materials**
I did a framegraph (attached) and found that comparing to v0.18, v0.19 spent a lot of time on `wgpu::ComputePipeline::get_bind_group_layout`.

### v0.18
![v0 18](https://github.com/gfx-rs/wgpu/assets/16053640/fc103abe-df14-4725-900f-b671c2e054fc)

### v0.19
![v0 19](https://github.com/gfx-rs/wgpu/assets/16053640/598451cf-aa2a-4178-a24d-c9b6ee146960)

**Platform**
- OS: Windows 10
- WGPU: v0.19.1
- GPU: NVidia RTX 3080, 4090

nical:
@cwfitzgerald looks like it might have been caused by the bindgroup layout dedup refactor ?

cwfitzgerald:
Double checked the trace - I think this is actually arcanization.

Based on the trace provided, `wgpu_core::identity::IdentityValues::alloc` seems to be calling `max_by` on a slice. This is what is taking the time. I think this is one of a few leaks combined with linear behavior during identity value allocation.
- #5029 
- #4912

cwfitzgerald:
For those hitting this problem, are you calling get_bind_group_layout every frame? To be clear this is a bug on our side, but reducing calls to get_bind_group_layout should reduce the problem.

nathanielsimard:
[...]

PR #5229, opened 2024-02-09
Simplify the ID allocation in IdentityValues

nical:
**Connections**

Addresses part of #5196 

**Description**

Our Id allocation logic keeps track of all used IDs, grouped by epoch. There are two issues with that:
 - It's slow when there are many IDs (which can happen because of some unfortunate ID leaks)
 - It doesn't quite match the mental model of a number of the maintainers including myself.

This PR switches to a very simple and fast logic for allocating IDs, which works under the assumption that we either always allocate IDs, or always provide them externally.

~Unfortunately right now this assumption breaks in one place in the tests, so we have to fix that.` Edit: False alert, it seems to work.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


wgpu-core/src/identity.rs
```diff
@@ -37,9 +37,12 @@ use std::{fmt::Debug, marker::PhantomData};
...
 pub(super) struct IdentityValues {
     free: Vec<(Index, Epoch)>,
-    //sorted by Index
-    used: FastHashMap<Epoch, Vec<Index>>,
+    next_index: Index,
     count: usize,
+    // Sanity check: The allocation logic works under the assumption that we don't
+    // do a mix of allocating ids from here and providing ids manually for the same
+    // storage container.
+    allocate_ids: Option<bool>,
```
cwfitzgerald: Could we use an enum here? It would really help the readability of the uses to have an `enum IdSource { Allocated, UserAllocated }`

wgpu-core/src/identity.rs
```diff
@@ -48,28 +51,32 @@ impl IdentityValues {
     /// The backend is incorporated into the id, so that ids allocated with
     /// different `backend` values are always distinct.
     pub fn alloc<T: Marker>(&mut self, backend: Backend) -> Id<T> {
+        assert!(
+            self.allocate_ids != Some(false),
+            "Mix of internally allocated and externally provided IDs"
+        );
+        self.allocate_ids = Some(true);
```
cwfitzgerald:
```suggestion
        assert_eq!(
            self.allocate_ids.get_or_insert(true),
            true,
            "Mix of internally allocated and externally provided IDs"
        );
```

Another form which feels personally cleaner.

cwfitzgerald (changes requested):

cwfitzgerald (approved):

PR #5211, opened 2024-02-06
add test that verifies that we can drop the queue before using the device to create a command encoder

Davidster:
I guess the issue would actually need to be fixed before this test gets merged?

**Connections**
n/a

**Description**
I noticed some code in the wgpu-profiler tests where the queue is dropped before we call Device::create_command_encoder. It produces the following panic:

```
thread 'src::errors::end_frame_unresolved_query' panicked at C:\Users\s\.cargo\registry\src\index.crates.io-6f17d22bba15001f\wgpu-core-0.19.0\src\storage.rs:113:39:
Queue[Id(0,1,vk)] does not exist
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library\std\src\panicking.rs:645
   1: core::panicking::panic_fmt
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library\core\src\panicking.rs:72
   2: wgpu_core::storage::Storage<wgpu_core::device::queue::Queue<wgpu_hal::vulkan::Api>,wgpu_core::id::Id<wgpu_core::device::resource::Device<wgpu_hal::empty::Api> > >::get<wgpu_core::device::queue::Queue<wgpu_hal::vulkan::Api>,wgpu_core::id::Id<wgpu_core::dev
             at C:\Users\s\.cargo\registry\src\index.crates.io-6f17d22bba15001f\wgpu-core-0.19.0\src\storage.rs:113
   3: wgpu_core::storage::Storage<wgpu_core::device::queue::Queue<wgpu_hal::vulkan::Api>,wgpu_core::id::Id<wgpu_core::device::resource::Device<wgpu_hal::empty::Api> > >::get_owned<wgpu_core::device::queue::Queue<wgpu_hal::vulkan::Api>,wgpu_core::id::Id<wgpu_cor
             at C:\Users\s\.cargo\registry\src\index.crates.io-6f17d22bba15001f\wgpu-core-0.19.0\src\storage.rs:128
   4: wgpu_core::registry::Registry<wgpu_core::id::Id<wgpu_core::device::resource::Device<wgpu_hal::empty::Api> >,wgpu_core::device::queue::Queue<wgpu_hal::vulkan::Api> >::get<wgpu_core::id::Id<wgpu_core::device::resource::Device<wgpu_hal::empty::Api> >,wgpu_co
[...]

Issue #5120, opened 2024-01-22
Resolve cycles in wgpu-core using weak references


nical:
At the moment the device has string reference to the resources and the resources have strong references to the device. This cycle is manually broken in `triage_suspected` by looking at the reference count.

We want to remove this manual reference count management and only rely on weak refst to break the cycles.

@cwfitzgerald put together a diagram that illustrates the end result: 

![Diagram of future strong and weak reference in wgpu-core](https://github.com/gfx-rs/wgpu/assets/264854/0e15f6fc-8ab5-444c-8bd8-f13fc03d217a)

nical:
@cwfitzgerald how does the queue fit in the above diagram?  The WebGPU spec says that the device has a strong reference to the queue (but we can make it work in the browser either way by having the JS-facing objects have the strong references in the right order even if wgpu-core internally has it the other way around).
Currently the queue has a strong reference in to the device in wgpu-core.

See also #5230.

cwfitzgerald:
@nical I believe it's "queue owns device", "device has weakref to queue". This doesn't really vibe with the idempotent queue method call on device in WebGPU, but I don't think we need to have that in wgpu-core.

If we were to model the WebGPU way, it would be "device owns InternalQueue", "queue owns InternalQueue". This way the device can hand out the same internal queue over and over to separate `Queue` objects.

PR #5230, opened 2024-02-09
Store the device's queue via a weak ref instead of an ID

nical:
**Connections**

 - #5120
 - #5211

**Description**

This is one of the internal uses of IDs instead of Arc/Weak references.

With this PR the Queue has an strong reference to the device and the device has a weak ref to the queue.
I think that to align with the WebGPU spec we should do the opposite (The device owns the queue). However that's more complicated to do this way because of the way the device and queue are destroyed currently (not impossible but more complicated).

This does not quite fix the test  from #5211, but it turns the panic into a validation error in `create_command_encoder`.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -13,7 +13,6 @@ use crate::{
     hal_api::HalApi,
     hal_label,
     hub::Hub,
-    id::QueueId,
     init_tracker::{
         BufferInitTracker, BufferInitTrackerAction, MemoryInitKind, TextureInitRange,
         TextureInitTracker, TextureInitTrackerAction,
@@ -56,7 +55,7 @@ use std::{
 
 use super::{
     life::{self, ResourceMaps},
-    queue::{self},
+    queue::{self, Queue},
     DeviceDescriptor, DeviceError, ImplicitPipelineContext, UserClosures, ENTRYPOINT_FAILURE_ERROR,
     IMPLICIT_BIND_GROUP_LAYOUT_ERROR_LABEL, ZERO_BUFFER_SIZE,
 };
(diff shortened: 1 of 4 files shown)


wgpu-core/src/device/resource.rs
```diff
@@ -359,6 +358,14 @@ impl<A: HalApi> Device<A> {
         }
     }
 
+    pub fn get_queue(&self) -> Option<Arc<Queue<A>>> {
+        self.queue.read().as_ref()?.upgrade()
+    }
+
+    pub fn set_queue(&self, queue: Arc<Queue<A>>) {
+        self.queue.write().replace(Arc::downgrade(&queue));
```
cwfitzgerald: I feel like there should be an assert that this is None - this function should never be called more than once. In fact we could use a OnceCell for the Arc instead of an Mutex<Option<>> - that more actually models what we want (late init)

wgpu-core/src/device/resource.rs
```diff
@@ -89,7 +88,7 @@ use super::{
 pub struct Device<A: HalApi> {
     raw: Option<A::Device>,
     pub(crate) adapter: Arc<Adapter<A>>,
-    pub(crate) queue_id: RwLock<Option<QueueId>>,
+    pub(crate) queue: RwLock<Option<Weak<Queue<A>>>>,
```
cwfitzgerald: This isn't a problem in this PR, but I think this is a problem in this design. I think we can get in a situation where the device exists but the queue has been dropped. I do think we need to re-think about how the device/queue is modeld.

cwfitzgerald (changes requested):

cwfitzgerald (approved):
Very nice

Issue #5217, opened 2024-02-07
Allow recycling IDs as soon as the `drop_*` wgpu-core function is called 


nical:
We recently found out that the hook for telling wgpu-core users that allocate their own IDs that an index is safe to reuse has been refactored away at some point. So Gecko currently never reuse any index. Oops.

We could re-introduce a hook for communicating reusable registry indices, but I'd like to simplify things to the point that we don't have to.

Currently the `Drop` impl for `ResourceInfo` is [what tells the registry that an index can be reused](https://github.com/gfx-rs/wgpu/blob/20fda698341efbdc870b8027d6d49f5bf3f36109/wgpu-core/src/resource.rs#L79). It is called when the reference-counted resource struct is dropped which in principle should happen after we call wgpu-core's `<resource>_drop` function... Except if the resource's `Arc` is moved out of the registry before `<resource>_drop` in which case the index might be made reusable to soon. That's what causing the headaches in #5141.
My main issue with the current system is that by having reference-counting in between the registry and its recycling logic, we introduced an invariant that was easy to break and broke it without noticing. Instead I would like the registry to completely own the ID reuse logic without being affected by reference counting that is hard to predict.

# Proposal

As soon as the wgpu-core resource_drop functions are called, we make the ID reusable. To do this we need:

- [ ] Gecko to remove wpgu-core code in the content process (patches being reviewed)
- [ ] Remove all internal usage of IDs in wgpu-core
- [ ] Reuse Ids as soon as `<resource>_drop`. For Gecko this means we have a simple rule for when it is safe to reuse IDs.

nical:
This would also fix our current issue with leaked deduplicated bindgroup registry slots.

Issue #4912, opened 2023-12-21
Deduplicated Bind Group Layout Ids Aren't Recycled.


cwfitzgerald:
We do not properly recycle BGL IDs that are deduplicated. The reason for this is that the ResourceInfo only points to the one id, so only recycles that one ID.

Will work on this as part of the BGL deduplication work.

PR #5242, opened 2024-02-12
Store Arcs instead of IDs in render bundles

nical:
**Connections**

- https://github.com/gfx-rs/wgpu/issues/5120
- https://github.com/gfx-rs/wgpu/issues/5217
- https://github.com/gfx-rs/wgpu/pull/5141
- https://github.com/gfx-rs/wgpu/issues/4912

**Description**

This is a step towards not relying on IDs internally. It is also a dependency to not using IDs in trackers and fix our deduplicated resource leaks since the current bundle code uses trackers to store and get the actual resource arcs via IDs.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/command/draw.rs
+++ b/wgpu-core/src/command/draw.rs
@@ -2,16 +2,21 @@
 !*/
 
 use crate::{
-    binding_model::{LateMinBufferBindingSizeMismatch, PushConstantUploadError},
+    binding_model::{BindGroup, LateMinBufferBindingSizeMismatch, PushConstantUploadError},
     error::ErrorFormatter,
+    hal_api::HalApi,
     id,
+    pipeline::RenderPipeline,
+    resource::{Buffer, QuerySet},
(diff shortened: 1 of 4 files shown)


wgpu-core/src/command/draw.rs
```diff
@@ -245,3 +250,115 @@ pub enum RenderCommand {
     EndPipelineStatisticsQuery,
     ExecuteBundle(id::RenderBundleId),
 }
+
+/// Equivalent to `RenderCommand` with the Ids resolved into resource Arcs.
+#[doc(hidden)]
+#[derive(Clone, Debug)]
+pub enum ArcRenderCommand<A: HalApi> {
```
cwfitzgerald, line 257: As long as the long term goal of this is to remove the duplication, I'm fine with this for now.

cwfitzgerald (approved):

PR #5244, opened 2024-02-12
Use a unique tracker index per resource instead of the ID in trackers

### LINKED ISSUE #4912: Deduplicated Bind Group Layout Ids Aren't Recycled.

### LINKED ISSUE #5217: Allow recycling IDs as soon as the `drop_*` wgpu-core function is called

nical:
**Connections**

Fixes #5217
Fixes #4912
Required for fixing #5141
An important step towards #5120

**Description**

Before this PR, trackers use the ID's index to manage per-resource information. This gets in the way of a lot of things (see linked issues). This PR adds a per-resource tracker index that is independent from the resource ID. Separating the two concepts lets us handle deduplicated IDs which point to the same resource and is an important step towards removing the IDs in favor of exposing the Arcs.

This PR also lets IDs be recycled as soon as their user-facing handle is dropped, which fixes the de-duplicated resource leaks and will allow Gecko to fix the recycling of handles.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -260,7 +260,7 @@ impl Global {
                 .trackers
                 .lock()
                 .buffers
-                .insert_single(id, resource, buffer_use);
+                .insert_single(resource, buffer_use);
 
             return (id, None);
         };
@@ -527,7 +527,7 @@ impl Global {
                 .lock_life()
                 .suspected_resources
                 .buffers
-                .insert(buffer_id, buffer);
+                .insert(buffer.info.tracker_index(), buffer);
         }
 
         if wait {
@@ -571,11 +571,11 @@ impl Global {
             let (id, resource) = fid.assign(texture);
             api_log!("Device::create_texture({desc:?}) -> {id:?}");
 
-            device.trackers.lock().textures.insert_single(
-                id,
-                resource,
-                hal::TextureUses::UNINITIALIZED,
-            );
+            device
+                .trackers
+                .lock()
+                .textures
+                .insert_single(resource, hal::TextureUses::UNINITIALIZED);
 
             return (id, None);
         };
@@ -645,11 +645,11 @@ impl Global {
             let (id, resource) = fid.assign(texture);
             api_log!("Device::create_texture({desc:?}) -> {id:?}");
 
-            device.trackers.lock().textures.insert_single(
-                id,
-                resource,
-                hal::TextureUses::UNINITIALIZED,
-            );
+            device
+                .trackers
+                .lock()
+                .textures
+                .insert_single(resource, hal::TextureUses::UNINITIALIZED);
 
             return (id, None);
         };
@@ -702,7 +702,7 @@ impl Global {
                 .trackers
                 .lock()
                 .buffers
-                .insert_single(id, buffer, hal::BufferUses::empty());
+                .insert_single(buffer, hal::BufferUses::empty());
 
             return (id, None);
         };
@@ -762,7 +762,7 @@ impl Global {
                         .lock_life()
                         .suspected_resources
                         .textures
-                        .insert(texture_id, texture.clone());
+                        .insert(texture.info.tracker_index(), texture.clone());
                 }
             }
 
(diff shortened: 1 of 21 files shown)


nical:
This PR contains https://github.com/gfx-rs/wgpu/pull/5242

wgpu-core/src/track/mod.rs
```diff
@@ -122,6 +118,104 @@ pub(crate) use texture::{
...
+            textures: Arc::new(Mutex::new(TrackerIndexAllocator::new())),
+            texture_views: Arc::new(Mutex::new(TrackerIndexAllocator::new())),
+            samplers: Arc::new(Mutex::new(TrackerIndexAllocator::new())),
+            bind_groups: Arc::new(Mutex::new(TrackerIndexAllocator::new())),
+            bind_group_layouts: Arc::new(Mutex::new(TrackerIndexAllocator::new())),
+            compute_pipelines: Arc::new(Mutex::new(TrackerIndexAllocator::new())),
+            render_pipelines: Arc::new(Mutex::new(TrackerIndexAllocator::new())),
+            pipeline_layouts: Arc::new(Mutex::new(TrackerIndexAllocator::new())),
+            bundles: Arc::new(Mutex::new(TrackerIndexAllocator::new())),
+            query_sets: Arc::new(Mutex::new(TrackerIndexAllocator::new())),
```
cwfitzgerald: Could we internalize these locks? It would make the call-sites simpler and prevent long holding locks.

wgpu-core/src/device/global.rs
```diff
@@ -1472,7 +1468,7 @@ impl Global {
                 .lock_life()
                 .suspected_resources
                 .render_bundles
-                .insert(render_bundle_id, bundle.clone());
+                .insert(bundle.info.tracker_index(), bundle.clone());
```
cwfitzgerald, line 1471: This feels like tracker indices shouldn't go in suspected resources, but I think this is okay, as I believe suspected resources is going away soon?

cwfitzgerald (changes requested):

wgpu-core/src/device/global.rs
nical: I think that they should, or at least the idea is that resource IDs don't go in there anymore so something else has to.

> I believe suspected resources is going away soon?

That would be ideal

wgpu-core/src/track/mod.rs
nical: Done.

cwfitzgerald (approved):
Woot!

PR #5248, opened 2024-02-13
Add an advanced_debugging preset including wgpu base validation

nical:
And remove it from the default debugging configuration.

**Description**

GPU based validation can have a very large overhead (in the order of minutes to show the first frame of a simple demo on my windows laptop with d3d12). It's a great tool but it's too expensive to be the recommended configuration for debug builds. This PR:
 - Removes gpu based validation from `InstanceFlags::debuggin()` (and transitively `InstanceFlags::from_build_config()` in debug builds.
 - Adds `InstanceFlags::advanced_debugging()` which includes gpu based validation.
 - Makes our test use advanced debugging to preserve the full suite of validation on CI.
 - Add an environment variable for gpu based validation in InstanceFlags::with_env`.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -924,7 +924,12 @@ impl Default for InstanceFlags {
 impl InstanceFlags {
     /// Enable recommended debugging and validation flags.
     pub fn debugging() -> Self {
-        InstanceFlags::DEBUG | InstanceFlags::VALIDATION | InstanceFlags::GPU_BASED_VALIDATION
+        InstanceFlags::DEBUG | InstanceFlags::VALIDATION
+    }
+
+    /// Enable advanced debugging and validation flags (potentially very slow).
+    pub fn advanced_debugging() -> Self {
+        Self::debugging() | InstanceFlags::GPU_BASED_VALIDATION
     }
 
     /// Infer good defaults from the build type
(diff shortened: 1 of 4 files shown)


cwfitzgerald (approved):

wgpu-types/src/lib.rs
```diff
@@ -964,6 +969,9 @@ impl InstanceFlags {
         if let Some(bit) = env("WGPU_DEBUG") {
             self.set(Self::DEBUG, bit);
         }
+        if let Some(bit) = env("WGPU_GPU_BASED_VALIDATION") {
+            self.set(Self::GPU_BASED_VALIDATION, bit);
+        }
```
ErichDonGubler: issue: uh

This already exists a few lines below. 😅

nical: wooops

wgpu-types/src/lib.rs
ErichDonGubler, line 1: issue: We should change the verbiage in the `CHANGELOG`, since it notes this being the default (which would be incorrect once this merges). I guess you could just append this PR number to that `CHANGELOG` entry, too.

ErichDonGubler (changes requested):
Just a couple of comms. things, otherwise LGTM!

ErichDonGubler (approved):

PR #5249, opened 2024-02-14
Implement override-expression evaluation

teoxoy:
Mainly implements override-expression evaluation for initializers of override declarations.


naga/src/valid/mod.rs
```diff
@@ -172,6 +172,7 @@ pub struct Validator {
     switch_values: FastHashSet<crate::SwitchValue>,
     valid_expression_list: Vec<Handle<crate::Expression>>,
     valid_expression_set: BitSet,
+    override_ids: BitSet,
```
jimblandy: Since pipeline constant ids can range from 0 to 65535 and aren't generally expected to be contiguous, it seems like this should probably be a `FastHashSet`, since a `BitSet` will expand out to i/8 bytes, for a maximum set member `i`.

naga/src/front/wgsl/lower/mod.rs
```diff
@@ -1619,6 +1609,9 @@ impl<'source, 'temp> Lowerer<'source, 'temp> {
                     LoweredGlobalDecl::Const(handle) => {
                         Typed::Plain(crate::Expression::Constant(handle))
                     }
+                    LoweredGlobalDecl::Override(handle) => {
+                        Typed::Plain(crate::Expression::Override(handle))
+                    }
```
jimblandy, line 1671: Just checking: this is unrelated to the rest of the commit, right?

Suggestion, for the future's sake:

```
                     LoweredGlobalDecl::Override(handle) => {
                         Typed::Plain(crate::Expression::Override(handle))
                     }
-                    _ => {
+                    LoweredGlobalDecl::Function(_)
+                    | LoweredGlobalDecl::Type(_)
+                    | LoweredGlobalDecl::EntryPoint => {
                         return Err(Error::Unexpected(span, ExpectedToken::Variable));
                     }
                 };
```

naga/src/proc/constant_evaluator.rs
```diff
@@ -585,9 +585,9 @@ impl<'a> ConstantEvaluator<'a> {
     /// [`ZeroValue`], and [`Swizzle`] expressions - to the expression arena
     /// `self` contributes to.
     ///
-    /// If `expr`'s value cannot be determined at compile time, return a an
-    /// error. If it's acceptable to evaluate `expr` at runtime, this error can
-    /// be ignored, and the caller can append `expr` to the arena itself.
+    /// If `expr`'s value cannot be determined at compile time, return an
+    /// error. If it's acceptable to evaluate `expr` at runtime, the `expr`
+    /// will be appended to the arena and no error will be returned.
```
jimblandy: Suggest:

```suggestion
    /// If `expr`'s value cannot be determined at compile time, and `self` is
    /// contributing to some function's expression arena, then append `expr` to
    /// that arena unchanged (and thus unevaluated). Otherwise, `self` must be
    /// contributing to the module's constant expression arena; since `expr`'s
    /// value is not a constant, return an error.
```

naga/src/proc/constant_evaluator.rs
```diff
@@ -599,6 +599,22 @@ impl<'a> ConstantEvaluator<'a> {
...
     pub fn try_eval_and_append(
+        &mut self,
+        expr: Expression,
+        span: Span,
+    ) -> Result<Handle<Expression>, ConstantEvaluatorError> {
+        let res = self.try_eval_and_append_impl(&expr, span);
+        if self.function_local_data.is_some() {
+            match res {
+                Ok(h) => Ok(h),
+                Err(_) => Ok(self.append_expr(expr, span, false)),
```
jimblandy: I don't think this is correct. Division by zero errors, for example, need to be reported, not delayed until runtime.

I think we need to somehow split up `ConstantEvaluatorError` into two categories: errors because the expression cannot be evaluated until runtime, and errors that occurred even though the expression is a proper constant expression. Then we can treat those two differently here.

jimblandy (changes requested):
Still reviewing, but I found one thing that (I think?) needs to change.

naga/src/front/wgsl/lower/mod.rs
teoxoy: Yeah, this should have ben part of be0c12289a83f79a1d43b0428fa3524e8c5efb3f not 397b6731980dbb06ffb750e97784e1174eb7703d.

naga/src/proc/constant_evaluator.rs
teoxoy: Good catch! I think this is an existing issue though, I will open an issue.

teoxoy: Actually I think be0c12289a83f79a1d43b0428fa3524e8c5efb3f took care of this. We now force evaluation to succeed if the expression was a const-expression. This is how I discovered we were not evaluating bool constructors properly (which I fixed in 33b392467919917a462d10f99f63cb707b8922f7).

jimblandy:
@teoxoy I pushed some comments - could you see if they're correct?

teoxoy:
They look great, thank you!

jimblandy (approved):
Okay, this looks good, let's merge it.

ErichDonGubler:
FTR: CI was failing because of https://github.com/gfx-rs/wgpu/pull/5341, but that's 👌🏻 okay, already been fixed because it was merged.

jimblandy:
> FTR: CI was failing because of #5341, but that's 👌🏻 okay, already been fixed because it was merged.

I pushed [a074d82](https://github.com/gfx-rs/wgpu/pull/5249/commits/a074d82966da808d1f6afc331d4b539f4d017faf) to fix that. All checks are now green.

Also, note that I've only landed this in `gfx-rs:pipeline-constants`, not `trunk`.

PR #5251, opened 2024-02-14
Free command encoders' platform resources on drop on _all_ platforms

### LINKED ISSUE #3193: Early frees on CPU Implementations

ErichDonGubler:
**Connections**

This is my attempt to investigate and resolve <https://github.com/gfx-rs/wgpu/issues/3193>. ~I believe this code is ready to go, but have not yet confirmed internal details to my satisfaction. I'm leaving this as WIP for now because I will be unavailable to work on finalizing details until \~2024-02-21. If somebody wants to pick it up before then, great!~

**Description**

I believe that #3193 is caused by not freeing command encoders early enough. This is apparent on the DX12 backend, where the DX12 API actually emits an error to inform us of our mistake. This is likely causing issues in other backends (i.e., Vulkan), where the symptoms aren't as obvious. Metal has already been cared for by @bradwerth with #2077, so only the other backends need attention here.

Interestingly, we don't currently do any clean-up of command encoders when they are dropped, but we do when they're `finish`ed. This seems wrong. So, this change adds a `CommandEncoder::discard_encoding` call for command encoders within `wgpu_core::Global::command_encoder_drop`. This change also adds a similar call to the `Drop` implementation of backends that appear to be idempotent in their `CommandEncoder::discard_encoder` calls (N.B., this does not include Vulkan). This appears to resolve DX12 API errors.

To be completely clear about how each backend is changed here:

| Backend | Status before this PR     | Status after this PR |
| ------- | ------------------------- | -------------------- |
| Vulkan  | Possibly has issues       | ✅Fixed.             |
| DX12    | Reporting hard errors     | ✅No more errors!    |
| Metal   | Already handled by #2077. | -                    |
| GLES    | Possibly has issues       | ✅Fixed.             |

**Testing**

~A test (`wgpu_test::buffer::command_encoder_without_finish_fine`) has been added, demonstrating both the expected failure before the fix and resolution to "just" passing.~

This change changes numerous expected error cases for tests for `DX12`, most notably `wgpu_tests::encoder::drop_encoder_after_error`, to simply be passing.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -1377,6 +1377,7 @@ impl Global {
             .command_buffers
             .unregister(command_encoder_id.transmute())
         {
+            cmd_buf.data.lock().as_mut().unwrap().encoder.discard();
             cmd_buf
                 .device
                 .untrack(&cmd_buf.data.lock().as_ref().unwrap().trackers);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,7 @@ Bottom level categories:
 - Fix panic when creating a surface while no backend is available. By @wumpf [#5166](https://github.com/gfx-rs/wgpu/pull/5166)
 - Correctly compute minimum buffer size for array-typed `storage` and `uniform` vars. By @jimblandy [#5222](https://github.com/gfx-rs/wgpu/pull/5222)
 - Fix timeout when presenting a surface where no work has been done. By @waywardmonkeys in [#5200](https://github.com/gfx-rs/wgpu/pull/5200)
+- Fix an issue where command encoders weren't properly freed if an error occurred during command encoding. By @ErichDonGubler in [#5251](https://github.com/gfx-rs/wgpu/pull/5251).
 
 #### WGL
 
--- a/tests/tests/bind_group_layout_dedup.rs
+++ b/tests/tests/bind_group_layout_dedup.rs
@@ -1,9 +1,6 @@
 use std::num::NonZeroU64;
 
-use wgpu_test::{
-    fail, gpu_test, FailureCase, GpuTestConfiguration, TestParameters, TestingContext,
-};
-use wgt::Backends;
+use wgpu_test::{fail, gpu_test, GpuTestConfiguration, TestParameters, TestingContext};
 
 const SHADER_SRC: &str = "
 @group(0) @binding(0)
@@ -307,18 +304,10 @@ fn bgl_dedupe_derived(ctx: TestingContext) {
     ctx.queue.submit(Some(encoder.finish()));
 }
 
-const DX12_VALIDATION_ERROR: &str = "The command allocator cannot be reset because a command list is currently being recorded with the allocator.";
-
 #[gpu_test]
 static SEPARATE_PROGRAMS_HAVE_INCOMPATIBLE_DERIVED_BGLS: GpuTestConfiguration =
     GpuTestConfiguration::new()
-        .parameters(
-            TestParameters::default()
-                .test_features_limits()
-                .expect_fail(
-                    FailureCase::backend(Backends::DX12).validation_error(DX12_VALIDATION_ERROR),
-                ),
-        )
+        .parameters(TestParameters::default().test_features_limits())
         .run_sync(separate_programs_have_incompatible_derived_bgls);
 
 fn separate_programs_have_incompatible_derived_bgls(ctx: TestingContext) {
@@ -376,13 +365,7 @@ fn separate_programs_have_incompatible_derived_bgls(ctx: TestingContext) {
 #[gpu_test]
 static DERIVED_BGLS_INCOMPATIBLE_WITH_REGULAR_BGLS: GpuTestConfiguration =
     GpuTestConfiguration::new()
-        .parameters(
-            TestParameters::default()
-                .test_features_limits()
-                .expect_fail(
-                    FailureCase::backend(Backends::DX12).validation_error(DX12_VALIDATION_ERROR),
-                ),
-        )
+        .parameters(TestParameters::default().test_features_limits())
         .run_sync(derived_bgls_incompatible_with_regular_bgls);
 
 fn derived_bgls_incompatible_with_regular_bgls(ctx: TestingContext) {
--- a/tests/tests/encoder.rs
+++ b/tests/tests/encoder.rs
@@ -21,16 +21,9 @@ static DROP_QUEUE_BEFORE_CREATING_COMMAND_ENCODER: GpuTestConfiguration =
                 device.create_command_encoder(&wgpu::CommandEncoderDescriptor::default());
         });
 
-// This test crashes on DX12 with the exception:
-//
-// ID3D12CommandAllocator::Reset: The command allocator cannot be reset because a
-// command list is currently being recorded with the allocator. [ EXECUTION ERROR
-// #543: COMMAND_ALLOCATOR_CANNOT_RESET]
-//
-// For now, we mark the test as failing on DX12.
 #[gpu_test]
 static DROP_ENCODER_AFTER_ERROR: GpuTestConfiguration = GpuTestConfiguration::new()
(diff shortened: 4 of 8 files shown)


wgpu-core/src/device/global.rs
```diff
@@ -1377,6 +1377,7 @@ impl Global {
             .command_buffers
             .unregister(command_encoder_id.transmute())
         {
+            cmd_buf.data.lock().as_mut().unwrap().encoder.discard();
```
ErichDonGubler, line 1380: This `wgpu-core`-centric change might be relevant to the issue that #4023 was trying to solve in Metal; ~~perhaps that fix is no longer necessary, with this in place~~? CC @bradwerth.

- [x] Investigate.

EDIT: Changed my mind, see below.

ErichDonGubler: This also means that the test I had added before is redundant, woot!

ErichDonGubler: I think the appropriate solution is actually to have every encoder for a specific back end discard itself on drop. That makes sense for _any_ stand-alone structure that can definitively conclude OS platform resource usage on `Drop` (minus maybe needing diagnostic context to report issues). We could also have the core abstraction discard the encoding—it would be redundant, technically, but I don't see that as a bad thing.

ErichDonGubler: Added a `Drop` impl. for DX12. Also added one for GLES.

It seems that Vulkan specifically does not appear to track its current state, relying on correct usage from `wgpu-core`. I won't add a `Drop` call to `discard_encoding` for that backend.

nical (approved):

wgpu-core/src/device/global.rs
ErichDonGubler: Added a follow-up for Vulkan at <https://github.com/gfx-rs/wgpu/issues/5255> .

torokati44:
*anxiously waiting for a release with this fix in, preferably in the `0.19` series*
https://github.com/ruffle-rs/ruffle/actions/runs/8046091966/job/21972653869

ErichDonGubler:
@torokati44: Have you confirmed that this fix addresses an issue in Ruffle? It's unclear if the CI issue you're linking is caused by this fix, or fixed by this fix. 😅

torokati44:
Since Ruffle is on `v0.19.1`, which was released before this got merged, the CI issue is definitely _not caused_ by this PR, but _hopefully fixed_ by it.
And, no, haven't verified it yet, but I'd give it more than 60% chance, just by gut feeling. :no_mouth:

torokati44:
So this did not end up being backported into `0.19.2`? :face_with_spiral_eyes:

ErichDonGubler:
@cwfitzgerald ☝🏻

cwfitzgerald:
Agh - I honestly thought this PR was caught up in all the other internal work, I didn't realize it was so backportable. There will be a 0.19.3 decently soon, so I'll include this in that

PR #5254, opened 2024-02-15
Add a vertex pulling flag to Metal pipelines.

### LINKED ISSUE #2432: Vertex pulling on Metal
In order to make vertex attribute data safe, we need to generate MSL that does vertex pulling manually based on the vertex index. Most of the work here would fall on Naga, but wgpu needs to provide the necessary plumbing.

bradwerth:
Add an experimental vertex pulling flag to Metal pipelines.

This proves a flag in msl::PipelineOptions that attempts to write all Metal vertex entry points to use a vertex pulling technique. It does this by:

1) Forcing the _buffer_sizes structure to be generated for all vertex entry points.
2) Adding new args to vertex entry points for the vertex id and for the bound buffers.
3) Adding code at the beginning of the function for vertex entry points to compare the vertex id against the lengths of all the bound buffers, and force an early-exit if the bounds are violated.
4) Extracting the raw bytes from the vertex buffer(s) and unpacking those bytes into the bound attributes with the expected types.
5) Replacing the varyings input and instead using the unpacked attributes to fill any structs-as-args that are rebuilt in the entry point.

A new naga test is added which exercises this flag and demonstrates the effect of the transform. The msl generated by this test  passes validation.

Future work will add reftests confirming that transformed shaders produce the same result as untransformed shaders. Eventually this transformation will be the default behavior and the flag will be removed.

**Connections**
Resolves https://github.com/gfx-rs/wgpu/issues/2432.

**Description**
This is infrastructure to support Metal OOB index security checks by transforming all generated msl shaders to use a vertex pulling strategy.

**Testing**
The only testing is a new naga test, vertex-pulling-transform, which merely confirms that the output msl is valid, not that it is useful.

**Checklist**

- [X] Run `cargo fmt`.

### DIFF
--- a/naga/src/back/msl/mod.rs
+++ b/naga/src/back/msl/mod.rs
@@ -222,6 +222,113 @@ impl Default for Options {
     }
 }
 
+/// Corresponds to [WebGPU `GPUVertexFormat`](
+/// https://gpuweb.github.io/gpuweb/#enumdef-gpuvertexformat).
+#[repr(u32)]
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq)]
+#[cfg_attr(feature = "serialize", derive(serde::Serialize))]
+#[cfg_attr(feature = "deserialize", derive(serde::Deserialize))]
+pub enum VertexFormat {
+    /// Two unsigned bytes (u8). `vec2<u32>` in shaders.
+    Uint8x2 = 0,
+    /// Four unsigned bytes (u8). `vec4<u32>` in shaders.
+    Uint8x4 = 1,
+    /// Two signed bytes (i8). `vec2<i32>` in shaders.
+    Sint8x2 = 2,
+    /// Four signed bytes (i8). `vec4<i32>` in shaders.
+    Sint8x4 = 3,
+    /// Two unsigned bytes (u8). [0, 255] converted to float [0, 1] `vec2<f32>` in shaders.
+    Unorm8x2 = 4,
+    /// Four unsigned bytes (u8). [0, 255] converted to float [0, 1] `vec4<f32>` in shaders.
+    Unorm8x4 = 5,
+    /// Two signed bytes (i8). [-127, 127] converted to float [-1, 1] `vec2<f32>` in shaders.
+    Snorm8x2 = 6,
+    /// Four signed bytes (i8). [-127, 127] converted to float [-1, 1] `vec4<f32>` in shaders.
+    Snorm8x4 = 7,
+    /// Two unsigned shorts (u16). `vec2<u32>` in shaders.
+    Uint16x2 = 8,
+    /// Four unsigned shorts (u16). `vec4<u32>` in shaders.
+    Uint16x4 = 9,
+    /// Two signed shorts (i16). `vec2<i32>` in shaders.
+    Sint16x2 = 10,
+    /// Four signed shorts (i16). `vec4<i32>` in shaders.
+    Sint16x4 = 11,
+    /// Two unsigned shorts (u16). [0, 65535] converted to float [0, 1] `vec2<f32>` in shaders.
+    Unorm16x2 = 12,
+    /// Four unsigned shorts (u16). [0, 65535] converted to float [0, 1] `vec4<f32>` in shaders.
+    Unorm16x4 = 13,
+    /// Two signed shorts (i16). [-32767, 32767] converted to float [-1, 1] `vec2<f32>` in shaders.
+    Snorm16x2 = 14,
+    /// Four signed shorts (i16). [-32767, 32767] converted to float [-1, 1] `vec4<f32>` in shaders.
+    Snorm16x4 = 15,
+    /// Two half-precision floats (no Rust equiv). `vec2<f32>` in shaders.
+    Float16x2 = 16,
+    /// Four half-precision floats (no Rust equiv). `vec4<f32>` in shaders.
+    Float16x4 = 17,
+    /// One single-precision float (f32). `f32` in shaders.
+    Float32 = 18,
+    /// Two single-precision floats (f32). `vec2<f32>` in shaders.
+    Float32x2 = 19,
+    /// Three single-precision floats (f32). `vec3<f32>` in shaders.
+    Float32x3 = 20,
+    /// Four single-precision floats (f32). `vec4<f32>` in shaders.
+    Float32x4 = 21,
+    /// One unsigned int (u32). `u32` in shaders.
+    Uint32 = 22,
+    /// Two unsigned ints (u32). `vec2<u32>` in shaders.
+    Uint32x2 = 23,
+    /// Three unsigned ints (u32). `vec3<u32>` in shaders.
+    Uint32x3 = 24,
+    /// Four unsigned ints (u32). `vec4<u32>` in shaders.
+    Uint32x4 = 25,
+    /// One signed int (i32). `i32` in shaders.
+    Sint32 = 26,
+    /// Two signed ints (i32). `vec2<i32>` in shaders.
+    Sint32x2 = 27,
+    /// Three signed ints (i32). `vec3<i32>` in shaders.
+    Sint32x3 = 28,
+    /// Four signed ints (i32). `vec4<i32>` in shaders.
+    Sint32x4 = 29,
+    /// Three unsigned 10-bit integers and one 2-bit integer, packed into a 32-bit integer (u32). [0, 1024] converted to float [0, 1] `vec4<f32>` in shaders.
+    #[cfg_attr(feature = "serde", serde(rename = "unorm10-10-10-2"))]
+    Unorm10_10_10_2 = 34,
+}
+
+/// A mapping of vertex buffers and their attributes to shader
+/// locations.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serialize", derive(serde::Serialize))]
+#[cfg_attr(feature = "deserialize", derive(serde::Deserialize))]
+pub struct AttributeMapping {
+    /// Shader location associated with this attribute
+    pub shader_location: u32,
+    /// Offset in bytes from start of vertex buffer structure
+    pub offset: u32,
+    /// Format code to help us unpack the attribute into the type
+    /// used by the shader. Codes correspond to a 0-based index of
+    /// <https://gpuweb.github.io/gpuweb/#enumdef-gpuvertexformat>.
+    /// The conversion process is described by
+    /// <https://gpuweb.github.io/gpuweb/#vertex-processing>.
+    pub format: VertexFormat,
+}
+
+/// A description of a vertex buffer with all the information we
+/// need to address the attributes within it.
+#[derive(Debug, Default, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serialize", derive(serde::Serialize))]
+#[cfg_attr(feature = "deserialize", derive(serde::Deserialize))]
+pub struct VertexBufferMapping {
+    /// Shader location associated with this buffer
+    pub id: u32,
+    /// Size of the structure in bytes
+    pub stride: u32,
+    /// True if the buffer is indexed by vertex, false if indexed
+    /// by instance.
+    pub indexed_by_vertex: bool,
+    /// Vec of the attributes within the structure
+    pub attributes: Vec<AttributeMapping>,
+}
+
 /// A subset of options that are meant to be changed per pipeline.
 #[derive(Debug, Default, Clone)]
 #[cfg_attr(feature = "serialize", derive(serde::Serialize))]
@@ -234,6 +341,17 @@ pub struct PipelineOptions {
     ///
     /// Enable this for vertex shaders with point primitive topologies.
     pub allow_and_force_point_size: bool,
+
+    /// If set, when generating the Metal vertex shader, transform it
+    /// to receive the vertex buffers, lengths, and vertex id as args,
+    /// and bounds-check the vertex id and use the index into the
+    /// vertex buffers to access attributes, rather than using Metal's
+    /// [[stage-in]] assembled attribute data.
+    pub vertex_pulling_transform: bool,
+
+    /// vertex_buffer_mappings are used during shader translation to
+    /// support vertex pulling.
+    pub vertex_buffer_mappings: Vec<VertexBufferMapping>,
 }
 
 impl Options {
--- a/naga/src/back/msl/writer.rs
+++ b/naga/src/back/msl/writer.rs
@@ -75,6 +75,14 @@ fn put_numeric_type(
     }
 }
 
+const fn scalar_is_int(scalar: crate::Scalar) -> bool {
+    use crate::ScalarKind::*;
+    match scalar.kind {
+        Sint | Uint | AbstractInt | Bool => true,
+        Float | AbstractFloat => false,
+    }
+}
+
 /// Prefix for cached clamped level-of-detail values for `ImageLoad` expressions.
 const CLAMPED_LOD_LOAD_PREFIX: &str = "clamped_lod_e";
 
@@ -85,6 +93,22 @@ struct TypeContext<'a> {
     access: crate::StorageAccess,
     binding: Option<&'a super::ResolvedBinding>,
     first_time: bool,
+}
+
+impl<'a> TypeContext<'a> {
+    fn scalar(&self) -> Option<crate::Scalar> {
+        let ty = &self.gctx.types[self.handle];
+        ty.inner.scalar()
+    }
+
+    fn vertex_input_dimension(&self) -> u32 {
+        let ty = &self.gctx.types[self.handle];
+        match ty.inner {
+            crate::TypeInner::Scalar(_) => 1,
+            crate::TypeInner::Vector { size, .. } => size as u32,
+            _ => unreachable!(),
+        }
+    }
 }
 
 impl<'a> Display for TypeContext<'a> {
@@ -3001,7 +3025,7 @@ impl<W: Write> Writer<W> {
                     // follow-up with any global resources used
                     let mut separate = !arguments.is_empty();
                     let fun_info = &context.expression.mod_info[function];
-                    let mut supports_array_length = false;
+                    let mut needs_buffer_sizes = false;
                     for (handle, var) in context.expression.module.global_variables.iter() {
                         if fun_info[handle].is_empty() {
                             continue;
@@ -3015,10 +3039,10 @@ impl<W: Write> Writer<W> {
                             }
                             write!(self.out, "{name}")?;
                         }
-                        supports_array_length |=
+                        needs_buffer_sizes |=
                             needs_array_length(var.ty, &context.expression.module.types);
                     }
-                    if supports_array_length {
+                    if needs_buffer_sizes {
                         if separate {
                             write!(self.out, ", ")?;
                         }
@@ -3417,11 +3441,20 @@ impl<W: Write> Writer<W> {
                 }
             }
 
-            if !indices.is_empty() {
+            let mut buffer_indices = vec![];
+            for vbm in &pipeline_options.vertex_buffer_mappings {
+                buffer_indices.push(vbm.id);
+            }
+
+            if !indices.is_empty() || !buffer_indices.is_empty() {
                 writeln!(self.out, "struct _mslBufferSizes {{")?;
 
                 for idx in indices {
                     writeln!(self.out, "{}uint size{};", back::INDENT, idx)?;
+                }
+
+                for idx in buffer_indices {
+                    writeln!(self.out, "{}uint buffer_size{};", back::INDENT, idx)?;
                 }
 
                 writeln!(self.out, "}};")?;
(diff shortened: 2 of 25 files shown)


cwfitzgerald:
Putting this in draft until it's ready to land

bradwerth:
> Putting this in draft until it's ready to land

Thanks, makes sense. I'll ensure the the generated msl is valid before seeking review again.

bradwerth:
This is passing checks and I'd like to get it in before starting in on the code-heavy data unpacking effort.

naga/src/back/msl/mod.rs
```diff
@@ -220,6 +220,40 @@ impl Default for Options {
...
+    /// Shader location associated with this attribute
+    pub id: u32,
+    /// Offset in bytes from start of vertex buffer structure
+    pub offset: u32,
+    /// Format code to help us unpack the attribute into the type
+    /// used by the shader. Codes correspond to a 0-based index of
+    /// <https://gpuweb.github.io/gpuweb/#enumdef-gpuvertexformat>.
+    /// The conversion process is described by
+    /// <https://gpuweb.github.io/gpuweb/#vertex-processing>.
+    pub format: u32,
```
teoxoy: I think we should duplicate `wgpu_types::VertexFormat` instead of using an index.

naga/src/back/msl/mod.rs
```diff
@@ -220,6 +220,40 @@ impl Default for Options {
...
 }
 
+/// A mapping of vertex buffers and their attributes to shader
+/// locations.
+#[derive(Debug, Default, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serialize", derive(serde::Serialize))]
+#[cfg_attr(feature = "deserialize", derive(serde::Deserialize))]
+pub struct AttributeMapping {
+    /// Shader location associated with this attribute
+    pub id: u32,
```
teoxoy:
```suggestion
    pub shader_location: u32,
```

naga/src/back/msl/writer.rs
```diff
@@ -3995,11 +3996,42 @@ impl<W: Write> Writer<W> {
...
+                    let buffer_stride = mapping.stride;
+                    let buffer_name = self.namer.call(format!("v_{buffer_id}_in").as_str());
+                    let buffer_ty = self.namer.call(format!("v_{buffer_id}_in_type").as_str());
+
+                    // Define a structure of bytes of the appropriate size.
+                    // When we access the attributes, we'll be unpacking these
+                    // bytes at some offset.
+                    writeln!(
+                        self.out,
+                        "struct {buffer_ty} {{ metal::uchar data[{buffer_stride}]; }};"
```
teoxoy, line 5024: The spec allows the stride to be 0 which means this array will have a size of 0.

I think we can get away without declaring this new type by calculating the offset into the struct manually (`vertex_index * stride`).

naga/src/back/msl/mod.rs
```diff
@@ -220,6 +220,40 @@ impl Default for Options {
...
+/// A description of a vertex buffer with all the information we
+/// need to address the attributes within it.
+#[derive(Debug, Default, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serialize", derive(serde::Serialize))]
+#[cfg_attr(feature = "deserialize", derive(serde::Deserialize))]
+pub struct VertexBufferMapping {
+    /// Shader location associated with this buffer
+    pub id: u32,
+    /// Number of elements in this buffer
+    pub count: u32,
```
teoxoy: The size of the vertex buffers won't be known when we generate the shader (at pipeline creation time).

We will have to pass the size of these buffers via another uniform buffer (constant buffer in metal).
We have this mechanism already in place. Here is some of the relevant code:

https://github.com/gfx-rs/wgpu/blob/8e15707631f1db5612824dea122ea8b6a4e9b271/naga/src/back/msl/writer.rs#L3240-L3259

teoxoy (changes requested):

naga/src/back/msl/writer.rs
bradwerth: Seems like the only reason why stride would be zero is if it is a specific array stride when the pipeline is using instance stride, or something. It seems like once we've gotten to a vertex_buffer_layout that we *must* have some kind of stride to make sense of what's happening. So I've put this check as an assert when copying the stride into our pipeline_options (all of this protected by the pref). If this is wrong, we'll sort it out later.

naga/src/back/msl/mod.rs
bradwerth: I'll do that later as part of the work to do real unpacking. It will be time consuming to do it here and without actually using it, I'll have no confidence that I have set it up correctly.

naga/src/back/msl/mod.rs
bradwerth: Thanks for this direction. I've updated the code to use the _buffer_sizes param.

wgpu-hal/src/metal/device.rs
```diff
@@ -830,8 +833,32 @@ impl crate::Device for super::Device {
...
+                            shader_location: attribute.shader_location,
+                            offset: attribute.offset as u32,
+                            format: attribute.format as u32,
+                        });
+                    }
+
+                    assert!(
+                        vbl.array_stride > 0,
+                        "Vertex pulling requires a non-zero buffer stride."
+                    );
```
teoxoy: We can't assert this here since it might be 0 and is not surrounded by the pref.

naga/src/back/msl/writer.rs
```diff
@@ -3801,11 +3803,12 @@ impl<W: Write> Writer<W> {
             );
 
             // Is any global variable used by this entry point dynamically sized?
-            let supports_array_length = module
-                .global_variables
-                .iter()
-                .filter(|&(handle, _)| !fun_info[handle].is_empty())
-                .any(|(_, var)| needs_array_length(var.ty, &module.types));
+            let supports_array_length = pipeline_options.experimental_vertex_pulling_transform
```
teoxoy: We should rename this var.

naga/src/back/msl/writer.rs
```diff
@@ -3288,7 +3288,9 @@ impl<W: Write> Writer<W> {
         {
             let mut indices = vec![];
             for (handle, var) in module.global_variables.iter() {
-                if needs_array_length(var.ty, &module.types) {
+                if pipeline_options.experimental_vertex_pulling_transform
```
teoxoy: I don't think this is right, we have to instead generate one field for each vertex buffer in `pipeline_options.vertex_buffer_mappings`.

teoxoy (dismissed):

naga/src/back/msl/writer.rs
bradwerth: Is that because the buffers are not necessarily global variables? I guess I need to understand better how this structure is supplied to the entry point at the point of calling.

teoxoy: They are represented as global variables at the IR level but `module.global_variables` won't contain the vertex buffers we bound for vertex pulling since we do that after we generate the IR (i.e. `global_variables` gets populated earlier by the WGSL front-end).

wgpu-hal/src/metal/device.rs
bradwerth: Removed, assert moved to `write_functions`.

naga/src/back/msl/writer.rs
bradwerth: Renamed to `needs_buffer_sizes`, here and in other places.

naga/src/back/msl/writer.rs
bradwerth: Updated to get the buffer indices from `pipeline_options.vertex_buffer_mappings`. Still need to figure out how to supply this data when the entry point is invoked.

bradwerth:
Things are changing fast as I add more material. Moved back to draft.

naga/src/back/msl/mod.rs
```diff
@@ -222,6 +222,151 @@ impl Default for Options {
...
+    /// Three signed ints (i32). `vec3<i32>` in shaders.
+    Sint32x3 = 28,
+    /// Four signed ints (i32). `vec4<i32>` in shaders.
+    Sint32x4 = 29,
+    /// Three unsigned 10-bit integers and one 2-bit integer, packed into a 32-bit integer (u32). [0, 1024] converted to float [0, 1] `vec4<f32>` in shaders.
+    #[cfg_attr(feature = "serde", serde(rename = "unorm10-10-10-2"))]
+    Unorm10_10_10_2 = 34,
+}
+
+impl From<u32> for VertexFormat {
```
teoxoy: We should do the mapping in wgpu hal so that we can ensure all variants of the original enum are handled.

naga/src/back/msl/mod.rs
```diff
@@ -222,6 +222,151 @@ impl Default for Options {
...
 }
 
+/// Corresponds to [WebGPU `GPUVertexFormat`](
+/// https://gpuweb.github.io/gpuweb/#enumdef-gpuvertexformat).
+#[repr(u32)]
+#[derive(Copy, Clone, Debug, Default, Hash, Eq, PartialEq)]
+#[cfg_attr(feature = "serialize", derive(serde::Serialize))]
+#[cfg_attr(feature = "deserialize", derive(serde::Deserialize))]
+pub enum VertexFormat {
+    #[default]
```
teoxoy: I don't think implementing default for this is appropriate since none of the variants seem like a good default.

naga/src/back/msl/writer.rs
```diff
@@ -3357,11 +3357,21 @@ impl<W: Write> Writer<W> {
...
+                }
+            }
+
+            if pipeline_options.experimental_vertex_pulling_transform {
+                // Only include the sizes of the vertex buffers if
+                // we're attempting vertex pulling, otherwise the
+                // the shaders will be getting a struct that is sized
+                // differently than expected.
+                for vbm in &pipeline_options.vertex_buffer_mappings {
+                    indices.insert(vbm.id);
```
teoxoy: These will surely collide. The first loop will insert global variable handle indices and the second inserts vertex buffer ids. We should consider splitting the sizes struct/buffer if we can't find a nice way to reuse it.

naga/src/back/msl/writer.rs
```diff
@@ -3720,6 +3730,734 @@ impl<W: Write> Writer<W> {
...
+                                self.out,
+                                "{}return metal::half4(as_type<metal::half>(b0 << 8 | b1), \
+                                                       as_type<metal::half>(b2 << 8 | b3), \
+                                                       as_type<metal::half>(b4 << 8 | b5), \
+                                                       as_type<metal::half>(b6 << 8 | b7));",
+                                back::INDENT
+                            )?;
+                            writeln!(self.out, "}}")?;
+                            (name, 8)
+                        }
```
teoxoy: I think the return type of all unpacking functions need to be 32-bit numbers (or vectors of 32-bit numbers).

From: https://gpuweb.github.io/gpuweb/#vertex-processing & https://gpuweb.github.io/gpuweb/wgsl/#channel-formats

WGSL without the `f16` extension (which we haven't implemented) only supports 32-bit numbers.

naga/src/back/msl/writer.rs
```diff
@@ -3720,6 +3730,734 @@ impl<W: Write> Writer<W> {
...
+                            )?;
+                            writeln!(
+                                self.out,
+                                "{}return unpack_unorm10a2_to_float(b0 << 24 | b1 << 16 | b2 << 8 | b3);",
+                                back::INDENT
+                            )?;
+                            writeln!(self.out, "}}")?;
+                            (name, 4)
+                        }
+                    };
```
teoxoy: All functions seem to do big-endian unpacking but the data in the buffers is meant to be unpacked as little-endian. This is covered in https://gpuweb.github.io/gpuweb/wgsl/#internal-value-layout (not very easy to find).

naga/src/back/msl/writer.rs
```diff
@@ -3720,6 +3730,734 @@ impl<W: Write> Writer<W> {
...
+                                              uint b6, \
+                                              uint b7, \
+                                              uint b8, \
+                                              uint b9, \
+                                              uint b10, \
+                                              uint b11, \
+                                              uint b12, \
+                                              uint b13, \
+                                              uint b14, \
+                                              uint b15) {{"
```
teoxoy: I was thinking it would be nicer to pass these as an array but I don't know how callers can get a sub-array from the initial array in MSL.

teoxoy (commented):
This is some good progress!
Left some more feedback.

naga/src/back/msl/writer.rs
bradwerth: But at this point, all the types and math are being done in MSL, using MSL types. wgsl types shouldn't matter anymore.

naga/src/back/msl/mod.rs
bradwerth: Default is gone. Not sure why I thought it was necessary.

naga/src/back/msl/writer.rs
bradwerth: Fixed by adding a new structure to carry the VertexBufferMapping into the StageInfo structure, and the calculated sizes of the buffers into the CommandState.

naga/src/back/msl/writer.rs
bradwerth: Fixed.

naga/src/back/msl/writer.rs
bradwerth: I don't see how to do this.

naga/src/back/msl/writer.rs
bradwerth: Re-open if I'm missing something, but hopefully any problems here will be revealed by tests.

bradwerth:
Getting a test in here is made more tricky by the limited exposure of the `PipelineOptions` object. That object is filled out in `load_shader`. Ideally we need to pass some configuration in the call to `create_render_pipeline` through some to-be-determined mechanism (modifying `RenderPipelineDescriptor`?). I don't have a solution yet.

naga/src/back/msl/mod.rs
bradwerth: I'm not clear on what this would entail, but I'm happy to do it with a bit more guidance on what a correct solution would look like.

naga/src/back/msl/writer.rs
bradwerth: I see the problem now, sorry, yes. Since Naga is generating all the MSL, and it only generates int, uint and float types (and vector variants), there's no value in these unpacking functions returning anything else. It also causes errors with Metal validation. For example, you can't assign a char4 to an int4. I'll fix this.

naga/src/back/msl/mod.rs
teoxoy: Moving it somewhere in `wgpu-hal/src/metal` would allow us to match directly on `wgpu_types::VertexFormat` instead of a `u32` making things less error-prone.

We can't have it be an `impl From<wgpu_types::VertexFormat> for naga::back::msl::VertexFormat` due to rust's "orphan rule" but we can just make it a function instead.

wgpu-hal/src/metal/device.rs
```diff
@@ -661,11 +664,12 @@ impl crate::Device for super::Device {
...
-            // handle the sizes buffer assignment and shader overrides
-            if info.sizes_count != 0 {
-                info.sizes_buffer = Some(info.counters.buffers);
-                info.counters.buffers += 1;
-            }
+            // Always set aside space for the sizes_buffer, which will
+            // be required if we have variable-length buffers, or to
+            // support vertex pulling.
+            info.sizes_buffer = Some(info.counters.buffers);
+            info.counters.buffers += 1;
```
teoxoy: Maybe we should instead initialize `sizes_count` like so:

```rust
sizes_count: if stage == naga::ShaderStage::Vertex { 1 } else { 0 },
```

to avoid reserving space for the `sizes_buffer` for the other stages.

We also seem to effectively use `sizes_count` as a `bool`; maybe we should change it to a `bool`.

naga/src/back/msl/writer.rs
```diff
@@ -3720,6 +3729,746 @@ impl<W: Write> Writer<W> {
...
+                    attributes: &vbm.attributes,
+                });
+
+                // Iterate the attributes and generate needed conversion functions.
+                for attribute in &vbm.attributes {
+                    if unpacking_functions.contains_key(&attribute.format) {
+                        continue;
+                    }
+
+                    let (name, byte_count) = match attribute.format {
```
teoxoy: This match is quite indented, could we pull it into its own function?

naga/src/back/msl/writer.rs
```diff
@@ -3720,6 +3729,746 @@ impl<W: Write> Writer<W> {
...
+                                self.out,
+                                "metal::float2 {name}(metal::ushort b0, \
+                                                      metal::ushort b1, \
+                                                      metal::ushort b2, \
+                                                      metal::ushort b3) {{"
+                            )?;
+                            writeln!(
+                                self.out,
+                                "{}return metal::float2(float(b1 << 8 | b0) / 65535.0f, \
+                                                        float(b3 << 8 | b1) / 65535.0f);",
```
teoxoy:
```suggestion
                                                        float(b3 << 8 | b2) / 65535.0f);",
```

naga/src/back/msl/writer.rs
teoxoy: Did you have a chance to test this? As far as I know the stride passed to hal can be 0.

https://github.com/gfx-rs/wgpu/blob/77a83fb0dd9f2295f25e99a850b9a031738925c3/wgpu-hal/src/metal/device.rs#L967-L983

(https://github.com/gfx-rs/wgpu/pull/3538)

naga/src/back/msl/writer.rs
bradwerth: Ugh, thanks for the careful read. Thanks for catching that.

bradwerth: Thanks for finding this tricky typo. I really appreciate it.

naga/src/back/msl/writer.rs
bradwerth: I put in a stride calculation in `load_shader` to cover this case.

jimblandy:
I'm happy with Teo's reviewing this.

naga/tests/out/msl/vertex-pulling-transform.msl
```diff
@@ -0,0 +1,65 @@
...
+    metal::float2 texcoord [[user(loc1), center_perspective]];
+};
+struct vb_1_type { metal::uchar data[36]; };
+vertex render_vertexOutput render_vertex(
+  uint v_existing_id [[vertex_id]]
+, constant metal::float4x4& mvp_matrix [[user(fake0)]]
+, const device vb_1_type* vb_1_in [[buffer(1)]]
+, constant _mslBufferSizes& _buffer_sizes [[user(fake0)]]
+) {
+    if (v_existing_id >= (_buffer_sizes.buffer_size1 / 36)) { return {}; }
```
teoxoy: I don't think we should immediately return here with a zero-initialized vertex output struct.

Point 2.4.2. in the ["process vertices" algorithm](https://gpuweb.github.io/gpuweb/#abstract-opdef-process-vertices) says:

> If this results in an out-of-bounds access, the resulting value is determined according to WGSL’s [invalid memory reference](https://gpuweb.github.io/gpuweb/wgsl/#invalid-memory-reference) behavior.

That link doesn't directly lead to the answer but there is more info below in https://gpuweb.github.io/gpuweb/wgsl/#invalid-load.

The body of the vertex shader must be executed even if the attributes were OOB.

I also filed https://github.com/gpuweb/gpuweb/issues/4668 to update the spec allowing us to treat all attributes from a vertex buffer as being OOB if at least one is so that we can still have one `if` per vertex buffer.

So, I think to get the desired behavior we can first zero-initialize the `VertexInput` struct then, if we are in-bounds for reading the vertex data we read it and set the fields of `VertexInput`.

naga/tests/out/msl/vertex-pulling-transform.msl
```diff
@@ -0,0 +1,65 @@
...
+    const vb_1_type vb_1_elem = vb_1_in[v_existing_id];
+    const metal::float4 position_1 = unpackFloat32x4_(vb_1_elem.data[0], vb_1_elem.data[1], vb_1_elem.data[2], vb_1_elem.data[3], vb_1_elem.data[4], vb_1_elem.data[5], vb_1_elem.data[6], vb_1_elem.data[7], vb_1_elem.data[8], vb_1_elem.data[9], vb_1_elem.data[10], vb_1_elem.data[11], vb_1_elem.data[12], vb_1_elem.data[13], vb_1_elem.data[14], vb_1_elem.data[15]);
+    const metal::float3 normal_1 = unpackFloat32x3_(vb_1_elem.data[16], vb_1_elem.data[17], vb_1_elem.data[18], vb_1_elem.data[19], vb_1_elem.data[20], vb_1_elem.data[21], vb_1_elem.data[22], vb_1_elem.data[23], vb_1_elem.data[24], vb_1_elem.data[25], vb_1_elem.data[26], vb_1_elem.data[27]);
+    const metal::float2 texcoord = unpackFloat32x2_(vb_1_elem.data[28], vb_1_elem.data[29], vb_1_elem.data[30], vb_1_elem.data[31], vb_1_elem.data[32], vb_1_elem.data[33], vb_1_elem.data[34], vb_1_elem.data[35]);
```
teoxoy: In this case the assignments are correct since the fields with the `@location` attribute have the same type as the return types of the unpacking functions; however that might not always be the case (they are allowed to vary in the nr of components).

Point 2.4.6. in the ["process vertices" algorithm](https://gpuweb.github.io/gpuweb/#abstract-opdef-process-vertices) has the details on how to do the conversions.

The first example in the ["Vertex Formats" section](https://gpuweb.github.io/gpuweb/#vertex-formats) shows this behavior in action.

teoxoy (changes requested):
I double-checked the behavior required by the spec and have some more feedback.

naga/tests/out/msl/vertex-pulling-transform.msl
bradwerth: The `VertexInput` structure is built from the attributes, so I'll zero-initialize all the attributes, set them if indices are in boudns, then let `VertexInput` be built from the attributes as usual.

naga/src/back/msl/writer.rs
```diff
@@ -4426,6 +5370,144 @@ impl<W: Write> Writer<W> {
...
+                            continue;
+                        }
+                        let am = am_option.unwrap();
+                        let attribute_ty_name = &am.ty_name;
+                        let attribute_name = &am.name;
+                        let attribute_zero_value = &am.zero_value;
+
+                        writeln!(
+                            self.out,
+                            "{}{attribute_ty_name} {attribute_name} = {attribute_zero_value};",
```
teoxoy:
```suggestion
                            "{}{attribute_ty_name} {attribute_name} = {{}};",
```

We could instead rely on the built-in zero-init behavior of `= {}`. We use this in other parts of the MSL backend.

naga/src/back/msl/writer.rs
```diff
@@ -4426,6 +5370,144 @@ impl<W: Write> Writer<W> {
...
+                    };
+                    let separator = if is_first_comparison { "" } else { " && " };
+                    write!(
+                        self.out,
+                        "{separator}{index_name} < (_buffer_sizes.buffer_size{idx} / {stride})"
+                    )?;
+                    is_first_comparison = false;
+                }
+
+                writeln!(self.out, ") {{")?;
```
teoxoy: We should have one check per vertex buffer since even with https://github.com/gpuweb/gpuweb/issues/4668 addressed, the OOB access will only "spread" to attributes coming from the same vertex buffer as the attribute that caused the OOB access (not to other attributes from other vertex buffers).

It seems appropriate to also add a 2nd vertex buffer for the snapshot test.

naga/src/back/mod.rs
```diff
@@ -241,6 +241,15 @@ const fn vector_size_str(size: crate::VectorSize) -> &'static str {
     }
 }
 
+/// Helper function that returns the u32 corresponding to the [`VectorSize`](crate::VectorSize)
+const fn vector_size_u32(size: crate::VectorSize) -> u32 {
```
teoxoy: `VectorSize` is `#[repr(u8)]`, so we can always do `size as u32` instead.

naga/src/back/msl/writer.rs
```diff
@@ -87,6 +104,62 @@ struct TypeContext<'a> {
...
+    }
+
+    fn scalar(&self) -> Option<crate::Scalar> {
+        let ty = &self.gctx.types[self.handle];
+        match ty.inner {
+            crate::TypeInner::Scalar(scalar) | crate::TypeInner::Atomic(scalar) => Some(scalar),
+            crate::TypeInner::Vector { scalar, .. } => Some(scalar),
+            crate::TypeInner::Matrix { .. } => Some(crate::Scalar::F32),
+            _ => None,
+        }
```
teoxoy:
```suggestion
        ty.inner.scalar()
```

naga/src/back/msl/writer.rs
```diff
@@ -87,6 +104,62 @@ struct TypeContext<'a> {
...
+    fn dimension(&self) -> u32 {
+        let ty = &self.gctx.types[self.handle];
+        match ty.inner {
+            crate::TypeInner::Vector { size, .. } => crate::back::vector_size_u32(size),
+            crate::TypeInner::Matrix { columns, rows, .. } => {
+                crate::back::vector_size_u32(columns) * crate::back::vector_size_u32(rows)
+            }
+            _ => 1,
+        }
+    }
```
teoxoy:
```suggestion
    fn vertex_input_dimension(&self) -> u32 {
        let ty = &self.gctx.types[self.handle];
        match ty.inner {
            crate::TypeInner::Scalar(_) => 1,
            crate::TypeInner::Vector { size, .. } => size as u32,
            _ => unreachable!(),
        }
    }
```

I think it's worth making this function more specialized/constrained since normally the type system is strict and we'd never treat scalars as vectors of size 1.

teoxoy (approved):
Looks great, let's land it!

PR #5265, opened 2024-02-18
Docs surface creation no longer unsafe

hackaugusto:
**Connections**
None

**Description**
Small changes to the docs

**Testing**
None

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu/src/macros.rs
+++ b/wgpu/src/macros.rs
@@ -78,7 +78,16 @@ macro_rules! include_spirv_raw {
     };
 }
 
-/// Macro to load a WGSL module statically.
+/// Load WGSL source code from a file at compile time.
+///
+/// The loaded path is relative to the path of the file containing the macro call, in the same way
+/// as [`include_str!`] operates.
+///
+/// ```ignore
+/// fn main() {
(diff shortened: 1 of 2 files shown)


wgpu/src/macros.rs
```diff
@@ -78,7 +78,14 @@ macro_rules! include_spirv_raw {
     };
 }
 
-/// Macro to load a WGSL module statically.
+/// Load WGSL source code from a file at compile time, the loaded path is relative to the source
+/// path.
```
kpreid: I think it'd be useful to mention the `std` behavior that this depends on, since that should have the most precise explanation. Also, "the source path" is a bit ambiguous if you don't already know what it means.

```suggestion
/// Load WGSL source code from a file at compile time.

/// The loaded path is relative to the path of the file containing the macro call,
/// in the same way as [`include_str!`] operates.
```

hackaugusto: updated :)

Wumpf (approved):
neat, thank you!

hackaugusto:
it seems there is a flaky text in the windows build. I can't restart the CI though.

Issue #5247, opened 2024-02-13
Plan for When `Features` overflows `u64`


Wumpf:
We just ran out of bits in the u64 feature flags in
* #5188 

Using 128bit integers might be problematic on the web and is generally not scalable going forward. We should instead try to use `enumset`

See also [discussion in Matrix](https://matrix.to/#/!FZyQrssSlHEZqrYcOb:matrix.org/$fVc-TvibKjMJT1mqkgS7iMH-cN8UzU0-ITIotZ8lHQY?via=matrix.org&via=mozilla.org&via=envs.net)

nical:
From a glance at the code it looks like `enumset` uses `u128` if you need more than 64 bits, and then arrays of `u64` if you need more than 128 bits. So we'd probably have the same ffi issues with u128 we are currently running into in gecko.

nical:
Due to the issue with `u128` and ffi my suggestion is to split into the standard flags and native-only ones into two types. We are pretty safe with 64 bits for the standard flags for a long while I think and they are the ones Gecko relies on in its FFI. If it doesn't break wgpu-native we can consider non-ffi-safe solutons for native-only flags in the future but splitting them off already gives us 30 or so bits to spare so we are good for a while too.

Wumpf:
sounds good! better to not take another dependency

cwfitzgerald:
I would still prefer to use `enumset` if we can - more specifically I really don't want to split the features, this is making the users bare the brunt of a minor internal issue. Many people's renderers have a single slot for features, one for limits, and this would force them to take on another slot for no gain on their end.

grovesNL:
I think it's likely that we'll eventually run out of space for `u128` too. Maybe we could consider moving these into a big feature flags struct (could be directly exposed to FFI)? We could still have convenience methods to get roughly the same ergonomics as the current setup. Or we could expose a struct publicly and internally use something else if we want to keep the representation compact.

nical:
I just tried to see if I could alter `enumset` to produce something that would work over ffi ( I don't think getting it to use arrays of u64 would be difficult), unfortunately just depending on `enumset` causes cbindgen to crash. Here is the quick and dirty experiment if anyone wants to play around: https://github.com/nical/ffi_flags_test

[...]

PR #5268, opened 2024-02-19
Make Features fit in 64 bits again

nical:
**Description**

`u128` is not safe to pass in ffi, this causes issues in Gecko's ffi glue and blocks Firefox from updating to a newer wgpu-core revision.

There's #5247 on file about using the `enumset` crate but the latter also seems to use u128 when the number of bits required is between 64 and 128 so in the interest of unblocking gecko, this PR packs the used bits into `u64`, which is a simple transformation since we still have a fair amount of unused bits.
The PR also reorders the standard features so that they appear in the same order as in the spec.

To get more bits in the future I think that we should split webgpu and native features into separate types. That will give us a fair amount of extra bits on each side and if some day the native-only bits are full again we can consider having them use something that isn't ffi-safe (if that doesn't cause issues with wgpu-native). without causing as much trouble on the gecko side. At this rate it is going to take a very long time for the webgpu flags to overflow 64 bits if it ever happens and gecko needs them to be ffi-safe.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -246,7 +246,7 @@ bitflags::bitflags! {
     #[repr(transparent)]
     #[derive(Default)]
     #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
-    pub struct Features: u128 { // TODO(https://github.com/gfx-rs/wgpu/issues/5247): consider making this an `enumset`
+    pub struct Features: u64 {
         //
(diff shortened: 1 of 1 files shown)


cwfitzgerald:
The failure is the GBV timeout issue

cwfitzgerald (approved):

PR #5269, opened 2024-02-19
Disable GPU based validation in tests by default but keep it on CI

nical:
**Description**

With gpu based validation, `cargo xtask test` immediately fills all of my RAM and makes the desktop unusable (forcing me to hard-reboot the computer) on the two computers I have handy (they have 32 and 64 GB of RAM).

I don't know if it's a mesa or AMD specific thing but it's so disruptive that I think that we need to take a more careful approach to enabling the the advanced validation stuff. This PR turns it off by default for people running tests locally (can be re-enabled using the `WGPU_GPU_BASED_VALIDATION` environment variable), but leaves it enabled on CI for now.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -527,6 +527,8 @@ jobs:
         run: |
           set -e
 
+          export WGPU_WGPU_GPU_BASED_VALIDATION=1
+
           cargo xtask test --llvm-cov
 
       - name: check naga snapshots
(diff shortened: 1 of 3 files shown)


cwfitzgerald:
Let's pull it everywhere - #5154 is having issues with dx12 GBV, and it's already been taking way too long.

The failures you're getting now is because we're now missing the additional validation errors when GBV is enabled.

cwfitzgerald (changes requested):
Clicking button

cwfitzgerald:
Tried a different strategy in #5271

PR #5271, opened 2024-02-19
Remove GPU based validation in Tests

cwfitzgerald:
**Connections**

Supersedes #5269 

**Description**

This disabled GBV everywhere. This should fix:
- DX12 tests taking multiple minutes
- DX12 Crashes in upcoming PR in CI #5154
- Vulkan using massive amounts of memory.

This leaves synchronization validation enabled, as that's still of significant value to us.

It also cleans up some of the vulkan initialization code.

**Testing**

Tested locally on MVK. @nical can you verify this fixes the memory explosion?


nical (approved):
Confirmed it fixes the memory explosion for me.

PR #5273, opened 2024-02-20
Add feature flags in hal to panic when running into some types of errors

nical:
**Description**

This is for debugging purposes and does not have to be long term. Right now hal backends don't report much information about what failed and why. This makes debugging some issues a bit difficult.

Having panics with with more context (the stacktrace and HResult code for example) would pretty dramatically speed up the process of debugging the d3d12 issues I'm investigating right now.

This PR adds three feature flags in hal, each causing panics for their respective category of error close to where the error is produced. 
 - `oom_panic`
 - `device_lost_panic`
 - `internal_error_panic`

I implemented the panics in vulkan and d3d12 but did not add the corresponding glue for the metal and GL backends.

Longer term we could address this better by producing more descriptive errors in hal.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-hal/Cargo.toml
+++ b/wgpu-hal/Cargo.toml
@@ -63,6 +63,18 @@ dxc_shader_compiler = ["hassle-rs"]
 renderdoc = ["libloading", "renderdoc-sys"]
 fragile-send-sync-non-atomic-wasm = ["wgt/fragile-send-sync-non-atomic-wasm"]
 link = ["metal/link"]
+# Panic when running into an out-of-memory error (for debugging purposes).
+#
+# Only affects the d3d12 and vulkan backends.
+oom_panic = []
+# Panic when running into a device lost error (for debugging purposes).
+# Only affects the d3d12 and vulkan backends.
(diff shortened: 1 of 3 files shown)


nical:
This is similar to https://github.com/gfx-rs/wgpu/pull/5099 with fewer backends but more useful info. Turns out wiring device lost errors will take some time, I need to debug some stuff now and also need to catch other types of errors.

Wumpf:
makes me wonder if the feature shouldn't be instead about having a callstack and description string on those errors enums, seems that's what's actually needed?

nical:
In the long run yes, it would be better to have the error types contain all of the relevant information, and be able to surface filter/process that data accordingly in gecko. In the short term there is a rather pressing issue that I am not able to reproduce and crashes are the easy way for the reproducer(s) to get back to me with a crash report containing all of the info I need (because of how firefox is set up).

cwfitzgerald (approved):

PR #5274, opened 2024-02-20
Validate that the view dimension of a multisampled texture binding is 2d

nical:
**Connections**

CTS test: `webgpu:api,validation,createBindGroupLayout:multisampled_validation:viewDimension=*`
Spec: https://gpuweb.github.io/gpuweb/#dom-gpudevice-createbindgrouplayout

**Description**

The spec says that the view dimension of a multi-sampled texture must be 2d. This implements the validation accordingly.

**Testing**

Covered by the CTS.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


Wumpf (approved):
looking good!

PR #5280, opened 2024-02-21
fix doc link

HexoKnight:
Fixes a link in the docs for `Instance::create_surface` that pointed to `Instance::create_surface` instead of `Instance::create_surface_unsafe`.


cwfitzgerald (approved):
Nice!

PR #5282, opened 2024-02-21
fix: always check buffer clear `offset` for OOB

ErichDonGubler:
Recommended review experience is commit-by-commit.

**Connections**

This fix was prompted by Firefox's [bug 1881065](https://bugzilla.mozilla.org/show_bug.cgi?id=1881065).

**Description**

Fuzz testing in Firefox encountered crashes for calls of `Global::command_encoder_clear_buffer` where:

* `offset` is greater than `buffer.size`, but…
* `size` is `None`.

Oops! We should _always_ check this (i.e., even when `size` is `None`), because we have no guarantee that `offset` and the fallback value of `size` is in bounds. 😅 So, we change validation here to unconditionally compute `size` and run checks we previously gated behind `if let Some(size) = size { … }`.

For convenience, the spec. link for this method: <https://gpuweb.github.io/gpuweb/#dom-gpucommandencoder-clearbuffer>

**Testing**

The following new tests have been added, and can be tested as arguments to `cargo xtask test`:

* `wgpu_test::buffer::clear_offset_outside_resource_bounds`
* `wgpu_test::buffer::clear_offset_plus_size_outside_u64_bounds`

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/tests/tests/buffer.rs
+++ b/tests/tests/buffer.rs
@@ -353,3 +353,35 @@ static CLEAR_OFFSET_OUTSIDE_RESOURCE_BOUNDS: GpuTestConfiguration = GpuTestConfi
             "Clear of 20..20 would end up overrunning the bounds of the buffer of size 16"
         ));
     });
+
+#[gpu_test]
+static CLEAR_OFFSET_PLUS_SIZE_OUTSIDE_U64_BOUNDS: GpuTestConfiguration =
+    GpuTestConfiguration::new()
+        .parameters(TestParameters::default())
+        .run_sync(|ctx| {
+            let buffer = ctx.device.create_buffer(&wgpu::BufferDescriptor {
+                label: None,
+                size: 16, // unimportant for this test
+                usage: wgpu::BufferUsages::COPY_DST,
+                mapped_at_creation: false,
+            });
+
+            let max_valid_offset = u64::MAX - (u64::MAX % wgpu::COPY_BUFFER_ALIGNMENT);
+            let smallest_aligned_invalid_size = wgpu::COPY_BUFFER_ALIGNMENT;
+
+            ctx.device.push_error_scope(wgpu::ErrorFilter::Validation);
+            ctx.device
+                .create_command_encoder(&Default::default())
+                .clear_buffer(
+                    &buffer,
+                    max_valid_offset,
+                    Some(smallest_aligned_invalid_size),
+                );
+            let err_msg = pollster::block_on(ctx.device.pop_error_scope())
+                .unwrap()
+                .to_string();
+            assert!(err_msg.contains(concat!(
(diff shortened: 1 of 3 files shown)


tests/tests/buffer.rs
```diff
@@ -326,3 +326,62 @@ static MINIMUM_BUFFER_BINDING_SIZE_DISPATCH: GpuTestConfiguration = GpuTestConfi
             let _ = encoder.finish();
         });
     });
+
+#[gpu_test]
+static CLEAR_OFFSET_OUTSIDE_RESOURCE_BOUNDS: GpuTestConfiguration = GpuTestConfiguration::new()
+    .parameters(TestParameters::default().test_features_limits())
```
cwfitzgerald: We don't need test_features_limits() here, just default

tests/tests/buffer.rs
```diff
@@ -326,3 +326,62 @@ static MINIMUM_BUFFER_BINDING_SIZE_DISPATCH: GpuTestConfiguration = GpuTestConfi
...
+        ctx.device.push_error_scope(wgpu::ErrorFilter::Validation);
+        ctx.device
+            .create_command_encoder(&Default::default())
+            .clear_buffer(&buffer, out_of_bounds, None);
+        let err_msg = pollster::block_on(ctx.device.pop_error_scope())
+            .unwrap()
+            .to_string();
+        assert!(err_msg.contains(
+            "Clear of 20..20 would end up overrunning the bounds of the buffer of size 16"
+        ));
```
cwfitzgerald, line 354: Could we work this into the `fail` function in some way?

cwfitzgerald (dismissed):
Love to see it!

tests/tests/buffer.rs
ErichDonGubler: Ah, okay. Fixed with 2e6138e169a2b1fa718402f4cd88bb3d57701c76 and 17c8f32cad0a1ce34ac2e978b5124fc9a1915775.

tests/tests/buffer.rs
ErichDonGubler: Ah, I actually intended to bring this up (though as follow-up work) and had forgotten about it. 😅 It seems like narrowing the specificity of a validation error would give a lot more confidence in a test being what we want. Do you consider this blocking for this PR? 🤔

While I'm thinking about it: I think adding a third argument is a good solution, unless we want to smith a one-off `Option`-alike with more descriptive variant names. I think confronting test authors with "please consider adding a message" is a good thing; I'd prefer to avoid a builder for this one.

ErichDonGubler: Filed a separate PR: https://github.com/gfx-rs/wgpu/pull/5282

nical (approved):

PR #5290, opened 2024-02-22
Add `expected_msg_substring` arg. to `wgpu_test::fail`

ErichDonGubler:
**Connections**

Motivated by discussion in (and based on code from) another PR: https://github.com/gfx-rs/wgpu/pull/5282#discussion_r1498481164

- [x] After merge: file `good first issue` for converting more tests to use specific validation error messages: https://github.com/gfx-rs/wgpu/issues/5727

**Description**

Until now, we accept _any_ validation failure inside a `callback`  passed to `wgpu_test::fail`. This is a _very_ broad error condition, and may allow errors for the wrong reason. Now, we'd like to tighten up the failures that we actually expect from these tests. Also, this would let us factor out some error scope boilerplate introduced for the tests in [#5282](https://github.com/gfx-rs/wgpu/pull/5282).

**Testing**

This is a refactor of tests to make them more narrow, and they still pass. ✅

**Checklist**

- [x] Run `cargo fmt`.
- [x] ~~Add change to `CHANGELOG.md`. See simple instructions inside file.~~ Seems like none should be necessary here?

### DIFF
--- a/tests/src/lib.rs
+++ b/tests/src/lib.rs
@@ -27,10 +27,28 @@ pub use run::{execute_test, TestingContext};
 pub use wgpu_macros::gpu_test;
 
 /// Run some code in an error scope and assert that validation fails.
-pub fn fail<T>(device: &wgpu::Device, callback: impl FnOnce() -> T) -> T {
+pub fn fail<T>(
+    device: &wgpu::Device,
+    callback: impl FnOnce() -> T,
+    expected_msg_substring: Option<&'static str>,
+) -> T {
     device.push_error_scope(wgpu::ErrorFilter::Validation);
     let result = callback();
-    assert!(pollster::block_on(device.pop_error_scope()).is_some());
+    let validation_error = pollster::block_on(device.pop_error_scope())
+        .expect("expected validation error in callback, but no validation error was emitted");
+    if let Some(expected_msg_substring) = expected_msg_substring {
+        let lowered_expected = expected_msg_substring.to_lowercase();
+        let lowered_actual = validation_error.to_string().to_lowercase();
+        assert!(
+            lowered_actual.contains(&lowered_expected),
+            concat!(
+                "expected validation error case-insensitively containing {:?}, ",
+                "but it was not present in actual error message:\n{:?}"
+            ),
+            expected_msg_substring,
+            validation_error
+        );
+    }
 
     result
 }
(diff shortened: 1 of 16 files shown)


tests/src/lib.rs
```diff
@@ -46,7 +64,7 @@ pub fn valid<T>(device: &wgpu::Device, callback: impl FnOnce() -> T) -> T {
 /// provided `should_fail` boolean.
 pub fn fail_if<T>(device: &wgpu::Device, should_fail: bool, callback: impl FnOnce() -> T) -> T {
     if should_fail {
-        fail(device, callback)
+        fail(device, callback, None)
```
ErichDonGubler: - [x] todo: Let's expose this argument in `fail_if`. I don't see a reason not to migrate its callers now.

tests/tests/buffer.rs
```diff
@@ -217,15 +217,19 @@ static MINIMUM_BUFFER_BINDING_SIZE_LAYOUT: GpuTestConfiguration = GpuTestConfigu
...
+            || {
+                ctx.device
+                    .create_compute_pipeline(&wgpu::ComputePipelineDescriptor {
+                        label: None,
+                        layout: Some(&pipeline_layout),
+                        module: &shader_module,
+                        entry_point: "main",
+                    });
+            },
+            None,
```
ErichDonGubler, line 233: - [x] todo: I'd like to ask for others' thoughts on taking the time to narrow the accepted validation error message in tests with this PR. I think a separate change to update each call to `wgpu_test::fail` to actually specify an expected validation message and maybe even force `expected_msg_substring` to be required programatically would be good, but it doesn't _have_ to be now (and I don't consider it a personal priority.

CC @jimblandy, @cwfitzgerald, @teoxoy, @nical.

ErichDonGubler: I'm gonna punt forcing concrete validation error messages for now. Maybe a `good first issue`? 🙂

Wumpf (approved):
nice! I did something similar here, but in a less intrusive way (because the pr shouldn't also drag in such a large refactor)

https://github.com/gfx-rs/wgpu/pull/5671/files#diff-aeefe6b3d3f8b1b5248d6bb4c2888cf2d6802dbf82515f7b02b6a89a1400228fR35

(will happily move my stuff to this new variant once we get the stuff that comes before it merged :/)

ErichDonGubler:
Rebasing required migrating new tests from <https://github.com/gfx-rs/wgpu/pull/5715>, resulting in 7d40e5b8f and 7bccffb9b.

PR #5291, opened 2024-02-23
[naga msl-out] Tolerate use of `IndexMap::remove`, for Firefox.

jimblandy:
`IndexMap::remove` was deprecated in `indexmap` 2.2.0, with `swap_remove` as the appropriate replacement for its use by Naga's `NamedExpressions` type. Unfortunately, Firefox [patches] `indexmap` to always use version 1, so Naga can't use any `indexmap` 2 features until that patch is removed.

I believe Firefox's patch is there to avoid vendoring two copies of `indexmap`. Once all of Firefox's uses move to indexmap 2, Firefox's patch will be removed, and we can remove this commit and update Naga's `IndexMap` usage.

[patches]: https://searchfox.org/mozilla-central/rev/da49863c3d6f34038d00f5ba701b9a2ad9cbadba/Cargo.toml#156-157

### DIFF
--- a/naga/src/back/msl/writer.rs
+++ b/naga/src/back/msl/writer.rs
@@ -3048,6 +3048,13 @@ impl<W: Write> Writer<W> {
         for statement in statements {
             if let crate::Statement::Emit(ref range) = *statement {
                 for handle in range.clone() {
+                    // `IndexMap::remove` was deprecated in `indexmap` 2.2.0,
+                    // with `swap_remove` as the appropriate replacement here.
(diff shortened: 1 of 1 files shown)


cwfitzgerald:
Counterintuitively, [swap_remove does exist in indexmap 1.9.3](https://docs.rs/indexmap/1.9.3/indexmap/map/struct.IndexMap.html#method.swap_remove), so we can just follow the deprecation message.

cwfitzgerald (changes requested):
.

PR #5300, opened 2024-02-25
Fix Integer Clamp

cwfitzgerald:
**Connections**

Discovered during my work on #3159.

**Description**

`clamp(3, 2, 1)` for integers is explicitly defined by the [wgsl spec](https://gpuweb.github.io/gpuweb/wgsl/#clamp) as returning 1. In practice this happens on OpenGL on nvidia cards and on vulkan on AMD cards.

**Testing**

Have a PR coming up with tests for int clamp, as well as other shader functions

### DIFF
--- a/naga/src/back/glsl/mod.rs
+++ b/naga/src/back/glsl/mod.rs
@@ -3138,7 +3138,29 @@ impl<'a, W: Write> Writer<'a, W> {
                     Mf::Abs => "abs",
                     Mf::Min => "min",
                     Mf::Max => "max",
-                    Mf::Clamp => "clamp",
+                    Mf::Clamp => {
+                        let scalar_kind = ctx
+                            .resolve_type(arg, &self.module.types)
+                            .scalar_kind()
+                            .unwrap();
+                        match scalar_kind {
+                            crate::ScalarKind::Float => "clamp",
+                            // Clamp is undefined if min > max. In practice this means it can use a median-of-three
+                            // instruction to determine the value. This is fine according to the WGSL spec for float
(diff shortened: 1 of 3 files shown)


naga/src/back/glsl/mod.rs
```diff
@@ -3138,7 +3138,29 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                        match scalar_kind {
+                            crate::ScalarKind::Float => "clamp",
+                            // Clamp is undefined if min > max. In practice this means it can use a median-of-three
+                            // instruction to determine the value. This is fine according to the WGSL spec for float
+                            // clamp, but integer clamp _must_ use min-max. As such we write out min/max.
+                            _ => {
+                                write!(self.out, "min(max(")?;
+                                self.write_expr(arg, ctx)?;
+                                write!(self.out, ", ")?;
+                                self.write_expr(arg1.unwrap(), ctx)?;
```
Wumpf, line 3155: Naga beginner question: can we know for sure at this point that `arg1` and `arg2` are `Some`, i.e. have we already made sure that the input is valid?

Wumpf (approved):
huh, interesting detail. Nothing is ever easy!

Wumpf:
warrants a changelog entry imho

naga/src/back/glsl/mod.rs
cwfitzgerald: Yeah, the Naga validator should have made sure that the argument count is correct

teoxoy (approved):

PR #4190, opened 2023-09-30
Subgroup Operations

exrook:
**Checklist**

- [x] Run `cargo clippy`.
- [x] Run `cargo clippy --target wasm32-unknown-unknown` if applicable.
- [x] Add change to CHANGELOG.md. See simple instructions inside file.

**Connections**
merged with https://github.com/gfx-rs/naga/pull/2523
Closes: #4428

**Description**
Allows wgsl shaders to perform subgroup operations

Adds wgpu features:
 -  `SUBGROUP_COMPUTE`
 -  `SUBGROUP_FRAGMENT`
 -  `SUBGROUP_VERTEX`

Each feature enables `BASIC`, `VOTE`, `ARITHMETIC`, `SHUFFLE`, and `SHUFFLE_RELATIVE` operations to be used in their respective stage. See https://github.com/gfx-rs/wgpu/pull/4190#issuecomment-1777752121 for a short discussion of the portability of this feature set.

Adds new naga capability `SUBGROUP`, required for a shader to use subgroup builtin functions or parameters.

Adds new naga validator settings `subgroup_operations` and `subgroup_stages` determining which sets of the below subgroup operations are valid, and which stages they are valid in.

## `BASIC` operations:
```
# Performs a control and memory barrier across all invocations in the subgroup
subgroupBarrier()
```

## `VOTE` operations:
```
subgroupAll(bool) -> bool
subgroupAny(bool) -> bool
```

## `ARITHMETIC` operations:
```
# Operations on scalars and vectors of f32, i32, u32:
# Computes a single result value using values from all active lanes
subgroupAdd(value) -> value
subgroupMul(value) -> value
subgroupMin(value) -> value
subgroupMax(value) -> value
# Operations on scalars and vectors of i32, u32:
# Computes a single result value using values from all active lanes
subgroupAnd(value) -> value
subgroupOr(value) -> value
subgroupXor(value) -> value
# Computes a prefix scan across all active lanes
subgroupPrefixInclusiveAdd(value) -> value
subgroupPrefixInclusiveMul(value) -> value
subgroupPrefixExclusiveAdd(value) -> value
subgroupPrefixExclusiveMul(value) -> value
```

## `BALLOT` operations:
```
# Computes a result using a single bit from every active lane
subgroupBallot() -> vec4<u32>
subgroupBallot(bool) -> vec4<u32>
# Operations on scalars and vectors of f32, i32, u32:
# Reads a value from the first active lane into every other active lane
subgroupBroadcastFirst(value) -> value
[...]

PR #5301, opened 2024-02-25
Subgroup Operations

### LINKED ISSUE #4428: Support subgroup/wave operations
Extremely useful for things like parallel reductions, single pass downscalers, etc.

https://github.com/gpuweb/gpuweb/issues/3950

https://www.khronos.org/blog/vulkan-subgroup-tutorial
https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/hlsl-shader-model-6-0-features-for-direct3d-12
https://developer.apple.com/documentation/metal/compute_passes/creating_threads_and_threadgroups#2928931

Lichtso:
Additional PR for code review as requested in https://github.com/gfx-rs/wgpu/pull/4190#issuecomment-1962736887.

Resolves https://github.com/gfx-rs/wgpu/issues/4428


wgpu-hal/src/dx12/adapter.rs
```diff
@@ -127,6 +127,11 @@ impl super::Adapter {
             )
         });
 
+        // If we don't have dxc, we reduce the max to 5.1
+        if dxc_container.is_none() {
+            shader_model_support.HighestShaderModel = d3d12_ty::D3D_SHADER_MODEL_5_1;
```
atlv24, line 132: good addition, reminder to myself to update #5154 to check `shader_model_support.HighestShaderModel >= d3d12_ty::D3D_SHADER_MODEL_6_0` instead of whether dxc is present for 64bit support

atlv24 (commented):
looks good to me

vgarleanu:
Awesome work! Is there an ETA for when this will be merged? How can we speed things up?
cc: @rodolphito

wgpu-hal/src/dx12/adapter.rs
```diff
@@ -298,6 +303,11 @@ impl super::Adapter {
         // float32-filterable should always be available on d3d12
         features.set(wgt::Features::FLOAT32_FILTERABLE, true);
 
+        features.set(
+            wgt::Features::SUBGROUP_COMPUTE | wgt::Features::SUBGROUP_FRAGMENT,
+            shader_model_support.HighestShaderModel >= d3d12_ty::D3D_SHADER_MODEL_6_0,
+        );
```
teoxoy: We also need to check for `WaveOps` in `D3D12_FEATURE_DATA_D3D12_OPTIONS1`.

wgpu-hal/src/metal/adapter.rs
```diff
@@ -807,6 +807,12 @@ impl super::PrivateCapabilities {
                 None
             },
             timestamp_query_support,
+            supports_simd_scoped_operations: family_check
+                && (device.supports_family(MTLGPUFamily::Metal3)
+                    || device.supports_family(MTLGPUFamily::Mac2)
+                    || device.supports_family(MTLGPUFamily::Apple7)
+                    || device.supports_family(MTLGPUFamily::Apple8)
+                    || device.supports_family(MTLGPUFamily::Apple9)),
```
```suggestion
                    || device.supports_family(MTLGPUFamily::Apple7)),
```

All `Apple8+` support `Apple7`.

wgpu-hal/src/metal/adapter.rs
```diff
@@ -888,6 +894,10 @@ impl super::PrivateCapabilities {
         features.set(F::RG11B10UFLOAT_RENDERABLE, self.format_rg11b10_all);
         features.set(F::SHADER_UNUSED_VERTEX_OUTPUT, true);
 
+        if self.supports_simd_scoped_operations {
+            features.insert(F::SUBGROUP_COMPUTE | F::SUBGROUP_FRAGMENT | F::SUBGROUP_VERTEX);
+        }
```
teoxoy: Could you provide a reference for the availability of subgroup ops in the vertex stage? I looked at the MSL spec and it only mentions kernel & fragment.

teoxoy:
I started to look at it today.

wgpu-types/src/lib.rs
```diff
@@ -873,6 +873,32 @@ bitflags::bitflags! {
...
+        const SUBGROUP_COMPUTE = 1 << 55;
+        /// Allows fragment shaders to use the subgroup operation built-ins
+        ///
+        /// Supported Platforms:
+        /// - Vulkan
+        /// - DX12
+        /// - Metal
+        ///
+        /// This is a native only feature.
+        const SUBGROUP_FRAGMENT = 1 << 56;
```
teoxoy: I think we can combine these 2 into a `SUBGROUPS` feature since there are very few devices that only support them in the compute stage.

wgpu-types/src/lib.rs
```diff
@@ -873,6 +873,32 @@ bitflags::bitflags! {
...
+        /// Allows vertex shaders to use the subgroup operation built-ins
+        ///
+        /// Supported Platforms:
+        /// - Vulkan
+        /// - Metal
+        ///
+        /// This is a native only feature.
+        const SUBGROUP_VERTEX = 1 << 57;
```
teoxoy: I think this is only available in Vulkan. Is it valuable to expose it?

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -648,6 +678,8 @@ pub struct PhysicalDeviceCapabilities {
     descriptor_indexing: Option<vk::PhysicalDeviceDescriptorIndexingPropertiesEXT>,
     acceleration_structure: Option<vk::PhysicalDeviceAccelerationStructurePropertiesKHR>,
     driver: Option<vk::PhysicalDeviceDriverPropertiesKHR>,
+    subgroup: Option<vk::PhysicalDeviceSubgroupProperties>,
+    /// The effective driver api version supported by the physical device.
```
```suggestion
```

Seems this came back in one of the merges.

tests/tests/subgroup_operations/shader.wgsl
```diff
@@ -0,0 +1,173 @@
...
+    @builtin(num_subgroups) num_subgroups: u32,
+    @builtin(subgroup_id) subgroup_id: u32,
+    @builtin(subgroup_size) subgroup_size: u32,
+    @builtin(subgroup_invocation_id) subgroup_invocation_id: u32,
+) {
+    var passed = 0u;
+    var expected: u32;
+
+    var mask = 1u << 0u;
+    passed |= mask * u32(num_subgroups == 128u / subgroup_size);
```
cwfitzgerald: Lets introduce a function that properly does this dance:

```
fn add_result_to_mask(mask: ptr<function, u32>, index: u32, value: bool) {
   (*mask) |= u32(value) << index;
}
```

I think this would clean up this code a bunch

cwfitzgerald (changes requested):

naga/src/lib.rs
```diff
@@ -1284,6 +1289,46 @@ pub enum SwizzleComponent {
...
+    ShuffleXor(Handle<Expression>),
+}
+
+#[derive(Clone, Copy, Debug, Hash, Eq, Ord, PartialEq, PartialOrd)]
+#[cfg_attr(feature = "serialize", derive(Serialize))]
+#[cfg_attr(feature = "deserialize", derive(Deserialize))]
+#[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
+pub enum SubgroupOperation {
+    All = 0,
+    Any = 1,
```
teoxoy, line 1310: I think these should get their own enum as they are not arithmetic and don't need to be paired with a `CollectiveOperation`.

We should also add `Elect` to the new enum.

naga/src/lib.rs
```diff
@@ -1284,6 +1289,46 @@ pub enum SwizzleComponent {
...
+#[cfg_attr(feature = "serialize", derive(Serialize))]
+#[cfg_attr(feature = "deserialize", derive(Deserialize))]
+#[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
+pub enum GatherMode {
+    BroadcastFirst,
+    Broadcast(Handle<Expression>),
+    Shuffle(Handle<Expression>),
+    ShuffleDown(Handle<Expression>),
+    ShuffleUp(Handle<Expression>),
+    ShuffleXor(Handle<Expression>),
```
teoxoy: We should add some docs saying what the expressions here are (id/mask/delta).

naga/src/front/wgsl/parse/conv.rs
```diff
@@ -252,3 +257,26 @@ pub fn map_conservative_depth(
...
+        "subgroupMul" => (sg::Mul, co::Reduce),
+        "subgroupMin" => (sg::Min, co::Reduce),
+        "subgroupMax" => (sg::Max, co::Reduce),
+        "subgroupAnd" => (sg::And, co::Reduce),
+        "subgroupOr" => (sg::Or, co::Reduce),
+        "subgroupXor" => (sg::Xor, co::Reduce),
+        "subgroupPrefixExclusiveAdd" => (sg::Add, co::ExclusiveScan),
+        "subgroupPrefixExclusiveMul" => (sg::Mul, co::ExclusiveScan),
+        "subgroupPrefixInclusiveAdd" => (sg::Add, co::InclusiveScan),
+        "subgroupPrefixInclusiveMul" => (sg::Mul, co::InclusiveScan),
```
teoxoy:
```suggestion
        "subgroupSum" => (sg::Add, co::Reduce),
        "subgroupProduct" => (sg::Mul, co::Reduce),
        "subgroupMin" => (sg::Min, co::Reduce),
        "subgroupMax" => (sg::Max, co::Reduce),
        "subgroupAnd" => (sg::And, co::Reduce),
        "subgroupOr" => (sg::Or, co::Reduce),
        "subgroupXor" => (sg::Xor, co::Reduce),
        "subgroupExclusiveSum" => (sg::Add, co::ExclusiveScan),
        "subgroupExclusiveProduct" => (sg::Mul, co::ExclusiveScan),
        "subgroupInclusiveSum" => (sg::Add, co::InclusiveScan),
        "subgroupInclusiveProduct" => (sg::Mul, co::InclusiveScan),
```

We should keep the naming in sync with the [upstream proposal](https://github.com/gpuweb/gpuweb/blob/main/proposals/subgroups.md).

naga/src/lib.rs
```diff
@@ -1284,6 +1289,46 @@ pub enum SwizzleComponent {
...
+
+#[repr(u8)]
+#[derive(Clone, Copy, Debug, Hash, Eq, Ord, PartialEq, PartialOrd)]
+#[cfg_attr(feature = "serialize", derive(Serialize))]
+#[cfg_attr(feature = "deserialize", derive(Deserialize))]
+#[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
+pub enum CollectiveOperation {
+    Reduce = 0,
+    InclusiveScan = 1,
+    ExclusiveScan = 2,
```
teoxoy, line 1327: Does this need to be `repr(u8)`? Same question for `SubgroupOperation`.

naga/src/back/hlsl/writer.rs
```diff
@@ -1213,6 +1227,53 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
+                                ep.workgroup_size[0] * ep.workgroup_size[1] * ep.workgroup_size[2]
+                            )?,
+                            crate::BuiltIn::SubgroupId => {
+                                writeln!(
+                                    self.out,
+                                    "(__local_invocation_id.z * {}u + __local_invocation_id.y * {}u + __local_invocation_id.x) / WaveGetLaneCount();",
+                                    ep.workgroup_size[0] * ep.workgroup_size[1],
+                                    ep.workgroup_size[1],
+                                )?;
+                            }
```
teoxoy: The calculations for `num_subgroups` & `subgroup_id` make sense to me but do you know if they are guaranteed by DX12?

naga/src/back/hlsl/writer.rs
```diff
@@ -1213,6 +1227,53 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
+                                self.out,
+                                "({}u + WaveGetLaneCount() - 1u) / WaveGetLaneCount();",
+                                ep.workgroup_size[0] * ep.workgroup_size[1] * ep.workgroup_size[2]
+                            )?,
+                            crate::BuiltIn::SubgroupId => {
+                                writeln!(
+                                    self.out,
+                                    "(__local_invocation_id.z * {}u + __local_invocation_id.y * {}u + __local_invocation_id.x) / WaveGetLaneCount();",
+                                    ep.workgroup_size[0] * ep.workgroup_size[1],
+                                    ep.workgroup_size[1],
```
teoxoy: Using `SV_GroupIndex` would be more straightforward.

naga/src/back/hlsl/writer.rs
```diff
@@ -1263,14 +1324,20 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
         func_ctx: &back::FunctionCtx,
         module: &Module,
     ) -> bool {
-        self.options.zero_initialize_workgroup_memory
+        func.arguments.iter().any(|arg| {
+            matches!(
+                arg.binding,
+                Some(crate::Binding::BuiltIn(crate::BuiltIn::NumSubgroups))
+            )
+        }) || (self.options.zero_initialize_workgroup_memory
```
teoxoy: Returning true here doesn't just emit `uint3 __local_invocation_id : SV_GroupThreadID`, it also injects an if and a barrier.
Please move this somewhere more appropriate. We can also then emit a `SV_GroupIndex` instead.

naga/src/back/hlsl/writer.rs
```diff
@@ -1213,6 +1227,53 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
+            let ep = &module.entry_points[ep_index as usize];
+            for (index, arg) in func.arguments.iter().enumerate() {
+                if let Some(crate::Binding::BuiltIn(builtin)) = arg.binding {
+                    if matches!(
+                        builtin,
+                        crate::BuiltIn::SubgroupSize
+                            | crate::BuiltIn::SubgroupInvocationId
+                            | crate::BuiltIn::NumSubgroups
+                            | crate::BuiltIn::SubgroupId
+                    ) {
```
teoxoy: The builtins could also be part of a struct, the current code doesn't handle this case.

naga/src/valid/function.rs
```diff
@@ -399,6 +414,106 @@ impl super::Validator {
...
+                if { &context.types[ty].inner == argument_inner } => {}
+            _ => {
+                return Err(SubgroupError::ResultTypeMismatch(result)
+                    .with_span_handle(result, context.expressions)
+                    .into_other())
+            }
+        }
+        Ok(())
+    }
+    fn validate_subgroup_broadcast(
```
teoxoy:
```suggestion
    fn validate_subgroup_gather(
```

naga/src/valid/function.rs
```diff
@@ -399,6 +414,106 @@ impl super::Validator {
...
+                    .into_other());
+            }
+        };
+
+        use crate::ScalarKind as sk;
+        use crate::SubgroupOperation as sg;
+        match (scalar.kind, *op) {
+            (sk::Bool, sg::All | sg::Any) if is_scalar => {}
+            (sk::Sint | sk::Uint | sk::Float, sg::Add | sg::Mul | sg::Min | sg::Max) => {}
+            (sk::Sint | sk::Uint | sk::Bool, sg::And | sg::Or | sg::Xor) => {}
```
teoxoy: MSL and HLSL don't have logical and/or/xor ops.
This would have to be gated by a capability (can be done later by interested parties).

naga/src/valid/function.rs
```diff
@@ -613,8 +728,27 @@ impl super::Validator {
                     stages &= super::ShaderStages::FRAGMENT;
                     finished = true;
                 }
-                S::Barrier(_) => {
+                S::Barrier(barrier) => {
                     stages &= super::ShaderStages::COMPUTE;
+                    if barrier.contains(crate::Barrier::SUB_GROUP) {
```
teoxoy, line 756: HLSL doesn't have a subgroup barrier.
This would have to be gated by a capability (and a wgpu feature).

naga/src/valid/function.rs
```diff
@@ -399,6 +414,106 @@ impl super::Validator {
         }
         Ok(())
     }
+    fn validate_subgroup_operation(
+        &mut self,
+        op: &crate::SubgroupOperation,
+        _collective_op: &crate::CollectiveOperation,
```
teoxoy: We should validate the `SubgroupOperation-CollectiveOperation` matrix.

- `Reduce` is available on all backends for all operations.
- `ExclusiveScan` & `InclusiveScan` is available on all backends only for sum and product.

Additional functionality needs to be gated behind capabilities (can be done later by interested parties).

naga/src/back/hlsl/writer.rs
```diff
@@ -2000,6 +2067,129 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
+                        write!(self.out, "WavePrefixProduct(")?
+                    }
+                    (crate::CollectiveOperation::InclusiveScan, crate::SubgroupOperation::Add) => {
+                        self.write_expr(module, argument, func_ctx)?;
+                        write!(self.out, " + WavePrefixSum(")?;
+                    }
+                    (crate::CollectiveOperation::InclusiveScan, crate::SubgroupOperation::Mul) => {
+                        self.write_expr(module, argument, func_ctx)?;
+                        write!(self.out, " * WavePrefixProduct(")?;
+                    }
```
teoxoy, line 2148: Since we write `argument` twice, the expr needs to be baked. See `fn update_expressions_to_bake()`.

teoxoy (changes requested):
Looks good overall!

Would you also be interested in adding the quad ops listed in the [upstream proposal](https://github.com/gpuweb/gpuweb/blob/main/proposals/subgroups.md)?

naga/src/front/wgsl/parse/conv.rs
Lichtso: WGSL has always copied the GLSL naming scheme. Thus I think we should stick with "Add" and "Mul" and [change the upstream proposal](https://github.com/gpuweb/gpuweb/pull/4528) instead. I removed the "Prefix" though as that is not present in GLSL.

naga/src/lib.rs
Lichtso: `And`, `Or`, `Xor` are not arithmetic either.
But they are all paired with a `CollectiveOperation`.

In fact:
- `All` is the same as `And` with `CollectiveOperation::Reduce` and scalar boolean operands.
- `Any` is the same as `Or` with `CollectiveOperation::Reduce` and scalar boolean operands.

`Elect` does not fit at all as it takes no operands. It is more like `GatherMode::BroadcastFirst` as it can be emulated as `broadcastFirst(subgroup_invocation_id) == subgroup_invocation_id`.

naga/src/back/hlsl/writer.rs
Lichtso: The SPIR-V spec (which can be targeted by both GLSL and HLSL) says it is guaranteed that:
`LocalInvocationIndex = LocalInvocationID.z * WorkGroupSize.x * WorkGroupSize.y + LocalInvocationID.y * WorkGroupSize.x + LocalInvocationID.x;`

And if `WorkGroupSize.x % SubgroupSize == 0` then it is also guaranteed that:
`LocalInvocationIndex = SubgroupLocalInvocationId + SubgroupId * SubgroupSize`

naga/src/valid/function.rs
raphlinus: Putting in a vote for having a subgroup barrier when available, as it's useful for sorting (particular the WLMS algorithm).

JMS55: I'm not the author, but maybe worth doing this in a separate PR to avoid scope creep. There's a lot of useful stuff unlocked by this even without subgroup barriers.

naga/src/back/hlsl/writer.rs
```diff
@@ -1144,9 +1218,20 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
             back::FunctionType::EntryPoint(ep_index) => {
                 if let Some(ref ep_input) = self.entry_point_io[ep_index as usize].input {
                     write!(self.out, "{} {}", ep_input.ty_name, ep_input.arg_name,)?;
+                    if func.arguments.iter().any(|arg| {
+                        matches!(
+                            arg.binding,
+                            Some(crate::Binding::BuiltIn(crate::BuiltIn::SubgroupId))
```
teoxoy: The binding could also be present as a field of a struct passed to the entry point.

We could add a field to `EntryPointBinding` which will get set in `write_interface_struct` if we see a `BuiltIn::SubgroupId`.

naga/src/back/hlsl/conv.rs
```diff
@@ -179,6 +179,10 @@ impl crate::BuiltIn {
             // to this field will get replaced with references to `SPECIAL_CBUF_VAR`
             // in `Writer::write_expr`.
             Self::NumWorkGroups => "SV_GroupID",
+            Self::SubgroupSize
+            | Self::SubgroupInvocationId
+            | Self::NumSubgroups
+            | Self::SubgroupId => return Err(Error::Unimplemented(format!("builtin {self:?}"))),
```
teoxoy:
```suggestion
            // These builtins map to functions
            Self::SubgroupSize
            | Self::SubgroupInvocationId
            | Self::NumSubgroups
            | Self::SubgroupId => unreachable!(),
```

naga/src/back/hlsl/writer.rs
```diff
@@ -1160,13 +1245,11 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
 
-                        if let Some(ref binding) = arg.binding {
-                            self.write_semantic(binding, Some((stage, Io::Input)))?;
-                        }
+                        self.write_semantic(&arg.binding, Some((stage, Io::Input)))?;
                     }
 
                     if need_workgroup_variables_initialization {
-                        if !func.arguments.is_empty() {
+                        if arg_num > 0 {
```
teoxoy:
```suggestion
                        if !func.arguments.is_empty() {
```

naga/src/back/hlsl/writer.rs
```diff
@@ -563,7 +591,13 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
     ) -> Result<EntryPointInterface, Error> {
         Ok(EntryPointInterface {
-            input: if !func.arguments.is_empty() && stage == ShaderStage::Fragment {
+            input: if !func.arguments.is_empty()
+                && (stage == ShaderStage::Fragment
+                    || func
+                        .arguments
+                        .iter()
+                        .any(|arg| is_subgroup_builtin_binding(&arg.binding)))
+            {
```
teoxoy, line 616: We should update the comment on this fn.

teoxoy (changes requested):

naga/src/valid/function.rs
teoxoy: This PR already adds support for the subgroup barrier.
It's just that they have to be gated behind another capability (and wgpu feature) since HLSL doesn't have it.

naga/src/back/hlsl/writer.rs
teoxoy: I think the problem is that it's not guaranteed that the value returned by `WaveGetLaneCount()` is static across the whole workgroup.

I think it can only be guaranteed by using the [`WaveSize` attribute](https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_WaveSize.html).

Lichtso: Ah, ok. As far as I am aware, heterogeneous subgroup sizes happen when a compute shader dispatch knows the grid / workitem counts and then partially deactivates the subgroups around the edges of the grid if the workitem count is not divisible by the workgroup size. But DirectX only knows the workgroup count at compute shader dispatch, unlike e.g. OpenCL. So it shouldn't be a problem for the wgpu backend anyway. Additionally the [DirectX wiki](https://github.com/Microsoft/DirectXShaderCompiler/wiki/Wave-Intrinsics) says that `WaveGetLaneCount()` is unaffected by partial deactivation.

naga/src/back/hlsl/writer.rs
```diff
@@ -442,17 +470,30 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
             }
             self.write_type(module, m.ty)?;
             write!(self.out, " {}", &m.name)?;
-            if let Some(ref binding) = m.binding {
-                self.write_semantic(binding, Some(shader_stage))?;
-            }
+            self.write_semantic(&m.binding, Some(shader_stage))?;
             writeln!(self.out, ";")?;
         }
+        if func.arguments.iter().any(|arg| {
```
teoxoy:
```suggestion
        if members.iter().any(|arg| {
```

Good idea to put this directly in the flattened struct! But this should be `members` since those are the params that have been flattened.

naga/src/back/hlsl/conv.rs
teoxoy: Could you also change this from the "unimplemented" error to `unreachable!()`? Since they have been implemented just in a different way.

naga/src/back/hlsl/writer.rs
teoxoy: My worry is that instead of launching same sized subgroups with inactive lanes, different sized subgroups would be launched (still not necessarily with all lanes active but at least less inactive ones).

The only indication of the former is the DXC wiki mentioning

> Returns the number of lanes in a wave on this architecture.

teoxoy: https://github.com/microsoft/DirectXShaderCompiler/issues/4731 & https://github.com/microsoft/hlsl-specs/issues/16#issuecomment-1636547485 do make it sound like all subgroups should have the same size. But since "driver-defined constants" are not yet a thing and `WaveGetLaneCount()` is just a function it would be conceivable that some driver would launch different sized subgroups.

I'm not sure what the best solution would be here. Maybe we should at least have a test making sure `subgroup_size` returns the same value for a bunch of odd variations of workgroup and dispatch sizes.

teoxoy: Looking around some more, everyone seems to rely on `WaveGetLaneCount()` returning a static value.

But Jesse Natalie says:

> > then you can get index of wave inside group by dividing group thread id by wave width
>
> No, there's no such guaranteed mapping

_from https://discord.com/channels/590611987420020747/996417435374714920/1084919934125088868_

Then says how VkOn12 works around this.

I looked at mesa's dozen driver (aka VkOn12) and the polyfill looks similar to the one they shared on discord with the only difference being that it also just divides the local invocation index by the subgroup size but only for 1D workgroup sizes.

Here is the permalink:
https://gitlab.freedesktop.org/mesa/mesa/-/blob/c2ec0111b21667d1db9d6ceae295b033f044ed87/src/microsoft/compiler/dxil_nir.c#L1937

Looking at the blame, it seems the fastpath [was added](https://gitlab.freedesktop.org/mesa/mesa/-/commit/4f56cede6deff561a8cfa3fe893cf085cb8b2f1c) because DOOM Eternal's shaders were relying on that property 😅.

Lichtso: So they implement ticked number drawing via atomic operations on workgroup memory.
Meaning their assignment of workitem-index in a workgroup to subgroup-index in a workgroup is chaotic / random.

I added a test to make sure the `subgroup_size` stays consistent within a workgroup.

teoxoy: > So they implement ticked number drawing via atomic operations on workgroup memory.
Meaning their assignment of workitem-index in a workgroup to subgroup-index in a workgroup is chaotic / random.

That explains the "corruption/flickering".

> I added a test to make sure the `subgroup_size` stays consistent within a workgroup.

Thanks, it looks good!

So given that games are relying on this behavior I guess our mapping is probably fine (at least for 1D workgroup sizes).

My last question would be: What about 2D/3D workgroup sizes? Do we need the more complex polyfill dozen is using?

Lichtso: Multidimensional cases are certainly tricky because there are many possible serializations and we don't know what the driver and the firmware on the GPU is going to use. Also, our tests currently only have 1D workgroups.

naga/src/back/pipeline_constants.rs
```diff
@@ -637,6 +639,31 @@ fn adjust_stmt(new_pos: &[Handle<Expression>], stmt: &mut Statement) {
...
+            ref mut result,
+            ..
+        } => {
+            adjust(argument);
+            adjust(result);
+        }
+        Statement::SubgroupGather {
+            ref mut argument,
+            ref mut result,
+            ..
```
teoxoy: `mode` has handles to expressions as well.

naga/src/back/hlsl/writer.rs
teoxoy: I think we should (at least for now) not allow using `SubgroupId` in entry-points with non-1D workgroup sizes. What do you think?

naga/tests/out/glsl/subgroup-operations-s.main.Compute.glsl
```diff
@@ -0,0 +1,58 @@
...
+    uint _e31 = subgroupShuffleDown(_e6, 1u);
+    uint _e32 = subgroupShuffleUp(_e6, 1u);
+    uint _e34 = subgroupShuffleXor(_e6, (_e5 - 1u));
+    return;
+}
+
+void main() {
+    uint num_subgroups = gl_NumSubgroups;
+    uint subgroup_id = gl_SubgroupID;
+    uint subgroup_size = gl_SubgroupSize;
```
raphlinus, line 50: This mapping is potentially incorrect unless the VK_EXT_subgroup_size_control extension is enabled and the VK_PIPELINE_SHADER_STAGE_CREATE_ALLOW_VARYING_SUBGROUP_SIZE_BIT_EXT flag is set. Both are on by default in Vulkan 1.3, but I'm pretty sure this code is intended to run on pre-1.3 versions.

I personally would recommend enabling the extension and turning on the flag on instance creation, and only advertising the subgroup extension when it is available. Failing that, there are workarounds which might have problems, such as dividing the workgroup size by the number of subgroups; this has been tested to work on Intel GPUs without the size control extension turned on.

Lichtso: Can you elaborate how the mapping is incorrect / not what is expected?

I understand subgroup size to mean the number of possible lanes, not the (potentially smaller) number of active lanes. That can instead be obtained by doing popcount on a subgroup ballot.

Also, I can only think of these scenarios is which a compute shader / kernel is launched with partially deactivated subgroups:
1. The grid size is not divisible by the subgroup size
2. The workgroup size is not divisible by the subgroup size
3. The workgroup size is smaller than the subgroup size
4. Register pressure is too high

Other than these I never saw an implementation do actually heterogeneous subgroup sizes (fully activated but different groupings) in the same kernel launch. E.g. the RDNA3 ISA:

> The shader supports both waves of 32 work-items ("wave32") and waves of 64 work-items ("wave64").
> Both wave sizes are supported for all operations, but shader programs must be compiled for and run as a particular wave size, regardless of how many work-items are active in any given wave.

raphlinus:
Another question about this PR. It seems to be missing the implementation of `subgroupMinSize` and `subgroupMaxSize` from the gpuweb proposal. Is that intentional? Is there a plan to add them later? These are also extremely difficult to get right, but on Vulkan with subgroup size control it is accurate and useful.

naga/tests/out/glsl/subgroup-operations-s.main.Compute.glsl
raphlinus: I wrote about this a bit in my [Prefix sum on Vulkan](https://raphlinus.github.io/gpu/2020/04/30/prefix-sum.html#subgroups-and-subgroup-size) blog post, and in followup issues.

My personal take on this is that modeling the subgroup size as always 32 on Intel, yet masking lanes so only 8 or 16 is active, is a polite fiction to work around a design limitation of Vulkan 1.1, fixed as of Vulkan 1.3, and should be rejected as a mistake, undesirable to ever expose to a WebGPU client. Writing code that is defensive against all possible subgroup masks is basically impossible. Doing popcount on a ballot is definitely another possible workaround for determining `subgroup_size`, and may be a better approach to the if subgroup size control is not available, but I think there's a risk it can't be recognized as a compile time constant.

Lichtso: Lets say the workgroup size is 15 and the hardware is running at a subgroup size of 8. But subgroups are partially deactivated from the start because of register pressure. The workitem in workgroup indices are given in hex or `#` if inactive.
```
[0 1 2 3 # # # # ]
        [4 5 6 7 # # # # ]
                [8 9 A B # # # # ]
                        [C D E # # # # # ]
```

So what do you expect `subgroup_size` to be in the given example?
1. `8` for all subgroups (which you reject as a mistake)
2. `4` for all subgroups
3. `4` for the first 3 subgroups and `3` for the last (which you could also get from `subgroupBallotBitcount`)

raphlinus: I expect 8 for all subgroups, and I don't agree that I classified that as a mistake - I want `subgroup_size` to reflect the hardware subgroup size, and in particular to match the Vulkan 1.3 SubgroupSize semantics.

I'm not sure where your picture comes from, I'm not aware of any case in which Intel will do that. And if the subgroup size is actually 4 (which may be the case for SwiftShader), then I'd expect a subgroup size of 4. Technically it would be allowed behavior with a subgroup size of 8 if the FULL_SUBGROUPS flag is not set, but to me that would be very surprising behavior from any actual GPU.

> If [subgroup size control is enabled] the SubgroupSize decorated variable will contain the subgroup size for each subgroup that gets dispatched. This value must be between [minSubgroupSize](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#limits-minSubgroupSize) and [maxSubgroupSize](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#limits-maxSubgroupSize) and must be uniform with [subgroup scope](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#shaders-scope-subgroup). The value may vary across a single draw call, and for fragment shaders may vary across a single primitive. In compute dispatches, SubgroupSize must be uniform with [command scope](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#shaders-scope-command).

So this clearly rules out your choice 3.

I believe the Vulkan spec provides sufficiently strong guarantees that the popcount of a ballot will match the subgroup size when FULL_SUBGROUPS is enabled. It will, I believe, empirically be true for compute shaders in practice, though I certainly approve of relying on the actual guarantee of the API in cases when the API has taken care to document its guarantees (as Vulkan does).

> Developers can also specify a new pipeline shader stage create flag that requires the implementation to have fully populated subgroups within local workgroups. This requires the workgroup size in the X dimension to be a multiple of the subgroup size.

Validation [02758](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPipelineShaderStageCreateInfo.html#VUID-VkPipelineShaderStageCreateInfo-flags-02758) further narrows this by saying that the workgroup X dimension must be a multiple of maxSubgroupSize when FULL_SUBGROUPS is set. (Note that this will be enforced for cooperative matrices, so it's worth looking ahead. Also note that life becomes more difficult for the programmer if subgroupMaxSize is not made available.)

Whether the application can request FULL_SUBGROUPS is a TODO in the current [gpuweb proposal](https://github.com/gpuweb/gpuweb/blob/main/proposals/subgroups.md). I would vote yes, because it seems to be needed if you want any kind of guarantee your code will work. The guarantee can reliably be offered on Vulkan; as the proposal states, other APIs are unclear.

There is *strong* value in having the subgroup size potentially available as a compile time constant; the robustly portable code for workgroup-wide prefix sum will have cases for 2, 3, and 4 levels for a workgroup size of 256, and you want those extra branches compiled out. It's an interesting empirical question whether the popcount of the ballot can get constant-folded, perhaps worth testing with microbenchmarks on different GPUs.

wgpu-hal/src/dx12/adapter.rs
```diff
@@ -377,6 +389,8 @@ impl super::Adapter {
                         .min(crate::MAX_VERTEX_BUFFERS as u32),
                     max_vertex_attributes: d3d12_ty::D3D12_IA_VERTEX_INPUT_RESOURCE_SLOT_COUNT,
                     max_vertex_buffer_array_stride: d3d12_ty::D3D12_SO_BUFFER_MAX_STRIDE_IN_BYTES,
+                    min_subgroup_size: features1.WaveLaneCountMin,
```
raphlinus: I've observed a reported value of 16 on Intel GPUs that can actually produce a subgroup size of 8, so best to hardcode this at 4.

The proposal's guidance is, I believe, incorrect. There may be some situations in which the reported values may be considered reliable (possibly in practice newer drivers, conjecturing those that support SM6.6, but I think empirical testing would be needed).

wgpu-hal/src/metal/adapter.rs
```diff
@@ -952,6 +960,8 @@ impl super::PrivateCapabilities {
                 max_vertex_buffers: self.max_vertex_buffers,
                 max_vertex_attributes: 31,
                 max_vertex_buffer_array_stride: base.max_vertex_buffer_array_stride,
+                min_subgroup_size: 4,
```
raphlinus, line 963: Sigh. All Apple Silicon devices have a fixed subgroup size of 32, though it's plausible that this may change in the future. I really wish there were a principled way to set this; it's too tempting to just sniff the device name.

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -987,6 +1022,14 @@ impl PhysicalDeviceProperties {
                 .min(crate::MAX_VERTEX_BUFFERS as u32),
             max_vertex_attributes: limits.max_vertex_input_attributes,
             max_vertex_buffer_array_stride: limits.max_vertex_input_binding_stride,
+            min_subgroup_size: self
+                .subgroup_size
+                .map(|subgroup_size| subgroup_size.min_subgroup_size)
+                .unwrap_or(0),
```
raphlinus, line 1050: I think these can be 4 and 128 if size control is not available (Vulkan 1.1 and friends).

wgpu-types/src/lib.rs
```diff
@@ -1136,6 +1160,11 @@ pub struct Limits {
     /// The maximum value for each dimension of a `ComputePass::dispatch(x, y, z)` operation.
     /// Defaults to 65535. Higher is "better".
     pub max_compute_workgroups_per_dimension: u32,
+
+    /// Minimal number of workitems in a subgroup. Lower is "better".
```
raphlinus: This is backwards from the limit comparison below. I think higher is better, and lower for max, as a narrower range of possibilities is easier for the programmer.

Also this should probably be "invocations" to match the spec, I don't see evidence of "workitems" being used elsewhere.

naga/tests/out/glsl/subgroup-operations-s.main.Compute.glsl
Lichtso: > I expect 8 for all subgroups, and I don't agree that I classified that as a mistake - I want subgroup_size to reflect the hardware subgroup size, and in particular to match the Vulkan 1.3 SubgroupSize semantics.

Sorry, I might have misinterpreted what you were referring to with:
> should be rejected as a mistake, undesirable to ever expose to a WebGPU client

> I'm not sure where your picture comes from, I'm not aware of any case in which Intel will do that.

The example was not specific to Intel, just a general example to tease out the possible different meanings of "subgroup size".

> SubgroupSize must be uniform with [command scope](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#shaders-scope-command). So this clearly rules out your choice 3.

You are right. But then I don't get what `VK_PIPELINE_SHADER_STAGE_CREATE_ALLOW_VARYING_SUBGROUP_SIZE_BIT` is about, if "SubgroupSize must be uniform with command scope".

> Validation [02758](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPipelineShaderStageCreateInfo.html#VUID-VkPipelineShaderStageCreateInfo-flags-02758) further narrows this by saying that the workgroup X dimension must be a multiple of maxSubgroupSize when FULL_SUBGROUPS is set. (Note that this will be enforced for cooperative matrices, so it's worth looking ahead. Also note that life becomes more difficult for the programmer if subgroupMaxSize is not made available.)

This does thread into the other discussion above about multi dimensional workgroup sizes and how the logical layout / mapping of workitems to subgroup would work in that case. We have not decided on that yet.

> There is strong value in having the subgroup size potentially available as a compile time constant

Agreed, cascades of subgroup operations are a hassle in general. But I don't see how we could achieve this across all backends.

wgpu-hal/src/metal/adapter.rs
Lichtso: You are forgetting about quad operations. According to the MSL spec:
> A quad-group is a SIMD-group with the thread execution width of 4.

Anyway, any suggested change here code wise?

naga/tests/out/glsl/subgroup-operations-s.main.Compute.glsl
raphlinus: "Varying" is I think confusing naming. For compute shaders, it means that SubgroupSize (gl_SubgroupSize) can be the the actual subgroup size chosen by the shader compiler, as opposed to the previous semantics which forced it to be equal to `subgroupSize` in [VkPhysicalDeviceSubgroupProperties](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceSubgroupProperties.html). That latter value is constant for the physical device, and cannot vary for different shader compilations.

> This does thread into the other discussion above about multi dimensional workgroup sizes and how the logical layout / mapping of workitems to subgroup would work in that case. We have not decided on that yet.

I agree that this is tricky. The Vulkan spec is clear on making no mapping guarantees - In Vulkan, when using subgroups it's best to rely on `subgroup_invocation_id` and use that for buffer indexing, etc, not relying on a specific mapping between that and LocalInvocationId. But then Vulkan provides SubgroupID, which is missing from HLSL. This is a TODO item in the proposal. In practice, for 1D dispatches, I think SubgroupInvocationID + SubgroupID * SubgroupSize will equal LocalInvocationID.x. But also note that this will go horribly wrong if you're on Intel and don't have subgroup size control - SubgroupSize will be 32 and so this will overflow your buffer.

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -936,6 +936,13 @@ impl PhysicalDeviceProperties {
             extensions.push(vk::KhrDrawIndirectCountFn::name());
         }
 
+        // Require `VK_EXT_subgroup_size_control` if the associated feature was requested
+        // Even though Vulkan 1.3 has promoted the extension to core, we must require the extension to avoid
```
Lichtso, line 941: Looks like this won't do and we have to use `PhysicalDeviceVulkan13Features` after all. @teoxoy Thoughts?

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -987,6 +1036,14 @@ impl PhysicalDeviceProperties {
...
             max_vertex_attributes: limits.max_vertex_input_attributes,
             max_vertex_buffer_array_stride: limits.max_vertex_input_binding_stride,
+            min_subgroup_size: self
+                .subgroup_size_control
+                .map(|subgroup_size| subgroup_size.min_subgroup_size)
+                .unwrap_or(4),
+            max_subgroup_size: self
+                .subgroup_size_control
+                .map(|subgroup_size| subgroup_size.max_subgroup_size)
+                .unwrap_or(128),
```
```suggestion
            min_subgroup_size: self
                .subgroup_size_control
                .map(|subgroup_size| subgroup_size.min_subgroup_size)
                .unwrap_or(0),
            max_subgroup_size: self
                .subgroup_size_control
                .map(|subgroup_size| subgroup_size.max_subgroup_size)
                .unwrap_or(0),
```

Should be 0 since we won't expose subgroup functionality without the extension (or vk 1.3).

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -773,6 +812,11 @@ pub struct PhysicalDeviceProperties {
     /// `VK_KHR_driver_properties` extension, promoted to Vulkan 1.2.
     driver: Option<vk::PhysicalDeviceDriverPropertiesKHR>,
 
+    /// Additional `vk::PhysicalDevice` properties from the
+    /// `VK_EXT_subgroup_size_control` extension, promoted to Vulkan 1.3.
+    subgroup: Option<vk::PhysicalDeviceSubgroupProperties>,
+    subgroup_size_control: Option<vk::PhysicalDeviceSubgroupSizeControlProperties>,
```
teoxoy, line 826: Looks like the doc comment is on the wrong field.

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -638,6 +652,31 @@ impl PhysicalDeviceFeatures {
...
+                        .supported_stages
+                        .contains(vk::ShaderStageFlags::COMPUTE | vk::ShaderStageFlags::FRAGMENT),
+                );
+                features.set(
+                    F::SUBGROUP_VERTEX,
+                    subgroup
+                        .supported_stages
+                        .contains(vk::ShaderStageFlags::VERTEX),
+                );
+                features.insert(F::SUBGROUP_BARRIER);
```
teoxoy, line 682: We should only advertise those features if `VK_EXT_subgroup_size_control` is present or we are on vk 1.3.

teoxoy (changes requested):

Lichtso:
I fiddled with the Vulkan backend since yesterday and it is still complaining that:

> Validation Error: Validation Error: [ VUID-VkPipelineShaderStageCreateInfo-flags-02784 ] | MessageID = 0xd65f4ec3 | vkCreateGraphicsPipelines(): pCreateInfos[0].pStages[0].flags includes VK_PIPELINE_SHADER_STAGE_CREATE_ALLOW_VARYING_SUBGROUP_SIZE_BIT_EXT, but the subgroupSizeControl feature was not enabled. The Vulkan spec states: If flags has the VK_PIPELINE_SHADER_STAGE_CREATE_ALLOW_VARYING_SUBGROUP_SIZE_BIT flag set, the subgroupSizeControl feature must be enabled (https://vulkan.lunarg.com/doc/view/1.3.268.0/linux/1.3-extensions/vkspec.html#VUID-VkPipelineShaderStageCreateInfo-flags-02784)

Even though I coupled the setting of `PipelineShaderStageCreateFlags` directly to the presence of the feature flag:
```
subgroup_size_control: phd_features
     .subgroup_size_control
     .map_or(false, |ext| ext.subgroup_size_control == vk::TRUE),

if self.shared.private_caps.subgroup_size_control {
     flags |= vk::PipelineShaderStageCreateFlags::ALLOW_VARYING_SUBGROUP_SIZE;
}
```

@teoxoy Ideas?

teoxoy:
Ah yes, we need to explicitly enable it when creating the device.

I think the only thing missing is actually pushing the struct.

https://github.com/gfx-rs/wgpu/blob/82dead09e492eed4adab73d8aa253a03645fbc29/wgpu-hal/src/vulkan/adapter.rs#L150

naga/tests/out/glsl/subgroup-operations-s.main.Compute.glsl
Lichtso: Makes sense now, thanks for the explanation!

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -773,6 +818,14 @@ pub struct PhysicalDeviceProperties {
     /// `VK_KHR_driver_properties` extension, promoted to Vulkan 1.2.
     driver: Option<vk::PhysicalDeviceDriverPropertiesKHR>,
 
+    /// Additional `vk::PhysicalDevice` properties from the
+    /// `VK_EXT_subgroup_size_control` extension, promoted to Vulkan 1.3.
```
```suggestion
    /// Additional `vk::PhysicalDevice` properties from Vulkan 1.1.
```

teoxoy:
I think the last remaining thing is https://github.com/gfx-rs/wgpu/pull/5301#discussion_r1556025602.

naga/src/valid/interface.rs
```diff
@@ -613,6 +615,17 @@ impl super::Validator {
...
+                && matches!(
+                    fa.binding,
+                    Some(crate::Binding::BuiltIn(crate::BuiltIn::SubgroupId))
+                        | Some(crate::Binding::BuiltIn(
+                            crate::BuiltIn::SubgroupInvocationId
+                        ))
+                )
+            {
+                return Err(EntryPointError::InvalidMultiDimensionalSubgroupBuiltIn.with_span());
+            }
```
teoxoy: We should do this in `VaryingContext.validate_impl` or else we will miss validating this in structs.

Also, do we need to disallow `SubgroupInvocationId` as well? I thought we didn't since it has equivalents in all backends.

atlv24:
Merge conflicts and compile errors fixed here: https://github.com/Lichtso/wgpu/pull/1

naga/src/valid/interface.rs
atlv24: I looked into resolving this feedback, but I dont understand. Why would structs need validation about subgroup functions?
Also, I checked and `SubgroupInvocationId` is indeed implemented on all backends, but I think this validation is unrelated to that. Its validating the dimensionality of workgroups to make sure subgroup operations make sense, I think. What I dont understand is why subgroups arent supported on workgroup dimensionalities that are not of the form (X, 1, 1).

Lichtso: > What I dont understand is why subgroups arent supported on workgroup dimensionalities that are not of the form (X, 1, 1).

That has to do with the discussion in http://github.com/gfx-rs/wgpu/pull/5301#discussion_r1534415235 , specifically the mapping between invocation in workgroup and invocation in subgroup:

> And if WorkGroupSize.x % SubgroupSize == 0 then it is also guaranteed that:
> LocalInvocationIndex = SubgroupLocalInvocationId + SubgroupId * SubgroupSize

Subgroups are supported for multi-dimensional workgroup sizes but the built-ins SubgroupLocalInvocationId and SubgroupId not.

> Also, do we need to disallow SubgroupInvocationId as well? 

Don't have to but I think we should for now. Because of the mapping above, it makes little sense to have SubgroupInvocationId without SubgroupId. And allowing it later is easier than the other way around.

teoxoy (approved):
I think it's time 🎉 

Thank you for this substantial contribution!

teoxoy:
@cwfitzgerald do you want to give this another look? Since you've previously requested changes.

cwfitzgerald (approved):
Amazing work everyone!

Lichtso:
@exrook We did it! The subgroup feature is merged.

@teoxoy You might want to adjust the CHANGELOG, I think we accidentally categorized it as "Bug fix" not "New features".

cwfitzgerald:
I can touch up the changelog, there is a merge conlfict on trunk between this and #5508 that I'm fixing

teoxoy:
I put together a new issue to track the differences between our implementation and the WebGPU proposal https://github.com/gfx-rs/wgpu/issues/5555.

9291Sam:
Are there plans to implement subgroupElect?

ErichDonGubler:
@9291Sam: See <https://github.com/gfx-rs/wgpu/issues/5555> for more details; it might not yet have the answer to the question you're asking, but it _is_ where we should discuss it.

PR #5305, opened 2024-02-26
Fix `extractBits` and `insertBits` When `offset + count` Overflows the Bit Width

cwfitzgerald:
**Connections**

Found in my push for #3159.

**Description**

All of the underlying implementations (including our polyfill) do not provide any guarantees about behavior when the offset/count overflows the bit width of the value being inserted/extracted. To counteract this, we first sanitize the offset/count to be within bounds before issuing the underlying call.

**Testing**

This is tested by a future PR submitting various integer tests for #3159.


teoxoy (approved):
Great stuff!

PR #5306, opened 2024-02-26
Make `GPUProgrammableStage.entryPoint` optional in `wgpu-core`

### LINKED ISSUE #5145: `GPUProgrammableStage.entryPoint` should be optional

ErichDonGubler:
**Connections**

Part of the work necessary for <https://github.com/gfx-rs/wgpu/pull/5148>, which resolves <https://github.com/gfx-rs/wgpu/issues/5145>. Unblocks agents like Firefox implementing this in their own environment as a first, easier step.

**Description**

An optional entry point is now part of the WebGPU standard, and we should adhere to it; see also <https://github.com/gpuweb/gpuweb/issues/4342>.

**Testing**

Some tests in `player/tests/` now exercise this via traces' specification of `entry_point: None`. However, there's no `wgpu-core` API tests exercising this, ATM. I think this is fine until such time that <https://github.com/gfx-rs/wgpu/pull/5148> has merged, which _should_ have more comprehensive testing over `wgpu`.

I expect that some coverage of this will be added over time; tracking this at <https://github.com/gpuweb/cts/issues/3432>.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,7 @@ Bottom level categories:
 ```
 - `wgpu::Id` now implements `PartialOrd`/`Ord` allowing it to be put in `BTreeMap`s. By @cwfitzgerald and @9291Sam in [#5176](https://github.com/gfx-rs/wgpu/pull/5176)
 - `wgpu::CommandEncoder::write_timestamp` requires now the new `wgpu::Features::TIMESTAMP_QUERY_INSIDE_ENCODERS` feature which is available on all native backends but not on WebGPU (due to a spec change `write_timestamp` is no longer supported on WebGPU). By @wumpf in [#5188](https://github.com/gfx-rs/wgpu/pull/5188)
+- Breaking change: [`wgpu_core::pipeline::ProgrammableStageDescriptor`](https://docs.rs/wgpu-core/latest/wgpu_core/pipeline/struct.ProgrammableStageDescriptor.html#structfield.entry_point) is now optional. By @ErichDonGubler in [#5305](https://github.com/gfx-rs/wgpu/pull/5305).
 
 #### GLES
 
--- a/wgpu-core/src/validation.rs
+++ b/wgpu-core/src/validation.rs
@@ -283,6 +283,16 @@ pub enum StageError {
     },
     #[error("Location[{location}] is provided by the previous stage output but is not consumed as input by this stage.")]
     InputNotConsumed { location: wgt::ShaderLocation },
+    #[error(
+        "Unable to select an entry point: no entry point was found in the provided shader module"
+    )]
+    NoEntryPointFound,
+    #[error(
+        "Unable to select an entry point: \
+        multiple entry points were found in the provided shader module, \
+        but no entry point was specified"
+    )]
+    MultipleEntryPointsFound,
 }
 
 fn map_storage_format_to_naga(format: wgt::TextureFormat) -> Option<naga::StorageFormat> {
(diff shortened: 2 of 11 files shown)


wgpu-core/src/validation.rs
```diff
@@ -1290,3 +1329,23 @@ pub fn validate_color_attachment_bytes_per_sample(
...
+    fn deref(&self) -> &Self::Target {
+        self.inner.as_ref()
+    }
+}
+
+impl FinalizedEntryPointName {
+    pub(crate) fn new(inner: String) -> Self {
+        Self { inner }
+    }
+}
```
ErichDonGubler: thought: I'm not sure that this newtype is worth it. I'm considering getting rid of it, and just using `String`s after finalizing the entry point name:

* Pro: It lets us include another flow hint as a type. This helps us to avoid a mistake with forgetting to finalize.
* OTOH, calling code would already have to figure out a conversion from `Option<&str>` to `String` without this, anyway.

nical: I would prefer to not have the newtype. I think that, as you said, converting from the option to the string is enough prevent us from forgetting that something has to happen.

nical (approved):
R+ although I'd prefer without the newtype.

wgpu-core/src/validation.rs
ErichDonGubler: Removed with 20fa2a84f.

CHANGELOG.md
```diff
@@ -102,6 +102,7 @@ Bottom level categories:
 ```
 - `wgpu::Id` now implements `PartialOrd`/`Ord` allowing it to be put in `BTreeMap`s. By @cwfitzgerald and @9291Sam in [#5176](https://github.com/gfx-rs/wgpu/pull/5176)
 - `wgpu::CommandEncoder::write_timestamp` requires now the new `wgpu::Features::TIMESTAMP_QUERY_INSIDE_ENCODERS` feature which is available on all native backends but not on WebGPU (due to a spec change `write_timestamp` is no longer supported on WebGPU). By @wumpf in [#5188](https://github.com/gfx-rs/wgpu/pull/5188)
+- BREAKING CHANGE: [`wgpu_core::pipeline::ProgrammableStageDescriptor`](https://docs.rs/wgpu-core/latest/wgpu_core/pipeline/struct.ProgrammableStageDescriptor.html#structfield.entry_point) is now optional. By @ErichDonGubler in [#5305](https://github.com/gfx-rs/wgpu/pull/5305).
```
cwfitzgerald: Merge when ready, but this probably shouldn't be broadly advertised as breaking, as this hasn't hit users yet.

cwfitzgerald (approved):

CHANGELOG.md
ErichDonGubler: Do you think we should remove `BREAKING CHANGE` in all caps, or that we could qualify with "this shouldn't affect `wgpu` users"?

cwfitzgerald: I would say remove the caps, you can keep the change itself.

ErichDonGubler: Lowercased with ffbf60904.

PR #5309, opened 2024-02-27
Fix staging buffer ID allocation mismatch

nical:
**Connections**

Followup from #5244 and #5229

**Description**

Two problems:
 - There was a mix of externally and internally allocated buffer IDs: regular buffers could be externally allocated but the temporary buffer used to initialize non-mappable-but-mapped-at-creation buffers was always internally allocated. The assertion added in #5229 caught this in gecko. To my understanding this was incorrect before since externally allocated IDs can't know about the IDs that are internally produced so they would silently stomp over one another in the registry. This is fixed by the first commit.
 - The trackers for render bundle execution were sized from the wrong source (the registry storage instead of the tracker index allocators). This is an oversight from #5244 and is fixed in the second commit.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`


ErichDonGubler (approved):
LGTM!

PR #5310, opened 2024-02-27
Fix docs.rs Builds

cwfitzgerald:
**Connections**

Between this and https://github.com/gfx-rs/metal-rs/pull/303, all targets should work on docs.rs again.

**Description**

We were just missing the `--cfg web_sys_unstable_apis` for wasm.


ErichDonGubler (approved):
lezduit

PR #5312, opened 2024-02-27
Fix Presentation

cwfitzgerald:
**Connections**

Followup to #5244 

**Description**

Just missed two small cases as surfaces aren't currently covered by tests.

**Testing**

Manually ran the examples.

cc @nical for vis


ErichDonGubler (approved):
This seems consistent with #5309 (~~CC @nical~~ EDIT: D'oh, just saw that he was already tagged in the OP. 😅).

PR #5297, opened 2024-02-24
Add multiple render targets example

kaphula:
**Connections**
None (I think).

**Description**
This PR adds new example that demonstrates how to use multiple render targets when you want to render to two or more targets from the fragment shader. 

As far as I know, none of the existing examples demonstrate this, and furthermore, finding information about how to do this with wgpu seems to be way too challenging, or at least it was for me. I had to ask help from multiple people who knew how to do this, and before that I was not even sure if this was possible at all.

**Testing**
I am not sure if this should or can be tested, or if the core feature has already been tested. Suggestions or instructions are welcome on how to implement the required tests.

Edit: I am not sure why it is showing the previous commits from other people, is that a problem?

PR #5313, opened 2024-02-27
Add multiple render targets example (v2)

kaphula:
**Connections**
This PR depends on https://github.com/gfx-rs/wgpu/pull/5312

This is a new version of https://github.com/gfx-rs/wgpu/pull/5297 which had some issues.

**Description**
This PR adds a new example that demonstrates how to use multiple render targets when you want to render to two or more targets from the fragment shader. 

As far as I know, none of the existing examples demonstrate this, and furthermore, finding information about how to do this with wgpu seems to be way too challenging, or at least it was for me. I had to ask help from multiple people who knew how to do this, and before that I was not even sure if this was possible at all.

**Testing**
I am not sure if this should or can be tested, or if the core feature has already been tested. Suggestions or instructions are welcome on how to implement the required tests.

**Checklist**

- [x] Run `cargo fmt`.


examples/src/multiple_render_targets/mod.rs
```diff
@@ -0,0 +1,536 @@
+use glam::Vec2;
+use std::borrow::Cow;
+use wgpu::{
+    Adapter, BindGroup, BindGroupLayout, ColorTargetState, CommandEncoder, Device, Extent3d, Queue,
+    RenderPassColorAttachment, RenderPipeline, Sampler, ShaderModule, SurfaceConfiguration,
+    Texture, TextureAspect, TextureDimension, TextureFormat, TextureUsages, TextureView,
+    TextureViewDescriptor, TextureViewDimension,
+};
+use winit::event::WindowEvent;
```
cwfitzgerald: Going to be _very_ annoying, and say we probably shouldn't import these names and always qualify with wgpu or the library they're from. It helps readability a ton when looking at a small section of code.

examples/src/multiple_render_targets/mod.rs
```diff
@@ -0,0 +1,536 @@
...
+                for x in 0..height {
+                    let cur_pos = Vec2::new(x as f32, y as f32);
+                    let distance_to_center_normalized =
+                        1.0 - (cur_pos - center).length() / half_distance;
+                    let val: u8 = (u8::MAX as f32 * distance_to_center_normalized) as u8;
+                    img_data.push(val)
+                }
+            }
+            img_data
+        }
```
cwfitzgerald: Let's make this a normal free function, not a nested one.

examples/src/multiple_render_targets/mod.rs
```diff
@@ -0,0 +1,536 @@
...
+            timestamp_writes: None,
+            occlusion_query_set: None,
+        });
+        rpass.set_pipeline(&self.pipeline);
+        rpass.set_bind_group(0, &self.bindgroup, &[]);
+        rpass.draw(0..3, 0..1);
+    }
+}
+
+fn create_target_textures(
```
cwfitzgerald: Lets make this a `TextureTarget::new` and move it next to the definition of TextureTargets

examples/src/multiple_render_targets/mod.rs
```diff
@@ -0,0 +1,536 @@
...
+                    write_mask: Default::default(),
+                })],
+            }),
+            primitive: wgpu::PrimitiveState::default(),
+            depth_stencil: None,
+            multisample: wgpu::MultisampleState::default(),
+            multiview: None,
+        });
+
+        let (ba, bb) =
```
cwfitzgerald:
```suggestion
        let (bg_a, bg_b) =
```

Or better yet bg_left bg_right.

cwfitzgerald (changes requested):
Looking good, some small comments, then probably one more small scale review, and this can land!

kaphula:
@cwfitzgerald Your requested changes are now implemented. Let me know if there is anything else.

kaphula:
I added another commit to make all the imports direct as I realized that is probably what you meant and not just the wgpu imports.

examples/src/multiple_render_targets/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+## To Run
+
+```
+cargo run --bin wgpu-examples multiple_render_targets
+```
+
+## Screenshots
+
+![Multi render target](./screenshot.png)
```
cwfitzgerald, line 21: Did you run the tests locally with `cargo xtask test`? The screenshot was never generated

examples/src/multiple_render_targets/mod.rs
```diff
@@ -0,0 +1,541 @@
...
+
+pub fn main() {
+    crate::framework::run::<Example>(EXAMPLE_NAME);
+}
+
+#[cfg(test)]
+#[wgpu_test::gpu_test]
+static TEST: crate::framework::ExampleTestParams = crate::framework::ExampleTestParams {
+    name: EXAMPLE_NAME,
+    // Generated on 1080ti on Vk/Windows
```
cwfitzgerald: These seem to just be blindly copied from other examples

cwfitzgerald (changes requested):
A couple issues with the screenshot and such. Additionally in general the code needs a lot more commentary about what's going on and why (adding things like method documentation would go a long way).

Additionally there seems the tests are failing inside the GL backend, we can mark it as failing to not block merging this, but should be filed as a follow up issue.

examples/src/multiple_render_targets/README.md
kaphula: The screenshot gets generated just fine for me, although bunch of tests fail on my machine overall. Do you want the generated screenshot to be part of the repository?

examples/src/multiple_render_targets/mod.rs
kaphula: Is that a problem? Can you give me some guidance on how the test should be configured for this specific example?

kaphula: As for the comments, I am not sure what should be commented better. The readme of the example describes what the example does and how it works. There are also various comments in the code. The render function has inline comments describing what the draw calls of the example do. Should these comments be on the methods instead?

Is the issue with GL failing something related to the internals of wgpu and not this example? If not, is there something I can do about it?

examples/src/multiple_render_targets/README.md
cwfitzgerald: Yeah - the fact that the screenshot is generated is how we ensure the screenshot stays up to date with the example

examples/src/multiple_render_targets/mod.rs
cwfitzgerald: > Is that a problem? Can you give me some guidance on how the test should be configured for this specific example?

Basically the idea is that you'll start with the lowest possible tolerance, then increase it as you find issues on specific gpu/api combinations. In your case, this should be listed as whatever card you ran it on to generate the tolerance. This way we have a record of what device is keeping the limit high, so we can lower it if needed.

cwfitzgerald:
I don't think I did a very good job of reviewing this the first time around and giving actionable advice, so I'm going to give this another go. It also completely fell through the cracks of reviewing - sorry for all the trouble, we try to be more on top of things than this.

examples/src/multiple_render_targets/mod.rs
cwfitzgerald: > As for the comments, I am not sure what should be commented better. The readme of the example describes what the example does and how it works. There are also various comments in the code. The render function has inline comments describing what the draw calls of the example do. Should these comments be on the methods instead?

Let me give this another, better look

examples/src/multiple_render_targets/mod.rs
```diff
@@ -0,0 +1,547 @@
...
+                    resource: wgpu::BindingResource::Sampler(&sampler),
+                },
+            ],
+            label: None,
+        });
+
+        let ts = target_states
+            .iter()
+            .map(|x| Some(x.clone()))
+            .collect::<Vec<_>>();
```
cwfitzgerald: I can't tell why this is needed - shouldn't you be able to pass target_states directly?

examples/src/multiple_render_targets/mod.rs
```diff
@@ -0,0 +1,547 @@
...
+            &self.bindgroup_layout,
+            texture_targets,
+            &self.sampler,
+        )
+    }
+}
+
+struct TextureTargets {
+    a_view: wgpu::TextureView,
+    b_view: wgpu::TextureView,
```
cwfitzgerald: I think these could have better names. Maybe `red_target_view`, or something similar. Same change for the locals in TextureTargets::new

cwfitzgerald (changes requested):
Alright, I have a few small comments, but I think this is largely ready to go! I've pushed a commit which fixes compilation issues that have happened in the interim.

Thanks for the patience :)

cwfitzgerald (approved):
Very nice!

kaphula:
Hey, no worries. Better prioritize the library itself over examples first.

I applied your suggested changes. Let me know what you think. Should the changelog addition be moved to next release or do we leave it under `## v0.19.4 (2024-04-17)`?

cwfitzgerald:
> I applied your suggested changes. Let me know what you think.

Nice!

> Should the changelog addition be moved to next release or do we leave it under ## v0.19.4 (2024-04-17)?

The next release

cwfitzgerald:
Amazing, lets land this!

PR #5319, opened 2024-02-28
Pipeline cache API and implementation for Vulkan

### LINKED ISSUE #5293: Pipeline Caching (between program executions)
In [Vello](https://github.com/linebender/vello/) on Android, the time taken to start the app up is unacceptably long (~2 seconds after linebender/vello#455). This is an order of magnitude longer than users would expect app startup to take.
The vast majority of this startup time is spent in calls to [`wgpu::Device::create_compute_pipeline`](https://docs.rs/wgpu/latest/wgpu/struct.Device.html#method.create_compute_pipeline). This is because each shader is being compiled to device microcode from scratch, each run.

We would like for `wgpu` to provide an unsafe API for [`VkPipelineCache`](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPipelineCache.html) objects, to allow reusing device microcode compilation between executions.

My proposed API would be for the application to provide a path to a directory for this cache to be stored in/retrieved from.
When creating a pipeline cache object, I would expect wgpu to attempt to read from the file `wgpu_vulkan.cache` (or alternative name) when initialised with a Vulkan backend, then create a pipeline cache from this value. 
This would also perform sanity checks on device version, probably by including an additional custom header (as discussed in [Creating a Robust Pipeline Cache with Vulkan](https://medium.com/@zeuxcg/creating-a-robust-pipeline-cache-with-vulkan-961d09416cda))

A method would then also be added to the cache to write the data from the cache back to disk.

Variations on the proposed solution:
1) The API could instead accept and output the data as an opaque blob. This would however run into complications if this were extended to include multiple backends, as the data for each backend would need to be loaded from disk, even when only one backend was being used.
2) The API could store the Vulkan specific data format - that is, saving the data which would be passed to [`PipelineCacheCreateInfoBuilder::initial_data`](https://docs.rs/ash/latest/ash/vk/struct.PipelineCacheCreateInfoBuilder.html#method.initial_data) directly in the file.  This would leave it up to applications to implement any sanity checking, beyond those provided by the drivers.
3) Combining 1 and 2, i.e. accepting a `&[u8]` which will be passed as-is to `PipelineCacheCreateInfoBuilder::initial_data`.
4) The APIs could be `fn(impl FnOnce(Backend)->Result<Vec<u8>, E>)->Result<Vec<u8>, E>` or `impl FnOnce(Backend, Vec<u8>) -> R`. This would alleviate most concerns, and is probably the right approach.

Alternative solutions:
1) wgpu could automatically implement this pipeline caching, without requiring manual implementation from user apps. I believe this is likely to be untenable for several reasons:
   1) wgpu cannot know where the cache data should be stored, without implementing suspect heuristics based on e.g. executable name
   2) wgpu cannot trust the cache data, as it could be modified/corrupted by the end user of the application or other programs. This means wgpu would have unavoidable unsoundness[^1]
   3) wgpu cannot know when to save this cache data, as the usage of wgpu is dependent on the application. E.g. some applications may initialise additional shaders later into their program execution

2) wgpu could allow us to pass an [`ash::vk::PipelineCache`](https://docs.rs/ash/latest/ash/vk/struct.PipelineCache.html) to `wgpu::Device::create_compute_pipeline` - either in the `ComputePipelineDescriptor` or through a new method. I suspect this is untenable, as it would specialise.
3) wgpu could allow us to pass an `ash::vk::PipelineCache` to [`wgpu::hal::vulkan::Device::create_compute_pipeline`](https://docs.rs/wgpu/latest/wgpu/hal/vulkan/struct.Device.html#method.create_compute_pipeline), and allow creating a `wgpu::ComputePipeline` from a `wgpu::hal::vulkan::ComputePipeline`. I don't know why this second aspect is currently not permitted.

**Additional context**
I have not researched other backends' caching APIs.

I have implemented an experiment to justify the requirement to. The code of that experiment can be found in linebender/vello#459 which depends on #5292.
On my device (a Google Pixel 6), this reduces pipeline creation time (on non-first boots) from ~2s to ~30ms, and empirically makes the test app launch as quickly as I'd expect an app to launch.

I suggest that the full API for the pipeline cache could look something like:
```rust
struct PipelineCache; // Probably not a unit struct in the final iteration
impl Clone for PipelineCache {}

impl Device {
    unsafe fn create_pipeline_cache(&self, folder: PathBuf) -> PipelineCache;
}
struct ComputePipelineDescriptor {
     cache: PipelineCache
}
impl PipelineCache {
     fn write_to_disk(&self) -> Result;
}
```
Future possibilities could allow using APIs such as [`vkMergePipelineCaches`](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkMergePipelineCaches.html).

I am willing to implement this in wgpu, but need guidance around how adding new resources should look, as well as the expected API

[^1]: This unsoundness is also unavoidable on the part of any programs using this feature of wgpu, but that's a tradeoff some users (including Vello) are able to justify.

DJMcNab:
**Connections**
- Fixes #5293 (for Vulkan)
- Required for https://github.com/linebender/vello/issues/289
- See also https://github.com/linebender/vello/pull/459 which validated that this was required

**Description**
Adds a `PipelineCache` resource to `wgpu`'s API, and wires it down to Vulkan [`vkCreatePipelineCache`](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCreatePipelineCache.html).
This cache is added as a parameter to [`ComputePipelineDescriptor`](https://docs.rs/wgpu/latest/wgpu/struct.ComputePipelineDescriptor.html) and [`RenderPipelineDescriptor`](https://docs.rs/wgpu/latest/wgpu/struct.RenderPipelineDescriptor.html).
This parameter is used as expected on Vulkan, and ignored on DirectX/Metal/WebGPU/gles.

**Testing**
Testing has been performed on a Pixel 6, which has validated that this works as expected between runs.
This is in [the branch](https://github.com/DJMcNab/vello/tree/proper-pipeline-pcache) in my fork of Vello.
This does only test the cache for Vulkan pipelining.

**Checklist**

- [x] Run `cargo fmt`.
  - [X] `--target wasm32-unknown-emscripten` - gets the same errors on `main`, so I assume it's fine
- [X] Run `cargo xtask test` to run tests - I get the same failures on `main`, so I assume it's fine

**Status**
This PR is now fully ready for review

**Future Work**
I am planning on implementing additional validation inside wgpu that the cache is suitable, as discussed in https://medium.com/@zeuxcg/creating-a-robust-pipeline-cache-with-vulkan-961d09416cda. That is mainly about storing the driver version and file size inline in the file. This should be done before this feature is released, but could be in a different pull request. Please advise if you have any preference here.

This could also be extended to different platforms/APIs - see e.g. https://github.com/gfx-rs/gfx/issues/2877
I am not intending to do this work myself


DJMcNab:
cc @expenses, as you had a previous investigation in https://github.com/gfx-rs/gfx/issues/3716

wgpu-core/src/device/resource.rs
```diff
@@ -2744,13 +2744,27 @@ impl<A: HalApi> Device<A> {
...
+                if cache.device.as_info().id() != self.as_info().id() {
+                    return Err(DeviceError::WrongDevice.into());
+                }
+                Some(cache)
+            } else {
+                None
+            }
+        } else {
+            None
+        };
```
cwfitzgerald: Lets replace this chain with a block with conditional breaks in them.

wgpu-core/src/pipeline.rs
```diff
@@ -383,6 +447,15 @@ pub struct RenderPipelineDescriptor<'a> {
     /// If the pipeline will be used with a multiview render pass, this indicates how many array
     /// layers the attachments will have.
     pub multiview: Option<NonZeroU32>,
+    pub cache: Option<PipelineCacheId>,
```
cwfitzgerald, line 384: Could we just copy the docs from wgpu here

wgpu-hal/src/vulkan/device.rs
```diff
@@ -1850,7 +1850,13 @@ impl crate::Device<super::Api> for super::Device {
             unsafe {
                 self.shared
                     .raw
-                    .create_graphics_pipelines(vk::PipelineCache::null(), &vk_infos, None)
+                    .create_graphics_pipelines(
+                        desc.cache
+                            .map(|it| it.raw)
+                            .unwrap_or(vk::PipelineCache::null()),
```
cwfitzgerald: Lets pull this out of the function call

wgpu-types/src/lib.rs
```diff
@@ -873,6 +873,11 @@ bitflags::bitflags! {
         /// - Vulkan (with dualSrcBlend)
         /// - DX12
         const DUAL_SOURCE_BLENDING = 1 << 54;
+        /// Allows the use of pipeline cache objects
+        ///
+        /// Supported platforms:
+        /// - Vulkan
+        const PIPELINE_CACHE = 1 << 55;
```
cwfitzgerald: I don't think we need this feature at all? Isn't an empty pipeline cache a valid one?

cwfitzgerald (changes requested):
Thanks for this! The code itself is fine, keep it up like this, and the documentation is _dreamy_.

The api has a few issues at multiple layers. At points wgpu-hal is expected to do validation, at other points the end user is just expected to know how to pass the correct thing. I think we should unify the api as follows.

## wgpu-hal

The wgpu-hal layer should just expect that the blob passed to it is correct for that platform and driver combination. It shouldn't do any additional validation. All of its internal code can expect that the cache key is honored.

It may append additional data structures to the blob which contain shader intermediate states (such as HLSL generated or DXBC/DXIL). 

There is a method on the Adapter which returns a backend-specific string of bytes to serve as a "cache key". This includes the wgpu-hal version.

## wgpu-core

Core does all of the validation that the key is honored and the file was not corrupted.

It adds metadata to the top of the blob that stores the magic number, file size, file hash, and pipeline key.

If this errors, it will create a new empty cache.

## wgpu

Passes through the core apis.

## User

The user will pull the pipeline key from the adapter, hash it, and use it as a filename to store the data.

Does this all make sense?

DJMcNab:
Thanks for the review @cwfitzgerald 

In terms of validation, my intent was to follow up with that, because I wasn't sure I was doing the right thing. So I can definitely add that in here.
So I can definitely add that validation in this PR, since the rest of the shape is approximately right.
One thing I need guidance on is how to hash the data in `wgpu_core`. Are you aware of a preferred library for this? It feels like the sort of thing which is quite opinionated

I just want some clarification about what you mean by:
> The user will pull the pipeline key from the adapter, hash it, and use it as a filename to store the data.

In my prototype integration, I wasn't hashing the filename. I find it unlikely that users would do so, either.
Should we pre-hash the value from the cache key method?

DJMcNab:
> It shouldn't do any additional validation

This is also slightly troubling, as different APIs have different validation needs. For example, Vulkan *probably* needs to validate the driver version is the same, whereas the concept of a driver version isn't exposed at the higher levels

wgpu-types/src/lib.rs
DJMcNab: The reason I thought we would have this is for consistency with other features which only exist for some backends.

It definitely makes the API easier to use to not have it though. I only added it to pre-empt a review comment asking for it to be added

xorgy:
> For example, Vulkan _probably_ needs to validate the driver version is the same, whereas the concept of a driver version isn't exposed at the higher levels

That's interesting... are there Vk drivers in the wild that accept incompatible pipeline caches from applications, or is this more preemptive (due to the quality issues with some drivers)?

DJMcNab:
According to the [linked article](https://medium.com/@zeuxcg/creating-a-robust-pipeline-cache-with-vulkan-961d09416cda#7837), there have been cases in the past where a driver either:
1) Doesn't validate the UUID which is meant to detect this at all
2) Fails to update the UUID when it would be incompatible

cwfitzgerald:
> This is also slightly troubling, as different APIs have different validation needs. For example, Vulkan probably needs to validate the driver version is the same, whereas the concept of a driver version isn't exposed at the higher levels

The idea is that the wgpu-hal decides on its pipeline key based on its validation needs, and wgpu-core is the one who does the comparison and rejection of the pipeline key. This way the driver version will be part of the wgpu-hal pipeline key.

> In my prototype integration, I wasn't hashing the filename. I find it unlikely that users would do so, either.
Should we pre-hash the value from the cache key method?

If we make cache key an opaque type we can put a `into_bytes` and `hash` method on it.

DJMcNab:
The key factor which needs to be accounted for is:
1) If the driver updates, the old cache you still have is *useless*. 

I'm planning on implementing that with effectively two keys, a "device" key (which identifies the device/backend combination) and a further validation key. The device key should related to the "file name" being used, whereas the validation key is just used to bail from using the cache. That is potentially what you are describing, but I think it's good to clarify that

DJMcNab:
This is now ready for follow-up review. I am going to now test this updated version in Vello, and also need some guidance on computing the `hash` (it currently uses a hardcoded value).

I'm unlikely to make further functional changes before another review

DJMcNab:
Having done some more thinking on this, I have decided to exclude the hashing of the cache data from this PR. The main reasons for this are:
- I believe it to be unlikely that filesystems would corrupt the data in a way which keeps the same length, but changes some of the content
- Any attempt at meddling in the file by other processes/application is already declared out of scope by our safety comment
- We do not already have a dependency which would compute a suitable hash. We could likely produce a hash designed for a hashmap, but these have significantly different requirements to our hashing need.

I also don't see the necessity in making the pipeline cache key an opaque set of bytes, rather than a readable string.
These files should be in a location which would be rarely accessed by end-users.
Additionally, if using these as a file name, an application would have to sanitise the bytes to be reasonable for use as a file name. I suspect at least some applications would not do so, and potentially lose some of portability which wgpu provides.
But I also think it's plausible that I've missed some important reason for this to be an incorrect assessment.

examples/src/mipmap/mod.rs
```diff
@@ -306,6 +307,7 @@ impl crate::framework::Example for Example {
             depth_stencil: None,
             multisample: wgpu::MultisampleState::default(),
             multiview: None,
+            cache: None,
```
cwfitzgerald, line 314: As discussed, this should go in the new compilation options struct

wgpu-core/src/pipeline_cache.rs
```diff
@@ -0,0 +1,530 @@
...
+            [1, 255, 255, 255, 0, 2, 0xFE, 0xED],         // Backend and Adapter key
+            [0xFE, 0xFE, 0xFE, 0xFE, 255, 255, 255, 255], // Backend and Adapter key
+            0xFFFFFFFF_FFFFFFFFu64.to_be_bytes(),         // Validation key
+            0x88888888_88888888u64.to_be_bytes(),         // Validation key
+            0x0u64.to_be_bytes(),                         // Data size
+            0xFEDCBA9_876543210u64.to_be_bytes(),         // Hash
+        ];
+        let expected = cache.into_iter().flatten().collect::<Vec<u8>>();
+
+        assert_eq!(result.as_slice(), expected.as_slice());
```
cwfitzgerald, line 338: These tests are thorough and amazing! One thing that could make it shorter is if you lift `cache` into a const, which you then modify inside the body. It's debatable if this is actually clearer. Thoughts?

wgpu/src/lib.rs
```diff
@@ -2703,6 +2839,90 @@ impl Device {
...
+    /// Create a [`PipelineCache`] with initial data
+    ///
+    /// This can be passed to [`Device::create_compute_pipeline`]
+    /// and [`Device::create_render_pipeline`] to either accelerate these
+    /// or add the cache results from those.
+    ///
+    /// # Safety
+    ///
+    /// The `data` field of `desc` must have previously been returned from a call
+    /// to [`PipelineCache::get_data`][^saving]. It's recommended to only `data` for the same
```
cwfitzgerald: This line has a typo - not entirely sure what you were trying to say.

wgpu-hal/src/vulkan/device.rs
```diff
@@ -1925,6 +1935,27 @@ impl crate::Device<super::Api> for super::Device {
         unsafe { self.shared.raw.destroy_pipeline(pipeline.raw, None) };
     }
 
+    unsafe fn create_pipeline_cache(
+        &self,
+        desc: &crate::PipelineCacheDescriptor<'_>,
+    ) -> Result<PipelineCache, crate::PipelineCacheError> {
+        let mut info = vk::PipelineCacheCreateInfo::builder();
+        // TODO: Add additional validation to the data, as described in https://medium.com/@zeuxcg/creating-a-robust-pipeline-cache-with-vulkan-961d09416cda
```
cwfitzgerald: Isn't this handled now?

wgpu/src/lib.rs
```diff
@@ -1575,10 +1672,49 @@ pub struct ComputePipelineDescriptor<'a> {
...
 #[cfg(send_sync)]
 static_assertions::assert_impl_all!(ComputePipelineDescriptor<'_>: Send, Sync);
 
+/// Describes a pipeline cache which reuses data from a previous run.
+///
+/// For use with [`Device::create_pipeline_cache_init`].
+///
+/// This type is unique to the Rust API of `wgpu`.
+#[derive(Clone, Debug)]
+pub struct PipelineCacheInitDescriptor<'a> {
```
cwfitzgerald: I feel like these two endpoints can be combined into one, with data being an option of a struct. This would be a lot cleaner than needing two entirely separate entry points.

cwfitzgerald (changes requested):
This is absolutely fabulous work!

I think the setup for caching is perfectly reasonable, and the core/hal relationship is exactly what i wanted to see!

Have some comments. I also would like to see an execution test that actually uses the pipeline cache. Maybe make a cache, make a pipeline, serialize the cache, re-parse it to a new cache, and make an identical pipeline. Just to make sure everything actually works and there aren't any api level issues.

wgpu/src/lib.rs
DJMcNab: We still want it to be unsafe, right?

This would be equivalent to the bounds check policy unchecked constructor?

cwfitzgerald: Yeah, I think it is perfectly reasonable to have all pipeline caches be unsafe to construct - constructing an empty one just isn't enough of a use case for it to really bother me.

wgpu-core/src/pipeline_cache.rs
DJMcNab: Maybe? I think dealing with the indices manually isn't great either, unfortunately.

cwfitzgerald: Eh, probably not worth it now that it's typed out.

DJMcNab:
Re-requesting review pending CI

tests/tests/pipeline_cache.rs
```diff
@@ -0,0 +1,204 @@
...
+
+use wgpu_test::{gpu_test, GpuTestConfiguration, TestParameters, TestingContext};
+
+/// We want to test that partial updates to push constants work as expected.
+///
+/// As such, we dispatch two compute passes, one which writes the values
+/// before a partial update, and one which writes the values after the partial update.
+///
+/// If the update code is working correctly, the values not written to by the second update
+/// will remain unchanged.
```
cwfitzgerald: Copy-pasta error

tests/tests/pipeline_cache.rs
```diff
@@ -0,0 +1,204 @@
...
+#[gpu_test]
+static PIPELINE_CACHE: GpuTestConfiguration = GpuTestConfiguration::new()
+    .parameters(
+        TestParameters::default()
+            .test_features_limits()
+            .features(wgpu::Features::PIPELINE_CACHE)
+            .skip(wgpu_test::FailureCase::adapter("llvmpipe")),
```
cwfitzgerald: Why does this fail?

tests/tests/pipeline_cache.rs
```diff
@@ -0,0 +1,204 @@
...
+        @compute @workgroup_size(1)
+        fn main() {{
+        {}
+        }}
+        "#,
+        (0..ARRAY_SIZE).fold(String::new(), |mut s, v| {
+            // "Safety": There will only be a single workgroup, and a single thread in that workgroup
+            writeln!(s, "    output[{v}] = {v}u;").expect("String");
+            s
+        })
```
cwfitzgerald: Could we construct this string in a normal for loop, then format it after?

tests/tests/pipeline_cache.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    let second_cache = unsafe {
+        ctx.device
+            .create_pipeline_cache(&wgpu::PipelineCacheDescriptor {
+                label: Some("pipeline_cache"),
+                data: first_cache_data.as_deref(),
+                fallback: false,
+            })
+    };
+    let start = std::time::Instant::now();
```
cwfitzgerald: Should use web_time::Instant (and we can import this)

tests/tests/pipeline_cache.rs
```diff
@@ -0,0 +1,204 @@
...
+            },
+        });
+    let duration = start.elapsed();
+    validate_pipeline(&ctx, first_pipeline, &bind_group, &gpu_buffer, &cpu_buffer).await;
+    if false {
+        // Ideally, we could make this assertion. However, that doesn't actually work, because drivers have
+        // their own internal caches. This does work on my machine if I set `MESA_DISABLE_PIPELINE_CACHE=1`
+        // before running the test; but of course that is not a realistic scenario
+        assert!(duration.as_millis() < first_pipeline_duration.as_millis());
+    }
```
cwfitzgerald: Lets just nuke the code, and leave the comment

tests/tests/pipeline_cache.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    {
+        let mut cpass = encoder.begin_compute_pass(&wgpu::ComputePassDescriptor {
+            label: Some("compute_pass"),
+            timestamp_writes: None,
+        });
+        cpass.set_pipeline(&pipeline);
+        cpass.set_bind_group(0, bind_group, &[]);
+
+        // -- Dispatch 0 --
```
cwfitzgerald: Copy-paste error

wgpu-hal/src/vulkan/device.rs
```diff
@@ -1863,12 +1863,23 @@ impl crate::Device for super::Device {
                 .build()
         }];
 
+        let vertex_cache = desc.vertex_stage.cache.map(|it| it.raw);
+        let fragment_cache = desc
+            .fragment_stage
+            .as_ref()
+            .and_then(|it| it.cache)
+            .map(|it| it.raw);
+        // TODO: What should the behaviour be when both are set and different?
```
cwfitzgerald: We should just assert they're the same, and document the requirement on the wgpu-hal interface - wgpu-core will always uphold it, so this is fine.

wgpu-types/src/lib.rs
```diff
@@ -914,6 +914,11 @@ bitflags::bitflags! {
         ///
         /// This is a native only feature.
         const SUBGROUP_BARRIER = 1 << 58;
+        /// Allows the use of pipeline cache objects
+        ///
+        /// Supported platforms:
+        /// - Vulkan
```
cwfitzgerald, line 920:
```suggestion
        /// Supported platforms:
        /// - Vulkan
        ///
        /// Unimplemented Platforms:
        /// - DX12
        /// - Metal
```

cwfitzgerald (changes requested):
Sorry for taking so long to get back to you on this - we're basically there, just some small nits at this point.

tests/tests/pipeline_cache.rs
DJMcNab: IIRC, the issue was with the compilation time testing, so it's no longer needed.

Thanks for catching this

tests/tests/pipeline_cache.rs
DJMcNab: No longer needed now that we're removing the checking code

wgpu-hal/src/vulkan/device.rs
DJMcNab: How should `wgpu-core` uphold it? The way that it works with `PipelineCompilationOptions` works, it's specified directly on each of the vertex and fragment shaders.

The problem becomes that `constants` are per-shader, whereas the pipeline cache is across the entire pipeline (at least *in Vulkan*).
I think the least bad option is to specify that the vertex cache will be prioritised on platforms where these are the same?

cwfitzgerald: Ah! Core should check to make sure the caches are the same - honestly it sounds like it needs to be part of the base pipeline descriptor instead of a per-entrypoint argument.

cwfitzgerald (changes requested):
Just need your throughs on where the pipeline cache should go, then we're good to land.

wgpu-hal/src/vulkan/device.rs
DJMcNab: Yeah, possibly adding it to the root descriptor is best.

But it's a bit annoying to ask every single user to need to add this field, when they already have a defaulted `PipelineCompilationOptions`.

I can also imagine these being meaningfully distinct, such as for producing DXIL (although combining them would also work in that case).

An alternative is to merge the caches on Vulkan, and make them alias the same raw cache object. There don't seem to be any good options, because of the semantics here

I won't be addressing this until at least Thursday, but I suppose putting in the top-level descriptor is actually best.

Hinting to users in the upgrade that the cache exists and that they should/could use it is possibly a good thing, anyway

cwfitzgerald: Sounds like root is the way to go. Give me a buzz when that's ready and we can get this in.

cwfitzgerald:
Looks to be an issue with the ash update

cwfitzgerald (approved):

cwfitzgerald:
I'll fix this up and get it in

PR #5326, opened 2024-03-01
wgpu-hal: add ndk-sys dependency to fix linking error.

ashdnazg:
**Description**
Compilation of an android library which depends on wgpu fails during linking with the error:
```
ld: error: undefined symbol: ANativeWindow_setBuffersGeometry
```

This isn't encountered in the tests, since the dev dependencies include `winit` which depends on `ndk-sys` when targeting android.

**Testing**
This was tested by succeeding to compile a project which did not compile previously.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-hal/src/gles/egl.rs
+++ b/wgpu-hal/src/gles/egl.rs
@@ -50,16 +50,6 @@ type WlEglWindowResizeFun = unsafe extern "system" fn(
 
 type WlEglWindowDestroyFun = unsafe extern "system" fn(window: *const raw::c_void);
 
-#[cfg(target_os = "android")]
-extern "C" {
-    pub fn ANativeWindow_setBuffersGeometry(
-        window: *mut raw::c_void,
-        width: i32,
-        height: i32,
-        format: i32,
-    ) -> i32;
-}
-
 type EglLabel = *const raw::c_void;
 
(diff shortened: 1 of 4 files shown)


wgpu-hal/src/gles/egl.rs
```diff
@@ -50,16 +50,6 @@ type WlEglWindowResizeFun = unsafe extern "system" fn(
 
 type WlEglWindowDestroyFun = unsafe extern "system" fn(window: *const raw::c_void);
 
-#[cfg(target_os = "android")]
```
ashdnazg, line 53: An alternative would be to add `#[link(name = "android")]` here, but I think delegating this to `ndk-sys` is more robust.

cwfitzgerald (approved):
Very nice

MarijnS95:
Seeing this from the changelog: any clue why this function is called in the first place? `git blame` doesn't help.

I've played with EGL on Android a fair bit and it always returns configs for _all_ available formats, on devices I've tested. Only when calling `ANativeWindow_setBuffersGeometry()` is the output limited to just EGL configs containing that format set on the "`Surface`".

Is the assumption that the user may have limited the format before, so that this function is used to change it again, or to make config filtering easier?

PR #5331, opened 2024-03-02
Add support for running on OpenGL 4.1 with a core profile on macOS

bes:
**Description**

When running wgpu with an OpenGL context on macOS that is created with a core profile and with the forward-compatibility bit set, the MAX_VARYING_COMPONENTS constant returns 0 when queried. The default value is 60, so we return the default value if the query returns 0.

We also need to use `#version 140` on macOS since `#version 130` isn't accepted. Since `#version 140` should be available from OpenGL 3.1, we use that everywhere. That way we don't need any specific macOS flags or features.

**Testing**

* Manually tested on macOS with an OpenGL 4.1 Context with a core profile that I got from GStreamer.
* Automatic tests (`cargo xtask test`)

**Checklist**

- [x] Run `cargo fmt`.


Wumpf (approved):
looks good to me! Can you please add a note in the changelog?

bes:
Thanks! Added a line in the changelog!

PR #5338, opened 2024-03-04
Fix output args for render_to_texture example

ratmice:
**Connections**

**Description**
The render_to_texture instructions specified render-to-texture as a `--bin render-to-texture` rather than `wgpu-examples`.
Since the example (e.g. `render_to_texture`) is now specified as the first argument, the output argument should now be the second one.

This also adds the default output to .gitignore.

Currently it will try to create a file named `render_to_texture` which when running under `examples/src` already exists and is a directory, which fails.

**Testing**
`cargo run --bin wgpu-examples -- render_to_texture "test.png"`

**Checklist**

- [ ] Run `cargo fmt`.


Wumpf (approved):
nice catch(es). Thanks for cleaning this up!

PR #5345, opened 2024-03-05
Don't depend on validation layers for setting object names

### LINKED ISSUE #3913: On Vulkan, debug labels shouldn't require validation layer
In #3809, I unintentionally made the population of `wgpu_hal::vulkan::InstanceShared::debug_utils` conditional on *both*:
- the availability of the `VK_EXT_debug_utils` extension *and*
- the availability of the `VK_LAYER_KHRONOS_validation`.

It it perfectly possible to report debug labels even when validation is not enabled: the extension and the layer are independent of each other.

DJMcNab:
**Connections**
Fixes #3913

**Description**
Before this PR, you needed to use the `VALIDATION` instance flag to get labels set on Vulkan objects properly, because of an oversight in the PR linked in the issue.
This changes it so that only the `DEBUG` flag is required again

**Testing**
I tested the water example in renderdoc (in release mode, so no debug_assertions) with `WGPU_DEBUG=1` with and without this PR, and saw that the labels were correctly present after this PR.

**Checklist**

- [X] Run `cargo fmt`.
- [X] Run `cargo xtask test` to run tests. Same failures as on `main`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -161,6 +161,10 @@ By @cwfitzgerald in [#5325](https://github.com/gfx-rs/wgpu/pull/5325).
 
 - Fixes for being able to use an OpenGL 4.1 core context provided by macOS with wgpu. By @bes in [#5331](https://github.com/gfx-rs/wgpu/pull/5331).
 
+#### Vulkan
+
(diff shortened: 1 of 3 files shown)


jimblandy (approved):
This looks fine - thank you!

PR #5346, opened 2024-03-05
Cache MAX_SAMPLES on gles backend

Dinnerbone:
**Connections**

**Description**
Believe it or not, this actually showed a little hot for us when measuring on Firefox webgl. Every call to `get_texture_format_features` ended up with an IPC to fetch the max sample count, which is a static immutable value.

Let's cache it on creation and keep things cheap. Also helps avoid a lock, which is always nice.

**Testing**
All existing tests pass? :D

**Checklist**

- [X] Run `cargo fmt`.

### DIFF
--- a/wgpu-hal/src/gles/mod.rs
+++ b/wgpu-hal/src/gles/mod.rs
@@ -251,6 +251,11 @@ struct AdapterShared {
     next_shader_id: AtomicU32,
     program_cache: Mutex<ProgramCache>,
     es: bool,
+
+    /// Result of `gl.get_parameter_i32(glow::MAX_SAMPLES)`.
+    /// Cached here so it doesn't need to be queried every time texture format capabilities are requested.
+    /// (this has been shown to be a significant enough overhead)
+    max_msaa_samples: i32,
 }
 
 pub struct Adapter {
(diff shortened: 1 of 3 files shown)


wgpu-hal/src/gles/mod.rs
```diff
@@ -251,6 +251,7 @@ struct AdapterShared {
     next_shader_id: AtomicU32,
     program_cache: Mutex<ProgramCache>,
     es: bool,
+    max_samples: i32,
```
Wumpf: let's call this `max_msaa_samples` so it's actually clear what is meant for the next poor soul stumbling in here :). Also a note on why this is cached can't hurt:

```suggestion
    /// Result of `gl.get_parameter_i32(glow::MAX_SAMPLES)`.
    /// Cached here so it doesn't need to be queried every time texture format capabilities are requested.
    /// (this has been shown to be a significant enough overhead)
    max_msaa_samples: i32,
```

Wumpf (changes requested):
nice catch, thanks for getting fix improvement in 👍 
Just a small change request on doc & naming to make future-me not confused about this

wgpu-hal/src/gles/mod.rs
Dinnerbone: Changed, thanks for the suggestion!

Wumpf (approved):

PR #5348, opened 2024-03-05
Don't create shader-clear program on GLES if it's not needed

Dinnerbone:
**Connections**
n/a

**Description**
We've noticed this before in Ruffle, that the built in "shader clear program" can contribute to the startup times on webgl. We've seen each shader program take as much as 100ms to compile before, and they really add up.

wgpu doesn't actually need this in most cases though, so it's wasteful to spend the time creating it. This reduces it down to an `Option`, that's only populated if the workaround flag is enabled.

I was contemplating removing `Workarounds.MESA_I915_SRGB_SHADER_CLEAR` and having it simply be "this value is `Some()` if needed, `None` if not", which feels much more natural - but the workaround is stored per Adapter and the program per Queue right now, so I opted to leave it\*. I'm also unsure if wgpu prefers to have these things listed as clear workarounds instead, as I do see the merit there.

*\*though this did leave me to question: why are we making so many resources per Queue on gles, rather than Adapter? If you opened up two queues you have two shader clear programs, two zero buffers, two draw fbos, etc. I feel like we can move those to Adapter and save some potential waste, in a future PR.*

**Testing**
It passes tests and seems to run fine!

**Checklist**

- [X] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -160,6 +160,7 @@ By @cwfitzgerald in [#5325](https://github.com/gfx-rs/wgpu/pull/5325).
 #### GLES
 
 - Fixes for being able to use an OpenGL 4.1 core context provided by macOS with wgpu. By @bes in [#5331](https://github.com/gfx-rs/wgpu/pull/5331).
+- Don't create a program for shader-clearing if that workaround isn't required. By @Dinnerbone in [#5348](https://github.com/gfx-rs/wgpu/pull/5348).
 
 ## v0.19.0 (2024-01-17)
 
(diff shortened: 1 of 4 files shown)


Wumpf (approved):
Thanks for the fix, looks good to me!

I agree with both of your points - the program should live on the adapter and the workaround should be incorporated by the optional clear program itself, saving us the `except` down the line (the Option<ShaderClearProgram> should be then appropriately named to indicate what workaround it is for).

The only reason I can think of why this doesn't live on the adapter is that it might be to protect against creating the shader alongside with the adapter itself which might be unnecessary overhead when creating the adapter just to probe for available adapters. So ideally we'd actually fully lazily create the shader clear program upon first usage instead.

Would be nice if you or someone else could look into all this in a follow-up :), but I think this PR is a clear improvement, so in it goes!

PR #5350, opened 2024-03-05
docs: mention primitive restart in the description of strip_index_format

cpsdqs:
**Connections**
N/A

**Description**
Currently, the wgpu documentation is rather unclear on why `PrimitiveState::strip_index_format` exists and what its purpose is, unless you’ve read the webgpu spec and know that it enables primitive restart. Since primitive restart is also never mentioned anywhere else in the documentation, I thought it would be nice to clarify this doc comment to include it here.

(Though, after a cursory look at the implementation, strip_index_format only seems to be actually required on Vulkan and DX12, as Metal has it [always enabled](https://developer.apple.com/documentation/metal/mtlrendercommandencoder/1515520-drawindexedprimitives#discussion)? I’m not sure if that would be worth mentioning as an implementation quirk…)

**Testing**
this PR contains no code changes, but all tests still pass!

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ By @cwfitzgerald in [#5325](https://github.com/gfx-rs/wgpu/pull/5325).
 ### Documentation
 
 - Document Wayland specific behavior related to `SurfaceTexture::present`. By @i509VCB in [#5092](https://github.com/gfx-rs/wgpu/pull/5092).
+- Add mention of primitive restart in the description of `PrimitiveState::strip_index_format`. By @cpsdqs in [#5350](https://github.com/gfx-rs/wgpu/pull/5350)
 
 ### New features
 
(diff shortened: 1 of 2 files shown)


Wumpf (commented):
Thank you! Good to have that in here 👍 

> Though, after a cursory look at the implementation, strip_index_format only seems to be actually required on Vulkan and DX12, as Metal has it [always enabled](https://developer.apple.com/documentation/metal/mtlrendercommandencoder/1515520-drawindexedprimitives#discussion)? I’m not sure if that would be worth mentioning as an implementation quirk…

That's an implementation detail, no point in mentioning it when it's actually important to the behavior. So good as is

Wumpf (approved):
oops, forgot to actually approve the PR which is why it didn't get auto-merged. Apologies!

PR #5351, opened 2024-03-05
gles: fix crash when holding multiple devices on wayland/surfaceless.

### LINKED ISSUE #5349: Crash when creating multiple GL devices on wayland/surfaceless.
**Description**
If you hold multiple devices with a GL backend on a wayland or surfaceless platform, you'll get a crash when they drop, citing `BadDisplay`:
```
called `Result::unwrap()` on an `Err` value: BadDisplay
stack backtrace:
   0: rust_begin_unwind
             at /rustc/2c9c764110ebbaffa2e248628ef8da36b69727bf/library/std/src/panicking.rs:647:5
   1: core::panicking::panic_fmt
             at /rustc/2c9c764110ebbaffa2e248628ef8da36b69727bf/library/core/src/panicking.rs:72:14
             at /rustc/2c9c764110ebbaffa2e248628ef8da36b69727bf/library/core/src/result.rs:1649:5
             at /rustc/2c9c764110ebbaffa2e248628ef8da36b69727bf/library/core/src/result.rs:1073:23
   4: wgpu_hal::gles::egl::EglContext::make_current
             at /home/ashdnazg/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-hal-0.19.3/src/gles/egl.rs:293:9
   5: wgpu_hal::gles::egl::AdapterContext::lock::{{closure}}
             at /home/ashdnazg/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-hal-0.19.3/src/gles/egl.rs:407:13
   6: core::option::Option<T>::map
             at /rustc/2c9c764110ebbaffa2e248628ef8da36b69727bf/library/core/src/option.rs:1072:29
   7: wgpu_hal::gles::egl::AdapterContext::lock
             at /home/ashdnazg/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-hal-0.19.3/src/gles/egl.rs:406:19
   8: wgpu_hal::gles::device::<impl wgpu_hal::Device<wgpu_hal::gles::Api> for wgpu_hal::gles::Device>::destroy_buffer
             at /home/ashdnazg/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-hal-0.19.3/src/gles/device.rs:622:23
   9: <wgpu_core::device::resource::Device<A> as core::ops::drop::Drop>::drop
             at /home/ashdnazg/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-core-0.19.3/src/device/resource.rs:155:13
```

I already investigated this and found the underlying issue.

The GL backend's context gets a display using [eglGetPlatformDisplay](https://registry.khronos.org/EGL/sdk/docs/man/html/eglGetPlatformDisplay.xhtml), whose documentation specifies:
> Multiple calls made to eglGetPlatformDisplay with the same parameters will return the same EGLDisplay handle.

Later, it initialises it using [eglInitialize](https://registry.khronos.org/EGL/sdk/docs/man/html/eglInitialize.xhtml) which comes with the following fine print:
> Initializing an already initialized EGL display connection has no effect

When dropped, it calls [eglTerminate](https://registry.khronos.org/EGL/sdk/docs/man/html/eglTerminate.xhtml).

The first context to drop kills the display, causing all subsequent egl calls of the other contexts to fail, since they share the same display.

This doesn't happen in X11, since `eglGetPlatformDisplay` is called with a distinct `native_display` parameter for every new device.

**Repro steps**
```rust
    #[tokio::test]
    async fn it_crashes() {
        async fn default_device_and_queue() -> (wgpu::Device, wgpu::Queue) {
            let _ = env_logger::builder().is_test(true).try_init();
                backends: wgpu::Backends::GL,
                flags: wgpu::InstanceFlags::debugging(),
                ..Default::default()
            });

            let adapter = instance
                .request_adapter(&wgpu::RequestAdapterOptions::default())
                .await
                .unwrap();

            adapter.request_device(&wgpu::DeviceDescriptor::default(), None).await.unwrap()
        }

        let _ = vec![default_device_and_queue().await, default_device_and_queue().await];
        // The above line crashes.
    }
```

AFAIK it should be allowed to hold multiple devices/adapters.

**Platform**
Ubuntu 22.04, wgpu 0.19.3

ashdnazg:
**Connections**
Closes #5349 

**Description**
See #5349 for the crash.

Unfortunately the global state of EGL forces us to do our own global state for reference counting the displays.

**Testing**
Added a test that crashed before and passes now.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/wgpu-hal/src/gles/egl.rs
+++ b/wgpu-hal/src/gles/egl.rs
@@ -1,7 +1,8 @@
 use glow::HasContext;
+use once_cell::sync::Lazy;
 use parking_lot::{Mutex, MutexGuard, RwLock};
 
-use std::{ffi, os::raw, ptr, rc::Rc, sync::Arc, time::Duration};
+use std::{collections::HashMap, ffi, os::raw, ptr, rc::Rc, sync::Arc, time::Duration};
 
 /// The amount of time to wait while trying to obtain a lock to the adapter context
 const CONTEXT_LOCK_TIMEOUT_SECS: u64 = 1;
@@ -432,6 +433,45 @@ struct Inner {
     srgb_kind: SrgbFrameBufferKind,
 }
 
+// Different calls to `eglGetPlatformDisplay` may return the same `Display`, making it a global
+// state of all our `EglContext`s. This forces us to track the number of such context to prevent
+// terminating the display if it's currently used by another `EglContext`.
+static DISPLAYS_REFERENCE_COUNT: Lazy<Mutex<HashMap<usize, usize>>> = Lazy::new(Default::default);
+
+fn initialize_display(
+    egl: &EglInstance,
+    display: khronos_egl::Display,
+) -> Result<(i32, i32), khronos_egl::Error> {
+    let mut guard = DISPLAYS_REFERENCE_COUNT.lock();
+    *guard.entry(display.as_ptr() as usize).or_default() += 1;
+
+    // We don't need to check the reference count here since according to the `eglInitialize`
+    // documentation, initializing an already initialized EGL display connection has no effect
+    // besides returning the version numbers.
+    egl.initialize(display)
+}
+
+fn terminate_display(
+    egl: &EglInstance,
+    display: khronos_egl::Display,
+) -> Result<(), khronos_egl::Error> {
+    let key = &(display.as_ptr() as usize);
+    let mut guard = DISPLAYS_REFERENCE_COUNT.lock();
+    let count_ref = guard
+        .get_mut(key)
+        .expect("Attempted to decref a display before incref was called");
+
+    if *count_ref > 1 {
+        *count_ref -= 1;
+
+        Ok(())
+    } else {
+        guard.remove(key);
+
+        egl.terminate(display)
+    }
+}
+
 impl Inner {
     fn create(
         flags: wgt::InstanceFlags,
@@ -439,7 +479,7 @@ impl Inner {
         display: khronos_egl::Display,
         force_gles_minor_version: wgt::Gles3MinorVersion,
     ) -> Result<Self, crate::InstanceError> {
-        let version = egl.initialize(display).map_err(|e| {
+        let version = initialize_display(&egl, display).map_err(|e| {
             crate::InstanceError::with_source(
                 String::from("failed to initialize EGL display connection"),
                 e,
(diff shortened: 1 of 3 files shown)


ashdnazg:
There's a bit of a mess with the tests, apparently the regular `#[test]` functions are not tested in the CI.
I converted the device ones to gpu tests so at least they run, but it's kinda ugly with the context initialising its own adapter and device in advance.

wgpu-hal/src/gles/egl.rs
```diff
@@ -608,8 +636,11 @@ impl Drop for Inner {
...
             log::warn!("Error in destroy_context: {:?}", e);
         }
-        if let Err(e) = self.egl.instance.terminate(self.egl.display) {
-            log::warn!("Error in terminate: {:?}", e);
+
+        if decrement_display_reference_count(self.egl.display) {
+            if let Err(e) = self.egl.instance.terminate(self.egl.display) {
+                log::warn!("Error in terminate: {:?}", e);
+            }
         }
```
Wumpf: This code is racy. After getting a decrement to hit zero, another thread might call `create` and bump the count up for that very display and call `egl.initialize` which then followed by this thread calling `terminate`, thus leaving the thread that just initialized with a broken display
(The logic doesn't work out like an an `arc` since in an `arc` there's a single defined creation via "new" whereas here we "new" on every reference count bump)

You'll have to do a lot more aggressive locking here: Decrement + `instance.terminate` and Increment + `egl.initialize` each have to be fully protected by the same mutex.

wgpu-hal/src/gles/egl.rs
```diff
@@ -432,13 +433,40 @@ struct Inner {
...
+
 impl Inner {
     fn create(
         flags: wgt::InstanceFlags,
         egl: Arc<EglInstance>,
         display: khronos_egl::Display,
         force_gles_minor_version: wgt::Gles3MinorVersion,
     ) -> Result<Self, crate::InstanceError> {
+        increment_display_reference_count(display);
+
```
Wumpf: The need for this reference count needs to be documented here. In particular why it's fine to call `egl.initialize` repeatedly but it's not to do so with `terminate`

Wumpf (changes requested):
thank you and great you added a test to catch this issue as well!
The locking logic needs another pass though I believe (see comment)

> There's a bit of a mess with the tests, apparently the regular #[test] functions are not tested in the CI.

yes, all that gpu testing is a bit tricky on CI (:

wgpu-hal/src/gles/egl.rs
ashdnazg: :facepalm: nice catch, thank you!

ashdnazg:
Should be fixed now.

Wumpf (approved):
great stuff, thanks for documenting everything in detail now!

PR #5357, opened 2024-03-06
Naga: GLSL 410 does not support layout(binding = ...)

bes:
**Description**
Naga assumed that GLSL 410 supported layout(binding = ...) but it does not, it only supports layout(location = ...). It is not possible to enable only layout(location = ...) currently, so we need to predicate the feature on GLSL 420 instead.

**Testing**
Tested locally using Bevy on macOS with OpenGL 4.1.

**Checklist**

- [x] Run `cargo fmt`.


teoxoy (approved):
Thanks!

PR #5358, opened 2024-03-07
Invoke a DeviceLostClosure immediately if set on an invalid device.

bradwerth:
**Connections**
N/A

**Description**
This handles another case in `device_set_device_lost_closure` where a `DeviceLostClosure` is given to wgpu but never called. The contract with these closures is that wgpu *must* call them before they are dropped.

**Testing**
This adds a new test DEVICE_INVALID_THEN_SET_LOST_CALLBACK which invalidates the device and then sets a device lost callback on it. Unfortunately, there was no existing way to make a device invalid other than dropping it, so it also adds a function `make_invalid` to replace the device with an error in the registry. A few functions had to be touched up to handle this unexpected error object in the registry, representing an invalid device.

**Checklist**

- [X] Run `cargo fmt`.

### DIFF
--- a/tests/tests/device.rs
+++ b/tests/tests/device.rs
@@ -560,10 +560,7 @@ static DEVICE_DROP_THEN_LOST: GpuTestConfiguration = GpuTestConfiguration::new()
         // Set a LoseDeviceCallback on the device.
         let callback = Box::new(|reason, message| {
             WAS_CALLED.store(true, std::sync::atomic::Ordering::SeqCst);
-            assert!(
-                matches!(reason, wgt::DeviceLostReason::Dropped),
-                "Device lost info reason should match DeviceLostReason::Dropped."
-            );
+            assert_eq!(reason, wgt::DeviceLostReason::Dropped);
             assert_eq!(message, "Device dropped.");
         });
         ctx.device.set_device_lost_callback(callback);
@@ -594,10 +591,7 @@ static DEVICE_INVALID_THEN_SET_LOST_CALLBACK: GpuTestConfiguration = GpuTestConf
         // Set a LoseDeviceCallback on the device.
         let callback = Box::new(|reason, _m| {
             WAS_CALLED.store(true, std::sync::atomic::Ordering::SeqCst);
-            assert!(
-                matches!(reason, wgt::DeviceLostReason::DeviceInvalid),
-                "Device lost info reason should match DeviceLostReason::DeviceInvalid."
-            );
+            assert_eq!(reason, wgt::DeviceLostReason::DeviceInvalid);
         });
         ctx.device.set_device_lost_callback(callback);
 
@@ -618,10 +612,7 @@ static DEVICE_LOST_REPLACED_CALLBACK: GpuTestConfiguration = GpuTestConfiguratio
         // Set a LoseDeviceCallback on the device.
         let callback = Box::new(|reason, _m| {
             WAS_CALLED.store(true, std::sync::atomic::Ordering::SeqCst);
-            assert!(
-                matches!(reason, wgt::DeviceLostReason::ReplacedCallback),
-                "Device lost info reason should match DeviceLostReason::ReplacedCallback."
-            );
+            assert_eq!(reason, wgt::DeviceLostReason::ReplacedCallback);
         });
         ctx.device.set_device_lost_callback(callback);
 
@@ -650,10 +641,7 @@ static DROPPED_GLOBAL_THEN_DEVICE_LOST: GpuTestConfiguration = GpuTestConfigurat
         // Set a LoseDeviceCallback on the device.
         let callback = Box::new(|reason, message| {
             WAS_CALLED.store(true, std::sync::atomic::Ordering::SeqCst);
-            assert!(
-                matches!(reason, wgt::DeviceLostReason::Dropped),
-                "Device lost info reason should match DeviceLostReason::Dropped."
-            );
+            assert_eq!(reason, wgt::DeviceLostReason::Dropped);
             assert_eq!(message, "Device is dying.");
         });
         ctx.device.set_device_lost_callback(callback);
--- a/tests/tests/device.rs
+++ b/tests/tests/device.rs
@@ -564,10 +564,7 @@ static DEVICE_DROP_THEN_LOST: GpuTestConfiguration = GpuTestConfiguration::new()
                 matches!(reason, wgt::DeviceLostReason::Dropped),
                 "Device lost info reason should match DeviceLostReason::Dropped."
             );
-            assert!(
-                message == "Device dropped.",
-                "Device lost info message should be \"Device dropped.\"."
-            );
+            assert_eq!(message, "Device dropped.");
         });
         ctx.device.set_device_lost_callback(callback);
 
@@ -657,11 +654,7 @@ static DROPPED_GLOBAL_THEN_DEVICE_LOST: GpuTestConfiguration = GpuTestConfigurat
                 matches!(reason, wgt::DeviceLostReason::Dropped),
                 "Device lost info reason should match DeviceLostReason::Dropped."
             );
-            assert!(
-                message == "Device is dying.",
-                "Device lost info message is \"{}\" and it should be \"Device is dying.\".",
-                message
-            );
+            assert_eq!(message, "Device is dying.");
         });
         ctx.device.set_device_lost_callback(callback);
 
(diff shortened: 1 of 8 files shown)


tests/tests/device.rs
```diff
@@ -550,6 +550,37 @@ static DEVICE_DROP_THEN_LOST: GpuTestConfiguration = GpuTestConfiguration::new()
...
+    .run_sync(|ctx| {
+        // This test checks that when the device is invalid, a subsequent call
+        // to set the device lost callback will immediately call the callback.
+        // Invalidating the device is done via a testing-only method. Fails on
+        // webgl because webgl doesn't implement make_invalid.
+
+        // Make the device invalid.
+        ctx.device.make_invalid();
+
+        let was_called = std::sync::Arc::<std::sync::atomic::AtomicBool>::new(false.into());
```
teoxoy:
```suggestion
        let was_called = std::sync::atomic::AtomicBool::new(false);
```

wgpu-core/src/device/global.rs
```diff
@@ -2224,6 +2224,15 @@ impl Global {
         }
     }
 
+    // This is a test-only function to force the device into an
+    // invalid state by inserting an error value in its place in
+    // the registry.
+    pub fn device_make_invalid<A: HalApi>(&self, device_id: DeviceId) {
```
teoxoy, line 2246:
```suggestion
    /// This is a test-only function to force the device into an
    /// invalid state by inserting an error value in its place in
    /// the registry.
    #[cfg(test)]
    pub fn device_make_invalid<A: HalApi>(&self, device_id: DeviceId) {
```

wgpu/src/backend/webgpu.rs
```diff
@@ -1948,6 +1948,10 @@ impl crate::context::Context for ContextWebGpu {
         create_identified(device_data.0.create_render_bundle_encoder(&mapped_desc))
     }
 
+    fn device_make_invalid(&self, _device: &Self::DeviceId, _device_data: &Self::DeviceData) {
```
teoxoy, line 1952:
```suggestion
    #[cfg(test)]
    fn device_make_invalid(&self, _device: &Self::DeviceId, _device_data: &Self::DeviceData) {
```

wgpu/src/backend/wgpu_core.rs
```diff
@@ -1346,14 +1346,16 @@ impl crate::Context for ContextWgpuCore {
             Err(e) => panic!("Error in Device::create_render_bundle_encoder: {e}"),
         }
     }
+    fn device_make_invalid(&self, device: &Self::DeviceId, _device_data: &Self::DeviceData) {
```
teoxoy, line 1350:
```suggestion
    #[cfg(test)]
    fn device_make_invalid(&self, device: &Self::DeviceId, _device_data: &Self::DeviceData) {
```

wgpu/src/lib.rs
```diff
@@ -2703,6 +2703,11 @@ impl Device {
             Box::new(callback),
         )
     }
+
+    /// Test-only function to make this device invalid.
+    pub fn make_invalid(&self) {
```
teoxoy:
```suggestion
    /// Test-only function to make this device invalid.
    #[cfg(test)]
    pub fn make_invalid(&self) {
```

wgpu/src/context.rs
```diff
@@ -267,6 +267,7 @@ pub trait Context: Debug + WasmNotSendSync + Sized {
         device_data: &Self::DeviceData,
         desc: &RenderBundleEncoderDescriptor<'_>,
     ) -> (Self::RenderBundleEncoderId, Self::RenderBundleEncoderData);
+    fn device_make_invalid(&self, device: &Self::DeviceId, device_data: &Self::DeviceData);
```
teoxoy, line 271:
```suggestion
    #[cfg(test)]
    fn device_make_invalid(&self, device: &Self::DeviceId, device_data: &Self::DeviceData);
```

wgpu/src/context.rs
```diff
@@ -1293,6 +1294,7 @@ pub(crate) trait DynContext: Debug + WasmNotSendSync {
         device_data: &crate::Data,
         desc: &RenderBundleEncoderDescriptor<'_>,
     ) -> (ObjectId, Box<crate::Data>);
+    fn device_make_invalid(&self, device: &ObjectId, device_data: &crate::Data);
```
teoxoy, line 1299:
```suggestion
    #[cfg(test)]
    fn device_make_invalid(&self, device: &ObjectId, device_data: &crate::Data);
```

wgpu/src/context.rs
```diff
@@ -2350,6 +2352,12 @@ where
         (render_bundle_encoder.into(), Box::new(data) as _)
     }
 
+    fn device_make_invalid(&self, device: &ObjectId, device_data: &crate::Data) {
```
teoxoy, line 2358:
```suggestion
    #[cfg(test)]
    fn device_make_invalid(&self, device: &ObjectId, device_data: &crate::Data) {
```

teoxoy (commented):
Looks good! I left a few suggestions, I can apply them if you think they are good.

tests/tests/device.rs
bradwerth: Sorry, can't get this to work. I need to make a clone of the bool for the closure, otherwise it gets moved into the closure and then I can't check it afterwards. It's possible I'm missing something about how to do this cleanly in Rust, but this change (alone) does not work.

bradwerth:
> Looks good! I left a few suggestions, I can apply them if you think they are good.

I would appreciate you applying these proposed changes, because I can't get them to work as-is, and my Rust-fighting skills are not strong enough to diagnose and fix the resulting errors.

tests/tests/device.rs
ErichDonGubler: I changed the `AtomicBool`s to be `static` items with 572efa1, e33c4c3, 6a3cf202c. They're `fixup!` commits, so you can squash 'em down with `git rebase --autosquash -i trunk` after you've had a gander. 🙂

ErichDonGubler:
Pushed up bb0331951 to abbreviate `assert!(a == b, …)` to `assert_eq!(a, b)`. `assert_eq` generally provides superior diagnostics, and, even better, it means you don't have to write your own failure text. 😀

wgpu-core/src/device/global.rs
ErichDonGubler: @teoxoy: This feedback (and others like it) couldn't ever work with an external test crate like `wgpu-test`. `wgpu-test` uses the public API of `wgpu`…so we need a different option.

In the most recent `wgpu maintainers meeting`, we [resolved](https://docs.google.com/document/d/1I3qh1kShDu48VehgFarftEpbEqZK8e8kWqaDo05e1jM/edit#heading=h.4boukitzny7l) to try just using `#[doc(hidden)]`. I've added this attribute to the suggested functions instead of `cfg(test)` with 76eec146f, which I think should resolve this and other feedback.

wgpu/src/backend/wgpu_core.rs
ErichDonGubler: Resolved; see <https://github.com/gfx-rs/wgpu/pull/5358#discussion_r1525688208>.

wgpu/src/context.rs
ErichDonGubler: Resolved; see <https://github.com/gfx-rs/wgpu/pull/5358#discussion_r1525688208>.

wgpu/src/context.rs
ErichDonGubler: Resolved; see <https://github.com/gfx-rs/wgpu/pull/5358#discussion_r1525688208>.

wgpu/src/context.rs
ErichDonGubler: Resolved; see <https://github.com/gfx-rs/wgpu/pull/5358#discussion_r1525688208>.

wgpu/src/lib.rs
ErichDonGubler: Resolved; see <https://github.com/gfx-rs/wgpu/pull/5358#discussion_r1525688208>.

wgpu/src/backend/webgpu.rs
ErichDonGubler: Resolved; see <https://github.com/gfx-rs/wgpu/pull/5358#discussion_r1525688208>.

ErichDonGubler:
Pushed [73d90fb](https://github.com/gfx-rs/wgpu/pull/5358/commits/73d90fb26f69403d42315289937ead6d651007fc) to further clean up tests being added with deriving `Eq` on `DeviceLostReason`, so we can leverage `assert_eq!(…)` s'more.

teoxoy (approved):

PR #5362, opened 2024-03-07
fix: move 0-dimension check for transfers to be after validation

ErichDonGubler:
**Connections**

Discovered with Firefox's [bug 1848866](https://bugzilla.mozilla.org/show_bug.cgi?id=1848866).

**Description**

**Testing**

**Checklist**

- [ ] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/command/transfer.rs
+++ b/wgpu-core/src/command/transfer.rs
@@ -649,12 +649,6 @@ impl Global {
             .into());
         }
 
-        if size == 0 {
-            log::trace!("Ignoring copy_buffer_to_buffer of size 0");
-            cmd_buf_data_guard.mark_successful();
-            return Ok(());
-        }
-
         // Make sure source is initialized memory and mark dest as initialized.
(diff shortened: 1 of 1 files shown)


teoxoy:
I don't think we have to remove these early returns, we just need to move them right before the calls that would execute the copies so that all the validation code still runs.

ErichDonGubler:
@teoxoy: Ah, yeah, I'd come to this same conclusion some time ago, but I haven't updated this draft yet. I'll change the title, and add a convo in my diff to not forget. 😅

wgpu-core/src/command/transfer.rs
```diff
@@ -704,11 +704,6 @@ impl Global {
             .into());
         }
 
-        if size == 0 {
-            log::trace!("Ignoring copy_buffer_to_buffer of size 0");
-            return Ok(());
-        }
-
```
ErichDonGubler: issue: This and the other hunk should be moved, not <em>re</em>moved.

PR #5370, opened 2024-03-09
GL actually supports DEPTH32FLOAT_STENCIL8

Dinnerbone:
**Connections**
n/a

**Description**
Noticed this recently, the GL backend didn't allow the use of `Depth32FloatStencil8` because the `DEPTH32FLOAT_STENCIL8` feature isn't claimed to be supported.

It actually is supported, so this fixes that and allows for the use of `Depth32FloatStencil8` on GLES/WebGL.

**Testing**
I manually changed the stencil test to use `Depth32FloatStencil8` and it works as expected.

The `wgpu_test::clear_texture::clear_texture_depth32_stencil8` test had to be configured to also require `wgpu::DownlevelFlags::DEPTH_TEXTURE_AND_BUFFER_COPIES` as that's not supported on the GLES backend, as otherwise that test tries to now run on GLES but fails.

**Checklist**

- [X] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,6 +122,7 @@ By @cwfitzgerald in [#5325](https://github.com/gfx-rs/wgpu/pull/5325).
 
 - Log an error when GLES texture format heuristics fail. By @PolyMeilex in [#5266](https://github.com/gfx-rs/wgpu/issues/5266)
 - Cache the sample count to keep `get_texture_format_features` cheap. By @Dinnerbone in [#5346](https://github.com/gfx-rs/wgpu/pull/5346)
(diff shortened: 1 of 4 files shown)


Wumpf (approved):
For reference:
* GLES3 documentation lists the format here https://registry.khronos.org/OpenGL-Refpages/es3.0/html/glTexImage2D.xhtml
* WebGL renderbufferStorage is documented to take `gl.DEPTH32F_STENCIL8` here https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/renderbufferStorage

So yeah, looks like an oversight!
Thanks for all this bite sized (Web)GL but great improvements @Dinnerbone ! 🙇

PR #5371, opened 2024-03-10
Minor rewording of ConfigureSurfaceError::TooLarge

rukai:
I hit this error:
`Surface width and height must be within the maximum supported texture size. Requested was (3840, 2119), maximum extent is 2048.`
I found the error confusing since its not clear how a single dimensional maximum should be used for a 2 dimensional value.

So this PR adjusts the error to read like:
`Surface width and height must be within the maximum supported texture size. Requested was (3840, 2119), maximum extent for either dimension is 2048.`

Alternatively we could do:
`Surface width and height must be within the maximum supported texture size. Requested was (3840, 2119), maximum extent is (2048, 2048).`
But I think that would not be as good since it implies that its possible for the maximum width and maximum height to differ.


Wumpf (approved):
seems better indeed 👍

PR #5378, opened 2024-03-11
fix: add missing deferred resource destruction in `Global::poll_all_devices`

ErichDonGubler:
**Connections**

* [Firefox's bug 1863872](https://bugzilla.mozilla.org/show_bug.cgi?id=1863872) is a result of this issue.
* This bug was introduced with #5216.

**Description**

#5216 
While debugging [Firefox's bug 1863872](https://bugzilla.mozilla.org/show_bug.cgi?id=1863872), @jimblandy and @nical discovered with @sotaroikeda's help that `TextureView`s appeared to be cleaned up far later than we expected. @nical had recently changed the path of clean-up for `TextureView`s with deferring until there was no lock contention. That PR added deferred resource clean-up to `Global::device_poll`, but, notably, _not_ `Global::poll_all_devices`, which was certainly intended.

With this change, we can no longer reproduce bug 1863872, and we can confirm getting `TextureView` clean-up when we expect it again.

**Testing**

TODO: write testing

**Checklist**

- [x] Run `cargo fmt`.
- [x] ~~Add change to `CHANGELOG.md`. See simple instructions inside file.~~

### DIFF
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -2151,6 +2151,10 @@ impl Global {
                 all_queue_empty = all_queue_empty && queue_empty;
 
                 closures.extend(cbs);
+
+                // Some deferred destroys are scheduled in maintain so run this right after
+                // to avoid holding on to them until the next device poll.
+                device.deferred_resource_destruction();
             }
         }
 
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -2097,16 +2097,16 @@ impl Global {
             .get(device_id)
             .map_err(|_| DeviceError::Invalid)?;
 
-        let (closures, queue_empty) = {
-            if let wgt::Maintain::WaitForSubmissionIndex(submission_index) = maintain {
-                if submission_index.queue_id != device_id.transmute() {
-                    return Err(WaitIdleError::WrongSubmissionIndex(
-                        submission_index.queue_id,
-                        device_id,
-                    ));
-                }
+        if let wgt::Maintain::WaitForSubmissionIndex(submission_index) = maintain {
+            if submission_index.queue_id != device_id.transmute() {
+                return Err(WaitIdleError::WrongSubmissionIndex(
+                    submission_index.queue_id,
+                    device_id,
+                ));
             }
+        }
 
+        let (closures, queue_empty) = {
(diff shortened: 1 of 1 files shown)


wgpu-core/src/device/global.rs
ErichDonGubler, line 1: I'm not sure if this needs a `CHANGELOG` entry. I don't see any descendant commits of 499bd64c292b2cd436b8c6aceafcf that have tags, ATM, and, IMO, that would be the only clear indicator that this would need one (since it would mean we'd released with buggy behavior).

ErichDonGubler, line 1: todo: write testing

wgpu-core/src/device/global.rs
```diff
@@ -2151,6 +2151,10 @@ impl Global {
                 all_queue_empty = all_queue_empty && queue_empty;
 
                 closures.extend(cbs);
+
+                // Some deferred destroys are scheduled in maintain so run this right after
+                // to avoid holding on to them until the next device poll.
+                device.deferred_resource_destruction();
```
ErichDonGubler: - [x] issue: This should be DRY. The reason we ran into this bug is because we're duplicating essential logic that we will need whenever we poll _any_ device, independent of whether it is done through `device_poll` and `poll_all_devices`.

nical (approved):

wgpu-core/src/device/global.rs
ErichDonGubler: @cwfitzgerald, @jimblandy: I'm going to plan on foregoing a `CHANGELOG` entry here. I'm happy to compose one post-merge if you feel it's really necessary.

ErichDonGubler: I don't see an easy way to test for resource clean-up with these APIs right now. 🫤 Asking for help [in Matrix](https://matrix.to/#/!FZyQrssSlHEZqrYcOb:matrix.org/$55AS31XckMVQv5FZwoXpMdPkK9wiVD9t-axDvW6Y1ps?via=matrix.org&via=mozilla.org&via=envs.net); I think I'm gonna punt on this. Will file an issue so we don't lose track of it.

ErichDonGubler: Filed https://github.com/gfx-rs/wgpu/issues/5381.

PR #5383, opened 2024-03-13
Add 64 bit atomics

atlv24:
**Connections**
Others in this series: #5155 #5154
Addresses #4424
Depends on #5498 and #5497

**Description**
Shaders currently do not support 64 bit integer atomics. Add support for these gated by a feature flag when the capability is present.

Metal 64bit atomics are quite different from VK/DX style atomics, they do not have a return/output value of any kind and are only available in min/max variants. To accommodate this, a separate feature was added for min/max 64 bit atomics, and AtomicFunctionNoReturn was created in the naga IR to represent these return-less atomics.

DirectX ShaderModel version detection was also implemented, as SM 6.0 was insufficient for this feature.

**Testing**
A couple snapshot tests have been added to test int64 atomic operations.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,8 +81,41 @@ for message in compilation_info
 
 By @stefnotch in [#5410](https://github.com/gfx-rs/wgpu/pull/5410)
 
-### New features
+#### 64 bit integer atomic support in shaders.
 
+Add support for 64 bit integer atomic operations in shaders.
+
+Add the following flags to `wgpu_types::Features`:
+
+- `SHADER_INT64_ATOMIC_ALL_OPS` enables all atomic operations on `atomic<i64>` and
+  `atomic<u64>` values.
+
+- `SHADER_INT64_ATOMIC_MIN_MAX` is a subset of the above, enabling only
+  `AtomicFunction::Min` and `AtomicFunction::Max` operations on `atomic<i64>` and
+  `atomic<u64>` values in the `Storage` address space. These are the only 64-bit
+  atomic operations available on Metal as of 3.1.
+
+Add corresponding flags to `naga::valid::Capabilities`. These are supported by the
+WGSL front end, and all Naga backends.
+
+Platform support:
+
+- On Direct3d 12, in `D3D12_FEATURE_DATA_D3D12_OPTIONS9`, if
+  `AtomicInt64OnTypedResourceSupported` and `AtomicInt64OnGroupSharedSupported` are
+  both available, then both wgpu features described above are available.
+
+- On Metal, `SHADER_INT64_ATOMIC_MIN_MAX` is available on Apple9 hardware, and on
+  hardware that advertises both Apple8 and Mac2 support. This also requires Metal
+  Shading Language 2.4 or later. Metal does not yet support the more general
+  `SHADER_INT64_ATOMIC_ALL_OPS`.
+
+- On Vulkan, if the `VK_KHR_shader_atomic_int64` extension is available with both the
+  `shader_buffer_int64_atomics` and `shader_shared_int64_atomics` features, then both
+  wgpu features described above are available.
+
+By @atlv24 in [#5383](https://github.com/gfx-rs/wgpu/pull/5383)
+
+### New features
 #### Vulkan
 
 - Added a `PipelineCache` resource to allow using Vulkan pipeline caches. By @DJMcNab in [#5319](https://github.com/gfx-rs/wgpu/pull/5319)
--- a/naga/src/lib.rs
+++ b/naga/src/lib.rs
@@ -1612,8 +1612,29 @@ pub enum Expression {
     },
     /// Result of calling another function.
     CallResult(Handle<Function>),
+
     /// Result of an atomic operation.
+    ///
+    /// This expression must be referred to by the [`result`] field of exactly one
+    /// [`Atomic`][stmt] statement somewhere in the same function. Let `T` be the
+    /// scalar type contained by the [`Atomic`][type] value that the statement
+    /// operates on.
+    ///
+    /// If `comparison` is `false`, then `ty` must be the scalar type `T`.
+    ///
+    /// If `comparison` is `true`, then `ty` must be a [`Struct`] with two members:
+    ///
+    /// - A member named `old_value`, whose type is `T`, and
+    ///
+    /// - A member named `exchanged`, of type [`BOOL`].
+    ///
+    /// [`result`]: Statement::Atomic::result
+    /// [stmt]: Statement::Atomic
+    /// [type]: TypeInner::Atomic
+    /// [`Struct`]: TypeInner::Struct
+    /// [`BOOL`]: Scalar::BOOL
     AtomicResult { ty: Handle<Type>, comparison: bool },
+
     /// Result of a [`WorkGroupUniformLoad`] statement.
     ///
     /// [`WorkGroupUniformLoad`]: Statement::WorkGroupUniformLoad
@@ -1890,15 +1911,66 @@ pub enum Statement {
     /// Atomic function.
     Atomic {
         /// Pointer to an atomic value.
+        ///
+        /// This must be a [`Pointer`] to an [`Atomic`] value. The atomic's
+        /// scalar type may be [`I32`] or [`U32`].
+        ///
+        /// If [`SHADER_INT64_ATOMIC_MIN_MAX`] or [`SHADER_INT64_ATOMIC_ALL_OPS`] are
+        /// enabled, this may also be [`I64`] or [`U64`].
+        ///
+        /// [`Pointer`]: TypeInner::Pointer
+        /// [`Atomic`]: TypeInner::Atomic
+        /// [`I32`]: Scalar::I32
+        /// [`U32`]: Scalar::U32
+        /// [`SHADER_INT64_ATOMIC_MIN_MAX`]: crate::valid::Capabilities::SHADER_INT64_ATOMIC_MIN_MAX
+        /// [`SHADER_INT64_ATOMIC_ALL_OPS`]: crate::valid::Capabilities::SHADER_INT64_ATOMIC_ALL_OPS
+        /// [`I64`]: Scalar::I64
+        /// [`U64`]: Scalar::U64
         pointer: Handle<Expression>,
-        /// Function to run on the atomic.
+
+        /// Function to run on the atomic value.
+        ///
+        /// If [`pointer`] refers to a 64-bit atomic value, then:
+        ///
+        /// - The [`SHADER_INT64_ATOMIC_ALL_OPS`] capability allows any [`AtomicFunction`]
+        ///   value here.
+        ///
+        /// - The [`SHADER_INT64_ATOMIC_MIN_MAX`] capability allows
+        ///   [`AtomicFunction::Min`] and [`AtomicFunction::Max`] here.
+        ///
+        /// - If neither of those capabilities are present, then 64-bit scalar
+        ///   atomics are not allowed.
+        ///
+        /// [`pointer`]: Statement::Atomic::pointer
+        /// [`SHADER_INT64_ATOMIC_MIN_MAX`]: crate::valid::Capabilities::SHADER_INT64_ATOMIC_MIN_MAX
+        /// [`SHADER_INT64_ATOMIC_ALL_OPS`]: crate::valid::Capabilities::SHADER_INT64_ATOMIC_ALL_OPS
         fun: AtomicFunction,
+
         /// Value to use in the function.
+        ///
+        /// This must be a scalar of the same type as [`pointer`]'s atomic's scalar type.
+        ///
+        /// [`pointer`]: Statement::Atomic::pointer
         value: Handle<Expression>,
+
         /// [`AtomicResult`] expression representing this function's result.
         ///
+        /// If [`fun`] is [`Exchange { compare: None }`], this must be `Some`,
+        /// as otherwise that operation would be equivalent to a simple [`Store`]
+        /// to the atomic.
+        ///
+        /// Otherwise, this may be `None` if the return value of the operation is not needed.
+        ///
+        /// If `pointer` refers to a 64-bit atomic value, [`SHADER_INT64_ATOMIC_MIN_MAX`]
+        /// is enabled, and [`SHADER_INT64_ATOMIC_ALL_OPS`] is not, this must be `None`.
+        ///
         /// [`AtomicResult`]: crate::Expression::AtomicResult
-        result: Handle<Expression>,
+        /// [`fun`]: Statement::Atomic::fun
+        /// [`Store`]: Statement::Store
+        /// [`Exchange { compare: None }`]: AtomicFunction::Exchange
+        /// [`SHADER_INT64_ATOMIC_MIN_MAX`]: crate::valid::Capabilities::SHADER_INT64_ATOMIC_MIN_MAX
+        /// [`SHADER_INT64_ATOMIC_ALL_OPS`]: crate::valid::Capabilities::SHADER_INT64_ATOMIC_ALL_OPS
+        result: Option<Handle<Expression>>,
     },
     /// Load uniformly from a uniform pointer in the workgroup address space.
     ///
--- a/naga/src/valid/mod.rs
+++ b/naga/src/valid/mod.rs
@@ -127,6 +127,18 @@ bitflags::bitflags! {
         const SUBGROUP = 0x10000;
         /// Support for subgroup barriers.
         const SUBGROUP_BARRIER = 0x20000;
+        /// Support for [`AtomicFunction::Min`] and [`AtomicFunction::Max`] on
+        /// 64-bit integers in the [`Storage`] address space, when the return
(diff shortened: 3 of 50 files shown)


wgpu-hal/src/dx12/adapter.rs
```diff
@@ -311,6 +311,23 @@ impl super::Adapter {
...
+            let mut features9: crate::dx12::types::D3D12_FEATURE_DATA_D3D12_OPTIONS9 =
+                unsafe { mem::zeroed() };
+            let hr = unsafe {
+                device.CheckFeatureSupport(
+                    37, // D3D12_FEATURE_D3D12_OPTIONS9
+                    &mut features9 as *mut _ as *mut _,
+                    mem::size_of::<crate::dx12::types::D3D12_FEATURE_DATA_D3D12_OPTIONS9>() as _,
+                )
+            };
+            hr == 0 && features9.AtomicInt64OnTypedResourceSupported != 0
```
cwfitzgerald: I don't think this check is right.

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -351,6 +355,17 @@ impl PhysicalDeviceFeatures {
             } else {
                 None
             },
+            shader_int64_atomic: if device_api_version >= vk::API_VERSION_1_1
```
cwfitzgerald: This seems to be checking if image atomics are supported, not if regular 64bit atomics are supported.

naga/tests/in/atomicCompareExchange-int64.wgsl
cwfitzgerald, line 1: These shaders need to be regenerated.

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -1458,6 +1483,11 @@ impl super::Adapter {
                 capabilities.push(spv::Capability::Int64);
             }
 
+            if features.contains(wgt::Features::SHADER_INT64_ATOMIC) {
+                capabilities.push(spv::Capability::Int64ImageEXT);
```
cwfitzgerald: Same wrt images.

wgpu-hal/src/dx12/adapter.rs
cwfitzgerald: I'm asking for clarification on the features.

cwfitzgerald (changes requested):

wgpu-hal/src/dx12/adapter.rs
cwfitzgerald: To quote Jesse:

> - Raw buffers are required to be supported by SM6.6 without any additional optional caps.
> - A typed resource would be a RWBuffer<sometype> or RWTexture2D<sometype>
> - A descriptor heap resource is one accessed via ResourceDescriptorHeap[index]

So if we just want buffer and group shared (which is what webgpu supports) we want to check only `AtomicInt64OnGroupSharedSupported`

naga/src/valid/mod.rs
```diff
@@ -110,6 +110,8 @@ bitflags::bitflags! {
         const CUBE_ARRAY_TEXTURES = 0x4000;
         /// Support for 64-bit signed and unsigned integers.
         const SHADER_INT64 = 0x8000;
+        /// Support for 64-bit signed and unsigned integers.
```
JMS55: Copy pasted comment, should mention atomics :)

wgpu-types/src/lib.rs
```diff
@@ -883,6 +883,15 @@ bitflags::bitflags! {
         ///
         /// This is a native only feature.
         const SHADER_INT64 = 1 << 55;
+        /// Allows shaders to use i64 and u64 atomics.
+        ///
+        /// Supported platforms:
+        /// - Vulkan
+        /// - DX12 (DXC only)
```
JMS55: Should maybe mention the VK extension required, and the DXC shader model/feature/thing? Idk how widespread support is / how new this is.

CHANGELOG.md
```diff
@@ -99,7 +99,8 @@ Bottom level categories:
   - As with other instance flags, this flag can be changed in calls to `InstanceFlags::with_env` with the new `WGPU_GPU_BASED_VALIDATION` environment variable.
 
   By @ErichDonGubler in [#5146](https://github.com/gfx-rs/wgpu/pull/5146), [#5046](https://github.com/gfx-rs/wgpu/pull/5046).
-- Signed and unsigned 64 bit integer support in shaders. By @rodolphito and @cwfitzgerald in [#5154](https://github.com/gfx-rs/wgpu/pull/5154)
+- Signed and unsigned 64 bit integer support in shaders. By @atlv24 and @cwfitzgerald in [#5154](https://github.com/gfx-rs/wgpu/pull/5154)
+- 64 bit integer atomic support in shaders. By @atlv24 in [#5383](https://github.com/gfx-rs/wgpu/pull/5383)
```
atlv24: i've changed my github username from rodolphito to atlv24

wgpu-core/src/device/resource.rs
```diff
@@ -1525,6 +1525,16 @@ impl<A: HalApi> Device<A> {
...
         );
+        caps.set(
+            Caps::SHADER_INT64_ATOMIC_MIN_MAX,
+            self.features
+                .contains(wgt::Features::SHADER_INT64_ATOMIC_MIN_MAX),
+        );
+        caps.set(
+            Caps::SHADER_INT64_ATOMIC_ALL_OPS,
+            self.features
+                .contains(wgt::Features::SHADER_INT64_ATOMIC_ALL_OPS),
```
cwfitzgerald: All on the name here feels weird considering you need the other feature to turn on Min/Max

cwfitzgerald (changes requested):
Looking good from the wgpu side, small comments.

There should be at least one execution test that makes sure everything at least seems to behave correctly and gets through all validation layers and such. Check calling atomicAdd N times results in a value of N, etc.

wgpu-core/src/device/resource.rs
atlv24: `SHADER_INT64_ATOMIC_ALL_OPS` does include min/max, though

atlv24: I made the code more explicit

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -448,6 +454,17 @@ impl PhysicalDeviceFeatures {
             } else {
                 None
             },
+            shader_int64_atomic: if device_api_version >= vk::API_VERSION_1_1
+                || enabled_extensions.contains(&vk::KhrGetPhysicalDeviceProperties2Fn::name())
+            {
+                Some(
+                    vk::PhysicalDeviceShaderAtomicInt64Features::builder()
+                        .shader_buffer_int64_atomics(true)
```
cwfitzgerald: This is checking for VK_KHR_get_physical_device_properties2 not VK_KHR_shader_atomic_int64?

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -109,6 +109,9 @@ pub struct PhysicalDeviceFeatures {
     zero_initialize_workgroup_memory:
         Option<vk::PhysicalDeviceZeroInitializeWorkgroupMemoryFeatures>,
 
+    /// Features provided by `VK_KHR_shader_atomic_int64`, promoted to Vulkan 1.3.
```
cwfitzgerald:
```suggestion
    /// Features provided by `VK_KHR_shader_atomic_int64`, promoted to Vulkan 1.2.
```

cwfitzgerald (dismissed):

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -110,6 +110,9 @@ pub struct PhysicalDeviceFeatures {
     zero_initialize_workgroup_memory:
         Option<vk::PhysicalDeviceZeroInitializeWorkgroupMemoryFeatures<'static>>,
 
+    /// Features provided by `VK_KHR_shader_atomic_int64`, promoted to Vulkan 1.2.
+    shader_int64_atomic: Option<vk::PhysicalDeviceShaderAtomicInt64Features<'static>>,
```
jimblandy: Please rename this `shader_atomic_int64`. I recognize that this matches the `wgpu_types::Features` name, but generally the `PhysicalDeviceFeatures` field for `vk::PhysicalDeviceMumbleFeatures` is named `mumble`, and I think that's the stronger precedent here.

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -423,6 +429,16 @@ impl PhysicalDeviceFeatures {
             } else {
                 None
             },
+            shader_int64_atomic: if device_api_version >= vk::API_VERSION_1_2
+                || enabled_extensions.contains(&khr::shader_atomic_int64::NAME)
+            {
+                Some(
+                    vk::PhysicalDeviceShaderAtomicInt64Features::default()
+                        .shader_buffer_int64_atomics(true),
```
jimblandy: Shouldn't this consult `requested_features`, rather than hard-wiring `true`?

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -563,6 +579,11 @@ impl PhysicalDeviceFeatures {
         features.set(F::SHADER_INT64, self.core.shader_int64 != 0);
         features.set(F::SHADER_I16, self.core.shader_int16 != 0);
 
+        features.set(
+            F::SHADER_INT64_ATOMIC_ALL_OPS | F::SHADER_INT64_ATOMIC_MIN_MAX,
+            caps.supports_extension(khr::shader_atomic_int64::NAME),
+        );
+
```
jimblandy: I think this isn't correct. Simply because a Vulkan physical device supports the `VK_KHR_shader_atomic_int64` extension, it doesn't follow that it actually supports 64-bit atomic operations. The Vulkan design is such that the Mumble extension could be present, but then the `VkPhysicalDeviceMumbleFeature` fields could still be `false`.

I think you need to handle this more like `F::MULTIVIEW` later in this function: check whether `self.shader_int64_atomic` is `Some`, and if so, set the `wgpu_types::Features` according to the fields in `vk::PhysicalDeviceShaderAtomicInt64Features`.

naga/src/lib.rs
```diff
@@ -1107,6 +1107,25 @@ pub enum AtomicFunction {
...
+
+impl AtomicFunctionNoReturn {
+    pub const fn with_return(self) -> AtomicFunction {
+        match self {
+            Self::Min => AtomicFunction::Min,
+            Self::Max => AtomicFunction::Max,
+        }
+    }
+}
+
```
jimblandy: This function probably belongs in `naga/src/proc/mod.rs`, not in `lib.rs`.

naga/src/valid/mod.rs
```diff
@@ -114,6 +114,10 @@ bitflags::bitflags! {
         const SUBGROUP = 0x10000;
         /// Support for subgroup barriers.
         const SUBGROUP_BARRIER = 0x20000;
+        /// Support for 64-bit integer atomics.
+        const SHADER_INT64_ATOMIC_MIN_MAX = 0x40000;
+        /// Support for 64-bit integer atomics.
+        const SHADER_INT64_ATOMIC_ALL_OPS = 0x80000;
```
jimblandy: These two comments aren't even different. Could you provide more detail?

jimblandy (changes requested):
Still reviewing, this is just the stuff I've noticed so far.

naga/src/lib.rs
```diff
@@ -1900,6 +1919,15 @@ pub enum Statement {
...
     },
+    /// Atomic function with no return value.
+    AtomicNoReturn {
+        /// Pointer to an atomic value.
+        pointer: Handle<Expression>,
+        /// Function to run on the atomic.
+        fun: AtomicFunctionNoReturn,
+        /// Value to use in the function.
+        value: Handle<Expression>,
+    },
```
jimblandy: Let's get some proper documentation on this. What does it do? What types does it support? What address spaces are permitted under which features? Why is it different from `Atomic`?

I've made one attempt at reading through the code in `valid/function.rs`, but I'm still struggling to understand exactly which behaviors this statement supports and does not support. Ideally the documentation in `lib.rs` should fully explain the behavior and constraints of each statement.

jimblandy (changes requested):
Before I review this further I'd like to see the doc comments filled in enough.

jimblandy:
@atlv24 Thanks very much for the new comments - these are helpful.

jimblandy:
I wanted to make some changes to this PR, but in the process came across a few things that need to be fixed first. #5735 was my misunderstanding, but I think #5741 and #5763 need to be addressed. Those are both in progress.

jimblandy:
@atlv24 I just pushed a new commit to this branch; the log message is below. Could you look it over and see what you think?

> Make `Statement::Atomic::result` optional.
> 
> Rather than introducing an entirely new statement variant `AtomicNoResult`, make `Statement::Atomic::result` an `Option`. This reduces the patch size [edit] a bit, and has no effect on snapshot output other than the direct IR dumps.
> 
> In HLSL output, don't bother to generated "discard" temporaries for functions where the final `original_value` parameter is not required. This improves snapshot output.
> 
> Improve documentation for `Expression::AtomicResult`, `Statement::Atomic`, and the new `valid::Capabilities` flags.
> 
> Rewrite `Atomic` statement validation. Consolidate validation of `AtomicResult` expressions with this; #5771 ensures that there is indeed some `Atomic` statement referring to every `AtomicResult` expression, so we can be sure it will be validated.
> 
> Fixes #5742.

jimblandy:
(No offense meant by removing your docs, after asking you to write them - they were helpful in getting started!)

jimblandy:
> but I think #5741 and #5763 need to be addressed.

#5741 is fixed. #5763 is not, but I think fixing #5742 should be good enough for now.

jimblandy:
Merged trunk so CI would pay attention

jimblandy (approved):
If my changes look good, then the rest of this looks good to me.

wgpu-hal/src/dx12/adapter.rs
```diff
@@ -354,6 +354,23 @@ impl super::Adapter {
...
+            let mut features9: crate::dx12::types::D3D12_FEATURE_DATA_D3D12_OPTIONS9 =
+                unsafe { mem::zeroed() };
+            let hr = unsafe {
+                device.CheckFeatureSupport(
+                    37, // D3D12_FEATURE_D3D12_OPTIONS9
+                    &mut features9 as *mut _ as *mut _,
+                    mem::size_of::<crate::dx12::types::D3D12_FEATURE_DATA_D3D12_OPTIONS9>() as _,
+                )
+            };
+            hr == 0 && features9.AtomicInt64OnGroupSharedSupported != 0
```
teoxoy: Shouldn't we check for `AtomicInt64OnTypedResourceSupported` as well?

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -559,6 +578,13 @@ impl PhysicalDeviceFeatures {
         features.set(F::SHADER_INT64, self.core.shader_int64 != 0);
         features.set(F::SHADER_I16, self.core.shader_int16 != 0);
 
+        if let Some(ref shader_atomic_int64) = self.shader_atomic_int64 {
+            features.set(
+                F::SHADER_INT64_ATOMIC_ALL_OPS | F::SHADER_INT64_ATOMIC_MIN_MAX,
+                shader_atomic_int64.shader_buffer_int64_atomics != 0,
```
teoxoy: Shouldn't we check for `shaderSharedInt64Atomics` as well?

teoxoy (commented):
Took a brief look at the feature detection code.

wgpu-hal/src/vulkan/adapter.rs
jimblandy: > - `shaderSharedInt64Atomics` indicates whether shaders can perform 64-bit unsigned and signed integer atomic operations on shared and payload memory.

Does "shared" mean WGSL `var<workgroup>`? What is "payload"?

/me searches the spec frantically

teoxoy: > Does "shared" mean WGSL `var<workgroup>`?

That was my read, D3D12's `AtomicInt64OnGroupSharedSupported` is more explicit.

> `AtomicInt64OnGroupSharedSupported` is a boolean that specifies whether 64-bit integer atomics are supported on `groupshared` variables.

> What is "payload"?

I have no idea 😅

jimblandy: I did not succeed in finding clearly relevant definitions of these terms.

jimblandy:
> If my changes look good, then the rest of this looks good to me.

@atlv24 This is Just waiting for your comments on my commit, and for you to look over @teoxoy's review.

atlv24:
@jimblandy I reviewed, looks great! I wish I had thought of using Option myself haha. Would have saved me a lot of copy pasting :)

Issue #4366, opened 2021-09-27
Pipeline-overridable constants


kvark:
Related to gfx-rs/naga#647
Required by https://github.com/gfx-rs/wgpu/issues/1762

MikhailGorobets:
Any updates?

teoxoy:
I recently started to work on it.

jimblandy:
See the `gfx-rs/pipeline-constants` branch for the work in progress.

Issue #1762, opened 2021-08-03
Pipeline overrides


kvark:
**Is your feature request related to a problem? Please describe.**
See https://github.com/gpuweb/gpuweb/pull/886

**Describe the solution you'd like**
Just follow the spec

**Describe alternatives you've considered**
?

**Additional context**

kvark:
Adding a milestone tentatively, let's see how difficult this is going to be.

kvark:
Too much work for this milestone, actually

aloucks:
Any update on this? It would be really nice to avoid string replacement on the shader source.

It looks like the syntax has changed since the linked PR:
https://gpuweb.github.io/gpuweb/#dictdef-gpuprogrammablestage

kvark:
Yep, that's why I'm not in a rush. The topic has been actively discussed recently. I think the main approach is figured out now, and just some details are missing.

adeline-sparks:
I'd really like to use this too! Do you think the spec is stable enough to implement it? I have some free time upcoming and might give it a try if so.

teoxoy:
@adeline-sparks I'm working on it.

PR #5387, opened 2024-03-14
Implement override-expression evaluation in functions

teoxoy:


jimblandy:
I think the `try_eval_and_append` refactor is a good improvement.

jimblandy:
@teoxoy I've pushed some new commits to the branch. Could you look them over?

I'm not sure about "Hoist ConstantEvaluator". It does make the loop body simpler, but maybe it's nice to make it explicit that these `ConstantEvaluator`s don't retain any state from one iteration to the next?

naga/src/back/pipeline_constants.rs
```diff
@@ -237,6 +250,72 @@ fn process_override(
...
+    let mut adjusted_local_expressions = Vec::with_capacity(function.expressions.len());
+
+    let mut local_expression_kind_tracker = crate::proc::ExpressionKindTracker::new();
+
+    let mut expressions = mem::take(&mut function.expressions);
+
+    // Dummy `emitter` and `block` for the constant evaluator.
+    // We can ignore the concept of emitting expressions here since
+    // expressions have already been covered by a `Statement::Emit`
+    // in the frontend.
```
jimblandy, line 278: Super-helpful. At the exact point where I'm thinking, "What in the world?", the comment points out to me what I was missing, and I can just move on.

naga/src/back/pipeline_constants.rs
```diff
@@ -237,6 +250,72 @@ fn process_override(
...
+
+    for (old_h, mut expr, span) in expressions.drain() {
+        if let Expression::Override(h) = expr {
+            expr = Expression::Constant(override_map[h.index()]);
+        }
+        adjust_expr(&adjusted_local_expressions, &mut expr);
+        let h = evaluator.try_eval_and_append(expr, span)?;
+        debug_assert_eq!(old_h.index(), adjusted_local_expressions.len());
+        adjusted_local_expressions.push(h);
+    }
```
jimblandy, line 301: The simplicity of this loop is encouraging, as it suggests that changing when constant evaluation occurs would not be such a big change: you'd simply have a loop like this after validation.

jimblandy:
@teoxoy Does this close #4366 and #1762?

jimblandy:
Accidentally whacked the upstream branch.

jimblandy:
Found some expression handles that weren't getting adjusted.

jimblandy (approved):
Okay! Please review my commits before merging, but with those fixes, this PR looks good to me.

teoxoy:
> Does this close #4366 and #1762?

Yeah but let's do that once we merge the branch into trunk.

teoxoy:
> Found some expression handles that weren't getting adjusted.

Good catch! We can hopefully avoid those cases once we have some common infrastructure to deal with IR mutation.

teoxoy:
> Okay! Please review my commits before merging, but with those fixes, this PR looks good to me.

They look good, thanks!

PR #5404, opened 2024-03-16
Bump version of all crates to match latest patched version

### LINKED ISSUE #5396: wgpu trunk version number is behind release, making it hard to patch
The current version in the `trunk` branch's `Cargo.toml` is `0.19.0`, which is less than the latest published release, `0.19.3`. This means that a Cargo `[patch]` to use `wgpu` from the repository may be disregarded.

It would be useful if the version number in `trunk` was always equal to or greater than the release.

(Of course, this is moot in the case of breaking changes. Still, in that case it would be more informative for the version to be `0.20.0` than `0.19.0`.)

Wumpf:
**Connections**
* Fixes #5396

**Description**
~Figured it would actually be a good idea if unreleased would be on an alpha version (that we don't actually ever have to push to crates.io), makes everyone's life easier as pointed out by #5396~
Bumped all versions to match latest patched version


kpreid:
> makes everyone's life easier as pointed out by https://github.com/gfx-rs/wgpu/issues/5396

Bumping to an _incompatible_ (different major) version number **doesn't** make anyone's life easier, except that it makes the problem more accurately-reported, because Cargo will still consider it incompatible for patching.

The optimal strategy would be to bump to the next minor version until an actually incompatible change happens. (But I understand not wanting to bother with that, and agree that doing just this is strictly better than the previous situation.)

Wumpf:
Hmmm, I see. Nothing much to add and I agree that it's impractical to track when trunk is no longer compatible with the previously released version, we might as well just always immediately change the minor version.

Wumpf:
changed everything now to match latest patch version. (Makes the branch name a misnomer but who cares ;))

cwfitzgerald (approved):
Yeet

PR #5408, opened 2024-03-17
[naga-cli] make CliError use Cow.

ratmice:
**Connections**
Intended for a subsequent patch, allowing `ShaderStage` to be parsed from either 
file extension or by command line argument.

**Description**
`CliError` is currently limited to `&'static str` which means that the errors returned by the
various `FromStr` implementations cannot be converted into `CliError`.

This patch attempts to alleviate that, also allowing the use of `format!` with the `str`.

**Testing**
`cargo run -p naga-cli` without arguments should be sufficient to produce a `CliError`.

**Checklist**

- [x] Run `cargo fmt`.
  - [ ] `--target wasm32-unknown-emscripten` (failed on trunk, before/after)
- [x] Run `cargo xtask test` to run tests. (4 failures before/after)

Wasn't sure a changelog entry was necessary, given it isn't user visible change?

### DIFF
--- a/naga-cli/src/bin/naga.rs
+++ b/naga-cli/src/bin/naga.rs
@@ -1,7 +1,7 @@
 #![allow(clippy::manual_strip)]
 #[allow(unused_imports)]
 use std::fs;
-use std::{error::Error, fmt, io::Read, path::Path, str::FromStr};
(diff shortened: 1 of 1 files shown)


ratmice:
Think i'm going to close this for now, as I'm not entirely happy with the quality of errors coming out of the subsequent PR this was for, and that'll probably impact what is done here.

PR #5409, opened 2024-03-17
Remove exposed C symbols from renderpass/computepass recording

### LINKED ISSUE #3105: Using multiple major versions of wgpu causes linker errors
**Description**
When a binary links against both wgpu 0.13 and wgpu 0.14, a linker error is caused by one-definition-rule violations, caused by a bunch of `#[no_mangle]` functions like here: https://github.com/gfx-rs/wgpu/blob/b28a87e2984181a5eaf659850f55b1e5aa1a461c/wgpu-core/src/command/render.rs#L2111-L2112

**Repro steps**

`Cargo.toml`:
```toml
[package]
name = "double-wgpu"
version = "0.1.0"
edition = "2021"

[dependencies]
wgpu14 = { version = "0.14", package = "wgpu" }
wgpu13 = { version = "0.13", package = "wgpu" }
```

`src/main.rs`:
```rust
fn main() {
    wgpu13::Instance::new(wgpu13::Backends::PRIMARY);
    wgpu14::Instance::new(wgpu14::Backends::PRIMARY);
}
```

This should just work, ideally, since wgpu is Rust code, not C.

I'm not sure why these are marked as `#[no_mangle]` – they're called by code in `wgpu` that imports these items by name, not by linking to them through an `extern "C" { }` block. Mangling should not be a problem for that. If they're intended to also be used by C code, perhaps the FFI wrapper should be moved to a small shim crate?

Wumpf:
**Connections**
* Most of #3105

**Description**
There was a seemingly random assortment of C symbols exported from wgpu around command encoding. The C methods were already hardly usable since they first require initialization of objects that can only be created from Rust.
I noticed this a few weeks prior as unintentionally exported symbols on applications and libraries that use wgpu and the issue described in #3105 is caused by this.
The wgpu crate's wgpu-core implementation as well as `wgpu-native` rely on these methods, but *not* the fact that they're exported and `#[no_mangle]`.

Removing the FFI opens a few small clean-up opportunities and less unsafe code here.

**EDIT:** This PR leaves intentionally leaves out bundles because firefox still relies on this. See [comment below](https://github.com/gfx-rs/wgpu/pull/5409#issuecomment-2016947420)

**Testing**
It compiles!

The real test here is the review - am I missing something and these need to stick around?

**Checklist**

- [x] Run `cargo fmt`.


cwfitzgerald:
I think Firefox relies on this, so we can't back this out without confirming with them.

Wumpf:
Pulled firefox source now to have a look at: Firefox has its own ffi layer https://hg.mozilla.org/mozilla-central/file/tip/gfx/wgpu_bindings/src/command.rs which again falls the methods from `compute_ffi`/`render_ffi` but does not rely on them being exported on their own.

But best for some Mozilla folk to chime in and confirm. @jimblandy maybe since you commented on [that ticket](https://github.com/gfx-rs/wgpu/issues/3105#issuecomment-1278383343) before?

wgpu-core/src/command/bundle.rs
```diff
@@ -1715,25 +1706,22 @@ pub mod bundle_ffi {
     ///
     /// This function is unsafe as there is no guarantee that the given `label`
     /// is a valid null-terminated string.
-    #[no_mangle]
-    pub unsafe extern "C" fn wgpu_render_bundle_push_debug_group(
+    pub unsafe fn wgpu_render_bundle_push_debug_group(
         _bundle: &mut RenderBundleEncoder,
         _label: RawString,
```
Wumpf: raw strings should be converted to regular strings everywhere

wgpu-core/src/command/render.rs
```diff
@@ -2851,8 +2820,7 @@ pub mod render_ffi {
     ///
     /// This function is unsafe as there is no guarantee that the given pointer is
     /// valid for `render_bundle_ids_length` elements.
-    #[no_mangle]
-    pub unsafe extern "C" fn wgpu_render_pass_execute_bundles(
+    pub unsafe fn wgpu_render_pass_execute_bundles(
         pass: &mut RenderPass,
         render_bundle_ids: *const id::RenderBundleId,
```
Wumpf: pointers like this should also be just references

jimblandy:
@Wumpf Thanks for the ping. Since @nical changed command recording to never use wgpu in the content process, I think we may not need these any more, but let me check.

jimblandy:
The `#[no_mangle]` declarations in `wgpu-core/src/command/render.rs` and `compute.rs` are no longer in use, and can be removed.

However, it seems like Firefox is still using `wgpu-core/src/command/bundle.rs`'s `#[no_mangle]` functions from C++ code [like this](https://searchfox.org/mozilla-central/rev/c949d9fcb5e686c4d1c4e2ff20bfacfaa9d77801/dom/webgpu/RenderBundleEncoder.cpp#169). It should be straightforward to fix these in the same way the render and compute recording function uses were removed.

jimblandy:
I assume resolving this issue requires *all* `#[no_mangle]` attributes to be removed from `wgpu-core`. I've filed [Bug 1887574](https://bugzilla.mozilla.org/show_bug.cgi?id=1887574) for getting that done.

Wumpf:
thank you! I have some ongoing work that requires changes to the compute functions specifically in compute & render (follow-up to #5432 ) so I'll change this PR to clean up only those and we can take bundles at a later point in time.

Wumpf:
all done and cleaned-up. Removed all FFI code now from compute & render, but didn't touch the bundle interface.
Removing ffi scaffolding resulted in quite a bit of streamlining!

cwfitzgerald (approved):
Amazing!

cwfitzgerald:
Blocked on #5549

PR #5410, opened 2024-03-17
Add get_compilation_info

### LINKED ISSUE #2130: Unable to handle shader compilation errors

stefnotch:
**Connections**
Fix #2130

**Description**
This adds an async `get_compilation_info` method to the `wgpu` crate, which either uses the native compilation info as reported by naga, or the compilation info that the browser's WebGPU backend reported.

The general idea is to note down all the relevant info when the compilation happens, and later extract them when the user calls `get_compilation_info`
- WebGPU backend with WGSL code: We store the Javascript promise, and later access it. We also store the source code, so that we can remap the reported offsets.
- WebGPU backend with other code: We translate the code to WGSL, and store the compilation and validation errors. And then we store the Javascript promise, so that we can also report any errors that the underlying WGSL compiler reports. Here, we don't keep the source code around, cause it'd be nonsensical. (Offsets are pointing at auto-generated WGSL code)
- Native backend: We immediately compile it, and store the errors.

Relevant bits about API design are
- Using UTF-8 for positions. We're purposefully deviating from the `GPUCompilationInfo` specification, which uses UTF-16 for reporting positions.
- For any given error, we're only reporting the main error location. If naga reports [multiple locations for a single error](https://github.com/gfx-rs/wgpu/blob/82fa5801525e7a32118ba3e02b9ce4adfb57112a/naga/src/front/wgsl/error.rs#L610), then the additional locations will be in the message.
  - For tooling that really wants to get all the error locations: Please directly use naga

Also moved the `ShaderError` struct from `wgpu_core` to `naga`, so that it can be used by both
- wgpu_core
- webgpu with certain feature flags

Any suggestions on how to simplify the code, or improve it are very appreciated. There are a few bits that feel like I'm duplicating code, or feel suboptimal.

**Testing**

Basic unit tests have been added.

**Checklist**

- [x] Run `cargo fmt`.
  - [ ] `--target wasm32-unknown-emscripten` I cannot get this to run locally without a million compile errors. I clearly missed some setup part.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -190,7 +190,7 @@ By @atlv24 and @cwfitzgerald in [#5154](https://github.com/gfx-rs/wgpu/pull/5154
 
 - Improved `wgpu_hal` documentation. By @jimblandy in [#5516](https://github.com/gfx-rs/wgpu/pull/5516), [#5524](https://github.com/gfx-rs/wgpu/pull/5524), [#5562](https://github.com/gfx-rs/wgpu/pull/5562), [#5563](https://github.com/gfx-rs/wgpu/pull/5563), [#5566](https://github.com/gfx-rs/wgpu/pull/5566), [#5617](https://github.com/gfx-rs/wgpu/pull/5617), [#5618](https://github.com/gfx-rs/wgpu/pull/5618)
 - Add mention of primitive restart in the description of `PrimitiveState::strip_index_format`. By @cpsdqs in [#5350](https://github.com/gfx-rs/wgpu/pull/5350)
-- Document precise behaviour of `SourceLocation`. By @stefnotch in [#5386](https://github.com/gfx-rs/wgpu/pull/5386)
+- Document and tweak precise behaviour of `SourceLocation`. By @stefnotch in [#5386](https://github.com/gfx-rs/wgpu/pull/5386) and [#5410](https://github.com/gfx-rs/wgpu/pull/5410)
 - Give short example of WGSL `push_constant` syntax. By @waywardmonkeys in [#5393](https://github.com/gfx-rs/wgpu/pull/5393)
 - Fix incorrect documentation of `Limits::max_compute_workgroup_storage_size` default value. By @atlv24 in [#5601](https://github.com/gfx-rs/wgpu/pull/5601)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -123,6 +123,8 @@ By @atlv24 and @cwfitzgerald in [#5154](https://github.com/gfx-rs/wgpu/pull/5154
   ```
 - Breaking change: [`wgpu_core::pipeline::ProgrammableStageDescriptor`](https://docs.rs/wgpu-core/latest/wgpu_core/pipeline/struct.ProgrammableStageDescriptor.html#structfield.entry_point) is now optional. By @ErichDonGubler in [#5305](https://github.com/gfx-rs/wgpu/pull/5305).
 - `Features::downlevel{_webgl2,}_features` was made const by @MultisampledNight in [#5343](https://github.com/gfx-rs/wgpu/pull/5343)
+- Breaking change: [`wgpu_core::pipeline::ShaderError`](https://docs.rs/wgpu-core/latest/wgpu_core/pipeline/struct.ShaderError.html) has been moved to `naga`. By @stefnotch in [#5410](https://github.com/gfx-rs/wgpu/pull/5410)
+- Add a `get_compilation_info` method to `wgpu` to get shader compilation errors. By @stefnotch in [#5410](https://github.com/gfx-rs/wgpu/pull/5410)
 - More as_hal methods and improvements by @JMS55 in [#5452](https://github.com/gfx-rs/wgpu/pull/5452)
   - Added `wgpu::CommandEncoder::as_hal_mut`
   - Added `wgpu::TextureView::as_hal`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,12 +80,13 @@ This allows you to get more structured information about compilation errors, war
 ...
 let lighting_shader = ctx.device.create_shader_module(include_wgsl!("lighting.wgsl"));
 let compilation_info = lighting_shader.get_compilation_info().await;
-for message in compilation_info.messages.iter() {
-    let line_number_to_highlight = message
-        .location
-        .map(|location| location.line_number)
-        .unwrap_or(1);
-    println!("Compile error at line {line_number_to_highlight}");
+for message in compilation_info
+    .messages
+    .iter()
+    .filter(|m| m.message_type == wgpu::CompilationMessageType::Error)
+{
+    let line = message.location.map(|l| l.line_number).unwrap_or(1);
+    println!("Compile error at line {line}");
 }
 ```
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,26 @@ Due to a specification change `write_timestamp` is no longer supported on WebGPU
 
 By @wumpf in [#5188](https://github.com/gfx-rs/wgpu/pull/5188)
 
+#### Querying shader compilation errors
+
+Wgpu now supports querying [shader compilation info](https://www.w3.org/TR/webgpu/#dom-gpushadermodule-getcompilationinfo).
+
+This allows you to get more structured information about compilation errors, warnings and info:
+```rust
+...
+let lighting_shader = ctx.device.create_shader_module(include_wgsl!("lighting.wgsl"));
+let compilation_info = lighting_shader.get_compilation_info().await;
+for message in compilation_info.messages.iter() {
+    let line_number_to_highlight = message
+        .location
+        .map(|location| location.line_number)
+        .unwrap_or(1);
+    println!("Compile error at line {line_number_to_highlight}");
+}
+```
+
+By @stefnotch in [#5410](https://github.com/gfx-rs/wgpu/pull/5410)
+
 
 #### Wgsl const evaluation for many more built-ins
 
@@ -124,7 +144,6 @@ By @atlv24 and @cwfitzgerald in [#5154](https://github.com/gfx-rs/wgpu/pull/5154
 - Breaking change: [`wgpu_core::pipeline::ProgrammableStageDescriptor`](https://docs.rs/wgpu-core/latest/wgpu_core/pipeline/struct.ProgrammableStageDescriptor.html#structfield.entry_point) is now optional. By @ErichDonGubler in [#5305](https://github.com/gfx-rs/wgpu/pull/5305).
 - `Features::downlevel{_webgl2,}_features` was made const by @MultisampledNight in [#5343](https://github.com/gfx-rs/wgpu/pull/5343)
 - Breaking change: [`wgpu_core::pipeline::ShaderError`](https://docs.rs/wgpu-core/latest/wgpu_core/pipeline/struct.ShaderError.html) has been moved to `naga`. By @stefnotch in [#5410](https://github.com/gfx-rs/wgpu/pull/5410)
-- Add a `get_compilation_info` method to `wgpu` to get shader compilation errors. By @stefnotch in [#5410](https://github.com/gfx-rs/wgpu/pull/5410)
 - More as_hal methods and improvements by @JMS55 in [#5452](https://github.com/gfx-rs/wgpu/pull/5452)
   - Added `wgpu::CommandEncoder::as_hal_mut`
   - Added `wgpu::TextureView::as_hal`
--- a/naga/src/front/glsl/error.rs
+++ b/naga/src/front/glsl/error.rs
@@ -145,15 +145,14 @@ impl Error {
     }
 }
 
-// TODO: Rename to ParseErrors?
 /// A collection of errors returned during shader parsing.
 #[derive(Clone, Debug)]
 #[cfg_attr(test, derive(PartialEq))]
-pub struct ParseError {
+pub struct ParseErrors {
     pub errors: Vec<Error>,
 }
 
-impl ParseError {
+impl ParseErrors {
     pub fn emit_to_writer(&self, writer: &mut impl WriteColor, source: &str) {
         self.emit_to_writer_with_path(writer, source, "glsl");
     }
@@ -181,19 +180,19 @@ impl ParseError {
     }
 }
 
-impl std::fmt::Display for ParseError {
+impl std::fmt::Display for ParseErrors {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         self.errors.iter().try_for_each(|e| write!(f, "{e:?}"))
     }
 }
 
-impl std::error::Error for ParseError {
+impl std::error::Error for ParseErrors {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
         None
     }
 }
 
-impl From<Vec<Error>> for ParseError {
+impl From<Vec<Error>> for ParseErrors {
     fn from(errors: Vec<Error>) -> Self {
         Self { errors }
     }
--- a/naga/src/front/glsl/error.rs
+++ b/naga/src/front/glsl/error.rs
@@ -1,4 +1,5 @@
 use super::token::TokenValue;
+use crate::SourceLocation;
 use crate::{proc::ConstantEvaluatorError, Span};
 use codespan_reporting::diagnostic::{Diagnostic, Label};
 use codespan_reporting::files::SimpleFile;
@@ -137,6 +138,14 @@ pub struct Error {
     pub meta: Span,
 }
 
+impl Error {
+    /// Returns a [`SourceLocation`] for the error message.
+    pub fn location(&self, source: &str) -> Option<SourceLocation> {
+        Some(self.meta.location(source))
+    }
+}
+
+// TODO: Rename to ParseErrors?
 /// A collection of errors returned during shader parsing.
 #[derive(Clone, Debug)]
 #[cfg_attr(test, derive(PartialEq))]
--- a/naga/src/front/wgsl/error.rs
+++ b/naga/src/front/wgsl/error.rs
@@ -13,6 +13,7 @@ use thiserror::Error;
 #[derive(Clone, Debug)]
 pub struct ParseError {
     message: String,
+    // TODO: Document that the first span should be the primary span, and the other ones should be complementary.
     labels: Vec<(Span, Cow<'static, str>)>,
     notes: Vec<String>,
 }
--- a/naga/src/front/wgsl/error.rs
+++ b/naga/src/front/wgsl/error.rs
@@ -13,7 +13,7 @@ use thiserror::Error;
 #[derive(Clone, Debug)]
 pub struct ParseError {
     message: String,
-    // TODO: Document that the first span should be the primary span, and the other ones should be complementary.
+    // The first span should be the primary span, and the other ones should be complementary.
     labels: Vec<(Span, Cow<'static, str>)>,
     notes: Vec<String>,
 }
--- a/wgpu-core/src/pipeline.rs
+++ b/wgpu-core/src/pipeline.rs
@@ -10,7 +10,8 @@ use crate::{
     resource_log, validation, Label,
 };
 use arrayvec::ArrayVec;
-use std::{borrow::Cow, error::Error, fmt, marker::PhantomData, num::NonZeroU32, sync::Arc};
+use naga::error::ShaderError;
+use std::{borrow::Cow, marker::PhantomData, num::NonZeroU32, sync::Arc};
 use thiserror::Error;
 
 /// Information about buffer bindings, which
@@ -107,77 +108,6 @@ impl<A: HalApi> ShaderModule<A> {
     }
 }
 
-#[derive(Clone, Debug)]
-pub struct ShaderError<E> {
-    pub source: String,
-    pub label: Option<String>,
-    pub inner: Box<E>,
-}
-#[cfg(feature = "wgsl")]
-impl fmt::Display for ShaderError<naga::front::wgsl::ParseError> {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let label = self.label.as_deref().unwrap_or_default();
-        let string = self.inner.emit_to_string(&self.source);
-        write!(f, "\nShader '{label}' parsing {string}")
-    }
-}
-#[cfg(feature = "glsl")]
-impl fmt::Display for ShaderError<naga::front::glsl::ParseError> {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let label = self.label.as_deref().unwrap_or_default();
-        let string = self.inner.emit_to_string(&self.source);
-        write!(f, "\nShader '{label}' parsing {string}")
-    }
-}
-#[cfg(feature = "spirv")]
-impl fmt::Display for ShaderError<naga::front::spv::Error> {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let label = self.label.as_deref().unwrap_or_default();
-        let string = self.inner.emit_to_string(&self.source);
-        write!(f, "\nShader '{label}' parsing {string}")
-    }
-}
-impl fmt::Display for ShaderError<naga::WithSpan<naga::valid::ValidationError>> {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        use codespan_reporting::{
-            diagnostic::{Diagnostic, Label},
-            files::SimpleFile,
-            term,
-        };
-
-        let label = self.label.as_deref().unwrap_or_default();
-        let files = SimpleFile::new(label, &self.source);
-        let config = term::Config::default();
-        let mut writer = term::termcolor::NoColor::new(Vec::new());
-
-        let diagnostic = Diagnostic::error().with_labels(
-            self.inner
-                .spans()
-                .map(|&(span, ref desc)| {
-                    Label::primary((), span.to_range().unwrap()).with_message(desc.to_owned())
-                })
-                .collect(),
-        );
-
-        term::emit(&mut writer, &config, &files, &diagnostic).expect("cannot write error");
-
-        write!(
-            f,
-            "\nShader validation {}",
-            String::from_utf8_lossy(&writer.into_inner())
-        )
-    }
-}
-impl<E> Error for ShaderError<E>
-where
-    ShaderError<E>: fmt::Display,
-    E: Error + 'static,
-{
-    fn source(&self) -> Option<&(dyn Error + 'static)> {
-        Some(&self.inner)
-    }
-}
-
 //Note: `Clone` would require `WithSpan: Clone`.
 #[derive(Debug, Error)]
 #[non_exhaustive]
@@ -207,17 +137,6 @@ pub enum CreateShaderModuleError {
         group: u32,
         limit: u32,
     },
-}
-
-impl CreateShaderModuleError {
-    pub fn location(&self, source: &str) -> Option<naga::SourceLocation> {
-        match *self {
-            #[cfg(feature = "wgsl")]
-            CreateShaderModuleError::Parsing(ref err) => err.inner.location(source),
-            CreateShaderModuleError::Validation(ref err) => err.inner.location(source),
-            _ => None,
-        }
-    }
 }
 
 /// Describes a programmable pipeline stage.
--- a/wgpu-core/src/pipeline.rs
+++ b/wgpu-core/src/pipeline.rs
@@ -117,7 +117,7 @@ pub enum CreateShaderModuleError {
     Parsing(#[from] ShaderError<naga::front::wgsl::ParseError>),
     #[cfg(feature = "glsl")]
     #[error(transparent)]
-    ParsingGlsl(#[from] ShaderError<naga::front::glsl::ParseError>),
+    ParsingGlsl(#[from] ShaderError<naga::front::glsl::ParseErrors>),
     #[cfg(feature = "spirv")]
     #[error(transparent)]
     ParsingSpirV(#[from] ShaderError<naga::front::spv::Error>),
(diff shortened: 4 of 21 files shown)


wgpu/src/backend/webgpu.rs
```diff
@@ -1493,7 +1493,10 @@ impl crate::context::Context for ContextWebGpu {
         if let Some(label) = desc.label {
             descriptor.label(label);
         }
-        create_identified(device_data.0.create_shader_module(&descriptor))
+        let shader_module = device_data.0.create_shader_module(&descriptor);
+        let _compilation_info = shader_module.get_compilation_info();
```
stefnotch: This is a Javascript promise. It has been causing me a lot of headaces, because it forces an asynchronous API.

wgpu/src/backend/wgpu_core.rs
```diff
@@ -852,6 +856,7 @@ impl crate::Context for ContextWgpuCore {
             device => self.0.device_create_shader_module(*device, &descriptor, source, None)
         );
         if let Some(cause) = error {
+            let _info = CompilationInfo::from(&cause);
```
stefnotch: Creating a compilation info from a wgpu-core error seems reasonably straightforward, see implementation below.

wgpu/src/backend/wgpu_core.rs
```diff
@@ -2991,6 +2997,77 @@ fn default_error_handler(err: crate::Error) {
...
+        let message = match value {
+            #[cfg(feature = "wgsl")]
+            CreateShaderModuleError::Parsing(e) => CompilationMessage {
+                message: e.to_string(),
+                message_type: CompilationMessageType::Error,
+                span: e
+                    .inner
+                    .labels()
+                    .next()
+                    .and_then(|(span, _)| span_to_source_span(&span)),
```
stefnotch: I wasn't entirely sure if I should only take the first span, or do something different.

wgpu/src/backend/wgpu_core.rs
```diff
@@ -2991,6 +2997,77 @@ fn default_error_handler(err: crate::Error) {
...
+            },
+            CreateShaderModuleError::MissingFeatures(_)
+            | CreateShaderModuleError::InvalidGroupIndex { .. } => CompilationMessage {
+                message: value.to_string(),
+                message_type: CompilationMessageType::Error,
+                span: None,
+            },
+            // Device errors are reported through the error sink, and are not compilation errors.
+            // Same goes for native shader module generation errors.
+            CreateShaderModuleError::Device(_) | CreateShaderModuleError::Generation => {
```
stefnotch, line 3036: The WebGPU spec says
> A shader-creation error is an error feasibly detectable at shader module creation time. Detection relies only on the WGSL module source text and other information available to the createShaderModule API method. Statements in this specification that describe something the program must do generally produce a shader-creation error if those assertions are violated.

I figured that this is in the spirit of the specification, but am happy to change it.

wgpu/src/lib.rs
```diff
@@ -632,6 +632,57 @@ impl Drop for ShaderModule {
...
+}
+
+/// Byte positions in an UTF-8 encoded string.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct SourceSpan {
+    /// 0-based offset in code units (in bytes) of the start of the span.
+    pub offset: u32,
+    /// Length in code units (in bytes) of the span.
+    pub length: u32,
+}
```
stefnotch, line 877: Was creating a new type here the correct move, or should I just have exposed the already existing Naga `Span` type?

Also, the WebGPU spec wants a `lineNum` and a `linePos`. Should I add those to this struct and thus redundantly store them? Or add a helper method to compute them? Or neither of those options?

wgpu/src/lib.rs
```diff
@@ -632,6 +632,57 @@ impl Drop for ShaderModule {
     }
 }
 
+impl ShaderModule {
+    /// Get the compilation info for the shader module.
+    pub async fn compilation_info(&self) -> CompilationInfo {
+        todo!()
```
stefnotch: Piping through the wgpu-core compilation message is easy enough.

But I have no idea how to deal with the Javascript promise and async Rust. Any pointers there would be hugely appreciated.

wgpu/src/backend/wgpu_core.rs
```diff
@@ -2994,6 +3023,62 @@ fn default_error_handler(err: crate::Error) {
     panic!("wgpu error: {err}\n");
 }
 
+impl From<&CreateShaderModuleError> for CompilationInfo {
```
stefnotch: The CI fails, because there's a configuration where need this trait in `wgpu/src/backend/webgpu.rs`, but `#[cfg(wgpu_core)]` doesn't apply. So it doesn't get compiled, and Rust rightfully throws an error.

Any suggestions on where to move it?

wgpu/src/lib.rs
```diff
@@ -632,6 +632,58 @@ impl Drop for ShaderModule {
...
+#[derive(Debug, Clone)]
+pub struct CompilationInfo {
+    /// The messages from the shader compilation process.
+    pub messages: Vec<CompilationMessage>,
+}
+
+/// A single message from the shader compilation process.
+///
+/// Roughly corresponds to [`GPUCompilationMessage`](https://www.w3.org/TR/webgpu/#gpucompilationmessage),
+/// except that there is only a `span`, with a byte positions in a UTF-8 encoded source code.
```
```suggestion
/// except that there is only a `span`, with a byte positions in the UTF-8 encoded source code that was passed for the corresponding shader.
```

wgpu-core/src/pipeline.rs
```diff
@@ -218,6 +229,32 @@ impl CreateShaderModuleError {
             _ => None,
         }
     }
+
+    pub fn main_range(&self) -> Option<std::ops::Range<usize>> {
```
Wumpf: not entirely clear from the name that this is a source code range this error is about (how about `source_range`?) but more importantly, I'm confused about how this relates to `location` just above. Isn't that what we're actually looking for here?

wgpu/src/backend/webgpu.rs
```diff
@@ -843,6 +909,41 @@ fn future_pop_error_scope(result: JsFutureResult) -> Option<crate::Error> {
...
+                }))
+                .collect()
+        }
+        Err(_v) => base_messages
+            .chain(std::iter::once(crate::CompilationMessage {
+                message: "Getting compilation info failed".to_string(),
+                message_type: crate::CompilationMessageType::Error,
+                span: None,
+            }))
+            .collect(),
```
Wumpf: nice idea to chain that error simply in :+1:

wgpu/src/backend/webgpu.rs
```diff
@@ -1417,83 +1523,176 @@ impl crate::context::Context for ContextWebGpu {
...
+                        block_ctx_dump_prefix: None,
+                    };
+                    let spv_parser = front::spv::Frontend::new(spv.iter().cloned(), &options);
+                    let spv_module = spv_parser
+                        .parse()
+                        .map_err(|inner| wgc::pipeline::ShaderError {
+                            source: String::new(),
+                            label: desc.label.map(|s| s.to_string()),
+                            inner: Box::new(inner),
+                        })
```
Wumpf: since you asked about less code dupl - this one could have a mapping utility function. A `From`/(`Into`) would be nice, but won't work unfortunately since each type is defined elsewhere :/

wgpu/src/backend/webgpu.rs
```diff
@@ -1417,83 +1523,176 @@ impl crate::context::Context for ContextWebGpu {
...
+                                message_type: crate::CompilationMessageType::Error,
+                                span: err
+                                    .inner
+                                    .errors
+                                    .first()
+                                    .and_then(|e| e.meta.to_range())
+                                    .map(|range| crate::SourceSpan {
+                                        offset: range.start as u32,
+                                        length: range.len() as u32,
+                                    }),
```
Wumpf: that one could be handled with an `From` implementation :)

wgpu/src/backend/webgpu.rs
```diff
@@ -1417,83 +1523,176 @@ impl crate::context::Context for ContextWebGpu {
...
+                            label: desc.label.map(|s| s.to_string()),
+                            inner: Box::new(inner),
+                        })
+                        .map_err(|err| crate::CompilationInfo {
+                            messages: vec![crate::CompilationMessage {
+                                message: err.to_string(),
+                                message_type: crate::CompilationMessageType::Error,
+                                span: err
+                                    .inner
+                                    .errors
```
Wumpf: err.to_string just concats the strings for all underlying glsl parsing errors if I follow it correctly. So why not do `err.inner.errors.map` and map each error to a `CompilationMessage`?

wgpu/src/backend/webgpu.rs
```diff
@@ -1417,83 +1523,176 @@ impl crate::context::Context for ContextWebGpu {
...
+                    label: desc.label.map(|s| s.to_string()),
+                    inner: Box::new(err),
+                })
+                .map_err(|err| crate::CompilationInfo {
+                    messages: vec![crate::CompilationMessage {
+                        message: err.to_string(),
+                        message_type: crate::CompilationMessageType::Error,
+                        span: err
+                            .inner
+                            .spans()
```
Wumpf: as before, why merge the errors and then take arbitrarily the first span instead of exposing all errors als separate messages

EDIT: ah I think this is different here because this is a `WithSpan<ValidationError>` if I'm not mistaken. Hmm still strange though to pick the first span and ignore the rest :/. Maybe nothing to be done about it right now.

wgpu/src/backend/webgpu.rs
```diff
@@ -106,6 +106,72 @@ impl crate::Error {
...
+    for (utf8_index, c) in text.char_indices() {
+        if utf16_i >= utf16_offset {
+            return utf8_index as u32;
+        }
+        utf16_i += c.len_utf16() as u32;
+    }
+    if utf16_i >= utf16_offset {
+        text.len() as u32
+    } else {
+        panic!("UTF16 offset out of bounds");
```
Wumpf: I don't think we should panic if the webgpu impl messes this up. let's return max int and call `log::error!` instead

wgpu/src/backend/webgpu.rs
```diff
@@ -106,6 +106,72 @@ impl crate::Error {
...
+        js_message: webgpu_sys::GpuCompilationMessage,
+        compilation_info: &WebShaderCompilationInfo,
+    ) -> Self {
+        let message_type = match js_message.type_() {
+            webgpu_sys::GpuCompilationMessageType::Error => crate::CompilationMessageType::Error,
+            webgpu_sys::GpuCompilationMessageType::Warning => {
+                crate::CompilationMessageType::Warning
+            }
+            webgpu_sys::GpuCompilationMessageType::Info => crate::CompilationMessageType::Info,
+            // TODO: What should be the default?
```
Wumpf: let's go with error - after all it's an error in wgpu or the webgpu impl if we don't know about the type :)

Wumpf (changes requested):
Looking great already! Left a few comments on things to improve / check upon.
Very nicely documented api on wgpu :+1:

Regarding tests: Yes plz! :). You can put them under tests/test s, maybe even part of tests/tests/shader, haven't checked if that makes sense. Would be great to have simple test cases for success, warning and error.

Also, please add a changelog entry. This is a great addition that shouldn't be missed there :)

stefnotch:
By the way, if I run
```
cargo check --target wasm32-unknown-emscripten
```
I get the following output. Am I doing something wrong?
```                                                                                                                                                                                                                            04/12/2024 10:45:49 AM AM
    Checking spirv v0.3.0+sdk-1.3.268.0
    Checking bytemuck v1.15.0
    Checking winit v0.29.15
    Checking player v0.19.3 (D:\Coding\GitHub\Libraries\wgpu\player)
error: The platform you're compiling for is not supported by winit
  --> C:\Users\Stefnotch\.cargo\registry\src\index.crates.io-6f17d22bba15001f\winit-0.29.15\src\platform_impl\mod.rs:67:1
   |
67 | compile_error!("The platform you're compiling for is not supported by winit");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
```

Wumpf:
uh yeah emscripten target seems to be broken again (much worse for me on trunk actually), ignore that for now. `wasm32-unknown-unknown` still seems to work fine (it also gets a lot more testing on ci)

wgpu-core/src/pipeline.rs
stefnotch: I added a separate method, since the existing `location` method requires a source string (and piping that one through is tough), and it computes a few pieces of information that I just end up throwing away. (line number and offset in the line)

If there's a better option that ultimately leads to simpler code, I'd love to hear about it.

wgpu-core/src/pipeline.rs
```diff
@@ -218,6 +229,32 @@ impl CreateShaderModuleError {
...
+        match self {
+            #[cfg(feature = "wgsl")]
+            CreateShaderModuleError::Parsing(err) => err
+                .inner
+                .labels()
+                .next()
+                .and_then(|(span, _)| span.to_range()),
+            #[cfg(feature = "glsl")]
+            CreateShaderModuleError::ParsingGlsl(err) => {
+                err.inner.errors.first().and_then(|e| e.meta.to_range())
```
stefnotch: This one feels odd, most of the error types report one error, right? Meanwhile ParsingGlsl seems to be able to return multiple errors?

wgpu/src/backend/webgpu.rs
stefnotch: True, good point. It feels a tad inconsistent that the glsl error type has multiple underlying errors while the other error types only seem to have one.
Should anything be done about that?

wgpu/src/backend/webgpu.rs
stefnotch: Yeah, I'm not certain what the best solution here is. I could
- do nothing and keep it as is
- add a Vec<> for the compilation spans to the `CompilationMessage` type, deviating from the standard
- find a few good examples and open an issue in the repository where the work on the WebGPU specification happens (surely there are cases where an error clearly has multiple spans, right?)
- report multiple spans as multiple errors? But that seems wrong
- join the spans? Also seems wrong

stefnotch:
I added tests for this 🎉  They're very barebones, but should be decent enough to verify that this feature isn't completely broken.

Out of curiosity, do those tests need to be gpu tests? Couldn't they theoretically run with a fake GPU backend?

Wumpf:
> Out of curiosity, do those tests need to be gpu tests? Couldn't they theoretically run with a fake GPU backend?

most of what ci runs is actually with software rasterizers by the respective apis. You're right that for these tests we don't actually need to go to a real backend but we're lacking the necessary mocking architecture to stub out wgpu_hal out of these.

wgpu-core/src/pipeline.rs
Wumpf: I see, makes sense!

wgpu-core/src/pipeline.rs
Wumpf: well the others are also returning iterators and you're only calling `next()` on them, thus only looking at the first one. All of them have several error spans.

cc: @teoxoy generally we should expose these errors one by one to webgpu right?

wgpu/src/backend/webgpu.rs
Wumpf: as commented above, I don't think glsl is that special here; it just has a slightly different interface in the errors. Also, it's not unlikely that all parsers expose multiple errors in the future.

Wumpf:
huh, sorry for the long delay - I missed the updates on this

Thanks for the improvements, looking better already. But I'm still not happy about any of the "pick first" code in here. If any of these were to remain we need a good (documented) reason each, otherwise I have to assume that we have a bug where we're accidentally hiding errors.

wgpu-core/src/pipeline.rs
teoxoy: Right now only the GLSL frontend reports multiple errors but the WGSL frontend will as well in the future.

WGSL errors are richer and contain multiple spans so that we can point to different parts of the shader, see for example the redefinition error:

https://github.com/gfx-rs/wgpu/blob/82fa5801525e7a32118ba3e02b9ce4adfb57112a/naga/src/front/wgsl/error.rs#L610

teoxoy: Is this new method necessary? It seems to be doing the same thing as the one above.

stefnotch: > WGSL errors are richer and contain multiple spans so that we can point to different parts of the shader, see for example the redefinition error:

Thank you for the example! How do I best report that as a WebGPU error?
I'm basically struggling with the fact that the official WebGPU `GPUCompilationMessage` only lets one include a single location
https://developer.mozilla.org/en-US/docs/Web/API/GPUCompilationMessage

One could
- add a `Vec<>` for the compilation spans to the wgpu `CompilationMessage` type, deviating from the standard
- open an issue in the repository where the work on the WebGPU specification happens, and making the case for "multiple locations make sense"
- report multiple spans as multiple errors (e.g. Make the first one an error, and the other ones warnings that pretty much say "error also happened here")



> Is this new method necessary? It seems to be doing the same thing as the one above.

I added a separate method, since the existing location method requires a source string, and it computes a few pieces of information that I just end up throwing away. (line number and offset in the line)

I'd be happy about better suggestions though. Maybe I could tweak the existing `location` methods to return a `Range` and defer resolving the `naga::SourceLocation` until a bit later?

teoxoy: > How do I best report that as a WebGPU error?
I'm basically struggling with the fact that the official WebGPU `GPUCompilationMessage` only lets one include a single location

I think a single location is fine, as for the error message we can use the `ParseError.emit_to_string()` method (it will return a nicely formatted error message that takes into account all the spans).

The implementation of `.location()` seems to use the first span which I think is correct? for now? I guess it makes sense that the first span would point to the issue at hand and the others are complementary but it doesn't seem to be documented.

> I added a separate method, since the existing location method requires a source string, and it computes a few pieces of information that I just end up throwing away. (line number and offset in the line)

Why do we need to throw some of the fields away? All `naga::SourceLocation`'s fields are part of `GPUCompilationMessage`.

stefnotch: @teoxoy
Sounds good.

Designing a good `CompilationMessage` for wgpu is a bit annoying, because we pretty much have to deviate from [the standard](https://gpuweb.github.io/gpuweb/#gpucompilationmessage) for it to be idiomatic Rust. Here's what I'm leaning towards at the moment, does this make sense?
```rust
pub struct CompilationMessage {
    pub message: String,
    // `message_type` is a name that doesn't clash with a reserved keyword, unlike `type`
    pub message_type: CompilationMessageType,
    /// An option, instead of the "Must be 0 if the message does not correspond to any specific point in the shader code." rule.
    pub location: Option<SourceLocation>,
}

// Exists so that we can put the location info in an option
pub struct SourceLocation {
    /// Assumes a UTF-8 encoded source instead of assuming UTF-16
    pub offset: u32,
    /// Assumes a UTF-8 encoded source instead of assuming UTF-16
    pub length: u32,
    /// This one actually matches the standard. Well, except that the name is `line_number` instead of `line_num`.
    pub line_number: u32,
    /// Uses Unicode code units, which are more comprehensible for humans instead of UTF-16 offsets.
    /// Naga does the same thing, hence this decision. I'm not sure if it's ideal though. 
    /// e.g. VSCode's column indicator actually uses UTF-16 to count, as far as I can tell. 
    /// And Unicode code units also aren't perfect for humans, since they don't match up with visual characters (graphemes) 
    pub line_position: u32,
}
```

Would love some feedback.

wgpu/src/backend/webgpu.rs
stefnotch: The other spans should be in the error message, which is hopefully good enough.

naga/src/front/glsl/error.rs
```diff
@@ -137,6 +138,14 @@ pub struct Error {
...
 }
 
+impl Error {
+    /// Returns a [`SourceLocation`] for the first label in the error message.
+    pub fn location(&self, source: &str) -> Option<SourceLocation> {
+        Some(self.meta.location(source))
+    }
+}
+
+// TODO: Rename to ParseErrors?
```
stefnotch: Would it make sense to rename this struct below to `ParseErrors`?

wgpu-core/src/pipeline.rs
teoxoy: I think that looks reasonable! I also don't know what would be best for `line_position`; we can go with "unicode code units" for now.

kpreid:
> Using Unicode code units for the line_position. 

>  `/// Uses Unicode code units, which are more comprehensible for humans instead of UTF-16 offsets.`

I think you mean code _points_ here. Counting code units is encoding-dependent. But beyond that terminological nitpick: surely on basic principles this should be counting UTF-8 bytes (aka UTF-8 code units) instead? That's the only way a Rust string can be indexed/sliced efficiently.

If the goal is to mark spans by counting "columns", then Unicode code points do not do that either — you need to execute the [`unicode-width`](https://crates.io/crates/unicode-width) algorithm to approximate that, or ask your font/terminal for feedback to get it perfect. I don't think that should be baked into the API, so it's better to present what is maximally useful for slicing, and let the caller compute the width it cares about.

Don't count code points, because there are almost no use cases where it is the _correct_ answer.

stefnotch:
@kpreid Yes, my bad, I meant code points.

I would, for a brief moment, have argued that counting code points makes sense, since VSCode seems to do that. Except that that's not true, since VSCode has a special case for tabs `\t`. Surprisingly, VSCode doesn't have a special case for emojis that consist of multiple code points. 

So I'm realising that it's bonkers and you absolutely have a point. And that the best possible option is outputting something that a computer can reasonably convert to whatever is most appropriate for a given case. 

Though at that point, one would usually directly use the `offset` field instead?

For what it's worth, I did copy it from the internal naga `SourceLocation` struct. Maybe the internal struct should also be changed?

stefnotch:
Okay, I fixed a few more issues. Importantly, I moved the `ShaderError` struct from `wgpu_core` to `naga`, so that it can be used by both
- wgpu_core
- webgpu with certain feature flags

I wonder if there's a simpler variation of those `cfg`s for conditional compilation than whatever I chose.

CHANGELOG.md
```diff
@@ -123,6 +123,8 @@ By @atlv24 and @cwfitzgerald in [#5154](https://github.com/gfx-rs/wgpu/pull/5154
   ```
 - Breaking change: [`wgpu_core::pipeline::ProgrammableStageDescriptor`](https://docs.rs/wgpu-core/latest/wgpu_core/pipeline/struct.ProgrammableStageDescriptor.html#structfield.entry_point) is now optional. By @ErichDonGubler in [#5305](https://github.com/gfx-rs/wgpu/pull/5305).
 - `Features::downlevel{_webgl2,}_features` was made const by @MultisampledNight in [#5343](https://github.com/gfx-rs/wgpu/pull/5343)
+- Breaking change: [`wgpu_core::pipeline::ShaderError`](https://docs.rs/wgpu-core/latest/wgpu_core/pipeline/struct.ShaderError.html) has been moved to `naga`. By @stefnotch in [#5410](https://github.com/gfx-rs/wgpu/pull/5410)
+- Add a `get_compilation_info` method to `wgpu` to get shader compilation errors. By @stefnotch in [#5410](https://github.com/gfx-rs/wgpu/pull/5410)
```
Wumpf: now that I already did the 0.20 prep for changelog, could you move this to the highlight section above `New features` with a brief usage example? Deserves one if it should make the cut for 0.20 :)

naga/src/front/glsl/error.rs
Wumpf: agreed `ParseErrors` (plural) makes more sense

naga/src/front/wgsl/error.rs
```diff
@@ -13,6 +13,7 @@ use thiserror::Error;
 #[derive(Clone, Debug)]
 pub struct ParseError {
     message: String,
+    // TODO: Document that the first span should be the primary span, and the other ones should be complementary.
```
Wumpf: If you remove the todo you practically did it? 😄

wgpu/src/backend/webgpu.rs
```diff
@@ -1420,83 +1542,136 @@ impl crate::context::Context for ContextWebGpu {
...
+                        },
+                    )
+                })
             }
             crate::ShaderSource::Dummy(_) => {
                 panic!("found `ShaderSource::Dummy`")
             }
         };
+
+        #[cfg(any(feature = "spirv", feature = "glsl", feature = "naga-ir"))]
```
Wumpf: you should check for the `naga` alias here instead, see https://github.com/gfx-rs/wgpu/blob/trunk/wgpu/build.rs#L16

wgpu/src/lib.rs
```diff
@@ -812,6 +812,140 @@ impl Drop for ShaderModule {
...
+    /// 1-based column of the start of this span, counted in Unicode code points.
+    pub line_position: u32,
+    /// 0-based Offset in code units (in bytes) of the start of the span.
+    pub offset: u32,
+    /// Length in code units (in bytes) of the span.
+    pub length: u32,
+}
+
+#[cfg(all(feature = "wgsl", wgpu_core))]
+impl From<&naga::error::ShaderError<naga::front::wgsl::ParseError>> for CompilationInfo {
```
Wumpf: nice, so much cleaner with these `From` implementations all listed out here!

The only thing I findodd is that they work on reference types. I expect from/into to be consuming - if the callsite needs it longer (for error handling as it happens in a few places) it's the callsite's job to do the necessary clone. You're also not entirely consistent with this here having `impl From<naga::SourceLocation> for SourceLocation` further down.

wgpu/src/backend/wgpu_core.rs
```diff
@@ -2996,6 +3023,35 @@ fn default_error_handler(err: crate::Error) {
...
```
Wumpf: might have missed this earlier, but similar to comment on `impl From<...> for SourceLocation` I'd prefer to avoid `From` implementations that operate on references. From/Into is supposed to take ownership of the thing it converts and callsites can get weird & confusing when it sometimes does and sometimes doesn't. (it still consumes the _reference_, but no longer the object itself)

Wumpf (changes requested):
Looking really good now. A few cosmetic notes that would be nice to address.
Thanks @teoxoy for the input - now that this touched Naga stuff more directly I'll wait for your additional sign-off before merge.

Could you try to avoid rebasing on the next iteration? Rebasing may look nice on the history but makes it a lot harder to incrementally review - I can't tell github to show me the changes since last time and instead have to start from scratch

naga/src/front/wgsl/error.rs
stefnotch: Perfect, did it. 

I also checked all the `ParseError`s that get constructed. They all obey the rule, except for the `Error::CyclicDeclaration` case. The cyclic declaration case doesn't really have a "primary span".

wgpu/src/lib.rs
stefnotch: I see. I'm guessing that convention is more widespread (like `into_iter` and `iter`) and should be worth remembering.

Will fix :)

stefnotch:
@kpreid @teoxoy I looked into the linePos (column number) a bit more. I also looked into the naga `SourceLocation` struct, since it also contains such info.

The `naga::SourceLocation` struct is mainly being used for emitting SpirV OpLine instructions. Those are for debug tools.
And inconveniently, [the specification](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpLine) says exactly nothing about what the column in an `OpLine` is supposed to represent. It's just supposed to be a positive integer.

So I looked at what some tools do
- SPIRV-Cross: Always emits `0` for the column number.
- DirectXShaderCompiler: [Uses the length of the string plus one](https://github.com/microsoft/DirectXShaderCompiler/blob/00cd8233d5d6755dae1218e5f84bafdad83c7fc5/tools/clang/lib/Basic/SourceManager.cpp#L1118). I couldn't definitively figure out which encoding is being used.
- Renderdoc: Couldn't figure out what it assumes, but it doesn't seem to be counting code points.

So my question: May I change the `naga::SourceLocation.line_position` to count UTF-8 code units instead of general Unicode code points? This only makes a difference when dealing with non-ASCII Unicode characters, so it doesn't actually come up very often.

CHANGELOG.md
```diff
@@ -71,6 +71,26 @@ Due to a specification change `write_timestamp` is no longer supported on WebGPU
...
+```rust
+...
+let lighting_shader = ctx.device.create_shader_module(include_wgsl!("lighting.wgsl"));
+let compilation_info = lighting_shader.get_compilation_info().await;
+for message in compilation_info.messages.iter() {
+    let line_number_to_highlight = message
+        .location
+        .map(|location| location.line_number)
+        .unwrap_or(1);
+    println!("Compile error at line {line_number_to_highlight}");
```
Wumpf: not all messages are errors, so this is a bit misleading

teoxoy:
> May I change the `naga::SourceLocation.line_position` to count UTF-8 code units instead of general Unicode code points?

Sounds good to me. It makes things more consistent.

CHANGELOG.md
stefnotch: Good point, thank you for catching this

stefnotch:
And it's ready for a (hopefully) final review. I'm glad that I got to learn a ton about wgpu internals.

Thank you for all the reviews and helping me get this PR into a decent shape. <3

Wumpf (approved):
all good from my side 👍 

Thanks again for taking this feature on working through all these iterations! :)

teoxoy (approved):

Wumpf:
uhoh, this went into 0.20 changelog but didn't get release with it

PR #5411, opened 2024-03-18
[naga-cli] Add `input-kind` and `shader-stage` args

ratmice:
**Connections**
None

**Description**
When dealing with .glsl files which have just a plain `.glsl` extension in git.
Renaming things such that they include the shader stage can be inconvenient.

This adds an argument `--shader-stage` and uses the `FromStr` implementation of that for both extension parsing and command line arguments.

This also allows `fragment`, `vertex`, `compute` to be accepted as well. 

In addition this also adds `--input-kind` argument which accepts `glsl`, etc...

:four_leaf_clover: 

**Testing**

```
touch foo{,{,.{bar,frag,fragment}}.glsl}
cargo run -p naga-cli -- foo.frag.glsl
cargo run -p naga-cli -- --shader-stage frag foo.glsl
cargo run -p naga-cli -- --shader-stage frag --input-kind glsl foo
cargo run -p naga-cli -- foo.fragment.glsl
# should error
cargo run -p naga-cli -- foo.bar.glsl
cargo run -p naga-cli -- --shader-stage bar foo.glsl
```

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,7 +68,7 @@ path = "./naga"
 version = "0.19.2"
 
 [workspace.dependencies]
-anyhow = "1.0"
+anyhow = "1.0.3"
 arrayvec = "0.7"
 bit-vec = "0.6"
 bitflags = "2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,7 +68,7 @@ path = "./naga"
 version = "0.19.2"
 
 [workspace.dependencies]
-anyhow = "1.0.3"
+anyhow = "1.0.23"
 arrayvec = "0.7"
 bit-vec = "0.6"
 bitflags = "2"
--- a/naga-cli/src/bin/naga.rs
+++ b/naga-cli/src/bin/naga.rs
@@ -353,7 +353,7 @@ impl fmt::Display for CliError {
 }
 impl std::error::Error for CliError {}
 
-fn run() -> Result<(), anyhow::Error> {
+fn run() -> anyhow::Result<()> {
     env_logger::init();
 
     // Parse commandline arguments
@@ -560,7 +560,7 @@ fn parse_input(
     input_path: &Path,
     input: Vec<u8>,
     params: &Parameters,
-) -> Result<Parsed, anyhow::Error> {
+) -> anyhow::Result<Parsed> {
     let input_kind = match params.input_kind {
         Some(kind) => kind,
         None => input_path
@@ -599,7 +599,7 @@ fn parse_input(
                     let file_stem = input_path
                         .file_stem()
                         .context("Unable to determine file stem from input filename.")?;
-                    // filename.shader_stage -> shader_stage.
+                    // filename.shader_stage -> shader_stage
                     let inner_ext = Path::new(file_stem)
                         .extension()
                         .context("Unable to determine inner extension from input filename.")?
@@ -641,7 +641,7 @@ fn write_output(
     info: &Option<naga::valid::ModuleInfo>,
     params: &Parameters,
     output_path: &str,
-) -> Result<(), anyhow::Error> {
+) -> anyhow::Result<()> {
     match Path::new(&output_path)
         .extension()
         .ok_or(CliError("Output filename has no extension"))?
--- a/naga-cli/src/bin/naga.rs
+++ b/naga-cli/src/bin/naga.rs
@@ -1,4 +1,5 @@
 #![allow(clippy::manual_strip)]
+use anyhow::anyhow;
 #[allow(unused_imports)]
 use std::fs;
 use std::{error::Error, fmt, io::Read, path::Path, str::FromStr};
@@ -300,7 +301,7 @@ impl fmt::Display for CliError {
 }
 impl std::error::Error for CliError {}
 
-fn run() -> Result<(), Box<dyn std::error::Error>> {
+fn run() -> Result<(), anyhow::Error> {
     env_logger::init();
 
     // Parse commandline arguments
@@ -504,7 +505,7 @@ fn parse_input(
     input_path: &Path,
     input: Vec<u8>,
     params: &Parameters,
-) -> Result<Parsed, Box<dyn std::error::Error>> {
+) -> Result<Parsed, anyhow::Error> {
     let (module, input_text) = match Path::new(&input_path)
         .extension()
         .ok_or(CliError("Input filename has no extension"))?
@@ -519,11 +520,11 @@ fn parse_input(
             match result {
                 Ok(v) => (v, Some(input)),
                 Err(ref e) => {
-                    let message = format!(
+                    let message = anyhow!(
                         "Could not parse WGSL:\n{}",
                         e.emit_to_string_with_path(&input, input_path)
                     );
-                    return Err(message.into());
+                    return Err(message);
                 }
             }
         }
@@ -579,7 +580,7 @@ fn write_output(
     info: &Option<naga::valid::ModuleInfo>,
     params: &Parameters,
     output_path: &str,
-) -> Result<(), Box<dyn std::error::Error>> {
+) -> Result<(), anyhow::Error> {
     match Path::new(&output_path)
         .extension()
         .ok_or(CliError("Output filename has no extension"))?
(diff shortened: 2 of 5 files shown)


naga-cli/src/bin/naga.rs
```diff
@@ -505,15 +559,32 @@ fn parse_input(
...
+        None => {
+            let ext = input_path
+                .extension()
+                .ok_or(CliError::from("Input filename has no extension"))?
+                .to_str()
+                .ok_or(CliError::from("Input filename not valid unicode"))?;
+            InputKind::from_str(ext).or_else(|_| {
+                InputKind::from_str(
+                    Path::new(ext)
+                        .extension()
```
ratmice: This second call of `extension()` is unnecessary because it actually `extension()` always returns the last part of the extension e.g. `.glsl` for `.frag.glsl`.  Fixed in [7ac8e39](https://github.com/gfx-rs/wgpu/pull/5411/commits/7ac8e39955695ed9ad1ee39105e0268e3b665625)

ratmice:
I'm a bit uncertain where I like the whole `context` thing, but it does seem to improve things.

One of the things I noticed doing this is the `FromStr` errors typically are embedding the `--argument` information.
But this seems redundant because the parameter parsing code is already including that information.

naga-cli/src/bin/naga.rs
```diff
@@ -259,14 +311,54 @@ fn main() {
...
 impl fmt::Display for CliError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}", self.0)
+        write!(f, "{}", self.msg)?;
+        if let Some(source) = self.source.as_ref() {
+            write!(f, " {}", source)?;
+        }
+        Ok(())
     }
 }
```
cwfitzgerald: This is basically rolling anyhow with extra steps - we probably should just use anyhow.

naga-cli/src/bin/naga.rs
```diff
@@ -465,15 +559,28 @@ fn parse_input(
...
-    {
-        "bin" => (bincode::deserialize(&input)?, None),
-        "spv" => naga::front::spv::parse_u8_slice(&input, &params.spv_in).map(|m| (m, None))?,
-        "wgsl" => {
+    let input_kind = match params.input_kind {
+        Some(kind) => kind,
+        None => InputKind::from_str(
+            input_path
+                .extension()
+                .ok_or(CliError::from("Input filename has no extension"))?
```
cwfitzgerald:
```suggestion
                .context("Input filename has no extension")?
```

With anyhow, this can just be

naga-cli/src/bin/naga.rs
```diff
@@ -487,48 +594,55 @@ fn parse_input(
...
+                    )
+                    .to_str()
+                    .ok_or(CliError::from("Input filename is not valid unicode"))?,
+                )
+                .map_err(|e| {
+                    CliError::from(e).context(CliError::from(format!(
+                        "for input filename {}",
+                        input_path.display()
+                    )))
+                })?,
```
cwfitzgerald: Could we de-nest this. I'd much favor having multiple `let` bindings and variables rather than trying to nest everything in a single expression

naga-cli/src/bin/naga.rs
```diff
@@ -487,48 +594,55 @@ fn parse_input(
                 }
             }
         }
-        ext @ ("vert" | "frag" | "comp" | "glsl") => {
+        InputKind::Glsl => {
+            let shader_stage = match params.shader_stage {
+                Some(shader_stage) => shader_stage,
+                None => ShaderStage::from_str(
```
cwfitzgerald: Once you have `from_str` defined, you can use `my_string.parse()` with return type deduction.

cwfitzgerald (changes requested):
This is definitely needed functionality! Have some comments about errors and code structure.

ratmice:
@cwfitzgerald I believe I have resolved all of your review concerns.

This leaves on the table some additional clean-ups regarding existing errors, and replacing them with `anyhow` usage.
I figured these should be left for a follow up patch.

naga-cli/src/bin/naga.rs
cwfitzgerald: This doesn't seem to have been addressed?

naga-cli/src/bin/naga.rs
cwfitzgerald: Same here

cwfitzgerald (changes requested):
Waiting for anyhow is fine, though it may just be easier to switch without needing to build out the anyhow-like infra. Up to you.

Requesting changes as some changes weren't made.

naga-cli/src/bin/naga.rs
ratmice: Oops. I think I rebased it back onto an old version of the branch when updating to the current trunk :confused:

naga-cli/src/bin/naga.rs
```diff
@@ -300,7 +353,7 @@ impl fmt::Display for CliError {
 }
 impl std::error::Error for CliError {}
 
-fn run() -> Result<(), Box<dyn std::error::Error>> {
+fn run() -> Result<(), anyhow::Error> {
```
cwfitzgerald:
```suggestion
fn run() -> anyhow::Result<()> {
```

ratmice:
Sorry about the branch mixup, this switches it to use anyhow, and should contain the fixes I was referring to previously.

What i'd meant previously is that there are further cleanups to existing code that can probably benefit from using anyhow.
But this patch just does the minimal changes needed for the new code to use `context()` etc.  With the intent to do those cleanups in a follow-up.

naga-cli/src/bin/naga.rs
```diff
@@ -504,71 +560,77 @@ fn parse_input(
...
-        ext @ ("vert" | "frag" | "comp" | "glsl") => {
+        InputKind::Glsl => {
+            let shader_stage = match params.shader_stage {
+                Some(shader_stage) => shader_stage,
+                None => {
+                    // filename.shader_stage.glsl -> filename.shader_stage
+                    let file_stem = input_path
+                        .file_stem()
+                        .context("Unable to determine file stem from input filename.")?;
+                    // filename.shader_stage -> shader_stage.
```
cwfitzgerald:
```suggestion
                    // filename.shader_stage -> shader_stage
```

naga-cli/src/bin/naga.rs
```diff
@@ -579,7 +641,7 @@ fn write_output(
     info: &Option<naga::valid::ModuleInfo>,
     params: &Parameters,
     output_path: &str,
-) -> Result<(), Box<dyn std::error::Error>> {
+) -> Result<(), anyhow::Error> {
```
cwfitzgerald: Same nit

Cargo.toml
```diff
@@ -68,7 +68,7 @@ path = "./naga"
 version = "0.19.2"
 
 [workspace.dependencies]
-anyhow = "1.0"
+anyhow = "1.0.23"
```
cwfitzgerald, line 71:
```suggestion
anyhow = "1.0"
```

cwfitzgerald (changes requested):
Looking a lot better, one more round, then just need a naga person to sign off on it, then we're good to land!

naga-cli/src/bin/naga.rs
cwfitzgerald: Ah yeah, it happens :)

Cargo.toml
ratmice: I believe this version bump was needed to pass CI, If I recall anyhow added a trait implementation of AsRef.
I can't seem to reproduce locally though.

I'll go ahead and bump it back down though and we can see what happens.  The earlier bump to 1.0.3 was also the minimum version which added some other trait impl.

cwfitzgerald: Ah alright, if it's intentional, then that's totally fine.

naga-cli/src/bin/naga.rs
ratmice: I also noticed one more of these but for `anyhow::Result<Parsed>`.

cwfitzgerald (approved):
Great - going to boop the naga people

teoxoy (approved):
Nice!

PR #5421, opened 2024-03-21
fix: don't depend on BG{,L} layout entry order in HAL

ErichDonGubler:
**Connections**

* [Firefox's bug 1877461, comment 3](https://bugzilla.mozilla.org/show_bug.cgi?id=1877461#c3)
* Backport to 0.19: https://github.com/gfx-rs/wgpu/pull/5455

**Description**

Multiple backends for `wgpu-hal` currently make assumptions about the ordering of descriptor entries provided for bind groups and bind group layouts that are invalid. In particular, it was assumed that the _element order_ of resource entries stored in `Vec`s in bind group layouts were the same for bind group resources entries stored in `Vec`s. This caused _binding slot order_ to become out-of-sync when actually creating bind groups from descriptors and bind group layouts with code like `desc.entries.iter().zip(desc.layout.entries.iter())`, bypassing essential validation for, i.e., type matching in resource entries, and usage masks for buffers and textures. At best, this caused crashes because of different numbers of types of resources being used (as observed in [Firefox's bug 1877461, comment 3](https://bugzilla.mozilla.org/show_bug.cgi?id=1877461#c3)). At worst, this caused resources of (of the same _and_ different) types to be bound in the wrong places. We in Firefox noticed because of crashes and putting graphics drivers into invalid states. I'm sure there are other potential types of damage that are even worse. 😬

**Testing**

The test `wgpu_test::different_bgl_order_bw_shader_and_api` has been added, which exercises the previously crashing case where more than one resource type was being used in a bind group layout. This doesn't cover cases of resources of the same type having their order mixed up, but it should be sufficient to prevent a significant regression.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-hal/src/gles/device.rs
+++ b/wgpu-hal/src/gles/device.rs
@@ -1125,8 +1125,10 @@ impl crate::Device for super::Device {
                 !0;
                 bg_layout
                     .entries
-                    .last()
-                    .map_or(0, |b| b.binding as usize + 1)
+                    .iter()
+                    .map(|b| b.binding)
+                    .max()
+                    .map_or(0, |idx| idx as usize + 1)
             ]
             .into_boxed_slice();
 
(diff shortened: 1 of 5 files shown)


wgpu-hal/src/gles/device.rs
```diff
@@ -1179,7 +1181,16 @@ impl crate::Device for super::Device {
...
+        let layout_and_entry_iter = desc.entries.iter().map(|entry| {
+            let layout = desc
+                .layout
+                .entries
+                .iter()
+                .find(|layout_entry| layout_entry.binding == entry.binding)
+                .expect("internal error: no layout entry found with binding slot");
+            (entry, layout)
+        });
+        for (entry, layout) in layout_and_entry_iter {
```
ErichDonGubler, line 1193: Face-to-face discussion with @jimblandy: This is a quadratic (i.e. `O(n^2)`) algorithm. Not sure if this will be a perf. issue, but it's remarkable here since this is in the path to constructing passes. We should file a separate issue for perf. follow-up.

jimblandy (approved):
We need a followup bug for the quadratic behavior, but other than that, this looks fantastic. Thank you so much for writing a test.

Edit from @ErichDonGubler: Added in https://github.com/gfx-rs/wgpu/issues/5473.

wgpu-hal/src/gles/device.rs
ErichDonGubler: Filed: https://github.com/gfx-rs/wgpu/issues/5473

PR #5424, opened 2024-03-22
Add parsing support for un/pack4xI/U8

### LINKED ISSUE #5423: Implement new builtin packing/unpacking functions from WGSL spec
I tried using [unpack4xU8](https://www.w3.org/TR/WGSL/#unpack4xU8-builtin) and [pack4xU8](https://www.w3.org/TR/WGSL/#pack4xU8-builtin), but naga refused to parse them

I would like them to be implemented

There is a problem, though. I tried implementing them myself and have found that [SPIR-V GLOp](https://registry.khronos.org/SPIR-V/specs/unified1/GLSL.std.450.html) doesn't have the same functions. I don't know how closely this repo is affiliated with the group that writes WGSL spec, so I don't know if they can be pressured to remove these functions from the spec so would WGSL more closely resemble SPIR-V.

**Additional context**
I have written parsing support that you can see in a PR that I will open referencing this issue.

VlaDexa:
**Connections**
- Closes #5423

**Description**
WGSL spec defined new builtin functions. This PR tries to implement them

Current writer impls:
- [x] SPIR-V
- [x] HLSL
- [x] GLSL
- [x] MSL
- [x] WGSL

**Testing**
I think I added the necessary tests to `naga/tests/in/bits.wgsl`. 

**Checklist**

- [x] Run `cargo fmt`.


VlaDexa:
I have no idea how to use those magic words that make the spir-v backend, so I really would appreciate some help here

teoxoy:
I would look at some of the existing polyfills. For ex.:

https://github.com/gfx-rs/wgpu/blob/c613fbbeedc4c171a4bba66e1f4fc62f110c9102/naga/src/back/spv/block.rs#L734

https://github.com/gfx-rs/wgpu/blob/c613fbbeedc4c171a4bba66e1f4fc62f110c9102/naga/src/back/spv/block.rs#L987

VlaDexa:
Currently sign extension for unpacking only works in SPIR-V writer

VlaDexa:
@teoxoy Hey, it's been a week since this pr has been finished. Do I need to signal someone that it's ready ready or just wait for it to be reviewed, squashed and merged?

teoxoy:
It's on my radar, there have been other PRs that need/needed attention as well.
I should get to it these coming days.

teoxoy:
I squashed all the commits, rebased and extracted the match refactor to make it easier to review.

naga/src/back/glsl/mod.rs
```diff
@@ -3463,12 +3468,66 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                            };
+                            write!(self.out, ", {}, 8)", i * 8)?;
+                            if i != 3 {
+                                write!(self.out, ", ")?;
+                            }
+                        }
+                        write!(self.out, ")")?;
+
+                        return Ok(());
+                    }
```
teoxoy, line 3649: Can we generate the same code as we do for the MSL and HLSL backends?

naga/src/back/spv/block.rs
```diff
@@ -1201,11 +1201,156 @@ impl<'w> BlockContext<'w> {
...
+                        MathOp::Custom(Instruction::quaternary(
+                            spirv::Op::CompositeConstruct,
+                            result_type_id,
+                            id,
+                            parts[0],
+                            parts[1],
+                            parts[2],
+                            parts[3],
+                        ))
+                    }
```
teoxoy, line 1355: Is there a reason the SPIR-V backend is using a different approach than the others?

naga/src/back/spv/block.rs
```diff
@@ -1201,11 +1201,156 @@ impl<'w> BlockContext<'w> {
...
+                                let casted = self.gen_id();
+                                block.body.push(Instruction::unary(
+                                    spirv::Op::Bitcast,
+                                    uint_type_id,
+                                    casted,
+                                    extracted,
+                                ));
+                                extracted = casted;
+                            }
+                            let new_preresult = self.gen_id();
```
teoxoy: We could check if this is the last iteration and use `id` instead (avoiding the `OpCopyObject`).

naga/src/back/spv/block.rs
```diff
@@ -1201,11 +1201,156 @@ impl<'w> BlockContext<'w> {
...
+
+                        MathOp::Custom(Instruction::quaternary(
+                            spirv::Op::CompositeConstruct,
+                            result_type_id,
+                            id,
+                            parts[0],
+                            parts[1],
+                            parts[2],
+                            parts[3],
+                        ))
```
teoxoy:
```suggestion
                        MathOp::Custom(Instruction::composite_construct(
                            result_type_id,
                            id,
                            &parts
                        ))
```

teoxoy (commented):
Since the code for the MSL, GLSL and HLSL backends is the same, we should try to put it in the `back` module instead.

naga/src/back/spv/block.rs
VlaDexa: It uses the same approach as the glsl backend. The msl and hlsl backends were written with bitshifting logic because they lack bitfieldExtract function.

naga/src/back/glsl/mod.rs
VlaDexa: Yeah, it can be done, but I assume that using bitfieldExtract is cleaner than using bitshifts to emulate sign extension. Also, does stuff in less instructions. If it's important to keep all of the implementations as similar as possible, I can rewrite them.

teoxoy: I usually lean towards consistency. I thought `bitfieldExtract` will get lowered into bitshifts anyway but it seems at least nvidia and amd have instructions in their ISA that correspond to `bitfieldExtract`. Even so I would imagine the hardware is more complex for `bitfieldExtract` since it's a more complex operation.

I propose we go with bitshifts everywhere unless `bitfieldExtract` is faster.

teoxoy:
It would be great if you could put together some execution tests that make sure the polyfills work properly.

There are some existing ones at this path: https://github.com/gfx-rs/wgpu/tree/54740d5982fa5ae004eee95dd6c037356ce9ea17/tests/tests/shader.

naga/src/back/glsl/mod.rs
cwfitzgerald: As far as I know bfi is faster, multiple shifts require multiple instructions, and bfi and shifts should work at the same rate.

VlaDexa: Still, we need to keep in mind that less instructions doesn't always mean that it's going to be faster. Maybe one shift uses one microinstruction and one bitfield operation uses 6, if not optimized on hardware level. So I agree that this choice requires benchmarking. 
The problem I have right now is that I don't have that much hardware to test on, so it may be good on my system, but worse on many others.
Also, in the future, bitfield operations may become faster if they are slow now, but that seems more like a problem for future people.

teoxoy: Let's keep things as they are for now since it's not clear what is best.

tests/tests/shader/data_builtins.rs
```diff
@@ -0,0 +1,127 @@
...
+#[allow(non_snake_case)]
+fn create_unpack4xU8_test() -> Vec<ShaderTest> {
+    let mut tests = Vec::new();
+
+    let input: u32 = 0xAABBCCDD;
+    let output: [u32; 4] = [0xDD, 0xCC, 0xBB, 0xAA];
+    let unpack_u8 = ShaderTest::new(
+            format!("unpack4xU8({input:x?}) == {output:x?}"),
+            String::from("value: u32"),
+            String::from("let a = unpack4xU8(input.value); output[0] = a[0];output[1] = a[1];output[2] = a[2];output[3] = a[3];"),
```
cwfitzgerald: For readability let's make this multiple lines
```suggestion
            String::from("
                let a = unpack4xU8(input.value);
                output[0] = a[0];
                output[1] = a[1];
                output[2] = a[2];
                output[3] = a[3];
            "),
```

tests/tests/shader/data_builtins.rs
```diff
@@ -0,0 +1,127 @@
...
+        (0x11223344, [0x44, 0x33, 0x22, 0x11]),
+        // sign extension
+        (0xFF, [-1, 0, 0, 0]),
+    ];
+
+    for (input, output) in values {
+        let unpack_i8 = ShaderTest::new(
+            format!("unpack4xI8({input:x?}) == {output:x?}"),
+            String::from("value: u32"),
+            String::from("let a = bitcast<vec4<u32>>(unpack4xI8(input.value)); output[0] = a[0];output[1] = a[1];output[2] = a[2];output[3] = a[3];"),
```
cwfitzgerald: Same here

tests/tests/shader/data_builtins.rs
```diff
@@ -0,0 +1,127 @@
...
+    )
+    .run_async(|ctx| {
+        shader_input_output_test(ctx, InputStorageType::Storage, create_pack4xU8_test())
+    });
+
+#[allow(non_snake_case)]
+fn create_pack4xI8_test() -> Vec<ShaderTest> {
+    let mut tests = Vec::new();
+
+    let input: [u32; 4] = [0xDD, 0xCC, 0xBB, 0xAA];
```
cwfitzgerald: Might test something about sign behavior here?

cwfitzgerald (changes requested):
Small nits, but wgpu/test side looks good, by and large!

Sorry for the delay on the naga review, all the naga reviewers are pretty busy right now.

tests/tests/shader/data_builtins.rs
VlaDexa: Done!

tests/tests/shader/data_builtins.rs
VlaDexa: Done!

tests/tests/shader/data_builtins.rs
VlaDexa: Since the packing operation is essentailly just copying the last 8 bits of the number, signess of a value shouldn't matter. But still, more tests could never hurt, so, done!

tests/tests/shader/data_builtins.rs
VlaDexa: Weird quotes placement is enforced by rustfmt, so it isn't exactly as your suggested change, but I assume that's okay

cwfitzgerald: Yeah, that's fine.

tests/tests/shader/data_builtins.rs
cwfitzgerald: Love the attitude, thanks!

cwfitzgerald (approved):
Approved from wgpu side :)

teoxoy (approved):
Looks good!

teoxoy:
Could you squash the last 3 commits so that we end up with the refactor + the implementation? We can then land this without squashing it.

VlaDexa:
Done

PR #5400, opened 2024-03-16
snatch: Use a reentrant but unfair read lock

SludgePhD:
**Connections**
Should fix https://github.com/gfx-rs/wgpu/issues/5279 

**Description**
`RwLock`s typically use some mechanism to ensure that readers cannot starve waiting writers, even if they continually re-acquire new read locks. This means that attempts to acquire a read lock can block when a writer is trying to acquire the lock, even if there is already an existing read lock.

This mechanism has the unfortunate consequence of making deadlocks possible the instant a single thread wants to acquire 2 reads locks at the same time (and wgpu is full of code that does that).

Luckily, parking_lot also offers an explicitly reentrant but unfair way to acquire a read lock. This PR switches the snatch lock to use that, hopefully resolving the deadlock.

**Testing**
Ran my workload. It didn't deadlock. Previously it deadlocked within a fraction of a second.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`. If applicable, add:
  - [ ] `--target wasm32-unknown-unknown`
  - [ ] `--target wasm32-unknown-emscripten`
- [x] Run `cargo xtask test` to run tests.
- [ ] Add change to `CHANGELOG.md`. See simple instructions inside file.

PR #5426, opened 2024-03-22
Avoid recursive snatch lock acquisitions

SludgePhD:
**Connections**
Bug report: https://github.com/gfx-rs/wgpu/issues/5279

This PR is an alternative to https://github.com/gfx-rs/wgpu/pull/5400.

**Description**
Pass `&SnatchGuard` down to all functions needing it, instead of having them recursively acquire one.

This also adds some instrumentation to the `SnatchLock` API that will panic on any attempts to acquire it recursively (gated behind `cfg(debug_assertions)`). That should allow the test suite to catch any future misuse of the API, instead of silently making wgpu more deadlock-prone in multi-threaded applications.

**Testing**
`cargo xtask test` locally and on CI

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -188,11 +188,19 @@ impl Global {
                 hal::BufferUses::empty()
             } else if desc.usage.contains(wgt::BufferUsages::MAP_WRITE) {
                 // buffer is mappable, so we are just doing that at start
+                let snatch_guard = device.snatchable_lock.read();
                 let map_size = buffer.size;
                 let ptr = if map_size == 0 {
                     std::ptr::NonNull::dangling()
                 } else {
-                    match map_buffer(device.raw(), &buffer, 0, map_size, HostMap::Write) {
+                    match map_buffer(
+                        device.raw(),
+                        &buffer,
+                        0,
+                        map_size,
+                        HostMap::Write,
+                        &snatch_guard,
+                    ) {
                         Ok(ptr) => ptr,
                         Err(e) => {
                             to_destroy.push(buffer);
@@ -2008,9 +2016,10 @@ impl Global {
                 }
 
                 // Wait for all work to finish before configuring the surface.
+                let snatch_guard = device.snatchable_lock.read();
                 let fence = device.fence.read();
                 let fence = fence.as_ref().unwrap();
-                match device.maintain(fence, wgt::Maintain::Wait) {
+                match device.maintain(fence, wgt::Maintain::Wait, &snatch_guard) {
                     Ok((closures, _)) => {
                         user_callbacks = closures;
                     }
@@ -2120,9 +2129,10 @@ impl Global {
         device: &crate::device::Device<A>,
         maintain: wgt::Maintain<queue::WrappedSubmissionIndex>,
     ) -> Result<DevicePoll, WaitIdleError> {
+        let snatch_guard = device.snatchable_lock.read();
         let fence = device.fence.read();
         let fence = fence.as_ref().unwrap();
-        let (closures, queue_empty) = device.maintain(fence, maintain)?;
+        let (closures, queue_empty) = device.maintain(fence, maintain, &snatch_guard)?;
 
         // Some deferred destroys are scheduled in maintain so run this right after
         // to avoid holding on to them until the next device poll.
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -188,11 +188,11 @@ impl Global {
                 hal::BufferUses::empty()
             } else if desc.usage.contains(wgt::BufferUsages::MAP_WRITE) {
                 // buffer is mappable, so we are just doing that at start
-                let snatch_guard = device.snatchable_lock.read();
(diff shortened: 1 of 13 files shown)


wgpu-core/src/device/global.rs
```diff
@@ -188,11 +188,19 @@ impl Global {
                 hal::BufferUses::empty()
             } else if desc.usage.contains(wgt::BufferUsages::MAP_WRITE) {
                 // buffer is mappable, so we are just doing that at start
+                let snatch_guard = device.snatchable_lock.read();
```
Wumpf: should only be takein inside the else block, no need to bother that lock otherwise

wgpu-core/src/device/queue.rs
```diff
@@ -1147,7 +1149,9 @@ impl Global {
 
             let device = queue.device.as_ref().unwrap();
 
-            let mut fence = device.fence.write();
+            let snatch_guard = device.snatchable_lock.read();
+
+            let mut fence = device.fence.write(); // snatch(r) -> fence(w)
```
Wumpf: I don't understand this comment and the move of the snatch guard. Having worked with wgpu for a while can speculate but it's not clear at all why the ordering is important, please elaborate a bit here.

wgpu-core/src/device/resource.rs
```diff
@@ -97,6 +97,7 @@ pub struct Device<A: HalApi> {
     pub(crate) command_allocator: Mutex<Option<CommandAllocator<A>>>,
     //Note: The submission index here corresponds to the last submission that is done.
     pub(crate) active_submission_index: AtomicU64, //SubmissionIndex,
+    // NOTE: if both are needed, the `snatchable_lock` must be acquired before the `fence`
```
Wumpf: similar to the note on queue.rs: this should elaborate a little bit on why this is important :)

wgpu-core/src/snatch.rs
```diff
@@ -59,6 +64,10 @@ impl<T> std::fmt::Debug for Snatchable<T> {
 
 unsafe impl<T> Sync for Snatchable<T> {}
 
+thread_local! {
+    static READ_LOCK_LOCATION: Cell<Option<(&'static Location<'static>, Backtrace)>> = const { Cell::new(None) };
+}
```
Wumpf, line 69: I don't quite trust rust to optimize this out when it's not used and thread locals can be a valuable resource. -> Can we put this entire thing under `#[cfg(debug_assertions)]`?

wgpu-core/src/snatch.rs
```diff
@@ -76,7 +85,24 @@ impl SnatchLock {
     }
 
     /// Request read access to snatchable resources.
+    #[track_caller]
     pub fn read(&self) -> SnatchGuard {
+        if cfg!(debug_assertions) {
+            let caller = Location::caller();
+            let backtrace = Backtrace::force_capture();
```
Wumpf: capturing a backtrace can be very expensive operation and we do a lot of snatchlock reading accross the codebase. People should expect debug assertions to not degrate performance that much, so let's do only
```suggestion
            let backtrace = Backtrace::capture();
```
to make the backtrace opt-in like most backtraces in Rust.

wgpu-core/src/snatch.rs
```diff
@@ -76,7 +85,24 @@ impl SnatchLock {
     }
 
     /// Request read access to snatchable resources.
+    #[track_caller]
     pub fn read(&self) -> SnatchGuard {
+        if cfg!(debug_assertions) {
```
Wumpf, line 90: very much appreciating this debugging mechanism!

Wumpf (changes requested):
Thank you very much for this fix!
Indeed not as nice to read as a solution than the recursive lock, but I think @jimblandy is right, this version is the way to go.
Extra thanks for adding the debug assertion in there to capture this sooner in the future.

Bunch of smaller things to fix up and then good to go!
Please also add a changelog entry about this fix.

wgpu-core/src/snatch.rs
SludgePhD: A thread local variable is just turned into an LLVM global, it has no trouble optimizing it out when not used (just like for other types of globals, like `static`s and literals)

Wumpf: I'm no expert in this, but that doesn't sound at all - the key might be just a regular global, but at least on some platforms (and circumstances) using a thread local means that there's an allocation of a thread local slot in the thread (see e.g. `pthread_key_create` to obtain a key in the first place; I guess Rust might use a single key for several thread locals and then "subkey", but I figure it has to happen at least once). If this was just a regular global variable, how would allocation per thread work if not being directed to a per thread table of sorts.

Anyways, tried a bit in CompilerExplorer and it seems to optimize out fine as long as the thread local is not `pub`, so the compiler knows that it's indeed not in use.
So no point in dying on this hill regardless (:

Wumpf (changes requested):
lgtm, but changelog entry is still missing!

Wumpf (approved):

Issue #5124, opened 2024-01-23
Make wgpu-core less generic (brainstorm)


nical:
I'll use this issue to explore various ways we could make wgpu-core less generic.

The motivations include reducing the clutter of overly generic code and more importantly to reduce binary size, compile times.

# Factor out non-generic code into separate functions

There is some amount of code that could be easily factored out, for example some of the validation. Isolating the validation also has the advantage that it can be laid out to match the specification so that we can more easily check that the validation is correct.
There is only a limited amount of bang for buck we can get

# Making the resources not generic

We'll get the biggest wins by far if we can erase the type of the resources. most of wgpu-core is generic only because something needs to hold on to a generic resource.

## Downcasting

Assuming we can accept the cost of downcasting every resource at the top of every hal functions, we could use trait objects and downcast via `Any`.

```rust
// in wgpu-hal:

/// Implemented by most hal types
pub trait Resource: 'static {
    fn as_any(&self) -> &dyn std::any::Any;
}

// Helper for downcasting. Failure to downcast is always an internal error
pub fn downcast<T:'static>(resource: &dyn Resource) -> &T {
    resource.as_any().downcast_ref::<T>().unwrap()
}
``` 

The gpu-hal APIs would take `&dyn Resource` parameters instead of some associated types, and the first thing they would do is downcast these dyn traits into the concrete types they are expecting.

Note that we call also break it down into plenty of traits like `TextureResource`, `BufferResource`, etc. That's a rather small detail once we have overcome the other obstacles.

## Boxed resources

The simple approach (which I find very unsatisfying) is to box all hal resources:

```rust
// in wgpu-core:

pub struct BindGroupLayout {
    // ...

    raw: Box<dyn hal::Resource>,
}
```

The wgpu-core and wgpu-hal structures are no longer next to each other in memory, that most likely has a cost but it is hard to estimate it without doing all of the work.

## Dynamically sized types

The last member of a struct can be a DST, which turns the structure into a dst as well:

```rust
pub struct BindGroupLayoutResource<Hal: ?Sized> {
    // ...
    pub raw: Hal
}

// The alias that is used in most places
pub type BindGroupLayout = BindGroupLayoutResource<dyn hal::Resource>;

pub fn create_bind_group_layout(device: &Device) -> Result<Arc<BindGroupLayout>, Error> {
    let typed_bgl = Arc::new(BindGroupLayoutResource {
        label: String::new(),
[...]

PR #5427, opened 2024-03-22
[wgpu-hal] migrate `A` from a generic parameter to an associated type

lylythechosenone:
**Connections**
#5124 

**Description**
As seen in https://github.com/gfx-rs/wgpu/issues/5124#issuecomment-1926553806, in order to blanket implement a `DynDevice` trait (and by extension any other `Dyn____` trait), `A` must be an associated type rather than a generic.

While this *is* a breaking change, it is likely not very much of a deal, as from what I can tell, the only code it would break is mine (another project). It should only affect projects that directly implement `wgpu-hal` traits, or those that use `wgpu-hal` traits in bounds. However, I do not believe any such projects exist, excluding mine.

**Testing**
This should only cause compile-time differences. It may be worth checking to see if these changes would break anything on crates.io.

**Checklist**

- [x] Run `cargo fmt`.

(I ran the changes through the repository's CI before creating this PR)

### DIFF
--- a/wgpu-hal/src/dx12/adapter.rs
+++ b/wgpu-hal/src/dx12/adapter.rs
@@ -432,7 +432,9 @@ impl super::Adapter {
     }
 }
 
-impl crate::Adapter<super::Api> for super::Adapter {
+impl crate::Adapter for super::Adapter {
+    type A = super::Api;
+
     unsafe fn open(
         &self,
         _features: wgt::Features,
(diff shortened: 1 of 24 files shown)


Wumpf (approved):
thanks for taking this task on and doing this incremental as @nical suggested
Read up a bit on the ticket and this seems like a great first step!

Wumpf:
> It may be worth checking to see if these changes would break anything on crates.io.

given that this didn't require any changes to wgpu-core I'm not worried. People who are that deep in won't find it hard to understand this change. Might have warranted a changelog entry though. Gonna merge this right away though, we can still add more changelog entries in a later PR.

PR #5429, opened 2024-03-23
Expose all ray-query flags in constants

kvark:
**Connections**
Port of https://github.com/gfx-rs/naga/pull/2375

**Description**
We are making the ray query flags a part of the API here.
Also bringing the basics of candidate intersections.

**Testing**
Not really tested.

**Checklist**

- [x] Run `cargo fmt`.


naga/src/lib.rs
```diff
@@ -2087,3 +2087,56 @@ pub struct Module {
...
+        /// Intersecting with triangles..
+        /// Matches RayQueryCommittedIntersectionTriangleKHR and RayQueryCandidateIntersectionTriangleKHR.
+        const TRIANGLE = 0x1;
+        /// Intersecting with generated primitives.
+        /// Matches RayQueryCommittedIntersectionGeneratedKHR.
+        const GENERATED = 0x2;
+        /// Intersecting with Axis Aligned Bounding Boxes.
+        /// Matches RayQueryCandidateIntersectionAABBKHR.
+        const AABB = 0x4;
+    }
```
teoxoy: The value of `RayQueryCandidateIntersectionTriangleKHR` is 0 and the value of `RayQueryCandidateIntersectionAABBKHR` is 1. I think we need different types for candidate and committed intersections.

image

teoxoy (changes requested):

cwfitzgerald:
Boop from triage

naga/src/lib.rs
kvark: Yes, some flags are different between committed/candidate. The idea here is that we know (at shader compile time) what kind of intersection this is, and we can inject code to transform the flag from SPIR-V semantics into WGSL. Having less different flags overall seems like a win, and the cost of the transform is negligible since we are talking about a ray query nearby, anyway.

teoxoy: I see but we don't do that as part of this PR so the flags won't work properly.

kvark: Note that the candidate intersections are not implemented yet:
```rust
                if !committed {
                    return Err(Error::FeatureNotImplemented("candidate intersection"));
                }
```

kvark: Basic implementation is now a part of this PR, please take another look!

kvark:
PR has seen a substantial revamp now. Please see it again!

teoxoy (approved):
Looks good!

Issue #1453, opened 2020-03-03
Lifetimes on `RenderPass` make it difficult to use. 


aloucks:
Related PRs:

https://github.com/gfx-rs/wgpu-rs/pull/155
https://github.com/gfx-rs/wgpu-rs/pull/168

For context, I was looking at updating [imgui-wgpu](https://github.com/Yatekii/imgui-wgpu-rs/blob/master/src/lib.rs#L369) to work with the current master.

Now that `set_index_buffer` (and similar methods) take the a `&'a Buffer` instead of `&Buffer`, it has a ripple effect of "infecting" the surrounding scope with this lifetime.

While attempting to iterate and add draw calls, you end up with either "multiple mutable borrow" errors on wherever you're storing the `Buffer` or similar lifetime errors like "_data from `self` flows into `rpass` here_"

I would think that by calling `set_index_buffer` (and similar methods) that it would increment the ref-count on the buffer so that this lifetime restriction isn't needed.

Birbe:
What's the status of this? I think it would be extremely useful for me. I'm writing a renderer that would allow users to create custom pipelines, and they'd be provided with the state of the renderer using a trait. The issue is that I'm getting massive lifetime issues that are caused by the fact that I'm using a lot of `RwLock`s in my renderer state, meaning that within that trait, I cannot give `RenderPass` data from within a RwLock guard, as that data gets dropped at the end of the trait. If I had a way to instead pass in `Arc`s of buffers, this would clear up all of my issues.

kvark:
At this point, any improvement here should be blocked on @pythonesque work in Arc-anizing the internals.

cwfitzgerald:
This is now blocked on #2710.

swiftcoder:
This is still pretty painful. The interactions with this lifetime mean that any object which embeds wgpu resources becomes untouchable the instant you record one of those resources into a renderpass. And since we can't clone wgpu resource ids in rust today, there's no way to work around this with temporary handles on the application side.

For a non-trivial application, I'm ending up having to wrap all the wgpu resource types in wrapper objects, define new handles for them, and pass those around in my code...

Birbe:
> This is still pretty painful. The interactions with this lifetime mean that any object which embeds wgpu resources becomes untouchable the instant you record one of those resources into a renderpass. And since we can't clone wgpu resource ids in rust today, there's no way to work around this with temporary handles on the application side.
> 
[...]

PR #5432, opened 2024-03-24
Separate out ComputeCommand id->arc resolve (a step towards no lifetimes on `wgpu::ComputePass`)

Wumpf:
**Connections**
* part of #1453

Part of a series towards lifetime removal:
* #5409
* ➡️ #5432
* #5570
* #5575

**Description**

----------

## Background

This is working towards removing the dreaded lifetimes on `wgpu::ComputePass` & `wgpu::RenderPass`. I'm starting with compute passes since they have a much smaller API surface area.

Today, the lifetime on the wgpu objects protect us from dropping resources that may be used during pass execution. For example if one passes a pipeline to a render/compute pass, it musn't be dropped until the pass is executed/ended since the pass currently only remembers the pipeline's id.
To allow dropping it (and thus removing the lifetime constraint), we have to bump the reference count of all resources going into a pass upon recording and not (as done today) upon finishing the pass. Ever since Arcanization landed this is technically possible with manageable overhead.

## This PR

This PR introduces `ArcComputeCommand` which is to `ComputeCommand` like the existing `ArcRenderCommand` to `RenderCommand` (`ArcRenderCommand` is only used in RenderBundle today, but this likely going to change soon as well!).

The goal in a **follow-up PR** is for `ComputePass` to record `ArcComputeCommand` directly, reserving `ComputeCommand` solely for tracing.
Instead, in this iteration recording operations on `ComputePass` still push `ComputeCommand` into an internal vector which then is converted just before finish to `ArcComputeCommand`, thus separating out id->Arc resolve of the actual execution. -> Only recording replay will need to do this bulk conversion in the future, but it's not going away until ids are gone. The pass execution itself now relies already on `ArcComputeCommand`.

### Alternatives considered

The original idea was to extend `ComputeCommand` to hold both ids as well as optional Arc in order to avoid duplication of the command enum. However what made me decide differently:
* This would have required making `ComputeCommand` generic over a `HalApi` making the trace infrastructure infected with this. It looked to me like some type erasure would have been needed
    *  @lylythechosenone 's ongoing work on reducing generics will likely alleviate that
* There's precedence with `ArcRenderCommand`
* While the redundancy in the type isn't pretty, keeping recording separate from non-recording is quite advantageous in many ways: the actually used command enum doesn't get any bulkier for the non-trace case and the trace-only datastructures can be separated out and stay under the recording feature (-> `ArcComputeCommand` can become just `ComputeCommand` and what's `ComputeCommand` today becomes `TraceComputeCommand`!)

-----------

**Testing**
Regular unit tests run compute, this should cover this well enough.

**Checklist**

- [x] Run `cargo fmt`.
- ~[ ] Add change to `CHANGELOG.md`. See simple instructions inside file.~


wgpu-core/src/command/compute.rs
```diff
@@ -483,13 +458,13 @@ impl Global {
         let mut pending_discard_init_fixups = SurfacesInDiscardState::new();
 
         for command in base.commands {
-            match *command {
-                ComputeCommand::SetBindGroup {
+            match command.clone() {
```
cwfitzgerald: Why do we need to clone this here?

wgpu-core/src/command/compute_command.rs
```diff
@@ -0,0 +1,324 @@
...
+                    }
+                })
+            })
+            .collect::<Result<Vec<_>, ComputePassError>>()?;
+        Ok(resolved_commands)
+    }
+}
+
+/// Equivalent to `ComputeCommand` but the Ids resolved into resource Arcs.
+#[doc(hidden)]
```
cwfitzgerald: Why are these doc hidden?

wgpu-core/src/track/metadata.rs
```diff
@@ -87,16 +87,19 @@ impl<T: Resource> ResourceMetadata<T> {
...
     #[inline(always)]
-    pub(super) unsafe fn insert(&mut self, index: usize, resource: Arc<T>) {
+    pub(super) unsafe fn insert(&mut self, index: usize, resource: Arc<T>) -> &Arc<T> {
         self.owned.set(index, true);
-        unsafe {
-            *self.resources.get_unchecked_mut(index) = Some(resource);
-        }
+        let resource_dst = unsafe { self.resources.get_unchecked_mut(index) };
+        *resource_dst = Some(resource);
+        resource_dst.as_ref().unwrap()
```
cwfitzgerald: Isn't this just Option::insert?

cwfitzgerald (changes requested):
Looking good, have some comments

wgpu-core/src/command/compute.rs
Wumpf: fixed now in [708890bd626ed1f9b4391fd42608a0d0f7ba2ed4](https://github.com/gfx-rs/wgpu/pull/5432/commits/708890bd626ed1f9b4391fd42608a0d0f7ba2ed4) with a comment that we should actually be draining the command queue instead - I have to do fine grained copies inside the match now, but looking at it it's actually still less copying overall, so a win.

this was mostly laziness on my part so far

wgpu-core/src/command/compute_command.rs
Wumpf: Regarded this as an internal type. But not really needed. Removed doc hiding!

wgpu-core/src/track/metadata.rs
Wumpf: oh. nice. Not one I use very often

wgpu-core/src/command/compute.rs
Wumpf: have a follow-up (to the other follow-up ;)) in the works that addresses this better with less copies on the non-recording-replay path https://github.com/Wumpf/wgpu/tree/compute-pass-end-refactor

EDIT: that's it's own pr by now! https://github.com/gfx-rs/wgpu/pull/5575

wgpu-core/src/command/compute.rs
```diff
@@ -347,19 +294,43 @@ impl Global {
...
+                commands: &resolved_commands,
+                dynamic_offsets: base.dynamic_offsets,
+                string_data: base.string_data,
+                push_constant_data: base.push_constant_data,
+            },
+            timestamp_writes,
+        )
+    }
+
+    #[doc(hidden)]
```
cwfitzgerald: This is unnecessary - it's already private.

cwfitzgerald (approved):
Nice!

PR #5435, opened 2024-03-25
[fix bug #5430] index_limit judgment error in drawIndexed(indexCount)

microaaron:
[fix bug #5430] index_limit judgment error in drawIndexed(indexCount)

**Connections**
#5430

**Description**
index_limit judgment error in drawIndexed(indexCount). 
An error should be reported only when last_index > index_limit.
https://github.com/gfx-rs/wgpu/blob/e102e59e477c54dd9c8913438b92ad97a81c491a/wgpu-core/src/command/bundle.rs#L169-L174

**Testing**
I'm not sure how to test it. Do I need to compile the entire firefox?

**Checklist**

- [ ] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -166,7 +166,7 @@ fn validate_indexed_draw<A: HalApi>(
 ) -> Result<(), DrawError> {
     let last_index = first_index as u64 + index_count as u64;
     let index_limit = index_state.limit();
-    if last_index <= index_limit {
+    if last_index > index_limit {
         return Err(DrawError::IndexBeyondLimit {
             last_index,
             index_limit,


Wumpf (changes requested):
as mentioned on the ticket, I think we should really add a regression test on this one

Wumpf:
> I'm not sure how to test it. Do I need to compile the entire Firefox?

no. For test instructions see the README. Any updates to wgpu will eventually be picked up by Firefox at some point, but wgpu is tested & released independently. Essentially, being the backbone of Firefox's WebGPU implementation is one of wgpu's purposes (one of the most important ones) but not the full story.

microaaron:
> no. For test instructions see the README. 

4failed
捕获4failed

Wumpf:
there should be more log describing the actual test failures. Not sure if we pass all tests right now on Windows right now, might depend also on installed driver version (we do run them on ci afaik, so that's fine) - this _is_ concerning, but in any case an orthogonal issue. After all, as long as we don't add a unit test that covers this issue this won't change the outcome much.

Don't worry if you're struggling to find how and where to do that, we'll definitely get this fix landed before the next release given that it's fairly critical (but if you want to learn have a browse through the existing regression tests - but I figure you ran into this via javascript from the browser end of things, so for this you'd need to code against wgpu and that's quite a lot to take in if you're new to it). Good chance I get to it later this week.

microaaron:
Thanks a lot.

Wumpf:
took this over in https://github.com/gfx-rs/wgpu/pull/5441

PR #5436, opened 2024-03-25
Object-safe traits within wgpu-hal

lylythechosenone:
**Connections**
This is the next step forward for #5124 

**Description**
This PR adds object-safe versions of the wgpu-hal traits, and blanket implements them for types that implement the default traits. It includes small breaking changes for users of wgpu-hal. Acceleration structures remain unimplemented currently, although they are not used in core anyway.

**Testing**
Currently, there are no tests. Once this is fully integrated into core, it will use core's tests.

**Checklist**

- [x] Run `cargo fmt`.


lylythechosenone:
Actually @Wumpf hold before you merge this. I may have (perhaps annoyingly) found a way to entirely eliminate it. If you'd like, you can move this to a draft.

Edit: damn it, this might also require reverting my other PR! Let me keep looking into it.

lylythechosenone:
Alright. I have to figure out what causes that ICE. Once I do that, I'll probably remove the `dynamic` module and move it into core, with minor changes. I'll illustrate my full plan on the issue if I have time. Let me know if you know the cause of this ICE.

lylythechosenone:
Alright @Wumpf if this check passes, this should be fine to merge.

lylythechosenone:
Wait, never mind, back to draft I go!

EDIT: this was a mistake, please ignore.

EDIT 2: I was actually right.

lylythechosenone:
I think that should do it!

#5124: all of the hal steps should now be done. For now, I just patched `HalApi` to ignore these changes, but I'll start the work on incrementally making core dynamic as well. That will be a much more complicated process, but I do have a plan.

cc @Wumpf

lylythechosenone:
I think we can do it without that change [`A` as an associated type], and as such, we should avoid that change in an effort to reduce breaking changes.

wgpu-hal/src/lib.rs
```diff
@@ -190,50 +191,196 @@ impl InstanceError {
...
+resource!(AccelerationStructure: fmt::Debug + 'static);
 
-pub trait Instance: Sized + WasmNotSendSync {
-    type A: Api;
+pub trait SurfaceTexture<T: Texture + ?Sized>: Resource + fmt::Debug + Borrow<T> {}
+impl<T: Resource + fmt::Debug + Borrow<Tex>, Tex: Texture> SurfaceTexture<Tex> for T {}
 
-    unsafe fn init(desc: &InstanceDescriptor) -> Result<Self, InstanceError>;
+pub trait Api: Clone + fmt::Debug + Sized {
+    type Instance: Instance<Self> + ?Sized;
```
Wumpf, line 236: style nit: I know that's not status quo either, but could we do alphabetical order within these blocks? quite a mess otherwise. Also, (more subjective still) I find the blocking up beyond surface/adapter/device and queue/commandencoder/commandbuffer quite arbitrary, wouldn't know where to draw consistent lines. Goes for a lot of these listings and is obviously not very important :)

wgpu-hal/src/lib.rs
```diff
@@ -190,50 +191,196 @@ impl InstanceError {
...
+    type BindGroup: BindGroup + ?Sized;
+    type PipelineLayout: PipelineLayout + ?Sized;
+    type ShaderModule: ShaderModule + ?Sized;
+    type RenderPipeline: RenderPipeline + ?Sized;
+    type ComputePipeline: ComputePipeline + ?Sized;
+
+    type AccelerationStructure: AccelerationStructure + ?Sized;
+}
+
+pub trait ConcreteApi:
```
Wumpf, line 276: not super happy with `Concrete` prefix namingwise. Struggling to come up with a better name though myself! The other way round is more easy though and I wonder why you ended up not going with `DynApi` + `Api`? I guess a good answer to that is that `dyn DynTexture` doesn't look great either :thinking:

wgpu-hal/src/lib.rs
```diff
@@ -190,50 +191,196 @@ impl InstanceError {
...
+
+resource!(AccelerationStructure: fmt::Debug + 'static);
 
-pub trait Instance: Sized + WasmNotSendSync {
-    type A: Api;
+pub trait SurfaceTexture<T: Texture + ?Sized>: Resource + fmt::Debug + Borrow<T> {}
+impl<T: Resource + fmt::Debug + Borrow<Tex>, Tex: Texture> SurfaceTexture<Tex> for T {}
 
-    unsafe fn init(desc: &InstanceDescriptor) -> Result<Self, InstanceError>;
+pub trait Api: Clone + fmt::Debug + Sized {
```
Wumpf, line 235: there should be some docs on how `Api` is different from `ConcreteApi` and what's the purpose of the two

Wumpf (approved):
Looking great imho, don't have any much to comment on except discussy-style stuff & ask for some docs
While I can roughly picture it with the comments on the ticket, I haven't played around yet with how we can use this in wgpu-core yet. Since you already started on that, can you share a link to a preview? :)

It's pretty nice that this solution doesn't require any duplication of the traits and is a very incremental step towards our goal!

Passing iterators via &mut seems annoying, but unless we take `Vec`s or slices more often this is unavoidable. That said, I think that would make a lot more sense for many of those, we simply have to check on wgpu-core's usage here. But let's do that in follow-up PRs if at all :)

heads-up: I believe at least @jimblandy also wanted to have look before we merge; can't hurt to have more eyes on this since this is a cornerstone for more invasive refactors to come

wgpu-hal/src/lib.rs
lylythechosenone: I'm pretty sure I just copied what hal already had, but I can do this instead.

wgpu-hal/src/lib.rs
lylythechosenone: I wanted to keep `Api` as is because it prevents a breaking change. I considered `SizedApi`, but I thought `ConcreteApi` sounded nicer.

Do note that the associated types on `ConcreteApi` are pretty much just an implementation detail. Due to the `=` bound, the regular `Api` associated types can be used in the same places.

lylythechosenone:
Currently my work on core is very much theoretical. The basic idea is that we slowly incorporate non-concrete Apis into core resources, using nical's DST ideas. It will require some framework shifts that I think might be a bit too verbose to talk about here, but I have a (very long) comment drafted that I'll probably add to the issue sometime soon. Let me know if you'd like me to add a bit more here.

jimblandy:
[edited: add `draw_on_texture` method]

I'm not sure why we need to remove the associated type from the hal traits. There's no problem making trait objects for such traits, as long as you choose a concrete type for each associated type:

```rust
use std::fmt::Display;
use std::any::Any;

trait Device {
    type Texture: Texture;
    fn create_texture(&self) -> Self::Texture;
    fn draw_on_texture(&self, texture: &Self::Texture);
}

trait Texture: Display + Any {
    fn as_any(&self) -> &dyn Any;
}

struct StaticThing;

impl Device for StaticThing {
    type Texture = i32;
    fn create_texture(&self) -> i32 {
        4
    }
    fn draw_on_texture(&self, texture: &i32) {
        println!("drawing on {texture}");
    }
}

impl Texture for i32 { 
    fn as_any(&self) -> &dyn Any { self }
}

struct DynamicThing<T>(T);

impl<T: Device> Device for DynamicThing<T> {
    type Texture = Box<dyn Texture + 'static>;
    fn create_texture(&self) -> Self::Texture {
        Box::new(self.0.create_texture())
    }
    fn draw_on_texture(&self, texture: &Self::Texture) {
        self.0.draw_on_texture((&**texture).as_any().downcast_ref::<T::Texture>().unwrap());
    }
}

impl Texture for Box<dyn Texture + 'static> {
    fn as_any(&self) -> &dyn Any { self }
}

type DynamicDevice = dyn Device<Texture = Box<dyn Texture + 'static>>;

fn f(d: &DynamicDevice) {
    let t = d.create_texture();
    println!("Displaying texture {t}");
    d.draw_on_texture(&t);
} 

fn main() {
    let d = DynamicThing(StaticThing);
    f(&d);
}
```

I think `f` is the kind of code that we want to be writing in wgpu-core.

The downcasting is inherently necessary if we want dynamic dispatch, because you must check that the `Texture` was created from this specific type of `Device`, not some unrelated implementation of `Device`.

The `as_any` kludge is annoying, but apparently necessary because even when `Trait` extends `Any`, you can't use `Any::downcast_ref` on `dyn Trait`. I borrowed the kludge from [downcast-rs](https://github.com/marcianx/downcast-rs), which provides a macro for this. It's all safe code, you just need the boilerplate `as_any` implementations in all your `Trait` implementations.

lylythechosenone:
> I'm not sure why we need to remove the associated type from the hal traits.

Creating the associated type was my own previous change. I'm reverting it because it's pointless.

EDIT: wait, you mean something else.

lylythechosenone:
@jimblandy the implementation in core is going to be a bit more complicated than this, because we want to keep resources and their metadata inline. This means turning e.g. `Texture<A>` into a potentially-dynamic sized type, based on A::Texture. You are correct about the downcasting system (look at dynamic.rs from earlier).

As for putting associated types directly on the traits, I don't really see the point. We created `Api` as a way to group the associated types, to simplify the code.

lylythechosenone:
P.S.: the reason that as_any is needed is that downcast_* are methods on dyn Any, not Any itself (that's how it stays object safe). With the new trait upcasting feature, we can cast &dyn Texture → &dyn Any, but that feature is only going to be stabilized in the latest version, so it'll take a while for our MSRV to reach it.

jimblandy:
> As for putting associated types directly on the traits, I don't really see the point. We created Api as a way to group the associated types, to simplify the code.

I think we should leave the associated types on `Instance`, `Surface`, etc. as they are on trunk now. A given backend type like `wgpu_hal::vulkan::Instance` only implements `wgpu_hal::Instance` for one `Api`, not for potentially many different `Api`s, and an associated type is what expresses that best.

lylythechosenone:
Oh, you did mean like that.

I don't like that idea, only because that's a breaking change. I don't see the point of breaking working code for a semantic difference like that one. Also, it's not entirely true - all `Instance`s should implement `Instance` for a dynamic `Api`, once that exists in core.

wgpu-hal/src/lib.rs
lylythechosenone: Done.

wgpu-hal/src/lib.rs
lylythechosenone: I think I'll just keep it as it was. Resolving.

wgpu-hal/src/lib.rs
lylythechosenone: @Wumpf what's your position on this? Are you alright with the current naming, or do you have another proposal?

Wumpf: let's keep the current naming

Wumpf:
thanks for adding those docs! let's still try to align on the issue of associated types. Don't have a ~strong~ proper opinion on this, but also feel like Jim's stake here is much higher than mine, so leaving that to the two of you

lylythechosenone:
@jimblandy do you have any, position on this, or are you good to merge?

(This meaning the spacing of associated types within Api)

lylythechosenone:
And as for the status of everything else, this PR shouldn't be too huge of a change. The only breaking change is that Api can have dynamic associated types, and ConcreteApi should probably be used by nearly everyone instead. (It has the same semantics as old Api)

cwfitzgerald:
Closing, as de-generification should have replaced this, I believe

PR #5447, opened 2024-03-27
Add recursion limit for curly braces in WGSL

### LINKED ISSUE #4352: [wgsl-in] Enforce WGSL statement nesting depth limit
A program like the following
```wgsl
fn f() {{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{
```
will cause naga to overflow the stack as it recurses in
https://github.com/gfx-rs/naga/blob/db845347f23f3a00add894d38cf62d28f47669ed/src/front/wgsl/mod.rs#L2987-L3012

There should probably be some arbitrary but fixed limit to nesting which is checked with a counter. Programs exceeding that should cause naga to fail safely with an appropriate error message.

Edit: Since this fails at a depth of ~230 for me, I'd suggest going for something like 64 (which is already pretty ridiculous anyways, but I think nobody should be complaining with 64, at least not about it being too little).

ErichDonGubler:
**Connections**

Intended to resolve https://github.com/gfx-rs/wgpu/issues/4352.

**Description**

**Testing**

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/naga/src/front/wgsl/error.rs
+++ b/naga/src/front/wgsl/error.rs
@@ -269,6 +269,10 @@ pub enum Error<'a> {
         scalar: String,
         inner: ConstantEvaluatorError,
     },
+    ExceededLimitForNestedBraces {
+        span: Span,
+        limit: u8,
+    },
 }
 
 impl<'a> Error<'a> {
@@ -769,6 +773,13 @@ impl<'a> Error<'a> {
                 notes: vec![
                     format!("the expression should have been converted to have {} scalar type", scalar),
                 ]
+            },
+            Error::ExceededLimitForNestedBraces { span, limit } => ParseError {
+                message: "brace nesting limit reached".into(),
+                labels: vec![(span, "limit reached at this brace".into())],
+                notes: vec![
+                    format!("nesting limit is currently set to {limit}"),
+                ],
             },
         }
     }
--- a/naga/src/front/wgsl/mod.rs
+++ b/naga/src/front/wgsl/mod.rs
@@ -44,6 +44,17 @@ impl Frontend {
     }
 }
 
+/// <div class="warning">
+// NOTE: Keep this in sync with `wgpu::Device::create_shader_module`!
+// NOTE: Keep this in sync with `wgpu_core::Global::device_create_shader_module`!
+///
+/// This function may consume a lot of stack space. Compiler-enforced limits for parsing recursion
+/// exist; if shader compilation runs into them, it will return an error gracefully. However, on
+/// some build profiles and platforms, the default stack size for a thread may be exceeded before
+/// this limit is reached during parsing. Callers should ensure that there is enough stack space
+/// for this, particularly if calls to this method are exposed to user input.
+///
+/// </div>
 pub fn parse_str(source: &str) -> Result<crate::Module, ParseError> {
     Frontend::new().parse(source)
 }
--- a/naga/src/front/wgsl/parse/mod.rs
+++ b/naga/src/front/wgsl/parse/mod.rs
@@ -1619,6 +1619,7 @@ impl Parser {
         lexer: &mut Lexer<'a>,
         ctx: &mut ExpressionContext<'a, '_, '_>,
         block: &mut ast::Block<'a>,
+        brace_nesting_level: u8,
     ) -> Result<(), Error<'a>> {
         self.push_rule_span(Rule::Statement, lexer);
         match lexer.peek() {
@@ -1627,7 +1628,7 @@ impl Parser {
                 self.pop_rule_span(lexer);
             }
             (Token::Paren('{'), _) => {
-                let (inner, span) = self.block(lexer, ctx)?;
+                let (inner, span) = self.block(lexer, ctx, brace_nesting_level)?;
                 block.stmts.push(ast::Statement {
                     kind: ast::StatementKind::Block(inner),
                     span,
@@ -1709,7 +1710,7 @@ impl Parser {
                         let _ = lexer.next();
                         let condition = self.general_expression(lexer, ctx)?;
 
-                        let accept = self.block(lexer, ctx)?.0;
+                        let accept = self.block(lexer, ctx, brace_nesting_level)?.0;
 
                         let mut elsif_stack = Vec::new();
                         let mut elseif_span_start = lexer.start_byte_offset();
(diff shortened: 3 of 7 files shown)


naga/src/front/wgsl/error.rs
```diff
@@ -770,6 +774,13 @@ impl<'a> Error<'a> {
                     format!("the expression should have been converted to have {} scalar type", scalar),
                 ]
             },
+            Error::ExceededLimitForNestedBraces { span, limit } => ParseError {
+                message: format!("bruh dis too many braces"),
+                labels: vec![(span, "dis da brace yo".into())],
```
jimblandy: SHIP IT

jimblandy (commented):
sweet and simple

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -1619,22 +1619,41 @@ impl Parser {
...
+                // > § 2.4. Limits
+                // >
+                // > …
+                // >
+                // > Maximum nesting depth of brace-enclosed statements in a function[:] 127
+                //
+                // _However_, we choose 64 instead because (a) it avoids stack overflows in CI and
+                // (b) we expect the limit to be decreased to 63 based on this conversation in
+                // WebGPU CTS upstream:
+                // <https://github.com/gpuweb/cts/pull/3389#discussion_r1543742701>
```
ErichDonGubler: FTR: https://github.com/gpuweb/cts/pull/3389#discussion_r1543742701

ErichDonGubler: Motivation is @jimblandy's comment at <https://github.com/gfx-rs/wgpu/issues/4352#issuecomment-2025545672>.

ErichDonGubler:
This is now ready for review, @jimblandy!

naga/src/front/wgsl/mod.rs
```diff
@@ -44,6 +44,18 @@ impl Frontend {
...
+/// <div class="warning">
+// NOTE: Keep this in sync with `wgpu::Device::create_shader_module`!
+// NOTE: Keep this in sync with `wgpu_core::Global::device_create_shader_module`!
+///
+/// Note that this function may overflow its stack if the level of recursion in shader parsing
+/// exceeds remaining stack space, resulting in an abortion of the process invoking this function.
+///
+/// Compiler-enforced limits for parsing recursion exist, and if shader compilation runs into
+/// them, it will return an error gracefully. However, you may overflow the stack first if you
+/// do not allocate enough stack space for the thread you invoke this function in!
```
ErichDonGubler: todo: Note what users can do. I think just stating, "Make sure you don't overflow the stack with your current shader workload. If you allow users to provide shaders, use enough memory that you can demonstrate running into compiler-enforced limits."

ErichDonGubler: Face-to-face feedback from @jimblandy: Please limit yourself to a smaller word budget here; we want to stay within reasonable cognitive resource spend here.

ErichDonGubler: Reworded into 6c7ea07ba.

naga/src/front/wgsl/mod.rs
```diff
@@ -44,6 +44,17 @@ impl Frontend {
...
 
+/// <div class="warning">
+// NOTE: Keep this in sync with `wgpu::Device::create_shader_module`!
+// NOTE: Keep this in sync with `wgpu_core::Global::device_create_shader_module`!
+///
+/// Compiler-enforced limits for WGSL parsing recursion exist; if shader compilation runs into
+/// them, it will return an error gracefully. However, on some build profiles and platforms, the
+/// default stack size for a thread may be exceeded before this limit is reached during parsing.
+/// Callers should ensure that there is enough stack space for this, particularly if calls to this
+/// method are exposed to user input.
```
jimblandy: This comment buries the lede. It should say:

> Note that this function may consume quite a lot of stack space. Compiler-enforced limits ...

jimblandy (changes requested):
This looks to me as if it wouldn't catch code like:

```
loop { loop { loop { loop { loop { loop { ...
```

is that correct?

naga/src/front/wgsl/mod.rs
ErichDonGubler: Oh, I thought the `warning` box was sufficient encouragement to read the whole thing, but I have no objection to actually stating the key takeaway right off. 🙂 Will change as suggested shortly.

ErichDonGubler: Fixed up with 8cdd967ca.

ErichDonGubler:
> This looks to me as if it wouldn't catch code like:
> 
> ```
> loop { loop { loop { loop { loop { loop { ...
> ```
> 
> is that correct?

Oh, heck, you're right, and that's a pretty fatal lack of validation. I feel silly for not thinking of this. I should do the brace checking upon consuming an opening brace, rather than waiting for a statement parse which might not happen (`loop` handles the case where whitespace and a closing brace immediately follow, it looks like).

I've replaced all sites where `brace_nesting_level + 1` was written to call to a new `increase_brace_nesting` helper which does the check before returning an increased nesting level (e152ca928). I've also added a new test case called `too_many_unclosed_loops` that didn't trigger the recursion limit prior to your suggestion and my change (b855b7059).

ErichDonGubler:
Ah, <https://github.com/gfx-rs/wgpu/commit/e152ca928a43a143cfa3781495f82ef662b9faaf> introduced a bug where the nesting limit was being increased in statement parsing where the _check on the incremented nesting level_ (but not a permanent increment) was happening. Fortunately, CI caught this for me! Fixed with c19fa79ce.

ErichDonGubler:
Squashed things down and enabled auto-merge, since I think @jimblandy's feedback has been satisfied.

jimblandy (approved):
Looks good to me!

Issue #4067, opened 2023-08-16
Proposal for Underlying Api Interoperability


cwfitzgerald:
It is decently common for people to want to integrate wgpu into a larger application that is using a graphics api or to use a library built around a graphics api to integrate with wgpu. There are many different ways and with a wide variety of resources that you could need to do interop with, so this proposal will be a set of smaller api changes that combine to have a unified picture to underlying apis.

**This API is sure to evolve as it is refined, this is just a first attempt to unify it**

# wgpu layer

We currently only support getting the underlying types for a wgpu type if we're running on wgpu-core/hal. We should also allow this for access to the underlying WebGPU resource. To facilitate this, we should have a trait that mirrors wgpu_hal::Api but only for associated types. This is only there for allowing generic functions, so no real trait bounds need to exist.

```rust
trait Api {
   type Instance;
   type Adapter;
   type Device;
   type Texture;
   type Buffer;
   ...
}

struct WebGPU;
pub use wgpu_hal::api::*;

impl Api for WebGPU {
    type Instance = ();
    type Adapter = GpuAdapter;
    ....
}

impl Api for A where A: wgpu_hal::Api {
    // These are not wgpu_hal::*::Instance. It's the actual underlying api instance type
    type Instance = A::RawInstance;
    type Adapter = A::RawDevice;
    ....
}
```

Second we change the as_hal interop functions to be as_inner_api and take `A: Api`

```rust
impl CommandEncoder {
    fn as_inner_api<A: Api>(&self, f: impl FnOnce(&mut A::CommandBuffer) -> R) -> R;
}
```

## Ownership

The ownership of wgpu-created resources always lies in wgpu, all client code must keep the wgpu object alive while they are using the object.

The ownership of all wgpu-imported resources will depend on the presence of a DropGuard. This drop guard is a `Box<dyn FnOnce(A::Raw*)>` which will be called when the resource is destroyed.

- If a drop guard is provided, wgpu does not own the resource, and it must be kept alive externally until the provided drop guard is call.
- If a drop guard is not provided, wgpu does own the resource, and it will be destroyed as if created by wgpu.

## Creation

For each importable object, we will add a associated type that gives all the information wgpu-hal needs to successfully import a type.

```rust
trait Api {
    type InstanceImportDescriptor;
    type AdapterImportDescriptor;
    type DeviceImportDescriptor;
    type TextureImportDescriptor;
    type BufferImportDescriptor;
    // Note _not_ command encoder.
[...]

PR #5452, opened 2024-03-29
Add more hal methods

JMS55:
**Connections**
Revival of https://github.com/gfx-rs/wgpu/pull/3966

**Description**
* Adds CommandEncoder::as_hal_mut()
* Adds TextureView::as_hal()
* Adds a return value to Texture::has_hal()

Enable some interop with native graphics APIs. I'm not convinced that this is in any way a good idea. Users have to be extremely careful about not accidentally taking locks on or destroying resources while using them from native dispatches. They also have to consider pipeline barriers and image transitions themselves, with no help from wgpu in tracking them. To work around this in Bevy, I plan to bind the resources to a dummy compute pass that does nothing, but that's run before+after my native graphics API to try and coerce wgpu into tracking the lifetimes and barriers appropriately. This is very much a band-aid patch while I wait for a proper fix in https://github.com/gfx-rs/wgpu/issues/4067.

**Testing**
Will be tested as part of Bevy's FSR integration https://github.com/bevyengine/bevy/pull/12768.

**Checklist**

- [ x ] Run `cargo fmt`.
- [ x ] Run `cargo clippy`. If applicable, add:
  - [ x ] `--target wasm32-unknown-unknown`
  - [ x ] `--target wasm32-unknown-emscripten`
- [ x ] Run `cargo xtask test` to run tests.
- [ x ] Add change to `CHANGELOG.md`. See simple instructions inside file.

### DIFF
--- a/wgpu-core/src/resource.rs
+++ b/wgpu-core/src/resource.rs
@@ -924,11 +924,11 @@ impl Global {
     /// # Safety
     ///
     /// - The raw texture handle must not be manually destroyed
-    pub unsafe fn texture_as_hal<A: HalApi, F: FnOnce(Option<&A::Texture>)>(
+    pub unsafe fn texture_as_hal<A: HalApi, F: FnOnce(Option<&A::Texture>) -> R, R>(
         &self,
         id: TextureId,
         hal_texture_callback: F,
-    ) {
+    ) -> R {
         profiling::scope!("Texture::as_hal");
 
         let hub = A::hub(self);
@@ -937,7 +937,7 @@ impl Global {
         let snatch_guard = texture.device.snatchable_lock.read();
         let hal_texture = texture.raw(&snatch_guard);
 
-        hal_texture_callback(hal_texture);
+        hal_texture_callback(hal_texture)
     }
 
     /// # Safety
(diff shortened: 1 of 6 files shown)


wgpu-core/src/resource.rs
```diff
@@ -1005,6 +1027,29 @@ impl Global {
 
         hal_surface_callback(hal_surface)
     }
+
+    /// # Safety
+    ///
+    /// - The raw command encoder handle must not be manually destroyed
+    pub unsafe fn command_encoder_as_hal_mut<
```
Wumpf, line 1034: the side effect of this requires it internally to start wgpu-core level recording if it hasn't started already should be documented both here and on the wgpu level

Wumpf (changes requested):
>  I'm not convinced that this is in any way a good idea. 

neither 😄, as you point out every single one of these apis is a massive footgun and requires intricate knowledge of wgpu almost down to the specific version. But these fit into what's already there and your usecase illustrates the need for workarounds like this.

Just a note on missing docs - I find it particularly scary that this bottoms to `self.raw.begin_encoding` and would like to have this visible from the top without going down into the code.

wgpu-core/src/resource.rs
JMS55: Added.

Wumpf (approved):

PR #5460, opened 2024-03-30
[hal/vulkan] Don't advertise features without prerequisites present.

jimblandy:
Don't advertise features like `STORAGE_RESOURCE_BINDING_ARRAY` unless at least one of the features it extends, like `BUFFER_BINDING_ARRAY` or `TEXTURE_BINDING_ARRAY`, is actually present.

See the comments on `PhysicalDevicefeatures::prequisites_satisfied` for more details.

### DIFF
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -560,21 +560,13 @@ impl PhysicalDeviceFeatures {
             self.core.shader_sampled_image_array_dynamic_indexing != 0,
         );
         features.set(F::SHADER_PRIMITIVE_INDEX, self.core.geometry_shader != 0);
-        if Self::all_features_supported(
-            &features,
-            &[
-                (
-                    F::BUFFER_BINDING_ARRAY,
-                    self.core.shader_storage_buffer_array_dynamic_indexing,
-                ),
-                (
-                    F::TEXTURE_BINDING_ARRAY,
-                    self.core.shader_storage_image_array_dynamic_indexing,
-                ),
-            ],
-        ) {
-            features.insert(F::STORAGE_RESOURCE_BINDING_ARRAY);
-        }
+        features.set(
+            F::STORAGE_RESOURCE_BINDING_ARRAY,
+            (features.contains(F::BUFFER_BINDING_ARRAY)
+                && self.core.shader_storage_buffer_array_dynamic_indexing != 0)
+                || (features.contains(F::TEXTURE_BINDING_ARRAY)
+                    && self.core.shader_storage_image_array_dynamic_indexing != 0),
+        );
         //if self.core.shader_storage_image_array_dynamic_indexing != 0 {
         //if self.core.shader_clip_distance != 0 {
         //if self.core.shader_cull_distance != 0 {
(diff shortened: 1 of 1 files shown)


wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -689,13 +689,39 @@ impl PhysicalDeviceFeatures {
...
+        // disjunction (at least one of them must have its antecedent
+        // satisfied), so the `Iterator` adapters don't clarify very much.
+        let mut any = false;
+        for &(wgpu_base_feature, vulkan_support) in implications.iter() {
+            if features.contains(wgpu_base_feature) && vulkan_support == 0 {
+                return false;
+            }
+            any = true;
+        }
+        any
```
teoxoy: This will return `true` if the Features in the `implications` are not present in `features`.

teoxoy (changes requested):
I feel like it would be easier to just inline the requirements.

```rust
features.set(F::STORAGE_RESOURCE_BINDING_ARRAY,
    (features.contains(F::BUFFER_BINDING_ARRAY) && self.core.shader_storage_buffer_array_dynamic_indexing != 0) ||
    (features.contains(F::TEXTURE_BINDING_ARRAY) && self.core.shader_storage_image_array_dynamic_indexing != 0)
);
features.set(F::SAMPLED_TEXTURE_AND_STORAGE_BUFFER_ARRAY_NON_UNIFORM_INDEXING,
    (features.contains(F::TEXTURE_BINDING_ARRAY) && descriptor_indexing.shader_sampled_image_array_non_uniform_indexing != 0) &&
    (features.contains(F::BUFFER_BINDING_ARRAY | F::STORAGE_RESOURCE_BINDING_ARRAY) && descriptor_indexing.shader_storage_buffer_array_non_uniform_indexing != 0)
);
features.set(F::UNIFORM_BUFFER_AND_STORAGE_TEXTURE_ARRAY_NON_UNIFORM_INDEXING,
    (features.contains(F::BUFFER_BINDING_ARRAY) && descriptor_indexing.shader_uniform_buffer_array_non_uniform_indexing != 0) &&
    (features.contains(F::TEXTURE_BINDING_ARRAY | F::STORAGE_RESOURCE_BINDING_ARRAY) && descriptor_indexing.shader_storage_image_array_non_uniform_indexing != 0)
);
```

jimblandy:
> I feel like it would be easier to just inline the requirements.

And remove the helper function altogether?  ... I strongly agree.

teoxoy:
Yeah, I think it's more straightforward. Especially here because `STORAGE_RESOURCE_BINDING_ARRAY` needs different logic compared to the other 2 (`OR` vs `AND`).

alphastrata:
any movement on this one?

wgpu-hal/src/vulkan/adapter.rs
jimblandy: That's true. But that's also true of the original, right?

Anyway, yet another reason to just rewrite all this more directly.

teoxoy (approved):
Nice!

teoxoy:
@jimblandy looks like CI is failing.

PR #5465, opened 2024-03-31
[wgpu] Document buffer mapping.

jimblandy:
Resurrecting an ancient branch.

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -196,12 +196,31 @@ pub struct SubmissionIndex(ObjectId, Arc<crate::Data>);
 #[cfg(send_sync)]
 static_assertions::assert_impl_all!(SubmissionIndex: Send, Sync);
 
-/// The main purpose of this struct is to resolve mapped ranges (convert sizes
-/// to end points), and to ensure that the sub-ranges don't intersect.
+/// The mapped portion of a buffer, if any, and its outstanding views.
+///
+/// This ensures that views fall within the mapped range and don't overlap, and
+/// also takes care of turning `Option<BufferSize>` sizes into actual buffer
+/// offsets.
 #[derive(Debug)]
 struct MapContext {
+    /// The overall size of the buffer.
+    ///
+    /// This is just a convenient copy of [`Buffer::size`].
     total_size: BufferAddress,
+
+    /// The range of the buffer that is mapped.
+    ///
+    /// This is `0..0` if the buffer is not mapped. This becomes non-empty when
+    /// the buffer is mapped at creation time, and when you call `map_async` on
+    /// some [`BufferSlice`] (so technically, it indicates the portion that is
+    /// *or has been requested to be* mapped.)
+    ///
+    /// All [`BufferView`]s and [`BufferViewMut`]s must fall within this range.
     initial_range: Range<BufferAddress>,
+
+    /// The ranges covered by all outstanding [`BufferView`]s and
+    /// [`BufferViewMut`]s. These are non-overlapping, and are all contained
+    /// within `initial_range`.
     sub_ranges: Vec<Range<BufferAddress>>,
 }
 
@@ -214,6 +233,7 @@ impl MapContext {
         }
     }
 
+    /// Record that the buffer is no longer mapped.
     fn reset(&mut self) {
         self.initial_range = 0..0;
 
@@ -223,12 +243,22 @@ impl MapContext {
         );
     }
 
+    /// Record that the `size` bytes of the buffer at `offset` are now viewed.
+    ///
+    /// Return the byte offset within the buffer of the end of the viewed range.
+    ///
+    /// # Panics
+    ///
+    /// This panics if the given range overlaps with any existing range.
     fn add(&mut self, offset: BufferAddress, size: Option<BufferSize>) -> BufferAddress {
         let end = match size {
             Some(s) => offset + s.get(),
             None => self.initial_range.end,
         };
         assert!(self.initial_range.start <= offset && end <= self.initial_range.end);
+        // This check is essential for avoiding undefined behavior: it is the
+        // only thing that ensures that `&mut` references to the buffer's
+        // contents don't alias anything else.
         for sub in self.sub_ranges.iter() {
             assert!(
                 end <= sub.start || offset >= sub.end,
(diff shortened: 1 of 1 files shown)


wgpu/src/lib.rs
```diff
@@ -260,6 +298,93 @@ impl MapContext {
...
 /// Corresponds to [WebGPU `GPUBuffer`](https://gpuweb.github.io/gpuweb/#buffer-interface).
+///
+/// # Mapping buffers
+///
+/// If a `Buffer` is created with the appropriate [`usage`], it can be *mapped*:
+/// you can make its contents accessible to the CPU as an ordinary `&[u8]` or
+/// `&mut [u8]` slice of bytes. Buffers created with the
+/// [`mapped_at_creation`][mac] flag set are also mapped initially.
+///
+/// Depending on the hardware, the buffer could be memory shared between CPU and
```
Wumpf, line 309: we're missing a note here on what happens on WebGPU where there's always another copy to a jsarray (ideally comes with links to the jsarraybuffer using "extension functions")

wgpu/src/lib.rs
```diff
@@ -260,6 +298,93 @@ impl MapContext {
...
+/// execution makes it hard to avoid friction altogether.
+///
+/// While a buffer is mapped, you must not submit any commands to the GPU that
+/// access it. You may record command buffers that use the buffer, but you must
+/// not submit such command buffers.
+///
+/// When you are done using the buffer on the CPU, you must call
+/// [`Buffer::unmap`] to make it available for use by the GPU again. All
+/// [`BufferView`] and [`BufferViewMut`] views referring to the buffer must be
+/// dropped before you unmap it; otherwise, [`Buffer::unmap`] will panic.
```
Wumpf, line 382: [offtopic / not actionable] I really hope we can work towards getting rid of all this panicing at some point, it's pretty awful to have your application crash left and right because of subtle mistakes like this and having no way to recover / ignore

wgpu/src/lib.rs
```diff
@@ -2917,6 +3066,18 @@ fn range_to_offset_size<S: RangeBounds<BufferAddress>>(
...
+///
+/// To get a `BufferView`, first [map] the buffer, and then
+/// call `buffer.slice(range).get_mapped_range()`.
+///
+/// `BufferView` dereferences to `&[u8]`, so you can use all the usual Rust
+/// slice methods to access the buffer's contents. It also implements
+/// `AsRef<[u8]>`, if that's more convenient.
+///
+/// A `BufferView` borrows a shared reference to the `Buffer`, but if you try to
+/// create overlapping views of a buffer, `get_mapped_range` will panic.
```
Wumpf: as long as the lifetime is around to ensure that the buffer lives it's a bit of an odd detail implementation detail to point out that this has a shared reference to the buffer

(I have an experimental branch around somewhere where I'm trying to get rid of the lifetime constraint because it prevents staging belt implementations from giving out live views, but that's nowhere near of shippable)

wgpu/src/lib.rs
```diff
@@ -2925,8 +3086,20 @@ pub struct BufferView<'a> {
...
+/// call `buffer.slice(range).get_mapped_range_mut()`.
+///
+/// `BufferViewMut` dereferences to `&mut [u8]`, so you can use all the usual
+/// Rust slice methods to access the buffer's contents. It also implements
+/// `AsMut<[u8]>`, if that's more convenient.
+///
 /// It is possible to read the buffer using this view, but doing so is not
 /// recommended, as it is likely to be slow.
+///
+/// A `BufferViewMut` borrows a shared reference to the `Buffer`, but if you try
```

wgpu/src/lib.rs
```diff
@@ -4628,6 +4801,9 @@ impl<'a> Drop for QueueWriteBufferView<'a> {
 impl Queue {
     /// Schedule a data write into `buffer` starting at `offset`.
     ///
+    /// This uses an internally managed staging buffer to make a copy of `data`
+    /// immediately, and arranges for it to be uploaded later.
```
Wumpf: since this is wgpu interface we _technically_ don't know how it's implemented as this may go to the WebGPU interface and therefore to an arbitrary WebGPU implementation

wgpu/src/lib.rs
```diff
@@ -2925,8 +3086,20 @@ pub struct BufferView<'a> {
 
 /// Write only view into mapped buffer.
 ///
+/// To get a `BufferViewMut`, first [map] the buffer, and then
+/// call `buffer.slice(range).get_mapped_range_mut()`.
+///
+/// `BufferViewMut` dereferences to `&mut [u8]`, so you can use all the usual
+/// Rust slice methods to access the buffer's contents. It also implements
+/// `AsMut<[u8]>`, if that's more convenient.
```
Wumpf, line 3113: the really important thing to note here is that it's not allowed to read from the buffer since for wgpu-core this may be arbitrarily slow (the memory might be write-combined) and the contents of the buffer are undefined

Wumpf (changes requested):
nice! a few nits

wgpu/src/lib.rs
jimblandy: It would be nice if we could make those static errors somehow.

wgpu/src/lib.rs
jimblandy: I don't actually know anything about this. Firefox doesn't use `wgpu` proper, only `wgpu-core`, and our mapping is cross-process, so there's *tons* of copies.

Since I don't understand the issues you're describing, I'd like to leave this point for someone else to explain.

wgpu/src/lib.rs
jimblandy: I think the point I was intending to convey with that detail is actually not true anyway.

wgpu/src/lib.rs
jimblandy: Oh, wait, now I know what you mean.

jimblandy: I don't think this content fits in here, so I added a separate section later.

wgpu/src/lib.rs
jimblandy: Okay - I've taken this out. I wanted to try to give the user some intuition about how this performs relative to mapping the buffer and writing to it. But I guess we really don't know enough about the implementation to offer that intuition.

wgpu/src/lib.rs
jimblandy: The next paragraph says,

> It is possible to read the buffer using this view, but doing so is not recommended, as it is likely to be slow.

Is that not what you're looking for?

Wumpf: ah, sorry, tunnelvision on changed lines

Right now we panic though on deref, right? That would the "it is possible" part wrong. (as per Matrix chat, we don't really want that, warning logged in debug mode at most)

Wumpf (approved):

wgpu/src/lib.rs
jimblandy: No, we only warn:

```rust
impl Deref for BufferViewMut<'_> {
    type Target = [u8];

    fn deref(&self) -> &Self::Target {
        if !self.readable {
            log::warn!("Reading from a BufferViewMut is slow and not recommended.");
        }

        self.data.slice()
    }
}
```

PR #5469, opened 2024-04-01
chore: update deno

crowlKats:
Updated #2170

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1077,7 +1077,7 @@ dependencies = [
 
 [[package]]
 name = "deno_webgpu"
-version = "0.85.0"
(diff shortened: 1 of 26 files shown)


littledivy (approved):
LGTM

crowlKats:
Blocked on an issue where running files using any of the webgpu related APIs will exit

crowlKats:
its fixed now

ErichDonGubler (approved):
Approving here, qualifying that I only took time to really review dependencies and surface-level checks. I think this is appropriate for how much we actually depend on `cts_runner` right now, (which, AIUI, is not at all yet).

PR #5475, opened 2024-04-02
fix: unlock guard for `release_gpu_resources` call in `Device::maintain`

### LINKED ISSUE #5471: Deadlock destroying resources of a destroyed device
If you call `wgpu_core::global::Global::device_destroy` and then `Global::device_poll` on a device that has a buffer and an empty submission queue, wgpu will deadlock. This is a regression from #5426.

- `Global::device_poll` calls `poll_single_device`, which acquires a read guard on the device's snatchlock, and passes it to `Device::maintain`.

- `Device::maintain` notices that `self` is no longer valid and that the submission queue is empty, and calls `Device::release_gpu_resources`.

- `Device::release_gpu_resources` calls `Buffer::destroy` and `Texture::destroy`.

- `Buffer::destroy` and `Texture::destroy` both try to acquire a write guard on the device's snatchlock, causing a deadlock.

ErichDonGubler:
**Connections**

Resolves <https://github.com/gfx-rs/wgpu/issues/5471>.

**Description**

Resolves <https://github.com/gfx-rs/wgpu/issues/5471> by unlocking the snatch guard that conflicts with the call to `Device::release_gpu_resources` inside of `Device::maintain`.

**Testing**

Add the new test case `wgpu_test::poll_destroyed_device` which exercises the deadlock prior to this fix. You can test like so:

```
cargo nextest run --all-features -p wgpu-test -- wgpu_test::device::poll_destroyed_device
```

**Checklist**

- [x] Run `cargo fmt`.
- [x] ~~Add change to `CHANGELOG.md`. See simple instructions inside file.~~ This fixes an unreleased fix with no `CHANGELOG` entry, so I don't think that this is necessary.

### DIFF
--- a/tests/tests/device.rs
+++ b/tests/tests/device.rs
@@ -776,3 +776,42 @@ fn vs_main() -> @builtin(position) vec4<f32> {
             ],
         });
     });
+
+#[gpu_test]
+static DEVICE_DESTROY_THEN_BUFFER_CLEANUP: GpuTestConfiguration = GpuTestConfiguration::new()
+    .parameters(TestParameters::default())
+    .run_sync(|ctx| {
+        // When a device is destroyed, its resources should be released,
+        // without causing a deadlock.
+
+        // Create a buffer to be left around until the device is destroyed.
+        let _buffer = ctx.device.create_buffer(&wgpu::BufferDescriptor {
+            label: None,
+            size: 256,
+            usage: wgpu::BufferUsages::MAP_READ,
+            mapped_at_creation: false,
+        });
+
+        // Create a texture to be left around until the device is destroyed.
+        let texture_extent = wgpu::Extent3d {
+            width: 512,
+            height: 512,
+            depth_or_array_layers: 1,
+        };
+        let _texture = ctx.device.create_texture(&wgpu::TextureDescriptor {
+            label: None,
+            size: texture_extent,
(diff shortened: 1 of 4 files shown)


tests/tests/device.rs
```diff
@@ -776,3 +776,24 @@ fn vs_main() -> @builtin(position) vec4<f32> {
...
+            mapped_at_creation: false,
+        });
+        // …but don't do anything with it.
+
+        ctx.device.destroy();
+
+        // We now have an empty submission queue, and the device will clean up `_buffer` on the
+        // next `poll` operation.
+
+        ctx.instance.poll_all(true);
```
ErichDonGubler: issue: This test doesn't exercise textures, as [@jimblandy's](https://github.com/gfx-rs/wgpu/issues/5471#issuecomment-2030503115) did. We should just use his instead.

jimblandy:
@ErichDonGubler I've pushed two commits:

- The first changes the test case. I adopted parts of yours that were clearer.
- The second takes a different approach to handling the snatchguard. I think this is simpler.

Please take a look and let me know what you think.

jimblandy (changes requested):
I added two commits that are suggested revisions. So review is sort of back in Erich's court now.

jimblandy (approved):
Erich approved my suggestions, so this is good to go.

PR #5479, opened 2024-04-03
refactor(deno): simplify error handling

crowlKats:
This refactors the error handling to be more verbatim to what the spec specifies, while simplifying the implementation.
The main difference is that its not promise based anymore, which i think is fine given that all errors previously were "fake" promises. 
Spec references: https://gpuweb.github.io/gpuweb/#abstract-opdef-dispatch-error


deno_webgpu/01_webgpu.js
```diff
@@ -1877,12 +1810,7 @@ class GPUDevice extends EventTarget {
         "OperationError",
       );
     }
-    const operations = SafePromiseAll(scope.operations);
-    return PromisePrototypeThen(
-      operations,
-      () => PromiseResolve(null),
-      (err) => PromiseResolve(err),
-    );
+    return PromiseResolve(scope.errors[0] ?? null);
```
crowlKats, line 1810: The spec states:
> Let error be any one of the items in scope.[[[errors]]](https://gpuweb.github.io/gpuweb/#dom-gpu-error-scope-errors-slot), or null if there are none.
 >For any two errors E1 and E2 in the list, if E2 was caused by E1, E2 should not be the one selected.
 
 it would be neat if we could use an AggregateError here instead, but i guess currently not possible. might be worth bringing up to webgpu spec.

cwfitzgerald: I don't really understand this, but would be worth filing an issue with more context if you think this is weird.

cwfitzgerald (approved):

Issue #4842, opened 2023-12-06
Vulkan color space is different when rendering to swap chain (sRGB)


valaphee:
**Description**
When comparing the cube and water example for different back-ends on window (screenshots even of Vulkan seems to be fine which is weird). The colors are noticeable different.

The colors on Vulkan are darker compared to the screenshots and GL, DX12.

**Repro steps**
Run the cube or water example with WGPU_BACKEND=dx12 and with WGPU_BACKEND=vulkan

**Expected vs observed behavior**
Darker colors compared to the baseline and other back-ends window output.

**Extra materials**
DX12/OpenGL (AMD RX 6900 XT, Win 10/11, No HDR), Vulkan/OpenGL (AMD RX 6900 XT, Linux, No HDR) (expected?):
![Screenshot 2023-12-14 163957](https://github.com/gfx-rs/wgpu/assets/32491319/7f8776d6-c9d2-4356-b0ae-a183972a38b9)
Vulkan (AMD RX 6900 XT, Win 11, No HDR):
![Screenshot 2023-12-14 164014](https://github.com/gfx-rs/wgpu/assets/32491319/ee5029f4-b02c-4b0b-bff3-4a8d29421294)
Vulkan (Nvidia RTX 3070, Win 10, No HDR):
![Unbenannt](https://github.com/gfx-rs/wgpu/assets/32491319/a3255b26-9b19-4582-afcc-fc0266c4b087)

[wgpu-info.json](https://github.com/gfx-rs/wgpu/files/13592890/wgpu-info.json)

**Platform**

valaphee:
Might be related to https://github.com/gfx-rs/wgpu/issues/1646 and https://github.com/gfx-rs/wgpu/issues/3449, but probably not. As the format seems correct and it also looks right in RenderDoc, but Vulkan seems to do some gamma correction on its own.

teoxoy:
@cwfitzgerald any ideas why this is happening?

cwfitzgerald:
This looks like a tale of different primaries. When outputting SDR, what the output of compositing isn't sRGB. It's your monitor's color space. 

What looks like it's happening is that DX12 is outputting sRGB color (by converting the sRGB into the monitor's color space), whereas vulkan is just yeeting the colors into the monitor and hoping for the best.

In short, I'm not sure there's much we can do about this.

valaphee:
[...]

PR #5482, opened 2024-04-03
Prefer OpenGL over OpenGL ES

### LINKED ISSUE #5481: Prefer OpenGL over OpenGL ES
While working with features which are only present on OpenGL, but not ES, I wondered why OpenGL ES was used on my desktop PC. This also seems to only be the case on Linux, as on Windows OpenGL is used.

After looking up how OpenGL profiles work, I found that `egl.bind_api(khronos_egl::OPENGL_ES_API).unwrap();` in https://github.com/gfx-rs/wgpu/blob/trunk/wgpu-hal/src/gles/egl.rs#L529 is called.

Changing `khronos_egl::OPENGL_ES_API` to `khronos_egl::OPENGL_API`. But I don't know if there are any implications with it. According to the documentation if this operation fails the rendering api will not be changed. (therefore the default `OPENGL_ES_API` will be used)

**Additional context**

valaphee:
**Connections**
resolves https://github.com/gfx-rs/wgpu/issues/5481

**Description**
While working with features which are only present on OpenGL, but not ES, I wondered why OpenGL ES was used on my desktop PC. This also seems to only be the case on Linux, as on Windows OpenGL is used.

**Testing**
Running any example with `WGPU_BACKEND=gl`; On cube for example the wireframe should be visibile if the device supports OpenGL, and not only OpenGL ES

**Checklist**

- [x] Run `cargo fmt`.


valaphee:
Ah good that it catched the sRGB problem, is basically the same as in https://github.com/gfx-rs/wgpu/issues/4842

Wumpf (changes requested):
as-is this certainly gonna be a problem for Android.

I believe it should be fine to do this conditionally for desktop platforms, but I lack the historical context here. Maybe @cwfitzgerald knows if there's more problems with binding the desktop OpenGL api.
The more obvious concern that comes to my mind is that by setting desktop opengl we more easily break webgl & android compatibility when testing locally. The prime usecase for OpenGL in wgpu is to be a fallback when other APIs are not available, so stepping this up to desktop OpenGL may actually not be desirable since any scenario where more features are required should instead use Vulkan/DX12/Metal. I'd also guess that on Windows it's only desktop because there's nothing else available in wgl.

valaphee:
Yea I thought so, on the other hand we support those features already anyway, but only on Windows, they are also gated accordingly.

But this PR should definitely be tested on Android though.

valaphee:
Yea I mean OpenGL and OpenGL ES share the same code path anyway, all features are gated accordingly, and allowing to support more modern features on Linux shouldn't be a problem as OpenGL is already used on Windows anyway, it's just for completeness and that I can test 64-bit vertex formats, as I don't have Windows.

Ah you're right, added a check if egl supports OpenGL, now it also works on Android (tested on Android 13).

valaphee:
I noticed when running `cargo xtask test` that `wgpu_test::clear_texture::clear_texture_uncompressed` now fails.

```
[2024-04-06T12:34:24Z ERROR wgpu_test::expectations] Expected to fail due to [FailureReason { kind: Some(Panic), message: Some("texture with format Rg8Snorm was not fully cleared") }, FailureReason { kind: Some(Panic), message: Some("texture with format Rgb9e5Ufloat was not fully cleared") }, FailureReason { kind: Some(ValidationError), message: Some("GL_INVALID_FRAMEBUFFER_OPERATION") }, FailureReason { kind: Some(ValidationError), message: Some("GL_INVALID_OPERATION") }], but did not fail
thread '<unnamed>' panicked at 'explicit panic', tests/src/run.rs:119:9
```
the test is expected to fail, but does no longer fail, which I guess is a good thing. The fail should only be expected when the backend is GLES.

wgpu-hal/src/gles/egl.rs
```diff
@@ -954,6 +972,10 @@ impl crate::Instance<super::Api> for Instance {
             })
         };
 
+        if !matches!(inner.srgb_kind, SrgbFrameBufferKind::None) {
+            unsafe { gl.enable(glow::FRAMEBUFFER_SRGB) };
+        }
```
Wumpf, line 1018: Can you please add a comment explaining this. Took me quite a while to figure out that this is needed because in GLES all GL_SRGB buffers do the expected conversions whereas on desktop GL one can opt out of it with this flag (wtf...)
(great catch btw.! :))

Wumpf (changes requested):
thanks for testing and fixing on Android. Thinking about this again, with this fix in place and the fact that we pass now even more tests there doesn't seem to be much reason not to enable desktop gl if possible :)

Can you change the test you mentioned from expected-to-fail to expected-to-succeed as part of this PR please?

valaphee:
I noticed some odd behavior, with Bevy using OpenGL

normally it should look like (Win+Vulkan, WebGL, Linux+Vulkan): 
Original
but with OpenGL on Linux I get
OpenGL

on Windows with OpenGL it doesn't run at all
```
Caused by:
      note: label = `pbr_opaque_mesh_pipeline`
    Internal error in ShaderStages(FRAGMENT) shader: gsamplerCubeArrayShadow isn't supported in textureGrad, textureLod or texture with bias
```

valaphee:
Ah correction, on Linux with OpenGL ES I get spammed with
```
2024-04-08T16:33:18.891346Z ERROR wgpu_hal::gles: wgpu-hal heuristics assumed that the view dimension will be equal to `Cube` rather than `CubeArray`.
`D2` textures with `depth_or_array_layers == 1` are assumed to have view dimension `D2`
`D2` textures with `depth_or_array_layers > 1` are assumed to have view dimension `D2Array`
`D2` textures with `depth_or_array_layers == 6` are assumed to have view dimension `Cube`
`D2` textures with `depth_or_array_layers > 6 && depth_or_array_layers % 6 == 0` are assumed to have view dimension `CubeArray`    
```
but seems to run as expected.

Fixed the test, added the comment for enabling sRGB conversion; GL now also fills out the AdapterInfo:
`driver: "OpenGL ES", driver_info: "3.2 Mesa 24.0.4-arch1.2",`
`driver: "OpenGL", driver_info: "4.6 (Compatibility Profile) Mesa 24.0.4-arch1.2"`
(looks the same as on Vulkan)

wgpu-hal/src/gles/egl.rs
```diff
@@ -954,6 +972,12 @@ impl crate::Instance<super::Api> for Instance {
             })
         };
 
+        // In contract with OpenGL ES, OpenGL requires explicitly enabling sRGB conversions,
```
Wumpf:
```suggestion
        // In contrast to OpenGL ES, OpenGL requires explicitly enabling sRGB conversions,
```

Wumpf (changes requested):
code lgtm to be me now, but those color differences are concerning (looks like a shader broke..?), we'll have to get to the bottom of this before landing otherwise anyone relying on gl on linux breaks.
Have you checked of the log spam on linux gl es happened before? Also, curious, how did you force gles with this patch now?

valaphee:
> code lgtm to be me now, but those color differences are concerning (looks like a shader broke..?), we'll have to get to the bottom of this before landing otherwise anyone relying on gl on linux breaks. Have you checked of the log spam on linux gl es happened before? Also, curious, how did you force gles with this patch now?

On Windows it doesn't run at all, because of missing shader features, and on Linux without this PR I get the errors in the log (wgpu-hal heuristics, also probably the shadow shader, even though it seems to work just fine)

With this pr, the error in the log is gone, but the result is just wrong. I'll try to isolate this issue.

Did it manually (by changing bind_api to OPENGL_ES), don't know if its worth exposing this, or always just prefer OpenGL.

Wumpf:
> Did it manually (by changing bind_api to OPENGL_ES), don't know if its worth exposing this, or always just prefer OpenGL.

don't think it's worth exposing either if we get this stuff fixed. Was just curious if there's some magic trick :)

valaphee:
It was an oversight, also got this issue on Windows. The problem was that draw indexed has a path for when the base_vertex is 0 and uses `draw_elements_instanced`, and first_instance gets always passed though as an uniform, regardless of if the shader is compiled with the draw_param feature or not.

The path also only seems to be relevant to ANGLE, and ANGLE is OpenGL ES anyway, therefore the feature will not be available there.

Wumpf (approved):
great stuff, thank you!

PR #5483, opened 2024-04-03
implement slice for QueueWriteBuffer

Smicry:
fix the slice panic


Wumpf:
that panic sure could use some docs, but it was put there intentionally since write buffers should be only accessed for writing. See https://github.com/gfx-rs/wgpu/blob/trunk/wgpu/src/lib.rs#L4559

Also, please add more description on PRs and don't ignore the PR template

PR #5486, opened 2024-04-03
[naga wgsl-in] Fill out error message for deep nesting.

jimblandy:
In the WGSL front end, clarify error message for deeply nested statements.

@ErichDonGubler I was going to push this as a suggestion to the PR but forgot to. Only take it if you like it.

### DIFF
--- a/naga/src/front/wgsl/error.rs
+++ b/naga/src/front/wgsl/error.rs
@@ -775,10 +775,10 @@ impl<'a> Error<'a> {
                 ]
             },
             Error::ExceededLimitForNestedBraces { span, limit } => ParseError {
-                message: "brace nesting limit reached".into(),
-                labels: vec![(span, "limit reached at this brace".into())],
+                message: "Statements nested too deeply".into(),
+                labels: vec![(span, "nesting limit reached at this brace".into())],
                 notes: vec![
-                    format!("nesting limit is currently set to {limit}"),
+                    format!("The maximum nesting depth of brace-enclosed statements in a function is {limit}."),
                 ],
             },
         }


naga/src/front/wgsl/error.rs
```diff
@@ -775,10 +775,10 @@ impl<'a> Error<'a> {
                 ]
             },
             Error::ExceededLimitForNestedBraces { span, limit } => ParseError {
-                message: "brace nesting limit reached".into(),
-                labels: vec![(span, "limit reached at this brace".into())],
+                message: "Statements nested too deeply".into(),
```
ErichDonGubler, line 778: issue: The capitalized casing, for better or worse, is _not_ consistent with other diagnostics.

```suggestion
                message: "statements nested too deeply".into(),
```

ErichDonGubler: This feedback also applies to the `notes` field.

naga/src/front/wgsl/error.rs
```diff
@@ -775,10 +775,10 @@ impl<'a> Error<'a> {
...
+                labels: vec![(span, "nesting limit reached at this brace".into())],
                 notes: vec![
-                    format!("nesting limit is currently set to {limit}"),
+                    format!("The maximum nesting depth of brace-enclosed statements in a function is {limit}."),
```
ErichDonGubler, line 781: issue: This new diagnostic takes pains to note that it is related to statements, but that's not true, AFAIK. The limit current works based on opening braces in _any_ parsing rule (i.e., incl. `loop`, as [you pointed out](https://github.com/gfx-rs/wgpu/pull/5447#pullrequestreview-1974886213) in the initial PR). Therefore, I think attributing this to statements is misleading.

ErichDonGubler (approved):
I'm currently working on a few of our Firefox timeline items right now, so I won't be prioritizing this for the next couple of days, but I definitely intend to get back to this!

I'm actually surprised that you wanted to tweak this. Could you help me understand what your goals for revising were?

ErichDonGubler (changes requested):
oderp, wrong button 😖

cwfitzgerald:
Boop from triage

jimblandy:
I don't want to change this any more.

PR #5492, opened 2024-04-04
Move the burden of evaluating override-expressions to users of naga's API

teoxoy:
This PR makes it possible to evaluate override-expressions for the GLSL backend as well.


ErichDonGubler:
Gave this a cursory look-over, to see what I could learn. I don't expect my review to supplant @jimblandy's, but it looks basically like I expected. 👍🏻

EDIT: Also, CI is failing? 🤔

jimblandy (approved):
Looks good to me!

Issue #5439, opened 2024-03-26
Naga should permit override expressions as `LocalVariable` initializers


jimblandy:
At present, Naga only allows constant expressions in `LocalVariable::init`, but there's no reason it couldn't permit override expressions as well.

https://github.com/jimblandy/wgpu/commit/7985ce1a802f186ccda95db4de18f133ae7858b5 seems to work fine.

jimblandy:
Note that this is unrelated to WGSL compliance. The WGSL front end already accepts any sort of expression as an initializer for `LocalVariable`s, but initialization by a runtime expression is handled via an explicit `Store` statement. The question here is merely which initializations we can render using `LocalVariable::init`, and which require `Store`s.

Using `init` produces slightly cleaner output.

PR #5495, opened 2024-04-04
[naga] Permit override expressions as `LocalVariable` initializers.

jimblandy:
Allow `LocalVariable::init` to be an override expression.

Note that this is unrelated to WGSL compliance. The WGSL front end already accepts any sort of expression as an initializer for `LocalVariable`s, but initialization by an override expression was handled in the same way as initialization by a runtime expression, via an explicit `Store` statement.

This commit merely lets us skip the `Store` when the initializer is an override expression, producing slightly cleaner output in some cases.

Resolves https://github.com/gfx-rs/wgpu/issues/5439.

### DIFF
--- a/naga/src/back/pipeline_constants.rs
+++ b/naga/src/back/pipeline_constants.rs
@@ -304,6 +304,13 @@ fn process_function(
 
     filter_emits_in_block(&mut function.body, &function.expressions);
 
+    // Update local expression initializers.
+    for (_, local) in function.local_variables.iter_mut() {
+        if let &mut Some(ref mut init) = &mut local.init {
+            *init = adjusted_local_expressions[init.index()];
+        }
+    }
+
     // We've changed the keys of `function.named_expression`, so we have to
     // rebuild it from scratch.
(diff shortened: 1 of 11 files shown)


naga/src/front/glsl/parser/declarations.rs
```diff
@@ -251,7 +251,12 @@ impl<'source> ParsingContext<'source> {
                     init.and_then(|expr| ctx.ctx.lift_up_const_expression(expr).ok());
                 late_initializer = None;
             } else if let Some(init) = init {
-                if ctx.is_inside_loop || !ctx.ctx.local_expression_kind_tracker.is_const(init) {
+                if ctx.is_inside_loop
+                    || !ctx
+                        .ctx
+                        .local_expression_kind_tracker
+                        .is_const_or_override(init)
+                {
```
teoxoy:
```suggestion
                if ctx.is_inside_loop || !ctx.ctx.local_expression_kind_tracker.is_const(init) {
```

I think we should keep this as it was since GLSL doesn't have overrides.

teoxoy (approved):

PR #5498, opened 2024-04-05
Add DirectX ShaderModel 6.1-6.7 detection

atlv24:
**Connections**
For #5383

**Description**
ShaderModel detection is hardcoded to just map Dxc => 6.0 and Fxc => 5.1. This means higher shader models are never detected.

**Testing**
Not tested explicitly but works in the context of https://github.com/gfx-rs/wgpu/pull/5383

**Checklist**

- [x] Run `cargo fmt`.


wgpu-hal/src/dx12/adapter.rs
```diff
@@ -181,6 +181,37 @@ impl super::Adapter {
             hr == 0 && features3.CastingFullyTypedFormatSupported != 0
         };
 
+        let shader_model = {
+            let mut sm: crate::dx12::types::D3D12_FEATURE_DATA_SHADER_MODEL =
+                unsafe { mem::zeroed() };
```
teoxoy: > before calling the function initialize the HighestShaderModel field to the highest shader model that your application understands. After the function completes successfully, the HighestShaderModel field contains the highest shader model that is both supported by the device and no higher than the shader model passed in.

I don't think this is correct given what the D3D12 docs say.

This is further complicated by:

 > ID3D12Device::CheckFeatureSupport returns E_INVALIDARG if HighestShaderModel isn't known by the current runtime. For that reason, we recommend that you call this in a loop with decreasing shader models to determine the highest supported shader model.

There is an example here: https://github.com/microsoft/DirectXTK12/wiki/Shader-Model-6

wgpu-hal/src/dx12/adapter.rs
```diff
@@ -181,6 +181,37 @@ impl super::Adapter {
...
+                    crate::dx12::types::D3D_SHADER_MODEL_6_5 => naga::back::hlsl::ShaderModel::V6_5,
+                    crate::dx12::types::D3D_SHADER_MODEL_6_6 => naga::back::hlsl::ShaderModel::V6_6,
+                    crate::dx12::types::D3D_SHADER_MODEL_6_7 => naga::back::hlsl::ShaderModel::V6_7,
+                    _ => unreachable!(),
+                }
+            } else {
+                match dxc_container {
+                    Some(_) => naga::back::hlsl::ShaderModel::V6_0,
+                    None => naga::back::hlsl::ShaderModel::V5_1,
+                }
```
teoxoy, line 216: If DXC is not present we should just use 5.1

teoxoy (changes requested):

wgpu-hal/src/dx12/adapter.rs
atlv24: Thats what I implemented, unless im mistaken

teoxoy: Sorry, I meant in the general case (not only when the fn returns an error). The runtime could support higher shader models but we still need to cap them to 5.1 if DXC is not present.

atlv24: Got it

teoxoy (approved):
Looks good!

PR #5499, opened 2024-04-05
[wgpu-core] pass resources as Arcs when adding them to the registry

### LINKED ISSUE #5493: Pass resources as Arcs when adding them to the registry
A step towards #5124 

Currently when creating a resource, the registry takes the resource by value, and produce an Arc.

It does this so that it can set the ID of the resource, however it doesn't mesh well with our plan of making the resources DST to reduce the amount of generics in wgpu-core.

See the linked issue. In the future `Arc<core::Texture<A>>` might become an alias that would look like `Arc<core::TextureSource<dyn hal::Texture>>`.

We could instead pass the resource as an Arc to the registry directly. The steps would become:
 - reserve a slot in the registry, get an ID
 - create the resource with the ID
 - put the resource in an Arc and place it in the registry with the ID

lylythechosenone:
**Connections**
Fixes #5493
Linked to #5124 

**Description**
Currently, `Registry` takes resources to insert by value. This is incompatible with dynamic-sized resources, which are the next step in #5124. This change instead takes resources already in their `Arc`s, and uses `.get_mut()` to modify them in place.

**Testing**
No testing currently, but this is largely just a refactor. I'm not sure we need testing for it.

**Checklist**

- [x] Run `cargo fmt`.


nical (approved):

lylythechosenone:
@nical wait, there are errors on wasm and Windows. Just fixed Windows and will push in a second, but I think we need to discuss the wasm one.

lylythechosenone:
FIxed now.

PR #3507, opened 2023-02-20
[wgpu-hal] Inline RayQuery Support 

daniel-keitel:
**Checklist**

- [x] Run `cargo clippy`.
- [x] Run `RUSTFLAGS=--cfg=web_sys_unstable_apis cargo clippy --target wasm32-unknown-unknown` if applicable.
- [x] Add change to CHANGELOG.md. See simple instructions inside file.
- [x] Consider `unsafe_ops_in_unsafe_fn` #3044
- [x] Consider DX12 for API compatibility
- [x] Acceleration Structure Updates

**Future Work**
- [ ] Acceleration Structure Compaction
- [ ] DX 12 Implementation
- [ ] Ray Tracing Pipeline #3607

**Connections**
#1040
#3020 original PR
gfx-rs/naga#2256

**Description**
Basic Vulkan ray query and acceleration structure support.

This pull request is the continuation of @expenses PR #3020.

The Api is designed to work with a DirectX12 Implementation.

image

**Testing**
Tested with example application (ray-traced-triangle)

PR #5504, opened 2024-04-06
[wgpu-hal] Upgrade to `ash 0.38`

MarijnS95:
https://github.com/ash-rs/ash/releases/tag/0.38.0

In this release a lot of breaking changes have been made: builder structs are dropped in favour of always having a lifetime parameter available on every raw Vulkan structure (when they contain one or more pointers).  This massively contributes to lifetime (and mutable aliasing) safety, but requires significant changes to some uses in `wgpu-hal`.

All function pointer loaders for both Vulkan core and extensions have moved into the root of `ash::`, making `ash::vk::` a more pure `sys`-like module.  Extensions have their own `ash::<prefix>::<ext>` module to clearly separate and group their items from the core.  Besides `NAME` and the usual `*Fn` wrappers, the handwritten `extensions` module is now only available via this path.  This to combat the previous inconsistency between `ash::KhrSomeExtFn::name()` vs `ash::extensions::khr::SomeExt::name()`.

The Vulkan core clearly splits functions across `device` and `instance` functions, to keep functions that can be loaded without a dispatch-table for a device via `get_device_proc_addr()` apart from instance functions. This concept has now been applied to extension functions, making it possible to load them optimized for a device too (when the shared table would previously include instance functions and require the whole thing to be loaded via `get_instance_proc_addr()`), and in the rare case of 3 hand-written extension wrappers: load instance functions via `get_device_proc_addr()` resulting in `NULL` pointers.

Finally, a few new helpers like `_as_slice()` and `_as_c_str()` are available (the former only for statically-sized struct-owned arrays with a length delimiter field) to simplify commonly written patterns.

**Connections**
https://github.com/gfx-rs/wgpu/pull/5501

**Description**
New crate release of `ash` that users will likely want to start using, especially since it exposes a many more Vulkan extensions (generated against a newer spec) and has significant lifetime-safety improvements and other usability-refactors.

**Testing**
`cargo r -p wgpu-examples` for a few of them.

**Checklist**

- [x] Run `cargo fmt`.
  This fails on the `water` test, but also on `trunk`. As seen when running `cargo r -p wgpu-examples water`, a lot of `SYNC-HAZARD-WRITE-AFTER-WRITE` validation issues are detected.

---

I realize that this is a bit of a sloppy PR with some `TODO`s, which I'd like to discuss to find the cleanest way forward :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -133,7 +133,7 @@ objc = "0.2.5"
 
 # Vulkan dependencies
 android_system_properties = "0.1.1"
-ash = "0.37.3"
+ash = "0.38.0"
 gpu-alloc = "0.6"
 gpu-descriptor = "0.3"
 
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -1,6 +1,6 @@
 use super::conv;
 
-use ash::{extensions::khr, vk};
+use ash::{amd, ext, khr, vk};
 use parking_lot::Mutex;
 
 use std::{
@@ -43,39 +43,40 @@ pub struct PhysicalDeviceFeatures {
     core: vk::PhysicalDeviceFeatures,
 
     /// Features provided by `VK_EXT_descriptor_indexing`, promoted to Vulkan 1.2.
-    pub(super) descriptor_indexing: Option<vk::PhysicalDeviceDescriptorIndexingFeaturesEXT>,
+    pub(super) descriptor_indexing:
+        Option<vk::PhysicalDeviceDescriptorIndexingFeaturesEXT<'static>>,
 
     /// Features provided by `VK_KHR_imageless_framebuffer`, promoted to Vulkan 1.2.
-    imageless_framebuffer: Option<vk::PhysicalDeviceImagelessFramebufferFeaturesKHR>,
+    imageless_framebuffer: Option<vk::PhysicalDeviceImagelessFramebufferFeaturesKHR<'static>>,
 
     /// Features provided by `VK_KHR_timeline_semaphore`, promoted to Vulkan 1.2
-    timeline_semaphore: Option<vk::PhysicalDeviceTimelineSemaphoreFeaturesKHR>,
+    timeline_semaphore: Option<vk::PhysicalDeviceTimelineSemaphoreFeaturesKHR<'static>>,
 
     /// Features provided by `VK_EXT_image_robustness`, promoted to Vulkan 1.3
-    image_robustness: Option<vk::PhysicalDeviceImageRobustnessFeaturesEXT>,
+    image_robustness: Option<vk::PhysicalDeviceImageRobustnessFeaturesEXT<'static>>,
 
     /// Features provided by `VK_EXT_robustness2`.
-    robustness2: Option<vk::PhysicalDeviceRobustness2FeaturesEXT>,
+    robustness2: Option<vk::PhysicalDeviceRobustness2FeaturesEXT<'static>>,
 
     /// Features provided by `VK_KHR_multiview`, promoted to Vulkan 1.1.
-    multiview: Option<vk::PhysicalDeviceMultiviewFeaturesKHR>,
+    multiview: Option<vk::PhysicalDeviceMultiviewFeaturesKHR<'static>>,
 
     /// Features provided by `VK_KHR_sampler_ycbcr_conversion`, promoted to Vulkan 1.1.
-    sampler_ycbcr_conversion: Option<vk::PhysicalDeviceSamplerYcbcrConversionFeatures>,
+    sampler_ycbcr_conversion: Option<vk::PhysicalDeviceSamplerYcbcrConversionFeatures<'static>>,
 
     /// Features provided by `VK_EXT_texture_compression_astc_hdr`, promoted to Vulkan 1.3.
-    astc_hdr: Option<vk::PhysicalDeviceTextureCompressionASTCHDRFeaturesEXT>,
+    astc_hdr: Option<vk::PhysicalDeviceTextureCompressionASTCHDRFeaturesEXT<'static>>,
 
     /// Features provided by `VK_KHR_shader_float16_int8` (promoted to Vulkan
     /// 1.2) and `VK_KHR_16bit_storage` (promoted to Vulkan 1.1). We use these
     /// features together, or not at all.
     shader_float16: Option<(
-        vk::PhysicalDeviceShaderFloat16Int8Features,
-        vk::PhysicalDevice16BitStorageFeatures,
+        vk::PhysicalDeviceShaderFloat16Int8Features<'static>,
+        vk::PhysicalDevice16BitStorageFeatures<'static>,
     )>,
 
     /// Features provided by `VK_KHR_acceleration_structure`.
-    acceleration_structure: Option<vk::PhysicalDeviceAccelerationStructureFeaturesKHR>,
+    acceleration_structure: Option<vk::PhysicalDeviceAccelerationStructureFeaturesKHR<'static>>,
 
     /// Features provided by `VK_KHR_buffer_device_address`, promoted to Vulkan 1.2.
     ///
@@ -91,7 +92,7 @@ pub struct PhysicalDeviceFeatures {
     ///
     /// [`Instance::expose_adapter`]: super::Instance::expose_adapter
     /// [`Features::RAY_TRACING_ACCELERATION_STRUCTURE`]: wgt::Features::RAY_TRACING_ACCELERATION_STRUCTURE
-    buffer_device_address: Option<vk::PhysicalDeviceBufferDeviceAddressFeaturesKHR>,
+    buffer_device_address: Option<vk::PhysicalDeviceBufferDeviceAddressFeaturesKHR<'static>>,
 
     /// Features provided by `VK_KHR_ray_query`,
     ///
@@ -102,27 +103,23 @@ pub struct PhysicalDeviceFeatures {
     /// However, we do populate this when creating a device if ray tracing is requested.
     ///
     /// [`Instance::expose_adapter`]: super::Instance::expose_adapter
-    ray_query: Option<vk::PhysicalDeviceRayQueryFeaturesKHR>,
+    ray_query: Option<vk::PhysicalDeviceRayQueryFeaturesKHR<'static>>,
 
     /// Features provided by `VK_KHR_zero_initialize_workgroup_memory`, promoted
     /// to Vulkan 1.3.
     zero_initialize_workgroup_memory:
-        Option<vk::PhysicalDeviceZeroInitializeWorkgroupMemoryFeatures>,
+        Option<vk::PhysicalDeviceZeroInitializeWorkgroupMemoryFeatures<'static>>,
 
     /// Features provided by `VK_EXT_subgroup_size_control`, promoted to Vulkan 1.3.
-    subgroup_size_control: Option<vk::PhysicalDeviceSubgroupSizeControlFeatures>,
+    subgroup_size_control: Option<vk::PhysicalDeviceSubgroupSizeControlFeatures<'static>>,
 }
-
-// This is safe because the structs have `p_next: *mut c_void`, which we null out/never read.
-unsafe impl Send for PhysicalDeviceFeatures {}
-unsafe impl Sync for PhysicalDeviceFeatures {}
 
 impl PhysicalDeviceFeatures {
     /// Add the members of `self` into `info.enabled_features` and its `p_next` chain.
-    pub fn add_to_device_create_builder<'a>(
+    pub fn add_to_device_create<'a>(
         &'a mut self,
-        mut info: vk::DeviceCreateInfoBuilder<'a>,
-    ) -> vk::DeviceCreateInfoBuilder<'a> {
+        mut info: vk::DeviceCreateInfo<'a>,
+    ) -> vk::DeviceCreateInfo<'a> {
         info = info.enabled_features(&self.core);
         if let Some(ref mut feature) = self.descriptor_indexing {
             info = info.push_next(feature);
@@ -170,8 +167,8 @@ impl PhysicalDeviceFeatures {
     /// Return a `PhysicalDeviceFeatures` value capturing all the Vulkan
     /// features needed for the given [`Features`], [`DownlevelFlags`], and
     /// [`PrivateCapabilities`]. You can use the returned value's
-    /// [`add_to_device_create_builder`] method to configure a
-    /// [`DeviceCreateInfoBuilder`] to build a logical device providing those
+    /// [`add_to_device_create`] method to configure a
+    /// [`vk::DeviceCreateInfo`] to build a logical device providing those
     /// features.
     ///
     /// To ensure that the returned value is able to select all the Vulkan
@@ -188,8 +185,7 @@ impl PhysicalDeviceFeatures {
     /// [`Features`]: wgt::Features
     /// [`DownlevelFlags`]: wgt::DownlevelFlags
     /// [`PrivateCapabilities`]: super::PrivateCapabilities
-    /// [`add_to_device_create_builder`]: PhysicalDeviceFeatures::add_to_device_create_builder
-    /// [`DeviceCreateInfoBuilder`]: vk::DeviceCreateInfoBuilder
+    /// [`add_to_device_create`]: PhysicalDeviceFeatures::add_to_device_create
     /// [`Adapter::required_device_extensions`]: super::Adapter::required_device_extensions
     fn from_extensions_and_requested_features(
         device_api_version: u32,
@@ -222,7 +218,7 @@ impl PhysicalDeviceFeatures {
         Self {
             // vk::PhysicalDeviceFeatures is a struct composed of Bool32's while
             // Features is a bitfield so we need to map everything manually
-            core: vk::PhysicalDeviceFeatures::builder()
+            core: vk::PhysicalDeviceFeatures::default()
                 .robust_buffer_access(private_caps.robust_buffer_access)
                 .independent_blend(downlevel_flags.contains(wgt::DownlevelFlags::INDEPENDENT_BLEND))
                 .sample_rate_shading(
@@ -291,11 +287,10 @@ impl PhysicalDeviceFeatures {
                 //.shader_resource_residency(requested_features.contains(wgt::Features::SHADER_RESOURCE_RESIDENCY))
                 .geometry_shader(requested_features.contains(wgt::Features::SHADER_PRIMITIVE_INDEX))
                 .depth_clamp(requested_features.contains(wgt::Features::DEPTH_CLIP_CONTROL))
-                .dual_src_blend(requested_features.contains(wgt::Features::DUAL_SOURCE_BLENDING))
-                .build(),
+                .dual_src_blend(requested_features.contains(wgt::Features::DUAL_SOURCE_BLENDING)),
             descriptor_indexing: if requested_features.intersects(indexing_features()) {
                 Some(
-                    vk::PhysicalDeviceDescriptorIndexingFeaturesEXT::builder()
+                    vk::PhysicalDeviceDescriptorIndexingFeaturesEXT::default()
                         .shader_sampled_image_array_non_uniform_indexing(
                             needs_sampled_image_non_uniform,
                         )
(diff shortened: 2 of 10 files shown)


wgpu-hal/src/vulkan/device.rs
```diff
@@ -1452,119 +1418,116 @@ impl crate::Device for super::Device {
...
+                    let local;
+                    (local, acceleration_structure_infos) =
+                        acceleration_structure_infos.split_first_mut().unwrap();
 
-                    extra_descriptor_count += 1;
+                    *local = vk::WriteDescriptorSetAccelerationStructureKHR::default()
+                        .acceleration_structures(local_raw_acceleration_structures);
 
-                    write.push_next(&mut acceleration_structure_infos[index])
+                    write.descriptor_count(entry.count).push_next(local)
```
MarijnS95: Casual bug-fix here: `write.descriptor_count` should be equal to `acceleration_structure_count` in `vk::WriteDescriptorSetAccelerationStructureKHR`. Unless the selected range of `desc.acceleration_structures` (is `entry.count`) is always 1?

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -874,22 +850,19 @@ impl crate::Instance for super::Instance {
...
                 self.create_surface_android(handle.a_native_window.as_ptr())
             }
-            #[cfg(windows)]
             (Rwh::Win32(handle), _) => {
-                use winapi::um::libloaderapi::GetModuleHandleW;
-
-                let hinstance = unsafe { GetModuleHandleW(std::ptr::null()) };
-                self.create_surface_from_hwnd(hinstance as *mut _, handle.hwnd.get() as *mut _)
+                let hinstance = handle.hinstance.expect("HINSTANCE is not set.");
+                self.create_surface_from_hwnd(hinstance.get(), handle.hwnd.get())
```
MarijnS95: Any reason to call the Windows function when the `raw-window-handle` contains `hinstance` already? Probably and oversight since, as history suggests, when this code was introduced in 2021 on `raw-window-handle` the field was already there.

Cargo.toml
```diff
@@ -134,7 +134,7 @@ objc = "0.2.5"
 
 # Vulkan dependencies
 android_system_properties = "0.1.1"
-ash = "0.37.3"
+# ash = "0.37.3" # TODO: workspace deps are unused?
```
teoxoy: They are only used for some of the crates due to a bug in cargo. The bug has been fixed but we didn't yet re-enable full workspace inheritance; see https://github.com/gfx-rs/wgpu/issues/3407.

wgpu-hal/src/vulkan/instance.rs
teoxoy: I think it was originally introduced in https://github.com/gfx-rs/gfx/pull/1653 and since then copy-pasted.

But given it's optional, could it be that it's not set? I see `winit` sets it but it doesn't seem to be a requirement of `raw-window-handle`.

wgpu-hal/src/vulkan/command.rs
```diff
@@ -40,8 +40,10 @@ impl super::Texture {
 }
 
 impl super::DeviceShared {
-    fn debug_messenger(&self) -> Option<&ext::DebugUtils> {
-        Some(&self.instance.debug_utils.as_ref()?.extension)
+    /// Returns loaded [`ext::debug_utils`] for the device
+    // TODO: Inline this now-shorter function?
```
teoxoy: I think we should inline it.

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -1509,33 +1476,43 @@ impl super::Adapter {
...
+        let swapchain_fn = khr::swapchain::Device::new(&self.instance.raw, &raw_device);
 
-        let indirect_count_fn = if enabled_extensions.contains(&khr::DrawIndirectCount::name()) {
-            Some(khr::DrawIndirectCount::new(&self.instance.raw, &raw_device))
+        // TODO: instance/device extension.
+        let debug_utils_fn = if enabled_extensions.contains(&ext::debug_utils::NAME) {
+            Some(ext::debug_utils::Device::new(
+                &self.instance.raw,
+                &raw_device,
+            ))
```
teoxoy: `enabled_extensions` are device extensions. I think we can consult `self.instance.extensions` for instance extensions.

wgpu-hal/src/vulkan/device.rs
```diff
@@ -1452,119 +1418,116 @@ impl crate::Device for super::Device {
...
         let mut writes = Vec::with_capacity(desc.entries.len());
-        let mut buffer_infos = Vec::with_capacity(desc.buffers.len());
-        let mut sampler_infos = Vec::with_capacity(desc.samplers.len());
-        let mut image_infos = Vec::with_capacity(desc.textures.len());
-        let mut acceleration_structure_infos =
-            Vec::with_capacity(desc.acceleration_structures.len());
-        let mut raw_acceleration_structures =
-            Vec::with_capacity(desc.acceleration_structures.len());
+        let mut buffer_infos = vec![Default::default(); desc.buffers.len()];
+        let mut buffer_infos = buffer_infos.as_mut_slice();
```
teoxoy: I think we should create a new data structure based on a `Box<[MaybeUninit<T>]>`. It would only need an extend method that returns a `&[T]`.

teoxoy (changes requested):

wgpu-hal/src/vulkan/device.rs
MarijnS95: Sure, I've thought of that in the beginning, but the nature of the borrow checker requires the storage (a `Vec` or `Box`) to live separately from the mechanism that's mutably borrowed _or moved_ to keep track of the remaining space of the slice to write. `split_first_mut()` and `split_at_mut()` are critical helpers here as they splice the lifetime of the underlying `Vec`.

Implementing a data structure like you suggest, with a signature like:

```rust
fn extend(&mut self, values: IntoIterator<Item = T>) -> &[T] {}
```

is unfortunately impossible as it makes the returned slice _which `ash`'s builder-pattern correctly borrows for the remaining duration of `writes: Vec<vk::WriteDescriptorSet<'here>>`_ also borrow on `self`, which cannot be reused anymore in the next iteration of the loop. That is also why this implementation is repeatedly splitting the slice and reassigning the remainder to neatly get the lifetimes correct _without `unsafe`_.

I've pushed a wrapper now which re-adds `MaybeUninit` behaviour, relying on `Vec::with_capacity(x).spare_capacity_mut()` to get access to a `&mut [MaybeUninit<T>]` unfortunately a similar initializer for `box` is still unstable:

https://doc.rust-lang.org/std/boxed/struct.Box.html#method.new_uninit_slice

Let me know if there are any more improvements to this system!

wgpu-hal/src/vulkan/instance.rs
MarijnS95: Yeah, I'm not actually certain why and in what case it is supposed or allowed to be left at zero. Nor if it makes sense for `wgpu` to fall back to `GetModuleHandleW()` if it's unset.

I've never seen it being `None`/`0`, at least not with `winit`.

[Vulkan spec demands it to be valid](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkWin32SurfaceCreateInfoKHR.html), and I've converted the unwrap into returning a proper `InstanceError`: should we do the same to the `expect()`s at `Xlib` and `Xcb`?

wgpu-hal/src/vulkan/adapter.rs
MarijnS95: Correct, this is an instance extension that should not be enabled on the `instance`, I just didn't realize that the list of extensions was this close in reach :)

wgpu-hal/src/vulkan/instance.rs
teoxoy: I'm not sure, are these errors recoverable? I would generally lean towards errors for libraries; although in this case `InstanceError` is opaque... 😅

teoxoy: I think we should keep things consistent though.

@cwfitzgerald do you have an opinion on panic vs error for these?

MarijnS95: Not sure either, I'm not a windows expert on this front. We could keep the `.expect()` and ask the `raw-window-handle` devs what made them consider this optional?

Seems like there are multiple ways to get this opaque handle anyway, via `GetModuleHandleW(NULL)` for the current module or [`GetWindowLongPtrA(hwnd, GWLP_HINSTANCE)`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getwindowlongptra)...

wgpu-hal/src/vulkan/device.rs
teoxoy: Right, this is a problematic case for rust's borrow checker.

I think the new data structure is still better than nothing, thanks for adding it!

MarijnS95: Correct, the pattern is tricky for the borrow checker but _at least_ thanks to functions like `split_at_mut()`/`split_first_mut()` we can make them work in a lifetime+uniqueness-checked way while only doing one pre-allocation.

Would still be better if some of the `MaybeUninit` helpers are stabilized, but so be it :)

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -1208,15 +1174,17 @@ impl super::Instance {
...
         let (phd_capabilities, phd_features) = self.shared.inspect(phd);
 
         let info = wgt::AdapterInfo {
             name: {
-                cstr_from_bytes_until_nul(&phd_capabilities.properties.device_name)
-                    .and_then(|info| info.to_str().ok())
+                phd_capabilities
+                    .properties
+                    .device_name_as_c_str()
+                    .unwrap()
```
teoxoy: I don't think we should panic here.

wgpu-hal/src/vulkan/device.rs
```diff
@@ -531,8 +517,8 @@ impl
 }
 
 struct CompiledStage {
-    create_info: vk::PipelineShaderStageCreateInfo,
-    _entry_point: CString,
+    create_info: vk::PipelineShaderStageCreateInfo<'static>,
+    _entry_point: Pin<CString>,
```
teoxoy: I don't think this needs to be pinned. `CString` implements `Unpin`.

The pointer returned by `CString.as_ptr()` will be valid and won't change as long as we don't drop the `CString`.

wgpu-hal/src/vulkan/device.rs
```diff
@@ -1452,119 +1419,155 @@ impl crate::Device for super::Device {
...
         let set = vk_sets.pop().unwrap();
         if let Some(label) = desc.label {
-            unsafe {
-                self.shared
-                    .set_object_name(vk::ObjectType::DESCRIPTOR_SET, *set.raw(), label)
-            };
+            unsafe { self.shared.set_object_name(*set.raw(), label) };
+        }
+
+        struct ExtendStack<'a, T> {
```
teoxoy, line 1442: Could you add a doc comment explaining a bit why we need this? For future reference.

wgpu-hal/src/vulkan/device.rs
```diff
@@ -1452,119 +1419,155 @@ impl crate::Device for super::Device {
...
+                    // mutable reference which is also guaranteed to be valid for writes.
+                    unsafe {
+                        &mut *(to_init as *mut [MaybeUninit<T>] as *mut [T])
+                    }
+                };
+                (Self { remainder }, init)
+            }
+        }
+
+        // TODO: Pin?
```
teoxoy: Do we need to pin something?

wgpu-hal/src/vulkan/device.rs
```diff
@@ -1452,119 +1419,155 @@ impl crate::Device for super::Device {
...
         let mut writes = Vec::with_capacity(desc.entries.len());
         let mut buffer_infos = Vec::with_capacity(desc.buffers.len());
-        let mut sampler_infos = Vec::with_capacity(desc.samplers.len());
-        let mut image_infos = Vec::with_capacity(desc.textures.len());
+        let mut buffer_infos = ExtendStack::from_vec_capacity(&mut buffer_infos);
+        let mut image_infos = Vec::with_capacity(desc.samplers.len() + desc.textures.len());
+        let mut image_infos = ExtendStack::from_vec_capacity(&mut image_infos);
+        let mut acceleration_structure_infos = Vec::with_capacity(desc.textures.len());
```
teoxoy:
```suggestion
        let mut acceleration_structure_infos = Vec::with_capacity(desc.acceleration_structures.len());
```

wgpu-hal/src/vulkan/device.rs
```diff
@@ -1452,119 +1419,155 @@ impl crate::Device for super::Device {
...
         let mut acceleration_structure_infos =
-            Vec::with_capacity(desc.acceleration_structures.len());
+            ExtendStack::from_vec_capacity(&mut acceleration_structure_infos);
+        let mut raw_acceleration_structures = Vec::with_capacity(desc.textures.len());
```
teoxoy:
```suggestion
        let mut raw_acceleration_structures = Vec::with_capacity(desc.acceleration_structures.len());
```

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -220,7 +203,7 @@ impl super::Instance {
     ) -> Result<Vec<vk::ExtensionProperties>, crate::InstanceError> {
         let instance_extensions = {
             profiling::scope!("vkEnumerateInstanceExtensionProperties");
-            entry.enumerate_instance_extension_properties(layer_name)
+            unsafe { entry.enumerate_instance_extension_properties(layer_name) }
```
teoxoy, line 206: What are the safety requirements here?

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -595,7 +569,7 @@ impl crate::Instance for super::Instance {
         })?;
         let version = {
             profiling::scope!("vkEnumerateInstanceVersion");
-            entry.try_enumerate_instance_version()
+            unsafe { entry.try_enumerate_instance_version() }
```
teoxoy, line 572: What are the safety requirements here?

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -636,7 +610,7 @@ impl crate::Instance for super::Instance {
 
         let instance_layers = {
             profiling::scope!("vkEnumerateInstanceLayerProperties");
-            entry.enumerate_instance_layer_properties()
+            unsafe { entry.enumerate_instance_layer_properties() }
```
teoxoy, line 613: and here?

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -1035,18 +1003,18 @@ impl super::InstanceShared {
...
                     >= vk::API_VERSION_1_2
-                    || capabilities.supports_extension(vk::KhrDriverPropertiesFn::name());
+                    || capabilities.supports_extension(khr::driver_properties::NAME);
 
                 let supports_acceleration_structure =
-                    capabilities.supports_extension(vk::KhrAccelerationStructureFn::name());
+                    capabilities.supports_extension(khr::acceleration_structure::NAME);
 
-                let mut builder = vk::PhysicalDeviceProperties2KHR::builder();
+                let mut builder = vk::PhysicalDeviceProperties2KHR::default();
```
teoxoy: Can we rename this to something else? `properties2`?

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -1098,11 +1064,11 @@ impl super::InstanceShared {
         features.core = if let Some(ref get_device_properties) = self.get_physical_device_properties
         {
             let core = vk::PhysicalDeviceFeatures::default();
-            let mut builder = vk::PhysicalDeviceFeatures2KHR::builder().features(core);
+            let mut builder = vk::PhysicalDeviceFeatures2KHR::default().features(core);
```
teoxoy: same here, `features2`?

teoxoy (changes requested):

wgpu-hal/src/vulkan/device.rs
teoxoy: Makes sense to me.

@daniel-keitel do you have more context from https://github.com/gfx-rs/wgpu/pull/3507?

wgpu-hal/src/vulkan/device.rs
MarijnS95: Why is why `Unpin` has never made sense to me, since `(C)String` contain a pointer internally. If I can perform destructive actions on the contents of `(C)String` (e.g. `mem::replace()`, or appending to the string causing a realloc), how can I guarantee the memory address remains pinned?

wgpu-hal/src/vulkan/device.rs
MarijnS95: If my understanding of `Pin` is correct (see above: **it is not**) it _should_ prevent us from causing destructive actions on the pointer (`.push()`, `.resize()`, `.clear()`) after having `let mut` access to a `Vec`, while we can still `Deref` it into a `&mut [T]` to modify the elements without affecting where they are allocated.

I don't think that deref mechanism works for `spare_capacity_mut()` though.

(And having mutable borrows via `ExtendStack` prevents us from modifying the `Vec` anyway)

Please correct me about this.

wgpu-hal/src/vulkan/instance.rs
MarijnS95: Same as any other Vulkan call through `ash`. They just didn't happen to be marked `unsafe` before.

wgpu-hal/src/vulkan/device.rs
teoxoy: I looked at the docs and there is no way to mutate `CString`'s content. You can only do `mem::replace()` on the `CString` itself so the only way the pointer returned by `.as_ptr()` becomes invalid is if you drop the `CString`.

I find rust's safety docs to be top-notch and the docs of `.as_ptr()` indicate that the only invariant you have to care about is keeping the `CString` alive.

wgpu-hal/src/vulkan/device.rs
teoxoy: As long as there are references to subslices of the `Vec` around, we can't call any of `Vec`'s methods that take `&mut Self`, so we should be good.

wgpu-hal/src/vulkan/device.rs
MarijnS95: You are right that `CString` makes this easy. In the broader picture of `Pin` and [`Unpin`, its example](https://doc.rust-lang.org/std/marker/trait.Unpin.html) suggests a `String` that seems to be mutated in place and voids its pointer placement? `String::as_ptr()` doesn't have any documentation like `CString`.

MarijnS95: Actually, never mind. In that example the `Pin` doesn't own the `String` but a borrow of the `String`, which is different. I think my understanding and interpretation of `Pin` still holds then. After all, `String` is safe to move around, the thing it points to internally is not.

cwfitzgerald:
Really appreciate you spearheading the work on upgrading! Some conflicts need to be resolved and want Teo to give it one last look, but then we should be g2g.

teoxoy (commented):
There are 2 remaining discussions that we should address
- [Should some missing params in surface creation be errors or panics?](https://github.com/gfx-rs/wgpu/pull/5504#discussion_r1555435083) cc @cwfitzgerald @Wumpf 
- [Removing the Pin around CString](https://github.com/gfx-rs/wgpu/pull/5504#discussion_r1559382929)

ErichDonGubler:
@MarijnS95: Do you think you'll be able to attend to this soon? I'm excited to see quite a few duplicate dependencies like `libloading` 0.7 disappear once this lands. 🙂 I filed #5649 in anticipation. No pressure, though!

MarijnS95:
Before continuing further, it has to be said that those repeated "Changes requested" are quite off-putting. In the projects I work on that button is reserved as a heavy ban-hammer that basically means a full "NAK, go back to the drawing board". Any other nits and smaller comments use just that: the "leave a comment" mode which is neither an approval nor a full-on rejection.

Especially if most review comments are merely answering my questions.

---

Looks like we only have the surface creation panics to discuss, and the occasional acceleration structure descriptor-set-write drive-by-bugfix.

cwfitzgerald:
> Before continuing further, it has to be said that those repeated "Changes requested" are quite off-putting. In the projects I work on that button is reserved as a heavy ban-hammer that basically means a full "NAK, go back to the drawing board". Any other nits and smaller comments use just that: the "leave a comment" mode which is neither an approval nor a full-on rejection.

Definitely don't intend them to be off-putting! We use it basically as the middle of three options: reject entirely, need modifications, and approve. The "reject entirely" option for us would be commenting our disapproval of the strategy, then closing the PR entirely. Requesting changes also kicks the PR off our team review queue until review is re-requested. I think we could do better at communicating that the expected flow is `PR -> request changes -> fix/discuss -> re-request review -> (repeat) -> approval -> merge`.

MarijnS95:
> Requesting changes also kicks the PR off our team review queue

That's something I haven't considered yet, mostly being a lone-wolf maintainer where the TODO backlog is always "somehow further all PRs that are still open".

jimblandy:
> Before continuing further, it has to be said that those repeated "Changes requested" are quite off-putting. In the projects I work on that button is reserved as a heavy ban-hammer that basically means a full "NAK, go back to the drawing board".

I certainly don't intend my "changes requested" to be discouraging to anyone! I've always took them to mean exactly what they say: I'm requesting changes. I think the distinction I'm hoping to make is: If I only have suggestions, then I'll note them in the comments and "Approve", but if I see a bug or something that I think needs to be done differently, then I mark it "Changes requested". If I really wanted to end discussion of the whole approach, I'd close the PR. (What Connor said, basically.)

Anyway - just to say it out loud: @MarijnS95, thank you very much for this PR! Having `wgpu` be an open-source project is such a welcome contrast to most of the Firefox code base, whose contributors are almost exclusively Mozilla employees. I don't think Firefox's WebGPU implementation could possibly succeed with such a small team (me, @teoxoy, @ErichDonGubler, and @nical part-time) if we didn't also have contributions coming in on the wgpu side.

edit: [this comment](https://github.com/gfx-rs/wgpu/pull/5504#issuecomment-2091294413) from Erich is also a reflection of the value Mozilla, for our sake, finds in these kinds of contributions.

ErichDonGubler:
I think our community chat would be a good place to continue this discussion about perceptions with PR statuses, so I started a thread in Matrix: https://matrix.to/#/!FZyQrssSlHEZqrYcOb:matrix.org/$_lc2jnjTVMqi3SD2nsjSYmFdSSAlt7cO0wBahOcfT-0?via=mozilla.org&via=matrix.org&via=envs.net

wgpu-hal/src/vulkan/instance.rs
cwfitzgerald: InstanceError is fine here I think.

cwfitzgerald:
I think once merge conflicts are resolved, it is ready to go

teoxoy (approved):
Looks good to me as well!

wgpu-hal/src/vulkan/command.rs
```diff
@@ -469,23 +453,24 @@ impl crate::CommandEncoder for super::CommandEncoder {
         for desc in descriptors {
             let (geometries, ranges) = match *desc.entries {
                 crate::AccelerationStructureEntries::Instances(ref instances) => {
-                    let instance_data = vk::AccelerationStructureGeometryInstancesDataKHR::builder(
+                    let instance_data = vk::AccelerationStructureGeometryInstancesDataKHR::default(
+                    // TODO: Code is so large that rustfmt refuses to treat this... :(
```
MarijnS95, line 457: :disappointed:

cwfitzgerald (approved):
CI is mad, but lets go :)

MarijnS95:
@cwfitzgerald Maybe I broke that with the change in features for `winapi`? Though the `libloaderapi` feature that I moved around doesn't seem to impact whether `errhandlingapi` is enabled.

It is also a tad unfortunate that local `cargo clippy` on Rust 1.78 shows a bunch of `unnecessary qualification` warnings despite #5651 being merged, in code that wasn't touched in this PR.

MarijnS95:
The cause of this error is now obvious. With this PR there's no direct dependency on `libloading 0.7` anymore (there is via `glutin`, but it doesn't seem to be enabled by default, or at least not for the default target(s)/crate(s)), which used to turn on `errhandlingapi` on `winapi`:

```console
$ cargo tree -e features --target x86_64-pc-windows-msvc
...
│   │   ├── wgpu-hal v0.20.0 (/home/marijn/Code/TraverseResearch/wgpu/wgpu-hal)
...
│   │   │   ├── ash feature "default"
│   │   │   │   ├── ash v0.37.3+1.3.251
│   │   │   │   │   └── libloading feature "default"
│   │   │   │   │       └── libloading v0.7.4
│   │   │   │   │           ├── winapi feature "default" (*)
│   │   │   │   │           ├── winapi feature "errhandlingapi"
│   │   │   │   │           │   └── winapi v0.3.9
...
```

So we now have one more explicit feature dependency to list in `Cargo.toml`!

cwfitzgerald:
Thanks for the work on this and seeing this through to the end!

MarijnS95:
No worries, I hope you can all benefit/enjoy the incremental improvements in the latest `ash` release now!

PR #5500, opened 2024-04-05
Pipeline constants

teoxoy:
It's time 🎉 

Resolves https://github.com/gfx-rs/wgpu/issues/4366 & resolves https://github.com/gfx-rs/wgpu/issues/1762.

All commits have been previously approved in the following PRs:
- https://github.com/gfx-rs/wgpu/pull/4730
- https://github.com/gfx-rs/wgpu/pull/4845
- https://github.com/gfx-rs/wgpu/pull/4793
- https://github.com/gfx-rs/wgpu/pull/4998
- https://github.com/gfx-rs/wgpu/pull/5249
- https://github.com/gfx-rs/wgpu/pull/5387
- https://github.com/gfx-rs/wgpu/pull/5419
- https://github.com/gfx-rs/wgpu/pull/5492
- https://github.com/gfx-rs/wgpu/pull/5495

There are still 2 features that remain to be implemented:
- https://github.com/gfx-rs/wgpu/issues/4450
- https://github.com/gfx-rs/wgpu/issues/5315

Commits should not be squashed (they are self-contained and passed CI individually).

Issue #4592, opened 2023-10-27
Speed up zero initialization of workgroup memory


raphlinus:
This is related to #4591; when forcing `spv::ZeroInitializeWorkgroupMemoryMode::Polyfill` in `device_from_raw()`, we observe very slow (but correct!) behavior for zeroing the workgroup shared array - all the work is done on one thread. It would be better to distribute this; in this case the array size and workgroup size match, so for each invocation to zero one array element would be simple and efficient.

[zerooooooo.zip](https://github.com/gfx-rs/wgpu/files/13191956/zerooooooo.zip)

Repro case is the same as the linked bug, but changing line 1307 of `vulkan/adapter.rs` to `Polyfill`.

cwfitzgerald:
For reference, the zero-init code:

SPIRV: https://github.com/gfx-rs/wgpu/blob/trunk/naga/src/back/spv/writer.rs#L1327  
MSL: https://github.com/gfx-rs/wgpu/blob/trunk/naga/src/back/msl/writer.rs#L4441-L4549  
HLSL: https://github.com/gfx-rs/wgpu/blob/trunk/naga/src/back/hlsl/writer.rs#L1280-L1305  
GLSL: https://github.com/gfx-rs/wgpu/blob/trunk/naga/src/back/glsl/mod.rs#L1688-L1718

I think the easiest lift thing to do is that, for top level arrays, use the local index to init that element of the array, masking off the higher invocations than the array length, and doing a compile time loop for arrays longer than the element count.

cwfitzgerald:
To be clear, I think the init shoudl look like this:

```wgsl
var<workgroup> array1: array<u32, 652>;
var<workgroup> array2: array<u32, 256>;
var<workgroup> array3: array<u32, 45>;
var<workgroup> non_array: u32;

@compute @workgroup_size(16, 16)
fn main(@builtin(local_index) local_index: u32) {
    // All unconditional array init
    // Do loop at compile time, just generate multiple writes for long arrays
    array1[local_index] = <zero init>;
    array1[local_index + 256] = <zero init>;
    array2[local_index] = <zero init>;
    if local_index < 140 {
        // Conditional part of array1 is in conditional
        array1[local_index + 512] = <zero init>;
        if local_index < 45 {
            array3[local_index] = <zero init>;
            if local_index < 1 {
                non_array = <zero init>;
            }
        }
    }
    workgroupBarrier();
}
```

FL33TW00D:
[...]

PR #5508, opened 2024-04-08
Allow configuring whether workgroup memory is zero initialised

DJMcNab:
**Connections**

This is related to #5319 - a key reason initialisation takes so long is the 5x increase in pipeline creation time added by this feature
See also #4592, which this *does not* fix.
I'm planning on racing this PR with a fix for #4592, in the hope that I can get one of them in 0.20
(Both PRs should be useful, however.)

**Description**
Allows skipping zero initialisation of pipeline workgroup variables.

Backends:

- [X] Vulkan
- [x] DirectX
- [x] GLES
- [x] Metal

The underlying pipeline creation time issue is a really significant issue for adoption of Vello and Xilem, as it makes our projects take an extremely long time to start up.
This decreases pipeline creation time by a factor of (nearly) 5 on Android, which makes this a worthwhile tradeoff for us.

Renderer creation time with different modes (all values in millseconds). `None` corresponds to the new `ZeroInitializeWorkgroupMemory::never()`. This is on a Pixel 6

|Threads| Single | Multi
|---|---|---
|Native | 4228 | 1829
|Polyfill | 6350 | 2772
|None | 790 | 370

This is a breaking change, but this breaking change is shared with #5500

**Testing**

I have tested the improperly wired up version of this in [Vello](https://github.com/linebender/vello/), to get the above numbers.

I have also tested this PR in Vello, which showed the expected decrease in startup time

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo xtask test` to run tests - same story as #5319 - no new failures


jimblandy:
It's not great that we're adding this field that needs to be set in every pipeline descriptor anyone will ever write. But I understand that Vello wants to be a good neighbor when sharing `Device`s with other code, so we can't make it a device-wide flag.

Could we see an alternative version of this PR where the flag is on `wgpu_core::pipeline::ShaderModuleDescriptor`, rather than `ComputePipelineDescriptor`?

jimblandy (dismissed):
It's not great that we're adding this field that needs to be set in every pipeline descriptor anyone will ever write. But I understand that Vello wants to be a good neighbor when sharing Devices with other code, so we can't make it a device-wide flag.

Could we see an alternative version of this PR where the flag is on wgpu_core::pipeline::ShaderModuleDescriptor, rather than ComputePipelineDescriptor?

DJMcNab:
> Could we see an alternative version of this PR where the flag is on wgpu_core::pipeline::ShaderModuleDescriptor, rather than ComputePipelineDescriptor?

Do you mean as a seperate PR, or reusing this same one?

teoxoy:
It's worth noting we don't have the `max_compute_workgroup_storage_size` limit fully implemented (it's not doing anything).
Nor the 16k "maximum byte-size of an array type instantiated in the workgroup address space" from the WGSL spec.

Are the shaders in question surpassing those limits?

teoxoy:
Using a loop to do array initialization might also side-step the issue (if it doesn't get unrolled that is).

DJMcNab:
As it happens, we are exactly on the limit for `maxComputeWorkgroupStorageSize`, based on a quick check (taking the values in https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/DirectX12.20extremely.20slow.20startup.20time/near/431594219 as my source of truth for that claim, we have 16 lots of 256 lots of 4 bytes). So maybe, if the 16384 is an exclusive maximum.

DJMcNab:
Yes, as I say, fixing #4592 would also likely resolve this issue, as it would cause us to output probably ~24 instructions rather than 16384

But this is much easier to implement, and the actual zero initialisation is superfluous work anyway

Besides which, wgpu would still use the native zero initialisation method on Vulkan Android, which would mean we would still need to disable it. The native method is at least faster than the polyfill. But it would be slower if #4592 reaches the same result

DJMcNab:
Moving into draft whilst I reconfigure as requested

CHANGELOG.md
DJMcNab, line 1: Oh, this is frustrating - format on save is often great, but sometimes it can be annoying :(

DJMcNab:
This now shares the breaking change with #5500, using the mechanism discussed on Matrix. So it is a breaking change, but the "same" breaking change was already needed from. CC @teoxoy , as you brought in the `constants` field which has been moved.

I'm now quite happy with how this API looks - it also removes the need for everyone to set `constants`, which was a bit weird as `&Default::default`

cwfitzgerald (approved):
Seems pretty uncontroversial!

PR #5509, opened 2024-04-09
[core] Replace id `transmute` method with explicit functions.

jimblandy:
Replace the `wgpu_core::id::Id::transmute` method, the `transmute` private module, and the `Transmute` sealed trait with some associated functions with obvious names.

### DIFF
--- a/player/src/bin/play.rs
+++ b/player/src/bin/play.rs
@@ -87,7 +87,7 @@ fn main() {
                 &desc,
                 None,
                 Some(id),
-                Some(id.transmute())
+                Some(id.into_queue_id())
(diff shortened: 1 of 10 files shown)


jimblandy:
(No change in behavior, just less rigamarole.)

If the reviewer would prefer, I'd be happy to turn this into `From` implementations instead. I chose this way because it seemed good to discourage accidental generic conversion, but either way would be an improvement.

cwfitzgerald (approved):
Much clearer!

sagudev:
queue_id -> device_id is not possible anymore :(

PR #5512, opened 2024-04-09
[core] Provide an explicit type for some CommandBuffer pointers.

jimblandy:
To make sure rust-analyzer's "find references" feature works, spell out the return type for some calls to `map_pass_err`, which it is for some reason unable to infer.

### DIFF
--- a/wgpu-core/src/command/compute.rs
+++ b/wgpu-core/src/command/compute.rs
@@ -36,6 +36,7 @@ use serde::Serialize;
 
 use thiserror::Error;
 
+use std::sync::Arc;
 use std::{fmt, mem, str};
 
 #[doc(hidden)]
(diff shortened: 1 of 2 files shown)


wgpu-core/src/command/compute.rs
```diff
@@ -365,7 +365,8 @@ impl Global {
 
         let hub = A::hub(self);
 
-        let cmd_buf = CommandBuffer::get_encoder(hub, encoder_id).map_pass_err(pass_scope)?;
+        let cmd_buf: std::sync::Arc<CommandBuffer<A>> =
```
cwfitzgerald: Could we import Arc

cwfitzgerald (approved):
Nice approved with nits, did a pass on these at one point

PR #5513, opened 2024-04-09
[core] Use `expect` instead of "if let else panic".

jimblandy:
Use `Option::expect` to check the result from `Arc::into_inner`, rather than `if let Some ... else panic!`.


cwfitzgerald (approved):
Nice

PR #5516, opened 2024-04-10
[core] Document command encoding and command buffers.

jimblandy:
Flesh out the documentation for `wgpu_core`'s `CommandBuffer`, `CommandEncoder`, and associated types.

### DIFF
--- a/wgpu-core/src/command/mod.rs
+++ b/wgpu-core/src/command/mod.rs
@@ -38,23 +38,115 @@ use crate::device::trace::Command as TraceCommand;
 
 const PUSH_CONSTANT_CLEAR_ARRAY: &[u32] = &[0_u32; 64];
 
+/// The current state of a [`CommandBuffer`].
 #[derive(Debug)]
 pub(crate) enum CommandEncoderStatus {
+    /// Ready to record commands. An encoder's initial state.
+    ///
+    /// Command building methods like [`command_encoder_clear_buffer`] and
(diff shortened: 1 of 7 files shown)


jimblandy:
It may look weird that this requirement got deleted from `Queue::submit`'s documentation in `wgpu_hal/src/lib.rs`:

> - all of the command buffers had `CommandBuffer::finish()` called.

However, there is no longer any method named `finish` (or any methods on `CommandBuffer` at all). With the current `CommandEncoder` trait, it's not possible to obtain a `CommandBuffer` at all until it's been finished.

wgpu-core/src/command/mod.rs
```diff
@@ -38,23 +38,115 @@ use crate::device::trace::Command as TraceCommand;
 
 const PUSH_CONSTANT_CLEAR_ARRAY: &[u32] = &[0_u32; 64];
 
+/// The current state of a [`CommandBuffer`].
 #[derive(Debug)]
 pub(crate) enum CommandEncoderStatus {
+    /// Ready to record commands. An encoder's initial state.
+    ///
+    /// Command building methods like [`command_encoder_clear_buffer`]
+    /// and `command_encoder_run_compute_pass` require the encoder to
```
Wumpf:
```suggestion
    /// and [`command_encoder_run_compute_pass`] require the encoder to
```

Wumpf (approved):
nice! even learned a few new things while reading
looks like a lot of things could be named better

PR #5517, opened 2024-04-11
[v0.19] Fix Buffer Mapping Deadlock

cwfitzgerald:
Found this while using wgpu-profiler.

Subtle temporary extension caused a deadlock in buffer mapping code.

### DIFF
--- a/wgpu-core/src/device/life.rs
+++ b/wgpu-core/src/device/life.rs
@@ -791,30 +791,35 @@ impl<A: HalApi> LifetimeTracker<A> {
                 *buffer.map_state.lock() = resource::BufferMapState::Idle;
                 log::trace!("Buffer ready to map {:?} is not tracked anymore", buffer_id);
             } else {
-                let mapping = match std::mem::replace(
+                // This _cannot_ be inlined into the match. If it is, the lock will be held
+                // open through the whole match, resulting in a deadlock when we try to re-lock
+                // the buffer back to active.
+                let mapping = std::mem::replace(
                     &mut *buffer.map_state.lock(),
                     resource::BufferMapState::Idle,
-                ) {
+                );
+                let pending_mapping = match mapping {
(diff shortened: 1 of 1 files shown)


Wumpf (changes requested):
oh gawd. this is insanely evil
you're missing a changelog note though, imho worth pointing out that there's a fix (we should give people reasons to update ;-))

Wumpf (approved):

Wumpf:
changelog got promised to be done upon the release processing :)

wgpu-core/src/device/life.rs
```diff
@@ -791,30 +791,35 @@ impl<A: HalApi> LifetimeTracker<A> {
                 *buffer.map_state.lock() = resource::BufferMapState::Idle;
                 log::trace!("Buffer ready to map {:?} is not tracked anymore", buffer_id);
             } else {
-                let mapping = match std::mem::replace(
+                // This _cannot_ be inlined into the match. If it is, the lock will be held
+                // open through the whole match, resulting in a deadlock when we try to re-lock
+                // the buffer back to active.
+                let mapping = std::mem::replace(
                     &mut *buffer.map_state.lock(),
```
REASY, line 798: Should [ReentrantMutex](https://docs.rs/parking_lot/latest/parking_lot/type.ReentrantMutex.html) used to prevent double locking causing deadlock?
```
A mutex which can be recursively locked by a single thread.
This type is identical to Mutex except for the following points:

- Locking multiple times from the same thread will work correctly instead of deadlocking.
- ReentrantMutexGuard does not give mutable references to the locked data. Use a RefCell if you need this.
```

cwfitzgerald: No - there is a cost to reentrant mutexes and none of our mutexes require reentrency.

We have a PR open which adds a debug check to make sure we don't lock locks in a way that could cause a deadlock.

cwfitzgerald: #5539

PR #5518, opened 2024-04-11
Fix Buffer Mapping Deadlock

cwfitzgerald:
Forward port of #5517


Wumpf (changes requested):
same as #5517, this should have a changelog note imho

ErichDonGubler (approved):
Approving from Mozilla's side, for bookkeeping.

Wumpf (approved):

cwfitzgerald:
Changelog covered by #5552

PR #5519, opened 2024-04-11
repo: add `shell.nix` and `.envrc`

MultisampledNight:
**Connections**
- Suggested by @cwfitzgerald on Discord after someone asked how to use wgpu on NixOS: https://discord.com/channels/273534239310479360/335502453371961344/1227753511094194287

**Description**

Makes usage on NixOS easier, by not having to drop into a manual hurry of `nix-shell -p lots of packages` every time someone wants to experiment around with wgpu.
The shell.nix also installs Rust according to the MSRV listed in `rust-toolchain.toml`, as well as a few helpers (notably, *not* `cargo-run-wasm` since it's not on nixpkgs. Could just run `cargo install cargo-run-wasm` in the shell hook, otoh building takes some time and can also be done on-demand).

The .envrc causes shell.nix to be eval'd automatically _iff_

1. [direnv](https://direnv.net/) is installed
2. the user `direnv allow`ed the shell.nix

**Testing**

Not at all. _You_ can test this by just running `nix-shell` with this PR checked out, and checking whether cargo and the like are available and the examples run. See https://nix.dev/install-nix#install-nix for installing nix.

**Checklist**

None are done, as none of these are applicable for a non-code change.

- [ ] Run `cargo fmt`.


NyxAlexandra:
is there a reason this isn't a Flake?

MultisampledNight:
> is there a reason this isn't a Flake?

@NyxAlexandra nope, no rational one! i don't use flakes personally as they're not stable, so i don't know enough about how to write them. hence you or someone else would need to adjust this accordingly

shell.nix
```diff
@@ -0,0 +1,47 @@
+{ pkgs ? import <nixpkgs> {} }:
```
cwfitzgerald, line 23: Could you just add some documentation in a comment what this is, why one would care about it - most people aren't on a platform where this matters, so having clear documentation is important.

shell.nix
```diff
@@ -0,0 +1,47 @@
...
+    shaderc directx-shader-compiler
+    libGL
+    vulkan-headers vulkan-loader
+    vulkan-tools vulkan-tools-lunarg
+    vulkan-validation-layers
+    vulkan-extension-layer
+
+    python3 valgrind
+    renderdoc
+    gnuplot
```
cwfitzgerald: it'd probably be good to document which of these are manditory and which are nice-to-haves. And why we need them.

cwfitzgerald (changes requested):
Just needs comments, then good to land!

cwfitzgerald (approved):
Exactly what I was looking for! Thanks!

MultisampledNight:
@cwfitzgerald Thank you for asking about inclusion of something you don't use personally, but have seen others to need it!!

eddyb:
**EDIT**: GitHub won't let me leave a PR review, so ftr everything below is about this:
https://github.com/gfx-rs/wgpu/blob/14abdd47547e0cd364e8431dec46c3d57be3e160/shell.nix#L60-L68

---

This `rustup default ...` invocation is pretty intrusive (it affects the user's personal `~/.rustup` which may have nothing to do with individual projects!).

I got really confused as I happened to have a much smaller `default.nix` (and the same `use nix` in `.envrc`), and just switching to the latest `wgpu` triggered all of this, but that's not really a big issue and I would probably use this `shell.nix` if it didn't mutate user-wide state.

---

If the concern is `rustup component add` needing to apply to a specific toolchain, you can pass `--toolchain=$RUSTC_VERSION` to it.
Otherwise, I would expect `rust-toolchain.toml` to handle all of the overriding necessary.

Worst case, you can `export RUSTUP_TOOLCHAIN="..."` (i.e. rename `RUSTC_VERSION`), which is the `rustup` internal mechanism (which allows e.g. `rustup`-wrapped `cargo` to invoke the same version of `rustc` no matter the working directory etc.)<sup>1</sup> - and it applies to `rustup` itself too:
```console
$ RUSTUP_TOOLCHAIN=beta rustup show active-toolchain
beta-x86_64-unknown-linux-gnu (environment override by RUSTUP_TOOLCHAIN)
```

<sub><sup>1</sup> (though I'm not 100% sure `rustup` always sets `RUSTUP_TOOLCHAIN`, IIRC there've been some platform-specific issues and back & forth - it may have been considered an optimization at some point?)</sub>


Oh and the `+` syntax still works if `$RUSTUP_TOOLCHAIN` is set, and takes over as usual, e.g.:
```console
$ RUSTUP_TOOLCHAIN=beta rustc -V
rustc 1.82.0-beta.3 (4976ae480 2024-09-09)
$ RUSTUP_TOOLCHAIN=beta rustc +stable -V
rustc 1.81.0 (eeb90cda1 2024-09-04)
```

So if you want a "dev shell scoped" toolchain override, `$RUSTUP_TOOLCHAIN` is probably the best option.

---

<del>Oh but there's still noise from `rustup component add` every time, I don't know a good way to automatically trigger a check with nothing printed out if nothing has changed.</del>

---

Ended up wasting a bunch of time golfing a mix of stdenv shell hacks and trivial builders:
https://gist.github.com/eddyb/aa5a1485d2fbc912d5664a252d85caf7
(with explanations in a gist comment, but also see [the diff from `wgpu` trunk](https://gist.github.com/eddyb/aa5a1485d2fbc912d5664a252d85caf7/revisions))

It's not worth a full PR without going through the deps and splitting them by usecase, just wanted to note it down because at least the `rustup default ...` part can be disruptive.

PR #5520, opened 2024-04-11
[spv-in] add support for specialization constants

### LINKED ISSUE #4462: Support for OpSpecConstantComposite [SPIR-V]

I've been trying to compile and run some OpenCL C kernels in `wgpu` as SPIR-V modules and have run into the following validation error within `wgpu`:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: UnsupportedInstruction(Type, SpecConstantComposite)', /Users/nicksarten/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.12.0/src/backend/direct.rs:1022:45
```

I've had a quick search around to see if anyone else has run into a similar issue, but the closest I found was [this issue in `naga`](https://github.com/gfx-rs/naga/issues/39) which seems to suggest that `OpConstantComposite` is supported by `naga` at least, but there's no mention of `OpSpecConstantComposite`.

`clspv` seems to be using this instruction to define the workgroup size variable, which is a `v3uint`.

I'd like to be able to run this kernel with validation turned on and not get an `UnsupportedInstruction` error for `SpecConstantComposite`.

I'm going to attempt to run the kernel using SPIRV-passthrough.

I've attempted tweaking my `clspv` invocation to change how the work group size is specified. Using `--vec3-to-vec4` to force lowering of `vec3` to `vec4` didn't seem to have any effect on the SPIR-V at all.

**Additional context**
I've attached the SPIRV disassembly for the relevant kernel.
[resize.txt](https://github.com/gfx-rs/wgpu/files/8076774/resize.txt)

teoxoy:
This PR should get the Dota 2 shaders to work again (tested via the lazy CI job).
Also resolves https://github.com/gfx-rs/wgpu/issues/4462.

Some more context:

While implementing proper support for pipeline-overridable constants we first removed the existing code from the spir-v frontend and IR dealing with spec constants (in https://github.com/gfx-rs/wgpu/commit/b3dfc40c9df03147f0047820937dbfac236cdc70 & https://github.com/gfx-rs/wgpu/commit/f949ea69c426ab564d991c3dfc33b73bdf4187cb#diff-da9de19bca31f63903511682e750d4a03d5ffaae563f0a95d23b3f6d41bc0391) since we took a slightly different approach.

We thought this was fine to remove and add back later because only the spir-v frontend and IR supported spec constants i.e. you couldn't do anything useful with spec constants before; but it seems the lazy CI job had some testcases that were relying on the spec constants at least to be parsed by the frontend.

cc @cwfitzgerald


jimblandy (approved):
Fantastic!

PR #5521, opened 2024-04-11
Improve the polyfill for workgroup variable zero initialization

### LINKED ISSUE #4592: Speed up zero initialization of workgroup memory

DJMcNab:
**Connections**
Fixes #4592 

**Description**
Implements the suggestion in https://github.com/gfx-rs/wgpu/issues/4592#issuecomment-1783348185

I have implemented this for Spir-v, GLSL and HLSL.

Having implemented this, I have found enough issues that I am unsure about continuing this:
1) It is an overall performance regression on my laptop for Vulkan
2) Therefore, deciding on when to apply the polyfill is non-trivial
3) #5508 turned out to be quite clean, and better solves the issue for Vello in the near term
4) Better heuristics are possible, but are not plausible in the current architecture, as they have to be implemented manually in each backend.

If someone else wants to take this on, feel free.

**Testing**
I have added a new  test which tests the worse case of the old initialisation.

**Checklist**

- [X] Run `cargo fmt`.
  - [X] `--target wasm32-unknown-unknown` N/A
  - [X] `--target wasm32-unknown-emscripten` N/A


DJMcNab:
Realistically, this isn't going anywhere. On the Vello side, we only depend on this in the web context, which isn't a current priority. This does still need resolving, but the infrastructure isn't there to do so easily.

PR #5526, opened 2024-04-12
[core] Make `wgpu_core::command::BakedCommands` pub(crate), not pub.

jimblandy:


Wumpf (approved):
I think we're going wrong about infecting everything with `pub(crate)`. The cleaner way it to use `pub` or nothing everywhere except `mod.rs` and `lib.rs` where the visibility rules from one module to the next can be specified more easily.

(that said, nothing wrong with this individual change)

PR #5532, opened 2024-04-14
Make `scalar_width` return size in bytes

### LINKED ISSUE #5278: Make TypeInner::scalar_width return the width in bytes
TypeInner::scalar_width currently returns the scalar width in bits. This is counter-intuitive, when the width is stored in bytes and assumed to be bytes everywhere else. If the current functionality is to be maintained, it should be renamed to `scalar_width_bits` and documented.

https://github.com/gfx-rs/wgpu/blob/trunk/naga/src/proc/mod.rs#L267

atlv24:
**Connections**
Resolves #5278

**Description**
`scalar_width` returns size in bits, despite the rest of the codebase dealing exclusively in byte scalar widths, and it does so with no documentation or specification in the function name. This has shot me in the foot 3 times already, so I figured its probably shooting other people's feet.

**Testing**
Existing test cases

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/naga/src/back/glsl/mod.rs
+++ b/naga/src/back/glsl/mod.rs
@@ -3418,7 +3418,8 @@ impl<'a, W: Write> Writer<'a, W> {
                         let scalar_bits = ctx
                             .resolve_type(arg, &self.module.types)
                             .scalar_width()
-                            .unwrap();
+                            .unwrap()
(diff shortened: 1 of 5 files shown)


teoxoy (approved):
Thanks!

PR #5533, opened 2024-04-14
Refactor MSL atomic writing

atlv24:
**Connections**
Just stuff I found while writing #5383

**Description**
MSL atomic writing code structure is inconsistent with glsl/hlsl/wgsl

**Testing**
No functional changes

**Checklist**

- [x] Run `cargo fmt`.


naga/src/back/msl/mod.rs
```diff
@@ -538,3 +538,19 @@ fn test_error_size() {
...
+        match self {
+            Self::Add => "fetch_add",
+            Self::Subtract => "fetch_sub",
+            Self::And => "fetch_and",
+            Self::InclusiveOr => "fetch_or",
+            Self::ExclusiveOr => "fetch_xor",
+            Self::Min => "fetch_min",
+            Self::Max => "fetch_max",
+            Self::Exchange { compare: None } => "exchange",
+            Self::Exchange { compare: Some(_) } => "", //TODO
```
teoxoy: I think this variant should return `Error::FeatureNotImplemented` itself.

teoxoy (changes requested):

naga/src/back/msl/mod.rs
atlv24: I don't fully agree but i've made the changes. It required removing `const`, and its not consistent with these other ones anymore:
https://github.com/gfx-rs/wgpu/blob/trunk/naga/src/back/hlsl/conv.rs#L223
https://github.com/gfx-rs/wgpu/blob/trunk/naga/src/back/glsl/mod.rs#L86

teoxoy: That's fair, I didn't notice those were like that.
Let's just merge this and once we implement that functionality it won't matter anymore.

teoxoy (approved):

PR #5534, opened 2024-04-14
Refactor WGSL atomic writing

atlv24:
**Connections**
Just stuff I found while writing #5383

**Description**
WGSL atomic writing code can be more concise

**Testing**
No functional changes

**Checklist**

- [x] Run `cargo fmt`.


teoxoy (changes requested):
This is more concise but we now need to keep those in sync with the ones in `to_atomic`.

I think we should instead move in this direction:

https://github.com/gfx-rs/wgpu/blob/c9212c6d463160d144dacdfbf1a2330f2abfe607/naga/src/front/wgsl/lower/mod.rs#L2055-L2057

Where we call `crate::AtomicFunction::from_wgsl()` and it returns an option.

teoxoy (approved):
Good stuff!

PR #5535, opened 2024-04-14
Fix surfaces only compatible with first enabled backend

### LINKED ISSUE #5289: [wgpu-core] request_adapter fails when requesting with a specific backend
**Description**
[Instance::request_adapter](https://docs.rs/wgpu-core/latest/wgpu_core/global/struct.Global.html#method.request_adapter) fails when instance is created with `Backends::all()`, and then a surface is initialized, and then request_adapter is called with a specific backend in an environment with 2 or more backends.

**Repro steps**
1. Ensure you have two or more different backends.
2. [Update `instance_request_adapter`](https://github.com/gfx-rs/wgpu/blob/440574f92777c1e42a3355aa394c317645a6531b/wgpu/src/backend/wgpu_core.rs#L562) to pass a single backend like `wgt::Backends::DX12/VULKAN/METAL` instead of `wgt::Backends::all()`.
3. Run the hello_triangle example
4. Notice request_adapter fails, if it doesn't fail then cycle through other available backends on Step 2.

request_adapter should succeed

**Platform**
macOS 14, Metal + MoltenVK

### LINKED ISSUE #5324: DX12 adapter does not support surface when multiple backends are specified in the instance
**Description**
When I create my instance with default options or something that specifies more than one backend such as `Backends::PRIMARY` the surface (a default window created with winit) would not be compatible with any one of the DX12 adapters returned by `enumerate_adapters`. Instead, if I created the instance specifying DX12 backend in the options, DX12 adapters work fine.

**Repro steps**
A minimal example can be found at [https://github.com/ItsSunnyMonster/wgpu_crash_minimal](https://github.com/ItsSunnyMonster/wgpu_crash_minimal)

The returned adapters should have the same behaviour regardless of what backends were specified when creating the instance.

**Platform**
**OS**: Windows 10 22H2 build 19045.4046 (Reproducible on Windows 11 and presumably other versions of windows)
**GPU**: AMD Radeon Vega 8 Graphics (Reproducible on machines with a discrete GPU, tested on NVidia RTX 4060 and NVidia Quadro P2200)
**wgpu version**: 0.19.1

Wumpf:
**Connections**
* Fixes #5324

**Description**
Both instances and surfaces are supposed to be created for all backends. This has some major drawbacks (see e.g. #2384 - creation of one halsurface may preclude creation of another halsurface), but this is how things were assumed to be working. Only that they didn't and we only created a single `AnySurface` with only the first successfully create hal surface!

This PR sets things right again and removing `AnySurface` in the process since there's not much use to it anymore.
We now create hal surface for each available hal instance. Error handling had to change a bit for this (not super happy with it, but will do).

Some small renames & doc additions as well.

**Testing**
* Checked the repro case in #5324
* confirmed to fix #5289

**Checklist**

- [x] Run `cargo fmt`.


wgpu-core/src/instance.rs
```diff
@@ -461,13 +466,21 @@ pub enum RequestAdapterError {
...
+    BackendNotEnabled(Backend),
+    #[error("Failed to create surface for any enabled backend.")]
+    InstanceError(HashMap<Backend, hal::InstanceError>),
 }
 
 impl Global {
+    /// Creates a new surface targeting the given display/window handles.
+    ///
+    /// Internally attempts to create hal surfaces for all enabled backends.
+    /// TODO DO NOT MERGE: DESCRIBE BONKERS ERROR HANDLING
```
Wumpf: ah yeah I should really do that

cwfitzgerald (approved):
LGTM

PR #5536, opened 2024-04-14
[core] Also trace snatch lock write guard acquisitions.

jimblandy:
In debug builds, check for attempts to acquire snatch lock read guards while holding write guards, not just attempts to acquire a second read guard.

It's because write locks weren't checked that the validation added in #5426 didn't catch #5471.

### DIFF
--- a/wgpu-core/src/snatch.rs
+++ b/wgpu-core/src/snatch.rs
@@ -64,8 +64,58 @@ impl<T> std::fmt::Debug for Snatchable<T> {
 
 unsafe impl<T> Sync for Snatchable<T> {}
 
+struct LockTrace {
+    purpose: &'static str,
+    caller: &'static Location<'static>,
+    backtrace: Backtrace,
+}
+
+impl std::fmt::Display for LockTrace {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "a {} lock at {}\n{}",
+            self.purpose, self.caller, self.backtrace
+        )
+    }
+}
+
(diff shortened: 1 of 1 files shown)


wgpu-core/src/snatch.rs
```diff
@@ -64,8 +64,61 @@ impl<T> std::fmt::Debug for Snatchable<T> {
...
+    }
+
+    fn exit() {
+        READ_LOCK_LOCATION.take();
+    }
+}
+
+#[cfg(not(debug_assertions))]
+impl LockTrace {
+    #[inline]
```
SludgePhD: (these aren't needed anymore on trivial functions, at least on current Rust versions)

SludgePhD (approved):

wgpu-core/src/snatch.rs
```diff
@@ -64,8 +64,58 @@ impl<T> std::fmt::Debug for Snatchable<T> {
...
+
+#[cfg(not(debug_assertions))]
+impl LockTrace {
+    fn enter(purpose: &'static str) {}
+    fn exit() {}
+}
+
 thread_local! {
-    static READ_LOCK_LOCATION: Cell<Option<(&'static Location<'static>, Backtrace)>> = const { Cell::new(None) };
+    static READ_LOCK_LOCATION: Cell<Option<LockTrace>> = const { Cell::new(None) };
```
cwfitzgerald:
```suggestion
    static SNATCH_LOCK_TRACE: Cell<Option<LockTrace>> = const { Cell::new(None) };
```

maybe?

cwfitzgerald (approved):
Good after nit

PR #5537, opened 2024-04-15
64 bit Image Atomics

### LINKED ISSUE #5009: Image (64bit) atomic support
Missing extension that's very useful for doing compute shader based software raster.

Support for https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_shader_image_atomic_int64.html

**Additional context**
https://advances.realtimerendering.com/s2021/Karis_Nanite_SIGGRAPH_Advances_2021_final.pdf#page=84

atlv24:
**Connections**
Fixes #5009
Depends on #6706

**Description**
Adds 64 bit image atomics to Vulkan, DirectX, and Metal backends. Emulates using RG32Uint on platforms where R64Uint is not available (directX, metal). 

**Testing**
naga tests and runtime tests are included

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/naga/src/back/glsl/mod.rs
+++ b/naga/src/back/glsl/mod.rs
@@ -4944,6 +4944,7 @@ fn glsl_storage_format(format: crate::StorageFormat) -> Result<&'static str, Err
         Sf::Rgb10a2Uint => "rgb10_a2ui",
         Sf::Rgb10a2Unorm => "rgb10_a2",
         Sf::Rg11b10Ufloat => "r11f_g11f_b10f",
+        Sf::R64Uint => "r64ui",
         Sf::Rg32Uint => "rg32ui",
         Sf::Rg32Sint => "rg32i",
         Sf::Rg32Float => "rg32f",
--- a/naga/src/back/msl/writer.rs
+++ b/naga/src/back/msl/writer.rs
@@ -1212,7 +1212,11 @@ impl<W: Write> Writer<W> {
     ) -> BackendResult {
         write!(self.out, "{level}")?;
         self.put_expression(image, &context.expression, false)?;
-        let op = fun.to_msl();
+        let op = if context.expression.resolve_type(value).scalar_width() == Some(8) {
+            fun.to_msl_64_bit()?
+        } else {
+            fun.to_msl()
+        };
         write!(self.out, ".atomic_{}(", op)?;
         // coordinates in IR are int, but Metal expects uint
         self.put_cast_to_uint_scalar_or_vector(address.coordinate, &context.expression)?;
--- a/naga/src/back/spv/image.rs
+++ b/naga/src/back/spv/image.rs
@@ -1253,6 +1253,10 @@ impl BlockContext<'_> {
             class: spirv::StorageClass::Image,
         }));
         let signed = scalar.kind == crate::ScalarKind::Sint;
+        if scalar.width == 8 {
+            self.writer
+                .require_any("64 bit image atomics", &[spirv::Capability::Int64Atomics])?;
+        }
         let pointer_id = self.gen_id();
         let coordinates = self.write_image_coordinates(coordinate, array_index, block)?;
         let sample_id = self.writer.get_constant_scalar(crate::Literal::U32(0));
--- a/naga/src/valid/function.rs
+++ b/naga/src/valid/function.rs
@@ -1282,6 +1282,34 @@ impl super::Validator {
                                         .with_span_handle(image, context.expressions));
                                     }
                                     match format {
+                                        crate::StorageFormat::R64Uint => {
+                                            if !self.capabilities.intersects(
+                                                super::Capabilities::TEXTURE_INT64_ATOMIC,
+                                            ) {
+                                                return Err(FunctionError::MissingCapability(
+                                                    super::Capabilities::TEXTURE_INT64_ATOMIC,
+                                                )
+                                                .with_span_static(
+                                                    span,
+                                                    "missing capability for this operation",
+                                                ));
+                                            }
+                                            match fun {
+                                                crate::AtomicFunction::Min
+                                                | crate::AtomicFunction::Max => {}
+                                                _ => {
+                                                    return Err(
+                                                        FunctionError::InvalidImageAtomicFunction(
+                                                            fun,
+                                                        )
+                                                        .with_span_handle(
+                                                            image,
+                                                            context.expressions,
+                                                        ),
+                                                    );
+                                                }
+                                            }
+                                        }
                                         crate::StorageFormat::R32Sint
                                         | crate::StorageFormat::R32Uint => {
                                             if !self
--- a/naga/tests/in/atomicTexture-int64.wgsl
+++ b/naga/tests/in/atomicTexture-int64.wgsl
@@ -1,0 +1,12 @@
+@group(0) @binding(0)
+var image: texture_storage_2d<r64uint, atomic>;
+
+@compute
+@workgroup_size(2)
+fn cs_main(@builtin(local_invocation_id) id: vec3<u32>) {
+    textureAtomicMax(image, vec2<i32>(0, 0), 1lu);
+
+    workgroupBarrier();
+
+    textureAtomicMin(image, vec2<i32>(0, 0), 1lu);
+}
--- a/naga/tests/out/hlsl/atomicTexture-int64.hlsl
+++ b/naga/tests/out/hlsl/atomicTexture-int64.hlsl
@@ -1,0 +1,17 @@
+struct NagaConstants {
+    int first_vertex;
+    int first_instance;
+    uint other;
+};
+ConstantBuffer<NagaConstants> _NagaConstants: register(b0, space1);
+
+RWTexture2D<uint64_t> image : register(u0);
+
+[numthreads(2, 1, 1)]
+void cs_main(uint3 id : SV_GroupThreadID)
+{
+    InterlockedMax(image[int2(0, 0)],1uL);
+    GroupMemoryBarrierWithGroupSync();
+    InterlockedMin(image[int2(0, 0)],1uL);
+    return;
+}
--- a/naga/tests/out/msl/atomicTexture-int64.msl
+++ b/naga/tests/out/msl/atomicTexture-int64.msl
@@ -1,0 +1,18 @@
+// language: metal3.1
+#include <metal_stdlib>
+#include <simd/simd.h>
+
+using metal::uint;
+
+
+struct cs_mainInput {
+};
+kernel void cs_main(
+  metal::uint3 id [[thread_position_in_threadgroup]]
+, metal::texture2d<ulong, metal::access::read_write> image [[user(fake0)]]
+) {
+    image.atomic_max(metal::uint2(metal::int2(0, 0)), 1uL);
+    metal::threadgroup_barrier(metal::mem_flags::mem_threadgroup);
+    image.atomic_min(metal::uint2(metal::int2(0, 0)), 1uL);
+    return;
+}
--- a/tests/tests/image_atomics/mod.rs
+++ b/tests/tests/image_atomics/mod.rs
@@ -6,17 +6,45 @@ use wgpu_test::{
 };
 
 #[gpu_test]
-static IMAGE_32_ATOMICS: GpuTestConfiguration = GpuTestConfiguration::new()
+static IMAGE_64_ATOMICS: GpuTestConfiguration = GpuTestConfiguration::new()
     .parameters(
         TestParameters::default()
-            .limits(wgt::Limits {
+            .limits(wgpu::Limits {
                 max_storage_textures_per_shader_stage: 1,
                 max_compute_invocations_per_workgroup: 64,
                 max_compute_workgroup_size_x: 4,
                 max_compute_workgroup_size_y: 4,
                 max_compute_workgroup_size_z: 4,
-                max_compute_workgroups_per_dimension: wgt::COPY_BYTES_PER_ROW_ALIGNMENT,
-                ..wgt::Limits::downlevel_webgl2_defaults()
+                max_compute_workgroups_per_dimension: wgpu::COPY_BYTES_PER_ROW_ALIGNMENT,
+                ..wgpu::Limits::downlevel_webgl2_defaults()
+            })
+            .features(
+                wgpu::Features::TEXTURE_ATOMIC
+                    | wgpu::Features::TEXTURE_INT64_ATOMIC
+                    | wgpu::Features::SHADER_INT64,
+            ),
+    )
+    .run_async(|ctx| async move {
+        test_format(
+            ctx,
+            wgpu::TextureFormat::R64Uint,
+            wgpu::include_wgsl!("image_64_atomics.wgsl"),
+        )
+        .await;
+    });
+
+#[gpu_test]
+static IMAGE_32_ATOMICS: GpuTestConfiguration = GpuTestConfiguration::new()
+    .parameters(
+        TestParameters::default()
+            .limits(wgpu::Limits {
+                max_storage_textures_per_shader_stage: 1,
+                max_compute_invocations_per_workgroup: 64,
+                max_compute_workgroup_size_x: 4,
+                max_compute_workgroup_size_y: 4,
+                max_compute_workgroup_size_z: 4,
+                max_compute_workgroups_per_dimension: wgpu::COPY_BYTES_PER_ROW_ALIGNMENT,
+                ..wgpu::Limits::downlevel_webgl2_defaults()
             })
             .features(wgpu::Features::TEXTURE_ATOMIC),
     )
@@ -36,8 +64,8 @@ async fn test_format(
 ) {
     let pixel_bytes = format.target_pixel_byte_cost().unwrap();
     let size = wgpu::Extent3d {
-        width: wgt::COPY_BYTES_PER_ROW_ALIGNMENT,
-        height: wgt::COPY_BYTES_PER_ROW_ALIGNMENT,
+        width: wgpu::COPY_BYTES_PER_ROW_ALIGNMENT,
+        height: wgpu::COPY_BYTES_PER_ROW_ALIGNMENT,
         depth_or_array_layers: 1,
     };
     let bind_group_layout_entry = wgpu::BindGroupLayoutEntry {
--- a/wgpu-hal/src/dx12/adapter.rs
+++ b/wgpu-hal/src/dx12/adapter.rs
@@ -390,6 +390,13 @@ impl super::Adapter {
         );
 
         features.set(
+            wgt::Features::TEXTURE_INT64_ATOMIC,
+            shader_model >= naga::back::hlsl::ShaderModel::V6_6
+                && hr.is_ok()
+                && features1.Int64ShaderOps.as_bool(),
+        );
+
+        features.set(
             wgt::Features::SUBGROUP,
             shader_model >= naga::back::hlsl::ShaderModel::V6_0
                 && hr.is_ok()
--- a/wgpu-hal/src/gles/conv.rs
+++ b/wgpu-hal/src/gles/conv.rs
@@ -50,6 +50,7 @@ impl super::AdapterShared {
                 glow::RGB,
                 glow::UNSIGNED_INT_10F_11F_11F_REV,
             ),
+            Tf::R64Uint => (glow::RG32UI, glow::RED_INTEGER, glow::UNSIGNED_INT),
             Tf::Rg32Uint => (glow::RG32UI, glow::RG_INTEGER, glow::UNSIGNED_INT),
             Tf::Rg32Sint => (glow::RG32I, glow::RG_INTEGER, glow::INT),
             Tf::Rg32Float => (glow::RG32F, glow::RG, glow::FLOAT),
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -406,7 +406,7 @@ bitflags::bitflags! {
         /// This is a web and native feature.
         const FLOAT32_FILTERABLE = 1 << 11;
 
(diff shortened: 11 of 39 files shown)


naga/src/back/glsl/mod.rs
```diff
@@ -4081,6 +4092,56 @@ impl<'a, W: Write> Writer<'a, W> {
...
+
+        // This will only panic if the module is invalid
+        let dim = match *ctx.resolve_type(image, &self.module.types) {
+            TypeInner::Image { dim, .. } => dim,
+            _ => unreachable!(),
+        };
+
+        // Begin our call to `imageStore`
+        let fun_str = fun.to_glsl();
+        write!(self.out, "imageAtomic{fun_str}(")?;
```
atlv24: https://docs.gl/sl4/imageAtomicMax

naga/src/back/hlsl/writer.rs
```diff
@@ -2207,6 +2207,35 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
+                image,
+                coordinate,
+                array_index,
+                fun,
+                value,
+            } => {
+                write!(self.out, "{level}")?;
+
+                let fun_str = fun.to_hlsl_suffix();
+                write!(self.out, "Interlocked{fun_str}(")?;
```
atlv24: https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/interlockedmax

JMS55: I think this may need to be suffixed by 64 like in https://github.com/JMS55/wgpu/blob/a391e14e5114c3ada42924408c60f725b6d9bcdc/naga/src/back/hlsl/writer.rs#L2020

naga/src/back/msl/writer.rs
```diff
@@ -1186,6 +1186,31 @@ impl<W: Write> Writer<W> {
...
+        level: back::Level,
+        image: Handle<crate::Expression>,
+        address: &TexelAddress,
+        fun: crate::AtomicFunction,
+        value: Handle<crate::Expression>,
+        context: &StatementContext,
+    ) -> BackendResult {
+        write!(self.out, "{level}")?;
+        self.put_expression(image, &context.expression, false)?;
+        write!(self.out, ".write(")?;
```
atlv24: https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf page 251

naga/src/back/spv/image.rs
```diff
@@ -1188,4 +1188,56 @@ impl<'w> BlockContext<'w> {
...
+        block: &mut Block,
+    ) -> Result<(), Error> {
+        let pointer_type_id = self.get_expression_type_id(&crate::proc::TypeResolution::Value(
+            crate::TypeInner::ValuePointer {
+                space: crate::AddressSpace::Uniform,
+                size: None,
+                scalar: crate::Scalar {
+                    kind: crate::ScalarKind::Uint,
+                    width: 8,
+                },
```
atlv24: this shouldn't be hardcoded

naga/src/front/wgsl/lower/mod.rs
```diff
@@ -2347,6 +2347,42 @@ impl<'source, 'temp> Lowerer<'source, 'temp> {
...
+                            let image = args.next()?;
+                            let image = self.expression(image, ctx)?;
+
+                            let coordinate = self.expression(args.next()?, ctx)?;
+
+                            let ty = ctx
+                                .ensure_type_exists(crate::TypeInner::Scalar(crate::Scalar::I32));
+                            // We fib in a zero value for sample because it is not supported
+                            let sample =
+                                ctx.append_expression(crate::Expression::ZeroValue(ty), span)?;
```
atlv24: the reason i include it at all is because I dont know how to create a ZeroValue expression later on at the write stage

naga/tests/out/hlsl/atomicTexture-int64.hlsl
```diff
@@ -0,0 +1,21 @@
...
+RWTexture2D<uint> image : register(u0);
+
+int ZeroValueint() {
+    return (int)0;
+}
+
+[numthreads(2, 1, 1)]
+void cs_main(uint3 id : SV_GroupThreadID)
+{
+    image.InterlockedMax64(int2(0, 0),1uL);
```
atlv24: error: no member named 'InterlockedMax64' in 'RWTexture2D<unsigned int>'
im not sure how to get a `RWByteAddressBuffer` from a 'RWTexture2D<unsigned int>'
https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_Int64_and_Float_Atomics.html#interlockedmax

naga/tests/out/msl/atomicTexture-int64.msl
```diff
@@ -0,0 +1,18 @@
...
+using metal::uint;
+
+
+struct cs_mainInput {
+};
+kernel void cs_main(
+  metal::uint3 id [[thread_position_in_threadgroup]]
+, metal::texture2d<uint, metal::access::write> image [[user(fake0)]]
+) {
+    atomic_max_explicit(&image[metal::uint2(metal::int2(0, 0))], 1uL, metal::memory_order_relaxed);
```
atlv24: error: type 'metal::texture2d<uint, metal::access::write>' (aka 'texture2d<unsigned int, metal::access::write>') does not provide a subscript operator
idk how to get a pointer to a texel on metal, and https://github.com/KhronosGroup/SPIRV-Cross/issues/835 does not inspire hope

naga/tests/out/hlsl/atomicTexture-int64.hlsl
JMS55: You don't, you use RWTexture2D like so: https://github.com/microsoft/DirectXShaderCompiler/blob/75ff50caa046a054747ae15b5c1910a4c8aa1917/tools/clang/test/HLSLFileCheck/hlsl/intrinsics/atomic/atomic.hlsl#L58

naga/tests/out/msl/atomicTexture-int64.msl
JMS55: Based on the spec, I believe it should be `image.atomic_max(coord, value);`
image

atlv24: i get now
```
error: no member named 'atomic_max' in 'metal::texture2d<unsigned int, metal::access::write>'
    image.atomic_max(metal::uint2(metal::int2(0, 0)), 1uL, metal::memory_order_relaxed);
    ~~~~~ ^
```

i think 64 bit atomics have different syntax than the 32 bit, https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf page 251
{5691D177-355B-40C8-927A-CFDA2C4B6300}

naga/tests/out/hlsl/atomicTexture-int64.hlsl
atlv24: This indeed fixed the hlsl!

naga/tests/out/msl/atomicTexture-int64.msl
JMS55: Those are for buffers I believe, not textures. Notice how it says Metal 2.4 and not 3.1.

You want atomic_max based on this section:
image

atlv24: im following this to the letter and still erroring
```
error: no member named 'atomic_max' in 'metal::texture2d<unsigned long, metal::access::write>'
    image.atomic_max(metal::uint2(metal::int2(0, 0)), 1uL);
    ~~~~~ ^
```

wgpu-hal/src/metal/conv.rs
```diff
@@ -27,6 +27,11 @@ pub fn map_texture_usage(
         format.is_combined_depth_stencil_format(),
     );
 
+    mtl_usage.set(
+        metal::MTLTextureUsage::ShaderAtomic,
+        format == wgt::TextureFormat::R64Uint,
```
atlv24: There should be a better check here

naga/src/back/spv/image.rs
```diff
@@ -1188,4 +1188,74 @@ impl<'w> BlockContext<'w> {
...
+        coordinate: Handle<crate::Expression>,
+        sample: Handle<crate::Expression>,
+        fun: crate::AtomicFunction,
+        value: Handle<crate::Expression>,
+        block: &mut Block,
+    ) -> Result<(), Error> {
+        let image_id = match self.ir_function.expressions[image] {
+            crate::Expression::GlobalVariable(handle) => {
+                self.writer.global_variables[handle].var_id
+            }
```
atlv24: what happens if its a FunctionArgument or Expression::Access?

naga/src/valid/function.rs
```diff
@@ -1139,6 +1140,106 @@ impl super::Validator {
...
+                        _ => {
+                            return Err(FunctionError::InvalidImageAtomic(
+                                ExpressionError::ExpectedImageType(var.ty),
+                            )
+                            .with_span()
+                            .with_handle(var.ty, context.types)
+                            .with_handle(image, context.expressions))
+                        }
+                    };
+
```
atlv24: Needs read_write validation on the texture storage access

naga/tests/in/atomicTexture-int64.wgsl
```diff
@@ -0,0 +1,12 @@
...
+var image: texture_storage_2d<r64uint, read_write>;
+
+@compute
+@workgroup_size(2)
+fn cs_main(@builtin(local_invocation_id) id: vec3<u32>) {
+    imageAtomicMax(image, vec2<i32>(0, 0), 1lu);
+
+    workgroupBarrier();
+
+    imageAtomicMin(image, vec2<i32>(0, 0), 1lu);
```
JMS55: What about non-atomic reads? Does it work, especially on platforms where it's emulated as Rg32Uint?

atlv24: haven't tested but it should, i think. good call though, i'll make sure to include tests for that

wgpu-hal/src/metal/conv.rs
atlv24: Im not sure what the best way to go about this would be. hal::TextureUses comes from wgt::TextureUsages (interesting that these are inconsistently named) which comes from webgpu spec apparently, which makes no accommodation for specifying atomic image usage.

naga/tests/in/atomicTexture-int64.wgsl
atlv24: seems to work... maybe?

wgpu-hal/src/metal/conv.rs
jimblandy: I think it's cleanest and best to just go ahead and add appropriate bits to `wgpu_types::TextureUsages` and then `wgpu_hal::TextureUses`. As long as they are clearly documented as extensions, with a reference to which `wgpu_types::Features` bit you need to turn on in order to use them, I think we're good.

Since these are the first extension bits we'd be adding to `wgpu_types::TextureUsages`, I'd suggest following the example of `wgpu_types::Features` which blocks off standard and extension portions:

```rust
        //
        // ---- Start numbering at 1 << 0 ----
        //
        // WebGPU features:
        //
...

        //
        // ---- Restart Numbering for Native Features ---
        //
        // Native Features:
        //
```

(I don't think that layout succeeds in making the section breaks stand out visually, but, whatever.)

It might also be good to start assigning bits after a gap. At the moment, `TextureUsages` only goes up to `1 << 4`, but it's reasonable to assume that the WebGPU standard will add more in the future. I mean, we don't promise any kind of binary interface stability, but it's still probably nice to avoid having our bits overlap with the standard's bits. So maybe `const SHADER_ATOMIC = 1 << 16;` or something.

(Although `wgpu_types::Features` does this too, it doesn't really matter, because the WebGPU spec never assigns bits to features, it only ever talks about them in terms of strings.)

tests/tests/image_atomics/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+            .create_command_encoder(&wgpu::CommandEncoderDescriptor::default());
+        let mut rpass = encoder.begin_compute_pass(&wgpu::ComputePassDescriptor {
+            label: None,
+            timestamp_writes: None,
+        });
+        rpass.set_pipeline(&pipeline);
+        rpass.set_bind_group(0, Some(&bind_group), &[]);
+        rpass.dispatch_workgroups(1, 1, 1);
+        drop(rpass);
+        ctx.queue.submit(Some(encoder.finish()));
```
jimblandy: This test should try to read back the data from the the texture and see that the contents are as expected.

I think the current shader doesn't leave the texture in a predictable state; I've made some suggestions on the `.wgsl` file that should change that, so you can anticipate exactly what value you'll get back.

tests/tests/image_atomics/image_atomics.wgsl
```diff
@@ -0,0 +1,14 @@
+@group(0) @binding(0)
+var image: texture_storage_2d<r64uint, read_write>;
+
+@compute
+@workgroup_size(4, 4, 4)
+fn cs_main(@builtin(local_invocation_id) id: vec3<u32>) {
+    let l = u64(textureLoad(image, id.xy).r);
+
+    imageAtomicMin(image, id.xy, u64(id.z) + l);
```
jimblandy: Since these are all adding something >= 0, and the initial contents of the image are zero (right?), none of these `imageAtomicMin` operations are going to have any effect on the image, is that right? This makes it not a great test case, as the expected behavior is the same as if the code were just deleted.

I would suggest:

- Apply `imageAtomicMax` first, with `id.z + 10000`. This way, we can definitely expect every texel to end up with an `r` value of 10003.

- Apply `imageAtomicMin` next, with `id.z + 1000`. This way, we can definitely expect every texel to end up with an `r` value of 1000.

Further, we're only running 64 invocations here. At least on my GPU, a wavefront is 32 invocations. So you're only getting two wavefronts executing, which seems like we're not exercising the inter-execution-unit wide-area behavior that we care about. You've got a 256x256 texture, so go ahead and run it across the whole texture.

wgpu-hal/src/metal/conv.rs
atlv24: thanks, that makes sense. done

jimblandy (dismissed):

jimblandy:
@atlv24 ﻿Earlier today you asked me for a suggestion about the texture usages, and a review of the tests. I've posted those now. Is there anything else you need before breaking this up into independent commits for review?

atlv24:
@jimblandy I've broken it into independent commits already, thanks for the feedback. I've implemented the texture usage stuff, now im working on the tests. :D

wgpu-hal/src/dx12/adapter.rs
```diff
@@ -402,6 +402,8 @@ impl super::Adapter {
             wgt::Features::SHADER_INT64_ATOMIC_ALL_OPS | wgt::Features::SHADER_INT64_ATOMIC_MIN_MAX,
             atomic_int64_on_typed_resource_supported,
         );
+        // Ruint64 textures are always emulated on d3d12
+        features.set(wgt::Features::TEXTURE_INT64_ATOMIC, true);
```
cwfitzgerald: This should be at the beginning list of always enabled features.

wgpu/src/backend/webgpu/webgpu_sys/gen_GpuTextureFormat.rs
```diff
@@ -60,6 +60,7 @@ pub enum GpuTextureFormat {
     Rgb10a2uint = "rgb10a2uint",
     Rgb10a2unorm = "rgb10a2unorm",
     Rg11b10ufloat = "rg11b10ufloat",
+    R64uint = "r64uint",
```
cwfitzgerald: This file really shouldn't be edited

cwfitzgerald (changes requested):
Looking at the wgpu side:

What would prevent you from doing atomics on an image without the image atomics flag set?

atlv24:
> What would prevent you from doing atomics on an image without the image atomics flag set?

The naga validator currently only permits image atomic operations on the r64uint texture format, and only when the feature is available. No provisions were made for other texture formats yet

cwfitzgerald:
I mean the texture creation flag - I believe if you don't include the flag on texture creation it'll still you do atomics on it, as those pieces of data never touch each other.

atlv24:
@cwfitzgerald i have split this into two PRs: #6706 adds image atomics without any of the 64 bit stuff, this one adds the 64 bit stuff. Only the second commit on this PR is relevant to it, as this PR's branch is based off of the other's, so it includes its commits.

cwfitzgerald:
Demoting to draft while dependent

tests/tests/image_atomics/mod.rs
```diff
@@ -5,6 +5,35 @@ use wgpu_test::{
...
+    .parameters(
+        TestParameters::default()
+            .limits(wgt::Limits {
+                max_storage_textures_per_shader_stage: 1,
+                max_compute_invocations_per_workgroup: 64,
+                max_compute_workgroup_size_x: 4,
+                max_compute_workgroup_size_y: 4,
+                max_compute_workgroup_size_z: 4,
+                max_compute_workgroups_per_dimension: wgt::COPY_BYTES_PER_ROW_ALIGNMENT,
+                ..wgt::Limits::downlevel_webgl2_defaults()
```
cwfitzgerald:
```suggestion
                max_compute_workgroups_per_dimension: wgpu::COPY_BYTES_PER_ROW_ALIGNMENT,
                ..wgpu::Limits::downlevel_webgl2_defaults()
```

Nit

wgpu-hal/src/gles/conv.rs
```diff
@@ -50,6 +50,7 @@ impl super::AdapterShared {
                 glow::RGB,
                 glow::UNSIGNED_INT_10F_11F_11F_REV,
             ),
+            Tf::R64Uint => (glow::RG32UI, glow::RED_INTEGER, glow::UNSIGNED_INT), //TODO?
```
cwfitzgerald: Todo?

wgpu-types/src/lib.rs
```diff
@@ -988,6 +988,16 @@ bitflags::bitflags! {
         ///
         /// [VK_KHR_external_memory_win32]: https://registry.khronos.org/vulkan/specs/latest/man/html/VK_KHR_external_memory_win32.html
         const VULKAN_EXTERNAL_MEMORY_WIN32 = 1 << 63;
+
+        /// Enables R64Uint image atomic min and max.
+        ///
+        /// Supported platforms:
+        /// - Vulkan (with VK_EXT_shader_image_atomic_int64)
+        /// - DX12 (with SM 6.6+ emulated via Rg32Uint texture)
+        /// - Metal (with MSL 3.1+ emulated via RG32Uint texture)
```
cwfitzgerald: I wouldn't call this emulated

tests/tests/image_atomics/mod.rs
```diff
@@ -5,6 +5,35 @@ use wgpu_test::{
     fail, gpu_test, image::ReadbackBuffers, GpuTestConfiguration, TestParameters, TestingContext,
 };
 
+#[gpu_test]
+static IMAGE_64_ATOMICS: GpuTestConfiguration = GpuTestConfiguration::new()
```
cwfitzgerald, line 9: Just double checking, we've tested this on all platforms right?

cwfitzgerald (changes requested):

wgpu-types/src/lib.rs
atlv24: what would you call it?

wgpu-hal/src/gles/conv.rs
atlv24: not sure what glow is or what this should be

tests/tests/image_atomics/mod.rs
atlv24: it works on mac windows and linux, haven't tried anything else

wgpu-hal/src/gles/conv.rs
JMS55: OpenGL

wgpu-types/src/lib.rs
cwfitzgerald:
```suggestion
        /// - DX12 (with SM 6.6+)
        /// - Metal (with MSL 3.1+)
```

Just completely omit it - it's not exposed to the user in any way.

cwfitzgerald:
Ran this locally, got this failure on all devices

```
--- STDERR:              wgpu-test::wgpu-test [Executed] [Dx12/NVIDIA GeForce RTX 4070/2] wgpu_test::image_atomics::image_64_atomics ---
[2025-01-14T03:59:58Z ERROR wgpu::backend::wgpu_core] Handling wgpu errors as fatal by default
thread '<unnamed>' panicked at wgpu\src\backend\wgpu_core.rs:1423:26:
wgpu error: Validation Error

Caused by:
  In Device::create_texture
    Texture usages TextureUsages(STORAGE_ATOMIC) are not allowed on a texture of type R64Uint due to downlevel restrictions


[2025-01-14T03:59:58Z ERROR wgpu_test::expectations] Panic: wgpu error: Validation Error

    Caused by:
      In Device::create_texture
        Texture usages TextureUsages(STORAGE_ATOMIC) are not allowed on a texture of type R64Uint due to downlevel restrictions


thread '<unnamed>' panicked at tests\src\run.rs:120:9:
tests\tests\image_atomics\mod.rs:9:49: test "wgpu_test::image_atomics::image_64_atomics" did not behave as expected
```

This does not appear in CI as image atomics is not supported on any device, in part because of https://github.com/gfx-rs/wgpu/issues/6911

cwfitzgerald (approved):
Seems alright to me!

Whatever you did fixed the problem locally

teoxoy (approved):
LGTM

Issue #5443, opened 2024-03-27
wgpu needs benchmarks


jimblandy:
We know that we need to make some key architectural changes to wgpu's resource management, but wgpu's performance is a key selling point, and it's difficult to assess the impact of the changes we're considering without having benchmarks at hand.

The ideal would be something in CI, but we're sort of at zero right now, and we need some way to answer engineering questions. The benchmarks could be informal, require manual steps, be microbenchmarks, require use of external tools like Tracy - all would be a step forward.

JMS55:
To add to this, (iirc) when profiling Bevy we've seen wgpu logging having a significant amount of overhead, even when logging is not enabled. I wanted to look into profiling this and see if I could replicate and fix it, but it's difficult to hook Bevy up to newer wgpu versions, and I ended up working on something else. Having builtin benchmarks would be very useful.

lylythechosenone:
We could start by just including the tracing crate and putting its attribute on functions for some simple profiling.

Ideally we can get some proper benchmarks with criterion eventually. We can probably pretty easily port some of WebGPU's official benchmarks.

FL33TW00D:
@lylythechosenone For reference, I've hooked up `wgpu` to `criterion` here: https://github.com/FL33TW00D/wgpu-bench/blob/master/src/lib.rs#L165

PR #5539, opened 2024-04-15
[core] Enforce a deadlock-free locking order for Mutexes.

jimblandy:
If `debug_assertions` or the `"validate-locks"` feature are enabled, change `wgpu-core` to use a wrapper around `parking_lot::Mutex` that checks for potential deadlocks.

### DIFF
--- a/wgpu-core/src/lock/mod.rs
+++ b/wgpu-core/src/lock/mod.rs
@@ -1,0 +1,41 @@
+//! Instrumented lock types.
+//!
+//! This module defines a set of instrumented wrappers for the lock
+//! types used in `wgpu-core` ([`Mutex`], [`RwLock`], and
+//! [`SnatchLock`]) that help us understand and validate `wgpu-core`
+//! synchronization.
+//!
+//! - The [`ranked`] module defines lock types that perform run-time
+//!   checks to ensure that each thread acquires locks only in a
+//!   specific order, to prevent deadlocks.
+//!
+//! - The [`vanilla`] module defines lock types that are
+//!   uninstrumented, no-overhead wrappers around the standard lock
+//!   types.
+//!
+//! (We plan to add more wrappers in the future.)
+//!
+//! If the `wgpu_validate_locks` config is set (for example, with
+//! `RUSTFLAGS='--cfg wgpu_validate_locks'`), `wgpu-core` uses the
+//! [`ranked`] module's locks. We hope to make this the default for
+//! debug builds soon.
+//!
+//! Otherwise, `wgpu-core` uses the [`vanilla`] module's locks.
+//!
+//! [`Mutex`]: parking_lot::Mutex
+//! [`RwLock`]: parking_lot::RwLock
+//! [`SnatchLock`]: crate::snatch::SnatchLock
+
+pub mod rank;
+
+#[cfg_attr(not(wgpu_validate_locks), allow(dead_code))]
+mod ranked;
+
+#[cfg_attr(wgpu_validate_locks, allow(dead_code))]
+mod vanilla;
+
+#[cfg(wgpu_validate_locks)]
+pub use ranked::{Mutex, MutexGuard};
+
+#[cfg(not(wgpu_validate_locks))]
+pub use vanilla::{Mutex, MutexGuard};
--- a/wgpu-core/src/lock/rank.rs
+++ b/wgpu-core/src/lock/rank.rs
@@ -1,0 +1,143 @@
+//! Ranks for `wgpu-core` locks, restricting acquisition order.
+//!
+//! See [`LockRank`].
+
+/// The rank of a lock.
+///
+/// Each [`Mutex`], [`RwLock`], and [`SnatchLock`] in `wgpu-core` has been
+/// assigned a *rank*: a node in the DAG defined at the bottom of
+/// `wgpu-core/src/lock/rank.rs`. The rank of the most recently
+/// acquired lock you are still holding determines which locks you may
+/// attempt to acquire next.
+///
+/// When you create a lock in `wgpu-core`, you must specify its rank
+/// by passing in a [`LockRank`] value. This module declares a
+/// pre-defined set of ranks to cover everything in `wgpu-core`, named
+/// after the type in which they occur, and the name of the type's
+/// field that is a lock. For example, [`CommandBuffer::data`] is a
+/// `Mutex`, and its rank here is the constant
+/// [`COMMAND_BUFFER_DATA`].
+///
+/// [`Mutex`]: parking_lot::Mutex
+/// [`RwLock`]: parking_lot::RwLock
+/// [`SnatchLock`]: crate::snatch::SnatchLock
+/// [`CommandBuffer::data`]: crate::command::CommandBuffer::data
+#[derive(Debug, Copy, Clone)]
+pub struct LockRank {
+    /// The bit representing this lock.
+    ///
+    /// There should only be a single bit set in this value.
+    pub(super) bit: LockRankSet,
+
+    /// A bitmask of permitted successor ranks.
+    ///
+    /// If `rank` is the rank of the most recently acquired lock we
+    /// are still holding, then `rank.followers` is the mask of
+    /// locks we are allowed to acquire next.
+    ///
+    /// The `define_lock_ranks!` macro ensures that there are no
+    /// cycles in the graph of lock ranks and their followers.
+    pub(super) followers: LockRankSet,
+}
+
+/// Define a set of lock ranks, and each rank's permitted successors.
+macro_rules! define_lock_ranks {
+    {
+        $(
+            $( #[ $attr:meta ] )*
+            rank $name:ident $member:literal followed by { $( $follower:ident ),* $(,)? };
+        )*
+    } => {
+        // An enum that assigns a unique number to each rank.
+        #[allow(non_camel_case_types, clippy::upper_case_acronyms)]
+        enum LockRankNumber { $( $name, )* }
+
+        bitflags::bitflags! {
+            #[derive(Debug, Copy, Clone, Eq, PartialEq)]
+            /// A bitflags type representing a set of lock ranks.
+            pub struct LockRankSet: u32 {
+                $(
+                    const $name = 1 << (LockRankNumber:: $name as u32);
+                )*
+            }
+        }
+
+        impl LockRankSet {
+            pub fn name(self) -> &'static str {
+                match self {
+                    $(
+                        LockRankSet:: $name => $member,
+                    )*
+                    _ => "<unrecognized LockRankSet bit>",
+                }
+            }
+        }
+
+        $(
+            // If there is any cycle in the ranking, the initializers
+            // for `followers` will be cyclic, and rustc will give us
+            // an error message explaining the cycle.
+            $( #[ $attr ] )*
+            pub const $name: LockRank = LockRank {
+                bit: LockRankSet:: $name,
+                followers: LockRankSet::empty() $( .union($follower.bit) )*,
+            };
+        )*
+    }
+}
+
+define_lock_ranks! {
+    rank COMMAND_BUFFER_DATA "CommandBuffer::data" followed by {
+        DEVICE_USAGE_SCOPES,
+        SHARED_TRACKER_INDEX_ALLOCATOR_INNER,
+        BUFFER_BIND_GROUP_STATE_BUFFERS,
+        TEXTURE_BIND_GROUP_STATE_TEXTURES,
+        BUFFER_MAP_STATE,
+        STATELESS_BIND_GROUP_STATE_RESOURCES,
+    };
+    rank STAGING_BUFFER_RAW "StagingBuffer::raw" followed by { };
+    rank COMMAND_ALLOCATOR_FREE_ENCODERS "CommandAllocator::free_encoders" followed by {
+        SHARED_TRACKER_INDEX_ALLOCATOR_INNER,
+    };
+    rank DEVICE_TRACKERS "Device::trackers" followed by { };
+    rank DEVICE_LIFE_TRACKER "Device::life_tracker" followed by {
+        COMMAND_ALLOCATOR_FREE_ENCODERS,
+        // Uncomment this to see an interesting cycle.
+        // DEVICE_TEMP_SUSPECTED,
+    };
+    rank DEVICE_TEMP_SUSPECTED "Device::temp_suspected" followed by {
+        SHARED_TRACKER_INDEX_ALLOCATOR_INNER,
+        COMMAND_BUFFER_DATA,
+        DEVICE_TRACKERS,
+    };
+    rank DEVICE_PENDING_WRITES "Device::pending_writes" followed by {
+        COMMAND_ALLOCATOR_FREE_ENCODERS,
+        SHARED_TRACKER_INDEX_ALLOCATOR_INNER,
+        DEVICE_LIFE_TRACKER,
+    };
+    rank DEVICE_DEFERRED_DESTROY "Device::deferred_destroy" followed by { };
+    #[allow(dead_code)]
+    rank DEVICE_TRACE "Device::trace" followed by { };
+    rank DEVICE_USAGE_SCOPES "Device::usage_scopes" followed by { };
+    rank BUFFER_SYNC_MAPPED_WRITES "Buffer::sync_mapped_writes" followed by { };
+    rank BUFFER_MAP_STATE "Buffer::map_state" followed by { DEVICE_PENDING_WRITES };
+    rank BUFFER_BIND_GROUPS "Buffer::bind_groups" followed by { };
+    rank TEXTURE_VIEWS "Texture::views" followed by { };
+    rank TEXTURE_BIND_GROUPS "Texture::bind_groups" followed by { };
+    rank IDENTITY_MANAGER_VALUES "IdentityManager::values" followed by { };
+    rank RESOURCE_POOL_INNER "ResourcePool::inner" followed by { };
+    rank BUFFER_BIND_GROUP_STATE_BUFFERS "BufferBindGroupState::buffers" followed by { };
+    rank STATELESS_BIND_GROUP_STATE_RESOURCES "StatelessBindGroupState::resources" followed by { };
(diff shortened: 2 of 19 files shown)


jimblandy:
This has already found a cycle, although it might take a lot of threads to hit it. If you uncomment out the indicated code in `wgpu-core/src/lock/rank.rs`, then you'll get this error mesage:

```
error[E0391]: cycle detected when const checking `lock::rank::COMMAND_BUFFER_DATA`
  --> wgpu-core/src/lock/rank.rs:28:35
   |
28 |                   followers: 0 $( | $follower.bit )*,
   |                                     ^^^^^^^^^
...
39 | / define_lock_ranks! {
40 | |     rank COMMAND_BUFFER_DATA 0 followed by {
41 | |         DEVICE_USAGE_SCOPES,
42 | |         SHARED_TRACKER_INDEX_ALLOCATOR_INNER,
...  |
83 | |     rank SURFACE_PRESENTATION 22 followed by { };
84 | | }
   | |_- in this macro invocation
   |
note: ...which requires const checking `lock::rank::BUFFER_MAP_STATE`...
  --> wgpu-core/src/lock/rank.rs:28:35
   |
28 |                   followers: 0 $( | $follower.bit )*,
   |                                     ^^^^^^^^^
...
39 | / define_lock_ranks! {
40 | |     rank COMMAND_BUFFER_DATA 0 followed by {
41 | |         DEVICE_USAGE_SCOPES,
42 | |         SHARED_TRACKER_INDEX_ALLOCATOR_INNER,
...  |
83 | |     rank SURFACE_PRESENTATION 22 followed by { };
84 | | }
   | |_- in this macro invocation
note: ...which requires const checking `lock::rank::DEVICE_PENDING_WRITES`...
  --> wgpu-core/src/lock/rank.rs:28:35
   |
28 |                   followers: 0 $( | $follower.bit )*,
   |                                     ^^^^^^^^^
...
39 | / define_lock_ranks! {
40 | |     rank COMMAND_BUFFER_DATA 0 followed by {
41 | |         DEVICE_USAGE_SCOPES,
42 | |         SHARED_TRACKER_INDEX_ALLOCATOR_INNER,
...  |
83 | |     rank SURFACE_PRESENTATION 22 followed by { };
84 | | }
   | |_- in this macro invocation
note: ...which requires const checking `lock::rank::DEVICE_LIFE_TRACKER`...
  --> wgpu-core/src/lock/rank.rs:28:35
   |
28 |                   followers: 0 $( | $follower.bit )*,
   |                                     ^^^^^^^^^
...
39 | / define_lock_ranks! {
40 | |     rank COMMAND_BUFFER_DATA 0 followed by {
41 | |         DEVICE_USAGE_SCOPES,
42 | |         SHARED_TRACKER_INDEX_ALLOCATOR_INNER,
...  |
83 | |     rank SURFACE_PRESENTATION 22 followed by { };
84 | | }
   | |_- in this macro invocation
note: ...which requires const checking `lock::rank::DEVICE_TEMP_SUSPECTED`...
  --> wgpu-core/src/lock/rank.rs:28:35
   |
28 |                   followers: 0 $( | $follower.bit )*,
   |                                     ^^^^^^^^^
...
39 | / define_lock_ranks! {
40 | |     rank COMMAND_BUFFER_DATA 0 followed by {
41 | |         DEVICE_USAGE_SCOPES,
42 | |         SHARED_TRACKER_INDEX_ALLOCATOR_INNER,
...  |
83 | |     rank SURFACE_PRESENTATION 22 followed by { };
84 | | }
   | |_- in this macro invocation
   = note: ...which again requires const checking `lock::rank::COMMAND_BUFFER_DATA`, completing the cycle
   = note: cycle used when running analysis passes on this crate
   = note: see https://rustc-dev-guide.rust-lang.org/overview.html#queries and https://rustc-dev-guide.rust-lang.org/query.html for more information
   = note: this error originates in the macro `define_lock_ranks` (in Nightly builds, run with -Z macro-backtrace for more info)
```

ErichDonGubler:
I'll take on review for this one. I've already done quite a bit of review with Jim as he's explained how the locking mechanism works. We'll both work to polish this up, and document and test properly. 🫡

wgpu-core/src/lock/validating.rs
```diff
@@ -0,0 +1,91 @@
+/*! Methods that validate lock acquisition order. */
+#![allow(dead_code)]
+
+use super::LockRank;
+use std::{cell::Cell, panic::Location};
+
+pub struct Mutex<T> {
+    inner: parking_lot::Mutex<T>,
+    rank: LockRank,
+}
```
ErichDonGubler: - [x] issue: We should add some documentation to this symbol. At the very least, we should point folks towards this and the non-validating implementation being selected based on Cargo features, preferably by linking to the details you've already written elsewhere. Maybe something like this:

```suggestion
/// Mutual exclusion lock variant for resources in core which validates lock ordering. See also [`super`].
pub struct Mutex<T> {
    inner: parking_lot::Mutex<T>,
    rank: LockRank,
}
```

wgpu-core/src/lock/rank.rs
```diff
@@ -0,0 +1,84 @@
...
+    rank TEXTURE_VIEWS 14 followed by { };
+    rank TEXTURE_BIND_GROUPS 15 followed by { };
+    rank IDENTITY_MANAGER_VALUES 16 followed by { };
+    rank RESOURCE_POOL_INNER 17 followed by { };
+    rank BUFFER_BIND_GROUP_STATE_BUFFERS 18 followed by { };
+    rank STATELESS_BIND_GROUP_STATE_RESOURCES 19 followed by { };
+    rank TEXTURE_BIND_GROUP_STATE_TEXTURES 20 followed by { };
+    rank SHARED_TRACKER_INDEX_ALLOCATOR_INNER 21 followed by { };
+    rank SURFACE_PRESENTATION 22 followed by { };
+}
```
ErichDonGubler: nitpick: RE: current commit contents ([0f1f20bbd55fb0f87d08b3d41f848cc2f9a392c8](https://github.com/gfx-rs/wgpu/commit/0f1f20bbd55fb0f87d08b3d41f848cc2f9a392c8)) and this DAG: You noted in first round of review on-call that writing this DAG was empirically driven. We should at least include that in the commit message (but maybe not a comment, since that seems both prone to being outdated and unimportant for actually working with it).

wgpu-core/src/lock/validating.rs
```diff
@@ -0,0 +1,91 @@
...
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.inner.deref_mut()
+    }
+}
+
+impl<T: std::fmt::Debug> std::fmt::Debug for Mutex<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.inner.fmt(f)
+    }
+}
```
ErichDonGubler: - [x] issue: I mention this elsewhere, but: we should add tests here. That can help folks wanting to understand this API much more easily, _and_ we should prove that we're actually preventing the deadlocks we say we are. Perhaps doc. tests would be appropriate for the most straightforward cases, with some `#[test]` functions for others?

    The only complication I can see here is that using something other than the current lock ranking requires an interchangeability of some kind, either via trait or a different value that can be passed in (maybe as a `cfg(test)`-only constructor?).

wgpu-core/src/lock/rank.rs
ErichDonGubler, line 1: thought: RE: performance: I won't block this PR based on performance concerns, since we don't have any benchmarks. _But_, for the record, I expect that this would have been something that would have been nice to evaluate in terms of performance. I'll link to #5443 FTR.

wgpu-core/src/lock/mod.rs
```diff
@@ -0,0 +1,22 @@
+/*! Deadlock prevention via lock ordering.
+
+This module defines wrappers for the lock types used in `wgpu`
+(`Mutex`, `RwLock`, and `SnatchLock`) that use runtime checks to
+ensure that locks are always acquired in a consistent order, ensuring
+that deadlock is impossible. These checks are enabled only in debug
+builds, and when the `"validate-locks"` feature is enabled.
```
ErichDonGubler: nitpick: You don't need double-quotes inside the inline code block here. I understand that that's what somebody would literally write in a `features` array element in a `Cargo.toml`, but it's inconsistent with the styling we use at the top level of `wgpu` ([example](https://docs.rs/wgpu/0.19.3/wgpu/index.html)) and `wgpu-core` ([example](https://docs.rs/wgpu-core/0.19.3/wgpu_core/index.html)).

```suggestion
builds, and when the `validate-locks` feature is enabled.
```

wgpu-core/src/lock/mod.rs
```diff
@@ -0,0 +1,22 @@
+/*! Deadlock prevention via lock ordering.
+
+This module defines wrappers for the lock types used in `wgpu`
+(`Mutex`, `RwLock`, and `SnatchLock`) that use runtime checks to
```
ErichDonGubler: nitpick: Links help us keep valid names, _and_ they're convenient for curious users. We should use them here.

TODO: suggestion

wgpu-core/src/lock/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+ensure that locks are always acquired in a consistent order, ensuring
+that deadlock is impossible. These checks are enabled only in debug
+builds, and when the `"validate-locks"` feature is enabled.
+
+See the [`validating`] module for details on how/why the validation works.
+
+*/
```
ErichDonGubler: thought(non-blocking): You are the only person I know personally that prefers multi-line module commentary, as opposed to prepending `//!` to everything. 🤔

wgpu-core/src/lock/non_validating.rs
```diff
@@ -0,0 +1,37 @@
+/*! Methods that do not validate lock acquisition order. */
+#![allow(dead_code)]
```
ErichDonGubler: nitpick: It doesn't make sense to me to `allow(dead_code)` in a new API that only `wgpu-core` consumes. Is there something I'm missing?

jimblandy:
One wrinkle: This has already detected a deadlock, and I haven't pushed very far yet, so I'd guess there'd be more to come. We'll have to fix them before we can actually land this.

[Edit: no, we can just make it only enabled when the feature is turned on, leaving both debug and release builds unaffected. This means we're not getting test coverage for the instrumented lock types, but at least we can have it in tree, rather than sitting on a branch as an indefinitely long list of deadlocks gets addressed.]

wgpu-core/src/lock/validating.rs
```diff
@@ -0,0 +1,163 @@
...
+impl<'a, T> Drop for MutexGuard<'a, T> {
+    fn drop(&mut self) {
+        let prior = LOCK_STATE.replace(self.saved);
+
+        // Although Rust allows mutex guards to be dropped in any
+        // order, this analysis requires that locks be acquired and
+        // released in stack order: the next lock to be released must be
+        // the most recently acquired lock still held.
+        assert_eq!(prior.depth, self.saved.depth + 1,
+                   "Lock not released in stacking order");
```
ErichDonGubler: question: Is there a reason we can't assert before actually changing state? It's usually better to avoid invalid states entirely, but I'm not sure if it's reasonable to do here.

ErichDonGubler: - [x] ~~issue: We should test this case specifically!~~ EDIT: Invalidated.

wgpu-core/src/lock/validating.rs
ErichDonGubler: - [x] This feedback also applies to the `non_validating::Mutex` symbol, of course.

ErichDonGubler (commented):
Not gonna commit to a review outcome yet, since this isn't out of draft yet. :^)

wgpu-core/src/lock/non_validating.rs
ErichDonGubler: Ah, you're doing this because this might not be used when not validating. Is there even a reason to compile the modules when one is not selected? At the very least, it seems we can make this suppression use some meaningful `cfg` conditions. Perhaps in the parent module declaration we could use `#[cfg_attr(…, allow(dead_code))]` instead to keep the mirrored conditions closer together?

wgpu-core/src/lock/validating.rs
jimblandy: I agree in general, but checking afterwards fits better with the `Cell` API, because we can do a single access with `replace`. My thought was, once a thread has panicked, its state doesn't matter much.

jimblandy: A test would need to check the thread's state after a panic, which doesn't seem that interesting.

wgpu-core/src/lock/mod.rs
jimblandy: I guess I wished I'd used it for the `wgpu-hal` module docs, because it would have made it easier to paste into the `README.md` file. (Yes, I know there is tooling for that.)

I've changed this to use `\\!`.

wgpu-core/src/lock/ranked.rs
```diff
@@ -0,0 +1,180 @@
...
+//!   waiting to acquire a lock, that lock must be held by a different thread: if
+//!   you were allowed to acquire a lock you already hold, that would be a cycle in
+//!   the graph.
+//! 
+//! - Furthermore, because the graph has no cycles, as we work our way from each
+//!   thread to the thread it is blocked waiting for, we must eventually reach an
+//!   end point: there must be some thread that is able to acquire its next lock, or
+//!   that is about to release a lock.
+//! 
+//! Thus, the system as a whose is always able to make progress: it is free of
```
cwfitzgerald:
```suggestion
//! Thus, the system as a whole is always able to make progress: it is free of
```

cwfitzgerald (approved):
This is absolutely amazing stuff!

Have a docs nit, but I love the docs, they are _very_ clear

jimblandy:
@cwfitzgerald Thanks for the review! @ErichDonGubler and I agreed this should have unit tests, so I'll work on that tomorrow and then ask Erich to take another look.

cwfitzgerald:
Sounds good, could @ErichDonGubler just mark it as request changes so I don't accidentally merge something I shouldn't :)

ErichDonGubler (changes requested):
Just waiting on feedback I've already left. 😊

wgpu-core/src/lock/validating.rs
jimblandy: All `Mutex` and guard types now have docs, and there's an explanation of how we select an implementation.

wgpu-core/src/lock/rank.rs
jimblandy: I've filed #5572 to track filling the ranking out.

jimblandy: I've mentioned #5572 in the commit message.

wgpu-core/src/lock/validating.rs
jimblandy: All these have docs now.

wgpu-core/src/lock/validating.rs
jimblandy: Okay, I think I've got some decent tests now.

jimblandy:
@ErichDonGubler, I took most of your suggestions but you had one that I didn't want to take. I've explained and left it unresolved; let me know what you think.

wgpu-core/src/lock/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+//! [`ranked`] module's locks. We hope to make this the default for
+//! debug builds soon.
+//!
+//! Otherwise, `wgpu-core` uses the [`vanilla`] module's locks.
+//!
+//! [`Mutex`]: parking_lot::Mutex
+//! [`RwLock`]: parking_lot::RwLock
+//! [`SnatchLock`]: crate::snatch::SnatchLock
+
+pub mod rank;
```
ErichDonGubler, line 29: nitpick: We're still unconditionally using `allow(dead_code)` in the `rank` module. I think we should have a similar `#[cfg_attr(feature = "validate-locks", allow(dead_code))]` here instead?

wgpu-core/src/lock/ranked.rs
```diff
@@ -0,0 +1,266 @@
...
+//! [`lock::rank`]: crate::lock::rank
+
+#![allow(dead_code)]
+
+use super::rank::LockRank;
+use std::{cell::Cell, panic::Location};
+
+/// A `Mutex` instrumented for deadlock prevention.
+///
+/// This is just a wrapper around a `parking_lot::Mutex`, along with
```
ErichDonGubler: nitpick(non-blocking): Linking to `parking_lot::Mutex` plz? 😀

EDIT: This also applies to `MutexGuard`.

ErichDonGubler (approved):
Only nitpick feedback left. Approving to enable progress, since I trust @jimblandy to duly consider the feedback I've left.

wgpu-core/src/lock/mod.rs
jimblandy: Hmm, it's funny, it's only the fields of `LockRank` that aren't read. We certainly use the rank constants unconditionally in `wgpu-core`.

jimblandy: It turns out that... the `cfg_attr` conditions were swapped??? Anyway, now that they're fixed, there are no extraneous `allow`s in `wgpu-core/src/lock`.

jimblandy:
@ErichDonGubler Hmm. Plot twist.

1) Just today I remarked on Rust's wisdom in flagging the use of `feature = "foo"` in the code when there's no `"foo"` feature in `Cargo.toml`. This is completely untrue: rustc just treats them as if they're not enabled.

2) We want our CI to run tests with `--all-features`, which would enable `"validate-locks"`, which causes a panic in ordinary operation.

I'm not sure what the best way forward is.

- The easiest is to omit the feature from `Cargo.toml`. If people want to experiment it, they'll have to add it back themselves.

- We could just comment it out, but that has the disadvantage that it's not clear what you need to uncomment to make things work. The nice thing about putting `feature = "validate-locks"`, even without a `Cargo.toml` entry, is that its use marks everything that needs to be adjusted, and it's pretty obvious what to do if you want to try it out.

Other ideas??

jimblandy:
In chat we decided to go with `cfg(validate_locks)`.

Edit: Err, cfgs are compilation-wide, so actually it's `cfg(wgpu_validate_locks)`.

PR #5552, opened 2024-04-18
Backport changelog from 0.19.4

cwfitzgerald:


Wumpf (approved):
👍 
we should also do the crate version bump again

Issue #5255, opened 2024-02-15
Make `wgpu_hal::vulkan::CommandEncoder::discard_encoding` idempotent so we can discard on `Drop`


ErichDonGubler:
In <https://github.com/gfx-rs/wgpu/pull/5251>, we changed all but the Vulkan backend's `CommandEncoder` implementations to call `CommandEncoder::discard_encoding`. The reason we/I forewent changing the Vulkan backend is because it doesn't appear idempotent. We'd like this to change, though, to keep things safer via [the `Drop` trait](https://doc.rust-lang.org/stable/std/ops/trait.Drop.html)! To make a concrete list, we should:

1. [`wgpu_hal::vulkan::CommandEncoder::discard_encoding`](https://docs.rs/wgpu-hal/0.19.1/wgpu_hal/trait.CommandEncoder.html#tymethod.discard_encoding) idempotent, like with other backends.
1. Call [`wgpu_hal::CommandEncoder::discard_encoding`](https://docs.rs/wgpu-hal/0.19.1/wgpu_hal/trait.CommandEncoder.html#tymethod.discard_encoding) in [`wgpu_hal::vulkan::CommandEncoder`](https://docs.rs/wgpu-hal/0.19.1/wgpu_hal/trait.CommandEncoder.html#)'s [`Drop`](https://doc.rust-lang.org/stable/std/ops/trait.Drop.html) implementation (likely in `wgpu-hal/src/vulkan/command.rs`, close to the other `impl super::CommandEncoder` block).
1. Adjust documentation of the trait definition of [`wgpu_hal::CommandEncoder::discard_encoding`](https://docs.rs/wgpu-hal/0.19.1/wgpu_hal/trait.CommandEncoder.html#tymethod.discard_encoding) to _require_ idempotency, since this seems like an important property for _any_ backend.
1. Adjust the documentation of [`wgpu_hal::CommandEncoder`](https://docs.rs/wgpu-hal/0.19.1/wgpu_hal/trait.CommandEncoder.html#) to recommend that implementers also implement [`Drop`](https://doc.rust-lang.org/stable/std/ops/trait.Drop.html).

jimblandy:
[...]

PR #5557, opened 2024-04-18
Add check to ensure `vulkan::CommandEncoder::discard_encoding` is not called multiple times in a row

villuna:
**Connections**
Addresses https://github.com/gfx-rs/wgpu/issues/5255

**Description**
I just followed the (very helpful) todo list in Erich's issue, so the info is mostly there.

The vulkan implementation for `CommandEncoder` was previously pushing its active CommandBuffer handle to its discard vector every time the `discard_encoding` function was called, even if the handle was null. So I made it only do this if the active handle is non-null, so as to make the function idempotent.

**Testing**
No new tests but I ran the existing test suite on Vulkan (on an Nvidia GTX1050) and it ran the same as on the trunk branch.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo xtask test` to run tests - I got one error every now and then, but this also happened on the main branch so I think it's fine.

### DIFF
--- a/wgpu-hal/src/lib.rs
+++ b/wgpu-hal/src/lib.rs
@@ -617,6 +617,10 @@ pub trait Queue: WasmNotSendSync {
 ///   live `CommandBuffers` built from it. All the `CommandBuffer`s
 ///   are destroyed, and their resources are freed.
 ///
+/// You may want to implement the [core::ops::Drop] trait to discard
+/// the current commands before the encoder is dropped (e.g. using
+/// [CommandEncoder::discard_encoding]).
+///
 /// # Safety
 ///
 /// - The `CommandEncoder` must be in the states described above to
@@ -651,6 +655,10 @@ pub trait CommandEncoder: WasmNotSendSync + fmt::Debug {
     /// is the only safe thing to do with the encoder.
     ///
     /// This puts this `CommandEncoder` in the "closed" state.
+    ///
+    /// Implementations of this function must be idempotent, i.e.
+    /// if the function has just been called, calling it again should
+    /// not do anything.
     ///
     /// # Safety
     ///
--- a/wgpu-hal/src/lib.rs
+++ b/wgpu-hal/src/lib.rs
@@ -657,8 +657,8 @@ pub trait CommandEncoder: WasmNotSendSync + fmt::Debug {
     /// This `CommandEncoder` must be in the "recording" state.
     ///
     /// Callers must not assume that implementations of this
-    /// function is idempotent. Calling it multiple times in a
-    /// row might not necessarily be safe.
+    /// function are idempotent, so should not call it
+    /// multiple times in a row.
     unsafe fn discard_encoding(&mut self);
 
     /// Return a fresh [`CommandBuffer`] holding the recorded commands.
--- a/wgpu-hal/src/lib.rs
+++ b/wgpu-hal/src/lib.rs
@@ -617,10 +617,6 @@ pub trait Queue: WasmNotSendSync {
 ///   live `CommandBuffers` built from it. All the `CommandBuffer`s
 ///   are destroyed, and their resources are freed.
 ///
-/// You may want to implement the [core::ops::Drop] trait to discard
-/// the current commands before the encoder is dropped (e.g. using
-/// [CommandEncoder::discard_encoding]).
-///
 /// # Safety
 ///
 /// - The `CommandEncoder` must be in the states described above to
@@ -656,13 +652,13 @@ pub trait CommandEncoder: WasmNotSendSync + fmt::Debug {
     ///
     /// This puts this `CommandEncoder` in the "closed" state.
     ///
-    /// Implementations of this function must be idempotent, i.e.
-    /// if the function has just been called, calling it again should
-    /// not do anything.
-    ///
     /// # Safety
     ///
     /// This `CommandEncoder` must be in the "recording" state.
+    ///
+    /// Callers must not assume that implementations of this
+    /// function is idempotent. Calling it multiple times in a
+    /// row might not necessarily be safe.
     unsafe fn discard_encoding(&mut self);
 
     /// Return a fresh [`CommandBuffer`] holding the recorded commands.
--- a/wgpu-hal/src/lib.rs
+++ b/wgpu-hal/src/lib.rs
@@ -645,7 +645,7 @@ pub trait CommandEncoder: WasmNotSendSync + fmt::Debug {
     /// This `CommandEncoder` must be in the "closed" state.
     unsafe fn begin_encoding(&mut self, label: Label) -> Result<(), DeviceError>;
 
-    /// Discard the command list under construction, if any.
+    /// Discard the command list under construction.
     ///
     /// If an error has occurred while recording commands, this
     /// is the only safe thing to do with the encoder.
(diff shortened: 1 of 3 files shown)


wgpu-hal/src/lib.rs
```diff
@@ -617,6 +617,10 @@ pub trait Queue: WasmNotSendSync {
 ///   live `CommandBuffers` built from it. All the `CommandBuffer`s
 ///   are destroyed, and their resources are freed.
 ///
+/// You may want to implement the [core::ops::Drop] trait to discard
+/// the current commands before the encoder is dropped (e.g. using
+/// [CommandEncoder::discard_encoding]).
```
cwfitzgerald: I'm not sure this note is correct - the user of the api shouldn't ever let a command encoder fall on the floor, they should all be calling https://wgpu.rs/doc/wgpu_hal/trait.Device.html#tymethod.destroy_command_encoder when it's time to delete the command encoder, correct?

cwfitzgerald (dismissed):
Looks good, going to flag Erich on this one

jimblandy:
I'm actually -0.5 on this PR. The idea is for `wgpu_hal` traits to permit implementations to track as little state as pratical, and put all state tracking, validation, resource management, etc. in common code in `wgpu-hal`. Properties like idempotence should be easily manageable in `wgpu-core`.

jimblandy (changes requested):
Having thought about this more, I'd like to put a temporary pause on this PR.

At least given my understanding of the role wgpu's core/hal distinction is supposed to play in keeping wgpu maintainable, adding obligations that `wgpu-hal` trait implementors must meet to make life easier for their users is putting the code in the wrong place. State tracking, validation, and resource management should be handled in common code within `wgpu-core`, not repeated in each `wgpu_hal` backend.

@villuna, apologies for the back-and-forth - there's nothing wrong with the code at all, and it looks like it correctly implements the intended change. I just want a day or so for the maintainers to talk things over and get a shared understanding.

wgpu-hal/src/lib.rs
```diff
@@ -652,6 +656,10 @@ pub trait CommandEncoder: WasmNotSendSync + fmt::Debug {
     ///
     /// This puts this `CommandEncoder` in the "closed" state.
     ///
+    /// Implementations of this function must be idempotent, i.e.
+    /// if the function has just been called, calling it again should
+    /// not do anything.
```
jimblandy: Not pushing null handles onto `CommandEncoder::discarded` seems clearly necessary, lest `begin_encoding` pop them back off and try to use them. And if it turns out that the easiest fix for some bug happens to make a `wgpu_hal` trait method idempotent, great.

But adding idempotency as a requirement for implementations of the `CommandEncoder::discard_encoding` trait method is not the right move. See the crate-level docs in `wgpu_hal/src/lib.rs` for a better sense of why. (Those docs only landed a few days ago, but they're just an expansion of the terse/cryptic docs that were there before.)

jimblandy:
For what it's worth, I've been looking into `wgpu-core`'s handling of hal command encoders over the last week, in the process of trying to get a handle on our locking/synchronization story (see #5539, for example). I'd started to put together some ideas on how to clarify the life cycle of command encoders in `core`, and make it "correct by construction" that the `wgpu_hal::CommandEncoder` contract is satisfied. Clearly I'm not the only person with this on their mind!

wgpu-hal/src/lib.rs
jimblandy: Like @cwfitzgerald, I also think this suggestion should be removed from the PR. From chat, my thinking was:

> The problem with this comment is that it neither tells the user of the trait that they can assume something is taken care of (say, "users can just drop `CommandEncoders`, implementations must handle that"), nor does it tell the implementor of the trait that something is taken care of ("Users must always pass this to `destroy_command_encoder`, implementors can just make sure that works and their job is done.").
>
> By just suggesting `Drop`, the comment muddies the waters about whose obligations are what. It's difficult to write code against this.

wgpu-hal/src/lib.rs
jimblandy: Given #5566, please flip this around to say, instead, that callers *must not* assume that this function is idempotent.

The fact that the PR does make the Vulkan implementation idempotent doesn't change the caller's obligations. Edit: and below I ask that it be made non-idempotent.

wgpu-hal/src/vulkan/command.rs
```diff
@@ -104,8 +111,10 @@ impl crate::CommandEncoder for super::CommandEncoder {
     }
 
     unsafe fn discard_encoding(&mut self) {
-        self.discarded.push(self.active);
-        self.active = vk::CommandBuffer::null();
+        if self.active != vk::CommandBuffer::null() {
+            self.discarded.push(self.active);
+            self.active = vk::CommandBuffer::null();
+        }
```
jimblandy: Contrary to the original comments in #5255, it would probably be best to assert here that `self.active` is not null. If `wgpu-core` is calling `discard_encoding` multiple times, that is a bug in `wgpu-core`. The rationale is explained in #5566.

jimblandy (changes requested):
It sounds like we've got some consensus around #5566, so in that context:

villuna:
Thanks for the comments, I've made the changes you requested. I've also removed the drop implementation, since it was calling discard_encoding twice and triggering the assert panic :sweat_smile:. I guess it's working

jimblandy (approved):
Looks good to me. I did some minor copy-editing.

danielhjacobs:
https://github.com/ruffle-rs/ruffle/issues/20943 seems like a case of the assert triggering a panic, would that be a wgpu or a Ruffle issue?

danielhjacobs:
https://github.com/gfx-rs/wgpu/blob/v25/wgpu-hal/src/vulkan/command.rs#L107

```
panicked at C:\Users\runneradmin.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\wgpu-hal-25.0.2\src\vulkan\command.rs:107:9:
assertion left != right failed
left: 0x0
right: 0x0
```

danielhjacobs:
This shows up in these issues:

- https://github.com/ruffle-rs/ruffle/issues/19620
- https://github.com/ruffle-rs/ruffle/issues/19625
- https://github.com/ruffle-rs/ruffle/issues/19654
- https://github.com/ruffle-rs/ruffle/issues/19855
- https://github.com/ruffle-rs/ruffle/issues/19937
- https://github.com/ruffle-rs/ruffle/issues/20007
- https://github.com/ruffle-rs/ruffle/issues/20090
- https://github.com/ruffle-rs/ruffle/issues/20261
- https://github.com/ruffle-rs/ruffle/issues/20267
- https://github.com/ruffle-rs/ruffle/issues/20324
- https://github.com/ruffle-rs/ruffle/issues/20329
- https://github.com/ruffle-rs/ruffle/issues/20338
- https://github.com/ruffle-rs/ruffle/issues/20428
- https://github.com/ruffle-rs/ruffle/issues/20472
- https://github.com/ruffle-rs/ruffle/issues/20593
- https://github.com/ruffle-rs/ruffle/issues/20863
- https://github.com/ruffle-rs/ruffle/issues/20910
- https://github.com/ruffle-rs/ruffle/issues/20943

danielhjacobs:
Would it be sufficient if `discard_encoding` just early returned if `self.active` was `vk::CommandBuffer::null()` instead of asserting them to not be equal and triggering a panic if they are? I know the PR says "The vulkan implementation for `CommandEncoder` was previously pushing its active CommandBuffer handle to its discard vector every time the `discard_encoding` function was called, even if the handle was null. So I made it only do this if the active handle is non-null, so as to make the function idempotent."

ErichDonGubler:
@danielhjacobs: This issue is _long_ dead. Please file a new issue if there's something wrong happening in WGPU 25!

danielhjacobs:
Filed at https://github.com/gfx-rs/wgpu/issues/7897 already, I'm just fairly sure this was the triggering PR for the issue so thought it may be good to mention here too.

PR #5558, opened 2024-04-18
[naga] Introduce `HashableLiteral`

### LINKED ISSUE #5544: Incorrect const eval equality in naga
**Description**
Incorrect const evaluation of `==` and `!=`. Other comparison operators seem to work fine.

**Repro steps**
Retranslation of this shader
```wgsl
@vertex 
fn vs() -> @builtin(position) vec4<f32> {
    return vec4(1f);
}

@fragment 
fn fs() -> @location(0) vec4<f32> {
    var x: f32;
    if 1 == 1 { // <- should be true
        x = 2f;
    } else {
        x = 3f;
    }

    return vec4f(x);
}
```
..will output this
```wgsl
@vertex 
fn vs() -> @builtin(position) vec4<f32> {
    return vec4(1f);
}

@fragment 
fn fs() -> @location(0) vec4<f32> {
    var x: f32;

    if false {
        x = 2f;
    } else {
        x = 3f;
    }
    let _e4 = x;
    return vec4(_e4);
}
```

From development version of naga
```sh
> naga --version
0.19.0
```

teoxoy:
The only reason `Literal` had to implement `Eq` and `Hash` was due to the SPV backend having to cache those in a `HashMap`.

The `PartialEq` impl was error prone due to the match not being exhaustive.

Fixes https://github.com/gfx-rs/wgpu/issues/5544.

### DIFF
--- a/naga/src/back/spv/mod.rs
+++ b/naga/src/back/spv/mod.rs
@@ -456,7 +456,7 @@ impl recyclable::Recyclable for CachedExpressions {
 
 #[derive(Eq, Hash, PartialEq)]
 enum CachedConstant {
-    Literal(crate::Literal),
+    Literal(crate::proc::HashableLiteral),
     Composite {
         ty: LookupType,
         constituent_ids: Vec<Word>,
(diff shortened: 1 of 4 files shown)


jimblandy (approved):
Cute.

PR #5562, opened 2024-04-19
[core] Make `cargo doc --document-private-items` work again.

jimblandy:
Add a CI job to check it, for the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,6 +78,7 @@ jobs:
     # runtime is normally 2-8 minutes
     #
     # currently high due to documentation time problems on mac.
+    # https://github.com/rust-lang/rust/issues/114891
     timeout-minutes: 30
 
     strategy:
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -229,6 +230,14 @@ jobs:
...
           # build docs
           cargo +${{ env.DOCS_RUST_VERSION }} doc --target ${{ matrix.target }} --all-features --no-deps
+      - name: check private item docs
+        if: matrix.kind == 'native'
+        shell: bash
+        run: |
+          set -e
+
+          # wgpu_core package
+          cargo +${{ env.DOCS_RUST_VERSION }} doc --target ${{ matrix.target }} --all-features --no-deps --package wgpu-core --document-private-items
```
cwfitzgerald, line 240: Any reason not to do the whole repo?

cwfitzgerald (approved):

.github/workflows/ci.yml
jimblandy: TOO MUCH WORK

jimblandy: That is, I doubt the other crates build without errors. I'll add more packages once I get them working. Naga is probably in the best shape, because I did test with `--document-private-items` until recently, I think.

Issue #5559, opened 2024-04-18
Vulkan Swapchain and Submit Synchronization Incorrect


cwfitzgerald:
We are currently using a single ping-pong semaphore (I'm calling this set of two semaphores a "relay semaphore") for all of our synchronization between acquire/present and between submissions. This is wrong for multiple reasons.

First using the same relay semaphore for both the submit -> present barrier and the submit -> submit barrier means that a single semaphore is being waited on twice. The act of waiting on a semaphore also resets the semaphore, so waiting on a single semaphore twice results in undefined behavior.

Second we are using the same relay semaphore for all acquire -> submit -> present cycles. This would be fine if we could guarantee that the previously used submit had finished its wait and the semaphore was reset by the time we scheduled the following acquire, but we can't. We need to have N relay semaphores, one per acquire cycle, until we can prove that the least-recently-used semaphore is finished being used.

I have provided the diagram of how submission synchronization needs to work with lines. Blue lines are swapchain relay semaphores (one set per frame). Orange lines are submit -> submit relay semaphore (a single one on the device.

```mermaid
flowchart LR
    A1[Acquire 1] --> S1[Submit 1]
    S1 --> P1[Present 1]
    S1 --> S21

    A2[Acquire 2] --> S21[Submit 2]
    S21 --> S22[Submit 3]
    S22 --> P2[Present 2]
    S22 --> S3[Submit 4]
    S3 --> S4
    
    A3[Acquire 3] --> S4[Submit 5]
    S4 --> P3[Present 3]

    subgraph Frame 3
        A3
        S4
        P3
    end

    subgraph Free Submission
        S3
    end

    subgraph Frame 2
        A2
        S21
        S22
        P2
    end

    subgraph Frame 1
        A1
        S1
        P1
    end

    linkStyle 2,4,6,7 stroke:#ffa600,whoops;
    linkStyle 0,1,3,5,8,9 stroke:#019297,whoops;
```

Possibly related:
- #5491 
- #5374
- #4919 
- #3452

jimblandy:
The use of relaxed atomics in there seems very strange to me. I can't imagine how one could get away with not caring about which values concurrent threads would see for `relay_index`.

> We need to have N relay semaphores, one per acquire cycle, until we can prove that the least-recently-used semaphore is finished being used.

[...]

PR #5563, opened 2024-04-19
[hal] Document `Queue::submit` ordering guarantees a bit.

jimblandy:


jimblandy:
Hmm... I think this could be phrased better.

wgpu-hal/src/lib.rs
```diff
@@ -553,6 +553,15 @@ pub trait Queue: WasmNotSendSync {
 
     /// Submits the command buffers for execution on GPU.
     ///
+    /// Command buffers execute in the order in which they appear in
+    /// `command_buffers`. All results produced by one buffer are visible to the
+    /// next.
+    ///
+    /// If the caller has ensured that one call to `Queue::submit` on a
```
JMS55: /// If `Queue::submit` is called on one `Queue`, and then again on a different queue,
    /// then all the command buffers from the first call will finish executing before any of the 
    /// command buffers from the second call begin.

jimblandy: Actually, there's no ordering between queues. I've rephrased to make this clearer.

jimblandy:
This will conflict with the other hal doc PRs, but it's fine to review.

Wumpf (approved):
tough one. I had to spend quite a bit of time convince myself of this bit `all complete execution before any of the second submission's commands begin`, but from the looks of it we really do enforce that strictly.

jimblandy:
> tough one. I had to spend quite a bit of time convince myself of this bit `all complete execution before any of the second submission's commands begin`, but from the looks of it we really do enforce that strictly.

Thank you for the close reading! I was unsure too, but I think we couldn't function without this requirement, because waiting for fence values only works if things actually complete in increasing fence value order. But see also #5559.

PR #5564, opened 2024-04-19
[naga] add `Expression::Override` to `needs_pre_emit`

teoxoy:
We need to add `Expression::Override` to `needs_pre_emit` or else we get `ExpressionError::NotInScope` on the updated snapshot.


jimblandy (approved):
Thanks for the fix. Sorry I didn't catch this when reviewing the original overrides PR.

PR #5566, opened 2024-04-20
[hal doc] Document the main traits, and hal's validation approach

jimblandy:
* Describe `wgpu-hal`'s validation principles.

* Describe `wgpu-hal`'s main traits in the top-level docs.

* Consistenly use `wgpu-hal` for the crate name.

### DIFF
--- a/wgpu-hal/src/lib.rs
+++ b/wgpu-hal/src/lib.rs
@@ -110,6 +110,46 @@
  *
  * [tdc]: wgt::DownlevelCapabilities
  *
+ * ## Traits
+ *
+ * The `wgpu-hal` crate defines a handful of traits that together
+ * represent a cross-platform abstraction for modern GPU APIs.
+ *
+ * - The [`Api`] trait represents a `wgpu-hal` backend. It has no methods of its
+ *   own, only a collection of associated types.
+ *
+ * - [`Api::Instance`] implements the [`Instance`] trait. [`Instance::init`]
+ *   creates an instance value, which you can use to enumerate the adapters
+ *   available on the system. For example, [`vulkan::Api::Instance::init`][Ii]
+ *   returns an instance that can enumerate the Vulkan physical devices on your
+ *   system.
+ *
+ * - [`Api::Adapter`] implements the [`Adapter`] trait, representing a
+ *   particular device from a particular backend. For example, a Vulkan instance
+ *   might have a Lavapipe software adapter and a GPU-based adapter.
+ *
+ * - [`Api::Device`] implements the [`Device`] trait, representing an active
+ *   link to a device. You get a device value by calling [`Adapter::open`], and
+ *   then use it to create buffers, textures, shader modules, and so on.
+ *
+ * - [`Api::Queue`] implements the [`Queue`] trait, which you use to submit
+ *   command buffers to a given device.
+ *
+ * - [`Api::CommandEncoder`] implements the [`CommandEncoder`] trait, which you
+ *   use to build buffers of commands to submit to a queue. This has all the
+ *   methods for drawing and running compute shaders, which is presumably what
+ *   you're here for.
+ *
+ * - [`Api::Surface`] implements the [`Surface`] trait, which represents a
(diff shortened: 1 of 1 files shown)


wgpu-hal/src/lib.rs
```diff
@@ -110,6 +110,90 @@
...
+ * to provide portable behavior, and report conditions that the user can't
+ * reasonably anticipate, like device loss or running out of memory.
+ *
+ * The `wgpu` crate collection is intended for use in security-sensitive
+ * applications, like web browsers, where the API is available to untrusted
+ * code. This means that `wgpu-core`'s validation is not simply a service to
+ * developers, to be provided opportunistically when the performance costs are
+ * acceptable and the necessary data is ready at hand. Rather, `wgpu-core`'s
+ * validation must be exhaustive, to ensure that even malicious content cannot
+ * provoke and exploit undefined behavior in the platform's graphics API.
```
Wumpf, line 168: this is so central and fundamental to the implementation philosophy of the project that I'm quite surprised we don't have anything like that on the front page.

wgpu-hal/src/lib.rs
```diff
@@ -110,6 +110,90 @@
...
+ * types within the backend.
+ *
+ * A desire for "defense in depth" may suggest performing additional validation
+ * in `wgpu-hal` when the opportunity arises, but this must be done with
+ * caution. New contributors infer the expectations their changes must meet by
+ * considering not just requirements made explicit in types, tests, assertions,
+ * and comments, but also those implicit in the surrounding code. If a
+ * contributor sees validation code in `wgpu-hal`, we must not let them
+ * conclude, "Oh, `wgpu-hal` checks for this, so `wgpu-core` needn't worry about
+ * it - that would be redundant!" The responsibility for exhaustive validation
```
Wumpf: I do think this clarifying block is necessary indeed but arguing via new contributors is a bit backwards. We do run into this problem as well (speaking for myself at least) so it's not just new contributors. The "there's a slippery slope and we musn't waver because this creates unwanted redundancy"-party is well motivated but pointing to the contribution process distracts from this core fact imho

Wumpf (commented):
love it

wgpu-hal/src/lib.rs
```diff
@@ -110,6 +110,90 @@
...
+ *
+ * [Ii]: Instance::init
+ *
+ * ## Validation is the user's responsibility, not `wgpu-hal`'s
+ *
+ * As much as possible, `wgpu-hal` traits place the burden of validation,
+ * resource tracking, and state tracking on the caller, not on the trait
+ * implementations themselves. Anything which can reasonably be handled in
+ * backend-independent code should be. A `wgpu_hal` backend's sole obligation is
+ * to provide portable behavior, and report conditions that the user can't
```
kpreid:
```suggestion
 * to provide portable behavior, and report conditions that the calling code can't
```

wgpu-hal/src/lib.rs
jimblandy: Do you mean, in the workspace's main README?

wgpu-hal/src/lib.rs
jimblandy: Okay - I've tried to address this in b2a702f55. Let me know what you think.

wgpu-hal/src/lib.rs
jimblandy: Yes, "the user" is sort of weird here. I wish we had a better term for this in general; "calling code" is not bad. I've tried to tighten this up in the various places, but left in a use of "the `wgpu-hal` user" because that seemed explicit enough.

cwfitzgerald (approved):
LGTM, definitely agree on all fronts.

wgpu-hal/src/lib.rs
Wumpf: nice, thank you

wgpu-hal/src/lib.rs
Wumpf: yes! :)

Wumpf (approved):
🚢

Issue #5234, opened 2024-02-10
ETC2 support for GL backend


tasogare3710:
**Description**

ETC2 is required if GL4.3 and higher or GL_ARB_ES3_compatibility extension is available.

**Repro steps**

Specify `Features::TEXTURE_COMPRESSION_ETC2` for `DeviceDescriptor::required_features`.

```
let (device, queue) = adapter
    .request_device(
        &wgpu::DeviceDescriptor {
        required_features: wgpu::Features::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES
                         | wgpu::Features::TEXTURE_COMPRESSION_ETC2,
        ...
        },
        None,
    )
```

**Expected vs observed behavior**

If ETC2 is supported by the GL backend, I want to use ETC2, but currently I get the following error when I try to enable ETC2.

```
Failed to create device: RequestDeviceError { inner: Core(UnsupportedFeature(Features(TEXTURE_COMPRESSION_ETC2))) }
```

**Extra materials**

https://github.com/gfx-rs/wgpu/blob/990324fc337a7c9d836a558f442c8a562fad44f9/wgpu-hal/src/gles/adapter.rs#L507-L508

L508 can be fixed as `supported((3, 0), (4, 3)) || extensions.contains("GL_ARB_ES3_compatibility");`.

I think there is a wrong in `not part of Desktop GL at all`. Or is there some reason?

**Platform**

- Windows [Version 10.0.19045.3930]
- Intel HD Graphics (Haswell GT1)
- GL_VERSION: "4.3.0 - Build 20.19.15.5171"
- GL_SHADING_LANGUAGE_VERSION: "4.30 - Build 20.19.15.5171"
- GL_RENDERER: "Intel(R) HD Graphics"
- GL_VENDOR: "Intel"

PR #5568, opened 2024-04-20
Allow ETC2 in OpenGL if ES3 compatible

valaphee:
**Connections**

**Description**
ETC is also usable on Desktop GL if `GL_ARB_ES3_compatibility` is present. See https://registry.khronos.org/OpenGL/extensions/ARB/ARB_ES3_compatibility.txt

**Testing**
Skybox example with only ETC2 used.

**Checklist**

- [x] Run `cargo fmt`.
- [ x Run `cargo clippy`. If applicable, add:


valaphee:
Btw, just noticed that `ANY_SAMPLES_PASSED_CONSERVATIVE` is theoretically also an ES3 feature, Should be replaced with `ANY_SAMPLES_PASSED`, but considering how few drivers don't support `GL_ARB_ES3_compatibility`...

Wumpf (approved):
huh super surprised this works given how rare etc support is on desktop e.g. according to the vulkan database, but that's what the `GL_ARB_ES3_compatibility` spec says should be possible!
Thanks!

valaphee:
Just figured out that etc is supported on all platforms, starting with OpenGL 4.3, GL_ARB_ES3_compatibility is for OpenGL 4.2 and prior, but GL_ARB_ES3_compatibility is present anyway.

teoxoy:
That seems to be on par with the findings in https://github.com/gfx-rs/wgpu/issues/5234.

PR #5574, opened 2024-04-21
When a `#[gpu_test]` test fails, print its source location.

jimblandy:
Use `std::panic::Location` to record the source location of each `#[gpu_test]` test, and if it fails, include that in the error output. This is not essential, but it should make working with failures a bit more comfortable.

### DIFF
--- a/tests/src/config.rs
+++ b/tests/src/config.rs
@@ -1,4 +1,4 @@
-use std::{future::Future, pin::Pin, sync::Arc};
+use std::{future::Future, panic::Location, pin::Pin, sync::Arc};
 
 use crate::{TestParameters, TestingContext};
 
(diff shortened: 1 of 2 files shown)


tests/src/config.rs
```diff
@@ -26,14 +26,17 @@ cfg_if::cfg_if! {
 #[derive(Clone)]
 pub struct GpuTestConfiguration {
     pub(crate) name: String,
+    pub(crate) location: &'static std::panic::Location<'static>,
```
cwfitzgerald: Could we import this?

tests/src/run.rs
```diff
@@ -116,7 +116,7 @@ pub async fn execute_test(
 
     // The call to matches_failure will log.
     if expectations_match_failures(&test_info.failures, failures) == ExpectationMatchResult::Panic {
-        panic!();
+        panic!("{}: test {:?} did not behave as expected", config.location, config.name);
```
cwfitzgerald: Needs a format

cwfitzgerald (approved):
Good after nits

PR #5570, opened 2024-04-20
Remove lifetime constraints from `wgpu::ComputePass` methods

Wumpf:
**Connections**
* Follow-up & based on #5432
     * ⚠️ Will rebase once #5432 lands as this contains all the changes again.
* Important puzzle piece of #1453

Part of a series towards lifetime removal on compute passes:
* #5409
* #5432
* ➡️ #5570
* #5575
* #5620
* #5671

**Description**
**Removes the need for lifetime dependency on any resource passed into a ComputePass.**
Lifetime constraints to the originating encoder & queries used during creation still apply for the moment (to be addressed in a follow-up.

Directly builds on #5432 by moving the creation of `ArcComputeCommand` into compute pass methods.

To accomplish this I had to add a HalApi generic dependency on `ComputePass`. Since I didn't want to introduce a new object to the id/objecttracking system, I instead decided to create a `DynComputePass` on `wgpu/wgpu_core.rs` side. Hopefully we can simplify this as part of the generic refactor (#5124). Note that the new lifetime dependency made boxing of the wgpu-core compute pass is unavoidable at this point and `DynComputePass` allows us to merge the otherwise required `gfx_select` & type cast into a single trait/virtual call.
An alternative I expolored was to do the type erasure internally on `BasePass<ArcComputeCommand<A>>` but I found this quite unwieldy, also this wouldn't cover timestamp writes which are intentionally left unaddressed in this PR.

**Testing**
Added a new test that ensures that resources passed into a ComputePass can be dropped before executing the compute pass.
Proved that the test would crash before these changes and passes with them.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`. If applicable, add:
  - [x] `--target wasm32-unknown-unknown`
  - [ ] `--target wasm32-unknown-emscripten`
- [x] Run `cargo xtask test` to run tests.
- [x] Add change to `CHANGELOG.md`. See simple instructions inside file.

PR #5575, opened 2024-04-21
Make compute pass end consume the pass

Wumpf:
**Connections**
* Follow-up & based on #5570
     * ⚠️ Will rebase once #5570lands as this contains all the changes again.
* Puzzle piece, or rather "convenience refactor" on the way to #1453

* #5409
* #5432
* #5570
* ➡️ #5575
* #5620
* #5671

**Description**

* **avoid copies on compute pass end:**
   Some work in #5432 highlighted that a compute pass (render pass as well, but let's ignore this for now) isn't fully consumed during pass ending which means we have to do a couple of copies despite the fact that a pass can not be consumed twice.
* **error on ending compute pass twice:**
Any operation on an ended compute pass now causes an validation error (as it should per spec!). As expected this allows to remove a few copies (including needless reference count increment + decrement!). Naturally, this means now that *all* operations on the compute pass are failable
* **rename run to end, make end operation part of compute pass:**
Previously, ending a pass was seen as an operation on the command encoder across the api/crate layers. Now it is part of the render/compute pass in accordance with the WebGPU spec. This is relatively minor but also comes with an error check.
* **improve error handling:**
Compute pass errors make now use of its label

**Testing**
Existing tests still pass, should be enough for this one.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/command/compute.rs
+++ b/wgpu-core/src/command/compute.rs
@@ -65,6 +65,10 @@ impl<A: HalApi> ComputePass<A> {
         self.parent_id
     }
 
+    pub fn label(&self) -> Option<&str> {
+        self.base.as_ref().and_then(|base| base.label.as_deref())
+    }
+
     fn base_mut<'a>(
         &'a mut self,
         scope: PassErrorScope,
--- a/wgpu-core/src/command/compute.rs
+++ b/wgpu-core/src/command/compute.rs
@@ -161,6 +161,12 @@ pub enum ComputePassErrorInner {
     Bind(#[from] BindError),
     #[error(transparent)]
     PushConstants(#[from] PushConstantUploadError),
+    #[error("Push constant offset must be aligned to 4 bytes")]
+    PushConstantOffsetAlignment,
+    #[error("Push constant size must be aligned to 4 bytes")]
+    PushConstantSizeAlignment,
+    #[error("Ran out of push constant space. Don't set 4gb of push constants per ComputePass.")]
+    PushConstantOutOfMemory,
     #[error(transparent)]
     QueryUse(#[from] QueryUseError),
     #[error(transparent)]
@@ -930,19 +936,19 @@ impl Global {
         let scope = PassErrorScope::SetPushConstant;
         let base = pass.base_mut(scope)?;
 
-        assert_eq!(
-            offset & (wgt::PUSH_CONSTANT_ALIGNMENT - 1),
-            0,
-            "Push constant offset must be aligned to 4 bytes."
-        );
-        assert_eq!(
-            data.len() as u32 & (wgt::PUSH_CONSTANT_ALIGNMENT - 1),
-            0,
-            "Push constant size must be aligned to 4 bytes."
-        );
-        let value_offset = base.push_constant_data.len().try_into().expect(
-            "Ran out of push constant space. Don't set 4gb of push constants per ComputePass.",
-        ); // TODO: make this an error that can be handled
+        if offset & (wgt::PUSH_CONSTANT_ALIGNMENT - 1) != 0 {
+            return Err(ComputePassErrorInner::PushConstantOffsetAlignment).map_pass_err(scope);
+        }
+
+        if data.len() as u32 & (wgt::PUSH_CONSTANT_ALIGNMENT - 1) != 0 {
+            return Err(ComputePassErrorInner::PushConstantSizeAlignment).map_pass_err(scope);
+        }
+        let value_offset = base
+            .push_constant_data
+            .len()
+            .try_into()
+            .map_err(|_| ComputePassErrorInner::PushConstantOutOfMemory)
+            .map_pass_err(scope)?;
 
         base.push_constant_data.extend(
             data.chunks_exact(wgt::PUSH_CONSTANT_ALIGNMENT as usize)
--- a/wgpu-core/src/command/compute.rs
+++ b/wgpu-core/src/command/compute.rs
@@ -279,11 +279,11 @@ impl Global {
         Box::new(ComputePass::<A>::new(parent_id, desc))
     }
 
-    pub fn command_encoder_run_compute_pass<A: HalApi>(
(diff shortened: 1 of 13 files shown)


wgpu-core/src/command/render.rs
```diff
@@ -247,10 +247,16 @@ impl RenderPass {
         }
     }
 
+    #[inline]
```
ErichDonGubler, line 250: question(non-blocking): Why add `#[inline]` here? Was it necessary to maintain some performance characteristic alongside these other changes?

wgpu-core/src/command/render.rs
```diff
@@ -1303,13 +1309,9 @@ impl<'a, 'd, A: HalApi> RenderPassInfo<'a, 'd, A> {
...
 
 impl Global {
-    pub fn command_encoder_run_render_pass<A: HalApi>(
-        &self,
-        encoder_id: id::CommandEncoderId,
-        pass: &RenderPass,
-    ) -> Result<(), RenderPassError> {
-        self.command_encoder_run_render_pass_impl::<A>(
-            encoder_id,
+    pub fn render_pass_end<A: HalApi>(&self, pass: &RenderPass) -> Result<(), RenderPassError> {
```
ErichDonGubler, line 1312: nitpick: Your commit subject for using labeled errors for render passes:

```
use handle_error_no_label instead of handle_error for render_pass_end
```

…seems backwards. That is, you are replacing `handle_error_no_label` with `handle_error`, not the other way around.

wgpu/src/backend/wgpu_core.rs
```diff
@@ -2361,7 +2336,13 @@ impl crate::Context for ContextWgpuCore {
...
             .set_bind_group(&self.0, index, *bind_group, offsets)
         {
-            self.handle_error_nolabel(&pass_data.error_sink, cause, "ComputePass::set_bind_group");
+            self.handle_error(
+                &pass_data.error_sink,
+                cause,
+                LABEL,
+                pass_data.pass.label(),
+                "ComputePass::set_bind_group",
+            );
```
ErichDonGubler, line 2397: praise: This is a nice win, getting labels into these errors (finally). I think we might already have been hurt by this in Firefox debugging. 😅

wgpu-core/src/command/compute.rs
```diff
@@ -429,7 +457,7 @@ impl Global {
             .flags
             .contains(wgt::InstanceFlags::DISCARD_HAL_LABELS);
         let hal_desc = hal::ComputePassDescriptor {
-            label: hal_label(base.label, self.instance.flags),
+            label: hal_label(base.label.as_deref(), self.instance.flags),
```
ErichDonGubler, line 460: nitpick: You introduced `….as_ref().map(|s| s.as_str())` in 2d1239a2e, then changed it to `….as_deref()` in 6fe51ff22. It took me a hot second to figure out if this was intentional or not. 😅

I've consolidated both changes into the commit introducing this code in the first place. No action needed here.

wgpu-core/src/command/compute.rs
```diff
@@ -897,33 +932,36 @@ impl Global {
...
-            "Push constant offset must be aligned to 4 bytes."
-        );
-        assert_eq!(
-            data.len() as u32 & (wgt::PUSH_CONSTANT_ALIGNMENT - 1),
-            0,
-            "Push constant size must be aligned to 4 bytes."
-        );
-        let value_offset = pass.base.push_constant_data.len().try_into().expect(
-            "Ran out of push constant space. Don't set 4gb of push constants per ComputePass.",
-        ); // TODO: make this an error that can be handled
```
ErichDonGubler, line 913: nitpick: I've moved the conversion of these panicking cases into errors as their own separate commit (see ecd969969). 

No action needed here.

ErichDonGubler (approved):
praise: These changes make sense to me. Glad to see them!

nitpick: I've taken the liberty of modifying history a bit here, so review is easier and that we might reasonably consider rebasing instead of squashing. I've made individual conversations where it made sense. For general changes, I've folded the following commits into parents: d781e30, e07ba12, 6315094

wgpu-core/src/command/render.rs
ErichDonGubler: I've edited history myself to remedy this. No action needed!

ErichDonGubler:
I'll leave this PR un-merged for now, not because I'm not confident it'll land as-is, but because I want to give @Wumpf a chance to look over changes I've made to this PR's history before they enter mainline.

Wumpf:
@ErichDonGubler thanks for the review!

ah apologies I didn't pay all that much attention to the history on this PR. Due to how I kept stacking changes and created cutoff points for reviewable pieces I didn't maintain a all too straight history line there. I'll make sure the commits make a little bit more sense for putting the next of the series out of Draft.
I appreciate your efforts of cleaning it up, but now that you already went through the pain of puzzling it together it's pretty much in vain since we squash on merge 🤷

ErichDonGubler:
@Wumpf: I don't feel like my time was wasted, since I felt that modifying the history was important for me developing understanding sufficient to sign off on the review. No worries! 😊

Nice history is nice tho. I have found it tends to speed everybody up at the review stage, so I look forward to the next steps.

wgpu-core/src/command/render.rs
Wumpf: admittedly didn't do any profiling at all here, more of a habit for these trivial getters so that I don't have to rely on link time optimizations and I did the same to the one on `ComputePass`, so I just ensured symmetry here

Wumpf: I figure since you wrote non-blocking I'll leave it be. It's borders on a style decision here since I have no proof for it having any direct value. I have a leaning towards this, seeing it as good practice as described here
https://matklad.github.io/2021/07/09/inline-in-rust.html#Inlining-in-Practice
> when building libraries, proactively add #[inline] to small non-generic function

but no strong opinion otherwise until I have actual proof for this to have any actual effect

PR #5590, opened 2024-04-23
[spv-out] Move `request_type_capabilities` outside of `LocalType` lookup

### LINKED ISSUE #5569: Atomics and Scalars are collapsed to the same LocalType
Atomics and Scalars are collapsed to the same LocalType
https://github.com/gfx-rs/wgpu/blob/trunk/naga/src/back/spv/mod.rs#L360
Type capability requests are based on TypeInner
https://github.com/gfx-rs/wgpu/blob/trunk/naga/src/back/spv/writer.rs#L835
But deduplicated by LocalType
https://github.com/gfx-rs/wgpu/blob/trunk/naga/src/back/spv/writer.rs#L977

This means if an `atomic<u64>` is seen after `u64`, then the necessary `Int64Atomics` capability doesn't get requested.

`LocalType` should represent Atomic and non-Atomic types with distinct values, but I'm not sure how. Do we want a `atomic: bool` or an `Atomic {}` variant?

atlv24:
**Connections**
Fixes #5569
Unblocks #5383

**Description**
Avoids incorrect caching as per #5569

**Testing**
Existing tests

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/naga/src/back/spv/writer.rs
+++ b/naga/src/back/spv/writer.rs
@@ -970,13 +970,12 @@ impl Writer {
         handle: Handle<crate::Type>,
     ) -> Result<Word, Error> {
         let ty = &arena[handle];
+        // If it's a type that needs SPIR-V capabilities, request them now.
+        // This needs to happen regardless of the LocalType lookup succeeding,
+        // because some types which map to the same LocalType have different
+        // capability requirements. See https://github.com/gfx-rs/wgpu/issues/5569
+        self.request_type_capabilities(&ty.inner)?;
         let id = if let Some(local) = make_local(&ty.inner) {
-            // If it's a type that needs SPIR-V capabilities, request them now.
-            // This needs to happen regardless of the LocalType lookup succeeding,
-            // because some types which map to the same LocalType have different
-            // capability requirements. See https://github.com/gfx-rs/wgpu/issues/5569
-            self.request_type_capabilities(&ty.inner)?;
-
             // This type can be represented as a `LocalType`, so check if we've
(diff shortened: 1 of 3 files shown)


naga/src/back/spv/writer.rs
```diff
@@ -971,6 +971,12 @@ impl Writer {
     ) -> Result<Word, Error> {
         let ty = &arena[handle];
         let id = if let Some(local) = make_local(&ty.inner) {
+            // If it's a type that needs SPIR-V capabilities, request them now.
+            // This needs to happen regardless of the LocalType lookup succeeding,
+            // because some types which map to the same LocalType have different
+            // capability requirements. See https://github.com/gfx-rs/wgpu/issues/5569
+            self.request_type_capabilities(&ty.inner)?;
+
```
teoxoy:
```suggestion
        // If it's a type that needs SPIR-V capabilities, request them now.
        // This needs to happen regardless of the LocalType lookup succeeding,
        // because some types which map to the same LocalType have different
        // capability requirements. See https://github.com/gfx-rs/wgpu/issues/5569
        self.request_type_capabilities(&ty.inner)?;
        let id = if let Some(local) = make_local(&ty.inner) {
```

I feel like we should take this out of the `if` as well in case we need to request capabilities for non local types.

teoxoy: I can apply the suggestion and merge this if there are no concerns with it.

teoxoy (approved):
Thanks!

PR #5594, opened 2024-04-24
[core] Properly recycle `Device::temp_suspected`.

jimblandy:
Change the various uses of `Device::temp_suspected` to correctly
recycle its allocations when possible.

In the absence of errors, `Device::temp_suspected` is only used as
temporary storage within particular function calls, and is a field of
`Device` only as a way to recycle its allocations.

However, the code in `Global::queue_submit` that tries to use
`Device::temp_suspected` in this way doesn't quite pull it off.
It unconditionally replaces any prior `ResourceMaps` stored there
with a fresh empty one, clears the prior one, and then drops it,
wasting its allocations.

Rather than discard everything in `temp_selected` on every call to
`Device::maintain`, change `ResourceMaps::extend` to take the source
via `&mut` rather than by value, and have callers only obtain a `&mut`
to `Device::temp_selected`, rather than using `Option::take` to take
ownership of the value.

### DIFF
--- a/wgpu-core/src/device/life.rs
+++ b/wgpu-core/src/device/life.rs
@@ -93,7 +93,7 @@ impl<A: HalApi> ResourceMaps<A> {
         destroyed_textures.clear();
     }
 
-    pub(crate) fn extend(&mut self, mut other: Self) {
+    pub(crate) fn extend(&mut self, other: &mut Self) {
         let ResourceMaps {
             buffers,
             staging_buffers,
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -485,12 +485,14 @@ impl<A: HalApi> Device<A> {
     }
 
     pub(crate) fn untrack(&self, trackers: &Tracker<A>) {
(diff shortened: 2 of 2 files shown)


ErichDonGubler:
I'll take on review for this tomorrow. 👍🏻

ErichDonGubler:
I promised review for yesterday, but it looks like this will be in my queue until Monday. Sorry for the wait! 😖

ErichDonGubler (changes requested):
This PR conflicts with <https://github.com/gfx-rs/wgpu/pull/5653>, it seems, particularly my thought about [removing `temp_suspected` altogether](https://github.com/gfx-rs/wgpu/pull/5653#discussion_r1589505878). I think this should be closed? 🤔

jimblandy:
Yes, most of this PR is obviated by #5653, but the change to `Device::untrack` still makes sense.

ErichDonGubler:
@jimblandy: Weeeell then let's rebase and re-review! 😀

ErichDonGubler (approved):
Approving contingent on the rebase that we've agreed on.

PR #5595, opened 2024-04-24
Clean up weak references to texture views and bind groups

xiaopengli89:
**Description**

Cleanup weak references to texture views and bind groups to prevent leak.

image

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/device/life.rs
+++ b/wgpu-core/src/device/life.rs
@@ -578,12 +578,15 @@ impl<A: HalApi> LifetimeTracker<A> {
     fn triage_suspected_texture_views(&mut self, trackers: &Mutex<Tracker<A>>) -> &mut Self {
         let mut trackers = trackers.lock();
         let resource_map = &mut self.suspected_resources.texture_views;
-        Self::triage_resources(
+        let removed_views = Self::triage_resources(
             resource_map,
             self.active.as_mut_slice(),
             &mut trackers.views,
             |maps| &mut maps.texture_views,
         );
+        for v in removed_views {
+            v.parent.views.lock().retain(|weak| weak.strong_count() > 0);
+        }
         self
     }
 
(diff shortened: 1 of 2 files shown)


cwfitzgerald:
Tagging jimb on this as he's most aware of the lifetime management in wgpu-core

wgpu-core/src/device/life.rs
```diff
@@ -545,13 +545,17 @@ impl<A: HalApi> LifetimeTracker<A> {
             &mut trackers.bind_groups,
             |maps| &mut maps.bind_groups,
         );
+        let mut used_buffers = vec![];
+        let mut used_textures = vec![];
```
jimblandy: I understand that you're holding these until we're done draining `removed_resource` because you need the strong references in the bind group to go away, to make the `strong_count` check more obviously correct.

But we're already sticking all these buffers and textures in their respective `suspected_resources` tables, which we're guaranteed to scan anyway from the following calls in `triage_suspected`. It would be neater to run the `retain` on the buffers' and textures' backlinks from `triage_suspected_buffers` and `triage_suspected_textures`.

Doing it there would have the added benefit that, if there are several bind groups / texture views freed at once, then we can clear out all their `Weak` pointers in a single pass, rather than making a separate pass over the backlinks for each target.

jimblandy (changes requested):
Thank you for tracking this down! Forgetting to ever actually trim those vectors was sloppy. I really hate backlinks like that, because they always seem to run into problems like this.

However, I think we can do a bit better.

jimblandy:
@xiaopengli89 I've pushed a commit that makes the change I suggested. Could you look it over and see if it seems okay, and verify that it does fix the leak?

xiaopengli89:
> @xiaopengli89 I've pushed a commit that makes the change I suggested. Could you look it over and see if it seems okay, and verify that it does fix the leak?

@jimblandy LGTM.

jimblandy (approved):

xiaopengli89:
@jimblandy Sorry, we cannot use the change because `suspected_resources` only contains resources whose user handle has died.

PR #5596, opened 2024-04-24
CreateBindGroup validation error on device mismatch

### LINKED ISSUE #5582: createBindGroup accepts textures from different device
**Description**
CreateBindGroup does not create ValidationError when textures (happens also when using StorageTexture) are from different devices. When using Samplers from different devices ValidationError is triggered.

**Repro steps**
Repro is here: https://github.com/sagudev/CBG

**Extra materials**
This causes `Set must be allocated from this allocator` crash in gpu-allocator while testing https://gpuweb.github.io/cts/standalone/?q=webgpu:api,validation,createBindGroup:binding_resources,device_mismatch:entry={%22texture%22:{%22multisampled%22:false}} in both servo and firefox.

**Platform**
Ubuntu 23.10
Wgpu 0.19.4 and trunk (4f3d9a2af52d2ab10042787a1115fe8c0779d4d9)

sagudev:
**Connections**
Fixes https://github.com/gfx-rs/wgpu/issues/5582

**Description**
Removed `assertDeviceMatch` from deno_webgpu as this should be checked in wgpu-core (this made two test pass in `webgpu:api,validation,createBindGroup:binding_resources,device_mismatch:*`). Then adding device mismatched check in TextureView case make all tests pass 🎉 .

**Testing**
This is now tested via cts_runner.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@ To run a given set of tests:
 
 ```
 # Must be inside the `cts` folder we just checked out, else this will fail
-cargo run --manifest-path ../Cargo.toml --bin cts_runner -- ./tools/run_deno --verbose "<test string>"
+cargo run --manifest-path ../Cargo.toml -p cts_runner --bin cts_runner -- ./tools/run_deno --verbose "<test string>"
 ```
 
 To find the full list of tests, go to the [online cts viewer](https://gpuweb.github.io/cts/standalone/?runnow=0&worker=0&debug=0&q=webgpu:*).
--- a/deno_webgpu/01_webgpu.js
+++ b/deno_webgpu/01_webgpu.js
@@ -1289,11 +1289,6 @@ class GPUDevice extends EventTarget {
       const resource = entry.resource;
       if (ObjectPrototypeIsPrototypeOf(GPUSamplerPrototype, resource)) {
         const rid = assertResource(resource, prefix, context);
-        assertDeviceMatch(device, resource, {
-          prefix,
-          resourceContext: context,
-          selfContext: "this",
-        });
         return {
           binding: entry.binding,
           kind: "GPUSampler",
@@ -1304,11 +1299,6 @@ class GPUDevice extends EventTarget {
       ) {
         const rid = assertResource(resource, prefix, context);
         assertResource(resource[_texture], prefix, context);
-        assertDeviceMatch(device, resource[_texture], {
-          prefix,
-          resourceContext: context,
-          selfContext: "this",
-        });
         return {
           binding: entry.binding,
           kind: "GPUTextureView",
@@ -1318,11 +1308,6 @@ class GPUDevice extends EventTarget {
         // deno-lint-ignore prefer-primordials
         const rid = assertResource(resource.buffer, prefix, context);
         // deno-lint-ignore prefer-primordials
-        assertDeviceMatch(device, resource.buffer, {
-          prefix,
-          resourceContext: context,
-          selfContext: "this",
-        });
         return {
           binding: entry.binding,
           kind: "GPUBufferBinding",
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -13,6 +13,7 @@ use crate::{
     hal_api::HalApi,
     hal_label,
     hub::Hub,
+    id,
     init_tracker::{
         BufferInitTracker, BufferInitTrackerAction, MemoryInitKind, TextureInitRange,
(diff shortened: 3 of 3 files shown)


sagudev:
I only do not know what to do with `TextureViewArray`. This isn't in spec (anymore?) so this might be wgpu specific? There is a lot of repetition in `*Array` match cases (for loops (should) have same content as there non-array complements), so it would make sense to extract them into separate method, right?

Cargo.toml
```diff
@@ -22,6 +22,7 @@ members = [
 ]
 exclude = []
 default-members = [
+    "cts_runner",
```
sagudev: This was needed or else:
```
cargo run --manifest-path ../Cargo.toml --bin cts_runner -- ./tools/run_deno 'webgpu:api,validation,createBindGroup:binding_resources,device_mismatch:*'
error: no bin target named `cts_runner`.
Available bin targets:
    naga
    spv_parser
    wgsl_parser
    glsl_parser
    ir
    play
    wgpu-examples
    wgpu-info
```
or alternative is to run with `cargo run --manifest-path ../cts_runner/Cargo.toml -- ./tools/run_deno 'webgpu:api,validation,createBindGroup:binding_resources,device_mismatch:*'`

teoxoy:
> I only do not know what to do with `TextureViewArray`. This isn't in spec (anymore?) so this might be wgpu specific?

It's a native only feature for now (called bindless). We should add the same validation for `TextureViewArray`, same for `Buffer` and `BufferArray` since it seems they are also missing this validation.

> There is a lot of repetition in `*Array` match cases (for loops (should) have same content as there non-array complements), so it would make sense to extract them into separate method, right?

Feel free to add a method, we should then be able to use it for all resource variants.

Cargo.toml
teoxoy: I think you can do `-p cts_runner` instead? I don't know why `cts_runner` is not a default member though. IIRC the `cts_runner` wasn't working but it seems to have been recently fixed in https://github.com/gfx-rs/wgpu/pull/5459.

cc @crowlKats @cwfitzgerald

sagudev:
> same for Buffer and BufferArray

This wasn't cached by cts because `Unexpected validation error occurred: Usage flags BufferUsages(UNIFORM) for buffer Id(2,1,vk) do not contain required usage flags BufferUsages(STORAGE)`

EDIT: Opened https://github.com/gpuweb/cts/issues/3707

sagudev:
```
TRY 3 FAIL [   0.598s] cts_runner::integration hello_compute_example
  TRY 3 FAIL [   3.623s] wgpu-examples [Executed Failure: BACKEND] [Vulkan/AMD Radeon Graphics (RADV RENOIR)/0] water
  TRY 3 FAIL [   4.027s] wgpu-examples [Executed Failure: BACKEND] [Vulkan/llvmpipe (LLVM 15.0.7, 256 bits)/1] water
  TRY 3 FAIL [   0.840s] wgpu-test::wgpu-test [Executed Failure: BACKEND] [Vulkan/AMD Radeon Graphics (RADV RENOIR)/0] wgpu_test::shader::struct_layout::uniform_input
  TRY 3 FAIL [   0.957s] wgpu-test::wgpu-test [Executed Failure: BACKEND] [Vulkan/llvmpipe (LLVM 15.0.7, 256 bits)/1] wgpu_test::shader::struct_layout::uniform_input
  TRY 3 FAIL [   0.567s] wgpu-test::wgpu-test [Executed] [Vulkan/llvmpipe (LLVM 15.0.7, 256 bits)/1] wgpu_test::subgroup_operations::subgroup_operations
```

EDIT: All those errors are also happening on trunk on my machine.

Cargo.toml
sagudev: I think this is because cts_runner requires v8_rusty that does not have artifacts for aarch64 (ios ans android) hence the build were failing. I just updated line in readme to use `-p cts_runner`.

Although that's to long to write; justfile or xtask should be used for this imo.

sagudev:
I splitted work into multiple independent commits and CI is working now: https://github.com/sagudev/wgpu/actions/runs/8820454948

teoxoy:
Looks good but I will have to squash this since https://github.com/gfx-rs/wgpu/pull/5596/commits/c39d5d8d372a968ff94eac5ff6bc3effea32a1bd won't compile by itself; it contains a subset of the `create_texture_binding` changes.

teoxoy (approved):
Good stuff, thanks!

PR #5601, opened 2024-04-25
Fix inconsistency in Limits defaults and clean up

atlv24:
**Description**
The docs claimed `max_compute_workgroup_storage_size` had a default value of 16352, but the actual default value in the code was 16384. I did a bit of digging to figure out which was the right value and found https://github.com/gpuweb/gpuweb/issues/2832 discussing it.

**Testing**
Existing tests

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-types/src/lib.rs
+++ b/wgpu-types/src/lib.rs
@@ -1143,9 +1143,7 @@ pub struct Limits {
     /// pipeline output data, across all color attachments.
     pub max_color_attachment_bytes_per_sample: u32,
     /// Maximum number of bytes used for workgroup memory in a compute entry point. Defaults to
-    /// 16352 because of [Apple2]. Higher is "better".
-    ///
-    /// [Apple2]: https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf#page=7
+    /// 16384 because of [Apple2]. Higher is "better".
     pub max_compute_workgroup_storage_size: u32,
     /// Maximum value of the product of the `workgroup_size` dimensions for a compute entry-point.
     /// Defaults to 256. Higher is "better".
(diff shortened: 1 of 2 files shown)


wgpu-types/src/lib.rs
```diff
@@ -1184,6 +1186,12 @@ pub struct Limits {
 
 impl Default for Limits {
     fn default() -> Self {
+        Self::defaults()
+    }
+}
+
+impl Limits {
+    const fn defaults() -> Self {
```
cwfitzgerald, line 1194: Could we make a note of why this is separate from default

wgpu-types/src/lib.rs
```diff
@@ -1198,18 +1206,18 @@ impl Default for Limits {
...
+            max_buffer_size: 256 << 20, // (256 MiB)
             max_vertex_attributes: 16,
             max_vertex_buffer_array_stride: 2048,
             min_uniform_buffer_offset_alignment: 256,
             min_storage_buffer_offset_alignment: 256,
             max_inter_stage_shader_components: 60,
             max_color_attachments: 8,
             max_color_attachment_bytes_per_sample: 32,
-            max_compute_workgroup_storage_size: 16384,
+            max_compute_workgroup_storage_size: 16352,
```
cwfitzgerald: This is unfortunately not the limit for WebGPU - https://gpuweb.github.io/gpuweb/#dom-supported-limits-maxcomputeworkgroupstoragesize

cwfitzgerald (changes requested):

wgpu-types/src/lib.rs
atlv24: yes, i saw that. if you look at older versions of the spec it does have that value however- it was changed. downlevel_defaults has 16352 and the docs claim also claim 16352. Higher is better, so lowering this is fine. If you want all of them changed to 16384 and ignore the apple requirement im fine with that too, but it has to be documented

cwfitzgerald: I think downlevel (i.e. old macs) having 16352 is fine, and non-downlevel (aka webgpu compliant) at 16384 is the right way to go. WebGPU has already decided that it doesn't mind losing those devices.

atlv24: Okay, I'll update the docs then!

cwfitzgerald (approved):

PR #5603, opened 2024-04-25
[core] Add checked `RwLock`.

jimblandy:
Add a ranked wrapper for `RwLock` to `wgpu_core::lock`, with `vanilla` and `ranked` implementations, the latter selected with the `wgpu_validate_locks` cfg. Use this wrapper everywhere in `wgpu-core` instead of `parking_lot::RwLock`.

-   [core] Use `lock::RwLock` in `RenderBundleScope`.
    
    This requires bumping the `LockRankSet` bitset from `u32` to `u64`.

-   [core] Use `lock::RwLock` in `Registry`.

-   [core] Use `lock::RwLock` in `Buffer` and `Texture`.

-   [core] Include `SnatchLock` in lock rankings.

-   [core] Add checked `RwLock`.

-   [core] Abstract out checking code in `lock::ranked`.
    
    Introduce two new private functions, `acquire` and `release`, to the
    `lock::ranked` module, to perform validation for acquiring and
    releasing locks. Change `Mutex::lock` and `MutexGuard::drop` to use
    those functions, rather than writing out their contents.

-   [core] Remove unnecessary `#[inline]` attributes in `lock` module.

-   [core] Remove semicolon from lock rank macro syntax.
    
    In `define_lock_ranks!` macro invocations, don't require a semicolon
    after each rank's "followed by" list.

### DIFF
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -11,6 +11,7 @@ use crate::{
     id::{self, AdapterId, DeviceId, QueueId, SurfaceId},
     init_tracker::TextureInitTracker,
     instance::{self, Adapter, Surface},
+    lock::{rank, RwLock},
     pipeline, present,
     resource::{self, BufferAccessResult},
     resource::{BufferAccessError, BufferMapOperation, CreateBufferError, Resource},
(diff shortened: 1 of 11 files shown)


cwfitzgerald:
Exquisite stuff, with an amazing PR history etiquette.

cwfitzgerald (approved):

PR #5606, opened 2024-04-25
Extract a naga pub create_validator function for use in Bevy

atlv24:
**Description**
Bevy currently duplicates all this logic because its private. Factor out a function and expose it to avoid this

**Testing**
Existing tests

**Checklist**

- [x] Run `cargo fmt`.


Wumpf (approved):
Nice, totally see how that's useful and even without that would be good refactor :+1:

PR #5608, opened 2024-04-25
[naga] fix the way we adjust constant initializers when processing overrides

### LINKED ISSUE #5598: [naga] ValidationError after processing pipeline constants on a SPIR-V from rust-gpu
**Description**
I'm trying to use a shader compiled from Rust to SPIR-V by rust-gpu, and after processing pipeline overridable constants, the shader no longer validates. 
After running `naga::back::pipeline_constants::process_overrides`, the validation fails with 
```
ValidationError(WithSpan { inner: ConstExpression { handle: [8], source: Compose(ComponentType { index: 0 }) }, spans: [(Span { start: 556, end: 584 }, "naga::Expression [8]")] })
```

My goal is to read the SPIR-V, set pipeline overrides and emit GLSL with evaluated constants

Input SPIR-V: [shader.zip](https://github.com/gfx-rs/wgpu/files/15097330/shader.zip)
Rust source:
```rust
pub fn fragment(
    #[spirv(frag_coord)] in_frag_coord: Vec4,
    #[spirv(spec_constant(id = 100, default = 2))] _test_constant: u32,
    output: &mut Vec4,
) {

    let mut test = vec2(in_frag_coord.x, in_frag_coord.y);
    if _test_constant == 1 {
        test = test / 2.0;
    } else {
        test = test / 4.0;
    }

    output.x = test.x;
    output.y = test.y;
}
```
If I remove the `if _test_constant == 1 {` or replace it with `if 1 == 1 {`, then there's no error from naga

**Repro steps**
I prepared a minimum reproducible example: [shader-builder.zip](https://github.com/gfx-rs/wgpu/files/15097368/shader-builder.zip)
cd to `shader_builder` and run `cargo run --release`

I expected it to successfully process all overrides and emit correct GLSL

**Platform**
Windows 11, NVIDIA RTX 3080 Ti

teoxoy:
Fixes https://github.com/gfx-rs/wgpu/issues/5598.

This fixes 2 issues:
- we used to index `adjusted_global_expressions` with the handle index of the constant instead of its initializer
- we used to adjust the initializer multiple times if the arena contained multiple `Expression::Constant`s pointing to the same constant


naga/tests/out/glsl/spec-constants-issue-5598.fragment.Fragment.glsl
```diff
@@ -0,0 +1,34 @@
...
+
+vec4 global_1 = vec4(0.0);
+
+layout(location = 0) out vec4 _fs2p_location0;
+
+void function() {
+    vec2 phi_52_ = vec2(0.0);
+    vec4 _e7 = global;
+    if (false) {
+        phi_52_ = vec2((_e7.x * 0.5), (_e7.y * 0.5));
```
AdrianEddy, line 16: Is eliminating the dead branch in scope for naga? Or that's a job for the target backend compiler?

jimblandy (approved):
Oh, that's silly. I'm amazed things worked at all.

naga/tests/out/glsl/spec-constants-issue-5598.fragment.Fragment.glsl
jimblandy: Naga isn't generally in the optimization business. That would be a huge expansion of scope. Most drivers are running LLVM or the equivalent anyway.

PR #5609, opened 2024-04-25
[core] Fix caller location tracking in `lock::ranked`.

jimblandy:
Properly track the location of callers of `Mutex::lock`, `RwLock::read`, and `RwLock::write`, for use in panic messages.

### DIFF
--- a/wgpu-core/src/lock/ranked.rs
+++ b/wgpu-core/src/lock/ranked.rs
@@ -113,9 +113,8 @@ impl LockState {
 /// update the per-thread state accordingly.
 ///
 /// Return the `LockState` that must be restored when this thread is released.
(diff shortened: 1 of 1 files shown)


ErichDonGubler (approved):

wgpu-core/src/lock/ranked.rs
```diff
@@ -113,9 +113,8 @@ impl LockState {
 /// update the per-thread state accordingly.
 ///
 /// Return the `LockState` that must be restored when this thread is released.
-fn acquire(new_rank: LockRank) -> LockState {
+fn acquire(new_rank: LockRank, location: &'static Location<'static>) -> LockState {
```
cwfitzgerald, line 116: I know this is already merged, but why not make this also track_caller?

PR #5618, opened 2024-04-27
[hal] Document `Api::Fence`, its users, and its Vulkan impl.

jimblandy:


wgpu-hal/src/vulkan/mod.rs
```diff
@@ -571,6 +609,13 @@ pub enum Fence {
 }
 
 impl Fence {
+    /// Return the highest [`FenceValue`] among the signalled fences in `active`.
+    ///
+    /// As an optimization, assume that the returned value should be at least
+    /// `max_value`, and don't bother checking fences whose values are less than
+    /// that.
```
Wumpf: very confused about the naming here. Why is it a `max_value` if we're ignoring anything below. Wouldn't that make it a `min_checked_value`, or to be more precise (and accounting better for off-by-one-confusion) `skip_checking_fences_below_value`?

It's a strange optimization - is it really that expensive to check the fence status?

Wumpf (approved):
all these subtle details 😓 . Appreciate the precise language!

wgpu-hal/src/vulkan/mod.rs
jimblandy: Well, yeah. In practice this argument is always the `last_completed` field of `Fence::FencePool`, so we're just culling out the fences we've already waited for but haven't been able to move to the `free` pool yet.

In that light, perhaps renaming the argument to `last_completed` would tell the story better?

jimblandy: Gave that a try, and tried to motivate the optimization a bit better.

jimblandy:
> Appreciate the precise language!

It's all for naught if it's not actually true! :D

PR #5620, opened 2024-04-28
Remove lifetime dependency of `ComputePass` to its parent command encoder

Wumpf:
**Connections**
* Follow-up & based on #5575
     * ⚠️ Will rebase once #5575 lands as this contains all the changes again.
* Puzzle piece of #1453

* #5409
* #5432
* #5570
* #5575
* ➡️ #5620
* #5671

**Description**

`ComputePass` had a static lifetime dependency to its parent encoder. This PR removes this (`ComputePass<'a>` -> `ComputePass`) by having `ComputePass` own an Arc to its parent encoder. A side-effect of this is that `ComputePass` creation can now fail in theory (if an invalid id is passed or commandbuffer allocation fails for some reason).

Another wide reaching side effect is that command encoders can now be in a `Locked` state. (via WebGPU spec) any operation on a locked command encoder fails and makes the encoder invalid.

The last remaining lifetime dependency to resolve on `ComputePass` after this PR is in `ComputePassTimestampWrites<'a>`: We currently don't bump the reference count on the referenced `QuerySet`. This is in fact more of a pre-existing bug since destroying a QuerySet after recording it in a ComputePass and before closing the pass already causes issues!

**Testing**
* added a test that verifies that a command encoder can be dropped child compute pass is still live
* added a that that while a compute pass is being recorded, trying to do any operation on the encoder fails and makes it invalid

**Checklist**

- [x] Run `cargo fmt`.
- [x] ~Add change to `CHANGELOG.md`. See simple instructions inside file.~

### DIFF
--- a/wgpu-core/src/command/mod.rs
+++ b/wgpu-core/src/command/mod.rs
@@ -422,15 +422,10 @@ impl<A: HalApi> CommandBuffer<A> {
 }
 
 impl<A: HalApi> CommandBuffer<A> {
-    /// Return the [`CommandBuffer`] for `id`, for recording new commands.
-    ///
-    /// In `wgpu_core`, the [`CommandBuffer`] type serves both as encoder and
-    /// buffer, which is why this function takes an [`id::CommandEncoderId`] but
-    /// returns a [`CommandBuffer`]. The returned command buffer must be in the
-    /// "recording" state. Otherwise, an error is returned.
-    fn get_encoder(
+    fn get_encoder_impl(
         hub: &Hub<A>,
         id: id::CommandEncoderId,
+        lock_on_acquire: bool,
     ) -> Result<Arc<Self>, CommandEncoderError> {
         let storage = hub.command_buffers.read();
         match storage.get(id.into_command_buffer_id()) {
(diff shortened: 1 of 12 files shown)


wgpu-core/src/command/mod.rs
```diff
@@ -422,15 +434,75 @@ impl<A: HalApi> CommandBuffer<A> {
...
+            }
             Err(_) => Err(CommandEncoderError::Invalid),
         }
     }
 
+    /// Return the [`CommandBuffer`] for `id` and if successful puts it into the [`CommandEncoderStatus::Locked`] state.
+    ///
+    /// See [`CommandBuffer::get_encoder`].
+    /// Call [`CommandBuffer::unlock_encoder`] to put the [`CommandBuffer`] back into the [`CommandEncoderStatus::Recording`] state.
+    fn lock_encoder(
```
cwfitzgerald, line 475: For code reuse purposes, could this be a call to get_encoder, then locking it? The bodies are basically identical.

cwfitzgerald (approved):
Most of this makes sense, have a small comment here and there.

Test looks great, thanks for including it!

wgpu-core/src/command/mod.rs
Wumpf: yeah that would work, but then I'd need to take the lock again. So this would like like:
```rust
    fn lock_encoder(
        hub: &Hub<A>,
        id: id::CommandEncoderId,
    ) -> Result<Arc<Self>, CommandEncoderError> {
        Self::get_encoder(hub, id).map(|result| {
            result.data.lock().as_mut().unwrap().status = CommandEncoderStatus::Locked;
            result
        })
    }
```
this works & passes tests but not quite happy with it :/

Wumpf: decided to solve it with
```rust
    fn get_encoder_impl(
        hub: &Hub<A>,
        id: id::CommandEncoderId,
        lock_on_acquire: bool,
    ) -> Result<Arc<Self>, CommandEncoderError> {
```
[72de2b1](https://github.com/gfx-rs/wgpu/pull/5620/commits/72de2b104af5658b8e0ebdc141ca71be5a263586)

PR #5625, opened 2024-04-29
Fix Failure Case for MacOS 14.3

cwfitzgerald:
MacOS 14 seemed to fix some issues around subgroup op re-convergence, so we have a test now succeeding when it didn't before.

Fixed the cases, but also added some commentary on what's going on.


teoxoy (approved):
LGTM

PR #5627, opened 2024-04-29
[hal] Document resource destruction methods, and a few other things.

jimblandy:
Document some more safety expectations for
- resource destruction methods
- `CommandEncoder` methods
- `Queue::submit`

Document `Fence` creation a bit.

Document the `Queue` trait a bit.

### DIFF
--- a/wgpu-hal/src/lib.rs
+++ b/wgpu-hal/src/lib.rs
@@ -529,6 +529,70 @@ pub trait Adapter: WasmNotSendSync {
     unsafe fn get_presentation_timestamp(&self) -> wgt::PresentationTimestamp;
 }
 
+/// A connection to a GPU and a pool of resources to use with it.
+///
+/// A `wgpu-hal` `Device` represents an open connection to a specific graphics
+/// processor, controlled via the backend [`Device::A`]. A `Device` is mostly
+/// used for creating resources. Each `Device` has an associated [`Queue`] used
+/// for command submission.
+///
+/// On Vulkan a `Device` corresponds to a logical device ([`VkDevice`]). Other
+/// backends don't have an exact analog: for example, [`ID3D12Device`]s and
+/// [`MTLDevice`]s are owned by the backends' [`wgpu_hal::Adapter`]
+/// implementations, and shared by all [`wgpu_hal::Device`]s created from that
+/// `Adapter`.
+///
+/// A `Device`'s life cycle is generally:
+///
+/// 1)  Obtain a `Device` and its associated [`Queue`] by calling
+///     [`Adapter::open`].
+///
+///     Alternatively, the backend-specific types that implement [`Adapter`] often
+///     have methods for creating a `wgpu-hal` `Device` from a platform-specific
+///     handle. For example, [`vulkan::Adapter::device_from_raw`] can create a
+///     [`vulkan::Device`] from an [`ash::Device`].
+///
+/// 1)  Create resources to use on the device by calling methods like
+///     [`Device::create_texture`] or [`Device::create_shader_module`].
+///
+/// 1)  Call [`Device::create_command_encoder`] to obtain a [`CommandEncoder`],
+///     which you can use to build [`CommandBuffer`]s holding commands to be
+///     executed on the GPU.
+///
+/// 1)  Call [`Queue::submit`] on the `Device`'s associated [`Queue`] to submit
+///     [`CommandBuffer`]s for execution on the GPU. If needed, call
+///     [`Device::wait`] to wait for them to finish execution.
+///
+/// 1)  Free resources with methods like [`Device::destroy_texture`] or
+///     [`Device::destroy_shader_module`].
+///
+/// 1)  Shut down the device by calling [`Device::exit`].
+///
+/// [`vkDevice`]: https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkDevice
+/// [`ID3D12Device`]: https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nn-d3d12-id3d12device
+/// [`MTLDevice`]: https://developer.apple.com/documentation/metal/mtldevice
+/// [`wgpu_hal::Adapter`]: Adapter
+/// [`wgpu_hal::Device`]: Device
+/// [`vulkan::Adapter::device_from_raw`]: vulkan/struct.Adapter.html#method.device_from_raw
+/// [`vulkan::Device`]: vulkan/struct.Device.html
+/// [`ash::Device`]: https://docs.rs/ash/latest/ash/struct.Device.html
+/// [`CommandBuffer`]: Api::CommandBuffer
+///
+/// # Safety
+///
+/// As with other `wgpu-hal` APIs, [validation] is the caller's
+/// responsibility. Here are the general requirements for all `Device`
+/// methods:
+///
+/// - Any resource passed to a `Device` method must have been created by that
+///   `Device`. For example, a [`Texture`] passed to [`Device::destroy_texture`] must
+///   have been created with the `Device` passed as `self`.
+///
+/// - Resources may not be destroyed if they are used by any submitted command
+///   buffers that have not yet finished execution.
+///
+/// [validation]: index.html#validation-is-the-calling-codes-responsibility-not-wgpu-hals
+/// [`Texture`]: Api::Texture
 pub trait Device: WasmNotSendSync {
     type A: Api;
 
@@ -721,22 +785,35 @@ pub trait Queue: WasmNotSendSync {
     /// themselves are unordered. If each thread uses a separate [`Fence`], this
     /// problem does not arise.
     ///
-    /// Valid usage:
+    /// # Safety
     ///
-    /// - All of the [`CommandBuffer`][cb]s were created from
-    ///   [`CommandEncoder`][ce]s that are associated with this queue.
+    /// - Each [`CommandBuffer`][cb] in `command_buffers` must have been created
+    ///   from a [`CommandEncoder`][ce] that was constructed from the
+    ///   [`Device`][d] associated with this [`Queue`].
     ///
-    /// - All of those [`CommandBuffer`][cb]s must remain alive until
-    ///   the submitted commands have finished execution. (Since
-    ///   command buffers must not outlive their encoders, this
-    ///   implies that the encoders must remain alive as well.)
+    /// - Each [`CommandBuffer`][cb] must remain alive until the submitted
+    ///   commands have finished execution. Since command buffers must not
+    ///   outlive their encoders, this implies that the encoders must remain
+    ///   alive as well.
     ///
-    /// - All of the [`SurfaceTexture`][st]s that the command buffers
-    ///   write to appear in the `surface_textures` argument.
+    /// - All resources used by a submitted [`CommandBuffer`][cb]
+    ///   ([`Texture`][t]s, [`BindGroup`][bg]s, [`RenderPipeline`][rp]s, and so
+    ///   on) must remain alive until the command buffer finishes execution.
+    ///
+    /// - Every [`SurfaceTexture`][st] that any command in `command_buffers`
+    ///   writes to must appear in the `surface_textures` argument.
+    ///
+    /// - Each [`SurfaceTexture`][st] in `surface_textures` must be configured
+    ///   for use with the [`Device`][d] associated with this [`Queue`],
+    ///   typically by calling [`Surface::configure`].
     ///
     /// [`Fence`]: Api::Fence
     /// [cb]: Api::CommandBuffer
     /// [ce]: Api::CommandEncoder
+    /// [d]: Api::Device
+    /// [t]: Api::Texture
+    /// [bg]: Api::BindGroup
+    /// [rp]: Api::RenderPipeline
     /// [st]: Api::SurfaceTexture
     unsafe fn submit(
         &self,


wgpu-hal/src/lib.rs
```diff
@@ -522,6 +535,60 @@ pub trait Adapter: WasmNotSendSync {
     unsafe fn get_presentation_timestamp(&self) -> wgt::PresentationTimestamp;
 }
 
+/// A connection to a GPU and a pool of resources to use with it.
+///
+/// A `wgpu-hal` `Device` represents an open connection to specific graphics
```
ErichDonGubler: typo:

```suggestion
/// A `wgpu-hal` `Device` represents an open connection to a specific graphics
```

wgpu-hal/src/lib.rs
```diff
@@ -522,6 +535,60 @@ pub trait Adapter: WasmNotSendSync {
...
+/// processor, controlled via the backend [`Device::A`]. This corresponds to a
```
ErichDonGubler: style: It's nice to use the rename-proof `Self`, but definitely a matter of taste:

```suggestion
/// processor, controlled via the backend [`Self::A`]. This corresponds to a
```

wgpu-hal/src/lib.rs
```diff
@@ -522,6 +535,60 @@ pub trait Adapter: WasmNotSendSync {
...
+/// backend object like a Vulkan logical device ([`vkDevice`]) or a Direct3D 12
+/// virtual adapter ([`ID3D12Device`]).
```
ErichDonGubler: question: Won't these links be broken unless the right `--features` are enabled? 🤔

wgpu-hal/src/lib.rs
```diff
@@ -522,6 +535,60 @@ pub trait Adapter: WasmNotSendSync {
...
+///
+/// A `Device`'s life cycle is generally:
+///
+/// - Obtain a `Device` and its associated [`Queue`] by calling
```
ErichDonGubler: nitpick: We should use an ordered list here, since this is an ordered sequence of events.

wgpu-hal/src/lib.rs
```diff
@@ -522,6 +535,60 @@ pub trait Adapter: WasmNotSendSync {
...
+///
+/// - Free resources with methods like [`Device::destroy_texture`] or
+///   [`Device::destroy_shader_module`].
+///
+/// - Shut down the device by calling [`Device::exit`].
+///
+/// [`vkDevice`]: https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkDevice
+/// [`ID3D12Device`]: https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nn-d3d12-id3d12device
+/// [`vulkan::Adapter::device_from_raw`]: vulkan/struct.Adapter.html#method.device_from_raw
+/// [`vulkan::Device`]: vulkan/struct.Device.html
```
ErichDonGubler, line 577: suggestion: You _should_ be able to use the module path here.

```suggestion
```

ErichDonGubler: - [x] issue: Won't links be broken here unless the `vulkan` feature is set?

wgpu-hal/src/lib.rs
```diff
@@ -522,6 +535,60 @@ pub trait Adapter: WasmNotSendSync {
...
+/// [`ash::Device`]: https://docs.rs/ash/latest/ash/struct.Device.html
```
ErichDonGubler, line 578: - [x] issue: It makes me nervous to have `latest` be the specified version here, but I also don't want to tie our current dependency version to correct docs. 😬💦

suggestion: I think we can just use the internal dep. module path instead (i.e., remove this entry from the table entirely)? That might break if the `vulkan` feature is off (like with my previous comment), but that's a potential problem we already have.

wgpu-hal/src/lib.rs
```diff
@@ -522,6 +535,60 @@ pub trait Adapter: WasmNotSendSync {
...
+/// [`CommandBuffer`]: Api::CommandBuffer
+///
+/// # Safety
+///
+/// As usual for `wgpu-hal`, [validation] is the caller's responsibility. Here
```
ErichDonGubler: nitpick: The phrasing "As usual…" doesn't seem appropriate for the target audience, which is to introduce to or remind of to safety considerations when using `wgpu-hal`. I feel like it might be better to be more concrete, like "As with other `wgpu-hal` APIs…".

```suggestion
/// As with other `wgpu-hal` APIs, [validation] is the caller's responsibility. Here
```

wgpu-hal/src/lib.rs
```diff
@@ -576,48 +647,96 @@ pub trait Device: WasmNotSendSync {
...
+    /// Destroy `module`, freeing its associated GPU resources.
+    ///
+    /// It is safe to destroy a shader module while [`RenderPipeline`]s and
+    /// [`ComputePipeline`]s built from it are still in use, and while commands
+    /// using its code are still executing on the GPU.
+    ///
+    /// [`RenderPipeline`]: Api::RenderPipeline
+    /// [`ComputePipeline`]: Api::ComputePipeline
     unsafe fn destroy_shader_module(&self, module: <Self::A as Api>::ShaderModule);
     unsafe fn create_render_pipeline(
```
ErichDonGubler: thought: While we're here, it'd be nice to double-newline-delimit the other methods in this trait, too. Fine to punt on, tho.

wgpu-hal/src/lib.rs
```diff
@@ -576,48 +647,96 @@ pub trait Device: WasmNotSendSync {
...
         &self,
         desc: &ShaderModuleDescriptor,
         shader: ShaderInput,
     ) -> Result<<Self::A as Api>::ShaderModule, ShaderError>;
+
+    /// Destroy `module`, freeing its associated GPU resources.
+    ///
+    /// It is safe to destroy a shader module while [`RenderPipeline`]s and
+    /// [`ComputePipeline`]s built from it are still in use, and while commands
+    /// using its code are still executing on the GPU.
```
ErichDonGubler: praise: TIL, thanks! 😁

wgpu-hal/src/lib.rs
```diff
@@ -576,48 +647,96 @@ pub trait Device: WasmNotSendSync {
...
     unsafe fn create_fence(&self) -> Result<<Self::A as Api>::Fence, DeviceError>;
+
+    /// Destroy `fence`, freeing its associated GPU resources.
+    ///
+    /// # Safety
+    ///
+    /// - There must be no pending events that will write to `fence`. For
+    ///   example, all command submissions for which `fence` was passed to
+    ///   [`Queue::submit`] must have finished execution. that have not yet
+    ///   finished execution.
```
ErichDonGubler: - [x] issue: `finished execution. that have…` is a broken sentence. Maybe you intended to remove everything after `execution.`?

wgpu-hal/src/lib.rs
```diff
@@ -700,20 +827,36 @@ pub trait Queue: WasmNotSendSync {
     ///
     /// Valid usage:
     ///
-    /// - All of the [`CommandBuffer`][cb]s were created from
-    ///   [`CommandEncoder`][ce]s that are associated with this queue.
+    /// - All of the [`CommandBuffer`][cb]s must have been created from
```
ErichDonGubler: - [x] issue: A definite article here is jarring, since no definite selection of a subject happened before it.

```suggestion
    /// - Each [`CommandBuffer`][cb] must have been created from
```

I know this is from older documentation, so I'm happy to iterate in my own PR, if you want.

wgpu-hal/src/lib.rs
```diff
@@ -700,20 +827,36 @@ pub trait Queue: WasmNotSendSync {
...
+    ///   once, even if prior submissions returned errors.
     ///
-    /// - All of those [`CommandBuffer`][cb]s must remain alive until
-    ///   the submitted commands have finished execution. (Since
-    ///   command buffers must not outlive their encoders, this
-    ///   implies that the encoders must remain alive as well.)
+    /// - All of those [`CommandBuffer`][cb]s must remain alive until the
+    ///   submitted commands have finished execution. (Since command buffers
+    ///   must not outlive their encoders, this implies that the encoders must
+    ///   remain alive as well.)
```
ErichDonGubler: nitpick: As the note about lifetime relative to encoders seems _pretty important_, I reject the usage of parentheses, which typically implies tangential information.

```suggestion
    ///   submitted commands have finished execution. Since command buffers
    ///   must not outlive their encoders, this implies that the encoders must
    ///   remain alive as well.
```

I know this is from older documentation, so I'm happy to iterate in my own PR, if you want.

wgpu-hal/src/lib.rs
```diff
@@ -700,20 +827,36 @@ pub trait Queue: WasmNotSendSync {
     ///
     /// Valid usage:
     ///
-    /// - All of the [`CommandBuffer`][cb]s were created from
-    ///   [`CommandEncoder`][ce]s that are associated with this queue.
+    /// - All of the [`CommandBuffer`][cb]s must have been created from
+    ///   [`CommandEncoder`][ce]s that were constructed from the [`Device`][d]
+    ///   associated with this [`Queue`].
+    ///
+    /// - A given [`CommandBuffer`][cb] must only be submitted for execution
```
ErichDonGubler: suggestion: In line with my other suggestion to use `Each…`:

```suggestion
    /// - Each [`CommandBuffer`][cb] must only be submitted for execution
```

I know this is from older documentation, so I'm happy to iterate in my own PR, if you want.

wgpu-hal/src/lib.rs
```diff
@@ -700,20 +827,36 @@ pub trait Queue: WasmNotSendSync {
...
+    ///   once, even if prior submissions returned errors.
     ///
-    /// - All of those [`CommandBuffer`][cb]s must remain alive until
-    ///   the submitted commands have finished execution. (Since
-    ///   command buffers must not outlive their encoders, this
-    ///   implies that the encoders must remain alive as well.)
+    /// - All of those [`CommandBuffer`][cb]s must remain alive until the
```
ErichDonGubler: suggestion: In line with my other suggestion to use `Each…`:

```suggestion
    /// - Each [`CommandBuffer`][cb] must remain alive until the
```

I know this is from older documentation, so I'm happy to iterate in my own PR, if you want.

wgpu-hal/src/lib.rs
```diff
@@ -700,20 +827,36 @@ pub trait Queue: WasmNotSendSync {
...
+    /// - All the resources used by a submitted [`CommandBuffer`][cb]
+    ///   ([`Texture`][t]s, [`BindGroup`][bg]s, [`RenderPipeline`][rp]s, and so
+    ///   on) must remain alive until the command buffer finishes execution.
+    ///
+    /// - All of the [`SurfaceTexture`][st]s in `surface_textures` must be
+    ///   configured for use with the [`Device`][d] associated with this
+    ///   [`Queue`], typically by calling [`Surface::configure`].
+    ///
+    /// - All of the [`SurfaceTexture`][st]s that the command buffers write to
+    ///   must appear in the `surface_textures` argument.
```
ErichDonGubler: style: I think definite articles in reference to resources here is superfluous.

```suggestion
    /// - All resources used by a submitted [`CommandBuffer`][cb]
    ///   ([`Texture`][t]s, [`BindGroup`][bg]s, [`RenderPipeline`][rp]s, and so
    ///   on) must remain alive until the command buffer finishes execution.
    ///
    /// - All [`SurfaceTexture`][st]s in `surface_textures` must be
    ///   configured for use with the [`Device`][d] associated with this
    ///   [`Queue`], typically by calling [`Surface::configure`].
    ///
    /// - All [`SurfaceTexture`][st]s that `command_buffers` write to
    ///   must appear in the `surface_textures` argument.
```

ErichDonGubler (approved):
:shipit:, minus nits I'm trusting to be handled.

wgpu-hal/src/lib.rs
jimblandy: Interesting. I think it's clearer for readers to just say `Device`.

wgpu-hal/src/lib.rs
jimblandy: No - the `vkDevice` and `ID3D12Device` links are both explicitly linked to the corresponding doc web sites. They're not links to Rust code.

wgpu-hal/src/lib.rs
jimblandy: Yes, they will. But docs.rs builds on Linux, so they'll work on the online documentation, and Vulkan is enabled on Windows, so it's only mac users looking at local doc builds that will be disappointed.

What might I do instead? I think from the doc reader's point of view it's okay to mention a specific backend for the sake of giving examples - and it seems bad not to have them be links in the cases where the links *would* work.

wgpu-hal/src/lib.rs
jimblandy: The intent was indeed to have a link that was valid on all platforms. It's really hard to imagine `ash` not having a `Device` item that is the right thing. And the link is only for the sake of an example. Is this really going to be a source of confusion?

wgpu-hal/src/lib.rs
jimblandy: Surprising, isn't it?

wgpu-hal/src/lib.rs
jimblandy: The reason I think parentheses were okay is that this is strictly a consequence of rules stated elsewhere, not new information. But omitting the parens is fine with me.

jimblandy:
@ErichDonGubler I've accepted most of your suggestions, but had some questions.

wgpu-hal/src/lib.rs
jimblandy: The idea of using the http: scheme was to avoid complaints from `cargo doc` even when Vulkan isn't enabled.

wgpu-hal/src/lib.rs
ErichDonGubler: Sigh, I'd noticed this, and forgotten to remove this conversation from my review draft. Sorry for the noise, and thank you for the patient answer!

wgpu-hal/src/lib.rs
ErichDonGubler: > …it seems bad not to have them be links in the cases where the links would work.

To be clear, I think we're in agreement here: it's better to have broken links in some cases than no link at all. 🙂

wgpu-hal/src/lib.rs
```diff
@@ -522,6 +535,61 @@ pub trait Adapter: WasmNotSendSync {
     unsafe fn get_presentation_timestamp(&self) -> wgt::PresentationTimestamp;
 }
 
+/// A connection to a GPU and a pool of resources to use with it.
+///
+/// A `wgpu-hal` `Device` represents an open connection to a specific graphics
+/// processor, controlled via the backend [`Device::A`]. This corresponds to a
+/// backend object like a Vulkan logical device ([`vkDevice`]) or a Direct3D 12
```
cwfitzgerald:
```suggestion
/// backend object like a Vulkan logical device ([`VkDevice`]) or a Direct3D 12
```

wgpu-hal/src/lib.rs
```diff
@@ -522,6 +535,61 @@ pub trait Adapter: WasmNotSendSync {
...
+/// virtual adapter ([`ID3D12Device`]).
```
cwfitzgerald: This is like, quazi true. In practice, the implementation of ExposedAdapter also needs a ID3D12Device, because we need that to extract limit values out of. I'm not sure if that's worth noting here, but it might make reading the d3d12 backend a bit confusing.

cwfitzgerald (approved):
Approved with nits

ErichDonGubler (approved):

wgpu-hal/src/lib.rs
jimblandy: If mentioning D3D12 stuff is confusing, I'll just leave it out.

jimblandy: Actually, since `wgpu_hal::dx12::Device` is definitely an `ID3D12Device` with metadata, I think this paragraph is true as written. The intent here is to give people familiar with one or the other platform API an intuition as to what one does with a `wgpu_hal::Device`, and `ID3D12Device` is the right analog.

So I'm going to leave this as is, and put a comment on `dx12::Adapter::device` to head off confusion.

jimblandy: On chat, cwfitzgerald explained to me that `ID3D12Device`s are singletons per adapter. I also looked through the Metal backend a bit. I think I've got some new text that covers the situation well enough.

jimblandy:
that didn't go well

jimblandy:
Okay, I think I've undone the damage.

PR #5636, opened 2024-04-29
[wgpu] Clarify documentation for `Queue` transfer methods.

jimblandy:
Explain more clearly that the `write_buffer`, `write_buffer_with`, and `write_texture` methods on `wgpu::Queue` do not immediately submit the transfers for execution. Provide sample code for flushing them.

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -4991,11 +4991,24 @@ impl<'a> Drop for QueueWriteBufferView<'a> {
 impl Queue {
     /// Schedule a data write into `buffer` starting at `offset`.
     ///
-    /// This method is intended to have low performance costs.
-    /// As such, the write is not immediately submitted, and instead enqueued
-    /// internally to happen at the start of the next `submit()` call.
-    ///
     /// This method fails if `data` overruns the size of `buffer` starting at `offset`.
+    ///
+    /// This does *not* submit the transfer to the GPU immediately. Calls to
+    /// `write_buffer` begin execution only on the next call to
+    /// [`Queue::submit`]. To get a set of scheduled transfers started
+    /// immediately, it's fine to call `submit` with no command buffers at all:
+    ///
+    /// ```no_run
+    /// # let queue: wgpu::Queue = todo!();
+    /// queue.submit([]);
+    /// ```
+    ///
+    /// However, `data` will be immediately copied into staging memory, so the
+    /// caller may discard it any time after this call completes.
+    ///
+    /// If possible, consider using [`Queue::write_buffer_with`] instead. That
+    /// method avoids an intermediate copy and is often able to transfer data
+    /// more efficiently than this one.
     pub fn write_buffer(&self, buffer: &Buffer, offset: BufferAddress, data: &[u8]) {
         DynContext::queue_write_buffer(
             &*self.context,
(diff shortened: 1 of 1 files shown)


wgpu/src/lib.rs
```diff
@@ -4991,11 +4991,20 @@ impl<'a> Drop for QueueWriteBufferView<'a> {
...
+    /// [`Queue::submit`]. To get a set of scheduled transfers started
+    /// immediately, it's fine to call `submit` with no command buffers at all:
+    ///
+    /// ```no_run
+    /// # let queue: wgpu::Queue = todo!();
+    /// queue.submit(std::iter::empty());
+    /// ```
+    ///
+    /// However, `data` will be immediately copied into staging memory, so the
+    /// caller may discard it any time after this call completes.
```
JMS55, line 5007: I think it's worth mentioning `write_buffer_with()` to avoid a copy from `user_buffer -> wgpu_internal_staging_buffer`. 

I think it's also worth mentioning that this maps to a vkCmdCopyBuffer and similar on the main (only) graphics queue, which takes time (there's not just CPU cost), and can block actual rendering work.

As well, afaik, wgpu does not reuse staging buffers, although the memory allocator wgpu uses might internally reuse the allocation.

---

As a concrete example I used to be uploading ~100 MB/frame of data through write_buffer(), which not only took an insane amount (~18ms) of CPU time due to the copies and buffer allocations and staging buffer creation, but also took ~8.4ms of CPU time in `staging_buffer -> my_buffer` copies due to sheer PCIE bandwidth limits.

ErichDonGubler (approved):
LGTM!

wgpu/src/lib.rs
```diff
@@ -4991,11 +4991,20 @@ impl<'a> Drop for QueueWriteBufferView<'a> {
...
     /// This method fails if `data` overruns the size of `buffer` starting at `offset`.
+    ///
+    /// This does *not* submit the transfer to the GPU immediately. Calls to
+    /// `write_buffer` begin execution only on the next call to
+    /// [`Queue::submit`]. To get a set of scheduled transfers started
+    /// immediately, it's fine to call `submit` with no command buffers at all:
+    ///
+    /// ```no_run
+    /// # let queue: wgpu::Queue = todo!();
+    /// queue.submit(std::iter::empty());
```
kpreid: This sample code could be made shorter:

```suggestion
    /// queue.submit([]);
```

I don't know if there are any relevant performance differences, or if this should be considered good style or not.

ErichDonGubler: +1! IME, there's no performance difference, and if there is, that sounds like a bug against `rustc`. 😅

I'd definitely prefer this form for folks new to Rust and/or graphics. This is simply easier to parse for anybody that's not familiar with `std::iter` (most Rustacean coworkers I've had have seldom used it).

wgpu/src/lib.rs
jimblandy: > I think it's worth mentioning `write_buffer_with()` to avoid a copy from `user_buffer -> wgpu_internal_staging_buffer`. 

Okay, I've done this. I've also added some text to the `write_buffer_with` docs explaining why it's probably faster.

> I think it's also worth mentioning that this maps to a vkCmdCopyBuffer and similar on the main (only) graphics queue, which takes time (there's not just CPU cost), and can block actual rendering work.

Hmm. `wgpu-hal` doesn't expose any queues other than the main one, so there isn't actually a better alternative to suggest if that's not what they want. Since it's not actionable, I'm not sure that this is the best place to get into that level of detail.

> As well, afaik, wgpu does not reuse staging buffers, although the memory allocator wgpu uses might internally reuse the allocation.

I don't really understand the issues here, so I'm not keen to try documenting them in this PR.

wgpu/src/lib.rs
jimblandy: Okay, I've made this change.

Issue #5578, opened 2024-04-22
Add tvOS support


darkwater:
**Is your feature request related to a problem? Please describe.**
I was trying to get Bevy to build for tvOS, and wgpu was the main source of errors.

**Describe the solution you'd like**
Add support for tvOS. Because it's so similar to iOS, this is mostly just changing `cfg(target_os = "ios")` into `cfg(any(target_os = "ios", target_os = "tvos"))`. I already made some progress in [this commit](https://github.com/darkwater/wgpu/commit/f5c82efe77ed7a70f18386e6627de7de908d9b41). I got a Bevy example to build and install, but it would immediately crash upon launch. See [bevy#13058](https://github.com/bevyengine/bevy/issues/13058)

**Describe alternatives you've considered**
Don't support tvOS.

**Additional context**
None

jinleili:
The tvOS target is currently supported as part of the [tier3](https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-3) support level, and only supports [no_std](https://rust-embedded.github.io/book/intro/no-std.html) development

darkwater:
It is indeed tier 3, but doesn't only support no_std. It seems [this PR](https://github.com/rust-lang/rust/pull/103503) added libstd support. Either way, I was able to compile and install a Bevy example. It would crash at some iOS-specific functionality like trying to hide the status bar.

VladasZ:
Hi @darkwater. I want to use `wgpu` with `tvOS` too. 
I tried to add `target = "tvos"` like you did in my `wgpu` fork but now I get this link error:
```
Undefined symbols for architecture arm64:
  "_MTLCopyAllDevices", referenced from:
      metal::device::Device::all::h8012909a44210ef8 in libtest_game.a[3065](metal-d800b64461c8f587.metal.73d4cb208f4c1603-cgu.01.rcgu.o)
  "_objc_msg_lookup", referenced from:
```
Have you seen something like that?

Also I think that you should create a PR with your progress to the main `wgpu` repo. Yes it is not a full support for `tvOS` yet but it is a start and it will help other developers to progress from that and eventually add full `tvOS` support.

madsmtm:
I'd recommend:
- macOS: `cfg(target_os = "macos")`
- iOS/tvOS/watchOS/visionOS/Mac Catalyst: `cfg(all(target_vendor = "apple", not(target_os = "macos")))`
- Metal: `cfg(target_vendor = "apple")`

[...]

PR #3853, opened 2023-06-09
feat: add link feature

crowlKats:
This first requires a release of metal-rs

PR #5641, opened 2024-04-30
[metal] Use `objc2-metal`

madsmtm:
### Description

Use the [`objc2-metal`](https://docs.rs/objc2-metal/latest/objc2_metal/) crate instead of the [`metal`](https://docs.rs/metal/latest/metal/) crate. This will:
- Improve memory management and soundness.
- Make it easier to quickly support new Metal APIs (they're either already generated for you, or is basically just an update of Xcode away).
- Possibly allow reducing the usage of `Arc`, as Metal objects are already reference-counted.
- Likely improve performance, we use `objc_retainAutoreleasedReturnValue` underneath the hood to avoid putting objects into the autorelease pool when possible, reducing memory pressure.
- Make it possible to properly support [tvOS](https://github.com/gfx-rs/wgpu/issues/5578) and visionOS.

### Background

The `metal` crate contains bindings to the Metal framework. This uses [`objc`](https://docs.rs/objc/latest/objc/) to manually perform the message sends, which is quite error-prone, see https://github.com/gfx-rs/metal-rs/issues/284 (a.k.a. #7357), https://github.com/gfx-rs/metal-rs/issues/319 and https://github.com/gfx-rs/metal-rs/issues/209 for a few examples of unsoundness (out of many).

To solve such problems in the Rust ecosystem in general, I created [a successor](https://github.com/SSheldon/rust-objc/issues/101) to `objc` called [`objc2`](https://docs.rs/objc2/latest/objc2/), which contains most notably the smart-pointer [`Retained`](https://docs.rs/objc2/0.6.0/objc2/rc/struct.Retained.html) and an improved [`msg_send!`](https://docs.rs/objc2/0.6.0/objc2/macro.msg_send.html) macro, that together ensure that Objective-C's memory-management rules are upheld correctly.

This is only part of the solution though - we'd still have to write the bindings manually. To solve this, I created a tool (planning to integrate it with `bindgen`, but that's likely a multi-year project) to generate such framework crates automatically. In acknowledgement that this tool is by far not perfect, and never will be, I've ensured that there's a bunch of options to modify each generated crate.

[The modifications for `objc2-metal` in particular](https://github.com/madsmtm/objc2/tree/8d214f5477365ffcbcbb7de058c86ed9a518efb7/framework-crates/objc2-metal/src) are currently just a few hundred lines of code, weak evidence that the generator is fairly good at this point. I'll also bring attention to [the file where `unsafe` methods are marked safe](https://github.com/madsmtm/objc2/blob/8d214f5477365ffcbcbb7de058c86ed9a518efb7/framework-crates/objc2-metal/translation-config.toml) - I have plans to investigate ways to semi-automatically figure out if something is safe or not, or at least reduce the burden of doing so, see https://github.com/madsmtm/objc2/issues/685, but it's a hard problem to ensure is completely sound, so for now it's a bit verbose.

### Connections
`gpu-allocator` has transitioned to `objc2-metal` in https://github.com/Traverse-Research/gpu-allocator/pull/225, and `blade` has transitioned in https://github.com/kvark/blade/pull/229.

https://github.com/gfx-rs/metal-rs/pull/241 is an old open PR for using `objc2` in `metal` internally instead. There currently isn't really a clear path forwards there, and it's a lot of work for less direct benefits to the ecosystem (`wgpu-hal` is by far the biggest user of `metal`). But more fundamentally IMO, it's a problem of separation of concerns; `metal` defines several Foundation types like `NSArray`, `NSString`, `NSError` and so on, and even `CAMetalLayer` from QuartzCore, and that's a bad idea for interoperability compared to having separate crates for each of these frameworks.

https://github.com/gfx-rs/wgpu/pull/5752 removed the `link` feature which is not available in the `objc2` crates.

https://github.com/gfx-rs/wgpu/pull/6210 transitively depends on `objc2`.

### Implementation

The first commit implements the actual migration, by using [a branch](https://github.com/madsmtm/objc2/tree/metal-wgpu) of `objc2`, with a method naming scheme that (more closely) matches `metal`, to make it easier to review and test what's changed. Especially confusing is arguments that are re-ordered between the Objective-C API, and the API that `metal-rs` exposes, see https://github.com/gfx-rs/metal-rs/issues/144.

The second commit moves to the real naming scheme that `objc2` uses.

I'd strongly recommend you review these two commits separately.

### Testing
Tested by using the checklist below, as well as running each example individually, and checking that they _seem_ to work.

During the development of this I made two quite critical typos, which were luckily found by the test suite, but there's bound to be at least one more lurking in here somewhere, please test this thoroughly!

### Checklist

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


wgpu-hal/src/metal/device.rs
```diff
@@ -1272,3 +1312,22 @@ impl crate::Device for super::Device {
...
+
+// TODO: `label` is not exposed on `MTLBuffer`, is this always correct?
+fn set_buffer_label(buffer: &ProtocolObject<dyn MTLBuffer>, label: Option<&NSString>) {
+    unsafe { msg_send![buffer, setLabel: label] }
+}
+
+// TODO: `label` is not exposed on `MTLTexture`, is this always correct?
+fn set_texture_label(buffer: &ProtocolObject<dyn MTLTexture>, label: Option<&NSString>) {
+    unsafe { msg_send![buffer, setLabel: label] }
+}
```
teoxoy: They both inherit from [`MTLResource`](https://developer.apple.com/documentation/metal/mtlresource?language=objc) which has a `label`.

wgpu-hal/src/metal/device.rs
```diff
@@ -1272,3 +1312,22 @@ impl crate::Device for super::Device {
...
+
+// TODO: `newComputePipelineStateWithDescriptor:error:` is not exposed on
+// `MTLDevice`, is this always correct?
+fn new_compute_pipeline_state_with_descriptor(
+    device: &ProtocolObject<dyn MTLDevice>,
+    descriptor: &MTLComputePipelineDescriptor,
+) -> Result<Retained<ProtocolObject<dyn MTLComputePipelineState>>, Retained<NSError>> {
+    unsafe { msg_send_id![device, newComputePipelineStateWithDescriptor: descriptor, error: _] }
+}
```
teoxoy: It's odd that it doesn't exist since `newRenderPipelineStateWithDescriptor:error:` does.
We could use `newComputePipelineStateWithDescriptor:options:reflection:error:` though and pass `MTLPipelineOptionNone` for `options` and `nil` for `reflection`.

teoxoy (commented):
I think this is a great step forward from having to maintain our own set of bindings manually.

I do have some questions/comments:

- We still seem to be using the `msg_send!` macro in a bunch of places, could we use the generated methods instead?
- We also use `autoreleasepool`s in a bunch of places because we had issues with leaks but as far as I understand from the docs of `Retained`, a lot of these (if not all) should now be unnecessary. Is this correct? I think it would make sense to remove them in this PR.
- While I appreciate the naming being more consistent with metal's, some of the enum variants now have the enum name as their prefix which feels redundant; but not always which feels odd. Examples:
  - `MTLFeatureSet`'s variants are not prefixed
  - `MTLLanguageVersion`'s variants are all prefixed
  - one of `MTLReadWriteTextureTier`'s variants is not prefixed while the other 2 are)
- Some of the function names are quite long now (ex: `copyFromTexture_sourceSlice_sourceLevel_sourceOrigin_sourceSize_toBuffer_destinationOffset_destinationBytesPerRow_destinationBytesPerImage_options` 😆) but I'm not sure how they can be shortened by the generator while also keeping things easy to search for on Apple's docs.
- It would be great if all objects had docs from Apple's docs website and/or at least a link to the page on said website but no hurry, just giving my +1 for it :).

crowlKats:
@madsmtm 

> The first commit removes the link feature which is not available in these crates. This was added by @crowlKats in https://github.com/gfx-rs/wgpu/pull/3853, but Deno doesn't actually use it from what I can tell, [they instead specify this using -weak_framework](https://github.com/denoland/deno/blob/8a636d0600dc7000d1e12b2fc4f4f46ecd70164a/.cargo/config.toml#L14-L24), which is the correct solution to this problem IMO. (If I'm wrong about this, please say so, I could be persuaded to add a similar feature to the objc2-... crates).

so we did use it for a while, but then were able to remove the use on it, so yes, its not necessary anymore

wgpu-hal/src/metal/device.rs
madsmtm: Right, I didn't think to import that, fixed now

madsmtm:
Thanks for taking a look!

> We still seem to be using the `msg_send!` macro in a bunch of places, could we use the generated methods instead?

I've already done this in a separate branch, but there were some issues around the semantics not being quite the same, so I wanted to keep it for a separate PR where that discussion could be more focused.

> We also use `autoreleasepool`s in a bunch of places because we had issues with leaks but as far as I understand from the docs of `Retained`, a lot of these (if not all) should now be unnecessary. Is this correct? I think it would make sense to remove them in this PR.

There's a lot of nuance to this question, including the fact that the optimization that allows us to avoid an autorelease is not guaranteed, only likely (depends on the exact emitted assembly, inlining, and the phase of the moon).

What _has_ changed though is that autorelease pools no longer have any _effect_ on the program, other than in terms of memory usage (autoreleased objects can be reclaimed sooner) and runtime performance (pushing and popping an autorelease pool has a slight overhead).

I'd really prefer to keep it out of this PR, mostly because it'll screw with the diff even more than it already is (indentation of large function bodies will change), but also partly because I do not know why each of these are here, and I'd like to retain that memory-usage profile until deemed wise to do otherwise.

> While I appreciate the naming being more consistent with metal's, some of the enum variants now have the enum name as their prefix which feels redundant; but not always which feels odd. Examples:
>       
> * `MTLFeatureSet`'s variants are not prefixed
> * `MTLLanguageVersion`'s variants are all prefixed
> * one of `MTLReadWriteTextureTier`'s variants is not prefixed while the other 2 are)

Yup, that's a bug, the logic for implementing this name translation is very simplistic and a bit hastily thrown together - Swift has [the correct rules written down](https://github.com/apple/swift/blob/swift-5.9.2-RELEASE/docs/CToSwiftNameTranslation.md#enum-style-prefix-stripping), but it'll be breaking to change, so I'll fix it in `v0.3` of the framework crates.

> Some of the function names are quite long now (ex: `copyFromTexture_sourceSlice_sourceLevel_sourceOrigin_sourceSize_toBuffer_destinationOffset_destinationBytesPerRow_destinationBytesPerImage_options` 😆) but I'm not sure how they can be shortened by the generator while also keeping things easy to search for on Apple's docs.

Yeah :/

Feel free to comment on https://github.com/madsmtm/objc2/issues/284 if you think of a good solution (or just any solution)!

> It would be great if all objects had docs from Apple's docs website and/or at least a link to the page on said website but no hurry, just giving my +1 for it :).

There's https://github.com/madsmtm/objc2/issues/309 open for it, the local Xcode documentation is stored in an undocumented format that I spent a few hours on, but couldn't immediately reverse-engineer.

Linking is similarly difficult, not for class names, but for methods, they have some ID in them, which is why I didn't pursue this - but I guess just linking to the class name would still be a step up, will try to prioritize it.

madsmtm:
I just tried to actually benchmark this, but it's not an art that I'm familiar with, and the results seemed inconclusive (both improvements and regressions, without any clear pattern that I could discern). I'd suggest that someone more familiar with that tries to benchmark this.

grovesNL:
@madsmtm For what it's worth we used to see a huge amount of time in `retain`/`release` for a lot of the sampling profilers we tried years ago, so reference counting overhead could be a useful data point.

madsmtm:
> > We still seem to be using the `msg_send!` macro in a bunch of places, could we use the generated methods instead?
>
> I've already done this in a separate branch, but there were some issues around the semantics not being quite the same, so I wanted to keep it for a separate PR where that discussion could be more focused.

I've opened https://github.com/gfx-rs/wgpu/pull/6107 to fix the semantic issues, btw. Whichever of these two PRs merge first, I'll update the other to remove the remaining `msg_send!`.

jimblandy:
This PR has been open since the spring, so I want to let folks know what's going on with it.

We (the wgpu maintainers) think `objc2` is the future of Objective-C bindings for Rust, and wgpu should use it, as soon as possible. The emphasis on soundness and idiomatic Rust are music to our ears. I'm really grateful to Mads for coming in and writing this PR; it's a very valuable contribution. Everything has to be reviewed, of course, but based on looking at a bit of his prior work, I'm expecting this PR to be very high quality.

Firefox has been incorporating wgpu `trunk` roughly weekly into our development source tree, Mozilla Central, which is the basis for Nightly Firefox releases. We (Firefox) don't import from some separate branch, because a delta between what Firefox sees and what wgpu's direct users see means a slip in collaboration and an increase in risk.

Firefox uses a supply-chain auditing tool called [`cargo vet`](https://mozilla.github.io/cargo-vet/), which ensures that *every line of Rust source* that we incorporate into Firefox from public crates has been audited by someone who understands Firefox's security constraints. (For generated code, Mozilla audits the generator, and if its output is checked in, we verify that the output is actually what the generator produces.) There's a lot of grandfathered Rust code in Firefox that hasn't been audited, but yes, for the last few years, we actually really have looked at every change to every random upstream crate, because, well, there's no responsible alternative. `cargo vet` has features that allow us to share the auditing load with Google and a few other trusted orgs. Other groups are free to publish their own audit lists, but incorporating those is a pretty serious policy decision that requires looking into their processes, track record, and so on.

But this means that taking this PR means that Firefox cannot update the wgpu sources in Mozilla Central until we have audited every line of `objc2-metal` and its dependencies (`block2`, etc.) This is our only process, at the moment, for getting bug fixes in wgpu-core into Firefox, so if we take this PR, then Firefox is dead in the water until that vetting is done. Obviously, we'd prefer simply to audit the objc2 crates in advance, land this PR, and then do delta audits to cover any changes between that first audit and whatever gets incorporated.

The next steps for this PR are:

- Get it audited for `cargo vet`. The Firefox WebGPU team is only three full-time people. However, I've started poking around within the org to get some additional people with the necessary Objective-C and Rust background to help us with this, and things are looking promising.

- Finish reviewing this PR. In a few days, @teoxoy is going to be on PTO for a few weeks, but I'm going to find someone else to do it. (I'm seriously behind in my review obligations already, so if I were to claim I was going to do it myself, people would just laugh.)

What I'm hoping to get across here is that, despite the delay, this PR is important to us, and although we have some very stern constraints, we're moving the process along.

madsmtm:
Thank you very much for the update @jimblandy! It's really nice to see that you're taking supply chain attacks seriously, it's something we very much need to be better at in our industry!

Regarding the ability to audit/review [the `objc2` codebase](https://github.com/madsmtm/objc2/tree/objc-0.5.2), I recognize that I'm the person "under scrutiny" so to say, but is there anything I can do to make things easier for you?

Code-wise, I can give a quick primer:
- The only thing running on the host machine is [a small build script in `objc-sys`](https://github.com/madsmtm/objc2/blob/objc-0.5.2/crates/objc-sys/build.rs), there's no proc-macros (yet).
- The hardest part to review is probably the `objc2` crate itself, and [the declarative macros therein](https://github.com/madsmtm/objc2/tree/objc-0.5.2/crates/objc2/src/macros), it consists of hacks, hacks and nothing but hacks. I apologize in advance to whoever has to look at it.
- Most of the code in the framework crates, for Wgpu's case `objc2-metal`, `objc2-quartz-core` and `objc2-foundation`, is autogenerated ahead of time, put in [a separate `git` repo](https://github.com/madsmtm/objc2-generated/tree/v0.2.2), and symlinked into each crate. You can verify the generated code by removing it, [re-generating it with `cargo run --bin header-translator`](https://github.com/madsmtm/objc2/tree/objc-0.5.2/crates/header-translator) (requires, currently, Xcode 15.4 to be the active developer directory), and checking `git status`.
- `block2`, `objc2-encode` and the manual fixes/additions in the framework crates should be comparatively easy to review.

Feel free to ask if there's something here that you need help with, or if there's something I can change to make things easier to audit (now and in the future)!

On the social/people side, would it help if we had an online meeting or something? Though I guess that could be construed as a form of social engineering attack, and wouldn't really add anything trust-wise.

In any case, thanks again for the update, and I'm totally fine with waiting a few more months, or however long it takes you!

jimblandy:
In our meeting today, the wgpu maintainers decided to mark this as draft, since we'll be getting back to this in 2025.

madsmtm:
Small update: I've released a new version of `objc2` and the framework crates, and rebased the PR, which fixes the enum variant names and added links to Apple's documentation on top-level types, as noted in https://github.com/gfx-rs/wgpu/pull/5641#pullrequestreview-2073499942.

I also tried running the benchmarks again, seems to be better than last time I tried it, now all changes are within ~2% of the current `trunk` - but still very unsure about this, someone better suited for that should verify!

madsmtm:
FYI, since https://github.com/gfx-rs/wgpu/pull/7026 landed, I have rebased this on top of https://github.com/gfx-rs/wgpu/pull/6210, so the first two commits actually belong to that.

madsmtm:
Made the diff a bit smaller by filing https://github.com/gfx-rs/wgpu/pull/7631 first.

wgpu-hal/src/metal/command.rs
```diff
@@ -60,47 +74,48 @@ impl super::CommandEncoder {
                 .contains(TimestampQuerySupport::ON_BLIT_ENCODER);
 
             if !self.state.pending_timer_queries.is_empty() && !supports_sample_counters_in_buffer {
-                objc::rc::autoreleasepool(|| {
-                    let descriptor = metal::BlitPassDescriptor::new();
+                autoreleasepool(|_| unsafe {
```
ErichDonGubler: nitpick: Can we please narrow the spans that this `unsafe` block applies to? If you're aware of the rationale, a `SAFETY` comment would be great, but I know that in other places you've disclaimed knowledge of the Metal API specifically.

wgpu-hal/src/metal/command.rs
```diff
@@ -348,15 +366,15 @@ impl crate::CommandEncoder for super::CommandEncoder {
                 // the amount of data to copy.
                 0
             };
-            encoder.copy_from_buffer_to_texture(
+            encoder.copyFromBuffer_sourceOffset_sourceBytesPerRow_sourceBytesPerImage_sourceSize_toTexture_destinationSlice_destinationLevel_destinationOrigin_options(
```
ErichDonGubler: thought: Oh wow, this (what I presume to be an) overload name is _awful_. 😆

wgpu-hal/src/metal/device.rs
```diff
@@ -41,27 +49,29 @@ fn create_stencil_desc(
...
-) -> metal::StencilDescriptor {
-    let desc = metal::StencilDescriptor::new();
-    desc.set_stencil_compare_function(conv::map_compare_function(face.compare));
-    desc.set_read_mask(read_mask);
-    desc.set_write_mask(write_mask);
-    desc.set_stencil_failure_operation(conv::map_stencil_op(face.fail_op));
-    desc.set_depth_failure_operation(conv::map_stencil_op(face.depth_fail_op));
-    desc.set_depth_stencil_pass_operation(conv::map_stencil_op(face.pass_op));
+) -> Retained<MTLStencilDescriptor> {
+    let desc = unsafe { MTLStencilDescriptor::new() };
```
ErichDonGubler: thought: It's sorta weird that constructing a new one of these is `unsafe`… 🤔 What preconditions could possibly need to be upheld? I say this as an Apple platform programming newbie.

wgpu-hal/src/metal/adapter.rs
```diff
@@ -91,6 +100,8 @@ impl crate::Adapter for super::Adapter {
             MTLReadWriteTextureTier::TierNone => (Tfc::empty(), Tfc::empty()),
             MTLReadWriteTextureTier::Tier1 => (Tfc::STORAGE_READ_WRITE, Tfc::empty()),
             MTLReadWriteTextureTier::Tier2 => (Tfc::STORAGE_READ_WRITE, Tfc::STORAGE_READ_WRITE),
+            // Fall back to `MTLReadWriteTextureTier2`
```
ErichDonGubler: nitpick: Can we note that this behavior feels appropriate because it's likely a _higher_ level of support, please?

wgpu-hal/src/metal/device.rs
```diff
@@ -1611,3 +1658,12 @@ impl crate::Device for super::Device {
...
 }
+
+// TODO: `newComputePipelineStateWithDescriptor:error:` is not exposed on
+// `MTLDevice`, is this always correct?
+fn new_compute_pipeline_state_with_descriptor(
+    device: &ProtocolObject<dyn MTLDevice>,
+    descriptor: &MTLComputePipelineDescriptor,
+) -> Result<Retained<ProtocolObject<dyn MTLComputePipelineState>>, Retained<NSError>> {
+    unsafe { msg_send![device, newComputePipelineStateWithDescriptor: descriptor, error: _] }
+}
```
ErichDonGubler: nitpick: I'd prefer that this be inlined to its single use site; the comment is welcome, and we should validate it (CC @gfx-rs/wgpu), but the function signature simply doesn't seem to carry its weight.

wgpu-hal/src/metal/mod.rs
```diff
@@ -12,6 +12,9 @@ resources, followed by other bind groups. The vertex buffers are bound at the ve
 end of the VS buffer table.
 
 !*/
+// Avoid noise, many objc2-metal functions are still `unsafe`.
+// See also https://github.com/madsmtm/objc2/issues/685.
+#![allow(unsafe_op_in_unsafe_fn)]
```
ErichDonGubler: thought: IMO, it's okay to have this noise, unless you're _really_ convinced that most of it will fall away. 🙂

wgpu-hal/src/metal/mod.rs
```diff
@@ -124,30 +139,47 @@ impl crate::Instance for Instance {
...
+        // MTLCopyAllDevices is always available on macOS and Mac Catalyst,
+        // but only available recently on iOS 18.0 / tvOS 18.0 / visionOS 2.0.
+        // Instead, we use MTLCreateSystemDefaultDevice on those platforms.
+        //
+        // TODO: Call MTLCopyAllDevices once objc2-metal does this internally:
+        // https://github.com/madsmtm/objc2/commit/35439408e684adddc54d7ce954ae2661716d9ef8
+        #[cfg(any(target_os = "macos", target_abi = "macabi"))]
+        let devices = objc2_metal::MTLCopyAllDevices(); // NSArray -> into_iter
+        #[cfg(not(any(target_os = "macos", target_abi = "macabi")))]
+        let devices = objc2_metal::MTLCreateSystemDefaultDevice(); // Option -> into_iter
```
ErichDonGubler: question: Is there a way to track this via an issue?

wgpu-hal/src/metal/surface.rs
```diff
@@ -34,87 +23,16 @@ impl super::Surface {
...
-    /// If not called on the main thread, this will panic.
-    #[allow(clippy::transmute_ptr_to_ref)]
-    pub unsafe fn from_view(view: NonNull<Object>) -> Self {
-        let layer = unsafe { Self::get_metal_layer(view) };
-        let layer = ManuallyDrop::new(layer);
-        // SAFETY: The layer is an initialized instance of `CAMetalLayer`, and
-        // we transfer the retain count to `MetalLayer` using `ManuallyDrop`.
-        let layer = unsafe { metal::MetalLayer::from_ptr(layer.cast()) };
-        Self::new(layer)
-    }
```
ErichDonGubler: question: Pinging @gfx-rs/wgpu here: is it problematic if `metal::Surface::from_view` goes away WRT user-exposed functionality?

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -904,17 +901,19 @@ impl crate::Instance for super::Instance {
...
+            #[cfg(target_vendor = "apple")]
             (Rwh::AppKit(handle), _)
                 if self.shared.extensions.contains(&ext::metal_surface::NAME) =>
             {
-                self.create_surface_from_view(handle.ns_view)
+                let layer = unsafe { raw_window_metal::Layer::from_ns_view(handle.ns_view) };
+                self.create_surface_from_layer(layer)
             }
-            #[cfg(all(any(target_os = "ios", target_os = "visionos"), feature = "metal"))]
+            #[cfg(target_vendor = "apple")]
```
ErichDonGubler, line 913: question: How is it correct to move from `target_os` checks against subsets of the Apple hardware ecosystem to `target_vendor = "apple"`? 🤔

ErichDonGubler (commented):
Alright, I think this thing's been open long enough, and needs some movement. I think it's easy to say that the vast majority of this code doesn't feel risky. There are some things that seem unexpected, but minor to resolve. I've filed these as individual conversations, and pinged attention from other maintainers where I thought it'd be helpful.

Once these are resolved, I'm inclined to approve, based on conversations here being resolved to everyone's satisfaction. After this, the only blocker to merging will be auditing the `objc2` ecosystem for its `unsafe` usage, to satisfy Mozilla's supply chain processes. This appears to be _very_ non-trivial, and I don't have bandwidth to do this myself, but hey, at least nobody has to go over the entire review again!

CC @jimblandy.

wgpu-hal/src/metal/device.rs
cwfitzgerald: Generally unsafe new means something about dropping being unsafe/unsound.

wgpu-hal/src/metal/mod.rs
cwfitzgerald: To this end, we actually put this lint on intentionally and migrated all of hal to use it before it was enabled by default, so we are quite happy with having unsafe around.

wgpu-hal/src/metal/surface.rs
cwfitzgerald: Yes, it is likely to be an issue

wgpu-hal/src/metal/command.rs
madsmtm: Yeah, if you have suggestions for better ways of doing this naming, feel free to drop them in https://github.com/madsmtm/objc2/issues/284.

wgpu-hal/src/metal/mod.rs
madsmtm: It was fixed in v0.3.1, just forgot to update it here.

wgpu-hal/src/metal/device.rs
madsmtm: I've inlined it now.

wgpu-hal/src/metal/mod.rs
madsmtm: I am fairly confident that a lot of the `unsafe` will disappear, and so I felt that adding that line was valuable, since it made the already large diff easier to review.

Though I guess adding `unsafe` now would make it easier for you to migrate once the things are marked safe? Another option would be to open a new tracking issue and assigning me to it.

Regardless, I'll try to prioritize this work in `objc2`.

wgpu-hal/src/metal/device.rs
madsmtm: Mostly same as in https://github.com/gfx-rs/wgpu/pull/5641#discussion_r2066231610: this `unsafe` will likely disappear some day.

wgpu-hal/src/metal/command.rs
madsmtm: Reason I did it this way is mostly the same as in https://github.com/gfx-rs/wgpu/pull/5641#discussion_r2066231610: this `unsafe` will likely disappear some day, and it kept the diff smaller.

wgpu-hal/src/metal/surface.rs
madsmtm: Ah, this is public API, didn't realize that since it isn't present on docs.rs.

I'll see what I can do about keeping some form of it around.

wgpu-hal/src/vulkan/instance.rs
madsmtm: There are a bunch of nuances to the relationship between UIKit/AppKit and the different Apple OSes. One of them is that `all(target_os = "ios", target_abi = "macabi")` actually also supports AppKit, so if Wgpu wants to be maximally usable, you'd actually have to use `#[cfg(any(target_os = "macos", all(target_os = "ios", target_abi = "macabi")))]` on the `AppKit` variant.

Instead of handling all that, we delegate to the producer of the handle; if Winit or similar produced an `AppKitWindowHandle`, we trust that the `ns_view` field in there really is an `NSView`, regardless of which Apple platform we're currently running on. `raw-window-metal` is also designed to be "platform agnostic" in the sense that it's APIs are available on all Apple platforms, regardless of whether `UIView`/`NSView` makes sense for that specific platform.

madsmtm:
Thanks for the review Erich.

> After this, the only blocker to merging will be auditing the `objc2` ecosystem for its `unsafe` usage, to satisfy Mozilla's supply chain processes. This appears to be _very_ non-trivial, and I don't have bandwidth to do this myself, but hey, at least nobody has to go over the entire review again!

Again, do feel free to say if there's something I can do to make that process easier!

wgpu-hal/src/metal/surface.rs
madsmtm: Have opened https://github.com/gfx-rs/wgpu/pull/7649 for this, will rebase once that lands.

ErichDonGubler:
In case it helps: After today's WGPU maintainers' meeting, I attempted to vendor in the tip of PR after rebasing (which I'll push shortly). I have a list of unaudited crates that would be needed to please `mach cargo vet` in `mozilla-central`:

- `block2` 0.6.1
- `dispatch2` 0.3.0
- `objc2` 0.6.1
- `objc2-core-foundation` 0.3.1
- `objc2-encode` 4.1.0
- `objc2-foundation` 0.3.1
- `objc2-metal` 0.3.1
- `objc2-quartz-core` 0.3.1
- `raw-window-metal` 1.1.0

ErichDonGubler:
Rebased to keep up with some of the changes being made.

jimblandy:
Mozilla's getting ready to start working on vetting objc2, since we want to release on Mac in 2025H2 (ssh, don't tell anyone).

wgpu-hal/src/metal/device.rs
MarijnS95: I've started marking a bunch of these safe in https://github.com/madsmtm/objc2/pull/773 based on Metal 4 porting. I have yet to go through our engine / the full bindings in general to identify more simple things like constructors and getters/setters, and will be sure to include `MTLStencilDescriptor` at least :tada:

madsmtm:
I just opened https://github.com/madsmtm/objc2/pull/783 for automatically marking `objc2-foundation` as safe. I would _really_ like to get your opinions on this PR, since it marks a policy change in `objc2-*` crates that will affect you.

If the community agrees with the policy change, I plan to open a similar PR for marking Metal APIs safe.

teoxoy:
I think it's the pragmatic thing to do. In the few cases where the functions have extra safety requirements only outlined in the docs users can open issues to mark the functions as unsafe.

madsmtm:
I have now opened https://github.com/madsmtm/objc2/pull/792 for marking a lot of APIs safe in `objc2-metal` that I talked about.

It's likely I'll be merging it rather quickly (I'm behind schedule on `v0.3.2`, and I'd like to get this out with that), but I'd appreciate if you could take a look at the reasoning I've written down in there (regardless of it being merged by the time you take a look).

wgpu-hal/src/metal/command.rs
madsmtm: I've made all the `unsafe` blocks smaller scoped now that [`objc2-metal` is marked safe in more places](https://github.com/madsmtm/objc2/pull/792).

wgpu-hal/src/metal/device.rs
madsmtm: The `unsafe` blocks in this PR are smaller scoped now that [`objc2-metal` is marked safe in more places](https://github.com/madsmtm/objc2/pull/792).

wgpu-hal/src/metal/mod.rs
madsmtm: After https://github.com/madsmtm/objc2/pull/792, a lot of it did fall away, though unfortunately not as much as I'd have liked, we're still working on that in https://github.com/madsmtm/objc2/issues/793. I have removed this `#![allow(unsafe_op_in_unsafe_fn)]` now though, and made the unsafe blocks more local instead.

madsmtm:
I have now released `objc2-metal v0.3.2`, it contains the improvements to safety discussed above, I have updated this PR to use those.

wgpu-hal/src/metal/device.rs
madsmtm: We should, but I'd rather leave it for a different PR, and retain the exact methods that we're calling in this PR.

andyleiserson:
I rebased this on latest trunk, and pushed it to https://github.com/gfx-rs/wgpu/compare/trunk...andyleiserson:wgpu:objc2-metal

madsmtm:
> I rebased this on latest trunk, and pushed it to [trunk...andyleiserson:wgpu:objc2-metal](https://github.com/gfx-rs/wgpu/compare/trunk...andyleiserson:wgpu:objc2-metal)

Thanks! I used that to rebase on top of the current latest `trunk`.

wgpu-hal/src/metal/conv.rs
```diff
@@ -341,6 +342,7 @@ pub fn map_render_stages(stage: wgt::ShaderStages) -> MTLRenderStages {
 
 pub fn map_resource_usage(ty: &wgt::BindingType) -> MTLResourceUsage {
     match ty {
+        #[allow(deprecated)]
         wgt::BindingType::Texture { .. } => MTLResourceUsage::Sample,
```
andyleiserson, line 346: Filed #8940 to review the deprecation allowances added in this PR. (There's no change to behavior, it's just accurately reflecting the deprecation status of the APIs, but I still felt it might be worth revisiting, e.g. should we map `BindingType::Texture` to `ResourceUsage::Read`?)

wgpu-hal/src/metal/mod.rs
```diff
@@ -875,6 +830,25 @@ struct PipelineStageInfo {
     work_group_memory_sizes: Vec<u32>,
 }
 
+// TODO(madsmtm): Derive this once https://github.com/madsmtm/objc2/issues/804 is resolved.
```
andyleiserson:
```suggestion
// TODO(madsmtm): Derive this when a release with https://github.com/madsmtm/objc2/issues/804
// is available (likely 0.4).
```

wgpu-hal/src/metal/mod.rs
```diff
@@ -815,11 +767,11 @@ impl Default for UseResourceInfo {
...
     counters: MultiStageResourceCounters,
     buffers: Vec<BufferResource>,
-    samplers: Vec<SamplerPtr>,
-    textures: Vec<TexturePtr>,
+    samplers: Vec<NonNull<ProtocolObject<dyn MTLSamplerState>>>,
+    textures: Vec<NonNull<ProtocolObject<dyn MTLTexture>>>,
 
-    argument_buffers: Vec<metal::Buffer>,
-    resources_to_use: HashMap<ResourcePtr, UseResourceInfo>,
+    argument_buffers: Vec<Retained<ProtocolObject<dyn MTLBuffer>>>,
```
andyleiserson, line 773: It looks like this one is `Retained` because this is our primary reference to the argument buffer (and in fact keeping the buffer alive may be the only purpose of stashing it here), while the others are raw pointers because their liveness is managed by `wgpu-core` and we don't want unnecessary reference counting overhead?

This might be worth a comment, but that could be done as a follow-up (it's not changed by this PR, I just wanted to understand what was going on).

wgpu-hal/src/metal/device.rs
```diff
@@ -1571,14 +1642,18 @@ impl crate::Device for super::Device {
...
+                descriptor.setLabel(Some(&NSString::from_str(name)));
             }
 
-            let raw = self
-                .shared
-                .device
-                .new_compute_pipeline_state(&descriptor)
-                .map_err(|e| {
+            // TODO: `newComputePipelineStateWithDescriptor:error:` is not exposed on
+            // `MTLDevice`, is this always correct?
```
andyleiserson:
```suggestion
            // TODO(https://github.com/gfx-rs/wgpu/issues/8944):
            // `newComputePipelineStateWithDescriptor:error:` is not exposed on
            // `MTLDevice`, is this always correct?
```

wgpu-hal/src/metal/device.rs
```diff
@@ -1451,32 +1517,40 @@ impl crate::Device for super::Device {
...
-                    self.shared.device.new_render_pipeline_state(&d)
-                }
+                MetalGenericRenderPipelineDescriptor::Standard(d) => self
+                    .shared
+                    .device
+                    .newRenderPipelineStateWithDescriptor_error(&d),
                 MetalGenericRenderPipelineDescriptor::Mesh(d) => {
-                    self.shared.device.new_mesh_render_pipeline_state(&d)
+                    // TODO: `newRenderPipelineStateWithMeshDescriptor:error:` is not exposed on
+                    // `MTLDevice`, is this always correct?
```
andyleiserson:
```suggestion
                    // TODO(https://github.com/gfx-rs/wgpu/issues/8944):
                    // `newRenderPipelineStateWithMeshDescriptor:error:` is not exposed on
                    // `MTLDevice`, is this always correct?
```

wgpu-hal/src/metal/command.rs
```diff
@@ -968,66 +1040,77 @@ impl crate::CommandEncoder for super::CommandEncoder {
         let offset_words = offset_bytes as usize / WORD_SIZE;
         state_pc[offset_words..offset_words + data.len()].copy_from_slice(data);
 
+        let bytes = NonNull::new(state_pc.as_ptr().cast_mut().cast()).unwrap();
```
andyleiserson, line 1043: [Looks like we're about to get Rust 1.90](https://github.com/gfx-rs/wgpu/pull/8943), which I think enables simplifying this mouthful to `NonNull::from_ref(state_pc).cast()`. I made a note to come back to this.

andyleiserson (approved):

wgpu-hal/src/metal/command.rs
madsmtm: Yeah, I'm also not a fan of this, the `NonNull` APIs can be quite a mouthful.

I have some vague plans for mapping `setBytes:length:atIndex:` as something like `unsafe fn setBytes_length_atIndex(bytes_and_length: NonNull<[u8]>, index: usize)`, that would allow it to be just:
```rust
let bytes = bytemuck::cast_slice::<u32, u8>(&state_pc);
let bytes = NonNull::from_ref(bytes);
compute.setBytes_length_atIndex(bytes, layout.immediates_infos.cs.unwrap().buffer_index as usize);
```

wgpu-hal/src/metal/mod.rs
madsmtm: I think that's the case, yeah, the only purpose of `argument_buffers` is to make sure that the buffer is kept alive. IMO `BufferResource` should just store `Retained<ProtocolObject<dyn MTLBuffer>>`, but yeah, I didn't want to make such changes in this PR.

Issue #5639, opened 2024-04-30
sRGB seemingly applied to non-sRGB render targets with 0.20 GL backend


vorporeal:
**Description**

Upgraded to the 0.20 release and noticed that, when using the GL backend with a non-sRGB texture format (specifically, `Rgba8Unorm`), colors render as though an sRGB texture format was selected.

Specifically, the rendering is different from Vulkan with the same texture format, but the _same_ as Vulkan when using `Bgra8UnormSrgb` (can't do a direct comparison due to adapter support).

**Repro steps**

Will try to repro with a `wgpu` example; wanted to report this as soon as I found it so y'all were aware.

**Expected vs observed behavior**

I expect the same color representation between the Vulkan and GL backends when using non-sRGB texture formats.

**Screenshots**

Vulkan w/ `Bgra8Unorm` (expected):

![image](https://github.com/gfx-rs/wgpu/assets/211194/2a5bd2bf-0390-49df-95c8-89887f13953f)

OpenGL w/ `Rgba8Unorm` (unexpected):

![image](https://github.com/gfx-rs/wgpu/assets/211194/14befa97-e87a-4174-8165-a4928987714c)

Vulkan w/ `Bgra8UnormSrgb`:

![image](https://github.com/gfx-rs/wgpu/assets/211194/7ac2f4ba-01d5-4465-80d4-6a7eb027dc08)

**Platform**

Linux $HOSTNAME 6.8.7-arch1-2 #1 SMP PREEMPT_DYNAMIC Fri, 19 Apr 2024 09:51:31 +0000 x86_64 GNU/Linux

Adapters:

> IntegratedGpu: Intel(R) Graphics (RPL-P)
> Backend: Vulkan
> Driver: Intel open-source Mesa driver (Mesa 24.0.6-arch1.2)

> IntegratedGpu: Mesa Intel(R) Graphics (RPL-P)
> Backend: Gl
> Driver: OpenGL (4.6 (Compatibility Profile) Mesa 24.0.6-arch1.2)

rustc 1.76.0
`wgpu` 0.20.0
`winit` 0.30.0

vorporeal:
I'm suspicious of the changes in https://github.com/gfx-rs/wgpu/pull/5482, where issues with sRGB were discussed during the review process.

Wumpf:
@valaphee any chance you could have a look at this? :)

valaphee:
Sure. (maybe I just should have copied the sRGB handling from wgl ^^')

[...]

PR #5642, opened 2024-05-01
Fix regression on OpenGL (EGL) where non-sRGB still used sRGB

valaphee:
**Connections**
https://github.com/gfx-rs/wgpu/issues/5639

**Description**
Its probably worth to also include it in a patch release.

**Testing**
Setting https://github.com/gfx-rs/wgpu/blob/trunk/examples/src/framework.rs#L13 (SRGB) to false, and it should use the non-srgb view format, and therefore look darker.

**Checklist**

- [x] Run `cargo fmt`.
- [ x Run `cargo clippy`. If applicable, add:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,10 @@ By @stefnotch in [#5410](https://github.com/gfx-rs/wgpu/pull/5410)
 
 ### Bug Fixes
 
+#### GLES / OpenGL
+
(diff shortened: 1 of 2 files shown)


vorporeal:
Can confirm that this fixes the issue I reported. :)

Wumpf (approved):
thanks for following up on this!
a bit annoying that we duplicate this between egl.rs and wgl.rs.

PR #5646, opened 2024-05-02
[core] Ensure all lock guards are properly nested.

### LINKED ISSUE #5610: Lock analysis is not prepared to handle SnatchLock usage
In #5475, we worked around a deadlock by passing a `SnatchGuard` by value to `maintain`. However, this means that `SnatchGuard`s and other kinds of lock guards are not acquired and dropped in a stack-like fashion, which is an assumption made by the locking analysis in `lock::ranked`.

Specifically, `Global::queue_submit` acquires a `SnatchGuard` on the device, and then acquires a read guard on `Device::fence`. However, the `SnatchGuard` is passed by value to `Device::maintain`, which drops it, while the read guard goes out of scope in `queue_submit` after `maintain` has returned.

Thus, the `SnatchGuard` is both acquired first and dropped first, which is a violation of the stacking order, which the lock rank checker assumes.

Kudos to @ErichDonGubler for noting in the review of #5539 that the checker assumes stack-like acquisition and dropping, and suggesting that we actually check that this is the case.

jimblandy:
The lock analyzers in the `wgpu_core::lock` module can be a bit
simpler if they can assume that locks are acquired and released in a
stack-like order: that a guard is only dropped when it is the most
recently acquired lock guard still held. So:

- Change `Device::maintain` to take a `RwLockReadGuard` for the device's hal fence, rather than just a reference to it.

- Adjust the order in which guards are dropped in `Device::maintain` and `Queue::submit`.

- Implement `downgrade` for the `lock` module's `RwLock`s.

    Implement `RwLockWriteGuard::downgrade` for `lock::vanilla` and `lock::ranked`, to downgrade a write lock to a read lock.

- Refactor `lock::ranked` to use `LockStateGuard`.

  Rather than implementing `Drop` for all three lock guard types to restore the lock analysis' per-thread state, let lock guards own values of a new type, `LockStateGuard`, with the appropriate `Drop` implementation. This is cleaner and shorter, and helps us implement `RwLock::downgrade` in a later commit.

Fixes #5610.

### DIFF
--- a/wgpu-core/src/device/queue.rs
+++ b/wgpu-core/src/device/queue.rs
@@ -14,7 +14,7 @@ use crate::{
     hal_label,
     id::{self, DeviceId, QueueId},
     init_tracker::{has_copy_partial_init_tracker_coverage, TextureInitRange},
-    lock::{rank, Mutex},
+    lock::{rank, Mutex, RwLockWriteGuard},
     resource::{
         Buffer, BufferAccessError, BufferMapState, DestroyedBuffer, DestroyedTexture, Resource,
         ResourceInfo, ResourceType, StagingBuffer, Texture, TextureInner,
@@ -1162,8 +1162,8 @@ impl Global {
             let snatch_guard = device.snatchable_lock.read();
 
             // Fence lock must be acquired after the snatch lock everywhere to avoid deadlocks.
-            let mut fence = device.fence.write();
-            let fence = fence.as_mut().unwrap();
+            let mut fence_guard = device.fence.write();
+            let fence = fence_guard.as_mut().unwrap();
             let submit_index = device
                 .active_submission_index
                 .fetch_add(1, Ordering::Relaxed)
@@ -1459,8 +1459,8 @@ impl Global {
                 }
             }
 
-            let mut pending_writes = device.pending_writes.lock();
-            let pending_writes = pending_writes.as_mut().unwrap();
+            let mut pending_writes_guard = device.pending_writes.lock();
+            let pending_writes = pending_writes_guard.as_mut().unwrap();
 
             {
                 used_surface_textures.set_size(hub.textures.read().len());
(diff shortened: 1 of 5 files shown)


jimblandy:
The interesting thing for reviewers to consider here is the shifting of the code that assigns to `pending_writes.temp_resources` from after the `maintain` to before. I believe this is safe because at this point `pending_writes` should be entirely empty, and it's fine if some other thread starts queuing up new stuff.

ErichDonGubler:
~~It looks like this will be in my queue until Monday. Sorry for the wait! 😖~~ EDIT: Plot twist, I have review bandwidth today!

wgpu-core/src/lock/ranked.rs
```diff
@@ -281,6 +281,15 @@ impl<T> RwLock<T> {
     }
 }
 
+impl<'a, T> RwLockWriteGuard<'a, T> {
+    pub fn downgrade(this: Self) -> RwLockReadGuard<'a, T> {
+        RwLockReadGuard {
```
ErichDonGubler, line 286: nitpick(non-blocking): Destructuring might let us catch fields we add later, if ever.

<sub>I'm gonna go write some feedback that maybe actually is important now. 👉🏻👈🏻</sub>

wgpu-core/src/device/queue.rs
```diff
@@ -1162,8 +1162,8 @@ impl Global {
             let snatch_guard = device.snatchable_lock.read();
 
             // Fence lock must be acquired after the snatch lock everywhere to avoid deadlocks.
-            let mut fence = device.fence.write();
-            let fence = fence.as_mut().unwrap();
+            let mut fence_guard = device.fence.write();
+            let fence = fence_guard.as_mut().unwrap();
```
ErichDonGubler, line 1166: thought: It woulda been nice to have mechanical renames in a separate commit, but no point in changing that now that I understand what's going on. 🫡

ErichDonGubler (approved):
shippit

PR #5649, opened 2024-05-02
[wgpu-hal] Upgrade to `glutin` 0.31.3, narrow `libloading` to 0.8

ErichDonGubler:
This eliminates a _ton_ of duplicate dependencies. Woot! Firefox is finally ready for incremental update of `libloading`, so we can make this change without any hassle. 😀

~~Depends on #5504, to have no dependency on `libloading` 0.7.~~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -75,10 +75,10 @@ dependencies = [
  "jni-sys",
  "libc",
  "log",
- "ndk 0.8.0",
+ "ndk",
  "ndk-context",
- "ndk-sys 0.5.0+25.2.9519653",
- "num_enum 0.7.2",
+ "ndk-sys",
+ "num_enum",
  "thiserror",
 ]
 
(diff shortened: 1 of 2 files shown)


cwfitzgerald:
This can be un-drafted

ErichDonGubler:
@cwfitzgerald: It seems we need to pick an optional backend feature for `glutin` now (as indicated by CI)? I don't know immediately what the right answer is, tho. 😖

teoxoy:
They get [automatically enabled](https://github.com/rust-windowing/glutin/blob/052399de0a4a7237b4dc12dd0fcdeb63f1c2af41/glutin/build.rs), the issue is that iOS support [was removed in v0.30](https://github.com/rust-windowing/glutin/blob/052399de0a4a7237b4dc12dd0fcdeb63f1c2af41/CHANGELOG.md?plain=1#L97).

So, I think we should just skip building the example on iOS.

teoxoy:
What about this? Should we remove it or bump it as well?

https://github.com/gfx-rs/wgpu/blob/77a83fb0dd9f2295f25e99a850b9a031738925c3/Cargo.toml#L154

cwfitzgerald (changes requested):
For tracking

MarijnS95:
@ErichDonGubler what's needed to finish this, port the `raw-gles.rs` example because `glutin` completely removed `winit` from its core and is now a proper wrapper around GL context creation (kind of doing the exact same thing as your `wgl.rs`, `egl.rs` etc)?

I'd love to push this through to further reduce dependencies on `winapi` :)

ErichDonGubler:
@MarijnS95: Honestly, this just needs somebody to drive it to completion. I'm happy to cede this PR to you, if you have the bandwidth.

MarijnS95:
@ErichDonGubler bandwidth is tight, but I've written and worked on my fair share of `glutin` codebases over the years so I might be able to pick this up and get it over with in a short amount of time.

Wumpf:
https://github.com/gfx-rs/wgpu/pull/5709 highlights that glutin drags in winit right now, making it a holdout for winit 0.27
(dev dependency only still though)

PR #5651, opened 2024-05-02
Satisfy Rust 1.78's latest easy lints

ErichDonGubler:
**Connections**

- Official Rust blog post: https://blog.rust-lang.org/2024/05/02/Rust-1.78.0.html
- Clippy release notes: https://github.com/rust-lang/rust-clippy/blob/20b085d500dfba5afe0869707bf357af3afe20be/CHANGELOG.md#rust-178

**Description**

This hopefully helps some eventual migration we'll do when we set the MSRV to Rust 1.78 or greater.

**Testing**

CI FTW!

**Checklist**

- [x] Run `cargo fmt`.
- [x] ~~Add change to `CHANGELOG.md`. See simple instructions inside file.~~ I don't think we need this.


cwfitzgerald:
Any reason not to make the repo msrv 1.78?

ErichDonGubler:
@cwfitzgerald: Mostly because [Firefox's current MSRV is 1.76.0](https://searchfox.org/mozilla-central/rev/7ddf1affe2c1fe192fdf88b70a21872af543666f/python/mozboot/mozboot/util.py#14), and I'm not sure that I'd want to go up to bat for bumping the MSRV in `mozilla-central` every time we run into new lints. 😅

Wumpf:
I'm also a bit wary of going with the latest right away on a library like wgpu. There's been bugfixes to a bunch of rust versions and it would be unfortunate to force users onto a new rust version too early

cwfitzgerald:
Context: Strong opinion weekly held, etc

@ErichDonGubler we have separate MSRVs for the repository as a whole and core/hal/types/naga - as long as the changes introduced by clippy don't cause the code to break on the lower msrvs (which we have a CI job to test) we can keep our repo msrv wherever we want.

@Wumpf counterpoint, no one is going to see this version of wgpu until at least 1.80 is out

Would just be nice to enforce these things in CI instead of sporadically

Wumpf:
Good counterpoint!

Since you bring up keeping ci at a different version:
I also would like to get new lints checked locally right away, but it would be super annoying to do a thing that works locally and then get told off by ci because it operates with a different Rust version. Also it means that it can happen that I check out `trunk` and it has warnings on a local machine from the get go.

Wumpf (approved):

cwfitzgerald:
> Since you bring up keeping ci at a different version:
I also would like to get new lints checked locally right away, but it would be super annoying to do a thing that works locally and then get told off by ci because it operates with a different Rust version. Also it means that it can happen that I check out trunk and it has warnings on a local machine from the get go.

I was suggesting bumping the repo msrv, which includes both the `toolchain.toml` and CI versions so local always == CI.

ErichDonGubler:
@cwfitzgerald: My only hesitation is when Clippy eagerly recommends something that only a newer version of Rust can do. I can't remember the concrete example ATM, but I remember this happening sometime in the last year or so.

ErichDonGubler:
Just waiting on @gfx-rs/naga here. Maybe @teoxoy could rubber-stamp? 😀

teoxoy (approved):
Nice!

PR #5653, opened 2024-05-03
[core] Don't check for uniquely referenced resources at submission.

### LINKED ISSUE #5647: [core] Deadlock between `Device::temp_suspected` and `CommandBuffer::data`
Two threads can deadlock as follows:

- The first thread calls `Global::command_encoder_drop`, which locks `CommandBuffer::data` while calling `Device::untrack`, which then locks `Device::temp_suspected`.
- The second thread calls `Queue::submit`, which locks `Device::temp_suspected` and then `CommandBuffer::data`.

The lock around `temp_suspected` should never need to be held for a long time. This is just a simple allocation reuse.

jimblandy:
Remove unreachable code from `Global::queue_submit` that checks whether the resources used by the command buffer have a reference count of one, and adds them to `Device::temp_suspected` if so.

When `queue_submit` is called, all the `Arc`s processed by this code have a reference count of at least three, even when the user has dropped the resource:

- `Device::trackers` holds strong references to all the device's resources.

- `CommandBufferMutable::trackers` holds strong references to all resources used by the command buffer.

- The `used_resources` methods of the various members of `CommandBufferMutable::trackers` all return iterators of owned `Arc`s.

Fortunately, since the `Global::device_drop_foo` methods all add the `foo` being dropped to `Device::suspected_resources`, and `LifetimeTracker::triage_suspected` does an adequate job of accounting for the uninteresting `Arc`s and leaves the resources there until they're actually dead, things do get cleaned up without the checks in `Global::queue_submit`.

Fixes #5647.

### DIFF
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -127,9 +127,6 @@ pub struct Device<A: HalApi> {
     pub(crate) tracker_indices: TrackerIndexAllocators,
     // Life tracker should be locked right after the device and before anything else.
     life_tracker: Mutex<LifetimeTracker<A>>,
-    /// Temporary storage for resource management functions. Cleared at the end
-    /// of every call (unless an error occurs).
-    pub(crate) temp_suspected: Mutex<Option<ResourceMaps<A>>>,
     /// Pool of bind group layouts, allowing deduplication.
     pub(crate) bgl_pool: ResourcePool<bgl::EntryMap, BindGroupLayout<A>>,
     pub(crate) alignments: hal::Alignments,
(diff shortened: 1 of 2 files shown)


jimblandy:
I removed another needless reference to `temp_suspected`.

wgpu-core/src/device/resource.rs
```diff
@@ -142,6 +139,10 @@ pub struct Device<A: HalApi> {
     #[cfg(feature = "trace")]
     pub(crate) trace: Mutex<Option<trace::Trace>>,
     pub(crate) usage_scopes: UsageScopePool<A>,
+
+    /// Temporary storage, cleared at the start of every call,
+    /// retained only to save allocations.
+    temp_suspected: Mutex<Option<ResourceMaps<A>>>,
```
ErichDonGubler, line 145: question(blocking): Should we actually even keep this field? The only remaining references to it that I see are setting it to `ResourceMaps::new()`, or clearing it. This makes me question the doc. commentary, too, because we're obviously not re-using anything.

ErichDonGubler (approved):
LGTM and approving, just one question in a convo I'd like handled (on your honor) before you merge.

praise: `Global::queue_submit` makes _much more sense to me now; it's a lot easier to understand _just_ updating the most recent submission index on everything.

wgpu-core/src/device/resource.rs
jimblandy: Well, that last reference in `Device::untrack` is still meaningful. The code is broken, but the remainder of #5594 that still applies after this PR would make it effective.

Whether it's worthwhile to save that allocation, I have no idea. We have no numbers!

PR #5654, opened 2024-05-03
Workaround shader compiler bugs with degenerate switch statements

### LINKED ISSUE #4485: [hlsl-out] "continue cannot be used in a switch" within loop
WGSL:

```wgsl
@compute
@workgroup_size(1)
fn main() {
    loop {
        switch (0) {
            default: {
                continue;
            }
        }
    }
}
```

HLSL:

```hlsl
[numthreads(1, 1, 1)]
void main()
{
    while(true) {
        switch(0) {
            default: {
                continue;
            }
        }
    }
    return;
}
```

The generated HLSL is rejected by FXC with the following error:

```
error X3708: continue cannot be used in a switch
```

I guess the `continue` would need to be compiled down to e.g. a flag that conditionally executes the remainder of the loop.

### LINKED ISSUE #4514: Don't switch on a single constant in output consumed by FXC
`glslc` emits code like `OpSwitch 0` for control flow reasons, for example when one branch ends with a `OpKill`: 20220604_02h00m28s_grim

this produces the rather dodgy glsl code:

```glsl
switch(int(0u)) {
        default:
            ...
            if (((_e46.w * _e49.w) < 0.5)) {
                discard;
            } else {
                break;
            }
            break;
    }
}
```

On certain GPUs I've had this not render the fragment at all (need to double check).

Ideally we should check if there's only one default case and just not produce the switch in that case.

Imberflur:
**Connections**
Fixes https://github.com/gfx-rs/wgpu/issues/4514
Fixes https://github.com/gfx-rs/wgpu/issues/4485

**Description**

As discussed in https://github.com/gfx-rs/wgpu/issues/4485, FXC has a bug when handling switches with all cases leading to one body (it deletes the switch). It was also mentioned that some glsl consumers might not handle this scenario well either (but I don't know which ones that might be).

So we lower these to `do {} while(false);` loops in hlsl-out and glsl-out.

While FXC already can't handle continue statements in switches (see #4485), I ended up including logic to properly forward continue statements to the outer loop here, since there would otherwise be regressions in glsl-out and maybe hlsl-out with DXC.

An alternative solution could be to switch on a known constant and have a dummy case with an empty body that is never hit. This is a bit less cautious in terms of avoiding potential compiler bugs since we are still switching on a constant. But the implementation would be simpler and the only issues I have observed so far with FXC are when all cases lead to a single body.

**Testing**

Some of these scenarios are already covered by existing naga snapshot tests, I extended `naga/tests/in/control-flow.wgsl` with a few more cases.

I also included some regression tests for the linked issues that will execute the shaders and assert on their output. Additionally, a new mechanism is introduced into the test framework for forcing the usage of FXC when on the Dx12 backend.

I ran `cargo xtask test` on a windows machine and a separate linux machine.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/naga/src/back/continue_forward.rs
+++ b/naga/src/back/continue_forward.rs
@@ -1,0 +1,311 @@
+//! Workarounds for platform bugs and limitations in switches and loops.
+//!
+//! In these docs, we use CamelCase links for Naga IR concepts, and ordinary
+//! `code` formatting for HLSL or GLSL concepts.
+//!
+//! ## Avoiding `continue` within `switch`
+//!
+//! As described in <https://github.com/gfx-rs/wgpu/issues/4485>, the FXC HLSL
+//! compiler doesn't allow `continue` statements within `switch` statements, but
+//! Naga IR does. We work around this by introducing synthetic boolean local
+//! variables and branches.
+//!
+//! Specifically:
+//!
+//! - We generate code for [`Continue`] statements within [`SwitchCase`]s that
+//!   sets an introduced `bool` local to `true` and does a `break`, jumping to
+//!   immediately after the generated `switch`.
+//!
+//! - When generating code for a [`Switch`] statement, we conservatively assume
+//!   it might contain such a [`Continue`] statement, so:
+//!
+//!   - If it's the outermost such [`Switch`] within a [`Loop`], we declare the
+//!     `bool` local ahead of the switch, initialized to `false`. Immediately
+//!     after the `switch`, we check the local and do a `continue` if it's set.
+//!
+//!   - If the [`Switch`] is nested within other [`Switch`]es, then after the
+//!     generated `switch`, we check the local (which we know was declared
+//!     before the surrounding `switch`) and do a `break` if it's set.
+//!
+//!   - As an optimization, we only generate the check of the local if a
+//!     [`Continue`] statement is encountered within the [`Switch`]. This may
+//!     help drivers more easily identify that the `bool` is unused.
+//!
+//! So while we "weaken" the [`Continue`] statement by rendering it as a `break`
+//! statement, we also place checks immediately at the locations to which those
+//! `break` statements will jump, until we can be sure we've reached the
+//! intended target of the original [`Continue`].
+//!
+//! In the case of nested [`Loop`] and [`Switch`] statements, there may be
+//! multiple introduced `bool` locals in scope, but there's no problem knowing
+//! which one to operate on. At any point, there is at most one [`Loop`]
+//! statement that could be targeted by a [`Continue`] statement, so the correct
+//! `bool` local to set and test is always the one introduced for the innermost
+//! enclosing [`Loop`]'s outermost [`Switch`].
+//!
+//! # Avoiding single body `switch` statements
+//!
+//! As described in <https://github.com/gfx-rs/wgpu/issues/4514>, some language
+//! front ends miscompile `switch` statements where all cases branch to the same
+//! body. Our HLSL and GLSL backends render [`Switch`] statements with a single
+//! [`SwitchCase`] as `do {} while(false);` loops.
+//!
+//! However, this rewriting introduces a new loop that could "capture"
+//! `continue` statements in its body. To avoid doing so, we apply the
+//! [`Continue`]-to-`break` transformation described above.
+//!
+//! [`Continue`]: crate::Statement::Continue
+//! [`Loop`]: crate::Statement::Loop
+//! [`Switch`]: crate::Statement::Switch
+//! [`SwitchCase`]: crate::SwitchCase
+
+use crate::proc::Namer;
+use std::rc::Rc;
+
+/// A summary of the code surrounding a statement.
+enum Nesting {
+    /// Currently nested in at least one [`Loop`] statement.
+    ///
+    /// [`Continue`] should apply to the innermost loop.
+    ///
+    /// When this entry is on the top of the stack:
+    ///
+    /// * When entering an inner [`Loop`] statement, push a [`Loop`][nl] state
+    ///   onto the stack.
+    ///
+    /// * When entering a nested [`Switch`] statement, push a [`Switch`][ns]
+    ///   state onto the stack with a new variable name. Before the generated
+    ///   `switch`, introduce a `bool` local with that name, initialized to
+    ///   `false`.
+    ///
+    /// When exiting the [`Loop`] for which this entry was pushed, pop it from
+    /// the stack.
+    ///
+    /// [`Continue`]: crate::Statement::Continue
+    /// [`Loop`]: crate::Statement::Loop
+    /// [`Switch`]: crate::Statement::Switch
+    /// [ns]: Nesting::Switch
+    /// [nl]: Nesting::Loop
+    Loop,
+
+    /// Currently nested in at least one [`Switch`] that may need to forward
+    /// [`Continue`]s.
+    ///
+    /// This includes [`Switch`]es rendered as `do {} while(false)` loops, but
+    /// doesn't need to include regular [`Switch`]es in backends that can
+    /// support `continue` within switches.
+    ///
+    /// [`Continue`] should be forwarded to the innermost surrounding [`Loop`].
+    ///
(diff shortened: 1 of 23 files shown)


Imberflur:
```
stderr:
error: line 127: Case construct that targets '74[%74]' has invalid branch to block '55[%55]' (not another case construct, corresponding merge, outer loop merge or outer loop continue)
  %74 = OpLabel
```

I guess this is one of the nested test cases I added.

Imberflur:
> ```
> stderr:
> error: line 127: Case construct that targets '74[%74]' has invalid branch to block '55[%55]' (not another case construct, corresponding merge, outer loop merge or outer loop continue)
>   %74 = OpLabel
> ```
> 
> I guess this is one of the nested test cases I added.

Found minimal case https://github.com/gfx-rs/wgpu/issues/5658 and was able to avoid the issue here without affecting the aspects I was trying to test.

naga/src/back/glsl/mod.rs
```diff
@@ -2077,42 +2083,110 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                // Some glsl consumers may not handle switches with a single body correctly.
+                // So write this as a `do {} while(false);` loop instead.
+                // This is the same fix as for FXC in hlsl, although here it
+                // wasn't specifically tested against any problematic consumers.
+                // See https://github.com/gfx-rs/wgpu/issues/4514
+                let one_body = cases
+                    .iter()
+                    .rev()
+                    .skip(1)
+                    .all(|case| case.fall_through && case.body.is_empty());
```
Imberflur: One issue I just realized is what if none of these switch cases would have matched, I need to check if that's possible.

Imberflur: Nvm looks like naga's switch statement requires a default case!

Imberflur: Maybe I should note that here?

tests/src/init.rs
```diff
@@ -27,7 +27,12 @@ pub fn initialize_instance() -> Instance {
     } else {
         Backends::all()
     };
-    let dx12_shader_compiler = wgpu::util::dx12_shader_compiler_from_env().unwrap_or_default();
+    // TODO: should env override force_fxc?
```
cwfitzgerald: No - tests need to be able to force demote it.

cwfitzgerald (approved):
Wgpu side looks amazing, minus a nit

naga/src/back/glsl/mod.rs
Imberflur: I added a note

tests/src/init.rs
Imberflur: Thanks! Removed TODO comment and added a small note about this.

Imberflur:
Let me know if/when conflicts need to be fixed.

naga/src/back/continue_forward.rs
```diff
@@ -0,0 +1,159 @@
...
+//! need to forward continue statements to the outer loop.
+//!
+//! See <https://github.com/gfx-rs/wgpu/issues/4514>
+
+enum Nesting {
+    /// Currently nested in at least one loop.
+    ///
+    /// `continue` should apply to the current loop.
+    ///
+    /// * When entering a nested switch, add a `Forward` state to the stack.
```
jimblandy: What is `Forward`? Docs are out of date?

jimblandy:
@Imberflur Started reviewing this. My impression so far is that this is excellent work. I especially appreciate the execution tests and the docs.

jimblandy:
@Imberflur Could you check out the docs I just pushed in ~~4c10a41~~ [b185798](https://github.com/gfx-rs/wgpu/pull/5654/commits/b1857983649f9e219b7e09f91f98f79f5e9af99f)?

jimblandy:
> Let me know if/when conflicts need to be fixed.

It seems like this needs a rebase.

edit: I did the rebase.

naga/src/back/continue_forward.rs
Imberflur: Yeah, looks out of date. This should be `Nesting::Switch` (or just `Switch`) (confirmed by looking at the behavior in `ContinueCtx::enter_switch`).

naga/src/back/continue_forward.rs
```diff
@@ -0,0 +1,204 @@
...
+//!
+//! - When a [`Switch`] statement contains such a [`Continue`] statement:
+//!
+//!   - If it's the outermost such [`Switch`] within the targeted [`Loop`], we
+//!     declare the `bool` local, initialized to `false`, and immediately after
+//!     the `switch`, we check the local and do a `continue` if it's set.
+//!
+//!   - If the [`Switch`] is nested within other [`Switch`]es, then after the
+//!     generated `switch`, we check the local (which we know was declared
+//!     before the surrounding `switch`) and do a `break` if it's set.
```
Imberflur: As a nit, I would say that this line might be clarified:

> When a [`Switch`] statement contains such a [`Continue`] statement:

To me, this could imply that the following points are only applied for switches containing continue statements, when it is really applied for all switches (that are inside a loop and can't use `continue` directly) since we don't do anything to detect whether a switch statement contains `continue`s.

But FWIW I think the false implication is fairly weak and not significant to understanding the mechanisms here and I'm happy with the current text.

Imberflur (commented):
The new docs look great! Thanks for adding them.

I included one note for consideration.

P.S. The "forwarding" terminology is somewhat ad hoc, so feel free to suggest something else (like "deferring") if you have opinions there.

Imberflur:
I'm somewhat curious about what conflicts came up in the rebase (in the interest of double checking them). But I can always re-run the rebase locally to find them.

jimblandy:
> I'm somewhat curious about what conflicts came up in the rebase (in the interest of double checking them). But I can always re-run the rebase locally to find them.

Only the absence of a `cache` field in the pipeline descriptors.

naga/src/back/continue_forward.rs
jimblandy: It's a fair point - it might make people expect we do such a scan somewhere. I made the docs describe what we really do.

jimblandy (changes requested):
This looks great, and I think it's correct, but I'd like to see it simplified in two ways:

- Don't bother counting nesting depth; just always push/pop new entries on the stack. It took me a bit to realize that the depth counters were just an optimization, not something that actually affected the code generated.

- Eliminate `ContinueCtx::next_id`, let `enter_switch` take a `&mut Namer` argument, and just use the namer to generate unique names directly. It took me a while to be sure that the names you were generating couldn't conflict; if the code just uses the namer, then that question won't even arise.

  This means that `Nesting::Switch` will need to actually own the `String`. I think it will need to be an `Rc<String>`, both so that nested `Switch` variants can share it, and also so that `exit_switch` can return it regardless of whether it's been popped. `ExitControl` flow will need to hold a clone of the `String`.

jimblandy:
I played with having the HLSL and GLSL writers just save the current nesting state in a local variable, and restore it on exit, so you could replace `ContinueCtx::stack` with a single `Nesting` value. But the API doesn't end up as nice as the one you've got, so I thought it would end up being more distracting in the end.

jimblandy:
@Imberflur The only other suggestion I might have - and maybe this is just too hairy to be worth it - is that we might put a flag in `Nesting::Switch` to say whether it actually encountered any `Continue` statements that needed to be forwarded. If not, then there's no need to generate the check after the `switch`. Omitting those checks might make it easier for dumb drivers to see that the bool is truly unused.

Imberflur:
>  just use the namer to generate unique names directly

Oh that sounds great! I didn't realize there was a mechanism for this.

Imberflur:
@jimblandy  I implemented suggestions:

* Removed nesting depth counter.
* Used `Namer` and `Rc<String>` for variable names.
* Omitted check generation if no continue statements are found.
* Rebased

jimblandy (approved):
Okay, I fiddled with the docs some more, and cleaned up a bit of the control flow, and I think this ready to land.

Imberflur:
Thanks!

PR #5657, opened 2024-05-03
Seperate GLES and GL version selection

valaphee:
**Connections**

**Description**
Was wondering why compatibility profile was selected, and its because it falsely used the GLES version for Desktop GL.

Now it only applies to GLES, and for Desktop the minimum required version is 3.3

**Testing**

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-hal/src/gles/egl.rs
+++ b/wgpu-hal/src/gles/egl.rs
@@ -550,21 +550,25 @@ impl Inner {
         let supports_khr_context = display_extensions.contains("EGL_KHR_create_context");
 
         //TODO: make it so `Device` == EGL Context
-        let mut context_attributes = vec![
-            khronos_egl::CONTEXT_MAJOR_VERSION,
-            3, // Request GLES 3.0 or higher
-        ];
-
-        if force_gles_minor_version != wgt::Gles3MinorVersion::Automatic {
+        let mut context_attributes = vec![];
+        if !supports_opengl {
+            context_attributes.push(khronos_egl::CONTEXT_MAJOR_VERSION);
+            context_attributes.push(3); // Request GLES 3.0 or higher
+            if force_gles_minor_version != wgt::Gles3MinorVersion::Automatic {
+                context_attributes.push(khronos_egl::CONTEXT_MINOR_VERSION);
+                context_attributes.push(match force_gles_minor_version {
+                    wgt::Gles3MinorVersion::Version0 => 0,
+                    wgt::Gles3MinorVersion::Version1 => 1,
+                    wgt::Gles3MinorVersion::Version2 => 2,
+                    _ => unreachable!(),
+                });
+            }
+        } else {
+            context_attributes.push(khronos_egl::CONTEXT_MAJOR_VERSION);
+            context_attributes.push(3);
             context_attributes.push(khronos_egl::CONTEXT_MINOR_VERSION);
-            context_attributes.push(match force_gles_minor_version {
-                wgt::Gles3MinorVersion::Version0 => 0,
-                wgt::Gles3MinorVersion::Version1 => 1,
-                wgt::Gles3MinorVersion::Version2 => 2,
-                _ => unreachable!(),
-            });
+            context_attributes.push(3);
         }
-
         if flags.contains(wgt::InstanceFlags::DEBUG) {
             if version >= (1, 5) {
                 log::debug!("\tEGL context: +debug");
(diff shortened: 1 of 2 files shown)


wgpu-hal/src/gles/egl.rs
```diff
@@ -550,21 +550,25 @@ impl Inner {
...
 
         //TODO: make it so `Device` == EGL Context
-        let mut context_attributes = vec![
-            khronos_egl::CONTEXT_MAJOR_VERSION,
-            3, // Request GLES 3.0 or higher
-        ];
-
-        if force_gles_minor_version != wgt::Gles3MinorVersion::Automatic {
+        let mut context_attributes = vec![];
+        if !supports_opengl {
```
Wumpf: please flip this to `supports_opengl`, reads a lot nicer

wgpu-hal/src/gles/egl.rs
```diff
@@ -550,21 +550,25 @@ impl Inner {
         let supports_khr_context = display_extensions.contains("EGL_KHR_create_context");
 
         //TODO: make it so `Device` == EGL Context
```
Wumpf: I think that todo comment is outdated. At least I have no idea what it wants

wgpu-hal/src/gles/egl.rs
```diff
@@ -550,21 +550,25 @@ impl Inner {
...
+        if !supports_opengl {
+            context_attributes.push(khronos_egl::CONTEXT_MAJOR_VERSION);
+            context_attributes.push(3); // Request GLES 3.0 or higher
+            if force_gles_minor_version != wgt::Gles3MinorVersion::Automatic {
+                context_attributes.push(khronos_egl::CONTEXT_MINOR_VERSION);
+                context_attributes.push(match force_gles_minor_version {
+                    wgt::Gles3MinorVersion::Version0 => 0,
+                    wgt::Gles3MinorVersion::Version1 => 1,
+                    wgt::Gles3MinorVersion::Version2 => 2,
+                    _ => unreachable!(),
```
Wumpf: let's make this an exhaustive pattern so it's easy to see _what_ is unreachable (I really hope the only thing that is in here is `Automatic` 😄 )
(I know this was like that before, but you touched it ;-))

wgpu-hal/src/gles/egl.rs
```diff
@@ -550,21 +550,25 @@ impl Inner {
...
+            context_attributes.push(khronos_egl::CONTEXT_MAJOR_VERSION);
+            context_attributes.push(3);
             context_attributes.push(khronos_egl::CONTEXT_MINOR_VERSION);
-            context_attributes.push(match force_gles_minor_version {
-                wgt::Gles3MinorVersion::Version0 => 0,
-                wgt::Gles3MinorVersion::Version1 => 1,
-                wgt::Gles3MinorVersion::Version2 => 2,
-                _ => unreachable!(),
-            });
+            context_attributes.push(3);
```
Wumpf: I think we should clarify on the user facing `InstanceDescriptor::gles_minor_version` that it is ignored when desktop OpenGL is chosen.
I'm also still pondering if this should file a warning if we're in OpenGL mode and and this is not set to automatic 🤔, might be annoying though.

Wumpf (changes requested):
makes sense!

I'm a bit rusty with these things, what is the effect of selecting compatibility mode when we don't want it in practical terms?

wgpu-hal/src/gles/egl.rs
valaphee: What would be the intention when forcing the minor version?

Theoretically we could also use this to force an OpenGL ES backend, but this behavior would differ from what WGL is doing (as it never uses OpenGL ES anyway)

valaphee:
> makes sense!
> 
> I'm a bit rusty with these things, what is the effect of selecting compatibility mode when we don't want it in practical terms?

Compatibility Mode just exposes some deprecated functionalities/extensions of OpenGL like fixed-function pipelines, etc. which we don't shouldn't use anyway.

https://docs.google.com/spreadsheets/d/1hwSMn4GCpyUk_xDDBN3en8wswemq9kkX7FtHDlcAX-8/edit#gid=0
Here is an overview

Wumpf:
thank for digging this up. Now it also makes sense to me why you opted out of adding to the changelog; it should't have any effect on users of wgpu but may cause implementors to do stuff that shouldn't be supported :)

wgpu-hal/src/gles/egl.rs
Wumpf: I don't know for sure but the docs imply to me that this was used as a workaround in Angle at some point or still is 🤔  https://github.com/gfx-rs/wgpu/blob/trunk/wgpu-types/src/lib.rs#L7071
Forcing a GLES backend might be useful in a similar vain, but if we want that we should take this separately :)

wgpu-hal/src/gles/egl.rs
valaphee: It's there since the beginning, no idea either especially in the context of `context_attributes`

Wumpf (approved):
thank you for fixing all those nits!
🚢

PR #5671, opened 2024-05-05
Fix QuerySet ownership of ComputePass

Wumpf:
**Connections**
* Follow-up & based on #5620
* Puzzle piece of #1453

* #5409
* #5432
* #5570
* #5575
* #5620
* ➡️ #5671

_last piece of solving lifetime constraints for `ComputePass`_

**Description**
Technically this PR is a bugfix: Previously, when destroying a `QuerySet` that is used in any way (either `write_timestamp`, `ComputePassTimestampWrites` or `begin_pipeline_statistics_query`) before executing the pass, we would crash.
This solves this as in the PRs before by taking ownership of the query the moment it is provided.

**Testing**
Adds two new tests that test for gracefully handling destroyed query set before the compute pass is executed.
Both tests failed before the fix.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,7 @@ fn independent_cpass<'enc>(encoder: &'enc mut wgpu::CommandEncoder) -> wgpu::Com
 This is very useful for library authors, but opens up an easy way for incorrect use, so use with care.
 `forget_lifetime` is zero overhead and has no side effects on pass recording.
 
-By @wumpf in [#5569](https://github.com/gfx-rs/wgpu/pull/5569), [#5575](https://github.com/gfx-rs/wgpu/pull/5575), [#5620](https://github.com/gfx-rs/wgpu/pull/5620), [#5768](https://github.com/gfx-rs/wgpu/pull/5768) (together with @kpreid).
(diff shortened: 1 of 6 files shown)


wgpu-core/src/command/query.rs
```diff
@@ -214,133 +213,110 @@ impl<A: HalApi> QuerySet<A> {
...
+        reset_state,
+    )?;
+
+    if let Some((_old, old_idx)) = active_query.take() {
+        return Err(QueryUseError::AlreadyStarted {
+            active_query_index: old_idx,
+            new_query_index: query_index,
+        });
+    }
+    let query_set = &active_query.insert((query_set, query_index)).0;
```
ErichDonGubler: nitpick: It's easier to follow this code when you mirror the tuple structure b/w LHS and RHS:

```suggestion
    let (query_set, _) = &active_query.insert((query_set, query_index));
```

Pushed a `fixup!` for this and the case in `validate_and_begin_occlusion_query` with dfcf44620.

CHANGELOG.md
ErichDonGubler, line 1: nitpick(non-blocking): I actually think this is a case where squashing the changes together would be appropriate, since the tests don't pass without the fix, and the `CHANGELOG` entry can't stand alone without that commit either.

ErichDonGubler (approved):
praise: This was easier to audit than I thought it would be, based on the LoC. 😀 Whew! Sorry for the delay in getting to this.

wgpu-core/src/command/query.rs
Wumpf: good style advice, should do that more often like so

PR #5679, opened 2024-05-08
meta: add `GOVERNANCE.md`

ErichDonGubler:


ErichDonGubler:
Currently in draft to reflect that I would like feedback from decision-making groups identified in this current revision. To wit: @gfx-rs/wgpu, @gfx-rs/naga

I've given my best first approximation of how governance currently works in the WGPU project, but I expect that there's quite a bit I'm missing, or that's not quite correct.

GOVERNANCE.md
```diff
@@ -0,0 +1,44 @@
...
+1. …serves as a platform of experimentation for:
+    1. …WebGPU standards development.
+    1. …native application authors.
+
+## Decision-making
+
+The WGPU community's decision-making is influenced by the following
+groups:
+
+* Connor Fitzgerald (@cwfitzgerald)
```
cwfitzgerald: Let's call this community leadership and have me, wumpf, groves

GOVERNANCE.md
```diff
@@ -0,0 +1,44 @@
...
+* Firefox's WebGPU team
+* Other users that ship applications based on WGPU
```
cwfitzgerald, line 41: We should mention deno and crowlkats

GOVERNANCE.md
```diff
@@ -0,0 +1,44 @@
+NOTE: The following is a snapshot of a perpetually changing state of affairs in
+the WGPU community. It is not a binding contract between users and
+decision-makers of the WGPU project.
```
cwfitzgerald: This should be at the end - this is legaleze no user cares about.

GOVERNANCE.md
ErichDonGubler: I was careful to avoid mentioning groups smaller than Firefox (because I don't want details that are prone to changing in a document that seems frankly unlikely to get updated often), but if you'd like them explicitly mentioned, then SGTM!

GOVERNANCE.md
ErichDonGubler: I think that having it at the front is the only way it'll ever be seen, but 🤷🏻‍♂️ I don't have particularly strong feelings about the position.

cwfitzgerald: This strikes me as the kind of thing that exists so we can point to when someone tries to be an ass and use this document against us.

GOVERNANCE.md
cwfitzgerald: We have an explicitly special relationship with deno (we host deno_webgpu!) which has existed for a few years, I don't expect this to really change. We could also put him under the community banner, as I think of him as part of community interest.

GOVERNANCE.md
ErichDonGubler: I think your analysis is on point. 🙂

GOVERNANCE.md
```diff
@@ -0,0 +1,44 @@
...
+
+---
+
+The **WGPU project** is an ecosystem that _enables application authors to write
+portable and performant graphics programs_. It was originally conceived to
+provide an implementation of WebGPU for Firefox as the standard evolved and
+settled into something that could be shipped on all web browsers. WGPU has also
+enjoyed much contribution and use from other projects that requires graphics
```
Wumpf:
```suggestion
enjoyed much contribution and use from other projects that require graphics
```

GOVERNANCE.md
```diff
@@ -0,0 +1,44 @@
...
+The WGPU community seeks to realize the following directives through the
+project: it…
+
+1. …provides an implementation of the WebGPU API that is…
+    1. …correct and fully conformant.
+    1. …portable across all major platforms.
+    1. …performant enough to enable demanding applications.
+1. …serves as a platform of experimentation for:
+    1. …WebGPU standards development.
+    1. …native application authors.
```
Wumpf: Very happy with the first point, but I feel a few important aspects of wgpu are not quite captured yet here:
* seamless use as a graphics backend in applications targeting **both native & web**
* provide a WebGPU _style_ graphics API abstraction to native _Rust_ applications 

currently 'native' is only under the "platform of experimentation" which sounds a bit like anything that isn't part of a WebGPU implementation isn't fully supported.
I'm still pondering a bit how much our many many native extensions are directly part of the mission: they're clearly in part (!) experimentation but realistically this is far outside of what can be expected to be part of WebGPU any time soon. But these are still important pieces in providing a flexible WebGPU-like api to native Rust users, so I believe this deserves a place here.
My understanding of wgpu's mission is first WebGPU impl (as already well described above), then native graphics abstraction library with seamless web support, then platform for experimentation as described. All of these are obviously interconnected (well otherwise we wouldn't put it into a single family of libraries), but they do occasionally pull into different directions and it's worth pointing them and the implicit hierarchy out

GOVERNANCE.md
```diff
@@ -0,0 +1,44 @@
...
+* Connor Fitzgerald (@cwfitzgerald)
+* Firefox's WebGPU team
+* Other users that ship applications based on WGPU
+
+It is no coincidence that these groups correspond to the historically most
+active and consistent contributors. In general, WGPU's community structure is
+meritocratic: social influence is granted proportionate to groups' contribution
+to and stake in WGPU's mission.
+
+These decision-making groups meet together regulrly to discuss issues of
```
nical: typo: regulrly -> regularly

GOVERNANCE.md
ErichDonGubler: Implemented with fe1d09d0a.

GOVERNANCE.md
ErichDonGubler: Noted Deno and @crowlKats with a62303575.

I've added Firefox contributor names with a62303575 (minus @bradwerth, unless he also wants to be on there), though I think maybe that's not a good idea for us, since membership and contribution can fluctuate over time? CC @jimblandy, @teoxoy.

GOVERNANCE.md
ErichDonGubler: Moved to the bottom with 465c75ddd.

GOVERNANCE.md
ErichDonGubler: D'oh, implemented with 465c75ddd.

GOVERNANCE.md
ErichDonGubler: D'oh, fixed with b2e21e072.

GOVERNANCE.md
ErichDonGubler: Resolving for now, but feel free to un-resolve if somebody has an objection to the current contents.

GOVERNANCE.md
ErichDonGubler: I've attempted a revision of this text along the lines of what it seems you want at cd8cf89d5, but I'm not confident I've captured things to your satisfaction. Whaddaya think?

GOVERNANCE.md
```diff
@@ -0,0 +1,54 @@
...
+
+1. …provides libraries for the WebGPU API that…
+    1. …is correct and fully conformant.
+    1. …is portable across all major platforms, that is, …
+        1. …`wgpu-core` enables JavaScript platforms to implement their own
+           proper WebGPU API.
+        1. …`wgpu` provides a WebGPU-style API library for native applications,
+           which allows shipping to all major platforms, including WebGPU's
+           JavaScript API.
+    1. …is performant enough to enable demanding applications.
```
cwfitzgerald: `/s/...is/...are` subject-verb agreement

GOVERNANCE.md
```diff
@@ -0,0 +1,54 @@
...
+## Decision-making
+
+The WGPU community's decision-making is influenced by the following
+groups:
+
+* Community leadership:
+    * Connor Fitzgerald (@cwfitzgerald)
+    * Joshua Groves (@groves)
+    * Andreas Reich (@wumpf)
+* Firefox's WebGPU team (@jimblandy, @teoxoy, @ErichDonGubler, and others)
```
cwfitzgerald: I will take mild offense on nical's behalf, he definitely belongs on this list :)

GOVERNANCE.md
```diff
@@ -0,0 +1,54 @@
+The **WGPU project** is an ecosystem that _enables application authors to write
```
kpreid: The word “ecosystem” seems wrong to me. In particular, to me it implies _at least_ a group of _separately administrated_ interrelated projects, whereas this document refers to a _uniformly_ administrated set.

How about something narrower and plainer like “The WGPU project is a collection of open-source libraries that…”

GOVERNANCE.md
```diff
@@ -0,0 +1,54 @@
...
+active and consistent contributors. In general, WGPU's community structure is
+meritocratic: social influence is granted proportionate to groups' contribution
+to and stake in WGPU's mission.
+
+These decision-making groups meet together regularly to discuss issues of
+importance to the community, with a focus on WGPU's [mission](#Mission).
+
+---
+
+NOTE: The following is a snapshot of a perpetually changing state of affairs in
```
kpreid:
```suggestion
NOTE: The above is a snapshot of a perpetually changing state of affairs in
```

GOVERNANCE.md
Wumpf: nice, has all the points baked in, I like it. 🚢

Wumpf (approved):
there's still good feedback coming in that needs to be incorporated as of writing, but on my side I'm quite happy with this otherwise. Thanks again!

GOVERNANCE.md
cwfitzgerald:
```suggestion
    1. …are correct and fully conformant.
    1. …are portable across all major platforms, that is, …
        1. …`wgpu-core` enables JavaScript platforms to implement their own
           proper WebGPU API.
        1. …`wgpu` provides a WebGPU-style API library for native applications,
           which allows shipping to all major platforms, including WebGPU's
           JavaScript API.
    1. …are performant enough to enable demanding applications.
```

In useful suggestion form, as opposed to non-useful unix-nerd form.

GOVERNANCE.md
ErichDonGubler: Derp, I had actually thought to make this edit, but dropped it mentally, for some reason. Thanks!

GOVERNANCE.md
ErichDonGubler: ofrick urite sorry @nical 😭

This is why I didn't wanna name people, yo.

Will fix soon.

ErichDonGubler: Fixed with 598bc0ce9.

GOVERNANCE.md
ErichDonGubler: 🤦🏻‍♂️Fixed with 76baf1055. That's what I get for picking a form that's requires fuller re-reading to catch, _and_ making two separate edits without double-checking the entire chain of grammar.

GOVERNANCE.md
ErichDonGubler: Fixed with d6a8c81c8.

GOVERNANCE.md
ErichDonGubler: Hmm. It took me some thinking about this deeply, but I think you're right. That's an interesting distinction I hadn't considered before.

Is WGPU not an ecosystem, though? There are several third-party crates that integrate directly with `wgpu` and `naga`. In that case, maybe rephrasing to "…_creates_ an ecosystem that…" (emphasis on edit, not to be propagated in final changes).

cwfitzgerald: Yeah, i would say it _has_ an ecosystem, but is not one itself

kpreid: > Is WGPU not an ecosystem, though?

Well, I would understand “the WGPU ecosystem” as the WGPU project plus those third-party crates. But “third-party” here means, at a minimum, “not governed by the WGPU project organization” — and to me, “ecosystem” further implies that those parties created the crates _by their own choices, to serve their own goals_ and as a consequence made WGPU more useful for everyone.

This document describes governance for the WGPU project, not things that are not the WGPU project. Governing the “ecosystem” by these principles and leadership is a logical contradiction, because if you could, it would cease to be an “ecosystem” and become part of the project.

ErichDonGubler: Changed as suggested with 7a73c1622.

ErichDonGubler:
I think the only feedback I'd like to get before merging is from individuals mentioned as influential members of the community in this new document. IOW, **I'd like to see if we can get a nice green checkmark from**:

* @cwfitzgerald
* @grovesNL
* @jimblandy
* @nical
* @teoxoy
* @crowlKats

I intend to merge this by end-of-week, or when the above individuals approve (whichever is sooner). I don't see a good reason to block for longer than that.

teoxoy (approved):

GOVERNANCE.md
```diff
@@ -0,0 +1,55 @@
+The **WGPU project** is a set of open-source libraries that _enables application
+authors to write portable and performant graphics programs_. It was originally
+conceived to provide an implementation of WebGPU for Firefox as the standard
+evolved and settled into something that could be shipped on all web browsers.
```
cwfitzgerald:
```suggestion
authors to write portable and performant graphics programs_. It was originally
conceived to provide an implementation of WebGPU for Firefox, and 
settled into something that could be shipped on all web browsers.
```

The original feels a bit run-on-ey

cwfitzgerald (approved):
Approved with nit!

GOVERNANCE.md
ErichDonGubler: Implemented as suggested with e71c68132.

crowlKats (approved):

Issue #5644, opened 2024-05-01
Semaphore must not have any pending operations


hecrj:
**Description**
I have updated `iced` to use the latest release (see https://github.com/iced-rs/iced/pull/2417). While `0.19` works perfectly fine, `0.20` is very sluggish and I seem to be constantly hitting a validation error:

```
2024-05-01T14:25:29.145657Z ERROR wgpu_hal::vulkan::instance: VALIDATION [VUID-vkAcquireNextImageKHR-semaphore-01779 (0x5717e75b)]
        Validation Error: [ VUID-vkAcquireNextImageKHR-semaphore-01779 ] Object 0: handle = 0x4fac1c0000000032, type = VK_OBJECT_TYPE_SEMAPHORE; | MessageID = 0x5717e75b | vkAcquireNextImageKHR():  Semaphore must not have any pending operations. The Vulkan spec states: If semaphore is not VK_NULL_HANDLE it must not have any uncompleted signal or wait operations pending (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkAcquireNextImageKHR-semaphore-01779)
2024-05-01T14:25:29.145682Z ERROR wgpu_hal::vulkan::instance:   objects: (type: SEMAPHORE, hndl: 0x4fac1c0000000032, name: ?)
```

The upgrade was straightforward—it only involved adding default `compilation_options` everywhere, so not quite sure what's up.

**Repro steps**
Most examples of the `iced` repository (like the `tour`) in the `wgpu-0.20` branch have the issue.

**Expected vs observed behavior**
It should work well like `0.19` and not produce validation errors.

**Extra materials**
Validation log pasted in description.

**Platform**
Arch Linux, X11, NVIDIA RTX 4090, proprietary drivers.

cwfitzgerald:
This is almost certainly https://github.com/gfx-rs/wgpu/issues/5559

hecrj:
@cwfitzgerald Was this introduced in `0.20`? `0.19` does not have the issue I reported.

I ran `cargo flamegraph` with one of the examples and noticed that almost all of the time is now spent in `Surface::present`:

![flamegraph](https://github.com/gfx-rs/wgpu/assets/518289/f1c0b832-0690-4c85-8cfa-42b9212face2)

hecrj:
Here is `0.19` with the same example, for comparison:

[...]

Issue #5491, opened 2024-04-04
On Nvidia 535, tearing/flickering observed with PresentMode::Fifo


acarl005:
**Description**

I am observing tearing between 2 frames on Linux with X11 and Nvidia 535 drivers.

**Repro steps**

I have modified the "shadow" example in order to accentuate this effect: acarl005/wgpu#1
This branch just renders 2 frames in rapid succession every 1 second.

**Expected vs observed behavior**

I observe the expected behavior on the integrated GPU.

```sh
WAYLAND_DISPLAY="" cargo run --bin wgpu-examples -- shadow
```

https://github.com/gfx-rs/wgpu/assets/8334252/aebd27a6-e492-49cc-b327-5911e4e3193f

However, when I switch to my Nvidia GPU, I instead see tearing and/or flickering.

```sh
WAYLAND_DISPLAY="" WGPU_POWER_PREF=high cargo run --bin wgpu-examples -- shadow
```

https://github.com/gfx-rs/wgpu/assets/8334252/5bfd1835-d4d4-402f-8a2d-b1dc9419d680

The examples use `PresentMode::Fifo`, which says:

> "No tearing will be observed."

**Platform**

Forked the current `trunk` branch of wgpu, commit `ed843f802974eae9d708d3b3895264e9efb01dc6`.

See acarl005/wgpu#1

```
uname -a
```
```
Linux queenie 6.5.0-21-generic #21~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Feb  9 13:32:52 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```
```
nvidia-smi
```
```
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.161.07             Driver Version: 535.161.07   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA GeForce RTX 3050 ...    Off | 00000000:01:00.0 Off |                  N/A |
| N/A   45C    P0              N/A /  25W |      3MiB /  4096MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+

[...]

Issue #5374, opened 2024-03-10
Window resizing lags with white rectangles on Windows


Bromles:
**Description**
When resizing a window on Windows, strange white rectangles underneath it become visible (example in the video in the extra materials). The problem is present with both Vulkan and DX12 backends. Also, I tested the same code on an M1 Mac and it doesn't have this issue. So I think this may be specific to Windows.

**Repro steps**
Despite the fact that the video was recorded on my project, the issue is present even when running unchanged examples from the wgpu repository, for instance, `hello_triangle`.

**Extra materials**

https://github.com/gfx-rs/wgpu/assets/34621477/1b98effb-11cb-478a-9cc2-407fed661a0d

**Platform**
- Windows 11 Enterprise 22H2 (22631.3235)
- Nvidia RTX 3090 Ti
- Studio Driver 551.61
- wgpu 0.19
- winit 0.29

Bromles:
I just tested the `hello_triangle` example on `trunk`, `v0.19`, `v0.18`, `v0.17` and `v0.16` branches, and the problem remains everywhere. It is possible that there is a bug in the drivers, or the problem was there from the beginning

Bromles:
Also, GL backend works just fine even on Windows, the issue persists only on Vulkan and DX12 backends

Bromles:
Made an interesting observation. These are not white rectangles, it looks like a white surface the size of a window that appears in it before the wgpu's surface is rendered. Initially, when the window appears, it is completely filled with white, and only then the image is drawn there. As you make the window smaller, there is a delay in reducing the white surface size, causing it to become visible. However, having decreased once, it does not increase back when the window size increases, and due to this, during subsequent resizes, the lagging part is transparent (if the window has transparency enabled) or black (if not)

I have attached a video illustrating this behavior

https://github.com/gfx-rs/wgpu/assets/34621477/df5a330f-5b41-4e52-8ebd-ba8b3245514c

Wumpf:
Similar to
* #3868
* #3756
* #1168

Gremious:
It happens to me as well, and it's pretty noticable. Found it as I was trying to make an iced app and the window resizing was very slow and laggy, even in release. Then found out the wgpu examples had the same issue.

https://github.com/gfx-rs/wgpu/assets/21365711/2e35276c-6b02-45d7-9a5b-533f867f20ce

(the video is running the example in debug but release still does much the same, just slightly less)

I also get tons of errors in the console:

```
[...]

PR #5681, opened 2024-05-08
[hal/vk] Rework Submission and Surface Synchronization

### LINKED ISSUE #5559: Vulkan Swapchain and Submit Synchronization Incorrect

### LINKED ISSUE #5693: Submitting multiple command encoders per frame causes hang on Vulkan
**Description**
Frames hanging for 5-10 seconds (presumably until timeout).
I bisected the issue and found it was introduced in #4967.

Possibly related to https://github.com/gfx-rs/wgpu/issues/5662

**Repro steps**
It only happens for me when submitting multiple command encoders with render passes in a single frame. 
I don't have a minimal reproduction yet, but I wanted to share in case anyone else has encountered the same issue. 

**Platform**
Ubuntu 22.04, nvidia 4090 - 535.86.10

### LINKED ISSUE #5828: FIFO on Vulkan showing old frames on high refresh rate monitors
**Description**
FIFO and FIFO_RELAXED for Vulkan are displaying old frames out of order, causing jitter. This happens if my monitor is set to 144hz, but if I lock it to 60 in windows it doesn't happen. I found this while working on my Bevy game, and was able to reproduce it in the Boids example. The Bevy reproduction even shows the frame count rendered on screen and you can see it go 3317, 3319, 3321, 3320. In the Boids gif attached below, it goes 1,2,3,4,3,5, no skipping a frame outright. 

**Repro steps**
Run Boids example. 
It can be harder to tell vs normal frame pacing issue, so it can make it easier to record and review to footage frame by frame

Expected: To not see a frame repeated after a newer one has been shown.
Observed: Frames going backwards.

**Extra materials**
Old_frame
Old_frame_3
Trying to find other people running into this issue and I found this. https://forums.blurbusters.com/viewtopic.php?t=13358. 0 idea if it's useful.

**Platform**
Windows 11 build 22631, Nvidia 4090 running 555.99. Found on wgpu 0.17.2 and 0.20.1

cwfitzgerald:
**Connections**

* Closes #5559
* Closes #5693 
* Maybe #4919
* Maybe #5374
* Maybe #5491
* Probably not #3452.

**Description**

As described in #5559, our submission and surface synchronization was totally messed up. I've revamped it, made smarter classes to help keep all the bookkeeping straight, and generally make the code cleaner and easier to understand.

**Testing**

works on my machine lel

### DIFF
--- a/wgpu-hal/src/lib.rs
+++ b/wgpu-hal/src/lib.rs
@@ -459,50 +459,101 @@ pub trait Instance: Sized + WasmNotSendSync {
 pub trait Surface: WasmNotSendSync {
     type A: Api;
 
-    /// Configures the surface to use the given device.
+    /// Configure `self` to use `device`.
     ///
     /// # Safety
     ///
-    /// - All gpu work that uses the surface must have been completed.
+    /// - All GPU work using `self` must have been completed.
     /// - All [`AcquiredSurfaceTexture`]s must have been destroyed.
     /// - All [`Api::TextureView`]s derived from the [`AcquiredSurfaceTexture`]s must have been destroyed.
-    /// - All surfaces created using other devices must have been unconfigured before this call.
+    /// - The surface `self` must not currently be configured to use any other [`Device`].
     unsafe fn configure(
         &self,
         device: &<Self::A as Api>::Device,
         config: &SurfaceConfiguration,
     ) -> Result<(), SurfaceError>;
 
-    /// Unconfigures the surface on the given device.
+    /// Unconfigure `self` on `device`.
     ///
     /// # Safety
     ///
-    /// - All gpu work that uses the surface must have been completed.
+    /// - All GPU work that uses `surface` must have been completed.
     /// - All [`AcquiredSurfaceTexture`]s must have been destroyed.
     /// - All [`Api::TextureView`]s derived from the [`AcquiredSurfaceTexture`]s must have been destroyed.
-    /// - The surface must have been configured on the given device.
+    /// - The surface `self` must have been configured on `device`.
     unsafe fn unconfigure(&self, device: &<Self::A as Api>::Device);
 
-    /// Returns the next texture to be presented by the swapchain for drawing
+    /// Return the next texture to be presented by `self`, for the caller to draw on.
     ///
-    /// A `timeout` of `None` means to wait indefinitely, with no timeout.
+    /// On success, return an [`AcquiredSurfaceTexture`] representing the
+    /// texture into which the caller should draw the image to be displayed on
+    /// `self`.
+    ///
+    /// If `timeout` elapses before `self` has a texture ready to be acquired,
+    /// return `Ok(None)`. If `timeout` is `None`, wait indefinitely, with no
+    /// timeout.
+    ///
+    /// # Using an [`AcquiredSurfaceTexture`]
+    ///
+    /// On success, this function returns an [`AcquiredSurfaceTexture`] whose
+    /// [`texture`] field is a [`SurfaceTexture`] from which the caller can
+    /// [`borrow`] a [`Texture`] to draw on. The [`AcquiredSurfaceTexture`] also
+    /// carries some metadata about that [`SurfaceTexture`].
+    ///
+    /// All calls to [`Queue::submit`] that draw on that [`Texture`] must also
+    /// include the [`SurfaceTexture`] in the `surface_textures` argument.
+    ///
+    /// When you are done drawing on the texture, you can display it on `self`
+    /// by passing the [`SurfaceTexture`] and `self` to [`Queue::present`].
+    ///
+    /// If you do not wish to display the texture, you must pass the
+    /// [`SurfaceTexture`] to [`self.discard_texture`], so that it can be reused
+    /// by future acquisitions.
     ///
     /// # Portability
     ///
-    /// Some backends can't support a timeout when acquiring a texture and
-    /// the timeout will be ignored.
-    ///
-    /// Returns `None` on timing out.
+    /// Some backends can't support a timeout when acquiring a texture. On these
+    /// backends, `timeout` is ignored.
     ///
     /// # Safety
     ///
-    /// - The fence must be the same fence passed to all [`Queue::submit`]s
-    ///   that used textures acquired from this surface.
+    /// - The surface `self` must currently be configured on some [`Device`].
+    ///
+    /// - The `fence` argument must be the same [`Fence`] passed to all calls to
+    ///   [`Queue::submit`] that used [`Texture`]s acquired from this surface.
+    ///
+    /// - You may only have one texture acquired from `self` at a time. When
+    ///   `acquire_texture` returns `Ok(Some(ast))`, you must pass the returned
+    ///   [`SurfaceTexture`] `ast.texture` to either [`Queue::present`] or
+    ///   [`Surface::discard_texture`] before calling `acquire_texture` again.
+    ///
+    /// [`texture`]: AcquiredSurfaceTexture::texture
+    /// [`SurfaceTexture`]: Api::SurfaceTexture
+    /// [`borrow`]: std::borrow::Borrow::borrow
+    /// [`Texture`]: Api::Texture
+    /// [`Fence`]: Api::Fence
+    /// [`self.discard_texture`]: Surface::discard_texture
     unsafe fn acquire_texture(
         &self,
         timeout: Option<std::time::Duration>,
         fence: &<Self::A as Api>::Fence,
     ) -> Result<Option<AcquiredSurfaceTexture<Self::A>>, SurfaceError>;
+
+    /// Relinquish an acquired texture without presenting it.
+    ///
+    /// After this call, the texture underlying [`SurfaceTexture`] may be
+    /// returned by subsequent calls to [`self.acquire_texture`].
+    ///
+    /// # Safety
+    ///
+    /// - The surface `self` must currently be configured on some [`Device`].
+    ///
+    /// - `texture` must be a [`SurfaceTexture`] returned by a call to
+    ///   [`self.acquire_texture`] that has not yet been passed to
+    ///   [`Queue::present`].
+    ///
+    /// [`SurfaceTexture`]: Api::SurfaceTexture
+    /// [`self.acquire_texture`]: Surface::acquire_texture
     unsafe fn discard_texture(&self, texture: <Self::A as Api>::SurfaceTexture);
 }
 
@@ -815,6 +866,9 @@ pub trait Queue: WasmNotSendSync {
     /// - Each [`SurfaceTexture`][st] in `surface_textures` must be configured
     ///   for use with the [`Device`][d] associated with this [`Queue`],
     ///   typically by calling [`Surface::configure`].
+    ///
+    /// - All calls to this function that include a given [`SurfaceTexture`][st]
+    ///   in `surface_textures` must use the same [`Fence`].
     ///
     /// [`Fence`]: Api::Fence
     /// [cb]: Api::CommandBuffer
--- a/wgpu-hal/src/lib.rs
+++ b/wgpu-hal/src/lib.rs
@@ -493,9 +493,15 @@ pub trait Surface: WasmNotSendSync {
     /// the timeout will be ignored.
     ///
     /// Returns `None` on timing out.
+    ///
+    /// # Safety
+    ///
+    /// - The fence must be the same fence passed to all [`Queue::submit`]s
+    ///   that used textures acquired from this surface.
     unsafe fn acquire_texture(
         &self,
         timeout: Option<std::time::Duration>,
+        fence: &<Self::A as Api>::Fence,
     ) -> Result<Option<AcquiredSurfaceTexture<Self::A>>, SurfaceError>;
     unsafe fn discard_texture(&self, texture: <Self::A as Api>::SurfaceTexture);
 }
@@ -762,7 +768,7 @@ pub trait Queue: WasmNotSendSync {
 
     /// Submit `command_buffers` for execution on GPU.
     ///
-    /// If `signal_fence` is `Some(fence, value)`, update `fence` to `value`
+    /// If `signal_fence` is `(fence, value)`, update `fence` to `value`
     /// when the operation is complete. See [`Fence`] for details.
     ///
     /// If two calls to `submit` on a single `Queue` occur in a particular order
@@ -822,7 +828,7 @@ pub trait Queue: WasmNotSendSync {
         &self,
         command_buffers: &[&<Self::A as Api>::CommandBuffer],
         surface_textures: &[&<Self::A as Api>::SurfaceTexture],
-        signal_fence: Option<(&mut <Self::A as Api>::Fence, FenceValue)>,
+        signal_fence: (&mut <Self::A as Api>::Fence, FenceValue),
     ) -> Result<(), DeviceError>;
     unsafe fn present(
         &self,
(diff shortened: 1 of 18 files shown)


JMS55:
Works fine for me running the cube example with VK validation layers enabled
```
[2024-05-09T06:03:33Z INFO  wgpu_core::instance] Adapter Vulkan AdapterInfo { name: "Intel(R) Arc(TM) Graphics", vendor: 32902, device: 32085, device_type: IntegratedGpu, driver: "Intel Corporation", driver_info: "101.5382", backend: Vulkan }
```

Vecvec:
This fixes an annoying access violation termination (that I thought was a driver bug) on program exit after a present.

hecrj:
Unfortunately, it does not fix #5644 for me. Same constant validation error on presentation:

```
[2024-05-09T08:16:24Z ERROR wgpu_hal::vulkan::instance] VALIDATION [VUID-vkAcquireNextImageKHR-semaphore-01779 (0x5717e75b)]
        Validation Error: [ VUID-vkAcquireNextImageKHR-semaphore-01779 ] Object 0: handle = 0xba7514000000002a, type = VK_OBJECT_TYPE_SEMAPHORE; | MessageID = 0x5717e75b | vkAcquireNextImageKHR():  Semaphore must not have any pending operations. The Vulkan spec states: If semaphore is not VK_NULL_HANDLE it must not have any uncompleted signal or wait operations pending (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkAcquireNextImageKHR-semaphore-01779)
[2024-05-09T08:16:24Z ERROR wgpu_hal::vulkan::instance]         objects: (type: SEMAPHORE, hndl: 0xba7514000000002a, name: ?)
```

The actual validation error only happens with a `debug` build, although `release` feels much slower than `0.19` as well. I believe something is very slow and the errors actually only show up in `debug` mode.

Flamegraph of `cube` seems to indicate this—with `present` taking a huge amount of time:

flamegraph

For comparison, same drivers with `0.19`:

flamegraph

Even if I enable `Immediate` present mode, frames seem to take twice as much time to be rendered with `0.20`.

cwfitzgerald:
@hecrj could you try again with these latest commits, someone else who can reproduce the validation error says that this fixed it.

hecrj:
@cwfitzgerald Yep! The latest changes fix the validation errors. Thanks!

I'm still noticing a bit of worse performance in `debug` mode, but it doesn't seem to be there in `release` mode anymore. Flamegraphs look the same as before, but I imagine it's expected since now synchronization works properly and there may be more waiting. No errors in any case!

wgpu-hal/src/lib.rs
```diff
@@ -486,9 +486,15 @@ pub trait Surface: WasmNotSendSync {
     /// the timeout will be ignored.
     ///
     /// Returns `None` on timing out.
+    ///
+    /// # Safety
+    ///
+    /// - The fence must be the same fence passed to all [`Queue::submit`]s
+    ///   that used textures acquired from this surface.
```
jimblandy: This is unclear. I'm using textures acquired from this surface all the time, with many different submissions.

wgpu-hal/src/vulkan/mod.rs
```diff
@@ -149,6 +140,103 @@ pub struct Instance {
     shared: Arc<InstanceShared>,
 }
 
+/// The semaphores used to synchronize the swapchain image acquisition.
+///
+/// One of these is created per swapchain image and is used for the swapchain
```
jimblandy: This is unclear. A swapchain has several images. Each has its own `SwapchainSemaphores`. What does it mean for all of them to be "used for the swapchain"?

jimblandy:
I'm actually rewriting these comments as I review so if you're not attached to them then don't worry too much about fixing anything.

cwfitzgerald:
Yeah that's fine - I was just trying to get something there.

wgpu-hal/src/lib.rs
```diff
@@ -675,7 +681,7 @@ pub trait Queue: WasmNotSendSync {
 
     /// Submit `command_buffers` for execution on GPU.
     ///
-    /// If `signal_fence` is `Some(fence, value)`, update `fence` to `value`
+    /// If `signal_fence` is `(fence, value)`, update `fence` to `value`
     /// when the operation is complete. See [`Fence`] for details.
```
Wumpf:
```suggestion
    /// Will update `fence` to `value` when the operation is complete.
    /// See [`Fence`] for details.
```

wgpu-hal/src/vulkan/mod.rs
```diff
@@ -359,18 +463,70 @@ pub struct Device {
     render_doc: crate::auxil::renderdoc::RenderDoc,
 }
 
+/// Semaphores that a given submission should wait on and signal.
+struct RelaySemaphoreState {
+    wait: Option<vk::Semaphore>,
+    signal: vk::Semaphore,
```
jimblandy: My reading is that you can actually just use one semaphore to order submits, for both the signal and wait operations.

The rules I can find about signaling and waiting just say they have to happen in the natural order, and a submission will definitely wait before signalling. So the semaphore will always be in the right state for the next operation.

It's not a problem just to enqueue multiple operations that will wait and signal the same semaphore - if it were, then you couldn't handle an arbitrary number of submissions just by swapping two semaphores back and forth. Certainly when you're doing that, half the submissions are waiting on the same semaphore. So if that's permitted, you might as well just use one semaphore for everything.

The text of `VUID-VkSubmitInfo2-semaphore-03881` explicitly mentions the possibility of semaphores appearing in both lists on the same submit:

> If the same semaphore is used as the semaphore member of both an element of `pSignalSemaphoreInfos` and `pWaitSemaphoreInfos`, and that semaphore is a timeline semaphore, the value member of the `pSignalSemaphoreInfos` element must be greater than the value member of the `pWaitSemaphoreInfos` element

That's describing timeline semaphores, but it shows that the authors are expecting this possibility. And there's no VUID saying that the same semaphore can't appear in both lists.

jimblandy:
@cwfitzgerald What do you think of this? https://github.com/gfx-rs/wgpu/commit/da543fb51

My hope was that this would make the states of `RelaySemaphores` a bit clearer to a new reader. If people should not use `wait` unless `should_wait` is true, then let's actually force people to check that condition before they can even get at the value: classic `Option`. But it does make `advance` fallible and require us to pass the device.

[edit: made this a little nicer still: `RelaySemaphoreState` is no longer necessary]

jimblandy:
@cwfitzgerald Here's an even more radical suggestion: https://github.com/gfx-rs/wgpu/commit/b5d52fed45e4f89213e6fe5870e99126865711aa

Just use a single semaphore for all queue ordering.

cwfitzgerald:
@jimblandy i think we can go for it, the only question is if we want to still avoid this deadlock in anv mentioned in the old code, it if we consider that old enough to ignore.    

```
/// It would be correct to use a single semaphore there, but 
/// [Intel hangs in `anv_queue_finish`](https://gitlab.freedesktop.org/mesa/mesa/-/issues/5508).
```

I should have perseved that comment somehow.

jimblandy:
@cwfitzgerald:

> I should have perseved that comment somehow.

Ahh, yes, I was reviewing by checking out the branch and just wandering around with rust-analyzer, so I didn't notice that the patch removed the comment. If that's the sole rationale for the entire semaphore-swapping doohickey then we had definitely better have the comment.

Kvark filed that bug in Oct 2021. That's not that long ago, so I think we'd better keep it. It sounds like a bear to debug, and I'd rather not have to debug it again.

jimblandy:
So I guess the remaining question is what you think of [this one](https://github.com/gfx-rs/wgpu/pull/5681#issuecomment-2118327977).

jimblandy:
So here's the latest version of my suggestion:
https://github.com/jimblandy/wgpu/compare/hal-vk-wait-semaphore-is-option%7E2...jimblandy:wgpu:hal-vk-wait-semaphore-is-option

cwfitzgerald:
Looks fine to me, could you either PR to my fork or just push to it?

jimblandy:
I still want to get a handle on the swapchain semaphores. I've got some work-in-progress docs for those too. I'll take care of this tomorrow morning.

wgpu-hal/src/vulkan/mod.rs
```diff
@@ -149,6 +140,99 @@ pub struct Instance {
...
+/// One of these is created per swapchain image and is used for the swapchain
+///
+/// The `acquire` semaphore is used to signal from the acquire operation to the first submit
+/// that uses the image.
+///
+/// The `present` semaphores are used to signal from each submit that uses the image to the
+/// present operation. We need use one per submit as we don't know at submit time which
+/// submission will be the last to use the image, so we add semaphores to them all.
+#[derive(Debug)]
+struct SwapchainSemaphores {
```
jimblandy: Could we rename this `SwapchainImageSemaphores`, since it's per image, not per swapchain?

wgpu-hal/src/vulkan/mod.rs
```diff
@@ -702,58 +871,85 @@ impl crate::Queue for Queue {
...
+                }
+                check.len() == surface_textures.len()
+            },
+            "More than one surface texture is being used from the same swapchain. This will cause a deadlock in release."
+        );
 
-        for &surface_texture in surface_textures {
-            wait_stage_masks.push(vk::PipelineStageFlags::TOP_OF_PIPE);
-            wait_semaphores.push(surface_texture.wait_semaphore);
+        // We lock access to all of the semaphores. This may block if two submissions are in flight at the same time.
```
jimblandy: Please check my understanding, but I think this can't actually block. All submissions using any given `SurfaceTexture` are required to use the same `Fence`, and as explained in the docs for `wgpu_hal::Queue::submit`, all submissions using the same `Fence` must be synchronized to occur in a particular order.

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -1011,12 +1016,28 @@ impl crate::Surface for super::Surface {
...
+        let locked_swapchain_semaphores = swapchain_semaphores_arc
+            .try_lock()
+            .expect("Failed to lock a SwapchainSemaphores.");
+
+        // Wait for the previously acquired image used by the semaphore to be available.
+        swapchain.device.wait_for_fence(
+            fence,
+            locked_swapchain_semaphores.previously_used_submission_index,
+            timeout_ns,
+        )?;
```
jimblandy: What does this actually accomplish? Aren't we going to wait for the `acquire` semaphore anyway?

jimblandy:
@cwfitzgerald I just pushed two doc commits. Could you look them over for accuracy?

wgpu-hal/src/vulkan/instance.rs
cwfitzgerald: We must wait until the cpu timeline is sure that the semaphore has been signaled and reset. Otherwise there will be two acquire operations that could signal the same semaphore.

wgpu-hal/src/vulkan/mod.rs
cwfitzgerald: Sure

wgpu-hal/src/vulkan/mod.rs
```diff
@@ -793,19 +1058,20 @@ impl crate::Queue for Queue {
     ) -> Result<(), crate::SurfaceError> {
         let mut swapchain = surface.swapchain.write();
         let ssc = swapchain.as_mut().unwrap();
+        let mut swapchain_semaphores = texture.surface_semaphores.lock();
+
+        // debug_assert_eq!(
+        //     Arc::as_ptr(&texture.surface_semaphores),
+        //     Arc::as_ptr(&ssc.surface_semaphores[ssc.next_semaphore_index]),
+        //     "Trying to use a surface texture that does not belong to the current swapchain."
+        // );
```
jimblandy: This should either be uncommented or removed, but we shouldn't leave it in limbo like this.

wgpu-hal/src/vulkan/instance.rs
jimblandy: So, in other words:

If we were to remove this wait, then acquisition, command submission, and presentation would all set stuff in motion in the presentation engine or on the GPU, but as far as the CPU is concerned, they'd all return without blocking, so the CPU could just run the process full speed ahead, blow through the entire swapchain, come around to the front of the ring buffer again, and make a fresh call to `vkAcquireNextImage` passing the same semaphore as last time. This wait is the only thing that rate-limits that process to match actual presentation speed.

Is that correct?

wgpu-hal/src/vulkan/mod.rs
jimblandy: As discussed in Matrix (my bad), technically what I said above is correct, but we want to keep the semaphore-swapping code for the time being because of https://gitlab.freedesktop.org/mesa/mesa/-/issues/5508 .

wgpu-hal/src/vulkan/instance.rs
jimblandy: I've pushed another commit documenting this. If my explanation is correct, please mark this as resolved.

wgpu-hal/src/lib.rs
jimblandy: This makes more sense to me now. Resolving.

wgpu-hal/src/lib.rs
jimblandy: I've incorporated a change similar to this, so I'm going to resolve this.

jimblandy:
rebased on current trunk

wgpu-hal/src/vulkan/mod.rs
cwfitzgerald: Lets remove it.

wgpu-hal/src/vulkan/instance.rs
cwfitzgerald: > This wait is the only thing that rate-limits that process to match actual presentation speed.

Yes this is correct.

wgpu-hal/src/vulkan/mod.rs
cwfitzgerald: Ah yes, good point, then lets move these all to try_lock.

jimblandy (approved):

PR #5688, opened 2024-05-11
Add support for pipeline-overridable constants in web backend

DouglasDwyer:
**Connections**
https://github.com/gfx-rs/wgpu/pull/5500 - Pipeline-overridable constants

**Description**
I was excited to try out the new pipeline-overridable constants that were added in version 0.20, but unfortunately, I discovered that they didn't work on the web! This change copies the `constants` dictionary into the Javascript-side pipeline descriptors so that pipeline-overridable constants work with WebGPU.

**Checklist**

- [x] Run `cargo fmt`.


wgpu/src/backend/webgpu.rs
```diff
@@ -1874,6 +1875,11 @@ impl crate::context::Context for ContextWebGpu {
         let module: &<ContextWebGpu as crate::Context>::ShaderModuleData =
             downcast_ref(desc.vertex.module.data.as_ref());
         let mut mapped_vertex_state = webgpu_sys::GpuVertexState::new(&module.0.module);
+        let _ = js_sys::Reflect::set(
+            &mapped_vertex_state,
+            &"constants".into(),
+            &hashmap_to_jsvalue(desc.vertex.compilation_options.constants),
+        );
```
Wumpf: I don't like ignoring the error on this. If the WebGPU api still changes here (unlikely but who knows) or there's a bug here we'll never hear notice that the constants have been swallowed.
It should be possible to create a utility that encompasses all three callsites which internally does nothing if no constants are passed and otherwise uses js_sys::Reflect with error logging if something goes wrong.

I digged around a bit why `constants` is not exposed by wasmbindgen: It is mentioned here https://github.com/rustwasm/wasm-bindgen/blob/main/crates/web-sys/webidls/unstable/WebGPU.webidl#L635 but not exposed because of https://github.com/rustwasm/wasm-bindgen/issues/3587

A note linking to this issues should go into that utility as well (or each of these usage sites if utility method doesn't work out)

Wumpf (changes requested):
thank you for looking into this!

wgpu/src/backend/webgpu.rs
DouglasDwyer: That's totally fair - I have added a utility method as requested!

wgpu/src/backend/webgpu.rs
```diff
@@ -3808,3 +3816,32 @@ impl Drop for BufferMappedRange {
...
+/// exposed by `wasm-bindgen`. See the following issues for details:
+/// - [gfx-rs/wgpu#5688](https://github.com/gfx-rs/wgpu/pull/5688)
+/// - [rustwasm/wasm-bindgen#3587](https://github.com/rustwasm/wasm-bindgen/issues/3587)
+fn insert_constants_map(target: &JsValue, map: &HashMap<String, f64>) {
+    if !map.is_empty() {
+        let result = js_sys::Reflect::set(target, &"constants".into(), &hashmap_to_jsvalue(map));
+
+        if let Err(error) = result {
+            log::error!("Failed to set constants map for pipeline: {error:?}");
+        }
```
cwfitzgerald:
```suggestion
        js_sys::Reflect::set(target, &"constants".into(), &hashmap_to_jsvalue(map))
            .expect("Setting the values in a Javascript object should never fail");
```

I don't think we need to deal with the error case.

cwfitzgerald (dismissed):

cwfitzgerald:

wgpu/src/backend/webgpu.rs
Wumpf: not a very strong stance here but I'd rather we don't do `expect` and keep the `log` (I'm guilty of suggesting this earlier on the PR). Is it likely this ever breaks? No, very very unlikely! But an `expect` already implies checking on the error (so does unwrap) and I'd rather not have us unrecoverable crash over this ever.

(and yes I hedge similar concerns about _a lot_ of code in wgpu and wgpu-core ;). Maybe we should talk this out at some point to have more unified stance on this stuff going forward. But again this specific instance is like one of the least concerning ones, so just a leaning here 🤷 )

Wumpf (approved):
Thanks for the neatly documented util, much nicer :)

🟢 from my side, but let's hear out @cwfitzgerald now that he's in on the fun :)

wgpu/src/backend/webgpu.rs
cwfitzgerald: > But an expect already implies checking on the error (so does unwrap) and I'd rather not have us unrecoverable crash over this ever.

I would much rather an expect than a log that everyone will ignore - more specifically, this code should not _ever_ fail unless something horribly wrong with the js reflect api exists. I agree we shouldn't unwrap if there's a chance for the user to want to fail-and-try-again.

cwfitzgerald (approved):
Yeet!

PR #5690, opened 2024-05-11
Reduce string allocations related to labels and logging.

kpreid:
**Description**
* `Resource::label()` is (occasionally, for destroyed resources specifically) called for logging, but always allocated a `String`, which would not necessarily be skipped by the logging macro (if the level filter is met, a `format_args!()` is constructed even if it is skipped). Avoid that by making it return `&str` instead.
    * This has one behavior change: `wgpu_core::command::CommandBuffer` will return the `"<CommandBuffer>"` placeholder string instead of `""`. This avoids taking a lock in addition to the allocation, but it also means `Registry::label_for_resource()` won't produce a placeholder with ID. But if that behavior is desirable, surely it should be done more broadly (have `Resource::label` return empty strings consistently), instead of only for `CommandBuffer`?
* Also avoid allocating a throwaway `String` when creating an unlabeled `CommandBuffer`.

**Testing**

Ran `cargo xtask test`. All failures on my machine are unrelated to label strings as far as I know.

**Checklist**

- [X] Run `cargo fmt`.
- [X] ~~Add change to `CHANGELOG.md`. See simple instructions inside file.~~ Not worth mentioning.

### DIFF
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -514,8 +514,8 @@ impl<A: HalApi> Resource for BindGroupLayout<A> {
         &mut self.info
     }
 
-    fn label(&self) -> String {
-        self.label.clone()
+    fn label(&self) -> &str {
+        &self.label
     }
 }
 impl<A: HalApi> BindGroupLayout<A> {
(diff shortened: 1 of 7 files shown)


cwfitzgerald (approved):
Nice

PR #5691, opened 2024-05-11
Avoid introducing spurious features for optional dependencies

bjorn3:
**Connections**
None

**Description**
If a feature depends on an optional dependency without using the dep: prefix, a feature with the same name as the optional dependency is introduced. This feature almost certainly won't have any effect when enabled other than increasing compile times and polutes the feature list shown by cargo add. Consistently use dep: for all optional dependencies to avoid this problem.

**Testing**

**Checklist**

- [x] Run `cargo fmt`.


Wumpf (approved):
Didn't know this was a thing! I tried before after `cargo add --path <my wgpu path> wgpu` and indeed the old version lists `naga` and `hal` when it really shouldn't.
Nice, thank you!

fintelia:
FYI: This is likely to cause breakage for some downstream users when they upgrade to the next release so it could be worth calling out more clearly in the CHANGELOG.

Specifically, the `naga` feature used to be required for anyone who wanted to use the `wgpu::naga` re-export with the 0.18 or earlier versions. It became a no-op in the 0.19 release, but many people upgrading probably didn't notice. For instance, [Bevy still enables the feature today ](https://github.com/bevyengine/bevy/blob/19d078c60953858d5cb7e006c5d90110712280a4/crates/bevy_render/Cargo.toml#L77) even though they target 0.20.

PR #5694, opened 2024-05-12
Add Benchmarks

### LINKED ISSUE #5443: wgpu needs benchmarks

cwfitzgerald:
**Connections**

Closes #5443 

**Description**

This is our first useful foray into benchmarking.

Adds benchmarking infrastructure and adds some more labeling within the code to make traces more useful.

image

I will write up some of my findings from this in an issue later. 

**Testing**

Should be compile and run tested by existing infra.


JMS55:
Can I please request some benchmarks? 

1. Making a very large (>500MB, 2GB is the limit iirc) buffer, then making a second, slightly bigger one, and destroying the original. My project allocates a lot of _very_ large buffers, but when I need to grow them I make a bigger one, copyBuffer to transfer the contents, and then destroy the old. I'm very curious how wgpu's memory allocators handle this.
2. Bind group creation benchmarks (currently we don't cache these between frames).

cwfitzgerald:
1. Should already be covered by the buffer creation benchmark
2. Can add that, yeah

benches/README.md
```diff
@@ -0,0 +1,91 @@
...
+## Usage
+
+```sh
+# Run all benchmarks
+cargo bench -p wgpu-benchmark
+# Run a specific benchmarks that contains "filter" in its name
+cargo bench -p wgpu-benchmark -- "filter"
+```
+
+## Benchmarks
```
nical, line 17: Nit: I see there is also a shader benchmark that should be mentioned here.

nical (approved):

PR #5700, opened 2024-05-13
feature: SHADER-F16

FL33TW00D:


FL33TW00D:
Closing in favour of #5701 as you need a directive to test with.

PR #5701, opened 2024-05-13
feat: implement F16 support in shaders

### LINKED ISSUE #4384: [meta] f16 support
From WGSL spec: https://gpuweb.github.io/gpuweb/wgsl/#floating-point-types

TODO

- [x] https://github.com/gfx-rs/wgpu/pull/5476
- [ ] wgsl-in
- [ ] validator
- [ ] msl-out (@FL33TW00D is interested in contributing this one)
- [ ] hlsl-out
- [ ] spv-out

### LINKED ISSUE #5476: WGSL: Support the `enable` directive
The WGSL specification has introduced the `enable` directive for to implementation-specific extensions (see https://www.w3.org/TR/WGSL/#enable-extensions-sec). Currently naga rejects these as an invalid token, for example:

```
enable f16;
^^^^^^ expected global item ('struct', 'const', 'var', 'alias', ';', 'fn') or the end of the file
```

Even if naga doesn't support a requested extension, it should recognize the `enable` directive. In my specific use case, naga is used to parse and preprocess WGSL which may end up getting passed on to a different implementation (i.e. Dawn/Chromium). It would be nice to be able to declare Chromium-specific enable directives for the wgsl-in -> wgsl-out case.

FL33TW00D:
**Connections**

- Hard dependency on <https://github.com/gfx-rs/wgpu/pull/6424>.
- Resolves https://github.com/gfx-rs/wgpu/issues/5476.
- Resolves https://github.com/gfx-rs/wgpu/issues/4384.

**Checklist for merge-ability**

- [x] Test multiple redundant directives (I think there may be a cleaner way to do this: https://github.com/gfx-rs/wgpu/pull/5701/files#diff-9ef835ffc5f7d7313243db5bfa371d3115b82c0705d4f1df75c98dfe07d927a1R2175)
- [x] MSL out
- [x] HLSL out
- [x] SPV out
- [x] Arbitrary for F16 on half-rs: https://github.com/starkat99/half-rs/pull/111. Currently we are using a git dep on my fork which is causing CI to fail.
- [x] ~hexf_parse f16 support: https://github.com/lifthrasiir/hexf. This one won't be fun. Did some work on it here: https://github.com/lifthrasiir/hexf/compare/master...FL33TW00D:hexf:feature/f16 .~ **Deferred to #7046**
- [x] We should make sure that we are properly validating As casts of matrices, because that code currently assumes that all matrices have the same kind (https://github.com/gfx-rs/wgpu/issues/6441).
- [x] Validate out `f16` inside push constants.
- [x] Continue tests for f16 layout.
- [x] wgsl-out needs to enable f16
- [x] Fix spirv output of matrix casts 
- [x] Aliases for f16 types still don't seem to work
- [x] Split out out-of-scope commits 

**Additional context**

F16 is available in >=SM6.2: https://github.com/microsoft/DirectXShaderCompiler/wiki/16-Bit-Scalar-Types


naga/src/front/wgsl/error.rs
```diff
@@ -791,6 +790,11 @@ impl<'a> Error<'a> {
                 )],
                 notes: vec![],
             },
+            Error::UnknownExtension(span, word) => ParseError {
+                message: format!("Unknown extension: {}. See available extensions at: https://www.w3.org/TR/WGSL/#enable-extension", word),
```
FL33TW00D: Would be nice to list supported extensions here but don't want to bring in `strum`.

examples/src/shader_f16/shader.wgsl
```diff
@@ -0,0 +1,7 @@
+@group(0) @binding(0)
```
FL33TW00D: Should result in a https://www.w3.org/TR/WGSL/#shader-creation-error as no enable directive. Sounds like this should be done in parsing but don't want to affect performance by checking if `f16` for every scalar.

naga/src/front/wgsl/error.rs
ErichDonGubler: Sounds like a good thing to consider in follow-up, perhaps?

FL33TW00D: @ErichDonGubler Sounds good to me!

examples/src/shader_f16/shader.wgsl
FL33TW00D: Resolved in [34c9ef9](https://github.com/gfx-rs/wgpu/pull/5701/commits/34c9ef9c13798d85e5eb1cc5b98aef762a7b336d)

Suggestions on a more elegant way to do it are encouraged.

image

FL33TW00D:
I've marked this as ready for review, as the wgpu specific logic is implemented and would be great to start iterating on it.

2 main blockers are:
1. Merging of the PR on `half-rs`.
2. `hexf-parse` improved implementation.

naga/src/proc/constant_evaluator.rs
```diff
@@ -1186,15 +1196,20 @@ impl<'a> ConstantEvaluator<'a> {
                 )
             }
             crate::MathFunction::Step => {
-                component_wise_float!(self, span, [arg, arg1.unwrap()], |edge, x| {
-                    Ok([if edge <= x { 1.0 } else { 0.0 }])
-                })
+                //component_wise_float!(self, span, [arg, arg1.unwrap()], |edge, x| {
+                //    Ok([if edge <= x { 1.0 } else { 0.0 }])
+                //})
+                todo!()
```
FL33TW00D: TODO (any pointers here appreciated)

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2167,6 +2167,34 @@ impl Parser {
...
+                            kind: ast::GlobalDirectiveKind::Enable(ast::EnableDirective {
+                                //TODO: LIST should allow multiple extensions
+                                enable_extension_list: vec![extension],
+                            }),
+                        },
+                        name_span,
+                    );
+                    lexer.expect(Token::Separator(';'))?;
+                }
+                other => panic!("Unexpected token: {:?}", other),
```
FL33TW00D: Shouldn't panic

FL33TW00D:
Dogfooding works! Fixed a few bugs in https://github.com/gfx-rs/wgpu/pull/5701/commits/30e12b5c4b68143a8fe037d93753215f34855056. Still waiting on upstream.

naga/src/proc/constant_evaluator.rs
teoxoy: We can match like in the other places? Looking at the spec we should error if the types of the 2 args are different.

naga/src/proc/constant_evaluator.rs
```diff
@@ -1467,7 +1485,7 @@ impl<'a> ConstantEvaluator<'a> {
                         Literal::U32(v) => v as i32,
                         Literal::F32(v) => v as i32,
                         Literal::Bool(v) => v as i32,
-                        Literal::F64(_) | Literal::I64(_) | Literal::U64(_) => {
+                        Literal::F64(_) | Literal::I64(_) | Literal::U64(_) | Literal::F16(_) => {
```
teoxoy: Looking at the spec this should work.

naga/src/proc/constant_evaluator.rs
```diff
@@ -1478,7 +1496,7 @@ impl<'a> ConstantEvaluator<'a> {
                         Literal::U32(v) => v,
                         Literal::F32(v) => v as u32,
                         Literal::Bool(v) => v as u32,
-                        Literal::F64(_) | Literal::I64(_) | Literal::U64(_) => {
+                        Literal::F64(_) | Literal::I64(_) | Literal::U64(_) | Literal::F16(_) => {
```
teoxoy: This too.

naga/src/proc/constant_evaluator.rs
```diff
@@ -1492,6 +1510,11 @@ impl<'a> ConstantEvaluator<'a> {
                         Literal::F64(v) => v as i64,
                         Literal::I64(v) => v,
                         Literal::U64(v) => v as i64,
+                        Literal::F16(v) => f16::to_i64(&v).ok_or(
+                            ConstantEvaluatorError::AutomaticConversionFloatToInt {
+                                to_type: "i64",
+                            },
+                        )?,
```
teoxoy: Looks like `to_i64` returns `None` only when the value is a `NaN` or `Inf`. But we already error on floats that are `NaN` or `Inf`. I think this should be an unwrap or internal error.

naga/src/proc/constant_evaluator.rs
```diff
@@ -1503,9 +1526,29 @@ impl<'a> ConstantEvaluator<'a> {
                         Literal::F64(v) => v as u64,
                         Literal::I64(v) => v as u64,
                         Literal::U64(v) => v,
+                        Literal::F16(v) => f16::to_u64(&v).ok_or(
+                            ConstantEvaluatorError::AutomaticConversionFloatToInt {
+                                to_type: "u64",
+                            },
+                        )?,
```
teoxoy: Same here.

naga/src/proc/constant_evaluator.rs
```diff
@@ -1503,9 +1526,29 @@ impl<'a> ConstantEvaluator<'a> {
...
+                        Literal::F32(v) => f16::from_f32(v),
+                        Literal::F64(v) => f16::from_f64(v),
+                        Literal::Bool(v) => f16::from_u32(v as u32).unwrap(),
+                        Literal::I64(_)
+                        | Literal::U64(_)
+                        | Literal::U32(_)
+                        | Literal::I32(_)
+                        | Literal::AbstractInt(_) => {
+                            return make_error();
+                        }
```
teoxoy: Looking at the spec these should work.

naga/src/proc/constant_evaluator.rs
```diff
@@ -1535,6 +1580,7 @@ impl<'a> ConstantEvaluator<'a> {
                         Literal::F64(_)
                         | Literal::I64(_)
                         | Literal::U64(_)
+                        | Literal::F16(_)
```
teoxoy: This too.

naga/src/lib.rs
```diff
@@ -970,6 +972,24 @@ pub struct ResourceBinding {
...
+
+/// Enable directive
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[cfg_attr(feature = "serialize", derive(Serialize))]
+#[cfg_attr(feature = "deserialize", derive(Deserialize))]
+#[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
+pub struct EnableDirective {
+    extension: Extension,
+}
+
```
teoxoy: These are only relevant to the WGSL frontend, they should go in the WGSL frontend.

naga/src/back/wgsl/writer.rs
teoxoy, line 1: The WGSL backend would need to enable the f16 extension.

wgpu-hal/src/dx12/adapter.rs
```diff
@@ -347,6 +347,11 @@ impl super::Adapter {
                 && features1.Int64ShaderOps != 0,
         );
 
+        features.set(
+            wgt::Features::SHADER_F16,
+            shader_model >= naga::back::hlsl::ShaderModel::V6_2 && hr == 0,
+        );
```
teoxoy, line 443: We need to also check for `Native16BitShaderOpsSupported`.

We also need to pass the `-enable-16bit-types` flag to DXC if the feature is enabled.

https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_feature_data_d3d12_options4
https://github.com/microsoft/DirectXShaderCompiler/wiki/16-Bit-Scalar-Types

naga/src/front/wgsl/parse/ast.rs
```diff
@@ -67,6 +68,17 @@ impl PartialEq for Dependency<'_> {
 
 impl Eq for Dependency<'_> {}
 
+/// A module-scope declaration.
```
teoxoy: Seems to have been copy-pasted.

naga/src/front/wgsl/parse/ast.rs
```diff
@@ -67,6 +68,17 @@ impl PartialEq for Dependency<'_> {
...
+#[derive(Debug)]
+pub struct GlobalDirective {
+    pub kind: GlobalDirectiveKind,
```
teoxoy: We should just make this struct an enum.

naga/src/front/wgsl/parse/ast.rs
```diff
@@ -164,6 +176,11 @@ pub struct ResourceBinding<'a> {
     pub binding: Handle<Expression<'a>>,
 }
 
+#[derive(Debug)]
+pub struct EnableDirective {
+    pub enable_extension_list: Vec<Extension>,
+}
```
teoxoy: Can we move this together with the other new declarations?

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2167,6 +2167,42 @@ impl Parser {
...
+        &mut self,
+        lexer: &mut Lexer<'a>,
+        out: &mut ast::TranslationUnit<'a>,
+    ) -> Result<(), Error<'a>> {
+        while let Token::Word("enable") = lexer.peek().0 {
+            let (_, enable_span) = lexer.next_ident_with_span()?;
+
+            let mut enable_extension_list = Vec::with_capacity(4);
+            let mut ready = true;
+            while !lexer.skip(Token::Separator(';')) {
```
teoxoy: The grammar requires at least one `enable_extension_name` to be present, this allows for 0.

naga/src/front/wgsl/parse/lexer.rs
```diff
@@ -376,14 +394,19 @@ impl<'a> Lexer<'a> {
...
+        let (scalar, span) = match self.next() {
+            (Token::Word(word), span) => conv::get_scalar_type(word)
+                .map(|scalar| (scalar, span))
+                .ok_or(Error::UnknownScalarType(span)),
             (_, span) => Err(Error::UnknownScalarType(span)),
         }?;
+
+        if matches!(scalar, Scalar::F16) && !self.extensions.contains(&Extension::F16) {
+            return Err(Error::ExtensionNotEnabled(span, Extension::F16));
+        }
```
teoxoy: `get_scalar_type` is called in 2 more places. I think we should move this check inside `get_scalar_type` an have it return a `Result` rather than `Option`.

naga/src/front/wgsl/parse/lexer.rs
```diff
@@ -204,6 +206,7 @@ pub(in crate::front::wgsl) struct Lexer<'a> {
     pub(in crate::front::wgsl) source: &'a str,
     // The byte offset of the end of the last non-trivia token.
     last_end_offset: usize,
+    extensions: SmallVec<[Extension; 4]>,
```
teoxoy: It seems appropriate to have an `Extensions` biflags type instead.

naga/src/front/wgsl/parse/ast.rs
```diff
@@ -1,10 +1,11 @@
 use crate::front::wgsl::parse::number::Number;
 use crate::front::wgsl::Scalar;
-use crate::{Arena, FastIndexSet, Handle, Span};
+use crate::{Arena, Extension, FastIndexSet, Handle, Span};
 use std::hash::Hash;
 
 #[derive(Debug, Default)]
 pub struct TranslationUnit<'a> {
+    pub directives: Arena<GlobalDirective>,
```
teoxoy: We don't use this field at all, we might in the future but I think it will end up looking differently. Extensions and language extensions can be bitflags, and global diagnostic modifiers will have their own shapes.

I think we should either remove this field or replace it with an extensions bitflags.

naga/src/back/msl/writer.rs
```diff
@@ -348,8 +350,12 @@ impl crate::Scalar {
         match self {
             Self {
                 kind: Sk::Float,
-                width: _,
+                width: 4 | 8,
```
teoxoy: We should `return Err(Error::CapabilityNotSupported(valid::Capabilities::FLOAT64))` if we see 8.

naga/src/back/msl/writer.rs
```diff
@@ -1290,6 +1296,21 @@ impl<W: Write> Writer<W> {
...
+                        write!(self.out, "{sign}INFINITY")?;
+                    } else if value.is_nan() {
+                        write!(self.out, "NAN")?;
+                    } else {
+                        let suffix = if value.fract() == f16::from_f32(0.0) {
+                            ".0"
+                        } else {
+                            ""
+                        };
+                        write!(self.out, "{value}{suffix}")?;
```
teoxoy: These need to be suffixed by an `h`.

naga/src/back/spv/writer.rs
teoxoy, line 1: We also need to request the `Float16` capability here:

https://github.com/gfx-rs/wgpu/blob/b9b7050bb93c005eeb0a8f7f3eade6e57971c53b/naga/src/back/spv/writer.rs#L822-L825

naga/src/proc/constant_evaluator.rs
FL33TW00D: Sorry not sure I follow, the macro is a little hairy.

FL33TW00D: Should be improved now, still not 100% certain on its correctness.

naga/src/back/spv/writer.rs
teoxoy: We also need to enable `StorageBuffer16BitAccess`, `UniformAndStorageBuffer16BitAccess` and `StorageInputOutput16`. It's probably fine to just enable those if we see the f16 type being used instead of having to trace the functionality they enable.

teoxoy: The Vulkan backend is also missing the `storageInputOutput16` feature. We should add it:

https://github.com/gfx-rs/wgpu/blob/d7a35ecda06742acb4cdd5a83c5b9737f375a70b/wgpu-hal/src/vulkan/adapter.rs#L658-L665

https://github.com/gfx-rs/wgpu/blob/d7a35ecda06742acb4cdd5a83c5b9737f375a70b/wgpu-hal/src/vulkan/adapter.rs#L383-L388

teoxoy (changes requested):

teoxoy:
Could you squash all the existing commits into 1? It will make it easier to look at the PR once the feedback is addressed.

FL33TW00D:
> Could you squash all the existing commits into 1? It will make it easier to look at the PR once the feedback is addressed.

@teoxoy 100% - thanks for the review let me address and squash!

teoxoy:
One thing I'm concerned is that the polyfills we have in some of the backends might not support `f16`. It would be worth adding more tests covering the built-in functions and operations that support `f16`. Execution tests would be great as well, they live in `tests/tests/shader`.

teoxoy:
It would be best for review to squash first so that the commits addressing various parts are by themselves.

naga/src/back/msl/writer.rs
FL33TW00D: Making this fallible will create a ton of work, can we keep the same behaviour as before (where it hits the unreachable if 8)?

naga/src/front/wgsl/parse/mod.rs
FL33TW00D: Nice catch! fixed

naga/src/back/msl/writer.rs
teoxoy: That's fine as well.

naga/src/proc/constant_evaluator.rs
FL33TW00D: Great!

wgpu-hal/src/dx12/adapter.rs
FL33TW00D: Ok now checking for `Native16BitShaderOpsSupported`.
I cannot get Rust to compile using `dx12` backend no matter what feature combination I supply on MacOS.
Still need to pass the flag to dxc.

naga/src/back/spv/writer.rs
FL33TW00D: "We also need to enable StorageBuffer16BitAccess, UniformAndStorageBuffer16BitAccess and StorageInputOutput16. It's probably fine to just enable those if we see the f16 type being used instead of having to trace the functionality they enable."

Can you expand more on this? Resolved the other 2.

naga/src/back/wgsl/writer.rs
FL33TW00D: This would require adding it to the `naga::Module`: https://github.com/gfx-rs/wgpu/blob/trunk/naga/src/lib.rs#L2254

Given it only applies to WGSL, didn't know how best to handle.

ErichDonGubler:
I have not reviewed this PR in-depth as @teoxoy has, but the broad strokes of the const. eval code look correct to me, FWIW. 🫡

naga/src/proc/constant_evaluator.rs
FL33TW00D: Fixed these but not 100% sure.

naga/src/lib.rs
FL33TW00D: Done!

naga/src/front/wgsl/parse/ast.rs
FL33TW00D: The other new decls are removed.

naga/src/front/wgsl/parse/lexer.rs
FL33TW00D: Done!

naga/src/front/wgsl/parse/lexer.rs
FL33TW00D: Done!

naga/src/back/spv/writer.rs
teoxoy: Each Vulkan feature has a corresponding SPIR-V capability in this case and they need to be enabled on both sides.

The comment was referring to SPIR-V's capabilities that we are not currently enabling. We can choose to enable each one of them whenever we detect that they are required or enable them all if we detect the presence of the `f16` type which should be a lot more straight forward.

FL33TW00D: Right! Thanks for the explanation - fixed!

naga/src/back/spv/writer.rs
```diff
@@ -823,6 +823,15 @@ impl Writer {
                 if bits == 64 {
                     self.capabilities_used.insert(spirv::Capability::Float64);
                 }
+                if bits == 16 {
+                    self.capabilities_used.insert(spirv::Capability::Float16);
+                    self.capabilities_used
+                        .insert(spirv::Capability::StorageBuffer16BitAccess);
+                    self.capabilities_used
+                        .insert(spirv::Capability::UniformAndStorageBuffer16BitAccess);
```
FL33TW00D, line 1162: @teoxoy not actually sure about this now.
https://github.com/gpuweb/gpuweb/issues/2512#issuecomment-1034413409
Seems some of these are optional. If you enable them all you lose most devices on Vulkan.

teoxoy: What ended up in the spec requires all of the Vulkan features. Some more context from the follow-up meetings:
https://github.com/gpuweb/gpuweb/issues/2512#issuecomment-1041207899
https://github.com/gpuweb/gpuweb/issues/2512#issuecomment-1057224605

naga/src/back/wgsl/writer.rs
teoxoy: I think we can do what we do on Vulkan where we check the type arena for the presence of the `f16` type.

teoxoy:
@FL33TW00D please request a re-review once ready.

FL33TW00D:
> @FL33TW00D please request a re-review once ready.

@teoxoy will do apologies this got deprioritised.

teoxoy:
No worries, I just want to make sure to get back to it.

tgross35:
Would it be feasible, either here or in a followup, to support the builtin `f16` as an alternative to `half`? It is still unstable so needs to be behind a feature gate, but it would be great to start working it into the ecosystem and help us find more bugs.

Bytemuck just needs its `nightly_float` feature enabled to support it.

teoxoy:
The buffers are all just a bunch of bytes, so you can use whichever type you want then put it in the buffer. The usage of `half` is internal to `naga`.

FL33TW00D:
Taking a look at this again @teoxoy on a rented EC2 machine to check Vulkan.
Seems like the NVDA Vulkan driver doesn't seem to allow F16 (which seems off):
```
HAS SHADER FLOAT16: Some((PhysicalDeviceShaderFloat16Int8Features { s_type: PHYSICAL_DEVICE_SHADER_FLOAT16_INT8_FEATURES, p_next: 0x7ffdc4f00a68, shader_float16: 1, shader_int8: 1, _marker: PhantomData<&()> }, PhysicalDevice16BitStorageFeatures { s_type: PHYSICAL_DEVICE_16BIT_STORAGE_FEATURES, p_next: 0x7ffdc4f00af8, storage_buffer16_bit_access: 1, uniform_and_storage_buffer16_bit_access: 1, storage_push_constant16: 1, storage_input_output16: 0, _marker: PhantomData<&()> }))
```

This is because the `storage_input_output16` is 0.

https://forums.developer.nvidia.com/t/nvidia-driver-should-enable-the-storageinputoutput16-feature/275835

Any ideas here? I'd expect a T4 to offer this?

EDIT:
wow turns out Turing didn't even support f16 and it isn't a bug.
```
Nvidia Turing-based hardware currently [reports support](https://vulkan.gpuinfo.org/displayreport.php?id=6907#extended) for shaderFloat16, shaderInt8, storageBuffer16BitAccess, uniformAndStorageBuffer16BitAccess, and storagePushConstant16, but not storageInputOutput16
```

teoxoy:
> It is possible to use FP16 in user-defined pipeline input/output variables. However some Vulkan devices, which support other usages in this extension like FP16 built-in function and storage buffer, don’t support this usage. For capabilities reasons, we may suggest not enabling using FP16 as pipeline input or output with this extension. However, this usage can also be emulated on Vulkan devices with no native support for this usage (`storageInputOutput16`).

_from https://github.com/gpuweb/gpuweb/issues/2512_

I think we have to do the casting ourselves. Dawn also seems to do this. We can do it in a different PR though.

What I find interesting is that the proposal talked about SM6.2 & Native16BitShaderOpsSupported enabling support for f16 IO, I guess the Nvidia compiler that ingests DXIL will do the casting. It probably does the casting regardless of frontend and that's why people report this working on Vulkan as well despite `storageInputOutput16` not being supported. While it's unlikely this will change we shouldn't rely on this behavior.

examples/src/shader_f16/README.md
```diff
@@ -0,0 +1,9 @@
+# shader-f16 
+
+Demonstrate the ability to perform compute in F16 using wgpu.
+
+## To Run
+
+```
+RUST_LOG=hello_compute cargo run --bin wgpu-examples shader_f16
```
ErichDonGubler: question: Why is `RUST_LOG` set to `hello_compute`, which is the name of a different example? Is this a typo?

ErichDonGubler:
I've taken the liberty of rebasing this PR, applying the following notable changes:

* Conflict resolution
	* `mem::size_of` imports
	* `windows` crate migration
	* Optional shader module entry point name in pipeline creation
* Changed our `half` dependency to be a non-`git` (viz., nominally Crates.io) dependency in `naga/Cargo.toml`, but modified by a `[patched]` entry PR in `Cargo.toml` to use @FL33TW00D's `arbitrary`. Once upstream merges `arbitrary` support, it should be easy to bump the `half` version number in `naga/Cargo.toml`, and discard the `[patch]` entry in `Cargo.toml`.

There are still numerous CI failures, but I believe that my rebasing onto `trunk` has not introduced new ones.

examples/src/shader_f16/README.md
FL33TW00D: Copy paste error!

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2474,6 +2522,7 @@ impl Parser {
 
         let mut lexer = Lexer::new(source);
         let mut tu = ast::TranslationUnit::default();
+        self.global_directive(&mut lexer, &mut tu)?;
```
ErichDonGubler: There's currently a conflict between this PR's addition of `enable` directive parsing and #6352's more general skeleton for directive parsing. I'm planning on reconciling this by rebasing on top of #6352, which should (1) bring in better diagnostics for unrecognized directives, and (2) set up a common structure for all directives and `enable` extensions going forward.

I've taken the liberty of breaking out the parsing and state tracking code into its own commit.

naga/src/front/wgsl/parse/ast.rs
ErichDonGubler: +1 to @teoxoy's thoughts here: I don't think we need the `{Global,Enable}Directive` type layers, because not all directives are going to be represented the same way (as I'm discovering in #6148). Let's just store a new `bitflags`-ish structure specific to `enable` extensions here.

ErichDonGubler (commented):
A few thoughts after reviewing this change myself.

naga/src/front/wgsl/parse/mod.rs
FL33TW00D: Nice sounds good

ErichDonGubler: Oh, shoot, I'd forgotten to complete my sentence and left (2) blank. Edited my message to complete it. 😅

ErichDonGubler: I've resolved this convo by rebasing on top of <https://github.com/gfx-rs/wgpu/pull/6424>, which handles the skeleton for general `enable` directive parsing. This narrows the scope of this PR to only the scope of moving `UnimplementedEnableExtension::F16` to the `ImplementedEnableExtension` `enum`. 😀

naga/src/front/wgsl/parse/ast.rs
ErichDonGubler: I've sorted this by making an `EnableExtensions` structure that stores `bool`s in named fields. Seems good enough for now; I don't think we even need the fanciness of a `bitflags`-ish structure, thinking about it!

naga/src/front/wgsl/parse/ast.rs
ErichDonGubler: I've rendered this conversation obsolete with my latest changes that alter the representation of a parsed `EnableExtension`/`EnableExtensions`.

examples/src/shader_f16/README.md
ErichDonGubler: Fixed to use `shader_f16` instead.

ErichDonGubler:
I've fixed up the outstanding conversations I have after rebasing onto #6424. @teoxoy, are there any more conversations you've opened that are resolved?

teoxoy:
No, they are up to date.

jimblandy (commented):
We should make sure that we are properly validating `As` casts of matrices, because that code currently assumes that all matrices have the same kind (#6441).

raphamorim:
Super awesome work folks, really appreciate it!! 👏 <3 
@FL33TW00D please lemme know if I can help in anything 🙏

wpmed92:
Is there an ETA for this to ship? We'd like to support `f16` in tinygrad, we love `wgpu`, but are considering switching to dawn if it's not expected to ship in the near future.

raphamorim:
> Is there an ETA for this to ship? We'd like to support f16 in tinygrad, we love wgpu, but are considering switching to dawn if it's not expected to ship in the near future.

same, i love wgpu and use on rio terminal. Please lemme know if I can help any way to ship this feature. Maybe can split some work

FL33TW00D:
@wpmed92 @raphamorim 

If either of you are using Vulkan or Windows backend, feel free to start testing support for F16 using this branch.

I use the Metal backend, implemented enough support so that all my own tests pass.

raphamorim:
Sounds good @FL33TW00D , i actually started testing this branch in Rio already.

Regarding https://github.com/starkat99/half-rs/pull/111 I have been thinking in publish a fork to crates io and maintain until your PR get merged

raphamorim:
Done: https://crates.io/crates/half-2
Hope unblock the CI issue for now, once your PR is merged then need to move back to half

raphamorim:
@FL33TW00D created a PR https://github.com/raphamorim/rio/pull/875 but having trouble with naga validation

```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 13.61s
     Running `target/debug/rio`
thread 'main' panicked at /Users/rapha/.cargo/git/checkouts/wgpu-48a004a2060e66f4/bfad205/wgpu/src/backend/wgpu_core.rs:3413:5:
wgpu error: Validation Error

Caused by:

Shader '' parsing error: expected global item ('struct', 'const', 'var', 'alias', ';', 'fn') or the end of the file, found 'enable'
  ┌─ wgsl:1:1
  │
1 │ enable f16;
  │ ^^^^^^ expected global item ('struct', 'const', 'var', 'alias', ';', 'fn') or the end of the file


      expected global item ('struct', 'const', 'var', 'alias', ';', 'fn') or the end of the file, found 'enable'
```

FL33TW00D:
> @FL33TW00D created a PR [raphamorim/rio#875](https://github.com/raphamorim/rio/pull/875) but having trouble with naga validation
> 
> ```
>     Finished `dev` profile [unoptimized + debuginfo] target(s) in 13.61s
>      Running `target/debug/rio`
> thread 'main' panicked at /Users/rapha/.cargo/git/checkouts/wgpu-48a004a2060e66f4/bfad205/wgpu/src/backend/wgpu_core.rs:3413:5:
> wgpu error: Validation Error
> 
> Caused by:
>   In Device::create_shader_module
> 
> Shader '' parsing error: expected global item ('struct', 'const', 'var', 'alias', ';', 'fn') or the end of the file, found 'enable'
>   ┌─ wgsl:1:1
>   │
> 1 │ enable f16;
>   │ ^^^^^^ expected global item ('struct', 'const', 'var', 'alias', ';', 'fn') or the end of the file
> 
> 
>       expected global item ('struct', 'const', 'var', 'alias', ';', 'fn') or the end of the file, found 'enable'
> ```

Just checked out 808db2481 and ran:
`RUST_LOG=shader_f16 cargo run --bin wgpu-examples shader-f16`
It works correctly.

You forgot to specify the branch in the cargo git dep!

raphamorim:
> You forgot to specify the branch in the cargo git dep!

You are indeed correct, thank you!

ErichDonGubler:
Just resolved the latest conflicts; CC @FL33TW00D and @raphamorim.

cwfitzgerald:
Getting this over the line is on my list, will probably be next week.

ErichDonGubler:
I've resolved conflicts with the latest `examples/` changes (I think, spot check me @cwfitzgerald), and added a `fixup!` that tentatively consumes @raphamorim's `half-2` crate.

cwfitzgerald:
I have split the hex parsing task into a different work item to follow this. See #7046.

ErichDonGubler:
Just realized that we didn't have the `vec{2,3,4}h` type aliases getting added to WGSL here. Pushed up 493e740f as a `fixup!` to remediate that.

cwfitzgerald:
Restored ownership of the commit, sorry about that :)

raphamorim:
man it's happening, let's go! ❤️ 

@cwfitzgerald @ErichDonGubler please lemme know if i can help somehow

examples/features/Cargo.toml
```diff
@@ -36,6 +36,10 @@ encase = { workspace = true, features = ["glam"] }
 flume.workspace = true
 getrandom.workspace = true
 glam = { workspace = true, features = ["bytemuck"] }
+half = { package = "half-2", version = "2.4.1", features = [
+    "arbitrary",
+    "bytemuck",
+] }
```
cwfitzgerald: Needs to be removed.

naga/Cargo.toml
```diff
@@ -98,6 +105,10 @@ petgraph = { version = "0.7", optional = true }
 pp-rs = { version = "0.2.1", optional = true }
 hexf-parse = { version = "0.2.1", optional = true }
 unicode-ident = { version = "1.0", optional = true }
+# TODO: Use `half` directly again after <https://github.com/starkat99/half-rs/pull/111> is released
+# upstream.
+half = { workspace = true, features = ["arbitrary", "num-traits"] }
+num-traits = "0.2"
```
cwfitzgerald: These being at the end isn't good.

cwfitzgerald: Validate that this is needed.

naga/src/back/wgsl/writer.rs
```diff
@@ -217,6 +222,41 @@ impl<W: Write> Writer<W> {
...
+                        writeln!(self.out, "enable f16;")?;
+                        written_declarations.insert(WrittenDeclarations::F16);
+                    }
+                }
+                _ => {}
+            }
+        }
+
+        if !written_declarations.is_empty() {
+            // Break for readability
```
cwfitzgerald:
```suggestion
            // New line for readability
```

naga/src/valid/type.rs
```diff
@@ -248,22 +259,37 @@ impl super::Validator {
         }
     }
 
-    pub(super) const fn check_width(&self, scalar: crate::Scalar) -> Result<(), WidthError> {
+    pub(super) const fn check_width(
+        &self,
+        scalar: crate::Scalar,
+    ) -> Result<PushConstantCompatibility, WidthError> {
+        let mut push_constant_compatibility = Ok(());
```
cwfitzgerald, line 268: This is the one change I wasn't sure about, but this solution does end up working fine.

naga/tests/example_wgsl.rs
```diff
@@ -1,55 +1,51 @@
 #![cfg(feature = "wgsl-in")]
 
 use naga::{front::wgsl, valid::Validator};
-use std::{fs, path::PathBuf};
+use std::{ffi::OsStr, fs, path::Path};
 
 /// Runs through all example shaders and ensures they are valid wgsl.
 #[test]
 pub fn parse_example_wgsl() {
-    let read_dir = match PathBuf::from(env!("CARGO_MANIFEST_DIR"))
```
cwfitzgerald: This needs to be factored out.

naga/tests/in/f16.wgsl
```diff
@@ -0,0 +1,128 @@
+enable f16;
+enable f16; //redundant directives are OK
```
cwfitzgerald: This should be a wgsl_errors.rs test

naga/tests/out/spv/atomicCompareExchange-int64.spvasm
```diff
@@ -1,17 +1,17 @@
 ; SPIR-V
 ; Version: 1.0
 ; Generator: rspirv
-; Bound: 125
+; Bound: 123
```
cwfitzgerald: Why did this change?

naga/Cargo.toml
cwfitzgerald: It is.

naga/tests/out/spv/atomicCompareExchange-int64.spvasm
cwfitzgerald: Oh I remember why - the old `As` code ended up allocating an `id` that it never used. The new code does not have this property, so there's some small renumberings in the other spvasm files.

naga/tests/example_wgsl.rs
cwfitzgerald: Split out as #7191

tests/tests/shader/struct_layout.rs
```diff
@@ -356,106 +461,232 @@ fn create_64bit_struct_layout_tests() -> Vec<ShaderTest> {
...
+                //
+                16, 17, 18, 19, // m4[0]
+                20, 21, 22, 23, // m4[1]
+                24, 25, 26, 27, // m4[2]
+                28, 29, 30, 31, // m4[3]
+            ],
+        ));
+    }
+
+    // // Nested struct and array test.
```
teoxoy, line 620: Looks copied from the 64bit test.

teoxoy: Looks copied from the 64bit test.

naga/src/back/spv/block.rs
```diff
@@ -1784,6 +1633,237 @@ impl BlockContext<'_> {
         Ok(())
     }
 
+    /// Helper which focuses on generating the `As` expressions and the various conversions
+    /// that need to happen because of that.
+    fn write_as_expression(
```
teoxoy, line 1779: I can't tell what the code changes are, could you split the refactor in its own commit?

naga/src/front/spv/mod.rs
```diff
@@ -80,6 +81,7 @@ pub const SUPPORTED_EXTENSIONS: &[&str] = &[
     "SPV_KHR_vulkan_memory_model",
     "SPV_KHR_multiview",
     "SPV_EXT_shader_atomic_float_add",
+    "SPV_KHR_16bit_storage",
```
teoxoy, line 86: We should also list the capabilities above.

naga/src/lib.rs
```diff
@@ -895,6 +896,7 @@ pub enum Literal {
     F64(f64),
     /// May not be NaN or infinity.
     F32(f32),
+    F16(f16),
```
teoxoy:
```suggestion
    /// May not be NaN or infinity.
    F16(f16),
```

naga/src/front/wgsl/error.rs
```diff
@@ -1008,19 +1006,18 @@ impl<'a> Error<'a> {
...
-                message: format!("`{}` enable-extension is not enabled", kind.to_ident()),
+                message: format!("the `{}` language extension is not enabled", kind.to_ident()),
                 labels: vec![(
                     span,
                     format!(
                         concat!(
-                            "the `{}` enable-extension is needed for this functionality, ",
-                            "but it is not currently enabled"
+                            "the `{0}` language extension is needed for this functionality, ",
+                            "but it is not currently enabled."
```
teoxoy, line 1033: "enable-extension" was correct.

naga/src/front/wgsl/parse/number.rs
```diff
@@ -367,7 +380,8 @@ fn parse_hex_float(input: &str, kind: Option<FloatKind>) -> Result<Number, Numbe
             // can only be ParseHexfErrorKind::Inexact but we can't check since it's private
             _ => Err(NumberError::NotRepresentable),
         },
-        Some(FloatKind::F16) => Err(NumberError::UnimplementedF16),
+        // TODO: f16 is not supported by hexf_parse
+        Some(FloatKind::F16) => Err(NumberError::NotRepresentable),
```
teoxoy, line 386: The error is misleading, we should add a new error variant.

Cargo.toml
```diff
@@ -100,6 +100,8 @@ flume = "0.11"
 futures-lite = "2"
 getrandom = "0.2"
 glam = "0.29"
+# TODO: Use `half` directly again after <https://github.com/starkat99/half-rs/pull/111> is released upstream.
+half = { package = "half-2", version = "2.4.1" }
```
teoxoy: The crate [`half` v1.8.2](https://searchfox.org/mozilla-central/rev/18bfd136df7f2fe618d8105549cccebb75ce92f5/Cargo.lock#2767) is already in Firefox's tree. I'd like to avoid having to bump its version (once that PR lands) since other crates would then need their versions bumped. Looking at the changelong of [`half` v2+](https://github.com/starkat99/half-rs/compare/v1.8.2...v2.4.1), I don't see anything critical that we would need. Another thing to consider is that rust will get an [`f16` type](https://github.com/rust-lang/rust/issues/116909) at some point in the future.

Could we instead use `half` v1.8.2 with a wrapper that implements `Arbitrary`?

teoxoy (dismissed):
This looks good overall!

Besides the other comments:

- The docs for the feature flag need to be updated.
- The metal backend seems to not advertise support for the feature.

naga/src/valid/type.rs
teoxoy: I think this info should be encoded as a new flag in `TypeFlags`.

cwfitzgerald:
Metal is enabled by https://github.com/FL33TW00D/wgpu/blob/feature/enable/wgpu-hal/src/metal/adapter.rs#L914

naga/src/front/wgsl/error.rs
cwfitzgerald: I think my main problem here is that the concept of an "enable-extension" exists nowhere else, so isn't likely to mean anything to a user. Maybe something like:

```
An `enable {0}` statement is required to use this functionality, but is not present in the shader.
```

How does that sound

Cargo.toml
cwfitzgerald: I'll see if that's possible

naga/src/back/spv/block.rs
cwfitzgerald: I'll try - might be kinda difficult

naga/src/front/wgsl/error.rs
teoxoy: enable-extension is what the spec uses though https://www.w3.org/TR/WGSL/#enable-extension. The error message has a note: "You can enable this extension by adding `enable {};` at the top of the shader."; which feels quite pragmatic to me.

ErichDonGubler: Not only is "enable-extension" the right term (from the standard) for `enable` directives ([link](https://www.w3.org/TR/WGSL/#enable-extensions-sec)), "language extension" is the _wrong_ term for this, because that corresponds to the `requires` directive instead ([link](https://www.w3.org/TR/WGSL/#language-extension)).

ErichDonGubler: RE: @cwfitzgerald's suggestion: "statement" (an instruction _inside a function_) is meaningfully distinct from a "directive" (an instruction before any global items in a module) in terms of WGSL's grammar. Enable-extensions are a directive, so please use the noun "directive" instead.

naga/src/front/wgsl/error.rs
```diff
@@ -1008,19 +1006,18 @@ impl<'a> Error<'a> {
                 )],
             },
             Error::EnableExtensionNotEnabled { kind, span } => ParseError {
-                message: format!("`{}` enable-extension is not enabled", kind.to_ident()),
+                message: format!("the `{}` language extension is not enabled", kind.to_ident()),
```
ErichDonGubler: nitpick: I don't think a definite article is valuable here. In my mind, headline error text should favor being brevity, while labels and notes should be the place for more natural-language explanations.

suggestion: Let's revert this hunk.

naga/src/front/wgsl/error.rs
ErichDonGubler: style: Please revert `{0}` back to `{}`; this change doesn't seem valuable, and it's inconsistent with surrounding code.

ErichDonGubler: thought: This change seems like a good candidate for being broken out into another PR. It's not necessary for overall functionality to land, and is subject to a lot of soft thinking which is valuable, but can take time to reach consensus.

ErichDonGubler (changes requested):
Adding a blocker for diagnostic adjustments specifically; I expect these to be easily resolved with a few (sync?) rounds of feedback between you, @teoxoy, and myself.

naga/src/front/wgsl/error.rs
```diff
@@ -1032,7 +1029,12 @@ impl<'a> Error<'a> {
                         kind.tracking_issue_num()
                     )]
                 } else {
-                    vec![]
+                    vec![
+                        format!(
+                            "You can enable this extension by adding `enable {};` at the top of the shader.",
```
ErichDonGubler: issue: "at the top of the shader" is a good informal abbreviation for where directives can be parsed, but for a new user that may try to put this, say, under their small `const MY_THING = 1;` on line 1 of their WGSL file, it's important to be precise in this hint to avoid frustration.

```suggestion
                            "You can enable this by adding `enable {};` before global items.",
```

naga/src/front/wgsl/parse/number.rs
```diff
@@ -16,12 +18,23 @@ pub enum Number {
...
+    /// Concrete f16
+    F16(f16),
     /// Concrete f32
     F32(f32),
     /// Concrete f64
     F64(f64),
 }
 
+impl Number {
+    pub(super) const fn required_enable_extension(&self) -> Option<ImplementedEnableExtension> {
```
ErichDonGubler: nitpick: I think present indicative is usually a better tense for query methods like this than past indicative, i.e., "requires", rather than "required".

suggestion: Let's use `requires_enable_extension`.

Cargo.toml
ErichDonGubler: I suspect both of you already know this, but just for reference: Mozilla _can_ use a fork of `half`, without much issue (because we can base changes we need on what we already use, making auditing trivial). I think it's WGPU's own requirements for posting on Crates.io that make this hard?

ErichDonGubler:
@raphamorim: Can you please post the Git history for `half-2`, presumably in a fork under your user? We noticed that the commit hash `367f64f7a23e8763d1662e7da6731d74422133b3` in [`half-2`'s `.cargo_vcs_info.json`](https://docs.rs/crate/half-2/latest/source/.cargo_vcs_info.json) is not accessible from the `starkat99` or `FL33TWOOD` repos.

raphamorim:
@ErichDonGubler I apologize, I haven't updated the fork on GH. Just ran a git push 🙏 

Regarding that hash I confess I did the fork and published as a test to if it was something worth to maintain and use for wgpu, but in case we plan to use half fork then I just updated the tree to only contain @FL33TW00D changes (along with Cargo.toml)

Also please bump to 2.4.2 https://crates.io/crates/half-2

naga/src/front/spv/mod.rs
cwfitzgerald: We already do list Float16, weirdly.

naga/src/front/wgsl/error.rs
cwfitzgerald: I've implemented a possible revision to this.

Cargo.toml
cwfitzgerald: We've decided in the meeting to go ahead with the newly updated `half` crate, hoping the audit load isn't too heavy.

cwfitzgerald:
All comments should be addressed.

FrankenApps:
Btw.: [`half`](https://crates.io/crates/half) was recently updated with the PR on which this depends.

cwfitzgerald:
Yup! I also updated it to use `half` 2.5 again

ErichDonGubler:
@gfx-rs/wgpu: I will take on Mozilla's audit workload for `half`.

ErichDonGubler (approved):
image

ErichDonGubler:
Resolved conflicts with the extraction of the `ir` module in `naga`, and added [efabbf6](https://github.com/gfx-rs/wgpu/pull/5701/commits/efabbf61ebfa52b4187aa0d5bfc11e4e45347e01) for a bit of clarity with the `num-traits` upgrade.

raphamorim:
> image

image

LET'S GOOOO, great job folks 🙏

cwfitzgerald:
Wooooooooooooo!

Issue #4489, opened 2023-04-07
[spv-in] Atomics support


schell:
Naga's SPIR-V frontend does not currently handle atomic operations. Using an atomic operation (in this case `spirv_std::arch::atomic_i_increment` in a `rust-gpu` shader) leads to a parse error: 
```
UnsupportedInstruction(Function, AtomicIIncrement)
```
I'm treating this as the tracking issue for adding atomics to the spv frontend.

## Milestone 1 

Completion of milestone 1 will be a proof-of-strategy, after which that strategy can be re-applied to complete atomics support for all operations. First atomic operations support will be documented in the [renderling devlog](https://renderling.xyz/devlog/index.html) as a method / PoC to help in the completion of other operations.

- https://github.com/gfx-rs/wgpu/pull/5702
- https://github.com/gfx-rs/wgpu/pull/5775

## Milestone 2 

Rinse and repeat, adding support for more atomic operations, until all have been exhausted.

- https://github.com/gfx-rs/wgpu/pull/5824

## Milestone 3 & Follow up work 

Cleaning up tests, bugfixes that might pop up, and generally making this feature more maintainable - tying up any loose ends.

- https://github.com/gfx-rs/wgpu/issues/6292
  - https://github.com/gfx-rs/wgpu/pull/6692
- https://github.com/gfx-rs/wgpu/issues/6296
  - https://github.com/gfx-rs/wgpu/pull/6590  
- https://github.com/gfx-rs/wgpu/issues/6683
  - https://github.com/gfx-rs/wgpu/pull/6693

schell:
I've submitted a proposal to NLNet to fund this issue. If they accept my proposal I'll likely pick this up after the holidays.

tombh:
So `AtomicIIncrement` is working in v22.0? I'm just wondering if Bevy's upcoming v0.15 release (the current release is still on naga v0.20.0) will support some SPIRV atomics. Though I understand we'll have to wait for #5824 to get more functions, like say `AtomicIDecrement`. Also, I wonder if that will be a patch bump so that we can use Cargo's `[patch.crates-io]` to get more atomics?

cwfitzgerald:
This is just parsing, not fully translating them, as far as im aware. @schell is that a correct assessment?

schell:
@tombh `AtomicIIncrement` should indeed be working in 0.22! Please let me know if `naga` does *not* accept your shader that uses this op!

[...]

PR #5702, opened 2024-05-14
spv-in parsing Op::AtomicIIncrement

schell:
**Connections**
Partially adresses #4489. 

This is mostly for @jimblandy and @cwfitzgerald to ensure that I'm not totally off course on my quest to add atomics. 
I'm trying to make my PRs as small as possible to reduce the overhead of reviewing these as they come in :)

**Description**
Begins work adding support for atomics in `naga`'s SPIR-V frontend.

* [x] adds a temporary feature `spv-in-atomics` that enables the new feature
* [x] adds support for parsing `Op::AtomicIIncrement`
  - [x] parse instruction
  - [x] insert result expression
  - [x] insert atomic statement
* [x] adds a lookup map to `Frontend` to track atomics use for a later pass

This _does not_ generate a valid `Module` as it predictably fails to type check ;) 
 
**Testing**
Added a test to parse and validate (with empty flags) a `.spv` file with an `AtomicIIncrement` instruction.

**Checklist**

- [x] Run `cargo fmt`.


cwfitzgerald (commented):
Great to see this! 

Can't comment on the code itself (though it generally looks good), but I don't think we need a special feature for it, this can just be part of the main spirv backend.

naga/src/front/spv/mod.rs
```diff
@@ -3944,7 +3961,68 @@ impl<I: Iterator<Item = u32>> Frontend<I> {
...
+                        scope_id,
+                        memory_semantics_id,
+                    };
+                    self.lookup_atomic.insert(pointer_id, atomic);
+
+                    log::trace!("\t\t\tlooking up expr {:?}", pointer_id);
+                    let (p_lexp_handle, p_lexp_ty_id) = {
+                        let lexp = self.lookup_expression.lookup(pointer_id)?;
+                        (lexp.handle, lexp.type_id)
+                    };
```
jimblandy: I think you're going to need the `get_expr_handle!` macro here. See the comments on `Frontend::get_expr_handle` for details on the cases that takes care of for you.

naga/src/front/spv/mod.rs
```diff
@@ -3944,7 +3961,68 @@ impl<I: Iterator<Item = u32>> Frontend<I> {
...
+                    log::trace!("\t\t\tlooking up type {:?}", pointer_id);
+                    let p_ty = self.lookup_type.lookup(p_lexp_ty_id)?;
+
+                    // Create an expression for our result
+                    let r_lexp_handle = {
+                        let expr = crate::Expression::AtomicResult {
+                            ty: p_ty.handle,
+                            comparison: false,
+                        };
```
jimblandy: I'm not sure I'm following this exactly, but I *believe* that `pointer_id` is going to be an `OpTypePointer`, so `p_ty.handle` will refer to a Naga `TypeInner::Pointer`. However, an `AtomicResult` expression's type should always be `i32` or `u32` for this instruction - that is, the pointer's base type. I think you can get this by looking up `p_ty.base_id`.

naga/src/front/spv/mod.rs
```diff
@@ -3944,7 +3961,68 @@ impl<I: Iterator<Item = u32>> Frontend<I> {
...
+                        );
+                        handle
+                    };
+
+                    // Create a literal "1" since WGSL lacks an increment operation
+                    // Create a statement for the op itself
+                    let one_lexp_handle = {
+                        let expr = crate::Expression::Literal(crate::Literal::I32(1));
+                        ctx.expressions.append(expr, self.span_from_with_op(start))
+                    };
```
jimblandy: Isn't it possible for the base type to be either `i32` or `u32`? Would it be possible to use `make_index_literal` here? That might help match the `crate::Literal` variant with the atomic's operand type.

naga/src/front/spv/mod.rs
```diff
@@ -5594,4 +5672,38 @@ mod test {
         ];
         let _ = super::parse_u8_slice(&bin, &Default::default()).unwrap();
     }
+
+    #[test]
+    fn atomic_i_inc() {
```
jimblandy, line 5689: This should become a snapshot test. I think it would be okay to add a `NO_VALIDATION` flag to `Targets` in `naga/tests/snapshots.rs` that directs `check_targets` to return early without trying to validate the module, and then have this test case handled by `convert_spv_all`, passing `Targets::IR | Targets::NO_VALIDATION`. Then you could at least see the IR generated for your SPIR-V.

jimblandy (changes requested):
This looks fine so far. Before we merge this I'd like the test to be turned into a snapshot test.

I'm kind of guessing at a bunch of stuff here so if I say something that doesn't sound right, definitely push back until it makes sense.

naga/src/front/spv/mod.rs
schell: @jimblandy - I'll add this as a snapshot test. Is it ok if I also keep it inline here, as this is an effective "staging ground" for my development workflow. From here I can easily play around with the frontend and the module and I can run `cargo watch -x 'test -p naga -- --nocapture atomic_i_inc'`. Otherwise I'll just keep it in a separate branch.

schell: Oh, also - it appears there is no `Targets::NO_VALIDATION` variant. I'm not sure how to keep the snapshot test from failing.

naga/src/front/spv/mod.rs
schell: Yup! You're right. It has to match the type of `pointer_id`.

naga/src/front/spv/mod.rs
schell: Ah, wait - I actually think that after adding this as a snapshot test it's correctly catching the case you mentioned above where the literal used as an increment amount is the wrong type. 👍

schell: Ok, after using `make_index_literal` to create a literal with a type that matches the atomic pointer's base type that error has cleared and the snapshot test is erring on the invalid atomic pointer, as expected.

schell: Ok, I've added `Targets::NO_VALIDATION` which is used to determine the `ValidationFlags` used in `naga::tests::snapshots::check_targets`, and now the snapshot passes with the `Targets::NO_VALIDATION` present.

cwfitzgerald (approved):
LGTM

naga/src/front/spv/mod.rs
```diff
@@ -3943,7 +3960,84 @@ impl<I: Iterator<Item = u32>> Frontend<I> {
...
+                    };
+                    log::trace!("\t\t\tlooking up type {pointer_id:?}");
+                    let p_ty = self.lookup_type.lookup(p_lexp_ty_id)?;
+                    let p_base_ty = if let Some(id) = p_ty.base_id {
+                        log::trace!("\t\t\tlooking up base type {id:?} of {p_ty:?}");
+                        self.lookup_type.lookup(id)?
+                    } else {
+                        log::warn!("\t\t\ttype {p_ty:?} has no base id");
+                        p_ty
+                    };
```
jimblandy: I think it should be fine to just return an error if `p_ty.base_id` is not `Some`, because the only way that `Frontend::parse_type_pointer` is ever going to put a `naga::TypeInner::Pointer` type in `self.lookup_type` in the first place is if it's got a base type.

(I assume the overall effort here is not aiming to add support for atomic operations on images - is that correct?)

naga/src/front/spv/mod.rs
```diff
@@ -3943,7 +3960,84 @@ impl<I: Iterator<Item = u32>> Frontend<I> {
...
+                                handle,
+                                type_id: result_type_id,
+                                block_id,
+                            },
+                        );
+                        handle
+                    };
+
+                    // Create a literal "1" since WGSL lacks an increment operation
+                    // Create a statement for the op itself
```
jimblandy: This "Create a statement" comment is now out of place.

jimblandy (approved):
Two small issues, but otherwise looks good for now. Once you've addressed these, give me a ping and I'll merge it.

naga/src/front/spv/mod.rs
schell: > I assume the overall effort here is not aiming to add support for atomic operations on images - is that correct?

Your assumption is correct in that I hadn't even considered it. I think that makes it out of scope for my project, but if I need it down the line you bet I'll try to cut a PR for it later :)

jimblandy:
This is fine for now, but ideally commit comments should be a clearer explanation of the changes, of the sort one might include in CHANGELOG.txt or something, but with more detail. For example, see: 8d73e5a9c

[edit: well, maybe that message talks a bit too much about a boring change. But the hope is that at least occasionally the commit message could save someone the time of actually reading the diff.]

jimblandy:
Thanks very much! Looking forward to the next episode!

PR #5708, opened 2024-05-15
Add infrastructure for counting and reporting internal resources

### LINKED ISSUE #5546: Add an API to expose core and hal internal debugging information
I have a work-in-progress [debug-overlay](https://github.com/nical/rust_debug/tree/master/font/overlay) for wgpu which I would like to use to investigate wgpu issues such as resource leaks. The overlay can display simple ascii text and graphs in a render pass.

While I'm polishing the overlay, one of the missing pieces is to surface wgpu's vitals in a way that is useful for debugging.

To feed the data to the overlay I would need some way to pull it out of wgpu-core at least. I propose that we keep track of useful info in a per-device struct that could be pulled The API could look like this:

```rust
pub struct HalCounters {
    // whatever information we want to get out of hal.
}

/// A structure that contains whatever debugging information is useful for wgpu developers and/or users.
pub struct InternalCounters {
    pub hal: HalCounters,
    pub allocated_textures: u32,
    pub allocated_buffers: u32,
    // ...
    // Option::None represents the case where a counter is "not set" for example if the information is about something that may not have happened.
    pub render_pass_encoding_time_ns: Option<u64>
    // ...
    // Also things that are more complicated than scalars, for example a data structure that describes the contents of the GPU memory allocator.
    // etc.
}

// Extracted this way:
let mut counters = wgpu_core::InternalCounters::default();
device.get_and_reset_internal_counters(&mut counters);
// or 
let counters = device.get_internal_counters(&mut counters);
```

Or

```rust
pub struct CountderId(usize);

pub struct InternalCounters {
    // The bulk of the data
    counters: Vec<Option<f32>>,
    // other more complicated data.
    // ...
}

impl InternalCounters {
    pub fn reset(&mut self) { ... }
    pub fn set(id: CounterId, value: Into<f32>) { ... }
    pub fn scalars(&self) -> impl Iterator<Item=Option<(CounterId, f32)>> { ... }
}

pub const ALLOCATED_TEXTURES: CounterId = CounterId(0);
pub const ALLOCATED_BUFFERS: CounterId = CounterId(1);
// etc.

// Extracted this way:
let mut counters = wgpu_core::InternalCounters::default();
device.get_and_reset_internal_counters(&mut counters);
// or 
let counters = device.get_internal_counters(&mut counters);
```

The second version seems more convoluted at first but doing it this way makes it possible to add counters in wgpu and have them integrate in the overlay automatically without manual intervention in the code that pulls the counters and feeds the overlay. That's how WebRender handles a similar functionality in Firefox.

nical:
... for debugging purposes

Fixes #5546

The PR is a first step, let's discuss the approach before I add more on top of it.

This adds an `InternalCounter` type which is internally a relaxed atomic isize when the `counters` feature is enabled and compiles to nothing otherwise.

These internal counters can be modified from anywhere with an `&self` reference. The idea is that a user of wgpu-core (and wgpu?) could query these internal counters and log them or display them using some custom [overlay](https://github.com/nical/rust_debug/tree/master/font/overlay) mechanism.

My primary motivation for this is to help with investigating resource lifetime/leak issues in wgpu, but I think that it would also be useful for users of wgpu.

The third commit might look a bit intimidating: I'm counting all textures, buffers, views, etc. in all hal backends. That's a fair amount of repetition, but I ended taking this route because wgpu-core internally creates some resources that aren't exposed to the API (like staging buffers), that would have to be tracked manually and generally I have more confidence in these counters staying up to date if they are in hal.
That said I expect to add more counters in both core and hal for whatever information will aid debugging.

Next steps:
- [x] Properly expose these in wgpu-core and wgpu.
- Add an API to generate memory reports that would be provided by the underlying gpu memory allocator (when there is one). I opened a PR in gpu-allocator https://github.com/Traverse-Research/gpu-allocator/pull/226 to expose the relevant information, I have to figure out if/why I have to do the same in the `gpu-alloc` crate that the vulkan backend uses.

### DIFF
--- a/wgpu-core/Cargo.toml
+++ b/wgpu-core/Cargo.toml
@@ -32,6 +32,9 @@ ignored = ["cfg_aliases"]
 [lib]
 
 [features]
+## Internally count resources and events for debugging purposes. If the counters
+## feature is disabled, the counting infrastructure is removed from the build and
+## the exposed counters always return 0.
 counters = ["wgt/counters"]
 
 ## Log all API entry points at info instead of trace level.
--- a/wgpu-core/Cargo.toml
+++ b/wgpu-core/Cargo.toml
@@ -32,6 +32,8 @@ ignored = ["cfg_aliases"]
 [lib]
 
 [features]
+counters = ["wgt/counters"]
+
 ## Log all API entry points at info instead of trace level.
 api_log_info = []
 
--- a/wgpu-hal/src/metal/device.rs
+++ b/wgpu-hal/src/metal/device.rs
@@ -8,7 +8,6 @@ use std::{
 
 use super::conv;
 use crate::auxil::map_naga_stage;
-use crate::InternalCounters;
 
 type DeviceResult<T> = Result<T, crate::DeviceError>;
 
@@ -1410,7 +1409,7 @@ impl crate::Device for super::Device {
         unimplemented!()
     }
 
-    fn get_internal_counters(&self) -> &InternalCounters {
-        &self.counters
+    fn get_internal_counters(&self) -> wgt::HalCounters {
+        self.counters.clone()
     }
 }
--- a/wgpu-hal/src/metal/device.rs
+++ b/wgpu-hal/src/metal/device.rs
@@ -347,13 +347,16 @@ impl crate::Device for super::Device {
             if let Some(label) = desc.label {
                 raw.set_label(label);
             }
+            self.counters.buffers.add(1);
             Ok(super::Buffer {
                 raw,
                 size: desc.size,
             })
         })
     }
-    unsafe fn destroy_buffer(&self, _buffer: super::Buffer) {}
+    unsafe fn destroy_buffer(&self, _buffer: super::Buffer) {
+        self.counters.buffers.sub(1);
+    }
 
     unsafe fn map_buffer(
         &self,
@@ -420,6 +423,8 @@ impl crate::Device for super::Device {
                 raw.set_label(label);
             }
 
+            self.counters.textures.add(1);
+
             Ok(super::Texture {
                 raw,
                 format: desc.format,
@@ -431,7 +436,9 @@ impl crate::Device for super::Device {
         })
     }
 
-    unsafe fn destroy_texture(&self, _texture: super::Texture) {}
+    unsafe fn destroy_texture(&self, _texture: super::Texture) {
+        self.counters.textures.sub(1);
+    }
 
     unsafe fn create_texture_view(
         &self,
@@ -491,9 +498,14 @@ impl crate::Device for super::Device {
             })
         };
 
+        self.counters.texture_views.add(1);
+
         Ok(super::TextureView { raw, aspects })
     }
-    unsafe fn destroy_texture_view(&self, _view: super::TextureView) {}
+
+    unsafe fn destroy_texture_view(&self, _view: super::TextureView) {
+        self.counters.texture_views.sub(1);
+    }
 
     unsafe fn create_sampler(
         &self,
@@ -550,15 +562,20 @@ impl crate::Device for super::Device {
             }
             let raw = self.shared.device.lock().new_sampler(&descriptor);
 
+            self.counters.samplers.add(1);
+
             Ok(super::Sampler { raw })
         })
     }
-    unsafe fn destroy_sampler(&self, _sampler: super::Sampler) {}
+    unsafe fn destroy_sampler(&self, _sampler: super::Sampler) {
+        self.counters.samplers.sub(1);
+    }
 
     unsafe fn create_command_encoder(
         &self,
         desc: &crate::CommandEncoderDescriptor<super::Api>,
     ) -> Result<super::CommandEncoder, crate::DeviceError> {
+        self.counters.command_encoders.add(1);
         Ok(super::CommandEncoder {
             shared: Arc::clone(&self.shared),
             raw_queue: Arc::clone(&desc.queue.raw),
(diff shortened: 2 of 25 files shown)


wgpu-core/Cargo.toml
```diff
@@ -33,6 +33,7 @@ ignored = ["cfg_aliases"]
 
 [features]
 default = ["link"]
+counters = ["wgt/counters"]
```
Wumpf, line 38: Features should be documented  - this is the right place to note down the impact of this and what it gives to a user

wgpu-types/src/lib.rs
```diff
@@ -7243,3 +7245,154 @@ pub enum DeviceLostReason {
     /// will call the callback immediately, with this reason.
     DeviceInvalid = 4,
 }
+
+/// An internal counter for debugging purposes
```
Wumpf: I know wgpu-types at this point has a long tradition of only being a single monstrous lib.rs file, but this seems like a nice reason to add  `counters.rs` to make things more readable

Wumpf (changes requested):
bunch of nits, but looking good to me overall. I also see the need for doing the counting at the hal level, but I'm concerned on how to reconciliate this stuff with the existing [`HubReport`](https://docs.rs/wgpu/latest/wgpu/core/hub/struct.HubReport.html).
What's happening in `InternalCounters` and `HubReport` is strongly overlapping from a users point of view even though the source of the information is quite different. Either we introduce clear distinction for what is what or (better imho) we merge these new counters into the existing hub report. Another interesting course of action would be to deprecate the hub report, but as long as we _have_ hubs they are quite useful.

wgpu-types/src/lib.rs
ErichDonGubler: +1 to not having a monstrous `lib.rs` file. This has made some things simple, but also, editor operations on this file feel heavy.

wgpu-hal/src/metal/device.rs
```diff
@@ -305,13 +306,16 @@ impl crate::Device for super::Device {
...
             }
+            self.counters.buffers.add(1);
             Ok(super::Buffer {
                 raw,
                 size: desc.size,
             })
         })
     }
-    unsafe fn destroy_buffer(&self, _buffer: super::Buffer) {}
+    unsafe fn destroy_buffer(&self, _buffer: super::Buffer) {
```
ErichDonGubler, line 356: style: You've taken pains to ensure two newlines between methods in other places, let's do that here, too. Pushed as a `fixup!` with 8a7b472a4a62b3ea27e3ff18910c1935249cf884.

wgpu-hal/src/metal/device.rs
```diff
@@ -508,15 +521,20 @@ impl crate::Device for super::Device {
             }
             let raw = self.shared.device.lock().new_sampler(&descriptor);
 
+            self.counters.samplers.add(1);
+
             Ok(super::Sampler { raw })
         })
     }
-    unsafe fn destroy_sampler(&self, _sampler: super::Sampler) {}
+    unsafe fn destroy_sampler(&self, _sampler: super::Sampler) {
```
ErichDonGubler, line 569: style: RE: two newlines: `fixup!`: 22a790c8af3e5bbf4b6812eb414b7ae065b5b879

ErichDonGubler (changes requested):

ErichDonGubler (changes requested):
🤦🏻 Oh man, I accidentally hit `Request changes` with just the `style` things written, but I intended more. I sure look like a jerk, eh?

I have more substantial feedback coming, but I guess I need to acknowledge my mistake _now_ to minimize the window of apparent jerk-ness.

wgpu-core/Cargo.toml
ErichDonGubler: question: Is there a reason we don't have `#[cfg(feature = "counters")]` in the new APIs and members we're adding here?

EDIT: Ah, I see that the inner implementation uses `cfg` internally. Clever! I don't have any objections at this time, then.

wgpu-hal/src/metal/device.rs
ErichDonGubler: Alright, I got tired of writing comments, just pushed up all-the-things whitespace fixes with b3f3c8dc3 (`fixup!` relevant to existing changes) and d230899a6 (not relevant, just annoyed enough to push it here, CC @Wumpf).

wgpu-types/src/lib.rs
```diff
@@ -7243,3 +7245,154 @@ pub enum DeviceLostReason {
...
+    }
+}
+
+/// `wgpu-hal`'s internal counters.
+#[derive(Clone, Default)]
+pub struct HalCounters {
+    // API objects
+    ///
+    pub buffers: InternalCounter,
+    ///
```
ErichDonGubler: issue: Newer Clippy versions will reject this (see also #5651); we should either document these here, or bite the bullet and use lint management to indicate intent without a warning:

```suggestion
    #[allow(empty_docs)]
```

wgpu-types/src/lib.rs
```diff
@@ -7243,3 +7245,154 @@ pub enum DeviceLostReason {
...
+    /// Get the counter's value.
+    #[cfg(feature = "counters")]
+    #[inline]
+    pub fn read(&self) -> isize {
+        self.value.load(Ordering::Relaxed)
+    }
+
+    /// Get the counter's value.
+    ///
+    /// Always returns 0 if the `counters` feature is not enabled.
```
ErichDonGubler: thought: I think it might be better to `cfg`-ify the body of these functions, and keep the docs. the same. This would also sidestep any problems if we accidentally let a function signature drift between feature selections.

ErichDonGubler (approved):
@Wumpf's blockage is a superset of anything I could think of blocking on, so approving here.

wgpu-types/src/lib.rs
ErichDonGubler: Implemented my suggestion with a491a5e46.

ErichDonGubler: style: It'd be nice to have another newline between the `// section header` comments that are also confusable with the doc. comments for fields.

Implemented this suggestion with 436f5e124; feel free to reject.

ErichDonGubler: 😮‍💨 Oh, sigh, both of these get reverted by the `rustfmt` we're using as part of `rust-toolchain.toml`. Guess I'll revert these.

ErichDonGubler:
Force-pushed and autosquashed my changes. Looks like some commit history could be squashed down a bit more, but I'll let @nical be the judge of how to do that.

nical:
Added the API at the `wgpu` level.

> I'm concerned on how to reconciliate this stuff with the existing [`HubReport`](https://docs.rs/wgpu/latest/wgpu/core/hub/struct.HubReport.html). What's happening in `InternalCounters` and `HubReport` is strongly overlapping from a users point of view even though the source of the information is quite different. Either we introduce clear distinction for what is what or (better imho) we merge these new counters into the existing hub report. Another interesting course of action would be to deprecate the hub report, but as long as we _have_ hubs they are quite useful.

I think that they should stay separate. They are two different ways to get different information. It looks like the information is similar but it's different important ways. Having two APIs is fine, and I'm about to a third for allocator reports.
The understanding you need to use either API is enough to get the difference between the two (if you know what the Hub is, then you know enough to guess that `get_internal_counters().hal.textures` will provide information from a different abstraction layer).
Also counters are cheap to query whereas hub report (and to a greater extent the upcoming allocator reports) require crawling potentially large data structures.

If they really have to be unified, then hub reports could be put into `CoreCounters` and would have to be gated by the `counters` feature.

Wumpf (approved):

Issue #2431, opened 2022-01-28
Ensure safety of indirect draw/dispatch


kvark:
When doing an indirect draw, the range of vertices/indices to draw needs to be within the bounds of the active vertex and index buffers. Similarly, indirect dispatches must be within the limits of the workgroup count.

teoxoy:
When we get to this, we should also validate usage of the `indirect-first-instance` feature.

> The value corresponding to firstInstance must be 0, unless the ["indirect-first-instance"](https://gpuweb.github.io/gpuweb/#indirect-first-instance) [feature](https://gpuweb.github.io/gpuweb/#feature) is enabled. If the ["indirect-first-instance"](https://gpuweb.github.io/gpuweb/#indirect-first-instance) [feature](https://gpuweb.github.io/gpuweb/#feature) is not enabled and firstInstance is not zero the [drawIndirect()](https://gpuweb.github.io/gpuweb/#dom-gpurendercommandsmixin-drawindirect) call will be treated as a no-op.

ErichDonGubler:
I've updated the OP with the scope of remaining changes I'm aware need to happen.

ErichDonGubler:
Downstream Firefox tracking for indirect dispatch: [bug 1921342](https://bugzilla.mozilla.org/show_bug.cgi?id=1921342)

ErichDonGubler:
Downstream Firefox tracking for indirect draws: [bug 1930972](https://bugzilla.mozilla.org/show_bug.cgi?id=1930972)

Issue #5320, opened 2024-02-29
Support diagnostic filters in WGSL (i.e., `diagnostic(…)`)


ErichDonGubler:
We currently do not support the `@diagnostic(…)` attribute/`diagnostic(…)` directive (AKA "diagnostic filters" in the WebGPU spec.). This is a rather complicated subsystem, but it's exciting to have robust management of diagnostics in WGSL [a la Rust](https://doc.rust-lang.org/reference/attributes/diagnostics.html). An example of how this is used from the spec.:

> EXAMPLE: Global diagnostic filter for derivative uniformity
>
> ```wgsl
> diagnostic(off,derivative_uniformity);
> var<private> d: f32;
> fn helper() -> vec4<f32> {
>   if (d < 0.5) {
>     // The derivative_uniformity diagnostic is disabled here
>     // by the global diagnostic filter.
>     return textureSample(t,s,vec2(0,0));
>   } else {
>     // The derivative_uniformity diagnostic is set to 'warning' severity.
>     @diagnostic(warning,derivative_uniformity) {
>       return textureSample(t,s,vec2(0,0));
>     }
>   }
>   return vec4(0.0);
> }
>
> ```

Relevant part of the spec.: [Diagnostics, section 2.3](https://gpuweb.github.io/gpuweb/wgsl/#diagnostics)

An initial implementation of this should include the `derivative_uniformity` rule, as the only current example of filterable trigger rules that can be managed per the spec.

Parse paths that we need to support to be spec.-compliant:

- [x] Directive: #6148
- [ ] Attributes placements (pasted from [the spec.](https://www.w3.org/TR/WGSL/#diagnostic-filtering)):
    - [ ] Beginning of a compound statement.
    - [ ] Beginning of a function declaration: #6353
    - [ ] Beginning of an `if` statement.
    - [ ] Beginning of a `switch statement`.
    - [ ] Beginning of a `switch_body`.
    - [ ] Beginning of a `loop` statement.
    - [ ] Beginning of a `while` statement.
    - [ ] Beginning of a `for` statement.
    - [ ] Immediately before the opening brace ('{') of the loop body of a loop, while, or for loop.
    - [ ] Beginning of a `continuing_compound_statement`.

ErichDonGubler:
Correspondent Firefox bug: [bug 1882800](https://bugzilla.mozilla.org/show_bug.cgi?id=1882800)

jimblandy:
This is blocking some BabylonJS demos.

[...]

Issue #5121, opened 2024-01-22
Remove IDs and Registries


nical:
This is a meta issue to describe the long-ish term vision for wgpu-core's resource management and track the steps towards it.

# Registries and IDs

Internally, all wgpu-core resources are stored behind `Arc`s. Externally, they are exposed as non-clonable structs containing an ID which conceptually corresponds to a single reference to the `Arc` of a resource.

# The goal

Instead of having wgpu public types wrap a non-clonabe ID, we would like them to wrap the `Arc` to the internal resource directly (and be clonable).

# TODO

- [x] #5120
- [x] Remove remaining internal use of the IDs.
- [x] #5124
- [x] Rewrite the record-reply functionality https://github.com/gfx-rs/wgpu/issues/5974
- [ ] #8911
- [ ] (Firefox): Move the `wgpu-core` part of command recording out of the content process
- [ ] Move logic present in the `Global`'s methods into methods on wgpu-core objects
- [ ] (Firefox): Stop providing the IDs during resource creation and do ID mapping in the gecko-side bindings.
- [ ] Remove the registries

# Other Related Issues

- [ ] #8234 
- [ ] #8837

sagudev:
> (Firefox): Stop providing the IDs during resource creation and do ID mapping in the gecko-side bindings.

How exactly would that look like? Would that be something like hashmap that would map id to Arc of wgpu object that would live in WebGPU thread?

I'm mostly asking for servo, as it would be nice to share ID mapping stuff.

nical:
> > (Firefox): Stop providing the IDs during resource creation and do ID mapping in the gecko-side bindings.
> 
> How exactly would that look like? Would that be something like hashmap that would map id to Arc of wgpu object that would live in WebGPU thread?

Yes, that's the general idea. Depending on how you generate the IDs you could get away with arrays instead of hash maps (like wgpu-core currently does).

sagudev:
There are some cases where this would also be useful in servo, for example when doing swapchain present we always need to send new encoder id from content process so we can create new encoder in wgpu_thread, but removing registries would allow us to have concept of wgc types that are not owned by JS (content process).

[...]

PR #5714, opened 2024-05-17
Ensure safety of indirect dispatch

teoxoy:
Ensure safety of indirect dispatch by injecting a compute shader that validates the content of the indirect buffer.

Part of https://github.com/gfx-rs/wgpu/issues/2431.

~~Depends on #6318.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -133,6 +133,7 @@ By @bradwerth [#6216](https://github.com/gfx-rs/wgpu/pull/6216).
 - Call `flush_mapped_ranges` when unmapping write-mapped buffers. By @teoxoy in [#6089](https://github.com/gfx-rs/wgpu/pull/6089).
 - When mapping buffers for reading, mark buffers as initialized only when they have `MAP_WRITE` usage. By @teoxoy in [#6178](https://github.com/gfx-rs/wgpu/pull/6178).
 - Add a separate pipeline constants error. By @teoxoy in [#6094](https://github.com/gfx-rs/wgpu/pull/6094).
+- Ensure safety of indirect dispatch by injecting a compute shader that validates the content of the indirect buffer. By @teoxoy in [#5714](https://github.com/gfx-rs/wgpu/pull/5714)
 
 #### GLES / OpenGL
 
--- a/tests/tests/dispatch_workgroups_indirect.rs
+++ b/tests/tests/dispatch_workgroups_indirect.rs
@@ -1,0 +1,241 @@
+use wgpu_test::{gpu_test, FailureCase, GpuTestConfiguration, TestParameters, TestingContext};
+
+/// Make sure that the num_workgroups builtin works properly (it requires a workaround on D3D12).
+#[gpu_test]
+static NUM_WORKGROUPS_BUILTIN: GpuTestConfiguration = GpuTestConfiguration::new()
+    .parameters(
+        TestParameters::default()
+            .features(wgpu::Features::PUSH_CONSTANTS)
+            .downlevel_flags(
+                wgpu::DownlevelFlags::COMPUTE_SHADERS | wgpu::DownlevelFlags::INDIRECT_EXECUTION,
+            )
+            .limits(wgpu::Limits {
+                max_push_constant_size: 4,
+                ..wgpu::Limits::downlevel_defaults()
+            })
+            .expect_fail(FailureCase::backend(wgt::Backends::DX12)),
+    )
+    .run_async(|ctx| async move {
+        let num_workgroups = [1, 2, 3];
+        let res = run_test(&ctx, &num_workgroups, false).await;
+        assert_eq!(res, num_workgroups);
+    });
+
+/// Make sure that we discard (don't run) the dispatch if its size exceeds the device limit.
+#[gpu_test]
+static DISCARD_DISPATCH: GpuTestConfiguration = GpuTestConfiguration::new()
+    .parameters(
+        TestParameters::default()
+            .features(wgpu::Features::PUSH_CONSTANTS)
+            .downlevel_flags(
+                wgpu::DownlevelFlags::COMPUTE_SHADERS | wgpu::DownlevelFlags::INDIRECT_EXECUTION,
+            )
+            .limits(wgpu::Limits {
+                max_compute_workgroups_per_dimension: 10,
+                max_push_constant_size: 4,
+                ..wgpu::Limits::downlevel_defaults()
+            })
+            .expect_fail(FailureCase::backend(wgt::Backends::DX12)),
+    )
+    .run_async(|ctx| async move {
+        let max = ctx.device.limits().max_compute_workgroups_per_dimension;
+
+        let res = run_test(&ctx, &[max, max, max], false).await;
+        assert_eq!(res, [max; 3]);
+
+        let res = run_test(&ctx, &[max + 1, 1, 1], false).await;
+        assert_eq!(res, [0; 3]);
+
+        let res = run_test(&ctx, &[1, max + 1, 1], false).await;
+        assert_eq!(res, [0; 3]);
+
+        let res = run_test(&ctx, &[1, 1, max + 1], false).await;
+        assert_eq!(res, [0; 3]);
+    });
+
+/// Make sure that resetting the bind groups set by the validation code works properly.
+#[gpu_test]
+static RESET_BIND_GROUPS: GpuTestConfiguration = GpuTestConfiguration::new()
+    .parameters(
+        TestParameters::default()
+            .features(wgpu::Features::PUSH_CONSTANTS)
+            .downlevel_flags(
+                wgpu::DownlevelFlags::COMPUTE_SHADERS | wgpu::DownlevelFlags::INDIRECT_EXECUTION,
+            )
+            .limits(wgpu::Limits {
+                max_push_constant_size: 4,
+                ..wgpu::Limits::downlevel_defaults()
+            }),
+    )
+    .run_async(|ctx| async move {
+        ctx.device.push_error_scope(wgpu::ErrorFilter::Validation);
+
+        let _ = run_test(&ctx, &[0, 0, 0], true).await;
+
+        let error = pollster::block_on(ctx.device.pop_error_scope());
+        assert!(error.map_or(false, |error| {
+            format!("{error}").contains("The current set ComputePipeline with '' label expects a BindGroup to be set at index 0")
+        }));
+    });
+
+async fn run_test(
+    ctx: &TestingContext,
+    num_workgroups: &[u32; 3],
+    forget_to_set_bind_group: bool,
+) -> [u32; 3] {
+    const SHADER_SRC: &str = "
+        struct TestOffsetPc {
+            inner: u32,
+        }
+
+        // `test_offset.inner` should always be 0; we test that resetting the push constant set by the validation code works properly.
+        var<push_constant> test_offset: TestOffsetPc;
+
+        @group(0) @binding(0)
+        var<storage, read_write> out: array<u32, 3>;
+
+        @compute @workgroup_size(1)
+        fn main(@builtin(num_workgroups) num_workgroups: vec3u, @builtin(workgroup_id) workgroup_id: vec3u) {
+            if (all(workgroup_id == vec3u())) {
+                out[0] = num_workgroups.x + test_offset.inner;
+                out[1] = num_workgroups.y + test_offset.inner;
+                out[2] = num_workgroups.z + test_offset.inner;
+            }
+        }
+    ";
+
+    let module = ctx
+        .device
+        .create_shader_module(wgpu::ShaderModuleDescriptor {
+            label: None,
+            source: wgpu::ShaderSource::Wgsl(SHADER_SRC.into()),
+        });
+
+    let bgl = ctx
+        .device
+        .create_bind_group_layout(&wgpu::BindGroupLayoutDescriptor {
+            label: None,
+            entries: &[wgpu::BindGroupLayoutEntry {
+                binding: 0,
+                visibility: wgt::ShaderStages::COMPUTE,
+                ty: wgpu::BindingType::Buffer {
+                    ty: wgpu::BufferBindingType::Storage { read_only: false },
+                    has_dynamic_offset: false,
+                    min_binding_size: None,
+                },
+                count: None,
+            }],
+        });
+
+    let layout = ctx
+        .device
+        .create_pipeline_layout(&wgpu::PipelineLayoutDescriptor {
+            label: None,
+            bind_group_layouts: &[&bgl],
+            push_constant_ranges: &[wgt::PushConstantRange {
+                stages: wgt::ShaderStages::COMPUTE,
+                range: 0..4,
+            }],
+        });
+
+    let pipeline = ctx
+        .device
+        .create_compute_pipeline(&wgpu::ComputePipelineDescriptor {
+            label: None,
+            layout: Some(&layout),
+            module: &module,
+            entry_point: Some("main"),
+            compilation_options: Default::default(),
+            cache: None,
+        });
+
+    let out_buffer = ctx.device.create_buffer(&wgpu::BufferDescriptor {
+        label: None,
+        size: 12,
+        usage: wgpu::BufferUsages::STORAGE | wgpu::BufferUsages::COPY_SRC,
+        mapped_at_creation: false,
+    });
+
+    let readback_buffer = ctx.device.create_buffer(&wgpu::BufferDescriptor {
+        label: None,
+        size: 12,
+        usage: wgpu::BufferUsages::COPY_DST | wgpu::BufferUsages::MAP_READ,
+        mapped_at_creation: false,
+    });
+
+    let bind_group = ctx.device.create_bind_group(&wgpu::BindGroupDescriptor {
+        label: None,
+        layout: &pipeline.get_bind_group_layout(0),
+        entries: &[wgpu::BindGroupEntry {
+            binding: 0,
+            resource: out_buffer.as_entire_binding(),
+        }],
+    });
+
+    let mut res = None;
+
+    for (indirect_offset, indirect_buffer_size) in [
+        // internal src buffer binding size will be buffer.size
+        (0, 12),
+        (4, 4 + 12),
+        (4, 8 + 12),
+        (256 * 2 - 4 - 12, 256 * 2 - 4),
+        // internal src buffer binding size will be 256 * 2 + x
+        (0, 256 * 2 * 2 + 4),
+        (256, 256 * 2 * 2 + 8),
+        (256 + 4, 256 * 2 * 2 + 12),
+        (256 * 2 + 16, 256 * 2 * 2 + 16),
+        (256 * 2 * 2, 256 * 2 * 2 + 32),
+        (256 + 12, 256 * 2 * 2 + 64),
+    ] {
+        let indirect_buffer = ctx.device.create_buffer(&wgpu::BufferDescriptor {
+            label: None,
+            size: indirect_buffer_size,
+            usage: wgpu::BufferUsages::COPY_DST | wgpu::BufferUsages::INDIRECT,
+            mapped_at_creation: false,
+        });
+
+        ctx.queue.write_buffer(
+            &indirect_buffer,
+            indirect_offset,
+            bytemuck::bytes_of(num_workgroups),
+        );
+
+        let mut encoder = ctx.device.create_command_encoder(&Default::default());
+        {
+            let mut compute_pass = encoder.begin_compute_pass(&Default::default());
+            compute_pass.set_pipeline(&pipeline);
+            compute_pass.set_push_constants(0, &[0, 0, 0, 0]);
+            if !forget_to_set_bind_group {
+                compute_pass.set_bind_group(0, Some(&bind_group), &[]);
+            }
+            compute_pass.dispatch_workgroups_indirect(&indirect_buffer, indirect_offset);
+        }
+
+        encoder.copy_buffer_to_buffer(&out_buffer, 0, &readback_buffer, 0, 12);
+
+        ctx.queue.submit(Some(encoder.finish()));
+
+        readback_buffer
+            .slice(..)
+            .map_async(wgpu::MapMode::Read, |_| {});
+
+        ctx.async_poll(wgpu::Maintain::wait())
+            .await
+            .panic_on_timeout();
+
+        let view = readback_buffer.slice(..).get_mapped_range();
+
+        let current_res = *bytemuck::from_bytes(&view);
+        drop(view);
+        readback_buffer.unmap();
+
+        if let Some(past_res) = res {
+            assert_eq!(past_res, current_res);
+        } else {
+            res = Some(current_res);
+        }
+    }
+
+    res.unwrap()
+}
--- a/wgpu-core/src/command/bind.rs
+++ b/wgpu-core/src/command/bind.rs
@@ -200,13 +200,17 @@ mod compat {
                 entries: (0..hal::MAX_BIND_GROUPS).map(|_| Entry::empty()).collect(),
             }
         }
-        fn make_range(&self, start_index: usize) -> Range<usize> {
+
+        pub fn num_valid_entries(&self) -> usize {
             // find first incompatible entry
-            let end = self
-                .entries
+            self.entries
                 .iter()
                 .position(|e| e.is_incompatible())
-                .unwrap_or(self.entries.len());
+                .unwrap_or(self.entries.len())
+        }
+
+        fn make_range(&self, start_index: usize) -> Range<usize> {
+            let end = self.num_valid_entries();
             start_index..end.max(start_index)
         }
 
--- a/wgpu-core/src/command/bind.rs
+++ b/wgpu-core/src/command/bind.rs
@@ -410,6 +410,14 @@ impl Binder {
             .map(move |index| payloads[index].group.as_ref().unwrap())
     }
 
+    #[cfg(feature = "indirect-validation")]
+    pub(super) fn list_valid<'a>(&'a self) -> impl Iterator<Item = (usize, &'a EntryPayload)> + '_ {
+        self.payloads
+            .iter()
+            .take(self.manager.num_valid_entries())
+            .enumerate()
+    }
+
     pub(super) fn check_compatibility<T: Labeled>(
         &self,
         pipeline: &T,
--- a/wgpu-core/src/command/compute.rs
+++ b/wgpu-core/src/command/compute.rs
@@ -18,7 +18,7 @@ use crate::{
     pipeline::ComputePipeline,
     resource::{
         self, Buffer, DestroyedResourceError, InvalidResourceError, Labeled,
-        MissingBufferUsageError, ParentDevice, Trackable,
+        MissingBufferUsageError, ParentDevice,
     },
     snatch::SnatchGuard,
     track::{ResourceUsageCompatibilityError, Tracker, TrackerIndex, UsageScope},
@@ -215,6 +215,8 @@ struct State<'scope, 'snatch_guard, 'cmd_buf, 'raw_encoder> {
     dynamic_offset_count: usize,
     string_offset: usize,
     active_query: Option<(Arc<resource::QuerySet>, u32)>,
+
+    push_constants: Vec<u32>,
 
     intermediate_trackers: Tracker,
 
@@ -442,6 +444,8 @@ impl Global {
             dynamic_offset_count: 0,
             string_offset: 0,
             active_query: None,
+
+            push_constants: Vec::new(),
 
             intermediate_trackers: Tracker::new(),
 
@@ -746,6 +750,21 @@ fn set_pipeline(
             }
         }
 
+        // TODO: integrate this in the code below once we simplify push constants
+        state.push_constants.clear();
+        // Note that can only be one range for each stage. See the `MoreThanOnePushConstantRangePerStage` error.
+        if let Some(push_constant_range) =
+            pipeline.layout.push_constant_ranges.iter().find_map(|pcr| {
+                pcr.stages
+                    .contains(wgt::ShaderStages::COMPUTE)
+                    .then_some(pcr.range.clone())
+            })
+        {
+            // Note that non-0 range start doesn't work anyway https://github.com/gfx-rs/wgpu/issues/4502
+            let len = push_constant_range.len() / wgt::PUSH_CONSTANT_ALIGNMENT as usize;
+            state.push_constants.extend(core::iter::repeat(0).take(len));
+        }
+
         // Clear push constant ranges
         let non_overlapping =
             super::bind::compute_nonoverlapping_ranges(&pipeline.layout.push_constant_ranges);
@@ -790,6 +809,10 @@ fn set_push_constant(
         offset,
         end_offset_bytes,
     )?;
+
+    let offset_in_elements = (offset / wgt::PUSH_CONSTANT_ALIGNMENT) as usize;
+    let size_in_elements = (size_bytes / wgt::PUSH_CONSTANT_ALIGNMENT) as usize;
+    state.push_constants[offset_in_elements..][..size_in_elements].copy_from_slice(data_slice);
 
     unsafe {
         state.raw_encoder.set_push_constants(
@@ -841,10 +864,6 @@ fn dispatch_indirect(
         .device
         .require_downlevel_flags(wgt::DownlevelFlags::INDIRECT_EXECUTION)?;
 
-    state
-        .scope
-        .buffers
-        .merge_single(&buffer, hal::BufferUses::INDIRECT)?;
     buffer.check_usage(wgt::BufferUsages::INDIRECT)?;
 
     if offset % 4 != 0 {
@@ -861,7 +880,6 @@ fn dispatch_indirect(
     }
 
     let stride = 3 * 4; // 3 integers, x/y/z group size
-
     state
         .buffer_memory_init_actions
         .extend(buffer.initialization_status.read().create_action(
@@ -870,12 +888,132 @@ fn dispatch_indirect(
             MemoryInitKind::NeedsInitializedMemory,
         ));
 
-    state.flush_states(Some(buffer.tracker_index()))?;
+    #[cfg(feature = "indirect-validation")]
+    {
+        let params = state.device.indirect_validation.as_ref().unwrap().params(
+            &state.device.limits,
+            offset,
+            buffer.size,
+        );
 
-    let buf_raw = buffer.try_raw(&state.snatch_guard)?;
-    unsafe {
-        state.raw_encoder.dispatch_indirect(buf_raw, offset);
+        unsafe {
+            state.raw_encoder.set_compute_pipeline(params.pipeline);
+        }
+
+        unsafe {
+            state.raw_encoder.set_push_constants(
+                params.pipeline_layout,
+                wgt::ShaderStages::COMPUTE,
+                0,
+                &[params.offset_remainder as u32 / 4],
+            );
+        }
+
+        unsafe {
+            state.raw_encoder.set_bind_group(
+                params.pipeline_layout,
+                0,
+                Some(params.dst_bind_group),
+                &[],
+            );
+        }
+        unsafe {
+            state.raw_encoder.set_bind_group(
+                params.pipeline_layout,
+                1,
+                Some(
+                    buffer
+                        .raw_indirect_validation_bind_group
+                        .get(&state.snatch_guard)
+                        .unwrap()
+                        .as_ref(),
+                ),
+                &[params.aligned_offset as u32],
+            );
+        }
+
+        let src_transition = state
+            .intermediate_trackers
+            .buffers
+            .set_single(&buffer, hal::BufferUses::STORAGE_READ);
+        let src_barrier =
+            src_transition.map(|transition| transition.into_hal(&buffer, &state.snatch_guard));
+        unsafe {
+            state.raw_encoder.transition_buffers(src_barrier.as_slice());
+        }
+
+        unsafe {
+            state.raw_encoder.transition_buffers(&[hal::BufferBarrier {
+                buffer: params.dst_buffer,
+                usage: hal::BufferUses::INDIRECT..hal::BufferUses::STORAGE_READ_WRITE,
+            }]);
+        }
+
+        unsafe {
+            state.raw_encoder.dispatch([1, 1, 1]);
+        }
+
+        // reset state
+        {
+            let pipeline = state.pipeline.as_ref().unwrap();
+
+            unsafe {
+                state.raw_encoder.set_compute_pipeline(pipeline.raw());
+            }
+
+            if !state.push_constants.is_empty() {
+                unsafe {
+                    state.raw_encoder.set_push_constants(
+                        pipeline.layout.raw(),
+                        wgt::ShaderStages::COMPUTE,
+                        0,
+                        &state.push_constants,
+                    );
+                }
+            }
+
+            for (i, e) in state.binder.list_valid() {
+                let group = e.group.as_ref().unwrap();
+                let raw_bg = group.try_raw(&state.snatch_guard)?;
+                unsafe {
+                    state.raw_encoder.set_bind_group(
+                        pipeline.layout.raw(),
+                        i as u32,
+                        Some(raw_bg),
+                        &e.dynamic_offsets,
+                    );
+                }
+            }
+        }
+
+        unsafe {
+            state.raw_encoder.transition_buffers(&[hal::BufferBarrier {
+                buffer: params.dst_buffer,
+                usage: hal::BufferUses::STORAGE_READ_WRITE..hal::BufferUses::INDIRECT,
+            }]);
+        }
+
+        state.flush_states(None)?;
+        unsafe {
+            state.raw_encoder.dispatch_indirect(params.dst_buffer, 0);
+        }
+    };
+    #[cfg(not(feature = "indirect-validation"))]
+    {
+        state
+            .scope
+            .buffers
+            .merge_single(&buffer, hal::BufferUses::INDIRECT)?;
+
+        use crate::resource::Trackable;
+        state.flush_states(Some(buffer.tracker_index()))?;
+
+        let buf_raw = buffer.try_raw(&state.snatch_guard)?;
+        unsafe {
+            state.raw_encoder.dispatch_indirect(buf_raw, offset);
+        }
     }
+
     Ok(())
 }
 
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -406,12 +406,12 @@ impl Global {
             trace.add(trace::Action::CreateBuffer(fid.id(), desc.clone()));
         }
 
-        let buffer = device.create_buffer_from_hal(Box::new(hal_buffer), desc);
+        let (buffer, err) = device.create_buffer_from_hal(Box::new(hal_buffer), desc);
 
         let id = fid.assign(buffer);
         api_log!("Device::create_buffer -> {id:?}");
 
-        (id, None)
+        (id, err)
     }
 
     pub fn texture_destroy(&self, texture_id: id::TextureId) -> Result<(), resource::DestroyError> {
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -31,7 +31,7 @@ use crate::{
         UsageScopePool,
     },
     validation::{self, validate_color_attachment_bytes_per_sample},
-    FastHashMap, LabelHelpers as _, PreHashedKey, PreHashedMap,
+    FastHashMap, LabelHelpers, PreHashedKey, PreHashedMap,
 };
 
 use arrayvec::ArrayVec;
@@ -144,6 +144,9 @@ pub struct Device {
     #[cfg(feature = "trace")]
     pub(crate) trace: Mutex<Option<trace::Trace>>,
     pub(crate) usage_scopes: UsageScopePool,
+
+    #[cfg(feature = "indirect-validation")]
+    pub(crate) indirect_validation: Option<crate::indirect_validation::IndirectValidation>,
 }
 
 pub(crate) enum DeferredDestroy {
@@ -175,6 +178,11 @@ impl Drop for Device {
         let fence = unsafe { ManuallyDrop::take(&mut self.fence.write()) };
         pending_writes.dispose(raw.as_ref());
         self.command_allocator.dispose(raw.as_ref());
+        #[cfg(feature = "indirect-validation")]
+        self.indirect_validation
+            .take()
+            .unwrap()
+            .dispose(raw.as_ref());
         unsafe {
             raw.destroy_buffer(zero_buffer);
             raw.destroy_fence(fence);
@@ -261,6 +269,25 @@ impl Device {
         let alignments = adapter.raw.capabilities.alignments.clone();
         let downlevel = adapter.raw.capabilities.downlevel.clone();
 
+        #[cfg(feature = "indirect-validation")]
+        let indirect_validation = if downlevel
+            .flags
+            .contains(wgt::DownlevelFlags::INDIRECT_EXECUTION)
+        {
+            match crate::indirect_validation::IndirectValidation::new(
+                raw_device.as_ref(),
+                &desc.required_limits,
+            ) {
+                Ok(indirect_validation) => Some(indirect_validation),
+                Err(e) => {
+                    log::error!("indirect-validation error: {e:?}");
+                    return Err(DeviceError::Lost);
+                }
+            }
+        } else {
+            None
+        };
+
         Ok(Self {
             raw: ManuallyDrop::new(raw_device),
             adapter: adapter.clone(),
@@ -306,6 +333,8 @@ impl Device {
             ),
             deferred_destroy: Mutex::new(rank::DEVICE_DEFERRED_DESTROY, Vec::new()),
             usage_scopes: Mutex::new(rank::DEVICE_USAGE_SCOPES, Default::default()),
+            #[cfg(feature = "indirect-validation")]
+            indirect_validation,
         })
     }
 
@@ -547,6 +576,13 @@ impl Device {
 
         let mut usage = conv::map_buffer_usage(desc.usage);
 
+        if desc.usage.contains(wgt::BufferUsages::INDIRECT) {
+            self.require_downlevel_flags(wgt::DownlevelFlags::INDIRECT_EXECUTION)?;
+            // We are going to be reading from it, internally;
+            // when validating the content of the buffer
+            usage |= hal::BufferUses::STORAGE_READ | hal::BufferUses::STORAGE_READ_WRITE;
+        }
+
         if desc.mapped_at_creation {
             if desc.size % wgt::COPY_BUFFER_ALIGNMENT != 0 {
                 return Err(resource::CreateBufferError::UnalignedSize);
@@ -586,6 +622,10 @@ impl Device {
         let buffer =
             unsafe { self.raw().create_buffer(&hal_desc) }.map_err(|e| self.handle_hal_error(e))?;
 
+        #[cfg(feature = "indirect-validation")]
+        let raw_indirect_validation_bind_group =
+            self.create_indirect_validation_bind_group(buffer.as_ref(), desc.size, desc.usage)?;
+
         let buffer = Buffer {
             raw: Snatchable::new(buffer),
             device: self.clone(),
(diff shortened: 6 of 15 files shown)


teoxoy:
Hmm, I'm not sure how we should deal with the `wgsl` feature not being available.

cwfitzgerald:
Let's just require it in core, honestly.

We could build a naga module in a build script, serialize it and store that in the binary, but that's wayyyy to much work for no real gain

teoxoy:
> We could build a naga module in a build script, serialize it and store that in the binary, but that's wayyyy to much work for no real gain

That's what I thought as well.

Requiring it in `wgpu-core` will make `wgpu`'s `wgsl` feature meaningless as well. Removing the `wgsl` feature from both `wgpu-core` and `wgpu` will effectively revert https://github.com/gfx-rs/wgpu/pull/2890.

That PR was prefaced with:

> It also might turn out entirely unnecessary if the compile-time savings aren't worth the added complexity.

This was also mentioned in https://github.com/gfx-rs/wgpu/issues/2549#issuecomment-1074525415:

> If you want to provide more concrete numbers on the binary size (that could be saved), that would help.

@daxpedda were the benefits of https://github.com/gfx-rs/wgpu/pull/2890 substantial in terms of compile time and binary size?

jimblandy (dismissed):
In our office hours meeting this morning, we came up with a better way to test, by requesting a device with a lower limit than the adapter offers. Testing this way would give a more positive indication that the limit is being imposed.

daxpedda:
> > If you want to provide more concrete numbers on the binary size (that could be saved), that would help.
> 
> @daxpedda were the benefits of #2890 substantial in terms of compile time and binary size?

I just tried a minimal example when using GLSL shaders on Wasm with WebGL.
Using fat LTO, a single codegen unit, reference types and `wasm-opt` for all scenarios.

|                		| O3 (O4 for  `wasm-opt`)	| Os     		          	| Oz               		|
|-				|-:						|-:					|-:					|
| without `wgsl` 	| 3 081 862 				| 2 070 971 			| 1 871 247 			|
| with `wgsl`    	| 3 428 880 				| 2 306 315 			| 2 084 794 			|
| diff           		| (11.26%) 347 018 		| (11.36%) 235 344 	| (11.41%) 213 547 	|

Compile times are similar. Total compile time with O3 goes from 12.21s to 13.48s, so again around a 10% increase.
For Naga alone (according to [Cargo timings](https://doc.rust-lang.org/cargo/reference/timings.html)) I'm getting an increase from 7.74s to 9.67s, so around 25% longer.

Keep in mind that this is a minimal example and might not be representative for a full blown application.

teoxoy:
I would say 10% is significant especially for web apps but I'd be curious how much it decreases with larger applications.

It's unfortunate though that since WebGL doesn't support indirect calls, this increase doesn't bring anything of value for this use case.

Maybe we can feature gate the validation on the `wgsl` feature and enable `wgc/wgsl` when `not(target_arch = "wasm32")` in `wgpu`?

The downside being that `wgpu`'s `wgsl` feature won't be respected (`wgsl` input will always be enabled) on native platforms (non-web).

teoxoy:
@jimblandy I lowered the limit required to 10 since the testing infra is not setup to be able to require limits based on what the adapter supports. I think 10 should be safe, the default value is 65535.

teoxoy:
I added another test for the `num_workgroups` builtin (expected to fail on dx12 - for now).

teoxoy:
I wanted to get CI working so I just implemented https://github.com/gfx-rs/wgpu/pull/5714#issuecomment-2122479252 for now.

teoxoy:
Filed https://github.com/gfx-rs/wgpu/issues/5739 for the metal issue which I've worked around in this PR.

teoxoy:
The `wgpu_test::compute_pass_resource_ownership::compute_pass_resource_ownership` test is reliably failing on all platforms.

And is because the internal call to `device_create_bind_group` tries to get the buffer via its id which should no longer be used.

I was hoping to get away with calling `wgpu-core`'s public APIs to implement this validation but it seems that won't work.

teoxoy:
Actually, I don't think the test is correct. Shouldn't our `Drop` implementation behave the same as calling `.destroy()` from the spec?

The `destroyed` property is checked by all `Queue` commands (notably `.submit()` in this case); the command only succeeding if none of the resources referenced by the command buffer have been tagged as `destroyed`.

The test was recently added in https://github.com/gfx-rs/wgpu/pull/5570.

> Added a new test that ensures that resources passed into a ComputePass can be dropped before executing the compute pass.
Proved that the test would crash before these changes and passes with them.

@Wumpf could you chime in on this? Did we use to crash or raise a validation error?

cwfitzgerald:
> Actually, I don't think the test is correct. Shouldn't our Drop implementation behave the same as calling .destroy() from the spec?

No, drop just removes the user's handle to it, doesn't make access to the resource invalid.

tests/tests/dispatch_workgroups_indirect.rs
```diff
@@ -0,0 +1,162 @@
...
+        @group(0) @binding(0)
+        var<storage, read_write> out: array<u32, 3>;
+
+        @compute @workgroup_size(1)
+        fn main(@builtin(num_workgroups) num_workgroups: vec3<u32>, @builtin(workgroup_id) workgroup_id: vec3<u32>) {
+            if (all(workgroup_id == vec3<u32>())) {
+                out[0] = num_workgroups.x;
+                out[1] = num_workgroups.y;
+                out[2] = num_workgroups.z;
+            }
```
jimblandy: Can we just match the type of `@builtin(num_workgroups)` by using a `vec3<u32>` here?

```suggestion
        var<storage, read_write> out: vec3<u32>;

        @compute @workgroup_size(1)
        fn main(@builtin(num_workgroups) num_workgroups: vec3<u32>, @builtin(workgroup_id) workgroup_id: vec3<u32>) {
            if (all(workgroup_id == vec3<u32>())) {
                out = num_workgroups;
            }
```

tests/tests/dispatch_workgroups_indirect.rs
```diff
@@ -0,0 +1,162 @@
...
+        .slice(..)
+        .map_async(wgpu::MapMode::Read, |_| {});
+
+    ctx.async_poll(wgpu::Maintain::wait())
+        .await
+        .panic_on_timeout();
+
+    let view = readback_buffer.slice(..).get_mapped_range();
+
+    bytemuck::from_bytes::<[u32; 3]>(&view).to_owned()
```
jimblandy: This seems simpler, since `[u8; N]` is `Copy`, and we can let type inference supply the type parameter:

```suggestion
    *bytemuck::from_bytes(&view)
```

tests/tests/dispatch_workgroups_indirect.rs
```diff
@@ -0,0 +1,162 @@
...
+            )
+            .limits(wgpu::Limits {
+                max_compute_workgroups_per_dimension: 10,
+                ..wgpu::Limits::downlevel_defaults()
+            }),
+    )
+    .run_async(|ctx| async move {
+        let max = ctx.device.limits().max_compute_workgroups_per_dimension;
+        let res = run_test(&ctx, &[max + 1, 1, 1], false).await;
+        assert_eq!(res, [0; 3]);
```
jimblandy, line 47: I don't know if it's worth it, but it's easy enough to check each dimension separately:

```suggestion

        let res = run_test(&ctx, &[max, max, max], false).await;
        assert_eq!(res, [max; 3]);

        let res = run_test(&ctx, &[max + 1, 1, 1], false).await;
        assert_eq!(res, [0; 3]);

        let res = run_test(&ctx, &[1, max + 1, 1], false).await;
        assert_eq!(res, [0; 3]);

        let res = run_test(&ctx, &[1, 1, max + 1], false).await;
        assert_eq!(res, [0; 3]);
```

tests/tests/dispatch_workgroups_indirect.rs
```diff
@@ -0,0 +1,162 @@
...
+    });
+
+    let readback_buffer = ctx.device.create_buffer(&wgpu::BufferDescriptor {
+        label: None,
+        size: 12,
+        usage: wgpu::BufferUsages::COPY_DST | wgpu::BufferUsages::MAP_READ,
+        mapped_at_creation: false,
+    });
+
+    const INDIRECT_OFFSET: u64 = 256 + 4; // unaligned offset (min_storage_buffer_offset_alignment is 256 by default)
```
jimblandy: It would be nice to have more explanation here about exactly why it helps to dispatch from 256 bytes into the buffer here. I see that `min_storage_buffer_offset_alignment` is 256, but how does that matter to the test?

jimblandy (commented):
Some thoughts so far:

tests/tests/dispatch_workgroups_indirect.rs
teoxoy: Metal has an issue with `vec3`s as the top level type of a host shareable `var` (the infamous `vec3` issues).
https://github.com/gfx-rs/wgpu/pull/5714#issuecomment-2129944114

tests/tests/dispatch_workgroups_indirect.rs
teoxoy: I was concerned with execution time but since `max` is now 10, this is probably fine.

tests/tests/dispatch_workgroups_indirect.rs
teoxoy: I added another paragraph, let me know if it looks good.

teoxoy:
> > Actually, I don't think the test is correct. Shouldn't our Drop implementation behave the same as calling .destroy() from the spec?
> 
> No, drop just removes the user's handle to it, doesn't make access to the resource invalid.

I see, I will try to convert the code doing bindgroup creation to use `Arc`s instead of `Id`s (moving in the direction of https://github.com/gfx-rs/wgpu/issues/5121) to see how feasible that is.

wgpu-core/src/device/resource.rs
```diff
@@ -96,6 +96,9 @@ pub struct Device<A: HalApi> {
     pub(crate) queue: OnceCell<Weak<Queue<A>>>,
     queue_to_drop: OnceCell<A::Queue>,
     pub(crate) zero_buffer: Option<A::Buffer>,
+    #[cfg(feature = "wgsl")]
+    pub(crate) dispatch_indirect_validation_pipeline:
+        Mutex<Option<(id::ComputePipelineId, id::BindGroupLayoutId)>>,
```
cwfitzgerald: Why can't we create this on device start?

wgpu-core/src/command/compute.rs
```diff
@@ -851,6 +889,211 @@ impl Global {
...
+                        if let Some(error) = error {
+                            return Err(ComputePassIndirectValidationError::Buffer(error))
+                                .map_pass_err(scope)?;
+                        }
+
+                        // align offset to min_storage_buffer_offset_alignment
+                        let aligned_offset = offset
+                            - offset % device.limits.min_storage_buffer_offset_alignment as u64;
+
+                        let (bind_group_id, error) = self.device_create_bind_group::<A>(
```
cwfitzgerald: When do these bind groups/buffers get dropped? As it looks now they will exist forever.

wgpu-core/src/command/compute.rs
```diff
@@ -851,6 +889,211 @@ impl Global {
...
+                        let (dst_buffer_id, error) = self.device_create_buffer::<A>(
+                            device_id,
+                            &crate::resource::BufferDescriptor {
+                                label: None,
+                                size: 4 * 3,
+                                usage: wgt::BufferUsages::INDIRECT | wgt::BufferUsages::STORAGE,
+                                mapped_at_creation: false,
+                            },
+                            None,
+                        );
```
cwfitzgerald: Creating a buffer and bind group every dispatch is going to be _very_ expensive. In this case we can make a single buffer/bind group per queue and a single bind group per indirect command buffer. We'd trade one bind group for two, but mean we never have to create resources at runtime.

I really wish we had root-uavs/bda so we didn't need a bind group here.

wgpu-core/src/device/global.rs
```diff
@@ -1767,6 +1767,135 @@ impl Global {
         (id, Some(error))
     }
 
+    #[cfg(feature = "wgsl")]
+    pub(crate) fn device_get_or_create_dispatch_indirect_validation_pipeline<A: HalApi>(
+        &self,
+        device_id: DeviceId,
+    ) -> Result<
+        (id::ComputePipelineId, id::BindGroupLayoutId),
```
cwfitzgerald: Could we wrap this functionality in an encapsulated type with constructor and execute functions, preferably in its own file?

wgpu-core/src/device/resource.rs
```diff
@@ -1902,7 +1916,19 @@ impl<A: HalApi> Device<A> {
...
+            bb.buffer_id,
+            buffer.usage
+                // Allow indirect buffers to be used as storage buffers, allowing
+                // us to bind them and validate their content. Note that we already
+                // pass this usage to hal at buffer creation.
+                | if buffer.usage.contains(wgt::BufferUsages::INDIRECT) {
+                    wgt::BufferUsages::STORAGE
+                } else {
+                    wgt::BufferUsages::empty()
+                },
```
cwfitzgerald: Do we need to be passing these into check_buffer_usage? If the user will do something wrong with usage, they will get an error message that includes a buffer usage that they did not request. We should just assume our code is right (or strict_assert maybe).

cwfitzgerald (dismissed):
Performance/code simplification thoughts.

Dispatches are unfortunately not trivial to validate all at once :(

wgpu-core/src/command/compute.rs
teoxoy: Wouldn't a single dst buffer (with its own bind group) per queue become a bottleneck in cases where a user issues 2 `dispatchIndirect` commands back to back each with their own indirect buffer? By only having one dst buffer, the 2 internal `dispatche`s we do for validation can't run in parallel and I think nor can the 2nd `dispatchIndirect` start before the first `dispatchIndirect` finished.

wgpu-core/src/device/resource.rs
teoxoy: Good point!

wgpu-core/src/device/resource.rs
teoxoy: We can, and it will simplify things - will do that.

wgpu-core/src/command/compute.rs
teoxoy: That's a good point, I was thinking it will happen eventually since we use `Arc`s but I missed we still have the registry and `Id`s 😅. I think what I have to do is just remove the id from the registry.

wgpu-core/src/device/resource.rs
teoxoy: This is turning out difficult to do because I have to convert all code to hal code.

teoxoy: I think I found a way to get this working by first initializing it with `None` and later when we have the `Arc<Device>` and can use all its methods, I set it to `Some`.

wgpu-core/src/command/compute.rs
teoxoy: We discussed this at our maintainers meeting and decided to go with Connor's suggestion. My concerns about not being able to run dispatches in parallel were not well-founded as most (if not all) IHVs will treat buffer barriers as "wait for idle" anyway.

wgpu-core/src/command/compute.rs
teoxoy: Resolving as this will no longer be relevant once I update the patch with the new implementation strategy.

wgpu-core/src/device/resource.rs
teoxoy: I had to add a flag on `BufferBinding`s or else we would have allowed users of the API to bind indirect buffers as storage buffers as well.

teoxoy:
I'm not happy about the new `allow_indirect_as_storage` & `ignore_push_constant_check` flags but I couldn't come up with a better idea.

wgpu-core/src/indirect_validation.rs
```diff
@@ -0,0 +1,423 @@
...
+        // => `binding_size = alignment * chunks + sr - alignment * chunks + alignment * u`
+        // => `binding_size = sr + alignment * u`
+        // => `min_aligned_binding_size <= sr + alignment * u`
+        // => `alignment + 8 <= sr + alignment * u`
+        // => `u` must be at least 2
+        // => `binding_size = sr + alignment * 2`
+
+        let binding_size = 2 * alignment + (buffer.size % alignment);
+        binding_size.min(buffer.size)
+    }
```
teoxoy: I hope this makes sense. It took me a while to figure out and write down what the constraints were and how we come up with a binding size that satisfies them.

teoxoy:
It seems this PR causes rustdoc on windows to time out (related https://github.com/gfx-rs/wgpu/issues/4905).

teoxoy:
We decided to disable rustdoc in CI for now. I will add a comment in https://github.com/gfx-rs/wgpu/issues/4905 saying that we should revert https://github.com/gfx-rs/wgpu/pull/5714/commits/f20d98bd7bdcf1c59e63bc33a0ca0fe20423eb19 once the issue gets fixed upstream.

teoxoy:
This was blocked because we still had a lot of internal usages of IDs and to move to a solution that uses `wgpu-core` resources, those had to go away. The reason for moving to a solution that uses `wgpu-core` resources was due to the inability of using externally generated (by Firefox for example) and internally generated IDs (by `wgpu-core`) at the same time which the initial solution required.

Now that internal usages of IDs are gone, I updated the PR to use `wgpu-core` resources but it turned out to still not be a satisfactory solution because of 2 main issues:
- user created indirect buffers had to hold an `Arc<BindGroup>` which itself contained a strong reference to the buffer causing both to leak
- limits validation was too constraining because things will get validated using the required limits passed when creating the device (I tried getting around this by adding a bunch of flags that ignore limit validation but it was too invasive)

In the end I updated the PR to just use `wgpu-hal` resources which solved both issues and feels overall more natural.

teoxoy:
There is one `TODO: reset push constants` left that I'll take care of on Monday.

teoxoy:
@cwfitzgerald now that this is using hal objects, I think it should be possible to optimize the impl by using 2 destination buffers. You mentioned previously on a call how that would work but it wasn't 100% clear to me. If you could outline it in a comment, maybe I can add it in this PR.

teoxoy:
I resolved the last TODO (https://github.com/gfx-rs/wgpu/pull/5714#issuecomment-2239517570).

teoxoy:
I implemented the double destination buffer approach and also rebased https://github.com/gfx-rs/wgpu/pull/5730 on this PR.

cwfitzgerald (commented):
I'll review this more later, but my vibe is no cfg for this - I'm in a very anti-cfg mood lately.

teoxoy:
Ok, so you are not persuaded by the numbers in https://github.com/gfx-rs/wgpu/pull/5714#issuecomment-2120913014?

cwfitzgerald:
Maybe I forgot about them.... :)

JMS55:
Can there be a way to turn this off?

teoxoy:
It's needed to get the `num_workgroups` builtin working (https://github.com/gfx-rs/wgpu/pull/5730). We could use the codepath only if `num_workgroups` is used but we'd like to see how big of a bottleneck it is first.

ErichDonGubler:
Rebased to resolve conflicts with mainline.

ErichDonGubler:
Dismissed current `Changes requested` reviews, since all feedback returned from them appears to be marked as `Resolved`.

wgpu-core/src/device/resource.rs
```diff
@@ -690,7 +735,11 @@ impl Device {
         self: &Arc<Self>,
         hal_buffer: Box<dyn hal::DynBuffer>,
         desc: &resource::BufferDescriptor,
-    ) -> Arc<Buffer> {
+    ) -> Result<Arc<Buffer>, resource::CreateBufferError> {
+        #[cfg(feature = "indirect-validation")]
+        let raw_indirect_validation_bind_group =
+            self.create_indirect_validation_bind_group(hal_buffer.as_ref(), desc.size, desc.usage)?;
+
         unsafe { self.raw().add_raw_buffer(&*hal_buffer) };
```
ErichDonGubler: @nical, was it correct to resolve a conflict here by ordering the `add_raw_buffer` counter increment after creating `raw_indirect_validation_bind_group`?

ErichDonGubler (commented):
Rebased onto `trunk` to resolve conflicts.

wgpu-core/src/pipeline.rs
```diff
@@ -92,7 +92,7 @@ impl ShaderModule {
 #[derive(Clone, Debug, Error)]
 #[non_exhaustive]
 pub enum CreateShaderModuleError {
-    #[cfg(feature = "wgsl")]
+    #[cfg(any(feature = "wgsl", feature = "indirect-validation"))]
```
ErichDonGubler, line 95: question: Would it be interesting to add a `cfg_alias` entry for this? I don't see any other usages of it, at this point, but having apparently unrelated features used here makes the footgun worth avoiding now.

wgpu-core/src/present.rs
ErichDonGubler, line 1: note: Split out the change to take `_guard` by mutable reference instead of by value in `Snatchable::snatch` into 8e6db0f8.

wgpu-core/src/command/bind.rs
```diff
@@ -200,13 +200,17 @@ mod compat {
                 entries: (0..hal::MAX_BIND_GROUPS).map(|_| Entry::empty()).collect(),
             }
         }
-        fn make_range(&self, start_index: usize) -> Range<usize> {
+
+        pub fn nr_of_valid_entries(&self) -> usize {
```
ErichDonGubler: note: I split the introduction of `BoundBindGroupLayouts::nr_of_valid_entries` out into d93deb0a.

wgpu/Cargo.toml
```diff
@@ -130,6 +130,12 @@ features = ["raw-window-handle"]
 workspace = true
 features = ["raw-window-handle"]
 
+# If we are not targeting WebGL, enable indirect-validation.
```
ErichDonGubler, line 133: nitpick: I think this covers more than just the WebGL target. I think this covers WASM _and_ Emscripten targets for WebGPU _and_ WebGL backends? Seems like something we should clarify?

wgpu-core/src/indirect_validation.rs
```diff
@@ -0,0 +1,369 @@
...
+/// This machinery requires the following limits:
+///
+/// - max_bind_groups: 2,
+/// - max_dynamic_storage_buffers_per_pipeline_layout: 1,
+/// - max_storage_buffers_per_shader_stage: 2,
+/// - max_storage_buffer_binding_size: 3 * min_storage_buffer_offset_alignment,
+/// - max_push_constant_size: 4,
+/// - max_compute_invocations_per_workgroup 1
+///
+/// Which are be all indirectly satisfied by `DownlevelFlags::INDIRECT_EXECUTION`.
```
ErichDonGubler: typo: `are be` should probably be `are`.

```suggestion
/// Which are all indirectly satisfied by `DownlevelFlags::INDIRECT_EXECUTION`.
```

ErichDonGubler: suggestion: This reference is oblique, and I think it might help the reader more if we note that it's supposed to be a _requirement_ for activating this functionality, i.e. (combined with the above typo suggestion, modulo a correct text break/reflow):

```suggestion
/// These are all indirectly satisfied by `DownlevelFlags::INDIRECT_EXECUTION`, which is also required for this module's functionality to work.
```

wgpu-core/src/indirect_validation.rs
```diff
@@ -0,0 +1,369 @@
...
+            }}
+        ");
+
+        let module = naga::front::wgsl::parse_str(&src).map_err(|inner| {
+            CreateShaderModuleError::Parsing(naga::error::ShaderError {
+                source: src.clone(),
+                label: None,
+                inner: Box::new(inner),
+            })
+        })?;
```
ErichDonGubler, line 91: thought: I wonder if (as a lofty, low-priority optimization) we could move the parsing step here to build time (maybe prepared by `build.rs`?). I don't have a reason to suspect that this module compilation is problematic, and certainly won't block this PR on account of that. I guess I just wanted to throw that out there.

wgpu-core/src/indirect_validation.rs
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    pub fn create_src_bind_group(
+        &self,
+        device: &dyn hal::DynDevice,
+        limits: &wgt::Limits,
+        buffer_size: u64,
+        buffer: &dyn hal::DynBuffer,
+    ) -> Result<Box<dyn hal::DynBindGroup>, DeviceError> {
+        let binding_size = calculate_src_buffer_binding_size(buffer_size, limits);
```
ErichDonGubler, line 287: TODO: validate this

wgpu-core/src/device/resource.rs
```diff
@@ -590,6 +629,10 @@ impl Device {
         };
         let buffer = unsafe { self.raw().create_buffer(&hal_desc) }.map_err(DeviceError::from)?;
 
+        #[cfg(feature = "indirect-validation")]
+        let raw_indirect_validation_bind_group =
+            self.create_indirect_validation_bind_group(buffer.as_ref(), desc.size, desc.usage)?;
+
```
ErichDonGubler, line 628: thought(non-blocking): It seems strange to "unconditionally" create this bind group. I'm not immediately sure how this could be improved, though; the only things I can think of to clear this up would be to effectively inline the helper.

CHANGELOG.md
```diff
@@ -97,6 +97,7 @@ By @teoxoy [#6134](https://github.com/gfx-rs/wgpu/pull/6134).
 - Fix error message that is thrown in create_render_pass to no longer say `compute_pass`. By @matthew-wong1 [#6041](https://github.com/gfx-rs/wgpu/pull/6041)
 - Add `VideoFrame` to `ExternalImageSource` enum. By @jprochazk in [#6170](https://github.com/gfx-rs/wgpu/pull/6170)
 - Document `wgpu_hal` bounds-checking promises, and adapt `wgpu_core`'s lazy initialization logic to the slightly weaker-than-expected guarantees. By @jimblandy in [#6201](https://github.com/gfx-rs/wgpu/pull/6201)
+- Ensure safety of indirect dispatch. By @teoxoy in [#5714](https://github.com/gfx-rs/wgpu/pull/5714)
```
ErichDonGubler: question: My hot take is that most users will either bounce off of this, or get spooked. Is that desired? Did we want to elaborate on why users might care about this?

ErichDonGubler (commented):
Review is still WIP for me, sadly. HOWEVER:

* I did split some commits out, and I'm feeling confident that they're truly orthogonal (which means I feel like I'm understanding the scope of individual commits!).
* I'm partway through the `ensure safety of indirect dispatch ` commit. I've got the resource validation, creation, and destruction all confirmed in my head, and I'm not concerned about the correctness of those anymore. My next step is to finish reviewing the executed commands in the middle, to make sure they all make sense to me.

praise: This looks hecka complicated, esp. the size calculation for buffers. Kudos for persevering! Hopefully your eyes didn't glaze over too much with all the algebra; I know mine would have (and probably will during review 😅).

ErichDonGubler:
CI is currently unhappy with the change of `ComputePass::set_bind_group` to be `Option`al, and some changes in error types. Not sure about the details of the latter yet, and I'm not sure if I'm going to be able to fix this until Wednesday.

ErichDonGubler:
Resolved compile issues with `set_bind_group`, and the shape of fallibility changing in `Device::create_buffer_from_hal`. There's still outstanding compilation errors with converting `CreateBufferError` to a `DeviceError`.

wgpu/Cargo.toml
teoxoy: That is true but `wgpu-core` is not used when `wgpu` targets WebGPU.

wgpu-core/src/indirect_validation.rs
teoxoy: It would be nice, it would also remove the need for the new `indirect-validation` feature flag. It's something I've never done before and wasn't sure how it would turn out. So, I can't assess it's worth right now.

wgpu-core/src/device/resource.rs
teoxoy: It can't be avoided and it's happening at buffer creation time so it shouldn't be surprising from a perf perspective.

The helper is used in 2 places (this being one of them), I'm not sure why it would be better to inline it?

CHANGELOG.md
teoxoy: I was thinking that interested people would check the PR directly and see:

> Ensure safety of indirect dispatch by injecting a compute shader that validates the content of the indirect buffer.

But maybe we can include that in the changelog directly?

ErichDonGubler: Yeah, I think the longer explanation is still short enough for the `CHANGELOG`. I'll edit the PR to include that. 👍🏻

wgpu-core/src/device/resource.rs
ErichDonGubler: Sorry, my feedback is not intended to comment on how semantics need to change; I think the code is doing the right thing. I think the way that the code is communicating ("let's 'create' something, but I under-the-hood might not actually create anything") is disorienting to somebody on both sides of a black box, but it's extremely minor. I agree, though, that I don't think there's a meaningful action to take here WRT clarifying; I can't think of something better to do.

I feel heard after expressing the above, so I'm going to `Resolve`.

wgpu-core/src/pipeline.rs
ErichDonGubler: I don't think the answer to this is critical for merging this PR, so I'm going to `Resolve` it.

wgpu-core/src/device/resource.rs
ErichDonGubler: Thinking out loud: I believe we shouldn't increment the counter after the only fallible operation that this PR introduces. If we fail, then we don't have a `Buffer` created.

wgpu-core/src/indirect_validation.rs
ErichDonGubler: I think I'll let maintainer and user pain drive this idea later, if it comes to that, then. 😅

CHANGELOG.md
ErichDonGubler: Implemented as agreed with a force-push of c851e983.

wgpu-core/src/indirect_validation.rs
ErichDonGubler: Implemented as agreed with a force-push of c851e983.

wgpu-core/src/device/resource.rs
```diff
@@ -272,6 +282,25 @@ impl Device {
...
+            .contains(wgt::DownlevelFlags::INDIRECT_EXECUTION)
+        {
+            match crate::indirect_validation::IndirectValidation::new(
+                raw_device.as_ref(),
+                &desc.required_limits,
+            ) {
+                Ok(indirect_validation) => Some(indirect_validation),
+                Err(e) => {
+                    log::error!("indirect-validation error: {e:?}");
+                    return Err(CreateDeviceError::Internal);
```
ErichDonGubler: issue: `CreateDeviceError` isn't even _used_ any more, and seems like it should have been removed as of <https://github.com/gfx-rs/wgpu/pull/6229> (CC @teoxoy as the author), so this addition of `CreateDeviceError::Internal` no longer makes sense.

issue: We have lots of other compilation errors after rebasing because we no longer `impl From<hal::DeviceError> for wgpu_core::DeviceError` (which is a good thing, also from <https://github.com/gfx-rs/wgpu/pull/6229>). We need to use `wgpu_core::DeviceError::from_hal`. This is appropriate, rather than `Device::handle_hal_error`, because the context is that we're trying to create a `Device` in the first place.

suggestion: I have resolved both of the above as follows with the force-pushed b6b2faed:

```diff
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -283,7 +283,7 @@
                 Ok(indirect_validation) => Some(indirect_validation),
                 Err(e) => {
                     log::error!("indirect-validation error: {e:?}");
-                    return Err(CreateDeviceError::Internal);
+                    return Err(DeviceError::Lost);
                 }
             }
         } else {
--- a/wgpu-core/src/indirect_validation.rs
+++ b/wgpu-core/src/indirect_validation.rs
@@ -107,7 +107,7 @@
             unsafe { device.create_shader_module(&hal_desc, hal_shader) }.map_err(|error| {
                 match error {
                     hal::ShaderError::Device(error) => {
-                        CreateShaderModuleError::Device(error.into())
+                        CreateShaderModuleError::Device(DeviceError::from_hal(error))
                     }
                     hal::ShaderError::Compilation(ref msg) => {
                         log::error!("Shader error: {}", msg);
@@ -133,7 +133,7 @@
         let dst_bind_group_layout = unsafe {
             device
                 .create_bind_group_layout(&dst_bind_group_layout_desc)
-                .map_err(DeviceError::from)?
+                .map_err(DeviceError::from_hal)?
         };
 
         let src_bind_group_layout_desc = hal::BindGroupLayoutDescriptor {
@@ -153,7 +153,7 @@
         let src_bind_group_layout = unsafe {
             device
                 .create_bind_group_layout(&src_bind_group_layout_desc)
-                .map_err(DeviceError::from)?
+                .map_err(DeviceError::from_hal)?
         };
 
         let pipeline_layout_desc = hal::PipelineLayoutDescriptor {
@@ -171,7 +171,7 @@
         let pipeline_layout = unsafe {
             device
                 .create_pipeline_layout(&pipeline_layout_desc)
-                .map_err(DeviceError::from)?
+                .map_err(DeviceError::from_hal)?
         };
 
         let pipeline_desc = hal::ComputePipelineDescriptor {
@@ -188,7 +188,7 @@
         let pipeline =
             unsafe { device.create_compute_pipeline(&pipeline_desc) }.map_err(|err| match err {
                 hal::PipelineError::Device(error) => {
-                    CreateComputePipelineError::Device(error.into())
+                    CreateComputePipelineError::Device(DeviceError::from_hal(error))
                 }
                 hal::PipelineError::Linkage(_stages, msg) => {
                     CreateComputePipelineError::Internal(msg)
@@ -208,7 +208,7 @@
             memory_flags: hal::MemoryFlags::empty(),
         };
         let dst_buffer =
-            unsafe { device.create_buffer(&dst_buffer_desc) }.map_err(DeviceError::from)?;
+            unsafe { device.create_buffer(&dst_buffer_desc) }.map_err(DeviceError::from_hal)?;
 
         let dst_bind_group_desc = hal::BindGroupDescriptor {
             label: None,
@@ -230,7 +230,7 @@
         let dst_bind_group = unsafe {
             device
                 .create_bind_group(&dst_bind_group_desc)
-                .map_err(DeviceError::from)
+                .map_err(DeviceError::from_hal)
         }?;
 
         Ok(Self {
@@ -272,7 +272,7 @@
         unsafe {
             device
                 .create_bind_group(&hal_desc)
-                .map_err(DeviceError::from)
+                .map_err(DeviceError::from_hal)
         }
     }
 
```

wgpu-core/src/indirect_validation.rs
ErichDonGubler: Ah, this was already discussed, sorry for missing it: https://github.com/gfx-rs/wgpu/pull/5714#issuecomment-2120047557

wgpu/Cargo.toml
ErichDonGubler: 🤷🏻‍♂️

wgpu-core/src/command/compute.rs
```diff
@@ -791,6 +810,11 @@ fn set_push_constant(
         end_offset_bytes,
     )?;
 
+    let offset_in_elements = (offset / wgt::PUSH_CONSTANT_ALIGNMENT) as usize;
+    let size_in_elements = (size_bytes / wgt::PUSH_CONSTANT_ALIGNMENT) as usize;
+    state.push_constants[offset_in_elements..offset_in_elements + size_in_elements]
+        .copy_from_slice(data_slice);
```
ErichDonGubler: nitpick: There's a (slightly) more brain-friendly way to write this!

```suggestion
    state.push_constants[offset_in_elements..][..size_in_elements].copy_from_slice(data_slice);

```

ErichDonGubler: Implemented as suggested with a force-push to c3d50b89.

wgpu-core/src/command/compute.rs
```diff
@@ -841,10 +864,6 @@ fn dispatch_indirect(
         .device
         .require_downlevel_flags(wgt::DownlevelFlags::INDIRECT_EXECUTION)?;
 
-    state
-        .scope
-        .buffers
-        .merge_single(&buffer, hal::BufferUses::INDIRECT)?;
```
ErichDonGubler, line 847: question: Why do we not call `merge_single` in the version of code with indirect validation, as with non-validated path? 🤔 Is that on purpose?

wgpu-core/src/command/compute.rs
```diff
@@ -870,12 +888,132 @@ fn dispatch_indirect(
...
+        let src_barrier =
+            src_transition.map(|transition| transition.into_hal(&buffer, &state.snatch_guard));
+        unsafe {
+            state.raw_encoder.transition_buffers(src_barrier.as_slice());
+        }
+
+        unsafe {
+            state.raw_encoder.transition_buffers(&[hal::BufferBarrier {
+                buffer: params.dst_buffer,
+                usage: hal::BufferUses::INDIRECT..hal::BufferUses::STORAGE_READ_WRITE,
```
ErichDonGubler, line 948: issue: This range seems invalid, maybe on purpose? `INDIRECT` is an element that is _after_ `STORAGE_READ_WRITE`. Was the intent to provide an empty range? If so, there's probably clearer way to do it (i.e., `BufferUses::empty()`). If not, theeeen oops!

ErichDonGubler (commented):
Mostly done with the `ensure safety of indirect dispatch` commit! 😭 🏁

wgpu-core/src/device/resource.rs
nical: The patch looks correct to me: if `create_indirect_validation_bind_group` fails then the `Buffer` is not added and `destroy_buffer` will not be called so the counter is balanced.

tests/tests/dispatch_workgroups_indirect.rs
```diff
@@ -0,0 +1,247 @@
...
+                max_push_constant_size: 4,
+                ..wgpu::Limits::downlevel_defaults()
+            }),
+    )
+    .run_async(|ctx| async move {
+        ctx.device.push_error_scope(wgpu::ErrorFilter::Validation);
+
+        let _ = run_test(&ctx, &[0, 0, 0], true).await;
+
+        let error = pollster::block_on(ctx.device.pop_error_scope());
```
ErichDonGubler, line 75: The `RESET_BIND_GROUPS` test fails because we handle validation errors fatally in `Queue::submit`. Because the current type signature unconditionally returns a `SubmissionIndex`, I need to change `wgpu_core`'s internals to return it even when there's an error, so we can handle it non-fatally and use the validation error sink (as is according to spec., I think?).

Tagging @cwfitzgerald, @Wumpf, and @jimblandy for validation on this approach. CC @teoxoy for 👀.

ErichDonGubler:
The test that is failing in <https://github.com/gfx-rs/wgpu/pull/5714#discussion_r1769372866> seems important, so I'm going to mark this as a draft until we can get that unblocked. I'm planning on continuing work on this during the coming week, though it will not be my highest priority until after #5320.

tests/tests/dispatch_workgroups_indirect.rs
ErichDonGubler: Tracking resolution with #6317, which is about to be resolved.

wgpu-core/src/indirect_validation.rs
```diff
@@ -0,0 +1,427 @@
...
+    // => `binding_size = alignment * chunks + sr - alignment * chunks + alignment * u`
+    // => `binding_size = sr + alignment * u`
+    // => `min_aligned_binding_size <= sr + alignment * u`
+    // => `alignment + 8 <= sr + alignment * u`
+    // => `u` must be at least 2
+    // => `binding_size = sr + alignment * 2`
+
+    let binding_size = 2 * alignment + (buffer_size % alignment);
+    binding_size.min(buffer_size)
+}
```
ErichDonGubler, line 378: issue(non-blocking): Sigh, I took more time trying to validate this than I'd like to admit. For me, it's difficult to understand, though admittedly I don't have a more refined problem statement for this logic than "we need to make sure we can fit indirect offsets into a buffer and address it via push constant, using no more memory than absolutely necessary". Assuming that @cwfitzgerald reviewed this algebra already, I'm going to have to pass on fully validating this; I don't think I can effectively add to the de-risking that existing review already has done.

As @teoxoy is out-of-office for at least another week, I'd like to confirm that @cwfitzgerald and/or @jimblandy have validated this algebra. If they have, then 👍🏻 I don't think my lack of understanding should block here.

tests/tests/dispatch_workgroups_indirect.rs
```diff
@@ -0,0 +1,247 @@
...
+            indirect_offset,
+            bytemuck::bytes_of(num_workgroups),
+        );
+
+        let mut encoder = ctx
+            .device
+            .create_command_encoder(&wgpu::CommandEncoderDescriptor::default());
+        {
+            let mut compute_pass =
+                encoder.begin_compute_pass(&wgpu::ComputePassDescriptor::default());
```
ErichDonGubler: nitpick: Any reason we don't use `Default::default` instead of explicitly naming these long-winded types?

suggestion:

```suggestion
        let mut encoder = ctx.device.create_command_encoder(&Default::default());
        {
            let mut compute_pass = encoder.begin_compute_pass(&Default::default());
```

ErichDonGubler: Implemented myself with a force-push to 3f96a622.

ErichDonGubler (approved):

wgpu-core/src/command/compute.rs
ErichDonGubler: This looks an inverted version of a similar call to `transition_buffers` that uses 	`usage: hal::BufferUses::STORAGE_READ_WRITE..hal::BufferUses::INDIRECT`.

tests/tests/dispatch_workgroups_indirect.rs
```diff
@@ -0,0 +1,244 @@
...
+
+        // `test_offset.inner` should always be 0; we test that resetting the push constant set by the validation code works properly.
+        var<push_constant> test_offset: TestOffsetPc;
+
+        @group(0) @binding(0)
+        var<storage, read_write> out: array<u32, 3>;
+
+        @compute @workgroup_size(1)
+        fn main(@builtin(num_workgroups) num_workgroups: vec3<u32>, @builtin(workgroup_id) workgroup_id: vec3<u32>) {
+            if (all(workgroup_id == vec3<u32>())) {
```
cwfitzgerald:
```suggestion
            if (all(workgroup_id == vec3u(0))) {
```

Just for readability

wgpu-core/src/command/bind.rs
cwfitzgerald:
```suggestion
        pub fn number_of_valid_entries(&self) -> usize {
```

`num` is also okay

wgpu-core/src/indirect_validation.rs
```diff
@@ -0,0 +1,427 @@
...
+            var<storage, read> src: array<u32>;
+            struct OffsetPc {{
+                inner: u32,
+            }}
+            var<push_constant> offset: OffsetPc;
+
+            @compute @workgroup_size(1)
+            fn main() {{
+                let src = vec3(src[offset.inner], src[offset.inner + 1], src[offset.inner + 2]);
+                let res = select(src, vec3<u32>(), src > vec3({max_compute_workgroups_per_dimension}u));
```
cwfitzgerald: This feels weird. We end up zeroing out only the dimensions that go over the limit. I guess this would result in zero invocations, but I kind of expect this to either be `src` or `res` in its entirety.

cwfitzgerald: Also does Mac allow you to issue an indirect dispatch of zero size? I know they have problems with direct dispatches.

tests/tests/dispatch_workgroups_indirect.rs
```diff
@@ -0,0 +1,244 @@
...
+
+        // `test_offset.inner` should always be 0; we test that resetting the push constant set by the validation code works properly.
+        var<push_constant> test_offset: TestOffsetPc;
+
+        @group(0) @binding(0)
+        var<storage, read_write> out: array<u32, 3>;
+
+        @compute @workgroup_size(1)
+        fn main(@builtin(num_workgroups) num_workgroups: vec3u, @builtin(workgroup_id) workgroup_id: vec3u) {
+            if (all(workgroup_id == vec3u())) {
```
ErichDonGubler, line 99: nitpick: @cwfitzgerald and I agreed that `s/vec3<u32>/vec3u` was nice.

Already pushed this up as a `fixup!` with 431ca4a0.

wgpu-core/src/command/compute.rs
ErichDonGubler: @cwfitzgerald explained to me that we're subverting the `Range` type to be a data structure of a "from" and "to" pair, rather than a range of values. That explains my misunderstanding!

I…think we should split `usage` into `*from` and `*to` pairs, because I would expect a similar confusion from anybody else with a background in idiomatic Rust. 😅 That need not block here, but it'd be a nice papercut to remove for later.

wgpu-core/src/command/compute.rs
ErichDonGubler: @cwfitzgerald explained to me that we're using several different layers of usage tracking here, and that we needed a different path when issuing these encoder commands under the hood. 👍🏻 I think we're good to go here?

wgpu-core/src/indirect_validation.rs
ErichDonGubler: @cwfitzgerald and I agreed to let @teoxoy deal with fixing this algebra if it's wrong. 😅👍🏻

wgpu-core/src/indirect_validation.rs
```diff
@@ -0,0 +1,429 @@
...
+    module: Box<dyn hal::DynShaderModule>,
+    dst_bind_group_layout: Box<dyn hal::DynBindGroupLayout>,
+    src_bind_group_layout: Box<dyn hal::DynBindGroupLayout>,
+    pipeline_layout: Box<dyn hal::DynPipelineLayout>,
+    pipeline: Box<dyn hal::DynComputePipeline>,
+    dst_buffer_0: Box<dyn hal::DynBuffer>,
+    dst_buffer_1: Box<dyn hal::DynBuffer>,
+    dst_bind_group_0: Box<dyn hal::DynBindGroup>,
+    dst_bind_group_1: Box<dyn hal::DynBindGroup>,
+    is_next_dst_0: AtomicBool,
```
ErichDonGubler: issue: @cwfitzgerald was concerned (and now I am, too) that this usage of `AtomicBool` is not correct for multi-threaded usage. Encoding of commands can happen on multiple different threads, and the swapping done here is greatly complicated by the fact that the commands that necessitate swapping may not be submitted in the order this boolean attempts to enforce.

ErichDonGubler: suggestion: Perhaps we can leave off the two-buffer optimization and the accompanying synchronization baggage for a follow-up PR/issue?

cwfitzgerald (dismissed):
See [Atomic issue](https://github.com/gfx-rs/wgpu/pull/5714#discussion_r1777618706) and other small nits

EDIT from @ErichDonGubler: added a link for atomic issue

tests/tests/dispatch_workgroups_indirect.rs
ErichDonGubler: Resolved in <https://github.com/gfx-rs/wgpu/pull/5714#discussion_r1777601203>.

wgpu-core/src/command/bind.rs
ErichDonGubler: Renamed to `num_valid_entries` with a force-push; relevant commit hashes are fcc254db and 5b62eea9.

wgpu-core/src/indirect_validation.rs
ErichDonGubler: At least on iOS, this seems to be problematic: https://github.com/linebender/vello/issues/637#issuecomment-2242428625

wgpu-core/src/device/resource.rs
teoxoy: Thanks!

wgpu-core/src/indirect_validation.rs
teoxoy: It did take me a while to figure all this out as well (https://github.com/gfx-rs/wgpu/pull/5714#discussion_r1635467462) and then even more to try to explain it but it seems it's still quite hard to explain (the problem being that it's not a simple equation) - I honestly don't know what kind/branch of mathematics this would go in.

wgpu-core/src/indirect_validation.rs
teoxoy: Yeah, implementing this optimization sounds more complicated than I thought. I think it would be a problem even in non multithreaded use-cases because even on a single thread users can create multiple command buffers that then get submitted in a different order (than the order they have been created in).

I will drop the commit and create a new issue to explore this further.

wgpu-core/src/indirect_validation.rs
teoxoy: > Also does Mac allow you to issue an indirect dispatch of zero size? I know they have problems with direct dispatches.

From my testing it does work (the new test also checks that this works).

https://github.com/linebender/vello/issues/637 is about direct dispatches.

teoxoy: > This feels weird. We end up zeroing out only the dimensions that go over the limit. I guess this would result in zero invocations, but I kind of expect this to either be `src` or `res` in its entirety.

I will update the code so that `res` = either `src` or `0 0 0`.

teoxoy:
I don't know why github it not letting me "rebase and merge":

image

Via the CLI is also not working.

```
gh pr merge 5714 -r -R gfx-rs/wgpu
GraphQL: This branch can't be rebased (mergePullRequest)
```

teoxoy:
I will just squash and merge, losing https://github.com/gfx-rs/wgpu/pull/5714/commits/6619f3a20b970f42bba52c04b0f13d6127a4cad8 is not the end of the world.

schell:
I'm tracking `trunk` in my project and just hit a panic in these changes: 
```
thread 'test::unlit_textured_cube_material' panicked at /Users/schell/.cargo/git/checkouts/wgpu-53e70f8674b08dd4/59f56e0/wgpu-core/src/indirect_validation.rs:288:58:
called `Option::unwrap()` on a `None` value
```

I'm guessing it's this binding `binding_size` on line 276 in the file mentioned:
```
        let binding_size = calculate_src_buffer_binding_size(buffer_size, limits);
```

being used on line 288:
```
                size: Some(NonZeroU64::new(binding_size).unwrap()),
```

...so for some reason in my case `binding_size` is `0`. Maybe the buffer passed is `0`? 🤷

vntec:
Any plans to allow opting out of this? Maybe by disabling a flag called "extensive validation" or something like that?
Don't get me wrong, I generally support the idea of more security checks like this (especially for WebGPU), but I feel like injecting a compute dispatch before every single indirect call could be a bit too intrusive for some release builds.

If that's not planned, I'd at least suggest adding an appropriate label to the compute pass? I was initially very confused about an unnamed dispatch showing up in RenderDoc :D.

JMS55:
I actually checked my renderer, and wasn't seeing any indirect validation dispatches after upgrading to wgpu 23. Not sure why.

vntec:
> I actually checked my renderer, and wasn't seeing any indirect validation dispatches after upgrading to wgpu 23. Not sure why.

That's odd. I definitely encountered these. After upgrading, my renderer stopped working and I noticed that I had been exceeding the 65k `max_compute_workgroups_per_dimension` limit quite frequently, which went unnoticed since my Vulkan driver supports 2 billion workgroups in the x-dimension.

So the feature is definitely useful for development and debugging, but I'd appreciate the option to disable it for some release builds. The performance impact seems quite small, but it may add up for indirect-heavy applications on low-end hardware.

teoxoy:
We will probably add a flag to disable this, see https://github.com/gfx-rs/wgpu/issues/6567. Though, the `num_workgroups` builtin won't work on D3D12 (https://github.com/gfx-rs/wgpu/pull/5730) without this.

vntec:
That's great to hear, thanks!

PR #5724, opened 2024-05-21
Buffer as hal

JasondeWolff:
**Connections**
It would be really nice to have access to wgpu buffers inside more low-level wgpu-hall passes as a user. This functionality is already exposed for the majority of structs where it's applicable, not exposing it for `Buffer` is confusing.

**Description**
Buffers created with wgpu cannot be accessed using wgpu-hal.

**Testing**
Was tested on my machine by writing and accessing into wgpu created buffers using the wgpu-hal bindgroups and command encoders.

**Checklist**

- [X] Run `cargo fmt`.
- [X] Run `cargo clippy`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,7 +78,7 @@ By @stefnotch in [#5410](https://github.com/gfx-rs/wgpu/pull/5410)
 
 #### General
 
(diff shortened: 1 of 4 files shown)


Wumpf (approved):
lgtm! thanks for filling out that gap

Issue #2471, opened 2022-02-10
[D3D12] Fix `vertex_index` and `instance_index` built-in inputs for indirect draw calls


FredrikNoren:
**Description**
`DrawIndexedIndirect.base_instance` is always 0 in the shader when using `draw_indexed_indirect` on DX12, regardless of the value provided. This works on Vulkan and Metal. Setting the `INDIRECT_FIRST_INSTANCE` flag doesn't help.

**Platform**
wgpu 0.12.0, Windows DX12

(Filed this issue by request from @kvark from a discussion in the matrix chat)

kvark:
Doesn't look like Dawn has this implemented either - https://bugs.chromium.org/p/dawn/issues/detail?id=548&q=548&can=2

jimblandy:
Dawn has fixed that bug.

teoxoy:
We should remember to also remove `DownlevelFlags::VERTEX_AND_INSTANCE_INDEX_RESPECTS_RESPECTIVE_FIRST_VALUE_IN_INDIRECT_DRAW` (added recently in https://github.com/gfx-rs/wgpu/pull/4723) once this is resolved.

mikialex:
Dx12 non-zero base_instance actually works but the result is just not exposed in the shader. Before this issue get fixed, one way to workaround is to manually generate(by compute shader) a base_instance buffer and bind it as an instance vertex buffer.

teoxoy:
This has a lower priority than https://github.com/gfx-rs/wgpu/issues/2431 but I will work on it right after draw indirect validation since it will be easier to do while I still have that context.

PR #5730, opened 2024-05-22
[d3d12] get `num_workgroups` builtin working for indirect dispatches

teoxoy:
**Connections**
Depends on https://github.com/gfx-rs/wgpu/pull/5714.
Follow-up to https://github.com/gfx-rs/wgpu/pull/1816.

**Description**
`num_workgroups` used to always be `0, 0, 0` for indirect dispatches; it now works properly.
This PR adds the logic needed to update the root constants with the right values for the `num_workgroups` builtin.

**Testing**
There is a test.

### DIFF
--- a/wgpu-core/src/indirect_validation.rs
+++ b/wgpu-core/src/indirect_validation.rs
@@ -1,3 +1,6 @@
+use std::mem::size_of;
+use std::num::NonZeroU64;
+
 use thiserror::Error;
 
 use crate::{
@@ -57,7 +60,7 @@ impl IndirectValidation {
         let src = format!(
             "
             @group(0) @binding(0)
-            var<storage, read_write> dst: array<u32, 3>;
+            var<storage, read_write> dst: array<u32, 6>;
             @group(1) @binding(0)
             var<storage, read> src: array<u32>;
             struct OffsetPc {{
@@ -74,13 +77,24 @@ impl IndirectValidation {
                     src.y > max_compute_workgroups_per_dimension ||
                     src.z > max_compute_workgroups_per_dimension
                 ) {{
-                    dst = array(0u, 0u, 0u);
+                    dst = array(0u, 0u, 0u, 0u, 0u, 0u);
                 }} else {{
-                    dst = array(src.x, src.y, src.z);
+                    dst = array(src.x, src.y, src.z, src.x, src.y, src.z);
                 }}
             }}
         "
         );
+
+        // SAFETY: The value we are passing to `new_unchecked` is not zero, so this is safe.
+        const SRC_BUFFER_SIZE: NonZeroU64 =
+            unsafe { NonZeroU64::new_unchecked(size_of::<u32>() as u64 * 3) };
+
+        // SAFETY: The value we are passing to `new_unchecked` is not zero, so this is safe.
+        const DST_BUFFER_SIZE: NonZeroU64 = unsafe {
+            NonZeroU64::new_unchecked(
+                SRC_BUFFER_SIZE.get() * 2, // From above: `dst: array<u32, 6>`
+            )
+        };
 
         let module = naga::front::wgsl::parse_str(&src).map_err(|inner| {
             CreateShaderModuleError::Parsing(naga::error::ShaderError {
@@ -133,7 +147,7 @@ impl IndirectValidation {
                 ty: wgt::BindingType::Buffer {
                     ty: wgt::BufferBindingType::Storage { read_only: false },
                     has_dynamic_offset: false,
-                    min_binding_size: Some(std::num::NonZeroU64::new(4 * 3).unwrap()),
+                    min_binding_size: Some(DST_BUFFER_SIZE),
                 },
                 count: None,
             }],
@@ -153,7 +167,7 @@ impl IndirectValidation {
                 ty: wgt::BindingType::Buffer {
                     ty: wgt::BufferBindingType::Storage { read_only: true },
                     has_dynamic_offset: true,
-                    min_binding_size: Some(std::num::NonZeroU64::new(4 * 3).unwrap()),
+                    min_binding_size: Some(SRC_BUFFER_SIZE),
                 },
                 count: None,
             }],
@@ -211,7 +225,7 @@ impl IndirectValidation {
 
         let dst_buffer_desc = hal::BufferDescriptor {
             label: None,
-            size: 4 * 3,
+            size: DST_BUFFER_SIZE.get(),
             usage: hal::BufferUses::INDIRECT | hal::BufferUses::STORAGE_READ_WRITE,
             memory_flags: hal::MemoryFlags::empty(),
         };
@@ -229,7 +243,7 @@ impl IndirectValidation {
             buffers: &[hal::BufferBinding {
                 buffer: dst_buffer.as_ref(),
                 offset: 0,
-                size: Some(std::num::NonZeroU64::new(4 * 3).unwrap()),
+                size: Some(DST_BUFFER_SIZE),
             }],
             samplers: &[],
             textures: &[],
@@ -271,7 +285,7 @@ impl IndirectValidation {
             buffers: &[hal::BufferBinding {
                 buffer,
                 offset: 0,
-                size: Some(std::num::NonZeroU64::new(binding_size).unwrap()),
+                size: Some(NonZeroU64::new(binding_size).unwrap()),
             }],
             samplers: &[],
             textures: &[],
--- a/wgpu-hal/src/dx12/command.rs
+++ b/wgpu-hal/src/dx12/command.rs
@@ -1210,11 +1210,17 @@ impl crate::CommandEncoder for super::CommandEncoder {
     }
 
     unsafe fn dispatch_indirect(&mut self, buffer: &super::Buffer, offset: wgt::BufferAddress) {
-        self.prepare_dispatch([0; 3]);
-        //TODO: update special constants indirectly
+        self.update_root_elements();
+        let cmd_signature = &self
+            .pass
+            .layout
+            .special_constants_cmd_signatures
+            .as_ref()
+            .unwrap_or_else(|| &self.shared.cmd_signatures)
+            .dispatch;
         unsafe {
             self.list.as_ref().unwrap().ExecuteIndirect(
-                &self.shared.cmd_signatures.dispatch,
+                cmd_signature,
                 1,
                 &buffer.resource,
                 offset,
--- a/wgpu-hal/src/dx12/command.rs
+++ b/wgpu-hal/src/dx12/command.rs
@@ -107,7 +107,13 @@ impl super::CommandEncoder {
                 );
             }
         }
-        if let Some(root_index) = self.pass.layout.special_constants_root_index {
+        if let Some(root_index) = self
+            .pass
+            .layout
+            .special_constants
+            .as_ref()
+            .map(|sc| sc.root_index)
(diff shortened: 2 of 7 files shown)


cwfitzgerald:
Moving to draft until dependents land

cwfitzgerald:
Isn't this still waiting for dependents?

teoxoy:
It is but I feel like keeping it as draft gives the wrong idea, it is ready to be reviewed.

wgpu-core/src/indirect_validation.rs
```diff
@@ -61,7 +61,7 @@ impl IndirectValidation {
 
         let src = format!("
```
ErichDonGubler: nitpick: It'd be nice to avoid the leading spaces in the literal with something like [`indoc`](https://docs.rs/indoc/2.0.5/indoc/index.html), or diminishing the indentation of the literal. I don't know if it actually makes a difference here, though.

wgpu-core/src/indirect_validation.rs
```diff
@@ -121,6 +124,8 @@ impl IndirectValidation {
                 }
             })?;
 
+        const DST_BUFFER_SIZE: NonZeroU64 = unsafe { NonZeroU64::new_unchecked(4 * 3 * 2) };
```
ErichDonGubler: nitpick: It'd be nice to have a comment to help a reader to understand how this value was computed.

ErichDonGubler: thought: It's a shame we can't use `NonZeroU64::new(…).unwrap()` here, and have to reach for `unsafe`, because it's not `const`. It should be safe and optimized out otherwise.

suggestion(blocking): We should add a `SAFETY` comment here, i.e., ``The value we are passing to `new_unchecked` is not zero, so this is safe.``

wgpu-hal/src/dx12/command.rs
```diff
@@ -1198,11 +1198,17 @@ impl crate::CommandEncoder for super::CommandEncoder {
...
-        self.prepare_dispatch([0; 3]);
-        //TODO: update special constants indirectly
+        self.update_root_elements();
+        let cmd_signature = if let Some(cmd_signatures) =
+            self.pass.layout.special_constants_cmd_signatures.as_ref()
+        {
+            &cmd_signatures.dispatch
+        } else {
+            &self.shared.cmd_signatures.dispatch
+        };
```
ErichDonGubler: style: I think it's (marginally) clearer to read the `cmd_signature` extraction this way:

```suggestion
        let cmd_signature = &self
            .pass
            .layout
            .special_constants_cmd_signatures
            .as_ref()
            .unwrap_or_else(|| &self.shared.cmd_signatures)
            .dispatch;
```

I can push this up as a `fixup!`, if you want.

ErichDonGubler:
Rebased to resolve conflicts with mainline.

wgpu-hal/src/dx12/device.rs
```diff
@@ -1111,6 +1115,63 @@ impl crate::Device for super::Device {
...
+                            DestOffsetIn32BitValues: 0,
+                            Num32BitValuesToSet: 3,
+                        },
+                    },
+                };
+                Some(super::CommandSignatures {
+                    draw: Self::create_command_signature(
+                        &self.raw,
+                        Some(&raw),
+                        12 + mem::size_of::<wgt::DrawIndirectArgs>(),
```
ErichDonGubler: question: Could we factor out the `12` in the `12 + size_of<…>()` calculation being done with these command signature creations? That might actually make it amenable to an expression and explanation, which I'd also prefer here.

wgpu-core/src/indirect_validation.rs
ErichDonGubler: suggestion: It'd be better to write this expression in the same terms as the shader it's supposed to correspond to, i.e., `size_of::<u32>() as u64 * 6`.

ErichDonGubler: suggestion: We should hoist `DST_BUFFER_SIZE` up to be closer to its sister definition in the shader.

ErichDonGubler: Combined suggestions, which I can push myself if you accept:

```diff
diff --git a/wgpu-core/src/indirect_validation.rs b/wgpu-core/src/indirect_validation.rs
index 51116b58c3..6b5d39ce4d 100644
--- a/wgpu-core/src/indirect_validation.rs
+++ b/wgpu-core/src/indirect_validation.rs
@@ -1,4 +1,4 @@
-use std::{num::NonZeroU64, sync::atomic::AtomicBool};
+use std::{mem::size_of, num::NonZeroU64, sync::atomic::AtomicBool};
 
 use thiserror::Error;
 
@@ -82,6 +82,13 @@
             }}
         ");
 
+        // SAFETY: The value we are passing to `new_unchecked` is not zero, so this is safe.
+        const DST_BUFFER_SIZE: NonZeroU64 = unsafe {
+            NonZeroU64::new_unchecked(
+                size_of::<u32>() as u64 * 6, // From above: `dst: array<u32, 6>``
+            )
+        };
+
         let module = naga::front::wgsl::parse_str(&src).map_err(|inner| {
             CreateShaderModuleError::Parsing(naga::error::ShaderError {
                 source: src.clone(),
@@ -124,8 +131,6 @@
                 }
             })?;
 
-        const DST_BUFFER_SIZE: NonZeroU64 = unsafe { NonZeroU64::new_unchecked(4 * 3 * 2) };
-
         let dst_bind_group_layout_desc = hal::BindGroupLayoutDescriptor {
             label: None,
             flags: hal::BindGroupLayoutFlags::empty(),
```

ErichDonGubler: praise: All this nitpicking made possible because you made this _much_ more accessible code by factoring out a constant. 😅 Thanks!

wgpu-hal/src/dx12/mod.rs
```diff
@@ -896,6 +898,7 @@ struct PipelineLayoutShared {
     signature: Option<Direct3D12::ID3D12RootSignature>,
     total_root_elements: RootIndex,
     special_constants_root_index: Option<RootIndex>,
+    special_constants_cmd_signatures: Option<CommandSignatures>,
```
ErichDonGubler: suggestion: It seems that `special_constants_*` members of this field are all or nothing, contrary to the individual `Option`ality for each. I think it'd be worth merging these behind a single `Option`, probably as a `struct`.

I have a `fixup!` ready to go for this.

wgpu-hal/src/dx12/device.rs
ErichDonGubler: question: What is the `12` supposed to correspond to? I'm not familiar enough with this code to tell quickly. 💦

wgpu-core/src/indirect_validation.rs
```diff
@@ -150,7 +155,7 @@ impl IndirectValidation {
                 ty: wgt::BindingType::Buffer {
                     ty: wgt::BufferBindingType::Storage { read_only: true },
                     has_dynamic_offset: true,
-                    min_binding_size: Some(std::num::NonZeroU64::new(4 * 3).unwrap()),
+                    min_binding_size: Some(NonZeroU64::new(4 * 3).unwrap()),
```
ErichDonGubler: suggestion: It's jarring to see `DST_BUFFER_SIZE` so nicely modeled above, and then see this constant. I think we should extract this constant—despite only being used in this one place—so that it's easy to keep track of `src` and `dst` buffer sizes in the same place, which is what I'd expect for this code. Concretely, I'd offer the name `SRC_BUFFER_SIZE`.

ErichDonGubler (commented):
LGTM so far, just have a question about how `special_constants_cmd_signatures` is calculated. If I can get that answered, I'd feel comfortable approving; most everything else is just suggested polish/hot take "but y tho" that could wait until another PR, if necessary.

wgpu-core/src/indirect_validation.rs
teoxoy: I don't think it matters that there is extra whitespace, we give this shader to naga to translate it.

wgpu-hal/src/dx12/command.rs
teoxoy: This does look better 👍

wgpu-core/src/indirect_validation.rs
teoxoy: Looks good, feel free to push it. Thanks!

wgpu-hal/src/dx12/mod.rs
teoxoy: I remember thinking this too, I think why I didn't do it is because I was planning to remove `PipelineLayoutFlags` altogether. In our usage these fields are always `Some(x)`.

wgpu-core/src/indirect_validation.rs
teoxoy: I'm open to having a `SRC_BUFFER_SIZE`, `DST_BUFFER_SIZE` is actually `2 * SRC_BUFFER_SIZE`.

wgpu-hal/src/dx12/device.rs
teoxoy: It's the size of the 3 fields of `RootElement::SpecialConstantBuffer`. Confusingly each field has a dual purpose but this is not documented; it's only apparent when you look at how they are used. For render pipelines, it's: `first_vertex`, `first_instance` and `empty`. For compute pipelines, it's: the `x`, `y` and `z` of the dispatch (corresponds to the `num_workgroups` WGSL builtin).

Note that we don't yet use the `draw` and `draw_indexed` `special_constants_cmd_signatures` (https://github.com/gfx-rs/wgpu/issues/2471 tracks this).

wgpu-core/src/indirect_validation.rs
ErichDonGubler: Pushed as afc11af2.

wgpu-hal/src/dx12/command.rs
ErichDonGubler: Pushed as 8e60dcb9.

wgpu-core/src/indirect_validation.rs
ErichDonGubler: Pushed a `fixup!` with e7ceae48.

wgpu-hal/src/dx12/mod.rs
ErichDonGubler: Would you like the change to get pushed in here, then? If you have concrete plans to still remove it, then 🤷🏻‍♂️ whatevs.

wgpu-hal/src/dx12/device.rs
ErichDonGubler: My attempt at documenting this in a hacky, but effective, way: c240b903

Also, my attempt at documenting what you just stated about those fields: 41a07d93

Pushed 'em, and I'll let you reject 'em if you want.

wgpu-hal/src/dx12/mod.rs
ErichDonGubler: Pushed, will let you reject if you don't want it: 00b542f8

ErichDonGubler (approved):

wgpu-hal/src/dx12/device.rs
ErichDonGubler: D'oh, fixed the build breakage I introduced in c240b90 because of a refutable pattern with 41a07d934. 😅

PR #5732, opened 2024-05-23
Make `DeviceLostReason` serializable and deserializable.

sagudev:
I'm not sure how is decided which types should be (de)serializable, but we want this in servo.

**Checklist**

- [x] Run `cargo fmt`.


Wumpf (approved):
👍 

> I'm not sure how is decided which types should be (de)serializable, 

It's mostly driven by the `trace`/`replay` feature and user requests.

PR #5744, opened 2024-05-25
Fix missing family check flag

VladasZ:
**Description**
iOS 12 doesn't have `supports_family` method and it crashes when this method is called. There was already a check for that using `family_check` flag. But it was missing in one place.

Xcode crash logs:
```
2024-05-25 20:52:31.043060+0300 TestGame[355:12589] [DYMTLInitPlatform] platform initialization successful
2024-05-25 20:52:31.154605+0300 TestGame[355:12422] Metal GPU Frame Capture Enabled
2024-05-25 20:52:31.157163+0300 TestGame[355:12422] Metal API Validation Enabled
2024-05-25 20:52:31.194143+0300 TestGame[355:12422] -[MTLDebugDevice supportsFamily:]: unrecognized selector sent to instance 0x102209200
2024-05-25 20:52:31.201718+0300 TestGame[355:12422] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[MTLDebugDevice supportsFamily:]: unrecognized selector sent to instance 0x102209200'
```

**Testing**
I ran it on my iPhone 5s and it didn't crash.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,10 @@ Bottom level categories:
 - Clean up weak references to texture views and bind groups to prevent memory leaks. By @xiaopengli89 in [#5595](https://github.com/gfx-rs/wgpu/pull/5595).
 - Fix segfault on exit is queue & device are dropped before surface. By @sagudev in [#5640](https://github.com/gfx-rs/wgpu/pull/5640).
 
+#### Metal
+
+- Fix unrecognized selector crash on iOS 12. By @vladasz in [#5744](https://github.com/gfx-rs/wgpu/pull/5744).
+
 #### Vulkan
 
 - Fix enablement of subgroup ops extension on Vulkan devices that don't support Vulkan 1.3. By @cwfitzgerald in [#5624](https://github.com/gfx-rs/wgpu/pull/5624).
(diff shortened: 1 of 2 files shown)


Wumpf (approved):
Good catch!
Wondering though why this line here isn't an issue? https://github.com/gfx-rs/wgpu/blob/trunk/wgpu-hal/src/metal/adapter.rs#L515 It seems that in order to determine `family_check` we first do a family check..?

VladasZ:
~~That is weird. Documentation says that `iOS 13.0+` is minimal supported version:~~
~~https://developer.apple.com/documentation/metal/mtldevice/3143473-supportsfamily .~~
~~Maybe actually the method does exist but `MTLGPUFamily` enum ends somewhere between `Apple2` and `Apple4~~
~~on iOS 12?~~

`version.major < 8` is false so `supports_family` is never called in this expression.

Friz64:
This only landed in the `v0.20` branch, was this intentional?

Wumpf:
oh shoot, I didn't pay enough attention there. Thank you so much @Friz64 for pointing this out!

@VladasZ we usually first push to `trunk` and then cherry-pick back for `main`. As is, this fix would get lost with the next major release. Can you please create a PR against trunk?

(luckily this patch would be perfectly fine for backporting, so it's alright to release it with 0.20.1)

VladasZ:
@Wumpf sorry. Didn't know that. I tried to use `trunk` at first but it had some breaking changes and I didn't want to use it. So I fixed it in `v0.20`. 
Here is the new PR: https://github.com/gfx-rs/wgpu/pull/5754

PR #5745, opened 2024-05-26
Prefer `default-features` instead of `default_features`

waywardmonkeys:
**Description**
The latter is being deprecated and not allowed in Rust 2024 Edition.

This fixes a warning when doing a build with a nightly compiler.

**Checklist**

- [x] Run `cargo fmt`.


cwfitzgerald (approved):
Neat!

PR #5747, opened 2024-05-26
chore: Fix `legacy_numeric_constants` lints

waywardmonkeys:
**Description**
These are being deprecated in the future in favor of the associated constants (which are already being used in some code here), so this consistently uses the preferred forms.

**Checklist**

- [x] Run `cargo fmt`.


Wumpf (approved):
thanks for the cleanup!

PR #5752, opened 2024-05-29
Remove the `link` Cargo feature

madsmtm:
**Connections**
Part of https://github.com/gfx-rs/wgpu/pull/5641, split out to make it so that that PR is a non-breaking change.

**Description**
Remove the `link` feature which is not available in the `objc2` crates. This was added by `@crowlKats` in https://github.com/gfx-rs/wgpu/pull/3853, but Deno doesn't actually use it, [they instead specify this using `-weak_framework`](https://github.com/denoland/deno/blob/8a636d0600dc7000d1e12b2fc4f4f46ecd70164a/.cargo/config.toml#L14-L24), which is the correct solution to this problem IMO.

**Testing**
Roughly as described in the changelog entry.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,15 @@ By @stefnotch in [#5410](https://github.com/gfx-rs/wgpu/pull/5410)
 
 - Avoid introducing spurious features for optional dependencies. By @bjorn3 in [#5691](https://github.com/gfx-rs/wgpu/pull/5691)
 
+#### Metal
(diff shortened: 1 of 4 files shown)


cwfitzgerald:
Code's fine, just want crowlKats to give a +1 - @crowlKats merge if this is fine, please

cwfitzgerald (approved):

crowlKats (approved):

PR #5754, opened 2024-05-29
Fix missing family check flag

VladasZ:
Merge: https://github.com/gfx-rs/wgpu/pull/5744 to `trunk` branch


cwfitzgerald (approved):
In the future the change should be cherry-picked/rebased, instead of merged across branches. in this case the diff is fine, so once we squash it, it'll be fine.

PR #5768, opened 2024-06-02
Reintroduce computepass->encoder lifetime constraint and make it opt-out via `wgpu::ComputePass::forget_lifetime`

Wumpf:
**Connections**
* Part of the wider story of #1453
* Follow-up to #5620

**Description**
As suggested on Matrix chat by @kpreid.

In  #5620, `wgpu::ComputePass` lost its explicit lifetime constraint to `wgpu::CommandEncoder`. This constraint prevented use of the command encoder while a pass is recorded. This is still an error, but now a runtime error. While removing this lifetime is very useful especially for library authors, it also prevented trivial incorrect api usage, i.e. that PR was great but also created a massive footgun ;-).

Here, we remedy this issue by re-introducing the exact same lifetime constraint but making it opt-out with a zero-overhead `forget_lifetime` method.

**Testing**
Existing tests have been adjusted - compilation failed as expected without these adjustments!

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,13 +47,18 @@ TODO(wumpf): This is still work in progress. Should write a bit more about it. A
 
 `wgpu::ComputePass` recording methods (e.g. `wgpu::ComputePass:set_render_pipeline`) no longer impose a lifetime constraint passed in resources.
 
-Furthermore, `wgpu::ComputePass` no longer has a life time dependency on its parent `wgpu::CommandEncoder`.
+Furthermore, you can now opt out of `wgpu::ComputePass`'s life time dependency on its parent `wgpu::CommandEncoder` using `wgpu::ComputePass::make_static`:
+```rust
+fn independent_cpass<'enc>(encoder: &'enc mut wgpu::CommandEncoder) -> wgpu::ComputePass<'static> {
+    let cpass: wgpu::ComputePass<'enc> = encoder.begin_compute_pass(&wgpu::ComputePassDescriptor::default());
+    cpass.make_static()
+}
+```
 ⚠️ As long as a `wgpu::ComputePass` is pending for a given `wgpu::CommandEncoder`, creation of a compute or render pass is an error and invalidates the `wgpu::CommandEncoder`.
-Previously, this was statically enforced by a lifetime constraint.
-TODO(wumpf): There was some discussion on whether to make this life time constraint opt-in or opt-out (entirely on `wgpu` side, no changes to `wgpu-core`).
-Lifting this lifetime dependencies is very useful for library authors, but opens up an easy way for incorrect use.
+This is very useful for library authors, but opens up an easy way for incorrect use, so use with care.
+`make_static` is low overhead and has no side effects on pass recording.
 
-By @wumpf in [#5569](https://github.com/gfx-rs/wgpu/pull/5569), [#5575](https://github.com/gfx-rs/wgpu/pull/5575), [#5620](https://github.com/gfx-rs/wgpu/pull/5620).
+By @wumpf in [#5569](https://github.com/gfx-rs/wgpu/pull/5569), [#5575](https://github.com/gfx-rs/wgpu/pull/5575), [#5620](https://github.com/gfx-rs/wgpu/pull/5620), [#5768](https://github.com/gfx-rs/wgpu/pull/5768)
 
 #### Querying shader compilation errors
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,18 +47,18 @@ TODO(wumpf): This is still work in progress. Should write a bit more about it. A
 
 `wgpu::ComputePass` recording methods (e.g. `wgpu::ComputePass:set_render_pipeline`) no longer impose a lifetime constraint passed in resources.
 
-Furthermore, you can now opt out of `wgpu::ComputePass`'s lifetime dependency on its parent `wgpu::CommandEncoder` using `wgpu::ComputePass::make_static`:  
+Furthermore, you can now opt out of `wgpu::ComputePass`'s lifetime dependency on its parent `wgpu::CommandEncoder` using `wgpu::ComputePass::forget_lifetime`:  
 ```rust
 fn independent_cpass<'enc>(encoder: &'enc mut wgpu::CommandEncoder) -> wgpu::ComputePass<'static> {
     let cpass: wgpu::ComputePass<'enc> = encoder.begin_compute_pass(&wgpu::ComputePassDescriptor::default());
-    cpass.make_static()
+    cpass.forget_lifetime()
 }
 ```
 ⚠️ As long as a `wgpu::ComputePass` is pending for a given `wgpu::CommandEncoder`, creation of a compute or render pass is an error and invalidates the `wgpu::CommandEncoder`.
 This is very useful for library authors, but opens up an easy way for incorrect use, so use with care.
-`make_static` is low overhead and has no side effects on pass recording.
+`forget_lifetime` is zero overhead and has no side effects on pass recording.
 
-By @wumpf in [#5569](https://github.com/gfx-rs/wgpu/pull/5569), [#5575](https://github.com/gfx-rs/wgpu/pull/5575), [#5620](https://github.com/gfx-rs/wgpu/pull/5620), [#5768](https://github.com/gfx-rs/wgpu/pull/5768)
+By @wumpf in [#5569](https://github.com/gfx-rs/wgpu/pull/5569), [#5575](https://github.com/gfx-rs/wgpu/pull/5575), [#5620](https://github.com/gfx-rs/wgpu/pull/5620), [#5768](https://github.com/gfx-rs/wgpu/pull/5768) (together with @kpreid).
 
 #### Querying shader compilation errors
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ TODO(wumpf): This is still work in progress. Should write a bit more about it. A
 
 `wgpu::ComputePass` recording methods (e.g. `wgpu::ComputePass:set_render_pipeline`) no longer impose a lifetime constraint passed in resources.
 
-Furthermore, you can now opt out of `wgpu::ComputePass`'s life time dependency on its parent `wgpu::CommandEncoder` using `wgpu::ComputePass::make_static`:
+Furthermore, you can now opt out of `wgpu::ComputePass`'s lifetime dependency on its parent `wgpu::CommandEncoder` using `wgpu::ComputePass::make_static`:  
 ```rust
 fn independent_cpass<'enc>(encoder: &'enc mut wgpu::CommandEncoder) -> wgpu::ComputePass<'static> {
     let cpass: wgpu::ComputePass<'enc> = encoder.begin_compute_pass(&wgpu::ComputePassDescriptor::default());
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -1286,10 +1286,20 @@ pub struct RenderPass<'a> {
 /// Corresponds to [WebGPU `GPUComputePassEncoder`](
 /// https://gpuweb.github.io/gpuweb/#compute-pass-encoder).
 #[derive(Debug)]
-pub struct ComputePass {
+pub struct ComputePass<'encoder> {
     id: ObjectId,
     data: Box<Data>,
     context: Arc<C>,
+
+    /// Whether end should be called on drop.
+    ///
+    /// Setting this to false is useful for manual ending and
+    /// for safe conversion of the [`ComputePass::encoder_guard`].
+    call_end_on_drop: bool,
+
+    /// The lifetime is just there to protect the [`CommandEncoder`] from being used
+    /// while the pass is alive.
+    encoder_guard: PhantomData<&'encoder ()>,
 }
 
 /// Encodes a series of GPU operations into a reusable "render bundle".
@@ -3866,6 +3876,10 @@ impl CommandEncoder {
     /// Begins recording of a render pass.
     ///
     /// This function returns a [`RenderPass`] object which records a single render pass.
+    ///
+    /// As long as the returned  [`RenderPass`] has not ended,
+    /// any recording operation on this command encoder causes an error and invalidates it.
+    /// (Note that the lifetime constraint protects against this.)
     pub fn begin_render_pass<'pass>(
         &'pass mut self,
         desc: &RenderPassDescriptor<'pass, '_>,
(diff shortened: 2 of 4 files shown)


wgpu/src/lib.rs
```diff
@@ -1286,10 +1286,20 @@ pub struct RenderPass<'a> {
...
+pub struct ComputePass<'encoder> {
     id: ObjectId,
     data: Box<Data>,
     context: Arc<C>,
+
+    /// Whether end should be called on drop.
+    ///
+    /// Setting this to false is useful for manual ending and
+    /// for safe conversion of the [`ComputePass::encoder_guard`].
+    call_end_on_drop: bool,
```
Wumpf: yes, this makes it trivial now to do add explicit end & discard calls which were requested at some point as well :)
Let's do that in a separate PR!

wgpu/src/lib.rs
```diff
@@ -3866,6 +3876,10 @@ impl CommandEncoder {
     /// Begins recording of a render pass.
     ///
     /// This function returns a [`RenderPass`] object which records a single render pass.
+    ///
+    /// As long as the returned  [`RenderPass`] has not ended,
+    /// any recording operation on this command encoder causes an error and invalidates it.
+    /// (Note that the lifetime constraint protects against this.)
```
kpreid: Out of context, this sounds weird to the moderately experienced Rust programmer: “You must not do this, but also, you can’t do this.” I think it should contain the same note as below. Suggested wording (not quite a copy of below but tweaked):

```suggestion
    /// (Note that the `'pass` lifetime relationship normally protects against this,
    /// but can be overridden by [`ComputePass::make_static()`].)
```

wgpu/src/lib.rs
```diff
@@ -3887,7 +3901,17 @@ impl CommandEncoder {
     /// Begins recording of a compute pass.
     ///
     /// This function returns a [`ComputePass`] object which records a single compute pass.
-    pub fn begin_compute_pass(&mut self, desc: &ComputePassDescriptor<'_>) -> ComputePass {
+    ///
+    /// As long as the returned  [`ComputePass`] has not ended,
+    /// any mutable operation on this command encoder causes an error and invalidates it.
```
kpreid:
```suggestion
    /// any mutating operation on this command encoder causes an error and invalidates it.
```

wgpu/src/lib.rs
```diff
@@ -3887,7 +3901,17 @@ impl CommandEncoder {
...
+    /// Note that the lifetime constraint protects against this, but it is possible to opt out of it
+    /// by calling [`ComputePass::make_static`].
+    /// This can be useful for runtime handling of the encoder->pass
```
kpreid, line 3911: However this is phrased exactly, I think `begin_render_pass` and `begin_compute_pass` should have **exactly the same wording**. The benefits of this are that:

* Users don't need to wonder if there's something subtly semantically different implied by the two wordings.
* Future edits are more likely to be done consistently in a find-and-replace fashion.

wgpu/src/lib.rs
```diff
@@ -3887,7 +3901,17 @@ impl CommandEncoder {
...
+    /// dependency e.g. when pass and encoder are stored in the same data structure.
+    pub fn begin_compute_pass<'a>(
+        &'a mut self,
+        desc: &ComputePassDescriptor<'_>,
+    ) -> ComputePass<'a> {
```
kpreid: For consistency:

```suggestion
    pub fn begin_compute_pass<'pass>(
        &'pass mut self,
        desc: &ComputePassDescriptor<'_>,
    ) -> ComputePass<'pass> {
```

wgpu/src/lib.rs
```diff
@@ -4739,7 +4765,42 @@ impl<'a> Drop for RenderPass<'a> {
     }
 }
 
-impl ComputePass {
+impl<'a> ComputePass<'a> {
+    /// Drops lifetime constraint to the parent command encoder, making usage of the encoder
+    /// while this pass is recorded a runtime error instead.
```
kpreid:
```suggestion
    /// Drops the lifetime relationship to the parent command encoder, making usage of
    /// the encoder while this pass is recorded a run-time error instead.
```

wgpu/src/lib.rs
```diff
@@ -4739,7 +4765,42 @@ impl<'a> Drop for RenderPass<'a> {
     }
 }
 
-impl ComputePass {
+impl<'a> ComputePass<'a> {
```
kpreid: How about using the same lifetime name everywhere? This helps readers keep track of the role the lifetime is playing.

```suggestion
impl<'pass> ComputePass<'pass> {
```

wgpu/src/lib.rs
```diff
@@ -4739,7 +4765,42 @@ impl<'a> Drop for RenderPass<'a> {
...
+    /// Drops lifetime constraint to the parent command encoder, making usage of the encoder
+    /// while this pass is recorded a runtime error instead.
+    ///
+    /// Attention: As long as the compute pass has not been ended, any mutable operation on the parent
+    /// command encoder will cause a runtime error and invalidate it!
+    /// By default the lifetime constraint protects from this, but it can be useful
+    /// to handle this at runtime, for example in order to store the pass and encoder in the same
```
kpreid:
```suggestion
    /// to handle this at run time, for example in order to store the pass and encoder in the same
```

(At least in the version of English in my head, it's properly spelled “at run time” like “at compile time”, and “runtime” is only the right spelling for “the runtime” in the sense of helper code.)

wgpu/src/lib.rs
```diff
@@ -4739,7 +4765,42 @@ impl<'a> Drop for RenderPass<'a> {
...
+    /// data structure.
+    ///
+    /// This operation has no effect on pass recording and is considered low overhead.
+    /// It's a safe operation, since [`CommandEncoder`] is in a locked state as long as the pass is active
+    /// regardless of here lifted lifetime constraint.
+    pub fn make_static(mut self) -> ComputePass<'static> {
```
kpreid: I think the name `make_static()` needs improvement. When looking here at the types it's obvious what it means, but out of context when reading code, it could be misleading. “static” doesn't just mean `'static`; it can imply unchanging, compile-time, or leaked. Also, the way it is _used_ is essentially making the program “more dynamic(ally checked)”.

Separately, “`make_`” isn't a usual prefix for this sort of operation; [the convention](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv) would be “`into_`”. In the standard library, `make_` is consistently used for `&mut self` operations that perform some mutation (and may then return another `&mut` obeying some constraint that has been applied).

How about calling this `without_lifetime()`? (The role `'static` is playing here is “no lifetime constraint” more than “something in here lives forever”.)

Or in a different direction, something like  `into_dynamically_checked()`?

wgpu/src/lib.rs
```diff
@@ -4739,7 +4765,42 @@ impl<'a> Drop for RenderPass<'a> {
...
+        // It's tempting to simply transmute the object, but that would be unsound
+        // since `repr(rust)` objects have no guarantees about padding or alignment
+        // even for seemingly equivalent types like `ComputePass<'a>` and `ComputePass<'static>`!
+
+        // Instead, make sure that dropping of `self` is a no-op.
```
kpreid:
```suggestion
        // Instead, make sure that dropping of `self` is a no-op other than dropping its pointers.
```

This distinction is important because if we didn't want to drop `context`, this would be a classic place to use `ManuallyDrop` instead of introducing a flag, but in fact doing that would be a leak.

kpreid (commented):
You pinged me, so you get my review 😁

The most important comment in this review bundle is that I think the name `make_static()` needs work.

wgpu/src/lib.rs
Wumpf: I agree in principle, but this was a deliberate choice for the moment: right now renderpass & computepass _are_ fundamentally different since renderpass doesn't take (shared) ownership of its encoder & passed in resources during recording like computepass does now.
Ideally, before 0.21 gets ready both behave the same and then we can c&p this over. I'll leave a todo note!

wgpu/src/lib.rs
Wumpf: makes sense, but I'll go with `'encoder` since really that's the lifetime we're tracking here

wgpu/src/lib.rs
Wumpf: Makes a lot of sense to me, I agree that we need something better.

`into_dynamically_checked` goes into the right direction imho since it is a "into" operation at its heart just as you say. Only it doesn't quite convey what is checked and it makes it sound like it adds overhead when it doesn't 🤔 

What do you think about `into_encoder_independent_lifetime`? Still not amazing, but it captures that it changes the lifetime associating to the parent encoder 🤔

wgpu/src/lib.rs
Wumpf: ~mhh probably easier to just have it be the same and then leave a note that the lifetime removal method is not yet available, less headache~

urgh but then again I can't link to a thing that doesn't exist yet. I'll just remove the change on RenderPass and only leave the todo note

wgpu/src/lib.rs
kpreid: > What do you think about `into_encoder_independent_lifetime`? Still not amazing, but it captures that it changes the lifetime associating to the parent encoder 🤔

“encoder independent lifetime” sounds like it has a lifetime that still means something specific, rather than one that is now insignificant. I really think that the name should make it clear that it's *removing* something, instead (that's why I suggested “`without_lifetime`”). The fact that there is now a `'static` lifetime is a necessary element of the implementation, but not significant itself. Analogy: Suppose we had a generic type `Foo<T>` and some way to type-erase it to `Foo<()>`. We wouldn’t call it something like “`into_unit_type_parameter()`”, but “`into_erased()`”, because the `()` is just filling the slot, not meaning something itself.

More brainstorming: 

* `.untie_lifetime()` (some people are starting to use "tied" as a description of a lifetime that appears in a function signature in a way that it actually links an input reference to an output reference, rather than, say, being only an input)
* `.forget_lifetime()`

CHANGELOG.md
```diff
@@ -47,13 +47,18 @@ TODO(wumpf): This is still work in progress. Should write a bit more about it. A
 
 `wgpu::ComputePass` recording methods (e.g. `wgpu::ComputePass:set_render_pipeline`) no longer impose a lifetime constraint passed in resources.
 
-Furthermore, `wgpu::ComputePass` no longer has a life time dependency on its parent `wgpu::CommandEncoder`.
+Furthermore, you can now opt out of `wgpu::ComputePass`'s life time dependency on its parent `wgpu::CommandEncoder` using `wgpu::ComputePass::make_static`:
```
kpreid:
```suggestion
Furthermore, you can now opt out of `wgpu::ComputePass`'s lifetime dependency on its parent `wgpu::CommandEncoder` using `wgpu::ComputePass::make_static`:
```

wgpu/src/lib.rs
```diff
@@ -4739,7 +4767,42 @@ impl<'a> Drop for RenderPass<'a> {
     }
 }
 
-impl ComputePass {
+impl<'encoder> ComputePass<'encoder> {
+    /// Drops the lifetime relationship to the parent command encoder, making usage of
+    /// the encoder while this pass is recorded a run-time error instead.
+    ///
+    /// Attention: As long as the compute pass has not been ended, any mutating operation on the parent
+    /// command encoder will cause a runtime error and invalidate it!
```
kpreid:
```suggestion
    /// command encoder will cause a run-time error and invalidate it!
```

wgpu/src/lib.rs
```diff
@@ -4739,7 +4767,42 @@ impl<'a> Drop for RenderPass<'a> {
...
+    /// By default the lifetime constraint protects from this, but it can be useful
```
kpreid:
```suggestion
    /// By default, the lifetime constraint prevents this, but it can be useful
```

wgpu/src/lib.rs
```diff
@@ -4739,7 +4767,42 @@ impl<'a> Drop for RenderPass<'a> {
...
+    /// to handle this at run time, for example in order to store the pass and encoder in the same
```
kpreid:
```suggestion
    /// to handle this at run time, such as when storing the pass and encoder in the same
```

wgpu/src/lib.rs
```diff
@@ -4739,7 +4767,42 @@ impl<'a> Drop for RenderPass<'a> {
...
+    /// data structure.
+    ///
+    /// This operation has no effect on pass recording and is considered low overhead.
+    /// It's a safe operation, since [`CommandEncoder`] is in a locked state as long as the pass is active
+    /// regardless of here lifted lifetime constraint.
```
kpreid:
```suggestion
    /// regardless of the lifetime constraint or its absence.
```

wgpu/src/lib.rs
```diff
@@ -4739,7 +4767,42 @@ impl<'a> Drop for RenderPass<'a> {
...
+    pub fn make_static(mut self) -> ComputePass<'static> {
+        // It's tempting to simply transmute the object, but that would be unsound
+        // since `repr(rust)` objects have no guarantees about padding or alignment
```
kpreid: Nitpick: I'm pretty sure it's impossible for lifetimes to make a difference in layout, because it _must_ be possible to pass a `&Foo<'static>` to a `for<'a, 'b> fn(&'a Foo<'b>)` — and there's no opportunity for the compiler to monomorphize anything there, so the layouts involved must be identical. I don't know that this has been officially written down anywhere, though.

However, I'm not suggesting that this *should* be written in the unsafe fashion; rather, a better reason not to do this is simply “avoid gratuitous unsafe code”. (But before spending any time on a new comment, see the below comment about a different implementation.)

wgpu/src/lib.rs
```diff
@@ -4739,7 +4767,42 @@ impl<'a> Drop for RenderPass<'a> {
...
+        // even for seemingly equivalent types like `ComputePass<'a>` and `ComputePass<'static>`!
+
+        // Instead, make sure that dropping of `self` is a no-op other than dropping its pointers.
+        // (but preserve drop-on-end behavior of the original object)
+        let call_end_on_drop_before = self.call_end_on_drop;
+        self.call_end_on_drop = false;
+
+        // The introduced some unnecessary overhead, but keeps everything safe!
```
kpreid: **This can be done with zero overhead** (no `Arc` clone, no flag update, and no call to `Drop`) and zero `unsafe`, by introducing an intermediate struct which separates the `Drop` implementor from the owner of `encoder_guard`, so that the former can be moved out of the latter. Self-contained proof of concept:

```rust
use std::marker::PhantomData;
use std::sync::Arc;
struct ObjectId;
struct Data;
struct C;

pub struct ComputePass<'encoder> {
    inner: ComputePassInner,
    encoder_guard: PhantomData<&'encoder ()>,
}

/// This struct’s job is to implement `Drop` so `ComputePass` can avoid doing so
/// and be destructurable.
struct ComputePassInner {
    id: ObjectId,
    data: Box<Data>,
    context: Arc<C>,
    call_end_on_drop: bool,
}

impl Drop for ComputePassInner {
    fn drop(&mut self) {}
}

impl<'encoder> ComputePass<'encoder> {
    fn make_static(self) -> ComputePass<'static> {
        let Self { inner, encoder_guard: _ } = self;
        ComputePass {
            inner,
            encoder_guard: PhantomData,
        }
    }
}
```

Wumpf: ah, nice idea. so much simpler!

wgpu/src/lib.rs
Wumpf: compelling case against `into_encoder_independent_lifetime`.
`untie_lifetime` makes sense to me as well, but I fear this has a bit too much unusual jargon.
`forget_lifetime` goes a little bit too much into `std::mem::forget`, but otherwise that's my favorite now, happy to leave it there

wgpu/src/lib.rs
Wumpf: removed again, let's take this separate

Wumpf:
thanks again, I'm much happier with this now!

kpreid (approved):

ErichDonGubler:
AFAICT, controls to `Approve` this PR are inaccessible. But ✅ let it be known that I would do so after reading through this today. Thanks for the help, @kpreid!

PR #5769, opened 2024-06-02
[naga] Validate that all order-sensitive expressions are emitted.

### LINKED ISSUE #5763: Naga doesn't check that subexpressions have been emitted.
Naga doesn't check that each subexpression of an expression used by a statement has been covered by an `Emit` statement.

When added to `naga/tests/validation.rs`, the following test should pass, but it fails:

```rust
/// Validation should reject expressions that refer to un-emitted
/// subexpressions.
#[test]
fn emit_subexpressions() {
    fn variant(
        emit: bool,
    ) -> Result<naga::valid::ModuleInfo, naga::WithSpan<naga::valid::ValidationError>> {
        let span = naga::Span::default();
        let mut module = Module::default();
        let ty_u32 = module.types.insert(
            Type {
                name: Some("u32".into()),
                inner: TypeInner::Scalar(Scalar::U32),
            },
            span,
        );
        let var_private = module.global_variables.append(
            naga::GlobalVariable {
                name: Some("private".into()),
                space: naga::AddressSpace::Private,
                binding: None,
                ty: ty_u32,
                init: None,
            },
            span,
        );

        let mut fun = Function::default();

        // These expressions are pre-emit, so they don't need to be
        // covered by any `Emit` statement.
        let ex_var = fun.expressions.append(Expression::GlobalVariable(var_private), span);
        let ex_42 = fun.expressions.append(Expression::Literal(naga::Literal::U32(42)), span);

        // This expression is neither pre-emit nor used directly by a
        // statement. We want to test whether validation notices when
        // it is not covered by an `Emit` statement.
        let ex_add = fun.expressions.append(Expression::Binary {
            op: naga::BinaryOperator::Add,
            left: ex_42,
            right: ex_42,
        }, span);

        // This expression is used directly by the statement, so if
        // it's not covered by an `Emit`, then validation will catch
        // that.
        let ex_mul = fun.expressions.append(Expression::Binary {
            op: naga::BinaryOperator::Multiply,
            left: ex_add,
            right: ex_add,
        }, span);

        if emit {
            // This `Emit` covers all expressions properly.
            fun.body.push(naga::Statement::Emit(naga::Range::new_from_bounds(ex_add, ex_mul)), span);
        } else {
            // This `Emit` covers `ex_mul` but not its subexpression `ex_add`.
            fun.body.push(naga::Statement::Emit(naga::Range::new_from_bounds(ex_mul, ex_mul)), span);
        }
        fun.body.push(naga::Statement::Store {
            pointer: ex_var,
            value: ex_mul,
        }, span);

        module.functions.append(fun, span);

        valid::Validator::new(
            valid::ValidationFlags::default(),
            valid::Capabilities::all(),
        )
        .validate(&module)
    }

    variant(true).expect("module should validate");
    variant(false).expect_err("validation should notice un-emitted subexpression");
}
```

jimblandy:
Ensure that every expression whose value is affected by statements' side effects is covered by an `Emit` statement.

See the comments on `Expression::is_order_sensitive` for details.

Fixes #5763.

### DIFF
--- a/naga/src/proc/mod.rs
+++ b/naga/src/proc/mod.rs
@@ -540,6 +540,27 @@ impl crate::Expression {
             _ => true,
         }
     }
+
+    /// Return true if this expression's value is sensitive to when it is evaluated.
+    ///
+    /// Expressions like [`Load`] and [`ImageLoad`] can produce different
+    /// values, even when their operands' values are unchanged, because they are
+    /// sensitive to side effects of other statements in the function.
+    ///
+    /// Such order-sensitive expressions must be covered by an [`Emit`]
+    /// statement, ensuring that front ends have fully specified the ordering
+    /// the input language requires, or chosen one arbitrarily if the input
+    /// language doesn't care.
+    ///
+    /// [`Load`]: crate::Expression::Load
+    /// [`ImageLoad`]: crate::Expression::ImageLoad
+    /// [`Emit`]: crate::Statement::Emit
+    pub const fn is_order_sensitive(&self) -> bool {
+        match *self {
+            Self::Load { .. } | Self::ImageLoad { .. } => true,
+            _ => false,
+        }
+    }
 }
 
 impl crate::Function {
--- a/naga/src/valid/function.rs
+++ b/naga/src/valid/function.rs
@@ -174,6 +174,8 @@ pub enum FunctionError {
     InvalidSubgroup(#[from] SubgroupError),
     #[error("Emit statement should not cover \"result\" expressions like {0:?}")]
     EmitResult(Handle<crate::Expression>),
+    #[error("Order-sensitive expression not covered by an `Emit` statement: {0:?}")]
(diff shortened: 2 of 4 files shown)


jimblandy:
Leaving this as a draft, because, naturally, our existing front ends don't actually meet the new requirement:

```console
--- STDERR:              naga::naga-test snapshots::convert_glsl_folder ---
thread 'snapshots::convert_glsl_folder' panicked at naga/tests/snapshots.rs:1054:10:
called `Result::unwrap()` on an `Err` value: WithSpan { inner: Function { handle: [1], name: "main", source: EmitMissing([7]) }, spans: [(Span { start: 65, end: 76 }, "naga::Function [1]"), (Span { start: 84, end: 87 }, "naga::Expression [7]")] }
```

cwfitzgerald:
@jimblandy do you want to keep this PR around? It's quite out of date at this point

cwfitzgerald:
Closing as stale.

PR #5771, opened 2024-06-03
[naga] Ensure that `FooResult` expressions are correctly populated.

jimblandy:
Make Naga module validation require that `CallResult` and `AtomicResult` expressions are indeed visited by exactly one `Call` / `Atomic` statement.


teoxoy (approved):
Looks good!

PR #5773, opened 2024-06-04
Remove vertex_pulling_transfrom from PipelineCompilationOptions.

bradwerth:
This option was only evaluated for Metal backends, and now it's required there so the option is going away. It is still configurable for tests via the PipelineOptions struct, deserialized from .ron files.

This also fixes some type problems with the unpack functions in writer.rs. Metal << operator extends operand to int-sized, which then has to be cast back down to the real size before as_type bit conversion.

**Connections**
N/A

**Description**
This is fully transitioning the Metal backend to use vertex pulling for all cases.

**Testing**
All existing Metal tests now use the vertex pulling transform.

**Checklist**

- [X] Run `cargo fmt`.

### DIFF
--- a/naga/src/back/msl/mod.rs
+++ b/naga/src/back/msl/mod.rs
@@ -354,7 +354,9 @@ pub struct PipelineOptions {
     /// to receive the vertex buffers, lengths, and vertex id as args,
     /// and bounds-check the vertex id and use the index into the
     /// vertex buffers to access attributes, rather than using Metal's
-    /// [[stage-in]] assembled attribute data.
+    /// [[stage-in]] assembled attribute data. This is true by default,
+    /// but remains configurable for use by tests via deserialization
+    /// of this struct. There is no user-facing way to set this value.
     pub vertex_pulling_transform: bool,
 
     /// vertex_buffer_mappings are used during shader translation to
--- a/naga/src/back/msl/writer.rs
+++ b/naga/src/back/msl/writer.rs
@@ -3953,8 +3953,8 @@ impl<W: Write> Writer<W> {
                 )?;
                 writeln!(
                     self.out,
-                    "{}return metal::float2((float(b0) - 128.0f) / 255.0f, \
-                                            (float(b1) - 128.0f) / 255.0f);",
+                    "{}return metal::float2(metal::max(-1.0f, as_type<char>(b0) / 127.0f), \
+                                            metal::max(-1.0f, as_type<char>(b1) / 127.0f));",
                     back::INDENT
                 )?;
                 writeln!(self.out, "}}")?;
@@ -3971,10 +3971,10 @@ impl<W: Write> Writer<W> {
                 )?;
                 writeln!(
                     self.out,
-                    "{}return metal::float4((float(b0) - 128.0f) / 255.0f, \
-                                            (float(b1) - 128.0f) / 255.0f, \
-                                            (float(b2) - 128.0f) / 255.0f, \
-                                            (float(b3) - 128.0f) / 255.0f);",
+                    "{}return metal::float4(metal::max(-1.0f, as_type<char>(b0) / 127.0f), \
+                                            metal::max(-1.0f, as_type<char>(b1) / 127.0f), \
+                                            metal::max(-1.0f, as_type<char>(b2) / 127.0f), \
+                                            metal::max(-1.0f, as_type<char>(b3) / 127.0f));",
                     back::INDENT
                 )?;
                 writeln!(self.out, "}}")?;
@@ -4033,8 +4033,8 @@ impl<W: Write> Writer<W> {
                 )?;
                 writeln!(
                     self.out,
-                    "{}return metal::int2(as_type<metal::short>(b1 << 8 | b0), \
-                                          as_type<metal::short>(b3 << 8 | b2));",
+                    "{}return metal::int2(as_type<short>(metal::ushort(b1 << 8 | b0)), \
+                                          as_type<short>(metal::ushort(b3 << 8 | b2)));",
                     back::INDENT
                 )?;
                 writeln!(self.out, "}}")?;
@@ -4055,10 +4055,10 @@ impl<W: Write> Writer<W> {
                 )?;
                 writeln!(
                     self.out,
-                    "{}return metal::int4(as_type<metal::short>(b1 << 8 | b0), \
-                                          as_type<metal::short>(b3 << 8 | b2), \
-                                          as_type<metal::short>(b5 << 8 | b4), \
-                                          as_type<metal::short>(b7 << 8 | b6));",
+                    "{}return metal::int4(as_type<short>(metal::ushort(b1 << 8 | b0)), \
+                                          as_type<short>(metal::ushort(b3 << 8 | b2)), \
+                                          as_type<short>(metal::ushort(b5 << 8 | b4)), \
+                                          as_type<short>(metal::ushort(b7 << 8 | b6)));",
                     back::INDENT
                 )?;
                 writeln!(self.out, "}}")?;
@@ -4117,8 +4117,7 @@ impl<W: Write> Writer<W> {
                 )?;
                 writeln!(
                     self.out,
-                    "{}return metal::float2((float(b1 << 8 | b0) - 32767.0f) / 65535.0f, \
-                                            (float(b3 << 8 | b2) - 32767.0f) / 65535.0f);",
+                    "{}return metal::unpack_snorm2x16_to_float(b1 << 24 | b0 << 16 | b3 << 8 | b2);",
                     back::INDENT
                 )?;
                 writeln!(self.out, "}}")?;
@@ -4139,10 +4138,8 @@ impl<W: Write> Writer<W> {
                 )?;
                 writeln!(
                     self.out,
-                    "{}return metal::float4((float(b1 << 8 | b0) - 32767.0f) / 65535.0f, \
-                                            (float(b3 << 8 | b2) - 32767.0f) / 65535.0f, \
-                                            (float(b5 << 8 | b4) - 32767.0f) / 65535.0f, \
-                                            (float(b7 << 8 | b6) - 32767.0f) / 65535.0f);",
+                    "{}return metal::float4(metal::unpack_snorm2x16_to_float(b1 << 24 | b0 << 16 | b3 << 8 | b2), \
+                                            metal::unpack_snorm2x16_to_float(b5 << 24 | b4 << 16 | b7 << 8 | b6));",
                     back::INDENT
                 )?;
                 writeln!(self.out, "}}")?;
@@ -4159,8 +4156,8 @@ impl<W: Write> Writer<W> {
                 )?;
                 writeln!(
                     self.out,
-                    "{}return metal::float2(as_type<metal::half>(b1 << 8 | b0), \
-                                            as_type<metal::half>(b3 << 8 | b2));",
+                    "{}return metal::float2(as_type<half>(metal::ushort(b1 << 8 | b0)), \
+                                            as_type<half>(metal::ushort(b3 << 8 | b2)));",
                     back::INDENT
                 )?;
                 writeln!(self.out, "}}")?;
@@ -4170,7 +4167,7 @@ impl<W: Write> Writer<W> {
                 let name = self.namer.call("unpackFloat16x4");
                 writeln!(
                     self.out,
-                    "metal::int4 {name}(metal::ushort b0, \
+                    "metal::float4 {name}(metal::ushort b0, \
                                         metal::ushort b1, \
                                         metal::ushort b2, \
                                         metal::ushort b3, \
@@ -4181,10 +4178,10 @@ impl<W: Write> Writer<W> {
                 )?;
                 writeln!(
                     self.out,
-                    "{}return metal::int4(as_type<metal::half>(b1 << 8 | b0), \
-                                          as_type<metal::half>(b3 << 8 | b2), \
-                                          as_type<metal::half>(b5 << 8 | b4), \
-                                          as_type<metal::half>(b7 << 8 | b6));",
+                    "{}return metal::float4(as_type<half>(metal::ushort(b1 << 8 | b0)), \
+                                          as_type<half>(metal::ushort(b3 << 8 | b2)), \
+                                          as_type<half>(metal::ushort(b5 << 8 | b4)), \
+                                          as_type<half>(metal::ushort(b7 << 8 | b6)));",
                     back::INDENT
                 )?;
                 writeln!(self.out, "}}")?;
@@ -4390,10 +4387,10 @@ impl<W: Write> Writer<W> {
                 let name = self.namer.call("unpackSint32");
                 writeln!(
                     self.out,
-                    "metal::int {name}(uint b0, \
-                                       uint b1, \
-                                       uint b2, \
-                                       uint b3) {{"
+                    "int {name}(uint b0, \
+                                uint b1, \
+                                uint b2, \
+                                uint b3) {{"
                 )?;
                 writeln!(
                     self.out,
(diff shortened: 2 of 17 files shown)


nical (approved):

naga/src/back/msl/mod.rs
```diff
@@ -346,7 +346,9 @@ pub struct PipelineOptions {
     /// to receive the vertex buffers, lengths, and vertex id as args,
     /// and bounds-check the vertex id and use the index into the
     /// vertex buffers to access attributes, rather than using Metal's
-    /// [[stage-in]] assembled attribute data.
+    /// [[stage-in]] assembled attribute data. This is true by default,
+    /// but remains configurable for use by tests via deserialization
+    /// of this struct. There is no user-facing way to set this value.
     pub vertex_pulling_transform: bool,
```
ErichDonGubler, line 360: question: Why not just remove this? 🤔 Are we planning to, maybe as a separate step?

naga/src/back/msl/writer.rs
```diff
@@ -4349,10 +4349,10 @@ impl<W: Write> Writer<W> {
                 let name = self.namer.call("unpackSint32");
                 writeln!(
                     self.out,
-                    "metal::int {name}(uint b0, \
-                                       uint b1, \
-                                       uint b2, \
-                                       uint b3) {{"
+                    "int {name}(uint b0, \
```
ErichDonGubler, line 4390: question: Why do we remove the `metal` path segment here (i.e., `s/metal::int/int/`)?

naga/src/back/msl/writer.rs
```diff
@@ -4140,10 +4140,10 @@ impl<W: Write> Writer<W> {
...
                 writeln!(
                     self.out,
-                    "{}return metal::int4(as_type<metal::half>(b1 << 8 | b0), \
-                                          as_type<metal::half>(b3 << 8 | b2), \
-                                          as_type<metal::half>(b5 << 8 | b4), \
-                                          as_type<metal::half>(b7 << 8 | b6));",
+                    "{}return metal::int4(as_type<metal::half>(metal::ushort(b1 << 8 | b0)), \
+                                          as_type<metal::half>(metal::ushort(b3 << 8 | b2)), \
+                                          as_type<metal::half>(metal::ushort(b5 << 8 | b4)), \
+                                          as_type<metal::half>(metal::ushort(b7 << 8 | b6)));",
```
ErichDonGubler: question: Why are `metal::ushort` constructions are being added here?

tests/tests/vertex_indices/mod.rs
```diff
@@ -356,20 +337,17 @@ async fn vertex_index_common(ctx: TestingContext) {
         )
         .create_view(&wgpu::TextureViewDescriptor::default());
 
-    let mut tests = Vec::with_capacity(5 * 2 * 2 * 2);
+    let mut tests = Vec::with_capacity(5 * 2 * 2);
```
ErichDonGubler: thought(non-blocking): I understand why we're diminishing the capacity here, but I think it's in the name of something that's not worth it.

issue(blocking): If the intent is to only allocate once for `tests`' storage (which is the only reason I can imagine we'd want to call `Vec::with_capacity` here), then this expression was wrong before, and is still wrong after. The old expression is correct _after_ removing `vertex_pulling_transforms`, i.e., this should have been `5 * 2 * 2 * 2 * 2` (case, ID source, draw call kind, encoder kind, _and_ vertex-pulling transform boolean) prior to the change.

suggestion(blocking): Let's just remove this hunk for now, _or_ (even better) add a commit fixing this mistake (i.e., adding `* 2`) before this change, and then immediately reverting it in the “intended” change's commit.

☝🏻CC:

* @cwfitzgerald as original author in <https://github.com/gfx-rs/wgpu/pull/4723>.
* @teoxoy as reviewer of <https://github.com/gfx-rs/wgpu/pull/4723> and the last changes made to this line at <https://github.com/gfx-rs/wgpu/pull/5254>, too.

ErichDonGubler: suggestion(non-blocking): I doubt we're affecting noticeably performance here, and I think in testing code like this, clarity trumps minor performance gains. Let's do one of two things:

1. Use `Itertools::cartesian_product` and `Iterator::collect` over these `ARRAY` constants (or something else that I'll mention in another comment about `strum`) to get what we want with fewer footguns and mental cache pressure:

    ```rust
    let tests = TestCase::ARRAY
        .cartesian_product(IdSource::ARRAY)
        .cartesian_product(DrawCallKind::ARRAY)
        .cartesian_product(EncoderKind::ARRAY)
        .map(|(((case, id_source), draw_call_kind), encoder_kind)| Test {
            case,
            id_source,
            draw_call_kind,
            encoder_kind,
        })
        .collect::<Vec<_>>();
    ```
2. Just replace `tests`' initial value with a `Vec::new`, and not worry about re-allocation with such few elements:

    ```rust
    let mut tests = Vec::new();
    ```


☝🏻CC folks who participated in introducing this code in <https://github.com/gfx-rs/wgpu/pull/4723> (@cwfitzgerald as author, @teoxoy as reviewer).

tests/tests/vertex_indices/mod.rs
```diff
@@ -356,20 +337,17 @@ async fn vertex_index_common(ctx: TestingContext) {
...
     for case in TestCase::ARRAY {
         for id_source in IdSource::ARRAY {
             for draw_call_kind in DrawCallKind::ARRAY {
                 for encoder_kind in EncoderKind::ARRAY {
```
ErichDonGubler: thought(non-blocking): These manual `ARRAY` constants are unfortunate. We actually don't need to deal with them; we can use the nice `strum::{IntoEnumIterator, EnumIter}` APIs to `derive` an `Iterator` over the elements of these different `enum`s instead, i.e.:

```diff
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, strum::EnumIter)] // adds `TestCase::iter()` instead
 enum TestCase {
     // …
 }
 
 impl TestCase {
-    const ARRAY: [Self; 5] = [
-        // …
-    ];
     // …
 }
```

We'll need to add a dependency to `strum` here, but we already have it in our lockfile, so 🤷🏻‍♂️ not worried about dependency bloat. It's a _very_ useful dep. to have.

☝🏻CC folks who participated in introducing this code in <https://github.com/gfx-rs/wgpu/pull/4723> (@cwfitzgerald as author, @teoxoy as reviewer). Also, CC @gfx-rs/wgpu for awareness.

ErichDonGubler (dismissed):
I'm leaning towards approving this, but there are a couple of things I think we should resolve:
* I have some questions about the Metal codegen changes; I didn't expect or understand them yet. This makes me uneasy; weren't we shipping something tested to be correct before? 🤔
* ~~I also noticed one incorrect thing with our usage of `Vec::with_capacity` that should be pretty trivial to resolve.~~ EDIT: Resolved on my end!

tests/tests/vertex_indices/mod.rs
ErichDonGubler: Resolved the blocking feedback I've raised in <https://github.com/gfx-rs/wgpu/pull/5773#discussion_r1645009650> with 53ddbc92 and 86fff753, and a new version of the original revision I've reviewed (ea77dc2a).

No further blocking feedback here; will file a follow-up for the remaining non-blocking stuff momentarily.

tests/tests/vertex_indices/mod.rs
ErichDonGubler: Let's move non-blocking feedback in this convo to <https://github.com/gfx-rs/wgpu/pull/5840>.

tests/tests/vertex_indices/mod.rs
ErichDonGubler: Let's move non-blocking feedback in this convo to <https://github.com/gfx-rs/wgpu/pull/5840>. Resolving.

naga/src/back/msl/mod.rs
ErichDonGubler: I guess @gfx-rs/naga might be interested in keeping this, so I'll invite their feedback, too.

cwfitzgerald:
For a change like this, possibly before this lands, definitely before we release next (1 mo ish), we need to reach out to our users for gpu side performance testing. Obviously enabling should be the default, but if this forces significant regressions, we should be exposing a escape hatch to the user.

I also feel like we should have runtime vertex format tests now that we're polyfilling them on a platform.

naga/src/back/msl/mod.rs
bradwerth: > question: Why not just remove this? 🤔 Are we planning to, maybe as a separate step?

Yeah, I'd like to do code tear-outs as a separate step. I'd like to keep this PR as targeted as possible.

naga/src/back/msl/writer.rs
bradwerth: Since MSL `int` matches the name and type of the C++ `int`, the `metal::` prefix is flagged as being unnecessary. I don't completely understand it, but we need to remove it to pass our MSL validation step.

naga/src/back/msl/writer.rs
bradwerth: The bit shift operators extend the result to u32, which can't be `as_type`'d into `metal::half`. So it has to be constrained back to `ushort` before the conversion.

bradwerth:
> I'm leaning towards approving this, but there are a couple of things I think we should resolve:
> 
>     * I have some questions about the Metal codegen changes; I didn't expect or understand them yet. This makes me uneasy; weren't we shipping something tested to be correct before? 🤔

We don't have complete test coverage of all the vertex formats. @cwfitzgerald notes in a commment on this PR that we should. I'll expand this PR to include a test update that exercises evertyhing, which will help justify the changes to the unpack functions.

bradwerth:
Converting to draft while I expand the test coverage of all the unpack functions.

tests/tests/vertex_formats/mod.rs
```diff
@@ -0,0 +1,393 @@
...
+    let attributes_block_4 = &wgpu::vertex_attr_array![
+        0 => Float32,
+        1 => Float32x2,
+        2 => Float32x3,
+        3 => Float32x4,
+        // TODO(SHADER_F16)
+        /*
+        4 => Float16x2,
+        5 => Float16x4,
+        */
```
teoxoy: These should actually work without the `f16` extension. On the shader side we can expect `vec2<f32>` and `vec4<f32>`.

tests/tests/vertex_formats/mod.rs
```diff
@@ -0,0 +1,393 @@
...
+                0u8, 0u8, 0u8, 191u8, 0u8, 0u8, 128u8, 64u8, // Float32x2 (-0.5, 4.0)
+                0u8, 0u8, 0u8, 192u8, 0u8, 0u8, 204u8, 194u8, 0u8, 0u8, 200u8,
+                66u8, // Float32x3 (-2.0, -102.0, 100.0)
+                0u8, 0u8, 92u8, 66u8, 0u8, 0u8, 72u8, 194u8, 0u8, 0u8, 32u8, 65u8, 0u8, 0u8, 128u8,
+                63u8, // Float32x4 (55.0, -50.0, 10.0, 1.0)
+                      // TODO(SHADER_F16)
+                      // Expected value is -1.5
+                      /*
+                      0u8, 60u8, 85u8, 53u8, // Float16x2 (1.0, 0.33)
+                      71u8, 57u8, 0u8, 192u8, 0u8, 188u8, 0u8, 184u8, // Float16x4 (0.66, -2.0, -1.0, -0.5)
```
teoxoy:
```suggestion
                      0u8, 60u8, 72u8, 53u8, // Float16x2 (1.0, 0.33)
                      72u8, 57u8, 0u8, 192u8, 0u8, 188u8, 0u8, 184u8, // Float16x4 (0.66, -2.0, -1.0, -0.5)
```

teoxoy (changes requested):
This is great!

teoxoy:
It took me a while to double-check the decimal to byte representations, maybe we can use rust's `v.to_le_bytes()` methods to improve the DX. Not a hard blocker though.

cwfitzgerald:
We can also use `half` and bytemuck, or other such things. These are tests, so new dependencies aren't a big deal.

tests/tests/vertex_formats/mod.rs
bradwerth: Yes, but then wgsl requires the attribute to be declared as `vec2<f32>` but the vertex format remains `Float16x2`, so the unpack functions try to assign a `half2` to a `float2`, which isn't allowed. Also, the vertex buffer then has to be padded out with empty bytes (because the input type/attributes are each f32), which the unpack functions then have to ignore/skip. It seems better to wait until we have `f16` support in wgsl.

tests/tests/vertex_formats/mod.rs
bradwerth: Done.

bradwerth:
Well, however the bytes are specified, the checksums are verifying that they compute to the correct values.

tests/tests/vertex_formats/mod.rs
teoxoy: The spec currently allows the mapping of `Float16xN` to `vecN<f32>`. These are the relevant sections: https://gpuweb.github.io/gpuweb/#vertex-processing & https://gpuweb.github.io/gpuweb/wgsl/#channel-formats.

And since we don't implement `f16` yet, the unpacking functions just return `f32`.

https://github.com/gfx-rs/wgpu/blob/aeb2067e8120c1ff480625c00b9571db8d01d5a4/naga/src/back/msl/writer.rs#L4151-L4163

teoxoy (changes requested):
We should resolve https://github.com/gfx-rs/wgpu/pull/5773#discussion_r1679085998.

tests/tests/vertex_formats/mod.rs
bradwerth: I'll try to do this, but I've hit snags that I'm still working through. Some of the issues:

1) Regardless of the type used in the shader, the `Float16x2` format in the vertex buffer is 4 bytes, not 8 bytes.
2) If the `location(n)` annotation is declared as `vec2<f32>`, then the shader is expecting 8 bytes for that vec2 value.
3) When the test fills out the vertex buffer, it has to provide all those bytes to make the offsets work out.
4) If the test writes extra bytes (zeroes?) then the unpack function has to be revised to accept and ignore those bytes.
5) Presuming I can get all of this to work, are these the same steps that end-users have to do to use `Float16x2` vertex format? If so, where are we going to document that?

Perhaps I'm misunderstanding what needs to be done here, but this seems like a real mess. I'll see what I can get working.

bradwerth: Huh. This time when I tried to make all the changes, it Just Worked. I guess my statement #3 above, is incorrect. Supplying 4 bytes for `Float16x2` works just fine.

teoxoy: Yup, that is how it's supposed to work. `Float16x2` is 4 bytes in the vertex buffer and when we read them out we "expand" them into a `vec2<f32>` (8 bytes).

teoxoy: I think the only changes need to get the new tests passing is to tweak the unpacking functions.

bradwerth:
> It took me a while to double-check the decimal to byte representations, maybe we can use rust's `v.to_le_bytes()` methods to improve the DX. Not a hard blocker though.

The syntax for this is just markedly worse. Since the numbers are a mix of 2-byte, 4-byte, and 6-byte, and 8-byte values, it's not easy to flatten them into a single `u8` array/Vec since there's no way to flatten those differently-types iterators. We end up with syntax like this:
```
let input_block_0 = &mut Vec::<u8>::new();
input_block_0.extend_from_slice(&0xC0C0C0C0u32.to_le_bytes()); // Unorm8x4 (0.5, 0.5, 0.5, 0.5)
input_block_0.extend_from_slice(&0xC000C000u32.to_le_bytes()); // Unorm16x2 (0.5, 0.5)
input_block_0.extend_from_slice(&0x4000400040004000u64.to_le_bytes()); // Unorm16x4 (0.25, 0.25, 0.25, 0.25)
input_block_0.extend_from_slice(&0x7F7F7F7Fu32.to_le_bytes()); // Snorm8x4 (1, 1, 1, 1)
input_block_0.extend_from_slice(&0xC000C000u32.to_le_bytes()); // Snorm16x2 (-1, -1)
input_block_0.extend_from_slice(&0x7FFF7FFF7FFF7FFFu64.to_le_bytes()); // Snorm16x4 (1, 1, 1, 1)
input_block_0.extend_from_slice(&0xFFFFu32.to_le_bytes()[0..2]); // Unorm8x2 (1, 1)
input_block_0.extend_from_slice(&0xC0C0u32.to_le_bytes()[0..2]); // Snorm8x2 (-1, -1)
```
I don't think this is improving readability. It's also actually wrong -- I got at least one of those values encoded incorrectly and it's messing up the checksums. I strongly urge that we leave the byte specification as-is. Re-open if you disagree.

bradwerth:
Didn't mean to close, sorry.

bradwerth:
> I don't think this is improving readability. It's also actually wrong -- I got at least one of those values encoded incorrectly and it's messing up the checksums. I strongly urge that we leave the byte specification as-is. Re-open if you disagree.

The error is trivial upon reflection: the `C0`s should be `80`s. I still think the syntax is worse.

teoxoy (approved):

PR #5775, opened 2024-06-05
spv-in upgrading AtomicIIncrement type

schell:
**Connections**
Partially addresses https://github.com/gfx-rs/wgpu/issues/4489. 
Follow up to https://github.com/gfx-rs/wgpu/pull/5702. 

This creates a new module `naga/src/front/atomic_upgrade.rs` that begins the upgrade process by recursing through types and expressions previously marked during parsing. 

~This results in a `Module` that contains all needed types, but still does not validate, as the upgrades to expressions need to be applied to any expressions that use those expressions until all are resolved.~
This results in a `Module` that validates, at least for the test case.

@jimblandy @cwfitzgerald :) 

**Checklist**

- [x] Run `cargo fmt`.


naga/src/front/atomic_upgrade.rs
```diff
@@ -0,0 +1,373 @@
...
+    pub pointer_type_handle: Handle<Type>,
+    /// Handle to the pointer expression in the module/function
+    pub pointer_handle: Handle<Expression>,
+}
+
+#[derive(Default)]
+struct Padding {
+    padding: Arc<AtomicUsize>,
+    current: usize,
+}
```
schell: `Padding` is just for my debugging.

naga/src/front/atomic_upgrade.rs
```diff
@@ -0,0 +1,373 @@
...
+            Expression::GlobalVariable(var) => {
+                Expression::GlobalVariable(self.upgrade_global_variable(var)?)
+            }
+            Expression::LocalVariable(var) => Expression::LocalVariable(
+                self.upgrade_local_variable(maybe_fn_handle.ok_or(Error::NoFunction)?, var)?,
+            ),
+            _ => {
+                return Err(Error::Unsupported);
+            }
+        };
```
schell: I'm purposefully only matching on the bare minimum here to make the diff as small as possible. I actually have the full match in a branch, but this way it's easier to see the structure of the PR.

naga/src/front/atomic_upgrade.rs
```diff
@@ -0,0 +1,373 @@
...
+            }
+
+            padding.debug("upgrading the pointer type:", op.pointer_type_handle);
+            let _new_pointer_type_handle = state.upgrade_type(op.pointer_type_handle)?;
+
+            padding.debug("upgrading the pointer expression", op.pointer_handle);
+            let _new_pointer_handle =
+                state.upgrade_expression(maybe_fn_handle, op.pointer_handle)?;
+        }
+
```
schell, line 211: The next step is to "spiral out" with a search/replace of expressions that use `op.pointer_handle`.

schell: (and "upgrade" those to use `_new_pointer_handle`)

naga/src/lib.rs
```diff
@@ -991,7 +991,7 @@ pub struct GlobalVariable {
 }
 
 /// Variable defined at function level.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
```
jimblandy: Rather than trying to apply equality to the entire `GlobalVariable`, I think it would be better for `upgrade_global_variable` to explicitly compare the new type and initializer expression handles to see if they're different. I don't think `PartialEq` on `GlobalVariable` really makes sense.

naga/tests/in/spv/atomic_i_increment.spv
jimblandy, line 1: Could you provide a `.spvasm` file along with the binary spv files? Just something from `spirv-dis` is fine.

naga/src/front/atomic_upgrade.rs
```diff
@@ -0,0 +1,334 @@
...
+/// along with enough information for us to find the items whose types
+/// we need to upgrade.
+///
+/// [at]: crate::TypeInner::Atomic
+/// [as]: crate::Statement::Atomic
+#[allow(dead_code)]
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct AtomicOp {
+    pub instruction: AtomicOpInst,
+    /// Handle to the pointer's type in the module
```
jimblandy: The important phrase in this comment is "the pointer" - the field's type tells us that it's a type handle, and the module is the only place where types appear. I'll push a suggested improvement.

naga/src/front/atomic_upgrade.rs
```diff
@@ -0,0 +1,334 @@
...
+    fn from(value: crate::arena::BadHandle) -> Self {
+        Error::MissingHandle(value)
+    }
+}
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum AtomicOpInst {
+    AtomicIIncrement,
+}
+
```
jimblandy: This information isn't actually used at the moment, right? It seems like all the atomic operations are going to be about the same.

naga/src/front/atomic_upgrade.rs
```diff
@@ -0,0 +1,334 @@
...
+    pub(crate) fn upgrade_atomics(
+        &mut self,
+        ops: impl IntoIterator<Item = AtomicOp>,
+    ) -> Result<(), Error> {
+        let mut state = UpgradeState {
+            padding: Default::default(),
+            module: self,
+        };
+
+        for op in ops.into_iter() {
```
jimblandy: `for` loops implicitly apply `into_iter` for you, so this call isn't needed.

naga/src/front/atomic_upgrade.rs
```diff
@@ -0,0 +1,334 @@
...
+            let padding = state.inc_padding();
+            padding.debug("op: ", op);
+
+            // Find the expression's enclosing function, if any
+            let mut maybe_fn_handle = None;
+            for (fn_handle, function) in state.module.functions.iter() {
+                log::trace!("function: {fn_handle:?}");
+                if function.expressions.try_get(op.pointer_handle).is_ok() {
```
jimblandy: This isn't going to work at all. Take a few minutes to just read through `arena.rs`. A `Handle<T>` is just a `usize` with pretensions. If it's `5`, then this is going to succeed for any `Function` whose `expressions` arena has five expressions in it.

Since `AtomicOp`s are one-to-one with `Atomic` statements, you should have a separate list of `AtomicOp`s for each `Function` and `EntryPoint`.

naga/src/front/atomic_upgrade.rs
```diff
@@ -0,0 +1,334 @@
...
+}
+
+impl<'a> UpgradeState<'a> {
+    fn inc_padding(&self) -> Padding {
+        self.padding.inc_padding()
+    }
+
+    /// Upgrade the type, recursing until we reach the leaves.
+    /// At the leaves, replace scalars with atomic scalars.
+    fn upgrade_type(&mut self, type_handle: Handle<Type>) -> Result<Handle<Type>, Error> {
```
jimblandy: We usually just use `ty` and `inner` for parameters/variables like this.

naga/src/front/atomic_upgrade.rs
```diff
@@ -0,0 +1,334 @@
...
+        let padding = self.inc_padding();
+        padding.debug("upgrading type: ", type_handle);
+        let type_ = self
+            .module
+            .types
+            .get_handle(type_handle)
+            .map_err(Error::MissingHandle)?
```
jimblandy: Since this is used by the SPIR-V front end, which just constructed the module, I think it's okay to just panic if the handle is bad. That's what the SPIR-V front end will do anyway if it's botched its own machinations. So this could simply be `self.module.types[type_handle]`.

naga/src/front/atomic_upgrade.rs
```diff
@@ -0,0 +1,334 @@
...
+    fn inc_padding(&self) -> Padding {
+        self.padding.inc_padding()
+    }
+
+    /// Upgrade the type, recursing until we reach the leaves.
+    /// At the leaves, replace scalars with atomic scalars.
+    fn upgrade_type(&mut self, type_handle: Handle<Type>) -> Result<Handle<Type>, Error> {
+        let padding = self.inc_padding();
+        padding.debug("upgrading type: ", type_handle);
+        let type_ = self
```
jimblandy: I discourage using `_` as a suffix for names that conflict with Rust reserved words. Everyone wants to pick a different convention, and then it's hard to guess what's going on. Instead, I recommend just using `r#type`. It's visually noisy but everyone knows exactly what's going on the moment they see it.

naga/src/front/atomic_upgrade.rs
```diff
@@ -0,0 +1,334 @@
...
+            },
+            TypeInner::Array { base, size, stride } => TypeInner::Array {
+                base: self.upgrade_type(base)?,
+                size,
+                stride,
+            },
+            TypeInner::Struct { members, span } => TypeInner::Struct {
+                members: {
+                    let mut new_members = vec![];
+                    for member in members.iter().cloned() {
```
jimblandy: Ultimately I don't think it's going to work to upgrade all the members of the struct. You're going to need to identify exactly which struct members need to be upgraded.

jimblandy: I'll have some more detailed comments on this in a bit.

naga/src/front/atomic_upgrade.rs
```diff
@@ -0,0 +1,334 @@
...
+            Expression::AccessIndex { base, index } => Expression::AccessIndex {
+                base: self.upgrade_expression(maybe_fn_handle, base)?,
+                index,
+            },
+            Expression::GlobalVariable(var) => {
+                Expression::GlobalVariable(self.upgrade_global_variable(var)?)
+            }
+            Expression::LocalVariable(var) => Expression::LocalVariable(
+                self.upgrade_local_variable(maybe_fn_handle.ok_or(Error::NoFunction)?, var)?,
+            ),
```
jimblandy: As far as I know, atomics cannot appear in local variables. Neither WGSL nor Naga IR permits this. And it doesn't seem meaningful anyway - only one thread can ever see a local variable anyway.

naga/src/lib.rs
jimblandy: Err, I mean, `LocalVariable`.

naga/src/front/atomic_upgrade.rs
```diff
@@ -0,0 +1,338 @@
...
+            name: var.name.clone(),
+            space: var.space,
+            binding: var.binding.clone(),
+            ty: self.upgrade_type(var.ty)?,
+            init: self.upgrade_opt_expression(None, var.init)?,
+        };
+        if new_var != var {
+            padding.debug("new global variable: ", &new_var);
+            let span = self.module.global_variables.get_span(handle);
+            let new_handle = self.module.global_variables.append(new_var, span);
```
jimblandy: You can't duplicate the global variable here - you need to modify the type of the existing `GlobalVariable` in place. This actually makes things simpler, because you don't need to change any `Expression`s, they're all pointing at the right `GlobalVariable` already.

jimblandy (changes requested):
There are bits and pieces of this that seem right - I think the type upgrading is almost right, for example - but I think the overall approach needs to be changed.

First of all, would be it feasible to restrict our focus for the time being to global variables with atomic or array-of-atomic types, leaving struct members and function arguments that are pointers to atomic values for later PRs?

In that case, what the SPIR-V front end needs to produce is a set of global variables whose types need to be upgraded. Atomics can only appear in `workgroup` and `storage`, so they can only really be held in global variables, we don't need to worry about locals or arguments. So instead of the `FastHashMap<..., AtomicOp>` structure you're producing now, you probably just need an `IndexSet<Handle<GlobalVariable>>`. The front end can discover the relevant global by looking into the `Expression` it just constructed for the atomic instruction's pointer operand.

(If the set will be iterated over, we prefer `IndexSet` over `HashSet`, to keep the iteration order deterministic.)

Then, the `upgrade_atomics` function needs to iterate over all those globals and upgrade their types. I *think* you shouldn't need to update any `Expression`s at all, so you shouldn't need to mess with any functions or expressions.

naga/src/lib.rs
jimblandy: Err, I mean, `LocalVariable`.

naga/tests/in/spv/atomic_i_increment.spv
schell: Sure thing. Is there a way to `parse` that?

naga/src/front/atomic_upgrade.rs
schell: Yes this is probably premature. I guess if I need an enum for this I should add it when it's needed.

naga/tests/in/spv/atomic_i_increment.spv
schell: Included 👍

naga/tests/in/spv/atomic_i_increment.spvasm
```diff
@@ -0,0 +1,62 @@
...
+         %30 = OpLabel
+         %24 = OpPhi %uint %18 %31 %34 %32
+         %36 = OpSelect %bool %29 %false %true
+               OpBranch %25
+         %25 = OpLabel
+               OpBranchConditional %36 %22 %26
+         %26 = OpLabel
+               OpReturn
+               OpFunctionEnd
+
```
schell, line 62: This SPIR-V is generated by `rust-gpu` for this rust code: 

```rust 
#[spirv(compute(threads(32)))]
pub fn test_atomic_i_increment(
    #[spirv(storage_buffer, descriptor_set = 0, binding = 0)] global_index: &mut u32,
    #[spirv(storage_buffer, descriptor_set = 0, binding = 1)] times: &u32,
) {
    let mut i = 0u32;
    loop {
        if i >= *times {
            break;
        }
        let _ = unsafe {
            spirv_std::arch::atomic_i_increment::<
                u32,
                { spirv_std::memory::Scope::Workgroup as u32 },
                { spirv_std::memory::Semantics::NONE.bits() },
            >(global_index)
        };
        i += 1;
    }
}
```

I did this because I want to make sure we can upgrade the atomic type here, without borking references to the type it was upgraded from (`u32`).

naga/src/front/atomic_upgrade.rs
schell: Oh, wow. That one little change (modifying the global var in place) actually makes the generated module validate!

naga/src/lib.rs
schell: And if we're not upgrading local variables (since atomics cannot be found in them), then we don't need this anyway.

naga/src/front/atomic_upgrade.rs
schell: Ah, yes.

naga/src/front/atomic_upgrade.rs
schell: Ok, that makes sense. The reason I (incorrectly) did it this way is because at parsing time we don't yet have a handle to a function.

schell: But now that I look at `BlockContext` it seems we have a `function_id`, so I'll see if I can make it work.

schell:
Ok @jimblandy I'll try what you suggest with regard to tracking `IndexSet<Handle<GlobalVariable>>` instead of a map of `AtomicOp`.

naga/tests/in/spv/atomic_i_increment.spvasm
jimblandy: We generally check in the disassembly, even when the SPV was generated by something else, like rust-gpu, just as a convenience to other devs wondering what the file is even doing.

Of course, they could just run spirv-dis themselves - and arguably they should, just to remove the possibility of things getting out of sync. But for now it looks to me like everything has an .spvasm file.

jimblandy:
The reason I think you won't need to update any expressions is that Naga IR Load expressions and Store statements both can operate on Atomics, so everything accessing the globals whose types you're whacking, whether Loads, Stores, or Atomics, should still be okay.

naga/tests/in/spv/atomic_i_increment.spv
jimblandy: Unfortunately, no - it would definitely be better to have a single source of authority. We just provide them both and hope they don't get out of sync.

naga/src/front/atomic_upgrade.rs
```diff
@@ -0,0 +1,238 @@
...
+        padding.trace("upgrading global variable: ", handle);
+
+        let var = self.module.global_variables[handle].clone();
+
+        let new_var = GlobalVariable {
+            name: var.name.clone(),
+            space: var.space,
+            binding: var.binding.clone(),
+            ty: self.upgrade_type(var.ty)?,
+            init: self.upgrade_opt_expression(None, var.init)?,
```
schell: If we don't need to upgrade the global variable's `init` here, then I think we can get rid of a lot more code. It would remove the need for `upgrade_expression` and `upgrade_opt_expression`.

jimblandy: That could be. Take a look at what the `valid` module does with `GlobalVariable::init`. You should be able to tell whether it will treat an `atomic<T>` variable like a `T` variable.

jimblandy: (I often resort to reading `naga/src/valid` when I want to understand the expectations of Naga IR. Ideally each time I did some exploration I would add docs to `lib.rs`...)

jimblandy: From chat:

> It seems like validate_global_var, at least, does not allow atomic globals to have initializers at all. Initializers are only allowed in the private and function address spaces, and atomics are not allowed in either of those.
>
> It seems that WGSL also only permits initializers for variables in the private or function address spaces, so that bit of Naga is faithfully enforcing WGSL's rules.
>
> For the sake of incremental development, it would probably make sense to just forbid initializers for now.

naga/src/front/spv/mod.rs
```diff
@@ -560,25 +561,40 @@ struct BlockContext<'function> {
...
 
+impl<'a> BlockContext<'a> {
+    /// Recurse into the expression with the given handle, locating any contained
+    /// global variables.
+    ///
+    /// This is used to track atomic upgrades.
+    fn get_contained_global_variables(
+        &self,
+        handle: Handle<crate::Expression>,
+    ) -> IndexSet<Handle<crate::GlobalVariable>> {
```
jimblandy: This is more complicated than you need. You never need to look at `index` operands, so there's only ever going to be one global variable, at the most. This function can simply be a loop, not recursive, and it can simply return an `Option<Handle<GlobalVariable>>`.

naga/src/front/spv/mod.rs
```diff
@@ -560,25 +561,40 @@ struct BlockContext<'function> {
...
+        let mut s = IndexSet::default();
+        if let Ok(expr) = self.expressions.try_get(handle) {
+            #[allow(clippy::pattern_type_mismatch)]
```
jimblandy: I know it's weird, but please go ahead and use `*expr`. It's Naga style. Maybe it will change in the future, but it's not going to change in this PR.

jimblandy (commented):
Two drive-by comments:

naga/src/front/atomic_upgrade.rs
```diff
@@ -0,0 +1,167 @@
...
+                stride,
+            },
+            TypeInner::Struct { mut members, span } => TypeInner::Struct {
+                members: {
+                    // In the future we should have to figure out which member needs
+                    // upgrading, but for now we'll only cover the single-member case
+                    if members.len() == 1 {
+                        members[0].ty = self.upgrade_type(members[0].ty)?;
+                    }
+                    members
```
schell: We still have to handle the single-member case, since most types end up bundled as members in structs, it seems.

naga/src/front/spv/mod.rs
```diff
@@ -560,25 +561,44 @@ struct BlockContext<'function> {
...
+    /// Descend into the expression with the given handle, locating a contained
+    /// global variable.
+    ///
+    /// This is used to track atomic upgrades.
+    fn get_contained_global_variable(
+        &self,
+        mut handle: Handle<crate::Expression>,
+    ) -> Option<Handle<crate::GlobalVariable>> {
+        log::debug!("\t\tlocating global variable in {handle:?}");
+        while let Ok(expr) = self.expressions.try_get(handle) {
```
jimblandy: As before, it's fine to assume that handles are valid in the front end. So this can simply be a `loop` that does
```
match self.expressions[handle]
```

naga/src/front/spv/mod.rs
```diff
@@ -590,8 +610,10 @@ pub struct Frontend<I> {
     future_member_decor: FastHashMap<(spirv::Word, MemberIndex), Decoration>,
     lookup_member: FastHashMap<(Handle<crate::Type>, MemberIndex), LookupMember>,
     handle_sampling: FastHashMap<Handle<crate::GlobalVariable>, image::SamplingFlags>,
-    // Used to upgrade types used in atomic ops to atomic types, keyed by pointer id
-    lookup_atomic: FastHashMap<spirv::Word, AtomicOp>,
+    /// An ordered set of any global variables that contain [`Atomic`](crate::Statement::Atomic)
+    /// statements we've generated, so we can upgrade the types of their operands.
```
jimblandy: Hmm, this looks like chimera text: global variables don't contain statements.

jimblandy (approved):
Okay, I've pushed a few more commits here. Please look them over, and if they seem all right, then we can land this.

Be sure to squash the changes when you land it!

jimblandy:
Oh, um, I guess I should actually fix my review comments if I'm going to approve this

schell (commented):
@jimblandy these changes all make sense to me. Thank you!

jimblandy:
Squashed and rebased.

Fixed up some `cargo doc` issues that I thought should have been caught by CI.

PR #5776, opened 2024-06-05
Fix double decoration if a binding array contains a struct with a runtime array

### LINKED ISSUE #5755: Binding array containing a struct with a runtime array causes invalid spirv
**Description**
When converting a wgsl shader with a binding array that contains a struct with a runtime array causes the spirv to have two `OpDecorate Block`s.
This is against the spirv specification https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Decoration and a similar (but much longer) piece of code causes a validation error
````cmd
VUID-VkShaderModuleCreateInfo-pCode-08737(ERROR / SPEC): msgNum: -1520283006 - Validation Error: [ VUID-VkShaderModuleCreateInfo-pCode-08737 ] | MessageID = 0xa5625282 | vkCreateShaderModule(): pCreateInfo->pCode (spirv-val produced an error):
ID '17' decorated with Block multiple times is not allowed.
  %MaterialsIdx = OpTypeStruct %_runtimearr_uint
. The Vulkan spec states: If pCode is a pointer to SPIR-V code, pCode must adhere to the validation rules described by the Validation Rules within a Module section of the SPIR-V Environment appendix (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkShaderModuleCreateInfo-pCode-08737)
````

**Repro steps**
converting this shader to spirv causes invalid spirv to generate
````wgsl
struct Array {
    arr: array<u32>,
}

@group(0) @binding(0)
var<storage> store: binding_array<Array>;
````

Naga should not generate invalid spirv

**Extra materials**
seems to be caused by 
https://github.com/gfx-rs/wgpu/blob/60a14c67fb43e68b7e64bee45ab3fac3f4253e72/naga/src/back/spv/writer.rs#L1045-L1047 and https://github.com/gfx-rs/wgpu/blob/60a14c67fb43e68b7e64bee45ab3fac3f4253e72/naga/src/back/spv/writer.rs#L1762-L1765

The invalid spirv generated
image

I'm not sure how this could be fixed, possibly with a hash-map of the id or a marker on the type id?

**Platform**
trunk as of 60a14c67fb43e68b7e64bee45ab3fac3f4253e72

Vecvec:
**Connections**
Fixes #5755 

**Description**
Previously if a binding array contained a struct with a runtime array it would generate two `OpDecorate Block`s which was not allowed by the spirv specification.

**Testing**
This shader no longer produces two `OpDecorate Block`s
````wgsl
struct Array {
    arr: array<u32>,
}

@group(0) @binding(0)
var<storage> store: binding_array<Array>;
````
and a similar one inside naga's tests (binding-buffer-arrays.spvasm) also no longer produces two `OpDecorate Block`s

**Checklist**

- [x] Run `cargo fmt`.
  - `--target wasm32-unknown-unknown` not applicable
  - `--target wasm32-unknown-emscripten`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -131,6 +131,10 @@ By @stefnotch in [#5410](https://github.com/gfx-rs/wgpu/pull/5410)
 
 - Added support for pipeline-overridable constants to the WebGPU backend by @DouglasDwyer in [#5688](https://github.com/gfx-rs/wgpu/pull/5688)
 
+#### Naga
+
+- In spv-out don't decorate a `BindingArray`'s type with `Block` if the type is a struct with a runtime array by @Vecvec in [#5776](https://github.com/gfx-rs/wgpu/pull/5776)
+
 ## v0.20.0 (2024-04-28)
 
 ### Major Changes
--- a/naga/src/back/spv/writer.rs
+++ b/naga/src/back/spv/writer.rs
@@ -1768,7 +1768,11 @@ impl Writer {
                         if let crate::TypeInner::Struct { members, .. } = &ty.inner {
                             // only the last member in a struct can be dynamically sized
                             if let Some(last_member) = members.last() {
-                                if let crate::TypeInner::Array { size: crate::ArraySize::Dynamic, .. } = &ir_module.types[last_member.ty].inner {
+                                if let crate::TypeInner::Array {
+                                    size: crate::ArraySize::Dynamic,
+                                    ..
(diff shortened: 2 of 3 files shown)


teoxoy (approved):
Thanks!

kvark:
@teoxoy are we planning to have this fix on crates soon?

cwfitzgerald:
This will land during the next wgpu release, 0.22, which is scheduled for next Wednesday.

PR #5778, opened 2024-06-05
Add a function to clear bindgroups from passes and bundles, with stub backend implementations.

bradwerth:
Since the spec allows null BindGroups to be passed to setBindGroup, we need a way to clear the BindGroup associated with an index. This patch implements that through a clear_bind_group function defined in various necessary places, and implements as much of it as possible, without getting into the wgpu-hal backend implementations, which are left as TODOs.

**Connections**
N/A

**Testing**
This code is not exercised, since all of our set_bind_group entry points require a BindGroup. The CTS contains tests that set null BindGroups.

**Checklist**

- [X] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ TODO(wumpf): This is still work in progress. Should write a bit more about it. A
 
 `wgpu::ComputePass` recording methods (e.g. `wgpu::ComputePass:set_render_pipeline`) no longer impose a lifetime constraint passed in resources.
 
-Furthermore, you can now opt out of `wgpu::ComputePass`'s lifetime dependency on its parent `wgpu::CommandEncoder` using `wgpu::ComputePass::forget_lifetime`:  
+Furthermore, you can now opt out of `wgpu::ComputePass`'s lifetime dependency on its parent `wgpu::CommandEncoder` using `wgpu::ComputePass::forget_lifetime`:
 ```rust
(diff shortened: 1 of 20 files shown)


ErichDonGubler:
I have several questions before I'm willing to dive into a full review of the implementation:

1. Is this actually shippable without having `hal` implementations?
2. Is there a reason that we shouldn't adhere more to the design of the JS API, specifically WRT `GPUDevice.setBindGroup` expressing this use case as an optional `bindGroup` argument, rather than a separate method?

bradwerth:
> I have several questions before I'm willing to dive into a full review of the implementation:
> 
>     1. Is this actually shippable without having `hal` implementations?

Probably not. Undoing the effects of a setBindGroup are subtle and require expertise per-backend; expertise I don't have.

>     2. Is there a reason that we shouldn't adhere more to the design of the JS API, specifically WRT `GPUDevice.setBindGroup` expressing this use case as an optional `bindGroup` argument, rather than a separate method?

Modifying the top-level entry point would be superior I think, but it would break all usage with the change, since every caller of setBindGroup would need to be modified. I also wasn't sure which of the many implementations of setBindGroup is the topmost, external-facing one. For those reasons, I offered the implementation up in this form.

bradwerth:
I'll create a new PR that changes the external-facing signature of `setBindGroup` to take an `Option<bindGroupId>`.

bradwerth:
The new PR is https://github.com/gfx-rs/wgpu/pull/6216.

PR #5780, opened 2024-06-06
[wgpu] Remove unused `ObjectId` field of `SubmissionIndex`.

jimblandy:
This removes warnings that field 0 of `SubmissionIndex` is never used.

I only see these warnings when referring to `wgpu` via a `path` dependency, not when fetching it from `crates.io`, and I don't really understand why. I wonder if this is related to https://github.com/rust-lang/cargo/issues/8546.

---

Drop the first field of `wgpu::SubmissionIndex`, as it is unused. Adjust uses of the types' other field.

Delete the `wgpu::context::Context` trait's `SubmissionIndex` associated type. Remove it from implementations.

Drop the first field of the `Context` trait's `queue_submit` method's return value. Adjust implementations.

### DIFF
--- a/wgpu/src/backend/webgpu.rs
+++ b/wgpu/src/backend/webgpu.rs
@@ -1158,7 +1158,6 @@ impl crate::context::Context for ContextWebGpu {
     type SurfaceData = Sendable<(Canvas, webgpu_sys::GpuCanvasContext)>;
 
     type SurfaceOutputDetail = SurfaceOutputDetail;
-    type SubmissionIndex = Unused;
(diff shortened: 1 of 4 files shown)


jimblandy:
I'm not sure about this PR at all. It seems like rustc or cargo are confused about something.

ErichDonGubler (approved):

ErichDonGubler:
IIRC, this also has something to do with the toolchain version? Not sure about the specifics, but I don't see a reason to keep an unused field around, in general. 🤷🏻‍♂️

PR #5785, opened 2024-06-09
[naga] Add snapshot test output omitted from #5702.

jimblandy:
cc @schell


teoxoy (approved):

teoxoy:
Hmm, I thought CI would catch these.

PR #5791, opened 2024-06-10
Compile with context

### LINKED ISSUE #5713: naga - compile with context

**Background - naga_oil**
I have been working on [naga_oil](https://github.com/bevyengine/naga_oil), which is a WGSL preprocessor, for a while. One important feature that naga_oil implements is shader imports.

For shader imports, we independently compile shaders using naga, and then correctly link them together. This works nicely, but comes with the limitation that we need to generate "header files" for the naga compiler.
For example, when we want to compile
```c
// cat.wgsl
#import foo::Bar;

struct Cat {
  a: Bar;
}
```
we first look at the `foo` file, and generate code along the lines of
```c
// cat.wgsl
// foo.wgsl header
struct Something { hi: f32 }
struct Bar { baz: f32, henlo: Something }

struct Cat {
  a: Bar;
}
```

And this scales really badly. As the import tree gets larger, we can end up with massive shaders that end up including almost everything else, just in case the compilation needs it.

So we're doing
1. Compile foo.wgsl to a naga module
2. Extract a header, and generate WGSL code 
3. Add the header to cat.wgsl, and compile that to a naga module
and repeat that, until we've compiled all shaders

It would be much nicer if we could do
1. Compile foo.wgsl to a naga module
2. Use the existing naga module to compile cat.wgsl

This request is asking for a way of pre-populating a naga module with some entries. I think this would be a useful building block for other tools, such as naga_oil.

Writing our own WGSL parser for performance, or accepting O(n^2) compile times hit for Bevy's shaders.

**Additional context**
There might also additional use cases for this, such as incremental compilation. But that's a different story.

stefnotch:
**Connections**
Fix #5713 

**Description**
This makes it possible to take a compiled module, and use it as a "base" for compiling another module. The semantics is
1. Compile the base module. Assume it exports every global (function, struct, alias, etc).
2. Compile the next module, and give it every global from the base module.

To only make individual items importable, one can first pre-process the base module. 
1. Generate UUID
2. Iterate over every global in the base module, and rename it to "UUID_name_of_the_global".

This particular implementation avoids exposing internals that weren't exposed before.

**Testing**
I added a comprehensive list of tests to naga/src/front/wgsl/tests.rs . There are 3 commented out tests that I would like to add, but currently cannot, due to other naga limitations. How should I best deal with those? 

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,27 @@ By @stefnotch in [#5410](https://github.com/gfx-rs/wgpu/pull/5410)
 
 - Implement `WGSL`'s `unpack4xI8`,`unpack4xU8`,`pack4xI8` and `pack4xU8`. By @VlaDexa in [#5424](https://github.com/gfx-rs/wgpu/pull/5424)
 - Began work adding support for atomics to the SPIR-V frontend. Tracking issue is [here](https://github.com/gfx-rs/wgpu/issues/4489). By @schell in [#5702](https://github.com/gfx-rs/wgpu/pull/5702).
+- Compile shaders with a "base module". This allows for cleanly implementing imports, or shadertoy-like environments. By @stefnotch in [#5791](https://github.com/gfx-rs/wgpu/pull/5791).
+  ```rust
+  use naga::front::wgsl::Frontend;
+
+  let base_module = Frontend::new()
+      .parse("
+      fn main_image(frag_coord: vec2f) -> vec4f {
+          return vec4f(sin(frag_coord.x), cos(frag_coord.y), 0.0, 1.0);
+      }").unwrap();
+  // For bonus points: Process the base_module to rename all globals except for `main_image`.
+  let result = Frontend::new()
+      .parse_to_ast("
+      @fragment
+      fn fs_main(@builtin(position) pos : vec4f, @location(0) uv: vec2f) -> vec4f {
+          return main_image(uv);
+      }")
+      .unwrap()
+      .to_module(Some(&base_module))
+      .unwrap();
+  ```
+
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,7 +104,7 @@ By @stefnotch in [#5410](https://github.com/gfx-rs/wgpu/pull/5410)
       fn main_image(frag_coord: vec2f) -> vec4f {
           return vec4f(sin(frag_coord.x), cos(frag_coord.y), 0.0, 1.0);
       }").unwrap();
-  // For bonus points: Process the base_module to rename all globals except for `main_image`.
+  // A full shadertoy implementation would rename all globals from base_module, except for `main_image`.
   let result = Frontend::new()
       .parse_to_ast("
       @fragment
--- a/naga/src/front/wgsl/lower/mod.rs
+++ b/naga/src/front/wgsl/lower/mod.rs
@@ -974,12 +974,12 @@ impl<'source, 'temp> Lowerer<'source, 'temp> {
                 ast::GlobalDeclKind::Type(ref t) => t.name,
             };
             if let Some(old) = ctx.globals.get(ident.name) {
-                let span = match old {
-                    LoweredGlobalDecl::Function(handle) => ctx.module.functions.get_span(*handle),
-                    LoweredGlobalDecl::Var(handle) => ctx.module.global_variables.get_span(*handle),
-                    LoweredGlobalDecl::Const(handle) => ctx.module.constants.get_span(*handle),
-                    LoweredGlobalDecl::Override(handle) => ctx.module.overrides.get_span(*handle),
-                    LoweredGlobalDecl::Type(handle) => ctx.module.types.get_span(*handle),
+                let span = match *old {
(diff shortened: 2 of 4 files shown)


naga/src/front/wgsl/lower/mod.rs
```diff
@@ -925,6 +926,70 @@ impl<'source, 'temp> Lowerer<'source, 'temp> {
...
+                if let Some(name) = t.name.as_ref() {
+                    ctx.globals.insert(name, LoweredGlobalDecl::Type(handle));
+                }
+            }
+            for entry_point in base_module.entry_points.iter() {
+                ctx.globals
+                    .insert(entry_point.name.as_str(), LoweredGlobalDecl::EntryPoint);
+            }
+            *ctx.global_expression_kind_tracker =
+                crate::proc::ExpressionKindTracker::from_arena(&ctx.module.global_expressions);
```
stefnotch, line 963: Restoring the context. Most parts of the lowerer's context do not need to be restored, from what I could gather.

naga/src/front/wgsl/mod.rs
```diff
@@ -32,18 +34,46 @@ impl Frontend {
...
+            index,
+        })
     }
 }
 
+pub struct ParsedWgsl<'a> {
+    source: &'a str,
+    translation_unit: TranslationUnit<'a>,
+    index: index::Index<'a>,
+}
```
stefnotch, line 61: This intermediate struct lets us repeatedly compile a module with different base modules. That's something that naga_oil may want, hence this API.

jimblandy:
@stefnotch What is this feature used for, exactly?

Where is this used in production?

CHANGELOG.md
```diff
@@ -95,6 +95,27 @@ By @stefnotch in [#5410](https://github.com/gfx-rs/wgpu/pull/5410)
 
 - Implement `WGSL`'s `unpack4xI8`,`unpack4xU8`,`pack4xI8` and `pack4xU8`. By @VlaDexa in [#5424](https://github.com/gfx-rs/wgpu/pull/5424)
 - Began work adding support for atomics to the SPIR-V frontend. Tracking issue is [here](https://github.com/gfx-rs/wgpu/issues/4489). By @schell in [#5702](https://github.com/gfx-rs/wgpu/pull/5702).
+- Compile shaders with a "base module". This allows for cleanly implementing imports, or shadertoy-like environments. By @stefnotch in [#5791](https://github.com/gfx-rs/wgpu/pull/5791).
+  ```rust
+  use naga::front::wgsl::Frontend;
```
jimblandy, line 100: We can't put documentation in `CHANGELOG.md`. If the feature is to be part of Naga, it has to be documented properly, the way we try to do for the rest of Naga.

jimblandy:
I think the main problem this PR is trying to address is that the WGSL front end really only operates on complete modules. If all the definitions aren't present, then the front end can't even lower the `ast` form to a `Module`. In fact, Naga's `Module` type itself is designed to ensure that all references to types, functions, variables, etc. are fully resolved and point directly to their definitions, which makes it useless for representing chunks of code that haven't yet been combined into a complete program.

If someone wants to experiment with WGSL module designs, they'd like to be able to borrow Naga's WGSL parser, rather than rewriting one from scratch, but since `naga::front::wgsl` wants to run `Index::generate` and figure out what everything refers to, that's kind of hard. It's a real impediment to experimentation.

My concern about this PR is that it doesn't seem like it really accomplishes much more than:
- validating the base module to make sure it's valid WGSL
- concatenating the base module source code with more code and validating again

In other words, it saves you having to re-parse the module, but it doesn't really enable new ways of using the code.

It's got a bunch of other issues, too, like, Naga spans don't have source ids, so if you have multiple source files Naga has no way to indicate which one a given span belongs to.

So although I think that this is an absolutely fantastic avenue for experimentation, and definitely a direction that's worth exploring, as it stands this doesn't look like something that is mature enough to merge into trunk.

If you would like to have a branch in `gfx-rs/wgpu` so that you can collaborate with others as you experiment with this design and others, that's definitely something we would be happy to do.

I definitely think we need to work with Bevy and others to understand what they're doing with their templating system, and set up some kind of process where we can iterate with developers to extend WGSL to better support complex shader development. It's something the WebGPU standards committee members also have very much on their minds, it came up in our office hours hangouts. But people have widely varying ideas about what support for complex shaders should look like; one person felt that C++-style [`if constexpr`](https://en.cppreference.com/w/cpp/language/if) was the way forward, whereas I thought something more like traits and generics would serve better.

The only way to answer these questions is by trying stuff out. Even very experienced people make lousy decisions trying to think it through in their heads. The way I see it, C was designed by a process of mutual evolution with the Unix kernel, pre-1.0 Rust was designed via a process of mutual evolution with Servo, and maybe Modular WGSL can be designed by a co-evolution process with Bevy.

stefnotch:
@jimblandy Sorry for the late response. First of all, thank you for the really good feedback. I didn't even think about the spans remapping, that's definitely a tough problem.

I would love to have a branch on this repository for collaboration and experiments. Now that I have actually tried out my code changes in a fork of naga_oil, I very much see all of the above mentioned limitations of this design (and some more). So a iterative "try out things and communicate" approach sounds like an excellent idea.

PR #5792, opened 2024-06-10
[ci] check for untracked naga snapshots

teoxoy:
`git diff` doesn't check untracked files, we need to stage those then compare with `HEAD`.

Avoids having to manually catch those issues (https://github.com/gfx-rs/wgpu/pull/5785).


cwfitzgerald (approved):
Could you just add a comment as to why we're doing this to the file, then g2g

cwfitzgerald:
Heh it found an issue lol

PR #5799, opened 2024-06-11
Deduplicate submitted surfaces to avoid a vulkan deadlock/crash

### LINKED ISSUE #5774: Crash on vulkan when multiple command encoders are used after #5681.
**Description**
Since https://github.com/gfx-rs/wgpu/pull/5681, wgpu now crashes with "More than one surface texture is being used from the same swapchain. This will cause a deadlock in release." if more than 1 command encoder (with a renderpass that has a texture view of your `Surface` texture) is used.

**Repro steps**
Create and submit a 2nd command encoder with a renderpass that uses a view of your surface texture (can create a new texture view or just re-use an existing one) when running on Vulkan.
Run https://github.com/Elabajaba/wgpu-vk-surface-crash

Expected: Doesn't crash
Observed: 
```
thread 'main' panicked at E:\temp\wgpu\wgpu-hal\src\vulkan\mod.rs:966:9:
More than one surface texture is being used from the same swapchain. This will cause a deadlock in release.
```

**Platform**
Happens on both Windows 11 + AMD gpu and Linux+Mesa RADV (amd) + Wayland.

Elabajaba:
**Connections**
Fixes https://github.com/gfx-rs/wgpu/issues/5774

**Description**
wgpu-hal's vulkan implementataion requires that we only submit each used surface texture once. If we submit the same surface texture multiple times, it'll deadlock.

To fix this, use a hashmap keyed on the surface texture arc's raw pointers instead of a smallvec to deduplicate the surface textures before we submit them.

**Testing**
Tested against https://github.com/Elabajaba/wgpu-vk-surface-crash

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/device/queue.rs
+++ b/wgpu-core/src/device/queue.rs
@@ -1448,7 +1448,7 @@ impl Global {
                             if raw.is_some() {
                                 // Compare the Arcs by pointer as Textures don't implement Eq
                                 submit_surface_textures_owned
-                                    .insert(Arc::as_ptr(&texture), texture.clone());
+                                    .insert(Arc::as_ptr(texture), texture.clone());
                             }
 
                             unsafe {
@@ -1493,7 +1493,7 @@ impl Global {
             let mut submit_surface_textures =
                 SmallVec::<[_; 2]>::with_capacity(submit_surface_textures_owned.len());
 
-            for (_, texture) in &submit_surface_textures_owned {
+            for texture in submit_surface_textures_owned.values() {
                 submit_surface_textures.extend(match texture.inner.get(&snatch_guard) {
                     Some(TextureInner::Surface { raw, .. }) => raw.as_ref(),
                     _ => None,
--- a/wgpu-core/src/device/queue.rs
+++ b/wgpu-core/src/device/queue.rs
@@ -1172,7 +1172,9 @@ impl Global {
 
             let mut used_surface_textures = track::TextureUsageScope::default();
 
-            let mut submit_surface_textures_owned = SmallVec::<[_; 2]>::new();
+            // Use a hashmap here to deduplicate the surface textures that are used in the command buffers.
+            // This avoids vulkan deadlocking from the same surface texture being submitted multiple times.
+            let mut submit_surface_textures_owned = FastHashMap::default();
 
             {
                 let mut command_buffer_guard = hub.command_buffers.write();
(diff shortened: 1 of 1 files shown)


wgpu-core/src/device/queue.rs
```diff
@@ -1263,7 +1265,9 @@ impl Global {
                                         Some(TextureInner::Native { .. }) => false,
                                         Some(TextureInner::Surface { ref raw, .. }) => {
                                             if raw.is_some() {
-                                                submit_surface_textures_owned.push(texture.clone());
+                                                // Compare the Arcs by pointer as Textures don't implement Eq.
+                                                submit_surface_textures_owned
+                                                    .insert(Arc::as_ptr(&texture), texture.clone());
```
cwfitzgerald, line 1270: Good thought on Arcs, just wondering if we should use the indexes already built into the texture.info.index or something like that.

wgpu-core/src/device/queue.rs
```diff
@@ -1172,7 +1172,9 @@ impl Global {
 
             let mut used_surface_textures = track::TextureUsageScope::default();
 
-            let mut submit_surface_textures_owned = SmallVec::<[_; 2]>::new();
+            // Use a hashmap here to deduplicate the surface textures that are used in the command buffers.
+            // This avoids vulkan deadlocking from the same surface texture being submitted multiple times.
+            let mut submit_surface_textures_owned = FastHashMap::default();
```
Elabajaba, line 1177: Should this be an IndexMap instead? I'm not sure if the ordering matters for it.

wgpu-core/src/device/queue.rs
Elabajaba: > Good thought on Arcs

I copied what you did in vulkan hal's `Queue::submit` lol

> just wondering if we should use the indexes already built into the texture.info.index

Isn't wgpu trying to remove all those indices and move to just using Arcs?

cwfitzgerald: Those indices are staying

wgpu-core/src/device/queue.rs
cwfitzgerald: No ordering doesn't matter, just uniqueness.

wgpu-core/src/device/queue.rs
cwfitzgerald: Let's leave this for now, for the sake of expediency.

cwfitzgerald (approved):
Thanks!

PR #5804, opened 2024-06-12
[core] Improve resource and api logging.

jimblandy:
Marked as draft, because this depends on #5802.

- Improve logging in `StatelessTracker::remove_abandoned` to show the
  outcome of the call.

- Add similar logging to `BufferTracker` and `TextureTracker`.

- Let `device_create_buffer`'s log only the new buffer's label, id,
  and whether it's mapped at creation. It used to show the whole
  descriptor, which is too much detail.

- Have `queue_submit` log the submission id, and have `device_poll`
  log what it was waiting for, and what actually got done.

- Have `Device::drop` log the destruction of the raw device when it
  actually happens, so it's properly ordered with respect to logging
  from other parts of the device, like `Device::command_allocator`.

### DIFF
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -260,7 +260,15 @@ impl Global {
             };
 
             let (id, resource) = fid.assign(Arc::new(buffer));
-            api_log!("Device::create_buffer({desc:?}) -> {id:?}");
+            api_log!(
+                "Device::create_buffer({:?}{}) -> {id:?}",
+                desc.label.as_deref().unwrap_or(""),
+                if desc.mapped_at_creation {
+                    ", mapped_at_creation"
+                } else {
+                    ""
+                }
+            );
 
             device
                 .trackers
@@ -2172,7 +2180,7 @@ impl Global {
         device_id: DeviceId,
         maintain: wgt::Maintain<queue::WrappedSubmissionIndex>,
     ) -> Result<bool, WaitIdleError> {
-        api_log!("Device::poll");
+        api_log!("Device::poll {maintain:?}");
 
         let hub = A::hub(self);
         let device = hub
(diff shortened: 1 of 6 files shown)


JMS55:
While you're here, maybe you feel like adding a cargo feature flag to remove logs statically? Iirc they have a fair amount of overhead just for checking if logging is enabled or not.

jimblandy:
> While you're here, maybe you feel like adding a cargo feature flag to remove logs statically? Iirc they have a fair amount of overhead just for checking if logging is enabled or not.

Sorry, I can't - this was just a drive-by.

Wumpf:
tbh I doubt that the checks on the log flags are significant at all given all the other things that happen in those methods, but if this actually turns out to be a measurable overhead the macros/feature flags here can be expanded accordingly
https://github.com/jimblandy/wgpu/blob/c7a16b36b16be7484c2306d1155705f3f6241bf4/wgpu-core/src/lib.rs#L309
(not in this pr obviously ;-))

wgpu-core/src/device/global.rs
```diff
@@ -260,7 +260,15 @@ impl Global {
             };
 
             let (id, resource) = fid.assign(Arc::new(buffer));
-            api_log!("Device::create_buffer({desc:?}) -> {id:?}");
+            api_log!(
+                "Device::create_buffer({:?}{}) -> {id:?}",
+                desc.label.as_deref().unwrap_or(""),
+                if desc.mapped_at_creation {
+                    ", m-a-c"
```
Wumpf: I'd never be able to figure out that `m-a-c` means "mapped at creation", why not spell it out

Wumpf (approved):

wgpu-core/src/device/global.rs
jimblandy: That's more than fair.

nical:
> While you're here, maybe you feel like adding a cargo feature flag to remove logs statically? Iirc they have a fair amount of overhead just for checking if logging is enabled or not.

`api_logs` are trace-level (the most verbose level) by default. Apps that depend on wgpu can filter out the logs at compile time by setting the appropriate [feature flags on the log crate](https://docs.rs/log/latest/log/#compile-time-filters).

PR #5806, opened 2024-06-13
[hal] Document buffer mapping and coherence.

jimblandy:


wgpu-hal/src/lib.rs
```diff
@@ -663,16 +661,80 @@ pub trait Device: WasmNotSendSync {
         desc: &BufferDescriptor,
     ) -> Result<<Self::A as Api>::Buffer, DeviceError>;
     unsafe fn destroy_buffer(&self, buffer: <Self::A as Api>::Buffer);
+
+    /// Return a pointer to CPU memory mapping the contents of `buffer`.
+    ///
+    /// Buffer mappings are persistent: the buffer may remain mapped on the CPU
+    /// while the GPU reads or writes to it.
```
cwfitzgerald: I'm not sure we actually use persistent mapping - I just took a look in wgpu-core and we only ever call hal map when the gpu is finished, and then is truly unmapped when the user unmaps it.

I'm not positive the GL can provide persistent mapping

cwfitzgerald (changes requested):
One comment, the rest looks good.

wgpu-hal/src/lib.rs
jimblandy: Interesting. This is something that I picked up from the original wgpu-hal comments, that tried to summarize the entire design in ten lines. So I just went with it.

jimblandy: I think the `wgpu_hal` documentation needs to document `wgpu_hal`, not `wgpu_core`'s use of it. We can certainly have a note saying, BTW, `wgpu_core` just unmaps everything, but if it's a fact that `wgpu_hal` supports persistent mapping, then that's what its docs should say.

And as far as I can see, it does. The `gles` backend [checks for the `GL_EXT_buffer_storage` extension](https://github.com/atlv24/wgpu/blob/4902e470ce61bd102a7d57f3e17bf15798095b6a/wgpu-hal/src/gles/adapter.rs#L584-L588) and if it is not available, [enables some sort of emulation](https://github.com/atlv24/wgpu/blob/4902e470ce61bd102a7d57f3e17bf15798095b6a/wgpu-hal/src/gles/device.rs#L532-L535) with a heap-allocated block of bytes. And [`wgpu_hal::gles::Api::Device`'s implementation of `wgpu_hal::Device::flush_mapped_ranges`](https://github.com/atlv24/wgpu/blob/4902e470ce61bd102a7d57f3e17bf15798095b6a/wgpu-hal/src/gles/device.rs#L709) calls [`glFlushMappedBufferRange`](https://registry.khronos.org/OpenGL-Refpages/es3.0/html/glFlushMappedBufferRange.xhtml).

So it seems pretty clear that `wgpu_hal` does behave the way described in this PR, even on `gles`, and even if `wgpu_core` doesn't use it that way.

cwfitzgerald (approved):

PR #5811, opened 2024-06-13
refactor: satisfy `clippy::never_loop`

ErichDonGubler:
Recommended review experience: commit-by-commit.

**Connections**

None of note.

**Description**

It turns out that many usages of `loop` in `wgpu-core` were written to access a `break` flow, but _not_ to actually jump to the beginning of a block again, i.e.:

```rust
let val = 'b: loop {
    if fails_validation(thing) {
        break 'b Err(…);
    }
    break 'b Ok(…);
};
```

This is the case for much validation code, seemingly as a stable alternative to the nascent [`try` block syntax](https://doc.rust-lang.org/beta/unstable-book/language-features/try-blocks.html). This particular usage of `loop` has been superseded (for a long time) by labeled scopes, which can also `break`, like so:

```rust
let val = 'b: {
    if fails_validation(thing) {
        break 'b Err(…);
    }
    Ok(…)
};
```

This is easier to read and understand, _and_ avoids a footgun of actually looping when we have no intention to do so.

**Testing**

I rest my case with CI's verdict. 🙂

**Checklist**

- [x] Run `cargo fmt`.
- [x] ~~Run `cargo clippy`. If applicable, add:~~
  - [x] ~~`--target wasm32-unknown-unknown`~~
  - [x] ~~`--target wasm32-unknown-emscripten`~~

### DIFF
--- a/naga/src/valid/interface.rs
+++ b/naga/src/valid/interface.rs
@@ -678,7 +678,7 @@ impl super::Validator {
         }
 
         {
-            let used_push_constants = module
+            let mut used_push_constants = module
                 .global_variables
                 .iter()
                 .filter(|&(_, var)| var.space == crate::AddressSpace::PushConstant)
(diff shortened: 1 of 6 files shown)


Vecvec:
Should 
https://github.com/gfx-rs/wgpu/blob/d7a35ecda06742acb4cdd5a83c5b9737f375a70b/wgpu-core/src/lib.rs#L23-L24
be removed, or is it required for other things?

ErichDonGubler:
@Vecvec: Actually, yes! I hadn't considered that we were silencing `never_loop`, and remember being puzzled yesterday that it wasn't firing in these cases. I should have investigated that, good catch.

ErichDonGubler:
Oh, sigh, b073744 ran into <https://github.com/rust-lang/rust-clippy/issues/12931>. Fun!

Wumpf (approved):
nice!

teoxoy (approved):

PR #5812, opened 2024-06-14
Do not feed &"" to D3DCompile

workingjubilee:
A recent change by rustc, now in 1.79-stable, makes empty str constants point to the same location: 0x01. This is an optimization of sorts, not stable behavior. Code must not rely on constants having stable addresses nor should it pass &"" to APIs expecting CStrs or NULL addresses. D3DCompile will segfault if you give it such a pointer, or worse: read random garbage addresses!

- Pass the NULL pointer to D3DCompile if wgpu lacks a decent CString.
- For {d,f}xc_compile, accept `source_name: Option<&CStr>`, as otherwise the API must be `unsafe`.

refs:
- https://learn.microsoft.com/en-us/windows/win32/api/d3dcompiler/nf-d3dcompiler-d3dcompile
- https://github.com/rust-lang/rust/issues/126442

### DIFF
--- a/wgpu-hal/src/dx12/shader_compilation.rs
+++ b/wgpu-hal/src/dx12/shader_compilation.rs
@@ -32,13 +32,21 @@ pub(super) fn compile_fxc(
     {
         compile_flags |= d3dcompiler::D3DCOMPILE_DEBUG | d3dcompiler::D3DCOMPILE_SKIP_OPTIMIZATION;
     }
+
+    // If no name has been set, D3DCompile wants the null pointer.
+    let source_name = if source_name.is_empty() {
+        ptr::null()
+    } else {
+        source_name.as_ptr()
+    };
+
     let mut error = d3d12::Blob::null();
     let hr = unsafe {
         profiling::scope!("d3dcompiler::D3DCompile");
         d3dcompiler::D3DCompile(
             source.as_ptr().cast(),
             source.len(),
-            source_name.as_ptr().cast(),
+            source_name.cast(),
             ptr::null(),
             ptr::null_mut(),
             raw_ep.as_ptr(),
--- a/wgpu-hal/src/dx12/shader_compilation.rs
+++ b/wgpu-hal/src/dx12/shader_compilation.rs
@@ -1,3 +1,4 @@
+use std::ffi::CStr;
 use std::ptr;
 
 pub(super) use dxc::{compile_dxc, get_dxc_container, DxcContainer};
@@ -14,7 +15,7 @@ use crate::auxil::dxgi::result::HResult;
 pub(super) fn compile_fxc(
     device: &super::Device,
     source: &str,
-    source_name: &str,
+    source_name: Option<&CStr>,
     raw_ep: &std::ffi::CString,
     stage_bit: wgt::ShaderStages,
     full_stage: String,
@@ -34,11 +35,7 @@ pub(super) fn compile_fxc(
     }
 
     // If no name has been set, D3DCompile wants the null pointer.
-    let source_name = if source_name.is_empty() {
-        ptr::null()
-    } else {
-        source_name.as_ptr()
-    };
+    let source_name = source_name.map(|cstr| cstr.as_ptr()).unwrap_or(ptr::null());
 
     let mut error = d3d12::Blob::null();
     let hr = unsafe {
@@ -86,6 +83,7 @@ pub(super) fn compile_fxc(
 // The Dxc implementation is behind a feature flag so that users who don't want to use dxc can disable the feature.
 #[cfg(feature = "dxc_shader_compiler")]
 mod dxc {
+    use std::ffi::CStr;
     use std::path::PathBuf;
 
     // Destructor order should be fine since _dxil and _dxc don't rely on each other.
@@ -140,7 +138,7 @@ mod dxc {
     pub(crate) fn compile_dxc(
         device: &crate::dx12::Device,
         source: &str,
-        source_name: &str,
+        source_name: Option<&CStr>,
         raw_ep: &str,
         stage_bit: wgt::ShaderStages,
         full_stage: String,
@@ -173,6 +171,10 @@ mod dxc {
             Ok(blob) => blob,
             Err(e) => return (Err(e), log::Level::Error),
         };
+
+        let source_name = source_name
+            .and_then(|cstr| cstr.to_str().ok())
+            .unwrap_or("");
 
         let compiled = dxc_container.compiler.compile(
             &blob,
@@ -271,6 +273,7 @@ mod dxc {
 // These are stubs for when the `dxc_shader_compiler` feature is disabled.
 #[cfg(not(feature = "dxc_shader_compiler"))]
 mod dxc {
+    use std::ffi::CStr;
     use std::path::PathBuf;
 
     pub(crate) struct DxcContainer {}
@@ -288,7 +291,7 @@ mod dxc {
     pub(crate) fn compile_dxc(
         _device: &crate::dx12::Device,
         _source: &str,
-        _source_name: &str,
+        _source_name: Option<&CStr>,
         _raw_ep: &str,
         _stage_bit: wgt::ShaderStages,
         _full_stage: String,
--- a/wgpu-hal/src/dx12/device.rs
+++ b/wgpu-hal/src/dx12/device.rs
@@ -259,12 +259,7 @@ impl super::Device {
             .as_ref()
             .map_err(|e| crate::PipelineError::Linkage(stage_bit, format!("{e}")))?;
 
-        let source_name = stage
-            .module
-            .raw_name
-            .as_ref()
-            .and_then(|cstr| cstr.to_str().ok())
-            .unwrap_or_default();
+        let source_name = stage.module.raw_name.as_deref();
 
         // Compile with DXC if available, otherwise fall back to FXC
         let (result, log_level) = if let Some(ref dxc_container) = self.dxc_container {


workingjubilee:
I had originally intended to test this more thoroughly but I stopped when I discovered I had to fuck around with patching multiple libraries, not *just* the Bevy monorepo, just to get the Bevy example in question to build.

wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -32,13 +32,21 @@ pub(super) fn compile_fxc(
     {
         compile_flags |= d3dcompiler::D3DCOMPILE_DEBUG | d3dcompiler::D3DCOMPILE_SKIP_OPTIMIZATION;
     }
+
+    // If no name has been set, D3DCompile wants the null pointer.
+    let source_name = if source_name.is_empty() {
+        ptr::null()
+    } else {
+        source_name.as_ptr()
+    };
```
RalfJung: This is still unsound though: `source_name` doesn't have a trailing 0. If D3DCompile expects a null-terminated string, then allocation is required here.

workingjubilee: > This is still unsound though: `source_name` doesn't have a trailing 0. If D3DCompile expects a null-terminated string, then allocation is required here.

Hmm, the source is a `CString` if it is not `&""`, so `D3DCompile` won't know the difference, and `DxcCompiler::compile` reallocates. I can of course alter the signature of `compile_fxc` and `compile_dxc`, but that seemed like a more invasive change.

workingjubilee: I do agree it is Morally Wrong, of course.

RalfJung: I deliberately used the term "unsound" -- the function is safe so it can't make assumptions about the callers, such as "this is secretly a CString". Also, if the caller guarantees that this is a CString, then your PR is unnecessary as a CString will always allocate at least one byte for the null terminator.

Also, under Stacked Borrows, if you get an `&str` then accessing the bytes outside the string length is UB.

workingjubilee: > Also, if the caller guarantees that this is a CString, then your PR is unnecessary as a CString will always allocate at least one byte for the null terminator.

It is an `Option<CString>`, just to make matters confusing!

RalfJung: Oh, the `&""` is hidden inside `unwrap_or_default`? I was wondering what the PR message talks about as there's no `&""` to be found in the code...

workingjubilee: YEP.

workingjubilee: I am slowly becoming of the opinion that `unwrap_or_default` was a mistake.

workingjubilee: @RalfJung I trust this passes your review with the new commit? I said I originally was reluctant to make the change but then I joined the Bevy Discord and saw there was a consensus of disapproving clucking regarding "passing a `&str` down a `&CStr`-shaped hole".

RalfJung: Without knowing any of the context, this looks right to me. All I can really check is whether this is the proper way to convert `Option<&CStr>` to a C pointer.

RalfJung: Using `&str` for the argument is also fine, but then it must allocate. As https://github.com/rust-lang/rust/issues/126442 shows, relying on the caller to uphold extra safety requirements (that are not documented!) is way too fragile (and violates Stacked Borrows).

wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -32,13 +33,17 @@ pub(super) fn compile_fxc(
...
     }
+
+    // If no name has been set, D3DCompile wants the null pointer.
+    let source_name = source_name.map(|cstr| cstr.as_ptr()).unwrap_or(ptr::null());
+
     let mut error = d3d12::Blob::null();
     let hr = unsafe {
         profiling::scope!("d3dcompiler::D3DCompile");
         d3dcompiler::D3DCompile(
             source.as_ptr().cast(),
```
msvbg, line 44: I realize this isn't where the segfault happens, but isn't this line relying on the same broken assumption?

RalfJung: This looks like a ptr+length kind of API, not something relying on the implicit null terminator?

workingjubilee: yeah.
```
HRESULT D3DCompile(
  [in]            LPCVOID                pSrcData,
  [in]            SIZE_T                 SrcDataSize,
```

teoxoy (approved):

teoxoy:
Thanks for the fix! We should backport this.

teoxoy:
Something else I noticed while looking at this is that `pSrcData` is documented as:

> A pointer to uncompiled shader data; either ASCII HLSL code or a compiled effect.

but we pass it UTF-8.

`pSourceName`, `pEntrypoint` and `pTarget` are also `LPCSTR` which is documented as:

> A pointer to a constant null-terminated string of 8-bit Windows (ANSI) characters. For more information, see [Character Sets Used By Fonts](https://learn.microsoft.com/en-us/windows/desktop/gdi/character-sets-used-by-fonts).

There is also [this page](https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page) which says you can change the default code page to UTF-8 and therefore rely on `LPCSTR` being UTF-8 but that's only possible for apps/binaries as far as I understand.

We can probably assume using ASCII for those is safe. Microsoft's docs don't seem to guarantee that those code pages are extended ASCII; [Wikipedia does though](https://en.wikipedia.org/wiki/Windows_code_page), maybe they went through [all the listed  ones](https://learn.microsoft.com/en-us/windows/win32/intl/code-page-identifiers) and checked that they are.

I will open a new issue to track this.

torokati44:
A new release with this fix would be greatly appreciated! :pray:

glandium:
I should probably open another issues, but I'll note here that the same function passes Rust Strings as C Strings, and that's undefined behavior, unless they have a NUL termination added on the Rust side.

jimblandy:
@glandium The arguments to `D3DCompile` are (per [docs](https://learn.microsoft.com/en-us/windows/win32/api/d3dcompiler/nf-d3dcompiler-d3dcompile)):

```c++
HRESULT D3DCompile(
  [in]            LPCVOID                pSrcData,
  [in]            SIZE_T                 SrcDataSize,
  [in, optional]  LPCSTR                 pSourceName,
  [in, optional]  const D3D_SHADER_MACRO *pDefines,
  [in, optional]  ID3DInclude            *pInclude,
  [in, optional]  LPCSTR                 pEntrypoint,
  [in]            LPCSTR                 pTarget,
  [in]            UINT                   Flags1,
  [in]            UINT                   Flags2,
  [out]           ID3DBlob               **ppCode,
  [out, optional] ID3DBlob               **ppErrorMsgs
);
```

So the textual arguments are as follows:

- The source code is passed via `pSrcData` and `SrcDataSize` as a pointer and byte length. This doesn't seem to require NUL termination.
- The filename `pSourceName` is a nul-terminated C string, which has been a `CString` for a long time.
- The entry point name `pEntryPoint` is constructed via `CString::new`, so it's NUL-terminated.
- The `pTarget` is constructed from a `String` whose creator has added an explicit `\0` character, which is gross but not wrong.

So it seems to me that all cases are now addressed. What specifically do you think is still wrong?

glandium-moz:
> The filename pSourceName is a nul-terminated C string, which has been a CString for a long time.

It was, in fact, changed to a CString in this PR. I had missed it.

> The pTarget is constructed from a String whose creator has added an explicit \0 character, which is gross but not wrong.

I must say I didn't look at callers to see what they were doing.

workingjubilee:
Hm, the `full_stage` should also be changed to be `&CStr` then.

RalfJung:
> The pTarget is constructed from a String whose creator has added an explicit \0 character, which is gross but not wrong.

That's still unsound in `compile_fxc`: the function relies on the caller to uphold certain invariants, or else there is UB. As this is a private function, the unsoundness is contained within the crate, but it is still generally considered a bug to have a private unsound function. You're needlessly giving up on having rustc help you with tracking the soundness requirements.

workingjubilee:
With #5836 `compile_fxc` has been made correct.

jimblandy:
> You're needlessly giving up on having rustc help you with tracking the soundness requirements.

I agree with all this - I didn't mean to suggest otherwise. I'm glad to see #5836.

jimblandy:
> It was, in fact, changed to a CString in this PR. I had missed it.

Again, I was looking at the actual origin of the data, which is this:

```
0a609e74c0 (Dzmitry Malyshau      2021-07-11 573) #[derive(Debug)]
0a609e74c0 (Dzmitry Malyshau      2021-07-11 574) pub struct ShaderModule {
0a609e74c0 (Dzmitry Malyshau      2021-07-11 575)     naga: crate::NagaShader,
726cf4f614 (Dzmitry Malyshau      2021-07-23 576)     raw_name: Option<ffi::CString>,
0a609e74c0 (Dzmitry Malyshau      2021-07-11 577) }
```

But yes, we agree that this PR fixed the actual value being passed to `D3DCompile`.

jimblandy:
Since I may have been misunderstood:

I think using terms like "morally wrong" for code like the original `compile_fxc` is fine as a joke - I'm sure I say stuff like that too, because I do think it's funny - but it may foster the impression that using argument types that properly capture the function's expectations is a matter of principle or style. The truth is that we have so much work to do in `wgpu` that we simply cannot afford the kind of distraction that comes from weird unstated safety contracts between functions and their callers.

A single instance of this kind of problem is not a big deal. A Rust update makes Bevy crash? Fine, we debug it, fix it, and move on. But if we make a practice of tolerating unsound code, we'll have not one but a steady stream of such distractions. And that will be a disaster, because `wgpu` will not be able to keep up with its users' expectations.

Properly used Rust is a huge improvement in *development velocity*, which we as a project can't afford to give up. People doing code review for `wgpu` should reject PRs that introduce unsoundness like that addressed in this PR and #5836, even when the code runs correctly.

PR #5818, opened 2024-06-14
[naga] Remove unneeded `PartialEq` derivations.

jimblandy:
Remove `PartialEq` derivations from various Naga IR types on which equality doesn't really make sense. In some cases, derive only in `cfg(test)` builds, so tests can check for expected output.

In the GLSL front end, use `append` to add new constants, not
`fetch_or_append`, since the latter requires `PartialEq` yet GLSL
doesn't permit duplicate declarations anyway.

### DIFF
--- a/naga/src/front/glsl/variables.rs
+++ b/naga/src/front/glsl/variables.rs
@@ -475,7 +475,7 @@ impl Frontend {
                     ty,
                     init,
                 };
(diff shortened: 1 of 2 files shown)


ErichDonGubler (approved):
lezduit

teoxoy (approved):

PR #5820, opened 2024-06-15
ensure that `wgpu::Error` is `Sync`

nolanderc:
**Connections**
N/A

**Description**
This PR adds a `Sync` bound to the boxed `std::error::Error` within `wgpu::Error`.
This makes it possible to wrap the error in `anyhow::Error` and `eyre::Report`, which require the inner error to be `Sync`.

**Testing**
There is a static assertion.

**Checklist**

- [x] Run `cargo fmt`.


ErichDonGubler (approved):
I see no reason this should be controversial, and it _is_ nice to be compatible with the current error-handling landscape recommended for binaries written in Rust.

PR #5824, opened 2024-06-16
spv-in parse more atomic ops

schell:
**Connections**
Part of
* https://github.com/gfx-rs/wgpu/issues/4489

Depends on
* https://github.com/gfx-rs/wgpu/pull/5775

**Description**

This adds parsing for the following atomic operations to `naga::spv::Frontend::next_block`: 

* [x] OpAtomicLoad
* [x] OpAtomicStore
* [x] OpAtomicExchange
* [ ] OpAtomicCompareExchange 
* [ ] ~OpAtomicCompareExchangeWeak~ (deprecated) 
* [ ] ~OpAtomicIIncrement~ was included in #5702
* [x] OpAtomicIDecrement
* [x] OpAtomicIAdd 
* [x] OpAtomicISub 
* [x] OpAtomicSMin 
* [x] OpAtomicUMin 
* [x] OpAtomicSMax 
* [x] OpAtomicUMax 
* [x] OpAtomicAnd 
* [x] OpAtomicOr 
* [x] OpAtomicXor 
* [x] OpAtomicFlagTestAndSet 
* [x] OpAtomicFlagClear

**Checklist**

- [x] Run `cargo fmt`.


jimblandy (commented):
The tests in `src/front/atomic_upgrade.rs` really belong in `src/front/spv`. We can't have SPIR-V specific code outside `src/front/spv` and `src/back/spv`.

naga/src/front/atomic_upgrade.rs
```diff
@@ -212,3 +212,82 @@ impl Module {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod test {
+    fn atomic_test(bytes: &[u8]) {
+        let _ = env_logger::builder().is_test(true).try_init();
```
jimblandy: Wouldn't it make more sense to `.expect` this `Result` than ignore it?

naga/src/front/spv/mod.rs
```diff
@@ -5710,32 +5964,44 @@ mod test {
...
-        log::info!("atomic_i_increment:\n{wgsl}");
+        ] {
+            let mut validator = crate::valid::Validator::new(vflags, Default::default());
+            match validator.validate(&m) {
+                Err(e) => {
+                    log::error!("SPIR-V validation {}", e.emit_to_string(""));
+                    should_panic = true;
+                }
+                Ok(i) => {
+                    wgsl = crate::back::wgsl::write_string(
```
jimblandy, line 5933: This won't compile if you only build with the `spv-in` feature, and not the `wgsl-out` feature, right? I think this function needs to be conditional on more cfg values. Note that this stuff changed recently; see the `cfg_aliases!` macro invocation in `naga/build.rs` for what you should use now.

naga/src/front/spv/mod.rs
```diff
@@ -5710,32 +5964,44 @@ mod test {
         let _ = super::parse_u8_slice(&bin, &Default::default()).unwrap();
     }
 
-    #[cfg(all(feature = "wgsl-in", wgsl_out))]
-    #[test]
-    fn atomic_i_inc() {
+    fn atomic_test(bytes: &[u8]) {
         let _ = env_logger::builder().is_test(true).try_init();
```
jimblandy, line 5916: Okay, this `Result` needs to be ignored because `env_logger` treats multiple initialization attempts as an error, right? I hope that won't stifle other errors. But I don't see anything better that can be done here.

naga/src/front/atomic_upgrade.rs
schell: I would say yes if this were `main`, where one would expect that logging would only be set up once. But in a test it may be that the logger has already been initialized by some other test, in which case I don't want to panic. 

Usually when using logging in a test I'll use `ctor` to ensure that we only init once, but that would turn on logging for all tests and introduce a new dependency.

schell:
@teoxoy do you think you would feel comfortable giving this a look over?

naga/src/front/atomic_upgrade.rs
jimblandy: Oh, right. I forgot about all that mess.

naga/src/front/spv/mod.rs
schell: Yes, exactly.

schell:
@jimblandy do you have any more comments? I think we're getting pretty close with this 🤞 .

naga/src/front/spv/mod.rs
```diff
@@ -1323,6 +1323,94 @@ impl<I: Iterator<Item = u32>> Frontend<I> {
...
+            pointer: p_lexp_handle,
+            fun: atomic_function,
+            value: v_lexp_handle,
+            result: Some(r_lexp_handle),
+        };
+        block.push(stmt, span);
+
+        // Store any associated global variables so we can upgrade their types later
+        self.upgrade_atomics
+            .extend(ctx.get_contained_global_variable(p_lexp_handle));
```
jimblandy: Each time the SPV front end calls `get_contained_global_variable`, we use `extend` to add the global to `Frontend::upgrade_atomics`, This means that if `get_contained_global_variable` can't identify the global to which this atomic is referring, we don't upgrade anything. Is this the correct behavior? Why aren't these calls returning an error?

jimblandy:
@schell Does this commit look okay? https://github.com/gfx-rs/wgpu/commit/87b01090af5363d77fb54292d8a94dc60f8a1e3b

naga/src/front/spv/mod.rs
schell: Previously this function recursed and returned `IndexSet<Handle<crate::GlobalVariable>>`, but then I changed that to `Option<Handle<crate::GlobalVariable>>` as per your suggestion. But I think it makes sense to error, as we'll always be dealing with a global variable.

schell:
@jimblandy that commit seems just fine 👍

jimblandy:
@schell So, right now this branch has a bunch of merges from trunk. We are generally trying to merge PRs by rebasing, because we don't want the work-in-progress commits in our history forever. Could you change this PR's branch to be merge-free?

Edit: I believe the standard `git rebase` behavior will just drop merges. Then you can fixup/squash as appropriate. cc our git authority: @ErichDonGubler

naga/src/front/spv/mod.rs
jimblandy: Without pretending to understand the consequences, just in case, here's a commit that makes it an error if we can't identify a global to be upgraded: https://github.com/gfx-rs/wgpu/commit/52d43fd5c245001585d8a0253c8d845e76e6ba7b

schell:
@jimblandy I rebased earlier today, so I think we're good to go - though now it's out of date again, so I'll do that again.

schell:
I think I also need to cherry-pick your spans commit?

schell:
Ok - my rebase earlier clobbered your "span" commit, so I cherry-picked it back in. The change of `get_contained_global_variable` to return a `Result` I made at the same time you did, and your change got clobbered as well, but my change covers that issue. So now I think we're g2g 👍

schell:
Tests still pass and everything :)

naga/src/front/spv/mod.rs
```diff
@@ -4027,31 +4192,141 @@ impl<I: Iterator<Item = u32>> Frontend<I> {
...
+                        Op::AtomicSMin => crate::AtomicFunction::Min,
+                        Op::AtomicUMin => crate::AtomicFunction::Min,
+                        Op::AtomicSMax => crate::AtomicFunction::Max,
+                        Op::AtomicUMax => crate::AtomicFunction::Max,
+                        Op::AtomicAnd => crate::AtomicFunction::And,
+                        Op::AtomicOr => crate::AtomicFunction::InclusiveOr,
+                        _ => crate::AtomicFunction::ExclusiveOr,
+                    },
+                )?,
+                Op::AtomicFlagTestAndSet => {
```
jimblandy: There's no test coverage for this, or for `Op::AtomicFlagClear`.

jimblandy (changes requested):
I think the flag ops are not going to work, because WGSL doesn't support `atomic<bool>`, so those `Literal::Bool` values are going to make something unhappy.

schell:
Ok, I'll add tests for those and debug it. As an alternative do you think we can make it use `u32`?

jimblandy:
Yes, treating the argument as `atomic<u32>` or `atomic<i32>` seems consistent with the SPIR-V ops, too, which say that the pointer should be a pointer "to a 32-bit integer type".

schell:
Yeah, you're right. I just missed that. Ok, I'll just stick to that then.

jimblandy:
If the tests pass with the code as it is, then if I'm reading correctly (and I'm often not), that would mean that the validator is not properly type-checking `Statement::Atomic`, because the pointer would be to an `atomic<i32 | u32>`, but the value would be a `bool`.

schell:
You know what - I don't even have the ability to create code that uses the flag ops (not from rust-gpu, at least). I say we skip these until a later date.

jimblandy (approved):
Okay, with the untested ops removed, this is good to go.

jimblandy (changes requested):
Sorry - I just noticed that some of the `.spvasm` files in `naga/tests/in/spv` do not match their `.spv` files. That needs to be squared away before we can merge.

jimblandy:
I've filed #6292 as a follow-up for converting the tests to snapshot tests.

jimblandy (approved):

PR #5836, opened 2024-06-18
Make `compile_fxc` accept `full_stage: &CStr`

workingjubilee:
As discussed in https://github.com/gfx-rs/wgpu/pull/5812#issuecomment-2174536813 the reason that it's acceptable to pass this String to an `LPCSTR` argument is because it's secretly a CStr. Make it be what it thinks it is, as otherwise, despite this having only one caller, it's not typesafe.

### DIFF
--- a/wgpu-hal/src/dx12/device.rs
+++ b/wgpu-hal/src/dx12/device.rs
@@ -273,6 +273,7 @@ impl super::Device {
                 dxc_container,
             )
         } else {
+            let full_stage = ffi::CStr::from_bytes_with_nul(full_stage.as_bytes()).unwrap();
             shader_compilation::compile_fxc(
                 self,
                 &source,
--- a/wgpu-hal/src/dx12/shader_compilation.rs
+++ b/wgpu-hal/src/dx12/shader_compilation.rs
@@ -18,7 +18,7 @@ pub(super) fn compile_fxc(
     source_name: Option<&CStr>,
     raw_ep: &std::ffi::CString,
     stage_bit: wgt::ShaderStages,
-    full_stage: String,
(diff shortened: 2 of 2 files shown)


workingjubilee:
- [x] ability to write type-safe code
- [ ] ability to remember import paths

Elabajaba:
Should I cherrypick this one for backporting as well?

workingjubilee:
No idea how a patch for DX12 on Windows is affecting Vulkan on Linux...?

@Elabajaba Unlike the other patch, where any version of wgpu with `compile_fxc` and *without* that patch is actively inducing reads of garbage data on Windows (or segfaults, which are much _less_ worse), this one shouldn't affect the actually-compiled programs. Rather, it affects whether `compile_fxc` is an `unsafe fn` and whether its callers must have secret magic knowledge. As this function is not public outside this codebase, the change here will have no effect on resulting programs, but will make it less likely a future change incorrectly violates invariants.

> Rather, it affects whether `compile_fxc` is an `unsafe fn` and whether its callers must have secret magic knowledge.

Currently: `compile_fxc` is not an `unsafe fn`, but its callers require secret knowledge of what happens to their strings, so it should be.

With this patch: `compile_fxc` does not need to be an `unsafe fn` as its callers no longer require secret knowledge of what happens to their strings.

workingjubilee:
"so is that a yes or no"
"idk lol I don't know what your policy is for backporting type-safety fixes?"

teoxoy:
> No idea how a patch for DX12 on Windows is affecting Vulkan on Linux...?

One of the tests is intermittently failing.

teoxoy (approved):

Issue #2521, opened 2022-03-03
Write timestamp lands on wrong side of barriers


fintelia:
**Description**
I've been experimenting with wgpu's `write_timestamp` functionality to measure the execution time of compute shaders. Unfortunately, the lack of control of when they happen in relation to pipeline barriers seems to be preventing them from being of much use. 

**Repro steps**
Execute two compute dispatches, and measure how long the second one takes:
```
cpass.dispatch(...);
cpass.write_timestamp(...);
cpass.dispatch(...);
cpass.write_timestamp(...);
```

**Expected vs observed behavior**
I would expect that the difference between the timestamps would reflect the execution time of the second dispatch. Instead, I observed that the measurement would sometimes be an order of magnitude larger, presumably including much of the time executing the first dispatch.

I assume this is related to vkCmdWriteTimestamp happening too early. Partial trace:
```
47 vkCmdBindPipeline(pipeline.generate.displacements)
48 vkCmdBindDescriptorSets(0, { bindgroup.generate.displacements })
49 vkCmdResetQueryPool(Query Pool 250)
50 vkCmdWriteTimestamp(Query Pool 250, 6)
51 vkCmdPipelineBarrier({ buffer.nodes })
52 vkCmdPipelineBarrier({ texture.tiles.displacements })
53 vkCmdDispatch(9, 9, 1)
54 vkCmdResetQueryPool(Query Pool 250)
55 vkCmdWriteTimestamp(Query Pool 250, 7)

```

**Platform**
I'm using Linux with an AMD GPU + mesa drivers, though I imagine this applies more broadly.

cwfitzgerald:
Yeah so this is an inherent problem with the current api -- because the barriers happen within the dispatch, if the barrier exists, it will always be included in the timestamp. The timestamp api is actually under a lot of debate within webgpu, and I don't know what the current status of it is. No matter what the upstream api ends up being, we should have an api that allows you to get the information you need to get this information.

Wumpf:
There's now timestamps on the borders of a Pass which are a bit more well defined. Also, timestamp taking within passes is a wgpu extension, not a feature of WebGPU itself.

I'm have no idea what we can do about this problem in general though: It would imply to schedule barriers earlier than we would without the write_timestamp call which then completely changes the perf characteristics of the thing that we want to measure 🤔 

[...]

PR #5838, opened 2024-06-18
Mark timestamp query inside encoders/passes tests as flaky

teoxoy:
We keep seeing these tests failing intermittently in CI, mark them as flaky since https://github.com/gfx-rs/wgpu/issues/2521 already tracks this issue.

### DIFF
--- a/examples/src/timestamp_queries/mod.rs
+++ b/examples/src/timestamp_queries/mod.rs
@@ -435,7 +435,7 @@ pub fn main() {
 
 #[cfg(test)]
 mod tests {
-    use wgpu_test::{gpu_test, GpuTestConfiguration};
+    use wgpu_test::{gpu_test, FailureCase, GpuTestConfiguration};
 
     use super::{submit_render_and_compute_pass_with_queries, QueryResults};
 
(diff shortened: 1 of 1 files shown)


examples/src/timestamp_queries/mod.rs
```diff
@@ -456,7 +456,9 @@ mod tests {
                 .features(
                     wgpu::Features::TIMESTAMP_QUERY
                         | wgpu::Features::TIMESTAMP_QUERY_INSIDE_ENCODERS,
-                ),
+                )
+                // see https://github.com/gfx-rs/wgpu/issues/2521
+                .skip(FailureCase::always().flaky()),
```
cwfitzgerald: Do we want to _skip_ or just mark as a flaky failure?

teoxoy: I wanted to `skip` them at first then I saw we have a `flaky` expectation and forgot to change `skip` to `expect_fail`.
Will fix.

examples/src/timestamp_queries/mod.rs
```diff
@@ -456,7 +456,9 @@ mod tests {
                 .features(
                     wgpu::Features::TIMESTAMP_QUERY
                         | wgpu::Features::TIMESTAMP_QUERY_INSIDE_ENCODERS,
-                ),
+                )
+                // see https://github.com/gfx-rs/wgpu/issues/2521
+                .expect_fail(FailureCase::always().flaky()),
```
cwfitzgerald: Ack, sorry one more comment, can you add a message to make sure that other issues with the test are caught, `flaky` is a big hammer (ignore status of the test) so we want to refine it as much as possible.

teoxoy: I added a "unexpected timestamp" panic message.

cwfitzgerald (approved):
Thanks!

torokati44:
FYI auto-merge is blocked by the unresolved conversation.

PR #5841, opened 2024-06-19
Device mismatch validation improvements

teoxoy:
At a high level this PR adds a new trait that children of the device implement which we use for device mismatch validation.

The validation now relies on `Arc::ptr_eq` rather than IDs.


nical (approved):
:tada:

PR #5843, opened 2024-06-19
In readme, use emoji literals instead of shortcodes

onkoe:
Good morning! Crates.io doesn't support emoji shortcodes like `:ok:`. While the readme's "supported platforms" listing looks great on GitHub, it's not so nice on Crates.io for that reason.

Such an issue can be moderately confusing for new users, and it's been bothering me for a while now.

Here's a minor fix to the readme that takes care of the problem! Please let me know if any changes are needed.


Wumpf (approved):
thx, makes sense!


this is killing me though 😄 
image

PR #5844, opened 2024-06-19
[naga] Use new `NonMaxU32` type for `Handle` indices.

jimblandy:
This is non-essential (hey, it's a holiday, I can work on whatever I want), but it looks to me like a reasonable cleanup to our `Handle` code. The upshot is, all the odd little +1s and -1s around handle indices have been isolated in a single module.

~~The only piece that's missing is, we still just derive `Serialize` and `Deserialize`, and `NonMaxU32` uses a 1-biased `NonZeroU32`, so the serialized forms of these types still show as 1-based indices, in snapshot test output, etc. It wouldn't be hard to write explicit implementations of those traits, and then we could truly forget about 1-based indexing.~~

Serialization and deserialization use the natural form, so that Naga snapshot `ir` and `analysis` files now use zero-based indexing, like everything else.

-----

-   Define a new type, `NonMaxU32`, that can represent any `u32` value
    except `u32::MAX`, defined such that `Option<NonMaxU32>` is still a
    32-bit value.

    Change `Handle` to use `NonMaxU32`. Adjust all code that works
    directly with handle indices to assume zero-based indices, not
    one-based indices.

-   Remove the phrase "zero-based" from comments and function names. Now
    that there is no mix of zero-based and one-based indices, there's no
    need to call out the distinction.

-   Delete unused constant `Handle::DUMMY`.

-   Manually implement `serde` traits for `NonMaxU32`.

    When the appropriate features are enabled, manually implement
    `serde::Serialize` and `serde::Deserialize`, such that the serialized
    form of `NonMaxU32::new(n).unwrap()` is the same as that of `n`.

    This eliminates the last trace of 1-based indices from Naga's snapshot
    tests, and aligns `std::fmt::Debug` with the serialized form.


jimblandy:
Note that there is already a [`nonmax` crate](https://crates.io/crates/nonmax). We don't need all the other types, and we don't need all the methods, so I think this is preferable to taking on the dependency.

jimblandy:
This may be better to review one commit at a time, as the changes that affect a lot of snapshots are isolated.

teoxoy (approved):
This is great! No longer do we have to figure out if we live in a 1-based world vs 0-based when looking at the snapshots.

PR #5845, opened 2024-06-20
[naga] Remove uses of `Handle::index` for identifier generation from backends

jimblandy:
Remove all uses of `Handle::index` for generating identifiers from Naga backends.

This is a step towards making `Handle::index` less prominent in the code base. Eventually, I'd like to use well-typed collections for everything, and have `Handle::index` be private to `arena.rs`.

-   [naga msl-out] Use `Handle::write_prefixed` instead of `index`.

    This replaces all uses of `Handle::index` for generating identifiers
    in `naga::back::msl` with uses of `Handle::write_prefixed`.

    There are still some uses of `Handle::index` remaining.

-   [naga dot-out] Use `Handle::write_prefixed` instead of `index`.

    This replaces all uses of `Handle::index` in `naga::back::dot` with
    uses of `Handle::write_prefixed`.

-   [naga hlsl-out] Use `Baked` for baked expressions.

    Make the HLSL backend more like other backends by using `back::Baked`
    to generate names for baked expression identifiers. This removes the
    final uses of `Handle::index` from the HLSL backend.

    This is separated out from the previous commit because it changes lots
    of snapshot tests, whereas the previous commit has no effect on Naga's
    output.

-   [naga] Introduce `Baked` newtype for writing baked expression names.

    Introduce a newtype `naga::back::Baked` that wraps a
    `Handle<Expression>` and formats with `std::fmt::Display` as a baked
    expression identifier. Use this in all backends for generating baked
    identifiers.

    Delete `BAKE_PREFIX`, as it's no longer used outside of `Baked`'s
    `Display` implementation.


teoxoy (approved):
Nice!

naga/src/back/dot/mod.rs
```diff
@@ -453,22 +479,23 @@ fn write_fun(
...
                 prefix,
                 from,
                 prefix,
-                to.index(),
+                Prefixed(to),
             )?;
         }
     }
 
+    assert!(sg.calls.is_empty());
```
dylanowen, line 491: Was this accidentally left in?

ErichDonGubler: Ohh, hmm, that _is_ weird. @teoxoy, @jimblandy, if `sg.calls` is supposed to be empty, why did we keep code that emits output based on it later? 🤔

teoxoy: I missed this while reviewing the PR, it's probably a leftover from debugging an issue.

dylanowen: I was having issues running the dot backend but removing that line fixed it.

teoxoy: Could you put up a PR for it?

PR #5846, opened 2024-06-20
[naga] Introduce `HandleVec`.

jimblandy:
Introduce a new type, `HandleVec<T, U>`, which is basically a `Vec<U>`, except that it's indexed by values of type `Handle<T>`. This gives us a more strictly typed way to build tables of data parallel to some other `Arena`.

Change `naga::back::pipeline_constants` to use `HandleVec` instead of `Vec`. This removes many calls to `Handle::index`, and makes the types more informative.

In `naga::back::spv`, change `Writer` and `BlockContext` to use `HandleVec` instead of `Vec` for various handle-indexed tables.

### DIFF
--- a/naga/src/arena.rs
+++ b/naga/src/arena.rs
@@ -783,3 +783,101 @@ where
         arbitrary::size_hint::and(depth_hint, (0, None))
     }
 }
+
+/// A [`Vec`] indexed by [`Handle`]s.
+///
+/// A `HandleVec<T, U>` is a [`Vec<U>`] indexed by values of type `Handle<T>`,
+/// rather than `usize`.
+///
+/// Rather than a `push` method, `HandleVec` has an [`insert`] method, analogous
+/// to [`HashMap::insert`], that requires you to provide the handle at which the
+/// new value should appear. However, since `HandleVec` only supports insertion
+/// at the end, the given handle's index must be equal to the the `HandleVec`'s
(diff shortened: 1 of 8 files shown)


naga/src/arena.rs
```diff
@@ -783,3 +783,101 @@ where
...
+///
+/// Rather than a `push` method, `HandleVec` has an [`insert`] method, analogous
+/// to [`HashMap::insert`], that requires you to provide the handle at which the
+/// new value should appear. However, since `HandleVec` only supports insertion
+/// at the end, the given handle's index must be equal to the the `HandleVec`'s
+/// current length; otherwise, the insertion will panic.
+///
+/// [`insert`]: HandleVec::insert
+/// [`HashMap::insert`]: std::collections::HashMap::insert
+pub(crate) struct HandleVec<T, U> {
```
nical, line 800: Should this be `HandleVec<T>` since `U` looks like it is always `Handle<T>`?

jimblandy: Not in the `spv` backend.

jimblandy: But also in `pipeline_constants.rs` we have:

```
    override_map: &mut HandleVec<Override, Handle<Constant>>,
```

teoxoy (approved):
Great stuff, makes things much saner to work with!

PR #5847, opened 2024-06-20
Check for subgroup support when creating a naga `Validator`

Elabajaba:
**Description**
Detecting support for subgroup ops in naga validators is currently pretty ugly, as it requires creating a validator then calling 2 separate functions on it. I found this really annoying and ugly when I updated `naga_oil` to naga 0.20.

Add `SUBGROUP_VERTEX_STAGE` to naga's `Capabilities` so that naga automatically detects subgroup support when creating a validator.

**Testing**
I ran the water example and it still worked.

**Checklist**

- [ ] Run `cargo fmt`.


teoxoy (approved):
LGTM, thanks!

PR #5848, opened 2024-06-20
Remove more internal uses of IDs

teoxoy:


ErichDonGubler:
🖐🏻I'll take on review for this.

nical (approved):

teoxoy:
looks like we had a race condition 😅

Wumpf:
_nice_

PR #5855, opened 2024-06-21
[naga] Add `packed` as a keyword for GLSL

### LINKED ISSUE #5853: Naga produces a reserved keyword when translating variable name `packed` to GLSL
**Description**

The following fragment of a WGSL shader:

```wgsl
fn unpack_components(packed: u32) -> vec2<u32> {
    return vec2<u32>(packed >> 8u, packed & 15u);
}
```

(Full shader: https://github.com/ruffle-rs/ruffle/blob/036839fb1f382852711e0c1e270fa3e69a3a540a/render/wgpu/shaders/filter/displacement_map.wgsl)

sometimes fails with `error: syntax error, unexpected PACKED_TOK, expecting ')'` (Linux) or `'packed' : reserved word ERROR: 0:42: 'packed' : syntax error syntax error` (Windows) on Open GL.

**Repro steps**

Use a variable `packed` in a WGSL shader and try to compile it using Open GL backend on a device where `packed` is a reserved keyword (I have no idea when `packed` is a keyword, it works for some people, for some doesn't).

The shader should always work without syntax errors.

**Extra materials**

Bugs reported for Ruffle with this issue: https://github.com/ruffle-rs/ruffle/issues/12844, https://github.com/ruffle-rs/ruffle/issues/16642.

A screenshot pasted at Ruffle's Discord:

<image width="300" src="https://github.com/gfx-rs/wgpu/assets/1507072/4cb42bdd-b274-4e54-b7dc-e29c13d81fe0"/>

My workaround that turned out to work:
* https://github.com/ruffle-rs/ruffle/pull/16805

**Platform**

I personally cannot reproduce it as Open GL does not work on my platform, but platform information is included in the issues referenced above.

kjarosh:
Turns out that sometimes `packed` is a keyword, and the produced GLSL had syntax errors due to that.

**Connections**
Fixes https://github.com/gfx-rs/wgpu/issues/5853.

**Description**
Sometimes `packed` is a keyword in GLSL, and naga was preserving variables named that way, which caused syntax errors.

**Testing**
I executed the following command and observed that the `packed` variable name was suffixed with an underscore. The input file was the original file that caused syntax errors for some users (https://github.com/ruffle-rs/ruffle/blob/036839fb1f382852711e0c1e270fa3e69a3a540a/render/wgpu/shaders/filter/displacement_map.wgsl). Others have confirmed that renaming the identifier to something else than `packed` fixes those issues.

```
cargo run --package naga-cli displacement_map.wgsl output.vert --entry-point main_vertex
```

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy` (it produces warnings but my patch does not introduce any). If applicable, add:
  - The same set of ~45 tests fail both before and after this PR, so... it's stable at least?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -175,6 +175,7 @@ By @atlv24 in [#5383](https://github.com/gfx-rs/wgpu/pull/5383)
 #### Naga
 
 - In spv-out don't decorate a `BindingArray`'s type with `Block` if the type is a struct with a runtime array by @Vecvec in [#5776](https://github.com/gfx-rs/wgpu/pull/5776)
+- Add `packed` as a keyword for GLSL by @kjarosh in [#5855](https://github.com/gfx-rs/wgpu/pull/5855)
 
 ## v0.20.0 (2024-04-28)
 
--- a/naga/src/back/glsl/keywords.rs
+++ b/naga/src/back/glsl/keywords.rs
@@ -473,6 +473,8 @@ pub const RESERVED_KEYWORDS: &[&str] = &[
     "anyInvocation",
     "allInvocations",
     "allInvocationsEqual",
+    // Sometimes "packed" is a keyword, see https://github.com/gfx-rs/wgpu/issues/5853
+    "packed",
     //
     // entry point name (should not be shadowed)
     //


torokati44:
I'm wondering whether there are any other similar keywords that should also be added. Such as the `std###` ones.

kjarosh:
I was wondering that too, but I couldn't find any evidence on the internet that any of them should be treated as a keyword. However, we may prefer falsely claiming that something is a keyword to falsely claiming that something isn't.

There's also an option to test those suspected keywords on a platform where `packed` is a keyword.

torokati44:
I tried looking at the documents linked in the comments within this list, and search for `packed` in them, seeing what's next to it in there.

kjarosh:
@sombraguerrero has confirmed that whilst `packed` is a keyword on their platform, other similar identifiers (`std140`,`std430`,`row_major`,`location`,`triangles`,`index`,`r8`) are not.

I did some tests myself (but not using wgpu, as OpenGL does not work for me in wgpu). And `packed` on my platform:

* is not a keyword in OpenGL,
* is not a keyword in OpenGL ES,
* is a keyword in WebGL 1.0,
* is not a keyword in WebGL 2.0.

I also tried other similar identifiers in WebGL, but it really seems that `packed` is the only one.

teoxoy (approved):
Thanks!

teoxoy:
I looked at the specs and we seem to be missing a few more, I will open a PR for the remaining ones.

PR #5856, opened 2024-06-21
[naga msl-out] Remove unnecessary block.

jimblandy:
Remove unnecessary curly brackets around some statements in `naga::back::msl::Writer::write`.


jimblandy:
Haste makes waste. This does at least free a `Vec` of stuff that isn't used.

PR #5858, opened 2024-06-22
buffer size mismatch: more detail in error message

nolanderc:
**Connections**
N/A

**Description**
In case there is a buffer size mismatch between the shader and bindings, prints both sizes in the error message.

**Testing**
N/A

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/validation.rs
+++ b/wgpu-core/src/validation.rs
@@ -147,10 +147,10 @@ pub enum BindingError {
         binding: naga::AddressSpace,
         shader: naga::AddressSpace,
     },
-    #[error("Buffer structure size {shader}, added to one element of an unbound array, if it's the last field, ended up greater than the given `min_binding_size`, which is {binding}")]
+    #[error("Buffer structure size {buffer_size}, added to one element of an unbound array, if it's the last field, ended up greater than the given `min_binding_size`, which is {min_binding_size}")]
     WrongBufferSize {
-        binding: wgt::BufferSize,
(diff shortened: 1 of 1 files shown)


wgpu-core/src/validation.rs
```diff
@@ -147,8 +147,11 @@ pub enum BindingError {
         binding: naga::AddressSpace,
         shader: naga::AddressSpace,
     },
-    #[error("Buffer structure size {0}, added to one element of an unbound array, if it's the last field, ended up greater than the given `min_binding_size`")]
-    WrongBufferSize(wgt::BufferSize),
+    #[error("Buffer structure size {shader}, added to one element of an unbound array, if it's the last field, ended up greater than the given `min_binding_size`, which is {binding}")]
+    WrongBufferSize {
+        binding: wgt::BufferSize,
+        shader: wgt::BufferSize,
+    },
```
Wumpf: those field names don't make sense to me. Both are sizes and not binding indices as the would names imply

```suggestion
    #[error("Buffer structure size {structure_size}, added to one element of an unbound array, if it's the last field, ended up greater than the given `min_binding_size`, which is {min_binding_size}")]
    WrongBufferSize {
        structure_size: wgt::BufferSize,
        min_binding_size: wgt::BufferSize,
    },
```

Wumpf (changes requested):
good addition, thanks!

Wumpf (approved):

PR #5862, opened 2024-06-23
[naga] Break `naga::arena` up into submodules.

jimblandy:
This is broken up into independent commits that are probably easier to review individually.

-   [naga] Use `Range::from_index_range` in `Arena::range_from`.

    Use the associated function `Range::from_index_range` to construct the
    return value of `Arena::range_from`, rather than constructing the
    `Range` directly, in preparation for those fields becoming private.

-   [naga] Use `Handle::new`, don't construct the `Handle` directly.

    In preparation for making `Handle` fields private to another module,
    have the `Iterator` implementation for `arena::Range` call
    `Handle::new` to construct the handle being produced, rather than
    building it using a struct literal.

-   [naga] Move `naga/src/arena.rs` to `naga/src/arena/mod.rs`.

-   [naga] Break `naga::arena` up into submodules.

    This commit is almost entirely code motion. The only meaningful
    changes should be:

    - changes to imports
    - changes to visibility
    - changes to use visible associated constructor functions instead of
      trying to construct values directly using now-invisible fields
    - moving the crate-level "Arena" docs into the `arena` module


teoxoy (approved):
Nice!

teoxoy:
It looks like there is a conflict since https://github.com/gfx-rs/wgpu/pull/5864 added `#[derive(Debug)]` to `HandleVec`.

PR #5870, opened 2024-06-24
[gl] reorder `program_cache` & `context` lock acquisition

teoxoy:
We are using `program_cache.try_lock()` when creting pipelines which is covered by a guard gotten from `context.lock()`. For the `.try_lock()` to always succeed we need to make sure that the other lock acquisitions are also covered by a `context.lock()`.

The `wgpu_examples::hello_compute::tests::multithreaded_compute` test has been failing intermittently in CI due to this ([recent example](https://github.com/gfx-rs/wgpu/actions/runs/9642173501/job/26589506810)).

Follow-up to https://github.com/gfx-rs/wgpu/pull/3380, relevant discussion: https://github.com/gfx-rs/wgpu/pull/3380#discussion_r1070791808.

### DIFF
--- a/wgpu-hal/src/gles/device.rs
+++ b/wgpu-hal/src/gles/device.rs
@@ -1409,17 +1409,17 @@ impl crate::Device for super::Device {
     }
 
     unsafe fn destroy_render_pipeline(&self, pipeline: super::RenderPipeline) {
-        let mut program_cache = self.shared.program_cache.lock();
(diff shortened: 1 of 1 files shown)


ErichDonGubler (approved):
Approving, but: Shouldn't @jimblandy's rules for lock ordering be updated with this, too? 🤔

teoxoy:
The rules are for the locks in `wgpu-core`; these locks are in the GL hal backend.

PR #5136, opened 2024-01-24
Snatch bind groups associated with destroyed textures and buffers

nical:
**Connections**

Rebased on top of #5131 

**Description**

Addresses the BindGroup part of https://github.com/gfx-rs/wgpu/issues/5079.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.
- [x] Run `cargo xtask test` to run tests.

PR #5871, opened 2024-06-24
Remove more internal uses of IDs #2

teoxoy:

### DIFF
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -475,7 +475,6 @@ pub struct BindGroupLayout<A: HalApi> {
     #[allow(unused)]
     pub(crate) binding_count_validator: BindingTypeMaxCountValidator,
     pub(crate) info: ResourceInfo<BindGroupLayout<A>>,
-    pub(crate) label: String,
 }
 
 impl<A: HalApi> Drop for BindGroupLayout<A> {
@@ -504,10 +503,6 @@ impl<A: HalApi> Resource for BindGroupLayout<A> {
 
     fn as_info_mut(&mut self) -> &mut ResourceInfo<Self> {
         &mut self.info
-    }
-
-    fn label(&self) -> &str {
-        &self.label
     }
 }
 
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -1,5 +1,3 @@
-#[cfg(feature = "trace")]
-use crate::device::trace;
 use crate::{
     device::{
         bgl, Device, DeviceError, MissingDownlevelFlags, MissingFeatures, SHADER_STAGE_COUNT,
@@ -486,11 +484,6 @@ impl<A: HalApi> Drop for BindGroupLayout<A> {
             self.device.bgl_pool.remove(&self.entries);
         }
         if let Some(raw) = self.raw.take() {
-            #[cfg(feature = "trace")]
-            if let Some(t) = self.device.trace.lock().as_mut() {
-                t.add(trace::Action::DestroyBindGroupLayout(self.info.id()));
-            }
-
             resource_log!("Destroy raw BindGroupLayout {:?}", self.info.label());
             unsafe {
                 use hal::Device;
@@ -638,11 +631,6 @@ impl<A: HalApi> Drop for PipelineLayout<A> {
     fn drop(&mut self) {
         if let Some(raw) = self.raw.take() {
             resource_log!("Destroy raw PipelineLayout {:?}", self.info.label());
-
-            #[cfg(feature = "trace")]
-            if let Some(t) = self.device.trace.lock().as_mut() {
-                t.add(trace::Action::DestroyPipelineLayout(self.info.id()));
-            }
 
             unsafe {
                 use hal::Device;
@@ -879,11 +867,6 @@ impl<A: HalApi> Drop for BindGroup<A> {
     fn drop(&mut self) {
         if let Some(raw) = self.raw.take() {
             resource_log!("Destroy raw BindGroup {:?}", self.info.label());
-
-            #[cfg(feature = "trace")]
-            if let Some(t) = self.device.trace.lock().as_mut() {
-                t.add(trace::Action::DestroyBindGroup(self.info.id()));
-            }
 
             unsafe {
                 use hal::Device;
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -483,7 +483,7 @@ impl<A: HalApi> Drop for BindGroupLayout<A> {
             self.device.bgl_pool.remove(&self.entries);
         }
         if let Some(raw) = self.raw.take() {
-            resource_log!("Destroy raw BindGroupLayout {:?}", self.info.label());
+            resource_log!("Destroy raw {}", self.error_ident());
             unsafe {
                 use hal::Device;
                 self.device.raw().destroy_bind_group_layout(raw);
@@ -625,8 +625,7 @@ pub struct PipelineLayout<A: HalApi> {
 impl<A: HalApi> Drop for PipelineLayout<A> {
     fn drop(&mut self) {
         if let Some(raw) = self.raw.take() {
-            resource_log!("Destroy raw PipelineLayout {:?}", self.info.label());
-
+            resource_log!("Destroy raw {}", self.error_ident());
             unsafe {
                 use hal::Device;
                 self.device.raw().destroy_pipeline_layout(raw);
@@ -861,8 +860,7 @@ pub struct BindGroup<A: HalApi> {
 impl<A: HalApi> Drop for BindGroup<A> {
     fn drop(&mut self) {
         if let Some(raw) = self.raw.take() {
-            resource_log!("Destroy raw BindGroup {:?}", self.info.label());
-
+            resource_log!("Destroy raw {}", self.error_ident());
             unsafe {
                 use hal::Device;
                 self.device.raw().destroy_bind_group(raw);
--- a/wgpu-core/src/command/bundle.rs
+++ b/wgpu-core/src/command/bundle.rs
@@ -423,8 +423,7 @@ impl RenderBundleEncoder {
                         .write()
                         .add_single(bind_group);
 
-                    self.check_valid_to_use(bind_group.device.info.id())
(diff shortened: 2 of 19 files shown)


wgpu-core/src/conv.rs
```diff
@@ -2,6 +2,13 @@ use wgt::TextureFormatFeatures;
 
 use crate::resource::{self, TextureDescriptor};
 
+pub fn bool_to_err<E, F>(cond: bool, err: F) -> Result<(), E>
+where
+    F: FnOnce() -> E,
+{
+    cond.then_some(()).ok_or_else(err)
+}
```
ErichDonGubler: nitpick(non-blocking): RE: [d67775dfff3203b87c8fd12bbdabd2a0b0a3cc7c](https://github.com/gfx-rs/wgpu/commit/d67775dfff3203b87c8fd12bbdabd2a0b0a3cc7c): I'm not persuaded that the `bool_to_err` helper actually provides value worth its weight; it seems it saves 4 characters, and swaps the order in which operations are written. 🤷🏻‍♂️

wgpu-core/src/resource.rs
```diff
@@ -843,11 +843,6 @@ impl<A: HalApi> Drop for DestroyedBuffer<A> {
         if let Some(raw) = self.raw.take() {
             resource_log!("Destroy raw Buffer (destroyed) {:?}", self.label());
 
-            #[cfg(feature = "trace")]
-            if let Some(t) = self.device.trace.lock().as_mut() {
-                t.add(trace::Action::DestroyBuffer(self.id));
-            }
-
```
ErichDonGubler, line 865: question: RE: hunk removing double `trace` actions here and in other places b/w [28ea0ed5dd35954883df06605e69034e12b1b17e](https://github.com/gfx-rs/wgpu/commit/28ea0ed5dd35954883df06605e69034e12b1b17e) and [73668b3ec3b99564c2f1f320ab33d21e7e902ed1](https://github.com/gfx-rs/wgpu/commit/73668b3ec3b99564c2f1f320ab33d21e7e902ed1): This is super weird. Why were these here to begin with? 🤔 How did we ever emit correct traces with these APIs?

wgpu-core/src/device/global.rs
ErichDonGubler, line 1: praise: RE: tracing earlier in [deab67c9650648516694bdca7828584025f0a71f](https://github.com/gfx-rs/wgpu/commit/deab67c9650648516694bdca7828584025f0a71f): I think that we might have actually run into cases from traces before where not all steps were captured that were necessary for reproducing, say, crashes. Like maybe with the recent [bug 1888178](https://bugzilla.mozilla.org/show_bug.cgi?id=1888178) and its companions.

wgpu-core/src/device/global.rs
```diff
@@ -496,7 +496,10 @@ impl Global {
             trace.add(trace::Action::FreeBuffer(buffer_id));
         }
 
-        let _ = buffer.unmap();
+        let _ = buffer.unmap(
+            #[cfg(feature = "trace")]
+            buffer_id,
+        );
```
ErichDonGubler, line 502: question: RE: passing `buffer_id` into `buffer.unmap(…)`: I feel like this would change behavior (probably fix something), rather than just be a refactor. I'm not sure what it would change, though; was this a crashing case previously?

wgpu-core/src/track/buffer.rs
```diff
@@ -304,17 +304,17 @@ impl<A: HalApi> ResourceTracker for BufferTracker<A> {
                 //so it's already been released from user and so it's not inside Registry\Storage
                 if existing_ref_count <= 2 {
                     resource_log!(
-                        "BufferTracker::remove_abandoned: removing {:?}",
-                        self.metadata.get_resource_unchecked(index).as_info().id()
+                        "BufferTracker::remove_abandoned: removing {}",
+                        self.metadata.get_resource_unchecked(index).error_ident()
```
ErichDonGubler, line 308: nitpick(non-blocking): I see that this `error_ident` API eventually prints some boilerplate text around the object's `label`, which means we no longer print an ID here. Wouldn't it be good to still print an ID, if we have one?

ErichDonGubler: issue(non-blocking): Docs. for `Resource::label` [state](https://github.com/ErichDonGubler/wgpu/blob/64a131feae3cc37fcfe003583e9a52be657abddd/wgpu-core/src/resource.rs#L182) that this will be non-empty unless provided a non-empty label from the user. However, `None` _can_ be specified as a `label`, in which case `ResourceInfo::new`  [appears to use an empty string](https://github.com/ErichDonGubler/wgpu/blob/64a131feae3cc37fcfe003583e9a52be657abddd/wgpu-core/src/resource.rs#L103). Between the docs., behavior, and my understanding, do you know which is incorrect?

wgpu-core/src/command/bundle.rs
```diff
@@ -817,14 +811,6 @@ impl RenderBundleEncoder {
         })
     }
 
-    fn check_valid_to_use(&self, device_id: id::DeviceId) -> Result<(), RenderBundleErrorInner> {
-        if device_id != self.parent_id {
-            return Err(RenderBundleErrorInner::NotValidToUse);
-        }
-
-        Ok(())
```
ErichDonGubler, line 831: praise: Nice, good catch on `check_valid_to_use` being redundant. 😀

wgpu-core/src/present.rs
```diff
@@ -225,7 +228,7 @@ impl Global {
                         mips: 0..1,
                     },
                     info: ResourceInfo::new(
-                        "<Surface Texture>",
+                        &Some(Cow::Borrowed("<Surface Texture>")),
```
ErichDonGubler, line 231: question(non-blocking): Should we expose an API for labeling surface textures? 🤔

ErichDonGubler (approved):
LGTM, minus nits and questions!

ErichDonGubler:
praise: Also, can I just say that the individual relatively small commits made review _zoom_ for me? 😀

wgpu-core/src/conv.rs
nical: The context is that I find `cond.then_some(()).ok_or_else(error)` to be wayt too much mental overhead over a simple branch, so Teo isolated it into a helper. But I'd also like this contraption to be replaced with the wordier but simpler:

```rust
if cond {
    Ok(())
} else {
    Err(err)
}
```

in the helper as well.

wgpu-core/src/binding_model.rs
```diff
@@ -477,21 +475,15 @@ pub struct BindGroupLayout<A: HalApi> {
     #[allow(unused)]
     pub(crate) binding_count_validator: BindingTypeMaxCountValidator,
     pub(crate) info: ResourceInfo<BindGroupLayout<A>>,
-    pub(crate) label: String,
 }
 
 impl<A: HalApi> Drop for BindGroupLayout<A> {
     fn drop(&mut self) {
+        resource_log!("Drop {}", self.error_ident());
```
nical: nit: The `resource_log` being in the branch was meant to have logging when something interesting happens (the bind group layout actually gets destroyed). If we hoist it out of the branch, then we won't be able to tell whether the bind group was snatched.

teoxoy: If that was the intent, I can move all inside the `if`. Some of them were outside and I wanted to make things consistent.

wgpu-core/src/resource.rs
teoxoy: They were somewhat new changes https://github.com/gfx-rs/wgpu/pull/5036, https://github.com/gfx-rs/wgpu/pull/5131 & https://github.com/gfx-rs/wgpu/pull/5136. We don't have end-to-end tests for tracing & replay so it's understandable that we didn't catch those.

wgpu-core/src/present.rs
teoxoy: I tried doing that but things got quite involved since the API has to be compatible with both the `wgpu-core` and `webgpu` backends of `wgpu`. On the web there is currently no way to label a canvas. I think we could add a label field to `.configure()` but we'd need to spec that first.

wgpu-core/src/device/global.rs
teoxoy: It won't change behavior as the `buffer_id` in this context is the same as the previous call to `self.info.id()`. One of the goals of this work is to eventually get rid of the `id` in the `ResourceInfo`.

wgpu-core/src/device/global.rs
teoxoy: Yeah, we should always trace as soon as possible.

wgpu-core/src/track/buffer.rs
teoxoy: > Wouldn't it be good to still print an ID, if we have one?

Once we remove the `id` field in `ResourceInfo` we will no longer have one. If we deem it necessary we can always introduce back a mechanism to add more data to `wgpu-core` objects that then gets printed in errors/logs.

> issue(non-blocking): Docs. for `Resource::label` [state](https://github.com/ErichDonGubler/wgpu/blob/64a131feae3cc37fcfe003583e9a52be657abddd/wgpu-core/src/resource.rs#L182) that this will be non-empty unless provided a non-empty label from the user. However, `None` _can_ be specified as a `label`, in which case `ResourceInfo::new`  [appears to use an empty string](https://github.com/ErichDonGubler/wgpu/blob/64a131feae3cc37fcfe003583e9a52be657abddd/wgpu-core/src/resource.rs#L103). Between the docs., behavior, and my understanding, do you know which is incorrect?

`None` and empty `String`/`str` are currently equivalent in terms of behavior. On the spec side of things, the label will be set to an empty string if it wasn't specified. In our rust API (`wgpu`) the label type is `Option<&'a str>`. I don't know why it's not just a `&str`. Maybe that's another thing worth changing.

wgpu-core/src/conv.rs
teoxoy: > and swaps the order in which operations are written

I wasn't a fan of this either.

Let's just remove `bool_to_err` and use `if`s.

wgpu-core/src/binding_model.rs
teoxoy: Done!

nical: Thanks!

PR #5872, opened 2024-06-24
Allow using include_wgsl! in const contexts

9SMTM6:
Its not possible to call into in const contexts. But what into does here under the hood is what i replaced in, and since thats just the creation of an enum variant, one can do that in const contexts.

**Connections**
Nada.

**Description**
I want my shader module in a constant (no lifetimes to deal with, can be static), and I'd like to use the provided macro for it.

**Testing**
Try to call the old include_wgsl! to create a constant, see that it doesn't work. Repeat with this version, hussa it works now.

**Checklist**

- [x] Run `cargo fmt`.


teoxoy (approved):
Nice, thanks!

PR #5874, opened 2024-06-25
Clean refs to died texture views

### LINKED ISSUE #5707: Texture view leaks
**Description**
Related to #5595 

https://github.com/gfx-rs/wgpu/pull/5595/commits/500804e01063c218353e0d9e2acc836075627069 only clean up the suspected resources, weak pointers will leak if the texture never be dropped.

**Repro steps**
Minimal reproducible example:

```rust
use std::{thread, time::Duration};

fn main() {
    let ins = wgpu::Instance::default();
    let (_adp, dev, que) = pollster::block_on(async {
        let adp = ins.request_adapter(&<_>::default()).await?;
        let (dev, que) = adp.request_device(&<_>::default(), None).await.ok()?;
        Some((adp, dev, que))
    }).unwrap();
    let size = wgpu::Extent3d {
        width: 32,
        height: 32,
        depth_or_array_layers: 1,
    };
    let tex = dev.create_texture(&wgpu::TextureDescriptor {
        label: None,
        size,
        mip_level_count: 1,
        sample_count: 1,
        format: wgpu::TextureFormat::R8Unorm,
        usage: wgpu::TextureUsages::TEXTURE_BINDING,
        view_formats: &[],
    });
    
    loop {
        let _view = tex.create_view(&<_>::default());
        que.submit(None);
        dev.poll(wgpu::Maintain::Wait);

        thread::sleep(Duration::from_secs_f32(1. / 60.));
    }
}
```

Expected: process memory keeps stable
Observed: process memory grows without bound

**Extra materials**
image

**Platform**
macOS, wgpu@77a83fb0d

xiaopengli89:
**Connections**
Close #5707

**Description**

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -841,6 +841,10 @@ impl Global {
 
             {
                 let mut views = texture.views.lock();
+
+                // Remove stale weak references
+                views.retain(|view| view.strong_count() > 0);
+
                 views.push(Arc::downgrade(&resource));
             }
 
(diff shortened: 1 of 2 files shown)


nical (changes requested):
Nice catch!

Most of the code in life.rs is quite tricky to understand and will hopefully go away in a not-so-distant future. I think that this fix would be easier to understand (and less likely to get accidentally removed along with the resource triaging code) if done just before inserting a view in `Texture::views` in `Global::texture_create_view`.

And while you are at it, could you also clean up `Texture::bind_groups` and `Buffer::bind_groups` the same way in `Global::create_bind_group`?

xiaopengli89:
> Nice catch!
> 
> Most of the code in life.rs is quite tricky to understand and will hopefully go away in a not-so-distant future. I think that this fix would be easier to understand (and less likely to get accidentally removed along with the resource triaging code) if done just before inserting a view in `Texture::views` in `Global::texture_create_view`.
> 
> And while you are at it, could you also clean up `Texture::bind_groups` and `Buffer::bind_groups` the same way in `Global::create_bind_group`?

Done.

nical:
Thanks!

nical (approved):

Issue #5869, opened 2024-06-24
Properly configure the memory allocators


nical:
The vulkan backend uses the `gpu-alloc` crate and [sets the memory allocator to the `i_am_prototyping` configuration](https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/native/vulkan/ResourceMemoryAllocatorVk.cpp) which is explicitly not meant to make it to production. The starting device memory chunk size for that config is 8 MiB and doubles every time a new chunk must be allocated.

The d3d12 backend uses the default settings provided by `gpu-allocator` (different crate), which is fixed 256 MiB memory chunks for device memory and 64 MiB for host memory. 256 MiB appears to also be the default size for VMA.

For reference, it looks like Dawn uses a fixed memory chunk size of 8MiB with a buddy allocator.

I think that wgpu should expose this configuration to the outside while providing some reasonable defaults. Firefox (which may create many devices at the same time) will want much smaller chunk sizes by default than a typical big game.

teoxoy:
Closing since https://github.com/gfx-rs/wgpu/pull/5875 landed.

PR #5875, opened 2024-06-25
Expose gpu allocation configuration options

nical:
**Connections**

Addresses #5869
See also https://github.com/Traverse-Research/gpu-allocator/issues/235

**Description**

This commit adds hints to control memory allocations strategies to device configuration options. These hints allow for automatic profiles such as optimizing for performance (the default, makes sense for a game), optimizing for memory usage (typically more useful for a web browser or UI library) and specifying settings manually.

The details of gpu allocation are still in flux. The goal is to switch vulkan and metal to gpu_allocator which is currently used with d3d12. gpu_allocator will also likely receive more configuration options, in particular the ability to start with smaller memory block sizes and progressively grow the block size. So the manual settings already provision for this upcoming option. Another approach could be to wait and add the manual option after the dust settles.

The reason for providing presets and defining values in the backends instead of in platform-independent code is that I am convinced that optimal configurations should take hardware capabilities into consideration. It's a deep rabbit hole, though, so that will be an exercise for later.

**Testing**

The examples have a mix of `Performance` and `MemoryUsage` presets which gives some very superficial coverage, however this can't properly be tested without large/realistic workloads.

**Checklist**

- [x] Run `cargo fmt`.


wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -1834,7 +1835,40 @@ impl super::Adapter {
 
         let mem_allocator = {
             let limits = self.phd_capabilities.properties.limits;
-            let config = gpu_alloc::Config::i_am_prototyping(); //TODO
+            // TODO: These configuration options should take hardware capabilities
```
nical: Note: vulkan's allocator parameters don't map super well to the manual hints, however the plan is to switch to `gpu_allocator` (the one we use with d3d12), which better aligns.

CHANGELOG.md
```diff
@@ -137,6 +137,11 @@ By @atlv24 in [#5383](https://github.com/gfx-rs/wgpu/pull/5383)
 #### General
 
 - Added `as_hal` for `Buffer` to access wgpu created buffers form wgpu-hal. By @JasondeWolff in [#5724](https://github.com/gfx-rs/wgpu/pull/5724)
+- Added memory allocation hints to `DeviceDescriptor` by @nical in [#5875](https://github.com/gfx-rs/wgpu/pull/5875)
+    - `MemoryHints::Performance`, the default, favors performance over memory usage and will likely cause large amounts of VRAM to be allocated up-front. This hint is typically good for games.
+    - `MemoryHints::MemoryUsage` favors memory usage over performance. This hint is typically useful for smaller applications or UI libraries.
+    - `MemoryHints::Manual` allows the user to specify parameters tofor the underlying GPU memory allocator. These parameters are subject to change.
+    - These hints may be ignored by some backends. Currently only the vulkan and d3d12 backends take them into account.
```
Wumpf:
```suggestion
    - `MemoryHints::Manual` allows the user to specify parameters for the underlying GPU memory allocator. These parameters are subject to change.
    - These hints may be ignored by some backends. Currently only the Vulkan and D3D12 backends take them into account.
```

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -1834,7 +1835,40 @@ impl super::Adapter {
...
+            // into consideration.
+            let mb = 1024 * 1024;
+            let perf_cfg = gpu_alloc::Config {
+                starting_free_list_chunk: 128 * mb,
+                final_free_list_chunk: 512 * mb,
```
Wumpf, line 1856: `perf_cfg` seems to be equivalent to `Config::i_am_prototyping` except that `final_free_list_chunk` is 512 instead of 128. Any specific motivation around these numbers, how did they come to be?

wgpu-hal/src/lib.rs
```diff
@@ -558,6 +558,7 @@ pub trait Adapter: WasmNotSendSync {
         &self,
         features: wgt::Features,
         limits: &wgt::Limits,
+        memory_hints: &wgt::MemoryHints,
```
Wumpf, line 565: taking a reference to `MemoryHints` seems a lil bit silly to me. Why not make it `Copy` and pass by value?

Wumpf (commented):
looking good overall, great to start to expose this!
The `Manual` configuration is indeed a bit unforunate that it doesn't map well to either backend 🤔 . Not too strong feelings about it since the path forward is pretty clear (switch to `gpu_allocator` and expose its config more or less verbatim)

wgpu-hal/src/vulkan/adapter.rs
nical: Some guesswork and a bit of looking at the vulkan memory allocator from AMD's SDK. But mostly guesswork. It's difficult to find good references about how to do this right.

`final_free_list_chunk` should be bigger than `starting_free_list_chunk` if we want the behavior of starting with smaller block sizes and using larger ones only after we observe that the small ones aren't enough, which I think is a good "I don't know what the workload is going to be like" approach.

For reference, `VMA`, and `gpu_allocator` both start with 256 MB blocks (then VMA doubles the block size each time it needs a new block). I kept the initial block size at 128 MB to avoid changing the current behavior too much without having a good sense of what the right parameters are and to limit risk. At some point it would be good to experiment with real workloads, perhaps with the help of folks in the bevy community, and get a better feel for what would make good defaults.

PS: I checked again and `gpu_alloc`'s default (prototyping) starting block size is 8MB. So my guess is that I picked 128 so that the performance mode is closer to what `VMA` and `gpu_allocator` do but didn't quite go all the way to 256 MB because of how big a difference it makes.

wgpu-hal/src/lib.rs
nical: My thinking was that I'm not sure how the parameters will evolve. If you feel strongly about it, I don't mind passing them by value and even make them copy. We don't have a commitment to not break the API.

Wumpf: 🤷 was mostly surprised. Not feeling strongly about either direction.

wgpu-hal/src/vulkan/adapter.rs
Wumpf: 👍 makes sense!
Leaving a comment of almost verbatim what you just wrote would be a good even if it boils down to "not quite sure what numbers to put here"

Wumpf (approved):
👍

PR #5882, opened 2024-06-26
Remove `BoundsCheckPolicies.image_store`

### LINKED ISSUE #5878: Naga worries too much about storage texture writes
I think it is the case that Naga generates bounds checks on storage texture writes, but this is unnecessary:
https://github.com/gpuweb/gpuweb/pull/4194

teoxoy:
Follow-up to https://github.com/gfx-rs/naga/pull/2265 & https://github.com/gfx-rs/wgpu/pull/3975.
Spec PR: https://github.com/gpuweb/gpuweb/pull/4194.
Resolves https://github.com/gfx-rs/wgpu/issues/5878.

### DIFF
--- a/naga/src/proc/index.rs
+++ b/naga/src/proc/index.rs
@@ -112,21 +112,15 @@ pub struct BoundsCheckPolicies {
     /// This controls the behavior of [`ImageLoad`] expressions when a coordinate,
     /// texture array index, level of detail, or multisampled sample number is out of range.
     ///
+    /// There is no corresponding policy for [`ImageStore`] statements. All the
+    /// platforms we support already discard out-of-bounds image stores,
+    /// effectively implementing the "skip write" part of [`ReadZeroSkipWrite`].
+    ///
     /// [`ImageLoad`]: crate::Expression::ImageLoad
+    /// [`ImageStore`]: crate::Statement::ImageStore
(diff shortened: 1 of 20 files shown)


naga/src/proc/index.rs
```diff
@@ -116,18 +116,6 @@ pub struct BoundsCheckPolicies {
...
-    /// of range?
-    ///
-    /// This controls the behavior of [`ImageStore`] statements when a coordinate,
-    /// texture array index, level of detail, or multisampled sample number is out of range.
-    ///
-    /// This policy should't be needed since all backends should ignore OOB writes.
-    ///
-    /// [`ImageStore`]: crate::Statement::ImageStore
-    #[cfg_attr(feature = "deserialize", serde(default))]
-    pub image_store: BoundsCheckPolicy,
```
jimblandy: Could you add a short paragraph to `image_load`'s comments explaining that all the hardware we support already discards out-of-bounds stores, so an `image_store` field is not necessary?

jimblandy (approved):
Looks great. I love deleting code without removing features.

naga/src/proc/index.rs
jimblandy: Actually I'll just do this.

jimblandy: Done in 6f0fd39.

PR #5883, opened 2024-06-26
Remove more internal uses of IDs #3

teoxoy:
The reason I extracted the code from passes was because `PassErrorScope` is `Copy` and I wanted to replace all the resources referenced by ids in its variants with `ResourceErrorIdent`s which is not `Copy`. After going through all the work to do that, I realized I could remove the ids from all variants anyway because the inner error types already contained `ResourceErrorIdent`'s where relevant; or if they didn't I added them. I think we should still land the code extractions since we can now deduplicate some of the code; it also removes the need to keep calling `.map_pass_err()` everywhere.

### DIFF
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -892,7 +892,6 @@ impl<A: HalApi> BindGroup<A> {
         &self,
         bind_group_index: u32,
         offsets: &[wgt::DynamicOffset],
-        limits: &wgt::Limits,
     ) -> Result<(), BindError> {
         if self.dynamic_binding_info.len() != offsets.len() {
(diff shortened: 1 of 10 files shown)


Wumpf:
Hmm even more conflicts with #5884, still haven't recovered from the last. Well, I guess I'll have to suck it up; that pr is still incomplete as there's a bunch of tests missing.

Edit: while I don't have time for a full review right now, skimming this some more it doesn't look all that bad since the core areas touched are different - good the the command arcanization thing went in before! I love the work you're doing here, more than willing to redo my draft on top!

teoxoy:
Sorry for the conflicts, I can help rebase, just let me know.

nical (approved):

PR #5794, opened 2024-06-10
Separate renderpass arc resolve & renderpass consume on end

Wumpf:
**Connections**

* part of #1453
* This is pretty much exactly like a combination of these two PRs, except we're doing it now for render passes!
  * https://github.com/gfx-rs/wgpu/pull/5432
  * https://github.com/gfx-rs/wgpu/pull/5575

**Description**

As before with ComputePass, we now convert all incoming `RenderCommand` first to `ArcRenderCommand` and then consume only those upon pass ending.
Unlike with `ComputePass`, we already had `ArcRenderCommand` around (because of render bundles), I "only" had to fill in the gaps here.

**Testing**
Existing tests pass. Similar to how to compute passes, more are to come for the actual lifetime removal

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`. If applicable, add:
  - [x] `--target wasm32-unknown-unknown`
  - [x] `--target wasm32-unknown-emscripten`
- [x] Run `cargo xtask test` to run tests.
- [x] Add change to `CHANGELOG.md`. See simple instructions inside file.

PR #5884, opened 2024-06-26
Renderpass take resource ownership

### LINKED ISSUE #1453: Lifetimes on `RenderPass` make it difficult to use.

Wumpf:
**Connections**
* Fixes https://github.com/gfx-rs/wgpu/issues/1453
    * 🥳🥳🥳  
* Follow-up to #5794
* Equivalent to previous compute pass prs:
   * https://github.com/gfx-rs/wgpu/pull/5575
   * https://github.com/gfx-rs/wgpu/pull/5620
   * https://github.com/gfx-rs/wgpu/pull/5671 
   * #5768

**Description**
Just like ComputePass before, we now take ownership of all ingoing resources to a RenderPass right away. We also take ownership of the parent encoder, making it possible to convert `RenderPass<'a>`->`RenderPass<'static>` (as opt-in)

**Testing**
New ownership tests added (very similar to what we added for compute pass previously)

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,24 +41,38 @@ Bottom level categories:
 
 ### Major Changes
 
-#### Remove lifetime bounds on `wgpu::ComputePass`
+#### Lifetime bounds on `wgpu::RenderPass` & `wgpu::ComputePass`
 
-TODO(wumpf): This is still work in progress. Should write a bit more about it. Also will very likely extend to `wgpu::RenderPass` before release.
+`wgpu::RenderPass` & `wgpu::ComputePass` recording methods (e.g. `wgpu::RenderPass:set_render_pipeline`) no longer impose a lifetime constraint to objects passed to a pass (like pipelines/buffers/bindgroups/query-sets etc.).
 
-`wgpu::ComputePass` recording methods (e.g. `wgpu::ComputePass:set_render_pipeline`) no longer impose a lifetime constraint passed in resources.
+This means the following pattern works now as expected:
+```rust
+let mut pipelines: Vec<wgpu::RenderPipeline> = ...;
+// ...
+let mut cpass = encoder.begin_compute_pass(&wgpu::ComputePassDescriptor::default());
+cpass.set_pipeline(&pipelines[123]);
+// Change pipeline container - this requires mutable access to `pipelines` while one of the pipelines is in use.
+pipelines.push(/* ... */);
+// Continue pass recording.
+cpass.set_bindgroup(...);
+```
+Previously, a set pipeline (or other resource) had to outlive pass recording which often affected wider systems,
+meaning that users needed to prove to the borrow checker that `Vec<wgpu::RenderPipeline>` (or similar constructs)
+aren't accessed mutably for the duration of pass recording.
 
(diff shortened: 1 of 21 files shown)


nical (approved):

John-Nagle:
It turns out that this change badly broke Rend3. See

https://users.rust-lang.org/t/unexpected-borrowed-data-escapes-outside-of-closure-for-generic/119501/2

(Rend3 was abandoned by its author around the time this went in. But I'm stuck using it, and now I have to fix it.)

grovesNL:
@John-Nagle This PR loosens the lifetime requirements on the wgpu side. The change to `'static` lifetime in egui makes you require `forget_lifetime` otherwise everything would work the same as usual.

You'll need to call `forget_lifetime` before you share renderpasses with egui. See https://github.com/emilk/egui/pull/5149 for background and examples.

John-Nagle:
I get that, mostly. The worst trouble is near here:

https://github.com/BVE-Reborn/rend3/blob/trunk/rend3/src/graph/graph.rs#L403

This code is trying to manage the lifetime of the render pass and encoder from unsafe code.
I'm still trying to figure this out. See the "SAFETY" comment:
```
// SAFETY: There are two things which borrow this encoder: the render pass and the node's
// encoder reference. Both of these have died by this point.
```
That's not an assurance that this is safe code. It's a warning that there are lifetime requirements imposed on the caller. Ones which are changing now.  The code in that function does considerable unsafe manipulation of encoder and render pass lifetimes. So changes in this area are likely to cause undefined behavior.

I really don't want to touch that code if I can possibly avoid doing so. Is that possible?

Wumpf:
@John-Nagle lifetime requirements only got relaxed and more things are now handled as a runtime error rather than a compile time error which is what allows `forget_lifetime` to be a safe operation. My expectation is that everything still works fine except that there's some lifetime annotations that can just go away - any lifetime annotation that ties a resource (e.g. pipeline) to the lifetime of a render pass should no longer be needed (and in some cases may cause compile errors).
From the little I know about rend3 via its author the unsafe code in this area is exactly because previously the ownership management was only via lifetime annotation & there was no `forget_lifetime`. But keeping the old unsafe code should still work fine afaik

John-Nagle:
_The unsafe code in this area is exactly because previously the ownership management was only via lifetime annotation & there was no forget_lifetime._

I think that's about right. It looks to me like the old WGPU lifetime system forced complex unsafe ownership management. I'm trying to figure that out now. I think ordinary Rc and RefCell can do the job now. Thanks.

PR #5889, opened 2024-06-27
Added Wireframe Vertex Pulling example

MrFastDie:
**Connections**
I did not came up with that idea but ported it from C
[https://github.com/samdauwe/webgpu-native-examples/blob/master/src/examples/wireframe_vertex_pulling.c](samdauwe/webgpu-native-examples)

**Description**
It shows how to draw Wireframes.

**Testing**
I added the tests for every executable path.

However the `wireframe-thick` does not work right now - whats strange because it is indeed working in my WebGPU only project. Maybe it's an issue related to my graphics card, it would be great if someone could double check.

**Checklist**

- [X] Run `cargo fmt`.
  - [ ] `--target wasm32-unknown-emscripten` - fails everywhere for me seems to be related to `serde`
- [ ] Run `cargo xtask test` to run tests. - fails at some `wgpu-test` but running `cargo xtask test --workspace wgpu-examples wireframe_vertex_pulling` works with no problem


jimblandy (commented):
The code looks good. However, the example is pretty big, and I'm not sure it covers much new ground. We want to accumulate examples strategically, not flood people. I'm inclined to --- with respectful appreciation --- turn down this PR. What do folks think?

Wumpf:
I also think that additionally the value of this sample versus the amount of effort to get it polished & maintained is a bit too limited:
To begin with, there's not a lot of documentation that lays out what the example does: it lacks explaining why vertex pulling is needed, what are the pro/cons versus using lines, how does it compare with the [native feature for wireframes](https://wgpu.rs/doc/wgpu/struct.Features.html#associatedconstant.POLYGON_MODE_LINE) etc.
Skimming the shader code I think it would need significant changes as well: there's a lot of buffer packing with unclear value (why not use arrays of structs) and the naming & documentation overall is very lacking for what should be an educational example.
I realize this is just taken over from webgpu-native-examples, but that brings up another point: this means we're depending on an external license here (webgpu-native-examples is apache licensed so at least we'd have to include that here). If I'm not mistaken that would be a first as well.

I appreciate the effort, but for projects like that it can also be often helpful to engage early with maintainers & community, a good place for that is the [wgpu dev chat room](https://matrix.to/#/#wgpu:matrix.org)

MrFastDie:
Sure thing but for some reason when in Browser I cannot request the polygon mode neither on Mac nor on my arch system. It seems not to be present there why I choose to implement this one.

MrFastDie:
Sure thing but for some reason when in Browser I cannot request the polygon mode neither on Mac nor on my arch system. It seems not to be present there why I choose to implement this one.

MrFastDie:
Sure thing but for some reason when in Browser I cannot request the polygon mode neither on Mac nor on my arch system. It seems not to be present there why I choose to implement this one.

Wumpf:
Hmmm that's actually a good point that I didn't consider.
Maybe there's some place for this after all, but then we should make the sample a lot (!) smaller to cater to that exact need, having it document the particular issue of emulating `POLYGON_MODE_LINE`. Also, as already mentioned, imho a lot of things should be redone to be much more educational and focused on that.
Personally, I don't have the bandwidth right now to provide guidance to get towards that goal, but maybe someone else has.

MrFastDie:
Fair enough. I am reworking my own app anyways so that I at least change the things you mentioned already. I will also remove the cube and the points. I will remove the thick wireframes aswell when I don't find why it it not working.

By more educational, what you mean in particular? I can try to explain what it does in the README

Wumpf:
> By more educational, what you mean in particular? I can try to explain what it does in the README

that and some commentary on key spots in the code on how & why they work like that

PR #5896, opened 2024-07-01
Fix soundness issue with Snatchable

nical:
The code was written with the incorrect assumption that if no lifetime is specified in a method definition, then all lifetimes are elided to the lifetime of self. In fact only lifetimes in the returned value are elided to the lifetime of self, and other parameters get their own lifetimes.

Kudos to @teoxoy for catching the issue!

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


teoxoy (approved):
LGTM

Issue #2343, opened 2022-01-02
Support configuring `Surface`s with `Device`s that don't share the same underlying WebGL2 context


vsaase:
**Description**
When creating multiple canvases on web and multiple corresponding surfaces, all rendering is done in the last canvas associated to the last surface created.

**Repro steps**
Here is a modified hello-triangle main.rs file, run it with  cargo run-wasm --example hello-triangle --features webgl

```
use std::borrow::Cow;
use winit::{
    event::{Event, WindowEvent},
    event_loop::{ControlFlow, EventLoop},
    window::Window,
};

async fn run(event_loop: EventLoop<()>, windows: [Window; 2]) {
    let size = windows[0].inner_size();
    let instance = wgpu::Instance::new(wgpu::Backends::all());
    let surfaces: Vec<_> = unsafe {
        windows
            .iter()
            .map(|window| instance.create_surface(&window))
            .collect()
    };
    let adapter = instance
        .request_adapter(&wgpu::RequestAdapterOptions {
            power_preference: wgpu::PowerPreference::default(),
            force_fallback_adapter: false,
            // Request an adapter which can render to our surface
            compatible_surface: Some(&surfaces[0]),
        })
        .await
        .expect("Failed to find an appropriate adapter");

    // Create the logical device and command queue
    let (device, queue) = adapter
        .request_device(
            &wgpu::DeviceDescriptor {
                label: None,
                features: wgpu::Features::empty(),
                // Make sure we use the texture resolution limits from the adapter, so we can support images the size of the swapchain.
                limits: wgpu::Limits::downlevel_webgl2_defaults()
                    .using_resolution(adapter.limits()),
            },
            None,
        )
        .await
        .expect("Failed to create device");

    // Load the shaders from disk
    let shader = device.create_shader_module(&wgpu::ShaderModuleDescriptor {
        label: None,
        source: wgpu::ShaderSource::Wgsl(Cow::Borrowed(include_str!("shader.wgsl"))),
    });

    let pipeline_layout = device.create_pipeline_layout(&wgpu::PipelineLayoutDescriptor {
        label: None,
        bind_group_layouts: &[],
        push_constant_ranges: &[],
    });

    let swapchain_format = surfaces[0].get_preferred_format(&adapter).unwrap();

    let render_pipeline = device.create_render_pipeline(&wgpu::RenderPipelineDescriptor {
        label: None,
        layout: Some(&pipeline_layout),
        vertex: wgpu::VertexState {
[...]

PR #5901, opened 2024-07-03
Require a compatible surface for `request_adapter()` on WebGL2

### LINKED ISSUE #5190: request_adapter requires specific configuration to succeed with webgl feature enabled
**Description**
When enabling `webgl` feature and running on wasm target, `Instance::request_adapter` requires one of the following to be true, otherwise it will fail with no informative error:
1. Either: *only* `backends: wgpu::Backends::GL` should be used when creating the instance (see https://github.com/gfx-rs/wgpu/issues/3847 )
2. Or: a `compatible_surface` must be provided when calling `request_adapter` (e.g. created from an `OffscreenCanvas` etc.).

Note the documentation of `request_adapter` doesn't specify `compatible_surface` may be required in some cases, and the documentation of `backends` doesn't imply that choosing more than just `GL` may cause no adapter to be provided.

**Repro steps**
Run the following in wasm with the `webgl` feature - the `expect` will fail.
Running without the `webgl` feature - it works. 
Creating a `Surface` and passing it in the `compatible_surface` would also make it work, which is very surprising.
Using `wgpu::Backends::GL` for `backends` would also make it work, which is strange - I would not have expected lowering restrictions to make it work.
```rust
  backends: wgpu::Backends::all(),
  flags: wgpu::InstanceFlags::empty(),
  dx12_shader_compiler: wgpu::Dx12Compiler::default(),
  gles_minor_version: wgpu::Gles3MinorVersion::Automatic,
});

let adapter = instance
  .request_adapter(&wgpu::RequestAdapterOptions {
      power_preference: wgpu::PowerPreference::HighPerformance,
      force_fallback_adapter: false,
      compatible_surface: None,
  })
  .await
  .expect("Failed to create wgpu::Adapter");
```

Expected adapter to be available, or for it to be clearly documented in either `Instance::new` or `Instance::request_adapter` that `compatible_surface` may be **required** in some cases, or that allowing more backends may unintuitively cause adapter to be unavailable.
Observed: fails every time, unless using one of the workarounds described above.

**Extra materials**
N/A

**Platform**
Tried with wgpu 0.18 and 0.19.1 - same observed behavior.
Tried with Chrome 120.0.6099.234 and Firefox 120.0.1 - same observed behavior.
Running on MacOS 13.6.1

teoxoy:
When targeting WebGL2, it was always the case that a surface had to be created first before calling `request_adapter()`. This PR makes it explicit that the adapter returned is compatible with the surface.

Fixes https://github.com/gfx-rs/wgpu/issues/5190.
Adds validation to prevent https://github.com/gfx-rs/wgpu/issues/2343#issuecomment-1859486498.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,6 +134,9 @@ By @atlv24 in [#5383](https://github.com/gfx-rs/wgpu/pull/5383)
 When targeting WebGL2, it has always been the case that a surface had to be created before calling `request_adapter()`.
 We now make this requirement explicit.
 
+Validation was also added to prevent configuring the surface with a device that doesn't share the same underlying
+WebGL2 context since this has never worked.
+
 Calling `enumerate_adapters()` when targeting WebGPU used to return an empty `Vec` and since we now require users
 to pass a compatible surface when targeting WebGL2, having `enumerate_adapters()` doesn't make sense.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,6 +129,16 @@ Platform support:
 
 By @atlv24 in [#5383](https://github.com/gfx-rs/wgpu/pull/5383)
 
+#### A compatible surface is now required for `request_adapter()` on WebGL2 + `enumerate_adapters()` is now native only.
+
+When targeting WebGL2, it has always been the case that a surface had to be created before calling `request_adapter()`.
+We now make this requirement explicit.
+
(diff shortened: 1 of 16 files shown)


cwfitzgerald (approved):
This _definitely_ needs a changelog entry, but woah does this mean multiple surfaces will work now?

cwfitzgerald:
Should this close #2343?

teoxoy:
Multiple surfaces will work more predictably but you still can't present to multiple surfaces via a single device (which I thought what #2343 was asking for, hence I renamed it).

Issue #1674, opened 2021-07-16
fullscreen egl


ec1oud:
I wish it was possible to run on the Linux console, without a windowing system, as with Qt's eglfs platform.

On a raspberry pi, I'd probably be using gles in this mode, since vulkan on the pi 4 doesn't actually seem to be working yet (despite the claims) and I want to run a 64-bit OS (like Arch) and also have it working on the pi 3.

kvark:
I don't know how this is supposed to work. If you can provide code samples and knowledge, we'd be happy to consider this! PRs would even be better :)

niklaskorz:
This looks like a good reference: https://github.com/astefanutti/kms-glsl

Essentially, wgpu would have to do what a window server is doing (Direct Rendering Manager -> Generic Buffer Management -> EGL).
Maybe it would be possible to pass a [GBM surface](https://docs.rs/gbm/0.7.0/gbm/struct.Surface.html) to `create_surface` and initialize EGL on that? Then the initialization of DRM and GBM could be done outside wgpu.

niklaskorz:
I think we would also need to adjust `Instance::init` to use [`eglGetPlatformDisplayEXT(EGL_PLATFORM_GBM_MESA)`](https://stackoverflow.com/a/23166541) for getting the display from GBM.

niklaskorz:
Thinking about it, this might also be a nice step to allow the implementation of Wayland compositors on top of wgpu :) I think being able to switch between GLES 3 and Vulkan there could be particularly useful

kvark:
Sounds like an interesting direction! Could you summarize the pros/cons of this?

niklaskorz:
If you mean the pros and cons of writing a wayland compositor on top of wgpu, I'm afraid I don't have enough Wayland experience to comment on the usefulness of using wgpu as a renderer there. It looks like https://github.com/Smithay/smithay is currently using OpenGL ES 2, and so does Sway / wlroots, but [wlroots is working on Vulkan support](https://github.com/swaywm/wlroots/issues/642). Maybe @PolyMeilex could chime in and give some thoughts about whether this would be interesting for the Smithay project?

Now regarding adding GBM surface support to wgpu, one issue I see is that only EGL supports GBM. Also, GBM is not supported by the proprietary NVIDIA drivers right now, but might be sometime in the future. [EGLStreams](https://github.com/aritger/eglstreams-kms-example) on the other hand are supported by NVIDIA as an EGL extension (but by no one else).

[...]

PR #5908, opened 2024-07-04
 Initial Support for DRM interface in vulkan backend

puppymati:
This pull request introduces initial support for the Linux [DRM interface](https://en.wikipedia.org/wiki/Direct_Rendering_Manager) in wgpu. This enables wgpu to be utilized in environments where a windowing system is unavailable and, more importantly, allows wgpu to function as a renderer for Wayland compositors.

For additional context, please refer to issue #1674.

The implementation is straightforward, though there are known limitations due to my limited familiarity with wgpu internals:
 - Physical device selection is mostly hardcoded, as the method to dynamically select the device used by wgpu is unclear to me.
 - Validation and error handling are lacking, making the code prone to crashes.
 - There is no mechanism to guide wgpu in selecting the appropriate display mode, defaulting to the first available mode, which may not be the correct one.

Despite these issues, the code functions, and I demonstrated this by porting the hello_triangle example, available here: https://github.com/verdiwm/diretto/blob/main/examples/wgpu.rs

I also created a [fork](https://github.com/morr0ne/raw-window-handle/tree/drm-connector) of raw-window-handle, adding the necessary connector_id for display initialization.

While this pull request is not ready for merging, I am submitting it to illustrate the feasibility of DRM support and to receive feedback from the wgpu development team.

### DIFF
--- a/wgpu-hal/src/vulkan/instance.rs
+++ b/wgpu-hal/src/vulkan/instance.rs
@@ -1,5 +1,7 @@
 use std::{
     ffi::{c_void, CStr, CString},
+    mem::MaybeUninit,
+    num::NonZeroU32,
     slice,
     str::FromStr,
     sync::Arc,
@@ -9,6 +11,7 @@ use std::{
 use arrayvec::ArrayVec;
 use ash::{ext, khr, vk};
 use parking_lot::RwLock;
+use raw_window_handle::DrmWindowHandle;
 
 unsafe extern "system" fn debug_utils_messenger_callback(
     message_severity: vk::DebugUtilsMessageSeverityFlagsEXT,
@@ -288,6 +291,10 @@ impl super::Instance {
         // so that we don't have to conditionally use the functions provided by the 1.1 instance
         extensions.push(khr::get_physical_device_properties2::NAME);
 
+        extensions.push(ext::acquire_drm_display::NAME);
+        extensions.push(khr::display::NAME);
+        extensions.push(ext::direct_mode_display::NAME);
+
         // Only keep available extensions.
         extensions.retain(|&ext| {
             if instance_extensions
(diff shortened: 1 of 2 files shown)


teoxoy:
Looking at those Vulkan functions, it seems important to know what `VkPhysicalDevice` the surface belongs to. I think we need a new surface creation function that users of this API can pass a wgpu Adapter (`VkPhysicalDevice`) to + any other hints to select the display mode.

cwfitzgerald:
I don't think we need a new function, just another variant in the unsafe enum, in which we can have an adapter passed.

notgull:
> I also created a fork of raw-window-handle, adding the necessary connector_id for display initialization.

Please open a PR against the `raw-window-handle` repo so we can discuss this in greater depth.

puppymati:
> I don't think we need a new function, just another variant in the unsafe enum, in which we can have an adapter passed.

I think it would be more appropriate to pass a device rather than an adapter since the actual device isn't created until request_device is called. Would it also be logical to pass mode information here? I’m still figuring out what’s needed to choose the correct mode, but it’s clear that some sort of information must be provided.

I the case of the device, it is also unclear to me how to obtain the raw vkPhysicalDevice from the generic Device struct

> Please open a PR against the `raw-window-handle` repo so we can discuss this in greater depth.

I've opened a PR here: https://github.com/rust-windowing/raw-window-handle/pull/170.

MarijnS95:
> Physical device selection is mostly hardcoded, as the method to dynamically select the device used by wgpu is unclear to me.

The passed File Descriptor should be the opened `/dev/dri/cardX` character device, to which a `/dev/char/{major}:{minor}` symlink exists. These correspond to the major and minor in  `VK_EXT_physical_device_drm`'s `VkPhysicalDeviceDrmPropertiesEXT` struct, to help you associate the right physical device.
(Do keep in mind that multiple Vulkan drivers might exist for the same DRM device, such as AMDVLK and RADV)

---

Note that you can already enumerate the available `VkDisplayKHR`s directly via the `VK_KHR_display` extension, _for a given physical device_. Furthermore, the `EXT` extensions required here are described to mostly be for HMDs (VR headsets) to get exclusive access (while other connectors are still owned by a running compositor afaik). I think this implementation should get away with only using `VK_KHR_display` (at least by default, but do read on!), which also doesn't require/allow you to use the connnector ID (and gets rid of the `raw-window-handle` patch and dependency), as you anyway already need to build out API to select a pipe/plane and mode.

While `VK_KHR_display` in design looks very similar to DRM/KMS, it doesn't provide any interop by itself so that this extension can be supported regardless of the host OS supporting DRM. However, it seems the [`VK_KHR_get_display_properties2`](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_get_display_properties2.html) extension was built specifically to query extra information when enumerating `VkDisplayKHR`, such as an extension structure containing the DRM connector ID for that display... but in the current `vk.xml` no structure exists that _extends `VkDisplayProperties2KHR`_ just yet.

---

Regarding `EXT` extensions, on my Linux machine with an AMD card two drivers are already behaving differently. On the AMDVLK driver `vkGetDrmDisplayEXT()` always returns `ERROR_EXTENSION_NOT_PRESENT`, but `vkGetPhysicalDeviceDisplayPropertiesKHR()` returns an array with a `VkDisplayKHR` handle.
On the Mesa RADV driver the opposite is true: `vkGetDrmDisplayEXT()` returns a `VkDisplayKHR` handle (for every connector ID, even unconnected ones), but `vkGetPhysicalDeviceDisplayPropertiesKHR()` always returns an empty list. So a fallback to the EXT might be desired to at least support RADV (but shouldn't always be required)?

wgpu-hal/src/vulkan/instance.rs
```diff
@@ -531,6 +538,100 @@ impl super::Instance {
...
+            unsafe {
+                self.shared
+                    .raw
+                    .get_physical_device_properties2(device, &mut properties2)
+            };
+
+            if drm_props.primary_major == major.into() && drm_props.primary_minor == minor.into() {
+                physical_device = Some(device);
+
+                break;
```
MarijnS95, line 598: As mentioned before there can be >1 physical device for which this is true.

puppymati:
closing in favor of #7212

PR #5915, opened 2024-07-05
WIP: refactor(types)!: `SampleCount` newtype

ErichDonGubler:
**Connections**

N/A

**Description**

A mostly refactoring (but breaking) change I thought might be interesting. I'm not significantly invested in it. Each commit is atomic, and so the suggested review experience is commit-by-commit.

**Testing**

I believe I'm preserving 

**Checklist**

- [x] Run `cargo fmt`.


teoxoy:
We shouldn't have this type in `wgpu-core`'s public API as the spec allows 0 to be passed for sample counts. If a user passes a 0, they will get a validation error.

We can still have this type at the `wgpu` level if users find it helpful, it could also check that the given sample count is a power of 2.

ErichDonGubler:
Gonna abandon this for now, but will re-open once I can prioritize this.

PR #5916, opened 2024-07-05
refactor: satisfy `missing_docs,clippy::empty_docs`

ErichDonGubler:
**Connections**

N/A

**Description**

I think the title is descriptive enough. 🙂

**Testing**

Things still compile!

**Checklist**

- [x] Run `cargo fmt`.
- [x] ~~Add change to `CHANGELOG.md`. See simple instructions inside file.~~ I don't think that this is necessary.

### DIFF
--- a/wgpu-types/src/counters.rs
+++ b/wgpu-types/src/counters.rs
@@ -98,34 +98,22 @@ impl std::fmt::Debug for InternalCounter {
 }
 
 /// `wgpu-hal`'s internal counters.
+#[allow(missing_docs)]
 #[derive(Clone, Default)]
 pub struct HalCounters {
     // API objects
-    ///
     pub buffers: InternalCounter,
-    ///
     pub textures: InternalCounter,
-    ///
     pub texture_views: InternalCounter,
-    ///
     pub bind_groups: InternalCounter,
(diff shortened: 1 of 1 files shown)


wgpu-types/src/counters.rs
```diff
@@ -101,31 +101,31 @@ impl std::fmt::Debug for InternalCounter {
 #[derive(Clone, Default)]
 pub struct HalCounters {
     // API objects
-    ///
+    #[allow(missing_docs)]
```
cwfitzgerald: We can just stick one of these on the struct and not need to duplicate it everywhere

cwfitzgerald (approved):
g2g after nit

wgpu-types/src/counters.rs
ErichDonGubler: butbutbut what if we forget the others that _need_ docs 😱

Guess this is reasonable, if not my preference. Fixed with [my latest push](https://github.com/gfx-rs/wgpu/compare/d652aa5eca6b80bdfe69d35128b1fea8cfdeed77..8ace14736f33002cf9e630118d9300c669c9b27f).

PR #5919, opened 2024-07-06
Naga features for platform dependent compilation of msl & hlsl

Wumpf:
**Connections**
* Quite related to https://github.com/gfx-rs/wgpu/issues/3514

**Description**

Adds `cfg_alias` indirection to Naga's x_out features which allows us to introduce new features for platform dependent compilation. This in turn allows wgpu-hal to request shading language output more conditionally.

Drawbacks:
* build.rs also on Naga
* new feature names are kinda clunky

**Testing**
Eyeballing on Windows. Single sample of `cargo clean && cargo build -p wgpu --timings` looked even more promising than I would have imagined:

Before:
image

After:
image

1.4s cpu time saved on this setup for not actually touching anything 😮. When building wgpu this is even on the hot path, meaning the the wgpu build as a whole got that much faster!
Edit: as expected lots of noise here. Running again on trunk I got 7.0s and 6.2s here, so above trunk compilation is likely with cold file caches. ymmv.

TODO:
* [ ] changelog entry
* [ ] test this manually on mac or linux and report impact

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2185,6 +2185,7 @@ dependencies = [
  "arrayvec 0.7.4",
  "bit-set",
  "bitflags 2.6.0",
+ "cfg_aliases",
  "codespan-reporting",
  "diff",
  "env_logger",
--- a/naga/Cargo.toml
+++ b/naga/Cargo.toml
@@ -56,6 +56,9 @@ hexf-parse = { version = "0.2.1", optional = true }
 unicode-xid = { version = "0.2.3", optional = true }
 arrayvec.workspace = true
 
+[build-dependencies]
+cfg_aliases.workspace = true
+
 [dev-dependencies]
 diff = "0.1"
 env_logger = "0.11"
--- a/naga/Cargo.toml
+++ b/naga/Cargo.toml
@@ -24,7 +24,18 @@ default = []
 dot-out = []
 glsl-in = ["dep:pp-rs"]
 glsl-out = []
+
+## Enables outputting to the Metal Shading Language (MSL).
+##
+## This enables MSL output regardless of the target platform.
+## If you want to enable it only when targeting iOS/tvOS/watchOS/macOS, use `naga/msl-out-if-target-apple`.
 msl-out = []
+
+## Enables outputting to the Metal Shading Language (MSL) only if the target platform is iOS/tvOS/watchOS/macOS.
+##
+## If you want to enable MSL output it regardless of the target platform, use `naga/msl-out`.
+msl-out-if-target-apple = []
+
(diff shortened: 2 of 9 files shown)


teoxoy (commented):
I thought the backends won't be enabled if they won't be used on the target but it seems not because the conditional target compilation is setup in `wgpu-core` and happens after feature selection. It doesn't feel right that naga gets those new cfgs though; I would say that `wgpu-hal` is at fault here for not conditionally enabling `naga`'s features based on the target. We already have some target cfgs in `wgpu-hal/Cargo.toml`, I think we should add naga in those instead of adding new features to `naga`. Thoughts?

Wumpf:
> I would say that wgpu-hal is at fault here for not conditionally enabling naga's features based on the target.

agreed, if possible we should solve it on that level instead 🤔 
The problem though is that it's not a pure platform conditional: If I build on Windows without dx12 I expect HLSL to be disabled. If we were to put the naga dependency _with_ `hlsl-out` under a windows target conditional, it would always enable `hlsl-out` though even the user left didn't specify `dx12`.
Putting the `hlsl-out` feature inside `dx12 = [....]` gets us back to status quo where we would enable `hlsl-out` if a user on linux/mac enabled `dx12`.
Any ideas how to get around this?

teoxoy:
> Any ideas how to get around this?

`wgpu-hal/Cargo.toml` has some deps behind target cfgs, I thought we could move the naga features under those as well.

Wumpf:
it works if a feature should be dragged in or not on a per-target basis. The problem here is that it's a combination of feature _plus_ target condition that we want this to be active upon. I.e. what we want is:

_iff BOTH `feature=dx12` is active AND `target=windows` enable naga's `hlsl-out`_

The left hand side is easy since cargo supports that directly: `my_feature = [other_crate/feature]`
The right hand side is harder but doable with deps behind target cfg. E.g. `khronos-egl` is used with different features depending on the platform in wgpu-hal
```toml
[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
khronos-egl = { version = "6", features = ["dynamic"], optional = true }
#...
[target.'cfg(target_os = "emscripten")'.dependencies]
khronos-egl = { version = "6", features = ["static", "no-pkg-config"] }
```

The thing I'm struggling here with is to evaluate both conditions at the same time

teoxoy:
Ah yeah, that's problematic. I previously didn't fully grasp what you meant in your previous comment (https://github.com/gfx-rs/wgpu/pull/5919#issuecomment-2213656514).

I can't come up with a better solution either. I will take another look at this tomorrow.

teoxoy (approved):

Issue #5749, opened 2024-05-28
Validate `MultisampleState::sample_count` against all texture formats before attachment formats in render pipeline creation


ErichDonGubler:
Once <https://github.com/gpuweb/gpuweb/pull/4681> lands, we should conform to the new spec. revision by validating `MultisampleState::sample_count` to be supported by the adapter when creating pipelines. When strictly complying with WebGPU, this set of values should only be 1 or 4. On native, this set will be larger.

We will likely need to add a test a la `wgpu_test::pipeline::no_targetless_render`, as introduced in <https://github.com/gfx-rs/wgpu/pull/5715>.

Note that we are already validating that multisample counts are compatible with render pipeline attachments, so that the absence of this validation isn't causing us to make invalid requests to `wgpu-hal`. But adding this validation will certainly improve diagnostics.

ErichDonGubler:
I'm happy to mentor somebody on this. Marking as a `good first issue`.

andristarr:
If this still applies, I would like to start working on this. 

So question goes, as I was trying to mimic how NO_TARGETLESS_RENDER is set up, I dont see any references how it is being used. Can you please elaborate how this is being called?

Additionally, can I extract the said sample_count from the `TestingContext` I didnt find it yet?

I can see `ctx.device.limits().max_sampled_textures_per_shader_stage` on the testing context for example, but if I understand this correctly, we are not talking about texture sampling, rather multisampling anti aliasing.

ErichDonGubler:
@andristarr:

> If this still applies, I would like to start working on this.

I've assigned the issue to you now. 🙂

> … I was trying to mimic how NO_TARGETLESS_RENDER is set up, I dont see any references how it is being used. Can you please elaborate how this is being called?

Yes! Tests are set up in the `tests/`/`wgpu-test` crate using the `#[gpu_test]` attribute. `NO_TARGETLESS_RENDER` is annotated with it, and can be run with `cargo nextest run -p wgpu-test -- wgpu_test::pipeline::no_targetless_render`.

> Additionally, can I extract the said sample\_count from the `TestingContext` I didnt find it yet?

[...]

Issue #3274, opened 2022-12-08
Strict WebGPU compliance mode


teoxoy:
We seem to only issue a warning if we end up with a downlevel adapter/device.

https://github.com/gfx-rs/wgpu/blob/3d33ace8d7f4c410894a6d63092870c7eab0388d/wgpu-core/src/instance.rs#L303-L313

We need a mechanism that errors in such an event.
I was thinking a feature flag.

cwfitzgerald:
What's the use case of this? You can view the downlevel flags on the adapter and error out if you only want compliant adapters.

teoxoy:
> You can view the downlevel flags on the adapter and error out if you only want compliant adapters.

That's indeed a possibility but I think we'll need this distinction anyway to support the proposed `strict_extensions` field on the WGSL `Parser` (see https://github.com/gfx-rs/naga/issues/2155).

cwfitzgerald:
Couldn't we automatically set that if all the native-only extensions are false on the device?

cwfitzgerald:
Hmm, I guess the error messages change - "feature X not enabled" vs "unknown feature X"

teoxoy:
We should also use this mechanism to avoid exposing features that are not implemented in naga.

For example, `SHADER_F16` is currently exposed due to it being available through spir-v passthrough (but not via WGSL).

teoxoy:
This can now be added to `InstanceFlags`.

PR #5920, opened 2024-07-07
Validating sample count based on texture features

andristarr:
**Connections**
#5749 

**Description**
We are validating if the sample count is valid for _any_ texture format.

**Checklist**

- [ ] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -3043,6 +3043,17 @@ impl<A: HalApi> Device<A> {
                         break 'error Some(pipeline::ColorStateError::FormatNotColor(cs.format));
                     }
 
+                    if !format_features
+                        .flags
+                        .supported_sample_counts()
+                        .into_iter()
+                        .any(|count: u32| format_features.flags.sample_count_supported(count))
+                    {
+                        break 'error Some(pipeline::ColorStateError::UnsupportedSampleCount(
+                            desc.multisample.count,
+                        ));
+                    }
+
                     if desc.multisample.count > 1
                         && !format_features
                             .flags
--- a/wgpu-core/src/pipeline.rs
+++ b/wgpu-core/src/pipeline.rs
@@ -444,6 +444,10 @@ pub enum ColorStateError {
     FormatNotBlendable(wgt::TextureFormat),
     #[error("Format {0:?} does not have a color aspect")]
     FormatNotColor(wgt::TextureFormat),
+    #[error(
+        "No texture format supports sample count {0} with the current set of adapter features."
+    )]
+    UnsupportedSampleCount(u32),
     #[error("Sample count {0} is not supported by format {1:?} on this device. The WebGPU spec guarantees {2:?} samples are supported by this format. With the TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES feature your device supports {3:?}.")]
     InvalidSampleCount(u32, wgt::TextureFormat, Vec<u32>, Vec<u32>),
     #[error("Output format {pipeline} is incompatible with the shader {shader}")]


andristarr:
@ErichDonGubler did you think something along this line?

ErichDonGubler:
@andristarr: Hey there! Finally getting to this in my personal time. Sorry for the delay. 🙂

> did you think something along this line?

Not quite! You're still checking the features of a single _specified_ format of a single color attachment at a time, with no checking of the depth-stencil attachment. This is redundant with the check immediately below it. The point is to check whether _any_ format could support the multisample count for the render pipeline.

**You will need to iterate over _all_ formats at some point and check whether at least one of them matches.** There are a few (mutually exclusive) approaches that I see for that:

1. Collect all possible sample counts for all texture formats on adapter init. as a `HashSet<u32>`[^1]. Pass this set to a `Device` when created. When creating a `RenderPipeline`, this set gets checked for membership. If the sample count for an attachment is not a member of the set, then reject with a validation error. 
2. Iterate over texture formats to find one that has a potentially matching sample count. If one is found, then proceed with creating the render pipeline. Otherwise, reject with a validation error.
3. Mix (1) and (2) by creating the `HashSet` only when a render pipeline is requested to be built, and check it as with (1).

The only significant difference between these is a trade-off between when work is done, and where that work is stored. **I'd recommend (1), since it should be a pretty small amount of data to add to a device**, while iterating over every texture format sounds pretty compute-intensive to do 3 times with (2). (3) sounds unnecessarily complicated; having variance in the computational load of validation in an API depending on whether you've called it before is a complication I'd rather avoid.

[^1]: We could later (as a separate step) optimize this to be a single-byte bitset, where each bit index corresponds to a power of two. This works because we only support powers of 2 up to 64 (2^8, or 8 bits).

ErichDonGubler (changes requested):
Requesting changes, as described in my previous comment.

I'm going to convert this back to a draft, since it seems obvious to me that this needs to be reworked.

teoxoy:
> **You will need to iterate over _all_ formats at some point and check whether at least one of them matches.**

When we previously talked about this I thought the plan was to error if the count is not 1 or 4 if we are in compliance mode (https://github.com/gfx-rs/wgpu/issues/3274) so that we match the spec validation as close as possible.

> Note that we are already validating that multisample counts are compatible with render pipeline attachments, so that the absence of this validation isn't causing us to make invalid requests to wgpu-hal. But adding this validation will certainly improve diagnostics.

Now that I reread the issue, I see the motivation was improving diagnostics but I don't see how adding this validation will improve the diagnostics. The validation error will say "sample count is not valid for any formats" instead of the current "sample count is not valid for format X". The fact that the sample count is not valid for any formats supported by the device doesn't seem relevant in this case since the user supplied a specific format.

> descriptor.[count](https://www.w3.org/TR/webgpu/#dom-gpumultisamplestate-count) must be either 1 or 4.

The reason the validation in the spec is written this way is because all formats in the spec support sample counts of only 1 and 4. Once the spec adds support for the optional sample counts of 2, 8 and 16 (https://github.com/gpuweb/gpuweb/issues/4446) the validation will be changed to what we currently implement.

I suggest we just keep the status quo.

andristarr:
@ErichDonGubler I am more than happy to do the work laid out by your comment (especially thank you for explaining because I wasn't familiar with the domain!), but I take it @teoxoy 's comment will render this issue obsolete.

Let me know if you want me to do the implementation or just drop the PR.

ErichDonGubler:
@andristarr: You are right; @teoxoy and I took some time to discuss this, and he's persuaded me that my idea for validation seems too heavyweight for the modest diagnostic improvement. We're now considering <https://github.com/gfx-rs/wgpu/issues/6039> instead, which is not _yet_ implementable. Let's drop this PR.

Sorry for taking your time this way. I think it's an error in our maintainership when somebody (like you) starts work on something that we didn't get consensus on properly, and I apologize for that.

PR #5929, opened 2024-07-09
[xtask test] Implement `--nocapture` flag.

jimblandy:


xtask/src/test.rs
```diff
@@ -81,8 +88,8 @@ pub fn run_tests(shell: Shell, mut args: Arguments) -> anyhow::Result<()> {
             "--retries",
             &retries,
         ])
+        .args(nocapture_flags)
         .args(args.finish())
-        .quiet()
```
ErichDonGubler, line 85: question: Why are we removing this flag?

ErichDonGubler (approved):
Tentatively approving, pending a question.

cwfitzgerald:
Why do we need this, it should just pass through excess flags

xtask/src/test.rs
jimblandy: Carelessness. I'm sorry.

jimblandy:
I'm sorry, I think this PR is unneeded and a waste of your time.

I think I must have passed `cargo xtask test -- --nocapture`, as is needed for `cargo test`, gotten the complaint "error: failed to parse test binary arguments `--nocapture`: arguments are unsupported", jumped to conclusions, dashed off to fix things, and inadvertently changed my invocation to `cargo xtask test --nocapture`, which works with or without the PR.

jimblandy:
I understand why we need `cargo xtask test`, but when I talk about how inattentive or naive users may be confused by custom harnesses --- I mean *me*

PR #5932, opened 2024-07-10
[wgpu-core] fix `.zip()` usages

teoxoy:
Fixes `.zip()` usages that were originally introduced in 4a19ac279c4f81aacedb1d215c884c10fe115275 and d8b0b5975d1c4957b177c8d450d19d3974ef7a0f (part of https://github.com/gfx-rs/wgpu/pull/5897).


nical (approved):
I'd prefer to use "skip" instead of "drop" since drop has a specific connotation in rust.

teoxoy:
Talked with Nical and we settled on `ZipWithProperAdvance`.

PR #5939, opened 2024-07-11
naga: Fix reference to `serde` feature.

waywardmonkeys:
**Description**
naga doesn't have a `serde` feature, instead having separate `serialize` and `deserialize` features, so things that want to modify the serde handling must check for either of those, not for `serde` itself.

**Testing**
Compilation + running the tests.

**Checklist**

- [x] Run `cargo fmt`.


teoxoy (approved):
Thanks!

PR #5941, opened 2024-07-11
Use `derive` feature on `serde` rather than `serde_derive`

waywardmonkeys:
**Description**
The current code works, but `serde` documents that the feature to use is `derive` (which then happens to use the `serde_derive` implicit feature).

See https://serde.rs/derive.html

**Testing**
Code compiles and passes tests.

**Checklist**

- [x] Run `cargo fmt`.


teoxoy (approved):
Thanks!

PR #5942, opened 2024-07-11
wgc: Use explicit feature for `raw-window-handle`

waywardmonkeys:
**Description**
The 2024 edition of Rust is planning to not allow implicit features. Here, we now use an explicit feature for `raw-window-handle` in `wgpu-core`. This also removes an implicit feature for `ron` which was not needed (shown when doing a `cargo add wgpu-core`).

This helps to prepare for the coming day when explicit features will be required.

**Testing**
Compiles and passes tests.

**Checklist**

- [x] Run `cargo fmt`.


teoxoy (approved):
Thanks!

PR #5946, opened 2024-07-12
[wgpu-core] Make the `StagingBuffer` implementation more robust

teoxoy:
Makes the `StagingBuffer` implementation more robust and uses of it more consistent.


nical (approved):
Nice!

PR #5952, opened 2024-07-13
Atomic64 fixes

### LINKED ISSUE #5784: `wgpu_types::Features::SHADER_INT64_ATOMIC_ALL_OPS` should also request `shader_shared_int64_atomics`
I missed this reviewing #5383, but in `PhysicalDeviceFeatures::from_extensions_and_requested_features` in `wgpu-hal/src/vulkan/adapter.rs`, when we populate `shader_atomic_int64`, we should set both `shader_buffer_int64_atomics` *and* `shader_shared_int64_atomics`, since we require both in `PhysicalDeviceFeatures::to_wgpu`.

cc @atlv24 @teoxoy

### LINKED ISSUE #5887: SHADER_INT64_ATOMIC_MIN_MAX can never be requested on Vulkan
SHADER_INT64_ATOMIC_MIN_MAX never seems to be able to be requested. Somewhere in wgpu-hal it thinks my GPU (RTX 3080) lacks support, but it shouldn't.

wgpu rev `abba12a`

### LINKED ISSUE #5888: [Naga] SHADER_INT64_ATOMIC_MIN_MAX issue
```rust
@group(0) @binding(8) var<storage, read_write> meshlet_visibility_buffer: array<atomic<u64>>;

@fragment
fn fragment(@builtin(position) position: vec4<f32>) {
    atomicMax(&meshlet_visibility_buffer[0u], u64(position.z));
}
```

```
error: 
  ┌─ foo.wgsl:5:16
  │
5 │     atomicMax(&meshlet_visibility_buffer[0u], u64(position.z));
  │                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ naga::Expression [3]

Entry point fragment at Fragment is invalid: 
        Expression [3] is invalid
        Used by a statement before it was introduced into the scope by any of the dominating blocks
```

Tested with naga-cli rev `82210e1c`

### LINKED ISSUE #5893: Naga emits wrong HLSL for atomicMax() on u64 buffer
**Description**
Naga should be emitting `buffer.InterlockedMax64`, not `buffer.InterlockedMax`.

This only applies to `RWByteAddressBuffer` according to the [spec](https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_Int64_and_Float_Atomics.html#interlockedmax).

This issue also applies to the other atomic functions.

**Repro steps**
```rust
@group(0) @binding(0) var<storage, read_write> buf: atomic<u64>;

atomicMax(&buf, 3147483647lu);
```

**Platform**
Naga rev `82210e1c`

JMS55:
**Connections**
* Fixes https://github.com/gfx-rs/wgpu/issues/5888
* Fixes https://github.com/gfx-rs/wgpu/issues/5887
* Fixes https://github.com/gfx-rs/wgpu/issues/5784
* Fixes https://github.com/gfx-rs/wgpu/issues/5893

**Description**
* Apply fix from issue written by @jimblandy 
* On Vulkan, check for atomic64 PHD feature when inspecting the instance to get supported adapter extensions (the code for vulkan features/extensions in wgpu is seriously confusing, and idk if PhysicalDeviceShaderAtomicInt64Features::default() is correct here, but it seems to work and follow the existing code pattern)
* On Vulkan, also request shader_shared_int64_atomics in addition to the buffer variant
* In the HLSL backend, write e.g. InterlockedMax64 instead of InterlockedMax when performing an atomic operation on a buffer

**Testing**
* Bevy shaders

**Checklist**

- [ x ] Run `cargo fmt`.
- [ x ] Run `cargo clippy`. If applicable, add:
  - [ x ] `--target wasm32-unknown-unknown`
  - [ x ] `--target wasm32-unknown-emscripten`
- [ x ] Run `cargo xtask test` to run tests.
- [ x ] Add change to `CHANGELOG.md`. See simple instructions inside file.


jimblandy:
@JMS55 So, why wasn't `tests/in/atomicOps-int64-min-max.wgsl`  broken by the absence of the fix in  `naga/src/front/wgsl/lower/mod.rs`?

JMS55:
I couldn't figure that out. My best guess is that maybe the fact that the second argument is not an expression matters. Maybe a constant vs expression in the second argument changes how things get parsed or something.

jimblandy:
The clippy warnings are specific to this branch, they don't happen on trunk. I can see them with:

```
cargo +1.76 clippy -p naga --features=hlsl-out
```

I would like to have a snapshot test that is affected by the change to `naga/src/front/wgsl/lower/mod.rs`. I'm uncomfortable that we have no tests that can detect that fix.

JMS55:
I didn't change that code though, so idk why clippy is complaining only on this branch... I mean I can make the fixes as part of this PR if you want, up to you.

jimblandy:
> I didn't change that code though, so idk why clippy is complaining only on this branch... I mean I can make the fixes as part of this PR if you want, up to you.

Yes, please change this PR so it does not break CI on trunk when it is merged.

jimblandy:
> I would like to have a snapshot test that is affected by the change to `naga/src/front/wgsl/lower/mod.rs`. I'm uncomfortable that we have no tests that can detect that fix.

I volunteered to work on this in chat.

jimblandy:
> @JMS55 So, why wasn't `tests/in/atomicOps-int64-min-max.wgsl` broken by the absence of the fix in `naga/src/front/wgsl/lower/mod.rs`?

It seems like the reason is, the final operands to the `atomicMax` and `atomicMin` calls in the snapshot are all simple expressions that don't need to be emitted. I've pushed a commit that uses a variety of final operands, and that causes a panic without the fix.

jimblandy (approved):
This looks good. I've pushed a commit to address the only nit I had.

PR #5956, opened 2024-07-14
[wgpu-hal] Migrate d3d12 backend over to `windows-rs`

### LINKED ISSUE #3207: Migrate to Windows-rs from winapi

### LINKED ISSUE #5813: The ComPtr API is unsound.
**Description**
The ComPtr API let's you de reference a null pointer in safe code.

**Repro steps**

1. Safely construct a null ComPtr with  the [`ComPtr::null`](https://github.com/gfx-rs/wgpu/blob/3b5051071a09e99cf44a29ddd4b06f79d7e115ba/d3d12/src/com.rs#L14-L16) function.
A.2 [Clone](https://github.com/gfx-rs/wgpu/blob/3b5051071a09e99cf44a29ddd4b06f79d7e115ba/d3d12/src/com.rs#L65-L73) your new pointer.
A.3 Look on in horror as [`as_unknown`](https://github.com/gfx-rs/wgpu/blob/3b5051071a09e99cf44a29ddd4b06f79d7e115ba/d3d12/src/com.rs#L47-L50) de references a null pointer.
B.2 Be a bit careless and let rust autoderef your pointer under your nose.
B.3 Watch as your pointer [`Deref` impl]((https://github.com/gfx-rs/wgpu/blob/3b5051071a09e99cf44a29ddd4b06f79d7e115ba/d3d12/src/com.rs#L85-L91)) executes a rapid unplanned program exit.

**Possible solutions**

The simplest solution would be two fold. The clone impl should only call `AddRef` when the pointer isn't null. The deref impl should be replaced by a deref function that's either safe and returns a `Option<&T>` or is unsafe and returns a &T.

I'm also attempting a solution where `ComPtr` is turned into a basic wrapper over a *mut ptr and a new type is introduced that always contains a valid (non null and pointing to a valid allocation) pointer to a Interface. The problem is it touches a lot of code and I'm not very familiar with this repo.

**Platform**
Windows

MarijnS95:
> [!CAUTION]
> This is a **draft** that is still a work-in-progress! See this as a reminder that work has actively been performed to get the migration done, and to familiarize yourself (as reader) with `windows-rs`!
>
> **I am still heavily self-reviewing, cleaning and refactoring this PR - reviewing will be a _waste of time_ until this is taken out of _draft_!**

**Connections**
Closes #3207 (once `wgl.rs` is also over)
Fixes #5813

**Description**
`winapi` is barely typed, doesn't care about object lifetime, and is unmaintaned above all. Microsoft officially provides up-to-date and completely typed bindings that make it much more natural, convenient, **and safe** to interface _any_ API on Windows systems, including DXGI and Direct3D12! By migrating we make it much easier to write, review and maintain the d3d12 backend, and have improved interoperability with support crates like `gpu-allocator` that already migrated long ago.

One thing you'll also see is the `windows` crate embracing `Result` and `#[must_use]` - there are still a hanful of clippy errors pointing out that the original code forgot to check the returned `HRESULT`, leading to potential UB.

**Testing**
Still in progress - a qick showdown was done using:

```sh
WGPU_BACKEND=d3d12 cargo run --bin wgpu-examples shadow
```

**Checklist**

- [x] Run `cargo fmt`.

**TODO**

- [ ] Clean up unused workspace imports?
- [ ] Migrate straight to `windows 0.67` (with e.g. https://github.com/rust-windowing/raw-window-handle/issues/171);
- [ ] Release `gpu-allocator`;
- [ ] Investigate (probably broken) lifetime and ownership of the various `enum Surface..` holders...


wgpu-hal/src/dx12/suballocation.rs
```diff
@@ -8,6 +8,9 @@ use committed as allocation;
 #[cfg(feature = "windows_rs")]
 use placed as allocation;
 
+// TODO: windows_rs is the default now, but we cannot actually remove the slow path as it is
+// used in a workaround for Intel Xe?  (without updating any of the comments/documentation)
+
 // This exists to work around https://github.com/gfx-rs/wgpu/issues/3207
 // Currently this will work the older, slower way if the windows_rs feature is disabled,
 // and will use the fast path of suballocating buffers and textures using gpu_allocator if
```
MarijnS95: @Elabajaba @xiaopengli89 is https://github.com/gfx-rs/wgpu/pull/3668 still a problem on Intel Xe? I'd like to clean up the now-meaningless `cfg(windows_rs)` feature, but probably have to keep `mod committed` around for this Xe workaround?

wgpu/src/api/surface.rs
MarijnS95, line 1: Can anyone help document what kind of ownership transition is expected in `SurfaceTargetUnsafe`? When passing raw pointers to specifically COM objects, are they a borrow, or are we given ownership over that COM object?

If we're given  ownership the handling code can easily call `ISomething::from_raw()`. Otherwise we'll have to wrap it in `ManuallyDrop` or (better yet) first increment the refcount (`Ixxx::from_raw_borrowed().unwrap().clone()`).

wgpu-hal/src/dx12/suballocation.rs
xiaopengli89: @MarijnS95 This seems to be a driver issue, and it would be best to keep a workaround via options.

MarijnS95: This workaround has since been refactored to not be as intrusive to the implementation. Note that `gpu-allocator` has both committed and placed resource support in its API [^1].

I'm still a bit surprised by this feature from @Elabajaba, because we've found that on some hardware and types of textures, placed resources are much slower, because the hardware doesn't have enough knowledge to put them in the right space of memory. If anything both committed and placed resources should be used throughout the codebase.

[^1]: I don't think there was much use in having this feature _directly inside `gpu-allocator`_ since committed resources heavily clobber/confuse our API without doing _anything_ on the allocated blocks (it's just a call to the `ID3D12Device`). The implementation does however count this memory usage towards a "compatible" memory type which might help @nical with the counters.

MarijnS95: @xiaopengli89 I'd love to know if the issue on Xe can be reproduced and how, your issue from https://github.com/gfx-rs/wgpu/issues/3552 doesn't contain easily-runnable sample code to demonstrate the problem.

I'd love to see if it can be solved by _only_ allocating render targets with committed resources, and keep the rest placed on the heap.

Elabajaba: > This workaround has since been refactored to not be as intrusive to the implementation. Note that `gpu-allocator` has both committed and placed resource support in its API [1](#user-content-fn-1-a4aaadbeda1e73fb5f4201416273c529).
> 
> I'm still a bit surprised by this feature from @Elabajaba, because we've found that on some hardware and types of textures, placed resources are much slower, because the hardware doesn't have enough knowledge to put them in the right space of memory. If anything both committed and placed resources should be used throughout the codebase.

I'm not a graphics programmer and didn't know that (my entire knowledge of gpu memory allocation is fuzzy memories of some asawicki blog posts). Also I don't think the webgpu api gives you that level of control to choose what type of allocation you want (and dx12 seems to make this a lot more painful than vulkan, where you can query what allocation type a resource would prefer).

The feature gating was just the simplest way I could come up with so that Mozilla wouldn't have to vendor windows-rs at the time, while also vastly improving performance vs only committed resources.

edit: Also some things are created with committed resources in the dx12 hal level, I just don't remember what they are and can't check currently.

MarijnS95: Yeah, on Vulkan there's an API that specifies if a dedicated allocation (synonimous to Dx12 committed resources) is preferred or even required. I don't know if `wgpu` and `gpu-alloc` take care of this, at least our `gpu-allocator` crate does.

You might have been reading https://www.asawicki.info/articles/memory_management_vulkan_direct3d_12.php even though it doesn't explicitly seem to list that placed resources are always better, but it becomes obvious that they allow for much finer grained placement and thus less wasted memory and likely also (opaque to us) extra driver overhead.

There's an old quote from an Nvidia [DX12 Do's And Don'ts](https://developer.nvidia.com/dx12-dos-and-donts) blog that is no longer available where it was stated that committed resources are always better. A newer post says that this is only the case for ucompressed resources _when aliasing_: https://developer.nvidia.com/blog/advanced-api-performance-memory-and-resources/

Elabajaba:
Tested some of the wgpu examples and it seems to work fine on windows 11 + amd, apart from the `todo!` panic on dropping window handles (in the `impl Drop for SurfaceTarget` block).

ErichDonGubler:
Oh, don't you worry about testing. We'll put this through the entire WebGPU CTS on Firefox for you, just to make sure this safe to merge. 🤪

I'm quite excited.

wgpu/src/api/surface.rs
MarijnS95: On the surface it may appear that `SurfaceTargetUnsafe` is not stored anywhere, and should hence be safe to carry temporary pointers that the caller has to keep valid while a call is ungoing (as usual). That is also what the changelog seems to suggest:
https://github.com/gfx-rs/wgpu/blob/6351a75b0cd9ec26948ca4934a5316fd781c567e/CHANGELOG.md?plain=1#L717-L718

However, reading further into `dx12`'s `Instance::create_surface_*()` functions that I had to touch in this PR, it becomes obvious that the values end up being copied into an `enum SurfaceTarget`. For COM objects, I can "`.clone()`" the pointer here which is equivalent to an `.AddRef()` call, but for `HANDLE`s it is not that easy.

However, all paths leading up to this (i.e. filling `SurfaceTargetUnsafe` and coming up with a custom `'window` bound) are `unsafe`, so this could also be taken as a safety precaution for the reader (i.e. `HANDLE` must not be freed during `'window`).

---

We could additionally also have the safe owned versions of at least the COM objects (`IDCompositionVisual` and `SwapChainPanel`) inside the safe `SurfaceTarget`?

MarijnS95:
@ErichDonGubler that would be very nice, but not before we answered some basic questions above about the validity of `SurfaceTargetUnsafe`. This is quite tricky because its values are stored inside `Surface`, requiring the caller to keep it alive, or us to extend the refcount. I'd love to receive some feedback on how `wgpu` expects callers to utilize this, and hence what we need to do on the backend.

ErichDonGubler:
@MarijnS95: Ah, of course. The `Surface` APIs are actually not used by Firefox, but we can still test `wgpu-core`-centric functionality. I think we can safely decouple Firefox test runs from figuring out questions with those APIs, though OFC merging a PR to mainline history would still be blocked on keeping things safe. 🙂

MarijnS95:
@ErichDonGubler correct, I don't think we can land this without settling, and documenting it. If my mind serves me correctly the original backend implementation does not assume ownership and does not drop the pointers, which we could replicate by wrapping the COM objects in `ManuallyDrop`.

wgpu/src/api/surface.rs
cwfitzgerald: Sorry for the delay in getting you this information.

I think we should have the unsafe contract be that "whatever handle you pass, the user _must_ keep it valid for the lifetime of 'window. By contract we do not own it, however if possible in the backend add and remove a reference for our own connivence.

We _can_ move IDCompositionVisual/SwapChainPanel and other COM objects into the safe enum and take them by value, as we can hold on to them ourselves.

MarijnS95: > I think we should have the unsafe contract be that "whatever handle you pass, the user must keep it valid for the lifetime of 'window. By contract we do not own it, however if possible in the backend add and remove a reference for our own connivence.

Those two points pretty much contradict themselves, as there's no point to keep the lifetime of the refcounted handle up when `wgpu-hal` does the same for both convenience and correctness.

I've already documented that in the most recent push to this PR a few weeks ago (and will push a rebase in a moment). Anticipating to push a change on top that moves `IDCompositionVisual`/`SwapChainPanel` from `SurfaceTargetUnsafe` to `SurfaceTarget`! That should put the burden on the caller instead to:

1. Use the `windows-rs` crate (if they don't already);
2. If not, correctly wrap their pointer using either `from_raw()` or `from_raw_borrowed()` depending on what they have.

Leaving just the `SurfaceHandle` from `DirectComposition` to be a "Borrow" (since I'm unclear if this `HANDLE` can be cloned or freed using standard APIs) and the `HWND` from the `RawWindowHandle` constructor.

cwfitzgerald: Sounds good, I'll take a more proper look at this PR tomorrow - thanks for keeping with it!

wgpu-hal/Cargo.toml
```diff
@@ -154,20 +157,12 @@ bit-set = { workspace = true, optional = true }
 range-alloc = { workspace = true, optional = true }
 gpu-allocator = { workspace = true, optional = true }
 hassle-rs = { workspace = true, optional = true }
+# For core macros. This crate is also reexported as windows::core.
+windows-core = { workspace = true, optional = true }
```
cwfitzgerald, line 161: I feel like I'd prefer to just use `windows::core`.

wgpu-hal/src/auxil/dxgi/factory.rs
```diff
@@ -51,50 +51,92 @@ fn should_keep_adapter(adapter: &dxgi::IDXGIAdapter1) -> bool {
...
+    pub fn unwrap_adapter2(&self) -> &Dxgi::IDXGIAdapter2 {
+        self.as_adapter2().unwrap()
+    }
+
+    // pub fn as_adapter6(&self) -> Option<&Dxgi::IDXGIAdapter6> {
+    //     match self {
+    //         Self::Adapter1(_) | Self::Adapter2(_) | Self::Adapter4(_) => None,
+    //         Self::Adapter6(f) => Some(f),
+    //     }
+    // }
```
cwfitzgerald: If we don't need this, lets just nuke it.

wgpu-hal/src/dx12/adapter.rs
```diff
@@ -76,45 +79,41 @@ impl super::Adapter {
...
-        unsafe {
-            adapter.unwrap_adapter2().GetDesc2(&mut desc);
-        }
+        let desc = unsafe { adapter.unwrap_adapter2().GetDesc2() }.unwrap();
 
         let device_name = auxil::dxgi::conv::map_adapter_name(desc.Description);
 
-        let mut features_architecture: d3d12_ty::D3D12_FEATURE_DATA_ARCHITECTURE =
+        let mut features_architecture: Direct3D12::D3D12_FEATURE_DATA_ARCHITECTURE =
             unsafe { mem::zeroed() };
```
cwfitzgerald: Can't this be default() instead of zeroed?

wgpu-hal/src/dx12/mod.rs
```diff
@@ -44,17 +44,254 @@ mod suballocation;
...
+///
+/// See also <https://github.com/microsoft/windows-rs/pull/2361#discussion_r1150799401> and
+/// <https://github.com/microsoft/windows-rs/issues/2386>.
+///
+/// # Safety
+/// Performs a [`mem::transmute_copy()`] on a refcounted [`Interface`] type.  The returned
+/// [`mem::ManuallyDrop`] should _not_ be dropped.
+pub unsafe fn borrow_interface_temporarily<Src: Interface, Dst>(
+    src: &Src,
+) -> mem::ManuallyDrop<Option<Dst>> {
```
cwfitzgerald: Probably worthwhile having a `for<'a> &'a Dst: From<&'a Src>` bound on this function to make limit the scale of mistake you can make with it.

cwfitzgerald (approved):
This is really amazing stuff! I have a few really small comments, but none that I think should block this landing. Let me make sure moz has everything they need to land this PR, and we can get rolling on this.

wgpu-hal/src/dx12/adapter.rs
MarijnS95: Thanks for the reminder, replaced all `unsafe { mem::zeroed() }` where sensible (and helping the compiler to catch one more trivial FRU initializer).

wgpu-hal/src/dx12/mod.rs
MarijnS95: I think I needed an interface cast in another crate where this was introduced (and should apply the suggestion there - quick reminder that interface hierarchies weren't a thing in `windows-rs` back then) but no such thing is ever needed in `wgpu` it seems so `Dst` is dropped entirely in favour of a renamed `Src`.

wgpu-hal/Cargo.toml
MarijnS95: Perhaps the comment should be clarified instead: this is an upstream bug where `::windows_core::` is always emitted instead of `::windows::core::` where possible.

Thanks to `^::` we can't `pub(crate) use windows::core as windows_core;` where the interface macro is called.

cwfitzgerald: Ahhh, the macros themselves need this dependency. Gottcha. Probably worth clarifying the comment

jimblandy:
@MarijnS95 Could `wgpu_hal::dx12::Device::mem_allocator` stop being an `Option` now?

It seems like `Device::exit` takes `self` by value, so assigning `None` to it there is just the same as dropping it, so that line can be deleted.

jimblandy (approved):
This looks fantastic. Assuming that this depends on windows 0.58, it should be fine for Firefox.

teoxoy (approved):
Things look straightforward now, great stuff!

teoxoy:
One thing I noticed is that we still use `libloading`. Is that necessary given that `libloading` itself depends on `windows_targets` which does the linking in the `windows` crate as well? I think we could instead call the functions exposed by `windows`.

MarijnS95:
> One thing I noticed is that we still use `libloading`. Is that necessary given that `libloading` itself depends on `windows_targets` which does the linking in the `windows` crate as well? I think we could instead call the functions exposed by `windows`.

It should only do the linking when those symbols are actually used/called, which they are currently not. I assumed `libloading` was preferred since `extern` functions would have been easy to use with `winapi` already, but this way makes the relevant DLLs optional to support (much?) older Windows?

If there's no particular reason to use runtime loading I'd be very happy to throw the extra code all right out, but it's a bit of a pity since it has already been written (ported from the linked functions inside the `windows` crate) in the first place.

cwfitzgerald:
Yeah the idea is to allow running on windows 7-8.1, as the d3d12 functions don't exist there. How valueable that is is debatable, but that's what we've been doing.

wgpu-hal/src/auxil/dxgi/result.rs
```diff
@@ -1,37 +1,46 @@
...
     fn into_device_result(self, description: &str) -> Result<O, crate::DeviceError>;
 }
-impl HResult<()> for i32 {
-    fn into_result(self) -> Result<(), Cow<'static, str>> {
-        if self >= 0 {
-            return Ok(());
-        }
+impl<T> HResult<T> for windows::core::Result<T> {
+    fn into_result(self) -> Result<T, Cow<'static, str>> {
+        // TODO: use windows-rs built-in error formatting?
```
MarijnS95, line 11: I think I forgot to look at simplifying this code...

wgpu-hal/src/auxil/dxgi/result.rs
```diff
@@ -1,37 +1,46 @@
...
     }
-    fn into_device_result(self, description: &str) -> Result<(), crate::DeviceError> {
+    fn into_device_result(self, description: &str) -> Result<T, crate::DeviceError> {
         #![allow(unreachable_code)]
 
+        let err_code = if let Err(err) = &self {
+            Some(err.code())
+        } else {
+            None
+        };
```
MarijnS95, line 29: ... And with that, getting rid of this ownership-workaround.

wgpu-hal/src/dx12/descriptor.rs
```diff
@@ -38,32 +40,30 @@ pub(super) struct GeneralHeap {
...
+                NodeMask: 0,
+            };
+            unsafe { device.CreateDescriptorHeap::<Direct3D12::ID3D12DescriptorHeap>(&desc) }
                 .into_device_result("Descriptor heap creation")?
         };
 
-        null_comptr_check(&raw)?;
-
         Ok(Self {
             raw: raw.clone(),
```
MarijnS95, line 60: No need to `clone()` (raise the refcount) if we construct `DualHandle` before `Self`.

wgpu-hal/src/dx12/descriptor.rs
```diff
@@ -250,30 +254,33 @@ unsafe impl Sync for CpuHeap {}
...
+            NodeMask: 0,
+        };
+        let raw = unsafe { device.CreateDescriptorHeap::<Direct3D12::ID3D12DescriptorHeap>(&desc) }
             .into_device_result("CPU descriptor heap creation")?;
 
-        null_comptr_check(&raw)?;
-
         Ok(Self {
             inner: Mutex::new(CpuHeapInner {
                 _raw: raw.clone(),
```
MarijnS95, line 273: `clone()` also not needed.

wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -16,57 +16,62 @@ pub(super) fn compile_fxc(
...
-    let mut error = d3d12::Blob::null();
+    let mut error = None;
     let hr = unsafe {
-        profiling::scope!("d3dcompiler::D3DCompile");
-        d3dcompiler::D3DCompile(
+        profiling::scope!("Direct3D::Fxc::D3DCompile");
+        Direct3D::Fxc::D3DCompile(
+            // TODO: Update low-level bindings to accept a slice here
             source.as_ptr().cast(),
             source.len(),
```
MarijnS95, line 47: Should report this upstream at https://github.com/microsoft/win32metadata.

wgpu-hal/src/dx12/command.rs
```diff
@@ -782,50 +792,54 @@ impl crate::CommandEncoder for super::CommandEncoder {
...
-                        pResource: resolve.dst.0.as_mut_ptr(),
-                        Subresource: resolve.dst.1,
-                        StateBefore: d3d12_ty::D3D12_RESOURCE_STATE_RENDER_TARGET,
-                        StateAfter: d3d12_ty::D3D12_RESOURCE_STATE_RESOLVE_DEST,
-                    }
+                let barrier = Direct3D12::D3D12_RESOURCE_BARRIER {
+                    Type: Direct3D12::D3D12_RESOURCE_BARRIER_TYPE_TRANSITION,
+                    Flags: Direct3D12::D3D12_RESOURCE_BARRIER_FLAG_NONE,
+                    Anonymous: Direct3D12::D3D12_RESOURCE_BARRIER_0 {
+                        //Note: this assumes `D3D12_RESOURCE_STATE_RENDER_TARGET`.
```
MarijnS95, line 816:
```suggestion
                        // Note: this assumes `D3D12_RESOURCE_STATE_RENDER_TARGET`.
```

And above.

wgpu-hal/src/dx12/command.rs
```diff
@@ -49,6 +53,7 @@ impl crate::BufferTextureCopy {
 
 impl super::Temp {
     fn prepare_marker(&mut self, marker: &str) -> (&[u16], u32) {
+        // TODO: Store in HSTRING
         self.marker.clear();
         self.marker.extend(marker.encode_utf16());
         self.marker.push(0);
```
MarijnS95, line 59: :eyes:

wgpu-hal/src/dx12/command.rs
```diff
@@ -290,22 +300,21 @@ impl crate::CommandEncoder for super::CommandEncoder {
...
-            .into_device_result("GraphicsCommandList::close")?;
+        unsafe { raw.Close() }.into_device_result("GraphicsCommandList::close")?;
         Ok(super::CommandBuffer { raw })
     }
     unsafe fn reset_all<I: Iterator<Item = super::CommandBuffer>>(&mut self, command_buffers: I) {
         for cmd_buf in command_buffers {
             self.free_lists.push(cmd_buf.raw);
         }
-        self.allocator.reset();
+        let _todo_handle_error = unsafe { self.allocator.Reset() };
```
MarijnS95, line 317: :eyes:

wgpu-hal/src/dx12/command.rs
```diff
@@ -719,7 +724,8 @@ impl crate::CommandEncoder for super::CommandEncoder {
                         cat.clear_value.b as f32,
                         cat.clear_value.a as f32,
                     ];
-                    list.clear_render_target_view(*rtv, value, &[]);
+                    // TODO: Empty slice vs None?
+                    unsafe { list.ClearRenderTargetView(*rtv, &value, Some(&[])) };
```
MarijnS95, line 728: :eyes:

wgpu-hal/src/dx12/mod.rs
```diff
@@ -44,17 +44,259 @@ mod suballocation;
...
-    Interface as _,
+use windows::{
+    core::{Interface, Param as _},
+    Win32::{
+        Foundation,
+        Graphics::{Direct3D, Direct3D12, DirectComposition, Dxgi},
+        System::Threading,
+    },
+};
+use windows_core::Free;
```
MarijnS95, line 59: Merge.

MarijnS95 (commented):
I didn't actually finish self-reviewing my PR. Here are some notes taken while very lightly scrubbing the changes, that I'll perhaps PR changes for later (or someone else can look into them).

wgpu-hal/src/dx12/command.rs
teoxoy: I feel like `None` would be more obvious.

wgpu-hal/src/auxil/dxgi/result.rs
teoxoy: I'm planning to resolve this one in an upcoming PR.

wgpu-hal/src/auxil/dxgi/result.rs
teoxoy: (same as above)

wgpu-hal/src/dx12/command.rs
teoxoy: I'm planning to resolve this one in the upcoming PR as well.

teoxoy:
@MarijnS95 thanks for being so thorough!

teoxoy:
I left some 👍 on the things I think would be good to address in a follow-up; if you'd be up for that :)

wgpu-hal/src/dx12/command.rs
MarijnS95: Looks like the old `clear_render_target_view()` in `d3d12/src/command_list.rs` used to map an empty slice to `ptr::null()`, so I'm changing it to `None` too. If anything this is one of the few strange _semantic differences_ to perhaps have an effect on https://github.com/gfx-rs/wgpu/issues/6162?

wgpu-hal/src/dx12/command.rs
```diff
@@ -732,46 +738,50 @@ impl crate::CommandEncoder for super::CommandEncoder {
...
-                );
+            if let Some(ds_view) = ds_view {
+                if flags != Direct3D12::D3D12_CLEAR_FLAGS::default() {
+                    unsafe {
+                        list.ClearDepthStencilView(
+                            ds_view,
+                            flags,
+                            ds.clear_value.0,
+                            ds.clear_value.1 as u8,
+                            &[],
```
MarijnS95, line 762: This one used to be `ptr::null()` too, but there's no `None` mapping in Direct3D12 here because someone only reported this issue (and hence fixed it) for `clearRenderTargetView()` 🤦🤦.

https://github.com/microsoft/win32metadata/pull/1971

MarijnS95:
Thanks @teoxoy: I've prepared a branch locally to batch some of these together, and will open a PR soon.

wgpu-hal/src/dx12/shader_compilation.rs
MarijnS95: https://github.com/microsoft/windows-rs/issues/3229. The annotation is there, `windows-rs` isn't using it (for possibly good reasons).

wgpu-hal/src/auxil/dxgi/result.rs
MarijnS95: The built-in formatting for `Error::from_hresult(E_INVALIDARG)` is:

- `Display`: `The parameter is incorrect. (0x80070057)`
- `Debug`: `Error { code: HRESULT(0x80070057), message: "The parameter is incorrect." }`

So the former is already better than the hardcoded `invalid argument` string - and we cover more than just 4 errors.

I don't know why I didn't solve this one straight away :)

MarijnS95: I've killed `into_result()` entirely because the only uses either:

- Discard the message (which is weird anyway?)
- Print it in a `log::warn!()`, which could `Display`-format the `windows::core::Error` directly anyway.

PR #5970, opened 2024-07-17
[core] Make `poll(Wait)` not hang after bad command submission.

### LINKED ISSUE #5969: `Device::poll(Wait)` hangs after failed command submission
The test below hangs on Vulkan backends, waiting for a submission index of `1` No commands have been successfully submitted to the queue, so this fence value will never be signaled.

Speculation: We assign submission indexes before we know whether the submission will succeed, because we need to (or used to need to) mark each wgpu_core resource with its submission index before hal submission. We implement `Maintain::Wait` by getting the current fence value and then waiting for it to be signalled. I speculate that, when submission fails, we increment the fence value, but then never submit any commands that would signal that value, so a subsequent `Wait` will wait for a value that cannot be signalled unless some other thread submits work to that queue.

```rust
/// Submit a command buffer to the wrong device. A wait poll shouldn't hang.
#[gpu_test]
static WAIT_AFTER_BAD_SUBMISSION: GpuTestConfiguration =
    GpuTestConfiguration::new().run_async(wait_after_bad_submission);

async fn wait_after_bad_submission(ctx: TestingContext) {
    let (device2, queue2) =
        wgpu_test::initialize_device(&ctx.adapter, ctx.device_features, ctx.device_limits.clone())
            .await;

    let command_buffer1 = ctx.device.create_command_encoder(&<_>::default()).finish();
    // This should fail, since the command buffer belongs to the wrong
    // device.
    wgpu_test::fail(
        &device2,
        || {
            queue2.submit([command_buffer1]);
        },
        Some("doesn't match Device"),
    );

    // This should not hang.
    //
    // Specifically, the failed submission should not cause a new fence value to
    // be allocated that will not be signalled until further work is
    // successfully submitted, causing a greater fence value to be signalled.
    device2.poll(wgpu::Maintain::Wait);
}
```

jimblandy:
Add `wgpu_core::device::Device::last_successful_submission_index`,
which records the fence value that `Maintain::Wait` should actually
wait for. See comments for details.

Fixes #5969.

### DIFF
--- a/tests/tests/poll.rs
+++ b/tests/tests/poll.rs
@@ -125,3 +125,37 @@ static WAIT_OUT_OF_ORDER: GpuTestConfiguration =
             .await
             .panic_on_timeout();
     });
+
+/// Submit a command buffer to the wrong device. A wait poll shouldn't hang.
+///
+/// We can't catch panics on Wasm, since they get reported directly to the
+/// console.
+#[gpu_test]
+static WAIT_AFTER_BAD_SUBMISSION: GpuTestConfiguration = GpuTestConfiguration::new()
+    .parameters(wgpu_test::TestParameters::default().skip(wgpu_test::FailureCase::webgl2()))
+    .run_async(wait_after_bad_submission);
+
+async fn wait_after_bad_submission(ctx: TestingContext) {
+    let (device2, queue2) =
+        wgpu_test::initialize_device(&ctx.adapter, ctx.device_features, ctx.device_limits.clone())
+            .await;
+
+    let command_buffer1 = ctx
+        .device
+        .create_command_encoder(&CommandEncoderDescriptor::default())
+        .finish();
+
+    // This should panic, since the command buffer belongs to the wrong
+    // device, and queue submission errors seem to be fatal errors?
+    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+        queue2.submit([command_buffer1]);
+    }));
+    assert!(result.is_err());
+
+    // This should not hang.
+    //
+    // Specifically, the failed submission should not cause a new fence value to
+    // be allocated that will not be signalled until further work is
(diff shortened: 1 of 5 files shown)


tests/tests/poll.rs
```diff
@@ -125,3 +125,30 @@ static WAIT_OUT_OF_ORDER: GpuTestConfiguration =
...
+#[gpu_test]
+static WAIT_AFTER_BAD_SUBMISSION: GpuTestConfiguration =
+    GpuTestConfiguration::new().run_async(wait_after_bad_submission);
+
+async fn wait_after_bad_submission(ctx: TestingContext) {
+    let (device2, queue2) =
+        wgpu_test::initialize_device(&ctx.adapter, ctx.device_features, ctx.device_limits.clone())
+            .await;
+
+    let command_buffer1 = ctx.device.create_command_encoder(&<_>::default()).finish();
```
cwfitzgerald: Could we spell this type out

wgpu-core/src/device/queue.rs
```diff
@@ -1062,7 +1062,7 @@ impl Global {
             let fence = fence_guard.as_mut().unwrap();
             let submit_index = device
                 .active_submission_index
-                .fetch_add(1, Ordering::Relaxed)
+                .fetch_add(1, Ordering::SeqCst)
```
cwfitzgerald, line 1072: I still don't think this is any more correct, but not with thinking about the cost, so we can leave the change

cwfitzgerald (approved):
Nice cleanups in here, small nits

wgpu-core/src/device/queue.rs
jimblandy: You're right, we need to make sure that submissions occur in order of increasing index, which means we need a lock on this entire thing. Using `SeqCst` here isn't good enough.

jimblandy: Filed as #5978

PR #5976, opened 2024-07-17
Remove latest_submission_index

cwfitzgerald:
**Description**

This PR flips how latest_submission_index is computed. Previously every submit told each resource that it was used. Now we run through the list of active submissions to see if any of them contain the buffer.

The act of incrementing the value for all resources was quite expensive for the few times we actually needed to know it. 

**Testing**

Tests pass, wrote comments explaining my code

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/device/life.rs
+++ b/wgpu-core/src/device/life.rs
@@ -11,7 +11,7 @@ use crate::{
 };
 use smallvec::SmallVec;
 
-use std::{mem, sync::Arc};
+use std::sync::Arc;
 use thiserror::Error;
 
 /// A command submitted to the GPU for execution.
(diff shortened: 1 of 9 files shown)


nical (approved):

wgpu-core/src/device/life.rs
```diff
@@ -248,21 +332,29 @@ impl<A: HalApi> LifetimeTracker<A> {
...
-                self.active.iter().position(|a| a.index == submit_index)
+                self.active
+                    .iter()
+                    .position(|a| Some(a.index) == submit_index)
             );
 
             self.active
                 .iter_mut()
-                .find(|a| a.index == submit_index)
+                .find(|a| Some(a.index) == submit_index)
```
teoxoy:
```suggestion
                .rev()
                .find(|a| a.contains_buffer(&buffer))
```

Since we already iterate through the submissions I think we can do this instead. We don't need to `mem::take` then either.

teoxoy (approved):
Good stuff! It's not as expensive as I thought either👍

teoxoy:
@cwfitzgerald should we merge this now to get it in the release? Even more perf improvements for `.submit()` 👀

Issue #5925, opened 2024-07-08
Switch the Vulkan backend from `gpu_alloc` to `gpu_allocator`.


nical:
It's what we use in the D3D12 backend.

Ideally, implement https://github.com/Traverse-Research/gpu-allocator/issues/235 before doing the switch to avoid changing the behavior too much.

Elabajaba:
I've started working on this.

I've got a super stale branch from when I did this a few years ago, but I'll just redo it from scratch.

PR #5982, opened 2024-07-18
Update gpu-allocator dependency to 0.27

nical:


ErichDonGubler:
Are there any interesting changes that are supposed to accompany this update? 👀

nical:
I added an API to provide allocation reports to gpu-allocator, but since we only use gpu-allocator on for d3d12 (see #5925) and most wgpu users are on vulkan and metal, there is no rush to merge this before the wgpu release that's coming soon.

That said it seems to simplify the dependency tree a bit so we might want to take it anyway.

In a followup patch I'll expose the allocation reports and that's changelog-worthy, but other than that I don't think there is much to report.

Wumpf:
having fewer versions of windows is a win, on full rebuilds I spend a lot of time building `windows`

Wumpf (approved):

Wumpf:
gpu-allocator 0.27 changelog https://github.com/Traverse-Research/gpu-allocator/releases/tag/0.27.0

PR #5992, opened 2024-07-19
Make some errors serializable

sagudev:
**Description**
In servo we need errors to be serializable.

Reviewable per commits (each commit builds by itself)

**Testing**
It compiles.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/instance.rs
+++ b/wgpu-core/src/instance.rs
@@ -23,6 +23,7 @@ type HalInstance<A> = <A as hal::Api>::Instance;
 type HalSurface<A> = <A as hal::Api>::Surface;
 
 #[derive(Clone, Debug, Error)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[error("Limit '{name}' value {requested} is better than allowed {allowed}")]
 pub struct FailedLimit {
     name: &'static str,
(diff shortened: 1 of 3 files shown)


wgpu-core/src/instance.rs
```diff
@@ -391,6 +392,8 @@ pub enum GetSurfaceSupportError {
 }
 
 #[derive(Clone, Debug, Error)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(bound(deserialize = "'de: 'static")))]
```
ErichDonGubler, line 396: question: These `bound` attributes seem unnecessary. Does [`serde::de::DeserializeOwned`](https://docs.rs/serde/1.0.204/serde/de/trait.DeserializeOwned.html) not resolve this problem?

ErichDonGubler:
I don't see any blocking issues with this. I want to tag in @cwfitzgerald for thoughts on supporting this as a temperature check, but I don't see any issues with the implementation beyond what I've submitted comments for.

ErichDonGubler (approved):

cwfitzgerald:
No problems with this landing, the only thing I'm unsure about is that these serializes have no in-codebase justification, so could be broken at some point. I don't know what to do about that problem, just noting it.

wgpu-core/src/instance.rs
sagudev: There is no way to derive `DeserializeOwned` (I think it only serves as a bound for consumers not implementers) or am I missing something?

sagudev:
> No problems with this landing, the only thing I'm unsure about is that these serializes have no in-codebase justification, so could be broken at some point. I don't know what to do about that problem, just noting it.

We could add test (or const asserts) that some types implements (de)serialize (like it's done for Sync/Send), but I think that would make things complicated. (de)serialize can be trivially added again if needed by consumers (mostly browsers).

cwfitzgerald (approved):
Lezdoit

Issue #5989, opened 2024-07-19
Cannot create GL ES 3.x adapter when GL < 3.3 exists


Rapdorian:
**Description**
I have a machine that supports GL 3.1 and GL ES 3.1  
When looking for adapters WGPU prefers using a GL adapter over a GL ES adapter, regardless of GL version. see #5481 
However in my case this causes an issue because my GL version is unsupported but I could use a GL ES adapter, which isn't found.

**Repro steps**
- On machine with GL <3.3 and GL ES 3.1+
- Create a GL adapter

**Expected vs observed behavior**
Observered: WGPU fails to create an adapter since it only supports GL 3.3+  
Expected: WGPU should create a GL ES 3.1 adapter.

**Extra materials**
```
$ glxinfo
name of display: :0
display: :0  screen: 0
direct rendering: Yes
Extended renderer info (GLX_MESA_query_renderer):
    Vendor: Mesa (0xffffffff)
    Device: Mali-G52 r1 (Panfrost) (0xffffffff)
    Version: 24.1.3
    Accelerated: yes
    Video memory: 7687MB
    Unified memory: yes
    Preferred profile: core (0x1)
    Max core profile version: 3.1
    Max compat profile version: 3.1
    Max GLES1 profile version: 1.1
    Max GLES[23] profile version: 3.1
OpenGL vendor string: Mesa
OpenGL renderer string: Mali-G52 r1 (Panfrost)
OpenGL core profile version string: 3.1 Mesa 24.1.3-arch1.1
OpenGL core profile shading language version string: 1.40
OpenGL core profile context flags: (none)

OpenGL version string: 3.1 Mesa 24.1.3-arch1.1
OpenGL shading language version string: 1.40
OpenGL context flags: (none)

OpenGL ES profile version string: OpenGL ES 3.1 Mesa 24.1.3-arch1.1
OpenGL ES profile shading language version string: OpenGL ES GLSL ES 3.10
```

```
$ cargo run --bin wgpu-info
[2024-07-19T04:32:09Z ERROR wgpu_hal::vulkan::instance] GENERAL [Loader Message (0x0)]
    	vkCreateInstance: Found no drivers!
[2024-07-19T04:32:09Z ERROR wgpu_hal::vulkan::instance] 	objects: (type: INSTANCE, hndl: 0xaaaafa787530, name: ?)
[...]

PR #5996, opened 2024-07-19
Check Opengl version is 3.3+ before creating a GL context over a GL ES context

Rapdorian:
**Connections**
#5989 

**Description**
When creating a GL context wgpu prefers GL over GL ES if GL is available, regardless of version. If GL ES 3.1 is available and GL 3.3 is not (but an older GL is) wgpu is not able to create the GL ES context

If GL is supported I am creating a context of any version to check the GL version supported, if less than GL 3.3 we try to build a GL ES context instead of a GL context.

**Testing**

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-hal/src/gles/egl.rs
+++ b/wgpu-hal/src/gles/egl.rs
@@ -550,26 +550,25 @@ impl Inner {
         let supports_khr_context = display_extensions.contains("EGL_KHR_create_context");
 
         let mut context_attributes = vec![];
-        if supports_opengl {
-            context_attributes.push(khronos_egl::CONTEXT_MAJOR_VERSION);
-            context_attributes.push(3);
-            context_attributes.push(khronos_egl::CONTEXT_MINOR_VERSION);
-            context_attributes.push(3);
-            if force_gles_minor_version != wgt::Gles3MinorVersion::Automatic {
-                log::warn!("Ignoring specified GLES minor version as OpenGL is used");
-            }
-        } else {
-            context_attributes.push(khronos_egl::CONTEXT_MAJOR_VERSION);
-            context_attributes.push(3); // Request GLES 3.0 or higher
-            if force_gles_minor_version != wgt::Gles3MinorVersion::Automatic {
-                context_attributes.push(khronos_egl::CONTEXT_MINOR_VERSION);
-                context_attributes.push(match force_gles_minor_version {
-                    wgt::Gles3MinorVersion::Automatic => unreachable!(),
-                    wgt::Gles3MinorVersion::Version0 => 0,
-                    wgt::Gles3MinorVersion::Version1 => 1,
-                    wgt::Gles3MinorVersion::Version2 => 2,
-                });
-            }
+        let mut gl_context_attributes = vec![];
+        let mut gles_context_attributes = vec![];
+        gl_context_attributes.push(khronos_egl::CONTEXT_MAJOR_VERSION);
+        gl_context_attributes.push(3);
+        gl_context_attributes.push(khronos_egl::CONTEXT_MINOR_VERSION);
+        gl_context_attributes.push(3);
+        if supports_opengl && force_gles_minor_version != wgt::Gles3MinorVersion::Automatic {
+            log::warn!("Ignoring specified GLES minor version as OpenGL is used");
+        }
+        gles_context_attributes.push(khronos_egl::CONTEXT_MAJOR_VERSION);
+        gles_context_attributes.push(3); // Request GLES 3.0 or higher
+        if force_gles_minor_version != wgt::Gles3MinorVersion::Automatic {
+            gles_context_attributes.push(khronos_egl::CONTEXT_MINOR_VERSION);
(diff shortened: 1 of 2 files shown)


wgpu-hal/src/gles/egl.rs
```diff
@@ -532,12 +532,42 @@ impl Inner {
                 .query_string(Some(display), khronos_egl::CLIENT_APIS)
                 .unwrap()
                 .to_string_lossy();
-            client_apis
+            if client_apis
                 .split(' ')
                 .any(|client_api| client_api == "OpenGL")
+            {
+                // check if GL version is supported
+                egl.bind_api(khronos_egl::OPENGL_API).unwrap();
```
Wumpf: this goes a bit beyond this PR since technically both the unwrap before and after are affected, but these unwraps make me paranoid! I think we'd better off handling each of them gracefully (more not-supported situations essentially)

wgpu-hal/src/gles/egl.rs
```diff
@@ -532,12 +532,42 @@ impl Inner {
...
+                let ctx = egl
+                    .create_context(display, config, None, &[khronos_egl::NONE])
+                    .map_err(|e| {
+                        crate::InstanceError::with_source(
+                            String::from("unable to create GL context"),
+                            e,
+                        )
+                    })?;
```
Wumpf: could it in theory happen that we fail creating the context while OPENGL_API is bound but wouldn't fail it if OPENGL_ES_API was bound?

wgpu-hal/src/gles/egl.rs
```diff
@@ -532,12 +532,42 @@ impl Inner {
...
+                let _ = egl.make_current(display, None, None, Some(ctx));
+                unsafe {
```
Wumpf: that unsafe scope is larger than it has to be, please shrink it down to the bare minimum

Wumpf (changes requested):
Beyond a few nits looking good!
Also, please add a changelog entry, this is definitely a change users want to be able to read up upon :)

Maybe @valaphee who did a lot of the non-es OpenGL support previously wants to have a look as well?

wgpu-hal/src/gles/egl.rs
Rapdorian: I've cut down the unsafe scope

valaphee:
Personally I would handle the error rather than of preventing it. (would also remove unsafe and unwrap, and `from_loader_function` is quite heavy, as it gathers all function pointers)

e.g. https://github.com/gfx-rs/wgpu/blob/trunk/wgpu-hal/src/gles/egl.rs#L609 fails because of https://github.com/gfx-rs/wgpu/blob/trunk/wgpu-hal/src/gles/egl.rs#L554

and if it fails and OpenGL is used, then I would bind the OPENGL_ES_API and try again as this is the unlikely-path anyway.

Rapdorian:
I've reworked the changes wrt @valaphee's recommendation.
The code now lets the GL context creation fails and then retrys with GLES.

wgpu-hal/src/gles/egl.rs
```diff
@@ -606,7 +605,29 @@ impl Inner {
             context_attributes.push(khr_context_flags);
         }
         context_attributes.push(khronos_egl::NONE);
-        let context = match egl.create_context(display, config, None, &context_attributes) {
+
+        gl_context_attributes.extend(&context_attributes);
+        gles_context_attributes.extend(&context_attributes);
+
+        let context = match egl
```
Wumpf: I had a pretty hard time parsing this with all the functionality crammed inside the head of a match statement. Also the error message down there isn't entirely correct anymore. How about

```rust
        let context = if supports_opengl {
            if let Ok(context) = egl.create_context(display, config, None, &gl_context_attributes) {
                context
            } else {
                // We tried OpenGL earlier, but failed. Try OpenGL ES instead.
                egl.bind_api(khronos_egl::OPENGL_ES_API).unwrap();
                egl.create_context(display, config, None, &gles_context_attributes)
                    .map_err(|err| {
                        crate::InstanceError::with_source(
                            String::from("unable to create either OpenGL and GLES3.x context"),
                            err,
                        )
                    })?
            }
        } else {
            egl.create_context(display, config, None, &gles_context_attributes)
                .map_err(|err| {
                    crate::InstanceError::with_source(
                        String::from("unable to create GLES 3.x context"),
                        err,
                    )
                })?
        };
```

Wumpf (changes requested):
ah this makes sense. Thank you @valaphee for pointing this out and @Rapdorian for improving accordingly!
Style suggestion for readability (anything but 20 line match head plz 😄 ), otherwise I think we're good

wgpu-hal/src/gles/egl.rs
Rapdorian: I've refactored it a bit to be a bit clearer.

Wumpf: ah neat, even better than what I suggested!

Wumpf (approved):
thanks again!

PR #5998, opened 2024-07-20
Split `wgpu` Crate into Modules

cwfitzgerald:
This PR is a bit of a flashbang, but this is the kind of transformation that needs to be done quickly.

I have re-organized wgpu's `lib.rs` into a series of modules in `api/*` where each major type in our API has a separate module.

Importantly I have not modified any of the code (outside of imports, and adding a `Id::new`) just re-arranged their locations. I think this is monumentally more clear than before.

For descriptor structs, they are defined close to the thing they crate. Helper structs like `MapContext` are next to the things that use them.

Unless anyone has major qualms with this, I suggest merging quickly, and iterating once it's in. This is going to bitrot extremely quickly if left.

**Checklist**

- [x] Run `cargo fmt`.


wgpu/src/api/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+//!
+//! # Conventions
+//!
+//! Each major type gets its own module. The module is laid out as follows:
+//!
+//! - The type itself
+//! - `impl` block for the type
+//! - `Drop` implementation for the type (if needed)
+//! - Descriptor types and their subtypes.
+//! - Any non-public helper types or functions.
```
Wumpf, line 12: very appreciated that you spell the convention out here :)

Wumpf (approved):
Putting a great amount of trust in that the moved code hasn't changed.

lib.rs & mod.rs looking great. Structure makes a lot of sense.

SOOO much better!

PR #5999, opened 2024-07-20
Convert some module doc comments

waywardmonkeys:
**Description**
Some module doc comments were using `/*! ... */` syntax and had leading ` *` prefixes on each line. This interferes with the tracking of `clippy::doc_lazy_continuation`, so switch those over to `//!` style comment blocks.

This leaves `/*! ... */` blocks alone which didn't prefix each line.

**Testing**

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/player/src/bin/play.rs
+++ b/player/src/bin/play.rs
@@ -1,5 +1,4 @@
-/*! This is a player for WebGPU traces.
-!*/
+//! This is a player for WebGPU traces.
 
 #[cfg(not(target_arch = "wasm32"))]
 fn main() {
--- a/player/src/lib.rs
+++ b/player/src/lib.rs
@@ -1,6 +1,5 @@
-/*! This is a player library for WebGPU traces.
- *
-!*/
+//! This is a player library for WebGPU traces.
+
 #![cfg(not(target_arch = "wasm32"))]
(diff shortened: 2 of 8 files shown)


cwfitzgerald:
Going to specifically cc @jimblandy on this one - I don't see why we shouldn't accept this though.

waywardmonkeys:
A couple of these blocks will have warnings in Rust 1.80 which is what led me down this path.

nical (approved):

ErichDonGubler (approved):

ErichDonGubler:
Going to specifically merge (🙃), and deal with any fallout that @jimblandy would like to impose. I think it's valuable to adhere to the convention that Rust upstream imposes by default, be internally consistent in the format of documentation that we have for ourselves, _and_ avoid warnings from `rustdoc`.

waywardmonkeys:
@ErichDonGubler Thanks. There are still a number of files that use a similar syntax, but they don't indent and put a ` * ` at the start of each line.

```
git grep "\/\*\!"
```

I targeted these as some of them introduced lint failures in 1.80 with `clippy::doc_lazy_continuation`.

PR #6002, opened 2024-07-20
wgpu-hal dynamic dispatch, towards fewer generics in wgpu-core

Wumpf:
**Connections**
* Towards (or the whole story? TBD!) of #5124 

**Description**

Still heavy work in progress!

---

This latest attempt to degenerification of wgpu core tries to be as straight forward as possible:
* introduce dynamic base types that cast to concrete types in HAL (`DynDevice`, `DynCommandEncoder` etc.)
* wherever wgpu-core passes around `&A::CommandEncoder` (and similar) instead pass around `& dyn DynCommandEncoder`
* start using a lot of `Box<&dyn DynBuffer>` (and similar)
* drop `<A: hal::API>` from all the wgpu-core types
* profit!

Previous more complex attempts tried to avoid boxing as much as possible, I'm deliberately doing here the opposite, well aware that this may cost us! But nobody actually measured how much yet and neither have I, so the jury is still out on this.

The hope is that this approach is simple enough that we can pull it off and cheap enough that we can worry about optimizing things later (it's not too hard to come up with ideas on how to tackle ensuing costs, we just have to know what to optimize first!).

---

I'm keeping this in a fairly landable state at every step (minus Metal & web things which I'm not testing all the time) and try to keep steps granular, BUT....
* there's still quite a bit of experimentation going on where I have to go back and change some things
     * I may or may not clean up the commit history to eliminate a few things where I went in circles.
* this overall approach has the risk of degrading performance more than we're willing to pay (we _are_ willing to pay a certain amount, since status quo is not great for all the reasons described in #5124), but it's hard to tell if it's bad or not without a full vertical slice
   * I may end up getting this almost to completion and then re-open a series of smaller, more reviewable PRs to land it

**Testing**
Hopefully covered by existing tests ;-)

**Checklist**

- [ ] Run `cargo fmt`.


Wumpf:
closing in favor of 3 part pr series coming up soon!

PR #6004, opened 2024-07-20
Allow 3D textures to be renderable

matthew-wong1:
**Description**
Hi! The WebGPU spec now allows textures with a 3D dimension to be renderable but the current wgpu implementation only allows for 2D textures to be renderable. I updated the checks to not throw an error if a 3D texture has GPUTextureUsage.RENDER_ATTACHMENT. I couldn't find any open issues or pull requests related to this. Sorry if it might be a work-in-progress or if it was a conscious decision not to enable 3D textures to be renderable yet!

I think the they updated the spec for 3D textures to be renderable around November 2023. 

**Testing**
I ran `cargo xtask test` on an M2 Mac. All tests passed except 3 (the same 3 also failed on the trunk branch of wgpu). 

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ Bottom level categories:
 - Fix profiling with `tracy`. By @waywardmonkeys in [#5988](https://github.com/gfx-rs/wgpu/pull/5988)
 - As a workaround for [issue #4905](https://github.com/gfx-rs/wgpu/issues/4905), `wgpu-core` is undocumented unless `--cfg wgpu_core_doc` feature is enabled. By @kpreid in [#5987](https://github.com/gfx-rs/wgpu/pull/5987)
 - Bump MSRV for `d3d12`/`naga`/`wgpu-core`/`wgpu-hal`/`wgpu-types`' to 1.76. By @wumpf in [#6003](https://github.com/gfx-rs/wgpu/pull/6003)
+- Allow for 3D textures to be rendered. By @matthew-wong1 in [#6004](https://github.com/gfx-rs/wgpu/pull/6004)
 
 ## 22.0.0 (2024-07-17)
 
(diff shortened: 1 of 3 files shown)


Wumpf (changes requested):
the change itself is looking good but...
* I think this should be covered by a unit test since there's no way of being sure that our backends actually support this (if any doesn't, we have to emulate this feature by doing copies etc.). I had a quick browse just now and it unforunately doesn't look like there's anything checking on something like that yet.
     * a test creating a renderpass to clear a specific layer of a small volume texture with a "marker color" and then reading back that only that layer has been cleared should be enough
     * bonus points if that test also tests array textures :)
* changelog entry please!

matthew-wong1:
Thanks for the feedback! Sorry for overlooking the changelog. 

I've been working on writing the test but I've been a bit slow as I'm still a bit new to Rust and using WebGPU through rust. My understanding is to create a texture view of a particular layer and then pass that into a render pass with a Clear operation, then copy the texture to a buffer and see if that layer is of the color that it was cleared to. 

Also realized I'll probably need to update a couple more places in the code as I've been running into unreachable errors related to 3D textures.

cwfitzgerald:
> Sorry for overlooking the changelog.

We're pretty bad about it ourselves, so you're in good company 😅

Feel free to come on the matrix to ask any questions if you need help with implementing tbf tests.

matthew-wong1:
It seems like the current backend doesn't fully support rendering 3D textures just yet. I probably should've looked a bit closer initially, but based on the WebGPU spec (https://www.w3.org/TR/webgpu/#dom-gpucommandencoder-beginrenderpass) beginning a renderpass with a 3D texture requires a new attribute `colorAttachment.depthSlice` to be included in the renderPassDescriptor. For me, implementing that is a bit out of my depth but I'm happy to help with writing any unit tests along the way. Sorry!

Wumpf:
Then it's probably best to shelve this for now.
Would you mind creating an issue detailing what needs to be done and what you've learned on the way so far?

PR #6006, opened 2024-07-21
[wgpu-hal] gles/wgl: Migrate from ancient/unmaintained `winapi` to `windows-rs`

MarijnS95:
**Connections**
Contributes to #3207
Same as PR #5956 but for GLES/WGL, instead of the Dx12 backend.

**Description**
Replaces ancient, deprecated, unmaintained `winapi` crate with `windows-rs`.  The full justification is in the linked issue and PR above.

**Testing**

```
WGPU_BACKEND=gles cargo run --bin wgpu-examples shadow
```

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 Please add your PR to the changelog! Choose from a top level and bottom
 level category, then write your changes like follows:
 
-- Describe your change in a user friendly format by @yourslug in [#99999](https://github.com/gfx-rs/wgpu/pull/99999)
+- Describe your change in a user friendly format. By @yourslug in [#99999](https://github.com/gfx-rs/wgpu/pull/99999)
 
 You can add additional user facing information if it's a major breaking change. You can use the following to help:
 
@@ -55,6 +55,12 @@ Bottom level categories:
 - Print requested and supported usages on `UnsupportedUsage` error. By @VladasZ in [#6007](https://github.com/gfx-rs/wgpu/pull/6007)
 - Fix function for checking bind compatibility to error instead of panic. By @sagudev [#6012](https://github.com/gfx-rs/wgpu/pull/6012)
 
+### Dependency Updates
+
+#### GLES
+
+- Replace `winapi` code in WGL wrapper to use the `windows` crate. By @MarijnS95 in [#6006](https://github.com/gfx-rs/wgpu/pull/6006)
+
 ## 22.0.0 (2024-07-17)
 
 ### Overview
@@ -64,7 +70,7 @@ Bottom level categories:
 For the first time ever, WGPU is being released with a major version (i.e., 22.* instead of 0.22.*)! Maintainership has decided to fully adhere to [Semantic Versioning](https://semver.org/)'s recommendations for versioning production software. According to [SemVer 2.0.0's Q&A about when to use 1.0.0 versions (and beyond)](https://semver.org/spec/v2.0.0.html#how-do-i-know-when-to-release-100):
 
 > ### How do I know when to release 1.0.0?
-> 
+>
 > If your software is being used in production, it should probably already be 1.0.0. If you have a stable API on which users have come to depend, you should be 1.0.0. If you’re worrying a lot about backward compatibility, you should probably already be 1.0.0.
 
 It is a well-known fact that WGPU has been used for applications and platforms already in production for years, at this point. We are often concerned with tracking breaking changes, and affecting these consumers' ability to ship. By releasing our first major version, we publicly acknowledge that this is the case. We encourage other projects in the Rust ecosystem to follow suit.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1522,7 +1522,7 @@ dependencies = [
  "libc",
  "log",
  "rustversion",
- "windows",
+ "windows 0.54.0",
 ]
 
 [[package]]
@@ -1686,7 +1686,7 @@ dependencies = [
  "presser",
  "thiserror",
  "winapi",
- "windows",
+ "windows 0.58.0",
 ]
 
 [[package]]
@@ -2608,7 +2608,7 @@ dependencies = [
  "redox_syscall 0.5.1",
  "smallvec",
  "thread-id",
- "windows-targets 0.52.5",
+ "windows-targets 0.52.6",
 ]
 
 [[package]]
@@ -4346,6 +4346,7 @@ dependencies = [
  "web-sys",
  "wgpu-types",
  "winapi",
+ "windows 0.58.0",
  "winit 0.29.15",
 ]
 
@@ -4473,8 +4474,18 @@ version = "0.54.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9252e5725dbed82865af151df558e754e4a3c2c30818359eb17465f1346a1b49"
 dependencies = [
- "windows-core",
- "windows-targets 0.52.5",
+ "windows-core 0.54.0",
+ "windows-targets 0.52.6",
+]
+
+[[package]]
+name = "windows"
+version = "0.58.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dd04d41d93c4992d421894c18c8b43496aa748dd4c081bac0dc93eb0489272b6"
+dependencies = [
+ "windows-core 0.58.0",
+ "windows-targets 0.52.6",
 ]
 
 [[package]]
@@ -4483,8 +4494,43 @@ version = "0.54.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "12661b9c89351d684a50a8a643ce5f608e20243b9fb84687800163429f161d65"
 dependencies = [
- "windows-result",
- "windows-targets 0.52.5",
+ "windows-result 0.1.2",
+ "windows-targets 0.52.6",
+]
+
+[[package]]
+name = "windows-core"
+version = "0.58.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6ba6d44ec8c2591c134257ce647b7ea6b20335bf6379a27dac5f1641fcf59f99"
+dependencies = [
+ "windows-implement",
+ "windows-interface",
+ "windows-result 0.2.0",
+ "windows-strings",
+ "windows-targets 0.52.6",
+]
+
+[[package]]
+name = "windows-implement"
+version = "0.58.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2bbd5b46c938e506ecbce286b6628a02171d56153ba733b6c741fc627ec9579b"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.72",
+]
+
+[[package]]
+name = "windows-interface"
+version = "0.58.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "053c4c462dc91d3b1504c6fe5a726dd15e216ba718e84a0e46a88fbe5ded3515"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.72",
 ]
 
 [[package]]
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 Please add your PR to the changelog! Choose from a top level and bottom
 level category, then write your changes like follows:
 
-- Describe your change in a user friendly format by @yourslug in [#99999](https://github.com/gfx-rs/wgpu/pull/99999)
+- Describe your change in a user friendly format. By @yourslug in [#99999](https://github.com/gfx-rs/wgpu/pull/99999)
```
Wumpf, line 7: oh, nice catch 😄

wgpu-hal/Cargo.toml
```diff
@@ -144,6 +147,8 @@ khronos-egl = { version = "6", features = ["static", "no-pkg-config"] }
 libloading = { version = ">=0.7, <0.9", optional = true }
 
 [target.'cfg(windows)'.dependencies]
+# backend: Dx12 and Gles
+windows = { version = "0.58", default-features = false, optional = true }
```
Wumpf: since there's more `windows` crate uses coming, let's put that into the workspace `Cargo.toml` (make sure it's also set with default-features false in there, otherwise it opts in default features for everything with now way to opt out)

```suggestion
windows = { workspace = true, optional = true }
```

Wumpf (approved):
Great stuff, thank you so much for taking care of the windows-rs migration everywhere!

Imho good to go after the workspace dependency nit (see comment)

wgpu-hal/Cargo.toml
MarijnS95: As before, I'm still confused by this unused yet _partially_ maintained workspace dependencies table :upside_down_face: 

Why are some but not all dependencies inherited from the workspace?

MarijnS95: Also, do folks want to start with Windows 0.57 or Windows 0.58? The latter has MSRV 1.70 but it seems `wgpu` is already on `1.76`.

Note that `default-features` only turns off the `std` feature: I'm not exactly sure what that implies but we may get worse error formatting - at least it was like that years ago.

Cargo.lock
```diff
@@ -1522,7 +1522,7 @@ dependencies = [
  "libc",
  "log",
  "rustversion",
- "windows",
+ "windows 0.54.0",
```
Wumpf, line 1525: might be worth digging what still uses 0.54 here and whether we could avoid the duplication, but doesn't have to be now

wgpu-hal/Cargo.toml
Wumpf: >  Why are some but not all dependencies inherited from the workspace?

wgpu is older than the ability to do workspace inheritance of crates and nobody every moved stuff over consistently 🤷 

> Also, do folks want to start with Windows 0.57 or Windows 0.58? The latter has MSRV 1.70 but it seems wgpu is already on 1.76.

Any reason in favor of Windows 0.57? As you pointed out from MSRV pov that's not an issue, so we should rather stay up to date? If there's a reason to stick to the older we also have to make this known to dependabot, otherwise an update will creep in latest next week ;-) 

> Note that default-features only turns off the std feature: I'm not exactly sure what that implies but we may get worse error formatting - at least it was like that years ago.

Good to know, maybe worth a comment on it in case you can confirm? Generally good practice for a library like wgpu not to drag in default features when they're not needed even if it's something like improved error handling, after all a user of wgpu can still do that just fine. But the other way round a user of wgpu wouldn't be able to opt-out of those default features at all

Cargo.lock
cwfitzgerald: Flagging this as a blocker for ff - i think we should be using as wide of a range dependency on windows as we can pull off, as having duplicates can range from being unfortunate to a hard blocker (Firefox)

cwfitzgerald (changes requested):
See comment above

Cargo.lock
ErichDonGubler: This is actually easy to answer with `cargo tree`:

```
$ cargo tree -i windows@0.54.0 -i windows@0.58.0 --all-features --all-targets

windows v0.54.0
└── generator v0.8.1
    └── loom v0.7.2
        └── tracy-client v0.17.1
            └── wgpu-benchmark v22.0.0 (E:\wgpu\benches)

windows v0.58.0
├── gpu-allocator v0.27.0
│   └── wgpu-hal v22.0.0 (E:\wgpu\wgpu-hal)
│       ├── wgpu v22.0.0 (E:\wgpu\wgpu)
│       │   ├── wgpu-benchmark v22.0.0 (E:\wgpu\benches)
│       │   ├── wgpu-examples v22.0.0 (E:\wgpu\examples)
│       │   ├── wgpu-info v22.0.0 (E:\wgpu\wgpu-info)
│       │   └── wgpu-test v22.0.0 (E:\wgpu\tests)
│       │       [dev-dependencies]
│       │       └── wgpu-examples v22.0.0 (E:\wgpu\examples)
│       └── wgpu-core v22.0.0 (E:\wgpu\wgpu-core)
│           ├── player v22.0.0 (E:\wgpu\player)
│           └── wgpu v22.0.0 (E:\wgpu\wgpu) (*)
└── wgpu-hal v22.0.0 (E:\wgpu\wgpu-hal) (*)
```

ErichDonGubler: `windows` 0.52 is what's currently in use in `mozilla-central`, and it has explicit use by _many_ crates. I'm not sure how easy this is to upgrade in `mozilla-central`, but I suspect it's non-trivial, _unless_ `windows` 0.52 is source-compatible _enough_ with 0.58.0.

CC @glandium and @jimblandy, who I'm sure would be interested in the pressure we're feeling on this front.

wgpu-hal/Cargo.toml
MarijnS95: > wgpu is older than the ability to do workspace inheritance of crates and nobody every moved stuff over consistently 🤷

It was previously told to me that workspace dependency inheritance was reverted, but the workspace table was kept in place so that it could ever get reenabled.

> Any reason in favor of Windows 0.57? As you pointed out from MSRV pov that's not an issue, so we should rather stay up to date? If there's a reason to stick to the older we also have to make this known to dependabot, otherwise an update will creep in latest next week ;-)

None really; except that I started the conversion on `0.57`. A lot of _my_ high-level API improvements in `Dxgi` via metadata annotations finally landed in `windows 0.58`, too. All dependencies like `gpu-allocator` have released with `windows 0.58` in their range as well.

---

I'll double-check the purpose of the `std` feature before marking this thread as resolved.

Cargo.lock
MarijnS95: > might be worth digging what still uses 0.54

No need to dig, it's pretty easy to see from the `Cargo.lock` context (or `cargo tree` otherwise) that this is [`generator`](https://crates.io/crates/generator).

`cargo tree` however points out that it's a seemingly-irrelevant dependency, exclusively involved in the `wgpu-benchmark` tree and not clobbering `wgpu`/`wgpu-hal`?

As for `0.52`: that really is quite a ways back and very unlikely to be version-range-compatible, especially for `Direct3D12`. It'll definitely work for our purposes, though. Let me know if you want me to downgrade in all PRs.

glandium: `windows-sys` 0.52 is what's used by plenty of crates. `windows` 0.52 is only used by gpu-allocator. Note the versions are *NOT* correlated. 0.52 is the current version for `windows-sys` and `windows` 0.58 depends on it. Anyways, there is some manual work involved in upgrading `windows` in mozilla-central, but nothing prevents it from happening.

MarijnS95: @glandium note that `windows` doesn't depend on `windows-sys`. It depends on `windows-targets`. Which `windows-sys` also depends on.

It is (unfortunately) not a classical `-sys` setup that helps deduplicating low-level definitions by splitting them out of the `windows` crate into `windows-sys` (i.e. having (vtable) function definitions there would be immensely helpful).

glandium: Sure, but that doesn't change the version requirements for mozilla-central.

wgpu-hal/Cargo.toml
cwfitzgerald: For context on the workspace table, we introduced it, but there was a bug in cargo-vendor that caused Firefox to not be able to injest it, so we reverted core, hal, and types (the only crates ff uses) back to standard dependencies. The bug has been long fixed, I just don't think anyone has put in the work to verify it's fine and move back to a workspace dep.

cwfitzgerald:
I suspect this windows version dance is going to be a recurring theme in our future. That is one upside off using a years-abandoned crate, nothing ever changes 😛

Anywho I think we (the maintainers) should figure out what the viable paths forward are that satisfy mozilla-central and wgpu's external users and then clearly communicate that, as I don't want Marijn to do duplicate work and I don't want this to explode mozilla-central.

- Ideally we can figure out a way to land 0.57/58. It's not improbable that nothing else in the stack breaks and mozilla-central can just update, though I know there will be politics around that
- if not, dropping back and supporting as wide of a range as we can, to minimize duplicate dependencies. We should earmark Marjin's improvements, so when et do update, we can re-implement them. 
- loom is indeed irrelevant
- if there is no range beyond 0.52

wgpu-hal/Cargo.toml
waywardmonkeys: I've opened #6020 to start cleaning up some of the workspace dependencies (if that's helpful).

MarijnS95: Thanks @waywardmonkeys, looking forward to see that PR fix at least the dependencies used here, so that this PR can simply be based on top. Looks like you have a lot of dependencies to go, though.

Cargo.lock
MarijnS95: > 0.52 is only used by gpu-allocator.

`gpu-allocator` is a crate that I maintain and recently released a new version for with support all the way up to `windows 0.58`.

The `mozilla-central` repo indeed [uses `windows 0.52`](https://searchfox.org/mozilla-central/rev/de46cd99a7d1634058b511a3f546a970763e5048/Cargo.lock#6915-6916) [for `gpu-allocator`](https://searchfox.org/mozilla-central/rev/de46cd99a7d1634058b511a3f546a970763e5048/Cargo.lock#2562) but it's [only used by `wgpu-hal`](https://searchfox.org/mozilla-central/rev/de46cd99a7d1634058b511a3f546a970763e5048/Cargo.lock#6819) whose version I am bumping **in this PR**, so this should be a **non-issue**.

It seems you [have a semver-compatible patch for `windows-targets`](https://searchfox.org/mozilla-central/rev/de46cd99a7d1634058b511a3f546a970763e5048/Cargo.toml#110-112) anyway.

glandium (approved):
No problem for me to upgrade to windows 0.58.

wgpu-hal/Cargo.toml
MarijnS95: > > Note that default-features only turns off the std feature: I'm not exactly sure what that implies but we may get worse error formatting - at least it was like that years ago.
> 
> Good to know, maybe worth a comment on it in case you can confirm? Generally good practice for a library like wgpu not to drag in default features when they're not needed even if it's something like improved error handling, after all a user of wgpu can still do that just fine. But the other way round a user of wgpu wouldn't be able to opt-out of those default features at all

Upon closer inspection, [the feature is broken](https://github.com/microsoft/windows-rs/issues/3163) ([fix here](https://github.com/microsoft/windows-rs/pull/3164)).

- `windows-core`: Only uses it to turn on `Event`, which is a wrapper for delegates. The feature is **not forwarded to windows-result` nor `windows-strings`** (which is a bug I'd say);
- `windows`: only forwards it to `windows-core`.

The transitive crates:

- `windows-result`: implements `std::error::Error` for `Error`, and conversions to/from `std::io::Error` (which is very similar on Windows);
- `windows-strings`: various conversion for `HSTRING` to and from `OsStr(ing)` and `Path`.

In the end it's nothing that seems to be hiding nicer internals such as error formatting from us. Everything that it exposes is explicit public API (and if we end up relying on it, turning off `std` will cause compiler errors).

Cargo.lock
cwfitzgerald: Heh, nice!

MarijnS95:
@cwfitzgerald since we've made the resolution to just go ahead and use `windows 0.58` directly (it seems), can we get reviews finished? Landing this PR first will make it much easier and cleaner to rebase #5956 on top.

cwfitzgerald (approved):
Great stuff! Got the all clear to merge from the moz team.

cwfitzgerald:
Ready to land, just needs merge conflicts resolved

PR #6007, opened 2024-07-21
Print requested and supported usages on UnsupportedUsage error

VladasZ:
**Description**
`ConfigureSurfaceError::UnsupportedUsage` error log message is not very informative.

**Testing**
This is how it looks with this change:
image

**Checklist**

- [x] Run `cargo fmt`.


cwfitzgerald (approved):
Nice catch!

PR #6010, opened 2024-07-22
Reduce the api_log overhead

nical:
**Description**

Some bevy users report that repetitively checking the global log level atomic adds up to a large overhead in wgpu  even when the log level is disabled (at runtime but still enabled din the build). Since the vast majority of it tends to be in render passes with many commands, it's easy enough to check the global once per render pass and skip the check for each individual command.

I initially thought that adding some sugar in the `api_log` macro to allow it to be conditional would  make a difference but in retrospect it looks like:

```rust
api_log!(if api_log_enabled, "Hello");
```

is not so different from:

```rust
if api_log_enabled {
    api_log!("Hello");
}
```

Let me know if you feel strongly one way or the other.

See also:
 - https://github.com/bevyengine/bevy/issues/14116
 - https://github.com/gfx-rs/wgpu/pull/5804

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.


jimblandy:
In the absence of measurements, we can't take PRs like this. This isn't a final decision - if you're able to post some benchmarks, then by all means we can re-open this.

PR #6015, opened 2024-07-22
[core] Use `ManuallyDrop` for `wgpu_core::device::Queue::raw`.

jimblandy:
Change the field `wgpu_core::device::Queue::raw` from an `Option<A::Queue>` to a `std::mem::ManuallyDrop<A::Queue>`. Replace various `.as_ref().unwrap()` chains with calls to a new accessor function `Queue::raw`.

An `Option` is misleading, as this field is always populated during the lifetime of a `Queue`. Instead, we simply have a field whose value needs to be moved in `<Queue as Drop>::drop`; `ManuallyDrop` is the Rust idiom for this situation.

### DIFF
--- a/wgpu-core/src/device/queue.rs
+++ b/wgpu-core/src/device/queue.rs
@@ -30,7 +30,7 @@ use smallvec::SmallVec;
 
 use std::{
     iter,
-    mem::{self},
+    mem::{self, ManuallyDrop},
     ptr::NonNull,
     sync::{atomic::Ordering, Arc},
 };
@@ -39,8 +39,21 @@ use thiserror::Error;
 use super::Device;
 
 pub struct Queue<A: HalApi> {
-    pub(crate) raw: Option<A::Queue>,
+    raw: ManuallyDrop<A::Queue>,
     pub(crate) device: Arc<Device<A>>,
+}
+
+impl<A: HalApi> Queue<A> {
+    pub(crate) fn new(device: Arc<Device<A>>, raw: A::Queue) -> Self {
+        Queue {
+            raw: ManuallyDrop::new(raw),
+            device,
+        }
+    }
+
+    pub(crate) fn raw(&self) -> &A::Queue {
+        &self.raw
+    }
 }
 
 crate::impl_resource_type!(Queue);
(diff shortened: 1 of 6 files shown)


wgpu-core/src/device/queue.rs
```diff
@@ -39,10 +39,23 @@ use thiserror::Error;
 use super::Device;
 
 pub struct Queue<A: HalApi> {
-    pub(crate) raw: Option<A::Queue>,
+    raw: std::mem::ManuallyDrop<A::Queue>,
```
cwfitzgerald: Could we import the ManuallyDrop

cwfitzgerald (approved):
Good, one nit

wgpu-core/src/device/queue.rs
```diff
@@ -56,7 +69,8 @@ crate::impl_storage_item!(Queue);
 impl<A: HalApi> Drop for Queue<A> {
     fn drop(&mut self) {
         resource_log!("Drop {}", self.error_ident());
-        let queue = self.raw.take().unwrap();
+        // Safety: we never access `self.raw` beyond this point.
```
ErichDonGubler: suggestion: We should use the same case as suggested by Rust upstream for `SAFETY` comments, i.e., uppercase.

```suggestion
        // SAFETY: we never access `self.raw` beyond this point.
```

ErichDonGubler: Pushed up <https://github.com/gfx-rs/wgpu/commit/825400c086fb029f93e08b8b4e0e005759a755d3> to fulfill the above.

ErichDonGubler: suggestion: The same page as ☝🏻 also shows an example of a `# Safety` section in API documentation, which doesn't include a trailing colon in the section name. We do this in some places. I've also pushed <https://github.com/gfx-rs/wgpu/commit/1787279aa3c16c0050dc148da2331586a1f0f865> to address this.

>

ErichDonGubler (approved):
Approving with the piggybacking of some safety-related style changes. CC @cwfitzgerald, who I found in Git history governing those.

jimblandy:
Changed this to rebase, the doc commits seem better as independent changes, since they cover stuff that has nothing to do with the ManuallyDrop logic.

PR #6016, opened 2024-07-22
[wgpu-tests] Use concrete error messages for failures

### LINKED ISSUE #5727: Good first issue: use concrete error messages for `wgpu-test` tests
After merging https://github.com/gfx-rs/wgpu/pull/5290, there are many tests in `tests/tests/` that do not specify the text of the expected validation error. This is actually a hole in coverage; if the reason for validation changes, it could potentially be a regression of coverage. We wouldn't be able to detect this without actually checking the content of the validation error!

We should narrow expected validation error texts for these tests.

@ErichDonGubler is happy to act as a mentor for whoever wants to take this on. 🙂

dv29:
Partially resolves #5727 (note: edited by @ErichDonGubler).

**Connections**
https://github.com/gfx-rs/wgpu/issues/5727

**Description**
Its adding better descriptions to failure tests to avoid breaking because of regression

**Testing**
`cargo xtask test`

**Checklist**

- [x] Run `cargo fmt`.
- [x] ~~Add change to `CHANGELOG.md`. See simple instructions inside file.~~ EDIT from @ErichDonGubler: Not necessary. 🙂

### DIFF
--- a/xtask/src/test.rs
+++ b/xtask/src/test.rs
@@ -18,12 +18,10 @@ pub fn run_tests(shell: Shell, mut args: Arguments) -> anyhow::Result<()> {
     };
     let llvm_cov_nextest_flags: &[_] = if llvm_cov {
         &["llvm-cov", "--no-cfg-coverage", "--no-report", "nextest"]
+    } else if list {
+        &["nextest", "list"]
     } else {
-        if list {
-            &["nextest", "list"]
-        } else {
-            &["nextest", "run"]
-        }
+        &["nextest", "run"]
     };
 
     log::info!("Generating .gpuconfig file based on gpus on the system");
--- a/tests/tests/bind_group_layout_dedup.rs
+++ b/tests/tests/bind_group_layout_dedup.rs
@@ -373,7 +373,7 @@ fn separate_pipelines_have_incompatible_derived_bgls(ctx: TestingContext) {
         || {
             drop(pass);
         },
-        None,
+        Some("label at index 0 is not compatible with the corresponding bindgrouplayout"),
     );
 }
 
@@ -445,7 +445,7 @@ fn derived_bgls_incompatible_with_regular_bgls(ctx: TestingContext) {
         || {
             drop(pass);
         },
-        None,
+        Some("label at index 0 is not compatible with the corresponding bindgrouplayout"),
     )
 }
 
--- a/tests/tests/buffer.rs
+++ b/tests/tests/buffer.rs
@@ -230,7 +230,7 @@ static MINIMUM_BUFFER_BINDING_SIZE_LAYOUT: GpuTestConfiguration = GpuTestConfigu
                         cache: None,
                     });
             },
-            None,
+            Some("shader global resourcebinding { group: 0, binding: 0 } is not available in the pipeline layout"),
         );
     });
 
@@ -335,7 +335,7 @@ static MINIMUM_BUFFER_BINDING_SIZE_DISPATCH: GpuTestConfiguration = GpuTestConfi
                 drop(pass);
                 let _ = encoder.finish();
             },
-            None,
+            Some("buffer is bound with size 16 where the shader expects 32 in group[0] compact index 0"),
         );
     });
 
--- a/tests/tests/buffer_copy.rs
+++ b/tests/tests/buffer_copy.rs
@@ -8,31 +8,62 @@ fn try_copy(
     ctx: &wgpu_test::TestingContext,
     offset: BufferAddress,
     size: BufferAddress,
-    should_fail: bool,
+    error_message: Option<&'static str>,
 ) {
     let buffer = ctx.device.create_buffer(&BUFFER_DESCRIPTOR);
     let data = vec![255; size as usize];
+
     fail_if(
         &ctx.device,
-        should_fail,
+        error_message.is_some(),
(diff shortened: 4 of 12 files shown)


xtask/src/test.rs
```diff
@@ -18,12 +18,10 @@ pub fn run_tests(shell: Shell, mut args: Arguments) -> anyhow::Result<()> {
...
         &["llvm-cov", "--no-cfg-coverage", "--no-report", "nextest"]
+    } else if list {
+        &["nextest", "list"]
     } else {
-        if list {
-            &["nextest", "list"]
-        } else {
-            &["nextest", "run"]
-        }
+        &["nextest", "run"]
```
ErichDonGubler, line 24: nitpick(non-blocking): This change looks like something that should be a separate commit, or maybe even a PR. I'm guessing `clippy` complained at you about this?

dv29: Yup it was `clippy`, I'll revert this

ErichDonGubler: You can even keep the change! It's good to apply `clippy` suggestions that fire for this code. 🙂 Just want to make sure individual commits have specific changes. Otherwise, it can be a hazard for source management operations like `git revert` if something needs to reverted.

ErichDonGubler:
@dv29: Just seeing <https://github.com/gfx-rs/wgpu/issues/5727#issuecomment-2243931638> now. Did you want to convert this to a draft and add some more changes (for a re-review) before merging this?

dv29:
> @dv29: Just seeing [#5727 (comment)](https://github.com/gfx-rs/wgpu/issues/5727#issuecomment-2243931638) now. Did you want to convert this to a draft and add some more changes (for a re-review) before merging this?

Yes it is Draft PR. Are the changes what you expected, if so I'll go ahead and do the other ones

ErichDonGubler:
@dv29:

> Yes it is Draft PR. Are the changes what you expected, if so I'll go ahead and do the other ones

Ah, sorry, I wasn't very explicit in confirming this, was I? Yes, this direction looks good, overall. 🙂 I'm not sure how I feel about the entire message being lowercased, but then again, we _are_ doing a case-insensitive match against the provided message.

Don't feel the need to copy the entire diagnostic text into an expected message argument; if you can reduce it to something that guarantees that the error has _some_ sensible content, then that also works. This is, however, subjective, so copying the entire diagnostic text is safe for merging, even if maybe more prone to needing updating over time.

ErichDonGubler:
@dv29: Looks like some messages are failing CI now. At least some of them are because the backend mentioned doesn't not apply on the tested platform.

🤔 Maybe we need a way of expressing expected message _patterns_, rather than dumb case-insensitive matches? CC @cwfitzgerald.

ErichDonGubler:
@dv29: If nothing else, we could punt on the problematic cases where an ID is interspersed in critical portions of the error message to another PR where we figure out the right solution. I suggest extracting the problematic changes into a commit outside this PR's branch for now.

ErichDonGubler:
Sorry, @JCapucho, I incorrectly clicked on your name in the `Assignees` list. 😓

dv29:
> @dv29: If nothing else, we could punt on the problematic cases where an ID is interspersed in critical portions of the error message to another PR where we figure out the right solution. I suggest extracting the problematic changes into a commit outside this PR's branch for now.

Gotcha, I'll try to see if I can resolve them for the platform (with what we have right now). If not I'll move them to another PR.

JCapucho:
> Sorry, @JCapucho, I incorrectly clicked on your name in the `Assignees` list. 😓

No problem, it happens :)

dv29:
@ErichDonGubler do you think we could do a regex error matches? i.e pass in regex expression string and it tries to match portion of it?

ErichDonGubler:
Hey @dv29! Sorry for taking so long to get back with you. I've been extremely heads-down on a large backlog of some WebGPU CTS work in Firefox that's been consuming all of my work time. 😅

I think case-sensitive regexes are a good idea, and certainly easier to understand than the current behavior. Any objections, @cwfitzgerald?

cwfitzgerald:
Yeah regex would be fine (maybe lets use regex-lite?) but I think that should be opt in with a different method, so that people don't need to worry about escaping when just string matching.

dv29:
> Yeah regex would be fine (maybe lets use regex-lite?) but I think that should be opt in with a different method, so that people don't need to worry about escaping when just string matching.

so we should support both string and regex input? and maybe convert the string to regex internally at the function level?

teoxoy:
The IDs in errors will go away soon (with the removal of the registries), so I don't think we should complicate things with regexes.

dv29:
Should I then undo the test cases that are failing and create another PR with a different solution for it and close this one?

teoxoy:
I'm not seeing any CI failures in this PR, can't we just merge it as is?

ErichDonGubler:
@teoxoy: I hadn't been paying much attention to the content of the PR and check status. I hadn't noticed that the question of platform-specific error messages had actually been resolved! 😅

I think we should be good with the current contents of this PR, and iterate further as we discover more pain points.

ErichDonGubler:
Squashed the fixup commit at the tip of the PR into the first commit, so we can rebase-and-merge.

ErichDonGubler (approved):
Fixed up one nit (`s/Option::None/None`).

PR #6020, opened 2024-07-23
Use workspace dependencies more.

waywardmonkeys:
**Description**
In the past, workspace dependencies couldn't be used everywhere due to complications in the Firefox environment. These are believed to be resolved and so we can try using them again to reduce version management overhead and duplicated information.

This patch makes some dependencies into workspace dependencies.

The `bitflags` dependency is changed to version `2.6` to match what was in naga rather than the bare `2` used elsewhere.

**Testing**
Compilation.

**Checklist**

- [x] Run `cargo fmt`.
- [X] ~~Add change to `CHANGELOG.md`. See simple instructions inside file.~~ EDIT by @ErichDonGubler: Not necessary. 🙂

### DIFF
--- a/wgpu-hal/Cargo.toml
+++ b/wgpu-hal/Cargo.toml
@@ -110,20 +110,20 @@ name = "raw-gles"
 required-features = ["gles"]
 
 [dependencies]
-bitflags = "2"
-parking_lot = ">=0.11, <0.13"
-profiling = { version = "1", default-features = false }
-raw-window-handle = "0.6"
-thiserror = "1"
-once_cell = "1.19.0"
+bitflags.workspace = true
+parking_lot.workspace = true
+profiling = { workspace = true, default-features = false }
+raw-window-handle.workspace = true
+thiserror.workspace = true
+once_cell.workspace = true
 
 # backends common
-arrayvec = "0.7"
-rustc-hash = "1.1"
-log = "0.4"
+arrayvec.workspace = true
+rustc-hash.workspace = true
+log.workspace = true
 
 # backend: Gles
-glow = { version = "0.14.0", optional = true }
+glow = { workspace = true, optional = true }
 
 [dependencies.wgt]
 package = "wgpu-types"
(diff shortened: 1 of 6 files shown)


waywardmonkeys:
This could be a draft and get updated as more dependencies are switched over, or we could use this to be sure that there aren't issues from enabling this.

wgpu-hal/Cargo.toml
```diff
@@ -132,20 +132,20 @@ version = "22.0.0"
...
 #Note: it's unused by emscripten, but we keep it to have single code base in egl.rs
-libloading = { version = ">=0.7, <0.9", optional = true }
+libloading = { workspace = true, optional = true }
 
 [target.'cfg(windows)'.dependencies]
 # backend: Dx12
-bit-set = { version = "0.8", optional = true }
+bit-set = { workspace = true, optional = true }
 range-alloc = { version = "0.1", optional = true }
 gpu-allocator = { version = "0.27", default-features = false, features = [
```
MarijnS95: These and everything below (outside of commentable diff range) are in the workspace table as well; any reason to _not_ inherit them?

waywardmonkeys: gpu-allocator is v0.26 in the workspace and I wanted to keep this as straight changes so far. (Nothing uses the v0.26 so there's no issue in updating it.)

I plan to do more, either as part of this PR or followups, but I wanted to make sure that there was no issue with me doing this to avoid potentially wasted effort.

Wumpf:
looping in @ErichDonGubler to confirm we can have this since as pointed out [here](https://github.com/gfx-rs/wgpu/pull/6006#discussion_r1687345503) this used to be an issue for Mozilla's vendoring

wgpu-hal/Cargo.toml
MarijnS95: Just like you did for `bitflags`, let's bring the version in the workspace table in sync since this is the only user?

teoxoy:
The was a bug with `cargo vendor` that required us to revert to non-inherited deps for the crates Firefox uses. It was fixed in rust 1.68 so we can now fully use workspace inheritance. https://github.com/gfx-rs/wgpu/issues/3407 is the bug tracking this.

ErichDonGubler:
+1 to @teoxoy's comment; AFAIK, there are no blockers to merging this now.

wgpu-hal/Cargo.toml
ErichDonGubler: Yeah, let's make merge the 0.27 over to `Cargo.toml`, perhaps as a separate PR to not block this one? 👍🏻

ErichDonGubler:
I'm going to merge this in a few hours, unless somebody has an objection (CC @gfx-rs/wgpu); I think the potential for conflicts is high, and I haven't felt or seen any controversy that _what's here already is fine by itself_.

ErichDonGubler (approved):

waywardmonkeys:
I've rebased this forward, added in the further work that I'd done.

cwfitzgerald (approved):

teoxoy:
Should this close https://github.com/gfx-rs/wgpu/issues/3407? I think so. I don't think we want all the deps in the workspace toml, only the ones present in more than one crate.

MarijnS95:
> Should this close #3407? I think so. I don't think we want all the deps in the workspace toml, only the ones present in more than one crate.

To me it seems like tracking all versions in one coherent place is nice. That way you also don't have to be aware of this and go move dependencies back and forth when they start being used in >1 place, or have their use removed and end up remaining used by just one crate.

teoxoy:
That's fair, let's keep it open for the time being then.

ErichDonGubler:
Just waiting on @gfx-rs/naga's approval here.

teoxoy (approved):

PR #6021, opened 2024-07-23
Undo doing less bench iterations under `cfg(test)`

Wumpf:
Messed this up recently, was under the impression that this limits the iteration count only when running tests. Need to find another way to do that.

Still don't quite understand why that happens, I was fairly convinced that `cfg(test)` [is not set for benchmarks in `benches` directory](https://users.rust-lang.org/t/what-are-the-rules-for-cfg-test/54122/2).

PR #6030, opened 2024-07-24
Reduce the number of iterations in benchmarks to a small number when …

nical:
**Connections**

#6021

**Description**

We want to run the benchmarks with a very small number of iterations on CI (not to get meaningful numbers out of it but to ensure they don't break). This is done by checking whether the "WGPU_TESTING" environment variable is set.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,7 @@ env:
   RUSTDOCFLAGS: -D warnings
   WASM_BINDGEN_TEST_TIMEOUT: 300 # 5 minutes
   CACHE_SUFFIX: c # cache busting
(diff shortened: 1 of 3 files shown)


teoxoy:
Could you do this for `DISPATCH_COUNT` & `DISPATCH_COUNT_BINDLESS` as well? `DISPATCH_COUNT` was the problematic one on macOS (https://github.com/gfx-rs/wgpu/pull/6021#issuecomment-2247444209).

nical:
ah yes of course :facepalm:

teoxoy (approved):
Thanks!

Issue #5890, opened 2024-06-28
Bunnymark example does not recompute its projection matrix and other parameters when the canvas size changes


mwyrzykowski:
**Description**
examples/src/bunnymark/mod.rs computes an Orthographic RH matrix on page load, but if the canvas is resized, the matrix is not recomputed. 

See all uses of config.width and config.height inside examples/src/bunnymark/mod.rs

**Repro steps**
(1) Open https://wgpu.rs/examples/?backend=webgpu&example=bunnymark in Chrome 126.0.6478.127
(2) Resize the window

**Expected vs observed behavior**
Observed: Bunnies bounce outside of the canvas or don't bounce to the top

Expected: bunnies bounce within the resized canvas

**Extra materials**

https://github.com/gfx-rs/wgpu/assets/105083895/a41bfd50-c2df-4200-937a-0ee0bcf3bc5a

**Platform**
macOS using Chrome Version 128.0.6562.0 (Official Build) canary (arm64)

dv29:
Can i take this up? has there been any discovery done on this?

Wumpf:
@dv29 doesn't look like, go ahead!

dv29:
I cannot replicate this issue on my machine with the following, it seems to be working as expected

```
OS:
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.4 LTS
Release:	22.04
Codename:	jammy

Google Chrome: Version 126.0.6478.182 (Official Build) (64-bit)
```

gabrieleforner:
Me too, I tried even by debugging it and it works nice. Try checking your GPU manufacturer drivers.


OS
Name: Apple macOS 14.5
Real. Number: 14.5
Codename: Sonoma

mwyrzykowski:
@dv29 @techforni  I am using Chrome Canary version 128.0.6610.0 (Official Build) canary (arm64) from macOS 15 beta and I think the issue or a similar one persists, specifically the implementation of `resize` is currently empty:

https://github.com/gfx-rs/wgpu/blob/101c996703ba5701cfe38efc677e76b9083592de/examples/src/bunnymark/mod.rs#L373

I would propose something along the lines of the following to examples/src/bunnymark/mod.rs, pardon my rust syntax as I have never written a line of rust before:
```
diff --git a/examples/src/bunnymark/mod.rs b/examples/src/bunnymark/mod.rs
index b5b33b54d..3823695d7 100644
[...]

PR #6034, opened 2024-07-24
fix: 5890 set upper bound for vertical jump

dv29:
**Connections**
#5890

**Description**
Its adding bounds to the vertical jump on bunnymark example. It will also update size of the surface on resize.

**Testing**
1. `cargo xtask run-wasm`
2. go to `http://localhost:8000/?backend=webgpu&example=bunnymark`
3. The bunnies should not jump out of the vertical bounds and should resize when the screen size changes.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/examples/src/bunnymark/mod.rs
+++ b/examples/src/bunnymark/mod.rs
@@ -34,12 +34,19 @@ impl Bunny {
         self.position[0] += self.velocity[0] * delta;
         self.position[1] += self.velocity[1] * delta;
         self.velocity[1] += GRAVITY * delta;
+
         if (self.velocity[0] > 0.0 && self.position[0] + 0.5 * BUNNY_SIZE > extent[0] as f32)
             || (self.velocity[0] < 0.0 && self.position[0] - 0.5 * BUNNY_SIZE < 0.0)
         {
             self.velocity[0] *= -1.0;
         }
+
         if self.velocity[1] < 0.0 && self.position[1] < 0.5 * BUNNY_SIZE {
+            self.velocity[1] *= -1.0;
+        }
+
+        // Top boundary check
+        if self.velocity[1] > 0.0 && self.position[1] + 0.5 * BUNNY_SIZE > extent[1] as f32 {
             self.velocity[1] *= -1.0;
         }
     }
@@ -47,6 +54,9 @@ impl Bunny {
 
 /// Example struct holds references to wgpu resources and frame persistent data
 struct Example {
+    view: wgpu::TextureView,
+    sampler: wgpu::Sampler,
+    global_bind_group_layout: wgpu::BindGroupLayout,
     global_group: wgpu::BindGroup,
     local_group: wgpu::BindGroup,
     pipeline: wgpu::RenderPipeline,
@@ -286,6 +296,7 @@ impl crate::framework::Example for Example {
             size: [BUNNY_SIZE; 2],
             pad: [0.0; 2],
         };
+
         let global_buffer = device.create_buffer_init(&wgpu::util::BufferInitDescriptor {
             label: Some("global"),
             contents: bytemuck::bytes_of(&globals),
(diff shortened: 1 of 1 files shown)


ErichDonGubler:
Based on https://github.com/gfx-rs/wgpu/issues/5890#issuecomment-2248051047, this appears to still be a draft. Marking as such.

examples/src/bunnymark/mod.rs
```diff
@@ -366,11 +373,11 @@ impl crate::framework::Example for Example {
 
     fn resize(
         &mut self,
-        _sc_desc: &wgpu::SurfaceConfiguration,
+        sc_desc: &wgpu::SurfaceConfiguration,
         _device: &wgpu::Device,
         _queue: &wgpu::Queue,
     ) {
-        //empty
```
mwyrzykowski, line 373: It would be good to recompute the orthographic matrix too, which is computed on line 277:
```
        let globals = Globals {
            mvp: glam::Mat4::orthographic_rh(
                0.0,
                config.width as f32,
                0.0,
                config.height as f32,
                -1.0,
                1.0,
```
and also update the GPUBuffer with `queue.writeBuffer`:
```
        let global_buffer = device.create_buffer_init(&wgpu::util::BufferInitDescriptor {
            label: Some("global"),
            contents: bytemuck::bytes_of(&globals),
            usage: wgpu::BufferUsages::COPY_DST | wgpu::BufferUsages::UNIFORM,
        });
```

otherwise the matrix will be fixed to the original window size which seems incorrect.

dv29: thats what I'm planning to do but there are few other things that depend on this so i'd have to extract it all out to reuse it

dv29:
This has been updated to apply the fix for resize, please take a look when you get a chance

examples/src/bunnymark/mod.rs
```diff
@@ -366,11 +380,51 @@ impl crate::framework::Example for Example {
 
     fn resize(
```
mwyrzykowski, line 381: I'm not sure if this is intentional, but unlike in https://github.com/gfx-rs/wgpu/issues/5890#issuecomment-2242094769, the bunnies do not get repositioned within the window's new vertical bounds.

So bunnies bouncing within a 150pt tall window will continue bouncing in that small 150pt vertical range even if the window is resized to 2000 points, for instance:


https://github.com/user-attachments/assets/75ec0d4a-2086-4eb1-83d0-c5f172896650

mwyrzykowski: The bunnies spawn correctly upon pressing spacebar, but the ones which already spawned bounce in this limited range

mwyrzykowski (commented):
This fixes the orthographic matrix, I'm not sure if this fix should also include repositioning existing bunnies on window resize.

examples/src/bunnymark/mod.rs
dv29: I'm taking a looking this weekend, sorry have been really busy

dv29: I think the jump speed is slow as it grows, try leaving it for a bit and it should jump higher. This is what i'm seeing on my side.

mwyrzykowski: I guess it depends on the intent of the sample. My expectation would be they would be repositioned relative to the new bounds of the canvas when the window is resized.

Either way is fine from my perspective 👍, I am not a wgpu contributor so ultimately your decision :)

dv29: I'm new here as well so I'll defer it to @ErichDonGubler

ErichDonGubler: bro idk 😆💦

Gonna punt to the rest of @gfx-rs/wgpu for opinion on this. I won't, however, punt on merging this, because stale PRs are no fun, and this behavior doesn't really seem risky, just subject to opinion.

@gfx-rs/wgpu: I'm gonna merge this. If you have qualms about current behavior, then let's fix/discuss it in a separate PR.

ErichDonGubler (approved):

PR #6041, opened 2024-07-24
Fix error message in create_render_pass

matthew-wong1:
**Description**
Hi! I came across an error message in create_render_pass that mentions `compute_pass` instead of `render_pass`, which I think might be a mistake. I updated it to say `render_pass` instead. 

**Testing**
Ran `cargo xtask test`

**Checklist**

- [x] Run `cargo fmt`.


ErichDonGubler (approved):
Thanks!

Issue #6043, opened 2024-07-25
Device::maintain overly verbose logging


AthosOfAthos:
After updating to wgpu 22.0.0 applications log at least one "Device::maintain: waiting for submission index x" Info log per frame.

**Repro steps**
The wgpu shadow example provides a good example of this.
```cargo run --bin wgpu-examples shadow```
[log.txt](https://github.com/user-attachments/files/16368885/log.txt)

**Expected vs observed behavior**
wgpu log messages drown out other log messages with log verbosity set to Level::Info. I believe this message should be either debug or trace with how verbose it is.

PR #6044, opened 2024-07-25
Reduced verbosity for 'Device::maintain: waiting for submission index…

AthosOfAthos:
Quick fix for #6043.

**Description**
Reduces verbosity of obnoxiously common log message from info to debug that was introduced in wgpu v22.0.0.

### DIFF
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -443,7 +443,7 @@ impl<A: HalApi> Device<A> {
                     .map_err(DeviceError::from)?
             };
         }
-        log::info!("Device::maintain: waiting for submission index {submission_index}");
+        log::debug!("Device::maintain: waiting for submission index {submission_index}");
 
         let mut life_tracker = self.lock_life();
         let submission_closures =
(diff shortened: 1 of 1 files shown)


nical (commented):
Thanks! Let's demote this all the way to `trace`-level.

Elabajaba:
> Thanks! Let's demote this all the way to `trace`-level.

Shouldn't this be a `resource_log`?

nical:
Resource logs are specific to the creation and destruction of objects and resources.

wgpu-core/src/device/resource.rs
```diff
@@ -443,7 +443,7 @@ impl<A: HalApi> Device<A> {
                     .map_err(DeviceError::from)?
             };
         }
-        log::info!("Device::maintain: waiting for submission index {submission_index}");
+        log::debug!("Device::maintain: waiting for submission index {submission_index}");
```
nical:
```suggestion
        log::trace!("Device::maintain: waiting for submission index {submission_index}");
```

nical (approved):

Kek5chen:
Quick question. Am not too deep in the project, but shouldn't this log statement also be inside of the if statement? Since that's where it's actually waiting? If the if statement is skipped it shouldn't print that it's waiting, since it's not. Or do I see this wrong? Can open an issue if you tell me to.

ErichDonGubler:
@Kek5chen: Filed a PR to ask the question a bit more pointedly (😉): https://github.com/gfx-rs/wgpu/pull/6303

PR #5640, opened 2024-04-30
Cast ptr to Device not Surface

sagudev:
Arc has device stored inside not surface.

fixes https://github.com/gfx-rs/wgpu/issues/5637

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`. If applicable, add:
  - [x] `--target wasm32-unknown-unknown`
  - [x] `--target wasm32-unknown-emscripten`
- [x] Run `cargo xtask test` to run tests.
- [ ] Add change to `CHANGELOG.md`. See simple instructions inside file.

Issue #5637, opened 2024-04-30
STATUS_ACCESS_VIOLATION on exit


daniel5gh:
**Description**
Since I have updated from 0.19 to 0.20, I am facing an issue similar to #1377.

When my program exits after the first use of the device, in a call `surface.configure(&device, &config)`, the program exits with STATUS_ACCESS_VIOLATION when exiting through an anyhow error bubbled up to pollster or a regular exit with an `Ok(())` result.

When I use the `panic!()` macro, it seems fine. See the various exit scenarios commented in the snippet below.

**Repro steps**

Extracted minimal reproduction from my Main Project in which I am using pollster, anyhow and winit.

```rust
use anyhow::{Context, Result};
use std::sync::Arc;
use wgpu::Features;
use winit::event_loop::{EventLoop};
use winit::window::WindowBuilder;


#[pollster::main]
async fn main() -> Result<()> {
    let event_loop = EventLoop::new()
        .context("Failed to create event loop")?;

    let window = Arc::new(WindowBuilder::new()
        .with_title("Panic! at the disco")
        .build(&event_loop)
        .context("Failed to create window")?);

    let instance = wgpu::Instance::new(wgpu::InstanceDescriptor {
        backends: wgpu::Backends::all(),
        ..Default::default()
    });

    let surface = instance.create_surface(window.clone())
        .context("Failed to create surface")?;

    let adapter = instance.request_adapter(
        &wgpu::RequestAdapterOptions {
            power_preference: wgpu::PowerPreference::HighPerformance,
            force_fallback_adapter: false,
            compatible_surface: Some(&surface),
        })
        .await
        .context("Failed to find an appropriate adapter")?;

    let (device, _queue) = adapter
        .request_device(
            &wgpu::DeviceDescriptor {
                label: None,
                required_features: Features::default(),
                required_limits: Default::default(),
            },
            None,
        )
        .await
        .context("Failed to create device / queue")?;

    let surface_caps = surface.get_capabilities(&adapter);
    let surface_format = surface_caps.formats.iter()
        .filter(|format| format.is_srgb())
        .next()
        .unwrap_or(surface_caps.formats.first().context("No surface formats found")?);

    let config = wgpu::SurfaceConfiguration {
        usage: wgpu::TextureUsages::RENDER_ATTACHMENT,
        format: *surface_format,
        width: window.inner_size().width,
        height: window.inner_size().height,
        present_mode: surface_caps.present_modes[0],
[...]

PR #6052, opened 2024-07-26
Fix AnyDevice drop implementation dropping the wrong thing

Wumpf:
**Connections**
* https://github.com/emilk/egui/issues/4874
* `AnyDevice` can hopefully go away soon as part of my ongoing generification work
     *  https://github.com/gfx-rs/wgpu/pull/6002

Post-merge edit:
* Corrected fix of https://github.com/gfx-rs/wgpu/pull/5640
    * Making this the actual fix for https://github.com/gfx-rs/wgpu/issues/5637

**Description**
There is a [crash report in egui](https://github.com/emilk/egui/issues/4874) after wgpu update. Quickly realized that this happens only when the surface is destroyed _after_ the all other resources are dropped. Which leads to this simple repro in `hello_triangle` (cc: @cwfitzgerald any idea how I can unit-testify this? :/)

```diff
diff --git a/examples/src/hello_triangle/mod.rs b/examples/src/hello_triangle/mod.rs
index 7c82d49cf..418bf5c66 100644
--- a/examples/src/hello_triangle/mod.rs
+++ b/examples/src/hello_triangle/mod.rs
@@ -1,4 +1,4 @@
-use std::borrow::Cow;
+use std::{borrow::Cow, sync::Arc};
 use winit::{
     event::{Event, WindowEvent},
     event_loop::EventLoop,
@@ -80,6 +80,8 @@ async fn run(event_loop: EventLoop<()>, window: Window) {
         .get_default_config(&adapter, size.width, size.height)
         .unwrap();
     surface.configure(&device, &config);
+    let surface = Arc::new(surface);
+    let surface_b = surface.clone();
 
     let window = &window;
     event_loop
@@ -143,6 +145,8 @@ async fn run(event_loop: EventLoop<()>, window: Window) {
             }
         })
         .unwrap();
+
+    drop(surface_b);
 }
 
 pub fn main() {
```

Unknown why this happened after wgpu 22.0.0 update, I strongly suspect the error is older and just surfaces (hah!) now because of improving ownership hygiene in the project.

A shining case of "How did this ever work?"
(A: this still drops the right ref count, so this is ONLY an issue if the last droping of `Arc<Device<A>>` happens through `AnyDevice` which is surprisingly rare)

**Testing**
Above manual test works now

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ Bottom level categories:
 - Print requested and supported usages on `UnsupportedUsage` error. By @VladasZ in [#6007](https://github.com/gfx-rs/wgpu/pull/6007)
 - Fix function for checking bind compatibility to error instead of panic. By @sagudev [#6012](https://github.com/gfx-rs/wgpu/pull/6012)
 - Deduplicate bind group layouts that are created from pipelines with "auto" layouts. By @teoxoy [#6049](https://github.com/gfx-rs/wgpu/pull/6049)
+- Fix crash when dropping the surface after the device. By @wumpf in [#6052](https://github.com/gfx-rs/wgpu/pull/6052)
 
 ### Dependency Updates
 
--- a/wgpu-core/src/device/any_device.rs
+++ b/wgpu-core/src/device/any_device.rs
@@ -34,7 +34,7 @@ impl AnyDevice {
         unsafe fn drop_glue<A: HalApi>(ptr: *mut ()) {
             // Drop the arc this instance is holding.
             unsafe {
-                _ = Arc::from_raw(ptr.cast::<A::Device>());
+                _ = Arc::from_raw(ptr.cast::<Device<A>>());
             }
         }
 


teoxoy (approved):

Wumpf:
(force push to fill in the changelog pr numbers, no other change)

sagudev:
> any idea how I can unit-testify this?

Simple test that inits every possible resource (device, buffer, texture, ...) and then runs all permutations of drops.

Wumpf:
that would be a good unit test to have regardless indeed! The main problem specifically here though is that we need to create a `Surface` and we're still having issues to do so on CI afaik

sagudev:
Maybe [xvfb-run](https://manpages.ubuntu.com/manpages/xenial/man1/xvfb-run.1.html) can help (for linux), that's what we [use in servo](https://github.com/servo/servo/blob/bc1c71bd4df9424e4cceed9dc8cc38efc591095b/.github/workflows/linux.yml#L117).

PR #6073, opened 2024-08-02
Tracker simplifications

teoxoy:
**Connections**
\-

**Description**
Various tracker simplifications (see each individual commit for details).

**Testing**
By existing tests in CI.

### DIFF
--- a/wgpu-core/src/track/mod.rs
+++ b/wgpu-core/src/track/mod.rs
@@ -681,25 +681,4 @@ impl<A: HalApi> Tracker<A> {
                 .set_and_remove_from_usage_scope_sparse(&mut scope.textures, &bind_group.textures)
         };
     }
-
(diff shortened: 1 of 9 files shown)


wgpu-core/src/track/mod.rs
```diff
@@ -443,9 +441,7 @@ impl<A: HalApi> BindGroupStates<A> {
     /// accesses will be in a constant ascending order.
     pub fn optimize(&mut self) {
         self.buffers.optimize();
-        self.textures.optimize();
         self.views.optimize();
-        self.samplers.optimize();
```
cwfitzgerald, line 448: Could we leave a note on all the places where we're omitting certain types of resources to know it was intentional and why it's valid

cwfitzgerald (approved):
Amazing stuff!

PR #6076, opened 2024-08-03
Handle TooManyAttachments in wgpu-core

### LINKED ISSUE #6075: wgpu-core: panic if more then 8 color attachment passed
**Description**
When too many attachment are passed to `begin_rander_pass` it panics on https://github.com/gfx-rs/wgpu/blob/7b4cbc26192d6d56a31f8e67769e656a6627b222/wgpu-core/src/command/render.rs#L1370-L1376
with: 
```
called `Result::unwrap()` on an `Err` value: CapacityError: insufficient capacity
```

because color_attachments is array with fixed size, bound check would be needed before, this validation is also required by spec: https://www.w3.org/TR/webgpu/#ref-for-abstract-opdef-gpurenderpassdescriptor-valid-usage, https://www.w3.org/TR/webgpu/#abstract-opdef-gpurenderpassdescriptor-valid-usage. This is not reproducible on high-level wgpu crate because:
https://github.com/gfx-rs/wgpu/blob/7b4cbc26192d6d56a31f8e67769e656a6627b222/wgpu/src/backend/wgpu_core.rs#L1926-L1934, so we just need to move this check to from wgpu to wgpu-core.

**Repro steps**
Here is a repro repo: https://github.com/sagudev/wgpu-problem/tree/panic-on-too-many-attachments

**Platform**
Happens on current trunk.

sagudev:
**Connections**
Fixes #6075 

**Testing**
Fixes repro.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/command/mod.rs
+++ b/wgpu-core/src/command/mod.rs
@@ -596,6 +596,8 @@ pub enum CommandEncoderError {
     InvalidTimestampWritesQuerySetId(id::QuerySetId),
     #[error("Attachment TextureViewId {0:?} is invalid")]
     InvalidAttachmentId(id::TextureViewId),
+    #[error(transparent)]
+    InvalidColorAttachment(#[from] ColorAttachmentError),
     #[error("Resolve attachment TextureViewId {0:?} is invalid")]
     InvalidResolveTargetId(id::TextureViewId),
     #[error("Depth stencil attachment TextureViewId {0:?} is invalid")]
--- a/wgpu-core/src/command/render.rs
+++ b/wgpu-core/src/command/render.rs
@@ -1342,17 +1342,17 @@ impl Global {
             hub: &crate::hub::Hub<A>,
             desc: &RenderPassDescriptor<'_>,
             arc_desc: &mut ArcRenderPassDescriptor<A>,
+            device: &Device<A>,
         ) -> Result<(), CommandEncoderError> {
             let query_sets = hub.query_sets.read();
             let texture_views = hub.texture_views.read();
 
-            // TODO: should be device.limits.max_color_attachments
-            let max_color_attachments = hal::MAX_COLOR_ATTACHMENTS;
+            let max_color_attachments = device.limits.max_color_attachments as usize;
             if desc.color_attachments.len() > max_color_attachments {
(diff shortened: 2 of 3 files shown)


wgpu-core/src/command/mod.rs
```diff
@@ -596,6 +596,8 @@ pub enum CommandEncoderError {
     InvalidTimestampWritesQuerySetId(id::QuerySetId),
     #[error("Attachment TextureViewId {0:?} is invalid")]
     InvalidAttachmentId(id::TextureViewId),
+    #[error(transparent)]
+    InvalidColorAttachment(#[from] ColorAttachmentError),
```
sagudev, line 600: Should we inline `TooMany` as it's the only variant that is used?

wgpu-core/src/command/render.rs
```diff
@@ -1346,6 +1346,15 @@ impl Global {
             let query_sets = hub.query_sets.read();
             let texture_views = hub.texture_views.read();
 
+            if desc.color_attachments.len() > hal::MAX_COLOR_ATTACHMENTS {
```
sagudev: Should this check use limits from device?

teoxoy: It should.

wgpu-core/src/command/mod.rs
teoxoy: I think it's fine as it is, it's not the only place where we only use a subset.

wgpu-core/src/command/render.rs
```diff
@@ -1447,7 +1458,7 @@ impl Global {
             Err(e) => return make_err(e, arc_desc),
         };
 
-        let err = fill_arc_desc(hub, desc, &mut arc_desc).err();
+        let err = fill_arc_desc(hub, desc, &mut arc_desc, &cmd_buf.device).err();
```
sagudev, line 1461: Should there also be `snatchable_lock.read()`?

teoxoy: The snatch lock is only needed for the `Snatchable` fields of a few types.

teoxoy (approved):
Thanks!

PR #6089, opened 2024-08-08
[wgpu-core] Call `flush_mapped_ranges` when unmapping write-mapped buffers

teoxoy:
**Connections**
\-

**Description**
`flush_mapped_ranges` needs to be called for non-coherent write-mappings so that the data becomes visible to the GPU.
We also need to call it for non-coherent read-mappings in cases where the memory needs to be 0 initialized.

This is only needed for Vulkan and GL but even then, I'm not sure how things worked so far.

Vulkan says:

> Unmapping non-coherent memory does not implicitly flush the host mapped memory, and host writes that have not been flushed may not ever be visible to the device.

GL says:

> If a buffer range is mapped with both GL_MAP_PERSISTENT_BIT and GL_MAP_FLUSH_EXPLICIT_BIT set, then these commands may be called to ensure that data written by the client into the flushed region becomes visible to the server.

Both specs seem to say that users that want to guarantee that the data is flushed should call the function. But they don't say that if users don't call the function the data won't be flushed.

So, it could be that the drivers are flushing internally?

**Testing**
Existing tests.


cwfitzgerald:
> This is only needed for Vulkan and GL but even then, I'm not sure how things worked so far.

Most devices have coherent memory in practice

ErichDonGubler (approved):

ErichDonGubler (approved):

PR #6090, opened 2024-08-08
Introduce `ExternalStagingBuffer` for staging buffer uses outside of `wgpu-core`

teoxoy:
**Connections**
\-

**Description**
I find that this refactor makes things clearer since most of the time we use staging buffers only internally.

**Testing**
Existing tests.

### DIFF
--- a/wgpu-core/src/device/queue.rs
+++ b/wgpu-core/src/device/queue.rs
@@ -17,8 +17,8 @@ use crate::{
     lock::RwLockWriteGuard,
     resource::{
         Buffer, BufferAccessError, BufferMapState, DestroyedBuffer, DestroyedResourceError,
-        DestroyedTexture, FlushedStagingBuffer, Labeled, ParentDevice, ResourceErrorIdent,
-        StagingBuffer, Texture, TextureInner, Trackable,
+        DestroyedTexture, ExternalStagingBuffer, FlushedStagingBuffer, Labeled, ParentDevice,
+        ResourceErrorIdent, StagingBuffer, Texture, TextureInner, Trackable,
     },
     resource_log,
     track::{self, Tracker, TrackerIndex},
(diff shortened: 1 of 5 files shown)


wgpu-core/src/device/queue.rs
```diff
@@ -483,6 +482,10 @@ impl Global {
             .and_then(Arc::into_inner)
             .ok_or_else(|| QueueWriteError::Transfer(TransferError::InvalidBufferId(buffer_id)))?;
 
+        // TODO: bubble up safety requirements, they currently hold but we should make functions
```
ErichDonGubler, line 485: question(blocking): Should this be resolved before merging?

ErichDonGubler (commented):
Current changes LGTM so far, but the unresolved `TODO` means I'm not sure if this should merge yet.

wgpu-core/src/device/queue.rs
teoxoy: I don't intend to resolve this in this PR. After degenerification and removal of registries we might not even need to resolve it since we can make the API safe by not giving users the pointer at all.

ErichDonGubler: Ah, shall we file an issue, then, and remove this comment? If we don't want an issue to track it, then I wonder if we should even include the comment. 

Will downgrade to non-blocking with a response to the above, trusting that you've thought this through.

ErichDonGubler:
@teoxoy: We gonna merge this? 👀

teoxoy:
It's not a priority right now, I was thinking I might get to remove the registries first which would make the situation clearer with regards to the safety discussion.

cwfitzgerald:
Demoting to draft due to plans to remove registries first.

cwfitzgerald:
I'm going to close this as stale - please update the PR and ask to be re-opened if you're interested in continuing to pursue this change.

PR #6092, opened 2024-08-08
Use `ManuallyDrop` in remaining places

teoxoy:


ErichDonGubler (approved):
My understanding of the benefit of these changes is that we continue to guarantee that `Drop` is the sole context that should have access to these resources when run, but we now no longer have to navigate an unnecessary `Option`al layer to get at resource value links that should always be present over the lifetime of these resources anyway. We make every other interaction now more ergonomic.

If the above sounds right, we're on the same page; shippit!

PR #6098, opened 2024-08-10
Degenerification part 1/3: `wgpu-hal` `DynDevice` & friends

Wumpf:
**Connections**
* Part of https://github.com/gfx-rs/wgpu/issues/5124
* Part of a series!
   * ➡️ #6098
   * #6099 
   * #6100 

**Description**
Introduces a new `Dyn` "backend" which uses dynamic dispatch & `std::any::Any` based casting while keeping changes to `wgpu-core` as minimal as possible.

Note that casts through `Any` could be avoided altogether, but right now part2 of the series relies on standard any casts, so I'd advocate to keep this as a future improvement.
Internal unboxing mechanism as proposed by @cwfitzgerald however sidesteps `Any` for the most part (except for debug assertion) which allows us to keep `wgpu-hal` backend unchanged for the most part which profits their isolation & direct consumers of ` wgpu-hal`. Concrete, this means that `destroy` methods continue to consume objects via move (which is both semantically & syntactically convenient).
Also note that this is intentionally not exposed as part of the public interface.

Speaking of future improvements:
Something that has been discussed in the maintainer group previously while this was already under way is to use `enum` based `DynResource` which would allow to trivially know the size of each resource and instead dispatch via `match`.
This PR series sticks with `Box` and `Any` in favor of expedience but is my belief that such refactors will be easier once this lands!

Commits have been cleaned up as good as possible to make them digestable for review, but there might be still some slight meandering in there hinting at earlier experimentations.

**Testing**
* Existing tests run.

**Checklist**

- [x] Run `cargo fmt`.
- ~[ ] Add change to `CHANGELOG.md`. See simple instructions inside file.~ changelog in Part 3

### DIFF
--- a/wgpu-hal/src/dx12/device.rs
+++ b/wgpu-hal/src/dx12/device.rs
@@ -689,7 +689,7 @@ impl crate::Device for super::Device {
 
     unsafe fn create_command_encoder(
         &self,
-        desc: &crate::CommandEncoderDescriptor<super::Api>,
+        desc: &crate::CommandEncoderDescriptor<super::Queue>,
     ) -> Result<super::CommandEncoder, DeviceError> {
         let allocator = self
             .raw
--- a/wgpu-hal/src/dx12/device.rs
+++ b/wgpu-hal/src/dx12/device.rs
@@ -1104,7 +1104,13 @@ impl crate::Device for super::Device {
 
     unsafe fn create_bind_group(
         &self,
-        desc: &crate::BindGroupDescriptor<super::Api>,
+        desc: &crate::BindGroupDescriptor<
+            super::BindGroupLayout,
+            super::Buffer,
+            super::Sampler,
+            super::TextureView,
+            super::AccelerationStructure,
+        >,
     ) -> Result<super::BindGroup, DeviceError> {
         let mut cpu_views = desc
             .layout
--- a/wgpu-hal/src/dx12/device.rs
+++ b/wgpu-hal/src/dx12/device.rs
@@ -1745,7 +1745,7 @@ impl crate::Device for super::Device {
 
     unsafe fn get_acceleration_structure_build_sizes<'a>(
         &self,
-        _desc: &crate::GetAccelerationStructureBuildSizesDescriptor<'a, super::Api>,
+        _desc: &crate::GetAccelerationStructureBuildSizesDescriptor<'a, super::Buffer>,
     ) -> crate::AccelerationStructureBuildSizes {
         // Implement using `GetRaytracingAccelerationStructurePrebuildInfo`:
         // https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#getraytracingaccelerationstructureprebuildinfo
--- a/wgpu-hal/src/dx12/device.rs
+++ b/wgpu-hal/src/dx12/device.rs
@@ -211,10 +211,10 @@ impl super::Device {
     /// allowed to be a subset of the vertex outputs.
     fn load_shader(
         &self,
-        stage: &crate::ProgrammableStage<super::Api>,
+        stage: &crate::ProgrammableStage<super::ShaderModule>,
         layout: &super::PipelineLayout,
         naga_stage: naga::ShaderStage,
-        fragment_stage: Option<&crate::ProgrammableStage<super::Api>>,
+        fragment_stage: Option<&crate::ProgrammableStage<super::ShaderModule>>,
     ) -> Result<super::CompiledShader, crate::PipelineError> {
         use naga::back::hlsl;
 
@@ -1320,7 +1320,11 @@ impl crate::Device for super::Device {
 
     unsafe fn create_render_pipeline(
         &self,
-        desc: &crate::RenderPipelineDescriptor<super::Api>,
+        desc: &crate::RenderPipelineDescriptor<
+            super::PipelineLayout,
+            super::ShaderModule,
+            super::PipelineCache,
+        >,
     ) -> Result<super::RenderPipeline, crate::PipelineError> {
         let (topology_class, topology) = conv::map_topology(desc.primitive.topology);
         let mut shader_stages = wgt::ShaderStages::VERTEX;
@@ -1515,7 +1519,11 @@ impl crate::Device for super::Device {
 
     unsafe fn create_compute_pipeline(
         &self,
-        desc: &crate::ComputePipelineDescriptor<super::Api>,
+        desc: &crate::ComputePipelineDescriptor<
+            super::PipelineLayout,
+            super::ShaderModule,
+            super::PipelineCache,
+        >,
     ) -> Result<super::ComputePipeline, crate::PipelineError> {
         let blob_cs =
             self.load_shader(&desc.stage, desc.layout, naga::ShaderStage::Compute, None)?;
@@ -1559,10 +1567,10 @@ impl crate::Device for super::Device {
     unsafe fn create_pipeline_cache(
         &self,
         _desc: &crate::PipelineCacheDescriptor<'_>,
-    ) -> Result<(), crate::PipelineCacheError> {
-        Ok(())
+    ) -> Result<super::PipelineCache, crate::PipelineCacheError> {
+        Ok(super::PipelineCache)
     }
-    unsafe fn destroy_pipeline_cache(&self, (): ()) {}
+    unsafe fn destroy_pipeline_cache(&self, _: super::PipelineCache) {}
 
     unsafe fn create_query_set(
         &self,
--- a/wgpu-hal/src/dx12/device.rs
+++ b/wgpu-hal/src/dx12/device.rs
@@ -776,7 +776,7 @@ impl crate::Device for super::Device {
 
     unsafe fn create_pipeline_layout(
         &self,
-        desc: &crate::PipelineLayoutDescriptor<super::Api>,
+        desc: &crate::PipelineLayoutDescriptor<super::BindGroupLayout>,
     ) -> Result<super::PipelineLayout, DeviceError> {
         use naga::back::hlsl;
         // Pipeline layouts are implemented as RootSignature for D3D12.
--- a/wgpu-hal/src/dynamic/command.rs
+++ b/wgpu-hal/src/dynamic/command.rs
@@ -1,5 +1,8 @@
+use std::ops::Range;
+
 use crate::{
-    BufferBarrier, BufferBinding, BufferCopy, CommandEncoder, DeviceError, Label, MemoryRange,
+    BufferBarrier, BufferBinding, BufferCopy, CommandEncoder, DeviceError, DynQuerySet, Label,
+    MemoryRange,
 };
 
 use super::{DynBuffer, DynResourceExt as _};
@@ -23,6 +26,19 @@ pub trait DynCommandEncoder: std::fmt::Debug {
     unsafe fn insert_debug_marker(&mut self, label: &str);
     unsafe fn begin_debug_marker(&mut self, group_label: &str);
     unsafe fn end_debug_marker(&mut self);
+
+    unsafe fn begin_query(&mut self, set: &dyn DynQuerySet, index: u32);
+    unsafe fn end_query(&mut self, set: &dyn DynQuerySet, index: u32);
+    unsafe fn write_timestamp(&mut self, set: &dyn DynQuerySet, index: u32);
+    unsafe fn reset_queries(&mut self, set: &dyn DynQuerySet, range: Range<u32>);
+    unsafe fn copy_query_results(
+        &mut self,
+        set: &dyn DynQuerySet,
+        range: Range<u32>,
+        buffer: &dyn DynBuffer,
+        offset: wgt::BufferAddress,
+        stride: wgt::BufferSize,
+    );
 
     unsafe fn set_index_buffer<'a>(
         &mut self,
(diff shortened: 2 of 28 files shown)


ErichDonGubler:
I'll commit to this first part in the next couple of days. 😀

wgpu-hal/src/dynamic/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+                fn as_any_mut(&mut self) -> &mut dyn std::any::Any {
+                    self
+                }
+            }
+        )*
+    };
+}
+pub(crate) use impl_dyn_resource;
+
+/// Extension trait for `DynResource` used by by implementations of various dynamic resource traits.
```
ErichDonGubler: typo: s/by by/by

```suggestion
/// Extension trait for `DynResource` used by implementations of various dynamic resource traits.
```

wgpu-hal/src/dynamic/mod.rs
```diff
@@ -0,0 +1,67 @@
+use std::any::Any;
+
+use wgt::WasmNotSendSync;
+
+/// Base trait for all resources, allows downcasting via `std::any::Any`.
```
ErichDonGubler: suggestion: Let's link!

```suggestion
/// Base trait for all resources, allows downcasting via [`std::any::Any`].
```

wgpu-hal/src/dynamic/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+
+/// Extension trait for `DynResource` used by by implementations of various dynamic resource traits.
+trait DynResourceExt {
+    fn expect_downcast_ref<T: DynResource>(&self) -> &T;
+    fn expect_downcast_mut<T: DynResource>(&mut self) -> &mut T;
+
+    /// Unboxes a `Box<dyn DynResource>` to a concrete type.
+    ///
+    /// # Safety
+    /// - The `Box<dyn DynResource>` must be the correct concrete type.
```
ErichDonGubler: style: newline b/w header and list plz

suggestion: I think it's clearer to simply refer to the argument `self` as such; the description of the trait does a pretty good job of making it clear that it's intended to be implemented for `Box<dyn DynResource>`s.

Combined suggestion hunk:

```suggestion
    /// # Safety
    ///
    /// - `self` must be the correct concrete type.
```

wgpu-hal/src/dynamic/mod.rs
```diff
@@ -0,0 +1,67 @@
+use std::any::Any;
+
+use wgt::WasmNotSendSync;
+
+/// Base trait for all resources, allows downcasting via `std::any::Any`.
+pub trait DynResource: Any + WasmNotSendSync + 'static {
```
ErichDonGubler, line 26: question(non-blocking): Why not name this a `Resource`, rather than `DynResource`? We don't need to distinguish from an existing `Resource` API that might compete with the name. I suppose that the `Dyn*` prefix makes this clear that it's a trait intended to be used with `dyn`, though. 🤔

wgpu-hal/src/dynamic/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+            }
+        )*
+    };
+}
+pub(crate) use impl_dyn_resource;
+
+/// Extension trait for `DynResource` used by by implementations of various dynamic resource traits.
+trait DynResourceExt {
+    fn expect_downcast_ref<T: DynResource>(&self) -> &T;
+    fn expect_downcast_mut<T: DynResource>(&mut self) -> &mut T;
```
ErichDonGubler: thought(non-blocking): I'm not sure why, but `expect_*` feels like an odd prefix (which I presume is included to preempt an assumption that this might return an `Option` like with `Any::downcast`).

ErichDonGubler: suggestion: According to [the Rust API guidelines' `C-FAILURE` section](https://rust-lang.github.io/api-guidelines/documentation.html#function-docs-include-error-panic-and-safety-considerations-c-failure), we should document that this will panic if `T` is not the correct type.

```suggestion
    /// # Panics
    ///
    /// - Panics if `self` is not downcastable to `T`.
    fn expect_downcast_ref<T: DynResource>(&self) -> &T;
    /// # Panics
    ///
    /// - Panics if `self` is not downcastable to `T`.
    fn expect_downcast_mut<T: DynResource>(&mut self) -> &mut T;
```

wgpu-hal/src/dynamic/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+    unsafe fn unbox<T: DynResource + 'static>(self: Box<Self>) -> T {
+        debug_assert!(
+            <Self as Any>::type_id(self.as_ref()) == std::any::TypeId::of::<T>(),
+            "Resource doesn't have the expected type, expected {:?}, got {:?}",
+            std::any::TypeId::of::<T>(),
+            <Self as Any>::type_id(self.as_ref())
+        );
+
+        let casted_ptr = Box::into_raw(self).cast::<T>();
+        *unsafe { Box::from_raw(casted_ptr) }
```
ErichDonGubler, line 99: issue: We should add a `SAFETY` comment here that justifies `Box::from_raw` as justified per the contract of this trait method.

wgpu-hal/src/dynamic/mod.rs
ErichDonGubler: Ah, I also suppose that having a family of `Dyn*`-prefixed traits has a sort of naming cohesion. 🤔

nical (approved):
Amazing

wgpu-hal/src/dx12/device.rs
```diff
@@ -1553,10 +1553,10 @@ impl crate::Device for super::Device {
     unsafe fn create_pipeline_cache(
         &self,
         _desc: &crate::PipelineCacheDescriptor<'_>,
-    ) -> Result<(), crate::PipelineCacheError> {
-        Ok(())
+    ) -> Result<super::PipelineCache, crate::PipelineCacheError> {
+        Ok(super::PipelineCache)
     }
-    unsafe fn destroy_pipeline_cache(&self, (): ()) {}
+    unsafe fn destroy_pipeline_cache(&self, _: super::PipelineCache) {}
```
ErichDonGubler, line 1573: question: I didn't understand why the commit titled `introduce DynResource & DynBuffer as first user` introduces new `PipelineCache` and `AccelerationStructure` types. They don't actually seem related: Should these be separate commits?

cwfitzgerald (approved):
I don't see anything that should block this

wgpu-hal/src/dx12/device.rs
ErichDonGubler: Will split myself and push soon. 🙂

ErichDonGubler (commented):
WIP review, here's the feedback I've got so far.

ErichDonGubler:
To be clear: I also see nothing blocking so far. 👏🏻

wgpu-hal/src/dynamic/mod.rs
teoxoy: I'm also not a fan of the `Dyn` prefix. What is the rationale behind it?

Wumpf: Felt like I had to prefix it with _something_ because just distinguishing by namespace gets confusing quickly imho and that's what I came up with 🤷 
open for suggestions!

teoxoy: I haven't looked at the PRs as others are reviewing them but don't these traits represent the full surface of the hal API? If so, I don't think we need a prefix at all.

Wumpf: It's almost the entire surface, yes. I think once the refactor is done they are no longer needed. The thing that I found them useful was probably more transitional!

Wumpf: Ah there's some name clashes though between the generic traits that HAL has. E.g. there's `Device<A>` and `DynDevice` 🤔

wgpu-hal/src/dynamic/mod.rs
Wumpf: yeah, that was the idea. I wanted to encapsulate the `expect` calls because this falls under the "HAL has many assumptions and is inherently unsafe, you have to use it correctly" umbrella

Wumpf: very good point!

wgpu-hal/src/dynamic/mod.rs
teoxoy: Oh so we still have a `Device<A>`? I should just take a look at the code.

Wumpf: oops. just deleted my previous comment because I thought wait this is silly and surely nobody read it yet:
Yes we still have the `Device<A>` baseclass, all this stuff hasn't changed. All individual hal implementors still do the same thing. But that nameclash isn't reaaaaallly a reason not to have this. I mean dx12 also has its own `Device`

Wumpf: We do however force everything on `Api` now to implement (e.g) `DynDevice` and then say that all `Device<A>` blanket implement `DynDevice`

teoxoy: I see, well we can remove the prefix in a follow-up.

wgpu-hal/src/dynamic/mod.rs
ErichDonGubler: Added with latest force-push.

wgpu-hal/src/dynamic/mod.rs
ErichDonGubler: Fixed with latest force-push.

wgpu-hal/src/dynamic/mod.rs
ErichDonGubler: Implemented with latest force-push.

wgpu-hal/src/dynamic/mod.rs
ErichDonGubler: Implemented with latest force-push.

wgpu-hal/src/dynamic/mod.rs
ErichDonGubler: Fixed with latest force-push.

wgpu-hal/src/dx12/device.rs
ErichDonGubler: Implemented with latest force-push. I ended up combining with the commit where this was needed in order to compile.

wgpu-hal/src/dynamic/command.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    unsafe fn set_vertex_buffer<'a>(
+        &mut self,
+        index: u32,
+        binding: BufferBinding<'a, dyn DynBuffer>,
+    ) {
+        let binding = binding.expect_downcast();
+        unsafe { self.set_vertex_buffer(index, binding) };
+    }
+}
```
ErichDonGubler: issue: We don't document a safety contract or adherence thereto. We should do this.

Wumpf:
merged in part 3

wgpu-hal/src/dynamic/command.rs
```diff
@@ -17,6 +38,58 @@ pub trait DynCommandEncoder {
...
+        }
+    }
+
+    unsafe fn end_debug_marker(&mut self) {
+        unsafe {
+            C::end_debug_marker(self);
+        }
+    }
+
     unsafe fn set_index_buffer<'a>(
```
ErichDonGubler, line 548: issue: We don't document a safety contract or adherence thereto for these methods. We should do this.

wgpu-hal/src/dynamic/command.rs
```diff
@@ -1,8 +1,29 @@
-use crate::{BufferBinding, CommandEncoder};
+use crate::{
+    BufferBarrier, BufferBinding, BufferCopy, CommandEncoder, DeviceError, Label, MemoryRange,
+};
 
-use super::DynBuffer;
+use super::{DynBuffer, DynResourceExt as _};
+
+pub trait DynCommandEncoder: std::fmt::Debug {
```
ErichDonGubler: issue: We don't document a safety contract or adherence thereto for these methods. We should do this.

wgpu-hal/src/dynamic/device.rs
```diff
@@ -0,0 +1,13 @@
...
+
+pub trait DynDevice {
+    unsafe fn destroy_buffer(&self, buffer: Box<dyn DynBuffer>);
+}
+
+impl<D: Device> DynDevice for D {
+    unsafe fn destroy_buffer(&self, buffer: Box<dyn DynBuffer>) {
+        unsafe { D::destroy_buffer(self, buffer.unbox()) };
+    }
+}
```
ErichDonGubler: issue: We don't document a safety contract or adherence thereto for these methods. We should do this.

wgpu-hal/src/dynamic/surface.rs
```diff
@@ -0,0 +1,71 @@
...
+#[derive(Debug)]
+pub struct DynAcquiredSurfaceTexture {
+    pub texture: Box<dyn DynSurfaceTexture>,
+    /// The presentation configuration no longer matches
+    /// the surface properties exactly, but can still be used to present
+    /// to the surface successfully.
+    pub suboptimal: bool,
+}
+
+pub trait DynSurface: DynResource {
```
ErichDonGubler, line 17: issue: We don't document a safety contract or adherence thereto for these methods. We should do this.

wgpu-hal/src/dynamic/queue.rs
```diff
@@ -0,0 +1,54 @@
+use crate::{
+    DeviceError, DynCommandBuffer, DynFence, DynResource, DynSurface, DynSurfaceTexture,
+    FenceValue, Queue, SurfaceError,
+};
+
+use super::DynResourceExt as _;
+
+pub trait DynQueue: DynResource {
```
ErichDonGubler, line 8: issue: We don't document a safety contract or adherence thereto for these methods. We should do this.

wgpu-hal/src/dynamic/device.rs
```diff
@@ -1,9 +1,14 @@
+// Box casts are needed, alternative would be a temporaries which are more verbose and not more expressive.
+#![allow(trivial_casts)]
+
```
ErichDonGubler, line 3: suggestion: This is actually not necessary; the offending lines of code can be fixed thusly:

```diff
-            .map(|b| Box::new(b) as Box<dyn DynCommandEncoder>)
+            .map(|b| -> Box<dyn DynCommandEncoder> { Box::new(b) })
```

Will push a fix myself anon.

wgpu-hal/src/dynamic/instance.rs
```diff
@@ -0,0 +1,53 @@
...
+
+pub struct DynExposedAdapter {
+    pub adapter: Box<dyn DynAdapter>,
+    pub info: wgt::AdapterInfo,
+    pub features: wgt::Features,
+    pub capabilities: Capabilities,
+}
+
+pub trait DynInstance: DynResource {
+    unsafe fn create_surface(
```
ErichDonGubler, line 16: issue: We don't document a safety contract or adherence thereto for these methods. We should do this.

ErichDonGubler (commented):
LGTM; really the only issue I noticed is that we don't have any docs for most of this. Buuut 🤷🏻‍♂️ I think it's better to defer to an issue, rather than block here and keep merge conflicts between this and probably most other PRs in total limbo.

PR #6099, opened 2024-08-10
Degenerification part 2/3: Removal of lifetimes from `wgpu-core`

Wumpf:
**Connections**
* Part of https://github.com/gfx-rs/wgpu/issues/5124
* Part of a series!
   * #6098
   * ➡️ #6099 
   *  #6100 

**Description**
This is where the magic happens!
Almost all interaction of `wgpu-hal` in `wgpu-core` use now the previously introduced `Dyn` backend abstraction.
To achieve this, all hal resources are allocated as `Box<dyn hal::DynResource>` (with more concrete types for safety & easy of use).

Since hal resource are no longer distinguish type, `Hubs` collapses to just a single `Hub`.
`AnyDevice` is no longer needed since every device is just a `Box<dyn hal::DynDevice>`

Since `wgpu-core` types are extremely intertwined (most of them have an `Arc<Device>`) this happens in very few commits. However, I split out a few rough iterations with two major "unravelings": first boxing & dyn passing, then hub simplification & actual generic removal.

In order to not have this PR spill into `wgpu`, `gfx_select` has been stripped down to pass-through, it will be removed entirely in Part 3.

**Testing**
* Existing tests run.

**Checklist**

- [x] Run `cargo fmt`.
- ~[ ] Add change to `CHANGELOG.md`. See simple instructions inside file.~ changelog in Part 3


wgpu-core/src/track/mod.rs
```diff
@@ -1,7 +1,7 @@
 /*! Resource State and Lifetime Trackers
 
 These structures are responsible for keeping track of resource state,
-generating barriers where needed, and making sure resources are kept
+generating barriers where needednd making sure resources are kept
```
nical, line 4: nit: looks like the next few comment modifications are accidental.

nical (approved):
Glorious

wgpu-core/src/track/mod.rs
Wumpf: oops, one of my many regex replacements must have gone sideways!

cwfitzgerald (approved):
My eyes glazed over by the 457th file, but this is looking good and nothing that deserves blocking. I could probably find something to bikeshed if I really tried, but lets favor landing.

Wumpf:
merged in part 3

PR #6100, opened 2024-08-10
Degenerification part 3/3: remove `gfx_select`, remove compute/render pass indirections

### LINKED ISSUE #4905: Rustdoc Builds Take 20-40 minutes.
For some reason, starting at #4880, all rustdoc builds started taking _forever_. This reproduces locally, though is fixed on nightly. Going to force rustdoc to nightly in the mean time.

### LINKED ISSUE #5124: Make wgpu-core less generic (brainstorm)

Wumpf:
**Connections**
* Fixes https://github.com/gfx-rs/wgpu/issues/5124
* Fixes #4905
* Part of a series!
   * #6098
   * #6099 
   *  ➡️ #6100 

**Description**
Finishes the series by ripping of various bandaids. Notably the dynamic indirections we had to use on compute & render passes go away again and `wgpu` can hold the pass objects by value again.
This also re-enables doc building on `wgpu-core`!

Please also have a look at the performance report [here](https://gist.github.com/Wumpf/b72b6273c50f38c699b5995a8b5ba2fc).
Note that slight changes have been made since these tests have been run, but the results should be representative.
If time permits, someone with convenient access should run Linux benchmarks as well.

**Testing**
* Existing tests run.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Add change to `CHANGELOG.md`. See simple instructions inside file


cwfitzgerald (approved):

nical (approved):
Sublime

PR #6103, opened 2024-08-12
Resolve lints in versions _newer_ than Rust 1.76 while still in Rust 1.76

ErichDonGubler:
Recommended review experience: commit-by-commit. These commits build individually, and do not need to be squashed.


naga/src/valid/expression.rs
```diff
@@ -1696,7 +1696,7 @@ pub fn check_literal_value(literal: crate::Literal) -> Result<(), LiteralError>
     Ok(())
 }
 
-#[cfg(all(test, feature = "validate"))]
+#[cfg(test)]
 /// Validate a module containing the given expression, expecting an error.
 fn validate_with_expression(
```
Wumpf, line 1701: for reference, seeing this stuff here caused me open https://github.com/gfx-rs/wgpu/issues/6106

but that's for another PR :)

naga/src/valid/expression.rs
```diff
@@ -1795,48 +1793,3 @@ fn f64_const_literals() {
     );
     assert!(result.is_ok());
 }
-
-/// Using I64 in a function's expression arena is forbidden.
-#[cfg(feature = "validate")]
-#[test]
-fn i64_runtime_literals() {
```
cwfitzgerald, line 1802: Why are these tests being removed, as opposed to just un-fearure-gated?

cwfitzgerald (dismissed):

naga/src/valid/expression.rs
ErichDonGubler: The answer is in the relevant commit message:

> chore(naga): remove broken Unsupported64Bit tests
> 
> Seemingly missed in https://github.com/gfx-rs/wgpu/commit/4e6f873da5e25acef9898f0f3490591c8be2e0b6.

This error variant was removed entirely a short while ago. Apparently, `naga`'s CI wasn't catching this! 😬 CC @jimblandy, @teoxoy.

Wumpf (approved):
lgtm!

ErichDonGubler:
Just waiting on @gfx-rs/naga here. I think it's important that they/we are informed, but I don't consider them a necessary blocking reviewer set; will merge in a couple of days if I don't get a response here.

ErichDonGubler:
@Wumpf: Just as a heads-up, I included 2bbfc855 and af81ed5c, which I think are uncontroversial.

ErichDonGubler:
@Wumpf: Also discovered that I could add 316e8450 at this point, rather than a later one. 🥳

teoxoy (approved):

Issue #3756, opened 2023-05-07
Surface lags behind on window resize


nohenry:
**Description**
When resizing the window, the wpgu surface does not keep up with the size of the window. This is possibly a problem with winit instead.

**Repro steps**
Run any example in the wgpu/examples folder

**Expected vs observed behavior**
I expect the surface to resize exactly to the current window dimensions:

https://user-images.githubusercontent.com/50533236/236695650-8513c44c-1d6d-4a2b-938a-3d436d090037.mov

But instead the surface lags behind the current window size:

https://user-images.githubusercontent.com/50533236/236695535-5674eca7-e02c-4d70-a277-1d2b57e86b35.mov

**Platform**
This is running on MacOS Ventura (Metal). Using wgpu on v0.16 git
I've not tested on Windows or Linux.

nohenry:
Found the solution:

```rust
#[cfg(target_os = "macos")]
unsafe {
    surface
        .surface
        .as_hal_mut::<wgpu_hal::api::Metal, _, _>(|surface| {
            if let Some(surface) = surface {
                surface.present_with_transaction = true
            }
        });
}
```

kpreid:
If it's that easy, perhaps this should be available as a _safe_ operation. Reopen this issue?

ifacodes:
> If it's that easy, perhaps this should be available as a _safe_ operation. Reopen this issue?

I agree, I feel like this issue has popped up a few times over the last few versions.

nohenry:
One thing I want to know is if this a problem on different platforms (also how easy it would be to fix for them). I can test on Windows 11 and maybe Linux.

In terms of implementation, would it be best to have a field in the `SurfaceConfiguration` that determines this behaviour? I suppose it depends if this issue arises in the surface for other platforms.

kpreid:
Another question is whether this should even be configurable, or just be the default. In other words: is there some situation where `present_with_transaction` should _not_ be true? (For example, does it reduce performance when _not_ resizing?)

cwfitzgerald:
`present_with_transaction` afaik really limits the amount of frames in flight to under one, so unless you're making a UI app or other thing that needs to be resized all the time, it should be false so that the gpu can run independently.

> One thing I want to know is if this a problem on different platforms (also how easy it would be to fix for them). I

[...]

Issue #249, opened 2019-07-17
How to handle window resizing on macOS


DagAgren:
I am playing around with using wgpu in a macOS project (and eventually iOS, but haven't tried it yet).

Right now I am struggling with how to handle window resizing. I'm currently rendering using CVDisplayLink, and for each frame checking if the size of the window has changed since last frame, and if so, I create a new swap chain.

This is causing double buffer flickering when resizing the window - every other frame is a frozen frame from the double buffer, and ever other is live updating and correct.

If I do not recreate the swap chain at all, it actually looks slightly better: Rendering seems to block completely for short times while resizing, and the contents of the previous frame is just resized instead.

Any clues what is going on? Am I doing something wrong, or is wgpu?

(Also, if I fullscreen, my framerate drops through the floor, and I get occasional squares of corruption on individual frames, but is that the same issue as https://github.com/gfx-rs/wgpu/issues/78 ?)

kvark:
Did you try running https://github.com/gfx-rs/wgpu-rs examples and see how they resize in your environment?

DagAgren:
No, I will try that.

DagAgren:
Sorry, I am not very familiar with rust, so, how do I actually build the examples? "cargo build" does not seem to do it?

kvark:
@DagAgren no worries, we'd be happy to walk you through. Sorry about inconvenience!
Basically, if you install Rust+Cargo (e.g. via https://rustup.rs/), you can do:
```bash
git clone https://github.com/gfx-rs/wgpu-rs
cd wgpu-rs
cargo run --example cube --features metal
```

DagAgren:
Thanks!

Resizing is very broken in a different way in the examples: The animation freezes, and the last frame is stretched to the window size.

DagAgren:
The examples also have the same problem where going to fullscreen will kill the framerate. The animation to fullscreen is of course also frozen the same way.

kvark:
Thank you for verifying! One of the things we are currently doing is stalling the CPU to wait for GPU on resizing: https://github.com/gfx-rs/wgpu/blob/master/wgpu-native/src/device.rs

There is also a number of issues with have with swapchain in gfx-rs Metal backend, including that fullscreen bug. We are looking for ways to resolve them, e.g. https://github.com/gfx-rs/gfx/pull/2882

DagAgren:
[...]

PR #6107, opened 2024-08-12
[metal] Improve layer initialization and resizing

### LINKED ISSUE #1168: Resizing on macOS is cursed again
**Description**
When resizing the window, the previous frame is stretched before any new frames are displayed. Or... maybe a new frame is drawn using an outdated uniform buffer and I can't update it in time. Who knows...

**Repro steps**
https://github.com/LoganDark/wgpu-stretch

The square in the top left should not stretch

**Extra materials**
wgpu stretch 2

The previous incarnation of this bug can be found at #249

**Platform**
macOS 10.14.6 Mojave
Tech stack is as described in the wgpu-stretch repository

madsmtm:
## Description
Overriding the `layer` on `NSView`, as is currently done, makes the view "layer-hosting", see [`wantsLayer`](https://developer.apple.com/documentation/appkit/nsview/1483695-wantslayer?language=objc).

This prevents Winit from emitting the `RedrawRequested` event at the desired frame, because overwriting the layer implicitly disables the [`drawRect:`](https://developer.apple.com/documentation/appkit/nsview/1483686-drawrect?language=objc)/[`updateLayer`](https://developer.apple.com/documentation/appkit/nsview/1483580-updatelayer?language=objc) mechanisms that we need to use. Instead, Winit is currently forced to emit the event after all other events have been processed, which is hacky, and does not integrate well with redraws.

In this PR, I've changed it so that `wgpu`, when it is given a view without a `CAMetalLayer`, instead creates a new sublayer that it renders into (as is already done on iOS).

Note that this might theoretically have a slight performance impact, since the system now has to do more state-tracking, but I suspect it's miniscule (wasn't able to measure it with `bunnymark`), and I'd argue that it's a small price to pay for correctness.

I guess that another solution would be to properly implement the `CALayerDelegate`, and call `drawRect:`/`updateLayer` in there ourselves - though that's a can of worms I'm not really prepared to open.

## Connections
Fixes https://github.com/gfx-rs/wgpu/issues/1168, mostly by using `kCAGravityResize`, but the situation will also be properly fixed in the future, since Winit can clean up its current redrawing hacks.

Related similar issues: https://github.com/gfx-rs/wgpu/issues/249, https://github.com/gfx-rs/wgpu-rs/issues/536, https://github.com/rust-windowing/glutin/issues/1340, https://github.com/rust-windowing/winit/pull/1901 and https://github.com/rust-windowing/winit/issues/2640.

Also makes the situation better for https://github.com/gfx-rs/wgpu/issues/3756, if the user enables `present_with_transaction`, resizing becomes perfectly smooth.

Finally, it removes our dependence on having access to `NSView`, paving the way for [a future where `raw-window-handle` may only provide `CALayer`](https://github.com/rust-windowing/raw-window-handle/issues/123#issuecomment-2308505694).

## Testing
See [this repo](https://github.com/madsmtm/wgpu-objc2-example) which contains roughly the "minimal" setup needed to get `wgpu` working on macOS and iOS without Winit. The crate renders a triangle whose top corner is at a fixed location from the right edge of the window, which is useful for testing that redrawing works as it should. It also has a few different features that you can try out to see how things look using some of the different methods for rendering that macOS/iOS provides. Use the `patch` key in `Cargo.toml` to test it with and without this PR.

I've also tested this with Winit on macOS and Mac Catalyst, by resizing the `hello_triangle` window, and moving it between an external monitors.

### Current behaviour:

https://github.com/user-attachments/assets/2cf9b55c-ea1f-4b9e-9e90-34c70b7176fc

### With this PR:

https://github.com/user-attachments/assets/2588ce7e-375b-4fb9-b8f9-3caccbe8138a

## Checklist

- [x] Run `cargo fmt`.

## TODO (though can still be reviewed in the meantime)
- [x] Test drawing in non-fullwindow views (and maybe scaled / transformed / rotated / weird views?).
  - Tested, it's broken, see https://github.com/gfx-rs/wgpu/pull/6107#issuecomment-2319599762
  - But will be fixed by using `raw-window-metal`, or by using .
- [x] Test passing a MTKView to `wgpu`.
  - See [sample repo](https://github.com/madsmtm/wgpu-objc2-example), pass `--features=mtkview`.
- [x] Test with SDL and other such "window providers" in the Rust ecosystem? I'll need some help with this!
  - `sdl2`'s [example](https://github.com/Rust-SDL2/rust-sdl2/blob/400e033411b239de10b680f20de800dcb70e9ade/examples/raw-window-handle-with-wgpu/main.rs) seems to work fine (they're using a custom view with a `CAMetalLayer` as the root layer, so shouldn't be affected by most of this).

### DIFF
--- a/wgpu-hal/src/metal/surface.rs
+++ b/wgpu-hal/src/metal/surface.rs
@@ -24,7 +24,7 @@ use parking_lot::{Mutex, RwLock};
 #[link(name = "QuartzCore", kind = "framework")]
 extern "C" {
     #[allow(non_upper_case_globals)]
-    static kCAGravityTopLeft: *mut Object;
+    static kCAGravityResize: *mut Object;
 }
 
 extern "C" fn layer_should_inherit_contents_scale_from_window(
@@ -239,7 +239,15 @@ impl super::Surface {
             let frame: CGRect = msg_send![root_layer, bounds];
             let () = msg_send![new_layer, setFrame: frame];
 
-            let _: () = msg_send![new_layer, setContentsGravity: unsafe { kCAGravityTopLeft }];
+            // The desired content gravity is `kCAGravityResize`, because it
+            // masks / alleviates issues with resizing when
+            // `present_with_transaction` is disabled, and behaves better when
+            // moving the window between monitors.
+            //
+            // Unfortunately, it also makes it harder to see changes to
+            // `width` and `height` in `configure`. When debugging resize
+            // issues, swap this for `kCAGravityTopLeft` instead.
+            let _: () = msg_send![new_layer, setContentsGravity: unsafe { kCAGravityResize }];
 
             // Set initial scale factor of the layer. This is kept in sync by
             // `configure` (on UIKit), and the delegate below (on AppKit).
--- a/wgpu-hal/src/metal/surface.rs
+++ b/wgpu-hal/src/metal/surface.rs
@@ -1,6 +1,7 @@
 #![allow(clippy::let_unit_value)] // `let () =` being used to constrain result type
 
 use std::ffi::c_uint;
+use std::mem::ManuallyDrop;
 use std::ptr::NonNull;
 use std::sync::Once;
 use std::thread;
@@ -14,7 +15,7 @@ use objc::{
     class,
     declare::ClassDecl,
     msg_send,
-    rc::autoreleasepool,
+    rc::{autoreleasepool, StrongPtr},
     runtime::{Class, Object, Sel, BOOL, NO, YES},
     sel, sel_impl,
 };
@@ -80,7 +81,9 @@ impl super::Surface {
         delegate: Option<&HalManagedMetalLayerDelegate>,
     ) -> Self {
         let layer = unsafe { Self::get_metal_layer(view, delegate) };
-        // SAFETY: The layer is an initialized instance of `CAMetalLayer`.
+        let layer = ManuallyDrop::new(layer);
+        // SAFETY: The layer is an initialized instance of `CAMetalLayer`, and
+        // we transfer the retain count to `MetalLayer` using `ManuallyDrop`.
         let layer = unsafe { metal::MetalLayer::from_ptr(layer.cast()) };
         let view: *mut Object = msg_send![view.as_ptr(), retain];
         let view = NonNull::new(view).expect("retain should return the same object");
@@ -107,7 +110,7 @@ impl super::Surface {
     pub(crate) unsafe fn get_metal_layer(
         view: NonNull<Object>,
         delegate: Option<&HalManagedMetalLayerDelegate>,
-    ) -> *mut Object {
+    ) -> StrongPtr {
         let is_main_thread: BOOL = msg_send![class!(NSThread), isMainThread];
         if is_main_thread == NO {
             panic!("get_metal_layer cannot be called in non-ui thread.");
@@ -141,7 +144,7 @@ impl super::Surface {
             // render directly into that; after all, the user passed a view
             // with an explicit Metal layer to us, so this is very likely what
             // they expect us to do.
-            root_layer
+            unsafe { StrongPtr::retain(root_layer) }
         } else {
             // The view does not have a `CAMetalLayer` as the root layer (this
             // is the default for most views).
@@ -246,7 +249,7 @@ impl super::Surface {
             if let Some(delegate) = delegate {
                 let () = msg_send![new_layer, setDelegate: delegate.0];
             }
-            new_layer
+            unsafe { StrongPtr::new(new_layer) }
         }
     }
 
--- a/wgpu-hal/src/metal/surface.rs
+++ b/wgpu-hal/src/metal/surface.rs
@@ -1,11 +1,15 @@
 #![allow(clippy::let_unit_value)] // `let () =` being used to constrain result type
 
-use std::{os::raw::c_void, ptr::NonNull, sync::Once, thread};
+use std::ffi::c_uint;
+use std::ptr::NonNull;
+use std::sync::Once;
+use std::thread;
 
 use core_graphics_types::{
     base::CGFloat,
     geometry::{CGRect, CGSize},
 };
+use metal::foreign_types::ForeignType;
 use objc::{
     class,
     declare::ClassDecl,
@@ -16,7 +20,6 @@ use objc::{
 };
 use parking_lot::{Mutex, RwLock};
 
-#[cfg(target_os = "macos")]
 #[link(name = "QuartzCore", kind = "framework")]
 extern "C" {
     #[allow(non_upper_case_globals)]
@@ -46,6 +49,7 @@ impl HalManagedMetalLayerDelegate {
             type Fun = extern "C" fn(&Class, Sel, *mut Object, CGFloat, *mut Object) -> BOOL;
             let mut decl = ClassDecl::new(&class_name, class!(NSObject)).unwrap();
             unsafe {
+                // <https://developer.apple.com/documentation/appkit/nsviewlayercontentscaledelegate/3005294-layer?language=objc>
                 decl.add_class_method::<Fun>(
                     sel!(layer:shouldInheritContentsScale:fromWindow:),
                     layer_should_inherit_contents_scale_from_window,
@@ -72,26 +76,15 @@ impl super::Surface {
     /// If not called on the main thread, this will panic.
     #[allow(clippy::transmute_ptr_to_ref)]
     pub unsafe fn from_view(
-        view: *mut c_void,
+        view: NonNull<Object>,
         delegate: Option<&HalManagedMetalLayerDelegate>,
     ) -> Self {
-        let view = view.cast::<Object>();
-        let render_layer = {
-            let layer = unsafe { Self::get_metal_layer(view, delegate) };
-            let layer = layer.cast::<metal::MetalLayerRef>();
-            // SAFETY: This pointer…
-            //
-            // - …is properly aligned.
-            // - …is dereferenceable to a `MetalLayerRef` as an invariant of the `metal`
-            //   field.
-            // - …points to an _initialized_ `MetalLayerRef`.
-            // - …is only ever aliased via an immutable reference that lives within this
-            //   lexical scope.
-            unsafe { &*layer }
-        }
-        .to_owned();
-        let _: *mut c_void = msg_send![view, retain];
-        Self::new(NonNull::new(view), render_layer)
+        let layer = unsafe { Self::get_metal_layer(view, delegate) };
+        // SAFETY: The layer is an initialized instance of `CAMetalLayer`.
+        let layer = unsafe { metal::MetalLayer::from_ptr(layer.cast()) };
+        let view: *mut Object = msg_send![view.as_ptr(), retain];
+        let view = NonNull::new(view).expect("retain should return the same object");
+        Self::new(Some(view), layer)
     }
 
     pub unsafe fn from_layer(layer: &metal::MetalLayerRef) -> Self {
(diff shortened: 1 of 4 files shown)


madsmtm:
Another solution than this would be to make `wgpu` handle when to render internally (e.g. provide a "render" callback). Not sure how that'd work on other platforms though, which is why I've gone this route instead.

wgpu-hal/src/metal/surface.rs
```diff
@@ -1,11 +1,17 @@
...
+use std::sync::Once;
+use std::thread;
 
 use core_graphics_types::{
     base::CGFloat,
     geometry::{CGRect, CGSize},
 };
+use metal::foreign_types::ForeignType;
+use objc::rc::StrongPtr;
 use objc::{
```
MarijnS95: Merge? There's `rc::releasepool` down below.

Wumpf:
I'm a bit short on time right now, but I'll try to have a deeper look later this week! Looks really well researched and documented, looking forward to it :)

wgpu-types/src/lib.rs
```diff
@@ -5463,8 +5463,16 @@ pub struct SurfaceConfiguration<V> {
     /// `Bgra8Unorm` and `Bgra8UnormSrgb`
     pub format: TextureFormat,
     /// Width of the swap chain. Must be the same size as the surface, and nonzero.
+    ///
+    /// If this is not the same size as the underlying surface (e.g. if it is
+    /// set once, and the window is later resized), the behaviour is defined
+    /// but platform-specific, and may change in the future.
```
Wumpf: I'm actually leaning towards going even further and start listing out the status quo per platform. Drawback of that is ofc that it crosslinks the docs with a piece of independent code and may thus easily outdate, also we'd be starting with just MacOS which just looks a bit bad.
It is not too unusual for us though to point out behavior of specific platforms in here since the benefits (user not having to dig deeply) outweigh the cost (maintenance, risk of outdated information). Usually we do so indirectly by referring to features and abilities, but that's not possible here obviously.

This is an improvement regardless, so I'd leave it to your judgement how much we want to leak the implementation details here.

wgpu-hal/src/metal/surface.rs
```diff
@@ -102,56 +98,167 @@ impl super::Surface {
...
+            // Note that there is a somewhat hidden design decision in this:
+            // We define the `width` and `height` in `configure` to control
+            // the `drawableSize` of the layer, while `bounds` and `frame` are
+            // outside of the user's direct control - instead, though, they
+            // can control the size of the view (or root layer), and get the
+            // desired effect that way.
+            //
+            // We _could_ also let `configure` set the `bounds` size, however
+            // that would be inconsistent with using the root layer directly
+            // (as we may do, see above).
```
Wumpf, line 222: clarifying question: if we _were_ to pass bounds we would have to create a sublayer even if there's already a metal layer as root layer, right? (which ofc might be rather unintuitive)

wgpu-hal/src/metal/surface.rs
```diff
@@ -102,56 +98,167 @@ impl super::Surface {
...
-            };
+
+            // The desired content gravity is `kCAGravityResize`, because it
+            // masks / alleviates issues with resizing when
+            // `present_with_transaction` is disabled, and behaves better when
+            // moving the window between monitors.
+            //
+            // Unfortunately, it also makes it harder to see changes to
+            // `width` and `height` in `configure`. When debugging resize
+            // issues, swap this for `kCAGravityTopLeft` instead.
```
Wumpf, line 243: seems like a minor concern in general but someone debugging this or needing more control may always provide their own readily configured metal layer. Appreciate the note though!

Wumpf (approved):
Excellent! This documents what's going on and all the considerations going on so much better, with the old code (not having read it before) I had no clue what was _actually_ going on.

I tested it on an M1 Sonoma 14.4.1 with a low-dpi screen attached and it performs as advertised in the PR description 👌 

Can land as-is, but I think @cwfitzgerald wanted to have look as well (?)
Regarding the todos left on the PR description: I see no problem in catching up on this later. Unfortunately I also don't have any of these slightly more unusual cases ready for testing

wgpu-hal/src/metal/surface.rs
madsmtm: I had trouble explaining my reasoning here, so clarifying questions are good!

> if we _were_ to pass bounds we would have to create a sublayer even if there's already a metal layer as root layer, right?

Not _necessarily_, we could track in a boolean whether we created the layer ourselves or not, and only update the bounds in `configure` if we created the layer. Or we could define that "the bounds of the layer that the user passed us is under our control". But I felt that overall, this was the cleanest solution, and `drawableSize` matched well with the name `Surface::configure` ("drawable" matches "surface").

If this part of my comments can be rewritten to make things clearer, please suggest how to do so!

wgpu-types/src/lib.rs
madsmtm: I felt like it was a lot of detail to put into a fairly niche area (not resizing when the documentation says you should do) - it gives it undue "weight" compared to the rest of the documentation, almost as-if we support such use-cases where I really don't want to - but that may be just a feeling, a small note like the following would probably be fine:

```suggestion
    /// set once, and the window is later resized), the behaviour is defined
    /// but platform-specific, and may change in the future (currently macOS
    /// scales the surface, other platforms may do something else).
```

Wumpf: Not a strong stance, but I would prefer the small note you propose :) 👍

madsmtm:
So, I've run into a macOS bug that has been driving me _insane_, because it took me so long to figure out that it's actually that, an OS bug, and not because I was doing something wrong. It happens when setting an `autoresizingMask` to something that auto-sizes (like we do with `kCALayerWidthSizable | kCALayerHeightSizable`), using a `contentsGravity` that auto-resizes (like we do with `kCAGravityResize`), and using auto layout on the owning view. I believe that it is an OS bug because I can reproduce it in a virtual machine on macOS 14.6 and on macOS 15 beta 3, but not on macOS 13.5.

I am fairly sure that what's happening is that AppKit's auto layout algorithm, when calculating the size on the containing `NSView`, sets the same size on the inner `CALayer`, and this size happens to contain a fractional value on high DPI screens (on macOS 13 and below, the auto layout algorithm did not create fractional view sizes, so it is not a problem there). Layers do not like fractional values in their width/height (documented [here](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CoreAnimation_guide/BuildingaLayerHierarchy/BuildingaLayerHierarchy.html#//apple_ref/doc/uid/TP40004514-CH6-SW5), though unknown _why_ this is the case), and so the auto-resize mechanism ends up rounding these values to not be fractional. This, in turn, causes there to be a difference between the layer's bounds, and the superlayer's bounds, and `kCAGravityResize` interprets it as-if we we want it to _maintain_ this difference, and then things blow up.

This is a problem when using Wgpu inside more complicated layouts, since the bounds of the layer will slowly drift out of sync with reality the more you resize, as can be seen in [this repo](https://github.com/madsmtm/wgpu-objc2-example) by using the `"two-triangles"` Cargo feature, or in the video below:

https://github.com/user-attachments/assets/e13a67db-ab21-41ab-b7cd-367acb12edb8

I would argue this is not a blocker for this PR, as it still improves things so far, but it's something that I'll try to work on resolving in any case.

scoopr:
Hm, it looks like I was less familiar with the AppKit view and layer management than I thought.

By not making it layer-hosting, does that mean that the layer should be _only_ updated in the `drawRect:`/`updateLayer`? 
I'm just slightly concerned that there is no undue restrictions when doing it like that. I'm thinking of maybe a case of using another thread to do `nextDrawable` in a custom loop, to do some fancy framepacing tricks (or maybe slow rendering without interfering other UI). Its just a bit unclear from the docs that what is required or allowed for layer-backed vs layer-hosted in the particular case of `CAMetalLayer`, though I bet it skips most of AppKit requirements.

I'm fairly sure I'm just paranoid but wanted to air out my concerns.

madsmtm:
> By not making it layer-hosting, does that mean that the layer should be _only_ updated in the `drawRect:`/`updateLayer`?

No no, layers can be updated safely anywhere, from any thread, regardless of whether the view they're contained in is layer-hosting or just layer-backed. Only issue is that you will get tearing and such unless you use Wgpu's `present_with_transaction`, or restrict your drawing inside `drawRect:`/`updateLayer`.

The difference between layer-hosting and layer-backed is only in terms of how the `NSView` talks to its `CALayer` - and with layer-hosting views, `NSView` functionality that Winit uses (i.e. `drawRect:`/`updateLayer`) is disabled, which is the primary motivation for this PR's changes.

Does that make sense?

madsmtm:
I've toiled further away on this, and I I've come to the conclusion that the only proper way to fix the issue I talked about in https://github.com/gfx-rs/wgpu/pull/6107#issuecomment-2319599762 is to use observers, that is, make the new layer that we create observe the `contentsScale` and `bounds` of the super layer, and update it's properties when those change.

You can see implementations of this idea https://github.com/rust-windowing/raw-window-metal/pull/19 and https://github.com/rust-windowing/softbuffer/pull/234, it's fairly tricky stuff since the Key-Value-Observing APIs are just very terrible outside of Objective-C.

I then tried to actually implement this in Wgpu, but since it's [still using the `objc` crate](https://github.com/gfx-rs/wgpu/pull/5641), I had to resort to all sorts of hacks to get it to work. In the end, I threw the code away; it's doable, but it's not pretty, and I would rather have Wgpu depend on `raw-window-metal` so that we can maintain this logic in a central place. That's a discussion for [a separate PR though](https://github.com/gfx-rs/wgpu/pull/6210), I think this PR is good to go now.

Wumpf:
sounds very reasonable to me. Thanks again for documenting everything so well both in comments as well as the thought process here on the PR!

jimblandy:
@madsmtm Thank you so much for the comments in `wgpu_hal::metal::Surface::get_metal_layer`! They're very helpful.

polina4096:
Not sure if anyone has noticed, yet for me it's quite obvious even on the video demonstration provided in the PR that after the fix the resizing is more jittery and laggy. It does fix the white blanks at the edges, yet introduces some delay.

I've spent about an hour just to find what was causing this undesirable behavior after updating dependencies in my personal project. I think that it's important to emphasize that the laggy resizing introduced by this PR is not what some users might want.

madsmtm:
> Not sure if anyone has noticed, yet for me it's quite obvious even on the video demonstration provided in the PR that after the fix the resizing is more jittery and laggy. It does fix the white blanks at the edges, yet introduces some delay.
> 
> I've spent about an hour just to find what was causing this undesirable behavior after updating dependencies in my personal project. I think that it's important to emphasize that the laggy resizing introduced by this PR is not what some users might want.

Hmm, that's probably because we're using `kCAGravityResize`, which might be forcing the system to do more buffering or something? I'm unsure!

Could I get you to try https://github.com/gfx-rs/wgpu/pull/6210? EDIT: Nvmd, [we raced](https://github.com/gfx-rs/wgpu/pull/6210#issuecomment-2499236366).

And separately, also try the following diff:
```diff
diff --git a/wgpu-hal/src/metal/surface.rs b/wgpu-hal/src/metal/surface.rs
index b35c73c91..5f0fda802 100644
--- a/wgpu-hal/src/metal/surface.rs
+++ b/wgpu-hal/src/metal/surface.rs
@@ -23,7 +23,7 @@ use parking_lot::{Mutex, RwLock};
 #[link(name = "QuartzCore", kind = "framework")]
 extern "C" {
     #[allow(non_upper_case_globals)]
-    static kCAGravityResize: *mut Object;
+    static kCAGravityTopLeft: *mut Object;
 }
 
 extern "C" fn layer_should_inherit_contents_scale_from_window(
@@ -243,7 +243,7 @@ impl super::Surface {
             // Unfortunately, it also makes it harder to see changes to
             // `width` and `height` in `configure`. When debugging resize
             // issues, swap this for `kCAGravityTopLeft` instead.
-            let _: () = msg_send![new_layer, setContentsGravity: unsafe { kCAGravityResize }];
+            let _: () = msg_send![new_layer, setContentsGravity: unsafe { kCAGravityTopLeft }];
 
             // Set initial scale factor of the layer. This is kept in sync by
             // `configure` (on UIKit), and the delegate below (on AppKit).
```

polina4096:
After applying the patch the result seems worse. Noticeable lag remains and now the window sometimes flashes at the top.


https://github.com/user-attachments/assets/1b6b9971-a7ea-4e0b-a08a-7e0771df5b94

madsmtm:
> After applying the patch the result seems worse

Huh!

Could you try with the following (should effectively reverse this PR), possibly in combination with the diff before:
```diff
diff --git a/wgpu-hal/src/metal/surface.rs b/wgpu-hal/src/metal/surface.rs
index b35c73c91..854b6deff 100644
--- a/wgpu-hal/src/metal/surface.rs
+++ b/wgpu-hal/src/metal/surface.rs
@@ -204,10 +204,10 @@ impl super::Surface {
 
             // Create a new sublayer.
             let new_layer: *mut Object = msg_send![class!(CAMetalLayer), new];
-            let () = msg_send![root_layer, addSublayer: new_layer];
+            let () = msg_send![view.as_ptr(), setLayer: new_layer];
 
             #[cfg(target_os = "macos")]
-            {
+            if false {
                 // Automatically resize the sublayer's frame to match the
                 // superlayer's bounds.
                 //
```

> for me it's quite obvious even on the video demonstration

Still not entirely sure I see the issue btw, could you explain further what you mean? As in, how would I verify whether the issue is present or not? The distance between the window corner and the cursor?

polina4096:
Wow, this seems like the perfect solution. Applying this patch atop of master gave the best results: no delay/lag when resizing and also no blanks/artifacts. However, when applied with the previous patch you sent the window has some artifacts at the top when resizing.

To determine whether the problem is still present, pay close attention to the triangle edges, or the window border. In the case of videos in the PR, it looks like the window itself is being resized at a "lower framerate". As for the triangle, its edges are updated to match the new size at a slower rate than in the other video.

To reiterate, applying the most recent patch atop of master fixes everything.

madsmtm:
Okay, so I think I see what you mean (and I think the effect is more visible when you actually do it, than in the videos).

But I couldn't see the effect when using https://github.com/gfx-rs/wgpu/pull/6210?

Tested with `cargo run --bin wgpu-examples -- hello_triangle` (no optimizations, though I found the same when enabling optimizations).

Current `trunk` (3016c56db674fb39e79327b52e85105c1020e766):

https://github.com/user-attachments/assets/a615dca4-5422-49f2-9d90-479e506a8c65

With [this patch](https://github.com/gfx-rs/wgpu/pull/6107#issuecomment-2499760574) applied:

https://github.com/user-attachments/assets/44f02b92-4685-4195-8bf3-3444e0c7c511

After https://github.com/gfx-rs/wgpu/pull/6210:

https://github.com/user-attachments/assets/c214394a-9206-4ab8-aafa-ac262987cb19

polina4096:
After testing it again with `cargo run --bin wgpu-examples -- hello_triangle` I concluded that patched and https://github.com/gfx-rs/wgpu/pull/6210 looks almost exactly the same. The main difference arises when the app constantly draws to the screen.

I've also tested it in my application which schedules redraws according to DisplayLink and https://github.com/gfx-rs/wgpu/pull/6210 appears to have undesirable lag when resizing, yet after the patch https://github.com/gfx-rs/wgpu/pull/6107#issuecomment-2499760574 applied to master it works flawlessly.

polina4096:
I'm not an expert in terms of how the rendering should work on macOS. However, through a lot of testing it seems that the Display Link approach minimizes the latency for applications that must redraw in sync with the display refresh rate, such as games. There may be a better approach available that I'm not aware of.

madsmtm:
You are right that Display Link is the "best" approach to rendering on macOS (with probably a bunch of caveats there), but it's not really a viable approach for Wgpu + Winit because it's not cross-platform.

I'd be interested in seeing your code, I really would've thought that #6210 wouldn't have such issues - unless the issue just is that it's more expensive to register that observer than I thought, and the cost ends up causing delays in your rendering code? I'd be interested in seeing a flamegraph of the two!

polina4096:
@madsmtm, I've pushed the code to https://github.com/polina4096/wgpu-pr6107-demo. There are two binaries, one is using the patched (https://github.com/gfx-rs/wgpu/pull/6107#issuecomment-2499760574) trunk, other is using #6210. For both, the rendering is scheduled using Display Link.

PR #6114, opened 2024-08-14
Make context current when dropping `glow::Context` to allow for debug callback cleanup

Imberflur:
**Description**

Previously when running tests via `cargo xtask test`, I noticed some of the tests randomly segfaulting on opengl. This was due to the debug message callback running after its state was freed when `glow::Context` is dropped (i.e. an instance of https://github.com/grovesNL/glow/issues/270).

The latest version of `glow` fixes this by clearing the debug message callback in its `Drop` impl. This requires the context to be current when dropping `glow::Context` (which is now documented in https://docs.rs/glow/0.14.0/glow/trait.HasContext.html#safety).

This PR adds some logic to make the context current when the `glow::Context` will be dropped and utilizes `ManuallyDrop` to control specifically when the drop occurs as well as preventing accidental drops from unwinding.

Presumably the context is always current for `gles/web.rs`?

**Testing**

Running `cargo xtask test` on a Linux machine.

Unfortunately in the latest version of `wgpu`, I'm not able to reproduce this segfault by running the tests, probably due to timing changes. So it is hard to confirm that this is fixing things. But it should theoretically be correct and my previous experimentation showed this fixing the bug.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-hal/src/gles/egl.rs
+++ b/wgpu-hal/src/gles/egl.rs
@@ -1,8 +1,10 @@
 use glow::HasContext;
 use once_cell::sync::Lazy;
-use parking_lot::{Mutex, MutexGuard, RwLock};
+use parking_lot::{MappedMutexGuard, Mutex, MutexGuard, RwLock};
 
-use std::{collections::HashMap, ffi, os::raw, ptr, rc::Rc, sync::Arc, time::Duration};
+use std::{
+    collections::HashMap, ffi, mem::ManuallyDrop, os::raw, ptr, rc::Rc, sync::Arc, time::Duration,
+};
 
 /// The amount of time to wait while trying to obtain a lock to the adapter context
 const CONTEXT_LOCK_TIMEOUT_SECS: u64 = 1;
@@ -295,6 +297,7 @@ impl EglContext {
             .make_current(self.display, self.pbuffer, self.pbuffer, Some(self.raw))
             .unwrap();
     }
+
     fn unmake_current(&self) {
         self.instance
             .make_current(self.display, None, None, None)
@@ -305,7 +308,7 @@ impl EglContext {
 /// A wrapper around a [`glow::Context`] and the required EGL context that uses locking to guarantee
 /// exclusive access when shared with multiple threads.
 pub struct AdapterContext {
-    glow: Mutex<glow::Context>,
+    glow: Mutex<ManuallyDrop<glow::Context>>,
     egl: Option<EglContext>,
 }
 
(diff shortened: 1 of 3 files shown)


wgpu-hal/src/gles/egl.rs
```diff
@@ -1073,13 +1100,15 @@ impl super::Adapter {
     /// - The underlying OpenGL ES context must be current.
     /// - The underlying OpenGL ES context must be current when interfacing with any objects returned by
     ///   wgpu-hal from this adapter.
+    /// - The underlying OpenGL ES context must be current when dropping this adapter and when
+    ///   dropping any objects returned from this adapter.
```
Imberflur, line 1109: This new point might be covered by "interfacing"? I want to make sure it is clear though.

wgpu-hal/src/gles/egl.rs
```diff
@@ -346,14 +349,34 @@ impl AdapterContext {
...
+        // requires the context to be current when anything that may be holding
+        // the `Arc<AdapterShared>` is dropped.
+        self.egl.as_ref().map(|egl| {
+            egl.make_current();
+        });
+        let glow = self.glow.get_mut();
+        // SAFETY: Field not used after this.
+        unsafe { ManuallyDrop::drop(glow) };
+        self.egl.as_ref().map(|egl| {
+            egl.unmake_current();
```
Imberflur: I just realized that if the drop panics for some reason then we won't call `unmake_current`, might want to adjust this to use a guard to call `unmake_current`?

wgpu-hal/src/gles/egl.rs
cwfitzgerald: More docs > less docs :)

cwfitzgerald (approved):
Looks good to me!

cwfitzgerald:
> Presumably the context is always current for gles/web.rs?

Yeah there's no concept of non-current contexts

wgpu-hal/src/gles/egl.rs
```diff
@@ -1052,6 +1082,8 @@ impl crate::Instance for Instance {
             unsafe { gl.debug_message_callback(super::gl_debug_message_callback) };
         }
 
+        // Avoid accidental drop when the context is not current.
+        let gl = ManuallyDrop::new(gl);
```
MarijnS95, line 1086: Is this a stale comment? You're not avoiding a drop when the context is _not current_, instead you're _forcing it to be current_ Before dropping.

Imberflur: This comment is noting the reason to wrap in `ManuallyDrop` before calling `unmake_current` on the line below. This prevents the `glow::Context` from being dropped if there is some panic between when we call `unmake_current` and when `gl` is placed into the `AdapterContext` (where the drop impl will handle making the context current before dropping `glow::Context`).

FWIW the comment could probably be clarified.

MarijnS95: Clarification is happening in https://github.com/gfx-rs/wgpu/pull/6152 where I ran into a few conflicts with this PR. Panics (except inside the drop handler) [should still cause `Drop` to be called while unwinding](https://doc.rust-lang.org/nomicon/unwinding.html). I'll add a mention that the only skipped case is between `let gl` and when `Self {}` is finally constructed.

Fwiw when a debug callback is specified by the user (in this case `wgpu`), it's common to skip messages when panicking:

https://github.com/gfx-rs/wgpu/blob/5deaef3b67259008844b15e1fd1d3a8924d1cde0/wgpu-hal/src/vulkan/instance.rs#L21-L23

Not sure if that can be done in EGL/WGL as well? Otherwise it seems this solution is a litle bit overengineered if it's _just_ about making the context current in `drop()`?

Imberflur: > Panics (except inside the drop handler) [should still cause Drop to be called while unwinding](https://doc.rust-lang.org/nomicon/unwinding.html)

Yes, this is why it needs to be wrapped in `ManuallyDrop` in the intermittent time before `AdapterContext` is constructed since panics would drop the `glow::Context` which could lead to UB if it isn't current.

> Fwiw when a debug callback is specified by the user (in this case wgpu), it's common to skip messages when panicking:

Something like this won't help here with the way the `Drop` impl for `glow::Context` works, by the time the callback is invoked there is already UB.

PR #6134, opened 2024-08-20
Remove IDs from wgpu traits

### LINKED ISSUE #6080: `wgpu::Buffer` implementing `PartialEq`, `Eq`, and `Hash`
I wanted to use `wgpu::Buffer` in a `HashSet` so I can cheaply determine which buffers are used in a generated `wgsl` shader before binding them. But you can't create `HashSet<&wgpu::Buffer>` or `HashSet<Rc<wgpu::Buffer>>` because `PartialEq`, `Eq`, and `Hash` are not implemented for the type.

I personally worked around this by following the [newtype pattern](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#using-the-newtype-pattern-to-implement-external-traits-on-external-types), which then also necessitated writing implementations for `Deref` and `DerefMut`, but it would be nice if this was just part of the `wgpu` package and I could drop the wrapper type.

I figure others generating shaders to execute on-the-fly may run into similar issues, which is why I'm opening the issue up for consideration. Because Buffer equality from a shader execution perspective requires that they are actually the same buffer on either side of the comparison, I simply did hashing and equality checking on `wgpu::Buffer::global_id` calls.

I'm willing to write the PR to do this if you agree it's useful, but I figured I shouldn't drive-by PR without getting that approval.

teoxoy:
Another stepping stone towards https://github.com/gfx-rs/wgpu/issues/5121.

The first commit is the result of a lot of regex replaces, the next 3 commits are manual fixes. The PR needs to be squashed but I figured it might be easier to review the steps individually.

Resolves https://github.com/gfx-rs/wgpu/issues/6080.


CHANGELOG.md
```diff
@@ -81,6 +81,7 @@ By @wumpf in [#6069](https://github.com/gfx-rs/wgpu/pull/6069), [#6099](https://
 
 - Reduce the amount of debug and trace logs emitted by wgpu-core and wgpu-hal. By @nical in [#6065](https://github.com/gfx-rs/wgpu/issues/6065)
 - `Rg11b10Float` is renamed to `Rg11b10UFloat`. By @sagudev in [#6108](https://github.com/gfx-rs/wgpu/pull/6108)
+- Removed `.global_id()` getters from wgpu objects. By @teoxoy [#6134](https://github.com/gfx-rs/wgpu/pull/6134)
```
cwfitzgerald: We can't just remove this - it's used externally for caching and stuff. We should generate ids like we do for wasm. Longer term this problem will be solved by our types implementing partialeq/hash

teoxoy: If we now have to generate IDs for both backends, wouldn't it make sense to have users interested in this functionality do this?

I looked at the first 3 pages of dependents on crates io and none of them use the `global_id`.

wgpu types don't implement Clone either, they are unique and so was their ID, so I'm not sure how the ID would be useful for caching.

Looping in @TheOnlyMrCat who was the author of https://github.com/gfx-rs/wgpu/pull/3104.

cwfitzgerald: > If we now have to generate IDs for both backends, wouldn't it make sense to have users interested in this functionality do this?

The problem is that they may not control all creation and deletion code (think middleware). This gives a pretty easy way to determine a wgpu resouece's identity.

teoxoy: I see, what do you think about using the pointers of the internal Boxes to implement `PartialEq`, `Eq`, and `Hash` for the resource types directly? That would also solve https://github.com/gfx-rs/wgpu/issues/6080.

cwfitzgerald: Sorry got called to an appointment in between responses 😛

Yeah that would work, we would need to make sure that all resources had those impls.

teoxoy: While we are at it, I feel like we should reconsider some of the objects implementing this and do it only for those that have use cases.

My proposal:

Yes:

- Buffer
- Texture
- TextureView
- Sampler
- QuerySet

Maybe?:

- BindGroupLayout (this won't be equal with others but will be compatible)
- BindGroup
- ShaderModule
- PipelineLayout
- RenderPipeline
- ComputePipeline
- RenderBundle

No:

- Surface
- Adapter
- Device
- Queue

Thoughts on this? cc @dfellis

dfellis: I agree with your Yes and No.

For the Maybe, the `RenderPipeline` and `ComputePipeline` immediately feel like "yes" to me: They could be useful as the keys to a `HashMap` that caches a `HashSet` of `BindGroup`s that have already been constructed for that operation, letting a bunch of setup work get short-circuited if this particular shader on this particular collection of Buffers/Textures/etc has been executed before. (This reasoning immediately marks `BindGroup` as "yes," as well.)

`BindGroupLayout` is also "yes," I think, because you can share the same bind group layout between pipelines on purpose. It could be useful to check different pipelines' `get_bind_group_layout` return values actually being the same underlying object so you know whether or not you can reuse bind groups between them.

I'm not so familiar with `RenderBundle` (it looks like some sort of hardware-specific optimization for the command encoder?) so I don't know if people would want to cache multiple kinds of these bundles and select one on the fly. I feel like the actual drawing to a screen surface would never change, but maybe if [it's drawing to a buffer](https://docs.rs/wgpu/latest/wgpu/struct.RenderBundleEncoder.html#method.draw_indirect) you could have different ones for different kinds of effects to create the textures or what-not.

teoxoy: I talked a bit with @cwfitzgerald and we decided to just do it for all device resources.

> `BindGroupLayout` is also "yes," I think, because you can share the same bind group layout between pipelines on purpose. It could be useful to check different pipelines' `get_bind_group_layout` return values actually being the same underlying object so you know whether or not you can reuse bind groups between them.

It's worth noting that this isn't and won't be the case, internally we share the same object but at the wgpu level we don't. This is also consistent with the WebGPU backend where you can't tell that 2 bind group layouts point to the same underlying object.

I don't think this is a really big deal, the bind group layouts will still be compatible with each other as per the WebGPU spec.

CHANGELOG.md
```diff
@@ -81,6 +81,8 @@ By @wumpf in [#6069](https://github.com/gfx-rs/wgpu/pull/6069), [#6099](https://
...
+- Implement `PartialEq`, `Eq`, `Hash`, `PartialOrd` and `Ord` for wgpu resources. By @teoxoy [#6134](https://github.com/gfx-rs/wgpu/pull/6134)
```
cwfitzgerald: I think this probably needs to go in a major change bulitin explaining the migration.

cwfitzgerald (approved):
Changelog thing, take or leave

CHANGELOG.md
kpreid: What *is* the migration? How are applications currently using `global_id()` going to get something with equivalent properties (reliably distinguishes resources, but *does not* keep the resource alive) without additional overhead? I use these a lot to ensure resources that depend on other resources (texture views, bind groups) are up-to-date.

Hm, I suppose applications could generate their own IDs and maybe even carry them in a wrapper implementing  `Deref` and get the same results. Still, I suspect something useful is being lost here.

cwfitzgerald: The main problem is after this PR we don't have such an ID anymore, so would need to generate it with an atomic on all platforms. 

If this is indeed required functionality we can absolutely have the wgpu layer do that, though they're might be other ways to solve that problem now that things impl PartialEq, Ord, and Hash

kpreid: Well, I wouldn't want to demand that wgpu produce something it doesn't use, so probably, application-generated IDs (or, in my case, perhaps merely generation numbers) are a reasonable solution for those applications that need them.

(Sorry for jumping in late and complaining a bit.)

cwfitzgerald: Sounds reasonable - thanks for the feedback, part of why I was unsure about this is I didn't exactly know if we were covering the use cases of people.

MarijnS95:
Cool that this _just_ got merged, but it also caused a regression on `as_hal()` it seems (if I didn't rebase before writing that code, it would've been easier to spot!). When I call `device.as_hal::<wgpu::hal::dx12::Api, _, _>(..)` on the output from `adapter.request_device()`, I get a panic of the form:

```
thread 'main' panicked at wgpu\src\context.rs:737:5:
assertion failed: data.is::<T>()
stack backtrace:
...
  16:     0x7ff6b986c4a3 - wgpu::context::downcast_ref<wgpu::backend::wgpu_core::Device>
                               at ...wgpu\wgpu\src\context.rs:737
  17:     0x7ff6b94e36ea - wgpu::api::device::impl$0::as_hal::closure$0<wgpu_hal::dx12::Api,wgpu_examples::hello_triangle::run::async_fn$0::closure_env$0,enum2$<core::result::Result<windows::Win32::Graphics::CompositionSwapchain::IPresentationFactory,windows_result::error::Error> >
                               at ...wgpu\wgpu\src\api\device.rs:444
  18:     0x7ff6b94b7cfe - enum2$<core::option::Option<ref$<wgpu::backend::wgpu_core::ContextWgpuCore> > >::map<ref$<wgpu::backend::wgpu_core::ContextWgpuCore>,enum2$<core::result::Result<windows::Win32::Graphics::CompositionSwapchain::IPresentationFactory,windows_result::error::Err
                               at /rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23\library\core\src\option.rs:1075
  19:     0x7ff6b94e3656 - wgpu::api::device::Device::as_hal<wgpu_hal::dx12::Api,wgpu_examples::hello_triangle::run::async_fn$0::closure_env$0,enum2$<core::result::Result<windows::Win32::Graphics::CompositionSwapchain::IPresentationFactory,windows_result::error::Error> > >
                               at ...wgpu\wgpu\src\api\device.rs:439
  20:     0x7ff6b94cbb63 - wgpu_examples::hello_triangle::run::async_fn$0
                               at ...wgpu\examples\src\hello_triangle\mod.rs:108
  21:     0x7ff6b94ad165 - pollster::block_on<enum2$<wgpu_examples::hello_triangle::run::async_fn_env$0> >
                               at C:\Users\Marijn\.cargo\registry\src\index.crates.io-6f17d22bba15001f\pollster-0.3.0\src\lib.rs:128
  22:     0x7ff6b940eda5 - wgpu_examples::hello_triangle::main
                               at ...wgpu\examples\src\hello_triangle\mod.rs:268
  23:     0x7ff6b93f1723 - wgpu_examples::main
                               at ...wgpu\examples\src\main.rs:229
...
error: process didn't exit successfully: `target\debug\wgpu-examples.exe hello_triangle` (exit code: 101)
```

ErichDonGubler:
@MarijnS95: Filed an issue to track that at <https://github.com/gfx-rs/wgpu/issues/6169>. Deets there plz?

CHANGELOG.md
kpreid: FWIW, here’s what the migration looks like for me, adding a wrapper with IDs to the places it was needed: https://github.com/kpreid/all-is-cubes/commit/87fb899aa0ef22489211257c4c851a3a7b9eed58. I tested it against wgpu trunk.

PR #6138, opened 2024-08-20
Dealing with data too large for a single buffer

alphastrata:
**Connections**
[discussion](https://github.com/gfx-rs/wgpu/discussions/6130)
[thread on matrix](https://matrix.to/#/!MFogdGJfnZLrDmgkBN:matrix.org/$-fQkQjOXWbzzJcEUvm9DP1maq1tub8kl6K2l7FapsRc?via=matrix.org&via=mozilla.org&via=kit.edu)

**Description**
The aim of this new example is to demonstrate taking a large input dataset, splitting it into chunks for the purpose of moving it onto the GPU, but then treating it as a single contiguous data structure once on the GPU.

**Testing**
There's a single test, that runs the same call made in the example itself, which allocates what _should_ be two buffers (on the CI GPU) worth of 0s, and then has the gpu add 1 to them.
lengths of the input and output are asserted to be equal.
contents of the returned array are asserted to all be 1.

**Checklist**

- [x] Run `cargo fmt`.


examples/src/big_compute_buffers/tests.rs
```diff
@@ -0,0 +1 @@
+
```
cwfitzgerald: Empty file? This example definitely needs tests

examples/src/big_compute_buffers/README.md
```diff
@@ -0,0 +1,98 @@
...
+- Returns those `1.0` values as a back to the HOST.
+
+The aim of this example is to demonstrate taking a large input dataset, splitting it into chunks for the purpose of moving it onto the GPU, but then treating it as a single contiguous data structure once on the DEVICE.
+
+## To Run
+As the maximum supported buffer size varies wildly per system, when you try to run this, then when it will likely fail, in-which-case read the error and update these `const`s accordingly:
+>`src/big_compute_buffers/mod.rs`
+```rust
+const MAX_BUFFER_SIZE: u64 = 1 << 27; // 134_217_728 // 134MB
+const MAX_DISPATCH_SIZE: u32 = (1 << 16) - 1; // 65_535
```
cwfitzgerald: These defaults should work everywhere, they're the minimum required by webgpu.

cwfitzgerald (changes requested):
Sorry for the long wait time for a review!

Frankly as it exists right now, we cannot accept this example. While it physically shows one strategy for dealing with large data sets, after reading it, the user doesn't get a good idea of _why_ that strategy should be used and what problems they are avoiding, compared to the more naive strategy of using larger and larger buffers. Through inline code comments and verbiage in the readme, the reader who has no idea about any of these topics (or even the details of memory allocation) should be able to understand why this is an effective strategy to utilize. 

Some things I think it should touch on:
- Large buffers may fail to allocate due to fragmentation
- Growing/shrinking a dataset with a buffer system requires copying the entire buffer contents, whereas pagenated data just requires rebuilding a bind group.

I'm not going to close this, as I do think this can be transformed into something that would be great to have.

Added a few incidental comments.

alphastrata:
> Sorry for the long wait time for a review!
> 
> Frankly as it exists right now, we cannot accept this example. While it physically shows one strategy for dealing with large data sets, after reading it, the user doesn't get a good idea of _why_ that strategy should be used and what problems they are avoiding, compared to the more naive strategy of using larger and larger buffers. Through inline code comments and verbiage in the readme, the reader who has no idea about any of these topics (or even the details of memory allocation) should be able to understand why this is an effective strategy to utilize. 
> 
> Some things I think it should touch on:
> - Large buffers may fail to allocate due to fragmentation
> - Growing/shrinking a dataset with a buffer system requires copying the entire buffer contents, whereas pagenated data just requires rebuilding a bind group.
> 
> I'm not going to close this, as I do think this can be transformed into something that would be great to have.
> 
> Added a few incidental comments.

Cheers, I'll keep working on it.

examples/src/big_compute_buffers/tests.rs
alphastrata: I could use a bit of advice on the test(s) for this: the included shader relies on traversing the contiguous array (made from the multiple buffers) via an `OFFSET` currently it needs to know this information ahead of compile time, hence the `const`s.
```rust
const OFFSET: u32 = 1u << 8u; 
const BUFF_LENGTH: u32 = 1u << 25u;
const NUM_BUFFERS: u32 = 2u;
const TOTAL_SIZE: u32 = BUFF_LENGTH * NUM_BUFFERS;
```
This is of course assuming these values would make sense for the test environment, [I've chosen these values based on looking at the test output from previous runs](https://github.com/gfx-rs/wgpu/actions/runs/11001397782/job/30546188996#step:12:1096). 

I do not know of a better way in `wgsl` to assign suitable values to these dynamically at runtime, as opposed to what I've done here.

If any maintainers know a cleaner, or more idiomatic way that we could tweak the shader to avoid doing this, that'd be awesome.

alphastrata: On the fragmentation and pagination notes: 
1) for the fragmentation is a warning similar to what's mentioned in the doc-string on `wgpu::Limits` enough? We could link to some external docs, perhaps https://developer.nvidia.com/docs/drive/drive-os/archives/6.0.4/linux/sdk/common/topics/graphics_content/avoiding_memory_fragmentation.html

2) Perhaps a pagination example is also required that this example could link to? [I would be interested in working on that]

alphastrata (commented):
Added a test.

Asked some questions.

Attempted to simplify.

ErichDonGubler:
@cwfitzgerald: Ping as the main reviewer on this PR.

cwfitzgerald:
Yup, on my list

examples/src/big_compute_buffers/tests.rs
cwfitzgerald: > I could use a bit of advice on the test(s) for this: the included shader relies on traversing the contiguous array (made from the multiple buffers) via an OFFSET currently it needs to know this information ahead of compile time, hence the consts.

the first issue seems like it has been fixed?

> Perhaps a pagination example is also required that this example could link to? [I would be interested in working on that]

Isn't this supposed to be an example of pagination?

> for the fragmentation is a warning similar to what's mentioned in the doc-string on wgpu::Limits enough? We could link to some external docs, perhaps

I would just discuss how larger allocations are more prone to memory fragmentation as you always need to find contiguous space for each allocation

alphastrata:
--Did not mean to close, sorry--

examples/src/big_compute_buffers/tests.rs
alphastrata: > > the first issue seems like it has been fixed?
>
> **Fixed**, yes, but it's neither nice, nor idiomatic I was hoping some maintainers may have a 'prettier' solution.

> > Isn't this supposed to be an example of pagination?
>
>  I have been doing CUDA for nearly 7 years, and have never heard it called this, we always call it 'chunking'. Asking around I didn't find anyone who recognised it outside of the context it's usually used in the web spheres for.. but ok.

> >I would just discuss how larger allocations are more prone to memory fragmentation as you always need to find contiguous space for each allocation
>
> I've added some cautionary notes & suggested further reading.

alphastrata:
> Yup, on my list

What else does the group want done with this one to get it into shape? ( assuming that none of this https://github.com/gfx-rs/wgpu/pull/6138#discussion_r1895477405 was serious deemed serious )

cwfitzgerald:
Sorry, this is on me, I didn't realize this was updated and didn't get back on my list, going to look at this early next week.

examples/src/big_compute_buffers/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+            label: Some("compute pass descriptor"),
+            timestamp_writes: None,
+        });
+        cpass.set_pipeline(&compute_pipeline);
+        cpass.set_bind_group(0, Some(&bind_group), &[]);
+
+        cpass.dispatch_workgroups(MAX_DISPATCH_SIZE.min(numbers.len() as u32), 1, 1);
+    }
+
+    storage_buffers.iter().zip(staging_buffers.iter()).for_each(
```
jimblandy: This should just be a `for` loop.

There's nothing glorious about `Iterator::for_each`, unless you have a very long chain of iterator combinators and want to make sure the variables for each iteration (`storage_buffer` and `staging_buffer`, in this example) are named closer to the code that uses them. That's not a problem here.

examples/src/big_compute_buffers/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+                0,
+                stg_size,
+            );
+        },
+    );
+
+    queue.submit(Some(encoder.finish()));
+
+    let mut buffer_slices = Vec::new();
+    staging_buffers.iter().for_each(|sb| {
```
jimblandy: Can this be a `collect`?

examples/src/big_compute_buffers/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+        buffer_slices.push(sb.slice(..));
+    });
+
+    let (sender, receiver) = flume::bounded(buffer_slices.len());
+    let sender = Arc::new(sender);
```
jimblandy: There's no reason to create an `Arc` of a `flume::Sender`, is there? It implements `Clone` just the way you'd want it to.

examples/src/big_compute_buffers/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    buffer_slices.iter().for_each(|bs| {
```
jimblandy: This should be a `for` loop.

examples/src/big_compute_buffers/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+        let sender = sender.clone();
+        bs.map_async(wgpu::MapMode::Read, move |v| {
+            sender.send(v).unwrap();
+        })
+    });
+
+    device.poll(wgpu::Maintain::wait());
+
+    if let Ok(Ok(())) = receiver.recv_async().await {
```
jimblandy: Maybe I'm misreading this, but... it seems like we send all our results to the flume channel, but we only check one of them? This doesn't make much sense. Also, I know that `Maintain::wait` will wait for all the `map_async` calls to finish, but it's just weird to go and assume that all the ranges are mapped when we technically haven't received their statuses from the channel.

So while this isn't technically wrong, it's really not what I want to see in an example.

examples/src/big_compute_buffers/README.md
```diff
@@ -0,0 +1,37 @@
+# big-compute-buffers
+
+This example assumes you're familiar with the other GP-GPU compute examples in this repository, if you're not you should go look at those first.
+
+This example also assumes you've specifically come here looking to do this, because you want at least the following:
+1. To be working on your 'data' in your shader treating it contiguously, not batching etc.
+2. The data you are wanting to work on does **not** fit within a single buffer on your device, see the [hello](https://github.com/gfx-rs/wgpu/tree/trunk/examples/src/hello) example for how to print information about your unique device to explore its maximum supported buffer size.
+
+Demonstrates how to split larger datasets (things too big to fit into a single buffer), across multiple buffers whilst treating them as a single, contiguous buffer on the GPU. This is known as 'chunking' or, sometimes 'pagination', not to be confused with the web kind, although they're not entirely dissimilar.
```
jimblandy: This doesn't accurately describe what the code does. The WGSL shader operates on a `binding_array<ContiguousArray, N>`, which is not a "single, contiguous buffer".

You could certainly say, "Well, it's basically the same idea." But the ability to just glide over the difference between what's described and what the code actually does depends on the skill you're trying to teach here. You're explaining something in a way that assumes the reader already understands it.

You should instead describe the code accurately, and say something to the effect of, "You can't get a buffer big enough, but you can use a `binding_array` of buffers instead, and then it's almost the same, you just need to split the index into two, or use a nested loop, check out my shader".

examples/src/big_compute_buffers/shader.wgsl
```diff
@@ -0,0 +1,34 @@
+const OFFSET: u32 = 1u << 8u;
+const BUFF_LENGTH: u32 = 1u << 25u;
```
jimblandy: It's probably worth explaining why this is `1 << 25u` but the number in the Rust code is `1 << 27u`.

examples/src/big_compute_buffers/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+            .iter()
+            .flat_map(|bs| {
+                let data = bs.get_mapped_range();
+                let result: Vec<f32> = bytemuck::cast_slice(&data).to_vec();
+                drop(data); // Drop to free buffer before unmap
+                result
+            })
+            .collect();
+
+        staging_buffers.iter().for_each(|sb| sb.unmap());
```
jimblandy: This should be a `for` loop.

examples/src/big_compute_buffers/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+        let total_bytes = (0.25 * bytes_per_gb as f32) as usize;
+        let elements = total_bytes / bytes_per_f32;
+
+        vec![0.0; elements]
+    };
+    assert!(numbers.iter().all(|n| *n == 0.0));
+    log::info!("All 0.0s");
+
+    let t1 = std::time::Instant::now();
+    let results = execute_gpu(&numbers).await.unwrap();
```
jimblandy: Rather than having everything return an `Option` that gets unwrapped here, I think you get better stack traces if you just use `panic!` or `unreachable!`. (Clippy says you shouldn't use `panic!` in production code, but 1) this isn't production code and 2) I'm not sure I agree with Clippy in principle.)

examples/src/big_compute_buffers/mod.rs
jimblandy: It is more legible to handle errors immediately, and let the function's main flow of control (the least indented) be the success path. That is:

```
let Ok(Ok(())) = receiver.recv_async().await else {
    log::error!("Failed to run compute on GPU!");
    return None;
};
```

and then have the success case, where we iterate over `buffer_slices`, follow directly.

examples/src/big_compute_buffers/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+    let max_elements_per_chunk = max_buffer_size as usize / std::mem::size_of::<f32>(); // Calculate max f32 elements per buffer
+    numbers.chunks(max_elements_per_chunk).collect()
+}
+
+fn create_storage_buffers(
+    device: &wgpu::Device,
+    numbers: &[f32],
+    input_size: u64, // bytes..
+) -> Vec<wgpu::Buffer> {
+    if input_size > MAX_BUFFER_SIZE {
```
jimblandy: Do we really need to have separate paths for small and large inputs? Won't the path that calls `calculate_chunks` work fine in both cases?

examples/src/big_compute_buffers/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+        })
+        .collect()
+}
+
+async fn run() {
+    let numbers = {
+        let bytes_per_gb = 1024 * 1024 * 1024;
+        let bytes_per_f32 = std::mem::size_of::<f32>();
+        let total_bytes = (0.25 * bytes_per_gb as f32) as usize;
+        let elements = total_bytes / bytes_per_f32;
```
jimblandy: What is this 0.25, and why isn't it redundant with the `/ bytes_per_f32` on the next line?

jimblandy (changes requested):
@cwfitzgerald asked me to have a look at this as well. I think it's great to have an example that shows a `binding_array` of buffers. There are a few things that I think could be improved, though.

CHANGELOG.md
```diff
@@ -520,6 +520,7 @@ By @MarijnS95 in [#6006](https://github.com/gfx-rs/wgpu/pull/6006).
 
 #### General
 
+- Added an example that splits up large data across multiple buffers and accesses them as a single contiguous array in the shader, demonstrating the naga/wgsl only `binding_array`, and some of the, less documented `wgpu::Features`. By @alphastrata in [#6138](https://github.com/gfx-rs/wgpu/pull/6138)
```
jimblandy: I made a similar comment in the main review, but: this just doesn't describe what the example actually does. I feel like something like this would be better:

```suggestion
- Added an example that shows how to handle datasets too large to fit in a single `GPUBuffer` by distributing it across many buffers, and then having the shader receive them as a `binding_array` of storage buffers.  By @alphastrata in [#6138](https://github.com/gfx-rs/wgpu/pull/6138)
```

alphastrata: Thanks @jimblandy, great review, lots of good catches --(totally disagree with you on the for loops :p but, it's not my repo so I'll get in line). I'll make all these changes at some stage over the next few days.

jimblandy: @alphastrata If you're curious, I put my rationale for preferring ordinary control flow (like `for` loops) over combinator chains here: https://github.com/gfx-rs/wgpu/pull/6148#discussion_r1826262913

examples/src/big_compute_buffers/mod.rs
alphastrata: Is there a convention that I can follow? 
I was trying do adhere to the pattern a few existing examples had used:
https://github.com/gfx-rs/wgpu/blob/93f64dc847a0e6788b17b292ae4d7d59361e3be0/examples/features/src/hello_synchronization/mod.rs#L185C1-L187C51

examples/src/big_compute_buffers/mod.rs
alphastrata: this is such a good catch

examples/src/big_compute_buffers/mod.rs
alphastrata: This is also a good catch.

(during testing [because I may or may not have been playing DOTA] I carelessly slapped this in there to reduce it to 1/4 of a GB and never took it back out ><)

examples/src/big_compute_buffers/mod.rs
alphastrata: the whole reason we use flume...

examples/src/big_compute_buffers/mod.rs
jimblandy: The code in `hello_synchronization/mod.rs` does a single `map_async` call, and a single `recv_async`, so it gets all the results. I think you want to have a loop that receives `buffer_slices.len()` results on the channel and checks them all.

examples/src/big_compute_buffers/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+        .unwrap();
+
+    execute_gpu_inner(&device, &queue, numbers).await.unwrap()
+}
+
+pub async fn execute_gpu_inner(
+    device: &wgpu::Device,
+    queue: &wgpu::Queue,
+    numbers: &[f32],
+) -> Option<Vec<f32>> {
```
jimblandy: This, too, should just return a `Vec<f32>`, without the `Option`. Instead of logging a message and returning `None` if there's an error, it should just use `panic!` to report that message.

examples/src/big_compute_buffers/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+    for bs in buffer_slices.iter() {
+        let sender = sender.clone();
+        bs.map_async(wgpu::MapMode::Read, move |v| {
+            sender.send(v).unwrap();
+        })
+    }
+
+    device.poll(wgpu::Maintain::wait());
+
+    let Ok(Ok(())) = receiver.recv_async().await else {
```
jimblandy: This should be a loop that receives `buffer_slices.len()` results from the channel and checks each of them.

jimblandy (dismissed):
Two more notes. There are still a few changes requested in the prior review that haven't been addressed yet.

examples/src/big_compute_buffers/mod.rs
alphastrata: you ok with 
```rust
assert_eq!(
        receiver.into_iter().map(|v| { v.unwrap() }).count(),
        buffer_slices.len()
    );
``` 
...or do we need to do something more robust? It's only receiving `()` unit types after all, the message arriving is basically all the info we'll get.

examples/src/big_compute_buffers/README.md
alphastrata: Yes. I see your point.

alphastrata:
Converting to draft -- As I've let this get too far behind the current latest (also to spare the pipeline).

examples/src/big_compute_buffers/mod.rs
alphastrata: removed entirely as not needed.

CHANGELOG.md
```diff
@@ -773,6 +774,7 @@ By @MarijnS95 in [#6006](https://github.com/gfx-rs/wgpu/pull/6006).
 
 #### General
 
+- Added an example that shows how to handle datasets too large to fit in a single `GPUBuffer` by distributing it across many buffers, and then having the shader receive them as a `binding_array` of storage buffers.  By @alphastrata in [#6138](https://github.com/gfx-rs/wgpu/pull/6138)
```
cwfitzgerald: Before we merge, can this be moved to the new `Unreleased` section?

cwfitzgerald (approved):
Thank you! Sorry for the long delay in merging!

CHANGELOG.md
alphastrata: Yep, ETA 2, 3 hrs.

cwfitzgerald:
CI is mad

alphastrata:
> CI is mad

this is a vulkan only feature -- am I missing something about having its tests excluded from wasm?  Should I just slap a `cfg_if!` macro in there?

PR #6144, opened 2024-08-23
Further error handling improvements

kornelski:
**Connections**
More of #6126

**Description**
* Removed boxing of `default_error_handler` in `ErrorSinkRaw`, which apart form avoiding an allocation in a common case, also lets `#[track_caller]` to work for that handler. Panics will show location of the function that triggered the error. This is going to be a function in wgpu, rather than user code, because user code goes through `dyn` calls that "forget" the caller.
* Reduced code size in `handle_error_inner`.

**Testing**
If you trigger an error with the default error handler, it'll point to `handle_error`'s caller, rather than the contextless `panic!("wgpu error: {err}\n")` line.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu/src/api/device.rs
+++ b/wgpu/src/api/device.rs
@@ -681,30 +681,31 @@ pub enum ErrorFilter {
 }
 static_assertions::assert_impl_all!(ErrorFilter: Send, Sync);
 
+/// Lower level source of the error.
+///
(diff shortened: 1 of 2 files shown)


ErichDonGubler (approved):
LGTM! Thanks. 🙂

PR #6146, opened 2024-08-23
Reduce size of parse error

kornelski:
**Description**
Reduces size of parse error from 112 bytes to 48 bytes. This should help with stack usage during parsing, which is especially important given [the huge warning about heavy stack usage](https://docs.rs/naga/latest/naga/front/wgsl/fn.parse_str.html).

**Testing**
There's a test asserting the new size.

It has a noticeable performance improvement:

```text
front/shader: wgsl-in   time:   [5.3159 ms 5.3196 ms 5.3245 ms]
                        thrpt:  [291.07 MiB/s 291.34 MiB/s 291.54 MiB/s]
                 change:
                        time:   [-6.7480% -4.1415% -1.9554%] (p = 0.00 < 0.05)
                        thrpt:  [+1.9944% +4.3204% +7.2363%]
                        Performance has improved.
```

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/naga/src/front/wgsl/error.rs
+++ b/naga/src/front/wgsl/error.rs
@@ -10,6 +10,9 @@ use std::ops::Range;
 use termcolor::{ColorChoice, NoColor, StandardStream};
 use thiserror::Error;
 
+#[cfg(test)]
+use std::mem::size_of;
+
 #[derive(Clone, Debug)]
 pub struct ParseError {
     message: String,
@@ -144,7 +147,7 @@ pub enum InvalidAssignmentType {
 }
 
 #[derive(Clone, Debug)]
-pub enum Error<'a> {
+pub(crate) enum Error<'a> {
     Unexpected(Span, ExpectedToken<'a>),
     UnexpectedComponents(Span),
     UnexpectedOperationInConstContext(Span),
@@ -154,8 +157,8 @@ pub enum Error<'a> {
     BadTexture(Span),
     BadTypeCast {
         span: Span,
-        from_type: String,
-        to_type: String,
+        from_type: Box<str>,
+        to_type: Box<str>,
     },
     BadTextureSampleType {
         span: Span,
@@ -188,8 +191,8 @@ pub enum Error<'a> {
     TypeNotInferable(Span),
     InitializationTypeMismatch {
         name: Span,
-        expected: String,
-        got: String,
+        expected: Box<str>,
+        got: Box<str>,
     },
     DeclMissingTypeAndInit(Span),
     MissingAttribute(&'static str, Span),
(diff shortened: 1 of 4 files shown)


naga/src/front/wgsl/error.rs
```diff
@@ -144,7 +144,7 @@ pub enum InvalidAssignmentType {
 }
 
 #[derive(Clone, Debug)]
-pub enum Error<'a> {
+pub(crate) enum Error<'a> {
```
ErichDonGubler, line 150: issue: I know that some folks in the ecosystem have built tooling on top of `naga` by inspecting the concrete `Error` variants emitted by it. Is there a reason you're restricting this to `pub(crate)`?

naga/src/front/wgsl/error.rs
```diff
@@ -738,44 +747,44 @@ impl<'a> Error<'a> {
                 )],
                 notes: vec![],
             },
-            Error::AutoConversion { dest_span, ref dest_type, source_span, ref source_type } => ParseError {
```
ErichDonGubler, line 741: nitpick: It's preferable to keep destructuring these errors types. That way, we can force ourselves to use the diagnostic information we have, if we ever add any (which happens often).

ErichDonGubler (changes requested):
I'm open to this sort of change, but I noticed a couple of things that I think need to change with this before merges.

I also want to confirm with @teoxoy and/or @jimblandy that this is a desirable change. I suspect that it might get rejected unless we can justify a quantitative gain here in terms of our own benchmarks (which should be easy?). @kornelski: Does this empirically reduce a memory high mark, or improve run times?

naga/src/front/wgsl/error.rs
kornelski: This type hasn't been exposed publicly. It's in a private module. 

It's not a change, but clarification of the existing behavior and a fix of misleading `pub` on a non-public type.

naga/src/front/wgsl/error.rs
kornelski: Rust doesn't have matching on `Box`, so the result looks a bit clunky.

kornelski:
It reduces stack usage of `too_many_unclosed_loops` test (#5447) from 1.6MB to under 1MB.

teoxoy:
How does this compare with boxing the error type itself? Maybe it can be done non-invasively via a newtype.

ErichDonGubler (approved):
My concerns have been resolved. I will defer to the rest of @gfx-rs/naga for merging.

kornelski:
@teoxoy I've tried boxing errors. It decreases perf by 2%, and the change is way more invasive —  needs changing all return types, all `return Err` expressions, and closures using `?` need type hints added.

There are some functions that match on the errors, so to avoid boxing, unboxing, and boxing again, they need to keep returning the unboxed error.

https://github.com/kornelski/temp-wgpu-pr/commit/c414134eadf92d5b626d72ed6f1947ac35329d16

teoxoy:
I see, thanks for giving it a shot! I think we can land this.

Issue #5295, opened 2024-02-23
Support multiple diagnostic items in shader compilation


Runi-c:
**Is your feature request related to a problem? Please describe.**
Right now naga-based tooling can only display one validation error at a time, it would be nice if such tooling could get access to all the errors naga can detect in order to ease working with wgsl files especially.

**Describe the solution you'd like**
It seems like [`Validator::validate`](https://github.com/gfx-rs/wgpu/blob/trunk/naga/src/valid/mod.rs#L331) could be feasibly modified to produce an iterator over validation errors (or more likely such behavior moved into something like `Validator::validate_all` or `Validator::iter` which `validate` calls and returns the first element).

**Describe alternatives you've considered**
[`wgsl_analyzer`](https://github.com/wgsl-analyzer/wgsl-analyzer) works around this for wgsl files by implementing its own validation layer, but this results in occasional mismatches with naga and lack of support for naga extensions like [`naga_oil`](https://github.com/bevyengine/naga_oil).

ErichDonGubler:
AFAIK, most WGPU maintainership (myself included) is aware of this and wants it, but we haven't prioritized it yet (and we're not sure when we will). I might add as an alternative that we could also accept an error receiver of some kind from client code, like `&dyn FnMut(<some-error-type>)`, instead of forcing collection of errors. The design space still needs to be explored, and consensus still needs to be driven, but we're happy to have it. I'd likely be the one to field such API discussions, so HMU in Matrix or keep replying here if you'd like to get the ball rolling!

CC @udoprog.

Runi-c:
@ErichDonGubler I feel I'm not experienced enough in Rust to be able to accurately assess design considerations for different API suggestions, but I definitely have an interest in pushing for this, for the sake of being able to better work with WGSL for my current project.

An error receiver is an interesting idea, but at least for the tooling I'm working on, I'd likely just use it to produce a collection anyway. Are there use-cases that would be underserved by a simple `Iterator<Item = WithSpan<ValidationError>>`? Perhaps both methods could be supported to avoid the possibility of forcing collection boilerplate?

ErichDonGubler:
FTR: On Firefox's side, this is estimated at `XL` ("multiple weeks of work"). I think most of the work is going to be:

1. Design a new diagnostic emission API to be used by `naga`.
[...]

PR #6148, opened 2024-08-23
WGSL: Implement `diagnostic(…);` directives and `derivative_uniformity` triggering rule

### LINKED ISSUE #3135: Feature request: disable uniformity analysis on shader creation

### LINKED ISSUE #5320: Support diagnostic filters in WGSL (i.e., `diagnostic(…)`)

ErichDonGubler:
**Connections**

Resolution:

* Partially resolves #5320.
* Resolves #3135.

Dependencies:

* Hard dependency on #6352.
* Soft dependency on #6349. 

**Description**

_**Recommended review experience is commit-by-commit. Each individual commit should pass CI!**_

This change implements a meaningful first iteration of support for `diagnostic(…);` directives; to wit:

1. Define the essential state that Naga needs to track for compilation messages controlled by `diagnostic(…)` rules.
2. Control Naga's current uniformity analysis errors via the `derivative_uniformity` rule. The current analysis is incorrect, and blocks users in issues like #3135. This can give them a way forward.

Several things are _out-of-scope_:

* There are more grammar rules where we should also `@diagnostic(…)` attributes, in order to conform fully with the WGSL spec. This PR does _not_ attempt to handle all of these cases, to limit the complexity of this PR.

    The exhaustive list of standards-mandated parse sites for `@diagnostic(…)` not implemented in this PR is tracked in <https://github.com/gfx-rs/wgpu/issues/5320>.
    
* Naga's current diagnostic model for shader compilation is too limited for us to comply with the spec.-mandated behavior with warnings. Naga offers only a `Result` to indicate operation success, or a _single_ fatal error message. This conflicts with the `Severity::Warning` model introduced in this PR, because standard WGSL compilation might report multiple diagnostics in cases where parsing the shader module…

    1. …emits multiple warning items.
    2. …emits warnings _in addition to_ a fatal error-level item.
    
    Because of this limitation, and because we want to limit complexity in this PR, `diagnostic(warn, …)` will not report a standard warning-level compilation item in cases that the standard dictates a warning-level compilation item. Instead, we will opt for `log::warn!(…)` entries. This is tracked as follow-up issues:

    - #5295
    - #6458 

	New cases where the above is relevant:

	* unrecognized diagnostic rule names
	* uniformity analysis diagnostics emitted at a warning level

**Testing**

Current test plan:

- [x] Demonstrate that the following CTS cases work:
	- [x] [`webgpu:shader,validation,parse,diagnostic:valid_locations:*`](https://gpuweb.github.io/cts/standalone/?q=webgpu:shader,validation,parse,diagnostic:*)
		- [x] `…:type="directive";location="module"`
	- [x] [`webgpu:shader,validation,parse,diagnostic:conflicting_directive:*`](https://gpuweb.github.io/cts/standalone/?q=webgpu:shader,validation,parse,diagnostic:conflicting_directive:*)
- [x] ~~Port a meaningful subset of the above to our own Naga test suite, to prevent the most impactful regressions.~~ Punted to <https://github.com/gfx-rs/wgpu/pull/6457>.

**Checklist**

- [x] Run `cargo fmt`.


naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2433,4 +2454,43 @@ impl Parser {
...
+        lexer.expect(Token::Separator(','))?;
+        let (diagnostic_rule_name, diagnostic_rule_name_span) = lexer.capture_span(|lexer| {
+            // TODO: ensure that errors are good
+            let mut segments = Vec::new();
+            let (first_segment, _span) = lexer.next_ident_with_span()?;
+            segments.push(first_segment);
+            while lexer.skip(Token::Separator('.')) {
+                let (segment, _span) = lexer.next_ident_with_span()?;
+                segments.push(segment);
+            }
```
ErichDonGubler: There are, at most, two segments here (called "diagnostic name tokens" upstream, maybe rename?), per [spec.](https://www.w3.org/TR/WGSL/#diagnostic-triggering-rule):

> The name of a [triggering rule](https://www.w3.org/TR/WGSL/#diagnostic-triggering-rule) is either:
> 
> -   a [diagnostic\_name\_token](https://www.w3.org/TR/WGSL/#syntax-[diagnostic_name_token](https://www.w3.org/TR/WGSL/#syntax-diagnostic_name_token)), or
>     
> -   two [diagnostic\_name\_token](https://www.w3.org/TR/WGSL/#syntax-[diagnostic_name_token](https://www.w3.org/TR/WGSL/#syntax-diagnostic_name_token)) tokens, separated by a period `'.'` (U+002E).

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2433,4 +2454,43 @@ impl Parser {
         }
         Ok(brace_nesting_level + 1)
     }
+
+    fn diagnostic_control<'a>(&self, lexer: &mut Lexer<'a>) -> Result<(), Error<'a>> {
+        // TODO(maybe): self.push_rule_span(Rule::???, lexer)
+        // TODO: I see `*_argument` helpers, but they're not used here. Guessing it's
+        // more valuable to make something easy to validate against the WGSL spec.?
```
ErichDonGubler: Wanted to ping @gfx-rs/naga on this question about `Lexer` usage:

```
        // TODO: I see `*_argument` helpers, but they're not used here. Guessing it's
        // more valuable to make something easy to validate against the WGSL spec.?
```

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2433,4 +2454,43 @@ impl Parser {
         }
         Ok(brace_nesting_level + 1)
     }
+
+    fn diagnostic_control<'a>(&self, lexer: &mut Lexer<'a>) -> Result<(), Error<'a>> {
+        // TODO(maybe): self.push_rule_span(Rule::???, lexer)
```
ErichDonGubler: @gfx-rs/naga: It's unclear to me when to use `Rule` spans. Should they be used for this new functionality? 🤔

jimblandy:
> Several things are out-of-scope:
>
> - There are more grammar rules where, in order to conform fully with the WGSL spec., we should optionally parse @diagnostic(…) attributes. This PR does not attempt to handle all of these cases, to limit the complexity of this PR. These cases should not be necessary up-front to deliver value to WGSL authors.

This is fine, but one of the reasons this issue is prioritized is that we want to be able to parse the shaders people are trying to run on the web. We can decide whether it's more convenient for us to address that in this PR or in follow-up PRs, but in order to deliver value to WGSL authors, the follow-up PRs may end up being just as urgent as this one.

jimblandy:
Overall, the scope and test plan in the OP looks fine to me.

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2433,4 +2454,43 @@ impl Parser {
...
+            "error" | "warning" | "info" | "off" => (/* neat */),
+            _ => todo!(
+                "diagnostic for invalid severity control name {:?}",
+                severity_control_name
+            ),
+        }
+        lexer.expect(Token::Separator(','))?;
+        let (diagnostic_rule_name, diagnostic_rule_name_span) = lexer.capture_span(|lexer| {
+            // TODO: ensure that errors are good
+            let mut segments = Vec::new();
```
jimblandy: Maybe use `Vec::with_capacity(2)`?

naga/src/front/wgsl/parse/mod.rs
jimblandy: I have always been very unsure what sorts of bugs the WGSL front end's `Rule` stack makes it easy to detect or fix. In a recursive descent parser, the parser's own Rust call stack indicates what we were in the midst of parsing, and the `Rule` stack is either identical to that, or misleading. 

It does seem like the code uses `push_rule_span` and `pop_rule_span` to build spans that cover non-trivial subtrees. But you can do that just as well by calling `Lexer::span_from` yourself; that's something you see in the code as well.

My personal preference would be to just use `Lexer::span_from` directly. Logically, it should be easier to debug span problems when there are fewer "players" influencing how they're produced.

naga/src/front/wgsl/parse/mod.rs
jimblandy: Honestly, I think when I see

```rust
self.expect(Token::Paren('('))?;
```

it's going to take me less time to understand and predict that code's behavior than if I see:

```rust
self.open_arguments()?;
```

I'd want to go check the definition of `open_arguments` to see if it was doing something unexpected or tricky. Neither is the case here, so I don't think `open_arguments` is providing net value to readers of Naga's WGSL front end.

naga/src/front/wgsl/parse/mod.rs
jimblandy: In WGSL, the idea is:

- If you see a single-component diagnostic rule name, it's an error if the name isn't recognized.
- If you see a double-component diagnostic rule name, it's *ignored* if the name isn't recognized. This lets you write `clippy.my_picky_lint` and have browsers ignore it.

So ultimately joining them into a `String` might not be what we want. But there's no need to fix that at this stage.

naga/src/front/wgsl/parse/mod.rs
ErichDonGubler: I was actually thinking of replacing this with an `ArrayVec`, pending tests with stack usage. But yes, failing that, setting capacity would be good.

ErichDonGubler:
@jimblandy:

> [O]ne of the reasons this issue is prioritized is that we want to be able to parse the shaders people are trying to run on the web.

Speaking as the person who's linked most (all?) concrete failures against [bug 1882800](
https://bugzilla.mozilla.org/show_bug.cgi?id=1882800) in Bugzilla, the cases I've actually observed blocking users are (1) directive usage, by a wide margin, and then (2) as an attribute of a function. I don't think I've observed other application uses except for parse testing, like with the CTS. Thus, I believe making this PR focus on (1) and then following up quickly with (2) should ease most of the pressure we have on this feature. After that point, I believe we can prioritize based on blockers reported by users.

naga/src/front/wgsl/parse/mod.rs
ErichDonGubler: That's how I felt too, with such trivial parse rules. Alright, I'll forgo using these helpers.

jimblandy:
>  the cases I've actually observed are (1) directive usage, by a wide margin, and then (2) as an attribute of a function

Okay, sounds like you're way ahead of me. +1

naga/src/front/wgsl/parse/mod.rs
ErichDonGubler: Invalidated by my latest force-push, which does the extract of a single diagnostic name token structurally instead.

naga/src/front/wgsl/parse/mod.rs
ErichDonGubler: Invalidated by my latest force-push, which does the extract of a single diagnostic name token structurally instead.

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2513,4 +2553,81 @@ impl Parser {
...
+    }
+
+    pub(crate) fn write_diagnostic_filters(
+        arena: &mut Arena<DiagnosticFilterNode>,
+        filters: DiagnosticFilterMap,
+        parent: Option<Handle<DiagnosticFilterNode>>,
+    ) -> Option<Handle<DiagnosticFilterNode>> {
+        filters
+            .into_iter()
+            .fold(parent, |acc, (triggering_rule, (new_severity, span))| {
```
ErichDonGubler: nitpick: s/acc/parent

naga/src/span.rs
ErichDonGubler, line 1: suggestion(blocking): With recent changes to use `Span`s only as supplementary information, changes to `Span` shouldn't be necessary any more. Let's not change it, if we can.

naga/src/valid/handles.rs
```diff
@@ -180,6 +183,20 @@ impl super::Validator {
             validate_type(ty)?;
         }
 
+        for (handle, _node) in diagnostic_filters.iter() {
+            handle.check_valid_for(diagnostic_filters)?;
+            let DiagnosticFilterNode {
+                inner: _,
+                parent: previous,
```
ErichDonGubler: nitpick: s/previous/parent

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2513,4 +2553,81 @@ impl Parser {
...
     }
+
+    fn diagnostic_filter<'a>(
+        &self,
+        lexer: &mut Lexer<'a>,
+    ) -> Result<Option<DiagnosticFilter>, Error<'a>> {
+        lexer.expect(Token::Paren('('))?;
+
+        let (severity_control_name, severity_control_name_span) = lexer.next_ident_with_span()?;
+        let new_severity = Severity::from_ident(severity_control_name).ok_or(
```
ErichDonGubler: nitpick: s/from_ident/from_word

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2474,6 +2481,39 @@ impl Parser {
 
         let mut lexer = Lexer::new(source);
         let mut tu = ast::TranslationUnit::default();
+        let mut diagnostic_filters = DiagnosticFilterMap::new();
+
+        // Parse directives.
+        while let Ok((ident, span)) = lexer.peek_ident_with_span() {
+            if let Some((kind, name)) = DirectiveKind::from_ident(ident) {
```
ErichDonGubler: nitpick: s/from_ident/from_word

naga/src/front/wgsl/parse/mod.rs
ErichDonGubler: Actually...nevermind.

naga/src/front/wgsl/parse/mod.rs
ErichDonGubler: Actually...nevermind.

naga/src/lib.rs
```diff
@@ -2271,4 +2272,10 @@ pub struct Module {
     pub functions: Arena<Function>,
     /// Entry points.
     pub entry_points: Vec<EntryPoint>,
+    /// Arena for all diagnostic filter rules parsed in this module, including those in functions.
+    pub diagnostic_filters: Arena<DiagnosticFilterNode>,
+    /// The head of a linked list of `diagnostic(…)` directives parsed in this module.
+    ///
+    /// TODO: doc more
```
ErichDonGubler: todo: do it

naga/src/valid/analyzer.rs
```diff
@@ -289,6 +290,9 @@ pub struct FunctionInfo {
 
     /// Indicates that the function is using dual source blending.
     pub dual_source_blending: bool,
+
+    // TODO: docs
```
ErichDonGubler: todo: do it

naga/src/front/wgsl/parse/ast.rs
```diff
@@ -26,6 +27,13 @@ pub struct TranslationUnit<'a> {
     /// These are referred to by `Handle<ast::Type<'a>>` values.
     /// User-defined types are referred to by name until lowering.
     pub types: Arena<Type<'a>>,
+
+    /// Arena for all diagnostic filter rules parsed in this module, including those in functions.
+    pub diagnostic_filters: Arena<DiagnosticFilterNode>,
+    /// The head of a linked list of `diagnostic(…)` directives parsed in this module.
+    ///
+    /// TODO: doc more
```
ErichDonGubler: todo: do it

naga/src/valid/analyzer.rs
```diff
@@ -842,8 +841,16 @@ impl FunctionInfo {
...
-                                    .with_span_handle(expr, expression_arena));
+                                let severity = DiagnosticFilterNode::search(
+                                    self.diagnostic_filter_head,
+                                    diagnostic_filter_arena,
+                                    FilterableTriggeringRule::DerivativeUniformity,
+                                );
+                                severity.report_diag(
+                                    FunctionError::NonUniformControlFlow(req, expr, cause)
+                                        .with_span_handle(expr, expression_arena),
+                                    |e, level| log::log!(level, "{e}"),
```
ErichDonGubler, line 863: todo(if-minor): This diagnostic has no source contextualization, and that makes it very difficult for a user to use this warning. Validation errors should be pretty easy to turn into full diagnostics; need to implement it, if it's easy.

naga/src/diagnostic_filter.rs
```diff
@@ -91,7 +111,108 @@ impl FilterableTriggeringRule {
...
+#[cfg_attr(feature = "serialize", derive(Serialize))]
+#[cfg_attr(feature = "deserialize", derive(Deserialize))]
+#[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
+pub struct DiagnosticFilterNode {
+    pub inner: DiagnosticFilter,
+    pub parent: Option<Handle<DiagnosticFilterNode>>,
+}
+
+impl DiagnosticFilterNode {
+    pub(crate) fn search(
```
ErichDonGubler, line 228: todo: docs

naga/src/front/wgsl/error.rs
```diff
@@ -300,12 +300,25 @@ pub(crate) enum Error<'a> {
...
+            triggering_rule,
+            triggering_rule_spans,
+        } = value;
+        Self::DiagnosticDuplicateTriggeringRule {
+            triggering_rule,
+            triggering_rule_spans,
+        }
+    }
+}
+
```
ErichDonGubler: thought: I'm not sure if this conversion is worth the ambiguity we're causing in other places, but I don't think it's unreasonable.

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -1967,9 +1965,9 @@ impl Parser {
                         let mut body = ast::Block::default();
                         if !lexer.skip(Token::Separator(';')) {
                             let (condition, span) = lexer.capture_span(|lexer| {
-                                let condition = self.general_expression(lexer, ctx)?;
-                                lexer.expect(Token::Separator(';'))?;
-                                Ok(condition)
+                                self.general_expression(lexer, ctx).and_then(|condition| {
+                                    lexer.expect(Token::Separator(';')).map(|()| condition)
+                                })
```
jimblandy: The original code seems much clearer to me than the `and_then` / `map` version.

jimblandy: Maybe I'm barking up the wrong tree, but I want to provide some background here:

The classic structured control flow syntax, with blocks of statements run in sequence, executed for their side effects, has a very fine theoretical pedigree; see Dijkstra's "A Discipline Of Programming", for example, which uses structured control flow as the basis for writing proofs about programs' behavior.

There is no deep reason to prefer 'functional' means like `and_then` and `map` to ordinary control. They are not inherently "cleaner" or "more well-defined". (These are not things you've said; I'm just going after arguments I remember advocates of functional programming having made in the past. For the record, I adore functional programming.) The choice should be made simply on the basis of which is quicker to read, and less distracting to the reader.

I have no opposition to `and_then` and `map` when they're clearer. But to me, the original code here - a series of statements, some of which have `?` operators in them - is much more straightforward.

naga/src/front/wgsl/error.rs
```diff
@@ -295,12 +295,29 @@ pub(crate) enum Error<'a> {
     DiagnosticInvalidSeverity {
         severity_control_name_span: Span,
     },
+    DiagnosticDuplicateTriggeringRule {
+        triggering_rule: FilterableTriggeringRule,
+        triggering_rule_spans: [Span; 2],
+    },
```
jimblandy: Would it make sense to simply let this be:

```rust
DiagnosticDuplicateTriggeringRule(ConflictingDiagnosticRuleError),
```

rather than duplicating the definition of `ConflictingDiagnosticRuleError` as the enum variant?

There's probably a reason, but I'm surprised we're not using `thiserror` here; I expected this to be a `#[from]` variant.

naga/src/valid/analyzer.rs
```diff
@@ -289,6 +290,9 @@ pub struct FunctionInfo {
 
     /// Indicates that the function is using dual source blending.
     pub dual_source_blending: bool,
+
+    // TODO: docs
+    diagnostic_filter_head: Option<Handle<DiagnosticFilterNode>>,
```
jimblandy: I would name this `diagnostic_filter_leaf`.

naga/src/lib.rs
```diff
@@ -2271,4 +2272,10 @@ pub struct Module {
...
+    pub diagnostic_filter_head: Option<Handle<DiagnosticFilterNode>>,
```
jimblandy: Similarly, I think `diagnostic_filter_leaf` is a better name here.

naga/src/lib.rs
```diff
@@ -2271,4 +2272,10 @@ pub struct Module {
     pub functions: Arena<Function>,
     /// Entry points.
     pub entry_points: Vec<EntryPoint>,
+    /// Arena for all diagnostic filter rules parsed in this module, including those in functions.
+    pub diagnostic_filters: Arena<DiagnosticFilterNode>,
+    /// The head of a linked list of `diagnostic(…)` directives parsed in this module.
```
jimblandy: This is probably fine, but I think it's helpful to draw at least a light line between Naga IR concepts and WGSL concepts. So this would be something like:

```rust
/// The leaf node of all module-wide diagnostic filters.
///
/// In WGSL, this includes all global diagnostic filter directives.
```

jimblandy (commented):
Some thoughts about naming.

naga/src/front/wgsl/error.rs
ErichDonGubler: Moving conversation to <https://github.com/gfx-rs/wgpu/pull/6148#discussion_r1826293651>.

naga/src/lib.rs
ErichDonGubler: Fixed with 04e248c66621fe0395532fed2d9a21a59e352bc6.

naga/src/valid/analyzer.rs
ErichDonGubler: Fixed with 04e248c6.

naga/src/front/wgsl/parse/ast.rs
ErichDonGubler: Fixed with 04e248c66621fe0395532fed2d9a21a59e352bc6.

naga/src/diagnostic_filter.rs
ErichDonGubler: Documented with the force-pushed d581c157.

naga/src/front/wgsl/parse/mod.rs
ErichDonGubler: The reason this "needed" to change was because I wanted to avoid specifying the error type for the closure, and this let me do it. With such a vociferous complaint, however, I see no reason not to just eat the return type. 😅

Fixed in the force-pushed 5fad2eff.

naga/src/front/wgsl/error.rs
ErichDonGubler: I raised the premise of your question in <https://github.com/gfx-rs/wgpu/pull/6148#discussion_r1823756480>. Not sure if you saw it, but 🤷🏻‍♂️ I'm fine moving this to simply be within `DiagnosticDuplicateTriggeringRule`.

Implemented as suggested with the force-pushed 4fc60fe9.

naga/src/valid/analyzer.rs
ErichDonGubler: Fixed in the force-pushed d581c157.

naga/src/lib.rs
ErichDonGubler: Fixed in the force-pushed d581c157.

naga/src/lib.rs
ErichDonGubler: Added s'more in 04e248c6.

naga/src/valid/analyzer.rs
ErichDonGubler: Unfortunately, this is not easy, because this context does not have access to source. It's assumed that a user will provide the source when rendering these diagnostics. Punting to <https://github.com/gfx-rs/wgpu/issues/6458>.

ErichDonGubler:
@jimblandy: This is ready for a full review!

ErichDonGubler:
Just validated that tests written in the (revised) `Testing` portion of the OP work as expected when consumed by Firefox. 👏🏻

CHANGELOG.md
```diff
@@ -40,11 +40,45 @@ Bottom level categories:
...
+### The `diagnostic(…);` directive is now supported in WGSL
+
+Naga now parses `diagnostic(…);` directives according to the WGSL spec. This
+allows users to control certain lints, similar to Rust's `allow`, `warn`, and
+`deny` attributes. For example, this snippet would emit a uniformity error:
+
+```wgsl
+@compute
+fn vs_main() {
+  TODO // Emits a derivative uniformity error during validation.
```
ErichDonGubler: @gfx-rs/naga: Do we have an example of code that should fire a uniformity analysis error here?

jimblandy: Edit: Sorry, you probably want something that gives you a suppressible error. I don't think the example below can be suppressed.

Here's an example:

```wgsl
@compute @workgroup_size(10)
fn f(@builtin(local_invocation_index) lii: u32) {
  if (lii == 0) {
    workgroupBarrier();
  }
}
```

Tint rejects this code:

```
-*- mode: compilation; default-directory: "~/play/uniformity-error/" -*-
Compilation started at Tue Nov  5 11:29:48

tint uniformity-error.wgsl
uniformity-error.wgsl:4:5 error: 'workgroupBarrier' must only be called from uniform control flow
    workgroupBarrier();
    ^^^^^^^^^^^^^^^^

uniformity-error.wgsl:3:3 note: control flow depends on possibly non-uniform value
  if (lii == 0) {
  ^^

uniformity-error.wgsl:3:7 note: builtin 'lii' of 'f' may be non-uniform
  if (lii == 0) {
      ^^^
```

jimblandy: [This one](https://gpuweb.github.io/gpuweb/wgsl/#example-e515f7e4) might be better. Since this PR doesn't implement `@diagnostic` attributes on statements, you'll probably want to take out the `else` clause.

```wgsl
diagnostic(off,derivative_uniformity);
var<private> d: f32;
fn helper() -> vec4<f32> {
  if (d < 0.5) {
    // The derivative_uniformity diagnostic is disabled here
    // by the global diagnostic filter.
    return textureSample(t,s,vec2(0,0));
  } else {
    // The derivative_uniformity diagnostic is set to 'warning' severity.
    @diagnostic(warning,derivative_uniformity) {
      return textureSample(t,s,vec2(0,0));
    }
  }
  return vec4(0.0);
}
```

ErichDonGubler: Neither of those are rejected on basis of derivative uniformity in Naga currently, so they're not super workable. 🙁 The second example doesn't even compile in Tint; neither `t` nor `s` are defined.

I'll see if I can make my own example. LMK if you have other ideas, though!

ErichDonGubler:
Yo @jimblandy, I can't merge this until you approve!

CHANGELOG.md
ErichDonGubler: Pushed a dissatisfying "here's a snippet that standard WGSL wouldn't accept" snippet that I derived from a CTS test case with 02daf28.

If we can find one that actually gets rejected by Naga later, then great, we can edit the `CHANGELOG` then.

jimblandy (dismissed):
As far as I can see, this never actually populates `naga::Module::diagnostic_filters` or `diagnostic_filter_leaf` with what was parsed by the WGSL front end. The PR description says this is supposed to control how validation reports `FunctionError::NonUniformControlFlow`, so I would have expected the diagnostics to have been passed through to the `Module`.

If this is correct, then that also indicates some missing tests.

ErichDonGubler:
@jimblandy: Ah, it seems I'd missed a couple of things:

1. Needed to remove `DISABLE_UNIFORMITY_REQ_FOR_FRAGMENT_STAGE`, which I've squashed into 2841f181. I added coverage to the existing `uniform_control_flow` test to check that one of the offending 

    It would be _much_ better if I had a WGSL snippet in hand that definitively fails, both for an end-to-end test and something for the `CHANGELOG`, but...we've struggled to find one, and I think the two hours I spent trying to figure it out needs to be price enough.
2. Needed to wire up the transfer of state from a `TranslationUnit` into a `Module`, which I've squashed in f950c596. I added a test to ensure that a module's IR actually contains a parsed diagnostic rule.

naga/src/valid/handles.rs
```diff
@@ -180,6 +183,19 @@ impl super::Validator {
             validate_type(ty)?;
         }
 
+        {
+            for (handle, _node) in diagnostic_filters.iter() {
+                handle.check_valid_for(diagnostic_filters)?;
+                let DiagnosticFilterNode { inner: _, parent } = diagnostic_filters[handle];
+                if let Some(handle) = parent {
+                    handle.check_valid_for(diagnostic_filters)?;
```
jimblandy: Should this be a check_dep?

naga/src/valid/analyzer.rs
```diff
@@ -15,19 +16,15 @@ use std::ops;
 
 pub type NonUniformResult = Option<Handle<crate::Expression>>;
 
-// Remove this once we update our uniformity analysis and
-// add support for the `derivative_uniformity` diagnostic
-const DISABLE_UNIFORMITY_REQ_FOR_FRAGMENT_STAGE: bool = true;
-
```
jimblandy, line 21: Hmm. This kludge was serving two purposes:
1) Cover for our inability to parse the directives (fixed by this PR).
2) Cover for the over-strictness of our analysis (not fixed by this PR).

I understand that we need to remove this if we want to be able to test the application of the directives, but I hope that point 2) doesn't cause us trouble.

jimblandy (approved):
Okay, this looks good. Just a few comments.

naga/src/diagnostic_filter.rs
```diff
@@ -91,7 +111,140 @@ impl FilterableTriggeringRule {
...
+#[cfg_attr(feature = "deserialize", derive(Deserialize))]
+#[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
 pub struct DiagnosticFilter {
     pub new_severity: Severity,
     pub triggering_rule: FilterableTriggeringRule,
 }
+
+/// A map of diagnostic filters to their severity and first occurrence's span.
+///
+/// Intended for front ends' first step into storing parsed [`DiagnosticFilter`]s.
```
jimblandy, line 124:
```suggestion
/// A map from diagnostic filters to their severity and span.
///
/// Front ends can use this to collect the set of filters applied to a
/// particular language construct, and detect duplicate/conflicting filters.
///
/// For example, WGSL has global diagnostic filters that apply to the entire
/// module, and diagnostic range filter attributes that apply to a specific
/// function, statement, or other smaller construct. The set of filters applied
/// to any given construct must not conflict, but they can be overridden by
/// filters on other constructs nested within it. A front end can use a
/// `DiagnosticFilterMap` to collect the filters applied to a single construct,
/// using the [`add`] method's error checking to forbid conflicts.
///
/// For each filter it contains, a `DiagnosticFilterMap` records the requested
/// severity, and the source span of the filter itself.
///
/// [`add`]: DiagnosticFilterMap::add
```

naga/src/valid/analyzer.rs
ErichDonGubler: I think we're in the clear with (2). If WGPU users run into errors that are disjoint with the spec., they should be able to `diagnostic(off, derivative_uniformity);`, and stay portable. This should work even in fully compliant environments, so I think this will be annoying for end users to deal with, but it will ultimately let them ship in more places.

naga/src/diagnostic_filter.rs
ErichDonGubler: I'll file a separate PR to follow up with this.

naga/src/valid/handles.rs
ErichDonGubler: Ah, yes, that would be better, wouldn't it? Each node's parent already gets checked in the iteration pass, so I should only need to call `diagnostic_filter_leaf.check_valid_for(…)` at the end.

Oh...come to think of it, I shouldn't need to do `handle.check_valid_for(…)` here, because the fact that the handle comes from an iterator over the arena makes the check redundant. 😅

naga/src/valid/analyzer.rs
ErichDonGubler: Fixed with the force-pushed 1ffd9c26366cfe7aabc06bc29884bb52740d1076.

naga/src/diagnostic_filter.rs
ErichDonGubler: Filed: https://github.com/gfx-rs/wgpu/pull/6524

PR #6149, opened 2024-08-23
hal: cargo feature to allow using `VK_GOOGLE_display_timing` unsafely

DJMcNab:
**Connections**
- https://raphlinus.github.io/ui/graphics/gpu/2021/10/22/swapchain-frame-pacing.html
- [#gpu > Latency Investigation](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Latency.20Investigation) on the Linebender Zulip.
- https://github.com/gfx-rs/wgpu/issues/2869

**Description**
In many applications, you need to do frame pacing properly to get acceptable latency performance.
On Android, this is exposed through the [VK_GOOGLE_display_timing](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_GOOGLE_display_timing.html) device extension. However, for applications using `wgpu`, it is currently impossible to use this extension for three reasons, each of which are addressed in this PR:
1) It is impossible to access the raw vulkan swapchain object (e.g. to call [`vkGetPastPresentationTimingGOOGLE`](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetPastPresentationTimingGOOGLE.html)
2) It is impossible to create a device with this extension enabled.
3) It is impossible to extend the `present` call with the [`VkPresentTimesInfoGOOGLE`](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPresentTimesInfoGOOGLE.html)

The solution here addresses the first of these by adding a `raw_swapchain` method to `wgpu_hal::vulkan::Surface`.
The remaining two are addressed using targeted hacks, both protected by the new perma-unstable `wgpu-hal` feature `unstable_vulkan_google_display_timing`. The extension is always enabled if available when the feature is set, and the additional present information for the "next" presentation is cached on the `wgpu-hal::vulkan::Surface` object, which will be accessed using [`wgpu::Surface::as_hal`](https://docs.rs/wgpu/latest/wgpu/struct.Surface.html#method.as_hal).

This PR does not change the high-level `wgpu` API, and instead requires users of this feature to add a dependency on `wgpu-hal` themselves.

**Testing**
This feature is currently not tested. My intention is to integrate this PR with [Vello](https://github.com/linebender/vello/) before this is merged.

I don't think it is possible to automate testing of this PR, as we do not have any tests running on Android, and only MoltenVK implements this outside of Android.

**Checklist**

- [X] Run `cargo fmt`.
- [ ] Run `cargo xtask test` to run tests. These kill my editor and any terminal emulator at the moment, but I can't imagine this changes anything.

### DIFF
--- a/wgpu-hal/Cargo.toml
+++ b/wgpu-hal/Cargo.toml
@@ -105,7 +105,6 @@ device_lost_panic = []
 # Only affects the d3d12 and vulkan backends.
 internal_error_panic = []
 
-
 [[example]]
 name = "halmark"
 
@@ -205,8 +204,8 @@ features = ["wgsl-in"]
 [dev-dependencies]
 cfg-if.workspace = true
 env_logger.workspace = true
-glam.workspace = true # for ray-traced-triangle example
-winit.workspace = true # for "halmark" example
+glam.workspace = true       # for ray-traced-triangle example
+winit.workspace = true      # for "halmark" example
 
 [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
 glutin.workspace = true # for "gles" example
--- a/wgpu-hal/Cargo.toml
+++ b/wgpu-hal/Cargo.toml
@@ -209,8 +209,8 @@ features = ["wgsl-in"]
 [dev-dependencies]
 cfg-if.workspace = true
 env_logger.workspace = true
-glam.workspace = true       # for ray-traced-triangle example
-winit.workspace = true      # for "halmark" example
+glam.workspace = true # for ray-traced-triangle example
+winit.workspace = true # for "halmark" example
 
 [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
 glutin.workspace = true # for "gles" example
--- a/wgpu-hal/Cargo.toml
+++ b/wgpu-hal/Cargo.toml
@@ -105,6 +105,11 @@ device_lost_panic = []
 # Only affects the d3d12 and vulkan backends.
 internal_error_panic = []
 
+# Enable the VK_GOOGLE_display_timing Vulkan extension if supported.
+# This API has no stability guarantees.
+# https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_GOOGLE_display_timing.html
+unstable_vulkan_google_display_timing = []
+
 [[example]]
 name = "halmark"
 
@@ -204,8 +209,8 @@ features = ["wgsl-in"]
 [dev-dependencies]
 cfg-if.workspace = true
 env_logger.workspace = true
-glam.workspace = true # for ray-traced-triangle example
-winit.workspace = true # for "halmark" example
+glam.workspace = true       # for ray-traced-triangle example
+winit.workspace = true      # for "halmark" example
 
 [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
 glutin.workspace = true # for "gles" example
--- a/wgpu-hal/Cargo.toml
+++ b/wgpu-hal/Cargo.toml
@@ -204,8 +204,8 @@ features = ["wgsl-in"]
 [dev-dependencies]
 cfg-if.workspace = true
 env_logger.workspace = true
-glam.workspace = true       # for ray-traced-triangle example
-winit.workspace = true      # for "halmark" example
+glam.workspace = true # for ray-traced-triangle example
+winit.workspace = true # for "halmark" example
 
 [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
 glutin.workspace = true # for "gles" example
--- a/wgpu-hal/Cargo.toml
+++ b/wgpu-hal/Cargo.toml
@@ -105,10 +105,6 @@ device_lost_panic = []
 # Only affects the d3d12 and vulkan backends.
 internal_error_panic = []
 
-# Enable the VK_GOOGLE_display_timing Vulkan extension if supported.
-# This API has no stability guarantees.
-# https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_GOOGLE_display_timing.html
-unstable_vulkan_google_display_timing = []
 
 [[example]]
 name = "halmark"
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -1,6 +1,6 @@
 use super::conv;
 
-use ash::{amd, ext, khr, vk};
+use ash::{amd, ext, google, khr, vk};
 use parking_lot::Mutex;
 
 use std::{collections::BTreeMap, ffi::CStr, sync::Arc};
@@ -1004,6 +1004,12 @@ impl PhysicalDeviceProperties {
             extensions.push(khr::shader_atomic_int64::NAME);
         }
 
+        #[cfg(feature = "unstable_vulkan_google_display_timing")]
+        // Support VK_GOOGLE_display_timing if it is requested *and* available.
+        if self.supports_extension(google::display_timing::NAME) {
+            extensions.push(google::display_timing::NAME);
+        }
+
         extensions
     }
 
@@ -1812,6 +1818,9 @@ impl super::Adapter {
             0, 0, 0, 0,
         ];
 
+        #[cfg(feature = "unstable_vulkan_google_display_timing")]
+        let has_display_timing = enabled_extensions.contains(&google::display_timing::NAME);
+
         let shared = Arc::new(super::DeviceShared {
             raw: raw_device,
             family_index,
@@ -1821,6 +1830,8 @@ impl super::Adapter {
             instance: Arc::clone(&self.instance),
             physical_device: self.raw,
             enabled_extensions: enabled_extensions.into(),
+            #[cfg(feature = "unstable_vulkan_google_display_timing")]
+            has_google_display_timing_extension: has_display_timing,
             extension_fns: super::DeviceExtensionFunctions {
                 debug_utils: debug_utils_fn,
                 draw_indirect_count: indirect_count_fn,
--- a/wgpu-hal/src/vulkan/adapter.rs
+++ b/wgpu-hal/src/vulkan/adapter.rs
@@ -1,6 +1,6 @@
 use super::conv;
 
-use ash::{amd, ext, khr, vk};
+use ash::{amd, ext, google, khr, vk};
 use parking_lot::Mutex;
 
 use std::{collections::BTreeMap, ffi::CStr, sync::Arc};
@@ -771,6 +771,11 @@ impl PhysicalDeviceFeatures {
             );
         }
 
+        features.set(
+            F::VULKAN_GOOGLE_DISPLAY_TIMING,
+            caps.supports_extension(google::display_timing::NAME),
+        );
+
         (features, dl_flags)
     }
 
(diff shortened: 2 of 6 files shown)


wgpu-hal/Cargo.toml
```diff
@@ -105,6 +105,11 @@ device_lost_panic = []
 # Only affects the d3d12 and vulkan backends.
 internal_error_panic = []
 
+# Enable the VK_GOOGLE_display_timing Vulkan extension if supported.
+# This API has no stability guarantees.
+# https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_GOOGLE_display_timing.html
+unstable_vulkan_google_display_timing = []
```
cwfitzgerald: I'm generally quite weary of adding more cargo features as it adds yet more ways that wgpu could end up not building.

I think we can do this entirely using `wgpu::Features` and I think this would be the preferred way. It could be something like `wgpu::Features::VULKAN_GOOGLE_DISPLAY_TIMING` to make it clear it's a vulkan only extension, and could be documented that you need to use wgpu-hal apis to make it work. If the feature is enable, we enable the extension and enable calling the display timing calls.

cwfitzgerald (changes requested):
I think we can activate the feature differently, but the code itself is fine.

wgpu-hal/src/vulkan/mod.rs
```diff
@@ -355,6 +355,11 @@ struct Swapchain {
     /// index as the image index, but we need to specify the semaphore as an argument
     /// to the acquire_next_image function which is what tells us which image to use.
     next_semaphore_index: usize,
+    /// The times which will be set in the next present times.
+    ///
+    /// SAFETY: This is only set if [wgt::Features::VULKAN_GOOGLE_DISPLAY_TIMING] is enabled, and
```
MarijnS95:
```suggestion
    /// SAFETY: This is only set if [`wgt::Features::VULKAN_GOOGLE_DISPLAY_TIMING`] is enabled, and
```

Fwiw safety docs usually use `/// # Safety`, whereas `// SAFETY:` is typically used in comments.

wgpu-hal/src/vulkan/mod.rs
```diff
@@ -355,6 +355,11 @@ struct Swapchain {
...
+    /// so `VK_GOOGLE_display_timing` is set.
```
MarijnS95: "is set" means "Vulkan extension is enabled"?

wgpu-hal/src/vulkan/mod.rs
```diff
@@ -375,6 +380,37 @@ pub struct Surface {
     swapchain: RwLock<Option<Swapchain>>,
 }
 
+impl Surface {
+    /// Get the raw Vulkan swapchain associated with this surface.
+    ///
+    /// Returns `None` if the surface is not configured.
```
MarijnS95:
```suggestion
    /// Returns [`None`] if the surface is not configured.
```

wgpu-hal/src/vulkan/mod.rs
```diff
@@ -1157,7 +1193,23 @@ impl crate::Queue for Queue {
...
+            debug_assert!(
+                ssc.device
+                    .features
+                    .contains(wgt::Features::VULKAN_GOOGLE_DISPLAY_TIMING),
+                "`next_present_times` should only be set if `VULKAN_GOOGLE_DISPLAY_TIMING` is enabled"
+            );
+            display_timing = vk::PresentTimesInfoGOOGLE::default();
+            present_times = [present_time];
+            display_timing = display_timing.times(&present_times);
+            // Safety: We know that VK_GOOGLE_display_timing is present because of the safety contract on `next_present_times`.
```
MarijnS95:
```suggestion
            // SAFETY: We know that VK_GOOGLE_display_timing is present because of the safety contract on `next_present_times`.
```

wgpu-types/src/lib.rs
```diff
@@ -952,6 +952,19 @@ bitflags::bitflags! {
         ///
         /// This is a native only feature.
         const SHADER_INT64_ATOMIC_ALL_OPS = 1 << 61;
+        /// Allows using the [VK_GOOGLE_display_timing] Vulkan extension.
+        /// This is used for frame pacing to reduce latency, and is generally only available on Android.
```
MarijnS95: Note that this will be formatted as one paragraph; I recommend to reflow the sentences.

wgpu-types/src/lib.rs
```diff
@@ -952,6 +952,19 @@ bitflags::bitflags! {
...
+        ///
+        /// This feature does not have a `wgpu`-level API, and must be accessed via `wgpu-hal`,
+        /// through various `as_hal` functions.
```
MarijnS95: Intradoc link to `as_hal()`?

wgpu-hal/src/vulkan/mod.rs
```diff
@@ -375,6 +380,37 @@ pub struct Surface {
...
+impl Surface {
+    /// Get the raw Vulkan swapchain associated with this surface.
+    ///
+    /// Returns `None` if the surface is not configured.
+    pub fn raw_swapchain(&self) -> Option<vk::SwapchainKHR> {
+        let read = self.swapchain.read();
+        read.as_ref().map(|it| it.raw)
+    }
+
+    /// Set the present timing information which will be used for the next presentation using `VK_GOOGLE_display_timing`.
```
MarijnS95: One more intradoc link?

Note that you didn't backtick the extension name below; can we settle on one and be consistent?

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -1004,6 +1009,11 @@ impl PhysicalDeviceProperties {
             extensions.push(khr::shader_atomic_int64::NAME);
         }
 
+        // Require `VK_GOOGLE_display_timing` if the associated feature was requested
+        if requested_features.contains(wgt::Features::VULKAN_GOOGLE_DISPLAY_TIMING) {
+            extensions.push(ash::google::display_timing::NAME);
```
MarijnS95: This wasn't qualified above, `google` is imported in scope. Doesn't `wgpu` check for `unused_qualifications`?

```suggestion
            extensions.push(google::display_timing::NAME);
```

wgpu-hal/src/vulkan/mod.rs
```diff
@@ -355,6 +355,11 @@ struct Swapchain {
     /// index as the image index, but we need to specify the semaphore as an argument
     /// to the acquire_next_image function which is what tells us which image to use.
     next_semaphore_index: usize,
+    /// The times which will be set in the next present times.
```
MarijnS95:
```suggestion
    /// The times which will be set in the next present times.
```

"the times ... next present times" -> this doesn't sound very clear to me. How about just copying the upstream `The earliest time image should be presented` documentation?

Also, you use the plural `times` everywhere when `VkPresentTimeGOOGLE` (as the singular `time` in the name implies) only contains a single name.

MarijnS95 (commented):
This looks interesting as a start but I wonder how useful it is without also providing the results from `vkGetPastPresentationTimingGOOGLE()`. Or are you driving these timings from `Choreographer` callbacks?

wgpu-hal/src/vulkan/mod.rs
DJMcNab: Already changed the wording here - how is the new one.

MarijnS95: Aynchronously writing reviews while a PR is actively force-pushed is very hard, yes - let me slowly close/rephrase whatever has already been changed since posting.
EDIT: I cannot "mark as resolved" my own review comments :upside_down_face: 

In hindsight I'm not sure about having `SAFETY` docs on a field in the first place: there's nothing unsafe happening on this line. Unwrapping on `Option` isn't unsafe either?

wgpu-hal/src/vulkan/mod.rs
```diff
@@ -355,6 +355,13 @@ struct Swapchain {
     /// index as the image index, but we need to specify the semaphore as an argument
     /// to the acquire_next_image function which is what tells us which image to use.
     next_semaphore_index: usize,
+    /// The present timing information which will be set in the next call to [`present`](crate::Queue::present).
+    ///
+    /// # SAFETY
```
MarijnS95:
```suggestion
    /// # Safety
```

wgpu-hal/src/vulkan/mod.rs
```diff
@@ -355,6 +355,13 @@ struct Swapchain {
     /// index as the image index, but we need to specify the semaphore as an argument
     /// to the acquire_next_image function which is what tells us which image to use.
     next_semaphore_index: usize,
+    /// The present timing information which will be set in the next call to [`present`](crate::Queue::present).
```
MarijnS95:
```suggestion
    /// The present timing information which will be set in the next call to [`present()`](crate::Queue::present()).
```

wgpu-hal/src/vulkan/mod.rs
```diff
@@ -1158,6 +1198,24 @@ impl crate::Queue for Queue {
...
+        let vk_info = if let Some(present_time) = ssc.next_present_times.take() {
+            debug_assert!(
+                ssc.device
+                    .features
+                    .contains(wgt::Features::VULKAN_GOOGLE_DISPLAY_TIMING),
+                "`next_present_times` should only be set if `VULKAN_GOOGLE_DISPLAY_TIMING` is enabled"
+            );
+            display_timing = vk::PresentTimesInfoGOOGLE::default();
+            present_times = [present_time];
+            display_timing = display_timing.times(&present_times);
```
MarijnS95:
```suggestion
            present_times = [present_time];
            display_timing = vk::PresentTimesInfoGOOGLE::default().times(&present_times);
```

wgpu-hal/src/vulkan/mod.rs
DJMcNab: Thanks for pointing this out - I'm used to following the usual Rust conventions on this, but I see that wgpu has their own.

wgpu-hal/src/vulkan/mod.rs
DJMcNab: > In hindsight I'm not sure about having `SAFETY` docs on a field in the first place: there's nothing unsafe happening on this line. Unwrapping on `Option` isn't unsafe either?

The field's value is relied upon for safety. There is a debug assertion protecting this condition.

However, it is UB in Vulkan to add the Present information without the extension enabled. We certainly could change the comment to not use the "# Safety" format, if that is preferable

DJMcNab:
> This looks interesting as a start but I wonder how useful it is without also providing the results from `vkGetPastPresentationTimingGOOGLE()`. Or are you driving these timings from `Choreographer` callbacks?

I have documented this in the PR on the feature now - is it clearer. Essentially, because this isn't a fully supported feature, I expect users to manually set these things up themselves (i.e. call [`ash::google::display_timing::Device::new()`](https://docs.rs/ash/latest/ash/google/display_timing/struct.Device.html#method.new), and handle the methods that way).

With the aim of keeping this change as focused as possible, I only provided  the absolute bare minimum of functionality to allow experimenting with this extension outside of wgpu - i.e. the functionality which could not be accessed from the outside.

DJMcNab:
I'm experimenting with this in my android-pacing branch, and indeed have it working as expected on Android https://github.com/DJMcNab/vello/commit/2673d0c3928d1ffdc097369fa47250a23e4a63a7

cwfitzgerald (approved):
Nice, thanks!

PR #5709, opened 2024-05-15
Update winit to 0.30 and refactor examples

anima-libera:
**Description**
*Attempted* to update winit to 0.30 and to refactor the examples accordingly. It does not work for the `wasm32` target (sorry, I had no idea what I was doing, trying to block in web browsers, not used to that); but for non-`wasm32` targets it *seems* to work (to the extent that I could test) so it may be useful in the end.

About the player crate, it is left untouched (so does not compile), but it can be ignored by changing [this line](https://github.com/gfx-rs/wgpu/blob/77a83fb0dd9f2295f25e99a850b9a031738925c3/player/Cargo.toml#L28) into
```rust
winit = { version = "0.29", features = ["android-native-activity"], optional = true }
```
(which may be a bad idea, since all the other dependencies seem in sync in the workspace) in order to focus on the examples or run xtask now.

There are 5 modules to partially refactor:
- the example `framework`
- the `hello_triangle`, `hello_windows` and `uniform_values` examples (they do not use the `framework` to handle the creation of their window(s) and event loop)
- the `player` crate

Note: I made the `hello_triangle` and `hello_windows` examples let the `WGPU_ADAPTER_NAME` env variable have a say in their choice of adapter (as the example framework does for all the examples that use it). (I did that to be able to run the examples that would no work with the default adapter on my machine, the issue there is most probably on my end and most probably not a concern.)

Note: There were mysterious numbers in the `uniform_values` example that could be labels referenced from some documentation that I could not find, they are mostly still above what they were labeling but since the code have moved around it may require whatever is referencing these labels to be partially rewritten to not rot.

**Testing**
This was tested on one machine (X11, Nvidia GPU) and works for the native target (some examples that use the framework were run, the three other examples were also run and they all seemed to work fine). Does not compile for `wasm32`. Even with `player`'s winit version set back down to 0.29 (to run xtask), xtask's output seems negative (and some part of that negativity is probably about `wasm32` target not even compiling in the `example` crate, hard to tell).

**Checklist**

- [X] Run `cargo fmt`.
- [X] Run `cargo clippy`. If applicable, add:
  - [ ] `--target wasm32-unknown-unknown`
  - [ ] `--target wasm32-unknown-emscripten`
- [ ] Run `cargo xtask test` to run tests.
- [ ] Add change to `CHANGELOG.md`. See simple instructions inside file.

PR #6150, opened 2024-08-23
[wgpu-hal] Upgrade `glutin` to `0.31`

MarijnS95:
**Connections**
Closes #5649 and should help towards #5709 where `raw-window-handle` can be used.

**Description**
`glutin 0.30` onwards completely refactored its internals to no longer be reliant on `winit`, as they (by default) have no direct relation except needing to perform _some_ operations (platform-specific) at strategic times in window creation and event loop handling.  Most of that is handled by a new `glutin-winit` introp crate, while the core `glutin` crate now exclusively focuses on wrapping the various OpenGL context APIs (CGL, EGL, WGL, ...).

This does result in a little more verbose handling to get the right `GLDisplay`, `GLConfig`, `GLContext` and `GLSurface`, but gives much more control and makes all intricacies more explicit.  Most of the code was copied from `glutin 0.31`'s example crate, with the code for transparency support removed.

Note that the example doesn't at all handle event loop events properly: resizes and redraws are not listened to, and mobile-specific surface events (`Resumed` and `Suspended`) are equally ignored.

**Testing**
`cargo r --example raw-gles` on `wayland` only with an AMD GPU and Mesa drivers (for now).

**Checklist**

- [x] Run `cargo fmt`.
  - Probably irrelevant for this kind of example-only change?

### DIFF
--- a/wgpu-hal/examples/raw-gles.rs
+++ b/wgpu-hal/examples/raw-gles.rs
@@ -12,11 +12,11 @@ extern crate wgpu_hal as hal;
 
 #[cfg(not(any(windows, target_arch = "wasm32", target_os = "ios")))]
 fn main() {
-    use std::ffi::CString;
+    use std::{ffi::CString, num::NonZeroU32};
 
     use glutin::{
         config::GlConfig as _,
-        context::{NotCurrentGlContext as _, Version},
+        context::{NotCurrentGlContext as _, PossiblyCurrentGlContext as _},
         display::{GetGlDisplay as _, GlDisplay as _},
         surface::GlSurface as _,
     };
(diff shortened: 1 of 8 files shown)


MarijnS95:
> Note that the example doesn't at all handle event loop events properly: resizes and redraws are not listened to, and mobile-specific surface events (`Resumed` and `Suspended`) are equally ignored.

~This is probably something we should get over with _now_ as it made/makes it easier to copy-paste the boiler-plate over from the `glutin` example, rather than trying to reconstruct/rethink it after the fact.~ EDIT: Done.

MarijnS95:
Regarding the build-failure, `glutin` is not supporting CGL on iOS, only MacOS: https://docs.rs/crate/glutin/0.31.3/source/build.rs

MarijnS95:
The duplicate `miniz_oxide` was introduced in https://github.com/gfx-rs/wgpu/pull/6135 but I guess the crate wasn't used until my PR here, which is why `cargo-deny` didn't stumble over it before?

wgpu-hal/examples/raw-gles.rs
```diff
@@ -117,10 +253,20 @@ fn main() {
...
 
-#[cfg(any(windows, all(target_arch = "wasm32", not(target_os = "emscripten"))))]
-fn main() {}
+#[cfg(any(
+    windows,
+    all(target_arch = "wasm32", not(target_os = "emscripten")),
+    target_os = "ios"
+))]
+fn main() {
+    eprintln!("This example is not supported on on Windows and non-emscripten wasm32")
```
cwfitzgerald:
```suggestion
    eprintln!("This example is not supported on Windows and non-emscripten wasm32")
```

wgpu-hal/examples/raw-gles.rs
```diff
@@ -10,61 +10,197 @@
...
-            .unwrap()
-    };
-    let inner_size = gl_context.window().inner_size();
+    let event_loop = winit::event_loop::EventLoop::new().unwrap();
+    // Only Windows requires the window to be present before creating the display.
+    // Other platforms don't really need one.
+    let window_builder = cfg!(windows).then(|| {
+        winit::window::WindowBuilder::new()
+            .with_title("WGPU raw GLES example (press Escape to exit)")
+    });
```
cwfitzgerald, line 35: Do we need this if Windows isn't supported in the example?

cwfitzgerald (changes requested):
Looking good in general, some small questions

wgpu-hal/examples/raw-gles.rs
MarijnS95: I'm adding that in #6152, fine to rip this out temporarily in this PR?

cwfitzgerald: Ah yup, no worries.

cwfitzgerald (approved):
Thanks!

PR #6152, opened 2024-08-24
[wgpu-hal] Allow importing external WGL contexts as with EGL

MarijnS95:
**Connections**
Depends on #6150

**Description**
Allows importing an external WGL context into `wgpu` to run the `raw-gles` example.

**Testing**
```shell
cargo r --example raw-gles
```

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-hal/src/gles/egl.rs
+++ b/wgpu-hal/src/gles/egl.rs
@@ -379,7 +379,7 @@ struct EglContextLock<'a> {
     display: khronos_egl::Display,
 }
 
-/// A guard containing a lock to an [`AdapterContext`]
+/// A guard containing a lock to an [`AdapterContext`], while the GL context is kept current.
 pub struct AdapterContextLock<'a> {
     glow: MutexGuard<'a, ManuallyDrop<glow::Context>>,
     egl: Option<EglContextLock<'a>>,
@@ -1082,7 +1082,9 @@ impl crate::Instance for Instance {
             unsafe { gl.debug_message_callback(super::gl_debug_message_callback) };
         }
 
-        // Avoid accidental drop when the context is not current.
+        // Wrap in ManuallyDrop to make it easier to "current" the GL context before dropping this
+        // GLOW context, which could also happen if a panic occurs after we uncurrent the context
+        // below but before AdapterContext is constructed.
         let gl = ManuallyDrop::new(gl);
         inner.egl.unmake_current();
 
--- a/wgpu-hal/src/gles/wgl.rs
+++ b/wgpu-hal/src/gles/wgl.rs
@@ -1,11 +1,3 @@
-use glow::HasContext;
-use glutin_wgl_sys::wgl_extra::{
-    Wgl, CONTEXT_CORE_PROFILE_BIT_ARB, CONTEXT_DEBUG_BIT_ARB, CONTEXT_FLAGS_ARB,
-    CONTEXT_PROFILE_MASK_ARB,
-};
-use once_cell::sync::Lazy;
-use parking_lot::{Mutex, MutexGuard, RwLock};
-use raw_window_handle::{RawDisplayHandle, RawWindowHandle};
 use std::{
     collections::HashSet,
     ffi::{c_void, CStr, CString},
@@ -19,6 +11,15 @@ use std::{
     thread,
     time::Duration,
 };
+
+use glow::HasContext;
+use glutin_wgl_sys::wgl_extra::{
+    Wgl, CONTEXT_CORE_PROFILE_BIT_ARB, CONTEXT_DEBUG_BIT_ARB, CONTEXT_FLAGS_ARB,
+    CONTEXT_PROFILE_MASK_ARB,
+};
+use once_cell::sync::Lazy;
+use parking_lot::{Mutex, MutexGuard, RwLock};
+use raw_window_handle::{RawDisplayHandle, RawWindowHandle};
 use wgt::InstanceFlags;
 use windows::{
     core::{Error, PCSTR},
@@ -48,7 +49,10 @@ impl AdapterContext {
     }
 
     pub fn raw_context(&self) -> *mut c_void {
-        self.inner.lock().context.context.0
+        match self.inner.lock().context {
+            Some(ref wgl) => wgl.context.0,
+            None => ptr::null_mut(),
+        }
     }
 
     /// Obtain a lock to the WGL context and get handle to the [`glow::Context`] that can be used to
@@ -62,7 +66,9 @@ impl AdapterContext {
             .try_lock_for(Duration::from_secs(CONTEXT_LOCK_TIMEOUT_SECS))
             .expect("Could not lock adapter context. This is most-likely a deadlock.");
 
-        inner.context.make_current(inner.device.dc).unwrap();
+        if let Some(wgl) = &inner.context {
+            wgl.make_current(inner.device.dc).unwrap()
+        };
 
         AdapterContextLock { inner }
     }
@@ -79,14 +85,15 @@ impl AdapterContext {
             .try_lock_for(Duration::from_secs(CONTEXT_LOCK_TIMEOUT_SECS))
             .expect("Could not lock adapter context. This is most-likely a deadlock.");
 
-        inner
-            .context
-            .make_current(device)
-            .map(|()| AdapterContextLock { inner })
+        if let Some(wgl) = &inner.context {
+            wgl.make_current(device)?;
+        }
+
+        Ok(AdapterContextLock { inner })
     }
 }
 
-/// A guard containing a lock to an [`AdapterContext`]
+/// A guard containing a lock to an [`AdapterContext`], while the GL context is kept current.
 pub struct AdapterContextLock<'a> {
     inner: MutexGuard<'a, Inner>,
 }
@@ -101,7 +108,9 @@ impl<'a> std::ops::Deref for AdapterContextLock<'a> {
 
 impl<'a> Drop for AdapterContextLock<'a> {
     fn drop(&mut self) {
-        self.inner.context.unmake_current().unwrap();
+        if let Some(wgl) = &self.inner.context {
+            wgl.unmake_current().unwrap()
+        }
     }
 }
 
@@ -136,7 +145,7 @@ unsafe impl Sync for WglContext {}
 struct Inner {
     gl: ManuallyDrop<glow::Context>,
     device: InstanceDevice,
-    context: WglContext,
+    context: Option<WglContext>,
 }
 
 impl Drop for Inner {
@@ -150,8 +159,14 @@ impl Drop for Inner {
 
         // Context must be current when dropped. See safety docs on
         // `glow::HasContext`.
-        self.context.make_current(self.device.dc).unwrap();
-        let _guard = CurrentGuard(&self.context);
+        //
+        // NOTE: This is only set to `None` by `Adapter::new_external` which
+        // requires the context to be current when anything that may be holding
+        // the `Arc<AdapterShared>` is dropped.
+        let _guard = self.context.as_ref().map(|wgl| {
+            wgl.make_current(self.device.dc).unwrap();
+            CurrentGuard(wgl)
+        });
         // SAFETY: Field not used after this.
         unsafe { ManuallyDrop::drop(&mut self.gl) };
     }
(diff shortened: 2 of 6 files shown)


MarijnS95:
Still WIP because ad-hoc (un)currenting the imported context creates a mess - I have to check how EGL importing is implemented.

If only `wgpu` used `glutin`... 🤤

wgpu-hal/src/gles/wgl.rs
```diff
@@ -101,7 +108,10 @@ impl<'a> std::ops::Deref for AdapterContextLock<'a> {
 
 impl<'a> Drop for AdapterContextLock<'a> {
     fn drop(&mut self) {
-        self.inner.context.unmake_current().unwrap();
+        // EGL performs a .take() here but that causes issues with drop ordering
+        if let Some(wgl) = &self.inner.context {
```
MarijnS95: I had a `.take()` here that may be the cause of all the spurious CI failures at https://github.com/gfx-rs/wgpu/actions/runs/10584695195/job/29329702561.

EGL does it too though, and unmaking the context current here pretty much defeats the purpose of #6114 doesn't it?

Imberflur: The patterns used for this lock guard between EGL and WGL aren't really consistent. The `.take()` in EGL isn't taking from the `egl` field in `AdapterContext` (which would have a similar effect to doing `self.inner.context.take()` here), instead it is a field in a different type that just holds a reference and I think a copy of the Display pointer (tbh it doesn't seem like the take is actually necessary in EGL since the used values are `Copy`).

MarijnS95:
@ErichDonGubler can you review this one? I prefer to clean up the `.take()` before merging though.

wgpu-hal/src/gles/wgl.rs
MarijnS95: Right, yes. Perhaps we should create a `WglContextLock` too and/or map the lock to `glow::Context` instead of lockign all of `Inner`?

It seems on EGL only `glow::Context` is in the lock whereas on WGL all of `glow:Context`, `InstanceDevice`, and `WglContext` are under the lock. We should make this more consistent.

ErichDonGubler:
> @ErichDonGubler can you review this one? I prefer to clean up the `.take()` before merging though.

😬 Iiii don't consider myself really qualified to review GL-related code, since I'm thoroughly unfamiliar with GL abstractions, but I'm happy to give it a shot and disclaim what I don't know! I've migrated some `winit` and `glutin` code, but that was mostly a transformation of existing code.

In particular, I don't know much about what conditions GL contexts need to be made current, when they might not be current, and what is safe and unsafe to do when they're not current.

Will finalize review after that (https://github.com/gfx-rs/wgpu/pull/6152#discussion_r1734430026), then. For now, I don't have any questions I expect will be useful to ask until more substantial review is satisfied.

wgpu-hal/src/gles/wgl.rs
Imberflur: > Right, yes. Perhaps we should create a `WglContextLock` too and/or map the lock to `glow::Context` instead of lockign all of `Inner`?
>

Imberflur: Perhaps yeah. An alternative is moving the EGL implementation to closer match the WGL one. I don't know if there are any differences between EGL and WGL that pushed them to these different approaches though.

> unmaking the context current here pretty much defeats the purpose of https://github.com/gfx-rs/wgpu/pull/6114 doesn't it?

Is this question still relevant or is it resolved now that the `take` behavior is clarified?

wgpu-hal/src/gles/wgl.rs
```diff
@@ -570,6 +587,43 @@ impl crate::Instance for Instance {
...
+impl super::Adapter {
+    /// Creates a new external adapter using the specified loader function.
+    ///
+    /// # Safety
+    ///
+    /// - The underlying OpenGL ES context must be current.
+    /// - The underlying OpenGL ES context must be current when interfacing with any objects returned by
+    ///   wgpu-hal from this adapter.
+    /// - The underlying OpenGL ES context must be current when dropping this adapter and when
+    ///   dropping any objects returned from this adapter.
```
Imberflur, line 594: On wgl I think it is not an `ES` context?

wgpu-hal/src/gles/egl.rs
```diff
@@ -1082,7 +1082,8 @@ impl crate::Instance for Instance {
             unsafe { gl.debug_message_callback(super::gl_debug_message_callback) };
         }
 
-        // Avoid accidental drop when the context is not current.
+        // Wrap in ManuallyDrop to make it easier to "current" the
+        // GL context before dropping this GLOW context.
         let gl = ManuallyDrop::new(gl);
         inner.egl.unmake_current();
```
Imberflur: This comment was trying to point out why we wrap in `ManuallyDrop` before calling `unmake_current`. I feel like the new comment is more about the role `ManuallyDrop` plays in the drop implementation of `AdapterContext`. While these are somewhat linked, I think it would be valuable for the comment here to focus specifically on the relation to the `unmake_current` call.

Perhaps something like:
```rs
// The GL context must be current when dropping `glow::Context`. Before `unmake_current`, wrap in
// `ManuallyDrop` to prevent unexpected panics from dropping this while the context is not current.
```

MarijnS95: @Imberflur we already discussed in https://github.com/gfx-rs/wgpu/pull/6114#discussion_r1733498062 that this comment is going to change.

wgpu-hal/src/gles/wgl.rs
MarijnS95: We definitely request it via Glutin and get one:

```console
$ RUST_LOG=trace cargo r --example raw-gles
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.41s
     Running `target\debug\examples\raw-gles.exe`
Initializing external GL context
Picked a config with 16 samples
Hooking up to wgpu-hal
[2024-08-29T21:19:06Z DEBUG wgpu_hal::gles::adapter] Vendor: Intel
[2024-08-29T21:19:06Z DEBUG wgpu_hal::gles::adapter] Renderer: Intel(R) Arc(TM) A770 Graphics
[2024-08-29T21:19:06Z DEBUG wgpu_hal::gles::adapter] Version: OpenGL ES 3.0 - Build xxxx
[2024-08-29T21:19:06Z DEBUG wgpu_hal::gles::adapter] SL version: OpenGL ES GLSL ES 3.00 - Build xxxx
[2024-08-29T21:19:06Z DEBUG wgpu_hal::gles::adapter] Supported GL Extensions: {
...
}
```

wgpu-hal/src/gles/egl.rs
Imberflur: Ah okay! I wasn't sure if that was still being done so I added this comment 😅

wgpu-hal/src/gles/wgl.rs
MarijnS95: Making WGL look more like EGL is mostly an annoying refactor, but I don't see any reason why it cannot be done or why it wasn't like this in the first place.

wgpu-hal/src/gles/wgl.rs
Imberflur: I see. I guess the WGL backend can use either type? Since iiuc the non-external WGL one isn't an ES context.

wgpu-hal/src/gles/wgl.rs
ErichDonGubler: @MarijnS95: Is this matter resolved? I wish I understood it well enough to tell immediately.

MarijnS95: @ErichDonGubler it's working as-is but the two backend implementations are unnecessarily hard to compare.

MarijnS95: In other words, I'm not sure if I should strive to bring any consistency/improvement here (mainly: seeing why different parts are behind or outside the `Mutex<>` in either backend) but that could/should strictly go in a separate followup PR.

ErichDonGubler: If it really is just a refactor to make things "nice" outstanding, I'm inclined to review and (when we get to that point) merge, then free up resources for the refactor in everybody's brains later.

wgpu-hal/src/gles/wgl.rs
```diff
@@ -565,6 +582,43 @@ impl crate::Instance for Instance {
...
```
ErichDonGubler, line 594: When designing `unsafe` APIs, there are two rules we should uphold:

1. Code inside `unsafe` has special considerations, and `unsafe` must be required _at any site where calling code can cause unsafety, if they do not follow the required safety con_.
2. Safe code (i.e., code _not_ in an `unsafe` block) should otherwise be able to rely on the Rust compiler to ensure safety; safe code should not need to be concerned with upholding safety invariants.

These are otherwise known as the Soundness Property:

> No matter what, Safe Rust can't cause Undefined Behavior.

See also the Rustonomicon's pages [Meet Safe and Unsafe](https://doc.rust-lang.org/nomicon/meet-safe-and-unsafe.html#meet-safe-and-unsafe) (where the Soundness Property is formally introduced) and [How Safe and Unsafe Interact](https://doc.rust-lang.org/nomicon/safe-unsafe-meaning.html#how-safe-and-unsafe-interact).

---

issue: This is not an `unsafe` API design that respects the Soundness Property, and that's not something (IMO) we can ship. The first condition seems to fit the rules of the Soundness Property. However:

- The second condition here:

	> ```rust
	>     /// - The underlying OpenGL ES context must be current when interfacing with any objects returned by
	>     ///   wgpu-hal from this adapter.
	> ```

    …implies safety conditions to uphold in subsequent calls, rather than this one. An `unsafe` API design that adheres to rule (2), then, should make it `unsafe` to perform these subsequent calls, rather than impose this obligation here.
- The third condition:

	> ```
	>    /// - The underlying OpenGL ES context must be current when dropping this adapter and when
	>    ///   dropping any objects returned from this adapter.
	> ```

	…imposes a safety condition on `Drop`, which is a safe API. This violates rule (2); we need a different (`unsafe`) API for safely dropping this context, so users can't write safe Rust that might violate this condition.

ErichDonGubler (changes requested):

wgpu-hal/src/gles/wgl.rs
MarijnS95: > and that's not something (IMO) we can ship

EGL shipped with exactly this though?

As said before I'm merely clearing up some platform parity and postponing any larger refactors of the existing broken and UB GLES implementations for WGL and EGL. But will gladly close this if you think the above is acceptable for EGL but not for WGL and move on.

MarijnS95: @ErichDonGubler note that this is why I suggested to use `wglShareLists()` (and the EGL equivalent) before (like what GStreamer does) to make sure the contexts can be currented independently and on separate threads while sharing resources such as textures and surfaces for rendering.

ErichDonGubler: Ah, if that's the case, then let's file an issue and sort it out later. 👍🏻 I hadn't realized that we'd already shipped some APIs that don't respect the Soundness Property. I'll file a follow-up issue, unless there's one I'm not aware of you're already working against?

ErichDonGubler (approved):
Since it [seems we've already shipped `unsafe` APIs that conflict with the Soundness Property](https://github.com/gfx-rs/wgpu/pull/6152#discussion_r1743946280), I'm willing to defer reconciliation with it until after this PR.

makes my teeth chatter tho 😬

wgpu-hal/src/gles/wgl.rs
MarijnS95: No open issue as far as I'm aware. I may have mentioned this while reviewing #6114 which tried to work around UB.

MarijnS95:
> makes my teeth chatter tho 😬

I was similarly frightened when opening this code for the first time, hence brought up `glutin` in Matrix. That doesn't solve this issue though, as it only concerns itself with the context API but not anything that's subsequently loaded/executed on it (which is deferred to `glow` or crates from `gl-rs`).

wgpu-hal/src/gles/wgl.rs
ErichDonGubler: Filed <https://github.com/gfx-rs/wgpu/issues/6211>.

PR #6161, opened 2024-08-26
Vulkan: create texture from d3d11 shared handle

xiaopengli89:
**Description**
[VK_KHR_external_memory_win32](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_external_memory_win32.html) extension enables application to import Vulkan memory objects from Windows handles.

**Testing**
1. Create D3D11 texture with [D3D11_RESOURCE_MISC_SHARED_NTHANDLE](https://learn.microsoft.com/en-us/windows/desktop/api/d3d11/ne-d3d11-d3d11_resource_misc_flag) and [D3D11_RESOURCE_MISC_SHARED_KEYEDMUTEX](https://learn.microsoft.com/en-us/windows/desktop/api/d3d11/ne-d3d11-d3d11_resource_misc_flag) flags
2. Create shared handle with [IDXGIResource1::CreateSharedHandle](https://learn.microsoft.com/en-us/windows/desktop/api/dxgi1_2/nf-dxgi1_2-idxgiresource1-createsharedhandle)
3. Create vulkan texture with `texture_from_d3d11_shared_handle`

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-hal/src/vulkan/device.rs
+++ b/wgpu-hal/src/vulkan/device.rs
@@ -686,6 +686,7 @@ impl super::Device {
         super::Texture {
             raw: vk_image,
             drop_guard,
+            external_memory: None,
             block: None,
             usage: desc.usage,
             format: desc.format,
@@ -693,6 +694,127 @@ impl super::Device {
             copy_size: desc.copy_extent(),
             view_formats,
         }
+    }
+
+    #[cfg(windows)]
+    unsafe fn find_memory_type_index(
+        &self,
+        type_bits: u32,
+        flags: vk::MemoryPropertyFlags,
+    ) -> Option<usize> {
+        let mem_properties = self
+            .shared
+            .instance
+            .raw
+            .get_physical_device_memory_properties(self.shared.physical_device);
+
+        for (i, mem_ty) in mem_properties
+            .memory_types_as_slice()
+            .into_iter()
+            .enumerate()
+        {
+            if type_bits & (1 << i) != 0 && mem_ty.property_flags & flags == flags {
+                return Some(i);
+            }
+        }
+
+        None
+    }
+
+    /// # Safety
+    ///
+    #[cfg(windows)]
+    pub unsafe fn texture_from_d3d11_shared_handle(
+        &self,
+        d3d11_shared_handle: vk::HANDLE,
+        desc: &crate::TextureDescriptor,
+    ) -> ash::prelude::VkResult<super::Texture> {
+        let copy_size = desc.copy_extent();
+
+        let mut raw_flags = vk::ImageCreateFlags::empty();
+        if desc.is_cube_compatible() {
+            raw_flags |= vk::ImageCreateFlags::CUBE_COMPATIBLE;
+        }
+
+        let original_format = self.shared.private_caps.map_texture_format(desc.format);
+        let mut vk_view_formats = vec![];
+        let mut wgt_view_formats = vec![];
+        if !desc.view_formats.is_empty() {
+            raw_flags |= vk::ImageCreateFlags::MUTABLE_FORMAT;
+            wgt_view_formats.clone_from(&desc.view_formats);
+            wgt_view_formats.push(desc.format);
+
+            if self.shared.private_caps.image_format_list {
+                vk_view_formats = desc
+                    .view_formats
+                    .iter()
+                    .map(|f| self.shared.private_caps.map_texture_format(*f))
+                    .collect();
+                vk_view_formats.push(original_format)
+            }
+        }
+        if desc.format.is_multi_planar_format() {
+            raw_flags |= vk::ImageCreateFlags::MUTABLE_FORMAT;
+        }
+
+        let mut vk_info = vk::ImageCreateInfo::default()
+            .flags(raw_flags)
+            .image_type(conv::map_texture_dimension(desc.dimension))
+            .format(original_format)
+            .extent(conv::map_copy_extent(&copy_size))
+            .mip_levels(desc.mip_level_count)
+            .array_layers(desc.array_layer_count())
+            .samples(vk::SampleCountFlags::from_raw(desc.sample_count))
+            .tiling(vk::ImageTiling::OPTIMAL)
+            .usage(conv::map_texture_usage(desc.usage))
+            .sharing_mode(vk::SharingMode::EXCLUSIVE)
+            .initial_layout(vk::ImageLayout::UNDEFINED);
+
+        let mut format_list_info = vk::ImageFormatListCreateInfo::default();
+        if !vk_view_formats.is_empty() {
+            format_list_info = format_list_info.view_formats(&vk_view_formats);
+            vk_info = vk_info.push_next(&mut format_list_info);
+        }
+
+        let mut external_memory_image_info = vk::ExternalMemoryImageCreateInfo::default()
+            .handle_types(vk::ExternalMemoryHandleTypeFlags::D3D11_TEXTURE_KMT);
+        vk_info = vk_info.push_next(&mut external_memory_image_info);
+
+        let raw = unsafe { self.shared.raw.create_image(&vk_info, None)? };
+        let req = unsafe { self.shared.raw.get_image_memory_requirements(raw) };
+
+        let mut import_memory_info = vk::ImportMemoryWin32HandleInfoKHR::default()
+            .handle_type(vk::ExternalMemoryHandleTypeFlags::D3D11_TEXTURE_KMT)
+            .handle(d3d11_shared_handle);
+
+        let Some(mem_type_index) = self
+            .find_memory_type_index(req.memory_type_bits, vk::MemoryPropertyFlags::DEVICE_LOCAL)
+        else {
+            return Err(vk::Result::ERROR_UNKNOWN);
+        };
+
+        let memory_allocate_info = vk::MemoryAllocateInfo::default()
+            .allocation_size(req.size)
+            .memory_type_index(mem_type_index as _)
+            .push_next(&mut import_memory_info);
+        let memory = unsafe {
+            self.shared
+                .raw
+                .allocate_memory(&memory_allocate_info, None)?
+        };
+        unsafe { self.shared.raw.bind_image_memory(raw, memory, 0)? };
+
+        Ok(super::Texture {
+            raw,
+            drop_guard: None,
+            external_memory: Some(memory),
+            block: None,
+            usage: desc.usage,
+            format: desc.format,
+            raw_flags,
+            copy_size,
+            view_formats: wgt_view_formats,
+        })
     }
 
     /// # Safety
@@ -1118,6 +1240,7 @@ impl crate::Device for super::Device {
         Ok(super::Texture {
             raw,
             drop_guard: None,
+            external_memory: None,
             block: Some(block),
             usage: desc.usage,
             format: desc.format,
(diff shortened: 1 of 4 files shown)


wgpu-hal/src/vulkan/device.rs
```diff
@@ -685,6 +686,176 @@ impl super::Device {
...
+            self.create_image_without_memory(desc, Some(&mut external_memory_image_info))?;
+
+        let mut import_memory_info = vk::ImportMemoryWin32HandleInfoKHR::default()
+            .handle_type(vk::ExternalMemoryHandleTypeFlags::D3D11_TEXTURE_KMT)
+            .handle(d3d11_shared_handle as _);
+
+        let Some(mem_type_index) = self
+            .find_memory_type_index(req.memory_type_bits, vk::MemoryPropertyFlags::DEVICE_LOCAL)
+        else {
+            return Err(crate::DeviceError::ResourceCreationFailed);
```
MarijnS95: Might `.ok_or(..)?;` be more concise here?

wgpu-hal/src/vulkan/device.rs
```diff
@@ -685,6 +686,176 @@ impl super::Device {
...
+        &self,
+        d3d11_shared_handle: *mut std::ffi::c_void,
+        desc: &crate::TextureDescriptor,
+    ) -> Result<super::Texture, crate::DeviceError> {
+        if !self.shared.private_caps.external_memory_win32 {
+            return Err(crate::DeviceError::ResourceCreationFailed);
+        }
+
+        let mut external_memory_image_info = vk::ExternalMemoryImageCreateInfo::default()
+            .handle_types(vk::ExternalMemoryHandleTypeFlags::D3D11_TEXTURE_KMT);
```
MarijnS95: You should document in the doc-comment that this uses `_KMT` which has different semantics from the non-`_KMT` version.

wgpu-hal/src/vulkan/device.rs
```diff
@@ -685,6 +686,176 @@ impl super::Device {
...
+
+        let memory_allocate_info = vk::MemoryAllocateInfo::default()
+            .allocation_size(req.size)
+            .memory_type_index(mem_type_index as _)
+            .push_next(&mut import_memory_info);
+        let memory = unsafe {
+            self.shared
+                .raw
+                .allocate_memory(&memory_allocate_info, None)
+                .map_err(super::map_host_device_oom_err)?
```
MarijnS95: This seems to be a general mistake in the Vulkan backend, but I'd love to see the `.map_err(...)?` part outside of the `unsafe {}` block (and in turn some lines my become more concise too).

wgpu-hal/src/vulkan/device.rs
```diff
@@ -685,6 +686,176 @@ impl super::Device {
...
+        };
+
+        unsafe {
+            self.shared
+                .raw
+                .bind_image_memory(raw, memory, 0)
```
MarijnS95: This is probably the case since we've got a handle to a texture and not some heap that contains a texture (probably not even a thing back in D3D11), but just making sure that we should always bind it at offset `0`?

wgpu-hal/src/vulkan/device.rs
```diff
@@ -685,6 +686,176 @@ impl super::Device {
...
+            self.shared
+                .raw
+                .create_image(&vk_info, None)
+                .map_err(map_err)?
+        };
+        fn map_err(err: vk::Result) -> crate::DeviceError {
+            // We don't use VK_EXT_image_compression_control
+            // VK_ERROR_COMPRESSION_EXHAUSTED_EXT
+            super::map_host_device_oom_and_ioca_err(err)
+        }
```
MarijnS95: ~Inline and have the comment above the `.map_err()` perhaps?~ EDIT: This was original moved code.

wgpu-hal/src/vulkan/device.rs
xiaopengli89: Yes, the memory only contains this texture.

wgpu-hal/src/vulkan/device.rs
xiaopengli89: Thanks for your reminder, we should use the non-`_KMT` version here.

wgpu-hal/src/vulkan/device.rs
```diff
@@ -685,6 +686,162 @@ impl super::Device {
...
+    ) -> Option<usize> {
+        let mem_properties = unsafe {
+            self.shared
+                .instance
+                .raw
+                .get_physical_device_memory_properties(self.shared.physical_device)
+        };
+
+        for (i, mem_ty) in mem_properties.memory_types_as_slice().iter().enumerate() {
+            if type_bits & (1 << i) != 0 && mem_ty.property_flags & flags == flags {
```
cwfitzgerald: Could we break up this expression into separate booleans with descriptive names/comments? I have no idea what this is doing just by looking at it.

wgpu-hal/src/vulkan/device.rs
```diff
@@ -685,6 +686,162 @@ impl super::Device {
...
+        desc: &crate::TextureDescriptor,
+        external_memory_image_create_info: Option<&mut vk::ExternalMemoryImageCreateInfo>,
+    ) -> Result<
+        (
+            vk::Image,
+            vk::MemoryRequirements,
+            crate::CopyExtent,
+            Vec<wgt::TextureFormat>,
+            vk::ImageCreateFlags,
+        ),
```
cwfitzgerald: Lets define a struct for this information

wgpu-hal/src/vulkan/device.rs
```diff
@@ -685,6 +686,162 @@ impl super::Device {
...
+    /// - The `d3d11_shared_handle` must be valid and respecting `desc`
+    /// - `VK_EXTERNAL_MEMORY_HANDLE_TYPE_D3D11_TEXTURE_BIT` flag is used because we need to hold a reference to the handle
+    #[cfg(windows)]
+    pub unsafe fn texture_from_d3d11_shared_handle(
+        &self,
+        d3d11_shared_handle: *mut std::ffi::c_void,
+        desc: &crate::TextureDescriptor,
+    ) -> Result<super::Texture, crate::DeviceError> {
+        if !self.shared.private_caps.external_memory_win32 {
+            return Err(crate::DeviceError::ResourceCreationFailed);
```
cwfitzgerald, line 811: Lets log an error here

wgpu-hal/src/vulkan/device.rs
```diff
@@ -685,6 +686,162 @@ impl super::Device {
...
+
+    /// # Safety
+    ///
+    /// - Vulkan 1.1+ (or VK_KHR_external_memory)
+    /// - The `d3d11_shared_handle` must be valid and respecting `desc`
+    /// - `VK_EXTERNAL_MEMORY_HANDLE_TYPE_D3D11_TEXTURE_BIT` flag is used because we need to hold a reference to the handle
+    #[cfg(windows)]
+    pub unsafe fn texture_from_d3d11_shared_handle(
+        &self,
+        d3d11_shared_handle: *mut std::ffi::c_void,
```
cwfitzgerald: My preference would be that this would be a strongly typed https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Foundation/struct.HANDLE.html, but it's probably not worth the extra dependency, especially as pulling out a c_void from it is so easy.

cwfitzgerald (changes requested):
Looking good! Some comments

wgpu-hal/src/vulkan/device.rs
xiaopengli89: I referred to the `findProperties` function of https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceMemoryProperties.html

wgpu-hal/src/vulkan/device.rs
xiaopengli89: Done.

wgpu-hal/src/vulkan/device.rs
xiaopengli89: Done.

xiaopengli89:
@cwfitzgerald Does this PR need any other improvements? This feature is already working properly in our product.

ErichDonGubler:
@cwfitzgerald ☝🏻

Tsuguri:
Hello. I've been working on sharing textures with wgpu using both Dx12 and Vulkan but without doing any changes to wgpu itself.  My memory allocation code ended up looking very similar with important difference: `vk::ExternalMemoryHandleTypeFlags::D3D12_RESOURCE` instead of `vk::ExternalMemoryHandleTypeFlags::D3D11_TEXTURE`. Other types could potentially be also usable for storing textures. If possible I would encourage to add it as a parameter next to d3d11_shared_handle.

In fact this new method is called `texture_from_d3d11_shared_handle` which makes my comment a little misplaced. Still, adding this parameter would make this code more usable.

Change to adapter to add `external_memory_win32` capabilities is very nice - otherwise one has to extract raw adapter from wgpu, copy code from [here](https://github.com/gfx-rs/wgpu/blob/v22/wgpu-hal/src/vulkan/adapter.rs#L1961), add required enabled_extensions to the list and then inject it back with ```adapter .create_device_from_hal``` which is a pain.

Other considerations: If I'm not wrong, NT handle holds the resource alive and it won't be cleaned up unless you call CloseHandle. I may be wrong here but I think Vulkan will NOT close the handle itself so user should do that manually. In my specific situation (sending textures between processes) I'm managing handles outside of wgpu - I remove injected shared textures from wgpu and it gets sent to different process. I think that this ownership situation should be reflected in documentation of this method (but please don't quote me on that, I'm not 100% sure how exactly this mechanism works after reading documentation and my limited tests).

AdrianEddy:
I too have the interop code [here](https://github.com/gyroflow/gyroflow/blob/master/src/core/gpu/wgpu_interop_directx.rs) if that's helpful to anyone. One thing I had to do is copy the texture format (DXGI_FORMAT) conversion functions, it would be nice if they were exposed from wgpu

cwfitzgerald:
Sorry my time basically disappeared suddenly - looking now.

wgpu-hal/src/vulkan/device.rs
```diff
@@ -1028,74 +1195,22 @@ impl crate::Device for super::Device {
...
-            super::map_host_device_oom_and_ioca_err(err)
-        }
-        let req = unsafe { self.shared.raw.get_image_memory_requirements(raw) };
+        let ImageWithoutMemory {
+            raw,
+            requirements,
+            copy_size,
+            view_formats,
+            raw_flags,
+        } = self.create_image_without_memory(desc, None)?;
```
cwfitzgerald: Nit: I think it would be cleaner if we don't destructure this in either place. And call this `image`

wgpu-hal/src/vulkan/device.rs
```diff
@@ -695,6 +696,172 @@ impl super::Device {
...
+        let mut external_memory_image_info = vk::ExternalMemoryImageCreateInfo::default()
+            .handle_types(vk::ExternalMemoryHandleTypeFlags::D3D11_TEXTURE);
+
+        let ImageWithoutMemory {
+            raw,
+            requirements,
+            copy_size,
+            view_formats,
+            raw_flags,
+        } = self.create_image_without_memory(desc, Some(&mut external_memory_image_info))?;
```
cwfitzgerald: Here

cwfitzgerald (approved):
Approving with nits

xiaopengli89:
@Tsuguri `VK_EXTERNAL_MEMORY_HANDLE_TYPE_D3D11_TEXTURE_BIT` owns a reference to the memory used by the Direct3D resource, after importing the handle can be released safely.

xiaopengli89:
@MarijnS95 Any other suggestions?

xiaopengli89:
Can this PR be merged before #6465?

PR #6164, opened 2024-08-26
[wgpu-hal] Change the `DropGuard` API in Vulkan and GLES to a callback-based one.

### LINKED ISSUE #6142: [wgpu-hal][Vulkan] Possibly changing the difference in behavior of `drop_guard` between `Instance::from_raw` and `Device::texture_from_raw`
When talking about `Instance`, `Adapter` and `Device` i mean `wgpu_hal::vulkan::{Instance, Adapter, Device}`

Currently, both `Instance::from_raw` and `Device::texture_from_raw` accept a `drop_guard` as a parameter. The behavior of the drop guard is completely different though.

When a texture is created with a drop guard, `wgpu_hal` **will not** destroy the underlying vulkan image when dropping the texture. It will destroy the drop guard though, thus signalling that wgpu is not using the resource anymore. This is what I expected would happen in both cases.

When an instance is created with a drop guard, the behavior is reversed: if the drop guard is present, wgpu will destory the drop guard **and** the underlying vulkan instance, and when the drop guard is absent the instance will not be destroyed. This is not intuitive to me at all, it's hard for me to find a use case for that behavior.

Additionally, the function `Adapter::device_from_raw` does not take a drop guard, but accepts a `handle_is_owned` bool, which controlls whether wgpu will destroy the underlying vulkan device when dropping the created `Device`.

Ideally, I would like all three of these functions to share the same behavior: the one the `Device::texture_from_raw` has. I'm working on a project which needs to initialize vulkan by itself and I would like to provide my API clients access to a wgpu device created with the same underlying vulkan device. The API inconsistency makes this difficult.

I looked at the PR that introduced these behaviors: #1850. It seems these are tailored to one specific use case mentioned in the PR description.

Is it possible for this to be changed? I'm happy to work on this myself if the change is possible.

jerzywilczek:
This patch also unifies the behavior of the the `DropGuard`-ish constructs in the Vulkan and GLES implementations.

**Connections**
A change discussed in #6142. Closes #6142.

**Description**
Previously, when creating some wgpu structs from raw Vulkan or GLES counterparts, it was sometimes possible to provide a `DropGuard`, which would signal when the wgpu struct was dropped and the underlying structure could be cleaned up. Unfortunately, the behavior was very inconsistent across different `*_from_raw` functions. This patch unifies this behavior switching to a callback-based approach.

**Testing**
Things still compile and there is no difference in test passes.

**Checklist**

- [x] Run `cargo fmt`.
    This just fails with many errors
  41 tests fail on my PC, but they also fail without my changes.

### DIFF
--- a/wgpu-hal/src/gles/device.rs
+++ b/wgpu-hal/src/gles/device.rs
@@ -110,22 +110,21 @@ impl super::Device {
     ///
     /// - `name` must be created respecting `desc`
     /// - `name` must be a texture
-    /// - If `drop_guard` is [`None`], wgpu-hal will take ownership of the texture. If `drop_guard` is
-    ///   [`Some`], the texture must be valid until the drop implementation
-    ///   of the drop guard is called.
+    /// - If `drop_callback` is [`None`], wgpu-hal will take ownership of the texture. If
+    ///   `drop_callback` is [`Some`], the texture must be valid until the callback is called.
     #[cfg(any(native, Emscripten))]
     pub unsafe fn texture_from_raw(
         &self,
         name: std::num::NonZeroU32,
         desc: &crate::TextureDescriptor,
-        drop_guard: Option<crate::DropGuard>,
+        drop_callback: Option<crate::DropCallback>,
     ) -> super::Texture {
         super::Texture {
             inner: super::TextureInner::Texture {
                 raw: glow::NativeTexture(name),
                 target: super::Texture::get_info_from_desc(desc),
             },
-            drop_guard,
+            drop_guard: crate::DropGuard::from_option(drop_callback),
             mip_level_count: desc.mip_level_count,
             array_layer_count: desc.array_layer_count(),
             format: desc.format,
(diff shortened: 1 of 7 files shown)


wgpu-hal/src/gles/device.rs
```diff
@@ -110,22 +110,22 @@ impl super::Device {
     ///
     /// - `name` must be created respecting `desc`
     /// - `name` must be a texture
-    /// - If `drop_guard` is [`None`], wgpu-hal will take ownership of the texture. If `drop_guard` is
-    ///   [`Some`], the texture must be valid until the drop implementation
+    /// - If `drop_callback` is [`None`], wgpu-hal will take ownership of the texture. If
+    ///   `drop_callback` is [`Some`], the texture must be valid until the drop implementation
```
cwfitzgerald: This comment is out of date

wgpu-hal/src/vulkan/adapter.rs
```diff
@@ -1812,12 +1814,14 @@ impl super::Adapter {
             0, 0, 0, 0,
         ];
 
+        let drop_guard = drop_callback.map(|callback| crate::DropGuard { callback });
```
cwfitzgerald: Might be worth a crate::DropGuard::from_option which takes an Option<DropCallback> and returns Option<DropGuard> to make this code a bit simpler

cwfitzgerald (changes requested):
Looking good, small comments

wgpu-hal/src/gles/device.rs
jerzywilczek: I don't really understand why or what to change it to ;(

jerzywilczek: can you elaborate?

cwfitzgerald: It refers to the drop implementation of the drop guard, as opposed to the new callback api

jerzywilczek: Oh, sure, I missed that

cwfitzgerald (approved):
Nice!

ErichDonGubler:
Pushed c2563dc8ca8da28d9093bbb71850be66687f3015 to the impending squash-merge so that CI passes and actually does the merge.

PR #5325, opened 2024-03-01
Vendor WebGPU Bindings from web_sys

cwfitzgerald:
**Connections**

Closes https://github.com/gfx-rs/wgpu/issues/5258
Closes https://github.com/gfx-rs/wgpu/pull/5224

**Description**

This PR:
- adds a script to our xtask which automatically vendors the web_sys bindings from a particular checkout of wasm-bindgen.
- converts our webgpu bindings to those bindings.
- vastly simplifes the xtask code

We _do not_ use it public-ally, so this is not a breaking change. I plan to backport this.

cc @daxpedda 

**Testing**

Manual testing of WebGPU, CI, and building manually.

PR #6170, opened 2024-08-27
Add `VideoFrame` to `ExternalImageSource` enum

### LINKED ISSUE #6167: Add `ExternalImageSource::VideoFrame`
WebCodecs API introduces `VideoFrame` (https://w3c.github.io/webcodecs/#videoframe), which is a valid `ExternalImageSource` in both WebGL (`WebGLRenderingContext.texImage2D`) and WebGPU (`GPUQueue.copyExternalImageToTexture`) contexts.

jprochazk:
**Connections**
* Closes https://github.com/gfx-rs/wgpu/issues/6167

**Description**
Adds missing `VideoFrame` type as a possible external image source.

The enum variant is gated behind `cfg(web_sys_unstable_apis)`, as `web_sys::VideoFrame` is unavailable without it.

**Testing**
Tested manually by using it in place of this hack: https://github.com/rerun-io/rerun/pull/7261/files#diff-213f45076b49c7bbe3d7a1f0c7383b3b9e0edb4c02c21bab32709573155a2e63R403-R414
It would be nice to have a Wasm-only example that renders a media stream using `VideoDecoder`, but I'm not sure how adding such an example would work.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,7 @@ By @teoxoy [#6134](https://github.com/gfx-rs/wgpu/pull/6134).
 - Deduplicate bind group layouts that are created from pipelines with "auto" layouts. By @teoxoy [#6049](https://github.com/gfx-rs/wgpu/pull/6049)
 - Fix crash when dropping the surface after the device. By @wumpf in [#6052](https://github.com/gfx-rs/wgpu/pull/6052)
 - Fix error message that is thrown in create_render_pass to no longer say `compute_pass`. By @matthew-wong1 [#6041](https://github.com/gfx-rs/wgpu/pull/6041)
+- Add `VideoFrame` to `ExternalImageSource` enum. By @jprochazk in [#6170](https://github.com/gfx-rs/wgpu/pull/6170)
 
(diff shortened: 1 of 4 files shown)


jprochazk:
I'm not sure what the policy is for unstable APIs, but this requires `--cfg web_sys_unstable_apis`. I gated the new enum variant behind that cfg, so that `wgpu-types` continues to build without it. Using `VideoFrame` at all requires using that cfg anyway, so hopefully it's fine!

cwfitzgerald (approved):
Perfect, thanks!

emilk:
I thought `web_sys_unstable_apis` was removed in https://github.com/gfx-rs/wgpu/pull/5325 🤔

jprochazk:
It was removed, but this PR doesn't invalidate that one, because `wgpu` continues to build without requiring the cfg to be set. But if you want `web_sys::VideoFrame` to be accessible (in any crate, not just `wgpu`), you need to set `--cfg web_sys_unstable_apis`. So to satisfy both constraints, the enum variant added here is gated behind the same cfg.

PR #6171, opened 2024-08-27
Accept only `vec3` (not `vecN`) for the `cross` built-in

### LINKED ISSUE #6153: Vec4 cross product does not check types and crashes the program
**Description**
When computing a cross product (with the cross() function) in a shader with invalid types (e.g two vec4<f32>) the program crashes with error code 0xC000041D

**Repro steps**
Adding the following code make the program crash silently
```wgsl
let a = cross(vec4(0., 1., 2., 3.), vec4(0., 1., 2., 3.));
```

There should be a type error, as cross() only works with vec3

**Extra materials**
 I coundn't get any error/logs other than a global crash

**Platform**
Windows with vulkan.

(this is my first github issue, tell me if im not in the right place)

ErichDonGubler:
**Connections**

Resolves #6153.

**Description**

Previously, we only checked that a single type of vector of floats was being used as input for the `cross` built-in. Only `vec3` should be permitted; so, narrow our checking of the vector type to 3 elements.

**Testing**

Tests have been added to ensure coverage of what we should and should not accept for the `cross` built-in.

Open questions:

- [x] ~~Do we need to permit more than just `vec3` from language frontends other than WGSL?~~ Nope!

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,10 @@ By @teoxoy [#6134](https://github.com/gfx-rs/wgpu/pull/6134).
 
 - Fix incorrect hlsl image output type conversion. By @atlv24 in [#6123](https://github.com/gfx-rs/wgpu/pull/6123)
 
+#### Naga
+
+- Accept only `vec3` (not `vecN`) for the `cross` built-in. By @ErichDonGubler in [#6171](https://github.com/gfx-rs/wgpu/pull/6171).
+
 #### General
 
 - If GL context creation fails retry with GLES. By @Rapdorian in [#5996](https://github.com/gfx-rs/wgpu/pull/5996)
--- a/naga/src/valid/expression.rs
+++ b/naga/src/valid/expression.rs
@@ -1161,7 +1161,7 @@ impl super::Validator {
                             ));
                         }
                     }
-                    Mf::Outer | Mf::Cross | Mf::Reflect => {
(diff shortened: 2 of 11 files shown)


teoxoy (approved):

ErichDonGubler:
@teoxoy: Thoughts on this question in the OP?

>  Do we need to permit more than just `vec3` from language frontends other than WGSL?

teoxoy:
I don't know, I'd have to check.

teoxoy:
GLSL and SPIR-V say they must be `vec3`s as well.

ErichDonGubler:
Ah, yep, all these sources state only accepting 3-component vectors:

* HLSL: https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-cross
* SPIR-V: <https://registry.khronos.org/SPIR-V/specs/unified1/GLSL.std.450.html>, <kbd>Ctrl</kbd> + <kbd>F</kdb> for `cross`.
* MSL: <https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf>, `Table 6.11. Geometric functions in the Metal standard library`, `cross` entry

There seems to be some divergence in the float types accepted on these platforms, but that's orthogonal to this PR. 😀 :shipit:

Thanks to @teoxoy for onboarding me onto the language spec. resources I cited above. 🍻

PR #6174, opened 2024-08-28
fix: compile `raw-gles` example again after #6150

ErichDonGubler:
**Connections**

I think this was regressed by #6150?

**Description**

See above.

**Testing**

It compiles now!

**Checklist**

- [x] Run `cargo fmt`.

    Probably not necessary? Might want to add on to the `glutin` `CHANGELOG` entry. 🤔

### DIFF
--- a/wgpu-hal/examples/raw-gles.rs
+++ b/wgpu-hal/examples/raw-gles.rs
@@ -71,7 +71,9 @@ fn main() {
     // Glutin tries to create an OpenGL context by default.  Force it to use any version of GLES.
     let context_attributes = glutin::context::ContextAttributesBuilder::new()
         // WGPU expects GLES 3.0+.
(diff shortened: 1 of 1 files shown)


ErichDonGubler:
@MarijnS95 ☝🏻

MarijnS95:
Doesn't your CI catch this?

Seems like I squashed this into the wrong PR (https://github.com/gfx-rs/wgpu/pull/6152/files#diff-ac2020a9d8bd724543c3bf9ff45b699d1ae9b15a9f6d77c38a604c7ff73cb716R19) as I was rebasing this from a Windows machine, which isn't supported in #6150 yet but only starting with #6152.

ErichDonGubler:
@MarijnS95: I'm not sure why our CI doesn't catch it. @cwfitzgerald, is there a reason CI doesn't at least try to compile examples?

PR #6178, opened 2024-08-28
[wgpu-core] when mapping buffers for reading, mark buffers as initialized only when they have `MAP_WRITE` usage

### LINKED ISSUE #6137: `wgpu_hal` tries to call `glFlushMappedBufferRange` on buffer without setting `GL_MAP_FLUSH_EXPLICIT_BIT`
Simple programs using the `gles` backend can get errors like this:

```
[... ERROR wgpu_hal::gles] GLES: [API/Error] ID 2 : GL_INVALID_OPERATION in glFlushMappedBufferRange(GL_MAP_FLUSH_EXPLICIT_BIT not set)
```

In GLES 3.2, calling `glFlushMappedBufferRange` is only allowed if the buffer was previously mapped with the `GL_MAP_FLUSH_EXPLICIT` flag set. [doc](https://registry.khronos.org/OpenGL-Refpages/es3/html/glFlushMappedBufferRange.xhtml) But `wgpu_hal` as driven by `wgpu_core` violates this rule.

In `wgpu_hal`, the code in `<gles::Device as Device>::create_buffer` sets `GL_MAP_FLUSH_EXPLICIT_BIT` only if `BufferUses::MAP_WRITE` is set. But when `wgpu_core::device::map_buffer` is presented with a `HostMap::Read` mapping of a buffer with uninitialized ranges, it calls `wgpu_hal::Device::flush_mapped_ranges`, which calls `glFlushMappedBufferRange` even though the buffer was created only with `BufferUses::MAP_READ`.

teoxoy:
Fixes https://github.com/gfx-rs/wgpu/issues/6137.
I couldn't test this locally as my buffer mappings seem to be coherent but the new `wgpu_test::dispatch_workgroups_indirect::reset_bind_groups` testcase in https://github.com/gfx-rs/wgpu/pull/5714 runs into this.

### DIFF
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -336,19 +336,41 @@ fn map_buffer(
 
     let mapped = unsafe { std::slice::from_raw_parts_mut(mapping.ptr.as_ptr(), size as usize) };
 
-    for uninitialized in buffer
-        .initialization_status
-        .write()
-        .drain(offset..(size + offset))
+    // We can't call flush_mapped_ranges in this case, so we can't drain the uninitialized ranges either
+    if !mapping.is_coherent
+        && kind == HostMap::Read
+        && !buffer.usage.contains(wgt::BufferUsages::MAP_WRITE)
     {
-        // The mapping's pointer is already offset, however we track the
-        // uninitialized range relative to the buffer's start.
-        let fill_range =
-            (uninitialized.start - offset) as usize..(uninitialized.end - offset) as usize;
-        mapped[fill_range].fill(0);
+        for uninitialized in buffer
+            .initialization_status
+            .write()
+            .uninitialized(offset..(size + offset))
+        {
+            // The mapping's pointer is already offset, however we track the
+            // uninitialized range relative to the buffer's start.
+            let fill_range =
+                (uninitialized.start - offset) as usize..(uninitialized.end - offset) as usize;
+            mapped[fill_range].fill(0);
+        }
+    } else {
+        for uninitialized in buffer
+            .initialization_status
+            .write()
+            .drain(offset..(size + offset))
+        {
+            // The mapping's pointer is already offset, however we track the
+            // uninitialized range relative to the buffer's start.
+            let fill_range =
+                (uninitialized.start - offset) as usize..(uninitialized.end - offset) as usize;
+            mapped[fill_range].fill(0);
 
-        if !mapping.is_coherent && kind == HostMap::Read {
-            unsafe { raw.flush_mapped_ranges(raw_buffer, &[uninitialized]) };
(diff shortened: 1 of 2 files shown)


wgpu-hal/src/gles/device.rs
```diff
@@ -603,8 +603,8 @@ impl crate::Device for super::Device {
 
         unsafe { gl.bind_buffer(target, None) };
 
-        if !is_coherent && desc.usage.contains(crate::BufferUses::MAP_WRITE) {
-            map_flags |= glow::MAP_FLUSH_EXPLICIT_BIT;
+        if !is_coherent {
+            map_flags |= glow::MAP_FLUSH_EXPLICIT_BIT | glow::MAP_WRITE_BIT;
```
ErichDonGubler: issue: CI is rejecting this change, though I don't understand why.

ErichDonGubler (changes requested):

wgpu-hal/src/gles/device.rs
ErichDonGubler: question: I don't know hardly anything about GL, so this is probably a dumb question, but: shouldn't setting `MAP_WRITE_BIT` be avoided unless we have `BuffersUses::MAP_WRITE` specified on the buffer?

teoxoy: It's not great but that's the current requirement of the buffer initialization code.

Maybe we should only do it if `MAPPABLE_PRIMARY_BUFFERS` is enabled. If it's not, I think we can avoid flushing altogether as the buffer in question can only have `MAP_READ` & `COPY_DST` usages.

Thoughts? @Wumpf @cwfitzgerald

teoxoy: Actually we need to do this since the GL backend doesn't support `MAPPABLE_PRIMARY_BUFFERS` and now the WebAssembly CI job is failing.

ErichDonGubler: Okay. Well, I'm inclined to trust the collectively invoked judgment of other contributors in this matter; once CI passes, I'm happy. 🤷🏻‍♂️

ErichDonGubler (approved):
Approving to leave the current conversation and CI as the only blockers I intend.

ErichDonGubler:
@teoxoy: If @Wumpf and @cwfitzgerald don't have anything to say by ~Wednesday, I think it'd be fair to just merge, if not earlier.

jimblandy:
I'll run the test case in #6137 against this on my machine.

jimblandy:
This patch does indeed silence the GLES validation error provoked on trunk by the test in #6137.

jimblandy:
I was going to suggest including the test from #6137 in this PR, but the OP says that #5714 will also cover it, so I think that should be good enough. Tests are not product.

PR #6105, opened 2024-08-12
Raise MSRV to Rust 1.80

ErichDonGubler:
- [ ] Depends on Firefox updating to Rust 1.80 or later.
- ~~Depends on #6147.~~
- ~~Depends on #6180.~~
- Part of #6379.

PR #6180, opened 2024-08-28
Yet more preempting of Rust 1.80 lints

ErichDonGubler:
**Connections**

Offloading some diff. complexity from #6105.

**Description**

Fixes more lints that will definitely affect us with Rust upgrades later.

**Testing**

Things still compile, so woot!

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/examples/src/boids/mod.rs
+++ b/examples/src/boids/mod.rs
@@ -2,7 +2,7 @@
 // adapted from https://github.com/austinEng/webgpu-samples/blob/master/src/examples/computeBoids.ts
 
 use nanorand::{Rng, WyRand};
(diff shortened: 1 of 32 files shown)


cwfitzgerald:
What's the problem this is solving?

ErichDonGubler:
@cwfitzgerald:

> What's the problem this is solving?

Individual motivations are already in the OP and individual commits, but to summarize: most of this changes code to add `use std::mem::{size_of, size_of_val}`, which is intended to head off most of the `unused_qualifications` warnings that will be introduced by Rust 1.80 adding them to the prelude. See also #6105.

nical (approved):

PR #6181, opened 2024-08-29
feat(wgsl): add `first` and `either` sampling for `@interpolate(flat, …)`

### LINKED ISSUE #6122: Support `first` and `either` sampling in `@interpolate(flat, …)`
The WebGPU spec. made some changes to WGSL with [`gpuweb`#4639](https://github.com/gpuweb/gpuweb/pull/4639) that we need to handle; it…

- [ ] …added the `first` and `either` interpolation sampling to the `flat` interpolation type, with `first` as the default.
- [ ] …added validation that `perspective` and `linear` interpolation types are constrained to the set of `{center,centroid,sample}`, to exclude the new interpolation sampling options.

ErichDonGubler:
**Connections**

Resolves #6122.

**Description**

Context is in the linked issue.

**Testing**

Added some coverage for both (1) new valid parses as snapshot testing and (2) invalid parses as validation testing.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/naga/src/back/glsl/mod.rs
+++ b/naga/src/back/glsl/mod.rs
@@ -498,6 +498,9 @@ pub enum Error {
     Custom(String),
     #[error("overrides should not be present at this stage")]
     Override,
+    /// [`crate::Sampling::First`] is unsupported.
+    #[error("`{:?}` sampling is unsupported", crate::Sampling::First)]
+    FirstSamplingNotSupported,
 }
 
 /// Binary operation with a different logic on the GLSL side.
@@ -1534,7 +1537,7 @@ impl<'a, W: Write> Writer<'a, W> {
         // here, regardless of the version.
         if let Some(sampling) = sampling {
             if emit_interpolation_and_auxiliary {
-                if let Some(qualifier) = glsl_sampling(sampling) {
+                if let Some(qualifier) = glsl_sampling(sampling)? {
                     write!(self.out, "{qualifier} ")?;
                 }
             }
@@ -4771,14 +4774,15 @@ const fn glsl_interpolation(interpolation: crate::Interpolation) -> &'static str
 }
 
 /// Return the GLSL auxiliary qualifier for the given sampling value.
-const fn glsl_sampling(sampling: crate::Sampling) -> Option<&'static str> {
+const fn glsl_sampling(sampling: crate::Sampling) -> BackendResult<Option<&'static str>> {
     use crate::Sampling as S;
 
-    match sampling {
-        S::Center => None,
+    Ok(match sampling {
+        S::First => return Err(Error::FirstSamplingNotSupported),
+        S::Center | S::Either => None,
         S::Centroid => Some("centroid"),
         S::Sample => Some("sample"),
-    }
+    })
 }
 
 /// Helper function that returns the glsl dimension string of [`ImageDimension`](crate::ImageDimension)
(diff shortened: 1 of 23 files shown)


naga/src/back/glsl/mod.rs
```diff
@@ -4774,7 +4774,7 @@ const fn glsl_sampling(sampling: crate::Sampling) -> Option<&'static str> {
     use crate::Sampling as S;
 
     match sampling {
-        S::Center => None,
+        S::Center | S::First | S::Either => None,
```
teoxoy: We should error if we see `first` in the GLSL backend.

ErichDonGubler: I'm thinking of using a new error variant with this message:

```
`First` sampling is unsupported in the GLSL backend
```

Eventually, we can add a "reason" field to the error message, which we could use to differentiate rejection between this case and using WebGPU compat.

ErichDonGubler: ...oh, but this is a backend context, where we wouldn't be checking for `compat`. Nevermind, then, just gonna add something to `naga::back::glsl::Error`, then. 😅

teoxoy: Yeah, I was thinking of adding a new variant to GLSL's error type. GLSL will always be compat so that's fine.

naga/tests/validation.rs
```diff
@@ -311,3 +311,116 @@ fn main() {{
...
+                        naga::Sampling::Either => "either",
+                        naga::Sampling::Center => "center",
+                        naga::Sampling::Centroid => "centroid",
+                        naga::Sampling::Sample => "sample",
+                    }
+                ),
+            };
+            let member_type = match interpolation {
+                naga::Interpolation::Perspective | naga::Interpolation::Linear => "f32",
+                naga::Interpolation::Flat => "f32",
```
ErichDonGubler: I originally intended to use `u32` for some variety here, but this is `f32` like the other branch. Oops!

ErichDonGubler: Fixed with latest force-push to 0591c9a5.

naga/src/back/glsl/mod.rs
ErichDonGubler: With my latest force push to 0591c9a5, I've added some test coverage to ensure that we're getting this:

* A new `validation::no_flat_first_in_glsl` test that ensures we emit a new `naga::back::glsl::Error::FirstSamplingNotSupported` variant, as we discussed above.
* Removal of the GLSL backend from `tests/in/interpolate.wgsl` snapshots, since `@interpolate(flat, first)` isn't testable in GLSL.
* A new `tests/in/interpolate_compat.wgsl` snapshot test that's like `interpolate.wgsl`, but omits the `@interpolate(flat, first)` test case and _does_ still snapshot GLSL output.

naga/tests/validation.rs
```diff
@@ -311,3 +311,180 @@ fn main() {{
...
+        let module = naga::front::wgsl::parse_str(&source).unwrap();
+
+        Self {
+            source,
+            module,
+            interpolate_attr,
+            entry_point: "main",
+        }
+    }
+}
```
ErichDonGubler, line 490: @teoxoy: I don't think this is too interesting, but just wanted to give you a heads-up that I extracted this `DummyInterpolationShader` API from the now-two tests being introduced with 0591c9a5.

teoxoy (approved):

PR #6188, opened 2024-08-31
[naga] Allow dynamic indexing of array

### LINKED ISSUE #4337: Support dynamic indexing of by-value matrices and arrays, per WGSL

sagudev:
**Connections**
For arrays, at least, this fixes https://github.com/gfx-rs/wgpu/issues/4337, second commit is just https://github.com/gfx-rs/naga/pull/723 ported to trunk.

**Description**
See commits and linked PR.

**Testing**
There are some new tests and I did CTS run in servo.

**Checklist**

- [x] Run `cargo fmt`.


sagudev:
Servo's CTS run: https://github.com/sagudev/servo/actions/runs/10656885185/job/29537501039 (click view raw log, then find string `Stable unexpected results`, from there results start):
A lot of passes (do note that servo hit flakyl CRASH `webgpu:shader,execution,expression,*` due to https://github.com/servo/servo/issues/31397) so I will only extract bad unexpected results:
```
TIMEOUT /_webgpu/webgpu/cts.https.html?q=webgpu:shader,execution,expression,call,builtin,textureSample:sampled_2d_coords:*
[2024-09-01T20:57:19Z ERROR wgpu_core::device::global] Device::create_shader_module error:
           Shader validation error:
              ┌─ :34:23
              │
         34 │           data[ndx] = textureLoad(tex, global_invocation_id.xy, mipLevel);
              │                                ^^^^^^^^^^^ naga::Expression [22]
       
 
    TIMEOUT [expected PASS] subtest: :format="bc1-rgba-unorm";sample_points="texel-centre"     
    NOTRUN [expected PASS] subtest: :format="bc1-rgba-unorm";sample_points="spiral"
    ...
TIMEOUT /_webgpu/webgpu/cts.https.html?q=webgpu:shader,execution,expression,call,builtin,textureSample:sampled_3d_coords:*
[2024-09-01T20:44:34Z ERROR wgpu_core::device::global] Device::create_shader_module error:
           Shader validation error:
              ┌─ :35:11
              │
         35 │           textureLoad(
              │           ^^^^^^^^^^^ naga::Expression [23]
    PASS [expected FAIL] subtest: :format="r8unorm";viewDimension="3d";sample_points="texel-centre"
    ...
    TIMEOUT [expected FAIL] subtest: :format="bc1-rgba-unorm";viewDimension="cube";sample_points="cube-edges" 
    NOTRUN [expected FAIL] subtest: :format="bc1-rgba-unorm";viewDimension="cube";sample_points="texel-centre"
    ...

**Comment by sagudev** [comment]
First CTS failures is from https://github.com/gpuweb/cts/blob/9b30f7a02d172ce36d138c02614d0a5a1edbfa72/src/webgpu/shader/execution/expression/call/builtin/texture_utils.ts#L3318, because servo does not support GPUExternalTexture, second is just timeout because tests would need to be spillted more (or servo should be faster).

TL;DR all ok

**Comment by ErichDonGubler** [comment]
Planning on reviewing this, and I heard @teoxoy also wanted to take a look. I'll consider this a learning experience for myself, since I'm still getting familiar with Naga, while @teoxoy would be the authoritative "LGTM". 🙂

**Comment by sagudev** [comment]
> I'll consider this a learning experience for myself, since I'm still getting familiar with Naga, while @teoxoy would be the authoritative "LGTM". 🙂

I am relatively new to naga too (my first commit to naga is 34bb9e4ceb45a5b1cfc5df6aa2b2e201cc55372c, which landed 5 days ago).

File: naga/tests/out/wgsl/access.wgsl [hunk 1]
```diff
@@ -121,11 +121,15 @@ fn assign_through_ptr_fn(p: ptr<function, u32>) {
...
-    (*foo_2) = array<vec4<f32>, 2>(vec4(1f), vec4(2f));
+fn assign_array_through_ptr_fn(foo_3: ptr<function, array<vec4<f32>, 2>>) {
+    (*foo_3) = array<vec4<f32>, 2>(vec4(1f), vec4(2f));
     return;
 }
 
+fn array_by_value(a_1: array<i32, 5>, i: i32) -> i32 {
+    return a_1[i];
+}
+
```
ErichDonGubler, line 132: question: @gfx-rs/naga, should this be showing up in final WGSL output at all if it's not used? It doesn't seem to be consumed at all in the rest of the shader.

ErichDonGubler (approved):
LGTM; let's see what @teoxoy has to say now. 😀

naga/tests/out/wgsl/access.wgsl
ErichDonGubler: Answer: Ah, this is here because it's explicitly defined in the `in` shader, and AFAIK we don't have compaction enabled for these snapshots. Nevermind!

jimblandy:
I think the sentence "Fixes #4337 for arrays" in the description of #6188 is confusing GitHub.

jimblandy:
Note that this PR must be squashed when merged.

naga/src/back/spv/block.rs
```diff
@@ -346,14 +346,32 @@ impl<'w> BlockContext<'w> {
...
+                            base_id,
+                            &self.fun_info[base].ty,
+                            index_id,
+                            ty_element,
+                            block,
+                        )?;
+                        self.function.internal_variables.push(variable);
+                        id
+                    }
+                    // TODO: support `crate::TypeInner::Matrix`
```
jimblandy: Please don't add TODO comments. Issues are easier to track, prioritize, and assign. Adding a comment that mentions an issue number is fine, if you like. (I know there are lots of other extant TODOs; the confusion is our fault.)

I'll fix this one myself.

naga/src/back/spv/writer.rs
```diff
@@ -135,6 +138,62 @@ impl Writer {
...
+        block: &mut Block,
+    ) -> Result<(Word, LocalVariable), Error> {
+        let container_type_id = self.get_expression_type_id(container_resolution);
+        let pointer_type_id = self.id_gen.next();
+        Instruction::type_pointer(
+            pointer_type_id,
+            spirv::StorageClass::Function,
+            container_type_id,
+        )
+        .to_words(&mut self.logical_layout.declarations);
```
jimblandy: It turns out that, if the thing being indexed is a `TypeInner::Array`, that will never be a `TypeResolution::Local`, there will always be a `Handle`. So if we change this to expect a `Handle<Type>` instead of `container_resolution`, we can use `get_pointer_id` here as well.

I have a change that does this (and documents `TypeResolution::Local` a bit better) which I'll push to the branch tomorrow.

naga/src/back/spv/writer.rs
```diff
@@ -135,6 +138,62 @@ impl Writer {
...
+            variable.id,
+            &[index_id],
+        ));
+        let id = self.id_gen.next();
+        block.body.push(Instruction::load(
+            result_type_id,
+            id,
+            element_pointer_id,
+            None,
+        ));
```
jimblandy, line 186: I think doing things this way means that, if we have something like:

```
fn (aaa: array<array<array<f32, 3>, 3> 3>, i: i32, j: i32, k: i32) -> f32 {
    return aaa[i][j][k];
}
```

then we'll end up spilling each intermediate array into its own SPIR-V variable, even though we could handle the access normally once we've spilled the outermost thing.

I don't have the wits at the moment to fix this. It's probably fine to address it in a follow-up PR.

jimblandy (commented):
Overall, this looks really reasonable. Not approving yet because:
- I have some more tweaks I want to push
- I haven't actually looked over the test changes yet.
I should be able to finish those tomorrow morning.

naga/src/back/spv/writer.rs
ErichDonGubler: +1 for iterating! 😀

sagudev: Indeed.

sagudev:
What blocking this? Is there something I need to do?

ErichDonGubler:
@sagudev: I don't see anything that's blocking; I see no objection to this PR merging, concrete or abstract, and there's an approving review (from me). We definitely dropped the ball on this one, sorry about that!

@gfx-rs/naga, unless you object in the next 24ish hours, I'm gonna merge this.

jimblandy (approved):
This looks good to me.

jimblandy:
Filed #6358 for the follow-up work.

naga/src/back/spv/writer.rs
jimblandy: Filed #6358 for the follow-up work.

PR #6194, opened 2024-09-02
fix wrong count buffer id

### LINKED ISSUE #6182: multi_draw_indirect_count set incorrect count buffer id.
**Description**
multi_draw_indexed_indirect_count & multi_draw_indirect_count api is regressed since upgrade to v22.0, the root cause is that the count buffer has been set wrong buffer id as code shown in snapshot.

**Repro steps**
Ideally, a runnable example we can check out.

observed behavior: the count buffer has been set to be the same as indirect buffer, which will cause wrong command count in gpu side.
expected behavior: retrieve the count buffer with correct count buffer id.

**Extra materials**
企业微信截图_1724915160935

**Platform**
platform: windows
graphics api: vulkan/dx12
wgpu version: 22.1.0

BlackStone1123:
**Connections**
fixes https://github.com/gfx-rs/wgpu/issues/6182

**Description**

**Testing**

**Checklist**

- [ ] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/command/render.rs
+++ b/wgpu-core/src/command/render.rs
@@ -3162,7 +3162,7 @@ impl Global {
             .map_err(|_| RenderPassErrorInner::InvalidBuffer(buffer_id))
             .map_pass_err(scope)?;
         let count_buffer = buffers
-            .get_owned(buffer_id)
+            .get_owned(count_buffer_id)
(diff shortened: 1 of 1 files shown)


teoxoy (approved):

teoxoy:
Thank you for the fix!

PR #6197, opened 2024-09-02
Properly handle the case where Navigator.gpu is undefined and WebGPU is the only compiled backend

### LINKED ISSUE #6196: `Instance::request_adapter` silently fails when browser doesn't support WebGPU and `wgpu` compiled without `webgl` feature.
**Description**

When `wgpu` is compiled with `webgpu` support but *not* with `webgl` support, `Instance::request_adapter` will silently fail in the JavaScript bindings because the `gpu` property of the `Navigator` object is undefined:

image

```
[eframe::web::web_painter_wgpu] eframe-0.28.1/src/web/web_painter_wgpu.rs:87: Creating wgpu painter [eframe_playground-4f32d0779d5ce992.js:355:17](http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992.js)
[eframe::web::web_painter_wgpu] eframe-0.28.1/src/web/web_painter_wgpu.rs:108: Creating wgpu instance with backends Backends(VULKAN | GL | METAL | DX12 | BROWSER_WEBGPU) [eframe_playground-4f32d0779d5ce992.js:355:17](http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992.js)
[eframe::web::web_painter_wgpu] eframe-0.28.1/src/web/web_painter_wgpu.rs:133: Attempting to create WebGPU adapter to check for support. [eframe_playground-4f32d0779d5ce992.js:355:17](http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992.js)
Uncaught TypeError: getObject(...) is undefined
   __wbg_requestAdapter_e6f12701c7a38391 http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992.js:593
    eframe_playground-48dc8705e159ba07.wasm.<wgpu::backend::webgpu::ContextWebGpu as wgpu::context::Context>::instance_request_adapter::hadf591a74423c947 http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:2242345
    eframe_playground-48dc8705e159ba07.wasm.<T as wgpu::context::DynContext>::instance_request_adapter::h1703fd936174fd32 http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:2323304
    h6fe9b80c9ca94af0 http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:2321486
    h647f146a6e64bf74 http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:132062
    h8db8326439fec966 http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:379486
    h502255be13bbe6eb http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:567807
    h96675c010ef6ac33 http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:873248
    hbfe20424c414c393 http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:1840992
    hd9850ef0b1f32ef7 http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:2317335
    eframe_playground-48dc8705e159ba07.wasm.<dyn core::ops::function::FnMut<(A,)>+Output = R as wasm_bindgen::closure::WasmClosure>::describe::invoke::hf44c2d097134f68e http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:2440395
    __wbg_adapter_41 http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992.js:237
    real http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992.js:200
    (Async: VoidFunction)
    __wbg_queueMicrotask_12a30234db4045d3 http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992.js:896
    hade5a56ce01b34cb http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:2107289
    h8109c4df9a7e514c http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:2240165
    h50cd6afe6cb6ebc3 http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:2358847
    h6a88423b0971bda0 http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:1982201
    h6db9c6983f17410f http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:2412272
    h85d20172bbd387bb http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:2412116
    h04ef72bb8494faea http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:2320016
    h74eda17738e2d5ac http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:2187987
    h95f28351807ad7f3 http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:2314583
    main http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:2440064
    wasm http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992_bg.wasm:2445515
    __wbg_finalize_init http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992.js:1633
    __wbg_init http://127.0.0.1:8080/eframe_playground-4f32d0779d5ce992.js:1681
    InterpretGeneratorResume self-hosted:1420
    AsyncFunctionNext self-hosted:807
    (Async: async)
    <anonymous> http://127.0.0.1:8080/:16
```

**Repro steps**

I am brand new to compiling web apps with Rust, so I don't know how to minimize my reproduction. I discovered this bug by trying to use [`eframe_template`](https://github.com/emilk/eframe_template).

If you clone [this branch of my `eframe_playground` repository](https://github.com/BGR360/eframe_playground/tree/wgpu-firefox-bug), you should be able to reproduce the problem on the latest (not-nightly) FireFox.

The repo is built with `Trunk`:

```
$ git clone -b wgpu-firefox-bug https://github.com/BGR360/eframe_playground.git
$ cd eframe_playground
$ trunk serve
<...>
2024-09-02T00:02:33.965564Z  INFO ✅ success
2024-09-02T00:02:33.965609Z  INFO 📡 serving static assets at -> /
2024-09-02T00:02:33.965654Z  INFO 📡 server listening at:
2024-09-02T00:02:33.965656Z  INFO     🏠 http://127.0.0.1:8080/
2024-09-02T00:02:33.965658Z  INFO     🏠 http://[::1]:8080/
```

The behavior I would expect in this case is that `Instance::request_adapter` returns `None`. Instead, the WASM crashes in the JavaScript bindings due to invoking bindings on an `undefined` GPU object.

**Platform**

MacOS Sonoma 14.4.1 (23E224), Apple Silicon
FireFox 129.0.2
Chrome 127.0.6533.89 (Official Build) (arm64)

The latest `eframe` uses `wgpu` version `0.20.1`, but I'm fairly certain the bug still exists on the latest `wgpu`, see below.

**Root cause**

Here's what I believe is happening:

In `Instance::new`, `is_only_available_backend` is true and `support_webgpu` is false (because `gpu.is_undefined()` is true):
https://github.com/gfx-rs/wgpu/blob/26398ea0db98f28c5d129a26f6c5e88442747628/wgpu/src/api/instance.rs#L117-L129

Because `is_only_available_backend` overrides the other check, `ContextWebGpu::init` is called even though `Navigator.gpu` is undefined.

In `ContextWebGpu::init`, it checks whether the `gpu` property can be fetched from the current thread, but it does not check whether `gpu` is undefined:
https://github.com/gfx-rs/wgpu/blob/26398ea0db98f28c5d129a26f6c5e88442747628/wgpu/src/backend/webgpu.rs#L1116-L1123

Finally, we hit this line in `ContextWebGpu::request_adapter` which calls the JS binding with an `undefined` arg0:
https://github.com/gfx-rs/wgpu/blob/26398ea0db98f28c5d129a26f6c5e88442747628/wgpu/src/backend/webgpu.rs#L1192

**Possible Solutions**

The easiest thing to do would be to panic inside `ContextWebGpu::init` if `gpu.is_undefined()`. But this doesn't seem like the right thing to do.

The better thing to do would be to check if `self.0.is_undefined()` before calling out to the JS bindings, and to return a Future that immediately resolves to `None` if it is undefined. We should probably also put a check-for-undefined in `ContextWebGpu::surface_get_capabilities`. In that case I think it's okay to panic.

I can send a PR sketching out a solution.

BGR360:
**Connections**
Fixes #6196.

**Description**
This PR fixes a bug where `wgpu` would "crash" in the browser (JavaScript `TypeError` exception) when compiled without `webgl` support and the browser doesn't support WebGPU.

**Implementation**
This PR changes `ContextWebGPU` to avoid calling `wgpu_sys` with an `undefined` `Gpu` object. In `ContextWebGpu::request_adapter`, this involves immediately returning `None` if `self.gpu` is undefined in the browser.

**Testing**
Please advise. I have no idea how to write an automated regression test for this kind of change. I could not find any `CONTRIBUTING.md` or likewise providing any guidelines.

Additionally, I have not yet manually tested that this works. I will work on that.
EDIT: Okay I applied the patch to `wgpu` 0.20.1 locally and patched `eframe` to use it, and it does successfully get past the problem.

**Checklist**

- [x] Run `cargo fmt`.
  - [ ] **??** Do I need to do this on `--target wasm32-unknown-unknown` as well **??**

### DIFF
--- a/wgpu/src/api/instance.rs
+++ b/wgpu/src/api/instance.rs
@@ -95,7 +95,7 @@ impl Instance {
     ///   [`Backends::BROWSER_WEBGPU`] takes a special role:
     ///   If it is set and WebGPU support is detected, this instance will *only* be able to create
     ///   WebGPU adapters. If you instead want to force use of WebGL, either
-    ///   disable the `webgpu` compile-time feature or do add the [`Backends::BROWSER_WEBGPU`]
+    ///   disable the `webgpu` compile-time feature or don't add the [`Backends::BROWSER_WEBGPU`]
     ///   flag to the the `instance_desc`'s `backends` field.
     ///   If it is set and WebGPU support is *not* detected, the instance will use wgpu-core
     ///   to create adapters. Meaning that if the `webgl` feature is enabled, it is able to create
--- a/wgpu/src/api/instance.rs
+++ b/wgpu/src/api/instance.rs
@@ -118,8 +118,9 @@ impl Instance {
         {
             let is_only_available_backend = !cfg!(wgpu_core);
             let requested_webgpu = _instance_desc.backends.contains(Backends::BROWSER_WEBGPU);
-            let support_webgpu =
-                crate::backend::get_browser_gpu_property().map_or(false, |gpu| !gpu.is_undefined());
+            let support_webgpu = crate::backend::get_browser_gpu_property()
+                .map(|maybe_gpu| maybe_gpu.is_some())
+                .unwrap_or(false);
 
             if is_only_available_backend || (requested_webgpu && support_webgpu) {
                 return Self {
--- a/wgpu/src/backend/webgpu.rs
+++ b/wgpu/src/backend/webgpu.rs
@@ -41,6 +41,7 @@ unsafe impl<T> Send for Sendable<T> {}
 unsafe impl<T> Sync for Sendable<T> {}
 
 pub(crate) struct ContextWebGpu {
+    /// `None` if browser does not advertise support for WebGPU.
     gpu: Option<DefinedNonNullJsValue<webgpu_sys::Gpu>>,
 }
 #[cfg(send_sync)]
--- a/wgpu/src/backend/webgpu.rs
+++ b/wgpu/src/backend/webgpu.rs
@@ -37,7 +37,11 @@ unsafe impl<T> Send for Sendable<T> {}
 #[cfg(send_sync)]
 unsafe impl<T> Sync for Sendable<T> {}
 
-pub(crate) struct ContextWebGpu(webgpu_sys::Gpu);
+pub(crate) struct ContextWebGpu {
+    // Invariant: `Some(gpu)` implies that `gpu`` is not `undefined` (i.e., the
+    // browser advertises support for WebGPU).
+    gpu: Option<webgpu_sys::Gpu>,
+}
 #[cfg(send_sync)]
 unsafe impl Send for ContextWebGpu {}
 #[cfg(send_sync)]
@@ -187,6 +191,36 @@ impl<F, M> MakeSendFuture<F, M> {
 
 #[cfg(send_sync)]
 unsafe impl<F, M> Send for MakeSendFuture<F, M> {}
+
+/// Wraps a future that returns `Option<T>` and adds the ability to immediately
+/// return None.
+pub(crate) struct OptionFuture<F>(Option<F>);
+
+impl<F: Future<Output = Option<T>>, T> Future for OptionFuture<F> {
+    type Output = Option<T>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<Self::Output> {
+        // This is safe because we have no Drop implementation to violate the Pin requirements and
+        // do not provide any means of moving the inner future.
+        unsafe {
+            let this = self.get_unchecked_mut();
+            match &mut this.0 {
+                Some(future) => Pin::new_unchecked(future).poll(cx),
+                None => task::Poll::Ready(None),
+            }
+        }
+    }
+}
+
+impl<F> OptionFuture<F> {
+    fn some(future: F) -> Self {
+        Self(Some(future))
+    }
+
+    fn none() -> Self {
+        Self(None)
+    }
+}
 
 fn map_texture_format(texture_format: wgt::TextureFormat) -> webgpu_sys::GpuTextureFormat {
     use webgpu_sys::GpuTextureFormat as tf;
@@ -1095,9 +1129,11 @@ impl crate::context::Context for ContextWebGpu {
     type SubmissionIndexData = ();
     type PipelineCacheData = ();
 
-    type RequestAdapterFuture = MakeSendFuture<
-        wasm_bindgen_futures::JsFuture,
-        fn(JsFutureResult) -> Option<Self::AdapterData>,
+    type RequestAdapterFuture = OptionFuture<
+        MakeSendFuture<
+            wasm_bindgen_futures::JsFuture,
+            fn(JsFutureResult) -> Option<Self::AdapterData>,
+        >,
     >;
     type RequestDeviceFuture = MakeSendFuture<
         wasm_bindgen_futures::JsFuture,
@@ -1119,7 +1155,9 @@ impl crate::context::Context for ContextWebGpu {
                 "Accessing the GPU is only supported on the main thread or from a dedicated worker"
             );
         };
-        ContextWebGpu(gpu)
+
+        let gpu = if !gpu.is_undefined() { Some(gpu) } else { None };
+        ContextWebGpu { gpu }
     }
 
     unsafe fn instance_create_surface(
@@ -1189,12 +1227,15 @@ impl crate::context::Context for ContextWebGpu {
         if let Some(mapped_pref) = mapped_power_preference {
             mapped_options.power_preference(mapped_pref);
         }
-        let adapter_promise = self.0.request_adapter_with_options(&mapped_options);
-
-        MakeSendFuture::new(
-            wasm_bindgen_futures::JsFuture::from(adapter_promise),
-            future_request_adapter,
-        )
+        if let Some(gpu) = &self.gpu {
+            let adapter_promise = gpu.request_adapter_with_options(&mapped_options);
+            OptionFuture::some(MakeSendFuture::new(
+                wasm_bindgen_futures::JsFuture::from(adapter_promise),
+                future_request_adapter,
+            ))
+        } else {
+            OptionFuture::none()
+        }
     }
 
     fn adapter_request_device(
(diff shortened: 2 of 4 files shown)


wgpu/src/backend/webgpu.rs
```diff
@@ -1315,9 +1356,11 @@ impl crate::context::Context for ContextWebGpu {
         let mut mapped_formats = formats.iter().map(|format| map_texture_format(*format));
         // Preferred canvas format will only be either "rgba8unorm" or "bgra8unorm".
         // https://www.w3.org/TR/webgpu/#dom-gpu-getpreferredcanvasformat
-        let preferred_format = self.0.get_preferred_canvas_format();
-        if let Some(index) = mapped_formats.position(|format| format == preferred_format) {
-            formats.swap(0, index);
+        if let Some(gpu) = &self.gpu {
```
BGR360: I think I could panic here instead. It shouldn't be possible to get here since `request_adapter` would return `None` if `self.gpu == None`. Should I do that instead?

wgpu/src/backend/webgpu.rs
```diff
@@ -37,7 +37,11 @@ unsafe impl<T> Send for Sendable<T> {}
 #[cfg(send_sync)]
 unsafe impl<T> Sync for Sendable<T> {}
 
-pub(crate) struct ContextWebGpu(webgpu_sys::Gpu);
+pub(crate) struct ContextWebGpu {
+    // Invariant: `Some(gpu)` implies that `gpu`` is not `undefined` (i.e., the
```
a1phyr:
```suggestion
    // Invariant: `Some(gpu)` implies that `gpu` is not `undefined` (i.e., the
```

wgpu/src/backend/webgpu.rs
```diff
@@ -37,7 +37,11 @@ unsafe impl<T> Send for Sendable<T> {}
...
+    // browser advertises support for WebGPU).
+    gpu: Option<webgpu_sys::Gpu>,
```
Wumpf: wrapping this in an option looks nice, but doesn't change the fact that the value inside the option may _still_ be undefined. Or in other words, the invariant documented here can't be statically enforced, so might as well just assert or handle the undefined cases individually

wgpu/src/backend/webgpu.rs
Wumpf: an `assert` on the definedness would make sense, that also allows then to document why we think that it can't be undefined at this point

Wumpf (changes requested):
great catch! Not convinced of the `Option` wrapper on gpu, but looking good otherwise.
Unfortunate that this needs the `OptionFuture` wrapper, but don't see a better solution either

wgpu/src/backend/webgpu.rs
BGR360: > can't be statically enforced, so might as well just assert

I disagree. The invariant only needs to be enforced in one place (`init()`), and anybody in the future that writes new code trying to use `self.gpu` will be forced by the compiler to deal with the fact that the thing might be undefined.

I can change if you really would rather me do so, but I do believe that writing it this way makes future bugs harder to write.

Wumpf: I'm not convinced, especially given that the same type (`Option<webgpu_sys::Gpu>`) is already used with different semantics on  `get_browser_gpu_property` (that's worth considering changing, it's imho a bit wonky how the kinds of None/undefined/defined of that method have different meanings :/ [.......]).

To me it really sounds like you want something like `std::ptr::NonNull` - in many ways a `JsValue` is like a pointer. 
So maybe we can compromise by adding a simple `DefinedJsValue<T>` which comes with a `new` method that returns an option of itself (doing the defined check) and a `Deref` impl for convenience.
Should be a trivial type and I'm sure we can justify the added complexity by using it in more places :)
I'm almost surprised wasm_bindgen doesn't come with it already, but maybe I missed it.

BGR360: I think that's a fine compromise. Where do you suggest I put that `DefinedJsValue<T>` type?

Also, should it be `DefinedNonNullJsValue`? Or is null-ness not really so much of a problem?

> I'm almost surprised wasm_bindgen doesn't come with it already, but maybe I missed it.

I can't find anything of the sort.

Wumpf: I think you're right, we want all the way here - `DefinedNonNullJsValue` is what we need!
I'm a big fan of fine grained modules - makes it easy to move somewhere else when needed. I'd put it in backend/webgpu right now since this is the place where all other webgpu related javascript stuff lives.

Thanks for bearing with me on this!

BGR360: Okay, I added `DefinedNonNullJsValue` and used it in `ContextWebGpu`.

I also used it in the signature of `get_browser_gpu_property()` to try to make it less wonky :)

wgpu/src/api/instance.rs
```diff
@@ -95,7 +95,7 @@ impl Instance {
     ///   [`Backends::BROWSER_WEBGPU`] takes a special role:
     ///   If it is set and WebGPU support is detected, this instance will *only* be able to create
     ///   WebGPU adapters. If you instead want to force use of WebGL, either
-    ///   disable the `webgpu` compile-time feature or do add the [`Backends::BROWSER_WEBGPU`]
+    ///   disable the `webgpu` compile-time feature or don't add the [`Backends::BROWSER_WEBGPU`]
```
Wumpf, line 98: oops, great catch

wgpu/src/backend/webgpu.rs
```diff
@@ -1189,12 +1235,16 @@ impl crate::context::Context for ContextWebGpu {
...
-            future_request_adapter,
-        )
+        if let Some(gpu) = &self.gpu {
+            let adapter_promise = gpu.request_adapter_with_options(&mapped_options);
+            OptionFuture::some(MakeSendFuture::new(
+                wasm_bindgen_futures::JsFuture::from(adapter_promise),
+                future_request_adapter,
+            ))
+        } else {
+            // Gpu is undfined; WebGPU is not supported in this browser.
```
Wumpf:
```suggestion
            // Gpu is undefined; WebGPU is not supported in this browser.
```

Wumpf (approved):
nice!

PR #6199, opened 2024-09-02
[naga wgsl-in] allow global vars without explicit type

### LINKED ISSUE #4601: [wgsl-in] Naga doesn't infer types for global vars in the private address space
Naga's WGSL front end currently requires a type on all global `var` declarations. WGSL allows `var<private>` declarations to omit the type, if an initializer is given, in which case the type of the variable is the concretized type of the initializer.

sagudev:
**Connections**
Fixes https://github.com/gfx-rs/wgpu/issues/4601

**Description**
Per spec each declaration only needs explicit type or initializer, so based on code for LocalVar I made type infer also work at module scope (GlobalVar).

**Testing**
Existing tests are updated.

**Checklist**

- [x] Run `cargo fmt`.


teoxoy (approved):
Nice!

PR #6200, opened 2024-09-02
D3D12 cleanup

teoxoy:
**Connections**
Follow-up to https://github.com/gfx-rs/wgpu/pull/5956.

**Description**
Mainly cleans-up usage of the D3D12 API.
Also fixes an issue where only high performance adapters would be enumerated if DXGI 1.6 is available.
I will have another follow-up to this that will go over and handle errors on a case-by-case basis (similar to https://github.com/gfx-rs/wgpu/pull/6119).

**Testing**
Existing tests. I also tested DXC locally with the changes to shader compilation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -149,7 +149,6 @@ gpu-descriptor = "0.3"
 bit-set = "0.8"
 gpu-allocator = { version = "0.27", default-features = false }
 range-alloc = "0.1"
-hassle-rs = "0.11.0"
 windows-core = { version = "0.58", default-features = false }
 
 # Gles dependencies
--- a/wgpu-hal/src/dx12/mod.rs
+++ b/wgpu-hal/src/dx12/mod.rs
@@ -116,7 +116,7 @@ impl D3D12Lib {
             riid: *const windows_core::GUID,
             ppdevice: *mut *mut core::ffi::c_void,
         ) -> windows_core::HRESULT;
-        let func: libloading::Symbol<Fun> = unsafe { self.lib.get(b"D3D12CreateDevice") }?;
+        let func: libloading::Symbol<Fun> = unsafe { self.lib.get(b"D3D12CreateDevice\0") }?;
 
         let mut result__ = None;
 
@@ -148,7 +148,7 @@ impl D3D12Lib {
             pperrorblob: *mut *mut core::ffi::c_void,
         ) -> windows_core::HRESULT;
         let func: libloading::Symbol<Fun> =
-            unsafe { self.lib.get(b"D3D12SerializeRootSignature") }?;
+            unsafe { self.lib.get(b"D3D12SerializeRootSignature\0") }?;
 
         let desc = Direct3D12::D3D12_ROOT_SIGNATURE_DESC {
             NumParameters: parameters.len() as _,
@@ -187,7 +187,7 @@ impl D3D12Lib {
             riid: *const windows_core::GUID,
             ppvdebug: *mut *mut core::ffi::c_void,
         ) -> windows_core::HRESULT;
-        let func: libloading::Symbol<Fun> = unsafe { self.lib.get(b"D3D12GetDebugInterface") }?;
+        let func: libloading::Symbol<Fun> = unsafe { self.lib.get(b"D3D12GetDebugInterface\0") }?;
 
         let mut result__ = None;
 
@@ -217,7 +217,7 @@ impl DxgiLib {
             riid: *const windows_core::GUID,
             pdebug: *mut *mut core::ffi::c_void,
         ) -> windows_core::HRESULT;
-        let func: libloading::Symbol<Fun> = unsafe { self.lib.get(b"DXGIGetDebugInterface1") }?;
+        let func: libloading::Symbol<Fun> = unsafe { self.lib.get(b"DXGIGetDebugInterface1\0") }?;
 
         let mut result__ = None;
 
@@ -239,7 +239,7 @@ impl DxgiLib {
             riid: *const windows_core::GUID,
             ppfactory: *mut *mut core::ffi::c_void,
         ) -> windows_core::HRESULT;
-        let func: libloading::Symbol<Fun> = unsafe { self.lib.get(b"CreateDXGIFactory2") }?;
+        let func: libloading::Symbol<Fun> = unsafe { self.lib.get(b"CreateDXGIFactory2\0") }?;
 
         let mut result__ = None;
 
@@ -261,7 +261,7 @@ impl DxgiLib {
             riid: *const windows_core::GUID,
             ppfactory: *mut *mut core::ffi::c_void,
         ) -> windows_core::HRESULT;
-        let func: libloading::Symbol<Fun> = unsafe { self.lib.get(b"CreateDXGIFactory1") }?;
+        let func: libloading::Symbol<Fun> = unsafe { self.lib.get(b"CreateDXGIFactory1\0") }?;
 
         let mut result__ = None;
 
--- a/wgpu-hal/src/dx12/mod.rs
+++ b/wgpu-hal/src/dx12/mod.rs
@@ -66,20 +66,46 @@ use crate::auxil::{
 };
 
 #[derive(Debug)]
+struct DynLib {
+    inner: libloading::Library,
+}
+
+impl DynLib {
+    unsafe fn new(filename: &'static str) -> Result<Self, libloading::Error> {
+        unsafe { libloading::Library::new(filename) }.map(|inner| Self { inner })
+    }
+
+    unsafe fn get<T>(
+        &self,
+        symbol: &[u8],
+    ) -> Result<libloading::Symbol<'_, T>, crate::DeviceError> {
+        unsafe { self.inner.get(symbol) }.map_err(|e| match e {
+            libloading::Error::GetProcAddress { .. } | libloading::Error::GetProcAddressUnknown => {
+                crate::DeviceError::Unexpected
+            }
+            libloading::Error::IncompatibleSize
+            | libloading::Error::CreateCString { .. }
+            | libloading::Error::CreateCStringWithTrailing { .. } => crate::hal_internal_error(e),
+            _ => crate::DeviceError::Unexpected, // could be unreachable!() but we prefer to be more robust
+        })
+    }
+}
+
+#[derive(Debug)]
 struct D3D12Lib {
-    lib: libloading::Library,
+    lib: DynLib,
 }
 
 impl D3D12Lib {
     fn new() -> Result<Self, libloading::Error> {
-        unsafe { libloading::Library::new("d3d12.dll").map(|lib| D3D12Lib { lib }) }
+        unsafe { DynLib::new("d3d12.dll").map(|lib| Self { lib }) }
     }
 
     fn create_device(
         &self,
         adapter: &DxgiAdapter,
         feature_level: Direct3D::D3D_FEATURE_LEVEL,
-    ) -> Result<windows_core::Result<Direct3D12::ID3D12Device>, libloading::Error> {
+    ) -> Result<Direct3D12::ID3D12Device, crate::DeviceError> {
         // Calls windows::Win32::Graphics::Direct3D12::D3D12CreateDevice on d3d12.dll
         type Fun = extern "system" fn(
             padapter: *mut core::ffi::c_void,
@@ -90,14 +116,18 @@ impl D3D12Lib {
         let func: libloading::Symbol<Fun> = unsafe { self.lib.get(b"D3D12CreateDevice") }?;
 
         let mut result__ = None;
-        Ok((func)(
+
+        (func)(
             unsafe { adapter.param().abi() },
             feature_level,
             // TODO: Generic?
             &Direct3D12::ID3D12Device::IID,
             <*mut _>::cast(&mut result__),
         )
-        .map(|| result__.expect("D3D12CreateDevice succeeded but result is NULL?")))
+        .ok()
+        .into_device_result("Device creation")?;
+
+        result__.ok_or(crate::DeviceError::Unexpected)
     }
 
     fn serialize_root_signature(
@@ -114,11 +144,8 @@ impl D3D12Lib {
             ppblob: *mut *mut core::ffi::c_void,
             pperrorblob: *mut *mut core::ffi::c_void,
         ) -> windows_core::HRESULT;
-        let func: libloading::Symbol<Fun> = unsafe { self.lib.get(b"D3D12SerializeRootSignature") }
-            .map_err(|e| {
-                log::error!("Unable to find serialization function: {:?}", e);
-                crate::DeviceError::Lost
-            })?;
+        let func: libloading::Symbol<Fun> =
+            unsafe { self.lib.get(b"D3D12SerializeRootSignature") }?;
 
         let desc = Direct3D12::D3D12_ROOT_SIGNATURE_DESC {
             NumParameters: parameters.len() as _,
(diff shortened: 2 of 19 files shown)


nical (approved):
Glorious.

wgpu-hal/src/dx12/mod.rs
```diff
@@ -65,39 +65,72 @@ use crate::auxil::{
...
+        (func)(
+            adapter.as_raw(),
             feature_level,
             // TODO: Generic?
             &Direct3D12::ID3D12Device::IID,
             <*mut _>::cast(&mut result__),
         )
-        .map(|| result__.expect("D3D12CreateDevice succeeded but result is NULL?")))
+        .ok()
+        .into_device_result("Device creation")?;
```
teoxoy, line 131: This change is causing CI to panic (due to `device_lost_panic`). It seems that creating a device from the WARP adapter fails with `DXGI_ERROR_DEVICE_RESET` in the `DEVICE_LIFETIME_CHECK` & `MULTIPLE_DEVICES` tests.

teoxoy: These 2 tests are not set up properly actually. They create a new instance, adapter and device from env or with defaults. I suspect that when we see they pass on WARP in CI, they actually use llvmpipe (confirmed in https://github.com/gfx-rs/wgpu/pull/6203).

ErichDonGubler: Does this mean we were never properly using WARP adapters? 👀

teoxoy: I'm not sure, creating the first device from the WARP adapter seems to work but creating a 2nd one seems to fail with `DXGI_ERROR_DEVICE_RESET`.

teoxoy: The only thing I could find online that is related to this is:

> Cheaper adapter enumeration
> For Direct3D 12, it’s no longer possible to backtrack from a device to the [IDXGIAdapter](https://learn.microsoft.com/en-us/windows/win32/api/DXGI/nn-dxgi-idxgiadapter) that was used to create it. It’s also no longer possible to provide D3D_DRIVER_TYPE_WARP into [D3D12CreateDevice](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-d3d12createdevice). To make development easier, you can use [IDXGIFactory4](https://learn.microsoft.com/en-us/windows/win32/api/DXGI1_4/nn-dxgi1_4-idxgifactory4) to deal with both of these. [IDXGIFactory4::EnumAdapterByLuid](https://learn.microsoft.com/en-us/windows/win32/api/DXGI1_4/nf-dxgi1_4-idxgifactory4-enumadapterbyluid) (designed to be paired with [ID3D12Device::GetAdapterLuid](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12device-getadapterluid)) enables an app to retrieve information about the adapter where a Direct3D 12 device was created. [IDXGIFactory4::EnumWarpAdapter](https://learn.microsoft.com/en-us/windows/win32/api/DXGI1_4/nf-dxgi1_4-idxgifactory4-enumwarpadapter) provides an adapter which can be provided to D3D12CreateDevice to use the WARP renderer.

_from https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/dxgi-1-4-improvements#cheaper-adapter-enumeration_

We should run some experiments to see what works and what doesn't. I will open an issue to track this.

For this PR, I modified the tests to reuse the instance/adapter and they are now passing.

Cargo.toml
```diff
@@ -149,7 +149,6 @@ gpu-descriptor = "0.3"
 bit-set = "0.8"
 gpu-allocator = { version = "0.27", default-features = false }
 range-alloc = "0.1"
-hassle-rs = "0.11.0"
```
MarijnS95, line 152: For future reference, the only reason why `hassle-rs` cannot use `windows-rs` as bindings is because the `windows-rs` maintainers refuse to allow basic Linux support. DirectxShaderConverter works on Linux via the same COM bindings otherwise.

(Either way the abstraction provided by `hassle-rs` is far from ideal)

wgpu-hal/src/dx12/mod.rs
```diff
@@ -137,8 +167,6 @@ impl D3D12Lib {
             <*mut _>::cast(&mut error),
         )
         .ok()
-        // TODO: If there's a HRESULT, error may still be non-null and
-        // contain info.
         .into_device_result("Root signature serialization")?;
```
MarijnS95, line 170: The idea was to check the `error` blob before bailing out with `into_device_result()`, missed this during the previous cleanup. The TODO remains valid.

wgpu-hal/src/dx12/mod.rs
```diff
@@ -147,104 +175,102 @@ impl D3D12Lib {
...
+        let mut result__ = None;
+
+        (func)(&Direct3D12::ID3D12Debug::IID, <*mut _>::cast(&mut result__))
+            .ok()
+            .into_device_result("GetDebugInterface")?;
 
-        let mut result__ = core::ptr::null_mut();
-        Ok((func)(&Direct3D12::ID3D12Debug::IID, &mut result__)
-            .and_then(|| unsafe { windows_core::Type::from_abi(result__) }))
+        result__.ok_or(crate::DeviceError::Unexpected)
```
MarijnS95, line 198: Note that a lot of this was intentionally verbatim-copied from `windows-rs` to allow easier updating.

wgpu-hal/src/dx12/mod.rs
```diff
@@ -147,104 +175,102 @@ impl D3D12Lib {
...
-        let mut result__ = core::ptr::null_mut();
-        Ok(
-            (func)(factory_flags, &Dxgi::IDXGIFactory4::IID, &mut result__)
-                .and_then(|| unsafe { windows_core::Type::from_abi(result__) }),
+        (func)(
+            factory_flags,
+            &Dxgi::IDXGIFactory4::IID,
+            <*mut _>::cast(&mut result__),
         )
+        .ok()
```
MarijnS95, line 251: I should report over at `windows-rs` that the naming choice of this function is less than ideal. I keep mistaking it for https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.ok which _discards_ `Error` information by turning `Result` into `Option` which instantly triggers during review.

wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -1,11 +1,11 @@
...
 use std::ffi::CStr;
-use std::ptr;
+use std::path::PathBuf;
+use windows::{
+    core::{Interface, PCSTR, PCWSTR},
+    Win32::Graphics::Direct3D::{Dxc, Fxc},
+};
 
-pub(super) use dxc::{compile_dxc, get_dxc_container, DxcContainer};
-use windows::Win32::Graphics::Direct3D;
```
MarijnS95, line 5: Doesn't `wgpu` use `StdExternalCrate` formatting?

Cargo.toml
teoxoy: I found it more straightforward to just use the bindings directly since we have to deal with them for D3D12 anyway.

wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -75,234 +73,237 @@ pub(super) fn compile_fxc(
...
+    fn new(lib_path: Option<PathBuf>, lib_name: &'static str) -> Result<Self, libloading::Error> {
+        let lib_path = if let Some(lib_path) = lib_path {
+            if lib_path.is_file() {
+                lib_path
+            } else {
+                lib_path.join(lib_name)
             }
+        } else {
+            PathBuf::from(lib_name)
         };
```
MarijnS95, line 109: I wouldn't have copied this weird bit from `hassle-rs` though, it has caused us quite a few headaches over the years.

wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -75,234 +73,237 @@ pub(super) fn compile_fxc(
...
-                hassle_rs::HassleError::ValidationError(_e) => unimplemented!(),
-                hassle_rs::HassleError::CompileError(_e) => unimplemented!(),
-            }
-        }
+impl OPCWSTR {
+    fn new(s: &str) -> Self {
+        let mut inner: Vec<_> = s.encode_utf16().collect();
+        inner.push(0);
+        Self { inner }
     }
```
MarijnS95, line 187: Didn't `HSTRING` work for you?

wgpu-hal/src/dx12/mod.rs
teoxoy: I see, but I think it's more valuable to have those be consistent. I'm also not a fan of using the `from_abi` function which maps null to `Error::empty()`.

wgpu-hal/src/dx12/mod.rs
teoxoy: Hmm, is that something we should handle? What would be an example?

The only scenario I can think of is: the blob getting partially filled with errors after which we get an OOM but it's probably ok to error with an OOM instead of reporting the partial errors.

I wish the exact error codes were documented (like in Vulkan), now we have to guess which variants we might run into...

Cargo.toml
MarijnS95: Yeah the abstraction by `hassle-rs` isn't that useful. It's mostly creating the bindings but we have those now (except not for Linux 😭) in `windows-rs`.

wgpu-hal/src/dx12/mod.rs
MarijnS95: Right yes, that is unfortunate.

Perhaps this also has to do with SAL annotations, if I remember correctly there are different annotations that allow return values to be `NULL`, versus them only being `NULL` when a non-`SUCCESS` `RESULT` is returned.

The original code now no longer relies on deref semantics, unfortunately.

wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -75,234 +73,237 @@ pub(super) fn compile_fxc(
...
-    ) -> Result<Option<DxcContainer>, crate::DeviceError> {
-        // Falls back to Fxc and logs an error.
-        log::error!("DXC shader compiler was requested on Instance creation, but the DXC feature is disabled. Enable the `dxc_shader_compiler` feature on wgpu_hal to use DXC.");
-        Ok(None)
+fn get_output<T: Interface>(
+    res: &Dxc::IDxcResult,
+    kind: Dxc::DXC_OUT_KIND,
+) -> Result<T, crate::DeviceError> {
+    let mut result__: Option<T> = None;
+    unsafe { res.GetOutput::<T>(kind, &mut None, <*mut _>::cast(&mut result__)) }
```
MarijnS95, line 199: This could coerce because it takes a `*mut Option<T>`

```suggestion
    unsafe { res.GetOutput::<T>(kind, &mut None, &mut result__) }
```

wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -75,234 +73,237 @@ pub(super) fn compile_fxc(
...
-        _dxil_path: Option<PathBuf>,
-    ) -> Result<Option<DxcContainer>, crate::DeviceError> {
-        // Falls back to Fxc and logs an error.
-        log::error!("DXC shader compiler was requested on Instance creation, but the DXC feature is disabled. Enable the `dxc_shader_compiler` feature on wgpu_hal to use DXC.");
-        Ok(None)
+fn get_output<T: Interface>(
+    res: &Dxc::IDxcResult,
+    kind: Dxc::DXC_OUT_KIND,
+) -> Result<T, crate::DeviceError> {
+    let mut result__: Option<T> = None;
```
MarijnS95, line 198: May just be me who prefers this short-hand:

```suggestion
    let mut result__ = None::<T>;
```

Although the type can be inferred:

```suggestion
    let mut result__ = None;
```

Not that it matters too much, this signature will change when upstream metadata fixes land.

wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -75,234 +73,237 @@ pub(super) fn compile_fxc(
...
-        }))
+        unsafe { crate::dx12::DynLib::new(lib_path).map(|lib| Self { lib }) }
+    }
+
+    pub fn create_instance<T: DxcObj>(&self) -> Result<T, crate::DeviceError> {
+        type Fun = extern "system" fn(
+            rclsid: *const windows_core::GUID,
+            riid: *const windows_core::GUID,
+            ppv: *mut *mut core::ffi::c_void,
+        ) -> windows_core::HRESULT;
```
MarijnS95, line 118: Any reason to not use the existing `DxcCreateInstanceProc` typedef? Because it's wrapped in `Option<>`?

wgpu-hal/src/dx12/shader_compilation.rs
```diff
@@ -75,234 +73,237 @@ pub(super) fn compile_fxc(
...
+        Size: source.len(),
+        Encoding: Dxc::DXC_CP_UTF8.0,
+    };
+
+    let compile_res: Dxc::IDxcResult = unsafe {
+        dxc_container
+            .compiler
+            .Compile(&buffer, Some(&compile_args), None)
+    }
+    .into_device_result("Compile")?;
```
MarijnS95, line 266: Like the old API, I wonder if `Compile` returns an error code while `IDxcResult` is _also_ returned with valid errors (the root cause) in `DXC_OUT_ERRORS`.

Or if that's still communicated by `IDxcOperationResult::GetResult()` which you're not checking here?

wgpu-hal/src/dx12/mod.rs
MarijnS95: I was expecting a similar setup like `Compile()` the error doesn't need to be OOM.

For `Compile()`, I'm not sure which `HRESULT`s to check either. There is:

- The result from `Compile()`;
- Non-null `error` blob;
- Null `blob`;
- Result from `IDxcOperationResult::GetStatus()`.

(and any getters on `IDxcOperationResult`/`IDxcResult` can fail as well)

wgpu-hal/src/dx12/shader_compilation.rs
teoxoy: I can imagine, but I didn't want to change the behavior (yet) as it is documented in our docs.

wgpu-hal/src/dx12/shader_compilation.rs
teoxoy: It doesn't seem that we do.

wgpu-hal/src/dx12/shader_compilation.rs
teoxoy: I saw that it exists but wasn't 100% sure how that would behave with libloading. It's not something I've seen in their docs/examples.

wgpu-hal/src/dx12/shader_compilation.rs
teoxoy: It seemed to be even more complex for this purpose (ref-counted, no easy way to create it from a `&str`, no easy way to get a `PCWSTR` from it).

wgpu-hal/src/dx12/shader_compilation.rs
teoxoy: From my testing, `IDxcCompiler3::Compile` seems to succeed even if there was a compilation error.

Getting the blob via `IDxcResult::GetOutput(DXC_OUT_ERRORS)` always succeeds. If there is a compilation error its size/length will be non 0 and will contain the error message.

The wiki seems to say that `Compile` will only fail if the compiler itself ran into an internal error.
https://github.com/microsoft/DirectXShaderCompiler/wiki/Advanced-Error-Handling#internal-compiler-errors

wgpu-hal/src/dx12/mod.rs
teoxoy: Left a reply in https://github.com/gfx-rs/wgpu/pull/6200#discussion_r1744221644.

PR #6201, opened 2024-09-02
[core, hal, types] Clarify `wgpu_hal`'s bounds check promises.

### LINKED ISSUE #1813: Robust Buffer Access conflicts with lazy initialization
WebGPU allows for RBA behavior on out-of-bound accesses into buffers. This behavior says that 0, 1, or any value from the underlying buffer (not binding!) can be returned or written.
However, if the buffer ranges are lazily initialized to zero, we can end up in a situation where shader reads values outside of the initialized range. This is not acceptable.

A quick solution would be to consider any storage buffer bindings to cover the whole buffer for the purpose of lazy initialization. cc @Wumpf 

Note also that the extent of the problem depends on the bounds checking policy: https://github.com/gfx-rs/wgpu/blob/3afab9c5aaba0f87be539fc656c3aabf245d2503/wgpu-hal/src/vulkan/adapter.rs#L843 . If it's `ReadZeroSkipWrite` then we can keep doing what we do today. cc @jimblandy

jimblandy:
- [core, hal, types] Clarify `wgpu_hal`'s bounds check promises.

    In `wgpu_hal`:
    
    - Document that `wgpu_hal` guarantees that shaders will not access buffer
      contents beyond the bindgroups' bound regions, rounded up to some
      adapter-specific alignment. Introduce the term "accessible region" for
      the portion of the buffer that shaders can actually get at.
    
    - Document that all bets are off if you disable bounds checks with
      `ShaderModuleDescriptor::runtime_checks`.
    
    - Provide this alignment in `wgpu_hal::Alignments`. Update all backends
      appropriately.
    
    - In the Vulkan backend, use Naga to inject bounds checks on buffer accesses
      unless `robustBufferAccess2` is available; `robustBufferAccess` is not
      sufficient. Retrieve `VK_EXT_robustness2`'s properties, as needed to discover
      the alignment above.
    
    In `wgpu_core`:
    
    - Use buffer bindings' accessible regions to determine which parts of the buffer
      need to be initialized.
    
    In `wgpu_types`:
    
    - Document some of the possible effects of using
      `ShaderBoundsChecks::unchecked`.

- [core]: Let `Device::create_buffer_binding` get `limits` from `self`.

  Rather than passing `self.limits` to `Device::create_buffer_binding`
  as an argument, let it simply refer to `self.limits` itself.

- [core] Remove unnecessary `pub(crate)` from `Device` methods.
  
  Remove the `pub(crate)` visibility marking from various associated
  functions of `Device` that are defined in, and not used outside of,
  the `wgpu_core::device::resource` module.

- [core] Simplify `self` types in `device::resource`.
  
  Change various functions that have no need to create an owning
  reference to the `Device` to accept `&self` instead of `&Arc<Self>`.
  
  Change `ParentDevice::same_device` to accept `&Device` as the point of
  comparison, not `&Arc<Device>`. Call sites will use Deref conversion,
  so no callers need to be changed.

- [core] Rename `map_buffer` to `resolve_buffer`, etc.
  
  In `Device::create_bind_group`, name the functions that convert
  resource ids to Arcs `resolve_foo`, not `map_foo`:
  
  - The types that hold Arcs are usually called `ResolvedBlah`.
  
  - The name `map_buffer` is misleading.

Fixes #1813.

**Checklist**

- [X] Run `cargo fmt`.
- [X] Run `cargo clippy`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,7 @@ By @teoxoy [#6134](https://github.com/gfx-rs/wgpu/pull/6134).
 - Fix crash when dropping the surface after the device. By @wumpf in [#6052](https://github.com/gfx-rs/wgpu/pull/6052)
 - Fix error message that is thrown in create_render_pass to no longer say `compute_pass`. By @matthew-wong1 [#6041](https://github.com/gfx-rs/wgpu/pull/6041)
 - Add `VideoFrame` to `ExternalImageSource` enum. By @jprochazk in [#6170](https://github.com/gfx-rs/wgpu/pull/6170)
+- Document `wgpu_hal` bounds-checking promises, and adapt `wgpu_core`'s lazy initialization logic to the slightly weaker-than-expected guarantees. By @jimblandy in [#6201](https://github.com/gfx-rs/wgpu/pull/6201)
 
 #### GLES / OpenGL
 
--- a/wgpu-core/src/binding_model.rs
+++ b/wgpu-core/src/binding_model.rs
@@ -884,6 +884,16 @@ pub(crate) fn buffer_binding_type_alignment(
     }
 }
 
+pub(crate) fn buffer_binding_type_bounds_check_alignment(
+    alignments: &hal::Alignments,
+    binding_type: wgt::BufferBindingType,
+) -> wgt::BufferAddress {
+    match binding_type {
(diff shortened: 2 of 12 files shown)


nical (approved):
Beautiful.

teoxoy (approved):
Looks great!

PR #6203, opened 2024-09-03
test

teoxoy:
Testing https://github.com/gfx-rs/wgpu/pull/6200#discussion_r1742167268.


teoxoy:
WARP fails with "Panic: failed to create adapter", so it seems that we were running the 2 tests with llvmpipe...

PR #6209, opened 2024-09-04
[hal] Document some `CommandEncoder` trait methods.

jimblandy:
Add documentation for the following `CommandEncoder` trait methods:
- `set_bind_group`
- `begin_render_pass`
- `begin_compute_pass`
- `end_render_pass`
- `end_compute_pass`

### DIFF
--- a/wgpu-hal/src/lib.rs
+++ b/wgpu-hal/src/lib.rs
@@ -51,8 +51,8 @@
 //!   must use [`CommandEncoder::transition_buffers`] between those two
 //!   operations.
 //!
-//! - Pipeline layouts are *explicitly specified* when setting bind
-//!   group. Incompatible layouts disturb groups bound at higher indices.
+//! - Pipeline layouts are *explicitly specified* when setting bind groups.
+//!   Incompatible layouts disturb groups bound at higher indices.
 //!
 //! - The API *accepts collections as iterators*, to avoid forcing the user to
 //!   store data in particular containers. The implementation doesn't guarantee
@@ -1242,8 +1242,40 @@ pub trait CommandEncoder: WasmNotSendSync + fmt::Debug {
 
     // pass common
 
-    /// Sets the bind group at `index` to `group`, assuming the layout
-    /// of all the preceding groups to be taken from `layout`.
+    /// Sets the bind group at `index` to `group`.
+    ///
+    /// If this is not the first call to `set_bind_group` within the current
+    /// render or compute pass:
+    ///
+    /// - If `layout` contains `n` bind group layouts, then any previously set
+    ///   bind groups at indices `n` or higher are cleared.
+    ///
+    /// - If the first `m` bind group layouts of `layout` are equal to those of
+    ///   the previously passed layout, but no more, then any previously set
+    ///   bind groups at indices `m` or higher are cleared.
+    ///
+    /// It follows from the above that passing the same layout as before doesn't
+    /// clear any bind groups.
+    ///
+    /// # Safety
+    ///
+    /// - This [`CommandEncoder`] must be within a render or compute pass.
+    ///
+    /// - `index` must be the valid index of some bind group layout in `layout`.
+    ///   Call this the "relevant bind group layout".
+    ///
+    /// - The layout of `group` must be equal to the relevant bind group layout.
+    ///
+    /// - The length of `dynamic_offsets` must match the number of buffer
+    ///   bindings [with dynamic offsets][hdo] in the relevant bind group
+    ///   layout.
+    ///
+    /// - If those buffer bindings are ordered by increasing [`binding` number]
+    ///   and paired with elements from `dynamic_offsets`, then each offset must
+    ///   be a valid offset for the binding's corresponding buffer in `group`.
+    ///
+    /// [hdo]: wgt::BindingType::Buffer::has_dynamic_offset
+    /// [`binding` number]: wgt::BindGroupLayoutEntry::binding
     unsafe fn set_bind_group(
         &mut self,
         layout: &<Self::A as Api>::PipelineLayout,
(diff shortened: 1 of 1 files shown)


jimblandy:
@cwfitzgerald Could you check my description of `set_bind_group`? I think I got the rules right, but I'm not sure.

wgpu-hal/src/lib.rs
```diff
@@ -1295,11 +1325,50 @@ pub trait CommandEncoder: WasmNotSendSync + fmt::Debug {
...
+    /// - [`set_render_pipeline`]
+    /// - [`set_index_buffer`]
+    /// - [`set_vertex_buffer`]
+    ///
+    /// [`set_bind_group`]: CommandEncoder::set_bind_group
+    /// [`set_push_constants`]: CommandEncoder::set_push_constants
+    /// [`begin_query`]: CommandEncoder::begin_query
+    /// [`set_render_pipeline`]: CommandEncoder::set_render_pipeline
+    /// [`set_index_buffer`]: CommandEncoder::set_index_buffer
+    /// [`set_vertex_buffer`]: CommandEncoder::set_vertex_buffer
```
ErichDonGubler: style: This form seems unnecessarily duplicative when compared to providing the link target on the same line (i.e., ``[`ident`](link)``). The current style is consistent with other linking forms, but I think that the consistency would only be valuable if the links were kept in a single region (i.e., for each of updating symbols).

suggestion:

```suggestion
    /// - [`set_bind_group`](CommandEncoder::set_bind_group)
    /// - [`set_push_constants`](CommandEncoder::set_push_constants)
    /// - [`begin_query`](CommandEncoder::begin_query)
    /// - [`set_render_pipeline`](CommandEncoder::set_render_pipeline)
    /// - [`set_index_buffer`](CommandEncoder::set_index_buffer)
    /// - [`set_vertex_buffer`](CommandEncoder::set_vertex_buffer)
```

wgpu-hal/src/lib.rs
```diff
@@ -1365,11 +1434,46 @@ pub trait CommandEncoder: WasmNotSendSync + fmt::Debug {
...
+    ///
+    /// - [`set_bind_group`]
+    /// - [`set_push_constants`]
+    /// - [`begin_query`]
+    /// - [`set_compute_pipeline`]
+    ///
+    /// [`set_bind_group`]: CommandEncoder::set_bind_group
+    /// [`set_push_constants`]: CommandEncoder::set_push_constants
+    /// [`begin_query`]: CommandEncoder::begin_query
+    /// [`set_compute_pipeline`]: CommandEncoder::set_compute_pipeline
```
ErichDonGubler: style: Ditto the above feedback for ~~duplicity~~ duplication with `begin_render_pass`.

suggestion:

```suggestion
    /// - [`set_bind_group`](CommandEncoder::set_bind_group)
    /// - [`set_push_constants`](CommandEncoder::set_push_constants)
    /// - [`begin_query`](CommandEncoder::begin_query)
    /// - [`set_compute_pipeline`](CommandEncoder::set_compute_pipeline)
```

wgpu-hal/src/lib.rs
```diff
@@ -1242,8 +1242,38 @@ pub trait CommandEncoder: WasmNotSendSync + fmt::Debug {
...
+    ///
+    /// If this is not the first call to `set_bind_group` within the current
+    /// render or compute pass:
+    ///
+    /// - If `layout` contains `n` bind group layouts, then any previously set
+    ///   bind groups at indices `n` or higher are cleared.
+    ///
+    /// - If the first `m` bind group layouts of `layout` are equal to those of
+    ///   the previously passed layout, but no more, then any previously set
+    ///   bind groups at indices `m` or higher are cleared.
```
ErichDonGubler, line 1255: ~~question: Shouldn't this be `any previously set bind groups at indices higher than `m` are cleared`, since the condition implies that index `m` is also still equal?~~

Answer from reading more thoroughly: We're using 0-based indices, so no, `m` is the index above the count.

wgpu-hal/src/lib.rs
```diff
@@ -1242,8 +1242,38 @@ pub trait CommandEncoder: WasmNotSendSync + fmt::Debug {
...
+    /// - The layout of `group` must be equal to the relevant bind group layout.
+    ///
+    /// - The length of `dynamic_offsets` must match the number of buffer
+    ///   bindings [with dynamic offsets][hdo] in the relevant bind group
+    ///   layout.
+    ///
+    /// - If those buffer bindings are ordered by increasing [`binding` number]
+    ///   and paired with elements from `dynamic_offsets`, then each offset must
+    ///   be a valid offset for the binding's corresponding buffer in `group`.
+    ///
```
ErichDonGubler, line 1276: suggestion: Because we're talking about using this in the context of a pass (`within the current render or compute pass`), it seems like we might be missing an explicit safety precondition here for "needs to have begun (but not ended) a render or compute pass". Thoughts?

ErichDonGubler: suggestion: If you agree with ☝🏻 the above suggestion, we might amend the summary line of the method to be:

```diff
-    /// Sets the bind group at `index` to `group`.
+    /// Sets the bind group of the current render or compute pass at `index` to `group`.
```

ErichDonGubler (approved):
praise: This was good for fleshing out pass begin/end expectations I'd sorta had in my head myself, but wasn't confident enough to document. Yay validation! What you wrote in `set_bind_group` seems consistent with what you were telling the Firefox WebGPU team verbally yesterday, so 👍🏻 I think I'm understanding the conditions you're setting forth there.

LGTM on my end, tho I understand if you want to wait for @cwfitzgerald's validation for `set_bind_group` specifically.

wgpu-hal/src/lib.rs
teoxoy, line 1255: Do all native APIs we target clear all `n`/`m` or higher bind groups?

I thought only Vulkan required users to rebind the `n`/`m` or higher bind groups.

jimblandy: `wgpu_hal` is supposed to be a portability layer, so can only promise what the weakest implementation can deliver. We could place the burden on Vulkan to track more state, but this seems like the kind of work that is best addressed in `wgpu_core`.

But the punch line is, unless I'm misunderstanding things, Metal actually needs this requirement too. Each MSL entry point refers to all its resources via a flat numbering scheme (`[[texture(0)]]`, `[[texture(1)]]`, `[[texture(2)]]`), with no analogue of WGSL's `@group` / `@binding` two-level scheme. At the Metal API level, the `set_bind_group` implementation just makes a bunch of calls like `[encoder setTexture:t atIndex:2]`. We assign MSL indices consecutively, so the indices of resources supplied by higher-numbered bind groups are influenced by the number of those resources present in lower-numbered bind groups. This means that, after we switch pipelines, the MSL numbering of resources supplied by higher-numbered bind groups could be off.

So MSL effectively has the same requirement as Vulkan, arising from our desire to densely number resources.

(I suspect that the Vulkan requirement exists to allow implementations to "stack up" resource pointers or descriptors in a contiguous block of memory, so the reasons behind the restrictions in both cases are similar.)

jimblandy: Note that if we were to switch `wgpu_hal::metal` to use argument buffers, Metal would no longer have this restriction, because this would actually introduce this two-level numbering: the argument buffers themselves would have `[[buffer(g)]]` indices corresponding to `@group(g)` indices, and the `[[id(b)]]` attributes within the argument buffer would correspond to `@binding(b)` indices. Then higher-level bind groups could be stable across pipeline layout changes.

wgpu-hal/src/lib.rs
jimblandy: > we might be missing an explicit safety precondition

Oh, we definitely are. I have some other commits in progress that will extend the "states" described in the `CommandEncoder` documentation to include "recording render pass" and "recording compute pass". But I'd like to leave that for a separate PR, because it touches pretty much all the `CommandEncoder` methods if done right.

For now, I'll add a safety condition.

wgpu-hal/src/lib.rs
jimblandy: However, `wgpu_hal` still probably shouldn't promise more bind group stability, since there isn't a natural way to implement it on Vulkan. So the above doesn't serve as a reason to prioritize using argument buffers on Metal.

teoxoy: > `wgpu_hal` is supposed to be a portability layer, so can only promise what the weakest implementation can deliver. We could place the burden on Vulkan to track more state, but this seems like the kind of work that is best addressed in `wgpu_core`.

I agree; it wasn't my intention to put the burden on Vulkan to track more state.

My comment was targeted at the word "cleared". If not all backends "clear", then `wgpu-hal` shouldn't promise that.

I think we should instead require users of the API to rebind the bind groups but not promise that they will be cleared.

jimblandy: I see what you're saying. I think I agree with you about what `wgpu_hal` should promise - but I still think there's a piece we're missing. However, I want a better place for the discussion than this issue; I'll bring it up in chat.

PR #6210, opened 2024-09-04
[metal] Use `raw-window-metal` to do layer creation

madsmtm:
**EDIT: `raw-window-handle`'s implementation, as described below, was used in https://github.com/gfx-rs/wgpu/pull/7026. This PR now only updates to use the actual `raw-window-handle` crate.**

You know, all that code and documentation I just wrote in https://github.com/gfx-rs/wgpu/pull/6107? Let's throw it ~away~ into a library, so that `wgpu`, `ash`, `vulkano`, `i-slint-renderer-skia` and so on can all benefit from it.

[`raw-window-metal`](https://docs.rs/raw-window-metal) is that library (existed previously, I've updated it in https://github.com/rust-windowing/raw-window-metal/pull/19).

This makes Wgpu depend transitively on `objc2`, so the concerns about doing that apply here as well, see also https://github.com/gfx-rs/wgpu/pull/5641.

**Description**
Use the `raw-window-metal` crate to do layer creation. This uses much the same approach as we do currently, except that it also uses observers internally, which fixes resizing on high DPI screens when using Wgpu together with an `NSView` using auto-layout.

Though it probably won't matter much, it's also more reliable to update the scale factor this way, rather than haphazardly in `configure`.

The logic in `raw-window-metal` _could_ be re-implemented in Wgpu to avoid the dependency, however I deemed it complex enough that I thought it wiser to centralize the implementation in one place.

**Testing**
See https://github.com/gfx-rs/wgpu/pull/6107#issuecomment-2319599762.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`.
  - I did not deem this to be necessary, since the bulk of this has already landed in https://github.com/gfx-rs/wgpu/pull/7026.


Wumpf (commented):
very appreciated, having that code as part of `raw-window-metal` sounds great to me :)

As pointed out this will have to wait for the objc2 pr to be approved. cc: @jimblandy 
(therefore not putting ✔️ just yet so this doesn't get merged ahead of time)

polina4096:
I've tested this and it still adds some noticeable (especially in comparison) with the way it was before previous PR (#6107) had been merged. I've gone into a bit further details here https://github.com/gfx-rs/wgpu/pull/6107#issuecomment-2499212851.

jimblandy:
This depends on #5641. In our meeting today, the wgpu maintainers decided to mark #5641 as draft, since we'll be getting back to that in 2025. So, marking this as draft as well.

andyleiserson (approved):

PR #6214, opened 2024-09-04
chore: satisfy `clippy::ref_as_ptr`

ErichDonGubler:
**Connections**

N/A

**Description**

Rust 1.76 added the `std::ptr::{from_ref,from_mut}` APIs, which allow several `as` casts to be eliminated as potential footguns. Because our MSRV is Rust 1.76, we can (and, IMO, should) migrate to these APIs. This PR just takes on the `from_ref` case; I haven't tried `from_mut` yet.

Rust 1.78's Clippy adds (allow-by-default) lints to check for cases where these APIs can be used instead (i.e., [`clippy::ref_as_ptr`](https://rust-lang.github.io/rust-clippy/stable/index.html#/ref_as_ptr)), but as our MSRV does not use Rust 1.78 yet, this PR does not set these lints to warn.

**Testing**

Things still compile, tests still run!

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/lock/observing.rs
+++ b/wgpu-core/src/lock/observing.rs
@@ -369,16 +369,16 @@ impl ObservationLog {
         self.write_location(new_location);
         self.write_action(&Action::Acquisition {
             older_rank: older_lock.rank.bit.number(),
-            older_location: older_lock.location as *const _ as usize,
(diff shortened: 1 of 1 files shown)


teoxoy (commented):
Maybe an `as_address` helper fn would be more fitting?

teoxoy (approved):

Issue #4738, opened 2023-11-21
Bind group and layout entries should be optional


nical:
Spec:

- https://gpuweb.github.io/gpuweb/#gpubindingcommandsmixin-setbindgroup
- https://www.w3.org/TR/webgpu/#dom-gpupipelinelayoutdescriptor-bindgrouplayouts

Bind group layouts and bind groups should be `null`able, which in Rust parlance means we should take an `Option` of them. This lets app authors “unset” a bind group at an index.

jimblandy:
It's unclear what the underlying platform APIs' expectations are around removing bindgroups, so we're going to need to investigate this a bit before we can tell how much work it will be.

teoxoy:
Relevant spec PR: https://github.com/gpuweb/gpuweb/pull/3938

Wumpf:
this got fixed in
* https://github.com/gfx-rs/wgpu/pull/6216

Wumpf:
ah I spoke too soon, HAL doesn't handle this yet it seems?

ErichDonGubler:
Note that this should include optional [`PipelineLayoutDescriptor::bind_group_layouts`](https://docs.rs/wgpu/27.0.1/wgpu/struct.PipelineLayoutDescriptor.html#structfield.bind_group_layouts).

ETA: Folded into OP.

ErichDonGubler:
Firefox downstream tracking: [bug 1992771](https://bugzilla.mozilla.org/show_bug.cgi?id=1992771)

andyleiserson:
Also discussed in #1489. (I am not going to mark a duplicate right now because there is some discussion in both issues that might be worth consolidating.)

PR #6216, opened 2024-09-04
Make bind group an Option for set_bind_group calls.

bradwerth:
This is just an API change for all the "set_bind_group" calls. Calls that pass a `Some()` argument have unchanged behavior. The `None` cases are left as TODOs.

**Connections**
Issue #4738

**Description**
`set_bind_group` now takes `Option<&BindGroup>`.

**Testing**
Existing tests, player scenes, and examples are all updated to use the new syntax.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,15 @@ traits that have now been implemented for `wgpu` resources.
 
 By @teoxoy [#6134](https://github.com/gfx-rs/wgpu/pull/6134).
 
+#### `set_bind_group` now takes an `Option` for the bind group argument.
+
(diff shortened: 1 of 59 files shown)


nical (approved):

JMS55:
I'd rather use Into<Option<T>> so that the caller does not need to wrap it in Some.

ErichDonGubler:
@JMS55:

> I'd rather use Into<Option> so that the caller does not need to wrap it in Some.

I don't think this feedback should block this PR, but if you feel strongly about it, I'd encourage you to file an issue for further discussion.

PR #6224, opened 2024-09-05
[ci] Re-enable `cargo doc --document-private-items` in CI

jimblandy:
Re-enable CI checks that naga, wgpu-core, and wgpu-hal documentation builds without warnings, even with `--document-private-items`. This was previously disabled due to #4905, which causes long build times, but with the de-generification of `wgpu_core`, that's no longer a problem: a build of all three crates' docs now takes only 7s.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -220,20 +220,18 @@ jobs:
           # build docs
           cargo doc --target ${{ matrix.target }} --all-features --no-deps
 
-      # wgpu-core docs are not feasible due to <https://github.com/gfx-rs/wgpu/issues/4905>
-      #
-      # - name: check private item docs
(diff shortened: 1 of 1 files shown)


nical:
Looks like you need to re add `DOCS_RUST_VERSION`, here is a revision where it was present: https://github.com/gfx-rs/wgpu/blame/1ced6c6165842d0ffeed2c707d66cb7d74f6b69f/.github/workflows/ci.yml

ErichDonGubler:
@nical: Pretty sure the whole reason we introduced `RUST_DOCS_VERSION` was to work around the issue mentioned in the OP. Just pushed up `fixup!` 6bc11f3 to remove it entirely; it should now run `cargo doc` at MSRV.

ErichDonGubler (approved):

PR #6225, opened 2024-09-05
Resolve lints for Rust 1.78-1.81 that can be preempted before upgrade

ErichDonGubler:
**Connections**

#6105 handles some migrations that are necessary for Rust 1.80 also. Unlike in this PR, it is only possible to implement the migrations there with an upgrade to the MSRV.

**Description**

These outstanding lints are obstacles to eventual upgrade to these Rust versions, and can be resolved now. We should do so!

Each commit is intended to be [atomic](https://www.aleksandrhovhannisyan.com/blog/atomic-git-commits/) for eventual rebase-and-merge.

**Testing**

Things still compile and are green in CI.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/benches/benches/computepass.rs
+++ b/benches/benches/computepass.rs
@@ -447,7 +447,7 @@ fn run_bench(ctx: &mut Criterion) {
             };
 
             group.bench_function(
-                &format!("{cpasses} computepasses x {dispatch_per_pass} dispatches ({label})"),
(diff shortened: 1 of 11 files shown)


nical (approved):
It's a bit annoying to have to annotate things when it's dead code just for emscripten (I blame rustc here). But nice to see some actual dead code going away!

ErichDonGubler:
Screenshot 2024-09-06 at 8 21 32 AM

tfw squash-and-merge 😭

PR #6229, opened 2024-09-06
Invalidate the device when we encounter driver-induced device loss or on unexpected errors

### LINKED ISSUE #1624: Figure out how to deal with device loss
A prerequisite for safely using a lot of APIs is that you (ideally) need to handle device loss somewhat gracefully.  However you obviously can't make sure a GPU doesn't crash between when you check it and when you submit your command (without high precision timers or synchronous communication there is a theorem about this!), so presumably what's needed is to be able to recover gracefully where possible, and abort if not.  Apparently wgpu is not really doing this yet.

From @kvark 
>  I think it's a combination of things. Like at least:

> 1. if we previously got an error code for device lost, we are no longer trying to use it
> 2. we set up the context parameters in a way that handles device loss gracefully, i.e. OpenGL robust context
> But I don't have a fully clear picture about this yet.

So let's get a clearer picture (and figure out what the challenges are on different platforms).

### LINKED ISSUE #5132: Mark the device as lost when hal produces a device lost error
We currently do not detect when hal produces a device lost error. In this situation we should call `Global::device_mark_lost` to fire the device lost callback and prevent the device from issuing hal commands again (other than potentially destroying resources).

See also #4907 which has a broader scope.

teoxoy:
Fixes https://github.com/gfx-rs/wgpu/issues/1624.
Fixes https://github.com/gfx-rs/wgpu/issues/5132.

### DIFF
--- a/wgpu-core/src/command/clear.rs
+++ b/wgpu-core/src/command/clear.rs
@@ -163,7 +163,7 @@ impl Global {
 
         // actual hal barrier & operation
         let dst_barrier = dst_pending.map(|pending| pending.into_hal(&dst_buffer, &snatch_guard));
-        let cmd_buf_raw = cmd_buf_data.encoder.open()?;
+        let cmd_buf_raw = cmd_buf_data.encoder.open(&cmd_buf.device)?;
         unsafe {
             cmd_buf_raw.transition_buffers(dst_barrier.as_slice());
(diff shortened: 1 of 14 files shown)


ErichDonGubler:
Tentatively assigning and requesting review from @jimblandy.

nical (approved):

PR #6241, opened 2024-09-09
[d3d12] handle known error variants returned by `D3D12CreateDevice`

teoxoy:
Follow-up to https://github.com/gfx-rs/wgpu/commit/dd01b6d20967f1b4a74e9a75baec6b54545ebfec (https://github.com/gfx-rs/wgpu/pull/6200).
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1917102#c13


ErichDonGubler (approved):

jimblandy:
This is surprising:

```rust
.ok()??
```

But given that `Result<Option<_>>` needs to be turned into `Option<Option<_>>` in order to use `?` here at all, I guess this is actually the most reasonable approach.

PR #6243, opened 2024-09-09
Move out invalidity from the `Registry`

teoxoy:
Another step towards https://github.com/gfx-rs/wgpu/issues/5121.

This moves out resource invalidity from the `Registry`; by storing a new `Fallible` type in the `Registry` instead.
Note that there are a few types that can't be invalid or have internal invalidity (ex `Device`), those don't need to be wrapped in a `Fallible`.


nical (approved):
Splendid

PR #6249, opened 2024-09-10
Fix initial texture state for injected hal textures

### LINKED ISSUE #6248: Crash when using texture created with create_texture_from_hal
**Description**
Recently I've been creating prototype of texture sharing between processes. I've used my old renderer for quick start - it was based on wgpu 0.18 though which was recently yanked. My code used `device.as_hal` to get vulkan device and then `device.create_texture_from_hal` to inject texture back into wgpu for further usage. wgpu 0.18 was yanked so I decided to upgrade to 22.1. After upgrade I noticed that `device.as_hal` now panics if invoked with wrong api (it's now fixed on trunk, I've managed to work around this). After fixing I noticed that:
1.  queue.write_texture fails with panic! There was [a change](https://github.com/gfx-rs/wgpu/commit/dc55cb436cdc4918b7f03500735b55fe02c4c177#) in texture initialization and now [textures from hal have TextureInitTracker with 0 mips and 0 layers](https://github.com/gfx-rs/wgpu/commit/dc55cb436cdc4918b7f03500735b55fe02c4c177#diff-43224d1cf0ac3a19a362f6a01c57eb66518b946cdab5c28494a3eeee60466befR983) which causes out of bounds access to mips array [here](https://github.com/gfx-rs/wgpu/blob/trunk/wgpu-core/src/device/queue.rs#L709)
2. After doing this change:
```
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -656,7 +656,7 @@ impl<A: HalApi> Device<A> {
             desc,
             format_features,
             resource::TextureClearMode::None,
-            false,
+            true,
         );

         let texture = Arc::new(texture);
```
which makes `create_texture_from_hal` use normal TextureInitTracker queue.write_texture works (at least is not crashing) BUT my receiving process can't read from it. Problems seems to be that create_texture_from_hal doesn't mark texture as initialized any more - it worked in version 0.18. There is no parameter in `create_texture_from_hal` to indicate whether texture is initialized but documentation mentions that "hal_texture must be initialized" so I would assume read from such a texture should be possible.

Error being shown: 
```
thread 'main' panicked at "..."\wgpu\wgpu-core\src\command\memory_init.rs:326:21:
Texture with 'vulkan allocated texture' label can not be cleared
```

**Repro steps**
I've created repro project here: https://github.com/Tsuguri/wgpu_as_hal_crash/blob/master/src/main.rs
My original solution have two processes: One creates vk texture and win32 handle for it and does queue.write_texture and sends it, then receiving process converts win32 handle to vk texture, injects it into wgpu and tries to sample it in a shader. For demonstration purposes here I'm just creating empty vulkan texture, it works (crashes) the same if it's imported from win32 handle.

My expectations:
- no panic at runtime :)
- being able to sample from texture created with `device.create_texture_from_hal`.

**Platform**
I'm working on windows (using KHR_external_memory_win32 vulkan extension), but repro project should probably run on anything since problem seems to be in wgpu logic.

I hope my description makes sense. I'm happy to help in actual solving of this problem but I need some pointers on how to properly do it since I'm not knowledgable enough about wgpu texture status tracking yet.

Tsuguri:
**Connections**
Fixes: https://github.com/gfx-rs/wgpu/issues/6248

**Description**
Brings back TextureInitTracker initial state from before https://github.com/gfx-rs/wgpu/commit/dc55cb436cdc4918b7f03500735b55fe02c4c177

Suggested [here](https://github.com/gfx-rs/wgpu/issues/6248#issuecomment-2340433765)
**Testing**
I've tested this change with test project created for #6248 and inside original project that I worked on.

**Checklist**

- [ ] Run `cargo fmt`.

### DIFF
--- a/wgpu-core/src/resource.rs
+++ b/wgpu-core/src/resource.rs
@@ -1057,7 +1057,7 @@ impl Texture {
                 if init {
                     TextureInitTracker::new(desc.mip_level_count, desc.array_layer_count())
                 } else {
-                    TextureInitTracker::new(0, 0)
+                    TextureInitTracker::new(desc.mip_level_count, 0)
                 },
             ),
             full_range: TextureSelector {


teoxoy (approved):
Thanks!

PR #6256, opened 2024-09-12
spv-out: configure source language in debug info

kvark:
**Connections**
Resolves #6252

**Description**
Tools are sensitive to the kind of shader they get on the input. Setting UNKNOWN was confusing NSight.

**Testing**
Build with debug info, load into NSight.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,7 @@ By @teoxoy [#6134](https://github.com/gfx-rs/wgpu/pull/6134).
 #### Naga
 
 - Accept only `vec3` (not `vecN`) for the `cross` built-in. By @ErichDonGubler in [#6171](https://github.com/gfx-rs/wgpu/pull/6171).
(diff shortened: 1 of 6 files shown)


JMS55:
Does this mean we get the original WGSL source code when profiling shaders in NSight? (on the Vulkan backend of course)

kvark:
Well, *almost*. It turns out that NSight doesn't accept WGSL right now. So the user would have to force GLSL there. That is more an issue with NSight than it is with wgpu though - hopefully they'll unblock it soon.
The goal of the PR is to make it configurable in Naga, I don't have strong preference on what wgpu should do with it.

cwfitzgerald:
I don't think there's any harm in lying that it's glsl, until someone complains as tooling is starting to support it

jimblandy:
I'd like to avoid TODO comments in the source code, because they're just not as good for tracking and prioritizing as filed issues. So I filed #6266 and changed the TODO comment to refer to that.

jimblandy (approved):
LGTM, thank you!

kvark:
Thanks Jim!

PR #6257, opened 2024-09-12
Vulkan: Avoid undefined behaviour with adversarial debug label

DJMcNab:
**Connections**

This method was first added in https://github.com/gfx-rs/wgpu/pull/1471
There was no discussion at the time about this use of unsafe, and the [original code in `gfx-hal`](https://github.com/gfx-rs/gfx/blob/master/src/backend/vulkan/src/lib.rs#L911) does not have this unsoundness.

**Description**
In Rust, it is [valid](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b3cc55a7df1db3a36fd6c9f96f7bb25d) for an `&str` to contain an interior nul byte.
However, the Vulkan debug object name code uses [`CStr::from_bytes_with_nul_unchecked`](https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html#method.from_bytes_with_nul_unchecked) with a user-provided `&str`, on which as far as I can see there is no additional validation.

This is technically an unsoundness fix, but I don't think it's a significant security issue, and so would not recommend backporting this fix.

The tradeoff I have chosen to make in the case of an interior nul-byte is to truncate to that byte.
Alternatives include:
- Replacing it with another ascii byte (`0`?)
- Panicking

It's also not clear to me why `set_object_name` is unsafe, as there are no provided preconditions.
I guessed at some, but would be happier just removing the `unsafe` marker entirely.

**Testing**
This change is untested.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`
- [x] Run `cargo xtask test` to run tests. (Same behaviour as on main of crashing my editor)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,10 @@ By @bradwerth [#6216](https://github.com/gfx-rs/wgpu/pull/6216).
 
 - Fix GL debug message callbacks not being properly cleaned up (causing UB). By @Imberflur in [#6114](https://github.com/gfx-rs/wgpu/pull/6114)
 
+#### Vulkan
+
+- Vulkan debug labels assumed no interior nul byte. By @DJMcNab in [#6257](https://github.com/gfx-rs/wgpu/pull/6257)
+
 ### Changes
 
 - `wgpu_hal::gles::Adapter::new_external` now requires the context to be current when dropping the adapter and related objects. By @Imberflur in [#6114](https://github.com/gfx-rs/wgpu/pull/6114).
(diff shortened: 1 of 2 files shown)


nical (commented):
Oh wow, thanks for catching that. I think that we have to add the trailing null byte here, since the string comes from the wgpu API which expects regular rust strings.

DJMcNab:
I don't understand what you're asking me to change? The code above this line already adds a nul byte - this line of code just doesn't take into account the possibility that there's already an extant nul byte.

nical:
Oh sorry I skimmed over this too quickly. You are right, the fix makes sense. Thanks!

nical (approved):

PR #6272, opened 2024-09-14
chore!: remove `#[no_mangle] … extern "C" …` from `fn`s in `bundle_ffi`

### LINKED ISSUE #3105: Using multiple major versions of wgpu causes linker errors

ErichDonGubler:
**Connections**

Resolves #3105.

**Description**

Remove the last remaining cases that should trigger #3105 in the `wgpu_core::command::bundle::bundle_ffi` module.

**Testing**

If things still compile, we just gotta check that this doesn't break downstream. @cwfitzgerald, halp plz?

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,7 @@ By @bradwerth [#6216](https://github.com/gfx-rs/wgpu/pull/6216).
 - Fix error message that is thrown in create_render_pass to no longer say `compute_pass`. By @matthew-wong1 [#6041](https://github.com/gfx-rs/wgpu/pull/6041)
 - Document `wgpu_hal` bounds-checking promises, and adapt `wgpu_core`'s lazy initialization logic to the slightly weaker-than-expected guarantees. By @jimblandy in [#6201](https://github.com/gfx-rs/wgpu/pull/6201)
 - Raise validation error instead of panicking in `{Render,Compute}Pipeline::get_bind_group_layout` on native / WebGL. By @bgr360 in [#6280](https://github.com/gfx-rs/wgpu/pull/6280).
+- **BREAKING**: Remove the last exposed C symbols in project, located in `wgpu_core::render::bundle::bundle_ffi`, to allow multiple versions of WGPU to compile together. By @ErichDonGubler in [#6272](https://github.com/gfx-rs/wgpu/pull/6272).
 
 #### GLES / OpenGL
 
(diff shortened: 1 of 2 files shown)


Wumpf (approved):
lgtm and good riddance!
But I thought @teoxoy still saw those being used after all in firefox?
Also, @crowlKats can you confirm this either isn't an issue for Deno (or one that you can fix easily enough)

ErichDonGubler:
Ah, using [SearchFox](https://searchfox.org/), I can confirm that we _are_ using these symbols, and you're right. I claimed that FF doesn't use them, and I was/am wrong.

I don't know if that's a deal-breaker, though, because we in Firefox _could_ just force ourselves to move the symbols with our next re-vendor…

Thoughts, @jimblandy and @teoxoy?

nical:
You could (should) do the gecko-side of this before the next re-vendor, so that if teh change requires new APIs whoever re-vendors next isn't stuck going back and forth between the two repos (for something they didn't start).

ErichDonGubler:
I have a patch ready to go on the Firefox side, and I've confirmed it compiles: [D223052](https://phabricator.services.mozilla.com/D223052)

So we should be good to merge this! 😀

ErichDonGubler:
Applying auto-merge, since @crowlKats hasn't responded to the [only open question](https://github.com/gfx-rs/wgpu/pull/6272#pullrequestreview-2305206276) in ~6 days. I believe this should Just Work™, since I see Deno directly using Rust paths to these symbols, i.e.:

https://github.com/gfx-rs/wgpu/blob/14abdd47547e0cd364e8431dec46c3d57be3e160/deno_webgpu/bundle.rs#L150

Issue #6095, opened 2024-08-09
get_bind_group_layout fails when layout is auto


sagudev:
**Description**
When layout is not explicitly provided when creating pipeline, `get_bind_group_layout` fails with `Error reflecting bind group 0: Invalid group index 0`, but explicitly providing it makes it work.

**Repro steps**
Relevant CTS test that fails in firefox (and servo): [webgpu:api,validation,getBindGroupLayout:index_range,auto_layout:*](https://gpuweb.github.io/cts/standalone/?debug=1&runnow=1&q=webgpu:api,validation,getBindGroupLayout:index_range,auto_layout:*
)

I also extracted relevant failing CTS subtest in rust/wgpu repro: https://github.com/sagudev/wgpu-problem/tree/get-auto-bind-group-layout (also contains explicit code that makes it work using `explicit` feature).

**Platform**
Happens on current trunk.

sagudev:
This happens on both compute/render pipeline.

Wumpf:
Duplicate of
* https://github.com/gfx-rs/wgpu/issues/4167

sagudev:
I tested repro with https://github.com/gfx-rs/wgpu/pull/6280 and while it does not panic it still raises validation error (per CTS it should return appropriate bind group), so please reopen the issue.

sagudev:
The problem is that `usage = 0`:
https://github.com/gfx-rs/wgpu/blob/9f85f8aeea6c43c1a412bafc8fbcfb43aad0dd20/wgpu-core/src/validation.rs#L897-L900
for compute pipeline with:
```wgsl
@group(0) @binding(0) var<uniform> binding: f32;

@compute @workgroup_size(1) fn main(
  @builtin(global_invocation_id) id: vec3<u32>
) {
    _ = binding;
}
```

sagudev:
More specifically the problem is that naga eliminates phony assigement, converting wgsl from above to wgsl returns:
```wgsl
@group(0) @binding(0) 
var<uniform> binding: f32;

@compute @workgroup_size(1, 1, 1)
fn main(@builtin(global_invocation_id) id: vec3<u32>) {
}
```

sagudev:
this is regression naga-cli v0.12.3` -> `naga-cli v0.13.0

sagudev:
I think it's https://github.com/gfx-rs/naga/pull/2378, but also we remove expressions on compacting ...

Issue #3767, opened 2023-05-10
Clarify error handling in documentation


cwfitzgerald:
Stemmed from https://github.com/gfx-rs/wgpu/issues/3764

It is quite easy for new users to get baffled by the non-rustic error handling in wgpu right now. We should talk about this more in the documentation so that people can find their way to the appropriate error handling mechanism on their own.

Error handling is handled through three functions:

- [`Device::push_error_scope`](https://docs.rs/wgpu/latest/wgpu/struct.Device.html#method.push_error_scope)
- [`Device::pop_error_scope`](https://docs.rs/wgpu/latest/wgpu/struct.Device.html#method.pop_error_scope)
- [`Device::on_uncaptured_error`](https://docs.rs/wgpu/latest/wgpu/struct.Device.html#method.on_uncaptured_error)

nolanderc:
Something that should be clarified in the docs: what if two threads push multiple error scopes at the same time? Are scopes local to the current thread, or will errors be sent to the thread which opened a scope last, or something else?

On the web, this is less of an issue since everything is single-threaded, but what happens on native?

cwfitzgerald:
This actually wasn't really previously defined - webgpu.h we decided they're going to be thread local. We need to redefine the api in wgpu to use a thread local.

PR #6275, opened 2024-09-14
Document default pipeline layout behavior.

### LINKED ISSUE #6254: Implicit layout behavior of `create_render_pipeline` is not documented.
I was reading wgpu examples when I saw this code:

https://github.com/gfx-rs/wgpu/blob/f3cbd6ce54690544871c8008eb9b2f1b4238e0a8/examples/src/mipmap/mod.rs#L89-L114

This code supplies `layout: None` in `RenderPipelineDescriptor` and then uses `pipeline..get_bind_group_layout(0)` to get a `BindGroupLayout`.

This is new to me; all the examples of wgpu I had seen online up to this point manually specified `BindGroupLayout`s in the `RenderPipelineDescriptor`.

I didn't know what this code should do when reading it. Does `None` mean that the pipeline expects no bindings? Does `None` mean that wgpu will deduce the layout from the shader module(s)? I guessed that it must be the latter, since the shaders in this example do require bindings.

So I went to the docs and looked up [`RenderPipeline::get_bind_group_layout`](https://docs.rs/wgpu/22.1.0/wgpu/struct.RenderPipeline.html#method.get_bind_group_layout) and [`RenderPipelineDescriptor::layout`](https://docs.rs/wgpu/22.1.0/wgpu/struct.RenderPipelineDescriptor.html#structfield.layout), but neither of them mention anything about this implicit layout behavior. In the source code though, I found evidence for "implicit layout" behavior.

### My request: please document the implicit layout behavior of `RenderPipelineDescriptor`

Furthermore, it's not documented what happens if you call `RenderPipeline::get_bind_group_layout` with a bind group index that doesn't exist in the layout.

### Request 2: Document behavior of `RenderPipeline::get_bind_group_layout` when index is out of bounds

Here's what I suggest based on my current understanding. I don't know if it's all correct.

```rust
pub struct RenderPipelineDescriptor<'a> {
    ...
    /// The layout of bind groups for this pipeline.
    ///
    /// If set, `wgpu` will validate that the layout matches what the shader modules expect,
    /// otherwise [`Device::create_render_pipeline`] will panic.
    ///
    /// If `None`, `wgpu` will deduce the pipeline layout from the shader modules. The deduced
    /// bind group layouts can then be read by calling [`RenderPipeline::get_bind_group_layout`].
    pub layout: Option<&'a PipelineLayout>,
```

```rust
impl RenderPipeline {
    /// Get an object representing the bind group layout at a given index.
    ///
    /// If [`RenderPipelineDescriptor::layout`] was set to `None` when creating this pipeline,
    /// this method will return the bind group layout that `wgpu` deduced for this index based
    /// on the shader modules. Otherwise, the returned layout will match the one at this index
    /// in the `RenderPipelineDescriptor`.
    ///
    /// # Panics
    /// If the pipeline has no bind group at the provided `index`, this method will panic.
    pub fn get_bind_group_layout(&self, index: u32) -> BindGroupLayout {
        ...
    }
}
```

BGR360:
**Connections**
Fixes #6254.

**Description**
The behavior of setting `{Render,Compute}PipelineDescriptor::layout` to `None` was not documented. It is now documented, with links to the WebGPU spec providing more information.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo doc -p wgpu --no-deps --lib`

### DIFF
--- a/wgpu/src/api/compute_pipeline.rs
+++ b/wgpu/src/api/compute_pipeline.rs
@@ -20,6 +20,13 @@ super::impl_partialeq_eq_hash!(ComputePipeline);
 
 impl ComputePipeline {
     /// Get an object representing the bind group layout at a given index.
+    ///
+    /// If this pipeline was created with a [default layout][ComputePipelineDescriptor::layout],
+    /// then bind groups created with the returned `BindGroupLayout` can only be used with this
+    /// pipeline.
+    ///
+    /// # Panics
+    /// - There is no bind group layout at the given `index`.
     pub fn get_bind_group_layout(&self, index: u32) -> BindGroupLayout {
         let context = Arc::clone(&self.context);
         let data = self
@@ -48,6 +55,25 @@ pub struct ComputePipelineDescriptor<'a> {
     /// Debug label of the pipeline. This will show up in graphics debuggers for easy identification.
     pub label: Label<'a>,
     /// The layout of bind groups for this pipeline.
+    ///
+    /// If this is set, then `wgpu` will validate that the layout matches what the shader module(s)
+    /// expect, otherwise [`Device::create_render_pipeline`] will panic.
+    ///
+    /// Using the same [`PipelineLayout`] for many [`RenderPipeline`] or [`ComputePipeline`]
+    /// pipelines guarantees that you don't have to rebind any resources when switching between
+    /// those pipelines.
+    ///
+    /// ## Default pipeline layout
+    ///
+    /// If `layout` is `None`, then the pipeline has a [default layout] created and used instead.
+    /// The default layout is deduced from the shader modules.
+    ///
+    /// You can use [`ComputePipeline::get_bind_group_layout`] to create bind groups for use with
+    /// the default layout. However, these bind groups cannot be used with any other pipelines. This
+    /// is convenient for simple pipelines, but using an explicit layout is recommended in most
+    /// cases.
+    ///
+    /// [default layout]: https://www.w3.org/TR/webgpu/#default-pipeline-layout
     pub layout: Option<&'a PipelineLayout>,
     /// The compiled shader module for this stage.
     pub module: &'a ShaderModule,
--- a/wgpu/src/api/compute_pipeline.rs
+++ b/wgpu/src/api/compute_pipeline.rs
@@ -24,9 +24,6 @@ impl ComputePipeline {
     /// If this pipeline was created with a [default layout][ComputePipelineDescriptor::layout],
     /// then bind groups created with the returned `BindGroupLayout` can only be used with this
     /// pipeline.
-    ///
-    /// # Panics
-    /// - There is no bind group layout at the given `index`.
     pub fn get_bind_group_layout(&self, index: u32) -> BindGroupLayout {
         let context = Arc::clone(&self.context);
         let data = self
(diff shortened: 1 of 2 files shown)


wgpu/src/api/compute_pipeline.rs
```diff
@@ -20,6 +20,13 @@ super::impl_partialeq_eq_hash!(ComputePipeline);
 
 impl ComputePipeline {
     /// Get an object representing the bind group layout at a given index.
+    ///
+    /// If this pipeline was created with a [default layout][ComputePipelineDescriptor::layout],
+    /// then bind groups created with the returned `BindGroupLayout` can only be used with this
+    /// pipeline.
+    ///
+    /// # Panics
+    /// - There is no bind group layout at the given `index`.
```
Wumpf: [not actionable, mostly note to self]

urgh this _actually_ panics. 
For future reference: This should be a validation error and return an invalid object, see https://developer.mozilla.org/en-US/docs/Web/API/GPUComputePipeline/getBindGroupLayout#validation

wgpu/src/api/compute_pipeline.rs
```diff
@@ -48,6 +55,25 @@ pub struct ComputePipelineDescriptor<'a> {
     /// Debug label of the pipeline. This will show up in graphics debuggers for easy identification.
     pub label: Label<'a>,
     /// The layout of bind groups for this pipeline.
+    ///
+    /// If this is set, then `wgpu` will validate that the layout matches what the shader module(s)
+    /// expect, otherwise [`Device::create_compute_pipeline`] will panic.
```
Wumpf:
```suggestion
    /// expect, otherwise [`Device::create_compute_pipeline`] will cause a validation error.
```

Should only panic with the default error handler.
(see also https://github.com/gfx-rs/wgpu/issues/3767)

wgpu/src/api/render_pipeline.rs
```diff
@@ -121,6 +127,24 @@ pub struct RenderPipelineDescriptor<'a> {
     /// Debug label of the pipeline. This will show up in graphics debuggers for easy identification.
     pub label: Label<'a>,
     /// The layout of bind groups for this pipeline.
+    ///
+    /// If this is set, then `wgpu` will validate that the layout matches what the shader module(s)
+    /// expect, otherwise [`Device::create_render_pipeline`] will panic.
```
Wumpf: as above:
```suggestion
    /// expect, otherwise [`Device::create_render_pipeline`] will cause a validation error.
```

Wumpf (changes requested):
very nice thank you!
There's some fine print about `panic` here. Tbh I'm not entirely sure how to rectify this best: mismatching layout _is_ a panic by default. But as a user it doesn't have to be and that's a fairly important distinction. The full story is a bit lengthy though and (see #3767) not super well documented
Imho the suggestion I wrote in the comment works, but is not great.

wgpu/src/api/compute_pipeline.rs
BGR360: > this *actually* panics

By this, do you mean that there's a bug here? I.e. `wgpu` should be catching a validation error and doing the proper thing with it, but it's not?

Or do you just mean to say "ah yes, I suppose this *does* panic if the user doesn't set any error scopes."

Wumpf: Looked closer again to be sure: Today it panics on wgpu-core which is a simply a bug. Turns out this also got filed already (didn't check before).. twice in fact. 
* https://github.com/gfx-rs/wgpu/issues/4167
* https://github.com/gfx-rs/wgpu/issues/6095

...closing one of them as dupl

Considering that, I think the doc here should say that this is a validation error and then add a note that there is a known bug that this panics

BGR360: Just FYI, I actually didn't go and research for myself whether it panics. I just made an assumption and hoped that my code reviewer would correct me.

Looking at the implementation, it looks like it does the proper thing on WebGPU, but not on wgpu_core. Given that, I don't think we should put anything in the public docs until we fix wgpu_core to raise this as a validation error.

BGR360: Oops, I responded before I refreshed and saw your reply. The known bug option sounds fine too.

BGR360: Okay actually I changed my mind, I'm going to omit any documentation about panics or errors until webgpu and wgpu-core have the same behavior.

Wumpf: fine by me! Didn't want to bog down this otherwise fine doc addition :)
Bit of a wild west situation this particular thing at this point T_T

Wumpf (approved):
thanks again, very appreciated!

PR #6276, opened 2024-09-15
naga: Don't consider per-polygon inputs to be subgroup uniform

### LINKED ISSUE #6270: Non uniform indices are not always properly handled by naga
**Description**
We had an off error that only occured on AMD GPUs (Vega, RDNA1 and RDNA2), where under Vulkan and DX12 textures where rendered with the wrong texel information (see attached screenshot). We use the "SAMPLED_TEXTURE_AND_STORAGE_BUFFER_ARRAY_NON_UNIFORM_INDEXING" feature to select textures based on non-uniform vertex data inside the fragment shader. We found out, that the AMD devices sometimes used the wrong texture for some texels. We debugged the problem under Vulkan and saw, that sometimes in the SPIRV output the indexes used for indexing the binding array witht the texture where not properly markes as non-uniform indexes.

We traces the problem back to naga, which will not emmit the necessary marker for non-unified indexes when using fragment shader with no wrapper struct for the function parameter. We reproduces the problem using the texture_arrays example.

**Repro steps**
Use the current trunk branch (eb18854b46a4d6c973a1981be6739ec19282f2f3) and remove the wrapper struct for the fragment input:

```wgsl
@group(0) @binding(0)
var texture_array_top: binding_array<texture_2d<f32>>;
@group(0) @binding(1)
var texture_array_bottom: binding_array<texture_2d<f32>>;
@group(0) @binding(2)
var sampler_array: binding_array<sampler>;

@fragment
fn non_uniform_main(
    @location(0) tex_coord: vec2<f32>,
    @location(1) index: i32,
) -> @location(0) vec4<f32> {
    var outval: vec3<f32>;
    if tex_coord.y <= 0.5 {
        outval = textureSampleLevel(
            texture_array_top[index],
            sampler_array[index],
            tex_coord,
            0.0
        ).rgb;
    } else {
        outval = textureSampleLevel(
            texture_array_bottom[index],
            sampler_array[index],
            tex_coord,
            0.0
        ).rgb;
    }

    return vec4<f32>(outval.x, outval.y, outval.z, 1.0);
}
```

This will result in SPIRV code which does not have the expected non-uniform index marker in it. It looks like this in HLSL when de-compiled with SPIRV cross:

```hlsl
void frag_main()
{
    float3 outval = 0.0f.xxx;
    if (tex_coord.y <= 0.5f)
    {
        outval = texture_array_top[index].SampleLevel(sampler_array[index], tex_coord, 0.0f).xyz;
    }
    else
    {
        outval = texture_array_bottom[index].SampleLevel(sampler_array[index], tex_coord, 0.0f).xyz;
    }
    _29 = float4(outval.x, outval.y, outval.z, 1.0f);
}
```

Normally the indexes should be marked as non-uniform indexes like (using the original texture_arrays code):

```hlsl
void frag_main()
{
    float3 outval = 0.0f.xxx;
    FragmentInput _24 = { tex_coord, index };
    if (_24.tex_coord.y <= 0.5f)
    {
        outval = texture_array_top[NonUniformResourceIndex(_24.index)].SampleLevel(sampler_array[NonUniformResourceIndex(_24.index)], _24.tex_coord, 0.0f).xyz;
    }
    else
    {
        outval = texture_array_bottom[NonUniformResourceIndex(_24.index)].SampleLevel(sampler_array[NonUniformResourceIndex(_24.index)], _24.tex_coord, 0.0f).xyz;
    }
    _31 = float4(outval.x, outval.y, outval.z, 1.0f);
}
```

It is also possible to skip the wgpu part and directly convert both WGSL shader versions with the help of naga-cli:

```sh
naga non_uniform_indexing.wgsl non_uniform_indexing.hlsl
```

Non-uniform indexes should properly marked as such in the resulting shader modules.

**Extra materials**
XInRsc6

**Platform**
WGPU stable (22.1) and WGPU trunk. Nvidia and Intel devices will accept the SPIRV modules and render them without any problem, but AMD devices on either Windows and Linux will render with artifacts (see screenshot).

magcius:
Implementations can absolutely pack multiple triangles per subgroup.

Fixes #6270.

**Connections**

**Description**

**Testing**

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,7 @@ By @bradwerth [#6216](https://github.com/gfx-rs/wgpu/pull/6216).
 
 - Accept only `vec3` (not `vecN`) for the `cross` built-in. By @ErichDonGubler in [#6171](https://github.com/gfx-rs/wgpu/pull/6171).
 - Configure `SourceLanguage` when enabling debug info in SPV-out. By @kvark in [#6256](https://github.com/gfx-rs/wgpu/pull/6256)
+- Per-polygon and flat inputs should not be considered subgroup uniform. By @magcius in [#6276](https://github.com/gfx-rs/wgpu/pull/6276).
 
 #### General
 
--- a/naga/src/valid/analyzer.rs
+++ b/naga/src/valid/analyzer.rs
@@ -589,23 +589,16 @@ impl FunctionInfo {
                     requirements: UniformityRequirements::empty(),
                 }
             }
-            // depends on the builtin or interpolation
+            // depends on the builtin
             E::FunctionArgument(index) => {
                 let arg = &resolve_context.arguments[index as usize];
                 let uniform = match arg.binding {
                     Some(crate::Binding::BuiltIn(
-                        // per-polygon built-ins are uniform
-                        crate::BuiltIn::FrontFacing
                         // per-work-group built-ins are uniform
-                        | crate::BuiltIn::WorkGroupId
+                        crate::BuiltIn::WorkGroupId
                         | crate::BuiltIn::WorkGroupSize
-                        | crate::BuiltIn::NumWorkGroups)
-                    ) => true,
-                    // only flat inputs are uniform
-                    Some(crate::Binding::Location {
-                        interpolation: Some(crate::Interpolation::Flat),
-                        ..
-                    }) => true,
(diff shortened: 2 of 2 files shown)


cwfitzgerald (approved):
I think I have the experience to give this one a boop

Issue #4444, opened 2021-11-16
Make buffers coherent by default


kvark:
See https://github.com/gpuweb/gpuweb/issues/1621 and gfx-rs/naga#1488
TL;DR: default behavior in WebGPU is making buffers coherent, so this task at the minimum should add the logic to all backends to enforce coherence. At the maximum, we can control it in IR.

jimblandy:
Here are the recent changes to Dawn that implemented this for Vulkan:

https://dawn-review.googlesource.com/c/dawn/+/202995
https://dawn-review.googlesource.com/c/dawn/+/203495
https://dawn-review.googlesource.com/c/dawn/+/203654

They say that they believe nothing is needed for HLSL.

PR #6277, opened 2024-09-15
Coherent storage

JMS55:
**Connections**
Related to https://github.com/gfx-rs/wgpu/issues/4444

**Description**
Implement coherent read_write storage buffers (and storage textures?). Writes to this variable from one workgroup should be visible to another workgroup.

Syntax is `var<storage, read_write, coherent>: foo`.

**Questions/TODO**
* Where in naga's AST should this qualifier live?
* Is this allowed on storage textures?
* We need a new capability in naga, and then where do we validate this?
* Where do we validate that this is only valid with read_write access mode?
* Metal support? (requires metal 3.2)

Generally everything around the frontend I have little idea what I'm doing. 

The backend is very easy for spv (add a decoration) and hlsl (prefix RWStructuredBuffer/RWByteAddressBuffer/RWTexture2D), and mostly on glsl (goes somewhere in the variable qualifiers, order unclear), and metal I haven't even looked at.

**Testing**

**Checklist**

- [ ] Run `cargo fmt`.

### DIFF
--- a/naga/src/back/glsl/mod.rs
+++ b/naga/src/back/glsl/mod.rs
@@ -1166,8 +1166,12 @@ impl<'a, W: Write> Writer<'a, W> {
             }
         }
 
-        if let crate::AddressSpace::Storage { access } = global.space {
+        if let crate::AddressSpace::Storage { access, coherent } = global.space {
             self.write_storage_access(access)?;
+
+            if coherent {
+                write!(self.out, "coherent ")?;
+            }
         }
 
         if let Some(storage_qualifier) = glsl_storage_qualifier(global.space) {
--- a/naga/src/back/hlsl/help.rs
+++ b/naga/src/back/hlsl/help.rs
@@ -1052,7 +1052,7 @@ impl<'a, W: Write> super::Writer<'a, W> {
                         }
                     };
                     let storage_access = match global_var.space {
-                        crate::AddressSpace::Storage { access } => access,
+                        crate::AddressSpace::Storage { access, .. } => access,
                         _ => crate::StorageAccess::default(),
                     };
                     let wal = WrappedArrayLength {
--- a/naga/src/back/hlsl/writer.rs
+++ b/naga/src/back/hlsl/writer.rs
@@ -829,13 +829,14 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
                 write!(self.out, "cbuffer")?;
                 "b"
             }
-            crate::AddressSpace::Storage { access } => {
+            crate::AddressSpace::Storage { access, coherent } => {
+                let globallycoherent = if coherent { "globallycoherent " } else { "" };
                 let (prefix, register) = if access.contains(crate::StorageAccess::STORE) {
                     ("RW", "u")
                 } else {
                     ("", "t")
                 };
-                write!(self.out, "{prefix}ByteAddressBuffer")?;
+                write!(self.out, "{globallycoherent}{prefix}ByteAddressBuffer")?;
                 register
             }
             crate::AddressSpace::Handle => {
@@ -3463,7 +3464,7 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
                 };
 
                 let storage_access = match var.space {
-                    crate::AddressSpace::Storage { access } => access,
+                    crate::AddressSpace::Storage { access, .. } => access,
                     _ => crate::StorageAccess::default(),
                 };
                 let wrapped_array_length = WrappedArrayLength {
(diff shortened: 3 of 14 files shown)


jimblandy (changes requested):
So, this is something I'm not really familiar with either but:

I think what we want here is something much simpler: just have the SPIR-V back end slap the `Coherent` decoration on all writable storage buffers. This seems to be [what Dawn does](https://dawn.googlesource.com/dawn/+/main/src/tint/lang/spirv/writer/printer/printer.cc#2329).

jimblandy:
I'm not seeing any code in Dawn to add the qualifiers in HLSL or GLSL. Is Bevy sure those are necessary?

JMS55:
> I think what we want here is something much simpler: just have the SPIR-V back end slap the Coherent decoration on all writable storage buffers

We could do that... and wgpu should for the spec, according to #4444. But it would probably cost a good bit of performance. From what I understand, it causes a cache flush every time you write to the buffer. Not great for general purpose, hence why I wanted this to be configurable (and the WebGPU backend could default to just making everything read_write_coherent in the future).

Coherent buffers basically solve the problem that when you do a buffer write, it's not visible to any other threads. You can call storageBarrier() and force it to be visible to the workgroup, but it's still not visible to other workgroups. Coherent buffers flush the caches to force visibility to other workgroups, which is useful for some advanced techniques.

I think the spv backend might actually be a bit more complicated though, there's implications with the vulkan memory model (no idea if wgpu has this enabled or not) that complicate things

> Coherent is not allowed if the declared [memory model](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Memory_Model) is Vulkan. The [memory operand](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Memory_Operands) bits MakePointerAvailable and MakePointerVisible or the [image operand](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Image_Operands) bits MakeTexelAvailable and MakeTexelVisible can be used instead.

Tldr; this is a useful feature, and needed for WebGPU apparently, but costs perf and is complicated and I don't quite know what I'm doing

jimblandy:
Yes, eventually we will want a way to disable it selectively. But I don't think that's what Bevy needs, right? If that's so, then let's leave user control to some later PR, when someone needs it.

jimblandy:
> From what I understand, it causes a cache flush every time you write to the buffer.

I'm almost positive this is not the case. The committee would never have settled on coherence as the default if it had this kind of cost.

jimblandy:
It's forbidden to use the `Coherent` decoration in the Vulkan memory model, and our SPIR-V output always requests the [GLSL 450 memory model](https://github.com/gfx-rs/wgpu/blob/fc85e4f970a74e69916550a24769065914be8b29/naga/src/back/spv/writer.rs#L2070-L2076).

cwfitzgerald:
From https://gpuweb.github.io/gpuweb/wgsl/#private-vs-non-private

> All non-atomic [read accesses](https://gpuweb.github.io/gpuweb/wgsl/#read-access) in the [storage](https://gpuweb.github.io/gpuweb/wgsl/#address-spaces-storage) or [workgroup](https://gpuweb.github.io/gpuweb/wgsl/#address-spaces-workgroup) address spaces are considered [non-private](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-model-non-private) and correspond to read operations with NonPrivatePointer | MakePointerVisible memory operands with the Workgroup scope.

> All non-atomic [write accesses](https://gpuweb.github.io/gpuweb/wgsl/#write-access) in the [storage](https://gpuweb.github.io/gpuweb/wgsl/#address-spaces-storage) or [workgroup](https://gpuweb.github.io/gpuweb/wgsl/#address-spaces-workgroup) address spaces are considered [non-private](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-model-non-private) and correspond to write operations with NonPrivatePointer | MakePointerAvailable memory operands with the Workgroup scope.

This seems to suggest the coherency is just within the workgroup level? My hunch is that this requires Coherent on vulkan but not globallycoherent on HLSL. We should chat with the dawn folks, who have thought about this more.

JMS55:
Hmm looking at the glsl spec for coherent, you also need an appropriate barrier. I'm not quite sure how any of this works anymore 😓 .

JMS55:
Going to close this, as I've run out of free time to work on it. That said, I'm remembering that the reason I started this was because it blocks some stuff in my engine... I might either come back to this, or have to use spv-passthrough for now.

PR #6285, opened 2024-09-16
[naga msl-out] Defeat the MSL compiler's infinite loop analysis.

### LINKED ISSUE #4972: Metal Shading Language compiler treats infinite loops as UB

jimblandy:
See the comments in the code for details.

This patch emits the definition of the macro only when the first loop is encountered. This does make that first loop's code look a bit odd: it would be more natural to define the macro at the top of the file. (See the modified files in `naga/tests/out/msl`.)

Rejected alternatives:

- We could emit the macro definition unconditionally at the top of the file. But this changes every MSL snapshot output file, whereas only eight of them actually contain loops.

- We could have the validator flag modules that contain loops. But the changes end up being not small, and spread across the validator, so this seems disproportionate. If we had other consumers of this information, it might make sense.

- We could change the MSL backend to allow text to be generated out of order, so that we can decide whether to define the macro after we've generated all the function bodies. But at the moment this seems like unnecessary complexity, although it might be worth doing in the future if we had additional uses for it - say, to conditionally emit helper function definitions.

Fixes #4972.

### DIFF
--- a/naga/src/back/msl/writer.rs
+++ b/naga/src/back/msl/writer.rs
@@ -376,6 +376,11 @@ pub struct Writer<W> {
     /// Set of (struct type, struct field index) denoting which fields require
     /// padding inserted **before** them (i.e. between fields at index - 1 and index)
     struct_member_pads: FastHashSet<(Handle<crate::Type>, u32)>,
+
+    /// Name of the loop reachability macro.
+    ///
+    /// See `emit_loop_reachable_macro` for details.
+    loop_reachable_macro_name: String,
 }
 
 impl crate::Scalar {
@@ -665,6 +670,7 @@ impl<W: Write> Writer<W> {
             #[cfg(test)]
             put_block_stack_pointers: Default::default(),
             struct_member_pads: FastHashSet::default(),
+            loop_reachable_macro_name: String::default(),
         }
     }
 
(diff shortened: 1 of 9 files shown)


naga/src/back/msl/writer.rs
```diff
@@ -675,6 +681,125 @@ impl<W: Write> Writer<W> {
...
+
+        self.loop_reachable_macro_name = self.namer.call("LOOP_IS_REACHABLE");
+        let loop_reachable_volatile_name = self.namer.call("unpredictable_jump_over_loop");
+        writeln!(
+            self.out,
+            "#define {} if (volatile bool {} = true; {})",
+            self.loop_reachable_macro_name,
+            loop_reachable_volatile_name,
+            loop_reachable_volatile_name,
+        )?;
```
ErichDonGubler, line 798: thought: IIRC, this approach is effectively cribbed from the Chromium WebGPU team's Dawn component (i.e., https://github.com/gfx-rs/wgpu/issues/4972#issuecomment-2326848790). It feels important to cite this as an influence in our implementation of this workaround, and not allow a reader to assume that this understanding is entirely our own (which is what I think a typical reader will do with the lengthy explanation above).

☝🏻This can be punted to a follow-up commit.

ErichDonGubler (approved):

naga/src/back/msl/writer.rs
jimblandy: Absolutely - I neglected to give credit where it was due.

PR #6287, opened 2024-09-17
Update parking_lot dependency

mahkoh:
**Connections**

None

**Description**

wgpu-hal does not compile with parking_lot versions that don't contain const Mutex::new.

**Testing**

Not at all. I used the github web UI for this change.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -144,6 +144,10 @@ By @bradwerth [#6216](https://github.com/gfx-rs/wgpu/pull/6216).
 
 - Replace `winapi` code to use the `windows` crate. By @MarijnS95 in [#5956](https://github.com/gfx-rs/wgpu/pull/5956) and [#6173](https://github.com/gfx-rs/wgpu/pull/6173)
 
+#### HAL
+
+- Update `parking_lot` to `0.12`. By @mahkoh in [#6287](https://github.com/gfx-rs/wgpu/pull/6287)
+
 ## 22.0.0 (2024-07-17)
 
 ### Overview
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -105,7 +105,7 @@ noise = { version = "0.8", git = "https://github.com/Razaekel/noise-rs.git", rev
 nv-flip = "0.1"
 obj = "0.10"
 once_cell = "1.19.0"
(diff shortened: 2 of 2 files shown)


Wumpf:
You literally replaced `parking_lot = ">=0.11, <0.13"` with `parking_lot = "0.12.1"`. I don't get the point: Before you could already set this version in your application without duplicated dependencies. This just narrows the range to a single already included version

Re-open if there's some actual issue with the current version range.

mahkoh:
>Re-open if there's some actual issue with the current version range.

wgpu-hal does not compile with parts of the  version range.

Wumpf:
Then it's probably better to fix that instead?
Can you provide more detail on which versions don't work out and what exact errors you're getting?

mahkoh:
>Can you provide more detail on which versions don't work out and what exact errors you're getting?

Any version which does not contain https://github.com/Amanieu/parking_lot/commit/cbab7b0f99097f864e53bccdcf8b47846e1e332f. The error is that you try to invoke a non-const function in const context.

cwfitzgerald:
This update is fine, one concern 

@ErichDonGubler just double checking we're good to force ff on parking_lot 0.12?

ErichDonGubler (approved):

ErichDonGubler:
Pretty sure we should be good; `parking_lot_core`'s latest version is 0.9.9, which uses `windows-targets` 0.52, which Firefox has a fake force-upgrade crate for (see [`build/rust/windows-targets/Cargo.toml`:3](https://searchfox.org/mozilla-central/rev/f549a50b1e39b1e6bea19912d92545c4c0a06b7b/build/rust/windows-targets/Cargo.toml#3)). CC @jimblandy and @teoxoy for 👀.

<sub>If it doesn't work, I'll be learning a lot as I effect a reversion! 🤪 </sub>

ErichDonGubler:
@mahkoh: Could you please add a `CHANGELOG` entry before we merge this? Thanks for the contribution, and in advance for the entry! 😃

Wumpf:
(apologies for the noise earlier, should have read this more properly!)

mahkoh:
> @mahkoh: Could you please add a `CHANGELOG` entry before we merge this? Thanks for the contribution, and in advance for the entry! 😃

Done

ErichDonGubler (approved):

PR #3631, opened 2023-04-01
[wgpu-core] Inline RayQuery Support

daniel-keitel:
**Checklist**

- [ ] Run `cargo clippy`.
- [ ] Run `RUSTFLAGS=--cfg=web_sys_unstable_apis cargo clippy --target wasm32-unknown-unknown` if applicable.
- [ ] Add change to CHANGELOG.md. See simple instructions inside file.

**Future Work**
- [ ] Acceleration Structure Compaction
- [ ] Ray tracing pipelines
- [ ] Ray tracing resource management

**Connections**
#1040
Depends on: #3507 (vulkan hal)
[Matrix Room Thread](https://matrix.to/#/!FZyQrssSlHEZqrYcOb:matrix.org/$vuQn2Lx30kOuFttuHHTpVZbKcZvCMScAMzTDi8r3s2A?via=matrix.org&via=mozilla.org&via=ralith.com)

**Description**
Ray tracing support for wgpu and wgpu-core.
This PR implements support for acceleration structures and ray queries.

It provides a create function each for bottom and top level acceleration structures,
and a build function for the two.
Top level acceleration structures can be used in bind groups.
Shader support is already implemented in naga. 

image

**TODO**
- [x] more validation
- [ ] as_hal/into_hal
- [x] tracing
- [ ] more examples
- [ ] more tests
- [ ] documentation
- [ ] specification
- [ ] procedural geometry
- [x] the safe and validated build function

**Testing**
Tested with example application (ray-cube) and test.
Needs more tests and examples.

Issue #1040, opened 2020-11-23
Ray Tracing Support


5nefarious:
Khronos has [recently released](https://www.khronos.org/blog/vulkan-ray-tracing-final-specification-release) the final specification for ray tracing on Vulkan. At this point, DX12, Vulkan, and [Metal](https://developer.apple.com/documentation/metalperformanceshaders/metal_for_accelerating_ray_tracing) all seem to have some form of acceleration for ray tracing. Are there any plans to eventually consolidate and expose these APIs in `wgpu` and `wgpu-rs`?

I foresee real-time rendering making increasing use of ray tracing in the future, so this may be an essential feature to have. However, I imagine it may be very difficult to support this on the DX11 and OpenGL backends through software emulation.

kvark:
See also:
- https://github.com/gfx-rs/wgpu-rs/issues/247 (dupe in wgpu-rs)
- https://github.com/gfx-rs/gfx/issues/2418 (graphics abstraction we use)
- https://github.com/gpuweb/gpuweb/issues/535 (Web upstream feature request)

It's too early to consider this an essential feature on all backends, but rolling it out at least where the backends *do* have support for it seems very useful today.

Sebbl0508:
Are there any updates on this?

cwfitzgerald:
Not currently. There's interest from upstream WebGPU, but that would be well after v1. 

We could make it a native extension and I think it would be helpful, but the apis are quite large (including shader side transformations), and given very few of us actually have hardware to support this, the chance of it happening any time soon without a champion is low.

OriginLive:
Hi, when is this comming out?

cwfitzgerald:
There's no one actively working on it to my knowledge.

OriginLive:
It would be nice to atleast get native extension or some bindings. Would be a nice start

expenses:
I'm curious about getting basic ray-tracing support working. I think that a first draft wouldn't be too hard, with some limitations. I think we'd initially want to:

- Focus on Vulkan
- Ignore the full ray tracing pipeline (https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_ray_tracing_pipeline.html) and just focus on ray queries (https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_ray_query.html)
[...]

PR #5760, opened 2024-05-31
Add missing acceleration structure buffer flag

Vecvec:
**Connections**

**Description**
Newer versions of the vulkan validation layers give this error when running wgpu hal ray-traced-triangle:
````cmd
[2024-05-31T05:50:53Z ERROR wgpu_hal::vulkan::instance] VALIDATION [VUID-vkGetAccelerationStructureDeviceAddressKHR-pInfo-09542 (0xe4b63bce)]
        Validation Error: [ VUID-vkGetAccelerationStructureDeviceAddressKHR-pInfo-09542 ] | MessageID = 0xe4b63bce | vkGetAccelerationStructureDeviceAddressKHR(): pInfo->accelerationStructure.buffer was created with usage flag(s) VK_BUFFER_USAGE_2_ACCELERATION
_STRUCTURE_STORAGE_BIT_KHR. The Vulkan spec states: The buffer on which pInfo->accelerationStructure was placed must have been created with the VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT usage flag (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkGetAccelerationStructureDeviceAddressKHR-pInfo-09542)
````
This adds the `VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT` flag to fix this error

**Testing**
No other validation error when running ray-traced-triangle.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`. If applicable, add:
   not applicable because this is not on the web
  - `--target wasm32-unknown-unknown`
  -  `--target wasm32-unknown-emscripten` 
- Run `cargo xtask test` to run tests.
- [ ] Add change to `CHANGELOG.md`. See simple instructions inside file. - Should this be done for such a small change?

PR #6291, opened 2024-09-18
Ray Queries

Vecvec:
**Connections**
Works towards #1040 

**Description**
This PR provides BLASes (bottom level acceleration structures), TLASes (top level acceleration structures), TLAS instances (which contain BLASes and data, like transforms, about them), TLAS packages (which contain vectors of TLAS instances). 

**Testing**
Running tests & examples included in PR

This a updated version of #3631 and is intended to replace it.

**Checklist**

- [x] Run `cargo fmt`.
  - n/a `--target wasm32-unknown-unknown`
  - n/a `--target wasm32-unknown-emscripten`
- [x] More tests & examples.
- [x] More docs

**Later** (follow-up PR)
- Acceleration structure compaction.
- Procedural geometry.
- `as_hal` methods


JMS55:
I'm super excited for this!

What makes this PR a draft? What do you think is missing?

Vecvec:
Examples, there is no case for no index buffer (but I'm working on it, hopefully done in a day or so).

nical:
This looks very exciting. This is a pretty large PR and to my knowledge, none of the wgpu maintainers are ray tracing experts. So to facilitate the review process, while we try to ramp up our knowledge of the domain, could you summarize here the context (that is scattered in #1040 and other places)? In particular:
 - The general ideas behind the design of the API compared to DX12/vulkan/metal:
   - What subset of the functionality does the design cover.
   - Tradeoffs (if any) compared to the underlying ray tracing APIs.
 - The scope of this work (for example, only implementing a specific portion of the overall plan).
 - Validation:
   - What are the important aspects of validation to be aware of for safety? For example preventing undefined behavior and out-of-bound memory accesses in shaders.
   - Any missing safety validation in this PR? Not saying the initial implementation should be bullet-proof enough to be a candidate for the WebGPU specification, but we should be aware of the shortcuts we take if any.
 - Any other important things to know (For example tricky aspects of mapping to this or that API)?

Vecvec:
To preface this, I came later to this (after most of the design was finalized), so I may not know all of the missing validation, tradeoffs, etc. I will be talking about the _safe_ tlas building (blas building is always safe but there is a extra unsafe function to do unsafe tlas building which might (?) be faster)

To my knowledge the only missing pieces of validation are ~~the vertex formats when creating a BLAS, metal (not yet implemented, but a consideration) only allows f32x3~~ (edit : finished), and tlas packages (which are only in wgpu) being able to have a different max instances from the max instances a tlas was created with (though I can't remember if it gets checked when building), this one is hard to validate, and later it might be a good idea to unify the tlases and tlas packages (especially if build acceleration structures unsafe gets removed).

The main tradeoffs are not exposing the scratch buffers because they are harder to validate (possibly) and could cause confusion (especially since most of that is lower level than wgpu provides, and also could be accidently be read back, giving garbage info and leaving the user confused) , which is why in an attempt to mitigate the allocation time blas building and tlas building are unified. 

This implementation lacks compaction, which was in the initial API design, but needs another HAL function. Procedural geometry (naga needs to implement this, and if most people want to partially support this they will mainly use triangles). I believe there is a full wgpu API (though without ray-tracing pipelines)

Extra things to note: this is just for wgpu and wgpu-core, naga support (for ray queries) was added earlier. There is a ray-tracing pipeline that would be nice to have (it could allow a ray-tracer - e.g path-tracer - to be extended). 

Edit: wgpu-hal support was also added earlier, but only for vulkan (so some things here still have to consider metal and dx12, though dx12 also was considered)

The naming is slightly confusing, this might be called ray-tracing, but something that uses this API to create something that simulates light bouncing around a scene might also be said to be ray-tracing (this does not do the second thing).

I think that is everything. Tell me if there are other questions about the API.

Edit # 2:  mostly api is from https://github.com/gfx-rs/wgpu/issues/1040#issuecomment-1484274170

Vecvec:
I've realised my previous comment requies quite some preexisting raytracing knowledge, so here are the API concepts:
- BLASes
    - A BLAS contains geometry in a device readable format, you can't interact directly with this, instead you have to build the BLAS with the buffers containing triangles or AABB.
- TLAS instances
    - A TLAS instance contains 
        - A BLAS (or in native APIs it's handle)
        - A transformation matrix for the BLAS
        - Culling flags (if shader cull mask & instance.mask != 0 [from vulkan](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkAccelerationStructureInstanceKHR.html) this can be hit)
        - A custom index (this is data provided into the shader, unused elsewhere)
- TLASes
    - A TLAS contains TLAS instances in a device readable form, you cant interact directly with these either, instead you have to build the TLAS with TLAS instances
    - Note: most of the native APIs require the TLAS instances on a buffer. Wgpu's safe API however requires TLAS instances on the CPU, this is because native APIs have a handle to the BLAS in the TLAS instance, but this is almost impossible to validate. Having TLAS instances in a buffer is supported by `build_acceleration_structures_unsafe_tlas` and is why it is unsafe.
- TLAS packages
    - A TLAS package contains
        - A TLAS
        - TLAS Instances
    - Note: because TLASes build with their data in a buffer this does not exist in native APIs and is only used in safe building in wgpu

All metal, vulkan, and DXR use roughly these concepts (and they are based on one another).

Vecvec:
I've noticed that a validation fix for spv (in trunk now see #5760) was duplicated in this PR, so I've removed that. (I think this was why gfx-rs/naga was requested for review)

wgpu-core/src/command/ray_tracing.rs
```diff
@@ -0,0 +1,1225 @@
...
+                + 1,
+        )
+        .unwrap();
+
+        #[cfg(feature = "trace")]
+        let trace_blas: Vec<crate::ray_tracing::TraceBlasBuildEntry> = blas_iter
+            .map(|x| {
+                let geometries = match x.geometries {
+                    BlasGeometries::TriangleGeometries(triangle_geometries) => {
+                        crate::ray_tracing::TraceBlasGeometries::TriangleGeometries(
```
nical: nitpick: let's `use` symbols like `crate::ray_tracing::TraceBlasGeometries` at the top of the file to avoid the long names.
Let's also `use crate::device::trace` to shorten these as well (keeping the `trace` part visible in the code is valuable in this case).

wgpu-core/src/command/ray_tracing.rs
```diff
@@ -0,0 +1,1225 @@
...
+
+        let build_command_index = NonZeroU64::new(
+            device
+                .last_acceleration_structure_build_command_index
+                .fetch_add(1, std::sync::atomic::Ordering::Relaxed)
+                + 1,
+        )
+        .unwrap();
+
+        #[cfg(feature = "trace")]
```
nical, line 104: I think that it would be easier to read if this sequence of conditionally compiled code was all in a 

```rust
#[cfg(feature = "trace")]
{
     // move the code in here
}
```

block.

wgpu-core/src/command/ray_tracing.rs
```diff
@@ -0,0 +1,1225 @@
...
+                if let Some(last) = buf_storage.last_mut() {
+                    last.5 = Some(blas.clone());
+                }
+            }
+        }
+    }
+    Ok(())
+}
+
+/// Iterates over the buffers generated [iter_blas] and convert the barriers into hal barriers, and the triangles into [hal::AccelerationStructureEntries] (and also some validation).
```
nical:
```suggestion
/// Iterates over the buffers generated in [iter_blas], convert the barriers into hal barriers, and the triangles into [hal::AccelerationStructureEntries] (and also some validation).
```

wgpu-core/src/device/ray_tracing.rs
```diff
@@ -0,0 +1,376 @@
...
+        blas_desc: &resource::BlasDescriptor,
+        sizes: wgt::BlasGeometrySizeDescriptors,
+    ) -> Result<Arc<resource::Blas>, CreateBlasError> {
+        let size_info = match &sizes {
+            wgt::BlasGeometrySizeDescriptors::Triangles { desc } => {
+                let mut entries =
+                    Vec::<hal::AccelerationStructureTriangles<dyn hal::DynBuffer>>::with_capacity(
+                        desc.len(),
+                    );
+                for x in desc {
```
nical: Let's give `x` a descriptive name.

```suggestion
                for desc in descriptors {
```

wgpu-core/src/ray_tracing.rs
```diff
@@ -0,0 +1,336 @@
...
+};
+use std::sync::Arc;
+/// Ray tracing
+/// Major missing optimizations (no api surface changes needed):
+/// - use custom tracker to track build state
+/// - no forced rebuilt (build mode deduction)
+/// - lazy instance buffer allocation
+/// - maybe share scratch and instance staging buffer allocation
+/// - partial instance buffer uploads (api surface already designed with this in mind)
+/// - ([non performance] extract function in build (rust function extraction with guards is a pain))
```
nical: It's a bit odd how this comment for the file sits between two use statements and seems to apply to the `use std::{num::NonZeroU64, slice};` under it. Let's move it to the first line and also make it a regular `//` comment instead of a doc comment.

wgpu-types/src/lib.rs
```diff
@@ -7423,6 +7437,101 @@ impl Default for InstanceDescriptor {
...
+#[derive(Clone, Debug)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+/// Descriptor for all size definiing attributes of all geometries inside a bottom level acceleration structure.
+pub enum BlasGeometrySizeDescriptors {
+    /// Triangle geometry version.
+    Triangles {
+        /// Descriptor for each triangle geometry.
+        desc: Vec<BlasTriangleGeometrySizeDescriptor>,
+    },
+}
```
nical: This descriptor should take a slice instead of a vector.

Also, wgpu so far has avoided chopping off words in the public API. I think that it's a good thing, and the plural form is informative.

```suggestion
pub enum BlasGeometrySizeDescriptors<'a> {
    /// Triangle geometry version.
    Triangles {
        /// Descriptor for each triangle geometry.
        descriptors: &'[BlasTriangleGeometrySizeDescriptor],
    },
}
```

For consistency with the rest of the API, BlasGeometrySizeDescriptors should be passed by reference (for example in `create_blas`.

api_spec.rs
```diff
@@ -0,0 +1,154 @@
+// Ray tracing api proposal for wgpu (underlining Vulkan, Metal and DX12 implementations)
```
nical: This file should be removed from the PR before merging.

.gitattributes
```diff
@@ -1,3 +1,8 @@
+<<<<<<< HEAD
+*.mtl binary -diff
+*.obj binary -diff
+=======
 *.mtl binary
 *.obj binary
 wgpu/src/backend/webgpu/webgpu_sys/** linguist-generated=true
+>>>>>>> trunk
```
nical: Looks like this change is not intentional.

nical (commented):
I did a first pass, it looks like a solid contribution, thanks a lot! Since I'm still wrapping my head around the domain, what is coming out of this first review is the usual round of stylistic changes. There is probably going to be more of those in the next pass and hopefully soon a more thorough review of the more interesting/technical aspects of the work.

wgpu-types/src/lib.rs
Vecvec: This is required to have no lifetimes because it is used in the wgpu-core BLAS here
https://github.com/Vecvec/wgpu/blob/0fcc749efda0f22ccbcd5dffed3442d68fd17850/wgpu-core/src/resource.rs#L1924-L1925
doing this would require a duplicate size descriptor for wgpu-core that was a vector.

Edit: This sounds like I'm disagreeing, which I am not, I just want to know whether I should still do this if it had a reason.

wgpu-core/src/command/ray_tracing.rs
Vecvec: This is complicated, due to iterators taking things by value the normal BLAS and TLAS iterators currently need to be recreated, which cannot be done this way, if you prefer this way, this could go like queue submit with the last level taking a slice.

wgpu-types/src/lib.rs
nical: I suspected something like that. My inclination is that wgpu's `create` functions are usually cold code and that we can pay for an extra allocation and copy for the sake of consistency, after all the internal detail that motivate passing a vector may change in the future. But that's something we can easily revisit later so it's fine to move on and land a first implementation this way.

wgpu-core/src/command/ray_tracing.rs
nical: Fair enough.

wgpu-types/src/lib.rs
Vecvec: So (just checking) does that mean that this should or shouldn't have something done about it here?

nical: Sorry, that was ambiguous. Let's leave it the way it is for now.

Vecvec:
> and tlas packages (which are only in wgpu) being able to have a different max instances from the max instances a tlas was created with (though I can't remember if it gets checked when building), this one is hard to validate

this does get validated when building. (see BuildAccelerationStructureError::TlasInstanceCountExceeded)

I've thought that the TLAS instances max count could be made sure to be the same in wgpu though. You could hold the value passed into create TLAS in the TLAS so the TLAS package can read it when it gets created. This would make the API differ from the proposed API a bit though.

wgpu-core/src/device/ray_tracing.rs
Vecvec: I am confident I have found all of these.

wgpu-core/src/command/ray_tracing.rs
Vecvec: I think I have resolved all of these

Vecvec:
@JMS55 I've pushed a commits removes the need for max instance when creating tlas package, but this is a deviation from the spec that this is implementing from. Tell me what you think.

wgpu/src/ray_tracing.rs
```diff
@@ -205,8 +206,8 @@ static_assertions::assert_impl_all!(TlasPackage: WasmNotSendSync);
 impl TlasPackage {
     /// Construct TlasPackage consuming the Tlas (prevents modification of the Tlas without using this package).
     /// (max_instances needs to fit into tlas)
```
JMS55: Probably need to remove the max_instances comment.

JMS55 (changes requested):
I have no preference on TlasPackage taking max_instances or not. If it's safer/easier to validate to take it from the Tlas, then feel free to do that. I never used the ability to have a different amount of tlas items. We can always change it later.

https://github.com/JMS55/bevy/blob/solari3/crates/bevy_pbr/src/solari/scene/bind_group.rs#L101-L103

Vecvec:
It's mainly a user code thing, we still need the validation (new_with_instances doesn't do this because it can't raise an validation error in wgpu). Personally I just found it annoying to have to write it in.

Vecvec:
> I never used the ability to have a different amount of tlas items.

there is also still the ability to have a none variant of a tlas instance this means that you can still have more or less of the tlas instances just it will now automatically have a cap to the tlas's max instance count

Vecvec:
I have added more detail to some of the docs to make them more like a spec (since this is not in the WGSL spec).

examples/src/ray_cube_compute/mod.rs
```diff
@@ -0,0 +1,632 @@
...
+                    vertex_buffer: &vertex_buf,
+                    first_vertex: 0,
+                    vertex_stride: mem::size_of::<Vertex>() as u64,
+                    index_buffer: Some(&index_buf),
+                    index_buffer_offset: Some(0),
+                    transform_buffer: None,
+                    transform_buffer_offset: None,
+                }]),
+            }),
+            // iter::empty(),
```
cwfitzgerald: Stale comment

examples/src/ray_scene/shader.wgsl
```diff
@@ -0,0 +1,168 @@
...
+    first_vertex: u32,
+    first_geometry: u32,
+    last_geometry: u32,
+    _pad: u32
+};
+
+struct Material{
+    r: f32,
+    m: f32,
+    s: f32,
```
cwfitzgerald: We should use better name here

examples/src/ray_scene/shader.wgsl
```diff
@@ -0,0 +1,168 @@
...
+
+struct Uniforms {
+    view_inv: mat4x4<f32>,
+    proj_inv: mat4x4<f32>,
+};
+
+struct Vertex {
+    pos:     vec3<f32>,
+    normal:  vec3<f32>,
+    uv:      vec2<f32>,
```
cwfitzgerald: While this is pretty, we should remove these to keep things consistent

examples/src/ray_scene/shader.wgsl
```diff
@@ -0,0 +1,168 @@
...
+    instance_id: u32,
+    sbt_record_offset: u32,
+    geometry_index: u32,
+    primitive_index: u32,
+    barycentrics: vec2<f32>,
+    front_face: bool,
+    object_to_world: mat4x3<f32>,
+    world_to_object: mat4x3<f32>,
+}
+*/
```
cwfitzgerald: Are these blocks just reminding the user the contents of the ray query? If so, we should had a bit of explanatory verbiage

examples/src/ray_scene/shader.wgsl
```diff
@@ -0,0 +1,168 @@
...
+            color = vec4<f32>(normal, 1.0);
+        }
+    }
+
+    // let vertex = vertices[0];
+    // let index = indices[0];
+    // let geometry = geometries[0];
+    // let instance = instances[0];
+
+    return color; // vec4<f32>(vertex.tex_coords, 1.0, 1.0);
```
cwfitzgerald: Stale comments?

examples/src/ray_scene/screenshot.png
cwfitzgerald, line 1: I wonder if we can start everything's rotation such that the screenshot is upright (and when the example starts they're upright.

examples/src/ray_traced_triangle/mod.rs
```diff
@@ -0,0 +1,460 @@
...
+    tlas_package: TlasPackage,
+    compute_pipeline: wgpu::ComputePipeline,
+    blit_pipeline: wgpu::RenderPipeline,
+    bind_group: wgpu::BindGroup,
+    blit_bind_group: wgpu::BindGroup,
+    storage_texture: wgpu::Texture,
+    start: Instant,
+}
+
+#[allow(dead_code)]
```
cwfitzgerald: All of these examples have allow(dead_code) which surprises me. Is that actually necessary.

tests/tests/ray_tracing/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+                    vertex_buffer: &vertex_buf,
+                    first_vertex: 0,
+                    vertex_stride: mem::size_of::<Vertex>() as u64,
+                    index_buffer: Some(&index_buf),
+                    index_buffer_offset: Some(0),
+                    transform_buffer: None,
+                    transform_buffer_offset: None,
+                }]),
+            }),
+            // iter::empty(),
```
cwfitzgerald: Old comment

tests/tests/ray_tracing_use_after_free/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+                index_buffer: None,
+                index_buffer_offset: None,
+                transform_buffer: None,
+                transform_buffer_offset: None,
+            }]),
+        }),
+        iter::empty(),
+    );
+    ctx.queue.submit(Some(encoder.finish()));
+    drop(blas);
```
cwfitzgerald: This doesn't actually ensure that the drop has happened. We should also device.poll(Wait) to make sure it fully dies.

tests/tests/ray_tracing_use_after_free/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+    AccelerationStructureFlags, AccelerationStructureGeometryFlags, BlasGeometrySizeDescriptors,
+    BlasTriangleGeometrySizeDescriptor, BufferAddress, BufferUsages, CommandEncoderDescriptor,
+    CreateBlasDescriptor, CreateTlasDescriptor, VertexFormat,
+};
+
+fn required_features() -> wgpu::Features {
+    wgpu::Features::RAY_QUERY | wgpu::Features::RAY_TRACING_ACCELERATION_STRUCTURE
+}
+
+fn execute(ctx: TestingContext) {
```
cwfitzgerald: It would be great to get a bit of commentary within the function what exactly we're testing here and which points are important to look at. I personally like breaking up the test into sections delineated by comments, this helps make clear what is going on.

wgpu-core/src/device/ray_tracing.rs
```diff
@@ -0,0 +1,376 @@
...
+                .create_acceleration_structure(&hal::AccelerationStructureDescriptor {
+                    label: desc.label.as_deref(),
+                    size: size_info.acceleration_structure_size,
+                    format: hal::AccelerationStructureFormat::TopLevel,
+                })
+        }
+        .map_err(DeviceError::from_hal)?;
+
+        let instance_buffer_size = get_raw_tlas_instance_size(self.backend())
+            * std::cmp::max(desc.max_instances, 1) as usize;
```
cwfitzgerald: nit: I think calling `desc.max_instances.max(1)` would be a little more expected

wgpu-core/src/resource.rs
```diff
@@ -981,6 +983,44 @@ impl Drop for FlushedStagingBuffer {
...
+        })
+    }
+    pub(crate) fn raw(&self) -> &dyn hal::DynBuffer {
+        self.raw.as_ref()
+    }
+}
+
+impl Drop for ScratchBuffer {
+    fn drop(&mut self) {
+        resource_log!("Destroy raw StagingBuffer");
```
cwfitzgerald: todo: the log is wrong here, should be ScratchBuffer

wgpu-core/src/resource.rs
```diff
@@ -981,6 +983,44 @@ impl Drop for FlushedStagingBuffer {
...
+        // SAFETY: We are in the Drop impl and we don't use self.raw anymore after this point.
+        let raw = unsafe { ManuallyDrop::take(&mut self.raw) };
+        unsafe { self.device.raw().destroy_buffer(raw) };
+    }
+}
+
```
cwfitzgerald: Could we plop this in its own module? In generally trying to pare down module sizes, if possible.

wgpu-types/src/lib.rs
```diff
@@ -98,7 +98,7 @@ pub const QUERY_SIZE: u32 = 8;
 /// Backends supported by wgpu.
 #[repr(u8)]
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
```
cwfitzgerald: Quite a few unintentional changes here. I think they probably should be left as is due to the conditional compilation.

cwfitzgerald (changes requested):
This is an absolutely monster PR, and it looks very good! Pretty amazing work that everyone has done, and thank you for shepherding it in! We're going to try our best to keep up with it so we can actually land it.

I haven't fully reviewed everything, but I have done a lot of spot checking for a first round of stuff. Going to review the API itself and the documentation more thoroughly in the next pass.

I really appreciate the testing and the testing of the examples, it means I have a lot more confidence that the code is actually correct!

I do think we need a little more testing, particularly on the shader side.

We should add execution tests to ensure reasonable behavior for:
- Forgetting to call rayQueryInitialize, but still calling Execute and getCommittedIntersection
- An additional call to rayQueryInitialize _after_ Execute and Intersection have been done.
- We call getCommittedIntersection without any calls to init or execute.

wgpu-core/src/resource.rs
Vecvec: Alternatively we could put it in the pre-existing ray_tracing.rs, it feels a little small to be on it's own.

examples/src/ray_traced_triangle/mod.rs
Vecvec: I thought it would since it's only written to, but apparently not.

edit: maybe something in bytemuck does something?

wgpu-core/src/resource.rs
cwfitzgerald: Honestly, I'd rather too small than too large at this point - it's not really inherently related to rt, just a nice way to get a scratch buffer

cwfitzgerald:
(Note: we don't use coverage as some kind of barrier-to-entry, but it's a good tool for finding where things might be missing)

https://app.codecov.io/gh/gfx-rs/wgpu/pull/6291 suggests there are a few larger methods or branches that just aren't being exercised in the tests, probably worth trying those out.

examples/src/ray_scene/screenshot.png
Vecvec: I'm not sure how this would be done with timings, but I can do it by adding Pi.

examples/src/ray_scene/shader.wgsl
Vecvec: I don't actually know what these were for, but having at least one somewhere was nice for a reference. I could move them to a simpler example (e.g. wgpu ray-traced-triangle)

cwfitzgerald: That would be nice. More generally we probably need somewhere to store all this api information as we're off in spec-less land, but let's not worry about that for now.

examples/src/ray_scene/screenshot.png
cwfitzgerald: Sounds good

examples/src/ray_scene/shader.wgsl
Vecvec: So that means I don't remove this?

examples/src/ray_scene/shader.wgsl
Vecvec: I'm confused what this means, these are part of the `Vertex` struct and are all used in the shader, could you please tell me what to do here?

examples/src/ray_scene/shader.wgsl
Vecvec: I'll copy the names from mod.rs

examples/src/ray_scene/shader.wgsl
cwfitzgerald: Just remove the whitespace between the variable name and the type

examples/src/ray_scene/shader.wgsl
cwfitzgerald: I think leaving them in each file is fine, just would preface the comment with "The contents of the RayQuery struct are as follows" or something like that.

Vecvec:
> https://app.codecov.io/gh/gfx-rs/wgpu/pull/6291 suggests there are a few larger methods or branches that just aren't being exercised in the tests, probably worth trying those out.

Yes, it seems that `command_encoder_build_acceleration_structures_unsafe_tlas` is not being tested, I'll modify one example to do basic testing (but since it's unsafe, it shouldn't need to much testing because it's guarantees are from the user).

Vecvec:
> I do think we need a little more testing, particularly on the shader side.
> We should add execution tests to ensure reasonable behavior for:
> - Forgetting to call rayQueryInitialize, but still calling Execute and getCommittedIntersection
> -  An additional call to rayQueryInitialize after Execute and Intersection have been done.
> -   We call getCommittedIntersection without any calls to init or execute.

This seems unrelated to this PR, that is naga testing for a feature already in trunk, should that be in this PR? Though it is a feature that would be opened up by this PR, I feel that creating extra test cases for code already checked in in this PR, which is already massive, would be bad. If this is wanted I can probably create something though.

tests/tests/ray_tracing_use_after_free/mod.rs
Vecvec: I think I've added enough commentary for this.

cwfitzgerald:
I'll leave this up to you if you are interested in doing this. I don't want to overly burden you with testing that code too, we just can currently only do execution tests with the wgpu-level api, so those tests couldn't have been written in the previous PR.

I can do a follow up PR to this to add those tests (and fix any issues).

examples/src/ray_scene/shader.wgsl
Vecvec: Ok, I missed that they had lots of whitespace.

Vecvec:
> I don't want to overly burden you with testing that code too, we just can currently only do execution tests with the wgpu-level API, so those tests couldn't have been written in the previous PR.

I think that since I have been trying to keep down the size of this PR (It's almost 10,000 lines!), tests for naga code should be left to a follow-up PR.

cwfitzgerald:
Sounds good.

Vecvec:
I'm trying to keep the size of this PR down, so not having those test would be good, I have already a few things to do for follow up PRs (making scratch buffer offset dynamic based on device instead of maximum, compaction, get vertex, procedural geometry), so more is probably fine. I am also expecting this to likely be revised.

wgpu/src/ray_tracing.rs
```diff
@@ -0,0 +1,616 @@
...
+    pub vertex_buffer: &'a Buffer,
+    /// Offset into the vertex buffer as a factor of the vertex stride.
+    pub first_vertex: u32,
+    /// Vertex stride.
+    pub vertex_stride: wgt::BufferAddress,
+    /// Index buffer (optional).
+    pub index_buffer: Option<&'a Buffer>,
+    /// Index buffer offset in bytes (optional, required if index buffer is present).
+    pub index_buffer_offset: Option<wgt::BufferAddress>,
+    /// Transform buffer containing 3x4 (rows x columns, row mayor) affine transform matrices `[f32; 12]` (optional).
```
cwfitzgerald:
```suggestion
    /// Transform buffer containing 3x4 (rows x columns, row major) affine transform matrices `[f32; 12]` (optional).
```

wgpu/src/ray_tracing.rs
```diff
@@ -0,0 +1,616 @@
...
+    /// This can be done by ordering instances from the most to the least used.
+    pub fn get_mut_single(&mut self, index: usize) -> Option<&mut Option<TlasInstance>> {
+        if index >= self.instances.len() {
+            return None;
+        }
+        if index as u32 + 1 > self.lowest_unmodified {
+            self.lowest_unmodified = index as u32 + 1;
+        }
+        Some(&mut self.instances[index])
+    }
```
cwfitzgerald: These apis are a bit weird to me compared to taking the TlasInstances at the start - is the idea here that you can update this later, and then the "update" operation will be cheaper?

examples/src/ray_traced_triangle/mod.rs
```diff
@@ -0,0 +1,459 @@
...
+        wgpu::Limits::default()
+    }
+
+    fn init(
+        config: &wgpu::SurfaceConfiguration,
+        _adapter: &wgpu::Adapter,
+        device: &wgpu::Device,
+        queue: &wgpu::Queue,
+    ) -> Self {
+        let mut index_buffer = false;
```
cwfitzgerald: I think having both options makes the example less clear for the total newbie - I would probably force index buffer on, then note in the `BlasTriangleGeometrySizeDescriptor` that just like vertex buffers, you can go without them.

wgpu/src/ray_tracing.rs
```diff
@@ -0,0 +1,616 @@
...
+        &self.instances
+    }
+
+    /// Get a mutable slice to a range of instances.
+    /// Returns None if the range is out of bounds.
+    /// All elements from the lowest accessed index up are marked as modified.
+    // this recommendation is not useful yet, but is likely to be when ability to update arrives or possible optimisations for building get implemented.
+    /// For best performance it is recommended to prefer access to low elements and modify higher elements as little as possible.
+    /// This can be done by ordering instances from the most to the least used.
+    pub fn get_mut_slice(&mut self, range: Range<usize>) -> Option<&mut [Option<TlasInstance>]> {
```
cwfitzgerald: It might be worth adding impls for `IndexMut<usize>` and `IndexMut<Range<usize>>` to mirror standard containers, making some of the example code clearer.

wgpu/src/ray_tracing.rs
```diff
@@ -0,0 +1,616 @@
...
+    pub(crate) lowest_unmodified: u32,
+}
+
+/// Utility module to add traits for the device and command encoder.
+pub mod traits {
+    pub use super::{CommandEncoderRayTracing as _, DeviceRayTracing as _};
+}
+
+/// Trait to add ray tracing functions to a [`Device`].
+pub trait DeviceRayTracing {
```
cwfitzgerald: We don't need to make this a special trait, we can use an impl block with some documentation. See https://docs.rs/wgpu/latest/src/wgpu/lib.rs.html#4261-4284 for reference.

This way the user doesn't need to import a trait to use them.

examples/src/ray_traced_triangle/mod.rs
```diff
@@ -0,0 +1,459 @@
...
+    blit_bind_group: wgpu::BindGroup,
+    storage_texture: wgpu::Texture,
+    start: Instant,
+}
+
+#[repr(C)]
+#[derive(bytemuck::Pod, bytemuck::Zeroable, Clone, Copy, Debug)]
+struct Uniforms {
+    view_inverse: [[f32; 4]; 4],
+    proj_inverse: [[f32; 4]; 4],
```
cwfitzgerald: These can be glam::Mat4 - glam impls bytemuck.

cwfitzgerald:
When running on llvmpipe on windows, I get some validation errors (and a segfault, probably related) on all the example tests. 

```
Validation Error: Validation Error: [ VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03710 ] Object 0: handle = 0x27d00c65740, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x63ff8904 | vkCmdBuildAccelerationStructuresKHR(): pInfos[0].scratchData.deviceAddress (2735913731904) must be a multiple of minAccelerationStructureScratchOffsetAlignment (128). The Vulkan spec states: For each element of pInfos, its scratchData.deviceAddress member must be a multiple of VkPhysicalDeviceAccelerationStructurePropertiesKHR::minAccelerationStructureScratchOffsetAlignment (https://vulkan.lunarg.com/doc/view/1.3.290.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03710
```

Seems we're not minding minAccelerationStructureScratchOffsetAlignment?

wgpu/src/ray_tracing.rs
cwfitzgerald, line 1: I don't think we should expose this as a separate module, it should be part of the main api. `wgpu`'s move of keeping the surface api to all be flat has generally regarded as a good move, and we should continue that.

The implementation (in wgpu/src/api) should probably be split up into `blas.rs` `tlas.rs` then put the impl blocks in the corresponding `deivce.rs` and `command_encoder.rs`. This separates out the raytracing code a bit, but allows types to be grouped in together (all the Blas Types are together) which really improves find ability.

wgpu-core/src/command/ray_tracing.rs
```diff
@@ -0,0 +1,1223 @@
...
+
+        let hub = &self.hub;
+
+        let cmd_buf = hub
+            .command_buffers
+            .get(command_encoder_id.into_command_buffer_id());
+
+        let buffer_guard = hub.buffers.read();
+        let blas_guard = hub.blas_s.read();
+        let tlas_guard = hub.tlas_s.read();
```
cwfitzgerald: Oddly, we generally prefer not having long standing locks here - the hubs should have a `get` or similar, which will give you an Arc<T> directly, doing all the locking internally. This will also simplify the types you need to pass into various functions.

wgpu-core/src/command/ray_tracing.rs
```diff
@@ -0,0 +1,1223 @@
...
+        let mut tlas_storage = Vec::<(
+            Arc<Tlas>,
+            hal::AccelerationStructureEntries<dyn hal::DynBuffer>,
+            u64,
+        )>::new();
+        let mut tlas_buf_storage = Vec::<(
+            Arc<Buffer>,
+            Option<PendingTransition<BufferUses>>,
+            TlasBuildEntry,
+        )>::new();
```
cwfitzgerald: I think it is worth defining structs for these, so the members get names and we don't get these complicated types.

wgpu-core/src/command/ray_tracing.rs
```diff
@@ -0,0 +1,1223 @@
...
+                            action.blas.error_ident(),
+                        ));
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
+
+    // makes sure a tlas is build before it is used
```
cwfitzgerald:
```suggestion
    // makes sure a tlas is built before it is used
```

cwfitzgerald (changes requested):
Things are shaping up nicely. Have another round of comments, this time focusing on the API itself and some bits of the implementation. Let me know if you have any questions.

After this round, I think I'm going to do a pass and a few commits myself where I will fix some smaller formatting things as well as writing some more internal documentation, as a means of me better understanding the code and what's going on.

Once that is finished, I think we will be able to land this, with the intent to land some additional testing afterwards.

wgpu/src/ray_tracing.rs
Vecvec: This is a potential optimization, though is not currently implemented.

examples/src/ray_traced_triangle/mod.rs
Vecvec: If this is forced it should be off because it's the only example that doesn't get compression from the index buffer, and so all other examples have index buffers, therefore if these are forced on we lose some testing coverage.

Vecvec:
> Seems we're not minding minAccelerationStructureScratchOffsetAlignment?

That's odd, at the top of the file it's forced to 256 (the maximum of any device), so this should work fine...
from command/ray_tracing.rs
````rust
// TODO: Get this from the device (e.g. VkPhysicalDeviceAccelerationStructurePropertiesKHR.minAccelerationStructureScratchOffsetAlignment) this is currently the largest possible some devices have 0, 64, 128 (lower limits) so this could create excess allocation (Note: dx12 has 256).
const SCRATCH_BUFFER_ALIGNMENT: u32 = 256;
````

File: wgpu-core/src/command/ray_tracing.rs (hunk 30 above)
Comment (Vecvec) [reply wgpu-core/src/command/ray_tracing.rs]: I've used types for some others, should those also be structs?

File: examples/src/ray_traced_triangle/mod.rs (hunk 25 above)
Comment (cwfitzgerald) [reply examples/src/ray_traced_triangle/mod.rs]: My personal preference would be to have an index buffer in the example (to show the user how they work) and have a simple test which covers the non-index case.

File: wgpu-core/src/command/ray_tracing.rs (hunk 30 above)
Comment (cwfitzgerald) [reply wgpu-core/src/command/ray_tracing.rs]: If they're big like this, sure.

File: wgpu-core/src/command/ray_tracing.rs [hunk 32]
```diff
@@ -0,0 +1,1223 @@
...
+    Arc<Blas>,
+    hal::AccelerationStructureEntries<'a, dyn hal::DynBuffer>,
+    u64,
+)>;
+
+// TODO: Get this from the device (e.g. VkPhysicalDeviceAccelerationStructurePropertiesKHR.minAccelerationStructureScratchOffsetAlignment) this is currently the largest possible some devices have 0, 64, 128 (lower limits) so this could create excess allocation (Note: dx12 has 256).
+const SCRATCH_BUFFER_ALIGNMENT: u32 = 256;
+
+impl Global {
+    pub fn command_encoder_build_acceleration_structures_unsafe_tlas<'a>(
```
cwfitzgerald, line 74: These functions are _very_ similar, is there any way we can define them in terms of each other?

Vecvec: I could try to compress some things into other functions e.g. checking device and updating it's index, but in the middle there is a large block of different code (where the tlas code is has about 100 lines different).

wgpu-core/src/command/ray_tracing.rs
Vecvec: They were bigger, clippy was complaining

wgpu-core/src/command/ray_tracing.rs
Vecvec: I didn't realise those existed.

wgpu-core/src/command/ray_tracing.rs
Vecvec: Defining one in terms of the other could be a lot of work and I'm not entirely sure of the use-case of `command_encoder_build_acceleration_structures_unsafe_tlas` vs adding `as_hal` methods. I think that if we *need* to define one in terms of the other it might be more worthwhile to just add `as_hal` methods instead, but perhaps @JMS55 could comment on the use-case for this function?

edit: forgot the method was `as_hal` not `into_hal`

examples/src/ray_traced_triangle/mod.rs
Vecvec: Implementing this gives a compile error saying that glam::Mat4 doesn't support Pod & Zeroable

wgpu/src/ray_tracing.rs
Vecvec: I think this is resolved.

examples/src/ray_traced_triangle/mod.rs
cwfitzgerald: We just need to add the bytemuck feature to https://github.com/gfx-rs/wgpu/blob/trunk/examples/Cargo.toml#L37. This will enable their impl of Pod & Zeroable.

wgpu-core/src/command/ray_tracing.rs
cwfitzgerald: Alright, could you just add a note as to why they're currently separate, and we can re-evaluate later.

examples/src/ray_traced_triangle/mod.rs
```diff
@@ -0,0 +1,442 @@
...
+                y: -1.0,
+                z: -2.0,
+            })
+            .transpose()
+            .to_cols_array()[..12]
+                .try_into()
+                .unwrap(),
+            0,
+            0xff,
+        ));
```
cwfitzgerald: typo: all of these have index 0, which is causing the test to fail.

tests/tests/ray_tracing/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+
+fn required_features() -> wgpu::Features {
+    wgpu::Features::TEXTURE_BINDING_ARRAY
+        | wgpu::Features::STORAGE_RESOURCE_BINDING_ARRAY
+        | wgpu::Features::VERTEX_WRITABLE_STORAGE
+        | wgpu::Features::RAY_QUERY
+        | wgpu::Features::RAY_TRACING_ACCELERATION_STRUCTURE
+}
+
+fn execute<const USE_INDEX_BUFFER: bool>(ctx: TestingContext) {
```
cwfitzgerald: We can just use a normal argument, we don't need a generic here

tests/tests/ray_tracing/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+#[gpu_test]
+static RAY_TRACING: GpuTestConfiguration = GpuTestConfiguration::new()
+    .parameters(
+        TestParameters::default()
+            .test_features_limits()
+            .features(required_features()),
+    )
+    .run_sync(execute::<false>);
```
cwfitzgerald:
```suggestion
#[gpu_test]
static RAY_TRACING: GpuTestConfiguration = GpuTestConfiguration::new()
    .parameters(
        TestParameters::default()
            .test_features_limits()
            .features(required_features()),
    )
    .run_sync(|ctx| {
        for value in [false, true] {
            execute(ctx, value);
        }
    });
```

Or something like that

Vecvec: Ok, I'd thought that the other way could have been better

examples/src/ray_traced_triangle/mod.rs
Vecvec: ah, thats probably my fault just when changing to index_mut

Vecvec:
@cwfitzgerald I've been thinking about the issue you are having with llvmpipe, it seems to be that the scratch buffer's device address is not properly aligned, that sounds to me like it's a wgpu-hal issue not a wgpu-core issue. To help check whether it is as I think, could you try the hal ray-traced-triangle? 

edit: comment is
> When running on llvmpipe on windows, I get some validation errors (and a segfault, probably related) on all the example tests.
> Validation Error: Validation Error: [ VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03710 ] Object 0: handle = 0x27d00c65740, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x63ff8904 | vkCmdBuildAccelerationStructuresKHR(): pInfos[0].scratchData.deviceAddress (2735913731904) must be a multiple of minAccelerationStructureScratchOffsetAlignment (128). The Vulkan spec states: For each element of pInfos, its scratchData.deviceAddress member must be a multiple of VkPhysicalDeviceAccelerationStructurePropertiesKHR::minAccelerationStructureScratchOffsetAlignment (https://vulkan.lunarg.com/doc/view/1.3.290.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03710
Seems we're not minding minAccelerationStructureScratchOffsetAlignment?

cwfitzgerald:
Good idea - I can also debug this a bit to see what's going on.

Vecvec:
Hmm... this line (https://github.com/gfx-rs/wgpu/blob/859dd8817e7484b51823d443d7cac93c6e9a7ef2/wgpu-hal/src/vulkan/device.rs#L918-L925) has no mention of scratch buffer usage, could you also try hard-coding it to
````rust
let alignment_mask = if desc.usage.intersects(
            crate::BufferUses::TOP_LEVEL_ACCELERATION_STRUCTURE_INPUT
                | crate::BufferUses::BOTTOM_LEVEL_ACCELERATION_STRUCTURE_INPUT,
        ) {
            16
        } else if desc.usage.contains(crate::BufferUses::ACCELERATION_STRUCTURE_SCRATCH) {
            128
        } else {
            req.alignment
        } - 1;
````

File: wgpu-core/src/command/ray_tracing.rs (hunk 32 above)
Comment (JMS55) [reply wgpu-core/src/command/ray_tracing.rs:74]: No idea either, I didn't use the unsafe functions ever.

However the general problem wgpu's as_hal() functions have is that you have no way of hooking into wgpu's synchronization code, so there's no real way to use the hal objects anyways.

Comment (Vecvec) [reply wgpu-core/src/command/ray_tracing.rs:74]: I don't think there is much synchronization code in build_acceleration_structures, but there is no `as_hal` method currently so this should be reconsidered in a later PR. I will add a todo talking about reconsidering this once `as_hal` for blases and tlases is in.

Comment (JMS55) [reply wgpu-core/src/command/ray_tracing.rs:74]: It's more like if you build the acceleration structures, you don't want the build commands to be reordered after the draw commands that use them.

Can be worked-around with multiple command encoders though probably, and you really want to build BLAS's on an async compute queue anyways, although that's tricky with wgpu vulkan due to vulkan not letting you request additional queues past device initialization.

**Comment by cwfitzgerald** [comment]
I'm fairly certain that this is actually a lavapipe bug. Even if we properly manage the alignment, the resulting buffer device address pointer is not actually aligned as much as the memory is.

This seems to be technically allowed for the spec, but it's probably just a spec oversight. I'm going to put together a bug report tomorrow and file it against Mesa to see what they say.

**Comment by Vecvec** [comment]
> I'm fairly certain that this is actually a lavapipe bug

that would explain why I couldn't repro this even though I have the same minAccelerationStructureScratchOffsetAlignment as mesa.

**Comment by cwfitzgerald** [comment]
Filed https://gitlab.freedesktop.org/mesa/mesa/-/issues/11940

File: examples/src/ray_traced_triangle/mod.rs (hunk 28 above)
Comment (cwfitzgerald) [reply examples/src/ray_traced_triangle/mod.rs]: This one still needs attention

Comment (Vecvec) [reply examples/src/ray_traced_triangle/mod.rs]: this should be fixed

**Comment by cwfitzgerald** [comment]
Looking good - I'm going to do a documentation and formatting pass, then check in with the other maintainers tomorrow, but I'm expecting to get this in by the end of the week!

**Comment by cwfitzgerald** [comment]
Unfortunately I don't seem to be able to push to your branch, I keep getting 403's even though you have the "Maintainers are allowed to edit this pull request" selected. I suspect it might be because it's a fork-of-a-fork, but not sure, github isn't giving me more information.

I've raised a PR against your fork as a next-best. It includes a resolve of the merge conflicts

**Comment by cwfitzgerald** [comment]
Unfortunately I have also managed to break the validation scheme as well as raise some vulkan validation errors. I have ideas on how to fix this, and will continue hacking on them tomorrow.

Issue 1: These show up when you don't turn on the RayQuery feature. We are doing barriers wrong in some way in the vulkan backend.

```
Validation Error: [ VUID-vkCmdPipelineBarrier-srcAccessMask-06257 ] Object 0: handle = 0x2610bd14440, name = TLAS 1, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x235b7e76 | vkCmdPipelineBarrier(): pMemoryBarriers[0].srcStageMask contains pipeline stages VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT|VK_PIPELINE_STAGE_VERTEX_SHADER_BIT|VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT|VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT. The Vulkan spec states: If the rayQuery feature is not enabled and a memory barrier srcAccessMask includes VK_ACCESS_ACCELERATION_STRUCTURE_READ_BIT_KHR, srcStageMask must not include any of the VK_PIPELINE_STAGE_*_SHADER_BIT stages except VK_PIPELINE_STAGE_RAY_TRACING_SHADER_BIT_KHR (https://vulkan.lunarg.com/doc/view/1.3.290.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdPipelineBarrier-srcAccessMask-06257)
[2024-09-25T03:21:20Z ERROR wgpu_hal::vulkan::instance]         objects: (type: COMMAND_BUFFER, hndl: 0x2610bd14440, name: TLAS 1)
```

Issue 2: The following test passes without raising a validation error:

```rust
    //
    // Encode the BLAS build before the TLAS build, but submit them in the wrong order.
    //

    let mut encoder_blas = ctx
        .device
        .create_command_encoder(&CommandEncoderDescriptor {
            label: Some("BLAS 2"),
        });

    encoder_blas.build_acceleration_structures([&as_ctx.blas_build_entry()], []);

    let mut encoder_tlas = ctx
        .device
        .create_command_encoder(&CommandEncoderDescriptor {
            label: Some("TLAS 2"),
        });

    encoder_tlas.build_acceleration_structures([], [&as_ctx.tlas_package]);

    fail(
        &ctx.device,
        || {
            ctx.queue
                .submit([encoder_tlas.finish(), encoder_blas.finish()]);
        },
        None,
    );
```

cwfitzgerald:
Ayy, thanks for the invite ;) Pushed directly

Vecvec:
I think the second one is actually valid, ordering:
1. build blas: built index 1
2. build tlas: checks blas, it has been build w/ index 1
3. build tlas: checks blas, it has been build w/ index 1
4. build blas: this will invalidate the tlas, so any new usage will fail (ray query)

edit: If you create a new ASContext before step 3 this will work.
edit # 2: I'll fix this

cwfitzgerald:
**mumble mumble** mutable damn state

cwfitzgerald:
Okay nice, it does seem to work. 

I am also getting some intermittent 

```
VALIDATION [VUID-vkDestroyAccelerationStructureKHR-accelerationStructure-02442 (0xaf03fd73)]
        Validation Error: [ VUID-vkDestroyAccelerationStructureKHR-accelerationStructure-02442 ] | MessageID = 0xaf03fd73 | vkDestroyAccelerationStructureKHR():  can't be called on VkAccelerationStructureKHR 0xcad092000000000d[TLAS] that is currently in use by VkCommandBuffer 0x1e1b8536c60[TLAS 1]. The Vulkan spec states: All submitted commands that refer to accelerationStructure must have completed execution (https://vulkan.lunarg.com/doc/view/1.3.290.0/windows/1.3-extensions/vkspec.html#VUID-vkDestroyAccelerationStructureKHR-accelerationStructure-02442)
[2024-09-25T04:21:55Z ERROR wgpu_hal::vulkan::instance] VALIDATION [VUID-vkDestroyBuffer-buffer-00922 (0xe4549c11)]
```

Which means the tlas drop isn't properly deferring until the command encoder is done with it.

Vecvec:
I suspect the second error is that if we are breaking with an error it fails since it has drained the actions.

Vecvec:
Looking into queue submit, it seems that earlier errors (https://github.com/gfx-rs/wgpu/blob/8e787eb70a98c260f44b129128037d9a28d2c9ae/wgpu-core/src/device/queue.rs#L1098-L1111) destroy the command buffer and continue, while later ones (https://github.com/gfx-rs/wgpu/blob/8e787eb70a98c260f44b129128037d9a28d2c9ae/wgpu-core/src/device/queue.rs#L1137-L1145) break out of the error context instead (this is because these should only return hal errors). I think that the validate functions are now out of place and should be in `validate_command_buffer`.

Edit: never mind, this didn't help the validation errors

Vecvec:
also for the first validation error we place some acceleration structure barriers that go between build and shader input
(line 336, 719, 1238) these should only be put in if also ray_query (or later ray_tracing_pipeline) appear.

I think this needs something in hal that conditionally adds that based on feature support (or we expand shader input to vertex, fragment, compute, and later ray-tracing-pipeline inputs).

cwfitzgerald:
Chatted with the other maintainers, and two main things to do before we can land, then we're good to land with follow up testing for any naga stuff.

Both of these things I can take on:

- More testing; some validation testing would be nice to make sure common error patterns are easily caught.
- The validation should be refactored away from storing mutable state on the TLAS/BLAS itself and move to a tracker-like way of validating state. We've learned from experience with the texture initialization code that having state inside the resources itself is an area that is ripe for bugs and we want to move towards validation where information all "flows" from command encoder into the queue. This is a lot easier to get right and to keep right. Even if the validation as it stands today is correct, we're very liable to break it in the future, by accident. I'm not sure if you wrote the valdiation for this or someone else did, but I can do this conversion, just wanted you to know what my plan was.

Vecvec:
I'm fine to work on some of the testing, I assume common errors are thing like:
- Invalid vertex format
- Mismatched index buffers
- ect.

Vecvec:
After using advanced debugging (putting an `eprintln!` at the end of the test) :) I've discovered that the second set of vulkan validation errors are happening at the end of the test. Shouldn't all encoders have been reset by the wgpu validation errors generated in queue submit?

Vecvec:
It seems like command buffers might be leaking if a validation error is generated
````cmd
[2024-09-26T08:58:50Z INFO  wgpu_core::device::global] CommandEncoder::drop Id(1,1)
[2024-09-26T08:58:50Z INFO  wgpu_core::command] Drop CommandBuffer with 'TLAS 1' label <--- command buffer freed here
[2024-09-26T08:58:50Z INFO  wgpu_core::device::global] Device::create_shader_module -> Id(0,1)
[2024-09-26T08:58:50Z INFO  wgpu_core::device::global] Device::create_compute_pipeline -> Id(0,1)
[2024-09-26T08:58:50Z INFO  wgpu_core::device::global] Device::create_bind_group -> Id(0,1)
[2024-09-26T08:58:50Z INFO  wgpu_core::device::global] BindGroupLayout::drop Id(0,1)
[2024-09-26T08:58:50Z INFO  wgpu_core::device::global] Device::create_command_encoder -> Id(1,2)
[2024-09-26T08:58:50Z INFO  wgpu_core::device::queue] Queue::submit Id(0,1)
[2024-09-26T08:58:50Z INFO  wgpu_core::device::global] CommandBuffer::drop Id(1,2)
[2024-09-26T08:58:50Z INFO  wgpu_core::device::global] CommandEncoder::drop Id(1,2)
[2024-09-26T08:58:50Z INFO  wgpu_core::command] Drop CommandBuffer with '' label
[2024-09-26T08:58:50Z INFO  wgpu_core::device::global] BindGroup::drop Id(0,1)
[2024-09-26T08:58:50Z INFO  wgpu_core::binding_model] Destroy raw BindGroup with '' label
[2024-09-26T08:58:50Z INFO  wgpu_core::device::global] ComputePipeline::drop Id(0,1)
[2024-09-26T08:58:50Z INFO  wgpu_core::pipeline] Destroy raw ComputePipeline with '' label
[2024-09-26T08:58:50Z INFO  wgpu_core::binding_model] Destroy raw PipelineLayout with '' label
[2024-09-26T08:58:50Z INFO  wgpu_core::binding_model] Destroy raw BindGroupLayout with '' label
[2024-09-26T08:58:50Z INFO  wgpu_core::device::global] ShaderModule::drop Id(0,1)
[2024-09-26T08:58:50Z INFO  wgpu_core::pipeline] Destroy raw ShaderModule with 'shader.wgsl' label
[2024-09-26T08:58:50Z INFO  wgpu_core::device::global] CommandEncoder::drop Id(0,1)
[2024-09-26T08:58:50Z INFO  wgpu_core::command] Drop CommandBuffer with 'BLAS 1' label
[2024-09-26T08:58:50Z INFO  wgpu_core::device::global] Buffer::drop Id(0,1)
[2024-09-26T08:58:50Z INFO  wgpu_core::resource] Destroy raw Tlas with 'TLAS' label
[2024-09-26T08:58:50Z ERROR wgpu_hal::vulkan::instance] VALIDATION [VUID-vkDestroyAccelerationStructureKHR-accelerationStructure-02442 (0xaf03fd73)]
    	Validation Error: [ VUID-vkDestroyAccelerationStructureKHR-accelerationStructure-02442 ] | MessageID = 0xaf03fd73 | vkDestroyAccelerationStructureKHR():  can't be called on VkAccelerationStructureKHR 0xcad092000000000d[TLAS] that is currently in use by VkCommandBuffer 0x7f8b0de60af0[TLAS 1]. The Vulkan spec states: All submitted commands that refer to accelerationStructure must have completed execution (https://vulkan.lunarg.com/doc/view/1.3.290.0/linux/1.3-extensions/vkspec.html#VUID-vkDestroyAccelerationStructureKHR-accelerationStructure-02442) 
               ^--- vulkan validation error about command buffer being still active
````
~~I'll see if I can create a memory leak repro (without acceleration structures).~~ Never mind, remembered command encoders get reused.

Edit: wondering if #6323 might be related

Vecvec:
I've put a workaround in the tests that has stopped the vulkan validation errors as I believe that the errors from the vulkan validation layers are the same as those in #6323.

edit: seems buffers used pending writes to do this so actually never mind

Vecvec:
> The validation should be refactored away from storing mutable state on the TLAS/BLAS itself and move to a tracker-like way of validating state.

@cwfitzgerald what would this be like (would there be a `FastHashMap<TrackerIndex, Option<NonZeroU64>>`?)

Vecvec:
I have noticed that quite a few resources use snatchables and it seems that they are used in resource destruction if the resource is in something such as a bind group, I will implement snatch guards for BLASes and TLASes while @cwfitzgerald works on moving the validation away from a mutable state on the BLAS/TLAS.

Vecvec:
@cwfitzgerald you were working on moving the validation away from a mutable state on the BLAS/TLAS, do you know how long that might take? It seems that other people are working off this branch (for example: https://github.com/gfx-rs/wgpu/issues/1040#issuecomment-2463690213), so it would be useful to get this branch into trunk.

JMS55:
I coincidentally had just asked @cwfitzgerald about this on matrix. Posted their response below, with the first part paraphrased.

> [explaining that they don't have time to work on this anymore]. However we don't actually want to block it at this stage so our resolution to this problem is to let it in basically as is. The one change that we want to make is prefix the feature flag with EXPERIMENTAL_ so that people will expect it to not work sometimes or break or break your computer [or] whatever.

So sounds like we can merge once you rename the feature flag :)

cwfitzgerald (approved):
Good after failing CI fixed

JMS55:
Thanks so much @Vecvec! What a great birthday present ❤️

teoxoy:
I noticed that this PR makes use of `pending_writes` but the operations adding things in pending writes are on the command encoder. Pending writes is only needed for queue operations that require a hal command encoder to be executed (ex `write_buffer`). I think the way the new `command_encoder_build_acceleration_structures` use pending writes is not sound or at least not what it was designed for. Shouldn't we add things to the command encoder itself rather than into pending writes?

teoxoy:
`Tlas.destroy` don't seem to check if the `Tlas` is used in a bind group of an active submission. The only reason we need the `destroy` methods for textures and buffers is because they allow users to eagerly release memory in browser implementations. I think we can remove the destroy methods on the acceleration structures for now as they complicate the picture without any gain. If they will be needed for Firefox we can add them back.

teoxoy:
I put up a PR with these changes ^. Could someone test it?

PR #6303, opened 2024-09-21
fix: only log `Device::maintain` waits if they happen

ErichDonGubler:
**Connections**

Follow-up to @Kek5chen's question at https://github.com/gfx-rs/wgpu/pull/6044#issuecomment-2335175795, which seems pretty reasonable to me.

**Description**

Only emits a `Device::maintain` wait message if we actually make a call to wait for a submission to finish.

**Testing**

This still passes CI, so I think we're de-risked there.

**Checklist**

- [x] Run `cargo fmt`.
- [x] ~~Add change to `CHANGELOG.md`. See simple instructions inside file.~~ Not necessary, IMO.

### DIFF
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -457,13 +457,13 @@ impl Device {
 
         // If necessary, wait for that submission to complete.
         if maintain.is_wait() {
+            log::trace!("Device::maintain: waiting for submission index {submission_index}");
             unsafe {
                 self.raw()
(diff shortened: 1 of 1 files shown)


Kek5chen (approved):
Yea, that's exactly what I meant. Seems correct to me. If the wait call stalls for whatever reason, at least you have trace output. Otherwise you won't.

Wumpf (approved):
makes sense to me!

PR #6305, opened 2024-09-22
style: use `concat!(…)` in `DOWNLEVEL_*_MESSAGE`

ErichDonGubler:
This plays _much_ better with `rustfmt`, and we should likely be using this in many more places.


cwfitzgerald (approved):
ooo, nice

PR #6315, opened 2024-09-23
Only Initialize a Single Backend in Tests

cwfitzgerald:
Changes tests from matching on a strict adapter index to matching on the full adpater info.

This lets us only initialize the backends that are needed to find that particular adapter, providing massive speed gains, and allows `WGPU_BACKEND` to work again. 

This brings the test suite from 130s to 100s on my machine.

If I limit to only the vulkan backend, it brings the vulkan tests from 55s -> 35s. 

Pretty snazzy.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/tests/src/native.rs
+++ b/tests/src/native.rs
@@ -19,15 +19,16 @@ struct NativeTest {
 }
 
 impl NativeTest {
+    /// Adapter index is only used for naming the test, the adapters are matched based on the adapter info.
     fn from_configuration(
         config: GpuTestConfiguration,
-        adapter: &AdapterReport,
+        adapter_report: AdapterReport,
         adapter_index: usize,
     ) -> Self {
-        let backend = adapter.info.backend;
-        let device_name = &adapter.info.name;
+        let backend = adapter_report.info.backend;
+        let device_name = &adapter_report.info.name;
 
-        let test_info = TestInfo::from_configuration(&config, adapter);
+        let test_info = TestInfo::from_configuration(&config, &adapter_report);
 
         let full_name = format!(
             "[{running_msg}] [{backend:?}/{device_name}/{adapter_index}] {base_name}",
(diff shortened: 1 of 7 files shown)


tests/src/native.rs
```diff
@@ -50,10 +51,12 @@ impl NativeTest {
 
                 let env_value = if metal_validation { "1" } else { "0" };
                 std::env::set_var("MTL_DEBUG_LAYER", env_value);
-                // Metal Shader Validation is entirely broken in the paravirtualized CI environment.
-                // std::env::set_var("MTL_SHADER_VALIDATION", env_value);
+                if std::env::var("GITHUB_ACTIONS").as_deref() != Ok("true") {
+                    // Metal Shader Validation is entirely broken in the paravirtualized CI environment.
+                    std::env::set_var("MTL_SHADER_VALIDATION", env_value);
+                }
```
Wumpf, line 57: neat

tests/tests/create_surface_error.rs
```diff
@@ -6,7 +6,7 @@
 #[wasm_bindgen_test::wasm_bindgen_test]
 fn canvas_get_context_returned_null() {
     // Not using the normal testing infrastructure because that goes straight to creating the canvas for us.
-    let instance = wgpu_test::initialize_instance(false);
+    let instance = wgpu_test::initialize_instance(wgpu::Backends::GL, false);
```
Wumpf: comment above specifically says this test is for WebGL _and_ WebGPU, so why only GL here?

cwfitzgerald: I guess should be both then

cwfitzgerald:
```suggestion
    let instance = wgpu_test::initialize_instance(wgpu::Backends::all(), false);
```

Wumpf (approved):

PR #6316, opened 2024-09-23
style: simplify string formatting for readability

hamirmahal:
**Description**
The goal of this pull request is to improve code readability and maintainability.

**Testing**
I ran `cargo xtask test` locally.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/player/tests/test.rs
+++ b/player/tests/test.rs
@@ -58,7 +58,7 @@ struct Test<'a> {
 
 fn map_callback(status: Result<(), wgc::resource::BufferAccessError>) {
     if let Err(e) = status {
-        panic!("Buffer map error: {}", e);
+        panic!("Buffer map error: {e}");
     }
 }
 
@@ -88,7 +88,7 @@ impl Test<'_> {
             .iter()
             .map(|feature| {
                 wgt::Features::from_name(feature)
-                    .unwrap_or_else(|| panic!("Invalid feature flag {}", feature))
+                    .unwrap_or_else(|| panic!("Invalid feature flag {feature}"))
             })
             .fold(wgt::Features::empty(), |a, b| a | b);
         Test {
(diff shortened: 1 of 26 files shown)


cwfitzgerald (approved):
Nice, surprised clippy doesn't have a thing for that, actually

a1phyr:
[It does](https://rust-lang.github.io/rust-clippy/master/index.html#/uninlined_format_args), but was moved from warn-by-default to allow-by-default because rust-analyzer didn't support them well at the time, which made them painful to have in a codebase.

ErichDonGubler:
I think we'd be open to follow-up work that set the lint to warn-by-default! 😀

hamirmahal:
Sure, do you want that follow-up work in this PR, or a separate one?

cwfitzgerald:
Let me merge this, then lets do a follow-up.

player/tests/test.rs
```diff
@@ -186,8 +186,7 @@ impl Test<'_> {
 
             if &expected_data[..] != contents {
                 panic!(
-                    "Test expectation is not met!\nBuffer content was:\n{:?}\nbut expected:\n{:?}",
-                    contents, expected_data
+                    "Test expectation is not met!\nBuffer content was:\n{contents:?}\nbut expected:\n{expected_data:?}"
```
ErichDonGubler, line 189: context: `rustfmt` refuses to format runs of expressions with lines that exceed the default 100-column limit which it can't break up. We often run into this when we're writing shaders in string literals, or have a long diagnostic message we want to present.

issue(non-blocking): By lengthening this formatting spec. string literal to go past that limit here, we're making `rustfmt` run into this case. 

Since this is already merged, I'm gonna open a new PR and fix this and other cases up: <https://github.com/gfx-rs/wgpu/pull/6349>

I wish this weren't such a hard-to-detect problem; the only reason I recognize it is because I've done this dozens of times to myself.

PR #6318, opened 2024-09-23
fix: handle `Queue::submit` non-fatally

### LINKED ISSUE #6317: `Queue::submit` should not panic on validation errors
**Description**

When an invalid `CommandBuffer` is fed into `Queue::submit` on `trunk` (but, notably, not on 22.1.0), it panics on errors after they are forwarded to our fatal error machinery. This is wrong; we should be handling errors non-fatally, i.e., via the error sink set up by `Device::on_uncaptured_error` or the default panic handler.

**Repro steps**

A Cargo project with the following contents reproduces it for me:

* `Cargo.toml`:

	```
	[package]
	name = "bruh-idk"
	version = "0.1.0"
	edition = "2021"
	
	[dependencies]
	pollster = "0.3.0"
	wgpu = { git = "https://github.com/gfx-rs/wgpu", rev = "859dd8817e7484b51823d443d7cac93c6e9a7ef2" }
	```

* `src/main.rs`

	```rust
	use pollster::FutureExt as _;

	fn main() {
	    let instance = wgpu::Instance::new(Default::default());
	
	    let adapter = instance
	        .request_adapter(&Default::default())
	        .block_on()
	        .unwrap();
	
	    let (device, queue) = adapter
	        .request_device(&Default::default(), None)
	        .block_on()
	        .unwrap();
	
	    let shader_source = "\
	@group(0) @binding(0)
	var<storage, read_write> stuff: u32;
	
	@compute @workgroup_size(1) fn main() {
	    stuff = 2u;
	}
	";
	
	    let module = device.create_shader_module(wgpu::ShaderModuleDescriptor {
	        label: None,
	        source: wgpu::ShaderSource::Wgsl(shader_source.into()),
	    });
	
	    let compute_pipeline = device.create_compute_pipeline(&wgpu::ComputePipelineDescriptor {
	        label: None,
	        layout: None,
	        module: &module,
	        entry_point: None,
	        compilation_options: Default::default(),
	        cache: Default::default(),
	    });
	
	    device.push_error_scope(wgpu::ErrorFilter::Validation);
	    {
	        let mut command_encoder = device.create_command_encoder(&Default::default());
	        {
	            let mut render_pass = command_encoder.begin_compute_pass(&Default::default());
	            render_pass.set_pipeline(&compute_pipeline);
	
	            // NOTE: We deliberately don't set a bind group here, to provoke a validation error.
	
	            render_pass.dispatch_workgroups(1, 1, 1);
	        }
	
	        let _idx = queue.submit([command_encoder.finish()]);
	    }
	}
	```

The error scope in the `device.push_error_scope` from repro steps should absorb the validation errors emitted by the `RenderPass::end` and all subsequent operations in which the invalidation propagates, including `Queue::submit`. `Queue::submit`, however, panics instead. Observe that the above example produces the following output:

```
Error in Queue::submit: Validation Error

Caused by:
  Command encoder is invalid

```

**Extra materials**

Discovered in <https://github.com/gfx-rs/wgpu/pull/5714#discussion_r1769372866> as a blocker for that PR.

**Platform**

\-

ErichDonGubler:
**Connections**

Fixes #6317, unblocks #5714.

**Description**

* Change the signature of `wgpu_core::Global::queue_submit` to return a `(SubmissionIndex, …)` in addition to its current error type.
* Change the control flow of errors in `Queue::submit` to break to the end of a block. This is similar to what we already do in many APIs in `wgpu_core`.
* Hoist the scope of the local `submit_index` binding so it can be used at the point where we need to convert current error paths to also return the submission index.

Later, we will likely want to avoid actually retrieving a new submission index so we can minimize the critical section of code. We'll need to figure out a strategy for returning a valid (but not necessarily unique) index in the case of failures that prevent successful submission.

**It is strongly recommended that a reviewer [review this diff with whitespace changes ignored](https://github.com/gfx-rs/wgpu/pull/6318/files?w=1).**

**Testing**

Exercised by #6317. If a reviewer feels strongly, we can take the example from #5714 and turn it into a test.

**Checklist**

- [x] Run `cargo fmt`.
- [x] ~~Add change to `CHANGELOG.md`. See simple instructions inside file.~~ I don't think that this is necessary, since we haven't shipped this defect to Crates.io.

### DIFF
--- a/wgpu-core/src/device/queue.rs
+++ b/wgpu-core/src/device/queue.rs
@@ -1027,11 +1027,13 @@ impl Global {
         &self,
         queue_id: QueueId,
         command_buffer_ids: &[id::CommandBufferId],
-    ) -> Result<SubmissionIndex, QueueSubmitError> {
+    ) -> Result<SubmissionIndex, (SubmissionIndex, QueueSubmitError)> {
         profiling::scope!("Queue::submit");
         api_log!("Queue::submit {queue_id:?}");
 
-        let (submit_index, callbacks) = {
+        let submit_index;
+
+        let res = 'error: {
             let hub = &self.hub;
 
             let queue = hub.queues.get(queue_id);
@@ -1042,7 +1044,7 @@ impl Global {
 
             // Fence lock must be acquired after the snatch lock everywhere to avoid deadlocks.
             let mut fence = device.fence.write();
-            let submit_index = device
+            submit_index = device
                 .active_submission_index
                 .fetch_add(1, Ordering::SeqCst)
                 + 1;
@@ -1119,18 +1121,29 @@ impl Global {
                         }
 
                         // execute resource transitions
-                        unsafe {
+                        if let Err(e) = unsafe {
                             baked.encoder.begin_encoding(hal_label(
                                 Some("(wgpu internal) Transit"),
                                 device.instance_flags,
                             ))
                         }
-                        .map_err(|e| device.handle_hal_error(e))?;
+                        .map_err(|e| device.handle_hal_error(e))
+                        {
+                            break 'error Err(e.into());
+                        }
 
                         //Note: locking the trackers has to be done after the storages
                         let mut trackers = device.trackers.lock();
-                        baked.initialize_buffer_memory(&mut trackers, &snatch_guard)?;
-                        baked.initialize_texture_memory(&mut trackers, device, &snatch_guard)?;
+                        if let Err(e) = baked.initialize_buffer_memory(&mut trackers, &snatch_guard)
+                        {
+                            break 'error Err(e.into());
+                        }
+                        if let Err(e) =
+                            baked.initialize_texture_memory(&mut trackers, device, &snatch_guard)
+                        {
+                            break 'error Err(e.into());
+                        }
+
                         //Note: stateless trackers are not merged:
                         // device already knows these resources exist.
                         CommandBuffer::insert_barriers_from_device_tracker(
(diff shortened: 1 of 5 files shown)


cwfitzgerald:
> If a reviewer feels strongly, we can take the example from https://github.com/gfx-rs/wgpu/pull/5714 and turn it into a test.

It's me, I'm a reviewer :)

wgpu-core/src/device/queue.rs
```diff
@@ -1054,238 +1054,247 @@ impl Global {
             // This avoids vulkan deadlocking from the same surface texture being submitted multiple times.
             let mut submit_surface_textures_owned = FastHashMap::default();
 
-            {
-                let command_buffer_guard = hub.command_buffers.read();
+            (|| {
```
cwfitzgerald: Could we use the block-break form that we do elsewhere instead of a IILE

ErichDonGubler: I did this for ease of authoring and validating that this works, but I'd also prefer consistency with the rest of the codebase. 🙂👍🏻 Will change to use forward-moving label breaks instead.

ErichDonGubler: Resolved with a force-push to 8836d7f.

ErichDonGubler:
@cwfitzgerald:

> > If a reviewer feels strongly, we can take the example from #5714 and turn it into a test.
> 
> It's me, I'm a reviewer :)

Just added a converted test.

wgpu-core/src/device/queue.rs
```diff
@@ -1055,237 +1057,271 @@ impl Global {
             let mut submit_surface_textures_owned = FastHashMap::default();
 
             {
-                let command_buffer_guard = hub.command_buffers.read();
+                {
```
cwfitzgerald: We seem to have added a layer of indentation? what's that about?

If it's possible to remove this, or generally mitigate rightward creep, could we do that, otherwise g2g.

cwfitzgerald (approved):

wgpu-core/src/device/queue.rs
ErichDonGubler: Oh, shoot, this is a mistake, and I just fixed it with a force-push to 816c139. Good catch!

cwfitzgerald:
Nice, just started to need this in #6291 :)

PR #6322, opened 2024-09-25
metal: fix query set result copies

kvark:
**Connections**
Just a drive-by fix

**Description**
The NSRange expects length, not end element. See https://github.com/gfx-rs/metal-rs/blob/9bbe74b1d3706e46ddf41bc8aad58ee74b0bf844/src/lib.rs#L51

**Testing**
Untested

**Checklist**

- [ ] Run `cargo fmt`.

### DIFF
--- a/wgpu-hal/src/metal/command.rs
+++ b/wgpu-hal/src/metal/command.rs
@@ -490,7 +490,7 @@ impl crate::CommandEncoder for super::CommandEncoder {
             wgt::QueryType::Timestamp => {
                 encoder.resolve_counters(
                     set.counter_sample_buffer.as_ref().unwrap(),
-                    metal::NSRange::new(range.start as u64, range.end as u64),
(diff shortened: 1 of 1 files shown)


Wumpf (approved):
Thanks!

For posterity: I was weirded out by this why this works in the first place (although to be fair we see a lot of weird timer results and that might be one of the reasons!).
So checking on https://developer.apple.com/documentation/metal/mtlblitcommandencoder/3194347-resolvecounters it actually looks like this is a different range type. However, https://developer.apple.com/documentation/metal/mtlblitcommandencoder/3194347-resolvecounters?language=objc does indeed take an NSRange and that's also what's in the generated api used here https://github.com/gfx-rs/wgpu/pull/5641/files#diff-c89d45595abad64eeb8de352df1de6c8092928e3efdb1421d037c0d536115f42R521 happens.

Would be nice to check if this fixes stuff on metal and add it to the changelog. But I don't want to discourage quick fixes, so let's land this :)

PR #6308, opened 2024-09-22
[naga] Add phony as naga-ir statement

sagudev:
**Connections**
Fix #6095

**Description**
Currently phony statements get removed when compacting and when emitting, we can solve both problems by introducing new phony statement that are not compacted away and are always emitted.

Why not reusing emit? Because it makes reasoning easier (we would need to somehow encode phonies in emit statements and this gets very ugly) and phonies are just different enough to deserve special treatment. Also per emit docs: `Emit a range of expressions, visible to all statements that follow in this block`, but phony statements are not visible to other statements. This was also discussed in https://github.com/gfx-rs/naga/pull/1866#discussion_r859601915, but naga changed (and regressed) as compacting was introduced, so instead of hacking around more I think this is the most elegant solution.

**Testing**
Existing tests, new tests and servo CTS run.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`. If applicable, add:
  - [ ] `--target wasm32-unknown-unknown`
  - [ ] `--target wasm32-unknown-emscripten`
- [x] Run `cargo xtask test` to run tests.
- [ ] Add change to `CHANGELOG.md`. See simple instructions inside file.

PR #6328, opened 2024-09-26
[naga] Handle phony statements properly by treating them as named expressions

### LINKED ISSUE #6095: get_bind_group_layout fails when layout is auto

sagudev:
**Connections**
Alternative to https://github.com/gfx-rs/wgpu/pull/6308 that fixes https://github.com/gfx-rs/wgpu/issues/6095

**Description**
Currently phony statements get removed when compacting, we bypass this by making them named expressions with `phony` name.

**Testing**
New tests added and [CTS run in servo](https://github.com/sagudev/servo/actions/runs/11045854357/job/30685772542).

**Checklist**

- [x] Run `cargo fmt`.


jimblandy:
Yeah, this is another reasonable approach.

naga/src/back/wgsl/writer.rs
```diff
@@ -627,6 +627,12 @@ impl<W: Write> Writer<W> {
                 for handle in range.clone() {
                     let info = &func_ctx.info[handle];
                     let expr_name = if let Some(name) = func_ctx.named_expressions.get(&handle) {
+                        if name.starts_with("_naga_phony") {
```
jimblandy: What happens if we don't special-case this?

naga/src/front/wgsl/parse/ast.rs
```diff
@@ -284,7 +284,7 @@ pub enum StatementKind<'a> {
     },
     Increment(Handle<Expression<'a>>),
     Decrement(Handle<Expression<'a>>),
-    Ignore(Handle<Expression<'a>>),
+    Phony(Handle<Expression<'a>>, Span),
```
jimblandy: The rename is good: we should hew to the spec's terminology as much as possible.

Why does this need a Span when no other variant of `StatementKind` does? Why isn't the `Span` in `struct Statement` good enough?

naga/src/front/wgsl/lower/mod.rs
```diff
@@ -1778,12 +1778,14 @@ impl<'source, 'temp> Lowerer<'source, 'temp> {
...
-            ast::StatementKind::Ignore(expr) => {
+            ast::StatementKind::Phony(expr, span) => {
                 let mut emitter = Emitter::default();
                 emitter.start(&ctx.function.expressions);
 
-                let _ = self.expression(expr, &mut ctx.as_expression(block, &mut emitter))?;
+                let value = self.expression(expr, &mut ctx.as_expression(block, &mut emitter))?;
                 block.extend(emitter.finish(&ctx.function.expressions));
+                ctx.named_expressions
+                    .insert(value, (format!("_naga_phony_{}", value.index()), span));
```
jimblandy: I *think* if you just use the empty string for the name, everything will just work. Backends are supposed to apply the namer to everything from `naga::Function::named_expressions`, as done [here](https://github.com/gfx-rs/wgpu/blob/765c20235ea712fd14ad1b9e36bbb2071a86e319/naga/src/back/wgsl/writer.rs#L634), and that will return a fresh name for every string, including "".

Or, for legibility, just use "phony" as the name for all of them.

Please expand the snapshot test to show that these actually don't clash.

naga/tests/out/wgsl/phony_assignment.wgsl
```diff
@@ -0,0 +1,7 @@
+@group(0) @binding(0) 
+var<uniform> binding: f32;
+
+@compute @workgroup_size(1, 1, 1) 
+fn main(@builtin(global_invocation_id) id: vec3<u32>) {
+    _ = binding;
```
jimblandy: Whoa, how come `_ = a;` does away, but `_ = binding;` does not?

naga/src/back/wgsl/writer.rs
sagudev: it gets emitted as `let` instead of `_`

naga/src/front/wgsl/parse/ast.rs
sagudev: This span refers to `_` and is used when adding named expressions where we usually add span of var name, but it would make sense for phony statements to use whole statement span (I didn't know we got it).

naga/tests/out/wgsl/phony_assignment.wgsl
sagudev: because `let x = a` would also get optimized away (by compat), because it's constant (and has no side effects so this is actually ok).

naga/src/back/wgsl/writer.rs
sagudev: this is now removed

naga/src/front/wgsl/lower/mod.rs
sagudev: done in https://github.com/gfx-rs/wgpu/pull/6328/commits/0036482fe5993f3ab5f47eef10a7a720f8740861

naga/tests/out/wgsl/phony_assignment.wgsl
```diff
@@ -0,0 +1,15 @@
...
+fn five() -> i32 {
+    return 5i;
+}
+
+@compute @workgroup_size(1, 1, 1) 
+fn main(@builtin(global_invocation_id) id: vec3<u32>) {
+    let phony = binding;
+    let phony_1 = binding;
+    let _e6 = five();
+    let _e7 = five();
```
jimblandy, line 13: Okay, I think I see what's going on here: these are generated for the two `Statement::Call`s, but then the `_ = five();` and the `let b = five();` don't actually generate any code.

jimblandy (approved):
No IR or back end changes necessary! I love it.

PR #6333, opened 2024-09-27
[naga] Permit only structs as binding array elements.

jimblandy:
Require `T` to be a struct in `binding_array<T, ...>`; do not permit arrays.

In #5428, the validator was changed to accept binding array types that the SPIR-V backend couldn't properly emit. Specifically, the validator was changed to accept `binding_array<array<T>>`, but the SPIR-V backend wasn't changed to wrap the binding array elements in a SPIR-V struct type, as Vulkan requires. So the type would be accepted by the validator, and then rejected by the backend.

### DIFF
--- a/naga/src/valid/type.rs
+++ b/naga/src/valid/type.rs
@@ -680,7 +680,6 @@ impl super::Validator {
                     // Currently Naga only supports binding arrays of structs for non-handle types.
                     match gctx.types[base].inner {
                         crate::TypeInner::Struct { .. } => {}
-                        crate::TypeInner::Array { .. } => {}
                         _ => return Err(TypeError::BindingArrayBaseTypeNotStruct(base)),
                     };
                 }
--- a/naga/tests/validation.rs
+++ b/naga/tests/validation.rs
@@ -496,3 +496,113 @@ fn main(input: VertexOutput) {{
         }
     }
 }
+
+#[allow(dead_code)]
+struct BindingArrayFixture {
+    module: naga::Module,
+    span: naga::Span,
+    ty_u32: naga::Handle<naga::Type>,
+    ty_array: naga::Handle<naga::Type>,
+    ty_struct: naga::Handle<naga::Type>,
+    validator: naga::valid::Validator,
+}
+
+impl BindingArrayFixture {
+    fn new() -> Self {
+        let mut module = naga::Module::default();
+        let span = naga::Span::default();
+        let ty_u32 = module.types.insert(
+            naga::Type {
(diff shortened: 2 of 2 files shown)


ErichDonGubler:
CC @teoxoy as reviewer and @kvark as author of the regressor being fixed here.

ErichDonGubler (approved):
LGTM!

question: Are there plans in the future to relax this restriction, since the OP implies it's possible?

JMS55:
Wait, so can you no longer do `binding_array<array<T>>`? That's very unfortunate. I was using that with ray tracing (soon to be merged...) to bind multiple vertex data buffers easily. I _can_ switch to a single uberbuffer of all mesh data, but it's a lot more annoying to manage.

jimblandy:
@JMS55 wrote:
> Wait, so can you no longer do `binding_array<array<T>>`? That's very unfortunate. I was using that with ray tracing (soon to be merged...) to bind multiple vertex data buffers easily. I _can_ switch to a single uberbuffer of all mesh data, but it's a lot more annoying to manage.

I'm confused. That has never worked:

```console
$ cat in.wgsl
@group(0) @binding(0)
var<storage> a: binding_array<array<u32>>;

fn f(i: i32, j: i32) -> u32 {
   return a[i][j];
}
$ naga in.wgsl out.spv
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
     Running `/home/jimb/rust/wgpu/target/debug/naga in.wgsl out.spv`
$ spirv-val out.spv
error: line 16: Block decoration on target <id> '4[%_runtimearr_uint]' must be a structure type
  OpDecorate %_runtimearr_uint Block

$ 
```

As the commit message and PR message explains, Naga fails to wrap the `array<u32>` in a struct, so this doesn't validate.

And this restriction is easy to work around: just put the runtime-sized array in a struct in the WGSL:

```console
$ cat fixed.wgsl
struct S {
  arr: array<u32>,
}

@group(0) @binding(0)
var<storage> a: binding_array<S>;

fn f(i: i32, j: i32) -> u32 {
   return a[i].arr[j];
}
$ naga fixed.wgsl fixed.spv
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
     Running `/home/jimb/rust/wgpu/target/debug/naga fixed.wgsl fixed.spv`
$ spirv-val fixed.spv
$ 
```

JMS55:
Oh I see, yeah wrapping it in a struct is what I was doing. Didn't realize that still worked, thanks!

PR #6334, opened 2024-09-27
[naga spv-out] Document and refactor `write_runtime_array_length`.

jimblandy:
* Document and refactor `naga::back::spv::BlockContext::write_runtime_array_length`.

  Don't try to handle finding the length of a particular element of a `binding_array<array<T>>`. The SPIR-V backend doesn't wrap that type correctly anyway; #6333 changes the validator to forbid such types. Instead, assume that the elements of a `binding_array<T>` are always structs whose final members may be a runtime-sized array.

  Pull out consistency checks after the analysis of the array expression, so that we always carry out all the checks regardless of what path we took to produce the information.

* Update Vulkan spec section number, and provide link.

* Replace `match` with equivalent `!=`.

### DIFF
--- a/naga/src/back/spv/block.rs
+++ b/naga/src/back/spv/block.rs
@@ -200,10 +200,7 @@ impl<'w> BlockContext<'w> {
     fn is_intermediate(&self, expr_handle: Handle<crate::Expression>) -> bool {
         match self.ir_function.expressions[expr_handle] {
             crate::Expression::GlobalVariable(handle) => {
-                match self.ir_module.global_variables[handle].space {
-                    crate::AddressSpace::Handle => false,
-                    _ => true,
-                }
+                self.ir_module.global_variables[handle].space != crate::AddressSpace::Handle
             }
             crate::Expression::LocalVariable(_) => true,
             crate::Expression::FunctionArgument(index) => {
--- a/naga/src/back/spv/helpers.rs
+++ b/naga/src/back/spv/helpers.rs
@@ -85,7 +85,7 @@ impl crate::AddressSpace {
 
 /// Return true if the global requires a type decorated with `Block`.
 ///
-/// Vulkan spec v1.3 §15.6.2, "Descriptor Set Interface", says:
+/// In the Vulkan spec 1.3.296, the section [Descriptor Set Interface][dsi] says:
 ///
 /// > Variables identified with the `Uniform` storage class are used to
 /// > access transparent buffer backed resources. Such variables must
@@ -98,6 +98,8 @@ impl crate::AddressSpace {
 /// > -   laid out explicitly using the `Offset`, `ArrayStride`, and
 /// >     `MatrixStride` decorations as specified in §15.6.4, "Offset
 /// >     and Stride Assignment."
+///
+/// [dsi]: https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#interfaces-resources-descset
 // See `back::spv::GlobalVariable::access_id` for details.
 pub fn global_needs_wrapper(ir_module: &crate::Module, var: &crate::GlobalVariable) -> bool {
     match var.space {
--- a/naga/src/back/spv/index.rs
+++ b/naga/src/back/spv/index.rs
@@ -38,98 +38,162 @@ impl<'w> BlockContext<'w> {
     ///
     /// Given `array`, an expression referring a runtime-sized array, return the
     /// instruction id for the array's length.
+    ///
+    /// Runtime-sized arrays may only appear in the values of global
+    /// variables, which must have one of the following Naga types:
+    ///
+    /// 1. A runtime-sized array.
+    /// 2. A struct whose last member is a runtime-sized array.
+    /// 3. A binding array of 2.
+    ///
+    /// Thus, the expression `array` has the form of:
+    ///
+    /// - An optional [`AccessIndex`], for case 2, applied to...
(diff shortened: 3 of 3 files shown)


naga/src/back/spv/block.rs
```diff
@@ -200,10 +200,7 @@ impl<'w> BlockContext<'w> {
     fn is_intermediate(&self, expr_handle: Handle<crate::Expression>) -> bool {
         match self.ir_function.expressions[expr_handle] {
             crate::Expression::GlobalVariable(handle) => {
-                match self.ir_module.global_variables[handle].space {
-                    crate::AddressSpace::Handle => false,
-                    _ => true,
-                }
+                self.ir_module.global_variables[handle].space != crate::AddressSpace::Handle
```
ErichDonGubler, line 203: thought: This is also a good place to use `matches!(…)`, though I think the `PartialEq` implementation isn't going away anytime soon (so there's no reason we _have_ to use it here).

naga/src/back/spv/helpers.rs
```diff
@@ -85,7 +85,7 @@ impl crate::AddressSpace {
 
 /// Return true if the global requires a type decorated with `Block`.
 ///
-/// Vulkan spec v1.3 §15.6.2, "Descriptor Set Interface", says:
+/// In the Vulkan spec, the section [Descriptor Set Interface][dsi] says:
```
ErichDonGubler: question(non-blocking): I think it's good to avoid presenting details that need explicit maintenance like versions and section numbers if they can be encoded into a link, but I just wanted to check: was that intentional? Was this section number maybe incorrect?

naga/src/back/spv/index.rs
```diff
@@ -38,98 +38,160 @@ impl<'w> BlockContext<'w> {
     ///
     /// Given `array`, an expression referring a runtime-sized array, return the
     /// instruction id for the array's length.
+    ///
+    /// Runtime-sized arrays may only appear in the values of global
+    /// variables, which must have one of the following Naga types:
+    /// 1) A runtime-sized array.
+    /// 2) A struct whose last member is a runtime-sized array.
+    /// 3) A binding array of 2).
```
ErichDonGubler: thought: I was greatly surprised to learn that the `1)` form is accepted by Rustdoc as valid ordered list syntax!

style(non-blocking): I think it'd be better to use the `1.` syntax form, as that's more ubiquitous, but I notice that _most_ (not all 🤔) of Naga uses the `1)` form.

style(non-blocking): I find it nicer to have a newline between a list and other surrounding block content. This matches better with the styling presented in most tools I'm aware of, incl. Rustdoc.

naga/src/back/spv/index.rs
```diff
@@ -38,98 +38,160 @@ impl<'w> BlockContext<'w> {
...
                     } => match self.ir_function.expressions[base_outer] {
+                        // An `AccessIndex` of an `AccessIndex` must be a
+                        // binding array holding structs whose last members are
+                        // runtime-sized arrays.
                         crate::Expression::GlobalVariable(handle) => {
                             let index_id = self.get_index_constant(index_outer);
-                            (Some(index_id), handle, Some(index))
+                            binding_array_index_id = Some(index_id);
+                            global_handle = handle;
+                            opt_last_member_index = Some(index);
```
ErichDonGubler, line 97: nitpick(non-blocking): No action required—I understand what I'm looking at now—but it woulda been nice if the change from `let (…) = …` to `let $ident; …` had been kept separate from the documentation and error content changes.

ErichDonGubler (approved):
praise: Yay for clarity, TIL some things! Shippit!

naga/src/back/spv/block.rs
jimblandy: I did think about that. But when I see `matches!` instead of `!=`, I expect pattern-matching behavior to be relevant. In this case, it's not, so it seemed like `!=` was less distracting.

naga/src/back/spv/helpers.rs
jimblandy: I believe what happened in this case was that the section number actually *changed*. I assume the Vulkan spec editors try to keep fragment ids stable. So yes, maybe just dropping the section number would be best.

However, I think the version number is relevant. The Vulkan spec web site doesn't seem to support permalinks, so if a future version changes the language I'm quoting substantially, it seems helpful to let people know what I was actually quoting. But in that case, a full three-component spec version would be better.

So: I'll include a detailed version number, and delete the section number, and try to follow that practice elsewhere in the future.

jimblandy: I think I misremembered what this commit actually does.

1) Yes, the section number was incorrect. I believe this is because the spec changed, so to avoid having to revisit this I switched to the fragment ID, which I assume will be more stable.

2) I deleted the version number because it seemed like the link covered that. But as I said in the prior comment, in the absence of a permalink facility I can imagine it being valuable to actually say what exact version of the doc I was looking at. The link doesn't cover that.

naga/src/back/spv/index.rs
jimblandy: Yes, this could have been separated out better. Everything was written at once, and I was afraid I would introduce bugs or confuse myself trying to reconstitute them as a series of individual commits. Some of the changes invalidated some of the old comments (variable renames), so the code and doc changes couldn't really be cleanly separated out, line by line. Of course, I could have just rewritten the change from scratch, comparing with the final result, but I felt that had crossed the line of diminishing returns.

PR #6336, opened 2024-09-28
[naga spv-out] Consolidate explanation of global wrapping.

jimblandy:
Consolidate the explanation of why and how we wrap Naga global variables in structs to satisfy Vulkan's requirements, and include it in the documentation for `back::spv::GlobalVariable`.

Clarify `GlobalVariable`'s members documentation.

### DIFF
--- a/naga/src/back/spv/index.rs
+++ b/naga/src/back/spv/index.rs
@@ -67,13 +67,14 @@ impl<'w> BlockContext<'w> {
     /// - An optional [`Access`] or [`AccessIndex`], for case 3, applied to...
     /// - A [`GlobalVariable`].
     ///
-    /// The SPIR-V generated takes into account wrapped globals; see
-    /// [`global_needs_wrapper`].
+    /// The generated SPIR-V takes into account wrapped globals; see
+    /// [`back::spv::GlobalVariable`] for details.
     ///
     /// [`GlobalVariable`]: crate::Expression::GlobalVariable
     /// [`AccessIndex`]: crate::Expression::AccessIndex
     /// [`Access`]: crate::Expression::Access
     /// [`base`]: crate::Expression::Access::base
+    /// [`back::spv::GlobalVariable`]: super::GlobalVariable
     pub(super) fn write_runtime_array_length(
         &mut self,
         array: Handle<crate::Expression>,
--- a/naga/src/back/spv/mod.rs
+++ b/naga/src/back/spv/mod.rs
@@ -467,38 +467,75 @@ enum CachedConstant {
     ZeroValue(Word),
 }
 
+/// The SPIR-V representation of a [`crate::GlobalVariable`].
+///
+/// In the Vulkan spec 1.3.296, the section [Descriptor Set Interface][dsi] says:
+///
+/// > Variables identified with the `Uniform` storage class are used to access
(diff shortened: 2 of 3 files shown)


naga/src/back/spv/mod.rs
```diff
@@ -466,38 +466,75 @@ enum CachedConstant {
     ZeroValue(Word),
 }
 
+/// How a Naga global variable is represented in SPIR-V.
```
ErichDonGubler: style: I strongly suspect this would serve better as a non-punctuated header; suggested edit:


```suggestion
/// # How a Naga global variable is represented in SPIR-V
```

naga/src/back/spv/index.rs
```diff
@@ -50,15 +50,16 @@ impl<'w> BlockContext<'w> {
     ///
     /// - An optional [`AccessIndex`], for case 2, applied to...
     /// - An optional [`Access`] or [`AccessIndex`], for case 3, applied to...
-    /// - A [`GlobalVariable`].
+    /// - A [`naga::GlobalVariable`].
```
ErichDonGubler: question: Why specify `naga::…` here? 🤔 Almost all the other Naga symbols around here aren't qualified, and I don't understand why this line merits an exception.

ErichDonGubler (approved):
LGTM, some non-blocking nits.

naga/src/back/spv/index.rs
jimblandy: Unfortunately, we have two types named `GlobalVariable` in play: the Naga IR's (`naga::GlobalVariable`) and the SPIR-V backend's (`naga::back::spv::GlobalVariable`). So pretty much every reference to `GlobalVariable` in the docs needs to be considered for ambiguity.

In this case, since we're talking about `naga::Expression`, it's probably clear enough, and the qualification is more of a distraction than a clarification.

naga/src/back/spv/mod.rs
jimblandy: The first paragraph of a doc string is used as the summary shown during search. It seems like having a header show up there would be weird?

jimblandy: I changed the first paragraph to be a bit less awkward.

Here's how search results look when the summary is an ordinary paragraph:
global-variable-pre

Here's how the page itself looks when the summary is an ordinary paragraph:
global-var-page-pre

Here's how search results look when the summary is a header:
global-variable-post

Basically, there's no difference, except for the missing period.

And here's how the page itself looks when the summary is a header:
global-var-page-post

Overall, I think having a summary paragraph be just a paragraph is better.

ErichDonGubler: Ahhh, the flavor of the edit to the summary paragraph makes it definitely feel like a summary of the type now, and that feels good. Previously, I thought it had maybe been a contextual phrase for the following content, rather than a summary of what the type is. 🤦🏻‍♂️

jimblandy: Yeah, that's fair. I think the "How a..." phrasing is a bit off, for the summary paragraph role.

Actually, it probably shouldn't even be a "paragraph" proper. "Noun clause" is closer. "The SPIR-V representation of a Naga global variable" is a noun clause. "How a ..." was sort of a noun-ification of an action, and I feel it was pushing the boundaries.

PR #6341, opened 2024-09-29
Fix unaligned slice::from_raw_parts in gles push contant handling.

### LINKED ISSUE #6327: Missing precondition for `slice::from_raw_parts` in gles backend
We got a report on this here https://gitlab.com/veloren/veloren/-/issues/2015 where some debug assertions catch a missing precondition. However, note this is on an older version and the `get_data` method has changed a bit.

A quick look at the current version suggests that the alignment might not be checked and the function itself has a note about how it should be replaced: https://github.com/gfx-rs/wgpu/blob/e7f891bf2b455927b98e70301c707428c7adf62a/wgpu-hal/src/gles/queue.rs#L1608-L1609

Imberflur:
**Connections**
Fixes https://github.com/gfx-rs/wgpu/issues/6327

**Description**

Fixes calling `slice::from_raw_parts` with an unaligned pointer by using `bytemuck::pod_read_unaligned` instead (which copies the value to the stack). I assume this is fine for values of the sizes used in push constants. If it seems useful to try to avoid this, then the push constants need to be stored aligned instead (actually looking at this, it would probably be easy to make `CommandBuffer::add_push_constant_data` include some padding to align the data).

This adds a new dependency on `bytemuck` in `wgpu-hal`. I don't know whether adding this is acceptable (e.g. for firefox).

**Testing**

I updated to a recent rustc nightly (`rustc 1.83.0-nightly (2bd1e894e 2024-09-26)`) and used `rustup override set nightly` followed by `cargo xtask test` on a linux system where the gles backend is present.

Before the fix I saw panics from new debug assertions in `slice::from_raw_parts` with these tests:
* `wgpu_test::regression::issue_3349::multi_stage_data_binding`
* `wgpu_test::shader::struct_layout::push_constant_input`

After the fix these test no longer panic.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,7 +76,7 @@ arrayvec = "0.7"
 bincode = "1"
 bit-vec = "0.8"
 bitflags = "2.6"
-bytemuck = { version = "1.18", features = ["derive"] }
+bytemuck = { version = "1.18" }
 cfg_aliases = "0.1"
 cfg-if = "1"
 criterion = "0.5"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -126,6 +126,7 @@ By @bradwerth [#6216](https://github.com/gfx-rs/wgpu/pull/6216).
 #### GLES / OpenGL
 
 - Fix GL debug message callbacks not being properly cleaned up (causing UB). By @Imberflur in [#6114](https://github.com/gfx-rs/wgpu/pull/6114)
+- Fix calling `slice::from_raw_parts` with unaligned pointers in push constant handling. By @Imberflur in [#6341](https://github.com/gfx-rs/wgpu/pull/6341)
 
 #### WebGPU
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3681,6 +3681,7 @@ dependencies = [
  "bit-set",
  "bitflags 2.6.0",
  "block",
+ "bytemuck",
  "cfg-if",
  "cfg_aliases",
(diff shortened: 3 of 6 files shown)


Cargo.toml
```diff
@@ -76,7 +76,7 @@ arrayvec = "0.7"
 bincode = "1"
 bit-vec = "0.8"
 bitflags = "2.6"
-bytemuck = { version = "1.18", features = ["derive"] }
+bytemuck = { version = "1.18" }
```
Wumpf, line 79: nice!

Wumpf (approved):
thank you! Agreed, the copy should be fine for push constants

> This adds a new dependency on bytemuck in wgpu-hal. I don't know whether adding this is acceptable (e.g. for firefox).

Should be fine, it's already used in wgpu-core if `spv-in` is enabled, although that feature is probably not enabled in Firefox?
cc: @ErichDonGubler to confirm that this won't be a problem

ErichDonGubler:
@Wumpf: Doesn't seem problematic to me. The biggest obstacle to most new dependencies is the work to vet the soundness of the crate, and `bytemuck` has been audited by the Chromium team/Google in a way that's shared with Firefox/Mozilla:

https://github.com/google/rust-crate-audits/blob/a0fe5aac7671f88eb32eea18dd0df1c95d0f1312/audits.toml#L1228-L1233

CC @jimblandy, who is slated to handle a re-vendor from WGPU into Firefox next.

PR #6349, opened 2024-10-01
Unbreak `rustfmt` in some places

ErichDonGubler:
**Connections**

Discovered during #6148.

**Description**

Hunks of Rust source from <https://github.com/gfx-rs/wgpu/pull/6316> and prior have some lines with string literals that are so long that `rustfmt` refuses to reformat both the literal and surrounding runs of expressions. Fix this, and keep chugging with delicious `rustfmt`'d code.

**Testing**

Still green in CI, still good!

**Checklist**

- [x] Run `cargo fmt`.
- [x] ~~Add change to `CHANGELOG.md`. See simple instructions inside file.~~ Definitely not necessary.

### DIFF
--- a/naga/src/back/pipeline_constants.rs
+++ b/naga/src/back/pipeline_constants.rs
@@ -14,7 +14,10 @@ use thiserror::Error;
 pub enum PipelineConstantError {
     #[error("Missing value for pipeline-overridable constant with identifier string: '{0}'")]
     MissingValue(String),
-    #[error("Source f64 value needs to be finite (NaNs and Inifinites are not allowed) for number destinations")]
+    #[error(
+        "Source f64 value needs to be finite ({}) for number destinations",
+        "NaNs and Inifinites are not allowed"
+    )]
     SrcNeedsToBeFinite,
     #[error("Source f64 value doesn't fit in destination")]
     DstRangeTooSmall,
--- a/naga/src/span.rs
+++ b/naga/src/span.rs
@@ -11,6 +11,7 @@ pub struct Span {
 
 impl Span {
     pub const UNDEFINED: Self = Self { start: 0, end: 0 };
+
     /// Creates a new `Span` from a range of byte indices
     ///
     /// Note: end is exclusive, it doesn't belong to the `Span`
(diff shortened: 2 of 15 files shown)


naga/src/span.rs
```diff
@@ -11,6 +11,7 @@ pub struct Span {
 
 impl Span {
     pub const UNDEFINED: Self = Self { start: 0, end: 0 };
+
```
ErichDonGubler, line 14: This change here is only tangentially related to the subject of this PR, but I figured it didn't hurt to include.

naga/src/back/pipeline_constants.rs
```diff
@@ -14,7 +14,10 @@ use thiserror::Error;
 pub enum PipelineConstantError {
     #[error("Missing value for pipeline-overridable constant with identifier string: '{0}'")]
     MissingValue(String),
-    #[error("Source f64 value needs to be finite (NaNs and Inifinites are not allowed) for number destinations")]
+    #[error(
+        "Source f64 value needs to be finite ({}) for number destinations",
+        "NaNs and Inifinites are not allowed"
+    )]
```
jimblandy, line 20: I'm sure there's a reason, but, explain to me why we can't just do

```rust
#[error(
    "Source f64 value needs to be finite \
     (NaNs and Infinities are not allowed) \
     for number destinations"
)]
```

cwfitzgerald: It will (marginally) help debugging if you say what the value is so you know what to look for.

ErichDonGubler: The reason is that I simply hadn't considered it, TBH. I susepct it might still break `rustfmt`, but worth a try!

jimblandy: > It will (marginally) help debugging if you say what the value is so you know what to look for.

@cwfitzgerald "Please File An Issue"

ErichDonGubler: Confirmed that this breaks `rustfmt`'s indentation, though not surrounding expressions. The literal will simply never have the indentation of the second and subsequent lines after the first line break.

I also tried `"{}", concat!(…)`, but this also breaks `rustfmt`. 😩

ErichDonGubler:
@jimblandy: Approval plz?

jimblandy (approved):

PR #6351, opened 2024-10-01
[naga spv-out] bounds-check runtime-sized array access correctly.

### LINKED ISSUE #4441: SPIR-V back end skips bounds checks for `AccessIndex` when applied to runtime-sized arrays
Naga's SPIR-V back end fails to generate a bounds check for an `AccessIndex` expression that is indexing into a runtime-sized array.

For example:

```
[[block]]
struct Unsized {
    arr: array<f32>;
};

[[group(0), binding(0)]] var<storage> unsized: Unsized;

fn fetch(i: i32) -> f32 {
   return unsized.arr[3] + unsized.arr[i];
}
```

When the `ReadZeroSkipWrite` policy is applied to buffers, Naga generates the following SPIR-V for the body:

```
      %fetch = OpFunction %float None %13
          %i = OpFunctionParameter %int
         %10 = OpLabel
               OpBranch %14
         %14 = OpLabel
         %20 = OpAccessChain %_ptr_StorageBuffer_float %unsized %uint_0 %uint_3
         %21 = OpLoad %float %20
         %22 = OpArrayLength %uint %unsized 0
         %23 = OpULessThan %bool %i %22
               OpSelectionMerge %27 None
               OpBranchConditional %23 %28 %27
         %28 = OpLabel
         %25 = OpAccessChain %_ptr_StorageBuffer_float %unsized %uint_0 %i
         %29 = OpLoad %float %25
               OpBranch %27
         %27 = OpLabel
         %30 = OpPhi %float %26 %14 %29 %28
         %31 = OpFAdd %float %21 %30
               OpReturnValue %31
               OpFunctionEnd
```

Note that the the `unsized.arr[3]` access has no bounds check, while the `unsized.arr[i]` access does. Since it is not known until run time whether `3` is a valid index, both accesses should be checked.

Two factors limit the impact of this bug:

- The problem is restricted to runtime-sized arrays: the validator [performs bounds checks](https://github.com/gfx-rs/naga/blob/e69a70bfb7b3dff1391d5189d5c68fb798c02040/src/valid/expression.rs#L254) for `AccessIndex` expressions applied to any type whose length is statically known.
- We expect that many targets will support a form of robust buffer access that would make bounds checks unnecessary on arrays in the `storage` storage class, which is the only storage class that can hold dynamically sized arrays.

However, Naga is expected to be used in situations where that kind of robust buffer access is not supported, so this is still important to fix.

jimblandy:
Do not neglect to apply bounds checks to indexing operations on runtime-sized arrays, even when they are accessed via an `AccessIndex` instruction.

Before this commit, `BlockContext::write_expression_pointer` would not apply bounds checks to `OpAccessChain` indices provided by an `AccessIndex` instruction, apparently with the rationale that any out-of-bounds accesses should all have been reported by constant evaluation.

While it is true that the `index` operand of an `AccessIndex` expression is known at compile time, and that the WGSL constant evaluation rules require accesses that can be statically determined to be out-of-bounds to be shader creation or pipeline creation time errors, accesses to runtime-sized arrays don't follow this pattern: even if the index is known, the length with which it must be compared is not.

Fixes #4441.

-   [naga] Extend snapshot tests for bounds checks.
    
    Extend the snapshot tests for bounds checking to cover the case where a runtime-sized array is indexed by a constant.

-   [naga spv-out] Let `BoundsCheckResult` supply the index value.
    
    Let `BoundsCheckResult::Conditional` provide both the condition to check before carrying out the access, and the index to use for that access. The `Conditional` variant indicates that we generated a runtime bounds check, which implies we must have had a SPIR-V id for the index to pass to that check, so there's no reason not to provide that to the callers - especially if the bounds check code was able to reduce it to a known constant.
    
    At the moment, this is not much of a refactor, but later commits will use `GuardedIndex` in more places, at which point this will avoid a re-matching and assertion.

-   [naga spv-out] Abstract out `OpAccessChain` index production.
    
    Abstract the code from `write_expression_pointer` to handle one indexing operation out into its own function, `BlockContext::write_access_chain_index`.

-   [naga spv-out] Consolidate code to find index values.
    
    Let the SPIR-V backend use `GuardedIndex::try_resolve_to_constant`, rather than writing out its definition in `write_restricted_index` and `write_index_comparison`.
    
    Call `try_resolve_to_constant` in one place, in `write_bounds_check`, and simply pass the `GuardedIndex` into subroutines.
    
    Reduce `write_restricted_index` and `write_index_comparison` to case analysis and code generation.
    
    Note that this commit does have a benign effect on SPIR-V snapshot output for programs like this:
    
        let one_i = 1i;
        var vec0 = vec3<i32>();
        vec0[one_i] = 1;
    
    The value indexing `vec0` here is an `i32`, but after this commit, the operand to `OpAccessChain` becomes a `u32` constant (with the same value).
    
    This is because `write_bounds_check` now calls `try_resolve_to_constant` itself, rather than deferring this work to its callees, so it may return `BoundsCheckResult::KnownInBounds` even when the `Unchecked` policy is in force. This directs the caller, `write_expression_pointer`, to treat the `OpAccessChain` operand as a fresh `u32` constant, rather than simply passing through the original `i32` expression.

-   [naga spv-out] Abstract extending a bounds check condition chain.
    
    Introduce a new function, `BlockContext::extend_bounds_check_condition_chain`, which adds a new boolean condition to the chain of bounds checks guarding an `OpAccessChain` instruction.

-   [naga spv-out] Doc fix: typo

-   [naga spv-out] Abstract out non-uniform binding array access test.
    
    Introduce a new function, `BlockContext::is_nonuniform_binding_array_access`, which determines whether a given array access expression means that the `OpAccessChain` instruction must have a `NonUniform` decoration.

### DIFF
--- a/naga/src/back/spv/block.rs
+++ b/naga/src/back/spv/block.rs
@@ -1935,11 +1935,14 @@ impl<'w> BlockContext<'w> {
                 Ok(self.writer.get_constant_scalar(scalar))
             }
             BoundsCheckResult::Computed(computed_index_id) => Ok(computed_index_id),
-            BoundsCheckResult::Conditional(comparison_id) => {
-                self.extend_bounds_check_condition_chain(accumulated_checks, comparison_id, block);
+            BoundsCheckResult::Conditional {
+                condition_id: condition,
+                index_id: index,
+            } => {
+                self.extend_bounds_check_condition_chain(accumulated_checks, condition, block);
 
                 // Use the index from the `Access` expression unchanged.
-                Ok(self.cached[index])
+                Ok(index)
             }
         }
     }
--- a/naga/src/back/spv/block.rs
+++ b/naga/src/back/spv/block.rs
@@ -1828,19 +1828,8 @@ impl<'w> BlockContext<'w> {
         let root_id = loop {
             expr_handle = match self.ir_function.expressions[expr_handle] {
                 crate::Expression::Access { base, index } => {
-                    if let crate::Expression::GlobalVariable(var_handle) =
-                        self.ir_function.expressions[base]
-                    {
-                        // The access chain needs to be decorated as NonUniform
-                        // see VUID-RuntimeSpirv-NonUniform-06274
-                        let gvar = &self.ir_module.global_variables[var_handle];
-                        if let crate::TypeInner::BindingArray { .. } =
-                            self.ir_module.types[gvar.ty].inner
-                        {
-                            is_non_uniform_binding_array =
-                                self.fun_info[index].uniformity.non_uniform_result.is_some();
-                        }
-                    }
+                    is_non_uniform_binding_array |=
+                        self.is_nonuniform_binding_array_access(base, index);
 
                     let index_id = match self.write_bounds_check(base, index, block)? {
                         BoundsCheckResult::KnownInBounds(known_index) => {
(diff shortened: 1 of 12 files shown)


ErichDonGubler (commented):
I got up to (but not including) the `[naga spv-out] Consolidate code to find index values.` So far, the review experience has been nice!

teoxoy (approved):
Looks great!

jimblandy:
Hmm, this needs to be rebased across #6188.

Issue #6350, opened 2024-10-01
WGSL: Support the `requires` directive


ErichDonGubler:
Standard WGSL supports something called the `requires` global directive, which allows shaders to declare that certain extensions either need to be supported by the WebGPU host, or shader compilation should fail. It's used like this:

```wgsl
requires readonly_and_readwrite_storage_textures; // single extension
requires packed_4x8_integer_dot_product, unrestricted_pointer_parameters; // multiple extensions, comma-separated
```

WGSL standard link: https://www.w3.org/TR/WGSL/#requires-directive

PR #6352, opened 2024-10-01
Add a skeleton for directives that reports nice errors

ErichDonGubler:
**Connections**

Dependency of #6148.

May conflict with changes in #5701. If this merges first, I'll resolve conflicts there.

Lays groundwork for:

* #5476 (see also <https://github.com/gfx-rs/wgpu/pull/5701>)
* #6148
* #6350

**Description**

**Recommended review experience is commit-by-commit. Each individual commit should pass CI!**

There's not a parse path for _any_ directives, let alone what I'm working on with the `diagnostic(…)` directive in #6148. It turns out that adding the starting point with some diagnostics is enough code that a PR checkpoint sounds wise. As a bonus, this can also serve as the base for a full solution to <https://github.com/gfx-rs/wgpu/issues/5476> and <https://github.com/gfx-rs/wgpu/issues/6350>.

**Testing**

Tests will be added before this PR comes out of draft, to ensure that the error cases added here are covered.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,7 @@ By @bradwerth [#6216](https://github.com/gfx-rs/wgpu/pull/6216).
 - Support local `const` declarations in WGSL. By @sagudev in [#6156](https://github.com/gfx-rs/wgpu/pull/6156).
 - Implemented `const_assert` in WGSL. By @sagudev in [#6198](https://github.com/gfx-rs/wgpu/pull/6198).
 - Support polyfilling `inverse` in WGSL. By @chyyran in [#6385](https://github.com/gfx-rs/wgpu/pull/6385).
+- Add an internal skeleton for parsing `requires`, `enable`, and `diagnostic` directives that don't yet do anything besides emit nicer errors. By @ErichDonGubler in [#6352](https://github.com/gfx-rs/wgpu/pull/6352).
 
 #### General
 
--- a/naga/src/front/wgsl/mod.rs
+++ b/naga/src/front/wgsl/mod.rs
@@ -58,3 +58,21 @@ impl Frontend {
 pub fn parse_str(source: &str) -> Result<crate::Module, ParseError> {
     Frontend::new().parse(source)
 }
+
+#[cfg(test)]
+#[track_caller]
+pub fn assert_parse_err(input: &str, snapshot: &str) {
+    let output = parse_str(input)
+        .expect_err("expected parser error")
+        .emit_to_string(input);
+    if output != snapshot {
+        for diff in diff::lines(snapshot, &output) {
+            match diff {
+                diff::Result::Left(l) => println!("-{l}"),
+                diff::Result::Both(l, _) => println!(" {l}"),
+                diff::Result::Right(r) => println!("+{r}"),
+            }
+        }
+        panic!("Error snapshot failed");
+    }
+}
--- a/naga/src/front/wgsl/parse/directive.rs
+++ b/naga/src/front/wgsl/parse/directive.rs
@@ -1,0 +1,179 @@
+//! WGSL directives. The focal point of this API is [`DirectiveKind`].
+//!
+//! See also <https://www.w3.org/TR/WGSL/#directives>.
+
+/// A parsed sentinel word indicating the type of directive to be parsed next.
+#[derive(Clone, Copy, Debug, Hash, Eq, PartialEq)]
+pub enum DirectiveKind {
+    Unimplemented(UnimplementedDirectiveKind),
+}
+
+impl DirectiveKind {
+    const DIAGNOSTIC: &'static str = "diagnostic";
+    const ENABLE: &'static str = "enable";
+    const REQUIRES: &'static str = "requires";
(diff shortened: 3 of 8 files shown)


naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2474,6 +2480,29 @@ impl Parser {
...
+            if let Ok(ident) = lexer.peek_ident() {
+                self.push_rule_span(Rule::Directive, &mut lexer);
+                if let Some((kind, name)) = DirectiveKind::from_ident(ident.name) {
+                    match kind {
+                        DirectiveKind::Unimplemented(unimplemented) => {
+                            return Err(Error::DirectiveNotYetImplemented {
+                                kind: name,
+                                span: ident.span,
+                                tracking_issue_num: unimplemented.tracking_issue_num(),
+                            })
```
ErichDonGubler: nitpick: I think I should consolidate the `kind` (`&str`) and `tracking_issue_num` fields into the `kind` of the unimplemented `enum`. The `&str` and `u16` can be derived from the `enum`.

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2474,6 +2480,29 @@ impl Parser {
...
+                            span,
+                            tracking_issue_num: unimplemented.tracking_issue_num(),
+                        })
+                    }
+                }
+                self.pop_rule_span(&lexer);
+            } else {
+                break;
+            }
+            // TODO: add directives to set of expected things in errors later
```
ErichDonGubler: note: I couldn't think of a reasonable, but I think it's fine to land this MVP without it. Filed https://github.com/gfx-rs/wgpu/issues/6421 for follow-up.

naga/src/front/wgsl/mod.rs
```diff
@@ -58,3 +58,21 @@ impl Frontend {
...
+        for diff in diff::lines(snapshot, &output) {
+            match diff {
+                diff::Result::Left(l) => println!("-{l}"),
+                diff::Result::Both(l, _) => println!(" {l}"),
+                diff::Result::Right(r) => println!("+{r}"),
+            }
+        }
+        panic!("Error snapshot failed");
+    }
+}
```
ErichDonGubler, line 78: note: I hated duplicating the `check` function in `naga`'s `tests/` directory, but this was the only solution I could find that made it possible to contain `derive(strum::EnumIter)` on the new `enum`s being added within a `cfg_attr(test, …)`. I haven't figured out a way to keep this function here, and to expose it to those `tests/` later.

CHANGELOG.md
```diff
@@ -85,6 +85,7 @@ By @bradwerth [#6216](https://github.com/gfx-rs/wgpu/pull/6216).
 - Support local `const` declarations in WGSL. By @sagudev in [#6156](https://github.com/gfx-rs/wgpu/pull/6156).
 - Implemented `const_assert` in WGSL. By @sagudev in [#6198](https://github.com/gfx-rs/wgpu/pull/6198).
 - Support polyfilling `inverse` in WGSL. By @chyyran in [#6385](https://github.com/gfx-rs/wgpu/pull/6385).
+- Add an internal skeleton for parsing `requires`, `enable`, and `diagnostic` directives that don't yet do anything besides emit nicer . By @ErichDonGubler in [#6352](https://github.com/gfx-rs/wgpu/pull/6352).
```
teoxoy:
```suggestion
- Add an internal skeleton for parsing `requires`, `enable`, and `diagnostic` directives that don't yet do anything besides emit nicer errors. By @ErichDonGubler in [#6352](https://github.com/gfx-rs/wgpu/pull/6352).
```

naga/src/front/wgsl/parse/directive.rs
```diff
@@ -0,0 +1,179 @@
+//! WGSL directives. The focal point of this API is [`DirectiveKind`].
+//!
+//! See also <https://www.w3.org/TR/WGSL/#directives>.
+
+/// A parsed sentinel word indicating the type of directive to be parsed next.
+#[derive(Clone, Copy, Debug, Hash, Eq, PartialEq)]
+pub enum DirectiveKind {
+    Unimplemented(UnimplementedDirectiveKind),
```
teoxoy, line 8: I feel like we can have all 3 variants here even if they are not implemented but I won't block on this given that we will implement all soon enough.

teoxoy (approved):

CHANGELOG.md
ErichDonGubler: oderp how did i miss this 😭

naga/src/front/wgsl/parse/directive.rs
ErichDonGubler: I actually think the `Unimplemented*` type state has been helpful for avoiding extraneous `match`ing in other logic, specifically for the diagnostics. Do you think it's not carrying its weight? 🤔

teoxoy:
@ErichDonGubler CI is stuck on a typo.

PR #6353, opened 2024-10-01
feat(wgsl-in): parse `diagnostic(…)` attributes on `fn`s

### LINKED ISSUE #5320: Support diagnostic filters in WGSL (i.e., `diagnostic(…)`)

ErichDonGubler:
**Connections**

* Depends on <https://github.com/gfx-rs/wgpu/pull/6533>.
* Partially resolves #5320.

**Description**

Building on #6148 and #6533, we use all the plumbing we made for tracking diagnostic rules, and add a (successful) parse path for `@diagnostic(…) fn function() { … }`. This and #6148's parse paths are what I've observed to be the next most common places for users to apply diagnostic filtering, by a wide margin. I expect we will 

**Testing**

- [x] Demonstrate that the following CTS cases work:
	- [x] [`webgpu:shader,validation,parse,diagnostic:valid_locations:*`](https://gpuweb.github.io/cts/standalone/?q=webgpu:shader,validation,parse,diagnostic:valid_locations:*)
		- [x] `…:type="attribute";location="function"`
	- [x] [`webgpu:shader,validation,parse,diagnostic:duplicate_attribute_same_location:*`](https://gpuweb.github.io/cts/standalone/?q=webgpu:shader,validation,parse,diagnostic:duplicate_attribute_same_location:*)
		- `…:loc="function";same_rule=true`: ✅
		- `…:loc="function";same_rule=false`: ❌ → ✅
	- [x] ~~[`webgpu:shader,validation,parse,diagnostic:conflicting_attribute_different_location:*`](https://gpuweb.github.io/cts/standalone/?q=webgpu:shader,validation,parse,diagnostic:conflicting_attribute_different_location:*)~~ Only tests statement-position attributes. 🙁
	- [x] ~~At least one case in [`webgpu:shader,validation,parse,diagnostic:diagnostic_scoping:*`](https://gpuweb.github.io/cts/standalone/?q=webgpu:shader,validation,parse,diagnostic:diagnostic_scoping:*), specific case TBD.~~ Only tests statement-position attributes. 🙁
- [x] ~~Port a meaningful subset of the above to our own Naga test suite, to prevent the most impactful regressions.~~ Punted for later.

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,12 +85,12 @@ fn main(@builtin(position) p : vec4f) -> @location(0) vec4f {
 
 There are some limitations to keep in mind with this new functionality:
 
-- We do not yet support `diagnostic(…)` rules in attribute position (i.e., `@diagnostic(…) fn my_func { … }`). This is being tracked in <https://github.com/gfx-rs/wgpu/issues/5320>. We expect that rules in `fn` attribute position will be relaxed shortly (see <https://github.com/gfx-rs/wgpu/pull/6353>), but the prioritization for statement positions is unclear. If you are blocked by not being able to parse `diagnostic(…)` rules in statement positions, please let us know in that issue, so we can determine how to prioritize it!
+- We support `@diagnostic(…)` rules as `fn` attributes, but prioritization for rules in statement positions (i.e., `if (…) @diagnostic(…) { … }` is unclear. If you are blocked by not being able to parse `diagnostic(…)` rules in statement positions, please let us know in <https://github.com/gfx-rs/wgpu/issues/5320>, so we can determine how to prioritize it!
 - Standard WGSL specifies `error`, `warning`, `info`, and `off` severity levels. These are all technically usable now! A caveat, though: warning- and info-level are only emitted to `stderr` via the `log` façade, rather than being reported through a `Result::Err` in Naga or the `CompilationInfo` interface in `wgpu{,-core}`. This will require breaking changes in Naga to fix, and is being tracked by <https://github.com/gfx-rs/wgpu/issues/6458>.
 - Not all lints can be controlled with `diagnostic(…)` rules. In fact, only the `derivative_uniformity` triggering rule exists in the WGSL standard. That said, Naga contributors are excited to see how this level of control unlocks a new ecosystem of configurable diagnostics.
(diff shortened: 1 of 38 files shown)


ErichDonGubler:
I haven't tried consuming this in Firefox yet, but I intend to, to test the CTS. Requesting review now for overall code and `CHANGELOG`.

ErichDonGubler:
Populated the OP with test results.

teoxoy (approved):
LGTM

PR #6354, opened 2024-10-01
[naga spv-out] Don't emit unreachable blocks that jump into loops.

### LINKED ISSUE #6220: Unable to break out of the loop in SPIR-V
**Description**
Getting invalid SPIR-V out of Naga:
>error: line 36: Block '23[%23]' branches to the loop continue target '16[%16]', but is not contained in the associated loop construct '13[%13]'
  OpBranch %16

**Repro steps**
Program:
```wgsl
fn main() {
    for (var i = 0; i < 4; i += 1) {
      break;
    }
}
```

Expected to produce valid SPIR-V

jimblandy:
When generating SPIR-V, avoid generating unreachable blocks following statements like `break`, `return`, and so on that cause non-local exits. These unreachable blocks can cause SPIR-V validation to fail.

Fixes #6220.


jimblandy:
This needs a snapshot test, but I think it's mostly ready for review.

[edit: snapshot test added]

jimblandy:
That comment in `write_block` is awfully long. Maybe it belongs in the issue, and the comment could just refer to that.

jimblandy:
I think the other criticism of this PR is that, if it's that important to avoid creating unreachable blocks, every call to `write_block` probably needs some attention.

I did look at the other calls, and I think in many cases there's no change needed: the "unreachable" block is listed as some OpThingyMerge instruction's operand, so even if control can't reach it, it's still "structurally reachable" in SPIR-V's terms.

But even if that's so, those sites need a comment to that effect.

naga/src/back/spv/block.rs
```diff
@@ -2077,14 +2096,60 @@ impl<'w> BlockContext<'w> {
...
+                        // loop's continuing block.
+                        //
+                        // The result is that we will introduce an unreachable SPIR-V
+                        // block that branches into the loop construct.
+                        //
+                        // It seems to me that if a block is unreachable, its behavior
+                        // shouldn't affect the validity of the SPIR-V module. The newer
+                        // versions of the SPIR-V specification talk about "structural
+                        // reachability" that I would have expected to address this. But
+                        // the SPIR-V validator still complains.
```
ErichDonGubler: style: Please don't use singular first-person in a shared codebase; it doesn't really have meaning in code like this. 😅

naga/src/back/spv/block.rs
```diff
@@ -2034,14 +2034,33 @@ impl<'w> BlockContext<'w> {
...
+    /// [`Break`]: Statement::Break
+    /// [`Continue`]: Statement::Continue
     pub(super) fn write_block(
         &mut self,
         label_id: Word,
         naga_block: &crate::Block,
         exit: BlockExit,
         loop_context: LoopContext,
         debug_info: Option<&DebugInfoInner>,
-    ) -> Result<(), Error> {
```
ErichDonGubler, line 2070: nitpick(non-blocking): I think it's difficult to keep in mind what this `bool` means while reading code outside of the signature.

suggestion: Let's use a one-off `enum`:

```rust
enum ControlReachesEndOfBlock {
    Yes,
    No,
}
```

naga/src/back/spv/block.rs
```diff
@@ -2693,6 +2758,6 @@ impl<'w> BlockContext<'w> {
         };
 
         self.function.consume(block, termination);
-        Ok(())
+        Ok(true)
```
ErichDonGubler: nitpick(non-blocking): I am concerned that while it was easiest to author this PR with `Ok(true)`, the easiest thing to read and verify later is likely to be a form where we are forced to decide per variant what the answer to the question of "do we expect control flow to connect to the end of the block".

suggestion: Let's change this code to use a forced assignment of the "end of block" control flow query in each variant of `Statement` matching.

I'm happy to make the change myself, if initial effort is the only objection.

naga/src/back/spv/block.rs
```diff
@@ -2034,14 +2034,33 @@ impl<'w> BlockContext<'w> {
...
+    ///
+    /// If the block contains [`Break`] or [`Continue`] statements,
+    /// `loop_context` supplies the labels of the SPIR-V blocks to jump to. If
+    /// either of these labels are `None`, then it should have been a Naga
+    /// validation error for the corresponding statement to occur in this
+    /// context.
+    ///
+    /// Return `true` if control can reach the end of the block. This means that
+    /// `exit` was used. Otherwise, return `false`, indicating that `exit` was
+    /// unused.
```
ErichDonGubler: nitpick(non-blocking): It initially caused me some confusion that the return type of this function was only consumed by itself recursively. Maybe a start-of-recursion wrapper around this that discards the return type (and making this function itself private, renaming to indicate it's a helper) would be interesting?

ErichDonGubler (approved):
LGTM, minus non-blocking nits.

naga/src/back/spv/block.rs
jimblandy: Okay. I'd considered doing that. `bool` is usually not great in these situations.

naga/src/back/spv/block.rs
jimblandy: I don't think that would be an improvement. For readers, that's effectively adding a note to most arms of this match saying, "This concern, which you might not have been aware of, isn't relevant here."

More generally: Being aware of all concerns all the time would certainly eliminate bugs, but it's not a feasible way for humans to operate. Helping people ignore what's not relevant is the name of the game.

Your earlier suggestion actually shows a much better way to address this concern. If we introduce a new enum to use in the function's return type, then a contributor cannot add an early `return` (thus skipping the `exit` processing) without having to consider which enum variant to return. This forces them to consider the question at exactly and only the points where it becomes relevant.

naga/src/back/spv/block.rs
jimblandy: I'll implement the suggestion above, and then let's look this over.

As I remarked in an earlier comment, I think the other call sites might want to be more explicit about why they're ignoring the return value.

naga/src/back/spv/block.rs
jimblandy: I think my other commit history makes it clear that, if I'm trying to optimize for ease of authoring, I'm pretty bad at that. 

But what I hear you saying is, this PR was a pain to make sense of.

naga/src/back/spv/block.rs
jimblandy: Okay, I've done this, in a prep refactoring commit. It let us make two other types private, and reduce the number of arguments to the public function, so this was a win.

jimblandy:
Should have used checkboxes for these:

> This needs a snapshot test, but I think it's mostly ready for review.

A snapshot test has been added.

> That comment in `write_block` is awfully long. Maybe it belongs in the issue, and the comment could just refer to that.

The comment was moved to the snapshot test input file, along with a reference to the issue.

naga/src/back/spv/block.rs
```diff
@@ -2039,14 +2063,33 @@ impl<'w> BlockContext<'w> {
...
```
ErichDonGubler: nitpick: This last paragraph no longer applies, now that we've refactored to use `BlockExitDisposition`. I think we can just remove it?

Pushed suggestion as a `fixup!` with 1505975d.

naga/src/back/spv/block.rs
```diff
@@ -59,6 +59,30 @@ enum BlockExit {
     },
 }
 
+/// What code generation did with a provided [`BlockExit`] value.
+///
+/// A function that accepts a [`BlockExit`] argument should return a value of
```
ErichDonGubler, line 64: nitpick(non-blocking): We only use this for a single function, so it's sorta weird to say "a function" like it might be typical to use in multiple places. Not really a concern, tho.

ErichDonGubler (approved):
I like the new `enum`! 😀 lezduit, pushed a `fixup!` that you can just autosquash-and-rebase-and-merge if you're happy with it.

jimblandy:
> I like the new `enum`! 😀 lezduit, pushed a `fixup!` that you can just autosquash-and-rebase-and-merge if you're happy with it.

Oh, good catch, thanks *sigh*

PR #6356, opened 2024-10-02
Remove redundant let binding

eliemichel:
**Description**
Isn't this `if let` condition redundant with the `match` case?

I know this is a bit of a detail, but I also do this PR to check that I correctly understand what's going on.


nical (approved):

nical:
Indeed!

Issue #6358, opened 2024-10-03
[naga spv-out] Indexing arrays by value causes unnecessary copies


jimblandy:
Given code like this:

```wgsl
fn (aaa: array<array<array<f32, 3>, 3> 3>, i: i32, j: i32, k: i32) -> f32 {
    return aaa[i][j][k];
}
```

the changes in #6188 will copy each intermediate indexing result into a separate variable. It would be much nicer to simply spill the value into a variable the moment we know it will be indexed dynamically, and then always use it from there.

sagudev:
More exact link to comment: https://github.com/gfx-rs/wgpu/pull/6188#discussion_r1746348184

PR #6362, opened 2024-10-03
[naga] Support dynamic indexing of by-value matrices

### LINKED ISSUE #4337: Support dynamic indexing of by-value matrices and arrays, per WGSL

sagudev:
**Connections**
Fix #4337

**Description**
Add support for dynamic indexing of by-value matrices, by introducing `get_local_pointer_id` and then follow #6188.

**Testing**
New tests added and because there was no expectation changes in servo's CTS run I wrote new test: https://github.com/gpuweb/cts/pull/3982 (which passes with this PR).

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/naga/src/back/spv/block.rs
+++ b/naga/src/back/spv/block.rs
@@ -348,14 +348,7 @@ impl<'w> BlockContext<'w> {
                     } => {
                         let index_id = self.cached[index];
                         let base_id = self.cached[base];
-                        let base_ty = match self.fun_info[base].ty {
-                            TypeResolution::Handle(handle) => handle,
-                            TypeResolution::Value(_) => {
-                                return Err(Error::Validation(
-                                    "Array types should always be in the arena",
-                                ))
-                            }
-                        };
+                        let base_ty = &self.fun_info[base].ty;
                         let (id, variable) = self.writer.promote_access_expression_to_variable(
                             &self.ir_module.types,
                             result_type_id,
@@ -371,14 +364,7 @@ impl<'w> BlockContext<'w> {
                     crate::TypeInner::Matrix { scalar, rows, .. } => {
                         let index_id = self.cached[index];
                         let base_id = self.cached[base];
-                        let base_ty = match self.fun_info[base].ty {
-                            TypeResolution::Handle(handle) => handle,
-                            TypeResolution::Value(_) => {
-                                return Err(Error::Validation(
-                                    "Array types should always be in the arena",
-                                ))
-                            }
-                        };
+                        let base_ty = &self.fun_info[base].ty;
                         let (id, variable) = self.writer.promote_access_expression_to_variable(
                             &self.ir_module.types,
                             result_type_id,
--- a/naga/src/back/spv/block.rs
+++ b/naga/src/back/spv/block.rs
@@ -362,13 +362,39 @@ impl<'w> BlockContext<'w> {
                             base_id,
                             base_ty,
                             index_id,
-                            ty_element,
+                            LookupType::Handle(ty_element),
                             block,
                         )?;
                         self.function.internal_variables.push(variable);
                         id
                     }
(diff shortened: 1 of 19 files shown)


sagudev:
Because there was no expectation changes in CTS I wrote new test: https://github.com/gpuweb/cts/pull/3982

naga/src/back/spv/writer.rs
```diff
@@ -169,8 +169,14 @@ impl Writer {
...
-        let element_pointer_type_id =
-            self.get_pointer_id(ir_types, element_ty, spirv::StorageClass::Function)?;
+        let element_pointer_type_id = match element_ty {
+            LookupType::Handle(handle) => {
+                self.get_pointer_id(ir_types, handle, spirv::StorageClass::Function)?
+            }
+            LookupType::Local(local_type) => {
+                self.get_local_pointer_id(local_type, spirv::StorageClass::Function)
+            }
+        };
```
sagudev: Should we pack this in one single `get_pointer_id` method?

BTW `get_pointer_id` never returns Err.

teoxoy: I think we should update `get_pointer_id` and address the TODO that's currently in `get_local_pointer_id`.

sagudev: > address the TODO that's currently in get_local_pointer_id

This is hard as we cannot pack LookupType in pointer (due to infinite struct size), except if we box it. Another alternative is to simply put u32 for local type, but this could get pretty chaotic.

I think boxing solution is the most nice one. What do you think?

naga/src/back/spv/block.rs
```diff
@@ -362,21 +362,47 @@ impl<'w> BlockContext<'w> {
...
+                        id
+                    }
+                    crate::TypeInner::Matrix { scalar, rows, .. } => {
+                        let index_id = self.cached[index];
+                        let base_id = self.cached[base];
+                        let base_ty = match self.fun_info[base].ty {
+                            TypeResolution::Handle(handle) => handle,
+                            TypeResolution::Value(_) => {
+                                return Err(Error::Validation(
+                                    "Array types should always be in the arena",
```
teoxoy:
```suggestion
                                    "Matrix types should always be in the arena",
```

I don't think this is always true:

https://github.com/gfx-rs/wgpu/blob/43cb730d588a06ae64d91ce98afbef09d5613596/naga/src/proc/typifier.rs#L98-L98

teoxoy (changes requested):

naga/src/back/spv/writer.rs
teoxoy: What I hand in mind is adding a new `ValuePointer` variant. Similar to:

https://github.com/gfx-rs/wgpu/blob/43cb730d588a06ae64d91ce98afbef09d5613596/naga/src/lib.rs#L751

naga/src/back/spv/block.rs
sagudev: This was copy&paste from array, I will check and fix.

naga/src/back/spv/writer.rs
sagudev: LocalType has pointers encoded in LocalType::Value field.

jimblandy:
I think actually just fixing #6358 for both matrices and arrays might be simpler than the combination of this and #6188. I'm taking a shot at it now, but if it's hard, then we'll take this PR instead.

sagudev:
> I think actually just fixing #6358 for both matrices and arrays might be simpler than the combination of this and #6188. I'm taking a shot at it now, but if it's hard, then we'll take this PR instead.

Sounds good. Ping me on PR, so I will see it (and test it with servo against CTS).

jimblandy:
Closing, since #6390 has landed.

PR #6364, opened 2024-10-03
[wgsl-in,ir] Add support for parsing rust-style doc comments

ThierryBerger:
**Connections**
None as far as I know.

**Description**
Generating automated documentation à la https://docs.rs is very useful, but currently difficult.

naga lacks information about comments to be too helpful, so this pull requests adds support to parsing those.

**Testing**
- [wgmath documentation](https://github.com/dimforge/wgmath.rs) uses https://github.com/jannik4/shader_docs/pull/1 to  add documentation strings to it, it's based on an [older version of this PR, based on wgpu 0.22 branch](https://github.com/Vrixyz/wgpu/tree/token-comment-0.22).
- A `cargo xtask test` run doesn't yield more errors than trunk.

**Status** 

- [x] Add a token `DocComment`
- [x] Add a token `DocCommentModule`
- [x] Rust syntax was chosen: https://github.com/gfx-rs/wgpu/pull/6364#issuecomment-2679268882
  - [x] Parse `///` comments
  - [x] Parse `/** */` comments
  - [x] Parse `//!` module comments
  - [x] Parse `/*! */` module comments
- [x] Add comments to `ast::Struct`
- Support comments for multiple items
  - [x] struct
    - [x] struct members
  - [x] var (global variables)
  - [x] consts
  - [x] functions
  - module level comment
    - [x] For maximum compatibility with naga_oil, this is implemented with its own syntax: they should start with `//!`.
      - because naga_oil injects the imported modules at the beginning of the file, and not necessarily where the import line was written exactly. That prompted that commit: 02caf3a6df0a98ed93d710080f7af377b4b022b2
  - [ ] alias ; :no_entry_sign: I consider this out of scope for this PR.
- [x] Add comments list in `naga::Module`
- [x] Added more test and adapted existing tests

**Checklist**

- [x] Run `cargo fmt`.


jimblandy:
This doesn't get the comments out of the front end, though, does it?

It seems to me it could be fine to just add a field like:

```rust
    doc_comments: Option<Box<DocComments>>,
```

to `naga::Module`, and then have something like

```rust
struct DocComments {
    types: FastIndexMap<Handle<Type>, Vec<Span>>,
    global_variables: FastIndexMap<Handle<GlobalVariable>, Vec<Span>>,
    ...
}
```

I think it would make sense for Naga to adopt Rust's `///` syntax as a WGSL extension.

ThierryBerger:
> This doesn't get the comments out of the front end, though, does it?

Correct, I'm looking into propagating them to the module now ; my first intuition was to add comments to `naga::TypeInner::Struct` and the likes ; but I fear for performance.

Your idea of a "flat" `DocComments` with an indexmap to Spans is interesting and probably a better idea! Thanks for suggesting. I'll go in that direction.

naga/src/lib.rs
```diff
@@ -2263,4 +2263,27 @@ pub struct Module {
     pub functions: Arena<Function>,
     /// Entry points.
     pub entry_points: Vec<EntryPoint>,
+    /// Comments, usually serving as documentation
+    pub comments: Comments,
```
jimblandy: I think this should be an `Option<Box<Comments>>`, so that users who aren't interested in collecting these comments don't need to spend memory on them.

naga/src/lib.rs
```diff
@@ -2263,4 +2263,27 @@ pub struct Module {
...
+#[cfg_attr(feature = "serialize", derive(Serialize))]
+#[cfg_attr(feature = "deserialize", derive(Deserialize))]
+#[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
+pub struct Comments {
+    pub types: FastIndexMap<Handle<Type>, Vec<String>>,
+    // The key is:
+    // - key.0: the handle to the Struct
+    // - key.1: the index of the `StructMember`.
+    pub struct_members: FastIndexMap<(Handle<Type>, usize), Vec<String>>,
+    pub functions: FastIndexMap<String, Vec<String>>,
```
jimblandy: We usually refer to functions by `Handle<Function>`.

However, for reasons that aren't clear to me, we refer to entry points by index; there's no such thing as a `Handle<EntryPoint>`. So you'll need a separate `FastIndexMap<usize, Vec<String>>` for the entry points.

naga/src/front/wgsl/tests.rs
jimblandy, line 1: It would be better to add this as a new snapshot test in `naga/tests/in`. You'll need to edit `naga/tests/snapshots.rs` and add the file to the list in `convert_wgsl`.

naga/src/lib.rs
```diff
@@ -2263,4 +2263,27 @@ pub struct Module {
...
+
+/// Comments preceding items.
+///
+/// These can be used to generate automated documentation,
+/// IDE hover information or translate shaders with their context comments.
+#[derive(Debug, Default, Clone)]
+#[cfg_attr(feature = "serialize", derive(Serialize))]
+#[cfg_attr(feature = "deserialize", derive(Deserialize))]
+#[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
+pub struct Comments {
```
jimblandy: Please put this above `struct Module`, so that remains the last thing in `lib.rs`.

naga/src/valid/handles.rs
```diff
@@ -39,6 +39,8 @@ impl super::Validator {
             ref types,
             ref special_types,
             ref global_expressions,
+            // TODO: validate comments (shouldn't have invalid handle or spans ?)
+            ..
```
jimblandy: Please do not add a `..` here. The reason we're doing this weird whole-struct match thing is exactly to ensure we're not skipping members, and adding `..` would defeat that.

Validation will need to iterate over the keys of the hash tables in `Comments` and ensure that they are all valid handles/indices into their respective arenas, by calling `Self::validate_foo_handle`. This validation could come last.

naga/src/front/wgsl/parse/lexer.rs
```diff
@@ -81,28 +83,45 @@ fn consume_token(input: &str, generic: bool) -> (Token<'_>, &str) {
             let og_chars = chars.as_str();
             match chars.next() {
                 Some('/') => {
-                    let _ = chars.position(is_comment_end);
-                    (Token::Trivia, chars.as_str())
+                    if let Some(end_position) = input
+                        .char_indices()
+                        .find(|char_indices| is_comment_end(char_indices.1))
+                    {
+                        let end_position = end_position.0 + 1;
```
jimblandy: If the comment is terminated by U+0085, U+2028, or U+2029, those characters have multi-byte encodings in UTF-8, so this `... + 1` will produce an invalid character boundary.

jimblandy (changes requested):
Some more comments here.

You'll also need to change `compact::compact` to adjust the type handles, even though only named types can have comments and compaction never removes named types, because compaction may remove anonymous types, and that renumbers the handles.

naga/src/front/wgsl/parse/lexer.rs
```diff
@@ -81,28 +83,45 @@ fn consume_token(input: &str, generic: bool) -> (Token<'_>, &str) {
...
+                        return (
+                            if chars.next() == Some('!') {
+                                Token::CommentModule(&input[..end_position])
+                            } else {
+                                Token::Comment(&input[..end_position])
+                            },
+                            &input[end_position..],
+                        );
+                    }
+                    (Token::Comment(input), &input[input.len() - 1..])
```
jimblandy: This doesn't look right. Doesn't it just return the last byte of `input` as the following text?

This code is only exercised by `front::wgsl::parse::lexer::test_comments`, which only uses a `//` at the end of the input, so it wouldn't notice if the rest of the text were truncated.

jimblandy (changes requested):

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2172,6 +2188,9 @@ impl Parser {
         lexer: &mut Lexer<'a>,
         out: &mut ast::TranslationUnit<'a>,
     ) -> Result<(), Error<'a>> {
+        // Save a lexer to be able to backtrack comments if need be.
+        let mut lexer_comments = lexer.clone();
```
jimblandy: This approach means that we end up re-scanning doc comments unnecessarily. Instead, you should use the `peek` functions to decide when the doc comment ends.

jimblandy (changes requested):

jimblandy:
> Note: I branched off 0.22 because I'm not confident with jumping on the maybe unstable trunk.

It should be fine to work on `trunk`. This PR should be rebased on that.

jimblandy:
@Vrixyz Since this isn't passing CI, to make it easier to track stuff I need to review, I'm going to convert this to a draft PR until those issues are sorted out. Feel free to flip it back into non-draft when you're ready.

naga/src/front/wgsl/parse/mod.rs
ThierryBerger: I'm not sure about "peeking" being better..

First, peeking is implemented by "cloning" the `Lexer` and call `. next()` on it. So by only cloning, I avoid the `next()` which I might not need.

- An alternative would be to build a list of comments eagerly, but maybe dismiss them if the next item doesn't support them, that would be quite bad :-1: 
- A better alternative would be to peek for next item and only build that comments list when next item support them, that's better, but now the following lines of searching for next item are useless because we already know when the next item starts, so that would end up in some refactoring, for I believe no meaningful performance changes.


<details><summary>More reasoning I had before thinking more about it</summary>
<p>

doc comments are actually bypassed because of `start_byte_offset`, which I updated to ignore doc comments.

**edit: "bypassed" is not really correct, they are still scanned, and I believe that's what Jim was commenting on)**

I updated it to ignore doc comments to avoid side effect breakage, so its behavior is the same as before.

I added a `start_byte_offset_and_aggregate_comment` which is it's counterpart, but aggregates the comments. It is only called when the item encountered can support a doc comment.

while I agree this looks a bit convoluted, scanning comments there would infer to build up a list of comments which we're not sure to use: i.e: In case comments are in the end of the struct, we don't really want to parse them because we wouldn't know where to store them.

We could also only peek next attribute to see if it can support comment, then parse comments, but the end result would be similar. We could avoid a `. clone()`, but a `Lexer.clone()` is quite cheap and the changeset needed to make that work scares me.

</p>
</details> 

I'd like to keep it as a follow up.

Let me know if performance analysis could help with the decision.

naga/src/lib.rs
ThierryBerger: I did the change ; we could also add a feature for that ?

naga/src/front/wgsl/parse/mod.rs
ThierryBerger: Current implementation is wrong though, `validation::validation_error_messages` seems to infer that span is incorrect

ThierryBerger:
I rebased on trunk (github comment reviews are degraded, but I adressed those :crossed_fingers:), tests are encouraging, there's a major bug I'm investigating still (*validation_error_messages*): I think spans are incorrect, which I hope to fix by working on [Jim's feedback](https://github.com/gfx-rs/wgpu/pull/6364/files#r1906174647).

:question: I'm curious about "pushing rules": I imagine parsing should contain a rule for parsing comments ? so we push that rule when checking for comments.

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2557,9 +2575,23 @@ impl Parser {
...
+            let mut module_comments: Vec<_> = module_comments
+                .into_iter()
+                .map(|comment_span| lexer.source.index(comment_span))
+                .collect();
+            out.comments.append(&mut module_comments);
+        }
+        let token_span = lexer.next();
+
         // read item
         let start = lexer.start_byte_offset();
```
ThierryBerger: `start` should be computed before calling `lexer.next();`.

naga/src/front/wgsl/parse/mod.rs
jimblandy: The idea is, it should be cleaner to look at the next token to decide when the series of doc comments ends. This is the way the parser hands other kinds of lists that are terminated by something that has meaning beyond simply being a terminator.

I doubt it has much of an observable performance impact. But I feel that --- to a limited extent --- code that does redundant work is more confusing, because the reader will not generally expecting the redundancy. For example, they might be surprised by the way the state is rewound many tokens before the current point of attention. But I concede that arguments based on "what people expect" are hard to judge; if one isn't persuaded, one simply isn't persuaded.

Implementing `peek` by cloning the `Lexer` made more sense when the `Lexer` was nothing more than two `&str` fields. But it's starting to get a little heavier, and I think that implementation might not be ideal any more. Many parsers use a peek operation, but it's usually done by having an `Option<Token>` field somewhere that you just save the next token in.

naga/src/valid/handles.rs
```diff
@@ -254,6 +255,76 @@ impl super::Validator {
...
+                                    || "members length incorrect".to_string(),
+                                    |name| name.to_string(),
+                                ),
+                                source: TypeError::InvalidData(comment_struct_member_doc.0 .0),
+                            })?;
+                    }
+                    crate::TypeInner::Scalar(_) => {
+                        // TODO: debug why this can happen, if we systematically add all couples "Type/index" for struct members.
+                        // This has been mitigated by not adding them when there is no comments, but it's probably not enough.
+                    }
```
Vrixyz, line 297: This

naga/src/lib.rs
ThierryBerger: Thanks, I did the change, that's better this way!

jimblandy:
> ❓ I'm curious about "pushing rules": I imagine parsing should contain a rule for parsing comments ? so we push that rule when checking for comments.

You mean `Parser::rules`? You shouldn't need to worry about that at all. You should just call `start_byte_offset` and `span_from` directly.

The rule stack is currently being used as a rough approximation to WGSL's [template list discovery](https://gpuweb.github.io/gpuweb/wgsl/#template-list-discovery), which you definitely do not want to get tangled up with, and which shouldn't affect your work at all. And it's used for some sanity checks.

jimblandy:
I rebased this on trunk (7e42040fa, 2025-2-19).

naga/src/front/wgsl/parse/mod.rs
jimblandy: Okay, I think I understand a little better now what the original intention was. Under the assumption that cloning a lexer is cheap, you're simply *noting* where doc comments *would* begin, and then only actually recording them in the module if you later decide that you've started to parses something that can accept doc comments. This makes more sense.

naga/src/lib.rs
jimblandy: Maybe I didn't really say what I meant:

I think users should be able to choose to collect or ignore comments at run time. That is, `naga::front::wgsl::FrontEnd::new` should take an `Options` type, or `FrontEnd` should have a `new_with_comments` constructor - or whatever, something boring but tasteful - and then parsing should return a `Module` whose `comments` field is either populated or not.

naga/src/lib.rs
```diff
@@ -2476,4 +2498,13 @@ pub struct Module {
...
     pub diagnostic_filter_leaf: Option<Handle<DiagnosticFilterNode>>,
+    /// Comments, usually serving as documentation
+    pub comments: Option<Box<Comments>>,
+}
+
+impl Module {
+    pub fn get_comments_or_insert_default(&mut self) -> &mut Box<Comments> {
+        self.comments
+            .get_or_insert_with(|| Box::new(Comments::default()))
+    }
```
jimblandy: It's valuable to frequent contributors to keep `Module`'s definition as the last thing in this file. I think once you make comment collection selectable at runtime this is going to change anyway, but if it sticks around let's put it someplace else -  `naga/src/front/mod.rs`, perhaps.

naga/src/compact/mod.rs
```diff
@@ -264,6 +264,65 @@ pub fn compact(module: &mut crate::Module) {
...
+            module: _,
+            types: ref mut comment_types,
+            struct_members: ref mut comment_struct_members,
+            entry_points: _,
+            functions: _,
+            constants: ref mut comment_constants,
+            global_variables: _,
+        } = **comments;
+        log::trace!("adjusting comments for types");
+        for comment_type_handle in comment_types.keys().cloned().collect::<Vec<_>>() {
```
jimblandy: I think in general we should be able to `std::mem::take` the `IndexMap`, and then do an owning iteration (just `for (key, value) in map { ... }`) and check, adjust, and reinsert the values. This should be simpler than `swap_remove`.

jimblandy (dismissed):
This is getting there.

- Can we record comments as Spans instead of Strings?

- Users should be able to choose at runtime whether to collect comments or not, as I explain here: https://github.com/gfx-rs/wgpu/pull/6364#discussion_r1964353109

- The CI failures will need to be addressed. I saw some Clippy style concerns.

naga/src/compact/mod.rs
ThierryBerger: Thanks! I adressed this in 684e7e46c

naga/src/lib.rs
ThierryBerger: Thanks ! I addressed this in 57865be9495fbcfaf0f3969309253b83afdfdb4e ; I'm open to suggestions about the implementation.

I updated a few tests to test both options.

naga/src/lib.rs
ThierryBerger: Ok, I'll trust your judgement here: 953d8fb49741479a7f39922c2e33cd0a4479b5fd

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -261,6 +261,21 @@ impl<'a> BindingParser<'a> {
     }
 }
 
+/// Configuration for the whole parser run.
+pub struct ParserOptions {
+    /// Controls whether the parser should parse comments.
+    pub parse_comments: bool,
+}
+
+impl ParserOptions {
```
jimblandy: This should just be named `Options`; the fact that it's from `naga::front::wgsl` makes it clear that it contains parsing options.

Rather than having a `new` method, `Options` structs should implement `Default`.

ThierryBerger: Good point about prefix being redundant 👍

I can implement default, but it's not const so this was complaining from an existing caller.

- Seems relevant: https://github.com/rust-lang/rust/issues/67792

ThierryBerger:
>     * Can we record comments as Spans instead of Strings?

I don't think we can or should.

I consider comments quite similare to [`Ident`](https://github.com/Vrixyz/wgpu/blob/d0299b65c3c72d2b18211e31ef02c3c9b738c736/naga/src/front/wgsl/parse/ast.rs#L48), where it could be any values, currently `Ident` is stored with a `name: &str` and a `Span`. Retrieving that `&str` from the original source + `Span` doesn't seem a used practice (and I'm not sure even feasible), so I opted for keeping it consistent.

What I could do is add the `Span` alongside the comment contents. I'm not sure what is the advantage of that though: I had the impression that that span was used mostly for error reporting, which would not occur on comment parsing (?), or mayyybe an internal error around https://github.com/Vrixyz/wgpu/blob/d0299b65c3c72d2b18211e31ef02c3c9b738c736/naga/src/valid/handles.rs#L284.

Another use case for a span for comments would be to make a tool to parse a wgsl then remove comments only, or link to comment source.

Let me know if "adding" that span (opposed to replacing `&str`)  in the comments is needed to make this PR progress.

If I'm wrong, I'm happy to be corrected, thanks again for your diligent reviews!

jimblandy:
@Vrixyz In Rust, only comments starting with `///` or `/**` are considered to be doc comments, and others are just ordinary comments. The syntax you've chosen doesn't make any distinction between the two. What are your thoughts about Rust's design? Would it be appropriate for WGSL?

ThierryBerger:
> What are your thoughts about Rust's design? Would it be appropriate for WGSL?

@jimblandy personally I'd advocate for mimicking Rust's syntax rules, and I'd be ok with implementing it :+1: (i.e ignore comments not starting by `///` or `/**`).

But I chose to parse every comment because [wgsl spec](https://www.w3.org/TR/WGSL/#comments) doesn't acknowledge "comment docs", as all comments are considered documentation.

That being said, I didn't quite follow my objective, because I didn't succeed in implementing module comments without adding a specific syntax: `//!`, which is similar to rust's syntax.

I made this decision because it made things simpler when working with `naga_oil` when multiple files are stitched together.

jimblandy:
@Vrixyz Let's go with the Rust syntax. Can we take care of it in this PR?

jimblandy:
> > ```
> > * Can we record comments as Spans instead of Strings?
> > ```
> 
> I don't think we can or should.

Okay, this is fine. We can revisit this question in the future if necessary.

naga/src/front/wgsl/parse/mod.rs
jimblandy: Sometimes we've marked functions as `const` just because Clippy complained, and not because we actually wanted to use that function in a constant context. So if the fix is simply removing `const` from `const fn`s, that's probably fine.

ThierryBerger: Thanks for that information.

I don't think `const` was added without reason here, it goes back to https://github.com/gfx-rs/wgpu/commit/6035b07b786135f65cb4140f99cf67cc0381b4c3#diff-46894d640690ef35dff3deaf0bb5416be6e4d00eaba890e3181628d769f6f413R1735 ; and the commit message has a `Fixes https://github.com/gfx-rs/wgpu/pull/2071: Global item does not support 'const'` (the linked issue doesn't seem quite relevant though, not sure).

I'd prefer to keep this PR as close to existing code as possible.

ThierryBerger:
> Let's go with the Rust syntax. Can we take care of it in this PR?

I did that in [9cb6d53](https://github.com/gfx-rs/wgpu/pull/6364/commits/9cb6d53334c0eee160feb2b20bf38d55f47fcb0b) 👍

naga/src/front/wgsl/parse/lexer.rs
```diff
@@ -270,7 +321,29 @@ impl<'a> Lexer<'a> {
...
+            // Eat all trivia because `next` doesn't eat trailing trivia.
+            let (token, rest) = consume_token(self.input, false, self.save_comments);
+            if let Token::CommentDoc(_) = token {
+                self.input = rest;
+                let next = self.current_byte_offset();
+                comments.push(Span::new(start as u32, next as u32));
+            } else if let Token::Trivia = token {
+                self.input = rest;
+            } else if let Token::CommentDocModule(_) = token {
+                self.input = rest;
```
teoxoy: Should it be valid to have these items in-between the doc comment and the item itself? It seems odd to me.

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -1211,12 +1232,21 @@ impl Parser {
             let ty = self.type_decl(lexer, ctx)?;
             ready = lexer.skip(Token::Separator(','));
 
+            let mut comments = Vec::new();
+            lexer_comments.start_byte_offset_and_aggregate_comment(&mut comments);
+
+            let comments = comments
+                .into_iter()
+                .map(|comment_span| lexer.source.index(comment_span))
+                .collect();
```
teoxoy: We don't seem to ever use the return type of `start_byte_offset_and_aggregate_comment`, could you rename it to something like `accumulate_doc_comments` that returns the `Vec<&str>` directly?

naga/src/front/wgsl/parse/lexer.rs
```diff
@@ -82,20 +83,64 @@ fn consume_token(input: &str, generic: bool) -> (Token<'_>, &str) {
...
+                    let mut char_indices = input.char_indices();
+
+                    // Skip '/' and '*'
+                    char_indices.next();
+                    char_indices.next();
+
+                    let mut constructing_token = if !save_comments {
+                        Token::Trivia
+                    } else {
+                        let mut peeker = char_indices.clone().peekable();
```
teoxoy: We can make `char_indices` `peekable`, avoiding the `clone`.

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -1182,6 +1200,9 @@ impl Parser {
                     ExpectedToken::Token(Token::Separator(',')),
                 ));
             }
+            // Save a lexer to be able to backtrack comments if need be.
+            let mut lexer_comments = lexer.clone();
```
teoxoy: We can then just call `accumulate_doc_comments` here instead.

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2762,14 +2839,34 @@ impl Parser {
...
+            // Parse module comments.
+            let mut comments = Vec::new();
+
+            fn peek_any_next<'a>(lexer: &'a Lexer) -> (Token<'a>, Span) {
+                let mut cloned = lexer.clone();
+                let token = cloned.next_until(|_| true, false);
+                token
+            }
+            while let (Token::CommentDocModule(_), span) = peek_any_next(&lexer) {
+                comments.push(lexer.source.index(span));
```
teoxoy: Doesn't `Token::CommentDocModule(_)` already contain the `&str`?

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2762,14 +2839,34 @@ impl Parser {
...
+        if options.parse_comments {
+            // Parse module comments.
+            let mut comments = Vec::new();
+
+            fn peek_any_next<'a>(lexer: &'a Lexer) -> (Token<'a>, Span) {
+                let mut cloned = lexer.clone();
+                let token = cloned.next_until(|_| true, false);
+                token
+            }
+            while let (Token::CommentDocModule(_), span) = peek_any_next(&lexer) {
```
teoxoy: We could add a `next_if` method on the lexer for this, making the code easier to look at.

naga/src/front/wgsl/lower/mod.rs
```diff
@@ -1065,6 +1069,29 @@ impl<'source, 'temp> Lowerer<'source, 'temp> {
...
+                                        f.comments.iter().map(|s| s.to_string()).collect(),
+                                    );
+                            }
+                            LoweredGlobalDecl::EntryPoint(index) => {
+                                ctx.module
+                                    .get_comments_or_insert_default()
+                                    .entry_points
+                                    .insert(
+                                        index,
+                                        f.comments.iter().map(|s| s.to_string()).collect(),
```
teoxoy: I'm not sure how the doc comments will be consumed but wouldn't it make sense if they didn't contain the `///`, `//!` (and so on) inside? We could then concatenate them so that each item has a single `String` as its doc comment.

teoxoy (changes requested):

teoxoy:
The overall approach looks good, I mainly left some comments that I think would improve the impl and a one last question about usage.

naga/src/front/wgsl/parse/mod.rs
ThierryBerger: The reason I went this route was because not all items should support comments, so if we try to accumulate on everything, there'd be a `Vec` instantiated to store the result, then we realize we don't need it, so we drop it. I suspect this would impact performance, but I didn't measure it.

naga/src/front/wgsl/parse/mod.rs
ThierryBerger: Correct! as I'm peeking I can't use it though because the borrow checker is unhappy, I think I have to make a `next_module_comment` function for the lexer.

naga/src/front/wgsl/parse/mod.rs
teoxoy: Empty `Vec`s don't allocate so I don't think it should be a problem.

naga/src/front/wgsl/parse/mod.rs
ThierryBerger: Actually, it's not 100% trivial, changing this would either:

- require a `peek` before, then getting the comment if relevant.
- or use an `accumulate_comments` for module docs, then backtracking to get the last token which resulted in leaving that `accumulate_comments`.
- change `accumulate_comment` implementation to return the last `(token, rest)`.

naga/src/front/wgsl/parse/mod.rs
ThierryBerger: `accumulate_doc_comments` leaves the lexer in a "bad" state though, where it consumes 1 item "too many" in order to know it's stopped. I could change this behaviour to `peek` before consuming, or return the latest consumed item (which is not a comment) for further parsing :thinking:

naga/src/front/wgsl/parse/lexer.rs
ThierryBerger: `char_indices` is actually an iterator (from rustlib), so the clone is not expensive. We could make the api more ergonomic but ultimately I think that would lead to the same generated code.

naga/src/front/wgsl/parse/lexer.rs
ThierryBerger: you're probably right, I thought it wouldn't hurt to not fail, but rust does indeed raise an error if a module comment is at the wrong place. I can do that.

naga/src/front/wgsl/parse/mod.rs
ThierryBerger: thanks, done

naga/src/front/wgsl/lower/mod.rs
ThierryBerger: I have no strong feelings about it, but I took the most "conservative" and unopiniated approach.

Some comments look like 
```rs
////////
// SOME HEADING
////////
```

or sometimes ascii art is in comments, and messing up with what user wrote may be surprising, so I kept it.

Oh and also, there are edge cases between `// comment` and `//comment` which could lead to unwanted space at the beginning, or we should take the most indented... cases add up fast and start to be opinionated quickly.

naga/src/front/wgsl/parse/lexer.rs
teoxoy: What about `Token::Trivia` (comments) does rust allow those in-between doc comments?

naga/src/front/wgsl/parse/mod.rs
teoxoy: `accumulate_doc_item_comments` doesn't set `self.input = rest` when it encounters a token it can't consume and so it won't advance the tokenizer.

naga/src/front/wgsl/lower/mod.rs
teoxoy: Ok, we can change this later if needed.

naga/src/front/wgsl/parse/lexer.rs
teoxoy: That's fair, I should have mentioned my main concern was about readability as it took me a while to figure out what the code was doing, though I realize making `char_indices` `peekable` doesn't improve the flow much.

teoxoy: Actually, we are already cloning the iterator, we shouldn't need to make it peekable at all, we can just call next on it.

naga/src/front/wgsl/parse/mod.rs
teoxoy: A similar method to `accumulate_doc_item_comments` would work and we wouldn't need to backtrack.

teoxoy:
I squashed the commits (since there were too many commits to rebase individually), rebased the PR and made some more modifications.

@Vrixyz let me know if the PR looks good, if so I think we can land it

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2750,6 +2771,8 @@ impl Parser {
         lexer: &mut Lexer<'a>,
         out: &mut ast::TranslationUnit<'a>,
     ) -> Result<'a, ()> {
+        let doc_comments = lexer.accumulate_doc_comments();
```
ThierryBerger, line 2774: This approach is interesting, as we could use it to emit warning when doc comments are encountered when it shouldn't have been: check that `doc_comments` is empty for items that don't support it (separator, word...).

I'd consider this for another PR though, as it will lead to questions about error reporting, etc...

ThierryBerger:
Thanks @teoxoy for championing it! Looks great to me.

naga/src/front/wgsl/parse/mod.rs
teoxoy: Yeah, it would be nice to warn/error in these cases, through we need to make sure to only do so based on a setting or else this would make naga not spec compliant.

teoxoy: I guess we could just do it based on `options.parse_doc_comments`.

teoxoy (approved):

jimblandy:
@ThierryBerger I'm just curious - what is this math? It appears in a test case for the doc comment parsing in `naga/src/front/wgsl/parse/lexer.rs`:

```
#[test]
fn test_doc_comment_long_character() {
    sub_test_with_and_without_doc_comments(
        "/// π/2
        ///     D(𝐡) = ───────────────────────────────────────────────────
///            παₜα_b((𝐡 ⋅ 𝐭)² / αₜ²) + (𝐡 ⋅ 𝐛)² / α_b² +`
    const a : i32 = 2;",
```

ThierryBerger:
@jimblandy most likely coming from bevy, such as https://github.com/bevyengine/bevy/blob/main/crates/bevy_pbr/src/render/pbr_lighting.wgsl#L154 ; they are isolated snippets which caused ~~tests~~ lexer to fail at some point where I was probably incorrectly handling characters with multiple bytes.

For the exact meaning of the math I won't be able to explain it but hopefully the context is enough breadcrumbs to follow :)

PR #6371, opened 2024-10-05
Add utils for WebGPU support detection & Instance creation

### LINKED ISSUE #6166: webgl fallback fails in some browsers.
**Description**
My understanding is that when the `webgpu` and `webgl` cargo features are enabled and `Backends::all()` is set, the adapter should attempt to be created with webgpu and then failing that fallback to webgl.
However that doesnt seem to be happening in chrome on my machine (linux) and chrome on a users machine (windows 10)
It does however seem to work fine on firefox on my machine.

This issue is not present in the wgpu examples, but I cant figure out what is different between the example and my (isolated) code

**Repro steps**
```
git clone -b fail_to_create_adapter_on_chrome https://github.com/rukai/wgpu_bug_report
cd wgpu_bug_report
cargo run-wasm --package wgpu_foo
```

I expect the project to work on chrome on my machine, it does not.

**Extra materials**
Fails on chrome:
image
Succeeds on firefox:
image

Wumpf:
**Connections**
* Fixes  #6166

**Description**
Targeting both WebGPU & WebGL at the same time works but is a pain right now. These new utilities and doc lines should make this a lot easier!

Would be nice to make use of this in the official examples, but don't quite know yet how - the examples on web today are geared towards explicit selection.

**Testing**

Testing done via https://github.com/Wumpf/minifb_wgpu_web_and_desktop/tree/better-webgpu-selection

* [x] test on native (should no-op)
* [x] test browser without `navigator.gpu`
* [x] test browser with `navigator.gpu` no actual webgpu support (Chrome on Linux!)
* [x] test on browser with webgpu support

**Checklist**

- [x] Run `cargo fmt`.


Wumpf:
@rukai can you check if this solves the problem for you? I don't have a good repro setup for it right now
Either by using the new `new_instance_with_webgpu_detection` method yourself or by checking with running `cargo xtask run-wasm` on https://github.com/Wumpf/minifb_wgpu_web_and_desktop/tree/better-webgpu-selection
Thank you!

wgpu/src/backend/webgpu.rs
```diff
@@ -1088,8 +1088,12 @@ pub struct BrowserGpuPropertyInaccessible;
 /// Returns the browser's gpu object or `Err(BrowserGpuPropertyInaccessible)` if
 /// the current context is neither the main thread nor a dedicated worker.
 ///
-/// If WebGPU is not supported, the Gpu property is `undefined`, and so this
-/// function will return `Ok(None)`.
+/// If WebGPU is not supported, the Gpu property may (!) be `undefined`,
+/// and so this function will return `Ok(None)` function will return `Ok(None)`.
```
teoxoy:
```suggestion
/// and so this function will return `Ok(None)`.
```

teoxoy (approved):
Much clearer than before!

Wumpf:
Let's hold off with merging: Tried to test this myself now in Chromium on Wsl which _seems_ to exhibit the issue, but I wasn't able to demonstrate it being fixed with using the new utility here

Wumpf:
phew! Figured it out quite quickly, [`bfd7879`](https://github.com/gfx-rs/wgpu/pull/6371/commits/bfd7879cbe3beff3741b7f0b42572be38d92ddd1) fixes it. Very happy I checked

rukai:
I can confirm this successfully fixes webgl fallback on linux+chrome.
Thankyou so much for this.

PR #6366, opened 2024-10-04
Enforce formatting of TOML with `taplo`

ErichDonGubler:
**Connections**

\-

**Description**

Title says all.

**Testing**

CI!

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `cargo clippy`. If applicable, add:
  - [x] `--target wasm32-unknown-unknown`
  - [x] `--target wasm32-unknown-emscripten`
- [x] Run `cargo xtask test` to run tests.
- [x] ~~Add change to `CHANGELOG.md`. See simple instructions inside file.~~ Not necessary.

PR #6376, opened 2024-10-07
Keep TOML fields alphabetized

cwfitzgerald:
#6366 but with config

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,22 @@
 [workspace]
-resolver = "2"
+default-members = [
+    "benches",
+    "examples",
+    "lock-analyzer",
+    "naga",
+    "naga-cli",
+    "naga/fuzz",
+    "naga/hlsl-snapshots",
+    "player",
+    "tests",
+    "wgpu",
+    "wgpu-core",
+    "wgpu-hal",
+    "wgpu-info",
+    "wgpu-macros",
+    "wgpu-types",
+]
+exclude = []
 members = [
     "cts_runner",
     "deno_webgpu",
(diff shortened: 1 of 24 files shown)


taplo.toml
```diff
@@ -0,0 +1,6 @@
+[formatting]
+align_comments = true
```
ErichDonGubler: question: This is already the default. Is there a reason you want to specify it explicitly?

cwfitzgerald: Dunno :) maybe I liked seeing both align options

benches/Cargo.toml
```diff
@@ -1,19 +1,19 @@
 [package]
-name = "wgpu-benchmark"
-version.workspace = true
```
teoxoy, line 3: It's odd seeing those ordered alphabetically. I think usually people expect to see name, version, then everything else.

Cargo.toml
```diff
@@ -1,54 +1,54 @@
 [workspace]
-resolver = "2"
-members = [
```
teoxoy, line 3: Same here.

ErichDonGubler: I think this particular case can we worked around by putting another newline in between `resolver` and the subsequent fields?

cwfitzgerald:
This isn't landing as is

ErichDonGubler:
@cwfitzgerald: I'm not sure on what basis you're claiming that, but from my perspective, we could work this out with consensus building. Maybe talk about this in the next maintainer's meeting? 🤔

cwfitzgerald:
Didn't mean as anything more than "this PR needs to be rehashed before it can land, and I don't necessarily am the one to do it" 😅 - purely administrative

Issue #6290, opened 2024-09-17
Remove `SupportedLimits.maxInterStageShaderComponents`


ErichDonGubler:
This was removed from the WebGPU spec. in <https://github.com/gpuweb/gpuweb/pull/4783>, and we should, too!

- [x] Remove from WebGPU-on-web backend:
	- #6377
	- #6380
- [x] Remove validation and `wgpu_types::SupportedLimits::maxInterStageShaderComponents`: #8652

ErichDonGubler:
Firefox downstream tracking: [bug 1919404](https://bugzilla.mozilla.org/show_bug.cgi?id=1919404)

codeart1st:
My test already failing in firefox: https://github.com/codeart1st/wgpu-layers/actions/runs/11090086957/job/30811980649

```
---- render_test::empty output ----
    info output:
        surface: Surface { context: ContextWebGpu { type: "Web" }, _handle_source: "None", id: ObjectId { id: Some(1), global_id: Some(1) }, data: Any { .. }, config: Mutex { data: None } }
        adapter: Adapter { context: ContextWebGpu { type: "Web" }, id: ObjectId { id: Some(1), global_id: Some(2) }, data: Any { .. } }
    
    error output:
        panicked at /home/runner/work/wgpu-layers/wgpu-layers/src/renderer.rs:77:8:
        Device can't be created.: RequestDeviceError { inner: WebGpu(JsValue(OperationError: requestDevice: Limit 'maxInterStageShaderComponents' not recognized.
        __wbg_get_imports/imports.wbg.__wbg_requestDevice_1c8e4f0fe8729328/<@http://127.0.0.1:45803/wasm-bindgen-test:1401:37
```

ErichDonGubler:
@codeart1st: I've removed the exposed limit in Firefox using the above link, yes! That's odd, though, because if we were failing in Firefox, then surely that means we were also failing in Chrome, no? 🤔

codeart1st:
The integrationtest is using https://download-installer.cdn.mozilla.net/pub/firefox/nightly/latest-mozilla-central/firefox-132.0a1.en-US.linux-x86_64.tar.bz2 - chrome is a different story and currently not possible to run headless integrationtests on it under linux, so I don't know for now 😄
The last working run was 4 days ago: https://github.com/codeart1st/wgpu-layers/actions/runs/11060882848/job/30732372163

ErichDonGubler:
[...]

PR #6377, opened 2024-10-07
fix(web)!: remove `maxInterStageShaderComponents` from vendored `web-sys`

ErichDonGubler:
**Connections**

Works around #6290. CC @daxpedda, who will likely be fielding a review in which I upstream this to `web-sys` eventually.

**Description**

\-

**Testing**

Tested locally that we are successfully working around the problem. Can be tested with `cargo xtask run-wasm` and opening the `hello_compute` example.

**Checklist**

- [ ] Run `cargo fmt`.


teoxoy (approved):

jimblandy (approved):
If I understand correctly, the reason we vendor web-sys is for exactly this kind of situation. So this looks good to me.

PR #5751, opened 2024-05-28
Support `texture-compression-bc-sliced-3d` in wgpu

mehmetoguzderin:
Fixes https://github.com/gfx-rs/wgpu/issues/5750

Since it has been a while since I created a PR for wgpu, I might have overlooked some aspects. Your guidance would be appreciated. Thank you!

PR #6381, opened 2024-10-08
Document rationale for vendoring WebGPU bindings.

jimblandy:
This commit is almost entirely autogenerated changes.


Wumpf (approved):
nice writedown, thank you!

daxpedda (commented):
You may want to mention that this also serves as a workaround for `--cfg=web_sys_unstable_apis`, which would be required when depending on unstable APIs in `web-sys`.

jimblandy:
It turns out that #5751 manually patched the vendored files, so my re-vendoring broke uses of `GpuFeatureName::TextureCompressionBcSliced3d`. I'm going to patch it back in manually to let this PR land, and then file a follow-up issue to remove it.

@mehmetoguzderin @teoxoy We can't generally have local changes to the files in `wgpu/src/backend/webgpu/webgpu_sys`, unfortunately.

Issue #4330, opened 2021-05-05
[msl-out] Decide whether we're going to handle `MathFunction::Inverse` for matrices 


expenses:
MSL doesn't natively have any functions for computing matrix inversions. `spirv-cross` handles this by adding functions that compute the inverse and calling those: https://github.com/KhronosGroup/SPIRV-Cross/blob/72a2ec4c1b56ce233e0da97a36f87af98927256c/spirv_msl.cpp#L4941-L5079.

The code for these functions is quite long, and I think it's worth thinking about whether we even want to handle this case in the first place.

Xaeroxe:
I'm running into this issue. Personally I'd have preferred it if the MSL output included the needed function definition.

```
panicked at 'wgpu error: Validation Error

Caused by:
    In Device::create_render_pipeline
      note: label = `horizon_pipeline`
    Internal error in VERTEX shader: MSL: UnsupportedCall("Inverse")
 
```

PR #6382, opened 2024-10-08
[naga/wgsl-out] allow user-provided polyfills of `inverse` and `outer` 

chyyran:
**Connections**
* https://github.com/gfx-rs/wgpu/issues/4330
* https://github.com/gpuweb/gpuweb/issues/4115
* https://github.com/gfx-rs/naga/issues/893

**Description**
`inverse` and `outer` are not currently supported by WGSL and might never be. However this leaves shaders that do need these functions fundamentally incompatible with WGSL unless they get polyfilled.

This provides an escape hatch for the user to provide their own polyfill for this function. The user is responsible for appending valid WGSL code to the output source that implements these polyfills.

Validation is unaffected because validation treats `inverse` and `outer` as normal `MathFunction`s. Providing an invalid polyfill or not appending the implementations is user error.

If something like this is desireable to have in naga, I wouldn't mind looking into similar functionality for msl-out as well.

**Testing**
I'm not sure how to utilize the existing test harness to test this, but I'm using something like this right now. Proper tests should be added before moving forward with this but I'll need some pointers on how or where to write these tests.

```rust
fn write_wgsl(module: &Module, info: &ModuleInfo) -> Result<String, ShaderCompileError> {
  let mut polyfills: naga::FastHashMap<_, _> = Default::default();
  polyfills.insert(
          PolyfillOverload::Inverse {
                    dimension: VectorSize::Quad,
                    width: 4,
                },
                String::from("inverse_4"),
   );

   let mut writer = naga::back::wgsl::Writer::new_with_polyfills(
            String::new(),
            WriterFlags::empty(),
            polyfills,
   );
   writer.write(module, info)?;
   let output = writer.finish();
   Ok(output)
}
```

**Checklist**

- [x] Run `cargo fmt`.

### DIFF
--- a/naga/src/back/wgsl/mod.rs
+++ b/naga/src/back/wgsl/mod.rs
@@ -8,6 +8,7 @@ mod writer;
 
 use thiserror::Error;
 
+pub use writer::PolyfillOverload;
 pub use writer::{Writer, WriterFlags};
 
 #[derive(Error, Debug)]
--- a/naga/src/back/wgsl/writer.rs
+++ b/naga/src/back/wgsl/writer.rs
@@ -2,7 +2,7 @@ use super::Error;
 use crate::{
     back::{self, Baked},
     proc::{self, ExpressionKindTracker, NameKey},
-    valid, Handle, Module, ShaderStage, TypeInner,
+    valid, Handle, MathFunction, Module, ScalarKind, ShaderStage, TypeInner,
 };
 use std::fmt::Write;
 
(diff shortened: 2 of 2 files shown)


teoxoy:
What is the rationale for having users provide a polyfill rather than naga injecting one?

chyyran:
> What is the rationale for having users provide a polyfill rather than naga injecting one?

Two reasons 

1. Punting on the shape/implementation of the polyfill. From the discussion on the gpuweb issue, it seems this is a concern. Having users provide a polyfill defers responsibility from naga for the implementation to the user.
2. Allow an explicit opt-in to polyfilling an expensive function

Of course 2 could be done if naga handled the injection. To be honest that would be preferable to me as well but I don’t want to end up bikeshedding the polyfill implementations.

teoxoy:
I would prefer that we add polyfills in naga rather than this mechanism. As long as our implementation of these functions satisfy the SPIR-V/GLSL specs I see no reason for bikeshedding.

chyyran:
Sounds good, I’ll rework this into having naga provide the polyfills.

PR #6384, opened 2024-10-08
[spv-in] sampling from depth textures returns a scalar and is splatted

### LINKED ISSUE #4551: [spv-in] Sampling results for depth textures need to be changed to scalar floats
This is a followup to https://github.com/gfx-rs/naga/pull/2341. Consider the following SPIR-V shader snippet:

```
<snip>
%type_2d_image = OpTypeImage %float 2D 1 0 0 1 Unknown
<snip>
         %68 = OpSampledImage %type_sampled_image %66 %67
         %69 = OpImageSampleExplicitLod %v4float %68 %65 Lod %float_0
         %70 = OpCompositeExtract %float %69 0
<snip>
```

`OpTypeImage` has a `1` after `2D` to indicate that it is a depth texture. When it is sampled the sample typed is a 4-component vector as [required by the spec](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpImageSampleExplicitLod).

This shader doesn't pass naga validation though because [the wgsl spec requires that depth textures sample to scalar floats](https://www.w3.org/TR/WGSL/#texturesamplelevel):

```
[2023-05-25T23:00:43Z ERROR naga::proc::typifier] Access index type Scalar { kind: Float, width: 4 }
Function [1] 'dof_downsample_with_coc' is invalid: 
        Expression [45] is invalid
        Type resolution failed
        Invalid access into expression [44], indexed: true
Generating wgsl output requires validation to succeed, and it failed in a previous step
```

If I manually mess with the the SPIR-V assembly to change the result type of the sample to a scalar float then naga processes it fine (note that this fails SPIR-V validation):

```
<snip>
%type_2d_image = OpTypeImage %float 2D 1 0 0 1 Unknown
<snip>
         %68 = OpSampledImage %type_sampled_image %66 %67
         %70 = OpImageSampleExplicitLod %float %68 %65 Lod %float_0
<snip>
```

What we need to do in the above case is, upon sampling a depth texture, we need to change the sample type to be a scalar float then insert a composite instruction that creates a 4-component vector with each field set to the scalar sample result (to not break later instructions).

schell:
This wraps the image expression in a splat expression when the texture being sampled/loaded is a depth texture.

**Connections**
This fixes #4551.

**Description**
Sampling from a depth texture fails because SPIR-V expects the result of sampling to be a vector, whereas WGSL expects it to be a scalar.

**Testing**
I ran this on some input shaders.

**Checklist**

- [x] Run `cargo fmt`.


schell:
@jimblandy @cwfitzgerald @teoxoy @Wumpf - I just want to put this on your radar :)

naga/src/front/spv/image.rs
```diff
@@ -377,36 +377,54 @@ impl<I: Iterator<Item = u32>> super::Frontend<I> {
...
         };
+        let image_load_handle = ctx
+            .expressions
+            .append(image_load_expr, self.span_from_with_op(start));
+
+        let handle = if is_depth {
+            let splat_expr = crate::Expression::Splat {
+                size: crate::VectorSize::Quad,
+                value: image_load_handle,
+            };
```
teoxoy: The result type of `OpImageRead` is documented as:

> Result Type must be a scalar or vector of floating-point type or integer type.

So I think we should splat based on the nr of components in the return type.

teoxoy (changes requested):

teoxoy:
Besides the comment I left the changes look correct but could you also add a snapshot test for this?

naga/src/front/spv/image.rs
```diff
@@ -377,36 +377,60 @@ impl<I: Iterator<Item = u32>> super::Frontend<I> {
...
             sample,
             level,
         };
+        let image_load_handle = ctx
+            .expressions
+            .append(image_load_expr, self.span_from_with_op(start));
+
+        let handle = if is_depth {
+            let result_ty = self.lookup_type.lookup(result_type_id)?;
+            match ctx.type_arena[result_ty.handle].inner {
```
teoxoy, line 416:
```suggestion
            // The return type of `OpImageRead` can be a scalar or vector.
            match ctx.type_arena[result_ty.handle].inner {
```

naga/src/front/spv/image.rs
```diff
@@ -656,10 +682,27 @@ impl<I: Iterator<Item = u32>> super::Frontend<I> {
             level,
             depth_ref,
         };
+        let image_sample_handle = ctx.expressions.append(expr, self.span_from_with_op(start));
+        let handle = if is_depth && depth_ref.is_none() {
+            let result_ty = self.lookup_type.lookup(result_type_id)?;
+            match ctx.type_arena[result_ty.handle].inner {
+                crate::TypeInner::Vector { size, .. } => {
```
teoxoy: Do we have to check this here too? I thought all sampling functions return `vec4`s.

schell: I wasn't sure, so I decided to do the safe thing.

teoxoy: I guess it doesn't hurt but might be misleading. I checked yesterday that all of SPIR-V's sampling functions that we handle all return `vec4`s, could you revert this?

schell: Of course 👍

schell:
Ok, g2g @teoxoy :)

teoxoy (approved):

PR #6385, opened 2024-10-09
[naga/wgsl-out]: polyfill `inverse` function

chyyran:
**Connections**
* https://github.com/gfx-rs/wgpu/issues/4330
* https://github.com/gpuweb/gpuweb/issues/4115
* https://github.com/gfx-rs/naga/issues/893

**Description**
`inverse` is not currently supported by WGSL and might never be. However this leaves shaders that do need these functions fundamentally incompatible with WGSL unless they get polyfilled. This PR allows `wgsl-out` to emit polyfills for GLSL `inverse` functions.

The polyfills provided are ported from [an old Mesa commit](https://gitlab.freedesktop.org/mesa/mesa/-/commit/e9df9636b986b09ba37095d963b388571f528e1b), and have been pre-validated with `naga-cli`.

`outer` is not included as part of this patch as it's much more rarer than `inverse`, and another approach would probably be better. For example SPIRV-Cross just [expands the calculation inline](https://github.com/KhronosGroup/SPIRV-Cross/blob/f8b084de61523957546429c627950adf1803afc3/spirv_msl.cpp#L4166).

Pre-emptive support for f16 is included as well. Note that the polyfill files are intentionally not templated to reduce large allocations during shader compilation. This allows `InversePolyfill` to be 2 `’static` pointers and no larger.

**Testing**
Added an `inverse-polyfill.frag` test to the `in/glsl` folder. 

This test is then run to validate the changes. 

```
$ cargo test convert_glsl_folder --features="glsl-in,wgsl-out"
```

**Checklist**

- [x] Run `cargo fmt`.


naga/src/back/wgsl/writer.rs
```diff
@@ -58,6 +59,8 @@ bitflags::bitflags! {
     pub struct WriterFlags: u32 {
         /// Always annotate the type information instead of inferring.
         const EXPLICIT_TYPES = 0x1;
+        /// Insert polyfills for matrix inverse functions.
+        const POLYFILL_INVERSE = 0x2;
```
teoxoy: I don't think we need this new flag, if it's not enabled and the function is used we will error. I don't see a situation where that would be more useful than just injecting the polyfill (considering the SPIR-V/GLSL -> WGSL use case).

cc @jimblandy for thoughts

naga/src/back/wgsl/writer.rs
```diff
@@ -1751,6 +1776,16 @@ impl<W: Write> Writer<W> {
                         }
                         write!(self.out, ")")?
                     }
+                    Function::InversePolyfill(inverse) => {
+                        write!(self.out, "{}(", inverse.fun_name)?;
+                        self.write_expr(module, arg, func_ctx)?;
+                        for arg in IntoIterator::into_iter([arg1, arg2, arg3]).flatten() {
+                            write!(self.out, ", ")?;
+                            self.write_expr(module, arg, func_ctx)?;
+                        }
```
teoxoy:
```suggestion
```
The function only requires one arg.

chyyran:
Fixed clippy errors and applied the suggestion from @teoxoy. Will await consensus from @jimblandy before moving forward with whether to keep the flag. I don't mind either way, but the idea to have the flag stemmed from concerns in https://github.com/gfx-rs/wgpu/issues/4330 where OP seemed to have reservations about the injected polyfill being "quite long".

naga/src/back/wgsl/writer.rs
jimblandy: Yeah, if we're going to generate code that uses `Inverse`, then we might as well just supply it.

jimblandy (dismissed):
Unless I'm missing something, the polyfill names all need to start with `__`, a double underscore. A single underscore isn't a reserved prefix (see `back::wgsl::Writer::reset`'s call to `Namer::reset`), so these injected polyfills could conflict with a definition supplied by the user for `_naga_inverse_blah`.

naga/src/back/wgsl/writer.rs
chyyran: Removed as of b698a8 to unconditionally generate the polyfill

chyyran:
@jimblandy Made the emitted polyfills start with a double underscore and further namespaced them as `__naga_polyfill`. I've had to add an exception to wgsl-in to allow `__naga_polyfill` in the lexer otherwise `wgsl-out` would output WGSL that `wgsl-in` would reject.

jimblandy:
@chyyran No, you're right, double underscore isn't something that the code that we generate can define. We should be producing WGSL that other WGSL implementations would accept. Sorry for the bad advice.

I think instead you should change the polyfill names back, so they start with `_naga` again, and just add "_naga" to the list of reserved prefixes that the WGSL back end passes to `Namer::reset`. The namer will take care of consistently adjusting any identifiers coming from the `naga::Module`, so that should just work. Then we can use `_naga` in the WGSL backend for any future injected stuff.

chyyran:
@jimblandy dropped the changes to add the double underscore and added `_naga` to the list of reserved prefixes in `Namer` in 8309067

teoxoy (approved):
LGTM, thanks!

teoxoy:
@chyyran it looks like the snapshots need updating.

chyyran:
Looks like its sensitive to insertion order. I’ll switch it to use an indexset.

chyyran:
@teoxoy switched `required_polyfills` to an indexset in a0a81ac, could you rerun workflows?

PR #6387, opened 2024-10-09
[naga spv-out] Expand LocalType to permit pointers to matrices.

jimblandy:
In `back::spv`:

- Factor out the numeric variants of `LocalType` into a new enum, `NumericType`.

- Split the `Value` variant into `Numeric` and `LocalPointer` variants, and let `LocalPointer` point to any numeric type, including matrices.

  In subsequent PRs, we'll need to spill matrices out into temporary local variables. This means we'll need to generate SPIR-V pointer-to-matrix types, so `LocalType` needs to be able to represent that.

- Rename `make_local` to `LocalType::from_inner`.

  Change the free function `back::spv::make_local` into an associated function `LocalType::from_inner`.

### DIFF
--- a/naga/src/back/spv/block.rs
+++ b/naga/src/back/spv/block.rs
@@ -4,7 +4,7 @@ Implementations for `BlockContext` methods.
 
 use super::{
     helpers, index::BoundsCheckResult, selection::Selection, Block, BlockContext, Dimension, Error,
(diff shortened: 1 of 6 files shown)


jimblandy:
This PR is big, but boring. All it's doing is rearranging some enums... that happen to be widely used :sob:

teoxoy (approved):
This is great! I like how it simplifies things as well.

PR #6388, opened 2024-10-09
[naga spv-out] Simplify `Writer::get_pointer_id`.

jimblandy:
Simplify the definition of `naga::back::spv::Writer::get_pointer_id` by using `get_type_id`'s ability to handle `LocalType::Pointer`.

This means that `get_pointer_id` is no longer fallible, and no longer needs a type arena. Simplify callers, as well as the `BlockContext::get_pointer_id` convenience function.


teoxoy (approved):
Nice!

Issue #4390, opened 2022-05-14
Relax vector, matrix and array OOB validation


teoxoy:
Relax vector, matrix and array OOB validation by handling OOB accesses at runtime.

The following example should work; however, validation currently errors because the index is out of bounds.

```
@compute @workgroup_size(1)
fn main() {
    let idx = 9;
    let v = vec2<f32>();
    let m = mat2x2<f32>();
    let a = array<f32,1>();
    let vi = v[idx];
    let mi = m[idx];
    let ai = a[idx];
}
```

Related

- gfx-rs/wgpu#4381
- gfx-rs/wgpu#4389

jimblandy:
Note that this issue only covers the excessive validation. There may be missing bounds checks, and those should be filed as separate issues.

nical:
See also: https://github.com/gfx-rs/wgpu/issues/4337

bavalpey:
Curious, why should this work?
Do you just mean it should pass validation? And then the bounds-check policy is used to determine what should happen with `vi`, `mi`, etc.?

I think this would be a regression. If you can detect OOB accesses, as these clearly are, at shader translation time, why not report them?

ErichDonGubler:
@bavalpey: It's for compliance with the WebGPU spec., which forces safe (but implementation-defined) behavior: https://www.w3.org/TR/WGSL/#out-of-bounds-access-sec

andyleiserson:
Some relevant CTS tests:

```
[fail] webgpu:shader,validation,expression,access,array:early_eval_errors:case="runtime_oob_neg" (12ms). Log:
[fail] webgpu:shader,validation,expression,access,array:early_eval_errors:case="runtime_oob_pos" (14ms). Log:
```

PR #6390, opened 2024-10-10
[naga spv-out] Spill arrays and matrices for runtime indexing.

### LINKED ISSUE #4337: Support dynamic indexing of by-value matrices and arrays, per WGSL

### LINKED ISSUE #6358: [naga spv-out] Indexing arrays by value causes unnecessary copies

jimblandy:
I tried to break this up into easy-to-review commits, but ymmv.

- [naga spv-out] Spill arrays and matrices for runtime indexing.

  Improve handling of `Access` expressions whose base is an array or
  matrix (not a pointer to such), and whose index is not known at
  compile time. SPIR-V does not have instructions that can do this
  directly, so spill such values to temporary variables, and perform the
  accesses using `OpAccessChain` instructions applied to the
  temporaries.

  When performing chains of accesses like `a[i].x[j]`, do not reify
  intermediate values; generate a single `OpAccessChain` for the entire
  thing.

  Remove special cases for arrays; the same code now handles arrays and
  matrices.

  Update validation to permit dynamic indexing of matrices.

  For details, see the comments on the new tracking structures in
  `naga::back::spv::Function`.

  Add snapshot test `index-by-value.wgsl`.

  Fixes #6358.
  Fixes #4337.
  Alternative to #6362.

- [naga spv-out] Add some tracing output to `Writer::write_function`.

- [naga spv-out] Introduce `Writer::get_resolution_pointer_id`.

  Introduce a new helper function,
  `naga::back::spv::Writer::get_resolution_pointer_id`. Use it in
  `BlockContext::write_expression_pointer`.

- [naga] Add new function, `GuardedIndex::from_expression`.

  Pull out the code to build a `naga::proc::index::GuardedIndex` from a
  `Handle<Expression>` into its own function,
  `GuardedIndex::from_expression`. Use that function in
  `GuardedIndex::try_resolve_to_constant`.

- [naga spv-out] Abstract out `NumericType::from_inner`.

  Pull out the code for building a `naga::back::spv::NumericType` from a
  `TypeInner` into its own function, `NumericType::from_inner`. Use that
  in `LocalType::from_inner`.

- [naga spv-out] Use `crate::proc::index::GuardedIndex`.

  Replace qualified paths with a `use` directive.

- [naga spv-out] Gather array, matrix, and vector cases.

  This commit is just code motion.

- [naga] Test access to a member/element through a pointer.

- [naga spv-out] Clean up `write_expression_pointer` type adjustment.

  Replace the `return_type_override` argument of
  `BlockContext::write_expression_pointer` with an enum that says how to
  derive the return type from `expr_handle`'s type.

  Introduce a new type, `AccessTypeAdjustment`, that covers possible
  derivation rules.

  This simplifies callers and the callee, in part by making the possible
  alternatives less general, and by giving them explicit names (the
  variants of the `AccessTypeAdjustment` enum).

- [naga spv-out] Move code to load a pointer into its own function.

  Introduce a new function,
  `naga::back::spv::BlockContext::write_checked_load`, that does the
  work of `Expression::Load`.

  This change is just code motion, and should have no effect on
  behavior. The new function will be used in later commits.

Depends on:
- [x] #6387
- [x] #6388
- [x] #6389


jimblandy:
cc @sagudev This is mostly right, I think, but the types on the `OpAccessChain` instructions are wrong. I think that should be easy to fix.

jimblandy:
@sagudev I think this is ready to test.

In the end, I think this wasn't simpler than yours, but I don't think it's too hard to understand, and I got a bunch of nice cleanups in there in the process.

jimblandy:
Tweaked comments, and moved out another logically independent change. I think the final commit should be a pretty clear explanation of the idea now.

teoxoy (approved):
This looks great!

teoxoy:
Should we move the test added by the initial PR into the new snapshot?

https://github.com/gfx-rs/wgpu/pull/6188/files#diff-f63b9d8b161661aa5df1539e321be5f259f9c2e508f3ac970a62f8a050b0fd36

naga/src/back/spv/block.rs
```diff
@@ -293,29 +341,70 @@ impl<'w> BlockContext<'w> {
...
-                    // path, since we handled the `Pointer` case above.
+                    crate::TypeInner::Array { .. } | crate::TypeInner::Matrix { .. } => {
+                        // See if `index` is known at compile time.
+                        match GuardedIndex::from_expression(index, self.ir_function, self.ir_module)
+                        {
+                            GuardedIndex::Known(value) => {
+                                // If `index` is known, we can just use `OpCompositeExtract`.
+                                //
+                                // We never need bounds checks for these cases: everything
+                                // size is statically known and checked in validation.
```
teoxoy: For linking purposes so that we remember to update this in the future: This is not spec compliant but we already have https://github.com/gfx-rs/wgpu/issues/4390 tracking this.

jimblandy: I went ahead and filed a separate issue specifically for this spot.

sagudev:
> @sagudev I think this is ready to test.
> 
> In the end, I think this wasn't simpler than yours, but I don't think it's too hard to understand, and I got a bunch of nice cleanups in there in the process.

I tested in servo and it works as expected🎉.

PR #6391, opened 2024-10-10
Move more code into methods on wgpu-core objects

teoxoy:
Another step towards https://github.com/gfx-rs/wgpu/issues/5121.

### DIFF
--- a/wgpu-core/src/command/transfer.rs
+++ b/wgpu-core/src/command/transfer.rs
@@ -47,8 +47,6 @@ pub enum TransferError {
     MissingBufferUsage(#[from] MissingBufferUsageError),
     #[error(transparent)]
     MissingTextureUsage(#[from] MissingTextureUsageError),
-    #[error("Destination texture is missing the `RENDER_ATTACHMENT` usage flag")]
-    MissingRenderAttachmentUsageFlag(TextureId),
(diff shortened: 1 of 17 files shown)


wgpu-core/src/command/transfer.rs
```diff
@@ -48,7 +48,7 @@ pub enum TransferError {
     #[error(transparent)]
     MissingTextureUsage(#[from] MissingTextureUsageError),
     #[error("Destination texture is missing the `RENDER_ATTACHMENT` usage flag")]
-    MissingRenderAttachmentUsageFlag(TextureId),
+    MissingRenderAttachmentUsageFlag,
```
Wumpf: any way to preserve the extra error information of which texture had the issue? 🤔

Wumpf (approved):
nice! things getting more and more clearer, like a fog slowly lifting :)
extra cool that you got to replace a unwrap surface case with more graceful error handling in the process!

wgpu-core/src/command/transfer.rs
teoxoy: I fixed this one since it was easy but not `InvalidDimensionExternal` since lots of other variants of `TransferError` don't include the texture label. We should find a better mechanism to include the label for all remaining variants.

PR #6413, opened 2024-10-15
Wait for submissions to complete on `Queue` drop

teoxoy:
This PR addresses leaks caused by circular references due to the `Device` still holding strong references to resources when removed from the registry and while submissions are still active. `Global::device_drop` was wrongly assuming `device_poll` with `Maintain::Wait` was called but this is not a documented invariant and only `wgpu` was upholding this.

Instead of calling `device_poll` in `device_drop` (which would solve the issue) this PR takes a different approach since we want to remove the registries in the future (https://github.com/gfx-rs/wgpu/issues/5121):

- it moves the `LifetimeTracker` and `PendingWrites` into the `Queue` so that we never have circular references
- it adds the wait for active submissions in `Queue`'s `Drop` impl

One thing to note is that ideally we shouldn't be waiting in the `Queue`'s `Drop` implementation but that's what `wgpu` was previously doing (by calling `device_poll`) and the alternative is more involved: We could put the burden of keeping the `Queue` alive on `wgpu-core` users if there are any active submissions. With the changes in this PR we will panic if we hit a timeout or ran into any errors; this would solve that too. I want to talk about this approach at the next maintainers call before making any changes though.

### DIFF
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -2078,20 +2078,7 @@ impl Global {
         profiling::scope!("Device::drop");
         api_log!("Device::drop {device_id:?}");
 
-        let device = self.hub.devices.remove(device_id);
-        let device_lost_closure = device.lock_life().device_lost_closure.take();
-        if let Some(closure) = device_lost_closure {
-            closure.call(DeviceLostReason::Dropped, String::from("Device dropped."));
-        }
-
-        // The things `Device::prepare_to_die` takes care are mostly
-        // unnecessary here. We know our queue is empty, so we don't
-        // need to wait for submissions or triage them. We know we were
-        // just polled, so `life_tracker.free_resources` is empty.
-        debug_assert!(device.lock_life().queue_empty());
-        device.pending_writes.lock().deactivate();
-
-        drop(device);
+        self.hub.devices.remove(device_id);
     }
 
     // This closure will be called exactly once during "lose the device",
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -219,9 +219,11 @@ impl Global {
         device.check_is_valid()?;
         buffer.check_usage(wgt::BufferUsages::MAP_WRITE)?;
 
-        let last_submission = device
-            .lock_life()
-            .get_buffer_latest_submission_index(&buffer);
+        let last_submission = device.get_queue().and_then(|queue| {
+            queue
+                .lock_life()
+                .get_buffer_latest_submission_index(&buffer)
+        });
 
         if let Some(last_submission) = last_submission {
             device.wait_for_submit(last_submission)?;
--- a/wgpu-core/src/device/global.rs
+++ b/wgpu-core/src/device/global.rs
@@ -2090,14 +2090,14 @@ impl Global {
     ) {
         let device = self.hub.devices.get(device_id);
 
-        let mut life_tracker = device.lock_life();
-        if let Some(existing_closure) = life_tracker.device_lost_closure.take() {
-            // It's important to not hold the lock while calling the closure.
-            drop(life_tracker);
-            existing_closure.call(DeviceLostReason::ReplacedCallback, "".to_string());
-            life_tracker = device.lock_life();
+        let old_device_lost_closure = device
+            .device_lost_closure
+            .lock()
+            .replace(device_lost_closure);
+
+        if let Some(old_device_lost_closure) = old_device_lost_closure {
+            old_device_lost_closure.call(DeviceLostReason::ReplacedCallback, "".to_string());
         }
-        life_tracker.device_lost_closure = Some(device_lost_closure);
     }
 
     pub fn device_destroy(&self, device_id: DeviceId) {
--- a/wgpu-core/src/device/mod.rs
+++ b/wgpu-core/src/device/mod.rs
@@ -298,24 +298,25 @@ impl DeviceLostClosure {
     }
 }
 
-fn map_buffer(
-    raw: &dyn hal::DynDevice,
+pub(crate) fn map_buffer(
     buffer: &Buffer,
     offset: BufferAddress,
     size: BufferAddress,
     kind: HostMap,
     snatch_guard: &SnatchGuard,
 ) -> Result<hal::BufferMapping, BufferAccessError> {
+    let raw_device = buffer.device.raw();
     let raw_buffer = buffer.try_raw(snatch_guard)?;
     let mapping = unsafe {
-        raw.map_buffer(raw_buffer, offset..offset + size)
+        raw_device
+            .map_buffer(raw_buffer, offset..offset + size)
             .map_err(|e| buffer.device.handle_hal_error(e))?
     };
 
     if !mapping.is_coherent && kind == HostMap::Read {
         #[allow(clippy::single_range_in_vec_init)]
         unsafe {
-            raw.invalidate_mapped_ranges(raw_buffer, &[offset..offset + size]);
+            raw_device.invalidate_mapped_ranges(raw_buffer, &[offset..offset + size]);
         }
     }
 
@@ -370,7 +371,7 @@ fn map_buffer(
                 && kind == HostMap::Read
                 && buffer.usage.contains(wgt::BufferUsages::MAP_WRITE)
             {
-                unsafe { raw.flush_mapped_ranges(raw_buffer, &[uninitialized]) };
+                unsafe { raw_device.flush_mapped_ranges(raw_buffer, &[uninitialized]) };
             }
         }
     }
--- a/wgpu-core/src/device/queue.rs
+++ b/wgpu-core/src/device/queue.rs
@@ -14,7 +14,7 @@ use crate::{
     hal_label,
     id::{self, QueueId},
     init_tracker::{has_copy_partial_init_tracker_coverage, TextureInitRange},
-    lock::{rank, Mutex, RwLockWriteGuard},
+    lock::{rank, Mutex, MutexGuard, RwLockWriteGuard},
     resource::{
         Buffer, BufferAccessError, BufferMapState, DestroyedBuffer, DestroyedResourceError,
         DestroyedTexture, Fallible, FlushedStagingBuffer, InvalidResourceError, Labeled,
@@ -37,12 +37,13 @@ use std::{
 };
 use thiserror::Error;
 
-use super::Device;
+use super::{life::LifetimeTracker, Device};
 
 pub struct Queue {
     raw: ManuallyDrop<Box<dyn hal::DynQueue>>,
     pub(crate) device: Arc<Device>,
     pub(crate) pending_writes: Mutex<ManuallyDrop<PendingWrites>>,
+    life_tracker: Mutex<LifetimeTracker>,
 }
 
 impl Queue {
@@ -94,11 +95,17 @@ impl Queue {
             raw: ManuallyDrop::new(raw),
             device,
             pending_writes,
+            life_tracker: Mutex::new(rank::QUEUE_LIFE_TRACKER, LifetimeTracker::new()),
         })
     }
 
     pub(crate) fn raw(&self) -> &dyn hal::DynQueue {
         self.raw.as_ref()
+    }
+
+    #[track_caller]
+    pub(crate) fn lock_life<'a>(&'a self) -> MutexGuard<'a, LifetimeTracker> {
+        self.life_tracker.lock()
     }
 }
 
@@ -1292,7 +1299,7 @@ impl Queue {
             profiling::scope!("cleanup");
 
             // this will register the new submission to the life time tracker
-            self.device.lock_life().track_submission(
+            self.lock_life().track_submission(
                 submit_index,
                 pending_writes.temp_resources.drain(..),
                 active_executions,
@@ -1341,7 +1348,7 @@ impl Queue {
     ) -> Option<SubmissionIndex> {
         api_log!("Queue::on_submitted_work_done");
         //TODO: flush pending writes
-        self.device.lock_life().add_work_done_closure(closure)
+        self.lock_life().add_work_done_closure(closure)
     }
 }
 
--- a/wgpu-core/src/device/queue.rs
+++ b/wgpu-core/src/device/queue.rs
@@ -14,7 +14,7 @@ use crate::{
     hal_label,
     id::{self, QueueId},
     init_tracker::{has_copy_partial_init_tracker_coverage, TextureInitRange},
-    lock::RwLockWriteGuard,
+    lock::{rank, Mutex, RwLockWriteGuard},
     resource::{
         Buffer, BufferAccessError, BufferMapState, DestroyedBuffer, DestroyedResourceError,
         DestroyedTexture, Fallible, FlushedStagingBuffer, InvalidResourceError, Labeled,
@@ -42,14 +42,59 @@ use super::Device;
 pub struct Queue {
     raw: ManuallyDrop<Box<dyn hal::DynQueue>>,
     pub(crate) device: Arc<Device>,
+    pub(crate) pending_writes: Mutex<ManuallyDrop<PendingWrites>>,
 }
 
 impl Queue {
-    pub(crate) fn new(device: Arc<Device>, raw: Box<dyn hal::DynQueue>) -> Self {
-        Queue {
+    pub(crate) fn new(
+        device: Arc<Device>,
+        raw: Box<dyn hal::DynQueue>,
+    ) -> Result<Self, DeviceError> {
+        let pending_encoder = device
+            .command_allocator
+            .acquire_encoder(device.raw(), raw.as_ref())
+            .map_err(DeviceError::from_hal);
+
+        let pending_encoder = match pending_encoder {
+            Ok(pending_encoder) => pending_encoder,
+            Err(e) => {
+                device.release_queue(raw);
+                return Err(e);
+            }
+        };
+
+        let mut pending_writes = PendingWrites::new(pending_encoder);
+
+        let zero_buffer = device.zero_buffer.as_ref();
+        pending_writes.activate();
+        unsafe {
+            pending_writes
+                .command_encoder
+                .transition_buffers(&[hal::BufferBarrier {
+                    buffer: zero_buffer,
+                    usage: hal::BufferUses::empty()..hal::BufferUses::COPY_DST,
+                }]);
+            pending_writes
+                .command_encoder
+                .clear_buffer(zero_buffer, 0..super::ZERO_BUFFER_SIZE);
+            pending_writes
+                .command_encoder
+                .transition_buffers(&[hal::BufferBarrier {
+                    buffer: zero_buffer,
+                    usage: hal::BufferUses::COPY_DST..hal::BufferUses::COPY_SRC,
+                }]);
+        }
+
+        let pending_writes = Mutex::new(
+            rank::QUEUE_PENDING_WRITES,
+            ManuallyDrop::new(pending_writes),
+        );
+
+        Ok(Queue {
             raw: ManuallyDrop::new(raw),
             device,
-        }
+            pending_writes,
+        })
     }
 
     pub(crate) fn raw(&self) -> &dyn hal::DynQueue {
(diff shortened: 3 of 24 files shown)


teoxoy:
CI was failing because when targeting WebGL we were timing out. eade5cfb4027095accec7ca31af1b3a00872aaa3 fixes CI and behaves the same as current trunk where the timeout is ignored (related: https://github.com/gfx-rs/wgpu/issues/3601 & https://github.com/gfx-rs/wgpu/issues/4589) but is not correct.

This is another argument for requiring users of `wgpu-core` to keep the `Queue` alive and keep polling it until submissions complete.

teoxoy:
Conclusions after discussing the idea of requiring users of `wgpu-core` to keep the `Queue` alive and keep polling it until submissions complete with the other maintainers:
- Most users don’t call poll so we shouldn’t make it a requirement (maintain gets called implicitly on submit).
- @cwfitzgerald said it should be fine to drop everything early on WebGL since WebGL implementations will still keep the objects alive (I will double check with @kdashg).
- We should still implement the idea for Firefox since we shouldn't be blocking the GPU process.

teoxoy:
Some follow-up work is still needed though, instead of panicking, we need to add a retry mechanism if we time out or if we run into OOMs. Device loss needs to be propagated to the device (making it invalid).

teoxoy:
I revised the comment explaining why it's actually ok that we time out on WebGL.

teoxoy:
I added the retry mechanism as well, this should be ready for review.

ErichDonGubler:
Looks like this has compilation issues after being rebased onto #6291.

teoxoy:
The last commit (see its description for details) fixes the issue that CI was blocked on in https://github.com/gfx-rs/wgpu/actions/runs/11799523804/job/32869472729

I have a follow-up that cleans up command buffers/encoders since they are currently more complicated than they can be; will put up the PR once this one lands.

wgpu-core/src/device/resource.rs
```diff
@@ -171,6 +171,12 @@ impl std::fmt::Debug for Device {
 impl Drop for Device {
     fn drop(&mut self) {
         resource_log!("Drop {}", self.error_ident());
+
+        let device_lost_closure = self.lock_life().device_lost_closure.take();
+        if let Some(closure) = device_lost_closure {
+            closure.call(DeviceLostReason::Dropped, String::from("Device dropped."));
+        }
+
```
ErichDonGubler: thought: I suspect that moving the `device_lost_closure` call to `Drop` will resolve some issues I've seen in CTS indicating that it wasn't expecting to see the device lost callback being invoked.

wgpu-core/src/resource.rs
```diff
@@ -778,17 +779,20 @@ impl Buffer {
...
-            let mut life_lock = device.lock_life();
-            let last_submit_index = life_lock.get_buffer_latest_submission_index(self);
-            if let Some(last_submit_index) = last_submit_index {
-                life_lock.schedule_resource_destruction(temp, last_submit_index);
+        if let Some(queue) = device.get_queue() {
+            let mut pending_writes = queue.pending_writes.lock();
+            if pending_writes.contains_buffer(self) {
+                pending_writes.consume_temp(temp);
+                return Ok(());
             }
```
ErichDonGubler: thought: This and the below changes are interesting, because previously, `device.pending_writes` had a lock region that overlapped the `life_lock` region. Now, they're not overlapping at all.

ErichDonGubler: question: Actually, I'm surprised this doesn't result in a deletion of an entry in the `QUEUE_PENDING_WRITES` entry in `wgpu-core/src/lock/rank.rs`. I'm guessing that this wasn't done because there are other cases where the `life_lock` is acquired within the region of a `pending_writes` lock?

wgpu-core/src/device/queue.rs
```diff
@@ -42,14 +42,59 @@ use super::Device;
 pub struct Queue {
     raw: ManuallyDrop<Box<dyn hal::DynQueue>>,
     pub(crate) device: Arc<Device>,
+    pub(crate) pending_writes: Mutex<ManuallyDrop<PendingWrites>>,
```
ErichDonGubler, line 46: question: RE: the `move PendingWrites into the Queue` commit: The commit message claims that it avoids (removes?) strong ref. cycles between a `Device` and resources containing backlinks to it. I don't see how moving `pending_writes` from a `Device` to a `Queue` (which seems to be held by a `Device`, and also hold a strong ref. to a `Device) does anything other than increase the number of steps in a cycle. Am I missing something?

ErichDonGubler: Ah, I think I understand the answer: a `Queue` is held behind a `Weak` ref., rather than an `Arc`, in a `Device`. So, this should break a cycle, because the ref. link to `Queue` is comparatively weaker.

wgpu-core/src/device/resource.rs
```diff
@@ -57,24 +57,6 @@ use super::{
...
-///
-/// Right now avoid locking twice same resource or registry in a call execution
-/// and minimize the locking to the minimum scope possible
-/// Unless otherwise specified, no lock may be acquired while holding another lock.
-/// This means that you must inspect function calls made while a lock is held
-/// to see what locks the callee may try to acquire.
-///
-/// Important:
-/// When locking pending_writes please check that trackers is not locked
-/// trackers should be locked only when needed for the shortest time possible
```
ErichDonGubler, line 79: note: CC @jimblandy for the ` Remove outdated locking comments ` commit specifically.

wgpu-core/src/device/global.rs
```diff
@@ -2090,14 +2091,14 @@ impl Global {
...
-            drop(life_tracker);
-            existing_closure.call(DeviceLostReason::ReplacedCallback, "".to_string());
-            life_tracker = device.lock_life();
+        let old_device_lost_closure = std::mem::replace(
+            device.device_lost_closure.lock().deref_mut(),
+            Some(device_lost_closure),
+        );
+
+        if let Some(old_device_lost_closure) = old_device_lost_closure {
+            old_device_lost_closure.call(DeviceLostReason::ReplacedCallback, "".to_string());
```
ErichDonGubler, line 2101: suggestion(non-blocking): I think the ` move device_lost_closure on the Device ` commit title is less clear than, say, ``move `device_lost_closure` from `Device::life_tracker`  to `Device``

wgpu-core/src/device/global.rs
```diff
@@ -2090,14 +2091,14 @@ impl Global {
...
-        let mut life_tracker = device.lock_life();
-        if let Some(existing_closure) = life_tracker.device_lost_closure.take() {
-            // It's important to not hold the lock while calling the closure.
-            drop(life_tracker);
-            existing_closure.call(DeviceLostReason::ReplacedCallback, "".to_string());
-            life_tracker = device.lock_life();
+        let old_device_lost_closure = std::mem::replace(
+            device.device_lost_closure.lock().deref_mut(),
+            Some(device_lost_closure),
+        );
```
ErichDonGubler: nitpick: This seems like a re-invention of `Option::replace`:


```suggestion
        let old_device_lost_closure = device
            .device_lost_closure
            .lock()
            .replace(device_lost_closure);
```

wgpu-core/src/device/resource.rs
```diff
@@ -147,8 +152,7 @@ impl Drop for Device {
     fn drop(&mut self) {
         resource_log!("Drop {}", self.error_ident());
 
-        let device_lost_closure = self.lock_life().device_lost_closure.take();
```
ErichDonGubler: question: This change eliminates coverage of a `life_tracker` lock region over some cleanup operations after this hunk. I presume this is safe (from a place of ignorance), because these should be exclusively held resources. Is that right? I'm nervous about the `fence` in particular, because we acquire a `write` lock on it before `ManuallyDrop::take`ing it.

wgpu-core/src/resource.rs
```diff
@@ -634,7 +634,12 @@ impl Buffer {
             .buffers
             .set_single(self, internal_use);
 
-        let submit_index = device.lock_life().map(self).unwrap_or(0); // '0' means no wait is necessary
+        let submit_index = if let Some(queue) = device.get_queue() {
+            queue.lock_life().map(self).unwrap_or(0) // '0' means no wait is necessary
+        } else {
+            // TODO: map immediately
```
ErichDonGubler: question: Is there a reason we can't immediately implement `TODO: map immediately` in the ` move LifetimeTracker into the Queue ` commit?

I'll likely see the answer to this myself shortly, since I'm not all the way through commits, and I see this `TODO` doesn't survive in the overall diff.

wgpu-core/src/device/mod.rs
```diff
@@ -298,24 +298,25 @@ impl DeviceLostClosure {
     }
 }
 
-fn map_buffer(
-    raw: &dyn hal::DynDevice,
+pub fn map_buffer(
```
ErichDonGubler: nitpick: It seems like this only needs to be promoted to `pub(crate)`? 🤔

wgpu-core/src/resource.rs
ErichDonGubler: Yep, answered in the next commit. 🫡

wgpu-core/src/device/queue.rs
```diff
@@ -143,12 +143,56 @@ crate::impl_storage_item!(Queue);
 impl Drop for Queue {
     fn drop(&mut self) {
         resource_log!("Drop {}", self.error_ident());
+
+        let last_successful_submission_index = self
+            .device
+            .last_successful_submission_index
+            .load(Ordering::Acquire);
```
ErichDonGubler, line 149: todo: I want to double-check atomic orderings again, just to make sure I feel confident with this. I don't trust myself to eyeball this. 😅

wgpu-core/src/device/queue.rs
```diff
@@ -143,12 +143,56 @@ crate::impl_storage_item!(Queue);
...
+            Ok(true) => {}
+            // Note: If we don't panic here we are in UB land (destroying resources while they are still in use by the GPU).
+            Ok(false) => {
+                panic!("We timed out while waiting on the last successful submission to complete!");
+            }
+            Err(e) => {
+                panic!(
+                    "We ran into an error while waiting on the last successful submission to complete! - {e}"
+                );
+            }
```
ErichDonGubler: note: panicking in `Drop` can result in an abortion of the program if `Drop` was being called in the middle of a panic (see [`std::Drop`'s `Panics` section](https://doc.rust-lang.org/std/ops/trait.Drop.html#panics)).

wgpu-core/src/device/queue.rs
```diff
@@ -143,12 +143,56 @@ crate::impl_storage_item!(Queue);
...
     fn drop(&mut self) {
         resource_log!("Drop {}", self.error_ident());
+
+        let last_successful_submission_index = self
+            .device
+            .last_successful_submission_index
+            .load(Ordering::Acquire);
+
+        let fence = self.device.fence.read();
+        let wait_res = unsafe {
```
ErichDonGubler: nitpick: Missing a `SAFETY` comment here.

wgpu-core/src/device/queue.rs
```diff
@@ -143,12 +143,56 @@ crate::impl_storage_item!(Queue);
...
+                panic!(
+                    "We ran into an error while waiting on the last successful submission to complete! - {e}"
+                );
+            }
+        }
+
+        let snatch_guard = self.device.snatchable_lock.read();
+        let (submission_closures, mapping_closures, queue_empty) =
+            self.maintain(last_successful_submission_index, &snatch_guard);
+        drop(snatch_guard);
```
ErichDonGubler: nitpick: Manual `drop`s of locks are sometimes necessary, but this doesn't seem like a case where it _is_ necessary.	

suggestion: I think it's strictly easier to perform assignment of these variables inside a block, so we can't forget to unlock locks.

```suggestion
        let wait_res;
        {
            let fence = self.device.fence.read();
            wait_res = unsafe {
                self.device.raw().wait(
                    fence.as_ref(),
                    last_successful_submission_index,
                    crate::device::CLEANUP_WAIT_MS,
                )
            };
        }

        match wait_res {
            Ok(true) => {}
            // Note: If we don't panic here we are in UB land (destroying resources while they are still in use by the GPU).
            Ok(false) => {
                panic!("We timed out while waiting on the last successful submission to complete!");
            }
            Err(e) => {
                panic!(
                    "We ran into an error while waiting on the last successful submission to complete! - {e}"
                );
            }
        }

        let (submission_closures, mapping_closures, queue_empty);
        {
            let snatch_guard = self.device.snatchable_lock.read();
            (submission_closures, mapping_closures, queue_empty) =
                self.maintain(last_successful_submission_index, &snatch_guard);
        }
```

wgpu-core/src/device/queue.rs
```diff
@@ -163,6 +166,13 @@ impl Drop for Queue {
             Ok(true) => {}
             // Note: If we don't panic here we are in UB land (destroying resources while they are still in use by the GPU).
             Ok(false) => {
+                // It's fine that we timed out on WebGL; GL objects can be deleted early as they
+                // will be kept around by the driver if GPU work hasn't finished.
+                // Moreover, the way we emulate read mappings on WebGL allows us to execute map_buffer earlier than on other
+                // backends since getBufferSubData is synchronous with respect to the other previously enqueued GL commands.
+                // TODO: Relying on this behavior breaks the clean abstraction wgpu-hal tries to maintain and
+                // we should find ways to improve this.
```
ErichDonGubler: todo: I don't think we should leave a `TODO` here on the final merge; we can just delete these lines after we file an issue.

wgpu-hal/src/gles/device.rs
```diff
@@ -1563,6 +1563,10 @@ impl crate::Device for super::Device {
     ) -> Result<bool, crate::DeviceError> {
         if fence.last_completed.load(Ordering::Relaxed) < wait_value {
             let gl = &self.shared.context.lock();
+            // MAX_CLIENT_WAIT_TIMEOUT_WEBGL is:
+            // - 1s in Gecko https://searchfox.org/mozilla-central/rev/754074e05178e017ef6c3d8e30428ffa8f1b794d/dom/canvas/WebGLTypes.h#1386
+            // - 0 in WebKit https://github.com/WebKit/WebKit/blob/4ef90d4672ca50267c0971b85db403d9684508ea/Source/WebCore/html/canvas/WebGL2RenderingContext.cpp#L110
+            // - 0 in Chromium https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webgl/webgl2_rendering_context_base.cc;l=112;drc=a3cb0ac4c71ec04abfeaed199e5d63230eca2551
             let timeout_ns = if cfg!(any(webgl, Emscripten)) {
```
ErichDonGubler, line 1562: question: How is the above commentary relevant to the `cfg!(…)` below? These are user agents, but we're not branching on anything like a user agent report.

wgpu-hal/src/gles/mod.rs
```diff
@@ -295,6 +295,13 @@ pub struct Device {
     counters: wgt::HalCounters,
 }
 
+impl Drop for Device {
+    fn drop(&mut self) {
+        let gl = &self.shared.context.lock();
+        unsafe { gl.delete_vertex_array(self.main_vao) };
```
ErichDonGubler, line 301: question: Why was the body of `gles::Device` not migrated in its entirety here?

answer: Oh, I see that it got split between `impl Drop for Device` and `impl Drop for Queue`.

question: This means that these operations were re-ordered a bit. Is this okay?

wgpu-core/src/device/queue.rs
```diff
@@ -149,37 +149,75 @@ impl Drop for Queue {
...
+                    hal::DeviceError::ResourceCreationFailed => unreachable!(),
+                    hal::DeviceError::Unexpected => {
+                        panic!(
+                            "We ran into an unexpected error while waiting on the last successful submission to complete!"
+                        );
+                    }
+                },
             }
         }
+        drop(fence);
```
ErichDonGubler, line 220: question: If we panic here, but don't abort, then we may have poisoned the `fence` lock here. Should we `drop` it before panicking?

wgpu-core/src/device/queue.rs
```diff
@@ -1209,6 +1209,13 @@ impl Queue {
                             }
                         }
 
+                        if first_error.is_some() {
+                            if let Ok(cmd_buf_data) = cmd_buf_data {
+                                cmd_buf_data.destroy(&command_buffer.device);
```
ErichDonGubler, line 1214: thought: I guess that this is technically more correct than before. The diagnostics seem like they would be wrong if a user attempts to use these command buffers again, though (i.e., "You can't use this, this is destroyed." as opposed to "this is invalid").

Not concerned either way, though.

ErichDonGubler (commented):
Inclined to approve if we can get my questions answered.

wgpu-core/src/resource.rs
teoxoy: Maybe, but I didn't check.

teoxoy: One of the next commits that moves the `LifetimeTracker` into the `Queue` changes this back to the previous shape.

wgpu-core/src/device/resource.rs
teoxoy: As far as I understand in this case the lifetime of the guard didn't have to be extended, it's in a temporary scope.

wgpu-core/src/device/queue.rs
teoxoy: I guess I should be wrapping them in `if !thread::panicking() {`?

wgpu-core/src/device/queue.rs
teoxoy: I don't know what the safety requirements of calling this function are... I'd have to check and document all the backends for this. All wgpu-hal functions are marked as unsafe without a safety contract. We should document the safety contract at some point but I don't think this is the time for that.

wgpu-hal/src/gles/device.rs
teoxoy: It's info as to why the timeout needs to be 0 on WebGL. I was confused at first why that was given that there was `MAX_CLIENT_WAIT_TIMEOUT_WEBGL` that apps could query for but it turns out to be a static 0 almost everywhere.

wgpu-hal/src/gles/mod.rs
teoxoy: I checked and it's fine.

wgpu-core/src/device/queue.rs
teoxoy: This is a `parking_lot::RwLock` which says "No poisoning, the lock is released normally on panic.".

wgpu-core/src/device/queue.rs
teoxoy: From the user's POV command buffers can't be destroyed. They always become invalid once submitted. Here we were just forgetting to destroy the underlying command buffer/encoder.

wgpu-core/src/device/queue.rs
teoxoy: I prefer to have the explicit `drop` calls rather than adding new scopes to trigger `drop` implicitly.

wgpu-core/src/device/queue.rs
teoxoy: Actually I think aborting in this case is appropriate. If we don't, that's UB and we shouldn't continue unwinding.

wgpu-core/src/device/queue.rs
teoxoy: Filed: https://github.com/gfx-rs/wgpu/issues/6538

wgpu-core/src/device/queue.rs
ErichDonGubler: [Definitely not blocking](https://web.archive.org/web/20241002184501/https://conventionalcomments.org/#labels), by any means. Running into an abort here means (1) a user was panicking for some reason and (2) an error was returned in a function that has no means of dealing with an error.

If somebody runs into errors, do we have a clear workaround identified? If so, I'm a _lot_ less concerned about this, though I will not block even if there is not (since I know there's yet more follow-up work to do that should resolve it anyway).

wgpu-core/src/device/queue.rs
ErichDonGubler: Seems right to me, resolving!

ErichDonGubler:
LGTM, with the discussions we've had. If you're still interested in @jimblandy's feedback, I'd suggest giving 'im...maybe 24 hours, if you're feeling patient? Otherwise, I think it's reasonable to do review after merging, if really necessary.

ErichDonGubler (approved):

wgpu-core/src/device/queue.rs
teoxoy: I implemented a retry mechanism for the cases that we can recover from. I can't think of other ways to deal with these errors.

teoxoy:
Let's land it, I already had to rebase it and fix conflicts a few times.

PR #6419, opened 2024-10-17
Minimize moves by introducing `WeakVec`

teoxoy:
This should improve the situation in https://github.com/bevyengine/bevy/issues/15893.

Follow-up to https://github.com/gfx-rs/wgpu/pull/5874.

@DGriffin91 could you give this PR a try?

### DIFF
--- a/wgpu-core/src/device/resource.rs
+++ b/wgpu-core/src/device/resource.rs
@@ -31,6 +31,7 @@ use crate::{
         UsageScopePool,
     },
     validation::{self, validate_color_attachment_bytes_per_sample},
+    weak_vec::WeakVec,
     FastHashMap, LabelHelpers, PreHashedKey, PreHashedMap,
 };
 
@@ -42,7 +43,7 @@ use wgt::{
 
 use std::{
     borrow::Cow,
-    mem::ManuallyDrop,
+    mem::{self, ManuallyDrop},
     num::NonZeroU32,
     sync::{
         atomic::{AtomicBool, AtomicU64, Ordering},
@@ -150,8 +151,8 @@ pub struct Device {
 }
 
 pub(crate) enum DeferredDestroy {
-    TextureView(Weak<TextureView>),
-    BindGroup(Weak<BindGroup>),
+    TextureViews(WeakVec<TextureView>),
+    BindGroups(WeakVec<BindGroup>),
 }
 
 impl std::fmt::Debug for Device {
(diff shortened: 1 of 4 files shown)


wgpu-core/src/weak_vec.rs
```diff
@@ -0,0 +1,66 @@
...
+                    *value = None;
+                    self.empty_slots.push(i);
+                }
+            }
+        }
+        if let Some(i) = self.empty_slots.pop() {
+            self.inner[i] = Some(value);
+        } else {
+            self.inner.push(Some(value));
+        }
```
ErichDonGubler: perf: We already build the knowledge we need to find an empty slot in the stale clearing iteration. Because of this, it seems completely unnecessary to keep a free list.

teoxoy: Right, good point.

ErichDonGubler: I have some code already! Just pushed up 1f94aad6 to resolve this; whaddaya think?

wgpu-core/src/weak_vec.rs
```diff
@@ -0,0 +1,70 @@
+//! Module containing the [`WeakVec`] API.
```
ErichDonGubler, line 1: note: I added this trivial module doc. to orient newcomers.

wgpu-core/src/weak_vec.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub(crate) fn new() -> Self {
+        Self { inner: Vec::new() }
+    }
+
+    /// Pushes a new element to this collection, dropping older elements that no longer have
+    /// a strong reference to them.
+    ///
+    /// NOTE: The length and capacity of this collection do not change when old elements are
+    /// dropped.
+    pub(crate) fn push(&mut self, value: Weak<T>) {
```
ErichDonGubler: nitpick: This comment about what the function does would really be better served as a doc. comment.

Pushed as a `fixup!` with ca93e72b.

wgpu-core/src/weak_vec.rs
teoxoy: Hmm, I was in process of addressing this with:

```rust
pub(crate) fn push(&mut self, value: Weak<T>) {
    let mut to_insert = Some(value);
    for slot in &mut self.inner {
        if let Some(w) = slot {
            if w.strong_count() == 0 {
                *slot = to_insert.take();
            }
        } else {
            *slot = to_insert.take();
        }
    }
    if let Some(to_insert) = to_insert {
        self.inner.push(Some(to_insert));
    }
}
```

Comparing the complied code it looks like this function has less jumps and moves.
Probably because the approach in https://github.com/gfx-rs/wgpu/commit/1f94aad66aa777b0adb12c2e7756012cd3d5f295 is not as straight forward to the compiler?

https://rust.godbolt.org/z/zKrE4xPKK
https://rust.godbolt.org/z/W7nW68WPW

https://www.diffchecker.com/XWsoSpeZ/

teoxoy: I will revise the patch to use the code above.

wgpu-core/src/device/resource.rs
ErichDonGubler, line 1: note: I initially worried that by removing these `Vec::retain` calls, we remove sites where these `Vec`s of weak refs. may have their lengths decreased. This, in turn, could put pressure on re-allocations. I actually think now that it's even less likely with current code. I guess another way of saying this is that I feel good that this PR achieves its goals!

wgpu-core/src/weak_vec.rs
```diff
@@ -0,0 +1,67 @@
...
```
ErichDonGubler, line 37: question(non-blocking): Having an API that _only_ exposes a `push` and an `IntoIterator` interface makes it easy to ask: How are we ensuring that a `WeakVec` gets its capacity shrunk down again if it ever gets large? I presume that this is done by consuming the `WeakVec`s with `IntoIterator`.

We had this problem to consider before this PR was filed, so I won't block merging with this question. I intend to read code to answer this, but I want to confirm this understanding "out loud", as it were. We can punt to an issue, if we discover something substantial with this discussion.

ErichDonGubler (approved):
Conditionally approving based on user confirmation that this does, indeed, solve a problem for them. I'm open to merging in the absence of that feedback, but I think waiting a day or so to confirm would be good.

wgpu-core/src/weak_vec.rs
teoxoy: I would imagine it's not common for apps to create lots of bind groups and texture views then consistently use less over time. `Weak` is also not large at 8 bytes.

DGriffin91:
Unfortunately it seems like bevy still has the issue with this PR. It might be slightly faster though. I have the [bevy branch with this pr](https://github.com/DGriffin91/bevy/tree/minimize-moves) up if you wanted to play with it.

First frame time with `cargo run --example many_cubes --release -- --vary-material-data-per-instance`

119619ms: [bevy w/ minimize-moves](https://github.com/DGriffin91/bevy/tree/minimize-moves)
129969ms: bevy main

vtune filtered in on just the part that is hanging w/ minimize-moves:
image

tychedelia:
The problem here as far as I can tell is that we are creating 160k unique bind groups that all share the same texture in a loop. Even without the call to retain, `WeakVec.push(..)` iterates over `self.inner` referencing all the bind groups we've created so far, which ends up looking exponential:
image

teoxoy:
> The problem here as far as I can tell is that we are creating 160k unique bind groups that all share the same texture in a loop.

This is worst case scenario for this code path. We can alleviate this issue further by running the code every X invocations of push rather than on every invocation - I will give it a try.

teoxoy:
I pushed another commit that should be a good balance between perf and memory use.

@DGriffin91 & @tychedelia could you give it a try?

tychedelia:
@teoxoy Yay! Validated on Metal and Vulkan That seems to do it. 🎉✨

Issue #6402, opened 2024-10-11
Converge dual-source blending with the WebGPU standard


ErichDonGubler:
https://www.w3.org/TR/webgpu/#dom-gpufeaturename-dual-source-blending

Firefox downstream tracking: [bug 1924328](https://bugzilla.mozilla.org/show_bug.cgi?id=1924328)

teoxoy:
We already support dual-source blending (https://github.com/gfx-rs/wgpu/issues/1804) but I think our validation is not 1:1 with the spec.

ErichDonGubler:
@teoxoy: Ah, yes. I've renamed the issue to note that the scope is to converge with the spec.

jimblandy:
So the work here is:
- Figure out what the spec says
- Figure out what we're doing wrong, exactly

Teo suspects that our validation is too strict.

We are using the wrong names for things.

We don't have the `enable` implemented in the WGSL front end.

There are probably API bits too.

jimblandy:
Also, we're probably exposing this even when the feature isn't enabled.

Wumpf:
started looking into this because I need it for a toy project :)
Expect my first Naga PR soon 😄

Issue #6236, opened 2024-09-08
support `enable clip_distances` extension


EriKWDev:
**Is your feature request related to a problem? Please describe.**
I have implemented planar reflections in our engine which would benefit from HW clipping plane support but I currently have to use fragment-shader discard instead. Being able to set the `@builtin(clip-distances)` output of the vertex shader would fit this perfectly and hopefully be more performant

**Describe the solution you'd like**
support the `clip_distances` extension per https://gpuweb.github.io/gpuweb/#dom-gpufeaturename-clip-distances

**Describe alternatives you've considered**
Not supporting `clip_distances`, making me sad

**Additional context**
It seems to my understanding that parsing of the `enable` directive got mangled up with the implementation of f16 even though there are other extensions that could perhaps be worked on as well:

---

Remaining work:

- implementing `enable` and `fp16`: #5701 
- `clip-distances` discussion for spec: https://github.com/gpuweb/gpuweb/pull/4588
- enable directive issue: #5476
- https://github.com/gfx-rs/wgpu/pull/9267
- https://github.com/gfx-rs/wgpu/pull/8762
- https://github.com/gfx-rs/wgpu/pull/9270
  - Firefox bug to begin consuming this: https://bugzilla.mozilla.org/show_bug.cgi?id=1931629
- Remaining scope broken out to these remaining issues at time of closing:
  - https://github.com/gfx-rs/wgpu/issues/10393
  - https://github.com/gfx-rs/wgpu/issues/10394

inner-daemons:
I'll take a look at this and might implement it myself at some point.

ErichDonGubler:
@inner-daemons: Most, if not all, of this has been implemented. That's why #8762 exists. 😅

teoxoy:
Support for this is still missing in the D3D12 and Metal backends.

ErichDonGubler:
@teoxoy: In Metal's case, I think it's as simple as detecting support. I do see existing logic for `clip_distance` usage in Naga, though I'm not sure if it's complete.

I can find documented supported all the way back to [Metal 1.2](https://web.archive.org/web/20161029044059/https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf). Perhaps we could just unconditionally enable it?

ETA: Ah, looks like Dawn unconditionally enables it: https://dawn-review.googlesource.com/c/dawn/+/204436/10/src/dawn/native/metal/PhysicalDeviceMTL.mm

[...]

PR #6424, opened 2024-10-18
WGSL: Add a skeleton for enable-extensions/`enable …;` that reports nice errors

ErichDonGubler:
**Connections**

- Based on <https://github.com/gfx-rs/wgpu/pull/6352>.
- Paves the way for #5701, #6236, and #6402.

**Description**

This PR is much like #6352, but for the `enable` directive; it unblocks work on `enable` extensions in WGSL by serving as a foundation on which to build individual extension functionality. It does the following:

1. Build on #6352 by parsing an `enable` extension list after we encounter an `enable` in the directive parse path keyword, instead of emitting an unimplemented directive error.
2. Create an `enum` modeling all known extensions, including standard-but-unimplemented ones.
3. Add nice diagnostics for known and unknown extensions. If a known, but unimplemented, extension is requested, provide a hint that points users towards the appropriate Naga issue, so they can let us know they were blocked.

Note that this PR does not actually implement any extensions yet—see also the `UnimplementedEnableExtension::tracking_issue_num` method added in this PR for references to follow-up work for those.

**Testing**

- This functionality is exercised in #5701.
- More exhaustive tests will be added in <https://github.com/gfx-rs/wgpu/pull/6444>. 

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `taplo format`.

### DIFF
--- a/naga/src/front/wgsl/error.rs
+++ b/naga/src/front/wgsl/error.rs
@@ -1,3 +1,6 @@
+use crate::front::wgsl::parse::directive::enable_extension::{
+    EnableExtension, UnimplementedEnableExtension,
+};
 use crate::front::wgsl::parse::directive::{DirectiveKind, UnimplementedDirectiveKind};
 use crate::front::wgsl::parse::lexer::Token;
 use crate::front::wgsl::Scalar;
@@ -186,6 +189,7 @@ pub(crate) enum Error<'a> {
     UnknownType(Span),
     UnknownStorageFormat(Span),
     UnknownConservativeDepth(Span),
+    UnknownEnableExtension(Span, &'a str),
     SizeAttributeTooLow(Span, u32),
     AlignAttributeTooLow(Span, Alignment),
     NonPowerOfTwoAlignAttribute(Span),
@@ -274,6 +278,10 @@ pub(crate) enum Error<'a> {
     },
     DirectiveAfterFirstGlobalDecl {
         directive_span: Span,
+    },
+    EnableExtensionNotYetImplemented {
+        kind: UnimplementedEnableExtension,
+        span: Span,
     },
 }
 
@@ -524,6 +532,14 @@ impl<'a> Error<'a> {
                 message: format!("unknown type: '{}'", &source[bad_span]),
                 labels: vec![(bad_span, "unknown type".into())],
                 notes: vec![],
+            },
+            Error::UnknownEnableExtension(span, word) => ParseError {
+                message: format!("unknown enable-extension `{}`", word),
+                labels: vec![(span, "".into())],
+                notes: vec![
+                    "See available extensions at <https://www.w3.org/TR/WGSL/#enable-extension>."
+                        .into(),
+                ],
             },
             Error::SizeAttributeTooLow(bad_span, min_size) => ParseError {
                 message: format!("struct member size must be at least {min_size}"),
@@ -906,6 +922,28 @@ impl<'a> Error<'a> {
                     "maybe hoist this closer to the top of the shader module?"
                 )
                 .into()],
+            },
+            Error::EnableExtensionNotYetImplemented { kind, span } => ParseError {
+                message: format!(
+                    "the `{}` extension is not yet supported",
+                    EnableExtension::Unimplemented(kind).to_ident()
+                ),
+                labels: vec![(
+                    span,
+                    concat!(
+                        "this extension specifies standard functionality ",
+                        "which is not yet implemented in Naga"
+                    )
+                    .into(),
+                )],
+                notes: vec![format!(
+                    concat!(
+                        "Let Naga maintainers know that you ran into this at ",
+                        "<https://github.com/gfx-rs/wgpu/issues/{}>, ",
+                        "so they can prioritize it!"
+                    ),
+                    kind.tracking_issue_num()
+                )],
             },
         }
     }
--- a/naga/src/front/wgsl/error.rs
+++ b/naga/src/front/wgsl/error.rs
@@ -112,6 +112,8 @@ impl std::error::Error for ParseError {
 pub enum ExpectedToken<'a> {
     Token(Token<'a>),
     Identifier,
+    AfterIdentListComma,
+    AfterIdentListArg,
     /// Expected: constant, parenthesized expression, identifier
     PrimaryExpression,
     /// Expected: assignment, increment/decrement expression
@@ -345,6 +347,12 @@ impl<'a> Error<'a> {
                     ExpectedToken::Type => "type".to_string(),
                     ExpectedToken::Variable => "variable access".to_string(),
                     ExpectedToken::Function => "function name".to_string(),
+                    ExpectedToken::AfterIdentListArg => {
+                        "next argument, trailing comma, or end of list (',' or ';')".to_string()
+                    }
+                    ExpectedToken::AfterIdentListComma => {
+                        "next argument or end of list (';')".to_string()
+                    }
                 };
                 ParseError {
                     message: format!(
--- a/naga/src/front/wgsl/error.rs
+++ b/naga/src/front/wgsl/error.rs
@@ -283,6 +283,10 @@ pub(crate) enum Error<'a> {
         kind: UnimplementedEnableExtension,
         span: Span,
     },
+    EnableExtensionNotEnabled {
+        kind: EnableExtension,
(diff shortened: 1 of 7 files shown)


naga/src/front/wgsl/error.rs
```diff
@@ -899,6 +919,55 @@ impl<'a> Error<'a> {
...
+                            "but it is not currently enabled"
+                        ),
+                        kind.to_ident()
+                    )
+                    .into(),
+                )],
+                notes: if let EnableExtension::Unimplemented(kind) = kind {
+                    vec![format!(
+                        concat!(
+                            "This extension is not yet implemented.",
```
ErichDonGubler: nitpick(blocking): Need to add a space here:

```suggestion
                            "This extension is not yet implemented. ",
```

naga/src/front/wgsl/error.rs
```diff
@@ -899,6 +919,55 @@ impl<'a> Error<'a> {
                 )
                 .into()],
             },
+            Error::EnableExtensionNotYetImplemented { kind, span } => ParseError {
+                message: format!(
+                    "the `{}` extension is not yet supported (sorry!)",
```
ErichDonGubler: suggestion: We probably don't need the ` (sorry!)` parenthetical; it's generally more interesting to keep the opening description of the error succinct, IME.

```suggestion
                    "the `{}` extension is not yet supported",
```

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2480,14 +2508,28 @@ impl Parser {
...
         while let Ok((ident, span)) = lexer.peek_ident_with_span() {
             if let Some(kind) = DirectiveKind::from_ident(ident) {
                 self.push_rule_span(Rule::Directive, &mut lexer);
                 let _ = lexer.next_ident_with_span().unwrap();
                 match kind {
+                    DirectiveKind::Enable => {
+                        enable_extensions.add(self.enable_extension_ident(&mut lexer)?);
+                        while lexer.skip(Token::Separator(',')) {
+                            enable_extensions.add(self.enable_extension_ident(&mut lexer)?);
+                        }
```
ErichDonGubler: issue: This parsing of the comma-separated list is actually incorrect; we're not permitting a trailing comma here (CC @FL33TW00D, since this is their initial implementation). I think a human-friendly description of the parsing algorithm would be:

1. Parse a word token, and validate it as an extension we recognize. If we one of those steps fail, bail with the corresponding error.
2. Peek at input. If the next token is a comma, then consume it, and peek at the next token:
	1. If it's a word, go to step 1.
	1. If it's not a semicolon, bail with an error.
3. Parse a semicolon, or bail with an error.

I'll fix this in a bit.

naga/src/front/wgsl/parse/directive/enable_extension.rs
```diff
@@ -0,0 +1,112 @@
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct EnableExtensions {}
+
+impl EnableExtensions {
+    pub(crate) const fn empty() -> Self {
+        Self {}
+    }
+
+    /// Add an enable extension to the set requested by a module.
+    #[allow(unreachable_code)]
```
teoxoy, line 16:
```suggestion
```

naga/src/front/wgsl/parse/directive/enable_extension.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub(crate) fn add(&mut self, ext: ImplementedEnableExtension) {
+        let _field: &mut bool = match ext {};
+        *_field = true;
```
teoxoy, line 19: This doesn't seem to do anything but how does it compile? I thought `match ext {}` would return `()`.

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2258,6 +2269,34 @@ impl Parser {
...
+                    continue 'next_arg;
+                }
+                ExpectedToken::AfterIdentListComma
+            } else {
+                ExpectedToken::AfterIdentListArg
+            };
+
+            let (Token::Separator(';'), _span) = lexer.next() else {
+                return Err(Error::Unexpected(span, expected_token));
+            };
```
teoxoy: I find this use of `let` `else` quite confusing as it's only used for pattern matching. Can we use a `match` or `if matches!` instead? There is another instance of this above with `if let` `else`.

I'm not a fan of `if let` `else` because the thing being matched is after the pattern (reversed compared to a `match`) which disrupts the flow when reading the code. I find `let` `else` to be even worse in this regard because there is no indication that it's a condition unless you look for the `else`.

I see value in using `if let` `else` & `let` `else` when they are used for pattern matching + binding but I think there are better alternatives when pattern matching.

I can't seem to find a clippy lint for this. I usually don't comment on style but I find this is important.

teoxoy (changes requested):

naga/src/front/wgsl/parse/directive/enable_extension.rs
ErichDonGubler: `match`ing on an uninhabited `enum` actually returns `!`, which can be coerced to any type. The explicit type does the coercion, instead of letting it fall back to `()`.

naga/src/front/wgsl/parse/directive/enable_extension.rs
ErichDonGubler: This is necessary because of the `match ext {}` below.

Is this blocking?

naga/src/front/wgsl/parse/mod.rs
ErichDonGubler: Switched to `!matches!(…)` with 959517559ca9.

ErichDonGubler:
@teoxoy: This PR is exercised by #5701; any objection to punting tests to a separate PR, so we can keep yet another blocker for it resolved?

naga/src/front/wgsl/parse/mod.rs
teoxoy: What about the other instance of this above with `if let` `else`?

naga/src/front/wgsl/parse/directive/enable_extension.rs
teoxoy: Not really, I thought it's not necessary.

naga/src/front/wgsl/parse/directive/enable_extension.rs
teoxoy: As a reader of the code I find that surprising, do we need this code at all? I guess it doesn't really matter either way since the next PR will add the proper logic here right?

teoxoy:
That's fine with me.

naga/src/front/wgsl/parse/directive/enable_extension.rs
ErichDonGubler: @teoxoy: Yeah, the intent was to minimize the diff. between here and the next PR. However, it doesn't look like F16 extension support is going to land for a while, based on the momentum I perceive. We _could_ simply remove the body of this function for now, but 🤷🏻‍♂️ I honestly just consider it a logistic detail, at this point.

We could also add a test-only variant to ensure that this works as intended, but IDK if it's worth the effort over "just" implementing a real extension.

naga/src/front/wgsl/parse/mod.rs
ErichDonGubler: Ack, sorry, that's my lack of attention to detail, not anything malicious. 😖 Moved from `if-let-else` to `match` with 5702588b.

naga/src/front/wgsl/parse/directive/enable_extension.rs
teoxoy: I would prefer we remove the body but up to you.

teoxoy (approved):
LGTM

naga/src/front/wgsl/parse/mod.rs
ErichDonGubler: RE: Clippy lints: Actually, this is basically an application of [`clippy::single_match_else`](<https://rust-lang.github.io/rust-clippy/stable/index.html#/single_match_else>), which aligns with the preference I had expressed originally for reducing nesting.

Whatever, tho. 🤷🏻‍♂️🙂

PR #6427, opened 2024-10-20
Optimise GL fence checking by querying less

### LINKED ISSUE #5179: GL edge case, overhead of Fence::get_latest() quadratic over number of submits
Hey, we just encountered a weird edge case: during a single frame, we do a bunch (100+) of submits for draws to a texture target. We observed that with each submit, the list of pending fences here:

https://github.com/gfx-rs/wgpu/blob/d7bbababa7ffd2708e994df472aae27697de3bfa/wgpu-hal/src/gles/mod.rs#L603-L605

kept increasing and never got cleared (until the end of the frame). With each submit, this entire list gets iterated (note: `get_latest` is called twice during one submit) and checked with `getSyncParameter()`:

https://github.com/gfx-rs/wgpu/blob/d7bbababa7ffd2708e994df472aae27697de3bfa/wgpu-hal/src/gles/mod.rs#L626-L629

Which means, that the number of calls to `gl.getSyncParameter` is quadratic over number of submits.
This wasn't an issue on most platforms as this is supposed to be fast anyway - except for Firefox, where for some reason the IPC overhead of this call is nontrivial, and was enough to slow down the entire frame by an order of magnitude.

FF Profile:  https://share.firefox.dev/3HIrLby 
FF issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1875046

I don't know if this is actually actionable on wgpu side; my understanding is that usage that would have led to this situation, shouldn't be done in the first place. But still, thought that this is worth pointing out.

Dinnerbone:
**Connections**
This fixes https://github.com/gfx-rs/wgpu/issues/5179

**Description**
By keeping better track of which GL sync object last signalled, we can avoid repeatedly querying the status of them over and over. Additionally, bail early when encountering an unsignalled object as we know the remaining objects can't be signalled.

This has a noticable impact on performance when there are many `Queue::submit()`s per frame. For an extreme and contrived case, testing with the code below shows a reduction in frame time from 2 minutes to 2 seconds for me (with firefox, which seems to have a high cost of each sync check due to IPC)

_(It's likely we can apply the same optimisation to Vulkan's FencePool too, but, I doubt the cost is nearly as high there as WebGL.)_

**Testing**
Passes tests, and manually tested the impact by adding the following code to `hello_triangle`:
```rust
for _ in 0..1000 {
    let encoder = device.create_command_encoder(&Default::default());
    queue.submit(Some(encoder.finish()));
}
```

**Checklist**

- [X] Run `cargo fmt`.
- [ ] Run `taplo format`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -142,6 +142,7 @@ By @bradwerth [#6216](https://github.com/gfx-rs/wgpu/pull/6216).
 
 - Fix GL debug message callbacks not being properly cleaned up (causing UB). By @Imberflur in [#6114](https://github.com/gfx-rs/wgpu/pull/6114)
 - Fix calling `slice::from_raw_parts` with unaligned pointers in push constant handling. By @Imberflur in [#6341](https://github.com/gfx-rs/wgpu/pull/6341)
+- Optimise fence checking when `Queue::submit` is called many times per frame. By @dinnerbone in [#6427](https://github.com/gfx-rs/wgpu/pull/6427)
 
 #### WebGPU
 
--- a/wgpu-hal/src/gles/device.rs
+++ b/wgpu-hal/src/gles/device.rs
@@ -8,6 +8,7 @@ use std::{
     sync::{Arc, Mutex},
 };
 
+use crate::AtomicFenceValue;
 use arrayvec::ArrayVec;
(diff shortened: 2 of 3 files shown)


cwfitzgerald (approved):
Very nice improvement

PR #6431, opened 2024-10-21
[hlsl-out] Add support for restricting indexing to avoid OOB accesses

teoxoy:


naga/src/back/hlsl/writer.rs
```diff
@@ -2587,24 +2587,66 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
-                    let non_uniform_qualifier = match *resolved {
+                    let (indexing_binding_array, non_uniform_qualifier) = match *resolved {
                         TypeInner::BindingArray { .. } => {
                             let uniformity = &func_ctx.info[index].uniformity;
 
-                            uniformity.non_uniform_result.is_some()
+                            (true, uniformity.non_uniform_result.is_some())
                         }
-                        _ => false,
+                        _ => (false, false),
```
ErichDonGubler, line 2596: style: I think the new code here is harder understand (and easier to mess up) because these `bool`s are type-interchangeable. I don't think it's at a threshold that requires action yet, but I just wanted to call this out.

tests/tests/oob_indexing.rs
```diff
@@ -51,7 +51,11 @@ static RESTRICT_WORKGROUP_PRIVATE_FUNCTION_LET: GpuTestConfiguration = GpuTestCo
         drop(view);
         test_resources.readback_buffer.unmap();
 
-        assert_eq!([1; 12], current_res);
+        if ctx.adapter_info.backend == Backend::Dx12 {
```
ErichDonGubler, line 54: thought: Hmm, this is specific to FXC, per the below comment, right? Do we not have the ability to query which compiler we were using? 🤔 It might be worth getting that granularity and adding an issue for getting DXC coverage in CI.

wgpu-hal/src/dx12/mod.rs
```diff
@@ -953,6 +953,7 @@ impl crate::DynPipelineLayout for PipelineLayout {}
 pub struct ShaderModule {
     naga: crate::NagaShader,
     raw_name: Option<ffi::CString>,
+    runtime_checks: bool,
```
ErichDonGubler, line 956: suggestion: It'd be good to document or rename so that it's clear that this means that runtime checks were requested for compilation. I don't think it's clear, just from the naming, but I understand now that I've been reading how it's used.

wgpu-hal/src/metal/device.rs
```diff
@@ -117,7 +117,7 @@ impl super::Device {
         let ep_resources = &layout.per_stage_map[naga_stage];
 
         let bounds_check_policy = if stage.module.runtime_checks {
-            naga::proc::BoundsCheckPolicy::ReadZeroSkipWrite
+            naga::proc::BoundsCheckPolicy::Restrict
```
ErichDonGubler, line 120: question: I'm minorly confused by this change, though I think it's correct. Did we have an issue with MacOS, too?

ErichDonGubler (approved):
Approving pending resolution of conversations that are blocking (i.e., `question`s).

naga/src/back/hlsl/writer.rs
teoxoy: Ideally we would restrict binding arrays as well but I'm not sure about all details and didn't want to do it in this PR. I also don't think it's worth adding more types for this yet.

wgpu-hal/src/metal/device.rs
teoxoy: The other backends all use restrict and so the new test was failing since it's testing that the write to `container[len]` wrote to `container[len - 1]` (same for reads), read zero skip write is also more costly in terms of perf.

wgpu-hal/src/dx12/mod.rs
teoxoy: The other backends also have this field, I'm not sure what it should be renamed to, do you have a suggestion?

teoxoy: Maybe `inject_runtime_checks_for_indexing`?

tests/tests/oob_indexing.rs
teoxoy: We don't currently have the ability to tell which shader compiler we are using. I'm hoping that in the not too distant future we will drop support for FXC as it's not worth adding and maintaining workarounds for it.

wgpu-hal/src/dx12/mod.rs
ErichDonGubler: Oh, if this is already in other backends, then it's fine to be consistent as a first iteration. Might be interesting to clarify in follow-up work, though!

It's interesting that you reword the field as an imperative phrase. I thought it was indicative, i.e., "this was compiled with runtime checks". Maybe my understanding is wrong?

ErichDonGubler:
Just resolved all conversations. Happy merging! 🙂

wgpu-hal/src/dx12/mod.rs
teoxoy: The checks will be injected later when we call naga's backend. The module here is the IR generated by the WGSL frontend.

PR #6432, opened 2024-10-21
Add bounds check to Buffer slice method

### LINKED ISSUE #6428: Buffer slice method should panic when index is out of range (?)
**Description**
The slice method on Buffer does not check whether the argument is in a valid range for the buffer. In order to be consistent with the Array and Vec slicing methods of Rust, attempting to take an out-of-range slice on a buffer should panic. The slice method is defined here: https://github.com/gfx-rs/wgpu/blob/b3f665be7d9127d8048ec3b556e2a50868e951df/wgpu/src/api/buffer.rs#L243-L250

**Repro steps**
Create a buffer of size 8, then try buffer.slice(0..16).

The slice method should panic since the argument is out of range. Instead, it allows creating an out-of-range Buffer slice.

**Platform**
wgpu 22.1.0, Windows 11, Rust 1.82.0

beholdnec:
**Connections**
https://github.com/gfx-rs/wgpu/issues/6428

**Description**
The Buffer slice method currently doesn't perform bounds checks on the offset and size arguments. This is inconsistent with Rust's Array and Vec, which panic when attempting to take a slice out of bounds.

I recently encountered a confusing issue caused by attempting to map a buffer out of bounds. Making the Buffer slice method consistent with Rust's standard libraries by panicking in such cases would help debug such issues.

**Testing**
This PR includes test cases for the bounds-checking function, including tests for integer wraparound.

**Checklist**

- [X] Run `cargo fmt`.
- [ ] Run `taplo format`.

### DIFF
--- a/wgpu/src/api/buffer.rs
+++ b/wgpu/src/api/buffer.rs
@@ -242,6 +242,7 @@ impl Buffer {
     /// end of the buffer.
     pub fn slice<S: RangeBounds<BufferAddress>>(&self, bounds: S) -> BufferSlice<'_> {
         let (offset, size) = range_to_offset_size(bounds);
+        check_buffer_bounds(self.size, offset, size);
         BufferSlice {
             buffer: self,
             offset,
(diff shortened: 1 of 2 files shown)


wgpu/src/api/buffer.rs
```diff
@@ -673,6 +674,31 @@ impl Drop for Buffer {
...
+        panic!(
+            "slice offset {} is out of range for buffer of size {}",
+            offset, buffer_size
+        );
+    }
+
+    if let Some(size) = size {
+        // Detect integer overflow.
+        let end = offset.checked_add(size.get());
+        if end.is_none() || end.unwrap() > buffer_size {
```
Wumpf: nit: this avoids the unwrap
```suggestion
        if end.map_or(true, |end| end > buffer_size) {
```

Wumpf (approved):
very nicely done! Thank you so much!

PR #6451, opened 2024-10-23
refactor texture format to scalar conversion

atlv24:
**Connections**
commit pulled from #5537

**Description**
storage width of 4 bytes is hardcoded which makes adding 64bit texture formats annoying

**Testing**
existing tests pass

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `taplo format`.
- [x] Run `cargo clippy`. If applicable, add:
  - [x] `--target wasm32-unknown-unknown`
  - [x] `--target wasm32-unknown-emscripten`
- [x] Run `cargo xtask test` to run tests.
- [x] Add change to `CHANGELOG.md`. See simple instructions inside file.

PR #6453, opened 2024-10-23
Add R64Uint texture format

atlv24:
**Connections**
commits pulled from #5537

**Description**
add r64uint texture format, wgpu feature, naga capability, and adapter support detection for metal, directx, and vulkan

**Testing**
tests are in #5537

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `taplo format`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -136,7 +136,7 @@ winit = { version = "0.29", features = ["android-native-activity"] }
 # Metal dependencies
 block = "0.1"
 core-graphics-types = "0.1"
-metal = { version = "0.29.0" }
+metal = { version = "0.29.0", git = "https://github.com/gfx-rs/metal-rs.git", rev = "ae7030be0edff4cda88ece74137e5bcd28ef48fa" }
 objc = "0.2.5"
 
 # Vulkan dependencies
--- a/naga/src/back/hlsl/conv.rs
+++ b/naga/src/back/hlsl/conv.rs
@@ -122,8 +122,9 @@ impl crate::StorageFormat {
             Self::R16Float | Self::R32Float => "float",
             Self::R8Unorm | Self::R16Unorm => "unorm float",
             Self::R8Snorm | Self::R16Snorm => "snorm float",
-            Self::R8Uint | Self::R16Uint | Self::R32Uint | Self::R64Uint => "uint",
+            Self::R8Uint | Self::R16Uint | Self::R32Uint => "uint",
             Self::R8Sint | Self::R16Sint | Self::R32Sint => "int",
+            Self::R64Uint => "uint64_t",
 
             Self::Rg16Float | Self::Rg32Float => "float2",
             Self::Rg8Unorm | Self::Rg16Unorm => "unorm float2",
--- a/naga/src/back/hlsl/conv.rs
+++ b/naga/src/back/hlsl/conv.rs
@@ -122,7 +122,7 @@ impl crate::StorageFormat {
             Self::R16Float | Self::R32Float => "float",
             Self::R8Unorm | Self::R16Unorm => "unorm float",
             Self::R8Snorm | Self::R16Snorm => "snorm float",
-            Self::R8Uint | Self::R16Uint | Self::R32Uint => "uint",
+            Self::R8Uint | Self::R16Uint | Self::R32Uint | Self::R64Uint => "uint",
             Self::R8Sint | Self::R16Sint | Self::R32Sint => "int",
 
             Self::Rg16Float | Self::Rg32Float => "float2",
--- a/naga/src/proc/mod.rs
+++ b/naga/src/proc/mod.rs
@@ -66,7 +66,11 @@ impl From<super::StorageFormat> for super::Scalar {
             Sf::Rgba16Unorm => Sk::Float,
             Sf::Rgba16Snorm => Sk::Float,
         };
-        super::Scalar { kind, width: 4 }
+        let width = match format {
+            Sf::R64Uint => 8,
+            _ => 4,
+        };
+        super::Scalar { kind, width }
     }
 }
 
--- a/naga/src/proc/mod.rs
+++ b/naga/src/proc/mod.rs
@@ -49,6 +49,7 @@ impl From<super::StorageFormat> for super::ScalarKind {
             Sf::Rgb10a2Uint => Sk::Uint,
             Sf::Rgb10a2Unorm => Sk::Float,
             Sf::Rg11b10Ufloat => Sk::Float,
+            Sf::R64Uint => Sk::Uint,
             Sf::Rg32Uint => Sk::Uint,
             Sf::Rg32Sint => Sk::Sint,
             Sf::Rg32Float => Sk::Float,
--- a/wgpu-hal/src/dx12/adapter.rs
+++ b/wgpu-hal/src/dx12/adapter.rs
@@ -316,8 +316,6 @@ impl super::Adapter {
             | wgt::Features::RG11B10UFLOAT_RENDERABLE
             | wgt::Features::DUAL_SOURCE_BLENDING
             | wgt::Features::TEXTURE_FORMAT_NV12
-            // Ruint64 textures are always emulated on d3d12
-            | wgt::Features::TEXTURE_INT64_ATOMIC
             // float32-filterable should always be available on d3d12
             | wgt::Features::FLOAT32_FILTERABLE;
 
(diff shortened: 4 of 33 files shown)


wgpu-hal/src/metal/adapter.rs
```diff
@@ -828,6 +835,7 @@ impl super::PrivateCapabilities {
                 && ((device.supports_family(MTLGPUFamily::Apple8)
                     && device.supports_family(MTLGPUFamily::Mac2))
                     || device.supports_family(MTLGPUFamily::Apple9)),
+            int64_image_atomics: family_check && device.supports_family(MTLGPUFamily::Apple6),
```
teoxoy: Metal's feature set tables say this should be the same as the one above (`int64_atomics`).

wgpu-hal/src/metal/adapter.rs
```diff
@@ -1041,6 +1053,7 @@ impl super::PrivateCapabilities {
             Tf::Rgb10a2Uint => RGB10A2Uint,
             Tf::Rgb10a2Unorm => RGB10A2Unorm,
             Tf::Rg11b10Ufloat => RG11B10Float,
+            Tf::R64Uint => RG32Uint,
```
teoxoy, line 1056: Could you add a similar comment to the one in the D3D12 backend?

naga/src/back/hlsl/conv.rs
```diff
@@ -122,7 +122,7 @@ impl crate::StorageFormat {
             Self::R16Float | Self::R32Float => "float",
             Self::R8Unorm | Self::R16Unorm => "unorm float",
             Self::R8Snorm | Self::R16Snorm => "snorm float",
-            Self::R8Uint | Self::R16Uint | Self::R32Uint => "uint",
+            Self::R8Uint | Self::R16Uint | Self::R32Uint | Self::R64Uint => "uint",
```
teoxoy: Shouldn't this be `uint64_t`?

According to this it seems so https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_Int64_and_Float_Atomics.html

naga/src/proc/mod.rs
```diff
@@ -49,6 +49,7 @@ impl From<super::StorageFormat> for super::ScalarKind {
             Sf::Rgb10a2Uint => Sk::Uint,
             Sf::Rgb10a2Unorm => Sk::Float,
             Sf::Rg11b10Ufloat => Sk::Float,
+            Sf::R64Uint => Sk::Uint,
```
ErichDonGubler, line 52: question: Doesn't this conflict with #6451?

wgpu-hal/src/metal/adapter.rs
```diff
@@ -183,6 +189,7 @@ impl crate::Adapter for super::Adapter {
                 flags.set(Tfc::STORAGE, pc.format_rg11b10_all);
                 flags
             }
+            Tf::R64Uint => Tfc::COLOR_ATTACHMENT | Tfc::STORAGE | image_atomic,
```
teoxoy: Will this work as a color attachment?

wgpu-types/src/lib.rs
```diff
@@ -3440,6 +3458,7 @@ impl TextureFormat {
             Self::Rgb10a2Uint =>          (        msaa, attachment),
             Self::Rgb10a2Unorm =>         (msaa_resolve, attachment),
             Self::Rg11b10Ufloat =>        (        msaa,   rg11b10f),
+            Self::R64Uint =>              (        noaa, attachment),
```
teoxoy: This should probably be `TextureUsages::STORAGE_BINDING | TextureUsages::SHADER_ATOMIC`.

wgpu-hal/src/gles/conv.rs
```diff
@@ -50,6 +50,7 @@ impl super::AdapterShared {
                 glow::RGB,
                 glow::UNSIGNED_INT_10F_11F_11F_REV,
             ),
+            Tf::R64Uint => unreachable!(),
```
ErichDonGubler: issue: Seeing a panic in this conversion makes me uneasy. Is it possible to reach this path from safe code? If so, we should probably return a validation error of some sort instead.

ErichDonGubler (changes requested):
In addition to the other things I've raised here: `CHANGELOG` entry please!

wgpu-types/src/lib.rs
```diff
@@ -3561,6 +3580,7 @@ impl TextureFormat {
             | Self::Rg16Uint
             | Self::Rgba16Uint
             | Self::R32Uint
+            | Self::R64Uint
```
teoxoy, line 3584: We should probably return `None` here if this is only meant to work with atomic ops.

wgpu-types/src/lib.rs
```diff
@@ -3860,6 +3881,7 @@ impl TextureFormat {
             Self::R32Uint
             | Self::R32Sint
             | Self::R32Float
+            | Self::R64Uint
```
teoxoy, line 3885: Same here.

wgpu-types/src/lib.rs
```diff
@@ -3780,6 +3800,7 @@ impl TextureFormat {
             | Self::Rgba16Unorm
             | Self::Rgba16Snorm
             | Self::Rgba16Float
+            | Self::R64Uint
```
teoxoy, line 3804: Same here.

wgpu-hal/src/dx12/adapter.rs
```diff
@@ -315,7 +315,11 @@ impl super::Adapter {
             | wgt::Features::SHADER_PRIMITIVE_INDEX
             | wgt::Features::RG11B10UFLOAT_RENDERABLE
             | wgt::Features::DUAL_SOURCE_BLENDING
-            | wgt::Features::TEXTURE_FORMAT_NV12;
+            | wgt::Features::TEXTURE_FORMAT_NV12
+            // Ruint64 textures are always emulated on d3d12
+            | wgt::Features::TEXTURE_INT64_ATOMIC
```
teoxoy: Doesn't this require SM 6.6 though?

teoxoy (changes requested):

naga/src/proc/mod.rs
atlv24: it does not conflict, but now that its merged, width should be a match statement to return 8 for R64

teoxoy:
Please add an error when creating a texture with this format (for people trying to use this new feature) because there are currently no checks in naga that this format is only used with atomic ops (which are currently not implemented).

wgpu-hal/src/gles/conv.rs
atlv24: theres already a panic for NV12 down below, so theres some precedent, but im not sure if this is actually unreachable

wgpu-types/src/lib.rs
atlv24: regular sampling should work with this format

wgpu-hal/src/gles/conv.rs
atlv24: i've followed example set by BGRA8 and just did a best effort and left a `// TODO:` instead.

wgpu-types/src/lib.rs
teoxoy: Will it on the APIs we emulate it on?

Cargo.toml
```diff
@@ -136,7 +136,7 @@ winit = { version = "0.29", features = ["android-native-activity"] }
 # Metal dependencies
 block = "0.1"
 core-graphics-types = "0.1"
-metal = { version = "0.29.0" }
+metal = { version = "0.29.0", git = "https://github.com/gfx-rs/metal-rs.git", rev = "ae7030be0edff4cda88ece74137e5bcd28ef48fa" }
```
ErichDonGubler, line 139: issue: We have a release impending, and we can't release to Crates.io with Git dependencies. This needs to be resolved upstream; I'm going to investigate releasing the current mainline history of Metal to see if this can be resolved before a WGPU release.

ErichDonGubler (changes requested):
Changelog LGTM, but I noticed we have a Git dependency on `metal` (which we discussed in-meeting, just wanted to call that out here so it's not a surprise to anyone).

I encourage @teoxoy to not wait on resolution of the `metal` dep. blocker I've raised to keep moving on other feedback here.

wgpu-types/src/lib.rs
atlv24: I need it to, so consider that a blocking issue on the PR that actually implements this. Doesnt matter that it doesnt work with this PR because creating this texture errors every time anyways

Cargo.toml
ErichDonGubler: I have a PR open upstream: https://github.com/gfx-rs/metal-rs/pull/341

teoxoy:
We were hopeful to have https://github.com/gfx-rs/wgpu/pull/5537 as part of a minor release if we could get the breaking changes in before we release v23 but what I expected was that said breaking changes were refactors of existing code (https://github.com/gfx-rs/wgpu/pull/6451) and adding a few enum variants but this PR and https://github.com/gfx-rs/wgpu/pull/6459 are not doing just that, they are partially implementing the feature and I have a very hard time reviewing partial implementations.

As @ErichDonGubler also mentioned on matrix, we shouldn't wait for PRs to land in order to release (unless they solve critical issues).

Cargo.toml
ErichDonGubler: [`metal` v0.30.0](https://crates.io/crates/metal/0.30.0) is now released!

PR #6456, opened 2024-10-23
 WGSL: Add a skeleton for diagnostic filter rules/`diagnostic(…)`; that reports nice errors

ErichDonGubler:
**Connections**

- Part of #5320.
- Unblocks #6148.

**Description**

_**Recommended review experience is commit-by-commit. Each individual commit should pass CI!**_ 

Finishes building out a parsing skeleton for all directives (see also #6352, #6424, #6350). Paves the way for us to actually track diagnostic filters, and apply them.

**Testing**

Testing has been done manually to @ErichDonGubler's satisfaction. Test coverage for this parsing will be added in #6457.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `taplo format`.
- [x] Run `cargo clippy`. If applicable, add:
  - [x] `--target wasm32-unknown-unknown`
  - [x] `--target wasm32-unknown-emscripten`
- [x] Run `cargo xtask test` to run tests.
- [x] Add change to `CHANGELOG.md`. See simple instructions inside file.

PR #6464, opened 2024-10-25
feat(naga-cli)!: init. logger to `INFO` by default

ErichDonGubler:
**Connections**

- Discovered over the course of testing #6456, which needs to emit a warning.

**Description**

Many of Naga's internals print to `stderr` with `eprintln!(…)`. I'm sure that part of the reason is to avoid. I'm _also_ sure that users are missing vital information when Naga CLI doesn't emit warnings by default; how else would they become aware of them, if not being emitted by default?

**Testing**

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `taplo format`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1912,6 +1912,7 @@ dependencies = [
  "bincode",
  "codespan-reporting",
  "env_logger",
+ "log",
  "naga",
 ]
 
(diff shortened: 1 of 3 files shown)


ErichDonGubler:
If @gfx-rs/naga's disposition is to merge this, I'll be happy to go to the effort of a `CHANGELOG` entry.

jimblandy (approved):

PR #6472, opened 2024-10-28
add **/Cargo.lock

mahit56:
**Connections**

**Description**

**Testing**

**Checklist**

- [ ] Run `cargo fmt`.
- [ ] Run `taplo format`.


cwfitzgerald:
Why is this needed?

ErichDonGubler:
I can't speak for the OP (@mahit56, please respond to @cwfitzgerald's prompt!), but I think of a few reasons this sort of change could be good (with a tiny bit more work):

1. Rust upstream announced a change in official guidance on when to check in `Cargo.lock` with [this Rust blog post](https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html#why-the-change), which is now "just do it always, end users have the power to update deps. to newer versions under their own power if they need it".
2. This change will make it possible for `naga-cli` to be installed via `cargo install --locked`. We actually _do_ ship binaries from this repository, now that we've brought Naga in-tree, but we never adjusted to uphold official guidance for shipping binaries once that changed.
3. `Cargo.lock` updates from Dependabot tend to reduce downstream surprises, since it has a history and "sudden" changes in transitive deps. become easier to diagnose.
4. We need a checked-in `Cargo.lock` to do lockfile-only updating from Dependabot anyway.

In order for any of these arguments to be valid, though, we need to also actually generate and check in a resolved `Cargo.lock` file, which OP has not done. Happy to do that, if we get consensus on proceeding with a checked-in lockfile.

CC @gfx-rs/wgpu and @gfx-rs/naga for thoughts.

jimblandy:
Also, there is a comment that says:

> # Include the root lockfile but not the others

I guess Erich's comment explains what this was hinting at, but either way it needs to be updated.

ErichDonGubler:
Ah, actually, I think I've misunderstood the intent of this PR. I'm not sure how the existing `*/Cargo.lock` is distinct from `**/Cargo.lock`, except that it would also include the root.

I'm going to close this PR. If I've misunderstood something, please LMK, @mahit56!

PR #6479, opened 2024-10-30
[naga spv-out] Rename `write_expression_pointer` to `write_access_chain`

jimblandy:
This function's sole purpose is to generate `OpAccessChain` instructions, so there's no point in being coy about it.

### DIFF
--- a/naga/src/back/spv/block.rs
+++ b/naga/src/back/spv/block.rs
@@ -24,7 +24,7 @@ fn get_dimension(type_inner: &crate::TypeInner) -> Dimension {
 /// types are simply the direct SPIR-V analog of the Naga IR's. But in some
 /// cases, the Naga IR and SPIR-V types need to diverge.
 ///
-/// This enum specifies how [`BlockContext::write_expression_pointer`] should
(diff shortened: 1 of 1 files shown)


teoxoy (approved):

naga/src/back/spv/block.rs
```diff
@@ -24,7 +24,7 @@ fn get_dimension(type_inner: &crate::TypeInner) -> Dimension {
 /// types are simply the direct SPIR-V analog of the Naga IR's. But in some
 /// cases, the Naga IR and SPIR-V types need to diverge.
 ///
-/// This enum specifies how [`BlockContext::write_expression_pointer`] should
+/// This enum specifies how [`BlockContext::write_access_chain`] should
```
ErichDonGubler, line 27: @jimblandy: Heads up, docs weren't updated. I just added a fix via force-push to 
[7c7da5b](https://github.com/gfx-rs/wgpu/compare/4db573d7dfadcf6eeaa41b0384358cdb477771d1..7c7da5b74ba429065816608292489f51acbb1eb1).

PR #6480, opened 2024-10-30
fix(naga-cli): allow output files with `--stdin-file-path`

### LINKED ISSUE #6474: [naga-cli] Use of `--stdin-file-path` prevents writing output files
**Description**
It's impossible to translate stdin files with naga-cli and specify the output file. It looks like the switch for parsing the ``--stdin-file-path`` argument is only invoked if the first file argument is Some: ``args.files.first()``.  However, this means that it's actually impossible to have naga translate a file from stdin.

**Repro steps**
```
$cat file.wgsl | naga --input-kind wgsl --stdin-file-path file.wgsl file.spv
```

Using ``stdin-file-path`` should not cause naga to treat the first file name in the argument list as an input file.

ErichDonGubler:
**Connections**

- Fixes #6474.

**Description**

Change file handling in non-`bulk-validation` cases to be based on an iterator, and use it to correctly treat `args.files`' first entry as an input or output path in response to `--stdin-file-path` being specified.

**Testing**

Ran things manually with `--stdin-file-path` and without. I encourage reviewers to do their own testing, since 

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `taplo format`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,11 +40,12 @@ Bottom level categories:
 
 ## Unreleased
 
-## New Features
+### New Features
 
-### Naga
+#### Naga
 
 - Parse `diagnostic(…)` directives, but don't implement any triggering rules yet. By @ErichDonGubler in [#6456](https://github.com/gfx-rs/wgpu/pull/6456).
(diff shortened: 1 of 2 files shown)


jimblandy (approved):
Looks good to me!

PR #6482, opened 2024-10-31
[wgpu-hal] Change the `DropCallback` API to use `FnOnce` instead of `FnMut`

jerzywilczek:
**Connections**
This refines the change I made in #6164 

**Description**
I initially made the `DropCallback` API use `FnMut`. This seemed like a good choice at the time, since we only get `&mut self` access to the `DropGuard` in its `Drop` implementation. After some time it became apparent to me that at least a part of the users of this API will want to supply a closure which just holds some kind of a reference-counted wrapper around the corresponding resource that would drop it when called, thus freeing the resource. With the current API they have to do weird stuff to achieve this behavior, because `FnMut` cannot consume its captures:

```rs
let mut vulkan_device_clone: Option<Arc<VulkanDevice>> = Some(vulkan_device.clone());
let wgpu_device = unsafe {
    wgpu_adapter.adapter.device_from_raw(
        ...
        Some(Box::new(move || { vulkan_device_clone.take(); })),
        ...
    )?
};
``` 

With this change they could consume the capture:

```rs
let vulkan_device_clone: Arc<VulkanDevice> = vulkan_device.clone();
let wgpu_device = unsafe {
    wgpu_adapter.adapter.device_from_raw(
        ...
        Some(Box::new(move || { vulkan_device_clone; })),
        ...
    )?
};
// or more explicitly:
let wgpu_device = unsafe {
    wgpu_adapter.adapter.device_from_raw(
        ...
        Some(Box::new(move || { drop(vulkan_device_clone); })),
        ...
    )?
};
``` 

This seems a lot more intuitive.

**Testing**
 Things still compile and there is no difference in test passes.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `taplo format`.
  This just fails
129 tests fail on my machine, with or without the changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,12 @@ Bottom level categories:
 
 - Parse `diagnostic(…)` directives, but don't implement any triggering rules yet. By @ErichDonGubler in [#6456](https://github.com/gfx-rs/wgpu/pull/6456).
 
+### Changes
+
+#### HAL
+
+- Change the `DropCallback` API to use `FnOnce` instead of `FnMut`. By @jerzywilczek in [#6482](https://github.com/gfx-rs/wgpu/pull/6482)
+
 ## 23.0.0 (2024-10-25)
 
 ### Themes of this release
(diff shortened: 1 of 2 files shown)


cwfitzgerald (approved):
Nice!

PR #6483, opened 2024-11-01
naga: Fix textureNumLevels in the GLSL backend

### LINKED ISSUE #6435: Possibly incorrect WGSL->GLSL transformation: `textureNumLayers -> textureSize`
**Description**
I make use of `wgpu` transitively via `bevy`.

When I build my `bevy` app for WebGL2, my use of `textureNumLayers` is transformed to a use `textureSize` that does not seem to be compatible with GLSL 3.00. 

The usage in the source looks like this:

```wgsl
const MAX_MAP_LAYERS: u32 = 4;

@group(2) @binding(4) var map_texture: texture_2d_array<u32>;

fn sample_map(source_xy: vec2<u32>, length: vec2<f32>, layer_mask: u32, palette: vec3<u32>, uv: vec2<f32>) -> vec4<f32> {
    let map_layers: u32 = min(textureNumLayers(map_texture), MAX_MAP_LAYERS);
    // ...
}
```

And the resulting GLSL looks like this:

```glsl
vec4 sample_map(uvec2 source_xy_1, vec2 length_1, uint layer_mask, uvec3 palette_3, vec2 uv_2) {
    // ...
    uint map_layers = min(uint(textureSize(_group_2_binding_4_fs).z), MAX_MAP_LAYERS);
    // ...
}
```

Note that the resulting call to `textureSize` is only passed one argument. This produces an error that there is no matching overload for `textureSize`. Upon consulting the spec, `textureSize` does indeed seem to expect a second positional argument for the mip level:

```glsl
highp ivec2 textureSize (gsampler2D sampler, int lod)
highp ivec3 textureSize (gsampler3D sampler, int lod)
highp ivec2 textureSize (gsamplerCube sampler, int lod)

highp ivec2 textureSize (sampler2DShadow sampler, int lod)
highp ivec2 textureSize (samplerCubeShadow sampler, int lod)
highp ivec3 textureSize (gsampler2DArray sampler, int lod)
highp ivec3 textureSize (sampler2DArrayShadow sampler, int lod)
```

However, `textureNumLayers` does _not_ accept a similar second argument (https://www.w3.org/TR/WGSL/#texturenumlayers).

**Repro steps**

1. Use `textureNumLayers` in a WGSL shader

The resulting GLSL is compatible with WebGL2 in web browsers, however it is not.

**Extra materials**
Here is the full transformed output of the offending fragment shader (captured via Chrome dev tools): https://gist.github.com/cdata/22a420fee91854366e13f01983e04dfd

**Platform**
Chrome 129, Linux / Wayland
bevy 0.14.2, wgpu 0.20.1

magcius:
... and reactivate the GLSL test for it.

**Connections**
Fixes https://github.com/gfx-rs/wgpu/issues/6435.

**Description**

**Testing**

**Checklist**

- [x] Run `cargo fmt`.
- [ ] Run `taplo format`.

### DIFF
--- a/naga/src/back/glsl/mod.rs
+++ b/naga/src/back/glsl/mod.rs
@@ -3095,7 +3095,7 @@ impl<'a, W: Write> Writer<'a, W> {
                         self.write_expr(image, ctx)?;
                         // All textureSize calls requires an lod argument
                         // except for multisampled samplers
-                        if class.is_multisampled() {
+                        if !class.is_multisampled() {
                             write!(self.out, ", 0")?;
                         }
                         write!(self.out, ")")?;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ Bottom level categories:
 
 - Parse `diagnostic(…)` directives, but don't implement any triggering rules yet. By @ErichDonGubler in [#6456](https://github.com/gfx-rs/wgpu/pull/6456).
 - Fix an issue where `naga` CLI would incorrectly skip the first positional argument when `--stdin-file-path` was specified. By @ErichDonGubler in [#6480](https://github.com/gfx-rs/wgpu/pull/6480).
+- Fix textureNumLevels in the GLSL backend. By @magcius in [#6483](https://github.com/gfx-rs/wgpu/pull/6483).
 
 #### General
 
--- a/naga/tests/in/image.param.ron
+++ b/naga/tests/in/image.param.ron
@@ -3,5 +3,11 @@
 		version: (1, 1),
 		debug: true,
 	),
-	glsl_exclude_list: ["depth_load", "depth_no_comparison", "levels_queries"]
+	glsl: (
+		version: Desktop(430),
+		writer_flags: (""),
+		binding_map: {},
+		zero_initialize_workgroup_memory: true,
+	),
+	glsl_exclude_list: ["depth_load", "depth_no_comparison"]
 )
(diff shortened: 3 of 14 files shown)


ErichDonGubler (commented):
LGTM, but I'd like eyes from somebody else in @gfx-rs/wgpu before merging. I'm not familiar _at all_ with GL.

naga/src/back/glsl/mod.rs
```diff
@@ -3095,7 +3095,7 @@ impl<'a, W: Write> Writer<'a, W> {
                         self.write_expr(image, ctx)?;
                         // All textureSize calls requires an lod argument
                         // except for multisampled samplers
-                        if class.is_multisampled() {
+                        if !class.is_multisampled() {
```
teoxoy, line 3098: right 😅

naga/src/back/glsl/features.rs
```diff
@@ -124,11 +124,10 @@ impl FeaturesManager {
...
             _ => check_feature!(MULTI_VIEW, 140, 310),
         };
-        // Only available on glsl core, this means that opengl es can't query the number
-        // of samples nor levels in a image and neither do bound checks on the sample nor
-        // the level argument of texelFecth
+        // Only available on GLSL core, this means that OpenGL ES can't query the number
+        // of samples in a image and neither do bound checks on the sample.
         check_feature!(TEXTURE_SAMPLES, 150);
-        check_feature!(TEXTURE_LEVELS, 130);
+        check_feature!(TEXTURE_LEVELS, 130, 300);
```
teoxoy: `textureQueryLevels` doesn't seem to be available in es 3.

magcius: Wow. I had the ES 3.00 PDF open when I wrote this and I swear it was in there, but it's clearly not when I just checked right now.

teoxoy (approved):

teoxoy:
Thanks for the fix!

PR #6494, opened 2024-11-04
Fix Issue 6467: Prevent MacOS crash on invalid workgroup size definition

### LINKED ISSUE #6467: Invalid `dispatch_workgroups` arguments can crash MacOS
**Description**
If you set any of the workgroup sizes to 0, instead of just not running your compute shader, MacOS will freeze up and eventually the computer will reboot.

**Repro steps**
@cwfitzgerald said I didn't need to. 😉 

On Windows and Linux, the shader execution simply does nothing, but on MacOS, the machine crashes. It should probably also just do nothing, or only exit the application.

**Extra materials**

Here's a picture of the crashing code being stopped from crashing the machine when metal debugging is enabled:

IMG20241025141640

**Platform**
Running on the latest commit of `wgpu` on macOS Monterrey 12.7.6 on an old 2014 Mac Mini I got for testing purposes. :)

dfellis:
**Connections**
This should resolve #6467 but I haven't yet been able to get the test suite working on the Mac Mini the problem is occurring on.

**Description**
If you make a mistake and define invalid workgroup dimensions (basically zero any of them), on Vulkan and DX12 it just ignores your compute shader, but on Metal it triggers a bug that eventually freezes the machine and causes a reboot.

**Testing**
I added a regression test, but something with the test harness is preventing it from running. Running the full test suite also causes a freeze and reboot. I am putting this up here in case your CI system is able to test this on a Mac and we can see if the same behavior shows up or not.

**Checklist**

- [ ] Run `cargo fmt`.
- [ ] Run `taplo format`.

### DIFF
--- a/tests/tests/regression/issue_6467.rs
+++ b/tests/tests/regression/issue_6467.rs
@@ -9,12 +9,14 @@ use wgpu_test::{gpu_test, GpuTestConfiguration, TestParameters};
 /// The following test should successfully do nothing on all platforms.
 #[gpu_test]
 static ZERO_WORKGROUP_SIZE: GpuTestConfiguration = GpuTestConfiguration::new()
-    .parameters(TestParameters::default()
-        .limits(wgpu::Limits::default()))
+    .parameters(TestParameters::default().limits(wgpu::Limits::default()))
     .run_async(|ctx| async move {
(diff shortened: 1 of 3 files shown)


teoxoy (approved):
LGTM, thanks!

PR #6500, opened 2024-11-06
Make `Surface::as_hal` take an immutable reference to the surface.

jerzywilczek:
**Connections**
None

**Description**
`Surface::as_hal` right now takes a mutable reference to the surface. I don't understand why, and no documentation I found mentions why, so I propose to change that to take an immutable reference instead.

**Testing**
Everything still compiles and number of test passes doesn't change.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `taplo format`.
  This fails on my PC
129 tests fail on my PC, with or without the change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,10 @@ Bottom level categories:
 
 ### Changes
 
+#### General
+
+- Make `Surface::as_hal` take an immutable reference to the surface. By @jerzywilczek in [#9999](https://github.com/gfx-rs/wgpu/pull/9999)
+
 #### HAL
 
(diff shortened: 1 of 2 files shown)


teoxoy:
This is most likely an oversight.

teoxoy (approved):
Thanks!

xiaopengli89:
@jerzywilczek @teoxoy Mutable reference makes it possible to modify `present_with_transaction` on macOS, which is important for UI apps, my app is broken now, could we revert it?

cwfitzgerald:
@xiaopengli89 the callback only ever gave you a &A::Surface so reverting this wouldn't help. Maybe we need an as_hal_mut or something. Could you please file a new issue about this?

PR #6502, opened 2024-11-08
Dx12 HAL texture: expose raw resource

Tsuguri:
**Description**
In project I'm working on we want to temporarily use wgpu textures directly with dx12. Vulkan seems to have analogous access: https://github.com/gfx-rs/wgpu/blob/trunk/wgpu-hal/src/vulkan/mod.rs#L798
Device and Queue are available in similar fashion: https://github.com/gfx-rs/wgpu/blob/trunk/wgpu-hal/src/dx12/device.rs#L352

**Checklist**

- [x] Run `cargo fmt`.
- [ ] Run `taplo format`.

### DIFF
--- a/wgpu-hal/src/dx12/mod.rs
+++ b/wgpu-hal/src/dx12/mod.rs
@@ -777,6 +777,12 @@ pub struct Texture {
     allocation: Option<suballocation::AllocationWrapper>,
 }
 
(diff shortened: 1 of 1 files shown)


Wumpf (approved):
thank you 👍

Issue #6504, opened 2024-11-08
Query set creation appears to not respect platform limits


ErichDonGubler:
**Description**

WGPU panics on this line when some large number of query sets are requested with `Device::create_query_set` (exact threshold unknown still):

https://github.com/gfx-rs/wgpu/blob/6a75a73300053f6b6b786f3f650c135120139471/wgpu-core/src/device/resource.rs#L3608

**Repro steps**

<details><summary>MVRE `main.rs` based on `wgpu` 23.0.0:</summary>

```rust
use pollster::FutureExt as _;

fn main() {
    let instance = wgpu::Instance::default();

    let adapter = instance
        .request_adapter(&Default::default())
        .block_on()
        .unwrap();

    let (device, _queue) = adapter
        .request_device(
            &wgpu::DeviceDescriptor {
                required_features: wgpu::Features::all_webgpu_mask()
                    | wgpu::Features::TIMESTAMP_QUERY,
                ..Default::default()
            },
            None,
        )
        .block_on()
        .unwrap();

    // Kaboom!
    let _ = device.create_query_set(&wgpu::QuerySetDescriptor {
        ty: wgpu::QueryType::Timestamp,
        count: 8192,
        label: None,
    });
}
```

</details>

**Expected vs observed behavior**

We should not panic here, nor is even an internal error acceptable. We should be able to respect platform limits and enforce them via validation.

**Extra materials**

\-

**Platform**

Originally reported as a Firefox crash in [bug 1930015](https://bugzilla.mozilla.org/show_bug.cgi?id=1930015).

I was able to reproduce this on my M1 MacBook Pro. Current reports from also only originate from macOS devices. It is unclear whether this is reproducible in other environments.

- [x] ~~TODO: try to reproduce on other platforms~~ Didn't feel necessary after seeing [upstream discussion](https://github.com/gfx-rs/wgpu/issues/6504#issuecomment-2470206587).

ErichDonGubler:
Copying [my comment from bug 1930015](https://bugzilla.mozilla.org/show_bug.cgi?id=1930015):

> I was able to reproduce this crash stack on my M1 MacBook Pro by running the following script:
>
> ```js
> const adapter = await navigator.gpu.requestAdapter()
> const device = await adapter.requestDevice({ requiredFeatures: ["timestamp-query"] })
> device.createQuerySet({ type: "timestamp", count: 8192 })
> ```
>
[...]

PR #6505, opened 2024-11-08
fix(core): handle HAL errors from query set creation

ErichDonGubler:
**Connections**

- Makes #6504 less severe.

**Description**

When query set creation fails, return an internal error and lose the device, rather than panic.

**Testing**

\-

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `taplo format`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,10 @@ Bottom level categories:
 
 ### Bug Fixes
 
+#### General
+
+- Handle query set creation as an internal error that loses the `Device`, rather than panicking. By @ErichDonGubler in [#6505](https://github.com/gfx-rs/wgpu/pull/6505).
+
 #### Naga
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -64,6 +64,10 @@ Bottom level categories:
 
 ### Bug Fixes
 
+#### General
+
+- Handle query set creation as an internal error that loses the `Device`, rather than panicking. By @ErichDonGubler in [#6505](https://github.com/gfx-rs/wgpu/pull/6505).
```
Wumpf:
```suggestion
- Handle query set creation failure as an internal error that loses the `Device`, rather than panicking. By @ErichDonGubler in [#6505](https://github.com/gfx-rs/wgpu/pull/6505).
```

Wumpf (approved):

Wumpf:
even better would be ofc if we also handle all sort of things that could lead to that error. But "not crashing" is obviously a massive improvement

Issue #6506, opened 2024-11-08
naga: index out of bound in constant_evaluator.rs:217


wgslfuzz:
Translating the following wgsl with `naga a.wgsl a.hlsl` results in a panic on naga 23.0.0

```wgsl
fn a() {
    const b = vec4(f32(), f32());
    saturate(b);
}
```

```
thread 'main' panicked at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/naga-23.0.0/src/proc/constant_evaluator.rs:217:1:
index out of bounds: the len is 2 but the index is 2
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
````

ErichDonGubler:
I can reproduce on my machine. The issue is that `component_wise_*` functions in the `const_eval` module presume that a `Compose` expression of a `Vector` of N elements (in the OP example, `vec4`) has N arguments in the IR. `Splat` and `Zero` expressions (viz., single- and no-argument overloads of constructors) are expanded into this form, but manually written arguments might not necessarily be.

We can make this not panic, and I've done so with <https://github.com/gfx-rs/wgpu/pull/6508>. However, after examining code, I think we have two further issues with const.-evaluation of vector constructors where:

- The arg. count is less than N, but is a valid overload of the `Compose` constructor (like the overloads that offer a mix of scalars and vectors available for `vec3` and `vec4`). For example, this should be const.-evaluated properly:

	```wgsl
	fn func() {
		_ = saturate(vec4(vec2(1., 1.), 1., 1.)); // works
		_ = saturate(vec4(vec2(1.), 1., 1.)); // works
		_ = saturate(vec4(vec2(), 1., 1.)); // should be fine, but it's not 😭	
		_ = saturate(vec4(vec2(), vec2())); // should be fine, but it's not 😭	
	}
	```

	…but is not.
- This code responsible for flattening `Compose` expressions into the full N-arguments form:

	https://github.com/gfx-rs/wgpu/blob/64a61ee5c69569bbb3db03563997e88a229eba17/naga/src/proc/mod.rs#L806

	…`take`s only N expressions from the components of the `Compose` expression. This silently ignores all arguments after the Nth argument, which causes code like this to pass validation:

	```wgsl
	fn func() {
		_ = saturate(vec4(1., 1., 1., 1., 1., 1.)); // too many args 😭
	}
	```

teoxoy:
[...]

PR #6508, opened 2024-11-08
fix(const_eval): don't panic when a `vecN` constructor's arg. count is less than N

ErichDonGubler:
**Connections**

- Fixes the crash noted in #6506.

**Description**

From my comment in the issue:

> … `component_wise_*` functions in the `const_eval` module presume that a `Compose` expression of a `Vector` of N elements (in the OP example, `vec4`) has N arguments in the IR.

Stop presuming this, and return an error gracefully if the condition is violated.

**Testing**

Tested manually.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `taplo format`.


teoxoy (approved):
Nice!

PR #6510, opened 2024-11-10
Fix Vulkan surface capabilities being advertised when its query failed. 

Wumpf:
**Connections**
I'm sure there's related tickets, but it's hard to dig up :/

**Description**
Observed in the wild Vulkan adapter gets chosen despite the surface being not compatible (e.g. causing a Vulkan adapter to be chosen over a working GL adapter). This happend because Vulkan backend violated the contract of returning _usable_ surface configs.

Extending `GetSurfaceSupportError` is somewhat unrelated. More of a desperate attempt to output more information.

**Testing**
I unfortunately don't have a repro case for any of these failures.

**Checklist**

- [x] Run `cargo fmt`.
- [ ] Run `taplo format`.


teoxoy (approved):
Makes sense, good catch!

PR #6513, opened 2024-11-10
naga: Add support for GLSL signed/unsigned samplers

DavidPeicho:
**Description**

Add support for `usampler*` and `isampler` for GLSL input. Example of valid input:

```glsl
#version 440 core

layout(set = 0, binding = 0) uniform utexture2D tex;
layout(set = 0, binding = 1) uniform sampler samp;

layout(local_size_x = 8, local_size_y = 8) in;
void
main()
{
  vec4 data = vec4(texture(usampler2D(tex, samp), vec2(0.0, 1.0)));
}
```

**Testing**

Updated the test `samplers.frag`.

**Checklist**

- [X] Run `cargo fmt`.
- [ ] Run `taplo format`.
- [X] Run `cargo clippy`.


naga/src/front/glsl/builtins.rs
```diff
@@ -462,72 +462,85 @@ fn inject_standard_builtins(
...
-                            name,
-                            "sampler1DArray"
-                                | "sampler2DArray"
-                                | "sampler2DMSArray"
-                                | "samplerCubeArray"
-                        ),
-                        class: ImageClass::Sampled {
-                            kind: Sk::Float,
-                            multi: matches!(name, "sampler2DMS" | "sampler2DMSArray"),
+    let sampler_info = if name.starts_with("sampler") {
```
DavidPeicho: I am not really opinionated about the best way to fix it. Could also check for `sampler` on `&name[1..]` instead, and check the first char then.

naga/src/front/glsl/builtins.rs
```diff
@@ -462,72 +462,85 @@ fn inject_standard_builtins(
...
+                    ],
+                    MacroCall::Sampler,
+                ))
+            }
+            "sampler1DShadow"
+            | "sampler1DArrayShadow"
+            | "sampler2DShadow"
+            | "sampler2DArrayShadow"
+            | "samplerCubeShadow"
+            | "samplerCubeArrayShadow" => {
```
teoxoy: These don't have integer versions. We should probably match on `sampler_info`.

teoxoy (changes requested):

naga/src/front/glsl/builtins.rs
DavidPeicho: Or I guess just move them into the original match over `name`?

teoxoy: Sure.

DavidPeicho: Done, should we add some failing test check for unsigned/signed shadow samplers or? There are no failing/output error check tests I think for glsl yet

teoxoy (approved):
Thanks!

PR #6519, opened 2024-11-11
[naga] Implement quantizeToF16

### LINKED ISSUE #4875: [wgsl-in] Implement `quantizeToF16`
Implement the `quantizeToF16` built-in function from the WGSL spec.

jamienicol:
**Connections**
Fixes #4875 

**Description**
Adds `MathFunction::QuantiseToF16` and required typification and validation. Implement WGSL frontend and WGSL, SPIR-V, HLSL, MSL, and GLSL backends. WGSL and SPIR-V backends natively support the instruction. MSL and HLSL emulate it by casting to f16 and back to f32. GLSL does similar but must (mis)use `(un)pack2x16` to do so.

**Testing**
Manual inspection of `naga/tests/out/*/math-functions.*` files. Ran validation. Compiled gecko with changes and ran relevant CTS tests.

**Checklist**

- [x] Run `cargo fmt`.
- [ ] Run `taplo format`.

### DIFF
--- a/naga/src/back/glsl/mod.rs
+++ b/naga/src/back/glsl/mod.rs
@@ -1332,7 +1332,8 @@ impl<'a, W: Write> Writer<'a, W> {
                     crate::MathFunction::Pack4xI8
                     | crate::MathFunction::Pack4xU8
                     | crate::MathFunction::Unpack4xI8
-                    | crate::MathFunction::Unpack4xU8 => {
+                    | crate::MathFunction::Unpack4xU8
+                    | crate::MathFunction::QuantizeToF16 => {
                         self.need_bake_expressions.insert(arg);
                     }
                     crate::MathFunction::ExtractBits => {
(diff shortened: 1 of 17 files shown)


naga/src/front/wgsl/parse/conv.rs
```diff
@@ -255,6 +255,8 @@ pub fn map_standard_fun(word: &str) -> Option<crate::MathFunction> {
         "unpack2x16float" => Mf::Unpack2x16float,
         "unpack4xI8" => Mf::Unpack4xI8,
         "unpack4xU8" => Mf::Unpack4xU8,
+        // FIXME: move to correct category
+        "quantizeToF16" => Mf::QuantizeToF16,
```
jamienicol: unsure which category to add this under (if that wasn't obvious from the FIXMEs :smile: )

naga/src/back/glsl/mod.rs
```diff
@@ -3722,6 +3722,49 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                            self.write_expr(arg, ctx)?;
+                            write!(self.out, "))")?;
+                            return Ok(());
+                        }
+                        crate::TypeInner::Vector {
+                            size: crate::VectorSize::Tri,
+                            ..
+                        } => {
+                            write!(self.out, "vec3(unpackHalf2x16(packHalf2x16(")?;
+                            self.write_expr(arg, ctx)?;
```
jamienicol: I don't love this bit, as we write `arg` twice, which might be undesirable if arg is a complex expression? It might be nicer to create a `quantizeToF16` function that can use a temporary variable. I see we do that for modf and frexp, however that's because they need a custom return type whereas this just returns a float/vecN

naga/src/front/wgsl/parse/conv.rs
teoxoy: `computational` looks like the best fit to me. Arguably they are all computing things so, `computational` is more like the misc bucket.

naga/src/back/glsl/mod.rs
teoxoy: We usually mark the `arg` as baked, that will make sure that the expression gets written to a temporary variable that we can then use multiple times. Example:

https://github.com/gfx-rs/wgpu/blob/6a60458790b70b584c5e5cfd534d63a884aa0294/naga/src/back/glsl/mod.rs#L1336

naga/src/front/wgsl/parse/conv.rs
jamienicol: Sounds good thanks

naga/src/back/glsl/mod.rs
jamienicol: Ah perfect that works great

jamienicol:
Thanks @teoxoy ! I think this is ready for a proper review now

teoxoy (approved):
Looks great!

Issue #6283, opened 2024-09-16
Unable to pass an element of binding array as an argument to function


kvark:
**Description**
Passing a single, say, texture as an argument to a function works. Passing one from a specific index in an array doesn't.

**Repro steps**
```wgsl
fn foo(t: texture_2d<f32>) -> vec4<f32> {
    return textureLoad(t, vec2<i32>(0), 0);
}

@group(0) @binding(7)
var standalone: texture_2d<f32>;
foo(standalone); // this works

@group(0) @binding(6)
var textures_unbounded: binding_array<texture_2d<f32>>;
foo(textures_unbounded[0]); // this doesn't
```

**Expected vs observed behavior**
Expected to work. Instead, we are producing invalid SPIR-V
```
error: line 257: OpFunctionCall Argument <id> '147[%147]'s type does not match Function <id> '32[%_ptr_UniformConstant_18]'s parameter type.
  %148 = OpFunctionCall %v4float %foo %147
```
Specifically, we are issuing `OpLoad` on the indexed texture. But instead we should have been passing a pointer (result of `OpAccessChain`) to the function argument.

**Extra materials**
I believe the relevant logic is inside `fn cache_expression_value`, this block specifically:
```rust
// Only binding arrays in the Handle address space will take this path (due to `is_intermediate`)
                    crate::TypeInner::BindingArray { } => {... }
```

**Platform**
SPIR-V

Issue #4857, opened 2023-12-11
[naga/spv-in] spv-in can not consume SPIR-V where arrays of textures are passed as function arguments


chyyran:
**Description**
SPIR-V with an `OpFunctionCall` with an image array fails to compile.

**Repro steps**
Shader Playground link: https://shader-playground.timjones.io/d87861b3fcc78063347366c88651b276

1. Compile this GLSL to SPIR-V with `glslang`
```glsl
#version 450

layout(location = 0) out vec4 color;
layout(set = 0, binding = 1) uniform texture2D[2] tex;
layout(set = 0, binding = 2) uniform sampler _sampler;

void access(texture2D[2] tex1, sampler arg_sampler) {
    color = texture(sampler2D(tex1[0], arg_sampler), vec2(0.0));
}

void main() {
    access(tex, _sampler);
}
```

2. Use Naga to parse SPIR-V and it fails with `InvalidGlobalVar(FunctionArgument(0))`.

chyyran:
https://shader-playground.timjones.io/2d02a87f59ed403ffef77b83d45a9b51

Putting the `OpAccessLoad` within the outer function also does not work

```glsl
#version 450

layout(location = 0) out vec4 color;
layout(set = 0, binding = 1) uniform texture2D[2] tex;
layout(set = 0, binding = 2) uniform sampler _sampler;

void access(texture2D tex1, sampler arg_sampler) {
    color = texture(sampler2D(tex1, arg_sampler), vec2(0.0));
}

void main() {
    access(tex[0], _sampler);
}
```

PR #6523, opened 2024-11-12
[naga] Implement binding_array function arguments

davnotdev:
**Connections**
Fixes both cases described in #4857.
Also related to #6283.

**Description**
There are two cases addressed here:

1. Allowing binding arrays to appear as function arguments by implementing various missing branches that this case triggers.
2. Fix issue where `OpFunctionCall` uses the value of an opaque type loaded via `OpAccessChain` + `OpLoad` is used as the argument when only the pointer to that opaque type is valid. This is implemented by keeping track of the pointer id of a loaded variable when `OpAccessChain` is called.

I'm not two happy with the 2nd fix because it uses an extra map for a relatively obscure edge case. Hopefully, there is a function that does the equivalent.

**Testing**
To test these changes, I ran the following shaders through naga multiple times (`SPV => spirv-val => WGSL => SPV => spirv-val => WGSL => SPV => spirv-val`).

The cases described in the WGPU issue above:
https://shader-playground.timjones.io/8a6382938dcc19b6d86bb569a93f1872
https://shader-playground.timjones.io/e37907fe8b30779a53201ef5ba4e1809

The issue I am running into:
https://shader-playground.timjones.io/656df9d61201d86379fe5c6f6b04d43a
(For this shader, the assumption is that `SPV_EXT_descriptor_indexing` is removed after every translation.)

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `taplo format`.

### DIFF
--- a/naga/src/back/mod.rs
+++ b/naga/src/back/mod.rs
@@ -264,6 +264,7 @@ impl crate::TypeInner {
             crate::TypeInner::Image { .. }
             | crate::TypeInner::Sampler { .. }
             | crate::TypeInner::AccelerationStructure { .. } => true,
+            crate::TypeInner::BindingArray { .. } => true,
             _ => false,
         }
     }
--- a/naga/src/back/spv/block.rs
+++ b/naga/src/back/spv/block.rs
@@ -423,8 +423,19 @@ impl<'w> BlockContext<'w> {
                         };
 
                         let binding_type_id = self.get_type_id(LookupType::Handle(binding_type));
-
                         let load_id = self.gen_id();
+
+                        // Map `binding_type_id` back to the original pointer if it is an opaque
+                        // type.
+                        match self.ir_module.types[binding_type].inner {
+                            crate::TypeInner::Image { .. }
+                            | crate::TypeInner::Sampler { .. }
+                            | crate::TypeInner::AccelerationStructure => {
+                                self.function_arg_ids.insert(load_id, result_id);
+                            }
+                            _ => {}
+                        }
+
                         block.body.push(Instruction::load(
                             binding_type_id,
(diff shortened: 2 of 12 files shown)


naga/src/back/mod.rs
```diff
@@ -264,6 +264,7 @@ impl crate::TypeInner {
             crate::TypeInner::Image { .. }
             | crate::TypeInner::Sampler { .. }
             | crate::TypeInner::AccelerationStructure { .. } => true,
+            crate::TypeInner::BindingArray { .. } => true,
```
jimblandy, line 267: I don't think this is right. The documentation for `TypeInner::BindingArray` says:

> The array's element type `base` may be [`Image`], [`Sampler`], or any type that would be permitted for a global in the [`Uniform`] or [`Storage`] address spaces.

So just because `TypeInner` is a `BindingArray` doesn't mean it's a handle.

jimblandy (changes requested):
These changes need to be covered by snapshot tests; see `naga/tests/snapshot.rs`, in particular:
- `convert_wgsl`, which converts WGSL inputs to all possible output formats. Please hand-check the `.spvasm` to ensure you're getting what you expect.
- `convert_spv_all`, which converts SPV inputs into WGSL. You should provide an `.spvasm` file to accompany each `.spv` input that is exactly what `spirv-dis` outputs when run on that `.spv` file.

We don't do every combination of input and output formats, because that's just too much noise, just WGSL->everything and everything->WGSL.

Please flag for re-review when these issues are addressed.

jimblandy:
We haven't heard from the author for a while, so we're closing this as stale. But please feel free to re-open it if you're able to work on it more!

PR #6529, opened 2024-11-12
[wgsl-in] Ensure textureSampleLevel's level argument is an integer for depth textures

### LINKED ISSUE #4548: `textureSampleLevel(texture_depth_2d_array, ..)` expects a float for level when it should be an integer
Hi there!

We've recently updated to wgpu 0.16 and are trying to fix regressions in our code. We encountered gfx-rs/naga#1218 and are trying to fix it by switching to `textureSampleLevel`:

```wgsl
let screen_depth = textureSample(solids_screen_depth, default_sampler, screen_tc);
if pos_ndc.z >= screen_depth && pos_ndc.z < screen_depth * 1.001 && screen_depth < 0.999 {
    return textureSample(solids_screen_color, default_sampler, screen_tc).rgb;
}
```
to
```wgsl
let screen_depth = textureSampleLevel(solids_screen_depth, default_sampler, screen_tc, 0u);
if pos_ndc.z >= screen_depth && pos_ndc.z < screen_depth * 1.001 && screen_depth < 0.999 {
    return textureSampleLevel(solids_screen_color, default_sampler, screen_tc, 0.0).rgb;
}
```

This works just fine on the web (using Chrome + Tint), but fails to compile on desktop:
```
Shader validation error: 
     ┌─ standard_shader_water_shader_true:1233:1
     │  
1233 │ ╭ fn screen_space_reflections(world_position: vec3<f32>, screen_ray_dir: vec3<f32>, normal: vec3<f32>, screen_size: vec2<f32>, screen_position: vec2<f32>) -> vec3<f32> {
1234 │ │     var reflected_dir = reflect(screen_ray_dir, normal);
1235 │ │     reflected_dir.z = max(0., reflected_dir.z);
1236 │ │ 
     · │
1254 │ │         let screen_depth = textureSampleLevel(solids_screen_depth, default_sampler, screen_tc, 0u);
     │ │                            ^^^^^^^^^^^^^^^^^^ naga::Expression [66]
     · │
1263 │ │     let sky = get_sky_color(global_params.camera_far, world_position, reflected_dir);
1264 │ │     return sky;
     │ ╰───────────────^ naga::Function [93]


    Function [93] 'screen_space_reflections' is invalid
    Expression [66] is invalid
    Sample level (exact) type [65] is not a scalar float
```

It appears that Naga believes the level for a depth texture is floating-point (as is the case for regular textures), but the WebGPU spec defines it as an integer:

| Parameterization | Overload |
| - | - |
| `L` is `i32`, or `u32` | `@must_use fn textureSampleLevel(t: texture_depth_2d, s: sampler, coords: vec2<f32>, level: L) -> f32` |

(from <https://www.w3.org/TR/WGSL/#texturesamplelevel>)

~~Is there a way around this without patching Naga so that we can maintain compatibility with both desktop and web? I'm content to do conditional compilation or such; I'm just not sure if that's something we need to do, or if Naga/WGSL already have some way to do so.~~ We already have a preprocessor, we'll use it to switch the zero used based on platform.

---

P.S. As suggested in gfx-rs/naga#1218, I tried lifting the `textureSample` out of the `if` (so that the sampling always occurs), which works on desktop but not on web:
```wgsl
let screen_depth = textureSample(solids_screen_depth, default_sampler, screen_tc);
let screen_color = textureSample(solids_screen_color, default_sampler, screen_tc).rgb;
if pos_ndc.z >= screen_depth && pos_ndc.z < screen_depth * 1.001 && screen_depth < 0.999 {
    return screen_color;
}
```

```
localhost/:1 1 error(s) generated while compiling the shader:
:1254:28 error: 'textureSample' must only be called from uniform control flow
        let screen_depth = textureSample(solids_screen_depth, default_sampler, screen_tc);
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

:1256:38 note: control flow depends on possibly non-uniform value
        if pos_ndc.z >= screen_depth && pos_ndc.z < screen_depth * 1.001 && screen_depth < 0.999 {
                                     ^^

:1254:28 note: return value of 'textureSample' may be non-uniform
        let screen_depth = textureSample(solids_screen_depth, default_sampler, screen_tc);
```

The following also has the same issue:
```wgsl
let screen_depth = textureSample(solids_screen_depth, default_sampler, screen_tc);
if pos_ndc.z >= screen_depth && pos_ndc.z < screen_depth * 1.001 && screen_depth < 0.999 {
    return textureSampleLevel(solids_screen_color, default_sampler, screen_tc, 0.0).rgb;
}
```

It would appear that the `screen_depth` needs to be level-sampled in order to be used in a conditional (which makes sense), but this results in the original issue here.

jamienicol:
**Connections**
Fixes #4548 

**Description**
The [spec](https://www.w3.org/TR/WGSL/#texturesamplelevel) specifies that `textureSampleLevel()` takes an integer type as the `level` parameter for depth textures (and a float for non-depth textures). We currently accept a float for all texture types. This makes us require the correct type. As a result we now need to convert the value in the SPIR-V backend. For HLSL and MSL, `SampleLevel()` and `metal::level()` appear to implicitly convert the integer to floats just fine. And as far as I can tell we do not support non-_compare_ sampling of depth textures on GLSL?

**Testing**
Manual inspection of output shaders. Shader validation succeeds. Built gecko with change and verified relevant CTS tests (`webgpu:shader,validation,expression,call,builtin,textureSampleLevel:level_argument`) now pass.

**Checklist**

- [x] Run `cargo fmt`.
- [ ] Run `taplo format`.

### DIFF
--- a/naga/src/back/spv/image.rs
+++ b/naga/src/back/spv/image.rs
@@ -924,7 +924,43 @@ impl<'w> BlockContext<'w> {
                     depth_id,
                 );
 
-                let lod_id = self.cached[lod_handle];
+                let mut lod_id = self.cached[lod_handle];
+                // SPIR-V expects the LOD to be a float for all image classes.
+                // lod_id, however, will be an integer for depth images,
+                // therefore we must do a conversion.
+                if matches!(
+                    self.ir_module.types[image_type].inner,
+                    crate::TypeInner::Image {
+                        class: crate::ImageClass::Depth { .. },
+                        ..
+                    }
+                ) {
+                    let lod_f32_id = self.gen_id();
+                    let f32_type_id = self.get_type_id(LookupType::Local(LocalType::Numeric(
+                        NumericType::Scalar(crate::Scalar::F32),
+                    )));
+                    let convert_op = match *self.fun_info[lod_handle]
+                        .ty
+                        .inner_with(&self.ir_module.types)
+                    {
+                        crate::TypeInner::Scalar(crate::Scalar {
+                            kind: crate::ScalarKind::Uint,
+                            width: 4,
+                        }) => spirv::Op::ConvertUToF,
+                        crate::TypeInner::Scalar(crate::Scalar {
+                            kind: crate::ScalarKind::Sint,
+                            width: 4,
+                        }) => spirv::Op::ConvertSToF,
+                        _ => unreachable!(),
+                    };
+                    block.body.push(Instruction::unary(
+                        convert_op,
+                        f32_type_id,
+                        lod_f32_id,
+                        lod_id,
+                    ));
+                    lod_id = lod_f32_id;
+                }
                 mask |= spirv::ImageOperands::LOD;
                 inst.add_operand(mask.bits());
                 inst.add_operand(lod_id);
--- a/naga/src/front/spv/image.rs
+++ b/naga/src/front/spv/image.rs
@@ -30,6 +30,7 @@ impl<'function> super::BlockContext<'function> {
         match self.expressions[handle] {
             crate::Expression::GlobalVariable(handle) => Ok(self.global_arena[handle].ty),
             crate::Expression::FunctionArgument(i) => Ok(self.arguments[i as usize].ty),
+            crate::Expression::Access { base, .. } => Ok(self.get_image_expr_ty(base)?),
             ref other => Err(Error::InvalidImageExpression(other.clone())),
         }
     }
@@ -460,6 +461,7 @@ impl<I: Iterator<Item = u32>> super::Frontend<I> {
         } else {
             None
         };
+        let span = self.span_from_with_op(start);
 
         let mut image_ops = if words_left != 0 {
             words_left -= 1;
@@ -486,9 +488,34 @@ impl<I: Iterator<Item = u32>> super::Frontend<I> {
                     let lod_lexp = self.lookup_expression.lookup(lod_expr)?;
                     let lod_handle =
                         self.get_expr_handle(lod_expr, lod_lexp, ctx, emitter, block, body_idx);
+
+                    let is_depth_image = {
+                        let image_lexp = self.lookup_sampled_image.lookup(sampled_image_id)?;
+                        let image_ty = ctx.get_image_expr_ty(image_lexp.image)?;
+                        matches!(
+                            ctx.type_arena[image_ty].inner,
+                            crate::TypeInner::Image {
+                                class: crate::ImageClass::Depth { .. },
+                                ..
+                            }
+                        )
+                    };
+
                     level = if options.compare {
                         log::debug!("Assuming {:?} is zero", lod_handle);
                         crate::SampleLevel::Zero
+                    } else if is_depth_image {
+                        log::debug!(
+                            "Assuming level {:?} converts losslessly to an integer",
+                            lod_handle
+                        );
+                        let expr = crate::Expression::As {
+                            expr: lod_handle,
+                            kind: crate::ScalarKind::Sint,
+                            convert: Some(4),
+                        };
+                        let s32_lod_handle = ctx.expressions.append(expr, span);
+                        crate::SampleLevel::Exact(s32_lod_handle)
                     } else {
                         crate::SampleLevel::Exact(lod_handle)
                     };
--- a/naga/src/valid/expression.rs
+++ b/naga/src/valid/expression.rs
@@ -106,7 +106,7 @@ pub enum ExpressionError {
     InvalidGatherComponent(crate::SwizzleComponent),
     #[error("Gather can't be done for image dimension {0:?}")]
     InvalidGatherDimension(crate::ImageDimension),
-    #[error("Sample level (exact) type {0:?} is not a scalar float")]
+    #[error("Sample level (exact) type {0:?} has an invalid type")]
     InvalidSampleLevelExactType(Handle<crate::Expression>),
     #[error("Sample level (bias) type {0:?} is not a scalar float")]
     InvalidSampleLevelBiasType(Handle<crate::Expression>),
@@ -530,11 +530,24 @@ impl super::Validator {
                     crate::SampleLevel::Auto => ShaderStages::FRAGMENT,
                     crate::SampleLevel::Zero => ShaderStages::all(),
                     crate::SampleLevel::Exact(expr) => {
-                        match resolver[expr] {
-                            Ti::Scalar(Sc {
-                                kind: Sk::Float, ..
-                            }) => {}
-                            _ => return Err(ExpressionError::InvalidSampleLevelExactType(expr)),
+                        match class {
+                            crate::ImageClass::Depth { .. } => match resolver[expr] {
+                                Ti::Scalar(Sc {
+                                    kind: Sk::Sint | Sk::Uint,
+                                    ..
+                                }) => {}
+                                _ => {
+                                    return Err(ExpressionError::InvalidSampleLevelExactType(expr))
+                                }
+                            },
+                            _ => match resolver[expr] {
+                                Ti::Scalar(Sc {
+                                    kind: Sk::Float, ..
+                                }) => {}
+                                _ => {
+                                    return Err(ExpressionError::InvalidSampleLevelExactType(expr))
+                                }
+                            },
                         }
                         ShaderStages::all()
                     }
--- a/naga/tests/out/hlsl/image.hlsl
+++ b/naga/tests/out/hlsl/image.hlsl
@@ -369,5 +369,6 @@ float4 depth_no_comparison() : SV_Target0
     float2 tc_3 = (0.5).xx;
     float s2d_1 = image_2d_depth.Sample(sampler_reg, tc_3);
     float4 s2d_gather = image_2d_depth.Gather(sampler_reg, tc_3);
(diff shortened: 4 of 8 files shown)


naga/src/back/spv/image.rs
```diff
@@ -924,7 +924,43 @@ impl<'w> BlockContext<'w> {
...
+                        }) => spirv::Op::ConvertSToF,
+                        _ => unreachable!(),
+                    };
+                    block.body.push(Instruction::unary(
+                        convert_op,
+                        f32_type_id,
+                        lod_f32_id,
+                        lod_id,
+                    ));
+                    lod_id = lod_f32_id;
```
jamienicol, line 962: I would love to know if there's a better way of doing this?

teoxoy: There isn't, at least not right now, which is why the SPIR-V backend is the biggest one 😅

naga/tests/out/hlsl/image.hlsl
```diff
@@ -369,5 +369,6 @@ float4 depth_no_comparison() : SV_Target0
     float2 tc_3 = (0.5).xx;
     float s2d_1 = image_2d_depth.Sample(sampler_reg, tc_3);
     float4 s2d_gather = image_2d_depth.Gather(sampler_reg, tc_3);
-    return ((s2d_1).xxxx + s2d_gather);
+    float s2d_level = image_2d_depth.SampleLevel(sampler_reg, tc_3, 1);
```
teoxoy, line 372: > For HLSL and MSL, `SampleLevel()` and `metal::level()` appear to implicitly convert the integer to floats just fine.

I worry that the way we test it here works because the `1` gets interpreted as a float. If you change the input WGSL to be `var level = 1;` instead of `let level = 1;`, does it still work? That should cause the variable to show up in the output.

teoxoy:
> And as far as I can tell we do not support non-compare sampling of depth textures on GLSL?

Looking at the code in the backend I think we do but there are caveats/restrictions.

naga/src/valid/expression.rs
```diff
@@ -521,11 +521,24 @@ impl super::Validator {
...
+                        match class {
+                            crate::ImageClass::Depth { .. } => match resolver[expr] {
+                                Ti::Scalar(Sc {
+                                    kind: Sk::Sint | Sk::Uint,
+                                    ..
+                                }) => {}
+                                _ => {
+                                    return Err(ExpressionError::InvalidSampleLevelExactType(expr))
+                                }
+                            },
```
teoxoy, line 542: Ideally we would modify the GLSL & SPIR-V frontends as well to not regress. I think we need to round the lod then cast it to a u32 so that the behavior is in line with a non-filtering sampler.

naga/tests/out/hlsl/image.hlsl
jamienicol: That does indeed cause the variable to show up in the output, but both metal and dxc are still fine with that.

teoxoy: Ok, if that works then I guess it's fine. I didn't know HLSL and MSL did automatic conversions.

naga/src/valid/expression.rs
jamienicol: I don't think any GLSL command can be converted to a depth texture `textureSampleLevel()`. GLSL's `textureLod()` with a shadow texture takes a vector with an extra component for the coords, and this component is used as the `compare` value when translated to `textureSampleCompareLevel()`.

textureSampleCompareLevel always has a LOD of zero, so the LOD specified in the GLSL also appears to be lost in translation. I think that occurs [here](https://github.com/gfx-rs/wgpu/blob/a2044aefe7fd0dc4be42684d3bded2c645dc5d92/naga/src/front/glsl/builtins.rs#L1653). So as I can tell I don't think there's anything to do for GLSL, but this is all very new to me so could be wrong!

jamienicol:
> > And as far as I can tell we do not support non-compare sampling of depth textures on GLSL?
> 
> Looking at the code in the backend I think we do but there are caveats/restrictions.

@teoxoy Could you point me towards this? I'm not very familiar with these things, but my understanding is GLSL's `textureLod()` or just `texture()` take a vec with an extra component when using a shadow sampler. This extra component is the compare value. We have code to write this in glsl back [here](https://github.com/gfx-rs/wgpu/blob/a2044aefe7fd0dc4be42684d3bded2c645dc5d92/naga/src/back/glsl/mod.rs#L2906) but only when `depth_ref` is Some, ie using `textureSampleCompare`.

If I try to translate some wgsl with `textureSampleLevel()` then we generate GLSL with a vec of the wrong size, as we didn't merge the depth_ref

naga/src/valid/expression.rs
teoxoy: Looking at the GLSL frontend, that seems accurate. The SPIR-V frontend can produce a `textureSampleLevel()` with a depth texture though.

teoxoy:
Oh wow that's a "nice" inconsistency between GLSL's sampling functions, some take the depth ref as a separate argument (ex `textureGather`), so I never expected some to take it as the last element of the vec. It also doesn't seem to be documented in the specs.

On the spec side of things, we are also working on [WebGPU Compat](https://github.com/gpuweb/gpuweb/blob/main/proposals/compatibility-mode.md) (WebGPU on top of OpenGL ES 3.1) and this hasn't been brought up by anyone yet, I will open a new issue.

naga/src/front/spv/image.rs
```diff
@@ -486,9 +488,34 @@ impl<I: Iterator<Item = u32>> super::Frontend<I> {
...
+                    } else if is_depth_image {
+                        log::debug!(
+                            "Assuming level {:?} converts losslessly to an integer",
+                            lod_handle
+                        );
+                        let expr = crate::Expression::As {
+                            expr: lod_handle,
+                            kind: crate::ScalarKind::Sint,
+                            convert: Some(4),
+                        };
```
teoxoy, line 516: I think that if we `round` it first this will behave the same as the sampler doing it. Thoughts?

jamienicol: My assumption was that a fractional lod interpolated between two mip levels. But I can't see anything in the spec which says that (but nor that it rounds the lod). Do you happen to know what the spec says?

teoxoy: Depth textures can only be paired with a non-filtering (or comparison) sampler which means that the `mipmapFilter` field of the sampler descriptor can't be `linear`, it must be `nearest` which sounds like it `round`s instead.

A few links:

https://www.w3.org/TR/webgpu/#ref-for-dom-gpusamplerbindingtype-filtering%E2%91%A2
https://www.w3.org/TR/webgpu/#dom-gpusamplerbindingtype-non-filtering
https://www.w3.org/TR/webgpu/#ref-for-dom-gpusampler-isfiltering-slot
https://www.w3.org/TR/webgpu/#enumdef-gpumipmapfiltermode

jamienicol: So that's a constraint in webgpu, but is the important thing here not what (if any) constraints exist in SPIR-V so we can best match the behaviour of the input SPIR-V shader when converting to WGSL? If SPIRV allows floating point values then we're potentially changing the behaviour of the input shader, regardless of whether we round to nearest or just cast (which rounds towards zero?)

FWIW I think tint simply casts: https://dawn.googlesource.com/dawn/+/refs/heads/main/src/tint/lang/spirv/reader/ast_parser/function.cc#5661

I don't think it makes much difference either way so happy to add a round, but just want to check I'm understanding the problem correctly.

teoxoy: > If SPIRV allows floating point values then we're potentially changing the behaviour of the input shader, regardless of whether we round to nearest or just cast (which rounds towards zero?)

That's fair. I found this in the docs of `VK_FORMAT_FEATURE_SAMPLED_IMAGE_FILTER_LINEAR_BIT`:

> If the format being queried is a depth/stencil format, this bit only specifies that the depth aspect (not the stencil aspect) of an image of this format supports linear filtering, and that linear filtering of the depth aspect is supported whether depth compare is enabled in the sampler or not. Where depth comparison is supported it <strong class="purple">may</strong> be linear filtered whether this bit is present or not, but where this bit is not present the filtered value <strong class="purple">may</strong> be computed in an implementation-dependent manner which differs from the normal rules of linear filtering. The resulting value <strong class="purple">must</strong> be in the range <span class="eq">[0,1]</span> and <strong class="purple">should</strong> be proportional to, or a weighted average of, the number of comparison passes or failures.

It sounds like depth textures can support linear filtering with non-comparison samplers in Vulkan. So yes, there doesn't seem to be a way to preserve the same semantics of SPIR-V & Vulkan. This is only a problem if naga is used with something else other than wgpu since it would be a validation error in wgpu.

I'm not sure what the best solution is here, maybe we leave things as they are and if someone runs into this we can consider other solutions then. If you think so too, resolve this conversation and the PR will land (I will approve the PR and enable auto-merge).

teoxoy (approved):

naga/src/front/spv/image.rs
jamienicol: That sounds like a reasonable approach to me. Thanks!

PR #6503, opened 2024-11-08
Isolate WGSL-specific concepts governing diagnostic filters

ErichDonGubler:
**Connections**

Resolves #6485.

**Description**

\-

**Testing**

If this compiles, I consider it tested. 🙂

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `taplo format`.
- [x] Run `cargo clippy`. If applicable, add:
  - [x] `--target wasm32-unknown-unknown`
  - [x] `--target wasm32-unknown-emscripten`
- [x] Run `cargo xtask test` to run tests.
- [x] ~~Add change to `CHANGELOG.md`. See simple instructions inside file.~~ No user-facing changes, so none needed.

Issue #6485, opened 2024-11-01
`naga/src/diagnostic_filter.rs` is in language-independent code, but defines WGSL syntax


jimblandy:
I'm concerned that `src/diagnostic_filter.rs` is where we define the WGSL names of `FilterableTriggeringRule`s. Specifically, `FilterableTriggeringRule`'s `from_ident` and `to_ident` rules are WGSL-specific.

Ideally, Naga IR is a language-independent representation of a shader module. Certainly its semantics are tailored for the languages we want to represent, but the specific spelling of Naga IR features should be localized to front and back ends. This principle isn't followed perfectly by existing code, but it's what we're aiming for.

It seems right to put `FilterableTriggeringRule` in language-independent code, like `src/diagnostic_filter.rs`. But the `from_ident` and `to_ident` methods probably belong somewhere in `src/front/wgsl`, and should be named `from_wgsl` and `to_wgsl`.

Moving that code to the WGSL front end would also make the `#[cfg(feature = "wgsl-in")]` conditional on `tracking_issue_num` unnecessary.

When we get to the point of propagating diagnostic filters into WGSL output, then we may have reached the point of needing a home for WGSL information used by both the front and back ends. #6443 also wants such a home, for holding the WGSL spellings of Naga IR's builtin functions.

cc @ErichDonGubler

jimblandy:
This concern also arises in #6148.

ErichDonGubler:
@jimblandy:

> This concern also arises in #6148.

I have a PR open for this issue (https://github.com/gfx-rs/wgpu/pull/6503), and I have conflict resolution already ready to go between this issue and #6148. So, should be good there.

ErichDonGubler:
Resolved by <https://github.com/gfx-rs/wgpu/pull/6533>, which #6148's review got consolidated into.

PR #6533, opened 2024-11-12
Emit unimplemented errors for several `@diagnostic(…)` attribute parse sites

ErichDonGubler:
**Connections**

- Hard dependency and based on <https://github.com/gfx-rs/wgpu/pull/6503>.

**Description**

\-

**Testing**

- I've tested this manually, and plan on following up with actual test coverage in a separate PR.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `taplo format`.

### DIFF
--- a/naga/src/front/wgsl/error.rs
+++ b/naga/src/front/wgsl/error.rs
@@ -1028,7 +1028,7 @@ impl<'a> Error<'a> {
                 ParseError {
                     message: format!(
                         "found conflicting `diagnostic(…)` rule(s) for `{}`",
-                        triggering_rule.to_ident()
+                        triggering_rule.to_wgsl_ident()
                     ),
                     labels: vec![
                         (first_span, "first rule".into()),
--- a/naga/src/front/wgsl/error.rs
+++ b/naga/src/front/wgsl/error.rs
@@ -296,6 +296,14 @@ pub(crate) enum Error<'a> {
         severity_control_name_span: Span,
     },
     DiagnosticDuplicateTriggeringRule(ConflictingDiagnosticRuleError),
+    DiagnosticAttributeNotYetImplementedAtParseSite {
+        site_name_plural: &'static str,
+        spans: Vec<Span>,
+    },
+    DiagnosticAttributeNotSupported {
+        on_what_plural: &'static str,
+        spans: Vec<Span>,
+    },
 }
 
 impl<'a> From<ConflictingDiagnosticRuleError> for Error<'a> {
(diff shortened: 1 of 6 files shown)


naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2324,36 +2350,55 @@ impl Parser {
...
         self.push_rule_span(Rule::Attribute, lexer);
         while lexer.skip(Token::Attribute) {
-            match lexer.next_ident_with_span()? {
-                ("binding", name_span) => {
+            let (name, name_span) = lexer.next_ident_with_span()?;
+            if let Some(DirectiveKind::Diagnostic) = DirectiveKind::from_ident(name) {
+                if let Some(filter) = self.diagnostic_filter(lexer)? {
+                    let span = self.peek_rule_span(lexer);
+                    diagnostic_filters.add(filter, span)?;
+                }
```
teoxoy, line 2373: I think this needs to break out of the loop or else it reaches `_ => return Err(Error::UnknownAttribute(name_span))`.

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2407,12 +2452,16 @@ impl Parser {
                 return Err(Error::DirectiveAfterFirstGlobalDecl { directive_span });
             }
             (Token::Word("struct"), _) => {
+                ensure_no_diag_attrs("`struct`s", diagnostic_filters)?;
```
teoxoy, line 2459: Github doesn't let me comment above but: Doesn't the line `(Token::Separator(';'), _) => None,` also need `ensure_no_diag_attrs`?

It might be better to return an additional bool from the match statement so that we don't forget to think about this in the future.

naga/src/front/wgsl/error.rs
```diff
@@ -1043,6 +1056,50 @@ impl<'a> Error<'a> {
...
+                message: format!(
+                    "`@diagnostic(…)` attribute(s) {on_what_plural}, which are not supported"
+                ),
+                labels: spans
+                    .iter()
+                    .cloned()
+                    .map(|span| (span, "".into()))
+                    .collect(),
+                notes: vec![
+                    "`@diagnostic(…)` attributes are only permitted on `fn`s and statements."
```
teoxoy:
```suggestion
                    "`@diagnostic(…)` attributes are only permitted on `fn`s, some statements and switch/loop body."
```

Do we want to list supported statement kinds?

teoxoy (changes requested):

naga/src/front/wgsl/parse/mod.rs
ErichDonGubler: Ah, yep, you're right. I missed this case.

Fixed with the force-pushed f9ac43af.

naga/src/front/wgsl/parse/mod.rs
ErichDonGubler: I actually had discovered this with testing right after I submitted this PR, but hadn't had a chance to re-submit this. Good catch!

Fixed with the force-pushed f9ac43af.

naga/src/front/wgsl/error.rs
ErichDonGubler: Just force-pushed e5120a3ae98dacabc2d54d1e27af293fe687888d incorporating your suggestion, with some tweaks to noun plurality (s/body/bodies), punctuation (added an Oxford comma), and styling (backticks).

naga/src/front/wgsl/parse/mod.rs
```diff
@@ -2402,17 +2448,24 @@ impl Parser {
         // read item
         let start = lexer.start_byte_offset();
         let kind = match lexer.next() {
-            (Token::Separator(';'), _) => None,
+            (Token::Separator(';'), _) => {
+                ensure_no_diag_attrs("`struct`s", diagnostic_filters)?;
```
teoxoy:
```suggestion
                ensure_no_diag_attrs("`;`s", diagnostic_filters)?;
```

naga/src/front/wgsl/parse/mod.rs
teoxoy: Thoughts on returning a bool?

teoxoy (approved):

naga/src/front/wgsl/parse/mod.rs
ErichDonGubler: Ah, yikes, sorry, trying to operate too quickly. 😓 Fixed to say `semicolons` and added grammar tweaks to the error messages being added with 9141923e5b15dbec2e8b8cacc96c92b0ba170ee3.

This diagnostic is weird, because there might be a case where somebody _just_ did `@diagnostic(…);`. It'd be nice to offer advice, indicating that they may have wanted a directive instead.

This is a bit complex to handle in the face of other parsed attributes, so I won't do that here, but I'll file a follow-up issue: <https://github.com/gfx-rs/wgpu/issues/6536>

naga/src/front/wgsl/parse/mod.rs
ErichDonGubler: I'm not sure what you're suggesting, but I can imagine returning `let diag_attrs_disallowed_case: Option<&str> = match { /* … */ };`. I don't think that this is better than calling `ensure_no_diag_attrs` in each branch, though, because each branch has its own parsing code after the check. IOW, moving the check to be after the `match` would defer errors on earlier spans of the code until after parsing of, e.g., the `struct` body was done.

ErichDonGubler:
Just added this PR to the existing `Major changes` entry for diagnostic filters in the `CHANGELOG` with 4b99d667d8cb481175562bf725d0f313404cb5c6.

ErichDonGubler:
@teoxoy: Did you review the first commits in this PR that are also in #6503? If so, I can close that PR, noting that you already reviewed those commits.

teoxoy:
Yeah, it looks consistent with Jim's request in https://github.com/gfx-rs/wgpu/issues/6485.

PR #6535, opened 2024-11-13
don't set autoresizing mask on ios as that's not supported

mockersf:
**Connections**
https://github.com/bevyengine/bevy/issues/16363

**Description**
After updating to wgpu 23, Bevy crashes on iOS with `fatal runtime error: Rust cannot catch foreign exceptions`

This started with https://github.com/gfx-rs/wgpu/pull/6107

By adding logging between all lines of code in the `get_metal_layer`, it came from
https://github.com/gfx-rs/wgpu/blob/ae6c6fbea4ee9dff911961512f128cfbfadb02b9/wgpu-hal/src/metal/surface.rs#L226

new_layer is a `CAMetalLayer` which is available on iOS https://developer.apple.com/documentation/quartzcore/cametallayer
but `autoresizingMask ` is only available on macOS https://developer.apple.com/documentation/quartzcore/calayer/1410877-autoresizingmask

this PR stops setting the `autoresizingMask` on iOS

**Testing**

I tested the same fix on the 23 tag with Bevy and it works

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `taplo format`.

### DIFF
--- a/wgpu-hal/src/metal/surface.rs
+++ b/wgpu-hal/src/metal/surface.rs
@@ -207,23 +207,26 @@ impl super::Surface {
             let new_layer: *mut Object = msg_send![class!(CAMetalLayer), new];
             let () = msg_send![root_layer, addSublayer: new_layer];
 
-            // Automatically resize the sublayer's frame to match the
-            // superlayer's bounds.
-            //
-            // Note that there is a somewhat hidden design decision in this:
-            // We define the `width` and `height` in `configure` to control
-            // the `drawableSize` of the layer, while `bounds` and `frame` are
-            // outside of the user's direct control - instead, though, they
-            // can control the size of the view (or root layer), and get the
-            // desired effect that way.
-            //
-            // We _could_ also let `configure` set the `bounds` size, however
-            // that would be inconsistent with using the root layer directly
-            // (as we may do, see above).
-            let width_sizable = 1 << 1; // kCALayerWidthSizable
-            let height_sizable = 1 << 4; // kCALayerHeightSizable
-            let mask: c_uint = width_sizable | height_sizable;
-            let () = msg_send![new_layer, setAutoresizingMask: mask];
+            #[cfg(not(target_os = "ios"))]
+            {
+                // Automatically resize the sublayer's frame to match the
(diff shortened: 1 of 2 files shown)


Wumpf:
great find, thank you!
@madsmtm do you think we need to do something else on iOS instead?

wgpu-hal/src/metal/surface.rs
```diff
@@ -207,23 +206,26 @@ impl super::Surface {
...
-            // desired effect that way.
-            //
-            // We _could_ also let `configure` set the `bounds` size, however
-            // that would be inconsistent with using the root layer directly
-            // (as we may do, see above).
-            let width_sizable = 1 << 1; // kCALayerWidthSizable
-            let height_sizable = 1 << 4; // kCALayerHeightSizable
-            let mask: c_uint = width_sizable | height_sizable;
-            let () = msg_send![new_layer, setAutoresizingMask: mask];
+            #[cfg(not(target_os = "ios"))]
```
madsmtm: It's only available on macOS (and Mac Catalyst):
```suggestion
            #[cfg(target_os = "macos")]
```

madsmtm (approved):
Ah damn, I think I relied on testing on Mac Catalyst a little bit too much, and then only tested previous iterations (but not the final PR) in the simulator - sorry!

>  @madsmtm do you think we need to do something else on iOS instead?

The details are a bit out of my cache here, but yes, we _should_ be doing something different on iOS - but that's what https://github.com/gfx-rs/wgpu/pull/6210 is for.

In total, with https://github.com/gfx-rs/wgpu/pull/6107 and this PR, we're not _worse_ off regarding resizing on iOS (it didn't work before). So 👍 from me.

wgpu-hal/src/metal/surface.rs
mockersf: ha I hesitated in which to use 🙂

Wumpf (approved):

Wumpf:
Backported to v23 here without PR https://github.com/gfx-rs/wgpu/commit/78fbdfc0a25e042947be67eae769565bb4a114a0

PR #6537, opened 2024-11-13
fix(wgsl-in): include user and unknown rules in `diagnostic(…)` tracking

ErichDonGubler:
**Connections**

- Contributes (but does not fully resolve) #5320.

**Description**

Prior to this PR, unknown and user `diagnostic(…)` triggering rules for filters would get ignored, and certain important validations like this were not rejecting the program properly:

```wgsl
fn myfunc() {
	if (true) @diagnostic(off, my.lint) {
		//    ^^^^^^^^^^^^^^^^^^^^^^^^^ not yet supported, should report an error
	}
}
```

```
fn myfunc() {
	if (true) @diagnostic(off, wtf) {
		//    ^^^^^^^^^^^^^^^^^^^^^ should emit a warning
	}
}
```

```wgsl
diagnostic(off, my.lint);
diagnostic(warning, my.lint); // should report a conflict error
```

```wgsl
diagnostic(off, wat_is_this);
diagnostic(warning, wat_is_this); // should report a conflict error
```

**Testing**

- Tests have been added for the above cases, and a bit more. N.B. that operation and validation coverage is missing for standard rule cases still.

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `taplo format`.

### DIFF
--- a/naga/src/diagnostic_filter.rs
+++ b/naga/src/diagnostic_filter.rs
@@ -54,7 +54,7 @@ impl Severity {
 /// A filterable triggering rule in a [`DiagnosticFilter`].
 ///
 /// <https://www.w3.org/TR/WGSL/#filterable-triggering-rules>
-#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
 #[cfg_attr(feature = "serialize", derive(Serialize))]
 #[cfg_attr(feature = "deserialize", derive(Deserialize))]
 #[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
@@ -140,7 +140,7 @@ impl DiagnosticFilterMap {
             triggering_rule,
         } = diagnostic_filter;
 
-        match diagnostic_filters.entry(triggering_rule) {
+        match diagnostic_filters.entry(triggering_rule.clone()) {
             Entry::Vacant(entry) => {
                 entry.insert((new_severity, span));
             }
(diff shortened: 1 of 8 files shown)


naga/src/diagnostic_filter.rs
```diff
@@ -107,51 +121,66 @@ impl DiagnosticFilterMap {
...
+                    if first_severity != new_severity {
+                        return Err(ConflictingDiagnosticRuleError {
+                            triggering_rule,
+                            triggering_rule_spans: [first_span, span],
+                        });
+                    }
+                }
+            },
+            DiagnosticFilter::User(span) => user_rule_spans.push(span),
+            DiagnosticFilter::Unknown(span) => unknown_rule_spans.push(span),
```
ErichDonGubler: todo: Gotta rebase this on the conflicting directive fix in #6353.

naga/src/diagnostic_filter.rs
```diff
@@ -59,22 +59,35 @@ impl Severity {
 #[cfg_attr(feature = "deserialize", derive(Deserialize))]
 #[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
 pub enum FilterableTriggeringRule {
+    Standard(StandardFilterableTriggeringRule),
+    Unknown(Box<str>),
+    User([Box<str>; 2]),
```
jimblandy: Just a thought: `[Box<str>; 2]` makes `FilterableTriggeringRule` five words long, since `Box<str>` is a fat pointer, and the enum discriminant ends up being expanded to a full word to preserve alignment. But using `Box<String>` and `Box<[String; 2]>` would let `FilterableTriggeringRule` remain two words long, and these are probably easier to work with.

The `Unknown` and `User` cases are ideally going to be rare: anything that does become widespread we'd add support for, with a proper enum variant of its own. So the impact of the double indirection should be low.

jimblandy (approved):
Looks good to me - thanks!

naga/src/diagnostic_filter.rs
ErichDonGubler: Agreed on indirection to minimize stack usage of the common case being worth it. I feel silly I hadn't thought to push the two `User` tokens behind a `Box`, honestly.

ErichDonGubler: Fixed with a force-push to 100b63d4.

Issue #5683, opened 2024-05-09
Push constant of non-struct type fails to compile on DX12


JasondeWolff:
**Description**
Using a `var<push_constant>` of type `mat4x4<f32>` fails to compile on DX12.

**Repro steps**
Wgsl Shader:
``` wgsl
struct VertexOutput {
    @builtin(position) position: vec4<f32>,
};

@group(0)
@binding(0)
var<uniform> view_proj: mat4x4<f32>;

var<push_constant> model: mat4x4<f32>;

@vertex
fn vs_main(
    @location(0) position: vec4<f32>,
    @location(1) normal: vec4<f32>,
    @location(2) tangent: vec4<f32>,
    @location(3) color: vec4<f32>,
    @location(4) tex_coord: vec4<f32>,
    @location(5) joints: vec4<u32>,
    @location(6) weights: vec4<f32>,
) -> VertexOutput {
    var result: VertexOutput;
    result.position = view_proj * model * vec4<f32>(position.xyz, 1.0);
    return result;
}

@fragment
fn fs_main(vertex: VertexOutput) -> @location(0) vec4<f32> {
    return vec4<f32>(1.0, 0.0, 1.0, 1.0);
}
```

Naga Hlsl Output:
``` hlsl
struct NagaConstants {
    int first_vertex;
    int first_instance;
    uint other;
};
ConstantBuffer<NagaConstants> _NagaConstants: register(b2, space1);

struct VertexOutput {
    float4 position : SV_Position;
};

cbuffer view_proj : register(b1, space1) { row_major float4x4 view_proj; }
ConstantBuffer<row_major float4x4> model: register(b0);

struct VertexOutput_vs_main {
    float4 position_1 : SV_Position;
};

struct FragmentInput_fs_main {
    float4 position_2 : SV_Position;
};

VertexOutput_vs_main vs_main(float4 position : LOC0, float4 normal : LOC1, float4 tangent : LOC2, float4 color : LOC3, float4 tex_coord : LOC4, uint4 joints : LOC5, float4 weights : LOC6)
{
    VertexOutput result = (VertexOutput)0;

    float4x4 _expr10 = view_proj;
    float4x4 _expr12 = model;
    result.position = mul(float4(position.xyz, 1.0), mul(_expr12, _expr10));
    VertexOutput _expr18 = result;
    const VertexOutput vertexoutput = _expr18;
    const VertexOutput_vs_main vertexoutput_1 = { vertexoutput.position };
[...]

Issue #5571, opened 2024-04-20
Crash compiling and using HLSL shaders


JMS55:
Source shader: https://paste.rs/CAVpW.txt
Output HLSL and crash: https://paste.rs/nW7t8.txt

Source shader: https://paste.rs/13TBc.txt
Output HLSL and crash: https://paste.rs/A2Vhb.txt

cwfitzgerald:
Yeah we need to wrap scalar push constants in structs in the hlsl backend.

cwfitzgerald:
Closing in favor of #5683

Issue #4502, opened 2023-09-20
Pad push constants if needed


teoxoy:
A variable in the push constant address space might need padding in the front if the range on the API side doesn't start at 0.

See https://github.com/gfx-rs/wgpu/issues/4099#issuecomment-1702428118 for more details.

teoxoy:
```
Validation Error: [ VUID-VkComputePipelineCreateInfo-layout-07987 ] Object 0: handle = 0x980f360000000011, type = VK_OBJECT_TYPE_SHADER_MODULE; Object 1: handle = 0xd5b26f0000000010, type = VK_OBJECT_TYPE_PIPELINE_LAYOUT; | MessageID = 0x92cd2576 | vkCreateComputePipelines(): pCreateInfos[0].stage SPIR-V (VK_SHADER_STAGE_COMPUTE_BIT) has a push constant buffer Block with range [0, 4] which outside the pipeline layout range of [4, 8]. The Vulkan spec states: If a push constant block is declared in a shader, a push constant range in layout must match both the shader stage and range (https://vulkan.lunarg.com/doc/view/1.3.280.0/linux/1.3-extensions/vkspec.html#VUID-VkComputePipelineCreateInfo-layout-07987)
```

I ran into this while trying to test some functionality in https://github.com/gfx-rs/wgpu/pull/5714.

Given that the space in push constants is limited, I don't think we should complicate things by supporting non-0 start of ranges on the API side. The proposal upstream https://github.com/gpuweb/gpuweb/pull/4612 also doesn't have this.

PR #6540, opened 2024-11-13
Fix `set_push_constants` for render bundles

### LINKED ISSUE #2683: Push constants in render bundles are half-implemented
`RenderBundleEncoder::finish` propagates `SetPushConstant` commands from the encoder to the final `RenderBundle`, but the bundle's `BasePass`'s `push_constant_data` vector is always empty.

### LINKED ISSUE #6532: RenderBundleEncoderSetPushConstants always crashes
There appears to be a bug in `RenderBundleEncoder.finish` that prevents `SetPushConstants` from working correctly.  If I understand this bug correctly, `SetPushConstants` has never worked for `RenderBundleEncoder`s.

I get the same failure on both a MacOS and on Linux.

Attached is some Rust code showing the bug.  The same code in run twice, the first time we draw using  a `RenderPass`, and the second time we draw using a`RenderBundleEncoder`, which we then `finish()` and execute on the `RenderPass`.  The drawing commands are:
```
            encoder.set_pipeline(&pipeline);
            encoder.set_push_constants(ShaderStages::VERTEX, 0, data_as_u8);
            encoder.draw(0..4, 0..1);
 ```
It runs fine the on the `RenderPass, and crashes on the `RenderBundleEncoder`.  The crash occurs during the execution of `            render_pass.execute_bundles([&render_bundle]);`

Internally the crash occurs on this [line](https://github.com/gfx-rs/wgpu/blob/4311091cce7e417fa6cf41f94d4199d67959b63c/wgpu-core/src/command/bundle.rs#L1044) of wgpu-core/src/command/bundle.rs.  The code is (correctly) trying to grab the push constants that is supposed to be stored in `self.base.push_constant_data`, but that variable (incorrectly) is empty.

I *believe* the bug is [here](https://github.com/gfx-rs/wgpu/blob/4311091cce7e417fa6cf41f94d4199d67959b63c/wgpu-core/src/command/bundle.rs#L546) on lines 546-547.  The code is copying the `RenderBundleEncoder` to the `RenderBundle`, but the `push_constant_data` is not being copied.

I'm not sure if `string_data` (on the previous line) also needs to be fixed. I am not fluent in Rust or in the internals of wgpu, so I'm not quite sure of the correct way to copy this variable.

[BugReport.zip](https://github.com/user-attachments/files/17722969/BugReport.zip)

### PR DESCRIPTION
**Pull request by fyellin**
Fixes https://github.com/gfx-rs/wgpu/issues/6532.
Fixes https://github.com/gfx-rs/wgpu/issues/2683.

The use of `RenderBundleEncoder.SetPushConstants` would cause a panic on the subsequent call to `RenderPass.ExecuteBundles()`.  The `.finish()` method was failing to copy its `push_constant_data` from the render bundle encoder to the render bundle.

Currently `string_data` is unused, but it seemed safest to copy it to prevent future bugs.

I have confirmed with a toy program that the old code panics and the new code works (and correctly accesses the pushed constants).

### CODE REVIEW

File: wgpu-core/src/command/bundle.rs [hunk 1]
```diff
@@ -543,8 +543,8 @@ impl RenderBundleEncoder {
                 label: desc.label.as_ref().map(|cow| cow.to_string()),
                 commands,
                 dynamic_offsets: flat_dynamic_offsets,
-                string_data: Vec::new(),
-                push_constant_data: Vec::new(),
+                string_data: self.base.string_data.clone(),
```
nical: ~I think that this should be `std::mem::take(&mut self.base.string_data)` (and same thing for the push constant data.~

Edit:

err, actually, just moving the two vectors should work since `self` is dropped at the end of the frame.

fyellin:
`self` is immutable in the function `finish()`  [see line 345], so I cannot
take these vectors from `self`.  Since this is my first time modifying
wgpu-core, I don't feel confident enough to change the signature of the
method and make `self` mutable.

I also just verified that I get a compiler error when I change `self` to
`&mut self` because at least one of the callers to `finish` wasn't
expecting it to be mutable.

I'll rely on your judgment on how to proceed.

fyellin:
I realized the argument is self.  So we own it.  I can just use `self.base.<field>`.  Also added tests to show that things are working now.

fyellin:
My changes fixed one bug, but it seems there is another bug in the Vulkan and DX12 implementation.  For now, I can fix the code so that it works on everything else, but I don't know enough about the internal implementations of Vulkan or DX12.  For Vulcan there is an error message that makes no sense. For DX12, naga seems to produce bad code, and there is a compiler failure.

My code in the test suite is just a basic use of SetPushConstant on a RenderPassEncoder and a RenderBundleEncoder.  The fact that the code is now running successfully on RenderBundleEncoders means that I've fixed the generic bug.   Someone else is going to need to fix the implementation specific bugs, since I don't know about the internals.

teoxoy:
The DX12 issue is https://github.com/gfx-rs/wgpu/issues/5683 / https://github.com/gfx-rs/wgpu/issues/5571, can you wrap the push constants in the shader in a struct? That should get things working.

I think the Vulkan validation error is https://github.com/gfx-rs/wgpu/issues/4502. Can you change the test to not use non-0 start ranges? This will most likely become a validation error emitted by us in the future as the [WebGPU spec proposal](https://github.com/gpuweb/gpuweb/blob/main/proposals/push-constants.md) also doesn't allow users to specify non-0 start ranges.

fyellin:
I've converted the code to use a struct.  The naga compiler does the right thing now.

Since push_constants no longer exists in WGPU or WebGPU, I was using the Vulkan [documentation](https://vkguide.dev/docs/chapter-3/push_constants/) as my guide.  According to those specs, one can separate the vertex shader constants from the fragment shader constants, and specify that a certain byte-range of the constants are are only for the vertex shader or only for the fragment shader.  When doing this, you are also required to call set_push_constants separately for each of the byte-ranges.

So I performed multiple tests.

1. Separate constants for the vertex shader and fragment shader.
    - Both Linux/Vulkan and Windows/DX12 failed.  Linux gave the message that the "set_push_constant" was out of range.  Windows
      gave the wrong answer.

2. Create a single constant pool shared by both shaders.  This seems to be the intent of the spec proposal you mentioned above.  
   - Linux  and Metal passed.  Windows/DX12 ran, but gave the wrong answer.

3. Create a single constant pool, but use two separate set_push_constant operations to separately push the constants intended for the vertex shader and those intended for the fragment shader.
   - The results were the same as previously.  Windows/DX12 ran, but gave the wrong answer.
 
It seems that there are two possibly new bugs:
- Linux/Vulkan cannot handle having the constants separated by stages, as per the Vulkan spec.
-  Windows/DX12 is just broken.
It also seems that there is no problem with using a non-zero offset.  Vulkan's problem is dealing with multiple stages.

I will resubmit my PR using #2 above, and exclude DX12.

fyellin:
I have opened two additional bugs:

BUG #6558.  SetPushConstant on Dx12 gives wrong results
BUG #6562.  Unable to do SetPushConstant on Vulcan with separate Vertex and Fragment Stages

This PR fixes yet a third bug and adds a test that succeeds on GL and Metal and actually tests SetPushConstant.
BUG #6532.  SetPushConstant always crashes.

How do I proceed?  Right now, the test suite isn't even running.

teoxoy:
Thanks for looking into these issues!

I left a comment in https://github.com/gfx-rs/wgpu/issues/6558#issuecomment-2482643043, could you try seeing if avoiding arrays gets things working on DX12?

teoxoy (approved):
Looks good, thank you!

PR #6541, opened 2024-11-13
[wgpu-core] Replace usage of `PreHashedMap` with `FastHashMap`

### LINKED ISSUE #6373: Uses of `PreHashedMap` are probably not sound
The `PreHashedMap` type in `wgpu_core` is probably not used in a sound way, in particular, its [use in `Device::derive_pipeline_layout`](https://github.com/gfx-rs/wgpu/blob/ee0d1703e5f4a267ce9b87d50b824190b45b5a79/wgpu-core/src/device/resource.rs#L2605) is almost certainly incorrect.

The idea is that a `PreHashedMap` key is just a copy of the true key's `u64` hash value. But this doesn't provide any way to distinguish keys that have the same hash value: `PartialEq` on a `PreHashedKey` just [compares the cached hash values](https://github.com/gfx-rs/wgpu/blob/ee0d1703e5f4a267ce9b87d50b824190b45b5a79/wgpu-core/src/hash_utils.rs#L41-L45). This will lead to incorrect lookups and insertions.

(I don't think `PreHashedMap` makes much sense: it's only sound if you can guarantee a distinct hash value for distinct keys, which means these "hash values" are almost certainly not produced by hashing: you're using some identifier or pointer that was assigned in a way that guarantees uniqueness. In that case, you should simply use that identifier as the key directly.)

jamienicol:
**Connections**
Fixes #6373 

**Description**
PreHashedMap was introduced as a performance optimization. It is, however, potentially unsound as it cannot distinguish between keys that are not equal yet have the same hash. This patch replaces its usage with a simple FastHashMap, which should be good enough. If this is found to cause real life performance issues down the line then we can figure out a better solution.

**Testing**
cargo xtask test passes

**Checklist**

- [x] Run `cargo fmt`.
- [x] Run `taplo format`.


ErichDonGubler (approved):
LGTM!

ErichDonGubler:
@gfx-rs/wgpu: I'm not sure if we need a `CHANGELOG` entry here. Might be worth nothing as a `General` bug fix? 🤔

Wumpf:
yeah would be nice to have it listed as a bugfix!

jamienicol:
Added the changelog entry (and consolidated the duplicated `General` categories while I was there)

ErichDonGubler:
Just needs conflicts resolved, and auto-merge should do the rest.

jamienicol:
> Just needs conflicts resolved, and auto-merge should do the rest.

done

> and consolidated the duplicated General categories while I was there

heh, this was already done I just hadn't rebased

